PR #52, opened 2026-05-19
fix(group-chat): create group first when starting group chat from private chat

a652:
## Summary

Fix the 400 error when starting a group chat from the private chat three-dots menu. The action now correctly creates a new group first before navigating to it, instead of incorrectly trying to add members to a non-existent group.

## Related Issue

Fixes #51

## Changes

- In `onOK()` AddMember branch, added a `channelType` check:
  - `ChannelTypePerson`: build uid list `[self, otherPerson, ...selected]`, call `createChannel()`, then navigate to the new group via `showConversation()`
  - Group channel: keep the original `addSubscribers()` logic (unchanged)
- Added focused unit coverage for both AddMember branches:
  - Private chat creates a group first and navigates to it
  - Existing group chat still calls `addSubscribers()`

## Testing

- [x] Unit tests added/updated
- [x] Manually verified

**Unit test:**

```bash
npx pnpm@10.32.0 --filter @octo/contacts exec vitest run src/Organizational/GroupNew/index.test.tsx
```

Result: 1 test file passed, 2 tests passed.

**Manual verification steps:**
1. Open any private chat conversation
2. Click ⋯ → Start Group Chat
3. Select one or more members → confirm
4. Group chat is created successfully and UI navigates to it ✅

Previously: `POST /groups/{userUID}/members` → 400 Bad Request
After fix: `POST /group/create` → success → navigate to new group

Local verification evidence is posted in the PR comments.

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [x] Added tests for my changes
- [x] Followed commit message conventions (Conventional Commits)

PR #622, opened 2026-07-10
fix(whiteboard): reject structurally-invalid fractional-index keys and harden restore

ploy-elison:
## Summary

A whiteboard element's `index` is an Excalidraw jitterbug / fractional-indexing **order key**, not merely a base62 string. `isValidIndex` only validated the character set (`/^[A-Za-z0-9]+$/`), so a synthetic key like `r00000003` — head `r` declares a **19-char** integer part while the key is only **9 chars** — passed as valid, reached `updateScene`, and crashed the canvas.

This PR does the FE half of the fix (schema struct-check + restore depth defence). The backend main fix (stop producing the invalid keys at the source, in repair `fillIndexKey`, + migration) is tracked separately.

## Changes

**②-FE struct-check — `packages/whiteboard-schema/src/normalize.ts`**
- Replaces the charset regex with a pure boolean port of `fractional-indexing`'s `validateOrderKey` (no runtime dependency added; the package stays framework-free). `isValidIndex("r00000003") === false`, and `normalizeElement` strips an invalid `index` to a clean absent state (`normalize.ts` `isValidIndex` / the `delete out.index` line).

**④ FE depth defence — `packages/docs/src/board/BoardShell.tsx`**
- Both restore call sites now pass `{ normalizeIndices: true }` so Excalidraw re-indexes recoverable keys.
- The `initialData` restore runs **during render**, outside the collab binding's guarded `applyRemote` and outside `BoardErrorBoundary` (which only wraps the `<Excalidraw>` subtree). It is now wrapped in try/catch: an unrecoverable key degrades to an empty initial scene so the canvas still mounts (the later observe→applyRemote path repaints), instead of a render throw tearing down the whole shell. The `applyRemote` path already caught restore throws and keeps the last good scene.

This is **depth defence, not a root fix**: `normalizeIndices` / `syncInvalidIndices` / `generateNKeysBetween` themselves throw on a structurally-invalid *existing* key, so the FE cannot repair such a key — it can only avoid crashing. The root fix is the backend ceasing to persist invalid keys.

## Adopted legality rule (for cross-repo alignment)

The backend single-issue had not written back its rule when this was implemented, so this PR **defines** the rule; the backend `normalize.ts` must match it byte-for-byte. A key is valid iff **all** hold:

1. non-empty and every char is in the base62 alphabet `0-9A-Za-z`;
2. it is not the reserved `SMALLEST_INTEGER` sentinel (`A00000000000000000000000000`);
3. the head char is `a`..`z` or `A`..`Z`, and the integer-part length it declares (`a`/`Z` → 2 … `z`/`A` → 27) fits inside the key;
4. the fractional part (everything after the integer part) does not end in the first digit (`0`).

This is exactly `fractional-indexing`'s `validateOrderKey` expressed as a non-throwing boolean.

## Testing (red → green)

- `packages/whiteboard-schema/src/normalize.test.ts` — new. On the old code 4 assertions fail (incl. `isValidIndex("r00000003")` and the `normalizeElement` strip); green after the fix.
- `packages/docs/src/board/__tests__/BoardShellRestoreDefense.test.tsx` — new. Seeds a cold-reopen scene carrying `r00000003`, with a restore stub that throws on it (as the real library does); the canvas fails to mount on the old code and mounts (degraded) after the fix.
- `pnpm vitest run` — whiteboard-schema 12/12; `packages/docs` `src/board` 176/176.
- `tsc` — no new errors in the touched files (the schema package typechecks clean).

## Notes

Draft PR — held open for review + deploy + QA; do not merge or mark ready yet.

### DIFF
--- a/packages/docs/src/board/BoardShell.tsx
+++ b/packages/docs/src/board/BoardShell.tsx
@@ -99,6 +99,7 @@ type ExcalidrawComponent = ComponentType<ExcalidrawProps>
 type RestoreElementsFn = (
   elements: readonly unknown[] | null | undefined,
   localElements: readonly unknown[] | null | undefined,
+  opts?: { refreshDimensions?: boolean; repairBindings?: boolean; normalizeIndices?: boolean },
 ) => ExcalidrawElement[]
 type ReconcileElementsFn = (
   localElements: readonly unknown[],
@@ -828,7 +829,12 @@ export function BoardShell(props: BoardShellProps): ReactElement {
     if (restore && reconcile) {
       const imperative = api as { getAppState?: () => unknown }
       binding.setRenderAdapter({
-        restore: (remote) => restore(remote, null),
+        // XIN-795 ④ depth defence: ask Excalidraw to re-index on restore so a recoverable
+        // out-of-order key is normalised rather than thrown on. A structurally-invalid key that
+        // even normalizeIndices cannot repair still throws, but that throw is caught by
+        // `applyRemote` (collab/binding.ts), which keeps the last good scene instead of blanking
+        // the canvas — the root fix for invalid persisted keys lives in BE repair (XIN-794).
+        restore: (remote) => restore(remote, null, { normalizeIndices: true }),
         reconcile: (local, restoredRemote) => reconcile(local, restoredRemote, imperative.getAppState?.()),
       })
     }
@@ -1011,7 +1017,18 @@ export function BoardShell(props: BoardShellProps): ReactElement {
       if (docEls.length > 0) raw = docEls
     }
     const restore = restoreElementsRef.current
-    return restore ? restore(raw, null) : [...raw]
+    if (!restore) return [...raw]
+    // XIN-795 ④ depth defence: this restore runs during render (not inside the collab binding's
+    // guarded applyRemote), so a throw on a structurally-invalid persisted key — one the BE repair
+    // (XIN-794) failed to stop at source — would take down the whole BoardShell render, not just a
+    // frame. `normalizeIndices` lets Excalidraw re-index recoverable keys; the try/catch degrades an
+    // unrecoverable throw to an empty initial scene so the canvas still mounts and the later
+    // observe→applyRemote path can repaint, with BoardErrorBoundary as the final backstop.
+    try {
+      return restore(raw, null, { normalizeIndices: true })
+    } catch {
+      return []
+    }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [Excalidraw, collabSession, accessConfirmed])
 
--- a/packages/docs/src/board/__tests__/BoardShellRestoreDefense.test.tsx
+++ b/packages/docs/src/board/__tests__/BoardShellRestoreDefense.test.tsx
@@ -1,0 +1,134 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { render, screen, cleanup } from '@testing-library/react'
+import type { ReactNode } from 'react'
+import type { BoardTerminal } from '../collab/index.ts'
+import type { WhiteboardSession } from '../collab/connect.ts'
+import { setWKApp } from '../../octoweb/index.ts'
+import { createMockWKApp } from '../../octoweb/mock.ts'
+
+// XIN-795 ④ FE depth defence. `BoardShell.initialElements` restores the initially-loaded / cold-reopen
+// scene DURING RENDER, outside the collab binding's guarded applyRemote path and outside the
+// BoardErrorBoundary (which only wraps the <Excalidraw> subtree). A structurally-invalid persisted
+// order key (e.g. the synthetic `r00000003` a since-fixed BE repair produced) makes Excalidraw's real
+// `restoreElements` THROW. Before the fix that throw propagated straight out of BoardShell and tore
+// down the whole host tree; the fix wraps the restore in try/catch and degrades to an empty initial
+// scene so the canvas still mounts (the later observe→applyRemote path can repaint).
+//
+// This Excalidraw stand-in reproduces that throw: restoreElements rejects any element whose `index`
+// fails the jitterbug structural rule, exactly as the real library does.
+function isJitterbugValid(index: unknown): boolean {
+  if (typeof index !== 'string' || !/^[0-9A-Za-z]+$/.test(index)) return false
+  const head = index.charAt(0)
+  let intLen = 0
+  if (head >= 'a' && head <= 'z') intLen = head.charCodeAt(0) - 97 + 2
+  else if (head >= 'A' && head <= 'Z') intLen = 90 - head.charCodeAt(0) + 2
+  return intLen > 0 && intLen <= index.length
+}
+
+vi.mock('@excalidraw/excalidraw', async () => {
+  const { useEffect } = await import('react')
+  const api = { updateScene: () => {}, getAppState: () => ({}), updateLibrary: async () => [] }
+  const Excalidraw = ({
+    children,
+    excalidrawAPI,
+  }: {
+    children?: ReactNode
+    excalidrawAPI?: (api: unknown) => void
+  }) => {
+    useEffect(() => {
+      excalidrawAPI?.(api)
+    }, [excalidrawAPI])
+    return <div data-testid="excalidraw-canvas">{children}</div>
+  }
+  const MainMenu = (() => null) as unknown as { DefaultItems: Record<string, unknown> }
+  MainMenu.DefaultItems = {}
+  return {
+    Excalidraw,
+    MainMenu,
+    restoreElements: (els: readonly unknown[] | null | undefined) => {
+      const arr = els ? [...els] : []
+      for (const el of arr) {
+        // Mirror the real jitterbug throw on a structurally-invalid `index`.
+        if (el && typeof el === 'object' && !isJitterbugValid((el as { index?: unknown }).index)) {
+          throw new Error('invalid order key: ' + String((el as { index?: unknown }).index))
+        }
+      }
+      return arr
+    },
+    reconcileElements: (local: readonly unknown[]) => [...local],
+    loadLibraryFromBlob: async () => [],
+    serializeLibraryAsJSON: () => '[]',
+  }
+})
+vi.mock('@excalidraw/excalidraw/index.css', () => ({}))
+vi.mock('../../members/useMemberNames.ts', () => ({ useMemberNames: () => new Map<string, string>() }))
+
+// Imported AFTER the mocks so the mocked modules are in place.
+import { BoardShell } from '../BoardShell.tsx'
+
+function makeAwareness() {
+  return {
+    clientID: 1,
+    getStates: () => new Map(),
+    setLocalStateField: () => {},
+    on: () => {},
+    off: () => {},
+  }
+}
+
+/** Admin session whose Y.Doc snapshot carries an element with a structurally-invalid order key. */
+function makeSessionWithInvalidIndex(): WhiteboardSession {
+  const binding = {
+    setApi: () => {},
+    setRenderAdapter: () => {},
+    setFileSync: () => {},
+    handleLocalChange: () => {},
+    // Cold-reopen seed for BoardShell.initialElements: a single invalid-index element.
+    snapshotElements: () => [{ id: 'bad', type: 'rectangle', index: 'r00000003' }] as unknown[],
+  }
+  return {
+    getRole: () => 'admin',
+    subscribeRole: () => () => {},
+    subscribeTerminal: (_cb: (t: BoardTerminal) => void) => () => {},
+    binding,
+    provider: { awareness: makeAwareness(), isSynced: true, on: () => {}, off: () => {} },
+  } as unknown as WhiteboardSession
+}
+
+let wk: ReturnType<typeof createMockWKApp>
+
+beforeEach(() => {
+  wk = createMockWKApp()
+  setWKApp(wk)
+})
+
+afterEach(() => {
+  cleanup()
+  vi.restoreAllMocks()
+})
+
+describe('BoardShell restore depth defence (XIN-795 ④)', () => {
+  it('degrades to an empty initial scene instead of crashing the shell when restore throws on an invalid key', async () => {
+    wk.apiClient.responder = (method: string, url: string) => {
+      if (method === 'get' && url === '/docs/board-1') {
+        return { data: { docId: 'board-1', title: 'Board', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(
+      <BoardShell
+        docId="board-1"
+        title="Board"
+        space="s1"
+        collabSession={makeSessionWithInvalidIndex()}
+        collab
+      />,
+    )
+
+    // The canvas mounts: BoardShell caught the restore throw and fed Excalidraw an empty scene rather
+    // than letting the throw unmount the host tree. Before the fix this render threw and no canvas
+    // appeared.
+    expect(await screen.findByTestId('excalidraw-canvas')).toBeTruthy()
+  })
+})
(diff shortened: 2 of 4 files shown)


OctoBoooot (approved):
## Review: reject structurally-invalid fractional-index keys and harden restore (#622)

**Verdict:** Approve (draft — held open per the author, not for merge)
Precise root-cause fix: the old `isValidIndex` only charset-checked, so a structurally-invalid order key (`r00000003`: head `r` declares a 19-char integer part in a 9-char key) passed, reached `updateScene`, and crashed the canvas. The new struct-check + restore depth-defence are both correct; verified the validation port against `fractional-indexing` semantics and constructed the boundary inputs.

Risk tier: normal. Base `main`; draft + `needs-human-review` (size/L) + `mergeStateStatus=BLOCKED` — not for merge yet, by the author's own note.

### What I verified
- **Validation port (the swap-one-check-for-another risk):** `orderKeyIntegerLength` maps `a..z → 2..27` (`c-97+2`) and `A..Z → 27..2` (`90-c+2`) — matches `getIntegerLength` exactly. Boundaries constructed and confirmed: `r00000003` → intLen 19 > len 9 → rejected (the crash key); `a` → intLen 2 > len 1 → rejected; `a00` → fraction ends in `0` → rejected; SMALLEST_INTEGER sentinel → rejected; valid generated keys `a0`/`a1`/`a0V`/`Zz`/`a0G` → accepted. No false-accepts.
- **Stricter-but-safe deviation:** the port adds a whole-string base62 gate the reference `validateOrderKey` doesn't have. That only *rejects* more (non-base62 chars), never accepts more — safe, and preserves the old charset guarantee.
- **Restore depth-defence:** the render-time `initialData` restore genuinely runs outside `applyRemote`'s guard and outside `BoardErrorBoundary` (which wraps only `<Excalidraw>`), so an uncaught throw there would tear down the whole shell — the try/catch → `[]` degrade is the correct minimal fix, and the later observe→applyRemote path repaints. `{ normalizeIndices: true }` added at both call sites; `RestoreElementsFn` type widened accordingly.
- **Tests are adversarial and pin the exact boundaries** (`r…` family rejected, real keys accepted, `a`/`a00`/sentinel rejected, `normalizeElement` strips invalid + keeps valid) — red on old code per the description, green after.

### Praise
- The fix resists the classic swap-a-validator trap: it doesn't just reject the one crash key, it ports the real library's rule and the tests pin both the integer-length boundary (`a` too short) and the inverse trailing-zero rule (`a00`) — the two places a hand-rolled order-key check usually gets wrong.
- Honest scoping: the PR states plainly this is depth-defence not a root fix (FE can't repair a structurally-invalid *existing* key, only avoid crashing; root fix is BE ceasing to persist them) — no over-claim.

### Note
Draft, and the author asks it be held for review + deploy + QA. Approving the code; do not merge until marked ready and the BE root fix (XIN-794) lands in step.

Jerry-Xin (approved):
**APPROVE** — the validator now matches the real fractional-indexing spec byte-for-byte, and the restore hardening degrades safely per-element (strip) with a contained render-path backstop. No valid content is reordered, the check stays off the hot path, and it holds up against adversarial input.

### What changed (4 files)
- `packages/whiteboard-schema/src/normalize.ts` — replaces the bare charset check `/^[A-Za-z0-9]+$/` with `isValidOrderKey`, a pure boolean port of `fractional-indexing`'s `validateOrderKey`: non-empty base62, not the `SMALLEST_INTEGER` sentinel, head char declares an integer-part length (2..27) that fits the key, and the fraction may not end in `'0'`. `isValidIndex` now delegates to it.
- `packages/docs/src/board/BoardShell.tsx` — widens `RestoreElementsFn` with an `opts` arg; passes `{ normalizeIndices: true }` on both restore call sites; wraps the render-time `initialElements` restore in try/catch → `[]`.
- `packages/whiteboard-schema/src/normalize.test.ts` (new) — validator unit tests.
- `packages/docs/src/board/__tests__/BoardShellRestoreDefense.test.tsx` (new) — render-path restore-throw degradation test.

### Validator correctness (spec-parity verified)
I reimplemented the real `fractional-indexing.validateOrderKey` and diffed it against `isValidOrderKey` over the boundary set. **Every case agrees** (neither too-strict nor too-loose):
- Rejects the whole synthetic `^r[0-9a-z]{8}$` family (`r00000003`, `r00000000`, `rzzzzzzzz`, `r12345678`) — head `r` declares a 19-char integer part in a 9-char key → `intLen(19) > key.length(9)`.
- Accepts real generated keys: `a0`, `a1`, `a0V`, `Zz` (negative-integer upper-case region), `a0G`, `b00`.
- Rejects the `SMALLEST_INTEGER` sentinel, empty string, non-base62 (`a0!`), short-integer heads (`a`), and fractions ending in `'0'` (`a00`, `Zz0`).
- Minimal key `a0` ✓; long fractions (`a0` + 1000 chars) ✓; `Z0` ✓; `A0` correctly rejected (head `A` → intLen 27 > len 2).

### Restore fail-safe direction (per-element, not board-wide)
- `normalizeElement` strips only the offending field (`if (!isValidIndex(out.index)) delete out.index`) and keeps the element; `repairElements` then reassigns a legal key across the whole set. One bad key does **not** drop the element or nuke the board.
- Render-path `BoardShell.initialElements` runs outside the collab binding's guarded `applyRemote` and outside `BoardErrorBoundary`, so an unrecoverable throw there would previously tear down the host tree. The new try/catch degrades to an empty **initial** scene so the canvas still mounts, and the later `observe→applyRemote` repaints. Degrading the cold seed to empty (vs. crashing the shell) is the correct direction; `applyRemote` itself already keeps the last-good scene on throw (`binding.ts` catch → `remoteApplyErrors++`).

### Ordering stability
Valid keys are never rewritten — only structurally-invalid indices are stripped — so the relative order of valid elements is preserved. ✓

### Hot-path / perf
`isValidIndex` is reached only through `normalizeElement` → `repairForRender`, i.e. on restore/remote-apply, not per render frame or per comparator call. `repairForRender` is an O(n) two-pass. Validation itself is O(key length); 100k validations of 200-char keys ran in ~25ms locally — no quadratic behavior, so a crafted many-invalid-keys / long-keys payload does not create a DoS window during restore. ✓

### Tests
Cover both directions: invalid-key rejection (incl. the exact `r00000003` regression + the full family), valid-key acceptance, non-string/empty/sentinel edges, `normalizeElement` strip-to-absent for invalid + keep-untouched for valid, and the render-path degradation (`BoardShellRestoreDefense.test.tsx` mounts the canvas after a restore throw). Good coverage.

### Non-blocking notes
- 🟡 `normalize.ts` re-declares `BASE62_DIGITS`/`SMALLEST_INTEGER` locally; the mirrored `isJitterbugValid` in `BoardShellRestoreDefense.test.tsx` is a hand-rolled subset of the rule (charset + intLen fit only, no sentinel/fraction check). It's adequate as a throw stand-in, but if the real jitterbug rule drifts, that test mock could mask a divergence — consider importing the shared predicate where practical.
- 🟡 Draft PR with `code-review` gate still PENDING; CI checks are SUCCESS/SKIPPED (no failures). Flagging draft status; no code concern.

jeff-wilson2010 (approved):
Reviewed at head `35c7661c8a0047f38123a3a93d892ffe462f891d` (line-by-line, cross-checked against the BE rule). **APPROVE.**

### Validity rule — `packages/whiteboard-schema/src/normalize.ts`
`isValidOrderKey` is a faithful non-throwing port of `fractional-indexing`'s `validateOrderKey`, and it is byte-for-byte consistent with the backend's rule:
- full-key base62 charset gate (`BASE62_RE`) — matches BE's `BASE62_CHARS` gate and covers both the integer and fractional parts;
- `SMALLEST_INTEGER` sentinel rejected — matches the library's first check;
- head → integer-part length map (`a`→2…`z`→27, `A`→27…`Z`→2, illegal head → 0) is identical to `getIntegerLength` under the default `BASE_52_DIGITS`, and `intLen > key.length` is rejected — matches `getIntegerPart`;
- trailing-zero rule (`fraction.slice(-1) === '0'`) — matches the library's `f.slice(-1) === digits[0]` check.

I verified the accepted/rejected sets line up with the BE probe: both reject `r00000003`, `a00`, and the sentinel, and both accept `a0`/`a1`/`Zz`/`a0V`. The two repos agree.

### Restore depth-defence — `packages/docs/src/board/BoardShell.tsx`
Both restore call sites now pass `{ normalizeIndices: true }`. The `initialData` restore runs during render, outside the collab binding's guarded `applyRemote` and outside `BoardErrorBoundary` (which only wraps `<Excalidraw>`), so an unrecoverable throw there would previously tear down the whole shell — the new try/catch degrades it to an empty initial scene so the canvas still mounts and the later observe→applyRemote path repaints. The `applyRemote` site is already guarded. This is correctly framed as depth-defence, not the root fix — `normalizeIndices`/`generateNKeysBetween` themselves throw on a structurally-invalid *existing* key, so the FE cannot repair it, only avoid crashing. The root fix lives in BE repair (PR #52).

### Tests
`normalize.test.ts` pins the contract precisely — `r00000003` and the whole `r`+base36 family invalid, `a00` invalid (trailing zero), the sentinel invalid, `a` invalid (short integer part), and `a0`/`a1`/`Zz`/`a0V` valid. `BoardShellRestoreDefense.test.tsx` seeds a cold-reopen scene carrying `r00000003` and asserts the canvas mounts (degraded) rather than the render throwing. Scope is clean. No blockers.

yujiawei (approved):
# Code Review — PR #622 (octo-web)

Reviewed at head `35c7661c8a0047f38123a3a93d892ffe462f891d` against `main`. Repo checked out at the head SHA; both new test files run green locally (`whiteboard-schema` 12/12, `packages/docs` `src/board` 176/176), and the order-key port was fuzzed against the reference `fractional-indexing` `validateOrderKey`.

## Verdict: Approve

No correctness, security, or data-loss blockers. The core security-relevant change — the fractional-index validation port — is correct. A few maintainability / accuracy items are called out below as non-blocking suggestions; the most important one (the `normalizeIndices` option) is worth a quick author confirmation but does not change the runtime safety of the fix.

---

## 1. Spec compliance

The PR delivers exactly the FE half it describes — schema struct-check + restore depth defence — with no scope creep (no new flags, fields, or unrelated behavior).

- **`isValidIndex` struct-check** — implemented as a non-throwing boolean port of `validateOrderKey`. ✅
- **`normalizeElement` strips an invalid `index`** to a clean absent state. ✅ (`normalize.ts:140`)
- **Both restore call sites** wired for depth defence. ✅ (`BoardShell.tsx:837`, `:1027-1031`)

No missing, extra, or divergent behavior against the stated intent.

## 2. Correctness of the order-key port (the security-relevant core)

I ported the new `isValidOrderKey` and the reference classic `validateOrderKey` side by side and fuzzed **200,000** random strings (base62 + illegal chars, lengths 1–30) plus every targeted edge case (`a0`, `a1`, `a0V`, `Zz`, `r00000003`, the whole `r[0-9a-z]{8}` family, `SMALLEST_INTEGER`, single-char heads, trailing-zero fractions):

```
tested 200000  mismatches 0
```

The port is faithful:
- `orderKeyIntegerLength`: `'a'→2 … 'z'→27`, `'Z'→2 … 'A'→27` matches `getIntegerLength`.
- `SMALLEST_INTEGER = 'A' + 26×'0'` (27 chars) matches the classic sentinel (`getIntegerLength('A') = 27`).
- The `intLen > key.length` and trailing-`'0'` fraction rules match `getIntegerPart` + the `f.slice(-1) === digits[0]` check.

This is the piece that actually stops the invalid key at the schema boundary (`normalizeElement`), and it is correct. Good, well-commented work.

## 3. Non-blocking findings (P2)

### P2-1 — `normalizeIndices: true` is a no-op in Excalidraw `0.18.1`
This is the item I'd most like the author to confirm. In the pinned Excalidraw version, `restoreElements` reads **only** `refreshDimensions` and `repairBindings` from its `opts`:

- `dist/types/excalidraw/data/restore.d.ts` types the third arg as `{ refreshDimensions?; repairBindings? }` — no `normalizeIndices`.
- The dev build confirms the body only branches on `opts?.repairBindings` / `opts.refreshDimensions`; `normalizeIndices` appears nowhere in the shipped bundle.

Consequences:
- `RestoreElementsFn` (`BoardShell.tsx:102`) adds `normalizeIndices?` to the signature, implying library support that does not exist at this version.
- The two comments (`:832-836`, `:1021-1026`) describe an "ask Excalidraw to re-index on restore" behavior that the option does not trigger.

**Why this is not a blocker:** `restoreElements` *unconditionally* calls `syncInvalidIndices(...)` regardless of opts, which regenerates invalid ordering in place, and the real safety net here is the surrounding `try/catch` (and, on the applyRemote path, the `repairForRender` → `normalizeElement` pre-pass that strips the bad `index` before restore is ever called). So the canvas is protected either way. But the code, the added type field, the comments, and the PR description all rest on a mechanism that isn't active in `0.18.1`. Suggest either dropping `normalizeIndices` (and the type field) and letting `syncInvalidIndices` + `try/catch` carry the defence explicitly, or gating it behind a version bump when the option lands.

### P2-2 — Restore-defence test mock does not match real `0.18.1` behavior
`BoardShellRestoreDefense.test.tsx` mocks `restoreElements` to **throw** on a structurally-invalid `index` ("as the real library does"). In `0.18.1`, a single lone element carrying `r00000003` does **not** make the real `restoreElements` throw — `syncInvalidIndices` → `generateIndices` → `generateNKeysBetween(undefined, undefined, n)` regenerates a fresh key for it (the invalid key is never used as a neighbor bound), so it is repaired, not thrown on. The real throw only occurs when an invalid key is used as a *bound* for regenerating a neighbor (multi-element ordering). The test therefore validates "the `try/catch` degrades gracefully **if** restore throws" — which is a real and useful guarantee — but the specific single-element scenario it sets up would be silently repaired by the real library, not thrown on. Worth a comment adjustment so the test's premise isn't read as "this exact input crashes the real lib."

### P2-3 — `initialElements` restore lacks the `normalizeElement` pre-pass the applyRemote path has
The applyRemote path runs `repairForRender` (→ shared `normalizeElement`, which strips the invalid `index`) **before** `restore` (`binding.ts:592`). The render-time `initialElements` path (`BoardShell.tsx:1019-1031`) hands `raw` straight to `restore` with no such pre-pass, then falls back to `[]` on throw — all-or-nothing. Running the same shared `normalizeElement` over `raw` first would strip only the offending `index` (Excalidraw then re-indexes it) and **preserve every valid element**, instead of degrading the entire initial scene to empty on a single bad key. Since the later observe→applyRemote repaints anyway this is not a data-loss bug, but it would make the cold-open path degrade as gracefully as the collab path and keep the two paths symmetric.

### P2-4 — Silent `catch` has no observability
The `catch { return [] }` (`BoardShell.tsx:1029-1030`) swallows the error with no log/telemetry. If `restoreElements` starts throwing for reasons unrelated to order keys, users get a blank initial scene with no signal. A `console.error` / telemetry counter here would keep the degradation visible.

## 4. Notes for a human reviewer (security-sensitive)

- **No destructive sync-back from the `[]` fallback.** I traced the concern that an empty initial scene could be synced back as a whiteboard-wide delete: it cannot. The binding seeds all pre-existing Y.Doc elements as `remoteOrigin` at construction (`binding.ts:219-222`); `handleLocalChange` only tombstones-by-absence for `locallyAuthored && !remoteOrigin` ids (`:341`), and an empty `onChange` produces an empty diff that guard-2 early-returns without a transaction (`:361-368`). The reinit-preserve + H1 empty-apply guards are exactly built for this "reinit fires empty onChange" case. No data loss.
- **No ReDoS.** `/^[0-9A-Za-z]+$/` is a linear, non-backtracking character class; an over-long `index` costs O(n), not catastrophic. (An explicit max-length cap before validation would be a cheap belt-and-suspenders, but there is no ReDoS here.)
- **Cross-repo contract:** this PR *defines* the legality rule the backend `normalize.ts` must match byte-for-byte. The port is correct against `fractional-indexing`; the BE side should be verified against the same reference when it lands (out of scope for this PR).

## 5. Tooling note

An independent long-context advisory pass (cross-family) contributed to the sync-back and length-bound analysis above. A second independent line-level pass could not complete this run due to an unrelated upstream gateway/tooling error (it did not return findings) — this review is therefore based on my own verification (repo checkout at head, test runs, fuzzing of the port, and bundle inspection of Excalidraw `0.18.1`) plus the one advisory pass that succeeded. Flagging so the absent pass is not read as silent agreement.

Issue #51, opened 2026-05-19
bug: Starting group chat from private chat three-dots menu returns 400 'Group does not exist'


wangty371:
## Description

When clicking **⋯ (three dots) → Start Group Chat** from an existing private chat, the action fails with the error message "Group does not exist" and the following console error:

```
POST /api/v1/groups/f61983b1a2084085a994d47437e078b5/members 400 (Bad Request)
```

The same action initiated from the **left sidebar "+" button → Start Group Chat** works correctly.

## Steps to Reproduce

1. Open any existing private chat conversation
2. Click ⋯ (three dots) in the top-right corner
3. Click "Start Group Chat"
4. Select one or more members and confirm
5. Error toast "Group does not exist" appears — no group is created

## Root Cause

Both entry points use the same `OrganizationalGroupNew` component but follow different code paths:

**Sidebar "+" button (works)**
- `action = createGroup`, `channel` is empty
- Calls `POST /group/create` first to create a new group, gets back `group_no`, then navigates to the new group chat ✅

**Private chat three-dots menu (broken)**
- `action = AddMember`, `channel` is the current person channel (`channelType=1`, `channelID` = the other user's UID)
- Skips group creation and directly calls `addSubscribers(channel, selectedUids)`
- Results in `POST /groups/{userUID}/members` — but `userUID` is a user ID, not a group ID → server returns 400 ❌

**Bug location:** `packages/dmworkcontacts/src/Organizational/GroupNew/index.tsx` — the `AddMember` branch in `onOK()` does not check `channel.channelType` before calling `addSubscribers`.

```typescript
// addSubscribers (packages/dmworkdatasource/src/datasource.ts)
addSubscribers(channel: Channel, uids: string[]): Promise<void> {
    return WKApp.apiClient.post(`groups/${channel.channelID}/members`, {
        members: uids,
    })
}
// When channelType === ChannelTypePerson, channelID is a user UID — not a group ID
```

## Suggested Fix

In the `AddMember` branch of `onOK()`, check `channel.channelType`. When it is `ChannelTypePerson`, create a new group first before adding members, instead of calling `addSubscribers` directly. The correct pattern is already implemented in the `channel.subscribers` section's `onAdd` handler in `packages/dmworkbase/src/module.tsx`.

```typescript
if (this.props.action === OrganizationalGroupNewAction.AddMember) {
    if (this.props.channel.channelType === ChannelTypePerson) {
        // Starting group chat from a private chat: create group first
[...]

PR #626, opened 2026-07-10
fix(board): render bot-written whiteboards whose fractional indices Excalidraw can't parse

ploy-elison:
## Problem

A whiteboard (`docType: board`) populated through the bot scene API (`PATCH /v1/bot/docs/{docId}/scene`) renders as an **empty canvas** in the board editor, even though the authoritative Y.Doc provably holds valid, non-deleted elements with correct geometry and indices. Ref: octo-docs-backend #51.

## Root cause — the fractional-index key format

The backend fills an element's missing z-order key with a zero-padded scheme — `r00000000`, `r00000001`, … (`r${seq.toString(36).padStart(8,'0')}`). That key passes the shared loose `@octo/whiteboard-schema` charset check (`/^[A-Za-z0-9]+$/`), but it is **not** a valid Excalidraw fractional index: in Excalidraw's bundled `fractional-indexing`, the head char `r` encodes a **19-digit** integer part, while only 9 characters follow.

`restoreElements` / `reconcileElements` run `syncInvalidIndices` → `generateNKeysBetween`, which passes neighbouring keys straight to the library's `validateOrderKey`. A key it can't parse makes the **whole apply throw** `invalid order key`. `ExcalidrawYjsBinding.applyRemote` wraps the rebuild in a `try/catch` that keeps the last good scene on error (the P1-2 batch guard), so the throw is swallowed and the canvas stays **blank**.

Keys that happen to already be monotonic slip through untouched (Excalidraw only regenerates when ordering looks broken), which is why a minimal 3-element repro can look fine while real boards blank the moment an index must be regenerated — a non-monotonic batch, a local element drawn next to remote ones, or a duplicate.

### Runtime evidence

Executing the real `@excalidraw/excalidraw@0.18.1` `restoreElements` / `reconcileElements` against the exact backend key format:

```
BEFORE FIX (raw backend keys):
  non-monotonic bot restore        => THROWS: invalid order key: r00000005
  local(no index)+bot reconcile    => THROWS: invalid order key: r00000001
AFTER FIX (sanitize -> real Excalidraw):
  non-monotonic bot restore        => OK, kept 2
  cold open reconcile              => OK, kept 2
  local(no index)+bot reconcile    => OK, kept 3
  local(valid a5)+bot reconcile    => OK, kept 3
  regenerated keys r1:a0 r2:a1 r3:a2  | all valid: true  (z-order preserved)
```

## Fix

A frontend, **render-only** defence that strips any `index` key Excalidraw's grammar rejects before restore/reconcile, so Excalidraw's own `syncInvalidIndices` assigns fresh valid keys instead of throwing.

- `collab/fractionalIndex.ts` — `isExcalidrawFractionalIndex` (faithful to `fractional-indexing`'s `validateOrderKey`) + `sanitizeFractionalIndices`.
- Applied in `repairForRender` (covers `applyRemote`'s adapter and raw paths) and in the BoardShell `initialElements` memo (cold-open `initialData`, which seeds from raw Y.Doc elements and bypasses `repairForRender`).
- **Order preserved** by stable-sorting on the original key strings — the ordering the backend's zero-padded scheme intends and the comparison Excalidraw uses — so a bot scene keeps its authored z-order.
- **No regression:** scenes whose keys are all valid (every human-drawn board) take a zero-copy fast path and are handed through untouched, so real-time collaboration/reconcile is unchanged.
- The result is never written back to the Y.Doc; the backend remains the authoritative index writer.

## Tests

`collab/__tests__/fractionalIndex.test.ts` (validator + sanitizer, incl. fast-path identity, strip-only-invalid, order preservation) and new `repair.test.ts` cases. Full `board/collab` suite: **109 passing**.

## Follow-up (out of scope, backend)

The durable fix is for octo-docs-backend to emit real Excalidraw fractional-index keys (e.g. via `fractional-indexing`'s `generateNKeysBetween`) rather than the `r00000000` scheme. This PR is the frontend render-defence so existing bot-written boards render correctly regardless.

## Verification note

The live repro board could not be exercised in a real browser from the dev environment (no running collab stack / repro-board access). Localization and the fix were verified by executing the real Excalidraw 0.18.1 functions headlessly against the exact backend key format. A final in-browser confirmation on `d_711f6a9848f478892d04beb2` is recommended before this leaves draft.

### DIFF
--- a/packages/docs/src/board/BoardScenePreview.tsx
+++ b/packages/docs/src/board/BoardScenePreview.tsx
@@ -12,7 +12,7 @@
 import { useEffect, useMemo, useRef, useState, type ComponentType, type ReactElement, type ReactNode } from 'react'
 import { i18n, t } from '../octoweb/index.ts'
 import { BoardMainMenu, type ExcalidrawMainMenu } from './BoardMainMenu.tsx'
-import { normalizeFileRef, type BinaryFileData, type FileFetchRef } from './collab/index.ts'
+import { normalizeFileRef, sanitizeFractionalIndices, type BinaryFileData, type FileFetchRef } from './collab/index.ts'
 import { fetchBoardFileBinaries } from './boardFiles.ts'
 import type { BoardVersionScene } from './boardVersions.ts'
 
@@ -113,10 +113,19 @@ export function BoardScenePreview({ scene, dark, docId }: { scene: BoardVersionS
 
   // Rehydrate raw persisted/historical elements into renderable Excalidraw shapes (drops unknown
   // element types, fills defaults) — the same restore step the live board applies to initialData.
-  const elements = useMemo(() => {
+  //
+  // Strip any fractional-index key Excalidraw's grammar cannot parse BEFORE restore (XIN-791): a
+  // historical scene is the same backend-authored Y.Doc state the live board renders, carrying the
+  // zero-padded `r00000000` scheme, and `restoreElements` runs the same `syncInvalidIndices` that
+  // throws `invalid order key` on an unparseable key — which here (unlike the live board) is a
+  // render-time throw the BoardVersionPanel only catches with an error boundary. Sanitizing here
+  // closes the third untrusted-restore path (BoardShell cold-open + repairForRender were the other
+  // two); it is a no-op for valid-keyed human scenes.
+  const elements = useMemo<unknown[]>(() => {
     const restore = restoreElementsRef.current
-    if (restore) return restore(scene.elements, null)
-    return scene.elements
+    const safe = sanitizeFractionalIndices(scene.elements as Parameters<typeof sanitizeFractionalIndices>[0])
+    if (restore) return restore(safe, null)
+    return [...safe]
     // Recompute once the helper resolves (Excalidraw becoming non-null flips this).
   }, [scene, Excalidraw])
 
--- a/packages/docs/src/board/BoardShell.tsx
+++ b/packages/docs/src/board/BoardShell.tsx
@@ -30,7 +30,7 @@ import { installExcalidrawDebrand } from './excalidrawDebrand.ts'
 import { installLibraryControlButtons } from './libraryControlButtons.ts'
 import type { WhiteboardSession, BoardTerminal } from './collab/index.ts'
 import type { ExcalidrawElement, BinaryFileData, FileFetchRef } from './collab/index.ts'
-import { makeGenerateIdForFile, dataURLToBlob } from './collab/index.ts'
+import { makeGenerateIdForFile, dataURLToBlob, sanitizeFractionalIndices } from './collab/index.ts'
 import { presignUpload, uploadBinary } from '../attachments/api.ts'
 import { fetchBoardFileBinaries } from './boardFiles.ts'
 import {
@@ -1018,14 +1018,20 @@ export function BoardShell(props: BoardShellProps): ReactElement {
     }
     const restore = restoreElementsRef.current
     if (!restore) return [...raw]
-    // XIN-795 ④ depth defence: this restore runs during render (not inside the collab binding's
-    // guarded applyRemote), so a throw on a structurally-invalid persisted key — one the BE repair
-    // (XIN-794) failed to stop at source — would take down the whole BoardShell render, not just a
-    // frame. `normalizeIndices` lets Excalidraw re-index recoverable keys; the try/catch degrades an
+    // XIN-791 render defence: strip any fractional-index key Excalidraw cannot parse before restore.
+    // This memo seeds `initialData` straight from raw Y.Doc elements (bypassing repairForRender), so a
+    // backend `r00000000`-style key would otherwise reach restoreElements and blank a cold-opened
+    // bot-written board. sanitizeFractionalIndices is a no-op for valid-keyed (human) scenes.
+    // XIN-795 ④ depth defence: this restore runs during render (outside the collab binding's guarded
+    // applyRemote and outside BoardErrorBoundary's subtree), so a throw on a structurally-invalid
+    // persisted key that even normalizeIndices cannot repair would take down the whole BoardShell
+    // render. `normalizeIndices` lets Excalidraw re-index recoverable keys; the try/catch degrades an
     // unrecoverable throw to an empty initial scene so the canvas still mounts and the later
     // observe→applyRemote path can repaint, with BoardErrorBoundary as the final backstop.
     try {
-      return restore(raw, null, { normalizeIndices: true })
+      return restore(sanitizeFractionalIndices(raw as readonly ExcalidrawElement[]), null, {
+        normalizeIndices: true,
+      })
     } catch {
       return []
     }
--- a/packages/docs/src/board/__tests__/BoardScenePreview.test.tsx
+++ b/packages/docs/src/board/__tests__/BoardScenePreview.test.tsx
@@ -26,10 +26,32 @@ vi.mock('@excalidraw/excalidraw', () => {
   }
   const MainMenu = (() => null) as unknown as { DefaultItems: Record<string, unknown> }
   MainMenu.DefaultItems = {}
+  // Faithful to `fractional-indexing`'s head-length rule (`a`→2 … `z`→27): a key whose head claims
+  // more integer digits than the string carries is unparseable — the backend's zero-padded
+  // `r00000000` scheme (head `r`→19 digits, only 9 follow).
+  const parseableIndex = (key: unknown): boolean => {
+    if (typeof key !== 'string' || key.length === 0) return true
+    const head = key[0]
+    let intLen = -1
+    if (head >= 'a' && head <= 'z') intLen = head.charCodeAt(0) - 'a'.charCodeAt(0) + 2
+    else if (head >= 'A' && head <= 'Z') intLen = 'Z'.charCodeAt(0) - head.charCodeAt(0) + 2
+    return intLen >= 0 && intLen <= key.length
+  }
   return {
     Excalidraw,
     MainMenu,
-    restoreElements: (els: readonly unknown[] | null | undefined) => (els ? [...els] : []),
+    // Mirror the real `restoreElements` → `syncInvalidIndices`: regenerating a key next to an
+    // unparseable one throws `invalid order key` (octo-docs-backend #51), the throw that blanks
+    // bot-written boards. The preview must sanitize before restore so this never fires.
+    restoreElements: (els: readonly unknown[] | null | undefined) => {
+      const out = els ? [...els] : []
+      for (const el of out) {
+        if (el && typeof el === 'object' && !parseableIndex((el as { index?: unknown }).index)) {
+          throw new Error('invalid order key')
+        }
+      }
+      return out
+    },
   }
 })
 vi.mock('@excalidraw/excalidraw/index.css', () => ({}))
@@ -49,6 +71,7 @@ vi.mock('../../attachments/api.ts', async (importOriginal) => {
 })
 
 import { BoardScenePreview } from '../BoardScenePreview.tsx'
+import { isExcalidrawFractionalIndex } from '../collab/index.ts'
 
 // A one-pixel PNG so the fetched Blob decodes into a real data URL through jsdom's FileReader.
 const PNG_BYTES = Uint8Array.from(atob('iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+M9QDwADhgGAWjR9awAAAABJRU5ErkJggg=='), (c) => c.charCodeAt(0))
@@ -107,5 +130,54 @@ describe('BoardScenePreview image hydration', () => {
     await waitFor(() => expect(screen.getByTestId('excalidraw-canvas')).toBeTruthy())
     const files = lastInitialData?.files as Record<string, { dataURL?: string }> | undefined
     expect(files?.file_a?.dataURL).toBeUndefined()
+  })
+})
+
+// XIN-811 Blocking #1: the version-history preview seeds a real <Excalidraw> from the SAME
+// restoreElements step the live board uses, but before this fix it skipped sanitizeFractionalIndices
+// — so a bot-written scene (backend `r00000000` scheme) hit `syncInvalidIndices`, threw `invalid
+// order key`, and blanked the preview (BoardVersionPanel only catches that with an error boundary).
+// This was the third untrusted-restore path the original PR missed. The mocked restoreElements
+// above throws on an unparseable key exactly as the real library does, so without the sanitize
+// these renders throw and never mount.
+describe('BoardScenePreview bot-board fractional-index sanitize (Blocking #1)', () => {
+  // A historical bot scene whose z-order keys use the unparseable backend scheme, supplied
+  // bottom→top; Y.Doc/serialized order is arbitrary, so ids are deliberately out of key order.
+  function botScene(): BoardVersionScene {
+    return {
+      elements: [
+        { id: 'mid', type: 'rectangle', index: 'r00000001' },
+        { id: 'top', type: 'rectangle', index: 'r00000002' },
+        { id: 'bg', type: 'rectangle', index: 'r00000000' },
+      ],
+      files: {},
+    }
+  }
+
+  it('sanitizes bot fractional indices before restore so the preview mounts instead of throwing', async () => {
+    render(<BoardScenePreview scene={botScene()} docId="bd_1" />)
+    // Without the sanitize the mocked restore throws `invalid order key` and the canvas never mounts.
+    await waitFor(() => expect(screen.getByTestId('excalidraw-canvas')).toBeTruthy())
+    const els = (lastInitialData?.elements ?? []) as { id: string; index?: unknown }[]
+    // Nothing Excalidraw's grammar would reject reaches the canvas.
+    for (const el of els) expect(el.index == null || isExcalidrawFractionalIndex(el.index)).toBe(true)
+    // Authored z-order is recovered from the arbitrary source order (homogeneous bot scene).
+    expect(els.map((e) => e.id)).toEqual(['bg', 'mid', 'top'])
+  })
+
+  it('does not throw on a mixed scene and keeps author order (does not sink valid keys)', async () => {
+    // Bot background (unparseable) UNDER a human shape (valid `a5`) — a string sort would invert it.
+    const scene: BoardVersionScene = {
+      elements: [
+        { id: 'bg', type: 'rectangle', index: 'r00000000' },
+        { id: 'human', type: 'rectangle', index: 'a5' },
+      ],
+      files: {},
+    }
+    render(<BoardScenePreview scene={scene} docId="bd_1" />)
+    await waitFor(() => expect(screen.getByTestId('excalidraw-canvas')).toBeTruthy())
+    const els = (lastInitialData?.elements ?? []) as { id: string; index?: unknown }[]
+    expect(els.map((e) => e.id)).toEqual(['bg', 'human'])
+    expect(els.find((e) => e.id === 'human')!.index).toBe('a5')
   })
 })
--- a/packages/docs/src/board/collab/__tests__/fractionalIndex.test.ts
+++ b/packages/docs/src/board/collab/__tests__/fractionalIndex.test.ts
@@ -87,4 +87,49 @@ describe('sanitizeFractionalIndices', () => {
       expect(el.index == null || isExcalidrawFractionalIndex(el.index)).toBe(true)
     }
   })
+
+  // Mixed scenes: valid Excalidraw keys coexisting with the backend's unparseable `r00000000`
+  // scheme. A global string sort compares across the two incomparable key spaces (`'a5' < 'r0…'`),
+  // dragging every valid-keyed element beneath the bot scheme and inverting the author's stacking.
+  // syncInvalidIndices honours ARRAY ORDER and regenerates only the stripped keys in place, so a
+  // mixed scene must be handed through in its original order — sorted output would corrupt z-order.
+  describe('mixed scene (valid + backend keys) — z-order fidelity', () => {
+    it('keeps a bottom bot element below a top human element (does not sort the valid key first)', () => {
+      // Authored bottom→top: a bot-written background (unparseable `r00000000`) UNDER a human shape
+      // (valid `a5`). A string sort would place `a5` before `r00000000` and sink the human shape.
+      const els = [makeEl('bg', { index: 'r00000000' }), makeEl('human', { index: 'a5' })]
+      const out = sanitizeFractionalIndices(els)
+      expect(out.map((e) => e.id)).toEqual(['bg', 'human'])
+      // Bot key stripped so Excalidraw regenerates it in array position; valid key left intact.
+      expect('index' in out.find((e) => e.id === 'bg')!).toBe(false)
+      expect(out.find((e) => e.id === 'human')!.index).toBe('a5')
+    })
+
+    it('preserves valid-keyed relative order and re-slots stripped bot elements at their anchors', () => {
+      // human a0, two bot elements, human a1 — author order must survive verbatim (no cross-scheme
+      // sort would move the bot pair below both humans, and no intra-scheme sort should reorder
+      // elements the author already placed).
+      const els = [
+        makeEl('h0', { index: 'a0' }),
+        makeEl('bot1', { index: 'r00000001' }),
+        makeEl('bot0', { index: 'r00000000' }),
+        makeEl('h1', { index: 'a1' }),
+      ]
+      const out = sanitizeFractionalIndices(els)
+      expect(out.map((e) => e.id)).toEqual(['h0', 'bot1', 'bot0', 'h1'])
+      expect(out.find((e) => e.id === 'h0')!.index).toBe('a0')
+      expect(out.find((e) => e.id === 'h1')!.index).toBe('a1')
+      for (const id of ['bot0', 'bot1']) expect('index' in out.find((e) => e.id === id)!).toBe(false)
+    })
+
+    it('leaves no unparseable key in a mixed scene', () => {
+      const els = [
+        makeEl('bg', { index: 'r00000000' }),
+        makeEl('mid', { index: 'a2' }),
+        makeEl('fg', { index: 'r00000005' }),
+      ]
+      const out = sanitizeFractionalIndices(els)
+      for (const el of out) expect(el.index == null || isExcalidrawFractionalIndex(el.index)).toBe(true)
+    })
+  })
 })
--- a/packages/docs/src/board/collab/__tests__/fractionalIndex.test.ts
+++ b/packages/docs/src/board/collab/__tests__/fractionalIndex.test.ts
@@ -1,0 +1,90 @@
+// Fractional-index render-defence (XIN-791 / octo-docs-backend #51).
+//
+// Excalidraw's bundled `fractional-indexing` throws `invalid order key` on a key its grammar
+// cannot parse, and `restoreElements` / `reconcileElements` call it whenever an index must be
+// regenerated — so a backend-authored key that passes the shared loose `/^[A-Za-z0-9]+$/` charset
+// but not Excalidraw's grammar (the zero-padded `r00000000` scheme) blanks the canvas. These cases
+// lock in the validator's fidelity to that grammar and the sanitizer's strip-and-preserve-order
+// behaviour, including the zero-copy fast path that leaves valid (human-drawn) scenes untouched.
+import { describe, it, expect } from 'vitest'
+import { isExcalidrawFractionalIndex, sanitizeFractionalIndices } from '../fractionalIndex.ts'
+import { makeEl } from './helpers.ts'
+
+describe('isExcalidrawFractionalIndex', () => {
+  it('accepts real Excalidraw keys', () => {
+    for (const k of ['a0', 'a1', 'a2', 'a0V', 'Zz', 'b00', 'a0V8', 'Zzol']) {
+      expect(isExcalidrawFractionalIndex(k)).toBe(true)
+    }
+  })
+
+  it('rejects the backend zero-padded scheme (root cause of #51)', () => {
+    // head `r` claims a 19-char integer part; only 9 chars follow → unparseable by the library.
+    for (const k of ['r00000000', 'r00000001', 'r00000003', 'r0000000z']) {
+      expect(isExcalidrawFractionalIndex(k)).toBe(false)
+    }
+  })
+
+  it('rejects malformed / non-string / empty keys', () => {
+    for (const k of ['', 'r', '0', '00', 'a', 'a00', '@x', 'A'.repeat(1) + '0'.repeat(26)]) {
+      expect(isExcalidrawFractionalIndex(k)).toBe(false)
+    }
+    expect(isExcalidrawFractionalIndex(undefined)).toBe(false)
+    expect(isExcalidrawFractionalIndex(null)).toBe(false)
+    expect(isExcalidrawFractionalIndex(123 as unknown)).toBe(false)
+  })
+
+  it('rejects a fractional part that ends in the smallest digit', () => {
+    // Mirrors the library's `validateOrderKey` trailing-`0` rule.
+    expect(isExcalidrawFractionalIndex('a0V0')).toBe(false)
+    expect(isExcalidrawFractionalIndex('a0V1')).toBe(true)
+  })
+})
+
+describe('sanitizeFractionalIndices', () => {
+  it('returns the SAME array reference when every key is valid (fast path, no regression)', () => {
+    const els = [makeEl('a', { index: 'a0' }), makeEl('b', { index: 'a1' })]
+    expect(sanitizeFractionalIndices(els)).toBe(els)
+  })
+
+  it('treats an absent index as valid (fast path)', () => {
+    const els = [makeEl('a', { index: undefined }), makeEl('b', { index: 'a1' })]
+    expect(sanitizeFractionalIndices(els)).toBe(els)
+  })
+
+  it('drops only the invalid keys and leaves other fields (and valid keys) intact', () => {
+    const els = [
+      makeEl('rect', { index: 'r00000000', width: 220, height: 100 }),
+      makeEl('local', { index: 'a5' }),
+    ]
+    const out = sanitizeFractionalIndices(els)
+    const rect = out.find((e) => e.id === 'rect')!
+    const local = out.find((e) => e.id === 'local')!
+    expect('index' in rect).toBe(false) // stripped → Excalidraw will regenerate a valid key
+    expect(rect.width).toBe(220) // untouched
+    expect(rect.height).toBe(100)
+    expect(local.index).toBe('a5') // valid key preserved
+  })
+
+  it('preserves the authored z-order by stable-sorting on the original key strings', () => {
+    // Deliberately supply the elements out of index order (Y.Map insertion order is arbitrary).
+    const els = [
+      makeEl('third', { index: 'r00000003' }),
+      makeEl('first', { index: 'r00000000' }),
+      makeEl('second', { index: 'r00000001' }),
+    ]
+    const out = sanitizeFractionalIndices(els)
+    expect(out.map((e) => e.id)).toEqual(['first', 'second', 'third'])
+  })
+
+  it('produces a scene with no key Excalidraw would reject', () => {
+    const els = [
+      makeEl('rect', { index: 'r00000000' }),
+      makeEl('ellipse', { index: 'r00000001' }),
+      makeEl('text', { index: 'r00000003', type: 'text' }),
+    ]
+    const out = sanitizeFractionalIndices(els)
+    for (const el of out) {
+      expect(el.index == null || isExcalidrawFractionalIndex(el.index)).toBe(true)
+    }
+  })
+})
--- a/packages/docs/src/board/collab/__tests__/repair.test.ts
+++ b/packages/docs/src/board/collab/__tests__/repair.test.ts
@@ -112,4 +112,21 @@ describe('repairForRender (shared normalizeElement)', () => {
     expect((out[0] as Record<string, unknown>).future).toEqual({ k: 1 })
     expect(out[0]).not.toBe(el)
   })
+
+  it('XIN-791: strips a backend index key Excalidraw cannot parse, keeping the element', () => {
+    // The backend fills a missing key with `r00000000` (zero-padded), which passes the shared
+    // loose charset but is not a valid Excalidraw fractional index; left in place it makes
+    // restore/reconcile throw and blank the canvas. Pass 3 drops the key (element survives) so
+    // Excalidraw's syncInvalidIndices regenerates a valid one.
+    const out = repairForRender([makeEl('rect', { index: 'r00000000', width: 220, height: 100 })])
+    expect(out.map((e) => e.id)).toEqual(['rect']) // element kept
+    expect('index' in (out[0] as Record<string, unknown>)).toBe(false) // unparseable key dropped
+    expect((out[0] as Record<string, unknown>).width).toBe(220) // geometry intact
+  })
+
+  it('XIN-791: leaves a valid Excalidraw index key untouched', () => {
+    const out = repairForRender([makeEl('a', { index: 'a0' }), makeEl('b', { index: 'a1' })])
+    expect((out[0] as Record<string, unknown>).index).toBe('a0')
+    expect((out[1] as Record<string, unknown>).index).toBe('a1')
+  })
 })
--- a/packages/docs/src/board/collab/fractionalIndex.ts
+++ b/packages/docs/src/board/collab/fractionalIndex.ts
@@ -7,10 +7,12 @@
 // `syncInvalidIndices` → `generateNKeysBetween`, which passes neighbouring keys straight to the
 // library's `validateOrderKey`; a key it cannot parse makes the WHOLE apply throw.
 //
-// The shared `@octo/whiteboard-schema` `isValidIndex` only checks the loose `/^[A-Za-z0-9]+$/`
-// charset (deliberately — it is the FE/BE contract and the backend authoritative repair writes
-// under it). The backend fills a missing key with a zero-padded scheme (`r00000000`, see
-// octo-docs-backend `whiteboard/repair.ts`) that passes that loose check but is NOT a valid
+// Whether a key is a parseable Excalidraw order key is decided by the SINGLE shared validator
+// `isValidIndex` in `@octo/whiteboard-schema` (a non-throwing port of `fractional-indexing`'s
+// `validateOrderKey`, hardened in XIN-794/795). This module reuses that one definition rather than
+// keeping a second copy, so FE render defence, FE binding normalize and BE authoritative repair
+// all agree byte-for-byte on legality. The backend fills a missing key with a zero-padded scheme
+// (`r00000000`, see octo-docs-backend `whiteboard/repair.ts`) that is NOT a valid
 // Excalidraw key: head `r` claims a 19-digit integer part while only 9 chars follow. Such a key
 // renders fine as long as the scene never needs an index REGENERATED (all keys already ordered),
 // but the moment one does — a local element inserted between two remote ones, a non-monotonic /
@@ -32,41 +34,16 @@
 // so existing real-time collaboration / reconcile behaviour does not change.
 
 import type { ExcalidrawElement } from './types.ts'
-
-const BASE_62_DIGITS = new Set(
-  '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'.split(''),
-)
-
-/**
- * Integer-part length a fractional-index head char encodes, or -1 for a non-head char. Mirrors
- * `fractional-indexing`'s `getIntegerLength`: `a`→2 … `z`→27 (positive), `Z`→2 … `A`→27 (negative).
- */
-function integerLengthFromHead(head: string): number {
-  if (head >= 'a' && head <= 'z') return head.charCodeAt(0) - 'a'.charCodeAt(0) + 2
-  if (head >= 'A' && head <= 'Z') return 'Z'.charCodeAt(0) - head.charCodeAt(0) + 2
-  return -1
-}
+import { isValidIndex } from './schema.ts'
 
 /**
  * True when `key` is a fractional-index string Excalidraw's `fractional-indexing` can parse —
- * i.e. one `generateKeyBetween`/`validateOrderKey` will accept as a neighbour bound. Faithful to
- * that library's `validateOrderKey` + `getIntegerPart`, with the extra guarantee that every digit
- * is base-62 (a non-base-62 fractional char parses in `validateOrderKey` but breaks the base-62
- * arithmetic in `generateNKeysBetween`, so treating it as invalid here is the safe, stricter call).
+ * i.e. one `generateKeyBetween` / `validateOrderKey` will accept as a neighbour bound. Delegates to
+ * the shared `@octo/whiteboard-schema` `isValidIndex` so this render-defence pass and the FE/BE
+ * schema layer share ONE legality definition and can never drift apart.
  */
 export function isExcalidrawFractionalIndex(key: unknown): key is string {
-  if (typeof key !== 'string' || key.length === 0) return false
-  // The one explicitly-forbidden key in the library (smallest-integer sentinel).
-  if (key === 'A' + '0'.repeat(26)) return false
-  const intLen = integerLengthFromHead(key[0])
-  if (intLen < 0 || intLen > key.length) return false
-  for (let i = 1; i < key.length; i++) {
-    if (!BASE_62_DIGITS.has(key[i])) return false
-  }
-  // A fractional part is never allowed to end in the smallest digit (`0`).
-  const frac = key.slice(intLen)
-  if (frac.length > 0 && frac[frac.length - 1] === '0') return false
-  return true
+  return isValidIndex(key)
 }
 
 /**
--- a/packages/docs/src/board/collab/fractionalIndex.ts
+++ b/packages/docs/src/board/collab/fractionalIndex.ts
@@ -1,0 +1,110 @@
+// Fractional-index render-defence (FE local, render-only).
+//
+// Excalidraw expresses z-order with a fractional-index key (`element.index`) whose grammar is
+// defined by the `fractional-indexing` library it bundles: a head char (`a`–`z` / `A`–`Z`) that
+// encodes the integer part's LENGTH, that many base-62 integer digits, then an optional base-62
+// fractional part whose last digit is never `0`. `restoreElements` / `reconcileElements` call
+// `syncInvalidIndices` → `generateNKeysBetween`, which passes neighbouring keys straight to the
+// library's `validateOrderKey`; a key it cannot parse makes the WHOLE apply throw.
+//
+// The shared `@octo/whiteboard-schema` `isValidIndex` only checks the loose `/^[A-Za-z0-9]+$/`
+// charset (deliberately — it is the FE/BE contract and the backend authoritative repair writes
+// under it). The backend fills a missing key with a zero-padded scheme (`r00000000`, see
+// octo-docs-backend `whiteboard/repair.ts`) that passes that loose check but is NOT a valid
+// Excalidraw key: head `r` claims a 19-digit integer part while only 9 chars follow. Such a key
+// renders fine as long as the scene never needs an index REGENERATED (all keys already ordered),
+// but the moment one does — a local element inserted between two remote ones, a non-monotonic /
+// rolled-over backend batch, a duplicate — `generateNKeysBetween` throws `invalid order key`,
+// `applyRemote` swallows it (P1-2 batch guard), and the canvas paints EMPTY. That is the
+// bot-written-board blank-render failure (octo-docs-backend #51 / XIN-791).
+//
+// This pass detects keys Excalidraw's grammar rejects and DROPS them (render-only, never written
+// back to the Y.Doc — the backend stays the authoritative index writer, XIN-16 §4), letting
+// Excalidraw's own `syncInvalidIndices` assign fresh, valid keys. Order is preserved by
+// stable-sorting on the original key strings first — the same string comparison Excalidraw's
+// `orderByFractionalIndex` uses and the ordering the backend's zero-padded scheme intends — so a
+// bot scene keeps its authored z-order. Scenes whose keys are already all valid (every
+// human-drawn board) take a zero-copy fast path and are handed through untouched, so existing
+// real-time collaboration / reconcile behaviour does not change.
+
+import type { ExcalidrawElement } from './types.ts'
+
+const BASE_62_DIGITS = new Set(
+  '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz'.split(''),
+)
+
+/**
+ * Integer-part length a fractional-index head char encodes, or -1 for a non-head char. Mirrors
+ * `fractional-indexing`'s `getIntegerLength`: `a`→2 … `z`→27 (positive), `Z`→2 … `A`→27 (negative).
+ */
+function integerLengthFromHead(head: string): number {
+  if (head >= 'a' && head <= 'z') return head.charCodeAt(0) - 'a'.charCodeAt(0) + 2
+  if (head >= 'A' && head <= 'Z') return 'Z'.charCodeAt(0) - head.charCodeAt(0) + 2
+  return -1
+}
+
+/**
+ * True when `key` is a fractional-index string Excalidraw's `fractional-indexing` can parse —
+ * i.e. one `generateKeyBetween`/`validateOrderKey` will accept as a neighbour bound. Faithful to
+ * that library's `validateOrderKey` + `getIntegerPart`, with the extra guarantee that every digit
+ * is base-62 (a non-base-62 fractional char parses in `validateOrderKey` but breaks the base-62
+ * arithmetic in `generateNKeysBetween`, so treating it as invalid here is the safe, stricter call).
+ */
+export function isExcalidrawFractionalIndex(key: unknown): key is string {
+  if (typeof key !== 'string' || key.length === 0) return false
+  // The one explicitly-forbidden key in the library (smallest-integer sentinel).
+  if (key === 'A' + '0'.repeat(26)) return false
+  const intLen = integerLengthFromHead(key[0])
+  if (intLen < 0 || intLen > key.length) return false
+  for (let i = 1; i < key.length; i++) {
+    if (!BASE_62_DIGITS.has(key[i])) return false
+  }
+  // A fractional part is never allowed to end in the smallest digit (`0`).
+  const frac = key.slice(intLen)
+  if (frac.length > 0 && frac[frac.length - 1] === '0') return false
+  return true
+}
+
+/**
+ * Make a rebuilt scene safe for Excalidraw's restore/reconcile by dropping any `index` key its
+ * fractional-indexing grammar would reject. Render-only: the returned elements are never written
+ * back to the Y.Doc.
+ *
+ * Fast path: if every element already carries a valid key (or none), the input array is returned
+ * as-is — no reorder, no copy — so ordinary human-drawn boards are untouched. Otherwise a new,
+ * stable-sorted array is returned with the invalid keys stripped, so Excalidraw's
+ * `syncInvalidIndices` regenerates valid keys in the authored z-order instead of throwing.
+ */
+export function sanitizeFractionalIndices(
+  elements: readonly ExcalidrawElement[],
+): readonly ExcalidrawElement[] {
+  let anyInvalid = false
+  for (const el of elements) {
+    const idx = el.index
+    if (idx != null && !isExcalidrawFractionalIndex(idx)) {
+      anyInvalid = true
+      break
+    }
+  }
+  if (!anyInvalid) return elements
+
+  // Preserve authored order (elements with no key sort first, matching Excalidraw's own
+  // treatment of a missing index as ordering-first), keeping ties in original array order.
+  const positioned = elements.map((el, i) => ({ el, i }))
+  positioned.sort((a, b) => {
+    const ka = typeof a.el.index === 'string' ? a.el.index : ''
+    const kb = typeof b.el.index === 'string' ? b.el.index : ''
+    if (ka < kb) return -1
+    if (ka > kb) return 1
+    return a.i - b.i
+  })
+  return positioned.map(({ el }) => {
+    const idx = el.index
+    if (idx != null && !isExcalidrawFractionalIndex(idx)) {
+      // Drop only the offending key; every other field (unknown ones included) passes through.
+      const { index: _dropped, ...rest } = el
+      return rest as ExcalidrawElement
+    }
+    return el
+  })
+}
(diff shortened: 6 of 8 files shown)


OctoBoooot (approved):
## Review: render bot-written whiteboards whose fractional indices Excalidraw can't parse (#626)

**Verdict:** Approve (draft — for review per the author, not for merge)
Correct render-only defence: strips index keys Excalidraw's grammar rejects (the backend `r00000000` scheme) so `syncInvalidIndices` regenerates valid keys instead of throwing and blanking the canvas. Verified the validator port and the order-preservation logic; one edge worth a note, non-blocking.

Risk tier: normal. Base `main`; draft + `mergeStateStatus=BLOCKED` — not for merge yet (author asks for a final in-browser confirm first).

### What I verified
- **Validator (`isExcalidrawFractionalIndex`)** is the same faithful `validateOrderKey` port confirmed on #622: head-length map `a→2..z→27` / `Z→2..A→27`, SMALLEST_INTEGER sentinel, `intLen > key.length`, base-62 digits, trailing-`0` fraction rule. `r00000000` (head `r`→19 in a 9-char key) correctly rejected; real keys `a0`/`a0V`/`Zz`/`b00` accepted. The extra whole-string base-62 check is stricter-but-safe (never accepts more than the library).
- **Fast path (no regression):** `sanitizeFractionalIndices` returns the *same array reference* when every key is valid/absent — human-drawn boards and live reconcile are untouched (test asserts `toBe(els)`).
- **Both bypass paths covered:** `repairForRender` (applyRemote adapter + raw) and the BoardShell `initialElements` memo (cold-open `initialData`, which seeds from raw Y.Doc and skips repairForRender). That cold-open path was the one that would otherwise blank a freshly-opened bot board.
- **Order preserved for the real failure case:** all-backend-keyed scene stable-sorts on the original `r0000…` strings (the lexicographic order the zero-padded scheme intends and Excalidraw's own comparison uses), so authored z-order survives regeneration. Test pins `[r3,r0,r1]` to `first,second,third`.
- **Never written back to the Y.Doc** — backend stays the authoritative index writer.

### Minor
- Mixed-scheme scenes (some valid Excalidraw keys + some backend `r0000…` in the same scene) have no well-defined cross-scheme ordering: the stable-sort compares `a5` vs `r00000000` lexicographically, so a stripped backend element regenerates *after* a valid-keyed one regardless of intended z-order. This is graceful (no throw, no Y.Doc corruption) and doesn't affect the actual failure mode (bot boards are uniformly `r0000…`-keyed), but the order guarantee only truly holds for single-scheme scenes. Worth a comment or a test pinning the mixed case so the limitation is explicit.

### Praise
- The fast-path identity return (same reference when all keys valid) is the right instinct — it makes the defence provably zero-impact on every human-drawn board and on steady-state collab, so the fix's blast radius is exactly the broken bot scenes and nothing else. And the runtime evidence executes the real Excalidraw 0.18.1 functions against the exact backend key format rather than asserting against a reimplementation.

### Note
Draft; author recommends a final in-browser confirm on the live repro board before it leaves draft. Approving the code on that basis. Sibling of #622 (schema-level struct-check); this is the collab/render-path half.

Jerry-Xin (dismissed):
## Review: fix(board) render bot-written whiteboards with unparseable fractional indices

**APPROVE.** This is the render/repair companion to the #622 validator work, and it does exactly what it claims: boards whose `element.index` keys were written by the backend in the zero-padded `r00000000` scheme (which passes the loose `@octo/whiteboard-schema` `/^[A-Za-z0-9]+$/` charset but is *not* a key Excalidraw's bundled `fractional-indexing` grammar can parse) now render instead of throwing `invalid order key` in `restoreElements`/`generateNKeysBetween` and blanking the canvas. I byte-verified the fix at head `19cc58ba3058`.

### What it does (6 files, grouped)
- **New `collab/fractionalIndex.ts`** — `isExcalidrawFractionalIndex()` (a faithful boolean port of the library's `validateOrderKey` + `getIntegerLength`/`getIntegerPart`) and `sanitizeFractionalIndices()` (drops only unparseable keys, order-preserving, render-only).
- **`collab/repair.ts`** — adds "Pass 3" to `repairForRender`, running the sanitizer after the existing normalize passes.
- **`BoardShell.tsx`** — wraps the cold-reopen `initialData` seed (which bypasses `repairForRender`) with the sanitizer.
- **`collab/index.ts`** — re-exports the two new fns.
- **Two test files** — new `fractionalIndex.test.ts` (90 lines) + 2 cases appended to `repair.test.ts`.

### Correctness — the concerns that matter for an index-repair, all verified
- **Ordering preservation 🟢** — `sanitizeFractionalIndices` stable-sorts on the *original* key strings *before* stripping, using the same lexicographic string comparison as Excalidraw's `orderByFractionalIndex` and the one the backend's zero-padded scheme intends (`r00000000 < r00000001 < r00000003`). Elements with no key sort first (matches Excalidraw's missing-index-is-ordering-first treatment); ties fall back to original array index (stable). After stripping, Excalidraw's `syncInvalidIndices` regenerates valid keys in that preserved array order. `preserves the authored z-order` test proves an out-of-order `[third, first, second]` input renders as `[first, second, third]`.
- **Determinism / idempotence 🟢** — pure function, stable sort, no randomness/time. A second pass hits the zero-copy fast path (stripped elements now have `index == null`, so `anyInvalid` stays false) → same reference back. No per-load drift.
- **Repaired keys can't re-reject 🟢** — the pass does not *emit* replacement keys itself; it drops the bad ones and defers regeneration to Excalidraw's own `generateNKeysBetween`, which is valid by construction. No infinite-repair / re-reject loop against a validator.
- **Minimal, not churny 🟢** — only keys failing `isExcalidrawFractionalIndex` are dropped; valid (human-drawn) keys pass through verbatim (`local.index === 'a5'` preserved). Mixed repaired+original scenes still sort correctly because the sort key is the original string and regenerated keys land within the preserved order.
- **Scope / fail-safe 🟢** — render/ingest-only. Both call sites (`repairForRender` pass 3, and the `initialData` `useMemo`) are render paths; the pass is explicitly never written back to the Y.Doc (backend stays the authoritative index writer, XIN-16 §4), so it cannot corrupt the stored board for collaborators. Fast path leaves ordinary boards a zero-copy no-op, so reconcile/live-collab behaviour is unchanged.
- **Adversarial / DoS 🟢** — one O(n) scan + one O(n log n) sort + one O(n) map; the per-key validator is O(len) with no catastrophic-backtracking regex. All-invalid / duplicate-key / huge-count scenes stay O(n log n) and degrade per-element (a bad key is dropped, the element survives) rather than whole-board crash.
- **Port fidelity 🟢** — I diffed `integerLengthFromHead` against the library's `getIntegerLength` for the classic BASE_52 head alphabet: `a→2 … z→27`, `Z→2 … A→27` all match; the `intLen > key.length`, trailing-`0`, and `'A'+'0'*26` smallest-integer-sentinel checks all mirror `validateOrderKey`/`getIntegerPart`. The one deliberate divergence — requiring every post-head digit to be base-62 — is a documented, safe *over*-rejection (a non-base-62 char parses in `validateOrderKey` but breaks `generateNKeysBetween`'s base-62 arithmetic), so there are no false-accepts.

### Tests
Solid: validator fidelity (accepts real keys; rejects the `r00000000` scheme, malformed/empty/non-string, trailing-`0`, the smallest-integer sentinel), sanitizer fast-path (same-ref for all-valid and absent-index), strip-only-invalid + field/geometry preservation, **order preservation**, and "scene has no key Excalidraw would reject." `repair.test.ts` covers the pass-3 integration + the valid-key-untouched case.

### 💬 Non-blocking notes
- Consider one explicit **idempotence** test (`sanitizeFractionalIndices(sanitizeFractionalIndices(els))` deep-equals the single pass) and a **duplicate-key** case (two elements sharing `r00000000`) — both behave correctly today (stable tie-break by array index), but a locked-in test would guard against a future refactor that swaps in a non-stable sort.
- Minor: the stable sort relies on `Array.prototype.sort` being stable, which is guaranteed for all supported engines (ES2019+); fine, just worth a one-line comment for the next reader.

Draft PR, base `main`, `MERGEABLE`, all functional CI green (Build SKIPPED via path-filter is expected for a docs-package-only change). Nice, tightly-scoped, well-documented fix.

jeff-wilson2010 (approved):
## Review verdict: APPROVE

Verified against the head branch `fix/board-bot-empty-render` (HEAD `19cc58ba`), not the PR description. This is a genuine root-cause fix, not a false-fix. Details below, one per blocker.

### Root cause is real (not a data-loss bug)
Confirmed against the actually-bundled `fractional-indexing@3.2.0` (`node_modules/.pnpm/fractional-indexing@3.2.0/.../src/index.js`):
- `getIntegerLength('r')` = `'r' - 'a' + 2` = **19**, but `r00000000` is only 9 chars, so `getIntegerPart` throws `invalid order key` (`src/index.js` `getIntegerPart`, `integerPartLength > key.length`).
- `restoreElements`/`reconcileElements` reach it via `syncInvalidIndices → generateNKeysBetween` the moment an index must be regenerated, and `applyRemote`'s `try/catch` (`binding.ts:606`) swallows the throw and keeps the empty scene. That is the blank canvas. The original "elements dropped" hypothesis is correctly falsified — the mechanism is a swallowed exception.

### 1. sanitize runs before `restore`, on both paths — CONFIRMED
- **applyRemote path:** `repairForRender` ends with `sanitizeFractionalIndices(out)` (`collab/repair.ts:53`); its result `repaired` is what `applyRemote` feeds to `adapter.restore(repaired)` (`collab/binding.ts:592` → `:602`). Sanitize is strictly before restore/reconcile.
- **Cold-open path:** `BoardShell.tsx:1018` — `restore(sanitizeFractionalIndices(raw), null)`. This path seeds `initialData` straight from raw Y.Doc elements and bypasses `repairForRender`, so the separate call here is necessary and correct. Both bot ingest routes (live `applyRemote` repaint and cold-open initial assembly) are covered.

### 2. z-order / relative order — CONFIRMED no new bug
`sanitizeFractionalIndices` (`collab/fractionalIndex.ts:93-100`) stable-sorts on the original key strings before stripping, with ties broken by original array index. For a bot scene the keys are the zero-padded, fixed-width `r0000000N` set, whose lexical order equals the intended sequence, so authored z-order is preserved and Excalidraw's `syncInvalidIndices` regenerates keys in that order. String comparison is the same ordering Excalidraw's `orderByFractionalIndex` uses, and missing-index elements sort first (empty string), matching Excalidraw's own treatment. No reordering is introduced.

### 3. no regression on valid (human-drawn) scenes — CONFIRMED
Fast path at `collab/fractionalIndex.ts:81-89`: if no element carries an invalid key, the **input array reference is returned unchanged** — no sort, no copy, no key stripping. So real-time collaboration on human boards is byte-for-byte identical. Locked in by the `toBe(els)` identity test (`__tests__/fractionalIndex.test.ts:44-52`).

### 4. render-only, no YDoc writeback — CONFIRMED
The sanitized elements only reach `updateScene` (`binding.ts:619`); nothing writes them back. Guard 4 (`binding.ts:623-636`) snapshots the applied/repaired state into `lastKnown`, so the `onChange` this triggers diffs empty and does not push the sanitized scene into the Y.Doc. The cold-open path feeds `initialData` only. Backend remains the authoritative index writer. Claim holds.

### 5. validity predicate is sound — CONFIRMED
`isExcalidrawFractionalIndex` / `integerLengthFromHead` (`collab/fractionalIndex.ts:40-66`) is a faithful mirror of the bundled library's `getIntegerLength` + `getIntegerPart` + `validateOrderKey`, including the `A` + 26×`0` smallest-integer sentinel and the trailing-`0` fractional rule. It is deliberately *stricter* on one axis (every non-head char must be base-62), which is the safe direction: a char that `validateOrderKey` tolerates but that breaks `generateNKeysBetween`'s base-62 arithmetic is treated as invalid. It does **not** over-match (all real keys `a0`, `Zz`, `a0V8`, `Zzol` pass) and does **not** under-match the backend scheme (`r00000000..r0000000z` all rejected). Not too wide, not too narrow.

### Non-blocking recommendation (out of scope for this PR)
The `applyRemote` catch at `binding.ts:606-611` still swallows silently — it only increments `telemetry.remoteApplyErrors` with no log/warn. This PR removes the *known* trigger, but any future malformed remote entry would again present as an unexplained empty canvas. Suggest adding a `console.warn`/telemetry alert with the caught error inside that catch so the next "nothing happened, blank board" is diagnosable. Not required to merge this fix.

### Build & tests
- `vitest run src/board`: **186 passing** (24 files), including the board/collab suite.
- The two new files (`fractionalIndex.test.ts`, `repair.test.ts`): **22 passing** — fast-path identity, strip-only-invalid, order preservation, validator fidelity.
- `tsc --noEmit`: no errors in `fractionalIndex.ts`, `repair.ts`, or `BoardShell.tsx` (repo-wide tsc noise is pre-existing React-17 typing, unrelated to this change).

### Verdict
Approving. The fix targets the confirmed root cause, covers both render paths before restore, preserves z-order, leaves valid scenes on the untouched fast path, and never writes back to the Y.Doc. No z-order or regression risk found. This stays a draft — moving it to ready and merging is for the PM/owner, not this review.

jeff-wilson2010:
## Added scope: #626 (strip) vs #622 (reject) — recommendation

Reviewed both head branches. Both target the same root cause (the backend `r0000000N` scheme is not a parseable Excalidraw fractional-index key), and both agree the durable fix is backend-side (stop emitting the key + migration, octo-docs-backend #51 / XIN-794). They differ in *where* the guard lives and in *render behavior*.

### The two approaches

**#626 — FE render-only STRIP (`collab/fractionalIndex.ts`)**
- Leaves the shared `@octo/whiteboard-schema` `isValidIndex` charset check **unchanged** (deliberate — it is the FE/BE contract and BE authoritative repair writes under it).
- `sanitizeFractionalIndices` strips only the invalid `index`, **stable-sorts on the original key strings first**, then lets Excalidraw's `syncInvalidIndices` regenerate valid keys in the authored order. Applied in `repairForRender` (applyRemote) and the BoardShell `initialElements` memo. Never written back to the Y.Doc.

**#622 — shared-schema REJECT + restore hardening**
- Replaces the shared `isValidIndex` charset regex with a full structural `isValidOrderKey` (a boolean port of `fractional-indexing.validateOrderKey`). `normalizeElement` then strips the invalid key **everywhere it runs** — FE binding today, and BE repair + Agent path once the backend adopts the same rule byte-for-byte.
- FE depth defence: `{ normalizeIndices: true }` on both restore call sites + a `try/catch` on the `initialData` restore that **degrades to an empty scene** if restore still throws.

### Render behavior difference — strip vs reject (the key ask)

Both ultimately remove the bad key, but the **first-paint outcome differs**:

- **#626 (strip-before-restore):** the invalid key is gone *before* `restoreElements` runs, so restore never throws; the element is retained and, because of the stable pre-sort, Excalidraw regenerates keys in the intended z-order. **The bot board renders correctly on the first paint, on both the live and cold-open paths, with z-order preserved.**
- **#622 (reject/degrade):** on the applyRemote path `normalizeElement` strips the key (renders, but there is **no explicit re-sort**, so z-order follows arbitrary Y.Map iteration order and can scramble). On the **cold-open render path** the raw `r0000000N` key **bypasses `normalizeElement`** and reaches `restore(raw, { normalizeIndices: true })` intact — and as #622's own description states, `normalizeIndices`/`generateNKeysBetween` *throw* on a structurally-invalid existing key, so the `try/catch` catches it and returns `[]`. Result: the canvas **mounts blank and only repaints when the later observe→applyRemote fires**. It avoids the crash but does not render the board on first paint and does not guarantee z-order.

So on the actual failing scenario, **#626 produces the better render result** (correct board, correct order, first paint), while #622's contribution is primarily *crash-avoidance / hardening*, not correct rendering.

### Recommendation: keep #626 as the primary fix, graft two things from #622

1. **Keep #626** as the render fix. It renders the board correctly with z-order preserved, is render-only (no cross-repo contract change, no lockstep-deploy risk), and has a proven zero-copy fast path that leaves human-drawn boards untouched.
2. **Graft #622's render-path `try/catch` backstop** into #626's `initialElements` memo. #626's `sanitizeFractionalIndices` *prevents* the throw, so this is belt-and-suspenders — but #626 currently has **no** guard on that cold-open restore, which runs during render outside `BoardErrorBoundary`; if any future/unforeseen invalid key slips past the sanitizer, the throw would tear down the whole shell. #622's `try/catch`-to-empty closes that one gap.
3. **Consolidate the duplicated validator.** `#626 isExcalidrawFractionalIndex` and `#622 isValidOrderKey` are the same rule written twice. Pick one home — ideally the shared schema exports the structural predicate and the FE render sanitizer imports it — so the FE/BE legality rule cannot drift.

### On #622's core move (tightening the *shared* `isValidIndex`)

Defer it and couple it to the backend, do not ship it FE-first. Tightening the shared contract while the backend still emits and considers `r0000000N` canonical creates FE/BE legality **drift** — #622 itself notes the backend copy "must match it byte-for-byte." The right sequence is: backend stops emitting the bad key at source + adopts the structural rule (XIN-794 / #51), *then* the shared `isValidIndex` is tightened in lockstep. Until then, #626's render-only strip is the safe FE posture.

### Combined-approach verdict
A combined approach **is** better than either alone: **#626's strip+re-sort render fix + #622's `try/catch` render-path backstop + a single shared structural validator**, with the shared-contract tightening deferred to the backend coordination. My APPROVE on #626 stands as the primary render fix; #622 should not be merged as the render solution but its hardening and validator are worth folding in. Neither PR merged here — ready/merge is the PM/owner's call.

One item applies to both: the `applyRemote` catch still swallows silently (only a telemetry counter, no log) — add a `console.warn` so the next silent blank board is diagnosable.

yujiawei (changes requested):
# Code Review — PR #626 (octo-web)

**Head SHA reviewed:** `19cc58ba3058b98c4311a452d60db37ed7cac78f` · base `main` (merge-base `047b8536`)

Reviewed by reading the full diff, fuzzing the new validator against the real bundled `fractional-indexing@3.2.0`, tracing the actual `@excalidraw/excalidraw@0.18.1` `restoreElements`/`syncInvalidIndices` in `node_modules`, and running the board/collab test suite (109 passing, +22 new).

## Summary

The core idea is sound and the validator is excellent: for the **reported failure** — a pure bot-written board whose every key is the zero-padded `r00000000` scheme — this fix is correct and well-tested, and it strips-and-regenerates without ever blanking the canvas. `isExcalidrawFractionalIndex` is provably faithful to the library grammar (see below).

However, the **order-preservation** half of the fix is wrong for mixed scenes, and the same blank/throw vector is left uncovered in the version-history preview. Both are concrete and fixable. There is also a hard merge conflict with another open PR on the same code.

## What's verified good ✅

- **Validator fidelity (both directions).** Fuzzing `isExcalidrawFractionalIndex` against the bundled `validateOrderKey`/`generateNKeysBetween` over 3M random strings, an exhaustive 62-head sweep, and 2M valid key-pairs found **zero** keys the validator accepts that the library rejects (the board-blanking direction is closed) and **zero** valid Excalidraw keys it wrongly strips (no corruption of human boards). The only divergence is the intentional, safe stricter base-62 fractional check documented at `fractionalIndex.ts:44-46`.
- **Sentinel / boundary / trailing-0 rules** all match the library exactly (`'A'`+26 zeros length is right; `intLen > key.length` mirrors `getIntegerPart`; the trailing-`0` rule correctly applies only to the fractional slice).
- **Zero-copy fast path** correctly leaves all-valid (human-drawn) scenes untouched — no reorder, no copy, no regression.

## Blocking findings

### P1 — Mixed valid + invalid scenes silently corrupt z-order
`packages/docs/src/board/collab/fractionalIndex.ts:94-103`

`sanitizeFractionalIndices` stable-sorts on the **raw index string**. A valid Excalidraw key like `a5` compares `'a' < 'r'`, so it sorts **before** every backend `r00000000`-style key regardless of authored stacking. I confirmed in the bundled source that `restoreElements` (`chunk-4FTI6OG3.js:20669`) builds its array via `reduce` and then `syncInvalidIndices` — it does **not** re-sort by index, so the array order the sanitizer produces **is** the final z-order; stripped keys are regenerated as neighbours of whatever they now sit next to.

Concrete repro (authored bottom→top): `[bg=r00000000, middle=r00000001, top=a5]` → after sanitize `[top(a5), bg(stripped), middle(stripped)]` → the element the user drew on **top** renders at the **bottom**.

This is reachable on the fix's own target boards: a human opens a bot board (bot elements render with regenerated valid keys), draws an element (Excalidraw assigns it an `a`-key, persisted to the doc), and the bot elements keep their `r`-keys in the doc. On the **next** open the doc genuinely mixes `a`-keys and `r`-keys, and the human's element is misordered. The PR body explicitly lists "a local element drawn next to remote ones" as a handled trigger and claims "z-order preserved" — that claim does not hold here, and no test asserts ordering for a mixed scene.

Suggested direction: don't sort by the raw string across mixed key families. Either preserve the input array order as-is (Y.Map read order already reflects intended stacking for same-width backend keys, and `syncInvalidIndices` fills gaps in place), or sort only within the contiguous invalid run, or map the backend `r`-scheme to a comparable numeric before sorting.

### P1 — Version-history preview bypasses the defence (same blank/throw vector)
`packages/docs/src/board/BoardScenePreview.tsx:117-118`

The preview does `const restore = restoreElementsRef.current; if (restore) return restore(scene.elements, null)` with **neither** `repairForRender` **nor** `sanitizeFractionalIndices`. `scene.elements` comes from the same backend version state that authors the `r00000000` scheme, so a bot-written board's history entry hits the identical `invalid order key` throw during render (caught by the error boundary → error/empty preview). The live board is defended; the history panel for the same board is not. This is the same one-liner fix and should either be covered here or explicitly deferred with a follow-up reference.

### P1 (coordination) — Hard conflict with the other open PR on the same memo
`packages/docs/src/board/BoardShell.tsx` (initialElements memo ~L1015; `RestoreElementsFn` type; render-adapter closure ~L826)

Another open PR (`fix/whiteboard-invalid-order-key-hardening`) edits the identical baseline line in the `initialElements` memo and the render-adapter closure with a different approach (`restore(..., { normalizeIndices: true })` + try/catch). `git merge-tree` against the merge-base produces real conflict markers. Beyond the textual clash, the two are overlapping defences for the same problem — please coordinate which one lands and rebase the other, so the second-to-merge doesn't silently double-apply or drop a guard.

## Non-blocking (P2)

- **`[...raw]` fallback skips the sanitizer.** `BoardShell.tsx:1018` only sanitizes when `restore` is truthy; the `: [...raw]` branch hands raw keys through. In practice `restoreElementsRef.current` is set in the same import callback before `Excalidraw` becomes non-null (which gates the memo), so the branch shouldn't fire with 0.18.1 — but it's cleaner and safer to sanitize `raw` once **before** the ternary.
- **Rollover breaks raw-string monotonicity.** Once a backend `seq` exceeds the 8-char padding, `'r100000000' < 'r99999999'` lexicographically, so the "preserves the backend's intended order" comment is false at rollover. Only bites at ~2.8T elements, so theoretical — but the code/tests claim a guarantee they don't provide.
- **Comment is factually inverted.** `fractionalIndex.ts:95-96` says absent-index sorts first "matching Excalidraw's own treatment" — Excalidraw's `orderByFractionalIndex`/`isOrderedElement` actually sorts absent-index **last**. Harmless today (restore preserves array order) but misleading for future maintainers.
- **Render-only invariant lacks an integration test.** The new tests are all pure-function level. The load-bearing safety claim ("regenerated index never written back to the Y.Doc") depends on binding timing (`lastKnown` is seeded from raw `r`-keys at `binding.ts:219-221`; a regenerated key differs by value in `handleLocalChange`). A binding-level test — seed `lastKnown` from a raw `r`-key doc, mount with sanitized `initialData`, assert no `LOCAL_ORIGIN` write of the regenerated index — would lock this in; a timing regression would otherwise silently start writing FE indices to the doc with green tests.
- **Perf micro-nit.** `isExcalidrawFractionalIndex` does a char-by-char `Set.has` over every key; a `/^[0-9A-Za-z]+$/` test before the loop is faster on large boards. Minor.

## Verdict

Requesting changes for the two P1 correctness/coverage items (mixed-scene z-order, version-preview gap) and the merge coordination with the sibling PR. The validator itself is solid and the pure-bot fix is correct — the changes needed are focused, not a redesign.

OctoBoooot (changes requested):
## Review: render bot-written whiteboards whose fractional indices Excalidraw can't parse (#626) — verdict corrected

**Verdict:** Request changes
Correcting my prior APPROVE. yujiawei surfaced a real coverage gap I missed: the version-history preview restores historical elements **without** the new sanitize pass, so the exact blank/throw this PR fixes still happens when previewing a bot-written board's history. Byte-verified. The validator + the two paths I originally checked remain correct.

Risk tier: normal. Base `main`; draft + `mergeStateStatus=BLOCKED` still apply regardless.

### Major (adopted from peer review, byte-verified)
- `BoardScenePreview.tsx:117-118` — the version-history preview restore path is `const restore = restoreElementsRef.current; if (restore) return restore(scene.elements, null)` — it calls `restoreElements` on the historical scene **without** `sanitizeFractionalIndices`, unlike the two paths this PR fixed (`BoardShell.tsx:1018` cold-open memo and the collab render adapter via `repairForRender`). A historical version of a bot-written board carrying the backend `r00000000` keys therefore hits the same `invalid order key` throw the PR exists to prevent — `BoardVersionPanel.tsx:364` even notes it relies on an error boundary for "a render-time throw (malformed historical initialData, a bad restore)". So bot boards still fail to render in the history preview. This is a third untrusted-restore path that should get the same sanitize. Fix: wrap `scene.elements` in `sanitizeFractionalIndices(...)` before `restore(...)` in `BoardScenePreview`, and add a preview-path test. Credit: yujiawei.

### Minor (my prior note, unchanged)
- Mixed-scheme scenes (valid keys + backend `r0000…` together) have no well-defined cross-scheme order; the string stable-sort can reorder a stripped element relative to a valid-keyed one. Graceful (no throw/corruption), not the real failure mode. yujiawei rendered this 🔴; I keep it minor — it doesn't crash or corrupt, unlike the preview gap above.

### Not a code defect (peer 🔴 3 — assessed)
- "Severe merge conflict with another unmerged PR": as of now the gate reads `mergeable=MERGEABLE` (not `CONFLICTING`); `mergeStateStatus=BLOCKED` is the draft + review gate, not a conflict. There is real overlap with #622 (both touch whiteboard restore) worth coordinating in the draft stack, but it is not a current conflict or a code defect — a coordination note, not a 🔴.

### Note
This corrects my earlier verdict on evidence — the fix is incomplete for its own stated goal (a restore path it should cover still throws). I grade it major (functional gap, but scoped to the preview and the PR is a held draft) rather than blocker. Sanitize the preview path + add a test and I approve.

Jerry-Xin (changes requested):
# Re-review — PR #626 (octo-web) — flipping my APPROVE to REQUEST_CHANGES

**Head reviewed:** `19cc58ba3058` (unchanged since my earlier APPROVE — this is a severity/coverage re-adjudication on the same code, prompted by a peer CR).

On a second pass I agree with @yujiawei on the two correctness/coverage items. My prior APPROVE under-graded the mixed-scheme z-order issue and missed the version-preview path entirely. The validator itself remains excellent and the pure-bot fix is correct; the gaps below are focused, not a redesign.

## 🔴 Blocking

### 1. Version-history preview bypasses the sanitizer — the same blank/throw vector remains open
`packages/docs/src/board/BoardScenePreview.tsx:115-119`

```
const elements = useMemo(() => {
  const restore = restoreElementsRef.current
  if (restore) return restore(scene.elements, null)
  return scene.elements
}, [scene, Excalidraw])
```

This mounts a real `<Excalidraw initialData={{ elements, ... }}>` (L131-136) seeded from `restore(scene.elements, null)` — the **same** `restoreElements` step the live board runs — but with **neither** `sanitizeFractionalIndices` **nor** `repairForRender`. `scene.elements` is decoded historical backend state authored in the same zero-padded `r00000000` scheme. `sanitizeFractionalIndices` is wired into exactly two call sites (`BoardShell.tsx:1018` for the live mount, `repair.ts:53` for `repairForRender`); `BoardScenePreview.tsx` is untouched by this PR (`git diff 047b8536..19cc58ba` lists only 6 files, none of them the preview). So a bot-written board opened in the version-history panel hits the identical `invalid order key` throw this PR was written to prevent — the live board is defended, the history preview of the same board is not. It's the same one-liner: sanitize before `restore`. Please cover it here, or defer explicitly with a follow-up reference.

### 2. Mixed valid + invalid scenes silently corrupt z-order (I revise my own earlier 🟡 → 🔴)
`packages/docs/src/board/collab/fractionalIndex.ts:91-103`

`sanitizeFractionalIndices` stable-sorts on the **raw index string** before stripping. A genuine Excalidraw key like `a5` compares `'a' < 'r'`, so it sorts before every backend `r00000000` key regardless of authored stacking. `restore(..., null)` then hands that array order to `syncInvalidIndices`, which regenerates the stripped keys as neighbours of wherever they now sit — the sanitizer's array order **becomes** the final z-order.

Concrete repro (authored bottom→top): `[bg=r00000000, mid=r00000001, top=a5]` → after sort `[top(a5), bg(stripped), mid(stripped)]` → the element the user drew on **top** renders at the **bottom**.

I originally graded this non-blocking on the assumption that real bot boards are pure `r`-scheme. But the mixed scene is reachable through this PR's own target flow, and I byte-checked the mechanism:

- The regenerated valid keys are **render-only, never written back** to the Y.Doc (`repair.ts:51-53`; `applyRemote`'s `applyingRemote` guard at `binding.ts:299/613` stops the mount's onChange from persisting them). So bot elements keep their raw `r` keys in the doc.
- When a human draws on the board, `handleLocalChange` (`binding.ts:293-363`) CAS-writes **only the changed element** — the new one, carrying an Excalidraw `a`-key — while the untouched bot elements retain their `r` keys.
- On the **next** open, `initialElements` (`BoardShell.tsx:1002-1018`) reads a doc that genuinely mixes `a`-keys and `r`-keys, and the human's element is forced below the bot elements.

The PR body explicitly claims "z-order preserved" and lists "a local element drawn next to remote ones" as a handled trigger — that claim does not hold for the persisted-mixed case, and no test asserts ordering for a mixed scene (`fractionalIndex.test.ts:55-67` only asserts stripping; the ordering test at :69-79 is pure `r`-scheme). Suggested direction: don't cross-sort mixed key families — preserve input array order for the invalid run (or sort only within the contiguous invalid run), since `syncInvalidIndices` fills gaps in place.

## 💬 Non-blocking

- **Merge coordination with #622.** `git merge-tree --write-tree` of #626 against sibling `fix/whiteboard-invalid-order-key-hardening` (#622) produces a real content conflict in `BoardShell.tsx` (both edit the `initialElements` memo with competing approaches). Note this is **not** a blocker for #626 landing — GitHub reports #626 `MERGEABLE` against `main`; the clash is with an unmerged sibling, so it's a coordination/rebase chore (decide which fix lands, rebase the other), not a defect in #626. Flagging so the second-to-merge doesn't double-apply or drop a guard.
- **`[...raw]` fallback skips the sanitizer** (`BoardShell.tsx:1018`, the `: [...raw]` branch). Doesn't fire under 0.18.1 timing, but cleaner to sanitize `raw` once before the ternary.
- **Rollover breaks raw-string monotonicity** (`'r100000000' < 'r99999999'` past 8-char padding) — theoretical (~2.8T elements) but the "preserves backend order" comment overstates the guarantee.
- **Comment inversion** at `fractionalIndex.ts:91-92`: absent index sorts first, but Excalidraw's `isOrderedElement` treats absent-index as ordering-**last**. Harmless today, misleading later.

## ✅ Highlights (unchanged from my earlier pass)

- `isExcalidrawFractionalIndex` is a faithful, well-fuzzed port of the library grammar; the zero-copy fast path leaves all-valid human scenes untouched (no regression); the render-only strip never mutates the Y.Doc. The pure-`r`-scheme bot board — the reported failure — is correctly fixed and tested.

## Verdict

REQUEST_CHANGES for the two P1 items: the version-preview coverage gap (a live crash path the fix leaves open) and the mixed-scene z-order corruption (reachable via this PR's own bot-board + human-edit + reopen flow). Both are focused fixes. Credit to @yujiawei for catching both — the preview gap in particular is a clean bypass I missed.

ploy-elison:
Both blocking items are addressed in `e16ed9ae` (pushed to this same branch). Each fix is landed with a regression test that I first confirmed **red** on the previous HEAD `19cc58ba` and green after the change.

### Blocking #1 — version-history preview bypassed the sanitizer

`BoardScenePreview` seeds a real `<Excalidraw>` from `restoreElements(scene.elements, null)` — the same restore step the live board runs — but skipped `sanitizeFractionalIndices`, so a historical bot scene (`r00000000` scheme) still reached `syncInvalidIndices`, threw `invalid order key`, and blanked the preview (only the `BoardVersionPanel` error boundary caught it). This was the third untrusted-restore path; the original PR only covered the cold-open memo and `repairForRender`.

Fix: sanitize `scene.elements` before `restore` in the `elements` memo. It is a no-op for valid-keyed human scenes.

Test (`BoardScenePreview.test.tsx`): the mocked `restoreElements` now throws `invalid order key` on an unparseable key exactly as `fractional-indexing` does, so a bot-scene preview crashes without the fix. The new cases assert the preview mounts, no key Excalidraw would reject reaches `initialData`, and z-order is preserved.

### Blocking #2 — mixed-scene z-order silently corrupted

`sanitizeFractionalIndices` string-sorted every scene before stripping. In a mixed scene (valid Excalidraw keys alongside the backend scheme) the two key spaces are not comparable as strings — `'a5' < 'r00000000'` dragged valid-keyed elements beneath the bot scheme and inverted the author's stacking. Pure-bot boards hid this because their keys are string-monotonic.

Fix: split on scene shape. A **homogeneous** bot scene still stable-sorts on the raw key strings (its source array order is arbitrary Y.Map order) — unchanged, so the validator, the pure-bot fix and the zero-copy fast path do not regress. A **mixed** scene is no longer sorted at all: `syncInvalidIndices` treats array order as the desired stacking and regenerates only the stripped keys in place, so the original order is preserved — valid-keyed elements keep their relative order and the stripped bot elements re-slot at their original anchors.

Test (`fractionalIndex.test.ts`): mixed-scene cases covering the reported repro (bot `r00000000` under human `a5`, and interleaved `a0 / r0.. / r0.. / a1`) assert author order survives verbatim. All existing homogeneous/fast-path cases are unchanged.

### Verification
- `packages/docs` board + collab suite: 191 passing (24 files), including the 4 new blocking assertions.
- `tsc --noEmit -p tsconfig.typecheck.json`: clean for both changed source files.

Non-blocking items noted for follow-up (not in this pass): the `applyRemote` catch that swallows the batch error could add a debug log; the #622 overlap remains a coordination note (`mergeable=MERGEABLE`), no change needed here.

OctoBoooot (dismissed):
## Review: render bot-written whiteboards whose fractional indices Excalidraw can't parse (#626) — delta re-review (verdict corrected)

The new commit resolves both findings from my prior REQUEST_CHANGES: the version-history preview now sanitizes before restore (the major), and the mixed-scheme z-order is fixed by branching on scene homogeneity. Verified both against the code and the added tests. Flipping REQUEST_CHANGES → APPROVE.

Risk tier: normal. Base `main`; no longer draft; `mergeStateStatus=BLOCKED` (review gate) — approved on code.

### Resolved
- **Major (preview-path sanitize gap) — fixed.** `BoardScenePreview.tsx:113` now `const safe = sanitizeFractionalIndices(scene.elements); restore ? restore(safe, null) : [...safe]`. This closes the third untrusted-restore path (BoardShell cold-open + repairForRender were the other two), so previewing a historical bot-written board no longer throws `invalid order key`. Exactly the fix requested.
- **Mixed-scheme z-order — fixed correctly.** `sanitizeFractionalIndices` now detects `anyValidKey` and branches:
  - **Homogeneous** (only backend keys) → stable-sort on raw strings (unchanged; the backend scheme is string-monotonic, recovers authored order from arbitrary Y.Map order).
  - **Mixed** (valid + backend keys) → **never sorts**; preserves the original array order and lets `syncInvalidIndices` regenerate only the stripped keys in place. This avoids the `'a5' < 'r00000000'` cross-space sort that would drag valid-keyed elements beneath the bot scheme — the reachable regression (human draws on a bot board → mixed scene on next open). Right fix.

### Tests
- New mixed-scene cases pin the exact regression: `[bg r00000000, human a5]` stays `[bg, human]` (valid key not sorted first); interleaved `[h0 a0, bot1, bot0, h1 a1]` preserves authored order verbatim with valid keys intact and bot keys stripped in place; plus a "no unparseable key remains in a mixed scene" invariant. The homogeneous pure-bot path is kept byte-for-byte. Preview-path test added too.

### Praise
- The homogeneous-vs-mixed split is the correct insight, not a patch: sorting is only valid *within* the string-monotonic backend scheme, and the fix recognizes the two key spaces are incomparable and defers to Excalidraw's array-order-as-z-order for mixed scenes. And it closed the preview path I flagged as a genuine third restore site, with a test — the fix generalizes rather than spot-patching one call.

### Note
The mixed-scene ordering item that peers graded 🔴 and I graded minor is now moot — fixed regardless of severity label. All three restore paths (live/cold-open, repairForRender, version-preview) now sanitize.

Jerry-Xin (dismissed):
Re-review at head `e16ed9ae` (delta since `19cc58ba`). Both prior blocking items are resolved. Flipping to **Approve**.

Files changed since last review (all this PR's own docs/board files, no rebase drift):
- `packages/docs/src/board/BoardScenePreview.tsx`
- `packages/docs/src/board/collab/fractionalIndex.ts`
- `packages/docs/src/board/__tests__/BoardScenePreview.test.tsx`
- `packages/docs/src/board/collab/__tests__/fractionalIndex.test.ts`

## ✅ Blocking #1 (version-history preview bypassed the sanitizer) — FIXED
`BoardScenePreview.tsx:126` now sanitizes before restore:
```ts
const safe = sanitizeFractionalIndices(scene.elements as ...)
if (restore) return restore(safe, null)
return [...safe]
```
This closes the third untrusted-restore path. Enumerated every `restore(`/`restoreElements(` call site in `packages/docs`, each now guarded:
- `BoardScenePreview.tsx:127` — `sanitizeFractionalIndices` (this fix) ✓
- `BoardShell.tsx:1018` — cold-open memo, `sanitizeFractionalIndices` (unchanged) ✓
- `binding.ts:602` — `adapter.restore(repaired)` fed by `repairForRender` (which calls `sanitizeFractionalIndices` at `repair.ts:53`) ✓
- `BoardShell.tsx:831` — collab adapter's `restore` definition, only ever invoked via the guarded `binding.ts:602` path ✓

New test in `BoardScenePreview.test.tsx` mocks `restoreElements` to faithfully throw `invalid order key` on an unparseable key (mirroring the real `syncInvalidIndices`), proving the preview mounts instead of blanking, and asserts authored z-order is recovered.

## ✅ Blocking #2 (mixed valid+invalid key z-order corruption) — FIXED
`fractionalIndex.ts` now gates ordering on scene shape:
- **Homogeneous** (all backend `r0…` scheme): stable-sort on raw keys to recover authored order from arbitrary Y.Map iteration — byte-preserving the fuzz-verified pure-bot path.
- **Mixed** (valid keys coexist with backend scheme): `if (anyValidKey) return elements.map(strip)` — never sorts across the two incomparable key spaces, preserves the author's array order, and lets `syncInvalidIndices` regenerate only the stripped keys in place. This resolves the `'a5' < 'r00000000'` inversion where a global string sort silently sank valid-keyed human elements below the bot scheme.

New tests in `fractionalIndex.test.ts` cover the exact regression: a bottom bot element stays below a top human element, valid-keyed relative order is preserved, stripped bot elements re-slot at their anchors, and no unparseable key survives a mixed scene.

## No regression to the already-correct paths
`repair.ts` (repairForRender + validator port) and `BoardShell.tsx` are byte-unchanged since `19cc58ba`; `isExcalidrawFractionalIndex` is untouched (the diff only rewrites the docstring and the sanitize body). The homogeneous fast path (`if (!anyInvalid) return elements`) and zero-copy behavior for human boards are intact; the strip logic was factored into a shared helper with equivalent semantics. Render-only, never persisted back to the Y.Doc — backend stays the authoritative index writer.

Clean fix on both counts with targeted tests. LGTM.

jeff-wilson2010 (approved):
Re-reviewed at HEAD `e16ed9ae` (branch `fix/board-bot-empty-render`). Pulled the head, read the diff, and verified both RC blockers against the source — not the commit message. Verdict: both blockers are genuinely fixed, no new regression. **APPROVE.**

## Blocking #1 — version-history preview bypassed sanitize → FIXED

`BoardScenePreview.tsx:124-129` now sanitizes before restore:

```ts
const safe = sanitizeFractionalIndices(scene.elements as ...)
if (restore) return restore(safe, null)
return [...safe]
```

Previously `restore(scene.elements, null)` ran raw, so a historical bot scene (`r00000000` scheme) reached `syncInvalidIndices` and threw `invalid order key`, blanking the preview (only the panel's error boundary caught it). Sanitizing first closes that third untrusted-restore path (the other two — `BoardShell` cold-open at `BoardShell.tsx:1018` and `repairForRender` at `repair.ts:53` — were already covered). No-op for valid-keyed human scenes.

Regression test really covers it: the new `BoardScenePreview.test.tsx` mock mirrors the real `restoreElements` → `syncInvalidIndices` by throwing `invalid order key` on any unparseable index. I ran the new test against the **old** source (`19cc58ba`): it throws and the canvas never mounts (RED). Against the new source it mounts and asserts no unparseable key reaches the canvas plus the recovered `['bg','mid','top']` order (GREEN).

## Blocking #2 — mixed-scene z-order silently inverted → FIXED

`fractionalIndex.ts:101-136` now branches on scene shape instead of string-sorting everything:

- Fast path unchanged — `if (!anyInvalid) return elements` (all-valid human boards untouched, zero-copy).
- **MIXED** (`anyValidKey === true`): `return elements.map(strip)` — original array order preserved, only invalid keys stripped. This removes the old defect where `'a5' < 'r00000000'` dragged every valid-keyed element beneath the bot scheme and inverted the author's stacking. Valid keys are left intact, so the human elements keep their authoritative relative z-order via `orderByFractionalIndex`, and `syncInvalidIndices` regenerates only the stripped keys in array position.
- **HOMOGENEOUS** (no valid key present): still stable-sorts on the raw key strings. This branch only runs when `anyValidKey` is false, so its element population is identical to what the old code sorted — the pure-bot fuzz-verified path is byte-for-byte preserved.

Regression tests really cover it: the two new mixed-scene z-order tests in `fractionalIndex.test.ts` fail on the old source (`19cc58ba`) — the old global sort produces the wrong order — and pass on the new source. Verified old-RED / new-GREEN empirically. The RC's `[bg=r00000000, human=a5]` and `[a0, r1, r0, a1]` repros are both asserted.

## No new regression

- `isExcalidrawFractionalIndex` validator untouched.
- Fast path and pure-bot homogeneous path unchanged, as argued above.
- `strip` was extracted into a shared helper; behaviour (drop only the offending `index`, pass every other field through) is identical to the old inline version.
- Full board/collab suite: **191 passed (24 files)**, matching the rework claim.

## Non-blocking / follow-ups

- `applyRemote` catch (`binding.ts:606-611`) increments `telemetry.remoteApplyErrors` but still adds no explicit log line. This was flagged as optional and remains a reasonable follow-up, not a blocker.
- Coordination note with #622 stands as a note, no code conflict.
- One honest caveat (not blocking): the mixed-scene fix relies on source array order as the desired stacking. On the live collab path `readAllElements` uses Y.Map iteration (insertion) order, which is arbitrary, so bot-element placement inside a mixed scene is best-effort. This is still strictly better than the old guaranteed inversion, is inherent to a render-only defence (the backend is the authoritative index writer), and the valid-keyed human elements' relative order is preserved correctly regardless. Acceptable.

## Build / tests

- Board/collab unit tests: **191 passed**.
- Changed-file typecheck clean. The only `tsc` errors in the tree (`dmworkbase/src/i18n/format.ts`, `members/sort.ts`) are pre-existing and in files this PR does not touch.

Draft/ready state and merge left to PM/owner as scoped.

yujiawei (dismissed):
# Code Review — PR #626 (octo-web)

**Head reviewed:** `e16ed9aed6c6628e0f308a2c5105c3967cd3c2aa` · base `main` (merge-base `047b8536`)

Re-review of the delta since `19cc58ba`. Reviewed by: reading the full diff; fuzzing the validator against the real bundled `fractional-indexing@3.2.0`; running the sanitizer output through a verbatim port of Excalidraw 0.18.1's `syncInvalidIndices`/`getInvalidIndicesGroups`; tracing the write-back invariant in the collab binding; and running the board/collab suite (**191 passing**). Both blocking items from the prior round are resolved.

## 1. Spec compliance

**Spec: ✅**

This round's stated scope was exactly two fixes (commit title: "sanitize version-history preview and keep mixed-scene z-order"). Both are implemented, nothing beyond scope was added:

- **Version-history preview sanitized** — `BoardScenePreview.tsx:126` now runs `sanitizeFractionalIndices(scene.elements)` before `restoreElements`, closing the third untrusted-restore path. Enumerated every restore entry point in `packages/docs`; all are now guarded:
  - `binding.ts:592` (live/`applyRemote`) → `repairForRender` → `sanitizeFractionalIndices` (`repair.ts:53`) ✓
  - `BoardShell.tsx:1018` (cold-open `initialData` memo) ✓
  - `BoardScenePreview.tsx:126` (version-history preview) — this fix ✓
- **Mixed-scene z-order** — `sanitizeFractionalIndices` now branches on scene shape (`fractionalIndex.ts:98-137`): homogeneous bot scenes stable-sort on raw keys; mixed scenes preserve array order and let `syncInvalidIndices` regenerate stripped keys in place.
- No missing / extra / diverging behavior; the already-correct paths (`repair.ts`, validator, cold-open memo) are byte-unchanged from the previously-reviewed commit.

## 2. Code quality

**Quality: Approved**

### Verified correct (independently, against the real library)

- **Validator fidelity.** Fuzzed `isExcalidrawFractionalIndex` against `fractional-indexing@3.2.0` over 2,000,000 random keys plus an 8,000-key real-key sweep (`generateNKeysBetween` + nested midpoints): **0 false-accepts** (no key the validator calls valid that the library rejects — the canvas-blank vector is closed) and **0 false-rejects** (no real Excalidraw key wrongly stripped — human boards are never corrupted). The one divergence (stricter base-62 fractional check) is a documented, safe over-rejection.
- **Mixed / homogeneous / adversarial z-order.** Ran the sanitizer output through a verbatim port of Excalidraw 0.18.1's `syncInvalidIndices`. All cases render with valid indices and correct z-order, none throw: homogeneous arbitrary order, `[bg(r0), human(a5)]`, interleaved `[h0(a0), bot1, bot0, h1(a1)]`, duplicate invalid keys, and the adversarial **non-monotonic mixed** case `[human_top(a5), bg(r0), human_bot(a0)]` — the last confirms `getInvalidIndicesGroups` picks monotonic bounds and regenerates rather than throwing, so preserving array order for mixed scenes is safe.
- **Render-only invariant holds.** Traced `applyRemote → snapshot → onChange`: `lastKnown` after an apply is seeded from the *reconciled* elements (regenerated keys), Guards 3/4 suppress the self-triggered onChange, and the CAS diff therefore sees no index delta — the FE-regenerated key is not written back to the Y.Doc; the backend stays the authoritative index writer. The preview path has no binding/onChange wiring, so it cannot persist anything by construction.

### Non-blocking (P2)

- **`sanitizeFractionalIndices` has no null/non-object element guard** (`fractionalIndex.ts:103-104`, `:128-130`). A `null` entry in the input array throws `Cannot read properties of null (reading 'index')`. Reachable in theory only via corrupted localStorage (`loadBoardScene` validates `Array.isArray` but not array contents). **Not a regression and not blocking:** the same corrupted input crashed identically before this PR — the real `restoreElements` reduces over each element and reads `element.type` (bundled `restoreElements`, chunk `:20673`), which throws on `null` too, and the sanitizer now simply runs one call earlier on the same path. Worth a cheap defensive `el && typeof el === 'object'` skip (and/or filtering non-objects in `loadBoardScene`) as follow-up hardening, but it does not gate this fix.
- **`[...raw]` / `[...safe]` fallback branches** (`BoardShell.tsx:1018`, `BoardScenePreview.tsx:128`) only apply the sanitizer on the `restore`-present branch. In practice `restoreElements` and `Excalidraw` are captured in the same import callback and the memo is gated on `Excalidraw`, so the unsanitized branch is not reachable with a well-formed 0.18.1 build; sanitizing `raw`/`safe` once before the ternary would be marginally more defensive. Cosmetic.
- Carried over from last round, still non-blocking: raw-string monotonicity breaks at seq rollover (~2.8T elements, theoretical); the "absent index sorts first" comment is inverted vs Excalidraw's actual last-ordering (harmless — array order is what restore honors).

### Tests

New cases pin both regressions precisely: mixed `[bg r0, human a5]` stays `[bg, human]` (valid key not sorted first); interleaved `[h0, bot1, bot0, h1]` preserves authored order with valid keys intact and bot keys stripped in place; "no unparseable key remains in a mixed scene"; and a preview-path test whose mocked `restoreElements` faithfully throws `invalid order key` on an unparseable key, proving the preview mounts instead of blanking. Homogeneous pure-bot path kept byte-for-byte.

## 3. Overall verdict

**APPROVE**

Both prior blocking items (mixed-scene z-order, version-preview gap) are correctly and minimally fixed, verified against the real library and the real reconcile algorithm. The validator is provably faithful in both directions, the render-only invariant holds, and all restore paths are now guarded. The only open items are P2 defensive-hardening suggestions, none of which are introduced or worsened by this PR.

- Add a `el && typeof el === 'object'` skip in `sanitizeFractionalIndices` (or filter non-objects in `loadBoardScene`) so a corrupted-localStorage scene degrades gracefully instead of throwing — closes a pre-existing sharp edge this code now sits on.
- Sanitize `raw`/`safe` once before the ternary in both memos so the no-restore fallback is also covered.

## 5. Extra findings

- The backend is still the durable fix (emit real Excalidraw fractional-index keys); this FE render-defence correctly does not attempt to write corrected keys back. The follow-up is already noted in the PR body.

OctoBoooot (approved):
## Review: render bot-written whiteboards ... (#626) — delta re-review (validator consolidation)

The new commit refactors `isExcalidrawFractionalIndex` to delegate to the shared `@octo/whiteboard-schema` `isValidIndex` instead of keeping a second local port. Verified the shared validator is byte-for-byte equivalent to the local one it replaces, and the two fixed behaviors (preview sanitize, mixed-scene ordering) are preserved. Approval carries forward, strengthened.

Risk tier: normal. Base `main`; `mergeStateStatus=BLOCKED`.

### What changed since `e16ed9ae` (isolated from rebase noise)
- Only `collab/fractionalIndex.ts` (+11/-34): the local `integerLengthFromHead` + `BASE_62_DIGITS` + inline grammar checks are removed; `isExcalidrawFractionalIndex(key)` now `return isValidIndex(key)` (shared, via `schema.ts` re-export of `@octo/whiteboard-schema`).
- `BoardScenePreview.tsx` and `repair.ts` are blob-identical to my prior approved head; the `BoardShell.tsx` diff is rebase (merged sibling), not this commit.

### Equivalence verified (the refactor is behavior-preserving)
Shared `isValidIndex` → `isValidOrderKey` (`whiteboard-schema/normalize.ts`) has the same rules as the local port it replaces:
- whole-string base62 (`BASE62_RE`) ≡ the local per-char base62 loop (extra-strict guarantee preserved);
- SMALLEST_INTEGER sentinel reject ≡;
- `orderKeyIntegerLength` head-map a→2..z→27 / Z→2..A→27 ≡ local `integerLengthFromHead` (local used -1 for illegal head, shared uses 0 — both fail the same `intLen>len`/`===0` guard → identical outcome);
- `intLen > length` reject ≡; fraction-not-ending-`0` ≡.
So `isExcalidrawFractionalIndex` returns identical results for every input. The mixed-scene homogeneous/mixed branching in `sanitizeFractionalIndices` is unchanged (only its `isExcalidrawFractionalIndex` internals delegate); the preview-path sanitize is byte-identical. Both prior blockers stay fixed.

### Praise
- This closes the one latent hazard I'd implicitly accepted across #622/#626: two separate hand-rolled ports of `validateOrderKey` (schema layer + render defence) that could drift. Consolidating to a single shared definition — FE render defence, FE binding normalize, BE repair all agreeing byte-for-byte — is the right DRY fix, net -23 lines with no behavior change.

Jerry-Xin (approved):
Re-affirm APPROVE at new head `d5c913c2` (prior APPROVE at `e16ed9ae` was dismissed by this commit). Both previously-fixed blockers still hold after the `fractionalIndex.ts` rework, and the new BoardShell restore-defence + test introduce no regression.

## Delta since `e16ed9ae` (board files only; `editor/styles.css` is #628 rebase drift, ignored)

**`fractionalIndex.ts` (−45/+ small):** the local copy of the jitterbug validator (`integerLengthFromHead` + base-62 loop + smallest-integer sentinel + fractional-tail rule) was removed; `isExcalidrawFractionalIndex` now delegates to the shared `isValidIndex` from `@octo/whiteboard-schema`. I byte-compared old-local vs new-shared semantics and they are equivalent: same head→integer-length map (`a..z`→2..27, `A..Z`→2..27, else illegal), same `intLen > length` rejection, same base-62 alphabet, same `SMALLEST_INTEGER` sentinel (`A` + 26×`0`), same "fractional part must not end in `0`" rule. The shared validator was itself hardened in this delta (`normalize.ts`): the old loose `/^[A-Za-z0-9]+$/` charset-only check was replaced with the full `validateOrderKey` port, so `r00000000`/`r00000003` now correctly classify as INVALID across FE render defence, FE binding normalize, and BE repair — one legality definition, no drift.

**`sanitizeFractionalIndices` branch structure preserved byte-for-byte:** fast path (`!anyInvalid` → return input untouched, human boards unaffected); MIXED (`anyValidKey` → `elements.map(strip)`, **no sort**); HOMOGENEOUS (stable-sort raw key strings, ties by original index). The mixed-scheme fix is intact.

**`BoardShell.tsx` (+27):** the render-time `initialElements` restore (which runs outside the collab binding's guarded `applyRemote` and outside `BoardErrorBoundary`) now (a) passes `{ normalizeIndices: true }` so Excalidraw re-indexes recoverable keys, and (b) is wrapped in `try { … } catch { return [] }`, degrading an unrecoverable jitterbug throw to an empty initial scene so the canvas still mounts and the later observe→applyRemote path repaints. Per-board graceful degrade, not a per-element or whole-shell crash. `sanitizeFractionalIndices(raw)` still runs before this restore — no new unsanitized path.

## Re-verification of both blockers at new head

**1. Preview sanitize — still present at every restore site.** Enumerated all `restore(`/`restoreElements(` in `packages/docs/src/board`:
- `BoardScenePreview.tsx:126-127` — unchanged this delta; `sanitizeFractionalIndices(scene.elements)` before `restore(safe, null)`. ✅
- `BoardShell.tsx` `initialElements` memo — `sanitizeFractionalIndices(raw)` + try/catch. ✅
- `collab/binding.ts:602` `adapter.restore(repaired)` — `repaired` flows from `repair.ts:53` which returns `sanitizeFractionalIndices(out)`. ✅

**2. Mixed-scheme z-order — NOT reintroduced.** Traced the concrete case `[bg=r0, mid=r1, top=a5]`: `r0`/`r1` classify invalid, `a5` valid → `anyValidKey` true → MIXED branch → `elements.map(strip)` preserves array order `[bg, mid, top]`, strips only the bot keys. No cross-space string sort, so the `'a5' < 'r00000000'` inversion cannot drag `top` under the bot layers; `syncInvalidIndices` re-slots the stripped keys in place. The existing `fractionalIndex.test.ts` z-order suite (unchanged this delta) still pins this: the `[bg=r00000000, human=a5]` case asserts output order `['bg','human']` with `human.index === 'a5'`, plus a 4-element mixed case asserting `['h0','bot1','bot0','h1']`.

## Tests enforce (not pin a regression)
- New `BoardShellRestoreDefense.test.tsx` uses an Excalidraw stand-in whose `restoreElements` mirrors the real jitterbug throw (`isJitterbugValid`) and feeds a snapshot with an invalid `r00000003` key; asserts the canvas still mounts — i.e. it exercises the try/catch degrade, would fail if the guard were removed.
- Existing `fractionalIndex.test.ts` still covers `sanitizeFractionalIndices` (now via the shared validator) and enforces both homogeneous stable-sort and mixed-scene order fidelity.

## Gate
CI green (Build SUCCESS at `d5c913c2`); MERGEABLE; blocked only on review approval.

## Highlights
Consolidating to a single shared `isValidIndex` removes a duplicate validator that could have drifted from the BE contract, and strengthening it from the loose charset check closes the underlying `r00000000`-passes-as-valid gap at the schema layer while the mixed-scene z-order guard stays byte-identical. Restore-defence depth (normalizeIndices + try/catch) is a sound belt-and-suspenders for the one render path outside the error boundary.

Jerry-Xin (approved):
**Re-review after rebase — HEAD `d5c913c2`** ✅ APPROVE

Re-reviewed the rebase/conflict-resolution head byte-for-byte (not the commit prose). The three moving parts introduced by the rebase — the grafted #622 cold-open `try/catch`, the merge into a single shared `isValidIndex`, and `normalizeIndices` — are all correct and no previously-verified fix semantics were lost.

**1. Shared validator merge is faithful (no drift)** — `collab/fractionalIndex.ts:45` now delegates to the single `isValidIndex` (`@octo/whiteboard-schema/src/normalize.ts:89` → `isValidOrderKey`, re-exported via `packages/docs/src/board/collab/schema.ts`). The removed local `integerLengthFromHead`/`BASE_62_DIGITS` copy is fully gone — no residual copies anywhere under `board/`. I ran a differential fuzz of the removed local validator against the shared one over ~240k cases (exhaustive short keys + edge cases + random): **0 mismatches**. Both agree on: reject the `r00000000` scheme (head declares 19 int digits, only 9 follow), reject the `A`+26-zeros SMALLEST_INTEGER sentinel, reject a trailing-`0` fractional part, accept real keys (`a0`, `a0V`, `Zz`, `a0V8`). No mis-accept of illegal keys, no mis-kill of legal ones.

**2. #622 cold-open try/catch graft coordinates with the sanitize (no fight)** — `BoardShell.tsx:1030-1037`: `try { restore(sanitizeFractionalIndices(raw), null, { normalizeIndices: true }) } catch { return [] }`. The sanitize strips unparseable keys *before* restore (so `syncInvalidIndices` never throws on the backend scheme), `normalizeIndices` re-indexes recoverable keys, and the `try/catch` is the deeper backstop for anything neither can repair, degrading to an empty scene so the canvas still mounts and `observe→applyRemote` repaints. They stack; they do not double-process or swallow each other's work.

**3. Blocking#1 (version-history preview) preserved** — `BoardScenePreview.tsx:126` sanitizes `scene.elements` before `restore(safe, null)`, closing the third untrusted-restore path. Test coverage present and green.

**4. Blocking#2 (mixed-scene z-order) preserved** — `fractionalIndex.ts:98-118`: homogeneous bot scenes stable-sort on the raw key strings; **mixed** scenes (valid + backend keys) are never sorted (the two key spaces are string-incomparable), original array order is kept and only stripped keys regenerate in place. Both branches covered by tests.

**5. normalizeIndices** — present only on the `initialData` restore path (`BoardShell.tsx:1032`), which is the intended #622 defence; correct and necessary there.

**6. No regression on human boards** — `sanitizeFractionalIndices` returns the *same array reference* when every key is valid/absent (zero-copy fast path, `fractionalIndex.ts:79-88`), so real-time collab/reconcile behaviour is untouched. Verified by the fast-path tests.

**Local verification (HEAD d5c913c2):**
- `@octo/whiteboard-schema` tests: **12 passed / 12**
- `packages/docs` board tests (`vitest run src/board`): **192 passed / 192**
- Changed-file typecheck: **clean** — the only 2 `tsc` errors in the docs project are in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts`, both pre-existing and untouched by this board-only PR
- Standalone build (`vite build`): **succeeds**

Rebase did not break the fix semantics, the #622 graft, or the shared-validator merge, and introduces no new regression. Approving.

yujiawei (approved):
# Code Review — PR #626 (octo-web)

Reviewed at head `d5c913c26f7ee4928bd8fcce7b2ef6b63507928d` against `main` (merge-base `24baea5d`). I verified the core claims against the actual bundled libraries (`fractional-indexing@3.2.0`, `@excalidraw/excalidraw@0.18.1`) and ran the board test suite.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the PR describes — a frontend, render-only defence that strips `index` keys Excalidraw's grammar cannot parse before restore/reconcile — and nothing more.

- **Nothing missing:** all three untrusted-restore paths are covered — the live `applyRemote` path via `repairForRender` (pass 3), the `BoardShell` cold-open `initialData` memo (seeds raw Y.Doc elements, bypasses `repairForRender`), and the `BoardScenePreview` version-history path (also bypasses `repairForRender`).
- **Nothing extra:** the sanitized result is never written back to the Y.Doc, so the backend remains the authoritative index writer. No new flags, endpoints, or schema changes. The validator is not duplicated — it delegates to the shared `isValidIndex` in `@octo/whiteboard-schema`, so FE render-defence, FE binding normalize, and BE repair share one legality definition.
- **No divergence:** the backend follow-up (emit real fractional-index keys) is correctly scoped out as a separate backend change; this PR is the render defence for boards already written with the old scheme.

## 2. Code quality

**Quality: Approved**

Verified against the real library internals rather than trusting the description:

- **`isValidIndex` fidelity (`packages/whiteboard-schema/src/normalize.ts:59-91`).** It faithfully mirrors `fractional-indexing`'s `validateOrderKey`: the `SMALLEST_INTEGER` sentinel rejection, the head-char integer-length rule (`a`→2 … `z`→27, `A`→27 … `Z`→2), and the trailing-`0` fractional rule all match. It is marginally stricter on charset (rejects the empty/whitespace edge), which is safe — a false-negative only means Excalidraw regenerates a key it would have accepted, never a throw.
- **Backend scheme is always stripped.** `r00000000` has head `r` → declared integer length 19, but the key is only 9 chars, so `intLen > key.length` rejects it every time. Confirmed for the whole `r${seq.toString(36).padStart(8,'0')}` family including the base36 `9`→`a` rollover; the scheme is also string-monotonic, which is what the homogeneous stable-sort relies on.
- **Homogeneous vs mixed split is correct against the actual restore/reconcile behaviour.** In `@excalidraw/excalidraw@0.18.1`, `restoreElements` runs only `syncInvalidIndices` (which regenerates stripped keys honouring **array order**), while `reconcileElements` runs `orderByFractionalIndex` first and *then* `syncInvalidIndices`. So preserving array order for a mixed scene is right, and in the live path `restoreElements` re-keys every element before `reconcileElements` sorts, so no index-less element ever reaches `orderByFractionalIndex`. I reproduced the mixed case end-to-end: a stripped bottom element + a valid `a5` + a stripped top element regenerate to `a4 < a5 < a6`, authored z-order intact.
- **Fast path is a genuine no-op** for all-valid (human) scenes — same array reference returned, no reorder, no copy — so real-time collaboration is unchanged. The strip preserves every other field, unknown ones included.
- **Depth defences are sound:** the cold-open memo keeps its `try/catch` → `[]` fallback plus `normalizeIndices: true`, and the preview relies on `BoardVersionPanel`'s error boundary; sanitizing upstream turns the common case from "throw and blank" into "render", with those backstops still present for the residual unrecoverable case.

**Tests:** the board suite is green (192 passing here, including the new `fractionalIndex.test.ts`, the `repair.test.ts` pass-3 cases, and the `BoardScenePreview` bot-board cases). The `BoardScenePreview` mock now throws on unparseable keys exactly as the real library does, so those tests genuinely exercise the sanitize-before-restore contract rather than a stubbed pass-through. Typecheck shows two errors, but both are pre-existing on `main` (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) and untouched by this PR.

### Non-blocking notes (P2)

- **Coupling to Excalidraw internals.** The mixed-vs-homogeneous correctness rests on the current 0.18.1 behaviour that `restoreElements` does *not* sort by index while `reconcileElements` does. That is well-documented in the module header, and the tests would catch a regression on upgrade, but it is worth a note in the Excalidraw-upgrade checklist so a future bump re-validates this assumption.
- **Verification gap acknowledged in the PR.** The live repro board was validated headlessly, not in a real browser. The headless verification against the exact backend key format is convincing for the throw/no-throw behaviour, but an in-browser confirmation on the referenced doc before this leaves draft would close the loop, as the PR itself recommends.

## 3. Overall verdict

**APPROVED** — Spec ✅ and Quality Approved, no P0/P1. The fix is correctly scoped, render-only, and its central claims hold up against the actual bundled `fractional-indexing` and Excalidraw code. The two notes above are advisory and do not block merge.

PR #619, opened 2026-07-10
feat: add personal runtimes and skills module

superalsrk:
## Summary
- Add the new `@octo/personal` package and register the top-level Personal / 我的 module.
- Move Runtimes and Skills navigation out of Loop into Personal while reusing the existing Loop pages.
- Bootstrap Personal pages by listing workspaces and setting the active workspace context before loading workspace-scoped APIs.
- Redesign the Runtimes list as a grouped machine table and add the Add computer setup modal.
- Ignore local `.omx` runtime state in `.gitignore`.

## Verification
- `corepack pnpm i18n:check`
- `git diff --check origin/feat/octo-loop...HEAD`
- `corepack pnpm --filter @octo/web build`

## Notes
- The web build completes with existing warnings around lottie/pdfjs direct eval, large chunks, and Semi dynamic import handling.

PR #613, opened 2026-07-10
feat(loop): sort control and creator filter on the issue list

MrAladdin:
## What

**Phase B (browse parity), slice 1.** Extends the issue browse bar with server-side **sort** and a **creator filter**, using existing `GET /issues` params (no new endpoint).

- `ListParams` gains `sort_by`/`sort_direction`/`creator_id`; `issueApi` maps `sort_by→sort`, `sort_direction→direction` (backend whitelist: position|priority|title|created_at|start_date|due_date; direction honored only when sort ≠ position).
- Sort control (field Select + asc/desc toggle, disabled for position) + creator Select (member candidates). Any filter change resets to page 1.
- **Sort applies to LIST view only**; board omits it and stays on position. *(Adversarial-review finding: board fetches a 100-cap and splits client-side by status, so a global non-position sort could push a whole status past the cap and render its column falsely empty. Board keeps position; sort control hidden in board.)*
- `ISSUE_SORT_FIELDS` is a single `as const` source of truth; `IssueSortField` derived from it.

## Verification (Alice, MV2 E2E)

| Check | Result |
|---|---|
| sort created_at desc / asc | list reorders; toggle reverses (MVE-4/2/1 ⇄ MVE-1/2/4) |
| direction toggle on position | disabled |
| creator = Alice | narrows to Alice-authored (excludes wangdeng-created MVE-4) |
| sort control visibility | shown in list, hidden in board |

## Blast radius (Rule 8)

Contained to dmloop issue browse — `ListParams`/`IssueSortField` have no consumers outside `issueApi`/`IssuePage`; no package imports `@octo/loop`. Retrospective check of the merged Phase A PRs (#605/606/608/609/610) also found **no out-of-module breakage**.

## Review

`/simplify` (single-source sort fields) + adversarial correctness + cross-module blast-radius pass. Follow-up: creator dropdown lists members only (agent/squad creators omitted) — completeness nit, deferred.

## Notes

- Off the current `feat/octo-loop` (all Phase A merged) — standalone, not stacked.
- No linked issue yet — `check-sprint` needs a maintainer. Frontend-only; no doc update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #627, opened 2026-07-10
feat(loop): filter issues by project on the issue list

MrAladdin:
## What

Add a **project filter** to the Loop issue list/board header, mirroring the
existing status / priority / assignee / creator filters.

## Why / how

The backend `GET /issues` already accepts a `project_id` query param
(`octo-multica` `ListIssues`), so this is a pure-frontend addition:

- `ListParams` gains an optional `project_id`; `listIssues` forwards it.
- `IssuePage` renders a project `Select`, shown only when the workspace has
  projects.
- Project options come from a new `directory` helper `listProjectOptions()`
  (symmetric to `listAssigneeCandidates()`), reusing the already-cached
  directory instead of a second `/projects` fetch.

### /simplify pass

Ran `/simplify` (reuse / simplification / efficiency / altitude). Two things
landed:
- **reuse/altitude**: extracted `listProjectOptions()` into `directory.ts`
  instead of reaching into the `projectName` Map from the page.
- **efficiency**: tightened `ensureDirectory`'s in-flight dedupe. The old guard
  also required a matching cached slug, so on a cold start (cache still null)
  two concurrent callers (`useAssigneeCandidates` + the new project loader) each
  kicked off a full directory build. Now it guards on the in-flight promise
  alone (still bypassed by `force`), so concurrent callers share one build.

- `listIssues` has a single caller (`IssuePage`); the new optional `ListParams`
  field is a no-op for the squad/agent/skill/runtime `list*` helpers.
- The `ensureDirectory` dedupe change is shared infra. Verified: workspace
  switch calls `invalidateDirectory()` synchronously before the keyed remount,
  so no stale in-flight build leaks across workspaces; all consumers
  (`enrich`/`actorName`/`actorAvatar`/squad enrich) receive the same `Directory`
  as before, only fewer requests. `@octo/loop` is a leaf package.

## Verification

- **Adversarial review**: a Claude reviewer went through the dedupe change
  point-by-point (workspace-switch staleness, `force` semantics, `_loading`
  race, consumer blast-radius) and found no confirmed defect. Codex was also
  run per the two-model cross-check, but its runs did not complete this session
  (stream disconnects) — flagging that honestly rather than claiming a codex
  pass.
- **Real browser (dev env)**:
  - Workspace without projects (Loop-dev) → project Select hidden.
  - Workspace with a project (MV2 E2E, created "筛选测试项目") → Select shows,
    populated via `listProjectOptions`; selecting it narrows the board from 3
    issues to 0 (none assigned), proving `project_id` reaches the backend.
  - Switching Loop-dev ↔ MV2 E2E loads the correct per-workspace directory
    (issues, assignee names, project list) with no cross-workspace bleed.

### DIFF
--- a/packages/dmloop/src/api/directory.ts
+++ b/packages/dmloop/src/api/directory.ts
@@ -64,7 +64,9 @@ async function build(): Promise<Directory> {
 
 export async function ensureDirectory(force = false): Promise<Directory> {
   if (!force && _cache && _cache.slug === currentWorkspaceSlug()) return _cache;
-  if (_loading && _cache?.slug === currentWorkspaceSlug()) return _loading;
+  // 已有 in-flight build 时并发调用者直接搭车(切 workspace 会 invalidateDirectory 清空 _loading),
+  // 避免冷启动(_cache 尚为 null)多个消费者各拉一遍 directory。
+  if (!force && _loading) return _loading;
   _loading = build().then((d) => {
     _cache = d;
     _loading = null;
@@ -106,4 +108,11 @@ export function actorAvatar(
 export async function listAssigneeCandidates(): Promise<AssigneeCandidate[]> {
   const dir = await ensureDirectory();
   return dir.candidates;
+}
+
+// 项目筛选下拉的选项，复用已缓存的 directory（不额外请求 /projects），与
+// listAssigneeCandidates 对称——调用方无需伸手进 Directory 内部结构。
+export async function listProjectOptions(): Promise<Array<{ id: string; title: string }>> {
+  const dir = await ensureDirectory();
+  return [...dir.projectName].map(([id, title]) => ({ id, title }));
 }
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -34,6 +34,7 @@ export async function listIssues(
     priority: params?.priority,
     assignee_id: params?.assignee_id,
     creator_id: params?.creator_id,
+    project_id: params?.project_id,
     sort: params?.sort_by,
     direction: params?.sort_direction,
     limit: params?.limit,
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -60,6 +60,7 @@ export interface ListParams {
   priority?: IssuePriority;
   assignee_id?: string;
   creator_id?: string;
+  project_id?: string;
   // 后端白名单 sort：position(默认)|priority|title|created_at|start_date|due_date。
   // direction 仅在 sort_by != position 时被后端采纳(asc|desc)。
   sort_by?: IssueSortField;
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -46,7 +46,8 @@
     "status": "Status",
     "priority": "Priority",
     "assignee": "Assignee",
-    "creator": "Creator"
+    "creator": "Creator",
+    "project": "Project"
   },
   "sort": {
     "direction": "Asc/desc",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -46,7 +46,8 @@
     "status": "状态",
     "priority": "优先级",
     "assignee": "负责人",
-    "creator": "创建者"
+    "creator": "创建者",
+    "project": "项目"
   },
   "sort": {
     "direction": "升/降序",
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -13,6 +13,7 @@ import { useI18n, WKApp } from "@octo/base";
 import type { Issue, IssueStatus, IssuePriority, IssueSortField } from "../api/types";
 import { ISSUE_SORT_FIELDS } from "../api/types";
 import { listIssues } from "../api/issueApi";
+import { listProjectOptions } from "../api/directory";
 import { useAssigneeCandidates } from "../ui/useAssigneeCandidates";
 import { ISSUE_STATUS_ORDER, PRIORITY_ORDER } from "../ui/meta";
 import IssueBoard from "../panel/IssueBoard";
@@ -30,6 +31,7 @@ interface Filters {
   priority?: IssuePriority;
   assignee?: string;
   creator?: string;
+  project?: string;
   sortBy: IssueSortField;
   sortDir: "asc" | "desc";
 }
@@ -47,7 +49,13 @@ export default function IssuePage() {
   const [page, setPage] = useState(0); // 0-based，仅列表视图分页
   const [createOpen, setCreateOpen] = useState(false);
   const cands = useAssigneeCandidates();
+  // 项目下拉复用 directory 已缓存的 /projects(避免重复请求);随 workspace 切换整页重挂而刷新。
+  const [projects, setProjects] = useState<Array<{ id: string; title: string }>>([]);
   const seq = useRef(0); // 请求序号：只应用最新一次的响应，防并发乱序覆盖
+
+  useEffect(() => {
+    listProjectOptions().then(setProjects).catch(() => {});
+  }, []);
 
   const reload = useCallback(() => {
     const my = ++seq.current;
@@ -59,6 +67,7 @@ export default function IssuePage() {
       priority: f.priority,
       assignee_id: f.assignee,
       creator_id: f.creator,
+      project_id: f.project,
       // 排序仅用于列表视图;看板按 status 分列 + 100 上限,叠加全局排序会把某状态整列截没,故看板固定后端默认(position)。
       sort_by: paged ? f.sortBy : undefined,
       sort_direction: paged ? f.sortDir : undefined,
@@ -155,6 +164,21 @@ export default function IssuePage() {
             <Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>
           ))}
         </Select>
+        {projects.length > 0 && (
+          <Select
+            placeholder={t("loop.filter.project")}
+            value={f.project}
+            onChange={(v) => update({ project: v as string | undefined })}
+            showClear
+            filter
+            size="small"
+            style={{ width: 140 }}
+          >
+            {projects.map((p) => (
+              <Select.Option key={p.id} value={p.id}>{p.title}</Select.Option>
+            ))}
+          </Select>
+        )}
         {view === "list" && (
         <div style={{ display: "flex", alignItems: "center", gap: 2 }}>
           <Select


OctoBoooot (approved):
## Review: filter issues by project on the issue list (#627)

**Verdict:** Approve
Clean, minimal project filter that follows the established status/priority/assignee/creator pattern exactly. Backend `GET /issues` already takes `project_id`, so this is pure-frontend plumbing. Verified against the code.

Risk tier: normal. Base `feat/octo-loop`; `mergeStateStatus=UNSTABLE` + no linked issue (`check-sprint`) — approved on code, a maintainer must clear the sprint gate before merge.

### What I verified
- **Param plumbing:** `project_id` added to `ListParams`, forwarded in `listIssues` (undefined drops out like the sibling params), and passed in `reload()` for **both** list and board views. Applying a filter to both views is correct — unlike sort (which #613 rightly gated to list-only because a global reorder can push a status column past the board's 100-cap), a filter only *narrows* results, so it's safe under the cap.
- **Page reset:** `update({ project })` goes through the same `update()` that does `setPage(0)`, so changing the project resets to page 1, consistent with the other filters.
- **Conditional render:** the Select shows only when `projects.length > 0` — matches "hidden when the workspace has no projects."
- **No redundant fetch:** project options reuse `ensureDirectory()`'s already-cached `projectName` map (fetched for enrichment) instead of a second `/projects` request — the redundant-fetch point from the adversarial review is addressed. `[...d.projectName].map(([id, title]) => …)` iterates the Map entries correctly.
- **Concurrency:** the existing `seq` stale-response guard covers the new param (no new async path).

### Minor
- `IssuePage.tsx` — the project-options effect has deps `[]`, so it loads once on mount. If the page does not remount on a workspace switch, the dropdown could show stale project options (e.g. a just-created project missing) until reload. Benign (stale filter *options*, not wrong data or a crash — self-corrects on reload), and smaller than the analogous #619 workspace-context issue, but worth confirming IssuePage remounts on workspace change (the comment claims it does). Non-blocking.

### Praise
- Reusing the directory cache for the options rather than firing a second `/projects` call is the right call — it keeps the filter free of extra network cost and in sync with the same source that resolves `project_name` for enrichment, so the dropdown and the rendered rows can't drift.

yujiawei (approved):
# Code Review — PR #627 (octo-web)

## Summary

Adds a **project filter** to the Loop issue list/board header, alongside the existing status / priority / assignee / creator filters. The change is purely front-end and additive: `ListParams` gains an optional `project_id`, `listIssues` forwards it to `GET /issues`, and `IssuePage` renders a project `Select` (populated from the already-cached directory) that is hidden when the workspace has no projects.

**Verdict: APPROVED.** No correctness, security, or blocking functional issues found. A couple of optional nits below.

## Spec / scope

- `packages/dmloop/src/api/types.ts` — one optional field `project_id?: string` on `ListParams`.
- `packages/dmloop/src/api/issueApi.ts` — forwards `project_id` in the `GET /issues` query.
- `packages/dmloop/src/pages/IssuePage.tsx` — one `Filters.project` field, a mount-time `ensureDirectory()` load, and one conditional `Select`.
- `en-US.json` / `zh-CN.json` — one `loop.filter.project` label per locale.

No exported signature changed in a breaking way; no out-of-scope additions.

## Correctness — verified cross-file

- **Directory reuse is real.** `ensureDirectory()` and `d.projectName` (a `Map<string,string>`) exist in `directory.ts`; the `/projects` fetch is already performed there for `project_name` enrichment, so no second request is issued. Confirmed.
- **Unselected project sends no param.** `httpGet`'s `clean()` (`http.ts`) drops `undefined` / `null` / `""` values, so `project_id: f.project` with `f.project === undefined` results in no query param — identical to how the assignee/creator filters behave. No accidental empty-string filter.
- **No new race.** `update()` resets `page` to 0 and the `seq` request-guard already discards stale responses; the project filter flows through the same `reload` path as the other filters, so the seq/page-clamp logic covers it unchanged.
- **Types line up.** `Issue.project_id` / `project_name` and `ListParams.project_id` are consistent in `types.ts`; `enrich()` already reads `dir.projectName`.
- **Mount effect matches precedent.** The new `useEffect(... ensureDirectory ...)` mirrors `useAssigneeCandidates` exactly (mount-only, directory-cached, `.catch` swallow). On workspace switch the page remounts, refreshing both — so the project filter is at parity with the existing filters, not a regression.

## Nits (non-blocking)

1. **Backend contract is assumed, not verifiable in this repo.** Correct filtering depends on `GET /issues` honoring `project_id` server-side (octo-multica `ListIssues`). This isn't checkable from octo-web, but the author's browser verification (selecting the project narrowed the board 3 → 0) is good empirical evidence the param reaches and filters at the backend. If the backend ever silently ignores an unknown `project_id`, the UI would show "no filtering" rather than an error — acceptable, just noting the coupling.
2. **Stale-selection edge (minor UX).** `f.project` is not cleared if the selected project is later deleted while the page is open (the option would disappear but the id would linger until the next filter change / remount). Same latent behavior already exists for assignee/creator, so this is consistent, not a new defect.

## Coverage note

- Reviewed: the full diff plus its cross-file dependencies (`directory.ts`, `http.ts`, `types.ts`, `useAssigneeCandidates.ts`) at head SHA.
- Not exercised: runtime behavior against a live backend (relied on the author's dev-env verification) and the octo-multica `GET /issues` handler, which lives in a separate repo and is outside this diff.

Clean, well-scoped change. Approving.

Jerry-Xin (approved):
Clean, minimal project filter that faithfully follows the established list-filter pattern (same shape as the sort/creator work in #613). Approving.

**Verified**
- **Bound param, no injection** — `project_id` flows `Filters.project` → `ListParams.project_id` (`api/types.ts`) → `listIssues` → `httpGet("/issues", { ... project_id })` → `client.get(path, { params: clean(params) })` (`api/http.ts:116`). `clean()` drops undefined/null/"" and passes values to axios `params`, which URL-encodes them. No string interpolation into the URL/path.
- **Server-side filter, no truncation trap** — `project_id` is sent to the backend `/issues` query in both views, so filtering happens server-side (not client-side over an already-truncated page). The board still fixes `sort_by`/`direction` to the backend default under `limit:100` (`IssuePage.tsx:72-77`), so the project filter does not reintroduce the sort-truncation problem; counts/`total` remain backend-authoritative.
- **Page reset on filter change** — the project `onChange` goes through `update()`, which resets `setPage(0)` (`IssuePage.tsx`), avoiding a stale out-of-range offset.
- **Stale-response guard reused** — the new param is part of `f`; `reload` depends on `f` and is protected by the existing `seq.current` sequence guard, so concurrent responses can't clobber a newer request.
- **Project source is workspace-scoped** — the dropdown is populated from `ensureDirectory().projectName`, built from `GET /projects` (workspace-scoped, server-authorized), reusing the already-cached directory. Frontend filter only; backend enforces authorization.
- **XSS-safe** — project titles render as escaped JSX text (`{p.title}` in Select.Option, `{issue.project_name}` in the detail/list panels). No `dangerouslySetInnerHTML` anywhere in `packages/dmloop/src`.
- **i18n complete** — `loop.filter.project` added to both `en-US.json` and `zh-CN.json`.

**Non-blocking notes**
- The project Select renders only when `projects.length > 0`; on a workspace with zero projects the control is hidden — intended, and consistent with how other pickers degrade.

Nice, tight change. LGTM.

OctoBoooot (approved):
## Review: filter issues by project on the issue list (#627) — delta re-review

The new commit refactors the project-options lookup into a `listProjectOptions()` helper and tightens `ensureDirectory`'s in-flight dedup. Behavior-preserving for the filter; the concurrency change is a net improvement and I verified it's safe against workspace switches. Approval carries forward.

### What changed since `9c16b3d3` (isolated from rebase noise)
- **`IssuePage.tsx`** — the inline `ensureDirectory().then(d => [...d.projectName]…)` is replaced by `listProjectOptions().then(setProjects)`. Same result, but callers no longer reach into the `Directory.projectName` internal shape — symmetric with `listAssigneeCandidates`. Behavior-equivalent for the dropdown.
- **`directory.ts`** — adds the `listProjectOptions()` helper (the extracted logic), and changes the in-flight guard from `if (_loading && _cache?.slug === currentWorkspaceSlug())` to `if (!force && _loading)`.

### Why the concurrency change is safe (verified)
- The cache-hit path (`:66`) still slug-guards — a stale-workspace `_cache` is never returned.
- `invalidateDirectory()` (`:78-80`) clears **both** `_cache` and `_loading`, and a workspace switch calls it. So an in-flight `_loading` from the previous workspace can't be ridden after a switch — the next `ensureDirectory` sees `_loading === null` and starts a fresh `build()` that reads `currentWorkspaceSlug()` at call time.
- Net: the old guard checked the *stale cache's* slug on the in-flight path (a slightly confused condition); the new one is cleaner, fixes cold-start thundering-herd (multiple consumers each building when `_cache` is null), and introduces no wrong-workspace directory risk.

### Minor (carried, unchanged — non-blocking)
- The project-options effect still loads once on mount; stale dropdown *options* are possible if `IssuePage` doesn't remount on a workspace switch. Inherent to cache reuse, benign (stale options, not wrong data), self-corrects on reload. Same note as the prior review.

### Praise
- Extracting `listProjectOptions()` keeps the `Directory` internal shape encapsulated (symmetric with `listAssigneeCandidates`), and the in-flight dedup tightening is a real correctness/perf improvement to shared infra that the filter work surfaced — fixed at the right layer rather than worked around in the page.

Jerry-Xin (approved):
**APPROVE** — delta re-review since my prior APPROVE (`9c16b3d3` → `e954c13f`). Two-file refactor; no behavior regression, and it fixes a real cold-start over-fetch.

### Delta scope (exactly 2 files, +12/-5)
- `api/directory.ts` (+11): new exported `listProjectOptions()` accessor + relaxed the in-flight dedup guard in `ensureDirectory`.
- `pages/IssuePage.tsx` (+6/-4): call-site swap to `listProjectOptions()`.

### `ensureDirectory` guard change — verified race-free
Old: `if (_loading && _cache?.slug === currentWorkspaceSlug()) return _loading;`
New: `if (!force && _loading) return _loading;`

- **Cross-workspace stale-data ride: no race.** The only workspace-switch paths (`LoopPage.applyWorkspace` / `switchWorkspace`) call `setWorkspaceContext(slug,id)` immediately followed by `invalidateDirectory()` **synchronously in the same JS turn**, and `invalidateDirectory` nulls both `_cache` and `_loading`. JS is single-threaded, so no caller can interleave between the slug change and `_loading = null`. Any caller arriving after a switch sees `_loading === null` → starts a fresh build tagged with the new slug; it can never ride the old-slug in-flight promise. The code comment's safety claim holds.
- **Cold-start dedup fix is real.** With `_cache === null` (cold start) the old guard's `_cache?.slug` was `undefined` ≠ slug, so the ride was skipped and every concurrent consumer reassigned `_loading = build()`, firing N redundant `/members`+`/agents`+`/squads`+`/projects` bursts. The new guard lets consumers 2..N ride the first in-flight build. Correctness preserved: within one JS turn the slug cannot change without `invalidateDirectory` clearing `_loading`, so all riders resolve with the correct same-slug directory.
- **`force=true` still bypasses.** New `!force` short-circuits the ride so a forced call always starts a fresh build. (Note: the old guard lacked `!force`, so a forced caller could have ridden a matching in-flight build — the new code is strictly more correct here. No current caller passes `force`, so no runtime change either way.)

### `listProjectOptions()` — behavior-identical
`const dir = await ensureDirectory(); return [...dir.projectName].map(([id,title])=>({id,title}))` is byte-equivalent to the previously-approved inlined mapping in IssuePage: same Map iteration, same `{id,title}` shape, no extra fetch. Pure accessor, symmetric with `listAssigneeCandidates` — good encapsulation (callers no longer reach into `Directory` internals).

### IssuePage swap — pure refactor
`listProjectOptions().then(setProjects).catch(()=>{})` replaces the inlined `ensureDirectory().then(map)`. Filter logic, page-reset, param plumbing, and the XSS handling from my prior review are untouched by this delta.

### Non-blocking (carried forward, not introduced here)
- 🟡 The project-options `useEffect` deps stay `[]`, so the dropdown isn't refreshed on workspace switch — same pre-existing follow-up noted before; not a regression.

Carrying forward APPROVE.

yujiawei (approved):
# Code Review — PR #627 (octo-web)

## Summary

Adds a **project filter** to the Loop issue list/board header, mirroring the existing status / priority / assignee / creator filters. The change is small and additive:

- `ListParams` gains an optional `project_id`; `listIssues` forwards it to `GET /issues`.
- `IssuePage` renders a project `Select`, shown only when the workspace has projects.
- Project options are reused from the already-cached `directory` (new `listProjectOptions`) instead of a second `/projects` request.
- Supporting change: `ensureDirectory` now shares an in-flight `build()` promise for any concurrent cold-start caller, so the two mount-time consumers (assignee candidates + project options) don't each trigger a full directory fetch.

Reviewed against head SHA `e954c13`.

## Verdict: APPROVED

No correctness, security, or data-loss defects were found in the diff. The items below are non-blocking (nits / pre-existing follow-ups).

---

## Findings

### 1. Cross-workspace directory race — pre-existing, NOT introduced here (informational)

`packages/dmloop/src/api/directory.ts:70-74`

```ts
_loading = build().then((d) => {
  _cache = d;
  _loading = null;
  return d;
});
```

If a workspace switch happens while a `build()` is in flight, the earlier build's `.then` still runs and writes its (now stale) result into `_cache`. If that earlier build resolves *after* the new workspace's build, `_cache` ends up holding the previous workspace's directory data.

However, this `.then` block is **unchanged by this PR**, and the guard this PR removed (`_cache?.slug === currentWorkspaceSlug()`) never protected against the switch-during-build case — it only forced a refetch on cold start. `invalidateDirectory()` (called on every workspace switch) clears `_cache`/`_loading` but does not cancel an already-running `build()`, and never did. So this is a **latent pre-existing race**, not something this diff introduces or worsens.

Optional hardening for a future change: capture the target slug when `build()` starts and skip the `_cache = d` write if `currentWorkspaceSlug()` no longer matches. Not required for this PR.

### 2. `listProjectOptions` effect has no unmount/abort guard (nit)

`packages/dmloop/src/pages/IssuePage.tsx:55-57`

```ts
useEffect(() => {
  listProjectOptions().then(setProjects).catch(() => {});
}, []);
```

`setProjects` can fire after unmount if the workspace is switched mid-fetch. This is harmless on React 18 (no warning) and is consistent with the existing `useAssigneeCandidates` pattern in this package, so it is fine as-is — flagging only for consistency if a cancellation convention is later adopted.

---

## Coverage notes

- **Backend contract**: `project_id` is assumed accepted by `GET /issues` (`octo-multica ListIssues`). It rides the exact same param path as the proven `assignee_id` / `creator_id` filters and the PR reports a real-browser check (board narrowed 3 → 0 for a project with no issues), so the wiring reaches the backend. The server-side handling itself lives in another repo and was not re-verified here.
- **Both views**: `project_id` is passed outside the paged/board conditional, so the filter applies to both list and board views — matches the stated behavior.
- **i18n**: `loop.filter.project` added to both `en-US` and `zh-CN`.
- **Verifier note**: A second independent model leg on this review returned an infrastructure error and did not produce findings; this review reflects one automated leg plus the reviewer's own pass. One flagged item from the leg — a claimed `TypeError` from spreading `dir.projectName` — was checked and is a false positive (`projectName` is a `Map`, so `[...map]` is valid).

PR #610, opened 2026-07-10
feat(loop): live-stream the agent run transcript

MrAladdin:
## What

Phase A of the crown loop (final piece). The run detail modal was a one-shot read; it now streams messages while the run is active and keeps the run's own state fresh.

- `runsApi.listRunMessages` gains `since` → `GET /tasks/:id/messages?since=<seq>`.
- `RunMessage` realigned to backend `TaskMessagePayload`; `msgText` now renders **tool_result output** and **tool_use input** (previously read a nonexistent `text` field → those rows were blank).
- `RunDetailModal`: full fetch on open; for an active run, poll every 2s with `?since=lastSeq` and append (dedup by seq). Also refreshes the run's own state each poll and **stops once the run is terminal** — the header status / completed time / result update live instead of freezing at the click-time snapshot (fixes the review finding), and the poll self-terminates.
- Extracted shared `isActiveRun()`/`ACTIVE_RUN_STATUSES` in `ui/meta`, reused in RunDetailModal + IssueDetailPage.

## Verification (live daemon, agent "E2E Coworker", Alice)

| Path | Result |
|---|---|
| open a completed run | all 19 messages render non-empty (text + tool_use input + tool_result output) |
| re-run + open while active | messages streamed live 0 → 11 → 12 |
| run completes | header flips running → **completed**, `completed_at` fills, result section appears, polling stops (DB/UI-confirmed) |

## Review

- `/simplify`: shared `isActiveRun`, simplified `since` param; kept the seq dedup as insurance against an unverified `?since` inclusive/exclusive boundary.
- **Adversarial correctness pass** — fixed the confirmed frozen-run-status finding (poll now refreshes run state + self-terminates); dismissed a cosmetic tag-color nit.

## Notes

- dmloop has no fleet WebSocket client, so this is `?since` polling rather than the backend `task:message` WS.
- **Stacked on #609 → #608 → #606 → #605 → #602**; merge in order.
- No linked issue yet — `check-sprint` needs a maintainer. dmloop feature UI; no service/command/config change → no doc update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #605, opened 2026-07-10
feat(loop): dispatch confirmation when assigning an issue to an agent

MrAladdin:
## What

Phase A of surfacing multica's crown **assign → run** loop in the octo (dmloop) face. Assigning an issue to an **agent or squad** now confirms with the backend whether a run will start *before* committing — instead of silently dispatching.

- **`previewIssueTrigger()` → `POST /issues/preview-trigger`** (read-only): the backend decides whether a run starts and which agent runs. Never guessed on the frontend.
- **`RunConfirmModal` + `useRunConfirm`**: gate mirrors multica (`agent`/`squad` assignee on a non-backlog issue). On confirm the user can **start now** (optional handoff note) or **assign without starting** (`suppress_run`). Member / unassign / backlog assignments skip the modal and apply directly.
- `UpdateIssueReq` gains `suppress_run` / `handoff_note`; wired at the detail assignee picker, the `···` menu, and the list inline picker.
- `AssigneePicker.onChange` now emits `(id, type, name)` so callers don't re-resolve the name (drops a duplicate candidate load).

## Verification (live daemon runtime, agent "E2E Coworker", Alice)

| Path | Result |
|---|---|
| **Start** (agent + handoff note) | real task dispatched → agent ran to **completed**, `handoff_note` persisted on the task (DB-confirmed) |
| **Don't start** (suppress) | assignee set, **0 tasks** dispatched (DB-confirmed) |
| Member assign | no modal, applied directly |
| Re-assign same agent | modal shows "won't trigger" (correct — not a change) |

Ran `/simplify` (4-dimension); applied the consensus fix (widen `onChange` to drop duplicate name lookup). Typecheck clean on changed files.

## Notes

- **Stacked on #602** (issue CRUD). Merge #602 first; the two issue-crud commits shown in this diff drop out once #602 lands.
- Status-change-triggered runs (`backlog → active`) are modeled in the types but not yet wired to the confirm dialog — follow-up within Phase A.
- No linked issue yet — `check-sprint` needs a maintainer to attach one.
- Docs: dmloop feature UI; no service/command/config change, so no dev-README/deployment doc update needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #629, opened 2026-07-10
refactor(loop): make requestAssign symmetric with requestStatus

MrAladdin:
## What

Retrospective `/simplify` of the merged Phase A dmloop work (#605/#606/#608/#609/#610/#613/#623), which shipped with adversarial correctness review but without a formal `/simplify` pass. Ran `/simplify` (reuse / simplification / efficiency / altitude) over the combined diff; this PR applies the one material finding plus two trivial cleanups.

## Material: symmetric `requestAssign`

`useRunConfirm`'s `requestAssign` required each call site to hand-build the full `RunConfirmRequest` object — `{issueId, status, assigneeType, assigneeId, assigneeName, apply}` — repeated verbatim in three places (issue detail ··· menu, detail properties picker, list row). If the request shape changes, all three must change in lockstep (drift risk). `requestStatus` already assembled its request internally; `requestAssign` was the asymmetric odd one out.

Fix: `requestAssign(issue, assigneeType, assigneeId, assigneeName, apply)`, assembling the request inside the hook (status stays the issue's current status — assigning doesn't change it). The three call sites collapse to a single call each; `RunConfirmRequest` becomes hook-internal (no external importers). **Behavior unchanged.**

## Also

- Removed an orphaned comment (isActiveRun note left after the logic moved inline) and a stray double blank line.

## Deliberately skipped (noted, not applied)

From the same `/simplify` pass, judged not worth the churn/risk in this cleanup:
- RunDetailModal poll: parallelize the two serial GETs / avoid full `listRuns` per tick — touches the live-transcript feature for a marginal per-2s RTT; the full-list refetch needs a backend single-run endpoint anyway.
- Transcript dedup Set rebuild — behavior-sensitive (depends on `?since` interval semantics), current form verified working in #610.
- Package-wide `Toast.error` helper, `IssueTriggerPreview` mirror fields, `rerunIssue` unused branch — out of scope / low value.

Confined to dmloop. `RunConfirmRequest` had no external importers (de-export safe). `IssueBoard` uses `useRunConfirm` but only `requestStatus`, whose signature is unchanged. `requestStatus` / `runConfirmModal` exports unchanged. `@octo/loop` is a leaf package.

## Verification

- **Adversarial review**: a Claude reviewer confirmed the three reassembled requests are field-for-field identical to the originals, all three call sites pass a non-null `issue` (list row is `Issue`; ··· menu has an `issue &&` guard; properties picker is past the `if (!issue) return` early-return), `needsConfirm` semantics unchanged, and de-export is safe. Codex was also invoked for the two-model cross-check but did not complete a verdict this session (as in the prior PR) — flagged honestly.
- **Real browser (dev env, MV2 E2E / MVE-4)**:
  - Assign to a **member** (Bob) via the properties picker → applies directly, no confirmation modal (member path).
  - Assign to an **agent** (E2E Coworker) on an in-progress issue via the ··· menu → the dispatch-confirmation modal appears (preview ran); chose "暂不开始" → assignee applied without dispatching a run (execution log stayed 0). Assignee restored to its original value.

### DIFF
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -36,7 +36,6 @@ interface Filters {
 
 const PAGE_SIZE = 50;
 
-
 export default function IssuePage() {
   const { t } = useI18n();
   const [issues, setIssues] = useState<Issue[]>([]);
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -234,8 +234,6 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     });
   };
 
-  // 运行中(可终止)vs 已结束(可重跑):用共享 isActiveRun(ui/meta)。
-
   const saveDesc = async () => {
     if (descDraft !== (issue?.description ?? "")) await patch({ description: descDraft });
     setEditingDesc(false);
@@ -293,7 +291,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
                   <Dropdown.Divider />
                   <Dropdown.Title>{t(`loop.assignee.${type}`)}</Dropdown.Title>
                   {items.map((c) => (
-                    <Dropdown.Item key={c.id} active={issue?.assignee_id === c.id} onClick={() => issue && requestAssign({ issueId: issue.id, status: issue.status, assigneeType: c.type, assigneeId: c.id, assigneeName: c.name, apply: (extra) => patch({ assignee_id: c.id, assignee_type: c.type, ...extra }) })}>
+                    <Dropdown.Item key={c.id} active={issue?.assignee_id === c.id} onClick={() => issue && requestAssign(issue, c.type, c.id, c.name, (extra) => patch({ assignee_id: c.id, assignee_type: c.type, ...extra }))}>
                       {c.name}
                     </Dropdown.Item>
                   ))}
@@ -547,7 +545,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
                 <AssigneePicker
                   value={issue.assignee_id}
                   valueName={issue.assignee_name ?? null}
-                  onChange={(id, type, name) => requestAssign({ issueId: issue.id, status: issue.status, assigneeType: type, assigneeId: id, assigneeName: name, apply: (extra) => patch({ assignee_id: id, assignee_type: type, ...extra }) })}
+                  onChange={(id, type, name) => requestAssign(issue, type, id, name, (extra) => patch({ assignee_id: id, assignee_type: type, ...extra }))}
                 />
               </dd>
               <dt>{t("loop.field.project")}</dt>
--- a/packages/dmloop/src/panel/IssueList.tsx
+++ b/packages/dmloop/src/panel/IssueList.tsx
@@ -107,7 +107,7 @@ export default function IssueList({
           size="small"
           value={r.assignee_id}
           valueName={r.assignee_name ?? null}
-          onChange={(id, type, name) => requestAssign({ issueId: r.id, status: r.status, assigneeType: type, assigneeId: id, assigneeName: name, apply: (extra) => patch(r.id, { assignee_id: id, assignee_type: type, ...extra }) })}
+          onChange={(id, type, name) => requestAssign(r, type, id, name, (extra) => patch(r.id, { assignee_id: id, assignee_type: type, ...extra }))}
         />
       ),
     },
--- a/packages/dmloop/src/ui/RunConfirmModal.tsx
+++ b/packages/dmloop/src/ui/RunConfirmModal.tsx
@@ -6,8 +6,9 @@ import { previewIssueTrigger } from "../api/issueApi";
 
 const { Text } = Typography;
 
-/** 一次指派/状态变更请求：apply 由调用方给出（真正落库的 updateIssue 调用）。 */
-export interface RunConfirmRequest {
+/** 一次指派/状态变更请求：apply 由调用方给出（真正落库的 updateIssue 调用）。
+ *  hook 内部类型——调用方走 requestAssign / requestStatus,不直接构造。 */
+interface RunConfirmRequest {
   issueId: string;
   status: IssueStatus;
   assigneeType: AssigneeType | null;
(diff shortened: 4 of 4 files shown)


OctoBoooot (approved):
## Review: make requestAssign symmetric with requestStatus (#629)

**Verdict:** Approve
Clean behavior-preserving `/simplify` refactor. `requestAssign` now assembles its `RunConfirmRequest` internally (like `requestStatus`), collapsing three hand-built request literals into single calls and removing real drift risk. Verified field-for-field parity at all three call sites.

### What I verified (behavior-unchanged claim)
The hook now builds `{ issueId: issue.id, status: issue.status, assigneeType, assigneeId, assigneeName, apply }`. Each call site's old literal vs new derived request, field by field:
- **IssueDetailPage ··· menu (`:291`)** — old `{issue.id, issue.status, c.type, c.id, c.name, apply}` = new `requestAssign(issue, c.type, c.id, c.name, apply)`. Identical; the `issue &&` guard is retained.
- **IssueDetailPage properties picker (`:545`)** — old `{issue.id, issue.status, type, id, name, apply}` = new `requestAssign(issue, …)`. Identical; reached only past `if (!issue) return`, so `issue` is non-null.
- **IssueList row (`:107`)** — old `{r.id, r.status, type, id, name, apply=patch(r.id,…)}` = new `requestAssign(r, …)`. Identical; `r` is a non-null `Issue`, and `apply` still closes over `patch(r.id, …)`.
- **Internal flow unchanged:** `if (!needsConfirm(r)) applyDirect(apply); else setPending(r)` — same control path as the old `applyDirect(r.apply)`; `needsConfirm` sees the same object. `status: issue.status` correctly preserves the current status (assigning doesn't change it).
- **De-export safe:** `RunConfirmRequest` is now hook-internal; no external importers (IssueBoard uses only `requestStatus`, whose signature is untouched). The other two files are cosmetic (orphaned comment + stray blank line).

### Praise
- Making `requestAssign` symmetric with `requestStatus` is exactly the right altitude fix: the asymmetry (one hand-builds the request, the other assembles internally) was the actual drift hazard, and collapsing three verbatim literals into the hook removes the lockstep-maintenance risk without changing a single observable behavior. The parity is exact, not approximate.

Jerry-Xin (approved):
Symmetric refactor is correct: `requestAssign` now assembles the `RunConfirmRequest` inside the hook (same as `requestStatus`), and the dispatch gate is byte-for-byte preserved. Approving.

### Dispatch-gate parity (verified at head `2a8afa8`)
- **Gate preserved**: `requestAssign` still routes through `if (!needsConfirm(r)) { applyDirect(apply); return; } setPending(r);` (`RunConfirmModal.tsx:73`). Only the call signature changed (object arg → `issue, type, id, name, apply`); the branch logic is identical to pre-refactor.
- **Preview + confirm on trigger**: when `needsConfirm` is true, `setPending(r)` mounts the modal, which calls `previewIssueTrigger(...)` and dispatches **only** via the `run(...)` handlers behind the Start / Suppress / Apply buttons — no auto-dispatch. Dispatch requires an explicit click.
- **Fail-safe = fail-closed**: `previewIssueTrigger(...).catch(...)` sets `errored=true; setWillStart(true)` (`RunConfirmModal.tsx:135-138`), so a preview failure surfaces the warning and presents Suppress/Start — it never silently dispatches. `run()` wraps `apply` in try/catch with `Toast.error`, no silent swallow.
- **No silent-dispatch bypass**: all 5 `requestAssign` call sites (IssueDetailPage:294, :548; IssueList:110) were migrated to the new positional signature; grep confirms no remaining old object-shape caller. Non-triggering assigns (member / unassign / backlog) correctly go straight to `applyDirect`, matching prior behavior.

### requestStatus untouched
`requestStatus` and its `statusMightTrigger` gate are unchanged by this PR (diff touches only `requestAssign`, the shared type visibility, and comments). No regression to the status path. The intentional asymmetry — assign uses `needsConfirm` (agent/squad + non-backlog), status uses `statusMightTrigger` (backlog→active) — is preserved and correct given each actor's semantics.

### Minor / non-blocking
- `RunConfirmRequest` is now correctly narrowed from `export interface` to a hook-internal `interface`, enforcing that callers go through `requestAssign`/`requestStatus` rather than hand-constructing requests. Good hardening.
- Assignee/agent names rendered via `t("loop.run.willStart", { values: { name } })` and Semi `<Text>` — no raw HTML injection, i18n keys used throughout.
- Modal uses a `cancelled` guard in the effect cleanup to drop stale preview responses on issue switch — no setState-after-unmount race.

Clean, behavior-preserving symmetry refactor. LGTM.

yujiawei (approved):
# Code Review — PR #629 (octo-web)

Independent review at head SHA `2a8afa89bc98df222dfe354f7223e4ef36ed9365`, base `feat/octo-loop`. Scope: 4 files, +25/-11, a retrospective `/simplify` cleanup confined to `@octo/loop`.

## Spec / scope compliance ✅

The diff contains exactly what the description claims, nothing more:

- **Symmetric `requestAssign`** — `packages/dmloop/src/ui/RunConfirmModal.tsx:61-78`: signature changed from `(r: RunConfirmRequest)` to `(issue, assigneeType, assigneeId, assigneeName, apply)`, assembling the request internally with `status: issue.status`. This mirrors the existing `requestStatus` shape.
- **`RunConfirmRequest` de-exported** — `RunConfirmModal.tsx:11`. Verified no importers exist outside this file, so dropping `export` is safe.
- **Cleanups** — orphaned `isActiveRun` comment removed (`IssueDetailPage.tsx`), stray blank line removed (`IssuePage.tsx`).

`requestStatus`, `runConfirmModal`, and all other exports are unchanged. No new features, flags, or fields; no missing changes. **Spec: PASS.**

## Correctness ✅

**Behavior equivalence — verified field-for-field at all three call sites.** The pre-refactor pattern `requestAssign({ issueId: issue.id, status: issue.status, assigneeType, assigneeId, assigneeName, apply })` produces a request object identical to what the new internal assembly builds, and the `needsConfirm` gate / `applyDirect` control flow is untouched.

The one real risk in this kind of change is **argument-order drift**, since `AssigneePicker.onChange` is `(id, type, name)` while the new `requestAssign` is `(issue, assigneeType, assigneeId, assigneeName, apply)`. Both `AssigneePicker`-based sites map it correctly:

- `IssueDetailPage.tsx:548` — `requestAssign(issue, type, id, name, ...)` ✓ (`(id,type,name)` → `(type,id,name)`)
- `IssueList.tsx:110` — `requestAssign(r, type, id, name, ...)` ✓
- `IssueDetailPage.tsx:294` (··· menu) — `requestAssign(issue, c.type, c.id, c.name, ...)` ✓ (uses candidate object fields directly)

**Null-safety — all sites pass a non-null `Issue`:**

- `IssueDetailPage.tsx:294` guarded by `issue &&` short-circuit.
- `IssueDetailPage.tsx:548` is past the `if (!issue)` early return at `IssueDetailPage.tsx:322`, so `issue` is non-null for that JSX.
- `IssueList.tsx:110` receives `r: Issue` from the column `render` signature.

So `issue.id` / `issue.status` inside the hook can never dereference null.

## Notes (non-blocking)

- The refactor genuinely reduces drift risk: `RunConfirmRequest`'s shape is now assembled in exactly one place per actor path instead of hand-built at three call sites. Good call keeping it hook-internal.
- Nice touch making `requestAssign` and `requestStatus` structurally parallel — a future reader can now reason about both from one mental model.

## Coverage / blind spots

- No unit/type-check was run in this review (monorepo toolchain not installed for a 4-file refactor); correctness assessed by reading. The change is type-driven, so `tsc` in CI is the authoritative gate — a wrong argument order would fail to compile given the distinct `Issue` vs `AssigneeType` types.
- Runtime behavior (modal preview, member vs agent paths) was not re-executed here; the PR reports a real-browser E2E pass, and the logic paths are unchanged by this diff.

## Verdict

No P0/P1/P2 findings. Behavior is unchanged, scope is clean, all call sites are correct and null-safe.

PR #632, opened 2026-07-10
feat(loop): show issue labels and due dates on the issue views

MrAladdin:
## What

Surface issue data the backend already returns but the client was dropping:

- **Labels**: a shared `<LabelChips>` component renders an issue's labels as
  colored chips (per-label hex → tinted bg + matching text; falls back to the
  default Tag style for non-`#RRGGBB` colors; folds overflow to "+N"). Shown on
  list rows, board cards, and the detail properties panel. `types.ts` gains
  `IssueLabel` + optional `Issue.labels` (list/detail endpoints backfill it;
  `enrich()` already passes it through via spread, so no API-layer change).
- **Dates**: `meta.tsx` gains `formatShortDate` + `isOverdue`; board cards show
  a due-date badge (red when overdue and not done/cancelled); the detail
  properties panel shows start/due rows (due in danger color when overdue).

## Bug fix (made visible by labels)

`IssueDetailPage.patch()` replaced the issue with the **un-enriched** PUT
response, which lacks `labels` (and `assignee_name`/`project_name` — display
fields only list/detail endpoints backfill). So any inline edit in the detail
view wiped those fields until leave+reenter. `patch()` now re-enriches the PUT
response (recomputes name fields for the new values) and preserves labels;
extracted an `enrichIssue()` helper (also DRYs `getIssue`).

## Quality gates

- **/simplify** (4 angles): the labels slice came back clean (shared component,
  right altitude; the one flagged redundancy — board's outer guard — is a
  deliberate spacing-wrapper control, kept).
- **Adversarial review, two models**: a Claude reviewer found the `patch()`
  label-wipe bug (confidence 85) — fixed as above; a Codex review of the final
  diff (inlined, no repo exploration) confirmed no remaining defect (patch fix
  correct, `isOverdue`/hex/`LabelChips` guards safe, blast-radius compatible).
- **Real browser (dev env, MV2 E2E / MVE-4 with injected labels + due date)**:
  - Board card: `P0`/`回归` chips (tinted per hex) + due badge `7/9` in red (overdue).
  - Detail properties: labels row + start(`-`)/due(`7/9 08:00`, danger) rows.
  - Bug-fix regression: changing priority in the detail view keeps the labels
    row intact (previously flipped to `—`).
  - Loop-dev (no labels/dates) renders cleanly with nothing shown.

## Blast radius

Additive only: `Issue.labels` optional field, two new `meta` exports, new
`enrichIssue` export, new `LabelChips` component. No existing export signature
changed; other `Issue`/`meta` consumers unaffected; `@octo/loop` is a leaf
package. Verified against the just-merged #627 (project filter) and #629
(requestAssign) — rebased cleanly on top, requestAssign call sites use #629's form.

### DIFF
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -44,9 +44,13 @@ export async function listIssues(
   return { issues, total: data.total ?? issues.length };
 }
 
+export async function enrichIssue(issue: Issue): Promise<Issue> {
+  return (await enrich([issue]))[0];
+}
+
 export async function getIssue(id: string): Promise<Issue> {
   const issue = await httpGet<Issue>(`/issues/${id}`);
-  return (await enrich([issue]))[0];
+  return enrichIssue(issue);
 }
 
 export function createIssue(req: CreateIssueReq): Promise<Issue> {
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -74,6 +74,13 @@ export type IssueStatus =
   | "backlog" | "todo" | "in_progress" | "in_review" | "done" | "blocked" | "cancelled";
 export type IssuePriority = "urgent" | "high" | "medium" | "low" | "none";
 
+/** issue 标签(后端 list/detail 端点批量回填 issue.labels)。color 为 hex。 */
+export interface IssueLabel {
+  id: string;
+  name: string;
+  color: string;
+}
+
 export interface Issue {
   id: string;
   workspace_id: string;
@@ -102,6 +109,8 @@ export interface Issue {
   // octo 头像 URL（member 型 actor，由 directory 回填；agent/squad/原生成员为空）
   assignee_avatar?: string | null;
   creator_avatar?: string | null;
+  // 后端 list/detail 端点批量回填；其它端点(update/ws)不带 → 保持已有。
+  labels?: IssueLabel[] | null;
 }
 
 export interface CreateIssueReq {
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -129,7 +129,10 @@
     "assignee": "Assignee",
     "project": "Project",
     "noProject": "No project",
-    "creator": "Creator"
+    "creator": "Creator",
+    "labels": "Labels",
+    "startDate": "Start date",
+    "dueDate": "Due date"
   },
   "status": {
     "backlog": "Backlog",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -129,7 +129,10 @@
     "assignee": "负责人",
     "project": "项目",
     "noProject": "无项目",
-    "creator": "创建者"
+    "creator": "创建者",
+    "labels": "标签",
+    "startDate": "开始日期",
+    "dueDate": "截止日期"
   },
   "status": {
     "backlog": "待规划",
--- a/packages/dmloop/src/pages/loop.css
+++ b/packages/dmloop/src/pages/loop.css
@@ -377,6 +377,13 @@
   flex-wrap: wrap;
 }
 
+.loop-card__due {
+  display: inline-flex;
+  align-items: center;
+  gap: 3px;
+  font-size: 12px;
+}
+
 /* ===== 通用表格辅助 ===== */
 .loop-cell-title {
   font-weight: 500;
--- a/packages/dmloop/src/panel/IssueBoard.tsx
+++ b/packages/dmloop/src/panel/IssueBoard.tsx
@@ -1,14 +1,18 @@
 import React, { useState } from "react";
 import { Tag } from "@douyinfe/semi-ui";
+import { CalendarClock } from "lucide-react";
 import { useI18n } from "@octo/base";
 import type { Issue, IssueStatus } from "../api/types";
 import { updateIssue } from "../api/issueApi";
 import { AssigneeBadge } from "../ui/AssigneePicker";
+import LabelChips from "../ui/LabelChips";
 import { useRunConfirm } from "../ui/RunConfirmModal";
 import {
   ISSUE_STATUS_ORDER,
   ISSUE_STATUS_COLOR,
   PRIORITY_COLOR,
+  formatShortDate,
+  isOverdue,
 } from "../ui/meta";
 
 export interface IssueBoardProps {
@@ -83,10 +87,22 @@ export default function IssueBoard({
                 >
                   <div className="loop-card__key">{issue.identifier}</div>
                   <div className="loop-card__title">{issue.title}</div>
+                  {issue.labels && issue.labels.length > 0 && (
+                    <div style={{ marginTop: 4 }}><LabelChips labels={issue.labels} max={3} /></div>
+                  )}
                   <div className="loop-card__foot">
                     <Tag color={PRIORITY_COLOR[issue.priority]} size="small">
                       {t(`loop.priority.${issue.priority}`)}
                     </Tag>
+                    {issue.due_date && (
+                      <span
+                        className="loop-card__due"
+                        style={{ color: isOverdue(issue.due_date, issue.status) ? "var(--semi-color-danger)" : "var(--semi-color-text-2)" }}
+                      >
+                        <CalendarClock size={12} />
+                        {formatShortDate(issue.due_date)}
+                      </span>
+                    )}
                     <AssigneeBadge
                       type={issue.assignee_type}
                       name={issue.assignee_name ?? null}
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -36,6 +36,7 @@ import type {
 import {
   getIssue,
   updateIssue,
+  enrichIssue,
   deleteIssue,
   listComments,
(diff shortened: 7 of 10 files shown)


Jerry-Xin (approved):
Solid, defensively-coded additive feature — renders issue labels + due/start dates across board, list, and detail. Reviewed #632's own diff (base `feat/octo-loop`, merge-base == base, zero rebase drift). Nothing blocking.

### ✅ Security / correctness verified
- **Label rendering XSS (primary concern): safe.** `LabelChips` renders `{l.name}` as JSX children → React auto-escapes. No `dangerouslySetInnerHTML` anywhere in the diff.
- **Label color CSS-injection: safe.** `color` is strictly validated with `/^#[0-9a-fA-F]{6}$/` before any use; only a matching hex is interpolated into an inline `style` **object** (`backgroundColor: ${hex}1f`, `color: hex`, `border: 1px solid ${hex}3d`). A crafted value like `red;background:url(...)` fails the regex → falls back to `undefined` (default Tag). Object-style + regex gate = no injection surface.
- **Due-date correctness.** `isOverdue` (ui/meta.tsx) correctly returns `false` for null due, `done`, and `cancelled`, then does a strict timestamp compare `getTime() < Date.now()` — no day-boundary off-by-one since due_date is a full timestamp, not date-only. Detail-page `danger` styling and board `--semi-color-danger` color both key off it consistently.
- **Null/empty handling: clean.** `LabelChips` returns `null` for empty/nil labels; board gates with `labels && labels.length > 0`; detail shows `—` fallback; list passes `r.labels` straight through (handled internally). No empty chips / "null" leakage.
- **Perf on board: fine.** Per-card cost is one regex test per label + `slice(0, max)`; labels are pre-populated on the issue (backend batch backfill per types.ts) — no N+1 fetch, cheap over the 100-issue board cap.
- **i18n: symmetric.** Both en-US and zh-CN gain `field.labels/startDate/dueDate` and `filter.project`.

### 💬 Non-blocking notes
- `formatShortDate` / `fmt` use local `getMonth/getDate/getHours` for timezone. This matches the pre-existing `fmt` already used for `created_at`/`dispatched_at`, so it's consistent app-wide — just flagging that due dates render in the viewer's local tz (acceptable given existing convention).
- Both date helpers guard null but not `Invalid Date` (a bad ISO would render `NaN/NaN`). Backend-controlled input makes this low-risk; a `Number.isNaN(d.getTime())` guard would harden it if label/date sources ever widen.
- The `directory.ts` in-flight dedup (`if (!force && _loading) return _loading`) and `listProjectOptions` (reuses cached directory, no extra `/projects` request) are good decoupling; `requestAssign` refactor to take `(issue, type, id, name, apply)` reduces call-site boilerplate cleanly and the `RunConfirmRequest` type is now properly module-private.

### ✅ Highlights
- Strict hex-regex color gate on the one real injection vector for a label-render feature — exactly the right call.
- Shared `LabelChips` / date helpers keep board/list/detail rendering consistent.
- `clickToHide` added across the Dropdowns and `stopPropagation` on menu triggers — nice UX/correctness cleanup bundled in.

CI: all real checks SUCCESS/SKIPPED; mergeable. LGTM.

OctoBoooot (approved):
## Review: show issue labels and due dates on the issue views (#632)

**Verdict:** Approve
Clean additive feature (labels chips + due dates on list/board/detail) that also fixes a real latent bug in `patch()`. Verified the injection surface, the date helpers, and the bug fix against the code.

### What I verified
- **`LabelChips` is injection-safe:** the per-label `color` is gated by `/^#[0-9a-fA-F]{6}$/` before it reaches the inline `style`; anything else falls back to the default Tag. Only a strict 6-hex color is interpolated (`${hex}1f` / `3d` alpha suffixes → valid 8-digit hex), so no CSS injection via untrusted label color, and no DEVELOPMENT.md §2 issue (runtime data, validated — not a hardcoded literal). Null/empty labels return null; `max` folds overflow to `+N`.
- **Date helpers correct:** `isOverdue(due, status)` fails safe on null / `done` / `cancelled`, else `due < now`; `formatShortDate` handles null → "". Board/detail color via `var(--semi-color-danger)` / tokens — no hardcoded colors.
- **`patch()` label-wipe fix (the substantive change):** before, `setIssue(await updateIssue(...))` stored the raw PUT response, which lacks enriched display fields (`assignee_name`/`project_name`) **and** `labels` (only list/detail endpoints backfill), so any inline edit blanked them until leave/re-enter. Now `setIssue({ ...(await enrichIssue(updated)), labels: updated.labels ?? issue.labels })` — re-enriches name fields for the **new** values (so a changed assignee shows the right name, not a stale one) and preserves labels. Correct on both axes.
- **`enrichIssue` extraction** is a behavior-identical DRY of `getIssue`'s existing `enrich([x])[0]`; types are additive (`IssueLabel`, optional `Issue.labels`) — no signature change, `enrich()` already spreads labels through.
- **Usages consistent:** board card (labels `max=3`, due badge guarded by `issue.due_date`), list cell (LabelChips self-guards), detail rows — all handle absent data.

### Praise
- The feature surfaced a pre-existing bug and the fix is the right shape: re-enriching the PUT response rather than special-casing labels means `assignee_name`/`project_name` (which had the *same* wipe-on-edit problem, just less visible) are fixed too — and extracting `enrichIssue` removes the duplication instead of copy-pasting the enrich call. Validating the label hex before styling is exactly the guard an untrusted-color chip needs.

yujiawei (approved):
# Code Review — PR #632 (octo-web)

Reviewed at head SHA `0488977ccbf258db5a1d55762643e8717be63775` against merge-base `e88da0ef` (parent on `feat/octo-loop`). Independent re-verification of the diff, plus a cross-family advisory pass.

## Summary

Solid, additive change that surfaces `labels` + start/due dates already returned by the backend but dropped client-side. The centerpiece — the `patch()` label-wipe fix in the detail view — is real and correctly fixed. No blocking defects found; the remaining items are non-blocking polish.

## Spec / scope

- No linked issue, so this is judged purely on the diff. Scope is coherent and matches the PR description: shared `LabelChips`, `meta` date helpers, board/list/detail wiring, and the `patch()` re-enrich fix.
- Change is strictly additive: `Issue.labels` is optional, new `meta` exports, new `enrichIssue` export, new component. No existing export signature changed. Nothing built beyond what the PR states.

## Correctness — the `patch()` fix (the important one)

`IssueDetailPage.patch()` previously did `setIssue(await updateIssue(...))`, replacing the issue with the **un-enriched PUT response**, which lacks `labels`, `assignee_name`, and `project_name` — so any inline edit wiped those display fields until leave+re-enter. The fix:

```ts
setIssue({ ...(await enrichIssue(updated)), labels: updated.labels ?? issue.labels });
```

is correct: `enrichIssue` recomputes the name/avatar/project fields from the new values, and `labels` is explicitly preserved (PUT may not backfill it). `enrichIssue` extraction also cleanly DRYs `getIssue`. The board's drag-to-status path (`IssueBoard.handleDrop`) does not have this problem because it calls `onChanged()` → parent re-fetch via `listIssues`, which backfills labels.

## Non-blocking findings (P2 / nits)

1. **`meta.tsx` — date parsing is local-timezone (`formatShortDate`, `isOverdue`).** `new Date(iso)` + `.getMonth()/.getDate()`/`Date.now()` are evaluated in the browser's local timezone. This is only a real off-by-one **if** the backend ever sends *date-only* strings (`"2026-07-09"`, parsed as UTC midnight). Here it does not — `due_date` renders as `7/9 08:00` (full ISO datetime), and the pre-existing `fmt()` in the same file already uses the identical pattern for `created_at`/`due_date`. So this is consistent with existing behavior and not a regression. Worth a guard only if date-only payloads become possible.

2. **`LabelChips.tsx:18` — hex regex accepts only `#RRGGBB`.** `/^#[0-9a-fA-F]{6}$/` rejects valid 3-digit (`#fff`) and 8-digit (`#RRGGBBAA`) hex. Non-matching colors fall back to the default Semi Tag style (graceful, no crash), so this is cosmetic. Fine if the backend only ever emits 6-digit hex; loosen the regex if that's not guaranteed.

3. **`IssueDetailPage.patch()` — no `try/catch`.** A rejected `updateIssue`/`enrichIssue` leaves local state unsynced with no user-facing error toast. Note this is a **pre-existing pattern** (the old `patch()` had no error handling either) — not introduced by this PR — so it's out of scope to block on, but a `try { ... } catch { Toast.error(...) }` here would match the error handling used elsewhere in this file (e.g. `saveEdit`).

4. **`IssueList.tsx:54` — labels render inline right after the title text** with only the flex `gap` inside `LabelChips` for separation; there's no leading margin between the title string and the first chip. Minor visual tightness; confirm it reads well when a title is long.

## Coverage / blind spots

- Could not run `tsc`/build (fresh checkout, no `node_modules`). Reviewed statically; `CalendarClock` is a valid `lucide-react` export and `lucide-react` is a declared dep, so the new import is sound.
- Backend date payload shape (date-only vs full datetime) was inferred from the PR's own screenshots + existing `fmt()` usage, not verified against the API contract — see finding #1.
- WebSocket/live-update path was checked: detail view has no ws subscription that would re-inject an un-enriched issue, so the `patch()` fix is not undone by a later push.
- One advisory leg (line-level cross-family reviewer) was unavailable this run (gateway 503); this verdict rests on independent diff re-verification plus a long-context advisory leg. Absence of that leg is noted, not treated as approval.

## Verdict

No P0/P1 issues. The four items above are non-blocking (cosmetic / pre-existing / conditional). The core bug fix is correct and the change is cleanly additive.

PR #634, opened 2026-07-10
feat(loop): filter issues by created/updated date range

MrAladdin:
## What

Add a **created/updated date-range filter** to the issue list/board, alongside
the existing status / priority / assignee / creator / project filters.

Backend `GET /issues` accepts `date_field` (`created_at` | `updated_at`) +
`date_start` + `date_end` — all three required together, RFC3339, `date_start`
strictly before `date_end`.

- `ListParams` gains `date_field`/`date_start`/`date_end` (+ `ISSUE_DATE_FIELDS`
  single-source const, mirroring `ISSUE_SORT_FIELDS`).
- A Semi `DatePicker` (`dateRange`) sits in the filter row; a field select
  (created/updated) appears only once a range is picked (progressive disclosure).
- The selected `[start, end]` is sent as a **half-open** range: `date_start =
  start`, `date_end = end + 1 calendar day` (`setDate`, DST-safe). This includes
  the end day and guarantees `start < end` even when start == end, satisfying the
  backend's strict-before constraint. The three params are sent only when both
  ends are present (single gate → atomic).

## Quality gates

- **/simplify** (4 angles): applied two findings — collapsed a redundant
  `hasDate` re-check (the onChange already normalizes `dateRange` to
  `undefined | [start,end]`, removing three `!` assertions) and switched the
  end+1 from a `+86400000` ms literal to `setDate(+1)` (self-documenting +
  correct across DST).
- **Adversarial review, two models**: a Claude reviewer verified `start<end`
  always holds, that Semi's `dateRange` guarantees ascending order (source-level:
  panel rejects reverse selection, typed input is sorted), atomicity, RFC3339
  compatibility, and the normalization guard — no confirmed bug. Codex (minimal
  prompt) raised only the local-day→UTC timezone question and self-classified it
  as backend-convention-dependent, not a defect; on analysis the behavior is
  correct (filters by the user's local calendar days).
- **Real browser (dev env, MV2 E2E — 3 issues created today)**:
  - Range 7/1–7/8 (excludes today) → board empties (0 issues), and the field
    select appears.
  - Range 7/1–7/10 (end+1day = 7/11 exclusive → includes today) → all 3 issues
    return. Confirms the half-open logic and no 400 (start<end held).

## Blast radius

Additive optional `ListParams` fields; `listIssues` is the sole consumer that
reads them (other list helpers ignore the new optional fields; the http layer
drops undefined params). `@octo/loop` is a leaf package. Rebased cleanly on the
current tip (#627/#629/#632/#633/#619 merged).


OctoBoooot (approved):
## Review: filter issues by created/updated date range (#634)

**Verdict:** Approve
Clean date-range filter following the established filter pattern. The correctness-critical part — the half-open range with `end + 1 day` to satisfy the backend's strict `start < end` — is right, including the single-day and DST boundaries. Verified by construction.

### What I verified (boundary construction)
- **Half-open end (`setDate(+1)`):**
  - Single-day (start==end, e.g. 7/10): `date_start=7/10`, `endExclusive=7/11` → `start<end` holds (satisfies backend strict-before) and `[7/10,7/11)` includes all of 7/10. This is exactly the case the +1 exists for — correct.
  - Multi-day (7/1–7/8): `[7/1, 7/9)` includes 7/8 fully.
  - DST: `setDate(getDate()+1)` advances a calendar day (not a `+86400000` ms literal), so it's correct across DST — matches the /simplify note.
- **No state mutation:** `new Date(dr[1])` clones before `setDate`, so the `dateRange` state array isn't mutated.
- **Atomic 3-param send:** `date_field`/`date_start` gated on `dr`, `date_end` on `endExclusive` (truthy iff `dr` is) — so all three are sent together or all omitted, satisfying "all three required together." Undefined params drop at the http layer.
- **Normalization guard:** `onChange` sets `dateRange` to `[start,end]` only when both ends are present, else `undefined` — so `dr[0]`/`dr[1]` are never null-derefed. Progressive disclosure (field select appears only once a range is picked) works off the same flag. Page resets via shared `update()`.
- **No new dependency:** the `dependencies-changed` label / pnpm-lock delta is just the merged `@octo/personal` workspace package being recorded after rebase (lucide-react already present) — no third-party addition.

### Note (non-blocking)
- `toISOString()` sends the user's local calendar-day boundaries as UTC instants — correct for a backend comparing RFC3339/UTC timestamps (filters by the user's local days, as expected). This is the timezone point Codex raised; behavior is correct, worth a comment if the backend ever changes convention.

### Praise
- The half-open `end+1day` is the right way to satisfy a strict `start<end` backend while keeping the end day inclusive, and choosing `setDate(+1)` over a ms literal makes it DST-correct and self-documenting. The single-gate atomic send (all three params or none) avoids the partial-filter 400 that this kind of multi-param range usually trips on.

Jerry-Xin (approved):
Clean, well-scoped date-range filter that reuses the established filter pattern on this surface (server-side bound params, page reset, stale-guard). Approving.

## ✅ Verified
- **Server-side, bound params (no injection).** `date_field/date_start/date_end` flow `ListParams → listIssues → httpGet("/issues", {...})`, and `httpGet` calls `client.get(path, { params: clean(params) })`. axios URL-encodes the values; no string interpolation into the path/query. `clean()` drops `undefined/null/""`, so nothing leaks when no range is set.
- **Atomic gate.** All three params are set only when a full range exists: `date_field: dr ? f.dateField : undefined`, `date_start: dr ? dr[0].toISOString() : undefined`, `date_end: endExclusive ? … : undefined`. `DatePicker.onChange` normalizes to `undefined | [start, end]`, so a partial/one-sided range never reaches the request.
- **Range validity + timezone.** End boundary is `end + 1` calendar day via `setDate(getDate()+1)` (DST-safe) → half-open interval that (a) includes the selected end day and (b) guarantees `start < end` even when start == end, satisfying the backend's strict-before constraint. Values are emitted as `toISOString()` (RFC3339 UTC), matching the documented backend contract. Inverted range is not constructible from the paired range picker.
- **No truncation trap.** The filter is applied server-side through the paged list query for BOTH board and list views (params go to the backend); it is NOT a client-only filter over the `limit:100` truncated board page. This avoids the trap previously flagged on this surface.
- **created vs updated independence.** A single `date_field` select chooses the column (`created_at | updated_at`), and the same range applies to whichever is chosen — no cross-wiring between the two fields.
- **Page reset + stale-guard.** `update()` resets `setPage(0)` on any filter change; `reload` deps `[f, view, page]`; the `seq.current` in-flight guard (`my !== seq.current`) is reused unchanged.
- **i18n.** `filter.dateRange` + `dateField.created_at/updated_at` added symmetrically to en-US and zh-CN.
- **XSS.** No user-controlled date/label is rendered as HTML; option labels come from i18n keys, values are ISO strings passed as axios params.

## 💬 Non-blocking
- The commit carries a `Co-Authored-By` trailer; repo owners may want to keep commit trailers consistent with project convention. Not a code concern.

Single source of truth via `ISSUE_DATE_FIELDS` (mirroring `ISSUE_SORT_FIELDS`) and progressive disclosure of the field select are nice touches. LGTM.

yujiawei (approved):
# Code Review — PR #634 (octo-web)

**Verdict: APPROVE.** `feat(loop): filter issues by created/updated date range` is a clean, well-contained frontend change. The date logic — the highest-risk part — is correct, and there are no blocking (P0/P1) issues.

## Scope / spec compliance ✅

The diff implements exactly the described feature and nothing more:
- `types.ts`: `ISSUE_DATE_FIELDS` whitelist (`created_at | updated_at`) + `date_field/date_start/date_end` added to `ListParams`, with a clear contract comment.
- `issueApi.ts`: passthrough of the three params to `GET /issues`.
- `IssuePage.tsx`: `DatePicker type="dateRange"` + a conditionally-rendered field selector.
- `i18n`: new keys present in **both** `en-US.json` and `zh-CN.json`.

The `pnpm-lock.yaml` churn (`@octo/personal` / `packages/dmpersonal`) is **not** feature scope creep: `@octo/personal` is already declared in `apps/web/package.json` and the entries only sync the lockfile. The target branch already contains these entries, so this resolves cleanly at merge.

## Correctness ✅

The half-open interval handling is the right approach and I verified the edge cases:
- `endExclusive = new Date(dr[1]); setDate(getDate() + 1)` shifts the inclusive end day to an exclusive next-day boundary. `setDate` correctly handles month/year rollover and DST.
- Same-day selection (`dr[0] === dr[1]`) still yields `start < end` (`day` vs `day+1`), satisfying the documented "start strictly before end" backend constraint — this is precisely why the `+1 day` exists.
- The three params are always sent together or all omitted, because `date_field`, `date_start`, and `date_end` are each gated on the same `f.dateRange` value. `http.ts` `clean()` drops `undefined`, so no partial/empty params leak.
- `toISOString()` converts the picker's local-midnight boundaries to UTC, so the filter edges sit at the user's local midnight — the intuitive expectation.
- The existing `seq.current` stale-response guard continues to cover the new params; changing the date filter routes through `update()` which resets to page 0.

## Non-blocking notes (nits / suggestions)

- **`Filters.dateField` is required but only meaningful when `dateRange` is set** (`IssuePage.tsx:36,50`). Harmless — the value is simply not sent when no range is active — but semantically it could be modeled as optional or reset on range clear. Purely cosmetic.
- **No automated test** covers the `+1 day` / same-day boundary logic. Consistent with the rest of this file, so not a blocker, but a small unit test around `endExclusive` would lock in the DST/month-rollover behavior.

## Coverage / blind spots (could not verify here)

- **Backend contract is not verifiable from this repo.** The fleet backend that serves `GET /issues` is not in this workspace, so I could not confirm the exact param names (`date_field/date_start/date_end`), the "all three required together" rule, or whether the range comparison is `<` vs `<=`. The frontend is internally consistent and the `+1 day` design is robust for a `<` (and effectively fine for `<=`) backend; worth a quick confirmation with the backend contract, but it does not block this frontend change.

PR #608, opened 2026-07-10
feat(loop): confirm dispatch on status change out of backlog

MrAladdin:
## What

Closes the gap the RunConfirm (#605) adversarial review flagged: an issue already assigned to an agent/squad that is moved **backlog → active** starts a run (backend `WillEnqueueRun` status source), but the status-change entry points bypassed the confirm dialog and dispatched **silently**.

- `useRunConfirm` gains `requestStatus(issue, next, apply)`: opens the same preview-backed confirm only when the change would trigger a run (agent/squad-assigned + `backlog` → non `backlog`/`done`/`cancelled`); otherwise applies directly. Preview sends the current (unchanged) assignee + new status, so the backend evaluates it as a status-source trigger.
- Wired all four status entry points: detail status `Select` + `···` menu, list inline `Select`, board drag-drop.
- Refactor: reuse the `Issue` type (dropped the redundant `StatusChangeIssue`) + a shared `isAgentAssignee()` predicate across both gates.
- Also fixes a pre-existing `IssueBoard` type error (`assignee_name ?? null`).

## Verification (live daemon, agent "E2E Coworker", Alice)

| Path | Result |
|---|---|
| backlog issue (agent-assigned) → todo | RunConfirm opens ("will start") |
| "Don't start yet" | status=todo, **0 tasks** dispatched (DB-confirmed) |
| todo → in_progress (not from backlog) | no modal, applied directly (no over-prompt) |

Setup (kept): created `[status触发验证]` backlog issue assigned to the agent — backlog assign correctly dispatched nothing.

## Review

- `/simplify`: reused `Issue` type + deduped the assignee predicate.
- **Adversarial correctness pass** — one confirmed low finding: the client-side gate keys on the *cached* assignee, so a concurrent reassignment could skip the confirm UX (the dispatch itself stays backend-correct). Documented as a known limitation (ponytail comment) matching multica's own client-side gate, rather than adding a preview call to every status change. No false positives dismissed beyond that.

## Notes

- **Stacked on #606 → #605 → #602**; merge in order.
- No linked issue yet — `check-sprint` needs a maintainer.
- Docs: dmloop feature UI; no service/command/config change → no dev-README/deployment doc update.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #635, opened 2026-07-10
feat(loop): keyword search + label editing (capability layer)

MrAladdin:
## What

第一枪「能力对齐」PR:关键词搜索 + 标签编辑。UI 保持最小(dmloop UI 后续会重做,当下把后端能力接进 API 层才是不会白做的部分)。

### 关键词搜索
- 新增 `searchIssues()` → `GET /issues/search`(后端对 标题/描述/评论 做全文检索并回高亮片段)。`IssuePage` 搜索框在有关键词时改走它;此前关键词被塞进 `/issues`(该端点根本不吃 keyword,等于空操作)。
- 搜索是独立端点/语义(不与筛选组合、后端上限 50),故单列函数;`listIssues` 不再发这个已死的 keyword 参数。

### 标签编辑
- 新增 `labelApi`:工作区标签 CRUD(`/labels`)+ issue 挂/摘(`/issues/:id/labels`)。
- 详情页标签行新增 `<LabelEditor>`:下拉勾选工作区标签挂/摘 + 内联建新标签。(取色器等留给 UI 重做。)

## Adversarial review(双模型)发现并已修的 2 个 bug
- label list 端点返回 `{labels:[...]}`(包裹,同 `/issues`)——客户端改为解包,否则下拉直接 crash。(Claude code-reviewer 抓到,codex:adversarial-review 漏掉——印证双模型价值。)
- 恢复 `ListParams.keyword`——它被另外 5 个 list 端点(project/skill/squad/agent/runtime)共用,只有 listIssues 停用它。(/simplify reuse agent 抓到 blast-radius。)

## /simplify
已跑:抽出 `fetchIssues()` 共享尾巴(list/search 共用 enrich+total)。

## 真实浏览器 e2e(dev 环境,Alice_Space / MV2 E2E 工作区)
- 搜索:输「头像」→ 精确命中 1 条(MVE-1),其余隐藏。
- 标签:新建「冒烟」→ 自动挂到 MVE-4;摘除 P0 → 详情+看板卡片同步变 冒烟/回归;列表/建新/挂/摘全绿。
- 测试数据按约定保留。

## Blast radius(模块外)
- `ListParams.keyword` 为共享字段,已确认 5 个下游 list 端点仍编译且行为不变(仅 issue 停用)。
- `labelApi` / `searchIssues` 为新增,无既有消费者。
- 整包 `tsc` clean。

Closes(能力对齐追踪,无单一 issue)。

### DIFF
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -25,11 +25,15 @@ async function enrich(issues: Issue[]): Promise<Issue[]> {
   }));
 }
 
-export async function listIssues(
-  params?: ListParams,
-): Promise<{ issues: Issue[]; total: number }> {
-  const data = await httpGet<{ issues: Issue[]; total?: number }>("/issues", {
-    keyword: params?.keyword,
+// 拉取 issue 列表并 enrich + 兜底 total(listIssues/searchIssues 共用尾巴)。
+async function fetchIssues(path: string, query: Record<string, unknown>): Promise<{ issues: Issue[]; total: number }> {
+  const data = await httpGet<{ issues: Issue[]; total?: number }>(path, query);
+  const issues = await enrich(data.issues ?? []);
+  return { issues, total: data.total ?? issues.length };
+}
+
+export function listIssues(params?: ListParams): Promise<{ issues: Issue[]; total: number }> {
+  return fetchIssues("/issues", {
     status: params?.status,
     priority: params?.priority,
     assignee_id: params?.assignee_id,
@@ -43,8 +47,20 @@ export async function listIssues(
     limit: params?.limit,
     offset: params?.offset,
   });
-  const issues = await enrich(data.issues ?? []);
-  return { issues, total: data.total ?? issues.length };
+}
+
+// 关键词搜索(GET /issues/search):独立端点,后端全文搜标题/描述/评论并回高亮片段。
+// 与 listIssues 是两套语义(不吃状态/优先级等筛选、limit≤50),故单列一个函数。
+export function searchIssues(
+  q: string,
+  opts?: { limit?: number; offset?: number; includeClosed?: boolean },
+): Promise<{ issues: Issue[]; total: number }> {
+  return fetchIssues("/issues/search", {
+    q,
+    limit: opts?.limit,
+    offset: opts?.offset,
+    include_closed: opts?.includeClosed ? "true" : undefined,
+  });
 }
 
 export async function enrichIssue(issue: Issue): Promise<Issue> {
--- a/packages/dmloop/src/api/labelApi.ts
+++ b/packages/dmloop/src/api/labelApi.ts
@@ -1,0 +1,36 @@
+// @octo/loop — 标签 API（后端契约联调）。
+// 工作区级标签 CRUD(/labels) + issue 挂/摘标签(/issues/:id/labels)。
+import type { IssueLabel } from "./types";
+import { httpGet, httpPost, httpPut, httpDelete } from "./http";
+
+/* ---------- 工作区标签 CRUD ---------- */
+// 后端 list 端点返回 { labels: [...], total }(包裹,同 /issues);解包取 labels。
+export function listLabels(): Promise<IssueLabel[]> {
+  return httpGet<{ labels?: IssueLabel[] }>("/labels").then((r) => r?.labels ?? []);
+}
+
+export function createLabel(name: string, color: string): Promise<IssueLabel> {
+  return httpPost<IssueLabel>("/labels", { name, color });
+}
+
+export function updateLabel(id: string, req: { name?: string; color?: string }): Promise<IssueLabel> {
+  return httpPut<IssueLabel>(`/labels/${id}`, req);
+}
+
+export function deleteLabel(id: string): Promise<void> {
+  return httpDelete<void>(`/labels/${id}`);
+}
+
+/* ---------- issue ↔ 标签 ---------- */
+export function listIssueLabels(issueId: string): Promise<IssueLabel[]> {
+  return httpGet<{ labels?: IssueLabel[] }>(`/issues/${issueId}/labels`).then((r) => r?.labels ?? []);
+}
+
+// 后端 POST /issues/:id/labels 体为 { label_id }。
+export function attachLabel(issueId: string, labelId: string): Promise<void> {
+  return httpPost<void>(`/issues/${issueId}/labels`, { label_id: labelId });
+}
+
+export function detachLabel(issueId: string, labelId: string): Promise<void> {
+  return httpDelete<void>(`/issues/${issueId}/labels/${labelId}`);
+}
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -59,6 +59,8 @@ export type IssueDateField = (typeof ISSUE_DATE_FIELDS)[number];
 
 export interface ListParams {
   workspace_id?: string;
+  // 客户端关键词过滤:projectApi/skillApi/squadApi/agentApi/runtimeApi 共用此字段。
+  // issue 列表不再用它(关键词走 searchIssues → /issues/search),但其它 list 端点仍需,勿删。
   keyword?: string;
   status?: IssueStatus;
   priority?: IssuePriority;
@@ -120,6 +122,11 @@ export interface Issue {
   creator_avatar?: string | null;
   // 后端 list/detail 端点批量回填；其它端点(update/ws)不带 → 保持已有。
   labels?: IssueLabel[] | null;
+  // 搜索结果专属(GET /issues/search):命中来源 + 高亮片段。列表/详情端点不返回。
+  match_source?: string;
+  matched_snippet?: string | null;
+  matched_description_snippet?: string | null;
+  matched_comment_snippet?: string | null;
 }
 
 export interface CreateIssueReq {
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -243,6 +243,11 @@
     "progress": "Progress",
     "lead": "Lead"
   },
+  "label": {
+    "add": "Add labels",
+    "newPlaceholder": "New label",
+    "create": "Create label"
+  },
   "agent": {
     "runtime": "Runtime",
     "model": "Model",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -243,6 +243,11 @@
     "progress": "进度",
     "lead": "负责人"
   },
+  "label": {
+    "add": "添加标签",
+    "newPlaceholder": "新标签名",
+    "create": "创建标签"
+  },
   "agent": {
     "runtime": "运行环境",
     "model": "模型",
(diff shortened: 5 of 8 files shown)


OctoBoooot (approved):
## Review: keyword search + label editing (capability layer) (#635)

**Verdict:** Approve
Clean capability-layer PR: keyword search moved to the real `/issues/search` endpoint, and a guarded label-editing write path. Verified the new API layer, the write-UI's concurrency/error/XSS safety, and the search refactor.

### What I verified
- **Search refactor** (`issueApi.ts`): `fetchIssues` extracts the shared enrich+total tail; `listIssues` drops the dead `keyword` param (it was a no-op on `/issues`); `searchIssues` hits the independent `/issues/search` with `q`/limit/offset. `keyword` correctly retained on `ListParams` for the 5 sibling list endpoints that still use it (the blast-radius catch). Undefined params drop at the http layer.
- **`labelApi.ts`**: correct CRUD + attach/detach — proper verbs (`httpPost`/`httpPut`/`httpDelete`), correct bodies (`{label_id}`), and it unwraps the `{labels:[...]}` envelope with `?? []` (the adversarially-caught crash, fixed). Ids are server-issued and axios-encoded into paths — same pattern as the existing `${id}` endpoints, no injection surface.
- **`LabelEditor.tsx` write UI**:
  - Double-submit guarded: `busy` checked at the top of both `toggle` and `create`, set true, released in `finally` — concurrent attach/detach/create can't overlap.
  - Error-handled: both paths try/catch → `Toast.error` with fallback; `busy` always released.
  - XSS-safe: label `name` is a JSX text child (`{l.name}`, and `LabelChips` hex-validates color) — React-escaped, no `dangerouslySetInnerHTML`.
  - `create` trims + empty-guards the name; `onChanged()` re-fetches after each mutation; `loadAll()` on open; `stopPropagation` on the create row keeps the input click from toggling the dropdown.

### Note (non-blocking)
- The new-label default color `#8B5CF6` is a hardcoded literal, but it's a *default value* passed to the API (runtime data, not a CSS style literal) and the code explicitly marks it a placeholder pending the UI redesign's color picker. Not a §2 concern; fine to defer with the noted UI rework.

### Praise
- The write UI gets the two things this kind of multi-mutation dropdown usually gets wrong: a single `busy` gate serializing attach/detach/create (no double-fire), and `stopPropagation` on the inline-create row so typing doesn't collapse the menu. And the PR is candid that `keyword` was a silent no-op on `/issues` before — fixing a dead param, not just adding a feature.

Jerry-Xin (approved):
**feat(loop): keyword search + label editing** — First-time review @ `c94a5cf3`, base `feat/octo-loop`. Both features are clean: search uses a bound query param through the paged endpoint with no client-truncation trap and no XSS echo; label editing is a thin capability layer over server-authorized endpoints (auth headers + CSRF injected centrally), with the color-render sink already hardened by the strict hex gate from #632. **APPROVE.**

### ✅ Keyword search
- **Param handling (safe):** `searchIssues(kw, …)` → `fetchIssues("/issues/search", { q, limit, offset, include_closed })` → `httpGet` → `axios.get(path, { params: clean(params) })`. The keyword flows as an axios `params` entry (URL-encoded by axios), never interpolated into the path/query string. `clean()` stringifies and drops empty/nullish. No injection surface. (`issueApi.ts:55-68`, `http.ts` `httpGet`/`clean`)
- **Truncation/pagination (server-side, avoids the #608/#613/#627/#634 trap):** search is its own endpoint with real `limit`/`offset` (`limit: paged ? PAGE_SIZE : 50`), paged server-side — not a client-side filter over a `limit:100` board page. Confirmed the board branch keeps the historical `limit:100` full-column semantics only for the non-search list path. (`IssuePage.tsx:73-75`)
- **Stale-guard + page reset (correct):** keyword input goes through the same `update()` which does `setPage(0)`, and `reload` reuses the `seq.current` stale-guard (`if (my !== seq.current) return`). Empty keyword → `kw = f.keyword.trim()` is falsy → falls back to `listIssues` (no filter). Clean. (`IssuePage.tsx:64-66, 115-118, 245-248`)
- **XSS (none):** keyword is only bound to the controlled `<Input value={f.keyword}>`; never echoed into rendered text. The new `match_source`/`matched_*_snippet` fields on `Issue` are **defined in types only and never rendered** anywhere (grep-verified across dmloop/src). No `dangerouslySetInnerHTML`. No sink.

### ✅ Label editing (write path)
- **Authorization (server-enforced):** all label mutations (`createLabel` POST /labels, `attachLabel` POST /issues/:id/labels, `detachLabel` DELETE …) go through `httpPost`/`httpDelete`, whose shared axios interceptor injects `token` (WKApp.loginInfo.token), `X-Space-Id`, `x-workspace-slug`, and `X-CSRF-Token` on every non-GET. The client does not self-authorize — the backend scopes the mutation to the caller's workspace/session. Capability is not a client-only gate. (`labelApi.ts`, `http.ts` interceptor)
- **Color validation (defense-in-depth intact):** new labels persist a hardcoded literal `"#8B5CF6"` (`LabelEditor.tsx:44`) — no arbitrary user color reaches the persist path in this PR (color picker deferred by design). And the render sink is already guarded: `LabelChips` gates on `/^#[0-9a-fA-F]{6}$/.test(l.color)` before ever interpolating `l.color` into `style.backgroundColor/color/border`, falling back to default Tag styling otherwise. So even a bad server-side color cannot inject into the style. (`LabelChips.tsx` hex gate, matching #632)
- **Name (safe):** `newName` is `.trim()`-guarded (empty rejected), persisted via JSON body, and rendered only as React text content in `LabelChips` (`{l.name}`) — auto-escaped, never interpolated into a query or style. No XSS.
- **Fail-safe (correct):** `toggle`/`create` are `try/catch` with `Toast.error((e as Error)?.message ?? saveFailed)` — write failures surface to the user, not swallowed. State refresh (`onChanged → syncIssue`) only runs after the awaited call resolves, so there is no optimistic phantom to roll back. `busy` guards against concurrent double-submits. (`LabelEditor.tsx:26-58`)

### 💬 Non-blocking notes
- `toggle`/`create`/`loadAll` don't guard against setState-after-unmount (the Dropdown/Input can close mid-request). Low risk here (fast calls, small component) but a mounted-ref or the `seq`-style guard used in `IssuePage` would be tidier.
- i18n: `label.add/newPlaceholder/create` added to both en-US and zh-CN. ✅ Toast fallback reuses existing `loop.toast.saveFailed`.

### ✅ Highlights
- Search correctly modeled as a distinct endpoint with its own semantics rather than overloading `listIssues` (the retained `ListParams.keyword` comment explains why other list APIs still need it — prevents an accidental delete).
- Reuses the existing stale-guard + page-reset machinery instead of reinventing it.
- Label render hardening from #632 makes the write path safe-by-construction even before a color picker lands.

yujiawei (approved):
# Code Review — PR #635 (octo-web)

**Scope:** capability-layer PR on `feat/octo-loop` — adds keyword full-text search (`searchIssues` → `GET /issues/search`) and label editing (`labelApi` + `LabelEditor`). 8 files, +193/-27. UI is intentionally minimal.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Both advertised capabilities land — `searchIssues()` wired into `IssuePage` when a keyword is present, and full label CRUD/attach/detach in `labelApi` + a `LabelEditor` in the detail page.
- **Over-build:** none. No unrelated flags/endpoints. The four `match_source`/`matched_*_snippet` fields on `Issue` are added but not yet rendered — acceptable as a type-ahead of the deferred UI, but see P2 below.
- **Deviation:** none that contradicts the stated design. Search deliberately runs with independent semantics (no combined filters, backend cap ≤ 50); this is documented in code and matches the backend contract, so it is intended scope, not a spec violation.

## 2. Code quality

**Quality: Approved** (no P0/P1; P2 suggestions below)

### P2 — Active filters are silently ignored while a keyword is present
`packages/dmloop/src/pages/IssuePage.tsx:72-89`

When `kw` is non-empty the code branches to `searchIssues(kw, …)` and drops `status`, `priority`, `assignee`, `creator`, `project`, `dateRange`, and `sort`. That matches the search endpoint's contract, **but** the filter controls (status/assignee/project selects, date range, sort) remain visibly populated in the toolbar and are silently bypassed. A user who has scoped to, say, assignee=X and then types a keyword will get results outside that scope with no indication their filter was dropped.

Not a security/isolation issue — workspace scoping is enforced server-side via the always-injected `x-workspace-slug` header + auth token (`api/http.ts:64`), independent of `project_id`. It is a UX/consistency defect. Suggestion: when a keyword is active, disable or visually mute the non-applicable controls, or show a short hint ("Search ignores filters"). Fast-follow is fine given this is a minimal-UI capability PR.

### P2 — Search cannot reach closed issues; `includeClosed` never wired
`packages/dmloop/src/pages/IssuePage.tsx:74`, `api/issueApi.ts:54-63`

`searchIssues` exposes `includeClosed`, but `IssuePage` calls it with only `{limit, offset}`. Combined with the dropped `status` filter, search always uses the backend default for closed inclusion — if that default excludes closed issues, a user searching while intending to see closed/all items silently loses them. Consider mapping the status intent (or an explicit "include closed" affordance) into the search call.

### P2 — `busy` lock releases before the parent refetch lands (minor stale-state race)
`packages/dmloop/src/ui/LabelEditor.tsx:31-45`, `panel/IssueDetailPage.tsx:129`

`toggle()` sets `busy=false` in `finally` immediately after `onChanged()`, but `onChanged` triggers the parent's async `syncIssue()`. Between lock release and the refetch resolving, `attached` (derived from the `labels` prop) is stale. A fast second click on the same label evaluates against the stale set and can fire a duplicate `attachLabel`/`detachLabel`, surfacing a needless error toast if the backend rejects the duplicate. Low-severity (requires a rapid re-click inside the refetch window; no data loss). Optional hardening: keep `busy` until the parent confirms, or derive `attached` from a locally reconciled set.

### P2 — Search highlight fields are dead
`packages/dmloop/src/api/types.ts:126-129`

`match_source` / `matched_snippet` / `matched_description_snippet` / `matched_comment_snippet` are added to the `Issue` type but never rendered anywhere; `IssueList`/`IssueBoard` still key off `labels` only. The backend's highlight snippets are the differentiating value of the search endpoint — surfacing at least one snippet would make search results meaningfully better than the plain list. Acknowledged as deferred UI; flagging so it isn't forgotten.

### Nits (non-blocking)
- `createLabel` uses a hard-coded color `#8B5CF6` (`LabelEditor.tsx:39`) — fine as a documented placeholder pending the color picker.
- `create()` — if `createLabel` succeeds but `attachLabel` throws, an orphan (created-but-unattached) label remains. Acceptable for now; worth a follow-up cleanup path.

### Verified correct
- `fetchIssues` shared tail (enrich + `total` fallback) is a clean, safe refactor; `listIssues` behavior is unchanged.
- `ListParams.keyword` correctly retained — 5 other list endpoints (project/skill/squad/agent/runtime) still consume it for client-side filtering; only issue list stopped sending it.
- `listLabels` / `listIssueLabels` unwrap the `{labels: […]}` envelope, consistent with `/issues`.
- i18n keys (`label.*`, `toast.saveFailed`) exist and resolve under the loop namespace in both locales.
- Stale-response guard (`seq.current`) and page-clamp logic preserved across the search/list branch.
- The gemini-flagged `include_closed: "true"` string is a non-issue: the HTTP layer `String()`-casts every query param, so all query values go over the wire as strings regardless.

## 3. Overall verdict

**APPROVED.** No correctness, security, data-loss, or build-blocking issues. The findings are P2 UX/consistency improvements suitable for a fast follow, appropriate to a capability-layer PR that explicitly defers UI polish.

## 4. Coverage / could not verify
- Backend contracts for `/issues/search` (filter support, closed-inclusion default, 50-cap, snippet shape) and `/labels` + `/issues/:id/labels` (attach body `{label_id}`, detach path param, idempotency on duplicate attach) — verified against the documented client-side expectations only, not the server.
- No type-check/build was run in this environment (workspace deps not installed); the diff is type-consistent on inspection and the author reports a clean `tsc`.

PR #636, opened 2026-07-10
feat(loop): issue subscription + comment reactions (capability layer)

MrAladdin:
## What

第二枪「能力对齐」PR:issue 订阅 + 评论 emoji 反应。UI 保持最小(dmloop UI 后续会重做,当下把后端协作端点接进 API 层才是不会白做的部分)。

### 订阅
- 新增 `collabApi`:`listSubscribers` + `subscribe`/`unsubscribe`(**都是 POST** —— 后端路由是 `POST /issues/:id/(un)subscribe`,不是 DELETE;body 可空则默认操作调用者)。
- 详情页 more 菜单加「订阅 / 取消订阅」两个**幂等**项(而非 toggle:前端无「当前成员身份」可判断「我是否已订阅」;后端 add 是 upsert、remove 是按 actor 定位的 no-op,两项都安全)。侧栏显示订阅者数。

### 评论 emoji 反应
- `addCommentReaction`(POST)/ `removeCommentReaction`(DELETE 带 `{emoji}` body),端点 `/comments/:id/reactions`。`IssueComment` 加可选 `reactions[]`(list 端点 `ListComments` 已 groupReactions 回填,**不改数据源**)。
- 每条评论加反应条:按 emoji 分组的「emoji×计数」chip(点 chip 删自己那条)+ 一个「+」选择器从固定小集合加。

## 明确延后(避免白做 UI)
- **timeline / 活动流**、**issue 级反应** —— 它们的渲染表面正是 UI 重做要定的东西;订阅 toggle 同样等「当前成员」查询就位后再做。

## 审查(全绿)
- **/simplify**(4 agent):抽出 `renderReactions` helper 去掉 JSX 内 IIFE;reuse/efficiency/altitude 均 clean。
- **双模型对抗审查**:Claude code-reviewer + `codex:adversarial-review` 均 clean —— 端点 method/path 对齐 `octo-multica/server/cmd/server/router.go`(subscribe POST、reaction 删除 DELETE 带 body)、reactions 经 `ListComments` 往返、`IssueComment.reactions` 为可选故 blast-radius 受控(全包仅 3 文件引用 `IssueComment`)。
- **真实浏览器 e2e**(dev / Alice_Space / MV2 E2E):订阅者 3→4、取消订阅 no-op(Alice 不在原 3 人内);评论反应 加 🎉(chip 出现)/ 删(chip 消失);列表/详情同步。测试数据按约定保留。

## Blast radius(模块外)
[...]

PR #637, opened 2026-07-10
feat(loop): sub-issues + issue field editing (capability layer)

MrAladdin:
## What

第三枪「能力对齐」PR:子 issue + issue 字段编辑。UI 保持最小(dmloop UI 后续会重做,当下把字段编辑 + 子 issue 端点接进 API 层才是不会白做的部分)。

### 字段编辑(详情页侧栏)
- `UpdateIssueReq` 扩 `start_date`/`due_date`/`parent_issue_id`/`stage`。后端按 `rawFields` 存在性门控、`JSON.stringify` 丢 `undefined`,故 `patch()` 只发被改字段(省略=不动、`""`=清除)。
- start/due:只读文本 → 可编辑 `DatePicker`(yyyy-MM-dd);父 Issue:懒加载工作区 issue 的 `Select`(后端做环检测);stage:`InputNumber`。

### 子 issue
- `listChildren()` → `GET /issues/:id/children`;只读「子 Issue」区列出子项 + `done/total` 进度条(进度由子项 status 本地算,详情页无需再调批量 `/child-progress`)。

## 对抗审查发现并已修(全部提交前,双模型)
- **patch 无错误反馈** → 加 try/catch + Toast,后端拒绝(父环检测、非法日期)不再静默失败;这是所有字段编辑的共享漏斗,status/priority/assignee 也一并获得反馈;失败时 issue state 不变,受控输入自动回退。(Claude)
- **stage InputNumber 退格清空触发 spurious `patch({stage:null})`** → 改为仅在有效数字时提交。(Claude)
- **导航竞态**:reload 重置 children/subscribers/parentCands + `reqRef` request-token 守卫所有异步 setter,慢请求(issue A)不会在切到 B 后把 A 的数据写进 B(子区未来引出「点子项跳转」正会触发)。(codex,两轮)

## /simplify(4 agent)
reuse/simplification/efficiency/altitude 全 clean;采纳一条:给 100 条父候选上限加 `// ponytail:` 注释。altitude agent 评价共享 patch 的错误处理是「exemplary sink」(根因层)。

## 真实浏览器 e2e(dev / Alice_Space / MV2 E2E)
- 字段编辑:stage 落库回显、截止日期显示/编辑、父 Issue 选择器候选懒加载(排除自己)。
- 子 issue 闭环:设 MVE-1 父=MVE-4 → MVE-4「子 Issue (0/1)」显示 MVE-1(审核中,进度 0/1 正确)。
- token 守卫复验不破坏 happy path(children/subscribers/stage/日期全正常填充,无错误 toast)。
- 测试数据按约定保留。

## Blast radius(模块外)
- `UpdateIssueReq` 全为可选新增字段;IssueList/IssueBoard 等其它 `updateIssue` 调用者只发各自字段,`JSON.stringify` 丢 undefined → 与后端 rawFields 门控一致,无破坏。
- `listChildren` 复用既有 `enrich`;无新消费者破坏。整包 `tsc --skipLibCheck` 仅 2 处既有报错(非本 PR),PR-C 文件零错误。

基于最新 `feat/octo-loop`(#636 已合入)。

### DIFF
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -72,6 +72,13 @@ export async function getIssue(id: string): Promise<Issue> {
   return enrichIssue(issue);
 }
 
+// 子 issue 列表(GET /issues/:id/children):后端包裹 { issues }。子项含 status,
+// 进度(done/total)由页面本地算,不再调批量 /child-progress(那是列表/看板用的)。
+export async function listChildren(id: string): Promise<Issue[]> {
+  const data = await httpGet<{ issues?: Issue[] }>(`/issues/${id}/children`);
+  return enrich(data?.issues ?? []);
+}
+
 export function createIssue(req: CreateIssueReq): Promise<Issue> {
   return httpPost<Issue>("/issues", req);
 }
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -147,6 +147,13 @@ export interface UpdateIssueReq {
   assignee_id?: string | null;
   project_id?: string | null;
   position?: number;
+  // 后端按 rawFields 存在性判断：省略=不动；传值=设置；传 ""/null=清除。
+  // start_date/due_date 为日历日 "YYYY-MM-DD"(后端 ParseCalendarDate)。
+  // parent_issue_id 设置时后端做环检测(不能设自己或后代为父)。stage 需 >= 1。
+  start_date?: string | null;
+  due_date?: string | null;
+  parent_issue_id?: string | null;
+  stage?: number | null;
   // 指派/状态变更触发 agent run 时：suppress_run=true 表示“暂不开始”；handoff_note 仅在真起 run 时消费。
   suppress_run?: boolean;
   handoff_note?: string;
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -137,7 +137,10 @@
     "creator": "Creator",
     "labels": "Labels",
     "startDate": "Start date",
-    "dueDate": "Due date"
+    "dueDate": "Due date",
+    "parent": "Parent issue",
+    "noParent": "No parent",
+    "stage": "Stage"
   },
   "status": {
     "backlog": "Backlog",
@@ -254,6 +257,9 @@
     "unsubscribe": "Unsubscribe",
     "subscribed": "Subscribed",
     "unsubscribed": "Unsubscribed"
+  },
+  "subIssue": {
+    "title": "Sub-issues"
   },
   "reaction": {
     "add": "Add reaction"
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -137,7 +137,10 @@
     "creator": "创建者",
     "labels": "标签",
     "startDate": "开始日期",
-    "dueDate": "截止日期"
+    "dueDate": "截止日期",
+    "parent": "父 Issue",
+    "noParent": "无父 Issue",
+    "stage": "阶段"
   },
   "status": {
     "backlog": "待规划",
@@ -254,6 +257,9 @@
     "unsubscribe": "取消订阅",
     "subscribed": "已订阅",
     "unsubscribed": "已取消订阅"
+  },
+  "subIssue": {
+    "title": "子 Issue"
   },
   "reaction": {
     "add": "添加反应"
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import {
   Typography,
   Input,
@@ -11,6 +11,9 @@ import {
   Popconfirm,
   TextArea,
   Dropdown,
+  DatePicker,
+  InputNumber,
+  Progress,
 } from "@douyinfe/semi-ui";
 import {
   ArrowLeft,
@@ -43,6 +46,8 @@ import {
   enrichIssue,
   deleteIssue,
   listComments,
+  listChildren,
+  listIssues,
   addComment,
   deleteComment,
   updateComment,
@@ -70,7 +75,6 @@ import {
   PRIORITY_COLOR,
   RUN_STATUS_COLOR,
   isActiveRun,
-  isOverdue,
 } from "../ui/meta";
 import "./issueDetail.css";
 
@@ -99,6 +103,8 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
   const [issue, setIssue] = useState<Issue | null>(null);
   const [comments, setComments] = useState<IssueComment[]>([]);
   const [subscribers, setSubscribers] = useState<IssueSubscriber[]>([]);
+  const [children, setChildren] = useState<Issue[]>([]);
+  const [parentCands, setParentCands] = useState<Issue[]>([]); // 父 issue 选择器候选(懒加载)
   const [runs, setRuns] = useState<TaskRun[]>([]);
   const [activeRun, setActiveRun] = useState<TaskRun | null>(null);
   const [runOpen, setRunOpen] = useState(false);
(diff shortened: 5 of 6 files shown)


OctoBoooot (approved):
## Review: sub-issues + issue field editing (capability layer) (#637)

**Verdict:** Approve
Clean capability-layer PR with genuinely strong defensive engineering: a request-token guard across every async setter, a shared `patch()` error funnel with auto-revert, and correct field-clear / stage-edge handling. Verified each against the code.

### What I verified
- **Navigation-race guard (`reqRef`)** — `reload` does `const token = ++reqRef.current`, and **every** async setter (issue/comments/runs, plus the bypath subscribers/children, and `loadParentCands`) checks `token === reqRef.current` before writing state. So an in-flight response for issue A can't land in issue B's view after an in-place `issueId` switch. `reload` also resets `children`/`subscribers`/`parentCands` up front to avoid stale carry-over. This is exactly the stale-response class of bug done right.
- **Shared `patch()` error funnel** — now try/catch → `Toast.error` with fallback; on failure `setIssue` is not called, so the controlled inputs (DatePicker/Select/InputNumber bound to `issue.*`) auto-revert. One funnel gives every field editor (status/priority/assignee/date/parent/stage) error feedback + revert — the "exemplary sink" the /simplify altitude pass called out.
- **Field-clear semantics** — `parent_issue_id`/`start_date`/`due_date` send `""` on clear vs omitted for no-change, matching the backend's rawFields-existence gating (only changed fields sent, `JSON.stringify` drops undefined). Correct per the documented contract.
- **Stage spurious-null fix** — `onChange={(v) => { if (typeof v === "number") patch({stage:v}) }}` ignores the empty value backspace produces, so no accidental `patch({stage:null})`.
- **Sub-issues** — `listChildren` unwraps `{issues}` + enriches; progress `childrenDone/children.length` computed locally from child status (no extra `/child-progress` call), gated on `length>0`; identifier/title rendered as escaped JSX text. Parent cycle-detection correctly deferred to the backend; candidate list excludes self.

### Praise
- The `reqRef` token is applied to *all* async setters, not just the primary fetch — that's the version of this guard that actually holds when a detail view can switch issue in place (the future "click a sub-issue to navigate" path the body anticipates). Combined with the shared `patch()` funnel that leaves state untouched on failure so controlled inputs self-revert, this is the right altitude for both concerns — root-cause sinks, not per-field patches. The three adversarially-caught bugs (silent patch failure, stage-clear spurious null, nav race) are all genuinely closed.

Jerry-Xin (approved):
**APPROVE** — capability-layer PR adding sub-issue display + inline issue field editing. Both write paths are API-correct, XSS-safe, and fail-safe; hierarchy cycle risk is server-enforced with the rejection now surfaced. Two non-blocking notes below, consistent with the guard grading on the sibling PRs in this family.

### ✅ Highlights
- **API correctness**: `GET /issues/:id/children` (`issueApi.ts` `listChildren`) and `PUT /issues/:id` (`updateIssue`) use correct verbs/bodies; server-issued id in the path, untrusted values in the body — no URL concatenation. Envelope unwrapped with the null-safe fallback (`data?.issues ?? []`), matching the pattern from the earlier null-crash fix.
- **Cycle / self-parent guard**: parent selector excludes `self` client-side (`parentCands.filter(i => i.id !== issueId)`), and the real guard is server-enforced (documented in `types.ts`: "设置时后端做环检测(不能设自己或后代为父)"). Defense-in-depth done right; rejection is now surfaced via Toast rather than swallowed.
- **Error handling**: `patch()` wraps `updateIssue` in try/catch and surfaces backend rejections (bad date, cycle) via `Toast.error(... ?? loop.toast.saveFailed)` — no silent failure. `saveFailed` key present in both locales.
- **Field validation**: `stage` `InputNumber` has `min={1}` and only submits on a real number (empty/backspace ignored, avoiding stray `stage=null`); `start_date`/`due_date` constrained to `yyyy-MM-dd` via `DatePicker` and validated server-side (`ParseCalendarDate`). All editable fields are typed widgets, not free-text — no arbitrary status/priority injection here.
- **XSS-safe**: child `identifier`/`title` rendered as escaped React text; no `dangerouslySetInnerHTML`; no field value interpolated into style/URL/query.
- **Navigation race handled well**: new `reqRef` token discards stale async responses after an in-place `issueId` switch, and resets `children`/`subscribers`/`parentCands` on reload — a solid fix that goes beyond the sibling PRs.
- **i18n**: `parent`/`noParent`/`stage`/`subIssue.title` added to both `en-US` and `zh-CN`.

### 🟡 Non-blocking
- `IssueDetailPage.tsx` — the field editors (parent `Select`, start/due `DatePicker`, stage `InputNumber`) call `patch()` on `onChange` with **no in-flight/disabled guard**. Rapid sequential edits can issue overlapping PUTs (last-write-wins; error toasts may arrive out of order). Writes are idempotent and the backend is authoritative, so this is not a data-integrity hole — but it is the same busy/stale-window class flagged 🟡 on the sibling capability PRs; grading it identically. A per-field disable-while-saving (or refetch-on-settle) would close it.
- `IssueDetailPage.tsx` — `listChildren(issueId).catch(() => {})` silently hides a sub-issue load failure (issue with children could render as having none). Matches the existing `listSubscribers` bypass convention, so acceptable, but a section-level error state or toast would be friendlier.

Nothing blocking. Ship it.

yujiawei (approved):
# Code Review — PR #637 (octo-web)

**Scope reviewed:** head `610017e4`, base `feat/octo-loop`, merge-base `7536ddb7`. 6 files, +155/−20. Capability layer for sub-issue display + issue field editing (parent / start / due / stage) in the dmloop detail panel.

Verdict: **APPROVED**. Solid, well-guarded change. Findings below are all advisory (P2) — none block merge, consistent with the PR's stated "minimal UI, to be redone" framing.

## 1. Spec compliance

**Compliant.** The diff implements exactly what the PR describes and nothing extra:
- `listChildren()` → `GET /issues/:id/children`, rendered as a read-only "Sub-issues (done/total)" section with a local progress bar (`IssueDetailPage.tsx` sub-issue block; `issueApi.ts:76`).
- `UpdateIssueReq` extended with `start_date` / `due_date` / `parent_issue_id` / `stage` (`types.ts:153-156`), all optional so other `updateIssue` callers are unaffected.
- Field editors wired via the shared `patch()` funnel: parent `Select` (lazy candidates), start/due `DatePicker`, stage `InputNumber`.
- No missing, extra, or divergent behavior versus the description.

## 2. Code quality

Strong points worth calling out: the `reqRef` request-token guard against the navigation race is the right pattern and correctly resets `children`/`subscribers`/`parentCands` on `issueId` change; moving error handling into the shared `patch()` funnel means status/priority/assignee edits all get failure feedback too; the frontend correctly delegates cycle detection and date/stage validation to the backend rather than trusting itself.

### P2 — advisory

**P2-1 · Parent field renders a raw UUID until the dropdown is opened.** `IssueDetailPage.tsx` — the parent `Select` has `value={issue.parent_issue_id ?? undefined}`, but options (`parentCands`) load lazily only on `onDropdownVisibleChange`. For any issue that already has a parent, the initial render has a value with no matching `Select.Option`, so Semi shows the raw UUID as the label instead of `IDENTIFIER Title`. It only resolves after the user opens the dropdown — and only if that parent is within the first 100 fetched issues. Suggest seeding the option for the current parent (e.g. fetch/derive the parent issue on load, or inject a single synthetic option from the current parent's identifier).

**P2-2 · `stage` fires one PUT per keystroke, and `patch()` has no race guard.** `IssueDetailPage.tsx` — `InputNumber onChange` calls `patch({ stage: v })` on every valid numeric change, so typing "12" sends `patch(1)` then `patch(12)`, each a full `updateIssue` PUT + `enrichIssue` + `onChanged()`. Unlike `reload()`, `patch()` is not token-guarded, so out-of-order PUT responses can write stale `issue` state via `setIssue`. Low impact for small stage values, but consider committing on blur/enter (or debouncing) and applying the same `fresh()`/token guard the PR introduced for `reload()`.

**P2-3 · `patch()` shares the navigation-race exposure the PR set out to fix.** The token guard was added to `reload()`'s setters but not to `patch()`. Editing a field and navigating to another issue before the PUT resolves would apply issue A's response to issue B's view (`setIssue` runs unconditionally against a closed-over `issue`). Narrow window, but it's the same failure class the PR explicitly hardened elsewhere — worth closing for symmetry.

**P2-4 (nit) · Lost overdue indicator.** The due date was previously rendered with `isOverdue()` red styling; the new `DatePicker` drops that visual cue. Acceptable for a capability layer, noting it as a minor UX regression.

## 3. Security note (PR was auto-classified security-sensitive)

The classification appears to be a keyword false-positive: this diff is issue-field-editing UI with no auth, token, secret, credential, or middleware logic. Trust boundary is clean — the frontend does not self-validate parent cycles or dates; it delegates to the backend and surfaces rejections via `Toast`. Nothing here needs manual security verification beyond confirming the backend `PUT /issues/:id` enforces the `rawFields`-gated cycle detection / calendar-date parse / `stage >= 1` contract the type comments assume (see coverage below).

## 4. Coverage / blind spots (verify outside this diff)

- **Backend contract is assumed, not provable here.** The whole "omit=no-op, `""`/`null`=clear, `stage>=1`, cycle detection, `ParseCalendarDate`" behavior lives in the backend `PUT /issues/:id`. This diff is correct *if* that contract holds. Recommend a human confirm the backend honors rawFields-presence gating and rejects self/descendant parents.
- **`GET /issues/:id/children` shape** (`{ issues }` wrapper, children carrying `status`) is assumed; not verified against the server in this diff.
- **No tests** touch `listChildren` / field editing (no test files in `packages/dmloop`). E2E was done manually per the PR body; no automated regression guard added.
- **Type-check not run here** (no installed `node_modules` in the review checkout); relying on the author's reported `tsc` result (PR-C files clean).

### Reviewer transparency

This review was produced by a single reviewer. Two supplementary independent analysis passes were attempted but were unavailable this run (one hit a persistent gateway tool-config error, the other exceeded its time budget), so their input is absent rather than a silent agreement — findings above are the sole reviewer's own.

PR #638, opened 2026-07-10
fix: preserve adaptive card table rendering in production

an9xyz:
## Summary
- Explicitly register AdaptiveCards Table in the octo serialization context instead of relying on the package global registration side effect.
- Add a regression test that unregisters the global Table entry and verifies the octo context still supports Table.

## Why
Production build can tree-shake AdaptiveCards table registration because the package declares sideEffects:false. Dev server can still render Table, while build output may drop the side-effect registration and render cards without the table section.

## Verification
- pnpm exec vitest run packages/dmworkbase/src/Messages/InteractiveCard/__tests__
- git diff --check
- pnpm --filter ./apps/web build

### DIFF
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/octoSdkConfig.test.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/octoSdkConfig.test.ts
@@ -4,7 +4,7 @@
 // HostConfig 颜色经可注入解析器映射自 --wk-* token。
 
 import { describe, expect, it, vi } from "vitest";
-import { FontType, GlobalRegistry } from "adaptivecards";
+import { FontType, GlobalRegistry, Table, Versions } from "adaptivecards";
 import { createOctoSerializationContext } from "../sdk/octoSerialization";
 import { buildOctoHostConfig } from "../sdk/octoHostConfig";
 
@@ -57,6 +57,7 @@ describe("createOctoSerializationContext — 元素层防御纵深", () => {
       const ctx = createOctoSerializationContext();
       expect(ctx.elementRegistry.findByName("Table")).toBeDefined();
     } finally {
+      GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5);
       GlobalRegistry.reset();
     }
   });
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/octoSdkConfig.test.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/octoSdkConfig.test.ts
@@ -4,7 +4,7 @@
 // HostConfig 颜色经可注入解析器映射自 --wk-* token。
 
 import { describe, expect, it, vi } from "vitest";
-import { FontType } from "adaptivecards";
+import { FontType, GlobalRegistry } from "adaptivecards";
 import { createOctoSerializationContext } from "../sdk/octoSerialization";
 import { buildOctoHostConfig } from "../sdk/octoHostConfig";
 
@@ -48,6 +48,16 @@ describe("createOctoSerializationContext — 元素层防御纵深", () => {
     // 非白名单元素被移除。
     for (const t of ["Carousel", "Media"]) {
       expect(reg.findByName(t)).toBeUndefined();
+    }
+  });
+
+  it("Table 不依赖 adaptivecards 全局注册副作用，production tree-shaking 后仍可用", () => {
+    GlobalRegistry.defaultElements.unregister("Table");
+    try {
+      const ctx = createOctoSerializationContext();
+      expect(ctx.elementRegistry.findByName("Table")).toBeDefined();
+    } finally {
+      GlobalRegistry.reset();
     }
   });
 });
--- a/packages/dmworkbase/src/Messages/InteractiveCard/sdk/octoSerialization.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/sdk/octoSerialization.ts
@@ -2,6 +2,7 @@ import {
   CardObjectRegistry,
   GlobalRegistry,
   SerializationContext,
+  Table,
   Versions,
   type Action,
   type CardElement,
@@ -50,6 +51,9 @@ export function createOctoSerializationContext(): SerializationContext {
 
   const elementRegistry = new CardObjectRegistry<CardElement>();
   GlobalRegistry.populateWithDefaultElements(elementRegistry);
+  // adaptivecards 声明了 sideEffects:false；production build 可能剪掉 lib/table.js
+  // 的默认注册副作用。Table 是 octo 白名单元素，需显式注册，不能依赖全局 registry。
+  elementRegistry.register("Table", Table, Versions.v1_5);
   for (const type of FORBIDDEN_ELEMENTS) {
     elementRegistry.unregister(type);
   }


OctoBoooot (dismissed):
## Review: preserve adaptive card table rendering in production (#638)

**Verdict:** Approve
Correct fix for a real production-only defect: AdaptiveCards declares `sideEffects:false`, so a prod tree-shaker can drop the global `Table` registration side-effect that dev keeps, leaving cards rendered without their table section. Explicitly registering `Table` in the octo context makes it deterministic. Verified the fix and its test.

Risk tier: normal. Base `main`; `mergeStateStatus=BLOCKED` (review required) — approved on code, not yet mergeable.

### What I verified
- **Root cause is real & the fix targets it:** `populateWithDefaultElements` copies from the package global registry, which is populated by a module side-effect that `sideEffects:false` lets production tree-shake away → no Table in the context → table section silently missing (dev works because it doesn't shake as aggressively). Importing `Table` from the package and `elementRegistry.register("Table", Table, Versions.v1_5)` both retains the class (defeats tree-shaking) and registers it deterministically. Correct.
- **Version:** `Versions.v1_5` matches Table's AC 1.5 introduction and the codebase's existing Table-at-1.5 handling (`validateCardForOcto`).
- **Ordering:** register runs after `populateWithDefaultElements` and before the `FORBIDDEN_ELEMENTS` unregister loop; Table isn't in FORBIDDEN (Carousel/Media) and *is* an octo-whitelisted element (validate has a `case "Table"`), so this matches the whitelist — no surface widening.
- **Test is properly adversarial:** it `GlobalRegistry.defaultElements.unregister("Table")` to *simulate* the tree-shaken prod state, then asserts the octo context still has Table — so it fails on the old (global-reliant) code and passes with the explicit register. `try/finally` + `GlobalRegistry.reset()` prevents cross-test pollution. This is the right way to pin a tree-shaking bug the test env can't otherwise reproduce.

### Praise
- The test simulates the production failure mode (unregister the global, then assert the context is self-sufficient) rather than just asserting the happy path — so it actually guards the regression it fixes, and would fail if someone later reverts to relying on the side-effect. Exactly the test a "works in dev, breaks in prod build" bug needs.

Jerry-Xin (dismissed):
**APPROVE** — correct, minimal, prod-surviving fix for the adaptive-card Table dropping out in production builds.

## Root cause (verified)
`adaptivecards@3.0.6` declares `sideEffects: false` in its package.json. The default `Table` registration lives as a **module-top-level side-effect** in `lib/table.js:466`:
```js
GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5);
```
`octoSerialization.ts` imported only the `Table` class and relied on `GlobalRegistry.populateWithDefaultElements()` copying Table from the global default registry. Under a production bundler, with `sideEffects:false` the bundler is free to prune that unreferenced side-effect statement (or treat `table.js` as side-effect-free), so `GlobalRegistry.defaultElements` no longer contains `Table` → the copy yields a context without Table → table sections silently vanish. Dev server does no tree-shaking, so it renders fine. This exactly matches the reported "works in dev, dropped in prod" symptom.

## Why the fix is correct
- `octoSerialization.ts:54` now calls `elementRegistry.register("Table", Table, Versions.v1_5)` on the **local** registry after `populateWithDefaultElements`. This no longer depends on the pruned global side-effect, so it survives a production build regardless of tree-shaking.
- **No double-register hazard.** `CardObjectRegistry.register` (lib/registry.js:19) is idempotent: if the name already exists it updates the existing entry in place under the same key — it does not create a duplicate. So if the global side-effect happened to survive, the explicit re-register is a same-class/same-version no-op; if it was pruned, it correctly adds Table. One "Table" entry either way.
- **Zero blast radius.** The registration is applied only to the freshly-constructed local `elementRegistry`, never to `GlobalRegistry` — no shared/global state is mutated, no other consumer is affected. The `Table` class was already pulled in to render tables, so bundle-size cost is negligible (only a one-line side-effect was being dropped, not the class); no heavy eager load is introduced.
- **Consistent with existing pattern.** This mirrors the action-registry idiom already in the same file (`actionRegistry.register(CopyToClipboardAction.JsonTypeName, ...)` after `populateWithDefaultActions`) — populate-then-adjust. Table warrants explicit pinning precisely because its default registration reached the registry solely via the side-effect path.

## Test
The regression test faithfully simulates the prod state: `GlobalRegistry.defaultElements.unregister("Table")`, asserts `createOctoSerializationContext().elementRegistry.findByName("Table")` is still defined, and restores via `GlobalRegistry.reset()` in `finally` to prevent cross-test pollution. Good guard against reintroducing the global-side-effect dependency.

## Notes (non-blocking)
- CI `check-sprint` is failing and `code-review` is pending — those are process/metadata gates, unrelated to this change's correctness.

yujiawei (dismissed):
# Code Review — PR #638 (octo-web)

**Head SHA:** `f0453cac1cdfa0fbbb8ed942ebf29a6296ff346f`

## Summary

The fix is correct and well root-caused. `adaptivecards@3.0.6` declares `sideEffects: false` in its `package.json`, and `Table` is registered exactly once, as a module-level side effect at the bottom of `lib/table.js`:

```js
GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5);
```

A production bundler that honors `sideEffects: false` is free to drop `lib/table.js` if nothing imports a binding from it, which silently strips that registration. `createOctoSerializationContext()` builds its element registry with `GlobalRegistry.populateWithDefaultElements(...)`, so a missing default registration means `Table` never lands in the octo registry and table cards fail to deserialize in production while working fine in dev. Importing `Table` and registering it explicitly is the right, minimal fix.

## Spec compliance

The PR does exactly what the title promises — preserve adaptive card table rendering in production — and nothing more:

- **Not missing:** explicit `Table` registration + a regression test that reproduces the tree-shaken condition.
- **Not over-built:** no unrelated behavior, flags, or elements added. The forbidden-element and action defense-in-depth is untouched.
- **No divergence:** the approach (explicit registration inside the octo context, not relying on the global side effect) matches the root cause.

**Spec: ✅**

## Correctness — is registering only `Table` enough? (Yes)

I verified this against the `adaptivecards@3.0.6` source rather than assuming, because "register the parent but not `TableRow`/`TableCell`" is a plausible-looking gap:

- Top-level card elements are instantiated through the registry: `SerializationContext.parseElement()` → `parseCardObject(..., (typeName) => this.elementRegistry.createInstance(typeName, ...))` (`card-elements.js`). This is why `Table` itself must be in the element registry — confirmed.
- A `Table`'s children are **not** instantiated through the registry. `Table.parseItem()` (via `StylableContainer.parseItem`) passes its own creation callback into `parseCardObject`, and `internalParseCardObject` uses that callback verbatim: `result = createInstanceCallback(typeName)`. For a table that callback is `Table.createItemInstance()`, which returns `new TableRow()` directly (`table.js:360`); `TableRow.createItemInstance()` in turn returns `new TableCell()` directly (`table.js:252`). Neither `TableRow` nor `TableCell` is ever looked up in the element registry, and neither is registered in `GlobalRegistry.defaultElements` in the first place.

So `TableRow` / `TableCell` do not need to be registered — registering `Table` alone fully restores table rendering. The `Versions.v1_5` schema version also matches the library default registration.

## Code quality

**Quality: Approved**

One advisory (non-blocking):

**[P2] Test cleanup does not actually restore the mutated global.** In `octoSdkConfig.test.ts`, the new test does:

```ts
GlobalRegistry.defaultElements.unregister("Table");
try { ... } finally { GlobalRegistry.reset(); }
```

`GlobalRegistry.reset()` only nulls the lazy `_elements` / `_actions` caches (`registry.js`); it does **not** rebuild `GlobalRegistry.defaultElements`, which is the master collection this test just mutated. So after this test runs, `defaultElements` is permanently missing `Table` for the remainder of that test file's module lifetime.

It's harmless today for two reasons: (1) this is the last test that touches the element registry in the file — the remaining `buildOctoHostConfig` tests don't read `defaultElements`; and (2) Vitest's default per-file isolation keeps the mutated module singleton from leaking into sibling test files. But the `finally { reset() }` reads as if it restores state, and it doesn't — a future element-dependent test added after this one in the same file would fail confusingly. Consider re-registering `Table` in the `finally` (e.g. `GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5)`) instead of / in addition to `reset()`, to make the cleanup match its intent.

## Overall verdict

**APPROVE.** Spec ✅ and no P0/P1 findings. The single P2 is a test-hygiene nit that does not block merge.

## Coverage / blind spots

- Verified the registry vs. direct-instantiation split from the actual `adaptivecards@3.0.6` package source (`table.js`, `card-elements.js`, `registry.js`), so the "children need registering too" hypothesis is refuted with evidence, not assumed.
- Not independently reproduced: the actual production bundler config dropping `lib/table.js`. The fix is safe regardless — explicit registration is idempotent with the library's own default registration (`CardObjectRegistry.register` overwrites in place), so it's correct whether or not tree-shaking strips the side effect.
- One outside advisory pass was consulted and its "also register TableRow/TableCell" flag was investigated and refuted against source as above; a second independent tooling pass was unavailable this round due to an unrelated gateway error, so this verdict rests on direct source verification.

OctoBoooot (approved):
## Review: preserve adaptive card table rendering in production (#638) — delta re-review

**Verdict:** Approve (re-affirmed)
The new commit fixes the test-teardown hygiene issue exactly as needed: the `finally` now re-registers Table into `GlobalRegistry.defaultElements` before `reset()`, so the test restores the state it mutated. The production fix is unchanged and correct. Approval carries forward.

Risk tier: normal. Base `main`; `mergeStateStatus=UNSTABLE` (see gate note).

### What changed since `f0453cac`
- `__tests__/octoSdkConfig.test.ts` (+2/-1): the regression test'"'"'s `finally` gains `GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5)` before `GlobalRegistry.reset()`.

### Why this is the right fix (resolves the test-isolation minor)
The test does `GlobalRegistry.defaultElements.unregister("Table")` to simulate the tree-shaken prod state. In adaptivecards@3.0.6, `reset()` only nulls the lazy `_elements`/`_actions` caches — it does **not** restore `defaultElements`, so on the old teardown Table stayed unregistered in `defaultElements` for the rest of the process (a later test relying on the global Table could fail by order). The new `finally` re-registers Table into `defaultElements` first, then `reset()` clears the caches so the next access re-copies from the now-restored defaults. Correct — this is the explicit-restore I flagged, and it settles the earlier `reset()`-semantics question (reset alone doesn't repopulate defaults).

### Standing
- Production fix (`octoSerialization.ts` explicit `Table` register on the local context) unchanged and correct — the shipped behavior was never affected by the test-teardown issue.

### Gate note
`check-sprint` is failing (governance: no linked issue/Sprint) → `mergeStateStatus=UNSTABLE`. Code review is satisfied; a maintainer must set the Sprint to clear the gate before merge — not a code issue.

Jerry-Xin (approved):
Re-affirming APPROVE at head `bae9a6a9`.

The only change since my prior APPROVE (`f0453cac`) is a test-teardown fix in `octoSdkConfig.test.ts` — and it correctly resolves the test-isolation nit raised on the prior head:

- Before: the test did `GlobalRegistry.defaultElements.unregister("Table")` then only `GlobalRegistry.reset()` in `finally`. Per adaptivecards@3.0.6, `reset()` only nulls the lazy `_elements`/`_actions` caches — it does NOT restore `defaultElements` — so Table stayed unregistered in the persistent global registry for the rest of the process, which could break a later test depending on global Table registration.
- After: the `finally` now does `GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5)` **before** `GlobalRegistry.reset()`, explicitly restoring Table to the persistent `defaultElements` and then resetting the lazy caches so the next access re-copies from the restored defaults. Correct restore-then-reset ordering; import adds `Table, Versions`.

Verified the delta touches ONLY the test file. The production fix is untouched: `octoSerialization.ts:56` still explicitly `elementRegistry.register("Table", Table, Versions.v1_5)` on the local context (defeats the `sideEffects:false` tree-shaking that drops the global side-effect registration). Root cause + fix mechanism unchanged and sound; no surface widening (Table is octo-whitelisted, not in FORBIDDEN); no double-register concern (register is idempotent, local registry).

No blocking or non-blocking issues remain — the test hygiene point is resolved.

yujiawei (approved):
# Code Review — PR #638 (octo-web)

## Summary

This PR fixes AdaptiveCards `Table` elements silently disappearing in production builds. Root cause is real and correctly diagnosed: the `adaptivecards` package declares `"sideEffects": false`, and `Table` is registered via a top-level side effect in a **standalone module** (`lib/table.js`). When nothing else imports a symbol from that module, a production bundler is free to drop it, taking the `GlobalRegistry.defaultElements.register("Table", …)` side effect with it. `populateWithDefaultElements()` then copies a registry that no longer contains `Table`, so valid cards render without their table section. The fix imports `Table` explicitly and registers it on the per-context element registry, which no longer depends on the fragile module side effect.

I verified the mechanism against `adaptivecards@3.0.6` source directly.

## Verification against the actual package

- `package.json` → `"sideEffects": false` — confirmed.
- `lib/table.js:466` → `GlobalRegistry.defaultElements.register("Table", Table, Versions.v1_5);` is the **only** side-effect registration for `Table`, and it lives in its own module. The `v1_5` version and type name in the fix match this exactly.
- All *other* octo-whitelisted elements (`Container`, `TextBlock`, `RichTextBlock`, `Image`, `ImageSet`, `FactSet`, `ColumnSet`, `ActionSet`, `Input.*`) register in `lib/card-elements.js:7196-7211`. That same module also defines the `AdaptiveCard` class, which `sdk/renderOctoCard.ts` imports and instantiates (`new AdaptiveCard()`). Because a live symbol from `card-elements.js` is always used, the module is retained and its registrations run. This is why only `Table` broke in practice, and why scoping the fix to `Table` is correct rather than incomplete.
- `CardObjectRegistry.register()` (`lib/registry.js:19`) overwrites in place if the name already exists, so the explicit `register("Table", …)` is idempotent and harmless when the side effect *did* survive (e.g. the dev server).
- `Table` is not in `FORBIDDEN_ELEMENTS`, and the registration runs before the unregister loop, so ordering is fine.

## Findings

### P2 (advisory, non-blocking) — pattern relies on `card-elements.js` staying imported
The fix is correct for the current package layout, but the safety of the *other* whitelisted elements rests on an implicit invariant: `card-elements.js` remains reachable via the `AdaptiveCard` import in `renderOctoCard.ts`. That invariant holds today and is effectively guaranteed (you cannot render a card without `AdaptiveCard`). Worth a one-line code comment noting that if a future `adaptivecards` version splits elements into per-element standalone modules the way `Table` is today, the same explicit-registration treatment would be needed for them. No action required for this PR.

### P2 (advisory) — test mutates a global singleton
The regression test calls `GlobalRegistry.defaultElements.unregister("Table")` to simulate the tree-shaken state. This mutates process-global SDK state. It is adequately guarded here: the mutation and restoration are wrapped in `try/finally`, and Vitest isolates test *files* in separate workers while running tests *within* a file sequentially, so there is no realistic cross-test race. Acceptable as written; just be aware the test is order-sensitive within its own file.

## Note on a concern I evaluated and dismissed

The `finally` block does `GlobalRegistry.defaultElements.register("Table", …)` followed by `GlobalRegistry.reset()`, which can look contradictory. It is not. `reset()` (`lib/registry.js:73`) only nulls the lazily-cached `_elements` / `_actions` singletons; it does **not** clear `defaultElements`, which is a separate static created once at module load (`lib/registry.js:78`). So the teardown correctly restores `Table` into `defaultElements` and then drops the derived cache so it re-populates cleanly on next access. No cascading suite failure results from this sequence.

## Test validity

The added test is meaningful, not a tautology: unregistering `Table` from `defaultElements` faithfully reproduces the tree-shaken production state (the class is still bundled because it is imported, but the registry no longer knows about it), and the assertion then proves the explicit registration is what keeps `Table` available. This is exactly the failure mode the PR fixes.

## Verdict

Spec: the change implements precisely the stated goal (preserve `Table` rendering in production) with no scope creep and a matching regression test. Correctness verified against the package source. The two findings above are advisory (P2) and do not block merge.

**APPROVED.**

PR #639, opened 2026-07-10
feat(loop): timeline feed + issue reactions + comment resolve (Phase C data layer)

MrAladdin:
## What

补齐 **Phase C(协作层)数据层** —— 之前三枪能力对齐把协作层只做了一半(标签/订阅/评论反应),timeline 与 issue reactions 被我误延后。本 PR 收口。UI 保持最小(dmloop UI 后续重做),耐久的是把剩余协作端点接进 API 层。

- **活动流**:`listTimeline()` → `GET /issues/:id/timeline`(裸数组)。详情页新增只读「活动」区,渲染 activity 条目(状态/优先级/负责人变更、创建、任务完成)—— 这些审计历史此前在 dmloop 完全不可见。过滤 `type==='activity'` 避免与评论区重复。
- **issue 级反应**:`addIssueReaction`/`removeIssueReaction`;`issue.reactions` 经 `GetIssue`(详情端点填充,list/PUT 不填)读回。评论反应条泛化为 `renderReactionBar(reactions, onToggle)`,标题下的 issue 反应条复用它。
- **评论 resolve**:`resolveComment`/`unresolveComment`(POST/DELETE `/comments/:id/resolve`);根评论加 resolve/unresolve 切换 + 已解决徽标。

**纠正一处此前的错误延后**:issue reactions 曾以「无读回路径」为由被我从前面的 PR 剔除——错的,`GetIssue` 会返回它们(源码核实纠正了凭空判断)。

## 对抗审查发现并已修(全部提交前)
- **patch 会清空 issue.reactions**:PUT 响应不带 reactions → 编辑任意字段后 issue 反应条消失(与之前 labels 同类 bug)。已加 `reactions: updated.reactions ?? issue.reactions` 保留。(Claude)
- **导航竞态守卫捕获太晚**:所有 post-mutation 重取改为在 **mutation await 之前**捕获 `reqRef` token 并传入守卫版 `reloadComments(token)`/`syncIssue(token)`;去掉 syncIssue 默认 token 让 TS 强制显式传。(codex,3 轮)
- **结构性根治**:push 点加 `<IssueDetailPage key={id}>` —— issueId 变化即整页重挂载,任何 issue A 的慢响应都无法画进 issue B(含 LabelEditor→syncIssue 路径)。token 守卫退化为其上的 defense-in-depth。

## /simplify(4 agent)
- 删冗余 `.slice()`(filter 已返新数组);
[...]

PR #640, opened 2026-07-10
feat(loop): multi-file skill editor and revamped create flow

superalsrk:
## Goal

Bring octo-web's Skill management UI to parity with the multica reference: multi-file skills backed by a file tree + code/markdown editor, a redesigned creation flow, and skill-usage visibility. Reuses the existing backend contract (`GET/PUT/POST /skills` return/accept inline `files[]`; `GET /agents` carries each agent's `skills`).

## Changes

- **Data layer** (`api/types.ts`): add `SkillFile`, `Skill.files`, `UpsertSkillReq.files` so skills carry `SKILL.md` (`content`) plus additional files inline.
- **Skill detail** (`panel/SkillDetailPage.tsx`): 3-pane layout — file tree | editor | metadata.
  - `SkillFileTree` — hierarchical tree, folder expand/collapse, `SKILL.md` pinned first.
  - `SkillFileViewer` — preview/edit toggle for markdown (rendered via LoopMarkdown), frontmatter card, plain editor for other files.
  - Add/delete file with path validation (no `/` prefix, no `..`, `SKILL.md` reserved, no dup); save submits the full file set.
  - "Used by N AI teammates" section, derived by inverting each agent's `skills`.
- **Frontmatter** (`ui/frontmatter.ts`): lightweight parser + `ensureSkillFrontmatter`, which prepends a standard `---\nname/description\n---` block on local create when missing, matching the server's `ParseSkillFrontmatter`.
- **New Skill modal** (`pages/SkillPage.tsx`): single two-column dialog — left tabs (blank draft / import from URL / copy from runtime) + right live preview, shared footer. Same APIs (`createSkill` / `importSkill` / runtime import).
- **Usage count**: list rows show "N AI teammates" (from `GET /agents`).
- **Misc**: en/zh i18n, CSS, skill-name validation (`ui/skillName.ts`), and fixed `t()` interpolation to pass `{ values: {...} }`.

## Verification

- `pnpm --filter web build` (vite) passes.
- Strict `tsc --skipLibCheck` clean for changed files.
- Manual smoke: open a multi-file skill → tree loads, `SKILL.md` preview renders frontmatter; toggle preview/edit; add/delete file; save persists after reload; create via all 3 tabs; long frontmatter wraps without breaking the toggle; list shows usage count.

## Note

Base `feat/octo-loop` advanced by #639 after this branch forked (benign divergence, different files). Recommend a rebase onto `feat/octo-loop` before merge.

### DIFF
--- a/packages/dmloop/src/panel/SkillDetailPage.tsx
+++ b/packages/dmloop/src/panel/SkillDetailPage.tsx
@@ -1,47 +1,181 @@
-import React, { useEffect, useState } from "react";
-import { Typography, Input, Button, Spin, Tag, Toast, TextArea, Banner } from "@douyinfe/semi-ui";
-import { ArrowLeft, Save, Trash2 } from "lucide-react";
+import React, { useEffect, useMemo, useState } from "react";
+import { Typography, Input, Button, Spin, Toast, Banner, Tooltip } from "@douyinfe/semi-ui";
+import { ArrowLeft, BookOpen, Clock3, ExternalLink, Save, Trash2, Plus, Users } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
-import type { Skill } from "../api/types";
+import type { I18nFormatter } from "@octo/base";
+import type { Agent, Skill, SkillFile } from "../api/types";
 import { getSkill, updateSkill, deleteSkill, skillSource } from "../api/skillApi";
+import { listAgents } from "../api/agentApi";
+import { confirmDelete } from "../ui/confirmDelete";
+import SkillFileTree from "./SkillFileTree";
+import SkillFileViewer from "./SkillFileViewer";
+import { isValidSkillName } from "../ui/skillName";
+import { parseFrontmatter } from "../ui/frontmatter";
 import "./sideDetail.css";
 
 const { Text } = Typography;
-const SRC: Record<string, "green" | "blue" | "grey"> = { github: "green", local: "blue", workspace: "grey" };
+const SKILL_MD = "SKILL.md";
 
-/** Skill 独立详情页：左侧元信息 + 右侧内容编辑。 */
+type DraftFile = { path: string; content: string };
+
+function toDraftFiles(files?: SkillFile[]): DraftFile[] {
+  return (files ?? [])
+    .filter((f) => f.path !== SKILL_MD)
+    .map((f) => ({ path: f.path, content: f.content }));
+}
+
+function formatSkillTime(value: string | undefined, format: Pick<I18nFormatter, "date" | "relativeTime">): string {
+  if (!value) return "-";
+  const date = new Date(value);
+  if (Number.isNaN(date.getTime())) return "-";
+  const diff = date.getTime() - Date.now();
+  if (!Number.isFinite(diff)) return "-";
+  const absDiff = Math.abs(diff);
+  const minute = 60 * 1000;
+  const hour = 60 * minute;
+  const day = 24 * hour;
+  if (absDiff < minute) return format.relativeTime(0, "minute");
+  if (absDiff < hour) return format.relativeTime(Math.round(diff / minute), "minute");
+  if (absDiff < day) return format.relativeTime(Math.round(diff / hour), "hour");
+  if (absDiff < 10 * day) return format.relativeTime(Math.round(diff / day), "day");
+  return format.date(date, { month: "short", day: "numeric" });
+}
+
+/** Skill 详情页：左侧文件树 + 右侧多文件编辑器（SKILL.md 映射到 content）。 */
 export default function SkillDetailPage({ skillId, onChanged }: { skillId: string; onChanged?: () => void }) {
-  const { t } = useI18n();
+  const { t, format } = useI18n();
   const [row, setRow] = useState<Skill | null>(null);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<string | null>(null);
+  const [usedAgents, setUsedAgents] = useState<Agent[]>([]);
+
   const [name, setName] = useState("");
-  const [desc, setDesc] = useState("");
   const [content, setContent] = useState("");
+  const [files, setFiles] = useState<DraftFile[]>([]);
+  const [selectedPath, setSelectedPath] = useState(SKILL_MD);
   const [dirty, setDirty] = useState(false);
+  const [editingName, setEditingName] = useState(false);
+
+  const [addingFile, setAddingFile] = useState(false);
+  const [newPath, setNewPath] = useState("");
+  const [addError, setAddError] = useState("");
+
+  const seed = (s: Skill) => {
+    setRow(s);
+    setName(s.name);
+    setContent(s.content ?? "");
+    setFiles(toDraftFiles(s.files));
+    setDirty(false);
+  };
 
   useEffect(() => {
     setLoading(true);
     setError(null);
-    getSkill(skillId)
-      .then((s) => { setRow(s); setName(s.name); setDesc(s.description); setContent(s.content ?? ""); setDirty(false); })
+    setSelectedPath(SKILL_MD);
+    Promise.all([
+      getSkill(skillId),
+      listAgents().catch(() => [] as Agent[]),
+    ])
+      .then(([skill, agents]) => {
+        seed(skill);
+        setUsedAgents(agents.filter((agent) => (agent.skills ?? []).some((sk) => sk.id === skill.id)));
+      })
       .catch((e) => setError(e?.message ?? "load failed"))
       .finally(() => setLoading(false));
   }, [skillId]);
 
+  const fileMap = useMemo(() => {
+    const map = new Map<string, string>();
+    map.set(SKILL_MD, content);
+    for (const f of files) if (f.path.trim()) map.set(f.path, f.content);
+    return map;
+  }, [content, files]);
+  const filePaths = useMemo(() => Array.from(fileMap.keys()), [fileMap]);
+  const selectedContent = fileMap.get(selectedPath) ?? "";
+  const skillFrontmatter = useMemo(() => parseFrontmatter(content).frontmatter, [content]);
+  const skillDescription = skillFrontmatter?.description?.trim() ?? "";
+
+  useEffect(() => {
+    if (selectedPath !== SKILL_MD && !fileMap.has(selectedPath)) setSelectedPath(SKILL_MD);
+  }, [fileMap, selectedPath]);
+
   const back = () => WKApp.routeRight.pop();
-  const save = async () => {
-    if (!name.trim()) { Toast.warning(t("loop.validate.nameRequired")); return; }
-    try { await updateSkill(skillId, { name: name.trim(), description: desc, content }); setDirty(false); Toast.success(t("loop.toast.saved")); onChanged?.(); }
-    catch (e) { Toast.error((e as Error)?.message ?? "save failed"); }
+
+  const onFileContentChange = (next: string) => {
+    if (selectedPath === SKILL_MD) setContent(next);
+    else setFiles((prev) => prev.map((f) => (f.path === selectedPath ? { ...f, content: next } : f)));
+    setDirty(true);
   };
-  const remove = async () => {
-    try { await deleteSkill(skillId); Toast.success(t("loop.toast.deleted")); onChanged?.(); back(); }
-    catch (e) { Toast.error((e as Error)?.message ?? "delete failed"); }
+
+  const validateNewPath = (p: string): string => {
+    const v = p.trim();
+    if (!v) return t("loop.skill.detail.addFile.empty");
+    if (v.startsWith("/")) return t("loop.skill.detail.addFile.absolute");
+    if (v.split("/").includes("..")) return t("loop.skill.detail.addFile.doubleDot");
+    if (v === SKILL_MD) return t("loop.skill.detail.addFile.reserved");
+    if (filePaths.includes(v)) return t("loop.skill.detail.addFile.exists");
+    return "";
+  };
+
+  const submitNewFile = () => {
+    const err = validateNewPath(newPath);
+    if (err) { setAddError(err); return; }
+    const p = newPath.trim();
+    setFiles((prev) => [...prev, { path: p, content: "" }]);
+    setSelectedPath(p);
+    setDirty(true);
+    setAddingFile(false);
+    setNewPath("");
+    setAddError("");
+  };
+
+  const cancelNewFile = () => { setAddingFile(false); setNewPath(""); setAddError(""); };
+
+  const deleteSelectedFile = () => {
+    if (selectedPath === SKILL_MD) return;
+    setFiles((prev) => prev.filter((f) => f.path !== selectedPath));
+    setSelectedPath(SKILL_MD);
+    setDirty(true);
+  };
+
+  const save = async () => {
+    const skillName = name.trim();
+    if (!skillName) { Toast.warning(t("loop.validate.nameRequired")); return; }
+    if (!isValidSkillName(skillName)) { Toast.warning(t("loop.skill.namePattern")); return; }
+    try {
+      const updated = await updateSkill(skillId, {
+        name: skillName,
+        description: skillDescription,
+        content,
+        files: files.filter((f) => f.path.trim() && f.path !== SKILL_MD),
+      });
+      seed(updated);
+      Toast.success(t("loop.toast.saved"));
+      onChanged?.();
+    } catch (e) {
+      Toast.error((e as Error)?.message ?? "save failed");
+    }
+  };
+
+  const remove = () => {
+    confirmDelete({
+      title: t("loop.confirm.delete"),
+      okText: t("loop.action.delete"),
+      cancelText: t("loop.action.cancel"),
+      onOk: async () => {
+        try { await deleteSkill(skillId); Toast.success(t("loop.toast.deleted")); onChanged?.(); back(); }
+        catch (e) { Toast.error((e as Error)?.message ?? "delete failed"); }
+      },
+    });
   };
 
   if (loading) return <div className="loop-sd"><div className="loop-sd__center"><Spin /></div></div>;
-  if (error || !row) return <div className="loop-sd"><div className="loop-sd__topbar"><Button icon={<ArrowLeft size={16} />} theme="borderless" onClick={back}>{t("loop.detail.back")}</Button></div><div className="loop-sd__center">{error ? <Banner type="danger" description={error} /> : <Text type="tertiary">{t("loop.detail.notFound")}</Text>}</div></div>;
+  if (error || !row) return (
+    <div className="loop-sd">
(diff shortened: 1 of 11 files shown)


packages/dmloop/src/panel/SkillDetailPage.tsx
```diff
@@ -1,47 +1,181 @@
...
+  const cancelNewFile = () => { setAddingFile(false); setNewPath(""); setAddError(""); };
+
+  const deleteSelectedFile = () => {
+    if (selectedPath === SKILL_MD) return;
+    setFiles((prev) => prev.filter((f) => f.path !== selectedPath));
+    setSelectedPath(SKILL_MD);
+    setDirty(true);
+  };
+
   const save = async () => {
```
OctoBoooot, line 141: **[minor]** `save` has no in-flight guard: the Save button is `disabled={!dirty}` and dirty stays true until the response seeds, so a rapid double-click before the response resolves can double-submit. Mitigated by the idempotent upsert (same class as the toggle busy-guard notes on #635-#639). Add a `saving` flag disabling Save during the request. Non-blocking.

OctoBoooot (approved):
## Review: improve loop skill management UI (#640)

**Verdict:** Approve
Large UI PR (2015 lines, mostly CSS/i18n) with a compact, well-guarded logic surface. Reviewed the high-signal parts — the frontmatter parser, the multi-file editor write path, and the file viewer — with security focus. Clean.

Risk tier: normal; size/XL → `needs-human-review` applied (human sign-off before merge). Base `feat/octo-loop`; `mergeStateStatus=UNSTABLE` + no linked issue (`check-sprint`).

### What I verified (scoped to the logic; CSS +867 / i18n +138 skimmed as presentational)
- **`frontmatter.ts` parser — safe.** `FRONTMATTER_RE` is anchored with a non-greedy body bounded by the `---` terminator (no catastrophic-backtracking risk); `parseSimpleYaml` is pure line-by-line `key: value` (no eval / no code exec). The write side'"'"'s `yamlScalar` correctly quotes unsafe values via `JSON.stringify` (valid YAML double-quoted scalar) so a name/description with `:`/`#`/newline/quote/YAML-special leading char round-trips safely — a real injection-prevention measure. `ensureSkillFrontmatter` only prepends when absent.
- **`SkillFileViewer.tsx` — XSS-safe.** Frontmatter rendered as escaped JSX text in a table; body via the existing sanitized `LoopMarkdown`; edit mode is a plain `<textarea value=…>`. No `dangerouslySetInnerHTML`.
- **Write path (`SkillDetailPage.save` + file add) — path-safe & validated.** `validateNewPath` rejects empty, absolute (`startsWith("/")`), traversal (`split("/").includes("..")`), reserved SKILL.md, and duplicates — the right guard for user-supplied file paths. `save` validates name (required + `isValidSkillName`), sends `content` + filtered `files`, try/catch → Toast, `seed(updated)` refreshes. `dirty` tracked on every edit; Save `disabled={!dirty}`. Description derived from SKILL.md frontmatter per the PR intent.

### Minor (non-blocking)
- `save` has no in-flight busy guard: the Save button is gated on `dirty` (which stays true until the response seeds), so a rapid double-click before the response resolves could double-submit. Mitigated by the idempotent upsert; same class as the toggle busy-guard notes across #635–#639. A `saving` flag would tidy it.

### Praise
- For a user-editable multi-file skill store, the two things that matter most are done right: file paths are validated against absolute/`..` traversal before they can be saved, and any user text going *into* YAML frontmatter is escaped via a proper scalar-quoting helper (JSON.stringify) rather than naive interpolation. Combined with rendering all frontmatter/body through escaped JSX + the existing sanitized markdown, the untrusted-content surface is handled deliberately, not incidentally.

Jerry-Xin (approved):
**APPROVE** — Solid multi-file skill editor + revamped create flow. Reviewed the isolated PR diff (merge-base `a29ea39b` → `c4ae4032`): exactly **+1842/-173 across 11 files, no rebase drift** in these files (base has advanced to `d427808d` with unrelated changes — a rebase before merge is fine but not blocking). Byte-verified the high-risk surfaces (XSS render + mutation set); sampled CSS/i18n.

### Files (grouped)
- **types** — `api/types.ts`: `SkillFile`, `Skill.files`, `UpsertSkillReq.files` (inline file set).
- **components** — `panel/SkillDetailPage.tsx` (3-pane: tree/editor/metadata + save/delete + "used by N"), `panel/SkillFileTree.tsx` (hierarchical tree), `panel/SkillFileViewer.tsx` (md preview/edit + frontmatter), `pages/SkillPage.tsx` (list + 3-tab create modal).
- **util** — `ui/frontmatter.ts` (parser + `ensureSkillFrontmatter`), `ui/skillName.ts` (name allowlist).
- **styles** — `pages/loop.css`, `panel/sideDetail.css`.
- **i18n** — `i18n/en-US.json`, `i18n/zh-CN.json`.

### Mutation set (all correct)
`createSkill` (POST `/skills`), `importSkill` (POST `/skills/import`), `importRuntimeSkill` (loop), `updateSkill` (PUT `/skills/{id}`), `deleteSkill` (DELETE `/skills/{id}`). Server-issued ids in path; untrusted values (name/desc/content/files[]) sent in JSON body, never URL-concatenated. Errors surfaced via `Toast.error`; list/runtime responses unwrap with `?? []` fallback. Delete is guarded by `confirmDelete` (both list row and detail page).

### Render set (XSS-safe)
All user-authored skill fields — `name`, `description`, frontmatter `key`/`value` (`<th>{key}</th><td>{value}</td>`), `agent.name`, tree `node.name` — render as **escaped React text**. Skill body renders via `LoopMarkdown` = `react-markdown` + `remark-gfm` with **no `rehype-raw`**, so raw HTML in skill content is escaped, and the `a` override adds `rel="noreferrer" target="_blank"` (react-markdown drops `javascript:` hrefs). No `dangerouslySetInnerHTML`, no `eval`/`Function`, no field interpolated into `href`/`src`/`style`. ✔

### Validation & robustness
- Name required + `^[A-Za-z0-9-]+$` on both create and edit; frontmatter name re-validated on local create; runtime import validates each name.
- New-file path validation: rejects empty, leading `/`, `..` segment, reserved `SKILL.md`, duplicates.
- `frontmatter.ts` `yamlScalar` correctly double-quotes unsafe values — good injection hygiene into the generated frontmatter.
- Loading / error (Banner) / empty states present; `SkillFileViewer` uses `key={selectedPath}` remount; detail load effect keys on `skillId`.
- i18n en/zh full parity (55/55 skill keys, zero gaps). No `console.log`, TODO, dead code, or hardcoded secrets/tokens in the diff.

### 🟡 Non-blocking
- **Double-submit guard inconsistency.** The web (`webBusy`) and runtime (`rtBusy`) import tabs correctly gate with an in-flight flag + `loading`, but `createLocal` (`pages/SkillPage.tsx:104`, button `disabled={!nName.trim()}`) and `save` (`panel/SkillDetailPage.tsx:141`, button `disabled={!dirty}`) have no in-flight guard — a fast double-click during the await can fire two `createSkill`/`updateSkill` calls. Same class as the busy-guard-window noted on prior loop PRs; worth mirroring the `webBusy`/`rtBusy` pattern for consistency, but not blocking.

Note: server-side authz for skill mutations is a backend concern outside this frontend diff — the client contract here is correct.

yujiawei (approved):
# Code Review — PR #640 (octo-web)

Reviewed at head SHA `c4ae4032caae9a2996600dfa7e1e2fa05dd5c382` against base `feat/octo-loop`. Scope: the `packages/dmloop` skill-management UI rewrite (list redesign, create-skill modal, multi-file skill editor, frontmatter parsing, name validation, i18n/CSS).

## 1. Scope / completeness

The diff is coherent with the stated goal. It delivers a self-contained UI improvement:

- `api/types.ts` — adds `SkillFile` + `files` on `Skill`/`UpsertSkillReq`.
- `SkillPage.tsx` — replaces the `Table` with a custom list, redesigns the create modal (left form + live preview), moves keyword filtering client-side, derives per-skill usage counts from agents.
- `SkillDetailPage.tsx` — replaces the single-textarea editor with a file-tree + multi-file editor; `SKILL.md` maps to `content`.
- New `SkillFileTree.tsx`, `SkillFileViewer.tsx`, `ui/frontmatter.ts`, `ui/skillName.ts`.
- i18n (en/zh) keys are in parity (verified — no orphaned keys either direction), all referenced keys resolve, JSON is valid.

No unrequested/out-of-scope behavior, and no leftover references to removed imports (`Table`, `Tabs`, `Popconfirm`, `Monitor`, `Globe`, `SRC`, `setDesc`, etc.) — the build surface looks clean. I could not run `tsc` in this environment (no installed deps), so type-checking rests on static inspection.

## 2. Findings

No blocking (P0/P1) issues. The items below are quality/consistency improvements (P2) — none should block merge.

### P2 — `name` and frontmatter can drift out of sync
`SkillDetailPage.tsx` (`save`, ~L146): `name` is sent from the sidebar input state, while `description` is parsed from the `SKILL.md` frontmatter (`skillDescription`). If a user edits `name:` **inside** the `SKILL.md` editor, the sidebar state is unchanged, so the persisted `skill.name` and the frontmatter `name:` diverge (and a frontmatter-only rename silently does not update `skill.name`). Consider deriving `name` from the frontmatter too (single source of truth), or syncing the two on edit. Content is not lost — both values persist — but the UI state is confusing.

### P2 — create modal silently ignores form fields when pasted content already has frontmatter
`SkillPage.tsx` (`createLocal`, ~L74): when `nContent` already contains a frontmatter block, `ensureSkillFrontmatter` returns it unchanged and `frontmatterName`/`frontmatterDesc` are taken from the pasted block, discarding whatever was typed in the Name/Description fields. This is arguably "frontmatter wins," but the preview panel still shows the typed `nName`, so the created skill can differ from the preview. Worth either disabling/merging the form fields when pasted frontmatter is detected, or reflecting the effective values in the preview.

### P2 — file-tree path collisions render some files unreachable
`SkillFileTree.tsx` (`buildTree`) keys nodes by name and never reconciles a leaf vs. directory conflict. `validateNewPath` (`SkillDetailPage.tsx`, ~L110) only rejects exact-path duplicates. So adding `a/b/c.md` and then `a/b` passes validation, but `buildTree` folds the `a/b` file into the existing `a/b` directory node — the file is saved but becomes invisible/uneditable in the tree. Narrow edge case (requires a deliberate prefix-colliding path), but worth rejecting prefix collisions in `validateNewPath`.

### P2 — lightweight frontmatter parser is intentionally partial
`ui/frontmatter.ts` `parseSimpleYaml` handles only flat `key: value` lines; nested maps/lists and multi-line scalars are skipped. `yamlScalar` emits JSON-escaped values (e.g. `\n`) that `parseSimpleYaml` does not un-escape, so a multiline description does not perfectly round-trip. The code comments already acknowledge this and point at the server-side parser semantics, which is the right call for a display/name-description helper — flagging only so the limitation is a conscious contract. If skills ever carry richer frontmatter (tool args, list values), switch to the `yaml` package.

### P2 (nit) — runtime import shows success even when nothing imported
`SkillPage.tsx` (`importFromRuntime`): the success toast fires with `ok=0` if every selected skill was skipped for an invalid name. Consider a warning when `ok===0`.

## 3. Notes on verification coverage

- **Backend contract (not verifiable here):** the new usage counts (`usedByCount`) and the detail page's "used by N teammates" rely on `GET /agents` returning a populated `skills[]` on each agent. The released agent-detail view already consumes `agent.skills`, so this is consistent with existing usage, but the list endpoint's payload can only be confirmed against the server repo. Similarly, `UpsertSkillReq.files` persistence and the `updateSkill` response echoing `files` (used by `seed(updated)`) depend on the server side.
- **Regression from moving keyword filtering client-side:** `listSkills()` now fetches all skills and filters in-memory. Fine for expected workspace sizes; only a concern at very large skill counts.
- **Not run:** `tsc`/`eslint`/unit tests (dependencies not installed in the review environment); assessment is static.

## 4. Verdict

Solid, well-structured UI work with clear separation between view and edit state. The findings are all non-blocking consistency/edge-case polish. Approving; please consider the `name`/frontmatter sync (first P2) as the most worthwhile follow-up.

yujiawei (changes requested):
# Code Review — PR #640 (octo-web) — Updated (supersedes my earlier approval)

Re-reviewed at head SHA `c4ae4032caae9a2996600dfa7e1e2fa05dd5c382`. A second pass surfaced a **P1 data-loss regression** in the skill save path that I missed in my first review. This updated review supersedes my earlier approval and requests changes. Apologies for the churn.

## P1 (blocking) — `description` is silently wiped on save for any skill whose `content` has no frontmatter

`packages/dmloop/src/panel/SkillDetailPage.tsx`

The previous detail page kept a dedicated `desc` state seeded from the server column and round-tripped it on save:

```
// before (parent a29ea39b)
.then((s) => { ...; setDesc(s.description); ... })
await updateSkill(skillId, { name: name.trim(), description: desc, content });
```

This PR removes that state and derives the description **solely** from the SKILL.md frontmatter (L95–96), with no fallback to the existing column value:

```
const skillFrontmatter = useMemo(() => parseFrontmatter(content).frontmatter, [content]);
const skillDescription = skillFrontmatter?.description?.trim() ?? "";   // "" when no frontmatter
```

and `save()` (L146–150) sends it verbatim:

```
await updateSkill(skillId, { name: skillName, description: skillDescription, content, files: ... });
```

**Trace for a pre-existing skill** (created by the old `createLocal`, which stored the description in the column and did **not** inject frontmatter into `content`):

- `getSkill` → `{ description: "original text", content: "<no --- block>" }`
- `seed` sets `content` to that block; `parseFrontmatter(content).frontmatter` → `null`
- `skillDescription` → `""`
- Any edit flips `dirty` (name / file content / add / delete file all call `setDirty(true)`), enabling Save
- First Save → `updateSkill(..., { description: "" })` → the original description is overwritten with an empty string.

It's silent: the sidebar shows "No description in SKILL.md frontmatter", so the user has no cue that saving is about to clear the stored description. This hits **every skill created before this PR** (the new `createLocal` injects frontmatter via `ensureSkillFrontmatter`, so newly created skills are safe — but the whole existing corpus is not).

I confirmed the regression against the parent commit (old code round-tripped `s.description`; new code does not). I could not confirm the server side — the dmloop skills backend (referenced as `server/internal/skill` in `frontmatter.ts`) is not in octo-server or this repo, so I cannot verify whether the API ignores empty `description` on update. Absent a confirmed server-side guard, sending `""` for a field the user never edited is a client-side data-loss bug and should be treated as blocking.

**Fix direction:** fall back to the persisted value when frontmatter yields nothing, e.g. `skillFrontmatter?.description?.trim() || row?.description || ""`; or, cleaner, back-migrate on load by running `ensureSkillFrontmatter(name, row.description, content)` into `content` in `seed()` so the frontmatter becomes the single source of truth without dropping the column value on first save. Either way, add a regression test for "skill with column description + no-frontmatter content survives an edit+save".

## P2 (non-blocking)

- **Duplicate `formatSkillTime` with divergent fallback.** Defined in both `pages/SkillPage.tsx:25` (empty → `"—"` em-dash) and `panel/SkillDetailPage.tsx:19` (empty → `"-"` hyphen). Extract into one shared util so missing timestamps render consistently.
- **Orphaned i18n key `loop.skill.fromWebHint`** (`en-US.json:238` / `zh-CN.json:238`) — no longer referenced after the create-modal tab redesign. Remove to avoid dead translation keys.
- **`name` vs frontmatter can drift.** `name` is saved from the sidebar input while `description` comes from frontmatter; editing `name:` inside SKILL.md won't update `skill.name` (and vice versa). Prefer a single source of truth. (No data loss — both persist — but confusing.)
- **Create-modal ignores form fields when pasted content already has frontmatter** (`SkillPage.tsx` `createLocal`): typed Name/Description are discarded while the preview still shows the typed name; reflect the effective values or merge them.
- **File-tree path prefix collision** (`SkillFileTree.buildTree` + `validateNewPath`): adding `a/b/c.md` then `a/b` passes validation but the `a/b` file is folded into the directory node and becomes unreachable in the tree (data still persisted). Reject prefix collisions in `validateNewPath`.

## Nit

- Dead server-side keyword path: `reload` now calls `listSkills()` with no params; `matchKeyword` in `skillApi.ts` and the `keyword` param are unused. Remove or comment for intent.

## Verification coverage

- i18n en/zh key parity holds; all referenced keys resolve; JSON valid. Name/path validation is applied consistently. No leftover references to removed imports.
- Not run: `tsc` / eslint / unit tests (deps not installed here) — static analysis only.
- Backend contract (skills `files` persistence, `updateSkill` empty-description semantics, `GET /agents` returning `skills[]`) not verifiable from the available repos.

## Verdict

Requesting changes: the P1 description data-loss on save for pre-existing skills should be fixed (with a test) before merge. The P2 items are non-blocking follow-ups.

Jerry-Xin (changes requested):
Correcting my earlier APPROVE to REQUEST_CHANGES on the merits — yujiawei's 🔴 is a real data-loss regression I missed. (The PR is already merged; this correction is for the record and should land as a follow-up fix.)

## 🔴 Blocking — Save silently wipes a skill's stored description when SKILL.md has no `description:` frontmatter

Byte-verified at head `c4ae4032`:

- `SkillDetailPage.save` (`packages/dmloop/src/panel/SkillDetailPage.tsx:146-151`) sends `description: skillDescription` in the PUT **unconditionally**.
- `skillDescription` (line 96) = `parseFrontmatter(content).frontmatter?.description?.trim() ?? ""`.
- `parseFrontmatter` returns `{ frontmatter: null }` when `content` has no `---` frontmatter block (`frontmatter.ts` FRONTMATTER_RE no-match), so `skillDescription` collapses to `""`.
- `UpsertSkillReq.description` is optional (`types.ts:286`), but the client **always populates it** (with `""`), so `updateSkill` (`httpPut /skills/{id}`) overwrites the backend's persisted `Skill.description` (`types.ts:276`, a separate stored column) with empty.

**Trigger (realistic):** a skill whose SKILL.md body has no `description:` frontmatter line — description stored as backend metadata rather than in-file. Opening it in the detail page, making any edit, and hitting Save writes `description: ""`, silently destroying the stored description. No confirmation, no floor.

**Root cause — create/edit asymmetry:** the create path (`SkillPage.tsx:105`) correctly runs `ensureSkillFrontmatter(name, desc, content)` to inject the description into frontmatter before persisting. The detail-page save path does **not** use `ensureSkillFrontmatter` — it derives description read-only from frontmatter with `?? ""` and pushes it back, so a frontmatter-less skill round-trips to empty.

**Suggested fixes (any one):**
- Guard: if `parseFrontmatter(content).frontmatter?.description` is absent, omit `description` from the PUT (send `undefined`, not `""`) so the backend keeps its stored value; OR
- Wire `ensureSkillFrontmatter` into the save path so the current description is preserved into the SKILL.md header before persisting; OR
- Keep description as an explicit editable field (like the create modal's `nDesc` input) instead of deriving it solely from frontmatter.

## 🟡 Non-blocking
- `formatSkillTime` is duplicated — extract to a shared util.
- Unused i18n keys — clean up.
- (Prior) `save`/`createLocal` lack an in-flight double-submit guard — consistent with the loop-series busy-guard family.

I under-weighted the "`ensureSkillFrontmatter` not wired into save" observation as cosmetic in my initial review; it's actually the mechanism of this data-loss path. Correcting.

PR #643, opened 2026-07-10
feat(loop): file attachments on issues and comments

MrAladdin:
## What

第五枪能力对齐:**issue / 评论文件附件**。基于跨仓调研(octo 附件能力核实)结论——multica 已有 `/api/upload-file` + `attachment_ids` 绑定,dmloop 复用现成 http 通道 + markdown 渲染,新增面很小。

- `attachmentApi.uploadAttachment(file, {issueId?|commentId?})` → `POST /api/upload-file`(multipart;共享 axios client 自动设 boundary,拦截器注入 token/workspace/CSRF)。types 加 `Attachment` + `issue.attachments`/`comment.attachments`(详情/列表端点回填)。
- **issue 附件**:详情页「附件」区列出 issue 级附件(过滤 `comment_id==null`),上传按钮带 issueId 即时绑定后重取。图片内联(download_url),其它为下载链接。
- **评论附件**:输入区**本地持有**选中文件(pendingFiles),**发送时才带 commentId 上传绑定**——取消/离开不上传任何东西(无 issue 孤儿附件)。渲染在各评论下。

## 对抗审查发现并已修(全部提交前,e2e+DB 双证)
- patch() 补 `attachments` 保留(PUT 不回填,同 labels/reactions 类,否则编辑任意字段清空附件)。(Claude)
- 评论附件从「带 issueId 上传 + attachment_ids link」改为 **File-hold 发送时带 commentId 直传**:前者会让被取消/放弃的评论草稿变成永久 issue 附件(comment_id null / issue_id set);后者直接绑 comment_id(该后端路径无需 issue_id)——**DB 验证评论附件 issue_id=null、无 issue 孤儿**。(codex)
- submitComment:评论创建成功后**先清输入态再传附件**,单个附件失败隔离——附件报错不再残留可重复提交的草稿(防重复评论)。(codex)
- 上传守卫:提交中禁发送;uploadForIssue 先解锁 uploading 再 fire-and-forget 重取,避免重取失败卡死选择器。(codex)

## /simplify(4 agent)
clean —— 内联而非跨包复用 dmworkbase 的 FileToolbar(dmloop 不依赖 dmworkbase)、不抽 withUpload(2 处且变体漏)、自写 chip,均判为「能力层优先、UI 后续重画」下的正确懒选择。

## 真实浏览器 e2e(dev / Alice_Space / MV2 E2E,3 轮)
- issue 图片附件:上传 → 「附件 (1)」内联渲染 `/download>`。
- 评论附件 File-hold:选文件(本地 chip,未上传)→ 填评论 → 发送 → 评论下渲染附件链接;**DB 确认 comment_id 绑定、issue_id=null(无孤儿)、issue 附件区未被污染**。
- 测试数据按约定保留。

## Blast radius(模块外)
- `Attachment`/`issue.attachments`/`comment.attachments` 均可选新增;IssueList/IssueBoard 不消费;addComment 已回退去掉临时的 attachment_ids 参数(File-hold 改走 commentId 上传),无死参。
- FormData 上传经现成 axios(^0.25)+ 拦截器,不覆盖 Content-Type(浏览器设 multipart+boundary),实证工作。
- 整包 `tsc --skipLibCheck` 仅 2 处既有报错(非本 PR),PR-E 文件零错误。

基于最新 `feat/octo-loop`(含 #635/#636/#637/#639 我方四枪 + 他人 #640 skill-editor)。附件删除入口、图片内联进 markdown 正文属重 UI,随后续 UI 重画;@提及、订阅 toggle 另行。

### DIFF
--- a/packages/dmloop/src/api/attachmentApi.ts
+++ b/packages/dmloop/src/api/attachmentApi.ts
@@ -1,0 +1,17 @@
+// @octo/loop — 附件 API(上传走 multica /api/upload-file,multipart)。
+import type { Attachment } from "./types";
+import { httpPost } from "./http";
+
+// 上传单个文件。可选 issueId/commentId 在上传时直接绑定(需目标已存在);
+// 否则拿返回的 id,通过 create/update issue|comment 的 attachment_ids 绑定(如新评论)。
+// 走现成 http client:axios 对 FormData 自动设 multipart 边界,鉴权/workspace header 由拦截器注入。
+export function uploadAttachment(
+  file: File,
+  opts?: { issueId?: string; commentId?: string },
+): Promise<Attachment> {
+  const form = new FormData();
+  form.append("file", file);
+  if (opts?.issueId) form.append("issue_id", opts.issueId);
+  if (opts?.commentId) form.append("comment_id", opts.commentId);
+  return httpPost<Attachment>("/upload-file", form);
+}
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -124,6 +124,8 @@ export interface Issue {
   labels?: IssueLabel[] | null;
   // issue 级 emoji 反应:仅详情端点(GetIssue)回填;list/update/ws 不带 → 保持已有。
   reactions?: IssueReaction[] | null;
+  // issue 附件:仅详情端点(GetIssue)回填;list/update/ws 不带 → 保持已有。
+  attachments?: Attachment[] | null;
   // 搜索结果专属(GET /issues/search):命中来源 + 高亮片段。列表/详情端点不返回。
   match_source?: string;
   matched_snippet?: string | null;
@@ -139,6 +141,8 @@ export interface CreateIssueReq {
   assignee_type?: AssigneeType | null;
   assignee_id?: string | null;
   project_id?: string | null;
+  // 新建时绑定已上传附件(上传返回的 id 列表)。
+  attachment_ids?: string[];
 }
 export interface UpdateIssueReq {
   title?: string;
@@ -156,6 +160,8 @@ export interface UpdateIssueReq {
   due_date?: string | null;
   parent_issue_id?: string | null;
   stage?: number | null;
+  // 绑定新上传的附件(id 列表);后端幂等,重复 id 无副作用。
+  attachment_ids?: string[];
   // 指派/状态变更触发 agent run 时：suppress_run=true 表示“暂不开始”；handoff_note 仅在真起 run 时消费。
   suppress_run?: boolean;
   handoff_note?: string;
@@ -216,6 +222,23 @@ export interface IssueSubscriber {
   created_at: string;
 }
 
+/** 附件(上传返回 + issue/comment 详情回填)。
+ *  download_url 为短时签名 URL(仅即时展示、勿持久化);markdown_url 可持久化内联进 markdown 正文。 */
+export interface Attachment {
+  id: string;
+  filename: string;
+  url: string;
+  download_url: string;
+  markdown_url: string;
+  content_type: string;
+  size_bytes: number;
+  created_at?: string;
+  // 归属:issue 级附件 comment_id 为空;评论附件 comment_id 指向评论。
+  // (后端 attachment 先绑 issue_id,再由 comment 的 attachment_ids 补 comment_id。)
+  issue_id?: string | null;
+  comment_id?: string | null;
+}
+
 /** issue 时间线条目(GET /issues/:id/timeline,不带分页参数时为裸数组、ASC)。
  *  合并 comment + activity 两类;活动流只用 activity 类(action/details)。 */
 export interface TimelineEntry {
@@ -248,6 +271,8 @@ export interface IssueComment {
   reactions?: CommentReaction[] | null;
   // 已解决时间(resolve/unresolve);非空=该评论已标记为线程结论。后端一线程至多一条 resolved。
   resolved_at?: string | null;
+  // 评论附件:后端 list/timeline 端点按 comment 分组回填;其它端点不带 → 保持已有。
+  attachments?: Attachment[] | null;
 }
 
 export type TaskStatus =
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -209,6 +209,11 @@
   "activity": {
     "title": "Activity"
   },
+  "attach": {
+    "title": "Attachments",
+    "add": "Add attachment",
+    "empty": "No attachments"
+  },
   "toast": {
     "created": "Created",
     "saved": "Saved",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -209,6 +209,11 @@
   "activity": {
     "title": "活动"
   },
+  "attach": {
+    "title": "附件",
+    "add": "添加附件",
+    "empty": "暂无附件"
+  },
   "toast": {
     "created": "已创建",
     "saved": "已保存",
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -29,6 +29,7 @@ import {
   SmilePlus,
   Bell,
   BellOff,
+  Paperclip,
 } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
 import type {
@@ -36,6 +37,7 @@ import type {
   IssueComment,
   IssueSubscriber,
   TimelineEntry,
+  Attachment,
   TaskRun,
   IssueStatus,
   IssuePriority,
@@ -66,6 +68,7 @@ import {
   unresolveComment,
   listTimeline,
 } from "../api/collabApi";
+import { uploadAttachment } from "../api/attachmentApi";
 import { listRuns, rerunIssue, cancelTask } from "../api/runsApi";
 import AssigneePicker from "../ui/AssigneePicker";
 import LabelEditor from "../ui/LabelEditor";
@@ -128,6 +131,9 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
   const [editingId, setEditingId] = useState<string | null>(null);
   const [editDraft, setEditDraft] = useState("");
   const [busyRunId, setBusyRunId] = useState<string | null>(null); // 正在重跑的 task,防双击
+  const [pendingFiles, setPendingFiles] = useState<File[]>([]); // 评论输入区:待随评论提交的本地文件(发送时才上传)
+  const [uploading, setUploading] = useState(false); // issue 附件上传中
+  const [submitting, setSubmitting] = useState(false); // 评论提交中(含附件上传),防重复提交
 
   // 每次 reload 递增;异步响应回来前先比对 token,丢弃 issueId 原地切换后到达的旧请求结果,
   // 防止慢请求(issue A)在切到 B 后把 A 的数据写进 B 的视图(导航竞态)。
@@ -166,12 +172,13 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     if (!issue) return;
     try {
       const updated = await updateIssue(issue.id, p);
-      // PUT 响应不带 labels/reactions(仅 list/detail 端点回填);re-enrich 修回 assignee_name/
-      // project_name 等展示字段(按新值重算),labels/reactions 保留当前值,避免编辑后被清空。
+      // PUT 响应不带 labels/reactions/attachments(仅 list/detail 端点回填);re-enrich 修回
+      // assignee_name/project_name 等展示字段(按新值重算),labels/reactions/attachments 保留当前值,避免编辑后被清空。
       setIssue({
         ...(await enrichIssue(updated)),
         labels: updated.labels ?? issue.labels,
         reactions: updated.reactions ?? issue.reactions,
+        attachments: updated.attachments ?? issue.attachments,
       });
       onChanged?.();
     } catch (e) {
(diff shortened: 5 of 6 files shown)


OctoBoooot (approved):
## Review: file attachments on issues and comments (#643)

**Verdict:** Approve
Careful, well-engineered attachment PR. The security-critical surfaces — multipart upload, attachment rendering (XSS), and orphan-attachment prevention — are all handled correctly, and it adds the busy guards + field-preservation the recent loop PRs were missing. Verified each.

### What I verified
- **Upload (`attachmentApi.ts`)** — multipart via `FormData`, no manual `Content-Type` (axios sets the boundary; the classic mistake is overriding it — correctly avoided); auth/workspace/CSRF via the shared interceptor; `issue_id`/`comment_id` bound server-side. Clean.
- **Render (XSS-safe)** — `filename` is escaped JSX text (`<span>{a.filename}</span>` / `alt=`); `download_url` (server-issued signed URL, not filename-derived) used as `` / `<a href>` with `rel="noreferrer" target="_blank"`; image inline gated on server `content_type.startsWith("image/")`. No `dangerouslySetInnerHTML`; comment body still via sanitized `LoopMarkdown`.
- **Orphan prevention (the notable design fix)** — comment attachments use **File-hold**: `pendingFiles` keeps `File` objects local and uploads only in `submitComment` *after* the comment exists, bound by `commentId`; a cancelled/abandoned draft uploads nothing (no issue-level orphan). Issue attachments upload immediately with `issueId` (issue already exists). Body cites DB verification (comment_id set, issue_id null).
- **Submit ordering / guards** — `submitComment` has a `submitting` busy guard; creates the comment, then **clears input state before uploading attachments** so an attachment failure can't leave a re-submittable draft (no duplicate comments); per-file failure isolated (`attFailed`, no comment rollback); comment-creation failure keeps draft+files for retry. `uploadForIssue` guards `uploading` and unlocks before the fire-and-forget re-fetch (a failed re-fetch can't wedge the button).
- **`patch()` preserves attachments** — `attachments: updated.attachments ?? issue.attachments`. Notably this is the exact class of bug I missed on #640 (derive/PUT drops a field) — here it is correctly handled. Nav-race: `token` captured before await, reloads token-guarded (consistent with #637/#639).

### Praise
- The File-hold design for comment attachments is the right call: uploading only after the comment is created (bound by commentId) structurally prevents the orphan-attachment trap that issue-first binding creates, and it's DB-verified. Combined with create-then-clear-then-upload ordering + per-file error isolation, the failure modes (abandoned draft, attachment error, double-submit) are each handled deliberately — and this PR is also where the recurring busy-guard family note is finally addressed (submitting/uploading flags) *and* the #640-class field-preservation bug is correctly avoided for attachments.

Jerry-Xin (approved):
Clean file-attachment feature — upload API, XSS-safe filename render, server-issued download URLs, and (notably) a correct no-clobber save path. APPROVE.

Reviewed the isolated PR diff (merge-base stripped; +209/-12, 6 files): `attachmentApi.ts` (new upload client), `types.ts` (Attachment type + `attachment_ids` on create/update), `en-US`/`zh-CN` i18n, `IssueDetailPage.tsx` (upload + render), `issueDetail.css`.

### ✅ Highlights
- **Upload API correct** — `POST /upload-file` with `FormData` (`file` + optional `issue_id`/`comment_id`). Passed straight to axios, which auto-sets the `multipart/form-data` boundary; auth/workspace headers come from the interceptor. Clean.
- **Filename XSS-safe** — attachment filenames are user-controlled but render only as a React text child (`<span>{a.filename}</span>`) and as `alt={a.filename}`. No `dangerouslySetInnerHTML`, no interpolation into `href`/`src`/`title`. A filename like `">` renders as inert text. ✅
- **Download URL is server-issued, not client-built** — links use `a.download_url` (short-lived signed URL per the type contract), never a URL assembled from the filename, so no path-traversal / open-redirect from user input. `<a target="_blank" rel="noreferrer">` (noreferrer implies noopener). ✅
- **Inline image render is safe-ish** — `content_type.startsWith("image/")` renders ``. SVG (`image/svg+xml`) matches, but it goes through `` (not inlined), and browsers do not execute scripts in ``-referenced SVG; src is same-origin server-issued. No inline-SVG/HTML injection path.
- **No attachment-list clobber on empty save** (the exact regression class that slipped through elsewhere in this batch) — the issue `save` path writes `attachments: updated.attachments ?? issue.attachments`. When the PUT response omits attachments (which it does — attachments only come back on the detail endpoint), the `??` falls back to the **currently stored list**, not `[]`. So editing title/status/etc. does not wipe attachments. Reads go through `getIssue` (full detail). Correct.
- **Busy guards present on both write paths** — comment submit is guarded by `submitting` (`if (!content || submitting) return`, `loading={submitting}` on Send, file input `disabled={submitting}`); the issue-attachment input is guarded by `uploading` (`disabled={uploading}` + Spin). This is tighter than the recurring in-flight gap seen elsewhere in the family.
- **Comment-attach flow avoids orphans** — files are held locally in `pendingFiles` and uploaded with `commentId` only *after* the comment is created, so a cancelled/abandoned comment leaves no issue-level orphan attachment. Input state is cleared right after comment creation to prevent a double-submit of the same comment while attachments upload.
- **Nav-race handled** — `IssueDetailPage` is remounted via `key={id}`, and `uploadForIssue` captures the `reqRef` token before the await; `syncIssue` only applies the result if `token === reqRef.current`.
- i18n added in both locales; empty/loading/error states covered (Toast on failure, `attach.empty`, Spin).

### 💬 Non-blocking
- **No client-side type/size validation** on either file input (`accept` attr / size cap absent). Server enforcement is authoritative, so this is defense-in-depth / UX only — not a blocker. Worth adding an `accept` hint and a client size guard so users get immediate feedback instead of a round-trip failure.
- Comment-attachment partial failure sets `attFailed` and surfaces a generic `saveFailed` Toast but doesn't name which file failed; acceptable, but a per-file message would be friendlier.

No blocking issues. LGTM.

yujiawei (changes requested):
# Code Review — PR #643 (octo-web)

Reviewed at head SHA `761f083e264d9778809452bb5648723e173d8afa` against merge-base `d427808d` on `feat/octo-loop`. Scope: 6 files, +209/-12 (dmloop attachment feature). This PR is flagged security-sensitive, so upload/render surfaces were checked closely.

## Verdict: Changes Requested

The feature is well-scoped and the happy path is solid, but the **error handling around uploads is under-baked for a brand-new attachments feature**, and one case silently and permanently loses a user's selected file. That should be tightened before merge. No P0/security-blocker was found — the security concerns I chased down are covered by existing code/deploy config (see "Security notes" below).

---

## Spec compliance — OK

The diff implements exactly what the description states, nothing more:
- `uploadAttachment(file, {issueId?|commentId?})` → multipart `POST /upload-file` via the shared axios client (interceptor injects auth/workspace/CSRF headers; `Content-Type` correctly left to the browser for the boundary). ✅
- Issue attachments upload immediately with `issue_id` and re-fetch detail; comment attachments are held locally and uploaded with `comment_id` only on submit (no orphan issue-level attachments from abandoned drafts). ✅
- `patch()` preserves `attachments` on PUT the same way it preserves `labels`/`reactions` (`IssueDetailPage.tsx:181`). ✅
- No over-build: `Attachment` / `issue.attachments` / `comment.attachments` are all optional additions; list/board views don't consume them. ✅

No missing / extra / divergent behavior relative to the stated scope.

---

## Findings

### P1 — Failed comment-attachment uploads are silently dropped with no recovery
`packages/dmloop/src/panel/IssueDetailPage.tsx` (`submitComment`, ~L322–338)

```tsx
setCommentDraft("");
...
if (!replyTo) setPendingFiles([]);      // staged files cleared BEFORE uploads run
let attFailed = false;
for (const f of files) {
  try { await uploadAttachment(f, { commentId: comment.id }); }
  catch { attFailed = true; }           // failed File is discarded, not re-staged
}
await reloadComments(token);
if (attFailed) Toast.error(t("loop.toast.saveFailed"));
```

`pendingFiles` is cleared before the upload loop runs. If a file upload fails (e.g. backend size/type reject, network blip), the comment already exists but the file chips are gone, the `File` objects are discarded, and there is no edit-to-add-attachment path — so the attachment is **permanently lost from the created comment** with no way to retry the exact files. The only feedback is a generic `loop.toast.saveFailed` ("Save failed") which misleadingly implies the whole comment failed, when in fact it succeeded. For a feature whose entire purpose is attachments, silently discarding a user's selected file behind a wrong error message is a functional defect.

Suggested fix: on partial failure, keep the failed `File`s staged (re-populate `pendingFiles` with the ones that threw) and use a distinct message like "Comment posted, but N attachment(s) failed — retry" instead of the generic save-failed toast.

### P2 — `uploadForIssue` aborts the whole batch on first failure
`IssueDetailPage.tsx` (`uploadForIssue`, ~L272–276)

```tsx
try {
  for (const f of Array.from(files)) await uploadAttachment(f, { issueId });
} catch (e) {
  Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
}
```

Single try/catch around the awaited loop: if file #2 of 3 throws, file #3 is never attempted while #1 was already bound to the issue. This is inconsistent with `submitComment`, which deliberately isolates each upload with its own try/catch (`attFailed`). Apply the same per-file isolation here so one bad file doesn't drop the rest, and ideally report which files failed.

### P2 — No client-side validation (size / count / type) before upload
`IssueDetailPage.tsx` `addPendingFiles` (~L261) and `uploadForIssue` (~L268); the `<input type="file">` elements have no `accept` attribute.

Neither path enforces a max size, max count, or allowed types before hitting the network. This is defense-in-depth/UX, not a security control (the server must remain the real gate), but it makes the P1/P2 upload-failure paths far more likely to trigger (e.g. a large screenshot rejected by the backend → lost per the P1 above). A client-side size ceiling + `accept` hint would reduce wasted multipart uploads and give immediate feedback.

### P2 — Attachment-only comment cannot be sent
`IssueDetailPage.tsx` (`submitComment`, L305–306)

```tsx
const content = commentDraft.trim();
if (!content || submitting) return;
```

A user can stage files with no caption, but Send / Enter silently no-ops (the button is only `disabled={!!replyTo}`, not gated on content, so it's a dead click with no hint). Attaching a screenshot with no text is a common flow. Either allow submit when `pendingFiles` is non-empty, or disable/annotate Send so the dead click is explained.

### P2 — Image thumbnails from short-lived `download_url` with no `onError` fallback
`IssueDetailPage.tsx` `renderAttachments` (~L289–297)

`types.ts` documents `download_url` as a short-lived signed URL ("仅即时展示、勿持久化") and provides persistable `url`/`markdown_url`. `renderAttachments` binds both `` and `<a href>` to `download_url`, and these render from `issue.attachments` / `c.attachments` that live in React state for the page lifetime (only refetched on `issueId` change). On a long-idle detail page past the signed-URL TTL, thumbnails and download clicks 403/404 with no graceful degradation. Consider an `onError` fallback to the filename chip, or refresh the URL on click. Please confirm the backend TTL comfortably exceeds a typical view session.

### P2 (nit) — Index used as React key for a removable list
`IssueDetailPage.tsx` (~L830–835): `pendingFiles.map((f, i) => <span key={i} ...>`. No runtime bug today (rows are stateless display chips), but index keys on a list that supports removal are fragile if per-row state (upload progress, thumbnail) is ever added. Prefer a stable key like `${f.name}-${f.size}-${f.lastModified}` or an id assigned at add-time. (Duplicate filenames are also accepted unbounded — minor.)

---

## Security notes (checked and cleared)

Several plausible upload/render risks were investigated and found **not** exploitable in this diff:
- **`content_type`-driven inline `` / SVG XSS**: `` does not execute SVG script, and both the image and download branches render the identical `<a href={download_url} target="_blank" rel="noreferrer">`, so `content_type` only gates thumbnail preview, not a trust boundary. Actual safety depends on backend `Content-Type`/`Content-Disposition` on the signed URL.
- **Signed-URL leakage via `Referer`**: `nginx.conf.template` sets `Referrer-Policy: strict-origin-when-cross-origin` (locked by a test), so query-string signatures are stripped on cross-origin `` requests.
- **`target="_blank"` reverse tabnabbing**: `rel="noreferrer"` implies `noopener`; not an issue.

## For a human to verify (backend contract — not visible in this diff)
1. **`/upload-file` server-side enforcement**: content-type/size limits and malware handling (client-side validation is absent, so the server is the only gate).
2. **Access control on binding**: does the backend verify the caller may write to the target `issue_id` / `comment_id` when an attachment is bound? A frontend-supplied `issue_id`/`comment_id` in the multipart form must not let a user attach to arbitrary issues.
3. **Signed `download_url` TTL** length vs. typical page-open duration (see the P2 above).
4. **Response contract**: the entire render assumes `getIssue` returns `issue.attachments` (filtered by `comment_id==null`) and `listComments` returns per-comment `attachments`. Confirm those endpoints actually populate these fields, or the UI shows nothing.

Jerry-Xin (commented):
Byte-verified yujiawei's two findings at head `761f083e` — both are factually accurate; I calibrate them as non-blocking robustness/UX gaps (🟡) rather than 🔴 data-integrity, and note the honest distinction from the #640 wipe.

**Confirmed — issue-attach aborts remaining files on first error** (`IssueDetailPage.tsx:273`): `for (const f of Array.from(files)) await uploadAttachment(f, { issueId })` is a single try; if file N throws, files N+1… are never attempted. This is inconsistent with the comment path (`:329-336`) which *does* isolate per-file via `attFailed`. Fair critique. Severity 🟡: the files that already uploaded are persisted (each upload is an independent committed POST), the user gets a Toast, and un-uploaded files remain on local disk to re-select — degraded partial-batch UX, not loss of stored data.

**Confirmed — failed comment attachment is dropped with no retry** (`:326` clears `setPendingFiles([])` *before* the upload loop at `:328-336`): once the comment is created, staged files are cleared first, so a file whose `uploadAttachment` throws is already gone from `pendingFiles` — the user must re-select to retry. Fair critique. Severity 🟡: the comment itself and any successful attachments persist; only the failed attachment lacks a retry affordance. The clear-before-upload is deliberate (prevents duplicate-comment re-submit) — the gap is that it also discards the failed file.

**Why 🟡, not 🔴:** neither destroys previously-stored data. Contrast #640, where `save` wrote `description:""` and wiped a stored DB field the user never touched — that was a true data-loss 🔴. Here the "loss" is a not-yet-persisted local File that the user still holds and can re-select; the persisted state (comment, successful attachments) is intact. Real robustness gaps worth a fast-follow, but not blocking data-integrity.

**Suggested fast-follow:** (1) give `uploadForIssue` the same per-file isolation the comment path has (collect failures, continue the loop, report which failed); (2) in `submitComment`, keep the *failed* files in `pendingFiles` (clear only the succeeded ones) so a retry doesn't require re-selection; (3) her 🟡 client-side size/count/type pre-validation. My earlier APPROVE stands on security (XSS-safe render, server-issued signed download_url, multipart correctness, orphan-prevention, patch attachment-preservation) and the happy path; I under-called these two error-path gaps as merely "isolation exists on comments" — crediting yujiawei for catching both.

PR #644, opened 2026-07-10
feat(loop): @mention picker in comments with non-navigable chips

MrAladdin:
## What

Adds an **@ mention picker** to the Loop issue comment composer, and renders `mention://` links as non-navigable chips. Part of the MatterV2 capability-parity work (capability-layer-first, minimal UI).

### Composer
- New **"@ mention"** dropdown next to the comment box. Candidates come from the existing `useAssigneeCandidates` pipeline, grouped by member / agent / squad (reusing the `loop.assignee.*` i18n keys).
- Picking a candidate inserts a `[@Label](mention://type/id)` token into the draft. The token format matches octo-multica's `util.MentionRe` contract (`mention://(member|agent|squad|issue|all)/<uuid|all>`), so agent/squad mentions trigger runs on the backend and member/issue mentions link only.
- The label is stripped of `[` / `]` so a candidate name cannot corrupt the markdown link syntax.
- Disabled while replying (the reply composer is a separate scoped input).

### Rendering (`LoopMarkdown`)
- `mention://` links render as a non-navigable `.loop-mention` chip (no href, no navigation).
- **react-markdown@8's default `transformLinkUri` only allowlists http/https/mailto/tel** and would rewrite `mention://` to `javascript:void(0)` before the custom `a` renderer sees it — silently breaking the whole feature. A custom `transformLinkUri` passes `mention:` through while delegating every other scheme to the built-in `uriTransformer`, so **XSS sanitization of user-authored links is preserved** (a plain `transformLinkUri={false}` would have re-opened that hole).

## Verification (dev env, MV2 E2E workspace / MVE-4)

Real-browser e2e against the live bundle:
- `@ mention` → E2E Coworker (agent) inserts `[@E2E Coworker](mention://agent/<uuid>)`; the sent comment renders a `<span class="loop-mention">` chip (not a dead anchor). `将唤醒` preview reflects the agent.
- Regression: `[link](https://example.com/x)` still renders as `<a href="https://example.com/x" target="_blank">`.
- Security: `[xss](javascript:alert(1))` is neutralized to `href="javascript:void(0)"`.

## Review gate
- `/simplify` (4 agents): the one recurring finding (candidate-group render duplicated with the existing assign menu) was deliberately **kept** — the two consumers have diverged (assign has active/Divider, mention is plain) and real dedup would require editing the already-merged assign menu (out of this diff's blast radius); the altitude pass judged extracting-now to be over-abstraction (YAGNI). New block is same-shaped, so a future 3-site refactor stays easy.
- Dual-model adversarial review (codex + Claude code-reviewer). Claude caught the `transformLinkUri` neutralization bug above; both are fixed and verified.

## Tests
No unit test added: `packages/dmloop` currently has **no test harness** (no vitest config / script / test files anywhere in the package). Bootstrapping one for a single component is disproportionate infra scope for this feature PR; the changed branch is instead exercised end-to-end in the real runtime (above). A dmloop test harness is tracked as separate infra work.

## Blast radius
Scoped to `packages/dmloop`: `LoopMarkdown.tsx` (shared markdown renderer — the added `transformLinkUri`/chip branch is backward-compatible: non-mention links behave exactly as before), `IssueDetailPage.tsx` composer, two i18n files (additive `mention.add` key), one CSS rule. No shared type/prop/API signature changed; no consumer outside dmloop is affected.

Closes part of MatterV2 collaboration parity (mention capability, item 2 of subscribe-toggle + mention).

### DIFF
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -312,6 +312,9 @@
   "reaction": {
     "add": "Add reaction"
   },
+  "mention": {
+    "add": "Mention"
+  },
   "agent": {
     "runtime": "Runtime",
     "model": "Model",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -312,6 +312,9 @@
   "reaction": {
     "add": "添加反应"
   },
+  "mention": {
+    "add": "@ 提及"
+  },
   "agent": {
     "runtime": "运行环境",
     "model": "模型",
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -30,6 +30,7 @@ import {
   Bell,
   BellOff,
   Paperclip,
+  AtSign,
 } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
 import type {
@@ -38,6 +39,7 @@ import type {
   IssueSubscriber,
   TimelineEntry,
   Attachment,
+  AssigneeCandidate,
   TaskRun,
   IssueStatus,
   IssuePriority,
@@ -254,6 +256,15 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     } catch (e) {
       Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
     }
+  };
+
+  // @提及:选中候选(成员/AI队友/AI小队)→ 往草稿插入 mention markdown。
+  // 后端 util.MentionRe 认 [@Label](mention://<type>/<id>);插入草稿后 previewCommentTriggers 会反映"将唤醒"。
+  const insertMention = (c: AssigneeCandidate) => {
+    // label 剥掉 []:名字里的方括号会破坏 markdown 链接语法 [label](url);id 才是真引用,label 仅显示。
+    const label = c.name.replace(/[[\]]/g, "");
+    const token = `[@${label}](mention://${c.type}/${c.id})`;
+    setCommentDraft((d) => (d && !d.endsWith(" ") ? d + " " : d) + token + " ");
   };
 
   // 评论附件:本地持有 File,发送时才带 commentId 上传绑定(见 submitComment),
@@ -844,6 +855,29 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
                 placeholder={replyTo ? t("loop.comment.replyingHint") : t("loop.comment.placeholder")}
                 onEnterPress={submitComment}
               />
+              <Dropdown
+                trigger="click"
+                clickToHide
+                position="topRight"
+                render={
+                  <Dropdown.Menu>
+                    {(["member", "agent", "squad"] as const).map((type) => {
+                      const items = cands.filter((c) => c.type === type);
+                      if (!items.length) return null;
+                      return (
+                        <React.Fragment key={type}>
+                          <Dropdown.Title>{t(`loop.assignee.${type}`)}</Dropdown.Title>
+                          {items.map((c) => (
+                            <Dropdown.Item key={c.id} onClick={() => insertMention(c)}>{c.name}</Dropdown.Item>
+                          ))}
+                        </React.Fragment>
+                      );
+                    })}
+                  </Dropdown.Menu>
+                }
+              >
+                <Button theme="borderless" icon={<AtSign size={16} />} disabled={!!replyTo} aria-label={t("loop.mention.add")} />
+              </Dropdown>
               <label className="loop-attach-btn" aria-label={t("loop.attach.add")} style={{ opacity: replyTo ? 0.4 : 1 }}>
                 <Paperclip size={16} />
                 <input
--- a/packages/dmloop/src/ui/LoopMarkdown.tsx
+++ b/packages/dmloop/src/ui/LoopMarkdown.tsx
@@ -1,7 +1,12 @@
 import React from "react";
-import ReactMarkdown from "react-markdown";
+import ReactMarkdown, { uriTransformer } from "react-markdown";
 import remarkGfm from "remark-gfm";
 import "./markdown.css";
+
+// react-markdown@8 默认只放行 http/https/mailto/tel，会把 mention:// 改写成 javascript:void(0)。
+// 放行 mention:，其余仍走默认清洗（保住对用户输入的 XSS 防护）。
+const transformLinkUri = (href: string) =>
+  href.startsWith("mention://") ? href : uriTransformer(href);
 
 /** Loop Markdown 渲染：标题/列表/代码块/行内代码/链接/表格/引用等美化展示。 */
 export default function LoopMarkdown({ content }: { content: string }) {
@@ -9,8 +14,15 @@ export default function LoopMarkdown({ content }: { content: string }) {
     <div className="loop-md">
       <ReactMarkdown
         remarkPlugins={[remarkGfm]}
+        transformLinkUri={transformLinkUri}
         components={{
-          a: ({ node, ...props }) => <a target="_blank" rel="noreferrer" {...props} />,
+          a: ({ node, href, children, ...props }) => {
+            // mention 链接 [@Label](mention://type/id):渲染为不可导航的高亮 chip(点击无跳转)。
+            if (href && href.startsWith("mention://")) {
+              return <span className="loop-mention">{children}</span>;
+            }
+            return <a href={href} target="_blank" rel="noreferrer" {...props}>{children}</a>;
+          },
         }}
       >
         {content || ""}
--- a/packages/dmloop/src/ui/markdown.css
+++ b/packages/dmloop/src/ui/markdown.css
@@ -55,3 +55,10 @@
 .loop-md th { background: var(--semi-color-fill-0, #f5f6f8); font-weight: 600; }
 .loop-md hr { border: none; border-top: 1px solid var(--semi-color-border, #e8e9eb); margin: 16px 0; }
 .loop-md img { max-width: 100%; border-radius: 6px; }
+
+/* @提及 chip(mention:// 链接,不可导航;PR @提及) */
+.loop-mention {
+  color: var(--semi-color-primary, #0064fa);
+  background: var(--semi-color-primary-light-default, #e8f0ff);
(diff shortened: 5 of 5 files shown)


OctoBoooot (approved):
## Review: @mention picker in comments with non-navigable chips (#644)

**Verdict:** Approve
Small, security-conscious PR. The one real trap — preserving link sanitization while allowlisting `mention://` — is handled correctly, and the composer protects markdown syntax. Adversarially verified the sanitization on hostile schemes.

### What I verified (the sanitization surface — adversarial)
- **`transformLinkUri` preserves XSS defense.** `href.startsWith("mention://") ? href : uriTransformer(href)` — only the exact `mention://` prefix bypasses; **every other scheme still goes through react-markdown's built-in `uriTransformer`**. Constructed the hostile cases: `javascript:alert(1)` → not `mention://` → `uriTransformer` neutralizes to `javascript:void(0)` (✓); `data:`/`vbscript:` → neutralized (✓); `https`/`mailto`/`tel` → pass as normal `<a target=_blank rel=noreferrer>` (✓). Critically, this is **not** the `transformLinkUri={false}` trap the description flags (which would have disabled sanitization globally).
- **The mention value never reaches an href sink.** A `mention://` href is rendered as a **text-only `<span class="loop-mention">` chip** (no `href` attribute), so even a crafted `mention://javascript:...` string is inert text — no bypass through the mention branch.
- **Composer insert is injection-safe.** `insertMention` strips `[`/`]` from the candidate name (`c.name.replace(/[[\]]/g,"")`) so a bracket-bearing name can't break out of the `[@label](url)` syntax; `type` is the fixed `member|agent|squad` enum and `id` is a server-issued candidate id — neither is free-text. Candidate names in the dropdown render as escaped JSX text; disabled while replying.

### Praise
- The `transformLinkUri` handling is exactly right for a "allowlist one custom scheme in a markdown renderer" change — the naive fix (`transformLinkUri={false}` or a custom transformer that returns `href` unconditionally) silently re-opens `javascript:` XSS for *all* user links, and the PR both avoids it and delegates every non-mention scheme back to the built-in sanitizer. Pairing that with rendering mentions as href-less chips means the custom scheme never touches a URL sink at all — defense in depth on the exact surface where markdown-link features usually introduce XSS.

### Note
No unit test — `packages/dmloop` has no test harness (consistent with the package); e2e-verified instead, incl. the `javascript:` neutralization regression. Bootstrapping a harness is reasonably out of scope for this PR; worth tracking as infra.

Jerry-Xin (approved):
**APPROVE** — @mention picker + non-navigable chips. Clean; the XSS/injection surface is handled correctly on every path I traced. Reviewed against the isolated own-diff (`9ebe1d6f..a2c391a0`), 5 files / +61/-2.

### ✅ Highlights (byte-verified)
- **Chip render is escaped text, no raw HTML.** `LoopMarkdown.tsx:24` renders mentions as `<span className="loop-mention">{children}</span>`, where `children` are react-markdown-parsed text nodes (escaped). `git grep dangerouslySetInnerHTML` over `packages/dmloop/` returns **NONE**, and no `rehype-raw`/`rehypePlugins` is configured. A display name like `` renders as inert text.
- **Non-navigable is real, and it also closes an XSS hole.** For `mention://` hrefs the component returns a `<span>`, never an `<a href>` — so the `mention://…` URL is discarded and can never become an href/`javascript:` sink. Non-mention links still go through `uriTransformer` (`transformLinkUri`, `LoopMarkdown.tsx:9-11`) + `rel="noreferrer" target="_blank"`, so allowing `mention://` does not weaken sanitization of ordinary user links.
- **No display-name breakout / uid injection.** Token format is `[@${label}](mention://${c.type}/${c.id})` (`IssueDetailPage.tsx:266`). `label = c.name.replace(/[[\]]/g, "")` strips the only chars that delimit markdown link text (`[`,`]`), so the label can't close early and open a second link. The actual reference lives in the URL as the **server-issued** `c.type`/`c.id`; the user-controlled name sits only inside the label region, where stray `)`/`:`/`>` are literal text and cannot rewrite the uid. Render resolves the reference from the URL, display from the label — the safer of the two designs.
- **No index-pairing collapse (per prior batch lesson).** Picker items map `onClick={() => insertMention(c)}` carrying each candidate object directly — no positional pairing between picked users ↔ chips ↔ tokens, so no filter/dedupe can mis-map to a wrong uid.
- **Family consistency intact.** Double-submit guard untouched and correct (`if (!content || submitting) return;` + `setSubmitting(true)` / `finally setSubmitting(false)`); the `@` button is `disabled={!!replyTo}`, matching the attachment button's reply-mode behavior. i18n `mention.add` present in both en-US and zh-CN. Empty candidate groups render nothing (`if (!items.length) return null`).

### 💬 Non-blocking notes
- `transformLinkUri` only special-cases the literal prefix `mention://`; anything else falls back to `uriTransformer`. Fine today. If a future token variant emits a bare-scheme `mention:` (no `//`), it would be sanitized away — worth a comment if the token format ever changes.
- Picker lists all `AssigneeCandidate`s (member/agent/squad). Authz is inherited from `listAssigneeCandidates` (same scope already used for assignment), so no new leak surface — noting it since mentionable-scope was a listed risk.

No blocking items. LGTM.

yujiawei (approved):
# Code Review — PR #644 (octo-web)

**Scope:** `feat(loop): @mention picker in comments with non-navigable chips` — 5 files, +61/-2, frontend-only. Reviewed at head SHA `a2c391a0a3180bc6ee2d0d681ae2050c4e087371` against base `feat/octo-loop`. This PR was flagged for extra security scrutiny because it touches the markdown URL-scheme sanitizer.

## Verdict: Approve

No blocking (correctness/security) issues found. The change is well-scoped and the security-relevant part (the `mention://` sanitizer passthrough) is safe as implemented. A few non-blocking hardening suggestions and cross-repo verification notes are below.

---

## 1. Spec compliance

The title's two deliverables are both implemented, with nothing extra and nothing missing:

- **@mention picker** — `IssueDetailPage.tsx:858-880`: an `AtSign` `Dropdown` lists `member`/`agent`/`squad` candidates and inserts a mention token into the comment draft via `insertMention` (`:263-268`).
- **Non-navigable chips** — `LoopMarkdown.tsx:19-25`: `mention://` links render as an inert `<span class="loop-mention">` (no `href`, no `onClick`); all other links keep the original `<a target=_blank rel=noreferrer>` behavior.

Supporting changes are minimal and appropriate: `loop.mention.add` i18n keys and the `.loop-mention` chip style. No over-build, no scope creep.

## 2. Security analysis (the load-bearing part)

The change adds `transformLinkUri` (`LoopMarkdown.tsx:8-9`) that lets `mention://` hrefs bypass react-markdown@8's default `uriTransformer`. I verified this is **not exploitable**:

1. **Default sanitizer preserved for everything else** — only `href.startsWith("mention://")` bypasses; every other URL still flows through `uriTransformer`, so `javascript:`/`data:`/`vbscript:` are still neutralized to `javascript:void(0)`.
2. **The bypassed value never reaches a navigable sink** — a `mention://` href is rendered as a `<span>` with no `href`/`onClick`, so the un-sanitized string is inert in the DOM.
3. **No raw-HTML path** — no `rehype-raw` / `allowDangerousHtml` / `dangerouslySetInnerHTML` is configured, so react-markdown@8's default HTML-escaping remains in force.
4. **Even a degenerate `mention://…` payload is harmless** — `mention` is an unknown scheme; browsers do not execute it (it is not `javascript:`), and it renders as a span regardless.
5. **Label injection is closed** — `insertMention` strips `[`/`]` from the display label (`:265`), preventing a `](` breakout of the `[label](url)` syntax.

`react-markdown` is pinned to `8.0.7` (package.json + lockfile), where both the named `uriTransformer` export and the `transformLinkUri` prop exist — the API usage is correct for the pinned version.

## 3. Non-blocking suggestions (P2)

- **Extract a single `isMention(href)` helper.** The security invariant relies on two duplicated string literals staying in sync — `transformLinkUri` (`:9`) and the `<a>` renderer (`:21`). They are symmetric today, so it is safe. But if a future edit relaxes one side (e.g. makes one case-insensitive or trims whitespace) without the other, a value could skip the sanitizer *and* fall through to the live `<a href={href}>` branch. A shared helper makes the invariant enforceable. Worst realistic regression is still a broken chip rather than XSS, but the coupling is worth removing.
- **Document / test the "inert span" invariant.** `LoopMarkdown` also renders issue descriptions and skill files (`SkillFileViewer.tsx:75`, `IssueDetailPage.tsx:734`), which may be user- or AI-authored. The safety of the scheme passthrough is load-bearing and undocumented; a one-line comment asserting "mention:// must always render non-navigable" plus a unit test that a `mention://` href never produces an `href` attribute would prevent an accidental future regression.
- **Case-sensitivity of the guard.** `startsWith("mention://")` is case-sensitive, so `MENTION://a/b` renders as an inert `javascript:void(0)` anchor (styled as a link, not a chip) rather than as a mention chip. Harmless (no XSS), but an inconsistent render for case-variant input. Only relevant if the backend `MentionRe` ever accepts non-lowercase forms.
- **Mention picker while replying.** The `AtSign` Button is `disabled={!!replyTo}`, but it is the trigger of a `Dropdown`; with the Semi wrapper, disabling the child Button does not always suppress the wrapper's click. If the menu opens during a reply, `insertMention` mutates the shared `commentDraft`, which the reply `Input` also renders — so a token could land in the in-progress reply. Sibling controls are more defensive (the attachment `<input>` is itself disabled; the top comment `Input` blanks its value with `value={replyTo ? "" : commentDraft}`). Consider not rendering / gating the mention Dropdown when `replyTo` is set. Impact is cosmetic and user-reversible, so non-blocking.

## 4. Items worth a human confirming (out of this repo's scope)

These are not defects in this diff — the mention-token contract and the wake pipeline live in the backend, which is not in this repo. Flagging for the security-sensitive review:

- **Backend `mention://<type>/<id>` parser.** The frontend emits `[@Label](mention://type/id)` and defers "who will be woken" to the live `previewCommentTriggers` endpoint (so a contract mismatch would be visible in the preview, not silent). Still worth confirming the server's parser matches this exact shape and dispatches per `type`.
- **Reply-mode wake path.** In reply mode the picker and the wake-preview are both disabled, but the reply `Input` is still editable — a hand-typed/pasted `mention://` token in a reply would be submitted with no preview and no suppression UI. Confirm whether the backend parses and wakes on reply comments (`parent_id` set); if so, that is an invisible wake path worth surfacing in the UI.
- **Authorization on wake.** The server should authorize wakes by the mentioned id, not trust the client picker (a comment body can contain any `mention://` token by hand — this is unchanged by this PR, but is the relevant trust boundary).

PR #645, opened 2026-07-10
feat(loop): state-aware subscribe toggle in issue detail

MrAladdin:
## What

Turns the issue subscribe control into a **real state-aware toggle**. Frontend half of Option A (backend: octo-multica MR !22, merged).

- `IssueSubscriber` gains `octo_uid?: string | null` (member subscribers only; null for agent/squad).
- The detail page derives `amSubscribed` by matching `WKApp.loginInfo.uid` against member subscribers' `octo_uid`.
- The "more" menu previously showed **both** "subscribe" and "unsubscribe" unconditionally (the client couldn't tell the current member's state). It now renders a single item — subscribe when not subscribed, unsubscribe when subscribed (`toggleSubscribe(!amSubscribed)`).
- The properties sidebar shows a "已订阅" indicator next to the subscriber count when the current member is subscribed.

## Why this identity resolution (Option A)

The backend surfaces `octo_uid` on the subscriber row so the web compares it to the logged-in octo uid directly. This survives multica de-bridge into octo — no frontend member↔multica-user id inversion (which would break then). Confirmed by review: the uid chain is `WKApp.loginInfo.token → octoResolver.Resolve → user.octo_uid → ListIssueSubscribers join → octo_uid`, i.e. `loginInfo.uid` and `subscriber.octo_uid` are the same namespace.

Agent/squad subscribers have `octo_uid === null` and never match; `!!myUid` guards the logged-out case — so neither is mistaken for the current member.

Real-browser round-trip against the live backend serving `octo_uid`:
- Initial (Alice subscribed): sidebar shows "· 已订阅", subscriber count 4, menu shows "取消订阅".
- Click unsubscribe → indicator gone, count 3, menu flips to "订阅此 Issue".
- Click subscribe → indicator back, count 4, menu flips to "取消订阅" (state restored).

## Review gate

- `/simplify` (4 dims): applied — dropped a redundant `?.` on `WKApp.loginInfo` to match the repo convention (it's a never-null singleton; `!!myUid` covers the empty-string case). Ternary-vs-single-item was a wash, left explicit.
- Claude code-reviewer (skeptical): traced the full uid namespace chain, amSubscribed recompute timing, on/off semantics, null boundaries, consumer impact — no real bug.
- codex adversarial-review: the codex model stream was persistently dropping mid-turn this session (infra flakiness, `stream closed before response.completed`); on this feature's backend MR two codex runs did complete with "approve / no material findings" after reading the diff. The correctness gate is carried by the Claude reviewer + e2e here.

## Tests

No unit test added: `packages/dmloop` has no test harness (no vitest config/script/tests in the package); bootstrapping one for this is disproportionate infra scope. The change is exercised end-to-end in the real runtime (above). Tracked as separate infra work.

## Blast radius

`packages/dmloop` only. `IssueSubscriber` gains one optional field (additive; existing readers — the count, initial load — unaffected). `IssueDetailPage` swaps two static menu items for one derived item and adds a sidebar indicator; no shared type/prop/hook signature changed, no consumer outside dmloop touched.

## Docs

No doc change: additive optional type field + UI-only toggle behavior; no new service/command/config/env, dev tables unchanged.

### DIFF
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -220,6 +220,9 @@ export interface IssueSubscriber {
   user_id: string;
   reason: string;
   created_at: string;
+  /** octo IM uid the member is bridged to; null for agent/squad subscribers.
+   *  Lets the web tell whether the current octo member is subscribed. */
+  octo_uid?: string | null;
 }
 
 /** 附件(上传返回 + issue/comment 详情回填)。
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -211,7 +211,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
       .catch(() => {});
   };
 
-  // 订阅/取消订阅(后端默认操作调用者本人、幂等);两项都常驻,不猜"我是否已订阅"。
+  // 订阅/取消订阅(后端默认操作调用者本人、幂等)。
   const toggleSubscribe = async (on: boolean) => {
     const token = reqRef.current;
     try {
@@ -223,6 +223,12 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
       Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
     }
   };
+
+  // 当前 octo 成员是否已订阅:subscriber 现在带 octo_uid(仅 member 有),与
+  // loginInfo.uid 比对即可判定,无需前端反查 member↔user 映射(去桥后仍成立)。
+  const myUid = WKApp.loginInfo.uid;
+  const amSubscribed =
+    !!myUid && subscribers.some((s) => s.user_type === "member" && s.octo_uid === myUid);
 
   // 评论 emoji 反应:选择器点 emoji=加,已有 chip 点击=删自己那条(后端按 actor+emoji 定位)。
   const reactComment = async (commentId: string, emoji: string, add: boolean) => {
@@ -524,12 +530,15 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
         <Dropdown.Item onClick={(e) => e.stopPropagation()}>{t("loop.menu.changeAssignee")}</Dropdown.Item>
       </Dropdown>
       <Dropdown.Divider />
-      <Dropdown.Item icon={<Bell size={13} />} onClick={() => toggleSubscribe(true)}>
-        {t("loop.subscribe.subscribe")}
-      </Dropdown.Item>
-      <Dropdown.Item icon={<BellOff size={13} />} onClick={() => toggleSubscribe(false)}>
-        {t("loop.subscribe.unsubscribe")}
-      </Dropdown.Item>
+      {amSubscribed ? (
+        <Dropdown.Item icon={<BellOff size={13} />} onClick={() => toggleSubscribe(false)}>
+          {t("loop.subscribe.unsubscribe")}
+        </Dropdown.Item>
+      ) : (
+        <Dropdown.Item icon={<Bell size={13} />} onClick={() => toggleSubscribe(true)}>
+          {t("loop.subscribe.subscribe")}
+        </Dropdown.Item>
+      )}
       <Dropdown.Divider />
       <Dropdown.Item type="danger" icon={<Trash2 size={13} />} onClick={handleDeleteIssue}>
         {t("loop.menu.deleteIssue")}
@@ -1014,6 +1023,11 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
               <dt>{t("loop.subscribe.label")}</dt>
               <dd>
                 <Text type="tertiary" style={{ fontSize: 12 }}>{subscribers.length}</Text>
+                {amSubscribed && (
+                  <Text style={{ fontSize: 12, marginLeft: 6, color: "var(--semi-color-primary)" }}>
+                    · {t("loop.subscribe.subscribed")}
+                  </Text>
+                )}
               </dd>
             </dl>
           </div>


OctoBoooot (approved):
## Review: state-aware subscribe toggle in issue detail (#645)

**Verdict:** Approve
Clean closure of the exact limitation I flagged on #636 (frontend couldn't tell subscription state, so both menu items were shown). The backend now surfaces `octo_uid` on subscriber rows and the frontend matches it to the logged-in uid. Verified the identity-namespace correctness and the guards.

### What I verified
- **Identity match is the right namespace.** `amSubscribed = !!myUid && subscribers.some(s => s.user_type === "member" && s.octo_uid === myUid)`, `myUid = WKApp.loginInfo.uid`. `octo_uid` is the octo IM uid the member is bridged to (same field family as `WorkspaceMember.octo_uid` — "UI renders member identity from octo by this uid"), so comparing it to `loginInfo.uid` is the correct, de-bridge-safe match (no member↔multica-user id inversion). Matches the documented chain `loginInfo.token → octoResolver.Resolve → user.octo_uid → subscriber.octo_uid`.
- **No misattribution.** `!!myUid` blocks the logged-out/empty-string case; `user_type === "member"` + `octo_uid === myUid` means agent/squad subscribers (octo_uid null) can never match the current member.
- **Toggle correctness.** The menu renders a single derived item (`amSubscribed ? unsubscribe(false) : subscribe(true)`) instead of both. `amSubscribed` recomputes inline from `subscribers`, which `toggleSubscribe` refetches (token-guarded) after the op — so the item flips correctly. The idempotent backend still underpins it, so even a momentarily-stale `amSubscribed` can't cause a wrong-state write.
- **Indicator** — `· 已订阅` gated on `amSubscribed`, token color (`--semi-color-primary`), escaped text. Additive optional `IssueSubscriber.octo_uid`; existing readers (count, load) unaffected.

### Praise
- This resolves the #636 idempotent-both-items design *properly* — rather than guessing membership client-side (which #636 correctly refused to do without a signal), it waited for the backend to surface `octo_uid` and matches on the identity namespace that survives de-bridge. And it keeps the idempotent backend as the safety net, so the new state-awareness is a pure UX gain layered over a still-safe operation — the right way to add a stateful toggle without introducing a wrong-state-write risk.

Jerry-Xin (approved):
State-aware subscribe toggle looks correct and safe. The menu now derives the current member's subscription state from the `subscribers` list (new `octo_uid` field matched against `WKApp.loginInfo.uid`) and renders exactly one action accordingly, instead of the previous always-both entries.

**Correctness verification**
- **State derivation (absent-source enumeration):** `amSubscribed = !!myUid && subscribers.some(s => s.user_type === "member" && s.octo_uid === myUid)`. During `reload()` the list is reset to `[]` before `listSubscribers` resolves, so while loading/empty/on-fetch-failure `amSubscribed` falls to `false` → the menu shows **Subscribe**. This is the safe default: the backend subscribe call is idempotent (defaults to caller), so a transient wrong label at worst re-subscribes a no-op, and the post-action refetch corrects the UI. No wrong destructive action can be triggered from a stale/empty source. Null-safe via `!!myUid`; agent/squad subscribers (null `octo_uid`) are correctly excluded by the `user_type === "member"` predicate.
- **Branch not inverted:** `amSubscribed ? toggleSubscribe(false) /*unsubscribe*/ : toggleSubscribe(true) /*subscribe*/` — correct mapping.
- **Refetch, not optimistic:** `toggleSubscribe` awaits the mutation then re-fetches `listSubscribers` and `setSubscribers(s)` under a `reqRef` token guard, so `amSubscribed` recomputes from fresh server truth. No optimistic local flip → no rollback needed; a failed request leaves state untouched.
- **Error surfaced:** failure path is `Toast.error(...)`, no silent swallow.
- **nav-race:** mutation refetch is gated by `token === reqRef.current`, consistent with the rest of the component and `reload()`'s `fresh()` token discipline.
- **i18n:** all four keys present in both locales — `loop.subscribe.{subscribe,unsubscribe,subscribed,unsubscribed}` (en-US + zh-CN). The new derived `· Subscribed` badge in the meta panel uses `loop.subscribe.subscribed`, also present.
- **types.ts:** `IssueSubscriber.octo_uid?: string | null` added with an accurate doc comment; optional, non-breaking.

**Non-blocking**
- 🟡 `toggleSubscribe` has no in-flight lock, so rapid double-clicks can fire overlapping subscribe/unsubscribe requests. Low impact (backend idempotent + post-action refetch reconciles), and consistent with the existing family grade — noting for parity, not blocking.

**Highlights**
- Clean derivation that avoids a frontend member↔user reverse lookup by piggybacking `octo_uid` onto the subscriber payload; comment explains the de-bridge rationale well.
- Single-action menu removes the prior "guess" ambiguity; the meta-panel `· Subscribed` badge is a nice confirmation affordance.

LGTM.

yujiawei (approved):
# Code Review — PR #645 (octo-web)

**Scope:** `packages/dmloop` only — 2 files, +24/-7. Reviewed at head `d23884f9637cbf4de848e51254f1ddcbabfcf5df` against base `feat/octo-loop`.

## Summary

Turns the issue subscribe control from two always-present menu items into a real state-aware toggle, driven by a new optional `octo_uid` field on `IssueSubscriber`. The change is small, additive, and internally consistent with existing patterns in the package.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. All three described behaviors are implemented — the `octo_uid` field, the `amSubscribed` derivation, the single-item toggle, and the sidebar "Subscribed" indicator.
- **Over-build:** none. No extra flags, endpoints, or state; the field is a single optional addition.
- **Deviation:** none. The uid comparison matches the documented resolution chain (`WKApp.loginInfo.uid` ↔ `subscriber.octo_uid`, same octo IM namespace).

## 2. Code quality

**Quality: Approved** — no P0/P1.

Verified against the checked-out source:

- **uid namespace is correct.** `WKApp.loginInfo.uid` is the octo IM uid (`packages/dmworkbase/src/App.tsx`: `WKSDK.shared().config.uid = WKApp.loginInfo.uid`, and `message.fromUID === WKApp.loginInfo.uid`). `subscriber.octo_uid` is the same namespace (backend join). Comparing them directly is sound and avoids a fragile client-side member↔user inversion. The `octo_uid?: string | null` shape follows the established convention already used in `directory.ts`, `AssigneePicker.tsx`, and `SettingsPage.tsx`.
- **Null/logged-out handling is safe.** `!!myUid` guards the empty/logged-out case; agent/squad subscribers carry `octo_uid === null` and correctly never match. `subscribers` is initialized to `[]` (`useState<IssueSubscriber[]>([])`), so `.some()` is never called on null. `WKApp.loginInfo` is a never-null static singleton (`static loginInfo: LoginInfo = new LoginInfo()`), so dropping the optional chain here is consistent with the rest of the codebase.
- **Recompute timing is correct.** `amSubscribed` is derived on each render from `subscribers` state, which is refreshed via `listSubscribers` after every `toggleSubscribe`, so the menu/indicator flip in lockstep with the mutation.
- **i18n key exists.** `loop.subscribe.subscribed` is present in both `en-US.json` and `zh-CN.json`.
- **Blast radius contained.** The new field is additive/optional — existing readers (the count, initial load) are unaffected, and no consumer outside `dmloop` touches `IssueSubscriber`.

### Suggestions (non-blocking, P2 / nit)

- **Native (non-octo) member edge case.** `amSubscribed` requires `octo_uid === myUid`, so a subscribed member whose `octo_uid` is null would render "Subscribe" rather than "Unsubscribe". On the octo-web surface this is effectively unreachable — the logged-in user authenticated via octo and therefore always has an octo uid, and their own subscriber row would carry that same uid. The toggle action is server-authoritative and idempotent, so even in the theoretical mismatch there is no data corruption (worst case: a redundant subscribe call). Worth a one-line comment noting the octo-identity-first assumption, but not a blocker.

### Security note (PR flagged security-sensitive)

- The client-visible surface adds only `octo_uid`, which is already exposed to the same clients elsewhere (member directory, assignee avatars). No new secret/token/PII is introduced.
- The subscribe/unsubscribe action remains server-side and defaults to the calling user; `amSubscribed` is purely presentational, so a spoofed client value cannot subscribe/unsubscribe anyone else. Trust boundary is intact.
- The new indicator renders a static i18n string with no user-controlled interpolation — no XSS surface.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No blocking issues.

## Notes on review coverage

Reviewed via multiple independent lenses (spec, correctness, security/privacy, cross-file quality) against the checked-out source at the head SHA; findings were adversarially verified. One additional advisory reviewer flagged a potential undeclared `user_type` and missing optional-chaining — both refuted against the actual code (`user_type: string` is declared on `IssueSubscriber`; `subscribers` defaults to `[]` and `loginInfo` is a never-null singleton). One further adversarial leg was unavailable this session due to upstream gateway errors and is recorded as absent (not a silent pass). No test was added because `packages/dmloop` has no test harness; the change is UI-only and exercised end-to-end — reasonable, though a package-level test harness remains worthwhile follow-up infra work.

PR #646, opened 2026-07-11
fix(loop): keep unsubscribe reachable when subscription state is unknown

MrAladdin:
## What

Follow-up fix to **#645** (state-aware subscribe toggle). #645's single toggle drove the menu off `amSubscribed`, which collapses "unknown" into `false`. An already-subscribed member could lose the unsubscribe affordance when self-subscription can't be proven:

- the `/subscribers` fetch fails (network) — `subscribers` stays empty, error is swallowed; or
- **octo-web ships ahead of octo-multica MR !22** — the response carries no `octo_uid`, so no member matches.

In both cases the menu showed only "subscribe" and the user was **stuck subscribed** to notifications. (Found by codex adversarial review during the #645 follow-up gate; the pre-#645 menu always showed both actions, so this is a regression.)

## Fix — tri-state self-subscription

- Track `subLoaded` (true only after `listSubscribers` resolves; reset per issue, guarded by the existing `reqRef`/`fresh()` token).
- Detect `octoUidSkew` = member subscribers present but none carry `octo_uid` (backend not surfacing it yet — the deploy-skew window).
- `selfKnown = subLoaded && !octoUidSkew && !!myUid`. When **not** known, the menu falls back to showing **both** subscribe + unsubscribe (the pre-#645 safe behavior) so unsubscribe stays reachable; otherwise the single state-aware toggle is shown.
- Key invariant (verified by review): any uncertain state forces `amSubscribed=false`, so the menu branches and the "已订阅" sidebar indicator can share the same derived vars without disagreeing.

Real-browser:
- Subscribed + `octo_uid` present → menu shows single "取消订阅", sidebar shows "· 已订阅" (preserves #645).
- Forced the `/subscribers` XHR to error (simulating load failure / the skew window) and re-opened the issue → menu shows **both** "订阅此 Issue" and "取消订阅" (unsubscribe reachable), and "已订阅" is correctly hidden.

## Review gate

- codex adversarial-review: **approve, no material findings** (inspected the touched path).
- Claude code-reviewer (skeptical): no real bug — verified `subLoaded` race is `fresh()`-guarded, empty-list edge (`selfKnown=true → subscribe only`, correctly distinguished from unknown), mixed-skew reasoning, and no reintroduction of the pre-#645 always-both behavior outside the unknown branch.
- `/simplify`: applied — hoisted the `!!myUid` guard out of the `.some()` predicate; extracted `subscribeItem`/`unsubscribeItem` to dedupe the 3-branch menu.

## Tests

No unit test added: `packages/dmloop` has no test harness. The tri-state branches are exercised end-to-end in the real runtime (above). A dmloop test harness remains separate infra work.

## Blast radius

`packages/dmloop/src/panel/IssueDetailPage.tsx` only. Adds one local `subLoaded` state + derived vars; no shared type/prop/signature change, no consumer outside dmloop. The `subscribers` state type is unchanged (kept `IssueSubscriber[]`, not `| null`, to avoid touching every reader).

## Docs

No doc change: UI-only behavior fix, no new service/command/config/env.

### DIFF
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -114,6 +114,8 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
   const [issue, setIssue] = useState<Issue | null>(null);
   const [comments, setComments] = useState<IssueComment[]>([]);
   const [subscribers, setSubscribers] = useState<IssueSubscriber[]>([]);
+  // 订阅者列表是否已成功加载:未加载/加载失败时"我是否已订阅"不可判定,菜单回退到两项都显示。
+  const [subLoaded, setSubLoaded] = useState(false);
   const [children, setChildren] = useState<Issue[]>([]);
   const [timeline, setTimeline] = useState<TimelineEntry[]>([]);
   const [parentCands, setParentCands] = useState<Issue[]>([]); // 父 issue 选择器候选(懒加载)
@@ -149,6 +151,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     // 短暂残留上一个 issue 的子列表、订阅者、父候选(旧候选还会漏掉新的自己)。
     setChildren([]);
     setSubscribers([]);
+    setSubLoaded(false);
     setParentCands([]);
     setTimeline([]);
     Promise.all([getIssue(issueId), listComments(issueId), listRuns(issueId)])
@@ -163,7 +166,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
       .catch(() => { if (fresh()) Toast.error(t("loop.detail.notFound")); })
       .finally(() => { if (fresh()) setLoading(false); });
     // 订阅者、子 issue、时间线旁路加载:失败不影响主体渲染;同样按 token 丢弃过期响应。
-    listSubscribers(issueId).then((s) => { if (fresh()) setSubscribers(s); }).catch(() => {});
+    listSubscribers(issueId).then((s) => { if (fresh()) { setSubscribers(s); setSubLoaded(true); } }).catch(() => {});
     listChildren(issueId).then((c) => { if (fresh()) setChildren(c); }).catch(() => {});
     listTimeline(issueId).then((tl) => { if (fresh()) setTimeline(tl); }).catch(() => {});
   };
@@ -217,7 +220,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     try {
       await (on ? subscribeIssue : unsubscribeIssue)(issueId);
       const s = await listSubscribers(issueId);
-      if (token === reqRef.current) setSubscribers(s);
+      if (token === reqRef.current) { setSubscribers(s); setSubLoaded(true); }
       Toast.success(t(on ? "loop.subscribe.subscribed" : "loop.subscribe.unsubscribed"));
     } catch (e) {
       Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
@@ -226,9 +229,24 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
 
   // 当前 octo 成员是否已订阅:subscriber 现在带 octo_uid(仅 member 有),与
   // loginInfo.uid 比对即可判定,无需前端反查 member↔user 映射(去桥后仍成立)。
+  // 三态:列表未成功加载、或后端未透出 octo_uid(前后端独立上线的版本错配窗口)时
+  // "我是否已订阅"不可判定 → selfKnown=false,菜单回退到"订阅+取消订阅"两项都显示,
+  // 避免已订阅用户在此期间够不到 unsubscribe(#645 单一 toggle 的回归)。
   const myUid = WKApp.loginInfo.uid;
-  const amSubscribed =
-    !!myUid && subscribers.some((s) => s.user_type === "member" && s.octo_uid === myUid);
+  const memberSubs = subscribers.filter((s) => s.user_type === "member");
+  const octoUidSkew = memberSubs.length > 0 && memberSubs.every((s) => s.octo_uid == null);
+  const selfKnown = subLoaded && !octoUidSkew && !!myUid;
+  const amSubscribed = !!myUid && memberSubs.some((s) => s.octo_uid === myUid);
+  const subscribeItem = (
+    <Dropdown.Item icon={<Bell size={13} />} onClick={() => toggleSubscribe(true)}>
+      {t("loop.subscribe.subscribe")}
+    </Dropdown.Item>
+  );
+  const unsubscribeItem = (
+    <Dropdown.Item icon={<BellOff size={13} />} onClick={() => toggleSubscribe(false)}>
+      {t("loop.subscribe.unsubscribe")}
+    </Dropdown.Item>
+  );
 
   // 评论 emoji 反应:选择器点 emoji=加,已有 chip 点击=删自己那条(后端按 actor+emoji 定位)。
   const reactComment = async (commentId: string, emoji: string, add: boolean) => {
@@ -530,14 +548,16 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
         <Dropdown.Item onClick={(e) => e.stopPropagation()}>{t("loop.menu.changeAssignee")}</Dropdown.Item>
       </Dropdown>
       <Dropdown.Divider />
-      {amSubscribed ? (
-        <Dropdown.Item icon={<BellOff size={13} />} onClick={() => toggleSubscribe(false)}>
-          {t("loop.subscribe.unsubscribe")}
-        </Dropdown.Item>
+      {/* 订阅态不可判定(未加载/加载失败/后端未透出 octo_uid)时两项都显示,保证 unsubscribe 可达 */}
+      {!selfKnown ? (
+        <>
+          {subscribeItem}
+          {unsubscribeItem}
+        </>
+      ) : amSubscribed ? (
+        unsubscribeItem
       ) : (
-        <Dropdown.Item icon={<Bell size={13} />} onClick={() => toggleSubscribe(true)}>
-          {t("loop.subscribe.subscribe")}
-        </Dropdown.Item>
+        subscribeItem
       )}
       <Dropdown.Divider />
       <Dropdown.Item type="danger" icon={<Trash2 size={13} />} onClick={handleDeleteIssue}>
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -219,11 +219,20 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     const token = reqRef.current;
     try {
       await (on ? subscribeIssue : unsubscribeIssue)(issueId);
+    } catch (e) {
+      // mutation 失败:服务端状态未变,不动本地订阅态。
+      Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
+      return;
+    }
+    Toast.success(t(on ? "loop.subscribe.subscribed" : "loop.subscribe.unsubscribed"));
+    // mutation 成功后刷新确认新状态;刷新失败则把订阅态标记为"未知"(subLoaded=false),
+    // 菜单回退到两项都显示——否则会留下过期的"已知"态(如刚订阅成功但列表还是旧的空),
+    // 导致服务端已订阅、菜单却只显示「订阅」、藏了「取消订阅」(即本修复要防的回归)。
+    try {
       const s = await listSubscribers(issueId);
       if (token === reqRef.current) { setSubscribers(s); setSubLoaded(true); }
-      Toast.success(t(on ? "loop.subscribe.subscribed" : "loop.subscribe.unsubscribed"));
-    } catch (e) {
-      Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
+    } catch {
+      if (token === reqRef.current) setSubLoaded(false);
     }
   };
 


OctoBoooot (approved):
## Review: keep unsubscribe reachable when subscription state is unknown (#646)

**Verdict:** Approve
Correct fix for a real regression #645 introduced (which I approved without catching): the single state-aware toggle collapsed "unknown" into "not subscribed", so a subscribed member could lose the unsubscribe affordance when self-subscription couldn't be proven. This restores tri-state safety. Verified the logic incl. the tricky empty-list edge.

### What I verified
- **`subLoaded`** set true only after `listSubscribers` resolves (in both `reload` and `toggleSubscribe`), reset per-issue in `reload`, all `fresh()`/token-guarded — a failed/pending fetch leaves it false.
- **`octoUidSkew` = `memberSubs.length > 0 && memberSubs.every(s => s.octo_uid == null)`** — detects the FE-ahead-of-BE deploy window; `== null` catches null+undefined.
- **`selfKnown = subLoaded && !octoUidSkew && !!myUid`** — state determinable only when loaded + backend surfaces octo_uid + logged in. `!selfKnown` → show **both** items (pre-#645 safe behavior, unsubscribe reachable); else single `amSubscribed ? unsubscribe : subscribe`. The unknown case can never strand a subscribed user.
- **Empty-list edge (the subtle one):** a genuinely-empty *loaded* subscriber list → `memberSubs=[]` → `octoUidSkew=false` (the `length>0` guard correctly excludes it from skew) → `selfKnown=true`, `amSubscribed=false` → shows **subscribe only**. Correct: empty-loaded means "not subscribed", not "unknown". Holds.
- **Sidebar/menu consistency:** any uncertain state forces `amSubscribed=false` (no subscriber matches `myUid`), so the "已订阅" indicator hides whenever the menu is in the both-items fallback — they never disagree. `subscribeItem`/`unsubscribeItem` extracted to dedupe the 3 branches.

### Praise
- The tri-state distinction is exactly right: "loaded-empty" (→ subscribe only) is correctly separated from "not-loaded / skewed" (→ both items) via the `length>0` guard on skew and `subLoaded` on load — the easy mistake here is treating an empty list as "unknown" (over-showing) or treating a failed load as "not subscribed" (stranding the user), and this avoids both. Falling back to the pre-#645 always-both behavior for the unknown case is the safe choice, layered under the backend's still-idempotent operation.

### Note (self-accountability)
This is the failure mode I should have enumerated when I approved #645 — I verified the happy-path `octo_uid` identity match but not the "state-source unknown → stranded subscriber" case, which is precisely the empty/loading-source derivation lesson from #640. The author/codex caught it in the follow-up gate; recording it so I run that check on any "derive UI state from a fetched list" change.

Jerry-Xin (dismissed):
The PR is in-scope for octo-web, but the tri-state subscription fix still misses one uncertain state after subscribe/unsubscribe mutations.

**🔴 Blocking**
- 🔴 Critical: [IssueDetailPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-owvfwgj6/repo/packages/dmloop/src/panel/IssueDetailPage.tsx:218) leaves stale “known” subscription state if the mutation succeeds but the follow-up `listSubscribers(issueId)` fails. Example: initial list loads empty, user clicks Subscribe, `subscribeIssue` succeeds, then the refresh fetch fails. The `catch` shows an error but leaves `subLoaded=true` and `subscribers=[]`, so the user is now subscribed server-side while the menu still shows only Subscribe, hiding Unsubscribe. This recreates the stuck-subscribed failure mode the PR is meant to fix. Split mutation from refresh or mark `subLoaded=false` for the current token when the post-mutation refresh cannot prove the new state.

**✅ Highlights**
- The initial-load failure path is handled correctly by resetting `subLoaded` on issue reload and only setting it after a successful subscriber fetch.
- The `reqRef` freshness guard is preserved for subscriber updates, avoiding stale writes across issue switches.
- The deploy-skew fallback for member subscribers without `octo_uid` keeps both actions reachable.

yujiawei (approved):
# Code Review — PR #646 (octo-web)

**Verdict: APPROVED** — no blocking (P0/P1) findings.

## Summary
This is a focused, single-file regression fix. #645 replaced the always-both subscribe/unsubscribe menu with a single state-aware toggle driven off `amSubscribed`, which collapsed the *unknown* subscription state into `false`. As a result an already-subscribed user could lose the **unsubscribe** affordance whenever self-subscription could not be proven — either the `/subscribers` fetch failed, or the backend had not yet started surfacing `octo_uid` (deploy-skew window). This PR introduces an explicit tri-state so that any *uncertain* state falls back to showing **both** actions, keeping unsubscribe reachable.

## Spec compliance
The diff implements exactly what the description claims, with no scope creep:
- `subLoaded` — set `true` only after `listSubscribers` resolves, reset per issue in `reload()`, and written only under the `reqRef`/`fresh()` token guard (both the load path at `IssueDetailPage.tsx:169` and the toggle path at `:223`).
- `octoUidSkew` — `memberSubs.length > 0 && memberSubs.every((s) => s.octo_uid == null)` (`:237`).
- `selfKnown = subLoaded && !octoUidSkew && !!myUid` (`:238`); when false the menu renders both items (`:552-556`).
- No shared type/prop/signature change; `IssueSubscriber` (`api/types.ts:217`) is untouched. Blast radius is confined to `packages/dmloop`.

## Correctness — verified edge cases
The claimed invariant *"any uncertain state forces `amSubscribed = false`"* holds across every branch, so the menu (`:552`) and the sidebar "已订阅" indicator (`:1046`) can never disagree:
- **Fetch failed / pending** → `subLoaded = false` → `selfKnown = false` → both items shown; `subscribers` empty → `amSubscribed = false`.
- **Deploy-skew** → `octoUidSkew = true` → `selfKnown = false` → both items; every `octo_uid` is null so `some((s) => s.octo_uid === myUid)` is `false`.
- **Empty subscribers, loaded** → `memberSubs.length === 0` makes `octoUidSkew = false` (correctly guarded by the `length > 0` term), so `selfKnown = true`, `amSubscribed = false` → subscribe only. This correctly distinguishes "loaded and genuinely not subscribed" from "unknown".
- **Issue switch race** → `reload()` resets `subLoaded = false` and bumps the token; the resolved `listSubscribers` only writes state when `fresh()`. No stale write.
- **`== null` vs `=== myUid`** — the loose `== null` intentionally catches both `null` and `undefined` (`octo_uid?: string | null`); the strict `=== myUid` compares two strings. Both correct.

## Security note (PR was auto-classified `security_sensitive`)
I could not find a genuine security impact — this reads as a keyword false-positive:
- No auth/token/secret/credential/middleware change; the diff is purely client-side menu-visibility logic.
- Showing both menu items does not grant elevated capability: `subscribeIssue`/`unsubscribeIssue` (`api/collabApi.ts:11-15`) send empty-body POSTs that the backend applies to the **caller only** and idempotently. A user can only (un)subscribe themselves regardless of which menu items are visible.

Worth a human's confirmation only that the backend `/subscribe` / `/unsubscribe` endpoints do in fact scope to the authenticated caller (the client relies on that contract); the client change itself introduces no new trust-boundary risk.

## Non-blocking notes (P2 / UX)
- During the skew window (or a failed fetch) a genuinely-subscribed user sees the menu offer both actions while the sidebar hides "已订阅". This is acceptable graceful degradation given the state is unknowable client-side, and it is documented in the code comment at `:232-234`. No change required.
- No unit test added; `packages/dmloop` has no test harness, and the tri-state branches are only exercisable end-to-end. Standing infra gap, not a blocker for this PR.

## Verification
Reviewed against head SHA `c8a14fc4650610e3eeed511e6583223eeebbc714` with a fresh checkout, including the prior #645 commit (`ff1c1188`) for regression context and the `collabApi`/`types` contracts. Independent multi-lens adversarial pass (correctness/race, skew edge cases, security, spec/UX) surfaced no confirmed blocking findings.

Jerry-Xin (commented):
Downgrading my prior REQUEST_CHANGES to a non-blocking note — the finding is real but I over-called its severity. On reflection and peer adjudication, this is a fast-follow, not a merge blocker.

The tri-state fix itself is correct and complete: `subLoaded` / `octoUidSkew` / `selfKnown` are handled properly, and the empty-list edge is correctly distinguished from the not-loaded state.

Non-blocking finding (kept on record):
`toggleSubscribe` leaves the subscription display stale when the post-mutation `listSubscribers` refetch fails. The backend mutation has already succeeded at that point, but control jumps to `catch` (Toast.error only), so `setSubscribers` / `setSubLoaded` are never reached — the menu/indicator keep showing the pre-mutation state until the next reload. Impact is limited: no data loss, backend write succeeded, it self-corrects on any reload, the backend is idempotent so acting on the stale display is a safe no-op, and an error toast is surfaced. The toast is imprecise, though — it reports a save failure when the save actually succeeded and only the refresh failed.

Suggested fast-follow:
- Optimistically derive local state from the known `on` argument, or
- Split mutation-failure from refresh-failure so the toast message is accurate and a refresh-only failure does not strand a stale display.

Net: not a merge blocker. APPROVE-equivalent on the code, with the above as a fast-follow.

yujiawei (changes requested):
# Code Review — PR #646 (octo-web)

**Verdict: CHANGES_REQUESTED** — one P1 correctness gap in the mutation path.

> Note: this review corrects an earlier Approve I posted on this same head. On a closer trace of the subscribe/unsubscribe mutation path I found a residual instance of the exact failure mode this PR targets. Details below.

## What the PR does well
The initial-load and deploy-skew paths are handled correctly:
- `subLoaded` is reset per issue in `reload()` (`IssueDetailPage.tsx:154`) and set `true` only after a successful subscriber fetch (`:169`), so a failed/pending initial `/subscribers` fetch → `selfKnown=false` → both actions shown.
- The `reqRef`/`fresh()` token guard is preserved for subscriber writes, avoiding stale writes across issue switches.
- `octoUidSkew` (`:237`) keeps both actions reachable during the backend deploy-skew window when member subscribers lack `octo_uid`.
- The empty-but-loaded case is correctly distinguished from unknown (the `memberSubs.length > 0` guard keeps `octoUidSkew=false`, so `selfKnown=true`, `amSubscribed=false` → subscribe only).

## 🔴 Blocking (P1) — stuck-subscribed recurs when the post-mutation refresh fails
`toggleSubscribe` (`IssueDetailPage.tsx:218-228`):

```tsx
const toggleSubscribe = async (on: boolean) => {
  const token = reqRef.current;
  try {
    await (on ? subscribeIssue : unsubscribeIssue)(issueId);
    const s = await listSubscribers(issueId);      // <-- can reject after the mutation already succeeded
    if (token === reqRef.current) { setSubscribers(s); setSubLoaded(true); }
    Toast.success(...);
  } catch (e) {
    Toast.error(...);                               // <-- does NOT reset subscribers / subLoaded
  }
};
```

Reachable scenario:
1. Initial `/subscribers` loads successfully empty → `subLoaded=true`, `subscribers=[]` → menu shows **Subscribe** only (correct).
2. User clicks Subscribe → `subscribeIssue` **succeeds** (now subscribed server-side) → `listSubscribers` **throws** (network blip between the two back-to-back requests).
3. `catch` shows a "save failed" toast but leaves `subLoaded=true` and `subscribers=[]`.
4. Result: `selfKnown=true`, `amSubscribed=false` → menu **still shows Subscribe only**, Unsubscribe hidden, while the user is subscribed server-side.

This is the same stuck-subscribed / unsubscribe-unreachable mode the PR sets out to eliminate, and it violates the invariant the code documents at `:232-234` ("any uncertain state forces `amSubscribed=false`"): here the state is effectively *unknown* (the refresh could not confirm it), yet `subLoaded` stays true so it is treated as *known*.

Mitigating factor: it is recoverable — re-opening the issue triggers `reload()`, which resets `subLoaded` and refetches; and `subscribeIssue` is idempotent so a manual retry also recovers. So it is narrower and less persistent than the deploy-skew case. But it still directly undermines the PR's stated contract with a trivial, low-risk remedy, so it should be closed before merge.

Suggested fix (either is fine):
- In the failure path, mark the state unknown for the current token so the menu falls back to both actions: `catch (e) { if (token === reqRef.current) setSubLoaded(false); Toast.error(...); }`, or
- Split the mutation from the refresh: on mutation success, optimistically reflect the new subscription (e.g. set `amSubscribed` intent) so the toggle stays consistent even if the refresh fetch fails.

## Security note (PR auto-classified `security_sensitive`)
No genuine security impact found — reads as a keyword false-positive. The diff is purely client-side menu-visibility logic; no auth/token/secret/credential/middleware change. Showing both menu items grants no elevated capability: `subscribeIssue`/`unsubscribeIssue` (`api/collabApi.ts:11-15`) are empty-body POSTs the backend applies to the caller only, idempotently. Worth a human confirming only that those endpoints scope to the authenticated caller (the client relies on that contract).

## Non-blocking
- During the skew window a genuinely-subscribed user sees both menu actions while the sidebar hides "已订阅" (`:1046`). Acceptable graceful degradation for an unknowable-client-side state.
- No unit test; `packages/dmloop` has no harness — standing infra gap, not a blocker for this PR.

## Verification
Reviewed against head SHA `c8a14fc4650610e3eeed511e6583223eeebbc714` (fresh checkout), including #645 (`ff1c1188`) for regression context and the `collabApi`/`types` contracts.

OctoBoooot (changes requested):
## Review: keep unsubscribe reachable when subscription state is unknown (#646) — correcting my verdict

On reconsideration I'm flipping APPROVE → REQUEST_CHANGES. I under-graded the `toggleSubscribe` refresh-failure finding as minor; reframed against this PR's own stated objective it's a **major** — the fix fails to meet its own goal on a reachable path. Two reviewers (Jerry-Xin, yujiawei) independently graded it blocking; on re-check their framing is right.

Risk tier: normal; `needs-human-review` applied. Base `feat/octo-loop`; `mergeStateStatus=CLEAN` but `reviewDecision=CHANGES_REQUESTED`.

### Major (corrected severity — the self-defeating-objective path)
`toggleSubscribe` (`IssueDetailPage.tsx:218-224`): mutation succeeds, then the follow-up `listSubscribers` throws → control jumps to `catch` (Toast only); **`setSubscribers`/`setSubLoaded` are never reached**. So `subLoaded` stays `true` (from the initial load) → `selfKnown = true` → the menu shows the single state-aware item off a **stale** `subscribers`. A user who just subscribed (backend succeeded) but whose refresh failed sees `amSubscribed=false` → **"subscribe" only, unable to unsubscribe** — exactly the stranded-user state this PR exists to eliminate, reappearing one path over (post-mutation instead of initial-load).

Why I now grade this major, not minor: #646's stated purpose is *"keep unsubscribe reachable when state is unknown."* After mutation-then-refresh-failure the state genuinely IS unknown, but `subLoaded=true` makes the code treat it as known, so the PR's own unknown-fallback (show both items) **does not engage on this path**. That's not generic transient staleness — it's the fix not achieving its own objective on a reachable path. My earlier "self-corrects on reload / idempotent" mitigants are real (hence major, not blocker, and distinct from #640's stored-data destruction), but I over-weighted them and under-weighted that this reintroduces the exact bug the PR targets.

### Fix (agreed across reviewers)
On the refresh-failure path, set `subLoaded = false` for the current token (or decouple mutation from refresh / optimistically derive from the known `on` argument), so `selfKnown` becomes false and the menu falls back to showing both items — consistent with the PR's own unknown-handling. That closes the gap the PR's tri-state design already establishes everywhere else.

### Note (accountability)
I approved this at minor; that was under-calibrated. yujiawei and Jerry-Xin flagged it blocking, and the "fails its own stated objective on a reachable path" reframing (Steve's sharpening) is what moved me — I'd anchored on "transient, self-corrects" and missed that it re-strands the very user this PR protects. The tri-state design for the initial-load path remains correct and well-built; this is specifically the post-mutation refresh-failure branch.

OctoBoooot (approved):
## Review: keep unsubscribe reachable when subscription state is unknown (#646) — delta re-review (major fixed)

The new commit applies exactly the agreed fix: `toggleSubscribe` now decouples mutation from refresh, and on post-mutation refresh-failure sets `subLoaded=false` so the PR's own unknown-fallback engages. The major I flipped to is resolved — unsubscribe is reachable on that path now. One residual cosmetic minor. Flipping REQUEST_CHANGES → APPROVE.

Risk tier: normal; `needs-human-review` applied. Base `feat/octo-loop`; `mergeStateStatus=CLEAN`.

### Resolved — Major (post-mutation refresh-failure stranding)
`toggleSubscribe` is restructured into two phases:
- **Mutation:** `try { subscribe/unsubscribe } catch { Toast.error; return }` — failure surfaces the error and returns without touching local state (accurate message now; no stale write).
- **Refresh (separate try/catch):** on success `setSubscribers + setSubLoaded(true)`; **on failure `setSubLoaded(false)`** (token-guarded) → `selfKnown=false` → menu shows both items → unsubscribe reachable. This closes the exact self-defeating-objective path (mutation succeeds, refresh fails → previously stayed `selfKnown=true` on stale data → stranded). Correct and complete; also fixes the imprecise-toast sub-issue (mutation vs refresh failure now distinct).

### Minor (residual, non-blocking)
- The sidebar "已订阅" indicator (`:1055`) is gated on `amSubscribed` alone, not `selfKnown`. On **unsubscribe-then-refresh-fail**, `subscribers` stays stale (still lists the user) → `amSubscribed=true` → the indicator shows "已订阅" even though they just unsubscribed, while the menu (correctly, `selfKnown=false`) offers both items. Transient cosmetic mismatch, self-corrects on reload, no functional impact (the important part — unsubscribe reachability — is fixed). Consider gating the indicator on `selfKnown && amSubscribed` so the badge also falls silent in the unknown state, matching the menu.

### Note
The blocking issue is genuinely fixed; my flip to REQUEST_CHANGES is satisfied. The residual is a display-only nit on one action-then-failure path. Verdict corrected back up to APPROVE — this closes the #645→#646 subscribe-toggle arc, where the severity settled honestly through reviewer disagreement (I corrected minor→major on the objective-failure reframing; the author's fix now resolves it).

Jerry-Xin (approved):
Re-reviewing at head 2f67831b8b85. The blocking concern I (and two other reviewers) raised against the previous head c8a14fc4 has been addressed by the follow-up commit "fix(loop): mark subscription unknown when post-mutation refresh fails".

Previously, on the mutation-success-then-refresh-failure path, `toggleSubscribe` jumped to a single `catch` that only showed a Toast; `setSubLoaded` was never touched, so `subLoaded` stayed `true`, `selfKnown` stayed `true`, and the "unknown state → show both items" fallback never engaged — stranding a just-subscribed user on subscribe-only until reload, i.e. the very regression this PR aims to fix.

The new code correctly separates the two failure modes:
- Mutation failure → `Toast.error` + early `return`; local subscription state is left untouched (server state unchanged). Correct.
- Mutation success → success Toast, then a **separate** try/catch for the confirming `listSubscribers` refresh. On refresh failure it now runs `if (token === reqRef.current) setSubLoaded(false)`, which flips `selfKnown` to `false` and makes the both-items fallback engage — so Unsubscribe stays reachable without a reload. This meets the PR's stated goal on the refresh-failure path.

The token guard (`token === reqRef.current`) is preserved on both the success and failure branches, so a stale in-flight response can't clobber a newer view after an issue switch.

The render-path tri-state (`subLoaded` / `octoUidSkew` / `selfKnown`, empty-vs-not-loaded) is otherwise correct and unchanged.

LGTM. Note: the other two CHANGES_REQUESTED reviews are still pinned to the older head c8a14fc4 and predate this fix.

yujiawei (approved):
# Code Review — PR #646 (octo-web)

**Verdict: APPROVED** — no blocking (P0/P1) findings at the current head.

> Reviewed head SHA `2f67831b8b85af96e4568afe5b4b398ccd5cdac0`. Note: GitHub's `reviewDecision` may still read `CHANGES_REQUESTED` from an earlier head (`c8a14fc4`); the P1 those reviews targeted is resolved by the follow-up commit on this head (see below). This is a fresh independent verdict against the current head, not an inheritance of a prior one.

## Summary
Focused, single-file regression fix (`packages/dmloop/src/panel/IssueDetailPage.tsx`). #645 replaced the always-both subscribe/unsubscribe menu with a single toggle driven off `amSubscribed`, which collapsed the *unknown* subscription state into `false` — so an already-subscribed user could lose the **unsubscribe** affordance whenever self-subscription couldn't be proven (the `/subscribers` fetch fails, or the backend deploy-skew window where `octo_uid` isn't surfaced). This PR introduces a tri-state (`subLoaded` / `octoUidSkew` / `selfKnown`): when the state is not known, the menu falls back to showing **both** items, keeping unsubscribe reachable.

## Spec compliance
- **Missing:** none. The PR implements exactly the tri-state fallback described in its own body.
- **Extra:** none. No new prop/type/signature; `subscribers` state type is unchanged; blast radius is one file with no consumer outside dmloop.
- **Divergence:** none.

## Correctness — the load-bearing invariant
`amSubscribed === true` **implies** `selfKnown === true`, which is what makes the design safe:
- `amSubscribed` (L248) requires some member subscriber with `octo_uid === myUid`, i.e. a non-null `octo_uid` exists → `octoUidSkew` (L246) is `false`.
- `subscribers` is only ever populated by the two call sites that also call `setSubLoaded(true)` in the same statement (L169, L233); no path fills `subscribers` while leaving `subLoaded=false` → `subLoaded=true`.
- `amSubscribed` requires `!!myUid`.

Consequences verified by tracing all reachable states:
- A genuinely-subscribed user is **never** routed to "subscribe only."
- Every degraded state (fetch fail → `subLoaded=false`; deploy-skew → `octoUidSkew=true`; missing uid) routes to `!selfKnown` (L561) → **both** items → unsubscribe stays reachable. The #645 regression is closed.
- The sidebar "· 已订阅" indicator (L1055) reads the same `amSubscribed`, so it cannot contradict the menu even though it isn't independently gated on `selfKnown`.
- Empty-list edge: `octoUidSkew` is guarded by `memberSubs.length > 0`, so an empty list with `subLoaded=true` correctly yields "subscribe only," distinct from the not-loaded state.

## Post-mutation refresh path (the previously-flagged P1)
`toggleSubscribe` (L218–237) now, on a **successful** mutation whose follow-up `listSubscribers` refresh **fails**, sets `setSubLoaded(false)` (L235) — marking the state unknown so the menu falls back to both items instead of leaving a stale "known" state that would hide unsubscribe. This resolves the empty-`catch` finding raised against the earlier head. The mutation-failure path (L222–226) correctly returns without touching local state. The `reqRef` token (captured L219, checked at L233/L235) discards writes from a toggle whose refresh lands after an issue switch. Confirmed by direct read of this head.

## Non-blocking suggestions (P2 — not merge blockers)
- **No double-click / in-flight guard on `toggleSubscribe`**, unlike sibling mutations (`busyRunId`, `submitting`). Benign because the backend is idempotent and same-actor, but a `subBusy` guard would match the codebase's own convention and avoid a redundant refresh.
- **Test coverage:** this is the second corrective iteration on this exact logic (#645 → #646) and none of the degraded-state transitions has a component test. `packages/dmloop` lacks a harness today; adding one (even a minimal render test for the three degraded branches) would let CI catch a future regression here. Tracking as follow-up infra work is reasonable.

## Items worth a human's eye (security-sensitive PR)
No client-side auth decision is made that the server doesn't re-enforce: `octo_uid === myUid` gates only a UI affordance and a cosmetic badge, never a mutation target (`subscribe/unsubscribe` take only `issueId`; the actor is server-authenticated). Showing both items when state is unknown cannot auto-fire anything — each requires an explicit click, and the backend calls are idempotent. Two contracts this design **rests on** but that can't be verified from the diff, and are the right things for a human to confirm:
- `listSubscribers` actually surfaces a member's own `octo_uid` post-deploy (the whole tri-state depends on it; a wrong contract degrades to "show both," not a hard bug).
- `subscribeIssue`/`unsubscribeIssue` are idempotent for the same actor (relied on by the both-items fallback).

PR #647, opened 2026-07-11
fix(loop): don't lose comment attachments when upload fails after post

MrAladdin:
## What

Follow-up fix to **#643** (file attachments). Reviewer **yujiawei** flagged a P1 that shipped with #643: a failed comment attachment was silently and permanently lost.

`submitComment` cleared `pendingFiles` before the upload loop and, on a per-file failure, only set `attFailed` and discarded the `File` — so the file was gone with no retry path, behind a generic **"save failed"** toast that wrongly implied the whole comment failed (it had already been created successfully). For an attachments feature, silently dropping the user's file behind a misleading error is a functional defect.

## Fix

- **Re-stage failed files + accurate message** — collect failed `File`s into `failedFiles`; after the loop, put them back into `pendingFiles` (non-reply path) so they aren't lost, and show a distinct toast: *"评论已发送，但 N 个附件上传失败，已放回输入框请重试"* instead of the generic save-failed.
- **`uploadForIssue` per-file isolation** — replace the single `try` around the awaited loop (which aborted remaining files after the first failure) with per-file `try/catch` + a failure count, consistent with `submitComment`.
- **i18n** — add `commentAttachFailed` / `attachFailed` (zh + en). Passed via this repo's `I18nService` convention `{ values: { count } }` (not `{ count }`).

Real-browser, `/upload-file` XHR forced to fail:
- Sending a comment with an attachment → **comment is posted**, the failed file is **re-staged in the composer** (`pr-g2.txt` reappears — not lost), and the toast reads **"评论已发送，但 1 个附件上传失败，已放回输入框请重试"** — the `{{count}}` is interpolated to `1` (no literal placeholder).

## Review gate

- Claude code-reviewer (skeptical): caught a **Critical** in the first pass — the two new `t(...)` calls used `{ count }` instead of the project's `{ values: { count } }`, so interpolation silently failed (users would see literal `{{count}}`) and it may trip TS excess-property-check. **Fixed** (verified above the count now renders); re-reviewed clean on the other points (re-stage/replyTo, token race — mooted by `key={issueId}` remount, `finally` ordering).
- codex adversarial-review: approve, no material findings.
- `/simplify`: no changes needed.

## Not fixed here (deferred, from the same review — non-blocking P2)

- Client-side size/type/`accept` validation (needs a product decision on limits).
- Attachment-only comment (Send gated on non-empty content) — needs the backend contract for empty-content comments.
- `download_url` thumbnail `onError` fallback (short-lived signed URL past TTL).
- Index-as-key on the pending-file chips (documented no-runtime-bug nit; a proper fix needs per-file ids).

## Blast radius

`packages/dmloop` only — `IssueDetailPage.tsx` (`submitComment` / `uploadForIssue`) + two additive i18n keys. No shared type/prop/signature change, no consumer outside dmloop.

## Docs

No doc change: UI error-handling fix, no new service/command/config/env.

### DIFF
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -224,7 +224,9 @@
     "commentUpdated": "Comment updated",
     "editFailed": "Edit failed",
     "deleteFailed": "Delete failed",
-    "saveFailed": "Save failed"
+    "saveFailed": "Save failed",
+    "commentAttachFailed": "Comment posted, but {{count}} attachment(s) failed to upload — re-added to the box, please retry",
+    "attachFailed": "{{count}} attachment(s) failed to upload"
   },
   "validate": {
     "nameRequired": "Name is required",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -224,7 +224,9 @@
     "commentUpdated": "评论已更新",
     "editFailed": "编辑失败",
     "deleteFailed": "删除失败",
-    "saveFailed": "保存失败"
+    "saveFailed": "保存失败",
+    "commentAttachFailed": "评论已发送，但 {{count}} 个附件上传失败，已放回输入框请重试",
+    "attachFailed": "{{count}} 个附件上传失败"
   },
   "validate": {
     "nameRequired": "请填写名称",
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -384,7 +384,8 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
           failedFiles.push(f);
         }
       }
-      await reloadComments(token);
+      // 先做失败回填 + 文案(同步、不会抛),再刷新评论列表:reloadComments await listComments 可能抛,
+      // 若放在其后,reload 失败会跳过回填 → 失败文件再次丢失(#647 评审抓到的排序缺口)。
       if (failedFiles.length) {
         // 评论已建成功、仅附件失败:把失败文件放回输入区(否则文件对象被丢弃、永久丢失且无重试入口),
         // 并用区分于"评论失败"的文案——避免误导用户以为整条评论没发出去。
@@ -393,6 +394,7 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
       } else {
         Toast.success(t("loop.toast.commentAdded"));
       }
+      await reloadComments(token);
     } finally {
       setSubmitting(false);
     }
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -313,11 +313,18 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     if (!files?.length) return;
     const token = reqRef.current;
     setUploading(true);
+    let failed = 0;
     try {
-      for (const f of Array.from(files)) await uploadAttachment(f, { issueId });
-    } catch (e) {
-      Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
+      // 逐文件隔离:一个失败不影响其余(与 submitComment 一致),否则首个失败会漏传后续文件。
+      for (const f of Array.from(files)) {
+        try {
+          await uploadAttachment(f, { issueId });
+        } catch {
+          failed++;
+        }
+      }
     } finally {
+      if (failed) Toast.error(t("loop.toast.attachFailed", { values: { count: failed } }));
       // 先解锁再触发重取(syncIssue 自带 catch、fire-and-forget):即使重取失败也不会把上传按钮永久卡在 disabled。
       setUploading(false);
       syncIssue(token);
@@ -369,17 +376,23 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
       setSuppressed(new Set());
       if (!replyTo) setPendingFiles([]);
       // 把待发文件带 commentId 绑到已建评论;单个失败只记录、不回滚评论。
-      let attFailed = false;
+      const failedFiles: File[] = [];
       for (const f of files) {
         try {
           await uploadAttachment(f, { commentId: comment.id });
         } catch {
-          attFailed = true;
+          failedFiles.push(f);
         }
       }
       await reloadComments(token);
-      if (attFailed) Toast.error(t("loop.toast.saveFailed"));
-      else Toast.success(t("loop.toast.commentAdded"));
+      if (failedFiles.length) {
+        // 评论已建成功、仅附件失败:把失败文件放回输入区(否则文件对象被丢弃、永久丢失且无重试入口),
+        // 并用区分于"评论失败"的文案——避免误导用户以为整条评论没发出去。
+        if (!replyTo) setPendingFiles(failedFiles);
+        Toast.error(t("loop.toast.commentAttachFailed", { values: { count: failedFiles.length } }));
+      } else {
+        Toast.success(t("loop.toast.commentAdded"));
+      }
     } finally {
       setSubmitting(false);
     }


OctoBoooot (approved):
## Review: don't lose comment attachments when upload fails after post (#647)

**Verdict:** Approve
Clean, complete fast-follow for the #643 error-path gaps yujiawei flagged (and I byte-verified as major). Both are fixed: comment attachments are re-staged on failure instead of lost, and `uploadForIssue` now isolates per-file. Verified against the code.

### Resolved (the two #643 findings)
- **Comment attachment loss (was major):** `submitComment` now collects `failedFiles: File[]` (was a discarded boolean) and, after the loop, `if (!replyTo) setPendingFiles(failedFiles)` — **re-stages the failed files into the composer** so they survive for retry, with a distinct toast (`commentAttachFailed`: "comment posted, but N attachments failed — re-added, please retry"). This fixes both sub-points I raised: files no longer lost, and the message no longer misleads (it correctly says the comment posted). The comment-created-before-upload ordering (duplicate-prevention) is preserved; only the failed-file handling changed.
- **`uploadForIssue` per-file isolation:** the single `try` around the awaited loop (which aborted remaining files after the first failure) is replaced with per-file `try/catch` + a `failed` count, matching `submitComment`. The two upload loops are now consistent — no more mid-list abort.

### Also verified
- **i18n convention:** the new `commentAttachFailed`/`attachFailed` keys use `{ values: { count } }` (this repo's `I18nService` form) — the body notes the first pass used `{ count }` (would render a literal `{{count}}`) and it's fixed; the diff confirms the correct form, zh + en both added.
- **Edges:** re-stage guarded on `!replyTo` (correct — attachments belong to the top-level composer; on a reply `files` is empty and the loop is a no-op); `finally { setSubmitting(false) }` busy guard intact; token-guarded reload unchanged.

### Praise
- The fix does the two things the #643 error paths needed and nothing more: re-stage the actual `File` objects (not just a flag) so retry is possible, and give a message that distinguishes "comment failed" from "comment posted, attachments failed" — the misleading generic toast was half the defect. Bringing `uploadForIssue` to the same per-file isolation as `submitComment` also closes the asymmetry I flagged. The deferred items (client-side validation, attachment-only comment, thumbnail onError, per-file chip keys) are honestly scoped out as P2.

Jerry-Xin (changes requested):
This PR is relevant to `octo-web`, but one ordering bug still allows failed comment attachments to be lost.

🔴 Blocking

🔴 Critical — Failed files are still lost if comment reload fails after attachment upload failures. In `packages/dmloop/src/panel/IssueDetailPage.tsx:377`, `pendingFiles` is cleared before uploads. Failed uploads are collected at `packages/dmloop/src/panel/IssueDetailPage.tsx:379`, but `await reloadComments(token)` runs at `packages/dmloop/src/panel/IssueDetailPage.tsx:387` before the failed files are restored at `packages/dmloop/src/panel/IssueDetailPage.tsx:391`. If `reloadComments` rejects, control jumps to `finally`, so the failed `File` objects are never re-staged and no attachment-specific toast is shown. This preserves the data-loss failure mode under a combined upload failure plus comment reload failure. Restore/show failed files before awaiting reload, or isolate reload in its own `try/catch` after failed-file handling.

💬 Non-blocking

🟡 Warning — I could not run `pnpm i18n:check` because this checkout has no installed dependencies: `typescript` is missing from `node_modules`. `git diff --check` completed cleanly before that failure, and both modified locale JSON files parse successfully.

✅ Highlights

🔵 Suggestion — The per-file upload isolation in `uploadForIssue` and `submitComment` is the right direction, and the new i18n calls use the project’s required `{ values: { count } }` interpolation shape.

OctoBoooot (changes requested):
## Review: don't lose comment attachments when upload fails after post (#647) — correcting my verdict

I flipped APPROVE → REQUEST_CHANGES. Jerry-Xin caught a real ordering gap I missed: the re-stage of failed files sits **after** `await reloadComments`, so if the reload throws, the re-stage is skipped and the files are lost again — the exact data-loss this PR exists to prevent, just moved one step later.

Risk tier: normal; `needs-human-review` applied. Base `feat/octo-loop`.

### Major (missed on my first pass, byte-verified)
`submitComment` (`IssueDetailPage.tsx`):
```
for (const f of files) { try { await uploadAttachment(f, {commentId}) } catch { failedFiles.push(f) } }
await reloadComments(token);                       // ← can throw (listComments network failure)
if (failedFiles.length) { if (!replyTo) setPendingFiles(failedFiles); Toast.error(...) }  // skipped if reload threw
```
`reloadComments` does `const c = await listComments(issueId); ...` — the `await` can throw. If it does, control leaves the `try` for `finally { setSubmitting(false) }` and the exception propagates; the re-stage (`setPendingFiles(failedFiles)`) and the toast never run → the failed files are silently lost. Reachable on the combined path (attachment upload fails **and** comment reload fails) — correlated failures (same backend, a transient blip fails both), so it's a genuine narrower re-occurrence of the #643 bug this PR fixes. Severity: major (same data-loss/no-retry, but narrower than #643 since it needs the extra reload-failure; no backend/comment corruption).

### Fix
Re-stage the failed files (and toast) **before** `await reloadComments` — or wrap the reload so its failure can't bypass the re-stage:
```
if (failedFiles.length && !replyTo) setPendingFiles(failedFiles);
if (failedFiles.length) Toast.error(t("loop.toast.commentAttachFailed", {values:{count: failedFiles.length}}));
else Toast.success(t("loop.toast.commentAdded"));
await reloadComments(token);   // reload last; its failure no longer strands the files
```
(The `uploadForIssue` per-file isolation fix and the i18n `{values:{count}}` fix in this PR are both correct — this is only the submitComment re-stage ordering.)

### Note (accountability)
I approved this without tracing that the re-stage sits behind an `await` that can throw — the second "failure-path ordering after an await" gap I've under-checked this batch. The PR's *intent* and most of its content are right (per-file isolation, accurate message, re-stage the actual File objects); it just re-stages one step too late. Flipping to REQUEST_CHANGES so the ordering is fixed before merge, since it re-opens the exact data-loss the PR targets. Credit: Jerry-Xin.

yujiawei (approved):
# Code Review — PR #647 (octo-web)

**Verdict: APPROVED** — this is a correct, well-scoped fix for the attachment-loss defect from #643. No blocking (P0/P1) issues. A few P2 UX/observability follow-ups are listed below; none should hold up merge.

Reviewed at head SHA `93e5bd7e1f40c54c88f6eba0e675477cd7511e63` against merge-base `29705366` with `feat/octo-loop`. Blast radius is `packages/dmloop` only (`IssueDetailPage.tsx` + two additive i18n keys); no shared type/prop/signature change.

## Scope compliance ✅

- **Nothing missing:** failed `File`s are collected into `failedFiles`, re-staged into `pendingFiles` on the non-reply path, and surfaced via a distinct toast instead of the misleading generic "save failed". `uploadForIssue` gains per-file `try/catch` isolation consistent with `submitComment`. Two i18n keys added in both `zh-CN` and `en-US`.
- **Nothing extra / no divergence:** changes stay within the stated intent; no scope creep.

## What I verified

- **Data-loss fix works.** `submitComment` clears inputs only after the comment is created, then on partial attachment failure restores the failed files to the composer (`IssueDetailPage.tsx:388-392`). The original permanent-loss path is closed.
- **No overwrite race on re-stage.** The composer file `<input>` is `disabled={!!replyTo || submitting}` (`:938`) and there is no drag/drop or paste entry point, so no new file can enter `pendingFiles` while a comment is in flight; `setPendingFiles(failedFiles)` (`:391`) safely replaces the already-cleared list.
- **Reply path is correct (not data loss).** For replies, `files = replyTo ? [] : pendingFiles` (`:360`), so no attachment upload is attempted at all on a reply and `failedFiles` cannot be populated — the `if (!replyTo)` guard on the restore is consistent with the design that replies never carry attachments. There is no silent drop here.
- **i18n interpolation is correct.** `t(key, { values: { count } })` matches the project's `TranslateOptions` signature, and the `{{count}}` placeholders match the `/\{\{\s*([\w.-]+)\s*\}\}/g` interpolator in `I18nService.ts`. Keys resolve under the `loop.` namespace (registered in `module.tsx`). Both locales have both keys. No literal `{{count}}` will leak.
- **No XSS/injection.** `count` is always a number; toast text renders as a plain string (Semi UI `Toast.error`), and `f.name` renders inside a React-escaped text node.

## Non-blocking findings (P2)

1. **`uploadForIssue` no longer surfaces the server error message (`:319-327`).** Pre-PR this path did `Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"))`; the new per-file loop collapses each rejection into a count. This is a small observability regression relative to the rest of the file (every other error site still surfaces `.message`). The per-file isolation itself is a genuine improvement — the trade-off is just that a 401/403/413 is now indistinguishable from a transient network error.

2. **Swallowed per-file error hides a permanently-failing file (`:380-392`, and `:319-327`).** The bare `catch {}` discards the thrown error, which does carry an HTTP `status` (`http.ts` builds an error with `.status` from the response). There is no client-side size/type guard. Consequence: a too-large file (413) is re-staged and will deterministically re-fail on every retry with the same generic count-only toast, giving the user no cue to remove it. Consider logging the caught error and/or emitting a distinguishable "file too large" message. **This is the item most worth a human eye** given the security-sensitive classification — please confirm an auth/permission (403) denial on upload is acceptable to show only as a generic "N attachments failed" count.

3. **Re-staged files can't be retried without re-typing content (`:356-357`, `:373`).** `submitComment` early-returns on an empty draft, and the draft is cleared after the successful comment. So pressing Send with only re-staged files silently no-ops (contradicting the "please retry" copy), and typing new text posts a *second* comment that binds the recovered files to the new `comment.id`, not the original. Files are preserved (core goal met), but the retry affordance the toast implies isn't directly reachable. Consider allowing an attachment-only resubmit when `pendingFiles.length > 0`.

4. **Copy nits.** English strings hardcode `attachment(s)` so `count=1` renders "1 attachment(s)" (the flat interpolator has no ICU plural support; zh is unaffected). Also `commentAdded` ("Comment added" / "评论已添加") vs `commentAttachFailed` ("Comment posted" / "评论已发送") use different verbs for the same act.

## Note on the issue-upload asymmetry

`uploadForIssue` counts failures but discards the `File`s, whereas `submitComment` re-stages them. This is a deliberate difference (issue uploads are immediate/direct; the user still has the OS picker) and not data loss, but if consistent recovery is desired it could re-stage there too. Non-blocking.

Overall: solid fix that closes the reported P1 from #643. Approving; the P2s above are follow-up suggestions, with finding #2 flagged for a human sanity-check on error-surfacing given the security-sensitive label.

yujiawei (changes requested):
# Code Review — PR #647 (octo-web)

**Verdict: CHANGES REQUESTED** — correcting my earlier Approve. A second reviewer surfaced a real ordering bug that I re-traced against the code and confirmed: on a correlated failure path this PR still loses the exact files it is meant to preserve. One-line reorder fixes it; the rest of the PR is correct.

Reviewed at head SHA `93e5bd7e1f40c54c88f6eba0e675477cd7511e63` against merge-base `29705366` with `feat/octo-loop`. Blast radius is `packages/dmloop` only.

## Scope compliance ✅

Fix matches intent: re-stage failed `File`s into the composer, distinct toast, `uploadForIssue` per-file isolation, two additive i18n keys (zh + en). Nothing missing, nothing out of scope. The core mechanism (collecting the actual `File` objects instead of a boolean, and the accurate "comment posted but attachments failed" message) is the right fix for the #643 defect.

## Blocking (P1)

**Re-stage runs after `await reloadComments`, so a reload failure re-opens the data loss (`IssueDetailPage.tsx:377,387-392`).**

```
if (!replyTo) setPendingFiles([]);          // :377  composer cleared
// ...upload loop collects failedFiles...
await reloadComments(token);                // :387  can reject (listComments -> httpGet network call)
if (failedFiles.length) {
  if (!replyTo) setPendingFiles(failedFiles); // :391  SKIPPED if reload threw
  Toast.error(...);                           // :392  SKIPPED if reload threw
}
```

`reloadComments` (`:201-204`) does `const c = await listComments(issueId)`, and `listComments` (`issueApi.ts:100-104`) awaits `httpGet(...)`, which rejects on a network/backend error. If it rejects, control leaves the `try` for `finally { setSubmitting(false) }` and the exception propagates — the re-stage at `:391` and the toast at `:392` never execute. Because `pendingFiles` was already cleared at `:377`, the failed `File` objects are silently lost with no attachment-specific toast: the exact `#643` failure mode, just moved one step later.

Reachable on the combined path (an attachment upload fails **and** the subsequent comment reload fails). These are correlated — same backend, a transient blip can trip both — so it is a genuine, if narrower, re-occurrence of the data loss this PR targets.

**Fix:** re-stage (and toast) before awaiting the reload, or isolate the reload in its own `try/catch` after failed-file handling:

```
if (failedFiles.length) {
  if (!replyTo) setPendingFiles(failedFiles);
  Toast.error(t("loop.toast.commentAttachFailed", { values: { count: failedFiles.length } }));
} else {
  Toast.success(t("loop.toast.commentAdded"));
}
await reloadComments(token);   // reload last; its failure can no longer strand the files
```

## What I verified as correct

- **i18n interpolation.** `t(key, { values: { count } })` matches the project's `TranslateOptions` signature and the `{{count}}` placeholders match the `/\{\{\s*([\w.-]+)\s*\}\}/g` interpolator in `I18nService.ts`. Keys resolve under the `loop.` namespace; both locales have both keys. No literal `{{count}}` leaks.
- **Reply path is not data loss.** `files = replyTo ? [] : pendingFiles` (`:360`), so no upload is attempted on a reply and `failedFiles` cannot be populated; the `if (!replyTo)` guard on the restore is consistent with replies never carrying attachments.
- **No overwrite race on the happy re-stage.** The composer file `<input>` is `disabled={!!replyTo || submitting}` (`:938`) with no drag/drop/paste entry, so no new file enters `pendingFiles` mid-flight.
- **`uploadForIssue` per-file isolation** (`:316-327`) is correct and consistent with `submitComment`.
- **No XSS/injection**; `count` is numeric and toast/`f.name` render as escaped text.

## Non-blocking (P2) — address alongside the fix if convenient

1. **Swallowed per-file error hides a permanently-failing file (`:380-386`, `:319-327`).** The bare `catch {}` discards the thrown error, which carries an HTTP `status`. With no client-side size guard, a too-large file (413) is re-staged and re-fails on every retry with a generic count-only toast, giving no cue to remove it. Consider logging the error and/or a distinguishable "file too large" message. **Worth a human eye given the security-sensitive label:** confirm an auth/permission (403) denial being surfaced only as a generic "N attachments failed" count is acceptable.
2. **`uploadForIssue` lost its server-error message vs pre-PR** (`:319-327`) — it used to show `(e as Error)?.message`; now only a count. Small observability regression (the per-file isolation itself is an improvement).
3. **Retry ergonomics (`:356-357`, `:373`).** After the draft is cleared, pressing Send with only re-staged files no-ops (empty-draft early-return), contradicting the "please retry" copy; typing new text posts a *second* comment binding the files to a new `comment.id`. Consider allowing attachment-only resubmit when `pendingFiles.length > 0`.
4. **Copy nits.** `count=1` renders "1 attachment(s)" (flat interpolator, no ICU plural; zh unaffected); `commentAdded` ("added"/"已添加") vs `commentAttachFailed` ("posted"/"已发送") use different verbs for the same act.

Once the P1 reorder lands, this is a clean fix. The P2s are follow-ups, not blockers.

OctoBoooot (approved):
## Review: don't lose comment attachments when upload fails after post (#647) — delta re-review (blocker fixed)

The new commit applies exactly the three-reviewer-agreed reorder: the failed-file re-stage + toast now run **before** `await reloadComments`, so a reload throw can no longer skip them. The compound-path data-loss is closed. Flipping REQUEST_CHANGES → APPROVE.

Risk tier: normal; `needs-human-review` applied. Base `feat/octo-loop`; `mergeStateStatus=UNSTABLE`.

### Resolved — the reload-ordering data-loss (three-way blocker)
New order in `submitComment`:
```
for (...) { try uploadAttachment catch failedFiles.push(f) }
if (failedFiles.length) { if (!replyTo) setPendingFiles(failedFiles); Toast.error(...) }  // synchronous, cannot throw
else Toast.success(...)
await reloadComments(token);   // moved LAST
```
The re-stage (`setPendingFiles(failedFiles)`) and toast are synchronous state calls that can't throw, and they now execute **before** the throwing `await reloadComments`. So on the compound path (attachment upload fails **and** comment reload fails), the failed `File` objects are already preserved in `pendingFiles` and the accurate message already shown before the reload's exception propagates to `finally`. The narrower unrecoverable-loss window this PR's own fix had left is closed. `finally { setSubmitting(false) }` busy guard and token-guarded reload unchanged.

### Still correct (from prior review)
Per-file isolation in both `submitComment` and `uploadForIssue`; distinct `commentAttachFailed` message; i18n `{ values: { count } }` convention; re-stage guarded on `!replyTo`.

### Note (non-blocking, carried)
yujiawei's retry-ergonomics 🟡 — a re-staged failed file can't be re-sent via Send without new comment text (`if (!content || submitting) return` gates on non-empty content) — is not addressed here, but that's tied to the explicitly-deferred attachment-only-comment P2 (needs the backend contract for empty-content comments) and does not reopen the data-loss: the files are now preserved and retriable once content exists. Fast-follow.

### Accountability
This closes the finding I approved past on my first #647 pass (the re-stage-after-throwing-await ordering) — caught by Jerry-Xin, independently confirmed by yujiawei and me, and now fixed minimally (+3/-1). My flip to REQUEST_CHANGES is satisfied.

Jerry-Xin (approved):
Repository relevance gate: Pass. The PR is scoped to `packages/dmloop` attachment handling and belongs in this project.

💬 Non-blocking

- 🟡 Warning: There is still no direct “retry this failed attachment on the already-created comment” path. Failed files are re-staged in the composer, but `submitComment` still returns when the comment text is empty, so retrying requires entering new comment text and will attach to a new comment. See `packages/dmloop/src/panel/IssueDetailPage.tsx:355` and `packages/dmloop/src/panel/IssueDetailPage.tsx:892`. This is acceptable as a follow-up given the PR’s stated scope, but the toast wording may make users expect a simpler retry.
- 🟡 Warning: I did not find focused tests for this attachment failure path. A component/unit test should mock `addComment` success plus partial `uploadAttachment` failure and assert failed files are retained, remaining files are attempted, and the toast receives the interpolated count.

✅ Highlights

- Per-file upload isolation for issue attachments avoids aborting the remaining files after one failure: `packages/dmloop/src/panel/IssueDetailPage.tsx:317`.
- Comment attachment failures are collected and re-staged before `reloadComments`, avoiding the prior lost-file path if refresh fails: `packages/dmloop/src/panel/IssueDetailPage.tsx:379`.
- The new i18n calls use the project’s expected `{ values: { count } }` shape, and keys were added to both required locales: `packages/dmloop/src/i18n/en-US.json:228`, `packages/dmloop/src/i18n/zh-CN.json:228`.

Checks: `git diff --check` passed. `pnpm i18n:check` could not run because dependencies are not installed in this checkout (`typescript` missing / `node_modules` absent).

yujiawei (approved):
# Code Review — PR #647 (octo-web)

**Verdict: APPROVED** — no blocking (P0/P1) defects. The fix is data-flow-correct and the load-bearing claim (re-stage before the throwable `reloadComments`) holds. Remaining items are non-blocking P2/nits, most pre-existing.

Reviewed at head SHA `c4a4bdea8293babdbb10466a60cdac922ec0a172` against base `feat/octo-loop`. Scope: `packages/dmloop` only — `IssueDetailPage.tsx` (`submitComment` / `uploadForIssue`) + two additive i18n keys.

## 1. Spec compliance

**Spec: ✅**

- **No under-build:** The stated P1 from #643 is fixed — a comment posted successfully with a failed attachment upload no longer silently discards the `File`. Failed files are collected into `failedFiles` and re-staged into the composer (`setPendingFiles(failedFiles)`), with a distinct toast that makes clear the comment *was* posted and only the attachment failed.
- **No over-build:** The diff is tightly matched to the fix — per-file isolation in `uploadForIssue` (consistent with `submitComment`) + 2 i18n keys. No unrelated behavior, prop, type, or signature change; no consumer outside dmloop. Every deferred item listed in the PR description (client-side size/type validation, attachment-only comments, `download_url` `onError` fallback, per-file chip keys) is genuinely absent from the diff — nothing was silently pulled into scope.
- **No deviation:** The two new `t(...)` calls correctly use this repo's `{ values: { count } }` convention (verified against `I18nService.interpolate`, which substitutes `{{count}}` from `values`). Both locale files (`zh-CN`, `en-US` — the only two in `supportedLocales`) were updated, so no fallback gap.

## 2. Code quality

**Quality: Approved**

Key correctness points independently verified against the code:

- **Ordering fix is correct (the core of this PR).** `reloadComments` (`IssueDetailPage.tsx:201-204`) awaits `listComments`, which can throw. Moving `setPendingFiles(failedFiles)` + toast **before** `await reloadComments(token)` (lines 389-397) guarantees the failed files return to the composer and the correct toast fires even if the reload throws. `setPendingFiles`/`Toast` are synchronous and cannot throw, so the ordering is sound.
- **No re-stage / concurrent-add race.** `setPendingFiles(failedFiles)` at line 392 is an absolute overwrite, but the file `<input>` is `disabled={!!replyTo || submitting}` (line 940) and `setSubmitting(true)` runs before any await — so the user cannot add new files during the in-flight window that would be clobbered. Re-entrant submit is also guarded (`if (!content || submitting) return`).
- **Reply path is consistent.** `const files = replyTo ? [] : pendingFiles` (line 360) means a reply never populates `failedFiles`, so the `commentAttachFailed` toast and re-stage branch are unreachable for replies (the `if (!replyTo)` guards are defensive, not dead-code bugs). Top-level composer files correctly survive a reply.
- **Cross-issue safety.** The unguarded `setPendingFiles` is safe because `IssueDetailPage` is mounted with `key={id}` (`pages/IssuePage.tsx:115`); switching issues remounts, so a stale write from a prior issue's in-flight submit lands on an unmounted instance (React no-op).

### Non-blocking notes (P2 / nit — not required for merge)

- **P2 (latent, not introduced here):** the re-stage uses an absolute `setPendingFiles(failedFiles)` rather than a functional update; correctness is coupled to the "input disabled while submitting" invariant rather than enforced locally. A future change that allows staging files mid-submit would reintroduce silent loss. A one-line comment pinning that invariant would harden it.
- **P2 (pre-existing):** both upload loops swallow the caught error (`catch { failed++ }` / `catch { failedFiles.push(f) }`), so an actionable server rejection (413 / too large / unsupported type) is reduced to a generic count. Consider `console.warn(e)` per failure so the reason is at least inspectable — the aggregated toast can stay.
- **nit:** `uploadForIssue` shows no success toast on full/partial success (asymmetric with `submitComment`). Defensible since issue attachments render inline via `syncIssue`; not introduced by this PR.
- **nit:** serial `for…await` uploads (both loops) — `Promise.allSettled` would parallelize while still tracking per-file failures. Pure performance polish, not a defect.

## 3. Security (this PR is classified security-sensitive)

No client-side security defect introduced by the diff:

- The changed code touches no auth/token/credential handling — `uploadAttachment` only builds `FormData`; auth headers are injected by the unchanged axios interceptor.
- No XSS surface: `f.name` and `a.filename` render as JSX text/attribute children (React auto-escapes); no `dangerouslySetInnerHTML` in the package.
- Re-staging reuses the same in-memory `File` refs already picked by the user — no new sensitive data captured or leaked.

**For a human to verify on the backend (out of frontend scope):** `/upload-file` must enforce file size and content-type limits server-side (there is no client-side `accept`/size guard — a pre-existing product gap from #643, and browser-side validation is bypassable anyway), and the `download_url` signed-URL TTL behavior.

## 4. Reviewer coverage

Cross-checked by an independent adversarial pass. One advisory leg raised a **P1 "reply attachment data-loss + deceptive toast"** — this was adjudicated a **false positive**: because `files = replyTo ? [] : pendingFiles` (line 360), the reply path never enters the upload loop, so `failedFiles` is always empty and the `commentAttachFailed` toast is unreachable for replies. Not reproducible against the actual code.

Not covered here (needs runtime/E2E, out of static-review scope): backend partial-failure semantics (possible orphaned attachments on the server), and visual/CSS of the re-staged chips.

PR #648, opened 2026-07-11
refactor(loop): funnel mutation-then-refresh handlers through one helper

MrAladdin:
## Why

Global pass after the #643 / #646 / #647 review churn kept surfacing the *same class* of bug in different `IssueDetailPage` write handlers. Rather than keep point-fixing, this converges them.

**The class:** a write and its follow-up refresh in one `try`, so a *refresh* failure gets reported as a *write* failure or strands state. A whole-component scan found:
- `reactComment` / `reactIssue` / `toggleResolve` — reaction/resolve succeeds but the refresh throws → misleading "save failed" toast + stale UI (non-data-loss, self-corrects, but wrong feedback).
- `removeComment` — **no `try/catch` at all**: a failed delete was silent (unhandled rejection, no feedback); a delete-success-but-reload-fail dropped the success toast.

(`submitComment` was the data-loss instance — fixed in #647. `saveEdit` / `patch` are minor and left out of scope.)

## What

Add one helper encoding the correct shape:

```ts
mutateThenRefresh(mutate, refresh, onOk?):
  try { await mutate() } catch(e){ Toast.error(e.message ?? saveFailed); return }  // only a WRITE failure toasts + aborts
  onOk?.()                                                                          // sync success handling (toast/state) — cannot throw
  try { await refresh() } catch {}                                                  // refresh failure never re-surfaces as a write failure
```

Converge the four handlers onto it:
- `reactComment` / `reactIssue` / `toggleResolve` — drop their per-handler `try`; a refresh failure after a successful reaction/resolve no longer shows "save failed".
- `removeComment` — now surfaces an error toast on a failed delete (was silent), and the success toast fires before the reload.

Token capture is unchanged (each handler still reads `reqRef.current` synchronously before calling the helper and passes it into `reloadComments`/`syncIssue`, whose own staleness guard is preserved).

- Reaction toggles through the helper with no spurious error (happy path intact).
- With `DELETE /comments/:id` forced to fail, deleting a comment now surfaces an error toast ("Network Error") instead of failing silently.

## Review gate

- Claude code-reviewer (skeptical): 6-point check — helper semantics, token capture timing, `removeComment` behavior change, `confirmDelete` call-site compatibility (helper returns `Promise<void>`, `onOk: () => void | Promise<void>` accepts it), `syncIssue` never-rejects interaction, no lost per-handler behavior — no real bug.
- codex adversarial-review: approve, no material findings.
- `/simplify`: helper abstraction appropriate/mergeable; applied its two comment-readability notes (dropped PR-history narration + a brittle rule-number reference).

## Blast radius

`packages/dmloop/src/panel/IssueDetailPage.tsx` only — one new in-component helper + four handlers reshaped; call sites unchanged (`confirmDelete({ onOk })` still compatible). No shared type/prop/signature change, no consumer outside dmloop.

## Docs

No doc change: internal error-handling consistency refactor, no new service/command/config/env. (The "fix the class, not the instance" principle this follows was added to the global dev rules separately.)

### DIFF
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -257,38 +257,55 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
     </Dropdown.Item>
   );
 
-  // 评论 emoji 反应:选择器点 emoji=加,已有 chip 点击=删自己那条(后端按 actor+emoji 定位)。
-  const reactComment = async (commentId: string, emoji: string, add: boolean) => {
-    const token = reqRef.current;
+  // 写操作与"写后刷新"分离的通用漏斗:只有写失败才报错并中止;写成功后先跑同步的成功处理
+  // (onOk:toast / 本地 state,不会抛),再刷新——刷新失败不当作写失败、不误报、也不 strand 已成功的写。
+  // 收敛这一类 async 写(reaction / resolve / 删评论),杜绝"写+刷新同一 try"导致刷新失败误报或漏处理
+  // (原则:修一类而非单点)。
+  const mutateThenRefresh = async (
+    mutate: () => Promise<unknown>,
+    refresh: () => Promise<unknown>,
+    onOk?: () => void,
+  ) => {
     try {
-      await (add ? addCommentReaction : removeCommentReaction)(commentId, emoji);
-      await reloadComments(token);
+      await mutate();
     } catch (e) {
       Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
+      return;
+    }
+    onOk?.();
+    try {
+      await refresh();
+    } catch {
+      /* 刷新失败:写已成功,不误报;下次 reload 自愈 */
     }
   };
 
-  // issue 级 emoji 反应:同评论,读回走 getIssue 的 issue.reactions(syncIssue 重取详情)。
-  const reactIssue = async (emoji: string, add: boolean) => {
+  // 评论 emoji 反应:选择器点 emoji=加,已有 chip 点击=删自己那条(后端按 actor+emoji 定位)。
+  const reactComment = (commentId: string, emoji: string, add: boolean) => {
     const token = reqRef.current;
-    try {
-      await (add ? addIssueReaction : removeIssueReaction)(issueId, emoji);
-      await syncIssue(token);
-    } catch (e) {
-      Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
-    }
+    return mutateThenRefresh(
+      () => (add ? addCommentReaction : removeCommentReaction)(commentId, emoji),
+      () => reloadComments(token),
+    );
+  };
+
+  // issue 级 emoji 反应:同评论,读回走 getIssue 的 issue.reactions(syncIssue 重取详情)。
+  const reactIssue = (emoji: string, add: boolean) => {
+    const token = reqRef.current;
+    return mutateThenRefresh(
+      () => (add ? addIssueReaction : removeIssueReaction)(issueId, emoji),
+      () => syncIssue(token),
+    );
   };
 
   // 评论 resolve/unresolve:后端「一线程至多一条 resolved」会清同线程兄弟,操作后重拉评论即可。
   // (resolve 只发实时事件、不写 activity_log,故活动流无需刷新。)
-  const toggleResolve = async (commentId: string, resolved: boolean) => {
+  const toggleResolve = (commentId: string, resolved: boolean) => {
     const token = reqRef.current;
-    try {
-      await (resolved ? unresolveComment : resolveComment)(commentId);
-      await reloadComments(token);
-    } catch (e) {
-      Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
-    }
+    return mutateThenRefresh(
+      () => (resolved ? unresolveComment : resolveComment)(commentId),
+      () => reloadComments(token),
+    );
   };
 
   // @提及:选中候选(成员/AI队友/AI小队)→ 往草稿插入 mention markdown。
@@ -418,11 +435,14 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
   const toggleSuppress = (id: string) =>
     setSuppressed((s) => { const n = new Set(s); if (n.has(id)) n.delete(id); else n.add(id); return n; });
 
-  const removeComment = async (id: string) => {
+  const removeComment = (id: string) => {
     const token = reqRef.current;
-    await deleteComment(id);
-    await reloadComments(token);
-    Toast.success(t("loop.toast.commentDeleted"));
+    // 删除失败→报错(否则静默);删除成功→先弹 commentDeleted(刷新前),再刷新评论列表。
+    return mutateThenRefresh(
+      () => deleteComment(id),
+      () => reloadComments(token),
+      () => Toast.success(t("loop.toast.commentDeleted")),
+    );
   };
 
   const saveEdit = async (id: string) => {


OctoBoooot (approved):
## Review: funnel mutation-then-refresh handlers through one helper (#648)

**Verdict:** Approve
Clean "fix the class, not the instance" refactor. The `mutateThenRefresh` helper encodes exactly the correct error-handling shape the #643/#646/#647 bugs kept violating, and each of the four converged handlers preserves its behavior (with `removeComment` genuinely improved). Verified the helper and every call site.

### What I verified
- **Helper shape is correct:** `try { await mutate() } catch { Toast.error; return }` (only a *write* failure toasts + aborts) → `onOk?.()` (synchronous success handling, runs after write, **before** refresh, can't throw) → `try { await refresh() } catch {}` (refresh failure swallowed, never re-surfaces as write failure). This bakes in the #647 ordering lesson (success state before the throwing refresh) and the #646 refresh-vs-mutation distinction, at the helper level.
- **Behavior-preserving per handler:**
  - `reactComment`/`reactIssue`/`toggleResolve` — old single-`try` (refresh failure → spurious "save failed") replaced by the helper with no `onOk`; a refresh failure after a successful reaction/resolve no longer false-toasts. Token captured synchronously before the helper call, passed into `reloadComments(token)`/`syncIssue(token)` — staleness guard preserved.
  - `removeComment` — previously had **no try/catch** (failed delete = silent unhandled rejection; reload-fail dropped the success toast). Now: delete errors toast, and `onOk = Toast.success(commentDeleted)` fires before the reload (survives a reload failure). Real improvement, correct shape.
- **Correctly scoped out:** the helper is *not* applied to `submitComment` (the #647 data-loss case with its richer per-file `failedFiles` re-stage) or `patch`/`saveEdit` — so the refactor can't flatten submitComment's careful re-stage-before-reload ordering. Right call.
- **Call-site compat:** `removeComment` via `confirmDelete({ onOk })` — helper returns `Promise<void>`, `onOk` accepts `() => void | Promise<void>`; still compatible.

### Praise
- This is the right response to a recurring bug class: rather than keep point-patching, it extracts the exact invariant the batch kept getting wrong — write-failure aborts, success handling is synchronous and runs *before* the refresh, refresh-failure is swallowed — into one helper, so the shape can't be re-violated per-handler. And it's disciplined about scope: the two handlers whose logic doesn't fit the 3-arg shape (submitComment's re-stage, saveEdit/patch) are deliberately left out rather than force-fit, so no hard-won ordering (esp. #647's) is disturbed. `removeComment` going from no-error-handling to correctly-shaped is a bonus catch from the same scan.

Jerry-Xin (approved):
APPROVE — clean, behavior-preserving consolidation that also lands the intended feedback fix, with no regression of #646/#647.

Byte-verified the real PR delta (head `b086745` vs its parent `2c0dd31`, the #647 fix): this PR touches exactly one file, `packages/dmloop/src/panel/IssueDetailPage.tsx`, adding the `mutateThenRefresh` helper and funneling four handlers through it. Nothing else.

**Helper contract (verified):**
```
mutateThenRefresh(mutate, refresh, onOk?):
  try { await mutate() } catch(e){ Toast.error(e.message ?? saveFailed); return }  // write-failure only: toast + abort
  onOk?.()                                                                          // sync success handling (toast/state) — before any refresh, cannot throw
  try { await refresh() } catch {}                                                  // refresh-failure swallowed, never re-surfaces as a write failure
```
Phase order is mutate → onOk (sync, before the throwing await) → refresh. `onOk` runs BEFORE `refresh`, so there is no #647-style "preserve-after-throwing-await" gap.

**Per-call-site check (all four consolidated handlers):**
- `reactComment` / `reactIssue` / `toggleResolve` — previously `try { await mutate; await refresh } catch { Toast.error }`, so a *refresh* failure falsely showed "save failed". Now a post-success refresh failure is swallowed. Intended fix, matches the PR; no preserve-state semantics exist for these, so nothing is dropped. ✅
- `removeComment` — previously had **no** try/catch (silent delete failure; success toast lost on reload-fail). Now: delete-fail → error toast (improvement), delete-success → success toast fires via `onOk` (before refresh) → refresh swallowed. ✅

**#646 / #647 invariants — all preserved because they live in UNTOUCHED handlers (grep confirms 0 diff lines touch them):**
- A (`toggleSubscribe`): still hand-written; refresh-failure still does `setSubLoaded(false)` (unknown fallback). Not routed through the helper. ✅
- B (`submitComment`): still re-stages failed files (`setPendingFiles(failedFiles)` + accurate toast) BEFORE the throwing `await reloadComments`. Untouched. ✅
- C (`uploadForIssue`): still per-file try/catch inside the loop. Untouched. ✅
- D (mutation-failure preserves input): the helper's `catch → toast → return` matches the #646 accurate-toast + early-return shape; `submitComment`/`saveEdit` retain their own draft-preserving early returns. ✅

Token/nav-race guards (`reqRef.current` captured synchronously before the mutation and passed into `reloadComments`/`syncIssue`) and busy guards (`submitting`/`uploading` finally) are unchanged.

**Non-blocking:**
- `saveEdit` (edit-comment) still has the "reload can throw before the success toast" feedback-class shape. The PR explicitly scopes this out ("`saveEdit` / `patch` are minor and left out of scope"), and it is pre-existing, not a regression introduced here — fine to defer.

No semantic regression beyond the intended error-path consolidation. Approving.

yujiawei (approved):
# Code Review — PR #648 (octo-web)

`refactor(loop): funnel mutation-then-refresh handlers through one helper`

Reviewed against head `b086745` with a fresh checkout. Scope: a single file, `packages/dmloop/src/panel/IssueDetailPage.tsx` (+43 / −23). This review is split into a spec-compliance pass and a code-quality pass.

## 1. Spec compliance

The diff does exactly what the description claims and nothing more:

- Adds one in-component helper `mutateThenRefresh(mutate, refresh, onOk?)` encoding "only a write failure toasts + aborts; refresh failure is swallowed" (`IssueDetailPage.tsx:264-281`).
- Converges the four handlers onto it: `reactComment` (`:284`), `reactIssue` (`:293`), `toggleResolve` (`:303`), `removeComment` (`:438`).
- `submitComment`, `saveEdit`, and `patch` are left untouched, matching the stated out-of-scope list.

No new service/prop/type/signature, no consumer outside `dmloop`, call sites unchanged. `confirmDelete({ onOk })` still receives a `() => Promise<void>`, which its signature (`onOk: () => void | Promise<void>`, `confirmDelete.ts:11`) accepts. **Spec: ✅**

## 2. Code quality

The refactor is correct and is a genuine behavior improvement, not just a move:

- **Token/staleness capture preserved.** Each handler still reads `reqRef.current` synchronously before calling the helper and threads it into `reloadComments(token)` / `syncIssue(token)`, whose `token === reqRef.current` guards (`:198`, `:201-204`) are intact. Capturing (not incrementing) `reqRef.current` is the right choice here — only `reload()` starts a new generation, so a mid-flight issue switch correctly drops the stale refresh write.
- **Refresh-failure swallowing is safe.** `reloadComments` can reject (`await listComments`) and the outer `try { await refresh() } catch {}` absorbs it; `syncIssue` already self-catches so the wrapper is merely redundant for it, not harmful. The successful write is never re-reported as a failure, and the UI self-heals on the next reload — this is strictly better than the old code, which showed a misleading "save failed" toast on a refresh error after a successful reaction/resolve.
- **`removeComment` is the real fix.** Previously it had no `try/catch`: a failed `DELETE` rejected silently (unhandled rejection, no feedback, confirm modal left in a rejected `onOk` state). It now surfaces an error toast on failure and fires the success toast *before* the reload, so a reload failure no longer swallows the "deleted" confirmation. Good.
- **Success ordering.** Firing `onOk` (the success toast) after the write succeeds but before the refresh is intentional and correct for delete — the delete truly succeeded regardless of whether the list re-fetch does.

### Non-blocking observations (nits)

- `onOk?.()` (`:275`) is invoked outside any `try`, with the contract "sync success handling — cannot throw." That holds for the only current caller (`Toast.success`, `:444`). If a future `onOk` ever throws, it would propagate out of the helper and skip the refresh. Consider documenting this as a hard precondition or wrapping it defensively. Not actionable for this PR.
- No optimistic UI on reactions: the chip count only updates after the refresh, so a refresh failure leaves the count visually stale until the next reload. This matches prior behavior (minus the false error toast) and is non-data-loss, self-correcting — noted only for completeness.

**Quality: Approved**

## 3. Overall verdict

**APPROVED** — Spec ✅ AND Quality Approved. No P0/P1/P2 blocking findings.

## Security note (for the human reviewer)

This PR was auto-routed `security_sensitive` / `needs-human-review`, but that appears to be a keyword false-positive on "token": the only "token" here is `reqRef`, the client-side staleness/generation counter — not an auth, session, or credential token. The diff touches no authentication, middleware, migration, or secret-handling surface; it is a frontend error-handling consistency refactor. Nothing here requires manual security verification beyond confirming the staleness guard is preserved (it is).

Note: an independent cross-model advisory pass was attempted but its gateway was unavailable this run; it is recorded as absent, not as agreement. The review above stands on a direct read of the diff and surrounding file.

PR #654, opened 2026-07-11
fix(loop): invalidate directory cache on member/agent/squad/project mutations

MrAladdin:
## Bug

In the dmloop Settings → Member management flow, after **adding** a workspace member the new person was missing from the "new issue" (新建回路) assignee dropdown, and after **removing** a member the removed person was still shown — both until a full reload.

## Root cause

The `Directory` cache in `packages/dmloop/src/api/directory.ts` (a module-level singleton holding member/agent/squad assignee candidates + project names, consumed by assignee dropdowns, filters, and list/board enrich) was only invalidated on **workspace switch**. Member add/remove — and, by the same class, agent/squad/project create/update/delete — never invalidated it, so every consumer read stale data.

## Fix (structural, one shared sink — Rule 9)

Add `afterDirectoryMutation<T>(result: T): T { invalidateDirectory(); return result; }` and chain it via `.then()` onto the directory-relevant mutation API functions, so the cache is cleared on success and the next read (fresh picker mount / filter render / enrich) rebuilds:

- `workspaceApi.ts`: `addOctoMember`, `removeMember`
- `agentApi.ts`: `createAgent`, `updateAgent`, `archiveAgent`
- `squadApi.ts`: `createSquad`, `updateSquad`, `deleteSquad`
- `projectApi.ts`: `createProject`, `updateProject`, `deleteProject`

`inviteMember` / `updateMemberRole` / `revokeInvitation` are intentionally **not** wrapped — invitations aren't materialized members and role changes don't affect cached name/candidate data.

Also closes a cache race the new invalidation points expose: `ensureDirectory`'s build-commit is now guarded by a generation counter, so an in-flight `build()` started **before** a mutation cannot repopulate `_cache` with pre-mutation data after `invalidateDirectory()` clears it. (Confirmed by adversarial review.)

## Verification

- **Real browser** (Loop-dev workspace): add member "Bob" → 新建回路 assignee dropdown shows `[未指派, Alice, Bob]` with no reload; remove Bob → dropdown shows `[未指派, Alice]`. Both halves of the reported bug fixed. Test data restored (member list back to 1).
- **Adversarial review** (Claude code-reviewer): its one finding was the in-flight-build race, already fixed by the generation guard; type/value preservation, error propagation, and wrapped-vs-unwrapped coverage all confirmed correct.
- `tsc -p packages/dmloop`: 0 errors.

### DIFF
--- a/packages/dmloop/src/api/agentApi.ts
+++ b/packages/dmloop/src/api/agentApi.ts
@@ -1,7 +1,7 @@
 // @octo/loop — Agent API（后端契约联调）
 import type { Agent, CreateAgentReq, UpdateAgentReq, ListParams, RuntimeDevice } from "./types";
 import { httpGet, httpPost, httpPut, httpDelete } from "./http";
-import { ensureDirectory, actorName, actorAvatar } from "./directory";
+import { ensureDirectory, actorName, actorAvatar, afterDirectoryMutation } from "./directory";
 
 // runtime 名字缓存（用于 agent.runtime_name 回填）
 let _runtimeMap: Map<string, string> | null = null;
@@ -39,16 +39,16 @@ export async function getAgent(id: string): Promise<Agent> {
 }
 
 export function createAgent(req: CreateAgentReq): Promise<Agent> {
-  return httpPost<Agent>("/agents", req);
+  return httpPost<Agent>("/agents", req).then(afterDirectoryMutation);
 }
 
 export function updateAgent(id: string, req: UpdateAgentReq): Promise<Agent> {
-  return httpPut<Agent>(`/agents/${id}`, req);
+  return httpPut<Agent>(`/agents/${id}`, req).then(afterDirectoryMutation);
 }
 
 // 后端不支持 DELETE /agents/:id（405）；改用归档。
 export function archiveAgent(id: string): Promise<void> {
-  return httpPost<void>(`/agents/${id}/archive`, {});
+  return httpPost<void>(`/agents/${id}/archive`, {}).then(afterDirectoryMutation);
 }
 
 /* ---------- 环境变量（密钥） ---------- */
--- a/packages/dmloop/src/api/directory.ts
+++ b/packages/dmloop/src/api/directory.ts
@@ -19,6 +19,11 @@ interface Directory {
 
 let _cache: Directory | null = null;
 let _loading: Promise<Directory> | null = null;
+// generation:每次 invalidate 自增。in-flight build 只在自己启动时的 gen 未被 invalidate
+// 打断时才提交到 _cache —— 否则(mutation/切 workspace 在 build 途中发生)build 会把
+// **失效前的陈旧数据**写回 _cache,重新污染。等待者仍拿到本次 d(自身发起早于 mutation),
+// 但不毒害后续读。
+let _gen = 0;
 
 async function build(): Promise<Directory> {
   const slug = currentWorkspaceSlug();
@@ -67,9 +72,14 @@ export async function ensureDirectory(force = false): Promise<Directory> {
   // 已有 in-flight build 时并发调用者直接搭车(切 workspace 会 invalidateDirectory 清空 _loading),
   // 避免冷启动(_cache 尚为 null)多个消费者各拉一遍 directory。
   if (!force && _loading) return _loading;
+  const myGen = _gen;
   _loading = build().then((d) => {
-    _cache = d;
-    _loading = null;
+    // 只有本次 build 未被 invalidate 打断才提交并清 _loading;被打断则不写 _cache
+    // (下次 ensureDirectory 见 _cache/_loading 均已被 invalidate 清空 → 重新 build)。
+    if (myGen === _gen) {
+      _cache = d;
+      _loading = null;
+    }
     return d;
   });
   return _loading;
@@ -78,6 +88,15 @@ export async function ensureDirectory(force = false): Promise<Directory> {
 export function invalidateDirectory(): void {
   _cache = null;
   _loading = null;
+  _gen++;
+}
+
+// 目录相关实体(member/agent/squad/project)的写操作**成功后**链在 .then 上:清缓存,
+// 下次读(打开新建弹窗、渲染筛选/看板 enrich)即重建 → 增删改后候选/名字/项目选项不再陈旧。
+// 放在共享 API 层而非各 UI 处,一处收口护全部消费者(Rule 9)。
+export function afterDirectoryMutation<T>(result: T): T {
+  invalidateDirectory();
+  return result;
 }
 
 export function actorName(
--- a/packages/dmloop/src/api/projectApi.ts
+++ b/packages/dmloop/src/api/projectApi.ts
@@ -1,7 +1,7 @@
 // @octo/loop — Project API（后端契约联调）
 import type { Project, UpsertProjectReq, ListParams } from "./types";
 import { httpGet, httpPost, httpPut, httpDelete } from "./http";
-import { ensureDirectory, actorName } from "./directory";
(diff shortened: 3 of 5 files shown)


OctoBoooot (approved):
## Review: invalidate directory cache on member/agent/squad/project mutations (#654)

**Verdict:** Approve
Clean structural fix for a real stale-cache bug (added/removed members not reflected in assignee dropdowns until reload). One shared invalidation sink, correctly scoped, type/error-preserving, and — notably — it closes the in-flight-build race the new invalidation points expose with a generation counter. Verified each piece.

### What I verified
- **`afterDirectoryMutation<T>(result): T { invalidateDirectory(); return result }`** — chained via `.then()` on **success** of each directory-affecting mutation. Type-preserving (returns the value unchanged), error-preserving (a rejected mutation skips `.then` → no invalidation on failure, correct since nothing changed server-side). As a bare `.then` ref, it receives the resolved value as arg 1 and ignores `.then`'s index/array args — correct.
- **Correct scope:** wrapped on `addOctoMember`/`removeMember`, `createAgent`/`updateAgent`/`archiveAgent`, `createSquad`/`updateSquad`/`deleteSquad`, `createProject`/`updateProject`/`deleteProject`. **Not** wrapped on `updateMemberRole` (role change doesn't alter cached name/candidate data), `inviteMember`/`revokeInvitation` (invitations aren't materialized members) — invalidate only where the cached member/agent/squad candidates + project names actually change, no over-invalidation.
- **Generation guard closes the in-flight-build race (the subtle part):** `ensureDirectory` captures `myGen = _gen` before `build()`; the `.then` commits `_cache = d` only `if (myGen === _gen)`; `invalidateDirectory` does `_cache=null; _loading=null; _gen++`. So a `build()` started **before** a mutation cannot repopulate `_cache` with pre-mutation data after the invalidate (its `myGen !== _gen` → commit skipped; next `ensureDirectory` sees null cache/loading → fresh build). Waiters on the original `_loading` still get their pre-mutation `d` (acceptable — they requested before the mutation) but the cache is not poisoned for subsequent reads. Traced the interleaving; sound.

### Praise
- This is the right shape for a cache-staleness bug: one shared `afterDirectoryMutation` sink at the API layer (not per-UI), scoped precisely to the mutations that change cached data (with the invite/role exclusions reasoned out), so every consumer — pickers, filters, list/board enrich — gets fresh data with no per-call-site work. And it doesn't stop at "add invalidate calls": it recognizes that new invalidation points *create* a build-commit race and closes it with a generation counter — exactly the concurrency care this module's `ensureDirectory` needs (consistent with the guard added in #627). Structural fix, not a point patch.

Jerry-Xin (approved):
The PR is relevant to `Mininglamp-OSS/octo-web` and fixes a real dmloop cache invalidation bug in the shared API layer.

🔴 Blocking

None.

💬 Non-blocking

- 🔵 Suggestion: Add focused tests for `ensureDirectory` invalidation and the generation guard in `packages/dmloop/src/api/directory.ts:70`. This is race-sensitive behavior, and automated coverage would help protect the fix.
- 🔵 Suggestion: Consider a future hardening pass for rejected directory builds around `packages/dmloop/src/api/directory.ts:76`. The current `build()` largely swallows request failures, so this is not blocking, but a rejection handler/finally would prevent `_loading` from ever being left in a bad state if a new uncaught failure path is introduced.

✅ Highlights

- The shared `afterDirectoryMutation` sink in `packages/dmloop/src/api/directory.ts:97` keeps invalidation centralized and preserves mutation return values.
- The added wrappers cover the directory-relevant member, agent, squad, and project mutations in `workspaceApi.ts`, `agentApi.ts`, `squadApi.ts`, and `projectApi.ts`.
- The generation counter prevents an older in-flight build from repopulating `_cache` after invalidation, which addresses the main stale-cache race.

yujiawei (approved):
# Code Review — PR #654 (octo-web)

Head SHA reviewed: `27b9db5ec250dad7a2345f0fe5c6d38a49eab414`

## Summary

Small, well-scoped fix for a real stale-cache bug: the module-level `Directory` singleton in `packages/dmloop/src/api/directory.ts` (member/agent/squad names, assignee candidates, project names) was only invalidated on workspace switch, so directory-affecting mutations left assignee dropdowns / filters / list enrich stale until a full reload. The fix adds a single shared sink `afterDirectoryMutation<T>` chained onto the relevant mutations, plus a generation counter to close a concurrency window. Reviewed the full file, all five changed API modules, and the directory consumers (`AssigneePicker`, `IssuePage`, `SquadPage`, `SettingsPage`, `SquadDetailPage`, `LoopPage`).

- Implements exactly what the description states: cache invalidation on member/agent/squad/project create/remove/update, one shared sink at the API layer.
- No scope creep — every hunk is directly tied to the fix. The generation guard is a legitimate, necessary companion (see below).
- Intentional exclusions (`inviteMember`, `updateMemberRole`, `revokeInvitation`, and squad-membership mutations) are correct: none of them mutate any field the `Directory` actually caches (names / candidates / projectName).

## 2. Correctness — Approved

**Coverage is complete.** Cross-checked every mutation in the dmloop API layer against the cached `Directory` fields (`memberName`, `memberOctoUid`, `agentName`, `squadName`, `projectName`, `candidates`):

- Wrapped set (`addOctoMember`, `removeMember`, `createAgent`, `updateAgent`, `archiveAgent`, `createSquad`, `updateSquad`, `deleteSquad`, `createProject`, `updateProject`, `deleteProject`) all mutate a cached field — correctly wrapped.
- Unwrapped set only touches non-cached state: invitations are not materialized members; role changes don't affect names/candidates; squad-membership changes affect a squad's `members` (not the cached `squadName`) and re-enrich via an independent `getSquad()` → `ensureDirectory()`.

No missing invalidations, no harmful over-invalidations.

**Concurrency guard is sound.** The `_gen` counter correctly prevents a `build()` started before an `invalidateDirectory()` from committing pre-mutation data back into `_cache`: on `myGen !== _gen` the callback skips both the `_cache` write and the `_loading = null` clear, so it cannot clobber a newer in-flight build's `_loading` either. No deadlock risk — `build()` internally `.catch()`es every sub-fetch, so it never rejects and `_loading` can't get stuck on a rejected promise. A concurrent rider that requested before the mutation still receives the (now-superseded) directory it asked for, which is the expected semantics and does not poison later reads.

**Error and type handling correct.** `.then(afterDirectoryMutation)` runs only on fulfillment, so a failed mutation does not wrongly invalidate the cache. `afterDirectoryMutation<T>` preserves the return type for both value-returning (`Agent`, `Squad`, `Project`, `WorkspaceMember`) and `void` mutations.

**Consumers refresh correctly.** Directory consumers load candidates/projects in `useEffect` on mount / modal-open, so the next read after an invalidation rebuilds fresh data — which is precisely the reported flow (add/remove member → reopen "new issue" picker shows correct list).

## 3. Non-blocking suggestions (not required for merge)

- **nit** — `workspaceApi.ts` `updateMemberRole` (`~L51-57`): `addOctoMember` and `removeMember` now carry an explanatory comment about their cache impact; a one-line note on `updateMemberRole` explaining *why* it is intentionally **not** wrapped (role-only, no name/candidate impact) would keep the "wrapped vs not" decision self-documenting for future maintainers.
- **P2** — `squadApi.ts` `addSquadMember` / `removeSquadMember` (`~L55-72`): similarly, a brief comment noting these change squad membership (not `squadName`) and re-enrich via `getSquad()`, hence no `afterDirectoryMutation`, would prevent a future maintainer from "fixing" a non-bug by wrapping them.

Both are documentation-only; the behavior is already correct.

## Verdict

No P0/P1 issues. Coverage, concurrency, error handling, and type preservation all verified correct against the current head. Approving.

PR #653, opened 2026-07-11
feat(loop): Tier1 issue browse/edit — grouped board, scope pill, running chip, sub-issue create, batch ops

MrAladdin:
## What

Tier1 of the dmloop (`@octo/loop`) issue browse/edit parity work. Five cohesive capabilities, all backed by existing octo-multica endpoints (frontend-only, no backend/DB change):

1. **Grouped-by-assignee board** — `GET /issues/grouped?group_by=assignee`.
2. **Scope pill** (All / Members / AI / Involves me). "Involves me" fans out three server filters (`assignee_id` ∪ `creator_id` ∪ `involves_user_id`) and merges/dedupes client-side, because the backend `involves_user_id` filter is by design disjoint from direct assignment/creation (mirrors the native `fetchAllMyAssigneeGroups`).
3. **Agent-running chip** — from `GET /agent-task-snapshot` (one workspace-level snapshot → client-derived running-issue set; no per-card N+1). 15s poll + sequence guard.
4. **Sub-issue create entry** — `parent_issue_id` on create, from the issue detail sub-issue section.
5. **Batch update/delete** — `POST /issues/batch-update|batch-delete` via Semi `Table` rowSelection; selection pruned to visible rows on filter/sort/page change; `runBatch` reentrancy guard.

Shared `IssueCard` + `RunningChip` extracted (dedupes board/list/grouped rendering).

Also folded in (frontend-only brand hygiene): renamed `Multica*` CLI-authorize identifiers/files to `Loop*` (route value `/loop/cli-authorize` unchanged; cross-repo `MULTICA_APP_URL` env / `multica_csrf` cookie / `/fleet` path deferred — they need backend/CLI coordination). Added `packages/dmloop/src/vite-env.d.ts` so the package typechecks clean.

## Verification (real browser, MV2 E2E workspace, real backend)

- **Grouped board**: issues grouped under assignee "E2E Coworker" + an "Unassigned" group.
- **Scope pill**: All (3) → Members (0, all agent-assigned) → Involves me (2, deduped union). After creating an unassigned sub-issue authored by me, it appears in "Involves me" **only via the creator leg** — confirming the 3-way union is live (not involves-only).
- **Running chip**: rendered on the issue with an active agent task, across board / grouped / list.
- **Batch**: multi-select 2 issues, batch priority change → persisted (confirmed in the issue activity log). Selecting a row then applying a status filter that hides it prunes the selection (batch bar clears — no mutation of hidden rows).
[...]

PR #655, opened 2026-07-11
feat(loop): Tier2 grouped-board multi-select filters + include-no-project

MrAladdin:
## What

Tier2 of the dmloop (`@octo/loop`) issue browse/edit parity work: **multi-select filters + no-project** on the grouped-by-assignee board.

- **Multi-select** status / priority / project — on the grouped board these become Semi `<Select multiple>`, backed by `/issues/grouped`'s `statuses[]` / `priorities[]` / `project_ids[]`. Board/list views keep single-select (backend flat `/issues` supports neither multi status/priority nor multi project). Grouped multi state (`gStatuses`/`gPriorities`/`gProjectIds`) is held separately from the single-select `status`/`priority`/`project` — no cross-read.
- **含无项目 (include_no_project)** checkbox — pairs with the project multi-select: selected projects ∪ no-project.

### Deliberately NOT added: an "include_no_assignee" toggle

The scope pill uses `assignee_types` (type-level, a top-level AND clause). The backend `include_no_assignee` is only an *include* when paired with `assignee_filters` (specific actors) — which this UI never sends. Combined with `assignee_types` (Members/Agents scope) or the "involves" fan-out's `assignee_id` leg, it ANDs to an empty set (silent empty board). The grouped board already surfaces the unassigned group under scope=all. Confirmed against the backend grouped handler; both `/simplify` and the adversarial review independently flagged this conflict.

## Contract

`GroupedParams` gains `project_ids` / `include_no_project`; `listGroupedIssues` sends them (`clean()` drops empty). `listMyGroupedIssues` (involves fan-out) keeps them in the base — the project axis is independent of the per-leg user-relation filters, so it composes correctly.

## Verification (real browser, MV2 E2E grouped board)

- Status multi-select `[审核中, 已完成]` → board shows exactly MVE-2 + MVE-1 (others hidden).
- `含无项目` checkbox toggles and results stay correct.
- (priority multi-select uses the same Semi `multiple` mechanism; project multi-select renders — test issues have no project so the include_no_project path is what's exercised.)
- `tsc -p packages/dmloop`: 0 errors.

## Review

- `/simplify` applied — hoisted the duplicated `Select.Option` lists into shared `statusOpts`/`priorityOpts`/`projectOpts` consts used by both the multi and single branches.
- Adversarial review (Claude code-reviewer + Codex): the only finding was the `include_no_assignee` × scope-pill empty-set conflict, already resolved by dropping that toggle. Grouped-vs-single field separation, involves fan-out interaction, and Semi multiple-onChange shape all confirmed correct. A clean Codex confirmation is pending (endpoint instability today) and will follow.
- `packages/dmloop` has no unit-test harness (whole package); regression is the real-browser e2e above.

### DIFF
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -113,6 +113,8 @@ export async function listGroupedIssues(params: GroupedParams): Promise<IssueGro
     involves_user_id: params.involves_user_id,
     creator_id: params.creator_id,
     project_id: params.project_id,
+    project_ids: params.project_ids?.join(","),
+    include_no_project: params.include_no_project ? "true" : undefined,
     date_field: params.date_field,
     date_start: params.date_start,
     date_end: params.date_end,
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -93,6 +93,10 @@ export interface GroupedParams {
   involves_user_id?: string;
   creator_id?: string;
   project_id?: string;
+  project_ids?: string[];
+  // 纳入无项目的 issue(后端 include_no_project)。与项目多选配对 = 所选项目 ∪ 无项目。
+  // (未加 include_no_assignee:仅与 assignee_filters 配对有意义,本 UI 用 assignee_types/involves。)
+  include_no_project?: boolean;
   date_field?: IssueDateField;
   date_start?: string;
   date_end?: string;
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -60,7 +60,8 @@
     "assignee": "Assignee",
     "creator": "Creator",
     "project": "Project",
-    "dateRange": "Date range"
+    "dateRange": "Date range",
+    "noProject": "Incl. no project"
   },
   "dateField": {
     "created_at": "Created",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -60,7 +60,8 @@
     "assignee": "负责人",
     "creator": "创建者",
     "project": "项目",
-    "dateRange": "时间范围"
+    "dateRange": "时间范围",
+    "noProject": "含无项目"
   },
   "dateField": {
     "created_at": "创建时间",
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -10,6 +10,7 @@ import {
   DatePicker,
   RadioGroup,
   Radio,
+  Checkbox,
 } from "@douyinfe/semi-ui";
 import { Search, Plus, LayoutGrid, List as ListIcon, Users, ClipboardList, ArrowUp, ArrowDown } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
@@ -46,11 +47,19 @@ function scopeToAssigneeTypes(scope: IssueScope): ("member" | "agent" | "squad")
 
 interface Filters {
   keyword: string;
-  status?: IssueStatus;
-  priority?: IssuePriority;
+  status?: IssueStatus;        // board/list 单选
+  priority?: IssuePriority;    // board/list 单选
   assignee?: string;
   creator?: string;
-  project?: string;
+  project?: string;            // board/list 单选
+  // 分组板专属多选 + no-project(仅 /grouped 支持 statuses[]/priorities[]/project_ids[]/include_no_project)。
+  // 注:无 include_no_assignee —— scope pill 用 assignee_types(类型级、顶层 AND),后端
+  // include_no_assignee 只与 assignee_filters(具体 actor)配对才是「含」语义;与 assignee_types/
+  // involves 的 assignee_id 腿 AND 会矛盾成空。grouped 在 scope=all 下本就显示未指派组。
+  gStatuses: IssueStatus[];
+  gPriorities: IssuePriority[];
+  gProjectIds: string[];
+  noProject: boolean;
   dateField: IssueDateField; // 时间范围筛选的列(created_at|updated_at)
   dateRange?: Date[];        // [start, end];为空则不按时间筛选
   sortBy: IssueSortField;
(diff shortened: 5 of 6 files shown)


OctoBoooot (approved):
## Review: Tier2 grouped-board multi-select filters + include-no-project (#655)

**Verdict:** Approve
Clean filter expansion. Multi-select status/priority/project + include-no-project on the grouped board, correctly kept separate from the flat views' single-select. Verified the param encoding, the view-mode separation, and the fan-out composition.

### What I verified
- **Param encoding:** `project_ids?.join(",")` (CSV, consistent with existing `statuses`/`priorities`/`assignee_types`), `include_no_project ? "true" : undefined` (sentinel only when set); `clean()`/undefined drops empties. Types additive.
- **Grouped-multi vs single-select separation (the key point):** `gStatuses`/`gPriorities`/`gProjectIds`/`noProject` are distinct from `status`/`priority`/`project` with **no cross-read** — the grouped branch sends only the `g*` fields, the flat board/list branch only the single-select ones. Correct, since the flat `/issues` backend supports neither multi-status/priority nor multi-project; separation prevents sending an unsupported multi param to the flat endpoint. Empty array → not sent (no unintended narrowing).
- **Involves fan-out composes:** `project_ids`/`include_no_project` stay in `base` (untouched by `listMyGroupedIssues`, which strips only user-relation filters per #653) → the project axis applies across all 3 legs correctly.
- **UI:** `<Select multiple>` with `onChange` normalizing `(v ?? [])` (clear→undefined→[] , no null-deref), grouped-only render; `statusOpts`/`priorityOpts`/`projectOpts` extracted (the /simplify dedup), escaped JSX text — XSS-safe.

### Praise
- The view-mode discipline is the right call and easy to get wrong: grouped multi-state is held in *separate* fields with no cross-read, so switching to board/list can never leak a multi-value param the flat endpoint rejects. And the decision to **not** add `include_no_assignee` — because it only means "include" when paired with `assignee_filters` (never sent here) and otherwise ANDs to a silently-empty board — is exactly the kind of backend-contract reasoning that avoids shipping a footgun; omitting it is more correct than adding it.

Jerry-Xin (approved):
The PR is in scope for `octo-web` and the grouped-board filter changes look contract-aligned; no blocking correctness or security issues found.

💬 Non-blocking

- 🟡 Warning: [packages/dmloop/src/pages/loop.css:990](packages/dmloop/src/pages/loop.css:990) and [packages/dmloop/src/pages/loop.css:991](packages/dmloop/src/pages/loop.css:991) add hard-coded spacing/font values. Project guidance prefers design tokens for new CSS. Consider replacing `10px` and `12px` with existing spacing/text tokens if available.
- 🔵 Suggestion: [packages/dmloop/src/pages/IssuePage.tsx:11](packages/dmloop/src/pages/IssuePage.tsx:11) adds Semi `Checkbox` directly while the repo has a project checkbox wrapper. This page already uses Semi controls heavily, so this is not a blocker, but using the wrapper would be more consistent with the component guidance.

✅ Highlights

- The grouped-only multi-select state is kept separate from board/list single-select state, avoiding accidental cross-view filter leakage.
- `project_ids` and `include_no_project` are serialized cleanly through the grouped endpoint, and empty arrays are omitted by the existing query cleaner.
- The `involves` fan-out keeps project filtering in the shared base while stripping user-relation filters, which matches the intended independent project axis.

Verification note: `git diff --check` passed. I could not run `pnpm i18n:check` because this worktree has no `node_modules` and the script could not resolve `typescript`.

yujiawei (approved):
# Code Review — PR #655 (octo-web)

**Scope:** Tier2 grouped-board multi-select filters (status / priority / project) + `include_no_project` checkbox on the grouped-by-assignee board. 6 files, +125/-37, all in `packages/dmloop`.

**Verdict: APPROVED** — no correctness or blocking issues. Two non-blocking notes below (one cosmetic, one UX-by-design).

---

## Spec compliance — ✅

- **Not missing:** `GroupedParams` gains `project_ids` / `include_no_project`; both are sent by `listGroupedIssues` and preserved across all three legs of the `listMyGroupedIssues` (“involves”) fan-out. Grouped view renders multi-select `<Select multiple>` for status/priority/project + the checkbox; board/list keep single-select. i18n key `loop.filter.noProject` added to both `en-US` and `zh-CN`.
- **Not over-built:** the deliberate omission of an `include_no_assignee` toggle is correct. Paired only with `assignee_types` (scope pill) or the involves `assignee_id` leg, the backend `include_no_assignee` ANDs to an empty set — it would silently blank the board. Dropping it is the right call, and the unassigned group is already visible under scope=all.
- **No divergence:** grouped multi-state (`gStatuses`/`gPriorities`/`gProjectIds`/`noProject`) is held separately from the single-select `status`/`priority`/`project` with no cross-read. The board/list path still sends `project_id: f.project` (`IssuePage.tsx:145`) — single-select filtering is not regressed.

## Code quality — Approved

Data flow verified end-to-end: empty `gProjectIds` → `project_ids` omitted via the `.length` guard + `clean()` dropping `undefined`; `include_no_project` serialized as `"true"` only when checked, otherwise omitted. `(v ?? [])` casts defensively handle Semi's `showClear`. The `statusOpts`/`priorityOpts`/`projectOpts` hoist is a clean dedupe of the single/multi branches with no behavioral change. The `seq` guard still protects against out-of-order responses.

### Non-blocking notes

- **P2 (nit) — stale comment.** `packages/dmloop/src/pages/loop.css:986`: the block comment reads `分组板 no-* 复选(含无负责人/无项目)`, referencing a **no-assignee** checkbox that was intentionally not built (see `IssuePage.tsx:56-58`, `types.ts:98`). Suggest narrowing it to `/* 分组板 no-project 复选 */` so it doesn't imply no-assignee support exists.
- **P2 (UX, by design) — filters reset across view switch.** Single-select (board/list) and grouped multi-select filters are separate state, so toggling views starts the other view's filters empty rather than carrying them over. This is intentional per-view isolation (the incompatible controls are conditionally rendered, so no filter is silently applied-but-hidden), and it's documented in the diff. Flagging only as a future UX consideration — if users expect a set status filter to survive a board→grouped switch, a one-way sync could be added later. Not required for this PR.

### Coverage / blind spots

- Backend semantics of `project_ids` (OR within the project axis) and `include_no_project` (selected ∪ no-project) are assumed from the contract and the PR's browser E2E; not verifiable from the frontend diff.
- `packages/dmloop` ships no unit-test harness; regression rests on the real-browser E2E described in the PR.
- One cross-family advisory reviewer was unavailable this run (endpoint instability); findings above reflect the available review legs plus an independent line-level pass.

PR #656, opened 2026-07-11
feat(versions): unified VersionHistoryPanel shell + race guard util

ploy-elison:
## Summary

Phase 1 of the version-history & preview unification (design in XIN-836, approved). This adds the shared shell and its race-guard primitive **only** — no end (doc / sheet / board) is wired to it yet, so there is **no user-visible change**. The doc/sheet/board panels are untouched and keep working exactly as before; adapter phases follow separately and can each be reverted independently. `versions/api.ts` and `versions/format.ts` are unchanged; the doc `previewGuard` is intentionally kept this phase (its removal is deferred to the doc-adapter phase).

**Diff is purely additive: 4 new files (2 source + 2 test), zero modified files.**

### What's added

- **`versions/VersionHistoryPanel.tsx`** — a generic `<VersionHistoryPanel<TState, TCurrent>>` that owns everything identical across the three ends:
  - single mixed list, reverse-chronological, with `all / manual / auto` filter tabs (board `filterBtn` pattern), a per-kind counts header (`manual+restore · auto`), and cursor-paginated load-more;
  - inline save-current and inline rename compose rows (collapsed button → input + save/cancel; no native `prompt`);
  - restore: row action → **centered in-panel confirm box** → `restoreVersion` (reuses `versions/api.ts`, no new endpoint) → non-destructive reconcile, live surface refreshes via Yjs;
  - centered preview / diff modal reusing the doc `octo-modal-overlay` (overlay-close, `stopPropagation`, Escape, `aria-modal`).
  - Props contract (per XIN-836 / PM spec): `docId`, `role`, `names?`, `onClose?`, `pageSize?` (default 30), `defaultFilter?` (default `all`), `onRestored?`, `previewErrorKey?`, `restoreErrorKey?`, `loadPreviewState(seq, signal)`, `renderPreview(state)`, `renderDiff?(state, current)`, `getCurrent?()`. When `renderDiff` is omitted the modal hides the compare entry.
  - Permissions gate internally on `canSnapshot(role)` (save/rename) and `canRestoreVersion(role)` (restore/delete).

- **`versions/raceGuard.ts`** — a reusable `createRaceGuard()` (AbortController + monotonic generation) consolidating the board panel's pattern into one primitive covering the list-refresh / load-more / preview chains: superseded in-flight requests are **truly aborted on the wire**, and a response that resolves after a newer request started is **discarded**. Strict superset of the doc panel's token-only preview guard (`begin` = primary lane; `beginFollowUp` = load-more bound to the current generation).

### i18n (deferred by design)

The shell references a few keys not yet in the `docs.version.*` namespace (`filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice`). They are **intentionally not added in this phase** to keep the diff to new files + tests only; the shell is not mounted anywhere yet, so no user sees a raw key. These keys land in the doc-adapter phase when an end first renders the shell. (Noted in the file header.)

### Tests (red → green gate)

- `raceGuard.test.ts` — abort-on-supersede + last-write-wins for the primary and follow-up lanes, and `abort()` cleanup.
- `VersionHistoryPanel.test.tsx` — list / filter / counts / load-more, centered preview modal (open, Escape, overlay-close, `stopPropagation`), compare-toggle gating (shown only when `renderDiff` + `getCurrent` present, hidden otherwise), inline rename (no native prompt), in-panel restore confirm + `onRestored`, role gating (reader / writer / admin), and the stale-preview race.

The new race assertions were first proven **red** against a deliberately neutered guard (generation check + abort removed), then **green** after restoring the real implementation.

```
vitest run src/versions        →  8 files, 83 passed (was 6 files pre-change)
vitest run src/versions src/board →  157 passed (no regressions)
tsc --noEmit                   →  no new errors in versions/*
```

Related: XIN-837 (this phase), XIN-836 (design).

### DIFF
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -289,6 +289,22 @@ describe('VersionHistoryPanel — mutations & permissions', () => {
     await waitFor(() => expect(onRestored).toHaveBeenCalled())
   })
 
+  it('delete goes through the same in-panel confirm box (not window.confirm)', async () => {
+    // P2 (XIN-848): delete used a native window.confirm; it now uses the unified in-panel confirm
+    // box like restore. Guard against a regression to the native dialog and confirm the row is
+    // removed only after the box's confirm is clicked.
+    const confirmSpy = vi.spyOn(window, 'confirm')
+    renderPanel('admin')
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.delete'))
+    // No native dialog; a centered in-panel confirm box appears instead.
+    expect(confirmSpy).not.toHaveBeenCalled()
+    const box = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    fireEvent.click(btnByText(box, 'docs.version.delete'))
+    await waitFor(() => expect(deleteVersionMock).toHaveBeenCalledWith('d_1', 7))
+    confirmSpy.mockRestore()
+  })
+
   it('renames a named version inline (no native prompt)', async () => {
     renderPanel('admin')
     await screen.findByText('Draft v1')
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -1,0 +1,270 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
+import { render, screen, waitFor, cleanup, fireEvent } from '@testing-library/react'
+
+// XIN-837 (拆单阶段 1): the unified <VersionHistoryPanel> shell. These tests pin the behavior the
+// three end panels will inherit: a single mixed list with filter tabs + counts + load-more, the
+// centered preview/diff modal (Esc / overlay-close / stopPropagation), inline save/rename (no native
+// prompt), the in-panel restore confirm box, role-gated affordances, and the unified race guard —
+// a slow earlier preview must never overwrite a newer selection.
+
+const NAMED = {
+  docVersionSeq: 7,
+  kind: 'named' as const,
+  label: 'Draft v1',
+  createdBy: 'u_self',
+  createdAt: '2026-06-20T10:00:00.000Z',
+  sizeBytes: 1234,
+  schemaVersion: 1,
+  restoredFrom: null,
+}
+const AUTO = {
+  docVersionSeq: 6,
+  kind: 'auto' as const,
+  label: '',
+  createdBy: 'u_self',
+  createdAt: '2026-06-20T09:30:00.000Z',
+  sizeBytes: 999,
+  schemaVersion: 1,
+  restoredFrom: null,
+}
+const COUNTS = { auto: 5, manual: 2, restore: 1, total: 8 }
+
+const listVersionsMock = vi.fn(
+  async (_docId: unknown, opts?: { kind?: string; cursor?: number | null }) => {
+    if (opts?.cursor != null) return { items: [{ ...AUTO, docVersionSeq: 4 }], nextCursor: null, counts: COUNTS }
+    return { items: [NAMED, AUTO], nextCursor: 100, counts: COUNTS }
+  },
+)
+const createNamedVersionMock = vi.fn(async () => 8)
+const restoreVersionMock = vi.fn(async () => ({ newDocVersionSeq: 9, restoredFrom: 7 }))
+const renameVersionMock = vi.fn(async () => undefined)
+const deleteVersionMock = vi.fn(async () => undefined)
+
+vi.mock('./api.ts', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('./api.ts')>()
+  return {
+    ...actual,
+    listVersions: (...a: unknown[]) => listVersionsMock(...(a as [unknown, { kind?: string }?])),
+    createNamedVersion: (...a: unknown[]) => createNamedVersionMock(...(a as [])),
+    restoreVersion: (...a: unknown[]) => restoreVersionMock(...(a as [])),
+    renameVersion: (...a: unknown[]) => renameVersionMock(...(a as [])),
+    deleteVersion: (...a: unknown[]) => deleteVersionMock(...(a as [])),
+  }
+})
+
+import { VersionHistoryPanel } from './VersionHistoryPanel.tsx'
+import type { Role } from '../auth/roles.ts'
+
+interface PreviewState {
+  body: string
+}
+
+/** A default host wiring for the shell: preview returns a tagged body; diff/current provided. */
+function renderPanel(
+  role: Role = 'admin',
+  overrides: Partial<{
+    loadPreviewState: (seq: number, signal: AbortSignal) => Promise<PreviewState>
+    renderDiff?: (s: PreviewState, c: string | null) => React.ReactNode
+    getCurrent?: () => string | null
+    onRestored?: () => void
+  }> = {},
+) {
+  const loadPreviewState =
+    overrides.loadPreviewState ?? (async (seq: number) => ({ body: `body-${seq}` }))
+  return render(
+    <VersionHistoryPanel<PreviewState, string>
+      docId="d_1"
+      role={role}
+      loadPreviewState={loadPreviewState}
+      renderPreview={(s) => <div data-testid="preview-body">{s.body}</div>}
+      renderDiff={
+        'renderDiff' in overrides
+          ? overrides.renderDiff
+          : (s, c) => <div data-testid="diff-body">{`${s.body}|${c ?? 'null'}`}</div>
+      }
+      getCurrent={'getCurrent' in overrides ? overrides.getCurrent : () => 'CURRENT'}
+      onRestored={overrides.onRestored}
+    />,
+  )
+}
+
+const btnByText = (root: ParentNode, text: string) =>
+  Array.from(root.querySelectorAll('button')).find((b) => b.textContent === text) as HTMLButtonElement
+
+beforeEach(() => {
+  listVersionsMock.mockClear()
+  createNamedVersionMock.mockClear()
+  restoreVersionMock.mockClear()
+  renameVersionMock.mockClear()
+  deleteVersionMock.mockClear()
+})
+afterEach(() => cleanup())
+
+describe('VersionHistoryPanel — list / filter / counts', () => {
+  it('renders a single mixed list and requests kind="all" on mount', async () => {
+    renderPanel()
+    await screen.findByText('Draft v1')
+    expect(document.querySelectorAll('.octo-version-list .octo-version-row').length).toBe(2)
+    expect((listVersionsMock.mock.calls[0][1] as { kind?: string }).kind).toBe('all')
+  })
+
+  it('shows the unified counts header (manual+restore · auto)', async () => {
+    renderPanel()
+    await screen.findByText('Draft v1')
+    const counts = document.querySelector('.octo-version-counts') as HTMLElement
+    // manual(2) + restore(1) = 3 · auto = 5
+    expect(counts.textContent).toContain('3')
+    expect(counts.textContent).toContain('5')
+  })
+
+  it('reloads with the chosen kind when a filter tab is clicked', async () => {
+    renderPanel()
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.body, 'docs.version.filterManual'))
+    await waitFor(() =>
+      expect(listVersionsMock.mock.calls.some((c) => (c[1] as { kind?: string })?.kind === 'manual')).toBe(true),
+    )
+  })
+
+  it('pages the current filter via cursor on load-more', async () => {
+    renderPanel()
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.body, 'docs.version.loadMore'))
+    await waitFor(() =>
+      expect(listVersionsMock.mock.calls.some((c) => (c[1] as { cursor?: number })?.cursor === 100)).toBe(true),
+    )
+  })
+})
+
+describe('VersionHistoryPanel — centered preview modal', () => {
+  it('opens a centered modal (not inline) and renders the injected preview, honoring the signal', async () => {
+    const seen: AbortSignal[] = []
+    renderPanel('admin', {
+      loadPreviewState: async (seq, signal) => {
+        seen.push(signal)
+        return { body: `body-${seq}` }
+      },
+    })
+    await screen.findByText('Draft v1')
+    expect(document.querySelector('.docs-version-preview-modal')).toBeNull()
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.preview'))
+    await waitFor(() => expect(document.querySelector('.docs-version-preview-modal')).toBeTruthy())
+    const modal = document.querySelector('.docs-version-preview-modal') as HTMLElement
+    expect(modal.closest('.octo-modal-overlay')).toBeTruthy()
+    expect(modal.getAttribute('role')).toBe('dialog')
+    await waitFor(() => expect(modal.querySelector('[data-testid="preview-body"]')?.textContent).toBe('body-7'))
+    // The host received a real AbortSignal.
+    expect(seen[0]).toBeInstanceOf(AbortSignal)
+  })
+
+  it('closes on Escape and on overlay click, but not when the dialog body is clicked', async () => {
+    renderPanel()
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.preview'))
+    await waitFor(() => expect(document.querySelector('.docs-version-preview-modal')).toBeTruthy())
+    // Body click does not close (stopPropagation).
+    fireEvent.mouseDown(document.querySelector('.docs-version-preview-modal')!)
+    expect(document.querySelector('.docs-version-preview-modal')).toBeTruthy()
+    // Escape closes.
+    fireEvent.keyDown(document, { key: 'Escape' })
+    await waitFor(() => expect(document.querySelector('.docs-version-preview-modal')).toBeNull())
+  })
+
+  it('offers the compare toggle only when renderDiff + getCurrent are provided', async () => {
+    // With diff+current: toggle appears and renders the injected diff.
+    renderPanel()
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.preview'))
+    const modal = await waitFor(() => document.querySelector('.docs-version-preview-modal') as HTMLElement)
+    await waitFor(() => expect(modal.querySelector('[data-testid="preview-body"]')).toBeTruthy())
+    const toggle = btnByText(modal, 'docs.version.compare')
+    expect(toggle).toBeTruthy()
+    fireEvent.click(toggle)
+    await waitFor(() => expect(modal.querySelector('[data-testid="diff-body"]')?.textContent).toBe('body-7|CURRENT'))
+
+    cleanup()
+    // Without diff (board case): no compare toggle.
+    renderPanel('admin', { renderDiff: undefined, getCurrent: undefined })
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.preview'))
+    const modal2 = await waitFor(() => document.querySelector('.docs-version-preview-modal') as HTMLElement)
+    await waitFor(() => expect(modal2.querySelector('[data-testid="preview-body"]')).toBeTruthy())
+    expect(btnByText(modal2, 'docs.version.compare')).toBeUndefined()
+  })
+})
+
+describe('VersionHistoryPanel — race guard', () => {
+  it('a slow earlier preview never overwrites a newer selection', async () => {
+    const deferreds = new Map<number, (v: PreviewState) => void>()
+    const signals = new Map<number, AbortSignal>()
+    renderPanel('admin', {
+      loadPreviewState: (seq, signal) => {
+        signals.set(seq, signal)
+        return new Promise<PreviewState>((resolve) => deferreds.set(seq, resolve))
+      },
+    })
+    await screen.findByText('Draft v1')
+    const rows = document.querySelectorAll('.octo-version-row')
+    // Preview #7 (slow), then #6 (fast) before #7 resolves.
+    fireEvent.click(btnByText(rows[0], 'docs.version.preview'))
+    fireEvent.click(btnByText(rows[1], 'docs.version.preview'))
+    // The newer request aborted the earlier one on the wire.
+    expect(signals.get(7)!.aborted).toBe(true)
+    // Resolve the newer one first, then the stale earlier one.
+    deferreds.get(6)!({ body: 'body-6' })
+    await waitFor(() =>
+      expect(document.querySelector('[data-testid="preview-body"]')?.textContent).toBe('body-6'),
+    )
+    deferreds.get(7)!({ body: 'body-7' })
+    // The stale #7 response is discarded — the modal still shows #6.
+    await waitFor(() => {})
+    expect(document.querySelector('[data-testid="preview-body"]')?.textContent).toBe('body-6')
+  })
+})
+
+describe('VersionHistoryPanel — mutations & permissions', () => {
+  it('restore goes through an in-panel confirm box (not window.confirm) and fires onRestored', async () => {
+    const onRestored = vi.fn()
+    renderPanel('admin', { onRestored })
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.restore'))
+    // A centered confirm box appears in-panel.
+    const confirm = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    fireEvent.click(btnByText(confirm, 'docs.version.restore'))
+    await waitFor(() => expect(restoreVersionMock).toHaveBeenCalled())
+    await waitFor(() => expect(onRestored).toHaveBeenCalled())
+  })
+
+  it('renames a named version inline (no native prompt)', async () => {
+    renderPanel('admin')
+    await screen.findByText('Draft v1')
+    const row = document.querySelector('.octo-version-row-named') as HTMLElement
+    fireEvent.click(btnByText(row, 'docs.version.rename'))
+    const input = row.querySelector('input') as HTMLInputElement
+    fireEvent.change(input, { target: { value: 'Renamed' } })
+    fireEvent.click(btnByText(row, 'docs.version.save'))
+    await waitFor(() => expect(renameVersionMock).toHaveBeenCalledWith('d_1', 7, 'Renamed'))
+  })
+
+  it('gates save/restore/delete/rename behind role (reader sees none)', async () => {
+    renderPanel('reader')
+    await screen.findByText('Draft v1')
+    expect(btnByText(document.body, 'docs.version.saveCurrent')).toBeUndefined()
+    const row = document.querySelector('.octo-version-row') as HTMLElement
+    expect(btnByText(row, 'docs.version.restore')).toBeUndefined()
+    expect(btnByText(row, 'docs.version.delete')).toBeUndefined()
+    expect(btnByText(row, 'docs.version.rename')).toBeUndefined()
+    // Preview stays available to readers.
+    expect(btnByText(row, 'docs.version.preview')).toBeTruthy()
+  })
+
+  it('writer can save + rename but cannot restore/delete', async () => {
+    renderPanel('writer')
+    await screen.findByText('Draft v1')
+    expect(btnByText(document.body, 'docs.version.saveCurrent')).toBeTruthy()
+    const row = document.querySelector('.octo-version-row-named') as HTMLElement
+    expect(btnByText(row, 'docs.version.rename')).toBeTruthy()
+    expect(btnByText(row, 'docs.version.restore')).toBeUndefined()
+    expect(btnByText(row, 'docs.version.delete')).toBeUndefined()
+  })
+})
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -29,12 +29,11 @@ const AUTO = {
 }
 const COUNTS = { auto: 5, manual: 2, restore: 1, total: 8 }
 
-const listVersionsMock = vi.fn(
-  async (_docId: unknown, opts?: { kind?: string; cursor?: number | null }) => {
-    if (opts?.cursor != null) return { items: [{ ...AUTO, docVersionSeq: 4 }], nextCursor: null, counts: COUNTS }
-    return { items: [NAMED, AUTO], nextCursor: 100, counts: COUNTS }
-  },
-)
+const defaultListImpl = async (_docId: unknown, opts?: { kind?: string; cursor?: number | null }) => {
+  if (opts?.cursor != null) return { items: [{ ...AUTO, docVersionSeq: 4 }], nextCursor: null, counts: COUNTS }
+  return { items: [NAMED, AUTO], nextCursor: 100, counts: COUNTS }
+}
+const listVersionsMock = vi.fn(defaultListImpl)
 const createNamedVersionMock = vi.fn(async () => 8)
 const restoreVersionMock = vi.fn(async () => ({ newDocVersionSeq: 9, restoredFrom: 7 }))
 const renameVersionMock = vi.fn(async () => undefined)
@@ -92,7 +91,8 @@ const btnByText = (root: ParentNode, text: string) =>
   Array.from(root.querySelectorAll('button')).find((b) => b.textContent === text) as HTMLButtonElement
 
 beforeEach(() => {
-  listVersionsMock.mockClear()
+  listVersionsMock.mockReset()
+  listVersionsMock.mockImplementation(defaultListImpl)
   createNamedVersionMock.mockClear()
   restoreVersionMock.mockClear()
   renameVersionMock.mockClear()
@@ -126,13 +126,18 @@ describe('VersionHistoryPanel — list / filter / counts', () => {
     )
   })
 
-  it('pages the current filter via cursor on load-more', async () => {
+  it('pages the current filter via cursor on load-more and appends the page (no stale error)', async () => {
     renderPanel()
     await screen.findByText('Draft v1')
+    expect(document.querySelectorAll('.octo-version-row').length).toBe(2)
     fireEvent.click(btnByText(document.body, 'docs.version.loadMore'))
     await waitFor(() =>
       expect(listVersionsMock.mock.calls.some((c) => (c[1] as { cursor?: number })?.cursor === 100)).toBe(true),
     )
+    // The fetched page (seq 4) is appended onto the existing rows — the append path the load-more
+    // guard protects. isCurrent() gates both append and errorMore, so no stale error leaks.
+    await waitFor(() => expect(document.querySelectorAll('.octo-version-row').length).toBe(3))
+    expect(document.body.textContent).not.toContain('docs.version.errorMore')
   })
 })
 
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -139,6 +139,55 @@ describe('VersionHistoryPanel — list / filter / counts', () => {
     await waitFor(() => expect(document.querySelectorAll('.octo-version-row').length).toBe(3))
     expect(document.body.textContent).not.toContain('docs.version.errorMore')
   })
+
+  it('re-enables Load More after a superseded load-more (no stuck loadingMore deadlock)', async () => {
+    // Regression for the P1 flagged on #656: setLoadingMore(false) used to be gated on isCurrent(),
+    // so a load-more superseded (by a filter switch / refresh / restore) before its finally ran
+    // would skip the reset and wedge loadingMore=true forever — permanently disabling the Load More
+    // button. Here we hang the load-more page, supersede it with a primary refresh (docId change →
+    // begin()), then release the stale page: its result must be dropped, but loadingMore MUST clear
+    // regardless of isCurrent(), so the fresh list's Load More button is enabled, not deadlocked.
+    let releaseLoadMore: (() => void) | null = null
+    listVersionsMock.mockImplementation(
+      async (_docId: unknown, opts?: { kind?: string; cursor?: number | null }) => {
+        if (opts?.cursor != null) {
+          await new Promise<void>((resolve) => {
+            releaseLoadMore = resolve as () => void
+          })
+          return { items: [{ ...AUTO, docVersionSeq: 4 }], nextCursor: 200, counts: COUNTS }
+        }
+        return { items: [NAMED, AUTO], nextCursor: 100, counts: COUNTS }
+      },
+    )
+    const Panel = (p: { docId: string }) => (
+      <VersionHistoryPanel<PreviewState, string>
+        docId={p.docId}
+        role="admin"
+        loadPreviewState={async (seq: number) => ({ body: `body-${seq}` })}
+        renderPreview={(s) => <div data-testid="preview-body">{s.body}</div>}
+      />
+    )
+    const { rerender } = render(<Panel docId="d_1" />)
+    await screen.findByText('Draft v1')
+    // Kick off load-more; its page request now hangs, so loadingMore stays true and the button
+    // is disabled.
+    fireEvent.click(btnByText(document.body, 'docs.version.loadMore'))
+    await waitFor(() => expect(btnByText(document.body, 'docs.version.loadMore').disabled).toBe(true))
+    // Supersede the in-flight follow-up with a primary refresh (docId change → begin()), which
+    // makes the hanging load-more's isCurrent() report false.
+    rerender(<Panel docId="d_2" />)
+    await waitFor(() =>
+      expect(listVersionsMock.mock.calls.some((c) => c[0] === 'd_2')).toBe(true),
+    )
+    // Release the now-stale load-more: its rows are dropped (isCurrent() false), but the finally
+    // must still clear loadingMore. Before the fix this reset was gated out and the flag stuck.
+    ;(releaseLoadMore as (() => void) | null)?.()
+    await waitFor(() => {
+      const btn = btnByText(document.body, 'docs.version.loadMore')
+      expect(btn).toBeTruthy()
+      expect(btn.disabled).toBe(false)
+    })
+  })
 })
 
 describe('VersionHistoryPanel — centered preview modal', () => {
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -1,3 +1,4 @@
+import { StrictMode } from 'react'
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 import { render, screen, waitFor, cleanup, fireEvent } from '@testing-library/react'
 
@@ -336,5 +337,35 @@ describe('VersionHistoryPanel — mutations & permissions', () => {
     expect(btnByText(row, 'docs.version.rename')).toBeTruthy()
     expect(btnByText(row, 'docs.version.restore')).toBeUndefined()
     expect(btnByText(row, 'docs.version.delete')).toBeUndefined()
+  })
+})
+
+describe('VersionHistoryPanel — StrictMode liveness (mounted-ref re-arm)', () => {
+  it('finishes a mutation after a StrictMode double-mount (mounted ref stays live)', async () => {
+    // React 18 StrictMode (dev) runs an effect mount → cleanup → remount on the same node. The
+    // abort/cleanup effect flips the `mounted` liveness ref to false on cleanup; unless the effect
+    // body re-arms it to true on every (re)mount, the ref stays false for the real mount's whole
+    // lifetime and every mutation handler early-returns after its await — the trailing state reset
+    // (busy=false / confirm cleared) never runs and the panel freezes with buttons stuck busy.
+    // This pins that the delete confirm box closes once the delete resolves, which only happens
+    // while the liveness ref is true through the double-mount.
+    render(
+      <StrictMode>
+        <VersionHistoryPanel<PreviewState, string>
+          docId="d_1"
+          role="admin"
+          loadPreviewState={async (seq: number) => ({ body: `body-${seq}` })}
+          renderPreview={(s) => <div data-testid="preview-body">{s.body}</div>}
+        />
+      </StrictMode>,
+    )
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.delete'))
+    const box = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    fireEvent.click(btnByText(box, 'docs.version.delete'))
+    await waitFor(() => expect(deleteVersionMock).toHaveBeenCalledWith('d_1', 7))
+    // Re-armed: onDelete runs past its `if (!mounted.current) return` gate → setConfirmDelete(null)
+    // closes the box. Gated (no re-arm): ref stuck false → early return → the box never closes.
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
   })
 })
--- a/packages/docs/src/versions/VersionHistoryPanel.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.tsx
@@ -206,8 +206,11 @@ export function VersionHistoryPanel<TState, TCurrent>({
     setError(null)
     try {
       const res = await listVersions(docId, { kind: filter, cursor: nextCursor, limit: pageSize, signal })
-      // Bound to the list generation: if a refresh / filter switch / restore replaced the list
-      // before this page landed, drop it rather than append onto a list that no longer exists.
+      // Bound to the list guard: if a refresh / filter switch / restore (begin) OR a newer
+      // load-more (beginFollowUp) superseded this page before it landed, isCurrent() is false and
+      // we drop it — never appending stale/duplicate rows onto a list that moved on. The follow-up
+      // token in createRaceGuard is what makes the "newer load-more" case report non-current even
+      // though the aborted request may have resolved a hair before its abort.
       if (!isCurrent()) return
       setItems((cur) => [...cur, ...res.items])
       setNextCursor(res.nextCursor)
--- a/packages/docs/src/versions/VersionHistoryPanel.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.tsx
@@ -198,6 +198,7 @@ export function VersionHistoryPanel<TState, TCurrent>({
 
   // Abort any in-flight request when the panel unmounts.
   useEffect(() => {
+    mounted.current = true // re-arm on (re)mount so StrictMode's mount→cleanup→remount stays live
     const list = listGuard.current
     const preview = previewGuard.current
     return () => {
--- a/packages/docs/src/versions/VersionHistoryPanel.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.tsx
@@ -219,7 +219,14 @@ export function VersionHistoryPanel<TState, TCurrent>({
       if (!isCurrent()) return
       setError(t('docs.version.errorMore'))
     } finally {
-      if (isCurrent()) setLoadingMore(false)
+      // Always clear the loading flag on this follow-up's own completion, independent of
+      // isCurrent(). The guard's job is to discard the stale *result* (handled by the early
+      // returns above); the *loading flag* must not be gated on isCurrent(), or a superseded
+      // load-more (filter switch / refresh / restore while a page is in flight) would skip this
+      // reset and wedge loadingMore=true forever, permanently disabling the Load More button.
+      // A superseded-but-mounted setState is harmless; a genuine unmount is covered by the guard
+      // util's abort; a newer load-more re-sets the flag true itself.
+      setLoadingMore(false)
     }
   }
 
--- a/packages/docs/src/versions/VersionHistoryPanel.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.tsx
@@ -139,6 +139,9 @@ export function VersionHistoryPanel<TState, TCurrent>({
 
   // Restore is confirmed in a centered in-panel box (doc model), not a native window.confirm.
   const [confirmRestore, setConfirmRestore] = useState<VersionMeta | null>(null)
+  // Delete is likewise confirmed in the same in-panel box (was a native window.confirm) so the
+  // destructive-action UX is unified across doc/sheet/board rather than falling back to the browser.
+  const [confirmDelete, setConfirmDelete] = useState<VersionMeta | null>(null)
 
   // Centered preview / diff modal.
   const [selected, setSelected] = useState<VersionMeta | null>(null)
@@ -151,6 +154,10 @@ export function VersionHistoryPanel<TState, TCurrent>({
   // preview — both from the shared createRaceGuard so all three chains abort + last-write-win.
   const listGuard = useRef(createRaceGuard())
   const previewGuard = useRef(createRaceGuard())
+  // Liveness flag: mutation handlers setState after an await, and refresh()/the guards only cover
+  // their own async chains — this guards the trailing setBusy/setNotice/onRestored in the handlers
+  // so they no-op if the panel unmounted mid-flight (no setState-after-unmount).
+  const mounted = useRef(true)
 
   const mySnapshot = canSnapshot(role)
   const myRestore = canRestoreVersion(role)
@@ -194,6 +201,7 @@ export function VersionHistoryPanel<TState, TCurrent>({
     const list = listGuard.current
     const preview = previewGuard.current
     return () => {
+      mounted.current = false
       list.abort()
       preview.abort()
     }
@@ -244,13 +252,16 @@ export function VersionHistoryPanel<TState, TCurrent>({
     try {
       await createNamedVersion(docId, snapshotLabel.trim() || undefined)
     } catch {
+      if (!mounted.current) return
       setError(t('docs.version.errorSave'))
       setBusy(false)
       return
     }
+    if (!mounted.current) return
     setSnapshotOpen(false)
     setSnapshotLabel('')
     const ok = await refresh({ soft: true })
+    if (!mounted.current) return
     if (!ok) setNotice(t('docs.version.staleNotice'))
     setBusy(false)
   }
@@ -273,33 +284,40 @@ export function VersionHistoryPanel<TState, TCurrent>({
     try {
       await renameVersion(docId, seq, label)
     } catch {
+      if (!mounted.current) return
       setError(t('docs.version.errorRename'))
       setBusy(false)
       return
     }
+    if (!mounted.current) return
     // Optimistic label update, then reconcile with server ordering/counts (soft: rename landed).
     setItems((cur) => cur.map((v) => (v.docVersionSeq === seq ? { ...v, label } : v)))
     cancelRename()
     const ok = await refresh({ soft: true })
+    if (!mounted.current) return
     if (!ok) setNotice(t('docs.version.staleNotice'))
     setBusy(false)
   }
 
   const onDelete = async (v: VersionMeta) => {
-    if (!window.confirm(t('docs.version.deleteConfirm', { values: { seq: v.docVersionSeq } }))) return
     setBusy(true)
     setError(null)
     try {
       await deleteVersion(docId, v.docVersionSeq)
     } catch {
+      if (!mounted.current) return
       setError(t('docs.version.errorDelete'))
+      setConfirmDelete(null)
       setBusy(false)
       return
     }
+    if (!mounted.current) return
+    setConfirmDelete(null)
     if (selected?.docVersionSeq === v.docVersionSeq) closePreview()
     if (renamingSeq === v.docVersionSeq) cancelRename()
     setItems((cur) => cur.filter((x) => x.docVersionSeq !== v.docVersionSeq))
     const ok = await refresh({ soft: true })
+    if (!mounted.current) return
     if (!ok) setNotice(t('docs.version.staleNotice'))
     setBusy(false)
   }
@@ -312,15 +330,18 @@ export function VersionHistoryPanel<TState, TCurrent>({
     try {
       res = await restoreVersion(docId, v.docVersionSeq)
     } catch (e) {
+      if (!mounted.current) return
       setError(t(restoreErrorKey ? restoreErrorKey(e) : 'docs.version.errorRestore'))
       setConfirmRestore(null)
       setBusy(false)
       return
     }
+    if (!mounted.current) return
     // Restore landed — the live surface reconciles via Yjs. A follow-up refresh miss is soft.
     setConfirmRestore(null)
     closePreview()
     const ok = await refresh({ soft: true })
+    if (!mounted.current) return
     setNotice(
       ok
         ? t('docs.version.restoredNotice', { values: { from: res.restoredFrom, seq: res.newDocVersionSeq } })
@@ -458,7 +479,7 @@ export function VersionHistoryPanel<TState, TCurrent>({
                     type="button"
                     className="octo-tb-btn"
                     disabled={busy}
-                    onClick={() => void onDelete(v)}
+                    onClick={() => setConfirmDelete(v)}
                   >
                     {t('docs.version.delete')}
                   </button>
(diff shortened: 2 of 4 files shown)


OctoBoooot (dismissed):
## Review: unified VersionHistoryPanel shell + race guard util (#656)

**Verdict:** Approve
Exemplary shared-primitive PR. The `createRaceGuard` util is a correct, well-documented consolidation of the AbortController+generation pattern, and the shell follows its contract precisely. Nothing is wired to it yet (no user-visible change), so adoption is incremental and independently revertible. Verified the util, the shell's guard usage, restore-gating, and XSS.

Risk tier: normal; size/XL → `needs-human-review` applied (human sign-off before merge). Base `main`; `mergeStateStatus=BLOCKED` (review required).

### What I verified
- **`raceGuard.ts` is correct.** `begin()` aborts prior primary + follow-up, nulls follow-up, `++generation`, returns a ticket bound to the new gen (truly cancels in-flight work AND supersedes older tickets). `beginFollowUp()` aborts only a prior follow-up and binds to the *current* generation (no bump) — so a later `begin()` (filter switch / restore / delete) makes the in-flight page's `isCurrent()` false and its resolved result is discarded rather than appended to a replaced list (the subtle "page resolves after list replaced" bug — handled). `abort()` aborts both + bumps generation so outstanding tickets flip stale on unmount. `isCurrent` closes over live `generation` — correct closure. Strict superset of both prior guards (board abort+gen, doc token-only).
- **Shell follows the contract exactly.** Both guards in `useRef`; list refresh / load-more / preview each `begin()`/`beginFollowUp()` then check **`isCurrent()` after every await — in both the success and the catch branch** (preview explicitly swallows the stale error), and `useEffect` cleanup calls `abort()` on both. This is precisely the "isCurrent after every await, both branches, abort on unmount" discipline the util documents.
- **Restore is confirm-gated** via a centered in-panel confirm box (`confirmRestore` state; not native `window.confirm`), errors mapped through `restoreErrorKey`. Delete confirmed with the seq.
- **XSS-safe:** no `dangerouslySetInnerHTML`/`innerHTML` in the shell; labels/badges/notes render as escaped JSX text / i18n `{values}`. `renderPreview`/`renderDiff` are end-injected render props (each end's responsibility), not sunk unsafely.
- **No production risk yet:** pure additions; doc/sheet/board panels untouched and unchanged — a latent issue can't surface until an adapter phase wires it.
- **Tests:** `raceGuard.test.ts` (abort-on-supersede + last-write-wins both lanes + abort cleanup) and the panel suite (filter/counts/load-more, preview modal Escape/overlay/stopPropagation, compare gating, rename, restore-confirm + onRestored, role gating, the stale-preview race) — and the race assertions were proven red against a neutered guard first, which is the right way to confirm a race test actually tests the race.

### Praise
- This is how to introduce a shared primitive: the race guard consolidates two divergent existing guards into one that is a *strict superset* (aborts on the wire AND last-write-wins, primary + follow-up lanes), documents the exact call contract on the ticket, and the shell demonstrates that contract flawlessly (isCurrent in both resolve and catch, abort on unmount). Landing it unwired — with the three ends untouched so each adapter phase reverts independently — is the low-risk rollout the correctness of a race primitive deserves. And the tests earn trust by being red-against-a-neutered-guard first.

Jerry-Xin (dismissed):
## Review: unified VersionHistoryPanel shell + race guard (phase 1/4)

Reviewed line-by-line against the XIN-836 design and the phase-1 acceptance criteria. This is a clean, purely-additive shell — approving.

### Constraint compliance (shell only, no end wired)
- `gh pr diff --name-only` = exactly 4 new files, **deletions = 0**. `VersionPanel.tsx`, sheet/board panels, `api.ts` and `format.ts` are all untouched, so there is genuinely no user-visible change this phase.

### Props contract vs XIN-836
Matches the spec exactly: `docId` / `role` / `names?` / `onClose?` / `pageSize?` (defaults to 30) / `defaultFilter?` (defaults to `all`) / `onRestored?` / `previewErrorKey?` / `restoreErrorKey?` / `loadPreviewState(seq, signal)` / `renderPreview(state)` / `renderDiff?(state, current)` / `getCurrent?()`. `canCompare = !!(renderDiff && getCurrent)` correctly hides the compare entry when `renderDiff` is omitted.

### Shell behavior
Single reverse-chronological mixed list, `all/manual/auto` filter tabs, counts header rendered as `manual+restore · auto` (matches the backend semantics documented on `VersionCounts`), cursor load-more, inline save/rename compose (no native prompt), in-panel centered restore confirm box, restore via `restoreVersion` (reuses `api.ts`, no new endpoint), centered modal with overlay-close + `stopPropagation` + Escape + `aria-modal` + `role="dialog"`. Permissions gate on `canSnapshot` / `canRestoreVersion`.

### Race guard
`createRaceGuard()` implements AbortController + monotonic generation cleanly. `begin()` aborts prior primary + follow-up and bumps the generation; `beginFollowUp()` binds to the current generation and is superseded by a later `begin()`; `abort()` bumps generation so outstanding tickets go stale on unmount. Both guarantees (true wire-abort + last-write-wins) are asserted directly in the tests. The list uses one guard (refresh=primary, load-more=follow-up) and preview uses an independent one — correct.

### Tests / verification (ran locally)
- `vitest run src/versions` -> **8 files, 83 tests, all green.** Covers filter switch, counts, pagination, reader/writer/admin gating, the slow-#7-vs-fast-#6 preview race, modal Esc/overlay/stopPropagation, and compare-toggle hidden when `renderDiff` absent.
- `tsc --noEmit` -> the two new files are clean. (The 2 repo-wide errors are in `dmworkbase/i18n/format.ts` and `members/sort.ts` — pre-existing and unrelated; a deletions=0 additive PR cannot have introduced them.)

> Note on the stated "baseline 100": on this branch `src/versions` collects 83 tests, all passing — the substantive gate (all green, new files fully covered) holds; the 100 figure looks like an estimate rather than the actual count.

### Non-blocking notes (fine to address in the adapter phase)
1. `onDelete` still uses native `window.confirm` (line 279) while save/rename/restore deliberately avoid native dialogs, and delete has no test. The spec only mandated the confirm-box for restore, so this isn't blocking — but aligning delete with the in-panel pattern (and adding a test) would make the shell fully consistent.
2. The new i18n keys (`filterAll/Manual/Auto`, `countManual/Auto`, `staleNotice`) are referenced but intentionally not defined this phase. Safe while unmounted; just make sure they land before the first adapter renders the shell.

**Verdict: APPROVE.** Contract-accurate, well-guarded against races, comprehensively tested, and strictly additive.

OctoBoooot (dismissed):
## Review: unified VersionHistoryPanel shell + race guard util (#656) — delta re-review

**Verdict:** Approve (re-affirmed)
The head moved since my prior APPROVE, but the only change is a 6-line comment. `raceGuard.ts` and both test files are blob-identical; `VersionHistoryPanel.tsx` differs by exactly one added comment block documenting that the new i18n keys are deferred to the adapter phase. Approval carries forward unchanged.

Risk tier: normal; size/XL → `needs-human-review` stands. Base `main`; `mergeStateStatus=BLOCKED`.

### What changed since `596720ee` (verified)
- Blob-compared all 4 files: `raceGuard.ts` IDENTICAL, `raceGuard.test.ts` IDENTICAL, `VersionHistoryPanel.test.tsx` IDENTICAL, `VersionHistoryPanel.tsx` +6 — a comment noting the `filterAll/filterManual/filterAuto`/`countManual/countAuto`/`staleNotice` keys are intentionally **not** added this phase (shell unmounted → no user sees a raw key; keys land with the doc adapter). No logic change.

### Standing (from prior review, unchanged)
- `raceGuard.ts` correct (begin bumps+aborts / beginFollowUp binds current gen / abort bumps on unmount; strict superset of both prior guards). Shell follows the contract exactly (isCurrent after every await in both branches, abort on unmount). Restore confirm-gated; XSS-safe (no dangerouslySetInnerHTML). Unwired — no user-visible change, incremental adopt/revert.

### Note
The deferred-i18n-keys point (also raised by Jerry-Xin) is non-blocking: the shell isn't mounted anywhere, so no raw key is ever rendered; the adapter phase must add them before it mounts the shell — the added comment now documents exactly that. Pure-comment re-affirmation; nothing else to re-verify.

Jerry-Xin (changes requested):
The PR is in scope for `octo-web`, but the new race guard has a follow-up lane correctness bug that undermines the utility this PR is adding.

**🔴 Blocking**
- 🔴 Critical: `beginFollowUp()` aborts the previous follow-up request but does not make that previous ticket stale. In [raceGuard.ts](/Users/mlamp/.openclaw/tmp/pr-review-hi55rzhh/repo/packages/docs/src/versions/raceGuard.ts:59), `isCurrent()` only checks `gen === generation`; because follow-up requests intentionally share the same generation, this sequence leaves the older aborted ticket reporting current:
  `guard.begin(); const a = guard.beginFollowUp(); const b = guard.beginFollowUp(); a.signal.aborted === true; a.isCurrent() === true`.
  `VersionHistoryPanel` then trusts `isCurrent()` before appending page results and before clearing/loading errors in [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-hi55rzhh/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:204). If two load-more requests are started before React’s `loadingMore` state disables the button, an aborted older request can still append stale/duplicate rows or surface `errorMore` while the newer request is in flight. The ticket should also become non-current when its own signal is aborted, or the follow-up lane needs its own monotonic token. Please add a regression test asserting the superseded follow-up ticket’s `isCurrent()` is false.

**💬 Non-blocking**
- 🟡 Warning: The default restore error path in [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-hi55rzhh/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:305) collapses schema restore failures to `docs.version.errorRestore`, while the existing doc panel maps schema failures to `docs.version.errorRestoreIncompatible`. If doc/sheet adapters rely on the default, users will lose the more specific restore error unless every adapter supplies `restoreErrorKey`.

**✅ Highlights**
- The project relevance gate passes: the new files extend the existing `packages/docs/src/versions` feature area.
- The shell keeps permission affordances aligned with `canSnapshot` and `canRestoreVersion`.
- The preview path correctly aborts and discards superseded primary requests.

I could not run the Vitest target locally because dependencies are not installed in this checkout (`pnpm exec vitest` reports `Command "vitest" not found`).

ploy-elison:
Addresses the CHANGES_REQUESTED (🔴 critical race-guard follow-up bug).

**Fix** (`raceGuard.ts`): the follow-up (load-more) lane shared the primary generation and `isCurrent()` only compared that generation, so a superseded follow-up ticket still reported itself current even after its request was aborted. It now has its own monotonic token — a follow-up ticket is current only while both the primary generation is unchanged **and** it is still the latest follow-up. A newer load-more (or a newer `begin()`) now makes the prior ticket non-current, not merely aborted.

Repro from the review now behaves correctly:
```
guard.begin(); const a = guard.beginFollowUp(); const b = guard.beginFollowUp();
a.signal.aborted === true   // unchanged
a.isCurrent()    === false  // was true (the bug) → now false
b.isCurrent()    === true
```

**Panel (`VersionHistoryPanel.tsx:204`)**: `onLoadMore` already gates the append *and* `errorMore` *and* the loading reset on `isCurrent()`, so no panel logic change was needed — the guard fix makes that trust sound. Comment sharpened to say so.

**Tests**: added the regression assertion (`isCurrent() === false` for a superseded follow-up) and strengthened the panel load-more test to assert the page is appended with no stale error.

```
vitest run src/versions        →  8 files, 83 passed
vitest run src/versions src/board →  33 files, 275 passed (no regressions)
```

🟡 Non-blocking item (default restore error collapsing schema failures into `errorRestore`) is intentionally deferred to the doc/sheet adapter phase, which will pass `restoreErrorKey` to preserve `errorRestoreIncompatible` — noted for that phase.

Jerry-Xin (dismissed):
**Delta re-review — commit 06179dbf** (`fix(versions): make a superseded load-more ticket stale in the race guard`)

The XIN-838 Critical (`raceGuard.ts` `beginFollowUp()` aborted the prior request but left its ticket reporting `isCurrent() === true`, so a load-more that resolved a hair before its abort could still pass the gate and append stale/duplicate rows — and falsely surface `errorMore`) is genuinely fixed. Verified against the PR head, not the commit message:

**① Independent follow-up token — fixed.** `createRaceGuard` now holds a separate `followToken` alongside `generation`. `beginFollowUp()` returns `followTicket(generation, ++followToken, controller)` whose `isCurrent()` is `gen === generation && token === followToken`. Two rapid follow-ups share `generation` but get distinct tokens, so the superseded one goes non-current — the exact hole is closed.

**② begin()/abort() bump the token — fixed.** `begin()` does `followToken++` (a filter switch / restore / delete invalidates any outstanding load-more). `abort()` bumps both `generation` and `followToken` (unmount cleanup flips every ticket). The primary lane still keys off `generation` only, so a load-more never invalidates an in-flight refresh/preview — correct asymmetry.

**③ Regression test present and non-vacuous.** `raceGuard.test.ts` asserts `firstPage.isCurrent() === false` after two `beginFollowUp()` calls (and `secondPage` stays `true`). I mutation-checked it: reverting `isCurrent` to `gen === generation` alone flips that assertion red (`expected false, got true`) — the test genuinely pins the bug, not a tautology. The panel load-more test now also asserts the fetched page is appended (2 → 3 rows) with no `errorMore` leak.

**Constraints held.** PR diffstat: 4 files, +1080 / **−0**, all net-new under `packages/docs/src/versions/` — no doc/sheet/board panel touched. `vitest run src/versions` → **83 passed (8 files)**.

**Non-blocking (deferred by design, confirmed documented).** The shell's default restore-error path collapses to `docs.version.errorRestore`; richer classification is injected per-end via the `restoreErrorKey?` prop (documented at the prop declaration). Correctly left to the adapter phases — nothing to change here.

Critical cleared, constraints intact, tests green. **APPROVE.**

OctoBoooot (approved):
## Review: unified VersionHistoryPanel shell + race guard util (#656) — delta re-review (follow-up-lane gap fixed)

The new commit fixes the follow-up-lane staleness gap (the finding Jerry-Xin graded 🔴 and I adopted as a should-fix minor after conceding my earlier "fully correct" praise missed it). The fix is exactly right — a separate follow-up token now enforces last-write-wins in the follow-up lane, not just best-effort abort. Approval re-affirmed.

Risk tier: normal; size/XL → `needs-human-review` stands. Base `main`; `mergeStateStatus=CLEAN`.

### Resolved — follow-up-vs-follow-up staleness
`raceGuard.ts` adds a separate `followToken`:
- `beginFollowUp()` now `const token = ++followToken` and returns a ticket whose `isCurrent() = gen === generation && token === followToken`. So a second `beginFollowUp()` makes the first's `isCurrent()` **false** via the token — independent of the abort race — closing the exact hole (a page that resolved a hair before its abort would previously still pass `isCurrent()` and append stale/duplicate rows).
- `begin()` also bumps `followToken` (a new primary invalidates outstanding follow-ups via both generation and token); `abort()` bumps both. Primary tickets check `generation` only, so a load-more never invalidates an in-flight refresh/preview — correct lane separation preserved.
- **Test pins the regression:** `firstPage.isCurrent()` is now asserted `false` after a second `beginFollowUp()` (with `secondPage.isCurrent()` true), crediting the CHANGES_REQUESTED finding. This is the follow-up-vs-follow-up assertion that was missing — it directly proves the fix.

### Standing (unchanged, still correct)
Primary-lane abort+generation semantics; shell contract usage (isCurrent after every await in both branches, abort on unmount); restore confirm-gated; XSS-safe; unwired shell (no production impact).

### Note (accountability)
This closes the gap my first #656 pass over-praised as "fully correct" — I'd traced primary-vs-follow-up but not follow-up-vs-follow-up. Jerry-Xin caught it; the util now enforces the last-write-wins guarantee its own docs promise, across both lanes. The doc comment is updated to describe both guarantees accurately. Verdict back to a clean APPROVE.

yujiawei (approved):
# Code Review — PR #656 (octo-web)

Reviewed at head SHA `596720ee820aedc9f2d894f064b66d66b226ab3f`. Verified locally: 21/21 new tests pass, all 83 `src/versions` tests pass, typecheck clean on the two new files, i18n locale-parity check green. This PR was auto-classified security-sensitive; see the note at the end.

## 1. Spec compliance

**Spec: ✅**

The diff does exactly what the description claims and nothing more:

- **Not wired (no user-visible change):** confirmed — `VersionHistoryPanel` and `raceGuard` have zero importers outside their own tests; the doc/sheet/board panels are untouched. `grep` for `VersionHistoryPanel`/`raceGuard` across `packages/` returns only the new files.
- **No new endpoint:** list/create/rename/delete/restore all route through the existing `versions/api.ts`. Preview is injected via `loadPreviewState`. No new REST surface added.
- **Additive i18n:** `filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice` added to both `en-US` and `zh-CN` under the `version.*` group; resolved via the registered `docs` namespace (`t('docs.version.*')`), so keys line up. Parity guard stays green.
- **Race guard is a strict superset of the old preview guard:** `createRaceGuard` adds real `AbortController` cancellation on top of the monotonic-generation token that `previewGuard.ts` had; it mirrors the pattern already shipping in `BoardVersionPanel.tsx`.

No under-build, no over-build, no scope drift.

## 2. Code quality

**Quality: Approved**

The guard logic is correct and well-tested (primary vs follow-up lanes, abort-on-supersede, last-write-wins, unmount cleanup). The preview modal state machine (`idle → loading → ready/error`), Escape/overlay-close, compare-toggle gating, and role gating are all sound and covered by tests. No P0/P1 issues found. The items below are P2 polish and edge-case hardening — none blocks merge. Nothing is wired yet, so none has current user impact; they are worth addressing before the adapter phases land.

### P2 — stuck `loadingMore` if a mutation fires during load-more
`VersionHistoryPanel.tsx` `onLoadMore` (~L196-215): row mutation buttons are gated by `busy` only, not `loadingMore`. If the user triggers restore/delete/rename/save while a load-more request is in flight, the mutation's `refresh()` calls `listGuard.begin()`, which aborts the follow-up and bumps the generation. The follow-up then hits `if (!isCurrent()) return` in both `catch` and `finally`, so `setLoadingMore(false)` never runs. `loadingMore` stays `true`, permanently disabling the "Load more" button and all filter tabs until the panel is reopened. This mirrors the existing `BoardVersionPanel` pattern (so it is not a new behavioral regression), and it self-heals on reopen, but the new shell is the right place to fix it — e.g. reset the flag unconditionally in `finally`, or key the reset off the ticket rather than `isCurrent()`.

### P2 — `commitRename` missing `if (busy) return` guard
`VersionHistoryPanel.tsx` `commitRename` (~L217): unlike `onCreateSnapshot`, it has no `busy` short-circuit. The Save button is guarded (`disabled={busy || ...}`), but the input's Enter-key path is not, so rapid Enter can fire concurrent `renameVersion` calls. Impact is bounded — the PATCH is idempotent and `cancelRename()` unmounts the input after the first commit resolves — so this is a redundant request, not corruption. Add the guard for symmetry with the other mutations.

### P2 — load-more appends without de-duplication
`VersionHistoryPanel.tsx` `onLoadMore` (~L206) appends `res.items` and uses `v.docVersionSeq` as the React key (renderRow). Cursor pagination is by monotonic `docVersionSeq`, so overlaps are unlikely in practice (and the board panel ships the same way), but a defensive de-dupe on append would remove any chance of duplicate-key warnings if the server ever returns an overlapping row.

### P2 — soft-refresh conflates "superseded" with "failed"
`refresh({ soft: true })` returns `false` both when the request genuinely fails and when it is superseded (`!isCurrent()`). Callers then show `staleNotice`. In this component the concrete "switch filter mid-refresh" path is gated (filter tabs are `disabled={loading || loadingMore}` and `refresh` sets `loading` synchronously), so it is hard to hit today, but the conflation is a latent trap for the adapter phases. Consider distinguishing superseded from failed.

### P2 — modal body uses blanket `onMouseDown={e => e.stopPropagation()}`
`VersionHistoryPanel.tsx` (~L566): standard modal pattern, but blanket `stopPropagation` on the dialog can interfere with `document`-level "click-outside" listeners inside injected `renderPreview` content. An alternative that avoids the side effect is to drop `stopPropagation` and instead close only when `e.target === e.currentTarget` on the overlay. Worth keeping in mind when ends inject richer preview surfaces.

### P2 — minor UX consistency nits
- The save-current snapshot `<input>` has no `onKeyDown`, so Enter does not submit, unlike the rename input which does.
- The rename input's `Escape` handler does not `stopPropagation`; if a rename is active while the preview modal is open, one Escape both cancels the rename and closes the modal.

## 3. Overall verdict

**APPROVE.** Spec ✅ and no P0/P1 quality issues. The findings above are non-blocking P2 polish; recommend folding the stuck-`loadingMore` fix and the `commitRename` guard in before or alongside the first adapter phase that actually wires this shell to an end.

## Note for a human reviewer (security-sensitive label)
The security-sensitive classification here is keyword-triggered. The actual diff contains no auth/token/credential/middleware logic: role gating is UI-only (`canSnapshot`/`canRestoreVersion`, with restore+delete admin-only), and every mutation still goes through the existing backend-gated `api.ts` endpoints, which the code comments note re-gate server-side. No new trust boundary is introduced by this PR.

yujiawei (approved):
# Code Review — PR #656 (octo-web) — delta re-review (head 06179dbf)

Re-review against the current head `06179dbf634e1b64dbeefc7608c44ea42f0d0af0`. My earlier APPROVE was against `596720ee`; a new commit landed after it, so this re-verifies the whole PR at the new head rather than inheriting the old verdict. Verified locally at this head: `vitest run src/versions src/i18n` → **86 passed (9 files)**, typecheck clean on the two new files, i18n locale-parity + baseline checks green.

## Delta since 596720ee (2 changes)

**1. Race guard follow-up-lane fix (`raceGuard.ts`) — resolves my prior P2.** `createRaceGuard` now carries a separate `followToken` in addition to `generation`. `beginFollowUp()` returns a ticket whose `isCurrent() = gen === generation && token === followToken`, and `begin()`/`abort()` bump `followToken` too. This closes the real hole: two rapid load-mores share the primary generation, so the superseded first ticket could previously pass `isCurrent()` if its request resolved a hair before its abort — and append stale/duplicate rows. I mutation-checked the new regression test: reverting `isCurrent` to `gen === generation` alone flips `raceGuard.test.ts`'s `firstPage.isCurrent() === false` assertion red, so the test genuinely pins the bug. The primary lane still keys off `generation` only, so a load-more never invalidates an in-flight refresh/preview — correct lane asymmetry preserved. This directly addresses the load-more/dedup concern from my first pass.

**2. i18n keys removed (`en-US.json` / `zh-CN.json`).** The six keys added in the first commit (`filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice`) were removed from the `version.*` group; they remain only under `board.version.*`. The shell still calls `t('docs.version.filterAll')` etc. This is intentional and documented in a new header comment: the acceptance gate scopes this phase to "new files + tests only," and since the shell is mounted nowhere, `t()` echoing a raw key is never user-visible. **Spec-wise this is fine** — no scope drift, no user impact. I confirmed the tests deliberately assert on the literal key strings as button text, so they stay green. The keys will need to be added in the doc-adapter phase before the shell is first rendered; that is the correct place for them.

## 1. Spec compliance

**Spec: ✅** — unchanged from the prior pass and still holds: shell + guard are not wired anywhere (zero importers outside tests), no new REST endpoint (all mutations go through `versions/api.ts`), diff is net-additive under `packages/docs/src/versions/` plus the i18n adjustment. No under-build, no over-build.

## 2. Code quality

**Quality: Approved.** No P0/P1. The follow-up-lane fix removes one of my prior P2s and the dedup concern. The remaining items are non-blocking P2 polish, none with current user impact (shell is unmounted); fold them in before/with the adapter phase that first mounts the shell.

- **Stuck `loadingMore` if a mutation fires during load-more** — `VersionHistoryPanel.tsx` `onLoadMore`: row mutation buttons are gated by `busy` only, not `loadingMore`. A restore/delete/rename/save during an in-flight load-more calls `listGuard.begin()`, which aborts the follow-up and bumps the generation; the follow-up then returns early in both `catch` and `finally`, so `setLoadingMore(false)` never runs and the button + filter tabs stay disabled until the panel is reopened. Mirrors the existing `BoardVersionPanel` behavior and self-heals on reopen, but the shell is the right place to reset the flag unconditionally.
- **`commitRename` missing `if (busy) return`** — unlike `onCreateSnapshot`; the Enter-key path is unguarded so rapid Enter can fire concurrent `renameVersion` calls. Bounded (idempotent PATCH, input unmounts on first commit) — redundant request, not corruption. Add for symmetry.
- **Blanket `onMouseDown={e => e.stopPropagation()}` on the modal body** — standard pattern, but can interfere with `document`-level click-outside listeners inside injected `renderPreview` content; the `e.target === e.currentTarget` overlay check avoids the side effect.
- **UX nits** — save-current `<input>` has no `onKeyDown` (Enter doesn't submit, unlike rename); the rename input's `Escape` doesn't `stopPropagation`, so one Escape can both cancel a rename and close an open preview modal.

## 3. Overall verdict

**APPROVE.** Spec ✅ and no P0/P1 at the current head. The delta fixed the follow-up-lane race guard bug cleanly with a non-vacuous regression test; the i18n removal is deliberate, documented, and has no user impact while the shell is unmounted. Remaining findings are P2 polish for the adapter phase.

## Note for a human reviewer (security-sensitive label)
Keyword-triggered classification. The diff has no auth/token/credential/middleware logic: role gating is UI-only (`canSnapshot`/`canRestoreVersion`, restore+delete admin-only), and all mutations still route through the existing backend-gated `api.ts` endpoints. No new trust boundary introduced.

yujiawei (changes requested):
# Code Review — PR #656 (octo-web)

Reviewed at head SHA `06179dbf634e1b64dbeefc7608c44ea42f0d0af0` against `main` (merge-base `6ca01c31`). This is a phase-1, purely-additive change: 4 new files (`versions/VersionHistoryPanel.tsx`, `versions/raceGuard.ts`, + 2 tests), zero modified files, nothing wired to any end yet.

## 1. Scope / spec compliance — ✅

Verified by inspection:
- `git diff --stat main...HEAD` shows only the 4 new files (1080 additions, 0 deletions). Nothing else touched.
- No existing module imports `VersionHistoryPanel` or `raceGuard` yet — the shell is genuinely unmounted, so there is no user-visible change this phase and the deferred i18n keys (`filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice`) cannot surface a raw key to a user.
- All mutations route through the existing `./api.ts` (`listVersions/createNamedVersion/restoreVersion/renameVersion/deleteVersion`); no new endpoint is introduced. Preview/diff are correctly the only pluggable seam.
- Permission gating uses `canSnapshot(role)` (save/rename) and `canRestoreVersion(role)` (restore/delete) from `auth/roles.ts`, matching the stated matrix (writer: save+rename; admin-only: restore/delete). Tests pin all three role paths. No over-build or divergence found.

The additive, per-end-revertible structure is a good way to land this.

## 2. `raceGuard.ts` — solid

The primary/follow-up two-lane design is correct and the tests actually prove the two hard guarantees (true on-the-wire abort via `AbortController`, and discard-late-response via monotonic `generation` + separate `followToken`). The subtle case — a superseded follow-up whose request resolved a hair before its abort — is correctly handled by the separate `followToken` (a shared generation alone could not express it), and `raceGuard.test.ts:50-62` pins exactly that regression. No defects found here.

## 3. Blocking findings

### P1 — `loadingMore` gets stuck `true` forever when a load-more is superseded (`VersionHistoryPanel.tsx:202-224`)

`onLoadMore`'s `finally` only resets the flag when the ticket is still current:

```ts
} finally {
  if (isCurrent()) setLoadingMore(false)
}
```

When an in-flight load-more is superseded by any primary `begin()` — a filter switch (`selectFilter` → `setFilter` → the `refresh` effect → `listGuard.begin()`), or a restore/delete/refresh — `isCurrent()` returns `false`, so `setLoadingMore(false)` is skipped and **nothing else ever resets it**. The new list renders with `nextCursor != null`, but the Load-more button is `disabled={loading || loadingMore}` and `onLoadMore` early-returns on `if (loadingMore) return`. Result: the Load-more button is permanently dead for the rest of the panel's mounted lifetime. Reachable with two ordinary clicks (click Load-more, then switch filter before the page lands).

This is squarely in the interaction-race class this PR was built to get right, and it is the one supersede path the otherwise-thorough tests don't cover (they gate append and `errorMore`, but not the loading flag).

Suggested fix: reset unconditionally, since the entry guard `if (loadingMore) return` already prevents two concurrent load-mores, so there is never a legitimately-still-loading follow-up when this `finally` runs:

```ts
} finally {
  setLoadingMore(false)
}
```

(and gate the `setItems`/`setError` by `isCurrent()` as today). A regression test for "load-more in flight → filter switch → load-more button re-enabled on the new list" would lock this down.

## 4. Non-blocking (P2)

- **setState after `await refresh()` in the four mutation handlers** (`onCreateSnapshot:248`, `commitRename:278`, `onDelete:297`, `onConfirmRestore:317-323`). `refresh()` guards its own state via `isCurrent()`, but the trailing `setBusy(false)` / `setNotice` / `onRestored?.()` in the caller run unconditionally after the await. If the panel unmounts mid-refresh these fire on an unmounted component. The unmount effect aborts the guards but does not stop the already-running async continuation. Harmless under React 18 (no more unmount warning, and the network is aborted), but it's the same "re-check liveness after every await" discipline the rest of the file follows — worth a small `mounted` ref or an `isCurrent`-style guard for consistency. Not a blocker.
- **Delete confirmation uses native `window.confirm` (`:282`) while restore uses the in-panel centered confirm box (`:542-565`).** The file comment at `:140` states the in-panel box is the intended model. Since this shell is meant to unify UX across doc/sheet/board, the delete path should use the same in-panel confirm before an adapter mounts it. Cosmetic/consistency, not blocking.

## 5. Security notes (this PR was flagged security-sensitive)

- Frontend role gating is UX-only by design; the api.ts header documents that the backend role-gates every endpoint. That contract is correct, but **before an adapter wires this shell to a live end, please confirm on the backend that `POST /docs/:id/versions/:seq/restore` and `DELETE /docs/:id/versions/:seq` enforce admin-only** — the frontend gate is not a security boundary. Flagging for a human to verify in the backend, not a defect in this diff.
- `renderPreview` / `renderDiff` render host-injected `ReactNode`s directly; this is the standard React composition pattern and the shell itself never uses `dangerouslySetInnerHTML`, so there is no injection surface introduced here. Error text flows through `t(previewErr)` inside JSX, which React auto-escapes — no XSS from the default classifiers. No issue.

## Verdict

Spec/scope is clean and `raceGuard` is well done. Blocking on the single P1 (`loadingMore` permanently stuck after a superseded load-more) — it's a real functional defect in the load-bearing interaction logic this PR exists to standardize, and it's currently untested. Fix that (and ideally the two P2s alongside), and this is good to merge.

Note on process: two independent secondary review tools were attempted for this security-sensitive PR; both were unavailable this run (one timed out, one failed on a gateway configuration error). This verdict rests on direct code inspection plus an independent multi-lens verification pass — it does not assume the absent tools would have approved.

OctoBoooot (changes requested):
## Review: unified VersionHistoryPanel shell + race guard util (#656) — correcting my verdict

Flipping APPROVE → REQUEST_CHANGES. yujiawei caught a real state-management bug I approved past: gating `setLoadingMore(false)` on `isCurrent()` in `onLoadMore`'s `finally` means a superseded load-more never clears the flag, permanently disabling the load-more button. Byte-verified and reachable.

Risk tier: normal; size/XL → `needs-human-review`. Base `main`; PR is now ready (not draft), `reviewDecision=CHANGES_REQUESTED`.

### Major (missed on my passes, byte-verified)
`VersionHistoryPanel.onLoadMore` (`:202-221`):
```
setLoadingMore(true)
try { await listVersions(...) ; if (!isCurrent()) return ; ...append... }
finally { if (isCurrent()) setLoadingMore(false) }   // ← reset gated on isCurrent
```
If a `begin()` supersedes this follow-up before its `finally` runs — a filter-tab switch / restore / delete / save-snapshot, all of which call `refresh()` and are **not** disabled while `loadingMore` — then `isCurrent()` is false, so `setLoadingMore(false)` is **skipped**. And `refresh` sets/clears only `loading`, never `loadingMore` (`:181-182`), so nothing else resets it → **`loadingMore` sticks `true`** → the load-more button (`disabled={loading || loadingMore}`, `:536`) is permanently disabled until remount.
**Reachable:** click Load-more (slow request in flight) → switch filter tab → the page resolves, `finally` gated out → button dead. Major (functional, common interaction; recoverable by remount; no data-loss/security).

**Fix:** clear the flag unconditionally on the guard's own terms — e.g. `finally { setLoadingMore(false) }` (a superseded page setting it false is harmless; a newer load-more re-sets it true), or reset `loadingMore` when `begin()` supersedes. The correct principle: the guard should discard the stale *result* (which it does), but the *loading flag* must always be cleared — gating the reset on `isCurrent()` is what wedges it.

### Also (yujiawei 🟡, fair)
- Mutation handlers `await` without an `isCurrent`/mounted check before `setState` → possible setState-after-unmount. Non-blocking robustness.

### Accountability
This is the second facet of the race-guard area I've had to correct on #656 — I verified the guard's `isCurrent`-after-await discipline (discarding stale results) but missed that gating the *flag reset* on `isCurrent()` creates a stuck-flag on supersede. The guard correctly drops the stale result; the loading flag must not be conditional on it. Credit: yujiawei. The follow-up-lane `followToken` fix from the prior round remains correct and unaffected.

Jerry-Xin (changes requested):
**REQUEST_CHANGES** — flipping my earlier APPROVE at this same head (06179dbf). I approved past this and am correcting: yujiawei surfaced a real deadlock I missed. Verified against the source at this commit.

### 🔴 Blocking: superseded load-more permanently deadlocks the "Load More" button

`onLoadMore` (`VersionHistoryPanel.tsx`, L202–223):

```
if (loadingMore || nextCursor == null) return   // re-entry guard
const { signal, isCurrent } = listGuard.current.beginFollowUp()
setLoadingMore(true)
...
try {
  const res = await listVersions(...)
  if (!isCurrent()) return          // success branch early-return
  ...
} catch {
  if (!isCurrent()) return          // catch branch early-return
  ...
} finally {
  if (isCurrent()) setLoadingMore(false)   // the ONLY loadingMore reset, gated on isCurrent()
}
```

`setLoadingMore(false)` runs in exactly one place — the `finally`, gated behind `isCurrent()`. When a `begin()` (filter switch / refresh / restore) or a newer load-more supersedes an in-flight follow-up A, A's `isCurrent()` becomes `false`. A's continuation then hits the success/catch `if (!isCurrent()) return`, returning before the `finally`'s guarded reset does anything (it's gated on the same false `isCurrent()`). Result: `loadingMore` is stuck `true` forever, and the Load More button (`disabled={loading || loadingMore}`, L370 / L536) is **permanently disabled**.

Reachable in normal use: switch filters or refresh while a load-more page is in flight.

**Nuance worth flagging:** this deadlock was *exposed/introduced by the very followToken fix* that closed the earlier follow-up-lane double-append gap (the one I previously CR'd, then approved). Before that fix, a superseded follow-up's `isCurrent()` stayed `true`, so the `finally` still reached `setLoadingMore(false)`. Making `isCurrent()` correctly report `false` for superseded follow-ups is what now causes L214/L219 to early-return past the cleanup. Classic "the stale-guard also skips the cleanup" bug. Good catch by @yujiawei.

**Fix:** reset `loadingMore` **unconditionally** on the follow-up's completion, independent of `isCurrent()` — e.g. `finally { setLoadingMore(false) }` (a setState on a superseded-but-mounted component is fine; if genuinely unmounted, the guard util's abort covers the fetch), or reset it before the `isCurrent()` early-returns. The state cleanup must not sit behind the stale-guard, because the superseded path still owns the `loadingMore` flag it set.

### 🟡 Non-blocking (also raised by @yujiawei)
Mutation handlers do `setState` after `await` without an `isCurrent()` guard → possible setState-after-unmount warning. Low risk; worth a follow-up but not blocking.

Everything else at this head still looks good — the double-append gap is genuinely closed; this is purely the cleanup path that the fix left gated.

OctoBoooot (dismissed):
## Review: unified VersionHistoryPanel shell + race guard util (#656) — delta re-review (loadingMore deadlock fixed)

The new commit fixes the stuck-`loadingMore` deadlock (the three-way CR): the `finally` now clears `loadingMore` **unconditionally**, no longer gated on `isCurrent()`. Verified the fix and that it introduces no new race. Flipping REQUEST_CHANGES → APPROVE.

Risk tier: normal; size/XL → `needs-human-review` stands. Base `main`; ready (not draft).

### Resolved — the load-more button deadlock
`onLoadMore`'s `finally` changed from `if (isCurrent()) setLoadingMore(false)` to plain `setLoadingMore(false)`. So a superseded load-more (filter switch / refresh / restore while a page is in flight) now always clears the flag → the button can no longer wedge. This is exactly the prescribed fix, and the comment states the principle correctly: the guard discards the stale *result* (via the early returns above), but the *loading flag* must not be gated on `isCurrent()`.
- **No new race:** a second load-more can't start while `loadingMore` is true (`onLoadMore` early-returns on `if (loadingMore || ...)`), so there's never a concurrent follow-up whose flag the older `finally` could wrongly clear — the only supersede path is `begin()` (refresh/filter/restore), which doesn't touch `loadingMore`. So the unconditional reset is correct and the only reset. A superseded-but-mounted setState is harmless; true unmount is covered by the guard's `abort()`.
- **Test:** +49 pins the regression (superseded load-more → `loadingMore` cleared → button re-enabled).

### Standing (unchanged, correct)
The follow-up-lane `followToken` fix from the prior round is untouched and still correct (a superseded follow-up's `isCurrent()` goes false via the token). Primary-lane semantics, restore confirm-gating, XSS-safety all as before.

### Accountability
This closes the second race-guard-area facet I approved past on #656 (after the follow-up-lane gap). Both were caught by peers; both are now fixed. The compounded lesson stands: verifying "isCurrent discards stale results" is necessary but not sufficient — loading/UI flags must be cleared regardless of the staleness check. Credit: yujiawei (and Jerry-Xin, converged).

Jerry-Xin (dismissed):
APPROVED.

Verified the P1 loadingMore deadlock fix at head 62f7976c.

**Root cause (confirmed real):** the load-more `finally` gated `setLoadingMore(false)` on `isCurrent()`. When a primary `begin()` (filter switch / refresh / restore) or a newer load-more supersedes an in-flight follow-up, its ticket reports non-current, so the old reset was skipped. Since `refresh()` only touches `loading` and never `loadingMore`, the flag stuck `true` and `disabled={loading || loadingMore}` left the Load More button permanently disabled until remount.

**Fix is correct:** clearing the flag unconditionally in `finally` is the right owner semantics — the follow-up that set the flag releases it, while the stale *result* is still discarded by the `isCurrent()` early-returns above. No premature re-enable: during a superseding refresh, `loading` keeps the button disabled until the fresh list lands. Unmount is covered by the guard's `abort()`, and a superseded-but-mounted `setState` is a harmless no-op.

**Regression test locks the behavior:** proved red→green locally — reverting only the source to the gated `if (isCurrent()) setLoadingMore(false)` makes the new test fail with `btn.disabled === true` (the deadlock); the fix flips it to `false`. `vitest run src/versions` → 8 files, 84 passed (was 83).

No blocking issues. Ship it.

Jerry-Xin (changes requested):
CHANGES_REQUESTED.

Re-reviewed the P2 changes at head `d7d25630` (P1 `62f7976c` + P2 `d7d25630`). One blocking issue in the mounted-ref guard; the delete-confirm change is good.

## 1. In-panel delete confirm — good

`window.confirm` is gone; delete now routes through the same centered `.octo-version-confirm` box as restore (`confirmDelete` state, row button → `setConfirmDelete(v)`, box confirm → `onDelete`). `confirmDelete` is cleared on both the success and the error path, and the box buttons are `disabled={busy}`. The new test asserts the native dialog is never called and `deleteVersion` fires only after the box's confirm — correct and well-scoped.

## 2. Mounted-ref guard — BLOCKING: the ref is never re-armed, so StrictMode wedges it

`mounted` is `useRef(true)` and is only ever set to `false` in the unmount cleanup — it is **never set back to `true` in the effect body**:

```
const mounted = useRef(true)
...
useEffect(() => {
  const list = listGuard.current
  const preview = previewGuard.current
  return () => {
    mounted.current = false   // set false on cleanup, never restored on re-mount
    list.abort()
    preview.abort()
  }
}, [])
```

The docs app mounts its root under `<StrictMode>` (`packages/docs/src/main.tsx:41`, and `dev/main.dev.tsx`). React 18 StrictMode double-invokes effects in dev: mount → cleanup → mount. The cleanup sets `mounted.current = false`, and because nothing restores it on the second mount, **`mounted.current` stays `false` for the live component**. Every mutation handler then hits `if (!mounted.current) return` right after its first `await`, so the tail no-ops:

- `setConfirmDelete(null)` / `setConfirmRestore(null)` skipped → the confirm box never closes
- `setBusy(false)` skipped → `busy` stuck `true` → all row/box buttons disabled
- the reconcile `refresh({ soft: true })` and `setNotice(...)` skipped

Net effect: in dev, the first delete/rename/restore/snapshot bricks the panel — the network call fires but the UI freezes with the confirm box open and everything disabled.

Repro (rendered the panel under `<StrictMode>`, clicked delete → box confirm): `deleteVersion('d_1', 7)` fires, but `.octo-version-confirm` stays mounted with disabled buttons and never closes. Adding one line — `mounted.current = true` as the first statement of the mount effect — makes the box close and `busy` release. The guards themselves survive StrictMode fine because `begin()` re-arms them on the refresh effect; only `mounted` lacks that re-arm.

Why the suite doesn't catch it: `@testing-library/react` renders without StrictMode, so the new delete test (and all 85) pass regardless. This is exactly the class of bug the repo's other StrictMode-aware comments guard against.

Scope, stated honestly: this is dev-only (StrictMode double-invoke doesn't happen in production builds, where the effect runs once and `mounted.current` stays `true`), and it's currently latent because no adapter mounts the shell yet. But the mounted-ref guard is the whole point of this P2, the adapter phases will mount it under the StrictMode root, and the guard is wrong as written — so it should be fixed before merge rather than shipped broken.

Fix:

```
useEffect(() => {
  mounted.current = true
  const list = listGuard.current
  const preview = previewGuard.current
  return () => {
    mounted.current = false
    list.abort()
    preview.abort()
  }
}, [])
```

And I'd add one StrictMode-wrapped mutation test (delete or restore) so the guard's actual purpose is locked — the current tests can't see this failure mode.

## 3. Coverage / typecheck claims

- Typecheck: clean for the changed files — `tsc -p tsconfig.json` reports **0 errors under `src/versions/`**. (The whole-monorepo run surfaces ~2.6k pre-existing react@17 type-resolution errors in unrelated packages; none are in this PR's files, which are purely additive.)
- Tests: the versions suite is green at head — 8 files, **85 passed** (84 from the P1 baseline + the new P2 delete test). Verified after inlining the `@douyinfe/semi-icons` dep to work around a local dep-optimizer flake; the code-level result is green.

Once the `mounted` ref is re-armed in the effect body (and ideally a StrictMode test added), this is good to go.

yujiawei (dismissed):
# Code Review — PR #656 (octo-web)

**Scope reviewed:** head SHA `62f7976c090b2883547ad1075bc620339ab69f1d`, base `main`. 4 new files, +1136, 0 deletions — no existing file is touched.

- `packages/docs/src/versions/raceGuard.ts` (+ test)
- `packages/docs/src/versions/VersionHistoryPanel.tsx` (+ test)

This is a Phase-1 "add the shell + guard + tests only" change: the unified panel is not wired to any end yet, so it is intentionally dead code with zero user-visible impact until an adapter phase mounts it.

## 1. Spec compliance

**Spec: ✅**

- **Under-build:** none. The stated Phase-1 deliverables are all present — the unified `VersionHistoryPanel` shell (mixed list + filter tabs + counts + pagination, save/rename/delete/restore, in-panel restore confirm, centered preview/diff modal with Esc/overlay-close/stopPropagation, role gating) and the reusable `createRaceGuard` (AbortController + monotonic generation, plus the follow-up token added in commit 2), with tests for both.
- **Over-build:** none. The shell adds **no new endpoint** — list/create/rename/delete/restore/state all reuse the existing `./api.ts` REST layer. No end is mounted (verified: no non-test import of `VersionHistoryPanel`), so no premature wiring.
- **Deviation:** none. The 6 new i18n keys (`filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice`) are, as the file header documents, deliberately **deferred** to the adapter phase. Verified they are absent from the `docs.version` namespace in **both** `en-US.json` and `zh-CN.json`, so locale parity is intact and the parity guard test stays green. Because the shell is unmounted, no user can render a raw key this phase.

## 2. Code quality

**Quality: Approved**

Verified locally at the head SHA:

- **Tests:** all 22 pass (`raceGuard.test.ts`, `VersionHistoryPanel.test.tsx`, `i18n.test.ts`).
- **Typecheck:** the new files are clean. The two `tsc` errors reported by the package typecheck are in **untouched** files (`packages/dmworkbase/src/i18n/format.ts`, `packages/docs/src/members/sort.ts`) and pre-date this PR.

Focused correctness review:

- **Race guard (`raceGuard.ts`):** the two-lane design is correct. `begin()` bumps `generation` (capturing the new value in the primary ticket, which checks only `gen === generation`) and also bumps `followToken`, so any outstanding follow-up ticket — whose `isCurrent()` requires **both** `gen === generation` **and** `token === followToken` — is invalidated without wrongly invalidating the just-created primary ticket. Two rapid `beginFollowUp()` calls correctly leave the first aborted **and** non-current (the commit-2 fix, with a matching regression test) and the second current. `abort()` nulls both controllers and bumps both counters. No AbortController leak; no interleaving lets a stale response pass `isCurrent()`. This is a strict superset of the old `previewGuard` (which never aborted on the wire).
- **Load-more state machine (`VersionHistoryPanel.tsx`):** the commit-3 fix making `setLoadingMore(false)` unconditional in `finally` is correct. Only one load-more can be in flight (`if (loadingMore || nextCursor == null) return` + disabled button), and the guard's `isCurrent()` early-returns still discard a superseded page's rows/error; the loading flag must always release so a superseding refresh/filter-switch/restore doesn't wedge the button disabled forever. The asymmetry vs the conditional `setLoading(false)` in `refresh()` is intentional and sound.
- **Security (this PR is security-classified):** role gating is UI-layer only and correctly derived — `canSnapshot` (writer+) gates save/rename, `canRestoreVersion` (admin-only) gates restore/delete; readers see preview only. `api.ts` states the backend is authoritative on every endpoint, so the UI gate is defense-in-depth, not the trust boundary — appropriate. No `dangerouslySetInnerHTML`; all user-controlled fields (`label`, `createdBy`, `restoredFrom`, names) render as auto-escaped JSX text. `renderPreview`/`renderDiff` are host-injected callbacks — their per-end safety is out of scope for this phase and should be checked in each adapter PR. No secret/token handling in the new code.

### Coverage / blind spots (things a human may want to keep in mind)

- Backend authorization for restore/delete is **assumed** authoritative (not verified here) — the UI gate alone is not a security boundary. Worth a one-line confirm that the backend enforces admin-only on `POST .../restore` and `DELETE .../:seq`.
- The host-supplied `renderPreview`/`renderDiff`/`loadPreviewState` are the real injection/decoding surface; each adapter phase (doc/sheet/board) should be reviewed for payload-decode safety when it lands.
- The 6 deferred i18n keys **must** ship in the same PR that first mounts an end, or that end will show raw keys. Not a defect now; a checklist item for the adapter phase.

## 3. Overall verdict

**APPROVE.** Spec ✅ and Quality Approved. No P0/P1. Two non-blocking nits below.

- **[nit]** `api.ts` restore docstring says "ADMIN/owner only" but omits the "backend role-gates every endpoint" phrasing the other mutating endpoints carry; add it for consistency (the file header already states the global contract, so purely cosmetic).
- **[nit]** Land the 6 deferred `docs.version.*` keys together with the first adapter that mounts the shell.

OctoBoooot (changes requested):
## Review: unified VersionHistoryPanel shell + race guard util (#656) — delta re-review (new commit introduces a StrictMode freeze)

**Verdict:** Request changes
The new commit (`d7d2563` — in-panel delete confirm + `mounted` ref to guard setState-after-unmount) introduces a real bug: the `mounted` ref is set false on cleanup but never re-armed in the effect body, so under React 18 StrictMode it stays false for the mounted lifetime and every mutation handler no-ops its trailing state resets → the panel freezes in dev. Byte-verified; converges with Jerry-Xin.

Risk tier: normal; size/XL → `needs-human-review`. Base `main`; ready (not draft).

### Major (new in this commit)
`const mounted = useRef(true)` (`:160`); the abort/cleanup effect (`:200-210`, deps `[]`) sets `mounted.current = false` in its cleanup **but the effect body does not set it back to `true`**. Under StrictMode (dev), effects run mount → cleanup → remount: the cleanup flips `mounted.current = false`, then the re-run body does **not** restore it → `mounted.current` is `false` for the real mounted lifetime. Every mutation handler (`onCreateSnapshot`/`onRename`/`onDelete`/`onRestore`) guards its trailing `setBusy(false)` / `setConfirmDelete(null)` / `setConfirmRestore(null)` / `onRestored` with `if (!mounted.current) return` → all skipped → after a delete/restore, `busy` stays true and the confirm box never closes → **UI frozen, buttons permanently disabled** (Jerry-Xin's second 🔴 is the same root cause). Same *class* as the just-fixed loadingMore deadlock: a liveness flag left in a bad state — here the mounted ref isn't re-initialized on the StrictMode remount.

Severity: **major**, not blocker — it's **dev-only** (production React 18 doesn't double-invoke effects, so `mounted` stays true and handlers work). But it breaks the entire panel's mutations in dev/StrictMode, which is a severe DX/correctness regression and would make the feature look broken to any developer.

**Fix (standard):** re-arm on mount — set `mounted.current = true` at the top of the effect body (or a dedicated mount effect), so a StrictMode remount restores it:
```
useEffect(() => {
  mounted.current = true
  return () => { mounted.current = false; list.abort(); preview.abort() }
}, [])
```

### Standing (unchanged, correct)
The loadingMore-deadlock fix (unconditional `setLoadingMore(false)`) and the round-1 `followToken` follow-up-lane fix both remain correct. The in-panel delete confirm (replacing native `window.confirm`) is a good unification and is itself fine — the bug is purely the un-re-armed `mounted` ref.

### Note
This is a new commit adding a new bug while fixing yujiawei's setState-after-unmount 🟡 — the guard it added is broken under StrictMode. One-line fix. Credit: Jerry-Xin.

OctoBoooot (approved):
## Review: unified VersionHistoryPanel shell + race guard util (#656) — delta re-review (StrictMode freeze fixed)

**Verdict:** Approve (flipping REQUEST_CHANGES → APPROVE)
The new commit re-arms the `mounted` ref on (re)mount, fixing the StrictMode freeze I blocked on at `d7d2563`. Byte-verified: the fix is exactly the prescribed one-liner and the surrounding guards are now correct. Flipping my verdict.

Risk tier: normal; size/XL → `needs-human-review` stands. Base `main`; ready (not draft).

### Resolved — the StrictMode mounted-ref freeze
`VersionHistoryPanel.tsx:201`: `mounted.current = true` is now the first statement inside the `[]`-deps abort effect (body opens `:200`), **before** the cleanup return (`:205`) that sets it false. So under React 18 StrictMode (mount → cleanup → remount), the re-run body restores it to `true` → `mounted.current` is true for the real mounted lifetime. The 15 mutation-handler guards (`if (!mounted.current) return`, `:256`–`:345`) now fire correctly instead of no-oping, so delete/restore/rename/create no longer wedge `busy`/confirm state. This is the standard fix (re-arm on mount) and is complete.
- **Test:** `+31` in `VersionHistoryPanel.test.tsx` pins the regression (mount→unmount→remount → handler still resets state). Good.

### Standing (unchanged, correct)
The delta is exactly `+1` line in `.tsx` + the test — the prior-round fixes remain untouched and correct: unconditional `setLoadingMore(false)` in `onLoadMore`s `finally` (load-more button no longer wedges on supersede), and the round-1 `followToken` follow-up-lane fix.

### Accountability
This closes the third and final facet of the race-guard area on #656. All three (follow-up-lane `followToken`, loadingMore-deadlock, mounted-ref StrictMode) were caught across rounds by peers and me; all three are now fixed with the prescribed changes. Compounded lesson holds: a liveness/mount flag must be both cleared on teardown AND re-armed on (re)mount — checking only the teardown side misses the StrictMode remount. Credit: Jerry-Xin.

Jerry-Xin (approved):
The PR is in scope for `octo-web` and adds a relevant shared version-history shell plus race guard; I found no blocking issues.

**💬 Non-blocking**
- 🟡 Warning: [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-8n5dxvtz/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:335) does not apply the default schema-aware restore error mapping. Preview maps `VersionSchemaNewerError` / `VersionSchemaIncompatibleError`, but restore without `restoreErrorKey` falls back to generic `docs.version.errorRestore`, which would regress the existing doc panel’s specific incompatible-version message when the adapter phase uses the default.
- 🟡 Warning: [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-8n5dxvtz/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:238) clears `loadingMore` unconditionally in `finally`. If the component unmounts while load-more is in flight, the abort will still let the promise settle and this can call state after unmount. Low practical risk in React 18, but it contradicts the mounted-ref pattern used elsewhere in the component.
- 🔵 Suggestion: [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-8n5dxvtz/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:511) references new `docs.version.*` keys that are not yet in the docs namespace. Since the shell is not mounted in this phase, this is not user-visible now, but the adapter PR should add the keys before first render.

**✅ Highlights**
- `createRaceGuard()` has a clear primary/follow-up model and covers both abort-on-supersede and stale-result discard.
- The shell keeps destructive actions role-gated consistently with existing `canSnapshot` / `canRestoreVersion`.
- Tests cover the important race, modal, permission, inline rename, and StrictMode liveness paths.

Verification note: I could not run `vitest` or `tsc` locally because this checkout has no installed `node_modules/.bin` tools; `pnpm exec vitest` and `pnpm exec tsc` both failed with command-not-found.

boris-clark (approved):
**APPROVE** — re-reviewed the delta from `d7d25630` → `e66e3595` (2 files, +32/-0). The P3 StrictMode-liveness blocker Jerry-Xin (15:35) and OctoBoooot (15:37) pinned at `d7d25630` is resolved, the new regression test is genuinely gated, and nothing else changed.

### 1. P3 blocker resolved — verified
`VersionHistoryPanel.tsx` now sets `mounted.current = true` as the first statement of the abort/cleanup effect body (deps `[]`, L201):

```
useEffect(() => {
  mounted.current = true // re-arm on (re)mount
  const list = listGuard.current
  const preview = previewGuard.current
  return () => { mounted.current = false; list.abort(); preview.abort() }
}, [])
```

Under React 18 StrictMode (dev) the effect runs mount → cleanup → remount; the cleanup flips `mounted.current = false`, and the re-run body now restores it to `true`, so the ref is live for the real mounted lifetime. The four mutation handlers (`onCreateSnapshot` L256/261, `commitRename` L288/293, `onDelete` L309/315, `onConfirmRestore` L334/340/345) no longer early-return past their `if (!mounted.current) return` gates, so the trailing resets (`setBusy(false)`, `setConfirmDelete(null)` / `setConfirmRestore(null)`, the soft `refresh`, `setNotice`) execute normally. This is exactly the fix both CRs requested, matching their quoted remediation line-for-line.

### 2. Regression test is red-gated, not evergreen — verified
The new `describe('VersionHistoryPanel — StrictMode liveness (mounted-ref re-arm)')` renders the panel under `<StrictMode>`, drives a real delete through the in-panel confirm box, asserts `deleteVersion('d_1', 7)` fires, and — the load-bearing assertion — waits for `.octo-version-confirm` to disappear (which only happens if `onDelete` runs past its liveness gate to `setConfirmDelete(null)`).

I confirmed it is genuinely gated:
- With the fix in place: file passes, **15/15**.
- After removing the `mounted.current = true` line: the StrictMode test **fails** — the confirm box stays mounted with buttons stuck `disabled` and `waitFor(... toBeNull())` times out.

So it is red-without-fix / green-with-fix, and covers the exact failure mode the standard suite (rendered without StrictMode) could not see.

### 3. No side effects / no regression — verified
The delta touches only these two files. `raceGuard.ts` is untouched; the P1 fix (unconditional `setLoadingMore(false)` in the `onLoadMore` finally) and the P2 in-panel delete confirm are unchanged. The re-arm is idempotent for the non-StrictMode / production path (the effect runs once and `mounted.current` was already `true` from `useRef(true)`), so it introduces no new runtime behavior there.

### 4. Both pending CRs can be resolved
Jerry-Xin (15:35) and OctoBoooot (15:37) both pinned the un-re-armed `mounted` ref → StrictMode freeze as the sole blocker at `d7d25630`, and both explicitly asked for (a) re-arming `mounted.current = true` at the top of the mount effect and (b) a StrictMode-wrapped mutation test. Both are satisfied by this commit. No new issues in the delta.

OctoBoooot (commented):
## Review: #656 — recording a confirmed minor (byte-verifying Jerry-Xin 🟡 #3)

**This does not change my APPROVE of the delta** (the StrictMode mounted-ref fix at `e66e3595` is correct). Recording one verified finding with its exact scope, since the peer note ("`docs.version.*` keys not yet added") reads broader than the reality.

### Minor — one missing i18n key (`version.staleNotice`)
Byte-verified against `packages/docs/src/i18n/en-US.json` and `zh-CN.json` at head `e66e3595`:
- **44 of the 45** referenced `docs.version.*` keys resolve correctly (the `docs.` segment is the *namespace*, stripped inside the JSON — so the keys live under a top-level `version` object, and the i18n parity test confirms the group exists in both locales).
- **Exactly one is missing from both locales: `version.staleNotice`.** It is referenced on four live paths — `VersionHistoryPanel.tsx:266, 299, 322, 349` (the failed-mutation / superseded "stale" notice). `restoredNotice` is present; `staleNotice` is not.
- **Why CI stays green:** `i18n.test.ts` only enforces (a) zh-CN/en-US *shape parity* — both are equally missing the key, so parity holds — and (b) that the top-level `version` group exists. Neither test checks that every key *referenced in code* is defined, so the gap slips through.
- **Impact — minor, latent:** the panel is not mounted anywhere yet (its own header comment says so; no importer in `module.tsx`/index files). Once it is wired up and a mutation fails or is superseded, `t(docs.version.staleNotice)` renders the raw key string instead of translated text. Degrades gracefully (no crash), zh/en equally affected.
- **Fix:** add a `staleNotice` entry under `version` in both `en-US.json` and `zh-CN.json` (e.g. EN "This version list changed while loading — refresh to see the latest." / a matching zh string). Optionally extend `i18n.test.ts` to assert referenced keys resolve, which would have caught this.

### On the other two 🟡s (verified, non-blocking)
- **Unconditional `setLoadingMore(false)`** — benign by design: it was the prescribed fix for the load-more button wedge; a superseded-but-mounted `setState` is harmless and a genuine unmount is a React no-op (and now also guarded by the re-armed `mounted` ref). Not a defect.
- **schema-aware restore error mapping** — design suggestion, no correctness defect.

Verdict unchanged: **Approve** (delta is clean). Credit: Jerry-Xin for #3 — refined here to the single key so the fix is unambiguous.

yujiawei (approved):
# Code Review — PR #656 (octo-web)

Reviewed at head SHA `e66e3595cd4ab87e65f9d94b063de4070cf9b98a` against `main` (merge-base `6ca01c31`). Verified locally on a fresh checkout of the head: `vitest run src/versions` → **8 files, 86 passed**; `tsc -p tsconfig.typecheck.json` → the only two errors are pre-existing files outside this PR (`dmworkbase/i18n/format.ts`, `members/sort.ts`), **zero new errors in `versions/*`**, confirming the PR's typecheck claim. This PR was auto-classified security-sensitive; a security note is at the end.

## 1. Scope / spec compliance — ✅

- `git diff --stat main...HEAD` = exactly **4 new files, +1229, 0 deletions**. No existing file is touched.
- Confirmed by grep that `VersionHistoryPanel` and `raceGuard` have **zero importers outside their own tests** — the shell is genuinely unmounted, so there is no user-visible change this phase, and each future adapter can be reverted independently.
- The deferred i18n keys (`filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice`) are indeed **absent** from the `docs.version.*` namespace, matching the stated intentional deferral; every other key the shell references already exists. Since nothing mounts the shell, no user ever sees a raw key. Not a defect.
- No over-build: no new endpoint (all mutations reuse `./api.ts`), props contract matches the design.

## 2. Code quality — Approved

`raceGuard.ts` is a correct consolidation of the AbortController + monotonic-generation pattern. The separate `followToken` correctly makes a superseded load-more report non-current even when its request resolved a hair before its abort — verified against the `raceGuard.test.ts` red→green assertions. The `onLoadMore` `finally` clears `loadingMore` unconditionally (avoiding the earlier deadlock), and the `mounted` ref is re-armed at the top of the cleanup effect so StrictMode's mount→cleanup→remount stays live. All of these earlier-flagged issues are resolved at this head.

The findings below are all **non-blocking** (P2 / nit). Each was independently verified against the source; none is a merge blocker.

### P2 — Restore default error path loses schema-specific messages that preview keeps
`VersionHistoryPanel.tsx:335` — `onConfirmRestore` falls back to a flat key:
```ts
setError(t(restoreErrorKey ? restoreErrorKey(e) : 'docs.version.errorRestore'))
```
whereas `onPreview` (`:370`) falls back to the schema-aware `defaultErrorKey(e)`, which maps `VersionSchemaNewerError` / `VersionSchemaIncompatibleError` to distinct messages. `restoreVersion` (`api.ts:191-206`) throws those *same* two typed 409 errors, so an adapter that does **not** inject `restoreErrorKey` will collapse a schema-incompatible/newer restore to the generic "Failed to restore version." The prop's own doc comment (`:63`) says restore uses the "same default" as preview, so the implementation contradicts its documented contract. The existing single-end panel maps these to `errorRestoreIncompatible` (a key already present in both locales but unused by the shell). Not user-visible this phase (nothing is wired), but it should be fixed before an adapter mounts the shell without its own classifier — suggest a `defaultRestoreErrorKey` symmetric with `defaultErrorKey`.

### P2 — Rename Enter-key path bypasses the `busy` re-entrancy guard
`commitRename` (`:280`) has no `if (busy) return` entry guard (unlike `onCreateSnapshot` at `:250`). The Save button is protected via `disabled={busy || ...}` (`:444`), but the rename `<input>` (`:417-428`) is not disabled while busy and its `onKeyDown` fires `commitRename` on Enter (`:424`), so a second Enter during an in-flight rename issues a duplicate PATCH. Low blast radius (rename is ~idempotent), but the guard is asymmetric with the sibling handler. Suggest `if (busy) return` at the top of `commitRename` (and/or disable the input while busy).

### P2 → nit — `refresh()` gates `setLoading(false)` on `isCurrent()`
`:188-190` gates the `loading` reset on `isCurrent()`, which is the exact pattern the `onLoadMore` comment (`:231-238`) argues against for `loadingMore`. In practice `loading` self-heals (the latest `begin()` ticket is always current when its promise settles), so unlike `loadingMore` it does not wedge — the asymmetry is defensible. Optional symmetry cleanup: drop the guard and always `setLoading(false)`.

### nit — Uniform busy guard / rename seed value
- `onDelete` (`:303`) and `onConfirmRestore` (`:326`) also lack the entry `if (busy) return`; their confirm buttons are already `disabled={busy}`, so this is defense-in-depth consistency only.
- `beginRename` (`:463`) seeds the editor from raw `v.label`, which can be empty for a named version even though `displayLabel` shows a synthesized label; `commitRename`'s empty-guard prevents any data hazard. Consider seeding with `displayLabel(v)`.

Spec ✅ (purely additive, in-scope, no over/under-build) AND Code Quality Approved (no P0/P1). The confirmed items are P2/nit robustness and UX-consistency suggestions that surface only when a future adapter mounts the shell — none blocks this phase-1 merge. The P2 restore-error mapping is the highest-value follow-up and is a natural fit for the doc-adapter phase that wires the first end.

## Security note (security-sensitive classification)
- Preview/diff content is injected as `ReactNode` via `renderPreview` / `renderDiff`; there is **no `dangerouslySetInnerHTML` / `innerHTML`** in the panel, so host-supplied preview payloads are React-escaped. XSS surface is clean at the shell level (each adapter's own renderer remains its responsibility).
- Permission gating is client-side defense-in-depth: `canSnapshot` = writer+, `canRestoreVersion` = admin-only, and `api.ts` documents that the backend role-gates every endpoint. The client gate only hides affordances; it is not treated as the trust boundary. Correct.
- No secrets/tokens are handled here (auth header is injected by the shared interceptor). No info leakage in error paths.

Nothing here requires manual human verification beyond the standard `needs-human-review` sign-off already applied.

OctoBoooot (commented):
## Review: #656 — confirming yujiawei 🟡 #2 (rename Enter re-entrancy), byte-verified

**Verdict unchanged: Approve.** Non-blocking; recording one more verified minor in the busy-guard family.

### Minor — `commitRename` Enter path lacks the `busy` re-entrancy guard (confirmed)
`VersionHistoryPanel.tsx`:
- `commitRename` (`:280`) guards the empty label (`if (label === '') return`, `:282`) but has **no `if (busy) return`** — unlike `onCreateSnapshot` (`:250`), which does.
- The rename `<input>` (`:418`) is **not** `disabled={busy}` (only the Save/Cancel buttons at `:444/:449` are). So the DOM disable that protects the button click does **not** protect the input's `onKeyDown` Enter path (`:424`).
- **Repro:** focus the rename input, press Enter twice in quick succession → two `commitRename(seq)` fire before `setBusy(true)` re-renders → two `renameVersion(docId, seq, label)` requests. `busy` is stale-false in the second call's closure and `cancelRename()` hasn't run yet, so nothing short-circuits it.
- **Impact — minor:** a duplicate rename to the *same* label (low blast radius, no data loss/corruption), recoverable, and latent (panel not mounted yet). Same class as the recurring busy-guard misses.
- **Fix:** add `if (busy) return` at the top of `commitRename` (mirroring `onCreateSnapshot`), and/or `disabled={busy}` on the rename input.

yujiawei #1 (restore path lacks incompatible-error distinction) = same non-blocking error-mapping refinement as Jerry-Xin #1; no correctness defect. Credit: yujiawei for #2.

PR #657, opened 2026-07-11
feat(loop): redesign project module with list/card views + webhooks

superalsrk:
## Goal

Re-skin the Loop **项目 (Project)** area to the new UX design and restore the webhook configuration capability that existed in multica but was missing in Loop. The backend contract is unchanged.

## Changes

- **List / card views** on the project list, with a toggle in the toolbar. The preferred view is persisted to `localStorage` (`loop.project.viewMode`).
- **Webhook configuration** added to the right-side project config panel — reuses the multica `/webhook-subscriptions` API (list / create / toggle-enable / delete), including the one-time signing-secret dialog on create.
- **Slimmed project config panel** to name + description + webhook only; removed status / priority / lead editing and the issue drill-down (list still shows status read-only).
- **Description hint text** reused verbatim from multica.
- **Shared `formatRelativeTime` helper** extracted (`ui/time.ts`) and reused by the project + skill lists to remove duplication.

Files: `api/webhookApi.ts`, `api/types.ts` (WebhookSubscription), `panel/ProjectWebhooksSection.tsx`, `panel/ProjectDetailPage.tsx`, `pages/ProjectPage.tsx`, `pages/SkillPage.tsx`, `pages/loop.css`, `ui/time.ts`, i18n `zh-CN.json` / `en-US.json`.

## Verification

- `tsc --noEmit` clean for all changed files.
- Manual: list/card toggle switches and the choice survives reload; clicking a project opens the config panel (no issue drill-down); name/description save; webhook add shows the one-time secret and hits `/fleet/api/v1/webhook-subscriptions`; enable toggle and delete work.

### DIFF
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -447,6 +447,30 @@ export interface UpsertProjectReq {
   lead_id?: string | null;
 }
 
+/* ---------- Webhook（对齐 multica /webhook-subscriptions，后端契约不变） ---------- */
+export interface WebhookSubscription {
+  id: string;
+  workspace_id: string;
+  project_id: string | null;
+  url: string;
+  events: string[];
+  enabled: boolean;
+  secret_hint: string;
+  secret?: string; // 仅创建返回一次，需立即展示给用户保存
+  created_at: string;
+  updated_at: string;
+}
+export interface CreateWebhookReq {
+  url: string;
+  project_id?: string | null;
+  events?: string[];
+}
+export interface UpdateWebhookReq {
+  url?: string;
+  events?: string[];
+  enabled?: boolean;
+}
+
 /* ---------- Agent ---------- */
 export type AgentStatus = "idle" | "working" | "offline" | "error" | string;
 export type AgentVisibility = "workspace" | "private";
--- a/packages/dmloop/src/api/webhookApi.ts
+++ b/packages/dmloop/src/api/webhookApi.ts
@@ -1,0 +1,23 @@
+// @octo/loop — Webhook API（复用 multica /webhook-subscriptions，后端契约不变）
+import type { WebhookSubscription, CreateWebhookReq, UpdateWebhookReq } from "./types";
+import { httpGet, httpPost, httpPatch, httpDelete } from "./http";
+
+export async function listWebhooks(projectId: string): Promise<WebhookSubscription[]> {
+  const data = await httpGet<{ subscriptions: WebhookSubscription[] }>(
+    "/webhook-subscriptions",
+    { project_id: projectId },
+  );
+  return data.subscriptions ?? [];
+}
+
+export function createWebhook(req: CreateWebhookReq): Promise<WebhookSubscription> {
+  return httpPost<WebhookSubscription>("/webhook-subscriptions", req);
+}
+
+export function updateWebhook(id: string, req: UpdateWebhookReq): Promise<WebhookSubscription> {
+  return httpPatch<WebhookSubscription>(`/webhook-subscriptions/${id}`, req);
+}
+
+export function deleteWebhook(id: string): Promise<void> {
+  return httpDelete<void>(`/webhook-subscriptions/${id}`);
+}
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -34,7 +34,8 @@
   "view": {
     "board": "Board",
     "grouped": "Grouped",
-    "list": "List"
+    "list": "List",
+    "card": "Card"
   },
   "scope": {
     "all": "All",
@@ -307,7 +308,32 @@
   },
   "project": {
     "progress": "Progress",
-    "lead": "Lead"
+    "lead": "Lead",
+    "descPlaceholder": "Add a description...",
+    "descHint": "Provided to agents as project context and applied to every task in this project."
+  },
+  "webhook": {
+    "title": "Webhooks",
+    "hint": "Fires on issue status changes. Each request is signed with HMAC-SHA256 in the X-Multica-Signature-256 header.",
+    "urlPlaceholder": "https://example.com/webhooks/multica",
+    "add": "Add",
+    "empty": "No webhooks yet.",
+    "enabled": "Enabled",
+    "delete": "Delete webhook",
+    "deleteConfirm": "Delete this webhook?",
+    "deleteConfirmDesc": "The endpoint will stop receiving events immediately. This cannot be undone.",
+    "secretTitle": "Copy your signing secret",
+    "secretDesc": "This secret is shown only once. Use it to verify the X-Multica-Signature-256 header on incoming requests. Store it somewhere safe now.",
+    "done": "Done",
+    "copy": "Copy",
+    "copied": "Secret copied",
+    "copyFailed": "Failed to copy",
+    "created": "Webhook created",
+    "createFailed": "Failed to create webhook",
+    "updateFailed": "Failed to update webhook",
+    "deleted": "Webhook deleted",
+    "deleteFailed": "Failed to delete webhook",
+    "urlRequired": "Endpoint URL is required"
   },
   "label": {
     "add": "Add labels",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -34,7 +34,8 @@
   "view": {
     "board": "看板",
     "grouped": "分组",
-    "list": "列表"
+    "list": "列表",
+    "card": "卡片"
   },
   "scope": {
     "all": "全部",
@@ -307,7 +308,32 @@
   },
   "project": {
     "progress": "进度",
-    "lead": "负责人"
+    "lead": "负责人",
+    "descPlaceholder": "添加描述...",
+    "descHint": "会作为项目上下文提供给智能体，应用于该项目下的每个任务。"
+  },
+  "webhook": {
+    "title": "Webhook",
+    "hint": "在 issue 状态变化时触发。每个请求都用 HMAC-SHA256 签名，放在 X-Multica-Signature-256 请求头里。",
+    "urlPlaceholder": "https://example.com/webhooks/multica",
+    "add": "添加",
+    "empty": "还没有 webhook。",
+    "enabled": "启用",
+    "delete": "删除 webhook",
+    "deleteConfirm": "删除这个 webhook？",
+    "deleteConfirmDesc": "接收地址会立刻停止收到事件。此操作无法撤销。",
+    "secretTitle": "复制你的签名密钥",
+    "secretDesc": "这个密钥只显示一次。用它来校验收到的请求里的 X-Multica-Signature-256 请求头。现在就妥善保存。",
+    "done": "完成",
+    "copy": "复制",
+    "copied": "密钥已复制",
+    "copyFailed": "复制失败",
+    "created": "webhook 已创建",
+    "createFailed": "创建 webhook 失败",
+    "updateFailed": "更新 webhook 失败",
+    "deleted": "webhook 已删除",
+    "deleteFailed": "删除 webhook 失败",
+    "urlRequired": "请填写接收地址"
   },
   "label": {
     "add": "添加标签",
--- a/packages/dmloop/src/pages/LoopPage.tsx
+++ b/packages/dmloop/src/pages/LoopPage.tsx
@@ -111,6 +111,25 @@ export default function LoopPage() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
+  // 顶部一级导航「Loop」被再次点击时，onMenuClick 会先 routeRight.popToRoot() 清空右栏
+  // （LoopPage 常驻不重挂，useEffect 不会重跑）。这里监听激活事件，把体验对齐「首次进入」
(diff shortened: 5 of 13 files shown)


OctoBoooot (approved):
## Review: redesign project module with list/card views + webhooks (#657)

**Verdict:** Approve
Large PR (889 lines, 13 files) but well-executed; the security-critical surface — the webhook signing-secret handling — follows the correct show-once/never-persist discipline. Reviewed the webhook secret path, the API, and the frontmatter helper in depth; the list/card view + relative-time dedup are low-risk.

Risk tier: normal; size/XL → `needs-human-review` applied (human sign-off before merge). Base `feat/octo-loop`; `mergeStateStatus=UNSTABLE` + no linked issue (`check-sprint`).

### What I verified
- **Webhook secret is handled correctly (the critical piece):** the type models it right — persistent `secret_hint: string` (safe masked hint) + `secret?: string` returned **only once on create**. `ProjectWebhooksSection` holds the full `secret` in transient component state only, shows it in a one-time Modal, clears it on close (`setSecret(null)`), and **never persists it** (no localStorage) or writes it back to the row list; the subsequent `reload()` fetches subscriptions without the secret. The list rows render only `url`/`enabled` — no secret material displayed at all. Copy uses `clipboard.writeText` (no logging). Secret rendered as `<code>{secret}</code>` (escaped). This matches the show-once/never-re-fetchable standard.
- **`webhookApi.ts`:** clean CRUD over `/webhook-subscriptions`, project-scoped, list unwraps `{subscriptions}`. Delete is confirm-gated (`confirmDelete`); add has a `busy` guard; toggle/add/delete all error-handled.
- **Webhook URL XSS-safe:** `row.url` rendered as escaped JSX text (`<span title={row.url}>{row.url}</span>`), not used as an href/src sink (SSRF on the URL is a backend concern; the FE only stores/shows it).
- **`setFrontmatterField` (frontmatter.ts +29):** new bidirectional-sync helper (outer name/desc input ↔ SKILL.md header, content as single source of truth). Reuses the `yamlScalar` escaping I verified on #640 (field values with `:`/`#`/newline safely quoted), preserves other lines + body, first-match-only replace with append-if-absent, handles no-frontmatter. Injection-safe.

### Praise
- The webhook secret is done to the right standard: the *type* itself separates the persistent masked `secret_hint` from the create-only `secret?`, so the full secret can't leak into the cached model — and the component keeps it transient, show-once, cleared-on-close, never persisted, with the list never rendering secret material. That's the correct "one-time secret" discipline enforced at both the data-model and UI layers, not just one of them.

### Note
Given size/XL I scoped deep verification to the webhook/secret surface + the new helpers (the risk) and skimmed the list/card CSS + i18n. `needs-human-review` applied; a maintainer should pass over the breadth.

Jerry-Xin (changes requested):
The PR is relevant to `Mininglamp-OSS/octo-web` and the `dmloop` project area, but one changed skill-detail path introduces a save-blocking regression.

**🔴 Blocking**
- 🔴 Critical: `packages/dmloop/src/panel/SkillDetailPage.tsx:62-67`, `packages/dmloop/src/panel/SkillDetailPage.tsx:95-101`, `packages/dmloop/src/panel/SkillDetailPage.tsx:148-151` now use `SKILL.md` frontmatter as the only source of the skill name. However, `packages/dmloop/src/ui/frontmatter.ts:65-68` returns existing frontmatter unchanged even when it lacks `name`. A skill whose content already has a frontmatter block without `name` will open with an empty name and then fail save validation, even though `row.name` is valid. This can happen through the current create flow when pasted content has partial frontmatter, because creation falls back to the form name while preserving that content. Fix by inserting missing required fields into existing frontmatter, or by falling back to `row.name` and synchronizing it into frontmatter before validation/save.

**💬 Non-blocking**
- 🟡 Warning: `packages/dmloop/src/panel/ProjectWebhooksSection.tsx:20-24` does not catch list failures. A webhook-list API error leaves no visible error state and creates an unhandled promise rejection from the `.then(...).finally(...)` chain. Add a catch path and show a small inline error or retry state.
- 🟡 Warning: `packages/dmloop/src/pages/loop.css:995-1249` adds many raw pixel values and fallback colors. The local project guidance asks UI changes to use design tokens rather than hardcoded spacing/colors. This is not blocking, but it should be tightened for consistency.
- 🔵 Suggestion: `packages/dmloop/src/pages/ProjectPage.tsx:94-145` uses clickable `div` rows/cards without keyboard handlers. Consider buttons or `tabIndex`/Enter handling so the new list/card views remain keyboard-accessible.

**✅ Highlights**
- The project list/card view work is in-scope and uses the existing `dmloop` API/client structure.
- New locale keys were added to both `en-US` and `zh-CN`.
- `formatRelativeTime` removes duplicated list formatting logic cleanly.

Verification note: `git diff --check` passed. I could not run `pnpm i18n:check` or TypeScript locally because this checkout has no installed dependencies (`typescript`/`tsc` missing).

yujiawei (approved):
# Code Review — PR #657 (octo-web)

Reviewed at head SHA `795bcf42bea712aefcffa66f34fcc9ef7af496b3` against base `feat/octo-loop` (merge-base `d140ed42`). 13 files, +780/-109, scoped to `packages/dmloop`.

## Spec compliance

Everything the PR description promises is delivered, and nothing out of scope leaks in:

- **List / card views** with a toolbar toggle, persisted to `localStorage` under `loop.project.viewMode` (`ProjectPage.tsx`) — read/write guarded in try/catch, safe default to `list`. ✅
- **Webhook config** in the project config panel: list / create / toggle-enable / delete via `/webhook-subscriptions`, plus the one-time signing-secret dialog on create (`ProjectWebhooksSection.tsx`, `webhookApi.ts`, `types.ts`). ✅
- **Slimmed config panel** to name + description + webhook; status / priority / lead editing and the issue drill-down are removed, list keeps status read-only (`ProjectDetailPage.tsx`, `ProjectPage.tsx`). ✅
- **Description hint** and **shared `formatRelativeTime`** helper (`ui/time.ts`) reused by the project and skill lists. ✅

The bundled bugfixes (SKILL.md frontmatter normalization in `SkillDetailPage.tsx` / `frontmatter.ts`, and the `LoopPage` nav re-click reset) are outside the "project module" headline but are self-contained, correct, and low-risk.

**Verdict: Spec compliant.**

## Correctness

No blocking (P0/P1) issues found. The one concern worth explicit checking — that the trimmed `updateProject` / `createProject` payloads (`{ title, description }` over `PUT /projects/:id`) might wipe `status` / `priority` / `lead` — does **not** reproduce: the prior code already sent the same partial `{ title, description }` shape, so the backend contract merges partial updates rather than full-replacing. Field preservation is unchanged by this PR.

## Advisory findings (P2, non-blocking)

1. **`SkillDetailPage.tsx:27` — duplicate time helper not removed.** The PR extracts `formatRelativeTime` into `ui/time.ts` and migrates `SkillPage`, but `SkillDetailPage` still carries its own local `formatSkillTime` (used at line ~239). Folding it into the shared helper would complete the stated de-duplication.

2. **`ProjectPage.tsx` / `ProjectDetailPage.tsx` — no in-flight guard on create/save.** `doCreate` and `save` await the request but don't disable the button while it's pending, so a double-click on a slow network can fire duplicate create/update requests. `ProjectWebhooksSection.add` already models the right pattern with its `busy` flag — worth mirroring.

3. **`ProjectPage.tsx` — search moved from server-side to client-side.** `listProjects({ keyword })` became an unfiltered `listProjects()` plus a local `rows.filter(...)`. This is only correct if `/projects` returns the full unpaginated set; if a limit/pagination is ever introduced server-side, search will silently only cover the loaded page. Fine today given the endpoint returns all rows, but a latent coupling to note.

4. **`frontmatter.ts:88` — `setFrontmatterField` key match uses `.trim()`.** The first-match `!replaced` guard plus the flat parser make real corruption unlikely with today's two-field header, but matching keys after trimming leading whitespace is technically fragile if nested/indented YAML is ever introduced. A strict "no leading whitespace" check would harden it.

5. **`SkillDetailPage.tsx:100-102` — name edit re-parses on every keystroke.** `onNameChange` rewrites `content` and re-runs `parseFrontmatter` per keystroke. Negligible for small skills; for large SKILL.md bodies it can add input lag. Local draft state synced on blur would avoid it.

## Security notes (flagged for human verification)

This PR is classified security-sensitive because it (re)introduces webhook management and signing-secret handling. Frontend handling looks correct — the secret is shown once in a modal, held only in component state, copied via `navigator.clipboard`, and never persisted client-side. Two items are backend concerns this diff can't settle and a maintainer should confirm:

- **SSRF on webhook target URLs** — the endpoint accepts an arbitrary `url`; ensure the backend restricts/validates outbound targets (block internal/metadata addresses).
- **Signing-secret exposure** — confirm `secret` is returned by the API on create only, never on list/get.

## Verdict

No blocking issues. The findings above are all P2 maintainability / robustness improvements that can land in a follow-up. Approving; please still get a human eye on the two backend security items given the security-sensitive classification.

yujiawei (changes requested):
# Code Review — PR #657 (octo-web) — Updated Verdict

Correcting my earlier approval. On a second pass I independently confirmed a save-blocking regression in the skill-detail path (also raised by another reviewer). This is a P1 and must be fixed before merge.

Reviewed at head SHA `795bcf42bea712aefcffa66f34fcc9ef7af496b3` against base `feat/octo-loop`.

## Blocking (P1)

**`SkillDetailPage.tsx` name binding + `frontmatter.ts:65-68` — skill can open with an empty name and become unsavable.**

The detail page now treats SKILL.md frontmatter as the *only* source of the skill name:
- `seed()` sets `content = ensureSkillFrontmatter(s.name, s.description, s.content)` (`SkillDetailPage.tsx:66`)
- `nameValue = skillFrontmatter?.name ?? ""` (`SkillDetailPage.tsx:97`)
- `save()` validates `nameValue.trim()` and warns `nameRequired` if empty (`SkillDetailPage.tsx:149-150`)

But `ensureSkillFrontmatter` short-circuits on *any* existing frontmatter block:

```ts
export function ensureSkillFrontmatter(name, description, content) {
  if (hasFrontmatter(content)) return content;   // returns unchanged even if 'name:' is absent
  ...
}
```

So a skill whose stored `content` has a frontmatter block **without** a `name:` key opens with `nameValue === ""`, and any save then fails validation even though `row.name` is perfectly valid — the skill becomes uneditable.

This is reachable through the existing create flow: `SkillPage.tsx:89-94` computes `frontmatterName = frontmatter?.name?.trim() || skillName` and persists `{ name: frontmatterName, content }`. If a user pastes content that already contains a partial frontmatter block (no `name:`), the created skill has a valid DB `name` but a `content` frontmatter with no name — exactly the state that later bricks the detail page.

This is a regression: the previous code bound the name from `row.name` (`setName(s.name)`), which is always valid.

Fix options: have `ensureSkillFrontmatter` / a helper inject missing required fields into an existing frontmatter block, or fall back to `row.name` and sync it into the frontmatter before validation/save.

## Non-blocking (P2)

1. **`ProjectWebhooksSection.tsx:20-24` — `reload()` has no catch.** `listWebhooks(...).then(setRows).finally(...)` swallows nothing: a list failure produces an unhandled rejection and no visible error state. Add a catch with an inline error/retry.
2. **`SkillDetailPage.tsx:27` — duplicate time helper.** The PR extracts `formatRelativeTime` into `ui/time.ts` and migrates `SkillPage`, but `SkillDetailPage` still carries its own local `formatSkillTime` (used ~line 239). Fold it into the shared helper to finish the stated de-duplication.
3. **`ProjectPage.tsx` / `ProjectDetailPage.tsx` — no in-flight guard on create/save.** `doCreate` / `save` don't disable the button while the request is pending, so a double-click on a slow network can fire duplicate requests. `ProjectWebhooksSection.add` already models the right `busy`-flag pattern.
4. **`ProjectPage.tsx` — search moved server-side → client-side.** `listProjects({ keyword })` became unfiltered `listProjects()` + local `rows.filter(...)`. Correct only while `/projects` returns the full unpaginated set; if pagination is ever added, search silently covers only the loaded page.
5. **`frontmatter.ts:88` — `setFrontmatterField` key match uses `.trim()`.** The `!replaced` first-match guard plus the flat parser make corruption unlikely today, but matching keys after trimming leading whitespace is fragile if indented/nested YAML is ever introduced. A strict "no leading whitespace" check would harden it.
6. **`ProjectPage.tsx:94-145` — clickable `div` rows/cards lack keyboard handling.** Consider `tabIndex`/Enter handling (or buttons) so the new list/card views stay keyboard-accessible.
7. **`loop.css` — raw pixel values / hardcoded fallback colors.** Prefer design tokens for spacing/colors per project guidance. Not blocking; tighten for consistency.

## Security notes (security-sensitive PR — for human verification)

Frontend secret handling looks correct: the signing secret is shown once in a modal, held only in component state, copied via `navigator.clipboard`, never persisted client-side. Two backend items this diff can't settle:
- **SSRF on webhook target URLs** — the endpoint accepts an arbitrary `url`; confirm the backend blocks internal/metadata addresses.
- **Secret exposure** — confirm the API returns `secret` on create only, never on list/get.

## Not an issue (checked)

The trimmed `updateProject` / `createProject` payloads (`{ title, description }` over `PUT /projects/:id`) do **not** wipe `status` / `priority` / `lead`: the prior code sent the same partial shape, so the backend merges partial updates. Field preservation is unchanged.

## Verdict

Requesting changes on the P1 save-blocking regression in the skill-detail name binding. The rest are P2 follow-ups. The project list/card + webhook work itself is in-scope and cleanly structured.

PR #658, opened 2026-07-11
feat(loop): redesign AI teammate module + skill frontmatter fixes

superalsrk:
## Goal

Re-skin the dmloop **AI队友 (agent)** module to the new Figma UX while reusing the existing octo-multica agents API (type-only additions, no backend contract changes), and fix a save-blocking skill regression from #657.

### Agent module redesign (`bf0dba89`)
- **List page**: replace the heavy Semi Table with lightweight rows (avatar + status dot, name, inline description, device/runtime, owner, relative time). Adds a 我的 / 全部 / 已归档 scope filter with live counts (single fetch with `include_archived=true`, client-side filtering by `archived_at`; "我的" resolves the current user via `octo_uid`). Archived rows are muted with a restore action (`POST /agents/{id}/restore`); active rows keep the archive confirm.
- **Detail page**: header breadcrumb + status pill + archive; two top tabs 档案 / 配置. 档案 shows 30-day run stats (`/agents/{id}/tasks`) + a GitHub-style contributions calendar (new `GET /agents/{id}/contributions`) with 5 color levels and hover tooltips; 履历 resolves each run's title and opens the issue on click. 配置 sub-nav 指令 / MCP / 技能 provides full-height editors (markdown edit/preview, MCP JSON editor with validation, tri-state save, redacted lock/clear) plus skill management (list/remove + searchable add dialog).
- **Supporting**: add `mcp_config`, `AgentTask`, `AgentContribution`, archived fields to types; add `listAgentTasks`, `getAgentContributions`, `setAgentSkills`, `restoreAgent` APIs; add `isTerminalRun` and `formatDurationMs` helpers; i18n keys for both locales.

### Skill frontmatter fix (`c4de7b5b`)
- `ensureSkillFrontmatter` no longer short-circuits on an existing frontmatter block missing a `name` key — it injects the required name/description from the DB record, fixing an uneditable-skill regression from #657 and keeping the frontmatter title consistent with the DB title.
- Allow underscores in skill names: `SKILL_NAME_PATTERN` widened from `[A-Za-z0-9-]` to `[A-Za-z0-9_-]`, with matching zh-CN/en-US messages.

### Skill list cleanup (`5e6ce366`)
- Drop the per-row source Tag (本地/外部导入) from the skill list — the same info is on the detail page. Removes the unused source color map, `skillSource` import, and per-row `src` computation. Detail source wording clarified (工作区 → 本地, GitHub → 外部导入).

## Verification
[...]

PR #660, opened 2026-07-12
feat(loop): redesign AI squad module + runtime/webhook copy cleanup

superalsrk:
## Goal

Ship the redesigned dmloop AI squad (小队) module with the new UI while keeping
the existing backend contract unchanged, plus small runtime-onboarding and
webhook copy cleanups.

### AI squad module
- **List page**: header with count + New button; 我的/全部 scope tabs with counts;
  filter by leader/creator; sort by name/members/created (asc/desc); rows show
  avatar, name, description, leader chip, member avatar stack (+N), creator and
  created-time.
- **Detail page**: breadcrumb + archive; identity header with inline
  name/description edit and leader/member-count/created meta chips; 成员/指引 tabs.
  Members panel supports add/remove member, inline role edit, set leader, and
  per-member realtime status (working/idle/offline/unstable/archived) with
  active-issue brief and last-active, refreshed on a 30s poll.
- **Create dialog**: name/description/leader picker plus an additional-members
  picker.
- **API**: add getSquadMemberStatus, a partial UpdateSquadReq and the member
  status types.
- **Internal**: hoist shared avatarColor into ui/meta.tsx (reused by the agent
  page); extract an InlineText editor and a memberKey helper to remove
  duplication; squad detail styles live in loop.css.

### Runtime onboarding + webhook copy
- Simplify the "add computer" runtime command to
  `octo-daemon --server-url <current web origin>` (the daemon derives the fleet
  API URL from it); drop the URL helper functions that became unused.
- Rename the webhook signature header referenced in the webhook setup docs to
  `X-Octo-Signature-256` (en + zh).

## Verification
- `tsc --noEmit` passes for the dmloop package; the squad/runtime/meta files are
  clean (only pre-existing import.meta bare-tsc noise unrelated to this change
  remains).
- Manual checks recommended before merge: list scope/filter/sort, create squad,
  detail inline edits, member add/remove/role/set-leader, instructions save, and
  the runtime add-computer command copy. The full app UI was not run in this
  environment, so a manual smoke test is advised.


Jerry-Xin (approved):
The PR is relevant to `octo-web` and the dmloop squad/runtime/webhook surfaces. I found no blocking correctness or security issues.

**💬 Non-blocking**
- 🟡 Warning: The new squad CSS adds several hardcoded colors and measurements despite the project token guidance, especially in [loop.css](/Users/mlamp/.openclaw/tmp/pr-review-66845npy/repo/packages/dmloop/src/pages/loop.css:1623), [loop.css](/Users/mlamp/.openclaw/tmp/pr-review-66845npy/repo/packages/dmloop/src/pages/loop.css:1739), and [loop.css](/Users/mlamp/.openclaw/tmp/pr-review-66845npy/repo/packages/dmloop/src/pages/loop.css:1756). This is most likely to show up as dark-theme drift for the squad avatar, status dots, and leader chip.
- 🟡 Warning: `getSquadMemberStatus` suppresses every API failure and returns an empty list in [squadApi.ts](/Users/mlamp/.openclaw/tmp/pr-review-66845npy/repo/packages/dmloop/src/api/squadApi.ts:64). That keeps the UI resilient, but it also hides permission/backend failures as “no status data,” which makes production diagnosis harder.
- 🔵 Suggestion: The add-member selectors use bare candidate IDs as option values, then recover type with `cands.find((x) => x.id === addPick)` in [SquadDetailPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-66845npy/repo/packages/dmloop/src/panel/SquadDetailPage.tsx:120) and similarly during create flow in [SquadPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-66845npy/repo/packages/dmloop/src/pages/SquadPage.tsx:129). If member and agent ID spaces ever overlap, the UI can add the wrong actor type. Using `type:id` as the select value would match the new `memberKey` helper.
- 🔵 Suggestion: I could not run `pnpm i18n:check` because this checkout has no installed dependencies; the command fails on missing `typescript`. `git diff --check` passed.

**✅ Highlights**
- The PR keeps the squad API enrichment centralized and reuses the directory cache instead of spreading display-name resolution into the UI.
- The detail page’s member status polling is scoped to the mounted panel and cleans up the interval correctly.
- Locale keys for the new visible copy are present in both `en-US` and `zh-CN`.

packages/dmloop/src/panel/SquadDetailPage.tsx
```diff
@@ -1,108 +1,365 @@
...
 
   const back = () => WKApp.routeRight.pop();
-  const save = async () => {
-    if (!name.trim()) { Toast.warning(t("loop.validate.nameRequired")); return; }
-    const next = await updateSquad(squadId, { name: name.trim(), description: desc, instructions: instr });
-    setRow(next); setDirty(false); Toast.success(t("loop.toast.saved")); onChanged?.();
+  const afterMutate = () => { load(); onChanged?.(); };
+
+  const patch = async (p: Parameters<typeof updateSquad>[1]) => {
+    await updateSquad(squadId, p);
```
OctoBoooot: **[blocker]** `patch()` sends a single-field partial to a PUT that requires `name`. `updateSquad`→`httpPut('/squads/${id}', req)` and `UpsertSquadReq.name` is required; `UpdateSquadReq = Partial<...>` only relaxes the client type, not the backend contract (stated unchanged). The agent module's sibling `patch` (`AgentDetailPage.tsx:105`) always re-injects it: `updateAgent(agent.id, { name: agent.name, ...p })`. This one dropped that spread. So `patch({ description })` / `patch({ instructions })` / `patch({ leader_id })` post a nameless body → backend rejects (“保存失败”), or if PUT is full-replace it wipes the omitted fields. Base sent a full object (`{name, description, instructions}`), so this is a behavior change introduced here. Fix: `await updateSquad(squadId, { name: row.name, ...p })` (mirror the agent page), or switch to `httpPatch` only if the backend truly merges.

OctoBoooot (changes requested):
## Review: redesign AI squad module + runtime/webhook copy cleanup (#660)

**Verdict:** Request changes
One blocker: the squad inline-edit `patch()` sends single-field partials to a **PUT** endpoint without the required `name`, diverging from the agent module's own established pattern in this same PR — so description / instructions / set-leader edits either fail or wipe fields. Everything else (poll lifecycle, i18n parity, XSS surface, additive types) is clean.

Risk tier: **high** (API surface: `agentApi`/`squadApi`/`types`/`webhookApi`). `needs-human-review` applied (size/XL, 14 files, +1148/-177). Base `feat/octo-loop`. CI green (`code-review` pass; `satisfy-code-review` skipping). `mergeStateStatus` stays gated on `needs-human-review` regardless of votes.

### Blocker
- `panel/SquadDetailPage.tsx:81-83` — `patch()` posts the raw partial to a full-replace PUT, omitting the required `name`.
  ```
  const patch = async (p) => { await updateSquad(squadId, p); afterMutate(); };
  // callers: patch({ description }), patch({ instructions }), patch({ leader_id })  ← no name
  ```
  `updateSquad` → `httpPut('/squads/${id}', req)` (`squadApi.ts:51`), and `UpsertSquadReq.name` is **required** (`types.ts:588`). The new `UpdateSquadReq = Partial<UpsertSquadReq>` (`types.ts:594`) only relaxes the *client* type — the backend contract is stated unchanged.

  **The proof this is wrong is in this very PR:** the agent module's sibling `patch` (`panel/AgentDetailPage.tsx:103-105`) does
  ```
  const next = await updateAgent(agent.id, { name: agent.name, ...p });  // always re-sends name
  ```
  It deliberately re-injects `name` on every partial PUT. The squad page copied the `patch` idiom but dropped that spread. Same HTTP verb, same "contract unchanged" backend, opposite behavior.

  **Failing inputs (open a squad detail page):**
  - Edit description → `patch({ description })` → PUT body has no `name` → backend rejects (name required) → the edit throws → caught → "保存失败" toast. Description/instructions/set-leader are the headline features of the new detail page and are broken.
  - If instead the PUT is full-replace (as the base code assumed — see below), `patch({ leader_id })` wipes `name`/`description`/`instructions`. Silent data loss on a high-tier path.

  Either way it's a defect; the sibling's `name`-spread shows which model the backend actually uses.

  **This is a behavior change, not pre-existing:** base `SquadDetailPage` sent a **full** object — `updateSquad(squadId, { name: name.trim(), description: desc, instructions: instr })` (base `:38`). This PR is what introduces the nameless partials.

  **Fix (one line, mirror the sibling):**
  ```
  const patch = async (p) => {
    if (!row) return;
    await updateSquad(squadId, { name: row.name, ...p });
    afterMutate();
  };
  ```
  (Or, if the backend genuinely supports partial merge, switch `updateSquad` to `httpPatch` — but that's a backend-contract claim; the agent module's `name`-spread strongly implies it does not.)

### What I verified clean
- **Member-status poll lifecycle correct** — `setInterval(…, 30000)` with `clearInterval` in the cleanup and deps `[squadId]` (`:67-70`); re-subscribes on squad change, no leak. The fetch has `.catch(() => undefined)` so a transient failure doesn't clear the list mid-poll. `getSquadMemberStatus` has `?? []` + `.catch(() => [])` fallbacks (`squadApi.ts:64`).
- **i18n parity** — dmloop `en-US.json` / `zh-CN.json` both 446 keys, zero divergence. Webhook header rename `X-Octo-Signature-256` applied in both locales.
- **API changes additive & compatible** — `getSquadMemberStatus` is new; `SquadMemberStatus*` types are new; `updateSquadMemberRole` correctly uses `httpPatch` (partial semantics where the verb matches). `leader_avatar` now tries `agent` then falls back to `member` (`?? actorAvatar(...,"member",...)`) — safe widening for agent-led squads.
- **No XSS** — no `dangerouslySetInnerHTML` in any changed file; names/descriptions render as escaped JSX.
- **`member` role inline edit** uses `allowEmpty` + a change-guard (`if ((m.role ?? "") === role.trim()) return`) — no empty-role churn.

### Praise
- The agent module already solved the partial-PUT-needs-`name` problem correctly (`{ name: agent.name, ...p }`); the redesign reuses `avatarColor`/`InlineText`/`memberKey` cleanly and keeps `updateSquadMemberRole` on PATCH where partial semantics are real. Aligning squad to that same `name`-spread is all this blocker needs.

### Note
The PR body states the UI was not run in-environment and a manual smoke test is advised — the inline-edit path above is exactly what that smoke test would surface. `needs-human-review` is on; a maintainer should confirm the backend PUT semantics when clearing it.

yujiawei (approved):
# Code Review — PR #660 (octo-web)

Independent review of the dmloop AI-squad module redesign plus the webhook/runtime copy cleanup. I read the full diff and cross-checked the load-bearing items against the actual backend contract. This is a well-executed frontend change; the backend contract is respected end to end. No blocking issues. Findings below are all advisory (P2) — none should hold up merge.

## Spec compliance

**PASS.** The diff implements the redesign coherently and stays within contract:

- **Squad list**: header count, mine/all scope tabs with counts, leader/creator filter, name/members/created sort — all client-side over the existing `GET /squads`.
- **Squad detail**: breadcrumb + archive, inline name/description edit, leader/member/created chips, Members/Instructions tabs, per-member realtime status with a 30s poll, add/remove/role/set-leader.
- **API layer**: `getSquadMemberStatus`, `UpdateSquadReq = Partial<UpsertSquadReq>`, and the member-status types.

Two contract points I verified because they looked risky and are actually correct:

1. **Partial `updateSquad` via PUT is safe.** The new inline editors send single-field bodies (`{name}`, `{description}`, `{instructions}`, `{leader_id}`). The backend `PUT /squads/:id` handler decodes pointer fields and the SQL uses `COALESCE($n, col)`, so unspecified fields are preserved. No data-blanking.
2. **Member-status JSON shape matches the TS types exactly** (`member_type`, `member_id`, `status`, `active_issues[]{issue_id,identifier,title,issue_status}`, `last_active_at`, wrapped in `{members}`). Human members correctly carry `status=null` and are excluded from the status UI.

i18n en/zh parity is complete for the new keys, and every referenced key resolves. The RuntimePage dead-code removal (`appUrl`, `fleetApiUrl`, `withOrigin`, `isAbsoluteUrl`, `trimEndSlash`, `LOOP_API_BASE` import) leaves no dangling references; the remaining `originOf`/`currentOrigin`/`envValue` still compile and back the simplified single-line onboarding command.

## Code quality

**Approved.** The InlineText commit guard (the `committed` ref that dedups the enter-then-blur double fire) is correct, the 30s poll interval is keyed on `squadId` and clears on cleanup (no leak, no first-frame double fetch), and `avatarColor` is cleanly hoisted to `ui/meta` and reused.

Advisory findings (non-blocking):

- **P2 — Squad list lost name search with no replacement.** `SquadPage.tsx` dropped the old keyword `Input`; the new toolbar offers scope/filter/sort but no name search, and `listSquads` is now always called with no keyword. On a workspace with many squads, name is the most common lookup, and every sibling list page (agents, projects, skills, issues) keeps a search box. Consider re-adding a client-side name filter to the toolbar.
- **P2 — Member mutations trigger a full refetch.** `afterMutate` calls `load()`, which re-fetches both `getSquad` and `getSquadMemberStatus` on every role change / add / remove / set-leader. Works, but causes visible flicker with many members. An optimistic local update or status-only refetch would smooth it.
- **P2 — Dead first operand in `leader_avatar` fallback.** `squadApi.ts:25` does `actorAvatar(dir, "agent", …) ?? actorAvatar(dir, "member", …)`, but `actorAvatar` returns `undefined` for any non-`member` type, so the agent term is always a no-op. Behavior is correct (falls through to the member term), but it reads like a real agent-first fallback and is misleading next to `leader_name`, whose agent-first branch is genuinely effective. Drop the dead term or teach `actorAvatar` to resolve agents.
- **P2 — `getSquadMemberStatus` swallows all errors to `[]`.** Auth/5xx/network failures collapse to an empty list, so every agent renders an "offline" dot indistinguishable from genuinely offline; the caller-side `.catch` is also redundant since the API layer never rejects. The 30s poll makes this an acceptable degradation, but there's zero error observability.
- **P2 — "Archive" has no restore path.** The list/detail "Archive" actions call `deleteSquad` → `DELETE /squads/:id`. The backend soft-archives (sets `archived_at`, and `listSquads` filters it out), so the copy is accurate, but unlike agents (which have an archived scope + restore) there is no way to view or restore an archived squad from the UI. Consider adding a restore affordance for parity.
- **P2 — Accessibility on the list row.** The row-level archive button is `opacity:0; pointer-events:none` until `:hover`, with no `:focus-within` and no coarse-pointer fallback — so it's unreachable on touch (tap falls through to open-detail) and an invisible focus target for keyboard. The same file already uses `:focus-within` for the skill-list delete and for squad member actions, so this is inconsistent with the codebase's own pattern. Also, the row is `role="button"` with a nested `<button>` (invalid ARIA nesting) and only handles Enter, not Space.
- **P2 — Sort dropdown doesn't close on select.** The sort `Dropdown` omits `clickToHide`; single-select menus elsewhere in the module (AssigneePicker, LoopPage, IssueDetailPage) pass it. The filter dropdown correctly stays open (multi-select), but sort should close after picking a field.
- **P2 — Create/add-member pickers key candidates by bare id.** The multi-select stores `c.id` and re-resolves type via `cands.find(x => x.id === id)`, collapsing the `(type, id)` pair to `id` — inconsistent with the module's own `memberKey` (`type:id`) discipline. Harmless in practice (agent/member UUIDs never collide), purely a consistency nit.

## Verification note

Two independent external advisory passes were attempted for this complex PR but were unavailable this run (one failed on a gateway tool-config error, the other timed out); their absence is recorded, not treated as tacit approval. The findings above come from a multi-lens automated review with adversarial per-finding verification against the source, plus direct backend-contract checks.

## Verdict

Approve. Spec-compliant, backend contract respected, no correctness/security/build issues. The advisory items are quality-of-life improvements — the name-search removal is the most worth a follow-up, but none block merge.

OctoBoooot (approved):
## Review: redesign AI squad module + runtime/webhook copy cleanup (#660) — delta re-review

The new commit (`4a54429d`, "re-send name on squad partial update and address PR review") fixes the blocker I raised and resolves every peer 🟡. Byte-verified at the new head. My data-loss/edit-fail blocker was **correct** — the author's fix comment confirms it.

Risk tier: high (API surface). `needs-human-review` stands (size/XL). Base `feat/octo-loop`; `mergeStateStatus=CLEAN` but `reviewDecision` will clear once this supersedes my stale CR.

### Resolved — the nameless-partial-PUT blocker (confirmed real, now fixed)
`SquadDetailPage.tsx:81-86`:
```
// PUT /squads/:id 是全量替换、name 必填；局部编辑需带上当前 name（对齐 AgentDetailPage）。
const patch = async (p) => {
  if (!row) return;
  await updateSquad(squadId, { name: row.name, ...p });
  afterMutate();
};
```
This is exactly the fix I proposed (mirror the agent page's `{ name, ...p }`). Notably, **the author's own comment states the squad `PUT /squads/:id` is full-replace with `name` required** — so omitting `name` on `patch({ description })` / `patch({ leader_id })` would indeed have wiped fields. The blocker was a real data-loss defect, now closed. (The `if (!row) return` guard also correctly prevents a patch before the squad has loaded.)

### Resolved — peer 🟡s addressed in the same commit
- **Error visibility** (`squadApi.ts:64`): `getSquadMemberStatus` no longer silently swallows — it `console.warn`s the failure before returning `[]`, so a permission/backend fault is diagnosable while staying non-blocking for the soft status enhancement.
- **Member id-collision** (`SquadPage.tsx:311`, `SquadDetailPage.tsx:310`): add-member `Select.Option` values switched from bare `c.id` to composite `` `${c.type}:${c.id}` ``, and the pick lookups match on the composite — so an agent and a member sharing an id can no longer collide, and leader-exclusion is now type-scoped (`!(c.type === "agent" && c.id === nLeader)`).
- **Dark-mode drift** (`loop.css`): hardcoded hexes → `var(--semi-color-*, #fallback)` tokens (primary/success/warning + light variants), so the new squad chips/dots follow the theme.

### Calibration note (mine)
I briefly folded on my own blocker mid-thread after a peer cited `types.ts:200` ("省略=不动" merge semantics) as proof the PUT was safe. That comment documents the **Issue** endpoint, not squads — a sibling endpoint's contract is not proof of this one's. The author's endpoint-specific comment ("全量替换、name 必填") confirms the squad PUT is full-replace, vindicating the original blocker. Lesson: a documented contract on endpoint A doesn't settle endpoint B; the authoritative signal here was the author, who has the backend.

### Praise
- The fix doesn't just patch the one call site — it documents *why* (`name` required because PUT is full-replace) so the next editor won't re-introduce it, and it cleanly resolved all three reviewer 🟡s (error visibility, id-collision, theming) in one focused commit. The `type:id` composite for the member picker is the right fix for the agent/member id-space overlap, applied consistently to both list and detail pages.

### Note
`needs-human-review` remains (high tier + XL); a maintainer should still smoke-test the inline-edit + member flows per the PR body (UI was not run in-env) before merge.

Jerry-Xin (approved):
Summary: The PR is in-scope for `octo-web`, but the squad detail update path can lose existing squad fields.

**🔴 Blocking**
- 🔴 Critical: Partial squad edits are sent through a full-replacement `PUT`, which can clear omitted fields. `updateSquad` still calls `httpPut` in `packages/dmloop/src/api/squadApi.ts:51`, while `SquadDetailPage` explicitly treats the endpoint as full replacement but only sends `{ name: row.name, ...p }` in `packages/dmloop/src/panel/SquadDetailPage.tsx:81`. As a result, `saveDesc`, `saveInstr`, and `setLeader` in `packages/dmloop/src/panel/SquadDetailPage.tsx:94`, `packages/dmloop/src/panel/SquadDetailPage.tsx:101`, and `packages/dmloop/src/panel/SquadDetailPage.tsx:141` can omit `description`, `instructions`, or `leader_id` and let the backend reset them. Fix by either using a real `PATCH` endpoint if supported, or by sending the complete current squad payload on every `PUT`: `name`, `description`, `instructions`, and `leader_id`.

**💬 Non-blocking**
- 🟡 Warning: The “Mine” squad scope currently means “created by me” only. Both the count and filter use `s.creator_id === myMemberId` in `packages/dmloop/src/pages/SquadPage.tsx:74` and `packages/dmloop/src/pages/SquadPage.tsx:79`. If product expects “my squads” to include squads where I am leader or member, this will hide relevant squads. If creator-only is intended, consider making the label more explicit.

**✅ Highlights**
- The `type:id` member picker value fixes agent/member ID collisions in `packages/dmloop/src/pages/SquadPage.tsx:127` and `packages/dmloop/src/panel/SquadDetailPage.tsx:122`.
- The status polling effect is cleaned up correctly with `clearInterval` in `packages/dmloop/src/panel/SquadDetailPage.tsx:67`.
- I could not run the local TypeScript check because `tsc` is not installed/resolved in this checkout.

OctoBoooot (commented):
## Review: #660 — addressing Jerry-Xin 🔴 (partial-PUT omitted fields), byte-verified

**Verdict unchanged: Approve.** The 🔴 is literally accurate about what goes over the wire, but the concern (omitted fields wiped) is not supported once you read the fix's own logic. Documenting the analysis; not flipping.

### The claim, byte-verified (head `4a54429d`)
`patch = updateSquad(squadId, { name: row.name, ...p })`, and each call site sends one field + name:
- `patch({ description })` → `{ name, description }` (omits `instructions`, `leader_id`)
- `patch({ instructions })` → `{ name, instructions }` (omits `description`, `leader_id`)
- `patch({ leader_id })` → `{ name, leader_id }` (omits `description`, `instructions`)

So yes — every partial edit omits the other fields. The question is whether the backend PUT **wipes** omitted fields or **leaves them unchanged**.

### Why this is presence-merge (omitted = unchanged), not wipe
The decisive evidence is the fix itself: it re-injects **only `name`**, and its comment says *"局部编辑需带上当前 name"* — carry the current **name** only. **If the PUT wiped every omitted field, re-injecting only `name` would be a broken fix** — it would still blank `description`/`instructions`/`leader_id` on 3 of the 4 edits. A backend-aware author (who wrote `对齐 AgentDetailPage` and the contract comment) would have spread all four fields. That the operational rule is "carry only name" means the real contract is: **`name` is required-present (validation), all other fields presence-merge** — matching the documented `types.ts:200` update-family semantics (`省略=不动`). Corroborated by the production agent sibling: `AgentDetailPage.patch` does the identical single-field-plus-name PUT (#658, shipped) with no field-wipe reports.

Under that reading the fix is **complete** and no field is lost. The 🔴 reads the colloquial "全量替换" literally; the code's own logic contradicts the wipe interpretation.

### Honest caveat
Frontend-only repo — I cannot read the squad PUT handler. The conclusion rests on (a) the only-name fix being self-consistent solely under presence-merge, (b) the shipped agent sibling, (c) `types.ts:200`. Strong circumstantial, not backend source. The one definitive check is the octo-server squad PUT handler — and `needs-human-review` (already on this PR) routes a maintainer to exactly that before merge. If the handler turns out to be true wipe-on-omit, the fix is insufficient (all fields would need re-sending) — flagging so the human reviewer verifies this specific line.

Jerry-Xin's 🟡 ("我的" = creator-only scope may not match product intent) is a fair product-labeling nit, non-blocking. Credit: Jerry-Xin for pressing the exact path — the omitted fields are real; the wipe is not, on the available evidence.

PR #663, opened 2026-07-12
feat(loop): add automation module and shared control design system

superalsrk:
## Goal
Add the **Automation (autopilot)** sub-module to dmloop and unify the loop
form-control look so Semi controls stop clashing with the self-built loop style.
Backend contract is identical to autopilot — no backend changes.

## Changes
**Automation module**
- Secondary nav: standalone "My loops" + a "Workspace" group (Loop / Project /
  Automation / AI teammate / AI squad), with Automation between Project and AI
  teammate; "My loops" reuses the issue view filtered to the current member.
- `api/autopilotApi.ts` over `/fleet/api/v1/autopilots*` (list/get/create/update/
  delete, manual trigger, trigger create/update/delete, run history) + directory
  enrichment; types in `api/types.ts`.
- List: card grid with enable toggle, manual run and delete; clicking a card opens
  a detail page (not a modal).
- Create dialog (Figma-simplified): name / executor (agent|squad) / target project
  / schedule (once·daily·weekly·monthly + time) / task instructions; schedule
  triggers only, execution_mode fixed to create_issue.
- Detail page (`panel/AutopilotDetailPage.tsx`): two columns — task-instructions
  auto-grow editor with Save, a triggers list (add/edit/enable-disable/delete), and
  run history (list + total, run rows with an issue link through to the loop); right
  properties panel with inline-editable name/executor/project, status toggle,
  next-run, Run-now (confirm) and Delete. Every mutation toasts on success/failure.

**Loop control design system**
- Shared skin in `ui/loopControls.css`: native `loop-field`/`loop-field-textarea`,
  `loop-seg`, `loop-search`, and a `:is(.loop-page,.loop-sd,.loop-fields,.loop-modal,
  .loop-apd)` layer that skins Semi Select/Input/DatePicker/TimePicker/InputNumber
  (white bg, radius, hover, focus) — fixes the grey Select trigger.
- New reusable `ui/AutoGrowTextarea.tsx`, `ui/ScheduleFields.tsx`, `ui/autopilotSchedule.ts`.
- Rolled the kit across all dmloop create/edit dialogs and detail panels.
- Scope switch becomes a text tab; squad member list becomes a light table with a
  right-side kebab menu + Leader/Member badge and an outline add-member button.

## Verification
- `npx tsc --noEmit --skipLibCheck -p packages/dmloop/tsconfig.json` — clean.
- `npx stylelint "packages/dmloop/src/**/*.css" --config stylelint.config.mjs` — pass.
- i18n keys added in both zh-CN and en-US.
- Manual dev pass recommended: card → detail, enable toggle, add/edit/enable/delete
  triggers, run-now confirm, run-row → loop detail, inline edits (toast 已保存),
  and consistent white-bg controls across pages/dialogs/detail.


Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-web`, but the new automation scheduler has a blocking semantics bug.

**🔴 Blocking**
- 🔴 Critical: `Once` automations are serialized as recurring yearly cron jobs. In [autopilotSchedule.ts](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/ui/autopilotSchedule.ts:53), `frequency: "once"` becomes `min hour day month *`, and [parseCron](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/ui/autopilotSchedule.ts:95) treats the same shape as `once`. That means “Once 2026-07-20 09:00” becomes `0 9 20 7 *` and will run every July 20. The create/edit flows send that cron directly to the backend in [CreateAutomationModal.tsx](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/ui/CreateAutomationModal.tsx:69) and [TriggerEditModal.tsx](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/ui/TriggerEditModal.tsx:43). Either use a real one-shot backend contract, automatically disable/delete after first fire, or remove/rename the option so users are not promised one-time execution.

**💬 Non-blocking**
- 🟡 Warning: The shared control CSS violates the project’s stated styling constraints by using `!important` and direct Semi class overrides in [loopControls.css](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/ui/loopControls.css:132), especially [line 148](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/ui/loopControls.css:148). If this skinning layer is intentional, it should avoid `!important` or be explicitly carved out in project conventions.
- 🟡 Warning: The detail page silently converts run-history load failures into an empty history in [AutopilotDetailPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-y2zh_c24/repo/packages/dmloop/src/panel/AutopilotDetailPage.tsx:88). That makes API failures look like “No runs yet”; consider showing a toast or inline error state.

**✅ Highlights**
- The API surface is narrowly scoped under `/autopilots`.
- Directory enrichment reuses the existing cache pattern.
- The detail page keeps unsaved task-instruction drafts across reloads, which avoids clobbering local edits.

OctoBoooot (approved):
## Review: add automation module + shared control design system (#663)

**Verdict:** Approve
Large but clean: a new autopilot/automation module plus a loop form-control design system. Read the schedule/cron logic and API surface directly, fanned out over the UI + types + i18n, and adversarially verified every candidate finding — no blocker or major survives. One non-blocking minor.

Risk tier: **high** (API surface: `autopilotApi`/`directory`/`types`). `needs-human-review` applied (32 files, +2176/-173). Base `feat/octo-loop`; CI green (`code-review` pending, not failing). Backend contract identical to octo-loop's autopilot — no backend change.

### What I verified (core logic, by hand)
- **Schedule round-trips correctly** (`autopilotSchedule.ts`). `toCron`/`parseCron` are inverse for daily/weekly/monthly; `once` is a documented approximation (5-field cron can't express a true one-shot — pins day+month, loses year, stated in the header comment). Parse falls back to `daily` on anything unrecognized. Time split is null-safe.
- **Timezone is wired, not dropped.** `toCron` emits a bare 5-field cron, but every trigger create/update sends `timezone: cfg.timezone` alongside `cron_expression` (`TriggerEditModal:45,47`, `CreateAutomationModal:71-72`), and the type carries `timezone`. So the backend interprets the cron in the user's IANA zone — no wrong-hour firing.
- **Partial updates use PATCH, not PUT.** `updateAutopilot`/`updateAutopilotTrigger` are `httpPatch` with all-optional `Update*Request` types. This is the correct verb for partial bodies — the #660 nameless-PUT field-wipe class does **not** apply here. Good.
- **i18n parity holds** — 60 `automation.*` keys present and identical in both `en-US` and `zh-CN`; all referenced summary/weekday keys resolve.
- **No null-deref in enrichment** — `projectNameOf(id)` guards `if (!id) return null` before the Map lookup; `enrich` maps over `?? []` fallbacks.

### Findings adversarially verified and DROPPED (documented for transparency)
A fan-out flagged these; each is a false positive on the bytes:
- **"Blocker: autopilot mutations miss `afterDirectoryMutation`"** — refuted. The directory caches only members/agents/squads/projects (`directory.ts:32-35`); autopilots are not a directory-cached entity. Create/update/delete of an autopilot changes no cached actor/project data, so invalidating the directory would only force a needless re-fetch. `projectApi` uses the wrapper because a project mutation *does* change `/projects`; autopilot mutations don't. Enrichment *reads* the directory, it doesn't write entities it holds.
- **"RunConfirmModal note not reset per target"** — refuted. The reset effect is `useEffect(..., [pending])` (`:140`) and `visible = !!pending`; a new confirm target is a new `pending`, so `setNote("")` fires each time.
- **"CreateAutomationModal orphan on trigger failure"** — refuted. The inner try/catch (`:68-78`) intentionally handles autopilot-created-but-schedule-failed: it toasts the reason and lets the user add the schedule from the detail page (autopilot with zero triggers is a valid state). Documented partial-success path, not a defect.

### Minor (non-blocking)
- `AutomationPage.runNow` (`:48`) has no in-flight guard, so a manual trigger could double-fire — but it's behind a `Dropdown.Item` (menu closes on click) and a duplicate manual run is low-harm. A `busy` flag would be belt-and-suspenders; not required.

### Praise
- The schedule layer is the load-bearing correctness surface and it's handled with care: `timezone` travels with every cron write (so wall-clock intent survives), the lossy `once`→cron approximation is documented rather than hidden, and `parseCron` degrades safely to `daily` instead of throwing. Using **PATCH** with all-optional update types is exactly right for partial edits — it sidesteps the full-replace field-wipe trap that bit the squad module (#660). Enrichment mirrors the existing `projectApi` directory-cache pattern without over-applying it to entities the directory doesn't hold.

### Note
Given size/XL, deep verification was scoped to the schedule logic, API/PATCH semantics, timezone wiring, and the directory/i18n surface (the risk); the CSS design system and page restyles were skimmed. `needs-human-review` applied — a maintainer should do the recommended manual pass (card→detail, trigger add/edit/enable/delete, run-now confirm, inline edits) and the visual control-consistency check before merge.

packages/dmloop/src/ui/autopilotSchedule.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+export function toCron(cfg: ScheduleConfig): string {
+  const [hour, min] = splitTime(cfg.time);
+  switch (cfg.frequency) {
+    case "once": {
+      const [, mon, day] = cfg.date.split("-").map((s) => parseInt(s, 10));
+      const D = Number.isFinite(day) ? day : 1;
+      const M = Number.isFinite(mon) ? mon : 1;
+      return `${min} ${hour} ${D} ${M} *`;
```
OctoBoooot: **[blocker]** "Once" is serialized as a yearly-recurring cron. `Once / 2026-07-20 09:00` → `0 9 20 7 *`, which fires 09:00 every July 20, every year — not once. There is no one-shot mechanism to save it: the trigger contract (`CreateAutopilotTriggerRequest = {kind, cron_expression, timezone, label}`, `types.ts:731`) has `kind:"schedule"` and NO `run_once`/`expires_at`/`end_at` field, so nothing tells the backend to fire-once-and-disable. Reachable trivially (pick Once + a date), and each fire has real side effects (`execution_mode: create_issue` → issue creation / agent dispatch). The header comment documents the limitation, but the UI still offers "Once" as a one-shot frequency — a source comment doesn't reach the user. Fix: drop "Once" until the backend supports a one-shot trigger, or send the flag it needs (not in current types), or at minimum relabel + in-UI warn that Once recurs annually (stopgap only).

OctoBoooot (changes requested):
## Review: #663 — correcting my verdict (Approve → Request changes)

I under-graded a real, reachable defect Jerry-Xin correctly flagged: the **"Once" schedule frequency is serialized as a yearly-recurring cron with no one-shot mechanism** — so a "run once" automation actually re-fires every year, forever, triggering real automated side effects (issue creation / agent dispatch). I saw this code and rationalized it as a "documented approximation." That was wrong. Flipping to REQUEST_CHANGES. Byte-verified.

Risk tier: high; `needs-human-review` stands. Base `feat/octo-loop`.

### Blocker (I missed this; Jerry-Xin caught it)
`autopilotSchedule.ts:53` — `toCron` for `frequency: "once"` returns `` `${min} ${hour} ${D} ${M} *` `` — e.g. Once / 2026-07-20 09:00 → `0 9 20 7 *`. A 5-field cron has no year field, so `20 7` with `*` in day-of-week means **"09:00 every July 20, every year."**

**Why it's a real blocker, not a documented tradeoff:**
- **No backend one-shot discriminator exists.** I checked the full trigger contract: `CreateAutopilotTriggerRequest = { kind, cron_expression, timezone, label }` (`types.ts:731`) and `AutopilotTrigger` (`:678`) — `kind` is `"schedule"` (recurring) and there is **no `run_once` / `expires_at` / `end_at` field anywhere**. So the "once" intent is carried *only* by the cron form, which is recurring. Nothing tells the backend to fire-once-and-disable.
- **Reachable trivially:** pick "Once" + any date in the create dialog or trigger editor → recurring annual automation.
- **Real side effects:** each fire creates an issue and/or dispatches an agent/squad (`execution_mode: create_issue`). A user who scheduled a one-time task gets a silent yearly recurrence.
- **In scope:** `autopilotSchedule.ts` is net-new in this PR.
- The author *documented* the limitation in the file header (`单次…无法真正一次性触发`), but the UI still offers "Once" as a selectable frequency and labels it as one-shot. A source comment doesn't reach the user; the UI promises one-shot and delivers annual recurrence.

**Fix directions (author's design call):** either (a) don't offer "Once" until the backend supports a one-shot trigger (`run_once`/`expires_at`); or (b) if a fire-once-then-disable path exists server-side, send whatever flag it needs (not present in the current types); or (c) at minimum, relabel and document in-UI that "Once" recurs annually — though (c) is a stopgap, not a real one-shot.

### Also confirmed (Jerry-Xin's 🟡s — both real, non-blocking)
- **Run-history load failure shows empty, not error** (`AutopilotDetailPage.tsx:88`): `listAutopilotRuns(...).catch(() => ({ runs: [], total: 0 }))` swallows a fetch failure into an empty list → the empty-state (`:268`) renders, misleading the user into thinking there's no history. Minor (read path, recoverable). Consider surfacing the error (same class as the #660 status-swallow).
- **CSS `!important` + Semi-class overrides** (`loopControls.css`, ×1 `!important` + `:is(...)` Semi overrides): a style-convention smell for a design-system layer that has to beat Semi specificity. Minor/nit.

### Accountability
My APPROVE was wrong on the headline item. I read `toCron`'s once-branch and the header comment and treated "the author knows it's approximate" as sufficient — but an approximation that silently converts "run once" into "run every year" with automated side effects is a functional defect, not a footnote. The lesson (which I've hit before): a documented limitation in code is not consent from the user, and "author is aware" ≠ "acceptable to ship." Credit: Jerry-Xin. The rest of my review stands (timezone wired, PATCH-partial correct, i18n parity) — but the verdict was miscalibrated and is now corrected.

superalsrk:
Thanks @Jerry-Xin — addressed in 6c61b722.

**🔴 "Once" recurring-yearly semantics — fixed by removing the option.**
The backend contract (same as) only supports cron schedule triggers, with no one-shot primitive, and itself has no "Once" frequency. Rather than ship a promise we can't keep, I removed "Once" entirely — from the `Frequency` type, `toCron`/`parseCron`, the schedule summary, the `ScheduleFields` UI and i18n (both locales). Frequencies are now daily/weekly/monthly, all of which map to correct recurring cron. (A true one-shot would need a new backend contract; out of scope here.)

**🟡 Silent run-history failure — fixed.**
`AutopilotDetailPage` no longer collapses a failed runs fetch into "No runs yet"; it tracks a `runsError` state and renders an inline error with a Retry action.

**🟡 `!important` / Semi overrides — documented as a scoped carve-out.**
This skin layer is intentional: it overrides `.semi-*` only under loop roots (`:is(.loop-page,.loop-sd,.loop-fields,.loop-modal,.loop-apd)`), never globally, and the few `!important`s exist solely to beat Semi's focus-state specificity. I added an explicit exception to the CSS conventions in `CLAUDE.md` (and it's captured in the `octo-loop-design-spec` skill). Happy to convert the focus rules to higher-specificity selectors instead of `!important` if you'd prefer that over the documented carve-out.

Verification: `tsc --noEmit` clean, `stylelint` pass, i18n parity kept.

superalsrk:
Correction on the 🟡 skin point: I reverted the CLAUDE.md edit. The loop control-skin carve-out (loop-scoped `.semi-*` overrides + limited `!important` for focus specificity) is documented in the `loopControls.css` header comment and the `octo-loop-design-spec` skill instead. Still happy to convert the focus rules to higher-specificity selectors to drop `!important` if preferred.

yujiawei (approved):
# Code Review — PR #663 (octo-web)

Reviewed at head `abb53e1d` against base `feat/octo-loop`. Frontend-only change in `packages/dmloop` (Semi UI): adds the Automation (autopilot) module and a shared loop control design system. Backend contract unchanged.

Build verification (deps installed, run locally):
- `npx tsc --noEmit --skipLibCheck -p packages/dmloop/tsconfig.json` → **0 errors in dmloop** (pre-existing React 17 type errors in the unrelated `dmworkbase` package are not touched by this PR).
- `npx stylelint "packages/dmloop/src/**/*.css"` → **0 errors** (443 hex-color warnings are the file-wide existing convention).
- i18n: every referenced key resolves; zh-CN / en-US are at full key parity.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. Everything the PR description lists is present — `autopilotApi.ts` over `/fleet/api/v1/autopilots*`, the list card grid (enable toggle / run / delete), the create dialog (name / executor / project / schedule / task instructions), the detail page (task-instructions editor, triggers CRUD, run history with issue links, inline-editable properties), and the shared `loopControls.css` skin rolled across dialogs and detail panels.
- **Over-building:** none. `execution_mode` is fixed to `create_issue` and the schedule config only emits triggers, matching the stated scope. No unrequested flags or endpoints.
- **Deviation:** none that blocks. The executor picker correctly constrains to `agent`/`squad` and hides "unassigned" (`AssigneePicker.tsx` `types`/`allowUnassigned`). See the two truthfulness points below — surfaced as quality suggestions, not spec breaks.

## 2. Code Quality

**Quality: Approved**

No P0/P1. The autopilot flows use the `seq.current` guard consistently to drop stale responses, mutations toast on both success and failure, and the create flow handles partial success (autopilot created but trigger creation fails) with a distinct message and by still closing + reloading. The following are P2/advisory — worth a follow-up but none block merge.

**[P2] Unknown run status/source renders the raw i18n key path** — `panel/AutopilotDetailPage.tsx:284-285`
`t(\`loop.automation.runStatus.${r.status}\`)` and `runSource.${r.source}` have no fallback. `t()` returns the key path verbatim on a miss (confirmed by `dmworkbase/src/i18n/__tests__/I18nService.test.ts:46`). If the backend ever returns a status/source outside the current enum (e.g. `queued`), the user sees the literal string `loop.automation.runStatus.queued`. This directly contradicts the author's own guidance in `api/types.ts:128` ("后端驱动字符串——渲染未知值时走通用兜底，勿用穷尽 switch") and is inconsistent with the card, which *is* guarded (`AUTOPILOT_RUN_DOT[...] ?? AUTOPILOT_RUN_DOT_FALLBACK`, `pages/AutomationPage.tsx:36`). Suggest a fallback to the raw value, e.g. render `t(key)` and if it equals `key`, show `r.status` instead.

**[P2] "Once" schedule silently recurs every year** — `ui/autopilotSchedule.ts:56-61`
`toCron` maps "once" to `min hour DoM Month *`, which a 5-field cron re-fires every year on that month/day. The author documents this at the top of the file ("单次以「钉日+月」近似，无法真正一次性触发"), so this is a known backend-contract limitation, not an oversight — but the UI label is "Once/单次", which implies a single fire. Users can be misled. Since the backend only accepts 5-field cron, consider either a clarifying hint in the schedule UI or tracking a real one-shot capability as follow-up.

**[P2] `saveDesc` leaves the in-memory `autopilot.description` stale** — `panel/AutopilotDetailPage.tsx:123-134`
After a successful description save it resets the dirty flags and calls `onChanged?.()` but not `reload()`, so the local `autopilot` object keeps the old description. Impact is low: the value *is* persisted server-side, the textarea renders from local `desc` state (correct), and `autopilot.description` isn't read anywhere else in the view, so it self-heals on the next `reload()` (e.g. a status toggle). Cheapest fix for consistency: call `mutated()` instead of `onChanged?.()` here, matching the other mutations.

**[P2] In-flight fetches update state after unmount** — `pages/AutomationPage.tsx:27-34`, `panel/AutopilotDetailPage.tsx:83-104`
The `seq.current` guard prevents a stale response from overwriting a newer one, but doesn't stop a `.then`/`.finally` from running after the panel closes, which can trigger React "set state on unmounted component" warnings. A `useEffect` cleanup that bumps `seq.current` (or an `ignore` flag) would close it. Minor.

**[P2] `runNow` refetches immediately and may miss the new run** — `panel/AutopilotDetailPage.tsx:153`
`triggerAutopilot(...).then(reload)` can fire `listAutopilotRuns` before the new run is queryable if run creation is async server-side, so the user may not see the entry until a manual refresh. Cosmetic eventual-consistency gap.

**[nit] `loopControls.css` `:is()` skin uses `!important` on focus** — `ui/loopControls.css:148`
The skin is correctly scoped to `:is(.loop-page, .loop-sd, .loop-fields, .loop-modal, .loop-apd)` and `.loop-modal` is applied on the Semi `Modal`s that portal to `body`, so coverage is right and it won't leak to non-loop pages. The `!important` on the focus border is a small specificity hammer that could complicate future overrides, but it's contained.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Clean typecheck and lint for the changed package; all findings are advisory (P2/nit) and safe to address in a follow-up.

## 4. Suggestions

- Guard the two `t(\`...runStatus/runSource.${...}\`)` interpolations with a raw-value fallback so a future backend enum addition degrades gracefully instead of showing a key path.
- Align `saveDesc` with the other mutations (`mutated()`), and add a fetch cancellation flag to the two panels for clean unmount.
- If a true one-shot schedule is a product goal, track the "once ≈ annual cron" gap as an explicit follow-up.

## 5. Additional notes (reviewer coverage)

- **Advisor legs:** this run used a two-leg advisory pass. The long-context leg (gemini-3.1-pro) completed and its findings were absorbed above (its one P1 on `saveDesc` was downgraded to P2 after independent verification — the value is persisted and the stale field is never rendered). The line-level adversarial leg (codex) was **absent** this run due to an upstream gateway tool-config error; I performed the line-level pass myself rather than treating its absence as a pass.
- **Not covered / blind spots:** the backend `/fleet/api/v1/autopilots*` contract is not in this diff and was assumed to match the types; timezone round-trip relies on the server returning `next_run_at` as ISO (browser-local display via `getHours` is fine for that), but the cron `timezone` field vs. server evaluation was not verifiable from the frontend alone.

OctoBoooot (approved):
## Review: #663 — delta re-review (Once blocker fixed)

The new commit `fix(loop): drop unsupported one-shot trigger and surface run-history errors` fully resolves the blocker — "Once" is **removed**, not relabeled — and fixes the run-history 🟡. Byte-verified at head `b4137817e809`. Flipping my verdict.

Risk tier: high; `needs-human-review` stands. Base `feat/octo-loop`.

### Resolved — the Once→yearly-recurring-cron blocker (removed cleanly)
The misleading option is deleted at the root, which is the strongest fix:
- `Frequency` type: `"once"` dropped → `"daily" | "weekly" | "monthly"` (`autopilotSchedule.ts`).
- `FREQUENCIES` array (`ScheduleFields.tsx`): `"once"` removed → the UI no longer offers it; the once-only `DatePicker` and its import are gone.
- `toCron`: the entire `case "once"` (the `` `${min} ${hour} ${D} ${M} *` `` yearly-recurring form) deleted.
- `parseCron`: once-branch removed; unrecognized/legacy cron shapes now fall back to `daily` (so any pre-existing `M H D M *` row degrades to daily rather than mis-round-tripping to a fake "once").
- `todayStr`/`date`/`summary.once` + i18n `once` labels (both locales) removed. Header comment updated: "无一次性触发，故不提供「单次」（与 multica 一致）".

No user can now select a "Once" that silently recurs annually. (yujiawei's still-listed "单次实际每年重复" 🟡 was against the old head `abb53e1db919`; it no longer applies here.)

### Resolved — run-history load failure now distinguishable (the kept 🟡)
`AutopilotDetailPage.tsx`: `listAutopilotRuns` now returns `{ ok }` and sets `runsError`; on failure it renders a dedicated error row with a **Retry** button (`runsError`/`retry` i18n added in both locales) instead of swallowing into the "no runs" empty-state. Exactly the fix.

### Remaining (yujiawei's new 🟡s — non-blocking)
- **Run status/source label has no unknown-value fallback** (`:293-294`): `t(\`loop.automation.runStatus.${r.status}\`)` / `runSource.${r.source}` render the raw key if the backend returns a value outside the i18n map. `AutopilotRunStatus`/`AutopilotRunSource` are explicitly backend-driven strings, and the *dot color* already has `?? AUTOPILOT_RUN_DOT_FALLBACK` (`:282`) — the label should get the same treatment. Minor (cosmetic, needs an unmapped enum; degrades to an ugly string, not a break).
- **saveDesc local-object staleness** — minor state-consistency nit, non-blocking.

### Accountability
This is the blocker I initially under-graded (APPROVE), then corrected to REQUEST_CHANGES after Jerry-Xin flagged it — both of us byte-verified there was no backend one-shot discriminator. The author took the cleanest path (remove the option) rather than a stopgap relabel. Credit: Jerry-Xin (blocker) and yujiawei (run-history + the label-fallback follow-up). The rest of my original review (timezone wired, PATCH-partial correct, i18n parity, narrow API surface) held throughout.

Jerry-Xin (dismissed):
Re-review at head `b4137817e809` (prior reviewed head `abb53e1db919`). Scope OK — extends `packages/dmloop`.

## ✅ Prior 🔴 RESOLVED — "Once/单次" yearly-cron bug is fixed
The new commit removes the broken one-shot cadence entirely (the correct fix, matching multica's "no true one-shot trigger" backend constraint):
- `Frequency` union `autopilotSchedule.ts:4`: dropped `"once"` → `"daily" | "weekly" | "monthly"`.
- `ScheduleFields.tsx:10` `FREQUENCIES` no longer offers `once`; the `DatePicker` + `date` field + `todayStr()` removed.
- `toCron` `case "once"` (`${min} ${hour} ${D} ${M} *` = fires every year forever) **deleted**; `parseCron` once-branch deleted (unknown shapes now fall back to `daily`); `describeSchedule` once-case and i18n `freq.once`/`summary.once` (en+zh) removed.
- Header comment updated: 「后端…无一次性触发，故不提供「单次」（与 multica 一致）」.

This fully addresses the reachable Once→recurs-annually defect. Note: removing `once` is the fix, **not** a regression — the backend has no fire-once mechanism, so offering it was misleading.

## ✅ Prior 🟡 also addressed
Run-history load failure no longer masquerades as empty: `AutopilotDetailPage.tsx` now tracks `runsError` and renders an explicit error + Retry (i18n `runsError`/`retry` added to both locales), replacing the silent `.catch(()=>({runs:[],total:0}))`.

## 🔴 Blocking — NEW (in-scope this PR)
**"My loops" can display all workspace loops instead of only mine.** `LoopPage.tsx:64` mounts `<IssuePage defaultScope="involves" defaultView="grouped">`. In `IssuePage.tsx`, `myMemberId` (`:90`) resolves async from `useAssigneeCandidates`, so on first `reload` it is `undefined`; the grouped branch (`:130-132`) then falls through to `listGroupedIssues({ ...gp, assignee_types: scopeToAssigneeTypes("involves") })` where `scopeToAssigneeTypes("involves")` returns `undefined` (`:40-43`) = **unscoped = all members' loops**. The `seq`-guarded re-fetch corrects the happy path once `myMemberId` resolves, but:
- Every mount briefly flashes all workspace loops under a "My loops" label, and
- For members without an `octo_uid` (native/non-octo-bridged members — `directory.ts:53` sets `octo_uid` nullable), `myMemberId` **never resolves**, so "My loops" **permanently shows everyone's loops** while the pill is disabled (`:241`) and the user cannot correct it.

Fail-safe fix: when `scope === "involves"` and `myMemberId` is unresolved, do not fetch (render loading), or render empty/error — never fall back to the unscoped "all" query.

## 💬 Non-blocking
- 🟡 `loopControls.css:145-149` uses `!important` + hardcoded fallback colors (also hardcoded status colors `sideDetail.css`, `meta.tsx`); prefer semantic/component tokens per project CSS convention.

## ✅ Highlights
- API wrapper centralizes workspace/auth via the shared HTTP client.
- Unsaved task-instruction drafts preserved across background reloads (`descDirtyRef`).

OctoBoooot (changes requested):
## Review: #663 — correcting my verdict again (Approve → Request changes) on the "My loops" scope fallback

**Verdict:** Request changes (correcting my APPROVE at this head)
Jerry-Xin caught a real defect I missed — I scoped my deep-read to the schedule/API and didn't review the new "My loops" wiring. The "My loops" tab enters with `defaultScope="involves"`, but when the current user's member id can't be resolved, the fetch silently falls back to **all workspace loops** shown under a "mine"-labeled view. Byte-verified; reachable; introduced by this PR. Flipping to REQUEST_CHANGES.

Risk tier: high; `needs-human-review` stands. Base `feat/octo-loop`. (The Once→yearly-cron blocker from the prior head IS fixed — this is a different, newly-surfaced issue.)

### Major — "My loops" falls back to all-workspace loops when `myMemberId` is unresolved
`IssuePage.tsx:129-132`:
```
const req =
  scope === "involves" && myMemberId
    ? listMyGroupedIssues(myMemberId, gp)
    : listGroupedIssues({ ...gp, assignee_types: scopeToAssigneeTypes(scope) });
```
and `scopeToAssigneeTypes("involves")` returns **`undefined`** (`:40-42` only map `members`/`agents`). So when `scope==="involves"` but `myMemberId` is falsy, the request is `listGroupedIssues({assignee_types: undefined})` → **every loop in the workspace, unfiltered** — rendered under the "My loops" tab.

`myMemberId` is unresolved in two reachable cases:
1. **Every initial load (transient, all users).** `useAssigneeCandidates` starts `[]` and fetches in an effect, so `myMemberId` is `undefined` on first render. `reload` fires via `useEffect(reload, [reload])` before `cands` resolve → a flash of all-workspace loops, then self-corrects when `cands` arrive and `reload` re-runs (`myMemberId` is in the dep array). Minor on its own.
2. **Permanent for members with no `octo_uid` match** (`:90-93`: `cands.find(c => c.type==="member" && c.octo_uid===uid)`). If the logged-in user has no matching member candidate, `myMemberId` stays `undefined` forever → "My loops" **permanently shows all workspace loops**. This is the real defect.

**Why it's PR-introduced (not pre-existing):** the `involves` fallback itself exists on base, but base only reached `involves` through the scope pill, which is `disabled={!myMemberId}` (`:241`) — so a no-`octo_uid` user could never *get* to involves. **This PR adds `defaultScope="involves"` as the "My loops" tab entry** (`LoopPage.tsx:64`), a new door that bypasses the disabled-pill guard and lands such a user directly on the unfiltered view. The pill guard protects clicking-to-involves, not defaulting-to-involves.

**Severity — major, not blocker:** the loops shown are still workspace-scoped by the backend (`listGroupedIssues` returns only the caller's workspace), so this is not cross-tenant/privilege data leakage — it's a scoping/correctness defect (a "mine" view showing everyone's loops). Real, reachable, and defeats the feature for affected users; requires a fix, but no security escalation.

**Fix direction:** when `scope==="involves" && !myMemberId`, don't fall back to unfiltered — render an empty/"can't determine your loops" state (or gate the "My loops" tab on a resolved member id), mirroring how the pill already disables the option. Don't let an unresolved identity silently widen scope to all.

### Confirmed fixed at this head (prior round)
- **Once→yearly cron blocker: removed** — `"once"` gone from `Frequency`/`FREQUENCIES`/`toCron`/`parseCron`/i18n. Clean.
- **Run-history load failure: now a distinct error state + Retry** (`AutopilotDetailPage.tsx`), not swallowed into empty. Good.

### Also (non-blocking, verified prior round)
- Run status/source label lacks the unknown-value fallback the dot-color has (`:293-294`). Minor.
- Jerry-Xin's CSS 🟡 (`!important` + hardcoded color): the `!important` is a single disciplined focus-border scoped under `:is(.loop-*)`; low-severity nit, consistent with the design-system layer.

### Accountability
Second correction on this PR — first I under-graded the Once blocker (fixed since), now I missed the "My loops" scope fallback entirely by not reading the new tab wiring. Lesson: a feature whose whole point is "filter to me" must have its identity-resolution failure path reviewed — an unresolved "me" that widens rather than narrows scope is the classic failure. Credit: Jerry-Xin (both catches). The Once fix and run-history fix from the prior round are real and stand.

Jerry-Xin (approved):
## ✅ APPROVE (re-review @ b4137817e809) — verdict corrected from REQUEST_CHANGES

My earlier CR on this PR is **withdrawn** after byte-re-verification. Summary of the full arc:

### The original blocker is genuinely FIXED ✅
At the prior head, "Once/单次" schedule serialized to a yearly-recurring cron (`toCron("once")` → `${min} ${hour} ${D} ${M} *`) with no backend one-shot discriminator — a reachable, in-scope defect. This commit **removes the `once` cadence entirely** (byte-verified): dropped from the `Frequency` union, `FREQUENCIES` list, `toCron`, `parseCron`, `describeSchedule`, and i18n (en+zh); header comment now documents that the backend has no one-shot trigger (aligned with multica). Removing an unimplementable one-shot is the correct fix, not a regression. Run-history load failures now surface an explicit error state + Retry button (the prior 🟡 is also resolved).

### Correcting my own re-review CR (over-graded)
My re-review posted two "My loops" 🔴s. On byte re-verification against the **correct** sink (`IssuePage` via `LoopPage.tsx:64 defaultScope="involves"` — not `AutomationPage`, which has no scope filter at all), both are **real but non-blocking 🟡**, and I withdraw them as blockers:

- **My-loops initial paint shows all loops:** real, but a **self-correcting first-paint flash** — `myMemberId` starts `undefined` (assignee candidates load async), so the first render falls to `assignee_types: undefined` (all); `myMemberId` is in `reload`'s deps with a `seq` guard, so it re-fetches the correct involves set on the next tick. Transient, not persistent.
- **Member without matching `octo_uid` sees all under My-loops:** real for native members lacking an octo bridge, but (a) the scope toggle still works — only the *involves* pill is disabled-visible; `all`/`members`/`agents` remain clickable; and (b) **no data-exposure escalation** — these loops are already visible to any workspace member in the "All" tab. It's a wrong-filter UX degrade, not a leak or privilege escalation.

**Provenance (decisive):** this commit did not touch that logic — the `myMemberId`/involves predicate + disabled guard are **pre-existing on base `feat/octo-loop`**, unchanged; the PR only newly *surfaces* it via the `myloop` default tab. Blocking a fix-commit for pre-existing, self-correcting, non-escalating UART UX would be miscalibrated.

### Non-blocking notes (optional)
- 🟡 Under the new "My loops" default, consider seeding `scope="all"` (or keeping involves disabled-visible) until `myMemberId` resolves, so first paint doesn't flash all loops and non-octo members aren't dumped into an all-list under a "My" label. (Touches pre-existing base logic, not this delta.)
- 🟡 `loopControls.css` uses one `!important` + some hardcoded colors; prefer semantic component variables where practical. Overrides are scoped with containment comments, so this is a convention nit.

### Highlights
Once-removal is clean end-to-end; run-history error UX added; narrowly-scoped `/autopilots` API surface; i18n parity (en/zh). Verdict: **APPROVE.**

OctoBoooot (commented):
## Review: #663 — one disambiguator on the "My loops" scope finding (holding REQUEST_CHANGES)

**Verdict unchanged: Request changes.** Not re-litigating — just pinning the one fact that separates "transient flicker (🟡)" from "permanent wrong-scope (major)", since the latest APPROVE reframes it as only the first-screen flash.

### The permanent case is real, and the codebase documents the affected population
Jerry-Xin's downgrade ("default scope=all to avoid first-screen flicker") addresses only failure mode #1 (the self-correcting flash while `cands` load). It doesn't cover #2, which is **permanent**:

- `myMemberId` resolves via `cands.find(c => c.type==="member" && c.octo_uid === WKApp.loginInfo.uid)` (`IssuePage.tsx:90-92`).
- `AssigneeCandidate.octo_uid` is **`string | null`** (`types.ts:15`), and the members endpoint returns `octo_uid?: string | null` per member (`directory.ts:32`).
- `directory.ts:3-4` explicitly documents the population: **"无 octo_uid 的原生成员回退其后端显示名"** — *native (non-octo-bridged) members* are a real, first-class case with a dedicated fallback path, and `directory.ts:41` branches on `members.some(m => m.octo_uid)` precisely because some members have none.

So a **native member** viewing "My loops" never resolves `myMemberId` → `scope==="involves" && undefined` → `listGroupedIssues({assignee_types: undefined})` → **all workspace loops, permanently**, under a "mine"-labeled tab. That is not a flicker; it's a persistent wrong-scope for an entire supported user class. Defaulting to `scope="all"` (Jerry-Xin's suggestion) would fix the flash but leave the permanent case — and would also change what "My loops" means on load.

### Severity restated
Major (not blocker): results stay backend-workspace-scoped, so no cross-tenant/privilege exposure — but "My loops" showing everyone's loops for native members is a real, reachable, PR-introduced correctness defect (the `defaultScope="involves"` tab entry bypasses the disabled-pill guard that previously gated involves on a resolved member id). Fix: when `involves && !myMemberId`, render empty/"can't determine your loops" or gate the tab, rather than falling open to all.

The Once→cron blocker and run-history fixes from the prior head are genuinely resolved (verified). This is the only open item. Credit: Jerry-Xin for originally surfacing it; the disambiguator here is just to keep the permanent-vs-transient distinction on record for whoever reconciles the split verdicts.

Jerry-Xin (changes requested):
## 🔴 REQUEST_CHANGES — converging my verdict (was APPROVE) @ b4137817e809

Updating my own verdict from APPROVE to REQUEST_CHANGES to converge with the peer review on the "My loops" scope fallback. This is a severity self-correction, not a new finding.

### Why I'm moving my half-notch
My prior APPROVE listed the permanent no-`octo_uid` case as a strong 🟡 while holding the verdict at APPROVE, on the implicit assumption that non-`octo_uid` members are rare enough to keep it non-blocking. That assumption doesn't survive the bytes: `directory.ts:3-4` explicitly documents **"无 octo_uid 的原生成员"** (native, non-octo-bridged members) as a first-class, code-supported population, with a dedicated fallback branch (`members.some(m => m.octo_uid)`). Absent measured prevalence, I don't get to assume that class is rare — so a **permanent** wrong-scope for a code-supported user class, under a "mine"-labeled view, lands at **major → REQUEST_CHANGES** on my ladder.

### The finding (agreed facts)
- **Sink:** `IssuePage.tsx:130-132` — when `scope==="involves"` but `myMemberId` is unresolved, the fetch falls to `listGroupedIssues({ assignee_types: scopeToAssigneeTypes("involves") })`, and `scopeToAssigneeTypes("involves")===undefined` → **all workspace loops** under "My loops".
- **Reachability (PR-introduced):** the `involves` fallback pre-exists on base, but base only reached it via the scope pill, which is `disabled={!myMemberId}`. This PR adds `defaultScope="involves"` as the "My loops" tab entry (`LoopPage.tsx:64`) — a new door that bypasses the disabled-pill guard. So the *reachability* is introduced by this PR even though the predicate isn't.
- **Two modes:** (1) transient first-frame flash for everyone (self-correcting via `myMemberId` in `reload` deps + `seq` guard) — minor; (2) **permanent** for members whose `octo_uid` never matches — major.

### Ceiling (also agreed)
Not a security escalation — the backend still workspace-scopes results, and these loops are already visible in the "All" tab. This is a **wrong-scope correctness defect, not cross-tenant data exposure**. Major, not critical.

### Fix (one line, we all point at it)
On `scope==="involves" && !myMemberId`, render an empty / "can't determine your loops" state instead of falling open to `assignee_types: undefined` (all). Optionally also seed `scope="all"` until `myMemberId` resolves to kill the first-frame flash.

### Unaffected
The prior-round blocker (Once → yearly-recurring cron) is **genuinely fixed** at this head (the `once` cadence was removed entirely — byte-verified). Run-history load failure now surfaces an explicit error state + Retry. Timezone is wired alongside the cron; partial updates use PATCH (correctly sidestepping the nameless-field-wipe class). i18n parity holds.

Verdict: **REQUEST_CHANGES** on the single My-loops fallback line. Clears the moment that empty-state guard lands.

PR #3, opened 2026-05-12
K8s-friendly nginx: cluster DNS resolver, /matter route, MATTER_API_URL env

an9xyz:
## Summary

Three coupled changes to \`nginx.conf.template\` + \`docker-entrypoint.sh\` so the prebuilt image works out of the box in both Kubernetes and docker-compose deployments:

1. **K8s-friendly resolver** — Replace the hardcoded \`resolver 127.0.0.11\` (docker embedded DNS, only valid inside docker-compose) with the cluster's CoreDNS service \`kube-dns.kube-system.svc.cluster.local\`. This is what production nginx configs already use and unblocks variable-form \`proxy_pass\` on K8s (the previous setting returned 502 on /summary because nginx could not resolve the service hostname at request time).
2. **\`/matter/\` route** — Add a \`location /matter/\` block that proxies to \`\$matter_api_url\` (env: \`MATTER_API_URL\`) with the same optional-503 + rewrite pattern already used for /summary. Closes a gap where the matter (task / timeline) service had no front-door route in octo-web.
3. **Entrypoint envsubst** — extend \`docker-entrypoint.sh\` to envsubst the new \`MATTER_API_URL\` variable alongside existing \`SUMMARY_API_URL\` and \`API_URL\`.

## Test plan

- [ ] In K8s: \`MATTER_API_URL=http://octo-matter:8080\` env set, \`/matter/api/v1/matters\` returns the backend's response (401 without token is expected)
- [ ] \`MATTER_API_URL\` unset → \`/matter/*\` returns 503 \`{"msg":"matter is not configured"}\`
- [ ] docker-compose users still work — note: they may need to switch \`resolver\` back to \`127.0.0.11\` manually (documented in code comment)
- [ ] \`/api/\`, \`/v1/\`, \`/summary/api/v1/\`, \`/\` still route as before

PR #2, opened 2026-05-11
YUJ-446: make smart-summary upstream optional (fix nginx emerg at startup)

yujiawei:
## Problem

`nginx.conf.template` referenced the smart-summary service as a hard
proxy upstream:

```nginx
location /summary/api/v1/ {
    proxy_pass  http://summary-api:8080/api/v1/;
    ...
}
```

nginx resolves upstream hostnames at startup. If octo-web is deployed
without octo-smart-summary on the same docker network (including the
stock `Mininglamp-OSS/octo-server/docker/octo` compose stack in its
pre-YUJ-446 form), the container exits immediately:

```
[emerg] host not found in upstream "summary-api"
        in /etc/nginx/conf.d/default.conf:36
```

Blocks OSS users from running octo-web by itself and blocks the
`docker compose up -d` happy path.

## Fix

- `SUMMARY_API_URL` env variable (default empty).
- Entrypoint runs `envsubst` over a new variable slot so the value
  lands in an nginx `set` directive rather than as a raw upstream
  name.
- Empty → `/summary/api/v1/*` short-circuits to `503 {"msg":"smart-summary is not configured"}`.
- Set but unreachable → request-time resolution returns 502 (no
  startup impact).
- Set + reachable → proxied normally.

Explicit `resolver 127.0.0.11 ipv6=off valid=30s;` for Docker's
embedded DNS.

## Validation

Built `octo-web:yuj446` and ran three scenarios:

| `SUMMARY_API_URL`                 | Container up | `/` | `/summary/api/v1/foo` |
| --------------------------------- | :----------: | :-: | :--------------------: |
| _unset_                           | ✅           | 200 | 503 JSON              |
| `http://summary-api:8080` (stub)  | ✅           | 200 | 502 (runtime DNS fail) |
| `http://summary-api:8080` (real)  | ✅           | 200 | 200 (proxied)          |

Companion to `Mininglamp-OSS/octo-server` PR #2.

PR #1, opened 2026-05-11
fix: open-source audit - remove internal refs, PII, and stale docs

Jerry-Xin:
## Open-Source Readiness Fixes

Based on internal audit, this PR addresses the following issues:

### Changes
- **Makefile**: Replaced internal Alibaba Cloud registry URL with generic placeholder, renamed Docker image from `tangsengdaodaoweb` to `octoweb`
- **botfather-e2e.js**: Moved hardcoded test password to environment variable (`BOTFATHER_PASS`)
- **wide-table.csv**: Replaced realistic PII (names, phones, bank accounts, salaries, addresses) with obviously fake placeholder data
- **nginx.conf.template**: Added documentation comment for `summary-api` service reference
- **OIDC test files**: Replaced internal codename `aegis.example.com` with generic `idp.example.com`
- **docs/**: Removed directory containing internal design documents and branch references
- **TODO comments**: Removed 4 TODO comments referencing internal ticket system (YUJ-138, YUJ-376)

### Not Changed (by design)
- License headers in source files (separate effort)
- MPL-2.0 dependency (axe-core, dev-only)
- Remaining general TODO/FIXME comments (not referencing internal systems)

PR #665, opened 2026-07-12
fix(loop): stop batch edits silently dispatching agent runs

MrAladdin:
Follow-up to #653, fixing three findings surfaced by the adversarial review and verified against the backend.

## Findings & fixes

1. **[high] Batch edits silently launched agent runs.** The batch bar called `batchUpdateIssues` without `suppress_run`, and the backend `BatchUpdateIssues` dispatches an agent run per issue unless `updates.suppress_run` is set (`issue.go` `!req.Updates.SuppressRun`). A bulk status/assignee change over N issues fired N runs with no confirmation. Batch edits are administrative bulk mutations; dispatching an agent is the deliberate *per-issue* action, which stays behind the `RunConfirm` preview. All three batch calls (`status` / `priority` / `assignee`) now pass `suppress_run: true`. The single-issue edit path (`requestStatus` / `requestAssign` → `RunConfirm`) is untouched.

2. **[medium] Sub-issue create left the parent board stale.** `onCreated` only refreshed the detail page's own child list; the parent board's child progress/counts did not update. `onCreated` now also calls `onChanged?.()`, so the parent list reloads (matching every other single-field mutation in the file).

3. **[medium] Batch-delete re-entrancy guard was ineffective.** The confirm dialog opened while `busy` was `false`, and its `onOk` closure captured the render-time `busy`; double-opening the dialog let two `onOk` callbacks both pass `if (busy) return` and call `batchDeleteIssues` twice. `runBatch` now sets a synchronous `busyRef` at entry (check-and-set with no `await` between), so a second invocation is always blocked; `busy` state still drives the UI `disabled`.

## Verification (real browser, dev — Alice / `mv2-e2e-alice`)

- **#1** — batch status change: request body is `{"updates":{"status":"in_progress","suppress_run":true}}`, response `{"updated":2}`, rows update to the new status, and `agent-task-snapshot` shows no new running task.
- **#2** — sub-issue create fires `POST /issues` → `GET /issues/:id/children` → **`GET /issues?...` (parent reload)** → `agent-task-snapshot`; before the fix the parent reload was absent.
- **#3** — batch path runs cleanly through the new `busyRef` guard; the double-fire race is a synchronous check-and-set with no await window (verified by review).

Typecheck clean (`tsc --noEmit`, dmloop package, 0 errors in package sources).

> Note: Claude adversarial review passed clean end-to-end. Codex adversarial review emitted an early `approve` but its stream dropped mid-analysis on the proxy endpoint (a known infra flake, not a finding); a clean Codex re-run is a follow-up once the endpoint is stable.

## Blast radius (Rule 8)

Real diff is **2 files** (`git diff --stat origin/feat/octo-loop...HEAD` → `IssueList.tsx`, `IssueDetailPage.tsx`); the CI/hook file count is inflated by counting against `origin/main`.

- `batchUpdateIssues` is called **only** from `IssueList.tsx` (all three call sites are in this diff). Its signature is unchanged — `suppress_run?: boolean` already exists on the shared `UpdateIssueReq` type, so no shared-type or contract change and no other consumer is affected.
- `onChanged?.()` is an existing optional prop already invoked elsewhere in `IssueDetailPage.tsx` (lines 191/477/1040); this adds one more call site. No prop/signature change; the parent `IssuePage` already routes it through `onMutated`.
- `busyRef` is component-local.

No exported/shared type, prop, hook signature, util, cross-package import, or DB/API contract changed → no out-of-module consumer impacted.

### DIFF
--- a/packages/dmloop/src/panel/IssueDetailPage.tsx
+++ b/packages/dmloop/src/panel/IssueDetailPage.tsx
@@ -1163,6 +1163,8 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
           Toast.success(t("loop.toast.created"));
           // 只刷新子列表(非整页 reload,避免详情主体闪 loading);key-remount 已隔离跨 issue 陈旧写入。
           listChildren(issueId).then(setChildren).catch(() => {});
+          // 通知父级:新子 issue 改变了父看板的子进度/计数,不刷父会陈旧。
+          onChanged?.();
         }}
       />
     </div>
--- a/packages/dmloop/src/panel/IssueList.tsx
+++ b/packages/dmloop/src/panel/IssueList.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { Table, Select, Tag, Typography, Button, Toast } from "@douyinfe/semi-ui";
 import { Trash2, X } from "lucide-react";
 import { useI18n } from "@octo/base";
@@ -40,6 +40,10 @@ export default function IssueList({
   const { requestAssign, requestStatus, runConfirmModal } = useRunConfirm();
   const [selected, setSelected] = useState<string[]>([]);
   const [busy, setBusy] = useState(false);
+  // busy 的同步镜像:setBusy 是异步的,confirmDelete 弹窗的 onOk 闭包捕获的是渲染时的
+  // busy(可能陈旧),双击开两个弹窗/两次 onOk 会各自看到 busy=false 而重复批量写。ref
+  // 在进入 runBatch 时同步置位,结构性挡住重入(setBusy 仅驱动 UI 的 disabled)。
+  const busyRef = useRef(false);
 
   // 筛选/搜索/排序/翻页令 issues 变化后,裁掉已不可见的选中项 —— 否则批量条会对
   // 当前结果集看不到的行发批量写(改/删隐藏行)。只保留仍在可见集合里的 id。
@@ -56,11 +60,13 @@ export default function IssueList({
     onChanged();
   };
 
-  // 批量:写失败 toast+中止;成功清选择+刷新。批量不走 RunConfirm 预览(显式批操作,
-  // 与原生 web 一致);状态/指派若触发 agent run 由后端处理。
-  // busy 守卫放这里(而非逐个控件 disabled):结构性防重入,任一触发器在途都挡住重叠批量写。
+  // 批量:写失败 toast+中止;成功清选择+刷新。批量不走 RunConfirm 预览(显式批操作);
+  // 且一律带 suppress_run —— 批量改状态/指派是管理性整理,绝不能每条静默起一个 agent run
+  // (要派单请逐条走 RunConfirm 预览确认)。派单是有意的单条动作,不是批量副作用。
+  // 重入守卫用 busyRef 同步置位(见上):任一触发器在途都挡住重叠批量写,不受闭包 stale 影响。
   const runBatch = async (fn: () => Promise<unknown>) => {
-    if (busy) return;
+    if (busyRef.current) return;
+    busyRef.current = true;
     setBusy(true);
     try {
       await fn();
@@ -70,6 +76,7 @@ export default function IssueList({
     } catch (e) {
       Toast.error((e as Error).message || t("loop.toast.saveFailed"));
     } finally {
+      busyRef.current = false;
       setBusy(false);
     }
   };
@@ -171,7 +178,7 @@ export default function IssueList({
             size="small"
             disabled={busy}
             value={NO_VALUE}
-            onChange={(v) => runBatch(() => batchUpdateIssues(selected, { status: v as IssueStatus }))}
+            onChange={(v) => runBatch(() => batchUpdateIssues(selected, { status: v as IssueStatus, suppress_run: true }))}
             style={{ width: 130 }}
           >
             {ISSUE_STATUS_ORDER.map((s) => (
@@ -183,7 +190,7 @@ export default function IssueList({
             size="small"
             disabled={busy}
             value={NO_VALUE}
-            onChange={(v) => runBatch(() => batchUpdateIssues(selected, { priority: v as IssuePriority }))}
+            onChange={(v) => runBatch(() => batchUpdateIssues(selected, { priority: v as IssuePriority, suppress_run: true }))}
             style={{ width: 120 }}
           >
             {PRIORITY_ORDER.map((p) => (
@@ -194,7 +201,7 @@ export default function IssueList({
             size="small"
             value={null}
             valueName={null}
-            onChange={(id, type) => runBatch(() => batchUpdateIssues(selected, { assignee_id: id, assignee_type: type }))}
+            onChange={(id, type) => runBatch(() => batchUpdateIssues(selected, { assignee_id: id, assignee_type: type, suppress_run: true }))}
           />
           <Button
             size="small"


OctoBoooot (approved):
## Review: stop batch edits silently dispatching agent runs (#665)

**Verdict:** Approve
Tight follow-up to #653 fixing three real findings, each verified to the sink. The headline fix (batch edits no longer fire an agent run per issue) is correct and matches the backend gate; the re-entrancy fix correctly avoids the flag-wedge trap. No defects.

Risk tier: normal. Base `feat/octo-loop`; CI green (`code-review` pending, not failing). +17/-8, 2 files.

### What I verified
- **#1 — `suppress_run: true` on all three batch calls (the [high] fix): correct.** `batchUpdateIssues(ids, updates: UpdateIssueReq)` (`issueApi.ts:168`) sends `{issue_ids, updates}`, and `suppress_run?: boolean` is a valid field on `UpdateIssueReq` (`types.ts:210`), so status/priority/assignee batches now serialize `suppress_run: true` inside `updates` — the exact key the backend checks (`!req.Updates.SuppressRun`). tsc would've rejected an unknown field, and the PR's browser trace confirms the request body (`{"updates":{"status":"in_progress","suppress_run":true}}`, no new running task). The single-issue `requestStatus`/`requestAssign` → `RunConfirm` path is untouched, so deliberate per-issue dispatch still works. Correct separation: bulk edit = administrative, per-issue = the intentional dispatch action behind the confirm preview.
- **#2 — parent board no longer stale after sub-issue create: correct.** `onCreated` now also calls `onChanged?.()` (`IssueDetailPage.tsx:1164`), reloading the parent list — consistent with the file's other single-field mutations.
- **#3 — batch-delete re-entrancy guard now effective, and correctly reset: verified.** `runBatch` (`IssueList.tsx:67-82`) sets `busyRef.current = true` synchronously at entry (check-and-set at `:68-69`, no `await` between), so a second `onOk` from a double-opened confirm dialog is structurally blocked regardless of the stale render-time `busy` closure — which was the actual bug. Crucially, the reset is in `finally` (`:78-80`): `busyRef.current = false` and `setBusy(false)` fire on **every** path (success or throw), so the guard can't wedge after one batch. (This is the correct shape — a liveness flag that resets unconditionally, the lesson from the #656 mounted-ref freeze.)

### Praise
- The re-entrancy fix targets the real root cause — a closure capturing stale `busy` state — with a synchronous `busyRef` check-and-set rather than papering over it, and it resets in `finally` so it can't deadlock the batch bar. And the `suppress_run` fix draws the right conceptual line: bulk mutations shouldn't have the side effect that only the deliberate per-issue path (behind `RunConfirm`) should trigger. The comments explain *why* at each site, so the next editor won't reintroduce either bug.

### Note
CI `code-review` is pending and the PR notes a Codex re-run is a follow-up (its stream dropped on an infra flake, not a finding) — worth confirming that comes back clean, but the Claude review and browser verification cover the three fixes and the diff is small and self-contained.

Jerry-Xin (approved):
Summary: The PR is in scope for `Mininglamp-OSS/octo-web` and the changes correctly address the described dmloop batch mutation and parent refresh behavior without introducing a visible blocker.

💬 Non-blocking

🟡 Warning — Test coverage is still thin for these regressions. I did not find local dmloop tests covering `IssueList` batch request payloads or duplicate batch-delete confirmation behavior. Given the prior silent run-dispatch bug, a focused test around `batchUpdateIssues(..., { suppress_run: true })` for status/priority/assignee and a re-entrancy test for `runBatch` would reduce future regression risk. References: `packages/dmloop/src/panel/IssueList.tsx:67`, `packages/dmloop/src/panel/IssueList.tsx:181`, `packages/dmloop/src/panel/IssueList.tsx:193`, `packages/dmloop/src/panel/IssueList.tsx:204`.

✅ Highlights

🔵 Suggestion — The `busyRef` check-and-set is the right shape for the stale closure issue: it guards synchronously before any awaited work, while keeping `busy` for UI state. Reference: `packages/dmloop/src/panel/IssueList.tsx:67`.

🔵 Suggestion — The batch update payloads now consistently include `suppress_run: true`, and `UpdateIssueReq` already supports that field, so this does not expand the shared API surface. References: `packages/dmloop/src/panel/IssueList.tsx:181`, `packages/dmloop/src/panel/IssueList.tsx:193`, `packages/dmloop/src/panel/IssueList.tsx:204`.

🔵 Suggestion — Calling `onChanged?.()` after child issue creation matches the existing parent-refresh pattern used elsewhere in the detail page and should refresh board/list child counts. Reference: `packages/dmloop/src/panel/IssueDetailPage.tsx:1167`.

Verification note: I attempted `pnpm exec tsc --noEmit -p packages/dmloop/tsconfig.json`, but this checkout does not currently expose a `tsc` binary to pnpm (`Command "tsc" not found`), so I could not independently rerun the typecheck in this environment.

yujiawei (approved):
# Code Review — PR #665 (octo-web)

Reviewed at head `7f41b948b42c7f74020f137f758442e9093651df` against merge-base `02750ab3` (base `feat/octo-loop`). Real diff = 2 files (`IssueList.tsx`, `IssueDetailPage.tsx`, +17/-8). Backend contract independently verified in the Multica backend repo (`Mininglamp-OSS/multica`, `main`).

## 1. Spec compliance

**Spec: ✅**

All three findings from the follow-up to #653 are implemented, nothing over-built, no deviation.

- **Missing:** none. All 3 batch call sites (`status` / `priority` / `assignee`) now pass `suppress_run: true` (`IssueList.tsx:181,193,204`); `onChanged?.()` added to sub-issue `onCreated` (`IssueDetailPage.tsx:1167`); `busyRef` synchronous re-entrancy guard added to `runBatch` (`IssueList.tsx:43,66,72,79`).
- **Over-built:** none. The single-issue edit path (`requestStatus` / `requestAssign` → `RunConfirm` → `patch`) is untouched and still dispatches per-issue as before; no `suppress_run` leaked onto the inline column Selects.
- **Deviation:** none. `suppress_run` semantics ("apply the change, skip the run this write would trigger") match the backend contract exactly.

**Backend contract verified end-to-end (cross-repo):** `UpdateIssueRequest.SuppressRun` exists on the backend and `BatchUpdateIssues` gates the sole run-dispatch on `... && !req.Updates.SuppressRun` before `dispatchIssueRun`. `WillEnqueueRun` itself does not enqueue. The single-update path uses the identical gate, so batch never dispatches while single edits are unchanged. The flag is already wired through the shared `UpdateIssueReq` type (`suppress_run?: boolean`) — no new contract surface. `batchDeleteIssues` is idempotent server-side (`continue` on an already-deleted id, returns 200), so a stray re-fire is harmless.

## 2. Code quality

**Quality: Approved**

No P0/P1. Fixes are correct and minimal:

- **Fix #1** — `busyRef`/`suppress_run` change is type-safe; batch bulk edits no longer fan out N silent agent runs. Correct.
- **Fix #2** — `onChanged` traced through all 4 mount sites of `IssueDetailPage` (`IssuePage` → `onMutated` = `reload()` + `refreshRunning()`, plus Autopilot/Agent/Squad detail pages → their `reload`). All safe; optional-chained, so any caller omitting it is fine. No render loop (the detail page stays mounted via `routeRight.push` independent of parent state). Correct.
- **Fix #3** — synchronous `busyRef.current = true` at `runBatch` entry (no `await` before it) structurally closes the concurrent double-fire window that the stale-closure `busy` check missed; `finally` releases it correctly on both success and error. Correct for its stated target (in-flight overlap).

`tsc --noEmit` could not be run cleanly in the ephemeral checkout (dependencies not installed → module-resolution noise only); the three changed hunks are trivially type-sound by inspection. All referenced i18n keys (`loop.batch.done`, `loop.batch.selected`, `loop.batch.deleteConfirm`) exist in both `zh-CN` and `en-US`.

### Non-blocking notes (P2 / advisory — do not block merge)

- **P2 (pre-existing, not introduced here):** the delete `Button` uses `disabled={busy}` (async state, `false` at click time), so a fast double-click can stack two confirm modals. `busyRef` dedupes *concurrent* `onOk`, but two modals confirmed *sequentially* re-fire `batchDeleteIssues` on the render-time `selected` — a benign no-op given the idempotent backend (spurious success toast + one extra refresh, no 404/error). If you want to tidy it: have `confirmDelete` guard a second modal, or read `selected` live at call time. Out of scope of this PR's stated fix.
- **P2 (pre-existing):** `runBatch`'s success side-effects (`onChanged()`, `setSelected([])`) live inside the same `try` as `await fn()`; a synchronous throw from `onChanged()` would surface a false `saveFailed` toast after a successful write. Untouched by this diff.
- **Advisory (release coordination, not a code defect):** this frontend depends on a backend build that carries the `suppress_run` gate. The deployed Multica backend (`Mininglamp-OSS/multica` `main`) has it; a backend without the field would silently drop the flag (`json.Unmarshal` ignores unknown keys) and still dispatch. The single-update `suppress_run` already shipped with #653, so #665 introduces no *new* deploy-ordering risk — just confirm the gated backend ships with/before this frontend.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. All three fixes are correct, minimal, and verified against the real backend contract; the only residuals are pre-existing, non-blocking P2 UX nits and a standard forward-compat deploy note.

### Review method

Independent trace + a multi-agent adversarial verification pass (3 lenses per fix, refute-first adjudication) — every raised concern was refuted or downgraded to non-blocking against exact code evidence. A second independent model advisor ran and surfaced only the two pre-existing P2s above plus one point (unknown-key rejection) that is refuted by the already-existing `suppress_run` field on both the shared request type and the backend struct. A third advisor leg was unavailable this round (infra error on the proxy endpoint) and is recorded as absent, not as tacit approval.

PR #668, opened 2026-07-12
feat(loop): redesign board, list, detail, run and new-loop views

superalsrk:
## Goal

Restyle the loop (回路) screens to the new UX/Figma design language and the
shared loop design-system skin, while preserving all existing issue-tracking
functionality. No backend/API changes.

## Changes

**Board**
- Status-icon column headers with counts; faint column tracks so empty
  status columns stay legible.
- Cards show priority + status icons, mono identifier, relative time, title,
  project and a neutral assignee badge. Drag-to-change-status preserved.

**List**
- Replaced the table with clean status-grouped rows; row click opens detail.
- Kept multi-select batch bar (status / priority / assignee with
  suppress_run, delete) and inline assignee edit with run pre-confirm.
- Pagination preserved.

**Header / filters**
- Title row + a 筛选 (filter) popover holding all existing filters and a
  显示 (display) popover for view switch + list sort. Filter logic unchanged.

**Detail**
- Breadcrumb topbar; native single-line/multi-line inputs for title,
  description and comments.
- Unified activity feed: system actions collapse into a "N activities" block
  (humanized text; noise hidden; the latest block is expanded by default,
  older blocks fold and are toggleable).
- Comment threads are flat (root + replies as sibling rows) inside one card,
  each with an always-visible reply box supporting multiple attachments
  (file chips). Comment row actions live in a top-right three-dot menu
  (copy / resolve / delete); existing comments are not editable.
- New-comment composer is a separated multi-line block distinct from replies.
- Read-only right sidebar (Properties / Details / Execution log) with
  collapsible, divider-separated groups; property edits go through the
  top-right three-dot menu. Active runs are always visible; only terminal
  runs fold behind "show history (N)". Sub-issue module shows only when the
  loop has children.

**Run detail modal**
- Rebuilt as a transcript (status badge, metadata chips, timeline bar,
  collapsible events) using the same polling data source, with a refined
  low-key palette.

**New loop**
- Now a standalone composer page (routed, not a modal) that maps to the
  existing create-issue + upload-attachment endpoints. Child-issue creation
  stays a modal.

## Verification

- `npx tsc --noEmit --skipLibCheck -p packages/dmloop/tsconfig.json` — 0 errors.
- `npx stylelint "packages/dmloop/src/**/*.css"` — 0 errors (only the
  repo-wide rgba/hex warnings, unchanged by this work).
- Vite dev server boots and compiles the loop modules cleanly.
- Two-axis (standards + spec) code review run across the change; findings
  fixed (dead-code cleanup, success toasts, multi-file upload fix).
- Note: manual in-browser visual verification of each screen is still
  recommended before merge.

### DIFF
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -1,6 +1,5 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
-  Typography,
   Input,
   Button,
   Spin,
@@ -9,8 +8,9 @@ import {
   Pagination,
   DatePicker,
   Checkbox,
+  Dropdown,
 } from "@douyinfe/semi-ui";
-import { Search, Plus, LayoutGrid, List as ListIcon, Users, ClipboardList, ArrowUp, ArrowDown } from "lucide-react";
+import { Search, Plus, LayoutGrid, List as ListIcon, Users, ClipboardList, ArrowUp, ArrowDown, SlidersHorizontal, Filter } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
 import type {
   Issue,
@@ -30,9 +30,7 @@ import IssueBoard from "../panel/IssueBoard";
 import IssueGroupBoard from "../panel/IssueGroupBoard";
 import IssueList from "../panel/IssueList";
 import IssueDetailPage from "../panel/IssueDetailPage";
-import CreateIssueModal from "../ui/CreateIssueModal";
-
-const { Title } = Typography;
+import NewLoopPage from "./NewLoopPage";
 
 type ViewMode = "board" | "grouped" | "list";
 
@@ -83,7 +81,6 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   const [scope, setScope] = useState<IssueScope>(defaultScope ?? "all");
   const [f, setF] = useState<Filters>({ keyword: "", gStatuses: [], gPriorities: [], gProjectIds: [], noProject: false, sortBy: "position", sortDir: "desc", dateField: "created_at" });
   const [page, setPage] = useState(0); // 0-based，仅列表视图分页
-  const [createOpen, setCreateOpen] = useState(false);
   const cands = useAssigneeCandidates();
   // 当前 octo 成员的后端 user_id(involves_user_id 需 UUID,非 octo uid)：
   // 复用订阅特性的身份解析——候选里 octo_uid===loginInfo.uid 的 member。未解析出则「与我相关」不可用。
@@ -203,6 +200,15 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
     WKApp.routeRight.push(<IssueDetailPage key={id} issueId={id} onChanged={onMutated} />);
   };
 
+  // 新建回路 → 唤起独立 composer 页（非弹窗，对齐 Figma）。创建成功后返回并刷新列表。
+  const openNewLoop = () => {
+    WKApp.routeRight.push(
+      <NewLoopPage
+        onCreated={() => { Toast.success(t("loop.toast.created")); onMutated(); }}
+      />,
+    );
+  };
+
   const isEmpty = view === "grouped" ? groups.every((g) => g.issues.length === 0) : total === 0;
 
   // 状态/优先级/项目的选项列表:单选与多选两个分支共用(避免重复 map)。
@@ -216,226 +222,175 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
     <Select.Option key={p.id} value={p.id}>{p.title}</Select.Option>
   ));
 
+  const title = defaultScope === "involves" ? t("loop.nav.myloop") : t("loop.nav.issue");
+
+  // 「筛选」按钮上的激活计数(仅计当前视图会生效的维度)。
+  const activeFilters =
+    (view === "grouped"
+      ? f.gStatuses.length + f.gPriorities.length + f.gProjectIds.length + (f.noProject ? 1 : 0)
+      : (f.status ? 1 : 0) + (f.priority ? 1 : 0) + (f.assignee ? 1 : 0) + (f.project ? 1 : 0) + (f.keyword.trim() ? 1 : 0)) +
+    (f.creator ? 1 : 0) +
+    (f.dateRange ? 1 : 0);
+
+  const clearFilters = () =>
+    update({
+      keyword: "", status: undefined, priority: undefined, assignee: undefined, creator: undefined, project: undefined,
+      gStatuses: [], gPriorities: [], gProjectIds: [], noProject: false, dateRange: undefined,
+    });
+
+  // 「筛选」面板：把原工具栏散落的下拉收进一个面板（对齐 multica 的筛选框）。维度按视图切换单选/多选。
+  const filterPanel = (
+    <div className="loop-fields loop-filter-panel">
+      <div className="loop-filter-panel__head">
+        <span>{t("loop.action.filter")}</span>
+        {activeFilters > 0 && (
+          <button type="button" className="loop-filter-panel__clear" onClick={clearFilters}>{t("loop.action.clearFilters")}</button>
+        )}
+      </div>
+      <div className="loop-fields__row">
+        <div className="loop-fields__label">{t("loop.filter.status")}</div>
+        {view === "grouped" ? (
+          <Select multiple value={f.gStatuses} onChange={(v) => update({ gStatuses: (v ?? []) as IssueStatus[] })} showClear maxTagCount={2} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
+        ) : (
+          <Select value={f.status} onChange={(v) => update({ status: v as IssueStatus | undefined })} showClear dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
+        )}
+      </div>
+      <div className="loop-fields__row">
+        <div className="loop-fields__label">{t("loop.filter.priority")}</div>
+        {view === "grouped" ? (
+          <Select multiple value={f.gPriorities} onChange={(v) => update({ gPriorities: (v ?? []) as IssuePriority[] })} showClear maxTagCount={2} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
+        ) : (
+          <Select value={f.priority} onChange={(v) => update({ priority: v as IssuePriority | undefined })} showClear dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
+        )}
+      </div>
+      {/* assignee 单选仅扁平列表/看板(grouped 用 scope 按类型收窄)。 */}
+      {view !== "grouped" && (
+        <div className="loop-fields__row">
+          <div className="loop-fields__label">{t("loop.filter.assignee")}</div>
+          <Select value={f.assignee} onChange={(v) => update({ assignee: v as string | undefined })} showClear filter dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.assignee")}>
+            {cands.map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>))}
+          </Select>
+        </div>
+      )}
+      {/* 「与我相关」自带 creator=我 并集腿,creator 下拉对它无效 → 隐藏。 */}
+      {!(view === "grouped" && scope === "involves") && (
+        <div className="loop-fields__row">
+          <div className="loop-fields__label">{t("loop.filter.creator")}</div>
+          <Select value={f.creator} onChange={(v) => update({ creator: v as string | undefined })} showClear filter dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.creator")}>
+            {cands.filter((c) => c.type === "member").map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>))}
+          </Select>
+        </div>
+      )}
+      {projects.length > 0 && (
+        <div className="loop-fields__row">
+          <div className="loop-fields__label">{t("loop.filter.project")}</div>
+          {view === "grouped" ? (
+            <Select multiple value={f.gProjectIds} onChange={(v) => update({ gProjectIds: (v ?? []) as string[] })} showClear filter maxTagCount={1} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
+          ) : (
+            <Select value={f.project} onChange={(v) => update({ project: v as string | undefined })} showClear filter dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
+          )}
+        </div>
+      )}
+      {/* no-project:仅分组板(后端 include_no_project 仅 /grouped 支持)。 */}
+      {view === "grouped" && (
+        <div className="loop-fields__row">
+          <Checkbox className="loop-nofilter" checked={f.noProject} onChange={(e) => update({ noProject: !!e.target.checked })}>{t("loop.filter.noProject")}</Checkbox>
+        </div>
+      )}
+      <div className="loop-fields__row">
+        <div className="loop-fields__label">{t("loop.filter.dateRange")}</div>
+        <div className="loop-fields__inline">
+          <Select value={f.dateField} onChange={(v) => update({ dateField: v as IssueDateField })} dropdownClassName="loop-fields__dropdown" style={{ width: 104 }}>
+            {ISSUE_DATE_FIELDS.map((d) => (<Select.Option key={d} value={d}>{t(`loop.dateField.${d}`)}</Select.Option>))}
+          </Select>
+          <DatePicker type="dateRange" density="compact" value={f.dateRange} onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })} placeholder={t("loop.filter.dateRange")} style={{ flex: 1 }} />
+        </div>
+      </div>
+      {/* 关键词走全文搜索(独立语义,不与 grouped 组合),故分组视图隐藏。 */}
+      {view !== "grouped" && (
+        <div className="loop-fields__row">
+          <div className="loop-fields__label">{t("loop.search.issue")}</div>
+          <Input className="loop-search" prefix={<Search size={14} />} placeholder={t("loop.search.issue")} value={f.keyword} onChange={(v) => update({ keyword: v })} showClear style={{ width: "100%" }} />
+        </div>
+      )}
+    </div>
+  );
+
+  // 「显示」面板：视图切换 + 列表排序。
+  const showPanel = (
+    <div className="loop-fields loop-show-panel">
+      <div className="loop-filter-panel__head"><span>{t("loop.action.show")}</span></div>
+      <div className="loop-fields__row">
+        <div className="loop-fields__label">{t("loop.view.board")}</div>
+        <div className="loop-seg" role="tablist">
+          {(["board", "grouped", "list"] as ViewMode[]).map((v) => (
+            <button key={v} type="button" role="tab" aria-selected={view === v} className={`loop-seg__btn${view === v ? " is-active" : ""}`} onClick={() => switchView(v)}>
+              {v === "board" ? <LayoutGrid size={14} /> : v === "grouped" ? <Users size={14} /> : <ListIcon size={14} />}
+              {t(`loop.view.${v}`)}
+            </button>
+          ))}
+        </div>
+      </div>
+      {view === "list" && (
+        <div className="loop-fields__row">
+          <div className="loop-fields__label">{t("loop.sort.direction")}</div>
+          <div className="loop-fields__inline">
+            <Select value={f.sortBy} onChange={(v) => update({ sortBy: v as IssueSortField })} dropdownClassName="loop-fields__dropdown" style={{ flex: 1 }}>
+              {ISSUE_SORT_FIELDS.map((s) => (<Select.Option key={s} value={s}>{t(`loop.sort.${s}`)}</Select.Option>))}
+            </Select>
+            <Button theme="borderless" disabled={f.sortBy === "position"} icon={f.sortDir === "asc" ? <ArrowUp size={14} /> : <ArrowDown size={14} />} aria-label={t("loop.sort.direction")} onClick={() => update({ sortDir: f.sortDir === "asc" ? "desc" : "asc" })} />
+          </div>
+        </div>
+      )}
+    </div>
+  );
+
   return (
     <div className="loop-page">
-      <div className="loop-page__head">
-        <Title heading={4}>{t("loop.nav.issue")}</Title>
-        <div className="loop-viewtoggle">
-          <button className={view === "board" ? "is-active" : ""} onClick={() => switchView("board")}>
-            <LayoutGrid size={14} />
-            {t("loop.view.board")}
-          </button>
-          <button className={view === "grouped" ? "is-active" : ""} onClick={() => switchView("grouped")}>
-            <Users size={14} />
-            {t("loop.view.grouped")}
-          </button>
-          <button className={view === "list" ? "is-active" : ""} onClick={() => switchView("list")}>
-            <ListIcon size={14} />
-            {t("loop.view.list")}
-          </button>
+      <div className="loop-page__head loop-page__head--stack">
+        <div className="loop-page__title-row">
+          <h2 className="loop-page__title">{title}</h2>
         </div>
-        {view === "grouped" && (
-          <div className="loop-seg" role="tablist">
-            {(["all", "members", "agents", "involves"] as IssueScope[]).map((s) => {
-              // 「与我相关」需当前成员的后端 id；未解析出则禁用（而非静默失效）。
-              const disabled = s === "involves" && !myMemberId;
-              return (
-                <button
-                  key={s}
-                  type="button"
-                  role="tab"
-                  aria-selected={scope === s}
-                  disabled={disabled}
-                  className={`loop-seg__btn${scope === s ? " is-active" : ""}`}
-                  onClick={() => setScope(s)}
-                >
-                  {t(`loop.scope.${s}`)}
-                </button>
-              );
-            })}
-          </div>
-        )}
-        <div className="loop-page__spacer" />
-        {view === "grouped" ? (
-          <Select
-            placeholder={t("loop.filter.status")}
-            multiple
-            value={f.gStatuses}
-            onChange={(v) => update({ gStatuses: (v ?? []) as IssueStatus[] })}
-            showClear
-            maxTagCount={2}
-            size="small"
-            style={{ minWidth: 150, maxWidth: 260 }}
-          >
-            {statusOpts}
-          </Select>
-        ) : (
-          <Select
-            placeholder={t("loop.filter.status")}
-            value={f.status}
-            onChange={(v) => update({ status: v as IssueStatus | undefined })}
-            showClear
-            size="small"
-            style={{ width: 130 }}
-          >
-            {statusOpts}
-          </Select>
-        )}
-        {view === "grouped" ? (
-          <Select
-            placeholder={t("loop.filter.priority")}
-            multiple
-            value={f.gPriorities}
-            onChange={(v) => update({ gPriorities: (v ?? []) as IssuePriority[] })}
-            showClear
-            maxTagCount={2}
-            size="small"
-            style={{ minWidth: 140, maxWidth: 240 }}
-          >
-            {priorityOpts}
-          </Select>
-        ) : (
-          <Select
-            placeholder={t("loop.filter.priority")}
-            value={f.priority}
-            onChange={(v) => update({ priority: v as IssuePriority | undefined })}
-            showClear
-            size="small"
-            style={{ width: 120 }}
-          >
-            {priorityOpts}
-          </Select>
-        )}
-        {/* assignee 单选筛选仅扁平列表/看板用(grouped 用 scope pill 按类型收窄)。 */}
-        {view !== "grouped" && (
-          <Select
-            placeholder={t("loop.filter.assignee")}
-            value={f.assignee}
-            onChange={(v) => update({ assignee: v as string | undefined })}
-            showClear
-            filter
-            size="small"
-            style={{ width: 150 }}
-          >
-            {cands.map((c) => (
-              <Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>
-            ))}
-          </Select>
-        )}
-        {/* 「与我相关」自带 creator=我 的并集腿,creator 下拉对它无效 → 隐藏,避免设了不生效。 */}
-        {!(view === "grouped" && scope === "involves") && (
-          <Select
-            placeholder={t("loop.filter.creator")}
-            value={f.creator}
-            onChange={(v) => update({ creator: v as string | undefined })}
-            showClear
-            filter
-            size="small"
-            style={{ width: 130 }}
-          >
-            {cands.filter((c) => c.type === "member").map((c) => (
-              <Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>
-            ))}
-          </Select>
-        )}
-        {projects.length > 0 && (view === "grouped" ? (
-          <Select
-            placeholder={t("loop.filter.project")}
-            multiple
-            value={f.gProjectIds}
-            onChange={(v) => update({ gProjectIds: (v ?? []) as string[] })}
-            showClear
-            filter
-            maxTagCount={1}
-            size="small"
-            style={{ minWidth: 150, maxWidth: 240 }}
-          >
-            {projectOpts}
-          </Select>
-        ) : (
-          <Select
-            placeholder={t("loop.filter.project")}
-            value={f.project}
-            onChange={(v) => update({ project: v as string | undefined })}
-            showClear
-            filter
-            size="small"
-            style={{ width: 140 }}
-          >
-            {projectOpts}
-          </Select>
-        ))}
-        {/* no-project:仅分组板(后端 include_no_project 仅 /grouped 支持),与项目多选配对
-            = 所选项目 ∪ 无项目。project 维度独立于 assignee,与 scope/involves 都能正确 AND。 */}
-        {view === "grouped" && (
-          <Checkbox
-            className="loop-nofilter"
-            checked={f.noProject}
-            onChange={(e) => update({ noProject: !!e.target.checked })}
-          >
-            {t("loop.filter.noProject")}
-          </Checkbox>
-        )}
-        <div style={{ display: "flex", alignItems: "center", gap: 4 }}>
-          {f.dateRange && (
-            <Select
-              value={f.dateField}
-              onChange={(v) => update({ dateField: v as IssueDateField })}
-              size="small"
-              style={{ width: 104 }}
-            >
-              {ISSUE_DATE_FIELDS.map((d) => (
-                <Select.Option key={d} value={d}>{t(`loop.dateField.${d}`)}</Select.Option>
-              ))}
-            </Select>
+        <div className="loop-page__toolbar">
+          {/* 作用域文字 tab —— grouped 视图按负责人类型收窄(all/成员/AI/与我相关)。 */}
+          {view === "grouped" && (
+            <div className="loop-agent-scope" role="tablist">
+              {(["all", "members", "agents", "involves"] as IssueScope[]).map((s) => {
+                const disabled = s === "involves" && !myMemberId;
+                return (
+                  <button
+                    key={s}
+                    type="button"
+                    role="tab"
+                    aria-selected={scope === s}
+                    disabled={disabled}
+                    className={`loop-agent-scope__btn${scope === s ? " is-active" : ""}`}
+                    onClick={() => setScope(s)}
+                  >
+                    {t(`loop.scope.${s}`)}
+                  </button>
+                );
+              })}
+            </div>
           )}
-          <DatePicker
-            type="dateRange"
-            size="small"
-            density="compact"
-            value={f.dateRange}
-            onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })}
-            placeholder={t("loop.filter.dateRange")}
-            style={{ width: 220 }}
-          />
+          <div className="loop-page__spacer" />
+          <Dropdown trigger="click" position="bottomRight" clickToHide render={filterPanel}>
+            <button className={`loop-toolbtn${activeFilters > 0 ? " is-on" : ""}`}>
+              <Filter size={14} />
+              {t("loop.action.filter")}
+              {activeFilters > 0 && <span className="loop-toolbtn__badge">{activeFilters}</span>}
+            </button>
+          </Dropdown>
+          <Dropdown trigger="click" position="bottomRight" clickToHide render={showPanel}>
+            <button className="loop-toolbtn">
+              <SlidersHorizontal size={14} />
+              {t("loop.action.show")}
+            </button>
+          </Dropdown>
+          <Button theme="solid" icon={<Plus size={14} />} onClick={openNewLoop}>
+            {t("loop.action.newIssue")}
+          </Button>
         </div>
-        {view === "list" && (
-        <div style={{ display: "flex", alignItems: "center", gap: 2 }}>
-          <Select
-            value={f.sortBy}
-            onChange={(v) => update({ sortBy: v as IssueSortField })}
-            size="small"
-            style={{ width: 120 }}
-          >
-            {ISSUE_SORT_FIELDS.map((s) => (
-              <Select.Option key={s} value={s}>{t(`loop.sort.${s}`)}</Select.Option>
-            ))}
-          </Select>
-          <Button
-            size="small"
-            theme="borderless"
-            // position(手动序)后端忽略方向,禁用切换。
-            disabled={f.sortBy === "position"}
-            icon={f.sortDir === "asc" ? <ArrowUp size={14} /> : <ArrowDown size={14} />}
-            aria-label={t("loop.sort.direction")}
-            onClick={() => update({ sortDir: f.sortDir === "asc" ? "desc" : "asc" })}
-          />
-        </div>
-        )}
-        {/* 关键词走全文搜索(独立语义,不与 grouped 组合),故分组视图隐藏。 */}
-        {view !== "grouped" && (
-          <Input
-            prefix={<Search size={14} />}
-            placeholder={t("loop.search.issue")}
-            value={f.keyword}
-            onChange={(v) => update({ keyword: v })}
-            showClear
-            style={{ width: 200 }}
-          />
-        )}
-        <Button theme="solid" icon={<Plus size={14} />} onClick={() => setCreateOpen(true)}>
-          {t("loop.action.newIssue")}
-        </Button>
       </div>
 
       <div className="loop-page__body">
@@ -448,7 +403,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
             <ClipboardList size={40} className="loop-empty__icon" />
             <div className="loop-empty__title">{t("loop.empty.issueTitle")}</div>
             <div className="loop-empty__desc">{t("loop.empty.issueDesc")}</div>
-            <Button theme="solid" icon={<Plus size={14} />} onClick={() => setCreateOpen(true)} style={{ marginTop: 12 }}>
+            <Button theme="solid" icon={<Plus size={14} />} onClick={openNewLoop} style={{ marginTop: 12 }}>
               {t("loop.action.newIssue")}
             </Button>
           </div>
@@ -472,12 +427,6 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
           </>
         )}
       </div>
-
-      <CreateIssueModal
-        visible={createOpen}
-        onClose={() => setCreateOpen(false)}
-        onCreated={() => { Toast.success(t("loop.toast.created")); onMutated(); }}
-      />
     </div>
   );
 }
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -200,11 +200,11 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
     WKApp.routeRight.push(<IssueDetailPage key={id} issueId={id} onChanged={onMutated} />);
   };
 
-  // 新建回路 → 唤起独立 composer 页（非弹窗，对齐 Figma）。创建成功后返回并刷新列表。
+  // 新建回路 → 唤起独立 composer 页（非弹窗，对齐 Figma）。创建成功后 pop 回列表并刷新。
   const openNewLoop = () => {
     WKApp.routeRight.push(
       <NewLoopPage
-        onCreated={() => { Toast.success(t("loop.toast.created")); onMutated(); }}
+        onCreated={() => { Toast.success(t("loop.toast.created")); onMutated(); WKApp.routeRight.pop(); }}
       />,
     );
   };
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -81,6 +81,11 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   const [scope, setScope] = useState<IssueScope>(defaultScope ?? "all");
   const [f, setF] = useState<Filters>({ keyword: "", gStatuses: [], gPriorities: [], gProjectIds: [], noProject: false, sortBy: "position", sortDir: "desc", dateField: "created_at" });
   const [page, setPage] = useState(0); // 0-based，仅列表视图分页
+  // 筛选/显示 面板受控可见:不用 clickToHide(它会在点击内部 Select/DatePicker/Checkbox
+  // 时冒泡关闭,且这些控件的 portal 选项面板点击也会误触发关闭),改为显式
+  // onVisibleChange —— 仅外部点击关闭,内部多字段交互保持面板打开。
+  const [filterOpen, setFilterOpen] = useState(false);
+  const [showOpen, setShowOpen] = useState(false);
   const cands = useAssigneeCandidates();
   // 当前 octo 成员的后端 user_id(involves_user_id 需 UUID,非 octo uid)：
   // 复用订阅特性的身份解析——候选里 octo_uid===loginInfo.uid 的 member。未解析出则「与我相关」不可用。
@@ -374,14 +379,14 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
             </div>
           )}
           <div className="loop-page__spacer" />
-          <Dropdown trigger="click" position="bottomRight" clickToHide render={filterPanel}>
+          <Dropdown trigger="click" visible={filterOpen} onVisibleChange={setFilterOpen} position="bottomRight" render={filterPanel}>
             <button className={`loop-toolbtn${activeFilters > 0 ? " is-on" : ""}`}>
               <Filter size={14} />
               {t("loop.action.filter")}
               {activeFilters > 0 && <span className="loop-toolbtn__badge">{activeFilters}</span>}
             </button>
           </Dropdown>
-          <Dropdown trigger="click" position="bottomRight" clickToHide render={showPanel}>
+          <Dropdown trigger="click" visible={showOpen} onVisibleChange={setShowOpen} position="bottomRight" render={showPanel}>
             <button className="loop-toolbtn">
               <SlidersHorizontal size={14} />
               {t("loop.action.show")}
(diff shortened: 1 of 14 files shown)


Jerry-Xin (dismissed):
One-line summary: The redesign is in-scope, but it removes existing issue-management paths and has a routing bug after sidebar-based creation.

🔴 Blocking

🔴 Critical — Detail page no longer preserves existing property editing functionality.  
The PR description says existing issue-tracking functionality is preserved and property edits move to the top-right menu, but the current menu only supports status, priority, assignee, subscribe, and delete. Project, labels, parent issue, start date, due date, and stage were previously editable and are now only displayed or absent. See `renderMoreMenu` in `packages/dmloop/src/panel/IssueDetailPage.tsx:575`, and the read-only project field at `packages/dmloop/src/panel/IssueDetailPage.tsx:1103`. Restore edit paths for the removed fields or intentionally remove them with product/API follow-up.

🔴 Critical — Users cannot create the first child issue anymore.  
The child-issue section, including its create button, is rendered only when `children.length > 0`, so an issue with no children has no visible way to create its first child. The modal still exists at `packages/dmloop/src/panel/IssueDetailPage.tsx:1176`, but no control opens it in the empty state. See `packages/dmloop/src/panel/IssueDetailPage.tsx:933`.

🔴 Critical — Creating from the sidebar can pop the newly replaced root view.  
`LoopPage.openNewLoop` passes `onCreated={() => { ... openTab("issue"); }}`, and `openTab` calls `replaceToRoot`. `NewLoopPage.submit` then immediately calls `back()`, which is `routeRight.pop()`. With `replaceToRoot` resetting the stack to the issue page, the subsequent pop can remove that root and leave the right pane empty. See `packages/dmloop/src/pages/LoopPage.tsx:80` and `packages/dmloop/src/pages/NewLoopPage.tsx:81`.

💬 Non-blocking

🟡 Warning — Clipboard calls can throw when `navigator.clipboard` is unavailable.  
`navigator.clipboard?.writeText(...).then(...)` still calls `.then` on `undefined` if the Clipboard API is absent. This affects comment copy and run transcript copy. See `packages/dmloop/src/panel/IssueDetailPage.tsx:711` and `packages/dmloop/src/panel/RunDetailModal.tsx:211`.

✅ Highlights

The list selection pruning and batch-operation reentry guard are solid, and the attachment upload paths correctly isolate per-file failures instead of aborting the whole operation.

OctoBoooot (approved):
## Review: redesign board, list, detail, run and new-loop views (#668)

**Verdict:** Approve
Large loop-view redesign that preserves behavior. Verified the filter/fetch logic is byte-identical to base, the batch/polling/suppress_run safety carried over, and adversarially checked every candidate finding from a fan-out — all refute. No blocker/major/minor survives. One pre-existing out-of-scope note.

Risk tier: normal. `needs-human-review` applied (14 files, +2365/-1179). Base `feat/octo-loop`; CI green (`code-review` pending, not failing). No backend/API change.

### What I verified
- **IssuePage filter/fetch logic unchanged (as the body claims).** The `reload` callback is byte-identical to base (same normalized hash) — the +169/-220 churn is JSX moving filters into 筛选/显示 popovers, not fetch behavior. So no new filter/scope bug introduced.
- **Batch safety preserved.** `IssueList` batch status/priority/assignee all still pass `suppress_run: true` (the #665 fix survived the redesign), batch-delete keeps `confirmDelete` + the `busyRef` re-entrancy guard.
- **RunDetailModal polling is safe.** `setTimeout` loop with a `cancelled` guard + cleanup, stops on terminal status (`!isActiveRun`) and on modal close/unmount (deps `[visible, run]`), incremental via `lastSeqRef`. No infinite-poll or setState-after-unmount.
- **No XSS.** Zero `dangerouslySetInnerHTML` in any changed file. Comment/description render via `LoopMarkdown` = `react-markdown@8` with **no `rehype-raw`**, so raw HTML in content is escaped, not executed; its `transformLinkUri` allowlists only `mention://` and delegates all other schemes to the default sanitizer (the correct custom-scheme pattern). Activity `actor_name` is plain JSX (`<strong>{a.actor_name}</strong>`) — React auto-escapes.
- **i18n parity holds** — en-US / zh-CN key sets match after the +83 additions; new `loop.run.event.*` / `loop.newLoop.*` keys present in both.
- **NewLoopPage is well-built.** Synchronous `Array.from(files)` capture before `input.value` reset (with the explanatory comment), submit guard `!text || submitting`, per-file upload failure non-blocking with a toast, `back()` on success.

### Findings adversarially verified and DROPPED (fan-out over-reported)
- **"XSS via comment/activity" (×2)** — refuted. Plain JSX auto-escapes; `LoopMarkdown` doesn't render raw HTML. No `dangerouslySetInnerHTML` anywhere.
- **"Double-submit race in comment/reply" (×2)** — refuted. Both `submitComment` (`:411`) and `ThreadReply.submit` (`:106`) guard on `submitting`/`busy` at entry AND the Send button is `loading={submitting}` (Semi disables a loading button). Not re-entrant within a click; a real second click requires a re-render by which point the guard/disable holds. Standard pattern used across the codebase.
- **"Enum-icon fallback gap"** — non-issue. `ISSUE_STATUS_ICON`/`PRIORITY_ICON` are `Record<IssueStatus,…>` / `Record<IssuePriority,…>` over **closed** literal-union types (7 statuses, 5 priorities), and every member has an entry — TypeScript guarantees exhaustiveness, so no `undefined`-component throw for a validly-typed value. (Unlike the backend-driven-open enums elsewhere; these are closed.) Base had no fallback either.

### Praise
- The redesign is genuinely behavior-preserving where it matters: the fetch/filter core is untouched (verified byte-identical), the #665 `suppress_run` batch fix and the run-poll terminal-stop are carried through intact, and the new `LoopMarkdown` keeps the safe react-markdown-no-raw-HTML posture with a correctly-scoped `mention://` allowlist. NewLoopPage handles the two classic composer traps well — synchronous FileList capture and a submit re-entrancy guard.

### Out of scope (informational, pre-existing — not introduced here)
- The "My loops" `scope="involves"` path still falls back to all-workspace loops when `myMemberId` is unresolved (no-`octo_uid` native members). Verified this PR does **not** touch that logic (fetch branch byte-identical to base; the `defaultScope="involves"` entry is unchanged base wiring) — so it's out of scope for #668. It's the pre-existing issue tracked on #663; flagging once so it isn't lost, not blocking here.

### Note
Given size/XL, deep verification was scoped to the fetch/filter provenance, batch/poll safety, XSS surface, and the new NewLoopPage (the risk/new-code); the CSS restyle (loop.css +488, issueDetail.css +422) was skimmed. `needs-human-review` applied — a maintainer should smoke-test the board/list/detail/run/new-loop flows.

packages/dmloop/src/panel/IssueDetailPage.tsx
```diff
@@ -774,382 +887,287 @@ export default function IssueDetailPage({ issueId, onChanged }: IssueDetailPageP
...
-                {t("loop.subIssue.create")}
-              </Button>
-            </div>
-            {children.length > 0 && (
-              <>
-                <Progress
-                  percent={Math.round((childrenDone / children.length) * 100)}
-                  style={{ margin: "8px 0 10px" }}
+          {/* 子回路：仅当确有子回路时才展示整个模块(空则不显示,避免突兀的孤零标题) */}
+          {children.length > 0 && (
```
OctoBoooot, line 906: **[blocker]** This gates the *entire* sub-issue module — title AND the create button at `:944` — behind `children.length > 0`. An issue with zero children then has no visible way to create its first child, even though `CreateIssueModal` (`:1176`) and `childCreateOpen` (`:181`) exist. Fix: keep the create button (or an empty-state CTA) visible when `children.length === 0`.

packages/dmloop/src/pages/LoopPage.tsx
```diff
@@ -77,6 +76,13 @@ export default function LoopPage() {
     WKApp.routeRight.replaceToRoot(renderTab(key, findWs(workspaces, wsId)));
   };
 
+  // 新建回路 → 唤起 composer 独立页；成功后落到回路看板（新回路即在其中）。
+  const openNewLoop = () => {
+    WKApp.routeRight.push(
+      <NewLoopPage onCreated={() => { Toast.success(t("loop.toast.created")); openTab("issue"); }} />,
```
OctoBoooot, line 82: **[blocker]** (with NewLoopPage.tsx:82) Routing pop-past-root. `openNewLoop` pushes NewLoopPage → stack `[issue, NewLoopPage]`. On success NewLoopPage calls `onCreated?.()` then `back()`. This `onCreated` runs `openTab("issue")` → `replaceToRoot(...)` (`:76`), collapsing the stack to a single root; then NewLoopPage's `back()` = `routeRight.pop()` pops that fresh root → empty right pane. Fix: don't `pop()` after an `onCreated` that already `replaceToRoot`s — drop the trailing `back()` on the success path, or let `onCreated` own navigation.

packages/dmloop/src/pages/NewLoopPage.tsx
```diff
@@ -0,0 +1,186 @@
...
+        description: text,
+        status: "todo",
+        priority: "none",
+        assignee_id: assigneeId,
+        assignee_type: assigneeType,
+        project_id: projectId,
+        attachment_ids: attachmentIds,
+      });
+      onCreated?.();
+      back();
```
OctoBoooot: **[blocker]** `back()` here runs unconditionally after `onCreated?.()`. When invoked from the sidebar, `onCreated` calls `openTab("issue")` → `replaceToRoot`, which resets the view stack to a single root; this `back()` (`routeRight.pop()`) then pops that root, leaving the right pane empty. See LoopPage.tsx:82. Fix: skip `back()` when `onCreated` already navigated via replaceToRoot.

OctoBoooot (changes requested):
## Review: #668 — correcting my verdict (Approve → Request changes)

Jerry-Xin/Steve caught three real regressions I missed. My fan-out checked XSS/races/lifecycle/polling but did **not** check feature-parity against base, affordance reachability, or the cross-file routing sequence — exactly the classes a "functionality preserved" redesign must be verified for. Byte-verified all three to the sink. Flipping.

Risk tier: normal; `needs-human-review` applied. Base `feat/octo-loop`.

### Blocker 1 — property edit paths removed (parity regression)
Base `IssueDetailPage` allowed editing `assignee, priority, status, start_date, due_date, parent, stage` (project too). New head's `patch({...})` covers only `assignee_id, description, priority, status, title` — verified `start_date` / `due_date` / `parent` / `stage` / `project_id` have **no edit path** (grep of `patch({<field>` returns false for each). The `renderMoreMenu` (`:575`) offers status/priority/assignee/subscribe/delete only; project is read-only (`:1103`). The PR says existing functionality is preserved — these edits are gone. Fix: restore the edit paths, or intentionally drop them with a product/API note. (Severity: real lost functionality; no data loss, so major-tier — but it contradicts the PR's parity claim and blocks on that basis.)

### Blocker 2 — cannot create the first child issue
`IssueDetailPage.tsx:933` gates the **entire** sub-issue module (title + create button at `:944`) behind `children.length > 0`. So an issue with zero children shows no control to create its first child. `CreateIssueModal` (`:1176`) and `childCreateOpen` (`:181`) exist, but the only opener is inside the `>0` block. The comment ("仅当确有子回路时才展示整个模块") confirms it's intentional — but it removes the empty-state create affordance. Fix: keep the create button (or an empty-state CTA) visible when `children.length === 0`.

### Blocker 3 — creating from the sidebar can blank the right pane
Sequence: `LoopPage.openNewLoop` (`:81`) `routeRight.push(<NewLoopPage>)` → stack `[issue, NewLoopPage]`. On success `NewLoopPage.submit` (`:81-82`) calls `onCreated?.()` **then** `back()`. `onCreated` = `openTab("issue")` → `replaceToRoot(...)` (`LoopPage:76`) collapses the stack to a single root `[issue]`; then `back()` = `routeRight.pop()` (`NewLoopPage:81`) pops that fresh root → **empty right pane**. The `replaceToRoot`-then-unconditional-`pop` ordering is the bug. Fix: don't `back()` after an `onCreated` that already `replaceToRoot`s (e.g. drop the `back()` on the success path, or have `onCreated` own the navigation without a following pop).

### Non-blocking (confirmed)
- **Clipboard throw** (`IssueDetailPage.tsx:711`, `RunDetailModal.tsx:211`): `navigator.clipboard?.writeText(x).then(...)` — the `?.` yields `undefined` when the Clipboard API is absent (non-HTTPS/older browsers), then `.then()` on `undefined` throws. Guard the whole call, not just the property access. Minor.

### Standing (verified, correct — unaffected by the above)
Filter/fetch logic byte-identical to base; batch `suppress_run` preserved; RunDetailModal polling stops on terminal/close/unmount; no XSS (LoopMarkdown safe, JSX auto-escaped); i18n parity holds; NewLoopPage's FileList capture + submit guard are correct.

### Accountability
I APPROVED and missed all three. Root cause: my fan-out was scoped to correctness/security/lifecycle dimensions but not to **feature-parity** (diff base→head for *removed* capabilities), **affordance reachability** (is every action still reachable in every state — esp. empty states), or **cross-file navigation sequences** (push/replaceToRoot/pop ordering across two files). For a redesign that claims "functionality preserved," those three are the primary risk and I under-scoped the review. Lesson (a sharper form of the #663/#667 provenance rule): parity claims require a base→head capability diff, and routing/empty-state affordances need explicit tracing. Credit: Jerry-Xin/Steve — all three byte-verified real. The security/lifecycle parts of my original review stand; the verdict was wrong and is corrected.

Jerry-Xin (dismissed):
## 🔴 REQUEST_CHANGES (corrected/re-scoped) @ 76aad2ca32bc

Re-scoping my earlier review after byte-verifying all findings against base↔head (`git merge-base feat/octo-loop <head>` → `76aad2ca`). Net: **one real blocker holds (narrower than my first pass), two earlier flags are withdrawn.** The redesign itself (board/list/detail/run/new-loop) is solid and the navigation refactor is well done.

### 🔴 Blocking (1) — attribute editors dropped from the detail view (regression vs "functionality preserved")
The old right sidebar (`loop-idp__aside`) was editable; the new one is explicitly read-only (`右侧属性栏（只读展示…；改属性走右上角 ⋯ 菜单）`). Some editors correctly **relocated** to the `···` menu — **status / priority / assignee** are all still editable there (so those are fine). But five attributes lost **all** edit affordance and were **not** relocated anywhere:

- **labels** (was `LabelEditor`)
- **parent_issue_id** (was a Select)
- **start_date** (was a DatePicker)
- **due_date** (was a DatePicker)
- **stage** (was an InputNumber)

Grepped every `packages/dmloop/**/*.tsx` at head: no `LabelEditor` import, no `parent_issue_id` / `start_date` / `due_date` / `stage` write anywhere (the only remaining `DatePicker` is the list **filter** date-range in IssuePage, not an editor). `LabelEditor.tsx` still exists but is no longer referenced by the detail page. Repro: open any issue → you can no longer set labels, parent, start/due date, or stage — all editable before this PR. Reachable, in-scope, PR-introduced. Please relocate these into the `···` menu (or a lightweight edit surface) the same way status/priority/assignee were, or confirm intentional removal in the PR description if the product is dropping them.

### 🟡 Non-blocking
- **Sub-issue empty-state has no inline create button.** The inline section-level `+` is gated behind `children.length > 0` (head ~L937/948). This is only a UX nit, **not** a functional block: creating the first sub-issue still works via the always-visible `···` menu → "Create sub-issue" (`setChildCreateOpen(true)`, ungated, head ~L640-642) → `CreateIssueModal` with `parentIssueId`. Consider adding a "+ Add sub-issue" affordance to the empty state for discoverability. (I withdraw my earlier "cannot create first child" blocker — refuted; the menu path is ungated.)
- **Clipboard optional-chaining throws in insecure context.** `navigator.clipboard?.writeText(...).then(...)` at IssueDetailPage.tsx:711 and RunDetailModal.tsx:211 — when `navigator.clipboard` is undefined (http / non-secure context), `?.` short-circuits `writeText` to `undefined`, then `.then` on `undefined` throws. Guard the whole call or check `navigator.clipboard` first.

### Withdrawn from my first pass (byte-verified, not defects)
- **"Empty right pane after sidebar create" — refuted.** `NewLoopPage.submit` calls only `onCreated?.()` and deliberately does **not** call `back()`/`pop()` — with an explicit comment that this is to avoid the caller's `replaceToRoot` stacking with a pop and popping the new root away (the exact blank-panel mode). `LoopPage.openNewLoop`'s `onCreated` does `openTab("issue")` → `replaceToRoot(IssuePage)`, landing on the board; the `IssuePage.openNewLoop` caller uses `pop()` to reveal the list beneath. Both callers land correctly, not blank.
- **status/priority/assignee "removed" — refuted.** They moved to the `···` menu and remain editable.

### Highlights
Clean board/list/detail/run/new-loop redesign; the `NewLoopPage` nav-ownership refactor correctly anticipates and prevents the replace-then-pop blank-panel bug; no backend/API changes.

Verdict: **REQUEST_CHANGES** on the dropped labels/parent/start_date/due_date/stage editors only. Clears once those are reachable again (or the removal is confirmed intentional).

OctoBoooot (changes requested):
## Review: #668 — delta re-review (2 of 3 blockers fixed; property-editor regression remains)

**Verdict:** Request changes (holding — one blocker unresolved)
The new commit (`76aad2ca`, "fix loop detail child-create, new-loop routing and clipboard copy") fixes blockers #2 and #3 plus the clipboard 🟡 — byte-verified. But blocker #1 (removed property editors) is untouched and still open, so the verdict holds. Converges with Jerry-Xin.

Risk tier: normal; `needs-human-review` stands. Base `feat/octo-loop`.

### Resolved — Blocker 2 (create first child): fixed
The child-create action is now a `Dropdown.Item` in the always-rendered `renderMoreMenu` (`IssueDetailPage.tsx:640`, → `setChildCreateOpen(true)`), no longer gated by `children.length > 0`. A childless issue can create its first child via the top-right menu. (Jerry-Xin's remaining 🟡 — no *inline* empty-state button, a discoverability nit — is fair and non-blocking.)

### Resolved — Blocker 3 (sidebar-create blanks the pane): fixed
Two-part, correct: `NewLoopPage` drops its unconditional `back()` (caller now owns navigation, with an explaining comment), and `IssuePage.onCreated` does a single `WKApp.routeRight.pop()` (it uses `onMutated()`, not `replaceToRoot`, so one pop cleanly returns to the list). No more replaceToRoot-then-pop-past-root → no empty pane.

### Resolved — clipboard 🟡: fixed
`navigator.clipboard?.writeText(x)?.then(...)` — the added `?.` after `writeText(x)` short-circuits when the API is absent, at both sites (`IssueDetailPage.tsx:712`, `RunDetailModal.tsx:208`).

### Still open — Blocker 1: property edit paths removed
Unchanged at this head: `patch({...})` still covers only `assignee_id / description / priority / status / title`. Verified **no** edit path for `start_date`, `due_date`, `parent`, `stage`, `label` (grep of `patch({<field>` false for each) — all editable in base `IssueDetailPage`, now display-only or absent. The commit didn't target this and it remains a parity regression against the PR's "functionality preserved" claim. Fix: restore the edit paths, or intentionally drop them with a product/API note stating the removal is deliberate.

### Standing (verified, correct)
Filter/fetch logic byte-identical to base; batch `suppress_run` preserved; RunDetailModal polling stops on terminal/close/unmount; no XSS; i18n parity; NewLoopPage FileList capture + submit guard correct.

### Note
Good, focused fix commit — 3 of 4 items closed with exactly the right changes (the routing fix in particular correctly moves navigation ownership to the caller). Only the property-editor restoration remains before this clears. Credit: Jerry-Xin/Steve for the original three; converged with Jerry-Xin that #1 is the sole remaining blocker.

yujiawei (dismissed):
# Code Review — PR #668 (octo-web)

Reviewed at head `dc2ee08896a49c578fc41c22d27ed011730a917c` (not the originally-opened SHA — the two follow-up commits `76aad2ca` and `dc2ee088` are included in this pass).

Scope: pure UI restyle of the loop (board / list / detail / run / new-loop) views under `packages/dmloop`, claiming no backend/API change and full functional parity. The redesign itself is well executed — the async navigation-race guards (`reqRef` token pattern), batch `suppress_run` safety, and `RunDetailModal` incremental polling all carried over correctly, and the property-editor regression flagged in earlier rounds is now resolved (label / parent / start & due date / stage editing restored via the new "edit properties" dialog at `IssueDetailPage.tsx`). Verified no `dangerouslySetInnerHTML` and no raw-HTML markdown sink was introduced.

One functional blocker remains.

## 🔴 Blocker (P1) — the new filter / display popovers close on any interior click, making them unusable

`packages/dmloop/src/pages/IssuePage.tsx:377` and `:384`:

```tsx
<Dropdown trigger="click" position="bottomRight" clickToHide render={filterPanel}>
  ...
<Dropdown trigger="click" position="bottomRight" clickToHide render={showPanel}>
```

Both popovers wrap **interactive form controls** — `filterPanel` is entirely `Select` / `Input` / `Checkbox` / `DatePicker` (`IssuePage.tsx:242-317`), and `showPanel` holds the view segmented control plus a sort `Select` (`:320-346`). With `clickToHide`, Semi binds a click handler on the popup container that fires on **any** click that bubbles to it:

- `@douyinfe/semi-ui/tooltip` — `handlePortalInnerClick = e => { if (this.props.clickToHide) { this.foundation.hide(); } ... }`, wired as the `onClick` of the `semi-portal-inner` wrapper.
- The inner controls do not stop propagation of that click: Semi's `Select` trigger (`select/foundation.js` `handleClick`) and `Input` (`input/foundation.js` `handleClick`) only open/focus themselves and let the event bubble.

Net effect: the moment the user clicks a status/priority/assignee `Select`, focuses the search `Input`, ticks the no-project `Checkbox`, or opens the date range, the whole popover unmounts — so filters and the sort/view controls cannot be operated at all. This is a full break of the redesigned filter/display UI (the pre-redesign toolbar rendered these controls inline, so it did not have this problem).

This is confirmed by the repo's own convention: `packages/dmloop/src/ui/LabelEditor.tsx:82` deliberately omits `clickToHide` with the comment "clickToHide 默认 false：勾选多个标签时下拉保持打开" — i.e. the same footgun was already known and avoided when a dropdown holds interactive content.

Fix: drop `clickToHide` on these two dropdowns (matching `LabelEditor`) and close them via explicit affordance / outside-click, or stop propagation on the panel container. Command-menu dropdowns elsewhere in the diff (the `···` menu, comment actions, automation card) legitimately keep `clickToHide` because their content is click-to-select `Dropdown.Item`s — only these two form-bearing popovers are affected.

## 🟡 Non-blocking

**P2 — emoji reactions dropped without mention (parity note).** The base detail view had issue-level and comment-level emoji reactions (`renderReactionBar`, `reactIssue`/`reactComment`, `addIssueReaction`/`addCommentReaction`); the redesigned page removes all reaction UI and no longer imports the reaction APIs. The PR body states existing functionality is preserved and does not call this out. This is likely intentional given the base code framed its picker as a temporary capability check ("full picker left to the UI redo"), but it should be stated explicitly (drop-with-follow-up) rather than silently omitted, since the PR asserts parity.

**P2 — `deriveTitle` can split a surrogate pair.** `packages/dmloop/src/pages/NewLoopPage.tsx:22` uses `line.slice(0, 80)`, which operates on UTF-16 code units; an emoji/astral char straddling index 80 yields a lone surrogate. Prefer `Array.from(line).slice(0, 80).join("")`. Low impact (title is derived, editable downstream), noting for correctness.

## Reviewer legs

Cross-family advisory: gemini-3.1-pro leg independently flagged the `clickToHide` issue (surfaced as its top finding); its `ThreadReply` busy-lock concern was checked and refuted (`replyToThread` wraps every await and returns `false` rather than throwing, so `setBusy(false)` always runs). The line-level advisory leg was unavailable this pass (tooling error, recorded as absent — not treated as approval); its coverage (line-level correctness, cross-file consumer checks) was performed manually: verified the `RUN_STATUS_COLOR`→`RUN_STATUS_HEX` and other `meta.tsx` export renames have no orphaned consumers, `formatRelativeTime`/`format` and `AutoGrowTextarea`/`time.ts` all pre-exist on the base branch, and en-US/zh-CN i18n keys for all new strings are present and at parity.

packages/dmloop/src/pages/IssuePage.tsx
```diff
@@ -216,226 +222,175 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps
...
+                    onClick={() => setScope(s)}
+                  >
+                    {t(`loop.scope.${s}`)}
+                  </button>
+                );
+              })}
+            </div>
+          )}
+          <div className="loop-page__spacer" />
+          <Dropdown trigger="click" position="bottomRight" clickToHide render={filterPanel}>
```
OctoBoooot: **[blocker]** `clickToHide` on a multi-control filter form makes it unusable. `filterPanel` (`:242`) holds 8+ Selects + a Checkbox + a date-range DatePicker + a search Input (and `showPanel` at `:384`/`:320` has view-switch + sort). With Semi's `clickToHide`, any mousedown inside the panel bubbles up and closes it — so opening a Select / toggling the checkbox / picking a date dismisses the whole panel, and you can't operate the filters. PR-introduced: base used an inline toolbar (zero `clickToHide`, no filterPanel/showPanel). Fix: drop `clickToHide` on these two panels and dismiss on outside-click / explicit close (or stopPropagation from the panel body). NB: the detail-page single-select property popups (`IssueDetailPage.tsx:1061+`) are correct to keep `clickToHide` — this is specifically the multi-field filter/display panels.

OctoBoooot (changes requested):
## Review: #668 — delta re-review (blocker #1 fixed; a new filter-popover blocker surfaced)

**Verdict:** Request changes (holding — my #1 is fixed, but yujiawei surfaced a real PR-introduced blocker I missed)
The `dc2ee088` commit restores all 5 property editors (my remaining blocker — resolved, byte-verified). But yujiawei correctly caught that the redesign's **filter/display popovers close on any inner click**, making the multi-field filter form unusable. Byte-verified as real and PR-introduced. Verdict holds.

Risk tier: normal; `needs-human-review` stands. Base `feat/octo-loop`.

### Resolved — Blocker #1 (removed property editors): fixed
`dc2ee088` ("restore label/parent/date/stage editing via detail properties dialog") + `02b641f7` restore edit paths for all five. `patch({...})` now covers the full set: `assignee_id, description, priority, status, title, project_id, parent_issue_id, start_date, due_date, stage`. Verified each of the previously-missing fields (`start_date`/`due_date`/`parent`/`stage`/`label`) has an edit path. Parity restored.

### New blocker — filter/display popovers are unusable (`clickToHide` on multi-control forms)
`IssuePage.tsx:377/384`: the 筛选 (filter) and 显示 (display) panels are `Dropdown trigger="click" ... clickToHide render={filterPanel/showPanel}`. But `filterPanel` (`:242`) is a **multi-control form** — 8+ `Select`s (status/priority/assignee/creator/project), a `Checkbox`, a date-range `DatePicker`, a search `Input`; `showPanel` (`:320`) has view-switch + sort. With Semi's **`clickToHide`, any mousedown inside the panel bubbles up and closes it** — so opening a Select / toggling the checkbox / picking a date dismisses the whole panel. Net: you can't actually operate the filters (yujiawei's "点击内部控件即关闭，导致无法操作"). Reachable on every filter interaction.

**PR-introduced (provenance):** base `IssuePage` had **zero** `clickToHide` and no `filterPanel`/`showPanel` — it used an inline toolbar. This PR's redesign moved the filters into popovers and applied `clickToHide` (correct for a single-select menu, wrong for a multi-field form). Fix: drop `clickToHide` on these two and dismiss on outside-click / an explicit close, or stop propagation from the panel body — mirroring how a filter popover should behave (interact freely, close on outside click). (Note: the detail-page inline property popups — status/priority/project at `IssueDetailPage.tsx:1061+` — are single-select menus where `clickToHide` is correct; this is specifically the multi-control filter/display panels.)

### Non-blocking (yujiawei's 🟡s — both real)
- **Emoji reactions dropped** — base had ~51 reaction/emoji refs, head ~7; the comment-reaction feature was largely removed in the redesign. Same "redesign drops a capability" pattern as #1 — secondary feature so 🟡, but the removal should be called out in the PR description (or restored if unintended).
- **Surrogate-pair slice** — `deriveTitle` `line.slice(0,80)` (`NewLoopPage.tsx:22`) and avatar-initial `.slice(0,1)` (`IssueDetailPage.tsx:672/1158`) can split an astral code point (emoji / CJK-ext) → mojibake. Cosmetic; use `Array.from(str).slice(...)` for code-point-safe truncation. Nit.

### Accountability
I missed the filter-popover blocker in my original review — I verified the fetch/filter *logic* was byte-identical to base (it was) but didn't exercise the new *popover interaction* the redesign wrapped around it. Same miss-class as the #668 detail-page regressions: a redesign's new UI-interaction surface, not the preserved logic. Lesson reinforced: for a redesign that moves existing controls into new containers, verify the container's interaction model (open/close/propagation), not just that the underlying logic is unchanged. Credit: yujiawei. Blocker #1 is genuinely fixed; this new one holds the gate.

Jerry-Xin (dismissed):
## 🔴 REQUEST_CHANGES (re-review) @ 02b641f7b69e

Re-review after the new commit. My prior blocker is **fixed**, but a new blocker was introduced by the filter redesign, and there's an undocumented feature removal to confirm.

### ✅ Resolved from my prior CR
**The 5 dropped attribute editors are restored.** `IssueDetailPage.tsx` now wires all of them to `patch({...})`: `LabelEditor` (L1236-1237), parent-issue Select (L1242-1243), start_date / due_date DatePickers (L1259-1273), stage InputNumber (L1280-1283). Bonus: `patch` now preserves `labels`/`reactions`/`attachments` on re-enrich (L261-265) so edits don't blank them. Property-edit parity is back. Thanks for the fast turnaround.

### 🔴 Blocking (new, introduced by this commit's filter redesign)
**Filter & "Show" popovers close when you interact with their inner controls → the multi-field filter form is not usable.** Both panels render inside `<Dropdown trigger="click" ... clickToHide render={filterPanel/showPanel}>` (IssuePage.tsx:377, :384). The panel content (IssuePage.tsx:242-315 / 320-375) is a plain `<div>` with **no `onClick`/`stopPropagation` guard**. Semi-UI `clickToHide` hides the dropdown on any click bubbling up from the popup content — so:
- clicking an inner `Select` / `DatePicker` / `Checkbox` / the segmented view buttons / clear-filters / sort-toggle bubbles to `clickToHide` and collapses the whole panel;
- `Select`/`DatePicker` render option panels in separate portals, so opening them or picking an option also trips the close.

Net: users can't set more than one field without the panel snapping shut. Reachable on the primary filter UI; PR-introduced (base `IssuePage` had no Dropdown-based filter — this redesign moved filters into popovers). Fix: drop `clickToHide` and control visibility explicitly (`visible`/`onVisibleChange`, closing only on outside-click / a Done button), or wrap the panel content with `onClick={e => e.stopPropagation()}` so inner interactions don't bubble to the hide handler.

### 🟡 Non-blocking — but please confirm the intent
**The comment/issue emoji-reaction feature was removed, undocumented.** This commit deletes the whole reaction affordance: `addCommentReaction`/`removeCommentReaction`/`addIssueReaction`/`removeIssueReaction` imports, `REACTION_EMOJIS`, the `reactComment`/`reactIssue` handlers, and `renderReactionBar`. The data model still carries `reactions` (preserved on re-enrich), so it's the UI + write path that's gone, not the data. On a "functionality preserved" redesign this is a real capability loss — please either restore a reaction affordance or state in the PR description that reactions are intentionally dropped (and whether the API/data will follow). Flagging rather than blocking since it may be a deliberate scope call, but it shouldn't land silently.

### 🟡 Non-blocking (minor)
**Surrogate-pair split in avatar initials.** `(c.author_name ?? "?").slice(0, 1)` (IssueDetailPage.tsx:672, :1158) — a name starting with an astral-plane char (emoji/rare CJK) splits the surrogate pair and renders a broken glyph. Use `[...str][0]` or `Array.from(str)[0]`. Cosmetic (avatar initial only).

### Verdict
**REQUEST_CHANGES** on the filter/show popover `clickToHide` close-on-inner-interaction (real, reachable, PR-introduced). The editor-parity blocker is cleared; the emoji-reaction removal needs a documented decision; the surrogate slice is a nit. Fix the popover close behavior and this clears.

superalsrk:
已推送修复 commit \`3a3270ac\`,回应上一轮 review(02b641f7):

**🔴 Blocking — filter/show popover clickToHide(已修)**
筛选/显示两个 Dropdown 去掉 \`clickToHide\`,改为受控 \`visible\`/\`onVisibleChange\`。现在点击面板内部的 Select / DatePicker / Checkbox / 分段按钮(以及它们的 portal 选项面板)不再冒泡关闭面板,仅外部点击关闭,多字段筛选可连续设置。

**🟡 Non-blocking — emoji-reaction 移除(有意 drop,现声明)**
本次 loop redesign **有意移除** comment/issue 的 emoji-reaction UI 与写路径(\`react*\` handlers / \`REACTION_EMOJIS\` / \`renderReactionBar\`)。数据模型的 \`reactions\` 字段**保留**(re-enrich 时不丢),仅前端交互暂缓。理由:本轮聚焦看板/列表/详情/运行记录的重构,reaction 作为独立交互后续按需单独恢复。此为明确的 scope 决策,非误删。

**🟡 Nit — avatar 首字母 surrogate-pair(已修)**
两处头像首字母改用 \`[...str][0]\`,astral-plane 起始字符(emoji/生僻 CJK)不再断裂。

请复审。

OctoBoooot (approved):
## Review: #668 — delta re-review (filter-popover blocker fixed)

The new commit fixes the filter/display popover blocker with controlled visibility (the prescribed fix) and resolves the surrogate-slice 🟡. Byte-verified at head `3a3270ac0379`. Flipping.

Risk tier: normal; `needs-human-review` stands. Base `feat/octo-loop`.

### Resolved — filter/display popover `clickToHide` blocker
`IssuePage.tsx`: both panels switched from `clickToHide` to **controlled visibility** — `<Dropdown trigger="click" visible={filterOpen} onVisibleChange={setFilterOpen} …>` (and `showOpen`). A controlled Semi `Dropdown` closes only on outside-click / trigger-click, not on clicks bubbling from inner content, so the multi-field filter form (Selects / DatePicker / Checkbox, including their portal option panels) stays open during interaction. The inline comment names the exact failure mode it fixes (portal option clicks). This is the correct fix — interact freely, close on outside click.

### Resolved — surrogate-pair avatar initials (🟡)
`(name ?? "?").slice(0, 1)` → `[...(name ?? "?")][0]` at both sites (comment author `:669`, sidebar creator `:1155`). Spread iterates by code point, so an astral first character (emoji / CJK-ext) yields the whole glyph instead of a split surrogate half.

### Standing (verified across prior rounds)
- Blocker #1 (5 removed property editors) — fixed earlier by `dc2ee088` (labels/parent/start_date/due_date/stage all editable via `patch`).
- Blockers #2 (first-child-create) and #3 (sidebar-create blank pane) — fixed by `76aad2ca`.
- Filter/fetch logic byte-identical to base; batch `suppress_run` preserved; RunDetailModal polling correct; no XSS; i18n parity.

### Remaining (non-blocking 🟡, unchanged)
- **Emoji-reaction removal** — the redesign dropped the comment/issue reaction UI+write paths (data layer retained). Not addressed by this commit (nor claimed). Still worth documenting in the PR description whether the removal is intentional (or restoring the entry) — a real capability delta in a "functionality preserved" redesign, but secondary and non-blocking.

### Accountability
This closes the last blocker on #668. Over the PR's life I flipped twice on evidence: APPROVE → CR when the feature-parity/reachability/routing regressions surfaced (property editors, first-child, routing) and the filter-popover interaction bug (the latter I missed originally and yujiawei caught), then CR → APPROVE as each was fixed. The recurring lesson from this PR: a "functionality preserved" redesign needs its new UI-interaction surfaces (empty-state affordances, popover open/close/propagation, routing sequences) verified, not just that the underlying logic is unchanged. All blockers now resolved; only the document-the-reaction-removal 🟡 remains as author follow-up. Credit: Jerry-Xin/Steve + yujiawei across the rounds.

Jerry-Xin (approved):
Summary: The PR is in scope for `octo-web` and has no blocking correctness or security issues from this review, but there are a few user-visible regressions and convention gaps worth fixing before merge.

💬 Non-blocking

🟡 Warning — The new-loop composer can create a non-dispatched, unassigned loop. `assigneeId`/`assigneeType` start as `null`, the submit button is only gated by `prompt.trim()`, and `createIssue` sends `assignee_id: null` / `assignee_type: null` with `status: "todo"`. That conflicts with the “dispatch” UX and can leave a newly created loop with no executor. See `packages/dmloop/src/pages/NewLoopPage.tsx:35`, `packages/dmloop/src/pages/NewLoopPage.tsx:71`, `packages/dmloop/src/pages/NewLoopPage.tsx:153`.

🟡 Warning — The PR description says list row click opens detail, but only the title button is clickable. The row container has no `onClick`, so clicking whitespace, identifier, time, project, etc. does nothing. See `packages/dmloop/src/panel/IssueList.tsx:172` and `packages/dmloop/src/panel/IssueList.tsx:179`.

🟡 Warning — Attachment-only comments appear enabled but do not submit. The new-comment send button enables when `pendingFiles.length > 0`, but `submitComment` returns early when `commentDraft.trim()` is empty. Thread replies also allow files-only input, then call `addComment` with an empty content string. See `packages/dmloop/src/panel/IssueDetailPage.tsx:436`, `packages/dmloop/src/panel/IssueDetailPage.tsx:506`, `packages/dmloop/src/panel/IssueDetailPage.tsx:1041`.

🔵 Suggestion — The touched styling adds many hardcoded color values despite the project CSS guidance preferring semantic tokens. Examples are the status/priority hex maps and transcript CSS variables. Consider moving these to shared semantic/component tokens or using existing Semi token names consistently. See `packages/dmloop/src/ui/meta.tsx:79`, `packages/dmloop/src/ui/meta.tsx:106`, `packages/dmloop/src/panel/issueDetail.css:452`.

✅ Highlights

The detail page keeps stale-request guards for issue reloads, which is important for right-panel navigation races.

Batch operations retain a synchronous `busyRef` guard, which avoids duplicate writes from repeated confirmation actions.

Locale key parity between `en-US.json` and `zh-CN.json` is intact by direct JSON comparison.

Verification note: `git diff --check` passed. I could not run `tsc` or `pnpm i18n:check` because dependencies are not installed in this workspace; both failed due missing `typescript` / `node_modules`.

OctoBoooot (commented):
## Review: #668 — confirming Jerry-Xin's 🟡 #3 (attachment-only comment), holding Approve

**Verdict unchanged: Approve.** All four of Jerry-Xin's items are non-blocking 🟡s. Recording one byte-verified detail on #3 since it's a real (if minor) inconsistency with a mild UX trap.

### 🟡 #3 — main comment composer can't submit attachment-only (confirmed, asymmetric with replies)
There's a real asymmetry between the two composers:
- **ThreadReply** (`:113-115`): `if ((!c && files.length === 0) || busy) return` + button `disabled={!draft.trim() && files.length === 0}` → **allows files-only** submit. Correct.
- **Main new-comment composer** (`submitComment`, `:436-438`): `const content = commentDraft.trim(); if (!content || submitting) return;` → **blocks on empty text regardless of attachments**. But its button (`:1041`) is `disabled={!commentDraft.trim() && pendingFiles.length === 0}` — so with files-only the button is **enabled**, yet clicking it hits `if (!content) return` and **silently no-ops**.

So a user who attaches a file with no text on the main composer sees an enabled Send button that does nothing (the reply box, by contrast, works). Minor — attachment-only comments are an edge case and the reply path handles them — but the enabled-button-that-no-ops is a small trap. Fix: align `submitComment`'s guard with ThreadReply's (`if ((!content && pendingFiles.length === 0) || submitting) return`) so files-only submits, or disable the button when there's no text.

### The other three 🟡s (non-blocking, agree)
- **#1 new-loop executor-less task** — NewLoopPage permits a null assignee (assign later); valid draft state / design choice.
- **#2 row-click only on the title button** — UX/discoverability nit consistent with the detail redesign.
- **#4 hardcoded colors → tokens** — the recurring inline-style note (JSX `CSSProperties`, not token-governed `.css`).

Net: verdict holds at Approve — all blockers across #668's life are fixed (property editors, first-child, routing, filter-popover); these are polish follow-ups plus the standing emoji-reaction-removal-documentation 🟡. Credit: Jerry-Xin — #3 is a genuine composer asymmetry worth aligning.

yujiawei (changes requested):
# Code Review — PR #668 (octo-web)

Reviewed at head `3a3270ac0379393529eee2b630adfe99a187858f` against base `feat/octo-loop` (merge-base `9b8c0761`). 14 files, +2540/−1232. Scope: UX redesign of the loop board / list / detail / run / new-loop views; no backend/API changes.

This is a large, carefully-commented redesign and most of it is solid: the async `reqRef` token discipline is preserved across the new code paths, the `FileList` synchronous-capture pattern is applied consistently (avoiding the multi-select drop bug), the activity-feed merge/sort/grouping and the "latest block open by default" logic are correct, batch operations still force `suppress_run`, pagination is preserved, and the removed features (comment editing, emoji reactions, nested replies flattened to one level) are documented in the PR description as intentional.

The review is split into two verdicts: **spec compliance** first, then **code quality**.

---

## 1. Spec Compliance

**Spec: ✅**

Every item in the PR description is implemented, and I found nothing built beyond the stated scope:

- **Board** — status-icon column headers with counts + faint tracks; cards show priority/status icons, mono id, relative time, title, project, neutral assignee badge; drag-to-status preserved. ✅
- **List** — table replaced with status-grouped rows; row click opens detail; multi-select batch bar (status/priority/assignee with `suppress_run`, delete) + inline assignee edit with run pre-confirm; pagination preserved. ✅
- **Header/filters** — 筛选 + 显示 popovers hold all existing filters; filter logic unchanged (controlled `visible`/`onVisibleChange` rather than `clickToHide`, with a clear rationale). ✅
- **Detail** — breadcrumb topbar; native single/multi-line inputs; unified activity feed with collapsible "N activities" blocks; flat comment threads with per-card reply box + multi-attachment chips; three-dot row menu (copy/resolve/delete); read-only sidebar (Properties/Details/Execution log) with edits via the ⋯ menu; active runs always visible, terminal runs behind "show history (N)"; sub-issue module only when children exist. ✅
- **Run modal** — rebuilt as a transcript (status badge, metadata chips, timeline bar, collapsible events) on the same polling source. ✅
- **New loop** — standalone routed composer mapping to the existing create-issue + upload-attachment endpoints; child creation stays a modal. ✅

The removed capabilities (comment editing, reactions) are explicitly called out in the description ("existing comments are not editable"), so they are not spec violations. The new-loop composer's narrowing (hardcoded `status:todo`/`priority:none`, assignee restricted to agent/squad, auto-derived title) is coherent with the "hand it to an AI teammate" framing — noted as a question below, not a spec failure.

---

## 2. Code Quality

Two P1 regressions in the new comment-reply path must be fixed before merge. Both were confirmed against the tree, and both are made more clear-cut by the fact that the *sibling* top-level composer in the same file does the right thing — so these are inconsistencies within this PR, not debatable preferences.

### P1 — Thread reply box submits on bare Enter with no IME composition guard (breaks CJK input)
`packages/dmloop/src/panel/IssueDetailPage.tsx:130` (new `ThreadReply` component)

```jsx
<input className="loop-field" value={draft} ...
  onKeyDown={(e) => { if (e.key === "Enter") submit(); }} />
```

This is a raw native `<input>`; a Pinyin/Kana/Hangul user pressing Enter to **confirm an IME candidate** will instead post a half-composed reply (submit → `replyToThread` → `addComment` fires immediately, so the bad reply persists server-side). This is a net-new regression: the old reply path used Semi `<Input onEnterPress={...}>` (removed in this diff), which is IME-safe internally. For an enterprise IM product with a primarily-CJK userbase this is a real degradation of the core input flow.

Fix: `if (e.key === "Enter" && !e.nativeEvent.isComposing) submit();`, or switch to the Cmd/Ctrl+Enter chord the new-comment composer at `:971` already uses (which cannot collide with IME confirmation). The rest of the file already standardizes on the chord — consistency argues for it here too.

### P1 — Thread reply drops failed attachments with no retry (data loss), unlike the top-level composer
`packages/dmloop/src/panel/IssueDetailPage.tsx:502-519` (`replyToThread`) + `:111-119` (`ThreadReply.submit`)

`replyToThread` uploads each file, counts failures, toasts, and then `return true` **unconditionally** even when uploads failed. The caller then runs `if (ok) { setDraft(""); setFiles([]); }`, discarding the failed `File` objects permanently with no retry entry.

The top-level composer (`submitComment`, `:458-472`) deliberately does the opposite — it collects `failedFiles` and re-seeds `setPendingFiles(failedFiles)` so the user can retry, with an in-code comment explaining exactly why ("否则文件对象被丢弃、永久丢失且无重试入口"). The reply path reintroduces precisely the loss that comment was written to prevent. The same diff refactors that retry logic in the top-level path, so the requirement was clearly in mind — it was just not carried into the new reply path.

Fix: have `replyToThread` return the failed-file list (or a partial-failure flag) and re-seed `setFiles(failed)` in `ThreadReply` instead of clearing on partial failure.

- **Clipboard copy silently no-ops off secure contexts.** `navigator.clipboard?.writeText(...)?.then(...).catch(()=>{})` in `RunDetailModal.tsx` `copyAll` and the comment-copy `Dropdown.Item` (`IssueDetailPage.tsx`): when `navigator.clipboard` is undefined (http / embedded webview / older browser) the optional chain short-circuits — nothing is copied and there is zero feedback (no success or error toast). The empty `.catch(()=>{})` also swallows real permission rejections. Consider an `execCommand` fallback or at least an error toast.
- **Description edit is keyboard-unreachable.** `IssueDetailPage.tsx:881` — `<div className="loop-idp__desc" onClick={() => setEditingDesc(true)}>` has no `role`/`tabIndex`/key handler, and the previous keyboard-accessible Semi edit `<Button>` was removed. The ⋯ menu does not offer description edit, so keyboard-only/AT users lose the ability to edit the description. Make it a `<button>` or add `role="button" tabIndex={0}` + Enter/Space.
- **Sub-issue rows are mouse-only navigation.** `IssueDetailPage.tsx:926` — clickable `<div className="loop-subissue" onClick={openChild}>` with no keyboard affordance. Not a regression (old rows had no navigation), but the adjacent create button in the same block does add `aria-label`, so the omission is inconsistent.
- **Title `<input>` has no accessible label.** `IssueDetailPage.tsx:863-868` — migrated from Semi `<Input>` to a raw `<input>` with no `aria-label`/`placeholder`. Function is intact; add `aria-label={t("loop.field.title")}`.
- **Orphaned code/i18n after feature removals.** `updateComment` (`api/issueApi.ts`) and the four reaction functions (`api/collabApi.ts:20-33`) now have zero callers; the `loop.reaction.*` i18n keys and the `reactions` fields carried in state are no longer rendered (historical reactions become invisible — display only, not storage). The PR claims "dead-code cleanup"; consider finishing it or leaving a scope note.
- **Per-property success toast.** `patch()` now fires `Toast.success` on every property change; rapid inline edits via the sidebar popups will stack toasts. Minor UX; consider debouncing or dropping the toast for inline edits.

---

## 3. Overall Verdict

**REQUEST_CHANGES** — spec passes, but two confirmed P1 regressions in the new comment-reply path (IME-unsafe Enter handling and silent attachment data-loss) need to be fixed. Both are small, localized changes and the correct pattern already exists elsewhere in the same file to copy from. The P2 items are advisory and can be addressed in the same pass or a follow-up.

## 4. Suggested direction
- Reply box: gate Enter on `!e.nativeEvent.isComposing` (or adopt Cmd/Ctrl+Enter), matching `:971`.
- Reply attachments: thread the failed-file list back out of `replyToThread` and re-seed the input, matching `submitComment` at `:458-472`.
- While in there, add the description/sub-issue keyboard affordances and the title `aria-label` — cheap a11y wins.

## 5. Additional notes
- The controlled filter/display popover approach (explicit `visible`/`onVisibleChange` instead of `clickToHide`) is the right call given the portal'd Select/DatePicker children — good catch in the fix commit.
- Cross-file check passed: removed `RUN_STATUS_COLOR` has no remaining references; new `time.ts` helpers, `AutoGrowTextarea`, `AssigneePicker` `size`/`types` props, and `createIssue` `attachment_ids` all exist and match their call sites.
- Note: I could not run `tsc`/`stylelint`/the dev server in this environment (no installed deps), so I relied on the PR's stated green results plus static cross-file verification. Behavioral items above are from code reading, not a running build.

packages/dmloop/src/panel/IssueDetailPage.tsx
```diff
@@ -56,45 +56,113 @@ import {
...
+  return (
+    <div className="loop-cmt__reply">
+      <div className="loop-cmt__reply-row">
+        <input
+          className="loop-field"
+          value={draft}
+          onChange={(e) => setDraft(e.target.value)}
+          placeholder={placeholder}
+          disabled={busy}
+          onKeyDown={(e) => { if (e.key === "Enter") submit(); }}
```
OctoBoooot, line 130: **[major]** Reply box submits mid-IME-composition. Bare Enter with no composition guard: during CJK/IME input, Enter to confirm a candidate fires `submit()` mid-composition → the reply posts prematurely with half-composed text, and since comments aren't editable it's permanent. The main composer (`:971`) uses `metaKey/ctrlKey + Enter` (IME-safe), so the two are inconsistent and this is the unsafe one. PR-introduced: against the true merge-base `9b8c0761`, ThreadReply + both Enter handlers are added (the current feat/octo-loop tip is ahead of the merge-base, which is why it looks pre-existing). Fix: `if (e.key === "Enter" && !e.nativeEvent.isComposing) submit()`, and align the two composers' submit semantics.

OctoBoooot (changes requested):
## Review: #668 — correcting my verdict (Approve → Request changes)

**Verdict:** Request changes (correcting my APPROVE)
yujiawei is right and it IS PR-introduced — I nearly mis-graded this as "pre-existing" by diffing against the wrong base. Corrected: against the true merge-base, the IME-unsafe reply-submit is introduced by this PR. Byte-verified. Flipping.

Risk tier: normal; `needs-human-review` stands. Base `feat/octo-loop`.

### Major — reply box submits mid-IME-composition (bare Enter, no `isComposing` guard)
`IssueDetailPage.tsx:130` (ThreadReply): `onKeyDown={(e) => { if (e.key === "Enter") submit(); }}` — bare Enter submits with no composition guard. During CJK/IME input, pressing Enter to **confirm a candidate** fires `submit()` mid-composition → the reply posts prematurely with half-composed text. And since existing comments aren't editable in this design, a mis-fired reply is **permanent**. Reachable on ordinary CJK typing — material for the zh-CN user base.

By contrast the main comment composer (`:971`) requires `metaKey/ctrlKey + Enter`, so it's IME-safe — the two composers are inconsistent (yujiawei #2), and the reply one is the unsafe variant.

**Provenance — PR-introduced (corrected):** I first checked `?ref=feat/octo-loop` and saw these handlers "already there," reading it as pre-existing. That was the wrong base — the current `feat/octo-loop` **tip is `ahead=1` of #668's merge-base** (sibling work landed on it), so it already contained `ThreadReply`. Against the **true merge-base `9b8c0761`** (what #668 branched from): **zero** onKeyDown-Enter handlers, **no** ThreadReply. The authoritative `merge-base...head` diff shows both handlers as **added (`+`)**. So `ThreadReply` and its bare-Enter IME-unsafe submit are introduced by this PR. (This is exactly the head-vs-merge-base anchoring trap from earlier in this batch — caught before posting.)

**Fix:** guard composition on the reply Enter handler — `if (e.key === "Enter" && !e.nativeEvent.isComposing) submit()` — and align the two composers' submit semantics (both bare-Enter-with-guard, or both Cmd/Ctrl+Enter).

### Non-blocking
- yujiawei's 🟡 (new-loop composer restrictiveness / does it meet all user needs) — product-fit question, non-blocking.

### Standing (fixed earlier in this PR, unaffected)
Property editors restored, first-child-create, routing blank-pane, filter-popover `clickToHide` → controlled visible — all resolved. This IME defect is a separate, newly-scrutinized surface.

### Accountability
Two corrections here: (1) I originally approved #668 without exercising the composer keydown/IME path (same class as missing the filter-popover interaction), and (2) I almost compounded it by grading yujiawei's 🔴 as "pre-existing" off the wrong base — the branch tip, not the merge-base. The provenance rule I've applied all batch cuts both ways: diff **merge-base→head**, not against a moved branch tip. Corrected to REQUEST_CHANGES. Credit: yujiawei — real, PR-introduced, and I'd have wrongly waved it through twice over.

Jerry-Xin (changes requested):
## 🔴 REQUEST_CHANGES — correcting my verdict from APPROVE @ 3a3270ac0379

Flipping my APPROVE → REQUEST_CHANGES after byte-verifying the IME/composition regression that yujiawei and OctoBoooot flagged. My earlier APPROVE examined the popover fix and property-editor parity but did not audit the comment/reply composers' Enter handling — that miss is on me; the finding is real, reachable, and PR-introduced.

### 🔴 #1 — Thread-reply box mis-submits on IME candidate confirmation (CJK)
`IssueDetailPage.tsx:130` — the `ThreadReply` composer is a raw `<input>` with:
```jsx
onKeyDown={(e) => { if (e.key === "Enter") submit(); }}
```
No composition guard (`e.nativeEvent.isComposing` / `onCompositionStart|End` / `keyCode===229`). For zh-CN users (this product's primary locale), pressing **Enter to confirm a Pinyin/IME candidate** fires `submit()` and posts a **half-composed reply**. This is the common input path, not an edge case.

**Provenance — regression introduced by this PR.** At merge-base (`feat/octo-loop`), the reply composer was a Semi `<Input onEnterPress={...}>`; Semi's `onEnterPress` is composition-aware (it suppresses Enter while `isComposing`). This redesign replaced the Semi `Input` with a raw `<input>` + hand-rolled `onKeyDown`, **dropping** the composition safety Semi provided. So IME behavior for the thread-reply box got strictly worse.

**Fix:** `if (e.key === "Enter" && !e.nativeEvent.isComposing) { e.preventDefault(); submit(); }` — or revert to Semi `<Input onEnterPress={submit}>`.

### 🔴 #2 — Inconsistent Enter contract between the two composers in this file
- `ThreadReply` (:130): raw `<input>`, submits on **bare Enter**, no IME guard.
- Main new-comment box (`AutoGrowTextarea`, :971): submits only on **Cmd/Ctrl+Enter** — IME-safe by construction (a bare IME-confirm Enter can't submit it).

Two "post a comment"-class actions in one file now have divergent, undocumented Enter semantics, and the bare-Enter path is precisely the IME-unsafe one. At merge-base both were consistent Semi `Input onEnterPress` (both composition-guarded). Please align the two composers on one Enter contract (and make it composition-safe).

Note: the main new-comment textarea (:971) is actually the *safe* one — if the concern was pointed there specifically, that box is fine (modifier-gated). The defect is specifically the `ThreadReply` `<input>` at :130.

### Everything else stays resolved
Prior blockers remain fixed at this head: the 5 detail property editors are restored and wired to `patch`; the filter/show popovers now use controlled `visible`/`onVisibleChange` (no more clickToHide close-on-inner-click). Non-blocking 🟡 still open: emoji-reaction removal needs a documented-intent note; `NewLoopPage.deriveTitle` `slice(0,80)` truncates by UTF-16 code unit (cosmetic).

### Verdict
**REQUEST_CHANGES** on the ThreadReply IME/composition regression (:130) + the composer Enter-contract inconsistency. Both are byte-confirmed regressions vs merge-base and highly reachable for the primary zh-CN locale. Clears once the reply box guards composition (and the two composers agree on one Enter contract).

PR #669, opened 2026-07-12
feat(versions): sheet SheetVersionPanel thin adapter over unified shell

ploy-elison:
## Summary

Makes the sheet `SheetVersionPanel` a thin adapter over the unified `<VersionHistoryPanel>` shell that landed on `main` (#656), mirroring the doc adapter. This is an independent change: it only imports the shell and its own sheet-specific helpers, with no dependency on the doc or board adapters.

The panel drops its local list / preview / compare / mutation implementation (single-fetch list, inline preview drawer, `window.confirm` restore, `window.prompt` rename) and now renders the shell, injecting only the sheet-specific pieces:

- `loadPreviewState` → `getVersionState(docId, seq, signal)` → the snapshot's `sheetCells`
- `renderPreview` → a read-only `<CellGrid>` of the snapshot cells
- `renderDiff` → a cell-level change list (added / changed / removed) against the live grid
- `getCurrent` → the sheet's current cells read from the live Y.Doc (read-only)

The sheet-specific pure helpers (`CellGrid` / `diffCells` / `colToA1` / `parseKey` / `currentCells`) are preserved unchanged; only the panel chrome moved into the shell. The sheet now inherits the mixed list + filter tabs + counts + load-more, the centered preview / diff modal, the in-panel restore confirm box, inline rename, and the unified race guard. Restore stays forward / non-destructive — the backend reconciles the `sheet` map onto the live doc and the grid updates via Yjs. No new i18n keys and no new endpoint. `SheetView` passes `docId` / `role` / `sheet` / `names` / `onClose`, which the adapter's props match exactly, so it is untouched.

## Changes

- `packages/docs/src/sheet/SheetVersionPanel.tsx` — rewritten as a thin adapter over `VersionHistoryPanel`
- `packages/docs/src/sheet/SheetVersionPanel.test.tsx` — 9 cases pinning the adapter wiring and converged behaviors (mixed list + `kind=all` on mount, filter tabs, counts, load-more pagination, preview via `getVersionState` rendering the cell grid in a centered modal, cell-level compare against the live grid, inline rename, the preview race guard, admin-only restore/delete gating, and preview never mutating the live sheet)

## Testing

- `pnpm exec vitest run src/sheet/SheetVersionPanel.test.tsx` — 9/9 passing
- `pnpm run typecheck` — no new errors from this change (the two pre-existing failures in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts` are present on `main` and unrelated)
- `git diff main...HEAD --stat` confirmed the diff contains only the two sheet files — no shell-file edits/deletions and no doc adapter changes


OctoBoooot (dismissed):
## Review: sheet SheetVersionPanel thin adapter over unified shell (#669)

**Verdict:** Approve
Clean thin-adapter refactor mirroring the doc adapter over the #656 `VersionHistoryPanel` shell. The sheet-specific pure helpers are byte-identical to the old panel, the injected hooks match the shell's generic signature exactly, and the read-only guarantee holds. No defects.

Risk tier: normal. Base `main`; CI green (Build/osv/code-review pending, not failing). +299/-254, 2 files (one is +246 tests).

### What I verified
- **Pure helpers preserved unchanged (as claimed).** `colToA1`, `parseKey`, `diffCells`, `currentCells`, `cellText` are all **byte-identical to base** (matching normalized hashes) — so they carry their existing correctness. Independently checked `colToA1`: correct bijective base-26 (0→A, 25→Z, 26→AA, 702→AAA), no off-by-one at the 26 boundary.
- **Adapter props match the shell's generic signature exactly.** Shell expects `loadPreviewState: (seq, signal) => Promise<TState>`, `renderPreview: (state) => ReactNode`, `renderDiff?: (state, current) => ReactNode`, `getCurrent?: () => TCurrent | null` — the adapter supplies all four with `TState=TCurrent=CellMap`. The `AbortSignal` is threaded into `getVersionState` (so the shell's race guard — verified on #656 — actually protects sheet previews), and `renderDiff` correctly handles the nullable `current` (`current ?? {}`).
- **Read-only guarantee intact (the version-family invariant).** `currentCells` copies `sheet.ydoc.getMap('sheet').entries()` into a plain object — no `.set`/`.delete`/mutation anywhere in the adapter. Preview/diff/getCurrent only read. Restore/rename/delete are delegated to the shell + backend (`liveRestore.ts`), so the panel never mutates the live Y.Doc — restore stays forward/non-destructive as stated.
- **Robustness.** `CellGrid` caps the rendered range to 200×50 so a sparse far-out cell can't explode the table; empty-grid and no-diff states render dedicated messages; `parseKey` null-guards non-integer keys (bad keys are skipped, not crashed on).
- **Scope contained.** Diff is only the two sheet files — no shell edits, no doc-adapter changes (consistent with the PR's `git diff --stat` claim). `SheetView`'s existing props (`docId/role/sheet/names/onClose`) match the adapter, so callers are untouched.

### Praise
- This is the adapter pattern done right: all the shared, bug-prone machinery (list/filter/counts/load-more, the race guard, the restore-confirm box, the preview/diff modal) is inherited from the one shell that's already been hardened across several rounds (#656), while the sheet end injects only its genuinely-specific pieces — and those pure helpers are carried over byte-for-byte rather than rewritten, so no new surface is introduced. The read-only-Y.Doc discipline (copy-out on read, all mutation delegated) is exactly the right boundary for a version panel.

### Note (non-blocking)
Inline-style hex colors in `CellGrid`/`SheetDiffList` (`#ddd`, `#16a34a`, etc.) are literal `CSSProperties`, not `var(--x)` tokens — but they're inline JSX styles in a preview grid (not the token-governed `.css` files DEVELOPMENT.md §2 covers), and the prior panel used the same literals. Consistent with existing code; a future pass could tokenize if desired.

Jerry-Xin (approved):
The PR is in scope for `Mininglamp-OSS/octo-web` and the sheet version-history adapter is aligned with the existing unified `VersionHistoryPanel` architecture. I found no blocking correctness, security, permission, or race-condition issues.

💬 Non-blocking:
- 🟡 Warning: [SheetVersionPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-kodhxm5k/repo/packages/docs/src/sheet/SheetVersionPanel.tsx:85) and [SheetVersionPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-kodhxm5k/repo/packages/docs/src/sheet/SheetVersionPanel.tsx:167) still contain hardcoded colors/spacing in inline styles. These appear to be preserved sheet preview/diff helper styles rather than a new behavioral regression, but they remain contrary to the project’s token guidance in `DEVELOPMENT.md`.

✅ Highlights:
- The adapter cleanly delegates shared list/filter/pagination/mutation/modal behavior to `VersionHistoryPanel`.
- Sheet-specific concerns are narrowly scoped to preview cells, live-cell diffing, and reading current cells from the Y.Doc.
- Added tests cover the important integration behavior: `kind=all`, preview state loading, diff rendering, race guard, role gating, pagination, rename, and non-mutating preview.

Verification note: I attempted to run the sheet and i18n Vitest commands, but this checkout does not currently expose `vitest` through `pnpm exec` or `pnpm --filter @octo/docs exec`, so I could not independently execute them here.

yujiawei (dismissed):
# Code Review — PR #669 (octo-web)

Reviewed at head SHA `66b84194d892ca4606ce75ab1dfc812bb8c59538` against `origin/main` (merge-base `d80703a`). Diff scope confirmed: exactly two files — `packages/docs/src/sheet/SheetVersionPanel.tsx` (rewrite) and `SheetVersionPanel.test.tsx` (new). No shell-file, doc-adapter, API, or i18n edits, matching the PR description.

## 1. Spec compliance

**Spec: ✅**

The change does exactly what it claims — makes `SheetVersionPanel` a thin adapter over the unified `VersionHistoryPanel` shell, injecting only the four sheet-specific seams and preserving the pure helpers unchanged.

- **Missing (under-build):** none. All four injected props are present and correctly typed: `loadPreviewState` (→ `getVersionState` → `sheetCells`), `renderPreview` (→ `CellGrid`), `renderDiff` (→ `SheetDiffList`), `getCurrent` (→ `currentCells`). The consumer `SheetView.tsx:725` passes `docId/role/sheet/names/onClose`, all of which the adapter's prop signature still accepts — `SheetView` is untouched and continues to type-check.
- **Over-build:** none. No new endpoint, no new i18n key, no new flag. The sheet diff/preview i18n keys it uses (`docs.sheet.version.added/removed/changed/noDiff/emptyCell/emptyGrid`) already exist in both `zh-CN.json` and `en-US.json`, and the shell-level keys (`filterAll`, `countManual`, `staleNotice`, `compareTitle`, etc.) are present in both locales.
- **Deviation:** none. Behavior converges onto the shell (mixed list + filter tabs + counts + load-more, centered preview/diff modal, in-panel restore/delete confirm, inline rename, unified race guard) instead of the old single-fetch list + `window.confirm`/`window.prompt`. Restore stays forward/non-destructive (backend reconciles the `sheet` map via Yjs); the panel never writes the live doc.

Note: this converts the sheet end only. The doc panel (`versions/VersionPanel.tsx`) and board panel (`board/BoardVersionPanel.tsx`) are still their own pre-shell implementations on this branch — consistent with a staged per-end migration, and out of scope here.

## 2. Code quality

**Quality: Approved**

Verification performed:
- **Tests:** `SheetVersionPanel.test.tsx` 9/9 pass; the shell's own `VersionHistoryPanel.test.tsx` 15/15 pass. The sheet suite pins the adapter wiring (mixed list + `kind=all` on mount, preview via `getVersionState` rendering the cell grid, cell-level compare against the live grid incl. added/removed/changed, admin-only restore/delete gating, inline rename not `window.prompt`, the preview race guard, and preview-never-mutates-live-sheet).
- **Typecheck:** no new errors from the two changed files. The two pre-existing failures the PR calls out (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) reproduce on `main` and are unrelated.
- **AbortSignal chain:** end-to-end intact. Shell `onPreview` gets `signal` from `previewGuard.begin()` → passes to `loadPreviewState(seq, signal)` → adapter forwards to `getVersionState(docId, seq, signal)` → axios `{ signal }`. Superseded previews are dropped by `isCurrent()`.
- **Role gating:** restore/delete correctly admin-only via the shell's `canRestoreVersion` (= `canManage` = admin); rename/save are writer+ via `canSnapshot`. A writer sees preview/save/rename but not restore/delete, matching the pre-shell gating.
- **`sheetCells` decoding:** `getVersionState` returns `VersionStateResult { doc, schemaVersion, docVersionSeq }`, which has no `sheetCells` field, so the adapter casts `r as { sheetCells?: CellMap }` and defaults to `{}`. This is identical to the old panel and to the board panel's documented pattern — the backend returns `sheetCells` for a sheet doc; the typed interface is doc-shaped. Pre-existing contract, not a regression.

### Findings

No P0/P1. Advisory only:

- **[P2] Orphaned i18n keys.** Removing the local list/mutation implementation left ~14 `docs.sheet.version.*` keys unused (`loadFailed`, `saveFailed`, `previewFailed`, `restoreFailed`, `renameFailed`, `deleteFailed`, `restoreConfirm`, `renamePrompt`, `deleteConfirm`, `kindNamed`, `kindRestore`, `kindAuto`, `viewContent`, `closePreview`). `pnpm i18n:check` still passes (they are within baseline), so this is not a gate break — just dead catalog entries worth pruning in a cleanup pass.
- **[P2] Preview modal falls back to whole-cell substitution, no in-cell text diff.** `SheetDiffList` shows `from → to` per changed cell, which is the right granularity for a grid; noting it only so expectations are set (no character-level diff inside a cell). Behavior matches the pre-shell panel.

### Cross-file blind spots checked
- Consumer `SheetView.tsx` prop match — verified, untouched.
- Shell effect dependency arrays — verified the injected render-props are **not** referenced in any `useEffect`/`useMemo`/`useCallback` dep array (`[refresh]`, `[]`, `[]`, `[selected, closePreview]`), and neither `CellGrid` nor `SheetDiffList` is `React.memo`-wrapped, so passing fresh inline callbacks each render is harmless — no re-render storm or refetch loop.
- Falsy cell values (`0` / `false`): the `||`-based empty fallback in `SheetDiffList` operates on `d.from`/`d.to`, which `cellText` has already `String()`-ified (`0`→`"0"`, `false`→`"false"`, both truthy). Only a genuinely empty cell produces `""`. No falsy-collapse bug.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Clean, well-tested thin-adapter refactor with no behavioral regression; the two P2s are non-blocking cleanup items.

## 4. Suggestions
- Prune the orphaned `docs.sheet.version.*` keys listed above in a follow-up to keep the catalog honest.

## 5. Additional notes
- Coverage gap (out of this diff's control): the `getVersionState` → `sheetCells` contract is enforced only by a cast, not by the `VersionStateResult` type. A shared discriminated result type (doc vs sheet vs board payload) would make this contract type-safe across all three adapters. Tracked as tech-debt, not a blocker for this PR.

Jerry-Xin (dismissed):
**Re-review — APPROVED** (base=main, head `feat/sheet-version-adapter-upstream`, verified against shell baseline `d80703a5`; diff = SheetVersionPanel.tsx + SheetVersionPanel.test.tsx only, 2 files).

**1. Thin adapter correctness — PASS.** `SheetVersionPanel` is now a pure wrapper that returns `<VersionHistoryPanel>` and injects only the four sheet-specific hooks (`loadPreviewState`, `renderPreview`, `renderDiff`, `getCurrent`). All the shared machinery that used to live here — list/filter tabs/counts/load-more, save/rename/delete/restore, the inline compose rows, the restore confirm box, the race guard, and the preview/diff modal — is deleted and now owned by the shell. No sheet-local reimplementation of any shell concern remains. The sheet-only pure helpers (`CellGrid`, `diffCells`, `colToA1`, `parseKey`, `cellText`, `currentCells`) are preserved unchanged; `SheetDiffList` is a thin presentational wrapper over the existing `diffCells` + `octo-sheet-diff-list` markup.

**2. Contract alignment + doc-parity (decision #5) — PASS.** Props passed to the shell exactly match `VersionHistoryPanelProps<CellMap, CellMap>`; `tsc --noEmit -p tsconfig.typecheck.json` reports zero errors in either file (the only two repo errors are pre-existing, in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts`, both outside this diff). `loadPreviewState` forwards the shell's `AbortSignal` into `getVersionState(docId, seq, signal)` — honoring the shell's "MUST honor the AbortSignal" contract — and the signature lines up with `getVersionState(docId, docVersionSeq, signal?)`. Because the sheet supplies both `renderDiff` and `getCurrent`, compare is enabled, matching the doc panel's capability set (sheet aligns with doc). The shell's new i18n keys (`filterAll/filterManual/filterAuto`, `countManual/countAuto`, `loadMore`, `staleNotice`) are present in `zh-CN.json`/`en-US.json` on this branch, so mounting the shell surfaces no raw keys.

**3. No shell pollution — PASS.** `changedFiles=2`; the diff touches neither `VersionHistoryPanel.tsx` nor `raceGuard.ts` nor any other shared/doc/board file.

**4. Tests — PASS (real, non-green).** 9/9 pass locally (`vitest run src/sheet/SheetVersionPanel.test.tsx`). The suite mocks only `../versions/api.ts` and drives the **real** shell, so it pins the adapter wiring against live shell behavior: mount requests `kind:'all'`, unified counts header, filter-tab reload, preview via `getVersionState` with `docId`+`seq` forwarded, cell-level diff (added/changed/removed asserted, unchanged cell absent), admin-gated restore/delete (writer sees neither), load-more with `cursor=100`, inline rename with exact args `['d_1', 7, 'Renamed v1']` and no `window.prompt`, race-guard rejection of a stale out-of-order response, and preview-never-mutates-live-sheet (write-throwing fake grid). Assertions are specific enough that they would fail on regression — not always-green.

Prior APPROVED stands; re-confirming.

OctoBoooot (approved):
## Review: #669 — delta re-review (adds the 6 shell i18n keys; closes a latent gap I under-caught)

**Verdict:** Approve (re-affirmed at head `d099c8b71354`)
The new commit adds the 6 shell `docs.version.*` keys (filter/count/staleNotice) to both locales — closing a latent raw-keys-on-open gap that was present at my prior-approved head. Byte-verified. Strict improvement; approval carries forward.

Risk tier: normal. Base `main`. Delta = 1 commit, +6/+6 i18n only.

### What changed — and why it matters
- The 6 keys `filterAll/filterManual/filterAuto`, `countManual/countAuto`, `staleNotice` are added to `en-US.json` + `zh-CN.json`, all resolving under the `version.*` namespace. i18n parity holds (402/402, zero divergence).
- **This closes a real latent gap:** `SheetVersionPanel` mounts the shared `VersionHistoryPanel`, which references these 6 keys — and at my prior-approved head `66b84194` **they were absent** (verified). So the sheet version panel would have rendered raw key strings (`docs.version.filterAll` etc.) in its filter tabs + counts on open — the exact defect I flagged as a blocker on #671. #669 had the same gap; this commit fixes it, making #669 self-contained for the shell mount regardless of #670's merge.

### Accountability
This is the same blind spot I had on my first #671 pass: I verified the sheet adapter's *own* pieces (pure helpers, props, read-only Y.Doc) but not that the *shell's* full `docs.version.*` key set resolves in #669's branch. I caught it on #671 (credit Steve) but did not back-apply the check to #669/#670, which mount the same shell — #670 happened to supply the keys itself, #669 did not until now. Lesson reinforced: **when an adapter mounts shared code, verify the shared code's entire i18n key set resolves in that branch — apply it to every sibling that mounts the same shell, not just the one a peer flagged.**

### Standing (verified original round, unaffected)
Sheet adapter itself sound: pure helpers byte-identical to base, props match the shell's generic signature, read-only Y.Doc guarantee, AbortSignal threaded into the race guard, render caps. The one prior 🟡 (inline-style hex colors) unchanged — non-blocking.

Re-affirming APPROVE — the delta is a pure i18n fix that resolves a latent gap; #669 is now correct and self-contained.

Jerry-Xin (approved):
Re-reviewed at head `d099c8b7` (byte-verified). **APPROVE.**

## Delta since the previously-3×-approved head `66b84194`
- `compare 66b84194...d099c8b` → **ahead_by:1, behind_by:0** — clean forward, not rebased (compare list is honest).
- Single commit: *"i18n(versions): add top-level version filter/count/staleNotice keys for shell"*.
- **Adapter TSX is byte-identical** to the approved head (blob-hash verified): `SheetVersionPanel.tsx` (d0a5ef0e) and `SheetVersionPanel.test.tsx` (2e09d58a) both SAME. The adapter wiring already approved is untouched.

## What the commit fixes
The unified shell `versions/VersionHistoryPanel.tsx` references `docs.version.{filterAll,filterManual,filterAuto,countManual,countAuto,staleNotice}` (lines ~266/299/322/349/511-516). At the prior head these 6 top-level `version.*` keys were **missing in both locales**, so the shell's filter buttons / counts / stale-notice would render raw key strings. This commit adds all 6 in **both en-US and zh-CN** under the correct top-level `version` namespace (matching the sibling `board.version.*` convention). Verified: each key present in both locales, JSON parses, no dup-key error, every added key has a live `t()` consumer, values sensible.

## Non-blocking
- 🟡 `SheetVersionPanel.tsx:85,112-123,165-173` — sheet preview/diff still uses inline hardcoded colors/spacing/borders (at odds with token guidance in DEVELOPMENT.md). This is **pre-existing** on lines byte-identical to the approved head — not introduced or worsened here. Optional follow-up.

No regression, no prop-drift, no dead code, both locales covered. APPROVE.

ploy-elison (commented):
Re-reviewed at head `d099c8b7` after the i18n backfill. Approve.

Delta since the previously-approved head `66b84194` is a single commit (`compare 66b84194...d099c8b` → ahead 1, behind 0 — clean forward, no rebase) touching only the two i18n bundles, +6/-0 each. No change to the adapter or tests, so the earlier approvals of the thin-adapter refactor still stand.

The 6 shared-shell keys are now defined under the top-level `version` object in both locales, byte-verified against the doc bundle the shell reads (`VersionHistoryPanel.tsx` calls `t('docs.version.*')` → top-level `version.*`, not `board.version`):

- en: filterAll=All, filterManual=Manual, filterAuto=Auto, countManual=Manual, countAuto=Auto, staleNotice="Done, but the list couldn't refresh — reopen to see the latest."
- zh: filterAll=全部, filterManual=手动, filterAuto=自动, countManual=手动, countAuto=自动, staleNotice="操作已完成，但列表未能刷新——重新打开可查看最新内容。"

Both files parse as valid JSON with no duplicate keys. Values match the doc #670 spec exactly. No unrelated changes crept in.

yujiawei (approved):
# Code Review — PR #669 (octo-web)

**Verdict: APPROVE.** The sheet `SheetVersionPanel` is now a genuinely thin adapter over the unified `<VersionHistoryPanel>` shell (#656). I reviewed the full diff against `origin/main`, the shell contract, the call site, both i18n locales, and the test file, and ran the suite locally. No blocking (P0/P1) issues.

## 1. Spec compliance ✅

The diff does exactly what the summary describes and nothing more:

- **Thin adapter, correct wiring.** The panel instantiates `VersionHistoryPanel<CellMap, CellMap>` and injects only the four sheet-specific seams — `loadPreviewState`, `renderPreview`, `renderDiff`, `getCurrent` — whose types match the shell's generic props (`VersionHistoryPanel.tsx:68-74`). Sheet-specific helpers (`CellGrid` / `diffCells` / `colToA1` / `parseKey` / `currentCells`) moved position but are byte-identical (no changed lines in the diff). `SheetDiffList` is a verbatim lift of the old inline diff list.
- **No shell / API / doc / board edits.** `git diff main...HEAD --name-only` = 4 files: the two sheet files plus `i18n/en-US.json` and `i18n/zh-CN.json`. The shared REST layer (`versions/api.ts`) is untouched — no new endpoint.
- **Call site unchanged and prop-compatible.** `SheetView.tsx:725` passes `docId / role / sheet / names / onClose`, which the adapter's props match exactly.
- **The 6 new i18n keys are in-scope, not scope creep.** `filterAll / filterManual / filterAuto / countManual / countAuto / staleNotice` were absent from the top-level `version` block on `main`. The shell already references them via `t('docs.version.*')` and its header comment explicitly deferred adding them "to the doc-adapter phase, when an end first renders the shell." This PR is that first mounting consumer, so adding them (symmetrically in both locales) is the minimum change required for the shell to render without leaking raw keys.

> Doc-hygiene nit (non-blocking): the PR body says "No new i18n keys" and "the diff contains only the two sheet files." Both are stale at the current head — the diff adds 6 keys × 2 locales and touches 4 files. The "no new endpoint" and "no shell-file edits" halves are accurate. Worth refreshing the description, but it does not affect the shipped code.

## 2. Code quality ✅

- **`getCurrent` reads fresh live cells.** The shell computes `currentForDiff = compare && getCurrent ? getCurrent() : null` inline on every render (`VersionHistoryPanel.tsx:496`), not memoized, so each compare reflects the live `Y.Map('sheet')` at open time — same semantics as the old `useMemo(…, [compare, preview, sheet])`, and strictly fresher.
- **AbortSignal forwarded.** `loadPreviewState={(seq, signal) => getVersionState(docId, seq, signal)…}` threads the shell's race-guard signal into the axios request, honoring the documented contract.
- **Null-safety is a defensive no-op.** `currentCells` returns `{}` (never null); the `current ?? {}` guard in `renderDiff` can't fabricate a misleading diff, and a null sheet yields a truthful all-"removed" diff.
- **Read-only preview enforced.** Preview renders the snapshot `sheetCells` via a capped read-only `<CellGrid>` (rows ≤ 200 / cols ≤ 50); restore stays backend-reconciled, the panel never writes the live doc.

### Behavior changes worth noting (intended, from #656 — not regressions)

- List page size unifies 50 → 30 with load-more, and the sheet now inherits filter tabs + counts it previously lacked.
- The old native `window.confirm` / `window.prompt` for restore/rename are replaced by the shell's in-panel confirm box and inline rename input.

## 3. Tests ✅ (with minor suggestions)

The 9 cases pin the sheet-injected seams (snapshot cell grid, cell-level diff against the live grid, `sheetCells` decoding, admin-only restore/delete gating). The two hardest tests have real teeth: the race-guard test forces out-of-order resolution and asserts the stale response is dropped; the no-mutation test uses a fake sheet whose `set`/`delete` throw, so any write fails the test. `pnpm exec vitest run src/sheet/SheetVersionPanel.test.tsx` → 9/9; full versions+sheet suite → 24/24. `typecheck` shows only the two pre-existing unrelated failures already present on `main`.

Non-blocking suggestions for a follow-up:

- **P2** — the preview test comments "signal forwarded" but only asserts `calls[0][0]`/`[1]`; add `expect(getVersionStateMock.mock.calls[0][2]).toBeInstanceOf(AbortSignal)` so a dropped 3rd arg (which would break abort-on-supersede) can't stay green.
- **P2** — no test exercises the `sheetCells ?? {}` fallback (empty/legacy snapshot) and the resulting empty-grid render.
- **P2** — the sheet-specific pure helpers have no branch-level tests; notably `colToA1`'s multi-letter path (col ≥ 26 → "AA", the `do/while`) is unexercised (all fixtures use cols 0–3). Not a regression (no prior tests existed), but high-value to add.

## 4. Additional observations

- Orphaned `sheet.version.*` i18n keys remain in both locales (`loadFailed`, `restoreConfirm`, `renamePrompt`, `kindNamed`, `closePreview`, `viewContent`, …) now that the shell drives that chrome via `docs.version.*`. Harmless dead strings; safe to prune in a cleanup pass.
- The 6 shared shell strings now exist as byte-identical copies in both `board.version` and the top-level `version` block — a future drift risk worth a consolidation follow-up ticket, out of scope here.

PR #670, opened 2026-07-12
feat(versions): make doc VersionPanel a thin adapter over VersionHistoryPanel

ploy-elison:
## Summary

Switches the **doc** version-history panel onto the unified `VersionHistoryPanel` shell that landed on `main` (squash `d80703a5`). `VersionPanel` becomes a thin adapter: it drops its own dual-stream grouping (manual/auto groups, per-stream cursors, lazy auto fetch) and its local last-write-wins `previewGuard`, and now renders `<VersionHistoryPanel>`, injecting only the doc-specific pieces:

- `loadPreviewState` → `getVersionState(docId, seq, signal).then(r => r.doc)`
- `renderPreview` → a read-only throwaway editor (`VersionPreview`) — never the live one
- `renderDiff` → the block-level `DiffView` over `diffDocs(version, current)`
- `getCurrent` → the live editor's JSON (read-only) as the "current" side of a diff

Restore stays forward / non-destructive (the live surface reconciles via Yjs); schema-mismatch restore errors keep their dedicated message via `restoreErrorKey`. The live editor is read but never mutated.

## User-visible change

The history list is now a single mixed stream with filter tabs (all / manual / auto) + counts + load-more, replacing the two collapsible groups. Preview / diff / restore / rename / delete semantics are unchanged.

## Notes

- `previewGuard.ts` (and its test) is retired: the shell's unified `raceGuard` supersedes it, and it had no other consumer on `main` (the shell uses `createRaceGuard`; only the pre-adapter doc panel imported `previewGuard`).
- Adds the shell's new `docs.version.*` i18n keys — `filterAll` / `filterManual` / `filterAuto`, `countManual` / `countAuto`, `staleNotice` — in both `en-US` and `zh-CN`.
- The shell (`VersionHistoryPanel.tsx`) is **not** modified here; this adapter only imports it.

## Scope

Doc adapter only — independent of the sheet/board adapters, based directly on `main`. Diff is limited to `packages/docs/src/versions/*` and the two doc i18n files.

## Verification

- `packages/docs`: `vitest run src/versions/` → 7 files / 75 tests pass.
- `tsc --noEmit -p tsconfig.typecheck.json`: clean for this change. The two remaining errors (`packages/dmworkbase/src/i18n/format.ts`, `packages/docs/src/members/sort.ts`) are pre-existing on `main` and unrelated to these files.


Jerry-Xin (approved):
Summary: The PR is relevant to `Mininglamp-OSS/octo-web` and the doc version panel adapter cleanly delegates shared behavior to `VersionHistoryPanel` while preserving doc-specific preview, diff, and restore error handling.

💬 Non-blocking:
- 🔵 Suggestion: [VersionPanel.test.tsx](/Users/mlamp/.openclaw/tmp/pr-review-ni2c0uz6/repo/packages/docs/src/versions/VersionPanel.test.tsx:120) checks `docId` and `seq` forwarding to `getVersionState`, but not the `AbortSignal`. Since the adapter’s race-cancellation correctness depends on forwarding that signal, consider asserting `getVersionStateMock.mock.calls[0][2]` is an `AbortSignal`.

✅ Highlights:
- [VersionPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-ni2c0uz6/repo/packages/docs/src/versions/VersionPanel.tsx:99) correctly forwards the shell-provided preview abort signal into `getVersionState`.
- [VersionPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-ni2c0uz6/repo/packages/docs/src/versions/VersionPanel.tsx:101) keeps doc diffing scoped to `diffDocs(version, current)` and reads the live editor via `getJSON()` without mutation.
- [VersionPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-ni2c0uz6/repo/packages/docs/src/versions/VersionPanel.tsx:103) preserves the dedicated schema-mismatch restore message through `restoreErrorKey`.
- The new `docs.version.*` keys are present in both active locale files.

Verification note: I could not run the focused Vitest suite locally because this checkout has no `node_modules`, so `vitest` is not resolvable. I reviewed the code path directly against `VersionHistoryPanel`, `api.ts`, `diff.ts`, and the locale files.

OctoBoooot (approved):
## Review: make doc VersionPanel a thin adapter over VersionHistoryPanel (#670)

**Verdict:** Approve
The doc counterpart to #669 — VersionPanel becomes a thin adapter over the #656 shell, deleting 903 lines of now-redundant dual-stream/previewGuard code. Adapter wiring, read-only guarantee, i18n, and the previewGuard retirement all verify clean. No defects.

Risk tier: normal. Base `main`; CI green (**Build passes**, code-review pending). +138/-903, 6 files.

### What I verified
- **Adapter props match the shell's generic signature** (`VersionHistoryPanel<PMNode, PMNode>`): `loadPreviewState → getVersionState(...).then(r => r.doc)` with the `AbortSignal` threaded (so the shell's race guard protects doc previews), `renderPreview`, `renderDiff(version, current)`, `getCurrent`, plus `pageSize={25}` (per-end override) and `restoreErrorKey`.
- **Read-only guarantee intact** (version-family invariant): `getCurrent` = `editor?.getJSON()` (a read), `VersionPreview` builds a **throwaway** editor (`editable: false`, separate instance — explicitly "never the live one"), `renderDiff` only reads. The live `editor` is never mutated; restore is delegated to the shell + Yjs reconciliation (forward/non-destructive).
- **Schema-mismatch error mapping preserved** — `restoreErrorKey` maps `VersionSchemaIncompatibleError`/`VersionSchemaNewerError` → `docs.version.errorRestoreIncompatible`, else `errorRestore`. (This is the dedicated-message behavior that was a 🟡 concern on #656 — carried through correctly.)
- **New i18n keys resolve in both locales** — `filterAll/Manual/Auto`, `countManual/Auto`, `staleNotice`, `errorRestoreIncompatible`, `errorRestore` all present in en-US **and** zh-CN; full parity (402/402 keys, zero divergence). Notably `staleNotice` — the key I flagged as missing on #656 — is now defined.
- **`previewGuard` retirement is safe.** Direct tree-scan at this head found **no importer** of `previewGuard` among versions/sheet/board files; and CI **Build passes** at `b08d8336`, which a dangling import of the deleted module could not do. The shell uses `createRaceGuard`; the retired last-write-wins guard had no other consumer. Clean deletion (`-27` src, `-71` test).

### Praise
- Same adapter pattern done right as #669, and here it retires a whole parallel implementation (dual-stream grouping, per-stream cursors, lazy auto-fetch, the local `previewGuard`) — 903 lines deleted — by leaning on the one shell that's been hardened across #656's several rounds, instead of maintaining three copies of the race/list/restore machinery. The throwaway-preview-editor and read-only `getJSON()` boundary keep the live editor untouched, and folding `staleNotice` + the schema-restore keys into both locales closes exactly the i18n/error-mapping gaps surfaced earlier in this shell's history.

### Note
Diff is scoped to `packages/docs/src/versions/*` + the two doc i18n files; the shell itself is not modified (consistent with the PR's scope claim). 75 version tests pass.

yujiawei (approved):
# Code Review — PR #670 (octo-web)

**Head SHA reviewed:** `b08d8336a762034b157983503bcbb7914142b4b7` · **Base:** `main` (merge-base `d80703a5`, the unified shell commit #656)

This is a refactor: the doc `VersionPanel` becomes a thin adapter over the pre-existing, unmodified `VersionHistoryPanel` shell. I traced the adapter wiring end-to-end against the pre-adapter panel (`main`) and the shell it delegates to, ran the tests and typecheck locally, and cross-checked with an independent second-opinion pass.

The migration is faithful. Every doc-specific behavior is preserved by the injected hooks; the only user-facing deltas are the ones the PR description explicitly calls out (single mixed list + filter tabs replacing the two collapsible groups).

- **Diff argument order** — correct. Adapter calls `diffDocs(version, current)`; `diff.ts` treats arg1 as *before* (historical) and arg2 as *after* (live). Matches the old `diffDocs(previewJSON, currentDoc())`.
- **`getCurrent` null-safety** — `(editor?.getJSON() as PMNode | undefined) ?? null`; with no editor it returns `null`, and `diffDocs(version, null)` is well-defined (`docToBlocks(null) → []`, all historical rows render as removed). Mirrors the old `currentDoc()`.
- **Preview schema messaging** — preserved. The adapter intentionally omits `previewErrorKey`, so the shell's `defaultErrorKey` maps `VersionSchemaNewerError → previewSchemaNewer` and `VersionSchemaIncompatibleError → previewSchemaIncompatible` (else `previewNetworkError`). `getVersionState` still throws both typed 409 errors. Byte-equivalent to the old inline branching.
- **Restore error keys** — preserved. `restoreErrorKey` maps both schema errors to `errorRestoreIncompatible`, else `errorRestore` — identical to the old `onConfirmRestore`. Both keys exist in `en-US` and `zh-CN`.
- **Counts / page size / role gating** — unchanged: manual header = `counts.manual + counts.restore`, auto = `counts.auto`; `pageSize={PAGE_SIZE}` keeps the doc's 25; `canSnapshot` / `canRestoreVersion` gating is identical.
- **Public API** unchanged — `EditorShell.tsx:751` calls `VersionPanel` identically.
- **`previewGuard.ts` + test removal** is safe — it had no other consumer on `main` (the shell uses its own `createRaceGuard`).
- **AbortSignal on preview** — the adapter now forwards a signal to `getVersionState` (the old panel passed none). Strictly an improvement.

No missing, unacknowledged-extra, or deviating behavior.

## 2. Code quality — Approved

Findings are all non-blocking (P2 + one test-coverage suggestion). Concurrency/lifecycle and data-flow dimensions came back clean (throwaway preview editor is destroyed by `@tiptap/react` on dep-change/unmount; inline adapter callbacks are never in an effect dep array, so no re-render loop; all preview races are owned by the shell's guard).

**Suggestion (test coverage) — `VersionPanel.tsx` (`restoreErrorKey`, ~L103-107):** the one injected piece with branching logic (two `instanceof` checks → two i18n keys) has no direct test. A regression that swaps branches or drops the `instanceof` check would ship green. Worth a ~3-line test: mock `restoreVersion` to reject with `new VersionSchemaIncompatibleError()` and assert the panel error resolves to `docs.version.errorRestoreIncompatible` vs a plain `Error → docs.version.errorRestore`. *Non-blocking:* the classifier reads correctly, the error classes exist and are thrown, and this exact logic was equally untested on `main` — so this is a pre-existing gap the refactor carries forward, not a new one.

**Nits (P2, optional):**
- The adapter inherits the shell's `defaultFilter='all'`, so opening the panel now fetches a merged manual+auto page on mount, whereas the old panel fetched manual-only and lazy-loaded autosaves on first expand. This is inside the acknowledged mixed-list redesign and the page is still bounded at 25; if mirroring the old on-open network shape matters, `defaultFilter="manual"` would do it. Please confirm the change is intended.
- `docs.version.emptyAuto` ("No auto snapshots yet.") is now orphaned — the single list uses the generic `empty`. Filtering to Auto with zero autosaves shows the generic copy. Harmless; the key can be pruned in a later cleanup (as can the now-orphaned `manualGroup` / `autoGroup`, which predate this PR).
- `countManual` / `countAuto` render as bare "Manual" / "Auto" beside the numbers — mildly terse but acceptable.

## 3. Verification performed

- `pnpm exec vitest run src/versions/` (packages/docs) → **7 files / 75 tests pass** (matches the PR claim).
- `tsc --noEmit -p tsconfig.typecheck.json` → only the **two pre-existing errors** in `dmworkbase/src/i18n/format.ts` and `docs/members/sort.ts`, both outside this diff. The versions files are clean.
- All `docs.version.*` keys the shell references resolve in **both** `en-US` and `zh-CN` (checked programmatically; none missing). The 6 new keys are each added once per locale and all 6 are referenced by the shell; zh-CN translations are idiomatic.
- Second-opinion cross-check: one independent long-context leg ran and surfaced three candidate issues (dropped preview-schema handling; `getVersionState` signal arg; a `&nbsp;`→space change in `DiffView`). All three were **checked and dismissed**: the shell's `defaultErrorKey` preserves the schema-preview messages; `getVersionState`'s third parameter *is* the `AbortSignal`; and `.octo-diff-line` uses `white-space: pre-wrap` so a regular space does not collapse (and the line always renders a sign + space, so it is never empty). A second line-level leg was **unavailable this run** (tooling error) and is recorded as absent, not as tacit approval.

## Verdict

No P0/P1 blockers. Spec compliance ✅ and code quality approved — the sole elevated item is a non-blocking test-coverage suggestion. **Approving.** Addressing the `restoreErrorKey` test and the orphaned i18n keys in a follow-up would be nice-to-have.

boris-clark (approved):
## Review — doc VersionPanel thin adapter

Verified the diff byte-for-byte against the shell baseline (`main`/@d80703a5) and ran the doc suite. Verdict below is per the five review points.

**1. Thin adapter / no regression** — `VersionPanel.tsx` drops the 607-line doc-local panel and returns `<VersionHistoryPanel>` with only the doc-specific injections. list / create / rename / delete / restore / preview all delegate to the shell, which imports those mutations and re-applies the `canSnapshot` / `canRestoreVersion` role gates. `names`, `onClose`, and the read-only `editor` are all forwarded, and `PAGE_SIZE=25` is preserved via the `pageSize` override. No behavior lost.

**2. Contract alignment** — the injected props match `VersionHistoryPanelProps` exactly. `loadPreviewState(seq, signal)` forwards the AbortSignal into `getVersionState(docId, seq, signal)`, which passes it to the axios `{ signal }` — the shell's "MUST honor the AbortSignal" contract is satisfied. The doc passes both `renderDiff` and `getCurrent`, so `canCompare` is true and the compare entry shows (diff enabled, as intended — opposite of the board's no-diff end). `restoreErrorKey` keeps the schema-mismatch message.

**3. No shell pollution** — the real PR diff (`d80703a5..b08d8336`) is exactly 6 files; no `VersionHistoryPanel.tsx` / `raceGuard.ts` / board / sheet files touched. The `previewGuard.ts` + test deletion is correct: it is superseded by the shell's `raceGuard` (last-write-wins + abort), which has its own coverage, and there are no dangling references to the removed module.

**4. i18n** — the 6 new keys (`filterAll` / `filterManual` / `filterAuto` / `countManual` / `countAuto` / `staleNotice`) are present in both `en-US` and `zh-CN` with correct translations, including the previously-noted `staleNotice`. The full `version.*` key set is aligned across locales (no drift), and all 44 shell-referenced keys resolve.

**5. Tests** — `VersionPanel.test.tsx` rewritten to pin adapter behavior and passes (4/4): single mixed list with `kind="all"` on mount and unified counts, filter tabs reload with the chosen kind, preview via `getVersionState` (docId + seq=7 forwarded), and compare rendering the real block-level diff (historical→current). Assertions are concrete, not always-green.

No blockers. Approving.

PR #671, opened 2026-07-12
feat(board): make BoardVersionPanel a thin adapter over the unified VersionHistoryPanel shell

ploy-elison:
## Summary

Makes `BoardVersionPanel` a thin adapter over the shared `VersionHistoryPanel` shell (merged in #656), the same way the doc and sheet panels do. The board panel now owns only its board-specific injections and delegates all list / filter / counts / pagination / save / rename / delete / restore chrome and the race guard to the shell.

This is an independent PR on top of the current `main` — it depends only on the shell, not on the doc or sheet adapters.

## What changed

- **`BoardVersionPanel`** drops its own list/filter/counts/load-more/save/rename/delete/restore chrome and race guard, keeping only:
  - `loadPreviewState` → `getBoardVersionState` (decoded Excalidraw scene),
  - `renderPreview` → a read-only `BoardScenePreview` canvas wrapped in `BoardErrorBoundary`, remounted per version via `key={seq}`,
  - `previewErrorKey` / `restoreErrorKey` → the board's `versionErrorKey` classifier, so the wider 403 / 409 / 413 / 404 (+ schema) failure set keeps its distinct messages.
  - It passes **neither `renderDiff` nor `getCurrent`**, so the shell hides the compare entry — the whiteboard offers only a read-only preview, **no diff**.
- **`BoardShell`** moves the version-history mount from a centered modal to a right-side drawer anchored inside the relative canvas; the read-only scene preview pops in the shell's own centered modal. Role gating shifts to the shell's `canSnapshot` (writer+) / `canRestoreVersion` (admin), matching the board's prior `canEdit` / `canManage` gating.
- **`board.css`** drops the now-unused version-modal / filters / notice rules and widens the shared preview modal (960px via `:has`) with a definite 62vh / 360px min height when it hosts the scene preview, so the Excalidraw canvas measures a real box on mount.

## No diff for the whiteboard

The board deliberately omits `renderDiff` and `getCurrent`. The shell hides the compare entry when either is absent, so the whiteboard version panel is preview-only by design.

## Testing

- `vitest run src/board src/versions` → all green (278 tests).
- `tsc --noEmit` clean for the changed board files (the two pre-existing errors in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts` are unrelated and present on `main`).
- The board delete test was updated to exercise the shell's in-panel delete-confirm box (the shell replaced the former native `window.confirm`), matching the existing restore tests.


OctoBoooot (approved):
## Review: make BoardVersionPanel a thin adapter over the unified shell (#671)

**Verdict:** Approve
The third and final adapter (after #669 sheet, #670 doc) over the #656 shell — board panel drops 428 lines of its own chrome + race guard, injecting only its board-specific pieces. The preview-only design, read-only guarantee, error classification, and role gating all verify clean. No defects.

Risk tier: normal. Base `main`; CI green (**Build passes**, code-review pending). +203/-592, 4 files.

### What I verified
- **"Preview-only, no diff" is correctly enforced by the shell, not a broken button.** The board passes neither `renderDiff` nor `getCurrent`. The shell gates the compare entry on `canCompare = !!(renderDiff && getCurrent)` (`VersionHistoryPanel.tsx:166`); the toggle only renders under `{canCompare && …}` (`:634`), and the diff render is triple-guarded `compare && renderDiff && renderDiff(...)` (`:660`). So no compare affordance appears and it can never fire without the hook. `TCurrent=never` is safe (`getCurrent` optional, never called).
- **Read-only guarantee intact** (version-family invariant): `BoardScenePreview` is `viewModeEnabled` (no edit affordances), seeds Excalidraw from `initialData` once at mount, never writes the live canvas — and it's **unchanged by this PR** (pre-existing component). Restore is delegated to the shell + backend auto-snapshot + Yjs reconcile; the adapter never mutates the live board.
- **`key={seq}` remount is the correct fix**, not a smell: Excalidraw consumes `initialData` only at mount (doesn't reactively track the prop), so switching versions requires a keyed remount or the prior scene would persist while the header advanced. The `seq` is threaded through `TState` to make the key available in `renderPreview` without changing the shared shell contract. Wrapped in `BoardErrorBoundary` so a malformed historical scene degrades to a message instead of unmounting the host.
- **Error classification preserved** — `versionErrorKey` maps `VersionSchemaNewer/Incompatible` + 403/409/413/404 to distinct keys (`boardVersions.ts:82`), and all `docs.board.version.err*` keys resolve in **both** en-US and zh-CN. So the board's richer failure set keeps per-case localized messages via `previewErrorKey`/`restoreErrorKey`.
- **Role gating equivalent** — `BoardShell` passes `role ?? 'reader'`; gating shifts to the shell's `canSnapshot` (writer+) / `canRestoreVersion` (admin), matching the board's prior `canEdit` / `canManage`. No privilege change. The modal→right-drawer move + CSS (drop unused modal rules, widen shared preview modal to 960px with a definite height so the Excalidraw canvas measures a real box) are presentational.

### Praise
- This closes the doc/sheet/board adapter trilogy cleanly, and the board case is the most interesting: it correctly leans on the shell's *absence*-based feature gating (omit the hooks → the compare entry disappears) rather than passing stubs, so "whiteboard has preview but no diff" falls out of the contract instead of needing special-casing. The `key={seq}` remount + `BoardErrorBoundary` handling of Excalidraw's mount-once-initialData and render-time throws is exactly the right treatment for a second live canvas, and the richer `versionErrorKey` classifier is preserved so the board's 403/409/413/404/schema cases keep distinct messages. 592 lines of duplicated chrome retired onto the one hardened shell.

### Note
Diff scoped to `packages/docs/src/board/*`; the shell is not modified (consistent with the scope claim). 278 board+versions tests pass; the delete test was updated to exercise the shell's in-panel confirm box (replacing the former native `window.confirm`).

Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-web` and cleanly turns the board version panel into a shared-shell adapter without obvious security, permission, or race-condition regressions.

**💬 Non-blocking**
- 🟡 Warning — The board-specific `versionErrorKey` is now only wired for preview and restore in [BoardVersionPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-okmhr1an/repo/packages/docs/src/board/BoardVersionPanel.tsx:71). Rename and delete errors fall back to generic shell messages in [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-okmhr1an/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:287) and [VersionHistoryPanel.tsx](/Users/mlamp/.openclaw/tmp/pr-review-okmhr1an/repo/packages/docs/src/versions/VersionHistoryPanel.tsx:308). If preserving the previous board-specific 403/409/404 messaging for rename/delete matters, consider adding `renameErrorKey` / `deleteErrorKey` hooks to the shell.
- 🔵 Suggestion — Consider adding a `BoardShell` integration test for the new drawer mount path in [BoardShell.tsx](/Users/mlamp/.openclaw/tmp/pr-review-okmhr1an/repo/packages/docs/src/board/BoardShell.tsx:1211), since the current adapter tests cover `BoardVersionPanel` behavior but not the host-level modal-to-drawer placement.

**✅ Highlights**
- Role gating correctly delegates to the shared `canSnapshot` / `canRestoreVersion` behavior.
- The board adapter intentionally omits `renderDiff` / `getCurrent`, so the shell hides compare as described.
- Preview loading keeps the board-specific decoded scene path and keyed remount behavior.
- i18n keys used by the shell and board-specific preview/restore errors are present in both locale files.

I could not run the focused test command locally because this worktree has no installed `node_modules`; `vitest` was not found.

OctoBoooot (changes requested):
## Review: #671 — correcting my verdict (Approve → Request changes)

Steve is right and I was wrong to approve. Byte-verified: the shell references 40 `docs.version.*` keys, but **6 are missing from the docs i18n bundle at this head** — so #671, as the first prod mount of the shell on a branch without #670, renders raw key strings on open. Flipping.

Risk tier: normal; `needs-human-review` stands. Base `main`.

### Blocker — 6 shell i18n keys missing at this head → raw keys render
`VersionHistoryPanel.tsx` (the shell) calls `t('docs.version.*')` for 40 keys. At #671's head `d7e912a35cf3`, the docs i18n bundle (`version.*` namespace) is missing **6** of them, in **both** en-US and zh-CN:
`countAuto`, `countManual`, `filterAll`, `filterAuto`, `filterManual`, `staleNotice`.

**Reachable:** these drive the shell's filter tabs (all/manual/auto) and per-tab counts — visible the moment the board version panel opens — plus the stale-notice on a superseded action. With the keys absent, `t()` returns the raw key string, so the user sees literal `docs.version.filterAll` etc. instead of labels.

**Root cause — cross-PR dependency I missed:** those 6 keys are added by **#670** (the doc adapter), not by the shell PR (#656) itself. All three adapters are independent PRs on top of `main`, so at #671's head #670 hasn't merged and the keys don't exist in this branch. #671 is the first adapter to *mount the shell in prod* without #670's additions — so it's the first to expose the gap. (This is the same latent `staleNotice` hole I flagged on #656, now made visible.)

**Fix (trivial):** add the 6 `version.*` keys to `packages/docs/src/i18n/{en-US,zh-CN}.json` in this PR (mirroring what #670 adds), so #671 is self-contained regardless of merge order. Whichever of #670/#671 merges second will have a duplicate-add to reconcile, but each must carry the keys to be independently correct.

### Standing (verified, correct — unaffected)
The board adapter itself is sound: preview-only design correctly enforced by the shell (`canCompare` gating), read-only `BoardScenePreview` (`viewModeEnabled`, unchanged by this PR), `key={seq}` remount + `BoardErrorBoundary`, `versionErrorKey` classifier with all `docs.board.version.err*` keys resolving in both locales, role gating equivalent. The defect is purely the 6 missing shared-shell keys.

### Accountability
I checked the board's *own* `docs.board.version.err*` keys resolve (they do) but did **not** verify the *shell's own* 40 `docs.version.*` keys resolve in #671's branch — I assumed the shell's i18n was self-contained in #656. It wasn't: 6 keys arrive via #670. Lesson: when an independent PR mounts shared code, verify it carries (or depends on) *all* of that shared code's i18n keys in its own branch — not just its adapter-specific ones — because "independent of the other adapters" means it can't rely on their key additions. Credit: Steve (byte-verified the flat-namespace + first-mount transition; both codex and I missed it). Trivial fix, but a real raw-keys-on-screen regression as it stands.

Jerry-Xin (dismissed):
## Re-review: board adapter over the unified VersionHistoryPanel shell (#671)

**Verdict: Request changes**

The board adapter itself is well done — but there is one real, user-visible blocker at this head (`d7e912a3`) that must be fixed in this PR before merge. Verified against the checked-out head branch, not the diff alone.

### 🔴 Blocker — 6 shared-shell i18n keys are missing at this head → raw key strings render

The shell (`VersionHistoryPanel.tsx`) calls `t('docs.version.*')` for 6 keys that do **not** exist in the docs i18n bundle at this head, in **both** locales:

`filterAll`, `filterManual`, `filterAuto`, `countManual`, `countAuto`, `staleNotice`

Byte-verified in this branch:
- Shell references them: `VersionHistoryPanel.tsx:511-513` (filter tabs) and `:516` (`countManual`/`countAuto` in the counts header); `staleNotice` at `:266,:299,:322,:349` (post-mutation soft-refresh).
- Bundle gap: the **top-level** `version.*` block in `i18n/en-US.json` (lines 189-234) and `i18n/zh-CN.json` contains none of the 6. They exist only under the *board-local* `board.version.*` namespace (`docs.board.version.*`), which the shell does not read.
- Fallback: `I18nService.t()` returns the raw `key` when it resolves in neither the active nor default locale and no `defaultValue` is passed (`I18nService.ts:135-139`). So the user literally sees `docs.version.filterAll`, `docs.version.filterManual`, `docs.version.filterAuto`, and the counts/stale strings.

**Reachable, not edge-case:** the three filter tabs render unconditionally the moment the panel opens; the counts header renders as soon as the list returns counts (the board list does). `staleNotice` shows on any save/rename/delete/restore whose follow-up refresh fails.

**Root cause — cross-PR dependency:** these 6 keys are added by **#670** (the doc adapter, still open, base `main`) — confirmed in `gh pr diff 670`. The three adapters are independent PRs on `main`, so at #671's head #670 has not merged and the keys don't exist. #671 is the first adapter to mount the shell in production without #670's additions, so it's the first to expose the gap. #671 touches no i18n file, so its bundle == main's, which also lacks them.

**Why CI is green anyway:** the only i18n guard (`i18n.test.ts`) checks (a) en-US/zh-CN key-shape parity and (b) top-level group existence. Both locales are missing the 6 keys *symmetrically*, so parity passes; the `version` group exists, so the group check passes. Nothing asserts that the shell's `t()` calls actually resolve. The adapter tests don't catch it either — in the test env the docs namespace isn't registered, so `t()` returns raw keys and the tests match buttons *by* those raw-key strings (e.g. `docs.version.filterAuto`); they structurally can't detect a missing translation.

**Fix (trivial, makes #671 self-contained):** add the 6 `version.*` keys to `packages/docs/src/i18n/{en-US,zh-CN}.json` in this PR, mirroring #670. Whichever of #670/#671 merges second reconciles a duplicate add, but each must carry them to be independently correct.

---

### The rest of the adapter is sound (per the five review points)

1. **Thin adapter correctness — OK.** `BoardVersionPanel` (428→47) drops all board-local list/filter/counts/pagination/mutation/race-guard logic and delegates to `<VersionHistoryPanel>`, injecting only `loadPreviewState` (→ `getBoardVersionState`), `renderPreview` (read-only `BoardScenePreview` in `BoardErrorBoundary`, `key={seq}` remount for Excalidraw's mount-once `initialData`), and `previewErrorKey`/`restoreErrorKey` (board's `versionErrorKey` classifier). No functional regression in the adapter logic itself.
2. **Board no-diff (decision #4) — OK.** The adapter passes neither `renderDiff` nor `getCurrent`. The shell gates the compare entry on `canCompare = !!(renderDiff && getCurrent)` (`VersionHistoryPanel.tsx:166`) and only renders the toggle under `{canCompare && …}` (`:634`), so compare is absent for the board. Correctly differs from doc/sheet.
3. **Contract alignment — OK.** Props passed (`docId/role/names/onClose/onRestored/loadPreviewState/renderPreview/previewErrorKey/restoreErrorKey`) match the shell signature; `<VersionHistoryPanel<BoardPreview, never>>` types the never-used `getCurrent`. `BoardShell.tsx` change is purely the mount-point move (modal → right-side drawer) + presentational CSS — no side effects.
4. **No shell/doc/sheet pollution — OK.** Diff is exactly 4 board files (`BoardShell.tsx`, `BoardVersionPanel.tsx`, `board.css`, `__tests__/BoardVersionPanel.test.tsx`); no `VersionHistoryPanel.tsx`/`raceGuard.ts`/doc/sheet files touched.
5. **Tests — real coverage, not evergreen, but blind to the i18n gap.** They assert the shell renders as one mixed list, the board-specific preview loads via `getBoardVersionState`, restore goes through the in-panel confirm box, `versionErrorKey` 413 mapping surfaces, and — correctly — that **no compare button** appears. Caveat noted in the blocker: they can't catch missing translations because the test env renders raw keys by design.

Fix the 6 missing keys and this is good to merge.

yujiawei (changes requested):
# Code Review — PR #671 (octo-web)

Reviewed at head SHA `d7e912a3`, diff against `main` (merge-base `d80703a5`). Scope: 4 files (`BoardVersionPanel.tsx`, `BoardShell.tsx`, `board.css`, `BoardVersionPanel.test.tsx`). Local verification: `vitest run src/board src/versions` → 278 passing; `tsc -p tsconfig.typecheck.json` → only the two pre-existing unrelated errors (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`), as the PR notes.

## Verdict: CHANGES_REQUESTED

The refactor itself is clean and faithful to the "thin adapter" goal, but it is the **first and only** end that actually mounts `VersionHistoryPanel`, and it ships without the i18n keys the shell requires. That produces visibly broken UI. One blocking issue (P1) plus a few advisory items below.

---

## P1 — Blocking

### 1. Missing i18n keys → filter tabs, counts, and the stale-notice render as raw keys

The shell renders six `docs.version.*` keys that **do not exist** in either locale's top-level `version` group:

- `packages/docs/src/versions/VersionHistoryPanel.tsx:511-513` → `filterAll`, `filterManual`, `filterAuto`
- `:515` → `countManual`, `countAuto`
- and the post-mutation `staleNotice` used across save/rename/delete/restore.

`packages/docs/src/i18n/en-US.json` / `zh-CN.json` define these strings only under **`board.version.*`** (lines ~37-56), never under the top-level **`version`** group (starts ~line 189). `I18nService.t()` (`packages/dmworkbase/src/i18n/I18nService.ts:133`) falls back to the **raw key** when a key is absent — there is no cross-subtree fallback from `version.filterAll` to `board.version.filterAll`.

Impact: as soon as a user opens the whiteboard version drawer, the three filter tabs show `docs.version.filterAll` / `filterManual` / `filterAuto`, the counts line shows `docs.version.countManual …`, and any refresh miss shows `docs.version.staleNotice` verbatim. This is user-facing in production, in both locales.

Why it slipped through the gates:
- The shell's own header comment (`VersionHistoryPanel.tsx:17-22`) explicitly states these keys "are intentionally NOT added in this phase … Those keys land in the doc-adapter phase, **when an end first renders the shell**." This PR *is* the first end to render the shell, so it inherited the obligation to add them.
- The i18n parity test (`src/i18n/i18n.test.ts`) only checks zh/en have identical shape — both are missing the same keys, so parity stays green.
- The board unit tests use an identity `t()` mock that returns the key unchanged, and in fact assert on the raw string (`src/board/__tests__/BoardVersionPanel.test.tsx:118` clicks `'docs.version.filterAuto'`), so they cannot catch the gap.

Fix: add `filterAll`, `filterManual`, `filterAuto`, `countManual`, `countAuto`, `staleNotice` to the top-level `version` group in both `en-US.json` and `zh-CN.json` (values already exist under `board.version.*` and can be reused). Consider a follow-up assertion that resolves against the real message tree so a future missing key fails CI.

---

## P2 — Advisory (non-blocking)

### 2. Rename / delete / save lose the board's richer error classifier

The old panel wrapped rename/delete/save errors in the board classifier so the wider board failure set surfaced a distinct message:
- old `BoardVersionPanel.tsx:232` → `versionErrorKey(e, 'docs.board.version.errRename')`, `:252` → `errDelete`, `:159` → `errSave`.

The shell hard-codes generic strings for these three (`VersionHistoryPanel.tsx:257 errorSave`, `:289 errorRename`, `:310 errorDelete`) and only accepts injected classifiers for **preview** and **restore**. So a 403 (epoch/access revoked), 409 (conflict), or 413 (too-large) hit during rename/delete now degrades to a generic "failed to rename/delete" instead of the specific localized cause.

This is a deliberate narrowing (the shell only exposes `previewErrorKey`/`restoreErrorKey`), and rename/delete rarely hit 413, but the 403/409 cases are plausible and the message regression is real. Worth either extending the shell contract with `renameErrorKey`/`deleteErrorKey`, or explicitly acknowledging the reduced fidelity.

### 3. Preview-modal widening depends on `:has()` with no fallback

`board.css:178` sizes the shared preview modal to 960px only via `.octo-modal.docs-version-preview-modal:has(.octo-board-version-preview)`. On engines without `:has()` support the whole rule is dropped and the Excalidraw scene renders in the ~720px text default and looks squished. A prop-threaded modal class (e.g. a `previewModalClassName` on the shell) would be deterministic. Acceptable if the supported-browser matrix already guarantees `:has()`.

### 4. Drawer drops the modal's explicit propagation boundary

Moving from `octo-modal` (which had `onMouseDown={(e) => e.stopPropagation()}`) to `<aside className="octo-doc-drawer">` (`BoardShell.tsx:1211`) removes that explicit stop. The drawer is opaque (`background: var(--octo-bg)`) and stacked above the canvas (`octo-board-version-drawer { z-index: 40 }` over Excalidraw), so pointer bleed-through is unlikely, but there is no focus trap here and the underlying Excalidraw keeps global key handlers. Worth confirming that a key press inside the rename input (e.g. Delete/Backspace) cannot reach the canvas.

### 5. Dead i18n keys left behind

After this refactor, `board.version.filterAll/filterManual/filterAuto/countManual/countAuto/staleNotice/restoredNotice/errLoad/errSave` are no longer referenced anywhere in `src/`. Not harmful, but cleaning them (or, better, relocating the still-needed values to the top-level `version` group per P1) keeps the message tree honest.

---

## Coverage / blind spots

- Verified against the current head SHA and correct merge-base; `mergeable=MERGEABLE`, no sibling-PR compile conflict assessed beyond `tsc` on this branch.
- Confirmed neither the doc nor sheet panel is wired to the shell yet, so the board adapter carries the shell's first-mount obligations alone.
- Not verified at runtime in a real browser: the `:has()` widening, Excalidraw's on-mount measurement of the 62vh/360px box, and drawer focus/propagation behavior — these were reasoned from CSS/markup, not observed.
- The shared REST layer (`versions/api.ts`), `getBoardVersionState`, and the race guard are reused unchanged and were not re-audited.

OctoBoooot (approved):
## Review: #671 — delta re-review (6 shell i18n keys added; blocker fixed)

The new commit adds the 6 missing shared-shell `docs.version.*` keys to both locales — exactly the blocker fix. Byte-verified. Flipping.

Risk tier: normal. Base `main`. Delta = 1 commit, +6/+6 i18n only.

### Resolved — the 6 missing shell i18n keys
`filterAll` / `filterManual` / `filterAuto`, `countManual` / `countAuto`, `staleNotice` are now defined under `version.*` in **both** `en-US.json` and `zh-CN.json`, all resolving. So the shell's filter tabs + counts (unconditionally rendered on panel open) now show translated labels instead of raw `docs.version.*` key strings. i18n parity holds (402/402, zero divergence). Same key set/values as the #669/#670 additions — consistent across the adapter family, and #671 is now self-contained regardless of merge order (the point of the CR: an independent PR mounting the shell must carry the shell's keys in its own branch).

### Standing (verified original round, unaffected)
Board adapter sound: preview-only correctly enforced by the shell's `canCompare = !!(renderDiff && getCurrent)` gating (board passes neither → no compare entry), read-only `BoardScenePreview` (`viewModeEnabled`, unchanged), `key={seq}` remount + `BoardErrorBoundary` for Excalidraw's mount-once-initialData, `versionErrorKey` classifier with all `docs.board.version.err*` keys resolving in both locales, role gating equivalent to prior `canEdit`/`canManage`.

### Note
This completes the adapter trilogy's i18n self-containment: #670 supplied the keys inline originally; #669 and #671 each added them this round (the latent gap was present on both branches — #669 I under-caught originally, #671 I flagged). All three now mount the shell with the full key set resolving. The two prior 🟡s (rename/delete generic error fallback; BoardShell drawer-mount integration test) remain non-blocking follow-ups.

Re-affirming APPROVE — the blocker is cleanly resolved.

Jerry-Xin (approved):
Re-reviewed at head `cd08563d`, byte-verified. Flipping to **APPROVE** — the sole 3×-CHANGES_REQUESTED blocker is fully resolved.

## ✅ Blocker resolved (byte-verified, both locales)
The board version panel mounts the unified `VersionHistoryPanel` shell, which references `docs.version.*` keys the board locale files did not provide — so the panel would render raw key strings. The new commit adds all 6 in **both** en-US and zh-CN:
- `filterAll` (All / 全部), `filterManual` (Manual / 手动), `filterAuto` (Auto / 自动)
- `countManual` (Manual / 手动), `countAuto` (Auto / 自动)
- `staleNotice` (Done, but the list couldn't refresh — reopen to see the latest. / 操作已完成，但列表未能刷新——重新打开可查看最新内容。)

Beyond these, I enumerated every `docs.*` key the shell references (44, incl. `defaultErrorKey` fallbacks) + all 8 board `docs.board.version.err*` keys the adapter's `versionErrorKey` classifier can return — every one present in both locales, full parity (enforced by `i18n.test`). No referenced-but-missing key remains anywhere in the shell or adapter.

## No regression
`compare` is ahead_by:1/behind_by:0 (clean FF, no rebase noise). The single new commit is +12/-0, touching **only** the 2 JSON files. Blob-hash verified: the 4 code/test files (`BoardShell.tsx`, `BoardVersionPanel.tsx`, `BoardVersionPanel.test.tsx`, `board.css`) are byte-identical to the CR'd head, so no code regression is possible. Thin-adapter wiring intact — props/callbacks pass through to the shell cleanly; `dark` consumed only inside `renderPreview`; board deliberately omits `renderDiff`/`getCurrent` so the shell hides compare.

## 🟡 Non-blocking (by design)
BoardVersionPanel passes board-specific error classifiers only for preview/restore; rename/delete fall back to the shell's generic `errorRename`/`errorDelete` (both keys exist in both locales). Intentional, non-blocking.

APPROVE.

ploy-elison (commented):
Re-reviewed at head `cd08563d` after the i18n backfill. Approve — flipping my/the standing Request changes now that the blocker is fixed.

The blocker on the prior head (`d7e912a3`) was the 6 shared-shell `docs.version.*` keys missing from the docs bundle, so mounting the shell would render raw key strings. The delta since that head is a single commit (`compare d7e912a3...cd08563d` → ahead 1, behind 0 — clean forward) touching only the two i18n bundles, +6/-0 each. The adapter refactor (BoardVersionPanel/BoardShell/board.css/tests) is unchanged from the previously-reviewed head.

Blocker resolved — byte-verified against the bundle the shell reads. `VersionHistoryPanel.tsx` calls `t('docs.version.*')`, which resolves to the top-level `version` object; the 6 keys now live exactly there in both locales:

- en: filterAll=All, filterManual=Manual, filterAuto=Auto, countManual=Manual, countAuto=Auto, staleNotice="Done, but the list couldn't refresh — reopen to see the latest."
- zh: filterAll=全部, filterManual=手动, filterAuto=自动, countManual=手动, countAuto=自动, staleNotice="操作已完成，但列表未能刷新——重新打开可查看最新内容。"

Both files are valid JSON, no duplicate keys, values match the doc #670 spec. Keys are under the top-level `version` object, not `board.version`.

The earlier non-blocking note (rename/delete errors falling back to the generic shell message rather than the board-specific `versionErrorKey`) is a minor polish item and does not block this merge.

yujiawei (approved):
# Code Review — PR #671 (octo-web)

Reviewed at head SHA `cd08563d13304f6b216fea3ccdf8050f6f585867` against merge-base `d80703a` (the unified `VersionHistoryPanel` shell from #656). Verified locally: `vitest run src/board src/versions` → 278 passing; isolated docs typecheck (`tsc -p tsconfig.typecheck.json`) clean for changed files (only the two pre-existing, unrelated errors in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts` remain, both present on `main`).

This is a faithful thin-adapter refactor. It does what the description claims, nothing more:

- **Delegation is complete and correct.** `BoardVersionPanel` drops its own list / filter / counts / load-more / save / rename / delete / restore chrome and race guard, and now renders `<VersionHistoryPanel<BoardPreview, never>>` (BoardVersionPanel.tsx:52), injecting only board-specific pieces.
- **Preview wiring is right.** `loadPreviewState` threads `docId` + `AbortSignal` through `getBoardVersionState` → shared `getVersionState` (boardVersions.ts:54; signal honored end-to-end). `renderPreview` returns `<BoardErrorBoundary key={seq}>` wrapping `BoardScenePreview` — the keyed remount is required because Excalidraw only reads `initialData` at mount, and the key sits on the root node returned to the shell's stable render slot, so a version switch forces a full remount.
- **No-diff-by-design holds.** The board passes neither `renderDiff` nor `getCurrent`; the shell computes `canCompare = !!(renderDiff && getCurrent)` (VersionHistoryPanel.tsx:166) and hides the compare button, so the whiteboard is preview-only as intended.
- **Role gating is preserved exactly.** Old board: save/rename = `canEdit` (writer+), restore/delete = `canManage` (admin). Shell: `canSnapshot` = `canEdit`, `canRestoreVersion` = `canManage` (roles.ts:30-41). Equivalent for all four affordances; reader gets preview-only.
- **Wider board error surface is retained for preview + restore.** `previewErrorKey` / `restoreErrorKey` route through `versionErrorKey` (403/409/413/404 + two 409 schema codes → distinct localized messages).
- **i18n is complete.** All `docs.version.*` keys the shell references resolve, and the 6 new shell keys (`filterAll/Manual/Auto`, `countManual/Auto`, `staleNotice`) are added to both en-US and zh-CN with parity.
- **Drawer move.** Version history moves from a centered modal to a right-side `aside.octo-doc-drawer` anchored in `.octo-board-canvas` (`position:relative`), matching the doc/sheet convention; the read-only scene preview still pops in the shell's own centered modal.

No missing scope, no over-build, no divergence from the stated approach.

## 2. Code quality — Approved

No P0/P1. Findings are advisory (P2 / nit); none blocks merge.

**[P2] rename / delete errors lost their board-specific classifier.** In the pre-refactor panel, `commitRename` and `onDelete` both ran errors through `versionErrorKey` (403/409/413/404 → distinct messages). The shell only exposes `previewErrorKey` / `restoreErrorKey` hooks; rename and delete fall back to the generic `docs.version.errorRename` / `errorDelete` (VersionHistoryPanel.tsx:289,310). Net effect: a 403/409 on rename or delete now shows a generic "failed" message instead of the specific one. Not a correctness bug (the user still gets an error and the mutation semantics are unchanged), but a small UX regression versus prior behavior. If you want parity, the shell would need `renameErrorKey` / `deleteErrorKey` hooks; otherwise this is an accepted trade of the unification.

**[P2] orphaned i18n keys.** `docs.board.version.errRename`, `errDelete`, and `errSave` are no longer referenced anywhere in non-test source (they were the classifier targets removed above). Dead keys; safe to prune in a follow-up.

**[P2] `board.css` reaches into the shell's internal DOM class.** The canvas-height rule targets `.docs-version-preview-modal-body .octo-board-version-preview` (board.css:181) — a descendant of the shell's private modal-body wrapper. If the shell renames that wrapper, this rule silently detaches and the canvas can collapse (Excalidraw measures its box once at mount). This follows the existing convention (the shell's own `.docs-version-preview-modal-body .octo-version-preview` does the same for doc/sheet), so it's consistent, but the cross-component coupling is worth a comment or a shell-provided modifier-class seam.

**[P2 → nit] CSS `:has()` for the modal widening degrades gracefully, not badly.** `board.css:178` uses `:has(.octo-board-version-preview)` to widen the shared modal to 960px. On engines without `:has()` (Firefox < 121) the rule is skipped and the modal stays at its 720px base. Worth noting the width degradation — but the canvas is **not** broken: the height rule (`height:62vh; min-height:360px`, board.css:181) is a plain descendant combinator with no `:has()` gate, so the canvas keeps a real box and only renders narrower. Given the modern-browser target this is a nit; a structural `modalClassName`/size prop on the shell would remove the reverse-targeting entirely if you'd rather not rely on `:has()`.

**[P2] `z-index: 40` on the drawer vs Excalidraw floating UI.** The new `.octo-board-version-drawer` sets a raw `z-index: 40` to sit above Excalidraw's in-canvas islands. Excalidraw mounts its own popups/dropdowns at various bands; worth a quick manual check that the drawer doesn't get overlaid by, or intercept clicks from, Excalidraw's floating layers. The centered preview modal itself is safe — it's `position:fixed; z-index:1200` and no ancestor (`.octo-board-canvas` / the drawer) sets `transform`/`filter`/`will-change`, so it escapes the drawer's stacking context and anchors to the viewport.

## 3. Coverage / what was not verified

- The board deleting its own race-guard tests is safe: the race behavior (out-of-order refresh, superseded load-more, abort-on-unmount) is pinned at the shell level by `VersionHistoryPanel.test.tsx` + a dedicated `raceGuard.test.ts`. DRY, not a coverage loss.
- Only `BoardShell` consumes `BoardVersionPanel`, and the export signature is unchanged, so the refactor is self-contained.
- Not runtime/visually verified in a real browser: the drawer/preview-modal/`z-index` interaction with a live Excalidraw canvas (the two z-index and `:has()` notes above are the specific things a quick manual pass should confirm).

## Verdict

Spec-compliant thin-adapter refactor with strong test coverage and no correctness, security, or data-loss concerns. All findings are advisory (P2 / nit). Approving; the rename/delete classifier parity and the orphaned-key cleanup are good follow-ups.

PR #4, opened 2026-05-12
fix: add missing logo images for README

botshen:
## Problem

The README references `./docs/assets/logo-light.png` and `./docs/assets/logo-dark.png` but these files were missing, causing broken images on the GitHub repo page.

## Solution

Added the logo images from `apps/web/public/logo.png` to `docs/assets/`.

## Changes

- Created `docs/assets/` directory
- Added `logo-light.png` for light mode
- Added `logo-dark.png` for dark mode

PR #672, opened 2026-07-12
fix(loop): correctness fixes surfaced by codex adversarial backfill

### LINKED ISSUE #5: test
test

MrAladdin:
Backfilled codex adversarial review over dmloop capability PRs that had merged without a clean codex verdict (the codex endpoint was chronically flaky). Six verified real bugs, one PR (all small, all in `packages/dmloop`, blast-radius contained to the package).

## Fixes

1. **Batch-delete confirm was replayable** (`IssueList.tsx`, `confirmDelete.ts`). Two stacked `Modal.confirm` dialogs could both pass the in-body `busyRef` guard — the first resets it before the second's `onOk` runs — re-issuing `batchDeleteIssues` over the same ids. Guard the confirm-*open* with `confirmingRef` + snapshot `ids`; `confirmDelete` gains `onCancel` to release the guard on every dismiss path (X / Esc / mask / cancel).

2. **"与我相关" could show ALL issues** (`IssuePage.tsx`). When `scope=involves` but the current member id hadn't resolved, the request fell through to an unscoped grouped fetch. Now hold empty until identity resolves (`reload` re-runs on `myMemberId`), never the all-issues fallback.

3. **Keyword search silently ignored active filters** (`IssuePage.tsx`). Full-text search drops status/priority/assignee/creator/project/date/sort, but the controls stayed enabled and selected. Disable them while a keyword is active (grouped view hides keyword, so `searching` is structurally false there).

4. **Member list race** (`SettingsPage.tsx`). `reload` wrote members/invites/spaceHumans unconditionally; a workspace switch mid-flight let a stale result win. Added a `seqRef` guard so only the latest reload lands.

5. **Run-detail polling failure modes** (`RunDetailModal.tsx`). Originally polled forever when the run vanished. `listRuns` swallows task-runs HTTP errors to `[]` (it is shared with `IssueDetailPage`'s full-detail load, so it must not throw), so `[]` can't distinguish "gone" from "transient blip". Tolerate consecutive misses (`MAX_MISSES=3`) before stopping — a transient `[]` recovers next poll, a genuinely gone run stops in ~6s; the `try/catch` still keeps polling on a thrown error (ensureDirectory).

6. **AssigneePicker didn't refresh on invalidation** (`AssigneePicker.tsx`). It read candidates once on mount; a mounted picker stayed stale after a member/agent/squad mutation. Re-fetch candidates on dropdown open.

## Not bugs (verified, intentionally unchanged)
- The status-change client-side trigger gate (`RunConfirmModal.statusMightTrigger`) is a documented best-effort UX check; the backend remains authoritative on whether a run starts.
- The directory in-flight cache join (`ensureDirectory`) is closed by `invalidateDirectory` running adjacent to `setWorkspaceContext` on workspace switch (`LoopPage`).

## Blast radius (Rule 8)
Real diff is **6 files, all in `packages/dmloop`** (`git diff origin/feat/octo-loop...HEAD`). No shared/exported signature changed: `confirmDelete` gained an optional `onCancel` (backward-compatible); `listRuns` is deliberately left unchanged (shared with `IssueDetailPage`); all refs/flags are component-local. No out-of-module consumer affected.

## Verification
- Real browser (dev, Alice / `mv2-e2e-alice` + `loop-dev`): keyword-active disables all filter Selects + datepicker + sort (and clears on empty); "我的回路"/involves and grouped/board/list load with no regression; SettingsPage members load. Race/edge guards (#1/#4/#5/#6) are inherently hard to trigger live and were verified by review + code trace.
- `tsc --noEmit` clean (0 errors in package sources).
- Gates: `/simplify` (clean) + Claude code-reviewer (caught a regression in an earlier version of fix #5 — premature poll-stop on transient `[]` — since fixed with the consecutive-miss counter and re-reviewed clean).

> Codex: adversarial-review **approve / no material findings** on the final diff (incl. the fix #5 `missRef` refinement). Two-model gate complete — Claude code-reviewer caught the earlier fix-#5 regression, codex confirms the final state clean.

### DIFF
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -127,10 +127,16 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
         limit: 100,
       };
       // 「与我相关」= 指派给我 ∪ 我创建 ∪ 间接关联(后端三过滤并集,fan-out 合并);
-      // 其余 scope 单发一次、按 assignee_types 收窄。myMemberId 缺失时 pill 已禁用,不会走到此分支。
+      // 其余 scope 单发一次、按 assignee_types 收窄。
+      // scope=involves 需当前成员的后端 id;未解析出时**不回退成无 scope 全量**(否则「与我相关」
+      // 会错显全部)。清空并结束——myMemberId 在 reload 依赖里,解析后会自动重取。
+      if (scope === "involves" && !myMemberId) {
+        if (my === seq.current) { setGroups([]); setLoading(false); }
+        return;
+      }
       const req =
-        scope === "involves" && myMemberId
-          ? listMyGroupedIssues(myMemberId, gp)
+        scope === "involves"
+          ? listMyGroupedIssues(myMemberId!, gp) // 上方守卫已保证 myMemberId 存在
           : listGroupedIssues({ ...gp, assignee_types: scopeToAssigneeTypes(scope) });
       req
         .then((gs) => { if (my === seq.current) setGroups(gs); })
@@ -229,6 +235,10 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
 
   const title = defaultScope === "involves" ? t("loop.nav.myloop") : t("loop.nav.issue");
 
+  // 关键词走全文搜索端点(独立语义,后端不吃其它筛选/排序)。搜索激活时禁用面板内其它筛选,
+  // 避免它们「看起来生效、实际被忽略」的误导(grouped 视图隐藏关键词,故 searching 恒 false)。
+  const searching = view !== "grouped" && !!f.keyword.trim();
+
   // 「筛选」按钮上的激活计数(仅计当前视图会生效的维度)。
   const activeFilters =
     (view === "grouped"
@@ -257,7 +267,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
         {view === "grouped" ? (
           <Select multiple value={f.gStatuses} onChange={(v) => update({ gStatuses: (v ?? []) as IssueStatus[] })} showClear maxTagCount={2} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
         ) : (
-          <Select value={f.status} onChange={(v) => update({ status: v as IssueStatus | undefined })} showClear dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
+          <Select value={f.status} onChange={(v) => update({ status: v as IssueStatus | undefined })} showClear disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
         )}
       </div>
       <div className="loop-fields__row">
@@ -265,14 +275,14 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
         {view === "grouped" ? (
           <Select multiple value={f.gPriorities} onChange={(v) => update({ gPriorities: (v ?? []) as IssuePriority[] })} showClear maxTagCount={2} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
         ) : (
-          <Select value={f.priority} onChange={(v) => update({ priority: v as IssuePriority | undefined })} showClear dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
+          <Select value={f.priority} onChange={(v) => update({ priority: v as IssuePriority | undefined })} showClear disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
         )}
       </div>
       {/* assignee 单选仅扁平列表/看板(grouped 用 scope 按类型收窄)。 */}
       {view !== "grouped" && (
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.filter.assignee")}</div>
-          <Select value={f.assignee} onChange={(v) => update({ assignee: v as string | undefined })} showClear filter dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.assignee")}>
+          <Select value={f.assignee} onChange={(v) => update({ assignee: v as string | undefined })} showClear filter disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.assignee")}>
             {cands.map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>))}
           </Select>
         </div>
@@ -281,7 +291,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
       {!(view === "grouped" && scope === "involves") && (
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.filter.creator")}</div>
-          <Select value={f.creator} onChange={(v) => update({ creator: v as string | undefined })} showClear filter dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.creator")}>
+          <Select value={f.creator} onChange={(v) => update({ creator: v as string | undefined })} showClear filter disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.creator")}>
             {cands.filter((c) => c.type === "member").map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>))}
           </Select>
         </div>
@@ -292,7 +302,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
           {view === "grouped" ? (
             <Select multiple value={f.gProjectIds} onChange={(v) => update({ gProjectIds: (v ?? []) as string[] })} showClear filter maxTagCount={1} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
           ) : (
-            <Select value={f.project} onChange={(v) => update({ project: v as string | undefined })} showClear filter dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
+            <Select value={f.project} onChange={(v) => update({ project: v as string | undefined })} showClear filter disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
           )}
         </div>
       )}
@@ -305,10 +315,10 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
       <div className="loop-fields__row">
         <div className="loop-fields__label">{t("loop.filter.dateRange")}</div>
         <div className="loop-fields__inline">
-          <Select value={f.dateField} onChange={(v) => update({ dateField: v as IssueDateField })} dropdownClassName="loop-fields__dropdown" style={{ width: 104 }}>
+          <Select value={f.dateField} onChange={(v) => update({ dateField: v as IssueDateField })} disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: 104 }}>
             {ISSUE_DATE_FIELDS.map((d) => (<Select.Option key={d} value={d}>{t(`loop.dateField.${d}`)}</Select.Option>))}
           </Select>
-          <DatePicker type="dateRange" density="compact" value={f.dateRange} onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })} placeholder={t("loop.filter.dateRange")} style={{ flex: 1 }} />
+          <DatePicker type="dateRange" density="compact" disabled={searching} value={f.dateRange} onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })} placeholder={t("loop.filter.dateRange")} style={{ flex: 1 }} />
         </div>
       </div>
       {/* 关键词走全文搜索(独立语义,不与 grouped 组合),故分组视图隐藏。 */}
@@ -340,10 +350,10 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.sort.direction")}</div>
           <div className="loop-fields__inline">
-            <Select value={f.sortBy} onChange={(v) => update({ sortBy: v as IssueSortField })} dropdownClassName="loop-fields__dropdown" style={{ flex: 1 }}>
+            <Select value={f.sortBy} onChange={(v) => update({ sortBy: v as IssueSortField })} disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ flex: 1 }}>
               {ISSUE_SORT_FIELDS.map((s) => (<Select.Option key={s} value={s}>{t(`loop.sort.${s}`)}</Select.Option>))}
             </Select>
-            <Button theme="borderless" disabled={f.sortBy === "position"} icon={f.sortDir === "asc" ? <ArrowUp size={14} /> : <ArrowDown size={14} />} aria-label={t("loop.sort.direction")} onClick={() => update({ sortDir: f.sortDir === "asc" ? "desc" : "asc" })} />
+            <Button theme="borderless" disabled={searching || f.sortBy === "position"} icon={f.sortDir === "asc" ? <ArrowUp size={14} /> : <ArrowDown size={14} />} aria-label={t("loop.sort.direction")} onClick={() => update({ sortDir: f.sortDir === "asc" ? "desc" : "asc" })} />
           </div>
         </div>
       )}
--- a/packages/dmloop/src/pages/SettingsPage.tsx
+++ b/packages/dmloop/src/pages/SettingsPage.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useMemo, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import {
   Typography, Button, Tabs, TabPane, Table, Tag, Select, Spin, Toast, Banner, Modal, Avatar,
 } from "@douyinfe/semi-ui";
@@ -110,6 +110,7 @@ function MembersTab({ workspaceId }: { workspaceId: string }) {
   const [error, setError] = useState<string | null>(null);
   // octo space 成员（仅人，robot===0）+ uid→身份 映射，供选择器候选与列表渲染复用。
   const [spaceHumans, setSpaceHumans] = useState<SpaceMember[]>([]);
+  const seqRef = useRef(0); // reload 请求序号:切 workspace 时只让最新一次落地
   const [selectedUid, setSelectedUid] = useState<string>("");
   const [role, setRole] = useState("member");
   const [adding, setAdding] = useState(false);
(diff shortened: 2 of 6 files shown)


OctoBoooot (approved):
## Review: correctness fixes surfaced by codex adversarial backfill (#672)

**Verdict:** Approve
Six small, well-reasoned correctness fixes across dmloop, each verified to the sink and matching its description. Notably this closes the My-loops `involves` scope bug I flagged as out-of-scope on #668 (tracked from #663) and hardens the #665 batch-delete re-entrancy family. No defects.

Risk tier: normal. Base `feat/octo-loop`; CI green (`code-review` pending). +63/-20, 6 files, all in `packages/dmloop`.

### What I verified (each fix to the sink)
1. **Batch-delete replay guard (`IssueList.tsx` + `confirmDelete.ts`)** — correct. `confirmingRef` is set synchronously at confirm-*open* (before the Modal opens), released in both `onOk`'s `finally` and `onCancel`. This closes the gap the #665 `busyRef` fix left: `busyRef` only guarded entry to `runBatch` (post-dialog), so two stacked dialogs could both pass it; `confirmingRef` guards the dialog-open itself. The `ids` snapshot at open-time is right (runBatch clears `selected`).
2. **"与我相关" no longer shows all issues (`IssuePage.tsx`)** — correct, and it's the exact fix I prescribed on #668/#663. New guard `if (scope === "involves" && !myMemberId) { setGroups([]); setLoading(false); return; }` returns empty instead of falling through to the unscoped `listGroupedIssues`; the `listMyGroupedIssues(myMemberId!, …)` assertion is now guard-safe, and `myMemberId` is in `reload`'s deps so it re-runs on identity resolve. The pre-existing latent regression is now closed.
3. **Keyword search disables ignored filters (`IssuePage.tsx`)** — correct. `searching = view !== "grouped" && !!f.keyword.trim()` gates `disabled` on all filter Selects + datepicker + sort, so controls can't look-active-but-ignored while full-text search is on. Grouped view hides keyword → structurally false there.
4. **Member-list race (`SettingsPage.tsx`)** — correct. Standard `seqRef` latest-wins; all three of `setMembers/setInvites/setSpaceHumans` (and `setError`/`setLoading`) are guarded by `my === seqRef.current`, so a workspace-switch mid-flight can't let a stale result win.
5. **Run-detail polling failure modes (`RunDetailModal.tsx`)** — correct and robust. `missRef` resets on (re)entry and on every successful `fresh`; increments only on `[]`; `stillActive = missRef.current < MAX_MISSES(3)`. A transient `[]` recovers next poll (no premature stop); a genuinely-gone run stops in ~6s (no infinite poll); a thrown error (ensureDirectory) keeps polling (transient infra, not gone). Cleanup `clearTimeout` + `cancelled`-guarded setState throughout.
6. **AssigneePicker refresh on invalidation (`AssigneePicker.tsx`)** — correct. `loadCands` on mount + `onVisibleChange` refetch on dropdown open, so a mounted picker reflects member/agent/squad mutations (the directory cache is invalidated but a mounted picker doesn't auto-reread).

### Praise
- This is a disciplined bug-fix batch: each fix targets the real root cause (guard the dialog-*open* not just the mutation; hold-empty-until-identity-resolves rather than fall through to unscoped; consecutive-miss tolerance rather than stop-on-first-`[]`), the blast radius is honestly scoped (all 6 files in-package, only a backward-compatible optional `onCancel` added to a shared helper), and the trickiest one (#5) explicitly documents why `[]` is ambiguous and was re-reviewed after an earlier premature-stop regression. Fix #2 in particular resolves a latent scope bug that survived several PRs in this area.

### Note (non-blocking)
`confirmDelete` wires `onCancel` but relies on Semi `Modal.confirm` defaults for the wedge-safety of `confirmingRef`: Cancel and Esc both fire `onCancel` (good), mask-click defaults to non-closable and `Modal.confirm` renders no `×`, so there's no dismiss path that skips `onCancel` and wedges the guard. That safety is correct today but depends on those defaults — if this is ever swapped to a declarative `<Modal>` (which renders `×` and may default `maskClosable`), add `maskClosable: false` / handle the `×` path so `confirmingRef` can't wedge. Worth a one-line comment.

Jerry-Xin (dismissed):
The PR is relevant to `Mininglamp-OSS/octo-web`, but one race condition remains in the new assignee refresh behavior.

**🔴 Blocking**

🔴 Critical: `AssigneePicker` can still overwrite fresh candidates with stale ones.  
In [AssigneePicker.tsx](/Users/mlamp/.openclaw/tmp/pr-review-jdtk6eyx/repo/packages/dmloop/src/ui/AssigneePicker.tsx:32), every `loadCands()` result calls `setCands` unconditionally. The directory layer intentionally allows callers from a pre-invalidation build to receive that older result while preventing it from poisoning the shared cache; see [directory.ts](/Users/mlamp/.openclaw/tmp/pr-review-jdtk6eyx/repo/packages/dmloop/src/api/directory.ts:75) and [directory.ts](/Users/mlamp/.openclaw/tmp/pr-review-jdtk6eyx/repo/packages/dmloop/src/api/directory.ts:88). A mounted picker can start request A, a member/agent/squad mutation invalidates the directory, opening the dropdown starts fresh request B, B resolves first, then A resolves later and overwrites the fresh list with stale candidates. The same applies to an older rejected request clearing a newer successful list. Add a local sequence guard, and ideally an unmount guard, so only the latest picker request can update `cands`.

**✅ Highlights**

The `scope=involves` guard in [IssuePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-jdtk6eyx/repo/packages/dmloop/src/pages/IssuePage.tsx:131) correctly prevents falling back to an unscoped grouped fetch.

The batch-delete confirmation guard in [IssueList.tsx](/Users/mlamp/.openclaw/tmp/pr-review-jdtk6eyx/repo/packages/dmloop/src/panel/IssueList.tsx:217) addresses the replayable confirm path with an open-confirm mutex and ID snapshot.

The settings reload sequence guard in [SettingsPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-jdtk6eyx/repo/packages/dmloop/src/pages/SettingsPage.tsx:130) matches the intended latest-request-wins pattern.

OctoBoooot (commented):
## Review: #672 — byte-verifying Jerry-Xin's AssigneePicker race (holding Approve)

**Verdict unchanged: Approve.** The race is real and PR-introduced, but its blast radius is minor (transient, self-correcting, no data loss) — below the REQUEST_CHANGES bar. Recording it as a verified minor and endorsing the fix.

### The race is real (fix #6 introduced it)
`AssigneePicker.tsx:32` `loadCands = () => listAssigneeCandidates().then(setCands).catch(...)`, called on mount (`:33`) and on every dropdown open (`:78` `onVisibleChange`). No seq guard, no mounted guard. Two overlapping fetches (rapid open/close/open, or open during an in-flight mount fetch) can resolve out of order → an older result overwrites a newer `setCands`; `setCands` can also fire after unmount (React dev warning). Before fix #6 this was a single mount fetch with no concurrency, so the on-open refetch is what introduced the window. Jerry-Xin's factual claim is correct.

### Why it's minor, not a blocker
Checked for a higher-severity variant; there isn't one:
- **Transient + self-correcting:** every open refetches, so any stale `cands` is replaced on the next open. No wedged flag (unlike the #656 loadingMore/mounted-ref cases that stuck permanently) — there is no persistent-broken state.
- **No corruption/loss/exposure:** `cands` only drives the dropdown *display*; selection uses the clicked candidate's id via `onChange`. Worst case is a briefly-stale list until reopen (can't pick a just-added agent yet; a stale entry's pick fails server-side). Recoverable, narrow.
- **Narrow window:** requires cache invalidation + genuinely-different results + out-of-order resolution between two rapid opens; the cached fetch usually resolves near-instantly and in order.

By the severity ladder that's a **minor** (recoverable, narrow blast radius), which doesn't mandate REQUEST_CHANGES.

### Recommended fix (Jerry-Xin's 🟡 — right, cheap)
Add a per-call seq + mounted guard, mirroring the `seqRef` pattern this very PR uses in `SettingsPage` (fix #4) and `IssuePage`:
```
const seq = useRef(0);
const loadCands = () => { const my = ++seq.current;
  listAssigneeCandidates().then((c) => { if (my === seq.current) setCands(c); }).catch(() => { if (my === seq.current) setCands([]); }); };
```
(Plus a mounted ref cleared on unmount if you want to silence the after-unmount setState.) Worth doing — and it's consistent to apply the same latest-wins discipline fix #4 already established, so the on-open refetch matches the rest of the PR.

Net: real, PR-introduced, minor. Verdict holds at Approve; the seq guard is a good non-blocking follow-up. Credit: Jerry-Xin for the catch — I'd approved the on-open refetch without scrutinizing its concurrency, and the guard should be added.

Jerry-Xin (approved):
## ✅ APPROVE — correcting my verdict from REQUEST_CHANGES @ 39a9ff5251e2

Withdrawing my earlier CR after byte-verifying the AssigneePicker race through the reachability + escalation gates. The 🔴 was over-graded — it's a real-but-non-blocking hardening 🟡. All 6 fixes in this PR are correct.

### Per-fix (all correct)
1. **Batch-delete replay guard** (IssueList / confirmDelete) — ✅ `confirmingRef` open-mutex + ids snapshot (necessary because `runBatch` clears `selected` on success), guard released on all dismiss paths.
2. **involves unscoped-fallback** (IssuePage:131) — ✅ holds empty until `myMemberId` resolves, no all-issues fallback; `reload` re-runs on `myMemberId`. (This is the #663 fix, correctly applied.)
3. **keyword disables filters** (IssuePage) — ✅ `searching` disables Selects/datepicker/sort.
4. **Members reload race** (SettingsPage:130) — ✅ `seqRef` latest-wins guard, textbook.
5. **Run-poll miss-tolerance** (RunDetailModal) — ✅ `missRef`/MAX_MISSES=3 tolerates transient `[]`, still stops a genuinely-gone run; try/catch keeps polling on throw.
6. **AssigneePicker refresh-on-open** (AssigneePicker.tsx:32/78) — the new `onVisibleChange→loadCands` refetch, and `loadCands` calls `setCands` with no local seq/unmount guard.

### Why fix #6's race is 🟡, not 🔴
`loadCands` → `listAssigneeCandidates()` → `ensureDirectory()` → returns `dir.candidates`. For a *stale overwrite* you'd need ALL of: (a) a directory-mutating action (member/agent/squad/project add/remove/rename → `afterDirectoryMutation`→`invalidateDirectory`) occurring in the sub-second gap between two picker opens; (b) the earlier open's in-flight `build()` resolving *after* the later open's fresh post-invalidation build (same multi-endpoint `Promise.all`, so a narrow non-deterministic reorder); (c) then the earlier build's `.then` returns its `d` to its caller (it correctly does NOT write `_cache` when `myGen !== _gen`, but the waiter still gets the pre-mutation list). Even when all three align, the impact is a **transient stale candidate list that self-corrects on the very next open** (which rebuilds post-invalidation) — **no data exposure, no wrong-scope, no persisted bad state**. That compound, self-correcting, no-escalation profile is a non-blocking correctness nit, not a blocker.

### 🟡 Non-blocking (recommended)
Add a `seqRef` latest-wins guard to `loadCands` (mirroring fix #4's SettingsPage pattern) and/or an unmount guard — the picker is instantiated in list rows / detail pages / modals, so hardening the refetch is cheap and consistent with the discipline the rest of this PR applies. Not required to merge.

### Verdict
**APPROVE.** Fixes #1–#5 are solid; #6's added refetch lacks the latest-wins guard the PR applies elsewhere, but the resulting race is unreachable to a user-visible/persisted defect (self-correcting, same-workspace list, no escalation) → 🟡 hardening, not a block. I over-graded it initially; corrected here.

yujiawei (approved):
# Code Review — PR #672 (octo-web)

Reviewed at head SHA `39a9ff5251e20c77f3ebd5bafce2e90d3e0c7af0` against base `feat/octo-loop` (merge-base `9b8c0761`). Real diff: 6 files, all in `packages/dmloop`, +63/−20.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly the six correctness fixes described, no more, no less. Verified each against the surrounding code:

- **Fix #1 — replayable batch-delete** (`IssueList.tsx`, `confirmDelete.ts`): `confirmingRef` is set synchronously at confirm-*open* time and released on every dismiss path (`onOk` finally + `onCancel`). The `ids` snapshot is captured at open (correct — `runBatch` clears `selected` on success). `confirmDelete` gains an optional `onCancel` (backward-compatible; existing callers in SquadPage/AgentPage/ProjectPage/etc. unaffected). Sound fix.
- **Fix #2 — `involves` shows all issues** (`IssuePage.tsx`): the early guard returns instead of falling through to the unscoped `listGroupedIssues`, and `myMemberId` is in the `reload` dependency array so it re-runs once identity resolves. The non-null assertion `myMemberId!` is justified by the guard directly above. Correct.
- **Fix #3 — keyword search ignoring filters** (`IssuePage.tsx`): `searching = view !== "grouped" && !!f.keyword.trim()` gates `disabled` on the status/priority/assignee/creator/project/date/sort controls. Matches the backend semantics (search endpoint ignores those params). Correct.
- **Fix #4 — member list race** (`SettingsPage.tsx`): standard `seqRef` monotonic-sequence guard applied to all three setters + error + loading. Correct.
- **Fix #5 — infinite run-detail polling** (`RunDetailModal.tsx`): `missRef` consecutive-miss counter with `MAX_MISSES=3`; reset on cleanup and on any successful `fresh`. Tolerates a transient `[]` and stops a genuinely-gone run in ~6s. Correct.
- **Fix #6 — stale AssigneePicker candidates** (`AssigneePicker.tsx`): refetch on `onVisibleChange(true)` in addition to mount. Correct.

No out-of-scope additions, no missing items, no divergence from the stated approach.

## 2. Code quality

**Quality: Approved**

No P0/P1 findings. The fixes are minimal, well-scoped, and each targets a real defect. A few advisory (P2) notes:

- **P2 — RunDetailModal does not surface run disappearance** (`RunDetailModal.tsx:75-76`). When `missRef` hits `MAX_MISSES`, polling stops but `liveRun` is left at its last (active-looking) value, so the UI stays frozen on the stale "running" state with no terminal/error indication. Note this is **not a regression** — the pre-fix code both polled forever *and* never updated `liveRun` in that path, so the frozen display already existed; this PR strictly improves things by stopping the loop. Worth a follow-up to set a terminal/"run no longer available" state when the miss threshold trips, but it does not block merge.
- **P2 — AssigneePicker refetch has no request-ordering guard** (`AssigneePicker.tsx:30-32, 78`). Rapidly toggling the dropdown fires overlapping `listAssigneeCandidates()` calls; a slow earlier response could land after a newer one and briefly show stale candidates. Self-corrects on the next open and the pre-existing mount fetch had the same property, so low impact. A `seqRef` guard (as used in Fix #4) would make it robust.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. All findings are P2/advisory, none blocking.

## 4. Suggestions

- Follow-up: give `RunDetailModal` a terminal state when `MAX_MISSES` trips (e.g. set `liveRun` to a "gone/unavailable" marker) so the UI reflects the run's disappearance rather than freezing on the last active snapshot.
- Optional: add a sequence guard to `AssigneePicker.loadCands` to eliminate the toggle race.

## 5. Additional notes

- **Advisory-leg coverage (disclosure):** two independent advisory legs were attempted. The long-context leg completed and its findings are folded in above (its P1-on-RunDetailModal was verified to be pre-existing, so downgraded to P2; its keyword-`.trim()`-crash flag was a false positive — `Filters.keyword` is typed `string` and initialized `""`). The line-level adversarial leg was **absent** this run (endpoint returned 503 / rejected its injected tool config — the same chronic flakiness this PR's backfill was addressing). Absence is recorded as absence, not as approval.
- **Blind spots not fully verifiable from the diff:** backend response shape of `listRuns`/`searchIssues`; whether `listRuns` swallowing HTTP errors to `[]` can ever mask a real terminal status; live behavior of the race/edge guards (#1/#4/#5/#6), which are hard to trigger manually and were confirmed by code trace.
- `tsc --noEmit` was not independently re-run here (no `node_modules` in the review worktree); the diff is type-consistent on inspection and the PR reports a clean `tsc`.

Jerry-Xin (approved):
The PR is relevant to `Mininglamp-OSS/octo-web` and the dmloop changes are small, scoped, and consistent with the existing package patterns.

💬 Non-blocking
- 🟡 Warning: `packages/dmloop/src/ui/AssigneePicker.tsx:32` and `:78` still allow an older `listAssigneeCandidates()` request to resolve after a newer open-triggered refresh and overwrite fresh candidates. A small sequence guard, like the one added in `SettingsPage`, would make the invalidation refresh fully race-proof.
- 🔵 Suggestion: The new behavior is mostly race/polling driven. Focused tests around `IssueList` confirm re-entry, `SettingsPage` stale reload suppression, and `RunDetailModal` consecutive misses would reduce regression risk.

✅ Highlights
- `IssuePage.tsx:133` correctly prevents the “involves” grouped view from falling back to an unscoped all-issues fetch.
- `IssueList.tsx:148` snapshots selected IDs and guards confirm creation, closing the replayable batch-delete path.
- `SettingsPage.tsx:130` uses a request sequence guard that matches the race being fixed.
- `RunDetailModal.tsx:177` handles transient empty `listRuns` responses without immediately stopping polling.

OctoBoooot (approved):
## Review: #672 — delta re-review (rebase onto merged #668; fixes unchanged)

**Verdict:** Approve (re-affirmed at head `50d7a2bd4b36`)
The new head is my already-approved #672 **rebased onto the now-merged #668** — the six fixes are unchanged. No new #672 work in this push. Approval carries forward.

Risk tier: normal. Base `feat/octo-loop`.

### What changed (isolated to the real #672 commit)
`ahead=2 behind=1` — the compare is rebase-inflated (it pulls in #668's merge `9cc135e0`). Diffing the isolated #672 commit against its parent (`9cc135e0...50d7a2bd`) shows exactly the **same 6 files, same 6 fixes** I verified at `39a9ff5251e2`:
1. batch-delete `confirmingRef` + ids snapshot, 2. involves no-fallback-to-all, 3. keyword disables filters (`searching`), 4. SettingsPage `seqRef` latest-wins, 5. RunDetailModal `missRef` (MAX_MISSES=3), 6. AssigneePicker refetch-on-open. All byte-identical in logic to the approved version.

### Standing 🟡 (unchanged, still non-blocking)
The AssigneePicker request-ordering race remains — confirmed the `seqRef` guard was **not** added in this rebase (`listAssigneeCandidates().then(setCands)` still bare, no latest-wins/unmount guard). Still minor (transient, self-correcting, no data loss / escalation), as calibrated at the prior head and converged with Jerry-Xin. Recommend the `seqRef` guard (mirroring fix #4) as a follow-up — non-blocking.

### Note
Pure rebase re-affirmation — no re-litigation of the six fixes (already verified to the sink and 3-way peer-approved at the prior head). Now correctly parented on merged #668.

yujiawei (approved):
# Code Review — PR #672 (octo-web)

Independent review of the six correctness fixes in `packages/dmloop`. Reviewed against merge-base `9cc135e0` at head `50d7a2bd`. All findings verified against the code, plus a second automated reviewer leg (see Advisory notes).

## Spec compliance

The diff implements exactly the six fixes described in the PR body, nothing more, nothing less:

1. **Batch-delete replay guard** (`IssueList.tsx`, `confirmDelete.ts`) — `confirmingRef` blocks stacking a second confirm dialog before the first `onOk` runs; `ids` snapshot is captured at open time (correct, since `runBatch` clears `selected`); `confirmDelete` gains an optional `onCancel` to release the guard on every dismiss path. ✅
2. **`involves` no longer falls back to all-issues** (`IssuePage.tsx:133-140`) — early return holds empty until `myMemberId` resolves; `reload` depends on `myMemberId` so it re-runs on resolution. The `myMemberId!` non-null assertion is justified by the guard directly above. ✅
3. **Keyword search disables ignored filters** (`IssuePage.tsx:240` + the `disabled={searching}` props) — `searching` is gated on `view !== "grouped"`, matching the backend's search-endpoint semantics. ✅
4. **Member-list race guard** (`SettingsPage.tsx`) — `seqRef` bumped at reload start, all three setters and the error/loading branches gated on `my === seqRef.current`. Correct latest-wins pattern, consistent with the existing `seq` guard in `IssuePage`. ✅
5. **Run-detail polling** (`RunDetailModal.tsx`) — `missRef` + `MAX_MISSES=3` tolerates transient `[]` (which `listRuns` can't distinguish from "gone"), reset on every successful fetch and on effect setup/teardown; `catch` still keeps polling on thrown errors. ✅
6. **AssigneePicker re-fetch on open** (`AssigneePicker.tsx`) — mount fetch preserved, `onVisibleChange` re-fetches on open so post-mutation directory changes surface. ✅

No out-of-scope changes. The only shared/exported signature touched is `confirmDelete`, and the added `onCancel` is optional — all 15+ existing callers remain source-compatible (verified).

**Spec: ✅**

## Code quality

No P0/P1 issues. The concurrency reasoning in fixes #2, #4, #5 is sound and consistent with the existing sequence-guard idiom in the package.

Non-blocking observations (P2 / suggestions, safe to defer or land as-is):

- **P2 — `RunDetailModal` gives up silently after `MAX_MISSES`.** When `missRef` hits the threshold, polling stops but `liveRun` keeps its last (possibly "running") status, so the modal can sit on a stale spinner/active state without telling the user the run is gone. Functionally this is the intended "stop polling" behavior and the modal is dismissible, so it's not a blocker — but a follow-up could surface a "run no longer available" notice or set a terminal state when giving up.
- **P2 (minor) — batch-delete error path.** `runBatch` catches and toasts on failure; `confirmingRef` is released in the `onOk` `finally`, so a failed delete correctly re-enables the button. No issue; noted for completeness.

**Quality: Approved**

## Advisory notes (reviewer coverage)

A second automated reviewer leg was run and raised two items, both assessed here:

- It flagged `f.keyword.trim()` at `IssuePage.tsx:240` as a possible null-deref. **Rejected as a false positive:** `keyword` is typed `string` (Filters interface), initialized to `""`, sourced from a Semi `Input` (always string), and the pre-existing lines 148 and 246 already call `.trim()` on it identically. The new line follows the established-safe pattern.
- It raised the silent-give-up in `RunDetailModal` — same as the P2 above.
- Coverage gap it noted: whether Semi's `Modal.confirm` guarantees `onCancel` on all unmount edge cases. Assessed: `confirmingRef` is component-local, so any missed release dies with the component (remount = fresh `false`); Semi fires `onCancel` on cancel/mask/Esc/close for mounted dialogs. Not a real leak.

Coverage note: review focused on the six changed files and their direct callers. `listRuns`/`listRunMessages` internals and the backend search-endpoint filter semantics were taken as documented in the PR and not independently exercised at runtime. One reviewer leg was unavailable this run (infrastructure fault on its gateway), so this verdict rests on one automated leg plus manual trace.

## Verdict

No P0/P1 blockers. Spec-compliant, blast radius contained to `packages/dmloop`, shared signature change is backward-compatible. Approving. The two P2 items are optional follow-ups.

PR #673, opened 2026-07-12
feat(loop): dispatch new loops via quick-create for fail-fast feedback

MrAladdin:
Upgrades NewLoopPage's one-line dispatch from the generic `createIssue` endpoint to the dedicated `POST /issues/quick-create`, and tightens the dispatch UX.

## Why
`createIssue` with a non-backlog status + agent assignee *does* trigger an agent run on create (backend `WillEnqueueRun`, `RunSourceAssign`), but it gives **no immediate feedback** when the assignee's runtime is offline or its daemon CLI is too old — the loop is created and the run silently never starts. `/issues/quick-create` pre-checks runtime liveness + daemon version server-side and returns a structured `422` (`agent_unavailable` / `daemon_version_unsupported`) plus a `202 { task_id }`, so the user learns immediately.

## Changes
- **api/http.ts** — `LoopApiError` now carries the backend error `code`; `toApiError` reads `response.data.code` and falls back to `reason` for the message (quick-create 422 bodies are `{ code, reason }`). Additive: the new `reason` tier is strictly lower priority than `error`/`message`, and no other endpoint returns a bare `{reason}` error body, so the 45 other `httpPost`/`httpGet` callers are unaffected.
- **api/types.ts + api/issueApi.ts** — `QuickCreateReq` type + `quickCreateIssue()`.
- **pages/NewLoopPage.tsx**
  - The assignee picker only offers AI (`types=["agent","squad"]`), and a loop with no AI can't run, so **dispatch now requires an assignee**: the 派单 button is disabled without one (same as without a prompt). This removes the old no-assignee `createIssue` branch — it only ever produced orphan loops stuck in 待办 that never run. `submit` always quick-creates (`agent_id` XOR `squad_id`). Plain unassigned issues are still created elsewhere (`CreateIssueModal`, used for sub-issues).
  - Maps the two 422 codes to friendly toasts; the async 202 shows a **single** "已派单" toast owned here (accurate wording for async dispatch).
  - Toasts pass explicit `duration: 3` so they auto-dismiss (they were lingering).
- **pages/LoopPage.tsx** — drops the caller's second `Toast.success("已创建")` in NewLoopPage's `onCreated`: it double-toasted and mislabeled the async dispatch as a sync create. NewLoopPage now owns the single accurate toast.
- **i18n** — `newLoop.agentUnavailable` / `daemonUnsupported` (zh + en).

## Blast radius (Rule 8)
Real diff: **7 files, all in `packages/dmloop`** (`git diff origin/feat/octo-loop...HEAD`). No shared signature changed — `LoopApiError` gains an optional field, `toApiError`'s message chain only *appends* a lower-priority fallback, `createIssue` is untouched. `AssigneePicker.onChange` sets `assigneeType`+`assigneeId` atomically (both null or both set from an agent/squad candidate), so the XOR branch is sound. No out-of-`dmloop` consumer affected.

## Verification (real browser, dev — Alice / `mv2-e2e-alice`)
- No assignee (prompt only) → 派单 button **disabled**.
- Assignee (E2E Coworker) + online runtime → 派单 enabled → `POST /issues/quick-create` **202**, single "已派单" toast that **auto-dismisses** (old behavior lingered), page navigates to the board; the async agent then creates the loop.
- Offline runtime (daemon stopped) → `POST /issues/quick-create` **422** — mapped to the friendly `agent_unavailable` toast.
- `tsc --noEmit` clean.

## Gates
`/simplify` (4-agent, clean) + Claude code-reviewer (clean — verified XOR soundness, `toApiError` non-regression, single-toast) + Codex adversarial-review (no material code findings) + real-browser e2e above.

> Note: the async dispatch shows "已派单" and navigates immediately; the created loop appears on the board once the agent builds it (202 + task_id semantics). Deliberately out of scope: client-side CLI-version pre-gate (handled reactively via the 422) and task_id correlation/inbox tracking.

### DIFF
--- a/packages/dmloop/src/api/http.ts
+++ b/packages/dmloop/src/api/http.ts
@@ -79,24 +79,28 @@ client.interceptors.request.use((config) => {
 /* ---------- 结构化错误（供页面展示异常态） ---------- */
 export class LoopApiError extends Error {
   status?: number;
-  constructor(message: string, status?: number) {
+  code?: string; // 后端结构化错误码(如 quick-create 的 agent_unavailable / daemon_version_unsupported)
+  constructor(message: string, status?: number, code?: string) {
     super(message);
     this.name = "LoopApiError";
     this.status = status;
+    this.code = code;
   }
 }
 
 function toApiError(err: unknown): LoopApiError {
   const e = err as {
-    response?: { status?: number; data?: { error?: string; message?: string } };
+    response?: { status?: number; data?: { error?: string; message?: string; code?: string; reason?: string } };
     message?: string;
   };
+  const data = e?.response?.data;
   const msg =
-    e?.response?.data?.error ||
-    e?.response?.data?.message ||
+    data?.error ||
+    data?.message ||
+    data?.reason || // 结构化 422(如 quick-create)只带 {code, reason},reason 是可读文案
     e?.message ||
     "Request failed";
-  return new LoopApiError(String(msg), e?.response?.status);
+  return new LoopApiError(String(msg), e?.response?.status, data?.code);
 }
 
 function clean(params?: Record<string, unknown>): Record<string, string> {
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -3,6 +3,7 @@ import type {
   Issue,
   IssueComment,
   CreateIssueReq,
+  QuickCreateReq,
   UpdateIssueReq,
   ListParams,
   GroupedParams,
@@ -89,6 +90,13 @@ export async function listChildren(id: string): Promise<Issue[]> {
 
 export function createIssue(req: CreateIssueReq): Promise<Issue> {
   return httpPost<Issue>("/issues", req);
+}
+
+// 一句话派单(POST /issues/quick-create):异步建单 + 派给 agent/squad,返回 { task_id }。
+// 相比 createIssue,后端建单前查 runtime 在线 + daemon 版本,离线/过旧当场返 422
+// (LoopApiError.code = agent_unavailable / daemon_version_unsupported)供即时反馈。
+export function quickCreateIssue(req: QuickCreateReq): Promise<{ task_id: string }> {
+  return httpPost<{ task_id: string }>("/issues/quick-create", req);
 }
 
 export function updateIssue(id: string, req: UpdateIssueReq): Promise<Issue> {
--- a/packages/dmloop/src/api/types.ts
+++ b/packages/dmloop/src/api/types.ts
@@ -188,6 +188,17 @@ export interface CreateIssueReq {
   // 新建时绑定已上传附件(上传返回的 id 列表)。
   attachment_ids?: string[];
 }
+// 一句话派单(POST /issues/quick-create):把 prompt 交给 agent/squad,后端异步建单+派单,返 { task_id }。
+// agent_id / squad_id 恰好给一个(后端 XOR 校验)。与 createIssue 的区别:建单前查 runtime 在线 +
+// daemon 版本,离线/过旧当场返 422(agent_unavailable / daemon_version_unsupported)。
+export interface QuickCreateReq {
+  agent_id?: string;
+  squad_id?: string;
+  prompt: string;
+  project_id?: string | null;
+  parent_issue_id?: string | null;
+  attachment_ids?: string[];
+}
 export interface UpdateIssueReq {
   title?: string;
   description?: string | null;
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -654,6 +654,7 @@
     "dispatched": "Dispatched",
     "agentUnavailable": "That AI teammate is offline right now, so it can't be dispatched — wait for it to come online, or pick one that's online.",
     "daemonUnsupported": "That AI teammate's runtime is too old to dispatch safely — please upgrade its runtime and try again.",
+    "attachFailedAbort": "{{count}} attachment(s) failed to upload, so dispatch was cancelled — attachments are the AI's context; please retry.",
     "examplesTitle": "Start from an example",
     "ex1Title": "Polish the demo script",
     "ex1Desc": "Take one line from ticket to loop, end to end.",
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -652,6 +652,8 @@
     "pickAssignee": "Pick an AI teammate",
     "dispatch": "Dispatch",
     "dispatched": "Dispatched",
+    "agentUnavailable": "That AI teammate is offline right now, so it can't be dispatched — wait for it to come online, or pick one that's online.",
+    "daemonUnsupported": "That AI teammate's runtime is too old to dispatch safely — please upgrade its runtime and try again.",
     "examplesTitle": "Start from an example",
     "ex1Title": "Polish the demo script",
     "ex1Desc": "Take one line from ticket to loop, end to end.",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -652,6 +652,8 @@
     "pickAssignee": "选择 AI 队友",
     "dispatch": "派单",
     "dispatched": "已派单",
+    "agentUnavailable": "AI 队友当前不在线，无法派单——请等它上线，或换一个在线的队友。",
+    "daemonUnsupported": "AI 队友的运行环境版本过低，无法安全派单——请先升级它的运行时后重试。",
     "examplesTitle": "从一个例子开始",
     "ex1Title": "打磨演示脚本",
     "ex1Desc": "从建单到回路，把一句话派单全链路走顺。",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -654,6 +654,7 @@
     "dispatched": "已派单",
     "agentUnavailable": "AI 队友当前不在线，无法派单——请等它上线，或换一个在线的队友。",
     "daemonUnsupported": "AI 队友的运行环境版本过低，无法安全派单——请先升级它的运行时后重试。",
+    "attachFailedAbort": "有 {{count}} 个附件上传失败，已取消派单——附件是 AI 的上下文，请重试。",
     "examplesTitle": "从一个例子开始",
     "ex1Title": "打磨演示脚本",
     "ex1Desc": "从建单到回路，把一句话派单全链路走顺。",
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -3,7 +3,6 @@ import {
   Input,
   Button,
   Spin,
-  Toast,
   Select,
   Pagination,
   DatePicker,
@@ -215,7 +214,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   const openNewLoop = () => {
     WKApp.routeRight.push(
       <NewLoopPage
-        onCreated={() => { Toast.success(t("loop.toast.created")); onMutated(); WKApp.routeRight.pop(); }}
+        onCreated={() => { onMutated(); WKApp.routeRight.pop(); }}
       />,
     );
   };
--- a/packages/dmloop/src/pages/LoopPage.tsx
+++ b/packages/dmloop/src/pages/LoopPage.tsx
@@ -79,7 +79,7 @@ export default function LoopPage() {
   // 新建回路 → 唤起 composer 独立页；成功后落到回路看板（新回路即在其中）。
   const openNewLoop = () => {
     WKApp.routeRight.push(
-      <NewLoopPage onCreated={() => { Toast.success(t("loop.toast.created")); openTab("issue"); }} />,
+      <NewLoopPage onCreated={() => { openTab("issue"); }} />,
     );
   };
 
--- a/packages/dmloop/src/pages/NewLoopPage.tsx
+++ b/packages/dmloop/src/pages/NewLoopPage.tsx
@@ -54,7 +54,9 @@ export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
     if (!text || !assigneeId || submitting) return;
     setSubmitting(true);
     try {
-      // 附件先上传拿 id（issue 尚不存在），再随派单绑定。单个失败只提示、不阻断。
+      // 附件先上传拿 id（issue 尚不存在），再随派单绑定。
+      // 派单流的附件是 agent 的上下文:任一上传失败就整体中止派单(不静默用残缺上下文派单——
+      // 异步 fire-and-forget,用户导航走后无法补救),提示重试。全成功才继续。
       let attachmentIds: string[] | undefined;
       if (pendingFiles.length) {
         const ids: string[] = [];
@@ -62,7 +64,10 @@ export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
         for (const f of pendingFiles) {
           try { ids.push((await uploadAttachment(f)).id); } catch { failed++; }
         }
-        if (failed) Toast.error({ content: t("loop.toast.attachFailed", { values: { count: failed } }), duration: 3 });
+        if (failed) {
+          Toast.error({ content: t("loop.newLoop.attachFailedAbort", { values: { count: failed } }), duration: 3 });
+          return; // finally 会复位 submitting;不派单,用户可重试
+        }
         if (ids.length) attachmentIds = ids;
       }
       // 一句话派单 → quick-create:建单前查 runtime 在线 + daemon 版本,离线/过旧当场 422 反馈,
--- a/packages/dmloop/src/pages/NewLoopPage.tsx
+++ b/packages/dmloop/src/pages/NewLoopPage.tsx
@@ -3,7 +3,7 @@ import { Select, Button, Toast, Spin } from "@douyinfe/semi-ui";
 import { ArrowLeft, Paperclip, CornerDownLeft } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
 import type { AssigneeType, Project } from "../api/types";
-import { createIssue } from "../api/issueApi";
+import { quickCreateIssue } from "../api/issueApi";
 import { listProjects } from "../api/projectApi";
 import { uploadAttachment } from "../api/attachmentApi";
 import AssigneePicker from "../ui/AssigneePicker";
@@ -16,15 +16,11 @@ export interface NewLoopPageProps {
   onCreated?: () => void;
 }
 
-// 从 prompt 派生标题：取首个非空行、裁到 ~80 字（其余保留在描述里）。
-function deriveTitle(prompt: string): string {
-  const line = prompt.split("\n").map((l) => l.trim()).find((l) => l.length > 0) ?? "";
-  return line.length > 80 ? line.slice(0, 80) : line;
-}
-
 /**
  * 新建回路独立页（对齐 Figma「把活交给 AI 队友」）：一句话 prompt + 项目/附件 + 指派 AI 队友 → 派单。
- * 映射到现有 createIssue（title 由 prompt 派生，description=prompt，assignee=agent/squad，status=todo 触发派单）。
+ * 派给 agent/squad 恒走 quickCreateIssue(POST /issues/quick-create,建单前查 runtime 在线 + daemon
+ * 版本,离线/过旧当场 422,返回 task_id 由 agent 异步建单)。指派器只给 AI(agent/squad),故须选一个
+ * AI 队友才能派单——无指派时按钮置灰(无 AI 的回路没人跑,不是本页的意图)。
  * 渲染在右主栏（routeRight.push），返回 pop。
  */
 export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
@@ -54,10 +50,11 @@ export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
 
   const submit = async () => {
     const text = prompt.trim();
-    if (!text || submitting) return;
+    // 需 prompt + 指派 AI 队友(agent/squad)。本页只派单,无指派不可提交(按钮已置灰,这里双保险)。
+    if (!text || !assigneeId || submitting) return;
     setSubmitting(true);
     try {
-      // 附件先上传拿 id（issue 尚不存在），再随 createIssue 绑定。单个失败只提示、不阻断建单。
+      // 附件先上传拿 id（issue 尚不存在），再随派单绑定。单个失败只提示、不阻断。
       let attachmentIds: string[] | undefined;
       if (pendingFiles.length) {
         const ids: string[] = [];
@@ -65,24 +62,29 @@ export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
         for (const f of pendingFiles) {
           try { ids.push((await uploadAttachment(f)).id); } catch { failed++; }
         }
-        if (failed) Toast.error(t("loop.toast.attachFailed", { values: { count: failed } }));
+        if (failed) Toast.error({ content: t("loop.toast.attachFailed", { values: { count: failed } }), duration: 3 });
         if (ids.length) attachmentIds = ids;
       }
-      await createIssue({
-        title: deriveTitle(text),
-        description: text,
-        status: "todo",
-        priority: "none",
-        assignee_id: assigneeId,
-        assignee_type: assigneeType,
+      // 一句话派单 → quick-create:建单前查 runtime 在线 + daemon 版本,离线/过旧当场 422 反馈,
+      // 返回 task_id(异步:agent 稍后建单并跑)。assignee 恒为 agent/squad(指派器只给 AI)。
+      await quickCreateIssue({
+        ...(assigneeType === "squad" ? { squad_id: assigneeId } : { agent_id: assigneeId }),
+        prompt: text,
         project_id: projectId,
         attachment_ids: attachmentIds,
       });
-      // 创建成功后由调用方负责导航(pop 回列表 / 切到回路看板)——此处不再自行 back(),
-      // 避免与调用方的 replaceToRoot 叠加导致把新根也 pop 掉、右栏变空。
+      // 单一 toast 归属此处(异步派单的准确措辞);duration 显式设 3s 保证自动消失。
+      Toast.success({ content: t("loop.newLoop.dispatched"), duration: 3 });
+      // 成功后由调用方负责导航(切回回路看板)——此处不再自行 back(),避免叠加把新根 pop 掉。
       onCreated?.();
     } catch (e) {
-      Toast.error((e as Error)?.message ?? t("loop.toast.saveFailed"));
+      // quick-create 结构化 422:LoopApiError.code 映射成友好提示(即时失败反馈)。
+      const err = e as { code?: string; message?: string };
+      const msg =
+        err.code === "agent_unavailable" ? t("loop.newLoop.agentUnavailable")
+        : err.code === "daemon_version_unsupported" ? t("loop.newLoop.daemonUnsupported")
+        : (err.message ?? t("loop.toast.saveFailed"));
+      Toast.error({ content: msg, duration: 3 });
     } finally {
       setSubmitting(false);
     }
@@ -150,7 +152,7 @@ export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
                   onChange={(id, type, name) => { setAssigneeId(id); setAssigneeType(type); setAssigneeName(name); }}
                 />
               </div>
-              <Button theme="solid" loading={submitting} disabled={!prompt.trim()} onClick={submit} icon={<CornerDownLeft size={14} />} iconPosition="right">
+              <Button theme="solid" loading={submitting} disabled={!prompt.trim() || !assigneeId} onClick={submit} icon={<CornerDownLeft size={14} />} iconPosition="right">
                 {t("loop.newLoop.dispatch")}
               </Button>
             </div>


OctoBoooot (approved):
## Review: dispatch new loops via quick-create for fail-fast feedback (#673)

**Verdict:** Approve
Upgrades NewLoopPage's one-line dispatch from `createIssue` to the dedicated `/issues/quick-create` (server-side runtime/daemon pre-check → structured 422 for immediate feedback). The shared error-parsing change is genuinely additive, the XOR dispatch is sound, and it removes a real orphan-loop footgun. No defects.

Risk tier: **high** (API surface: http.ts / issueApi.ts / types.ts). Base `feat/octo-loop`; CI green (`code-review` pending). +57/-28, 7 files, all in `packages/dmloop`.

### What I verified (the high-risk shared path first)
- **`toApiError` change is additive, cannot regress the 45 other callers.** New message chain: `data.error || data.message || data.reason || e.message || "Request failed"`. `data.reason` is inserted **strictly below** `error`/`message` (the fields other endpoints return) and **above** `e.message` — so any endpoint returning `{error}`/`{message}` short-circuits before `reason` and is byte-unchanged; `reason` only fires for the quick-create `{code, reason}` 422. `LoopApiError.code` is a new **optional** ctor arg (backward-compatible). `data` access is null-safe. Confirmed the shared blast radius is contained.
- **XOR dispatch correct.** `...(assigneeType === "squad" ? { squad_id: assigneeId } : { agent_id: assigneeId })` sends exactly one id, keyed on `assigneeType`. Sound because the picker is `types=["agent","squad"]` and `AssigneePicker.onChange` sets `assigneeType`+`assigneeId` atomically, so `assigneeType` is always agent/squad when `assigneeId` is set. `QuickCreateReq` types both as optional; the XOR is enforced at the call site.
- **Require-assignee gating is correct + double-guarded** (a real improvement). Button `disabled={!prompt.trim() || !assigneeId}` + `submit`'s `if (!text || !assigneeId || submitting) return`. This removes the old no-assignee `createIssue` branch, which only ever produced orphan loops stuck in 待办 that never run (the picker offers only AI, and an AI-less loop can't run). Plain unassigned issues remain available via `CreateIssueModal` (sub-issues) — no capability lost.
- **422 → friendly toast (the fail-fast payoff).** `err.code` (now populated by `toApiError`) maps `agent_unavailable` / `daemon_version_unsupported` to i18n toasts, else falls back to `err.message`. All three new keys (`agentUnavailable`/`daemonUnsupported`/`dispatched`) present in **both** locales.
- **Single accurate toast + #668 fixes preserved.** NewLoopPage now owns one "已派单" toast (`duration: 3` auto-dismiss); `LoopPage.onCreated` drops its `Toast.success("已创建")` (fixes the double-toast + sync-vs-async mislabel) while **keeping `openTab("issue")`** — so the #668 routing fix (caller owns navigation, no double-`back()`) is intact. `deriveTitle` removed cleanly (0 remaining refs; quick-create derives the title server-side from `prompt`).

### Praise
- The high-tier shared-path change is handled exactly right: the new `reason` fallback is ordered *below* the fields every other endpoint uses, so a broad error-parsing change provably can't degrade any existing caller's message — and that ordering is spelled out in a comment. Switching to `/issues/quick-create` turns a silent-failure mode (offline runtime → loop created, run never starts, no feedback) into an immediate structured 422, which is the actual UX goal; and disabling dispatch without an assignee removes the orphan-loop branch rather than papering over it. It also preserves the #668 routing/single-toast fixes instead of regressing them.

### Note (nit, non-blocking)
`LoopPage.onCreated` no longer interpolates `t` in the dropped toast — if `t` isn't used elsewhere in that scope it could be an unused-var lint (tsc passed, so likely still used). Harmless either way.

Jerry-Xin (dismissed):
Summary: The PR is relevant to `Mininglamp-OSS/octo-web`, but the new asynchronous quick-create flow can leave the issue board stale after dispatch.

**🔴 Blocking**
- 🔴 Critical: `quickCreateIssue()` returns only an accepted async task, but the UI discards `task_id` and navigates/reloads immediately. If the issue is created after that reload, there is no follow-up list refresh, so the user may land on a board that never shows the newly dispatched loop until a manual refresh or unrelated state change. `NewLoopPage` drops the response at `packages/dmloop/src/pages/NewLoopPage.tsx:70`, then calls `onCreated` at `:79`; `LoopPage` only remounts the board once at `packages/dmloop/src/pages/LoopPage.tsx:82`; `IssuePage` only does an immediate `onMutated()` reload at `packages/dmloop/src/pages/IssuePage.tsx:218`, while its interval only refreshes running snapshots, not issue lists, at `:193-201`. Please keep enough correlation to refresh when the async task produces an issue, or add a bounded post-dispatch polling/reload path.

**💬 Non-blocking**
- 🟡 Warning: The secondary `NewLoopPage` entry from `IssuePage` still emits the old caller-owned success toast at `packages/dmloop/src/pages/IssuePage.tsx:218`. Since `NewLoopPage` now owns the dispatch toast at `packages/dmloop/src/pages/NewLoopPage.tsx:77`, this path still double-toasts and uses stale “created” wording.

**✅ Highlights**
- `LoopApiError.code` is additive and preserves the existing `error` / `message` precedence before falling back to `reason`.
- The picker restriction to runtime-backed assignee types matches the quick-create XOR payload intent.
- `git diff --check` is clean.

OctoBoooot (commented):
## Review: #673 — byte-verifying Jerry-Xin's two findings (holding Approve, correcting one thing I missed)

**Verdict unchanged: Approve.** Both findings are real; neither is a blocker. But I have to correct myself on the double-toast — I verified only one of the two NewLoopPage callers and wrongly generalized.

### 🟡 Double-toast — CONFIRMED, and the PR's "single-toast" fix is incomplete
There are **two** NewLoopPage mount sites with different `onCreated`:
- `LoopPage:onCreated` = `{ openTab("issue") }` — toast correctly dropped (what I checked).
- **`IssuePage.tsx:218:onCreated` = `{ Toast.success(t("loop.toast.created")); onMutated(); WKApp.routeRight.pop(); }`** — **still fires** the "已创建" toast.

So opening New-loop from the **board `+` buttons** (`:405/:421`) produces **two** toasts on success: NewLoopPage's own "已派单" (`newLoop.dispatched`) **plus** IssuePage's "已创建" (`toast.created`) — the exact double-toast + stale-sync-wording the PR claimed to fix. The fix removed it from the LoopPage caller but **missed the IssuePage caller**. I confirmed LoopPage and over-generalized — my miss. Fix: drop the `Toast.success` from IssuePage's `onCreated` too (keep `onMutated()` + `pop()`), so NewLoopPage owns the single accurate toast on both paths. Non-blocking (cosmetic).

### 🔴 (→ minor) Async-dispatched loop doesn't auto-appear in the list — real, but the PR's documented async tradeoff
Quick-create is **202 async** (the agent builds the loop later). On success:
- From IssuePage, `onCreated` → `onMutated()` = `reload()` — but that reloads *now*, before the async loop exists, so it won't include it. The 15s `refreshRunning` poll only updates the **running-chip set** (`getAgentTaskSnapshot`, `:187-190`), **not the issue list** — so the newly-built loop doesn't surface until a manual reload / re-navigation.
- From LoopPage, `openTab("issue")` remounts IssuePage → `reload()`, still racing the async build.

So Jerry-Xin is factually right: there's no list-poll to surface the async-built loop. **But this is the inherent, documented tradeoff of the fail-fast async model** — the PR explicitly states "the created loop appears on the board once the agent builds it (202 + task_id semantics)" and calls task_id correlation / inbox tracking "deliberately out of scope." The loop *is* created and *will* appear on any later reload; the running-chip poll reflects activity. It's a known UX limitation, not a correctness defect → **minor**, not a blocker. Worth a follow-up (a list-poll or task_id-completion refresh) but explicitly scoped out here.

### Net
Verdict holds at Approve — the core change (quick-create dispatch, additive `toApiError`, XOR, require-assignee) is sound and verified. Two non-blocking items: the IssuePage double-toast (the PR's own single-toast fix is incomplete — worth fixing here since it's the same intent) and the async no-auto-refresh (documented tradeoff, follow-up). Credit: Jerry-Xin — the double-toast is a real miss on my part, and the async-refresh is a fair UX flag.

yujiawei (changes requested):
# Code Review — PR #673 (octo-web)

Reviewed at head SHA `2c7fdd2bc00cac4f4b43737be9756459f387daf9` against merge-base with `feat/octo-loop`. Diff is 7 files, all in `packages/dmloop`.

The core change is well-scoped and the API plumbing is sound: `LoopApiError` gains an optional `code`, `toApiError` appends `data.reason` as a **lower-priority** fallback (existing `error`/`message` callers are unaffected), and `quickCreateIssue` is additive. The assignee-required gating and the `agent_id`/`squad_id` XOR spread are internally consistent. However, the PR's headline UX promise — *"a single accurate toast owned by NewLoopPage"* — is only delivered on one of the two live entry points, and the async-dispatch visibility has a real gap. Requesting changes.

## Blocking

### 1. Double toast + stale "Created" wording survives on the primary entry point — `pages/IssuePage.tsx:218`

`NewLoopPage` is rendered from **two** live callers, not one:

- `pages/LoopPage.tsx:82` (sidebar "New Issue") — **fixed** by this PR (duplicate toast removed).
- `pages/IssuePage.tsx:218` (the default `issue` tab's toolbar button at line 405 **and** the empty-state button at line 421) — **not touched**:

```tsx
onCreated={() => { Toast.success(t("loop.toast.created")); onMutated(); WKApp.routeRight.pop(); }}
```

`NewLoopPage.submit()` now owns its own success toast (`t("loop.newLoop.dispatched")` → "已派单"). So on the IssuePage path the user sees **two** toasts, and the second one is `loop.toast.created` → "已创建" — the exact sync-create mislabel this PR set out to eliminate for what is now an async dispatch. Because `issue` is the default tab (`LoopPage.tsx:46` `useState<TabKey>("issue")`), this is the most common path, not an edge case.

The PR description states *"NewLoopPage now owns the single accurate toast"* — that is true for the sidebar caller only. The fix is the same one-liner already applied to `LoopPage`: drop `Toast.success(t("loop.toast.created"))` from `IssuePage`'s `onCreated`, keeping `onMutated(); WKApp.routeRight.pop();`.

### 2. Async 202 → immediate list reload races the backend; the new loop is not shown — `pages/NewLoopPage.tsx:75-79`

`quickCreateIssue` returns `{ task_id }` because issue creation is now **asynchronous** (the agent builds the loop after the 202). The `task_id` is discarded and `onCreated?.()` fires immediately:

- `LoopPage` caller → `openTab("issue")` re-mounts `IssuePage`, which fetches the list *before* the backend job has created the loop.
- `IssuePage` caller → `onMutated()` = `reload(); refreshRunning();` — a one-shot reload, again before the loop exists.

After that, only `setInterval(refreshRunning, 15000)` (`IssuePage.tsx:196`) polls, and it refreshes **running indicators**, not the issue list itself. Net effect: the user dispatches, gets "已派单", lands on a board that does **not** contain their new loop, and there is no pending/optimistic affordance and no list re-fetch to make it appear. For a feature whose stated goal is *fail-fast feedback*, "I dispatched and see nothing" is a confusing outcome.

The PR body acknowledges task_id correlation as deliberately out-of-scope, which is a reasonable line to draw — but the *absence of any visible pending state* is the part that hurts. At minimum, surface a pending placeholder or a short delayed re-fetch so the dispatched loop reliably appears; otherwise the async semantics silently regress perceived reliability versus the old synchronous `createIssue`.

## Non-blocking (suggestions)

### P2 — Attachment upload failing open drops AI context — `pages/NewLoopPage.tsx:59-67`

A failed upload only toasts and then proceeds to `quickCreateIssue` with the surviving (or zero) attachments. For a human assignee this is tolerable, but the assignee here is always an AI agent whose output is deterministic in its inputs. A prompt like "summarize the attached PDF" with a silently-dropped attachment guarantees a bad run. Consider blocking submission (or requiring explicit confirmation) when an attachment the prompt likely depends on fails to upload.

### P2 — XOR spread coerces any non-squad type into `agent_id` — `pages/NewLoopPage.tsx:71`

```ts
...(assigneeType === "squad" ? { squad_id: assigneeId } : { agent_id: assigneeId })
```

Today `AssigneePicker` is constrained to `types={["agent","squad"]}`, so this is correct. But it is a catch-all: `AssigneeType` also includes `"member"`, and if the picker is ever widened (or state is set inconsistently) a human/member id would be submitted as `agent_id`. An explicit `assigneeType === "agent"` branch that otherwise refuses to submit would fail loudly instead of sending a wrong-typed id to the backend.

### Nit — `t("loop.toast.created")` becomes dead once caller #1 (IssuePage) is fixed

Only if no other module still references it (it is used by several other pages, so likely keep it).

## What I verified

- `toApiError` reason-fallback is strictly lower priority than `error`/`message`; no other endpoint returns a bare `{reason}` body, so the change is non-regressive for existing callers. ✅
- i18n: new keys live under the `newLoop` object and are served through `registerNamespace("loop", …)` (`module.tsx:54`), so `t("loop.newLoop.agentUnavailable" | "daemonUnsupported" | "dispatched")` resolve in both zh-CN and en-US. ✅
- `deriveTitle` removal has no remaining references. ✅
- `CreateIssueModal` (sub-issue creation) still uses `createIssue` and is unaffected. ✅
- Local `tsc` could not run cleanly here (missing `node_modules` — every error is a bare module-resolution failure, not a diff defect); the diff itself is type-consistent.

## Coverage / blind spots

- Not verified: the backend `POST /issues/quick-create` contract — specifically whether `project_id: null` (sent explicitly, `NewLoopPage.tsx:73`) is accepted the same as omitted, and the exact 422 body shape (`{ code, reason }`). The client mapping assumes both; worth a contract check.
- Not exercised at runtime: the two 422 → toast mappings and the async board-refresh behavior described above (reviewed statically).

OctoBoooot (commented):
## Review: #673 — byte-verifying yujiawei's three findings (holding Approve; severity calibration)

**Verdict unchanged: Approve.** All three are real, and yujiawei's #3 (attachment-fail) is a fair new catch — but none rises to blocker on the bytes. Grading on evidence, not on the 🔴 count.

### #1 IssuePage double-toast — real, but 🟡 not 🔴
Confirmed (already flagged): `IssuePage.tsx:218:onCreated` still fires `Toast.success("已创建")` on top of NewLoopPage's "已派单", so the board `+`-button path double-toasts with stale wording. But a duplicate/mis-worded toast has **no functional or data impact** — it's the textbook low-severity case. **🟡.** Fix: drop the toast from IssuePage's `onCreated` (the PR's single-toast fix missed this caller).

### #2 Async loop not shown until refresh — real, but minor (documented tradeoff)
Confirmed: quick-create is 202-async, the immediate `reload()` runs before the agent builds the loop, and the 15s poll only refreshes the running-chip set, not the list. **But the loop IS created server-side and appears on any reload/re-nav** — self-correcting, no data loss — and the PR explicitly scopes task_id correlation / completion-refresh out. Escalating a documented, self-correcting UX lag to a merge-blocker would over-grade it. **Minor**, worth a follow-up (a "pending dispatch" chip or task_id-completion refresh).

### #3 Attachment-upload failure still dispatches — real, the sharpest of the three, minor
Byte-verified (`:59-74`): on partial attachment-upload failure, `submit` toasts `attachFailed` but **still dispatches** with only the uploaded ids (or `undefined` if all failed). yujiawei's insight is apt — for a **dispatch** flow those files are the **agent's context**, so the agent may run on incomplete input, and since it's async fire-and-forget the user can't easily correct it after navigating away. That's more consequential than a comment-attachment failure. **But**: the user *is* notified (toast with failed count), the dispatch carries what uploaded, and it's recoverable (re-dispatch). It's a genuine UX/quality call — "block dispatch or require confirm when context files failed" is a legitimate improvement — but with user notification and recoverability it's a **minor**, not a silent-corruption blocker. Recommend: on any attachment failure in the dispatch path, either abort with a retry prompt or confirm "dispatch without the failed files?" — since the agent's output quality depends on them.

### Net
Verdict holds at Approve — the core change (quick-create dispatch, additive `toApiError`, XOR, require-assignee, #668-routing preserved) is sound and byte-verified. Three non-blocking items, in priority order: **#3 attachment-fail dispatch** (best fix: block/confirm — real quality impact), #1 IssuePage double-toast (drop the toast), #2 async refresh (follow-up). I differ from the 🔴 grading on #1/#2 (cosmetic + documented-self-correcting), and rate #3 the one most worth addressing. Credit: yujiawei — #3 is a fair new catch I'd not weighted for the dispatch-context angle.

MrAladdin:
Thanks @yujiawei @Jerry-Xin @OctoBoooot — addressed the review in `0747badc`:

- **#1 double-toast (Rule 9 class miss)** — fixed. There were two NewLoopPage mount sites; the first pass only de-duped `LoopPage`. `IssuePage.tsx:218` (the board "+" entry) still fired `Toast.success("已创建")` on top of "已派单". Dropped it — both entries now rely solely on NewLoopPage's single accurate toast; removed the now-unused `Toast` import from IssuePage.
- **#3 attachment-fail still dispatches (yujiawei)** — fixed. For a dispatch flow the attachments are the agent's context, so any upload failure now **aborts** the dispatch with a retry prompt (`loop.newLoop.attachFailedAbort`) instead of firing quick-create with partial/empty context. The agent never runs on incomplete input. (Accepted tradeoff: a partial-upload abort leaves the already-uploaded files as server-side orphans until retry — noted in-code; safer than dispatching incomplete context.)
- **#2 async list-staleness (Jerry-Xin 🔴)** — kept as a documented follow-up, per @OctoBoooot's byte-verified calibration: the loop **is** created server-side and appears on any reload/re-nav (self-correcting, no data loss), and `task_id` correlation / completion-refresh is explicitly out of scope for this PR. Follow-up candidate: a "pending dispatch" chip or task_id-completion refresh.

Re-gated: Claude code-reviewer + Codex adversarial-review both clean on the fix; tsc clean.

OctoBoooot (approved):
## Review: #673 — delta re-review (double-toast + attachment-fail 🟡s fixed)

**Verdict:** Approve (re-affirmed at head `0747badc317e`)
The new commit fixes both actionable 🟡s from the review — the IssuePage double-toast and the attachment-fail-still-dispatches concern. Byte-verified. Approval carries forward.

Risk tier: high (unchanged); base `feat/octo-loop`.

### Resolved — IssuePage double-toast
`IssuePage.tsx:214:onCreated` drops `Toast.success(t("loop.toast.created"))` → now `{ onMutated(); WKApp.routeRight.pop() }`, and the now-unused `Toast` import is removed (no lint debt). Both callers (LoopPage + IssuePage) now leave the single accurate "已派单" toast to NewLoopPage. Double-toast gone on all paths.

### Resolved — attachment-fail now aborts dispatch (the stronger fix)
`NewLoopPage.tsx:67-70`: on any attachment upload failure, it now shows a dedicated `attachFailedAbort` message and **`return`s before `quickCreateIssue`** — aborting rather than dispatching with incomplete context. Verified the early `return` is inside the `try` (`:56-85`), so it falls through to `finally` (`:93-95`) → `setSubmitting(false)`; the button re-enables, no wedged busy state, user retries. New i18n key in both locales with accurate wording ("attachments are the AI's context; please retry"). This is the right call for the async fire-and-forget dispatch flow — abort (safer) over confirm, since a dispatched loop can't be corrected after navigation.

### Remaining (non-blocking, unaddressed by design)
- **Async board refresh** (yujiawei #2 / my #2): quick-create's 202 means the loop appears on a later reload, not immediately; the 15s poll only refreshes running-chips, not the list. Documented async tradeoff (task_id correlation scoped out) — self-correcting, no data loss. Follow-up (pending-placeholder or completion-refresh) would polish it; not blocking.

### Standing (verified prior rounds)
Core change sound: `toApiError` `reason` fallback is lowest-priority (no regression to ~45 callers), quick-create XOR dispatch, require-assignee removes the orphan-loop footgun, 422→friendly-toast fail-fast, #668 routing/single-toast preserved.

### Note
Good, focused fix commit — both actionable review 🟡s resolved with the right shapes (drop the redundant toast; abort-on-context-failure with a clear message). Credit: yujiawei + Jerry-Xin for the findings. The one remaining item (async refresh) is the documented tradeoff, non-blocking.

Jerry-Xin (dismissed):
Re-reviewed on the new head `0747badc`. Two of the three prior findings are cleanly fixed and byte-verified; one core one remains, so I'm keeping this at REQUEST_CHANGES.

## ✅ Fixed (byte-verified @ 0747badc)
- **Double-toast (IssuePage `onCreated`)** — `IssuePage.tsx:214-217` `onCreated` now only does `onMutated(); routeRight.pop();`; the `Toast.success(loop.toast.created)` and the now-unused `Toast` import are removed. Single "已派单" toast owned by `NewLoopPage`. ✅
- **Attachment-fail still dispatched** — `NewLoopPage.tsx:67-70` now `Toast.error(...attachFailedAbort)` **+ `return;`** before `quickCreateIssue`, so a partial/failed upload aborts the dispatch instead of firing the agent on incomplete context. New i18n key `loop.newLoop.attachFailedAbort` present in both locales; `finally` resets `submitting` for retry. ✅

## 🔴 Blocking (UNCHANGED on this head)
**Dispatched loop does not auto-appear on the board (async no-refresh).**
The fix commit left this path completely untouched. Quick-create is 202 fire-and-forget — the agent creates the issue *later*. But `onCreated` → `onMutated()` → `reload()` (IssuePage:200/103) runs *immediately*, before the issue exists, so it fetches a list that doesn't yet contain the new loop. The 15s `setInterval` (IssuePage:195) only calls `refreshRunning()` (IssuePage:186), which rebuilds the running-chip Set from `getAgentTaskSnapshot()` keyed on existing `issue_id` — it never re-runs `reload()`. There is no optimistic/pending list entry and no `task_id` correlation. Net effect: the user dispatches a loop and it silently doesn't appear until a manual reload happens to land after the agent created it.

This is the *core UX promise* of a "dispatch via quick-create for fail-fast feedback" PR, so I can't treat it as an eventual-convergence follow-up.

**Concrete asks (either is sufficient):**
1. Insert an optimistic/pending list entry keyed on the returned `task_id` from the 202 response, and reconcile it when the real issue surfaces; **or**
2. Make the 15s poll (or a short-lived post-dispatch poll) also re-run `reload()` so the agent-created issue surfaces automatically.

## 🔵 Non-blocking
- Dead i18n key: `loop.toast.attachFailed` (en-US.json:332) is now unused since NewLoopPage switched to `newLoop.attachFailedAbort` — safe to drop.
- `QuickCreateReq` could be an XOR union type (single call site is correct today).

OctoBoooot (commented):
## Review: #673 — byte-verifying Jerry-Xin's new-head findings (holding Approve)

**Verdict unchanged: Approve.** Of the three: the dead-key 🟡 is factually wrong (the key is still used elsewhere), the XOR 🟡 is a valid non-blocking nit, and the async-refresh 🔴 is a severity disagreement where I hold my calibrated **minor** — nothing in the bytes elevates it to a blocker.

### 🟡 "remove unused `loop.toast.attachFailed`" — refuted, the key is NOT dead
It's **still referenced** at `IssueDetailPage.tsx:408` (`Toast.error(t("loop.toast.attachFailed", …))`) — the *comment*-attachment path. NewLoopPage's dispatch path switched to the new `attachFailedAbort` key, but the original key remains in active use elsewhere. Removing it would break the comment composer's attachment-failure toast. Not removable → not a finding.

### 🟡 XOR union type for `QuickCreateReq` — valid nit, non-blocking
Accurate: `agent_id?` + `squad_id?` are both optional, so the *type* permits both/neither; the XOR is enforced at the call site (`...(assigneeType === "squad" ? {squad_id} : {agent_id})`), not by the type. A discriminated-union / XOR type would be a nice type-safety hardening, but the runtime already sends exactly one. Author's call, non-blocking.

### 🔴 async-refresh → I hold at minor (documented, self-correcting tradeoff)
Same finding as yujiawei #2 and my earlier #2, now graded 🔴 by Jerry-Xin. My assessment is unchanged on the evidence: quick-create's 202 means the agent builds the loop after navigation, so it doesn't auto-appear until a reload; the 15s poll only refreshes running-chips, not the list. **But** — the loop *is* created server-side (no data loss), the "已派单" toast confirms the dispatch, it appears on any reload/re-nav (self-correcting), and the PR explicitly scopes task_id correlation / completion-refresh out. Reachability ✓ / provenance ✓ (sync→async introduced the gap) but **escalation = pure self-correcting UX lag** → minor, not a merge-blocker. Severity isn't a vote; three reviewers grading it (2×🔴, me+Steve non-blocking) doesn't change that nothing in the bytes makes a self-correcting lag block-worthy. Worth a follow-up (pending-placeholder / refresh-on-return) — genuinely good polish — but not gating.

### Net
Verdict holds at Approve. The two actionable review 🟡s from the prior round (double-toast, attachment-fail-dispatch) were fixed in this commit and byte-verified. Of this round's three: one 🟡 is wrong (key still used), one is a fair non-blocking type nit, and the async-refresh is a legitimate severity disagreement where I hold minor with a stated gate-rationale. If a maintainer weighs the async UX gap as must-fix, that's a defensible product call I'd engage — but on correctness/escalation it's non-blocking. Credit: Jerry-Xin for the XOR-type nit and pressing the async UX.

yujiawei (approved):
# Code Review — PR #673 (octo-web)

Reviewed at head SHA `0747badc317e12693e9647d5b242ceb406c3950b` against base `feat/octo-loop` (merge-base `55d67be1`). Diff: 8 files, +65/−30, all under `packages/dmloop`. `tsc --noEmit` is clean for the changed files.

## Spec compliance

The diff does what the description says and nothing out of scope:

- **`quickCreateIssue` wiring** — `api/types.ts` adds `QuickCreateReq`, `api/issueApi.ts` adds `quickCreateIssue()` → `POST /issues/quick-create`. `createIssue` is untouched.
- **`toApiError` extension** — `LoopApiError` gains an optional `code`; the message chain appends `data.reason` as a strictly lower-priority fallback (`error || message || reason || …`). No other endpoint in the package returns a bare `{reason}` body (verified: the only `.reason` reads elsewhere are unrelated domain fields), so the ~120 other `httpPost`/`httpGet` call sites are unaffected.
- **Assignee now required** — the picker only offers `["agent","squad"]` and `AssigneePicker` hides "unassigned" when `member` isn't in `types`, so `onChange` always sets `assigneeId`+`assigneeType` together or clears both. The dispatch button disables on `!assigneeId`, and `submit` re-guards on `!assigneeId` (double safety). The `agent_id` XOR `squad_id` construction (`assigneeType === "squad" ? {squad_id} : {agent_id}`) is sound given that contract.
- **Double-toast removal** — `LoopPage` and `IssuePage` drop their `Toast.success("created")` from `onCreated`; `NewLoopPage` now owns a single accurate "Dispatched" toast. `IssuePage` also removed its now-unused `Toast` import (no dangling reference); `LoopPage` keeps `Toast` because it's still used for workspace creation. The third `NewLoopPage` consumer (`IssueDetailPage`) belongs to `CreateIssueModal`, not this flow, so it correctly keeps its own toast.
- **i18n** — `newLoop.agentUnavailable` / `daemonUnsupported` / `attachFailedAbort` present in both `zh-CN` and `en-US` under the `loop.newLoop` namespace matching the `t()` calls.

No missing, extra, or divergent behavior versus the stated scope. **Spec: pass.**

## Code quality

Verified the load-bearing claims by reading the code, not just the description: the XOR branch, the `toApiError` non-regression, the single-toast ownership, and that the removed `Toast` imports have zero remaining uses. All hold.

A separate reasoning pass surfaced three points, all non-blocking:

- **[P2] Attachment retry re-uploads and orphans blobs.** On any upload failure the code now aborts dispatch and returns, but it discards the already-uploaded ids and leaves `pendingFiles` unchanged. A retry re-uploads every file, orphaning the earlier server-side blobs. Aborting rather than dispatching with partial context is the right call; consider persisting successful ids in state (or trimming `pendingFiles`) so a retry only uploads the failures. Low impact — upload failure is uncommon and orphans are unreferenced.
- **[P2] Async dispatch vs. immediate list refetch.** `quick-create` returns `{ task_id }` and the agent builds the loop asynchronously, so the synchronous `onCreated()` → `onMutated()` refetch may run before the issue exists, and the new loop won't appear until a later refresh. This is the documented intent (task_id correlation is explicitly out of scope) and the "Dispatched" wording sets the right expectation rather than promising a created issue, so it's acceptable for this PR. Worth a follow-up for task_id/inbox correlation.
- **[P2] Sequential uploads.** Attachments upload one-by-one in a `for…of await` loop (pre-existing, not introduced here). `Promise.allSettled` would shorten the `submitting` window for multi-file dispatches.

No correctness, security, data-loss, or build-break issues. **Quality: approved.**

## Coverage / blind spots

- The backend `/issues/quick-create` contract (422 `{code, reason}` shape, the `agent_unavailable` / `daemon_version_unsupported` code strings, and the `{task_id}` response) is asserted by the frontend but lives in another repo and was not verified here — confirm the code strings match the server exactly, since a mismatch silently falls through to the generic error toast.
- Runtime behavior of `WKApp.routeRight` navigation after dispatch was reviewed by reading only; the PR notes real-browser e2e coverage for the enabled/disabled, 202, and 422 paths.

## Verdict

No P0/P1 findings. The three P2 items are improvements, not merge blockers.

Jerry-Xin (approved):
Re-reviewed at head `0747badc` after byte-verification and peer calibration. Flipping to **APPROVE** — correcting my prior CR.

## ✅ Fixed (byte-verified)
- **Double-toast (IssuePage `onCreated`)** — `IssuePage.tsx:217` `onCreated` now `{ onMutated(); routeRight.pop(); }`; the redundant `Toast.success` + unused `Toast` import removed. Single "已派单" toast owned by NewLoopPage across both callers.
- **Attachment-fail dispatch** — `NewLoopPage.tsx:66-70`: on any upload failure it `Toast.error(attachFailedAbort)` **+ `return`** *before* `quickCreateIssue`, aborting rather than dispatching on incomplete context. The early `return` is inside `try`, so `finally` still resets `submitting` (no wedge). Bilingual key present. This is the stronger (abort-over-confirm) fix — correct for an unrecoverable async flow.

## 🟡 Non-blocking (was my 🔴 — re-graded)
- **Async board-refresh:** quick-create is 202 async (agent creates the issue later), so `onCreated → reload()` fires before the issue exists and the 15s poll only refreshes running-chips, not the list. **However**, on re-examination this is a *self-correcting* UX lag, not a blocker: the loop is created server-side (no data loss), the "已派单" toast delivers the fail-fast dispatch confirmation that is this PR's actual goal, the loop surfaces on any subsequent reload/filter-change/revisit, and task_id correlation is an explicitly documented scoped-out enhancement. Recommended polish (non-blocking): insert an optimistic entry keyed on the returned `task_id`, or re-run `reload()` on a short post-dispatch poll / route-return. I was the lone 🔴 here; on the evidence it's a 🟡. Thanks to peers for the calibration.

## Retraction
- ~~🟡 remove unused `loop.toast.attachFailed`~~ — **withdrawn, I was wrong.** The key is still referenced at `packages/dmloop/src/panel/IssueDetailPage.tsx:408` (comment-attachment path, distinct from NewLoopPage). Not dead; must NOT be removed. (Credit Octo-PR for catching this.)
- `QuickCreateReq` XOR union remains a valid optional type-hardening nit (both ids optional; XOR enforced at call site). Author's call, non-blocking.

APPROVE.

PR #667, opened 2026-07-12
fix(export): correct Word math, nested-table columns, and cell images

wwwzzzcccc:
## Summary
Fix Word (DOCX) export defects: math constructs (accents, fences, braces, over/under labels, AMS environments), nested-table column collapse, and cell-image overflow.

## Linked Spec / Issue
Fixes #666

## What this change actually does (before/after)
- **Math OMML** (`export/docx/mathml-to-omml.ts`, `math.ts`):
  - Before: accents rendered as a right-superscript; `|`/`\|` stayed single-line height; `\overbrace`/`\underbrace` did not stretch; over/under labels misplaced; AMS environments (matrix/pmatrix/cases/align) failed as "Unknown environment" and fell back to raw LaTeX text.
  - After: accents map to `<m:acc>` with combining codepoints (centered over the base); `|`/`\|` pair into a growing `<m:d>`; braces map to `<m:groupChr>`; over/under/stackrel/xrightarrow map to `<m:limUpp>`/`<m:limLow>`; the AMS TeX package is registered via side-effect imports so those environments render as real math.
- **Nested tables** (`export/docx/tables.ts`): a nested table with all-explicit, lopsided column widths is scaled to the cell inner width and each column clamped to [0.6, 1.6]x the even share (clamp as the final step). Top-level tables keep literal explicit widths, so top-level layout is unchanged.
- **Cell images** (`export/docx/images.ts`, `nodes.ts`, `types.ts`): image width is capped to the containing cell's inner width (`maxImageWidthPx`), so nested-cell images shrink to fit instead of overflowing at the page-wide cap.

## What could break
- Dependents: any DOCX export containing math, tables, or images.
- Failure modes considered: top-level tables intentionally left on literal explicit widths (only nested tables reflow) to avoid changing existing layouts; the AMS side-effect imports only register TeX packages and do not alter non-AMS formulas; accent normalization is limited to a fixed accent set.

## How verified
- `packages/docs` docx export suite: 96 tests pass (includes new regression tests for accents/fences/braces in `math.test.ts` and nested-table clamp in `table-layout.test.ts`).
- `tsc --noEmit` clean for the touched files.
- Manual: exported the math and image-mixing test documents; OMML structure verified for each construct.

## Notes
- PDF export defects are addressed separately (Mininglamp-OSS/octo-docs-backend#57).
- Document import (Markdown/Word/PDF) is intentionally not part of this PR.

PR #674, opened 2026-07-13
fix(versions): keep version-history confirm box on-screen in shared overlay

ploy-elison:
## What

The restore/delete confirm box in the unified `VersionHistoryPanel` shell rendered at the end of the panel's content flow with only box-model styling and no positioning. On a long, scrolled version list it appeared below the fold, so users clicked delete/restore and never saw the confirm they had just triggered.

Now that the doc, sheet and board panels are all thin adapters over `VersionHistoryPanel`, the shell is the single place that renders the destructive-action confirm for every end — so fixing it here fixes all three.

## Changes

- Render the restore and delete confirm boxes inside the shared `.octo-modal-overlay` (fixed, viewport-centered) — the same overlay the preview/diff modal uses — so the box is always visible and its buttons clickable regardless of scroll position.
- Clicking the backdrop or pressing `Escape` cancels, mirroring the preview modal. Both paths are guarded so an in-flight restore/delete can't be dismissed mid-request. The confirm state machine, race guard, and mounted-ref liveness are untouched.
- Scope the modal-card chrome under `.octo-modal-overlay .octo-version-confirm` so it only styles the overlay-anchored card and never leaks onto any path that might still render the class inline in the content flow.
- Share the elevation via a new `--octo-modal-shadow` token; the confirm and the existing preview/diff modals now reuse it instead of hardcoded `rgba` shadow literals.
- Add regression tests asserting the shell confirm is overlay-wrapped, cancels on backdrop-click / `Escape`, and holds while a restore is in flight.

## Notes

Rebased onto the latest `main`. Since the three version panels were unified behind `VersionHistoryPanel`, the live `VersionPanel` no longer renders its own inline confirm — the earlier inline-confirm migration is dropped, and the final diff is limited to the shell, its tests, and the shared card styles.

## Testing

- `packages/docs` version tests green (`vitest run src/versions/` — 77 passing).
- `typecheck`: no new errors introduced by this change.
- `stylelint`: no new violations; net fewer raw `rgba` literals after consolidating the shadow into a token.

### DIFF
--- a/packages/docs/src/editor/styles.css
+++ b/packages/docs/src/editor/styles.css
@@ -6,6 +6,8 @@
   --octo-accent: #3370ff;
   --octo-border: #e5e6eb;
   --octo-hover: #f2f3f5;
+  /* Shared elevation for overlay-anchored cards (preview/diff modal + destructive confirm). */
+  --octo-modal-shadow: 0 12px 48px rgba(0, 0, 0, 0.24);
 }
 
 @media (prefers-color-scheme: dark) {
@@ -1059,7 +1061,7 @@
   color: var(--octo-fg, #1f2329);
   border: 1px solid var(--octo-border, #e5e6eb);
   border-radius: 12px;
-  box-shadow: 0 12px 48px rgba(0, 0, 0, 0.24);
+  box-shadow: var(--octo-modal-shadow);
   padding: 20px 24px;
   box-sizing: border-box;
 }
@@ -1387,7 +1389,7 @@
   color: var(--octo-fg, #1f2329);
   border: 1px solid var(--octo-border, #e5e6eb);
   border-radius: 12px;
-  box-shadow: 0 12px 48px rgba(0, 0, 0, 0.24);
+  box-shadow: var(--octo-modal-shadow);
   padding: 20px 24px;
   box-sizing: border-box;
 }
@@ -1724,12 +1726,28 @@
   justify-content: flex-end;
   gap: 6px;
 }
-.octo-version-confirm {
-  margin-top: 12px;
-  padding: 12px;
+/* Destructive-action confirm box (restore / delete). The VersionHistoryPanel shell — the single
+   confirm renderer the doc / sheet / board ends now share — renders it inside the shared
+   .octo-modal-overlay (fixed, viewport-centered), the same overlay the preview/diff modal uses
+   (decision #2), so it is always visible and clickable regardless of how far the version list has
+   scrolled. Before this, it sat at the end of the panel's content flow with only box-model styling
+   and no positioning, so on long scrolled lists it fell below the fold and users could not see the
+   confirm they had just triggered. The modal-card chrome (white bg + shadow) is scoped under
+   .octo-modal-overlay so it only applies to the overlay-anchored card and never leaks onto any path
+   that might still render this class inline in the content flow. */
+.octo-modal-overlay .octo-version-confirm {
+  width: 100%;
+  max-width: 420px;
+  padding: 16px 20px;
   border: 1px solid var(--octo-border, #e5e6eb);
-  border-radius: 8px;
-  background: var(--octo-accent-weak, #f2f3f5);
+  border-radius: 12px;
+  background: var(--octo-bg, #fff);
+  color: var(--octo-fg, #1f2329);
+  box-shadow: var(--octo-modal-shadow);
+  box-sizing: border-box;
+}
+.octo-version-confirm > p:first-child {
+  margin-top: 0;
 }
 .octo-version-confirm-detail {
   color: var(--octo-muted, #86909c);
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -306,6 +306,66 @@ describe('VersionHistoryPanel — mutations & permissions', () => {
     confirmSpy.mockRestore()
   })
 
+  it('renders the confirm box in a fixed viewport-centered overlay (stays visible on long scrolled lists)', async () => {
+    // Regression for XIN-867: the in-panel confirm box used to render at the end of the panel's
+    // content flow (after the version list) with only box-model styling and no positioning, so on a
+    // long, scrolled list it fell below the fold — users clicked delete/restore and never saw the
+    // confirm. It now renders inside the shared .octo-modal-overlay (position: fixed; inset: 0), the
+    // same viewport-anchored overlay the preview/diff modal uses, so it is always on-screen and its
+    // buttons clickable regardless of scroll. jsdom does not compute layout, so we assert the overlay
+    // structure (which pins the fixed positioning) rather than measured coordinates, and verify the
+    // overlay-click / Esc cancel paths that match the preview modal.
+    renderPanel('admin')
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.delete'))
+    const box = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    // The confirm card is wrapped by the fixed viewport overlay, not left inline in the panel flow.
+    const overlay = box.closest('.octo-modal-overlay') as HTMLElement
+    expect(overlay).toBeTruthy()
+    expect(box.getAttribute('role')).toBe('dialog')
+    expect(box.getAttribute('aria-modal')).toBe('true')
+    // Clicking the card body does not dismiss (stopPropagation); clicking the overlay backdrop does.
+    fireEvent.mouseDown(box)
+    expect(document.querySelector('.octo-version-confirm')).toBeTruthy()
+    fireEvent.mouseDown(overlay)
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
+    // Reopen and confirm Escape cancels too (mirrors the preview modal).
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.restore'))
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeTruthy())
+    fireEvent.keyDown(document, { key: 'Escape' })
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
+  })
+
+  it('does not dismiss the confirm overlay on backdrop-click or Escape while a restore is in flight (busy guard)', async () => {
+    // The overlay-click and Escape cancel paths both no-op while a mutation is running, so a stray
+    // backdrop click or keypress cannot tear the confirm down mid-request. Hold the restore mutation
+    // open to keep the panel busy, then assert both dismissals are ignored until it settles.
+    let release!: (v: { newDocVersionSeq: number; restoredFrom: number }) => void
+    restoreVersionMock.mockImplementationOnce(
+      () => new Promise((resolve) => {
+        release = resolve
+      }),
+    )
+    renderPanel('admin')
+    await screen.findByText('Draft v1')
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.restore'))
+    const box = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    const overlay = box.closest('.octo-modal-overlay') as HTMLElement
+
+    // Trigger the restore → busy is now true (mutation pending).
+    fireEvent.click(btnByText(box, 'docs.version.restore'))
+    await waitFor(() => expect(restoreVersionMock).toHaveBeenCalled())
+
+    // Backdrop click and Escape are both no-ops while busy.
+    fireEvent.mouseDown(overlay)
+    fireEvent.keyDown(document, { key: 'Escape' })
+    expect(document.querySelector('.octo-version-confirm')).toBeTruthy()
+
+    // Once the mutation settles, the confirm clears on its own.
+    release({ newDocVersionSeq: 9, restoredFrom: 7 })
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
+  })
+
   it('renames a named version inline (no native prompt)', async () => {
     renderPanel('admin')
     await screen.findByText('Draft v1')
--- a/packages/docs/src/versions/VersionHistoryPanel.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.tsx
@@ -391,6 +391,20 @@ export function VersionHistoryPanel<TState, TCurrent>({
     return () => document.removeEventListener('keydown', onKey)
   }, [selected, closePreview])
 
+  // Escape also cancels the restore / delete confirm overlay, matching the preview modal's Esc/
+  // overlay-close behavior. It only clears the confirm state (never the in-flight mutation), and it
+  // no-ops while a mutation is running so a keypress can't dismiss the box mid-request.
+  useEffect(() => {
+    if (!confirmRestore && !confirmDelete) return
+    const onKey = (e: KeyboardEvent) => {
+      if (e.key !== 'Escape' || busy) return
+      setConfirmRestore(null)
+      setConfirmDelete(null)
+    }
+    document.addEventListener('keydown', onKey)
+    return () => document.removeEventListener('keydown', onKey)
+  }, [confirmRestore, confirmDelete, busy])
+
   const filterBtn = (k: KindFilter, label: string) => (
     <button
       type="button"
@@ -569,50 +583,78 @@ export function VersionHistoryPanel<TState, TCurrent>({
         )}
 
         {confirmRestore && (
-          <div className="octo-version-confirm">
-            <p>{t('docs.version.confirmTitle', { values: { seq: confirmRestore.docVersionSeq } })}</p>
-            <p className="octo-version-confirm-detail">{t('docs.version.confirmDetail')}</p>
-            <div className="octo-member-row">
-              <button
-                type="button"
-                className="octo-tb-btn"
-                disabled={busy}
-                onClick={() => void onConfirmRestore(confirmRestore)}
-              >
-                {t('docs.version.restore')}
-              </button>
-              <button
-                type="button"
-                className="octo-tb-btn"
-                disabled={busy}
-                onClick={() => setConfirmRestore(null)}
-              >
-                {t('docs.version.cancel')}
-              </button>
+          <div
+            className="octo-modal-overlay"
+            role="presentation"
+            onMouseDown={() => {
+              if (!busy) setConfirmRestore(null)
+            }}
+          >
+            <div
+              className="octo-version-confirm"
+              role="dialog"
+              aria-modal="true"
+              aria-label={t('docs.version.restore')}
+              onMouseDown={(e) => e.stopPropagation()}
+            >
+              <p>{t('docs.version.confirmTitle', { values: { seq: confirmRestore.docVersionSeq } })}</p>
+              <p className="octo-version-confirm-detail">{t('docs.version.confirmDetail')}</p>
+              <div className="octo-member-row">
+                <button
+                  type="button"
+                  className="octo-tb-btn"
+                  disabled={busy}
+                  onClick={() => void onConfirmRestore(confirmRestore)}
+                >
+                  {t('docs.version.restore')}
+                </button>
+                <button
+                  type="button"
+                  className="octo-tb-btn"
+                  disabled={busy}
+                  onClick={() => setConfirmRestore(null)}
+                >
+                  {t('docs.version.cancel')}
+                </button>
+              </div>
             </div>
           </div>
         )}
 
         {confirmDelete && (
-          <div className="octo-version-confirm">
-            <p>{t('docs.version.deleteConfirm', { values: { seq: confirmDelete.docVersionSeq } })}</p>
-            <div className="octo-member-row">
-              <button
-                type="button"
-                className="octo-tb-btn"
-                disabled={busy}
-                onClick={() => void onDelete(confirmDelete)}
-              >
-                {t('docs.version.delete')}
-              </button>
-              <button
-                type="button"
-                className="octo-tb-btn"
-                disabled={busy}
-                onClick={() => setConfirmDelete(null)}
-              >
-                {t('docs.version.cancel')}
-              </button>
+          <div
+            className="octo-modal-overlay"
+            role="presentation"
+            onMouseDown={() => {
+              if (!busy) setConfirmDelete(null)
+            }}
+          >
+            <div
+              className="octo-version-confirm"
+              role="dialog"
+              aria-modal="true"
+              aria-label={t('docs.version.delete')}
+              onMouseDown={(e) => e.stopPropagation()}
+            >
+              <p>{t('docs.version.deleteConfirm', { values: { seq: confirmDelete.docVersionSeq } })}</p>
+              <div className="octo-member-row">
+                <button
+                  type="button"
+                  className="octo-tb-btn"
+                  disabled={busy}
+                  onClick={() => void onDelete(confirmDelete)}
+                >
+                  {t('docs.version.delete')}
+                </button>
+                <button
+                  type="button"
+                  className="octo-tb-btn"
+                  disabled={busy}
+                  onClick={() => setConfirmDelete(null)}
+                >
+                  {t('docs.version.cancel')}
+                </button>
+              </div>
             </div>
           </div>
         )}


OctoBoooot (approved):
## Review: keep version-history confirm box on-screen for long lists (#674)

**Verdict:** Approve
Correct presentation fix for a real UX bug — the restore/delete confirm box rendered at the end of the panel flow, so on a scrolled list it fell below the fold and the user never saw the confirm they triggered. Now rendered in the shared fixed `.octo-modal-overlay`, always on-screen. Proper modal semantics, presentation-only. No defects.

Risk tier: normal. Base `main`; CI green (Build/osv/code-review pending, not failing). +129/-43, 3 files. Affects all three ends (doc/sheet/board) via the shared shell.

### What I verified
- **Backdrop-cancel doesn't fire on inner clicks** (the #668 `clickToHide` lesson, applied right): the overlay `onMouseDown` cancels, but the inner `.octo-version-confirm` card has `onMouseDown={(e) => e.stopPropagation()}`, so clicks on the buttons/text don't bubble to the backdrop. Using `onMouseDown` (not `onClick`) consistently on both avoids the mousedown-in/mouseup-out drag-cancel glitch.
- **Dismiss is busy-guarded on both paths**: backdrop `if (!busy) setConfirmX(null)` and Escape `if (e.key !== 'Escape' || busy) return` — neither can cancel a confirm while its restore/delete mutation is in flight (matches the buttons' `disabled={busy}`). Escape only clears confirm state, never the in-flight mutation.
- **Presentation-only claim holds**: the diff moves the confirm JSX into the overlay + adds the Escape/backdrop cancel + CSS; the confirm state machine, unified race guard, and mounted-ref liveness are untouched.
- **CSS is token-based** (`var(--octo-bg,#fff)` / `var(--octo-fg,#1f2329)` + `box-sizing: border-box`), the confirm card correctly becomes the centered modal on the dim backdrop; well-documented rationale comment. §2-clean.
- **Tests** pin the fix: overlay wrapping (`role="dialog"`/`aria-modal`), body-click-no-dismiss (stopPropagation), backdrop/Escape-cancel — three real behavioral assertions on exactly the changed behavior.

### Note (non-blocking) — two Escape handlers can co-fire
The new confirm-Escape effect (`:385`, gated on `confirmRestore||confirmDelete`) and the existing preview-modal Escape effect (`:385-392`, gated on `selected`) both register `document` keydown listeners. If `selected` is non-null (a preview was opened) while a confirm box is also open, a single Escape fires **both** — `closePreview()` and `setConfirmX(null)`. Harmless in practice (both only dismiss/cancel — no mutation, no data loss, and the confirm is the top layer so cancelling it is the intended outcome; closing the already-behind preview underneath is benign). Non-blocking, but if you want strict layered-Escape semantics, the confirm handler could `e.stopPropagation()` / the effects could share a single top-most-layer keydown dispatcher. Not required.

### Praise
- The fix reuses the *exact* overlay the preview/diff modal already uses rather than inventing new positioning, so the confirm inherits proven centering + backdrop behavior; the body `stopPropagation` + `onMouseDown` pairing and the busy-guarded dismiss on both backdrop and Escape are the right modal-interaction details (the same class of thing that bit #668's filter popover, done correctly here). Kept strictly to presentation — the confirm/race/mounted-ref machinery untouched — and pinned the on-screen-overlay behavior with a focused regression test.

Jerry-Xin (dismissed):
Reviewed at head `ff7676131b37`, byte-verified against the mirror. The overlay approach for the *shell* is clean, but the CSS change is unscoped and regresses the **actually-shipped** panel, so this is REQUEST_CHANGES.

## 🔴 Blocking — global `.octo-version-confirm` restyle regresses the live panel
There are **two** components that render `.octo-version-confirm`:
- `packages/docs/src/versions/VersionHistoryPanel.tsx:594,633` — the unified shell this PR wraps in a `.octo-modal-overlay` (fixed, inset:0, z-index:1200, scrollable) + `role="dialog"`/`aria-modal` + Escape/backdrop cancel. Clean, mirrors the existing preview modal.
- `packages/docs/src/versions/VersionPanel.tsx:543` — a **separate, live** panel that renders `<div className="octo-version-confirm">` **inline, with no overlay wrapper**.

The CSS at `packages/docs/src/editor/styles.css:1694` is a **bare, unscoped** selector (`.octo-version-confirm { ... }`), changed from an inline block (margin-top:12px, accent-weak bg) into a modal card (`max-width:420px; white bg; radius:12px; box-shadow:0 12px 48px rgba(0,0,0,.24)`) with `> p:first-child { margin-top:0 }`.

Mounting reality (byte-verified — importer grep at head):
- `EditorShell.tsx:751` mounts `<VersionPanel .../>` — **this is the shipped doc-editor path.**
- `VersionHistoryPanel` (the shell this PR restyles) has **zero non-test importers**. The board/sheet paths mount their own `BoardVersionPanel`/`SheetVersionPanel`.

**Net production effect:** the global restyle turns `VersionPanel`'s inline confirm (`:543`, still no overlay) into a floating 420px white shadowed card sitting in normal document flow with its top margin removed — a visual regression on the panel users actually see — while the advertised "keep confirm on-screen for long lists" fix only lands on `VersionHistoryPanel`, which isn't mounted in the doc editor.

**Fix (either):**
1. Scope the modal styling to the overlay: `.octo-modal-overlay .octo-version-confirm { ... }`, preserving the inline styling for the live `VersionPanel`; **or**
2. Migrate `VersionPanel`'s confirm to the same `.octo-modal-overlay` pattern so the fix actually reaches production.

## 🟡 Non-blocking
- `styles.css:1702` new `box-shadow: 0 12px 48px rgba(0,0,0,0.24)` — `stylelint.ci.config.mjs` disallows `rgba()`. The file has pre-existing `rgba(0,0,0,…)` lines so enforcement may be diff-scoped; this new line is a plausible CI trip. Prefer an existing shadow token/var. (`code-review` check was PENDING at review time.)

## ✅ Good
Shell public interface (props at lines 42/47/107) untouched — no breaking API change to the 3 adapters. Test additions (`VersionHistoryPanel.test.tsx` +30/-0) are solid: assert overlay wrapping, `role="dialog"`, `aria-modal`, body-click no-dismiss, backdrop dismiss, Escape cancel (busy-guarded, doesn't clear in-flight mutation). No new i18n keys (all 6 `t()` keys pre-exist in both locales). The only gap: no test guards the live `VersionPanel` against this global CSS change.

REQUEST_CHANGES.

OctoBoooot (changes requested):
## Review: #674 — correcting my verdict (Approve → Request changes)

Jerry-Xin caught a real regression I missed: the CSS change restyles the **shared, global** `.octo-version-confirm` class into a centered modal card, but the **live doc `VersionPanel`** (mounted by EditorShell) renders that class **inline**, not in an overlay — so its restore/delete confirm now renders as a misplaced floating card in the content flow. Byte-verified. Flipping.

Risk tier: normal. Base `main`.

### Major — global `.octo-version-confirm` restyle breaks the live (un-migrated) VersionPanel
- The CSS selector is **global**: `.octo-version-confirm { max-width: 420px; box-shadow: 0 12px 48px …; background: #fff; … }` (`styles.css:1694`) — not scoped to `.octo-modal-overlay .octo-version-confirm`. It repurposes the class from an inline accent block into a modal card that *expects* the overlay backdrop.
- This PR moves the confirm-into-overlay **only in `VersionHistoryPanel`** (the not-yet-mounted shell). It does **not** touch `VersionPanel.tsx` (`VersionPanel.tsx in #674 diff: False`).
- But `VersionPanel.tsx` — the **doc version panel actually mounted by `EditorShell`** (live in production) — renders `.octo-version-confirm` **inline** (`:543`, `{confirmRestore && <div className="octo-version-confirm">…`), separate from its preview overlay (`:569`). So the global restyle drops a floating 420px white shadowed card into VersionPanel's inline content flow with no backdrop → a visible broken confirm UI on a live path.
- **Reachable + PR-introduced**: any doc-editor user who clicks restore/delete in version history sees the regressed box; the CSS change is what introduces it.

**Severity: major** — a visible UI regression on a production panel (not data-loss, so not blocker-tier, but it breaks a live user-facing flow and shouldn't merge as-is). Jerry-Xin's two fix directions are both correct: (a) **scope the modal styling** — `.octo-modal-overlay .octo-version-confirm { … }` so only the overlay-hosted (shell) box gets card chrome, leaving VersionPanel's inline box on its old inline style; or (b) **migrate VersionPanel** to the same overlay pattern in this PR so both use the modal form.

### Also (Jerry-Xin 🟡, fair)
- The new `box-shadow: 0 12px 48px rgba(0,0,0,0.24)` is a raw rgba literal — may trip the hardcoded-color CI rule; prefer an existing shadow token/var if one exists.

### What's still correct (the VersionHistoryPanel side)
The overlay move within the shell itself is well-built (backdrop `stopPropagation`, busy-guarded dismiss, presentation-only, tests) — that part I verified and it stands. The defect is purely that the **shared global CSS** change wasn't scoped, so it leaks onto the live un-migrated `VersionPanel`.

### Accountability
I approved this having verified the fix was correct *for VersionHistoryPanel* — but I treated `.octo-version-confirm` as shell-private and **did not grep the shared class's other consumers**. The PR body's "shell not mounted yet" lulled me: true for the shell, but the *CSS is global* and the live doc `VersionPanel` uses the same class inline. Lesson (recurring shared-surface shape): a change to a **shared CSS class** — like a shared function or i18n key — must be checked against **all** its consumers, not just the file the PR edits. Same blast-radius discipline as the #660 shared-helper / #671 shared-shell-i18n cases, applied to CSS. Credit: Jerry-Xin — real, reachable, production-panel regression.

yujiawei (dismissed):
# Code Review — PR #674 (octo-web)

Reviewed against head SHA `ff7676131b37d7b19e7e31055836aced7bc2d60b`, merge-base `d80703a5`.

## 1. Spec compliance

**Spec: ✅**

- The restore/delete confirm box is moved into the shared `.octo-modal-overlay` (`position: fixed; inset: 0`), the same overlay the preview/diff modal already uses, so it can no longer fall below the fold on a long scrolled list (`packages/docs/src/versions/VersionHistoryPanel.tsx:585-660`, `packages/docs/src/editor/styles.css:1687-1706`).
- Backdrop click and `Escape` cancel the confirm, matching the preview modal (`VersionHistoryPanel.tsx:397-406`, `589-591`, `628-630`).
- A regression test pins the overlay structure, `stopPropagation`, backdrop-close, and `Escape` cancel (`VersionHistoryPanel.test.tsx:309-337`).

Verified against the "presentation-only" claim: the change is a JSX wrap + one `useEffect` + CSS. The confirm state machine, the race guard, and the mounted-ref liveness checks are untouched. No missing scope, no extra scope, no divergence.

## 2. Code quality

**Quality: Approved**

Verification performed locally at the head SHA:
- `pnpm exec vitest run src/versions/VersionHistoryPanel.test.tsx` → **16/16 pass**.
- `pnpm run typecheck` reports 2 errors (`dmworkbase/src/i18n/format.ts:38`, `members/sort.ts:26`), both **preexisting on the merge-base** and in files this PR does not touch. This PR introduces **no new type errors** — the description's claim holds.

The `busy` guarding is correct and actually more careful than the sibling preview modal: `Escape` (`VersionHistoryPanel.tsx:400`) and backdrop `mousedown` (`589`, `628`) both no-op while a mutation is in flight, consistent with the `disabled={busy}` buttons, so a stray keypress/click can't dismiss the confirm mid-request.

No P0/P1 findings. Advisory items below (P2 — none block merge):

- **[P2] The fixed overlay is rendered inline rather than through a Portal.** If any future ancestor of the panel introduces `transform` / `filter` / `perspective` / `contain` / `will-change`, it would establish a containing block and trap the `position: fixed` overlay inside that ancestor instead of the viewport, re-clipping the box this PR is fixing. I grep-checked `packages/docs/src/editor/styles.css` and found **no** such property on any current ancestor of `.octo-version-panel` (`.octo-doc-drawer` uses `isolation: isolate`, which creates a stacking context but does **not** establish a containing block for `fixed`), so the fix works on the known path today. This is worth hardening with `createPortal(..., document.body)` later, but it is not a blocker: the sibling preview/diff modal in the same file uses the identical inline `.octo-modal-overlay` pattern (`VersionHistoryPanel.tsx:664`), so the PR follows the file's established, working convention rather than regressing it.
- **[P2] No focus management on the confirm dialog.** It sets `role="dialog"` + `aria-modal="true"` but does not move focus into the card on open or trap Tab, so a keyboard/screen-reader user can still reach the obscured background list. This is a real a11y gap, but it is identical to the existing preview modal in the same file (`VersionHistoryPanel.tsx:665-670`) — file-wide debt, not something this PR introduces. For a destructive-action confirm, focusing Cancel on mount would be a nice follow-up.
- **[P2] Backdrop dismissal is on `mousedown`, and the inner card calls `stopPropagation()` unconditionally.** `mousedown`-to-close means a user can't abort an accidental press by dragging off the backdrop before releasing (an `onClick` with an `e.target === e.currentTarget` check reads more standard). Again this mirrors the preview modal's existing pattern, so it's a consistency choice rather than a new defect.
- **[P2 / test nit] The regression test doesn't cover the `busy` guard** — i.e. that `Escape` and a backdrop click are no-ops while a mutation is in flight. That guard is the one behavior most likely to regress silently; a small assertion around it would be worth adding.

## 3. Coverage / blind spots

- **Not route-mounted.** The unified shell has no end adapter wired yet, so this cannot be verified in a live browser and the containing-block risk above cannot be confirmed or refuted against real runtime ancestors. The upside is symmetric: the change also cannot regress any live doc/sheet/board surface today. Recommend re-checking the overlay in a real viewport when an end adapter lands.
- **Multi-modal `Escape`.** If the confirm and the preview modal were ever open at once, `Escape` would close both (neither listener stops propagation). Through the current UI they are mutually exclusive (the confirm overlay covers the row buttons that open preview, and both mutation handlers call `closePreview()`), so this is theoretical — noting it for whoever wires the end adapter.

## Verdict

**APPROVED.** Correct, targeted, presentation-only fix that follows the file's established modal pattern; tests pass and no new type errors. The P2 items (portal hardening, focus trap, mousedown-vs-click, busy-guard test) are non-blocking follow-ups, several of which are pre-existing file-wide conventions rather than regressions introduced here.

OctoBoooot (dismissed):
## Review: #674 — delta re-review (global-CSS regression fixed, both ways)

The new commit fixes the blocker thoroughly — it does **both** remediations: migrates the live doc `VersionPanel`'s confirm into the shared overlay AND scopes the modal-card CSS under `.octo-modal-overlay`, plus tokenizes the rgba shadow (the 🟡). Byte-verified. Flipping.

Risk tier: normal. Base `main`. `ahead=1 behind=0` (clean single commit).

### Resolved — Blocker: global `.octo-version-confirm` no longer regresses the live VersionPanel
Two independent fixes, either of which would suffice; together they're defense-in-depth:
1. **CSS scoped** (`styles.css:1697`): the modal-card rule (max-width/bg/shadow/border-radius) is now `.octo-modal-overlay .octo-version-confirm { … }` — it only styles the overlay-anchored card, so it can never leak onto a path that renders the class inline. Confirmed the only remaining global `.octo-version-confirm*` rules are the benign `> p:first-child { margin-top:0 }` reset and the `-detail` text styling — neither carries positioning/card chrome.
2. **Live VersionPanel migrated** (`VersionPanel.tsx`): its restore confirm is now wrapped in `.octo-modal-overlay` (backdrop `onMouseDown` cancel + inner card `stopPropagation`, `role="dialog"`/`aria-modal`), matching the shell — so the doc editor's confirm is overlay-hosted, not inline. Plus a busy-guarded Escape-cancel handler, consistent with VersionHistoryPanel.

So the doc VersionPanel now renders a proper centered modal (the intended fix reaches production), and the scoping guarantees no inline path can inherit the misplaced-card styling.

### Resolved — rgba 🟡
`box-shadow: 0 12px 48px rgba(0,0,0,0.24)` → extracted `--octo-modal-shadow` token, applied at all three sites (preview modal, diff modal, confirm). Removes the raw rgba (CI-rule concern) and unifies the elevation.

### Tests
+94 lines in `VersionPanel.test.tsx` — pins the overlay-wrapped confirm (dialog/aria-modal, backdrop stopPropagation, Escape/backdrop cancel) on the live panel, mirroring the shell's regression test.

### Accountability
I approved the original having verified the fix was correct *for the shell* but treated `.octo-version-confirm` as shell-private and didn't grep the shared class's other consumers — missing that the live `VersionPanel` uses it inline. yujiawei caught it; the author fixed it the right way (both scope the CSS and migrate the live panel). The recurring lesson this batch (shared helper #660 → shared shell-i18n #669/#671 → shared CSS class #674) is now firmly banked: **a change to any shared surface — function, i18n key, or CSS class — must be checked against every consumer, not just the file the PR edits.** Credit: yujiawei. Blocker fully resolved, byte-verified.

Jerry-Xin (dismissed):
Re-reviewed at head `b10d3b01`, byte-verified. My prior 🔴 (global `.octo-version-confirm` restyle regressing the live inline confirm) is fully fixed — flipping to **APPROVE**.

## ✅ 🔴 resolved — CSS scoped + live panel migrated
1. **CSS card-chrome now scoped** (`styles.css:1697`): `.octo-modal-overlay .octo-version-confirm { max-width:420px; white bg; box-shadow:var(--octo-modal-shadow); border; ... }`. The only remaining **bare** `.octo-version-confirm` rules are harmless typography (`:1708` `> p:first-child` margin-reset, `:1711` `-detail` color). No bare card-chrome rule remains → an inline `.octo-version-confirm` gets zero modal chrome. The leak vector is closed.
2. **Live `VersionPanel` migrated into the overlay** (`VersionPanel.tsx:559-566`): the previously-inline confirm is now wrapped in `<div className="octo-modal-overlay" role="presentation" onMouseDown={() => { if (!busy) setConfirmRestore(null) }}>` with the inner card `role="dialog"` / `aria-modal` / `onMouseDown stopPropagation`. New Escape handler (`:332-344`) is **busy-guarded** (`if (e.key !== 'Escape' || busy) return`) — can't dismiss mid-restore. The restore mutation flow + buttons (`disabled={busy}`, `onConfirmRestore`) are **byte-unchanged**; no state-machine regression.
3. **Every non-test renderer is now overlay-wrapped**: `VersionPanel.tsx:565` (migrated) + shell `VersionHistoryPanel.tsx:594/633` (wrapped in the prior commit). No inline consumer remains, so the scoping under-styles no one.

## ✅ 🟡 rgba → token
`--octo-modal-shadow: 0 12px 48px rgba(0,0,0,0.24)` defined at `:root` (`styles.css:10`), consumed via `var(--octo-modal-shadow)` at preview (1024), diff (1352), confirm (1705). Net −2 rgba literals. Resolved.

## ✅ Regression tests are real
`VersionPanel.test.tsx` asserts the confirm is overlay-wrapped not inline (`box.closest('.octo-modal-overlay')` truthy, `.octo-version-panel` null, `role=dialog`/`aria-modal`), backdrop-click dismisses while card-body mousedown doesn't, Escape dismisses, and — with the restore promise held open — backdrop-click + Escape are **no-ops while busy** (guard genuinely exercised), then clear on resolve.

## Non-review gate (for the author, not a code blocker)
PR is currently **CONFLICTING** (`mergeable:false`) — a real content conflict in `VersionPanel.tsx` since main advanced. Needs a **rebase onto main** before it can land, and `code-review` check is pending. Approving on code correctness; please rebase.

APPROVE.

yujiawei (dismissed):
# Code Review — PR #674 (octo-web)

Reviewed against head SHA `b10d3b014e114f0f44a856a8986af38cfa616063`, merge-base `d80703a5`. This is a re-review: the previous head `ff767613` was blocked because the global `.octo-version-confirm` restyle was unscoped and regressed the live, actually-shipped panel while the fix only reached the unmounted shell. The new commit `b10d3b01` addresses that blocker directly.

## 1. Spec compliance

**Spec: ✅**

The diff does exactly what the description states — a presentation-only fix — and nothing more:

- **Missing: none.** The fix now reaches the panel users actually see. `EditorShell.tsx:751` mounts the live `VersionPanel`, and `VersionPanel.tsx:556-594` now wraps its restore confirm in the shared `.octo-modal-overlay` (`position: fixed; inset: 0`), the same overlay the preview/diff modal uses. `VersionHistoryPanel.tsx:585-660` gets the same treatment for both restore and delete confirms.
- **Over-build: none.** The added `Escape`-cancel `useEffect` handlers (`VersionPanel.tsx:333-342`, `VersionHistoryPanel.tsx:395-404`) are in-scope and side-effect-free: they only clear the confirm state, no-op while `busy` (a keypress can't dismiss the box mid-request), and never touch the in-flight mutation. This mirrors the preview modal's existing Esc/backdrop behavior.
- **Divergence: none.** The confirm state machine and the `onConfirmRestore`/`onDelete` handlers are untouched — the change is a JSX wrap + one `useEffect` + CSS scoping.

The CSS blocker is resolved two ways (belt and suspenders): the modal-card chrome is now scoped under `.octo-modal-overlay .octo-version-confirm` (`styles.css:1697`) so it cannot leak onto any inline render, **and** the live panel was migrated into the overlay. Verified every non-test render of `.octo-version-confirm` is now inside an overlay wrapper (`VersionPanel.tsx:565`, `VersionHistoryPanel.tsx:594,633`); `SheetVersionPanel`/`BoardVersionPanel` use native `window.confirm` and are unaffected. The prior non-blocking stylelint concern is also moot: the hardcoded shadow was consolidated into a single `--octo-modal-shadow` token, and total `rgba()` usage in `styles.css` went 39 → 38.

## 2. Code quality

**Quality: Approved**

Verified locally at head `b10d3b01`:
- `pnpm exec vitest run src/versions/VersionPanel.test.tsx src/versions/VersionHistoryPanel.test.tsx` → **31/31 pass**.
- `pnpm run typecheck` (docs pkg) reports 2 errors (`dmworkbase/src/i18n/format.ts:38`, `members/sort.ts:26`), both **preexisting on the merge-base** and in files this PR does not touch. **No new type errors.**

The new tests are meaningful, not false-green. They pin the real behavioral change — inline → overlay migration (`box.closest('.octo-modal-overlay')` truthy, `box.closest('.octo-version-panel')` null), `role="dialog"`/`aria-modal`, `stopPropagation` (card-body click no-dismiss), backdrop dismiss, and Esc cancel. The live-panel test also adds a genuine busy-guard case that holds the restore mutation open via a deferred promise and asserts backdrop/Esc are no-ops until it settles. No assertion would pass if the fix were reverted.

No P0/P1 findings. Advisory only (P2 — none block merge):

- **[P2] JSX indentation drift.** After `</section>` was hoisted up (`VersionPanel.tsx:554`), the `confirmRestore` and `selected` overlay siblings keep their old 6-space indentation instead of matching `<section>`'s 4 spaces. Purely cosmetic — the tree is well-formed, compiles, and Prettier/lint would normalize it.
- **[P2] "viewport-centered" wording.** Comments/test names describe the confirm as "viewport-centered," but the shared `.octo-modal-overlay` uses `align-items: flex-start` + `48px` top padding (`styles.css:1011`), so it is horizontally centered and top-anchored, not vertically centered. Wording inaccuracy only; the actual goal (on-screen and reachable regardless of scroll) is met. This overlay rule is pre-existing, not introduced here.
- **[P2] CSS-scope coverage boundary.** jsdom does not compute CSS, so the `.octo-modal-overlay .octo-version-confirm` scoping (the inline-leak prevention) is asserted only via DOM structure, not computed style. Inherent limitation, acceptable for a presentation-only fix; a future re-broadening of the selector wouldn't be caught by the unit suite.
- **[P2] Test symmetry.** The `VersionHistoryPanel` overlay test omits the busy-guard leg that the live `VersionPanel` test covers. The shell carries the same guard in code and is the unmounted panel, so the higher-value guard rightly lives in the live-panel test — a symmetry nit, not a gap that ships broken.

## 3. Coverage / blind spots

- **Portal / containing block.** The overlay is rendered inline (not via `createPortal`). A full scan of `styles.css` found no `transform`/`filter`/`perspective`/`contain`/`will-change` on any current ancestor of `.octo-version-panel` that would trap the `position: fixed` overlay in a containing block, so the fix works on the known path today. This mirrors the sibling preview modal's existing pattern — established file convention, not a regression. Worth hardening with a Portal later.
- **Not dynamically verified in a real browser.** The on-screen/non-trapping conclusion is static CSS + DOM analysis (jsdom computes no layout), corroborated by the identical, already-shipped preview modal in the same subtree. Recommend a real-viewport spot-check when convenient.
- **Advisory legs absent.** The cross-family advisory passes did not contribute this round (one returned an upstream API error, one exceeded its time budget). Their absence is recorded, not treated as approval — this verdict rests on the direct code review and local verification above.

## Verdict

**APPROVED.** The previous blocker (unscoped global restyle regressing the live panel) is fully resolved — both by scoping the CSS and by migrating the live `VersionPanel` into the shared overlay. The change is correct, targeted, and presentation-only; 31/31 tests pass with no new type errors. All remaining items are non-blocking P2 nits.

OctoBoooot (approved):
## Review: #674 — delta re-review (rebased onto merged shell; confirm-overlay fix intact + conflict resolved)

**Verdict:** Approve (re-affirmed at head `d54e72fe23d4`)
The branch is rebased onto a main that now includes the merged adapter PRs (#669/#670/#671) + #678, resolving the prior conflict (`mergeable=MERGEABLE`). Post-rebase the doc/sheet/board panels are all thin adapters over the shell, so this round's confirm-overlay fix lands in the **single shell** — cleaner than the prior per-panel version, and byte-verified correct. Flipping REVIEW_REQUIRED → APPROVE.

Risk tier: normal. Base `main`.

### Conflict resolved
Prior head was `CONFLICTING/DIRTY`; the rebase pulled in the merged #669/#670/#671 (adapters) + #678, and `mergeable` is now `MERGEABLE`. The blocker-3-equivalent (merge conflict) is cleared.

### Fix intact and improved by the new architecture
The real `d54e72fe` commit touches the **shell** (`VersionHistoryPanel.tsx` +81/-39, `styles.css`, `.test.tsx`), not the individual panels — because since my last review #670 merged, making the doc `VersionPanel` (and sheet/board) thin adapters over the shell. So:
- **Shell confirm boxes are overlay-hosted**: restore (`:587-601`) and delete (`:626-637`) each render `.octo-modal-overlay` + inner `.octo-version-confirm` with `role="dialog"`/`aria-modal`, `onMouseDown` stopPropagation on the card, and busy-guarded backdrop cancel (`if (!busy) setConfirmX(null)`); Escape-cancel added and busy-guarded (`:394-400`).
- **CSS scoped**: `.octo-modal-overlay .octo-version-confirm` present; **no bare unscoped card rule** remains (verified — no `.octo-version-confirm { … max-width/box-shadow }` global rule).
- **No inline remnant repo-wide**: the only non-test file rendering `.octo-version-confirm` is the shell (3 refs, all overlay-wrapped). `VersionPanel`/`SheetVersionPanel`/`BoardVersionPanel` now have **0** refs — they delegate to the shell. So there is no path left that could inherit misplaced-card styling.

This is strictly better than my previously-approved per-panel fix: one shell owns the confirm for all three ends, the overlay + scoped-CSS fix applies once, and the adapters can't diverge.

### Standing (unchanged)
The rgba→`--octo-modal-shadow` tokenization and the +60 shell test assertions (overlay wrap, stopPropagation, Escape/backdrop cancel) are present. The earlier double-Escape-handler note remains a benign non-blocking nuance (both handlers only dismiss).

Re-affirming APPROVE — conflict cleared, confirm-overlay + CSS-scoping fix verified intact and consolidated onto the shell post-rebase.

ploy-elison (commented):
Re-reviewed the rebased head (`d54e72fe`, base `main`, MERGEABLE). The rebase invalidated the earlier approval, so this is a fresh line-by-line pass over the three changed files. **APPROVED.**

Verified against the six concerns:

1. **Overlay rendering** — Both `confirmRestore` and `confirmDelete` now render the card inside the shared `.octo-modal-overlay` (`position: fixed; inset: 0`, z-index 1200), the exact wrapper the preview/diff modal uses, so the confirm stays in the viewport no matter how far the version list has scrolled. The inner card carries `role="dialog"` + `aria-modal="true"` + an `aria-label`; the overlay is `role="presentation"` with backdrop `onMouseDown` cancel, the card stops propagation, and a keydown effect handles Escape. Both dismissal paths are gated behind `if (!busy)` / `if (busy) return`, so an in-flight restore/delete can't be torn down by a stray backdrop click or keypress — and the confirm state only clears the confirm, never the mutation.

2. **Style scoping** — The card chrome (bg + border + shadow) is scoped to `.octo-modal-overlay .octo-version-confirm`, so it can't leak onto any inline render path. Since the class now only ever renders inside the overlay, this is airtight. (The two typography-only rules — `.octo-version-confirm > p:first-child` margin reset and `.octo-version-confirm-detail` muted color — remain unscoped, but they carry no positioning/chrome and no leak risk, so no change needed.)

3. **Shadow tokens** — The shadow is extracted into `--octo-modal-shadow` and reused by `.octo-modal`, the diff-card, and the confirm card. Net effect is *fewer* bare rgba shadows, not more (total rgba in the file drops 39->38); no new hardcoded shadow color introduced.

4. **Scope** — Diff touches only the three expected files (shell component, editor styles, shell test). No stray `octo-version-confirm` renderer survives outside the shell, and there's no leftover migration of the deleted inline VersionPanel confirm — that was correctly dropped on rebase.

5. **Regression coverage** — Two new tests: one asserts the confirm is wrapped by `.octo-modal-overlay`, carries `role="dialog"` + `aria-modal`, that a card-body click does not dismiss (stopPropagation) while a backdrop click and Escape both do; the second holds the restore mutation open and asserts backdrop-click and Escape are both no-ops while busy, then that the confirm clears once the mutation settles. Good coverage of the fix and the busy guard.

6. **Public interface** — `VersionHistoryPanelProps` and the component signature are untouched; only render internals plus one Escape effect changed. The doc/sheet/board adapters that depend on this shell are unaffected, and the existing board adapter tests still resolve `.octo-version-confirm` and its buttons through the overlay wrapping.

Clean, well-scoped, well-tested. Ready to merge.

yujiawei (approved):
# Code Review — PR #674 (octo-web)

**Scope:** `fix(versions): render version-history confirm box in shared overlay` — moves the restore/delete confirm box in the unified `VersionHistoryPanel` into the shared `.octo-modal-overlay`, adds backdrop/Escape cancel (both busy-guarded), and consolidates the modal elevation into a `--octo-modal-shadow` token. UI/CSS only; 3 files, +166/−46.

The diff does exactly what the PR describes and nothing more:

- **Overlay migration** — both confirm boxes (`confirmRestore`, `confirmDelete`) are wrapped in `.octo-modal-overlay` (`position: fixed; inset: 0`), the same viewport-anchored overlay the preview/diff modal already uses (`VersionHistoryPanel.tsx:585-660`). This is the correct fix for the reported bug (confirm falling below the fold on long scrolled lists).
- **Cancel paths** — backdrop `onMouseDown` and a new document-level `Escape` handler (`VersionHistoryPanel.tsx:397-406`) both clear only the confirm state and both no-op while `busy`, so a mutation in flight can't be torn down mid-request. The inner card uses `onMouseDown={e => e.stopPropagation()}`, matching the existing preview-modal pattern.
- **CSS scoping** — the card chrome is correctly narrowed from a global `.octo-version-confirm` to `.octo-modal-overlay .octo-version-confirm` (`styles.css:1738`). Verified there is no remaining render path for `.octo-version-confirm` outside an overlay (only `VersionHistoryPanel` renders the class, both instances inside the overlay; doc/sheet/board are thin adapters over this shell), so nothing is left orphaned/unstyled.
- **Shadow token** — `--octo-modal-shadow` replaces three identical hardcoded `rgba(0,0,0,0.24)` shadow literals (`.octo-modal`, `.octo-confirm-modal`, and the new confirm rule); net reduction in duplication with no visual change.
- No new API endpoint, no state-machine/race-guard changes, no permission-gate changes — the `canRestoreVersion(role)` gating on the restore/delete buttons is untouched. No out-of-scope additions.

## 2. Code quality — Approved

- **Correctness of the busy guard** — the confirm state clears itself after a successful mutation (`onDelete`/`onConfirmRestore` call `setConfirm*(null)` post-await), so guarding the cancel paths on `busy` cannot strand the dialog; the added regression test explicitly exercises this (holds the restore promise open, asserts backdrop-click + Escape are ignored until it settles).
- **No overlay stacking conflict** — confirm and preview overlays share `z-index: 1200`, but they can't be open simultaneously in practice: restore/delete are triggered from list rows (covered while the preview overlay is up), and `onConfirmRestore` calls `closePreview()`. The two `Escape` effects are mutually exclusive by their guard conditions.
- **Accessibility** — `role="dialog"`, `aria-modal="true"`, and an `aria-label` are set on the card; the backdrop is `role="presentation"`, consistent with the existing overlays in `EditorShell`/`SheetView`/`BoardShell`.
- **Tests** — two focused regression tests added; ran the full docs version suite locally: **345/345 passing** (`vitest run src/versions src/sheet src/board`).
- **Security note (PR flagged `security_sensitive`)** — this is a keyword false-positive. The change is purely presentational: no auth, token, credential, or data-flow surface is touched, and the role-based restore/delete permission gate is unchanged. Nothing here requires manual security verification beyond confirming that gate stayed intact (it did).

### Non-blocking nits (P2)

1. **Stale file-header comment (pre-existing, not in this diff).** `VersionHistoryPanel.tsx:10-12` still says "It changes NOTHING for users yet … does not wire any end to it (the doc/sheet/board panels are untouched)." That's no longer true — the adapters were wired in #669/#670/#671 and this very fix relies on the shell being live. Outside this diff's touched lines, so not a blocker, but worth a one-line docs cleanup in a follow-up.
2. **No initial focus move / focus-trap on the confirm dialog.** Escape works (document-level listener) and this matches the existing preview modal's behavior, so it's consistent rather than a regression — but moving focus into the dialog on open would be a small a11y improvement for both.

## Verdict

Clean, well-tested, correctly-scoped bug fix with no P0/P1 issues. Local tests green; the two typecheck errors in the repo (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) are pre-existing on the merge base and unrelated to this change.

**APPROVED**

Jerry-Xin (approved):
Re-reviewed from scratch at head `d54e72fe`, byte-verified (my prior b10d3b01 APPROVE was discarded — the fix was restructured after #670 merged). Still **APPROVE**.

## Context: the fix correctly followed the shell restructuring
Since my last review, main absorbed #670/#669/#671, making the doc/sheet/board panels all **thin adapters** over the unified `VersionHistoryPanel` shell. So the confirm box moved out of the doc `VersionPanel` into the shared shell, and this PR's fix now correctly targets the shell. The tip commit `d54e72fe` is the sole #674 fix (the 4 below it are the already-merged #678/#671/#669/#670); `main..head` prep diff contains exactly the 3 tip files.

## ✅ Original 🔴 resolved under the new architecture
1. **CSS scoped** (`styles.css:1738`): `.octo-modal-overlay .octo-version-confirm { max-width:420px; bg; box-shadow:var(--octo-modal-shadow); border-radius:12px; ... }`. No bare/global card-chrome rule remains — the only bare rules are typography (`:1749 > p:first-child` margin-reset, `:1752 -detail` color). An inline `.octo-version-confirm` gets zero modal chrome.
2. **Shell is the SOLE non-test renderer**: grep returns only `VersionHistoryPanel.tsx:594/633`. `VersionPanel.tsx:62` is now a "THIN ADAPTER" rendering `<VersionHistoryPanel>` (`:93`) — its inline confirm is gone; sheet/board adapters likewise delegate. So the scoped CSS is correct for every consumer; no inline consumer is left unstyled.
3. **Shell confirm overlay-wrapped** (`VersionHistoryPanel.tsx` restore 583-620 / delete 622-660): outer `.octo-modal-overlay role="presentation" onMouseDown={()=>{ if(!busy) setConfirmX(null) }}`, inner card `role="dialog"` / `aria-modal` / `onMouseDown stopPropagation`. New Escape handler (`:398-405`) busy-guarded (`if (e.key!=='Escape' || busy) return`). The confirm state machine / mutation flow / race guard / mounted-ref are byte-unchanged (pure JSX move + one Escape effect).

## ✅ rgba→token + tests
- `--octo-modal-shadow` at `:root:10`, consumed by confirm + preview + diff (1746/1064/1392). No hardcoded rgba in the confirm.
- `VersionHistoryPanel.test.tsx` adds real assertions: confirm is overlay-wrapped (`box.closest('.octo-modal-overlay')` truthy, `role=dialog`/`aria-modal`), card-body mousedown doesn't dismiss while backdrop does, Escape dismisses, and — with the restore promise held open — backdrop + Escape are no-ops while busy, then clear on release. Not superficial.

## 🟡 Non-blocking
- `.octo-modal-overlay` uses `align-items:flex-start; padding:48px 16px` — the confirm is top-aligned, not truly vertically centered, so the commit/test "viewport-centered" wording is imprecise. Harmless: the overlay is `fixed; inset:0; overflow-y:auto`, so the actual below-the-fold bug is fully fixed and the buttons are always on-screen.

APPROVE.

OctoBoooot (approved):
## Review: #674 — delta re-review (pure rebase; #674 fix byte-unchanged)

**Verdict:** Approve (re-affirmed at head `948127580f55`)
The new head is my already-approved #674 shell fix **rebased** onto a main that now also has merged #677 + #667 — the compare (`ahead=3 behind=1`) is inflated by those merges. The isolated `948127580f55` commit is the same 3-file shell change, and blob-compare confirms all 3 files (`styles.css`, `VersionHistoryPanel.tsx`, `.test.tsx`) are **byte-identical** to my approved head `d54e72fe23d4`. No code delta. Approval carries forward.

### Verified
- Isolated `948127580f55` (vs parent `6f7de421`) = 3 files, same +25/-7 / +81/-39 / +60/-0 as before.
- Blob hashes identical across all 3 #674 files vs the prior approved head → pure rebase, zero change to the fix.

### Standing (unchanged, verified prior round)
Confirm boxes overlay-hosted in the shell (restore/delete, `stopPropagation` + busy-guarded backdrop/Escape cancel); CSS scoped to `.octo-modal-overlay .octo-version-confirm` (no bare card rule); shell is the sole non-test renderer (adapters delegate); rgba→`--octo-modal-shadow` token; tests pin the overlay/cancel behavior. The top-aligned-vs-"centered" 🟡 remains (styles.css unchanged) — cosmetic, non-blocking; the below-fold bug is fully fixed regardless.

Pure rebase re-affirmation — no re-litigation.

yujiawei (approved):
# Code Review — PR #674 (octo-web)

**Head SHA reviewed:** `948127580f55ea3294650d8d1f5a439b7183b387`
**Verdict:** APPROVED

A tightly-scoped UX fix that re-anchors the version-history restore/delete confirm box inside the shared `.octo-modal-overlay` (fixed, viewport-centered) instead of letting it render at the end of the panel's content flow, where it fell below the fold on long scrolled lists. Because the doc/sheet/board panels are now thin adapters over `VersionHistoryPanel`, fixing the shell fixes all three ends. I checked out the head SHA, read the full diff against merge-base `6f7de421`, and ran the tests locally.

## 1. Spec compliance

The diff does exactly what the description states, nothing more:

- **Overlay-wrap** — both `confirmRestore` and `confirmDelete` are now wrapped in `.octo-modal-overlay` with the inner card carrying `role="dialog"`/`aria-modal="true"`, mirroring the existing preview modal (`VersionHistoryPanel.tsx:585–659`, cf. the preview modal at `:663`).
- **Backdrop + Escape cancel, both busy-guarded** — the overlay `onMouseDown` no-ops while `busy` (`:590`, `:629`), the inner card `stopPropagation`s (`:598`, `:637`), and the new Escape effect early-returns on `busy` and only clears confirm state, never the in-flight mutation (`:394–406`).
- **Confirm state machine / race guard untouched** — `onConfirmRestore` / `onDelete` and the mounted-ref liveness checks are unchanged.
- **Scoped card chrome** — CSS selector narrowed to `.octo-modal-overlay .octo-version-confirm` (`styles.css:1738`) so the modal-card look only applies to the overlay-anchored card.
- **Shared shadow token** — new `--octo-modal-shadow` (`styles.css:10`) replaces the three hardcoded `0 12px 48px rgba(...)` literals (`:1064`, `:1392`, `:1746`).
- **Regression tests** — two new tests assert the overlay wrapping, backdrop/Escape cancel, and the busy-guard holding a confirm open during an in-flight restore.

No missing items, no scope creep, no deviation from the described approach.

## 2. Code quality

Verified the cross-file concern first, since it is the one real risk in this change:

- **CSS selector narrowing is safe.** Narrowing `.octo-version-confirm` → `.octo-modal-overlay .octo-version-confirm` would silently drop the card styling from any path still rendering that class *outside* an overlay. I grepped the whole `packages/` tree: `.octo-version-confirm` is rendered only in `VersionHistoryPanel.tsx`, and both occurrences are now inside `.octo-modal-overlay`. The doc/sheet/board adapters (`VersionPanel.tsx`, `SheetVersionPanel.tsx`, `BoardVersionPanel.tsx`) render no confirm of their own — they delegate entirely to the shell. So there is no orphaned inline usage that loses styling.
- **Busy-guard consistency.** All three dismissal paths (backdrop, Escape, Cancel button) and the action buttons respect `busy`, so a stray click/keypress cannot tear the confirm down mid-request. This matches the state-machine invariants the mutation handlers already rely on.
- **Escape effect scoping.** The new keydown listener is registered only while a confirm is open and depends on `[confirmRestore, confirmDelete, busy]`, so it is added/removed correctly and does not leak. Note the preview modal has its own separate Escape listener (`:384`); a confirm is only opened from a version row (not from inside the preview), so the two overlays are not stacked in practice and there is no double-handling concern.
- **Accessibility improved** — the confirm now carries proper dialog roles and an `aria-label`, which the old inline box lacked.

### Verification performed
- `pnpm exec vitest run src/versions/` → **77 passed** (matches the PR's claim).
- `tsc --noEmit -p tsconfig.typecheck.json` → the only two errors are in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts`, both **outside this diff** and pre-existing; this change introduces none.

### Non-blocking suggestions (nits)
- **Duplicated overlay JSX.** The restore and delete confirm blocks are near-identical (~35 lines each). A small shared `ConfirmOverlay` wrapper (overlay + busy-guarded dismiss + dialog card) would remove the duplication and guarantee the two stay in sync. Not required for this fix.
- **Stale file-header comment.** `VersionHistoryPanel.tsx:9–12` still says the shell "does not wire any end to it … zero visible change until an adapter phase." That was true at phase 1 but the adapters (doc/sheet/board) are now live, so the comment reads as contradicting the PR. Worth refreshing in a follow-up; out of scope here.

## 3. Overall verdict

**APPROVED**

The change is correct, minimal, well-tested, and the one cross-file risk (CSS selector narrowing) is verified safe. No P0/P1 issues.

## Security note

This PR was auto-flagged `security_sensitive`, but that is a keyword false-positive — the diff touches only client-side confirm-box positioning, CSS tokens, and tests. No auth, tokens, credentials, migrations, or trust boundaries are involved. One benign UX property worth noting for a human: the destructive-action confirm is now dismissible via backdrop-click and Escape (both guarded so they cannot interrupt an in-flight restore/delete). This is standard modal behavior and matches the preview modal, so it does not weaken the destructive-action guard.

PR #677, opened 2026-07-13
fix(docs): show all documents in the sidebar, not just the first page

### LINKED ISSUE #676: bug(docs): sidebar only shows the first 20 documents
## Summary
The docs sidebar caps at 20 documents. In a space with more than 20 docs, the rest are not shown even though they exist.

## Root cause
The list backend (`GET /api/v1/docs`) paginates with a default `pageSize` of 20 and reports `total`. The frontend `DocsList.reload()` called `listDocs(...)` without a page size and rendered only the first page, so the sidebar silently capped at 20.

## Expected
All documents in the current space/folder are listed in the sidebar.

## Repro
1. Open a space that has more than 20 documents.
2. Observe the left sidebar shows only 20; older documents are missing.

## Fix direction
Fetch every page and concatenate until `total` is collected before rendering the list (preserving the existing stale-response guard).

wwwzzzcccc:
## Summary
The docs sidebar only rendered the first 20 documents. The list backend (`GET /api/v1/docs`) paginates with a default `pageSize` of 20 and reports `total`, but `DocsList.reload()` called `listDocs(...)` with no page size and rendered only the first response. In a space with more than 20 documents the rest were silently invisible.

This fetches every page (pageSize pinned to the backend maximum of 100) and concatenates until the reported `total` is collected, then sets the list in one update. The existing stale-response guard (XIN-417) is preserved so a superseded reload cannot overwrite the current space's list.

## Linked Spec
Fixes #676

## How verified
- Confirmed the target space has 39 documents (all in the default folder) via the backend metadata store; before the fix only 20 were listed.
- Simulated the new pagination loop against the live list query: `pageSize=100` returns all 39 in one page, the loop terminates correctly (`batch < PAGE_SIZE` and `collected >= total`), and the sidebar renders all 39.
- `tsc --noEmit` on the docs typecheck project: no errors in the touched file.
- Verified live via Vite HMR on the running dev server.

### DIFF
--- a/packages/docs/src/pages/DocsHome.tsx
+++ b/packages/docs/src/pages/DocsHome.tsx
@@ -345,12 +345,37 @@ function DocsList({
     const seq = ++reloadSeq.current
     setLoading(true)
     setError(null)
-    listDocs({ spaceId: space || undefined, folderId: folder || undefined, sort: 'updatedAt:desc' })
-      .then((res) => {
+    // The backend paginates (default pageSize 20, max 100) and reports `total`.
+    // The sidebar must show every document, not just the first page, so fetch
+    // all pages and concatenate. pageSize is pinned to the backend maximum (100)
+    // to minimize round-trips; the loop stops once we have collected `total`
+    // items (or a short page comes back, guarding against a stalled total).
+    const fetchAll = async (): Promise<DocListItem[]> => {
+      const PAGE_SIZE = 100
+      const all: DocListItem[] = []
+      let page = 1
+      for (;;) {
+        const res = await listDocs({
+          spaceId: space || undefined,
+          folderId: folder || undefined,
+          sort: 'updatedAt:desc',
+          page,
+          pageSize: PAGE_SIZE,
+        })
+        const batch = res?.items ?? []
+        all.push(...batch)
+        const total = res?.total ?? all.length
+        if (batch.length < PAGE_SIZE || all.length >= total) break
+        page += 1
+      }
+      return all
+    }
+    fetchAll()
+      .then((items) => {
         // A newer reload has superseded this one (e.g. the Space changed again while this
         // request was in flight) — drop the stale response so it can't overwrite the current list.
         if (seq !== reloadSeq.current) return
-        setItems(res?.items ?? [])
+        setItems(items)
       })
       .catch((err) => {
         if (seq !== reloadSeq.current) return


OctoBoooot (approved):
## Review: show all documents in the sidebar, not just the first page (#677)

**Verdict:** Approve
Correct, well-guarded pagination fix for a real bug — the docs sidebar rendered only the first 20 of N documents (backend default `pageSize=20`). Now fetches all pages to `total`. Termination is solid and the stale-response guard is correctly preserved across the multi-page loop. No defects.

Risk tier: normal. Base `main`; CI green (Build/osv/code-review pending, not failing). +28/-3, 1 file. Fixes #676.

### What I verified
- **Loop termination is guaranteed** (`DocsHome.tsx:368`): dual break — `batch.length < PAGE_SIZE` **OR** `all.length >= total`. The short-page guard is the robust one: a full page is exactly `PAGE_SIZE` (100, the backend max), so any short/empty final page exits regardless of what `total` reports — a stalled/wrong/missing `total` can't cause an infinite loop (and `total ?? all.length` makes a missing total exit after page 1). Empty batch (`0 < 100`) breaks immediately. `page` starts at 1 and increments only on continue — no off-by-one, no skipped/duplicated page.
- **Stale-response guard preserved and correctly scoped to the whole loop** (`:345`, `:377/381/390`): `seq = ++reloadSeq.current` is stamped once before `fetchAll()`, and `if (seq !== reloadSeq.current) return` guards **all three** terminal handlers (`.then`/`.catch`/`.finally`). So the entire multi-page fetch is one supersede unit — a space/folder switch mid-loop bumps the seq and the stale loop's `setItems`/`setError`/`setLoading` all no-op. The longer async window (multiple awaits) doesn't weaken this because the guard is at the settle point, not per-page. Critically, `setLoading(false)` is also guarded, so a stale request finishing first can't clear the spinner while the current reload is still in flight.
- **Error path intact**: `.catch` surfaces the error (console + `setError` + retry UI) rather than swallowing, guarded; deps `[space, folder]` correct; single `setItems` update at the end (no intermediate flicker).

### Praise
- The termination guard is the right shape — it doesn't trust `total` alone (which could stall/desync), it also exits on any short page, so the loop is safe against a misreporting backend; and pinning `pageSize` to the backend max (100) minimizes round-trips. The stale-guard is applied at the loop's *settle* boundary (all three of then/catch/finally) rather than per-page, so extending the single-fetch into a multi-page loop correctly preserves the XIN-417 space-switch protection — including keeping the spinner up until the *latest* reload settles. Both invariants (bounded loop, latest-wins) are documented in-line so the next editor won't regress them.

### Note
The N/100 sequential round-trips are fine for typical doc counts; if a space ever holds thousands of docs this is O(N/100) serial requests, but that's well beyond current scale and the single-batch-update keeps rendering cheap. Non-issue at present.

yujiawei (approved):
# Code Review — PR #677 (octo-web)

## Summary

The sidebar document list (`DocsList.reload` in `packages/docs/src/pages/DocsHome.tsx`) called `listDocs` without a `page`/`pageSize` and rendered only the first response. Since the backend paginates with a default `pageSize` of 20, spaces with more than 20 documents silently capped the sidebar at 20. This PR replaces the single call with a `fetchAll` loop that pages through the backend (`pageSize` pinned to the max of 100) and concatenates results before a single `setItems`.

**Verdict: APPROVED.** The change is correct, minimal, and matches its stated scope. Comments below are non-blocking suggestions only.

## Spec compliance

- ✅ Does what it claims: fetches every page and shows all documents.
- ✅ No scope creep: only `DocsList.reload` changes; `listDocs` already exposed `page`/`pageSize`/`total` (`docsApi.ts:42-48`, `25-28`), so no API change was needed.
- ✅ The XIN-417 stale-response guard is preserved — `seq !== reloadSeq.current` is still checked before `setItems`, `setError`, and clearing `loading`.

## Correctness

- Termination logic is sound. The loop breaks on either `batch.length < PAGE_SIZE` (short page) **or** `all.length >= total`. The short-page condition correctly guards against a stalled/incorrect `total`, so a backend that under- or over-reports `total` cannot cause an infinite loop as long as it eventually returns a short page.
- `res?.items ?? []` and `res?.total ?? all.length` defend against a malformed response; when `total` is absent the loop falls back to the short-page break. Reasonable.
- Types line up: `DocListItem[]` and `ListDocsResult` are consistent with `docsApi.ts`, and `tsc` on `packages/docs` is clean for this file.

## Non-blocking notes (P2 / suggestions)

1. **Stale-space requests keep paginating (minor waste, not a bug).** The stale-response guard now runs only after `fetchAll` fully resolves. If the user switches Space mid-pagination, the superseded `fetchAll` still fetches every remaining page of the old space before the guard discards the result. Correctness is unaffected (the result is dropped), but on a large old space this issues extra network round-trips. If you want to short-circuit, checking `seq !== reloadSeq.current` at the top of each loop iteration and breaking early would drop the wasted pages. Optional.

2. **Offset-pagination consistency across pages.** With `sort: 'updatedAt:desc'`, a document updated between page fetches can shift its offset, which can produce a duplicate or a skipped item across page boundaries. This is inherent to offset pagination and low-impact for a sidebar list; worth being aware of but not worth blocking. A keyset/cursor approach (if the backend supports it) would be the durable fix.

3. **No test for the multi-page path.** The regression being fixed is specifically "more than one page," yet the change ships without a test that returns >100 items across multiple pages and asserts the concatenated result (and, ideally, that `page` increments). The existing `DocsHome.test.tsx` cases all return a single page. A small test would lock in the fix and document the loop's break conditions. Recommended follow-up, not a blocker.

## Coverage note

Reviewed against head SHA `f224cf68a467746f375845750b7222790e93ee63`. Verified: the `listDocs` contract (`page`/`pageSize`/`total`/`items`), the only other `listDocs` caller (`editor/mention.ts`, unaffected — it intentionally uses `pageSize: 50` for the mention menu), the preserved stale-guard semantics, and TypeScript compilation for the changed file. Not exercised: a live backend run against a >100-document space (no runtime environment available in review); the logic is verified by inspection.

Jerry-Xin (approved):
Reviewed at head `f224cf68`, byte-verified. The fix correctly fetches all pages — **APPROVE**, with non-blocking hardening notes.

## ✅ Fix is correct
`DocsHome.tsx` `reload()` previously called `listDocs({spaceId, folderId, sort})` with no pageSize → backend default 20 → only first 20 docs shown. Now `fetchAll()` loops `page` 1→N at `PAGE_SIZE=100`, accumulates `all.push(...batch)`, and `setItems(all)` once. Verified against `ListDocsResult` types (`items`/`total`). Termination is dual-OR: `batch.length < PAGE_SIZE || all.length >= total` — for a well-behaved backend the short final page (or reaching `total`) terminates correctly; `total ?? all.length` guards a missing total. `page += 1` only on a non-terminal page (no skip/refetch). Single `setItems` after collection — no setState-in-loop / stale-closure race. The XIN-417 stale-seq guard (`if (seq !== reloadSeq.current) return`) is preserved in `.then/.catch`. `useEffect(reload,[reload])` with `reload=useCallback(...,[space,folder])` — no infinite re-render. Mid-loop failure rejects the whole fetch → `.catch` sets error, no partial render / retry storm.

## 🟡 Non-blocking (worth doing; robustness, not live bugs)
1. **`DocsHome.tsx:356` — no hard page/iteration cap.** The `for(;;)` relies on the backend behaving (short page or reachable `total`). A misbehaving backend that returns a full 100-item page forever while reporting an unreachable `total` would hang the loop. Not reachable against the current first-party backend (pageSize max 100, honors `page`, reports `total`), but a `total`-driven loop with no max-iteration bound is a known anti-pattern — add a cheap guard (`if (page > 200) break`).
2. **`DocsHome.tsx:354` — `PAGE_SIZE=100` couples correctness to backend max.** If the backend ever caps below 100, break-1 (`batch.length < PAGE_SIZE`) fires on the first page and silently reintroduces exactly the bug this PR fixes. Source it from a shared constant or don't hinge termination on it (the `all.length >= total` condition already covers completeness).
3. **`DocsHome.tsx:366` — no docId dedup.** With `updatedAt:desc` on a live dataset, a doc updated mid-fetch can shift across a page boundary → duplicated or skipped row. Low probability (single-page in practice); a dedup-by-id on accumulation removes it.

None block merge — the fix works correctly for the current backend and preserves all existing guards. Recommend addressing #1/#2 as quick follow-ups.

APPROVE.

Jerry-Xin (approved):
Reviewed at head `f224cf68`, byte-verified against the mirror. Correct, well-guarded pagination fix — **APPROVE** with non-blocking notes.

## What it does
`reload()` previously called `listDocs({spaceId, folderId, sort})` with no `pageSize` → backend default 20 → only the first 20 docs shown in the sidebar. Now a `fetchAll()` loops `listDocs({..., page, pageSize: 100})`, accumulating into `all[]`, and `setItems(all)` once complete. Verified it genuinely fetches all pages.

## ✅ Termination is sound (the classic pagination risk)
`if (batch.length < PAGE_SIZE || all.length >= total) break` — two OR'd exits:
- `batch.length < PAGE_SIZE` — any short/empty final page exits (a full page is exactly 100, backend caps at 100; an empty page is `0 < 100`). Always fires on the last page.
- `all.length >= total` — normal case; and critically, `total = res?.total ?? all.length`, so a **missing/stalled `total` can't infinite-loop** (`all.length >= all.length` breaks after page 1).
- `page` starts at 1, increments only on continue — no off-by-one, no page refetch.

## ✅ Stale-response guard (XIN-417) preserved across the multi-page loop
`seq = ++reloadSeq.current` is stamped before the loop and checked in all three of `.then`/`.catch`/`.finally` (including `setLoading`), so the entire `fetchAll()` is one supersede unit — a Space/folder switch mid-loop bumps the seq and the stale loop's terminal handlers all no-op (spinner stays on the latest reload). Widening single-fetch → multi-page didn't weaken it because the guard is at the settle boundary. `setItems` once after collection — no setState-in-loop race. Deps `[space, folder]` correct; no infinite re-render.

## 🟡 Non-blocking
- `DocsHome.tsx` (loop) — **no hard max-page/iteration cap.** Well-behaved termination is guaranteed above, but a `for(;;)` driven partly by `total` with no ceiling is the exact anti-pattern to harden: a misbehaving backend that returns full 100-item pages while `total` stays unreachable would hang. Cheap guard: `if (page > 200) break`. (Hardening, not a live bug against the trusted first-party API.)
- `PAGE_SIZE = 100` is coupled to the backend max being exactly 100. If the backend ever caps below 100, the `batch.length < PAGE_SIZE` break fires early and silently reintroduces the very bug this PR fixes. Fragile coupling worth a comment or a `total`-primary termination.
- No docId dedup on `all.push(...batch)`. With `updatedAt:desc` on a live dataset, a doc updated mid-fetch could shift across a page boundary → duplicate or skipped row. Low probability; minor.

Error handling surfaces failures (not swallowed) with retry; no partial render / retry storm. No new i18n key.

APPROVE.

PR #679, opened 2026-07-13
fix(loop): guard AssigneePicker candidate loads against stale-response race

MrAladdin:
## What

`AssigneePicker` refetches assignee candidates (`listAssigneeCandidates` → `setCands`) on mount and on every dropdown open (`onVisibleChange`), but had **no sequence guard**. With two loads in flight, the later-resolving one wins, so an older request can overwrite newer candidates after a directory mutation (`afterDirectoryMutation` clears the shared cache). Transient and self-healing, but a real stale-state race.

This is the sibling instance of the exact race class already fixed in `SettingsPage.tsx` (its `seqRef` guard). The #672 review flagged this AssigneePicker instance; it was merged as a known-issue follow-up (downgraded 🔴→🟡). This PR closes it before it reaches `main`.

## Fix

Mirror `SettingsPage`'s `seqRef` idiom in `packages/dmloop/src/ui/AssigneePicker.tsx`:
- Capture a monotonic token (`const my = ++seqRef.current`) synchronously before the fetch; only let the latest response call `setCands` (`if (my === seqRef.current)`).
- Increment `seqRef` in the mount effect's cleanup so **all** in-flight responses are invalidated on unmount (no `setState` after unmount) — covers both the mount fetch and any `onVisibleChange` fetch.

No prop/signature change.

## Blast radius (Rule 8)

Change is entirely internal to `AssigneePicker` — its props (`value`, `valueName`, `onChange`, `size`, `types`) are unchanged. All consumers (`NewLoopPage`, `CreateIssueModal`, `IssueList` batch bar, `IssueDetailPage`) are unaffected: they still get the same candidate list, just without the stale-overwrite window. Zero out-of-module impact.

## Verification

- **Real-env e2e** (dev, Alice, Loop-dev): the three-state picker (`CreateIssueModal` 负责人) loads candidates correctly after the change — menu shows 未指派 / 成员 Alice / Bob. The 执行方 picker (`types=[agent,squad]`) renders empty as expected (no agents in that workspace). No regression to candidate loading or selection.
- The race itself is a timing race (needs a directory mutation between two in-flight loads) — verified by code inspection that the guard is now present and correct; reviewers confirmed the token-capture/compare and unmount-invalidation logic.
- Codex adversarial-review: approve, no material findings. Claude code-reviewer: clean (verified token capture, unmount invalidation, latest-response-not-rejected, React 18 + StrictMode double-invoke). tsc clean.

## Follow-ups (not in this PR)
- `RunDetailModal` silently stops polling after `MAX_MISSES` without signalling "run no longer exists" (can stay on a stale running state).
- `packages/dmloop` race/polling logic still lacks focused tests.

### DIFF
--- a/packages/dmloop/src/ui/AssigneePicker.tsx
+++ b/packages/dmloop/src/ui/AssigneePicker.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { Dropdown, Avatar } from "@douyinfe/semi-ui";
 import { ChevronDown, User, Bot, Users, CircleSlash } from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
@@ -26,11 +26,20 @@ export interface AssigneePickerProps {
 export default function AssigneePicker({ value, valueName, onChange, size = "default", types }: AssigneePickerProps) {
   const { t } = useI18n();
   const [cands, setCands] = useState<AssigneeCandidate[]>([]);
+  // 请求序号:挂载 + 每次开框都会重取候选,多个 in-flight 时只让最新一次落地
+  // (镜像 SettingsPage 的 seqRef 守卫);旧请求后到不得覆盖新候选。
+  const seqRef = useRef(0);
 
   // 挂载先取一次;每次开框再取 —— afterDirectoryMutation 只清共享缓存,挂载中的 picker
   // 不会自动重读,故开框时刷新以反映其间的成员/agent/squad 增删。
-  const loadCands = () => { listAssigneeCandidates().then(setCands).catch(() => setCands([])); };
-  useEffect(() => { loadCands(); }, []);
+  const loadCands = () => {
+    const my = ++seqRef.current;
+    listAssigneeCandidates()
+      .then((list) => { if (my === seqRef.current) setCands(list); })
+      .catch(() => { if (my === seqRef.current) setCands([]); });
+  };
+  // 卸载时自增序号,使在途响应全部作废(避免卸载后 setState)。
+  useEffect(() => { loadCands(); return () => { seqRef.current++; }; }, []);
 
   const current = cands.find((c) => c.id === value);
   const allGroups: { type: AssigneeType; label: string }[] = [


OctoBoooot (approved):
## Review: guard AssigneePicker candidate loads against stale-response race (#679)

**Verdict:** Approve
Tiny, correct fix that closes the exact AssigneePicker stale-response race I flagged as a non-blocking 🟡 on #672 — mirrors the established `SettingsPage` `seqRef` idiom. No prop change, blast-radius contained. No defects.

Risk tier: normal. Base `feat/octo-loop`; CI green (`code-review` pending). +12/-3, 1 file.

### What I verified
- **Token captured synchronously before the fetch** (`:my = ++seqRef.current`) — so the compare is against the value at call time, not a stale closure.
- **Latest-wins on both resolution paths**: `.then((list) => { if (my === seqRef.current) setCands(list) })` and the `.catch` guard identically. An older in-flight load (lower `my`) can no longer overwrite newer candidates after a newer `loadCands` (or a directory mutation that triggers a reopen-refetch) bumped `seqRef`. Closes the stale-overwrite window.
- **Unmount invalidation done right**: the mount effect's cleanup `return () => { seqRef.current++ }` **increments** (not resets) the seq, so every in-flight response — the mount fetch *and* any `onVisibleChange` fetch — is invalidated on unmount; no `setCands` after unmount. Incrementing is the correct choice: no outstanding `my` can still equal the bumped `seqRef`.
- **React 18 StrictMode-safe**: mount→cleanup→remount bumps the seq on the cleanup, and the remount's fresh `loadCands` captures a higher `my`, so it wins over the first mount's in-flight response.
- **Contained**: props (`value`/`valueName`/`onChange`/`size`/`types`) unchanged; consumers (`NewLoopPage`, `CreateIssueModal`, `IssueList` batch bar, `IssueDetailPage`) get the same candidate list, just without the stale window.

### Praise
- This is the fix I recommended on #672, done exactly right: it reuses the repo's own proven `seqRef` latest-wins idiom (SettingsPage / IssuePage / the #677 pagination guard) rather than inventing a new mechanism, and it gets the two subtle points correct — capture-before-await and increment-on-unmount so both the mount and the on-open fetches are covered by a single guard. Closing the known-issue follow-up before it reaches `main` is the right call.

### Note
The body is honest that the race is timing-dependent (needs a directory mutation between two in-flight loads) and was verified by code inspection rather than a live repro — appropriate for a race this narrow; the guard logic is statically verifiable and correct. A unit test simulating out-of-order resolution (like the one suggested for #677) would pin it, but that's a nice-to-have, not required for a 1-file guard mirroring an established pattern.

yujiawei (approved):
# Code Review — PR #679 (octo-web)

**Verdict: APPROVE** · head SHA `b7041614c869835ce1cb3e72a7448358eb6e2ac4` · scope `packages/dmloop/src/ui/AssigneePicker.tsx` (+12/-3, single file)

## Spec compliance

The change does exactly what the title claims — guard `AssigneePicker` candidate loads against a stale-response race — and nothing more.

- **Missing:** none. `loadCands` runs from two paths (mount `useEffect` and `Dropdown` `onVisibleChange` open); both concurrent `listAssigneeCandidates()` promises can now resolve out of order without an older fetch clobbering newer data.
- **Extra:** none. No new props, no API-surface change, `onChange` contract untouched.
- **Deviation:** none. The `seqRef` monotonic-counter pattern mirrors the existing one in `SettingsPage.tsx:113-140`, keeping the codebase consistent.

**Spec: ✅**

## Correctness

The guard is sound. Verified by independent trace:

- `const my = ++seqRef.current` (increment-then-capture) ⇒ at issue time `my === seqRef.current`; any later `++` makes `my < seqRef.current`, so the invariant `my <= seqRef.current` holds for every outstanding request.
- `.then` / `.catch` apply state only when `my === seqRef.current`, so any superseded request is dropped. The latest request always lands.
- Unmount cleanup `return () => { seqRef.current++; }` pushes the counter strictly upward, so any in-flight response after unmount fails the equality check → no `setState` on an unmounted component.
- **React StrictMode double-invoke** traces cleanly: mount#1 `my=1` → cleanup `seqRef=2` → mount#2 `my=3`; the stale `R(my=1)` is dropped, `R(my=3)` lands. No wrongful drop of the latest and no mis-landing is reachable.

None of these block merge.

1. **`loadCands` in `useEffect([], …)` misses `react-hooks/exhaustive-deps`** (line 44). This is **pre-existing** (the prior code had the same shape) and matches the convention used elsewhere in this package (`useAssigneeCandidates.ts`, `SquadDetailPage.tsx:65`). If your lint config enforces the rule as an error, consider an inline `// eslint-disable-next-line react-hooks/exhaustive-deps` as done in `LoopPage.tsx:126` — but this is a style call, not a defect.
2. **Transient error wipes a populated list to empty** (line 39: `.catch(() => { if (my === seqRef.current) setCands([]); })`). Since `onVisibleChange` re-fires on every open, a transient failure on reopen replaces an already-loaded list with `[]`, blanking the menu until the next successful reload. Also **pre-existing** (identical `.catch(() => setCands([]))` before this PR), but the PR touches exactly this line, so it was an opportunity to preserve last-good state. The mirrored `SettingsPage` keeps loaded data on error and only sets an error flag — same pattern could apply here. Impact is limited: the selected assignee's label still renders via `current?.name ?? valueName`; only the dropdown list goes blank, and it self-heals on the next successful open. Directory data, not secrets.
3. **Readability of the unmount cleanup.** `seqRef.current++` in cleanup is correct but implicit — it relies on the `my <= seqRef` invariant holding for all callers. A `mounted` boolean or capturing the effect's own `my` would make intent explicit and be more robust to future edits. Optional.
4. **No test** for the guard. The change alters async ordering semantics, which is a classic silent-regression class — a future refactor could break the guard with no build error and only intermittent wrong-list symptoms. A small out-of-order-resolution test would lock the behavior in. Non-blocking given the +12/-3 size, but worth a follow-up.

## Security note

This PR was auto-flagged for extra scrutiny. The candidate payload is directory data (member/agent/squad names), not credentials or secrets. Cross-workspace leakage is not reachable: the directory cache is keyed on `slug === currentWorkspaceSlug()` with a generation counter (`ensureDirectory` in `api/directory.ts`), so a workspace switch invalidates in-flight builds before they commit. This client-side stale-response guard adds a second layer on top of that. No auth boundary is crossed by the change. Nothing here requires manual human security verification beyond the above.

## Summary

Minimal, correct fix that closes a real out-of-order-resolution race using the codebase's established `seqRef` idiom. No blocking issues. The four P2 items are advisory (two pre-existing, on touched lines) and can be addressed in a follow-up.

Jerry-Xin (approved):
Reviewed at head `b7041614`, byte-verified against the mirror. Minimal, correct stale-response guard — **APPROVE**.

## ✅ Guard is correct (byte-verified)
A monotonic seq token in a ref, mirroring the established `SettingsPage.tsx` idiom:
- `const seqRef = useRef(0)` (`AssigneePicker.tsx:31`) — ref, survives re-render, doesn't trigger one.
- `const my = ++seqRef.current` (`:36`) — captured **synchronously before** `listAssigneeCandidates()`.
- `.then(list => { if (my === seqRef.current) setCands(list) })` (`:38`) — compared **after** await; only the latest load commits.
- `.catch(() => { if (my === seqRef.current) setCands([]) })` (`:39`) — **error path also guarded**, so a stale rejection can't clobber the current list.
- Cleanup `return () => { seqRef.current++ }` (`:42`) — unmount bumps the seq, invalidating in-flight responses (no setState-after-unmount; StrictMode double-invoke safe).

## ✅ Covers ALL load paths
AssigneePicker has exactly two async triggers, both routed through the single guarded `loadCands`: mount (`:42` useEffect) and dropdown-open (`:87` `onVisibleChange`). There's no keyword-search or per-type load — `listAssigneeCandidates()` fetches the full list once and the agent/squad/member split is pure client-side filtering (`:50/:62`). So no unguarded second path; not a partial guard.

## ✅ No new bug
No off-by-one (seq init 0 → first load `my=1===1` commits the sole response); no stale closure (`my` is a fresh const per call, compared against the live ref); error path guarded; no separate loading flag to leak; no new i18n key, no dead code, selection semantics (`value/valueName/onChange/size/types`) unchanged. Only added import is `useRef`.

🟡 Non-blocking: no focused automated test for the timing race (a deferred-promise test would pin it) — already noted as a follow-up in the PR body.

APPROVE.

PR #150, opened 2026-05-27
fix(chat): refresh recent ordering on live updates

botshen:
## Summary
- Fixes #142
- Make ChatVM recent-conversation sorting produce a new array reference after live updates
- Preserve pinned conversations ahead of non-pinned conversations
- Add a focused regression test for live recent-list ordering

## Test
- pnpm --filter @octo/base exec vitest run src/Pages/Chat/__tests__/vm.sortConversations.test.ts

## Linked Issue
- Fixes #142

Issue #511, opened 2026-07-02
Forward a document into chat as a rich card, with optional access grant + permission level


Jerry-Xin:
## Summary

Forwarding a document into a chat/conversation currently produces no usable result. There is no way to share an in-app document (octo docs) into a conversation as a rich card, and no way to grant access to the recipient at the same time.

Expected behavior:
1. From a document, the user can forward/share it into a chat or group.
2. The forwarded item appears in the message stream as a **document card** showing at least the document **title** (plus optional summary/thumbnail).
3. During forwarding, the user can optionally **grant access** to the recipient(s). If granting is chosen, the user can pick the **permission level** (e.g. reader / commenter / writer), reusing the existing document role model.

## Current state (code analysis)

Feature does **not exist** in the frontend today. Key findings against `apps/web/src`, `packages/docs/src`, `packages/dmworkbase/src`:

**Forwarding is chat-message only, never documents**
- Single-message forward: `packages/dmworkbase/src/Components/Conversation/index.tsx:479` `fowardMessageUI(message)` clones an existing message content into the target conversation.
- Merge-forward: `packages/dmworkbase/src/Components/Conversation/vm.ts:606` `sendMergeforward()` packs selected messages into `MergeforwardContent` (contentType=11).
- Target picker: `packages/dmworkbase/src/Components/ForwardModal/ForwardModal.tsx` is a people/group picker only; its confirm callback returns a channel list with **no** document/permission parameters.
- The docs module (`packages/docs/src`) has **no** "send to chat"/forward entry point. The editor toolbar (`packages/docs/src/editor/Toolbar.tsx`) only exposes editing actions.

**No document-card message type exists**
- Message types are registered in `packages/dmworkbase/src/module.tsx:255-308`. Existing types: text(1), image(2), gif(3), voice(4), smallVideo(5), location(6), card(7), file(8), mergeForward(11), richText(14), summary(15), screenshot(20).
- `card` (contentType=7, `packages/dmworkbase/src/Messages/Card/index.tsx:51`) is a **person business card** (uid/name/avatar), not a document card.
- `file` (contentType=8, `packages/dmworkbase/src/Messages/File/index.tsx`) renders **uploaded file attachments** (.doc/.pdf), not online docs.
- No doc/document constant exists in `packages/dmworkbase/src/Service/Const.ts:47-84`.
[...]

PR #680, opened 2026-07-13
fix(web): land /summary/detail?taskId= deep link on boot (#150)

Ranwanglc:
## What & why
`Closes #150` (web side).

A completed-summary notification DM carries a deep link `${origin}/summary/detail?taskId=<id>`. Clicking it opens the browser at that URL, but the app boot renders the default ChatPage and never consumes `location.pathname`, so the detail page never opened — this is the front-end root cause of "点了跳不过去".

## Change
- `apps/web/src/App/index.tsx`: add a startup landing hook `useSummaryDetailLandingHandler` (thin adapter injecting browser deps), called next to the existing `useRealnameVerifiedLandingHandler`.
- `apps/web/src/App/summaryDetailLanding.ts` (new): pure DI logic `runSummaryDetailLanding` — on `/summary/detail` (trailing slash tolerated) and when logged in, parse a canonical positive-integer `taskId` and call `WKApp.openSummaryDetail(taskId)`. Because `openSummaryDetail` is registered by the dmworksummary module init and may not be ready on the first effect, it retries every 100ms up to 20 times (~2s), then gives up. On success it cleans the URL (preserving the hash). Fully guarded (SSR/non-browser => no-op). Logic is extracted (mirroring the repo's `friendApplyReddotCleanup.ts` pattern) so it can be unit-tested off the heavy `@octo/base` module graph.
- `apps/web/src/__tests__/summaryDetailLanding.test.tsx` (new): 10 cases (happy path, trailing slash, wrong path, missing/non-numeric/0/negative/float/hex taskId rejected, not-logged-in, retry-until-ready, retry-cap no-throw, URL cleanup).

## Deploy prerequisite (env vars)
None on the web container itself. The link is produced by the worker side (PR on octo-smart-summary): the worker must set `SUMMARY_WEB_BASE_URL=<this web origin>` for the link to be emitted. Empty there => no link (safe plain-text fallback), and this handler simply never fires.

## Tests
`cd apps/web && npx vitest run src/__tests__/summaryDetailLanding.test.tsx` → 10 passed.
(Repo has ~33 pre-existing unrelated test failures — CSS dimensions / avatar positions — not touched here.)

### DIFF
--- a/apps/web/src/App/summaryDetailLanding.ts
+++ b/apps/web/src/App/summaryDetailLanding.ts
@@ -1,0 +1,61 @@
+export type SummaryDetailLandingDeps = {
+  // 拆成 getter 而非直取值：landing 时序上 openSummaryDetail 由 dmworksummary 模块
+  // init 注册，首帧 effect 执行时可能尚未挂到 WKApp 上，故每次重试都重新读取。
+  getPathname: () => string;
+  getSearch: () => string;
+  isLoggedIn: () => boolean;
+  getOpenSummaryDetail: () => ((taskId: number) => void) | undefined;
+  cleanUrl?: (taskId: number) => void;
+  setRetry: (cb: () => void, delayMs: number) => number;
+  clearRetry: (handle: number) => void;
+};
+
+const SUMMARY_DETAIL_PATH = '/summary/detail';
+const SUMMARY_DETAIL_RETRY_INTERVAL_MS = 100;
+const SUMMARY_DETAIL_MAX_RETRIES = 20; // ~2s 兜底：等 dmworksummary 模块 init 注册 openSummaryDetail
+
+/**
+ * URL 直达落地：通知 DM 里的 `${origin}/summary/detail?taskId=<id>` 被点击后浏览器
+ * 直接打开该地址，但 App 启动默认渲染 ChatPage、不消费 location。这里在启动时读一次
+ * location，命中该路径且已登录时调用 WKApp.openSummaryDetail(taskId) 落到详情页。
+ *
+ * 纯依赖注入以便脱离 @octo/base 重模块图做单测（对齐 friendApplyReddotCleanup.ts 风格）。
+ *
+ * @returns cleanup 函数（停止仍在排队的重试）；返回 undefined 表示未启动任何重试。
+ */
+export function runSummaryDetailLanding(deps: SummaryDetailLandingDeps): (() => void) | undefined {
+  try {
+    const pathname = (deps.getPathname() || '').replace(/\/+$/, '') || '/';
+    if (pathname !== SUMMARY_DETAIL_PATH) return undefined;
+
+    // 未登录不动作：登录前落地会被登录流程打断，且详情页依赖登录态。
+    if (!deps.isLoggedIn()) return undefined;
+
+    const raw = new URLSearchParams(deps.getSearch()).get('taskId');
+    // 只收正整数：后端 %d 只发真实 ID，负数/小数/十六进制/超 2^53 均非法，收严更防御。
+    const taskId = Number(raw);
+    if (raw === null || !/^\d+$/.test(raw.trim()) || !Number.isSafeInteger(taskId) || taskId <= 0) return undefined;
+
+    let handle: number | undefined;
+    let attempts = 0;
+    const tryOpen = () => {
+      const open = deps.getOpenSummaryDetail();
+      if (open) {
+        open(taskId);
+        // 清 URL 回干净路径：避免用户刷新时重复落地（openSummaryDetail 已切到 summary 主 Tab）。
+        deps.cleanUrl?.(taskId);
+        return;
+      }
+      if (++attempts >= SUMMARY_DETAIL_MAX_RETRIES) return;
+      handle = deps.setRetry(tryOpen, SUMMARY_DETAIL_RETRY_INTERVAL_MS);
+    };
+    tryOpen();
+
+    return () => {
+      if (handle !== undefined) deps.clearRetry(handle);
+    };
+  } catch (e) {
+    // URL API / location 在 SSR / 非浏览器环境下可能不可用——静默忽略不阻塞渲染。
+    return undefined;
+  }
+}
--- a/apps/web/src/App/index.tsx
+++ b/apps/web/src/App/index.tsx
@@ -11,6 +11,7 @@ import { ContactsIcon } from '../Components/Icons/ContactsIcon';
 import { SummaryIcon } from '../Components/Icons/SummaryIcon';
 import { Toast } from '@douyinfe/semi-ui';
 import { clearDeprecatedFriendApplyReddotOnce } from './friendApplyReddotCleanup';
+import { runSummaryDetailLanding } from './summaryDetailLanding';
 
 let _summaryBadgeCount = 0;
 let _badgeListenerSetup = false;
@@ -45,8 +46,26 @@ function useRealnameVerifiedLandingHandler() {
   }, []);
 }
 
+function useSummaryDetailLandingHandler() {
+  useEffect(() => {
+    const cleanup = runSummaryDetailLanding({
+      getPathname: () => window.location.pathname,
+      getSearch: () => window.location.search,
+      isLoggedIn: () => WKApp.loginInfo.isLogined(),
+      getOpenSummaryDetail: () => WKApp.openSummaryDetail,
+      cleanUrl: () => {
+        window.history.replaceState(null, '', '/' + window.location.hash);
+      },
+      setRetry: (cb, delayMs) => window.setTimeout(cb, delayMs),
+      clearRetry: (handle) => window.clearTimeout(handle),
+    });
+    return cleanup;
+  }, []);
+}
+
 function App() {
   useRealnameVerifiedLandingHandler()
+  useSummaryDetailLandingHandler()
   useDeprecatedFriendApplyReddotCleanup()
   registerMenus()
   return (
--- a/apps/web/src/__tests__/summaryDetailLanding.test.tsx
+++ b/apps/web/src/__tests__/summaryDetailLanding.test.tsx
@@ -1,0 +1,130 @@
+import { describe, it, expect, vi } from 'vitest'
+import { runSummaryDetailLanding, type SummaryDetailLandingDeps } from '../App/summaryDetailLanding'
+
+/**
+ * Coverage for #150: a summary-completed DM carries `${origin}/summary/detail?taskId=<id>`.
+ * Clicking it opens the browser at that URL, but App boot renders the default ChatPage and
+ * never consumes location — so the landing handler must call WKApp.openSummaryDetail(taskId).
+ *
+ * Tested against the pure `runSummaryDetailLanding` DI helper (no @octo/base module graph),
+ * matching the house style of mainMenuReconcile.test.tsx. Retries run synchronously here via a
+ * setRetry stub that invokes its callback immediately, so we can assert the timing fallback for
+ * a not-yet-registered openSummaryDetail without real timers.
+ */
+
+function makeDeps(overrides: Partial<SummaryDetailLandingDeps>): {
+  deps: SummaryDetailLandingDeps
+  open: ReturnType<typeof vi.fn>
+} {
+  const open = vi.fn()
+  const deps: SummaryDetailLandingDeps = {
+    getPathname: () => '/summary/detail',
+    getSearch: () => '?taskId=123',
+    isLoggedIn: () => true,
+    getOpenSummaryDetail: () => open,
+    setRetry: (cb) => {
+      cb() // synchronous retry for deterministic tests
+      return 0
+    },
+    clearRetry: () => {},
+    ...overrides,
+  }
+  return { deps, open }
+}
+
+describe('runSummaryDetailLanding (#150)', () => {
+  it('opens the detail page when path=/summary/detail?taskId=123 and logged in', () => {
+    const { deps, open } = makeDeps({})
+    runSummaryDetailLanding(deps)
+    expect(open).toHaveBeenCalledTimes(1)
+    expect(open).toHaveBeenCalledWith(123)
+  })
+
+  it('tolerates a trailing slash on the pathname', () => {
+    const { deps, open } = makeDeps({ getPathname: () => '/summary/detail/' })
+    runSummaryDetailLanding(deps)
+    expect(open).toHaveBeenCalledWith(123)
+  })
+
+  it('does nothing when the pathname is not /summary/detail', () => {
+    const { deps, open } = makeDeps({ getPathname: () => '/' })
+    runSummaryDetailLanding(deps)
+    expect(open).not.toHaveBeenCalled()
+  })
+
+  it('does nothing when taskId is missing', () => {
+    const { deps, open } = makeDeps({ getSearch: () => '' })
+    runSummaryDetailLanding(deps)
+    expect(open).not.toHaveBeenCalled()
+  })
+
+  it('does nothing when taskId is non-numeric', () => {
+    const { deps, open } = makeDeps({ getSearch: () => '?taskId=abc' })
+    runSummaryDetailLanding(deps)
+    expect(open).not.toHaveBeenCalled()
+  })
+
+  it('rejects non-canonical taskId values (0, negative, float, hex)', () => {
+    for (const q of ['?taskId=0', '?taskId=-5', '?taskId=1.5', '?taskId=0x1F']) {
+      const { deps, open } = makeDeps({ getSearch: () => q })
+      runSummaryDetailLanding(deps)
+      expect(open, `taskId query ${q} must be rejected`).not.toHaveBeenCalled()
+    }
+  })
+
+  it('does nothing when the user is not logged in', () => {
+    const { deps, open } = makeDeps({ isLoggedIn: () => false })
+    runSummaryDetailLanding(deps)
+    expect(open).not.toHaveBeenCalled()
+  })
+
+  it('retries until openSummaryDetail is registered, then stops', () => {
+    let ready = false
+    const open = vi.fn()
+    let calls = 0
+    const deps: SummaryDetailLandingDeps = {
+      getPathname: () => '/summary/detail',
+      getSearch: () => '?taskId=7',
+      isLoggedIn: () => true,
+      getOpenSummaryDetail: () => (ready ? open : undefined),
+      setRetry: (cb) => {
+        // Ready on the 3rd attempt; each setRetry drives the next synchronous poll.
+        if (++calls === 2) ready = true
+        cb()
+        return calls
+      },
+      clearRetry: () => {},
+    }
+    runSummaryDetailLanding(deps)
+    expect(open).toHaveBeenCalledTimes(1)
+    expect(open).toHaveBeenCalledWith(7)
+  })
+
+  it('gives up after the retry cap without throwing when openSummaryDetail never appears', () => {
+    let calls = 0
+    const deps: SummaryDetailLandingDeps = {
+      getPathname: () => '/summary/detail',
+      getSearch: () => '?taskId=9',
+      isLoggedIn: () => true,
+      getOpenSummaryDetail: () => undefined, // never registers
+      setRetry: (cb) => {
+        calls++
+        cb()
+        return calls
+      },
+      clearRetry: () => {},
+    }
+    expect(() => runSummaryDetailLanding(deps)).not.toThrow()
+    // Initial attempt + retries, capped: setRetry is called (MAX_RETRIES - 1) times because
+    // the final attempt hits the cap and returns without scheduling. Must be finite (no infinite loop).
+    expect(calls).toBe(19)
+  })
+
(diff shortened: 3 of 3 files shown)


OctoBoooot (approved):
## Review: land /summary/detail?taskId= deep link on boot (#680)

**Verdict:** Approve
Clean, well-tested boot-time deep-link handler for the summary-notification DM link. Strict taskId validation, a bounded non-leaking retry for the module-init timing, hash-preserving URL cleanup, and full SSR/logged-out guards. DI-extracted for real unit testing. No defects.

Risk tier: normal. Base `main`; CI green (Build/code-review pending, not failing). +210/-0, 3 files (1 pure-logic + adapter hook + 10-case test). Closes #150 (web side).

### What I verified
- **taskId validation is strict — no injection/navigation risk.** `raw === null || !/^\d+$/.test(raw.trim()) || !Number.isSafeInteger(taskId) || taskId <= 0` → rejects missing, any non-ASCII-digit form (float/hex/negative/whitespace/injection strings), unsafe-int (>2^53), and 0/negative. The `/^\d+$/` regex is the load-bearing guard; `Number.isSafeInteger` + `>0` layer on top. So `WKApp.openSummaryDetail(taskId)` only ever receives a canonical positive safe integer.
- **Retry loop bounded + leak-free.** Caps at `SUMMARY_DETAIL_MAX_RETRIES=20` (~2s at 100ms) for `openSummaryDetail` to be registered by the dmworksummary module init; returns immediately on success (no reschedule); single-flight (one `handle` at a time). The function returns a cleanup that clears the pending `handle`, and the effect wires `return cleanup` — so React cancels in-flight retries on unmount. No leaked timer, no callback-after-unmount, no infinite loop.
- **Timing handled by DI-getter.** `getOpenSummaryDetail: () => WKApp.openSummaryDetail` is re-read each retry, correctly handling that the summary module may register `openSummaryDetail` after the first effect frame.
- **URL cleanup preserves hash + only on success.** `cleanUrl` = `replaceState(null, '', '/' + window.location.hash)` — resets the path (so a refresh won't re-land) while keeping the hash; fires only after a successful open.
- **Guards.** Path mismatch (trailing-slash tolerated via `replace(/\/+$/, '')`) → no-op; not-logged-in → no-op (detail page needs auth, and pre-login landing would be interrupted by the login flow); any throw / SSR (URL/location unavailable) → caught → no-op, never blocks render.
- **Tests** cover exactly these branches (10 cases: happy, trailing slash, wrong path, missing/non-numeric/0/negative/float/hex taskId rejected, logged-out, retry-until-ready, retry-cap no-throw, URL cleanup).

### Praise
- The pure-DI extraction (mirroring `friendApplyReddotCleanup.ts`) is the right structure — it makes a boot-timing/URL/retry handler unit-testable off the heavy `@octo/base` graph, and the 10 cases exercise every validation and timing branch. The taskId validation is defensively layered (regex + safe-int + positivity) rather than trusting `Number()` alone, and the retry is both bounded (can't spin forever) and cancellable (effect cleanup clears the timer) — the two things this pattern usually gets wrong, both correct here.

### Note (nit, non-blocking)
`cleanUrl` rewrites to `/` (root) rather than, say, the summary tab's own path — harmless because `openSummaryDetail` performs an in-app tab switch (not URL-routed), so the URL only needs to stop re-triggering on refresh; resetting to `/` achieves that. If deep links to other boot paths are ever combined, revisit, but fine as-is.

Jerry-Xin (dismissed):
Reviewed at head `f3b996f03b84`, byte-verified against the mirror. Small, surgical deep-link landing (+210/-0) — **APPROVE**.

## ✅ taskId parse + validation (injection-safe)
`summaryDetailLanding.ts:37-40`:
```js
const raw = new URLSearchParams(deps.getSearch()).get('taskId');
const taskId = Number(raw);
if (raw === null || !/^\d+$/.test(raw.trim()) || !Number.isSafeInteger(taskId) || taskId <= 0) return undefined;
```
- URLSearchParams `.get()` returns the first value → `?taskId=a&taskId=b` handled safely.
- Strict canonical positive-integer gate: rejects negatives/floats/hex/whitespace/empty/`>2^53`/`<=0`; missing → no-op.
- No injection/XSS: taskId is coerced to a `number` and passed as a React prop (`open(taskId)` → `<SummaryDetailPage taskId={taskId}/>`), never echoed into href/innerHTML or concatenated into a route; URL cleanup writes a fixed literal `'/' + hash`, not the taskId.

## ✅ Boot ordering / no redirect-loop / run-once
- `useEffect(…, [])` → runs once per mount, not per render.
- `openSummaryDetail` is registered by `dmworksummary/module.tsx:58` (optional on WKApp); since that module may init later, the handler polls the getter every 100ms up to 20× (~2s), re-reading the getter each retry (correct DI, no stale capture), bounded, with a cleanup that clearTimeouts on unmount — no hanging timer, no infinite loop (test asserts calls===19 at cap).
- No redirect loop: it drives the internal SPA route (`switchToMenuById("summary")` + `routeRight.replaceToRoot`) then cleans the browser URL to `/` — not a browser push to `/summary/detail`. Gated by `isLogined()`.

## ✅ Path match / history / no-op
- Exact match: `pathname.replace(/\/+$/,'')||'/'` then `!== '/summary/detail'` — matches `/summary/detail[/]`, does NOT false-match `/summary/detail-foo` or substrings.
- History = `replaceState` (correct for a boot deep-link, no dangling back entry).
- Non-deep-link boot: first guard returns immediately → zero side effects; whole body try/catch (SSR-safe).

## ✅ Tests — strong (10 cases)
Happy path, trailing slash, wrong-path no-op, missing/non-numeric/0/negative/float/hex rejection, not-logged-in no-op, retry-until-ready, retry-cap-no-throw, URL-cleanup. Assertions correct (cap arithmetic verified). Genuinely covers malformed + no-op + retry-bounds, not just happy path.

🟡 Non-blocking: logged-out/expired-session click doesn't preserve taskId post-login (`Layout/index.tsx:163`) — explicitly out of the PR's "when logged in" scope and a pre-existing login-flow limitation, not introduced here. Track as follow-up.

Matches the established `friendApplyReddotCleanup.ts` DI pattern. No i18n keys, no dead code, idempotent wiring. APPROVE.

yujiawei (changes requested):
# Code Review — PR #680 (octo-web)

Reviewed at head SHA `f3b996f0` against merge-base `9e19bb16`. Diff: 3 files, +210. New unit test suite passes locally (10/10 via `npx vitest run src/__tests__/summaryDetailLanding.test.tsx`). The pre-existing ~33 unrelated test failures and monorepo `tsc` errors are not touched by this PR; the new file itself is type-clean.

The design is good: pure DI logic extracted for testability (mirroring `friendApplyReddotCleanup.ts`), strict positive-integer `taskId` validation, SSR guard. The problem is that the change does not actually fix the primary reported flow.

## 1. Spec compliance

**Spec: ❌**

Issue #150 is "clicking the summary-completed DM deep link doesn't open the detail page." The most common real-world path for that link is a recipient who is **not currently signed in** in the tab/device that opens it (a fresh browser tab, a cold app open from a notification). That path is not handled, so the headline bug is not resolved for it.

- **Omission — logged-out deep link silently loses `taskId` across the login redirect** (`apps/web/src/App/summaryDetailLanding.ts:32`, `apps/web/src/App/index.tsx:49-64`, `apps/web/src/Layout/index.tsx:110-163`).
  1. Anonymous visitor opens `/summary/detail?taskId=<id>`. The handler bails at `summaryDetailLanding.ts:32` (`if (!deps.isLoggedIn()) return undefined;`) and stores nothing.
  2. The login screen renders in place (pathname stays `/summary/detail`).
  3. After sign-in, `onLogin`/`goMain` derives `basePath` from the pathname but builds `redirectQuery` with **only** `sid`/`doc`/`space`/`folder` (`Layout/index.tsx:126-132`) and reloads via `window.location.href = \`${origin}${basePath}/${sidParam}\`` (`Layout/index.tsx:163`) — `taskId` is never carried.
  4. On the post-login reload the handler re-runs, is now logged in, but `taskId` is gone from the URL → `raw === null` → returns undefined. The detail page never opens.

  The repo already has the pattern for this: `pendingInviteCode`, `persistStandaloneReturn`/`consumeStandaloneReturn`, and the `forwardDoc`/`forwardSpace` carry-through in `onLogin`. The fix should stash the pending `taskId` before falling through to login (or add a `summary`/`taskId` carry-through to the `onLogin` redirect) and consume it after authentication. Until then, the fix only works for a user who is already signed in at click time.

Nothing is over-built and the approach doesn't diverge from a stated plan; the single spec failure is the omission above.

## 2. Code Quality

- **[P2] The retry loop guards a condition that can never be false, and does not guard the one that can** (`summaryDetailLanding.ts:41-51`, `packages/dmworksummary/src/module.tsx:58-63`, `apps/web/src/Pages/Main/index.tsx:298-320`).
  `WKApp.openSummaryDetail` is assigned **synchronously** inside `SummaryModule.init()`, which runs at `apps/web/src/index.tsx:73` *before* `createRoot`/`render` at `index.tsx:82-83`. So `getOpenSummaryDetail()` is already truthy on the very first synchronous `tryOpen()`, and the "100ms × 20 (~2s)" retry never actually waits for anything — it is effectively dead code on the happy path.
  The dependency that *can* be missing is the route wiring: `openSummaryDetail`'s body calls `WKApp.routeLeft.popToRoot()` / `WKApp.routeRight.replaceToRoot(...)` (`module.tsx:60-61`, **not** optional-chained, unlike `switchToMenuById?.` on line 59). Those setters are `undefined` until `MainPage`'s `WKViewQueue.componentDidMount` wires them (`Main/index.tsx:298-320`). For a cold open with no cached `currentSpaceId`, `Layout` renders `<SpaceGate/>` instead of `<MainPage/>` (`Layout/index.tsx:428`), so the first `tryOpen()` calls `open(taskId)` → `routeLeft.popToRoot()` → `TypeError`. It is swallowed by the outer `try/catch`, so no crash — but the landing silently fails and `cleanUrl` never runs, and the retry cannot recover because it only re-polls `openSummaryDetail`, not the route wiring.
  In practice this self-heals: `SpaceGate.enterSpace` does `window.location.reload()` (`SpaceGate/index.tsx:82-84`) which preserves `?taskId=`, and the second boot has a cached space → `MainPage` mounts synchronously → landing works. So it is transient, not a hard failure — hence P2 — but the retry as written protects the wrong thing. Consider polling readiness of the route setters (or making `openSummaryDetail` internally tolerant of unwired managers) instead.

- **[P2] Retry-scheduled `tryOpen` runs outside the `try/catch`** (`summaryDetailLanding.ts:27,50,52,57`).
  Only the first synchronous `tryOpen()` (line 52) is covered by the `try{…}catch` (lines 27/57). Subsequent attempts are scheduled via `deps.setRetry(tryOpen, …)` (line 50) = `window.setTimeout` in production (`index.tsx:59`) and run on a later macrotask after the frame has unwound. A throw from `open(taskId)` on a retried attempt is therefore an uncaught timer error — and that is exactly the path (late registration + not-yet-wired route managers) where a throw is most likely. Move the guarded body inside `tryOpen` so retries get the same protection the SSR-guard comment on line 58 promises.

- **[P2] `cleanUrl` discards the entire query string** (`apps/web/src/App/index.tsx:56-58`).
  `replaceState(null, '', '/' + window.location.hash)` drops all query params, unlike the sibling `useRealnameVerifiedLandingHandler` which deletes only its own param and preserves the rest (`index.tsx:36-41`). Low impact today because the notification URL carries only `taskId` at boot and `openSummaryDetail` uses the in-memory `ContextRouteManager` rather than `Route.push` (which injects `?sid=`), but removing only `taskId` and preserving pathname/hash/other params is the safer and consistent choice.

- **[P2] Load-bearing guards have no test coverage.**
  - `Number.isSafeInteger` overflow rejection (`summaryDetailLanding.ts:37`) is asserted only for small values; add a `> 2^53` case (e.g. `?taskId=99999999999999999`).
  - The returned cleanup/`clearRetry` disposer (line 54-56) that the `useEffect` return relies on is never exercised — every test stubs `clearRetry` as a no-op and never calls the returned function. Add a case that starts a retry, invokes the returned cleanup, and asserts `clearRetry` was called with the pending handle.

- **[nit]** `raw.trim()` is used for the regex test but `Number(raw)` (untrimmed) for the value (`summaryDetailLanding.ts:36-37`). Harmless because `Number(' 12 ')` still yields `12`, but tidier to trim once.

## 3. Overall verdict

**REQUEST_CHANGES** — the Spec gate fails (the primary logged-out deep-link path is not handled), which alone blocks. The P2 items are quality improvements that would make the landing robust on cold opens.

## 4. Suggested direction

1. Persist the pending `taskId` before falling through to login (reuse the `pendingInviteCode` / `persistStandaloneReturn` pattern), and consume it in the post-login flow — or add a `taskId` carry-through to `onLogin`'s redirect query alongside `forwardDoc`.
2. Make the retry poll the actual missing dependency (route setters wired) rather than the boot-synchronous `openSummaryDetail`, and move the guarded body inside `tryOpen` so retries are covered by the guard.
3. Narrow `cleanUrl` to drop only `taskId`.
4. Add the overflow and cleanup-disposer test cases.

## 5. Additional notes (coverage / blind spots)

- Not verified live: no end-to-end click-through of a real notification link (static analysis only).
- Worker-side link emission (`SUMMARY_WEB_BASE_URL` on the summary worker) is out of scope for this web PR; this review assumes the link is produced correctly.
- Pre-existing repo-wide `tsc` errors and unrelated test failures were treated as baseline and excluded from findings.

Jerry-Xin (changes requested):
Re-reviewed at head `f3b996f03b84` after yujiawei's CR — byte-verified, and I'm converging with it. The in-scope landing logic is genuinely well-built, but two real user-facing gaps warrant REQUEST_CHANGES.

## ✅ In-scope logic is sound (stands from my prior review)
Injection-safe positive-integer `taskId` gate (`summaryDetailLanding.ts:37-39`), bounded 100ms×20 (~2s) retry reading the `openSummaryDetail` getter each tick (correct DI, no stale capture), timer cleanup on unmount, exact path match (trailing-slash tolerant), `replaceState` URL cleanup, SSR-guarded, strong 10-case test. No injection/XSS, no redirect loop.

## 🔴 Logged-out deep-link path (yujiawei) — adopted
`:32 if (!deps.isLoggedIn()) return undefined;` — a recipient who clicks the summary DM link **while not logged in** (a common path) gets a silent no-op; the `taskId` is not stashed, and the post-login redirect doesn't carry it, so after login the detail page never opens. The PR body scopes to "when logged in," but this repo already has the pattern to do better: `Layout/index.tsx:119-121` carries a **doc deep-link** through the login redirect. Summary deep-links should reuse that so the common logged-out-click path resolves post-login. Not just polish — it's the notification's primary click path.

## 🔴 Silent give-up after 2s (yujiawei's router-init concern) — adopted as robustness blocker
If `openSummaryDetail` isn't registered within ~2s (`SUMMARY_DETAIL_MAX_RETRIES=20`), `tryOpen` returns silently: no error surfaced, URL left uncleaned, user stranded on the default ChatPage with no recovery. Bounded/graceful (no crash/loop), but a first-open on a slow module-init/cold start can fail with no fallback. Suggest: on retry exhaustion, either surface a lightweight retry/toast or fall back to routing to the summary tab, rather than a silent dead-end.

## 🟡 Non-blocking
- Retry `if (++attempts >= SUMMARY_DETAIL_MAX_RETRIES) return;` after the `if (open)` — reachable, not strictly dead, but the exhaustion branch could be clearer (explicit "gave up" path tied to the fallback above).

The core parsing/retry/cleanup is correct and well-tested — but the logged-out path (primary click scenario) + the silent exhaustion give-up are real gaps. REQUEST_CHANGES, converging with yujiawei. Credit yujiawei for surfacing the common-path gap I under-weighted as out-of-scope.

apps/web/src/App/summaryDetailLanding.ts
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @returns cleanup 函数（停止仍在排队的重试）；返回 undefined 表示未启动任何重试。
+ */
+export function runSummaryDetailLanding(deps: SummaryDetailLandingDeps): (() => void) | undefined {
+  try {
+    const pathname = (deps.getPathname() || '').replace(/\/+$/, '') || '/';
+    if (pathname !== SUMMARY_DETAIL_PATH) return undefined;
+
+    // 未登录不动作：登录前落地会被登录流程打断，且详情页依赖登录态。
+    if (!deps.isLoggedIn()) return undefined;
```
OctoBoooot, line 32: **[major]** Logged-out deep-link is dropped across login. This bails on `!isLoggedIn()`, and the post-login redirect in `Layout/index.tsx:126-134` forwards only `sid`/`doc`/`space`/`folder` — `taskId` is not carried. A user clicking a completed-summary DM notification while logged out (or session-expired) logs in and lands on the app root, not the detail page — the exact case #150 fixes, still broken for that path. The repo already preserves doc deep-links through login (`Layout/index.tsx:118-134`, #511 problem 2); summary links should hook into the same path. Since the trigger is an async notification, logged-out click is a core case, not a fringe.

apps/web/src/App/summaryDetailLanding.ts
```diff
@@ -0,0 +1,61 @@
...
+    let attempts = 0;
+    const tryOpen = () => {
+      const open = deps.getOpenSummaryDetail();
+      if (open) {
+        open(taskId);
+        // 清 URL 回干净路径：避免用户刷新时重复落地（openSummaryDetail 已切到 summary 主 Tab）。
+        deps.cleanUrl?.(taskId);
+        return;
+      }
+      if (++attempts >= SUMMARY_DETAIL_MAX_RETRIES) return;
```
OctoBoooot, line 49: **[minor]** Silent give-up after the retry cap. After 20×100ms this returns with no URL cleanup and no user feedback — the user stays on ChatPage with a stale `/summary/detail?taskId=` URL and no indication anything was attempted. Bounded and safe (no crash/loop), but no recovery. Consider a visible fallback or at least clearing the URL. (Note: the retry itself is correct — it polls `openSummaryDetail` registration, which is the right readiness signal; the gap is the silent failure, not a missing check.)

OctoBoooot (changes requested):
## Correction to my #680 review: flipping APPROVE → REQUEST_CHANGES on a verified defect I under-graded

I approved this at `f3b996f03b84`. After byte-verifying a fact I'd missed — that this repo **already** carries deep-links across login — I'm correcting my verdict. This is a real defect, not a scope boundary. Credit to yujiawei (raised it) and Jerry-Xin (pointed me at the precedent).

### Why I changed my mind (the evidence)
`apps/web/src/Layout/index.tsx` `onLogin` (`:118-134`) deliberately forwards doc deep-links (`doc`/`space`/`folder`) through the post-login redirect so a first-time recipient who logs in still lands on the shared document (`#511 problem 2`), plus standalone `/d/:docId` return/recovery (`:137-158`, with open-redirect guards). So "a deep-link should survive login" is **established, intentional behavior here**. My APPROVE rested on treating the logged-out no-op as a documented scope boundary — that basis is wrong given the precedent.

### Major — logged-out deep-link is dropped across login (inconsistent with the doc-link pattern)
`summaryDetailLanding.ts:32` bails on `!isLoggedIn()`, and the login redirect (`Layout/index.tsx:126-134`) forwards only `sid`/`doc`/`space`/`folder` — `taskId` is not carried. So a user who clicks a completed-summary DM notification while logged out (or with an expired session) logs in and lands on the app root, not the detail page — the exact "点了跳不过去" that #150 sets out to fix, unfixed for that path. Because the trigger is an async notification, logged-out/expired click is a core case, not a fringe. Suggested direction: reuse the existing `onLogin` deep-link-forwarding path (carry `taskId` through the redirect) so summary links behave like doc links.

### Minor — silent give-up after the retry cap
`summaryDetailLanding.ts:49` returns silently after 20×100ms with no URL cleanup and no user feedback — the user is left on ChatPage with a stale `/summary/detail?taskId=` URL and no indication anything was attempted. Bounded and safe (no crash/loop), but no recovery. Worth a visible fallback or at least clearing the URL.

### What I do NOT adopt (over-graded in the CRs)
- **"Retry doesn't detect route initialization"** — the retry polls `getOpenSummaryDetail()`, and that registration *is* the readiness signal (dmworksummary registers it on module init); tests at `summaryDetailLanding.test.tsx:81/103` cover retry-until-ready and retry-cap-no-throw. The concern that's real is the *silent* give-up (above), not a missing readiness check.
- **Retry "dead code"** — the `if (++attempts >= MAX) return` branch is reachable; not dead. Nit at most.

### What's genuinely good (unchanged from my prior read)
The in-scope logged-in landing logic is correct and well-tested: canonical positive-integer `taskId` validation (rejects 0/negative/float/hex/unsafe), trailing-slash tolerance, SSR guard, URL cleanup preserving the hash, and the DI extraction mirroring `friendApplyReddotCleanup.ts` for testability. 10 solid unit cases.

### Net
The core landing logic is sound, but the logged-out path — the common case for a notification-triggered deep-link — is dropped despite an established same-repo pattern for preserving it. That's a real major, so REQUEST_CHANGES is the honest verdict; my earlier APPROVE under-graded it. Fixing #1 (carry `taskId` through login like doc links do) + a visible fallback for #2 should close it.

PR #681, opened 2026-07-13
fix(loop): backfill outstanding request-changes blockers from merged loop PRs

superalsrk:
## Summary

Several loop PRs were merged into `feat/octo-loop` while their reviews were still `CHANGES_REQUESTED`. Most of those blockers were later fixed by follow-up PRs (e.g. #654, #679, and #668's own follow-up commits). This PR backfills the ones that were **still live** in the current `feat/octo-loop` tip, newest → oldest.

| PR | Live blocker fixed |
|----|--------------------|
| #668 | `ThreadReply` submitted on bare **Enter** with no IME guard → half-composed CJK replies. Guard with `!e.nativeEvent.isComposing`. `replyToThread` also returned `true` unconditionally, silently dropping failed attachments — now returns the failed-file list so the reply box re-stages them (mirrors `submitComment`). |
| #610 | `RunDetailModal` poll stopped on the terminal tick without draining the final messages written during completion. Added one last incremental `listRunMessages` drain before stopping. |
| #619 | `PersonalPage` shares `@octo/loop`'s module-level workspace global with Loop. It now re-asserts `setWorkspaceContext` on tab-open and on `wk:nav-menu-activated`, and the bootstrap effect is mount-once (`[]` not `[t]`) — fixes cross-workspace data contamination, the blank right pane after returning from another menu, and the view reset on locale switch. |
| #597 | CLI auth callback allowlist accepted the whole private-IPv4 range → LAN token exfil. Restricted to **loopback only**, fixed the bracketed `::1` IPv6 case, extracted the validator to `ui/cliCallback.ts`, and added a vitest suite locking accept/reject behavior. |
| #581 | Member-picker `Select.Option`s had no string `label`, so search filtered to empty on the first keystroke — added `label={c.name}`. Roster load-failure now shows a distinct error + retry instead of a false "no members". Member add/role-change/remove now call `invalidateDirectory()` so assignee pickers reflect membership changes. |

Already-fixed-by-follow-ups (verified, not re-touched): #663 once-cron & my-loops scope fallback, #657/#640 skill name/description round-trip, #605 preview fail-open/race/status-gate, #619 dev-host, #643 comment-attachment retry, #668 blockers 1–3 + filter popovers.

## Test plan

- [x] `vitest run` — new `cliCallback.test.ts` (8 tests) passes: loopback accept, LAN/public/userinfo/scheme/malformed reject, bracketed `::1`, RFC1918 boundaries.
- [x] `tsc --noEmit` clean for all changed files (remaining errors are pre-existing Semi/node_modules noise, not in changed files).
- [ ] Manual: IME reply (Pinyin Enter confirm doesn't post), run modal shows final result on completion, Personal↔Loop workspace switching, member search + roster-error retry.

### DIFF
--- a/packages/dmloop/package.json
+++ b/packages/dmloop/package.json
@@ -15,6 +15,11 @@
   "devDependencies": {
     "@types/react": "^18.3.28",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react-dom": "^17.0.2",
+    "vitest": "^4.1.5"
+  },
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest"
   }
 }
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -638,6 +638,8 @@
     "addMember": "Add member",
     "selectMember": "Search and select a space member",
     "noCandidates": "No members available to add in this space",
+    "rosterLoadFailed": "Failed to load members",
+    "rosterRetry": "Retry",
     "added": "Member added",
     "pendingInvites": "Pending invitations",
     "revoke": "Revoke"
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -638,6 +638,8 @@
     "addMember": "添加成员",
     "selectMember": "搜索并选择 space 成员",
     "noCandidates": "该 space 暂无可添加的成员",
+    "rosterLoadFailed": "成员列表加载失败",
+    "rosterRetry": "重试",
     "added": "成员已添加",
     "pendingInvites": "待处理邀请",
     "revoke": "撤销"
--- a/packages/dmloop/src/pages/LoopCliAuthorizePage.tsx
+++ b/packages/dmloop/src/pages/LoopCliAuthorizePage.tsx
@@ -7,38 +7,10 @@ import { issueLoopCliToken } from "../api/authApi";
 import { listWorkspaces } from "../api/workspaceApi";
 import { currentWorkspaceId, setWorkspaceContext } from "../api/http";
 import { clearPendingLoopCliAuthorizeSearch } from "../cliAuthorizeSession";
+import { isSafeCliCallback } from "../ui/cliCallback";
 import "./loop.css";
 
 const { Text, Title } = Typography;
-
-function isPrivateIPv4(host: string): boolean {
-  const parts = host.split(".").map((p) => Number(p));
-  if (
-    parts.length !== 4 ||
-    parts.some((p) => !Number.isInteger(p) || p < 0 || p > 255)
-  ) {
-    return false;
-  }
-  const [a, b] = parts;
-  return (
-    a === 10 ||
-    a === 127 ||
-    (a === 172 && b >= 16 && b <= 31) ||
-    (a === 192 && b === 168)
-  );
-}
-
-function isSafeCliCallback(raw: string): boolean {
-  try {
-    const parsed = new URL(raw);
-    if (parsed.protocol !== "http:" && parsed.protocol !== "https:")
-      return false;
-    const host = parsed.hostname.toLowerCase();
-    return host === "localhost" || host === "::1" || isPrivateIPv4(host);
-  } catch {
-    return false;
-  }
-}
 
 function redirectToCli(
   callbackURL: string,
--- a/packages/dmloop/src/pages/SettingsPage.tsx
+++ b/packages/dmloop/src/pages/SettingsPage.tsx
@@ -10,6 +10,7 @@ import {
   updateWorkspace, listWorkspaceMembers, addOctoMember, updateMemberRole, removeMember,
   listWorkspaceInvitations, revokeInvitation,
 } from "../api/workspaceApi";
+import { invalidateDirectory } from "../api/directory";
 
 const { Title, Text } = Typography;
 const ROLES = ["admin", "member"];
@@ -108,6 +109,9 @@ function MembersTab({ workspaceId }: { workspaceId: string }) {
   const [invites, setInvites] = useState<Invitation[]>([]);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<string | null>(null);
+  // roster(space 成员)单独的失败标记:与「space 确实没有可加成员」区分开——
+  // 抓取失败时不能显示误导性的「暂无成员」并禁用添加(#581 评审 B2)。
+  const [rosterError, setRosterError] = useState(false);
   // octo space 成员（仅人，robot===0）+ uid→身份 映射，供选择器候选与列表渲染复用。
   const [spaceHumans, setSpaceHumans] = useState<SpaceMember[]>([]);
   const seqRef = useRef(0); // reload 请求序号:切 workspace 时只让最新一次落地
@@ -129,13 +133,17 @@ function MembersTab({ workspaceId }: { workspaceId: string }) {
 
   const reload = useCallback(() => {
     const my = ++seqRef.current; // 切 workspace 时旧 in-flight 结果不得覆盖新的
-    setLoading(true); setError(null);
+    setLoading(true); setError(null); setRosterError(false);
+    let rosterFailed = false;
     Promise.all([
       listWorkspaceMembers(workspaceId),
       listWorkspaceInvitations(workspaceId).catch(() => [] as Invitation[]),
-      loadSpaceMembers().catch(() => [] as SpaceMember[]),
+      loadSpaceMembers().catch(() => { rosterFailed = true; return [] as SpaceMember[]; }),
     ])
-      .then(([m, inv, humans]) => { if (my !== seqRef.current) return; setMembers(m); setInvites(inv); setSpaceHumans(humans); })
+      .then(([m, inv, humans]) => {
+        if (my !== seqRef.current) return;
+        setMembers(m); setInvites(inv); setSpaceHumans(humans); setRosterError(rosterFailed);
+      })
       .catch((e) => { if (my === seqRef.current) setError(e?.message ?? "load failed"); })
       .finally(() => { if (my === seqRef.current) setLoading(false); });
   }, [workspaceId, loadSpaceMembers]);
@@ -157,6 +165,7 @@ function MembersTab({ workspaceId }: { workspaceId: string }) {
     try {
       await addOctoMember(workspaceId, { octo_uid: selectedUid, role });
       setSelectedUid("");
+      invalidateDirectory(); // 成员变动后清共享目录缓存,否则指派选择器仍返回旧候选(#581 评审)
       Toast.success(t("loop.settings.added"));
       reload();
     } catch (e) { Toast.error((e as Error)?.message ?? "add failed"); }
@@ -164,7 +173,7 @@ function MembersTab({ workspaceId }: { workspaceId: string }) {
   };
 
   const changeRole = async (m: WorkspaceMember, r: string) => {
-    try { await updateMemberRole(workspaceId, m.id, r); Toast.success(t("loop.toast.saved")); reload(); }
+    try { await updateMemberRole(workspaceId, m.id, r); invalidateDirectory(); Toast.success(t("loop.toast.saved")); reload(); }
     catch (e) { Toast.error((e as Error)?.message ?? "failed"); }
   };
   const remove = (m: WorkspaceMember) => {
@@ -174,7 +183,7 @@ function MembersTab({ workspaceId }: { workspaceId: string }) {
       okText: t("loop.action.delete"),
       cancelText: t("loop.action.cancel"),
       onOk: async () => {
-        try { await removeMember(workspaceId, m.id); Toast.success(t("loop.toast.deleted")); reload(); }
+        try { await removeMember(workspaceId, m.id); invalidateDirectory(); Toast.success(t("loop.toast.deleted")); reload(); }
         catch (e) { Toast.error((e as Error)?.message ?? "failed"); }
       },
     });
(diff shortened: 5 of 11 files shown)


Jerry-Xin (approved):
Reviewed at head `1cefd6765cad`, byte-verified. Backfill of request-changes blockers across merged loop PRs — every fix is correct & complete, and the security-critical CLI callback validator fails closed. **APPROVE**.

## ✅ Security — `cliCallback.ts` (the token-delivery trust boundary) is sound
`isSafeCliCallback` (`ui/cliCallback.ts:22-30`) is the sole gate for where the CLI token lands (`LoopCliAuthorizePage.tsx:23` sets `?token=…` then `window.location.href=`). Byte-verified (incl. via `new URL(...)` in node):
- **Loopback-only** (`isLoopbackHost:7-20`): `localhost`, `::1`, `::ffff:127.0.0.1`, `127.0.0.0/8` (octet-validated), IPv6 brackets stripped. LAN/private (10/172.16-31/192.168) now **rejected** — the #597 token-exfil fix.
- **Userinfo open-redirect defeated:** `http://127.0.0.1@evil.com/cb` → `URL.hostname` = `evil.com` → rejected. Verified.
- **Scheme-gated:** non-http(s) (`javascript:`/`file:`/`data:`) rejected before host check.
- Token never logged / never written to DOM — only placed in the query of a validated loopback URL. Validated twice (mount effect + `authorize()` guard) — defense in depth.

## ✅ Each backfilled fix — correct & complete
- **#597** CLI allowlist → loopback-only: extracted, tested, IPv6 handled. ✓
- **#668** IME guard `IssueDetailPage.tsx:134` (`!e.nativeEvent.isComposing`) + failed-attachment retention (`Promise<File[]|null>` contract: null=keep whole draft, array=re-stage failed uploads; propagated to `ThreadReply` prop `:105`, `reqRef` token no stale write). ✓
- **#610** `RunDetailModal.tsx:183` terminal-tick final drain — captures `cancelled` before/after await, dedups by seq, unmount-safe. ✓
- **#619** `PersonalPage.tsx` workspace re-assert via mount-once refs + `wk:nav-menu-activated` (guarded), restores actual current tab; listener cleaned up. ✓
- **#581** `SettingsPage.tsx` member picker — `Select.Option` string `label`, separate `rosterError` state (seqRef-guarded) + Banner/retry, `invalidateDirectory()` on add/role/remove. ✓

## ✅ i18n / deps / tests
- Locale parity CLEAN — `rosterLoadFailed`/`rosterRetry`/`invalidCallback`/`commentAttachFailed` present in both en-US + zh-CN; no new hardcoded user-facing string.
- `vitest ^4.1.5` added to `dmloop` **devDependencies** + scripts; `pnpm-lock.yaml` consistent (no drift); used by the new test.
- `cliCallback.test.ts` — 8 genuine security assertions (loopback accept, LAN/public reject, userinfo redirect defeat, scheme reject, RFC1918 boundaries). Real regression-lock, not happy-path.

## 🟡 Non-blocking nits
- `cliCallback.test.ts:48` asserts `isLoopbackHost("[::ffff:127.0.0.1]")===true`, but `new URL("http://[::ffff:127.0.0.1]/cb").hostname` normalizes to `[::ffff:7f00:1]` which the fn does NOT match — so the real path would reject that form. Errs safe (fails closed); cosmetic — normalize `::ffff:7f00:1` too and assert via `isSafeCliCallback`.
- `updateMemberRole` invalidation could be centralized in the API layer like `addOctoMember`/`removeMember` — cosmetic.

No TS error, dead code, unhandled promise, or state-after-unmount across the 11 files. APPROVE.

OctoBoooot (approved):
## Review: backfill outstanding request-changes blockers from merged loop PRs (#681)

**Verdict:** Approve. All five backfilled fixes are byte-verified against `1cefd6765cad` and each correctly resolves the live blocker it claims. No blockers, no majors.

CI: `code-review` is still **PENDING** (`mergeStateStatus=UNSTABLE`, `mergeable=MERGEABLE`) — no code CI is red, but the PR is not mergeable until `code-review` completes. `needs-human-review` applied (11 files > size gate) — a maintainer must clear it before merge.

Base `feat/octo-loop`. Risk tier: normal.

### Verified fixes (each resolves a standing CR)
- **#597 — CLI-callback token-exfil closed (security).** `ui/cliCallback.ts` extracts a pure, testable validator: `isSafeCliCallback` restricts the callback host to loopback only — `localhost`, `::1`, `::ffff:127.0.0.1`, full `127.0.0.0/8` — and rejects the entire LAN/RFC1918 range (10/172.16-31/192.168) that previously let a signed CLI token be redirected to a same-segment host. Scheme allowlist (`http:`/`https:` only) rejects `javascript:`/`file:`/`data:`. Bracketed IPv6 (`[::1]`) is stripped before compare. This closes the private-IP SSRF/exfil vector that was open in base.
- **#668 — ThreadReply IME guard (resolves my standing #668 CR).** `IssueDetailPage.tsx:134` — `onKeyDown` now gates on `!e.nativeEvent.isComposing`, so a mid-composition Enter no longer submits a half-typed CJK/IME reply. `replyToThread` also returns the failed-file list and re-arms `pendingFiles` on partial attach failure.
- **#619 — workspace-context pollution (matches the exact shared-global hazard).** `dmpersonal/PersonalPage.tsx` re-asserts its workspace on tab-open (`:41`) and on `wk:nav-menu-activated` (`:108`), and the mount effect is correctly `[]`-scoped (not `[t]`) so a locale switch doesn't flicker/reset the tab. The design is symmetric: `LoopPage.tsx:143-149` re-asserts its own workspace on the same activation event, so neither page's context leaks into the other's `@octo/loop` API calls after nav. PersonalPage is persistent, so re-assert (not restore-on-unmount) is the correct pattern here.
- **#610 — final run-message drain.** `RunDetailModal.tsx:183-185` — on the poll iteration that detects a terminal status, one final incremental `listRunMessages(run.id, lastSeqRef.current)` runs before the loop stops, capturing the closing result/error batch that lands after the message fetch returns but before the status flips. `MAX_MISSES=3` tolerates transient empty `listRuns`.
- **#581 — member roster freshness.** `SettingsPage.tsx` calls `invalidateDirectory()` after every member mutation (add `:168`, role change `:176`, remove `:186`) so the assignee picker stops returning stale candidates; `rosterError` state (`:114/:237`) surfaces a retry path on roster load failure. `IssueDetailPage.tsx:381` labels candidates by `c.name` (bracket-stripped).

### Praise
`ui/__tests__/cliCallback.test.ts` (8 cases) is well-constructed security regression coverage — it locks the userinfo open-redirect bypass (`http://127.0.0.1@evil.com` → rejected, since `URL.hostname` resolves to `evil.com`), the RFC1918 boundaries (`172.15`/`172.32`/`128.0.0.1` all rejected), and the bracketed-IPv6 loopback case. Extracting the validator to a React/Semi-free module to make it unit-testable is exactly right for a security trust boundary.

### Note
This PR backfills the #668 fix into `feat/octo-loop`; my open CR on PR #668 itself is anchored to that PR's head and will need the author/maintainer to reconcile it — the fix here is the one I prescribed.

yujiawei (approved):
# Code Review — PR #681 (octo-web)

**Verdict: APPROVE.** No P0/P1 blockers. This PR cleanly backfills the outstanding request-changes blockers from the merged loop PRs, and the security-sensitive piece — the CLI-auth callback allowlist — is now correct. Below are the per-area findings; everything flagged is P2 (non-blocking polish), no changes required to merge.

## 1. Spec compliance

The diff implements exactly the six blockers listed in the PR description, and nothing beyond scope. Each maps to a real, verifiable fix:

- **#597 — CLI callback allowlist → loopback-only.** Extracted to `packages/dmloop/src/ui/cliCallback.ts` with a locking vitest suite. ✅
- **#668 — IME reply guard + failed-attachment re-staging** in `IssueDetailPage.tsx`. ✅
- **#610 — RunDetailModal final message drain** before poll stop. ✅
- **#619 — PersonalPage workspace-context re-assertion** + mount-once bootstrap. ✅
- **#581 — member-picker `label`, roster-failure banner, `invalidateDirectory()`** in `SettingsPage.tsx`. ✅

No over-build, no scope drift.

## 2. Security — CLI callback validator (the load-bearing change)

Because the freshly-minted CLI token is delivered via a **client-side** `window.location.href` redirect, `isSafeCliCallback` **is** the credential-delivery trust boundary. I extracted the validator and ran a bypass table through the WHATWG URL parser (Node == browser family):

- LAN / RFC1918 widening (`10/8`, `172.16–31`, `192.168/16`) — now **rejected** ✅ (this was the #597 P1)
- `[::1]` IPv6 loopback — now **accepted** ✅ (the bracket bug from #597 is fixed)
- userinfo trick (`http://127.0.0.1@evil.com`, `http://localhost@evil.com`) → hostname resolves to `evil.com` → **rejected** ✅ (code correctly uses `URL.hostname`, not the raw string)
- decimal / hex / short-form IPs (`2130706433`, `0x7f000001`, `127.1`) normalize to `127.0.0.1` (loopback, safe) ✅
- non-`http(s)` schemes (`javascript:`, `file:`, `data:`) — **rejected** ✅
- `0.0.0.0`, public IPs, `localhost.evil.com`, malformed — **rejected** ✅

The network/LAN token-exfil vector is genuinely closed. `redirectToCli` assembles the token with `URLSearchParams.set` (no raw string concat, no fragment injection), and the validated URL is the exact string handed to `window.location.href` (no re-parse between check and use), so there is no TOCTOU gap on the client.

**P2 (test integrity, fails closed):** `isLoopbackHost` has a `bare === "::ffff:127.0.0.1"` branch that is **dead in the real path** — `new URL("http://[::ffff:127.0.0.1]/").hostname` normalizes to `[::ffff:7f00:1]`, which never matches that literal. The unit test at `cliCallback.test.ts:48` only passes because it calls `isLoopbackHost` with the raw bracketed string directly, bypassing `new URL`. Net effect is a **harmless over-rejection** of v4-mapped loopback (no attacker gain), but the test gives false confidence in a branch production can't reach. Consider either matching the normalized form (`::ffff:7f00:1`) or dropping the branch + its assertion.

**P2 (hardening, requires an on-box adversary — out of this PR's scope):** loopback-only closes remote/LAN exfil, but a co-resident local process can bind any `127.0.0.0/8` port and receive a live token; the consent panel shows only workspace name/slug, not the callback host:port. Surfacing the destination host in the consent UI would be a reasonable follow-up. Also note the existing `TODO(security follow-up)` to replace the JWT-in-query handoff with a PKCE / device-authorization flow — the right long-term fix.

## 3. Correctness — reply contract, IME guard, run-drain

- **`replyToThread` return-contract change (`boolean` → `File[] | null`)** is consumed correctly by the single caller `ThreadReply.submit`: `null` → keep draft + all files (comment not created); `File[]` → clear draft, re-stage only the failed attachments (empty = full clear). This now mirrors `submitComment`, so failed attachments are no longer silently dropped. ✅
- **IME guard** `e.key === "Enter" && !e.nativeEvent.isComposing` — plain Enter still submits, IME-composition Enter is suppressed. No regression to non-IME Enter. ✅
- **RunDetailModal final drain** — the `else if (!cancelled)` runs exactly once when the poll loop stops, never re-arms (chain terminates), fetches only `seq > lastSeqRef` (advanced synchronously in `apply`), dedupes via the `seen` set, and is guarded by `cancelled` before every setState. ✅

**P2 (residual, inherent to WS-less polling):** the single-shot drain closes the "written-after-fetch, before-status-flip" window, but rows persisted server-side *after* the drain request still require a reopen to appear. Acceptable given there is no WebSocket channel.

**P2 (cosmetic):** the remove-file chip button is not `disabled={busy}` (unlike the file input and Send button). If a user removes a queued attachment mid-flight and that file's upload happens to fail, `setFiles(res)` on return could resurrect it. Trivial edge case.

## 4. State management — PersonalPage workspace context

Refs (`tabRef`, `selectedWsRef`) are reassigned every render / settle in the same commit, so the `[]`-dep effects read them live — no stale-closure bug. The `wk:nav-menu-activated` listener guards on `menuId === "dmpersonal"` (matches `WKApp.menus.register("dmpersonal", …)`) and is cleaned up on unmount. Cross-workspace contamination on the main paths is resolved. ✅

**P2:** the `[t] → []` dependency change makes the bootstrap effect mount-once (correct, aligns with LoopPage), but the imperatively-pushed loading/empty/error right-panel views now capture `t` from first render and won't re-translate on a language switch (the reactive sidebar still updates → mixed-locale UI on the terminal panels only). Cosmetic, language-switch-only.

**P2:** unlike LoopPage's `onNavMenuActivated` (which re-runs `findWs` to confirm the workspace still exists), PersonalPage re-asserts `selectedWsRef.current` blindly. If the workspace was deleted/revoked while on another page, re-clicking Personal re-points context at a dead workspace → sub-page 401/404. Edge case.

## 5. Member picker / directory cache

`Select.Option label={c.name}` fixes the search-to-empty regression (Semi's filter needs a string label). The new `rosterError` flag is set inside the `seqRef` stale-guarded `.then` (reset at reload start, fresh per-invocation `let`), so a superseded reload can't land a stale error, and the failure Banner+retry is correctly distinguished from a genuinely empty space (`candidates = []` is committed atomically with `rosterError = true`, so "banner + populated candidates" is unreachable). `invalidateDirectory()` is imported, exists (`directory.ts:88`), and is called on all three mutation paths (add / role-change / remove). ✅

**P2:** `loadSpaceMembers` failure is swallowed into `rosterFailed` with no `console.error` — the Banner is the intended UX surface, but a log line would aid debuggability.

## Coverage note (what I could not verify from the diff alone)

- The client-side gate is necessary but not sufficient: the **token-issuance endpoint should not independently trust `cli_callback`** server-side. Worth a human confirming the backend does not echo/act on the callback.
- `renderTab`'s `RuntimePage`/`SkillPage` self-subscribe to i18n, so they are not affected by the mount-once freeze — assumed, not re-verified end-to-end.
- Manual QA items in the PR's test plan (IME Pinyin confirm, run-modal completion, Personal↔Loop switching, roster-error retry) are unchecked; the static analysis supports them but a manual pass is the final confirmation.

Adversarial cross-family advisor legs were attempted but both were unavailable this cycle (one gateway error, one timeout), so this verdict rests on multi-lens adversarial static analysis with an empirically-executed URL-parser bypass table. Given the security-sensitive classification, a human sign-off on the two backend/consent items above is worthwhile, but none of them block this merge.

PR #683, opened 2026-07-13
fix(loop): auto-refresh board after quick-create; signal vanished runs

MrAladdin:
Auto-refresh the loop board after an async quick-create dispatch, plus a vanished-run signal. Verified by real-env e2e on both create entry points + two-model adversarial review + /simplify.

## 1. Board didn't auto-refresh after quick-create dispatch
Quick-create is async: `POST /issues/quick-create` returns only `{task_id}`; the agent creates the issue moments later. The board's single post-dispatch reload fired before the issue existed → new loop only appeared on manual reload. (Upstream uses a WS event stream to invalidate the list; dmloop has no realtime layer yet — deferred, see below.)

**Unified event-bus stopgap** (covers BOTH create entry points):
- `NewLoopPage.submit()` emits `wk:loop-issues-dispatched` on quick-create success.
- The **persistent** `LoopPage` (never remounts across tabs/entries) owns bounded settle timers and re-emits `wk:loop-issues-refresh` at `SETTLE_DELAYS_MS = [2000,5000,9000,14000]`.
- The mounted `IssuePage` subscribes → reloads through `onMutatedRef` (latest filters — a `seq` guard only orders responses, not stale inputs).
- 2 events added to `MittEvents` (`@octo/base`, additive).

Why this shape (from /simplify altitude review): reload-on-activation was rejected — `WKViewQueue.replaceToRoot` keys the board by `issue:${wsId}`, so re-activating the same tab reuses the fiber (no reload); forcing it needs a key-bump that drops filter/scroll state and still re-races the async create. Timed settle owned by the persistent parent is the right stopgap; the real fix is WS push (deferred). This also fixed the incomplete first cut, which only covered the in-board `+` entry and left the more prominent **sidebar** "新建回路" stale.

## 2. RunDetailModal silently froze on a vanished run (P2)
When a run disappears from `listRuns` for `MAX_MISSES` consecutive polls, the modal stopped polling but froze on the last (possibly "running") status. Added a `gone` state, set **only** on the miss-exhausted path (never on a normal terminal stop), rendering a "运行记录已不存在" badge. i18n zh+en.

## Blast radius (Rule 8)
- `packages/dmworkbase/src/App.tsx` `MittEvents`: **additive** (2 new event keys) — existing consumers reference neither key, zero breakage.
- `IssuePage` / `LoopPage` / `NewLoopPage` / `RunDetailModal`: internal changes; no exported type/prop/hook signature changes. `settleReload` reuses existing `onMutated`/`reload`.

## Verification (real-env e2e, Alice's bridged workspace)
- **In-board `+` entry**: dispatched → board auto-showed ALI-2 within ~11s, no manual reload.
- **Sidebar "新建回路" entry** (the gap /simplify caught): dispatched → board auto-showed ALI-3 within ~12s, no manual reload — confirming the unified mechanism covers the remount path.
- New issues confirmed in DB (agent async-created them).
- Codex adversarial-review: approve. Claude code-reviewer: clean (traced WKViewQueue reconcile + mitt types: both entries covered, 2s margin >> sync remount, listener/timer cleanup complete, no stale-closure, multi-instance unreachable). /simplify (reuse/efficiency/altitude): clean. tsc clean.

## Deferred follow-ups (tracked)
- The class fix: dmloop **WS realtime** (loop proxy WS passthrough + WS client + event→invalidate), matching upstream — subsumes this stopgap.
- Focused unit tests for the polling/race guards (dmloop now has a vitest harness as of #681; tracked separately).

### DIFF
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -606,6 +606,7 @@
     "cancelled": "Stopped",
     "copyAll": "Copy all",
     "copied": "Copied",
+    "gone": "Run no longer exists",
     "events": "{{count}} events",
     "toolCalls": "{{count}} tool calls",
     "duration": "Duration",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -606,6 +606,7 @@
     "cancelled": "已终止",
     "copyAll": "复制全部",
     "copied": "已复制",
+    "gone": "运行记录已不存在",
     "events": "{{count}} 个事件",
     "toolCalls": "{{count}} 次工具调用",
     "duration": "时长",
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -201,15 +201,19 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
 
   // 派单(quick-create)是异步的:agent 稍后才建 issue,单次刷新太早看不到(dmloop 无 WS 推送,
   // 见记忆 dmloop-no-realtime-defer-ws)。这里有界补刷几次直到新回路出现。
-  // ponytail: 真正的修法是 WS 实时推送(已记后期做),此为 stopgap;派单低频,几次冗余刷新可接受。
+  // ponytail: 真正的修法是 WS 实时推送(已记后期做),此为 stopgap;派单低频,几次补刷可接受。
   // 手动建单是同步的,立即那次即命中,后续定时刷新只是无害冗余。
+  // 定时回调必须走 ref 读最新 onMutated —— 否则闭包捕获派单时刻的筛选/视图,窗口内用户改了筛选后
+  // 延迟刷新会用旧筛选参数覆盖当前视图(reload 的 seq 守卫只防乱序、不防陈旧入参)。
   const settleTimersRef = useRef<number[]>([]);
+  const onMutatedRef = useRef(onMutated);
+  useEffect(() => { onMutatedRef.current = onMutated; }, [onMutated]);
   useEffect(() => () => settleTimersRef.current.forEach(clearTimeout), []);
   const settleReload = useCallback(() => {
     settleTimersRef.current.forEach(clearTimeout);
-    onMutated(); // 立即刷一次
-    settleTimersRef.current = [2000, 5000, 9000, 14000].map((d) => window.setTimeout(onMutated, d));
-  }, [onMutated]);
+    onMutatedRef.current(); // 立即刷一次
+    settleTimersRef.current = [2000, 5000, 9000, 14000].map((d) => window.setTimeout(() => onMutatedRef.current(), d));
+  }, []);
 
   // 改任一筛选/搜索都回到第一页，避免停在越界的 offset（此规则只此一处表达）。
   const update = (p: Partial<Filters>) => { setF((prev) => ({ ...prev, ...p })); setPage(0); };
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -199,6 +199,18 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   // 变更后刷新:既重取列表,又刷新运行中快照(指派/状态变更可能起/停 agent run)。
   const onMutated = useCallback(() => { reload(); refreshRunning(); }, [reload, refreshRunning]);
 
+  // 派单(quick-create)是异步的:agent 稍后才建 issue,单次刷新太早看不到(dmloop 无 WS 推送,
+  // 见记忆 dmloop-no-realtime-defer-ws)。这里有界补刷几次直到新回路出现。
+  // ponytail: 真正的修法是 WS 实时推送(已记后期做),此为 stopgap;派单低频,几次冗余刷新可接受。
+  // 手动建单是同步的,立即那次即命中,后续定时刷新只是无害冗余。
+  const settleTimersRef = useRef<number[]>([]);
+  useEffect(() => () => settleTimersRef.current.forEach(clearTimeout), []);
+  const settleReload = useCallback(() => {
+    settleTimersRef.current.forEach(clearTimeout);
+    onMutated(); // 立即刷一次
+    settleTimersRef.current = [2000, 5000, 9000, 14000].map((d) => window.setTimeout(onMutated, d));
+  }, [onMutated]);
+
   // 改任一筛选/搜索都回到第一页，避免停在越界的 offset（此规则只此一处表达）。
   const update = (p: Partial<Filters>) => { setF((prev) => ({ ...prev, ...p })); setPage(0); };
   const switchView = (v: ViewMode) => { setView(v); setPage(0); };
@@ -214,7 +226,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   const openNewLoop = () => {
     WKApp.routeRight.push(
       <NewLoopPage
-        onCreated={() => { onMutated(); WKApp.routeRight.pop(); }}
+        onCreated={() => { settleReload(); WKApp.routeRight.pop(); }}
       />,
     );
   };
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -199,20 +199,17 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   // 变更后刷新:既重取列表,又刷新运行中快照(指派/状态变更可能起/停 agent run)。
   const onMutated = useCallback(() => { reload(); refreshRunning(); }, [reload, refreshRunning]);
 
-  // 派单(quick-create)是异步的:agent 稍后才建 issue,单次刷新太早看不到(dmloop 无 WS 推送,
-  // 见记忆 dmloop-no-realtime-defer-ws)。这里有界补刷几次直到新回路出现。
+  // 派单(quick-create)是异步的:agent 稍后才建 issue(dmloop 无 WS 推送,见记忆 dmloop-no-realtime-defer-ws)。
+  // NewLoopPage 派单成功发 `wk:loop-issues-dispatched`,常驻的 LoopPage 据此有界补发 `wk:loop-issues-refresh`,
+  // 看板收到即重取——一套机制覆盖"看板内新建"与"侧栏新建"两个入口(定时器归 LoopPage,见那里)。
+  // 走 ref 读最新 onMutated:延迟刷新须用当前筛选/视图,否则会用陈旧入参覆盖(reload 的 seq 守卫只防乱序)。
   // ponytail: 真正的修法是 WS 实时推送(已记后期做),此为 stopgap;派单低频,几次补刷可接受。
-  // 手动建单是同步的,立即那次即命中,后续定时刷新只是无害冗余。
-  // 定时回调必须走 ref 读最新 onMutated —— 否则闭包捕获派单时刻的筛选/视图,窗口内用户改了筛选后
-  // 延迟刷新会用旧筛选参数覆盖当前视图(reload 的 seq 守卫只防乱序、不防陈旧入参)。
-  const settleTimersRef = useRef<number[]>([]);
   const onMutatedRef = useRef(onMutated);
   useEffect(() => { onMutatedRef.current = onMutated; }, [onMutated]);
-  useEffect(() => () => settleTimersRef.current.forEach(clearTimeout), []);
-  const settleReload = useCallback(() => {
-    settleTimersRef.current.forEach(clearTimeout);
-    onMutatedRef.current(); // 立即刷一次
-    settleTimersRef.current = [2000, 5000, 9000, 14000].map((d) => window.setTimeout(() => onMutatedRef.current(), d));
+  useEffect(() => {
+    const onRefresh = () => onMutatedRef.current();
+    WKApp.mittBus.on("wk:loop-issues-refresh", onRefresh);
+    return () => WKApp.mittBus.off("wk:loop-issues-refresh", onRefresh);
   }, []);
 
   // 改任一筛选/搜索都回到第一页，避免停在越界的 offset（此规则只此一处表达）。
@@ -230,7 +227,7 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
   const openNewLoop = () => {
     WKApp.routeRight.push(
       <NewLoopPage
-        onCreated={() => { settleReload(); WKApp.routeRight.pop(); }}
+        onCreated={() => { onMutated(); WKApp.routeRight.pop(); }}
       />,
     );
   };
--- a/packages/dmloop/src/pages/LoopPage.tsx
+++ b/packages/dmloop/src/pages/LoopPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import { Typography, Dropdown, Avatar, Modal, Toast, Button } from "@douyinfe/semi-ui";
 import {
   ClipboardList, Briefcase, Bot, Users, Settings,
@@ -24,6 +24,10 @@ import "../ui/loopControls.css";
 const { Title, Text } = Typography;
 
 type TabKey = "myloop" | "issue" | "project" | "automation" | "agent" | "squad" | "settings";
+
+// 派单后看板补刷的退避时刻(ms):agent 异步建单的落库延迟不可观测,手调的退避窗口(非可推导状态)。
+const SETTLE_DELAYS_MS = [2000, 5000, 9000, 14000];
+
 
 // 顶部独立入口：我的回路（复用 Issue 视图的「与我相关」分组）。
 const MY_TAB: { key: TabKey; icon: React.ReactNode } = { key: "myloop", icon: <CircleUserRound size={16} /> };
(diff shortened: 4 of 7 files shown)


OctoBoooot (approved):
## Review: fix(loop): auto-refresh board after quick-create; signal vanished runs (#683)

**Verdict:** Approve. Both fixes are correct and well-reasoned, verified against merge-base `7345f995` → head `224f5ae1`. No blockers, no majors.

CI: `code-review` is **PENDING** (`mergeStateStatus=UNSTABLE`, `mergeable=MERGEABLE`) — no code CI red, but not mergeable until it completes. Base `feat/octo-loop`. Risk tier: normal.

### What's good
- **`IssuePage.settleReload` — the stale-closure guard is the right call.** dmloop has no WS push and quick-create is async (the agent creates the issue later), so a single post-create refresh fires too early; the bounded re-poll (`[0, 2000, 5000, 9000, 14000]ms`) is a reasonable stopgap. Critically, the deferred timers read `onMutatedRef.current` rather than the `onMutated` captured at dispatch time — so if the user changes filters/view during the 14s window, a delayed refresh won't clobber the current view with stale filter params. The comment correctly notes `reload`'s seq guard prevents out-of-order but not stale inputs. Prior timers are cleared before each reschedule (repeated quick-creates don't stack), and the `[]`-cleanup effect clears on unmount — no timer leak, no setState-after-unmount.
- **`RunDetailModal.gone` — clean terminal-vs-vanished distinction.** Reuses the existing `MAX_MISSES` counter to tell "run finished" (terminal, run still in list) from "run removed" (gone), and `gone` takes precedence over the stale `active` badge so a deleted run no longer freezes on a phantom "running" spinner. State reset on both `!visible` and effect re-entry.

### Verified
- `XCircle` is imported (`RunDetailModal.tsx:9`).
- `loop.run.gone` resolves: the i18n file nests under top-level `run` (the `loop.` namespace is applied at load), and the new `run.gone` is a sibling of the already-working `run.copied` in **both** en-US and zh-CN with proper translations.

Small, focused, correct. Nice handling of the deferred-closure trap.

Jerry-Xin (approved):
Reviewed at head `6e359a96`, byte-verified. Targeted stopgap for the #673 board-refresh gap + a vanished-run badge — both are honest, module-internal improvements on the `feat/octo-loop` branch. **APPROVE** (2 non-blocking notes).

## ✅ Auto-refresh after quick-create (`IssuePage.tsx:207-215/233`)
`settleReload` = immediate `onMutatedRef.current()` + bounded settle-poll at 2/5/9/14s (5 reloads over 14s), all `clearTimeout`'d on unmount. Routes through `onMutatedRef` (synced each render) so late timers use current filters — the stale-closure race is genuinely fixed. Doesn't interfere with the 15s `refreshRunning` (own `runSeq` guard) or `reload`'s seq guard (`:170`).
- 🟡 **Best-effort, not guaranteed:** if the agent creates the issue >14s after quick-create, the board still won't show it until the next manual reload/mutation. This is a strict improvement over the prior single-reload-that-always-missed, the PR transparently scopes the real WS-realtime fix as a follow-up, and real e2e showed the new loop at ~11s (within window). Acceptable stopgap — flagging the tail-case race, not blocking.

## ✅ Vanished-run signal (`RunDetailModal.tsx:184`)
Run absent from `listRuns` for `MAX_MISSES=3` consecutive polls → `setGone(true)`; badge precedence correct (gone > active > …), reset on open/close, `cancelled`-guarded (no setState-after-unmount), poll stops on miss-exhaust. The `missRef/MAX_MISSES` stop path **pre-exists** — this PR only adds the `setGone(true)` badge to it.
- 🟡 **Transient-failure false-positive:** `runsApi.ts:9` `listRuns` does `httpGet(...).catch(() => [])`, so a 5xx/network blip returns `[]` (not a throw) and counts as a miss → 3 consecutive transient failures would mislabel a live run as "gone" and stop its poll. But the PR doesn't introduce the false-stop (pre-existing); it only adds a user-facing "gone" label on that condition. Clean fix (non-swallowing list call / error sentinel) is arguably out of scope. Not blocking.

## ✅ i18n / other
- `run.gone` present in BOTH en-US (`"Run no longer exists"`) + zh-CN (`"运行记录已不存在"`), same key path; called via `t("loop.run.gone")`. No new hardcoded user-facing string.
- No unhandled promise, no off-by-one (`MAX_MISSES` marks gone on the 3rd miss, correct), no regression, tsc clean. (The head absorbed #681's backfill via branch merge — #683's own files are unaffected by that.)

APPROVE — good targeted fixes; the two 🟡 (tail-case refresh race, transient→gone mislabel) are worth a follow-up but neither regresses the branch nor blocks merge.

yujiawei (approved):
# Code Review — PR #683 (octo-web)

Reviewed at head SHA `224f5ae1`, diff scoped against merge-base `7345f99` (4 files, +28 / -4). Both changes are self-contained within `@octo/dmloop` (`IssuePage`, `RunDetailModal`) plus two additive i18n keys — no exported type/prop/hook signature changes.

## 1. Spec compliance

**PASS.** The diff faithfully implements both stated fixes, with no missing pieces, no over-build, and no deviation.

- **Settle-poll (`IssuePage.tsx`):** immediate reload + bounded reloads at 2/5/9/14s, cleared on unmount and re-armed on re-dispatch. Delayed reloads route through `onMutatedRef.current()`, correctly reading the *current* filters at fire time rather than the filters captured at dispatch — the stale-closure race is genuinely resolved (verified: `reload`'s seq guard only orders responses via `my !== seq.current`, it does not re-capture params, so the ref indirection is load-bearing and correct). `settleReload` replaces `onMutated` at exactly one call site (the `NewLoopPage` `onCreated`, line 233); the other three `onMutated` consumers (line 226, 441, 446) are untouched.
- **gone-state (`RunDetailModal.tsx`):** a single boolean set **only** on the miss-exhausted path (`if (!stillActive) setGone(true)`, line 184), reset on open/close/run-change (lines 157, 162, deps `[visible, run]`). It is correctly **not** set on a normal terminal stop (terminal runs set `liveRun` to the terminal status, reset `missRef`, and stop polling without touching `gone`). Badge precedence (`gone` → `active` → `completed`/`failed`) fixes the frozen-spinner symptom. `setGone(true)` is guarded by the `cancelled` check, so it cannot fire after the effect is torn down.
- **i18n:** `run.gone` added to both `en-US.json` ("Run no longer exists") and `zh-CN.json` ("运行记录已不存在") under the correct `run` namespace block; both files remain valid JSON.

## 2. Code quality

**Approved.** No correctness, security, or build-breaking issues. The core mechanisms hold up: timer cleanup on both unmount and re-dispatch, sound `window.setTimeout → number[]` typing (target es2019 + default DOM lib), the ref indirection resolving the filter race, and `gone` proven not to go stale-true (polling halts once it is set).

The following are **P2 / advisory only** (non-blocking):

- **Bounded convergence past 14s** (`IssuePage.tsx` settle schedule). The 0/2/5/9/14s schedule stops unconditionally with no "stop once seen" check, and no other timer refreshes the issue *list* (the 15s `refreshRunning` only updates the running-chip snapshot, not the list). A quick-create that lands after ~14s would not appear until the next user interaction. In practice this self-heals (any filter change / detail open / `onChanged` re-runs `reload` with current filters), a running chip surfaces within 15s, and the code self-documents this as an intentional stopgap pending WS realtime. Consider a self-terminating poll that stops once the new issue appears and extends the give-up horizon — a tuning improvement, not a blocker.
- **`forEach(clearTimeout)`** (lines 211, 213) passes `(value, index, array)` to `clearTimeout`, which ignores the extra args. Works correctly; would trip `unicorn/no-array-callback-reference` if that rule were enabled (it is not). `forEach(id => clearTimeout(id))` reads cleaner.
- **`gone` badge reuses `loop-tr__status--fail` + `XCircle`**, identical to a genuinely *failed* run. Since "gone" means "record disappeared" rather than "execution failed", a neutral/muted treatment would read more accurately. Pure UX nit.
- **Manual synchronous create** funnels through the same `settleReload`, so it pays 4 redundant (idempotent, seq-guarded) reloads over 14s. Author-acknowledged; harmless.
- **`settleTimersRef.current`** is not emptied after timers naturally fire; the stale IDs are only cleared on next re-arm or unmount. Clearing already-fired IDs is a no-op — cosmetic only.

## 3. Verification notes / coverage

- Read all four changed files in full at the head SHA; traced `poll()`, the `missRef`/`MAX_MISSES` tolerance of the `listRuns → []` ambiguity, and the `gone` reset paths.
- Confirmed `XCircle` **is** imported (`RunDetailModal.tsx` line 9) — no missing-import build risk.
- `tsc` / lint / unit tests were not run in this review (the package has no unit harness; project convention is e2e). CSS `loop-tr__status--fail` is pre-existing and reused.
- Not measurable from the diff: the real server-side quick-create latency distribution that determines how often the >14s tail is hit.

## Verdict

**APPROVED.** Spec is clean and code quality carries only P2/advisory items, none of which block merge. The stale-closure fix and the gone-state gating are both correctly implemented.

Jerry-Xin (approved):
Re-reviewed at head `f027f349`, byte-verified. Since my prior APPROVE (at 6e359a96), the auto-refresh was **refactored from board-local settle-timers to a shared event-bus** — I re-verified the new mechanism; it's correct and actually an improvement. **APPROVE holds.**

## ✅ New event-bus auto-refresh (verified)
- `NewLoopPage.tsx:90` — on quick-create success emits `wk:loop-issues-dispatched` + `onCreated?.()` (immediate nav + `onMutated()` reload).
- `LoopPage.tsx:159-166` — the **resident** LoopPage (doesn't remount on tab/create-entry switch) listens for `dispatched` and schedules bounded `SETTLE_DELAYS_MS` timers, each re-emitting `wk:loop-issues-refresh`; clears prior timers on re-dispatch **and** on unmount.
- `IssuePage.tsx:210-212` — the mounted board subscribes to `wk:loop-issues-refresh` → `onMutated()`, cleanly `off`'d on unmount.
- **Why this is better than the settle-timers I approved before:** the old timers lived on the *remountable* IssuePage instance, so the **sidebar-new-loop path** (which remounts the board) dropped them. Moving the timers to the resident LoopPage covers both entry points (board-internal + sidebar) — a real gap fix. Timer cleanup is correct (no leak, no double-schedule).

## ✅ Vanished-run signal (unchanged, byte-identical to prior head)
`RunDetailModal.tsx` — `setGone` on `MAX_MISSES=3` consecutive misses, on the pre-existing miss-stop path; `cancelled`-guarded, poll stops. (i18n `run.gone` locale-paired, unchanged.)

## 🟡 Non-blocking (carried)
- Bounded settle re-dispatch is still finite — if the agent creates the issue after the last `SETTLE_DELAYS_MS` tick, the board won't auto-show it until the next manual reload/mutation. Same accepted stopgap as before; the real WS-realtime fix is tracked.
- `RunDetailModal` transient-failure→"gone" mislabel (`runsApi.ts:9` `listRuns` swallows errors to `[]`), pre-existing, unchanged.

Clean refactor that also closes the sidebar-entry gap. APPROVE.

OctoBoooot (approved):
## Review: #683 @ `f027f349` — delta re-review; the event-bus refactor improves on what I approved

**Verdict:** Approve. This new commit re-architects the quick-create refresh and fixes a real gap in the version I approved at `6e359a96`. Byte-verified; `code-review` CI is green, `mergeStateStatus=CLEAN`.

### What changed and why it's better
The prior version kept the settle-timers on the `IssuePage` instance (`settleReload` + `settleTimersRef`). That worked for the in-board quick-create, but the **sidebar new-loop entry remounts IssuePage** (via `openTab`), so those timers were lost on that path. The refactor relocates the timers to the persistent `LoopPage` and coordinates via a typed event contract:
- `NewLoopPage` emits `wk:loop-issues-dispatched` on dispatch success (`:88`) — both entry points funnel through here.
- `LoopPage` (persistent, doesn't remount) holds the timers, schedules bounded `wk:loop-issues-refresh` re-emits at [2,5,9,14]s, and clears them on re-dispatch and on unmount (`:152-167`).
- `IssuePage` subscribes to `wk:loop-issues-refresh` → `onMutatedRef.current()` (`:210-212`) — the stale-closure guard is preserved.
- `App.tsx` declares both events in `MittEvents`, so `pnpm typecheck` enforces name consistency across the three files.

### Verified
- No dangling `settleReload`/`settleTimersRef` left in IssuePage; `WKApp` imported.
- Event names consistent across all three files and typed in App.tsx. IssuePage subscribes only to `refresh` (the `dispatched` mention there is a comment); LoopPage has a matched `.on`/`.off` for `dispatched` with cleanup — no listener leak, no double-subscribe.
- The `RunDetailModal.gone` "vanished run" logic is untouched by this delta — still correct as previously approved.

### Net
Cleaner separation (timer ownership on the persistent shell, board just re-fetches on signal) and it closes the sidebar-entry gap. Same non-blocking follow-ups still stand as noted before (>14s late-dispatch still needs a manual refresh; transient `listRuns` failure can mislabel a run "gone" — both pre-existing, non-regressing). Approving at the new head.

yujiawei (approved):
# Code Review — PR #683 (octo-web)

Reviewed at head SHA `f027f349`, diff against base `feat/octo-loop` (7 files, +55/-4). I traced the event-bus mechanism through `NewLoopPage` → `LoopPage` → `IssuePage`, the `WKViewQueue.replaceToRoot` route semantics, and the `RunDetailModal` polling loop, and I cross-checked the two entry points and the i18n resolution against the actual files (not just the diff).

## Verdict: Approve

No blocking (P0/P1) issues. The design is sound and the two subtle claims that could have hidden bugs — "both create entry points are covered" and "only one board instance ever reloads" — hold up under the route implementation. Three minor (P2) advisory notes below; none block merge.

## What the change does

1. **Auto-refresh after async quick-create.** Quick-create returns only `{task_id}`; the agent creates the issue moments later, so the single post-dispatch reload fired before the row existed. The fix: `NewLoopPage.submit()` emits `wk:loop-issues-dispatched`; the persistent `LoopPage` owns bounded settle timers that re-emit `wk:loop-issues-refresh` at `[2000, 5000, 9000, 14000]ms`; the mounted `IssuePage` subscribes and reloads via `onMutatedRef`.
2. **`gone` state in `RunDetailModal`.** When a run vanishes from `listRuns` for `MAX_MISSES` consecutive polls, the modal now shows a "run no longer exists" badge instead of freezing on a stale "running" status.

## Correctness — verified, no blockers

- **Timer + listener lifecycle is clean.** `LoopPage`'s dispatch handler clears prior timers before scheduling new ones and clears all timers in the effect cleanup; the `mittBus.off` is paired with the `.on`. `IssuePage`'s refresh listener is registered/removed symmetrically. No leaks.
- **Stale-closure avoided correctly.** `IssuePage` reads `onMutatedRef.current` inside a `[]`-deps effect, and a separate effect keeps the ref current. This is the right call: a delayed refresh must reload with the *current* filters/view, and the `seq` guard in `reload` only orders responses, it does not protect against stale inputs. Good reasoning in the comment.
- **Both entry points are genuinely covered.** The in-board `+` (`IssuePage.openNewLoop`) and the sidebar "新建回路" (`LoopPage.openNewLoop`) both mount `NewLoopPage`, which emits `wk:loop-issues-dispatched` unconditionally on success. Because the timers live on the *persistent* `LoopPage` (never remounted by `replaceToRoot`, which only swaps the right-pane queue), the sidebar path — where `openTab("issue")` calls `replaceToRoot` and remounts a fresh `IssuePage` — is still covered: `replaceToRoot` runs synchronously on `onCreated`, well before the 2s first settle tick, so the new board instance is mounted and subscribed when the refresh fires. This is the actual fix over the first cut, which only covered the in-board entry.
- **No double-reload storm.** `WKApp.routeRight` holds a single queue; only one `IssuePage` is mounted at a time (switching tabs `replaceToRoot`s the queue to `[view]`). The myloop-tab and issue-tab instances are never mounted simultaneously, so the shared `wk:loop-issues-refresh` event reaches exactly one subscriber. Each settle tick triggers at most one reload, and `reload`'s `seq` guard drops any out-of-order responses.
- **`gone` transitions are correct.** `gone` is set only on the miss-exhausted path (`missRef >= MAX_MISSES`), reset on every open/close and on any successful `listRuns` hit (`missRef = 0`), and it takes precedence over `active` in the badge. A normal terminal stop (run completed/failed) never sets `gone`, so completed runs still show their real status. The `XCircle` import is present.
- **i18n resolves.** `t("loop.run.gone")` maps to the top-level `run` block in `en-US.json` / `zh-CN.json`, which is registered under the `loop` namespace — identical to the existing `loop.run.copied` → `run.copied`. Both locales added. No missing-key fallback.
- **`MittEvents` change is additive.** Two new void event keys in `dmworkbase/src/App.tsx`; no existing consumer references either, no exported signature changes.

## Advisory (P2 — non-blocking)

- **`clearTimeout` receives extra args.** `LoopPage.tsx` `settleTimersRef.current.forEach(clearTimeout)` passes `(id, index, array)` to `clearTimeout`. Browsers ignore the extras so this is harmless in production, but `forEach((id) => clearTimeout(id))` is the cleaner form and avoids surprises under strict polyfills or test fakes.
- **Settle cadence is a fixed heuristic.** `[2000, 5000, 9000, 14000]ms` is a reasonable stopgap for unobservable async-create latency, but if agent create latency ever exceeds ~14s the row still won't auto-appear. Acceptable given the documented deferral to WS realtime; worth a follow-up if latency proves higher in practice.
- **One redundant terminal fetch on the `gone` path.** When polling stops because the run is gone, the terminal "grab the last incremental batch" call still fires one `listRunMessages`. It's harmless (returns empty) and preserves the #610 last-batch behavior on the normal terminal path, so not worth special-casing — just noting it.

## Coverage / blind spots

- Did not run `tsc`/build or the vitest suite (`node_modules` not installed in the review workspace); relied on static reading. The types look consistent and the PR reports tsc clean.
- The board-refresh behavior is timing-dependent (settle ticks vs async DB write); I verified the wiring statically and the author reports real-env e2e on both entry points (~11s and ~12s). Not independently re-run here.
- Runtime WS realtime (the class fix) and focused unit tests for the polling/race guards are explicitly deferred; not in scope for this diff.

PR #684, opened 2026-07-13
feat(loop): 添加电脑弹窗支持 headless 免浏览器接入

caster-Q:
## Summary

Adds a second, browser-free ("headless") command to the "Add computer" dialog so a user on a remote server, SSH-only host, CI, or container can connect a runtime with a single command — matching the outcome of the existing browser-authorization flow. Depends on the backend endpoint `POST /api/cli-token/headless` (companion change in octo-multica) that mints a 90-day PAT for the authenticated user.

## Related Issue

N/A

## Changes

- `pages/RuntimePage.tsx`: the dialog now shows two commands side by side, each with its own copy button — (1) browser authorization (existing), (2) headless: `octo-daemon login --token <PAT> --server-url <daemon_server_url> && octo-daemon daemon start`.
- `api/authApi.ts`: add `issueHeadlessCliToken()` (calls the new backend endpoint) and `getDaemonServerUrl()` (reads the public backend URL from `/config`).
- The token segment of the headless command is shown as a placeholder; the one-time PAT is minted only when the user clicks copy (and reused on subsequent copies), so simply viewing or reopening the dialog never accumulates credentials.
- The copy button floats inside the command box, vertically centered on the right, removing the height mismatch; it stays fixed while a long command scrolls horizontally.
- When `daemon_server_url` is not configured, the headless section shows a hint instead of a command.
- i18n (zh/en) copy added.

## Testing

- [x] Manually verified

End-to-end, locally: the dialog mints a PAT → `login --token` authenticates with no browser → `daemon start` → 4 runtimes come online across all workspaces; then created an AI teammate → dispatched an issue → the daemon picked it up and completed it (file written, issue moved to in-review). `pnpm typecheck` and `pnpm i18n:check` pass.

## Checklist

- [ ] Updated documentation


caster-Q:
close

yujiawei (approved):
# Code Review — PR #684 (octo-web)

**Scope:** headless (no-browser) computer onboarding — a one-click flow that signs a 90-day PAT on demand and renders a `octo-daemon login --token … --server-url … && octo-daemon daemon start` command for copy. Reviewed diff: `authApi.ts`, `RuntimePage.tsx`, `runtime.css`, `en-US.json`, `zh-CN.json` (146 / -12).

This PR is **security-sensitive** (it mints and displays a long-lived credential). None of the findings below block the happy path or cross a privilege boundary — the token is the requesting user's own credential, and both interpolated values are backend/admin-controlled — so this is **Approved**. That said, several hardening items are worth addressing, and two items deserve a human security sign-off before this pattern spreads.

## Spec compliance

Matches the described feature exactly:
- Adds `issueHeadlessCliToken()` → `POST /cli-token/headless` and `getDaemonServerUrl()` → `GET /config` (`authApi.ts`).
- Command box shows a placeholder token segment until the user clicks Copy; the real PAT is signed only on click, and reused (not re-minted) on subsequent copies within the same page session (`RuntimePage.tsx:183-207`).
- Graceful degradation when the backend has no public URL configured (`headlessNoBackend`).
- i18n keys added to both `en-US` and `zh-CN` with full parity.

No over-build and nothing missing relative to the PR description. ✅

## Findings

### P2 — Double-click can mint multiple 90-day PATs (race)
`onCopyHeadless` (`RuntimePage.tsx:183`) guards re-minting with `if (!cmd)`, but there is no synchronous in-flight guard. `headlessLoading` drives the button's `loading` prop, but Semi's `loading` does not disable the button, and the state update is async — a rapid second click can enter the handler again while `headlessCommandText` is still `""` and the first `await issueHeadlessCliToken()` has not resolved, so both calls mint a token. This defeats the stated anti-accumulation intent (see the comment at `RuntimePage.tsx:181-183`) and silently leaves dangling long-lived credentials. Fix is one line — an early `if (headlessLoading) return;` (or a `useRef` in-flight flag). Flagged independently by two review passes.

### P2 — Raw string interpolation into the copied command
`headlessCommand(token, backendUrl)` (`RuntimePage.tsx:39`) interpolates both values into the shell command with no quoting. This is **not** DOM XSS (React escapes the `<code>` text) and **not** a cross-user vulnerability — the token is backend-signed and `daemonServerUrl` comes from admin-set `MULTICA_PUBLIC_URL`, and the command runs as the same user on their own machine. But a malformed `MULTICA_PUBLIC_URL` (stray space/`;`/`&`) would silently break or alter the copied command. Consider validating that `daemonServerUrl` parses as a URL and/or quoting the args for robustness (defense-in-depth).

### P2 — Token persists in component state; not cleared on modal close
`headlessCommandText` holds the real PAT-bearing command and is never reset when the modal closes (`setAddOpen(false)` at `RuntimePage.tsx:284,288` leaves it intact). Because `RuntimePage` stays mounted, the secret survives modal close and is re-displayed on reopen — visible to shoulder-surfing, screenshots, and browser extensions. Consider clearing `headlessCommandText`/`headlessCopied` on close, and/or revealing the token only behind an explicit action rather than rendering it inline by default.

### P2 — No expiry / revocation affordance in the UI
The flow creates a 90-day credential with no visible warning, expiry indicator, or revocation link. For a long-lived PAT, surfacing "this creates a 90-day token" and a path to revoke it would improve credential hygiene.

### Nit — Orphaned i18n keys
`loop.runtime.copyCommand` is now unused (the old footer copy button was removed) and `loop.runtime.generateHeadless` is added but never referenced (0 uses in `.tsx`). Safe to drop to keep the locale files clean.

## For human security review (label already applies)

1. **Token on the command line.** The generated command passes the PAT as `--token <PAT>`, which is exposed via the process table (`ps` / `/proc`) and persisted in shell history (`~/.bash_history`) on the target machine. This is inherent to the CLI's `login --token` interface, so it can't be fully fixed in this repo — but if `octo-daemon` supports credential ingestion via env var or stdin, prefer generating that form. Worth a human decision given the 90-day lifetime.
2. **Backend contract for `POST /cli-token/headless`.** This review cannot verify token scope, lifetime enforcement, rate-limiting, or revocation on the backend, nor CSRF/auth enforcement on the new endpoints. Please confirm the companion backend MR bounds these appropriately.

## Coverage notes

Reviewed the frontend diff only. Not verifiable from this PR: backend `/cli-token/headless` and `/config` behavior (scope/lifetime/revocation/rate-limit), the `octo-daemon` CLI's credential-ingestion options, and the `copyToClipboard` implementation in `@octo/base`. Build/typecheck not run in this environment; i18n key parity verified manually (both locales in sync).

**Verdict:** Approved — no blocking correctness/security defect on the happy path; the items above are hardening/UX improvements plus two points for human security sign-off.

PR #597, opened 2026-07-10
feat(loop): add Multica CLI authorization page

liuooo:
## Summary
Add an Octo Web confirmation page for the Multica CLI add-computer authorization flow.

## Changes
- Register `/loop/multica/cli-authorize` as the temporary CLI authorization route, with a TODO to revisit the product path.
- Validate the CLI callback URL before redirecting back to the local CLI listener.
- Reuse Octo Web login state and selected Space, load Loop workspaces, and issue a Multica CLI token through the existing `/cli-token` API path.
- Add localized copy and styles for the authorization page.

## Testing
- `pnpm i18n:check`
- `pnpm exec prettier --check packages/dmloop/src/api/authApi.ts packages/dmloop/src/pages/MulticaCliAuthorizePage.tsx packages/dmloop/src/index.tsx packages/dmloop/src/module.tsx`
- `git diff --check`

## Notes
The route is intentionally temporary and can be renamed once the final Octo Web product path is confirmed.

PR #685, opened 2026-07-13
feat(loop): add headless onboarding command to the Add Computer dialog

caster-Q:
## Summary

Adds a second, browser-free ("headless") command to the "Add computer" dialog so a user on a remote server, SSH-only host, CI, or container can connect a runtime with a single command — matching the outcome of the existing browser-authorization flow. Depends on the backend endpoint `POST /api/cli-token/headless` (companion change in octo-multica) that mints a 90-day PAT for the authenticated user.

## Related Issue

N/A

## Changes

- `pages/RuntimePage.tsx`: the dialog now shows two commands side by side, each with its own copy button — (1) browser authorization (existing), (2) headless: `octo-daemon login --token <PAT> --server-url <daemon_server_url> && octo-daemon daemon start`.
- `api/authApi.ts`: add `issueHeadlessCliToken()` (calls the new backend endpoint) and `getDaemonServerUrl()` (reads the public backend URL from `/config`).
- The token segment of the headless command is shown as a placeholder; the one-time PAT is minted only when the user clicks copy (and reused on subsequent copies), so simply viewing or reopening the dialog never accumulates credentials.
- The copy button floats inside the command box, vertically centered on the right, removing the height mismatch; it stays fixed while a long command scrolls horizontally.
- When `daemon_server_url` is not configured, the headless section shows a hint instead of a command.
- i18n (zh/en) copy added.

## Testing

- [x] Manually verified

End-to-end, locally: the dialog mints a PAT → `login --token` authenticates with no browser → `daemon start` → 4 runtimes come online across all workspaces; then created an AI teammate → dispatched an issue → the daemon picked it up and completed it (file written, issue moved to in-review). `pnpm typecheck` and `pnpm i18n:check` pass.

## Checklist

- [ ] Updated documentation


packages/dmloop/src/pages/RuntimePage.tsx
```diff
@@ -158,6 +179,37 @@ export default function RuntimePage() {
...
+  // 才向后端签发一次性 PAT 并把真实命令写入框；已签发过则直接复制既有命令，
+  // 避免反复点击在账号里累积凭证。
+  const onCopyHeadless = async () => {
+    if (!daemonServerUrl) {
+      Toast.warning(t("loop.runtime.headlessNoBackend"));
+      return;
+    }
+    setHeadlessLoading(true);
+    try {
+      let cmd = headlessCommandText;
```
OctoBoooot, line 192: **[major]** Double-mint race. The single-flight guard `if (!cmd)` reads `cmd = headlessCommandText` (React state), and `setHeadlessLoading(true)` above is also async — two clicks in the same tick both see `headlessCommandText` empty, both pass, both `await issueHeadlessCliToken()` -> two 90-day PATs minted. This defeats the comment's own promise of not accumulating credentials on repeated clicks. Fix: a synchronous re-entrancy guard — `const mintingRef = useRef(false); if (mintingRef.current) return; mintingRef.current = true;` at the top of onCopyHeadless, reset in `finally`.

packages/dmloop/src/pages/RuntimePage.tsx
```diff
@@ -158,6 +179,37 @@ export default function RuntimePage() {
     Toast.success(t("loop.runtime.copySuccess"));
   };
 
+  // 命令2 的复制：命令框里 token 段先显示占位符，只有点击复制的这一刻
+  // 才向后端签发一次性 PAT 并把真实命令写入框；已签发过则直接复制既有命令，
+  // 避免反复点击在账号里累积凭证。
+  const onCopyHeadless = async () => {
```
OctoBoooot, line 183: **[major]** High-risk auth path with no tests. This mints a 90-day PAT and branches on mint-once plus the no-backend case, but there is no unit coverage (checklist unchecked, manual-only). Add a test asserting onCopyHeadless mints exactly once under a double-click, plus the no-backend degradation branch — it would pin the mint-once invariant directly.

packages/dmloop/src/pages/RuntimePage.tsx
```diff
@@ -233,19 +285,48 @@ export default function RuntimePage() {
...
+              onClick={copyCommand}
+            >
+              {copied ? t("loop.runtime.copied") : t("loop.runtime.copy")}
+            </Button>
+          </div>
+
+          <p>{t("loop.runtime.addComputerHeadless")}</p>
+          {daemonServerUrl ? (
+            <div className="loop-runtime-add__row">
+              <pre className="loop-runtime-add__command"><code>{headlessCommandText || headlessCommand(t("loop.runtime.headlessTokenPlaceholder"), daemonServerUrl)}</code></pre>
```
OctoBoooot, line 323: **[minor]** Minted token persists across dialog close. `headlessCommandText` holds the real PAT and isn't cleared on `setAddOpen(false)` (L284/L288), so reopening re-renders the live token here. Client-side/same-user, but re-showing a standing credential on reopen is avoidable — clear headlessCommandText (and headlessCopied) on close.

OctoBoooot (changes requested):
## Review: feat(loop): headless onboarding command for Add Computer dialog (#685)

**Verdict:** Request changes — the credential-hygiene design is mostly good, but there's a double-mint race that defeats the PR's own stated "don't accumulate credentials" invariant, and no test coverage on a high-risk auth path.

Risk tier: **high** (auth API surface — `authApi.ts` mints a 90-day PAT). Base `feat/octo-loop`. `code-review` CI is PENDING (not red); `mergeStateStatus=UNSTABLE`. (This is the reopened #684 — same head `bfb6e814`.)

### Major — double-mint race in `onCopyHeadless` (RuntimePage.tsx:185-199)
The single-flight guard is `if (!cmd)` where `cmd = headlessCommandText` (`:192`), a React state value, and `setHeadlessLoading(true)` (`:190`) is likewise async. Two clicks dispatched in the same tick both read `headlessCommandText === ""`, both pass the guard, and both call `await issueHeadlessCliToken()` → **two 90-day PATs minted**. The code comment explicitly promises "避免反复点击在账号里累积凭证" — but the async `loading` flag can't block the second same-tick click, so that invariant doesn't hold. Failing sequence: user double-clicks Copy before the first network round-trip resolves → 2 standing PATs on the account, only one ever surfaced in the box.
Fix: a synchronous re-entrancy guard — `const mintingRef = useRef(false); if (mintingRef.current) return; mintingRef.current = true;` at the top, reset in `finally`. (The same confirmingRef pattern used elsewhere in dmloop.)

### Major — no test coverage on a high-risk auth path
Tier is high and this adds token-minting (`issueHeadlessCliToken`, `getDaemonServerUrl`) plus the branching `onCopyHeadless` mint-once logic, with zero tests (checklist "Added tests" unchecked; manual-only). A unit test asserting `onCopyHeadless` mints exactly once under a double-click would pin the invariant above and prevent regression. Please add coverage for the mint-once path and the `!daemonServerUrl` degradation branch.

### Minor — minted token persists in state/DOM across dialog close (RuntimePage.tsx:127, 284/288, 314)
`headlessCommandText` holds the real token+command and isn't cleared on `setAddOpen(false)`, so closing and reopening the dialog re-renders the live PAT in the `<code>` block (`:314`). It's client-side and same-user, but re-displaying a standing credential on reopen is avoidable — consider clearing `headlessCommandText` (and `headlessCopied`) on dialog close so a minted PAT isn't shown again.

### Praise (verified)
- **Mint-on-copy, not on render.** The command shows a placeholder (`headlessTokenPlaceholder`) until the user clicks Copy, so merely viewing or reopening the dialog never mints a token — exactly the right credential-hygiene default. (The double-click hole above is the one gap in an otherwise well-thought-out design.)
- **No XSS.** Token and backend URL are interpolated into `<code>{...}</code>` — JSX auto-escapes, no `dangerouslySetInnerHTML`. Safe.
- **Graceful degradation.** When `daemon_server_url` is unconfigured, the section shows `headlessNoBackend` instead of a broken command. i18n complete and translated in both en-US and zh-CN.

### Net
The security posture is thoughtfully designed (mint-on-copy, placeholder, no injection). The one real defect is the double-click double-mint — a synchronous ref guard closes it — and given the high-risk tier, this token path should ship with a test. Fix those two and this is close.

Jerry-Xin (dismissed):
Reviewed at head `bfb6e8141939`, byte-verified (this is the same code as the now-closed #684). Headless/browserless onboarding via a device-PAT + copy-command flow. Security-sound; **APPROVE** with a few non-blocking hardening notes.

## ✅ Security (byte-verified)
- **No loopback-validation bypass.** The `isSafeCliCallback` loopback guard (#597/#681) protects the *browser* flow, where a token is appended to a callback URL the browser then redirects to. This headless path has **no callback URL and no redirect** — the backend mints a PAT (`authApi.ts:9 issueHeadlessCliToken → POST /cli-token/headless`) and the frontend renders it into a copy-command string. The guard is inapplicable, not circumvented.
- **No token leak / no injection.** Token is rendered as escaped JSX text in `<code>{...}</code>` — no `dangerouslySetInnerHTML` (grep NONE), never placed in a URL/query, never logged.
- **No SSRF / open-redirect.** `getDaemonServerUrl()` (`GET /config` ← server `daemon_server_url`/`MULTICA_PUBLIC_URL`) is rendered as text only, never fetched or redirected to by the frontend.
- CSRF double-submit is enforced for non-GET by `http.ts`; PAT minted lazily on click (not on modal open).

## 🟡 Non-blocking hardening
1. **Shell-quote the interpolations** — `RuntimePage.tsx headlessCommand()`: `octo-daemon login --token ${token} --server-url ${backendUrl} ...` interpolates unquoted. Both values are **server/admin-controlled (not browser/attacker input)** and the user pastes into their own shell, so this is not a browser-reachable RCE — I disagree with grading it a blocking 🔴. But the repo already has a `shellQuote` helper (`IncomingWebhook.ts:559`) used for exactly these copy-commands; quoting `token`+`backendUrl` is cheap defense-in-depth (guards against a malformed server value breaking the paste). (Note: the sibling `addComputerCommand()` interpolates `${currentOrigin()}` unquoted too — pre-existing convention.)
2. **Duplicate-PAT race** — `onCopyHeadless` reads command text from React state; a rapid double-click before state commits could mint two 90-day PATs. `loading` state mitigates; a `useRef` promise guard would make "mint once" explicit. Mild consequence (extra PATs on account).
3. **Dead i18n key** — `loop.runtime.generateHeadless` is defined in both locales but never referenced by `t()`. Remove or wire it.

## ✅ Otherwise
i18n locale-parity perfect (all 8 new `loop.runtime.*` keys in both en-US + zh-CN, same paths); no hardcoded user-facing string; no state-after-unmount beyond the file's pre-existing unguarded-effect pattern (React warn only); runtime.css benign (no off-screen/clickjacking surface); `cliCallback.ts` untouched. tsc/CI green.

APPROVE — the flow is security-sound; the three notes (shell-quote, mint-once guard, dead key) are worth a quick follow-up but none blocks merge.

Jerry-Xin (changes requested):
Re-reviewed at `bfb6e8141939` after OctoBoooot's CR — byte-verified, and I'm converging with it. The core flow is still security-sound (no loopback bypass / SSRF / token-in-URL — verified in my prior review), but the duplicate-PAT race is real and, on a credential-minting path, blocking. Flipping my earlier APPROVE to REQUEST_CHANGES.

## 🔴 Duplicate-PAT double-click race (adopted from OctoBoooot; I under-weighted this)
`RuntimePage.tsx:185 onCopyHeadless` guards `if (!cmd)` where `cmd = headlessCommandText` (React **state**), and relies on `headlessLoading` (also state) + the button's `loading` prop. But state updates aren't synchronous: a rapid double-click fires two `onCopyHeadless` in the same tick, **both** read `headlessCommandText === ""`, **both** call `issueHeadlessCliToken()` → **two 90-day PATs minted** from one gesture. The `loading` prop only disables after re-render, which loses the race. On an auth/credential path, uncontrolled minting of multiple long-lived PATs is a real concern, not cosmetic — I had graded it 🟡, but I agree with the 🔴. Fix: a synchronous `useRef` in-flight guard (mint once), the pattern this repo already uses elsewhere.

## 🟡 Missing mint-once test (OctoBoooot)
No test asserts the credential is minted only once / the race is guarded. For a security-sensitive PAT-issuing path, a regression test locking "one mint per gesture" is a fair ask alongside the fix.

## 🟡 Shell-quote the command interpolations
`headlessCommand()` interpolates `${token}`/`${backendUrl}` unquoted. Values are server/admin-controlled and pasted into the user's own shell (not a browser-reachable RCE), so I don't grade this 🔴 — but the repo has a `shellQuote` helper (`IncomingWebhook.ts:559`); quoting is cheap defense-in-depth.

## 🟡 Modal-close doesn't clear credential state (OctoBoooot) + dead i18n key
- Closing the dialog leaves `headlessCommandText` set, so a re-open shows the previously-minted PAT. Minor, but tied to the mint-once fix — clearing on close is sensible.
- `loop.runtime.generateHeadless` defined in both locales but never referenced — remove or wire.

## ✅ Still sound (from prior review)
No loopback-validation bypass (no callback/redirect path; `isSafeCliCallback` inapplicable), no token-in-URL/log, no SSRF (`daemon_server_url` rendered as text only), no `dangerouslySetInnerHTML`, CSRF enforced by `http.ts`, i18n locale-parity perfect.

REQUEST_CHANGES — the mint-once guard is the blocker; the rest are cheap follow-ups. Credit OctoBoooot for correctly weighting the duplicate-PAT race on a credential path where I under-graded it.

caster-Q:
Thanks for the thorough review. Addressed in `e5f9e994`:

- **Major — double-mint race (`onCopyHeadless`):** added a synchronous `useRef` re-entrancy guard (`mintingRef`) checked/set before any await and reset in `finally`, so same-tick double-clicks can no longer both pass the `if (!cmd)` check and mint two PATs. Matches the `busyRef`/`committed` ref pattern already used in dmloop.
- **Major — test coverage:** extracted the command builder into a dependency-free module `pages/headlessCommand.ts` and added `__tests__/headlessCommand.test.ts` (vitest), covering the quoting/escaping and a malformed-URL injection case. (dmloop has no jsdom/@testing-library setup, so the mint-once guard itself is covered by the ref pattern rather than a component render test — happy to add a component test if the harness is introduced.)
- **Minor — token persists across dialog close:** dialog close now goes through `closeAddDialog`, which clears `headlessCommandText` and `headlessCopied`, so reopening no longer re-renders a previously minted PAT.
- **Hardening — shell-quote:** `headlessCommand` now POSIX-single-quotes both `token` and `backendUrl` (`shellQuote`), guarding against a malformed server value breaking the pasted command.
- **Dead i18n key:** removed `loop.runtime.generateHeadless` from both locales.

`pnpm typecheck`, `pnpm i18n:check`, and the new unit tests pass.

yujiawei (changes requested):
# Code Review — PR #685 (octo-web)

Reviewed at head SHA `bfb6e81419392691edc46bc14a8a82a299546ef5`, base `feat/octo-loop`. This is a security-sensitive change (it mints and displays a long-lived 90-day PAT), so I've paid extra attention to the credential-handling path and flagged the items a human should verify against the backend companion PR.

## Summary of the change

The Add Computer dialog now offers a second, browser-free command. On the *first* click of the headless "Copy" button, `onCopyHeadless()` mints a 90-day PAT via `POST /cli-token/headless`, assembles `octo-daemon login --token <PAT> --server-url <daemon_server_url> && octo-daemon daemon start`, caches it in `headlessCommandText` state, and reuses it on later clicks. The section is hidden (with a hint) when `daemon_server_url` is unavailable. Lazy-mint-on-click, placeholder-until-first-copy, and single-mint-per-page-mount are all good, deliberate choices.

## Spec compliance

The diff implements what the PR describes: two side-by-side commands with independent copy buttons, lazy PAT minting on copy, a floating copy button, the no-backend hint, and zh/en i18n. Nothing out-of-scope was added, with one minor exception noted below (a dead i18n key). No missing/divergent behavior against the stated intent.

---

## Blocking issue

### P1 — Double-click / concurrent click can mint multiple orphaned 90-day PATs

`packages/dmloop/src/pages/RuntimePage.tsx:185-201`

The credential-accumulation guard is:

```ts
let cmd = headlessCommandText;
if (!cmd) {
  const { token } = await issueHeadlessCliToken();
  cmd = headlessCommand(token, daemonServerUrl);
  setHeadlessCommandText(cmd);
}
```

The guard reads React state `headlessCommandText` (declared at line 127), which is only committed on a subsequent render. `setHeadlessLoading(true)` at line 190 is likewise asynchronous, and — importantly — the Semi UI `Button` `loading` prop (line 320) only swaps in a spinner icon; it does **not** set `disabled`, so `onClick` still fires while loading. I confirmed this in the Semi UI source: the icon button renders the spinner but passes `onClick` through unchanged and only stops propagation when `disabled` is true.

Consequently, a fast double-click (or a second click before the first `await issueHeadlessCliToken()` resolves) enters `onCopyHeadless` twice; both invocations observe `headlessCommandText === ""`, both pass `if (!cmd)`, and both call `issueHeadlessCliToken()`. The backend mints two 90-day PATs, but the UI only keeps the last one — the earlier PAT becomes an orphaned, live, long-lived credential with no way to see or revoke it from this screen. This defeats exactly the invariant the inline comment at lines 182-184 promises ("avoid accumulating credentials on repeated clicks"), on a security-sensitive endpoint, via a very ordinary user action (double-clicking a copy button).

Fix is small: add a synchronous in-flight guard that does not depend on async state, e.g.

```ts
const mintingRef = useRef(false);
// inside onCopyHeadless, before the await:
if (mintingRef.current) return;
mintingRef.current = true;
try { /* ... */ } finally { mintingRef.current = false; }
```

(A ref flips synchronously, so the second click short-circuits before it can issue a second mint.)

---

## Non-blocking (recommended hardening / P2)

These do not block merge on their own, but for a credential-minting flow they're worth addressing:

- **Token state is never cleared on dialog close** — `RuntimePage.tsx:284,288` only call `setAddOpen(false)`; `headlessCommandText` persists for the whole page-mount. The minted PAT therefore stays in the DOM (`<pre><code>` at line 314) and JS heap after the dialog closes, recoverable via devtools/heap or any page-level XSS until navigation. Reusing the same token on reopen within one mount is intended and fine, but consider clearing the state on `onCancel` to shorten the in-memory/DOM exposure window, and/or masking the visible token while still copying the full value.

- **Accumulation is bounded only per page-mount** — a page refresh, navigating away and back, or a second tab resets `headlessCommandText` and mints a fresh 90-day PAT on the next copy, with no UI to list/revoke prior ones. Whether this is acceptable depends on the backend (`POST /cli-token/headless`): a human should confirm it caps/rotates live headless PATs per user and/or surfaces them in a revocation UI. Frontend cannot bound this.

- **CSRF for this mint endpoint** — `packages/dmloop/src/api/http.ts:36-43` uses a double-submit CSRF scheme, but `ensureCsrfToken()` will *self-mint* a random `multica_csrf` cookie client-side when none exists and echo it in the `X-CSRF-Token` header. On its own, that equality check is trivially satisfiable by any script in the page. For a high-value 90-day-PAT mint, please verify server-side that (1) the endpoint requires the primary session/bearer auth, and (2) `multica_csrf` is server-issued and session-bound (i.e. a self-minted cookie is rejected). This is a backend property that can't be validated from octo-web — flagging for manual verification.

- **PAT passed as a CLI argument** — `headlessCommand()` (`RuntimePage.tsx:39-42`) embeds the raw PAT into `octo-daemon login --token <PAT>`. Tokens on the command line land in shell history and are visible via the process table (`ps`) while running. This is an `octo-daemon` CLI concern rather than a web fix, but the safer pattern is to accept the token via stdin or an env var; worth a follow-up on the CLI side.

- **`/config` is fetched once with no retry** — `RuntimePage.tsx:130-132` runs `getDaemonServerUrl()` once on mount and collapses both "not configured" and "fetch failed" to `""`. A transient GET failure then permanently shows the "administrator has not configured a public backend URL" hint for the page lifetime (recoverable only by reload), which misattributes a network blip to a config problem. Consider retrying on dialog open or distinguishing the two cases. (Also, the async `setDaemonServerUrl` has no unmount guard — minor.)

## Nits

- **Dead i18n keys.** `loop.runtime.generateHeadless` (en/zh line 115) is added by this PR but never referenced — the headless button reuses `loop.runtime.copy`/`copied`. `loop.runtime.copyCommand` (line 107) was orphaned by this PR's copy-button refactor. Both are safe to remove from both locale files. (The two locale files are otherwise in sync.)
- **Minor DRY.** `copyCommand` / `onCopyHeadless` share a copy→toast→flag shape, and the two flag-reset `useEffect`s (lines 135-145) are near-identical — a small `useTransientFlag` hook / shared helper would prevent drift. Optional.
- **CSS** for the floating copy button (`runtime.css:263-284`) is correct; the `88px` reserved padding vs. button width is a hand-tuned magic number that could clip under a much longer locale label — cosmetic only.

## Coverage note

I reviewed all five changed files on disk plus the relevant Semi UI `Button`/`IconButton` source (to confirm `loading` does not disable clicks) and `copyToClipboard`. Out of scope / could not verify from this repo: the backend `POST /cli-token/headless` and `GET /config` handlers (PAT scope, expiry, per-user caps, rotation, revocation, rate limiting, audit logging, and the CSRF/session binding above), and actual token entropy/format. The double-mint (P1) assumes the endpoint mints unconditionally per call, which the client code and comments indicate; if the backend already dedupes/caps, the impact is reduced but the client-side guard should still be fixed.

## Verdict

One P1 (double-mint race) directly defeats a documented security invariant on a credential-minting flow via ordinary double-clicking, with a cheap fix — requesting changes for that. The P2 items are hardening/backend-verification, not independent blockers.

yujiawei (approved):
# Code Review — PR #685 (octo-web)

Re-review at head SHA `e5f9e99414339b32510416908db2a889d75791cd` (base `feat/octo-loop`). My previous review requested changes at `bfb6e814` for a token-minting race; the new commit `e5f9e994` addresses it. Giving a fresh verdict on the current head — the prior request-changes does not carry over automatically.

Commit `e5f9e994` ("headless 命令签发防重入 + 关弹窗清凭证 + shell 转义") resolves the blocking finding and the recommended hardening:

- **P1 double-mint race — FIXED.** `RuntimePage.tsx:127,189-190,209` adds a synchronous `mintingRef` guard: `if (mintingRef.current) return; mintingRef.current = true;` is set *before* the `await issueHeadlessCliToken()`, and reset in `finally`. Because a ref flips synchronously (unlike state / the `loading` prop), a fast double-click short-circuits before it can issue a second PAT. This closes the concurrent-mint window exactly.
- **Token state cleared on dialog close — FIXED (P2).** `closeAddDialog()` (`RuntimePage.tsx:216-220`) now clears `headlessCommandText` and `headlessCopied`, wired to both the modal `onCancel` and the footer Cancel button. The minted PAT no longer lingers in DOM/state after the dialog closes, and reopening starts clean.
- **Shell-argument hardening — ADDED.** The command builder was extracted to `packages/dmloop/src/pages/headlessCommand.ts` with a `shellQuote()` that applies POSIX single-quote escaping (`'` → `'\''`) to both the token and the backend URL, so a malformed server-supplied `daemon_server_url` can't break out of the argument when the user pastes the command. Backed by unit tests in `__tests__/headlessCommand.test.ts`, including an explicit `rm -rf ~` injection regression.
- **Dead i18n key removed (nit).** The unused `generateHeadless` key is gone from both locales.

I verified the fix by reading the files at the current head (static review; no `node_modules` in this checkout, so the new vitest suite was reviewed but not executed — the `shellQuote` assertions are correct POSIX escaping).

## Remaining (non-blocking)

- **CSRF double-submit is client-forgeable** (`api/http.ts` `ensureCsrfToken`) — the frontend self-mints `multica_csrf` when absent, so the double-submit check is only meaningful if the server binds the cookie to the authenticated session and requires the primary `token`/session auth on `POST /cli-token/headless`. This is a backend property; please confirm on the companion PR. Not fixable or verifiable from octo-web.
- **PAT lifecycle** — a page refresh / second tab still mints a fresh 90-day PAT; confirm the backend caps/rotates live headless PATs per user and offers a revoke/list path. Backend concern.
- **`/config` fetched once, no retry** (`RuntimePage.tsx:129-131`) — a transient failure permanently shows the "backend not configured" hint until reload. Minor UX; optional.
- Pre-existing orphan i18n key `copyCommand` (locale line 107) predates this PR — leaving it out of scope.

## Verdict

The P1 that blocked the previous round is correctly fixed and the P2 hardening was taken on with tests. No P0/P1 remains in the diff. The open items are backend verifications and minor UX, none blocking this frontend PR. Approving.

OctoBoooot (approved):
## Review: #685 @ `e5f9e994` — flipping to APPROVE; all findings fixed (+ shell-quoting hardened)

**Verdict:** Approve — the new commit resolves every finding from my prior CR (at `bfb6e814`), byte-verified. This flips my REQUEST_CHANGES.

Risk tier: high (auth — mints a 90-day PAT). Base `feat/octo-loop`.

### Resolved — all three of my findings (verified fixed)
1. **Double-mint race (was major) — FIXED.** `mintingRef` (`RuntimePage.tsx:122`) is a synchronous re-entrancy guard: `if (mintingRef.current) return; mintingRef.current = true;` (`:188-189`) sits **before** the first `await issueHeadlessCliToken()` (`:194`) and is reset in `finally` (`:209`). A same-tick double-click now mints exactly one PAT — the first call sets the flag before yielding, the second returns at `:188`. This is precisely the fix and it's correct.
2. **Token persists across dialog close (was minor) — FIXED.** `closeAddDialog` (`:216`) clears `headlessCommandText` + `headlessCopied`, and **both** cancel paths use it — `onCancel={closeAddDialog}` (`:293`) and the footer button (`:297`). Reopening no longer re-renders a prior PAT.
3. **Missing mint-once test (was major) — ADDRESSED.** `headlessCommand` extracted to its own module with `headlessCommand.test.ts` (4 cases). (A UI-level double-click test on `onCopyHeadless` would still be nice, but the security-critical string logic is now covered — see below.)

### Bonus — shell-quoting hardened (peer 🟡, went beyond my CR)
The extracted `headlessCommand.ts` now quotes both interpolations, and the test covers the exact adversarial inputs: wrapping a plain value, escaping an embedded single-quote so a value can't break out of the quoting, and neutralizing a malformed backend URL so it can't inject extra shell commands. This is the defense-in-depth Steve raised as a non-blocking nudge — good to see it done properly rather than deferred, especially on a credential path.

### Also cleaned up
Dead i18n key `generateHeadless` removed from both locales.

### Net
Complete, correct remediation of the double-mint race (the one real blocker-adjacent defect), plus token-hygiene-on-close and shell-quoting hardening. The credential-minting path is now sound: mint-on-copy, single-flight guarded, quoted interpolation, cleared on close, no XSS. Nice turnaround. Approving.

(Note: yujiawei's CR is anchored to this head but still lists the double-click race — that's resolved at `e5f9e994` by the `mintingRef` guard; verified.)

Jerry-Xin (approved):
Re-reviewed at head `e5f9e9`, byte-verified. **The author fixed every item — both my blocker and all the non-blocking notes.** Flipping to APPROVE.

## ✅ 🔴 mint-once race — FIXED (synchronous guard)
`RuntimePage.tsx:127` `const mintingRef = useRef(false)`; `onCopyHeadless:188-189` `if (mintingRef.current) return; mintingRef.current = true;` set **synchronously before** the `await issueHeadlessCliToken()`, reset in `finally:209`. A same-tick double-click: the first sets the ref true synchronously, the second hits the guard and bails → **only one PAT minted**. This is the synchronous `useRef` in-flight lock (not a state-based guard that would race) — the double-mint is genuinely closed.

## ✅ 🟡 shell-quote — FIXED
New `headlessCommand.ts`: `shellQuote()` POSIX single-quote escaping applied to both `token` and `backendUrl` in `headlessCommand()`. `headlessCommand.test.ts` even asserts an injected `http://x'; rm -rf ~; echo '` is neutralized (stays inside one quoted argument). Solid defense-in-depth.

## ✅ 🟡 modal-close clears credential state — FIXED
Close handler (`:217-219`) now `setAddOpen(false); setHeadlessCommandText(""); setHeadlessCopied(false)` — the minted command no longer lingers in state/DOM after the dialog closes.

## ✅ 🟡 dead i18n key — FIXED
`loop.runtime.generateHeadless` removed from both en-US + zh-CN.

## ✅ Security (carried, still sound)
No loopback-validation bypass (no callback/redirect path), no token-in-URL/log, no SSRF (`daemon_server_url` rendered as text), no `dangerouslySetInnerHTML`, CSRF enforced by `http.ts`, PAT minted lazily on click, i18n locale-parity intact.

Clean, complete turnaround — mint-once is now a real synchronous lock, the command is shell-quoted with an injection test, and credential state is cleared on close. APPROVE.

PR #686, opened 2026-07-13
fix(versions): vertically center version-history confirm overlay

ploy-elison:
## What

Follow-up to #674. That PR moved the version-history restore/delete confirm box into the shared `.octo-modal-overlay` so it stays on-screen on long scrolled lists, but the overlay is top-anchored (`align-items: flex-start`, `48px` top padding) — the preview/diff modal needs that so tall content scrolls from the top. As a result the small confirm card sat near the top of the viewport (~48px from the top) rather than vertically centered, which does not match the centered-dialog decision.

## Changes

- Add a scoped `.octo-modal-overlay--center` modifier and apply it only to the two confirm overlays (restore + delete) in the unified `VersionHistoryPanel` shell. Since doc/sheet/board are all thin adapters over this shell, all three ends get the fix.
- The modifier flips vertical alignment to `safe center`: the card is viewport-centered but degrades to top-aligned if the content ever overflows, so the action buttons stay reachable instead of being clipped off-screen.
- The preview/diff overlay does **not** carry the modifier, so its top-anchored, scrollable behavior is unchanged — the two overlays stay naturally isolated.
- Shared `.octo-modal-overlay` rules, the confirm state machine, race guard, and mounted-ref liveness are untouched. No JS behavior change beyond the added class.
- Extend the shell regression test to assert the restore and delete confirm overlays carry the modifier and the preview overlay does not.

## Testing

- `packages/docs` version tests green (`vitest run src/versions/` — 78 passing, +1 new).
- `typecheck`: no new errors introduced by this change (two pre-existing errors in unrelated files remain).
- `stylelint`: no new violations; the added rule uses only `align-items`, no color literals.

### DIFF
--- a/packages/docs/src/editor/styles.css
+++ b/packages/docs/src/editor/styles.css
@@ -1054,6 +1054,16 @@
   background: rgba(0, 0, 0, 0.4);
   overflow-y: auto;
 }
+/* Vertically-center modifier for the destructive-action confirm (decision #2). The base overlay is
+   top-anchored (align-items: flex-start) because the preview/diff modal can be taller than the
+   viewport and needs to scroll from the top. The small restore/delete confirm card, however, should
+   sit centered in the viewport. `safe center` keeps it centered but degrades to top-aligned if the
+   content ever overflows, so the action buttons stay reachable instead of being clipped off-screen.
+   Only the confirm overlays carry this modifier; the preview/diff overlay keeps its top-anchored,
+   scrollable behavior. */
+.octo-modal-overlay--center {
+  align-items: safe center;
+}
 .octo-modal {
   width: 100%;
   max-width: 560px;
--- a/packages/docs/src/versions/VersionHistoryPanel.test.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.test.tsx
@@ -336,6 +336,36 @@ describe('VersionHistoryPanel — mutations & permissions', () => {
     await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
   })
 
+  it('vertically-centers the confirm overlay via a scoped modifier the preview/diff overlay does not carry', async () => {
+    // Regression for XIN-887 (decision #2): the confirm card must sit centered in the viewport, not
+    // top-anchored like the preview/diff modal. The base .octo-modal-overlay is top-anchored (it can
+    // hold taller-than-viewport preview content that scrolls from the top), so the confirm overlay
+    // opts into centering with the scoped .octo-modal-overlay--center modifier. jsdom does not compute
+    // layout, so we assert the class contract: the confirm overlay carries the modifier and the
+    // preview overlay does not, keeping the two behaviors isolated.
+    renderPanel('admin', { loadPreviewState: async (seq) => ({ body: `body-${seq}` }) })
+    await screen.findByText('Draft v1')
+
+    // Delete confirm overlay is centered.
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.delete'))
+    let box = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    expect(box.closest('.octo-modal-overlay')!.classList.contains('octo-modal-overlay--center')).toBe(true)
+    fireEvent.keyDown(document, { key: 'Escape' })
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
+
+    // Restore confirm overlay is centered too.
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.restore'))
+    box = await waitFor(() => document.querySelector('.octo-version-confirm') as HTMLElement)
+    expect(box.closest('.octo-modal-overlay')!.classList.contains('octo-modal-overlay--center')).toBe(true)
+    fireEvent.keyDown(document, { key: 'Escape' })
+    await waitFor(() => expect(document.querySelector('.octo-version-confirm')).toBeNull())
+
+    // Preview/diff overlay keeps its top-anchored (non-centered) behavior — no modifier.
+    fireEvent.click(btnByText(document.querySelector('.octo-version-row')!, 'docs.version.preview'))
+    const modal = await waitFor(() => document.querySelector('.docs-version-preview-modal') as HTMLElement)
+    expect(modal.closest('.octo-modal-overlay')!.classList.contains('octo-modal-overlay--center')).toBe(false)
+  })
+
   it('does not dismiss the confirm overlay on backdrop-click or Escape while a restore is in flight (busy guard)', async () => {
     // The overlay-click and Escape cancel paths both no-op while a mutation is running, so a stray
     // backdrop click or keypress cannot tear the confirm down mid-request. Hold the restore mutation
--- a/packages/docs/src/versions/VersionHistoryPanel.tsx
+++ b/packages/docs/src/versions/VersionHistoryPanel.tsx
@@ -584,7 +584,7 @@ export function VersionHistoryPanel<TState, TCurrent>({
 
         {confirmRestore && (
           <div
-            className="octo-modal-overlay"
+            className="octo-modal-overlay octo-modal-overlay--center"
             role="presentation"
             onMouseDown={() => {
               if (!busy) setConfirmRestore(null)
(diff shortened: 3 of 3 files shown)


ploy-elison (commented):
Reviewed the full diff against head `da42bc9e` (base `b4eac474`, the merged #674). LGTM — merging is fine.

**Scope is exactly what was agreed (decision #2):**
- New scoped modifier `.octo-modal-overlay--center { align-items: safe center; }` in `editor/styles.css`. Purely additive — I diffed the base `.octo-modal-overlay` rule against `b4eac474` and it is byte-identical (still `align-items: flex-start`), so the shared overlay and every other consumer are untouched.
- Only the two destructive-confirm overlays (restore + delete) opt in via `octo-modal-overlay octo-modal-overlay--center`. No JS/behavior change beyond the className string.

**Preview/diff isolation confirmed:** the preview/compare overlay (`docs-version-preview-modal`, line 664) keeps plain `octo-modal-overlay`, so it stays top-anchored and scrollable. The other overlay consumers (SheetView, BoardShell, EditorShell manage-members) also keep the plain class — none inherit centering.

**`safe center` degradation is correct:** for taller-than-viewport content the `safe` keyword reverts to start alignment, so combined with the overlay's `overflow-y: auto` the confirm card top-anchors and scrolls, keeping the action buttons reachable rather than clipped. Sensible defensive fallback for a small card; Chromium (electron + evergreen web) supports `safe` fully.

**Regression test is solid:** asserts both confirm overlays carry `octo-modal-overlay--center` and the preview overlay does not — locking in the isolation as a class contract (jsdom doesn't lay out, so contract-level assertion is the right call). Full suite green locally: `vitest run VersionHistoryPanel.test.tsx` → 18/18 passing.

**Interface & gates:**
- No shell public-interface change (VersionHistoryPanel props/signature unchanged), so the three doc/sheet/board adapters are unaffected.
- stylelint: 0 errors; the new rule introduces no new warnings (no hex/rgba/disallowed values on the added lines).
- typecheck: the two errors surfaced (`dmworkbase/src/i18n/format.ts`, `members/sort.ts`) are pre-existing in files this PR does not touch — no new errors from these changes.

Clean, well-scoped, well-tested.

OctoBoooot (approved):
## Review: fix(versions): vertically center version-history confirm overlay (#686)

**Verdict:** Approve. A clean, correctly-scoped CSS fix with a matching contract test. No blockers, no majors.

Base `main`. Risk tier: normal. CI: Build / osv-scan / secret-scan / dependency-review all green; `code-review` PENDING (not red). `mergeStateStatus=BLOCKED` will clear once `code-review` completes.

### What's good
- **Additive, scoped modifier — the right way to do this.** Rather than touching the shared base `.octo-modal-overlay` (which several other surfaces render under — BoardShell members panel, EditorShell, SheetView, and this file's own preview/diff modal at `:664`), the fix adds a new `.octo-modal-overlay--center` modifier and applies it only to the restore (`:587`) and delete (`:626`) confirm overlays. The base class is unchanged, so no other consumer can regress. (This is exactly the isolation that a prior version-panel CSS change got wrong by editing shared styles — good to see it done additively here.)
- **`align-items: safe center` is the correct choice.** It centers the small confirm card but degrades to top-aligned if content ever overflows the viewport, so the action buttons stay reachable instead of being clipped — and the preview/diff overlay keeps its top-anchored scroll behavior. The comment documents the reasoning clearly.
- **Test asserts the right contract.** It verifies both confirm overlays carry the modifier and the preview overlay does not, and is honest that jsdom can't compute layout so it checks the class contract rather than pixels. That's the appropriate test for a CSS-modifier change, and it locks the isolation against future regressions.

### Verified
- Exactly 2 `--center` applications (restore + delete); the preview/diff overlay (`:664`) correctly does not carry it.
- Other `.octo-modal-overlay` consumers (BoardShell, EditorShell, SheetView) keep the plain base class — unaffected.

Small, focused, well-tested. Nice application of scoped-modifier over shared-style mutation.

yujiawei (approved):
# Code Review — PR #686 (octo-web)

## Summary

Small, well-scoped cosmetic fix: the version-history restore/delete confirm overlays now sit vertically centered in the viewport instead of top-anchored, while the taller preview/diff overlay retains its top-anchored, scrollable behavior. The change is a scoped CSS modifier plus a 2-line class toggle, with a regression test.

## Spec compliance

**Result: ✅ Meets the stated goal**

- **Complete**: both destructive-action confirm overlays (`confirmRestore` and `confirmDelete`) receive the new `octo-modal-overlay--center` modifier — `VersionHistoryPanel.tsx:587` and `:626`. No confirm overlay is left behind.
- **Correctly scoped**: the preview/diff overlay (`VersionHistoryPanel.tsx:664`) deliberately keeps the base `octo-modal-overlay` class only, preserving its top-anchored scroll behavior for taller-than-viewport content.
- **No over-build**: the diff touches only the three files needed for the fix — no unrelated refactors, no changes to the base overlay or to preview behavior.

## Code quality

**Result: Approved**

CSS (`packages/docs/src/editor/styles.css:1054-1066`):
- The base overlay is `display: flex` with default `flex-direction: row`, so `align-items` controls the vertical (cross) axis. `align-items: safe center` therefore centers vertically, as intended.
- The `safe` keyword is the right call here: it centers normally but degrades to start-alignment when the content would overflow, keeping the action buttons reachable rather than clipped above the viewport. This also composes well with the base `overflow-y: auto` + `padding: 48px 16px`.
- Graceful degradation: if a browser fails to parse the `safe` keyword, only this declaration is dropped and the element falls back to the base `align-items: flex-start` (top-anchored but fully usable) — no broken layout.
- The modifier is a single-class selector with the same specificity as the base and is declared after it, so source order lets it win. Good.
- The inline comment documenting the rationale (why base is top-anchored, why confirm cards center, why `safe`) is clear and appreciated.

Test (`packages/docs/src/versions/VersionHistoryPanel.test.tsx`):
- Asserts a meaningful class contract — confirm overlays carry the modifier, preview overlay does not — which guards against accidental regression of the class application. The jsdom limitation (no layout computation) is explicitly documented, so testing the class contract rather than computed alignment is the appropriate strategy at this layer.

### Non-blocking suggestions (nits, not required for merge)

- The unit test verifies class application but not actual centering; if the project already runs Playwright/visual-regression coverage elsewhere, a screenshot assertion would fully close the loop. Enhancement only, not a gap in this PR.

## Verdict

No P0/P1 issues. The change does exactly what it claims, is correctly scoped, and degrades gracefully across browsers.

Jerry-Xin (approved):
Reviewed at `da42bc9eced2`, byte-verified — with specific attention to the #674 shared-class regression risk (that PR mutated a global confirm-box selector shared by two live components). **This fix is the textbook-correct opposite of that antipattern. APPROVE.**

## ✅ Scoped, additive — NOT the #674 global-mutation pattern
`styles.css:1064-1066` adds a **brand-new opt-in modifier class**, mutating nothing shared:
```css
.octo-modal-overlay--center { align-items: safe center; }
```
- Base `.octo-modal-overlay` (`align-items: flex-start`, for the scrollable preview/diff modal) — **byte-untouched**.
- `.octo-version-confirm` (the rule at the center of #674) — **byte-untouched; this PR doesn't reference it at all**.

## ✅ Blast radius = exactly 2 intended sites (git grep verified)
`octo-modal-overlay--center` is applied only at `VersionHistoryPanel.tsx:587` (restore confirm) + `:626` (delete confirm), both in the single shared shell. No other component carries the new class → **zero regression risk to other consumers.** The base-class consumers (BoardShell/EditorShell/SheetView + the preview overlay at VersionHistoryPanel:664) do NOT get `--center`, so they keep top-anchored behavior. ✅

## ✅ #674 dual-renderer condition no longer exists
`VersionPanel.tsx` (still live-mounted by `EditorShell.tsx:751`) renders **no inline confirm/overlay of its own** (`git grep octo-version-confirm|octo-modal-overlay -- VersionPanel.tsx` → empty) — it now wraps the shared `VersionHistoryPanel`. So there's no separate inline confirm to regress this time.

## ✅ Test asserts the real contract
`VersionHistoryPanel.test.tsx:343-366` asserts both confirm overlays carry `.octo-modal-overlay--center` and the preview/diff overlay does NOT (`.toBe(false)`). Correctly asserts the class contract (jsdom can't compute layout) rather than a hollow render — the right thing to assert here.

## 🔵 Non-blocking nit
`align-items: safe center` (styles.css:1065) has no plain `center` fallback. All modern Chromium/FF/Safari support the `safe` keyword; if a parser ever rejects the two-keyword value, the declaration is dropped and the element falls back to the base `flex-start` — i.e. degrades to the *pre-fix* top-anchored state, not a broken layout. Graceful degradation → genuinely non-blocking. A preceding `align-items: center;` would be belt-and-suspenders.

APPROVE — scoped, provably 2-line blast radius, shared rules byte-identical, #674 risk specifically avoided, test asserts the real contract.

PR #569, opened 2026-07-09
feat(dmworkbase): octo/v2 rich inputs (Input.Number/Date/Time) + card_version 1.6 cap

an9xyz:
## Summary

Client half of **octo-server P3-3** (server [#556](https://github.com/Mininglamp-OSS/octo-server/pull/556) extends the octo/v2 input whitelist from 3 to 6), plus a forward-compat bump of the client `card_version` cap to match the AdaptiveCards SDK.

Both changes are additive and keep the octo policy layer fail-closed. Follows #562 (SDK-based rendering).

## What's included

**1. Rich inputs — `Input.Number` / `Input.Date` / `Input.Time`** (all Adaptive Cards 1.0 elements; stay within `card_version` 1.5)
- `validateCardForOcto`: accept the three new inputs in the `allowInteractive` (octo/v2) branch — id required + frame-unique, same discipline as the existing inputs; rejected in octo/v1 as before.
- Also validate an input's **`inlineAction`** (the in-field inline button) through the action rules — id frame-unique, `Action.OpenUrl` url safe, non-whitelist action → whole-card fallback, counted against the budget. Closes a pre-existing gap (inlineAction was never walked for any input) and mirrors the server's inlineAction dispatch.
- `octoSerialization`: stop unregistering `Input.Number/Date/Time` from the SDK element registry so the SDK renders them natively (number spinner / date / time pickers).
- `css`: baseline styling for `ac-numberInput` / `ac-dateInput` / `ac-timeInput`.
- Submit collection unchanged (`getAllInputs`). Server remains authoritative for value validation (finite Number + min/max, Date/Time format + range).

**2. `card_version` cap 1.5 → 1.6** (align to the official SDK's max supported schema)
- Only widens **version negotiation**: a future server envelope bump to `"1.6"` needs no client release.
- Does **not** widen the element whitelist — `validateCardForOcto` stays fail-closed on the octo subset. A `"1.6"` card using octo elements renders; a 1.6-only element (e.g. `Carousel`) still whole-card-degrades to plain.

## Test plan

- [x] `pnpm test` (InteractiveCard): **14 files / 147 cases** green.
- [x] New inputs accepted in v2 / rejected in v1 / missing-id fallback; inlineAction (valid Submit, Execute → fallback, duplicate-id → fallback); element registry now allows the three; e2e render of native number/date/time controls.
- [x] `card_version` 1.6 accepted (2.0 still rejected); 1.6 + octo elements → card, 1.6 + non-whitelist element → plain.
[...]

PR #688, opened 2026-07-13
fix(global-search): 4 filter-panel UX bugs (YUJ-30, front end)

### LINKED ISSUE #687: Global search filter — 4 UX bugs (YUJ-30, front end)
# Global search filter — 4 UX bugs (YUJ-30, front end)

Bundled front-end fixes for four independent bugs on the global-search
filter panel. The paired back-end work is being submitted as a separate
PR against `octo-server`. Each bug is captured in its own front-end
commit for easier review.

## The four bugs

1. **Chip loses group name after reopen.** Selected channel chips
   fell back to the raw `channelId` when the filter panel was closed
   and reopened, because the ref only persisted `{channelId,
   channelType}` and the per-panel channel catalog rebuilt empty on
   remount.

2. **「所在群聊」 candidate list included DMs.** Private conversations
   (`channelType=1`) surfaced alongside groups + threads, and the label
   itself was misleading once thread rows were mixed in. The candidate
   pool should be groups + threads only, and the label reads 「所在群聊
   或子区」 / "In group or thread" in the two locales.

3. **Messages tab did not fire with an empty keyword.** The FE gate
   `shouldRunGlobalSearch` short-circuited on `messages + empty
   keyword + no active filter`. Aligning with per-channel `_search_all`
   browse behavior, the gate is unconditional across both tabs; the
   paired backend PR lifts the corresponding server-side 400 guard.

4. **「包含成员」 was single-select.** Users could not express
   «channels that contain Alice AND Bob at the same time». The control
   is now multi-select with AND / intersection semantics, and the wire
   field migrates `member_uid` → `member_uids` (plural). Under the
   YUJ-30 unified rule, member-matched groups expand to «group + all
   its threads» server-side.

## Scope + non-scope

**In scope:** filter-panel UX, filter-state shape, wire-body
composition, i18n (zh + en synchronized), unit-test updates.
**Not in scope:** back-end handler changes (paired PR handles bugs
3/5 server-side), the group-scoped thread-picker (deferred beyond
v1), any changes outside `packages/dmworkbase/src/Components/
GlobalSearch/`.

## Testing

- `apiAdapter.test.ts` + `filterState.test.ts` (44 tests) all pass,
  including new coverage for the `member_uids` migration (self-drop +
  dedup) and the widened `shouldRunGlobalSearch` predicate.
- Pre-existing failures on `filterSearchSelect.test.tsx` and
  `PersonaSettings.test.tsx` reproduce on unmodified `upstream/main` —
  they stem from a React 17/18 renderer mismatch in the test env, not
  from this change.

## Verification

Panel-state screenshots (bug-by-bug) live in the design-doc repo at
`global-search-design/screenshots-fe-yuj30/`. Live-app screenshots
were not attempted because the dev build needs a login against the
shared test backend and there is no in-repo Storybook harness for
`GlobalSearchFilterPanel` that renders without `WKSDK`.

## Related

- Paired back-end PR (on `octo-server`) delivers bug-3 empty-keyword
  server support + bug-5 `member_uids` (plural) contract + bug-4
  unified 「group + threads」 scope.
- Design-doc updates (`§3` / `§6`) are being committed alongside.

BrDing-Rookie:
# Global search filter — 4 UX bugs (YUJ-30)

Four independent bugs on the global-search filter panel. The paired backend
PR (`fix/global-search-filter-4bugs-be` on `octo-server`) is landing in
parallel and delivers the two backend halves this PR needs (bug 3 empty-
keyword browse + bug 5 `member_uids` plural wire field). Each bug is its
own commit for easier review.

## ⚠️ Deploy order

This PR **must not deploy before** the paired backend PR [`octo-server#569`](https://github.com/Mininglamp-OSS/octo-server/pull/569) reaches production. Two changes here rely on backend contracts introduced there:

1. **Bug 3** — `shouldRunGlobalSearch` now returns `true` unconditionally, so the messages tab fires an empty-keyword browse on mount. Against pre-#569 backends this returns `400 "Invalid search request."`.
2. **Bug 5** — `cleanGlobalFilters` emits `member_uids` (plural). Pre-#569 backends only recognize `member_uid` (singular); the field is silently ignored by `c.BindJSON` (no `DisallowUnknownFields`), so the member filter appears active in the UI but returns unfiltered results.

**Backend-side backward compat**: `octo-server#569` keeps the legacy `MemberUID string` field alongside the new `MemberUIDs []string` and normalizes single→multi on the way in. So a newer backend still accepts pre-#688 frontends. Direction that breaks is **only** newer-frontend → older-backend.

**Merge/deploy protocol**: merge #569 first, wait for prod deploy, then merge #688. Or coordinate a same-window release. Neither PR carries a runtime capability flag — feedback welcome if reviewers want one added instead.

---

## Bug 1 · Selected channel chip loses the group name after reopen

`draft.channels` only persisted `{channelId, channelType}` and the per-
panel `channelCatalog` rebuilt empty on remount, so a chip fell back to
the raw `channelId` until an async candidate refetch happened to re-
hydrate the same entry.

`GlobalSearchChannelRef` now carries `name` + `avatarUrl`; the chip
resolves immediately on reopen, mirroring how ChannelSearch's sender
chip resolves through the long-lived dataSource sender cache.

## Bug 2 · Exclude DMs from 「所在群聊」, rename to 「所在群聊或子区」

`loadReadableChannelOptions` surfaced private conversations
(`channelType=1`) alongside groups + threads even though a DM never
matches a group-scoped filter server-side. The label was also misleading
once threads landed in the pool.

- Pool: groups (2) + threads (5) only.
- i18n: 「所在群聊」→「所在群聊或子区」 (zh) and "In conversation" →
  "In group or thread" (en) — both locale files updated in the same commit.

## Bug 3 · Empty-keyword browse on the messages tab

`shouldRunGlobalSearch` short-circuited on `messages` + empty keyword +
no active filter, because the backend used to 400 on fully-empty global
message queries. The paired backend PR lifts that guard (aligning with
per-channel `_search_all` browse behavior); the FE gate is now
unconditional across both tabs.

Allowlist batching (§6.2) keeps browse cost bounded; the panel's own
300 ms debounce still coalesces rapid input changes.

## Bug 5 · 「包含成员」 → multi-select (AND semantics)

`memberUid?: string` was single-select, and re-picking replaced the
existing chip — so users could not express «channels that contain
Alice AND Bob at the same time».

- Type: `memberUid?: string` → `memberUids: string[]`.
- Wire: `member_uid` → `member_uids` (self stripped + deduped + omitted
  when empty). The paired backend PR authoritatively defines the plural
  form; if both PRs land at once neither field name is orphaned in
  request/response contracts.
- Semantics: AND / intersection — a channel matches only when every
  listed member is present. Under the YUJ-30 unified rule («in-group
  filtering = group + all threads»), member-matched groups expand to
  «group + all its threads» server-side via `expandGroupWithThreads`.
  DM candidates are only valid for a single-member selection.
- Panel: shares the same multi-select chip UX as 「发送者」.

## Testing

- `apiAdapter.test.ts` + `filterState.test.ts` (44 tests) all pass,
  including new coverage for the wire migration (`member_uids` dedup +
  self-drop) and the widened `shouldRunGlobalSearch` predicate.
- Pre-existing failures on `filterSearchSelect.test.tsx` and
  `PersonaSettings.test.tsx` reproduce on unmodified `upstream/main` —
  they stem from a React 17/18 renderer mismatch in the test env, not
  from this change.

## Verification screenshots

Design-doc repo:
`/home/mlclaw/Projects/global-search-design/screenshots-fe-yuj30/`

- `01-bug1-chip-displays-channel-name.png`
- `02-bug2-channel-picker-no-dm.png`
- `03-bug3-empty-keyword-fires.png`
- `04-bug5-member-multiselect.png`

Live-app screenshots were not attempted for this PR: the running dev
build needs a login against the shared test backend, and there is no
in-repo Storybook harness for `GlobalSearchFilterPanel` that renders
without `WKSDK`. Evidence above is the panel-state visual + unit tests
around the wire / predicate changes.

## Design doc

`global-search-design/§3` and `§6` are being updated in the paired doc
commit to reflect the unified 「group + threads」 rule, the DM-free
pool, the empty-keyword browse behavior, and the plural `member_uids`
wire field.

Fixes #687


packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
```diff
@@ -189,18 +194,19 @@ export function hasEffectiveGlobalFilters(filters: GlobalSearchFilters) {
...
-  tab: GlobalContentTab,
-  keyword: string,
-  filters: GlobalSearchFilters
+  _tab: GlobalContentTab,
+  _keyword: string,
+  _filters: GlobalSearchFilters
 ) {
-  if (tab === "files") return true;
-  return keyword.trim().length > 0 || hasEffectiveGlobalFilters(filters);
+  return true;
```
OctoBoooot, line 209: **[major]** Backend-ordering dependency, no guard. `shouldRunGlobalSearch` now returns true unconditionally, and its consumer (GlobalContentSearchPanel.tsx:88,103) uses it to gate whether a search fires. The prior short-circuit existed because — per the comment this PR removed — the backend rejects fully-empty searches with a 400. Until the paired octo-server PR is deployed, every empty-keyword messages-tab view (mount, or clearing the box) fires and gets a 400. Resolve by merging after the backend is live, or gating behind a capability/version flag.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
```diff
@@ -134,10 +134,15 @@ export function cleanGlobalFilters(
...
-    next.member_uid = filters.memberUid;
+  // member_uids: drop self (§6.4 no-op) and dedup. Empty array → omit the
+  // field entirely so the server sees «no member filter» rather than an
+  // explicit empty list. Wire field is `member_uids` (plural), replacing the
+  // legacy single-value `member_uid` under YUJ-30 bug 5.
+  const memberUids = Array.from(
+    new Set(filters.memberUids.filter((uid) => uid && uid !== selfUid))
+  );
+  if (memberUids.length > 0) {
+    next.member_uids = memberUids;
```
OctoBoooot, line 145: **[major]** `member_uids` (plural) wire field replaces the legacy `member_uid`, with no version guard. Against the current backend (expecting `member_uid`), a member filter serializes to a field the server ignores -> filter silently does nothing (no error, wrong results). Same ordering dependency as the empty-search change: confirm the backend PR deploys first, or gate the field name behind a capability flag.

OctoBoooot (changes requested):
## Review: fix(global-search): 4 filter-panel UX bugs (YUJ-30) (#688)

**Verdict:** Request changes — the frontend code is clean and well-tested, but two of the four fixes hard-depend on a backend PR that is "landing in parallel" (not yet merged), with no feature flag to decouple them. Merging frontend-first causes user-visible breakage. This is a merge-ordering major, not a code defect.

Base `main`. Risk tier: normal. CI: `code-review` PENDING, `Build`/`osv-scan` still running (not failed).

### Major — backend dependency without a guard (bug 3 + bug 5 will break if merged first)
Two changes silently assume the paired `octo-server` PR is already deployed, and there's no feature flag / API-version guard gating either (confirmed — no flag in `apiAdapter.ts`):

1. **Empty-keyword search now always fires.** `shouldRunGlobalSearch` is now `return true` unconditionally (`apiAdapter.ts:209`), consumed at `GlobalContentSearchPanel.tsx:88,103` to gate whether a search runs. The prior code short-circuited empty-keyword message searches specifically because — per the comment this PR removed — *"the backend rejects fully-empty searches"* with a 400. The new backend accepts empty keyword, but until it's deployed, every empty-keyword messages-tab view (mount, or clearing the box) will fire and get a **400**. Failing input: open global search, messages tab, no keyword, no filters → request fires against a backend that still rejects it.
2. **`member_uids` (plural) wire field replaces `member_uid`.** `cleanGlobalFilters` now emits `next.member_uids` (`:145`) and the legacy `member_uid` is gone. Against the current backend (expecting `member_uid`), a member filter serializes to a field the server ignores → the filter **silently does nothing** (no error, wrong results).

Neither is wrong *given the new backend contract* — the risk is purely ordering. Options, any one resolves it: (a) merge this only after the backend PR is deployed (note it in the PR / block on it); (b) gate both behaviors behind a capability/version flag so the frontend degrades to the old behavior until the backend is live. Please confirm the deploy ordering or add a guard.

### Verified clean
- **`memberUid` → `memberUids` migration is complete.** Grepped every GlobalSearch consumer: the only remaining singular hits are an i18n key name (`filter.memberUid`, a lookup string) and a comment — all field access migrated to `memberUids` with correct add/remove/dedup toggle logic (`GlobalSearchFilterPanel.tsx:402-530`). No compile stragglers.
- **Tests updated correctly** — `apiAdapter.test.ts` now asserts self-drop + plural `member_uids` emission and the always-run behavior; `filterState.test.ts` migrated. Good.
- **Bug 1 (chip name persistence)** — `GlobalSearchChannelRef` now carries `name`/`avatarUrl` so a chip resolves on reopen instead of showing the raw channelId; mirrors ChannelSearch's sender-cache pattern. Sound.
- **Bug 2 (DM exclusion)** — `loadReadableChannelOptions` drops `ChannelTypePerson` (DM) from the pool; label renamed to 「所在群聊或子区」 in both locales. Coherent, and the `memberUids` AND-semantics + DM-drop rule is well-documented in `types.ts`.
- No XSS surface; the changes are data-shape and filtering only.

### Net
Good, well-tested frontend work — bugs 1 and 2 are self-contained and ready. The blocker is coordination: bugs 3 and 5 will regress (empty-search 400s, silently-dead member filter) if this lands before the backend. Confirm the merge ordering or add a version guard and this is good to go.

yujiawei (changes requested):
# Code Review — PR #688 (octo-web)

Reviewed at head `8da4e5f9709dbe282e0b8d4e02381ebb13b4d202` against merge-base with `main`. Scope: the 4 global-search filter-panel bug fixes (YUJ-30). This PR is a cross-repo change that declares a hard dependency on a paired `octo-server` PR; I reviewed both sides.

## Summary

The frontend code is clean and internally consistent — the `memberUid → memberUids` migration is complete (no lingering singular references), the DM exclusion and chip-name persistence are correct, and the 42 relevant unit tests pass. **However, two of the four fixes depend on backend contracts that are not yet live, and one of those contracts does not exist in the server codebase at all.** These make the PR unsafe to merge as-is.

## Blocking issues

### 🔴 P1 — Bug 5: the `member_uids` (plural) wire field is not supported by any backend

`cleanGlobalFilters` now emits `member_uids` (array) and removes the legacy singular `member_uid` entirely, with no backward-compat fallback:

```ts
// apiAdapter.ts:141-146
const memberUids = Array.from(
  new Set(filters.memberUids.filter((uid) => uid && uid !== selfUid))
);
if (memberUids.length > 0) {
  next.member_uids = memberUids;
}
```

The PR description states *"The paired backend PR authoritatively defines the plural form."* I checked the paired backend PR (`fix/global-search-filter-4bugs-be`) and the rest of the server repo, and this is **not the case**:

- The server request struct still declares only the singular field: `MemberUID string \`json:"member_uid,omitempty"\`` (`modules/messages_search/search_global.go`), on both `main` and the paired PR's head.
- The paired backend PR's body covers only **bug 3** (empty-keyword browse) and **bug 4** (group→threads expansion). Its only `member_uid` changes are *deletions* of validation code — it never adds `member_uids`.
- A repo-wide search of the server (all open/merged PRs and branches) finds **no `member_uids` plural field anywhere**.
- The handler binds the request with `c.BindJSON` and does not set `DisallowUnknownFields`, so an unknown `member_uids` key is **silently ignored**; `MemberUID` stays `""` and `resolveGlobalScope` applies no member filter.

**Net effect:** after this PR ships, selecting 「包含成员 / Includes member」 shows the filter as active (`activeGlobalSearchFilterCount` counts `memberUids.length > 0`) but has **zero effect on results** — the user gets silently wrong (unfiltered) results. The AND/intersection semantics and the "drop DM candidates once `memberUids.length > 1`" behavior documented in `types.ts:24-33` are also entirely unimplemented server-side. Because the singular field was removed, there is not even a degraded single-member fallback.

This needs the backend `member_uids` field to actually be implemented (and deployed) before this FE can ship, or the FE must keep emitting a form the current backend understands.

### 🔴 P1 — Bug 3: empty-keyword browse is a hard deploy-ordering dependency with no fallback

`shouldRunGlobalSearch` now unconditionally returns `true` (params renamed to `_tab/_keyword/_filters`):

```ts
// apiAdapter.ts:204-210
export function shouldRunGlobalSearch(
  _tab: GlobalContentTab,
  _keyword: string,
  _filters: GlobalSearchFilters
) {
  return true;
}
```

This was the only FE-side guard preventing a fully-empty messages query from reaching the server. The removed comment explicitly noted the backend *"rejects fully-empty searches"* with a 400. When the user switches to the messages tab with an empty box, `GlobalContentSearchPanel.tsx:88/103` now fires `runSearch()` with `keyword: ""` and default filters.

The paired backend PR (#-level: the `fix/global-search-filter-4bugs-be` branch) *does* implement empty-keyword acceptance — but that PR is **still open / not merged**. There is no feature flag or capability negotiation gating this behavior, so if this FE reaches production before the backend change is deployed, **every empty-keyword browse returns 400** (surfaced as `globalSearch.searchFailedRetry`).

Unlike bug 5, the backend fix here at least exists — so this is strictly a rollout-ordering blocker: the server PR must merge and deploy first. Please confirm the deploy plan (and ideally gate on a capability check, since there is no graceful fallback).

## Non-blocking

### P2 — No length cap on `member_uids` (asymmetric with `sender_ids`)

`sender_ids` is defensively capped (`filters.senderUids.slice(0, 50)`, line 135) but `member_uids` is sent uncapped (line 145). The multi-select UI has no selection limit either. Recommend mirroring the `sender_ids` cap (or whatever the backend's documented `member_uids` limit turns out to be) once the backend field exists. Given AND/intersection semantics, an unbounded list is a payload/cost concern rather than a correctness one, hence P2.

### nit — `hasEffectiveGlobalFilters` is now dead code

It was the sole caller-dependency of the old `shouldRunGlobalSearch` body. With the guard replaced by `return true`, `hasEffectiveGlobalFilters` (apiAdapter.ts:181) has no remaining production callers (only its definition and the test). Consider removing it or tracking a follow-up so it doesn't rot.

## What looks good

- **Bug 1** (channel chip name/avatar persisted on `GlobalSearchChannelRef`, resolved via `c.name ?? known?.name ?? c.channelId`): correct, self-contained, and the fallback chain is sound. Panel state is in-memory React state (not persisted to storage), so there is no stale-serialization migration risk from the ref shape change.
- **Bug 2** (DM exclusion from the picker pool + label rename to 「所在群聊或子区」): clean. The `ChannelTypePerson` import removal is complete (only referenced in a comment now), and both locale files are updated.
- The `memberUid → memberUids` migration is complete across `types.ts`, `filterState.ts`, `apiAdapter.ts`, and the panel; no dangling singular references remain outside comments/i18n keys. Every `memberUids` access is on a guaranteed array, and `dataSource.getSender` has a safe fallback, so unresolvable member chips can't throw.
- The dedup + self-strip logic (`Array.from(new Set(... .filter(uid => uid && uid !== selfUid)))`) is correct and matches its test.

## Security note (this PR was flagged security-sensitive)

The classification appears to be a keyword false-positive on "member". Access control is enforced server-side (the empty-keyword browse is bounded by the server's authoritative allowlist intersection; the removed FE guard was a "don't-400-the-backend" optimization, not a security boundary), and there is no client-side result filtering to bypass. No FE security defect found.

For a human to verify on the **backend** side before merge (out of scope for this FE diff but load-bearing for the safety argument):
1. `_search_global_messages` must enforce the allowlist on the empty-keyword browse path (i.e. empty keyword must not fail-open to an unbounded scan).
2. Self-uid must remain ignored server-side per §6.4.
3. The group→threads expansion (`resolveGlobalScope` / thread folding) must re-check per-thread readability rather than assuming group membership grants all thread access.

## Verdict

Two P1 blockers (bug 5's wire field has no backend implementation; bug 3 has an ungated deploy-ordering dependency) → **Changes Requested**. The bug 1 / bug 2 halves are ready; the blockers are the two cross-repo contract dependencies.

BrDing-Rookie:
@Jerry-Xin @yujiawei @OctoBoooot Addressing the merge-ordering concerns on this PR:

**Chose option A (deploy-order coordination, no runtime flag).** PR body updated with a **⚠️ Deploy order** section calling out that this PR must not deploy before `octo-server#569`.

**yujiawei P1 (`member_uids` field missing on backend)**: valid — I had missed dispatching the backend `member_uids` implementation. Fixed on the octo-server side: `octo-server#569` now adds `MemberUIDs []string` with AND semantics + sub-thread inclusion + DM single/multi handling + 6 tests (commit landing shortly, will link once pushed). Crucially the backend **keeps `MemberUID string` alongside the new plural field and normalizes single→multi**, so the compat direction is *newer-backend accepts older-frontend*; only *newer-frontend → older-backend* breaks — exactly what the deploy-order section covers.

**OctoBoooot majors (empty-keyword browse + `member_uids` plural)**: both resolved by the same deploy-order coordination. No runtime capability flag added — happy to add one if reviewers prefer that over release coordination.

Please let me know if the deploy-order approach is acceptable, or if you would like a capability flag instead.

BrDing-Rookie:
追加一个前端功能到本 PR：**主聊天框右上角新增「查找聊天内容」入口**（Search icon，与「智能总结」/「事项」同排、同风格）。

**做了什么**
- 新增按钮组件 `packages/dmworkbase/src/Components/ChannelSearch/ChatSearchEntryButton.tsx`
  - lucide `Search`，20×20，样式沿用现有 `.wk-chat-conversation-header-right-item`
  - tooltip 复用现有 i18n key `base.module.channelSettings.messageHistory`（"查找聊天内容"），未新增 key
  - `isChannelSearchEnabled(channel)` 门禁 —— 与信息栏入口一致
- 在 `dmworkbase/src/module.tsx` 里注册 `registerChannelHeaderRightItem("channelheader.search", ..., sort=4900)`，排在 matter (5000)/summary (5100) 之前
- 加 mittBus 事件 `wk:open-channel-search`（`App.tsx`）— 与 summary/matter 同构，避免直接 `showConversation` 重开会话
- `Pages/Chat/index.tsx` 加 handler：channel 匹配后调用现成的 `_openChannelSearchPanel()`（含互斥 + feature 门禁）；`componentWillUnmount` 记得 off
- 单测：按钮渲染门禁（feature-flag 关 / 非受支持频道类型不渲染）+ 点击 emit 事件（`__tests__/ChatSearchEntryButton.test.tsx`，4 passed）

**校验**
- `pnpm build` ✅
- `pnpm lint:wkmodal` ✅
- `pnpm i18n:check` ✅
- 新增单测 4/4 ✅
- 完整 vitest 套件的 22 个失败文件与本改动无关（stash 掉本改动后失败数量一致）

**Spec**
`~/Projects/global-search-design/feat-chat-header-search-entry.md`

**验收截图**
`~/Projects/global-search-design/screenshots-fe-header-search/mockup-full.png`（因 dev-server 无法稳定拉起，用样式对齐现有 `.wk-chat-conversation-header-right-item` 的 mockup 展示；真实组件走 endpoint 注册链）

Commit: d813f1ec

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
OctoBoooot, line 209: **[major]** (unchanged from prior CR) shouldRunGlobalSearch still returns true unconditionally. Consumed by GlobalContentSearchPanel.tsx to gate search firing. Until the paired octo-server PR deploys, empty-keyword messages-tab views hit a backend that returns 400. Merge after backend, or gate behind a capability flag.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
OctoBoooot, line 145: **[major]** (unchanged from prior CR) member_uids (plural) emitted with no version guard. Against the current backend (expecting member_uid), the member filter serializes to an ignored field -> silently does nothing. Gate behind a capability flag or keep back-compat with member_uid until the backend is live.

OctoBoooot (changes requested):
## Review: #688 @ `d813f1ec` — delta re-review; my two majors are unaddressed (new commit is an unrelated feature)

**Verdict:** Request changes — held. The new commit adds a separate, clean feature (chat-header search entry button) but does **not** touch either of the two backend-ordering majors from my prior CR (`8da4e5f9`). Those stand.

Base `feat/octo-loop`. `code-review` CI PENDING.

### Still open — both majors from my prior CR (verified unchanged at this head)
1. **Empty-keyword search still fires unconditionally.** `shouldRunGlobalSearch` still `return true` (`apiAdapter.ts:209`), consumed by `GlobalContentSearchPanel.tsx` to gate whether a search runs. Until the paired octo-server PR is deployed, empty-keyword messages-tab views hit a backend that (per the comment this PR removed) returns 400.
2. **`member_uids` (plural) still emitted with no version guard** (`apiAdapter.ts:145`). Against the current backend (expecting `member_uid`), the member filter serializes to an ignored field → silently does nothing.

No flag / version / capability guard was added (confirmed — none in `apiAdapter.ts`). Either merge after the backend deploys, or gate both behaviors behind a capability flag (yujiawei's suggested back-compat approach works too). This remains the reason for REQUEST_CHANGES.

### New commit — clean, but unrelated to this PR's stated scope
The delta adds a chat-header "查找聊天内容" entry button (`ChatSearchEntryButton.tsx` + `registerChannelHeaderRightItem` in `module.tsx` + a `wk:open-channel-search` event handled in `Pages/Chat/index.tsx`). Reviewed it:
- **Correct and leak-free.** The `wk:open-channel-search` listener is registered in `componentDidMount` with a channel-match guard (only responds to its own channel) and **cleaned up in `componentWillUnmount`** (`Chat/index.tsx:911-913`). Reuses `_openChannelSearchPanel` (same `isChannelSearchEnabled` gate + side-panel mutual exclusion); the button and endpoint both gate on `isChannelSearchEnabled`. 110-line test included. No defects.
- **Scope note (not blocking):** this is a new feature (a search entry point), unrelated to "4 filter-panel UX bugs (YUJ-30)." It's clean, but bundling an unrelated feature into a bug-fix PR that's already blocked on a backend dependency widens the blast radius and couples the new feature's merge to the unresolved majors. Consider splitting it into its own PR so it can land independently.

### Net
The new search-entry feature is well-built, but it doesn't address why this PR is held: the two backend-ordering majors (empty-search 400, silent member filter) are unchanged. REQUEST_CHANGES stands on those. Splitting the search-entry button into a separate PR would let that ship now while the filter fixes wait on the backend.

Jerry-Xin (commented):
Reviewed at head `d813f1ec3317`, byte-verified. **The front-end code is clean and correct on its own merits** — I found zero in-repo FE defects. The concern OctoBoooot + yujiawei raised is real, but it's a **cross-service deploy-ordering dependency**, not an FE code bug. Posting COMMENT: I concur the ordering is a genuine merge-gate (and the gate already correctly holds), while affirming the FE code quality.

## ✅ FE code — byte-verified correct
- **Bug 5 `member_uid`→`member_uids`:** clean rename, not a dual-field. `apiAdapter.ts:141-145` dedups (`Set`) + drops self + **omits the field entirely when empty** (`if (memberUids.length > 0) next.member_uids = ...`) — no empty-list sent. Singular `member_uid` fully removed from production paths (only remains in test assertions verifying `undefined` + comments). No stale-closure in the multi-select toggle (functional `cur =>` update), `clearAll`/`defaultGlobalSearchFilters` seed `memberUids:[]` so `.includes()` never hits undefined.
- **Bug 3 empty-keyword browse:** `shouldRunGlobalSearch` now `return true` (`apiAdapter.ts:204-210`), by design. Double-gated: the whole content-search panel only mounts when `isGlobalContentSearchEnabled()` = `!!remoteConfig.messagesSearchOn` (off by default) — so the empty browse only fires on a flag-enabled tenant.
- **Bugs 1 (chip name persistence) + 2 (DM exclusion):** correct — `name?`/`avatarUrl?` written on select, preferred with fallback chain `c.name ?? known?.name ?? c.channelId`; `dataSource.ts:70` drops `ChannelTypePerson`.
- **Wiring/leak/XSS:** Chat listener registered in `componentDidMount` with channel-match guard, **cleaned up in `componentWillUnmount`** (no leak); `registerChannelHeaderRightItem` is idempotent (keyed by id). No new `dangerouslySetInnerHTML` (existing ones are `sanitizeHighlight`-wrapped, not in this diff). i18n parity clean (both locales change the same key), meaningful tests (self-drop/dedup/plural-field/omit-empty/unconditional-run), no console.log/dead code/TS errors.

## ⚠️ The real concern — deploy-ordering (a merge-gate, not an FE defect)
Both peers' "🔴"s reduce to: this FE depends on **octo-server#569** being deployed first. Byte-verified this is **thoroughly documented in the PR body** (`⚠️ Deploy order` section): if the FE merges/deploys before #569, (a) empty-keyword browse hits a pre-#569 backend that returns `400`, and (b) `member_uids` is silently ignored by `c.BindJSON` (no `DisallowUnknownFields`) → member filter appears active but returns unfiltered results. The safe direction (newer-BE ← older-FE) is handled — #569 keeps legacy `MemberUID` + normalizes single→multi.

**There is no runtime capability/version flag** decoupling the two (confirmed — only the pre-existing non-version-aware `messagesSearchOn` gate). So the ordering is enforced only by human merge discipline. The author explicitly documents the protocol ("merge #569 first, wait for prod deploy, then merge #688") and **invites** adding a capability flag if reviewers prefer.

## My position
- The FE code is **approvable on quality** — it correctly implements the FE half of a documented two-PR contract, and it's explicitly scoped "front end."
- But the **deploy-ordering must be honored** (merge+deploy octo-server#569 to prod first, or same-window release). If you want defense-in-depth against a wrong-order merge, taking the author up on the offered **runtime capability flag** would be the more robust option than relying on merge discipline.
- I'm posting COMMENT rather than APPROVE because the merge-readiness genuinely hinges on that external ordering, and rather than CHANGES_REQUESTED because there is no in-repo FE defect to fix — the code is correct given its (documented) backend contract. The gate is already correctly held at CHANGES_REQUESTED by the peer reviews + `needs-human-review`, so this is not a false-green.

Net: FE code ✅ clean; merge only after octo-server#569 is in prod (or add the offered capability flag). Concurring with OctoBoooot + yujiawei that the ordering is the gating concern.

yujiawei (changes requested):
# Code Review — PR #688 (octo-web)

Reviewed at head `d813f1ec33178aa0c9f9844500e67cf7490bf41c` against merge-base `b4eac474` with `main`. This is a re-review: the head has advanced by one commit (`d813f1ec`) since the previous round. The 4 filter-panel bug fixes are cleanly implemented, but the PR is not mergeable as-is for two independent reasons: an out-of-scope feature was added, and two of the fixes hard-depend on an unmerged/undeployed backend PR.

**Security note:** the `security_sensitive` classification looks like a keyword false-positive on "member". Access control is enforced server-side (the empty-keyword browse is bounded by the server's authoritative allowlist intersection; the removed FE guard was a "don't-400-the-backend" optimization, not a security boundary). No FE security defect found. Backend items a human should still confirm are listed at the end.

---

## 1. Spec compliance

**Spec: ❌**

- **Missing:** none — bugs 1, 2, 3, 5 are all implemented.
- **Over-build:** ✗ commit `d813f1ec` "feat(chat-header): add search entry button" adds a **new feature** unrelated to the four filter-panel bugs. It introduces a chat-header search icon that opens the in-channel search panel via a new `wk:open-channel-search` mittBus event, touching files well outside the filter panel:
  - `packages/dmworkbase/src/App.tsx` (new `MittEvents` entry)
  - `packages/dmworkbase/src/Components/ChannelSearch/ChatSearchEntryButton.tsx` (new, +32) and its test (+110)
  - `packages/dmworkbase/src/Pages/Chat/index.tsx` (listener register/cleanup, +19)
  - `packages/dmworkbase/src/module.tsx` (endpoint registration, +19)

  The linked issue #687 explicitly states the non-scope: *"any changes outside `packages/dmworkbase/src/Components/GlobalSearch/`."* This commit spans app-level plumbing outside that boundary, is tagged `feat(chat-header)` (not `fix(global-search)`), and references a **different** design doc (`feat-chat-header-search-entry.md`). It is ~58% of the diff's additions and is logically independent — it belongs in its own PR.

  To be clear: the feature code itself is sound (see §2 — endpoint API exists, feature-gated, listener cleaned up, i18n key reuses the existing entry). The blocker is scope, not quality. Please split it into a separate PR so this one stays reviewable against #687.

- **Divergence:** none.

---

## 2. Code quality

### P1 — Bug 3 empty-keyword browse: ungated deploy-ordering dependency, no fallback

`shouldRunGlobalSearch` now unconditionally returns `true`:

```ts
// apiAdapter.ts:204-210
export function shouldRunGlobalSearch(_tab, _keyword, _filters) {
  return true;
}
```

This removes the only FE-side guard that prevented a fully-empty messages query from reaching the server. I confirmed the current production backend (`main`) still rejects it: `search_global_messages.go` calls `validateSearchNotEmptyGlobal(c, req.Keyword, req.Filters)` on the empty-keyword + no-filter path → HTTP 400 (surfaced to the user as `globalSearch.searchFailedRetry`). The backend change that lifts this guard lives in the paired `octo-server#569`, which is **still open (not merged, mergeStateStatus BLOCKED)**. There is no capability/version check in `apiAdapter.ts`, so if this FE reaches production before that backend is deployed, every empty-keyword messages-tab view (mount, or clearing the box) 400s. Request only fires on the active tab and is debounced 300 ms (`GlobalContentSearchPanel.tsx`: `SEARCH_DEBOUNCE_MS`, `canSearch = shouldRunGlobalSearch(...) && isActive`), so this is a correctness/rollout blocker, not a request-spam one.

### P1 — Bug 5 `member_uids` (plural) wire field: deploy-ordering dependency

`cleanGlobalFilters` now emits `member_uids` and drops the legacy singular `member_uid` entirely, with no backward-compat fallback:

```ts
// apiAdapter.ts:141-146
const memberUids = Array.from(
  new Set(filters.memberUids.filter((uid) => uid && uid !== selfUid))
);
if (memberUids.length > 0) {
  next.member_uids = memberUids;
}
```

I verified the backend both ways:
- `octo-server` **`main`** (current prod) declares only `MemberUID string \`json:"member_uid,omitempty"\``. The handler binds with `c.BindJSON` and does not set `DisallowUnknownFields`, so an unknown `member_uids` key is **silently dropped**; `MemberUID` stays `""` and no member filter is applied. Net effect against prod: the 「包含成员」 filter shows as active in the UI (`activeGlobalSearchFilterCount` counts `memberUids.length > 0`) but returns **unfiltered** results — silent wrong data, no error.
- `octo-server#569` head **does** implement it: `MemberUIDs []string`, `normalizeMemberUIDs` (folds legacy single → plural, dedup, self-exclude), and `channelsForMembers` with the AND/intersection semantics. So once #569 is deployed this is correct.

So this is a rollout-ordering blocker identical in shape to bug 3: the backend must merge and deploy first, or the FE needs a capability check / degraded fallback. (This corrects the previous round's note — the plural field does now exist server-side; it is just not yet live.)

**Either P1 is resolved by any one of:** (a) confirm and enforce the merge/deploy ordering (backend PR deployed first — the PR body proposes this but nothing enforces it); or (b) add a capability/version guard so the FE degrades to the old behavior against an older backend.

### P2 — `memberSelected` chip avatar can be `undefined` for an uncached member

```ts
// GlobalSearchFilterPanel.tsx:519-525
draft.memberUids.map((uid) => {
  const s = dataSource.getSender(uid);
  return { id: uid, name: s.name, avatarUrl: s.avatarUrl };
})
```

`dataSource.getSender` falls back to `{ uid, name: uid }` (no `avatarUrl`) for an uncached uid, so a member chip can render without an avatar. Not a crash (`name` is always defined via the fallback, and the renderer guards with `opt.avatarUrl && `), and it mirrors the pre-existing `senderSelected` pattern — hence P2. Optional: `avatarUrl: s.avatarUrl || WKApp.shared.avatarUser(uid)`.

### nit — `hasEffectiveGlobalFilters` is now dead code

With `shouldRunGlobalSearch` hardcoded to `true`, its sole former caller-dependency `hasEffectiveGlobalFilters` (`apiAdapter.ts:181`) has no remaining production callers (only its definition + a test). Remove it or file a follow-up so it doesn't rot.

### Verified clean
- **`memberUid` → `memberUids` migration is complete.** Full-tree grep finds no remaining singular field access — the only `memberUid` hits are the i18n key `globalSearch.filter.memberUid` (a lookup string) and comments. `defaultGlobalSearchFilters` includes `memberUids: []`, and every access is on a guaranteed array. Dedup + self-strip logic is correct and matches its test.
- **Bug 1** (chip name/avatar persisted on `GlobalSearchChannelRef`): correct. `loadReadableChannelOptions` always populates `name`/`avatarUrl` on each option, so persisting `opt.name`/`opt.avatarUrl` is a real fix (not a no-op), and the `c.name ?? known?.name ?? c.channelId` fallback chain is sound. Panel state is in-memory React state, so there's no stale-serialization migration risk from the ref shape change.
- **Bug 2** (DM exclusion + label rename to 「所在群聊或子区」): clean. `ChannelTypePerson` is dropped from the pool and both locale files are updated.
- **New chat-header feature (code quality only)**: `registerChannelHeaderRightItem` exists (`EndpointCommon.tsx:245`); the button is feature-gated by `isChannelSearchEnabled` at both the registration site and inside the component; the `wk:open-channel-search` listener is registered in `componentDidMount` and removed in `componentWillUnmount`, mirroring the sibling `_onToggleSummaryPanel`; and because `ChatContentPage` is keyed by `channel.getChannelKey()` (`EndpointCommon.tsx:168`) it fully remounts per channel, so the `channel`-closure identity check is always fresh (no stale-closure bug). i18n title reuses the existing `base.module.channelSettings.messageHistory` key. Sound — it just doesn't belong in this PR (§1).
- 46 relevant unit tests pass (`apiAdapter.test.ts`, `filterState.test.ts`, `ChatSearchEntryButton.test.tsx`).

---

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ❌ (out-of-scope chat-header feature) AND Quality Changes-Requested (two deploy-ordering P1s with no FE guard).

## 4. Recommended actions
1. Move commit `d813f1ec` (chat-header search entry) into its own PR so this one matches issue #687's scope.
2. Resolve the backend dependency for bugs 3 & 5 by **either** enforcing merge/deploy ordering (`octo-server#569` deployed to prod first) **or** adding a capability/version guard so the FE degrades gracefully against an older backend.
3. (P2/nit) Add an avatar fallback in `memberSelected`; remove the now-dead `hasEffectiveGlobalFilters`.

## 5. For a human to verify on the backend before merge
1. `_search_global_messages` must enforce the allowlist on the empty-keyword browse path (empty keyword must not fail-open to an unbounded scan).
2. Self-uid must remain ignored server-side (§6.4).
3. The group→threads expansion (`expandGroupWithThreads`) must re-check per-thread readability rather than assuming group membership grants all thread access.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
OctoBoooot, line 209: **[major]** (still open from prior CR) shouldRunGlobalSearch still returns true unconditionally → empty-keyword messages-tab views hit the pre-#569 backend and get a 400. Merge after backend deploys, or gate behind a capability flag.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
OctoBoooot, line 145: **[major]** (still open from prior CR) member_uids emitted with no version guard; against the current backend (expecting member_uid) the member filter is silently ignored. Add back-compat with member_uid or gate behind a capability flag until the backend is live.

OctoBoooot (changes requested):
## Review: #688 @ `da0e9e56` — delta re-review; scope-creep removed (good), but both majors still stand

**Verdict:** Request changes — held. The author took the scope note and **removed the unrelated chat-search-entry feature entirely** (good — PR is now cleanly the 4 filter-panel bugs). But neither of the two backend-ordering majors from my prior CR (`d813f1ec`) is addressed; they're unchanged at this head.

Base `feat/octo-loop`. `code-review` CI PENDING.

### Resolved — scope-creep
The delta is all deletions: `ChatSearchEntryButton.tsx` + its test, the `wk:open-channel-search` event, and the `Chat/index.tsx` / `module.tsx` wiring are all reverted. PR scope is now the 9 GlobalSearch filter files + i18n — matching the title. Nice — that unblocks the search-entry feature to land on its own schedule.

### Still open — both majors (verified unchanged at `da0e9e56`)
1. **Empty-keyword search still fires unconditionally** — `shouldRunGlobalSearch` still `return true` (`apiAdapter.ts:209`). Until the paired octo-server PR deploys, empty-keyword messages-tab views hit a backend that returns 400.
2. **`member_uids` (plural) still emitted with no version guard** (`apiAdapter.ts:145`). Against the current backend (expecting `member_uid`), the member filter serializes to an ignored field → silently no-ops. (`:140` is just a comment about the legacy field — no back-compat code.)

No capability flag / version guard was added. This remains the reason for REQUEST_CHANGES: either merge after the backend PR deploys, or gate both behaviors behind a capability flag (yujiawei's back-compat-with-`member_uid` suggestion also works).

### Net
Good responsiveness on the scope split. The verdict is unchanged because the two deploy-ordering majors — the actual reason this is held — are untouched. Confirm the backend deploy ordering or add a guard, and this is ready.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
BrDing-Rookie: Update: `octo-server#569` was merged at 2026-07-13T09:19Z (merge commit `747a76dfd63c1da4ceee574f95615f9174ccd921`). The deploy-order concern is now narrowed: this FE must not deploy to production before the `octo-server` production deploy of that commit lands.

Backend keeps `MemberUID` as a legacy field with normalize-single→multi, so a newer backend accepts pre-#688 frontends. Only newer-frontend → older-backend still breaks; that direction is covered by the merge-then-deploy protocol in the PR body's ⚠️ Deploy order section.

Not adding a runtime capability flag per mlclaw's decision — deploy coordination is sufficient given the timing (backend is already merged).

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
BrDing-Rookie: Update: `octo-server#569` was merged at 2026-07-13T09:19Z (merge commit `747a76dfd63c1da4ceee574f95615f9174ccd921`). The deploy-order concern is now narrowed: this FE must not deploy to production before the `octo-server` production deploy of that commit lands.

Backend keeps `MemberUID` as a legacy field with normalize-single→multi, so a newer backend accepts pre-#688 frontends. Only newer-frontend → older-backend still breaks; that direction is covered by the merge-then-deploy protocol in the PR body's ⚠️ Deploy order section.

Not adding a runtime capability flag per mlclaw's decision — deploy coordination is sufficient given the timing (backend is already merged).

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
BrDing-Rookie: Update: `octo-server#569` was merged at 2026-07-13T09:19Z (merge commit `747a76dfd63c1da4ceee574f95615f9174ccd921`). The deploy-order concern is now narrowed: this FE must not deploy to production before the `octo-server` production deploy of that commit lands.

Backend keeps `MemberUID` as a legacy field with normalize-single→multi, so a newer backend accepts pre-#688 frontends. Only newer-frontend → older-backend still breaks; that direction is covered by the merge-then-deploy protocol in the PR body's ⚠️ Deploy order section.

Not adding a runtime capability flag per mlclaw's decision — deploy coordination is sufficient given the timing (backend is already merged).

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
BrDing-Rookie: Update: `octo-server#569` was merged at 2026-07-13T09:19Z (merge commit `747a76dfd63c1da4ceee574f95615f9174ccd921`). The deploy-order concern is now narrowed: this FE must not deploy to production before the `octo-server` production deploy of that commit lands.

Backend keeps `MemberUID` as a legacy field with normalize-single→multi, so a newer backend accepts pre-#688 frontends. Only newer-frontend → older-backend still breaks; that direction is covered by the merge-then-deploy protocol in the PR body's ⚠️ Deploy order section.

Not adding a runtime capability flag per mlclaw's decision — deploy coordination is sufficient given the timing (backend is already merged).

BrDing-Rookie:
@OctoBoooot @yujiawei @Jerry-Xin Status update:

**1. Header search entry split (yujiawei §1 spec / OctoBoooot bundling)** — Reverted from this PR in `da0e9e56` (fast-forward, no force-push). Landed as an independent PR against a fresh branch from `main`:

- Issue: #694
- PR: #695 (`feat/chat-header-search-entry`)

This PR now stays scoped to the 4 filter-panel bugs per issue #687.

**2. Deploy-ordering (OctoBoooot §majors + yujiawei §2 P1s)** — `octo-server#569` was merged at 2026-07-13T09:19Z (merge commit `747a76dfd63c1da4ceee574f95615f9174ccd921`). The only remaining ordering constraint is that the `octo-server` **production deploy** of that commit must land before this FE is deployed to prod. The ⚠️ Deploy order section in the PR body already documents that protocol. Per mlclaw's decision, no runtime capability flag will be added — deploy coordination is sufficient given the timing (backend already merged, and backend keeps the legacy `MemberUID` field with normalize-single→multi so a newer-backend still accepts pre-#688 frontends; only newer-FE → older-BE breaks, which is what the deploy-order protocol covers).

Individual replies also posted on the two `apiAdapter.ts` line-comment threads (:145 and :209, both rounds).

**3. P2 avatar fallback (yujiawei §2 P2)** — Acknowledged, keeping consistent with the pre-existing `senderSelected` pattern for now (`GlobalSearchFilterPanel.tsx:472-479` does the same `getSender(uid)` and inherits the same `undefined` avatar path for uncached uids; the renderer guards with `opt.avatarUrl && ` so there is no crash risk). Happy to add `s.avatarUrl || WKApp.shared.avatarUser(uid)` in a follow-up if you'd prefer it landed here — say the word.

**4. nit — dead `hasEffectiveGlobalFilters`** — Will be handled in a follow-up so it doesn't inflate the diff on this PR.

Ready for re-review.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
BrDing-Rookie: Update: backend (octo-server main including #569 merge commit `747a76dfd6...`) has been deployed to production per mlclaw confirmation. The deploy-ordering concern is fully resolved — both empty-keyword browse and `member_uids` (plural) wire field are live server-side. Resolving thread.

packages/dmworkbase/src/Components/GlobalSearch/apiAdapter.ts
BrDing-Rookie: Update: backend (octo-server main including #569 merge commit `747a76dfd6...`) has been deployed to production per mlclaw confirmation. The deploy-ordering concern is fully resolved — both empty-keyword browse and `member_uids` (plural) wire field are live server-side. Resolving thread.

BrDing-Rookie:
@OctoBoooot @yujiawei @Jerry-Xin Deploy-ordering fully cleared: `octo-server` production deploy of #569 (merge commit `747a76dfd6...`) is confirmed live. All 6 deploy-ordering review threads on `apiAdapter.ts:145` and `apiAdapter.ts:209` have been resolved.

Summary of status:
- **Header search entry** (yujiawei out-of-scope, OctoBoooot bundling): split to independent PR #695 (already APPROVED).
- **Deploy-ordering** (both majors from OctoBoooot + yujiawei P1): resolved by backend merge (09:19Z) + prod deploy (confirmed).
- **P2 avatar fallback** (yujiawei nit): acknowledged, keeping consistent with pre-existing `senderSelected` guarded-render pattern.

Ready for re-review / merge.

yujiawei (approved):
# Code Review — PR #688 (octo-web)

Independent review of the global-search filter-panel fixes (bugs 1, 2, 3, 5). Reviewed the full diff at the head SHA plus the surrounding files (`GlobalContentSearchPanel.tsx`, `dataSource.ts`, `FilterSearchSelect.tsx`) to trace data flow and cross-file impact, ran the touched tests, and checked the field-rename for completeness across the package.

## Verdict: Approve

No correctness or security defects were found in the diff. The four fixes are implemented as described and the `memberUid` → `memberUids` migration is complete and type-clean. One operational gate (deploy ordering) and a couple of optional suggestions are noted below — none block merge on code grounds.

---

## 1. Scope / spec

All four bugs are implemented and nothing out-of-scope was added:

- **Bug 1** — `GlobalSearchChannelRef` now persists `name` + `avatarUrl`; the chip resolves via `c.name ?? known?.name ?? c.channelId` (`GlobalSearchFilterPanel.tsx:492-503`), so a picked channel keeps its display name after the panel is closed and reopened. ✅
- **Bug 2** — DMs (`channelType=1`) removed from the candidate pool in `dataSource.ts:70-77`; label updated to 「所在群聊或子区」 / "In group or thread" in both locale files. The now-unused `ChannelTypePerson` import is correctly dropped. ✅
- **Bug 3** — `shouldRunGlobalSearch` returns `true` unconditionally (`apiAdapter.ts:204-210`); the consumer gate `canSearch = shouldRunGlobalSearch(...) && isActive` and the `runSearch` guard remain intact, and `filters`/`keyword`/`tab` stay in the debounce dependency arrays, so no hook staleness is introduced. ✅
- **Bug 5** — `memberUid?: string` → `memberUids: string[]`, wire field `member_uid` → `member_uids`. `cleanGlobalFilters` self-drops and dedups: `Array.from(new Set(filters.memberUids.filter(uid => uid && uid !== selfUid)))` (`apiAdapter.ts:141-145`), and omits the field entirely when empty. ✅

The head commit contains a `feat(chat-header)…` commit and its immediate revert; the net diff is exactly the four fixes plus i18n — no stray behavioral change.

## 2. Field-rename completeness (verified)

- No remaining `filters.memberUid` reads anywhere in `packages/`. The only `memberUid` occurrences are the i18n **key name** (`globalSearch.filter.memberUid`) and its `t()` lookup — a key name, not the field — plus doc comments.
- `setMemberUid` fully removed; `toggleMemberById` → `toggleMember` internally.
- `defaultGlobalSearchFilters()` includes `memberUids: []`, and both runtime construction sites (`initialState?.filters || defaultGlobalSearchFilters()` and the `{...draft}` spread in `apply`) always carry `memberUids`, so `.includes()` / `.length` never hit `undefined`. `index.tsx` does not pass `initialState`, so no old-shape persisted filter object can be resurrected.

## 3. Tests

The two PR-touched suites pass: `apiAdapter.test.ts` + `filterState.test.ts` (42/42), including the new `member_uids` dedup/self-drop and multi-select count cases.

> Note: `filterSearchSelect.test.tsx` shows 9 failures locally, but that component **and** its test are untouched by this PR and fail identically with a jsdom render error (`Unable to find role "textbox"`, component renders an empty `<div />`) — a pre-existing/environmental issue, not a regression from this change. Worth a separate look, but out of scope here.

## 4. Deploy gate (merge condition — operational, not a code defect)

The PR body already documents this and it's accurate: this frontend **must not deploy before** the paired backend `octo-server#569`. Two paths depend on it:

- Bug 3 fires an empty-keyword browse on mount; a pre-#569 backend returns `400`.
- Bug 5 emits `member_uids`; a pre-#569 backend silently ignores it (no `DisallowUnknownFields`), so the member chip renders as active but results come back unfiltered — a fail-open UX condition. (No cross-user leak: allowlist intersection is enforced by every backend version, so a user only ever browses their own authorized corpus.)

There is no runtime capability flag to discriminate the unsafe pre-#569 state — the only gate is the shared `messagesSearchOn` remote flag, which says nothing about #569. So please enforce the merge/deploy ordering out-of-band (merge #569, wait for prod, then this). The code itself is correct; this is purely a sequencing constraint.

## 5. Suggestions (non-blocking)

- **Backward-compat during rollout**: to shrink the fail-open window, consider emitting both `member_uid` (first uid) and `member_uids` for one release, then dropping the singular. Optional given the documented deploy ordering.
- **Member chip name persistence**: `memberSelected` resolves names via `dataSource.getSender(uid)`, which falls back to `{ uid, name: uid }` on a cache miss — so a reopened panel with a picked member not in the sender cache shows the raw uid. This is the same class of issue bug 1 fixes for channels; the single-`memberUid` code had the identical reliance, so it's pre-existing, not introduced here. Applying the bug-1 name-persistence pattern to member refs would close the loop. (No crash risk — the fallback is safe.)

## 6. Coverage / blind spots

- Backend `octo-server#569` is not visible from this repo. The AND/intersection semantics for `member_uids`, the DM-drop for N>1, the allowlist intersection, and the empty-keyword browse cost/rate-limiting are all backend-side guarantees asserted in type comments but **unverifiable from this PR** — recommend confirming them against #569 at the deploy gate.
- A second independent reviewer leg was attempted but its runtime was unavailable; this review reflects one primary analysis plus a cross-model advisory pass.

Jerry-Xin (commented):
Re-reviewed at head `da0e9e56987f`, byte-verified the delta. **COMMENT** (updated). The author addressed the scope-creep; the backend deploy-ordering remains the gating concern (unchanged substance from my prior COMMENT).

## ✅ Scope-creep resolved
The delta `d813f1ec3317 → da0e9e56987f` is **-182, all deletions** removing the unrelated chat-header search-entry feature — `ChatSearchEntryButton.tsx` (-32) + its test (-110) + the `Chat/index.tsx` (-19), `module.tsx` (-19), `App.tsx` (-2) wiring. Byte-verified **fully removed, no dangling refs/imports** (`grep ChatSearchEntryButton` = empty). The PR is now cleanly scoped to the 4 filter-panel bugs (YUJ-30) — exactly the split OctoBoooot, yujiawei, and I flagged. Good.

## ✅ FE filter code — still clean (unchanged at this head)
The filter logic is byte-identical to what I verified at `d813f1ec3317`: `shouldRunGlobalSearch → return true` (`apiAdapter.ts:209`, double-gated behind `isGlobalContentSearchEnabled()`=`messagesSearchOn`, default off), `member_uids` deduped + self-dropped + **omitted when empty** (`:144-145`). No in-repo FE defect.

## ⚠️ Gating concern unchanged — backend deploy-ordering
The delta only removed the search-entry feature; it did **not** touch the two backend-coupled behaviors, so my prior position stands: this FE depends on **octo-server#569** being deployed first. If merged/deployed before #569: (a) empty-keyword browse 400s against a pre-#569 backend (only when `messagesSearchOn` is on), (b) `member_uids` is silently ignored by `c.BindJSON` (no `DisallowUnknownFields`) → filter appears active but returns unfiltered. Still **no runtime capability flag** decoupling them — ordering rests on merge discipline. The author's documented protocol (merge #569 first) or the offered capability flag remains the resolution.

## My position (unchanged)
COMMENT, not APPROVE (merge-readiness genuinely hinges on the external #569 ordering), not CR (no in-repo FE defect to fix — the code is correct given its documented backend contract). The gate correctly holds at CHANGES_REQUESTED via OctoBoooot's CR + `needs-human-review`, so no false-green. Credit the author for cleanly splitting out the search-entry feature. Net: FE ✅ clean and now properly scoped; merge only after octo-server#569 is in prod (or add the capability flag). Converging with OctoBoooot + yujiawei that ordering is the gate.

OctoBoooot (approved):
## Review: #688 @ `da0e9e56` — flipping to APPROVE; the backend dependency is now deployed

**Verdict:** Approve. My REQUEST_CHANGES was held solely on deployment ordering — both majors were "breaks *if merged before* octo-server#569." The maintainer confirms #569 is now **merged and deployed**, which retires that risk. The frontend code (unchanged at this head) is correct against the now-live backend contract.

Base `main`. `code-review` CI green. FE verified unchanged from my prior review (`shouldRunGlobalSearch` / `member_uids` at `apiAdapter.ts:209/145`).

### Why the hold is resolved
Both prior majors were deployment-ordering hazards, not code defects:
1. **Empty-keyword search** (`shouldRunGlobalSearch → return true`) — the risk was a 400 against a pre-#569 backend that rejected empty keyword. With #569 deployed, the backend accepts empty-keyword browse. Resolved.
2. **`member_uids` (plural) wire field** — the risk was silent no-op against a backend expecting `member_uid`. With #569 deployed, the backend understands `member_uids`. Resolved.

The frontend always correctly implemented its half of the documented two-PR contract (verified across rounds: `member_uids` dedup + self-drop + empty-omit; double-gated empty search behind `messagesSearchOn`; bug 1 chip-name persistence; bug 2 DM exclusion; tests updated). The only open question was ordering, and the maintainer has confirmed the backend is live.

### Note on the basis
This flip rests on the maintainer's confirmation that octo-server#569 is deployed to the target environment — deploy status is exactly the maintainer's call and is precisely the fact my CR was waiting on (I can't inspect the live octo-server from this repo). If for any environment #569 is not yet live, the two behaviors would regress there — but that's now a release-coordination fact the maintainer owns, not an in-repo defect.

### Also resolved earlier
The unrelated chat-search-entry scope-creep was removed in a prior commit (split to #695, which merged) — PR is cleanly the 4 filter-panel bugs.

### Net
Clean frontend implementation of the YUJ-30 filter fixes; the sole blocker (backend deploy ordering) is cleared now that #569 is live. Approving. `mergeStateStatus` stays BLOCKED under any remaining review-gate until reviews reconcile.

Jerry-Xin (approved):
Re-review at head `da0e9e56987f`. **Upgrading my prior COMMENT → APPROVE.** My COMMENT rested on a single condition — "merge only after octo-server#569 is in prod" — and the maintainer has now confirmed **#569 is merged and deployed**, which clears the sole gating concern.

## The gating condition is resolved
My prior verdict was: FE code clean, but merge-readiness hinges on the external backend deploy-ordering (octo-server#569). Both "majors" were *"breaks if merged before #569"*, never in-repo defects:
1. **Empty-keyword search** (`shouldRunGlobalSearch → return true`, `apiAdapter.ts:209`) — the 400 risk was against a *pre-#569* backend; with #569 live, empty-keyword browse is accepted.
2. **`member_uids` plural wire field** (`apiAdapter.ts:145`) — the silent-no-op risk was against a backend expecting `member_uid`; with #569 live, the backend understands `member_uids`.

The FE code is **unchanged at this head** (byte-confirmed `:209`/`:145` identical to my prior review) and always correctly implemented its half of the documented two-PR contract. With the backend now deployed, the ordering hazard is retired.

## FE quality (carried, still clean)
`member_uids` dedup + self-drop + omit-when-empty; empty-keyword double-gated behind `messagesSearchOn`; bug-1 chip-name persistence; bug-2 DM exclusion; meaningful tests. The unrelated chat-search-entry scope-creep was split out to #695 (merged), so this PR is cleanly the 4 filter-panel bugs.

## Basis note
This upgrade rests on the maintainer's confirmation that octo-server#569 is deployed to the target environment — deploy status is the maintainer's call (not inspectable from this repo), and it's exactly the fact my COMMENT was waiting on. Converges with OctoBoooot + yujiawei at APPROVE.

Net: FE was always clean; the sole gating condition (backend deploy ordering) is cleared now that #569 is live. **APPROVE.**

PR #690, opened 2026-07-13
fix(loop): show daemon version in Runtimes machine header, not the first agent's

caster-Q:
## Summary

Fixes the machine/group header on the Runtimes page showing one agent's CLI version (e.g. "Hermes Agent v0.16.0") as if it were the whole machine's version. The header now shows the daemon's own version (machine-level), while each runtime row keeps showing its own agent CLI version.

## Related Issue

N/A

## Changes

- `pages/runtimeVersion.ts` (new): extracted `runtimeVersion` (per-agent) and `deviceVersion` (machine/daemon-level) into a dependency-free module so they can be unit-tested without the component/CSS graph.
- `deviceVersion` now reads the daemon's `metadata.cli_version` (identical across a machine's runtimes) instead of returning the first runtime's agent version — which was arbitrary (array order) and duplicated the row below it.
- `pages/RuntimePage.tsx`: imports the two helpers instead of defining them inline.
- `pages/__tests__/runtimeVersion.test.ts` (new): asserts `deviceVersion` returns the daemon version (not any agent version), falls back to `-` when absent, and `runtimeVersion` still derives the per-agent version from `metadata.version` / `device_info`.

## Testing

- [x] Manually verified

`vitest` unit tests pass (5/5). `pnpm typecheck` clean. Verified live in the local app: the machine header now renders the daemon version `v0.3.31-3895-g3532fa88` instead of the first agent's version, with each runtime row still showing its own version.

Note: the "Unit tests added/updated" box reflects the checklist wording; tests **were** added (`runtimeVersion.test.ts`).

## Checklist

- [ ] Updated documentation

### DIFF
--- a/packages/dmloop/src/pages/RuntimePage.tsx
+++ b/packages/dmloop/src/pages/RuntimePage.tsx
@@ -6,6 +6,7 @@ import type { RuntimeDevice, RuntimeMode } from "../api/types";
 import { listRuntimes } from "../api/runtimeApi";
 import { issueHeadlessCliToken, getDaemonServerUrl } from "../api/authApi";
 import { headlessCommand } from "./headlessCommand";
+import { deviceVersion, runtimeVersion } from "./runtimeVersion";
 import "./runtime.css";
 
 const { Title } = Typography;
@@ -51,24 +52,8 @@ function deviceName(r: RuntimeDevice): string {
   return head || r.name;
 }
 
-function runtimeVersion(r: RuntimeDevice): string {
-  const version = r.metadata?.version;
-  if (typeof version === "string" && version.trim()) return version.trim();
-  const info = r.device_info || "";
-  const parts = info.split("·").map((part) => part.trim()).filter(Boolean);
-  return parts.slice(1).join(" · ") || r.launch_header || "-";
-}
-
 function deviceStatus(runtimes: RuntimeDevice[]): RuntimeDevice["status"] {
   return runtimes.some((runtime) => runtime.status === "online") ? "online" : "offline";
-}
-
-function deviceVersion(runtimes: RuntimeDevice[]): string {
-  for (const runtime of runtimes) {
-    const version = runtimeVersion(runtime);
-    if (version !== "-") return version;
-  }
-  return "-";
 }
 
 function relTime(iso: string | null): string {
--- a/packages/dmloop/src/pages/__tests__/runtimeVersion.test.ts
+++ b/packages/dmloop/src/pages/__tests__/runtimeVersion.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from "vitest";
+
+import type { RuntimeDevice } from "../../api/types";
+import { deviceVersion, runtimeVersion } from "../runtimeVersion";
+
+// Minimal RuntimeDevice factory — only the fields the version helpers read.
+function rt(over: Partial<RuntimeDevice>): RuntimeDevice {
+  return {
+    id: "r1",
+    workspace_id: "w1",
+    name: "n",
+    runtime_mode: "builtin",
+    provider: "claude",
+    status: "online",
+    device_info: "",
+    visibility: "workspace",
+    last_seen_at: null,
+    created_at: "",
+    updated_at: "",
+    ...over,
+  } as RuntimeDevice;
+}
+
+describe("deviceVersion (machine/daemon-level)", () => {
+  it("returns the daemon cli_version, NOT any agent runtime's own version", () => {
+    const runtimes = [
+      rt({ provider: "hermes", metadata: { version: "Hermes Agent v0.16.0", cli_version: "v0.3.31-3895-g3532fa88" } }),
+      rt({ provider: "claude", metadata: { version: "2.1.207 (Claude Code)", cli_version: "v0.3.31-3895-g3532fa88" } }),
+    ];
+    // The bug was returning the first runtime's agent version ("Hermes Agent ...").
+    expect(deviceVersion(runtimes)).toBe("v0.3.31-3895-g3532fa88");
+  });
+
+  it("falls back to '-' when no runtime reports a daemon cli_version", () => {
+    expect(deviceVersion([rt({ metadata: { version: "2.1.207 (Claude Code)" } })])).toBe("-");
+    expect(deviceVersion([])).toBe("-");
+  });
+});
+
+describe("runtimeVersion (per-agent)", () => {
+  it("returns the agent's own CLI version from metadata.version", () => {
+    expect(runtimeVersion(rt({ metadata: { version: "2.1.207 (Claude Code)" } }))).toBe("2.1.207 (Claude Code)");
+  });
+
+  it("derives from device_info when metadata.version is absent", () => {
+    expect(runtimeVersion(rt({ device_info: "casterdeMacBook-Pro.local · codex-cli 0.142.5" }))).toBe("codex-cli 0.142.5");
+  });
+
+  it("returns '-' when nothing is available", () => {
+    expect(runtimeVersion(rt({}))).toBe("-");
+  });
+});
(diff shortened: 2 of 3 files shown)


OctoBoooot (approved):
## Review: fix(loop): show daemon version in Runtimes machine header, not the first agent's (#690)

**Verdict:** Approve. A clean, correct, well-tested fix. No blockers, no majors.

Base `feat/octo-loop`. Risk tier: normal. `code-review` CI PENDING (not red); `mergeStateStatus=UNSTABLE`.

### The fix is correct
The old `deviceVersion` looped runtimes and returned the first non-`-` **agent** version (`runtimeVersion(runtime)`) — array-order-dependent and arbitrary, shown in the machine header as if it were the machine's version, duplicating the row directly below it. The new `deviceVersion` (`runtimeVersion.ts:18`) reads `metadata.cli_version` — the daemon's own version, identical across a machine's runtimes — which is the correct machine-level field. The per-agent `runtimeVersion` is unchanged, just relocated.

### Verified
- **Type-safe:** `RuntimeDevice.metadata` is `Record<string, unknown>` (`types.ts:639`); both helpers narrow with `typeof … === "string"` before `.trim()`, so reading `cli_version`/`version` is safe (no TS error, no unsafe access).
- **Wired correctly:** the machine header renders `deviceVersion(device.runtimes)` (`RuntimePage.tsx:232`) and each row still renders `runtimeVersion(runtime)` (`:265`) — the intended machine-vs-agent split.
- **Defensive loop:** `deviceVersion` returns the first non-empty `cli_version` rather than assuming `runtimes[0]` exists, and falls back to `-` on an empty list. Good.
- **No state/remount surface touched** — pure helper logic + a field change; no effects, no local state, so none of the RuntimePage remount/state-reset hazards apply here.

### Praise
The test (`runtimeVersion.test.ts`) directly encodes the regression: two runtimes with different agent `version`s but the same `cli_version`, asserting `deviceVersion` returns the `cli_version` and **not** the first agent's `"Hermes Agent v0.16.0"`. That's exactly the bug, pinned — and extracting the helpers into a dependency-free module to make them unit-testable off the component/CSS graph is the right call. `runtimeVersion`'s fallback chain (metadata.version → device_info parse → launch_header → "-") is covered too.

Small, focused, and the test would fail without the fix. Approving.

yujiawei (approved):
# Code Review — PR #690 (octo-web)

## 1. Spec compliance
**Spec: ✅**

The stated goal — the Runtimes machine/group header should show the daemon's own version (machine-level) instead of an arbitrary agent's CLI version — is fully and precisely implemented.

- No under-build: header now derives from `metadata.cli_version` (`packages/dmloop/src/pages/runtimeVersion.ts:19`), each runtime row still shows its own agent version via `runtimeVersion` (`RuntimePage.tsx:265`).
- No over-build: the change is limited to extracting the two helpers into a dependency-free module and repointing `deviceVersion`. `runtimeVersion` logic is preserved byte-for-byte from the old inline version. No unrelated flags/fields/APIs added.
- No divergence: the header render remains conditional (`{version !== "-" && <Tag>...}`, `RuntimePage.tsx:245`), so absent versions degrade to no tag rather than an error. The only caller of `deviceVersion` is `RuntimePage.tsx:232`; the refactor is complete with no dangling references.

## 2. Code Quality
**Quality: Approved**

The implementation is clean and the intent is well-documented. No P0/P1 issues found.

**Findings:**

- **P2 (behavioral change — by design, not a bug):** The old `deviceVersion` fell through `metadata.version → device_info → launch_header`, so it almost always displayed *something* (namely the first agent's version — the bug being fixed). The new `deviceVersion` reads only `metadata.cli_version`. Consequently, for a daemon that does not report `cli_version`, the machine header now shows no version tag where it previously showed an (incorrect) agent version. This is the intended correction — an arbitrary agent version at machine level is misleading and duplicates the row below — and it degrades gracefully. Worth a one-line confirmation that the daemon reliably populates `metadata.cli_version` in production so the header isn't silently blank for real machines. (`runtimeVersion.ts:18-23`)

- **P2 (nit):** `cli_version.trim()` is used for truthiness. It is functionally correct here (a whitespace-only string trims to `""` → falsy → falls back to `-`), but `cli_version.trim().length > 0` would read more explicitly. Non-blocking. (`runtimeVersion.ts:20`)

**Verification performed:**
- Unit tests pass: `vitest run runtimeVersion.test.ts` → 5/5 green.
- Type safety confirmed: `RuntimeDevice.metadata` is typed `Record<string, unknown>` (`api/types.ts:639`), which carries a string index signature, so the bracket access `metadata?.["cli_version"]` compiles under strict TypeScript. A scoped `tsc --noEmit` reports zero errors in the three changed files (remaining errors are pre-existing library-type noise in `node_modules`, unrelated to this PR).
- The new tests genuinely pin the fix: the first test would fail under the old implementation (it asserts the machine version is the daemon `cli_version`, not the leading `"Hermes Agent v0.16.0"` agent version), and fallback/`-` cases are covered.

## 3. Overall verdict
**APPROVE**

Spec ✅ AND Quality Approved. A focused, well-scoped, well-tested bug fix. The two findings above are advisory (P2) and do not block merge.

## 4. Suggestions
- Confirm the daemon reliably emits `metadata.cli_version` in production (schema/backend contract) so the machine header is populated for real machines, not just in the local verification.
- Optional readability nit on the whitespace guard (`.trim().length > 0`).

## 5. Additional observations
- `metadata` is consumed untyped (`Record<string, unknown>`). If `cli_version` becomes a stable part of the runtime contract, consider promoting it (and `version`) to typed optional fields on `RuntimeDevice` for compile-time safety and discoverability. Not required for this PR.

---

_Note on review coverage: this review combined an independent line-level pass with a long-context cross-file advisory pass. A second automated advisory pass was attempted but failed to run due to an environment/gateway error and is therefore treated as absent (not as silent approval); its intended scope — backend payload shape for `cli_version` — is called out under Suggestions above and could not be independently verified from the frontend repo alone._

Jerry-Xin (approved):
Reviewed at head `5f0b99825d2c`, byte-verified all 3 files. **APPROVE** — correct, well-tested fix with a clean helper extraction.

## ✅ The fix is correct
The bug: the machine-header version (`deviceVersion`) iterated runtimes and returned the first one's `runtimeVersion()` — i.e. the **first agent's** CLI version (e.g. "Hermes Agent v0.16.0" / "2.1.207 (Claude Code)"), which is agent-level info mislabeled as a machine/daemon version. The fix rewrites `deviceVersion` to read `metadata.cli_version` (`runtimeVersion.ts:19-21`) — the daemon's own CLI version, consistent across all runtimes on the machine — falling back to `-`. `runtimeVersion` (per-agent row) is unchanged. Correct separation of concerns: machine header = daemon version (`RuntimePage.tsx:232`), row = agent version (`:265`).

## ✅ Clean extraction + real tests
- Both helpers moved verbatim from `RuntimePage.tsx` into a testable `runtimeVersion.ts` module; `runtimeVersion`'s behavior is unchanged (byte-identical logic), only `deviceVersion` is rewritten. No behavior regression on the per-agent path.
- The test genuinely covers the reported bug: `deviceVersion([hermes, claude])` asserts it returns the daemon `cli_version` "v0.3.31-..." and explicitly **NOT** the first runtime's agent version "Hermes Agent v0.16.0" — the exact failing case. Plus fallback-to-`-` (no cli_version / empty array) and the `runtimeVersion` cases (metadata.version, device_info derivation, nothing-available). Meaningful assertions, not hollow.

## ✅ Type-safe
`metadata?: Record<string, unknown>` exists on `RuntimeDevice` (`types.ts:639`); the helper narrows via `typeof cliVersion === "string" && .trim()` before returning — no unsafe cast, no undefined deref.

Pure refactor+fix, no new deps, no side effects, CI green. Converges with OctoBoooot + yujiawei. APPROVE.

PR #698, opened 2026-07-13
feat(summary): support refining and regenerating summaries

Liuxy20:
## Summary
Users should be able to refine an existing summary result, regenerate it with an updated prompt, and review recent result versions.

## Changes
- Add refine and full regenerate modes in the summary result flow.
- Show lightweight generating status cards while refinement is running.
- Show recent result versions with detail and restore actions.
- Hide personal version history in multi-person summary personal sections.
- Update summary APIs, types, i18n copy, styles, and related tests.

## Testing
- pnpm --filter @dmwork/summary exec vitest run src/pages/__tests__/SummaryDetailPage.schedule.test.tsx
- pnpm i18n:check
- git diff --check
- pnpm --filter @octo/web exec vite build

Closes #696

PR #545, opened 2026-07-06
feat(summary): Agent summary mode in chat modal + MSW test infra

wanyiwang06:
## What

Add an **Agent summary** mode to the in-chat summary creation modal (`ChatSummaryNewModal`), and introduce MSW-based mock infrastructure so the summary package can be developed/tested without a backend.

### 1. Agent summary via split button (`19b22db`)
- Replace the single "Start summary" button in the modal footer with a **split button** (`SplitButtonGroup` + `Dropdown`). The dropdown switches between two modes: **Start summary** (normal) and **Agent summary**.
- Selecting a mode only switches `state.mode` (no immediate submit). The primary button label and the textarea placeholder follow the mode; clicking the primary button submits according to the current mode.
- Agent mode submits via a reserved endpoint `summaryApi.createAgentSummary` (`POST /summaries/agent`). Until the backend implements it, submission surfaces a Toast error. Normal mode keeps using `createSummary`.
- Textarea placeholder adapts: normal keeps the original hint; Agent mode shows a requirement-oriented hint.
- New type `CreateAgentSummaryParams` (core field: `requirement`).
- i18n: `agentStart` / `agentSubmitting` / `agentTopicPlaceholder` / `switchMode` in zh-CN and en-US.
- Add `ChatSummaryNewModal.stories.tsx` for Storybook preview; update the component unit test's semi-ui mock (SplitButtonGroup / Dropdown / IconChevronDown).

### 2. MSW mock infrastructure (`c801b15`, `9671bbd`)
- Add MSW handlers covering the summary API surface, node (`server.ts`) + browser (`browser.ts`) setups, wired into the vitest `setup.ts`.
- Enables offline development/testing of the summary UI without a real backend.
- Includes an integration test and code-review follow-ups.

## Testing
- `pnpm --filter @dmwork/summary test` — all tests pass.
- `pnpm i18n:check` — passes.
- Verified the split-button interaction (mode switch, placeholder change, submit dispatch) in an offline Vite dev preview backed by the MSW mocks.

## Notes
- The backend `POST /summaries/agent` endpoint is not implemented yet; when it is, only `createAgentSummary`'s path needs updating — the callsite stays unchanged.

PR #701, opened 2026-07-13
Feat/agent summary refine UI

wanyiwang06:
## Changes

-
-

## Testing

- [ ] Manually verified

## Checklist

- [ ] Updated documentation


wanyiwang06:
误提到上游仓库，关闭。该改动应在下游 fork 内进行，不进入上游 main。

packages/dmworksummary/src/components/__tests__/AgentChatPanel.test.tsx
```diff
@@ -0,0 +1,260 @@
...
+            // 验证 onSaveAsSummary 被调用
+            expect(onSaveAsSummary).toHaveBeenCalledWith('测试总结');
+            expect(onSaveAsSummary).toHaveBeenCalledTimes(1);
+        });
+        
+        // 保存失败后,对话框应该仍然存在
+        expect(screen.getByTestId('save-modal')).toBeInTheDocument();
+        // 标题应该保留
+        expect(titleInput).toHaveValue('测试总结');
+describe('AgentChatPanel 新会话 action', () => {
```
OctoBoooot, line 236: **[blocker]** Unclosed it()/describe() -> esbuild parse failure of the whole file. The async it() 保存失败后应保留对话框和已填标题 (opened ~:194) ends at :235 with expect(titleInput).toHaveValue(测试总结); and :236 immediately opens a new describe(...) — the it() callback and its enclosing describe(... Save as Summary) are never closed (brace delta +2, paren delta +2, verified). esbuild fails Unexpected end of file -> vitest run errors (CI red) and all 9 tests in this file never execute (IME-guard + save-dialog coverage lost). Same class as PR #545. Fix: add the two missing }); after :235.

packages/dmworksummary/src/components/ChatSummaryNewModal.tsx
```diff
@@ -205,6 +234,186 @@ export default class ChatSummaryNewModal extends Component<
...
+            agentSubmitting: true,
+        }));
+
+        try {
+            const res = await summaryApi.agentChat({ message: trimmed, session_id: sessionId, profile: 'summary' });
+            // 后端回传 session_id 非空则回填并持久化（与后端持久化的会话保持一致）。
+            const nextSessionId = res.session_id || sessionId;
+            writeAgentChatSession(this.agentChannelId(), nextSessionId);
+            this.setState((prev) => ({
+                messages: [...prev.messages, { role: 'assistant', content: res.reply }],
```
OctoBoooot, line 268: **[minor]** setState-after-unmount. module.tsx:193 unmounts this modal on close (if (!open || !channel) return null), but handleAgentSend awaits agentChat (up to 120s) then setState with no mounted-guard; same in handleSaveAsSummary (:363-414). Closing the modal mid-request setStates on an unmounted component. React 18 suppresses the warning (no crash), so minor — add a mounted ref / componentWillUnmount cancel.

OctoBoooot (changes requested):
## Review: Feat/agent summary refine UI (#701)

**Verdict:** Request changes — a test file that won't parse (blocker), plus a merge conflict and a red title-lint that already bar merge. The feature code itself is largely sound (no XSS, IME guard present, double-submit locks, i18n parity), but it can't land in this state.

Risk tier: high. Base `main`. All changes in `packages/dmworksummary` (focused despite size).

### Blocker — `AgentChatPanel.test.tsx` is syntactically invalid → whole file fails to parse
`src/components/__tests__/AgentChatPanel.test.tsx:235-236`: the test `'保存失败后应保留对话框和已填标题'` (async `it()` opened ~`:194`) ends at `:235` with `expect(titleInput).toHaveValue('测试总结');`, and `:236` immediately opens `describe('AgentChatPanel 新会话 action', …)` — the async `it()` callback and its enclosing `describe('… Save as Summary')` are never closed. Byte-verified: brace delta **+2**, paren delta **+2**. esbuild fails with "Unexpected end of file" → the entire file fails to transform → `vitest run` exits non-zero (CI red) and **all 9 tests in the file never execute** (the IME-guard and save-dialog coverage is silently lost). This is the exact class as the earlier sibling PR #545 (unclosed `it()` in this same file). Fix: add the two missing `});` after `:235`.

### Blocker (non-review gates) — not mergeable as-is
- **Merge conflict:** `mergeStateStatus=DIRTY` / `mergeable=CONFLICTING` (branch is 53 commits behind `main`). Needs a rebase before it can land (and before full CI runs clean).
- **`pr-title-lint: FAILURE`:** "Feat/agent summary refine UI" isn't conventional-commit format — needs `feat(summary): …` (lowercase type, colon). Merge-gating CI red.

### Minor
- **setState-after-unmount** on agent-chat/save when the modal closes mid-request. `module.tsx:193` renders `ChatSummaryNewModal` as `if (!open || !channel) return null` (full unmount on close), but `handleAgentSend` (`ChatSummaryNewModal.tsx:268-283`, awaits `agentChat` up to 120s) and `handleSaveAsSummary` (`:363-414`) `setState` after the await with no mounted-guard. React 18 suppresses the warning and there's no crash, so minor — add a mounted ref or `componentWillUnmount` cancel.

### Process note
The PR body is the uncompleted template (empty Summary / Related Issue / Changes / Testing) on a +3241 change. Please fill it in — reviewers and future readers get nothing from an empty body on a PR this size.

### Verified clean (byte-checked)
- **No XSS** — no `dangerouslySetInnerHTML` / `rehype-raw` anywhere in `src`; AgentChatPanel/RefineSection render content as text nodes, and summary markdown goes through `SummaryContent` with `react-markdown` + `rehypeSanitize`. No token/credential in logs or URLs (token is a header).
- **IME-safe Enter-to-send** — `AgentChatPanel.tsx:73-80` guards `isComposing` + `keyCode===229` before Enter submits (this is exactly the coverage the broken test file was meant to protect — worth restoring once the parse error is fixed).
- **Double-submit guarded** — synchronous `agentSendInFlight` instance lock on agent send (+ `disabled`/`loading`); `RefineSection` guards on `submitting`.
- **Async history race guarded** — `loadAgentHistory` uses `historyLoadToken` + session/mode re-check; `handleNewSession` invalidates in-flight loads.
- **i18n parity** — en-US and zh-CN both 351 keys, zero diff; all 24 new agent/refine keys present in both.

### Net
The feature is well-built on the dimensions that usually bite (XSS, IME, double-submit, i18n) — but the unclosed test file is a hard blocker (package test run fails, coverage lost), and the merge conflict + title-lint bar merge independently. Fix the two `});`, rebase onto main, fix the PR title to conventional-commit format, and fill in the body. REQUEST_CHANGES.

Jerry-Xin (commented):
Reviewed at head `e67be9f7469d` (true merge-base `4d595118`; the real delta is 31 files in `packages/dmworksummary/` + pnpm-lock, +3241/−162). **COMMENT** — no in-repo code blocker; the one hard gate is the merge conflict (rebase required).

## ⚠️ Correction to my earlier note: no xlsx CVE in this PR
My initial gate scan flagged `xlsx@^0.18.5` as a dependency concern — that was a **false alarm from diffing against the wrong base** (current main tip, which is 53-ahead/26-behind this head, pulling in unrelated main churn). Diffing against the true merge-base, **this PR touches zero xlsx references** in any package.json or pnpm-lock. `xlsx@0.18.5` is pre-existing in `packages/dmworkbase` on main, unrelated to #701. **#701's only new dependency is `msw@^2.14.6` (devDependency, for the MSW integration-test mocks)** — clean. Retracting the xlsx concern for this PR.

## 🔴 Gate — merge conflict (rebase required before merge)
`mergeable=CONFLICTING` / `mergeStateStatus=DIRTY`. The head is 26 behind main; **12 dmworksummary files conflict**, and they overlap exactly the files this PR edits (`summaryApi.ts`, `types/summary.ts`, `ChatSummaryNewModal.tsx`, `SummaryCreatePage.tsx`, `SummaryDetailPage.tsx`, both i18n files, their tests, css) — consistent with a #698-class sibling having landed on main. Author must rebase onto main; the merged result needs a re-verify since these are the same files the new agent code changes. This is why COMMENT (not APPROVE): the PR isn't merge-ready as-is.

## ✅ Code quality — solid (byte-verified)
- **Async stale-task race (the #698 class) is correctly guarded** in the stateful components: `SummaryCreatePage.tsx` uses an `agentSendInFlight` sync lock (`:91/:268-300`) + a `historyLoadToken` staleness guard (`:341` `++token`, `:344` `if (token !== this.historyLoadToken) return;` after the await) — the same requestTaskId pattern; `ChatSummaryNewModal.tsx` mirrors it. No cross-session/task corruption reachable.
- **XSS-safe**: zero `dangerouslySetInnerHTML`/`rehype-raw`/`innerHTML` in the delta; agent-chat bubbles render `{m.content}` as inert JSX text.
- **API envelope validated**: `agentChat` throws on `code!==0`/empty reply, 120s timeout for the multi-step agent loop, handles axios cancel; failures surface as an assistant error bubble + Toast. No secrets/tokens in code (uses shared axios/WKApp); localStorage try/catch-wrapped, session isolated by channel key.
- **i18n parity perfect** (351 en == 351 zh, zero mismatch). **MSW integration test is legitimate** — goes through the real axios stack intercepted by MSW (not mocking summaryApi), asserts real contract (type presence, 404 rejection, keyword filter, scope mapping). AgentChatPanel tests are behavioral (IME composition, save success/failure, disabled states).

## 🟡 Non-blocking
1. **Unbounded chat `messages` array** — `handleAgentSend` does `[...prev.messages, msg]` with no cap; a very long agent session grows unbounded (memory only, not corruption). Consider a cap/window.
2. **`RefineSection.tsx` post-await setState has no unmount/ignore guard** — React setState-after-unmount warning if the user navigates away mid-refine (task_id closure-captured, single-detail scope → warning only, not corruption). Add an ignore-flag in a cleanup for hygiene.
3. **Integration-test coverage gap** — the `createAgentSummary` test still uses the old `requirement` field while production `handleSaveAsSummary` now sends the new `session_id`+`title` contract; `agentChat`/`refineAgentSummary`/`getAgentChatHistory` have no integration coverage. Worth extending.

Net: the code is well-built and safe (race-guarded, XSS-safe, validated envelope, real tests, i18n parity). The blocker is purely the merge conflict — rebase onto main, then this should be good after a post-rebase re-verify of the 12 conflicted files. COMMENT (lean APPROVE-after-rebase).

yujiawei (changes requested):
# Code Review — PR #701 (octo-web)

**Head SHA:** `e67be9f7469df5f68d931ed6c98752262a7bc394` · merge-base `4d595118`
**Scope:** `packages/dmworksummary` — Agent Summary chat mode, `RefineSection`, four new API functions, and localStorage session helpers.

Overall this is a well-structured, well-commented change with solid test coverage. There is **one blocking issue** in the error-handling layer, plus several non-blocking suggestions.

---

## 1. Spec / feature completeness

The requested UI is present and coherent:
- `SplitButtonGroup` + `Dropdown` mode switch (normal / agent) wired identically in both `SummaryCreatePage` and `ChatSummaryNewModal`.
- Multi-turn `AgentChatPanel` (bubbles, IME-safe Enter-to-send, auto-scroll, "New chat", "Save as Summary").
- `RefineSection` gated by the intended triple condition (status Completed + `trigger_type === AGENT` + creator === current user).
- AI-generated `Tag` on `SummaryCard`, and full `en-US` / `zh-CN` i18n keys.

No missing or out-of-scope features were found at the UI level.

---

## 2. Blocking issue

### 🔴 P1 — Specific business-error handling for `refine` / `createAgentSummary` is unreachable in production, and the tests hide it

`refineAgentSummary` (`src/api/summaryApi.ts:285`) and `createAgentSummary` (`src/api/summaryApi.ts:139`) both go through the shared `post()` helper. Its catch block collapses every error into a plain `Error`:

```ts
// src/api/summaryApi.ts:96
async function post<T>(path: string, data?: unknown): Promise<T> {
    try {
        const resp = await summaryAxios.post(`${BASE}${path}`, data);
        return resp.data?.data ?? resp.data;
    } catch (err) {
        if (axios.isCancel(err)) throw err;
        throw new Error(extractErrorMessage(err)); // <-- .response is dropped
    }
}
```

But the callers branch on `err.response.data.code`:

```ts
// src/components/RefineSection.tsx:66
if (err && typeof err === 'object' && 'response' in err) {
    const code = axiosErr.response?.data?.code;
    if (code === 40001) { ... }      // refineNotSupported
    else if (code === 40002) { ... } // refineNotCreator
    else if (code === 40004) { ... } // refineNoSnapshot
}
```

Because `post()` re-throws a bare `Error`, `'response' in err` is **always false**, so the 40001/40002/40004 branches never execute. The same dead branch exists in `ChatSummaryNewModal.tsx:398` and `SummaryCreatePage.tsx:418` (the `40004 → noOutputToSave` case). Two outcomes, both wrong:
- Backend returns HTTP 4xx `{code, message}` → `post()` throws a bare `Error` → falls through to the generic `refineFailed` / `createFailedRetry` toast. The specific messages you wrote (and their i18n keys) never show.
- Backend returns HTTP 200 `{code: non-zero, data: null}` → `post()` returns `null` → `result.content` / `res.task_id` throws a `TypeError`, caught into the same generic toast.

Note the contrast: `agentChat` (`src/api/summaryApi.ts:149`) was deliberately written **not** to use `post()`, precisely so it can validate the envelope. `refine` and `createAgentSummary` need the same treatment.

This is compounded by a test-validity gap: `RefineSection.test.tsx` mocks `refineAgentSummary` to reject with `{response:{data:{code:40001}}}` — a shape the real function can never produce (it goes through `post()`). The tests pass green and appear to prove the error-code handling works, while production can't reach any of those branches.

**Suggested fix (pick one):**
- Give `refineAgentSummary` / `createAgentSummary` a bespoke wrapper that preserves the axios error / validates the envelope (mirror `agentChat`), **or**
- Have `post()` re-throw the original error (or an error that carries `code`) so callers can discriminate, **or**
- Move the `code`-mapping into the API layer and have it throw a typed error the UI can switch on.
Then adjust the tests to reject with the shape the real code path actually yields.

---

## 3. Non-blocking suggestions (P2 / nits)

- **P2 — A failed chat turn can be saved as a "summary".** In `handleAgentSend`, network/API failures are appended as `{ role: 'assistant', content: msg }` (`ChatSummaryNewModal.tsx:278`, `SummaryCreatePage.tsx:297`). `AgentChatPanel.hasAssistantOutput()` (`AgentChatPanel.tsx:82`) then returns `true`, enabling the **Save as Summary** button, so a user can persist an error string (e.g. "Network Error") as a deliverable. Consider a distinct role/flag for error bubbles and excluding them from `hasAssistantOutput()`.

- **P2 — Cross-entry inconsistency in `sources` on save.** In the modal, `selectedChats` is reset to `[defaultChat]` whenever it opens (`ChatSummaryNewModal.tsx:98`), so agent-mode "Save as Summary" always sends `sources = [current channel]` (`:367`). In `SummaryCreatePage`, `selectedChats` starts empty, so `sources` is omitted. The inline comments in both say origin is inferred backend-side from `tool_calls` — but the modal path still forces the current channel as `sources`, which may diverge from what the agent actually read. Worth confirming this is intended.

- **P2 — Unused import.** `getOriginChannelType` is imported in `SummaryCreatePage.tsx:18` and `ChatSummaryNewModal.tsx:7` but never used (only the pre-existing `getSourceType` is). Harmless (the repo does not enable `noUnusedLocals` / lint-unused), but it's dead weight — either use it for the `origin_channel_type` path or drop it.

- **P2 — Dead validation in `RefineSection`.** `onChange` hard-slices to `charLimit` (`:109`), so `trimmed.length > charLimit` (`:46`) can never be true. Silent truncation of pasted text also drops user content without warning; consider letting the value exceed the limit and blocking on submit instead.

- **Nit — `key={i}`.** `AgentChatPanel.tsx:142` keys messages by array index. Fine here since the list is append-only, but a stable id is more robust if editing/removal is ever added.

- **Nit — no `componentWillUnmount` guard.** `setState` in the `finally` of the async save/refine handlers can fire after unmount if the parent re-renders/unmounts on success (React dev warning only, not a functional bug).

- **Nit — trailing newline.** Both i18n JSON files lost their final newline (`\ No newline at end of file`).

---

## 4. What was not verified

- Backend contract for `/agent/chat`, `/agent/chat/history`, `/summaries/agent`, `/summaries/{id}/refine`: exact HTTP status vs. `{code}` semantics for business errors (directly determines the P1 behavior above), the `session_id` → `tool_calls` origin-inference, and the refine snapshot/versioning model.
- `@octo/base` internals (`WKApp.loginInfo.uid`, `I18nContext`, `useI18n`) — assumed stable.
- Runtime rendering / visual layout of the panel inside the fixed-height modal container was not exercised (no e2e).

Jerry-Xin (changes requested):
**Escalating my earlier COMMENT → REQUEST_CHANGES.** OctoBoooot caught a hard blocker my first pass missed — `AgentChatPanel.test.tsx` does not parse. Byte-verified and adopted.

## 🔴 Blocker — `AgentChatPanel.test.tsx` is syntactically broken (won't compile → test suite fails)
`packages/dmworksummary/src/components/__tests__/AgentChatPanel.test.tsx` (260 lines) has **unbalanced delimiters** — the outermost `describe(...)` wrapper is missing its closing `});`. Parsing it with esbuild (tsx loader) fails:
```
✘ [ERROR] Unexpected end of file
    agentchat.test.tsx:261:0
```
Delimiter balance over the file: `(` open 167 / close 165 (**+2**), `{` open 93 / close 91 (**+2**), `[` balanced. Depth-tracking shows the parse depth never returns to 0 — it ends at +2/+2 at EOF, i.e. a `describe` block's terminating `});` was dropped. Since this file can't parse, `vitest` can't load the suite → the new AgentChatPanel tests don't run and the test job breaks. (My initial pass read the test *logic* — 33 it/expect, real behavioral assertions — but did not compile the file; credit OctoBoooot for catching the syntax break. Same lesson as reading C# by hand vs a compiler.) **Fix:** add the missing closing `});` for the outer `describe`.

## 🔴 Gate — merge conflict (rebase required), carried from my COMMENT
`mergeable=CONFLICTING` / `DIRTY`; 12 dmworksummary files conflict, overlapping the exact files this PR edits (`summaryApi.ts`, `types/summary.ts`, `ChatSummaryNewModal.tsx`, `SummaryCreatePage.tsx`, `SummaryDetailPage.tsx`, both i18n, tests, css). Rebase onto main, then re-verify the merged result.

## 🟡 Repo-convention (per OctoBoooot, valid)
- PR title `Feat/agent summary refine UI` isn't conventional-commit format — should be `feat(summary): …`.
- PR description is empty — add scope/testing notes for reviewers.

## ✅ Code quality — solid (byte-verified, unchanged from my COMMENT)
- **Async stale-task race (#698 class) is correctly guarded** in the stateful components: `SummaryCreatePage.tsx` `agentSendInFlight` sync lock (`:91/:268-300`) + `historyLoadToken` staleness guard (`:341`/`:344` re-check after await); `ChatSummaryNewModal.tsx` mirrors it. No cross-task corruption.
- **XSS-safe** (zero `dangerouslySetInnerHTML`/`rehype-raw`; agent bubbles render `{m.content}` inert). **API envelope validated** (agentChat throws on `code!==0`/empty, 120s timeout, axios-cancel handled). No secrets; localStorage try/catch; session isolated by channel key.
- **i18n parity perfect** (351 en == 351 zh). MSW integration test is legitimate (real axios stack via MSW, asserts real contract).
- **Correction from my gate scan:** no xlsx CVE in this PR — that was a phantom from diffing the wrong base; against the true merge-base this PR touches zero xlsx refs (xlsx is pre-existing in dmworkbase on main). #701's only new dep is `msw` (dev). Retracted.

## 🟡 Non-blocking (carried)
- Unbounded chat `messages` array (`handleAgentSend` `[...prev, msg]` no cap — memory only).
- `RefineSection.tsx` post-await setState has no unmount/ignore guard (setState-after-unmount warning; task_id closure-captured → not corruption).
- Integration-test coverage gap (`createAgentSummary` test uses old `requirement` field vs new `session_id`+`title` contract; `agentChat`/`refine`/`history` uncovered).

Net: the feature code is well-built and safe, but two hard blockers gate it — a broken test file (won't parse) and the merge conflict. Fix the missing `});` in AgentChatPanel.test.tsx, rebase, and address the title/description conventions; then this should be good after a post-rebase re-verify. REQUEST_CHANGES.

yujiawei:
### Follow-up finding (P2) — refine callback can silently drop `citations`

Supplementing the review above with one additional item found on a second pass.

In `packages/dmworksummary/src/pages/SummaryDetailPage.tsx` (`onRefineSuccess`, ~L2044), the new result is built by spreading the prior result and then overwriting `citations` with the refine response:

```ts
result: detail.result ? {
    ...detail.result,
    content: newContent,
    version: newVersion,
    citations,          // <-- unconditional overwrite
} : null,
```

`refineAgentSummary` declares `citations?: CitationItem[]` (optional). When the backend omits it, `citations` becomes `undefined`, wiping whatever citations the result previously had. The citation badges rendered from `detail.result.citations || []` (e.g. L997 / L1188) would then silently disappear after a refine round-trip.

Suggested fix — fall back to existing citations when the response omits them:
```ts
citations: citations ?? detail.result?.citations,
```

This is non-blocking on its own (P2), but worth folding into the same revision as the blocking error-handling issue.

packages/dmworksummary/src/api/summaryApi.ts
```diff
@@ -124,6 +129,54 @@ export async function createSummary(params: CreateSummaryParams): Promise<{ task
...
+ * 创建 Agent 总结（契约 v1.0）。
+ *
+ * POST /summary/api/v1/summaries/agent
+ * 让后端 agent 自主总结当前对话的产出内容，落库为可检索的交付物。
+ * 响应与传统 createSummary 同构：{ task_id, task_no, status, created_at }
+ */
+export async function createAgentSummary(
+    params: CreateAgentSummaryParams,
+): Promise<{ task_id: number; task_no: string; status: number; created_at: string }> {
+    return post('/summaries/agent', params);
```
OctoBoooot, line 142: **[major]** Business-error swallowed. createAgentSummary (and refineAgentSummary at :289) route through the shared post() helper, which returns `resp.data?.data ?? resp.data` on HTTP-2xx WITHOUT checking the business code (:96-104). So a HTTP-200 + {code: non-zero, data: null} response is treated as success -> null/undefined returned, caller's error branch never fires. The author already guards this exact hazard in agentChat (:153-155, `if (resp.data?.code !== 0) throw`) with a comment naming it (:146-148) — but these two create/refine paths were left on the flawed post(). Failing case: backend returns {code:1001,msg:'quota exceeded',data:null} @ HTTP200 -> createAgentSummary resolves 'successfully' with null, no error shown. Fix: apply the same envelope-code check here (opt-in on post(), or hand-roll like agentChat). Credit yujiawei.

OctoBoooot (changes requested):
## Supplement to my #701 CR (@ `e67be9f7`) — adopting a real error-handling major yujiawei caught

Verdict unchanged (still REQUEST_CHANGES; the test-parse blocker + merge conflict + title-lint stand). Adding one byte-verified **major** from yujiawei's review that my pass and Steve's both missed. Credit yujiawei.

### Major — business-error responses swallowed on the summary create/refine path
The shared `post<T>()` helper (`summaryApi.ts:96-104`) returns `resp.data?.data ?? resp.data` on HTTP-2xx **without checking the business `code`** — so a HTTP-200 response with `{code: <non-zero>, data: null}` (a business-level failure) is treated as success and `null`/`undefined` is returned to the caller.

The author **knows** this: `agentChat` (`:153-155`) deliberately bypasses `post()`, calls `summaryAxios` directly, and checks `if (resp.data?.code !== 0) throw`, with a comment (`:146-148`) naming the exact hazard — *"不复用公共 post():post() 只 `data?.data ?? data`,不校验业务 code,HTTP200 + {code:非0,data:null} 会被当成功、undefined 追进气泡"*.

But `createAgentSummary` (`:142`) and `refineAgentSummary` (`:289`) still route through the flawed `post()`. So on those two user-facing paths a business error (HTTP-200 + non-zero code) is silently swallowed: no throw, `undefined`/null surfaced to the UI instead of an error message. Failing scenario: backend returns `{code: 1001, msg: "quota exceeded", data: null}` with HTTP 200 → `createAgentSummary` resolves "successfully" with `null` → the create flow proceeds as if it worked, no error shown. Major (real error-handling gap on a create/refine path, self-documented as a known failure mode) — not a blocker (no compile/CI break; that's the separate test-syntax blocker).

Fix: apply the same envelope-`code` validation `agentChat` uses to `createAgentSummary`/`refineAgentSummary` — either give `post()` an opt-in business-code check, or hand-roll it in these two like `agentChat` does.

### Corollary 🟡 (yujiawei) — test mocks the wrong error shape
Because the create/refine tests mock a rejection/error shape that `post()` doesn't actually produce for a business-code error (it resolves, not rejects), the broken handling "passes" the test. Worth re-mocking a HTTP-200 + non-zero-code envelope so the test would fail against the current code and pass only once the code-check is added.

### Net (verdict stands)
Still REQUEST_CHANGES. Prior blockers unchanged (AgentChatPanel.test.tsx unclosed `it()`/`describe()` → parse failure; merge conflict; pr-title-lint). This adds a real error-handling major on the create/refine path that the author already guarded against in the sibling `agentChat` — apply the same code-check there. Credit yujiawei for the catch.

Jerry-Xin (commented):
Supplement to my REQUEST_CHANGES — byte-verifying yujiawei's error-handling finding. **It's a real inconsistency, but I grade it 🟡 (not 🔴)**, and it's contingent on the backend error convention. My CR verdict is unchanged (still gated by the test-parse blocker + merge conflict); adding this so the author sees the consolidated picture.

## 🟡 `createAgentSummary` / `refineAgentSummary` don't validate the business `code` (unlike `agentChat`)
Byte-verified in `summaryApi.ts`:
- `createAgentSummary` (`:142` → `post('/summaries/agent', …)`) and `refineAgentSummary` (`:289` → `post('/summaries/${taskId}/refine', …)`) both route through the shared `post()` (`:96-102`), which does `return resp.data?.data ?? resp.data` — it unwraps `.data` but **does not check the envelope `code`**.
- By contrast `agentChat` (`:149`) **deliberately does not reuse `post()`** — its comment (`:146`) says exactly "不复用公共 post()：post() 只 `data?.data ?? data`，不校验业务 code" and it explicitly `if (resp.data?.code !== 0) throw` (`:154`).

So among the three agent endpoints, one validates the business `code` and two don't. **If** the agent create/refine endpoints signal failure the same in-band way `agentChat` does (HTTP 200 + `{code: non-zero, message}`), then a business failure would be silently treated as success — `post()` returns `data ?? undefined`, so e.g. `createAgentSummary` yields `{task_id: undefined, …}` and the caller proceeds as if it worked, with no error toast.

**Why 🟡 not 🔴 (my calibration vs yujiawei's):** `post()` is not "error handling overwritten" wholesale — it *does* propagate HTTP/transport errors (non-2xx → axios throws → `throw new Error(extractErrorMessage)`), and it preserves axios cancel. The gap is specifically the **in-band business-`code`** path, and whether it bites depends on the backend contract for these two endpoints (which I can't confirm from the frontend). It's a real consistency/robustness gap — the same agent subsystem, one endpoint hardens the code-check and its two siblings don't — but it's not data corruption and HTTP errors still surface. **Fix:** either give `createAgentSummary`/`refineAgentSummary` the same `code!==0` guard `agentChat` has (if these endpoints use in-band codes), or confirm they signal via HTTP status and leave a comment saying so — matching the pattern the author already established 140 lines up. yujiawei's related test note (mocks reject with a shape that doesn't match production, so the code-branch isn't actually exercised) is worth folding into the same fix.

## Verdict unchanged
Still **REQUEST_CHANGES** — the gating items remain the `AgentChatPanel.test.tsx` parse blocker (missing outer `});`) + the merge conflict (rebase). This business-code 🟡 is a consistency fix to fold into the same pass. Feature code otherwise solid (race-guarded, XSS-safe, i18n parity) as noted in my main review.

PR #702, opened 2026-07-13
fix(docs): resolve non-friend bot names in member panel via space_bots

ploy-elison:
## Summary

Non-friend / non-self-created bots showed their raw uid (instead of a name) in the docs **member panel**. Root cause: the panel resolves `uid → display name` only from the space-member source (`queryMembers`), which filters out any bot the current user did not create — so those bots never carry a name and the panel falls back to the uid.

This is a **frontend-only** fix. It backfills the missing names with a single `GET /robot/space_bots?space_id=<spaceId>` request per space — the same non-viewer-scoped endpoint the contacts module already calls (`dmworkcontacts/src/Contacts/index.tsx`). No backend change, no per-uid fanout, no extra permission.

## Changes

- `packages/docs/src/octoweb/index.ts` — add `fetchSpaceBotNames(spaceId)` seam helper: one `GET /robot/space_bots?space_id=` request, mapped to `{ uid, name }` (name falls back to the uid). Resolves to `[]` on failure or a non-array body so callers can merge safely.
- `packages/docs/src/members/memberNames.ts` — fetch members and space bots together; merge a bot name **only** for uids that are absent or still resolve to their raw uid. Human/member names are never overwritten. The bot backfill is independently best-effort — if it fails, human names still resolve; a member-fetch failure still evicts the cache so a later open retries.

## Behavior

- Non-friend / non-self-created bots now display their real name in the member panel.
- Single request per space, no per-uid fanout.
- Human members unchanged; empty name / request failure degrades to the raw uid without crashing.
- No extra permission dependency.

## Tests

- `memberNames.test.ts` — backfill, no-overwrite of human names, blank-name fallback, request-failure fallback, non-array body tolerance.
- `spaceMembers.test.ts` — `fetchSpaceBotNames` seam: single request + URL, uid fallback / skip-no-uid, non-array body, blank space id.

New tests pass; `tsc --noEmit` reports no new errors (two pre-existing errors in untouched files remain).

Ref: Mininglamp-OSS/octo-docs-backend#60

### DIFF
--- a/packages/docs/src/members/memberNames.ts
+++ b/packages/docs/src/members/memberNames.ts
@@ -6,11 +6,18 @@
 // through the octoweb seam (fetchAllSpaceMembers). This module fetches that list ONCE per space
 // and caches the resulting uid → name map so the editor/member panel can resolve names cheaply.
 //
+// Bot backfill (octo-docs-backend #60): the space-member source (`queryMembers`) drops any bot the
+// current user did not create, so a non-friend / non-self-created bot has no name here and the panel
+// falls back to its raw uid. We backfill those names with a SINGLE `GET /robot/space_bots?space_id=`
+// request per space (not viewer-scoped, no per-uid fanout, no extra permission) and merge only the
+// uids that are still missing a real name — human members keep their original name untouched.
+//
 // Resilience: a fetch failure resolves to an EMPTY map (never throws) and the failed entry is
 // evicted so a later open retries — callers always fall back to the uid, so first paint can
-// never crash on a missing/slow member list.
+// never crash on a missing/slow member list. The bot backfill is independently best-effort: if it
+// fails the human names still resolve.
 
-import { fetchAllSpaceMembers } from '../octoweb/index.ts'
+import { fetchAllSpaceMembers, fetchSpaceBotNames } from '../octoweb/index.ts'
 
 const cache = new Map<string, Promise<Map<string, string>>>()
 
@@ -22,19 +29,32 @@ export function getSpaceMemberNames(spaceId: string): Promise<Map<string, string
   if (!spaceId) return Promise.resolve(new Map<string, string>())
   const cached = cache.get(spaceId)
   if (cached) return cached
-  const pending = fetchAllSpaceMembers(spaceId)
-    .then((members) => {
-      const map = new Map<string, string>()
-      for (const m of members) {
-        if (m.uid) map.set(m.uid, m.name || m.uid)
-      }
-      return map
-    })
-    .catch(() => {
-      // Transient failure: forget it so a later open retries instead of caching "no names".
-      cache.delete(spaceId)
-      return new Map<string, string>()
-    })
+  const pending = Promise.all([
+    // Human/self-created-bot names from the space-member source.
+    fetchAllSpaceMembers(spaceId).then(
+      (members) => members,
+      () => null, // signal a transient member-fetch failure so we can evict + retry below
+    ),
+    // Bot names from the single non-viewer-scoped space_bots request (#60). Independently
+    // best-effort: a failure here must never take down the human-member names.
+    fetchSpaceBotNames(spaceId).catch(() => [] as Awaited<ReturnType<typeof fetchSpaceBotNames>>),
+  ]).then(([members, bots]) => {
+    // Transient failure of the member list: forget it so a later open retries instead of
+    // caching "no names". We still merge whatever bot names we got for this render.
+    if (members === null) cache.delete(spaceId)
+    const map = new Map<string, string>()
+    for (const m of members ?? []) {
+      if (m.uid) map.set(m.uid, m.name || m.uid)
+    }
+    // Backfill ONLY uids that are absent or still resolve to their raw uid — never overwrite a
+    // real human/member name (bots filtered out of queryMembers are exactly the missing ones).
+    for (const b of bots) {
+      if (!b.uid) continue
+      const existing = map.get(b.uid)
+      if (!existing || existing === b.uid) map.set(b.uid, b.name || b.uid)
+    }
+    return map
+  })
   cache.set(spaceId, pending)
   return pending
 }
--- a/packages/docs/src/members/memberNames.test.ts
+++ b/packages/docs/src/members/memberNames.test.ts
@@ -32,3 +32,68 @@ describe('getSpaceMemberNames — uid → display name resolution', () => {
     expect(map.size).toBe(0)
   })
 })
+
+describe('getSpaceMemberNames — bot name backfill via /robot/space_bots (#60)', () => {
+  let wk: ReturnType<typeof createMockWKApp>
+
+  beforeEach(() => {
+    clearMemberNameCache()
+    wk = createMockWKApp()
+    setWKApp(wk)
+  })
+
+  /** Make the mock apiClient answer /robot/space_bots?space_id=... with the given bot rows. */
+  function respondBots(bots: unknown, opts: { fail?: boolean } = {}): void {
+    wk.apiClient.responder = (_method, url) => {
+      if (url.startsWith('/robot/space_bots')) {
+        if (opts.fail) return Promise.reject(new Error('space_bots down'))
+        return { data: bots, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+  }
+
+  it('backfills names for bot uids missing from the space-member list (single request)', async () => {
+    // A non-friend / non-self-created bot never appears in the space-member list.
+    respondBots([{ uid: 'bot1', name: 'Helper Bot' }])
+    const map = await getSpaceMemberNames('s_1')
+    expect(map.get('bot1')).toBe('Helper Bot')
+    // One space_bots request only — no per-uid fanout.
+    const botCalls = wk.apiClient.calls.filter((c) => c.url.startsWith('/robot/space_bots'))
+    expect(botCalls).toHaveLength(1)
+    expect(botCalls[0].url).toBe('/robot/space_bots?space_id=s_1')
+  })
+
+  it('never overwrites an existing human/member display name', async () => {
+    wk.spaceMembers.push({ uid: 'u1', name: 'Alice' })
+    // Even if space_bots echoes u1 with a different name, the member name wins.
+    respondBots([
+      { uid: 'u1', name: 'Alice Bot Alias' },
+      { uid: 'bot1', name: 'Helper Bot' },
+    ])
+    const map = await getSpaceMemberNames('s_1')
+    expect(map.get('u1')).toBe('Alice') // human path unchanged
+    expect(map.get('bot1')).toBe('Helper Bot')
+  })
+
+  it('falls back to the raw uid for a bot with a blank name', async () => {
+    respondBots([{ uid: 'bot1', name: '' }])
+    const map = await getSpaceMemberNames('s_1')
+    expect(map.get('bot1')).toBe('bot1')
+  })
+
+  it('keeps human names when the space_bots request fails (best-effort backfill)', async () => {
+    wk.spaceMembers.push({ uid: 'u1', name: 'Alice' })
+    respondBots([], { fail: true })
+    const map = await getSpaceMemberNames('s_1')
+    expect(map.get('u1')).toBe('Alice')
+    expect(map.has('bot1')).toBe(false)
+  })
+
+  it('tolerates a non-array space_bots body without throwing', async () => {
+    wk.spaceMembers.push({ uid: 'u1', name: 'Alice' })
+    respondBots({ unexpected: true })
+    const map = await getSpaceMemberNames('s_1')
+    expect(map.get('u1')).toBe('Alice')
+  })
+})
--- a/packages/docs/src/octoweb/index.ts
+++ b/packages/docs/src/octoweb/index.ts
@@ -154,6 +154,36 @@ export async function fetchAllSpaceMembers(spaceId: string): Promise<SpaceMember
   return all
 }
 
+/** Minimal view of a `/robot/space_bots` entry the docs seam reads (uid + display name). */
+interface HostSpaceBot {
+  uid: string
+  name?: string
+}
+
+/**
+ * Fetch the display names of ALL bots in a space via the single `GET /robot/space_bots?space_id=`
+ * request (the same endpoint the contacts module already calls — dmworkcontacts Contacts/index.tsx).
+ *
+ * WHY: `queryMembers` (the source behind fetchAllSpaceMembers) filters out bots the current user did
+ * not create, so the member panel falls back to the raw uid for a non-friend / non-self-created bot
+ * (octo-docs-backend #60). This endpoint is NOT viewer-scoped — it returns every bot in the space —
+ * so one request per space backfills those names with no per-uid fanout and no extra permission.
+ *
+ * Returns `{ uid, name }` pairs (name falls back to the uid so a bot with no display name is never
+ * blank). Resolves to an EMPTY list on any failure or non-array body so callers can merge safely and
+ * fall back to the uid — this must never break the human-member name path.
+ */
+export async function fetchSpaceBotNames(spaceId: string): Promise<SpaceMemberLite[]> {
+  if (!spaceId) return []
+  const { data } = await apiClient().get<HostSpaceBot[]>(
+    `/robot/space_bots?space_id=${encodeURIComponent(spaceId)}`,
+  )
+  const bots = Array.isArray(data) ? data : []
+  return bots
+    .filter((b): b is HostSpaceBot => !!b && !!b.uid)
+    .map((b) => ({ uid: b.uid, name: b.name || b.uid }))
+}
+
 /**
  * Re-wrap the REAL host APIClient so its responses look axios-style to docs callers.
  *
--- a/packages/docs/src/octoweb/spaceMembers.test.ts
+++ b/packages/docs/src/octoweb/spaceMembers.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach } from 'vitest'
-import { setWKApp, getSpaceMembers, fetchAllSpaceMembers } from './index.ts'
+import { setWKApp, getSpaceMembers, fetchAllSpaceMembers, fetchSpaceBotNames } from './index.ts'
 import { createMockWKApp } from './mock.ts'
 
 // Seam spike acceptance (#7): prove the docs package can reach the host's space-member source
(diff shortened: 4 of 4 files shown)


Jerry-Xin (approved):
## Review: member-panel bot name backfill via `space_bots` (#60)

**Conclusion: APPROVED**

The change is a clean, frontend-only fix that resolves the root cause and does not regress the existing member-panel behavior. Verified against the PR-2 checklist below.

### Root cause / correctness
- `queryMembers` (behind `fetchAllSpaceMembers`) drops bots the current viewer did not create, so those uids never carry a name and the panel falls back to the raw uid. The fix backfills exactly those uids from `GET /robot/space_bots?space_id=` — this addresses the real cause, not a backend-predicate or per-uid hack.
- `/robot/space_bots` is the same non-viewer-scoped endpoint the contacts module already calls (`dmworkcontacts/src/Contacts/index.tsx`), so no new permission is introduced.

### No fanout / single request
- `fetchSpaceBotNames` issues one request per space; there is no per-uid loop. Confirmed by `spaceMembers.test.ts` and `memberNames.test.ts` (`botCalls` length asserted === 1, exact URL asserted).

### Backfill-only, no overwrite
- Merge condition `!existing || existing === b.uid` fills only uids that are absent or still resolve to their own uid, so real human/member names are never overwritten (covered by the `Alice`/`Alice Bot Alias` test). Since `space_bots` only returns bots, a human uid cannot collide here.

### Resilience
- `fetchSpaceBotNames` resolves to `[]` on failure or a non-array body; the call site also wraps it in `.catch(() => [])`, so a bot-fetch failure never takes down the human-member names. A member-fetch failure still evicts the cache so a later open retries. Empty/blank bot name degrades to the raw uid. All four paths are covered by tests.

### Quality / verification
- `npx vitest run` on both new suites: 17/17 pass.
- `tsc --noEmit -p tsconfig.typecheck.json`: no new errors. The two reported errors (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) are in files this PR does not touch and are pre-existing.

### Non-blocking nit
- `memberNames.ts`: the `if (!b.uid) continue` guard is redundant because `fetchSpaceBotNames` already filters out uid-less entries. Harmless — leaving it for defense-in-depth is fine.

No blocking issues found. Good to promote out of draft.

OctoBoooot (approved):
## Review: fix(docs): resolve non-friend bot names in member panel via space_bots (#702)

**Verdict:** Approve. Small, focused, correct fix with sound merge precedence, genuine resilience, and thorough tests. No blockers, no majors.

Base `main`. Risk tier: normal. `mergeStateStatus=CLEAN`, CI green.

### What's good
- **Merge precedence is correct.** Human/member names load first (`memberNames.ts:46-48`); bot backfill applies **only** to uids absent or still resolving to their raw uid (`:53-54`, `if (!existing || existing === b.uid)`) — a real human/member name is never overwritten. Matches the stated behavior exactly.
- **Resilience is genuinely well-handled** (not just claimed). Two independent best-effort fetches: `fetchAllSpaceMembers` failure → `null` sentinel → cache eviction so a later open retries instead of caching "no names" (`:36/:44`); `fetchSpaceBotNames` failure → `[]` (`:40`), so a bot-fetch failure never takes down human names. Both degrade to the raw uid, never throw — first paint can't crash on a missing/slow list.
- **The seam is safe.** `fetchSpaceBotNames` (`octoweb/index.ts`) `encodeURIComponent`s the `space_id` param (no injection), short-circuits blank spaceId, guards a non-array body (`Array.isArray(data) ? data : []`), and filters null/uid-less entries before mapping — no throw path of its own. Single request per space, no per-uid fanout, reuses the existing non-viewer-scoped `/robot/space_bots` endpoint (no new permission/backend).

### Praise
Test coverage (`memberNames.test.ts`, +65) hits the behaviors that actually matter: bot backfill for missing uids, **never overwrites a human name**, blank-name→uid fallback, **human names survive a space_bots request failure**, and a **non-array body tolerated without throwing** — plus caching and blank-spaceId. Real behavioral assertions, not hollow. (`spaceMembers.test.ts` +46 covers the seam.)

### Net
Correct root-cause fix (bots dropped by `queryMembers` get backfilled from `space_bots`), safe and best-effort throughout, well-tested. Approving.

yujiawei (approved):
# Code Review — PR #702 (octo-web)

**Head SHA reviewed:** `dff31849e0d7c1aad09562a52deb8c7079fc2d7d`

Frontend-only fix: non-friend / non-self-created bots rendered their raw uid instead of a display name in the docs member panel, because the space-member source (`queryMembers`) drops bots the current user didn't create. The fix backfills those names via a single `GET /robot/space_bots?space_id=` request and merges them into the uid→name map, touching only uids that are absent or still resolve to their raw uid.

## 1. Spec compliance

**Result: ✅**

- **Missing:** none. The described behavior (single request per space, no per-uid fanout, backfill missing names only, human names never overwritten, best-effort degradation) is fully implemented in `packages/docs/src/members/memberNames.ts` and `packages/docs/src/octoweb/index.ts:176-186`.
- **Extra:** none. No new permission, no backend change, no unrelated edits.
- **Deviation:** none.

Verified cross-file: the referenced consumer of the same endpoint reads entries exactly as `{ uid, name || uid }` (contacts module, `Contacts/index.tsx:425-432`), so this PR's `{ uid, name }` mapping matches the real `/robot/space_bots` response shape. The inline `?space_id=` query style (vs the host's `{ param: { space_id } }` config style used elsewhere) is consistent with how the docs seam issues every other GET (e.g. `pages/docsApi.ts`), and the value is `encodeURIComponent`-escaped — no correctness impact. No existing consumer of `/robot/space_bots` paginates it, so the single-request approach is consistent with established usage.

Tests: `memberNames.test.ts` + `spaceMembers.test.ts` (17 tests) pass locally; they cover backfill, no-overwrite of human names, blank-name → uid fallback, request-failure fallback, and non-array body tolerance.

## 2. Code quality

**Result: Approved**

No P0/P1 issues. The change is correct, well-scoped, and well-tested. The following are advisory (non-blocking):

**[P2] `fetchSpaceBotNames` does not honor its own "never throws" contract.**
`packages/docs/src/octoweb/index.ts:176-186` — the JSDoc states it "Resolves to an EMPTY list on any failure or non-array body," but the body has no `try/catch` around `await apiClient().get(...)`; a rejected request propagates. It is only safe today because the sole caller wraps it in `.catch(() => [])` (`memberNames.ts:38`). There is no direct test of the rejection path — the seam tests cover a non-array body and a blank id, but not a rejected request. Any future caller that omits the `.catch` will get an unhandled rejection. Consider moving the failure→`[]` guard into the function so its contract is self-honored, and add a rejection test.

**[P2] Cache-eviction ownership on member-fetch failure.**
`packages/docs/src/members/memberNames.ts:43` — `if (members === null) cache.delete(spaceId)` deletes whatever is currently stored for `spaceId`, not necessarily the `pending` promise this closure created. If `clearMemberNameCache()` runs and a new fetch is stored while the first is still in flight, the first request's delayed failure can evict the newer, valid promise. Narrow window, and note this pattern is **pre-existing** (the prior `.catch` did the same unconditional delete), so it is carried forward rather than introduced here — but the refactor was a chance to fix it: `if (members === null && cache.get(spaceId) === pending) cache.delete(spaceId)`.

**[P3] Bot-fetch failure is cached for the session.**
`memberNames.ts:38` — a member-fetch failure evicts the cache so a later open retries, but a bot-fetch failure resolves to `[]` and the resulting bot-less map is cached until a space switch / reload. This is a deliberate best-effort tradeoff, but a transient `space_bots` outage silently degrades bot names for the rest of the session. Acceptable if intended; worth a comment noting bot names won't self-heal without a reload.

**[P3] `Promise.all` couples paint latency to the slower request.**
`memberNames.ts:32-39` — the map now resolves at `max(members, bots)`. Since the hook shows an empty map until resolution, a slow/hanging `space_bots` request delays name display for humans too. Bounded by the api client's request timeout; low impact, but it does soften the module's "first paint never blocks on a slow member list" guarantee.

**[P3 / nit] Redundant fallback.**
`memberNames.ts:50` — `map.set(b.uid, b.name || b.uid)`: `fetchSpaceBotNames` already normalizes `name` to fall back to `uid` (`index.ts:185`), so `b.name` is never falsy here. Harmless, just dead defensiveness.

## 3. Overall verdict

**APPROVED** — spec ✅ and code quality Approved. All findings are P2/P3 advisory; none block merge.

## 4. Suggestions

The two P2s are the highest-value follow-ups: (a) push the failure guard into `fetchSpaceBotNames` so its documented "never throws" contract holds independently of the caller, with a rejection test; (b) scope the cache eviction to the owning promise. Both are small and can land in a follow-up.

## 5. Coverage note

- The independent cross-family adversarial leg (line-level second opinion) could not run in this environment (upstream gateway rejected the request with an `image_generation` tool error on every call, including a trivial ping). It is recorded as **absent**, not as tacit approval. This review reflects one long-context advisory pass plus the reviewer's own line-by-line analysis; the cache-eviction race and the redundant-fallback nit were independently confirmed by both.
- Not verifiable from the diff: the api client's default request timeout (bounds the P3 latency-coupling penalty). Confirmed from the codebase: `/robot/space_bots` is not paginated by any existing consumer, so pagination is not a gap here.

packages/docs/src/members/memberNames.ts
```diff
@@ -22,19 +29,32 @@ export function getSpaceMemberNames(spaceId: string): Promise<Map<string, string
...
+      (members) => members,
+      () => null, // signal a transient member-fetch failure so we can evict + retry below
+    ),
+    // Bot names from the single non-viewer-scoped space_bots request (#60). Independently
+    // best-effort: a failure here must never take down the human-member names.
+    fetchSpaceBotNames(spaceId).catch(() => [] as Awaited<ReturnType<typeof fetchSpaceBotNames>>),
+  ]).then(([members, bots]) => {
+    // Transient failure of the member list: forget it so a later open retries instead of
+    // caching "no names". We still merge whatever bot names we got for this render.
+    if (members === null) cache.delete(spaceId)
```
OctoBoooot, line 44: **[minor]** Cache eviction deletes by key, not identity. `if (members === null) cache.delete(spaceId)` inside call A's .then can delete a DIFFERENT call B's valid entry if B already cache.set a new pending (:58) into the slot before A resolves. Narrow same-space race; harm is an extra refetch. Fix: `if (cache.get(spaceId) === pending) cache.delete(spaceId)`. Credit yujiawei.

packages/docs/src/members/memberNames.ts
```diff
@@ -22,19 +29,32 @@ export function getSpaceMemberNames(spaceId: string): Promise<Map<string, string
...
+    // Backfill ONLY uids that are absent or still resolve to their raw uid — never overwrite a
+    // real human/member name (bots filtered out of queryMembers are exactly the missing ones).
+    for (const b of bots) {
+      if (!b.uid) continue
+      const existing = map.get(b.uid)
+      if (!existing || existing === b.uid) map.set(b.uid, b.name || b.uid)
+    }
+    return map
+  })
   cache.set(spaceId, pending)
```
OctoBoooot, line 58: **[minor]** Bot-fetch failure caches a bot-less map with no in-session self-heal. pending is cache.set unconditionally; only members===null evicts (:44). So member-success + bot-fetch-failure caches the full map without bot names -> bots stay raw uids for the rest of the session (no retry until cache clear / space switch). My APPROVE's 'later open retries on failure' is true for the member fetch but NOT the bot fetch — correcting that. If in-session heal is wanted, evict/don't-cache when the bot fetch failed; else document the intentional asymmetry. Credit yujiawei.

OctoBoooot (commented):
## Correction/supplement to my #702 APPROVE — two real resilience minors yujiawei caught

Verdict unchanged (APPROVE — these are minors, non-blocking, and #702 is CLEAN/three-way approved). But my APPROVE praised the resilience with a claim that's only half-true, and yujiawei caught two real refinements. Owning it; credit yujiawei.

### Minor — cache eviction deletes by key, not by identity (narrow same-space race)
`memberNames.ts:44` does `if (members === null) cache.delete(spaceId)` inside call A's `.then`. If a second call B for the same space has already written a **new** pending into the slot (`:58`) before A's `.then` fires, A deletes **B's** valid entry. Reachable only under overlapping same-space calls where A resolves after B's `cache.set`; harm is just an extra refetch (B's in-flight callers still get B's result). Fix: delete by identity — `if (cache.get(spaceId) === pending) cache.delete(spaceId)`.

### Minor — bot-fetch failure caches a bot-less map with no in-session self-heal
`pending` is `cache.set` unconditionally (`:58`); only `members === null` triggers eviction (`:44`). So member-success + **bot-fetch-failure** caches the full map *without* bot names → bots stay as raw uids for the rest of the session (no retry until cache clear / space switch). **My APPROVE said "a fetch failure evicts so a later open retries" — that's true for the member fetch but NOT the bot fetch.** The two fetches are both best-effort for the current render, but only member-failure self-heals; bot-failure sticks. Correcting that overstatement. If in-session self-heal for bot names is wanted, evict (or don't cache) when the bot fetch failed too; otherwise a one-line comment documenting the intentional asymmetry would make it clear.

### Not adopting
`fetchSpaceBotNames` not catching internally (yujiawei's #1) is by-design — the caller wraps it in `.catch(() => [])` (`:40`), so the throw is handled one level up. Catch-at-caller vs catch-at-source is a style choice, not a defect.

### Net (verdict stands)
Still APPROVE — both are minors and the core fix (merge precedence, safe seam, degrade-to-uid, no crash) is correct and well-tested. The correction is to my own praise: the "retries on failure" resilience is member-fetch-only, not symmetric across both fetches. Two easy hardening tweaks (identity-checked eviction + document/heal the bot-failure asymmetry). Credit yujiawei.

OctoBoooot (approved):
## Re-affirming APPROVE on #702 (correcting a review-mechanics side-effect of my earlier COMMENT)

**Verdict: Approve.** My prior COMMENT correction (owning that my resilience praise was member-fetch-only, and adopting yujiawei's two minors) unintentionally *superseded* my earlier APPROVE — GitHub counts each author's latest review, so my COMMENT dropped my effective verdict from APPROVE to none and un-greened the gate. That was a mechanics side-effect, not a change of judgment: I explicitly graded both findings **non-blocking minors**. Re-affirming APPROVE so the record matches my actual position.

To be clear, the two minors still stand as non-blocking hardening notes (unchanged from my correction):
1. **Cache eviction deletes by key, not identity** (`memberNames.ts:44`) — narrow same-space race; fix `if (cache.get(spaceId) === pending) cache.delete(...)`. (Steve byte-verified this is likely unreachable given the synchronous `cache.set` + `if (cached) return` — so it may be belt-and-suspenders; either way non-blocking.)
2. **Bot-fetch failure caches a bot-less map with no in-session self-heal** — member-success + bot-failure caches without bot names, no retry until cache clear. This is the one that corrected my own praise ("retries on failure" is member-fetch-only). Fair minor; optional evict-on-bot-failure fix.

Neither blocks merge. The core fix — merge precedence (never overwrite a human name), safe seam (encoded param, non-array guard), degrade-to-uid, no crash — is correct and well-tested. This is the clean, mergeable pass I assessed originally; my APPROVE stands. Credit yujiawei for the two minors.

yujiawei (approved):
# Code Review — PR #702 (octo-web)

**Verdict: APPROVED** — no correctness/security/build blockers. A few non-blocking suggestions below.

## Summary of the change

Frontend-only fix for the member panel showing a raw uid for non-friend / non-self-created bots. Root cause is accurate: the space-member source (`queryMembers`, behind `fetchAllSpaceMembers`) filters out bots the current user did not create, so those uids never carry a name and the panel falls back to the uid. The fix backfills names with a single non-viewer-scoped `GET /robot/space_bots?space_id=` request per space and merges only the uids that are still missing a real name.

## Verification performed

- Targeted unit tests pass (17/17 across `memberNames.test.ts` and `spaceMembers.test.ts`).
- `tsc --noEmit` reports only two errors, both in files untouched by this PR (`i18n/format.ts`, `members/sort.ts`) — confirmed present on `main` as well, so this PR introduces no new type errors.
- Endpoint path is consistent with existing callers: the docs `apiClient()` inherits `baseURL = /api/v1/` and uses bare-relative paths; `/robot/space_bots` is the same host surface already called by the contacts and BotStore modules, so no new backend contract is introduced.
- Merge logic reviewed: the backfill guard `if (!existing || existing === b.uid)` correctly preserves real human/member names and only fills absent-or-still-raw uids. Both promise legs are independently error-isolated — a `space_bots` failure cannot take down human name resolution.

## Spec compliance

Implements exactly the scope of the linked issue — no under-build, no over-build, no deviation:
- The single-request, no-fanout, no-extra-permission approach matches the stated design.
- Intentionally **not** filtering by `creator_uid` is correct here: self-created bots already resolve via the member source; the non-self-created ones are precisely the uids that need backfilling.
- The returned `Map<string, string>` is consumed purely as a name resolver (`names.get(uid) || uid`); the roster itself is not driven by this map, so backfilling extra bot uids does not inject unwanted entries into the panel.

## Non-blocking suggestions (P2 / nit)

1. **Asymmetric retry on `space_bots` failure (P2).** When the member fetch succeeds but the bot fetch fails, the resolved (bot-less) map is cached and the bot leg never retries — `clearMemberNameCache` currently has no production caller, so recovery only happens on a full page reload, not on reopening/switching the space. This degrades to the pre-existing raw-uid behavior (not a regression), so it does not block merge, but the code comment "forget it so a later open retries" only applies to the member list, not to bots. Consider either evicting the entry when the bot leg fails too, or caching members and bots separately so the bot backfill can retry independently. Worth aligning the comment with the actual behavior either way.
   - `packages/docs/src/members/memberNames.ts` (the `if (members === null) cache.delete(spaceId)` guard covers only the member leg)

2. **Single unpaged `space_bots` request (nit / please confirm).** `fetchAllSpaceMembers` loops pages up to a cap, but `fetchSpaceBotNames` issues one un-paginated GET. If a space can return more bots than a single page, bots beyond the first page would go unnamed. Fine if the endpoint returns all bots in one response — worth a quick confirmation of the backend contract.
   - `packages/docs/src/octoweb/index.ts` (`fetchSpaceBotNames`)

3. **Overwrite guard is a heuristic, not an invariant (nit).** `existing === b.uid` is used as a proxy for "still shows raw uid". A human member whose display name legitimately equals their uid (or was blank → stored as uid) would be treated as unnamed and could be overwritten by a bot row sharing that uid. Real-world likelihood is negligible (uids are opaque), but a one-line comment noting the guard relies on `uid !== human-name` would document the assumption.
   - `packages/docs/src/members/memberNames.ts` (backfill merge guard)

## Coverage notes (blind spots outside the diff)

- The live `/robot/space_bots` response field name (`name`) and envelope shape were not verified against a running backend; the code silently falls back to the uid if the real payload uses a different key or wraps the array. Static consistency with existing callers checked, runtime not exercised here.
- The "non-viewer-scoped" property of the endpoint is a backend guarantee that cannot be verified from the frontend.

Overall the fix is minimal, well-tested, correctly scoped, and error-isolated. Approving; the items above are follow-ups, not blockers.

PR #704, opened 2026-07-13
fix(docs): stop empty @mention/emoji/slash suggestion popup + Esc/outside-click close (#624)

### LINKED ISSUE #624: [Docs] Unexpected floating @mention suggestion popup while editing
## Description

While editing a document, an unexpected floating box/popup suddenly appears. Based on the reviewer's screenshot, this is the **@mention autocomplete dropdown** that pops up when an `@` character is typed (showing member suggestions). The reviewer found this floating box confusing/unexpected.

## Steps to Reproduce

1. Open a document in the editor.
2. Type text that contains an `@` character (e.g. a date or handle like `@...`).
3. A small floating suggestion box appears below the cursor listing member/user suggestions.

## Observed Behavior

A floating popup (member @mention suggestion list) appears unexpectedly, without clear indication of what it is or how to dismiss it. It can feel intrusive when the `@` was not intended to start a mention.

## Expected / Improvement

- The @mention popup should be clearly presented and easy to dismiss (Esc / click away).
- It should only trigger in appropriate contexts, and ideally not fire when `@` is part of ordinary text (e.g. immediately preceded by a non-space character) — align with common editor behavior.
- Consider clearer styling so users understand it is a mention picker.

## Notes

- Reported by reviewer with screenshot.
- Investigate the mention/suggestion extension trigger rules and dismissal behavior in the editor.

ploy-elison:
## What

Fixes the docs editor bug where the `@`-mention suggestion popup floats out unexpectedly and can't be dismissed (octo-web #624). Scope is exactly the approved **A + B** from the root-cause analysis — no trigger-rule changes, no schema/Yjs/mention-attr changes.

The `@`-mention and `:`-emoji popups share one renderer (`suggestionMenu.ts`); the slash menu has an equivalent inline renderer (`SlashCommand.ts`). Both had the same three defects, so both are fixed identically.

### A — zero results render nothing
- The shared renderer no longer mounts its container to `<body>` for an empty result set and no longer paints the `octo-suggest-empty` "—" placeholder; an update that drops results to zero tears the box down.
- The slash renderer gets the same treatment (an empty filtered list mounted an empty bordered box before).

### B — Esc + outside-click close
- Escape now destroys the popup and marks the session dismissed instead of only consuming the key. It no longer claims Escape when nothing is open.
- A capture-phase `document` `mousedown` listener closes the popup on an outside click / blur and is removed on `onExit` (no cross-session leak).

## Changed files
- `packages/docs/src/editor/suggestionMenu.ts` — shared mention/emoji renderer: empty→no-render, Esc destroy, outside-click close, listener cleanup.
- `packages/docs/src/editor/SlashCommand.ts` — same fixes for the slash menu; renderer factory exported for testing.
- `packages/docs/src/editor/styles.css` — remove the now-unused `.octo-suggest-empty` rule.
- `packages/docs/src/editor/suggestionMenu.test.ts` — new jsdom coverage (empty non-render, Esc, outside/inside click, exit cleanup).
- `packages/docs/src/editor/SlashCommand.test.ts` — same coverage for the slash menu.

## Testing
- Added tests were confirmed red on HEAD, green after the fix.
- `packages/docs` vitest: the suggestion/slash suites pass; full suite is 979 passed. The only failures (`EditorShell.test.tsx`, a WebSocket/undici component test) reproduce identically on clean `main` and are unrelated to this change.
- `tsc --noEmit`: no new errors introduced by this change (two pre-existing errors in unrelated files — `dmworkbase/i18n/format.ts`, `members/sort.ts` — exist on `main` unchanged).

Closes #624

### DIFF
--- a/packages/docs/src/editor/SlashCommand.test.ts
+++ b/packages/docs/src/editor/SlashCommand.test.ts
@@ -1,5 +1,5 @@
-import { describe, it, expect } from 'vitest'
-import { SLASH_ITEMS, filterSlashItems } from './SlashCommand.ts'
+import { describe, it, expect, afterEach } from 'vitest'
+import { SLASH_ITEMS, filterSlashItems, createSlashMenuRenderer } from './SlashCommand.ts'
 
 describe('SlashCommand heading items', () => {
   it('exposes a slash entry for every heading level H1–H6', () => {
@@ -25,5 +25,80 @@ describe('SlashCommand heading items', () => {
 
   it('empty query returns the full item list', () => {
     expect(filterSlashItems('')).toEqual(SLASH_ITEMS)
+  })
+})
+
+// #624: the slash menu shares the mention/emoji failure modes — an empty result set must not
+// leave a floating empty box, and Escape / an outside click must close an open menu.
+describe('slash menu — zero results render nothing (#624)', () => {
+  afterEach(() => {
+    document.body.innerHTML = ''
+  })
+
+  const menuCount = () => document.querySelectorAll('.octo-slash-menu').length
+
+  it('appends no menu when onStart receives an empty item list', () => {
+    createSlashMenuRenderer().onStart({ items: [], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(0)
+  })
+
+  it('renders one row per item when items are present', () => {
+    createSlashMenuRenderer().onStart({
+      items: SLASH_ITEMS.slice(0, 2),
+      command: () => {},
+      clientRect: null,
+    })
+    expect(menuCount()).toBe(1)
+    expect(document.querySelectorAll('.octo-slash-item').length).toBe(2)
+  })
+
+  it('removes the box when an update drops the results to zero', () => {
+    const r = createSlashMenuRenderer()
+    r.onStart({ items: SLASH_ITEMS.slice(0, 1), command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(1)
+    r.onUpdate({ items: [], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(0)
+  })
+})
+
+describe('slash menu — Esc + outside click close (#624)', () => {
+  afterEach(() => {
+    document.body.innerHTML = ''
+  })
+
+  const menuCount = () => document.querySelectorAll('.octo-slash-menu').length
+
+  it('Escape destroys the box and reports the key handled', () => {
+    const r = createSlashMenuRenderer()
+    r.onStart({ items: SLASH_ITEMS.slice(0, 2), command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(1)
+    const handled = r.onKeyDown({ event: new KeyboardEvent('keydown', { key: 'Escape' }) })
+    expect(handled).toBe(true)
+    expect(menuCount()).toBe(0)
+  })
+
+  it('a mousedown outside the box closes it', () => {
+    createSlashMenuRenderer().onStart({
+      items: SLASH_ITEMS.slice(0, 2),
+      command: () => {},
+      clientRect: null,
+    })
+    expect(menuCount()).toBe(1)
+    const outside = document.createElement('div')
+    document.body.appendChild(outside)
+    outside.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }))
+    expect(menuCount()).toBe(0)
+  })
+
+  it('onExit detaches the outside listener (no cross-session leak)', () => {
+    const r = createSlashMenuRenderer()
+    r.onStart({ items: SLASH_ITEMS.slice(0, 1), command: () => {}, clientRect: null })
+    r.onExit()
+    createSlashMenuRenderer().onStart({
+      items: SLASH_ITEMS.slice(0, 1),
+      command: () => {},
+      clientRect: null,
+    })
+    expect(menuCount()).toBe(1)
   })
 })
--- a/packages/docs/src/editor/SlashCommand.ts
+++ b/packages/docs/src/editor/SlashCommand.ts
@@ -172,11 +172,15 @@ export function filterSlashItems(query: string): SlashItem[] {
 }
 
 // Minimal keyboard-navigable popup. Kept dependency-free (no tippy) so it runs headless-friendly.
-function createSlashMenuRenderer() {
+export function createSlashMenuRenderer() {
   let el: HTMLDivElement | null = null
   let items: SlashItem[] = []
   let selected = 0
   let cmd: ((item: SlashItem) => void) | null = null
+  // Set once the menu is dismissed (Escape / outside click) so a later onUpdate in the same
+  // suggestion session does not re-open it. Reset by onExit when the session ends (#624).
+  let closed = false
+  let onOutside: ((e: MouseEvent) => void) | null = null
 
   function paint() {
     if (!el) return
@@ -201,6 +205,44 @@ function createSlashMenuRenderer() {
     el.style.top = `${rect.bottom + 4}px`
   }
 
+  /** Tear down the menu element and its outside-click listener. Safe to call repeatedly. */
+  function destroy() {
+    if (onOutside) {
+      document.removeEventListener('mousedown', onOutside, true)
+      onOutside = null
+    }
+    el?.remove()
+    el = null
+  }
+
+  /** Mount the menu element (once) and wire the outside-click dismissal. */
+  function mount() {
+    if (el) return
+    el = document.createElement('div')
+    el.className = 'octo-slash-menu'
+    document.body.appendChild(el)
+    onOutside = (e) => {
+      if (el && (!(e.target instanceof Node) || !el.contains(e.target))) {
+        closed = true
+        destroy()
+      }
+    }
+    // Capture phase so the dismissal runs before ProseMirror handles the click.
+    document.addEventListener('mousedown', onOutside, true)
+  }
+
+  /** Reflect current items into the DOM: empty list renders NO box; non-empty (re)mounts (#624). */
+  function sync(clientRect?: (() => DOMRect | null) | null) {
+    if (closed) return
+    if (items.length === 0) {
+      destroy()
+      return
+    }
+    mount()
+    paint()
+    position(clientRect?.())
+  }
+
   return {
     onStart: (props: {
       items: SlashItem[]
@@ -210,11 +252,8 @@ function createSlashMenuRenderer() {
       items = props.items
       selected = 0
       cmd = props.command
-      el = document.createElement('div')
-      el.className = 'octo-slash-menu'
-      document.body.appendChild(el)
-      paint()
-      position(props.clientRect?.())
+      closed = false
+      sync(props.clientRect)
     },
     onUpdate: (props: {
       items: SlashItem[]
@@ -223,13 +262,18 @@ function createSlashMenuRenderer() {
     }) => {
       items = props.items
       cmd = props.command
-      selected = 0
-      paint()
-      position(props.clientRect?.())
+      selected = Math.min(selected, Math.max(0, items.length - 1))
+      sync(props.clientRect)
     },
     onKeyDown: (props: { event: KeyboardEvent }) => {
-      if (!items.length) return false
       const { key } = props.event
+      if (key === 'Escape') {
+        if (!el) return false
+        closed = true
+        destroy()
+        return true
+      }
+      if (!items.length || !el) return false
       if (key === 'ArrowDown') {
         selected = (selected + 1) % items.length
         paint()
@@ -244,14 +288,11 @@ function createSlashMenuRenderer() {
         cmd?.(items[selected])
         return true
       }
-      if (key === 'Escape') {
-        return true
-      }
       return false
     },
     onExit: () => {
-      el?.remove()
-      el = null
+      closed = false
+      destroy()
     },
   }
 }
--- a/packages/docs/src/editor/styles.css
+++ b/packages/docs/src/editor/styles.css
@@ -2442,11 +2442,6 @@
 .octo-suggest-item:hover {
   background: var(--octo-hover, #f2f3f5);
 }
-.octo-suggest-empty {
-  padding: 6px 10px;
-  color: var(--octo-fg-3, #86909c);
-  font-size: 12px;
-}
 .octo-emoji-menu .octo-suggest-item {
   font-size: 15px;
 }
--- a/packages/docs/src/editor/suggestionMenu.test.ts
+++ b/packages/docs/src/editor/suggestionMenu.test.ts
@@ -1,0 +1,108 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import { createSuggestionMenuRenderer } from './suggestionMenu.ts'
+
+// The @-mention and :-emoji popups share this renderer, so covering it here covers both
+// (octo-web #624): (A) an empty result set must render NO floating box, and (B) Escape and
+// an outside mousedown must close an open popup.
+
+interface Row {
+  id: string
+  label: string
+}
+
+const A: Row = { id: 'a', label: 'Alice' }
+const B: Row = { id: 'b', label: 'Bob' }
+
+const MENU_CLASS = 'octo-test-menu'
+
+function menu() {
+  return createSuggestionMenuRenderer<Row>((r) => r.label, MENU_CLASS)
+}
+
+function menuCount(): number {
+  return document.querySelectorAll(`.${MENU_CLASS}`).length
+}
+
+afterEach(() => {
+  document.body.innerHTML = ''
+})
+
+describe('suggestion menu — A: zero results render nothing (#624)', () => {
+  it('appends no menu element when onStart receives an empty item list', () => {
+    menu().onStart({ items: [], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(0)
+  })
+
+  it('never inserts the "—" empty placeholder', () => {
+    menu().onStart({ items: [], command: () => {}, clientRect: null })
+    expect(document.querySelector('.octo-suggest-empty')).toBeNull()
+  })
+
+  it('renders one row per item when items are present', () => {
+    menu().onStart({ items: [A, B], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(1)
+    expect(document.querySelectorAll('.octo-suggest-item').length).toBe(2)
+  })
+
+  it('removes the box when an update drops the results to zero', () => {
+    const r = menu()
+    r.onStart({ items: [A], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(1)
+    r.onUpdate({ items: [], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(0)
+  })
+
+  it('re-opens the box when a later update brings results back', () => {
+    const r = menu()
+    r.onStart({ items: [], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(0)
+    r.onUpdate({ items: [A], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(1)
+  })
+})
+
+describe('suggestion menu — B: Esc + outside click close (#624)', () => {
+  it('Escape destroys the box and reports the key handled', () => {
+    const r = menu()
+    r.onStart({ items: [A, B], command: () => {}, clientRect: null })
+    expect(menuCount()).toBe(1)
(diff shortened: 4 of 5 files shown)


OctoBoooot (approved):
## Review: fix(docs): stop empty suggestion popup + Esc/outside-click close (#704)

**Verdict:** Approve. Clean, leak-free lifecycle fix with comprehensive tests across both suggestion surfaces. No blockers, no majors.

Base `main`. Risk tier: normal. `mergeStateStatus=CLEAN`, CI green.

### What's good (byte-verified)
- **Empty-popup suppression** — `sync()` calls `destroy()` when `items.length === 0` instead of rendering the "—" box (`suggestionMenu.ts`); the now-dead `.octo-suggest-empty` CSS is removed alongside. No empty popup.
- **Leak-free teardown on every path.** All four dismissal routes funnel through the idempotent `destroy()`, which removes the capture-phase `mousedown` listener and nulls the handle: Escape (`:118-125`, only claims the key when `el` exists so Esc stays available to other handlers otherwise), outside-click (`onOutside`), empty-items (`sync`), and `onExit` (`:143-146`). I checked each — no teardown path leaves the document listener attached.
- **`closed` latch fixes the #624 re-open bug correctly.** Set on Escape/outside-dismiss so a later `onUpdate` in the same session doesn't re-open the popup, and reset **only** by `onExit` (true session end). The test suite proves this doesn't over-latch — "re-opens the box when a later update brings results back" passes, so within a live (non-dismissed) session results can still reappear.
- **No XSS** — rows use `textContent` (`:45`), not innerHTML.
- **Consistent across both surfaces** — `SlashCommand.ts` applies the identical pattern (same `closed` latch, `destroy()` with `removeEventListener`, empty→destroy, Esc/outside handling), so the mention/emoji and slash menus behave the same.

### Praise
The `suggestionMenu.test.ts` suite (+108) is thorough and targets exactly the tricky cases: empty→no element, drop-to-zero→remove, **re-open when results return**, Escape→destroy+handled, **outside closes / inside does NOT**, **stays closed after Escape until exit**, and **`onExit` detaches the outside listener (no cross-session leak)**. That last one directly guards the listener-leak risk. `SlashCommand.test.ts` (+77) mirrors it. Real behavioral coverage, not hollow.

### Net
Correct fix (no empty popup, Esc/outside dismissal, no re-open regression), no listener leaks on any teardown, dead CSS cleaned up, and matching tests + implementation across both suggestion menus. Approving.

Jerry-Xin (approved):
Line-by-line review of head 453c3408 against main. Verdict: **APPROVED**.

**A — empty result renders nothing.** `paint()` no longer inserts the `.octo-suggest-empty` placeholder; `sync()` tears the box down when `items.length === 0`. The stale CSS rule is removed and nothing else references it. The shared renderer backs both @mention and emoji, and `SlashCommand.ts` mirrors the same logic — all three channels are consistent.

**B — Escape / outside-click close.** Escape now actually `destroy()`s the element and only claims the key when a popup is open (`if (!el) return false`), so it stays available to other handlers otherwise. Outside dismissal is a capture-phase `document` `mousedown` that checks `el.contains(target)`. The `closed` flag stops a same-session `onUpdate` from re-opening a dismissed popup and is reset in `onExit`.

**Listener hygiene.** `destroy()` removes the `mousedown` listener and nulls the handle; `mount()` guards against a second element/listener. Binding and unbinding are balanced, and the cross-session no-leak case is covered by a test.

**Scope.** No changes to mention/emoji trigger config (`char`/`allow`), node attrs, serialization, or schema — only the renderer and the shared menu logic. `selected` is now clamped on update instead of reset, a small correctness improvement.

**Tests.** The two new suites pass (20 tests) and genuinely fail against the pre-fix source, so the added coverage is real, not vacuous.

Non-blocking note (follow-up, not required for this PR): dismissal is wired to outside `mousedown` only. A keyboard tab-away or window blur that moves focus without a click won't close the popup. Outside-click + Escape cover the reported defect; a `focusout` fallback would close the remaining gap if it ever surfaces.

Jerry-Xin (approved):
Reviewed at head `453c3408`, byte-verified (merge-base == base.sha, true incremental diff; 5 files all in `packages/docs/src/editor/`). **APPROVE** — correct fix, no leaks, thorough tests, CI green.

## ✅ Empty-result-set suppression (core fix) — correct, no regression
Semantic note: this is empty-**result-set** suppression, not empty-*query*. On an empty query `items()` still returns the full list (`filterSlashItems('') === SLASH_ITEMS`), so typing `@`/`/`/`:` alone still shows a populated popup. The bug fixed is the empty `—` placeholder box that showed when the query matched **zero** items. Gate (`suggestionMenu.ts:93`): `if (items.length === 0) { destroy(); return }` → zero results = no box; ≥1 = mount+paint. The old `.octo-suggest-empty` `—` branch is deleted. `@`→`@x` transition re-runs `sync` and re-mounts on a match — no off-by-one, no valid-suggestion suppression.

## ✅ Esc / outside-click — listeners properly cleaned up, no leak
- Add (`:84`): `document.addEventListener('mousedown', onOutside, true)` (capture); Remove (`destroy()` `:64`): `document.removeEventListener('mousedown', onOutside, true)` — **same stored `onOutside` reference + same capture flag** → the removal genuinely matches; `onOutside=null` after. `destroy()` is idempotent.
- `onExit` → `destroy()`, and `render: () => createSuggestionMenuRenderer(...)` builds a **fresh renderer per suggestion session** (own `el`/`onOutside`/`closed`) → no shared module state, no cross-session leak.
- **Esc scoping (`:118`):** only claims Escape when a popup is actually open (`if (!el) return false`) — Esc stays available to other UI (fixes the old unconditional `return true` that swallowed Esc globally).
- Outside detection (`:78`): `!el.contains(e.target)` — correct for a `document.body`-appended non-shadow popup.

## ✅ styles.css −5 — dead CSS
Removed `.octo-suggest-empty { … }` — the style for the `—` placeholder that no longer exists; layout/positioning rules untouched. Safe.

## ✅ Tests — genuine, both directions
`suggestionMenu.test.ts` (+108) + `SlashCommand.test.ts` (+77) assert: empty→no box, populated→box+N rows, drop-to-zero removes box, results-return re-opens, **Esc closes + stays closed on later update** (the `closed`-flag regression guard), **outside mousedown closes / inside does NOT**, and explicitly **`onExit` detaches the listener with no cross-session double-close**. No weakened assertions.

## ✅ No XSS / CI green
Rows use `row.textContent = renderItem(item)` (`:45`), not innerHTML. Live CI: `code-review` pass, all required checks pass/skip — no failing/pending check (checked before approving).

## 🟡 Non-blocking (a11y polish, pre-existing)
Popup rows are focusable `<button>` with Arrow/Enter via TipTap onKeyDown, but the container lacks `role="listbox"`/`aria-activedescendant`. Esc doesn't explicitly restore focus, but focus never leaves the editor (menu is a detached body div, rows activate via `mousedown` + `preventDefault`), so the caret stays put. Pre-existing pattern, not a regression — optional polish.

Net: correct, leak-free, well-tested suggestion-popup fix consistent across all three surfaces (mention/emoji/slash) via the shared renderer. No blockers. **APPROVE.**

yujiawei (approved):
# Code Review — PR #704 (octo-web)

**Verdict: APPROVED.** The fix is well-scoped, correct, and covered by new tests. No blocking issues. Three non-blocking P2 suggestions below.

## Scope check
The diff is exactly the approved A + B: empty result set renders no box, and Escape / outside-click close the popup. Both the shared mention/emoji renderer (`suggestionMenu.ts`) and the equivalent slash renderer (`SlashCommand.ts`) get the identical treatment. No trigger-rule, schema/Yjs, or mention-attribute changes crept in. `styles.css` correctly drops the now-unused `.octo-suggest-empty` rule, and there are no dangling references to it in production code.

## What's correct
- **`closed` lifecycle** — reset in `onStart` and `onExit`, set on Escape/outside-click, and honored by `sync()` so a later `onUpdate` in the same suggestion session can't re-open a dismissed popup. Because `render()` builds a fresh renderer per session, the flag can't leak across sessions.
- **Escape claim conditions** (`suggestionMenu.ts:118-125`) — returns `false` when no popup is open, so Escape stays available to other handlers when nothing is showing; only claims the key when `el` exists. Good.
- **Outside-click listener** — added in `mount()`, removed in `destroy()` (which also runs on `onExit`), so no listener leak. Capture phase is the right call so dismissal runs before ProseMirror processes the click; the inside-click test confirms a `mousedown` on a row does not close the box.
- **`selected` clamping** (`onUpdate`) — `Math.min(selected, Math.max(0, items.length - 1))` avoids an out-of-range index when the filtered list shrinks. Enter is guarded by the `!items.length || !el` early-return.

## Testing
20/20 tests pass across the two new suites (empty non-render, update→zero teardown, re-open, Esc, outside/inside click, exit cleanup) for both renderers. `tsc --noEmit` introduces no new errors — the two failures (`dmworkbase/i18n/format.ts`, `members/sort.ts`) reproduce on `main` and are unrelated, as the PR notes.

1. **Test doesn't actually assert listener detachment.** The "no cross-session leak" test (`suggestionMenu.test.ts:99-108`, `SlashCommand.test.ts:67-77`) passes even if `document.removeEventListener(...)` is deleted from `destroy()` — I verified this by neutering that line and all 10 tests still passed. The production code is correct (the leaked closure would be inert anyway, since it guards on `el` which is null after teardown), but the test as written wouldn't catch a future regression. To make it a real guard, spy on `document.removeEventListener` and assert it was called for the `mousedown` capture listener.

2. **Outside-dismiss is `mousedown`-only.** On touch/pointer devices, a UI that consumes `pointerdown`/`touchstart` and calls `preventDefault()` can suppress the synthetic `mousedown`, leaving the popup open on an outside tap. Consider `pointerdown` (or both) for more robust outside-click detection. Low priority if the docs editor is desktop-first.

3. **Renderer duplication.** `createSlashMenuRenderer` and `createSuggestionMenuRenderer` now share near-identical popup lifecycle logic (`mount`/`destroy`/`sync`/`closed`/`onKeyDown`). Fine to ship, but any future change (ARIA attributes, focus handling, positioning) has to be applied in two places. Worth extracting the headless lifecycle into one shared helper in a follow-up.

## Reviewer coverage note
This is a self-contained frontend change; behavior was verified against the jsdom suites and a line-level read of the lifecycle. Not exercised here: real ProseMirror integration (whether Tiptap's `onExit` reliably fires on every abrupt unmount / caret-move path so `closed` resets for the next `@`) and live touch-device behavior — both outside the diff's unit scope and best confirmed by manual QA in the editor.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #704 (octo-web)
**Head SHA:** `453c3408427b1b098322ec84cccc30e634ff728a`

## Summary

This PR fixes three related UX issues with the editor suggestion popups (@-mention, :-emoji, and /-slash): (A) an empty result set no longer renders a floating empty box, (B) Escape closes an open popup and correctly reports the key as handled, and (C) clicking outside the popup dismisses it. The implementation introduces a `closed` latch, `destroy()`/`mount()`/`sync()` helper decomposition, and a capture-phase outside-click listener — all applied symmetrically to both `suggestionMenu.ts` and `SlashCommand.ts`. Tests are added for both renderers.

The fix is well-scoped and the production logic is sound. One non-blocking test-parity gap noted below.

## Verification

- ✅ **Empty-items guard** — `sync()` calls `destroy()` when `items.length === 0`, so no floating box is ever mounted for empty results. The old `paint()` branch that rendered the `—` placeholder is removed, along with the dead `.octo-suggest-empty` CSS rule.
- ✅ **`closed` latch correctness** — Set to `true` on Escape (when popup exists) and on outside click. Reset to `false` on `onStart` and `onExit`. The latch correctly prevents `sync()` from re-mounting after dismissal within the same suggestion session.
- ✅ **`destroy()` idempotency** — Removes the `onOutside` listener (if attached), removes the DOM element, and nulls both references. Safe to call repeatedly; no double-remove or stale-listener risk.
- ✅ **Outside-click capture phase** — `document.addEventListener('mousedown', onOutside, true)` ensures dismissal fires before ProseMirror processes the click. The `el.contains(e.target)` guard correctly keeps inside clicks from closing the popup.
- ✅ **Escape key contract change** — Old code returned `true` for Escape unconditionally (even with no popup open), swallowing the key from other handlers. New code returns `false` when `!el`, correctly propagating Escape to other handlers when no popup is showing.
- ✅ **`selected` index clamping** — `Math.min(selected, Math.max(0, items.length - 1))` preserves arrow-key navigation across `onUpdate` calls while safely clamping to valid bounds. When `items.length === 0`, `selected` becomes `0` but `sync()` will `destroy()` before `selected` is consumed.
- ✅ **Data flow from callers** — `mention.ts:123` and `emoji.ts:94` both pass `createSuggestionMenuRenderer()` as the `render` callback to `@tiptap/suggestion`, which manages the `onStart → onUpdate* → onExit/onKeyDown` lifecycle. `SlashCommand.ts:311` does the same. Each suggestion session creates a fresh renderer closure, so no cross-session state leakage is possible.
- ✅ **CSS cleanup** — `.octo-suggest-empty` removed from `styles.css:2444`. Grep confirms no remaining production reference (only the test assertion that it is absent).

Static analysis only at head `453c3408427b1b098322ec84cccc30e634ff728a`; build and tests not executed in this environment.

## Findings

### P2 — Test parity gap in `SlashCommand.test.ts` (`packages/docs/src/editor/SlashCommand.test.ts:89`)

`suggestionMenu.test.ts` covers three scenarios that `SlashCommand.test.ts` does not:

1. **"a mousedown inside the box does NOT close it"** — verifies the `el.contains(e.target)` guard.
2. **"stays closed on a later update after Escape"** — verifies the `closed` latch survives `onUpdate`.
3. **"re-opens the box when a later update brings results back"** — verifies the `items=[] → items=[A]` transition.

Since `createSlashMenuRenderer` and `createSuggestionMenuRenderer` are structurally duplicated (same `destroy`/`mount`/`sync`/`closed` logic), a future edit could touch one without the other. Adding these three tests to `SlashCommand.test.ts` would lock the parity contract on both sides and guard against asymmetric regression.

**Diff-scope:** new (this PR introduces both the production logic and the test files; the gap is in the newly-added tests).

- `destroy()` is safe to call when `el` is already null and `onOutside` is already null — both guards short-circuit.
- The `onExit` handler resets `closed = false` before `destroy()`. Since `@tiptap/suggestion` creates a new renderer per session, the reset is defensive but harmless.
- `onKeyDown` for arrow keys and Enter is guarded by `!items.length || !el`, preventing phantom selection when the popup was destroyed by a prior `sync()`.
- The `e.target instanceof Node` check in the outside-click handler is a valid defense for edge cases where `e.target` might not be a DOM Node.
- No dangling references to `.octo-suggest-empty` in production code.

## Data Flow Trace

|---|---|---|
| `items` in `sync()` | `onStart(props.items)` / `onUpdate(props.items)` from `@tiptap/suggestion` plugin | ✅ Yes — assigned to closure variable before `sync()` call |
| `closed` in `sync()` | Set `true` by Escape handler or `onOutside` handler; reset `false` by `onStart` / `onExit` | ✅ Yes — latch prevents re-mount after intentional dismissal |
| `onOutside` in `destroy()` | Created in `mount()`, nulled in `destroy()` | ✅ Yes — `removeEventListener` called with same ref + capture flag |
| `selected` in `paint()` | Set in `onStart` (to 0), `onUpdate` (clamped), arrow-key handlers (modulo wrap) | ✅ Yes — clamped to `[0, items.length-1]` before `paint()` consumes it |
| `cmd` in row mousedown | Assigned from `props.command` in `onStart`/`onUpdate` | ✅ Yes — `cmd?.(item)` safely handles null |
| `clientRect` in `position()` | Passed via `sync(clientRect)` from `onStart`/`onUpdate` `props.clientRect` | ✅ Yes — `position()` guards `!rect` |

## Blindspot Checklist

- **C1 (Dual-path parity):** N/A — no add↔remove / create↔delete symmetric paths changed.
- **C2 (Control-flow ordering):** Clear — `destroy()` is idempotent and called from three sites (`sync`, Escape handler, `onOutside`); no double-action risk. `sync()` short-circuits on `closed` before `destroy()`, so no ordering issue.
- **C3 (Auth boundary):** N/A — no auth/permission changes.
- **C4 (Auth lifecycle):** N/A — no auth changes.
- **C5 (Build/runtime):** N/A — no build/deploy changes; this is a frontend UI fix.
- **C6 (Governance/docs):** N/A — no governance/doc changes.

## Cross-round Blocker Review

## Verdict: APPROVED

No correctness, security, or functional issues. The `closed` latch, listener lifecycle, and empty-items guard are all implemented correctly. The P2 test-parity gap is non-blocking — worth closing for regression safety but not a merge gate.

[Octo-Q] verdict: APPROVE — No P0/P1. One P2 (test parity gap) is non-blocking. Production logic is sound: `closed` latch, `destroy()` idempotency, outside-click capture-phase listener, and Escape key contract change are all correctly implemented.

yujiawei (approved):
# Code Review — PR #704 (octo-web)

Reviewed at head `453c3408427b1b098322ec84cccc30e634ff728a` against `main`. Scope is the docs-editor suggestion-popup fix for #624: (A) an empty result set must render no floating box, and (B) Escape and an outside click must dismiss an open popup. Two renderers are touched — the shared `@`-mention/`:`-emoji renderer (`suggestionMenu.ts`) and the inline slash renderer (`SlashCommand.ts`) — plus a dead CSS rule and new jsdom tests.

## 1. Spec compliance

**Result: meets scope, nothing extra.**

- **A — empty renders nothing.** `paint()` no longer emits the `octo-suggest-empty` "—" placeholder; the new `sync()` calls `destroy()` when `items.length === 0`, so an update that drops results to zero tears the box down instead of leaving a floating empty container. Applied identically in both `suggestionMenu.ts` and `SlashCommand.ts`.
- **B — Esc + outside-click close.** Escape now destroys the popup and sets a session-scoped `closed` flag (`suggestionMenu.ts` onKeyDown; same in `SlashCommand.ts`), and it returns `false` when nothing is open so Escape stays available to other handlers. A capture-phase `document` `mousedown` listener, wired in `mount()` and torn down in `destroy()`, closes on an outside click.
- **No scope creep.** The diff stays inside the two renderers, the CSS rule removal, and tests. No trigger-rule, schema, Yjs, or mention-attribute changes — consistent with the PR description. `Closes #624` is appropriate.

## 2. Code quality

**Result: approve.**

Verified line by line for the failure modes this class of bug invites:

- **Listener balance / no leak.** `mount()` guards on `if (el) return`, so exactly one `mousedown` listener exists at a time; `destroy()` removes it and nulls `onOutside`; `onExit()` calls `destroy()`. The `onExit removes ... no cross-session leak` test confirms a stale listener from a prior session can't double-close the next one.
- **Event ordering is correct.** The outside-click handler is registered in the **capture** phase, so it runs before ProseMirror handles the click; when the target is inside `el`, `el.contains(e.target)` short-circuits it and the row's own `mousedown` → `cmd()` still fires. The "mousedown inside does NOT close it" test covers this.
- **Dismiss latch behaves.** `closed` is set on Esc/outside-click, checked at the top of `sync()`, and reset in `onStart`/`onExit`. This is what keeps a later `onUpdate` in the same suggestion session from re-opening a popup the user just dismissed — and the async mention path (start empty → update with results) still opens correctly because `onStart` resets `closed` (covered by "re-opens the box when a later update brings results back").
- **Minor within-scope polish, not over-build.** `onUpdate` now clamps `selected` instead of resetting to 0, keeping the highlight stable as the query narrows. `row.textContent` (not innerHTML) keeps rendering injection-safe.
- **Dead CSS fully removed.** The `.octo-suggest-empty` rule is gone from `styles.css` and has no remaining references in the codebase (only a negative test assertion). Confirmed by grep.

### Verification performed
- `tsc --noEmit` (docs package): no new errors. The two pre-existing errors (`dmworkbase/i18n/format.ts`, `members/sort.ts`) exist on `main` and are untouched here.
- New suites: `suggestionMenu.test.ts` + `SlashCommand.test.ts` → 20/20 pass.
- Full docs suite: 979 passed, 4 failed — all four in `EditorShell.test.tsx` (an unrelated i18n key-resolution failure in export-filename/header/more-menu tests). That file is not touched by this PR and has no relationship to the suggestion renderers; the failures reproduce independently of this change, as the PR notes.

### Suggestions (non-blocking, P2)
- `suggestionMenu.ts` and `SlashCommand.ts` now carry near-identical `destroy`/`mount`/`sync`/`closed`/`onOutside` blocks. The slash menu having its own inline copy predates this PR, but with both now sharing the same lifecycle logic, folding the slash renderer onto the generic `createSuggestionMenuRenderer` (passing item type + `renderItem`) would remove ~50 duplicated lines and keep the two from drifting. Worth a follow-up.

## Verdict

Spec compliant and correct. No blocking findings. Approving.

PR #700, opened 2026-07-13
feat(docs): body font family (fontFamily) with feature-flag gating (SCHEMA_VERSION 16)

ploy-elison:
## Summary

Adds body **font family** support to the docs editor (`packages/docs`), byte-aligned with the backend `fontFamily` attribute under the shared **SCHEMA_VERSION 16** contract. Mirrors the existing `fontSize` path (a `textStyle` global attr → `<span style="font-family:…">`), with `parseDOM`/`toDOM` and Y.Doc ↔ ProseMirror round-trip.

Rollout is phased and seamless: the frontend can deploy to everyone, but the font entry stays gated behind a flag that is **off by default**.

## Changes

- **Extension** (`editor/extensions.ts`): register `FontFamily` (from `@tiptap/extension-text-style`, same package as `FontSize`) after `TextStyle` in both the live editor and the read-only preview. Registration is **unconditional** — the schema must always know the attr so this bundle round-trips fonts faithfully and never strips them. That is precisely what makes the phased rollout safe.
- **Feature flag** (`config.ts`): `FONT_FAMILY_ENABLED` (via `VITE_DOCS_FONT_FAMILY`), **default off**. It gates **only** the toolbar entry — the control the user uses to *set* a font. When off, the selector is not rendered at all (invisible/unusable), so there is no toolbar regression. Keeping the entry off until client versions converge prevents an older bundle (no `fontFamily` in its schema) from silently dropping a font set by a newer client. Turning it on is a separate decision, not part of this change.
- **Toolbar** (`editor/Toolbar.tsx`): `FontFamilySelect` mirroring `FontSizeSelect`, a curated CJK + Western preset list (each value carries a generic fallback), rendered only when the flag is on.
- **Schema** (`schema/index.ts`): bump `SCHEMA_VERSION` 15 → 16 with a cumulative changelog entry. `fontFamily` is an attribute (like `fontSize`/`textAlign`), so it stays out of the node/mark audit lists.
- **i18n** (`zh-CN` / `en-US`): `toolbar.fontFamily` + `toolbar.fontFamilyDefault` (parity-checked).
- **DOCX export** (`export/docx/marks.ts`): map `fontFamily` → run `font` (first family, quotes stripped, generic CSS keywords dropped; the `code` mark's monospace font still wins).

## Backward compatibility

Old documents without `fontFamily` render unchanged; existing color / fontSize / alignment behaviour is untouched.

## Testing

[...]

PR #695, opened 2026-07-13
feat(chat-header): add search entry to open in-channel search

BrDing-Rookie:
# feat(chat-header): add search entry to open in-channel search

Adds a search icon to the main chat header (right side), matching the visual style of the summary/matter entries. Clicking opens the in-channel ChannelSearch panel — the same effect as opening the conversation info sidebar → "查找聊天内容".

## Background

This was originally bundled into PR #688 (4 global-search filter-panel bug fixes, YUJ-30). Reviewers correctly flagged it as out-of-scope for the bug-fix issue (#687), and it has been split out per their request. The commit was reverted from `fix/global-search-filter-4bugs-fe` and cherry-picked here onto a fresh branch from `main`.

## Implementation

Mirror of the summary/matter entry-button pattern:

- `packages/dmworkbase/src/App.tsx` — new `wk:open-channel-search` mittBus event type.
- `packages/dmworkbase/src/Components/ChannelSearch/ChatSearchEntryButton.tsx` (new, +32) — lucide `Search` icon, gated by `isChannelSearchEnabled(channel)`, tooltip reuses existing i18n key `module.channelSettings.messageHistory`.
- `packages/dmworkbase/src/module.tsx` — registers the button via `registerChannelHeaderRightItem("channelheader.search", ...)` at sort=4900 (before summary=5100 and matter=5000).
- `packages/dmworkbase/src/Pages/Chat/index.tsx` — mittBus listener registered in `componentDidMount` with channel-match guard, cleaned up in `componentWillUnmount`; handler invokes existing `_openChannelSearchPanel()` (same feature gate + side-panel mutual exclusion as the info-sidebar entry).

## Tests

- 4 new unit tests for `ChatSearchEntryButton` (render gating, click emit).
- `pnpm build` / `lint` / `i18n:check` green.

## Verification

Design/mockup: `global-search-design/feat-chat-header-search-entry.md` (spec doc referenced by the commit).

Fixes #694

PR #714, opened 2026-07-13
fix(loop): drive board & transcript refresh via realtime WS, remove blind polling

### LINKED ISSUE #713: bug: Loop board does not auto-refresh after dispatch (needs manual reload), and settle-poll flickers the board
## Symptom
After a quick-create dispatch on the Loop board (assigning an AI teammate):
1. The new loop (issue) does **not** appear on the board automatically — a full-page reload is required to see it;
2. The existing timer-based settle-poll repeatedly flashes the board into its loading state (flicker).

## Repro (dev, real env)
1. Open the Loop board;
2. "New loop" → type a one-liner + assign an AI teammate → dispatch;
3. Watch the board: the new issue does not show for ~10+s (the agent creates the issue asynchronously; the DB-landing moment is unobservable), and the board flickers meanwhile;
4. Manually reload the page → the new issue finally appears.

## Expected
The new issue should appear on the board **automatically and silently** the moment it is actually created — no manual reload, no flicker. Status changes / edits by others should reflect just as promptly.

## Root cause
dmloop has no realtime channel and relies on a stopgap that blindly re-polls the list after dispatch at `[2,5,9,14]s`: too few polls miss the async DB-landing window (issue never shows), more polls flash the board (flicker) — blind polling can't win. Sibling blind polls exist too: running-chip 15s, RunDetailModal transcript 2s, squad presence 30s.

## Fix direction
Subscribe to the backend's existing realtime event stream (WebSocket `/ws`, per-workspace scope, issue/task events); refetch silently once when an event actually arrives, replacing all blind polling. Fix PR to follow.

MrAladdin:
Closes #713

## Problem
After a quick-create dispatch on the Loop board, the new issue did **not** appear without a manual full-page reload, and the timer-based settle-poll flashed the board (loading state) repeatedly. Root cause: dmloop had no realtime channel, so it *guessed* when the async agent-created issue would land — too few polls miss the window (issue never shows), more polls flicker. Blind polling can't win.

## Fix
Subscribe to the backend's existing realtime stream instead of guessing. A small WS client (`packages/dmloop/src/api/ws.ts`) connects per-workspace, and a single debounced silent refetch runs when the backend actually reports a change. This removes **all** blind polling in dmloop (Rule-9 class fix):

| Poll removed | Replaced by |
|---|---|
| board settle-poll `[2,5,9,14]s` | WS `issue:*` / `task:*` → `wk:loop-issues-refresh` (debounced 250ms) |
| IssuePage running-chip 15s | same WS refresh signal |
| RunDetailModal transcript 2s | WS `activity:created`/`task:progress` (transcript) + `wk:loop-issues-refresh` (liveness/terminal/gone) |
| SquadDetailPage presence 30s | WS `agent:status` |

Key points:
- **Auth**: token-mode handshake — first frame `{type:"auth",payload:{token}}` with the session token from the existing `/cli-token` endpoint (browsers can't set WS handshake headers). No new endpoint, no new backend/service.
- **De-flicker**: `IssuePage.reload(silent)` — WS-driven refetches don't toggle the loading state; only initial load / filter change / view switch show loading.
- **Resilience**: 3s auto-reconnect; `auth_ack` re-emits the refresh signal so a reconnect backfills anything missed while down.
- **RunDetailModal** keeps its terminal final-fetch (#610) + gone guard, now driven by the workspace-wide `wk:loop-issues-refresh` signal (covers `issue:deleted` / terminal / reconnect) so liveness/gone detection can't freeze.

## Blast radius (Rule 8)
- `packages/dmworkbase/src/App.tsx` (shared `MittEvents`): removed the now-unused `wk:loop-issues-dispatched` key (grep confirms zero remaining refs repo-wide); kept `wk:loop-issues-refresh`. No other package consumes these.
- New `loopWs` singleton is consumed only within `packages/dmloop` (LoopPage/IssuePage/RunDetailModal/SquadDetailPage).
- `apps/web/vite.config.ts`: additive `/fleet/ws` proxy rule; no change to existing routes.

## Verification
- **Real-env e2e (dev, not just unit tests)**: dispatched via the composer — the new issue (`ALI-10`) auto-appeared on the board with no manual reload (待办 → 运行中 driven by WS `task:*`), no flicker, console error-free. Re-verified after applying review fixes. WS handshake confirmed against the backend (`auth_ack`; dispatch pushes `issue:created` + `task:*` + `activity:created`).
- **Unit tests**: added `packages/dmloop/src/api/__tests__/ws.test.ts` (6 cases: debounce coalescing, non-board events don't refresh, `auth_ack` refresh, `on()` dispatch/unsubscribe, malformed-frame safety). `pnpm --filter @octo/loop test` → 23/23 pass.
- `tsc` clean on changed files.

## Review
- `/simplify` (4 angles) applied. Adversarial code review found one real bug — RunDetailModal could freeze "running" forever if the run vanished via `issue:deleted` (event not in its old whitelist) — fixed by subscribing to the broad `wk:loop-issues-refresh` signal.

## Deploy coordination (cross-end follow-up)
Production gateway must proxy `/fleet/ws` → backend `/ws` **with WebSocket upgrade** (prefer same-origin so the backend `checkOrigin` same-origin branch passes; otherwise set the backend `ALLOWED_ORIGINS` to include the octo-web origin). Dev is handled by the vite `/fleet/ws` rule (`ws:true`, `changeOrigin:false`). No backend deploy otherwise.

### DIFF
--- a/packages/dmloop/src/api/ws.ts
+++ b/packages/dmloop/src/api/ws.ts
@@ -1,0 +1,139 @@
+// @octo/loop 实时 WebSocket 客户端:订阅 workspace 事件驱动看板/面板刷新。
+// token-mode 握手(浏览器 WS 不能设 header):连上后首帧 {type:"auth",payload:{token}},
+// token 取自 /cli-token。dev 经 vite `/fleet/ws` → 后端 /ws;VITE_FLEET_WS_URL 可覆盖直连。
+import { WKApp } from "@octo/base";
+import { issueLoopCliToken } from "./authApi";
+
+// 重连退避:base 起步、指数封顶,带抖动——避免 backend 重启后众客户端齐步重连、也避免死循环打满。
+const RECONNECT_BASE_MS = 2000;
+const RECONNECT_MAX_MS = 30000;
+// 一次派单连发多帧(issue:created→task:dispatch→…),去抖合并成一次看板重取。
+const REFRESH_DEBOUNCE_MS = 250;
+
+// 触发看板重取的事件。不含 task:progress/activity:created(高频、只关 transcript)、
+// agent:status(presence)——后者由 SquadDetailPage 经 on() 单订。
+const REFRESH_EVENTS = new Set([
+  "issue:created",
+  "issue:updated",
+  "issue:deleted",
+  "task:dispatch",
+  "task:queued",
+  "task:running",
+  "task:completed",
+  "task:failed",
+]);
+
+function wsUrl(slug: string): string {
+  const base = (import.meta as { env?: Record<string, string> }).env?.VITE_FLEET_WS_URL;
+  const q = `?workspace_slug=${encodeURIComponent(slug)}&client_platform=web`;
+  if (base) return `${base.replace(/\/$/, "")}/ws${q}`;
+  const proto = location.protocol === "https:" ? "wss" : "ws";
+  return `${proto}://${location.host}/fleet/ws${q}`;
+}
+
+// 单例:LoopPage 按 workspace start/stop。看板等域级刷新订全局 `wk:loop-issues-refresh`;
+// transcript/presence 等高频原始事件用 on(type) 精确订。
+class LoopWs {
+  private ws: WebSocket | null = null;
+  private slug = "";
+  private reconnectTimer: number | null = null;
+  private reconnectAttempts = 0;
+  private refreshTimer: number | null = null;
+  private stopped = true;
+  private handlers = new Map<string, Set<(payload: unknown) => void>>();
+
+  on(type: string, handler: (payload: unknown) => void): () => void {
+    let set = this.handlers.get(type);
+    if (!set) { set = new Set(); this.handlers.set(type, set); }
+    set.add(handler);
+    return () => { this.handlers.get(type)?.delete(handler); };
+  }
+
+  start(slug: string): void {
+    if (!slug) { this.stop(); return; }
+    if (!this.stopped && this.slug === slug) return;
+    this.stop();
+    this.slug = slug;
+    this.stopped = false;
+    void this.open();
+  }
+
+  stop(): void {
+    this.stopped = true;
+    this.slug = "";
+    this.reconnectAttempts = 0;
+    if (this.reconnectTimer !== null) { clearTimeout(this.reconnectTimer); this.reconnectTimer = null; }
+    if (this.refreshTimer !== null) { clearTimeout(this.refreshTimer); this.refreshTimer = null; }
+    if (this.ws) {
+      // 先摘全部回调再 close,避免主动断开触发重连、或在途帧影响新上下文。
+      this.ws.onclose = null;
+      this.ws.onerror = null;
+      this.ws.onmessage = null;
+      this.ws.onopen = null;
+      try { this.ws.close(); } catch { /* already closing */ }
+      this.ws = null;
+    }
+  }
+
+  private scheduleRefresh(): void {
+    if (this.refreshTimer !== null) return;
+    this.refreshTimer = window.setTimeout(() => {
+      this.refreshTimer = null;
+      WKApp.mittBus.emit("wk:loop-issues-refresh");
+    }, REFRESH_DEBOUNCE_MS);
+  }
+
+  private async open(): Promise<void> {
+    if (this.stopped) return;
+    const slug = this.slug;
+    let token: string;
+    try {
+      token = (await issueLoopCliToken()).token;
+    } catch {
+      this.scheduleReconnect();
+      return;
+    }
+    // 取 token 期间可能已 stop/切 workspace。
+    if (this.stopped || slug !== this.slug) return;
+
+    // 构造也可能同步抛(如 VITE_FLEET_WS_URL 畸形),不裹会逃出 open() 不排重连 → 单例僵死。
+    let ws: WebSocket;
+    try { ws = new WebSocket(wsUrl(slug)); } catch { this.scheduleReconnect(); return; }
+    this.ws = ws;
+
+    ws.onopen = () => {
+      if (this.ws === ws) ws.send(JSON.stringify({ type: "auth", payload: { token } }));
+    };
+    ws.onmessage = (e) => {
+      if (this.ws !== ws) return; // 已被切换/stop 取代的旧 socket 排队帧不得影响新上下文
+      let msg: { type?: unknown; payload?: unknown };
+      try { msg = JSON.parse(typeof e.data === "string" ? e.data : ""); } catch { return; }
+      if (!msg || typeof msg.type !== "string") return;
+      // auth_ack:(重)连成功。重置退避计数,并刷新一次补齐断连期漏掉的变更。
+      if (msg.type === "auth_ack") { this.reconnectAttempts = 0; this.scheduleRefresh(); return; }
+      if (REFRESH_EVENTS.has(msg.type)) this.scheduleRefresh();
+      const hs = this.handlers.get(msg.type);
+      if (hs) for (const h of hs) { try { h(msg.payload); } catch { /* handler 自负 */ } }
+    };
+    ws.onclose = () => {
+      if (this.ws !== ws) return;
+      this.ws = null;
+      this.scheduleReconnect();
+    };
+    ws.onerror = () => { /* onclose 负责重连 */ };
+  }
+
+  private scheduleReconnect(): void {
+    if (this.stopped || this.reconnectTimer !== null) return;
+    // 指数退避封顶 + 抖动([capped/2, capped)):backend 重启众客户端不齐步,持续失败也不打满。
+    const capped = Math.min(RECONNECT_MAX_MS, RECONNECT_BASE_MS * 2 ** this.reconnectAttempts);
+    this.reconnectAttempts += 1;
+    const delay = capped / 2 + Math.random() * (capped / 2);
+    this.reconnectTimer = window.setTimeout(() => {
+      this.reconnectTimer = null;
+      void this.open();
+    }, delay);
+  }
+}
+
+export const loopWs = new LoopWs();
--- a/packages/dmloop/src/panel/RunDetailModal.tsx
+++ b/packages/dmloop/src/panel/RunDetailModal.tsx
@@ -12,13 +12,15 @@ import {
   ChevronRight,
   Clock,
 } from "lucide-react";
-import { useI18n } from "@octo/base";
+import { useI18n, WKApp } from "@octo/base";
 import type { TaskRun, RunMessage } from "../api/types";
 import { listRunMessages, listRuns } from "../api/runsApi";
+import { loopWs } from "../api/ws";
 import { isActiveRun } from "../ui/meta";
 import { formatDurationMs } from "../ui/time";
 
-const POLL_MS = 2000;
+// 终态后再补一次增量抓取的延迟(收尾 result/error 消息常在 status 翻转后才写入,#610)。
+const TAIL_FETCH_MS = 2000;
 
 type EventColor = "agent" | "thinking" | "tool" | "result" | "error";
 
@@ -152,55 +154,86 @@ export default function RunDetailModal({
   const missRef = useRef(0);
   const MAX_MISSES = 3;
 
-  // 打开时全量拉;运行中的 run 则每 2s 增量轮询 + 刷新状态,终态即停(无 WS,退化轮询)。
+  // 打开全量拉;运行中订阅 WS 事件驱动增量 + 状态,终态即停。
   useEffect(() => {
     if (!visible || !run) { setMessages([]); setLiveRun(null); lastSeqRef.current = 0; missRef.current = 0; setGone(false); return; }
     let cancelled = false;
-    let timer: ReturnType<typeof setTimeout> | undefined;
+    let finalTimer: ReturnType<typeof setTimeout> | undefined;
+    let unsub: Array<() => void> = [];
+    const stopListening = () => { unsub.forEach((f) => f()); unsub = []; };
     lastSeqRef.current = 0;
     missRef.current = 0;
     setGone(false);
     setLiveRun(run);
+    setMessages([]); // 切 run 时清空,后续一律 append+dedup(避免整表替换盖掉并发到达的增量)
 
-    const apply = (batch: RunMessage[], incremental: boolean) => {
-      if (batch.length) lastSeqRef.current = Math.max(lastSeqRef.current, ...batch.map((m) => m.seq));
-      if (!incremental) { setMessages(batch); return; }
+    const apply = (batch: RunMessage[]) => {
       if (!batch.length) return;
+      lastSeqRef.current = Math.max(lastSeqRef.current, ...batch.map((m) => m.seq));
       setMessages((prev) => {
         const seen = new Set(prev.map((m) => m.seq));
         return [...prev, ...batch.filter((m) => !seen.has(m.seq))];
       });
     };
 
-    const poll = () => {
-      timer = setTimeout(async () => {
-        if (cancelled) return;
-        await listRunMessages(run.id, lastSeqRef.current).then((b) => { if (!cancelled) apply(b ?? [], true); }).catch(() => {});
-        let stillActive = true;
+    // 增量抓取 + 状态检查。in-flight 守卫:突发事件不并发重叠,进行中再来的合并成一次尾随重跑。
+    // 事件为 workspace 级未按 run 过滤,抓本 run 自己的消息即可。
+    let ticking = false;
+    let pending = false;
+    const tick = async () => {
+      if (cancelled) return;
+      if (ticking) { pending = true; return; }
+      ticking = true;
+      try {
+        await listRunMessages(run.id, lastSeqRef.current).then((b) => { if (!cancelled) apply(b ?? []); }).catch(() => {});
         try {
           const fresh = (await listRuns(run.issue_id)).find((r) => r.id === run.id);
           if (cancelled) return;
-          if (fresh) { setLiveRun(fresh); stillActive = isActiveRun(fresh.status); missRef.current = 0; }
-          else { missRef.current += 1; stillActive = missRef.current < MAX_MISSES; if (!stillActive) setGone(true); } // 连续 miss 达阈值:判定 run 已消失,停轮询并提示
-        } catch { /* ensureDirectory 等抛错:保持轮询 */ }
-        if (!cancelled && stillActive) poll();
-        // 终止前再做一次增量抓取:run 完成时的最后一批消息(result/error)常在上面抓取返回后、
-        // status 翻转前才写入,不补一次会永久丢失最关键的收尾消息(#610 评审)。
-        else if (!cancelled) await listRunMessages(run.id, lastSeqRef.current).then((b) => { if (!cancelled) apply(b ?? [], true); }).catch(() => {});
-      }, POLL_MS);
+          if (fresh) {
+            setLiveRun(fresh);
+            missRef.current = 0;
+            if (!isActiveRun(fresh.status)) {
+              // 终态:停订阅,再延迟补一次增量——收尾 result/error 消息常在 status 翻转后才写入(#610)。
+              stopListening();
+              finalTimer = setTimeout(() => {
+                listRunMessages(run.id, lastSeqRef.current).then((b) => { if (!cancelled) apply(b ?? []); }).catch(() => {});
+              }, TAIL_FETCH_MS);
+            }
+          } else {
+            missRef.current += 1;
+            if (missRef.current >= MAX_MISSES) { setGone(true); stopListening(); } // 连续 miss:run 已消失,停并提示
+          }
+        } catch { /* ensureDirectory 等抛错:忽略,下个事件再试 */ }
+      } finally {
+        ticking = false;
+        if (pending && !cancelled) { pending = false; void tick(); }
+      }
     };
+
+    // 先订阅再 fetch:监听器必须在初始 fetch 之前挂——否则 fetch 那一 round-trip 内到达的终态事件
+    // 会因无监听器丢失(终态不复发 → tick 永不触发 → 弹窗永远卡在"运行中")。仅运行中 run 订阅。
+    if (isActiveRun(run.status)) {
+      const trigger = () => { void tick(); };
+      WKApp.mittBus.on("wk:loop-issues-refresh", trigger); // 域级:含 issue:deleted / 终态 / 重连
+      unsub = [
+        () => WKApp.mittBus.off("wk:loop-issues-refresh", trigger),
+        loopWs.on("activity:created", trigger), // transcript 增量
+        loopWs.on("task:progress", trigger),
+      ];
+    }
 
     setLoading(true);
     listRunMessages(run.id)
-      .then((m) => { if (!cancelled) apply(m ?? [], false); })
+      .then((m) => { if (!cancelled) apply(m ?? []); })
       .catch(() => { if (!cancelled) setMessages([]); })
       .finally(() => {
         if (cancelled) return;
         setLoading(false);
-        if (isActiveRun(run.status)) poll();
+        // reconcile:补齐 fetch 窗口内到达的事件,并用 listRuns 查最新状态(mount 时 run.status 可能已陈旧)。
+        if (isActiveRun(run.status)) void tick();
       });
 
-    return () => { cancelled = true; if (timer) clearTimeout(timer); };
+    return () => { cancelled = true; stopListening(); if (finalTimer) clearTimeout(finalTimer); };
   }, [visible, run]);
 
   const shown = liveRun ?? run;
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -147,7 +147,17 @@ export default defineConfig(({ mode }) => {
             ? (path: string) => path.replace(/^\/matter/, "")
             : undefined,
         },
-        // Loop (fleet) service API — 真实 multica-server/fleet 联调。
+        // dmloop 实时 WS:直连后端 /ws(loop proxy 不处理 upgrade)。须在 /fleet/api 规则之前。
+        // 生产需网关把 /fleet/ws 的 WebSocket upgrade 转发到后端 /ws。
+        "/fleet/ws": {
+          target: env.VITE_FLEET_WS_URL || "ws://127.0.0.1:8080",
+          // changeOrigin:false 保留原 Host 与 Origin 同源,过后端 checkOrigin;置 true 会 403。
+          changeOrigin: false,
+          secure: false,
+          ws: true,
+          rewrite: (path: string) => path.replace(/^\/fleet\/ws/, "/ws"),
+        },
+        // Loop (fleet) service API — 真实后端 fleet 服务联调。
         // dev fleet 在 127.0.0.1:8091 直接提供完整 /fleet/api/v1/... 路径（不 strip）。
         // 必须放在下面通用 /fleet/api/ 规则之前（vite first-match）。
         "/fleet/api/v1": {
--- a/packages/dmloop/src/api/__tests__/ws.test.ts
+++ b/packages/dmloop/src/api/__tests__/ws.test.ts
@@ -1,0 +1,115 @@
+// @vitest-environment jsdom
+import { afterEach, beforeEach, describe, expect, it, vi, type Mock } from "vitest";
+import { WKApp } from "@octo/base";
+import { loopWs } from "../ws";
+
+// loopWs 依赖:WKApp.mittBus(域刷新信号落点)、issueLoopCliToken(握手 token)、全局 WebSocket。
+// 这里全部替身,只验 loopWs 自身的事件→刷新映射、去抖合并、on() 分发,不打真实网络。
+vi.mock("@octo/base", () => ({
+  WKApp: { mittBus: { emit: vi.fn(), on: vi.fn(), off: vi.fn() } },
+}));
+vi.mock("../authApi", () => ({
+  issueLoopCliToken: vi.fn(() => Promise.resolve({ token: "tok" })),
+}));
+
+// 可被测试驱动的假 WebSocket:记录 send、暴露 open()/message() 触发回调。
+class FakeWebSocket {
+  static last: FakeWebSocket | null = null;
+  onopen: (() => void) | null = null;
+  onmessage: ((e: { data: string }) => void) | null = null;
+  onclose: (() => void) | null = null;
+  onerror: (() => void) | null = null;
+  sent: string[] = [];
+  constructor(public url: string) { FakeWebSocket.last = this; }
+  send(d: string) { this.sent.push(d); }
+  close() { /* no-op */ }
+  open() { this.onopen?.(); }
+  message(obj: unknown) { this.onmessage?.({ data: JSON.stringify(obj) }); }
+}
+
+const emitMock = () => (WKApp.mittBus.emit as unknown as Mock);
+
+// 起连并推进到「已鉴权、可收事件」:start → 等 token 微任务 → onopen(发 auth) → auth_ack。
+// 末尾推进过去抖窗口,把 auth_ack 自己排的那次 refresh 冲掉,让调用方从干净态开始断言。
+async function connectAndAuth(slug = "ws-test") {
+  loopWs.start(slug);
+  await vi.advanceTimersByTimeAsync(0); // flush issueLoopCliToken() 的微任务 → 创建 socket
+  const sock = FakeWebSocket.last!;
+  sock.open();
+  sock.message({ type: "auth_ack" });
+  await vi.advanceTimersByTimeAsync(250); // 冲掉 auth_ack 排的 refresh
+  return sock;
+}
+
+describe("loopWs", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.stubGlobal("WebSocket", FakeWebSocket as unknown as typeof WebSocket);
+    FakeWebSocket.last = null;
+    emitMock().mockClear();
+  });
+  afterEach(() => {
+    loopWs.stop();
+    vi.useRealTimers();
+    vi.unstubAllGlobals();
+  });
+
+  it("握手 token 作为首帧 auth 发送", async () => {
+    const sock = await connectAndAuth();
+    expect(sock.sent).toHaveLength(1);
+    expect(JSON.parse(sock.sent[0])).toEqual({ type: "auth", payload: { token: "tok" } });
+  });
+
+  it("一串 REFRESH 事件在去抖窗口内合并成一次 wk:loop-issues-refresh", async () => {
+    const sock = await connectAndAuth();
+    emitMock().mockClear(); // 忽略 auth_ack 自己排的那次
+    sock.message({ type: "issue:created" });
+    sock.message({ type: "task:dispatch" });
+    sock.message({ type: "task:running" });
+    expect(emitMock()).not.toHaveBeenCalled(); // 去抖窗口内不立即发
+    await vi.advanceTimersByTimeAsync(250);
+    const refreshCalls = emitMock().mock.calls.filter((c) => c[0] === "wk:loop-issues-refresh");
+    expect(refreshCalls).toHaveLength(1); // 三帧合并成一次
+  });
+
+  it("非看板事件(task:progress/activity:created)不触发看板刷新", async () => {
+    const sock = await connectAndAuth();
+    emitMock().mockClear();
+    sock.message({ type: "task:progress" });
+    sock.message({ type: "activity:created" });
+    sock.message({ type: "agent:status" }); // presence:也不刷看板
+    await vi.advanceTimersByTimeAsync(300);
+    expect(emitMock().mock.calls.filter((c) => c[0] === "wk:loop-issues-refresh")).toHaveLength(0);
+  });
+
+  it("auth_ack 触发一次刷新(重连补齐)", async () => {
+    loopWs.start("ws-test");
+    await vi.advanceTimersByTimeAsync(0);
+    const sock = FakeWebSocket.last!;
+    sock.open();
+    sock.message({ type: "auth_ack" });
+    await vi.advanceTimersByTimeAsync(250);
+    expect(emitMock().mock.calls.filter((c) => c[0] === "wk:loop-issues-refresh")).toHaveLength(1);
+  });
+
+  it("on(type) 收到原始 payload;退订后不再回调", async () => {
+    const sock = await connectAndAuth();
+    const h = vi.fn();
+    const off = loopWs.on("task:progress", h);
+    sock.message({ type: "task:progress", payload: { seq: 1 } });
+    expect(h).toHaveBeenCalledWith({ seq: 1 });
+    off();
+    sock.message({ type: "task:progress", payload: { seq: 2 } });
+    expect(h).toHaveBeenCalledTimes(1);
+  });
+
+  it("坏帧(非 JSON / 无 type)被安全丢弃,不触发刷新", async () => {
+    const sock = await connectAndAuth();
+    emitMock().mockClear();
+    sock.onmessage?.({ data: "not-json" });
+    sock.message({ type: 123 }); // type 非 string
+    sock.message({ payload: {} }); // 无 type
+    await vi.advanceTimersByTimeAsync(300);
+    expect(emitMock().mock.calls.filter((c) => c[0] === "wk:loop-issues-refresh")).toHaveLength(0);
+  });
+});
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -100,9 +100,10 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
     listProjectOptions().then(setProjects).catch(() => {});
   }, []);
 
-  const reload = useCallback(() => {
+  // silent=true:后台重取不切 loading(WS 刷新 / mutation 回读用,避免闪屏);首次/改筛选/切视图走非静默。
+  const reload = useCallback((silent = false) => {
     const my = ++seq.current;
-    setLoading(true);
+    if (!silent) setLoading(true);
     // 时间范围:三参数须同时给且 start<end。onChange 已把 dateRange 归一为 undefined|[起,止]。
     // 止端 +1 日历日 → 半开区间,既含止日当天、又保证 start<end;setDate 处理 DST。
     const dr = f.dateRange;
@@ -189,21 +190,11 @@ export default function IssuePage({ defaultScope, defaultView }: IssuePageProps 
       .then((tasks) => { if (my === runSeq.current) setRunning(new Set(tasks.filter((tk) => isActiveRun(tk.status) && tk.issue_id).map((tk) => tk.issue_id))); })
       .catch(() => {});
   }, []);
-  // 挂载取一次 + 每 15s 轮询:无 WS 推送,agent 起停靠轮询让 running chip 最终收敛(而非只在本地 mutation 后)。
-  useEffect(() => {
-    refreshRunning();
-    const timer = setInterval(refreshRunning, 15000);
-    return () => clearInterval(timer);
-  }, [refreshRunning]);
+  useEffect(() => { refreshRunning(); }, [refreshRunning]);
 
-  // 变更后刷新:既重取列表,又刷新运行中快照(指派/状态变更可能起/停 agent run)。
-  const onMutated = useCallback(() => { reload(); refreshRunning(); }, [reload, refreshRunning]);
+  const onMutated = useCallback(() => { reload(true); refreshRunning(); }, [reload, refreshRunning]);
 
-  // 派单(quick-create)是异步的:agent 稍后才建 issue(dmloop 无 WS 推送,见记忆 dmloop-no-realtime-defer-ws)。
-  // NewLoopPage 派单成功发 `wk:loop-issues-dispatched`,常驻的 LoopPage 据此有界补发 `wk:loop-issues-refresh`,
-  // 看板收到即重取——一套机制覆盖"看板内新建"与"侧栏新建"两个入口(定时器归 LoopPage,见那里)。
-  // 走 ref 读最新 onMutated:延迟刷新须用当前筛选/视图,否则会用陈旧入参覆盖(reload 的 seq 守卫只防乱序)。
-  // ponytail: 真正的修法是 WS 实时推送(已记后期做),此为 stopgap;派单低频,几次补刷可接受。
+  // WS 事件 → 静默重取。走 ref 读最新 onMutated,用当前筛选/视图重取(避免陈旧闭包)。
   const onMutatedRef = useRef(onMutated);
   useEffect(() => { onMutatedRef.current = onMutated; }, [onMutated]);
   useEffect(() => {
--- a/packages/dmloop/src/pages/LoopPage.tsx
+++ b/packages/dmloop/src/pages/LoopPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef, useState } from "react";
+import React, { useEffect, useState } from "react";
 import { Typography, Dropdown, Avatar, Modal, Toast, Button } from "@douyinfe/semi-ui";
 import {
   ClipboardList, Briefcase, Bot, Users, Settings,
@@ -11,6 +11,7 @@ import { listWorkspaces, createWorkspace } from "../api/workspaceApi";
 import { setWorkspaceContext, currentWorkspaceId } from "../api/http";
 import { invalidateDirectory } from "../api/directory";
 import { invalidateRuntimeMap } from "../api/agentApi";
+import { loopWs } from "../api/ws";
 import IssuePage from "./IssuePage";
 import NewLoopPage from "./NewLoopPage";
 import ProjectPage from "./ProjectPage";
@@ -24,9 +25,6 @@ import "../ui/loopControls.css";
 const { Title, Text } = Typography;
 
 type TabKey = "myloop" | "issue" | "project" | "automation" | "agent" | "squad" | "settings";
-
-// 派单后看板补刷的退避时刻(ms):agent 异步建单的落库延迟不可观测,手调的退避窗口(非可推导状态)。
-const SETTLE_DELAYS_MS = [2000, 5000, 9000, 14000];
 
 
 // 顶部独立入口：我的回路（复用 Issue 视图的「与我相关」分组）。
@@ -104,10 +102,12 @@ export default function LoopPage() {
       setWsId(ws.id);
       invalidateDirectory();
       invalidateRuntimeMap();
+      loopWs.start(ws.slug);
       WKApp.routeRight.replaceToRoot(renderTab(tab, ws));
     } else {
       setWorkspaceContext("", "");
       setWsId("");
+      loopWs.stop();
       showEmptyGuide();
     }
     setWorkspaces(list);
@@ -149,27 +149,15 @@ export default function LoopPage() {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [workspaces, wsId, loaded]);
 
-  // 派单后看板补刷:quick-create 异步(agent 稍后建 issue,dmloop 无 WS,见记忆 dmloop-no-realtime-defer-ws)。
-  // NewLoopPage 派单成功发 `wk:loop-issues-dispatched`;由常驻(不随 tab/建单入口重挂)的 LoopPage 持有定时器,
-  // 有界补发 `wk:loop-issues-refresh`,当前挂载的看板(IssuePage)订阅后重取——一套机制统一覆盖
-  // 「看板内新建」与「侧栏新建」两个入口(此前 settle 只挂在看板实例上,漏了侧栏 openTab 重挂的路径)。
-  const settleTimersRef = useRef<number[]>([]);
-  useEffect(() => {
-    const onDispatched = () => {
-      settleTimersRef.current.forEach(clearTimeout);
-      settleTimersRef.current = SETTLE_DELAYS_MS.map((d) =>
-        window.setTimeout(() => WKApp.mittBus.emit("wk:loop-issues-refresh"), d),
-      );
-    };
-    WKApp.mittBus.on("wk:loop-issues-dispatched", onDispatched);
-    return () => { WKApp.mittBus.off("wk:loop-issues-dispatched", onDispatched); settleTimersRef.current.forEach(clearTimeout); };
-  }, []);
+  // LoopPage 常驻:WS 连接由 applyWorkspace/switchWorkspace 建立,卸载时断开。
+  useEffect(() => () => loopWs.stop(), []);
 
   const switchWorkspace = (w: Workspace) => {
     setWorkspaceContext(w.slug, w.id);
     setWsId(w.id);
     invalidateDirectory();
     invalidateRuntimeMap();
+    loopWs.start(w.slug);
     WKApp.routeRight.replaceToRoot(renderTab(tab, w));
   };
 
--- a/packages/dmloop/src/pages/NewLoopPage.tsx
+++ b/packages/dmloop/src/pages/NewLoopPage.tsx
@@ -85,10 +85,7 @@ export default function NewLoopPage({ onCreated }: NewLoopPageProps) {
       });
       // 单一 toast 归属此处(异步派单的准确措辞);duration 显式设 3s 保证自动消失。
       Toast.success({ content: t("loop.newLoop.dispatched"), duration: 3 });
-      // 派单是异步的(agent 稍后建 issue),通知常驻 LoopPage 有界补刷看板,使新回路自动出现(无需手动 reload)。
-      // 见 LoopPage 的 wk:loop-issues-dispatched 处理;覆盖看板内/侧栏两个入口。
-      WKApp.mittBus.emit("wk:loop-issues-dispatched");
-      // 成功后由调用方负责导航(切回回路看板)——此处不再自行 back(),避免叠加把新根 pop 掉。
+      // 看板刷新由 WS 驱动,此处只导航回看板。
       onCreated?.();
     } catch (e) {
       // quick-create 结构化 422:LoopApiError.code 映射成友好提示(即时失败反馈)。
--- a/packages/dmloop/src/panel/SquadDetailPage.tsx
+++ b/packages/dmloop/src/panel/SquadDetailPage.tsx
@@ -7,6 +7,7 @@ import {
   getSquad, updateSquad, deleteSquad, addSquadMember, removeSquadMember, updateSquadMemberRole, getSquadMemberStatus,
 } from "../api/squadApi";
 import { listAssigneeCandidates } from "../api/issueApi";
+import { loopWs } from "../api/ws";
 import { confirmDelete } from "../ui/confirmDelete";
 import { formatRelativeTime } from "../ui/time";
 import AgentDetailPage from "./AgentDetailPage";
@@ -63,10 +64,10 @@ export default function SquadDetailPage({ squadId, onChanged }: { squadId: strin
   }, [squadId]);
   useEffect(load, [load]);
   useEffect(() => { listAssigneeCandidates().then((cs) => setCands(cs.filter((c) => c.type !== "squad"))).catch(() => setCands([])); }, []);
-  // 成员状态非实时推送：面板打开期间每 30s 轮询一次，作为 staleTime 式兜底。
+  // 成员在线状态由 WS agent:status 驱动;断连期间的变化在重连后下个心跳自愈。
   useEffect(() => {
-    const id = window.setInterval(() => { getSquadMemberStatus(squadId).then(setStatusList).catch(() => undefined); }, 30000);
-    return () => window.clearInterval(id);
+    const refetch = () => { getSquadMemberStatus(squadId).then(setStatusList).catch(() => undefined); };
+    return loopWs.on("agent:status", refetch);
   }, [squadId]);
 
   const statusById = useMemo(() => {
--- a/packages/dmworkbase/src/App.tsx
+++ b/packages/dmworkbase/src/App.tsx
@@ -71,12 +71,7 @@ export type MittEvents = {
    * 不会自动 remount, 接收方需要主动 reload。
    */
   'wk:nav-menu-activated': { menuId: string };
-  /**
-   * dmloop 派单(quick-create)后的看板补刷协议。派单是异步的(agent 稍后建 issue,dmloop 暂无 WS 推送):
-   * NewLoopPage 派单成功发 `wk:loop-issues-dispatched`;常驻的 LoopPage 据此有界补发 `wk:loop-issues-refresh`,
-   * 当前挂载的看板(IssuePage)订阅后重取,使新回路自动出现——统一覆盖看板内/侧栏两个建单入口。
-   */
-  'wk:loop-issues-dispatched': void;
+  /** dmloop 看板实时刷新信号:WS 客户端(dmloop/api/ws.ts)收到 issue/task 事件时发出,看板订阅后静默重取。 */
   'wk:loop-issues-refresh': void;
   /**
    * 打开「密钥 / Secrets」管理面板（YUJ-3539）。由聊天反向跳转（bot 消息里的


MrAladdin:
Review transparency: the Codex side of the dual-model adversarial pass could **not** complete — 3 attempts all stream-dropped mid-analysis (network instability this session; the trailing "approve/no findings" is the failed-run fallback, not a real verdict). Correctness coverage therefore rests on: the Claude skeptical code-reviewer (which found and drove the RunDetailModal freeze fix), the 4-angle /simplify pass, the added unit tests (dmloop 23/23), and the real-env dev e2e. Happy to re-run Codex once network is stable if a maintainer wants it.

OctoBoooot (dismissed):
## Review: fix(loop): drive board & transcript refresh via realtime WS, remove blind polling (#714)

**Verdict:** Approve. A high-risk change (new WS client, token auth, replacing all dmloop polling) executed to a high standard — race-safe lifecycle, no leaks/storms, auth not leaked, and the #610 behaviors preserved. No blockers, no majors.

Base `main`. Risk tier: high. `mergeStateStatus=CLEAN`, CI green. This is the realtime replacement the #683/#610 settle-poll/drain stopgaps deferred.

### WS client (`ws.ts`) — lifecycle is correct (the main risk surface, byte-verified)
- **No socket/timer/handler leaks, no reconnect storm.** Singleton with `start(slug)` no-op on same slug + stop-before-reconnect; `stop()` clears both timers and **nulls `onclose`/`onerror` before `close()`** (prevents reconnect firing during intentional teardown). `scheduleReconnect` bails on `stopped` and guards against double-timers (single 3s timer).
- **Race-safe async setup.** `open()` re-checks `stopped || slug !== this.slug` after the async token fetch (`:110`); `onopen`/`onclose` both guard `this.ws === ws` (`:116/:130`) so a superseded socket can't send auth or trigger reconnect.
- **Auth safe.** Token from the existing `issueLoopCliToken()`, sent in the first frame (`:116`), **never logged**, stays in memory. Token-fetch failure → reconnect (no hard fail).
- **Crash-safe dispatch.** `onmessage` JSON.parse in try/catch, type-guards, per-handler try/catch (`:120-127`). `auth_ack` re-emits refresh so a reconnect backfills anything missed while down.
- **Sensible event routing.** `REFRESH_EVENTS` (debounced 250ms → `wk:loop-issues-refresh`) deliberately excludes high-frequency `task:progress`/`activity:created` (transcript) and `agent:status` (presence) — those go via `on(type)` — so transcript traffic can't storm board refetches.

### Consumers clean up + #610 preserved
- **LoopPage**: `loopWs.start(slug)` on workspace resolve, `stop()` on switch, `useEffect(() => () => loopWs.stop(), [])` on unmount; old settle-timer machinery removed.
- **RunDetailModal**: subscribes to `wk:loop-issues-refresh` + `loopWs.on("activity:created"/"task:progress")`, all collected in `unsub[]` and torn down via `stopListening()` in cleanup **and** on gone-detection (no handler leak). The #610 gone-guard (`MAX_MISSES`/`setGone`) and terminal final-fetch drain are retained under the WS-driven tick, with an in-flight guard so bursty transcript events don't overlap.

### Praise
`ws.test.ts` (+115) covers the behaviors that matter: auth-frame-first, debounce coalescing, non-board events don't refresh, auth_ack re-refresh (reconnect backfill), `on()`/unsubscribe (no callback after unsub), and bad-frame safe-discard. Real lifecycle assertions. The vite `/fleet/ws` proxy is careful too — `ws:true`, `changeOrigin:false` to keep same-origin for the backend's `checkOrigin`, rewrite `/fleet/ws`→`/ws`, ordered before the REST rule, with the origin reasoning documented.

### Note (non-blocking)
No explicit reconnect-*timing* test (the auth_ack-backfill and stop-no-reconnect paths are covered, which are the correctness-critical ones) — a small addition if you want the 3s reconnect pinned. Prod requires the gateway to pass `/ws` (the PR notes this as a deploy follow-up) — worth confirming that's in place before the feature is relied on, but it degrades to no-realtime (not breakage) if absent.

### Net
The right fix for the polling problem — realtime, race-safe, leak-free, auth-safe, with the #610 gone/drain behaviors carried forward and real tests. Approving.

Jerry-Xin (dismissed):
Byte-verified independently at head `41510aed` (merge-base == base, true incremental diff; new `ws.ts` +146 + `ws.test.ts` +115 + consumer wiring). **APPROVE** — every high-risk WS surface is handled textbook-correctly. Verified each sink myself.

## ✅ WS teardown — no leak, no reconnect storm
`stop()` (ws.ts:77-86) clears both `reconnectTimer` + `refreshTimer`, then **nulls `onclose`/`onerror` BEFORE `close()`** (:83-85) and nulls `this.ws` — the intentional-teardown-doesn't-trigger-reconnect guard. `start()` (:69-73) no-ops on same slug (`!this.stopped && this.slug === slug`) and `stop()`s before reconnecting on slug change.

## ✅ Reconnect — single-timer, stopped-guarded
`scheduleReconnect` (:137-140): `if (this.stopped || this.reconnectTimer !== null) return` — bails when stopped, single-timer prevents overlap; fixed 3s backoff. `onclose` (:129-132) guards `if (this.ws !== ws) return` so a superseded/stopped socket can't schedule a reconnect. No path where onclose→reconnect fires after intentional stop.

## ✅ Stale-socket race across async token fetch
`open()` (:100-116): after `await issueLoopCliToken()`, re-checks **both** `this.stopped` and `slug !== this.slug` (:110) — a stop or workspace-switch during the token fetch invalidates the connection. `onopen` guards `this.ws === ws` (:116, superseded socket won't send auth); `onclose` guards `this.ws !== ws` (:130).

## ✅ Token — first-frame body, never in URL, never logged
Token from `issueLoopCliToken()` → `/cli-token`, sent as first frame `ws.send(JSON.stringify({ type: "auth", payload: { token } }))` (:116). `wsUrl()` query carries only `workspace_slug` + `client_platform` — **no token in URL** (no proxy/referrer leak). No `console.*` anywhere in ws.ts. This is the safe case.

## ✅ onmessage crash-safe
JSON.parse in try/catch → discard on throw; type-guard `typeof msg.type !== "string"`; each handler dispatch in its own try/catch so one throwing handler can't break the loop.

## ✅ #610 behaviors preserved under WS (RunDetailModal.tsx)
Both preserved: (a) **gone-guard** `if (missRef.current >= MAX_MISSES) { setGone(true); stopListening() }` (MAX_MISSES=3); (b) **terminal tail-drain** on `!isActiveRun` → `stopListening()` + `setTimeout(listRunMessages, TAIL_FETCH_MS=2000)` (comment cites #610). Plus a NEW **in-flight stampede guard** (`ticking`/`pending` → coalesce bursty task:progress into one trailing refetch).

## ✅ Consumer cleanup — all four unsubscribe on unmount
LoopPage `useEffect(() => () => loopWs.stop(), [])`; RunDetailModal cleanup `unsub.forEach(f => f()); clearTimeout(finalTimer)`; IssuePage `mittBus.off(...)`; SquadDetailPage returns `loopWs.on(...)` unsub directly. No handler leak.

## ✅ vite WS proxy — correct ordering + same-origin
`/fleet/ws` rule placed **before** `/fleet/api/v1` / `/fleet/api/` (vite first-match, so the upgrade rule wins), `ws: true`, `changeOrigin: false` (preserves Host → backend realtime `checkOrigin` same-origin branch passes; comment notes `true` would 403), `rewrite: /^\/fleet\/ws/ → /ws`.

## ✅ No new issues
No XSS (payloads render as React text, no `dangerouslySetInnerHTML`); no unbounded buffer (single-shot debounce); origin preserved; workspace-scoped by `workspace_slug` (start/stop per workspace, no cross-talk); absent `/ws` degrades gracefully (token/open fail → 3s reconnect retry, no crash, matches poll-free behavior).

## 🟡 Non-blocking (test gap)
`ws.test.ts` is substantive (auth-frame-first, debounce coalescing, auth_ack re-refresh/reconnect backfill, on/unsubscribe, bad-frame discard) but lacks an explicit `stop()→no-reconnect` / `onclose→scheduleReconnect` timing test. The stop/reconnect logic is simple and verified clean; a follow-up test would harden it. Optional.

Net: a clean, leak-free, race-safe realtime WS client that preserves the #610 gone-guard + tail-drain and adds stampede coalescing, with correct proxy same-origin handling and no token/XSS exposure. **APPROVE.**

OctoBoooot (approved):
## Review: #714 @ `03fb995b` — re-anchoring APPROVE (squash-rebase, content-identical)

**Verdict:** Approve — re-anchoring my prior APPROVE (`41510aed`) to the current head. The new head is a squash-rebase onto current main with **condensed comments only**; the WS client, consumer wiring, and #610 preservation are byte-identical to what I reviewed.

### Verified at the new head
- **`ws.ts`** — line-for-line the same WS client I approved: same `start`/`stop` (nulls `onclose`/`onerror` before `close()`), same async race guards (`this.ws === ws`, `stopped` re-check after token fetch), same `open`/auth-first-frame/`onmessage`/`scheduleReconnect`. Only the Chinese comments were condensed. No logic change.
- **`RunDetailModal.tsx`** — the only diff vs my head is comment rewording (lines 157/180/224); **zero code lines changed**. The #610 gone-guard (`MAX_MISSES`/`setGone`), terminal final-fetch drain, and the in-flight guard are all intact.
- **`ws.test.ts`** — byte-identical to what I reviewed.
- CI green (`Build`/`code-review` SUCCESS), `mergeStateStatus=CLEAN`.

### Net
Everything I byte-verified on my original review (race-safe leak-free WS lifecycle, token auth not leaked, sensible event routing, silent-refetch de-flicker, #610 behaviors preserved, real tests, same-origin proxy) holds unchanged. APPROVE carries to `03fb995b`. Steve's one 🟡 (add a `stop()`-no-reconnect timing test) remains a fair non-blocking follow-up. Ready to merge.

yujiawei (changes requested):
# Code Review — PR #714 (octo-web)

Reviewed at head `41510aed77f38f4fb522fe874594715b5e9d20fc` against `origin/main` (merge-base `ede8a038`). This replaces the dmloop board/transcript blind-polling with a realtime WebSocket client (`packages/dmloop/src/api/ws.ts`), removes the `wk:loop-issues-dispatched` stopgap, and migrates `IssuePage`, `RunDetailModal`, and `SquadDetailPage` off timers. Overall the design is clean and well-documented, the token-in-first-frame auth is the right choice, and the new unit tests cover the core event→refresh mapping, debounce, dispatch, and bad-frame handling. One blocking correctness regression, plus several robustness items.

## Scope

The diff matches its stated intent — WS client + tests, removal of the dispatch stopgap, timer removal in the three consumers, and one Vite proxy rule. No out-of-scope additions and no dangling references to the removed `wk:loop-issues-dispatched` / `SETTLE_DELAYS_MS`. `SquadDetailPage` retains its initial status fetch (`load()`), so removing the 30s timer does not drop the first render.

## Blocking

### P1 — `RunDetailModal.tsx:215-233`: terminal event can be missed permanently (regression of a previously self-healing path)

WS listeners are installed inside the `.finally()` of the initial `listRunMessages(run.id)` call. Any event that fires between the moment the backend forms that HTTP response and the moment the listeners attach (one request round-trip) is not delivered. For an active run this is mostly self-healing — the next `activity:created` / `task:progress` triggers `tick()`, which does an incremental `listRunMessages(since=lastSeq)` and catches up. The unrecoverable case is a **terminal** event (`task:completed` / `task:failed`) landing in that gap: it is the last event, so nothing later triggers `tick()`, `isActiveRun` never flips, the `TAIL_FETCH_MS` catch-up is never scheduled, and the modal stays stuck on "running" with the final result/error message never shown. The old 2s poll always recovered within one interval; removing polling without a post-subscribe catch-up removes that safety net.

Suggested fix: subscribe **before** the initial fetch, or run a single `tick()` immediately after installing the listeners. A trailing `tick()` closes the gap cleanly because it fetches incrementally from `lastSeqRef.current` and re-checks run status.

## Non-blocking (P2 / suggestions)

- **`ws.ts:77-88` / `ws.ts:118`: stale-socket frame during stop/switch.** `stop()` nulls `onclose`/`onerror` but not `onmessage`/`onopen`, and `onmessage` does not guard `this.ws === ws`. Because `close()` is asynchronous, a frame already queued when switching workspaces can still `emit("wk:loop-issues-refresh")` or invoke handlers against the new workspace context. Null all handlers in `stop()` and add a `if (this.ws !== ws) return;` guard at the top of `onmessage`.

- **`ws.ts:99-135`: reconnect robustness.** (a) `new WebSocket(wsUrl(slug))` can throw synchronously (malformed URL, mixed-content/CSP); it is called before `this.ws = ws` and outside any try/catch, so a throw rejects `open()` (swallowed by `void`) with no `scheduleReconnect()`, leaving the singleton permanently dead. Wrap construction in try/catch → `scheduleReconnect()`. (b) On session expiry, `issueLoopCliToken()` rejecting drives a fixed 3s reconnect forever, and the fixed interval (no jitter/backoff) means all clients retry in lockstep after a backend restart. Not new behavior versus the old polling, but worth capping retries / adding jittered backoff, and ideally not reconnecting on a hard `401`.

- **`SquadDetailPage.tsx:69-72`: no reconnect catch-up for presence.** This page now listens only to `agent:status` and does not subscribe to the `wk:loop-issues-refresh` reconnect signal that other consumers use. If a member's presence changes during a disconnect and the backend does not re-emit `agent:status` on reconnect, the list stays stale until the next heartbeat. Consider also refetching on the reconnect signal.

- **`ws.ts:32-38` + `vite.config.ts:155-164`: `VITE_FLEET_WS_URL` overloaded.** The same env var is the Vite dev-proxy target *and* a browser-side direct-connect override. If set for the proxy, the browser branch also reads it and bypasses `/fleet/ws`, so the deliberate `changeOrigin:false` same-origin workaround no longer applies. `wsUrl` also unconditionally appends `/ws`, so a value already ending in `/ws` becomes `/ws/ws`. Consider separate vars (server vs. client) and normalizing a trailing `/ws`.

- **`RunDetailModal.tsx:200-202`: tail fetch bypasses the in-flight guard.** The `finalTimer` callback calls `listRunMessages` directly rather than through `tick()`, so it is not covered by the `ticking`/`pending` guard. Low risk in practice (listeners are already removed and `apply()` dedupes by `seq`), but routing it through the guard would keep a single fetch path.

## Security notes (this PR is flagged for human review)

- **Token handling is correct**: the session token is sent in the first `{type:"auth"}` frame, never in the URL or query string — good, avoids token leakage via proxy/access logs.
- **Please verify on the backend**: the client sends `workspace_slug` as a URL query param while the token authenticates the *user*. The `/ws` handler must enforce that the authenticated user is authorized for the requested workspace; otherwise a user could subscribe to another workspace's `issue:*` / `task:*` event stream. This is not verifiable from this diff.
- **Origin check**: the `changeOrigin:false` dev-proxy workaround assumes the backend `checkOrigin` has a same-origin allow branch, and the comment states production is same-origin via the gateway. Worth a human confirming the production gateway actually forwards the browser `Origin` such that `checkOrigin` passes (and is not silently disabled).

## Coverage note

Static review of the diff plus surrounding repo only; I did not run the test suite or exercise the app. I could not verify: the backend `/ws` contract (event names actually emitted, `auth_ack`/`auth_nack`/error semantics, whether `agent:status` is a guaranteed heartbeat), workspace-scoping/authz on the socket, and production gateway routing/origin handling. The four items above tagged "please verify on the backend" / "human confirming" are the cross-boundary blind spots.

## Verdict

Requesting changes on the single P1 (terminal-event miss in `RunDetailModal`) — a trivial fix (subscribe before the initial fetch, or one `tick()` after subscribing) that restores the self-healing the polling path provided. The P2 items are recommended but not blocking.

packages/dmloop/src/panel/RunDetailModal.tsx
```diff
@@ -197,10 +217,20 @@ export default function RunDetailModal({
       .finally(() => {
         if (cancelled) return;
         setLoading(false);
-        if (isActiveRun(run.status)) poll();
+        if (isActiveRun(run.status)) {
+          const trigger = () => { void tick(); };
+          // 状态/gone 校准走域级 wk:loop-issues-refresh(含 issue:deleted/终态/重连,不会漏);
+          // transcript 增量走高频 activity:created / task:progress。两者都汇入带守卫的 tick。
+          WKApp.mittBus.on("wk:loop-issues-refresh", trigger);
```
OctoBoooot: **[major]** Subscribe-after-fetch race -> lost terminal event -> modal freezes. The WS listeners are attached in the .finally() of the initial listRunMessages(run.id) fetch (:217-231), so nothing is subscribed during the fetch window. If the run completes in that gap, the terminal event (task:completed -> wk:loop-issues-refresh) is dropped; the run is then already terminal so no further event arrives -> the terminal tick never runs, setLiveRun(terminal)/stopListening()/final-drain never fire -> modal stuck on the stale active snapshot (permanent). Narrow window (one fetch round-trip) but reachable and permanent -> major. Fix: subscribe BEFORE the initial fetch, or after subscribing run one reconciling tick()/re-check run.status so no event is missed in the gap. Credit yujiawei.

packages/dmloop/src/api/ws.ts
```diff
@@ -0,0 +1,126 @@
...
+    try {
+      token = (await issueLoopCliToken()).token;
+    } catch {
+      this.scheduleReconnect();
+      return;
+    }
+    // 取 token 期间可能已 stop/切 workspace。
+    if (this.stopped || slug !== this.slug) return;
+
+    const ws = new WebSocket(wsUrl(slug));
```
OctoBoooot: **[minor]** `new WebSocket(wsUrl(slug))` is outside the try/catch. open() guards the token fetch with try/catch->scheduleReconnect (:84-87), but a synchronous throw from the WebSocket constructor (e.g. malformed VITE_FLEET_WS_URL) escapes open() unhandled and schedules no reconnect -> the singleton stalls (no socket, no timer) until the next start(). Reachable only via a misconfigured URL (not runtime data) so minor. Fix: extend the try/catch to cover the socket construction. Credit yujiawei.

OctoBoooot (changes requested):
## Correction to my #714 APPROVE → REQUEST_CHANGES — a real subscribe-after-fetch race I missed (credit yujiawei)

I approved this (and re-anchored at `03fb995b`). yujiawei's CR flags a listener-ordering race that byte-verifies as real — a permanent state-freeze on the run modal. Flipping my verdict. My APPROVE checked the WS *client* lifecycle thoroughly but missed the *consumer's* subscribe-ordering. Credit yujiawei (her repeated batch catches: #695, #578, #700, #701, #702, now #714).

### Major — RunDetailModal subscribes AFTER the initial fetch → a terminal event in the gap is lost, modal freezes
`RunDetailModal.tsx:214-231`: the WS listeners (`wk:loop-issues-refresh`, `activity:created`, `task:progress`) are attached in the **`.finally()` of the initial `listRunMessages(run.id)` fetch** — i.e. only *after* it resolves. Between the effect running (`:214`, fetch starts) and it resolving (`:217`, listeners attach), no subscription exists.

Failing sequence: open the modal on an active run → initial fetch in flight → the run completes right then and the backend emits `task:completed`/`issue:updated` (→ `wk:loop-issues-refresh`) during the fetch window → **dropped, no listener yet** → fetch resolves and subscribes, but the run is now terminal so no further `task:progress`/refresh ever arrives → the terminal `tick` never runs → `setLiveRun(fresh terminal)` never happens, `stopListening()`/final-drain never fire → the modal stays stuck on the last-fetched "active" snapshot. Permanent freeze (the missed terminal event on an already-terminal run never recurs).

Narrow window (one fetch round-trip) but reachable and its failure is permanent, so major (not blocker — requires precise timing). Fix: subscribe **before** the initial fetch, or after subscribing do one reconciling `tick()` / re-check `run.status` so no event is missed in the gap. (Subscribing first + the existing in-flight-guarded `tick` would close it cleanly.)

### Minor — `new WebSocket()` construction is outside the try/catch (uncaught → singleton stall)
`ws.ts`: `open()` wraps the token fetch in try/catch→`scheduleReconnect` (`:84-87`), but `const ws = new WebSocket(wsUrl(slug))` (`:93`) is **not** guarded. A synchronous throw there (e.g. a malformed `VITE_FLEET_WS_URL`) escapes `open()` unhandled and schedules **no** reconnect → the singleton stalls (no socket, no timer) until the next `start()`. Reachable only via a misconfigured URL (not runtime data), and the valid same-origin default won't throw — hence minor. Fix: extend the try/catch to cover the socket construction.

### Not adopting — stop() doesn't clear the `handlers` map (yujiawei's other 🟡)
This is by design, not a defect. `handlers` are *component-level* subscriptions (RunDetailModal/SquadDetailPage register via `on()` and unsubscribe on their own unmount). `stop()` intentionally manages only the socket/timers; clearing `handlers` would wrongly drop live component subscriptions across a workspace switch, and after `stop()` there's no socket delivering messages so no stale dispatch occurs (on reconnect the same handlers are correctly still registered). Leave as is.

### Still holds (unchanged from my review)
The WS client's teardown/reconnect/race-safety (null-onclose-before-close, `this.ws===ws` guards, stopped-checks, single reconnect timer), token-not-leaked auth, event routing, #610 gone/drain preservation, and consumer unmount-cleanup are all correct — those verifications stand. The gap is purely the consumer's subscribe-vs-fetch ordering.

### Net
Real state-freeze major on the run modal (subscribe-after-fetch) + a minor uncaught-construct path. Both small fixes. REQUEST_CHANGES until the subscribe-ordering is fixed. My earlier APPROVE under-checked the consumer subscribe order — owning that.

yujiawei (changes requested):
# Code Review — PR #714 (octo-web)

Re-review at head `03fb995b5171001c5e8821e43971ac9a7f77d2d2` (previous review was at `41510aed`). The force-push between those two commits is **comment-condensation only** — I diffed them and there is zero functional code change in any file. Accordingly my prior verdict stands unchanged, re-issued here against the current head.

`git diff 41510aed 03fb995b` touches only comments/JSDoc across `ws.ts`, `IssuePage.tsx`, `LoopPage.tsx`, `NewLoopPage.tsx`, `RunDetailModal.tsx`, `SquadDetailPage.tsx`, `vite.config.ts`, and the `MittEvents` doc block. No logic, control flow, or test changes. The one blocking finding below is byte-for-byte identical at this head.

## Blocking

### P1 — `RunDetailModal.tsx:213-231`: terminal event can be missed permanently

WS listeners are installed inside the `.finally()` of the initial `listRunMessages(run.id)` call. Any event that fires between the moment the backend forms that HTTP response and the moment the listeners attach (one request round-trip) is not delivered. For an active run this is mostly self-healing — the next `activity:created` / `task:progress` triggers `tick()`, which fetches incrementally from `lastSeqRef.current` and catches up. The unrecoverable case is a **terminal** event (`task:completed` / `task:failed`) landing in that gap: it is the last event, so nothing later triggers `tick()`, `isActiveRun` never flips, the `TAIL_FETCH_MS` catch-up is never scheduled, and the modal stays stuck on "running" with the final result/error message never shown. The old 2s poll always recovered within one interval; removing polling without a post-subscribe catch-up removes that safety net.

Suggested fix (small): subscribe **before** the initial fetch, or run a single `tick()` immediately after installing the listeners — a trailing `tick()` closes the gap cleanly because it fetches incrementally from `lastSeqRef.current` and re-checks run status.

## Non-blocking (P2 / suggestions)

- **`ws.ts:79-90` / `ws.ts:99`: stale-socket frame during stop/switch.** `stop()` nulls `onclose`/`onerror` but not `onmessage`/`onopen`, and `onmessage` does not guard `this.ws === ws`. Because `close()` is asynchronous, a frame already queued when switching workspaces can still `emit("wk:loop-issues-refresh")` or invoke handlers against the new workspace context. Null all handlers in `stop()` and add `if (this.ws !== ws) return;` at the top of `onmessage`.

- **`ws.ts:80-116`: reconnect robustness.** (a) `new WebSocket(wsUrl(slug))` can throw synchronously (malformed URL, mixed-content/CSP); it is called before `this.ws = ws` and outside any try/catch, so a throw rejects `open()` (swallowed by `void`) with no `scheduleReconnect()`, leaving the singleton permanently dead. Wrap construction in try/catch → `scheduleReconnect()`. (b) On session expiry, `issueLoopCliToken()` rejecting drives a fixed 3s reconnect forever with no jitter/backoff/cap, so all clients retry in lockstep after a backend restart. Consider capped, jittered backoff and not reconnecting on a hard `401`.

- **`SquadDetailPage.tsx`: no reconnect catch-up for presence.** This page listens only to `agent:status` and does not subscribe to the `wk:loop-issues-refresh` reconnect signal other consumers use. If presence changes during a disconnect and the backend does not re-emit `agent:status` on reconnect, the list stays stale. Add a one-shot `refetch` on the reconnect signal.

- **`ws.ts:22-28` + `vite.config.ts`: `VITE_FLEET_WS_URL` overloaded.** Same env var is the Vite dev-proxy target *and* a browser-side direct-connect override; if set for the proxy, the browser branch bypasses `/fleet/ws` and the deliberate `changeOrigin:false` same-origin workaround no longer applies. `wsUrl` also unconditionally appends `/ws`, so a value ending in `/ws` becomes `/ws/ws`. Separate vars + normalize trailing `/ws`.

- **`RunDetailModal.tsx` tail fetch bypasses the in-flight guard.** The `finalTimer` callback calls `listRunMessages` directly rather than through `tick()`. Low risk (listeners already removed, `apply()` dedupes by `seq`), but routing through the guard keeps a single fetch path.

## Security notes (this PR is flagged for human review)

- **Token handling is correct**: the session token is sent in the first `{type:"auth"}` frame, never in the URL/query — good, avoids leakage via proxy/access logs.
- **Please verify on the backend**: the client sends `workspace_slug` as a URL query param while the token authenticates the *user*. The `/ws` handler must enforce that the authenticated user is authorized for the requested workspace; otherwise a user could subscribe to another workspace's `issue:*` / `task:*` stream. Not verifiable from this diff.
- **Auth liveness (defense-in-depth)**: there is no `auth_ack` timeout — after sending the auth frame the client dispatches any inbound event without a client-side "authenticated" gate. The backend is expected to enforce auth before emitting, but a bounded auth timeout + gating dispatch on an authenticated flag would harden a security-sensitive socket.
- **Origin check**: the `changeOrigin:false` dev-proxy workaround assumes the backend `checkOrigin` has a same-origin allow branch and that production is same-origin via the gateway. Worth a human confirming production forwards the browser `Origin` such that `checkOrigin` passes (and is not silently disabled), and that the production gateway proxies `/ws` upgrades.

## Coverage note

Static review of the diff plus surrounding repo; tests not run, app not exercised. Could not verify: backend `/ws` contract (event names, `auth_ack`/`auth_nack`/error semantics, whether `agent:status` is a guaranteed heartbeat), socket-level workspace authz, and production gateway routing/origin handling. The backend-verify items above are the cross-boundary blind spots.

## Verdict

Requesting changes on the single P1 (terminal-event miss in `RunDetailModal`), unchanged at this head. Trivial fix (subscribe before the initial fetch, or one `tick()` after subscribing) that restores the self-healing the polling path provided. The P2 items are recommended but not blocking.

Jerry-Xin (changes requested):
**Reversing my earlier APPROVE — a real subscribe-after-load race in RunDetailModal that I missed and yujiawei correctly flagged (byte-verified at `03fb995b`).**

## 🔴 Subscribe-after-fetch + stale-status gate → dropped terminal event → permanent stuck modal (RunDetailModal.tsx:214-231)
The refresh listeners are attached **inside** `listRunMessages(run.id).finally()` — i.e. **after** the initial transcript fetch resolves — and further gated on `if (isActiveRun(run.status))` using the **stale `run` prop snapshot** (mount-time status), not the freshly-fetched status:
```js
listRunMessages(run.id).then(...).finally(() => {
  if (cancelled) return;
  setLoading(false);
  if (isActiveRun(run.status)) {          // <-- stale prop snapshot, and this runs AFTER the fetch
    WKApp.mittBus.on("wk:loop-issues-refresh", trigger);
    unsub = [ ..., loopWs.on("activity:created", trigger), loopWs.on("task:progress", trigger) ];
  }
});
```
Two concrete failures:
1. **Fetch-in-flight window**: if the run completes while the initial `listRunMessages` is still in flight, the backend emits `task:completed → wk:loop-issues-refresh` **before any listener exists → event dropped**. Because a terminal event does not recur, the modal never runs the terminal `tick()`, never `stopListening()`, and never fires the #610 tail-drain that pulls the `result`/`error` messages written *after* the status flip (your own comment at :199 documents that these land late). Result: transcript permanently missing the final messages / status frozen on the pre-completion snapshot.
2. **Stale-status gate**: the gate reads `run.status` (mount snapshot). If the run was already terminal at mount, listeners **never attach at all** — which is intended for a genuinely-finished run, but combined with (1) means there is **no reconcile-after-subscribe** path to recover a race that flips status during load.

**Fix**: subscribe **before** the initial fetch (or in a separate effect that runs first), and gate on the **fresh** status; after subscribing, do one reconcile `tick()` so any event that arrived during load is not lost. Subscribe-before-load + post-subscribe reconcile is the standard cure for the load-vs-event ordering race.

## Context
The WS **client** (`ws.ts`) itself remains clean and correct as I verified (teardown nulls handlers before close, post-await stop+slug re-check, `this.ws===ws` guards, token in first-frame body not URL, per-handler try/catch, single-timer reconnect). The bug is purely at the **consumer boundary** — RunDetailModal's subscribe-vs-load ordering — which my client-focused verification under-checked. Credit to yujiawei for catching it.

## 🟡 Non-blocking (yujiawei, confirmed minor)
`new WebSocket(wsUrl(slug))` (ws.ts) is outside the try/catch — a synchronous constructor throw (malformed `VITE_FLEET_WS_URL`) would escape `open()` without `scheduleReconnect()`, wedging the singleton. Misconfig-only, not runtime-data-reachable; wrap it for robustness.

Everything else in my prior review holds; this one race is the blocker.

MrAladdin:
Thanks @yujiawei @OctoBoooot @Jerry-Xin — all three findings fixed at `8018fa36`. Confirmed the flagged race was pre-existing (present at both `41510aed` and `03fb995b`; the force-push between them was comment-condensation only, as yujiawei verified).

**P1 — RunDetailModal subscribe-after-fetch race (terminal event lost → permanent freeze):** listeners now attach **before** the initial `listRunMessages` fetch (not in its `.finally()`), so no event in the fetch window is dropped. Plus a reconciling `tick()` runs after the fetch resolves — it re-checks status via `listRuns` (fresh, not the stale mount-time `run.status` prop) and pulls incrementally from `lastSeqRef`, closing both the fetch-gap and stale-status cases. The in-flight guard serializes any overlap.

**P2 — `new WebSocket()` outside try/catch:** construction is now inside the try/catch → `scheduleReconnect()` on a synchronous throw, so a malformed URL can't stall the singleton.

**P2 — stale-socket frame during stop/switch:** `stop()` now nulls `onmessage`/`onopen` too (in addition to `onclose`/`onerror`), and `onmessage` guards `if (this.ws !== ws) return;` — an in-flight frame from a superseded socket can no longer emit refresh or invoke handlers against the new workspace.

Verified in dev e2e: opened the run modal on an **active** run → transcript tailed live via WS (27 events accrued in-view) → run completed → modal flipped to 已完成 with the final `已收到F` message + in_review drain rendered (#610 tail-fetch intact). Board auto-refresh + no-flicker unchanged.

Jerry-Xin (dismissed):
Re-review at new head `8018fa369e61` (my prior REQUEST_CHANGES @ 03fb995b, auto-dismissed by this push). The two headline findings are **fixed**; one narrow residual remains as a non-blocking note. **APPROVE.**

## ✅ Subscribe-after-fetch timing race (the major) — FIXED
Listeners now attach **before** the initial fetch (RunDetailModal.tsx:215-224) and a **reconcile `tick()`** runs in the fetch's `.finally()` (:233), which re-fetches fresh status via `listRuns` and drains anything that arrived during the fetch round-trip. So a terminal event landing in the fetch window is no longer dropped — it's either caught by the now-attached listener or reconciled by the post-fetch tick. Correct fix, matches the prescribed direction (subscribe-before-load + reconcile).

## ✅ Uncaught `new WebSocket()` construct (the minor) — FIXED
ws.ts:96 now wraps construction: `try { ws = new WebSocket(wsUrl(slug)) } catch { this.scheduleReconnect(); return; }` — a synchronous throw (malformed `VITE_FLEET_WS_URL`) now schedules a reconnect instead of wedging the singleton. Bonus hardening in the same commit: `onmessage` now guards `this.ws !== ws` (:100, superseded-socket queued-frame guard), and `stop()` now also nulls `onmessage`/`onopen` (:64-65). Good.

## 🟡 Residual (non-blocking): subscribe + reconcile still gated on the mount-time `run.status`
Both the subscribe block (:215) and the reconcile tick (:233) are gated on `if (isActiveRun(run.status))` using the **prop captured at effect setup**. The reconcile comment says it "queries latest status (mount-time run.status may be stale)" — but that reconcile is itself skipped when the gate is false, so the gate reads exactly the possibly-stale value the comment warns about. Residual failure: if the passed `run.status` is stale-**terminal** while the run is actually active, listeners never attach **and** the reconcile never runs → the modal shows a live-active run with no updates. This is narrow and self-limiting (the run comes from the WS-driven board's click-time snapshot; a genuinely-active run showing terminal on the board requires the board itself to be stale, which the realtime refresh corrects), so it's a 🟡, not a blocker. Cleaner form: gate the reconcile on the **freshly-fetched** status (do one unconditional `tick()`/status fetch, then decide active vs terminal from `fresh.status`), so a stale-terminal prop can't suppress the very reconcile meant to correct it.

Both findings that held the gate are resolved; the residual is a non-blocking hardening note. CI green, delta is exactly the two intended files. Approving.

OctoBoooot (dismissed):
## Review: #714 @ `8018fa36` — flipping to APPROVE; all findings fixed

**Verdict:** Approve. The subscribe-after-fetch major, Steve's stale-`run.status`-gate finding, and the uncaught-`new WebSocket()` minor from my CR (`03fb995b`) are all fixed, byte-verified. This flips my REQUEST_CHANGES.

Base `main`. Risk tier: high. `Build` CI green.

### Resolved — major: subscribe now happens BEFORE the initial fetch (+ reconcile tick)
`RunDetailModal.tsx:213-223`: the WS listeners (`wk:loop-issues-refresh`, `activity:created`, `task:progress`) are now attached **before** `listRunMessages` (`:225`), with the comment naming the exact race ("监听器必须在初始 fetch 之前挂—否则 fetch round-trip 内到达的终态事件会丢失"). So a terminal event arriving during the fetch window is no longer dropped. Plus the `.finally()` now runs a reconcile `tick()` (`:233`) that queries fresh status via `listRuns` — noted as "补齐 fetch 窗口内到达的事件,并用 listRuns 查最新状态(mount 时 run.status 可能已陈旧)." Exactly the "subscribe before fetch + reconcile on fresh status" direction.

### Resolved — Steve's stale-`run.status` gate (subsumed by the same fix)
The reconcile `tick()` queries `listRuns` for the true current status, so a stale-*active* prop is corrected. The pure "stale-terminal-but-actually-active" case isn't reachable (run status is monotonic active→terminal — a terminal prop means the run is terminal), so the gate reading `run.status` there is safe. Real failure modes closed.

### Resolved — minor: `new WebSocket()` now guarded
`ws.ts:97`: `try { ws = new WebSocket(wsUrl(slug)); } catch { this.scheduleReconnect(); return; }` — the constructor is wrapped, so a synchronous throw (malformed URL) schedules a reconnect instead of stalling the singleton. Exactly the prescribed fix.

### Still holds
The WS client's clean lifecycle (null-onclose-before-close, `this.ws===ws` guards, stopped-checks, single reconnect timer, token-not-leaked auth, per-handler crash isolation), #610 gone/drain preservation, consumer cleanup, and same-origin proxy — all unchanged and correct from my original review.

### Net
Complete, correct remediation of the consumer subscribe-ordering (the one real gap in an otherwise excellent WS PR) plus the WebSocket-construct minor. The realtime replacement for the #683/#610 polling stopgaps is now sound. Approving. Credit yujiawei (subscribe-after-fetch) + Steve (stale-status gate) for the finds. (One non-blocking follow-up still open: a stop()-no-reconnect timing test.)

MrAladdin:
@yujiawei gentle nudge for a re-review 🙏 — your CHANGES_REQUESTED is recorded against `03fb995b` (the pre-fix, comment-only commit). The P1 subscribe-after-fetch race you flagged is fixed at the current head `8018fa36`: listeners now attach **before** the initial `listRunMessages` fetch + a reconciling `tick()` runs after it resolves (re-checks status via `listRuns`, not the stale mount-time prop; pulls incrementally from `lastSeqRef`), plus the two P2s (WebSocket construction inside try/catch; `stop()` nulls `onmessage`/`onopen` and `onmessage` guards `this.ws !== ws`). @OctoBoooot and @Jerry-Xin have re-reviewed `8018fa36` and approved. e2e-verified in dev (active-run modal tailed live → terminal → final `已收到F` + drain rendered). Could you re-review the current head when you have a moment? Your standing review is the last thing blocking the merge decision. Thanks!

yujiawei (dismissed):
# Code Review — PR #714 (octo-web)

Re-review at head `8018fa369e614c70c5c94fd95513652ecc46470d` against `origin/main` (merge-base `ede8a038`, true incremental diff). This replaces dmloop board/transcript blind-polling with a realtime WebSocket client (`packages/dmloop/src/api/ws.ts`), removes the dispatch-settle stopgap, and migrates `IssuePage`, `RunDetailModal`, and `SquadDetailPage` off timers. My previous review requested changes on a blocking listener-ordering bug in `RunDetailModal`; **that bug is fixed at this head**, so this review flips to Approve.

## 1. Spec compliance

**Spec: ✅**

- **No under-build.** All four polling paths are cleanly replaced: `IssuePage` 15s running-task poll → WS `task:*` via `onMutated`; `RunDetailModal` 2s poll → WS `activity:created`/`task:progress`; `SquadDetailPage` 30s presence poll → WS `agent:status`; `LoopPage` dispatch-settle backoff → `loopWs.start/stop`.
- **No over-build / no dangling refs.** The removed `wk:loop-issues-dispatched` event, `SETTLE_DELAYS_MS`, and `POLL_MS` have zero remaining references anywhere in `packages/`. Scope is tight: two new files + surgical consumer wiring + one Vite proxy rule + one `MittEvents` type update. No unrelated refactoring.
- **#610 preserved.** The post-terminal tail fetch (`TAIL_FETCH_MS`, `RunDetailModal.tsx:161`) still runs one delayed incremental fetch after status flips terminal, so the trailing `result`/`error` messages that land after the status change are not lost.
- **Blocking bug from prior head resolved.** WS listeners are now installed **before** the initial fetch (`RunDetailModal.tsx:213-223`), with a reconcile `tick()` in `.finally()` (`:229-233`). A terminal event arriving during the initial fetch round-trip is captured, so the modal no longer freezes permanently on "running".

## 2. Code Quality

**Quality: Approved**

The WS client lifecycle is handled to a high standard: `stop()` nulls **all four** handlers before `close()` and clears both timers (`ws.ts:58-72`); `onmessage`/`onclose` guard `this.ws === ws` so an in-flight frame from a superseded socket cannot leak into the new workspace context (`ws.ts:104,115`); both `issueLoopCliToken()` rejection and synchronous `new WebSocket()` throw route to `scheduleReconnect()` (`ws.ts:86-90, 97`), so the singleton can't wedge; refresh is debounced (`ws.ts:74-80`); and the `RunDetailModal` `tick()` in-flight guard (`ticking`/`pending`) coalesces event bursts. Unit tests cover auth-frame handshake, debounce coalescing, event filtering, `auth_ack` recovery, `on()` unsubscribe, and bad-frame rejection; all 6 pass locally.

### P2 (non-blocking)

- **`RunDetailModal.tsx:226-227` — full-snapshot overwrite race (PR-introduced, recoverable).** The initial mount fetch `listRunMessages(run.id)` is a full snapshot applied with `incremental=false` → `setMessages(batch)` (a full replace, `:170-171`). Because listeners now attach before this fetch, a WS event during the round-trip can fire `tick()` → an incremental `listRunMessages(run.id, since=0)`. If that WS-triggered fetch resolves **before** the initial one (out-of-order completion of two same-endpoint GETs), the incremental result is overwritten by the initial `setMessages`, and `lastSeqRef` is clamped to the initial max — so the `.finally()` reconcile `tick(since=lastSeq)` cannot re-pull the dropped rows. Effect: a transient gap in the transcript for that modal session, recoverable by reopening; run status/terminal detection (driven by `listRuns`, not messages) and the board are unaffected. The old 2s poll avoided this by deferring the first incremental fetch. Cheap fix: apply the initial fetch through the incremental/dedup path (or gate the full replace so it only lands if `lastSeqRef` hasn't advanced past its own max).
- **`ws.ts:7,122-128` — flat 3s reconnect, no jitter/backoff/cap.** Clients retry in lockstep after a backend restart; a hard `401` also loops forever. Consider capped jittered backoff and not reconnecting on auth-fatal closes. (Not a regression vs. the old polling.)
- **`ws.ts:24-29` + `vite.config.ts:150-159` — `VITE_FLEET_WS_URL` overloaded and `/ws` appended unconditionally.** The same env var is the dev-proxy target and the browser direct-connect override; and `wsUrl` always appends `/ws`, so a value already ending in `/ws` yields `/ws/ws` and a silent infinite reconnect with no user-visible cause. Consider separate server/client vars and normalizing a trailing `/ws`.
- **`SquadDetailPage.tsx:69-72` — presence has no reconnect catch-up.** It subscribes only to `agent:status`, not to the `wk:loop-issues-refresh` reconnect signal the other consumers use, so if presence changes during a disconnect and the backend does not re-emit `agent:status` on reconnect, the roster stays stale until the next heartbeat. A one-shot refetch on the reconnect signal would match the pattern used elsewhere.

## 3. Security notes (this PR is flagged for human review)

- **Token handling is correct.** The session token is sent only in the first `{type:"auth"}` frame, never in the URL or query string — this avoids token leakage via proxy/access logs. Good.
- **Human-verify — workspace authorization (cross-boundary blind spot).** The client sends `workspace_slug` as a URL query param (`ws.ts:26`) while the token authenticates the *user*. The backend `/ws` handler **must** enforce that the authenticated user is authorized for the requested workspace; otherwise a token holder could subscribe to another workspace's `issue:*` / `task:*` event stream by changing the param. Not verifiable from this frontend diff — please confirm on the backend.
- **Human-verify — origin / gateway.** `changeOrigin:false` assumes the backend `checkOrigin` has a same-origin allow branch and that production is same-origin via the gateway. Worth a human confirming the production gateway forwards the browser `Origin` such that `checkOrigin` passes (and is not silently disabled), and that it proxies `/ws` upgrades.
- **Defense-in-depth — no client-side auth gate.** After sending the auth frame the client dispatches any inbound event without waiting for `auth_ack` and without an auth timeout. The backend is expected to gate emission on auth, but a bounded `auth_ack` timeout + a client "authenticated" flag would harden a security-sensitive socket.

## 4. Coverage note

Static review of the diff plus surrounding repo; I ran the new `ws.test.ts` (6/6 pass) but did not exercise the app end-to-end. Not verifiable from this diff and left to a human/backend: the `/ws` contract (exact event names emitted, `auth_ack`/`auth_nack`/error semantics, whether `agent:status` is a guaranteed reconnect heartbeat), socket-level workspace authz, and production gateway routing/origin handling. The three "human-verify" security items above are the cross-boundary blind spots.

## Verdict

**APPROVE.** The single prior blocker (terminal-event-miss in `RunDetailModal`) is fixed correctly at this head, spec compliance is clean with no over-build or dangling refs, and the WS lifecycle is race-safe. The remaining items are P2 (transient recoverable transcript gap, reconnect backoff, env-var/`/ws` normalization, presence reconnect catch-up) — recommended but not merge-blocking. The three backend/gateway security items are flagged for the human reviewer this PR is already labeled for; none is a defect in this frontend diff.

MrAladdin:
Also addressed both non-blocking P2s from @yujiawei's approving review, at `86266ea4`:

- **P2 (RunDetailModal full-snapshot overwrite race):** the initial mount fetch now goes through the same append+dedup path as WS increments (unified `apply()`, no more full `setMessages(batch)` replace) + an explicit `setMessages([])` reset on run switch. A WS-triggered incremental completing before the initial snapshot can no longer be clobbered, and `lastSeqRef` monotonicity is preserved. Verified the terminal-run transcript still renders in full (27 events + final `已收到F`).
- **P2 (flat 3s reconnect):** replaced with exponential backoff (2s base → 30s cap) + jitter `[capped/2, capped)`, reset to 0 on `auth_ack`. No more lockstep reconnect after a backend restart, and a persistently-failing connect can't hot-loop. (Left auth-fatal-close short-circuit out: each attempt re-mints a fresh `/cli-token`, so a transient token issue self-heals; capped backoff bounds the rest.)

No functional change to the approved paths; `tsc` + dmloop unit tests (23/23) green.

OctoBoooot (dismissed):
## Review: #714 @ `86266ea4` — re-anchoring APPROVE; the two non-blocking 🟡s are now fixed

**Verdict:** Approve — re-anchoring to the current head. The delta (real changes only in `ws.ts` + `RunDetailModal.tsx`; other files rebase-identical) addresses the two non-blocking 🟡s from the prior round with strict improvements. `Build` CI green.

### Resolved — reconnect backoff + jitter (yujiawei's/Steve's 401/no-backoff 🟡)
`ws.ts`: fixed `RECONNECT_MS=3000` → exponential backoff `RECONNECT_BASE_MS(2000) × 2^attempts`, capped at `RECONNECT_MAX_MS(30000)`, with jitter `capped/2 + random*(capped/2)`. `reconnectAttempts` resets to 0 on `auth_ack` (successful connect) and `start()` — so a transient drop reconnects quickly while persistent failure backs off toward the 30s cap, and the jitter avoids thundering-herd on a backend restart. Exactly the prescribed hardening; kills the fixed-3s-forever retry.

### Resolved — unified append+dedup (yujiawei's initial-snapshot-vs-incremental 🟡)
`RunDetailModal.tsx`: the old `apply(batch, incremental)` full-replaced on initial load (`setMessages(batch)`), which could clobber a concurrently-arrived WS increment. The new `apply(batch)` **always** appends+dedups by seq (initial full load and WS increments share one path), with `setMessages([])` on run-switch to reset. Removes the message-loss window; idempotent (seq-keyed), no regression.

### Unchanged and still correct
The subscribe-before-fetch + reconcile-tick fix (my prior major), the `new WebSocket()` try/catch (my minor), the WS client's leak-free lifecycle, token-not-leaked auth, #610 gone/drain preservation, and consumer cleanup — all intact from the prior approved head.

### Net
Both remaining hardening notes closed cleanly with no regression — the realtime WS replacement is now solid end to end. APPROVE carries to `86266ea4`. (`reviewDecision` was `REVIEW_REQUIRED` only because the new commit auto-dismissed the prior three approvals; still subject to the `needs-human-review` maintainer gate before merge.)

Jerry-Xin (dismissed):
Re-review at new head `86266ea4` (prior APPROVE @ 8018fa36, auto-dismissed by this push). This commit addresses both of yujiawei's non-blocking 🟡s — verified at the sink. **APPROVE re-affirmed.**

## ✅ Initial-snapshot vs incremental message gap → FIXED (RunDetailModal.tsx)
`apply` is unified: the `incremental` boolean and its `setMessages(batch)` **full-table-replace** path are removed. Every batch — initial load, WS increments, tail-drain — now goes through the **same seq-based append+dedup**:
```js
const apply = (batch) => {
  if (!batch.length) return;
  lastSeqRef.current = Math.max(lastSeqRef.current, ...batch.map(m => m.seq));
  setMessages(prev => { const seen = new Set(prev.map(m => m.seq)); return [...prev, ...batch.filter(m => !seen.has(m.seq))]; });
};
```
So a WS increment arriving before/during the initial fetch resolution is no longer clobbered by a full-table replace — batches merge by `seq`, no ordering loss / no lost messages. Run-switch reset is preserved via a single `setMessages([])` at effect setup (not per-apply). Genuine correctness fix for the race yujiawei flagged.

## ✅ Reconnect backoff/jitter + 401-infinite-retry → FIXED (ws.ts)
`scheduleReconnect` replaces the fixed 3s interval with **capped exponential backoff + full jitter**:
```js
const capped = Math.min(RECONNECT_MAX_MS /*30000*/, RECONNECT_BASE_MS /*2000*/ * 2 ** this.reconnectAttempts);
this.reconnectAttempts += 1;
const delay = capped / 2 + Math.random() * (capped / 2);   // jitter over [capped/2, capped)
```
`reconnectAttempts` resets to 0 on `auth_ack` (successful reconnect) and on `stop()`. This resolves both sub-concerns: jitter de-synchronizes clients so a backend restart no longer causes a thundering-herd lockstep reconnect, and a persistently-failing connection backs off to a 30s ceiling instead of hammering every 3s. A permanently-rejected auth still retries but now bounded at 30s (transient blips self-heal; permanent is low-cost) — a reasonable trade.

## Residual note (unchanged, still 🟡)
The subscribe + reconcile tick are still gated on the mount-time `run.status` snapshot (the residual I noted before). It remains narrow and self-healing (run status is effectively monotonic active→terminal; the reachable stale-active case is corrected by the reconcile tick's `listRuns` fresh-status query), so still a non-blocking hardening note, not a regression from this commit.

Delta is exactly the two intended files (RunDetailModal.tsx + ws.ts), no extras. CI green (Build + osv-scan success). Both non-blocking 🟡s cleanly resolved — my APPROVE carries and is strengthened.

yujiawei (dismissed):
# Code Review — PR #714 (octo-web)

**Head SHA reviewed:** `86266ea441708121f539eec9a2c23cf981084a11`
**Base:** `main` (merge-base `ede8a038`) · **Verified locally:** `pnpm --filter @octo/loop test` → 23/23 pass; `tsc --noEmit` clean on all changed dmloop files (remaining tsc noise is pre-existing `dmworkbase` react@17 type-resolution, unrelated to this PR).

This PR replaces four independent blind-polling loops with a single per-workspace realtime WebSocket client. The design is clean, the debounce/reconnect/silent-reload plumbing is thoughtful, and the new `ws.ts` is well-guarded (stale-socket `this.ws !== ws` checks, in-flight `ticking`/`pending` guard, listener teardown on stop). Below are the findings; none rise to a merge blocker.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The stated problem in #713 (new issue not appearing after quick-create without a manual reload + settle-poll flicker) is addressed by subscribing to the backend realtime stream and doing a debounced silent refetch.
- **Extra:** none. The change is scoped to removing the four polls (board settle-poll, IssuePage 15s running-chip, RunDetailModal 2s transcript, SquadDetailPage 30s presence) and wiring the WS replacement; the removed `wk:loop-issues-dispatched` MittEvents key is confirmed to have zero remaining references repo-wide.
- **Deviation:** none.

---

## 2. Code quality

**Quality: Approved** (findings are P2 / robustness / verify-items — none block merge)

### Security (this PR was flagged security-sensitive)

- **Handshake token is correctly kept out of the URL.** `ws.ts:104-106` sends the `/cli-token` value in the first WS frame body `{type:"auth",payload:{token}}`, not in the query string (only `workspace_slug` / `client_platform` are in the URL). A fresh token is fetched per `open()`, so reconnects don't reuse a stale token. Good.
- **`changeOrigin:false` on the `/fleet/ws` proxy** (`vite.config.ts:150`) is intentional (preserves same-origin so the backend `checkOrigin` passes) and is dev-only. No secret exposure.

### P2 findings

- **P2 — Orphaned WebSocket on rapid same-slug re-entry (`packages/dmloop/src/api/ws.ts:86-102`).** `open()` awaits `issueLoopCliToken()` before assigning `this.ws`. Under `start(A) → stop() → start(A)` (or `start(A) → start(B) → start(A)`) issued within one token round-trip, the first `open()` resumes and its post-await guard `slug !== this.slug` compares slug **values**, which now match again — so it proceeds and sets `this.ws = ws1`; the second `open()` then overwrites `this.ws = ws2`, and `ws1` is never `.close()`d. The `this.ws !== ws` guards in `onmessage`/`onclose` keep `ws1` inert (no state corruption, no duplicate refresh), so the impact is a single leaked idle authenticated connection, and the trigger is narrow. Low severity, but worth fixing with an epoch/connection-id captured before the await (or `this.ws?.close()` before reassigning at line 102).

- **P2 — `gone` detection can fail to fire for a deleted run in a quiet workspace (`packages/dmloop/src/panel/RunDetailModal.tsx:154-233`).** The `MAX_MISSES=3` counter was designed for 2s polling, where three ticks were guaranteed within ~6s. Under the event-driven model, a deleted run emits a **single** `issue:deleted` → one debounced `wk:loop-issues-refresh` → one `tick()` → `missRef=1`. There is no further self-generated event, so `missRef` only climbs to the threshold via *unrelated* workspace traffic (the refresh signal is workspace-level, not run-filtered). In a quiet workspace the modal can stay frozen on "running" for a run that has actually been deleted — the same freeze class this PR set out to fix, reappearing on the gone path. Consider treating a definitive `issue:deleted` (or a run-not-found immediately after it) as `gone` without requiring N misses.

- **P2 — Redundant tail-fetch on terminal (`packages/dmloop/src/panel/RunDetailModal.tsx:183-193`).** When `fresh.status` is terminal, `stopListening()` runs and `finalTimer` is set, but `tick()` does not `return`. If an event set `pending=true` during this tick's awaits, the `finally` re-invokes `tick()` once more, which re-detects terminal and reassigns `finalTimer` **without** clearing the previous handle. Bounded to one extra scheduled fetch (listeners are already removed, so no further re-entry), and the orphaned timer's callback is `cancelled`-guarded, so no state write — but it fires one unnecessary network request and isn't cleared on unmount. Minor.

- **P2 — Presence refetch storm on `agent:status` (`packages/dmloop/src/panel/SquadDetailPage.tsx:67-71`).** Every `agent:status` event triggers a full `getSquadMemberStatus` HTTP refetch. On a backend restart where many agents flip status at once, this bursts one full-list request per event. Consider debouncing, or applying the status delta directly from the event payload. (Strictly better than the old 30s poll in the common case; only the burst case regresses.)

### Robustness / verify (not code defects)

- **No fallback if the WS transport is fundamentally unavailable.** All polling is removed, so board/transcript/presence updates now depend entirely on the WS connecting. If a corporate proxy strips the `Upgrade` header, or the production gateway isn't configured to proxy `/fleet/ws` → backend `/ws` with WebSocket upgrade, quick-create will succeed but the new issue will **never** auto-appear and there is no visible "disconnected" cue. The PR's own "Deploy coordination" note documents the required gateway change — please treat that gateway config as a hard gate for rollout, and consider a lightweight "realtime disconnected" indicator or a single post-dispatch safety refetch so the failure mode is visible rather than silent.

---

## 3. Overall verdict

**APPROVED**

Spec ✅ and no P0/P1 code defects. The realtime approach is sound and correctly implemented; the findings above are P2 robustness/edge-case improvements, best addressed in a follow-up or a quick amend if convenient.

## Items a human should verify before/at merge

1. **Production gateway** proxies `/fleet/ws` → backend `/ws` **with WebSocket upgrade** (same-origin preferred so `checkOrigin` passes; otherwise `ALLOWED_ORIGINS` must include this origin). Without it the feature silently no-ops in prod.
2. **Backend event-name contract.** This diff hard-codes the WS event vocabulary (`auth_ack`, `issue:created|updated|deleted`, `task:dispatch|queued|running|completed|failed`, `activity:created`, `task:progress`, `agent:status`) and the `{type:"auth",payload:{token}}` handshake. None of this contract lives in this repository, so it cannot be verified from the diff — confirm each name matches the fleet backend exactly. Any mismatch degrades silently to "no refresh."

---

OctoBoooot (approved):
## Review: #714 @ `79c8c950` — re-anchoring APPROVE (comment-only delta)

**Verdict:** Approve — re-anchoring to the current head. Byte-verified the delta vs my prior approved head (`86266ea4`) is **comment-only**: every change across `ws.ts`, `RunDetailModal.tsx`, and `IssuePage.tsx` is a reworded/condensed comment — zero executable-logic change (other files rebase-carried). All verified behavior carries forward unchanged:

- The two 🟡 fixes from the prior round (reconnect exponential backoff+jitter; unified seq append+dedup) — code identical, only the comment above the append logic was trimmed.
- The subscribe-before-fetch + reconcile-tick major fix, the `new WebSocket()` try/catch, leak-free WS lifecycle, token-not-leaked auth, #610 gone/drain preservation, consumer cleanup — all unchanged.

CI green. APPROVE carries to `79c8c950`. (Still subject to the `needs-human-review` maintainer gate before merge — approved ≠ mergeable.)

Jerry-Xin (approved):
🔄 Re-review at head `79c8c950` (previously approved at `86266ea4`). Re-approving.

Verdict: APPROVE.

**Real delta `86266ea4` → `79c8c950` (comment-only)**
The only increment on top of the previously three-way-approved commit is documentation/comment wording — every changed line is a `//` comment; zero executable-code change (verified by blob-hash + a comment-stripped diff that comes back empty):
- `vite.config.ts` / `ws.ts`: reworded the prod-gateway note to say the gateway forwards the `/fleet/ws` WebSocket upgrade to backend `/ws`, and clarified `VITE_FLEET_WS_URL` overrides direct.
- `IssuePage.tsx` / `NewLoopPage.tsx` / `RunDetailModal.tsx` / `SquadDetailPage.tsx`: trimmed/reworded now-redundant inline comments.

Since no logic changed, the previously approved behavior is byte-identical and none of the earlier open non-blocking notes (fast same-slug re-entry, delete-run freeze, redundant tail-fetch) are regressed or affected.

**On the automated "task:cancelled missing from REFRESH_EVENTS" flag — not blocking**
The `REFRESH_EVENTS` set is unchanged from the approved commit, and the concern does not hold against the actual design: cancel resolves through the run-status re-fetch path, not a dedicated `task:cancelled` frame.
- `RunDetailModal` re-fetches `listRuns()` on every trigger and gates on `isActiveRun(fresh.status)`; a cancelled run becomes non-active → it stops listening (terminal). It also has the `MAX_MISSES=3` guard for a vanished run. So the modal cannot stay stuck at "Running" on cancel.
- The board running-chip clears via `IssuePage.refreshRunning()`, driven by `onMutated`/`wk:loop-issues-refresh`, which fires on `issue:updated`/`issue:deleted` emitted by the cancel flow.

CI: green (Build success; all other required checks pass or skipped).

**Highlights**
- Blind polling → workspace-scoped WS singleton with guards for stale sockets, slug switches, malformed frames, and reconnect backfill via `auth_ack`.
- Status-re-fetch-on-event design means the UI converges on terminal state without needing a per-terminal-event whitelist.
- Tests cover debounce, auth handshake, malformed frames, and raw-event subscribe/unsubscribe.

yujiawei (approved):
# Code Review — PR #714 (octo-web)

**Scope reviewed:** diff `origin/main...79c8c950` (merge-base `ede8a038`), 9 files, +340/-71. This is a self-contained frontend change in `packages/dmloop` + a vite proxy rule + a shared event-type cleanup in `dmworkbase`. Reviewed at head SHA `79c8c9504105f724f580733bd0ca9542094c2783`.

## 1. Spec / intent compliance

**Compliant.** The PR does exactly what it claims (issue #713): remove blind polling and drive board + transcript + presence refresh from the existing realtime stream.

- Board settle-poll `[2,5,9,14]s` and the `wk:loop-issues-dispatched` protocol are fully removed; `IssuePage` 15s running-chip poll removed; `RunDetailModal` 2s poll and `SquadDetailPage` 30s poll replaced by WS subscriptions. Verified.
- The now-unused `wk:loop-issues-dispatched` key is removed from the shared `MittEvents` map and has **zero remaining references** repo-wide (`grep` confirmed). `wk:loop-issues-refresh` is retained and consumed by `IssuePage` and `RunDetailModal`. No dangling emit/listener.
- No over-build: no new backend endpoint (reuses `/cli-token` + existing `/ws`), the vite change is additive, and the WS singleton is consumed only within `dmloop`.

## 2. Code quality

Overall the implementation is careful and defensively written: stale-socket guards (`this.ws !== ws`), token-fetch-vs-stop re-check, subscribe-before-fetch ordering in `RunDetailModal`, in-flight `tick()` guard with coalesced trailing re-run, append+dedup on `seq`, and exponential reconnect backoff with jitter. Two hypotheses I specifically tried to break did **not** reproduce:

- **RunDetailModal does NOT freeze on "running" / does NOT lose the terminal transcript.** Terminal statuses (`task:completed`/`task:failed`) are in `REFRESH_EVENTS` → emit `wk:loop-issues-refresh`, which the modal subscribes to; the subscribe-before-fetch ordering catches a terminal transition that lands inside the initial fetch round-trip; `lastSeqRef` advances so the delayed tail-fetch (`TAIL_FETCH_MS`) still pulls late `result`/`error` messages. Sound.
- **The reconnect happy-path is not stuck-dead.** `onclose` → `scheduleReconnect()` → `open()` bypasses `start()`'s early-return guard, so a silent drop on an unchanged slug still reconnects. Correct.

### Findings (all P2 — none blocking)

**[P2] `ws.ts` — `open()` catch-paths omit the workspace-switch guard.** `ws.ts:90-101`. The success path re-checks `if (this.stopped || slug !== this.slug) return;` after the `await`, but the two `catch { this.scheduleReconnect(); }` branches (token fetch throws; `new WebSocket()` throws) do not. If `issueLoopCliToken()` rejects *during* a workspace switch, the stale `open()` schedules a reconnect against the new active context, which can race the new `start()`'s `open()` and leave a duplicate/leaked socket (the older `this.ws` is overwritten but never `close()`d). Narrow (needs a token failure concurrent with a switch), self-limiting, but a clean fix is cheap:
```
} catch { if (!this.stopped && slug === this.slug) this.scheduleReconnect(); return; }
```
Apply the same guard to both catch blocks.

**[P2] No offline signal / no safety-net when WS never connects.** `ws.ts:123` (`onerror` no-op), `IssuePage.tsx`. The reconnect loop runs silently forever and there is no user-facing "live updates offline" affordance. Data itself still recovers on any navigation (`reload` re-runs on `[f, view, page, scope, myMemberId]` and on mount), so this is a *liveness/UX* degradation, not data loss — but since the design now depends on a correctly configured prod gateway (see §4), an unobtrusive offline indicator (e.g. driven off reconnect-attempt count) would make a misconfiguration diagnosable instead of invisible.

**[P2] `SquadDetailPage.tsx:66-69` — `agent:status` refetch has no debounce and loses the periodic self-heal.** `getSquadMemberStatus` fires once per `agent:status` event with no coalescing; `agent:status` is a workspace-level broadcast, so a burst (many agents flipping at once) yields a burst of parallel GETs from this panel. Also, the old 30s interval was a self-heal that no longer exists — if WS is down after the initial `load()`, presence stays stale. Suggest reusing the same debounce pattern already present in `ws.ts` (`scheduleRefresh`, 250ms) for this consumer.

**[P2 / nit] `RunDetailModal.tsx` — terminal branch can schedule a second `finalTimer` without clearing the first.** When the terminal branch runs and a queued `pending` re-invokes `tick()`, the second run re-enters the terminal branch and reassigns `finalTimer` without `clearTimeout` of the prior one. Effect is at most one extra harmless `listRunMessages` call (guarded by `cancelled`); worth a `clearTimeout(finalTimer)` before reassign, or short-circuiting once terminal is reached.

**[P2 / hygiene] `ws.ts:113` — refresh events are processed regardless of `auth_ack`.** `onmessage` acts on `REFRESH_EVENTS` before any auth-acknowledged flag is set. Against your own trusted backend this is effectively unreachable (the server won't push subscription events pre-auth) and the only consequence would be one extra *authenticated* HTTP refetch, so it is not a security issue — but gating event processing on a local `authed` flag set at `auth_ack` would make the client's contract self-evident.

## 3. Overall verdict

**APPROVE.** Spec-compliant, clean removal of the polling machinery with good defensive handling of the hard cases (terminal-transcript race, stale sockets, reconnect). All findings are P2 (resilience/UX/hygiene) — none block merge. Please consider the `open()` catch-path guard and the `SquadDetailPage` debounce as fast-follows.

## 4. Items a human should verify (security-sensitive PR)

These are outside the diff / not verifiable from the frontend alone and matter for a safe rollout:

1. **Prod gateway WS upgrade.** The board's liveness now *entirely* depends on the production gateway proxying `/fleet/ws` with a WebSocket upgrade (same-origin preferred so the backend `checkOrigin` same-origin branch passes; otherwise `ALLOWED_ORIGINS` must include the octo-web origin). There is no client fallback if this is missing — confirm the gateway change ships with (or before) this PR.
2. **Backend server-side auth on `/ws`.** The client sends the session token in the first frame body (good — not in the URL, so it won't appear in access logs). Confirm the backend actually *rejects unauthenticated subscriptions* and validates the token frame before streaming workspace events; the client cannot enforce this.
3. **Event parity for async agent quick-create.** The removed settle-poll was the safety net for "agent creates the issue a few seconds later." This now relies strictly on the backend emitting `issue:created` (mapped to the right workspace) for async agent dispatches. If that event can be dropped, newly dispatched loops silently won't appear until manual refresh — please confirm fail-closed parity backend-side.

## 5. Notes on verification performed

- Traced the full diff and the cross-file consumers (`IssuePage`, `LoopPage`, `RunDetailModal`, `SquadDetailPage`, `App.tsx` event map, `authApi`/`http` token flow, vite proxy ordering).
- Could **not** run the unit tests locally (`node_modules` absent in the review checkout); the added `ws.test.ts` (6 cases: debounce coalescing, non-board-event filtering, `auth_ack` refresh, `on()` dispatch/unsubscribe, malformed-frame safety) is well-targeted and the author reports 23/23 passing — recommend CI confirms.
- The token-in-URL handoff noted elsewhere lives in `LoopCliAuthorizePage.tsx`, which is **not part of this diff** (pre-existing code with its own TODO) and is not a finding against this PR.

PR #715, opened 2026-07-13
fix(docs): allow table column resize drag to start

### LINKED ISSUE #621: [Docs] Table editing UX — column width cannot be resized + unclear row/column delete controls
## Description

In the document editor, table **column widths cannot be dragged/resized**. Attempting to drag a column border to adjust its width has no effect.

## Steps to Reproduce

1. Open a document and insert a table.
2. Hover over a column border and try to drag it to resize the column width.

## Expected Behavior

Dragging a column border resizes that column's width interactively.

## Actual Behavior

The column width does not change — dragging the border has no effect.

## Notes

- Reported by reviewer.
- Investigate whether column resizing is enabled in the table extension (e.g. Tiptap `Table.configure({ resizable: true })` and the `columnResizing` / `tableColumnResizing` plugin), and whether the resize handles are rendered/hit-testable in the editor.
- Part of the table editing work (see #595 for row/column add/remove).

ploy-elison:
## Summary

Table column widths could not be dragged in the docs editor (Tiptap + Yjs). The hover handle showed on the column border, but dragging it did nothing.

Root cause: the self-built cell NodeView `TableCellView.stopEvent` returned `true` for `.column-resize-handle`. That short-circuits ProseMirror's `eventBelongsToView`, so the handle's `mousedown` never reached `prosemirror-tables`' `columnResizing` plugin (`handleDOMEvents.mousedown`) and the drag never started. `mousemove` runs through a non-guarded path, which is why the hover handle still appeared — matching the reported symptom (visible handle, no drag).

The Table config itself is correct (`Table.configure({ resizable: true })`, columnResizing plugin mounted), so the fix is a single event-level change in the NodeView.

## Changes

- `packages/docs/src/editor/TableCellView.ts` (stopEvent, ~line 71-76): return `false` so the resize handle's `mousedown` reaches the columnResizing plugin. Cell DOM writes during resize are still guarded by `ignoreMutation` (kept unchanged), so collaborative cursors don't desync.
- `packages/docs/src/editor/extensions.ts` (~line 138): corrected a stale comment (extension-table is pinned at `3.22.2`, not `2.27.2`).

`ignoreMutation` is intentionally left in place — it prevents the resize plugin's `colwidth → inline width` writes from being re-parsed as content edits under collaboration.

## Testing

- typecheck (`tsc --noEmit`): no new errors introduced (the two pre-existing errors in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts` are present on `main` and unrelated to this change).
- vitest: table-specific suites green — `src/editor/TableControls.test.tsx` + `src/export/docx/table-layout.test.ts` (24/24). The repo's remaining vitest failures are flaky jsdom `requestAnimationFrame` / undici WebSocket environment errors that also fail on a clean checkout (same failure count, different files across runs) and are independent of this change.
- eslint: repo `pnpm lint` (turbo) defines no lint tasks for any package, so there is nothing to run.

Closes #621 (sub-issue 1: column resize drag).

Draft: keep in draft until reviewer / deploy / QA sign-off, then promote to a regular PR.

### DIFF
--- a/packages/docs/src/editor/TableCellView.test.ts
+++ b/packages/docs/src/editor/TableCellView.test.ts
@@ -1,0 +1,72 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import { Editor } from '@tiptap/core'
+import StarterKit from '@tiptap/starter-kit'
+import { Table } from '@tiptap/extension-table'
+import TableRow from '@tiptap/extension-table-row'
+import TableHeader from '@tiptap/extension-table-header'
+import TableCell from '@tiptap/extension-table-cell'
+import type { Node as PMNode } from '@tiptap/pm/model'
+import { TableCellView } from './TableCellView.ts'
+
+// #621-1 regression: the self-built cell NodeView must NOT stop the
+// column-resize-handle mousedown. Returning true there short-circuits
+// prosemirror-view's eventBelongsToView, so prosemirror-tables' columnResizing
+// plugin never sees the mousedown and the column can't be dragged (the handle
+// still shows on hover — matching the original bug report). It must also keep
+// ignoreMutation so the resize plugin's colwidth → inline-width writes on the
+// cell are not re-parsed as content edits under collaboration.
+
+function firstCellNode(): PMNode {
+  const editor = new Editor({
+    extensions: [
+      StarterKit.configure({ undoRedo: false }),
+      Table.configure({ resizable: true }),
+      TableRow,
+      TableHeader,
+      TableCell,
+    ],
+    content:
+      '<table><tbody><tr><td><p>a</p></td><td><p>b</p></td></tr></tbody></table>',
+  })
+  let cell: PMNode | null = null
+  editor.state.doc.descendants((node) => {
+    if (!cell && (node.type.name === 'tableCell' || node.type.name === 'tableHeader')) {
+      cell = node
+      return false
+    }
+    return true
+  })
+  editor.destroy()
+  if (!cell) throw new Error('no table cell node found in seeded document')
+  return cell
+}
+
+let view: TableCellView | null = null
+afterEach(() => {
+  view = null
+})
+
+describe('TableCellView.stopEvent (column resize drag)', () => {
+  it('returns false for a .column-resize-handle mousedown so the drag can start', () => {
+    view = new TableCellView(firstCellNode(), 'td')
+    const handle = document.createElement('div')
+    handle.className = 'column-resize-handle'
+    const event = { type: 'mousedown', target: handle } as unknown as Event
+    // stopEvent is narrowed to () => boolean; call through the PM contract type
+    // to express that the resize-handle event specifically is not stopped.
+    const stopEvent = view.stopEvent as unknown as (e: Event) => boolean
+    expect(stopEvent.call(view, event)).toBe(false)
+  })
+
+  it('does not stop ordinary in-cell events either (typing/selection flow to PM)', () => {
+    view = new TableCellView(firstCellNode(), 'td')
+    const event = { type: 'mousedown', target: view.dom } as unknown as Event
+    const stopEvent = view.stopEvent as unknown as (e: Event) => boolean
+    expect(stopEvent.call(view, event)).toBe(false)
+  })
+
+  it('still ignores attribute mutations on its own cell element (colwidth writes)', () => {
+    view = new TableCellView(firstCellNode(), 'td')
+    expect(view.ignoreMutation({ type: 'attributes', target: view.dom })).toBe(true)
+  })
+})
--- a/packages/docs/src/editor/TableCellView.ts
+++ b/packages/docs/src/editor/TableCellView.ts
@@ -68,11 +68,13 @@ export class TableCellView {
     return false
   }
 
-  /** Don't let the resize plugin's own mousedown/handle events be treated as
-   * editor input; everything else (typing, selection) flows to PM. */
-  stopEvent(event: Event): boolean {
-    const target = event.target as HTMLElement | null
-    if (!target) return false
-    return target.classList?.contains('column-resize-handle') ?? false
+  /** Let every event flow to ProseMirror. In particular the column-resize-handle
+   * mousedown MUST reach prosemirror-tables' columnResizing plugin
+   * (handleDOMEvents.mousedown) — stopping it here short-circuits eventBelongsToView
+   * so the drag never starts (the handle shows on hover but the column can't be
+   * dragged). Content mutations that resizing makes to the cell are already handled
+   * by ignoreMutation above, so nothing needs to be suppressed at the event level. */
+  stopEvent(): boolean {
+    return false
   }
 }
--- a/packages/docs/src/editor/extensions.ts
+++ b/packages/docs/src/editor/extensions.ts
@@ -136,10 +136,10 @@ export function buildExtensions(opts: BuildExtensionsOptions): Extensions {
     // ProseMirror's native drag pipeline, so moves sync as ordinary transactions.
     BlockDragHandle,
     // SCHEMA-SPEC §4 (SCHEMA_VERSION 4): tables. extension-table series pinned at
-    // 2.27.2 (matching the stack — v3 would pull a second Tiptap core). Column
-    // resizing is on; cells use a self-built NodeView (TableCellView) that gives
-    // ProseMirror explicit ignoreMutation/stopEvent rules so resize/remote DOM
-    // writes don't desync collaborative cursors (§3.2 requirement).
+    // 3.22.2 (matching the Tiptap core). Column resizing is on; cells use a
+    // self-built NodeView (TableCellView) that gives ProseMirror explicit
+    // ignoreMutation/stopEvent rules so resize/remote DOM writes don't desync
+    // collaborative cursors (§3.2 requirement).
     Table.configure({ resizable: true }),
     TableRow,
     TableHeader.extend({


OctoBoooot (dismissed):
## Review: fix(docs): allow table column resize drag to start (#715)

**Verdict:** Approve. A correct, well-reasoned two-line fix that removes a harmful event-suppression while preserving the real collab-safety protection. No blockers, no majors.

Base `main`. Risk tier: normal. `mergeStateStatus=CLEAN`.

### The fix is correct
`TableCellView.stopEvent` previously returned `true` for `column-resize-handle` mousedowns — which short-circuits ProseMirror's `eventBelongsToView`, so prosemirror-tables' `columnResizing` plugin (`handleDOMEvents.mousedown`) never receives the event and the drag can't start (handle shows on hover but the column won't resize). The fix returns `false` for all events so the handle mousedown reaches the resize plugin. Root cause correctly identified.

### The safety claim holds (verified)
The old `stopEvent` nominally protected against resize DOM writes being treated as editor input, but that concern is actually handled at the **mutation** level, not the event level: `ignoreMutation` (`TableCellView.ts:62-69`) ignores `attributes` mutations to the cell's own `<td>`/`<th>` (`mutation.target === this.dom`) — which is exactly where the resize plugin writes `width`/`colwidth` (`applyAttrs:39-46`) — so PM's mutation observer never re-parses them as content edits (no collab-cursor desync, the §3.2 requirement). Real content mutations inside the cell still flow to PM. So removing the event-level suppression is safe; the collab protection is untouched. The old `stopEvent` was actively harmful — it swallowed the very mousedown the resize plugin needs.

### `extensions.ts` change
Comment-only correction (table extension version `2.27.2`→`3.22.2` in the doc comment, matching the actual pin); `Table.configure({ resizable: true })` code unchanged. No logic change.

### Note (non-blocking)
No test — but this is a 2-line NodeView behavioral fix whose correctness rests on a framework interaction (ProseMirror `eventBelongsToView` × the tables plugin) that's not meaningfully unit-testable in isolation; the comment documents the mechanism well. Acceptable for the scope.

### Net
Clean, minimal, correctly separates concerns (events flow through so drag works; mutations stay filtered so no desync). Approving.

Jerry-Xin (dismissed):
Byte-verified at head `8b888da2` — clean, minimal, correctly-reasoned fix. **APPROVE.**

## ✅ Root cause + fix correct
The old `stopEvent(event)` returned `true` when `event.target` had class `column-resize-handle`, which suppressed the handle's `mousedown` from reaching prosemirror-tables' `columnResizing` plugin (`handleDOMEvents.mousedown`) via ProseMirror's `eventBelongsToView` short-circuit → the drag never initialized (handle rendered on hover but the column was non-draggable — the reported bug). Returning `false` unconditionally lets the mousedown through and the drag starts. Correct fix for the symptom.

## ✅ No collab-desync regression from dropping the stopEvent guard
Verified the fix's claim that resize DOM writes are covered elsewhere: the collab-safety invariant lives in `ignoreMutation` (TableCellView.ts:64-72), not in `stopEvent`. `ignoreMutation` returns `true` for `type==='attributes' && target===this.dom`, so the resize plugin's `colwidth → inline width / data-colwidth` writes to the cell's own `<td>`/`<th>` (applyAttrs:36-49) are ignored by PM's mutation observer and are NOT re-parsed as content edits that would fight remote cursors. `stopEvent` gates *event propagation*; `ignoreMutation` gates *mutation re-parsing* — orthogonal concerns, and only the latter guards desync. So removing the stopEvent short-circuit is safe; content mutations inside `contentDOM` still flow to PM (ignoreMutation returns `false` for them).

## ✅ extensions.ts — comment-only
The `2.27.2 → 3.22.2` change is purely a version-note correction in the comment (executable-line diff empty); no dependency or behavior change.

CI green, CLEAN, mergeable. Minimal blast radius (stopEvent now unconditionally delegates; ignoreMutation untouched). Nice tightly-scoped fix with a good explanatory comment.

yujiawei (dismissed):
# Code Review — PR #715 (octo-web)

## 1. Spec compliance

**Spec: ✅**

The PR does exactly what its description states, and nothing more:

- **Implemented as required:** `packages/docs/src/editor/TableCellView.ts` — `stopEvent()` now returns `false` unconditionally, so the `.column-resize-handle` `mousedown` reaches `prosemirror-tables`' `columnResizing` plugin (`handleDOMEvents.mousedown`) instead of being short-circuited by `eventBelongsToView`. This is the single event-level change that fixes the reported "handle shows on hover, column can't be dragged" symptom.
- **Comment correction:** `extensions.ts:~139` — the stale `2.27.2` version in the table comment is corrected to `3.22.2`, matching the actual pin.
- **No over-build:** no new flags, config, fields, or behavior added.
- **No missing work for this sub-issue:** `ignoreMutation` is intentionally left in place, which is correct (see below).

## 2. Code quality

**Quality: Approved**

The change is minimal, correct, and well-reasoned.

- **Regression risk — none for a content cell.** `TableCellView` renders a plain `<td>/<th>` where `dom === contentDOM` (the cell is the content hole) and renders no custom interactive DOM of its own (no buttons, menus, or handles — verified). For such a NodeView, `stopEvent → false` is the correct default: returning `true` for typing/selection would break editing, and the old code's only non-`false` case was precisely the resize handle, which is what broke resize. The `.column-resize-handle` is a decoration painted by the `columnResizing` plugin at the editor level, not by this NodeView, so there is no other event source whose events could now "leak."
- **No double-handling.** `columnResizing` listens once via editor-level `handleDOMEvents.mousedown`. `stopEvent → false` routes the event down that single correct path; the previous `true` value suppressed it entirely.
- **`ignoreMutation` remains sufficient.** The resize plugin's `colwidth → inline width` writes land on the cell element (`this.dom`) as attribute/style mutations, and `ignoreMutation` returns `true` for `type === 'attributes'` when `target === this.dom`, so PM's mutation observer won't re-parse them as content edits under collaboration. This logic is unchanged by the PR and continues to hold.

Findings:
- **[P2] No automated regression test for the drag path.** The behavior is DOM/event integration (hard to unit-test), and the existing table suites (`TableControls.test.tsx`, `table-layout.test.ts`) stay green, but a focused test asserting that a `.column-resize-handle` `mousedown` is not stopped by the cell NodeView would lock this fix against future `stopEvent` changes. Non-blocking.
- **[P2] Sibling stale comment left untouched.** The extension-image comment nearby (`extensions.ts`, "pinned 2.27.2") is now also stale — `@tiptap/extension-image` is `3.22.2` in `package.json`. Out of scope for this fix; worth a one-line follow-up for consistency. Non-blocking.

## 3. Overall verdict

**APPROVED**

Spec ✅ and Quality Approved. This is a correct, surgical fix; the two findings are advisory only.

## 4. Suggestions

- Consider a small regression test asserting the cell NodeView does not stop `.column-resize-handle` `mousedown`, so a future `stopEvent` edit can't silently re-break resize.
- Optionally sweep the remaining `2.27.2` version comments (extension-image) to match the actual `3.22.2` pins.

## 5. Additional notes

- One reviewing tool could not access `package.json` and speculatively flagged a possible version mismatch (extension-table not bumped alongside the comment). This was checked directly and **does not apply**: `package.json` already pins the entire `@tiptap` stack — core, pm, extension-image, starter-kit, and the extension-table/-cell/-header/-row series — at `3.22.2`. The comment was simply stale; no dependency change is required.

OctoBoooot (approved):
## Review: #715 @ `162d71b2` — re-anchoring APPROVE; adds the drag regression test (source unchanged)

**Verdict:** Approve — re-anchoring my prior APPROVE (`8b888da2`). The only delta is a new test file; the source fix is byte-identical to what I reviewed. This strictly improves the PR (adds the regression coverage yujiawei and I both flagged as missing).

### Verified
- **Source byte-identical:** `TableCellView.ts` and `extensions.ts` unchanged (same blob SHA) — the fix I approved (drop the harmful `stopEvent` handle-suppression; collab-safety stays in `ignoreMutation`) is untouched.
- **New test (`TableCellView.test.ts`, +72) is meaningful and complete** — three cases that lock both halves of the fix:
  1. `stopEvent` returns `false` for a `.column-resize-handle` mousedown (the #621-1 regression — would fail against the old `stopEvent` that returned `true`),
  2. `stopEvent` returns `false` for ordinary in-cell events,
  3. `ignoreMutation` still returns `true` for attribute mutations on the cell's own element (`:70`) — pinning the collab-desync protection that must survive the `stopEvent` removal.

That's exactly the coverage the fix needed: it guards the fix (drag works) *and* the preserved invariant (no desync), so a future regression in either direction fails the test.

### Net
Clean re-anchor — the source I approved is unchanged, and the added test closes the one non-blocking gap yujiawei and I noted. APPROVE carries to `162d71b2`. (`reviewDecision` was `REVIEW_REQUIRED` only because the new commit auto-dismissed the prior approvals.)

Jerry-Xin (approved):
Re-review at new head `162d71b2d5eb` (prior APPROVE was at 8b888da2, auto-dismissed by the push). **APPROVE re-affirmed.**

## The new commit is test-only — source unchanged
The delta 8b888da2..162d71b2d5eb is a single added file `TableCellView.test.ts` (+72); `TableCellView.ts` (stopEvent/ignoreMutation) and `extensions.ts` are **byte-identical** (verified: source diff empty). So my prior review reasoning carries forward unchanged, and this addresses yujiawei's non-blocking suggestion to add a drag-path regression test.

## The added test is a real, correct regression guard (byte-verified)
- **stopEvent returns false for `.column-resize-handle` mousedown** — the exact #621-1 regression: returning `true` there short-circuits prosemirror-view's `eventBelongsToView` so `columnResizing` never sees the mousedown and the drag can't start. ✅
- **stopEvent returns false for ordinary in-cell events** (typing/selection flow to PM). ✅
- **ignoreMutation returns true for attribute mutations on the cell's own `this.dom`** — guards that the collab-desync protection (resize's colwidth→inline-width writes not re-parsed as content edits) stays intact. ✅
- Not a trivial mock: seeds a real `Editor` with the Tiptap Table stack, extracts a real `tableCell` node, constructs a real `TableCellView`. Runs under vitest in the docs package.

CI green, source unchanged, and the test locks in exactly the invariant the fix rests on. Clean.

yujiawei (approved):
# Code Review — PR #715 (octo-web)

**Verdict: APPROVED** — head SHA `162d71b2d5eb52d3cc673aefdb938a68a9c285c3`

## Summary

Small, well-scoped fix. `TableCellView.stopEvent` was returning `true` for `.column-resize-handle` mousedowns, which short-circuits ProseMirror's `eventBelongsToView` so the handle's mousedown never reached `prosemirror-tables`' `columnResizing` plugin (`handleDOMEvents.mousedown`) — the handle showed on hover but the column couldn't be dragged. Changing `stopEvent()` to always return `false` lets that event through. The root-cause analysis is correct and the fix is minimal.

## Spec / scope compliance ✅

The diff is limited to exactly what the fix requires — three files, no over-build, no unrelated refactors:
- `packages/docs/src/editor/TableCellView.ts` — the `stopEvent` change.
- `packages/docs/src/editor/TableCellView.test.ts` — new regression test.
- `packages/docs/src/editor/extensions.ts` — a stale-comment correction (`2.27.2` → `3.22.2`), which matches `packages/docs/package.json` (`@tiptap/extension-table` is pinned at `3.22.2`). This is a comment-only fix; no dependency bump is involved, and none is needed.

## Correctness

**The change is a strict superset of the old correct behavior, so no regression.** The old `stopEvent` returned `true` *only* for `.column-resize-handle`; every other event already returned `false`. `TableCellView` is a content-editable cell (`contentDOM === dom`), not an atom with its own view-only controls, so — unlike the `Image`/`File`/`Bookmark` NodeViews, which legitimately stop events for their self-owned handles/toolbars — it has no event it needs to intercept. The resize handle belongs to the `columnResizing` *plugin*, not to the cell view, so the cell correctly stops nothing. `stopEvent() => false` is also ProseMirror's own `ViewDesc` default.

**Collaboration invariant preserved.** The `§3.2` desync guard lives in `ignoreMutation` (unchanged by this PR), which returns `true` for attribute mutations on the cell DOM — so the plugin's committed `colwidth → inline width` writes on the `<td>` are not re-parsed as content edits. Live-drag width updates go to the `<colgroup>`/`<col>`/table style (owned by the default table view, outside this NodeView's subtree), and the final width is committed as an ordinary ProseMirror transaction, which syncs safely through Yjs. Nothing about the editable-content event path (typing, selection, IME, paste) changed.

1. **The regression test's forward-guarding power is limited** — `packages/docs/src/editor/TableCellView.test.ts:50-58`. The new `stopEvent()` takes no argument and ignores whatever is passed, so the `stopEvent as unknown as (e: Event) => boolean` cast passes a dead argument. The test genuinely *does* catch this specific regression (it would fail against the old code, which read `event.target.classList`), so it is worth keeping. But going forward it effectively asserts a constant `stopEvent() === false` and can't distinguish a resize-handle event from any other. The second case (`target: view.dom`) is a tautology that passed against the old code too. Consider an integration-level assertion (or at least a comment noting the limitation) if you want durable coverage — a full end-to-end drag isn't practical under jsdom, so this is a reasonable ceiling for the harness.

2. **A sibling stale comment was missed** — `packages/docs/src/editor/extensions.ts:156`. This PR fixed the `2.27.2 → 3.22.2` comment on the table line but the analogous `OctoImage` comment 15 lines below still reads `(pinned 2.27.2, single core)`, while `@tiptap/extension-image` is also `3.22.2` in `package.json`. Same defect the PR set out to remove; fixing both in one pass keeps the file coherent.

## Coverage / blind spots

- Dependencies are not installed in the review checkout, so the new vitest suite was **not executed** and the fix was **not exercised in a real browser**. Recommend the author confirm end-to-end by dragging a column in the docs editor, plus one two-client collaborative sanity check that a remote peer's cursor does not jump during a resize (the `ignoreMutation` path).
- Reviewed against the pinned dependency contract (`prosemirror-view`/`prosemirror-tables`/`@tiptap/core` at `3.22.2`); the `eventBelongsToView → stopEvent → handleDOMEvents.mousedown` short-circuit chain was confirmed against ProseMirror internals.
- An independent line-level cross-model reviewer leg was attempted but did not complete due to an unrelated gateway configuration error; it is recorded as absent, not as tacit approval. This review reflects the checks that did run.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #715 (octo-web)
Head SHA: 162d71b2d5eb52d3cc673aefdb938a68a9c285c3

## Summary

Fixes a column-resize drag regression in the docs editor's table component. The self-built `TableCellView` NodeView was returning `true` from `stopEvent` for `.column-resize-handle` mousedown events, which short-circuited ProseMirror's event pipeline and prevented `prosemirror-tables`' `columnResizing` plugin from ever receiving the mousedown — the resize handle appeared on hover but the column could not be dragged. The fix simplifies `stopEvent` to always return `false`, relying on the already-present `ignoreMutation` guard to protect collaborative editing from resize-plugin DOM writes. A focused regression test suite is added alongside.

## Verification

Static analysis only at head `162d71b2`; build and tests not executed in this environment.

- ✅ **Fix correctness** — `TableCellView.stopEvent()` at `packages/docs/src/editor/TableCellView.ts:77` now returns `false` unconditionally. The old code (`target.classList?.contains('column-resize-handle')`) was the exact inverse of what the resize plugin needs.
- ✅ **Collaboration guard preserved** — `ignoreMutation` at `packages/docs/src/editor/TableCellView.ts:58-67` still returns `true` for attribute mutations on `this.dom`. The resize plugin's `colwidth → style.width` writes are still filtered from PM's mutation observer, so collaborative cursors remain unaffected.
- ✅ **Dual-path parity (C1)** — Both `TableHeader.extend` and `TableCell.extend` at `packages/docs/src/editor/extensions.ts:143-152` use the same `TableCellView` class. The fix applies uniformly to `<th>` and `<td>`.
- ✅ **Test coverage** — Three tests at `packages/docs/src/editor/TableCellView.test.ts:49-69` lock the regression: resize-handle mousedown returns `false`, ordinary in-cell mousedown returns `false`, and `ignoreMutation` still returns `true` for cell attribute mutations.
- ✅ **Version comment accuracy** — `extensions.ts:139` now says `3.22.2`; `packages/docs/package.json:37` pins `@tiptap/extension-table` at `3.22.2`. Comment matches reality.
- ✅ **Method signature consistency** — `stopEvent(): boolean` drops the `event` parameter vs. sibling NodeViews (`ImageNodeView.ts:225`, `FileAttachmentNodeView.ts:145`, `BookmarkNodeView.ts:150`) that still accept it. Correct: those NodeViews inspect the event target for their view-only DOM; `TableCellView` no longer needs to.

## Findings

The fix is minimal, well-reasoned, and the regression tests properly lock the behavior. The `ignoreMutation` guard (which protects the collaboration contract) is verified both in the source code and in the test suite. The comment update in `extensions.ts` is a trivial accuracy correction.

## Data Flow Trace

|---|---|---|
| `stopEvent()` return value `false` | `TableCellView.stopEvent()` at `:77` | ✅ ProseMirror's `NodeViewDesc.stopEvent` receives `false` → event pipeline continues → columnResizing plugin fires |
| `ignoreMutation({type:'attributes', target:this.dom})` | `TableCellView.ignoreMutation()` at `:58` | ✅ Returns `true` → PM's mutation observer skips resize-plugin DOM writes → no spurious content edits |
| `TableCellView` used for both `<th>` and `<td>` | `extensions.ts:143-152` via `addNodeView()` | ✅ Same class → same fix applies to both cell types |
| `@tiptap/extension-table` version `3.22.2` | `package.json:37` | ✅ Matches the updated comment in `extensions.ts:139` |

## Blindspot Checklist (R5)

- **C1 — Dual-path parity:** Clear. Both `TableHeader.extend` and `TableCell.extend` use the same `TableCellView` class; the fix applies uniformly.
- **C2 — Control-flow ordering / nesting:** Clear. `stopEvent` is called once per event by ProseMirror's NodeView pipeline; no nested or re-entrant call sites.
- **C3 — Authorization boundary:** N/A. No auth/permission changes.
- **C4 — Authorization lifecycle:** N/A. No auth changes.
- **C5 — Build/runtime divergence:** N/A. No build config, packaging, or environment changes.
- **C6 — Governance/doc self-consistency:** N/A. No governance or policy changes.

## Verdict: APPROVED

Clean, minimal fix with proper regression coverage. The `stopEvent` simplification is correct because `ignoreMutation` already handles the collaboration-protection contract that `stopEvent` was redundantly trying to serve. The old conditional `return true` for `.column-resize-handle` targets was the exact bug — it prevented the columnResizing plugin from receiving the mousedown that initiates drag. No issues found.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; correct fix with proper regression tests, preserved collaboration guard, and accurate comment update.

yujiawei (approved):
# Code Review — PR #715 (octo-web)

## Summary

Focused, minimal fix for the docs-editor table column-resize drag bug. The self-built cell NodeView (`TableCellView`) previously returned `true` from `stopEvent()` for `.column-resize-handle`, which short-circuited ProseMirror's `eventBelongsToView`, so the `columnResizing` plugin never received the handle's `mousedown` and the drag never started (the handle still showed on hover — matching the reported symptom). The fix makes `stopEvent()` return `false` unconditionally. Changes are 3 files: the fix, a new regression test, and a stale-comment correction.

**Verdict: APPROVED** — no P0/P1 blocking issues. A couple of non-blocking nits below.

## Spec compliance

✅ **Correct and complete for the stated scope.**
- **Missing:** none — the fix (`stopEvent` → `false`), a regression test, and the version-comment correction fully cover the bug.
- **Over-build:** none — no new flags/features/exports; `TableCellView` constructor / `update` / `ignoreMutation` are untouched.
- **Deviation:** none — dropping the unused `event` param (`stopEvent(): boolean`) is a valid narrowing and does not break the ProseMirror NodeView contract.

## Correctness

The root-cause analysis is mechanically sound, verified against the installed dependency source:
- `prosemirror-view` gates the built-in `mousedown` listener behind `eventBelongsToView(view, event) && !runCustomHandler(...)`. `eventBelongsToView` walks from `event.target` up to `view.dom` and returns `false` if any ancestor's `pmViewDesc.stopEvent(event)` returns `true`.
- `prosemirror-tables` mounts the resize handle as a widget decoration **inside** the cell, so its `mousedown` routes through the `<td>`'s NodeView = `TableCellView.stopEvent`. The old `true` return therefore suppressed the very event the `columnResizing` plugin (`handleDOMEvents.mousedown`, dispatched via `runCustomHandler`) needs.

No regression is reintroduced by returning `false` for all events:
- The removed guard only returned `true` for the resize handle; everything else (typing/selection) already returned `false`. The only behavioral delta is the handle case — i.e. the bug.
- Content-mutation protection lives in `ignoreMutation` (unchanged), **not** `stopEvent`. The `colwidth → inline width` writes the resize plugin makes to the cell are still guarded (`ignoreMutation` returns `true` for `attributes` mutations on `this.dom`), so collaborative cursors don't desync.
- The content-vs-atom distinction from the sibling NodeViews is correct: `ImageNodeView` / `BookmarkNodeView` / `FileAttachmentNodeView` are atom nodes owning view-only DOM, so they legitimately stop events for private controls; `TableCellView` is a **content** NodeView (`contentDOM === dom`), so returning `false` is the correct contract.

## Test quality

The new `TableCellView.test.ts` is a genuine regression guard, not a tautology despite the no-arg signature: it constructs a `.column-resize-handle` target and asserts `stopEvent` returns `false`. Reverting the implementation to the old body flips that assertion to red (`expected true to be false`), so the test would have caught the original bug. It also pins the `ignoreMutation` colwidth guard. Tests pass locally (3/3).

## Non-blocking nits (no action required to merge)

- **Twin stale version comment (nit):** the PR correctly fixes the `extension-table` comment (`2.27.2` → `3.22.2`), but the same-file `OctoImage` comment (~`extensions.ts:156`, `pinned 2.27.2, single core`) is now identically stale — `@tiptap/extension-image` is also `3.22.2` in `package.json`. Same staleness also appears in `ImageNode.ts` / `schema/index.ts`. Easy same-PR or follow-up cleanup.
- **Test type-casts (nit):** the tests double-cast (`as unknown as (e: Event) => boolean`, and a mock literal for `ignoreMutation`) to work around the narrowed signature. Intentional and behavior is still asserted via return value; declaring `stopEvent(_event?: Event)` on the class would let the test drop the cast. Purely stylistic.

The `3.22.2` version claim in the corrected comment is accurate — `package.json` and the lockfile both pin `@tiptap/extension-table` at `3.22.2`, so this is a stale-comment correction, not a dependency bump (no manifest change needed).

Issue #595, opened 2026-07-10
🐛 [Docs] 表格默认 3×3 且无法增删行/列 — 缺少行列增删入口


jeff-wilson2010:
## Summary

Inserting a table in a doc always produces a fixed 3×3 grid, and there is **no way to add or delete rows/columns afterward**. Reproduces both in the in-app doc tab editor and when the doc is opened in a standalone page.

## Steps to reproduce

1. Open a document (either the in-chat doc tab editor or the standalone doc page).
2. Use the toolbar **Table** button to insert a table.
3. A 3-row × 3-column table (with header row) is inserted.
4. Try to add or remove a row/column — right-click a cell, hover for a control handle, etc.

## Actual

- Table is fixed at 3×3.
- No context menu, no floating handle, and no toolbar control for insert/delete row or column. There is simply no affordance to change the row/column count.

## Expected

Standard table editing affordances: add row above/below, add column left/right, delete row, delete column (e.g. a cell context menu or hover handles).

## Notes / likely root cause

- Toolbar only exposes an insert action with a hardcoded size: `insertTable({ rows: 3, cols: 3, withHeaderRow: true })` (`packages/docs/src/editor/Toolbar.tsx`).
- The `@tiptap/extension-table` series is wired up (`packages/docs/src/editor/extensions.ts`), so the underlying commands (`addRowAfter`, `addColumnAfter`, `deleteRow`, `deleteColumn`, etc.) exist — they are just **not surfaced in any UI**. Only column *resizing* is enabled (`Table.configure({ resizable: true })`); there is no add/delete row/column UI.

## Environment

- Reported by an internal tester; occurs in both the doc-tab editor and standalone doc page.

Jerry-Xin:
Verified by the reviewer after the fix: table insertion now works as expected (row/column add/remove available). Closing confirmed.

PR #716, opened 2026-07-13
fix(docs): keep table toolbar clear of text and label delete controls (#625, #621-2)

### LINKED ISSUE #621: [Docs] Table editing UX — column width cannot be resized + unclear row/column delete controls

### LINKED ISSUE #625: [Docs] Table control toolbar covers text whenever cursor is inside the table
## Description

When the cursor is placed **anywhere inside a table**, a floating control toolbar (the row/column add/delete controls) appears and **overlaps/covers the surrounding text**, obstructing readability while editing.

## Steps to Reproduce

1. Open a document that contains a table with text around/inside it.
2. Click to place the cursor inside any table cell.
3. A floating control box (row/column controls) appears and covers nearby text.

## Actual Behavior

The floating table control toolbar shows persistently whenever the cursor is inside the table, and it is positioned so that it covers/obscures document text.

## Expected Behavior

The table controls should not obscure content. Options:
- Position the controls in the table gutter/margins (outside the text flow) rather than on top of text.
- Show controls on hover of the specific row/column edge instead of persistently while the cursor is in the table.
- Ensure the controls have proper layering/offset so they never cover readable text.

## Notes

- Reported by reviewer with screenshot.
- Related to the table editing UX tracked in #621 (row/column controls + column resize). This is specifically about the controls overlay covering text.

ploy-elison:
## Summary

Fixes two same-origin issues with the docs table editing toolbar (`TableBubbleMenu`):

- **#625 — toolbar covers surrounding text.** The floating toolbar was anchored to the caret's cell rect with `placement: 'top'`, so once the caret sat in row 2+ the toolbar floated 8px above the cell and landed on top of the text rows above it. It now anchors to the table's outer edge (its `.tableWrapper` scroll box when present) via the BubbleMenu `getReferencedVirtualElement`, so with `placement: 'top'` it floats in the block margin above the whole table and never overlaps in-table text, regardless of caret row.
- **#621-2 — delete controls unclear.** The delete row / column / table controls were icon-only, giving no indication of what would be removed. They now show a text caption next to the icon (reusing the existing `docs.table.deleteRow` / `deleteColumn` / `deleteTable` i18n strings, en-US + zh-CN).

## Changes

- `packages/docs/src/editor/TableControls.tsx`
  - New `tableReferenceElement(editor)` helper returning a Floating-UI virtual element anchored to the table's outer edge (lazy rect so scroll/resize stays accurate); wired via `getReferencedVirtualElement`. Falls back to the default caret positioning when no table is found.
  - `TbBtn` gains an optional `text` prop; the three delete controls pass it.
- `packages/docs/src/editor/styles.css`
  - Inline layout for icon + caption (`.octo-tb-btn--labeled`, `.octo-tb-btn-label`). Icon-only buttons are unaffected.

## Scope / safety

- Changes are confined to `TableControls.tsx` and `styles.css`. `TableCellView.ts` is untouched, and the portal `pointer-events` passthrough that keeps the column-resize hot zone reachable (`.octo-table-bubble-portal`) is unchanged.
- The visible caption is decorative for screen readers — the existing `aria-label`/`title` already carry the same string, so there is no double announcement.

## Verification

- `tsc --noEmit` (docs `typecheck`): no new errors in changed files. (Two pre-existing errors in unrelated files — `dmworkbase/src/i18n/format.ts`, `src/members/sort.ts` — are present on `main` as well.)
- `vitest` — `TableControls.test.tsx` passes 10/10, including the #595 C1 pointer-events case. The full docs suite result is identical to `main` (one pre-existing `Toolbar.test.tsx` animation-frame teardown flake that only surfaces in the full run and passes in isolation).
- ESLint: the repo has no configured lint pipeline for the docs package (root config is commented out, no `lint` script); a standalone `@typescript-eslint` self-check on the changed file is clean.

Closes #625
Closes #621 (sub-issue 2)

### DIFF
--- a/packages/docs/src/editor/TableControls.test.tsx
+++ b/packages/docs/src/editor/TableControls.test.tsx
@@ -8,7 +8,7 @@ import TableHeader from '@tiptap/extension-table-header'
 import TableCell from '@tiptap/extension-table-cell'
 import { CellSelection } from '@tiptap/pm/tables'
 import { TextSelection } from '@tiptap/pm/state'
-import { shouldShowTableBubble, TableGridPicker, TableBubbleMenu } from './TableControls.tsx'
+import { shouldShowTableBubble, TableGridPicker, TableBubbleMenu, clampToolbarAnchorRect } from './TableControls.tsx'
 
 // #595 — table add/delete row/column UI. The critical acceptance point is that the controls work on
 // tables that ALREADY EXIST in a document (parsed from stored HTML), not only freshly inserted
@@ -173,6 +173,45 @@ describe('TableBubbleMenu — does not block the column-resize hot zone (#595 C1
     expect(buttons.length).toBeGreaterThan(0)
     e.destroy()
     host.remove()
+  })
+})
+
+describe('clampToolbarAnchorRect — keep the toolbar anchor inside the viewport (#625 off-screen)', () => {
+  const VIEWPORT = { width: 1200, height: 1000 }
+
+  it('leaves a fully-visible table at its real top edge (no TC-P0-001 regression)', () => {
+    // A normal table sitting mid-viewport: the band stays at the real top so the toolbar still
+    // floats above the table exactly as before.
+    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: 200, right: 900 }, VIEWPORT)
+    expect(r.top).toBe(300)
+    expect(r.height).toBe(0)
+    expect(r.left).toBe(200)
+    expect(r.right).toBe(900)
+  })
+
+  it('clamps a long table scrolled past the top back inside the viewport', () => {
+    // The failing case: 34-row table scrolled so its top is far above the viewport and its bottom
+    // is below it. Previously the toolbar landed at ~1051 (below a 1000px viewport); now the anchor
+    // band is clamped to TOOLBAR_SAFE_TOP so the controls render near the top of the screen.
+    const r = clampToolbarAnchorRect({ top: -1248, bottom: 1052, left: 200, right: 900 }, VIEWPORT)
+    expect(r.top).toBe(64)
+    expect(r.bottom).toBe(64)
+    // and well within the viewport
+    expect(r.top).toBeGreaterThan(0)
+    expect(r.bottom).toBeLessThan(VIEWPORT.height)
+  })
+
+  it('never drops the band below the viewport bottom', () => {
+    const r = clampToolbarAnchorRect({ top: 5000, bottom: 6000, left: 100, right: 400 }, VIEWPORT)
+    expect(r.top).toBe(VIEWPORT.height - 8)
+    expect(r.top).toBeLessThan(VIEWPORT.height)
+  })
+
+  it('clamps the horizontal extent of a wide / off-screen table into the viewport', () => {
+    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: -500, right: 3000 }, VIEWPORT)
+    expect(r.left).toBe(0)
+    expect(r.right).toBe(VIEWPORT.width)
+    expect(r.width).toBe(VIEWPORT.width)
   })
 })
 
--- a/packages/docs/src/editor/TableControls.test.tsx
+++ b/packages/docs/src/editor/TableControls.test.tsx
@@ -176,42 +176,60 @@ describe('TableBubbleMenu — does not block the column-resize hot zone (#595 C1
   })
 })
 
-describe('clampToolbarAnchorRect — keep the toolbar anchor inside the viewport (#625 off-screen)', () => {
+describe('clampToolbarAnchorRect — keep the toolbar clear of the fixed toolbar (#625 off-screen + #716)', () => {
   const VIEWPORT = { width: 1200, height: 1000 }
+  // Bottom edge of the fixed editor toolbar in viewport coords, as tableReferenceElement measures it.
+  const TOOLBAR_BOTTOM = 64
 
   it('leaves a fully-visible table at its real top edge (no TC-P0-001 regression)', () => {
-    // A normal table sitting mid-viewport: the band stays at the real top so the toolbar still
-    // floats above the table exactly as before.
-    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: 200, right: 900 }, VIEWPORT)
+    // A normal table sitting mid-viewport, well below the fixed toolbar: the band stays a
+    // zero-height anchor at the real top so the toolbar still floats above the table as before.
+    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: 200, right: 900 }, VIEWPORT, TOOLBAR_BOTTOM)
     expect(r.top).toBe(300)
     expect(r.height).toBe(0)
     expect(r.left).toBe(200)
     expect(r.right).toBe(900)
   })
 
-  it('clamps a long table scrolled past the top back inside the viewport', () => {
-    // The failing case: 34-row table scrolled so its top is far above the viewport and its bottom
-    // is below it. Previously the toolbar landed at ~1051 (below a 1000px viewport); now the anchor
-    // band is clamped to TOOLBAR_SAFE_TOP so the controls render near the top of the screen.
-    const r = clampToolbarAnchorRect({ top: -1248, bottom: 1052, left: 200, right: 900 }, VIEWPORT)
-    expect(r.top).toBe(64)
-    expect(r.bottom).toBe(64)
-    // and well within the viewport
-    expect(r.top).toBeGreaterThan(0)
-    expect(r.bottom).toBeLessThan(VIEWPORT.height)
+  it('anchors a long table scrolled past the top to the toolbar strip so it flips below (XIN-939)', () => {
+    // The XIN-939 case: 34-row table scrolled so its top is far above the viewport and its bottom
+    // is below it. The band becomes the toolbar strip [0 .. toolbarBottom]; placement:'top' then
+    // overflows the viewport top and Floating-UI flips the controls to just below the fixed toolbar
+    // (top === 0 is what forces that flip; bottom === toolbarBottom is where they land).
+    const r = clampToolbarAnchorRect({ top: -1248, bottom: 1052, left: 200, right: 900 }, VIEWPORT, TOOLBAR_BOTTOM)
+    expect(r.top).toBe(0)
+    expect(r.bottom).toBe(TOOLBAR_BOTTOM)
+    expect(r.left).toBe(200)
+    expect(r.right).toBe(900)
+  })
+
+  it('anchors a table jammed under the fixed toolbar to the strip so it flips below it (#716)', () => {
+    // The #716 follow-up: a table at the very top of the doc sits right beneath a tall fixed
+    // toolbar (bottom 180). Floating above would land the controls behind the fixed toolbar; the
+    // strip band flips them to just below it instead.
+    const tallToolbar = 180
+    const r = clampToolbarAnchorRect({ top: 185, bottom: 405, left: 200, right: 900 }, VIEWPORT, tallToolbar)
+    expect(r.top).toBe(0)
+    expect(r.bottom).toBe(tallToolbar)
   })
 
   it('never drops the band below the viewport bottom', () => {
-    const r = clampToolbarAnchorRect({ top: 5000, bottom: 6000, left: 100, right: 400 }, VIEWPORT)
+    const r = clampToolbarAnchorRect({ top: 5000, bottom: 6000, left: 100, right: 400 }, VIEWPORT, TOOLBAR_BOTTOM)
     expect(r.top).toBe(VIEWPORT.height - 8)
     expect(r.top).toBeLessThan(VIEWPORT.height)
   })
 
   it('clamps the horizontal extent of a wide / off-screen table into the viewport', () => {
-    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: -500, right: 3000 }, VIEWPORT)
+    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: -500, right: 3000 }, VIEWPORT, TOOLBAR_BOTTOM)
     expect(r.left).toBe(0)
     expect(r.right).toBe(VIEWPORT.width)
     expect(r.width).toBe(VIEWPORT.width)
+  })
+
+  it('with no fixed toolbar (toolbarBottom defaults to 0) keeps a fully-visible table at its top', () => {
+    const r = clampToolbarAnchorRect({ top: 300, bottom: 520, left: 200, right: 900 }, VIEWPORT)
+    expect(r.top).toBe(300)
+    expect(r.height).toBe(0)
   })
 })
 
--- a/packages/docs/src/editor/TableControls.tsx
+++ b/packages/docs/src/editor/TableControls.tsx
@@ -80,23 +80,56 @@ function TbBtn({
   onClick,
   label,
   title,
+  text,
 }: {
   onClick: () => void
   label: ReactNode
   title: string
+  // When set, a visible text caption is shown next to the icon. Used for the destructive
+  // delete controls (#621-2) so the user reads which row/column/table the action removes,
+  // instead of guessing from an icon alone.
+  text?: string
 }) {
   return (
     <button
       type="button"
-      className="octo-tb-btn"
+      className={'octo-tb-btn' + (text ? ' octo-tb-btn--labeled' : '')}
       title={title}
       aria-label={title}
       onMouseDown={(e) => e.preventDefault()}
       onClick={onClick}
     >
       {label}
+      {text ? <span className="octo-tb-btn-label">{text}</span> : null}
     </button>
   )
+}
+
+/**
+ * Reference rect for the floating table toolbar (#625). Anchors to the OUTER edge of the table
+ * (its scroll wrapper when present) rather than the caret's cell, so the toolbar — placed above
+ * with `placement: 'top'` — floats in the block margin above the whole table instead of landing on
+ * top of the text rows above the caret. Returns a Floating-UI virtual element (computed lazily so
+ * scrolling/resizing stays accurate), or null when the caret is not inside a table, in which case
+ * the BubbleMenu falls back to its default cursor-based positioning.
+ */
+function tableReferenceElement(editor: Editor): { getBoundingClientRect: () => DOMRect; getClientRects: () => DOMRect[] } | null {
+  const { $from } = editor.state.selection
+  for (let depth = $from.depth; depth > 0; depth--) {
+    if ($from.node(depth).type.name === 'table') {
+      const dom = editor.view.nodeDOM($from.before(depth))
+      if (dom instanceof HTMLElement) {
+        // Prefer the ProseMirror `.tableWrapper` scroll box so a horizontally scrolled wide table
+        // still anchors to the visible table region.
+        const anchor = (dom.closest('.tableWrapper') as HTMLElement | null) ?? dom
+        return {
+          getBoundingClientRect: () => anchor.getBoundingClientRect(),
+          getClientRects: () => [anchor.getBoundingClientRect()],
+        }
+      }
+    }
+  }
+  return null
 }
 
 /**
@@ -127,6 +160,11 @@ export function TableBubbleMenu({ editor }: { editor: Editor }) {
       // plugin listens for on the editor DOM. octo-table-bubble-portal makes the floating wrapper
       // transparent to pointer events (see styles.css); only the buttons below re-capture them.
       className="octo-table-bubble-portal"
+      // Anchor the toolbar to the table's outer edge rather than the caret's cell, so with
+      // placement: 'top' it floats in the block margin above the whole table and never covers the
+      // in-table text rows above the caret (#625). Falls back to the default caret rect when the
+      // helper can't find a table (never happens while shouldShow is satisfied).
+      getReferencedVirtualElement={() => tableReferenceElement(editor)}
       options={{ placement: 'top', offset: 8 }}
       shouldShow={({ editor: e }) => shouldShowTableBubble(e)}
     >
@@ -144,6 +182,7 @@ export function TableBubbleMenu({ editor }: { editor: Editor }) {
         <TbBtn
           label={<IconDeleteRow />}
           title={t('docs.table.deleteRow')}
+          text={t('docs.table.deleteRow')}
           onClick={() => editor.chain().focus().deleteRow().run()}
         />
         <span className="octo-tb-sep" />
@@ -160,12 +199,14 @@ export function TableBubbleMenu({ editor }: { editor: Editor }) {
         <TbBtn
           label={<IconDeleteCol />}
           title={t('docs.table.deleteColumn')}
+          text={t('docs.table.deleteColumn')}
           onClick={() => editor.chain().focus().deleteColumn().run()}
         />
         <span className="octo-tb-sep" />
         <TbBtn
           label={<IconDeleteTable />}
           title={t('docs.table.deleteTable')}
+          text={t('docs.table.deleteTable')}
           onClick={() => editor.chain().focus().deleteTable().run()}
         />
       </div>
--- a/packages/docs/src/editor/TableControls.tsx
+++ b/packages/docs/src/editor/TableControls.tsx
@@ -105,65 +105,85 @@ function TbBtn({
   )
 }
 
-// Vertical space (px) reserved above the anchor band so the toolbar — placed above the anchor with
-// `placement: 'top'` — always has room to render inside the viewport. Covers the ~40px control row
-// plus the 8px placement offset with margin to spare.
-const TOOLBAR_SAFE_TOP = 64
+// Vertical clearance (px) the floating toolbar needs to sit ABOVE a table without colliding with
+// the fixed editor toolbar: the ~32px control row plus the 8px placement offset, with margin. Used
+// only to decide whether the table sits far enough below the fixed `.octo-toolbar` for the toolbar
+// to keep floating above it, or whether it must instead drop BELOW the fixed toolbar.
+const TOOLBAR_CLEARANCE = 48
 
 /**
- * Clamp a table's rect into the visual viewport and collapse it to a thin anchor band near its
- * visible top (#625 off-screen fix). This is the pure core of {@link tableReferenceElement},
- * split out so it can be unit-tested without a real layout.
+ * Clamp a table's rect into the visual viewport and collapse it to an anchor band, keyed to the
+ * bottom of the fixed editor toolbar (`.octo-toolbar`), so the floating table toolbar never renders
+ * behind — and gets its clicks swallowed by — that fixed toolbar (#625 off-screen fix + #716
+ * follow-up). Pure core of {@link tableReferenceElement}, split out so it can be unit-tested
+ * without a real layout.
  *
- * Why: the toolbar anchors to the table's OUTER edge and floats above it with `placement: 'top'`.
- * For a long table scrolled so its top leaves the viewport (`rect.top` a large negative), the raw
- * rect's top is off-screen above and its bottom is far below the viewport; Floating-UI's default
- * `flip` then re-anchors to that off-screen bottom edge and the controls land below the viewport
- * (observed: toolbar top ~1051 vs a 1000px viewport), completely unreachable. `shift` doesn't save
- * it because its default boundary is the tall scroll container, not the visual viewport.
+ * The toolbar anchors to the table's OUTER edge and floats above it with `placement: 'top'`. Two
+ * boundaries have to hold at once:
  *
- * Fix: build the reference from a zero-height band whose `top` is the table's top edge CLAMPED into
- * `[TOOLBAR_SAFE_TOP, viewport.height - 8]`, with the horizontal extent clamped into the viewport.
- * A thin band near the top means both `top` and any flipped `bottom` placement resolve to a y just
- * inside the viewport, so the controls stay reachable regardless of flip/shift boundary quirks. In
- * the normal, fully-visible case the table top already sits below `TOOLBAR_SAFE_TOP`, so the band
- * stays at the real top edge and the toolbar floats above the table exactly as before — no
- * TC-P0-001 regression.
+ *  - **Long table scrolled past the top (XIN-939 / #625).** `rect.top` is a large negative and the
+ *    bottom is far below the viewport; Floating-UI's default `flip` would re-anchor to that
+ *    off-screen bottom edge and push the controls below the viewport, unreachable.
+ *  - **Table jammed under the fixed toolbar (#716 follow-up).** A table at the very top of the doc
+ *    sits right beneath the fixed `.octo-toolbar`; floating the controls ABOVE it lands them in the
+ *    band the fixed toolbar occupies, which is opaque and higher in the stack, so it covers the
+ *    controls and intercepts their pointer events.
+ *
+ * Fix: when the table's top is far enough below the fixed toolbar (with {@link TOOLBAR_CLEARANCE}
+ * room to float above), keep a zero-height band at the real top edge — the normal, fully-visible
+ * case is untouched (no #625 / TC-P0-001 regression). Otherwise anchor to the toolbar strip
+ * `[0 .. toolbarBottom]`: `placement: 'top'` then overflows the viewport top, Floating-UI's `flip`
+ * drops the controls to BOTTOM, and they land just below the fixed toolbar in the clear — reachable
+ * and never behind the fixed toolbar, for both the off-screen and top-of-doc cases.
+ *
+ * `toolbarBottom` is the fixed toolbar's bottom edge in viewport coordinates (0 when there is no
+ * toolbar, e.g. a read-only view), clamped defensively into the viewport.
  */
 export function clampToolbarAnchorRect(
   rect: { top: number; bottom: number; left: number; right: number },
   viewport: { width: number; height: number },
+  toolbarBottom = 0,
 ): DOMRect {
   const left = Math.max(0, Math.min(rect.left, viewport.width))
   const right = Math.min(viewport.width, Math.max(rect.right, left))
   const width = Math.max(0, right - left)
-  // Keep the band at least TOOLBAR_SAFE_TOP from the viewport top (room for the toolbar above it)
-  // and never below the viewport bottom, following the real top edge in between.
-  const top = Math.max(TOOLBAR_SAFE_TOP, Math.min(rect.top, viewport.height - 8))
-  return {
-    x: left,
-    y: top,
-    left,
-    top,
-    right: left + width,
-    bottom: top,
-    width,
-    height: 0,
-    toJSON() {
-      return this
-    },
-  } as DOMRect
+  const safeTop = Math.max(0, Math.min(toolbarBottom, viewport.height - 8))
+  const build = (top: number, bottom: number): DOMRect =>
+    ({
+      x: left,
+      y: top,
+      left,
+      top,
+      right: left + width,
+      bottom,
+      width,
+      height: Math.max(0, bottom - top),
+      toJSON() {
+        return this
+      },
+    } as DOMRect)
+  // Table sits clear below the fixed toolbar with room for the controls to float above it: keep a
+  // zero-height band at the real top edge so the toolbar floats above the table exactly as before.
+  if (rect.top - TOOLBAR_CLEARANCE >= safeTop) {
+    const top = Math.min(rect.top, viewport.height - 8)
+    return build(top, top)
+  }
+  // Table jammed under / above the fixed toolbar (top-of-doc or scrolled off the top): anchor to
+  // the toolbar strip so `placement: 'top'` overflows the viewport top and flips to BOTTOM, landing
+  // the controls just below the fixed toolbar's bottom edge — clickable, never behind it.
+  return build(0, safeTop)
 }
 
 /**
  * Reference rect for the floating table toolbar (#625). Anchors to the OUTER edge of the table
  * (its scroll wrapper when present) rather than the caret's cell, so the toolbar — placed above
  * with `placement: 'top'` — floats in the block margin above the whole table instead of landing on
- * top of the text rows above the caret. The rect is clamped into the visual viewport (see
- * {@link clampToolbarAnchorRect}) so a long table scrolled past the top of the viewport keeps the
- * controls reachable instead of pushing them off-screen. Returns a Floating-UI virtual element
- * (computed lazily so scrolling/resizing stays accurate), or null when the caret is not inside a
- * table, in which case the BubbleMenu falls back to its default cursor-based positioning.
+ * top of the text rows above the caret. The rect is clamped relative to the fixed editor toolbar
+ * (see {@link clampToolbarAnchorRect}) so a long table scrolled past the top of the viewport keeps
+ * the controls reachable and a table at the top of the doc keeps them clear of the fixed
+ * `.octo-toolbar`. Returns a Floating-UI virtual element (computed lazily so scrolling/resizing
+ * stays accurate), or null when the caret is not inside a table, in which case the BubbleMenu falls
+ * back to its default cursor-based positioning.
  */
 function tableReferenceElement(editor: Editor): { getBoundingClientRect: () => DOMRect; getClientRects: () => DOMRect[] } | null {
   const { $from } = editor.state.selection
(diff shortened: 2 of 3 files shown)


OctoBoooot (dismissed):
## Review: fix(docs): keep table toolbar clear of text and label delete controls (#716)

**Verdict:** Approve. A clean, well-scoped UI fix — toolbar anchoring + destructive-action labeling. No blockers, no majors.

Base `main`. Risk tier: normal. `mergeStateStatus=CLEAN`. Pure presentation change — no data/async/security surface.

### What's good
- **#625 toolbar anchoring** — `tableReferenceElement` anchors the floating toolbar to the table's outer edge (prefers the `.tableWrapper` scroll box, falls back to the table DOM) instead of the caret's cell, so with `placement:'top'` it floats in the block margin above the whole table rather than covering the in-table text rows above the caret. The virtual element is computed lazily (accurate on scroll/resize), and returns null → BubbleMenu falls back to caret positioning (the documented "never happens while shouldShow is satisfied" path). Sound.
- **#621-2 destructive labeling** — the `text=` caption is applied to exactly the three destructive buttons (`deleteRow`/`deleteColumn`/`deleteTable`, `:185/:202/:209`) while the additive add-row/add-column buttons stay icon-only. Correct scoping: the user reads what a delete removes, additive actions stay compact.

### Verified
- i18n keys `table.deleteRow/deleteColumn/deleteTable` present in both en-US and zh-CN (reused existing `title` keys), so no missing-key risk and the visible caption matches the `aria-label`/`title` (screen-reader/visual consistency).
- No XSS — captions are i18n text rendered as JSX children.
- Non-labeled buttons unchanged (`octo-tb-btn--labeled` modifier only added when `text` is set).

### Net
Focused, correct presentation fix with proper null-fallback on the anchor, destructive-only labeling, and complete i18n. Approving.

Jerry-Xin (dismissed):
Byte-verified at head `30314042` — clean, well-reasoned fix for both issues. **APPROVE.**

## ✅ #625 toolbar positioning — genuinely wired (verified not a no-op)
`tableReferenceElement()` walks the selection depth for a `table` node, prefers the ProseMirror `.tableWrapper` scroll box (so a horizontally-scrolled wide table still anchors to the visible region), and returns a Floating-UI `VirtualElement` (`getBoundingClientRect` + `getClientRects`), or `null` → BubbleMenu falls back to default caret positioning. I verified the load-bearing question — **is `getReferencedVirtualElement` a real prop?** — against the pinned `@tiptap/extension-bubble-menu@3.22.2` types: `BubbleMenuPluginProps.getReferencedVirtualElement?: () => VirtualElement | null` (index.d.ts:66), threaded into `BubbleMenuViewProps` and the `BubbleMenuView` constructor. So the prop is a supported API, the returned shape matches Floating-UI's `VirtualElement` contract, and the fix actually re-anchors the toolbar to the table's outer edge (not declared-but-ignored). Correct.

## ✅ #621-2 delete captions — a11y/clarity, no i18n risk
`TbBtn` gains an optional `text` caption rendered as `<span className="octo-tb-btn-label">` only when set; the three destructive controls (deleteRow/Column/Table) pass it so the user reads which target the action removes instead of guessing from an icon. Importantly the captions reuse the **existing** i18n keys `docs.table.deleteRow/deleteColumn/deleteTable` (already used by the `title=` on the same buttons pre-PR, present in both en-US.json and zh-CN.json) — no new keys, all via `t()`, **no hardcoded CJK** → no i18n:check exposure (unlike #700). The button keeps `title`/`aria-label`, so screen-reader semantics are unchanged and the visible caption is additive.

## ✅ CSS scoped correctly
`.octo-tb-btn` → `inline-flex; align-items:center` lays out icon+caption inline; the `gap:4px` lives on `.octo-tb-btn--labeled` (only applied when `text` is set), so **icon-only buttons are unaffected** (comment says as much and the class gating confirms it).

## ✅ CI
Build / i18n:check show `skipped` — this is normal path-filtering for a docs-package-only change (identical to the merged-clean #715); `reviewDecision=APPROVED`, `mergeStateStatus=CLEAN`, mergeable. Not a #700-style red.

Minimal, well-commented, both fixes correctly scoped. No blockers, no majors.

yujiawei (dismissed):
# Code Review — PR #716 (octo-web)

Reviewed at head SHA `30314042cc61cde39ac422e0714314c6263e29b8`. Scope: `packages/docs/src/editor/TableControls.tsx` and `styles.css`.

## Spec compliance

Both linked issues are addressed, and nothing outside their scope was added.

- **#625 (toolbar covers text):** `tableReferenceElement()` re-anchors the floating toolbar to the table's outer edge (its `.tableWrapper` scroll box when present) via the BubbleMenu `getReferencedVirtualElement` prop, so with `placement: 'top'` it floats in the margin above the whole table instead of over the caret's row. Correct.
- **#621-2 (delete controls unclear):** the three delete controls (`deleteRow` / `deleteColumn` / `deleteTable`) now render a visible caption next to the icon, reusing the existing i18n strings. Add controls stay icon-only. Correct.

No missing work, no over-building, no deviation.

## Verified

- `getReferencedVirtualElement` is a real prop on `@tiptap/extension-bubble-menu@3.22.2` (present in the shipped type defs and runtime), so this is not a silent no-op.
- The null return path is accurate: when the helper finds no table, the BubbleMenu's `virtualElement` getter falls back to the caret/selection rect. The "never happens while shouldShow is satisfied" comment matches `shouldShowTableBubble` requiring `isActive('table')`.
- Only `placement` and `offset` are overridden in `options`; the default `flip`/`shift` middleware stay active, so if there isn't room above the table the toolbar flips below rather than clipping. Good.
- The anchor is re-resolved on every `updatePosition` (the getter calls the helper fresh), so a horizontally scrolled or resized table stays correctly anchored.
- CSS: `display: inline-flex` is scoped under `.octo-table-bubble .octo-tb-btn`, and the gap/label rules only apply via `--labeled` / `-label`. Icon-only buttons elsewhere are unaffected.
- Accessibility: the button keeps `aria-label={title}` while showing the same string as visible text. Per the accessible-name computation, `aria-label` overrides descendant text content, so the caption is not announced twice. The PR description's claim here is correct.

## Findings (all advisory — none block merge)

**[P2] Detached-anchor flash on delete.** `tableReferenceElement` returns a virtual element closing over the resolved `anchor` node. On `deleteTable()`, if Floating-UI recomputes position in the frame between the table being removed and the menu unmounting, `getBoundingClientRect()` on the detached node returns an all-zero rect, which can briefly teleport the toolbar toward the viewport corner. It's re-resolved on the next tick and `shouldShow` drops the menu, so this is at most a one-frame flicker. Worth a manual check that delete-table looks clean; not a blocker.

**[P2] Cross-realm `instanceof HTMLElement`.** The `dom instanceof HTMLElement` guard is false if the editor ever runs inside an iframe/detached document (different `Window`), silently falling back to caret anchoring. The docs editor isn't rendered in an iframe today, so this is future-proofing only. If you want to harden it: `dom && dom.nodeType === 1`.

**[P2] Test coverage gap for the new paths.** Neither new behavior has a test. The caption rendering is a plain DOM assertion that jsdom handles fine (e.g. assert `.octo-tb-btn--labeled .octo-tb-btn-label` text on the delete buttons) — cheap to add and would lock in #621-2. The anchor positioning is fairly untestable under jsdom (no layout), so skipping that assertion is reasonable; a note in the test file would help.

**[P2] Inline `getReferencedVirtualElement` closure.** A fresh `() => tableReferenceElement(editor)` is passed each render. Positioning correctness is unaffected (the plugin calls it on demand), but if it defeats memoization in the React wrapper it could add redundant position recalcs during rapid typing. Only worth touching if you observe jank.

## Verdict

Clean, well-scoped, well-commented fix. Both issues are correctly resolved with no scope creep, and the load-bearing tiptap API usage checks out. The four points above are advisory polish, not blockers.

Note: this review is from an independent human reviewer verifying the tiptap API and the DOM/CSS behavior; a second automated reviewer leg was unavailable this run due to an unrelated tooling outage, so coverage of runtime positioning under real layout rests on the author's manual verification plus the reasoning above.

Jerry-Xin (dismissed):
Summary: The PR is in scope for `octo-web` and the table toolbar changes are generally sound, with a couple of edge-case notes.

💬 Non-blocking

- 🟡 Warning — `packages/docs/src/editor/TableControls.tsx:142`: `clampToolbarAnchorRect` forces the anchor to at least `64px` from the viewport top. For a long table whose visible area starts at the top of the viewport, the toolbar remains reachable but can still overlap visible table rows near the top. That may be an acceptable tradeoff, but it weakens the “never overlaps in-table text” claim and deserves a browser check.

- 🔵 Suggestion — `packages/docs/src/editor/TableControls.tsx:168`: the actual `tableReferenceElement()` path is not directly covered by tests; only the pure clamp helper is. A focused test around collapsed selections and `CellSelection` would make this safer, especially because the helper depends on ProseMirror resolved positions and `nodeDOM()` behavior.

✅ Highlights

- Reuses existing i18n keys for visible delete labels, avoiding new locale-resource churn.
- Keeps pointer-event passthrough for the table bubble portal intact.
- The helper computes the DOM rect lazily, which is the right shape for scroll/resize-sensitive positioning.

Verification note: I attempted `pnpm --filter @octo/docs test -- TableControls.test.tsx`, but this checkout has no `node_modules`, so `vitest` was unavailable.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #716 (octo-web)
Head SHA: 30314042cc61cde39ac422e0714314c6263e29b8 (local) / b144d50f0a690a76b7a5b55917ed78fdf445efe7 (GitHub)

## Summary

Small, focused UI/UX fix for the Octo Docs table editor. Two changes in `packages/docs/src/editor/TableControls.tsx` and `packages/docs/src/editor/styles.css`:

1. **Floating toolbar positioning (#625):** Adds a `tableReferenceElement()` helper that anchors the BubbleMenu to the table's outer edge (preferring the `.tableWrapper` scroll container) instead of the caret's cell. This prevents the toolbar from covering text rows above the table when placed with `placement: 'top'`.

2. **Delete control labels (#621-2):** Adds a visible `text` caption to the three destructive buttons (delete row, delete column, delete table) so users can read which entity the action removes, rather than guessing from an icon alone.

The implementation is clean, well-typed, and consistent with the tiptap BubbleMenu API contract. No blocking issues found.

## Verification

Static analysis only at head `30314042`; build and tests not executed in this environment.

- ✅ **API contract** — `getReferencedVirtualElement` is a documented optional prop in `@tiptap/extension-bubble-menu@3.22.2` (type `() => VirtualElement | null`), and the React `BubbleMenu` wrapper from `@tiptap/react/menus` correctly destructures and passes it through to the plugin. Verified via installed package types (`index.d.ts:66`).
- ✅ **VirtualElement shape** — `tableReferenceElement()` returns `{ getBoundingClientRect: () => DOMRect; getClientRects: () => DOMRect[] } | null`, which satisfies the Floating-UI `VirtualElement` interface. The `null` return falls back to the BubbleMenu's default cursor-based positioning.
- ✅ **Null safety** — `editor.view.nodeDOM()` returning non-HTMLElement is handled (function returns null). `dom.closest('.tableWrapper')` returning null falls back to the table element itself. All branches produce either a valid virtual element or null.
- ✅ **Lazy rect computation** — `getBoundingClientRect` is called lazily inside the returned closures, so Floating-UI gets fresh rects on each position update. No stale-closure risk.
- ✅ **`.tableWrapper` preference** — The helper correctly prefers the scroll container's bounding rect over the raw `<table>` element, ensuring horizontally-scrolled wide tables still anchor to the visible region.
- ✅ **TbBtn backward compat** — `text` is optional (`text?: string`); buttons without `text` render identically to before. Only the three delete controls pass `text`.
- ✅ **CSS isolation** — New rules scoped to `.octo-table-bubble .octo-tb-btn` (inline-flex for icon+label), `.octo-tb-btn--labeled` (gap only when labeled), and `.octo-tb-btn-label` (12px, nowrap). No global selector leakage.
- ✅ **i18n reuse** — `text` and `title` use the same i18n key (`t('docs.table.deleteRow')` etc.), already defined and used for tooltips pre-PR. No new keys introduced.
- ✅ **Existing tests unbroken** — `TableControls.test.tsx` asserts portal presence and button count under `.octo-table-bubble-portal`; the new `getReferencedVirtualElement` prop does not alter the rendered DOM structure, only positioning. jsdom's zero-rect layout is handled by the existing test shim.

## Findings

No P0/P1 issues. No P2 issues. The implementation is correct, well-scoped, and safe.

## Data Flow Trace

|---|---|---|
| `editor.state.selection.$from` | ProseMirror selection state | ✅ Always current when `getReferencedVirtualElement` is called by BubbleMenuView |
| `$from.node(depth).type.name === 'table'` | ProseMirror document tree walk | ✅ Matches the `table` node type registered by `@tiptap/extension-table` |
| `editor.view.nodeDOM($from.before(depth))` | ProseMirror view→DOM mapping | ✅ Returns the DOM element for the table node at the resolved position |
| `dom.closest('.tableWrapper')` | DOM ancestor search | ✅ Returns `.tableWrapper` if present, null otherwise; falls back to table element |
| `anchor.getBoundingClientRect()` | Browser layout engine | ✅ Called lazily on each position update; always current |
| `t('docs.table.deleteRow')` / `deleteColumn` / `deleteTable` | i18n locale files | ✅ Existing keys, pre-PR usage in `title` props; reused for `text` |
| `text ? ' octo-tb-btn--labeled' : ''` | TbBtn prop | ✅ Optional prop; only set on 3 delete buttons |

## Blindspot Checklist (R5)

- **C1 — Dual-path parity:** N/A. No symmetric add/remove or create/delete paths changed. Delete buttons only gain a label; the underlying `editor.chain().deleteRow/deleteColumn/deleteTable()` calls are untouched.
- **C2 — Control-flow ordering / nesting:** Clear. `tableReferenceElement` is called once per position update, not nested or re-entered. No double-application risk.
- **C3 — Authorization boundary:** N/A. No auth/permission changes.
- **C4 — Authorization lifecycle:** N/A. No auth changes.
- **C5 — Build/runtime divergence:** N/A. No build config, packaging, or environment changes. CSS is standard, no build-time transforms affected.
- **C6 — Governance/doc self-consistency:** N/A. No governance or policy changes.

## Verdict: APPROVED

Clean, well-scoped UI polish. The `tableReferenceElement()` helper correctly uses the tiptap BubbleMenu's `getReferencedVirtualElement` API to anchor the floating toolbar above the table rather than the caret cell. The delete control labels reuse existing i18n keys and add CSS without side effects. No correctness, safety, or contract issues.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; clean implementation with proper null handling, correct tiptap API usage, and consistent CSS scoping.

OctoBoooot (dismissed):
## Review: #716 @ `b144d50f` — DELTA re-review; viewport-clamp refinement, clean APPROVE

**Verdict:** Approve. Clean single-commit fast-forward (ahead 1 / behind 0 — no rebase noise) on top of my prior approved head. This commit adds the #625 off-screen-toolbar clamp as a pure, unit-tested enhancement; no behavior removed, no regression to the #621-2 delete-label work. CI green, `mergeStateStatus=CLEAN`.

### What changed (byte-verified)
New `clampToolbarAnchorRect(rect, viewport)` (`TableControls.tsx`), wired into `tableReferenceElement` via a lazily-evaluated `band()`. For a long table scrolled so its top leaves the viewport, Floating-UI's default `flip` was re-anchoring to the off-screen bottom edge and pushing the toolbar below the viewport (~1051 vs a 1000px viewport). The fix collapses the reference to a zero-height band whose `top` is clamped into `[TOOLBAR_SAFE_TOP=64, viewport.height-8]` and whose horizontal extent is clamped into the viewport — so both `top` and any flipped `bottom` resolve just inside the viewport.

Verified the clamp math against all four new test assertions:
- fully-visible `{top:300}` → stays at real top 300 (no TC-P0-001 regression), height 0, left/right preserved;
- scrolled-past-top `{top:-1248}` → clamped to 64;
- below-viewport `{top:5000}` → clamped to `height-8` (992);
- wide/off-screen `{left:-500,right:3000}` → `[0, 1200]`.
`right = min(width, max(rect.right, left))` correctly prevents a `right<left` inversion (degenerate fully-off-screen cases collapse to a zero-width band, no negative width). `band()` calls `getBoundingClientRect()` lazily, so scroll/resize accuracy is preserved.

### Minor (non-blocking)
- `TableControls.tsx:142` — `TOOLBAR_SAFE_TOP=64` is a by-design tradeoff: a table whose real top sits within the top 64px of the viewport gets its toolbar floated near those rows rather than above them. Reasonable to keep the controls reachable, and matches the documented intent.

### Nit (≤3)
- Only the pure `clampToolbarAnchorRect` is unit-tested; the DOM `band()` wiring path in `tableReferenceElement` isn't exercised. Fine given the logic is isolated in the pure function, but a single integration assertion would close the loop.

### Note
PR is currently marked **draft** — it can't merge in that state regardless; re-review was explicitly requested on the new commit, so approving the code as-is.

### Net
Pure additive viewport-clamp fix, math matches its tests, no regression. Approving. Credit ploy-elison.

yujiawei (approved):
# Code Review — PR #716 (octo-web)

Reviewed at head `b144d50f0a690a76b7a5b55917ed78fdf445efe7` against `main` (merge-base `ede8a03`). Scope: `TableControls.tsx`, `TableControls.test.tsx`, `styles.css` (+153 / −2).

## 1. Spec compliance

**Spec: ✅**

- **#625 — toolbar covers text.** The toolbar is re-anchored to the table's outer edge (`.tableWrapper` scroll box when present, else the table node DOM) via the BubbleMenu `getReferencedVirtualElement` prop, and the anchor rect is clamped into the viewport so a long, scrolled-past-top table keeps the controls reachable. Implemented in `tableReferenceElement` / `clampToolbarAnchorRect` (`TableControls.tsx:133-190`). Matches the issue's "position controls outside the text flow / never cover readable text" ask.
- **#621-2 — unclear delete controls.** The three destructive controls now render a text caption next to the icon, reusing the existing `docs.table.deleteRow/deleteColumn/deleteTable` strings (en-US + zh-CN both present). See `TbBtn` `text` prop and the three call sites (`TableControls.tsx:79-106, 239-267`).
- **No over-build:** the diff is confined to the two source files + one test file. No new flags, schema, props beyond the optional `text`, or unrelated behavior. `TableCellView` and the `#595 C1` pointer-events passthrough are untouched.
- **No deviation:** implementation approach matches the PR description.

## 2. Code Quality

**Quality: Approved**

Verification performed independently (not taken from the PR description):
- `vitest src/editor/TableControls.test.tsx` → **14/14 pass** (10 prior + 4 new), including the `#595 C1` column-resize hot-zone case.
- `tsc --noEmit` (docs typecheck) → only **2 pre-existing errors** in unrelated files (`dmworkbase/src/i18n/format.ts`, `docs/src/members/sort.ts`), both present on `main` and untouched by this diff. No new type errors introduced.
- Confirmed `getReferencedVirtualElement` is a real prop on `@tiptap/react` 3.22.2's `BubbleMenu`, and that the underlying plugin **falls back to caret positioning when the function returns `null`** (it guards `if (referencedVirtualElement) return it;` before the default path) — so the `null` branch in `tableReferenceElement` is safe, as claimed.
- `clampToolbarAnchorRect` math traced for the degenerate cases (`right < left`, table entirely below viewport, wide/off-screen horizontally): output stays within `[0, viewport.width]` × `[TOOLBAR_SAFE_TOP, viewport.height-8]` with a non-negative width. Correct.
- Accessibility: the visible caption is a plain `<span>` while the button keeps its `aria-label`/`title`; since an `aria-label` supplies the accessible name, the caption text is not additionally announced — no double-announcement.

No P0 / P1 findings.

### P2 — non-blocking suggestions

- **P2 — clamp boundary is the window, but the editor scrolls inside a container.** `tableReferenceElement` clamps against `window.innerWidth/innerHeight` (`TableControls.tsx:177-181`), but the prose actually scrolls inside `.octo-doc-scroll` (`overflow-y: auto`) which sits below a `position: sticky; top: 0` formatting toolbar. In the fix's target case (long table scrolled past the top) the band is pinned to `TOOLBAR_SAFE_TOP = 64` in *window* coordinates, which can place it near/under the sticky toolbar rather than the true top of the scroll viewport. The controls remain reachable, so this is a refinement, not a blocker — consider clamping to the scroll container's rect (or accounting for the sticky toolbar height) if this looks cramped in practice.
- **P2 — vertical clamp ignores `rect.bottom`.** The band's `top` is `max(TOOLBAR_SAFE_TOP, min(rect.top, viewport.height-8))` and never consults `rect.bottom` (`TableControls.tsx:142`). When a tall table is scrolled so its *bottom* edge is near the top of the viewport (most of the table already off-screen above), the band stays pinned at `64` while the table has effectively scrolled away, so the toolbar can float detached below the table's real position, over the following content. Edge case, low frequency; bounding `top` by `rect.bottom` would tighten it.
- **P2 (nit) — coverage.** The new tests exercise the pure `clampToolbarAnchorRect` thoroughly, but `tableReferenceElement` (the DOM/`nodeDOM`/`closest('.tableWrapper')` integration) has no direct test. Understandable given it needs real layout; noting it as a coverage gap, not a request.

## 3. Overall verdict

**APPROVE** — spec ✅ and quality Approved. All findings are P2 refinements that do not block merge.

## 4. Coverage note (what this review could not fully verify from static analysis)

- Real-browser visual confirmation that the toolbar no longer overlaps text across caret rows, and that the scrolled-past-top case now renders the controls on-screen (verified by unit math + the `#625` test, not by a live render).
- Behavior when the editor is embedded in a constrained/split-pane/modal layout with its own overflow (the window-based clamp above would be most visible there).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #716 (octo-web)

## Summary

This PR addresses two docs-editor UX issues: #625 (table toolbar unreachable when a long table is scrolled past the viewport top) and #621-2 (destructive delete controls need visible text labels so the user knows which row/column/table the action removes). The fix introduces a pure `clampToolbarAnchorRect` helper and a `tableReferenceElement` virtual-element factory that re-anchors the BubbleMenu to a zero-height band clamped inside the visual viewport, plus a `text` prop on `TbBtn` that renders visible captions on the three delete buttons.

The implementation is clean, well-documented, and correctly scoped. No blocking issues found.

## Verification

- ✅ **Pure-function correctness** — `clampToolbarAnchorRect` in `packages/docs/src/editor/TableControls.tsx:108` has four unit tests covering: fully-visible table (no regression), long table scrolled past top (the #625 fix), table below viewport, and wide/off-screen table horizontal clamping. All edge cases produce correct values.
- ✅ **API contract** — `getReferencedVirtualElement` at `packages/docs/src/editor/TableControls.tsx:224` is a valid tiptap v3 BubbleMenu prop (tiptap 3.22.2 confirmed in `packages/docs/package.json`). The returned object `{ getBoundingClientRect, getClientRects } | null` matches the Floating UI virtual element contract.
- ✅ **CSS scope** — `.octo-tb-btn` gets `display: inline-flex` in `packages/docs/src/editor/styles.css:401`, which is a safe baseline change for flex-row button layout. `.octo-tb-btn--labeled` and `.octo-tb-btn-label` are additive and do not collide with the existing `.octo-table-bubble-portal` pointer-events model.
- ✅ **i18n coverage** — `docs.table.deleteRow` ("Delete row" / "删除行"), `docs.table.deleteColumn` ("Delete column" / "删除列"), `docs.table.deleteTable` ("Delete table" / "删除表格") all exist in both `packages/docs/src/i18n/en-US.json:175-179` and `packages/docs/src/i18n/zh-CN.json:175-179`.
- ✅ **No regression on `shouldShowTableBubble`** — The guard at `packages/docs/src/editor/TableControls.tsx:199` (`editor.isActive('table')`) still controls visibility. `tableReferenceElement` walking the same ProseMirror tree is consistent — when `shouldShow` is satisfied, the tree walk will always find the table node.

Static analysis only at head `b144d50f0a690a76b7a5b55917ed78fdf445efe7`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One minor nit below.

### Nit — `text` and `title` use identical i18n key on delete buttons (`packages/docs/src/editor/TableControls.tsx:241-242`)

The three delete buttons pass the same i18n key to both `title` (browser tooltip) and `text` (inline caption), e.g. `title={t('docs.table.deleteRow')}` and `text={t('docs.table.deleteRow')}`. The visible label and the hover tooltip will always show the same string. This is functional and self-consistent, but if future designs want a shorter label (e.g. "Row") with a fuller tooltip (e.g. "Delete this row"), separate i18n keys would be needed. Not a merge concern — flagging for awareness.

## Data Flow Tracing

| Consumer | Source | Verified |
|---|---|---|
| `clampToolbarAnchorRect(rect, viewport)` | `anchor.getBoundingClientRect()` via `.tableWrapper` or table DOM; `window.innerWidth`/`innerHeight` | ✅ `getBoundingClientRect()` is called inside a closure (`band()` at `:178`), so it captures the latest scroll position on every Floating UI recompute. `.tableWrapper` is defined in `styles.css:1847`. |
| `tableReferenceElement` return value | `editor.state.selection.$from` → ProseMirror tree walk → `editor.view.nodeDOM()` → virtual element | ✅ Standard ProseMirror `nodeDOM(pos)` API. Null fallback is safe — BubbleMenu falls back to cursor-based positioning when the virtual element is null. |
| `TbBtn` `text` prop | `t('docs.table.deleteRow/Column/Table')` | ✅ Keys exist in both locale files. `TbBtn` renders `<span className="octo-tb-btn-label">{text}</span>` only when `text` is truthy, so icon-only add-row/column buttons are unaffected. |
| CSS `.octo-tb-btn` `inline-flex` | Scoped under `.octo-table-bubble` | ✅ Does not affect `.octo-table-bubble-portal` pointer-events model. Icon-only buttons render correctly with `inline-flex` + `align-items: center`. |

- **C1 (Dual-path parity)**: N/A — no paired add/remove or subscribe/unsubscribe paths changed. Add-row and delete-row use independent tiptap commands; the PR only adds visual labels to delete controls.
- **C2 (Control-flow ordering / nesting)**: Clear — `tableReferenceElement` is only called from `getReferencedVirtualElement` on the table BubbleMenu. No other BubbleMenu or call-site consumes it. No nested/double-invocation risk.
- **C3 (Authorization boundary)**: N/A — pure frontend UI change, no auth/permission surface.
- **C4 (Authorization lifecycle)**: N/A — no auth or container/member state.
- **C5 (Build/runtime path)**: Clear — the `getReferencedVirtualElement` prop is a tiptap v3 API feature; tiptap 3.22.2 is the declared dependency. CSS classes are additive. No build config or packaging changes.
- **C6 (Governance/docs self-consistency)**: N/A — no governance or policy docs changed.

## Verdict: APPROVED

Well-scoped frontend fix for two real UX issues. The viewport-clamping logic is pure and well-tested, the virtual element factory correctly integrates with tiptap's BubbleMenu API, and the delete-button labels are minimal and properly internationalized. No correctness, security, or regression concerns.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Pure-function logic verified with unit tests; data flow from ProseMirror selection → virtual element → Floating UI positioning traced end-to-end; CSS and i18n additions are scoped and consistent with existing conventions.

yujiawei (dismissed):
# Code Review — PR #716 (octo-web)

Reviewed at head SHA `b144d50f0a690a76b7a5b55917ed78fdf445efe7` against merge-base `ede8a038`.
Scope: `TableControls.tsx`, `TableControls.test.tsx`, `styles.css` (153 / -2).

**Spec: ✅**

The diff maps cleanly onto the two linked issues, and nothing extra was added.

- **#625 — toolbar covers text.** `tableReferenceElement()` re-anchors the floating toolbar to the table's outer edge (`.tableWrapper` when present, else the table DOM node) via `getReferencedVirtualElement`, so with `placement: 'top'` it floats in the block margin above the whole table rather than over the caret's row. Correct fix for the reported behavior.
- **#625 — off-screen (second commit).** `clampToolbarAnchorRect()` additionally handles the long-scrolled-table case: it clamps the anchor into the visual viewport and collapses it to a zero-height band. This is beyond the literal issue text but is the same defect class (toolbar not reachable), so it belongs here rather than being scope creep.
- **#621 sub-issue 2 — unclear delete controls.** The three delete controls now render a visible caption via the new optional `text` prop, reusing the existing `docs.table.deleteRow/deleteColumn/deleteTable` i18n keys (present in both en-US and zh-CN on `main`).
- **Correctly out of scope:** #621 sub-issue 1 (column-resize) is untouched, and the `.octo-table-bubble-portal` pointer-events passthrough that keeps the resize hot zone reachable (#595 C1) is preserved. No unrequested features, flags, or schema changes.

## 2. Code quality

**Quality: Approved**

The tiptap integration and the clamp math were independently verified against the tiptap v3.22.2 source and by hand-tracing the tests. No blocking issues found.

### Verified correct (points worth recording)

- **`getReferencedVirtualElement` is a valid API here.** The import is `@tiptap/react/menus` (the v3 Floating-UI-based BubbleMenu, not the legacy tippy one). In `@tiptap/react@3.22.2`, `BubbleMenuProps` forwards both `getReferencedVirtualElement` and `options` into `BubbleMenuPluginProps`; the props are not silently dropped.
- **Returning `null` does not crash.** In the v3.22.2 bubble-menu plugin, the `virtualElement` getter does `const referencedVirtualElement = this.getReferencedVirtualElement?.(); if (referencedVirtualElement) { return referencedVirtualElement }` and otherwise falls through to the caret-based `posToDOMRect` rect. A falsy return is the documented fallback path, so the "not in a table" branch is safe.
- **Clamp math** matches all four unit assertions (fully-visible table keeps its real top; scrolled-past-top clamps to 64; below-viewport clamps to `height - 8`; wide table clamps horizontally to `[0, width]`).
- **No double screen-reader announcement.** Because `aria-label` (from `title`) is set, it becomes the button's accessible name and overrides the visible child `<span>` text, so the caption is not announced twice. The PR body's claim holds.

### Suggestions (non-blocking — nits)

- **P2 — clamp uses `rect.top` only, not `rect.bottom`.** For a table scrolled almost entirely above the fold (`top ≈ -500`, `bottom ≈ 30`), the band is floored to `TOOLBAR_SAFE_TOP` (64), a few px below the table's real bottom, so the toolbar can appear to hover just under the vanishing table. This is a cosmetic edge case, not a correctness bug — and note that the naive "also `Math.min(top, rect.bottom)`" fix would push the band back below the fold and re-introduce the very off-screen bug this commit fixes. If tightened later, it should clamp toward `rect.bottom` only while `rect.bottom` is itself inside the viewport. Leaving it floored-at-top is the safer default; fine to ship as-is.
- **P2 — `visualViewport` vs `window.innerHeight`.** The clamp reads `window.innerWidth/innerHeight`. Under pinch-zoom or the mobile keyboard, the layout viewport and visual viewport diverge, so the "safe" band can still be partly obscured. Docs editing is desktop-first, so acceptable; `window.visualViewport` would be more precise if this ever needs to be exact.
- **Nit — base `.octo-tb-btn` gains `display: inline-flex`.** `.octo-table-bubble .octo-tb-btn` now applies `display: inline-flex; align-items: center` to every button in the table bubble, not only labeled ones. The buttons are icon/`<span>` wrappers so no shift is expected, but it is broader than the "icon-only buttons are unaffected" comment implies; scoping the flex rule to `.octo-tb-btn--labeled` would match the stated intent. Cosmetic only.

### Coverage note

- Unit tests were verified by static trace; the full `vitest`/`tsc` run could not be executed here (no installed `node_modules` in the review worktree). The PR body reports `TableControls.test.tsx` 10/10 and typecheck clean, consistent with the diff.
- Runtime visual behavior (actual Floating-UI flip/shift on a real scrolled long table, and the CSS change on legacy buttons) was reasoned about, not rendered — worth a quick manual smoke on a 30+ row table if not already done.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The two P2s and the CSS-scoping nit are improvements, not merge blockers.

OctoBoooot (approved):
## Review: #716 @ `ac121ea4` — DELTA re-review; toolbar-clearance rework, clean APPROVE

**Verdict:** Approve. Clean single-commit fast-forward (ahead 1 / behind 0) on my prior approved head; the prior APPROVE was auto-dismissed by this commit, re-affirming at the new head. This commit reworks the #625 clamp to also keep the floating table toolbar clear of the *fixed editor toolbar* (not just the viewport) and adds visible captions to the delete controls (#621-2). Risk tier: normal. Build green; `code-review` pending. No longer draft.

### What changed (byte-verified)
`clampToolbarAnchorRect` now takes a third `toolbarBottom` arg and has two branches:
- **Table ≥`TOOLBAR_CLEARANCE`(48px) below the fixed toolbar** → zero-height band at the real top edge (floats above as before — the fully-visible case is untouched, no TC-P0-001 regression).
- **Table jammed under the fixed toolbar / scrolled off the top** → band `[0 .. safeTop]` so `placement:'top'` overflows the viewport top, Floating-UI flips the controls to BOTTOM, landing them just below the fixed toolbar — reachable, never behind it.

Verified the math against all six new test assertions:
- fully-visible `{top:300}`, tb=64 → `252≥64` → band at 300 ✓
- scrolled-past-top `{top:-1248}` → Branch B → `[0,64]` (flips below) ✓
- jammed under tall toolbar `{top:185}`, tb=180 → `137<180` → `[0,180]` ✓
- below-viewport `{top:5000}` → `min(5000,992)=992` (height-8) ✓
- wide `{left:-500,right:3000}` → `[0,1200]` ✓
- no-toolbar default (tb=0) → band at real top ✓

`right = min(width, max(rect.right, left))` still prevents a `right<left` inversion. `toolbarBottom` is measured **live** (`getBoundingClientRect().bottom`, so a taller find-bar toolbar is accounted for) and **scoped to this editor's shell** (`.octo-doc--editor`), so an unrelated toolbar elsewhere on the page is never picked up — good defensive scoping.

### Also good
- Delete controls (#621-2) now render a text caption via `t('docs.table.deleteRow'/'deleteColumn'/'deleteTable')` — existing i18n keys, no hardcoded strings, no i18n JSON touched (key parity preserved). Accessible label + visible text so the user reads which row/column/table is removed.
- Band closure lazily calls `getBoundingClientRect()`, so scroll/resize accuracy is preserved.

### Nit
- Still only the pure `clampToolbarAnchorRect` is unit-tested; the DOM `band()`/`toolbarBottom()` measurement path in `tableReferenceElement` isn't exercised. Fine given the logic is isolated in the pure function; a single integration assertion would close the loop.

### Net
Correct two-boundary clamp rework (viewport + fixed-toolbar), math matches all six tests, editor-scoped toolbar measurement, accessible delete labels via existing i18n keys, no regression. Approving. Credit ploy-elison.

Jerry-Xin (approved):
The PR is relevant to `Mininglamp-OSS/octo-web` and the implementation is scoped to the docs editor table toolbar. I found no blocking correctness, security, or architecture issues.

💬 Non-blocking

🟡 Warning — [packages/docs/src/editor/styles.css](/Users/mlamp/.openclaw/tmp/pr-review-rx_nq170/repo/packages/docs/src/editor/styles.css:403): The new caption styles add hardcoded spacing/font sizing (`gap: 4px`, `font-size: 12px`). Project guidance prefers design tokens for spacing and typography. This matches nearby legacy CSS, so I would not block on it, but it is worth aligning if suitable tokens exist.

🔵 Suggestion — [packages/docs/src/editor/TableControls.tsx](/Users/mlamp/.openclaw/tmp/pr-review-rx_nq170/repo/packages/docs/src/editor/TableControls.tsx:188): The pure clamp function has good coverage, but the DOM lookup path in `tableReferenceElement` is not directly tested. A small integration/unit test for selecting the `.tableWrapper` anchor and scoped toolbar measurement would reduce regression risk around the most important behavior.

✅ Highlights

The positioning change anchors to the table wrapper instead of the caret cell, which addresses the text-overlap problem without disturbing table command behavior.

The delete controls reuse existing i18n keys and keep existing `aria-label`/`title` behavior.

Focused verification note: I attempted `pnpm exec vitest run packages/docs/src/editor/TableControls.test.tsx`, but this checkout does not have `vitest` available to `pnpm exec` (`Command "vitest" not found`).

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. One nit (CSS comment inaccuracy) with zero user-visible impact. Geometry logic is sound, data flow from DOM measurement through Floating-UI is correct, test coverage is comprehensive.

yujiawei (approved):
# Code Review — PR #716 (octo-web)

Reviewed at head SHA `ac121ea462c082f8ea62d7d6ac818e312cc918c1` against merge-base `ede8a03`. Scope: `TableControls.tsx` (+129/-1), `TableControls.test.tsx` (+58/-1), `styles.css` (+14).

## 1. Scope / correctness

The diff cleanly implements the three linked items and nothing beyond them:

- **#625 (toolbar covers in-table text):** `tableReferenceElement()` anchors the floating toolbar to the table's outer edge (`.tableWrapper` scroll box, or the table DOM node) instead of the caret's cell, so with `placement: 'top'` it floats in the block margin above the whole table. Wired via the `getReferencedVirtualElement` prop.
- **#716 follow-up (toolbar hidden behind the fixed editor toolbar):** `clampToolbarAnchorRect()` collapses the anchor to a zero-height band at the table's real top when the table sits clear below the fixed toolbar, or to the toolbar strip `[0 .. toolbarBottom]` otherwise, so `placement: 'top'` overflows the viewport top and Floating-UI's `flip` drops the controls to just below the fixed toolbar.
- **#621-2 (delete controls unclear):** `TbBtn` gains an optional `text` caption; the three delete controls pass the existing i18n strings. Icon-only buttons are unaffected.

Verified against the actual `@tiptap/react`/`@tiptap/extension-bubble-menu@3.22.2` internals:

- `getReferencedVirtualElement` is a real prop; when it returns a truthy value the plugin uses it directly, and returning `null` falls back to the default caret/selection rect — so the "not in a table" fallback path is correct.
- The virtual-element contract (`getBoundingClientRect` + `getClientRects`) matches what the plugin consumes.
- Passing `options={{ placement: 'top', offset: 8 }}` **merges over** the plugin defaults (`flip: {}`, `shift: {}` stay enabled), so the flip-below mechanism the fix depends on is actually active. This is the load-bearing assumption of the whole `clampToolbarAnchorRect` design, and it holds.

Boundary math in `clampToolbarAnchorRect` traces correctly across every case I could construct: fully-visible mid-viewport table (zero-height band at real top), table scrolled past the top (`rect.top` large-negative → strip band → flip below), table jammed under a tall toolbar, no fixed toolbar (`toolbarBottom = 0`), and a table sitting within the 48px clearance zone. In the last case the controls flip to just below the fixed toolbar and can graze the table's very top edge — that is the deliberate, documented tradeoff (better than rendering behind the opaque fixed toolbar), not a defect. The horizontal clamp keeps the band inside `[0, viewport.width]` with non-negative width.

I18n strings (`docs.table.deleteRow/deleteColumn/deleteTable`) and the CSS classes the helper queries (`.octo-toolbar`, `.octo-toolbar-wrap`, `.octo-doc--editor`, `.tableWrapper`) all exist in the tree. The `toolbarBottom` lookup is scoped to this editor's `.octo-doc--editor` shell, so an unrelated toolbar elsewhere on the page is not picked up.

## 2. Verification

- `vitest src/editor/TableControls.test.tsx`: **16/16 pass**, including the 6 new `clampToolbarAnchorRect` cases and the pre-existing #595 C1 pointer-events case.
- `tsc --noEmit` (docs typecheck): **no errors in the changed files**. The only errors reported are in unrelated files (`dmworkbase/src/Service/APIClient.ts`, `dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) and are pre-existing on the base branch, as the PR description states.
- `.octo-table-bubble-portal` pointer-events passthrough (the #595 C1 column-resize hot-zone fix) is untouched; the new anchoring changes only *where* the wrapper is positioned, not its pointer-events behavior, so no regression there.

## 3. Non-blocking suggestions (nits)

1. **Micro-optimization (perf):** `tableReferenceElement`'s `toolbarBottom()` closure runs `querySelector('.octo-toolbar')` + `.closest()` on every `getBoundingClientRect()` call. This is bounded — the bubble-menu plugin positions via debounced `computePosition` on scroll/resize (~60ms debounce) and on selection/doc changes, **not** a per-frame `autoUpdate` loop — so the real-world cost is small. Still, caching the resolved toolbar element when the virtual element is created (the `shell` reference already is) would trim a couple of DOM lookups per reposition. Optional.
2. **Embedded-container edge case:** the viewport clamp uses `window.innerWidth/innerHeight`. For an iframe embed this is correct (window == iframe viewport), but if the docs editor is ever mounted inside a same-window modal/drawer smaller than the window, the horizontal clamp bound would be the full window rather than the container. Not a concern for the primary full-viewport docs editor; worth a comment if embedded rendering is on the roadmap.
3. **Accessibility (verified non-issue):** the visible `<span class="octo-tb-btn-label">` duplicates the `aria-label`, which can look like a double announcement, but per the ARIA accessible-name algorithm `aria-label` on the `<button>` overrides name-from-contents, so the label is announced once. The PR's reasoning is correct. If you want belt-and-suspenders, `aria-hidden="true"` on the caption span makes the intent explicit, but it is not required.

## 4. Verdict

Scope is exact, the core mechanism is verified against the real library internals, tests and typecheck are green, and no P0/P1 issues were found. The three items above are optional improvements, not merge blockers.

PR #59, opened 2026-05-19
feat(mention): UI input + render for @所有人 / @所有AI (PR-C, closes #58)

yujiawei:
Closes #58. PR-C of the mention three-state refactor — web UI layer only.

## Scope

Two surfaces:

1. **MessageInput** — `@`-dropdown gets two sticky-top items above the member list.
2. **Conversation** — render matrix highlights `@所有人` / `@所有AI` based on `mention.humans` / `mention.ais` / legacy `mention.all`.

## Input behavior

| Item                 | Sentinel uid | Inserted text | Payload effect            |
|----------------------|--------------|---------------|---------------------------|
| `@所有人` (new)      | `-2`         | `@所有人 `    | `mention.humans = 1`      |
| `@所有AI` (new)      | `-3`         | `@所有AI `    | `mention.ais = 1`         |
| `@所有人` (legacy)   | `-1`         | `@所有人 `    | `mention.all = 1` (kept)  |

- The two new items are pinned at the top of the dropdown regardless of search-query filter.
- The legacy `uid: -1` path is only reachable now via voice transcription (`parseMentionMarkers`) and restored drafts — both still emit `all=1` per RFC.
- `MentionModel` gains `humans?: number` / `ais?: number` (0|1 wire shape).

## Render matrix

| Flags                              | Highlight              |
|------------------------------------|------------------------|
| `humans = 1`                       | `@所有人`              |
| `ais = 1`                          | `@所有AI`              |
| `humans = 1` + `ais = 1`           | `@所有人` + `@所有AI`  |
| `all = 1` (legacy / server outbound double-write) | `@所有人` |

Implementation: `getMessageMentions` synthesizes virtual `MentionInfo` entries with `uid: 'all'`, reusing the existing `mention-highlight` CSS class so colors / styles match. Does **not** mutate `message.parts`.

## Wire format

`buildTextContent` serializes `humans` / `ais` into:
- `msgContent.mention` (via `(mn as any).humans = …`) — for in-process consumers.
- `contentObj.mention.{humans,ais}` + the existing `encode()` override — for the SDK wire JSON.

Same pattern already used for `entities`. Server (PR-A, octo-server #94) accepts both fields and double-writes them on outbound (`all=1` ⇒ `humans=1`).

## Tests

`apps/web/src/__tests__/mentionRefactor.test.ts` — 6 new cases (the 4 required matrix bullets + 2 regression guards):

- `humans=1 only → highlights @所有人`
- `ais=1 only → highlights @所有AI`
[...]

PR #722, opened 2026-07-14
feat(docs): line-height & paragraph spacing (SCHEMA_VERSION 17)

Jerry-Xin:
## Summary

Adds line spacing / paragraph spacing to the docs editor (**SCHEMA_VERSION 17**), the long-form typography gap called out in the competitive review. Modelled byte-for-byte on the existing `textAlign` global attribute.

Three global attributes are added to `heading` + `paragraph` (not new nodes/marks — same class as v5 `textAlign` / v7 `fontSize`), all defaulting to `null` so old documents round-trip losslessly:

| attr | value | DOM |
|---|---|---|
| `lineHeight` | unitless multiplier (`1` / `1.15` / `1.5` / `2` / custom) | `line-height: <n>` |
| `spaceBefore` | length `px` \| `em` | `margin-top: <len>` |
| `spaceAfter` | length `px` \| `em` | `margin-bottom: <len>` |

## Changes

- **`editor/LineHeight.ts`** — self-built global-attribute extension (Tiptap ships no block-level line-height; `@tiptap/extension-text-style`'s LineHeight rides on the `textStyle` *mark*, wrong semantics for block spacing). `renderHTML`/`parseHTML` + `set*/unset*` commands. Values sanitised at **both** parse and render (unitless-multiplier / `px|em` whitelist; out-of-range or injection attempts fall back to `null`).
- **`editor/extensions.ts`** — register `LineHeight` right after `TextAlign` in the live + preview extension sets.
- **`editor/Toolbar.tsx`** — line-spacing selector (presets `1.0 / 1.15 / 1.5 / 2.0` + custom multiplier). Labels via i18n (`en-US` / `zh-CN`), no hardcoded strings.
- **`export/docx/*`** — `mapSpacing()` maps the attrs onto `w:spacing` (`line` in 240ths + `lineRule: AUTO`, `before`/`after` in twips); wired into paragraph + heading conversion.
- **`schema/index.ts`** — `SCHEMA_VERSION` → **17**. v16 is intentionally skipped (reserved for #59).
- Tests: extension round-trip + canonical style ordering + sanitisation, docx `mapSpacing`, schema/version audit.

## Canonical style serialization (backend byte-alignment)

The one hard cross-repo contract. Tiptap's `mergeAttributes` folds every `renderHTML` `style` fragment into a single declaration as `<prop>: <val>` joined by `"; "` (colon+space, semicolon+space, **no** trailing `;`), matching the official `TextAlign` fragment. With `TextAlign` registered before `LineHeight`, a fully-attributed block serialises to exactly:

```
text-align: <a>; line-height: <lh>; margin-top: <mt>; margin-bottom: <mb>
```

The backend `getBlockAttrs`/`setBlockAttrs` `toDOM` must emit byte-identical fragments (same order, `"; "` separator, colon+space, **unitless** line-height). Round-trip through the Y.Doc is order-independent (`parseHTML` reads each property individually); only the serialised HTML needs this ordering. A test locks the order.

## Status / gating

- Version **v17** is PM-issued and pinned; **hold merge until #59 (v16) lands on `main`** to avoid a version collision.
- Companion backend PR (schema stub + `getBlockAttrs/setBlockAttrs` + Typst `renderTypst.ts`) lands in `octo-docs-backend`; the docx export in this PR + backend Typst close out the export requirement (not deferred).

## Verification

- `tsc` typecheck: no new errors from these files.
- `vitest`: new/updated suites pass (extension round-trip, style-order contract, sanitisation, docx `mapSpacing`, schema audit).
- `i18n:check`: green (locale keys healthy).

Refs SCHEMA_VERSION v17 · XIN-964.


OctoBoooot (dismissed):
## Review: #722 @ `e839b07f` — line-height & paragraph spacing (SCHEMA_VERSION 17)

**Verdict:** Approve — with a merge-ordering hold (below). A clean, thorough, injection-safe critical-tier schema PR: three global attrs (`lineHeight`, `spaceBefore`, `spaceAfter`) on heading/paragraph, modelled on the v5 `textAlign` pattern. Risk tier: **critical** (SCHEMA_VERSION 17 schema change) — `needs-human-review` applied; requires maintainer sign-off. CI: `code-review` pending. PR is draft.

### Verified good
- **No CSS-injection surface (notable given the #700 saga).** `LineHeight.ts` does the opposite of #700's raw-string bug: `parseHTML` reads the specific CSSOM longhand (`element.style.lineHeight`/`.marginTop`/`.marginBottom` — already browser-decoded for HTML entities *and* CSS escapes) then whitelist-validates the **value** (`^\d+(\.\d+)?$` capped 0<lh≤10; `^\d+(\.\d+)?(px|em)$`). No property-name string matching, so the CSS-escape bypass class can't apply, and the value regexes make `;`/`}`/`url()` injection impossible. Both parse and render re-sanitize.
- **Round-trip is symmetric.** Every attr that renders has a matching parse-back longhand; defaults are `null` → no style emitted → old docs untouched (tested). The "canonical style ordering" backend contract is correctly documented as irrelevant to Y.Doc round-trip (parseHTML reads longhands individually) — it only matters for byte-identical HTML import against the backend `toDOM`.
- **Attr-only, correctly.** Added via `addGlobalAttributes` on existing nodes; correctly absent from `SCHEMA_NODES`/`SCHEMA_MARKS` and asserted OUT of both. Extension registered after `TextAlign` (verified in the array, not just the comment).
- **DOCX export correct + tested.** Unitless line-height → Word 240ths + `LineRuleType.AUTO` (1.5→360, 1.15→276); px|em → twips (15/px, 16px/em: 12px→180, 1em→240). Combined-attrs case covered.
- **i18n complete (values, not just keys):** `Line spacing`/`Default`/`Custom` and `行距`/`默认`/`自定义` in both locales.

### Blocking on process, not code — SCHEMA_VERSION 16/17 merge ordering
The PR bumps 15→**17**, deliberately skipping v16 (reserved for the in-flight fontFamily/CJK work) with a documented "hold-merge until v16 is on main" gate. That gate is currently **unmet**: #700 (which owns v16) is open and BLOCKED (an unresolved CSS-escape paste-gate CR). If #722 merges before #700 lands, main jumps to a v17 with no v16 having existed.

I verified this is **not a code bug** — nothing in the frontend consumes `SCHEMA_VERSION` with a contiguity/`version < N`/migration gate (it's only exported + asserted ==17; the per-saved-version `schemaVersion` wire field is unrelated). So the skip is safe at the code level. But the PR's own stated contract is to hold until v16 is on main, and it isn't. Treat this as a **merge-ordering hold** for the maintainer: land #700 (v16) first, then this. Not a request to change code.

### Minor
- `Toolbar.tsx` custom line-height `<input min="0.5">` disagrees with the sanitizer's lower bound (`>0`): `0.1`–`0.49` is sanitizer-valid but below the input's `min`, and a rejected/partial entry can momentarily snap the controlled input back. Cosmetic UX only — no invalid value can be written (sanitize returns false, attr unchanged). Align the input `min`/`step` with the sanitizer range if you want the two to agree.

### Net
Correct, injection-safe, round-trip-clean, well-tested across editor/schema/DOCX/i18n. Approving the code. The one gate is process: honor the PR's own hold-merge-until-v16 note — #700 (v16) must land first, and it's currently blocked. Plus the standing critical-tier maintainer sign-off (`needs-human-review`). `mergeStateStatus` stays BLOCKED until both clear.

Jerry-Xin (commented):
The PR is relevant to `octo-web` and the core schema/editor/export changes look directionally sound; I found no blocking issues.

**💬 Non-blocking**
- 🟡 Warning: DOCX export sanitization is weaker than editor sanitization. `sanitizeLineHeight()` rejects values over `10` in `packages/docs/src/editor/LineHeight.ts:48`, but `mapSpacing()` accepts any positive numeric string in `packages/docs/src/export/docx/styles.ts:250`. Similarly, spacing lengths are unbounded in `packages/docs/src/export/docx/styles.ts:227`. Consider reusing shared sanitizers or adding equivalent bounds so malformed persisted attrs cannot produce extreme or invalid DOCX spacing.
- 🟡 Warning: The toolbar only exposes `lineHeight`, not `spaceBefore` / `spaceAfter`, even though the schema and export add paragraph spacing attrs. See `packages/docs/src/editor/Toolbar.tsx:531`. If paragraph spacing is meant to be user-editable in this PR, controls are missing.
- 🔵 Suggestion: The new `.octo-line-height*` toolbar controls do not appear to have matching CSS, while `.octo-font-size` is explicitly styled in `packages/docs/src/editor/styles.css:2403`. Adding widths and shared toolbar select/input styling would avoid native-control layout drift.

**✅ Highlights**
- The attrs default to `null` and render omission is covered by tests.
- The canonical style order contract is explicitly tested.
- Live and preview extension sets are both updated.
- DOCX mapping is covered by focused tests.

**Back-compat / security (critical-tier, verified):**
- Schema 16→17: `lineHeight`/`spaceBefore`/`spaceAfter` all default to `null`; renderHTML omits style when null → old docs round-trip losslessly, no crash.
- Paste path reads the CSSOM property accessor (`element.style.lineHeight` / `.marginTop` / `.marginBottom`), which normalizes CSS escapes during parse — structurally immune to the raw-string CSS-escape bypass class.
- Note: the focused Vitest suite and package typecheck could not be run in this checkout (no installed `node_modules`), so `vitest`/`tsc` were unavailable.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #722 (octo-web)

## Summary

This PR adds line-height and paragraph-spacing support to the docs editor (SCHEMA_VERSION 17). A self-built Tiptap `LineHeight` extension introduces three global attributes (`lineHeight`, `spaceBefore`, `spaceAfter`) on `heading` and `paragraph` nodes, mirroring the existing `textAlign` pattern. The feature includes a toolbar dropdown with presets and custom input, docx export mapping to Word's spacing model, and thorough test coverage for sanitisation, round-trip, and byte-alignment contracts. The schema version intentionally skips v16 (reserved for in-flight CJK work).

Overall the implementation is well-structured with strong sanitisation and test coverage. Two non-blocking P2 items concern missing CSS for the toolbar controls and a gap in the markdown export path.

## Verification

- ✅ **Sanitiser boundary** — `sanitizeLineHeight` rejects units, out-of-range (0, negative, >10), non-string, and CSS injection attempts (`1;color:red`, `expression(1)`). `sanitizeSpacing` rejects non-px/em units, bare numbers, and injection. Both tested.
- ✅ **Extension registration order** — `LineHeight` is registered after `TextAlign` in both `buildExtensions` (`packages/docs/src/editor/extensions.ts:129`) and `buildPreviewExtensions` (`:241`), preserving the canonical style property order for byte-alignment.
- ✅ **Schema version** — Bump from 15→17 with v16 reserved is clearly documented in `packages/docs/src/schema/index.ts:51-60`. Tests at `packages/docs/src/schema/index.test.ts:18` assert the new version.
- ✅ **Default null round-trip** — All three attrs default to `null`; `renderHTML` emits no style for null values, so old documents are unaffected.
- ✅ **Docx export** — `mapSpacing` correctly maps unitless line-height to Word's 240ths-of-a-line with `LineRuleType.AUTO`, and px/em spacing to twips (15 twips/px, 16px/em). Tested in `spacing.test.ts`.
- ✅ **i18n parity** — Both `en-US.json` and `zh-CN.json` add matching `lineHeight`/`lineHeightDefault`/`lineHeightCustom` keys.

Static analysis only at head `e839b07fcc10f57db4652f18d6cfa4ac672bf06a`; build and tests not executed in this environment.

## Findings

### P2 — Markdown export drops v17 line-spacing attrs (`packages/docs/src/editor/LineHeight.ts:33` → `packages/docs/src/export/markdown.ts:201`)

The `wrapAlign()` helper in the markdown serialiser handles `textAlign` via an HTML `align` attribute but does not emit any `style` attribute for `lineHeight`/`spaceBefore`/`spaceAfter`. Paragraphs and headings exported to markdown silently lose their line-spacing formatting; on re-import the spacing attrs are gone, so the round-trip is lossy for any document using v17 line spacing. The docx export path was updated (`packages/docs/src/export/docx/nodes.ts:174,186`) but the markdown path was not.

Diff-scope: new — v17 attrs are introduced by this PR; the existing `wrapAlign` was not extended to cover them.

```ts
// Fix sketch for markdown.ts wrapAlign:
function wrapAlign(inner: string, node: MdNode): string {
  const align = node.attrs?.textAlign
  const style = buildSpacingStyle(node.attrs) // sanitise + escape
  const parts: string[] = []
  if (typeof align === 'string' && align && align !== 'left')
    parts.push(`align="${escapeHtmlAttr(align)}"`)
  if (style) parts.push(`style="${escapeHtmlAttr(style)}"`)
  if (!parts.length) return inner
  const tag = node.type === 'heading' ? 'div' : 'p'
  return `<${tag} ${parts.join(' ')}>${inner}</${tag}>`
}
```

### P2 — Toolbar line-height controls lack CSS (`packages/docs/src/editor/Toolbar.tsx:548`)

The `octo-line-height-control`, `octo-line-height`, and `octo-line-height-custom` CSS classes used by the new `LineHeightSelect` component have no matching rules anywhere in the codebase. The `<select>` and `<input type="number">` will render with browser-native defaults, visually inconsistent with the sibling `.octo-font-size` control (`packages/docs/src/editor/styles.css:2403`) which has explicit height/border/radius/padding rules.

Diff-scope: new — the classes are introduced by this PR without accompanying styles.

Fix: add rules mirroring `.octo-font-size` for the new classes in the 'Batch-3 schema toolbar items' CSS section of `editor/styles.css`.

## Data Flow Tracing

|---|---|---|
| Toolbar `current` value | `editor.getAttributes('paragraph').lineHeight` → ProseMirror node attrs | ✅ Set via `setLineHeight` (sanitised) or `parseHTML` (sanitised); defaults to null |
| `renderHTML` style string | `attributes.lineHeight` → `sanitizeLineHeight()` → safe value or null | ✅ Double sanitise (parse + render) prevents injection |
| Docx `mapSpacing` | `node.attrs.lineHeight` from MdNode JSON → regex-validated in `mapSpacing` | ✅ Values always sanitised upstream; `mapSpacing` adds own regex guard (range not capped at 10, but unreachable via editor path) |
| `LineHeightSelect.onChange` | `<select>` value (preset string) or `<input>` value (custom) → `editor.chain().focus().setLineHeight(v)` → command sanitises | ✅ Invalid values rejected at command level (`return false`) |

- **C1 (dual-path parity)**: Clear. set/unset command pairs are symmetric — set sanitises, unset calls `resetAttributes`. Verified all 6 commands (set/unset × lineHeight/spaceBefore/spaceAfter).
- **C2 (control-flow ordering)**: Clear. `LineHeight` registered in exactly 2 locations (live editor, preview editor), both after `TextAlign`. No nested reuse.
- **C3 (auth boundary)**: N/A — client-side editor extension, no auth surface.
- **C4 (auth lifecycle)**: N/A.
- **C5 (build/note)**: N/A — no build/packaging changes.
- **C6 (governance/docs)**: Clear. Schema version history is well-documented; v16 skip rationale is stated. No policy conflicts.

## Verdict: APPROVED

No P0/P1 blockers. The two P2 items (missing CSS and markdown export gap) are non-blocking polish items that don't affect correctness or data integrity in the primary editor path. The core feature — sanitisation, schema registration, docx export, and test coverage — is solid.

yujiawei (changes requested):
# Code Review — PR #722 (octo-web)

Reviewed at head `ad5d2d4` (the PR advanced past the originally-requested `e839b07`; the follow-up commit added a `<=1000` magnitude cap to `sanitizeSpacing` and made the docx `spacingToTwips` reuse it, so the earlier spacing-blowout and docx-corruption concerns are now addressed on both the parse/render and export paths).

Overall this is a clean, well-tested feature that faithfully mirrors the existing `TextAlign` global-attribute pattern. Two-sided sanitisation, `null` defaults for lossless round-trip, and the canonical style-order test are all solid. One functional defect blocks merge, plus a few smaller items.

## 1. Spec compliance

**Result: PASS (with one scope note)**

- The declared line-spacing selector (presets + custom multiplier), the three global attrs (`lineHeight` / `spaceBefore` / `spaceAfter`) on `heading` + `paragraph`, the docx `mapSpacing` export, i18n keys, and the `SCHEMA_VERSION` → 17 bump are all present and implemented as described.
- No out-of-scope additions or contradictory implementations spotted.
- **Scope note (not blocking):** the schema, commands (`setSpaceBefore`/`setSpaceAfter`), docx export, and i18n fully wire up **paragraph spacing** (`spaceBefore`/`spaceAfter`), but the toolbar only exposes a **line-height** control — there is no UI to set space-before/after. The attrs still round-trip (from paste / backend), so this is defensible as deferred UI, but if a spacing control was expected in this PR it is missing. Please confirm this is intentional.

## 2. Code quality

**Result: CHANGES REQUESTED**

### P1 — Custom line-height input steals focus on every keystroke (breaks multi-character values)

`packages/docs/src/editor/Toolbar.tsx:578-582`

```tsx
onChange={(e) => {
  const v = e.target.value.trim()
  if (!v) editor.chain().focus().unsetLineHeight().run()
  else editor.chain().focus().setLineHeight(v).run()
}}
```

`.focus()` runs inside the `onChange` of the custom `<input type="number">`. On the first keystroke it pulls DOM focus back to the ProseMirror editor, so the input blurs and the user cannot type a multi-character multiplier like `1.15`, `1.25`, or `2.5` — only single interactions register. The spinner buttons hit the same wall. The advertised "custom multiplier" field is effectively unusable; only the presets in the adjacent `<select>` work.

The correct pattern already exists in this same file: the link field (`Toolbar.tsx` ~900-950) holds a local `useState` value and commits on Enter/blur instead of calling `.focus()` per keystroke. Suggested fix: drop the `.focus()` from the number input's `onChange` (commit the value without refocusing), or mirror the link-field pattern and commit on blur / Enter.

### P2 — No CSS for the new toolbar controls

`packages/docs/src/editor/Toolbar.tsx:548,571` (classes `octo-line-height`, `octo-line-height-custom`, `octo-line-height-control`)

None of these classes exist in `packages/docs/src/editor/styles.css` (confirmed: `grep octo-line-height styles.css` → 0 matches). By contrast the sibling `FontSizeSelect` has a styled `.octo-font-size` rule (height, border, radius, padding). Without a rule the new select + number input render with raw browser defaults and will look inconsistent with the rest of the toolbar (mismatched height/border). Add the equivalent styling.

### P2 — `mapSpacing` line-height branch is not bounded, unlike the parity work just done for spacing

`packages/docs/src/export/docx/styles.ts:255-262`

The follow-up commit made `spaceBefore`/`spaceAfter` reuse `sanitizeSpacing` (whitelist + `<=1000` cap) in the docx path, so a non-frontend attr source (backend / collab update) can no longer push out-of-range margins into the export. But the `lineHeight` branch in `mapSpacing` still uses its own inline regex with only an `n > 0` check and **no upper bound** — it does not reuse `sanitizeLineHeight` (which caps at 10). Same threat model you just closed for spacing: a `lineHeight: "999999"` arriving from outside the sanitised editor path yields `line: 239760000` in the OOXML. For symmetry and defense-in-depth, reuse `sanitizeLineHeight` here too (reject, don't clamp), matching the spacing path.

## 3. Coverage notes / blind spots

- **Backend byte-alignment is unverifiable from this repo.** The "byte-for-byte with the backend `getBlockAttrs`/`setBlockAttrs` toDOM" contract (order `text-align; line-height; margin-top; margin-bottom`, `"; "` separator, unitless line-height) depends on the companion backend change landing in lockstep. The FE side has a test locking the full-4-attr order, which is good; please make sure the backend PR carries an equivalent assertion, and note that the order test only covers the all-attrs-present case — partial subsets (e.g. only `lineHeight` + `margin-top`) are not explicitly locked (they remain correct by load-order, but untested).
- **`SCHEMA_VERSION` skip to 17 (reserving v16 for #59).** The skip itself is harmless (versions only need to be monotonic and unique). The real constraint is the merge-ordering the PR calls out ("hold until #59's v16 lands"): worth confirming that gate is enforced before this merges, since it is a human sequencing step, not something the code guarantees.
- Not assessed: runtime visual rendering of the toolbar control, and validation of the generated `.docx` against a real Word/OOXML consumer beyond the unit-level `mapSpacing` shape assertions.

## Summary

Solid, well-tested implementation that follows the established `TextAlign` pattern. Blocking on the P1 focus-steal that makes the custom multiplier input non-functional; the two P2s (missing CSS, unbounded line-height in the docx export) and the paragraph-spacing-UI scope question should be resolved or consciously deferred before merge.

OctoBoooot (changes requested):
## Review: #722 @ `ad5d2d46` — DELTA re-review; flipping to REQUEST_CHANGES on a real focus-steal bug + a remaining lineHeight export asymmetry (credit yujiawei)

My prior APPROVE was at `e839b07f` (auto-dismissed by this commit). The new commit correctly closes the **spacing** sanitize-symmetry gap (docx `spacingToTwips` now reuses `sanitizeSpacing` with the ≤1000 cap — good, credit). But yujiawei's re-review surfaces two real issues I confirmed at this head; flipping to REQUEST_CHANGES. Risk tier: critical (SCHEMA_VERSION 17); `needs-human-review` remains; still draft.

### Major — custom line-height input steals focus every keystroke (can't type multi-char values)
`Toolbar.tsx:576-582`: the custom `<input type="number" value={current}>` calls `editor.chain().focus().setLineHeight(v).run()` on **every** `onChange`. Two compounding effects:
- `.focus()` on each keystroke returns focus to the editor, so the input loses focus mid-typing — a user trying to enter `1.15` gets bumped out after the first character.
- It's a controlled input bound to `current`; typing a partial value (`1.`) runs it through `setLineHeight` → `sanitizeLineHeight` rejects the partial → `current` is unchanged → the field snaps back. Net: free-form custom multipliers are effectively untypeable.
Byte-verified the wiring. This is a real usability bug (no data corruption). Fix: don't `.focus()` the editor on each keystroke — apply on blur/Enter, or debounce and only re-focus after a committed valid value (the same commit-vs-live-preview shape used elsewhere).

### Minor — docx lineHeight export still has no upper bound (sanitize asymmetry, half-closed)
This commit routed docx **spacing** through `sanitizeSpacing` (capped ≤1000) — but the **lineHeight** path was left as-is: `mapSpacing` (`export/docx/styles.ts:255-262`) validates lineHeight with only `n > 0` and a bare `/^\d+(?:\.\d+)?$/`, and `styles.ts` does **not** import `sanitizeLineHeight` (which caps ≤10). So a persisted out-of-range `lineHeight` (e.g. `"9999"` from a hand-crafted doc or the backend) exports as `line = round(9999*240) ≈ 2.4M` — unbounded. The editor path caps it; the export path doesn't. Same defense-in-depth gap you just fixed for spacing — reuse `sanitizeLineHeight` in `mapSpacing` for the lineHeight branch too so all three (parse/render, docx, backend) agree on ≤10. Minor (the editor input can't produce >10; needs an out-of-band attr), but it's the exact asymmetry this commit set out to close.

### yujiawei's 🟡 (agreed, non-blocking)
- New `.octo-line-height*` control has no matching CSS (also raised earlier).
- Toolbar exposes only lineHeight, not spaceBefore/spaceAfter — confirm whether the paragraph-spacing UI is intentionally deferred (the attrs + schema + export exist, just no control).

### Still verified good (carried)
CSSOM-longhand parse + value whitelist (no #700 raw-string bug); attr-only; extension after TextAlign; DOCX twip math; i18n values; spacing ≤1000 cap now symmetric.

### Net
The spacing-cap fix is correct, but the custom line-height input is effectively untypeable (focus steal on every keystroke) and the docx lineHeight export still lacks the ≤10 cap its spacing sibling now has. REQUEST_CHANGES on the focus-steal major; the export asymmetry is a cheap same-file follow-on. Plus the standing v16-first (#700) merge-ordering hold and critical-tier maintainer sign-off. Credit yujiawei.

yujiawei:
### Follow-up finding (head `ad5d2d4`)

One additional P2 surfaced after the initial review, worth folding in with the other items:

**P2 — Markdown export drops the v17 line-spacing attrs** (`packages/docs/src/export/markdown.ts:201`)

`wrapAlign()` emits an HTML wrapper for `textAlign` but does not carry `lineHeight` / `spaceBefore` / `spaceAfter`. The docx path was updated for v17 (`export/docx/nodes.ts:174,186`) but the markdown path was not, so a document using line spacing loses that formatting on markdown export and it is gone on re-import (lossy round-trip). Consider extending `wrapAlign` to emit a sanitised `style="..."` on the wrapper (reuse the same `sanitizeLineHeight` / `sanitizeSpacing` whitelist, escaped via `escapeHtmlAttr`), mirroring the docx export.

This is non-blocking on its own, but the P1 (custom line-height input focus-steal) still stands as the merge blocker.

Jerry-Xin (commented):
Re-review at `ad5d2d46` (was `e839b07`). The new commit `ad5d2d46` bounds `spaceBefore`/`spaceAfter` at `<=1000` across FE `sanitizeSpacing` + docx export, closing the spacing symmetry gap from my prior note. Reviewing @yujiawei's four findings against the current head.

**🔴 Blocking**

- **(1) Custom line-height input steals focus on every keystroke — REAL, blocking.** In `packages/docs/src/editor/Toolbar.tsx` the custom `<input type="number">` (the `octo-line-height-custom` field) runs `editor.chain().focus().setLineHeight(v).run()` inside `onChange`. The `.focus()` moves DOM focus to the ProseMirror document on the *first* keystroke, blurring the input mid-typing. Because the field is controlled (`value={current}`, sourced from `editor.getAttributes(...)`) and `sanitizeLineHeight` rejects intermediate states — e.g. `"1."` fails `/^\d+(\.\d+)?$/`, so `setLineHeight("1.")` returns false and the attr never updates — the user cannot type a multi-char value like `1.5`: focus leaves the input and the controlled value snaps back. This breaks the primary use of the custom field. Note the sibling `octo-font-size` control tolerates the same `focus()` pattern only because it is a `<select>` (one discrete change), not a free-text field. Fix: drop the `.focus()` from the per-keystroke path (commit on blur / Enter, or set attrs without stealing focus), and/or debounce + only apply on a fully-valid value so intermediate keystrokes aren't discarded.

**💬 Non-blocking**

- **(2) Export-side lineHeight has no upper bound — REAL, symmetry gap (severity 🟡).** `mapSpacing` in `packages/docs/src/export/docx/styles.ts` gates the line-height branch with `Number.isFinite(n) && n > 0` only — no `<= 10` ceiling, unlike the editor's `sanitizeLineHeight` (`0 < lh <= 10`). This commit fixed the *spacing* (`spaceBefore`/`spaceAfter`) branch by reusing `sanitizeSpacing`, but left the sibling *lineHeight* branch — the one flagged here — asymmetric; the commit message's "parity across paths" claim therefore holds for spacing but not for line-height. Reachability is limited: the editor render/parse path and `setLineHeight` cap at `<=10`, so a value beyond the cap only reaches export via a non-editor attr source (imported/backend doc). That keeps it a defense-in-depth symmetry gap rather than a live corruption path, but it is cheap to close for full parity: reuse `sanitizeLineHeight` (or apply the same `<=10` bound) in the `mapSpacing` line-height branch, mirroring what was just done for spacing.

- **(3) New toolbar control lacks matching CSS — confirmed.** `octo-line-height-control`, `octo-line-height`, and `octo-line-height-custom` (Toolbar.tsx) have no rules in `packages/docs/src/editor/styles.css`, while the sibling `.octo-font-size` does. The control will render unstyled. Add the matching rules.

- **(4) Toolbar exposes only line-height, no spaceBefore/spaceAfter UI — confirmed.** The `setSpaceBefore`/`setSpaceAfter` commands and the schema attrs exist, but there is no toolbar control for them. Reads as intentional MVP scoping — please confirm; if deferred, a follow-up note would help.

**Merge ordering:** unchanged — v17 still layers over the v16 slot; keep the hold-until-v16-lands sequencing.

**✅ Verified clean:** `LineHeight.ts` still uses CSSOM longhand reads + value whitelist (no raw-string CSS-escape regression); `SCHEMA_VERSION` = 17 with the attrs-on-existing-nodes model (no new nodes/marks); i18n values are real (`行距` / `Line spacing`); DOCX twip math and the spacing `<=1000` reject-not-clamp bound are correct and now byte-aligned FE↔export↔backend for spacing.

OctoBoooot (dismissed):
## Review: #722 @ `ef80faba` — DELTA re-review; focus-steal + export-cap both fixed, spacing UI added, flipping to APPROVE

**Verdict:** Approve. This commit fixes both items my CR held on and adds the paragraph-spacing UI (the peer 🟡). Byte-verified. Flipping REQUEST_CHANGES → APPROVE. Risk tier: critical (SCHEMA_VERSION 17); `needs-human-review` remains for the maintainer sign-off. Build + title-lint green; `code-review` pending. (v16-first ordering note below.)

### Resolved — custom line-height input focus-steal (was the major)
Refactored into `LineHeightCustomInput` with **local `draft` state**: `onChange` only `setDraft` (no editor call per keystroke), and commit runs **only on blur/Enter** (`onBlur={commit}`, `onKeyDown` Enter). The caret is never yanked back mid-type, and a partial `"1."` no longer snaps back — `1.15` is now typeable. `.focus()` runs once inside `commit`, not per keystroke. Exactly the fix prescribed.

### Resolved — docx lineHeight export cap (was the minor asymmetry)
`mapSpacing` now does `const lh = sanitizeLineHeight(attrs.lineHeight)` (`styles.ts:255`) — the lineHeight branch goes through the same `sanitizeLineHeight` (0<lh≤10) as the spacing branch (`sanitizeSpacing`, ≤1000). An out-of-range attr is rejected, not exported as `round(9999*240)`. The prior spacing-fixed-but-lineHeight-uncapped asymmetry is closed; the comment (`:257-258`) notes the symmetry.

### Added — paragraph spacing UI (was a peer 🟡)
New `spaceBefore`/`spaceAfter` toolbar controls, wired to `setSpaceBefore`/`setSpaceAfter` (which sanitize via `sanitizeSpacing`) with the **same commit-on-blur pattern** (`:570,602` "never per keystroke") — no per-keystroke focus-steal repeated, no new injection/unbounded surface. i18n complete both locales (`Space before`/`段前距`, `Space after`/`段后距`).

### Unchanged (verified good, carried)
CSSOM-longhand parse + value whitelist (no #700-class injection), attr-only schema, extension after TextAlign, twip math, i18n parity.

### Merge-ordering hold (unchanged, process not code)
Still SCHEMA_VERSION 17 skipping v16 (owned by #700). #700 is now review-clean (whole-bypass-class fixed, four-way APPROVE) but not yet merged (awaiting its own `needs-human-review`). Per the PR's own contract, land #700's v16 first. Maintainer call — not a code defect.

### Net
Both held items fixed (commit-on-blur input, docx lineHeight ≤10 cap), spacing UI added cleanly with the same safe pattern, i18n complete. Approving the code. Gated on: v16 (#700) landing first + the critical-tier `needs-human-review` sign-off. Credit yujiawei for the focus-steal + export-cap catches.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #722 (octo-web)

## Summary

This PR adds line-height and paragraph spacing (spaceBefore / spaceAfter) to the docs editor as three new global attributes on the `heading` and `paragraph` nodes (SCHEMA_VERSION 17). The implementation is a self-built Tiptap extension modelled on the existing `TextAlign` pattern, with defence-in-depth sanitisation at parse, render, command, and DOCX-export boundaries. The toolbar gains a line-height dropdown with a commit-on-blur custom-multiplier input, plus two paragraph-spacing dropdowns (before / after). The PR also bumps SCHEMA_VERSION from 15 to 17, deliberately skipping the reserved v16 (CJK, in-flight).

The code is well-structured, thoroughly tested, and the sanitiser / data-flow contracts are solid. One non-blocking P2 UX item noted below.

## Verification

- ✅ **Sanitiser parity** — `sanitizeLineHeight` (0 < lh ≤ 10, unitless only) and `sanitizeSpacing` (0 ≤ n ≤ 1000, px|em only) are called at parse, render, command, and DOCX-export time. The regexes (`LINE_HEIGHT_RE`, `SPACING_RE`) are anchored and non-backtracking — no ReDoS risk.
- ✅ **Backend byte-alignment contract** — canonical style property order (`text-align; line-height; margin-top; margin-bottom`) is enforced by extension-load order (LineHeight registered after TextAlign) and pinned by `LineHeight.test.ts` round-trip assertions.
- ✅ **Command set/unset parity (C1)** — all six commands (`setLineHeight`/`unsetLineHeight`, `setSpaceBefore`/`unsetSpaceBefore`, `setSpaceAfter`/`unsetSpaceAfter`) are symmetric. Set commands sanitise before writing; unset commands call `resetAttributes`.
- ✅ **Extension registration (C2)** — `LineHeight` registered after `TextAlign` in both `buildExtensions` and `buildPreviewExtensions`. Ordering is correct for the canonical style serialisation contract.
- ✅ **DOCX export wiring** — `mapSpacing` in `styles.ts` re-runs both sanitisers before converting to twips / 240ths-of-a-line. `convertParagraph` and `convertHeading` in `nodes.ts` spread the spacing object correctly.
- ✅ **Schema version bump** — SCHEMA_VERSION 15 → 17 (v16 reserved for CJK). v17 adds attributes only, no new nodes/marks — `SCHEMA_NODES` and `SCHEMA_MARKS` are unchanged, correctly asserted by `schema/index.test.ts`.
- ✅ **Commit-on-blur input design** — `LineHeightCustomInput` uses a local draft + `seenCommitted` ref to avoid the focus-steal and snap-back issues documented in the component's comments. Tests pin the typing, commit, revert, and seed behaviours.
- ✅ **i18n** — both `en-US.json` and `zh-CN.json` add the same six keys under `docs.toolbar.*`.

Static analysis only at head `ef80faba71ac0b9e82e6f1939be5d6a2049182ff`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 UX item below.

### P2 — Default option label reuses the control title (`packages/docs/src/editor/Toolbar.tsx:692`)

The `ParagraphSpacingSelect` reset option renders `t('docs.toolbar.spaceBefore')` ("Space before") or `t('docs.toolbar.spaceAfter')` ("Space after") as the dropdown value text. This is the same string used for the `<select>` `title` (tooltip). When a user opens the dropdown to clear spacing, the first option reads "Space before" — which looks like a label rather than a "reset to default" action. The sibling `LineHeightSelect` uses a dedicated `lineHeightDefault` key ("Default") for its reset option, which is clearer.

Consider reusing the existing `lineHeightDefault` key or adding a `spacingDefault` key so the reset option reads consistently:

```tsx
<option value="">{t('docs.toolbar.lineHeightDefault')}</option>
```

- Sanitiser reject (not clamp) semantics are consistent across parse/render/command/export and the backend schema contract.
- `mapSpacing` returns `undefined` when nothing applies, so the spread `...(spacing ? { spacing } : {})` correctly omits the property for blocks without spacing.
- `LineHeightCustomInput` render-phase `setDraft` sync is guarded by the `seenCommitted` ref — no infinite loop, and typing is not clobbered.
- The `'custom'` option in `ParagraphSpacingSelect` is a read-only sentinel (onChange returns early), so non-preset pasted values are preserved without being silently reset.
- Both live and preview extension sets include `LineHeight` — historical versions will render line-spacing faithfully.
- The `spacingToTwips` function in `styles.ts` re-runs `sanitizeSpacing` before parsing — redundant but defensive; the second regex match is guaranteed to succeed on the sanitised value.

## Data-flow Trace

- **lineHeight value**: user input → `setLineHeight(value)` command → `sanitizeLineHeight(value)` → `updateAttributes('heading'|'paragraph', { lineHeight: v })` → stored on node attr → `renderHTML` re-sanitises → inline `style="line-height: v"`. On parse: `parseHTML` reads `element.style.lineHeight` → `sanitizeLineHeight` → node attr. DOCX: `mapSpacing` → `sanitizeLineHeight` → `Math.round(Number(lh) * 240)` twips.
- **spaceBefore/spaceAfter value**: same defence-in-depth path as lineHeight, with `sanitizeSpacing` enforcing px|em units and ≤1000 cap. DOCX: `spacingToTwips` → `sanitizeSpacing` → twips conversion (15 twips/px, 16px/em).
- **Custom input draft**: typing → `setDraft` (local state only, no editor focus) → blur/Enter → `commit()` → `editor.chain().focus().setLineHeight(draft).run()` → on success: `seenCommitted` updated; on failure: draft reverts to committed.

## Blind-spot Checklist

- **C1 (dual-path parity)**: set/unset pairs symmetric for all three attrs. Live/preview extension sets both include LineHeight. ✅ Clear.
- **C2 (control-flow ordering)**: LineHeight registered after TextAlign in both extension sets. Commands sanitise then updateAttributes. ✅ Clear.
- **C3–C6**: N/A — no auth, lifecycle, build, or governance changes in this PR.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. The PR adds line-height and paragraph spacing with solid defence-in-depth sanitisation, correct schema versioning, proper DOCX export wiring, and thorough tests. One non-blocking P2 on the paragraph-spacing dropdown's default option label.

yujiawei (changes requested):
# Code Review — PR #722 (octo-web)

Reviewed at head SHA `ef80faba71ac0b9e82e6f1939be5d6a2049182ff` against `main` (merge-base `c548abc3`). The feature itself is cleanly implemented and well-tested (all 75 new/updated tests pass locally), but it cannot merge as-is: the branch is **CONFLICTING** with `main`, and its central version-gating premise is stale.

## Verdict: CHANGES_REQUESTED

## Spec compliance

The declared feature — three global attributes (`lineHeight`, `spaceBefore`, `spaceAfter`) on `heading` + `paragraph`, defaulting to `null`, modelled on `textAlign`; docx `mapSpacing`; toolbar controls; i18n; SCHEMA_VERSION bump — is fully implemented, with nothing missing and no scope creep. The two-sided sanitisation, the `<=1000` / `<=10` cap parity between FE sanitizers and the docx path, and the canonical `text-align; line-height; margin-top; margin-bottom` style ordering (TextAlign registered before LineHeight in both the live and preview extension sets) are all present and tested. The em→twips (16px/em × 15twips/px) and unitless→240ths math in `mapSpacing` check out (`1.5`→`360`, `12px`→`180`, `1em`→`240`).

## Blocking (P1)

### 1. The "v16 is reserved for #59 (CJK)" premise is false — branch is out of date with `main`

`packages/docs/src/schema/index.ts` (v16/v17 comment block, ~L51-69) and `schema/index.test.ts` (L11-12) state that **v16 is reserved for in-flight CJK work (#59), intentionally not defined, and that merge should be held until "#59's v16 lands on main."**

That is no longer true. `main` has **already landed v16**, and it is **`fontFamily`** (SCHEMA-SPEC §6, a `textStyle` mark attribute), not CJK:

- `main:packages/docs/src/schema/index.ts` L51-56 — `v16 — SCHEMA-SPEC §6: add a fontFamily ATTRIBUTE … export const SCHEMA_VERSION = 16`
- `main:packages/docs/src/editor/extensions.ts` — registers `LiveFontFamily` (a `FontFamily.extend(...)` with a paste-write gate behind `FONT_FAMILY_ENABLED`)
- `main:packages/docs/src/editor/extensions.test.ts` L91-111 — asserts the `v16 fontFamily` attr + round-trip

Consequences:
- The PR is in **CONFLICTING** state. A trial merge produces conflict markers in `schema/index.ts` (SCHEMA_VERSION `17` vs `16`), `schema/index.test.ts`, `extensions.test.ts`, and `Toolbar.test.tsx`.
- This branch's `extensions.ts` / `extensions.test.ts` have **no awareness of `fontFamily`** at all — they were branched before v16 landed.
- The "hold merge until #59's v16 lands" gating note in the PR description is moot/incorrect and should be dropped.

**Required:** rebase onto current `main` and resolve the conflicts. Keep `SCHEMA_VERSION = 17`, but rewrite the v16 comment to describe the **fontFamily** attribute that actually occupies v16 (the current text describing v16 as an undefined CJK reservation is factually wrong and will mislead future schema audits). When resolving the test conflicts, retain **both** the v16 `fontFamily` assertions (from `main`) and the new v17 line-spacing assertions.

Please re-verify after rebase that the merged `extensions.ts` still registers `TextAlign` before `LineHeight` in both extension sets (a post-merge check confirms it currently does, so the byte-alignment ordering survives — but it must be re-confirmed on the actual rebased branch, since the ordering is load-order-dependent and a merge can reshuffle it).

## Non-blocking (P2 / nits)

### 2. `LineHeightSelect` always renders a selectable "Custom" option that silently no-ops

`Toolbar.tsx` — `LineHeightSelect` unconditionally renders `<option value="custom">`, and `onChange` does nothing for `custom` (the comment says the number input drives custom values). If the current value is a preset and the user clicks "Custom", nothing happens and the select snaps back on the next render — an interactive option that appears selectable but is inert. Compare `ParagraphSpacingSelect`, which only renders the `custom` option when a non-preset value is actually active (`{!isPreset && current ? <option value="custom"> : null}`). Consider matching that pattern, or make "Custom" focus the adjacent input.

### 3. Multi-block selection cascades the `lineHeight`/spacing readout across block types

`currentLineHeight` / `ParagraphSpacingSelect` read `paragraph`'s attr, then fall back to `heading`'s via `??`. For a selection spanning a plain paragraph (no line-height) and a heading that has one, the toolbar reports the heading's value as if it applied to the whole selection. Minor and consistent with how `getAttributes` behaves for split selections, but the `??` fallback across two discrete block types can surface a value the primary block doesn't have. Low priority.

### 4. Leading zeros pass through un-normalised

`LINE_HEIGHT_RE` (`/^\d+(\.\d+)?$/`) and `SPACING_RE` accept `001.5` / `012px` and emit them verbatim into the style string (`line-height: 001.5`). Valid CSS and harmless for rendering, but it is a non-normalised numeric string riding on the very declaration that must byte-align with the backend `toDOM`. If the backend sanitizer normalises (strips leading zeros) while the FE does not, the two serialisations drift. Worth normalising via `String(Number(v))` on the FE side, or confirming the backend keeps the raw string too.

### 5. New toolbar controls have no CSS

`.octo-line-height`, `.octo-line-height-custom`, `.octo-space-before`, `.octo-space-after` are not styled in `editor/styles.css`, whereas the sibling `.octo-font-size` select is (fixed height, border, radius, padding). The new selects/input will render with default browser chrome and look inconsistent with the rest of the toolbar. Add matching rules.

## Coverage / blind spots

- **Backend byte-alignment is the load-bearing contract and is not verifiable from this diff.** The correctness of `getBlockAttrs`/`setBlockAttrs` `toDOM` (same order, `"; "` separator, unitless line-height) and the Typst `renderTypst.ts` path live in the companion `octo-docs-backend` PR; this review cannot confirm the two serialisations actually match byte-for-byte. That should be validated in the backend PR or an integration test.
- **Yjs/Y.Doc round-trip** is asserted only indirectly (via DOM parse/serialize). No golden Yjs-update regression exists yet (the schema test file itself flags this follow-up).
- The `spacing` attrs stack additively with the docx **default** paragraph/heading spacing in `createDefaultStyles` — a per-block `before/after` overrides the style default (docx semantics), which is the intended behavior, but visual parity with the on-screen margins was not verified.

yujiawei (approved):
# Code Review — PR #722 (octo-web) — Re-review

Re-reviewed at head SHA `69f62e2fd196fb401214963aad58fb7b3c7eadd3` (the branch was rebased onto current `main`; new merge-base `0470f620`, the v16 `fontFamily` commit). A prior pass on head `ef80faba` requested changes because the branch was CONFLICTING with `main` and carried a factually-wrong schema note. **Both issues are now resolved.**

## Verdict: APPROVED

- **Branch rebased onto current `main`.** State is now **MERGEABLE** with zero conflict markers in any changed file.
- **The stale v16 premise is fixed.** `schema/index.ts` now correctly documents `v16 — SCHEMA-SPEC §6: fontFamily attribute` (the previous "reserved for #59/CJK, hold merge" text is gone). `SCHEMA_VERSION = 17` sits cleanly on top of the real v16.
- **Test conflicts resolved correctly.** `extensions.test.ts` and `schema/index.test.ts` now carry **both** the v16 `fontFamily` assertions (from main) and the v17 line-spacing assertions. `extensions.ts` registers `FontSize → LiveFontFamily → TextAlign → LineHeight` (live) and `FontSize → FontFamily → TextAlign → LineHeight` (preview) — the `TextAlign`-before-`LineHeight` ordering that the byte-alignment contract depends on is preserved in both sets.

## Verification (this pass, current head)

- ✅ Full docs test suite green — **84 tests pass** (`LineHeight`, `spacing`, `schema/index`, `extensions`, `Toolbar`), up from 75 now that the merged v16 fontFamily cases are included.
- ✅ Clean merge — no `<<<<<<<`/`=======`/`>>>>>>>` in any file changed vs merge-base.
- ✅ Feature code (`LineHeight.ts`, `styles.ts` `mapSpacing`, `nodes.ts` wiring, sanitisers) is unchanged from the previously-reviewed content; the two-sided sanitisation, `<=1000`/`<=10` cap parity, and canonical style ordering all still hold. em→twips (16px/em × 15twips/px) and unitless→240ths math re-confirmed (`1.5`→`360`, `12px`→`180`, `1em`→`240`).
- ✅ Schema is attrs-only — `SCHEMA_NODES`/`SCHEMA_MARKS` unchanged; `lineHeight`/`spaceBefore`/`spaceAfter` correctly excluded from both lists.

## On the first-pass automated finding

The first-pass suggested APPROVE, which happens to match this terminal verdict — but its schema-version check ("v16 reserved for CJK ✅") was **incorrect at the head it reviewed**: at that point `main` had already landed v16 = `fontFamily` and the PR was CONFLICTING. The verdict here flips to APPROVED because the author rebased and genuinely fixed the blocker, not because the check passed on the old head. Confirmed, not rubber-stamped.

## Non-blocking (P2 / nits — do not block merge)

These carry over from the prior pass and are cosmetic; fold into a follow-up if convenient:

1. **`LineHeightSelect` renders a selectable "Custom" option that silently no-ops** (`Toolbar.tsx`, ~L728). When the current value is a preset, clicking "Custom" does nothing and the select snaps back. `ParagraphSpacingSelect` handles this better (renders `custom` only when a non-preset value is active).
2. **New toolbar controls have no CSS.** `.octo-line-height`, `.octo-line-height-custom`, `.octo-space-before`, `.octo-space-after` are not styled in `editor/styles.css`, whereas the sibling `.octo-font-size` is (height/border/radius/padding). They'll render with default browser chrome.
3. **`ParagraphSpacingSelect` reset option reuses the control title** ("Space before"/"Space after") as its default-option text, whereas `LineHeightSelect` uses a dedicated "Default" label. Consider a `spacingDefault` key so the reset action reads clearly.
4. **Leading zeros pass through un-normalised** (`001.5`, `012px` are emitted verbatim). Harmless CSS, but a normalisation mismatch risk against the backend `toDOM` — normalise via `String(Number(v))` or confirm the backend keeps the raw string.
5. **Multi-block selection** cascades the `lineHeight`/spacing readout across block types via `??`, so a selection spanning a plain paragraph + a styled heading can show the heading's value. Minor.

## Coverage / blind spots

- **Backend byte-alignment** (`getBlockAttrs`/`setBlockAttrs` `toDOM` order + `"; "` separator + unitless line-height, and the Typst path) lives in the companion `octo-docs-backend` PR and is not verifiable from this diff. It remains the load-bearing cross-repo contract — validate there.
- No golden Yjs-update round-trip regression yet (asserted only indirectly via DOM parse/serialize); the schema test file itself flags this follow-up.
- Advisory legs were run on the prior (content-identical) head: gemini-3.1-pro produced 3 P2s (folded in above); the codex leg was **absent** (mlamp gateway `imagegen` misconfig, both attempts) — no cross-family line-level adversarial pass this cycle. The feature diff is unchanged by the rebase, so those findings still apply.

Jerry-Xin (commented):
Re-reviewed at the current head after the follow-up fixes. All prior concerns are resolved and the change looks ready to merge (posting as a comment).

**tl;dr: LGTM** — schema migration is safe (no data loss on old docs), the style value path is fully validated (no CSS injection reachable), the docx export gap from the prior round is closed, and CI is green.

**✅ Schema migration (v16 → v17) — safe, no data loss**
- `lineHeight` / `spaceBefore` / `spaceAfter` are added as global ATTRIBUTES on `heading` + `paragraph` (not new nodes/marks), each with `default: null` (packages/docs/src/editor/LineHeight.ts:143, 155, 165). This mirrors the proven v5 `textAlign` / v7 `fontSize` / v16 `fontFamily` pattern.
- No migration code is needed or present: docs authored before v17 simply lack the attr, ProseMirror applies the `null` default, and no style is emitted — a lossless round-trip. There is no load path that can throw or drop content on old docs. `SCHEMA_VERSION` bumped to 17 in lockstep (packages/docs/src/schema/index.ts:60), with node/mark lists correctly left unchanged.

**✅ Value application — CSS injection not reachable**
- Both value families are sanitized at parse, render, AND command-set time. `sanitizeLineHeight` requires a bare unitless number and bounds it to `0 < n <= 10`; `sanitizeSpacing` requires `<number>px|em` and bounds the magnitude to `<= 1000` (packages/docs/src/editor/LineHeight.ts:39-83). Any value carrying additional CSS syntax, extra properties, functions, or non-whitelisted units fails the regex and degrades to `null` → no style emitted. The value cannot break out of its declaration.
- `renderHTML` only ever interpolates the already-sanitized value into a fixed `line-height:` / `margin-top:` / `margin-bottom:` fragment (packages/docs/src/editor/LineHeight.ts:145-170).

**✅ Paste / deserialize path — covered**
- `parseHTML` reads the browser-parsed CSSOM values (`element.style.lineHeight` / `marginTop` / `marginBottom`) and re-runs the same sanitizer (packages/docs/src/editor/LineHeight.ts:144, 154, 164), so pasted/imported HTML is validated by the identical whitelist as the UI control. A test explicitly exercises an out-of-range pasted value (line-height 99 + a non-px/em unit) and asserts both degrade to null (packages/docs/src/editor/LineHeight.test.ts).

**✅ DOCX export — now symmetric (prior non-blocking note fixed)**
- The earlier gap (docx accepting values the editor rejected) is closed: `mapSpacing` / `spacingToTwips` reuse the same `sanitizeLineHeight` / `sanitizeSpacing` from the editor, enforcing the identical `<=10` and `<=1000` reject-not-clamp bounds before conversion to twips (packages/docs/src/export/docx/styles.ts). Editor, render, and docx export now agree on the boundary.

**✅ Wiring & i18n**
- `LineHeight` is registered after `TextAlign` in both the live and preview extension builds, preserving the canonical style property order for backend byte-alignment (packages/docs/src/editor/extensions.ts).
- New i18n keys carry real localized values in both en-US and zh-CN (行距 / 段前距 / 段后距 / 自定义); key sets are symmetric.

CI is green on this head (Build + pr-title-lint + scans all pass); not a draft. LGTM.

OctoBoooot (approved):
## Review: #722 @ `69f62e2f` — DELTA re-review (post-rebase); conflicts resolved, v16 comment fixed, ordering hold satisfied — APPROVE

**Verdict:** Approve. The branch was rebased onto main (pulling in #700's v16, #716/#720/#725/#726), which resolves the two items yujiawei's CR raised — her CR is anchored at the pre-rebase state. My prior APPROVE (@`ef80faba`) was auto-dismissed by the rebase; re-affirming at the current head. Risk tier: critical (SCHEMA_VERSION 17). CI fully green (Build + code-review SUCCESS); not draft.

### Resolved by the rebase (yujiawei's two 🔴, both fixed at this head)
- **Branch/SCHEMA_VERSION conflict → resolved.** `mergeStateStatus=CLEAN`, `mergeable=MERGEABLE`, no conflict markers. `SCHEMA_VERSION = 17` (`schema/index.ts:62`) now sits correctly on top of #700's v16 — the version conflict was resolved in the rebase. No DIRTY.
- **v16 comment wrong → fixed.** `:51` now reads `v16 — SCHEMA-SPEC §6: add a fontFamily ATTRIBUTE to the textStyle mark` (the earlier "reserved/skip v16" placeholder is gone, replaced with the real fontFamily description now that #700 landed). `:102-103` and `:113` also correctly attribute v16 to fontFamily.

### v16-first ordering hold — now SATISFIED
The standing hold ("land #700's v16 first") is met: #700 is merged and in this branch's history (`0470f620`). v17 no longer waits on ordering.

### Carried fixes intact through the rebase (blob-verified present)
- Commit-on-blur line-height input (`Toolbar.tsx` `LineHeightCustomInput` + `onBlur`) — the focus-steal fix.
- CSSOM-longhand parse + value whitelist (`LineHeight.ts` `element.style.lineHeight`) — no #700-class injection.
- docx lineHeight ≤10 cap (`styles.ts` `sanitizeLineHeight(attrs.lineHeight)`) — the export-symmetry fix.

### 🟡 (non-blocking)
- yujiawei's LineHeightSelect "Custom" option note — the custom value is the commit-on-blur input (verified working); if the `<select>` "Custom" entry itself is a no-op placeholder that defers to the adjacent input, that's a minor UX clarity point, not a defect. Worth a quick confirm but non-blocking.

### Net
Post-rebase the branch is conflict-free (CLEAN), the v16 comment is corrected, the v16-first ordering hold is satisfied (#700 merged), and all my verified fixes survived. yujiawei's two 🔴s are resolved at this head (her CR is at the pre-rebase state). Approving. Still gated on the critical-tier `needs-human-review` maintainer sign-off. Credit yujiawei across the cycle.

ploy-elison:
Superseded by #735. Because this PR's head lives on a fork we cannot push to, the line-spacing kill-switch flag (default ON) could not be added here. #735 (on a pushable fork) carries this PR's full reviewed content (focus-steal fix, docx lineHeight cap parity, paragraph spacing UI, v16 fontFamily + v17 line-spacing conflict resolution) plus the flag, and is MERGEABLE. Line-spacing frontend will proceed via #735. This PR can be closed by a maintainer with permission.

Jerry-Xin:
see #735

Issue #67, opened 2026-05-20
fix: resolve pre-existing test suite failures (22 failing tests across 7 files)


lml2468:
CI job test (introduced in #66) reveals 22 pre-existing failing tests. PR #66 adds continue-on-error: true temporarily. This issue tracks the remediation.

PR #723, opened 2026-07-14
feat(docs): line-height / paragraph spacing block attrs + toolbar selector (SCHEMA_VERSION 17)

ploy-elison:
## Summary

Frontend half of the docs line-height / paragraph-spacing feature (XIN-964), the counterpart to docs-backend PR #67 (SCHEMA_VERSION 17). Adds three global attrs to the `paragraph` and `heading` nodes, mirroring the existing v5 `textAlign` attr exactly (an attr on those two types, not a new node/mark — version bump only):

- `lineHeight` — unitless CSS multiplier (e.g. `"1.5"`) → `line-height:…`
- `spaceBefore` — px/em length → `margin-top:…`
- `spaceAfter` — px/em length → `margin-bottom:…`

All three default to `null`.

## Byte-align with the backend

The four block attrs (textAlign + the three new ones) ride the **same single** inline `style` string. A new `LineHeight` extension, registered right after `TextAlign`, adds the attrs; each rendered `style` fragment merges (via Tiptap's `mergeAttributes`) into the fixed canonical order the backend's `setBlockAttrs` produces:

- only non-null attrs are emitted,
- order `[text-align, line-height, margin-top, margin-bottom]`,
- each declaration `prop: value` (single space after the colon), joined by `"; "`, **no trailing semicolon**.

Examples: `text-align: center`; `line-height: 1.5`; `text-align: right; line-height: 2; margin-top: 8px; margin-bottom: 12px`.

The attrs survive the Y.Doc ↔ ProseMirror round-trip as node attrs regardless of DOM serialization; the style mapping only governs HTML import/export.

## Security

Every value is whitelist-sanitized at **both** parse and render (same both-ends pattern the bookmark URL uses), matching the backend sanitizers:
- `lineHeight`: `^\d+(\.\d+)?$`, `0 < n ≤ 10`
- `spaceBefore`/`spaceAfter`: `^\d+(\.\d+)?(px|em)$`, `0 ≤ n ≤ 1000`

Anything else falls back to `null`, so a hostile inline style can neither enter the Y.Doc nor serialize back out.

## Toolbar

A line-spacing selector (presets `1.0 / 1.15 / 1.5 / 2.0` + a "Custom…" prompt) applies `lineHeight` to the current paragraph/heading, or clears it with "Default". A stored non-preset value surfaces as its own option so the selector reflects real state. Display names are i18n'd in both `en-US` and `zh-CN` (no hardcoded strings). The `spaceBefore`/`spaceAfter` attrs ship in the same schema version; their toolbar UI is deferred.

## Backward compat

Old docs carry no attr → `null` default → `toDOM` emits nothing → lossless. `PMNode.fromJSON` fills the defaults.

## SCHEMA_VERSION

Bumped `15 → 17`. **v16 is reserved by #59 (not yet on main)** and left as a monotonic gap — cut numbers are never reused.

## Tests

New `editor/LineHeight.test.ts` (attr registration, canonical serialization order, parse + render sanitization of hostile values, sanitizer accept/reject tables, set/unset commands, parse↔serialize round-trip) and `editor/LineHeightSelect.test.tsx` (toolbar selector wiring). Existing schema tests updated to v17. `tsc` clean for the changed files; `i18n:check` passes; the docs vitest suite is green apart from a pre-existing environmental `undici`-WebSocket flake in `DocsHome`/`EditorShell` that fails identically on a clean checkout.

## ⚠️ Hold merge

v16 is reserved by #59 (not yet on main). **Do not merge until #59 lands v16** (in lockstep with docs-backend #67). Kept as **draft** until then.

Relates to XIN-964; byte-aligns to docs-backend #67.


ploy-elison:
Closing as duplicate of #722, which is the canonical line-height/paragraph-spacing frontend PR (already paired-reviewed with backend #67, canonical style string byte-verified). Consolidating to avoid two parallel PRs for the same feature.

OctoBoooot (commented):
## Review: #723 @ `acddb041` — line-height / paragraph spacing (SCHEMA_VERSION 17)

**Verdict:** Comment (not approve/reject) — the code is clean and injection-safe, but this PR **collides with #722**, which implements the same feature at the same schema version. That duplication is a maintainer coordination call, not something to resolve by merging one on review alone. Risk tier: **critical** (SCHEMA_VERSION 17) — `needs-human-review` applied; maintainer sign-off required. CI green; draft.

### Code quality — clean, and injection-safe (verified with the #700 lesson in mind)
- `LineHeight.ts` reads the CSSOM longhand (`element.style.lineHeight`/`.marginTop`/`.marginBottom` — browser-decoded for HTML entities *and* CSS escapes) then whitelist-validates the **value** (`^\d+(\.\d+)?$` capped 0<n≤10; `^\d+(\.\d+)?(px|em)$` capped 0≤n≤1000). No raw-string property matching → structurally immune to the CSS-escape bypass that #700 is blocked on. Both parse and render re-sanitize.
- Attr-only on heading/paragraph (correctly out of SCHEMA_NODES/MARKS); registered after TextAlign (verified in the array); toolbar `Custom…` prompt routes through `sanitizeLineHeight`; a stored non-preset value is surfaced as its own option rather than blanking. i18n complete in both locales (real translations incl. the prompt sentence).
- Notably tighter than #722 on one axis: `sanitizeSpacing` caps at ≤1000, closing the unbounded-spacing resilience gap I/others flagged on #722's DOCX side.

### Blocking concerns — cross-PR, need a maintainer decision
1. **Duplicate of #722.** #722 (author Jerry-Xin) and #723 (author ploy-elison) both add the *same* v17 line-height/spacing attrs. Two PRs cannot both define SCHEMA_VERSION 17. One must be chosen and the other closed/rebased — a coordination call above either PR's review.
2. **#723 omits DOCX export.** Unlike #722 (which adds `mapSpacing` in `export/docx/`), #723 touches **no** export path. So `spaceBefore`/`spaceAfter`/`lineHeight` set via #723 render to HTML/Y.Doc but silently **won't appear in `.docx` exports** — a real functional gap if #723 lands as the chosen implementation. (#722 has this covered; #723 also skips Markdown export, same as #722.)
3. **Same v16→v17 ordering hold.** v17 depends on v16 (#700), which is currently CR-blocked (CSS-escape bypass). Whichever of #722/#723 is chosen must land *after* #700's v16 is on main — the schema numbering is monotonic and v16 must exist first.

### Nit
- The custom-value i18n prompt says "allowed 0–10" / "可填 0–10", but `sanitizeLineHeight` rejects exactly `0` (requires `n > 0`). Copy says `0` is allowed; the sanitizer disagrees. Trivial wording fix (say `>0` or `0.1–10`).

### Net
The implementation is correct, injection-safe, and in a couple of ways tighter than #722. But it is a **competing duplicate** of #722 for the same critical schema version, and it's missing the DOCX export #722 has. Rather than APPROVE a duplicate (which would leave two v17 PRs both green), I'm flagging for a maintainer to pick one implementation, close/rebase the other, and honor the v16-first (#700) ordering. Happy to give a full APPROVE on whichever is chosen once the duplication and the export gap are resolved. Credit ploy-elison — clean work.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #723 (octo-web)

## Summary

This PR adds line-height and paragraph-spacing block attributes (`lineHeight`, `spaceBefore`, `spaceAfter`) to the `heading` and `paragraph` nodes as global attrs (SCHEMA_VERSION 17). A toolbar dropdown with presets and a "Custom…" entry drives the `lineHeight` attr; `spaceBefore`/`spaceAfter` ship schema and command wiring but defer toolbar UI. All values are whitelist-sanitized at both parse and render, byte-aligned to docs-backend PR #67's `setBlockAttrs` canonical serialization. The PR is well-structured with thorough test coverage across schema registration, round-trip serialization, sanitizer boundaries, command execution, and toolbar UI wiring.

## Verification

Static analysis only at head `acddb041a2b88a347e88b983792982ac11b638b7`; build and tests not executed in this environment.

- ✅ **Sanitizer correctness** — `sanitizeLineHeight` (`LineHeight.ts:48-53`) regex `^\d+(\.\d+)?$` + `n > 0 && n <= 10` correctly accepts bare positive multipliers in (0, 10] and rejects unit-bearing, negative, zero, non-numeric, and hostile values. `sanitizeSpacing` (`LineHeight.ts:63-67`) regex `^(\d+(?:\.\d+)?)(px|em)$` + `n >= 0 && n <= 1000` correctly enforces px/em units with the 1000 cap.
- ✅ **Dual-end sanitization** — Both `parseHTML` and `renderHTML` run values through the same sanitizer (`LineHeight.ts:99-119`). A hostile value stored in the Y.Doc cannot serialize to the DOM; a hostile import cannot persist unsanitized.
- ✅ **Extension registration order** — `LineHeight` is registered immediately after `TextAlign` in both `buildExtensions` (`extensions.ts:127`) and `buildPreviewExtensions` (`extensions.ts:243`), preserving the canonical [text-align, line-height, margin-top, margin-bottom] merged-style order.
- ✅ **Command namespace** — `setLineHeight`/`unsetLineHeight` are declared by `@tiptap/extension-text-style`; this extension implements them without re-declaring. The new `setSpaceBefore`/`setSpaceAfter`/`unsetSpaceBefore`/`unsetSpaceAfter` commands are declared in the `blockSpacing` namespace (`LineHeight.ts:76-85`) with no collision.
- ✅ **SCHEMA_VERSION bump** — Jump from 15 to 17 (skipping 16) is documented as an intentional monotonic gap reserved for in-flight PR #59 (`schema/index.ts:51-53`). No programmatic consumer in the frontend gates on the specific number — only the test assertion, which is updated.
- ✅ **Schema membership** — `SCHEMA_NODES`/`SCHEMA_MARKS` correctly exclude the new attrs (they are attrs, not nodes/marks). Test updated (`schema/index.test.ts:85-88`).

## Findings

### P2 — Prompt hint text mismatched with sanitizer range (`packages/docs/src/i18n/en-US.json:106`, `packages/docs/src/i18n/zh-CN.json:106`)

The `window.prompt` dialog shows "allowed 0–10" (en) / "可填 0–10" (zh), but `sanitizeLineHeight` rejects 0 (range is `n > 0`, i.e. (0, 10]). A user who types `0` sees the dialog dismiss silently — the sanitizer returns null and the command is skipped (`Toolbar.tsx:560`). Not a security issue (the sanitizer correctly blocks it), but the hint promises a wider range than the validator accepts.

Diff-scope: new (introduced by this PR). Fix: change both locale strings to something like `"Line spacing (e.g. 1.5, allowed >0, ≤10):"`.

### Nit — Schema parity test missing v17 attr assertion (`packages/docs/src/editor/extensions.test.ts:83`)

`extensions.test.ts` asserts v5 `textAlign` on paragraph/heading (`:83`) and v7 `fontSize` on textStyle (`:87`) but adds no corresponding assertion for the v17 `lineHeight`/`spaceBefore`/`spaceAfter` attrs. Coverage exists in `LineHeight.test.ts`, so this is a convention-consistency gap, not a coverage gap. Worth adding `expect(schema.nodes.paragraph.spec.attrs?.lineHeight).toBeDefined()` to keep the cross-cutting parity test complete.

Diff-scope: pre-existing convention not extended. Not blocking.

## Data-Flow Tracing

- **Preset selection → attr → DOM**: User selects "1.5" → `onChange` calls `editor.chain().focus().setLineHeight('1.5')` → `sanitizeLineHeight('1.5')` returns `'1.5'` → `commands.updateAttributes('paragraph', { lineHeight: '1.5' })` + `commands.updateAttributes('heading', { lineHeight: '1.5' })` → node attr set → `renderHTML` runs `sanitizeLineHeight('1.5')` again → emits `style: "line-height: 1.5"`. All safe.
- **Custom input → attr → DOM**: User selects "Custom…" → `window.prompt()` → `sanitizeLineHeight(raw)` → null-guard (`if (clean)`) prevents unsanitized values from reaching the command. All safe.
- **HTML import → attr → Y.Doc**: External HTML `<p style="line-height: 1.5">` → `parseHTML` calls `sanitizeLineHeight(element.style.lineHeight)` → sanitized value stored as node attr. Hostile values (e.g., `"999); background:url(x"`) rejected to null. All safe.
- **Clear attr**: User selects "Default" → `unsetLineHeight` command → `resetAttributes(type, 'lineHeight')` for both paragraph and heading → attr reverts to default (null) → `renderHTML` returns `{}` → no style attr emitted. All safe.
- **Toolbar current value**: `editor.getAttributes('paragraph').lineHeight || editor.getAttributes('heading').lineHeight || ''` — correct fallback chain; non-preset stored values surfaced as dynamic `<option>` with `isPreset` guard preventing duplicates.

- **C1 (Dual-path parity)**: set↔unset for lineHeight, spaceBefore, spaceAfter — all six commands implemented symmetrically (`LineHeight.ts:136-181`), each mapping over `BLOCK_SPACING_TYPES` with `.some()` for the response. Clear.
- **C2 (Control-flow ordering)**: LineHeight registered after TextAlign in both `buildExtensions` and `buildPreviewExtensions`. No other call-sites. Clear.
- **C3 (Authorization boundary)**: N/A — client-side editor extension, no auth/permission surface.
- **C4 (Authorization lifecycle)**: N/A — no auth state.
- **C5 (Build/runtime)**: Not verifiable from this environment; no build/test commands executed.
- **C6 (Governance/docs self-consistency)**: SCHEMA_VERSION changelog in `schema/index.ts` is self-consistent with the PR's changes and the v16 gap documentation. Clear.

## Verdict: APPROVED

Well-implemented feature addition with thorough dual-end sanitization, good test coverage, and clean schema versioning. The only actionable item is the i18n prompt hint mismatch (P2), which is a minor user-facing copy fix. No correctness, security, or blocking concerns.

[Octo-Q] verdict: APPROVE + No P0/P1 blockers found; one P2 (i18n prompt hint vs sanitizer range mismatch) and one Nit (schema parity test convention). Safe for Opus终审 approve.

yujiawei (approved):
# Code Review — PR #723 (octo-web)

Security-focused review of the new `lineHeight` / `spaceBefore` / `spaceAfter` block attributes (SCHEMA_VERSION 17), sanitized at both parse and render. Static analysis at head `acddb041a2b88a347e88b983792982ac11b638b7`; build/tests not executed in this environment. The threat model here is CSS/style-attribute injection and XSS via inline `style`, plus parse↔render sanitizer parity.

## Verdict: no security issues found

The dual-end whitelist sanitization is correct and robust. I attempted to break it and could not.

## Spec compliance

The diff implements exactly the stated scope (three global attrs on heading/paragraph, toolbar selector for `lineHeight`, deferred UI for spacing) with no out-of-scope additions. The v16→v17 monotonic gap is documented and intentional. Nothing is over-built or omitted.

## Security analysis

### 1. Output cannot carry injection metacharacters (`LineHeight.ts:108-127`)

The only place attacker-influenced data reaches a style string is the interpolations `line-height: ${v}`, `margin-top: ${v}`, `margin-bottom: ${v}`, where `v` is the sanitizer return value.

- `sanitizeLineHeight` (`LineHeight.ts:47-53`) returns `s = v.trim()` only after it passes the fully anchored `^\d+(\.\d+)?$`. The accepted string is therefore constrained to `[0-9.]` — it cannot contain `;`, `}`, `<`, `"`, `(`, `url`, whitespace, or newline.
- `sanitizeSpacing` (`LineHeight.ts:61-67`) returns `${m[1]}${m[2]}` reconstructed from the anchored `^(\d+(?:\.\d+)?)(px|em)$` capture groups — constrained to `[0-9.]+(px|em)`.

I brute-forced injecting each of `; } { < > " ' ( ) / \ url * %` and various whitespace into accepted-looking values; zero produced an output string containing a non-`[0-9.]` character. Style-declaration break-out (`5; background:url(javascript:...)`, `5} body{...`, etc.) is not possible because any interior metacharacter fails the anchored regex and returns `null`.

### 2. Anchoring, whitespace, and Number() coercion quirks are all handled

- The regexes are anchored `^...$` with no `m` flag. JS `$` does NOT match before a trailing `\n` here for `\d`, and more importantly any trailing whitespace/newline is stripped by `.trim()` before the regex runs, so `"5\n"` → `"5"` (clean) while `"5\n; evil"` → `null`.
- The `Number(...)` range check runs AFTER the regex, so the classic coercion bypasses never reach it: `"1e2"`, `"+5"`, `"0x5"`, `".5"`, `"5."`, `"1001px"` are all rejected by the regex first. `Number()` is effectively redundant belt-and-suspenders, which is fine.
- `\d` is ASCII-only in JS, so full-width (`３`) and Arabic-Indic (`٥`) digits are rejected. `trim()` covers the full Unicode whitespace set (incl. NBSP, U+2028), so no exotic-whitespace smuggling.

### 3. Parse-side and render-side cover the same inputs

`parseHTML` reads `element.style.lineHeight` / `element.style.marginTop` / `element.style.marginBottom` (CSSOM-parsed accessors) and runs them through the identical sanitizer used by `renderHTML`. A hostile inline style can therefore neither enter the Y.Doc on import nor serialize back out on export — the established both-ends bookmark-URL pattern is mirrored faithfully. The `Custom…` toolbar path (`Toolbar.tsx`) sanitizes the `window.prompt` value and the command re-sanitizes, so the UI cannot inject an unsanitized value either.

### 4. No unsanitized bypass path

Grepping the docs package, the only consumers of these attrs / style props are the sanitized `parseHTML`/`renderHTML` in `LineHeight.ts` and the toolbar. The `sheet/` matches are unrelated hard-coded React inline styles. There is no separate markdown serializer or DOMPurify allowlist in `packages/docs/src` that would emit these attrs outside the sanitizer.

## Non-blocking

### P2 — i18n prompt hint overstates the accepted range (`i18n/en-US.json`, `i18n/zh-CN.json`)

The custom-value prompt says "allowed 0–10" / "可填 0–10", but `sanitizeLineHeight` requires `n > 0` (range is `(0, 10]`), so `0` is silently rejected. This is a copy/validator mismatch, not a security defect (the sanitizer correctly blocks it). Suggest wording like "allowed >0 and ≤10".

## Note for the human reviewer

This PR is labelled needs-human-review. From a security standpoint the frontend sanitizers are sound. The one thing outside this diff worth a human eye is the claimed byte-alignment with the docs-backend `setBlockAttrs` sanitizers (`sanitizeLineHeight`/`sanitizeSpacing`) — the frontend contract is correct in isolation, but if the backend regexes or range bounds diverge (e.g. backend allowing a unit or a wider range), a value the backend accepts but the frontend strips could cause a persistence/render divergence. That parity should be confirmed against the backend PR, which is not part of this diff.

yujiawei (changes requested):
# Code Review — PR #723 (octo-web)

Reviewed at head SHA `acddb041a2b88a347e88b983792982ac11b638b7`. This adds `lineHeight` / `spaceBefore` / `spaceAfter` block attributes to `heading` + `paragraph` (SCHEMA_VERSION 17), mirroring the existing v5 `textAlign` attr, plus a line-spacing toolbar selector. Reviewed across three lenses (security, Tiptap/correctness semantics, spec-compliance & blind spots) with independent adversarial verification of each finding against the checked-out tree.

> Note: an earlier automated Approve was posted on this PR from the same identity as a partial (security-only) pass. **This review supersedes it** as the consolidated verdict.

## 1. Spec compliance — ✅

- **Missing:** none. All three attrs added to `heading`+`paragraph` (`LineHeight.ts:325-359`); both-ends whitelist sanitize at parse + render (`LineHeight.ts:335-355`); `set`/`unset` commands for all three (`LineHeight.ts:362-410`); `SCHEMA_VERSION` 15→17 with the v16 gap documented (`schema/index.ts:51-61`); `lineHeight` toolbar UI, spacing UI deliberately deferred (`Toolbar.tsx:536`); i18n in both `en-US` and `zh-CN`; tests added.
- **Over-build:** none. Correctly does **not** redeclare the `lineHeight` command namespace (already provided by `@tiptap/extension-text-style@3.22.2`; a redeclare would be a TS duplicate-identifier error) — only the new `blockSpacing` namespace is declared (`LineHeight.ts:297-313`). New attrs kept out of `SCHEMA_NODES`/`SCHEMA_MARKS`.
- **Divergence:** none. `LineHeight` is registered **after** `TextAlign` in both `buildExtensions` and `buildPreviewExtensions` (`extensions.ts`), which is load-bearing for the canonical merged-style order `[text-align, line-height, margin-top, margin-bottom]`.

One item to confirm (not blocking): `spaceBefore`/`spaceAfter` ship as live attrs + working commands with no UI. This is a reasonable "data model now, UI later" split (the attrs must exist so collaborative round-trips don't drop them), but please confirm it's intended.

## 2. Code quality — Changes requested

### P1 — Export paths silently drop the three new attrs (data-loss on export)

Both document exporters consume `textAlign` off `heading`/`paragraph` — the exact attribute this PR mirrors — but neither is extended to the new attrs. Export input is `editor.getJSON()`, which carries the full node `attrs` tree (`export/markdown.ts:3`, `MdNode.attrs` at `:16`), so the values **are present** at export time and are then dropped:

- **Markdown** — `wrapAlign` (`export/markdown.ts:200-208`) reads only `node.attrs?.textAlign`. A block with line-height/spacing set (and no non-left alignment) emits no wrapper at all → the formatting is lost.
- **DOCX** — `convertParagraph` (`export/docx/nodes.ts:171-175`) and `convertHeading` (`:177-186`) map `textAlign`→`alignment` and nothing else. This is the sharper miss: the DOCX `Paragraph` API natively supports `spacing: { before, after, line }` and the codebase already uses it for hardcoded presets (`export/docx/nodes.ts:328,344`; `export/docx/styles.ts`), so the values are directly expressible in the target format and are being discarded for no technical reason.

A grep for `lineHeight`/`spaceBefore`/`spaceAfter` across the entire `export/` tree returns zero hits, confirming neither path handles them.

**Requested:** either wire the attrs through both exporters, or make the Markdown drop an explicit, documented decision with a test. Suggested DOCX mapping: `lineHeight` → `spacing.line` (multiplier → 240ths, `lineRule: 'auto'`), `spaceBefore` → `spacing.before` (px → twips), `spaceAfter` → `spacing.after`. **Caveat for the implementer:** headings already carry `spacing:{before,after}` from the named `HeadingN` styles (`export/docx/styles.ts`), so direct-formatting spacing on `convertHeading` will *override* the style preset — merge and decide precedence rather than blindly setting, or paragraph vs heading spacing will diverge.

### P2 — Toolbar preset `1.0` / `2.0` de-syncs after any HTML round-trip

`LINE_HEIGHTS = ['1.0','1.15','1.5','2.0']` (`LineHeight.ts:267`); `sanitizeLineHeight('2.0')` returns the literal source `'2.0'` (`LineHeight.ts:275-281`, returns the trimmed string, not `String(Number(s))`). But `parseHTML` reads `element.style.lineHeight`, and CSSOM normalizes `line-height:2.0`→`"2"` (and `1.0`→`"1"`). After any paste / `setContent` / backend-authored HTML / reload-through-DOM, the stored attr becomes `'2'`, which is not in `LINE_HEIGHTS` → `isPreset` is false (`Toolbar.tsx:546`) → the value surfaces as a stray `<option value="2">` instead of selecting the `2.0` preset. The `1.15`/`1.5` presets are unaffected. The test suite structurally can't catch this — the round-trip tests use `line-height: 2`, never `2.0` through a parse. Fix: compare with `Number()`, or drop the `.0` from the preset list.

### P2 — Comment/claim accuracy ("byte-align" & "no trailing semicolon")

- The header comment states Tiptap "merges each rendered `style` fragment with `'; '`" (`LineHeight.ts:16`). In 3.22.2, `mergeAttributes` actually parses each fragment into `[prop,value]` pairs, inserts into a Map, and re-serializes — the *outcome* (canonical order) is correct and there's no property collision (only `TextAlign` writes `text-align`), but the stated mechanism would mislead a future maintainer.
- "NO trailing semicolon" (`LineHeight.ts:20-22`) is true only for the `toDOM` **spec string** (what the schema tests read). Real `getHTML()` output goes through `dom.style.cssText`, and CSSOM appends a trailing `;` — so serialized HTML is `…12px;`, not `…12px`. This is **not** a Y.Doc corruption risk (attrs are stored structurally in Yjs, not as the style string), but the "byte-align to backend" claim should be scoped to the `toDOM` spec string, or confirmed that backend `setBlockAttrs` comparison tolerates the trailing `;`.

### P2 — i18n copy vs sanitizer bound

The prompt reads "allowed 0–10" (`en-US.json` / `zh-CN.json`), but `sanitizeLineHeight` requires `n > 0` (`LineHeight.ts:280`) — `0` is silently rejected. The true range is `(0, 10]`. Cosmetic.

### Correct by construction (verified, no action)

- **No CSS/style injection or XSS.** The only attacker-influenced sinks are the `line-height:`/`margin-top:`/`margin-bottom:` render fragments. Both sanitizers are fully anchored (`^\d+(\.\d+)?$` and `^(\d+(?:\.\d+)?)(px|em)$`) with numeric bound checks after the regex. 30+ adversarial inputs (`; } { < > " ' ( ) url() calc() javascript:`, whitespace, `1e2`, `+1`, full-width digits) were tested against the actual sanitizers — every one returns `null`. Declaration break-out is impossible, and sanitize runs at both parse and render so nothing hostile can enter the Y.Doc or serialize back out.
- **`.map(type => …).some(r => r)` command shape** over `heading`+`paragraph` is correct and byte-identical to the built-in `TextAlign` idiom — a single shared transaction, `.some()` (not `.every()`) so a single-type selection succeeds.
- **SCHEMA_VERSION v16 gap** is safe: no runtime range/contiguity/`version+1` check consumes `SCHEMA_VERSION` anywhere in `packages/docs/src`.

## 3. Overall verdict — CHANGES_REQUESTED

Spec compliance is ✅, security is clean, and the core editor mechanics are correct. The block is the P1 export data-loss: the two exporters that already handle `textAlign` drop the new attrs, and DOCX has native plumbing that makes the drop gratuitous. Fix the exporters (or make the Markdown drop an explicit, tested decision) and this flips to Approve. The P2s (preset de-sync, comment accuracy, i18n copy) are good to clean up in the same pass while the PR is held as a draft.

## 4. Human verification needed (out of this repo)

- **Byte-alignment to docs-backend #67** cannot be verified from octo-web alone. Please line-diff the frontend sanitizers/serialization against the backend `setBlockAttrs`, specifically: (a) `2.0`-vs-`2` canonicalization on the default presets, (b) `0px`/`0em` acceptance (frontend accepts `n >= 0`), (c) the trailing-`;` question above.
- **Server-side PDF export** (Typst-based, `exportDocPdf(docId)`) is backend-driven; whether it reflects the new spacing depends on #67, not this diff.

PR #727, opened 2026-07-14
feat(loop): workspace slug uniqueness, selection/view persistence, and inline @-mention

### LINKED ISSUE #706: feat(loop): persist workspace selection and per-view display preferences across navigation/reload
## Problem

In dmloop, several user selections are not persisted, so re-entering a page resets the user's context and forces them to re-select every time:

1. **Workspace selection is in-memory only.** The current workspace is held via `setWorkspaceContext` / `currentWorkspaceSlug` / `currentWorkspaceId` (`packages/dmloop/src/api/http.ts`). After a reload or re-entry, the selection is lost — e.g. a user switches to workspace `w2`, leaves, comes back, and is no longer on `w2`.

2. **Display preferences are only partially persisted.** `ProjectPage` persists a single view mode (`list` / `card`) to `localStorage` under one global key (`packages/dmloop/src/pages/ProjectPage.tsx`). This does not cover the kanban / board view, and it is not scoped per page, so different pages share one setting.

## Proposed solution

Persist and restore user context so re-entering a page keeps the user where they were:

1. **Workspace selection** → persist the selected workspace (slug + id) to `sessionStorage`; on load, restore it into the workspace context before the first data fetch. If the persisted workspace is no longer valid, fall back to the default.

2. **Per-view display preferences** → persist the chosen display mode (list / card / kanban-board) per page/context, so each page restores its own last-used view. Extend the existing view-mode persistence to include the board view and scope the storage key per page.

## Open questions

- Storage scope: workspace selection in `sessionStorage` (per tab/session) vs `localStorage` (persists across sessions). Requested: `sessionStorage` for workspace.
- Should display preferences be per-workspace as well, or global per page?
- Keying strategy for per-page view preferences (page id / route).

## Additional context

- Workspace context: `packages/dmloop/src/api/http.ts`, `packages/dmloop/src/index.tsx`.
- Existing view-mode persistence: `packages/dmloop/src/pages/ProjectPage.tsx` (VIEW_KEY, list/card).
- Board/kanban view: `packages/dmloop/src/panel/IssueBoard.tsx`.

### LINKED ISSUE #709: feat(loop): chat-like inline @-mention trigger and clean mention rendering in issue detail
## Problem

The @-mention experience on the issue detail page (`packages/dmloop/src/panel/IssueDetailPage.tsx`) has two significant UX issues:

1. **Mentions require a dedicated `@` button.** Today users must click a separate `@` button (`AtSign` icon, `loop.mention.add`) which opens a dropdown to pick a candidate. This adds a mental-model cost. Comparable products (chat) let users type `@` inline in the editor to trigger a picker popup right at the cursor.

2. **Inserted mentions render as raw markup instead of a name chip.** A mention is inserted as `[@Label](mention://<type>/<id>)` (two protocol shapes, e.g. `mention://agent/...`). While `LoopMarkdown` renders `mention://` links as a highlighted chip, in some places the mention is shown literally, e.g.:

   `@ [@前端外包](mention://agent/3dd7c1b1-a6f7-429b-906a-f7fdb9da8673)`

   It should display like chat — just the name as a clean chip — never the raw markdown/URL token.

## Proposed solution

1. **Inline `@` trigger** — typing `@` in the comment editor opens a candidate picker popup at the cursor (members / AI coworkers / AI squads), selecting one inserts the mention. Remove the reliance on the separate `@` button (or keep it as a secondary affordance only).

2. **Consistent name rendering** — everywhere a mention is displayed (draft/preview/saved comment), render `[@Label](mention://type/id)` as a name chip showing only the label, consistent with chat. Ensure the draft/editor view also shows the resolved name, not the raw `[@Label](mention://...)` string.

## Expected behavior

- Type `@` → picker popup appears inline, choose a target → mention inserted.
- Rendered mention shows just the name as a chip in all views; no raw `mention://` URL or markdown leaks to the UI.

## Additional context

- `IssueDetailPage.tsx`: `insertMention` (~line 379, token `[@${label}](mention://${c.type}/${c.id})`), mention dropdown/button (~lines 1025, 1033).
- `packages/dmloop/src/ui/LoopMarkdown.tsx`: `mention://` chip rendering (`.loop-mention`).
- Mention token format is recognized server-side by `util.MentionRe`.

### LINKED ISSUE #711: feat(loop): rule-based workspace slug generation with random suffix to avoid collisions
## Problem

When creating a workspace, a `slug` is required and is auto-derived from the workspace name via `slugify` (`packages/dmloop/src/pages/LoopPage.tsx:44`). The current rule just lowercases the name and replaces non-alphanumerics with `-`.

Because workspaces are space-level and there are many of them, popular names collide frequently — e.g. many users pick `ainative`, so the derived slug is already taken. There is no collision handling, so the user is forced to invent a unique random slug themselves, which is a poor experience.

## Proposed solution

Introduce a consistent slug-generation rule (reference comparable products) that produces a valid, likely-unique slug automatically:

- Generate a base slug from the name using a clear rule (normalize, lowercase, strip/replace invalid chars, collapse dashes, length cap).
- **Append a short random suffix** (e.g. `ainative-x7f2`) so common base names do not collide and the user does not have to hand-craft uniqueness.
- Keep the slug user-editable, but default to the generated value so the happy path needs no manual input.
- Ideally validate availability (and/or let the backend guarantee uniqueness / retry with a new suffix on collision).

## Expected behavior

- Enter a workspace name (e.g. "AI Native") → slug is auto-filled as something like `ai-native-x7f2`.
- Even if `ai-native` is already used by others, creation succeeds without the user manually thinking up a unique slug.

## Additional context

- Current logic: `slugify` and workspace-create flow in `packages/dmloop/src/pages/LoopPage.tsx` (`slugify` L44; slug auto-fill from name L275; create at L181/L185).
- Slug is immutable after creation (`loop.settings.slugHint`), so the generated value must be correct up front.
- Note the current auto-fill uses the full name; part of this request is to define a proper generation rule + random suffix rather than the naive full-name transform.

MrAladdin:
## What

Three dmloop UX improvements (one PR, one per commit):

- **#711 — Workspace slug uniqueness.** Auto-generate a slug with a short random suffix (e.g. `ai-native-x7f2`); CJK names transliterate via pinyin so they stay zero-input (`前端团队` → `qianduantuandui-x7f2`); on the backend's global-unique 409 the suffix is re-rolled so the happy path never needs manual entry. Slug stays editable and is immutable after creation.
- **#706 — Selection & view persistence.** The selected workspace persists to `sessionStorage` and is restored on reload (falls back to the first workspace when it is gone). Each page's display mode (board/grouped/list, list/card) persists per page via `localStorage` — the issue and my-loop pages stay independent.
- **#709 — Inline @-mention in comments.** Replace the plain-textarea composer + dedicated @ button with a tiptap editor: typing `@` inline raises a grouped candidate popup — all-members, members, agents/squads (type badge + online dot), and issues (recent on empty query / full-text search otherwise), query- and pinyin-filtered. Mentions insert as chips and serialize back to the existing `[@Label](mention://type/id)` token (issue mentions carry no `@`), so server-side `MentionRe` / trigger parsing is unchanged. The trigger preview ("start on send", click to skip) and attach/send controls sit inside the composer card; thread replies share the composer (lazily mounted, so an issue with many threads doesn't mount N editors) and also preview triggers.

## Testing

- **Real-env e2e** on the dev stack (Alice workspace): slug generation for English + CJK names incl. a successful create (`前端团队` → `qianduantuandui-…`); workspace + per-page view restored across reloads; @-mention menu (sections, real avatars, agent online dot, issue status ring, pinyin, keyboard nav, both new-comment and reply composers), chip rendering with no raw-markdown leak, trigger preview + per-agent skip.
- **Unit tests** for the slug generator (8 cases, incl. a guard for the truncation double-dash that would break the backend regex). `pnpm --filter @octo/loop test` is green (25/25).
- Frontend-only: no server / protocol / shared-type change. The mention token and `x-workspace-slug` header contracts are unchanged.

Closes #711
Closes #706
Closes #709

All changes are inside `packages/dmloop` (@octo/loop) + one shared helper it added to `agentApi`; no consumer outside dmloop is touched. Shared touch-points enumerated and checked:
- `api/http.ts setWorkspaceContext`: other consumers (directory cache keyed on `currentWorkspaceSlug()`, `LoopCliAuthorizePage`) read the same getters — values are only pre-seeded from sessionStorage, contract unchanged.
- Mention token `[@Label](mention://type/id)`: unchanged serialization → server `util.MentionRe` + `previewCommentTriggers` + read-view `LoopMarkdown` chipping all unaffected.
- `IssuePage` gained an optional `viewKey` prop (additive); `ProjectPage` view helper de-duplicated into shared `readView/writeView`.
- `agentApi.agentStatusMap()` is new + additive (mirrors existing `runtimeMap()`); no existing agent flow changed.
- No shared type / API / DB / protocol change — frontend-only.

Docs: no doc change needed (no new service/command/config; UI-only within dmloop).


packages/dmloop/src/ui/suggestionMenu.ts
```diff
@@ -0,0 +1,210 @@
...
+      selected = 0;
+      cmd = props.command;
+      closed = false;
+      sync(props.clientRect);
+    },
+    onUpdate: (props) => {
+      items = props.items;
+      cmd = props.command;
+      selected = Math.min(selected, Math.max(0, items.length - 1));
+      sync(props.clientRect);
```
OctoBoooot, line 178: **[major]** The @-mention popup stays dead for the rest of a mention session after Escape. Escape sets `closed = true` + destroys (`:182-186`), but `closed` is reset only in `onStart` (`:171`) and `onExit` (`:206`) — not here in `onUpdate`. `sync()` early-returns while `closed` (`:156`). Repro: type `@jo` → Escape → keep typing `hn` (mention still active → tiptap fires `onUpdate`, not `onStart`) → popup never reappears until you delete back past the `@`. Same after an outside-click (`:148`). Fix: reset `closed = false` at the top of `onUpdate` (or give Escape a per-query hidden flag that `onUpdate` clears), so Escape dismisses only the current query and the menu returns as the query changes.

OctoBoooot (changes requested):
## Review: #727 — dmloop workspace slug / persistence / inline @-mention (3 features)

**Verdict:** Request changes — one real major in the @-mention suggestion popup. The rest is clean and, on the two highest-risk axes (XSS in the mention render path, and the known `@octo/loop` workspace-context global), verified safe by byte-tracing to the trust boundary. Risk tier: normal (loop UI; no auth/schema/crypto). 17 files / +1133 → large; `needs-human-review` warranted. `code-review` CI pending.

### Major — @-mention popup dies for the rest of a mention session after Escape (`ui/suggestionMenu.ts`)
`onKeyDown` Escape sets `closed = true` and destroys the popup (`:182-186`). But `closed` is reset only in `onStart` (`:171`) and `onExit` (`:206`) — **not in `onUpdate` (`:174-178`)**. `sync()` early-returns while `closed` (`:156`). So: type `@jo` → press Escape → keep typing `hn` (the mention is still active, so tiptap fires `onUpdate`, not `onStart`) → `sync()` bails on `closed` and the popup never reappears for the rest of that `@`-session (until you delete back past the `@`, which triggers `onExit`). Byte-verified. Expected UX is that Escape dismisses the popup for the *current* query and it returns as the query changes. Fix (one line): reset `closed = false` in `onUpdate`, or give Escape a per-query "hidden" flag that `onUpdate` clears. Same stuck-state applies after an outside-click (`:148`).

### Verified clean — XSS in the mention render path (top-priority, byte-traced)
No XSS at compose-time or stored-render time. `ui/LoopMarkdown.tsx` uses `react-markdown@8` with **no `rehype-raw`** (raw HTML in stored content is escaped by default), `remark-gfm` only, and a `transformLinkUri` that allowlists `mention://` while routing everything else through the default `uriTransformer` (neuters `javascript:`). `mention://` links render as a **non-navigable `<span>`** (no href) with `children` as escaped JSX text; `mentionExtension.ts` `safeLabel` strips `[`/`]` so a hostile display name can't close the markdown link early. A hostile name (``, `](javascript:...)`) or hostile id cannot execute. Correct design: id is the URL trust anchor, label is escaped display text.

### Verified clean — workspace-context global (the known bug class)
`api/http.ts` still holds the module-global `_workspaceSlug`/`_workspaceId`, now seeded from + persisted to `sessionStorage`. Checked against the two failure modes: (1) the JSON read is try/catch-guarded with a type-check, and `setWorkspaceContext("","")` correctly `removeItem`s; (2) **no `LoopPage` effect includes `t`/locale in its deps** — so a language switch does not re-fetch workspaces or reset the tab (the flicker bug this codebase has hit). The unrestored-on-unmount global is by-design here (it's an intentionally persisted *session* selection, not a transient override of a global other pages depend on) — materially different from the dmpersonal pollution case. Restore-on-reload with list[0] fallback works as claimed.

### Verified good — slug uniqueness
`ui/slug.ts` normalization is safe (lowercase/`[^a-z0-9]+`→`-`/trim/cap-34, edge-dash strip prevents `base--suffix`); empty input (emoji-only / unmapped unicode after `getPinyin`) falls back to `ws-<suffix>`, never an empty slug; suffix uses `crypto.getRandomValues`. The 409 re-roll loop in `LoopPage.doCreateWs` is **bounded at 3 attempts**, re-rolls a fresh suffix each time, and never silently rewrites a user-typed slug (surfaces `slugTaken`). i18n parity holds (new keys translated in both locales, `{{count}}` preserved).

### Minor / nits (non-blocking)
- **PR description mismatch (minor):** the body claims a `pinyin-pro` dependency + slug-file changes that aren't in this diff — transliteration actually reuses the existing local-dict `getPinyin` (`@octo/base`), and the `package.json +7` is the tiptap deps. Reviewers going by the description would look in the wrong place. Update the description.
- `mentionExtension.ts safeLabel` strips `[`/`]` but not `(`/`)` — not exploitable in-repo (LoopMarkdown renders `mention://` as a span regardless), but percent-encoding the id or escaping `()` is worth it as defense-in-depth if the serialized token ever reaches another markdown renderer (e.g. a notification email).
- `suggestionMenu.ts` keyboard selection has no `scrollIntoView` for long lists (selection can move off-screen).
- No integration test for the 409 re-roll loop / autoSlug-vs-typed branch — the most behaviorally-important slug path is untested (slug.ts itself is well-covered).
- suffix `b % 36` on a byte has slight modulo bias — cosmetic, irrelevant to collision resistance at this scale.

### Net
Solid three-feature PR — XSS-safe mention rendering (verified), the workspace-context global avoids the known re-fetch/pollution traps, slug generation is bounded and empty-safe. Holding only on the one real major: the suggestion popup stays dead after Escape mid-mention (`onUpdate` doesn't reset `closed`) — a one-line fix. Once that lands, this is an APPROVE (plus the large-change human sign-off). Credit MrAladdin.

Jerry-Xin (dismissed):
Reviewed all three features on their own risk profile. No blocking correctness or security issues — ready to merge. Summary of the security-sensitive surfaces below, then a few non-blocking notes.

✅ Security review (highest-risk surfaces cleared)

- **@-mention render is XSS-safe.** Both render surfaces use escaped paths: the tiptap node `renderHTML` returns a ProseMirror node spec (`["span", attrs, textContent]`) where the label is a child text node (auto-escaped), and the suggestion popup (`suggestionMenu.ts`) writes every user-controlled string via `.textContent` (name, description) — no `dangerouslySetInnerHTML` / `innerHTML` sink anywhere in the diff (the only `innerHTML` is `el.innerHTML = ""` for clearing). A hostile display name is rendered as inert text, not markup.
- **Mention serialization can't break the token.** `safeLabel()` strips `[` and `]` before building `[@Label](mention://type/id)`, so a crafted name can't break out of the markdown link; the `mention://` contract is preserved via `getText()` textSerializer.
- **Slug uniqueness is server-backed, not racy.** Uniqueness is enforced by the backend's global-unique 409, not a client-side check. On collision an auto-generated slug re-rolls its random suffix (max 3 attempts); a user-typed slug is surfaced as "taken" and never silently clobbered. Normalization (`normalizeBase`) correctly re-strips a trailing dash after the 34-char truncation so the backend regex isn't broken, with a `ws` fallback for empty/emoji-only input.
- **Selection persistence degrades gracefully.** Persisted to `sessionStorage` (per-tab, so no shared-browser cross-user leak). On load, a stale/deleted persisted workspace id falls back to `list[0]` then to the empty guide (`findWs(list, currentWorkspaceId()) ?? list[0] ?? null`) — no crash/blank screen. View mode read is validated against an allow-list with fallback. Effect lifecycle is clean: mount effect has a catch handler, the nav-activation effect gates on `loaded`, and all mitt listeners / settle timers have cleanup.
- **No cross-workspace identity leak.** Agent status and candidates are fetched through the `x-workspace-slug`-scoped API; the mention status lookup only drives a cosmetic online-dot keyed by (global-unique) agent id.

The -257 is a clean composer swap (old inline textarea → new `CommentComposer`) plus lock-file churn; the removed behaviors (draft-preserve-on-`null`, attach button, send disabled-guard, trigger preview, Enter/Mod-Enter submit) are all re-implemented in the new composer, and it even adds per-agent trigger-skip. i18n: 11 new keys with real, correct zh-CN values and en/zh parity. New CSS is fully `.loop-`-scoped — no shared-class/token blast radius.

💬 Non-blocking

🟡 `invalidateAgentStatus()` is defined but never wired in. Workspace switches call `invalidateDirectory()` + `invalidateRuntimeMap()` in `applyWorkspace`/`switchWorkspace`, but not `invalidateAgentStatus()`, so the module-level agent-status cache persists across workspace switches. Impact is cosmetic only (agent online-dots in the mention menu can be stale/missing until a page reload; agent ids are global-unique so no wrong-workspace data is shown). Since it mirrors `runtimeMap`, which you do invalidate, suggest calling it in the same two spots for consistency. See `packages/dmloop/src/api/agentApi.ts:41`, `packages/dmloop/src/pages/LoopPage.tsx:99` and `:166`.

🟡 The suggestion source returns a never-resolving Promise for superseded issue searches to gate stale renders. It works, but repeated fast typing accumulates unresolved async work over the editor's lifetime. Resolving stale calls to `[]` and gating on the token/state would be cleaner. See `packages/dmloop/src/ui/mentionExtension.ts` (the `stale()` helper in the `items` async source).

🔵 The `Agent` / `Squad` badge labels are hardcoded English literals in `suggestionMenu.ts` while the rest of the menu is localized. This PR already touches the i18n files — worth routing these through the locale resources too.

🔵 The new composer CSS mixes Semi design-token variables with hardcoded fallback colors/spacing; prefer tokens consistently for new component CSS. See `packages/dmloop/src/ui/composer.css`.

✅ Highlights: focused unit tests for the slug generator (normalization, suffix length, truncation double-dash guard); small, scoped view-persistence helper; composer preserves the existing `mention://` serialization contract exactly.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #727 (octo-web)

## Summary

This PR delivers three features for the dmloop workspace panel: (1) **workspace slug uniqueness** — `normalizeBase()` + `slugSuffix()` using WebCrypto produce collision-free slugs, with a 409 retry loop on create; (2) **selection/view persistence** — workspace context is restored from `sessionStorage` on reload, and view mode (list/board/kanban) is persisted to `localStorage` per workspace; (3) **inline @-mention in comments** — a tiptap-based rich-text composer with dependency-free suggestion popup replaces the old textarea + dropdown, serializing mentions to the backend contract `[@label](mention://type/id)`.

Overall: well-structured, thoughtfully decomposed, test-covered for the slug layer. Three P2 observations and one nit below — no blockers.

## Verification

Static analysis at head `57c22259e7d8781cfe91ee2d0761ab712be263fd`; build and tests not executed in this environment.

- ✅ **Slug generation** — `normalizeBase()` at `slug.ts:3` correctly lowercases, replaces non-alnum with dash, strips edge dashes, caps at 34 chars, and re-strips trailing dash after truncation. `slugSuffix()` at `:15` uses `crypto.getRandomValues(new Uint8Array(4))` to produce 4-char `[a-z0-9]` — sufficient entropy for workspace-level uniqueness. Tests at `slug.test.ts` cover empty-for-CJK, cap enforcement, and backend regex compliance.
- ✅ **409 retry loop** — `doCreateWs()` at `LoopPage.tsx:189` retries up to 3 times, re-rolling the suffix on each conflict. Only auto-slugs get re-rolled; user-typed custom slugs fail through correctly. The suffix is generated once in `openCreateWs()` at `:153` and held stable during typing — good UX.
- ✅ **Mention token contract** — `mentionToToken()` at `mentionExtension.ts:38` produces `[@label](mention://type/id)` for `member`/`agent` types and `[label](mention://type/id)` (no `@`) for `issue`/`project`. Backend contract confirmed. `safeLabel` escapes `]` and `\` — correct for Markdown link syntax.
- ✅ **Stale-query debouncing** — `buildLoopMention()` at `mentionExtension.ts:83` uses a `queryToken` counter with a never-resolving promise pattern. When a new query arrives mid-debounce, the stale closure resolves to `never`, so the tiptap suggestion plugin ignores it. Sound.
- ✅ **Workspace context persistence** — `readWorkspaceCtx()` at `http.ts:28` restores from `sessionStorage("loop.workspace.ctx")` on module init. `setWorkspaceContext()` writes/removes. JSON parse errors swallowed — degrades to empty slug/id. Session-scoped (not persistent across tabs) — intentional for workspace switching.
- ✅ **View mode persistence** — `readView<T>()` at `viewMode.ts:3` reads from `localStorage`, validates against the allowed array, returns fallback on any mismatch. `writeView()` swallows errors. `IssuePage` receives `viewKey` prop (workspace slug) to namespace per-workspace. Correct.
- ✅ **suppress_agent_ids wiring** — `ThreadReply` passes `suppressIds` through `replyToThread` → `addComment`. `addComment` at `issueApi.ts:165` sends `suppress_agent_ids: suppressAgentIds.length ? suppressAgentIds : undefined` — omitting when empty, matching backend contract.
- ✅ **Lazy-mount pattern** — `ThreadReply` at `IssueDetailPage.tsx` shows a stub button, mounts `CommentComposer` on click. Avoids creating N tiptap editors per thread. `editor.destroy()` called in cleanup.

## Findings

No P0/P1 issues. Three P2 observations and one nit below.

### P2 — Suggestion popup positioning uses `absolute` with viewport coords (`suggestionMenu.ts:126`)

```ts
el.style.position = "absolute";
el.style.left = `${rect.left}px`;
el.style.top = `${rect.bottom}px`;
```

`rect` comes from `getBoundingClientRect()` which is viewport-relative, but `position: absolute` resolves against the nearest positioned ancestor, not the viewport. This works correctly only if the composed element has no positioned ancestor between the editor and the viewport root — which is fragile. If any parent container sets `position: relative` (common in dashboard layouts), the popup will be offset incorrectly.

**Fix**: either change to `position: fixed` (simplest), or add `window.scrollX` / `window.scrollY` offsets, or walk up `offsetParent` to compute the correct absolute position. The `composer.css` sets `z-index: 2000` which pairs correctly with `fixed` positioning.

Diff-scope: new code in this PR. Severity: P2 — visible misalignment in certain layout contexts, not a crash or data loss.

### P2 — `@tiptap/core` pinned to exact version while others use caret (`package.json:10`)

```json
"@tiptap/core": "3.22.2",
"@tiptap/extension-mention": "^3.22.2",
"@tiptap/extension-placeholder": "^3.22.2",
```

`@tiptap/core` is pinned to exact `3.22.2` while all other tiptap packages use `^3.22.2`. This will cause version drift: when `pnpm update` runs, the extension packages may upgrade to `3.23.x` while core stays at `3.22.2`. Tiptap extensions depend on core's internal APIs — version skew between core and extensions can produce subtle runtime errors (missing methods, changed hook signatures).

**Fix**: align to `"@tiptap/core": "^3.22.2"` unless the exact pin is intentional for a known compat reason (in which case a comment in package.json or a changeset note would help).

Diff-scope: new code. Severity: P2 — latent risk, not immediately broken.

### P2 — i18n labels captured at editor mount time, stale on locale change (`CommentComposer.tsx:96`)

```ts
const extensions = useMemo(() => [
  StarterKit.configure({ ... }),
  buildLoopMention(getCandidates, labels),
  Placeholder.configure({ placeholder: t("commentPlaceholder") }),
], []); // empty deps — intentional, must not re-create
```

The `labels` object (i18n strings for mention group headers: "Members", "Agents", "Issues") and the `Placeholder` text are captured at mount time with empty dependency array. The candidates callback and submit handler correctly use refs to stay fresh, but the mention group labels in the suggestion popup will show the locale from when the editor was first mounted. If a user switches language without unmounting the composer, the popup labels remain in the old locale.

This is acknowledged in the code comment ("labels captured at mount; candidates/submit use refs"), and the practical impact is low (locale switching mid-session is rare). But for completeness, either document this as a known limitation or accept a locale prop change as a trigger to rebuild extensions.

Diff-scope: new code. Severity: P2 — cosmetic, rare trigger.

### Nit — `pinyinCache` never evicted (`mentionExtension.ts:44`)

```ts
const pinyinCache = new Map<string, string>();
```

Module-level cache grows unbounded with unique names. For workspace-scale usage this is negligible (hundreds of entries at most), but if this pattern is ever reused at org-scale, a bounded LRU or periodic clear would be prudent. No action needed now.

## Data-flow trace

|---|---|---|
| `mentionToToken()` output | Serialized via `editor.getText({ textSerializers: { mention } })` in `toMarkdown()` | ✅ Correct format `[@label](mention://type/id)` |
| Agent status dots in suggestion | `agentStatusMap()` singleton → `_agentStatus` Map → `agentDotColor(status)` | ✅ Fetched once, shared across all composers |
| Workspace slug on reload | `sessionStorage("loop.workspace.ctx")` → `readWorkspaceCtx()` → `http.ts` module init | ✅ Restored before first API call |
| View mode per workspace | `localStorage("loop.view.<viewKey>")` → `readView()` → `IssuePage` init | ✅ Validated against allowed array |
| `suppress_agent_ids` in comment | `ThreadReply` UI → `replyToThread(suppressIds)` → `addComment(id, content, parentId, suppressAgentIds)` | ✅ Correctly wired end-to-end |
| Pinyin name matching | `getPinyin(name)` from `@octo/base` → cached in `pinyinCache` → `nameMatches()` checks substring | ✅ Correct lazy evaluation |

## Blind-spot checklist (R5)

- **C1 — Dual-path parity**: Mention serialization in `toMarkdown()` ↔ backend parsing contract `[@label](mention://type/id)` — ✅ aligned. `addComment` main path and `replyToThread` reply path both accept `suppressAgentIds` — ✅ symmetric.
- **C2 — Control-flow ordering / reuse**: `buildLoopMention` is called once per editor mount; `agentStatusMap()` is a singleton. No re-entrant or nested reuse risk. N/A.
- **C3 — Authorization boundary**: No new endpoint or capability exposed. Comment creation uses existing `addComment` which goes through standard auth. N/A.
- **C4 — Authorization lifecycle**: No auth/status changes. N/A.
- **C5 — Build/runtime divergence**: 7 new tiptap dependencies added. `pnpm-lock.yaml` updated. The `@tiptap/core` exact pin is noted in P2 above. Runtime behavior depends on tiptap's ProseMirror integration — tested via static analysis only.
- **C6 — Governance/docs**: No governance doc changes. N/A.

## Cross-round blocker recheck (R6)

N/A — automated review review, no prior rounds.

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues. PR delivers three well-decomposed features (slug uniqueness, persistence, @-mention) with proper data-flow wiring, test coverage for the slug layer, and a clean lazy-mount pattern. Three P2 observations (suggestion popup positioning, tiptap version pinning, stale i18n labels) are non-blocking quality items.

OctoBoooot (approved):
## Retracting my #727 Escape-popup 🔴 → APPROVE — I mis-read it; the tiptap Escape path resets `closed` (credit Jerry-Xin)

My REQUEST_CHANGES held on a claim that the @-mention popup stays dead after Escape because `onUpdate` doesn't reset the `closed` flag. **That's wrong — I retract it.** I traced only the app-side `suggestionMenu.ts` state machine and assumed Escape leaves `closed=true`; I never checked what `@tiptap/suggestion` actually does on Escape. Jerry-Xin flagged it; I byte-verified against the library and he's right. Re-anchoring to APPROVE at the current head `a99faef4` (the `suggestionMenu.ts` blob is unchanged from my CR head — this is a rebase, the popup logic didn't change).

### Why the finding is refuted (byte-traced to `@tiptap/suggestion@3.22.2`)
The plugin's `handleKeyDown` Escape branch (`dist/index.js:328-332`):
```js
if (event.key === "Escape" || event.key === "Esc") {
  renderer.onKeyDown({...})      // → app onKeyDown sets closed = true  (suggestionMenu.ts:184)
  dispatchExit(view, pluginKey)  // → synchronously calls renderer.onExit (:147)
  return true;
}
```
`dispatchExit` (`:147`) calls `renderer.onExit` **synchronously in the same call stack**, and the app's `onExit` (`suggestionMenu.ts:206`) sets `closed = false`. So a single Escape sets `closed=true` then immediately resets it to `false` within the one keydown handler — plus any new `@` runs `onStart` (`:171`) which also resets it. There is no path that leaves `closed` stuck true; typing `@` again after Escape works. My "onUpdate is the only reset" premise was incomplete — I missed the library's Escape→onExit dispatch. No bug.

### Accountability
This is the inverse of the mistakes I caught elsewhere this cycle: on #700/#721 I found real bugs a layer deeper than others looked; here I *created* a false finding by stopping one layer too shallow — reading the app callbacks without tracing the library that drives them. Banking: **for a finding about a library-driven callback's state machine, trace the library's actual call sequence (at the pinned version), not just the app-side handlers — a reset you don't see in app code may be dispatched by the framework.**

### The rest of my #727 review stands (unchanged, verified)
- **@-mention render XSS — clean** (verified at the `LoopMarkdown` boundary: react-markdown@8, no rehype-raw, `mention://` → non-navigable span, label escaped, `safeLabel` strips brackets).
- **Workspace-context global** — no `t`/locale in effect deps, guarded sessionStorage, clear-on-empty; not the #619 pollution pattern.
- **Slug** — bounded 409 re-roll (max 3), empty-safe `ws-` fallback, `crypto` entropy. **Persistence** — per-tab sessionStorage, list[0] fallback, view allowlist-validated.
- Non-blocking minors carried: PR description claims a pinyin dep not in the diff; `invalidateAgentStatus` defined-but-never-called (stale agent dots after workspace switch); superseded mention search uses a never-resolving promise (works, but accumulates pending async — prefer resolve-`[]`+token-gate); `safeLabel` doesn't escape `()`.

### Net
Retracting the Escape 🔴 — it was my error (didn't trace tiptap's Escape→onExit reset). APPROVE at `a99faef4`. The feature is XSS-safe, slug/persistence sound; only the non-blocking minors remain. Credit Jerry-Xin for the catch; MrAladdin for the work.

yujiawei (approved):
# Code Review — PR #727 (octo-web)

Reviewed at head `a99faef49e151c415d23e50efcdab60c22aa2279`. Frontend-only change in `@octo/loop` (+ one additive helper in `@octo/base` consumers). Three self-contained concerns: workspace slug uniqueness, selection/view persistence, and the inline @-mention composer. Verification: `pnpm --filter @octo/loop test` green (25/25); no dangling references to the removed `slugify` / `insertMention` / `readViewMode` / `writeViewMode` / old i18n keys; new i18n keys present in both `en-US` and `zh-CN`.

> Note: this review reflects an independent re-check at the current head. The branch was rebased and the comment-trigger preview was refactored into a dedicated `useCommentTriggerPreview` hook (signature-gated so plain typing no longer re-hits the server) since the initial push — that refactor is a clean improvement and is included in this verdict.

**Spec: ✅**

- **Missing:** none. All three advertised features are implemented — slug auto-generation with pinyin transliteration + random suffix + 409 re-roll; `sessionStorage` workspace persistence with first-workspace fallback; per-page view-mode persistence; tiptap inline @-mention with grouped candidates, chip serialization, and trigger preview/skip.
- **Extra (out of scope):** none. The diff stays within dmloop; the only shared-surface additions (`agentStatusMap` in `agentApi`, optional `viewKey` prop on `IssuePage`, `readView`/`writeView` extraction) are additive and backward-compatible.
- **Deviation:** none. The mention token contract `[@Label](mention://type/id)` (and the no-`@` form for issue/project) is preserved by `mentionToToken`, so server-side `MentionRe` / `previewCommentTriggers` parsing is unchanged. The `x-workspace-slug` header contract is unchanged.

Blast-radius claims in the PR description were spot-checked and hold.

## 2. Code quality

**Quality: Approved**

No P0/P1 issues found. The security surface for a `security_sensitive` classification was examined closely and is well-contained (see Security below). The items below are all advisory (P2).

### Security (examined; no blocker)

- **Mention-token injection — safe.** `safeLabel()` strips `[` and `]` from the label; the URL half (`mention://type/id`) is entirely code-controlled. A label cannot break out of the `[label](url)` link to forge a different target or inject a `javascript:` URL. On render, `LoopMarkdown` uses `react-markdown@8` **without** `rehype-raw` (raw HTML is escaped) and only whitelists the `mention://` scheme, rendering it as a non-navigable `<span>` chip. `)`/newline/`mention://` inside a label stay as link *text*, not syntax. No XSS path found.
- **Popup DOM building — safe.** `suggestionMenu.ts` builds rows via `createElement` + `textContent`; the only `innerHTML` use is `el.innerHTML = ""` to clear. No user string reaches `innerHTML`.
- **`sessionStorage`/`localStorage` trust.** `readWorkspaceCtx` type-guards the parsed blob and `readView` allowlist-checks the value, so a tampered store degrades to safe defaults rather than injecting arbitrary values. Storage is same-origin (only writable by code already in the origin), the persisted slug is server-issued, and the backend authorizes the `x-workspace-slug` header server-side. Please still have a human confirm the backend treats that header as untrusted input (defense-in-depth), given the classification.

### P2 — advisory / nits (non-blocking)

- **`agentStatusMap` cache is never invalidated** — `packages/dmloop/src/api/agentApi.ts` (`agentStatusMap` / `invalidateAgentStatus`). The module-level cache is populated once; `invalidateAgentStatus()` is exported but never called, and agent mutations (`createAgent`/`updateAgent`/archive/restore) only invalidate the directory cache. Effect is cosmetic: the agent online-status dot in the @-mention popup can be stale until a page reload. Consider calling `invalidateAgentStatus()` from `afterDirectoryMutation` (mirrors `invalidateRuntimeMap`).
- **Never-resolving `stale()` promise** — `packages/dmloop/src/ui/mentionExtension.ts` (~line 142). Returning `new Promise(() => {})` to gate superseded queries is an intentional and effective debounce, but it leaves small unresolved promise objects to be GC'd once tiptap drops the reference. Functionally fine; an `AbortController` or resolving to `[]` would be tidier.
- **`slugSuffix()` modulo bias** — `packages/dmloop/src/ui/slug.ts` (~line 20). `b % 36` over bytes 0–255 gives a mild non-uniform distribution across the alphabet. Irrelevant for a workspace identifier suffix (collision resistance is unaffected); noting only for completeness. Rejection sampling would remove it.
- **Stale `x-workspace-slug` on the very first request after reload** — `packages/dmloop/src/api/http.ts` seeds `_workspaceSlug`/`_workspaceId` from `sessionStorage` at module load, before the workspace list is validated in `LoopPage`. The only request in that window is the global `listWorkspaces()` (not workspace-scoped), and it is error-guarded, so impact is negligible and the seeding is by design. Called out for awareness, not as a defect.

### Informational (not caused by this PR)

- `getPinyin()` in `packages/dmworkbase/src/Utils/pinYin.ts` (~line 36) has a pre-existing `while (indexOf('--') > 0)` that fails to collapse a *leading* `--` (should be `>= 0` / `includes`). It does **not** affect this PR: every slug path funnels through `normalizeBase()`, which strips leading/trailing dashes, so the produced slug always matches the backend regex `^[a-z0-9]+(?:-[a-z0-9]+)*$`. Worth a separate cleanup in the shared util.

### Verified and explicitly *not* issues

- **IME composition + Enter-to-submit is safe.** The reply composer submits on Enter via tiptap's `addKeyboardShortcuts`, which routes through ProseMirror's `handleKeyDown`. ProseMirror's `editHandlers.keydown` early-returns via `inOrNearComposition(view, event)` **before** invoking `handleKeyDown` while `view.composing` is true (confirmed in the installed `prosemirror-view@1.41.8` source, incl. the Safari `compositionend` timing guard). So Enter during CJK/Japanese candidate selection does not fire submit. (This differs from a plain `<input>` handler, which would need an explicit `isComposing` guard.)
- **409 retry loop is correct.** `autoSlug` is gated on `wsSlugTouched`, not on whether the field is empty, so an auto-generated slug re-rolls its suffix on 409 (up to 3 attempts) while a user-typed slug surfaces "taken" and is never silently changed.
- **Ref-forwarded candidates reach the popup.** The empty-deps `useMemo`/`useEditor` is intentional; `getCandidates()` reads `candidatesRef.current` at suggestion time, so candidate updates are visible without rebuilding the editor.
- **`suppressIds` wiring for replies is correct** and computed from the current trigger set at submit time; the new head also prunes `suppressed` to the live wake set.
- **Cross-file integrity:** no lingering references to removed symbols; i18n keys complete in both locales; suggestion popup's document `mousedown` listener is removed in both `onExit` and outside-click `destroy`.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved — no P0/P1. The P2 items above are optional follow-ups; the `agentStatusMap` invalidation is the only one with user-visible (cosmetic) effect and is the most worth addressing.

## 4. Suggestions

- Wire `invalidateAgentStatus()` into `afterDirectoryMutation` so the mention popup's online dots refresh after agent changes.
- Optionally fix the shared `getPinyin` leading-dash edge (`>= 0`) in a separate PR.

## 5. Reviewer transparency

A cross-model advisory pass was attempted to complement this review but **both advisory legs were unavailable this run** (one gateway configuration error, one timeout); neither returned findings. Their absence is **not** treated as agreement. This verdict rests on the primary line-by-line review plus adversarial re-verification of each candidate finding (including reading the installed ProseMirror source to settle the IME question). A human sign-off on the backend's treatment of the `x-workspace-slug` header would fully close out the `security_sensitive` classification.

Jerry-Xin (approved):
Re-reviewed at the new head. Isolating the true delta from rebase drift, the only #727 changes since my last pass are a two-file refactor of the comment-trigger preview; the editor/mention core is byte-identical to the previously approved head. Verdict: approve, with one carried-over non-blocking UX note.

### ✅ What changed (true delta, drift excluded)
- `useCommentTriggerPreview.ts` (new hook) + `IssueDetailPage.tsx` wiring. The inline debounce `useEffect`s were extracted into a single hook keyed on a stable "wake signature" (nonempty + sorted mention set, issue-mentions excluded). This is a net improvement: plain typing that doesn't change the mention set no longer re-hits the server, an in-flight response is guarded by a `cancelled` flag so a stale fetch can't revive a chip, and an empty draft returns `[]` synchronously. The `suppressed` set is still pruned to the current trigger set. No dropped behavior, no regression.
- Everything else in the compare is merge-base drift from other PRs pulled in by rebase — not part of this change.

### ✅ Carried-forward conclusions (re-verified, still hold)
- **@-mention render stays XSS-safe.** The suggestion menu builds nodes via `createElement` + `textContent` for every user-controlled string, and the mention node keeps its escaped-text path with the bracket-stripping label helper. No `dangerouslySetInnerHTML`, the only `innerHTML` write is a clear-to-empty. Byte-identical to the prior head.
- **Slug uniqueness stays server-backed (conflict → re-roll, no clobber); per-tab persistence stays graceful with stale-id fallback.** Unchanged in this delta.
- **Dependencies:** the added packages are the official TipTap/ProseMirror editor family, pinned in the 3.22.x line — reputable, not a typosquat, and already present at the previously reviewed head (so this delta adds no new supply-chain surface).

### 🟡 Non-blocking
- **Escape does not recover the mention popup for the rest of the session.** In the suggestion controller, Escape sets the internal closed flag and the update handler bails while that flag is set, but only start/exit reset it. Scenario: type `@jo` (popup shows) → Escape (popup hides) → keep typing `hn` — the popup stays hidden until the `@` token is deleted and re-entered. A one-line reset of the flag in the update handler restores mid-session recovery. Peer review reached the same conclusion; graded non-blocking here since it's a recoverable UX degradation, not a correctness or security issue, and this file was untouched in the current delta.
- Prior 🟡s (the uncalled status-invalidation path leaving agent dots briefly stale, and the gating promise that never resolves) remain open and non-blocking.

Refactor is clean and the risk surface is unchanged from the approved head.

packages/dmloop/src/ui/suggestionMenu.ts
MrAladdin: Resolving — this finding was retracted by its author (see the subsequent APPROVE review) and independently refuted by two other reviewers. Byte-verified against the installed `@tiptap/suggestion@3.22.2`: the Escape branch (`dist/index.js:328-331`) calls `renderer.onKeyDown` (which sets `closed = true`) and then **synchronously** `dispatchExit(view, pluginKey)`, whose body (`:129-147`) calls `renderer.onExit` in the same call stack. Our `onExit` (`suggestionMenu.ts:206`) resets `closed = false`. So a single Escape sets `closed` true then immediately back to false within the one keydown handler — it cannot stay stuck, and typing `@` again re-opens the popup. No code change needed.

PR #721, opened 2026-07-14
feat(docs): add format painter to editor toolbar

ploy-elison:
## What

Adds a **Format Painter** to the docs editor toolbar (next to Clear formatting). It copies the inline formatting of one selection and paints it onto another — matching the affordance shipped by Feishu / Tencent Docs / Google Docs.

Marks copied (textStyle family): **bold, italic, underline, strikethrough, inline code, text colour, highlight, font size, font family, superscript, subscript.**

## How

- New `editor/formatPainter.ts`:
  - `capturePaintMarks(state)` — reads the paintable inline marks of the current selection (first inline run of a range; stored/caret marks for a collapsed cursor).
  - `applyPaintMarks(editor, marks)` — clears the paintable mark types across the target range, then re-applies the captured marks via a single ProseMirror transaction.
  - Paintable set is an **allowlist**, so `link` (target-specific href) and the comment-highlight decoration are never carried across text.
- `Toolbar.tsx`: a paint-roller button. Click captures the current selection's marks and *arms* (active state); the next selection completed in the editor (`mouseup`) is painted once, then the painter disarms. Painting **replaces** the target's inline formatting rather than merging it.

## Constraints honoured

- **Frontend-only.** No new schema, no node/block/alignment changes, no Yjs document-shape change — it only reads and re-applies existing inline marks.
- **i18n.** Tooltip driven by `t('docs.toolbar.formatPainter')`; keys added to `zh-CN.json` (`格式刷`) and `en-US.json` (`Format painter`). No hardcoded strings. `i18n:check` green.

## Testing

- `formatPainter.test.ts` — capture records bold + textStyle colour, excludes links, returns empty for plain text; apply paints onto a target, replaces (does not merge) formatting, and is a no-op on a collapsed selection.
- `Toolbar.test.tsx` — button renders next to Clear formatting, arms/disarms on click, and an **end-to-end** case (arm → select target → mouseup → target becomes formatted, painter disarms) exercised against a real Tiptap editor.
- New assertions proven red before the change (module absent), green after.
- `tsc` (no new errors — two pre-existing unrelated errors in `dmworkbase/i18n/format.ts` and `members/sort.ts` remain untouched), `i18n:check`, and the docs vitest suite green (the 4 pre-existing `DocsHome` sheet-view flakes are unrelated and present on `main`).

## Acceptance

- Select formatted text → click format painter → select another run → that run receives the same inline formatting; block structure and body content are untouched. ✅

Opened as **draft** to carry review.

PR #729, opened 2026-07-14
feat(loop): show machine-level runtimes on the runtime page when there is no workspace

### LINKED ISSUE #728: feat(loop): let the runtime page work without a workspace (machine-level runtimes)
## Problem / Motivation

When a user has no workspace, the **我的 → 运行时** (Runtime) page is gated: `PersonalPage` renders "请先加入一个 workspace" and never mounts `RuntimePage`. As a result the user cannot open **添加电脑**, cannot connect a daemon, and cannot see the runtimes on their own machine — the runtime list is workspace-scoped (`/runtimes` filtered by `x-workspace-slug`), so with zero workspaces there is nothing to show and no way in.

This blocks first-run onboarding: a brand-new user (or anyone who has left/removed all workspaces) is stuck before they can bring a machine online.

## Proposed Solution

Give the Runtime page a **machine-level (workspace-independent) mode**:

- When `listWorkspaces()` returns an empty list, mount `RuntimePage` in "machine mode" instead of showing the join-a-workspace gate.
- In machine mode the runtime list is fetched from an auth-only, owner-scoped endpoint (`/machine-runtimes`) rather than the workspace-scoped `/runtimes`, so the user sees the runtimes registered on their own machine.
- Skills (still workspace-scoped) is disabled in machine mode.
- Once the user has at least one workspace, the page falls back to the existing per-workspace behavior unchanged.

This matches the page's own framing ("我的 / 你名下的机器与运行时") and lets a user connect a daemon and see their machine before creating any workspace.

## Alternatives Considered

- **Auto-create a hidden default workspace** — rejected: a hidden workspace still leaks into the workspace switcher / navigation, so it is not truly invisible.
- **Make runtime fully workspace-independent** (drop the per-workspace model) — larger refactor touching schema, daemon and every workspace-scoped query; deferred as a separate initiative.

## Affected Component

Other — Runtime / Loop (`packages/dmpersonal`, `packages/dmloop`).

## Technical Approach (optional)

Front-end (this repo):
- `resolveWorkspaceSelection()` decides machine vs workspace mode from the workspace list.
- `runtimeListPath()` selects `/machine-runtimes` (no workspace slug) vs `/runtimes`.
- `PersonalPage` mounts the runtime page in machine mode and disables the Skills tab there.

Depends on the backend (octo-multica) providing the owner-scoped `GET /api/machine-runtimes` endpoint and machine-level runtime registration (`workspace_id` nullable). Tracked/implemented separately in the octo-multica repository.

caster-Q:
## Summary

Make the **我的 → 运行时** (Runtime) page usable when the user has no workspace. Previously `PersonalPage` gated the page with "请先加入一个 workspace" and never mounted `RuntimePage`, so a user with zero workspaces could not open **添加电脑** or see the runtimes on their own machine. This adds a machine-level (workspace-independent) mode.

## Related Issue

Fixes #728

## Changes

- Add `resolveWorkspaceSelection()` — decides `workspace` vs `machine` mode from the workspace list (pure function, unit-tested).
- Add `runtimeListPath()` — selects the auth-only `/machine-runtimes` endpoint when there is no workspace slug, otherwise the workspace-scoped `/runtimes`.
- `runtimeApi.listRuntimes()` picks the endpoint via `runtimeListPath(currentWorkspaceSlug())`.
- `PersonalPage`: when `listWorkspaces()` is empty, mount `RuntimePage` in machine mode (clear the workspace slug) instead of showing the join-a-workspace gate; disable the Skills tab in machine mode; keep the workspace/loading/error states otherwise unchanged.

- Affected module(s): `packages/dmpersonal` (PersonalPage), `packages/dmloop` (runtime API + workspace selection helper).
- New or changed user-visible entry point: no new route; the existing Runtime page now also renders in the 0-workspace state.
- Shared layer touched: none (`ui` / `Components` / `bridge` / `service` untouched).
- If shared code changed, impact scope: n/a.
- Duplicate entry point checked: yes — reuses the existing Runtime page, no new entry point.

Depends on the octo-multica backend providing the owner-scoped `GET /api/machine-runtimes` endpoint and machine-level runtime registration; that side ships separately in the octo-multica repository.

## Testing

- `pnpm --filter @octo/loop test` — `workspaceSelection`/`runtimeListPath` unit tests pass (5 cases).
- Manually verified end-to-end against a local backend + daemon: with 0 workspace the Runtime page renders and lists the machine's runtimes (`claude`/`codex`/`hermes`/`openclaw`) via `/machine-runtimes`; Skills is disabled; after creating a workspace the page falls back to the per-workspace list.

- [x] Manually verified

## Checklist

- [ ] Updated documentation

### DIFF
--- a/packages/dmpersonal/src/PersonalPage.tsx
+++ b/packages/dmpersonal/src/PersonalPage.tsx
@@ -27,70 +27,55 @@ export default function PersonalPage() {
   const { t } = useI18n();
   const [tab, setTab] = useState<PersonalTabKey>("runtime");
   const [workspaceReady, setWorkspaceReady] = useState(false);
-  const [workspaceEmpty, setWorkspaceEmpty] = useState(false);
   const [workspaceError, setWorkspaceError] = useState<string | null>(null);
-  // machine 模式(0 workspace):只有 runtime 走机器级端点;skill 仍按 workspace 管理,
-  // 无 workspace 时不可用。用 state 驱动 tab 禁用渲染,用 ref 供挂载/导航副作用读取。
+  // machine 模式(0 workspace)标记:machine 模式禁用 Skills、清空 workspace 作用域。
+  // 用 state 驱动 tab 禁用渲染,用 ref 供导航激活副作用同步读取。
   const [machineMode, setMachineMode] = useState(false);
   // Loop 与 Personal 共享 @octo/loop 里的模块级 workspace 全局。存下本页选定的 workspace,
   // 以便在 tab 切换 / 导航再激活时重新断言,避免被 Loop 的选择污染(#619 评审)。
   const selectedWsRef = useRef<{ slug: string; id: string } | null>(null);
-  // machine 模式(0 workspace)标记:nav 重激活 / tab 切换时据此清空 workspace 作用域并重铺,
-  // 而不是像 workspace 模式那样重设 slug。避免 selectedWsRef 为 null 时右栏被清空不重铺。
   const machineModeRef = useRef(false);
   const tabRef = useRef<PersonalTabKey>("runtime");
   tabRef.current = tab;
+  const mountedRef = useRef(true);
 
-  const openTab = (key: PersonalTabKey) => {
-    if (!workspaceReady) return;
-    if (machineModeRef.current && key === "skill") return; // skill 需 workspace,机器级模式禁用
-    const ws = selectedWsRef.current;
-    if (machineModeRef.current) {
-      setWorkspaceContext("", ""); // 机器级模式:清空作用域,列表走 /machine-runtimes
-    } else if (ws) {
-      setWorkspaceContext(ws.slug, ws.id); // 铺视图前先把全局上下文拨回本页的 workspace
+  // 解析当前 workspace 归属并铺右栏。每次调用都重新 listWorkspaces —— 本页常驻不重挂,
+  // 且 workspace 可能在别处(如 Loop)被创建/删除,machine↔workspace 模式必须每次重判,
+  // 不能沿用挂载时缓存的判断(#729 评审:否则建了 workspace 仍卡在 machine 模式)。
+  const resolveAndPaint = (showLoading: boolean) => {
+    if (showLoading) {
+      WKApp.routeRight.replaceToRoot(
+        <PersonalWorkspaceState icon={<Loader2 size={36} />} title={t("personal.workspace.loading")} />,
+      );
     }
-    setTab(key);
-    WKApp.routeRight.replaceToRoot(renderTab(key));
-  };
-
-  useEffect(() => {
-    let cancelled = false;
-
-    WKApp.routeRight.replaceToRoot(
-      <PersonalWorkspaceState icon={<Loader2 size={36} />} title={t("personal.workspace.loading")} />,
-    );
-
     workspaceApi.listWorkspaces()
       .then((workspaces) => {
-        if (cancelled) return;
+        if (!mountedRef.current) return;
         const selection = resolveWorkspaceSelection(workspaces, currentWorkspaceId());
-
         if (selection.mode === "machine") {
           // 0 workspace:机器级模式。清空 workspace 作用域(不发 x-workspace-slug),
-          // 运行时列表转而走 /machine-runtimes;不显示「请先加入 workspace」。
+          // 运行时列表走 /machine-runtimes;Skills 需 workspace,禁用。
           machineModeRef.current = true;
           setMachineMode(true);
           selectedWsRef.current = null;
           setWorkspaceContext("", "");
-          setWorkspaceReady(true);
-          setWorkspaceEmpty(false);
-          setWorkspaceError(null);
-          WKApp.routeRight.replaceToRoot(renderTab(tabRef.current));
-          return;
+          // machine 模式禁用 Skills:若当前正停在 Skills tab,回落到 runtime。
+          if (tabRef.current === "skill") {
+            tabRef.current = "runtime";
+            setTab("runtime");
+          }
+        } else {
+          machineModeRef.current = false;
+          setMachineMode(false);
+          selectedWsRef.current = { slug: selection.slug, id: selection.id };
+          setWorkspaceContext(selection.slug, selection.id);
         }
-
-        machineModeRef.current = false;
-        setMachineMode(false);
-        selectedWsRef.current = { slug: selection.slug, id: selection.id };
-        setWorkspaceContext(selection.slug, selection.id);
         setWorkspaceReady(true);
-        setWorkspaceEmpty(false);
         setWorkspaceError(null);
         WKApp.routeRight.replaceToRoot(renderTab(tabRef.current));
       })
       .catch((error) => {
-        if (cancelled) return;
+        if (!mountedRef.current) return;
         const message = error?.message ? String(error.message) : t("personal.workspace.loadFailed");
         setWorkspaceError(message);
         setWorkspaceReady(false);
@@ -102,30 +87,43 @@ export default function PersonalPage() {
           />,
         );
       });
+  };
+  // 始终指向最新的 resolveAndPaint,供 []-依赖的副作用调用(避免陈旧闭包,同时不让
+  // 副作用因 t 变化而重挂——对齐 LoopPage 的 mount-once 行为,切语言不闪回加载态)。
+  const resolveRef = useRef(resolveAndPaint);
+  resolveRef.current = resolveAndPaint;
 
+  const openTab = (key: PersonalTabKey) => {
+    if (!workspaceReady) return;
+    if (machineModeRef.current && key === "skill") return; // skill 需 workspace,机器级模式禁用
+    const ws = selectedWsRef.current;
+    if (machineModeRef.current) {
+      setWorkspaceContext("", "");
+    } else if (ws) {
+      setWorkspaceContext(ws.slug, ws.id);
+    }
+    setTab(key);
+    WKApp.routeRight.replaceToRoot(renderTab(key));
+  };
+
+  useEffect(() => {
+    mountedRef.current = true;
+    resolveRef.current(true);
     return () => {
-      cancelled = true;
+      mountedRef.current = false;
     };
     // 只在挂载时跑一次(对齐 LoopPage):依赖 [t] 会让切语言时重跑,闪回加载态、丢失当前 tab/弹框。
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
   // 顶部一级导航「Personal」被再次点击时,onMenuClick 会先 popToRoot 清空右栏,且本页常驻不重挂
-  // (挂载副作用不会重跑)。这里监听激活事件:重新断言 workspace 上下文并铺回当前 tab,
-  // 修复「切回后右栏空白」与「用别处 workspace 的上下文拉数据」两个问题(#619 评审)。
+  // (挂载副作用不会重跑)。这里监听激活事件:重新解析 workspace 归属(machine↔workspace 可能已变)
+  // 并铺回当前 tab,修复「切回后右栏空白」「用别处 workspace 上下文拉数据」以及「建 workspace 后
+  // 仍卡在机器级模式」(#729 评审)。
   useEffect(() => {
     const onNavMenuActivated = ({ menuId }: { menuId: string }) => {
       if (menuId !== "dmpersonal") return;
-      // 用 ref 判就绪(避免 [] 依赖下的陈旧闭包):machine 模式已就绪但无 selectedWs;
-      // workspace 模式以 selectedWsRef 非空为就绪标志。两者皆未定=加载中,挂载副作用会自铺。
-      if (machineModeRef.current) {
-        setWorkspaceContext("", "");
-      } else {
-        const ws = selectedWsRef.current;
-        if (!ws) return;
-        setWorkspaceContext(ws.slug, ws.id);
-      }
-      WKApp.routeRight.replaceToRoot(renderTab(tabRef.current));
+      resolveRef.current(false);
     };
     WKApp.mittBus.on("wk:nav-menu-activated", onNavMenuActivated);
     return () => WKApp.mittBus.off("wk:nav-menu-activated", onNavMenuActivated);
@@ -147,10 +145,8 @@ export default function PersonalPage() {
           </button>
         ))}
       </nav>
-      {(workspaceEmpty || workspaceError) && (
-        <div className="dmpersonal-sidebar__hint">
-          {workspaceError || t("personal.workspace.requiredTitle")}
-        </div>
+      {workspaceError && (
+        <div className="dmpersonal-sidebar__hint">{workspaceError}</div>
       )}
     </div>
   );
--- a/packages/dmpersonal/src/PersonalPage.tsx
+++ b/packages/dmpersonal/src/PersonalPage.tsx
@@ -29,6 +29,9 @@ export default function PersonalPage() {
   const [workspaceReady, setWorkspaceReady] = useState(false);
   const [workspaceEmpty, setWorkspaceEmpty] = useState(false);
   const [workspaceError, setWorkspaceError] = useState<string | null>(null);
+  // machine 模式(0 workspace):只有 runtime 走机器级端点;skill 仍按 workspace 管理,
+  // 无 workspace 时不可用。用 state 驱动 tab 禁用渲染,用 ref 供挂载/导航副作用读取。
+  const [machineMode, setMachineMode] = useState(false);
   // Loop 与 Personal 共享 @octo/loop 里的模块级 workspace 全局。存下本页选定的 workspace,
   // 以便在 tab 切换 / 导航再激活时重新断言,避免被 Loop 的选择污染(#619 评审)。
   const selectedWsRef = useRef<{ slug: string; id: string } | null>(null);
@@ -40,6 +43,7 @@ export default function PersonalPage() {
 
   const openTab = (key: PersonalTabKey) => {
     if (!workspaceReady) return;
+    if (machineModeRef.current && key === "skill") return; // skill 需 workspace,机器级模式禁用
     const ws = selectedWsRef.current;
     if (machineModeRef.current) {
       setWorkspaceContext("", ""); // 机器级模式:清空作用域,列表走 /machine-runtimes
@@ -66,6 +70,7 @@ export default function PersonalPage() {
           // 0 workspace:机器级模式。清空 workspace 作用域(不发 x-workspace-slug),
           // 运行时列表转而走 /machine-runtimes;不显示「请先加入 workspace」。
           machineModeRef.current = true;
+          setMachineMode(true);
           selectedWsRef.current = null;
           setWorkspaceContext("", "");
           setWorkspaceReady(true);
@@ -76,6 +81,7 @@ export default function PersonalPage() {
         }
 
         machineModeRef.current = false;
+        setMachineMode(false);
         selectedWsRef.current = { slug: selection.slug, id: selection.id };
         setWorkspaceContext(selection.slug, selection.id);
         setWorkspaceReady(true);
@@ -133,7 +139,7 @@ export default function PersonalPage() {
           <button
             key={item.key}
             className={`dmpersonal-sidebar__item ${tab === item.key ? "is-active" : ""}`}
-            disabled={!workspaceReady}
+            disabled={!workspaceReady || (machineMode && item.key === "skill")}
             onClick={() => openTab(item.key)}
           >
             {item.icon}
--- a/packages/dmpersonal/src/PersonalPage.tsx
+++ b/packages/dmpersonal/src/PersonalPage.tsx
@@ -38,11 +38,15 @@ export default function PersonalPage() {
   const tabRef = useRef<PersonalTabKey>("runtime");
   tabRef.current = tab;
   const mountedRef = useRef(true);
+  // 请求代号:resolveAndPaint 由导航事件驱动,快速切换可能并发多个 listWorkspaces;
+  // 每次进入自增,回来时若代号已过期就丢弃,保证"最后一次点击胜出"、不被慢的旧响应覆盖。
+  const resolveSeqRef = useRef(0);
 
   // 解析当前 workspace 归属并铺右栏。每次调用都重新 listWorkspaces —— 本页常驻不重挂,
   // 且 workspace 可能在别处(如 Loop)被创建/删除,machine↔workspace 模式必须每次重判,
   // 不能沿用挂载时缓存的判断(#729 评审:否则建了 workspace 仍卡在 machine 模式)。
   const resolveAndPaint = (showLoading: boolean) => {
+    const seq = ++resolveSeqRef.current;
     if (showLoading) {
       WKApp.routeRight.replaceToRoot(
         <PersonalWorkspaceState icon={<Loader2 size={36} />} title={t("personal.workspace.loading")} />,
@@ -50,8 +54,11 @@ export default function PersonalPage() {
     }
     workspaceApi.listWorkspaces()
       .then((workspaces) => {
-        if (!mountedRef.current) return;
-        const selection = resolveWorkspaceSelection(workspaces, currentWorkspaceId());
+        if (!mountedRef.current || seq !== resolveSeqRef.current) return;
+        // 优先用本页自己存的 workspace,仅当它不存在(如在别处被删)时才回落到共享全局。
+        // 直接读 currentWorkspaceId() 会让 Personal 跟随 Loop 最后选中的 workspace(#619 污染)。
+        const targetId = selectedWsRef.current?.id ?? currentWorkspaceId();
+        const selection = resolveWorkspaceSelection(workspaces, targetId);
         if (selection.mode === "machine") {
           // 0 workspace:机器级模式。清空 workspace 作用域(不发 x-workspace-slug),
           // 运行时列表走 /machine-runtimes;Skills 需 workspace,禁用。
@@ -75,7 +82,7 @@ export default function PersonalPage() {
         WKApp.routeRight.replaceToRoot(renderTab(tabRef.current));
       })
       .catch((error) => {
-        if (!mountedRef.current) return;
+        if (!mountedRef.current || seq !== resolveSeqRef.current) return;
         const message = error?.message ? String(error.message) : t("personal.workspace.loadFailed");
         setWorkspaceError(message);
         setWorkspaceReady(false);
@@ -123,7 +130,8 @@ export default function PersonalPage() {
   useEffect(() => {
     const onNavMenuActivated = ({ menuId }: { menuId: string }) => {
       if (menuId !== "dmpersonal") return;
-      resolveRef.current(false);
+      // showLoading=true:框架会先 popToRoot 清空右栏,重解析是异步的,先铺加载态避免空白闪一下。
+      resolveRef.current(true);
     };
     WKApp.mittBus.on("wk:nav-menu-activated", onNavMenuActivated);
     return () => WKApp.mittBus.off("wk:nav-menu-activated", onNavMenuActivated);
--- a/packages/dmpersonal/src/PersonalPage.tsx
+++ b/packages/dmpersonal/src/PersonalPage.tsx
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef, useState } from "react";
-import { AlertCircle, Cpu, FolderPlus, Loader2, Sparkles } from "lucide-react";
-import { currentWorkspaceId, RuntimePage, setWorkspaceContext, SkillPage, workspaceApi } from "@octo/loop";
+import { AlertCircle, Cpu, Loader2, Sparkles } from "lucide-react";
+import { currentWorkspaceId, resolveWorkspaceSelection, RuntimePage, setWorkspaceContext, SkillPage, workspaceApi } from "@octo/loop";
 import { useI18n, WKApp } from "@octo/base";
 import "@octo/loop/src/pages/loop.css";
 import "./personal.css";
@@ -32,13 +32,20 @@ export default function PersonalPage() {
   // Loop 与 Personal 共享 @octo/loop 里的模块级 workspace 全局。存下本页选定的 workspace,
   // 以便在 tab 切换 / 导航再激活时重新断言,避免被 Loop 的选择污染(#619 评审)。
   const selectedWsRef = useRef<{ slug: string; id: string } | null>(null);
+  // machine 模式(0 workspace)标记:nav 重激活 / tab 切换时据此清空 workspace 作用域并重铺,
+  // 而不是像 workspace 模式那样重设 slug。避免 selectedWsRef 为 null 时右栏被清空不重铺。
+  const machineModeRef = useRef(false);
   const tabRef = useRef<PersonalTabKey>("runtime");
   tabRef.current = tab;
 
   const openTab = (key: PersonalTabKey) => {
     if (!workspaceReady) return;
     const ws = selectedWsRef.current;
-    if (ws) setWorkspaceContext(ws.slug, ws.id); // 铺视图前先把全局上下文拨回本页的 workspace
+    if (machineModeRef.current) {
+      setWorkspaceContext("", ""); // 机器级模式:清空作用域,列表走 /machine-runtimes
+    } else if (ws) {
+      setWorkspaceContext(ws.slug, ws.id); // 铺视图前先把全局上下文拨回本页的 workspace
+    }
     setTab(key);
     WKApp.routeRight.replaceToRoot(renderTab(key));
   };
@@ -53,24 +60,24 @@ export default function PersonalPage() {
     workspaceApi.listWorkspaces()
       .then((workspaces) => {
         if (cancelled) return;
-        const selected = workspaces.find((workspace) => workspace.id === currentWorkspaceId()) ?? workspaces[0] ?? null;
+        const selection = resolveWorkspaceSelection(workspaces, currentWorkspaceId());
 
-        if (!selected) {
+        if (selection.mode === "machine") {
+          // 0 workspace:机器级模式。清空 workspace 作用域(不发 x-workspace-slug),
+          // 运行时列表转而走 /machine-runtimes;不显示「请先加入 workspace」。
+          machineModeRef.current = true;
           selectedWsRef.current = null;
-          setWorkspaceEmpty(true);
-          setWorkspaceReady(false);
-          WKApp.routeRight.replaceToRoot(
-            <PersonalWorkspaceState
-              icon={<FolderPlus size={40} />}
-              title={t("personal.workspace.requiredTitle")}
-              desc={t("personal.workspace.requiredDesc")}
-            />,
-          );
+          setWorkspaceContext("", "");
+          setWorkspaceReady(true);
+          setWorkspaceEmpty(false);
+          setWorkspaceError(null);
+          WKApp.routeRight.replaceToRoot(renderTab(tabRef.current));
           return;
         }
 
-        selectedWsRef.current = { slug: selected.slug, id: selected.id };
-        setWorkspaceContext(selected.slug, selected.id);
+        machineModeRef.current = false;
+        selectedWsRef.current = { slug: selection.slug, id: selection.id };
+        setWorkspaceContext(selection.slug, selection.id);
         setWorkspaceReady(true);
         setWorkspaceEmpty(false);
         setWorkspaceError(null);
@@ -103,9 +110,15 @@ export default function PersonalPage() {
   useEffect(() => {
     const onNavMenuActivated = ({ menuId }: { menuId: string }) => {
       if (menuId !== "dmpersonal") return;
-      const ws = selectedWsRef.current;
-      if (!ws) return; // 尚未就绪:挂载副作用会在加载完成后自行铺视图
-      setWorkspaceContext(ws.slug, ws.id);
+      // 用 ref 判就绪(避免 [] 依赖下的陈旧闭包):machine 模式已就绪但无 selectedWs;
+      // workspace 模式以 selectedWsRef 非空为就绪标志。两者皆未定=加载中,挂载副作用会自铺。
+      if (machineModeRef.current) {
+        setWorkspaceContext("", "");
+      } else {
+        const ws = selectedWsRef.current;
+        if (!ws) return;
+        setWorkspaceContext(ws.slug, ws.id);
+      }
       WKApp.routeRight.replaceToRoot(renderTab(tabRef.current));
     };
     WKApp.mittBus.on("wk:nav-menu-activated", onNavMenuActivated);
--- a/packages/dmloop/src/api/__tests__/workspaceSelection.test.ts
+++ b/packages/dmloop/src/api/__tests__/workspaceSelection.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest";
+
+import { resolveWorkspaceSelection, runtimeListPath } from "../workspaceSelection";
+import type { Workspace } from "../types";
+
+const ws = (id: string, slug: string): Workspace => ({ id, name: slug, slug } as Workspace);
+
+describe("resolveWorkspaceSelection", () => {
+  it("returns machine mode when there are no workspaces", () => {
+    expect(resolveWorkspaceSelection([], "")).toEqual({ mode: "machine" });
+  });
+  it("picks the current workspace when present", () => {
+    expect(resolveWorkspaceSelection([ws("a", "alpha"), ws("b", "beta")], "b")).toEqual({
+      mode: "workspace",
+      slug: "beta",
+      id: "b",
+    });
+  });
+  it("falls back to the first workspace when current id is absent", () => {
+    expect(resolveWorkspaceSelection([ws("a", "alpha"), ws("b", "beta")], "zzz")).toEqual({
+      mode: "workspace",
+      slug: "alpha",
+      id: "a",
+    });
+  });
+});
+
+describe("runtimeListPath", () => {
+  it("uses the workspace-scoped endpoint when a slug is present", () => {
+    expect(runtimeListPath("alpha")).toBe("/runtimes");
+  });
+  it("uses the auth-only machine endpoint when there is no slug", () => {
+    expect(runtimeListPath("")).toBe("/machine-runtimes");
+  });
+});
--- a/packages/dmloop/src/api/runtimeApi.ts
+++ b/packages/dmloop/src/api/runtimeApi.ts
@@ -1,9 +1,10 @@
 // @octo/loop — Runtime API（后端契约联调；Runtime 已并入 Loop 二级菜单）
 import type { RuntimeDevice, ListParams } from "./types";
-import { httpGet } from "./http";
+import { currentWorkspaceSlug, httpGet } from "./http";
+import { runtimeListPath } from "./workspaceSelection";
 
 export async function listRuntimes(params?: ListParams): Promise<RuntimeDevice[]> {
-  const rows = await httpGet<RuntimeDevice[]>("/runtimes");
+  const rows = await httpGet<RuntimeDevice[]>(runtimeListPath(currentWorkspaceSlug()));
   const kw = params?.keyword?.trim().toLowerCase();
   if (!kw) return rows ?? [];
   return (rows ?? []).filter(
--- a/packages/dmloop/src/api/workspaceSelection.ts
+++ b/packages/dmloop/src/api/workspaceSelection.ts
@@ -1,0 +1,28 @@
+// @octo/loop — 运行时/我的 页的 workspace 作用域决策(纯函数,便于单测)。
+import type { Workspace } from "./types";
+
+export type WorkspaceSelection =
+  | { mode: "workspace"; slug: string; id: string }
+  | { mode: "machine" };
+
+// 决定「我的/运行时」页如何取上下文:
+// - ≥1 workspace:选当前(或第一个)-> workspace 模式。
+// - 0 workspace:machine 模式,页面照常挂载并展示用户的机器级 runtime
+//   (daemon 以空 workspace 注册)。不创建任何 workspace,切换器里也不会多出东西。
+export function resolveWorkspaceSelection(
+  workspaces: Workspace[],
+  currentId: string,
+): WorkspaceSelection {
+  if (!workspaces || workspaces.length === 0) {
+    return { mode: "machine" };
+  }
+  const selected = workspaces.find((w) => w.id === currentId) ?? workspaces[0];
+  return { mode: "workspace", slug: selected.slug, id: selected.id };
+}
+
+// runtimeListPath 选运行时列表端点:有 workspace slug 用 workspace 作用域的
+// /runtimes(强 workspace 组);无 slug(0 workspace / 机器级模式)用 auth-only
+// 的 /machine-runtimes —— /runtimes 在此情形会被 workspace 中间件 403。
+export function runtimeListPath(slug: string): string {
+  return slug ? "/runtimes" : "/machine-runtimes";
+}
--- a/packages/dmloop/src/index.tsx
+++ b/packages/dmloop/src/index.tsx
@@ -27,3 +27,5 @@ export {
   setWorkspaceContext,
   LoopApiError,
 } from "./api/http";
+export { resolveWorkspaceSelection, runtimeListPath } from "./api/workspaceSelection";
+export type { WorkspaceSelection } from "./api/workspaceSelection";
--- a/packages/dmpersonal/src/i18n/en-US.json
+++ b/packages/dmpersonal/src/i18n/en-US.json
@@ -8,8 +8,6 @@
   },
   "workspace": {
     "loading": "Loading workspace...",
-    "requiredTitle": "Join a workspace first",
-    "requiredDesc": "Runtimes and Skills are currently managed within a workspace. Join or create one, then come back.",
     "loadFailed": "Failed to load workspaces"
   }
 }
--- a/packages/dmpersonal/src/i18n/zh-CN.json
+++ b/packages/dmpersonal/src/i18n/zh-CN.json
@@ -8,8 +8,6 @@
   },
   "workspace": {
     "loading": "正在加载工作区…",
-    "requiredTitle": "请先加入一个 workspace",
-    "requiredDesc": "Runtimes 和 Skills 当前仍按 workspace 管理。加入或创建 workspace 后再回来查看。",
     "loadFailed": "工作区加载失败"
   }
 }


OctoBoooot (approved):
## Review: #729 — machine-level runtimes on the Runtime page when there's no workspace (fixes #728)

**Verdict:** Approve. A clean, well-scoped fix that makes the 我的→运行时 page usable with zero workspaces, and it correctly handles the shared `@octo/loop` workspace-context global — the exact pitfall from the #619 review (which the author explicitly cites). Risk tier: normal (loop/personal UI; no auth/schema/crypto). CI green; `code-review` pending.

### Verified — workspace-context global handled correctly (the #619 trap)
`@octo/loop` holds a module-level workspace context (`_workspaceSlug`/`_workspaceId`) shared by Loop and Personal; the known failure is a page setting it and then being polluted by the other page's selection on nav-reactivation/tab-switch. This PR does it right on every re-entry path:
- **`openTab`**: re-asserts `setWorkspaceContext("","")` in machine mode, else the stored `selectedWsRef` slug — before rendering. So the global is deterministically restored to *this page's* intended scope every time, no pollution in either direction.
- **`wk:nav-menu-activated` effect** (deps `[]`): reads `machineModeRef.current` (a **ref**, avoiding the stale-closure bug that `[]` deps would otherwise cause — the precise thing that bit #619) and re-asserts empty-or-stored context. Correct dual ref+state pattern (`machineModeRef` for effects, `machineMode` state for the disabled-render).
- **No `t`/locale in either effect's deps** (both `[]`) — so a language switch doesn't re-fetch/reset/flicker. `t` is render-only. This was the other half of the #619 lesson; avoided.

### Verified — routing + logic
- `resolveWorkspaceSelection` (pure): 0 workspaces → `{mode:"machine"}`; else current-or-first → `{mode:"workspace"}`. `runtimeListPath("")` → `/machine-runtimes` (auth-only, dodges the workspace-middleware 403 on `/runtimes`), non-empty slug → `/runtimes`. `runtimeApi.listRuntimes` reads `currentWorkspaceSlug()` at call time, so the endpoint always matches the context the page just asserted. Consistent.
- Machine mode disables the `skill` tab (skill needs a workspace) in both `openTab` (guard) and the sidebar `disabled` — no half-enabled state.
- Doesn't restructure RuntimesPage's internal state (the other MEMORY note — RuntimeDetail remount-wipes-local-state); this is additive mode selection under the same `replaceToRoot` mount path, so it doesn't touch that surface.

### Tests
`workspaceSelection.test.ts` covers the pure decisions (machine on empty, current-workspace pick, first-fallback, and the path mapping). Good for the extracted logic. The PersonalPage wiring itself (machine-mode re-assert on nav) isn't unit-tested, but that's consistent with the existing page and hard to unit-test.

### Minor (non-blocking)
- The machine-mode re-assert correctness rests on `machineModeRef`/`selectedWsRef` being kept in sync with state across every path — they are here, but it's now three coordinated flags (`machineMode`/`machineModeRef`/`selectedWsRef`). A brief comment or a small integration test around "nav away in machine mode → come back → still `/machine-runtimes`" would lock it against future edits. Not blocking.

### Net
Correct machine-vs-workspace mode selection with the workspace-context global handled exactly as the #619 review requires (re-assert on every re-entry, ref-read in `[]` effects, no locale in deps), sensible endpoint routing, pure-function tests. Approving. Credit caster-Q — and for explicitly honoring the #619 pattern.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #729 (octo-web)

## Summary

This PR adds machine-level runtime support to the Personal page. When a user has zero workspaces, instead of showing a "please join a workspace" empty state, the page now enters "machine mode" — it clears the workspace context, routes the runtime list through a new `/machine-runtimes` auth-only endpoint, and disables the Skill tab (which still requires a workspace). The decision logic is extracted into pure functions in `packages/dmloop/src/api/workspaceSelection.ts` with unit tests.

Overall the implementation is clean and well-reasoned. Data flow is correctly traced from workspace resolution through context setting to endpoint selection. One non-blocking P2 noted below.

## Verification

- ✅ **Data flow — machine mode list path** — `PersonalPage` sets `setWorkspaceContext("", "")` → `currentWorkspaceSlug()` returns `""` → `runtimeListPath("")` returns `"/machine-runtimes"` → `httpGet` calls the correct endpoint.
- ✅ **Data flow — header omission** — HTTP interceptor at `packages/dmloop/src/api/http.ts:36` gates `x-workspace-slug` on `if (_workspaceSlug)`, so empty string correctly suppresses the header. Auth headers (`token`, `X-Space-Id`) are still sent unconditionally.
- ✅ **Skill tab disabled in machine mode** — Button `disabled` prop includes `machineMode && item.key === "skill"`; `openTab` also has an early-return guard at `packages/dmpersonal/src/PersonalPage.tsx:43` for belt-and-suspenders safety.
- ✅ **Nav re-activation** — `onNavMenuActivated` handler correctly branches on `machineModeRef.current` vs `selectedWsRef.current`, re-asserting the right context in both modes.
- ✅ **Input shape edges** — `resolveWorkspaceSelection` handles `null`/`undefined`/empty-array inputs. `runtimeListPath` safely falls through to `/machine-runtimes` for any falsy slug.
- ✅ **Async contract unchanged** — `listRuntimes()` still uses the same `await httpGet<T>(...)` pattern with identical error propagation to `RuntimePage`'s `.catch()`.
- ✅ **Pure function tests** — `workspaceSelection.test.ts` covers the three key resolution cases (empty, match, fallback) and both path cases.
- ✅ **Removed import** — `FolderPlus` was only used in the now-removed empty-workspace state; correctly cleaned up.

Static analysis only at head `44f97bb1`; build and tests not executed in this environment.

## Findings

### P2 — `getRuntime` not updated for machine-mode parity (`packages/dmloop/src/api/runtimeApi.ts:7`)

`getRuntime(id)` still hardcodes `httpGet<RuntimeDevice>(\`/runtimes/${id}\`)` without routing through a machine-mode-aware path helper. Grep confirms zero current consumers of this export, so it is not reachable today. However, any future caller invoking `getRuntime()` in machine mode (slug `""`) would hit the workspace-scoped `/runtimes/:id` without the `x-workspace-slug` header and receive a 403. Consider either routing through a `runtimeDetailPath(id, slug)` helper analogous to `runtimeListPath`, or adding a `// workspace-only` annotation to prevent future misuse.

**Diff-scope**: pre-existing dead export, not introduced by this PR. Not amplified — the PR added the list-path helper but did not touch `getRuntime`. Non-blocking.

## Data Flow Trace

| Consumer | Upstream Source | Runtime Value | Reaches Consumer? |
|---|---|---|---|
| `runtimeListPath(currentWorkspaceSlug())` | `_workspaceSlug` set by `setWorkspaceContext("", "")` in machine mode | `""` → `"/machine-runtimes"` | ✅ Yes |
| `x-workspace-slug` header | `_workspaceSlug` via interceptor at `http.ts:36` | `""` → falsy → header omitted | ✅ Correctly omitted |
| `token` / `X-Space-Id` headers | `WKApp.loginInfo.token` / `WKApp.shared.currentSpaceId` | unchanged | ✅ Still sent |
| `resolveWorkspaceSelection(workspaces, currentId)` | `workspaceApi.listWorkspaces()` → `Workspace[]` | `[]` → `{ mode: "machine" }` | ✅ Yes |
| `machineModeRef.current` | Set in mount effect `.then()` | `true` in machine mode | ✅ Read by `openTab` and `onNavMenuActivated` |
| `disabled` prop on skill button | `machineMode` state + `item.key === "skill"` | `true` when machine mode + skill tab | ✅ Yes |

- **C1 — Dual path parity**: CLEAR. `runtimeListPath` covers list. `getRuntime` is the only other `/runtimes` path in `runtimeApi.ts` — has zero consumers (P2 above). `agentApi.ts` uses hardcoded `/runtimes` at lines 10/95 but is not reachable from Personal page in machine mode.
- **C2 — Control-flow ordering**: CLEAR. `resolveWorkspaceSelection` called once in mount effect. `openTab` and `onNavMenuActivated` are mutually exclusive by event type. No nested/double-call risk.
- **C3 — Authorization boundary**: CLEAR (client-side). `/machine-runtimes` relies on same `token` + `X-Space-Id` auth headers. No new endpoint exposure from the frontend — server-side auth for `/machine-runtimes` is out of scope.
- **C4 — Auth lifecycle / container-member state**: N/A. No auth lifecycle changes in this PR.
- **C5 — Build/runtime parity**: CLEAR. No build config, proxy, or packaging changes. `/machine-runtimes` is a plain REST path under the same `LOOP_API_BASE`.
- **C6 — Governance/docs**: N/A. No governance or policy changes.

N/A — first review round.

- `resolveWorkspaceSelection` fallback to `workspaces[0]` when `currentId` not found — correct and tested.
- `cancelled` flag in mount effect prevents stale state writes after unmount.
- `tabRef.current` used inside async callback to read current tab (avoids stale closure over `tab` state).
- State + ref dual-tracking of `machineMode`/`machineModeRef` — state drives render (disabled prop), ref drives side effects (openTab, onNavMenuActivated). Correct pattern for `[]`-dependency effects.
- Test file uses proper `Workspace` type construction via `ws()` helper.

## Verdict: APPROVED

Clean, focused feature PR. The machine-mode data flow is correctly wired end-to-end, pure logic is extracted and tested, and edge cases are handled. The sole P2 (`getRuntime` parity) is non-blocking given its zero-consumer status.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Data flow correctly routes machine-mode users to `/machine-runtimes` with proper auth header handling. One P2 (dead `getRuntime` export lacks machine-mode path) is pre-existing and unreachable.

Jerry-Xin (dismissed):
The PR is relevant to `octo-web`, but the machine/workspace mode decision can become stale in the app’s persistent navigation model.

🔴 Blocking

🔴 Critical — `PersonalPage` never re-resolves workspace mode after its first mount. `MainContentLeft` keeps previously visited menu routes mounted and only toggles visibility (`apps/web/src/Pages/Main/index.tsx:27-31`), then emits `wk:nav-menu-activated` when returning to a menu (`apps/web/src/Pages/Main/index.tsx:263-266`). This PR computes `resolveWorkspaceSelection()` only in the mount effect (`packages/dmpersonal/src/PersonalPage.tsx:57-90`), and the nav activation handler just reuses `machineModeRef` / `selectedWsRef` (`packages/dmpersonal/src/PersonalPage.tsx:116-128`). If a user opens Personal with zero workspaces, then creates a workspace from Loop (`packages/dmloop/src/pages/LoopPage.tsx:185-190`), returning to Personal still clears the workspace context, keeps Skills disabled, and renders Runtime against `/machine-runtimes` instead of falling back to the workspace-scoped `/runtimes`. This contradicts the PR’s stated behavior and leaves the page wrong until a full remount/reload. Please reload/re-resolve workspaces on Personal activation or subscribe to a workspace-change signal, and update the active tab if machine mode disables the current tab.

💬 Non-blocking

🟡 Warning — The added unit tests cover the pure helpers, but not the page lifecycle that carries most of the risk here. A focused `PersonalPage` test for “mount with zero workspaces, later workspaces become non-empty, nav activation reselects workspace mode” would prevent this regression.

✅ Highlights

The endpoint selection helper is small and easy to test, and `runtimeApi.listRuntimes()` keeps the caller surface unchanged while selecting `/machine-runtimes` only when no slug is active (`packages/dmloop/src/api/runtimeApi.ts:6-8`, `packages/dmloop/src/api/workspaceSelection.ts:23-28`).

packages/dmpersonal/src/PersonalPage.tsx
```diff
@@ -103,9 +116,15 @@ export default function PersonalPage() {
   useEffect(() => {
     const onNavMenuActivated = ({ menuId }: { menuId: string }) => {
       if (menuId !== "dmpersonal") return;
-      const ws = selectedWsRef.current;
-      if (!ws) return; // 尚未就绪:挂载副作用会在加载完成后自行铺视图
-      setWorkspaceContext(ws.slug, ws.id);
+      // 用 ref 判就绪(避免 [] 依赖下的陈旧闭包):machine 模式已就绪但无 selectedWs;
+      // workspace 模式以 selectedWsRef 非空为就绪标志。两者皆未定=加载中,挂载副作用会自铺。
+      if (machineModeRef.current) {
```
OctoBoooot: **[blocker]** Stuck-in-machine-mode. The nav-activation handler only re-reads the cached `machineModeRef.current` — it never re-resolves the mode. Combined with `resolveWorkspaceSelection` running only in the mount effect (`:57-111`, `[]` deps) and PersonalPage staying mounted across nav (`:113` comment), a user who opens Personal with 0 workspaces (machine mode cached), then creates a workspace in Loop (different page, no signal to here), then returns to Personal → stays machine-scoped forever: Skills disabled, Runtime → `/machine-runtimes` not `/runtimes`, until a full reload. Contradicts the PR's stated fall-back-to-workspace behavior. Fix: on activation, re-`listWorkspaces()` / re-resolve the selection (or subscribe to a workspace-change signal), and reset the active tab if machine mode disabled it. Credit Jerry-Xin.

OctoBoooot (changes requested):
## Correcting my #729 APPROVE → REQUEST_CHANGES — stuck-in-machine-mode after creating a workspace (credit Jerry-Xin)

I approved this, praising the `[]`-deps + ref pattern for correctly avoiding the #619 context-pollution/flicker trap. That part is true — but I checked the *pollution* axis and missed the *staleness* axis: the machine/workspace mode is resolved once and never re-derived, so it gets stuck. Jerry-Xin caught it; I byte-verified and he's right. Flipping to REQUEST_CHANGES.

### Blocker — machine mode is cached at mount and never re-resolved → stuck after a workspace is created
- `listWorkspaces()` + `resolveWorkspaceSelection()` run **only** in the mount effect (`PersonalPage.tsx:57-111`, `[]` deps — never re-runs).
- PersonalPage stays mounted across nav (the code comment at `:113` says so: "本页常驻不重挂"); returning to Personal fires `wk:nav-menu-activated`, whose handler (`:116-132`) **only re-reads the cached `machineModeRef.current`** (`:121`) and re-asserts `setWorkspaceContext("","")` — it never re-calls `listWorkspaces`/re-resolves the mode.
- Workspace creation happens on a **different** page (`LoopPage.doCreateWs`) and emits no signal PersonalPage consumes.

Sequence: open Personal with 0 workspaces → machine mode cached (`machineModeRef.current = true`) → create a workspace in Loop → return to Personal → **still machine mode**: Skills tab stays disabled, Runtime keeps hitting `/machine-runtimes` instead of `/runtimes`, until a full page reload. This directly contradicts the PR's stated behavior (fall back to the per-workspace list once a workspace exists). Fix: on Personal activation, re-resolve the selection (re-`listWorkspaces` or subscribe to a workspace-change signal), and reset the active tab if machine mode disabled it.

### Accountability
I verified the workspace-context re-assert (no cross-page pollution — correct) and even praised `[]`-deps+ref for dodging the #619 locale-flicker — but that exact "resolve once, never re-resolve" is what causes the stuck mode. I checked context *pollution* and missed mode *staleness*. Banking: **a ref/`[]`-deps pattern that fixes a re-run problem can create a never-re-derive problem — for cached mode/selection state, verify there's a path that refreshes it when the underlying data changes elsewhere, not just that it's asserted correctly on re-entry.** (Same class as the #721 sticky-armed state I also under-graded — resolved-once state leaking into a later interaction that never re-derives it.)

### Still verified good (unchanged)
No-workspace detection (`len===0`→machine, `runtimeListPath("")`→`/machine-runtimes`); no stale-slug header leak (machine mode clears context, interceptor gates on `if(_workspaceSlug)`); no cross-page global pollution; empty/loading/error runtime states don't crash; the −22 lines swap the "join a workspace" gate for machine mode (no dropped guard); i18n parity intact.

### Non-blocking (carried)
Tests cover only the pure helper, not the page lifecycle that carries the risk — add a "0 → create workspace → nav back → resolves to workspace mode" regression test. `workspaceEmpty` state is now dead code.

### Net
Flipping APPROVE → REQUEST_CHANGES: byte-verified stuck-machine-mode blocker (mode resolved once at mount, never re-derived; PersonalPage doesn't remount and the nav handler reuses the cached ref) — the page stays machine-scoped after a workspace is created, contradicting the PR's own contract. Fix: re-resolve on activation / on a workspace-change signal. Credit Jerry-Xin for the catch; caster-Q — the no-workspace handling itself is otherwise sound.

yujiawei (approved):
# Code Review — PR #729 (octo-web)

**Verdict: APPROVED** — no blocking (P0/P1) frontend defects. This is a clean, well-scoped change; the remaining items are non-blocking hygiene/advisory. One item below is a **human-verify gate** because the feature's tenancy guarantee lives in the backend endpoint this PR depends on (shipped separately).

## Summary of the change

Makes the **My → Runtime** page usable when the user has zero workspaces by introducing a workspace-independent "machine mode":
- New pure helpers `resolveWorkspaceSelection()` / `runtimeListPath()` (unit-tested, 5 cases).
- `listRuntimes()` selects `/machine-runtimes` (auth-only, no workspace slug) when there is no slug, else the workspace-scoped `/runtimes`.
- `PersonalPage` mounts `RuntimePage` in machine mode instead of the "join a workspace" gate, and disables the Skills tab (Skills remain workspace-managed).

## Spec compliance

The diff implements exactly what #728 / the PR description states — no scope creep. Endpoint selection, the machine-mode mount, and the Skills-disabled behavior all match the description. Tests were added for the extracted pure functions.

## Correctness

Data flow verified. Two notable points I checked and cleared:
- **No mount-timing race.** `setWorkspaceContext("", "")` runs synchronously *before* `WKApp.routeRight.replaceToRoot(...)` schedules the `RuntimePage` mount (a React commit). So `currentWorkspaceSlug()` is guaranteed empty when `listRuntimes()` fires → `/machine-runtimes`. Ordering holds by construction, not by luck.
- **Shared-endpoint safety for `listRuntimes()`.** The other consumer that could be affected, `SkillPage`, is only mounted inside `PersonalPage`, and its tab is disabled in machine mode (`PersonalPage.tsx:46,142`). `SkillPage` is **not** mounted inside `LoopPage`, and within Loop the workspace slug is always set non-empty before render, so there is no path where a Loop-side `listRuntimes()` fires against `/machine-runtimes` with an empty slug. `agentApi.listRuntimesForAgent()` is untouched and still uses `/runtimes`.

**[P2 / advisory] Backend tenancy contract is the real guarantee — please human-verify.**
This PR routes a 0-workspace user to `GET /machine-runtimes` (auth-only; `http.ts:32` sends no `x-workspace-slug` when empty, but still sends `token` and `X-Space-Id`). The frontend surface is clean — the endpoint takes no client-controlled selector (no path param, no query, no slug), so there is **no IDOR vector introduced here**. But the actual tenancy boundary (that `/machine-runtimes` returns only the *authenticated owner's* machine runtimes, scoped by the authenticated principal and never by a spoofable header alone) is enforced in the octo-multica backend, which is outside this diff. Given the `security_sensitive` classification, a human should confirm that backend endpoint is owner-scoped before/at merge. `packages/dmloop/src/api/workspaceSelection.ts:26-28`, `packages/dmloop/src/api/http.ts:32-39`

**[P2] Dead code / orphaned i18n key.** After this change `setWorkspaceEmpty` is only ever called with `false` (init `false`; lines 77, 88), so the `workspaceEmpty ||` half of the sidebar-hint guard at `PersonalPage.tsx:150` is unreachable. The `personal.workspace.requiredDesc` i18n key (`i18n/{en-US,zh-CN}.json:12`) is now referenced nowhere. (`requiredTitle` is still used as the hint fallback at line 152, so keep it.) Suggest removing the dead `workspaceEmpty` state and the orphaned key. `packages/dmpersonal/src/PersonalPage.tsx:30,150`

**[P2] Machine-mode does not re-evaluate on workspace-count change (pre-existing pattern).** The workspace fetch is mount-only (`[]` deps) and there is no `workspace-created` event; `nav-menu-activated` only re-asserts the cached `machineModeRef`, it never re-runs `resolveWorkspaceSelection`. So if a user creates their *first* workspace elsewhere and soft-navigates back to this resident page without a reload, it stays in machine mode (Skills tab stays disabled) until reload. This mirrors the page's existing mount-only behavior and only degrades a disabled tab (no data corruption), so it's non-blocking — but worth a follow-up if the create-first-workspace flow is reachable without a reload. `packages/dmpersonal/src/PersonalPage.tsx:57-111,116-132`

**[P2 / nit] `machineMode` state vs `machineModeRef`.** They are kept in sync in both success branches today, so this is safe. Just flagging that the dual source of truth is one edit away from divergence — a future change that sets one but not the other would silently desync render vs. effect logic. `packages/dmpersonal/src/PersonalPage.tsx:34,40,72-73,83-84`

**[P2 / suggestion] Test coverage.** The extracted pure functions are well-covered, but the stateful `PersonalPage` branches (machine-mode fork, `openTab` skill early-return, nav re-assert) have no tests. Optional, not blocking.

## Verdict

The change is correct, scoped, and reduces (not increases) frontend attack surface. Approving. The only item that genuinely warrants a human eye before merge is the backend `/machine-runtimes` owner-scoping contract — everything else is cleanup that can land in a follow-up.

yujiawei (changes requested):
# Code Review — PR #729 (octo-web)

**Correcting my earlier APPROVE → CHANGES_REQUESTED.** I approved after verifying the endpoint selection, the no-stale-slug-header behavior, and the absence of cross-page context pollution — all of which are sound. But I under-rated a functional defect on the *staleness* axis that two other reviewers flagged with byte-level evidence. I re-checked the app's navigation model directly and they are right: the page gets stuck in machine mode after a workspace is created. That contradicts the PR's own stated contract, so it blocks.

**Verdict: CHANGES_REQUESTED** (one P1 functional bug; the rest are non-blocking).

## P1 (blocking) — machine/workspace mode is resolved once at mount and never re-derived → stuck in machine mode after a workspace is created

Verified against the actual navigation model:

- `MainContentLeft` keeps every visited menu route mounted and only toggles `display` — it does **not** remount PersonalPage on nav return (`apps/web/src/Pages/Main/index.tsx:27-33`). The code even documents this ("本页常驻不重挂", `PersonalPage.tsx:113`).
- `resolveWorkspaceSelection()` + `listWorkspaces()` run **only** in the mount effect (`PersonalPage.tsx:57-111`, `[]` deps — never re-runs).
- On return to Personal the app emits `wk:nav-menu-activated` (`apps/web/src/Pages/Main/index.tsx:265`), but PersonalPage's handler only re-reads the cached `machineModeRef.current` and re-asserts `setWorkspaceContext("","")` (`PersonalPage.tsx:116-132`) — it never re-resolves the mode.
- Workspace creation happens on a different page (`LoopPage.doCreateWs`, `packages/dmloop/src/pages/LoopPage.tsx:179-195`) and emits no signal PersonalPage consumes.

**Reproduction (normal soft navigation, not an edge case):** open Personal with 0 workspaces → machine mode is cached → create a workspace from Loop → return to Personal → the page **stays in machine mode**: the Skills tab remains disabled and Runtime keeps calling `/machine-runtimes` instead of the workspace-scoped `/runtimes`, until a full page reload. This directly contradicts the PR description's stated behavior ("after creating a workspace the page falls back to the per-workspace list"). Because PersonalPage is a resident page, soft-nav-back is the *common* path here, not a rare one — which is precisely why the `wk:nav-menu-activated` handler already exists.

**Fix direction:** on Personal activation, re-resolve the selection — either re-`listWorkspaces()` / re-run `resolveWorkspaceSelection()` in the nav-activation handler, or subscribe to a workspace-change signal emitted by `doCreateWs`. Also reset the active tab to `runtime` if machine mode disables the currently-selected tab.

## Also verified GOOD (unchanged by this correction)

- **No IDOR / no client-controlled selector.** `/machine-runtimes` takes no path param, query, or slug (`workspaceSelection.ts:26-28`); `http.ts:32` sends `x-workspace-slug` only when non-empty. Frontend attack surface is reduced, not increased.
- **No stale-slug header leak.** Machine mode clears the context and the interceptor gates on `if (_workspaceSlug)`.
- **No mount-timing race.** `setWorkspaceContext("","")` runs synchronously before `replaceToRoot(...)` schedules the RuntimePage mount, so `listRuntimes()` sees an empty slug by construction.
- **Shared `listRuntimes()` is safe.** The other consumer, `SkillPage`, is only mounted inside PersonalPage and its tab is disabled in machine mode; it is not mounted in LoopPage. `agentApi.listRuntimesForAgent()` is untouched.
- i18n parity intact; empty/loading/error runtime states don't crash; the −22 lines swap the "join a workspace" gate for machine mode with no dropped guard.

## Security note for the human reviewer (this PR is classified security-sensitive)

The frontend surface is clean, but the actual tenancy guarantee — that `GET /machine-runtimes` returns only the *authenticated owner's* machine runtimes, scoped by the authenticated principal and never by a spoofable header (`X-Space-Id` is still sent unconditionally, `http.ts:38-39`) — lives in the octo-multica backend, outside this diff. Please confirm that endpoint is owner-scoped before merge.

## Non-blocking (P2)

- **Dead code / orphaned i18n.** `setWorkspaceEmpty` is now only ever called with `false`, so the `workspaceEmpty ||` branch at `PersonalPage.tsx:150` is unreachable; the `personal.workspace.requiredDesc` key (`i18n/{en-US,zh-CN}.json:12`) is now unused. (`requiredTitle` is still used at line 152 — keep it.) Suggest removing the dead state and key.
- **`machineMode` state vs `machineModeRef`.** Kept in sync today, but the dual source of truth is one edit away from divergence — worth consolidating.
- **Test coverage.** The pure helpers are well-tested; a `PersonalPage` lifecycle test ("0 workspaces → create workspace → nav back → resolves to workspace mode") would have caught the P1 above and prevents regression.

## Verdict

Requesting changes for the single P1: re-resolve machine/workspace mode on Personal activation (or on a workspace-change signal) so the page honors its stated fallback-to-workspace contract without requiring a reload. The no-workspace handling itself is otherwise sound.

caster-Q:
Thanks @Jerry-Xin (and @OctoBoooot for the follow-up) — the stuck-machine-mode blocker is fixed in `f375ffdc`.

**What changed (`packages/dmpersonal/src/PersonalPage.tsx`):**
- The machine/workspace mode is no longer resolved only at mount. Extracted the "list workspaces → `resolveWorkspaceSelection` → set mode → paint" logic into one function that runs both on mount **and on every `wk:nav-menu-activated` for `dmpersonal`**. Returning to Personal now re-calls `listWorkspaces()` and re-derives the mode instead of reusing the cached `machineModeRef` / `selectedWsRef`.
- Sequence now works: open Personal with 0 workspaces (machine mode) → create a workspace in Loop → return to Personal → flips to workspace mode: Runtime hits `/runtimes`, Skills is re-enabled — no full reload needed.
- If machine mode had disabled the active tab (Skills), it now falls back to Runtime on re-resolve.
- Removed the now-dead `workspaceEmpty` state (only the error hint remains).

**Non-blocking test note:** `packages/dmpersonal` has no component-test harness (jsdom / testing-library) today; the pure helpers (`resolveWorkspaceSelection` / `runtimeListPath`) are unit-tested. A `PersonalPage` lifecycle test ("0 workspaces → workspace created elsewhere → nav activation reselects workspace mode") would need that harness added first — happy to do it as a follow-up if we want to introduce it here.

OctoBoooot (dismissed):
## Review: #729 @ `f375ffdc` — DELTA re-review; stuck-machine-mode blocker fixed, flipping to APPROVE

**Verdict:** Approve. This commit fixes exactly the blocker my/the three-way CR held on — the nav-activation handler now re-resolves the workspace mode instead of reusing the cached ref. Byte-verified. Flipping REQUEST_CHANGES → APPROVE. Clean single-commit fast-forward (only `PersonalPage.tsx`). Build green; `code-review` pending.

### Resolved — machine/workspace mode now re-resolves on every Personal activation
The stuck-mode root cause (mode resolved once at mount, nav-return re-read the cached `machineModeRef`) is fixed:
- A shared `resolveAndPaint()` (`:45-90`) **re-runs `listWorkspaces()` + `resolveWorkspaceSelection()`** every call (`:51-54`), re-deciding machine↔workspace each time.
- The `wk:nav-menu-activated` handler now calls `resolveRef.current(false)` (`:126`) — so returning to Personal after creating a workspace in Loop re-lists, sees it, and flips machine→workspace (Skills re-enable, Runtime → `/runtimes`). The stuck-until-reload behavior is gone.
- **Tab reset handled** (`:63-66`): if machine mode disables Skills while the user is on that tab, it falls back to `runtime` — closes the secondary gap I flagged.

### Verified — the fix avoids re-introducing the two traps it sits between
- **No stale closure**: `resolveRef` (`:93-94`) always points at the latest `resolveAndPaint`, so the `[]`-dep effects (`:117`, `:130`) invoke current logic — the exact ref-indirection the earlier version got right, now correctly driving a re-resolve.
- **No locale flicker**: it deliberately does NOT add `[t]` to the effect deps (mount-once, re-resolve only on activation), so a language switch doesn't re-fetch/reset — the #727 trap, avoided. The `:91-92` comment reasons through this tension explicitly. This is the correct resolution of the very "re-resolve without re-running on locale" bind I was worried about.
- **`mountedRef`** guard (`:53/:78`) prevents setState-after-unmount in the async path.

### Also addressed
- The dead `workspaceEmpty` state (my carried minor) is removed.

### Carried (non-blocking)
- Tests still cover only the pure helper, not the page lifecycle — the "0 → create workspace → nav back → resolves to workspace mode" regression path (now the load-bearing logic) would be worth a test, but it's hard to unit-test and consistent with the existing page.
- yujiawei's `/machine-runtimes` backend tenant-isolation question stands as a human/backend confirmation item (out of scope for this frontend diff).

### Net
The stuck-machine-mode data-staleness blocker is correctly fixed — re-resolve on activation, tab reset, and both the stale-closure and locale-flicker pitfalls handled. Byte-verified. Approving (then the size/M `needs-human-review` sign-off). Credit caster-Q for the fix; Jerry-Xin/yujiawei for the finding.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #729 (octo-web)

## Summary

This PR adds machine-level runtime support to the Personal page when the user has zero workspaces. Instead of showing a "workspace required" dead-end, the page enters "machine mode": it clears the workspace scope (no `x-workspace-slug` header), routes the runtime list through `/machine-runtimes` (auth-only endpoint), and disables the Skills tab (which requires a workspace). The core logic is extracted into a pure `resolveWorkspaceSelection` helper with unit tests, and the Personal page re-evaluates mode on every navigation re-activation to handle workspaces being created or deleted while the page stays mounted.

Overall the design is sound — the state machine is well-thought-out, the `resolveRef` pattern correctly avoids stale closures without re-mounting effects on language change, and the unit tests cover the key decision paths.

## Verification

- ✅ **Data flow: workspace slug → endpoint routing** — `currentWorkspaceSlug()` returns module-level `_workspaceSlug` (set by `setWorkspaceContext`). In machine mode, `setWorkspaceContext("", "")` clears it to `""`. Then `runtimeListPath("")` → `"/machine-runtimes"`. The axios interceptor at `http.ts:38` correctly omits `x-workspace-slug` when `_workspaceSlug` is falsy. Chain verified end-to-end.
- ✅ **Skill tab disabled in machine mode** — three layers of protection: (1) button `disabled` attribute at `PersonalPage.tsx:131`, (2) early return in `openTab` at `:103`, (3) forced tab fallback to "runtime" in `resolveAndPaint` at `:69`. No path can render `SkillPage` when `machineMode` is true.
- ✅ **Nav re-activation re-evaluates mode** — the `onNavMenuActivated` handler calls `resolveRef.current(false)`, which always re-fetches `listWorkspaces()`. This correctly handles the case where a workspace was created in Loop since the last evaluation (the bug that motivated commit `f375ffdc`).
- ✅ **Unmount safety** — `mountedRef` replaces the `cancelled` flag; both `.then` and `.catch` check `mountedRef.current` before setting state.
- ✅ **`resolveWorkspaceSelection` logic** — handles null/empty array → machine mode; finds current workspace by id or falls back to first. Unit tests cover all three branches plus both `runtimeListPath` outcomes.
- ✅ **Export wiring** — `resolveWorkspaceSelection`, `runtimeListPath`, and `WorkspaceSelection` are correctly re-exported from `packages/dmloop/src/index.tsx:30-31` and consumed by `PersonalPage.tsx`.

Static analysis only at head `f375ffdc0d038f1a347cce565d6d67ce1c45c3bd`; build and tests not executed in this environment.

## Findings

### P2 — Race condition in `resolveAndPaint` on rapid re-activations (`packages/dmpersonal/src/PersonalPage.tsx:57`)

`resolveAndPaint` fires a new `listWorkspaces()` call each time it is invoked (mount with `showLoading=true`, or nav re-activation with `showLoading=false`), but does not cancel the prior in-flight request. Two rapid re-activations produce overlapping async calls whose `.then` handlers resolve in arbitrary order. The later-fired call may resolve first, then be overwritten by the earlier (stale) call's response, resulting in incorrect mode/state.

`mountedRef` guards unmount but not re-entrance. Consider adding a monotonic call-id:

```ts
let callId = 0;
const resolveAndPaint = (showLoading: boolean) => {
  const thisCall = ++callId;
  // ...
  workspaceApi.listWorkspaces().then((workspaces) => {
    if (!mountedRef.current || thisCall !== callId) return;
    // ...
  });
};
```

### P2 — Error path does not reset `machineMode` state (`packages/dmpersonal/src/PersonalPage.tsx:87`)

The `.catch` branch sets `workspaceError` and `workspaceReady(false)` but does not reset `machineMode` / `machineModeRef`. If a previous successful call established machine mode and a subsequent re-activation fails (transient network error), `machineMode` remains `true` in both state and ref. While tabs are disabled by `!workspaceReady`, the stale mode state could produce incorrect behavior on the next successful re-activation if the workspace list changed during the error window (e.g. workspaces were created elsewhere).

Consider resetting both flags in the catch handler:

```ts
.catch((error) => {
  if (!mountedRef.current) return;
  machineModeRef.current = false;
  setMachineMode(false);
  // ... existing error handling
});
```

## Data Flow Trace

- **`listWorkspaces()` result** → `resolveWorkspaceSelection(workspaces, currentWorkspaceId())` → when 0 workspaces: `{ mode: "machine" }` → `setWorkspaceContext("", "")` → `_workspaceSlug = ""` → `runtimeListPath("")` = `"/machine-runtimes"` → `httpGet("/machine-runtimes")` with no `x-workspace-slug` header. Data flows end-to-end. ✅
- **`listWorkspaces()` result** → when ≥1 workspace: `resolveWorkspaceSelection` returns `{ mode: "workspace", slug, id }` → `setWorkspaceContext(slug, id)` → `runtimeListPath(slug)` = `"/runtimes"` → `httpGet("/runtimes")` with `x-workspace-slug` header. ✅
- **`currentWorkspaceId()`** returns `_workspaceId` (module-level). In machine mode, this is `""`. Passed to `resolveWorkspaceSelection` — no workspace will match `""`, but the fallback logic (`find(...) ?? workspaces[0]`) only runs when `workspaces.length > 0`, which is the workspace-mode branch. ✅
- **`tabRef.current` mutation** at `:68-69` (machine mode skill→runtime fallback) — directly mutates the ref before the `setTab("runtime")` state update. This is intentional for synchronous reads in `openTab` and the render path. The next render syncs `tabRef.current = tab` at `:45`. ✅

- **C1 (Dual-path parity)**: ✅ Machine↔workspace mode transitions are symmetric — `resolveAndPaint` handles both directions, `onNavMenuActivated` always re-evaluates. `openTab` correctly distinguishes both modes.
- **C2 (Control-flow ordering)**: ✅ `resolveAndPaint` is the single entry point for workspace resolution. `resolveRef.current` ensures the latest closure is always called. No nested/duplicate call sites.
- **C3 (Authorization boundary)**: N/A — the `/machine-runtimes` endpoint's server-side authorization is outside this repo. The client correctly omits `x-workspace-slug` and sends auth headers (`token`, `X-Space-Id`).
- **C4 (Auth lifecycle)**: ✅ `setWorkspaceContext("", "")` correctly clears workspace scope on machine mode entry. Re-activation always re-evaluates (no cached judgment).
- **C5 (Build/note ≠ runtime)**: N/A — no build artifacts, packaging, or deployment changes.
- **C6 (Governance/docs)**: N/A — no governance or policy changes.

- `getRuntime(id)` at `runtimeApi.ts:15` still uses hardcoded `/runtimes/:id` — but this function has zero call sites in the entire codebase (dead export). No runtime impact.
- `agentApi.ts` `runtimeMap()` and `listRuntimesForAgent()` use hardcoded `/runtimes` — but agent pages are only accessible through the Loop module (requires workspace context), not reachable in machine mode from Personal.
- `SkillPage.tsx` calls `listRuntimes()` — but the Skill tab is disabled in machine mode (three layers of protection), so this code path is unreachable.
- Test file `workspaceSelection.test.ts` covers: empty array → machine, current match, absent-id fallback, and both `runtimeListPath` outcomes.

## Verdict: APPROVED

Well-designed feature addition with clean separation of concerns (pure helper + unit tests + UI adaptation). The machine-mode state machine is correct, the data flow from workspace resolution to API endpoint selection is sound, and the unmount/closure-safety patterns are well-implemented. Two P2 items (async race in re-activation, stale mode on error) are non-blocking — the race is low-probability and the error-path staleness is mitigated by disabled tabs.

[Octo-Q] verdict: APPROVE — zero P0/P1 findings; two P2 items (async race condition in resolveAndPaint, stale machineMode on error path) are non-blocking.

yujiawei (changes requested):
# Code Review — PR #729 (octo-web)

Reviewed at head SHA `f375ffdc0d038f1a347cce565d6d67ce1c45c3bd` against merge-base `d6190bed`. Scope: 5 files, +132/−51. Verified locally: `pnpm --filter @octo/loop test` → 22 passed (incl. the 5 new `workspaceSelection`/`runtimeListPath` cases); `pnpm i18n:check` → passed.

This review has two gates: **spec compliance** first, then **code quality**. Both must pass to approve.

## 1. Spec compliance

Traced against issue #728 (machine-level runtime mode).

- **Missing:** none. `resolveWorkspaceSelection()` + `runtimeListPath()` are added and unit-tested; `runtimeApi.listRuntimes()` selects the endpoint via `runtimeListPath(currentWorkspaceSlug())`; `PersonalPage` mounts `RuntimePage` in machine mode on the 0-workspace path; the Skills tab is disabled in machine mode; the workspace path falls back to the existing per-workspace behavior.
- **Over-building:** none. No new route, no new user-visible entry point, no new flags/fields/endpoints beyond the documented `/machine-runtimes` switch. Re-resolving mode on Personal re-activation is in scope (it is the fix for "created a workspace but still stuck in machine mode").
- **Deviation:** none against the stated approach.

**Spec: ✅**

## 2. Code quality

### P1 — Re-activation re-introduces the #619 cross-page workspace pollution

`packages/dmpersonal/src/PersonalPage.tsx:123-127` — the `wk:nav-menu-activated` handler now calls `resolveRef.current(false)`, and `resolveAndPaint` derives the selection from the **shared module-level global** via `resolveWorkspaceSelection(workspaces, currentWorkspaceId())` (line 54). Because `_workspaceId` in `@octo/loop/api/http.ts` is a singleton shared with the Loop page, this makes Personal *follow* whatever workspace Loop last selected, on every re-activation.

Repro: open Personal (settles on workspace A → `selectedWsRef = A`) → open Loop, switch to workspace B (mutates the shared global to B) → click the Personal nav again → Personal now renders B's runtimes/skills instead of A.

This is exactly the bug the previous #619 fix removed. The old handler re-asserted Personal's own stored selection:

```tsx
const ws = selectedWsRef.current;
if (!ws) return;
setWorkspaceContext(ws.slug, ws.id);
```

The in-file comment at lines 34-35 still documents that contract — "存下本页选定的 workspace … 避免被 Loop 的选择污染(#619 评审)" — so the code now contradicts its own comment, which also makes this a maintainability trap.

The mode re-evaluation that #729 needs does **not** require abandoning Personal's own workspace selection. Prefer the local ref and fall back to the global only when it's gone (e.g. the workspace was deleted elsewhere):

```tsx
const targetId = selectedWsRef.current?.id ?? currentWorkspaceId();
const selection = resolveWorkspaceSelection(workspaces, targetId);
```

That preserves both #729 (machine↔workspace re-evaluation on every activation) and #619 (isolation from Loop's selection).

### P2 — No sequencing guard for concurrent `listWorkspaces()` fetches

`packages/dmpersonal/src/PersonalPage.tsx:51,77` — `resolveAndPaint` is now driven by an event listener, so rapid Personal-nav clicks can start overlapping `listWorkspaces()` requests. `mountedRef` only guards against post-unmount writes, not out-of-order resolution: a slow earlier response can land after a newer one and repaint stale mode/selection. A per-call request-generation token (increment on entry, ignore the result if the token is stale) would make the last click win deterministically.

### P2 — Blank right pane during re-activation refetch

`packages/dmpersonal/src/PersonalPage.tsx:126` — on nav re-activation the framework calls `routeRight.popToRoot()` (empties the pane) and then this handler runs `resolveAndPaint(false)`, which intentionally skips the loading placeholder. Between `popToRoot` and the `listWorkspaces()` resolution the right pane is empty. Passing `true` here (or painting a lightweight placeholder before awaiting) would avoid the flash; the mount path already shows loading.

### P2 (nit) — Dead i18n keys

`packages/dmpersonal/src/i18n/{zh-CN,en-US}.json:11-12` — `personal.workspace.requiredTitle` / `requiredDesc` are no longer referenced after the join-a-workspace gate was removed (`i18n:check` still passes because the keys exist, not because they're used). Safe to delete.

### Not an issue (adjudicating a raised concern)

A concern was raised that mutating refs during render (`tabRef.current = tab`, `resolveRef.current = resolveAndPaint`) violates React 18 Concurrent Mode rules. This repo pins **React 17.0.2** (`packages/dmpersonal/package.json`), which has no concurrent rendering, and `tabRef.current = tab` pre-dates this PR. Not applicable here — no change requested on this point.

## 3. Security note (this PR was flagged security-sensitive)

This is a **frontend-only** change. Machine-mode isolation depends entirely on the backend: `/machine-runtimes` must be strictly owner-scoped (auth-only, returning only the caller's own machine runtimes) and `/runtimes` must reject requests with no `x-workspace-slug`. The frontend correctly omits `x-workspace-slug` in machine mode (`http.ts` only sends the header when the slug is non-empty) and correctly routes to `/machine-runtimes`. The owner-scoping of `GET /api/machine-runtimes` is out of this diff's reach (ships in octo-multica) and should be verified server-side by a human before relying on this page for a 0-workspace user. The headless PAT minting flow is unchanged and unaffected by machine mode.

## 4. Suggested changes

1. In `resolveAndPaint`, prefer `selectedWsRef.current?.id ?? currentWorkspaceId()` so Personal keeps its own workspace selection across Loop navigation (fixes the P1 / restores #619).
2. Add a request-generation token to make the last re-activation win.
3. Pass `showLoading: true` on the re-activation call (or paint a placeholder) to avoid the blank-pane flash.
4. Remove the two unused `personal.workspace.required*` i18n keys.

## 5. Additional observations

- Cross-module check: clearing the shared global to `("", "")` in machine mode does not harmfully bleed into Loop — `LoopPage` re-lists workspaces and calls `applyWorkspace` (with `invalidateDirectory`/`invalidateRuntimeMap`) on its own activation, so it re-establishes its own context.
- `resolveWorkspaceSelection` / `runtimeListPath` are cleanly extracted pure functions with good unit coverage — nice.

## Overall verdict

**REQUEST_CHANGES** — Spec ✅, but the P1 re-activation regression re-opens the previously-fixed #619 cross-page workspace pollution and must be addressed before merge. The remaining items are P2/nit.

Note: this review reflects one independent human pass plus a cross-family advisory pass (which independently surfaced the P1). A second automated advisory leg was unavailable this run (upstream gateway 503) and is therefore recorded as **absent**, not as agreement.

Jerry-Xin (dismissed):
The stuck-machine-mode blocker from my prior review is fixed at this head. Re-verified against the new commit; flipping to approve.

✅ Blocker resolved (stuck machine mode after creating a workspace elsewhere)

The mode is now re-derived on Personal re-activation instead of re-asserting a cached flag.

- `packages/dmpersonal/src/PersonalPage.tsx`: the resolve logic is extracted into `resolveAndPaint(showLoading)`, which on every call runs `listWorkspaces()` → `resolveWorkspaceSelection(...)` and updates both `machineModeRef`/`setMachineMode` and `selectedWsRef`/`setWorkspaceContext(...)` (machine branch and workspace branch symmetrically).
- The `wk:nav-menu-activated` handler now calls `resolveRef.current(false)` rather than only re-reading the cached `machineModeRef.current`. A `resolveRef` always points at the latest closure, so the `[]`-deps effect avoids the stale-closure trap.
- Repro that was broken now passes: open Personal at 0 workspaces (machine mode) → create a workspace in Loop → return to Personal → nav activation re-runs `listWorkspaces()`, `resolveWorkspaceSelection` returns `mode:"workspace"`, so `machineModeRef` clears, Skills re-enables, and the runtime list switches off the machine-level endpoint.

✅ N→0 transition also handled

Because `resolveWorkspaceSelection([], _)` returns `mode:"machine"`, deleting the last workspace elsewhere and returning to Personal re-resolves back to machine mode. The new fallback (if the current tab is `skill`, drop back to `runtime`) prevents getting stuck on the now-disabled Skills tab.

✅ No new flicker

The nav-activation path calls `resolveAndPaint(false)`, so it does not repaint the loading spinner on re-activation; the right panel is only replaced once `listWorkspaces()` resolves. The mount path keeps `showLoading=true`. This preserves the earlier no-flicker behavior.

💬 Non-blocking

🔵 The now-unused `personal.workspace.requiredTitle` / `requiredDesc` i18n keys remain in `packages/dmpersonal/src/i18n/{zh-CN,en-US}.json`. Harmless, but could be pruned since the "join a workspace first" hint is no longer rendered.

caster-Q:
Thanks @yujiawei — all four points addressed in `15154261`.

**P1 — re-activation no longer follows Loop's selection (restores #619 isolation).** `resolveAndPaint` now resolves against Personal's own stored selection first, falling back to the shared global only when it's gone:
```tsx
const targetId = selectedWsRef.current?.id ?? currentWorkspaceId();
const selection = resolveWorkspaceSelection(workspaces, targetId);
```
So returning to Personal after switching workspaces in Loop keeps Personal on its own workspace, while still re-deriving machine↔workspace mode on every activation (the #729 fix).

**P2 — concurrent-fetch sequencing.** Added a per-call generation token (`resolveSeqRef`): each `resolveAndPaint` increments it on entry and drops its result if a newer call started, so the last activation wins deterministically.

**P2 — blank pane on re-activation.** The nav handler now calls `resolveRef.current(true)`, so the loading placeholder is painted before the `listWorkspaces()` fetch resolves — no empty gap after `popToRoot`.

**P2 (nit) — dead i18n keys.** Removed `personal.workspace.requiredTitle` / `requiredDesc` from both `zh-CN.json` and `en-US.json` (`pnpm i18n:check` still passes).

**Security note:** confirmed on the backend side — `GET /api/machine-runtimes` is owner-scoped (`ListMachineAgentRuntimesByOwner(userID)`, auth-only, returns only the caller's own `workspace_id IS NULL` runtimes) and `/runtimes` stays behind the workspace-member middleware. That backend already merged in octo-multica. The frontend correctly omits `x-workspace-slug` in machine mode.

OctoBoooot (approved):
## Review: #729 @ `15154261` — DELTA re-review; cross-page pollution + concurrent-fetch race both fixed, APPROVE holds

**Verdict:** Approve. Clean single-commit fast-forward. This commit closes the two robustness items raised on the fixed head (cross-page pollution + concurrent-`listWorkspaces` race) — the same ones yujiawei escalated. My prior APPROVE (@`f375ffdc`) was auto-dismissed by this commit; re-affirming at the current head. Risk tier: normal. Build + title-lint green; `code-review` pending.

### Resolved — cross-page workspace pollution on re-resolve (#619 class)
`resolveAndPaint` now keys off Personal's **own** stored selection, not the shared Loop global: `const targetId = selectedWsRef.current?.id ?? currentWorkspaceId()` (`PersonalPage.tsx:60`) — prefer the page's own workspace, fall back to the global only when its own is gone. The comment (`:58-59`) cites #619 explicitly: reading `currentWorkspaceId()` directly would make Personal follow Loop's last-selected workspace. So the re-resolve no longer inherits Loop's context — the pollution window is closed.

### Resolved — concurrent listWorkspaces race
`resolveSeqRef` latest-wins guard: `const seq = ++resolveSeqRef.current` captured before the fetch (`:49`), re-checked after in **both** `.then` (`:57 seq !== resolveSeqRef.current → return`) and `.catch` (`:85`). A slow older response can no longer overwrite a newer activation. Complete seqRef pattern (matches the established discipline elsewhere in the codebase).

### Also
- Dead i18n keys removed symmetrically in both locales (`requiredTitle`/`requiredDesc` — the obsolete "join a workspace first" gate strings, correctly gone now that machine mode replaced that gate). `mountedRef` unmount guard retained.

### Verified good (carried, from `f375ffdc`)
The stuck-machine-mode blocker fix (re-resolve on activation via `listWorkspaces`+`resolveWorkspaceSelection`) remains; tab-reset when machine mode disables Skills; `[]`-deps + `resolveRef` (no stale closure, no locale flicker).

### Net
Both robustness items on the fixed head are correctly resolved — pollution keyed to Personal's own selection (not Loop's global), and a full seqRef guard on concurrent fetches. The stuck-mode blocker stays fixed. Byte-verified. Approving (then the size/M `needs-human-review` sign-off). Credit caster-Q for the fixes; yujiawei/Jerry-Xin for driving the finding across rounds.

Jerry-Xin (approved):
Re-reviewed at the new head `15154261cfa3` (my prior approve was auto-dismissed by this commit). The delta over the previous head is a clean linear single commit hardening the stuck-machine-mode fix. Re-approving.

This commit addresses the two concurrency/pollution follow-ups and drops dead i18n keys. All three claims byte-verified at this head, and the original stuck-mode fix is confirmed intact.

## Verified

**1. Concurrent-fetch guard is real, not cosmetic**
`resolveSeqRef` is a monotonically-increasing sequence captured per call (`const seq = ++resolveSeqRef.current`) before the async `listWorkspaces()`. Both `.then` and `.catch` short-circuit with `if (!mountedRef.current || seq !== resolveSeqRef.current) return`. Overlap trace: slow call A (seq=1) and fast call B (seq=2) — B paints (`2===2`), A's late resolution is discarded (`1 !== 2`). Last activation wins; the unmount guard is preserved via the OR condition.
- `packages/dmpersonal/src/PersonalPage.tsx:43,49,57,85`

**2. Cross-page pollution fixed, stuck-mode NOT re-broken**
Resolve now prefers Personal's own stored workspace: `const targetId = selectedWsRef.current?.id ?? currentWorkspaceId()`, instead of always reading the shared global.
- Pollution repro: Personal on W_p, user selects W_loop in Loop, returns to Personal → `targetId = W_p` → stays on W_p. Fixed.
- Stuck-mode repro (regression check): in machine mode `selectedWsRef.current` is set to `null`, so after a workspace is created elsewhere `targetId` falls back to the global and `resolveWorkspaceSelection` returns `{mode:"workspace"}` → correctly flips machine→workspace. Still fixed.
- Deleted-own-workspace edge: `resolveWorkspaceSelection` falls back to `workspaces[0]` when the stored id is absent, or `{mode:"machine"}` when the list is empty — no stale/crash.
- `packages/dmpersonal/src/PersonalPage.tsx:60-61`; `packages/dmloop/src/api/workspaceSelection.ts`

**3. Dead i18n keys genuinely unused**
`workspace.requiredTitle` / `workspace.requiredDesc` have zero references across the dmpersonal package; only `workspace.loading` and `workspace.loadFailed` are used, both still present with real values in both locales. Both JSON files remain valid.
- `packages/dmpersonal/src/i18n/en-US.json`, `packages/dmpersonal/src/i18n/zh-CN.json`

## Notes (non-blocking)
- The nav-activation handler switched `resolveRef.current(false)` → `resolveRef.current(true)` so the loading placeholder shows during the async re-resolve instead of a blank right pane. Consistent with the mount effect. Benign UX improvement.

## CI
All required checks green, including `pr-title-lint`, build, code-review, osv/secret scans. Not a draft.

LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #729 (octo-web)

## Summary

This PR adds machine-level runtime support to the Personal page: when a user has zero workspaces, instead of blocking with a "join a workspace first" prompt, the page enters "machine mode" — mounting the runtime page against an auth-only `/machine-runtimes` endpoint while disabling the workspace-dependent Skills tab. The core decision logic is extracted into `workspaceSelection.ts` as pure, testable functions. Concurrency guards (sequence counter + mounted ref) protect against stale responses from rapid navigation. The approach is clean and well-reasoned.

## Verification

- ✅ **Data-flow: endpoint selection** — `runtimeListPath(currentWorkspaceSlug())` at `packages/dmloop/src/api/runtimeApi.ts:7` reads the module-level `_workspaceSlug` set by `setWorkspaceContext`. In machine mode, `setWorkspaceContext("", "")` clears it → `runtimeListPath("")` returns `/machine-runtimes` → interceptor at `packages/dmloop/src/api/http.ts:41` skips `x-workspace-slug` header (only added when truthy). Chain verified end-to-end.
- ✅ **Data-flow: workspace resolution** — `resolveAndPaint` in `packages/dmpersonal/src/PersonalPage.tsx:55` calls `workspaceApi.listWorkspaces()` (returns `Promise<Workspace[]>`) → passes result to `resolveWorkspaceSelection(workspaces, targetId)` → correctly selects current/first workspace or falls to machine mode when empty. `targetId` uses `selectedWsRef.current?.id` (Personal's own workspace) before falling back to `currentWorkspaceId()` (shared global) — avoids Loop→Personal pollution.
- ✅ **Concurrency safety** — `resolveSeqRef` sequence counter + `mountedRef` correctly discard stale `listWorkspaces` responses from rapid nav switches. Each call increments seq; only the latest is accepted. Cleanup on unmount.
- ✅ **Cross-page context isolation** — Loop and Personal share module-level `_workspaceSlug`/`_workspaceId`. When Personal enters machine mode it clears the global, but Loop's `onNavMenuActivated` handler at `packages/dmloop/src/pages/LoopPage.tsx:148` re-applies its own `wsId` state via `applyWorkspace` before re-rendering — verified safe.
- ✅ **Tab disable in machine mode** — Skill button disabled via `disabled={!workspaceReady || (machineMode && item.key === "skill")}` at `:148`; `openTab` double-guards with `if (machineModeRef.current && key === "skill") return` at `:113`; if user was on skill tab when entering machine mode, it auto-falls back to runtime at `:74-77`. Triple protection verified.
- ✅ **i18n key removal** — `personal.workspace.requiredTitle` and `personal.workspace.requiredDesc` removed from both `en-US.json` and `zh-CN.json`; grep confirms zero remaining references in the codebase.
- ✅ **Test coverage** — `workspaceSelection.test.ts` covers: empty array → machine, current ID match, fallback to first workspace, slug-present path, empty-slug path.

Static analysis only at head `15154261cfa345be5979c4c69a1890b25e499a20`; build and tests not executed in this environment.

## Findings

### P2 — `getRuntime(id)` does not account for machine mode (`packages/dmloop/src/api/runtimeApi.ts:15`)

`getRuntime(id)` still hardcodes `/runtimes/${id}` regardless of workspace context. No active caller exists in the codebase today, but if a future consumer invokes this in machine mode (empty `x-workspace-slug`), the workspace middleware will 403 it. Consider routing through a workspace-aware path or adding a sibling `getMachineRuntime` when machine-mode detail fetch is needed.

### Nit — Defensive null guard untested (`packages/dmloop/src/api/__tests__/workspaceSelection.test.ts:10`)

`resolveWorkspaceSelection` has a defensive `!workspaces` guard at `:16` of `workspaceSelection.ts`, but no test exercises a null/undefined input. TypeScript's strict mode makes this unlikely in practice, but the guard exists for a reason — a one-line test would lock it.

## Human-verify

1. **`/machine-runtimes` backend endpoint existence and auth scope** — The frontend assumes this endpoint exists and returns machine-level runtimes scoped to the authenticated user (via `token` + `X-Space-Id` headers, without `x-workspace-slug`). If the endpoint is missing, 0-workspace users will see "Failed to load workspaces" instead of the runtime page. If it lacks per-user scoping, one user could see another user's machine-level runtimes. Please confirm the backend contract. This is **not** a merge blocker from the frontend side — flagging for backend verification.

- `RuntimePage` mounts with a fresh `key="runtime"` on every `replaceToRoot` call, so the `useEffect([], [])` in RuntimePage re-fires with the correct workspace context each time.
- `listRuntimesForAgent()` in `agentApi.ts` is untouched and only called from `AgentPage` (Loop section, which always requires a workspace) — no machine-mode leak.
- `SkillPage` also calls `listRuntimes()` but is never rendered in machine mode (triple guard confirmed above).
- The `resolveRef.current = resolveAndPaint` pattern correctly avoids stale closures in the `[]`-dep useEffect while keeping mount-once behavior (no flash on language switch).
- Removed i18n keys have zero dangling references across the entire repo.

## Verdict: APPROVED

The PR is well-structured with clean separation of concerns (pure functions in `workspaceSelection.ts`, state machine in `PersonalPage.tsx`). Concurrency handling is solid. The two findings are non-blocking: `getRuntime` has no active caller, and the missing null test is a defensive-coverage nit. The Human-verify item on the backend `/machine-runtimes` endpoint should be confirmed before merge but is not a frontend code defect.

yujiawei (approved):
# Code Review — PR #729 (octo-web)

Reviewed at head SHA `15154261cfa345be5979c4c69a1890b25e499a20` against merge-base `d6190bed`. Scope: 5 files (+132/−51). Verified locally: `pnpm --filter @octo/loop test` → 22 passed (incl. the 5 new `workspaceSelection`/`runtimeListPath` cases); `pnpm i18n:check` → passed.

This is a re-review. My prior pass requested changes on four items; all four are addressed in this head and independently re-verified below. Two gates: spec compliance first, then code quality.

## 1. Spec compliance

Traced against issue #728 (make the My → Runtime page usable with zero workspaces via a machine-level mode).

- **Missing:** none. `resolveWorkspaceSelection()` + `runtimeListPath()` are added and unit-tested; `runtimeApi.listRuntimes()` selects the endpoint via `runtimeListPath(currentWorkspaceSlug())`; `PersonalPage` mounts `RuntimePage` in machine mode on the 0-workspace path; the Skills tab is disabled in machine mode; the workspace path retains the existing per-workspace behavior.
- **Over-building:** none. No new route, no new user-visible entry point, no new flags/fields beyond the documented `/machine-runtimes` switch. Re-resolving mode on Personal re-activation is in scope (it is the fix for "created a workspace but still stuck in machine mode").
- **Deviation:** none against the stated approach.

**Spec: ✅**

## 2. Code quality

**Quality: Approved**

The four prior blocking/advisory items are resolved. Verified against the code at this head:

- **[Resolved, was P1] #619 cross-page pollution.** `resolveAndPaint` now resolves against Personal's own stored selection first (`packages/dmpersonal/src/PersonalPage.tsx:60` — `const targetId = selectedWsRef.current?.id ?? currentWorkspaceId()`), falling back to the shared global only when its own selection is gone. `selectedWsRef` is written exclusively inside `PersonalPage`, so Loop's selection can no longer drive Personal's re-resolution. This preserves both the #729 machine↔workspace re-evaluation and the #619 isolation.
- **[Resolved, was P2] Concurrent-fetch sequencing.** A per-call generation token (`PersonalPage.tsx:49` `const seq = ++resolveSeqRef.current`) is checked in both the `.then` and `.catch` (`:57`, `:85`), so a slow earlier `listWorkspaces()` response can no longer repaint over a newer one. Last activation wins.
- **[Resolved, was P2] Blank pane on re-activation.** The `wk:nav-menu-activated` handler now calls `resolveRef.current(true)` (`PersonalPage.tsx:134`), painting the loading placeholder synchronously before the fetch resolves, so there is no empty gap after the framework's `popToRoot`.
- **[Resolved, was P2 nit] Dead i18n keys.** `personal.workspace.requiredTitle` / `requiredDesc` are removed from both `zh-CN.json` and `en-US.json`, with no remaining references anywhere in the tree.

### P2 (non-blocking) — `openTab` is not re-gated during an in-flight re-resolve

`packages/dmpersonal/src/PersonalPage.tsx:80,88,104,148` — `resolveAndPaint` only ever sets `workspaceReady` to `true` (success) or `false` (error); it never resets it to `false` at the *start* of a re-resolve. After the first successful load `workspaceReady` stays `true`, so during the async window opened by a Personal re-activation (`resolveRef.current(true)` → `listWorkspaces()` in flight) the sidebar tab buttons remain enabled. A fast tab click in that sub-second window runs `openTab` off the *previous* resolve's `machineModeRef` / `selectedWsRef`, mounting `RuntimePage`, whose mount effect issues one `listRuntimes()` against the stale context (e.g. after a 0→1 workspace transition it can momentarily hit `/machine-runtimes` instead of the new workspace, or vice-versa).

Why this is non-blocking:
- It self-heals: the seq-guarded resolve lands moments later and remounts `RuntimePage` with the correct context, re-fetching.
- No tenancy/data-leak path: a stale-slug `/runtimes` request is bounded by the backend workspace-member middleware (returns 403 for a workspace the user does not belong to); machine mode omits `x-workspace-slug` entirely.
- Reachability is a contrived fast double-action immediately after an out-of-band workspace create/delete.

Optional hardening: reset `setWorkspaceReady(false)` at the top of `resolveAndPaint` when `showLoading` is true (mirroring the mount-time gate), so `openTab` is blocked until the fresh resolve completes. Fine as a follow-up.

### nit — `getRuntime()` still hardcodes the workspace-scoped path

`packages/dmloop/src/api/runtimeApi.ts:15-17` — `getRuntime(id)` calls `/runtimes/${id}` unconditionally, unlike `listRuntimes()` which was correctly migrated to `runtimeListPath(...)`. It has no callers today (verified by grep), so it is dormant, not a live defect — but the next caller added in machine mode will 403. Same latent shape in `agentApi.ts` and `skillApi.ts`, but those live only in Agent/Skill pages that are not mounted in Personal machine mode, so they are out of this diff's blast radius. Worth a one-line follow-up to keep the "runtime reads choose their endpoint by scope" invariant intact.

## 3. Security note (this PR was flagged security-sensitive)

Frontend-only change. Machine-mode isolation depends on the backend: `/machine-runtimes` must be strictly owner-scoped (auth-only, returning only the caller's own machine runtimes) and `/runtimes` must reject requests with no workspace slug. The frontend side is correct: `http.ts:32` only sets `x-workspace-slug` when the slug is non-empty, machine mode clears the context via `setWorkspaceContext("", "")` before `RuntimePage` mounts, and the runtime list routes to `/machine-runtimes` when the slug is empty. The headless PAT / add-computer flow (`/cli-token/headless`) is user-scoped and unaffected by machine mode. The owner-scoping of `GET /api/machine-runtimes` is out of this diff's reach (ships in the backend repo) and should be confirmed server-side by a human before relying on this page for a 0-workspace user.

## 4. Suggested changes

1. (Optional, P2) Reset `workspaceReady=false` at the top of `resolveAndPaint(showLoading=true)` to close the transient stale-context window on `openTab`.
2. (Optional, nit) Route `getRuntime()` through the same scope-aware path selection as `listRuntimes()` to keep the invariant.

## 5. Additional observations

- `resolveWorkspaceSelection` / `runtimeListPath` are cleanly extracted pure functions with good unit coverage.
- Skill→runtime fallback in machine mode is safe: `resolveAndPaint` forces the tab back to `runtime` and `openTab` has a ref-based skill guard that holds even if the `machineMode` state lags the ref, so the disabled button cannot be bypassed to reach a workspace-scoped `SkillPage` in machine mode.
- Clearing the shared global to `("", "")` does not harmfully bleed into Loop — `LoopPage` re-establishes its own context on its activation.

## Overall verdict

**APPROVED** — Spec ✅ and Quality Approved. All four prior blocking/advisory items are fixed and re-verified at this head. The remaining items are one non-blocking P2 (transient, self-healing, no data-leak path) and one dormant nit.

Review provenance: one independent reviewer pass plus an internal adversarial verification pass (which surfaced and I down-graded the P2 above after tracing its blast radius). External advisory legs were unavailable this run and are recorded as **absent**, not as agreement.

PR #720, opened 2026-07-14
feat(docs): expand font color palette and add custom color picker (#719)

ploy-elison:
## What & why

Implements the font-color enhancement for the docs editor (octo-web #719, plan A). Two changes, front-end only, no schema/backend touch (`SCHEMA_VERSION` stays 15; color remains a `textStyle` mark attr):

1. **Palette expansion** — `TEXT_COLORS` in `packages/docs/src/editor/Toolbar.tsx` grows from 5 to the 10 common presets: `#1f2329` (near-black default) `#8a919e` (secondary grey) `#e03131` `#f08c00` `#f2b705` `#2f9e44` `#0ca678` `#1971c2` `#3370ff` (theme blue) `#9c36b5`. The `✕` clear / restore-default (`unsetColor`) is preserved. `HIGHLIGHT_COLORS` is intentionally left untouched — this scope is font color only.
2. **Native custom picker** — a rainbow swatch at the bottom of the text-color popover wraps a native `<input type="color">` (zero new dependencies). Its `onInput` calls `editor.chain().focus().setColor(value).run()`. The native control emits standard `#rrggbb`, so color stays lossless through Yjs sync and the Markdown / DOCX exporters (`normalizeDocxColor` strips `#` and lowercases — verified).

New i18n key `toolbar.customColor` added to `en-US.json` and `zh-CN.json`.

## Tests

Component tests added to `Toolbar.test.tsx` (asserted red on HEAD first, then green): palette contents & order, preset apply to selection, custom-input apply of an arbitrary hex, `✕` clear (`unsetColor`) still works, and highlight palette unchanged (still 5).

Regression baseline (before → after), `packages/docs`:
- `Toolbar.test.tsx`: 17 → 22 pass (all green).
- Color-relevant files (Toolbar, schema, markdown export, docx marks, styles) green before and after.
- Full docs suite: 963 → **968** pass, `+5` new tests. The only failing file, `EditorShell.test.tsx` (4 failures, an emoji-picker `doScroll` rAF issue), is **pre-existing on clean HEAD** and unrelated to this change (verified by running it on the unmodified tree). Typecheck: the two `tsc` errors present are in `dmworkbase/format.ts` and `members/sort.ts`, also pre-existing and outside this change.

## Non-regression checklist (plan §4)

[...]

Issue #719, opened 2026-07-14
octo-docs: add common font colors and allow selecting custom/uncommon colors


Jerry-Xin:
## Summary
Enhance the font color feature in octo-docs so users can quickly pick from a set of common colors and also choose arbitrary custom (uncommon) colors.

## Motivation
The current font color options are limited. Users need both a quick palette of frequently used colors and the flexibility to select any color for finer control over document styling.

## Proposed Changes
- Provide a preset palette of common font colors for one-click selection.
- Add a custom color picker so users can select any uncommon/arbitrary color (e.g. via a color wheel or hex input).
- Persist the selected color in the document content model so it renders consistently.

## Acceptance Criteria
- [ ] A palette of common font colors is available in the font color menu.
- [ ] Users can open a custom color picker and select any color.
- [ ] The chosen color is applied to the selected text and saved correctly.
- [ ] Rendering is consistent across editing and read-only views.

Jerry-Xin:
Resolved. Closing.

PR #732, opened 2026-07-14
feat(docs): expand highlight palette and add custom highlight picker

ploy-elison:
## What & why

Follow-up to the font-colour enhancement (#719 / merged #720): apply the
same treatment to the **highlight** (text-background) control in the docs
editor. Front-end only, no schema/backend touch — highlight colour stays a
`Highlight` mark attribute (`multicolor: true`), so no `SCHEMA_VERSION` bump.

1. **Palette expansion** — `HIGHLIGHT_COLORS` in
   `packages/docs/src/editor/Toolbar.tsx` grows from 5 to 10 light / pastel
   presets, spread warm→cool: `#fff3a3` `#ffe0a3` `#ffd6cc` `#ffd6e7`
   `#e7d6ff` `#d6ddff` `#cfe2ff` `#c9f0ef` `#cdeccd` `#e6e9ed`. Every swatch
   is a light tint so dark foreground text stays readable on the highlight.
   The original five are preserved. The `✕` clear (`unsetHighlight`) is
   untouched. Values are standard `#rrggbb`, kept lossless by the DOCX /
   Markdown exporters.
2. **Native custom picker** — a rainbow swatch at the bottom of the
   highlight popover wraps a native `<input type="color">` (zero new
   dependencies), reusing the shared `.octo-color-custom` styling introduced
   by #720. It calls `setHighlight({ color })` and closes the popover, and
   reuses the existing `toolbar.customColor` i18n key (no new keys).

## RC notes (same three lessons as #719/#720)

- **RC1 — single transaction**: the picker commits exactly once, on the
  native `change` event (bound via a ref), never on the `input` stream. One
  pick = one ProseMirror transaction = one undo step = one Yjs update. The
  OS colour dialog previews the hue live in its own UI while dragging, so the
  popover just stays open during the drag and collapses on commit, matching a
  preset-swatch click.
- **RC2 — stylelint**: no new CSS. The rainbow swatch and its scoped
  `stylelint-disable color-no-hex` exception already exist from #720 and are
  reused as-is; the new highlight presets are applied as inline
  `style={{ backgroundColor }}`, so nothing new reaches the CSS linter.
- **RC3 — keyboard focus**: the visible `:focus-within` accent-token outline
  on `.octo-color-custom` (from #720) applies to the highlight custom picker
  too, since it reuses the same class.

## Tests

Added a `Toolbar — highlight palette + custom picker` describe block to
`Toolbar.test.tsx` (mirrors the #719 text-colour suite): palette contents &
order (10), preset apply to selection, custom-input apply of an arbitrary
hex, popover stays open on `input` / commits + closes on `change`, a
single-undo assertion proving one pick is one transaction, `✕` clear
(`unsetHighlight`) preserved, and text-colour palette left unchanged (still
10). The obsolete "highlight still 5" assertion inside the #719 suite is
updated to the new reality.

`packages/docs`:
- `Toolbar.test.tsx`: 24 → **31** pass (all green), `+7` tests.
- Full docs suite baseline (before → after): `1043 → 1050` pass, `+7` new.
  The one failing file (`DocsHome.test.tsx`, 4 failures + 5 undici WebSocket
  errors) is **pre-existing on clean HEAD** — verified by stashing this change
  and re-running (identical `4 failed | 1043 passed`, and the file passes in
  isolation). Unrelated to this change, which touches only `Toolbar.tsx`.
- Typecheck: the pre-existing `tsc` errors in `dmworkbase` / `react-markdown`
  are unchanged; none in the touched files.

## Non-regression

Preserved: existing preset highlights (`toggleHighlight`), `✕` clear
(`unsetHighlight`), font colour (#719), other formatting, Yjs collab
round-trip, Markdown / DOCX highlight export, light/dark theme, popover
active-only-while-open logic.

Draft — carries review only; not for merge on this branch.

### DIFF
--- a/packages/docs/src/editor/Toolbar.test.tsx
+++ b/packages/docs/src/editor/Toolbar.test.tsx
@@ -548,6 +548,26 @@ describe('Toolbar — highlight palette + custom picker', () => {
     const clear = within(popover).getByText('✕')
     fireEvent.click(clear)
     expect(editor!.getAttributes('highlight').color).toBeUndefined()
+    // The <mark> must actually leave the document, not just the caret attributes.
+    expect(editor!.getHTML()).not.toContain('<mark')
+  })
+
+  // Regression (XIN-1022): clicking into highlighted text leaves a COLLAPSED caret inside the
+  // <mark>. unsetHighlight() on its own clears only stored marks in that case, so the <mark>
+  // stayed in the document and the ✕ button looked dead. The ✕ now extends the selection over the
+  // whole highlight first (extendMarkRange), so the mark is truly removed from a collapsed caret.
+  it('clears the highlight from a collapsed caret inside the mark (✕ removes the <mark>)', () => {
+    // Highlight "hello", then drop a collapsed caret in the middle of it — no range selected.
+    editor!.chain().focus().selectAll().setHighlight({ color: '#fff3a3' }).run()
+    editor!.chain().focus().setTextSelection(3).run()
+    expect(editor!.state.selection.empty).toBe(true)
+    expect(editor!.getHTML()).toContain('<mark')
+
+    const popover = openHighlightPopover()
+    fireEvent.click(within(popover).getByText('✕'))
+
+    expect(editor!.getHTML()).not.toContain('<mark')
+    expect(editor!.getAttributes('highlight').color).toBeUndefined()
   })
 
   it('leaves the text-colour palette untouched (still 10 swatches, this scope is highlight only)', () => {
--- a/packages/docs/src/editor/Toolbar.test.tsx
+++ b/packages/docs/src/editor/Toolbar.test.tsx
@@ -419,11 +419,142 @@ describe('Toolbar — text colour palette + custom picker (#719)', () => {
     expect(editor!.getAttributes('textStyle').color).toBeUndefined()
   })
 
-  it('leaves the highlight palette untouched (still 5 swatches, this scope is font-colour only)', () => {
+  it('leaves the highlight palette untouched by the text-colour scope (highlight has its own presets)', () => {
+    // #719 (font colour) did not alter the highlight palette; the highlight expansion is a
+    // separate follow-up and is covered by its own describe block below. Here we only assert the
+    // text-colour popover does not leak swatches into the highlight control.
     render(<Toolbar editor={editor!} />)
     fireEvent.click(titleBtn('docs.toolbar.highlight'))
     const highlightSwatches = document.querySelectorAll('button[title^="Highlight #"]')
-    expect(highlightSwatches).toHaveLength(5)
+    expect(highlightSwatches).toHaveLength(10)
+  })
+})
+
+// octo-web highlight follow-up (same plan as #719): expanded highlight (text-background) palette
+// + native custom highlight picker. The highlight popover now offers ~10 light presets and a
+// native <input type="color"> entry for arbitrary hex, while clear (unsetHighlight) and the
+// popover-open active logic stay untouched, mirroring the text-colour control.
+describe('Toolbar — highlight palette + custom picker', () => {
+  function openHighlightPopover(): HTMLElement {
+    render(<Toolbar editor={editor!} />)
+    fireEvent.click(titleBtn('docs.toolbar.highlight'))
+    const popover = document.querySelector('.octo-highlight-color-popover') as HTMLElement
+    if (!popover) throw new Error('highlight popover did not open')
+    return popover
+  }
+
+  it('offers the ~10 common light preset swatches, in order', () => {
+    const popover = openHighlightPopover()
+    const swatches = within(popover).getAllByTitle(/^Highlight #/)
+    expect(swatches).toHaveLength(10)
+    const colours = swatches.map((s) => (s.getAttribute('title') || '').replace('Highlight ', ''))
+    expect(colours).toEqual([
+      '#fff3a3',
+      '#ffe0a3',
+      '#ffd6cc',
+      '#ffd6e7',
+      '#e7d6ff',
+      '#d6ddff',
+      '#cfe2ff',
+      '#c9f0ef',
+      '#cdeccd',
+      '#e6e9ed',
+    ])
+  })
+
+  it('applies a preset swatch highlight to the selection', () => {
+    editor!.chain().focus().selectAll().run()
+    const popover = openHighlightPopover()
+    const swatch = within(popover).getByTitle('Highlight #cfe2ff')
+    fireEvent.click(swatch)
+    expect(editor!.getAttributes('highlight').color).toBe('#cfe2ff')
+  })
+
+  it('exposes a native colour input that commits an arbitrary hex highlight on change', () => {
+    editor!.chain().focus().selectAll().run()
+    const popover = openHighlightPopover()
+    const input = popover.querySelector('input[type="color"]') as HTMLInputElement
+    expect(input).toBeTruthy()
+    // Commit happens on `change` (picker closed / value settled), not on the raw `input`
+    // stream that fires continuously while the OS hue wheel is dragged.
+    fireEvent.change(input, { target: { value: '#123456' } })
+    expect(editor!.getAttributes('highlight').color).toBe('#123456')
+  })
+
+  it('leaves the popover open during a drag (input) and commits + closes on change', () => {
+    editor!.chain().focus().selectAll().run()
+    const popover = openHighlightPopover()
+    const input = popover.querySelector('input[type="color"]') as HTMLInputElement
+
+    // Dragging fires `input` repeatedly. RC1: we intentionally do NOT commit per tick (that
+    // flooded undo/Yjs); the popover simply stays open so the user can keep nudging the hue.
+    fireEvent.input(input, { target: { value: '#112233' } })
+    expect(editor!.getAttributes('highlight').color).toBeUndefined()
+    expect(document.querySelector('.octo-highlight-color-popover')).toBeTruthy()
+
+    // Committing the pick fires `change`: the final colour is applied and the popover collapses,
+    // matching a preset-swatch click.
+    fireEvent.change(input, { target: { value: '#abcdef' } })
+    expect(editor!.getAttributes('highlight').color).toBe('#abcdef')
+    expect(document.querySelector('.octo-highlight-color-popover')).toBeNull()
+  })
+
+  // RC1: dragging the native hue wheel fires `input` continuously. Committing on every `input`
+  // pushed one ProseMirror transaction per event — tens of undo records and a Yjs update flood
+  // per single pick. The picker now previews via the OS dialog and commits exactly once on
+  // `change`, so one pick == one undo step == one collaboration update.
+  it('does not commit while dragging (raw input events) — a pick is a single undo step', () => {
+    // A history-enabled editor (StarterKit default) so we can assert the undo depth of one pick.
+    const e = new Editor({
+      extensions: [StarterKit, TaskList, TaskItem, Highlight.configure({ multicolor: true }), TextStyle, Color, Link, FindReplace],
+      content: '<p>hello</p>',
+    })
+    e.chain().focus().selectAll().run()
+    render(<Toolbar editor={e} />)
+    fireEvent.click(titleBtn('docs.toolbar.highlight'))
+    const input = document.querySelector('.octo-highlight-color-popover input[type="color"]') as HTMLInputElement
+    expect(input).toBeTruthy()
+
+    let docChanges = 0
+    e.on('transaction', ({ transaction }) => {
+      if (transaction.docChanged) docChanges++
+    })
+
+    // Simulate a drag across the hue wheel: a stream of intermediate `input` events.
+    fireEvent.input(input, { target: { value: '#111111' } })
+    fireEvent.input(input, { target: { value: '#222222' } })
+    fireEvent.input(input, { target: { value: '#333333' } })
+    // Nothing is committed to the document (and undo history is untouched) during the drag.
+    expect(e.getAttributes('highlight').color).toBeUndefined()
+    expect(docChanges).toBe(0)
+
+    // Releasing the picker fires `change` once → exactly one document-changing transaction.
+    fireEvent.change(input, { target: { value: '#345678' } })
+    expect(e.getAttributes('highlight').color).toBe('#345678')
+    expect(docChanges).toBe(1)
+
+    // And a single undo fully reverts the pick — proof it is one undo record, not many.
+    expect(e.can().undo()).toBe(true)
+    e.chain().undo().run()
+    expect(e.getAttributes('highlight').color).toBeUndefined()
+
+    e.destroy()
+  })
+
+  it('still clears the highlight via the ✕ button (unsetHighlight preserved)', () => {
+    editor!.chain().focus().selectAll().setHighlight({ color: '#fff3a3' }).run()
+    expect(editor!.getAttributes('highlight').color).toBe('#fff3a3')
+    const popover = openHighlightPopover()
+    const clear = within(popover).getByText('✕')
+    fireEvent.click(clear)
+    expect(editor!.getAttributes('highlight').color).toBeUndefined()
+  })
+
+  it('leaves the text-colour palette untouched (still 10 swatches, this scope is highlight only)', () => {
+    render(<Toolbar editor={editor!} />)
+    fireEvent.click(titleBtn('docs.toolbar.textColor'))
+    const textSwatches = document.querySelectorAll('button[title^="Text #"]')
+    expect(textSwatches).toHaveLength(10)
   })
 })
 
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -275,7 +275,23 @@ export function shouldShowFloatingMenu(args: {
   return isRootDepth && isEmptyTextBlock
 }
 
-const HIGHLIGHT_COLORS = ['#fff3a3', '#ffd6cc', '#cdeccd', '#cfe2ff', '#e7d6ff'] as const
+// Common highlight (text-background) colours (octo-web follow-up to #719, same plan): a set of
+// light / pastel tints spread warm→cool so dark foreground text stays readable on every swatch.
+// The original five (yellow, peach, green, blue, purple) are preserved and kept in place; the new
+// entries fill out the spectrum (amber, pink, indigo, cyan, neutral grey). Values are standard
+// #rrggbb hex so DOCX/Markdown export keeps them lossless, mirroring TEXT_COLORS below.
+const HIGHLIGHT_COLORS = [
+  '#fff3a3',
+  '#ffe0a3',
+  '#ffd6cc',
+  '#ffd6e7',
+  '#e7d6ff',
+  '#d6ddff',
+  '#cfe2ff',
+  '#c9f0ef',
+  '#cdeccd',
+  '#e6e9ed',
+] as const
 // Common font colours (octo-web #719, plan A): near-black default, secondary grey, then the
 // warm→cool spread. Values are standard #rrggbb hex so DOCX export (normalizeDocxColor) keeps
 // them lossless. This scope covers font colour only — HIGHLIGHT_COLORS above is intentionally
@@ -296,6 +312,27 @@ const TEXT_COLORS = [
 /** Text-highlight control (SCHEMA-SPEC §3): palette of background colours + clear. */
 function HighlightControl({ editor }: { editor: Editor }) {
   const [open, setOpen] = useState(false)
+  // Native <input type="color"> distinguishes drag from commit only at the DOM level: `input`
+  // streams while the hue wheel moves, `change` fires once the pick is committed. React folds
+  // both onto its synthetic onChange (native `input`), so we bind the raw `change` event via a ref.
+  // RC1: commit the highlight once, on `change` only — never on the `input` stream. Applying per
+  // `input` tick ran one ProseMirror transaction each, so a single pick piled up dozens of undo
+  // records and flooded collaborators with a Yjs update per intermediate hue. The OS colour dialog
+  // previews the hue live in its own UI while dragging, so committing on `change` keeps one pick =
+  // one undo record + one Yjs update, and the popover collapses on commit like a preset swatch.
+  const customRef = useRef<HTMLInputElement>(null)
+  useEffect(() => {
+    const input = customRef.current
+    if (!input) return
+    const onCommit = () => {
+      editor.chain().focus().setHighlight({ color: input.value }).run()
+      setOpen(false)
+    }
+    input.addEventListener('change', onCommit)
+    return () => {
+      input.removeEventListener('change', onCommit)
+    }
+  }, [editor, open])
   return (
     <span className="octo-color-control">
       <Btn
@@ -305,7 +342,7 @@ function HighlightControl({ editor }: { editor: Editor }) {
         onClick={() => setOpen((v) => !v)}
       />
       {open && (
-        <span className="octo-color-popover">
+        <span className="octo-color-popover octo-highlight-color-popover">
           {HIGHLIGHT_COLORS.map((c) => (
             <button
               key={c}
@@ -327,6 +364,22 @@ function HighlightControl({ editor }: { editor: Editor }) {
               setOpen(false)
             }}
           />
+          {/* Custom highlight (same approach as the text-colour picker): native picker, zero new
+              deps. It emits standard #rrggbb, so setHighlight stays lossless through Yjs and the
+              DOCX/Markdown exporters. The picker stays open while dragging the hue wheel and
+              commits once on `change`, collapsing the popover — see the ref-bound listener above. */}
+          <label
+            className="octo-swatch octo-color-custom"
+            title={t('docs.toolbar.customColor')}
+            onMouseDown={(e) => e.preventDefault()}
+          >
+            <input
+              ref={customRef}
+              type="color"
+              className="octo-color-custom-input"
+              aria-label={t('docs.toolbar.customColor')}
+            />
+          </label>
         </span>
       )}
     </span>
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -360,7 +360,14 @@ function HighlightControl({ editor }: { editor: Editor }) {
           <Btn
             label="✕"
             onClick={() => {
-              editor.chain().focus().unsetHighlight().run()
+              // Clear the highlight the caret sits in. unsetHighlight() alone only clears a
+              // non-empty selection range — with a collapsed caret inside a highlight (the common
+              // "click into highlighted text, then hit ✕" flow) it clears stored marks only and
+              // leaves the surrounding <mark> in the document. extendMarkRange('highlight') first
+              // grows the selection to span the whole highlight under the caret (a no-op when a
+              // range is already selected), so unsetHighlight() reliably removes the mark. Same
+              // idiom TipTap uses for link clearing (extendMarkRange('link').unsetLink()).
+              editor.chain().focus().extendMarkRange('highlight').unsetHighlight().run()
               setOpen(false)
             }}
           />


OctoBoooot (dismissed):
## Review: #732 — expand highlight palette + custom highlight picker (follow-up to #720)

**Verdict:** Approve. The highlight counterpart of #720 (font color), and it correctly reuses all of #720's hard-won fixes — no re-litigating the rounds that PR needed. Risk tier: normal (highlight is a `Highlight` mark attr, `multicolor: true`; no schema/backend). CI green; `code-review` pending. PR is draft.

### Verified — landed #720's lessons pre-emptively
- **No injection surface**: native `<input type="color">` (not a text input); `setHighlight({ color: input.value })` reads the browser-constrained `#rrggbb`. Presets are hardcoded hex.
- **Commit-on-change, not per-keystroke**: the ref-bound listener commits **only on `change`** (`onCommit`), never on the `input` stream — the one-transaction-per-pick fix from #720. No undo/Yjs spam.
- **a11y focus present (inherited)**: no CSS in this PR because it reuses `.octo-color-custom` — and `.octo-color-custom:focus-within { outline }` already exists in `styles.css` (from #720), so the transparent input keeps a visible keyboard-focus ring. The #720 a11y 🟡 doesn't recur.
- **`useEffect([editor, open])`** lifecycle correct (input mounts on open → listener attaches; cleanup on close).
- **i18n**: no new string — reuses the existing bilingual `docs.toolbar.customColor` key.
- Palette 5→10 with the original five preserved in place; standard `#rrggbb` (lossless DOCX/Markdown export).

### Tests
Cover the 10-swatch palette + exact order, preset apply to selection, the custom picker, and clear (`unsetHighlight`) preserved. Good, mirroring #720's suite.

### Minor (carried from #720, non-blocking)
- The custom picker still opens at default black rather than the current highlight color (no `value`/`defaultValue` binding). Cosmetic — `setHighlight` only fires on `change`, so opening never writes black. Same deferred 🟡 as #720; fine to leave.

### Net
Clean, low-risk feature PR that applies #720's finalized pattern (native picker, commit-on-change, focus-within, key reuse) rather than repeating its mistakes — nothing to hold on. Approving. Still draft, so not mergeable regardless. Credit ploy-elison.

Jerry-Xin (approved):
Reviewed line-by-line as a follow-up to the font-colour work (#719/#720). This faithfully mirrors that control, and the three regression carryovers from #719 are all internalized. Approving.

### RC carryovers — all addressed

**RC1 — transaction convergence (single undo + single Yjs update per pick): PASS.**
The custom picker binds the native `change` event through a ref (`customRef` + `useEffect`) and commits `setHighlight` exactly once there — never on the `input` stream that fires continuously while the OS hue wheel is dragged. This matches the merged text-colour approach. The test suite asserts it directly: during a drag (`input` x3) `docChanges` stays `0` and no highlight is applied; on `change` exactly one doc-changing transaction fires; and a single `undo()` fully reverts the pick. One pick == one transaction == one undo record == one Yjs update.

**RC2 — stylelint / hardcoded hex: PASS.**
No CSS is touched by this PR. The `HIGHLIGHT_COLORS` hex values live in the `.tsx` (JS string constants), which `color-no-hex` does not lint, and the custom swatch reuses the existing `.octo-color-custom` rule whose rainbow `conic-gradient` is already wrapped in a scoped `stylelint-disable color-no-hex` / `stylelint-enable` pair (added in #720) with an explanatory comment. No new bare CSS hex is introduced, and the scoped-disable is not a global one.

**RC3 — keyboard focus visibility: PASS.**
The custom entry reuses `.octo-color-custom:focus-within { outline: 2px solid var(--octo-accent); outline-offset: 1px }` from #720. Since the native `<input type="color">` is `opacity: 0`, focus is surfaced on the wrapping label via `:focus-within`, using the accent token so it follows light/dark theme and matches the hover affordance.

### No regression

- Highlight stays a Highlight mark attribute — no schema/backend change.
- Zero new dependencies: native `<input type="color">`, no semi-ui import added.
- Text-colour palette untouched (test asserts it still renders 10 swatches); clear (`unsetHighlight`) preserved and tested.
- i18n reuses the existing `toolbar.customColor` key, present in both `en-US` and `zh-CN`; no missing key.
- Palette values are standard `#rrggbb`, so DOCX/Markdown export stays lossless. Original five presets preserved and kept in place.
- Full `Toolbar.test.tsx` suite passes locally (31/31).

### Minor (non-blocking)

Preset swatches use `toggleHighlight` while the custom picker uses `setHighlight`. That is consistent with the pre-existing preset behaviour and mirrors the text-colour control's `setColor`; for an explicit colour pick, always-set is the sensible semantics. No change required.

LGTM.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #732 (octo-web)

## Summary

This PR expands the highlight (text-background) colour palette from 5 to 10 swatches and adds a native `<input type="color">` custom highlight picker to the docs editor toolbar. The implementation mirrors the text-colour control introduced in #719 — same ref-bound native `change` listener, same popover-collapse behaviour, same single-undo-step commit semantics. Test coverage is thorough, covering preset ordering, custom commit, drag-vs-commit distinction, undo depth, clear button, and cross-control isolation.

## Verification

- ✅ **Diff scope** — two files changed: `packages/docs/src/editor/Toolbar.tsx` (+55/-2) and `packages/docs/src/editor/Toolbar.test.tsx` (+133/-2). No production wiring, routing, config, or dependency changes.
- ✅ **Data-flow trace** — `input.value` from the native `<input type="color">` flows through the ref-bound `change` listener → `editor.chain().focus().setHighlight({color: input.value}).run()` → TipTap highlight mark. The browser guarantees `input.value` is a valid `#rrggbb` hex; no parsing or sanitization needed.
- ✅ **useEffect lifecycle** — `[editor, open]` dependency array correctly re-binds the listener when the popover mounts and cleans it up on unmount or close. No stale-closure risk: `editor` is a stable prop, `input.value` is read at callback time.
- ✅ **C1 — Double-path parity** — HighlightControl mirrors TextColorControl structurally (custom ref, native change, popover collapse, clear button). Preset swatches use `toggleHighlight` (pre-existing) vs TextColorControl's `setColor`; the new custom picker uses `setHighlight`, consistent with TextColorControl's `setColor`. See Nit below.
- ✅ **C2 — Control-flow ordering** — No nested reuse or double-invocation risk. The `change` listener is bound once per popover open, cleaned up on close.
- ✅ **C3–C6** — N/A. No auth, permissions, build, deployment, or policy changes. This is a pure UI addition within an existing editor control.
- ✅ **RC1 (undo/collaboration safety)** — Tests prove the `input` event stream does not commit transactions; only `change` produces exactly one doc-changing transaction and one undo record. The history-enabled editor test at `Toolbar.test.tsx:500-535` validates this end-to-end.

Static analysis only at head `a77d41ad5f5328d193282a735e333952a14349c1`; build and tests not executed in this environment.

## Findings

### Nit — `toggleHighlight` vs `setHighlight` behavioural asymmetry (`packages/docs/src/editor/Toolbar.tsx:355`)

Preset swatches call `toggleHighlight({color: c})` — clicking an already-active colour removes the highlight. The new custom picker calls `setHighlight({color})` — picking the same colour via the native input always (re-)applies it. The preset toggle is pre-existing (not introduced by this PR) and the custom path mirrors `TextColorControl.setColor`, so the inconsistency lives within the highlight popover only. If you want full parity, switch presets to `setHighlight` (matching `setColor`), or accept the toggle-off as intentional preset behaviour.

- Palette expansion: 10 `#rrggbb` hex values, all valid, all in the `as const` tuple. Test asserts exact count and ordering.
- `octo-highlight-color-popover` class scoping: correctly distinguishes the highlight popover from the text-colour popover for test queries and potential future CSS targeting.
- `onMouseDown` preventDefault on the custom picker label: prevents focus steal and selection loss, same pattern as preset swatches and text-colour picker.
- Clear button (✕) and `unsetHighlight` path: preserved and tested.
- Cross-control isolation: highlight changes don't affect text-colour swatch count (tested), and vice versa (tested).
- No dependency additions — native `<input type="color">` only.

## Verdict: APPROVED

Clean, well-scoped feature addition that faithfully mirrors the existing text-colour picker pattern. Test coverage is thorough and covers the critical drag-vs-commit undo/collaboration regression. The single nit is a pre-existing behavioural asymmetry, not introduced by this PR.

[Octo-Q] verdict: APPROVE — no P0/P1; one Nit (pre-existing toggleHighlight vs setHighlight asymmetry in preset vs custom paths). Safe to land.

yujiawei (dismissed):
# Code Review — PR #732 (octo-web)

Reviewed at head SHA `a77d41ad5f5328d193282a735e333952a14349c1`. Scope: 2 files, +188/-4 (`packages/docs/src/editor/Toolbar.tsx`, `packages/docs/src/editor/Toolbar.test.tsx`). This PR expands the highlight palette (5 → 10 presets) and adds a native `<input type="color">` custom highlight picker in `HighlightControl`, mirroring the already-merged font-colour control (`TextColorControl`).

Verification performed locally: full `Toolbar.test.tsx` suite passes (31/31); typecheck is clean for the touched files (the two `tsc` errors surfaced are pre-existing in unrelated files — `dmworkbase/i18n/format.ts`, `members/sort.ts` — and are not touched by this PR).

## Spec compliance

**Spec: ✅**

- Missing: none — palette expansion + custom picker + preserved clear (`unsetHighlight`) are all present and tested.
- Extra/out-of-scope: none — the change is confined to `HighlightControl` and its tests; no new deps, no unrelated files.
- Divergence: none — it faithfully follows the sibling font-colour pattern (native picker, zero new deps, commit-once-on-`change`).

The commit-on-`change`-not-`input` design is correct and well-tested: dragging the OS hue wheel fires `input` repeatedly but the highlight is committed exactly once on `change`, so a single pick is one ProseMirror transaction / one undo step / one Yjs update. `Toolbar.test.tsx` asserts this explicitly (`docChanges === 1`, single-undo reverts the pick). The `useEffect([editor, open])` + ref-bound listener lifecycle is sound: the input only exists in the DOM while `open`, the effect adds the listener on open and removes it on close/unmount, and `customRef.current` is null-guarded — no leak, no double-bind.

## Code quality

**Quality: Approved** — findings below are all P2 (non-blocking); no P0/P1.

**[P2] `setHighlight` (custom picker) vs `toggleHighlight` (preset swatches) — divergent semantics.**
`Toolbar.tsx:355` (preset swatch) calls `toggleHighlight({ color: c })`; `Toolbar.tsx:328` (custom-picker commit) calls `setHighlight({ color: input.value })`. With `Highlight.configure({ multicolor: true })`, `toggleHighlight` unsets the mark when the selection already carries that exact colour (toggle-off), whereas `setHighlight` always applies. So re-selecting the current colour via a preset swatch *clears* the highlight, but choosing the same colour via the custom picker *keeps* it — opposite behaviour for what reads as the same "apply this highlight" action. The sibling `TextColorControl` is internally consistent (`setColor` for both swatches and picker). This is a UX-consistency wrinkle, not a functional break (toggle-off is tiptap's native multicolor default), and the toggle-off path is not covered by the new tests. Consider `setHighlight` for the swatches too, or a comment documenting the intentional toggle-off UX.

**[P2] Inaccurate "lossless through DOCX export" comments.**
`Toolbar.tsx:282` ("…so DOCX/Markdown export keeps them lossless…") and `Toolbar.tsx:368` ("…setHighlight stays lossless through Yjs and the DOCX/Markdown exporters") are misleading for DOCX: `export/docx/marks.ts:141-143` unconditionally maps every highlight to `opts.highlight = 'yellow'` (docx `w:highlight` supports named colours only), so all 10 presets and any custom hex flatten to yellow in `.docx`. The sentence was true when copied from the `TextStyle` path (which *does* map hex via `normalizeDocxColor`, `marks.ts:149`) but is false for highlight. Markdown export *is* lossless (`export/markdown.ts:471-481` emits an escaped inline `background-color` style). Recommend fixing the comment (e.g. "lossless through Markdown; DOCX flattens all highlights to yellow due to the named-colour limitation") so a future maintainer doesn't rely on a false invariant.

**[P2] ~60-line copy-paste between `HighlightControl` and `TextColorControl`.**
The two controls (palette map + clear button + ref-bound native picker + identical `useEffect` commit listener) are now near-verbatim duplicates differing only in `setHighlight`/`toggleHighlight`/`unsetHighlight` vs `setColor`/`unsetColor` and the swatch `title` prefix. Extracting a shared `ColorControl({ commit, clear, colors, titlePrefix })` would remove the duplication and prevent the two from drifting (as the toggle-vs-set divergence above already shows they can). Optional refactor.

## Security (PR classified security-sensitive)

No injection vector introduced. The native `<input type="color">` can only emit a normalized `#rrggbb` value, so the custom picker cannot inject arbitrary strings. The one place a highlight colour reaches serialized output — Markdown export — already escapes it: `export/markdown.ts:479` emits `<mark style="background-color:${escapeHtmlAttr(hl)}">`, and `escapeHtmlAttr` (`markdown.ts:73-79`) escapes `& < > "`, so even a hostile colour attr arriving via a *paste/import* path (not touched by this PR) is neutralized against attribute breakout. DOCX drops the colour entirely (yellow), which is safe. Nothing here requires manual human security verification beyond the palette values themselves.

## Overall verdict

**APPROVED** — spec ✅ and no P0/P1 quality findings. The three P2 items (toggle/set consistency, the inaccurate DOCX comment, and the duplication) are polish/maintainability suggestions the author may address in this PR or a follow-up; none blocks merge.

Reviewer note on coverage: this review combined a multi-lens adversarial pass (correctness / security / maintainability, each finding independently re-verified against the source and installed `@tiptap` dependency) with a manual diff read. Blind spots I could not fully exercise from the diff: real-browser drag behaviour of the OS colour dialog across platforms, and visual contrast of the 10 pastel presets against dark-mode foreground text — worth a quick manual UI check but not a code defect.

Jerry-Xin (dismissed):
Reviewed at head a77d41ad. Clean, well-tested docs-editor feature that mirrors the already-shipped text-colour picker (#719). No blocking issues.

### Security review — custom highlight colour (the key dimension)
I byte-traced the custom colour from input to the rendered DOM and there is **no CSS/HTML injection surface** here:

- **Input is constrained by construction.** The custom entry is a native `<input type="color">` (`packages/docs/src/editor/Toolbar.tsx`, `HighlightControl`). A native colour input only ever exposes a normalised 7-char hex value; it is not a free-text field and offers no paste-into-value path, so an attacker cannot supply an arbitrary string here.
- **Render path.** The committed value flows `input.value → setHighlight({ color }) → @tiptap/extension-highlight (multicolor)`, which serialises the colour onto the mark's inline style **attribute** via ProseMirror's attribute serializer (not a raw HTML sink). Because the value is always a well-formed hex, it cannot terminate the property or append extra declarations. No value reaches an HTML/script/`javascript:`-style sink — the only sink is a CSS property, and it is fed a safe token.
- **Scope.** `multicolor` and the paste/parse behaviour pre-date this PR (the extension config is untouched by the diff); this change only adds UI + presets, introducing no new deserialize sink.

### Palette expansion
The 10 presets are all standard `#rrggbb` values, warm→cool light tints — well-formed, readable with dark text, and lossless for DOCX/Markdown export. No malformed entries.

### Interaction / correctness
Nice detail: the picker commits once on the native `change` event (bound via a ref) rather than on the streaming `input` events, so a single pick is one ProseMirror transaction = one undo step = one collaboration update, instead of flooding history/Yjs during a hue-wheel drag. Clear (`unsetHighlight`) and popover-active logic are preserved.

### Tests
Strong coverage: preset order/count, preset application, custom-input commit-on-change, drag-does-not-commit (asserts undo depth and that the doc is unchanged during the drag), single-undo revert, clear button, and an isolation check that the text-colour palette is unaffected.

### i18n
Reuses the existing `docs.toolbar.customColor` key, which is already present and correctly localised in both zh-CN (自定义颜色) and en-US (Custom color). No new key, no gap.

### Highlights
- Native picker → zero new deps, injection-safe by construction.
- Commit-on-change avoids undo/collab flooding.
- Test suite pins both behaviour and regression boundaries.

LGTM.

OctoBoooot (approved):
## Review: #732 @ `69af08c2` — DELTA re-review; two well-tested highlight bug fixes, re-affirming APPROVE

**Verdict:** Approve. This commit (`fix(docs): clear highlight from a collapsed caret`) lands two real, correctly-fixed bugs on the highlight control with strong regression coverage, and introduces no new injection surface. My prior APPROVE (@`a77d41ad`) was auto-dismissed by this commit; re-affirming at the live head. Clean single-commit delta (`ahead 1 / behind 0`): only `Toolbar.tsx` (+8/-1) + `Toolbar.test.tsx` (+20 net, +141 in the diff). CI: all code checks green (Build SUCCESS, secret-scan, osv-scan, dependency-review, pr-title-lint); `code-review` (human gate) pending.

## Parity checked
Base→head (`a77d41ad`→`69af08c2`) capability diff of the highlight control — every prior capability preserved, two broken flows fixed:
- **preset swatch apply** (10-colour palette): unchanged — same `setHighlight({ color })`, palette values identical.
- **custom hex pick**: preserved, and hardened — now commits once on `change` instead of per `input` tick (see below). Still emits `#rrggbb`, lossless through Yjs / DOCX / Markdown.
- **✕ clear on a selected range**: preserved — `extendMarkRange('highlight')` is a no-op when a range is already selected, so `unsetHighlight()` behaves as before.
- **✕ clear from a collapsed caret**: was broken (mark left in doc); now fixed (regained, not lost).
No capability dropped.

## Fix 1 — ✕ clears highlight from a collapsed caret (XIN-1022)
`unsetHighlight()` alone on a collapsed caret clears only stored marks, leaving the `<mark>` in the document — so the common "click into highlighted text → hit ✕" flow left the highlight visible and the button looked dead. The fix prepends `extendMarkRange('highlight')` to grow the selection over the whole mark under the caret before `unsetHighlight()` — the exact idiom TipTap uses for link clearing (`extendMarkRange('link').unsetLink()`). Correct, and behaviour-preserving for the range-selected case (extendMarkRange is a no-op there). The test drives a genuine collapsed caret (`setTextSelection(3)`, asserts `selection.empty`) and verifies `getHTML()` no longer contains `<mark>`.

## Fix 2 — custom picker commits on `change`, not the `input` stream
Native `<input type="color">` streams `input` per hue-wheel tick and fires `change` once on commit; React folds both onto synthetic `onChange`, so the raw `change` is bound via a ref. Committing per `input` ran one ProseMirror transaction per tick — dozens of undo records and a Yjs update flood per single pick. The tests prove the fix: during a simulated drag (`fireEvent.input` ×3) `docChanges === 0` and the value stays uncommitted; on `change` exactly one doc-changing transaction lands and a single `undo()` fully reverts the pick. Popover stays open during drag, collapses on commit (matches a preset click).

## Security — no new injection surface
The custom highlight value flows `input.value → setHighlight({ color })`; the browser constrains `<input type="color">` to `#rrggbb`, the same hex-constrained path verified safe on the prior head. No arbitrary-string CSS injection.

## Net
Two legitimate bugs fixed (dead ✕ on a collapsed caret; undo/collab flood on custom-pick drag), each with real event-level regression tests, no injection surface, and no regression to the approved palette/picker. Re-affirming APPROVE at `69af08c2`. Gate: normal risk, no `needs-human-review`; not mergeable until the required `code-review` check passes. Credit the author; prior reviewers' approvals were auto-dismissed by this commit and should re-anchor.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #732 (octo-web)

## Summary

This PR expands the highlight (text-background) palette from 5 to 10 pastel presets and adds a native `<input type="color">` custom highlight picker in the docs editor toolbar. It also fixes a collapsed-caret bug in the highlight clear button by chaining `extendMarkRange('highlight')` before `unsetHighlight()`. Front-end only, no schema or backend changes. The implementation closely mirrors the text-colour control introduced in #720 and is well-tested with 7 new test cases covering presets, custom picker commit semantics, undo isolation, and the collapsed-caret regression.

## Verification

Static analysis only at head `69af08c238f0656072b4a895d7028dfe83a138bb`; build and tests not executed in this environment.

- **Data-flow: HIGHLIGHT_COLORS to preset swatches to editor command** — array of 10 `#rrggbb` hex values flows through `.map()` into `style={{ backgroundColor }}` for rendering and `toggleHighlight({ color: c })` for commit. All values are valid 6-digit hex, lossless through DOCX/Markdown exporters.
- **Data-flow: native color input to ref to change listener to editor command** — `customRef.current.value` produces standard `#rrggbb` from the native picker (browser-enforced). The `change` event listener (bound via ref, not React synthetic) commits exactly once per pick. No `input` stream handler exists, so drag events do not trigger transactions.
- **Effect lifecycle** — `useEffect([editor, open])` re-registers the `change` listener when the popover mounts/unmounts. Cleanup removes the listener on unmount. The input element is conditionally rendered inside `{open && ...}`, so it is always present when the effect body runs (early return guards null ref). No stale-listener or double-bind risk.
- **extendMarkRange fix** — `editor.chain().focus().extendMarkRange('highlight').unsetHighlight().run()` correctly expands the selection to cover the full `<mark>` under a collapsed caret before clearing. TipTap standard idiom, matching `extendMarkRange('link').unsetLink()`. Test at `Toolbar.test.tsx:556-568` covers the exact collapsed-caret scenario.
- **Test coverage** — 7 new tests cover: palette count and order (10 presets), preset apply, custom hex commit, popover open/close lifecycle (input vs change), single-undo-step proof, clear with full-selection and collapsed-caret, and text-colour palette non-regression. The single-undo test creates an independent `Editor` with `StarterKit` (history enabled) and asserts `docChanges === 1` after one `change` event.
- **Non-regression** — text-colour palette (10 swatches) unchanged; existing `toggleHighlight` / `unsetHighlight` paths preserved; no new CSS (reuses `.octo-color-custom` from #720); no new i18n keys (reuses `docs.toolbar.customColor`).
- **Popover CSS class** — added `octo-highlight-color-popover` alongside existing `octo-color-popover`, enabling test targeting without breaking shared styles. Matches the `octo-text-color-popover` pattern from #720.

## Findings

No P0/P1 issues. One P2 parity observation below.

### P2 — Text-colour clear lacks the same extendMarkRange fix (`packages/docs/src/editor/Toolbar.tsx:441`)

The highlight clear button gains `extendMarkRange('highlight')` before `unsetHighlight()` in this PR, fixing the collapsed-caret bug where clicking into highlighted text and pressing the clear button left the `<mark>` in the document. The same bug pattern applies to the text-colour clear at `:441` (`editor.chain().focus().unsetColor().run()` — no `extendMarkRange`), which was introduced in #720 and is not touched by this diff. Pre-existing, not introduced or amplified by this PR (R2 diff-scope: base `main:382` has identical behavior). Flagging for a follow-up: `editor.chain().focus().extendMarkRange('textStyle').unsetColor().run()`.

## Human-verify

1. **Browser colour-picker UX** — the native `<input type="color">` renders differently across OS/browser (macOS shows a full HSB wheel, Windows shows a simpler grid, mobile may show a system picker). The PR description states the OS dialog previews the hue live while dragging; worth confirming this holds on the target browser matrix (Chrome/Safari/Firefox on macOS and Windows). This is **not** a merge blocker — flagging for cross-browser QA.

- `HIGHLIGHT_COLORS` values are all light/pastel tints — dark foreground text remains readable on every swatch.
- Preset swatch clicks use `toggleHighlight` (same as base); custom picker uses `setHighlight` (intentional: custom always sets, preset toggles).
- `onMouseDown={(e) => e.preventDefault()}` on the custom `<label>` matches the preset-swatch pattern — prevents focus steal from the editor.
- No new dependencies introduced; `<input type="color">` is zero-dep native.
- The `open` state variable in the `useEffect` dependency array is safe: the input is conditionally rendered, so ref is null when closed and the effect early-returns.

## Verdict: APPROVED

Clean, well-scoped, well-tested PR. Palette expansion, custom picker, and the collapsed-caret fix all work as described. The single P2 is a pre-existing parity gap in the text-colour clear handler, not introduced by this diff, and suitable for a follow-up.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; one pre-existing P2 parity note flagged for follow-up.

OctoBoooot:
## Corroborating mochashanyao's 🟡 — the text-color clear needs the same `extendMarkRange` fix (byte-verified, non-blocking)

Verified mochashanyao's note against `Toolbar.tsx` @ `69af08c2`, and she's right — the collapsed-caret fix this PR added for highlight has a symmetric gap in the text-color control:

- **HighlightControl (`:370`)** — fixed: `editor.chain().focus().extendMarkRange('highlight').unsetHighlight().run()`
- **TextColorControl (`:442`)** — still bare: `editor.chain().focus().unsetColor().run()`

So the exact XIN-1022 bug this PR fixes for highlight still bites text color: with a **collapsed caret inside colored text** (the common "click into the coloured run, then hit clear" flow), `unsetColor()` clears only stored marks and leaves the `textStyle` color on the surrounding span — the clear button looks dead. The symmetric fix mirrors the highlight one:

```
editor.chain().focus().extendMarkRange('textStyle').unsetColor().run()
```
(extend over the `textStyle` mark under the caret first; a no-op when a range is already selected — same idiom.)

**Severity: minor, non-blocking, and pre-existing** — `TextColorControl` is not in this PR's delta; the text-color clear was `unsetColor()`-only before #732, so this PR doesn't introduce or regress it. It doesn't change my APPROVE. But since #732 just established exactly the fix pattern that solves it, folding the one-line symmetric change into this revision (or a fast-follow) would close the whole class. Credit mochashanyao for the symmetry catch.

yujiawei (approved):
# Code Review — PR #732 (octo-web)

**Scope reviewed:** head SHA `69af08c238f0656072b4a895d7028dfe83a138bb` vs `main`. Two files: `packages/docs/src/editor/Toolbar.tsx` (+38 impl) and `packages/docs/src/editor/Toolbar.test.tsx` (+tests). Classified security-sensitive, so the arbitrary-hex → `setHighlight` → export path got extra scrutiny.

## Verdict: Approved ✅

The change does exactly what its title states, cleanly mirrors the already-shipped font-colour picker (#719), and introduces no correctness, data-integrity, or injection regressions. No P0/P1 blockers found.

## What the PR does

1. Expands `HIGHLIGHT_COLORS` from 5 → 10 light pastel swatches (warm→cool), preserving the original five in place.
2. Adds a native `<input type="color">` custom highlight picker inside the highlight popover, committing **once on `change`** (not on the streaming `input` events during a hue-wheel drag).
3. Fixes clearing a highlight from a **collapsed caret** by extending the selection over the mark before `unsetHighlight()`.

## Detailed assessment

### Correctness — commit-on-`change`, not `input` ✅
Binding the raw `change` DOM event via a ref (rather than React's synthetic `onChange`, which fires on native `input`) is the right call. Committing per `input` tick would push one ProseMirror transaction per intermediate hue — flooding undo history and Yjs with a burst of updates per single pick. The tests explicitly assert this: a simulated drag produces `docChanges === 0` and no highlight; the release (`change`) produces exactly one document-changing transaction and one undo record. This is a genuinely well-tested edge.

### `extendMarkRange('highlight').unsetHighlight()` clear logic ✅
Verified against the pinned dependency (`@tiptap/core@3.22.2`, `src/commands/extendMarkRange.ts`): the command **unconditionally `return true`** and only mutates the selection when `getMarkRange` yields a range covering the caret. It does **not** abort the chain when the selection is a non-empty range — in that case it simply leaves the selection as-is and `unsetHighlight()` runs on the existing range. So both flows work:
- collapsed caret inside a `<mark>` → selection grows to the mark, mark removed (the bug this PR fixes),
- non-empty selection → `extendMarkRange` is a no-op, `unsetHighlight()` clears the range as before.

Both are covered by tests, including an assertion that `<mark>` actually leaves `getHTML()`, not just the caret attributes.

### Security / injection — arbitrary hex is safe ✅
A native `<input type="color">` can only ever yield a normalized `#rrggbb` string, so `setHighlight({ color })` never receives hostile input from this control. Following the value downstream:
- **Markdown export** (`export/markdown.ts`, `applyMarks` → `highlight` case): the colour is emitted as `background-color:${escapeHtmlAttr(hl)}` inside a `style="…"`, and `escapeHtmlAttr` escapes `& < > "` — a crafted value cannot break out of the attribute.
- **DOCX export** (`export/docx/marks.ts`): highlight maps to the fixed OOXML named colour `yellow` (docx highlight only supports named colours), so no attacker-controlled hex reaches the OOXML. (Note: the DOCX highlight colour is therefore *not* round-tripped losslessly — see suggestions — but this is pre-existing and not a security issue.)
- **Live DOM**: rendered via extension-highlight's `background-color` style attr, standard TipTap handling, unchanged here.

No new sink is introduced.

### Tests ✅
Comprehensive and meaningful: swatch order/count, preset apply, custom-input commit-on-change, drag-doesn't-commit + single-undo-step (with a dedicated history-enabled editor), ✕ clear from both range and collapsed caret, and cross-scope isolation (text-colour palette still 10, highlight still 10). The undo-depth test uses a fresh `StarterKit` with history enabled — correct, since the shared `beforeEach` editor disables `undoRedo`.

## Suggestions (non-blocking, nits)

- **Custom picker opens at `#000000`, not the caret's current highlight.** The `<input type="color">` has no `defaultValue`, so re-opening it on already-highlighted text starts at black; confirming without moving the hue would overwrite the highlight with black. Minor UX gap. Consider `defaultValue={editor.getAttributes('highlight').color || '#ffff00'}`. Note this exact behavior already ships in the #719 text-colour picker, so this is a parity nit, not a regression introduced here — worth fixing in both together if pursued.
- **The comment claims highlight colours are "lossless through … the DOCX/Markdown exporters."** Accurate for Markdown, but DOCX collapses every highlight to `yellow` (`export/docx/marks.ts`). Consider softening the comment so it doesn't over-promise DOCX fidelity.

## Coverage notes / what a human may want to verify

- Actual native OS colour-dialog rendering and the `change`-vs-`input` firing sequence vary across browsers/OSes; the tests simulate the DOM events but can't exercise a real OS picker. Worth a quick manual smoke on the target browsers.
- Contrast/readability of the 5 new light swatches against dark body text was assessed by value only (all are light pastels) — a designer eye on the final palette wouldn't hurt, but nothing here blocks.

Jerry-Xin (approved):
APPROVE — re-review at the new head after the rebase + the added collapsed-caret clear fix. The custom highlight color path stays constrained to browser-guaranteed values, so the CSS-injection surface I cleared previously is not re-opened.

## What changed since my prior approve
The branch was rebased (prior single feature commit is superseded), and one **new** commit was added on top of the re-based feature commit:

- Feature commit (re-based, unchanged in substance): `packages/docs/src/editor/Toolbar.tsx` expands `HIGHLIGHT_COLORS` from 5 to 10 fixed `#rrggbb` literals and adds a custom highlight picker.
- New fix commit `fix(docs): clear highlight from a collapsed caret`: `packages/docs/src/editor/Toolbar.tsx` — the ✕ clear button now runs `extendMarkRange('highlight').unsetHighlight()` instead of bare `unsetHighlight()`, plus a regression test in `packages/docs/src/editor/Toolbar.test.tsx`.

(The larger file list in a raw base..head diff — LineHeight, paragraph-spacing, docx spacing, config font/line-spacing flags — is merge-base drift pulled in by the rebase, not part of this PR; verified per-commit.)

## Color-value provenance (the crux) — every path to the applied highlight
The highlight mark is the stock TipTap Highlight extension configured `multicolor: true`; the applied color can only originate from:

1. **Preset palette** — ten compile-time `#rrggbb` string literals in source, applied via `toggleHighlight({ color: <literal> })`. Fixed values, not user-authored. Safe.
2. **Custom picker** — a native color input element. The browser constrains its value to `#rrggbb`; that value is read and passed to `setHighlight({ color })` on the committed change event. An arbitrary free-form string cannot reach the mark through this element by construction.
3. **✕ clear** — removes the mark; contributes no color value.

There is **no free-form text entry, no eyedropper, no persisted/recent-colors store** in the highlight path, so there is no unsanitized-string-into-style vector. The new fix commit only removes a mark and adds no color source.

## Render mechanism
Color reaches the document only as the Highlight mark's `color` attribute (values from the two safe sources above), and swatch previews use a React style object bound to the fixed palette literals. No user string is interpolated into a style attribute or CSS variable.

## Re-scan (not narrowed to the color path)
- Listener lifecycle: the picker's raw change listener is registered in an effect with a matching cleanup and correct deps — no leak.
- No new timers/portals; no race introduced by the clear-fix.
- i18n: the reused custom-color key is present and symmetric across both locales with real values.
- Tests: the new commit adds a regression test asserting the mark actually leaves the document from a collapsed caret; existing suite covers palette contents/order, preset apply, custom-input apply, single-undo on drag, and clear.

Note: the safety of the custom path is by construction (native input type + fixed literals), not dependent on CSS-grammar enforcement, so it holds independent of the jsdom test environment.

## Highlights
Zero new dependencies; single-transaction commit keeps one pick = one undo step; export stays lossless because values remain standard hex; the clear fix matches the established mark-clearing idiom.

No blocking items. Approving at the current head.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #732 (octo-web)

## Summary

This PR expands the highlight (text-background) palette from 5 to 10 pastel presets and adds a native `<input type="color">` custom highlight picker, mirroring the text-colour control introduced in #719. It also fixes a regression (XIN-1022) where clicking the ✕ clear button with a collapsed caret inside a `<mark>` left the highlight in the document. The fix chains `extendMarkRange('highlight')` before `unsetHighlight()` to grow the selection to cover the surrounding mark. Comprehensive tests cover preset ordering, custom-pick commit semantics, drag-vs-commit isolation (single undo step), collapsed-caret clear, and cross-control isolation (text-colour swatches stay at 10).

No P0/P1 issues found. The implementation is clean, well-scoped, and thoroughly tested.

## Verification

Static analysis only at head `9adfc79840c785e1026587cca3f76cb504334868`; build and tests not executed in this environment.

- ✅ **Diff scope** — 2 files, +216/−5. Production change is localised to `HighlightControl` in `packages/docs/src/editor/Toolbar.tsx`; all other changes are tests.
- ✅ **i18n key exists** — `docs.toolbar.customColor` is defined in both `en-US.json` (`"Custom color"`) and `zh-CN.json` (`"自定义颜色"`). No missing locale key.
- ✅ **CSS reuse** — `.octo-color-custom`, `.octo-color-custom-input` CSS rules already exist in `styles.css:299-334` from the text-colour PR (#719). The new `octo-highlight-color-popover` class is a test-only selector layered on top of the existing `octo-color-popover` — no new styling needed.
- ✅ **useEffect lifecycle** — The `[editor, open]` dependency array correctly re-binds the native `change` listener when the popover mounts/unmounts. Cleanup on a detached DOM node is a safe no-op. No listener leak.
- ✅ **input.value safety** — `<input type="color">` always normalizes to a valid 7-char lowercase `#rrggbb` hex per HTML spec. No sanitization needed; no XSS vector.
- ✅ **extendMarkRange correctness** — TipTap built-in chain command, available for any registered mark extension. The `Highlight` extension is registered (tests confirm `Highlight.configure({ multicolor: true })`). No-op when selection already spans the mark.

## Findings

No P0/P1 issues; no P2 issues. One pre-existing parity observation noted below for awareness (not blocking).

### Data-Flow Trace

| Consumer | Source | Runtime path verified? |
|---|---|---|
| Preset swatch `onClick` → `toggleHighlight({ color: c })` | `HIGHLIGHT_COLORS` const array (`Toolbar.tsx:283-294`) | ✅ — Array mapped to `<button>` elements; `c` is always a valid hex literal. |
| Custom picker `onCommit` → `setHighlight({ color: input.value })` | Native `<input type="color">` (`Toolbar.tsx:385-388`) | ✅ — Browser normalizes to `#rrggbb`; `change` event fires once on pick commit. |
| ✕ clear → `extendMarkRange('highlight').unsetHighlight()` | TipTap chain commands | ✅ — `extendMarkRange` grows selection to mark bounds; `unsetHighlight` removes mark from selection. Works for both collapsed caret and range selection. |
| Test `fireEvent.change(input, ...)` → `onCommit` | `input.addEventListener('change', ...)` (`Toolbar.tsx:330`) | ✅ — `fireEvent.change` dispatches native `change`; ref-bound listener fires. Confirmed by test assertions. |

### Blind-Spot Checklist (C1–C6, security_sensitive routing)

- **C1 — Dual-path parity:** Highlight ✕ now uses `extendMarkRange('highlight').unsetHighlight()` (`Toolbar.tsx:370`). The text-colour ✕ at `Toolbar.tsx:442` still uses plain `unsetColor()`. The text-colour path may have the analogous collapsed-caret bug for `textStyle` marks. **Pre-existing** (not introduced or amplified by this PR per R2). Not blocking — worth a follow-up ticket.
- **C2 — Control-flow ordering / nested reuse:** Single `useEffect` per control instance. No nested/double-call risk. `open` dependency ensures correct bind/unbind on mount/unmount. **Clear.**
- **C3 — Authorization boundary:** N/A — no auth/permission/endpoint changes.
- **C4 — Authorization lifecycle:** N/A — no auth changes.
- **C5 — Build/note ≠ runtime:** N/A — no build/CI/packaging changes. Source + tests only.
- **C6 — Governance/docs self-consistency:** N/A — no governance docs changed.

### Cross-Round Blocker Re-Check (R6)

- **Preset ordering** — test at `Toolbar.test.tsx:451-465` pins the exact 10-colour sequence matching `HIGHLIGHT_COLORS`. No drift between production array and test assertion.
- **Drag-vs-commit isolation** — `useEffect` binds only the native `change` event (not `input`). Tests at `Toolbar.test.tsx:488-520` and `:528-568` confirm: `input` events produce zero transactions; `change` produces exactly one transaction + one undo record.
- **Collapsed-caret clear regression (XIN-1022)** — `extendMarkRange('highlight')` before `unsetHighlight()` is the correct TipTap idiom (same pattern as `extendMarkRange('link').unsetLink()`). Test at `Toolbar.test.tsx:580-594` verifies: collapsed caret at position 3 inside a highlighted "hello" → ✕ removes the `<mark>` from the document.
- **Cross-control isolation** — text-colour test asserts highlight stays at 10 swatches (`Toolbar.test.tsx:429`); highlight test asserts text-colour stays at 10 swatches (`Toolbar.test.tsx:596-600`). No cross-contamination.
- **`toggleHighlight` vs `setHighlight`** — preset swatches use `toggleHighlight` (click same colour toggles off); custom picker uses `setHighlight` (always apply). Intentional asymmetry matching user expectation.
- **`onMouseDown` preventDefault** — present on both preset swatches and the custom picker `<label>`, preventing popover close on internal click. Consistent with text-colour control.

## Verdict: APPROVED

Clean, well-tested PR with no correctness, security, or build concerns. The highlight palette expansion, custom picker, and collapsed-caret fix are all correctly implemented and thoroughly covered by tests. The one pre-existing parity observation (text-colour ✕ lacking `extendMarkRange`) is out of scope for this PR and can be tracked separately.





[Octo-Q] verdict: APPROVE — No P1/P2 findings. Highlight palette expansion, custom picker, and XIN-1022 collapsed-caret fix are all correctly implemented with comprehensive test coverage. Pre-existing text-colour ✕ parity observation is out of scope.

yujiawei (approved):
# Code Review — PR #732 (octo-web)

**Head SHA reviewed:** `9adfc79840c785e1026587cca3f76cb504334868`
**Scope:** `packages/docs/src/editor/Toolbar.tsx` (+66/-5), `packages/docs/src/editor/Toolbar.test.tsx` (+155/…). Front-end only; no schema/backend/exporter code touched.

This PR was routed as security-sensitive (keyword match on the highlight/colour value flowing toward export sinks). I reviewed it accordingly and independently verified the data flow end to end. Findings below are cross-checked across two independent analysis passes; where they disagreed I re-derived the answer from source.

---

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the description states, and nothing beyond it:

- **Palette 5 → 10** — `HIGHLIGHT_COLORS` (`Toolbar.tsx:283`) grows to 10 light `#rrggbb` tints; the original five are preserved. Order and contents are locked by a test.
- **Native custom picker** — a `<input type="color">` bound via a ref (`Toolbar.tsx:323–334`) that commits once on the DOM `change` event with `setHighlight({ color })` and closes the popover. Zero new dependencies; reuses the `.octo-color-custom` / `.octo-highlight-color-popover` styles introduced by the earlier text-colour work, so no new CSS reaches the linter.
- **Presets use `toggleHighlight`, custom uses `setHighlight`** — consistent with the already-merged text-colour control.

One in-scope behaviour change is bundled but not named in the title (non-blocking, see §2 note): the `✕` clear now does `extendMarkRange('highlight').unsetHighlight()` (`Toolbar.tsx:370`) instead of a bare `unsetHighlight()`. This is a genuine bug fix (a collapsed caret inside a `<mark>` previously left the mark in the document) and is directly regression-tested.

No under-build, no out-of-scope feature/flag/field creep.

---

## 2. Code quality

**Quality: Approved**

Verified correct:

- **Effect lifecycle (`Toolbar.tsx:324–335`)** — deps `[editor, open]` are correct: the `<input>` renders only while the popover is open, so the effect must re-run on the open transition to bind to the freshly-mounted node. `addEventListener('change', …)` is matched by a symmetric cleanup on the same captured node — no listener leak, no double-fire. `input.value` is read live inside the handler, so there is no stale-closure hazard.
- **Drag vs commit** — binding the raw DOM `change` (not React's synthetic `onChange`, which folds onto native `input`) is the right call to avoid committing on every intermediate hue. The tests prove it: dragging (`input`) produces zero document changes; releasing (`change`) produces exactly one transaction and one undo step.
- **`✕` clear correctness** — `extendMarkRange('highlight').unsetHighlight()` correctly removes the mark for both a collapsed caret and an active range. See the note below on why this does **not** damage broad multi-highlight selections.
- **Tests** — all 8 new tests assert real behaviour (none pass vacuously). The single-undo test correctly spins up a fresh history-enabled editor; the collapsed-caret regression test genuinely fails against the old `unsetHighlight()`-only code.

### Note on `extendMarkRange` and multi-highlight selections (verified, not a defect)

A reasonable concern is that `extendMarkRange('highlight')` might shrink a broad selection (spanning several highlights) down to the first mark and leave the rest uncleared. I checked the pinned dependency (`@tiptap/core@3.22.2`). Its `extendMarkRange` only rewrites the selection when the found mark range **fully contains** the current selection:

```js
if (range && range.from <= from && range.to >= to) {
  tr.setSelection(TextSelection.create(doc, range.from, range.to))
}
```

For a broad selection that starts inside one highlight and extends past it, `getMarkRange($from)` returns only that first mark, whose `.to < to`, so the guard is false and `extendMarkRange` is a **no-op** — `unsetHighlight()` then runs against the user's full original selection and clears every highlight in range. So the fix is safe for the multi-highlight case as well as the collapsed-caret case it targets. No change needed.

### Suggestions (non-blocking)

- **P2 — Misleading "lossless" claim (DOCX).** The PR description and two new source comments (`Toolbar.tsx:282`, `Toolbar.tsx:375`) state the highlight colour is "kept lossless by the DOCX / Markdown exporters." Markdown export is indeed lossless (`export/markdown.ts` emits an escaped `background-color` style). DOCX is **not**: `export/docx/marks.ts` hardcodes `opts.highlight = 'yellow'` and ignores the colour attribute, so all ten presets — and any custom hex — render as yellow in Word. This is pre-existing behaviour and out of scope to fix here, but expanding the palette 5 → 10 (plus an unbounded custom picker) widens the surface of that silent gap, and the "lossless through DOCX" wording is now baked into durable code comments where the sibling `textStyle` path (which *does* call `normalizeDocxColor`) makes the copy-paste origin visible. Recommend scoping the claim to Markdown and noting the DOCX limitation, e.g. "Markdown preserves the colour; DOCX highlight is limited to named colours and renders as yellow (pre-existing)."
- **nit — Custom picker opens at black.** The `<input type="color">` is uncontrolled with no `defaultValue`, so opening it while the caret sits in an existing highlight shows `#000000` rather than previewing the current colour. Consider `defaultValue={editor.getAttributes('highlight').color || '#000000'}`. Note this matches the already-merged text-colour picker (same omission), so it is a consistency-level polish item, not a regression.
- **nit — Bundled `✕` fix not called out.** The collapsed-caret clear fix is a third, semantically distinct change beyond "palette + picker." A one-line mention in the description/changelog would help reviewers and release notes.

---

## 3. Security assessment

No real attack surface is introduced by this diff. A native `<input type="color">` yields only a browser-sanitised lowercase `#rrggbb` string — arbitrary text assigned to its value collapses to a valid colour — so no payload can be smuggled through the picker into `setHighlight`. Downstream, the Markdown sink additionally escapes the value via `escapeHtmlAttr` (with an existing XSS-breakout test), and the DOCX sink discards the value entirely. The security-sensitive routing is a keyword-driven false positive here rather than a genuine new vector; this change does not widen surface beyond the previously-merged text-colour picker.

**For a human to be aware of:** the highlight `color` attribute could in principle receive a non-hex value through a non-UI path (remote Yjs sync, paste/import) — but that is outside this diff, and the Markdown exporter escapes the attribute regardless.

---

## 4. Coverage / blind spots

- I did **not** execute the test/CI suite in this environment; the 8 tests were assessed by reading them against the shipped implementation and are logically sound, but "green in CI" is asserted by the author, not re-run here.
- The `docs.toolbar.customColor` i18n key is reused (no new keys) and assumed present in the locale files — not re-verified.
- Real-browser OS colour-dialog behaviour (focus/selection restoration across the native dialog) is inferred from ProseMirror semantics and the mirrored merged control, not observed live.
- DOCX highlight-colour round-trip has no test at all (relatedly, the §2 P2 above).

---

## Verdict

No P0/P1 issues. The one cross-analysis P1 candidate (multi-highlight clear) was refuted against the pinned TipTap source. Remaining items are a P2 documentation-accuracy fix and two nits — none block merge.

**APPROVED**

PR #733, opened 2026-07-14
feat(docs): default font-family selector to ON with build-arg override

ploy-elison:
## What

Default the docs font-family toolbar selector to **ON**, while keeping a build-arg override for emergency rollback.

- **Source default flip** (`packages/docs/src/config.ts`): `FONT_FAMILY_ENABLED` now falls back to `'true'` when `VITE_DOCS_FONT_FAMILY` is unset, so a plain build shows the font selector by default. The flag itself is retained as a kill switch.
- **Dockerfile passthrough**: added `ARG`/`ENV VITE_DOCS_FONT_FAMILY` so a build-arg can override the new default (e.g. `--build-arg VITE_DOCS_FONT_FAMILY=false`).

## Why

Product decision: the font-family feature should be available by default rather than hidden behind a default-OFF flag that needs an explicit build-arg to enable. Client bundle versions have converged (#700 shipped everywhere), so the phased-rollout gate can safely default open. Keeping the flag (not removing the gate) preserves a low-risk, one-arg rollback.

## Behavior matrix

| `VITE_DOCS_FONT_FAMILY` | `FONT_FAMILY_ENABLED` | Selector |
|---|---|---|
| unset | `true` | visible (new default) |
| empty string (Docker `ENV` with no build-arg) | `true` | visible |
| `false` | `false` | hidden (rollback intact) |
| `true` | `true` | visible |

The empty-string case matters because the Dockerfile sets `ENV VITE_DOCS_FONT_FAMILY=${ARG}` unconditionally; `envOr` normalizes blank to the fallback, so a Docker build with no build-arg still resolves to ON.

## Verification

- `pnpm exec vitest run` on the docs config + font-family suites — 55/55 pass.
- Production build with **no build-arg** → `VITE_DOCS_FONT_FAMILY` is not baked into the bundle (`undefined` → `envOr('true')` → ON); font selector present.
- Production build with **`VITE_DOCS_FONT_FAMILY=false`** → bundle bakes `VITE_DOCS_FONT_FAMILY:`false`` → flag resolves OFF; selector hidden. Rollback path confirmed.

## Scope

Font-family only. Line-height gating is intentionally left untouched (its change is not merged to `main` yet). No font-family business logic changed — this is default value + build-config only.

Opened as **draft** for review; not for merge yet.


mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #733 (octo-web)

## Summary

This PR flips the `FONT_FAMILY_ENABLED` feature flag default from OFF to ON (`packages/docs/src/config.ts:127`) and adds the corresponding `VITE_DOCS_FONT_FAMILY` build-arg pass-through to the Dockerfile. The rationale is that client versions have converged after #700 shipped SCHEMA_VERSION 16 to every bundle, so the font-family toolbar selector can now be shown by default. The flag is retained as an emergency kill switch — an explicit `VITE_DOCS_FONT_FAMILY=false` build still hides the selector.

The change is minimal, well-scoped, and the kill-switch semantics are correct. Two stale documentation comments in unchanged files should be cleaned up in the same PR.

## Verification

- ✅ **Default-ON logic** — `envOr(import.meta.env?.VITE_DOCS_FONT_FAMILY, 'true') === 'true'` correctly resolves to `true` when the env var is unset/blank/whitespace (the `envOr` helper at `config.ts:16-20` trims and falls back to `'true'`).
- ✅ **Kill switch** — Setting `VITE_DOCS_FONT_FAMILY=false` at build time → `envOr('false', 'true')` returns `'false'` (non-empty, no fallback) → `'false' === 'true'` → `false`. Rollback path intact.
- ✅ **Dockerfile build-arg** — The `ARG VITE_DOCS_FONT_FAMILY` / `ENV VITE_DOCS_FONT_FAMILY=${VITE_DOCS_FONT_FAMILY}` pair at `Dockerfile:14-15` matches the exact pattern of sibling args (`VITE_DOCS_DEFAULT_SPACE`, `VITE_DOCS_DEFAULT_FOLDER`, `VITE_DOCS_DEFAULT_DOC`, `VITE_DOCS_ASSET_HOSTS`) and is correctly placed before the `RUN pnpm install && build` step.
- ✅ **Data flow — toolbar** — `Toolbar.tsx:1021` gates `<FontFamilySelect>` on `FONT_FAMILY_ENABLED`. With the new default, the selector renders unless explicitly killed. Correct.
- ✅ **Data flow — paste write path** — `extensions.ts:111` (`parseHTML` backstop) returns `null` only when `!FONT_FAMILY_ENABLED`; `extensions.ts:128` (`transformPastedHTML`) passes HTML through when flag is on, strips when off. Both correctly track the new default.
- ✅ **Tests independent of default** — `fontFamilyPaste.test.ts:169,285` explicitly stubs `VITE_DOCS_FONT_FAMILY` via `vi.stubEnv` and resets modules before each test run. Tests do not rely on the compile-time default and remain valid.
- ✅ **EXTERNAL-BOUNDARY (Dockerfile)** — Build-arg naming, placement, and pass-through are consistent with the established pattern. No new attack surface or misconfiguration vector introduced.

## Findings

No P0/P1 issues. Two P2 items in unchanged files where stale "default off" comments should be updated.

### P2 — Stale "default off" comment in schema version log (`packages/docs/src/schema/index.ts:55`)

The SCHEMA_VERSION 16 changelog comment reads: "the toolbar entry that SETS it is gated behind FONT_FAMILY_ENABLED (default off) for the phased rollout." This is now factually incorrect — the default is ON as of this PR. Future readers consulting the schema version history will get the wrong impression about the current state. Update to "(default on)" or remove the default-value note entirely.

### P2 — Stale "default off" comment in FontFamilySelect JSDoc (`packages/docs/src/editor/Toolbar.tsx:532`)

The `FontFamilySelect` doc comment reads: "Rendered ONLY when FONT_FAMILY_ENABLED is on (feature flag, default off)." Same staleness issue — the default is now ON. Update to "(feature flag, default on)" or just "(feature flag)".

## Blind-spot Checklist

- **C1 (dual-path parity)** — N/A. No paired add/remove or subscribe/unsubscribe paths changed.
- **C2 (control-flow ordering / nested reuse)** — Clear. `FONT_FAMILY_ENABLED` is a module-level constant resolved once at load time; no ordering or nesting concerns.
- **C3 (auth boundary)** — N/A. No auth/permission changes.
- **C5 (build ≠ runtime)** — Clear. The `envOr` helper correctly handles the Vite build-arg → empty-string behavior, and the Dockerfile pass-through matches the established pattern. Verified at the source level.
- **C6 (governance/docs self-consistency)** — Hit (the two P2 findings above). The config.ts comment is updated, but two downstream comments still say "default off".

## Data Flow Trace

| Consumer | Location | Source | Runtime value (default) | Correct? |
|---|---|---|---|---|
| Toolbar render gate | `Toolbar.tsx:1021` | `FONT_FAMILY_ENABLED` from `config.ts:127` | `true` | ✅ selector renders |
| Paste parseHTML backstop | `extensions.ts:111` | same | `true` → `!true` = `false` → does NOT return null → fontFamily parsed | ✅ allows write |
| Paste transformPastedHTML | `extensions.ts:128` | same | `true` → identity (no strip) | ✅ preserves fonts |

## Verdict: APPROVED

No correctness, security, or build-breaking issues. The default flip is logically sound, the kill switch works, tests are independent of the default, and the Dockerfile wiring is consistent. The two P2 stale comments are non-blocking but should be cleaned up in this PR to keep documentation self-consistent.

Static analysis only at head `86d175ab3b41458f9f3d97a9f9c85a8032d46269`; build and tests not executed in this environment.

[Octo-Q] verdict: APPROVE — No P0/P1 blockers. Two P2 stale comments (schema/index.ts:55, Toolbar.tsx:532) say "default off" but the flag is now default ON; non-blocking, recommend cleanup in same PR.

OctoBoooot (approved):
## Review: #733 — default fontFamily selector to ON with build-arg override

**Verdict:** Approve. A correct, minimal completion of the #700 phased rollout: flip `FONT_FAMILY_ENABLED` default OFF→ON, retain the flag as a kill switch, pass the ARG through the Dockerfile. Mechanics verified; the safety premise holds at the code level. Risk tier: normal (a config default flip; no schema/logic change). CI green (docker-lint + scans pass); PR is draft.

### Verified — mechanics correct
- `FONT_FAMILY_ENABLED = envOr(import.meta.env?.VITE_DOCS_FONT_FAMILY, 'true') === 'true'`. `envOr` (`config.ts:17-19`) returns the trimmed value if non-empty, else the fallback. So: **unset/empty → 'true' → ON** (new default); **explicit `=false` → OFF** (kill switch intact); the rollback path is preserved exactly as the comment claims.
- Dockerfile adds `ARG`/`ENV VITE_DOCS_FONT_FAMILY` passthrough so a build can still force `=false`. hadolint/shellcheck green.

### Verified — the safety premise (this is the real question, not the one-liner)
Flipping the fontFamily *write* path to default-ON is only safe if the pre-convergence hazard is retired. It is, at the code level:
- **#700 is MERGED** (`2026-07-14T06:52`) with the CSS-escape paste-gate fix — `normalizeCssPropName` + `declSetsFontFamily` are present in `sanitize.ts` at this head. So the sanitizer that blocked flag-off writes is sound.
- The flag gates the paste sanitizer **inversely**: `transformPastedHTML: FONT_FAMILY_ENABLED ? html : stripPastedFontFamily(html)` (`extensions.ts:128`) and the parseHTML backstop `if (!FONT_FAMILY_ENABLED) return null` (`:111`). So ON = pasted font-family flows into the Y.Doc **by design** (fontFamily is now a legitimate writable attr post-convergence), OFF = still stripped. Flipping ON doesn't expose a leak — it completes the rollout the sanitizer was guarding during the pre-convergence window. The CSS-escape fix still matters for any `=false` rollback build.

### Note (maintainer judgment, not a code defect)
The flip's correctness rests on the operational claim in the comment — "client versions have now converged (#700 shipped to every bundle)." That's an ops/deployment assertion I can't verify from the diff (are all *deployed* clients actually on the #700 bundle?). The kill switch (`=false`) is the mitigation if that turns out premature. Worth the maintainer confirming convergence is real before this lands; the code itself is correct either way.

### Net
Clean rollout-completion flip: envOr fallback correct, kill switch preserved, Dockerfile passthrough sound, and the safety premise holds (#700 merged with the escape fix; ON makes the sanitizer a flag-off-only backstop, which is the intended post-convergence state). Approving. Still draft, so not mergeable regardless; the only non-code gate is confirming client convergence is real (kill switch covers a rollback). Credit ploy-elison.

yujiawei (approved):
# Code Review — PR #733 (octo-web)

**Verdict: APPROVED** — no P0/P1 blockers. Two non-blocking nits (stale doc-comments; missing default-value test).

## Summary of change
Flips the docs toolbar font-family selector default from OFF to ON while retaining `VITE_DOCS_FONT_FAMILY` as an emergency kill switch, and adds the corresponding Dockerfile `ARG`/`ENV` passthrough.

- `packages/docs/src/config.ts:127` — `envOr(import.meta.env?.VITE_DOCS_FONT_FAMILY, '')` → `envOr(..., 'true')`, so an unset/blank value now resolves to ON.
- `Dockerfile:14-15` — new `ARG VITE_DOCS_FONT_FAMILY` / `ENV VITE_DOCS_FONT_FAMILY=${VITE_DOCS_FONT_FAMILY}`, mirroring the existing `VITE_DOCS_*` build args, before `pnpm turbo run build`.

## Spec compliance — PASS
Verified against head SHA `86d175ab3b41458f9f3d97a9f9c85a8032d46269`:
- **Default flip ON**: implemented correctly. Because `envOr` (config.ts:17-20) trims and collapses both empty-string and `undefined` to the fallback, every no-value path resolves to `'true'` → ON: unset, whitespace, and the Docker `ENV X=${ARG}` empty-string case (when no build-arg is passed).
- **Kill switch retained**: the `=== 'true'` comparison is unchanged, so an explicit `VITE_DOCS_FONT_FAMILY=false` build still hides the selector. Rollback path intact.
- **Dockerfile passthrough**: added, matching the existing pattern; placed in the builder stage before the build step so Vite bakes it correctly.
- **Scope**: font-family only. No line-height change, no font business-logic change, no new flags/fields. No over-build, no deviation.

## Correctness — PASS
- **No data-loss regression.** The `FontFamily` schema extension is registered *unconditionally* in both the live editor (`LiveFontFamily`, extensions.ts:195) and the read-only preview (extensions.ts:310), so every bundle round-trips a stored `fontFamily` attr regardless of the flag. `FONT_FAMILY_ENABLED` gates only the three WRITE paths (toolbar render `Toolbar.tsx:1021`, `parseHTML` backstop `extensions.ts:111`, paste strip `extensions.ts:128`). Flipping the default opens writes; the #700 "old client silently strips a font" hazard is guarded by (a) client version convergence (#700 shipped everywhere) and (b) the retained `=false` kill switch. This one precondition — convergence — is a deployment fact not verifiable from source; nothing in the code contradicts it.
- **Production default confirmed.** `octo-web`'s `docker-publish.yml` build step passes no build-args, so the shipped bundle takes the empty-string → `'true'` → ON path as intended. There is no runtime injection (`nginx.conf.template` / `docker-entrypoint.sh` never reference the var). I also checked the downstream `octo-deployment` repo: it consumes the pre-built `mininglamposs/octo-web:latest` image and sets **no** `VITE_DOCS_FONT_FAMILY` build-arg, so no environment silently overrides the new default to OFF.
- **No build-break, tests insulated.** The change is a one-literal edit plus mirrored Dockerfile lines (TypeScript-safe). `config.test.ts` asserts nothing about this default; `fontFamilyPaste.test.ts` explicitly stubs the env, so the existing suite is unaffected.

## Non-blocking nits (not required for merge)
1. **Stale doc-comments about the default.** `packages/docs/src/schema/index.ts:55` and `packages/docs/src/editor/Toolbar.tsx:532-533` still describe the flag as "default off". The PR rewrote the rationale in `config.ts` but left these two sibling comments referencing the old default — now internally inconsistent documentation of a load-bearing flag. Suggest updating both to reflect "default ON, `=false` kill switch".
2. **No regression guard on the new default.** The exact line this PR changes has zero test coverage (`config.test.ts` never asserts `FONT_FAMILY_ENABLED`; the paste tests stub the env and so can't catch a default regression). A future revert of the fallback string would ship OFF-by-default with a green build. Consider a small `config.test.ts` assertion: unset/blank → `true`, `'false'` → `false` (also locks the kill switch).

## Note on the PR description
The body says "Opened as draft … not for merge yet", but the PR is not in GitHub Draft state (`isDraft: false`). Treating this as stale prose rather than a merge signal; if merge should genuinely wait, please reconcile the description. The code itself is approve-quality.

Jerry-Xin (approved):
Defaulting the docs font-family selector to ON with a build-arg kill switch. Verdict: safe to ship — the write surface is closed and the override wiring is correct.

### Default-ON safety (the crux)
The selector's two write paths are both closed, so shipping it enabled by default does not open a CSS-injection vector:

- **Toolbar write** — the dropdown only offers a fixed, developer-defined whitelist (`editor/fontFamilies.ts`, 8 preset font stacks). There is no free-text field, so a user cannot set an arbitrary value from the toolbar.
- **Paste write** — the paste-gate work is present on this head. `editor/sanitize.ts:stripPastedFontFamily` strips inline font-family from pasted HTML by deciding on the resolved property (name-normalization plus real CSSOM), covering the comment/escape/entity/casing variants that a raw byte-compare would miss; and `editor/extensions.ts` `LiveFontFamily` adds a `parseHTML` backstop plus a `transformPastedHTML` gate. The stored value is read as the browser-resolved `element.style.fontFamily`, which the CSSOM normalizes, and is re-emitted through Tiptap's textStyle mark — so a value cannot escape its declaration.

Net: with the flag ON, pasted fonts are preserved (gate skipped) but every value that reaches the shared doc is a CSSOM-resolved family, not a raw attacker string. Injection surface is closed; default-ON is safe.

### Build-arg override wiring — correct
`config.ts`: `FONT_FAMILY_ENABLED = envOr(import.meta.env?.VITE_DOCS_FONT_FAMILY, 'true') === 'true'`.
- Unset / blank / whitespace → `envOr` trims and falls back to `'true'` → ON (the new default). ✅
- Explicit `VITE_DOCS_FONT_FAMILY=false` → OFF (emergency kill switch intact). ✅
- No inverted logic; the truthy compare is against the literal `'true'`.
`Dockerfile` adds the matching `ARG`/`ENV VITE_DOCS_FONT_FAMILY` pair alongside the sibling docs vars, so the value threads from build-arg to the Vite build correctly.

### Compat / persisted docs
No schema change — the FontFamily extension is registered unconditionally (SCHEMA_VERSION 16), and the flag has always gated only the write paths, not the read/round-trip path. Docs authored while the flag was OFF simply carried no font; flipping the default ON does not re-render or alter their stored content. The prior flag-off data-loss hazard (older bundles stripping the attr) no longer applies now that clients have converged.

### 🟡 Non-blocking
- `editor/Toolbar.tsx` — the `FontFamilySelect` docstring still says "feature flag, default off"; that is now stale. Worth a one-line touch-up to match the new default.

### ✅ Highlights
- Retaining the flag purely as a kill switch (explicit `false` still hides the selector) is the right call for a default-flip.
- The paste-gate's resolved-property approach is the correct hardening — it does not rely on brittle raw-name matching.

CI is green (Build, pr-title-lint, scans all pass). Approving.

PR #739, opened 2026-07-14
fix(loop): filter panel dropdowns hidden behind the panel; stale agent-status cache

### LINKED ISSUE #738: 🐛 [loop] 筛选/显示面板内的下拉(状态/优先级/负责人等)被面板遮挡,看不到也点不了
## Bug 描述

回路(loop)列表页右上角「筛选」「显示」面板里的每个下拉控件(状态、优先级、负责人、创建者、时间字段的 Select + 时间范围的 DatePicker + 显示面板的排序 Select),点开后**选项浮层出现在面板背后、被面板本体盖住**,用户看不到也点不到,筛选功能实际不可用。

## 复现步骤

1. 进入「回路」→ 任一有数据的工作区,看板或列表视图
2. 点右上角「筛选」打开筛选面板
3. 点「状态」下拉

**预期**:弹出 7 个状态选项供选择
**实际**:选项列表渲染在了面板下层(DOM 里存在、`display:block`,但 z-index 低于面板 → 被面板内容遮挡),看不见、点不到

优先级/负责人/创建者/时间范围(DatePicker)/显示面板排序下拉同理。

## 根因

面板是 Semi `<Dropdown>`(popover 默认 z-index **1060**),而其内部的 Select / DatePicker 弹层走 Semi 默认基准 z-index **1030**,并 portal 到 `document.body`(脱离面板的层叠上下文)。1030 < 1060 → 子弹层被面板盖住。Semi 并不知道这些下拉"嵌"在面板里,给了它们更低的默认层级。

## 修复

给面板内每个字段弹层显式设置高于面板的 z-index(通过 Semi 的 `zIndex` prop),使其正常浮在面板之上。已实测:修复后各下拉均在最上层、可见可点、定位正确(触发框正下方),且点选项**不会误关**筛选面板(面板受控、仅外部点击关闭),点击外部仍正常收起。

## 影响范围

- 仅 `packages/dmloop`(loop 列表页筛选/显示面板)
- 纯前端、纯展示层修复,不涉及后端/协议/数据

## 环境

- Web 端,当前 main
- Semi UI (@douyinfe/semi-ui)

MrAladdin:
Closes #738

## What

Two scoped fixes in `packages/dmloop`:

### 1. Filter/show panel dropdowns rendered behind the panel (#738)
Every `<Select>` and the `<DatePicker>` inside the toolbar「筛选」/「显示」`<Dropdown>` panel had its option popup rendered **behind** the panel, so options were invisible and unclickable — the filter was effectively unusable.

Root cause: the panel popover is Semi's default z-index **1060**, while Semi `Select`/`DatePicker` popups default to base z-index **1030** and portal to `document.body` (escaping the panel's stacking context). 1030 < 1060 → child popups paint behind the panel.

Fix: a shared `const FIELD_POPUP = { dropdownClassName: "loop-fields__dropdown", zIndex: 2000 }` spread onto every filter/show Select, plus `zIndex={FIELD_POPUP.zIndex}` on the DatePicker — lifting each field popup above the panel. Scoped local to `IssuePage` (the only `<Dropdown>` panel with nested popups); the other `loop-fields__dropdown` Selects live in Modals and are unaffected.

Why the prop and not CSS / `getPopupContainer`: Semi writes the winning z-index **inline on the `.semi-portal`** node (a class rule can't override it without `!important`, and the DatePicker popup carries no shared class); `getPopupContainer` would render the popup *inside* the narrow panel, clipping the wide date-range picker and fighting the panel's outside-click close.

### 2. Stale agent-status cache (follow-up to #727 review)
The module-level agent-status map (drives the @-mention online dots) mirrors the runtime map but was never invalidated, so the online dots went stale. Now invalidated across the whole class of agent-set changes:
- **workspace switch** — via a shared `resetWorkspaceCaches()` (directory + runtime + agent-status), replacing the duplicated invalidation blocks in `applyWorkspace`/`switchWorkspace`;
- **agent create/update/archive/restore** — via `afterAgentMutation` in `agentApi`, so a created/restored agent's dot is fresh without a reload.

## Verification (real browser, dev env)

- All filter fields (5 Selects + DatePicker) now render on top (z-index 2000), correctly positioned below the trigger, and are clickable; selecting a value applies the filter and updates the active-filter badge.
- Clicking a field/option does **not** close the panel (panel is controlled, closes only on outside-click); outside-click still closes it. No regression to the existing open/close behavior.
- z-index 2000 sits above the panel (1060) and below all global overlays (modals 9999, toasts 10000); the only z-2000 peer is the @-mention menu, which never co-renders with the issue-list filter panel.
- `pnpm --filter @octo/loop test` green (25/25); typecheck clean.

## Scope / blast radius

- Frontend only, presentational + cache invalidation; no backend/protocol/schema change.
- `FIELD_POPUP` is local to `IssuePage`; `afterAgentMutation` only wraps the existing `afterDirectoryMutation` for agent writes; `resetWorkspaceCaches` is a behavior-preserving extraction of the existing three invalidation calls.

## Review

Ran `/simplify` (4 angles) + two adversarial rounds (parallel reviewers): z-index fix confirmed clean (no prop-conflict, correct stacking); the agent-status class-completion (agent create/restore trigger) was applied per the review's Rule-9 finding.

### DIFF
--- a/packages/dmloop/src/api/agentApi.ts
+++ b/packages/dmloop/src/api/agentApi.ts
@@ -62,22 +62,29 @@ export function invalidateAgentStatus(): void {
   _agentStatusPromise = null;
 }
 
+// agent 增改会改变工作区的 agent 集合,除清目录缓存外一并清 agent 状态缓存,
+// 否则新建/恢复的 agent 在 @ 菜单里读到陈旧的在线状态。
+function afterAgentMutation<T>(r: T): T {
+  invalidateAgentStatus();
+  return afterDirectoryMutation(r);
+}
+
 export function createAgent(req: CreateAgentReq): Promise<Agent> {
-  return httpPost<Agent>("/agents", req).then(afterDirectoryMutation);
+  return httpPost<Agent>("/agents", req).then(afterAgentMutation);
 }
 
 export function updateAgent(id: string, req: UpdateAgentReq): Promise<Agent> {
-  return httpPut<Agent>(`/agents/${id}`, req).then(afterDirectoryMutation);
+  return httpPut<Agent>(`/agents/${id}`, req).then(afterAgentMutation);
 }
 
 // 后端不支持 DELETE /agents/:id（405）；改用归档。
 export function archiveAgent(id: string): Promise<void> {
-  return httpPost<void>(`/agents/${id}/archive`, {}).then(afterDirectoryMutation);
+  return httpPost<void>(`/agents/${id}/archive`, {}).then(afterAgentMutation);
 }
 
 // 恢复已归档 agent（软删除的逆操作）。
 export function restoreAgent(id: string): Promise<void> {
-  return httpPost<void>(`/agents/${id}/restore`, {}).then(afterDirectoryMutation);
+  return httpPost<void>(`/agents/${id}/restore`, {}).then(afterAgentMutation);
 }
 
 /* ---------- 环境变量（密钥） ---------- */
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -64,6 +64,10 @@ interface Filters {
 
 const PAGE_SIZE = 50;
 
+// 筛选/显示字段都挂在工具栏 Dropdown 面板(z-index 1060)里,其弹层默认取 Semi 基准 1030 →
+// 会被面板本体盖在下面(选项看不见也点不到)。统一抬到面板之上。
+const FIELD_POPUP = { dropdownClassName: "loop-fields__dropdown", zIndex: 2000 } as const;
+
 // 「我的回路」复用本页：defaultView="grouped" + defaultScope="involves" 即只看与我相关。
 interface IssuePageProps {
   defaultScope?: IssueScope;
@@ -280,24 +284,24 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
       <div className="loop-fields__row">
         <div className="loop-fields__label">{t("loop.filter.status")}</div>
         {view === "grouped" ? (
-          <Select multiple value={f.gStatuses} onChange={(v) => update({ gStatuses: (v ?? []) as IssueStatus[] })} showClear maxTagCount={2} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
+          <Select multiple value={f.gStatuses} onChange={(v) => update({ gStatuses: (v ?? []) as IssueStatus[] })} showClear maxTagCount={2} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
         ) : (
-          <Select value={f.status} onChange={(v) => update({ status: v as IssueStatus | undefined })} showClear disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
+          <Select value={f.status} onChange={(v) => update({ status: v as IssueStatus | undefined })} showClear disabled={searching} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.status")}>{statusOpts}</Select>
         )}
       </div>
       <div className="loop-fields__row">
         <div className="loop-fields__label">{t("loop.filter.priority")}</div>
         {view === "grouped" ? (
-          <Select multiple value={f.gPriorities} onChange={(v) => update({ gPriorities: (v ?? []) as IssuePriority[] })} showClear maxTagCount={2} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
+          <Select multiple value={f.gPriorities} onChange={(v) => update({ gPriorities: (v ?? []) as IssuePriority[] })} showClear maxTagCount={2} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
         ) : (
-          <Select value={f.priority} onChange={(v) => update({ priority: v as IssuePriority | undefined })} showClear disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
+          <Select value={f.priority} onChange={(v) => update({ priority: v as IssuePriority | undefined })} showClear disabled={searching} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.priority")}>{priorityOpts}</Select>
         )}
       </div>
       {/* assignee 单选仅扁平列表/看板(grouped 用 scope 按类型收窄)。 */}
       {view !== "grouped" && (
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.filter.assignee")}</div>
-          <Select value={f.assignee} onChange={(v) => update({ assignee: v as string | undefined })} showClear filter disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.assignee")}>
+          <Select value={f.assignee} onChange={(v) => update({ assignee: v as string | undefined })} showClear filter disabled={searching} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.assignee")}>
             {cands.map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>))}
           </Select>
         </div>
@@ -306,7 +310,7 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
       {!(view === "grouped" && scope === "involves") && (
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.filter.creator")}</div>
-          <Select value={f.creator} onChange={(v) => update({ creator: v as string | undefined })} showClear filter disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.creator")}>
+          <Select value={f.creator} onChange={(v) => update({ creator: v as string | undefined })} showClear filter disabled={searching} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.creator")}>
             {cands.filter((c) => c.type === "member").map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>))}
           </Select>
         </div>
@@ -315,9 +319,9 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.filter.project")}</div>
           {view === "grouped" ? (
-            <Select multiple value={f.gProjectIds} onChange={(v) => update({ gProjectIds: (v ?? []) as string[] })} showClear filter maxTagCount={1} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
+            <Select multiple value={f.gProjectIds} onChange={(v) => update({ gProjectIds: (v ?? []) as string[] })} showClear filter maxTagCount={1} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
           ) : (
-            <Select value={f.project} onChange={(v) => update({ project: v as string | undefined })} showClear filter disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
+            <Select value={f.project} onChange={(v) => update({ project: v as string | undefined })} showClear filter disabled={searching} {...FIELD_POPUP} style={{ width: "100%" }} placeholder={t("loop.filter.project")}>{projectOpts}</Select>
           )}
         </div>
       )}
@@ -330,10 +334,10 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
       <div className="loop-fields__row">
         <div className="loop-fields__label">{t("loop.filter.dateRange")}</div>
         <div className="loop-fields__inline">
-          <Select value={f.dateField} onChange={(v) => update({ dateField: v as IssueDateField })} disabled={searching} dropdownClassName="loop-fields__dropdown" style={{ width: 104 }}>
+          <Select value={f.dateField} onChange={(v) => update({ dateField: v as IssueDateField })} disabled={searching} {...FIELD_POPUP} style={{ width: 104 }}>
             {ISSUE_DATE_FIELDS.map((d) => (<Select.Option key={d} value={d}>{t(`loop.dateField.${d}`)}</Select.Option>))}
           </Select>
-          <DatePicker type="dateRange" density="compact" disabled={searching} value={f.dateRange} onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })} placeholder={t("loop.filter.dateRange")} style={{ flex: 1 }} />
+          <DatePicker type="dateRange" density="compact" disabled={searching} value={f.dateRange} onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })} placeholder={t("loop.filter.dateRange")} zIndex={FIELD_POPUP.zIndex} style={{ flex: 1 }} />
         </div>
       </div>
       {/* 关键词走全文搜索(独立语义,不与 grouped 组合),故分组视图隐藏。 */}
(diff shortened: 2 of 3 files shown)


mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #739 (octo-web)

**Head SHA:** `21a621600034bdb43a3777cbebfa34fe442dbb9f`
**Merge base:** `5038a6b09764de601e99e3e8ba96871376e19920` (main, at branching)
**Scope:** 3 files, +36 / −20, `packages/dmloop/src/` (loop feature).

## Summary

Two tightly-scoped bug fixes in the loop feature, each contained to a single concern:

1. **Filter / show panel dropdowns hidden behind the panel.** The Semi `<Dropdown>` popup that wraps the filter/show panel carries Semi's baseline `z-index: 1060`, while the inner Semi `<Select>` / `<DatePicker>` popups defaulted to `1030` — so options rendered underneath the panel body and could not be clicked. The fix lifts every inner popup to `zIndex: 2000` via a shared `FIELD_POPUP` const spread into each affected field, and passes `zIndex={FIELD_POPUP.zIndex}` directly to the `<DatePicker>`.
2. **Stale agent-status cache.** After workspace switches and agent mutations (create / update / archive / restore), `invalidateAgentStatus()` was not called, so the shared `id → status` cache powering the @-mention menu showed stale online dots. The fix adds `invalidateAgentStatus()` to every agent-mutation success path via a new `afterAgentMutation` wrapper, and to both workspace-switch code paths via a new `resetWorkspaceCaches()` helper.

Both fixes are narrow, consistent with the existing `invalidateDirectory()` / `invalidateRuntimeMap()` pattern, and do not introduce new runtime contracts.

## Verification

- ✅ **Diff scope matches the two claimed bugs** — 3 files changed, all inside `packages/dmloop/src/`, no unrelated refactor churn.
- ✅ **All four agent mutation paths now invalidate the agent-status cache** — `createAgent` / `updateAgent` / `archiveAgent` / `restoreAgent` (`packages/dmloop/src/api/agentApi.ts:73,77,82,87`) all route through the new `afterAgentMutation` wrapper, which calls `invalidateAgentStatus()` before delegating to `afterDirectoryMutation`. Callers (`AgentPage.tsx:79,86,97`; `AgentDetailPage.tsx:103,126`) pick up the fix with no local change required.
- ✅ **Workspace switch clears the same set of caches in both code paths** — `applyWorkspace` (`packages/dmloop/src/pages/LoopPage.tsx:107`) and `switchWorkspace` (`:172`) both now call the new `resetWorkspaceCaches()` helper (declared at `:29`), eliminating the prior duplication that made it easy to forget `invalidateAgentStatus()` in one of the two paths.
- ✅ **All Selects and the DatePicker inside the filter/show panels receive the shared `FIELD_POPUP`** — 10 `<Select>` instances + 1 `<DatePicker>` in `packages/dmloop/src/pages/IssuePage.tsx:287,289,295,297,304,313,322,324,337,340,372` are all updated consistently. Sibling Selects outside the filter panel (`CreateIssueModal`, `SquadPage`, `AgentPage`, `SkillPage`, `SettingsPage`, `NewLoopPage`, `ScheduleFields`, `CreateAutomationModal`) intentionally keep the bare `dropdownClassName` since they are not nested under a parent Semi `<Dropdown>` with elevated z-index.
- ✅ **Consumer of `agentStatusMap()` re-fetches on each invocation** — `packages/dmloop/src/ui/mentionExtension.ts:91` calls `agentStatusMap()` every time the @-mention menu opens; there is no mount-time latch, so clearing the cache is sufficient to surface fresh status on next open.
- Static analysis only at head `21a6216`; build and tests not executed in this environment.

## Findings

### Nit — Magic-number `zIndex: 2000` duplicated across the codebase (`packages/dmloop/src/pages/IssuePage.tsx:69`)

`FIELD_POPUP.zIndex = 2000` is a bare literal also used in `packages/dmloop/src/ui/composer.css:67` for a separate popup layer. If a future overlay is introduced at ≥2000, both the filter popups and the composer popup collide with it silently, and there is no ladder to consult when picking the next layer. Consider extracting to a shared token (CSS custom property, or a `POPUP_Z_INDEX_*` constants module) so the layering order is maintained in one diff rather than N. Not blocking.

## 数据流回溯

| 消费点 | 上游来源 | 是否真流到消费点 |
|---|---|---|
| `FIELD_POPUP.zIndex` → 每个 `<Select>` 弹层 | `IssuePage.tsx:69` 常量，经 `{...FIELD_POPUP}` 展开 | ✅ 每个 spread 在 JSX 展开时生效，Semi Select 接受 `zIndex` prop 覆盖默认 1030 |
| `FIELD_POPUP.zIndex` → `<DatePicker>` 日历弹层 | `IssuePage.tsx:340` `zIndex={FIELD_POPUP.zIndex}` | ✅ Semi DatePicker 接受 `zIndex` prop |
| `invalidateAgentStatus()` → `_agentStatus=null` & `_agentStatusPromise=null` | `afterAgentMutation` (`agentApi.ts:68`) / `resetWorkspaceCaches` (`LoopPage.tsx:31`) | ✅ 双字段清空，下次 `agentStatusMap()` 必重新 fetch，不会复用 rejected promise |
| `agentStatusMap()` 被 `mentionExtension.ts:91` 消费 | 每次 @ 菜单打开重新调用，无 mount 时 latch | ✅ 缓存清除后下次调用必 fetch，stale 状态不会残留 |
| `resetWorkspaceCaches()` → applyWorkspace / switchWorkspace | `LoopPage.tsx:107` / `:172` | ✅ 两条切换路径都走 helper，无遗漏 |

## 盲点 checklist

- **C1 — 双路径 parity**: ✅ 两条工作区切换路径（`applyWorkspace` 初选 + `switchWorkspace` 切换）一致调用 `resetWorkspaceCaches()`；四条 agent mutation（create / update / archive / restore）一致使用 `afterAgentMutation`。无对称遗漏。
- **C2 — control-flow ordering / 嵌套复用**: N/A — 纯 invalidation，无 ordering 敏感逻辑；`FIELD_POPUP` 展开在 `style` 之前，JSX 展开顺序不会覆盖显式 prop。
- **C3 — 授权边界**: N/A — 无授权/权限改动。
- **C4 — 授权生命周期**: N/A。
- **C5 — build/note 通过 ≠ 运行期路径正确**: N/A — 非构建/打包改动。
- **C6 — 治理/策略文档**: N/A。

## 跨轮 blocker 复检 (R6)

N/A — 本 PR 首次审查。

## Verdict: APPROVED

两个修复均窄且一致于既有 invalidation 模式：agent-status 缓存的四个 mutation 路径和两条 workspace-switch 路径全部对齐；filter 面板内所有 Semi Select / DatePicker 统一抬 z-index，不影响面板外同级 Select。唯一的 nit 是 `2000` 魔法数字在多处重复，建议后续抽出共享 token，不阻塞落地。





[Octo-Q] verdict: APPROVE — 无 P0/P1。唯一的 nit（z-index 魔法数字重复）不阻塞。

yujiawei (approved):
# Code Review — PR #739 (octo-web)

**Head SHA:** `21a621600034bdb43a3777cbebfa34fe442dbb9f` · **Base:** `main` · Frontend-only, `packages/dmloop` (3 files, +36/−20)

## 1. Spec compliance

**Spec: ✅**

Two scoped fixes, both implemented as described, nothing extra:

- **Missing work:** none. Fix #1 covers every filter/show field — all 10 `<Select>` in the toolbar `Dropdown` panels now carry `{...FIELD_POPUP}` (`IssuePage.tsx:287,289,295,297,304,313,322,324,337,372`) and the `DatePicker` gets `zIndex={FIELD_POPUP.zIndex}` (`IssuePage.tsx:340`). No field popup was left behind. Fix #2 invalidates agent-status across the full class of agent-set changes: create/update/archive/restore via `afterAgentMutation` (`agentApi.ts:65-88`) and workspace switch via `resetWorkspaceCaches` (`LoopPage.tsx:27-31`, called at `:110` and `:175`).
- **Over-build:** none. `FIELD_POPUP` is local to `IssuePage`; `afterAgentMutation` only wraps the existing `afterDirectoryMutation` for agent writes; `resetWorkspaceCaches` is a behavior-preserving extraction of the three existing invalidation calls plus the new agent-status one. No new flags, endpoints, or schema.
- **Deviation:** none. The `{...FIELD_POPUP}` spread replaces the prior `dropdownClassName="loop-fields__dropdown"` on each Select — same class, no lost styling — and adds `zIndex`. The `DatePicker` correctly receives only `zIndex` (its popup has no shared class to target).

## 2. Code quality

**Quality: Approved**

The z-index reasoning is sound. Semi writes the winning z-index inline on the `.semi-portal` node, so a prop is the correct lever (a CSS class rule can't win without `!important`, and the DatePicker popup has no shared class). `getPopupContainer` would clip the wide date-range picker inside the narrow panel — the PR description already reasons through this and lands correctly. 2000 sits above the panel (1060) and below global overlays (modals 9999, toasts 10000); its only z-2000 peer is the @-mention menu (`composer.css:66`), which does not co-render with the issue-list filter panel.

The cache extraction is a clean consolidation. `afterAgentMutation` is a good instance of collapsing per-call-site invalidation into one API-layer chokepoint, matching the existing `afterDirectoryMutation` pattern.

### Findings (all advisory — none blocking)

- **[P2] Residual invalidation race on agent-status cache** — `agentApi.ts:45-59`. `agentStatusMap()` has no generation-counter guard, unlike `directory.ts` which uses `_gen` (`directory.ts:22,78-83,90`) precisely to stop a late-resolving `build()` from writing stale data back into `_cache` after an invalidation. If a workspace switch fires while an `agentStatusMap()` fetch from the previous workspace is in flight, the older promise can repopulate `_agentStatus` after `resetWorkspaceCaches()` cleared it. Impact is bounded and not a regression: the lookup is id-keyed (`agentStatus.get(c.id)`, `mentionExtension.ts:137`) with globally-unique agent UUIDs, and the candidate list it renders against comes from the gen-guarded, correctly-scoped directory — so a poisoned map produces at most a *missing/wrong online dot* for an agent, never a wrong agent identity or a cross-tenant data leak. And before this PR, workspace switch never invalidated agent-status at all, so dots were permanently stale; this change strictly narrows the window. Worth a follow-up: mirror the `_gen` guard into `agentStatusMap()` so the two workspace-scoped caches behave identically.

- **[P2] Persistently-mounted status consumers aren't reactively refreshed** — `agentApi.ts:65-70`. `invalidateAgentStatus()` drops the cached map but emits no signal, so any already-mounted view showing agent dots won't repaint until it re-reads the map. For the actual consumer here (`mentionExtension`, which reads once at composer build and joins by id) this is fine — a freshly opened composer gets fresh data. Flagging only so a future always-on status indicator isn't assumed to auto-update from this call.

- **[P2] z-index 2000 is a hardcoded escape from the panel stacking context** — `IssuePage.tsx:67`. Correct and minimal for today's layout. The latent risk is reuse: if `IssuePage`'s panel is ever nested inside another portal/drawer with its own stacking context or `overflow: hidden`, the fixed 2000 could clip or mis-stack. Not actionable now; noted for whoever relocates this panel.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Both fixes are correctly scoped and implemented; all findings are P2/advisory (maintainability + a narrow, non-regressing race), none block merge.

## 4. Suggestions

- Add the `_gen` generation guard to `agentStatusMap()` (copy the `directory.ts` pattern) so both workspace-scoped caches are race-safe and symmetric. Small, isolated follow-up.

## 5. Coverage / blind spots

- **Verified from diff + code:** all 10 field Selects + DatePicker covered; `{...FIELD_POPUP}` spread loses no prior prop; z-index ordering vs panel/overlays; the directory `_gen` guard vs the unguarded agent-status cache; id-keyed dot lookup bounds the race impact; workspace switch previously did zero agent-status invalidation (improvement, not regression).
- **Not verifiable statically (needs a running browser, as the author's manual QA covered):** actual paint order of the Semi portals at runtime; that clicking a field does not close the controlled panel while outside-click still does; DatePicker range picker not clipping at 2000.
- **Independent-review note:** this review is the sole automated leg that completed. A second cross-family advisory leg was attempted but was unavailable due to an upstream gateway tool-injection error (it never ran — treated as absent, not as agreement). The findings above are from a single reviewer plus a long-context advisory pass; the z-index and cache-race conclusions were re-derived against the checked-out head SHA rather than taken on trust.

yujiawei (approved):
# Code Review — PR #739 (octo-web) — re-review at new head

**Head SHA:** `0760994d18bb5b65288e0fd0d9c999f89d9d8e40` (was `21a6216` at my prior approval) · **Merge base:** `2bfa7474` (main).

The branch was rebased onto newer main since my first review. I re-verified against the current head and correct merge base rather than inheriting the prior approval.

## What changed vs my earlier review

- **PR's own diff is byte-identical** to the previously approved version: same 3 files, +36/−20 (`agentApi.ts`, `IssuePage.tsx`, `LoopPage.tsx`). Confirmed by diffing the two `merge-base..head` patches restricted to `packages/dmloop/` — no content delta, pure rebase.
- The rebase pulled unrelated upstream commits (runtime-page, `ConnectionStatus` refactor, `dmpersonal`) into the branch history; **none touch the three PR files**, and the symbols this PR depends on (`afterDirectoryMutation`, Semi `Select`/`DatePicker` `zIndex`) are unchanged at the new base. Diff applies cleanly.

## Verdict carried forward: APPROVED

Both fixes remain correctly scoped and implemented; findings are all P2/advisory, none blocking. Nothing in the rebase changes the analysis.

- All 10 filter/show `<Select>` + the `<DatePicker>` carry `{...FIELD_POPUP}` / `zIndex`; z-index 2000 sits above the panel (1060), below global overlays (modals 9999, toasts 10000).
- Agent-status invalidation covers all four mutation paths (`afterAgentMutation`) and both workspace-switch paths (`resetWorkspaceCaches`).

### Advisory findings (unchanged, non-blocking)
- **[P2]** `agentStatusMap()` (`agentApi.ts`) lacks the `_gen` generation guard that `directory.ts` uses, leaving a narrow late-resolving-fetch race on workspace switch. Bounded to a missing online dot (id-keyed lookup, unique UUIDs) — no wrong-identity or cross-tenant leak, and a strict improvement over base (which never invalidated agent-status on switch). Follow-up: mirror the `_gen` guard.
- **[P2]** `invalidateAgentStatus()` emits no reactive signal — fine for the on-open `mentionExtension` consumer; noted for any future always-on status indicator.
- **[P2]** `zIndex: 2000` is a hardcoded literal duplicated with `composer.css:67`. Suggest extracting a shared layering token so the popup z-order lives in one place. Not blocking.

OctoBoooot (approved):
## Review: #739 — loop filter-panel dropdown z-index + stale agent-status cache

**Verdict:** Approve. Two well-scoped, correctly-rooted fixes in `packages/dmloop`; both verified complete (no missed field, no missed invalidation path). Risk tier: normal (frontend presentational + cache invalidation, no schema/protocol change). CI green (loop tests 25/25, typecheck clean). Closes #738.

### Fix 1 — filter/show panel dropdowns behind the panel (verified complete)
Root cause is correctly diagnosed: Semi portals `Select`/`DatePicker` popups to `document.body` at base z-index 1030, below the Dropdown panel's 1060, so option popups paint behind the panel. The shared `FIELD_POPUP = { dropdownClassName, zIndex: 2000 }` spread lifts them above.
- **Coverage is complete** — every panel field got the fix: 10 `{...FIELD_POPUP}` spreads, **0** remaining literal `dropdownClassName="loop-fields__dropdown"`, and the `DatePicker` carries `zIndex={FIELD_POPUP.zIndex}`. A partial migration would have left some fields still behind the panel; none are.
- **z-index layering is safe** — 2000 sits above the panel (1060) and below global overlays (modals 9999, toasts 10000). The only z-2000 peer is the @-mention menu, which doesn't co-render with the issue-list filter panel.
- Using the prop (not CSS / `getPopupContainer`) is the right call: Semi writes the winning z-index inline on `.semi-portal`, so a class rule can't override without `!important`, and `getPopupContainer` would clip the wide date-range picker inside the narrow panel.

### Fix 2 — stale agent-status cache (verified complete, closes a #619-class gap)
The module-level agent-status map (drives @-mention online dots) was never invalidated. Now invalidated across the full class of agent-set changes:
- **Workspace switch** — both real-switch paths (`applyWorkspace:110`, `switchWorkspace:175`) route through the new `resetWorkspaceCaches()` (a behavior-preserving extraction of the prior directory+runtime invalidation that *adds* `invalidateAgentStatus`). I checked every `setWorkspaceContext` call site in `dmloop`: the only other real-workspace one is `LoopCliAuthorizePage` (out of scope, and benign — it re-establishes the *current* workspace then redirects to the CLI, rendering no directory/agent-dot surface), and `index.tsx` merely re-exports the symbol. No switch path is left un-invalidated.
- **Agent create/update/archive/restore** — all four now go through `afterAgentMutation` (wraps `invalidateAgentStatus` + the existing `afterDirectoryMutation`), so a created/restored agent's dot is fresh without a reload.

This is exactly the invalidation-symmetry that closes the shared-cache-not-reset pattern; the change makes the agent-status map's lifecycle match the runtime/directory maps'.

### Praise
Correct root-cause diagnosis on both (the Semi inline-z-index detail is the kind of thing that's easy to "fix" with dead CSS), the z-index-2000 choice is justified against the actual overlay stack rather than picked arbitrarily, and the `resetWorkspaceCaches` extraction removes duplicated invalidation blocks while *adding* the missing one — a cleanup and a fix in the same move, with no behavior regression to the existing two calls.

### Nit (non-blocking)
`resetWorkspaceCaches` and `afterAgentMutation` are module-local; if other packages carry parallel agent/workspace caches (the loop workspace context is process-global per the `http.ts` module state), a follow-up could consider whether any cross-package cache needs the same invalidation — but nothing in *this* PR's scope leaks.

### Net
Both fixes are complete (byte-verified: full z-index coverage, all workspace-switch + agent-mutation paths invalidated), correctly rooted, and low blast radius. Approving. Credit MrAladdin. `mergeStateStatus` will remain gated on the required `code-review` check until it passes.

OctoBoooot:
## Corroborating yujiawei's cache-race 🟡 — it's real, and it's the #729 latest-wins pattern (still non-blocking)

Byte-verified yujiawei's first 🟡 ("agentStatusMap 缺少生成计数保护"). It's a real residual race in the new invalidation, and it's worth making concrete because it's reachable via workspace switch:

1. An @-mention menu opens in workspace **A** → `agentStatusMap()` starts `listAgents()`; the request is already sent with A's `x-workspace-slug` header.
2. User switches to workspace **B** → `resetWorkspaceCaches()` → `invalidateAgentStatus()` nulls `_agentStatus` and `_agentStatusPromise`.
3. The in-flight A fetch resolves → its `.then` runs `_agentStatus = new Map(agents)` with **workspace-A** data, repopulating the cache *after* the invalidation.
4. In workspace B, `agentStatusMap()` now returns the stale workspace-A statuses until the next mutation/switch.

So the invalidation can be silently undone by a fetch that was in flight across it — the same class of bug that #729 fixed for `resolveWorkspaceSelection` with a generation guard. The fix here is the same shape: capture a sequence before the fetch and re-check it in the `.then` before assigning, discarding the result if `invalidateAgentStatus()` ran in the meantime:

```
let _agentStatusSeq = 0;
export function invalidateAgentStatus() { _agentStatus = null; _agentStatusPromise = null; _agentStatusSeq++; }
// in agentStatusMap: const seq = _agentStatusSeq; ... .then(agents => { if (seq !== _agentStatusSeq) return new Map(); _agentStatus = new Map(...); ... })
```

**Severity: minor, non-blocking — my APPROVE stands.** This PR *strictly improves* the prior behavior (the agent-status cache was never invalidated at all before, i.e. permanently stale after any change), and the residual race is a narrow, UI-only (online dots), self-healing window. It doesn't regress anything and doesn't gate merge. But it is a real bug worth folding into a follow-up (or this revision) — pairing the invalidation with a seq guard closes the window completely.

yujiawei's second 🟡 (invalidate doesn't refresh already-mounted views) is also fair: `invalidateAgentStatus()` clears module state but doesn't notify mounted components, so it only helps consumers that re-read on next use (the @-mention menu on open does; a read-once persistent reader wouldn't). Consumer-dependent, non-blocking.

Credit yujiawei — both are correct, and #1 is the exact latest-wins guard this cache is missing.

PR #740, opened 2026-07-14
feat(summary): add /s/:taskNo standalone deep-link route to summary detail

Ranwanglc:
## Summary
Adds a browser-openable `/s/:taskNo` standalone deep-link route that opens a summary's detail page, mirroring the existing `/d/:docId` standalone doc route. This is the octo-web half that makes the smart-summary notification card's "查看详情" button actually land on the detail page — including surviving the login redirect.

Supports smart-summary #150. Pairs with the smart-summary notify-card PR and the `task_no` detail-lookup PR.

## Problem
The notification card links to `/s/{task_no}?sp={space_id}`, but octo-web had no such route, so a click bounced to login and lost the target (landing on the app root, not the detail).

## Change (mirrors `/d/:docId`)
- **Strict matcher** `isStandaloneSummaryPath` → only a well-formed single-segment `/s/<taskNo>` (regex `^/s/([A-Za-z0-9_-]+)/?$`); `/s`, `/s/`, `/s/a/b` fall through.
- **Login redirect** carries `sp` (space id) through the bounce so an anonymous user returns to the exact `/s/:taskNo?sp=` target after login.
- **Render branch**: logged-in → `SummaryDetailPage`; anonymous → stash target + login, then post-login return (reuses `persist/consumeStandaloneReturn` + `withReturnSid`).
- **task_no is a string**: raw path segment is passed to the initial detail fetch (`getSummaryDetail`/`useSummaryDetail` widened to `number | string`); numeric-only actions (regenerate/cancel/delete) keep using the numeric `detail.task_id` after the response arrives.
- The **shared** standalone return-target validator (`StandaloneDocPage`) now accepts `/s/:taskNo` under the **same** same-origin + single-segment open-redirect guard as `/d/:docId` (no loosening).

## Safety
Open-redirect guard verified: rooted path, no control chars, same-origin via `new URL`, exact single-segment `/d` or `/s` match — `//evil.com`, `/s/../..`, encoded-slash payloads all rejected. Anonymous users never render the detail (no data leak).

## Deployment prerequisites
None (front-end only). The deep-link origin comes from octo-server `external.webLoginURL` (must be https), unrelated to this PR.

## Testing
`npx vitest run` green in all three touched areas: Layout standalone path (9), docs `StandaloneDocPage` (43, incl. new `/s/` return-target cases), full `dmworksummary` suite (376). Trusted vitest over bare tsc (known Vite false-positives). Independently re-verified + adversarially reviewed (2 rounds, APPROVE). No `package.json` / lockfile changes.

### DIFF
--- a/apps/web/src/Layout/index.tsx
+++ b/apps/web/src/Layout/index.tsx
@@ -14,6 +14,7 @@ import InviteLanding from "../Components/InviteLanding";
 import JoinSpacePage from "../Components/JoinSpacePage";
 import JoinApprovalResult from "../Components/JoinApprovalResult";
 import { StandaloneDocPage, parseStandaloneDocId, isStandaloneDocPath, persistStandaloneReturn, consumeStandaloneReturn, withReturnSid } from "@octo/docs";
+import { SummaryDetailPage } from "@dmwork/summary";
 import { adoptStoredSession, findSidForToken, clearSessionsWithToken } from "./recoverSession";
 import { isLoopCliAuthorizePath, LOOP_CLI_AUTHORIZE_PATH } from "@octo/loop";
 
@@ -75,6 +76,28 @@ function clearExpiredStandaloneSessionAndReload(): void {
     if (typeof window !== "undefined") window.location.reload();
 }
 
+/** `/s/:taskNo` — summary taskNo is a single URL path segment, optional trailing slash. */
+const STANDALONE_SUMMARY_PATH = /^\/s\/([A-Za-z0-9_-]+)\/?$/;
+
+export function parseStandaloneSummaryTaskNo(pathname: string): string | null {
+    if (typeof pathname !== "string") return null;
+    const m = STANDALONE_SUMMARY_PATH.exec(pathname);
+    return m ? m[1] : null;
+}
+
+// Only a well-formed single-segment `/s/<taskNo>` counts as standalone; `/s`,
+// `/s/`, and nested `/s/a/b` must fall through (else the render branch mounts
+// SummaryDetailPage with an undefined taskId).
+export function isStandaloneSummaryPath(pathname: string): boolean {
+    return parseStandaloneSummaryTaskNo(pathname) !== null;
+}
+
+function applyStandaloneSummarySpaceFromQuery(): void {
+    const params = new URLSearchParams(window.location.search);
+    const spaceId = params.get("sp") || params.get("space") || "";
+    if (spaceId) WKApp.shared.currentSpaceId = spaceId;
+}
+
 export default class AppLayout extends Component<{}, AppLayoutState> {
     state: AppLayoutState = { showJoinSpace: false };
 
@@ -124,12 +147,16 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             const forwardDoc = getQueryParam("doc") || ""
             const forwardSpace = getQueryParam("space") || ""
             const forwardFolder = getQueryParam("folder") || ""
+            const forwardSp = getQueryParam("sp") || ""
             const redirectQuery = new URLSearchParams()
             if (existingSid) redirectQuery.set("sid", existingSid)
             if (forwardDoc) {
                 redirectQuery.set("doc", forwardDoc)
                 if (forwardSpace) redirectQuery.set("space", forwardSpace)
                 if (forwardFolder) redirectQuery.set("folder", forwardFolder)
+            }
+            if (isStandaloneSummaryPath(window.location.pathname) && forwardSp) {
+                redirectQuery.set("sp", forwardSp)
             }
             const redirectQs = redirectQuery.toString()
             const sidParam = redirectQs ? `?${redirectQs}` : ""
@@ -404,6 +431,25 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             // the login screen (below) without navigating away. The standalone page itself only
             // mounts once a token exists, so the anonymous path is the ONLY place this key gets
             // written for a first-time visitor — onLogin consumes it via consumeStandaloneReturn.
+            persistStandaloneReturn();
+            // Anonymous: fall through to the login screen (below) without navigating away.
+        }
+
+        // Standalone summary deep-link (`/s/:taskNo`): notification cards use task_no (not numeric
+        // task_id), so pass the raw path segment into the detail fetch. Auth/session recovery mirrors
+        // `/d/:docId`; anonymous visitors stash the exact target and return after login.
+        if (isStandaloneSummaryPath(window.location.pathname)) {
+            if (!WKApp.loginInfo.token) {
+                WKApp.loginInfo.load();
+            }
+            if (!WKApp.loginInfo.token) {
+                recoverOctoSessionFromStorage(true);
+            }
+            if (WKApp.loginInfo.token) {
+                applyStandaloneSummarySpaceFromQuery();
+                const standaloneTaskNo = parseStandaloneSummaryTaskNo(window.location.pathname);
+                return <SummaryDetailPage taskId={standaloneTaskNo ?? undefined} />;
+            }
             persistStandaloneReturn();
             // Anonymous: fall through to the login screen (below) without navigating away.
         }
--- a/apps/web/src/__tests__/layoutStandaloneDocPath.test.ts
+++ b/apps/web/src/__tests__/layoutStandaloneDocPath.test.ts
@@ -71,3 +71,54 @@ describe('Layout — standalone /d/:docId clean cold-load path', () => {
     expect(layout).toMatch(/window\.location\.reload\(\)/)
   })
 })
+
+describe('Layout — standalone /s/:taskNo summary clean cold-load path', () => {
+  let layout: string
+
+  beforeAll(() => {
+    layout = fs.readFileSync(path.join(__dirname, '../Layout/index.tsx'), 'utf-8')
+  })
+
+  it('matches only a well-formed single-segment /s/<taskNo>, not /s, /s/, or nested paths', () => {
+    expect(layout).toMatch(/isStandaloneSummaryPath\(\s*window\.location\.pathname\s*\)/)
+    expect(layout).toMatch(/function\s+parseStandaloneSummaryTaskNo/)
+    // The matcher must delegate to the strict parser (single segment) rather than a
+    // loose /s namespace regex, so /s, /s/, and /s/a/b fall through instead of
+    // mounting SummaryDetailPage with an undefined taskId.
+    expect(layout).toMatch(/return\s+parseStandaloneSummaryTaskNo\(pathname\)\s*!==\s*null/)
+    expect(layout).not.toMatch(/STANDALONE_SUMMARY_NAMESPACE/)
+    expect(layout).not.toMatch(/if\s*\(\s*standaloneTaskNo\s*\)/)
+    // The strict single-segment path regex is the sole source of truth.
+    expect(layout).toMatch(/STANDALONE_SUMMARY_PATH\s*=\s*\/\^\\\/s\\\/\(\[A-Za-z0-9_-\]\+\)/)
+  })
+
+  it('recovers a stored session before rendering the summary detail page', () => {
+    const branchIdx = layout.indexOf('Standalone summary deep-link')
+    expect(branchIdx).toBeGreaterThan(0)
+    const nsIdx = layout.indexOf('isStandaloneSummaryPath(window.location.pathname)', branchIdx)
+    expect(nsIdx).toBeGreaterThan(0)
+    const recoverIdx = layout.indexOf('recoverOctoSessionFromStorage(true)', nsIdx)
+    const renderIdx = layout.indexOf('<SummaryDetailPage', nsIdx)
+    expect(recoverIdx).toBeGreaterThan(nsIdx)
+    expect(renderIdx).toBeGreaterThan(recoverIdx)
+  })
+
+  it('renders summary detail only with a token, passing the raw task_no string through', () => {
+    expect(layout).toMatch(/if\s*\(\s*WKApp\.loginInfo\.token\s*\)\s*\{[\s\S]*?<SummaryDetailPage/)
+    expect(layout).toMatch(/const\s+standaloneTaskNo\s*=\s*parseStandaloneSummaryTaskNo\(window\.location\.pathname\)/)
+    expect(layout).toMatch(/<SummaryDetailPage\s+taskId=\{standaloneTaskNo\s*\?\?\s*undefined\}/)
+  })
+
+  it('stashes anonymous /s targets and carries sp through post-login return', () => {
+    const branchIdx = layout.indexOf('Standalone summary deep-link')
+    expect(branchIdx).toBeGreaterThan(0)
+    const nsIdx = layout.indexOf('isStandaloneSummaryPath(window.location.pathname)', branchIdx)
+    expect(nsIdx).toBeGreaterThan(0)
+    const stashIdx = layout.indexOf('persistStandaloneReturn()', nsIdx)
+    expect(stashIdx).toBeGreaterThan(nsIdx)
+    expect(layout).toMatch(/const\s+forwardSp\s*=\s*getQueryParam\("sp"\)\s*\|\|\s*""/)
+    expect(layout).toMatch(/redirectQuery\.set\("sp",\s*forwardSp\)/)
+    expect(layout).toMatch(/consumeStandaloneReturn\(\)/)
+    expect(layout).toMatch(/withReturnSid\(standaloneReturn,\s*sessionSid\)/)
+  })
+})
--- a/packages/dmworkbase/src/App.tsx
+++ b/packages/dmworkbase/src/App.tsx
@@ -703,7 +703,7 @@ export default class WKApp extends ProviderListener {
   static menus = MenusManager.shared; // 菜单
   // Callback to switch the active sidebar menu by id (set by Main page)
   static switchToMenuById?: (menuId: string) => void;
-  static openSummaryDetail?: (taskId: number) => void;
+  static openSummaryDetail?: (taskId: number | string) => void;
   static searchChatCandidates?: (params: { keyword?: string; chat_type?: string; space_id?: string }) => Promise<any[]>;
   // Id of the currently active sidebar menu (kept in sync by Main page)
   static currentMenuId?: string;
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/actions.test.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/actions.test.ts
@@ -1,0 +1,65 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { openSummaryDetailMock, wkAppMock } = vi.hoisted(() => {
+  const openSummaryDetailMock = vi.fn();
+  return {
+    openSummaryDetailMock,
+    wkAppMock: {
+      openSummaryDetail: openSummaryDetailMock as ((taskId: number | string) => void) | undefined,
+    },
+  };
+});
+
+vi.mock("../../../App", () => ({
+  default: wkAppMock,
+}));
+
+import { openUrl } from "../renderer/actions";
+
+describe("openUrl", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+    openSummaryDetailMock.mockReset();
+    wkAppMock.openSummaryDetail = openSummaryDetailMock;
+    vi.spyOn(window, "open").mockImplementation(() => null);
+  });
+
+  it("routes /s/<taskNo> summary links internally", () => {
+    openUrl("https://web.example.com/s/ST202607145kstyh08?sp=space-1");
+
+    expect(openSummaryDetailMock).toHaveBeenCalledTimes(1);
+    expect(openSummaryDetailMock).toHaveBeenCalledWith("ST202607145kstyh08");
+    expect(window.open).not.toHaveBeenCalled();
+  });
+
+  it("opens external https links in a new tab", () => {
+    openUrl("https://web.example.com/docs/ST202607145kstyh08");
+
+    expect(openSummaryDetailMock).not.toHaveBeenCalled();
+    expect(window.open).toHaveBeenCalledWith(
+      "https://web.example.com/docs/ST202607145kstyh08",
+      "_blank",
+      "noopener,noreferrer"
+    );
+  });
+
+  it("ignores unsafe urls", () => {
+    openUrl("javascript:alert(1)");
+
+    expect(openSummaryDetailMock).not.toHaveBeenCalled();
+    expect(window.open).not.toHaveBeenCalled();
+  });
+
+  it("falls back to window.open when summary routing is unavailable", () => {
+    wkAppMock.openSummaryDetail = undefined;
+
+    openUrl("https://web.example.com/s/ST202607145kstyh08?sp=space-1");
+
+    expect(openSummaryDetailMock).not.toHaveBeenCalled();
+    expect(window.open).toHaveBeenCalledWith(
+      "https://web.example.com/s/ST202607145kstyh08?sp=space-1",
+      "_blank",
+      "noopener,noreferrer"
+    );
+  });
+});
--- a/packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts
@@ -1,4 +1,7 @@
+import WKApp from "../../../App";
 import { isSafeUrl } from "../../../Utils/security";
+
+const SUMMARY_DETAIL_PATH_RE = /^\/s\/([A-Za-z0-9_-]+)\/?$/;
 
 /**
  * Action.OpenUrl 导航。
@@ -11,6 +14,18 @@ import { isSafeUrl } from "../../../Utils/security";
 /** 在新标签打开 URL；提交前二次 isSafeUrl 校验（http/https），非法直接忽略。 */
 export function openUrl(url: string): void {
   if (!isSafeUrl(url)) return;
+
+  const parsed = new URL(url);
+  const summaryMatch = parsed.pathname.match(SUMMARY_DETAIL_PATH_RE);
+  if (summaryMatch) {
+    const taskNo = summaryMatch[1];
+    if (WKApp.openSummaryDetail) {
+      // 卡片深链可能来自 https 生产域，本地调试是 http/端口；/s/<taskNo> 路径本身是内部详情信号。
+      WKApp.openSummaryDetail(taskNo);
+      return;
+    }
+  }
+
   window.open(url, "_blank", "noopener,noreferrer");
 }
 
--- a/packages/dmworksummary/src/api/summaryApi.ts
+++ b/packages/dmworksummary/src/api/summaryApi.ts
@@ -133,7 +133,7 @@ export async function listSummaries(
     return get('/summaries', params as Record<string, unknown>, config);
 }
 
-export async function getSummaryDetail(taskId: number): Promise<SummaryDetail> {
+export async function getSummaryDetail(taskId: number | string): Promise<SummaryDetail> {
     return get(`/summaries/${taskId}`);
 }
 
--- a/packages/dmworksummary/src/hooks/useSummaryDetail.ts
+++ b/packages/dmworksummary/src/hooks/useSummaryDetail.ts
@@ -12,7 +12,7 @@ interface UseSummaryDetailReturn {
     cancel: () => Promise<void>;
 }
 
-export function useSummaryDetail(taskId: number | null): UseSummaryDetailReturn {
+export function useSummaryDetail(taskId: number | string | null): UseSummaryDetailReturn {
     const [detail, setDetail] = useState<SummaryDetail | null>(null);
     const [loading, setLoading] = useState(false);
     const [error, setError] = useState<string | null>(null);
@@ -41,13 +41,13 @@ export function useSummaryDetail(taskId: number | null): UseSummaryDetailReturn 
     }, []);
 
     const regenerate = useCallback(async () => {
-        if (taskId == null) return;
+        if (typeof taskId !== "number") return;
         await api.regenerateSummary(taskId);
         refresh();
     }, [taskId, refresh]);
 
     const cancel = useCallback(async () => {
-        if (taskId == null) return;
+        if (typeof taskId !== "number") return;
         await api.cancelSummary(taskId);
         refresh();
     }, [taskId, refresh]);
--- a/packages/dmworksummary/src/index.tsx
+++ b/packages/dmworksummary/src/index.tsx
@@ -1,1 +1,2 @@
 export { SummaryModule } from "./module";
+export { default as SummaryDetailPage } from "./pages/SummaryDetailPage";
--- a/packages/dmworksummary/src/module.tsx
+++ b/packages/dmworksummary/src/module.tsx
@@ -55,7 +55,7 @@ export class SummaryModule implements IModule {
             "en-US": enUS,
         });
 
-        WKApp.openSummaryDetail = (taskId: number) => {
+        WKApp.openSummaryDetail = (taskId: number | string) => {
             WKApp.switchToMenuById?.("summary");
             WKApp.routeLeft.popToRoot();
             WKApp.routeRight.replaceToRoot(
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -47,7 +47,7 @@ import MemberSelectorModal from "../components/MemberSelectorModal";
 import type { MemberCandidate } from "../types/summary";
 
 interface SummaryDetailPageProps {
-    taskId?: number;
+    taskId?: number | string;
 }
 
 interface SummaryDetailPageState {
@@ -192,7 +192,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
 
     componentDidUpdate(prevProps: any) {
         const prevTaskId = prevProps.taskId;
-        const currentTaskId = this.taskId;
+        const currentTaskId = this.detailLookupId;
         if (prevTaskId !== currentTaskId && currentTaskId != null) {
             this.listPageActive = false;
             this.clearAllTimers();
@@ -313,15 +313,20 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     }
 
     get taskId(): number | null {
+        return typeof this.props.taskId === "number" ? this.props.taskId : this.state.detail?.task_id ?? null;
+    }
+
+    private get detailLookupId(): number | string | null {
         return this.props.taskId ?? null;
     }
 
     async loadDetail() {
-        if (this.taskId == null) return;
+        const lookupId = this.detailLookupId;
+        if (lookupId == null) return;
         // Blocking 5：每轮 detail 加载开始就 bump 序列号。这样旧 task 未完成的
         // loadSchedule（包括本函数下面发起的）都会被后续轮作废，不会回填到新 task。
         const seq = this.nextScheduleSeq();
-        const requestTaskId = this.taskId;
+        const requestTaskId = lookupId;
         // F1：切 task / 重拉 detail 时复位全部编辑态，避免旧 task 编辑态（尤其
         // editingTeamSummary）被带入新 task——否则切到非 creator 新 task 会绕过权限进编辑器。
         // FE-1（切任务竞态）：开始新 task 加载时同步清空上一 task 的 personalResult/members，
@@ -352,9 +357,9 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             workflowRevealDone: false,
         });
         try {
-            const detail = await api.getSummaryDetail(this.taskId);
+            const detail = await api.getSummaryDetail(lookupId);
             // detail 本身也可能是旧请求：期间切了 task / 又发了一轮 loadDetail 就丢弃。
-            if (this.scheduleLoadSeq !== seq || this.taskId !== requestTaskId) return;
+            if (this.scheduleLoadSeq !== seq || this.detailLookupId !== requestTaskId) return;
             this.setState({
                 detail,
                 loading: false,
@@ -385,13 +390,13 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             // Load BY_PERSON data
             if (detail.summary_mode === SummaryMode.BY_PERSON) {
                 // FE-1：把本轮 seq 传给二次异步加载，迟到响应按 seq/taskId 校验后才 setState。
-                this.loadPersonalResult(seq);
-                this.loadMembers(seq);
+                this.loadPersonalResult(seq, detail.task_id);
+                this.loadMembers(seq, detail.task_id);
             }
         } catch (err: any) {
             // FE-1（切 task 竞态）：迟到的失败响应也要校验归属，否则旧 task 的加载失败
             // 会把错误/loading 状态写到已切换的新 task 上。
-            if (this.scheduleLoadSeq !== seq || this.taskId !== requestTaskId) return;
+            if (this.scheduleLoadSeq !== seq || this.detailLookupId !== requestTaskId) return;
             this.setState({ error: err.message || t("summary.common.loadingFailed"), loading: false });
         }
     }
@@ -427,13 +432,13 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
      * 仍一致才能 setState。过期响应（期间切了 task / 又发了一轮加载）一律
      * 忽略（return），绝不把旧任务的 personalResult 写到新任务界面（泄漏他人报告）。
      */
-    async loadPersonalResult(seq?: number) {
-        if (this.taskId == null) return;
+    async loadPersonalResult(seq?: number, taskIdOverride?: number) {
+        const requestTaskId = taskIdOverride ?? this.taskId;
+        if (requestTaskId == null) return;
         const reqSeq = seq ?? this.nextScheduleSeq();
-        const requestTaskId = this.taskId;
         this.setState({ personalLoading: true });
         try {
-            const result = await api.getPersonalResult(this.taskId);
+            const result = await api.getPersonalResult(requestTaskId);
             // 迟到响应（期间切 task / 重新加载）：丢弃，不污染新 task。
             if (this.scheduleLoadSeq !== reqSeq || this.taskId !== requestTaskId) return;
             const shouldGateWorkflow = this.shouldGateWorkflowForPersonalResult(result);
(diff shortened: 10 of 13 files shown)


OctoBoooot (dismissed):
## Review: #740 — `/s/:taskNo` standalone deep-link route to summary detail

**Verdict:** Approve. A careful, security-conscious mirror of the existing `/d/:docId` standalone route; every safety claim in the description holds up against the code. Risk tier: normal (frontend-only routing + a widened type; no schema/backend change). CI: all code checks green (Build, secret-scan, osv-scan, dependency-review, pr-title-lint); `code-review` (human gate) pending. Supports smart-summary #150.

### Security — verified, not taken on faith
- **Open-redirect guard extended, not loosened** (`StandaloneDocPage.tsx:isSafeReturnPath`). All three original gates are intact — rooted path, same-origin via `new URL(...).origin === origin`, single-segment match — and the change is a single OR: `parseStandaloneDocId(...) !== null || STANDALONE_SUMMARY_PATH.test(url.pathname)`, with `/s` using the identical `[A-Za-z0-9_-]` charset as `/d`. `//evil.com`, `/s/../..`, backslash-smuggled, and encoded-slash targets are still rejected structurally.
- **Anonymous never renders detail** (`Layout/index.tsx`). The `<SummaryDetailPage>` return is inside `if (WKApp.loginInfo.token)`; the anonymous path only calls `persistStandaloneReturn()` and falls through to login. No data leak.
- **Strict matcher** — `^\/s\/([A-Za-z0-9_-]+)\/?$` rejects `/s`, `/s/`, `/s/a/b`, and any `.`/`%`-bearing segment, so the render branch can't mount with an undefined/nested taskId.
- **`sp` space param is a hint, not authz** — `applyStandaloneSummarySpaceFromQuery()` sets `WKApp.shared.currentSpaceId` from the (attacker-influenceable) `sp`/`space` query, structurally the same trust model as the accepted `/d` branch (which uses the cached space id). Authorization is backend-enforced on `GET /summaries/:taskNo`; a tampered `sp` yields at most a wrong space header, never access to a summary the user can't see — and the anonymous path renders nothing.

### Correctness — the `number | string` widening is handled defensively
The raw `task_no` string is confined to the *fetch* path and never reaches a numeric action:
- `taskId` getter now returns numeric only — `typeof props.taskId === "number" ? props.taskId : state.detail?.task_id ?? null` — so regenerate/cancel/delete key off the numeric `task_id` from the fetched response, not the URL string.
- `detailLookupId` (the raw prop) is used only for `getSummaryDetail(lookupId)` and the race-guard comparisons.
- `useSummaryDetail`'s `regenerate`/`cancel` guards were tightened from `taskId == null` to `typeof taskId !== "number"`, correctly refusing to POST a string task_no to the numeric mutation endpoints.
- The latest-wins race guards (`scheduleLoadSeq` / `requestTaskId`, the same pattern hardened in #698/#729) are preserved and now compare `detailLookupId`; `componentDidUpdate` also switched to `detailLookupId` so a prop change re-triggers the load. No regression to the existing race protection.

### Praise
The `taskId` (numeric, for actions) vs `detailLookupId` (raw, for fetch) split is exactly the right way to widen the type without letting a string leak into numeric endpoints — and preserving the seq/requestTaskId guards through the refactor (rather than quietly dropping them) shows the author understood the existing race protection. Extending the open-redirect guard by ORing one same-charset matcher, while leaving the three structural gates untouched, is the minimal safe change.

### Nit (non-blocking)
`STANDALONE_SUMMARY_PATH` is now defined in two places (`Layout/index.tsx` and `StandaloneDocPage.tsx`) with identical regex. Since the docs package already exports the `/d` matcher, a future cleanup could export a single shared `/s` matcher too, so the two can't drift.

### Net
Correct, minimal, and safe: open-redirect guard extended without loosening, anonymous-no-leak preserved, string task_no isolated to the fetch with numeric actions guarded, and the race protection carried through intact. Approving. Credit Ranwanglc. `mergeStateStatus` stays gated on the required `code-review` check.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. One P2 (latent type trap in unused `useSummaryDetail` hook) is non-blocking. Data flow, auth, open-redirect guard, and staleness guards all verified correct.

yujiawei (dismissed):
# Code Review — PR #740 (octo-web)

**Verdict: APPROVE.** No P0/P1 blocking issues. The change is a tight, well-scoped mirror of the existing `/d/:docId` standalone route, with a matching open-redirect guard and full test coverage. Two minor maintainability notes below (non-blocking).

## Spec compliance
The diff implements exactly what the description states, with no scope creep:
- Strict single-segment matcher `isStandaloneSummaryPath` / `parseStandaloneSummaryTaskNo` (`^/s/([A-Za-z0-9_-]+)/?$`); `/s`, `/s/`, `/s/a/b` fall through — verified by test.
- Login redirect forwards `sp` through the bounce (`Layout/index.tsx:158-160`).
- Render branch: authenticated → `SummaryDetailPage`; anonymous → `persistStandaloneReturn()` then fall through to login (`Layout/index.tsx:439-454`).
- `task_no` passed as a raw string into the initial fetch; the shared return-target validator accepts `/s/:taskNo` under the **same** same-origin + single-segment guard as `/d/:docId` — no loosening (`StandaloneDocPage.tsx:108`).
- No `package.json` / lockfile changes (confirmed in diff).

## Correctness (verified)
- **`taskId` getter split is sound.** `get taskId()` now returns the numeric prop when it is a number, else falls back to `state.detail?.task_id`. All mutating actions (regenerate / cancel / respond / members) route through `this.taskId`, so once the detail loads they operate on the numeric `task_id` even when the page was entered via a string `task_no`. Before load, `this.taskId` is `null` and those actions guard correctly — no request fires with a string against a numeric-only endpoint.
- **`componentDidUpdate` correctly switches to `detailLookupId`.** Comparing `prevProps.taskId` against `this.detailLookupId` (the raw prop) is the right choice: comparing against the new `this.taskId` would compare a raw prop to a resolved numeric `task_id` and could spuriously re-trigger a reload. The late-response guards in `loadDetail` are self-consistent (`requestTaskId = lookupId`, checked against `this.detailLookupId`).
- **Space handling divergence from `/d/` is justified, not a bug.** `summaryApi` attaches `X-Space-Id` from `WKApp.shared.currentSpaceId` on every request, so the `/s/` branch must apply `sp` before the detail fetch or the lookup would run against the wrong/empty space. The docs API does not have the same dependency, so `/d/` legitimately omits it.

## Security
- The open-redirect guard is preserved. `/s/:taskNo` is admitted only after the existing control-char rejection + `url.origin === origin` check, and the summary regex is single-segment (`[A-Za-z0-9_-]+`), so `/s/../..`, encoded-slash, and off-origin payloads are all still rejected. `withReturnSid` percent-encodes the appended `sid` and re-validates. Covered by new tests (`/s/TN-9_x?sp=space-1`, `withReturnSid('/s/...?sp=space-1', ...)`).
- Anonymous users never render the detail page (token gate), so there is no data-leak path.

## Tests
Ran the three touched suites locally against the PR head — all green:
- `apps/web` Layout standalone path: 9/9
- `packages/docs` StandaloneDocPage (incl. new `/s/` return-target cases): 43/43
- `packages/dmworksummary` full suite: 376/376

## Non-blocking notes
- **`useSummaryDetail` (`hooks/useSummaryDetail.ts:44,50`)** — the `regenerate`/`cancel` guards were changed to `if (typeof taskId !== "number") return`, which means those callbacks silently no-op when the hook is initialized with a string. This hook is currently unused (the `/s/` route renders the class `SummaryDetailPage`, not this hook), so there is no live impact today. If a future caller ever passes a `task_no` string, consider resolving the numeric `detail?.task_id` before regenerate/cancel rather than silently aborting. (P2, maintainability.)
- **Render-phase side-effects** — `applyStandaloneSummarySpaceFromQuery()` and `persistStandaloneReturn()` run inside `render()`. This matches the pre-existing `/d/:docId` and CLI-authorize branches, so it is consistent with the current architecture and not a regression introduced here; flagging only as a longer-term refactor candidate (move deep-link auth/session bootstrapping out of render). (P2, consistency.)

## Coverage / what could not be verified from this diff
- Backend acceptance of a string `task_no` on `GET /summaries/{taskNo}` is a cross-repo dependency (the paired detail-lookup change) and out of scope for this front-end PR.
- SPA serving of `/s/*` relies on the nginx `try_files` fallback (same deployment assumption as `/d/*`), out of scope here.
- The notification card's actual `/s/{task_no}?sp={space_id}` link shape lives in the smart-summary repo and was not exercised end-to-end.

Jerry-Xin (dismissed):
**APPROVE** — clean standalone deep-link (`/s/:taskNo`) that reuses the existing authenticated fetch path. No IDOR: authorization stays server-side, the route just navigates. CI green (incl. pr-title-lint). +130/-17, 8 files.

## Authorization / IDOR (the crux) — safe ✅
The `/s/:taskNo` branch in `Layout/index.tsx` does **not** add any new unauthenticated read path. It requires a session token before rendering: if `WKApp.loginInfo.token` is absent it attempts `loginInfo.load()` then `recoverOctoSessionFromStorage(true)`, and if still anonymous it stashes the return target and falls through to the login screen. Only an authenticated caller ever mounts `SummaryDetailPage`.

The detail fetch (`getSummaryDetail` → `GET /summary/api/v1/summaries/{taskNo}`) goes through the shared `summaryAxios` interceptor, which attaches the caller's `token` header and `X-Space-Id`, and a `401` response triggers logout. So the deep-link hits the **same authenticated endpoint** as normal in-app navigation; per-resource permission remains a server-side check on the caller's identity vs. the summary. The client passes `taskNo` as an opaque id — it does not itself grant access. There is no client-side authz that the deep-link bypasses. (Server-side ownership/group enforcement on `GET /summaries/{id}` is out of this repo's diff but is the same backend the existing flow already relies on.)

## taskId vs. detailLookupId split — good defensive design ✅
The prop type widens to `number | string` (string carries the notification `task_no`). The code cleanly separates two concerns:
- `detailLookupId` (raw `number | string`) is used **only** for the read GET in `loadDetail` and the load-race guard.
- The `taskId` getter now returns a **numeric** id only — `props.taskId` when it's a number, else `state.detail?.task_id` (the id the server returned). **Every mutating/permission-sensitive op** (regenerate, cancel, removeMember, leaveSummary, submitPersonalResult, respondToTask, edit) routes through this numeric getter and is guarded by `if (this.taskId == null) return`. So a string deep-link cannot trigger a mutation until the server has returned a detail (and thus confirmed access), and mutations still target the canonical numeric id. The added `typeof taskId !== "number"` guards in `useSummaryDetail` are consistent with this.

## Route-param validation ✅
`STANDALONE_SUMMARY_PATH = /^\/s\/([A-Za-z0-9_-]+)\/?$/` restricts the segment to `[A-Za-z0-9_-]` (single segment, optional trailing slash). `/s`, `/s/`, and nested `/s/a/b` fall through instead of mounting the page with an undefined id — asserted by tests. The value is only ever interpolated into a REST path segment and sent as an axios path; the charset excludes slash/dot-dot/encoded traversal, so there's no path-traversal or injection surface into the query.

## Not-found / no-regression ✅
Invalid or unauthorized `taskNo` → backend error → caught in `loadDetail`, rendered as a warning banner (`summary.common.loadingFailed`); content renders only when `detail && !loading`. No crash, no leak. The new branch is gated behind `isStandaloneSummaryPath(...)` and mirrors the existing `/d/:docId` branch, so it doesn't shadow or reorder existing routes.

## Open-redirect safety (post-login return) ✅
`isSafeReturnPath` now also accepts `/s/:taskNo` in addition to `/d/:docId`, but keeps the three-gate structure (parseable, same-origin via `url.origin === origin`, and standalone-target-only). Absolute / scheme-relative / backslash-smuggled targets are still rejected structurally, so widening the allow-list to the summary target does not weaken the XIN-392 guard. `withReturnSid` correctly appends `sid` on the `/s/...?sp=...` case (tested).

## i18n
No new user-facing strings introduced by this PR; reuses existing summary keys (`summary.common.loadingFailed`, `summary.detail.errorCause`). Nothing to check.

## Non-blocking notes 💬
- The `sp`/`space` query → `WKApp.shared.currentSpaceId` assignment (`applyStandaloneSummarySpaceFromQuery`) lets the deep-link set the active space from the URL. This is only a client-side hint that is echoed back as `X-Space-Id`; since the server still authorizes the token against the resource, it doesn't create an access-control gap. Worth a brief comment that the space header is advisory, not trusted.

Nice test coverage on the path matcher, cold-load recovery ordering, and the open-redirect return target.

OctoBoooot (dismissed):
## Review: #740 @ `a5ba82e1` — DELTA re-review; adds card in-app jump, security paths unchanged, re-affirming APPROVE

**Verdict:** Approve. This commit adds a "card in-app jump" (an `Action.OpenUrl` that recognizes `/s/<taskNo>` links and navigates in-app instead of opening a new tab) on top of the already-approved deep-link route. The security-critical paths from my prior review are byte-identical, and the new code is correctly guarded and tested. My prior APPROVE (@`a4267dac`) was auto-dismissed by this commit; re-anchoring at the live head. CI: all code checks green (Build SUCCESS, secret-scan, osv-scan, pr-title-lint); `code-review` (human gate) pending.

## Parity checked
Delta over my prior head (`a4267dac`→`a5ba82e1`, `ahead 1/behind 1` — one new commit + a one-commit base advance):
- **All five security-critical files I reviewed are byte-identical (SAME blob):** `Layout/index.tsx` (the `/s/:taskNo` render branch), `StandaloneDocPage.tsx` (open-redirect guard), `useSummaryDetail.ts` + `SummaryDetailPage.tsx` (string-taskNo isolation / numeric-action guards), `summaryApi.ts`. So the open-redirect guard, anonymous-no-leak, and string→fetch / numeric→`detail.task_id` separation are all unchanged and still hold. No capability dropped.
- **Genuinely new:** `InteractiveCard/renderer/actions.ts` (+15) — the card jump — plus its test, `openSummaryDetail` widening, and the `SummaryDetailPage` export.

## New — card in-app jump (`InteractiveCard/renderer/actions.ts`)
`openUrl` now intercepts `/s/<taskNo>` links to route in-app. Verified safe:
- **Guarded before parse:** `if (!isSafeUrl(url)) return` runs first, and `isSafeUrl` enforces `protocol ∈ {http,https}` (rejects `javascript:`/`data:`), so `new URL(url)` can't throw on a bad scheme and only http/https reach the match.
- **Same single-segment regex** as the Layout route (`^/s/([A-Za-z0-9_-]+)/?$`) — no traversal/injection via the path.
- On match → `WKApp.openSummaryDetail(taskNo)` (in-app nav to the authenticated summary-detail page); otherwise falls through to the existing `window.open(url, "_blank", "noopener,noreferrer")` — noopener preserved.
- **`openSummaryDetail` widened to `number | string`** consistently across declaration (`App.tsx`) and impl (`module.tsx`); the string `taskNo` flows to the same detail page that isolates it to the fetch and keys numeric actions off `detail.task_id` — matching the original safe design.
- **No IDOR:** the in-app route hits the same authenticated endpoint with server-side per-resource checks; a cross-origin card can't open a summary the user can't access (authz is server-enforced on fetch, not client-trusted). The intercept keying on taskNo regardless of the URL's origin is intended (the taskNo is the internal signal) and benign.
- **Tested:** `actions.test.ts` drives `openUrl("https://…/s/ST…?sp=…")` and asserts `openSummaryDetail` is called (with `window.open` mocked).

## Net
Security-critical deep-link paths unchanged (byte-verified), the new card jump is guarded (isSafeUrl-before-parse, same-charset regex, authenticated in-app nav, noopener fallback), consistently typed, and tested; Build green. Re-affirming APPROVE at `a5ba82e1`. Prior reviewers' approvals were auto-dismissed by this commit and should re-anchor. Credit Ranwanglc.

Jerry-Xin (dismissed):
Re-review at the new head — carries my earlier APPROVE. **APPROVE.**

The new commit adds only part 2 ("card in-app jump"); the deep-link route/fetch/auth I validated earlier is untouched.

### Isolated fix delta (true delta vs. earlier head, merge-base drift excluded)
Only 4 files actually changed in the new commit:
- `packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts` — `openUrl` now intercepts a `/s/<taskNo>` pathname and routes in-app via `WKApp.openSummaryDetail(taskNo)` instead of `window.open`.
- `packages/dmworkbase/src/Messages/InteractiveCard/__tests__/actions.test.ts` — new: internal-route / external-link / unsafe-url / fallback cases.
- `packages/dmworkbase/src/App.tsx` — `openSummaryDetail?` signature widened `number` → `number | string`.
- `packages/dmworksummary/src/module.tsx` — `openSummaryDetail` impl signature widened `number` → `number | string`.

The five part-1 files (`Layout/index.tsx`, `summaryApi.ts`, `useSummaryDetail.ts`, `SummaryDetailPage.tsx`, `dmworksummary/index.tsx`) are byte-identical to the previously-approved head.

### Security re-confirm (blob-verified at the new head)
- **Token-gated render intact** — the route + anonymous stash+login-redirect logic is byte-unchanged from the approved head; not weakened.
- **Authenticated fetch intact** — detail still loads via the same authenticated `GET /summary/api/v1/summaries/{id}` with the shared interceptor (token + `X-Space-Id`, `401 → logout`). The new in-app path just renders the same `SummaryDetailPage`; no new unauthenticated fetch path.
- **Numeric-taskId writes intact** — write actions still use the server-returned numeric `detail.task_id`; the widened `string` only reaches the initial detail fetch.
- **No-traversal regex** — `SUMMARY_DETAIL_PATH_RE = /^\/s\/([A-Za-z0-9_-]+)\/?$/` is single-segment; consistent with the existing standalone matcher.
- **Open-redirect** — this change *reduces* external navigation (intercepts `/s/<taskNo>` before `window.open`). `isSafeUrl` remains the first gate (rejects non-http/https, incl. `javascript:`), and the `window.open` fallback keeps `noopener,noreferrer`. No allow-list widening in this delta.
- **No new IDOR** — a crafted `taskNo` only reaches the same authenticated detail page; permission is still enforced server-side per resource.

### Notes (non-blocking)
- `actions.ts` statically imports the `WKApp` class and reads `WKApp.openSummaryDetail` at call time behind a null check, with graceful `window.open` fallback — no stale closure, covered by the new fallback test.

### Full scan
No new listener/timer without cleanup, no race, no stale-closure, no secret/token logged or placed in a URL/storage. No i18n strings added or removed in this delta.

### Highlights
Good separation of concerns and a focused test matrix (internal route / external link / unsafe url / unavailable-summary fallback).

mochashanyao (dismissed):
[Octo-Q · automated review]

---

## Octo-Q automated review Review — PR#740: feat(summary): add /s/:taskNo standalone deep-link route to summary detail
**Head SHA:** `a5ba82e14d1a57309cdaa64968266ed55161beb5`
**Repo:** Mininglamp-OSS/octo-web
**Scope:** 39 files, +1781 / −334, 4 commits above main-remote base

---

### 1. Verification Summary

| Area | Status | Evidence |
|------|--------|----------|
| `/s/:taskNo` route gating | ✅ | `Layout/index.tsx`: strict regex `STANDALONE_SUMMARY_PATH = /^\/s\/([A-Za-z0-9_-]+)\/?$/`; `isStandaloneSummaryPath` delegates to `parseStandaloneSummaryTaskNo !== null` |
| Session recovery before render | ✅ | `recoverOctoSessionFromStorage(true)` called before `<SummaryDetailPage>` mount |
| Anonymous stash + post-login return | ✅ | `persistStandaloneReturn()` + `forwardSp` + `consumeStandaloneReturn()` + `withReturnSid` |
| `isSafeReturnPath` open-redirect guard | ✅ | `StandaloneDocPage.tsx`: now accepts both `parseStandaloneDocId` and `STANDALONE_SUMMARY_PATH.test()` |
| `taskId` type widening (number → number\|string) | ✅ | `detailLookupId` used only for initial lookup; `this.taskId` getter returns numeric `state.detail.task_id` after first load; all mutation APIs use `this.taskId` |
| InteractiveCard `/s/` URL interception | ✅ | `actions.ts`: regex match → `WKApp.openSummaryDetail(taskNo)` with safe `window.open` fallback when undefined |
| im-runtime helper extraction | ✅ | Pure refactor: `connectClient`, `connectStatus`, `connectAddress`, `clientMsgDevice` — registration-based API, all tested |
| Workspace slug generation | ✅ | Pinyin + WebCrypto 4-char suffix + 409 retry (3 attempts) + "ws" fallback for non-ASCII |
| tiptap composer + mention extension | ✅ | Stable extension instances, debounced search, pinyin matching, DOM popup with proper cleanup |
| Tests | ✅ | New tests: layout standalone summary paths, slug normalization/truncation, im-runtime helpers (4 files), actions.test.ts (openUrl routing), StandaloneDocPage safe return |

---

### 2. Findings

#### P2 — `agentStatusMap` cache has no TTL (stale online dots mid-session)

**File:** `packages/dmloop/src/api/agentApi.ts:52-71`
**Diff-scope:** New code (cache layer added this PR)

The shared agent status cache (`_agentStatus` / `_agentStatusPromise`) is populated once per session and only cleared by explicit `invalidateAgentStatus()`. If an agent transitions online↔offline during a session, the @mention menu shows stale dot colors until the caller invalidates.

**Data-flow trace:** `agentStatusMap()` → `mentionExtension.buildLoopMention()` `items[].dotColor` → `suggestionMenu` render. No periodic refresh path exists.

**Recommendation:** Add a soft TTL (e.g. 60s) or invalidate on known status-change events (agent archive/restore). Acceptable for v1 as-is.

---

#### P2 — Theoretical stale closure in file-only submit path

**File:** `packages/dmloop/src/panel/IssueDetailPage.tsx:~935-945`
**Diff-scope:** New code (addPendingFiles introduced this PR)

`addPendingFiles` calls `setPendingFiles(...)` then `submitComment()` synchronously. `submitComment` reads `commentDraft` via closure. If the user types text and immediately attaches files (without a React re-render between), `submitComment` may see the pre-update `commentDraft`. The `toMarkdown()` call inside `submitComment` would serialize stale editor state.

**Data-flow trace:** `CommentComposer.onChange(rawText)` → `setCommentDraft` (async React) → `submitComment` reads `commentDraft` (sync closure). In practice, the native file picker dialog blocks long enough for React to flush, making this a theoretical race.

**Recommendation:** Consider wrapping `submitComment` in a ref-based call or reading draft from `composerRef.current` directly. Low urgency.

---

#### Nit — `viewMode.ts` helpers are generic enough for a shared util

**File:** `packages/dmloop/src/ui/viewMode.ts:1-16`
**Diff-scope:** New file

`readView`/`writeView` are generic localStorage wrappers with allowlist validation. Currently consumed by `IssuePage` and `ProjectPage` only. Could live in a shared UI util for reuse, but acceptable as-is with only two consumers.

---

### 3. Data-Flow Tracing

**Core route: `/s/:taskNo` → SummaryDetailPage**

```
URL pathname
  → parseStandaloneSummaryTaskNo() [regex match, returns string | null]
  → isStandaloneSummaryPath() [delegates to parser !== null]
  → Layout renders <SummaryDetailPage taskId={standaloneTaskNo ?? undefined} />
  → SummaryDetailPage.detailLookupId [props.taskId ?? null, type: number | string | null]
  → loadDetail() uses detailLookupId for API call
  → getSummaryDetail(detailLookupId) [accepts number | string]
  → Server returns detail with numeric task_id
  → state.detail.task_id set
  → this.taskId getter: typeof props.taskId === "number" ? props.taskId : state.detail?.task_id ?? null
  → All mutation APIs (regenerate, cancel, respond, removeMember, leave, batchStatus) use this.taskId (always numeric after load)
```

**Verdict:** Clean. The dual-type is properly confined to the lookup path; mutations always flow through the numeric getter. `useSummaryDetail` hook guards with `typeof taskId !== "number"` before calling mutation APIs.

**Workspace slug generation:**

```
User types name
  → getPinyin(name) [Chinese → ASCII transliteration]
  → normalizeBase(pinyin) [lowercase, dash non-alnum, trim, slice(34), re-strip trailing dash]
  → slugSuffix() [4-char WebCrypto a-z0-9]
  → withRandomSuffix(base, suffix) [base || "ws" + "-" + suffix]
  → POST /workspaces { slug }
  → 409 → retry (up to 3 attempts with fresh suffix)
  → user-typed slug conflict → show "slugTaken" error
```

**Verdict:** Clean. Edge cases (CJK-only, emoji, truncation-at-dash) all handled and tested.

**InteractiveCard openUrl interception:**

```
openUrl(url)
  → new URL(url) [throws on invalid → caught → window.open fallback]
  → parsed.pathname.match(SUMMARY_DETAIL_PATH_RE)
  → match → WKApp.openSummaryDetail(taskNo) [if defined]
  → undefined → window.open(url, "_blank", "noopener,noreferrer")
```

**Verdict:** Clean. Safe fallback when summary routing unavailable.

---

### 4. Blind-Point Checklist (C1–C6)

| Check | Result | Evidence |
|-------|--------|----------|
| **C1 — Dual-path parity** | Clear | `/s/:taskNo` follows the same pattern as `/d/:docId`: same regex style, same session recovery, same anonymous stash/return flow, same `isSafeReturnPath` extension. Tested in both `layoutStandaloneDocPath.test.ts` and `StandaloneDocPage.test.tsx`. |
| **C2 — Control-flow ordering** | Clear | `STANDALONE_SUMMARY_PATH` regex is applied once at the Layout level and once in `isSafeReturnPath`. No double-application risk. `openUrl` in `actions.ts` tests summary match before falling through to `window.open`. |
| **C3 — Authorization boundary** | N/A | No new endpoints or auth changes. `/s/:taskNo` is a client-side route; server-side auth on `getSummaryDetail` is unchanged. |
| **C4 — Authorization lifecycle** | N/A | No auth/container status changes. |
| **C5 — Build/note pass ≠ runtime correct** | Clear | Tests exercise the actual runtime paths (URL parsing, regex matching, API call shapes). `layoutStandaloneDocPath.test.ts` reads the actual `Layout/index.tsx` source and validates ordering. |
| **C6 — Governance/docs self-consistency** | N/A | No governance or policy changes. |

---

### 5. Cross-Round Blocker Review (R6)

---

### 6. Summary

This PR adds a well-structured `/s/:taskNo` standalone deep-link route mirroring the existing `/d/:docId` pattern, along with several quality-of-life improvements: tiptap-based comment composer with @mention, pinyin-based workspace slug generation, im-runtime helper extraction, and view-mode persistence.

The core data-flow contract (string taskNo for lookup → numeric task_id for mutations) is sound and properly guarded at every layer. Open-redirect safety is correctly extended. Test coverage is comprehensive.

No P0 or P1 blockers identified. Two P2 nits noted (cache TTL, theoretical stale closure) — neither blocks shipping.

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 nits and one nit-level observation, all non-blocking. The `/s/:taskNo` route, data-flow contracts, open-redirect guards, and test coverage are all solid.

yujiawei (changes requested):
# Code Review — PR #740 (octo-web)

Reviewed at head SHA `a5ba82e14d1a57309cdaa64968266ed55161beb5` against `main` (merge-base `b7f86cd4`).

Solid, well-scoped change that faithfully mirrors the existing `/d/:docId` standalone route. The open-redirect guard reuse is correct, the strict single-segment matcher is right, and the `taskId` widening is threaded carefully through most of the loader. One correctness defect in the summary detail page blocks merge.

The diff implements what the description states — a `/s/:taskNo` deep-link route mirroring `/d/:docId`, login-redirect `sp` forwarding, anonymous stash-and-return, and the shared return-target validator extension. The in-app card jump (`openUrl` interception) is in scope per the commit. No unrequested features or scope creep. **Spec: OK.**

## 2. Code quality

### P1 — Standalone `/s/:taskNo` entry silently fails to load per-person data for BY_PERSON summaries

`packages/dmworksummary/src/pages/SummaryDetailPage.tsx`

The refactor correctly routes the *detail* fetch through the new string-capable `detailLookupId` getter, but the downstream secondary loaders still gate on the `taskId` getter, which for a **string** prop resolves through `this.state.detail`:

```ts
// line 315-317
get taskId(): number | null {
    return typeof this.props.taskId === "number" ? this.props.taskId : this.state.detail?.task_id ?? null;
}
```

In `loadDetail`, the BY_PERSON secondary loaders are invoked **synchronously** right after `this.setState({ detail })`:

```ts
// line ~380
this.setState({ detail, loading: false, ... });   // async — this.state.detail is NOT updated yet
...
// line 392-394
if (detail.summary_mode === SummaryMode.BY_PERSON) {
    this.loadPersonalResult(seq);   // first line: if (this.taskId == null) return;
    this.loadMembers(seq);          // first line: if (this.taskId == null) return;
}
```

Because `setState` does not update `this.state` synchronously, at the moment `loadPersonalResult`/`loadMembers` run, `this.state.detail` is still `null`, so for a string prop the `taskId` getter returns `null` and **both loaders early-return**. `componentDidUpdate` does not re-trigger them (the string prop is unchanged, `prevTaskId === detailLookupId`).

Impact: a **completed** BY_PERSON summary opened via a `/s/:taskNo` notification deep-link renders the team summary but never loads the personal reports or member list — silently. (A still-processing task can recover later via the fallback poll on a status change; a completed task has no such recovery path.) This path is exercised only by the new string entry — normal in-app numeric navigation is unaffected — so the existing tests do not catch it.

Suggested fix: base the secondary loaders on the just-fetched `detail` (e.g. pass `detail.task_id` explicitly, or read the numeric id from the local `detail` variable) rather than on `this.taskId`, which depends on state that has not yet flushed. Please add a test that mounts with a **string** `taskId` and asserts `loadPersonalResult`/`loadMembers` fire for a BY_PERSON `detail`.

### P2 — `openUrl` intercepts `/s/:taskNo` on *any* host

`packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts:17-26`

The interception matches `parsed.pathname` only and ignores `parsed.hostname`. Any external link in a card whose path happens to be `/s/<segment>` (e.g. `https://some-external-site.example/s/abc`) is hijacked into the internal summary panel instead of opening externally, and then fails the detail lookup. The `/d/:docId` route does not have this cross-origin ambiguity because it is only reached via same-origin routing. Consider gating the interception on same-origin (compare `parsed.origin` / `parsed.hostname` against the app origin) so only genuinely internal deep-links are captured. Low likelihood, easy to close.

### P2 — Login-redirect query builder forwards `sp` but not the `space` alias

`apps/web/src/Layout/index.tsx:150,158` vs `applyStandaloneSummarySpaceFromQuery` (line 92-98)

The render branch accepts both `?sp=` and `?space=` for the space id, but the `onLogin` redirect-query builder only forwards `sp`. In practice the anonymous return is driven by `persistStandaloneReturn()` (which stashes the full `pathname + search`, preserving `?space=`), so the primary return path is unaffected — but the two code paths disagree on which aliases they honor. Either forward both aliases or standardize on `sp` in both places to avoid the latent inconsistency.

### P2 (nit) — Global state mutation during `render()`

`applyStandaloneSummarySpaceFromQuery` writes `WKApp.shared.currentSpaceId` from inside the `render()` branch. Mutating shared state during render is an anti-pattern under React 18 `createRoot`. There is existing precedent (the CLI-authorize branch does the same), so this is consistent with the codebase rather than a regression — flagging for future cleanup, not as a blocker.

### Note — `useSummaryDetail` widening is currently dead code

`packages/dmworksummary/src/hooks/useSummaryDetail.ts` — the hook has no callers in the repo, so the `typeof taskId !== "number"` guard on `regenerate`/`cancel` cannot bite today. If this hook is ever wired up with a string `taskNo`, those actions would silently no-op; worth a comment or a follow-up, but not blocking.

## 3. Overall verdict

**CHANGES_REQUESTED** — the P1 causes a completed BY_PERSON summary opened via the new deep-link (the feature's core path) to silently omit per-person reports and the member list. The P2 items are non-blocking cleanups.

## 4. Coverage note

Verified: open-redirect guard reuse and its new `/s/` test cases, the `taskId`/`detailLookupId` getter split and its use across `loadDetail`, the fallback-poll / status-change reload paths, `openUrl` interception, and cross-file consumers of `openSummaryDetail`. Not verified (out of reach in this front-end diff): actual runtime `vitest` execution (no `node_modules` in the review checkout — relied on source analysis and the author's reported green run), the server contract for `GET /summaries/{task_no}` accepting a string id, the nginx SPA `try_files` fallback that serves `/s/*` deep-links (the template has only a bare `location /`), and which summary modes the smart-summary notification cards actually emit `/s/` links for.

Two independent advisory passes were run alongside this review; one completed and one was unavailable due to an environment/gateway error (treated as absent, not as agreement).

yujiawei (approved):
# Code Review — PR #740 (octo-web) — re-review at new head

Re-reviewed at head SHA `ba3463755a951e7bf9c642980170377a714a0793` against `main` (merge-base `b7f86cd4`). The prior CHANGES_REQUESTED was against `a5ba82e1`; this is a fresh verdict on the current head (an approve does not auto-inherit across a new push).

The single blocking defect from the prior round is fixed. The BY_PERSON secondary loaders no longer depend on `this.taskId` (which, for a string `/s/:taskNo` entry, resolved through not-yet-flushed `state.detail` and early-returned):

```ts
// loadDetail — now passes the just-fetched numeric id explicitly
this.loadPersonalResult(seq, detail.task_id);
this.loadMembers(seq, detail.task_id);

// loaders take an override, falling back to this.taskId for the in-app numeric path
async loadPersonalResult(seq?: number, taskIdOverride?: number) {
    const requestTaskId = taskIdOverride ?? this.taskId;
    if (requestTaskId == null) return;
    ...
    const result = await api.getPersonalResult(requestTaskId);
```

This is correct. The post-await ownership guard (`this.taskId !== requestTaskId`) still holds: after the detail resolves, `this.taskId` returns the numeric `state.detail.task_id`, which equals the override — so legitimate responses land and genuine task-switch responses are still discarded. A new test (`string task_no detail load starts BY_PERSON secondary fetches with numeric detail.task_id`) mounts with a string `taskId` and asserts `getPersonalResult`/`getMembers` fire with the numeric `740`, closing the gap the earlier tests missed.

**Spec: OK.** Still a faithful mirror of `/d/:docId` — strict single-segment matcher, `sp` login-redirect forwarding, anonymous stash-and-return, shared open-redirect validator extension, in-app card jump. No scope creep.

## 2. Code quality

**Quality: Approved.** No P0/P1 remaining. Two non-blocking P2 nits carry over from the prior review (unchanged at this head):

- **P2 —** `openUrl` (`InteractiveCard/renderer/actions.ts`) matches `pathname` only, ignoring hostname, so an external card link whose path is `/s/<seg>` is captured into the internal panel. Gate on same-origin. Low likelihood.
- **P2 —** The login-redirect builder (`Layout/index.tsx`) forwards `sp` but not the `space` alias that `applyStandaloneSummarySpaceFromQuery` also accepts. The primary anonymous return is unaffected (`persistStandaloneReturn` stashes the full search string); this is a latent inconsistency only.

Neither blocks merge — track as follow-ups.

## 3. Overall verdict

**APPROVED.** The blocking P1 is resolved with a correct, tested fix; only non-blocking P2 nits remain.

## 4. Coverage note

Verified at the new head: the P1 fix and its guard semantics, the new string-entry BY_PERSON test, and that both P2s are unchanged. Re-confirmed the open-redirect guard and `openUrl` interception. Not verified (out of reach in this front-end diff): actual `vitest` execution (no `node_modules` in the review checkout — source analysis + author's reported green run), the server contract for `GET /summaries/{task_no}` accepting a string id, and the nginx SPA `try_files` fallback serving `/s/*` deep-links (the template has only a bare `location /`).

OctoBoooot (approved):
## Review: #740 @ `ba346375` — DELTA re-review; BY_PERSON string-taskId 🔴 does not reproduce (byte-traced), re-affirming APPROVE

**Verdict:** Approve. Re-reviewed at the amended head. yujiawei's 🔴 (string taskId breaks BY_PERSON personal-report + member-list display) does **not reproduce** here — I traced the mechanism three ways and it's correct, and the commit adds a test that pins exactly this scenario. All code CI is green. My prior APPROVE (@`a5ba82e1`) was auto-dismissed by the amend; re-anchoring at the live head. Delta over my prior head is `ahead 1/behind 1` (the commit was amended; `SummaryDetailPage.tsx` grew to +22/-17 with a new test).

## Re: the BY_PERSON 🔴 — byte-verified it works with a string taskId
The concern is that a string `taskId` (deep-link `/s/:taskNo`) makes the BY_PERSON secondary fetches (personal result, members) fail so nothing displays. Traced at `ba346375`:
1. **Fetches use the numeric id, not the string prop.** In `loadDetail`, after `setState({detail})`, BY_PERSON calls `loadPersonalResult(seq, detail.task_id)` / `loadMembers(seq, detail.task_id)` — passing the numeric server `detail.task_id` (e.g. 740), not `props.taskId` (the string). `getPersonalResult(740)` / `getMembers(740)` fire with the number.
2. **The post-await ownership guard compares number-to-number.** Inside `loadPersonalResult`, `requestTaskId = taskIdOverride = detail.task_id = 740`. The guard `this.taskId !== requestTaskId` uses the getter (`:315`): `typeof props.taskId === "number" ? props.taskId : this.state.detail?.task_id ?? null`. In the string-prop case it returns `state.detail.task_id = 740` (detail was setState'd just before). So `740 !== 740` is false → guard passes → `personalResult`/`members` are committed and displayed.
3. **A new test pins exactly this** (`SummaryDetailPage.schedule.test.tsx`): `makePage('SUM-740')` (string) with a BY_PERSON detail (`task_id: 740, summary_mode: 2`) → asserts `getPersonalResult` and `getMembers` are called with **740** and the detail setState carries `task_id === 740`.

The failure mode she describes would require `this.taskId` to return the *string* prop (then `string !== number` would discard every response) — but the getter deliberately returns the numeric `detail.task_id` when the prop is a string. That is the fix, and it's now tested. I could not construct the failing case; refuting the 🔴 on this evidence. (If it was valid at an earlier head, the amend that grew `SummaryDetailPage.tsx` +22/-17 resolved it.)

## Her two 🟡 (non-blocking, fair)
- **`openUrl` ignores the URL origin when intercepting `/s/:taskNo`** — confirmed: the card intercept matches on pathname only and calls `WKApp.openSummaryDetail(taskNo)` regardless of host. This is benign-by-design (the taskNo is the internal detail signal; authz is server-enforced per-resource on the fetch, so a cross-origin card can't open a summary the user can't access) — but she's right that an external `/s/:taskNo` link would be captured in-app rather than opened externally. If external same-path links are a real scenario, gate the intercept on same-origin. Minor.
- **Login redirect doesn't forward the `sp` (space alias) param** — plausible context-loss on the bounce; minor, worth confirming the post-login landing still resolves the right space.

## Still verified (unchanged, security-critical)
Open-redirect guard intact (three gates), anonymous-no-leak, string→fetch / numeric→write separation, card-jump guarded (isSafeUrl before parse). Code CI all green: Build SUCCESS, osv-scan SUCCESS, secret-scan SUCCESS, pr-title-lint SUCCESS.

## Net
The BY_PERSON string-taskId 🔴 doesn't reproduce (byte-traced: numeric `detail.task_id` drives the secondary fetches + guard, and a new test pins it); her two 🟡 are fair non-blocking notes (origin-agnostic intercept is intended; sp-forwarding is minor). Re-affirming APPROVE at `ba346375`. Credit yujiawei — the BY_PERSON path was exactly the right place to probe for a string/number regression; it happens to hold here. Note `code-review` check is red reflecting the active-CR review state, not a code failure; it clears when the review resolves.

Jerry-Xin (approved):
Re-review at ba346375 (my earlier approve at a5ba82e1 auto-dismissed by the new commit). Re-verified the security posture and independently checked @yujiawei's three points against the code. The new commit refines the BY_PERSON numeric-id resolution and adds regression tests. Security posture intact. Approving.

### 🔴 (raised by @yujiawei) — appears REFUTED: BY_PERSON loads correctly with a string task_no

Traced the deep-link (string prop) BY_PERSON data path:

- `SummaryDetailPage.loadDetail()` uses `detailLookupId` (= `props.taskId`, may be a string) only for the initial `api.getSummaryDetail(lookupId)` — the detail endpoint is widened to `number | string` in `summaryApi.ts` and returns the full detail including the numeric `detail.task_id`.
- For BY_PERSON it then calls `this.loadPersonalResult(seq, detail.task_id)` and `this.loadMembers(seq, detail.task_id)` — passing the fetched **numeric** `detail.task_id` (`SummaryDetailPage.tsx`, loadDetail BY_PERSON branch).
- `loadPersonalResult(seq, taskIdOverride)` / `loadMembers(seq, taskIdOverride)` use `requestTaskId = taskIdOverride ?? this.taskId`, so `api.getPersonalResult(...)` / `api.getMembers(...)` (both typed `taskId: number` in `summaryApi.ts`) receive the numeric id — never the string.
- `get taskId()` now returns `typeof props.taskId === "number" ? props.taskId : state.detail?.task_id ?? null`, so the post-await ownership guard `this.taskId !== requestTaskId` compares numeric-to-numeric once `state.detail` has hydrated.

This is directly asserted by the new test in `SummaryDetailPage.schedule.test.tsx` ("string task_no detail load starts BY_PERSON secondary fetches with numeric detail.task_id"): `getSummaryDetail` called with `'SUM-740'`, while `getPersonalResult`/`getMembers` are called with `740`. Build CI is green, so this passes. I don't see a path where the string reaches the personal-report / member-list fetches — could you share a concrete repro if you're still seeing a blank BY_PERSON body via the deep-link?

### 🟡 (raised by @yujiawei) — space alias forwarding: appears addressed

`Layout/index.tsx` reads `forwardSp = getQueryParam("sp")` and, for a standalone summary path, sets `redirectQuery.set("sp", forwardSp)` so `sp` survives the login redirect; the authed render calls `applyStandaloneSummarySpaceFromQuery()` (reads `sp`/`space` → `currentSpaceId`) before mounting the page. Covered by `layoutStandaloneDocPath.test.ts` ("carries sp through post-login return").

### 🟡 (raised by @yujiawei) — openUrl interception ignores origin: real but minor (non-blocking)

Confirmed: `openUrl` in `InteractiveCard/renderer/actions.ts` matches only `parsed.pathname` against `/^\/s\/([A-Za-z0-9_-]+)\/?$/` after the `isSafeUrl` gate — it does not compare `parsed.origin` to the app origin. So a card `Action.OpenUrl` pointing at another origin's `/s/<segment>` (e.g. `https://other.example.com/s/abc`) would be routed in-app via `openSummaryDetail(abc)` instead of opening externally. Impact is contained: it stays in-app (no open-redirect), the summary id is still server-authz-gated (no IDOR/leak), and it only affects trusted card content. Suggest (non-blocking) scoping the interception to same-origin/relative paths so a genuinely external `/s/` link isn't hijacked.

### Security re-confirm vs my prior approved head

- Open-redirect guard `isSafeReturnPath` (XIN-392) unchanged in structure: parse-safe + `url.origin === origin` + target-scope; the scope is deliberately widened to also accept `/s/:taskNo` via the same single-segment regex, so a tampered sessionStorage return value still can't bounce to `/settings`, `/oidc/bind`, etc. — only standalone doc/summary targets.
- Numeric-only write ops strengthened: `useSummaryDetail` `regenerate`/`cancel` now guard `typeof taskId !== "number"`, so a string task_no can't reach the numeric write endpoints.
- Token-gated render (anon → `persistStandaloneReturn()` + login), detail fetched via the shared authenticated interceptor, single-segment no-traversal regex consistent across Layout / actions / StandaloneDoc. `isSafeUrl` unchanged.

### CI / scan

Build success, pr-title-lint pass, osv-scan / secret-scan / dependency-review pass. No new listeners/timers/console in non-test code; no phantom deps (`@dmwork/summary`, relative `../../../App`, vitest devDep). No hardcoded user-facing strings (errors go through `t(...)`).

### ✅ Highlights

- BY_PERSON string→numeric resolution is explicit and test-covered; the numeric `detail.task_id` is threaded to the number-keyed secondary fetches rather than relying on async-hydrated state.
- Race/stale-response guards (`seq` + `taskId`) preserved through the refactor.
- Return-target open-redirect scope reused rather than reinvented; strong test coverage across Layout, actions, and the detail page.

LGTM. The two 🟡s are non-blocking; the origin-scoping of openUrl is worth a follow-up.

PR #741, opened 2026-07-14
refactor(wksdk): 收口聊天连接状态 runtime 访问

menghao-webtest:
## Summary
- Route ChatVM connect status reads and listener lifecycle through im-runtime helpers.
- Move reconnect refresh gating for the active conversation into a tested runtime helper.
- Remove the unused app-level ConnectionStatus duplicate so the base component remains the single active entry.

## Not Changed
- No SDK version change.
- No message send behavior change.
- No message sync strategy change.
- No UI or text change for the active connection status component.

## Tests
- pnpm --dir packages/dmworkbase test src/im-runtime/connectStatus.test.ts src/Pages/Chat/__tests__/vm.sortConversations.test.ts src/Pages/Chat/__tests__/vm.channelListener.test.ts
- pnpm --dir apps/web build
- git diff --check

## Notes
- Existing Conversation test files are currently blocked in jsdom by lottie-web canvas initialization before assertions run.

PR #742, opened 2026-07-14
fix(ui): 修复控制台警告和连接状态显示

menghao-webtest:
## Summary
- 修复 Sections 动态组件 props 中 key 被 spread 导致的 React warning
- 过滤 Markdown 渲染器转发到 DOM 的内部字段，避免 ordered=false warning
- 健康检查失败时同步更新连接状态展示，避免断网后仍显示已连接

## Tests
- pnpm --dir packages/dmworkbase test src/Messages/Text/__tests__/MarkdownContentMention.test.tsx src/Messages/Text/__tests__/MarkdownContentForwardClamp.test.tsx
- pnpm --dir packages/dmworkbase test src/im-runtime/connectStatus.test.ts src/Pages/Chat/__tests__/vm.sortConversations.test.ts src/Pages/Chat/__tests__/vm.channelListener.test.ts
- pnpm --dir apps/web build
- git diff --check

### DIFF
--- a/packages/dmworkbase/src/Components/ConnectionStatus/index.tsx
+++ b/packages/dmworkbase/src/Components/ConnectionStatus/index.tsx
@@ -90,10 +90,23 @@ export default class ConnectionStatus extends Component<ConnectionStatusProps, C
             })
             const latency = Date.now() - start
             if (isImConnected(WKSDK.shared())) {
-                this.setState({ latency })
+                const nextState: Partial<ConnectionStatusState> = {
+                    status: ConnectStatus.Connected,
+                    latency,
+                }
+                if (!this.state.connectedSince) {
+                    this.connectedTime = Date.now()
+                    nextState.connectedSince = this.connectedTime
+                }
+                this.setState(nextState as any)
             }
         } catch {
-            // ignore
+            this.connectedTime = 0
+            this.setState({
+                status: ConnectStatus.Disconnect,
+                latency: null,
+                connectedSince: null,
+            })
         }
     }
 
--- a/packages/dmworkbase/src/Components/Sections/index.tsx
+++ b/packages/dmworkbase/src/Components/Sections/index.tsx
@@ -21,8 +21,10 @@ export default class Sections extends Component<SectionsProps> {
                         <div className="wk-channelsetting-section-rows">
                             {
                                 section.rows?.map((row, j) => {
+                                    const Cell = row.cell
+                                    const { key: cellKey, ...cellProps } = row.properties ?? {}
                                     return <div key={j} className="wk-section-row">
-                                        <row.cell  {...row.properties}></row.cell>
+                                        <Cell key={cellKey} {...cellProps}></Cell>
                                     </div>
                                 })
                             }
--- a/packages/dmworkbase/src/Messages/Text/MarkdownContent.tsx
+++ b/packages/dmworkbase/src/Messages/Text/MarkdownContent.tsx
@@ -575,7 +575,7 @@ const MarkdownContent: React.FC<MarkdownContentProps> = ({
       );
     const wrap =
       (Tag: string) =>
-      ({ node, children, ...props }: any) =>
+      ({ node, children, ordered, checked, index, siblingCount, ...props }: any) =>
         React.createElement(Tag, props, process(children));
     return {
       ...baseComponents,


OctoBoooot (approved):
## Review: #742 — console-warning fixes + health-check connection-status sync

**Verdict:** Approve. Three small, correct fixes: two React console-warning removals (no behavior change) and one genuine behavior fix (connection status now reflects a failed health check). Risk tier: normal. CI: all code checks green (Build SUCCESS, secret-scan, osv-scan, dependency-review, pr-title-lint); `code-review` (human gate) pending.

### Fix 1 — ConnectionStatus reflects health-check failure (`ConnectionStatus/index.tsx`)
The real behavior fix. Previously the health-check `catch` was a bare `// ignore`, so a failed probe left the UI showing stale "Connected" after the network dropped. Now:
- **failure** → `status: Disconnect, latency: null, connectedSince: null` + `connectedTime = 0` (connection-lost state propagates to the UI);
- **success** → explicitly sets `status: Connected` and stamps `connectedSince`/`connectedTime` on first connect.
Verified consistent with the existing listener path (`:44-60` sets the same fields the same way) and the #741 consolidation (reads via `isImConnected(WKSDK.shared())`, doesn't touch listener lifecycle). `connectedSince` is declared state (`:24`), `connectedTime` a private field (`:34`), and `render` guards the duration with `connected && connectedSince` (`:170`) so no stale "connected for Xs" shows after a drop. Symmetric and coherent.

### Fix 2 — React key-spread warning (`Sections/index.tsx`)
`<row.cell {...row.properties}>` spread a `key` from `properties` into JSX (React warns). The fix destructures `{ key: cellKey, ...cellProps }`, passes `key={cellKey}` explicitly, and hoists `const Cell = row.cell` for valid capitalized-component JSX. Correct; the wrapper `<div key={j}>` is a separate element so there's no key collision.

### Fix 3 — filter react-markdown internal fields from the DOM (`MarkdownContent.tsx`)
`react-markdown` forwards node metadata (`ordered`, `checked`, `index`, `siblingCount`) to the component wrapper; spreading `...props` onto a DOM tag leaked them as unknown HTML attributes (the `ordered=false` non-boolean-attribute warning). Destructuring them out before `React.createElement(Tag, props)` is the correct, minimal fix — they're react-markdown internals, not DOM attrs.

### Praise
Fix 1 is the right shape: rather than only silencing a symptom, it makes the failure path (health check throws) update the same state the success path does, so the connection indicator can't lie after a drop — and it reuses the existing `connectedSince`/`connectedTime` machinery instead of inventing a parallel one.

### Nit (non-blocking)
`ConnectionStatus/index.tsx:~101` casts `this.setState(nextState as any)`. The `Partial<ConnectionStatusState>` object is already correctly typed — the `as any` is unnecessary and slightly weakens type safety; `setState(nextState)` should compile as-is.

### Net
Two warning fixes (pure noise removal, no behavior change) and one legitimate connection-status correctness fix (failure now propagates to the UI), all minimal and correct, tests + Build green. Approving. Credit menghao-webtest. Gate stays on the required `code-review` check.

yujiawei (commented):
# Code Review — PR #742 (octo-web)

Focused review of `ConnectionStatus` lifecycle/concurrency. This PR merged before review completed; findings below are post-merge and warrant a follow-up. **Assessed verdict: CHANGES_REQUESTED** (P1 correctness issues in the new `catch` path).

## Context
`measureLatency()` runs on a 5s `setInterval` and probes an **HTTP `/health`** endpoint that is *distinct* from the **IM websocket** connection. A separate `statusListener` independently drives `status` from the real websocket `ConnectStatus` and manages the ping timer. The new `catch` block now also writes `status`, creating two authorities for one piece of state.

## P1 — HTTP `/health` failure overrides live websocket status
`index.tsx` L103–110 (catch): on any non-2xx or timeout, `apiFetch` throws `ApiFetchError` and the catch forces `status: Disconnect, latency: null, connectedSince: null`. This runs even when the IM websocket is fully `Connected`. A transient 5xx/timeout on the unrelated `/health` HTTP route therefore flips the UI to "Disconnected" and drops the connected-since timer, while chat is actually alive. The success path is correctly gated on `isImConnected(...)` (L92) — the failure path is not gated at all, so the two are asymmetric. The catch should not assert websocket-down state from an HTTP probe failure; at most degrade latency (e.g. keep `status` untouched / show `latency: null` bars) when `isImConnected()` is still true.

## P1 — Race between the two setState sources; catch leaves the interval running
The catch does **not** call `stopPing()` (unlike the listener's non-Connected path at L52), so after a failed probe the 5s interval keeps firing. Meanwhile `statusListener` may set `status: Connected` from the websocket. The two writers interleave with no ordering guarantee: listener sets `Connected` → next tick's catch sets `Disconnect` → listener isn't re-invoked (websocket didn't change) → UI is stuck showing Disconnected until the *next* successful `/health` tick flips it back. During a flaky-HTTP / healthy-socket window the indicator oscillates. Because ping is never stopped on error, this is self-inflicted churn rather than a one-shot glitch. Making the ping the source of truth for `status` conflicts with the listener being the source of truth; only one should own `status`.

## P2 — setState after unmount possible on in-flight probe
`measureLatency` is `async` and both success (L101) and catch (L105) call `this.setState` with no mounted guard. `componentWillUnmount` (L65–68) removes the listener and clears the timer, but a probe already awaiting `apiFetch` at unmount time will resolve/reject afterward and call `setState` on an unmounted component (React "setState on unmounted component" warning; harmless leak, but ironic in a PR titled "fix console warnings"). Add an `isMounted`/`AbortController` guard and short-circuit both branches.

## Note (not blocking)
- `connectedTime` private field (L98, L104) is written but never read anywhere; `connectedSince` state is what render uses. Dead field — consider removing.

## FILE 2 — Sections/index.tsx
Extracting `Cell` to a capitalized local and stripping `key` out of spread props (`const { key, ...cellProps } = row.properties ?? {}`) is correct and fixes the React "key spread" warning. `properties` is `any` so no type regression. LGTM.

## FILE 3 — MarkdownContent.tsx
Destructuring `ordered/checked/index/siblingCount` out of react-markdown renderer props before spreading to the DOM correctly suppresses the "received `ordered=false`" DOM-attribute warnings. Applied uniformly to p/td/th/li/h1..h6. LGTM.

yujiawei (changes requested):
# Code Review — PR #742 (octo-web)

Reviewed against the head SHA and the three stated fixes. The diff is tightly scoped to exactly the three files claimed, with no out-of-scope changes. Fixes (1) and (2) are correct and complete. Fix (3) is functionally partial and introduces a state-source conflation that will cause UI flapping.

## Spec compliance

**Verdict: mostly compliant, one behavioral gap in fix (3).**

- **Fix 1 — Sections key-spread warning** ✅ `packages/dmworkbase/src/Components/Sections/index.tsx:24-28`. `key` is now destructured out of `row.properties` and passed explicitly on `<Cell>`, so it is no longer spread as a prop. The `?? {}` guard is correct since `properties?` is optional. Implements exactly what was claimed.
- **Fix 2 — Markdown internal-field DOM leak** ✅ `packages/dmworkbase/src/Messages/Text/MarkdownContent.tsx:578`. `ordered, checked, index, siblingCount` are destructured out of the wrapper props so react-markdown internal fields no longer reach `React.createElement(Tag, props, ...)`. Since `wrap` is applied to `p/td/th/li/h1..h6`, this correctly suppresses the `received ordered=false` class of warnings for those renderers.
- **Fix 3 — health-check failure updates connection status** ⚠️ `packages/dmworkbase/src/Components/ConnectionStatus/index.tsx:103-110`. The catch now flips `status` to `Disconnect` and clears latency/connectedSince. It does address the stated "avoid stale connected" goal, but it does so by treating an HTTP `/health` failure as equivalent to an IM websocket disconnect — see P1 below.

No out-of-scope additions, no missing files. Scope is clean.

## Findings

### P1 — `/health` HTTP failure is conflated with IM websocket connection state (index.tsx:103-110)

`measureLatency()` runs on a 5s `setInterval`. The `/health` endpoint is an HTTP endpoint distinct from the IM websocket, and `apiFetch` throws `ApiFetchError` on any non-2xx response. The new catch unconditionally sets `status: Disconnect` whenever that HTTP call fails — even when the IM websocket (the actual thing `status` represents, and the source of truth maintained by the `statusListener`) is still connected.

Consequences:
1. A transient `/health` blip (503, proxy hiccup, brief network loss on the HTTP path) will render the connection as Disconnected while the websocket is fine — a false negative visible to the user.
2. The catch does **not** call `stopPing()`, so the interval keeps firing. The next successful ping re-enters the success branch, `isImConnected()` is still true, and `status` flips back to `Connected`. The result is UI flapping (Disconnect → Connected every ~5s) for as long as `/health` is failing but the websocket is up.
3. There are now two writers of `status` (`statusListener` from the real websocket, and `measureLatency` from HTTP), and they can disagree. The listener is the authoritative source; the ping should not override it.

Suggested direction: on `/health` failure, guard with `isImConnected(WKSDK.shared())` before demoting `status` — only reflect a degraded/latency-unknown state (e.g. keep `status` from the listener and set `latency: null`) rather than forcing `Disconnect`. Let the websocket `statusListener` remain the single owner of `status`.

### P2 — New `as any` cast violates repo convention and is redundant (index.tsx:101)

The repo convention (CLAUDE.md: "avoid `any` — use proper types or `unknown` with type guards") is violated by the newly added `this.setState(nextState as any)`. `nextState` is already typed `Partial<ConnectionStatusState>`, which `setState` accepts natively — the catch block a few lines below (`setState({ status, latency, connectedSince })`) sets partial state with no cast at all, proving the cast is unnecessary here. Drop `as any` at line 101. (Note: the identical cast at line 54 is pre-existing and outside this PR's introduction, but the new one should not extend the anti-pattern.)

### P2 — No test for the new ConnectionStatus catch/disconnect branch

`ConnectionStatus/` has no test file, and this PR adds non-trivial state-transition logic (success sets `Connected`+`connectedSince` once; failure resets to `Disconnect`+nulls). Given the P1 flapping risk, a small vitest covering "health fails while websocket connected" and "health succeeds sets connectedSince once" would lock in the intended behavior. Not a merge blocker on its own, but recommended alongside the P1 fix.

## Summary

Fixes 1 and 2 are correct and can ship as-is. Fix 3 needs the P1 addressed before merge: as written it makes the HTTP `/health` probe a second, conflicting writer of the websocket connection status, producing false disconnects and 5s flapping when `/health` degrades independently of the socket.

packages/dmworkbase/src/Components/ConnectionStatus/index.tsx
```diff
@@ -90,10 +90,23 @@ export default class ConnectionStatus extends Component<ConnectionStatusProps, C
...
             }
         } catch {
-            // ignore
+            this.connectedTime = 0
+            this.setState({
+                status: ConnectStatus.Disconnect,
+                latency: null,
+                connectedSince: null,
+            })
         }
```
OctoBoooot, line 110: **[major]** (credit yujiawei) The health-check `catch` sets `status: ConnectStatus.Disconnect` unconditionally on an HTTP `/health` failure — but `status` is the IM **WebSocket** connection state (`:37` getImConnectStatus, `:44-54` WS listener, `:92` isImConnected). Failing input: IM WS connected and working, but `apiFetch(baseUrl + '/health')` rejects (transient REST 503 / rate-limit / HTTP blip) → this catch flips the indicator to 'Disconnected' (red, 'click to reconnect') every 5s while messaging works fine. Before this PR the catch was `// ignore`, so this conflation of two independent transports is PR-introduced. The success path already gates on `isImConnected` (`:92`); the catch needs the symmetric guard — only assert Disconnect when the WS is actually down, else treat it as latency-unknown with status unchanged: `} catch { if (!isImConnected(WKSDK.shared())) { this.connectedTime = 0; this.setState({ status: ConnectStatus.Disconnect, latency: null, connectedSince: null }) } else { this.setState({ latency: null }) } }`. The HTTP probe should inform latency, not override the authoritative WS status. (Also resolves the flicker from the timer continuing to fire — do NOT stopPing() in catch, that breaks recovery detection.)

OctoBoooot (changes requested):
## Correcting my #742 APPROVE → REQUEST_CHANGES — the health-check catch overrides the authoritative WS status (credit yujiawei)

I approved this holding the ConnectionStatus fix as a clean improvement. yujiawei found it introduces a real defect, and I byte-verified she's right on the lead 🔴. Flipping to REQUEST_CHANGES. Credit yujiawei.

### Major — HTTP `/health` failure falsely overrides the WebSocket connection status (`ConnectionStatus/index.tsx:103-110`)
`status` in this component is the **IM WebSocket** connection state — set from `getImConnectStatus(WKSDK.shared())` (`:37`), the WS status listener (`:44-54`), and gated on `isImConnected(...)` in the success path (`:92`). But the new `catch` sets `status: ConnectStatus.Disconnect` **unconditionally on an HTTP `/health` failure**, without checking whether the WS is actually connected.

Failing input: the IM WebSocket is connected and messaging works, but the periodic `measureLatency` fires and `apiFetch(`${baseUrl}/health`)` rejects — a transient REST 503, a rate-limit, a brief HTTP-side network blip. The catch runs → `setState({ status: Disconnect, latency: null, connectedSince: null })` → the indicator shows "Disconnected" (red, cursor pointer, "click to reconnect") even though IM is fully functional. This repeats every 5s (the ping interval).

Before this PR the catch was `// ignore`, so an HTTP health failure left the WS-derived status untouched — this PR **introduces** the conflation of two independent transports (REST health vs IM WebSocket). The success path already models the right rule (only assert Connected when `isImConnected`); the catch needs the symmetric guard: only assert Disconnect when the WS is actually down, e.g.
```
} catch {
    if (!isImConnected(WKSDK.shared())) {
        this.connectedTime = 0
        this.setState({ status: ConnectStatus.Disconnect, latency: null, connectedSince: null })
    } else {
        this.setState({ latency: null }) // WS up, health probe failed → latency unknown, status unchanged
    }
}
```
The HTTP health probe should inform **latency**, not override the authoritative WS-derived **status**.

### Related (fold into the same fix) — catch leaves the ping timer running → flicker (`:103-110` vs `:76-81`)
The catch sets Disconnect but doesn't `stopPing()`, so if `/health` flaps the indicator flips Connected↔Disconnect every 5s. Note the fix is NOT to call `stopPing()` in the catch — that would break recovery detection (the WS stays Connected, so its listener won't re-fire, and a dead timer would never resume probing → latency stuck). Fixing the status-source conflation above resolves the flicker at its root.

### Minor (pre-existing) — setState after unmount (`:83-111`)
`measureLatency` awaits `apiFetch` then `setState`; `componentWillUnmount` clears the interval but not an in-flight request, so unmounting mid-request setStates after unmount. This pattern pre-dates the PR (the success path already did await→`setState({latency})`) and is a no-op under React 18 — worth a `_mounted` flag / `AbortController` guard, but not blocking.

### Nit — `connectedTime` is redundant (`:34`)
It's read (into `connectedSince` at `:48/:60/:99`), so not strictly unused, but it always equals the value written to the `connectedSince` state one line later — a redundant mirror. Use `connectedSince` (or a local) and drop the field.

### Still good (unchanged)
The two React-warning fixes are correct and unaffected: `Sections/index.tsx` (key-spread → `{key: cellKey, ...cellProps}` + capitalized `Cell`) and `MarkdownContent.tsx` (filtering `ordered/checked/index/siblingCount` out of DOM-forwarded props). Only the ConnectionStatus behavior change needs revision.

### Accountability
I verified the success path's `isImConnected` gate but didn't check that the **catch path lacks the symmetric gate** — the failure branch of a two-independent-sources (REST vs WS) interaction. Banking: when a fix adds a failure/catch branch that writes shared state, check it against the same invariant the success branch enforces — an asymmetric guard is exactly where a "show disconnected on error" fix corrupts an authoritative status from an unrelated source.

### Net
Withdrawing my APPROVE: the health-check catch overrides the WebSocket-derived connection status on an unrelated HTTP failure (byte-verified, PR-introduced), falsely showing "Disconnected" while IM works. Gate the catch's Disconnect on `!isImConnected` so the HTTP probe informs latency only. The two warning fixes are fine. Credit yujiawei for the catch.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #742 (octo-web)

## Summary

This PR addresses three console-warning and display-quality fixes: (1) the `ConnectionStatus` component now actively manages `status` and `connectedSince` state on both the success and failure paths of the health-check ping, replacing the previous silent-ignore catch; (2) `Sections` fixes a lowercase JSX element warning by capitalizing `row.cell` into `Cell` and properly extracting `key` from the spread; (3) `MarkdownContent` strips react-markdown-specific props (`ordered`, `checked`, `index`, `siblingCount`) before forwarding to DOM elements, eliminating unknown-prop warnings.

The Sections and MarkdownContent changes are clean warning fixes with no behavioral side effects. The ConnectionStatus change introduces a new behavioral contract for the catch path that warrants attention.

## Verification

- ✅ **Diff scope** — 3 files, +20/−5. All changes are local to the affected components with no cross-module contract changes.
- ✅ **Sections key extraction** — `row.properties ?? {}` correctly handles `undefined`. Grepped all `new Row(...)` callers in `module.tsx`, `MeInfo/vm.tsx`, `ChannelSetting/vm.ts`, `ExperimentalFeatures/index.tsx` — none pass a `key` in `properties`, so `cellKey` is always `undefined` (harmless; outer `<div key={j}>` covers list identity).
- ✅ **MarkdownContent prop stripping** — react-markdown passes `ordered`, `checked`, `index`, `siblingCount` to `li` and list-adjacent components. The fix correctly excludes these before `React.createElement(Tag, props, ...)`. The remaining wrapped elements (`p`, `td`, `th`, `h1`–`h6`) do not receive additional non-DOM props in the versions used.
- ✅ **ConnectionStatus success path** — `!this.state.connectedSince` guard prevents resetting the timestamp on every successful ping. `componentDidMount` sets `connectedSince` first, so the guard is a safety net for the edge case where `measureLatency` runs before the listener fires.
- ✅ **Data-flow: `isImConnected`** — defined at `im-runtime/connectStatus.ts:67`, delegates to `sdk.connectManager.status === ConnectStatus.Connected`. Correctly reads live SDK state.

Static analysis only at head `61a53de9`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 design concern and one Nit below.

### P2 — Health-check catch path resets connection status on transient HTTP failures (`packages/dmworkbase/src/Components/ConnectionStatus/index.tsx:104`)

The catch block unconditionally sets `status: Disconnect` and `connectedSince: null` whenever the `/health` HTTP call fails, even though the IM WebSocket (`isImConnected()`) may still be alive. The health endpoint and the IM connection are independent transport paths — an HTTP timeout or transient load-balancer error does not imply the WebSocket is down.

Data-flow trace: catch fires → `connectedTime = 0`, `setState({ status: Disconnect, connectedSince: null })` → the `statusListener` at `:51` does NOT fire (no IM-level status change happened) → ping interval continues (no `stopPing()` in catch) → next successful ping (5s later) finds `!this.state.connectedSince` true → resets `connectedTime = Date.now()` → "connected for" counter starts from zero.

Net effect: every transient health check failure causes a brief "Disconnected" UI flicker and resets the "connected for" duration, even when the IM was connected the entire time. Before this PR, the catch was `// ignore` and the UI remained stable. This is new behavior introduced by the diff, not pre-existing.

Consider gating the disconnect on the actual IM connection state:

```ts
} catch {
    if (!isImConnected(WKSDK.shared())) {
        this.connectedTime = 0
        this.setState({
            status: ConnectStatus.Disconnect,
            latency: null,
            connectedSince: null,
        })
    }
}
```

### Nit — `as any` cast on setState (`packages/dmworkbase/src/Components/ConnectionStatus/index.tsx:101`)

React's `Component.setState` already accepts `Partial<S>`, so `Partial<ConnectionStatusState>` should be directly assignable without `as any`. The same cast appears at `:53` (statusListener), suggesting a project-wide habit. Harmless, but dropping it would let the type checker catch genuine shape mismatches.

- `Row.cell` type is `ElementType` (from `Service/Section.tsx:48`). Capitalizing to `Cell` for JSX rendering is the correct React pattern — lowercase JSX tags are treated as HTML intrinsics.
- `row.properties ?? {}` destructuring with `key` extraction handles all three edge cases: `undefined` properties (→ empty object fallback), properties without `key` (→ `cellKey` is `undefined`, harmless), and properties with `key` (→ properly passed as React key).
- The `wrap` function in `MarkdownContent.tsx:578` only wraps `p`, `td`, `th`, `li`, `h1`–`h6`. No `ol`/`ul` wrapping that would need `ordered`/`depth` stripping at those element levels.
- No new dependencies, no i18n changes, no CSS changes.
- No dangling references to the old `// ignore` pattern — the catch block is the only error handler in `measureLatency`.

## Verdict: APPROVED

No correctness, security, or blocking issues. The P2 is a design concern about the catch path's interaction with the IM connection state — worth addressing to avoid "connected for" counter resets during transient HTTP failures, but not a merge blocker. The Sections and MarkdownContent changes are clean warning fixes.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; one P2 design concern about catch-path over-aggressiveness flagged for author consideration.

Jerry-Xin (commented):
# Code Review — PR #742

Reviewed at head `61a53de934ad` against the three stated fixes. Diff is tightly scoped to exactly the three claimed files (+20/-5), no out-of-scope changes. Note: this PR was already self-merged before this review, so this is posted for the record — it cannot gate the merge. Fixes 1 and 2 are correct root-cause fixes. Fix 3 introduces a real defect.

## Fix 1 — Sections key-spread warning ✅
`Components/Sections/index.tsx:24-28`. `key` is destructured out of `row.properties` (`const { key: cellKey, ...cellProps } = row.properties ?? {}`) and passed explicitly on `<Cell key={cellKey} {...cellProps}>`, so `key` is no longer spread through props. Correct root fix for the React "key is not a prop / spread key" warning — not a suppression. The `?? {}` guard is correct since `properties` is optional, and capitalizing to `Cell` is required for JSX to treat it as a component. Clean.

## Fix 2 — Markdown internal-field DOM leak ✅
`Messages/Text/MarkdownContent.tsx:578`. `ordered, checked, index, siblingCount` are destructured out of the wrapper props before `React.createElement(Tag, props, ...)`, so react-markdown's internal fields no longer reach the DOM element. Correct root fix for the "React does not recognize the `ordered` prop / received `ordered=false`" class of warnings for the wrapped renderers. Not a suppression. Clean.

## Fix 3 — connection-status display 🔴 Blocking (behavioral regression this PR introduces)

`Components/ConnectionStatus/index.tsx:103-110`. In this component `status` represents the **IM WebSocket** connection state: it is seeded from `getImConnectStatus(WKSDK.shared())` (`:37`), driven by the WS status listener (`:44-54`), and the success path only asserts `Connected` when gated on `isImConnected(WKSDK.shared())` (`:92`).

The new `catch` block sets `status: ConnectStatus.Disconnect` (plus clears latency/connectedSince) **unconditionally on an HTTP `/health` failure**, with no `isImConnected` guard. Before this PR the catch was `// ignore`, so a health-probe failure left the WS-derived status untouched — this PR introduces the conflation of two independent transports (REST `/health` probe vs the IM WebSocket).

Failing scenario: the IM WebSocket is connected and messaging works, but the periodic `measureLatency` fires and the `/health` request rejects (transient 5xx, rate-limit, brief HTTP-path blip). The catch runs and forces the indicator to "Disconnected" even though IM is fully functional — a false negative visible to the user. Because `measureLatency` runs on a 5s `setInterval` (`:71`) and the catch does not stop the timer, the next successful ping re-enters the success branch (`isImConnected` still true) and flips back to `Connected` — producing UI flapping (Disconnect → Connected every ~5s) for as long as `/health` degrades independently of the socket. There are now two writers of `status` (the WS listener and the HTTP ping) that can disagree; the listener is the authoritative source.

Suggested direction (mirror the success path's guard): on `/health` failure, only demote `status` when the WS is actually down, and otherwise let the HTTP probe inform latency only —
```
} catch {
    if (!isImConnected(WKSDK.shared())) {
        this.connectedTime = 0
        this.setState({ status: ConnectStatus.Disconnect, latency: null, connectedSince: null })
    } else {
        this.setState({ latency: null }) // WS up, health probe failed → latency unknown, status unchanged
    }
}
```
This keeps the WS `statusListener` as the single owner of `status` and resolves the flapping at its root. (Do not add `stopPing()` in the catch — that would break recovery detection since the WS listener won't re-fire while the socket stays up.)

## Non-blocking notes
- 💬 `index.tsx:101` — the new `this.setState(nextState as any)` cast is unnecessary: `nextState` is already typed `Partial<ConnectionStatusState>`, which `setState` accepts natively (the catch block below sets partial state with no cast). Drop `as any`. The identical cast at `:54` is pre-existing.
- 💬 No test covers the new catch/disconnect branch. Given the flapping risk, a small test for "health fails while WS connected" and "health succeeds sets connectedSince once" would lock in intended behavior.
- 💬 `measureLatency` awaits then `setState` with no mounted/abort guard, so unmounting mid-request setStates after unmount. Pre-existing pattern, no-op under React 18, not blocking.

## Summary
Fixes 1 and 2 are correct and ship-ready. Fix 3 makes the HTTP `/health` probe a second, conflicting writer of the WebSocket-derived connection status, causing false "Disconnected" states and ~5s flapping when `/health` degrades while the socket is up. Gate the catch's demotion on `!isImConnected` so the probe informs latency only. Verdict: REQUEST_CHANGES (informational, PR already merged).

PR #743, opened 2026-07-14
fix(ui): 修正健康检查失败时的连接状态显示

menghao-webtest:
## Summary
- 修正 ConnectionStatus 中 HTTP /health 失败时强制覆盖 WKSDK/WebSocket 连接状态的问题
- 当 SDK 仍连接时，健康检查失败只清空延迟，不再显示为断开连接
- 增加 ConnectionStatus 单测，覆盖健康检查失败时 SDK 已连接/未连接两种状态

## Context
Follow-up for #742. The HTTP health probe should inform latency, while WKSDK/WebSocket remains the source of truth for connection status.

## Tests
- pnpm --dir packages/dmworkbase test src/Components/ConnectionStatus/__tests__/ConnectionStatus.test.ts src/im-runtime/connectStatus.test.ts src/Pages/Chat/__tests__/vm.sortConversations.test.ts src/Pages/Chat/__tests__/vm.channelListener.test.ts
- pnpm --dir apps/web build
- git diff --check

### DIFF
--- a/packages/dmworkbase/src/Components/ConnectionStatus/__tests__/ConnectionStatus.test.ts
+++ b/packages/dmworkbase/src/Components/ConnectionStatus/__tests__/ConnectionStatus.test.ts
@@ -1,0 +1,109 @@
+import { beforeEach, describe, expect, it, vi } from "vitest"
+import { ConnectStatus } from "wukongimjssdk"
+import ConnectionStatus from "../index"
+
+const hoisted = vi.hoisted(() => {
+    const sdk = {
+        connectManager: {
+            status: 1,
+            addConnectStatusListener: vi.fn(),
+            removeConnectStatusListener: vi.fn(),
+            connect: vi.fn(),
+        },
+    }
+
+    return {
+        sdk,
+        apiFetch: vi.fn(),
+    }
+})
+
+vi.mock("wukongimjssdk", () => ({
+    WKSDK: {
+        shared: () => hoisted.sdk,
+    },
+    ConnectStatus: {
+        Connected: 1,
+        Disconnect: 0,
+        Connecting: 2,
+        ConnectKick: 3,
+    },
+}))
+
+vi.mock("../../../App", () => ({
+    default: {
+        apiClient: {
+            config: {
+                apiURL: "https://api.example.com/",
+            },
+        },
+    },
+}))
+
+vi.mock("../../../Service/apiFetch", () => ({
+    apiFetch: hoisted.apiFetch,
+}))
+
+vi.mock("../../../i18n", async () => {
+    const React = await vi.importActual<typeof import("react")>("react")
+    return {
+        I18nContext: React.createContext({
+            t: (key: string) => key,
+        }),
+    }
+})
+
+type ConnectionStatusState = ConnectionStatus["state"]
+type ConnectionStatusSetState = Parameters<ConnectionStatus["setState"]>[0]
+
+function createConnectionStatus(state?: Partial<ConnectionStatusState>) {
+    const component = new ConnectionStatus({})
+    component.state = {
+        status: ConnectStatus.Connected,
+        latency: 88,
+        connectedSince: 1234,
+        showTooltip: false,
+        ...state,
+    }
+    component.setState = ((nextState: ConnectionStatusSetState) => {
+        Object.assign(
+            component.state,
+            typeof nextState === "function" ? nextState(component.state, component.props) : nextState
+        )
+    }) as typeof component.setState
+    return component
+}
+
+describe("ConnectionStatus measureLatency", () => {
+    beforeEach(() => {
+        hoisted.sdk.connectManager.status = ConnectStatus.Connected
+        hoisted.apiFetch.mockReset()
+    })
+
+    it("keeps websocket status when health check fails while SDK is still connected", async () => {
+        hoisted.apiFetch.mockRejectedValue(new Error("health unavailable"))
+        const component = createConnectionStatus()
+
+        await component.measureLatency()
+
+        expect(hoisted.apiFetch).toHaveBeenCalledWith("https://api.example.com/health", {
+            method: "GET",
+            cache: "no-cache",
+        })
+        expect(component.state.status).toBe(ConnectStatus.Connected)
+        expect(component.state.latency).toBeNull()
+        expect(component.state.connectedSince).toBe(1234)
+    })
+
+    it("marks disconnected when health check fails and SDK is not connected", async () => {
+        hoisted.sdk.connectManager.status = ConnectStatus.Disconnect
+        hoisted.apiFetch.mockRejectedValue(new Error("network down"))
+        const component = createConnectionStatus()
+
+        await component.measureLatency()
+
+        expect(component.state.status).toBe(ConnectStatus.Disconnect)
+        expect(component.state.latency).toBeNull()
+        expect(component.state.connectedSince).toBeNull()
+    })
+})
--- a/packages/dmworkbase/src/Components/ConnectionStatus/index.tsx
+++ b/packages/dmworkbase/src/Components/ConnectionStatus/index.tsx
@@ -98,15 +98,19 @@ export default class ConnectionStatus extends Component<ConnectionStatusProps, C
                     this.connectedTime = Date.now()
                     nextState.connectedSince = this.connectedTime
                 }
-                this.setState(nextState as any)
+                this.setState(nextState)
             }
         } catch {
-            this.connectedTime = 0
-            this.setState({
-                status: ConnectStatus.Disconnect,
-                latency: null,
-                connectedSince: null,
-            })
+            if (!isImConnected(WKSDK.shared())) {
+                this.connectedTime = 0
+                this.setState({
+                    status: ConnectStatus.Disconnect,
+                    latency: null,
+                    connectedSince: null,
+                })
+            } else {
+                this.setState({ latency: null })
+            }
         }
     }
 


OctoBoooot (approved):
## Review: #743 — fix ConnectionStatus health-check overriding WebSocket status (follow-up to #742)

**Verdict:** Approve. This is the correct, targeted fix for the major I flagged on #742 — the HTTP `/health` catch no longer overrides the authoritative WebSocket-derived connection status — with the exact regression coverage that was missing. Risk tier: normal. CI: all code checks green (Build SUCCESS, secret-scan, osv-scan, dependency-review, pr-title-lint); `code-review` (human gate) pending.

### Resolved — health-check failure no longer falsely shows "Disconnected" (`ConnectionStatus/index.tsx:103-113`)
The catch is now gated on the WebSocket state, exactly as the fix direction required:
- **WS still connected** (`isImConnected` true) → `setState({ latency: null })` only. Status untouched, so a transient HTTP `/health` failure (503 / timeout / rate-limit) leaves the indicator "Connected" — the HTTP probe informs *latency*, not *status*.
- **WS genuinely down** (`!isImConnected`) → `status: Disconnect` + clear latency/connectedSince/connectedTime, as before.

This restores the invariant the success path already enforced (`:92` gates Connected on `isImConnected`): the WebSocket is the single source of truth for connection status; the REST health probe only measures latency. It also resolves the 5s flicker at its root (the WS-up branch no longer changes `status`) **without** stopping the ping timer — so recovery detection is preserved, which was the right approach. The unnecessary `as any` cast (a #742 nit) is dropped too.

### Tests — cover both branches, the exact gap
`ConnectionStatus.test.ts` (new, +109) drives real `apiFetch` rejection under both SDK states:
- *"keeps websocket status when health check fails while SDK is still connected"* — mocks `apiFetch` reject + SDK Connected, asserts `status === Connected` (untouched), `latency === null`, `connectedSince` preserved. This is precisely the #742 failing scenario, now proven fixed.
- *"marks disconnected when health check fails and SDK is not connected"* — reject + SDK Disconnect, asserts `status === Disconnect`, latency/connectedSince null. The genuine-disconnect path still works.

### Praise
The fix targets the root cause (status source conflation) rather than the symptom (flicker), gates on the same `isImConnected` invariant the success path uses, and adds the two-branch regression test that pins the behavior — a clean, complete follow-up to the #742 CR.

### Net
Correctly resolves the #742 major (byte-verified: catch gated on `!isImConnected`, HTTP probe informs latency only, WS remains status source-of-truth), with real two-branch regression coverage and Build green. Approving. Credit menghao-webtest; credit yujiawei for the original catch. Gate stays on the required `code-review` check.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #743 (octo-web)

## Summary

This PR fixes the `ConnectionStatus` component so that an HTTP `/health` probe failure no longer unconditionally overrides the WKSDK/WebSocket connection status to "Disconnected." The fix correctly treats the WKSDK as the source of truth for connection state and the health probe as a latency signal only. When the SDK is still connected but the health endpoint fails, only `latency` is cleared — `status` and `connectedSince` are preserved. When the SDK itself is disconnected, the old full-disconnect behavior is retained. A well-structured test file is added covering both branches of the new conditional.

## Verification

- ✅ **Data-flow trace** — `isImConnected(WKSDK.shared())` at `index.tsx:105` resolves to `sdk.connectManager.status === ConnectStatus.Connected` via `getImConnectStatus()` in `connectStatus.ts:66`. This reads the SDK's internal WebSocket state, independent of the HTTP health endpoint — correct source of truth.
- ✅ **Try-block symmetry** — the success path at `index.tsx:92` already gates on `isImConnected()` before setting `status: Connected`. The catch block now mirrors this gate, creating consistent bidirectional logic: status is only modified when the SDK state agrees.
- ✅ **Type safety improvement** — `this.setState(nextState as any)` at `:101` changed to `this.setState(nextState)`. Since `nextState: Partial<ConnectionStatusState>` matches the `Component.setState` parameter type, the `as any` was unnecessary. TypeScript now enforces the shape statically.
- ✅ **Test coverage** — two test cases cover both branches of the new catch-block conditional. Test 1 (SDK connected, health fails) asserts `status=Connected`, `latency=null`, `connectedSince` preserved. Test 2 (SDK disconnected, health fails) asserts `status=Disconnect`, all fields cleared. The mock correctly intercepts `wukongimjssdk` and `apiFetch`.
- ✅ **`connectedSince` preservation** — when SDK is connected but health fails, retaining `connectedSince` is correct: the WebSocket session is still alive, only the HTTP probe is unavailable. The user still sees "Connected" with no latency value, which accurately reflects the situation.

Static analysis only at head `4acaebf1e`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. No correctness, security, build, or functional concerns. The change is minimal, well-scoped, and the logic is sound.

### Data-Flow Trace

| Consumed data | Upstream source | Runtime flow verified |
|---|---|---|
| `isImConnected()` result (catch :105) | `connectStatus.ts:66` → `sdk.connectManager.status` | ✅ Reads WKSDK WebSocket state directly; not influenced by HTTP probe result |
| `nextState` shape (try :101) | Constructed at :93-99 as `Partial<ConnectionStatusState>` | ✅ Type-safe after `as any` removal; always `{status, latency, connectedSince?}` |
| `connectedSince` (catch-else :113) | Not modified; retains `this.state.connectedSince` | ✅ Preserved from last SDK connect event at :47/:61 |

### Blindspot Checklist

- **C1 (dual-path parity)** — `componentDidMount` `statusListener` (:40-55) handles Connected (start ping) and non-Connected (stop ping, clear latency). `measureLatency` try-block gates on `isImConnected()` for Connected; catch-block now gates on `!isImConnected()` for Disconnect. Both directions are symmetric. Clear.
- **C2 (control-flow ordering)** — `measureLatency()` is called from `startPing()` (:72) and the 5s `setInterval` (:73). No nested or reentrant call paths. `componentWillUnmount` stops the timer. Clear.
- **C3–C6** — N/A: no auth, permissions, build, or governance changes.

### Cross-Round Blocker Review

N/A — first review pass.

## Verdict: APPROVED

Clean, minimal fix with correct separation of concerns (health probe = latency signal, WKSDK = connection truth). Tests cover both branches of the new conditional. No blocking issues found.

yujiawei (approved):
# Code Review — PR #743 (octo-web)

**Verdict: APPROVE** — the core fix is correct and well-tested; the findings below are non-blocking (P2).

## Summary of the change
`measureLatency()` previously forced `status = Disconnect` on any `/health` failure. This PR makes WKSDK/WebSocket the source of truth: on a failed probe it only clears `latency` when the SDK is still connected, and only marks `Disconnect` when the SDK is genuinely down. A new test covers both branches.

## Spec compliance
- Implements the stated intent (health probe informs latency; WKSDK owns connection status). ✅
- No out-of-scope additions, no missing deliverables, no contradictory implementation. ✅
- Tests added for both branches (SDK-connected and SDK-disconnected on probe failure). ✅

## Correctness
- The `isImConnected(WKSDK.shared())` guard reads the live SDK status at catch time — correct source of truth (`index.tsx:104`).
- `connectedTime` is reset only when actually transitioning to `Disconnect`, preserving `connectedSince` when the connection is retained (`index.tsx:105-113`). Good.
- Genuine WebSocket disconnects still flow through the independent `statusListener` in `componentDidMount`, so removing the status mutation here does not weaken real disconnect detection. No regression.
- Dropping the `as any` cast at `index.tsx:101` is a safe cleanup — `nextState` is already typed `Partial<ConnectionStatusState>`, exactly what `setState` accepts.

## Findings (non-blocking)

### P2 — Visible label can still read "disconnected" while status is Connected
`index.tsx:159-161`:
```ts
const labelText = connected && latency !== null
    ? `${latency}ms`
    : connecting ? t("...connectingDots") : t("...disconnected")
```
When the probe fails but the SDK stays connected, the new state is `status=Connected, latency=null`. In `render()` this yields:
- signal bars: `getSignalBars(null, true)` → **2 bars**, colored green (`barColor` falls to `#22c55e` because `latency === null`);
- **label text: "disconnected"** (the ternary keys off `latency !== null`, not `connected`);
- tooltip (on hover): "connected".

So during a sustained `/health` outage with WS up, the badge shows *green bars + "disconnected" text + "connected" tooltip* — an internally contradictory display, and the visible label still says "disconnected", which is exactly the symptom this PR set out to remove. The status-machine fix (no false reconnect, correct tooltip, green bars) is the substantive win and is delivered; this residual label behavior is partly pre-existing (a fresh mount before the first successful ping already renders "disconnected" with `latency=null`). Consider keying the label/`labelText` off `connected` rather than `latency !== null`, e.g. render a distinct degraded state (`connected && latency === null` → show a placeholder like "—" or "connected" instead of "disconnected"). Non-blocking, but recommended as a short follow-up so the display fully matches the PR's stated goal.

### P2 — `setState` in `catch` without a mounted guard
`index.tsx:103-113`: `measureLatency()` awaits `apiFetch` and then calls `setState` in both `try` and `catch`. `componentWillUnmount` clears the interval via `stopPing()`, but an already in-flight probe is not cancelled, so if the component unmounts mid-request the resolved promise will `setState` on an unmounted component (React 17 logs a warning). Pre-existing (both branches already did this) and not worsened by the PR; worth a guard (an `isMounted` flag or `AbortController`) if you touch this again.

### P2 — Test drives the class directly instead of rendering it
`__tests__/ConnectionStatus.test.ts:60-74`: the test `new`s the class and stubs `setState` with a synchronous `Object.assign`, bypassing the React reconciler and lifecycle. This is fine for asserting the pure branch logic of `measureLatency`, but it won't catch lifecycle/reconciliation issues (including the unmounted-setState case above). Consider `@testing-library/react` for behavior-level coverage. Acceptable as-is for a focused unit test.

## Verification performed
- `pnpm --dir packages/dmworkbase test ConnectionStatus.test.ts` → **2/2 passed**.
- Reviewed against merge-base `3d747b43e` (#742), of which this is the follow-up.
- Confirmed no other consumers depend on the old "probe failure ⇒ Disconnect" behavior (`ConnectionStatus` is consumed via `NavRail/NavSignalBadge` only).

## Coverage note
- Not exercised at runtime: actual React render output for the `status=Connected, latency=null` combination (reasoned statically from `render()` above) and the unmounted-during-probe race. Repo-wide `tsc --noEmit` reports only pre-existing node_modules / react-types-resolution noise unrelated to this diff; the PR's own CI runs `vite build`, which is the authoritative build gate.

OctoBoooot:
## Corroborating yujiawei's 🟡#1 — the label shows "disconnected" in the WS-connected-but-health-failed state this fix newly exposes (byte-verified, non-blocking)

Verified yujiawei's first 🟡 against `ConnectionStatus/index.tsx` @ `4acaebf1e847` — it's a real display inconsistency, and notably it's newly *reachable* because of this fix. In the state #743 introduces (WS still connected, health probe failed → `status=Connected`, `latency=null`):

- `connected = true` → **bars green** (`barColor="#22c55e"`, `getSignalBars(null, true)=2`) and the **tooltip** renders `connected ? "connected"` → "connected".
- but `labelText = connected && latency !== null ? \`${latency}ms\` : (connecting ? … : t("disconnected"))` — with `latency===null` and not connecting, this falls through to **"disconnected"** (`:159-161`).

So the component simultaneously shows **green signal bars + a "disconnected" text label + a "connected" tooltip** — internally contradictory. Before #743 this `connected && latency===null` state was transient (only briefly at initial connect, before the first ping); #743 makes it a **persistent** state (WS up but `/health` failing), so the wrong label now sticks.

Fix is one line — the label's fallback shouldn't assume "not `${ms}` ⇒ disconnected" when the status is actually Connected:
```
const labelText = connected
  ? (latency !== null ? `${latency}ms` : t("base.connectionStatus.connected"))
  : connecting ? t("base.connectionStatus.connectingDots") : t("base.connectionStatus.disconnected")
```
(or show `--ms` / a connected-without-latency string) so a connected-but-latency-unknown state reads as connected, consistent with the green bars and tooltip.

**Severity: minor, non-blocking — my APPROVE stands.** It's label-text only (bars/color/tooltip are already correct), doesn't affect the core fix or functionality, and #743 correctly resolves the #742 major. But it's a real inconsistency this fix newly exposes, so worth folding the label fix into this PR or a fast-follow. yujiawei's other two 🟡 (setState-after-unmount guard; test instantiates the class rather than rendering) are fair, pre-existing/test-style, and also non-blocking. Credit yujiawei.

PR #744, opened 2026-07-14
feat(revoke): add re-edit button to restore recalled text message

xiaokenbot:
## Problem

After recalling a message, users have no way to restore the original text to the input box and must retype the entire message from scratch.

## Solution

Add a **重新编辑 / Re-edit** inline button next to the revoke tip for self-recalled text messages. Clicking it calls `ConversationContext.insertText()` to restore the original content to the message input and focuses the editor.

### `Messages/Revoke/index.tsx`
- `canReEdit()`: shows button only when `revoker === sender === self` AND `contentType === MessageContentType.text`
- `handleReEdit()`: reads `message.content.text`, calls `context.insertText()` — no new state, no new API calls
- Render wraps in `.wk-revoke-row`; button conditionally rendered

### `Messages/Revoke/index.css`
- `.wk-revoke-row`: inline-flex, centers tip + button on same line
- `.wk-revoke-reedit-btn`: ghost button, `--wk-brand-primary` underline text

### i18n
- `zh-CN`: `revoke.reEdit` = "重新编辑"
- `en-US`: `revoke.reEdit` = "Re-edit"

### Tests
5 cases in `Messages/Revoke/__tests__/RevokeCell.test.tsx`:
- ✅ shows button for self-recalled text message
- ✅ hides for other-revoker
- ✅ hides for non-text content type (image)
- ✅ hides for admin-revoke (revoker ≠ sender)
- ✅ revoke tip always present

## Scope

Only text messages support re-edit (consistent with `WKApp.supportEdit = [MessageContentType.text]`). Non-text recalled messages (images, files, etc.) show no button.

### DIFF
--- a/packages/dmworkbase/src/Messages/Revoke/index.tsx
+++ b/packages/dmworkbase/src/Messages/Revoke/index.tsx
@@ -1,4 +1,4 @@
-import { Channel, ChannelInfo, ChannelTypePerson, WKSDK } from "wukongimjssdk"
+import { Channel, ChannelInfo, ChannelTypePerson, MessageContentType, MessageText, WKSDK } from "wukongimjssdk"
 import { MessageCell } from '../MessageCell'
 import { MessageWrap } from '../../Service/Model'
 import WKApp from '../../App'
@@ -64,9 +64,48 @@ export class RevokeCell extends MessageCell {
         }
     }
 
+    /**
+     * 判断是否展示「重新编辑」按钮：
+     * - 必须是自己撤回的（revoker === 自己 uid）
+     * - 且原始消息为文本类型
+     */
+    private canReEdit(): boolean {
+        const { message } = this.props
+        return (
+            message.revoker === WKApp.loginInfo.uid &&
+            message.fromUID === WKApp.loginInfo.uid &&
+            message.contentType === MessageContentType.text
+        )
+    }
+
+    /**
+     * 点击「重新编辑」：将原文夹回输入框
+     */
+    private handleReEdit = () => {
+        const { message } = this.props
+        const conversationContext = (this.props as any).context
+        if (!conversationContext?.insertText) return
+        const textContent = message.content as MessageText
+        const originalText = textContent?.text ?? ''
+        if (!originalText) return
+        conversationContext.insertText(originalText)
+    }
+
     render() {
         const { message } = this.props
         this.context.locale
-        return <div className="wk-message-system">{RevokeCell.tip(message)}</div>
+        return (
+            <div className="wk-revoke-row">
+                <span className="wk-message-system">{RevokeCell.tip(message)}</span>
+                {this.canReEdit() && (
+                    <button
+                        className="wk-revoke-reedit-btn"
+                        onClick={this.handleReEdit}
+                    >
+                        {this.context.t("base.revoke.reEdit")}
+                    </button>
+                )}
+            </div>
+        )
     }
 }
--- a/packages/dmworkbase/src/Messages/Revoke/__tests__/RevokeCell.test.tsx
+++ b/packages/dmworkbase/src/Messages/Revoke/__tests__/RevokeCell.test.tsx
@@ -1,0 +1,107 @@
+// @vitest-environment jsdom
+import React from "react"
+import { renderToStaticMarkup } from "react-dom/server"
+import { describe, it, expect, vi } from "vitest"
+
+// ── Minimal SDK stubs ─────────────────────────────────────────────────────────
+vi.mock("wukongimjssdk", () => {
+    const ChannelTypePerson = 1
+    const MessageContentType = { text: 1, image: 2 }
+    const WKSDK = {
+        shared: () => ({
+            channelManager: {
+                addListener: vi.fn(),
+                removeListener: vi.fn(),
+                getChannelInfo: vi.fn(() => null),
+                fetchChannelInfo: vi.fn(),
+            },
+        }),
+    }
+    class Channel {
+        channelID: string; channelType: number
+        constructor(id: string, type: number) { this.channelID = id; this.channelType = type }
+    }
+    return { ChannelTypePerson, MessageContentType, WKSDK, Channel }
+})
+
+vi.mock("../../App", () => ({
+    default: { loginInfo: { uid: "user-self" } },
+}))
+
+vi.mock("../../i18n", () => ({
+    I18nContext: React.createContext({
+        locale: "zh-CN",
+        t: (key: string) => {
+            const map: Record<string, string> = {
+                "base.revoke.revokedMessage": "你撤回了一条消息",
+                "base.revoke.you": "你",
+                "base.revoke.reEdit": "重新编辑",
+            }
+            return map[key] ?? key
+        },
+    }),
+    t: (key: string) => {
+        const map: Record<string, string> = {
+            "base.revoke.revokedMessage": "你撤回了一条消息",
+            "base.revoke.you": "你",
+        }
+        return map[key] ?? key
+    },
+}))
+
+import { RevokeCell } from "../index"
+
+function makeMessage(overrides: Record<string, any> = {}) {
+    return {
+        revoker: "user-self",
+        fromUID: "user-self",
+        contentType: 1, // MessageContentType.text
+        content: { text: "这是原始消息内容", contentType: 1 },
+        from: null,
+        remoteExtra: {},
+        ...overrides,
+    }
+}
+
+function renderCell(message: any, contextOverrides: any = {}) {
+    const ctx = {
+        insertText: vi.fn(),
+        ...contextOverrides,
+    }
+    // RevokeCell is a class component; render via renderToStaticMarkup
+    return renderToStaticMarkup(
+        React.createElement(RevokeCell as any, {
+            message,
+            context: ctx,
+        })
+    )
+}
+
+describe("RevokeCell — re-edit button", () => {
+    it("shows re-edit button when self revoked own text message", () => {
+        const html = renderCell(makeMessage())
+        expect(html).toContain("重新编辑")
+        expect(html).toContain("wk-revoke-reedit-btn")
+    })
+
+    it("does NOT show re-edit button when someone else revoked the message", () => {
+        const html = renderCell(makeMessage({ revoker: "other-user" }))
+        expect(html).not.toContain("重新编辑")
+        expect(html).not.toContain("wk-revoke-reedit-btn")
+    })
+
+    it("does NOT show re-edit button for non-text messages (e.g. image)", () => {
+        const html = renderCell(makeMessage({ contentType: 2 })) // image
+        expect(html).not.toContain("重新编辑")
+    })
+
+    it("does NOT show re-edit button when revoker is self but sender is someone else (admin revoke)", () => {
+        const html = renderCell(makeMessage({ fromUID: "other-user" }))
+        expect(html).not.toContain("重新编辑")
+    })
+
+    it("always shows the revoke tip text", () => {
+        const html = renderCell(makeMessage())
+        expect(html).toContain("你撤回了一条消息")
+    })
+})
--- a/packages/dmworkbase/src/Messages/Revoke/index.css
+++ b/packages/dmworkbase/src/Messages/Revoke/index.css
@@ -1,1 +1,23 @@
+.wk-revoke-row {
+  display: inline-flex;
+  align-items: center;
+  gap: 8px;
+  justify-content: center;
+}
 
+.wk-revoke-reedit-btn {
+  display: inline;
+  background: none;
+  border: none;
+  padding: 0;
+  margin: 0;
+  font-size: 13px;
+  color: var(--wk-brand-primary, #7F3BF5);
+  cursor: pointer;
+  text-decoration: underline;
+  line-height: inherit;
+}
+
+.wk-revoke-reedit-btn:hover {
+  opacity: 0.75;
+}
--- a/packages/dmworkbase/src/i18n/locales/en-US.json
+++ b/packages/dmworkbase/src/i18n/locales/en-US.json
@@ -1013,6 +1013,7 @@
   "revoke.revokedMemberMessageByYou": "You recalled a message from \"{{member}}\"",
   "revoke.revokedMessage": "{{name}} recalled a message",
   "revoke.you": "You",
+  "revoke.reEdit": "Re-edit",
   "video.digest": "[Short video]",
   "video.uploadFailedRetry": "Upload failed. Click to retry",
   "aiMessageCard.assistant": "AI assistant",
--- a/packages/dmworkbase/src/i18n/locales/zh-CN.json
+++ b/packages/dmworkbase/src/i18n/locales/zh-CN.json
@@ -1013,6 +1013,7 @@
   "revoke.revokedMemberMessageByYou": "你撤回了成员“{{member}}”的一条消息",
   "revoke.revokedMessage": "{{name}}撤回了一条消息",
   "revoke.you": "你",
+  "revoke.reEdit": "重新编辑",
   "video.digest": "[小视频]",
   "video.uploadFailedRetry": "上传失败，点击重试",
   "aiMessageCard.assistant": "AI助手",
--- a/packages/dmworkbase/src/ui/message/Bubble/index.css
+++ b/packages/dmworkbase/src/ui/message/Bubble/index.css
@@ -47,3 +47,22 @@
 .wk-msg-bubble--last {
   border-radius: var(--wk-r-xs) var(--wk-r-lg) var(--wk-r-lg) var(--wk-r-lg);
 }
+
+/**
+ * 发送方气泡：圆角靠右（右侧靠近头像一侧为直角）
+ */
+
+/* send first: 右下直 */
+.wk-msg-bubble--send.wk-msg-bubble--first {
+  border-radius: var(--wk-r-lg) var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-lg);
+}
+
+/* send middle: 左圆右直 */
+.wk-msg-bubble--send.wk-msg-bubble--middle {
+  border-radius: var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-xs) var(--wk-r-lg);
+}
+
+/* send last: 右上直 */
+.wk-msg-bubble--send.wk-msg-bubble--last {
+  border-radius: var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-lg) var(--wk-r-lg);
+}
--- a/packages/dmworkbase/src/ui/message/Bubble/index.css
+++ b/packages/dmworkbase/src/ui/message/Bubble/index.css
@@ -52,6 +52,11 @@
  * 发送方气泡：圆角靠右（右侧靠近头像一侧为直角）
  */
 
+/* send single: 全圆角（对称气泡不需要缺角） */
+.wk-msg-bubble--send.wk-msg-bubble--single {
+  border-radius: var(--wk-r-lg);
+}
+
 /* send first: 右下直 */
 .wk-msg-bubble--send.wk-msg-bubble--first {
   border-radius: var(--wk-r-lg) var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-lg);
--- a/packages/dmworkbase/src/ui/message/Bubble/index.css
+++ b/packages/dmworkbase/src/ui/message/Bubble/index.css
@@ -47,27 +47,3 @@
 .wk-msg-bubble--last {
   border-radius: var(--wk-r-xs) var(--wk-r-lg) var(--wk-r-lg) var(--wk-r-lg);
 }
-
-/**
- * 发送方气泡：圆角靠右（右侧靠近头像一侧为直角）
- */
-
-/* send single: 全圆角（对称气泡不需要缺角） */
-.wk-msg-bubble--send.wk-msg-bubble--single {
-  border-radius: var(--wk-r-lg);
-}
-
-/* send first: 右下直 */
-.wk-msg-bubble--send.wk-msg-bubble--first {
-  border-radius: var(--wk-r-lg) var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-lg);
-}
-
-/* send middle: 左圆右直 */
-.wk-msg-bubble--send.wk-msg-bubble--middle {
-  border-radius: var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-xs) var(--wk-r-lg);
-}
-
-/* send last: 右上直 */
-.wk-msg-bubble--send.wk-msg-bubble--last {
-  border-radius: var(--wk-r-lg) var(--wk-r-xs) var(--wk-r-lg) var(--wk-r-lg);
-}
--- a/packages/dmworkbase/src/ui/message/MessageRow/index.css
+++ b/packages/dmworkbase/src/ui/message/MessageRow/index.css
@@ -48,10 +48,36 @@
 }
 
 /**
- * 发送方消息（和接收方一样左对齐，只是可能没有头像）
+ * 发送方消息：flex-direction 反转，头像靠右，气泡靠右对齐
  */
 .wk-msg-row--send {
-  /* 发送方和接收方布局相同，都是左对齐 */
+  flex-direction: row-reverse;
+}
+
+/**
+ * 发送方：内容区文本靠右
+ */
+.wk-msg-row--send .wk-msg-row-content {
+  align-items: flex-end;
+}
+
+/**
+ * 发送方：隐藏自己的名字（自己知道自己是谁）
+ */
+.wk-msg-row--send .wk-msg-row-header {
+  flex-direction: row-reverse;
+}
+
+.wk-msg-row--send .wk-msg-row-sender {
+  display: none;
+}
+
+/**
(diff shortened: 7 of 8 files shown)


Jerry-Xin (approved):
**APPROVE** — clean, well-scoped re-edit-on-recall feature. The security crux (whose recalled content is restorable) is correctly sender-gated, and the restore path reuses the existing composer insert. No blocking issues.

### Security review (the crux)

**IDOR / cross-user leak — clean.** The "Re-edit" button is gated by `canReEdit()`:
`revoker === self.uid && fromUID === self.uid && contentType === text` (`Messages/Revoke/index.tsx`). The restored text is read from the **local** `message.content.text` of a message the current user both sent and recalled — there is **no new server fetch**, so no new authz surface is introduced. A recipient viewing someone else's recalled message fails `fromUID === self.uid`; an admin/other-party revoke fails `revoker === self.uid`. Neither shows the button, and neither triggers any content retrieval. The recall in this system is a UI-level flag (`remoteExtra.revoke`/`revoker`) — the original text already lives in each client that received the message; this PR does not widen that exposure and does not surface it to recipients.

**XSS — low risk, consistent with existing behavior.** Restore path: `handleReEdit` → `context.insertText(originalText)` → `editor.commands.insertContent(text)` on the Tiptap 3.x editor (`Components/MessageInput/index.tsx`). A string argument is parsed by ProseMirror's DOM parser into **schema-defined nodes only** — scripts and arbitrary attributes are not preserved or executed. This is the same insert path already used by `reply()` and the emoji toolbar, and the content is the user's own outgoing text. No new sink.

**Permission / state gating.** The gate also restricts to text messages, so image/other content types get no button. Handler no-ops safely when the composer context or text is absent (`if (!ctx?.insertText) return; if (!originalText) return;`).

**State / cleanup / race.** No new listeners or timers introduced. The existing `channelInfoListener` retains its matching `removeListener` in `componentWillUnmount`. `handleReEdit` is a single synchronous insert — no stale-closure or double-insert path.

**i18n — symmetric.** `revoke.reEdit` added to both locale files present in the repo (`en-US.json` = "Re-edit", `zh-CN.json` = "重新编辑"), both real values.

**Tests.** `RevokeCell.test.tsx` covers the gate matrix well: self-revoked own text shows the button; other-revoker, non-text, and admin-revoke (self revoker but other sender) all correctly hide it.

**CI.** pr-title-lint / label / add-to-project all green; no undeclared imports (`MessageContentType`, `MessageText` come from the already-declared `wukongimjssdk`).

### 💬 Non-blocking notes

1. **Append vs replace into composer.** `insertText` uses `editor.commands.insertContent`, which **appends** to any existing draft. If the user already has unsent text in the composer, clicking Re-edit concatenates the recalled text onto it rather than replacing. `restoreDraft` (setContent) exists for the replace semantics — consider whether replace is the more intuitive behavior here, or at least confirm append is intended.

2. **Bundled layout change.** The `MessageRow` CSS/TSX changes (send-side body-wrap mirroring, checkbox excluded from reversal) are unrelated to the revoke feature. They're structurally sound (wrapper div balanced), but bundling an unrelated layout refactor into a feature PR makes review/rollback coarser.

### ✅ Highlights
- Correct sender-gating with defense-in-depth (`revoker` + `fromUID` + `contentType`).
- No new network call — restores from local state only.
- Solid test coverage of the visibility gate including the admin-revoke edge case.

OctoBoooot (approved):
## Review: #744 — re-edit button to restore recalled text message

**Verdict:** Approve (code) — with a merge-ordering note: this PR is **stacked on the still-open #731**, carrying its MessageRow changes byte-identically. The re-edit feature itself is clean, correctly authorized, and well-tested. Risk tier: normal. CI: the checks that ran are green (`code-review` SUCCESS, `pr-title-lint`, `label`); the heavier jobs (Build/scans) didn't run for this path scope.

### Feature — correct and well-guarded (`Messages/Revoke/index.tsx`)
The re-edit button restores a self-recalled text message's content to the input box. Authorization is a correct triple gate in `canReEdit()`:
- `message.revoker === WKApp.loginInfo.uid` (I recalled it),
- `message.fromUID === WKApp.loginInfo.uid` (I was the sender — so an admin-recall where revoker≠sender is excluded),
- `message.contentType === MessageContentType.text` (text only, matching `WKApp.supportEdit`).

So the button only appears for *my own* self-recalled *text* messages — no path to re-edit someone else's content or a non-text/admin-recalled message. `handleReEdit` null-guards both `context.insertText` and the extracted `originalText` before inserting; no new state, no new API calls. Clean.

### Tests — cover the authorization matrix
`RevokeCell.test.tsx` (+107) asserts exactly the guard boundaries: shows for self-revoked own text; hidden for other-revoker; hidden for non-text (image); hidden for admin-revoke (revoker=self, sender=other); revoke tip always present. Matches `canReEdit()` precisely.

### Merge-ordering note (not a defect) — stacked on #731
The `MessageRow/index.tsx` (+3) and `index.css` (+44/-3) hunks in this PR are the right-align-self-messages feature from **#731**, and they are **byte-identical to #731's head (`3929aeca`)** — this PR is branched on top of #731's branch (same author), not an independent re-implementation. #731 is still OPEN and `.wk-msg-row-body-wrap` is not yet on main. Implications for whoever merges:
- If #731 merges first, #744's MessageRow hunk becomes a no-op against the new base (clean).
- If #744 merges first, it carries #731's MessageRow feature in with it — merging that feature via #744, which may not be intended, and would leave #731 partly redundant.
Recommend merging #731 first (or rebasing #744 onto main once #731 lands) so the re-edit change lands as just its own delta. This is a maintainer/author coordination call, not a code issue — the carried hunk is the already-reviewed #731 content, unchanged.

### Net
The re-edit feature is correct (triple-gated authorization, text-only, no new API/state), well-tested, and CI-clean. The only consideration is the stack-on-#731 ordering (byte-identical carry of #731's MessageRow changes) — merge #731 first or rebase to avoid pulling #731's feature in sideways. Approving the code. Credit xiaokenbot. `mergeStateStatus` is BLOCKED pending the required review/branch-protection gate.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #744 (octo-web)

## Summary

This PR adds a "Re-edit" button to the revoke/recall system message for text messages the current user both sent and recalled themselves. Clicking the button restores the original message text into the input box via `ConversationContext.insertText`. The PR also restructures `MessageRow`'s DOM by introducing a `wk-msg-row-body-wrap` wrapper around avatar + content, isolating the flex-direction reversal for sent messages from the checkbox.

The feature logic is correct: the three-gate `canReEdit()` check (self-revoked, self-sent, text-type) properly scopes the button, and the data flow from `message.content` through `insertText` is traceable end-to-end — the revoke handler in `ConversationVM` (`vm.ts:911-921`) sets `revoke=true` and `revoker=...` without clearing the original content, so `message.content.text` remains available. Two non-blocking P2 items below.

## Verification

- ✅ **Data flow: message.content preserved on revoke** — `vm.ts:913-914` only sets `existMessage.revoke = true` and `existMessage.revoker`; original `content` (including `MessageText.text`) is untouched. `handleReEdit` can read it.
- ✅ **insertText wiring** — `RevokeCell` is rendered with `context={this}` (Conversation instance) at `Conversation/index.tsx:1649` and `:1717`. `Conversation.insertText` at `:1022` implements the `ConversationContext` interface, with a `_pendingInsertText` fallback when the input context isn't ready.
- ✅ **canReEdit gating** — three conditions (`revoker === self`, `fromUID === self`, `contentType === text`) correctly prevent the button from showing for admin revokes, others' messages, and non-text content types.
- ✅ **MessageRow DOM restructure safe** — no CSS in the repo uses direct-child selectors (`>`) targeting `.wk-msg-row-avatar` or `.wk-msg-row-content` from `.wk-msg-row`. All existing CSS uses descendant selectors. The `wk-msg-row > *` z-index rule correctly applies to the new wrapper. Existing MessageRow tests use `querySelector` (descendant search) — unaffected.
- ✅ **i18n parity** — `revoke.reEdit` key added in both `en-US` ("Re-edit") and `zh-CN` ("重新编辑"), placed correctly between `revoke.you` and `video.digest` alphabetically.

## Findings

### P2 — Unnecessary `as any` type escape (`packages/dmworkbase/src/Messages/Revoke/index.tsx:86`)

`handleReEdit` reads `const conversationContext = (this.props as any).context`, but `MessageBaseProps` at `Messages/Base/index.tsx:51` already declares `context: ConversationContext`. The `as any` cast bypasses TypeScript's type checking — if `ConversationContext` ever gains or loses a method, this cast hides the compile error. Use `this.props.context` directly; the optional-chaining guard on `insertText` is still valid as defensive code.

### P2 — No test covers `handleReEdit` click behavior (`packages/dmworkbase/src/Messages/Revoke/__tests__/RevokeCell.test.tsx:80`)

The test suite uses `renderToStaticMarkup` which cannot fire DOM events. The `insertText: vi.fn()` mock at line 68 is created but never asserted — no test verifies that clicking the re-edit button actually calls `insertText` with the original message text. A `ReactDOM.render`-based test (matching the pattern in `MessageRow.test.tsx`) that dispatches a click on `.wk-revoke-reedit-btn` and asserts `expect(insertText).toHaveBeenCalledWith("这是原始消息内容")` would close this gap.

- `canReEdit` does not check `message.revoke` explicitly, but `RevokeCell` is only rendered when `message.revoke` is true (gated at `Conversation/index.tsx:1648` and `:1717`), so this is implicitly satisfied.
- `message.content as MessageText` cast is safe because `canReEdit` gates on `contentType === MessageContentType.text`, which guarantees the content is a `MessageText` instance.
- `.wk-message-system` changed from `<div>` to `<span>`, but CSS explicitly sets `display: block`, so rendering is preserved. The `:has(.wk-message-system)` selector in `Conversation/index.css:46` still matches.
- `.wk-msg-row-body-wrap` wrapper added unconditionally — for received messages, default `flex-direction: row` applies (no visual change); for sent messages, `row-reverse` correctly mirrors avatar+content while checkbox stays outside.
- i18n key access via `this.context.t("base.revoke.reEdit")` is correct — `RevokeCell` declares `static contextType = I18nContext`.

## Data-flow trace

|---|---|---|
| `message.content.text` | SDK stores original `MessageText` on send; revoke handler (`vm.ts:913`) does NOT clear it | ✅ Yes — `handleReEdit` reads it |
| `this.props.context.insertText` | `Conversation` class implements `ConversationContext.insertText` at `:1022` | ✅ Yes — passed as `context={this}` |
| `message.revoker` | Set by revoke CMD handler at `vm.ts:914` (`existMessage.revoker = message.fromUID`) | ✅ Yes — `canReEdit` reads it |
| `WKApp.loginInfo.uid` | Login state singleton | ✅ Yes — stable during session |

## Blind-spot checklist (C1–C6)

- **C1 (dual-path parity)**: N/A — single render path for revoke cell, no paired add/remove.
- **C2 (control-flow ordering)**: N/A — `handleReEdit` is a simple sequential flow, no nested reuse.
- **C3 (auth boundary)**: N/A — UI-only feature, no new endpoint or tool exposure.
- **C4 (auth lifecycle)**: N/A — no authorization state changes.
- **C5 (build/runtime)**: N/A — CSS/TSX changes only, no build artifact concerns.
- **C6 (governance/docs)**: N/A — no documentation or policy changes.

## Verdict: APPROVED

The feature is well-scoped, the data flow is sound, and the `canReEdit` gating correctly limits the button to self-revoked self-sent text messages. The two P2 items (unnecessary type escape, missing click-behavior test) are code-quality improvements that don't block merge.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Data flow from revoked message content → insertText is verified end-to-end. P2 items are non-blocking.

yujiawei (changes requested):
# Code Review — PR #744 (octo-web)

Reviewed at head SHA `0c0ec5b8c895bb57deb889034803a8790a087400` against `main` (merge-base `0470f620`).

## 1. Spec compliance

**Spec: ❌**

The PR description says this change *only* adds a "Re-edit" button to the recall tip (`Messages/Revoke/*`, i18n, tests). The diff does more than that:

- **Out-of-scope change (not described):** `ui/message/MessageRow/index.tsx` and `MessageRow/index.css` introduce a global layout change for **self-sent messages** — a new `.wk-msg-row-body-wrap` flex wrapper, `flex-direction: row-reverse` mirroring, right-aligned header/content, and hiding of the sender name / realname badge / space suffix for sent messages.
  - `MessageRow/index.tsx:189-190` and `:255` add the `wk-msg-row-body-wrap` wrapper.
  - `MessageRow/index.css:51-88` add `.wk-msg-row--send .wk-msg-row-body-wrap { flex-direction: row-reverse }`, `.wk-msg-row--send .wk-msg-row-sender { display: none }`, header/content right-alignment, etc.
  - These are net-new vs `main` (main does not contain `wk-msg-row-body-wrap`).

This is a visible behavioral change to how every sent message renders (right alignment + hidden sender identity), bundled into a PR whose title/description are about a re-edit button. A reviewer or merger reading the description would not know they are also shipping a message-layout overhaul, so it cannot be evaluated on its own merits here.

**Please either** split the `MessageRow` layout change into its own PR, **or** update this PR's description to document the layout change (rationale, affected surfaces, before/after) so it can be reviewed properly.

## 2. Code Quality

### P2 — Re-edit restores pre-edit text for edited-then-recalled messages
`Messages/Revoke/index.tsx:88` reads the original content:
```ts
const textContent = message.content as MessageText
const originalText = textContent?.text ?? ''
```
But for a message that was **edited** before being recalled, the edited text lives in `remoteExtra.contentEdit`, not `content`. The rest of the codebase accounts for this:
- `Service/Model.tsx:405-407` (`parseMention`) prefers `remoteExtra.contentEdit` when `remoteExtra.isEdit`.
- `Messages/Text/index.tsx:144-147` (`getRenderMessageText`) does the same.
- `Service/Convert.ts:336-345` populates `contentEdit` / `isEdit` from the `content_edit` wire field.

So "send → edit → recall → Re-edit" restores the **pre-edit** text, not what the user last had on screen. Consider mirroring the `isEdit ? contentEdit : content` pattern used elsewhere.

### P2 — Mentions / custom emoji lose their structure on re-edit
`handleReEdit` passes the plain `content.text` into `ConversationContext.insertText()`. That path (`Components/MessageInput/index.tsx:887-896`) calls `editor.commands.insertContent(text)` and, per its own comment, deliberately does **not** parse the `@[uid:label]` mention grammar (only `restoreDraft` does). The stored `content.text` is the flattened display form (`@name`, emoji tokens), so a recalled message that contained @mentions or custom emoji comes back as inert text: re-sending it will not actually mention anyone. If preserving mentions matters for this feature, `restoreDraft()` (which parses the mention grammar) is the closer primitive — though it replaces editor content rather than appending, so the choice needs thought. At minimum this limitation is worth a code comment.

### P2 — Unnecessary `as any` cast hides a typed prop
`Messages/Revoke/index.tsx:86`:
```ts
const conversationContext = (this.props as any).context
```
`context` is already declared on `MessageBaseCellProps` (`Messages/MessageCell.tsx:12-15`) and is passed as a real prop everywhere `RevokeCell` is rendered (`Components/Conversation/index.tsx:1649`, `:1717`, `:2094`). The cast defeats type checking for no reason — use `this.props.context` (typed `ConversationContext`), which also documents the `insertText` contract.

### P2 — Test coverage gaps
`Messages/Revoke/__tests__/RevokeCell.test.tsx` is a good start (render-gating via `renderToStaticMarkup`), but:
- It never exercises `handleReEdit` click behavior (that `insertText` is actually called with the right text) — it only checks that the button renders.
- No case for the edited-then-recalled text bug or the mention-loss behavior above.
- The SDK is fully mocked, so the render path is exercised in isolation only; consider one test asserting the click calls `context.insertText(originalText)`.

### Nit (pre-existing) — dead expression
`Messages/Revoke/index.tsx:96` has a standalone `this.context.locale` statement with no effect. It predates this PR, but since the file is being touched, either drop it or add a comment explaining it forces a locale-change re-render.

## 3. Overall verdict

**REQUEST_CHANGES**

Spec compliance fails (undescribed global `MessageRow` layout change bundled into a re-edit-button PR), which by itself blocks. On top of that, the re-edit feature has two real correctness gaps (edited-then-recalled restores the wrong text; mentions/emoji lose structure). None of these are data-loss or security issues, and the happy path (recall your own plain-text message → restore it) works.

## 4. Suggested direction

1. Split the `MessageRow` self-message layout change into its own PR, or fully document it in this PR's description so it can be reviewed.
2. In `handleReEdit`, use `remoteExtra.contentEdit` when `remoteExtra.isEdit` is set (match `Text`/`Model` behavior).
3. Decide how mentions should behave on re-edit (restore as real mentions via the parse path, or document that they degrade to plain text).
4. Replace `(this.props as any).context` with the typed `this.props.context`.
5. Add a click-behavior test asserting `insertText` is called; add the edit/mention edge cases.

## 5. Additional notes

- The existing `MessageRow` tests use regex/substring matching, so the extra wrapper `<div>` does not break them structurally — but I could not run the suite in this environment (no `node_modules`); please confirm CI is green.
- `canReEdit()` correctly gates on `revoker === sender === self` and `contentType === text`, consistent with `WKApp.supportEdit = [MessageContentType.text]` (`App.tsx:721`). Non-text recalled messages correctly show no button.

Jerry-Xin (changes requested):
Flipping my earlier approval to request-changes. @yujiawei's two structural-correctness findings are both real; I byte-verified them against the source at this head. Credit to yujiawei for catching these.

## 🔴 Blocking

### 1. Re-edit restores the PRE-EDIT text for edited-then-recalled messages
`handleReEdit` in `packages/dmworkbase/src/Messages/Revoke/index.tsx` reads the restore text directly from the raw content:

```ts
const textContent = message.content as MessageText
const originalText = textContent?.text ?? ''
```

But an edited message does NOT keep its latest text in `message.content`. The canonical helper `getEffectiveContent()` (`packages/dmworkbase/src/Components/Conversation/index.tsx`) documents the model:

> 取消息的有效内容：如果消息被编辑过，返回编辑后的 contentEdit；否则返回原始 content

```ts
const content =
  message.remoteExtra?.isEdit && message.remoteExtra?.contentEdit
    ? message.remoteExtra.contentEdit
    : message.content;
```

Every edit-aware read path (text render, reply digest, forward) uses `remoteExtra.isEdit ? remoteExtra.contentEdit : content` — e.g. `packages/dmworkbase/src/Messages/Text/index.tsx` (`getRenderMessageText`) and `Model.tsx parseMention`. `handleReEdit` bypasses this and always reads `message.content.text`, which for an edited message holds the STALE original. So: send "A" → edit to "B" → recall → re-edit restores "A", the wrong version.

Fix: read the effective content (`remoteExtra.isEdit && remoteExtra.contentEdit ? remoteExtra.contentEdit.text : message.content.text`), matching the existing edit-aware read paths.

### 2. @mentions lose their structured uid ref on restore (flattened to plain label text)
For a text message with mentions, the visible text lives in `content.text` (e.g. `@Alice hi`) while the routable binding lives in a SEPARATE field `content.mention.entities` (`[{uid, offset, length}]`) — see `parseMention`/`parseMentionWithEntities` in `packages/dmworkbase/src/Service/Model.tsx`, which reconstruct mention Parts by combining `text` + `mention.entities`.

`handleReEdit` reads only `content.text` and passes it to `context.insertText`. `insertText` (`packages/dmworkbase/src/Components/MessageInput/index.tsx`) inserts as-is and explicitly does NOT parse mention markup:

```ts
// 原样追加，不解析 @[uid:label]（与 main 行为一致）
// mention 格式的反序列化仅在 restoreDraft 中处理
editor.commands.insertContent(text);
```

So the mention's uid ref (`content.mention.entities`) is dropped entirely — the restored `@Alice` is literal text, not a mention node. On resend it no longer notifies/links the user. The composer represents mentions as Tiptap mention nodes (`addMention` builds `{type:"mention", attrs:{id,label}}`), which a plain-text insert cannot recreate. Note `restoreDraft` (which parses `@[uid:label]` via `parseDraftToContent`) is the path that would preserve them, but re-edit does not use it.

(Custom emoji are inline tokens embedded in `content.text` and re-parsed by `parseEmoji`, so they largely round-trip; the mention-entity loss is the real correctness issue here.)

## Suggested direction
Reconstruct effective content and rebuild mention nodes on restore — either serialize the effective text + `mention.entities` into the `@[uid:label]` form and route through the draft-restore parse path (`parseDraftToContent`/`restoreDraft`), or insert structured mention nodes directly. Reading raw `content.text` via `insertText` is insufficient for edited messages and for mentions.

## ✅ Highlights
- `canReEdit` triple-gate (revoker===self && fromUID===self && contentType===text) and the test matrix are solid.
- Insert path parsing to schema nodes keeps the injection surface safe.

packages/dmworkbase/src/Messages/Revoke/index.tsx
```diff
@@ -64,9 +64,48 @@ export class RevokeCell extends MessageCell {
...
+     * 点击「重新编辑」：将原文夹回输入框
+     */
+    private handleReEdit = () => {
+        const { message } = this.props
+        const conversationContext = (this.props as any).context
+        if (!conversationContext?.insertText) return
+        const textContent = message.content as MessageText
+        const originalText = textContent?.text ?? ''
+        if (!originalText) return
+        conversationContext.insertText(originalText)
```
OctoBoooot, line 91: **[major]** (credit yujiawei) `handleReEdit` calls `context.insertText(originalText)` with the plain `message.content.text`. `insertText` (MessageInput/index.tsx:887) inserts literally — its own comment: `原样追加，不解析 @[uid:label]... mention 格式的反序列化仅在 restoreDraft 中处理`. So a recalled message with @-mentions or custom emoji is restored as literal `@[uid:label]` text, losing the mention chip/emoji structure and uid linkage (MessageText carries a structured `entities` array separate from `.text`). Use the purpose-built `restoreDraft` (parseDraftToContent + setContent) instead — it deserializes mentions/emoji to structured nodes AND replaces the composer content (also fixing the append-vs-replace behavior). Route re-edit through the parse-and-replace primitive, not literal insertText.

OctoBoooot (changes requested):
## Correcting my #744 APPROVE → REQUEST_CHANGES — re-edit uses the literal-insert primitive, losing mention/emoji structure (credit yujiawei)

I approved this having verified the re-edit button's authorization (triple-gated on self) and XSS-safety, but I did NOT check whether the restored content preserves structure. yujiawei did, and I byte-verified she's right. Flipping to REQUEST_CHANGES. Credit yujiawei.

### Major — mentions/emoji restored as literal text, not structured content (`Revoke/index.tsx:handleReEdit` + `MessageInput insertText`)
`handleReEdit` reads `message.content.text` (a plain string) and calls `context.insertText(originalText)`. The `insertText` implementation (MessageInput/index.tsx:887) is explicit about what it does:
```
// 原样追加，不解析 @[uid:label]（与 main 行为一致）
// mention 格式的反序列化仅在 restoreDraft 中处理
editor.commands.insertContent(text);
```
So it inserts the raw string **without parsing mention/emoji markup**. `MessageText` carries a structured `entities` array (`{uid, offset, length}` — see index.tsx:268) separate from `.text`, and re-edit reads only `.text`. Net: a recalled message containing an @-mention or custom emoji is restored with the mention/emoji rendered as **literal `@[uid:label]` text**, losing the mention chip / emoji structure (and the uid linkage).

There is a purpose-built primitive for exactly this — `restoreDraft` (MessageInput/index.tsx, right below `insertText`):
```
// 解析草稿中的 @[uid:label] 格式为 Tiptap 文档结构
const content = parseDraftToContent(text);
editor.commands.setContent(content);
```
Re-edit should route through `restoreDraft` (or expose it on `ConversationContext`) so mentions/emoji deserialize back to structured nodes, and so it **replaces** the composer content rather than appending (which also addresses the append-vs-replace behavior noted earlier). Using `insertText` here is the wrong primitive.

### Major (related) — restores pre-edit content / incomplete restore
yujiawei's 🔴 that the restore doesn't correctly handle edited content is corroborated by the same root cause: reading `.text` + literal `insertText` restores a flattened form, not the message's actual structured content. Confirm what `message.content.text` holds for an edited-then-recalled message and route through the parse-and-replace path.

### Major — PR description omits the MessageRow layout change / should split
The PR is scoped as "re-edit button," but it also carries the MessageRow right-align changes (+44/-3) — which are **byte-identical to the still-open #731** (this PR is stacked on #731's branch). The description doesn't mention them. Split (or document + sequence #731 → #744) so the re-edit change is reviewable/revertable on its own.

### Nits (carried, non-blocking)
- `(this.props as any).context` — add a typed `context` field to the props interface and drop the cast.
- No test exercises the click → `insertText` path (the 5 tests cover only the `canReEdit` visibility matrix); add one asserting the restored content — and it should assert structured mention/emoji restoration once the primitive is fixed.

### Accountability
I verified the dimension I anchored on (authorization + XSS) and reported the feature clean without checking content-fidelity — whether the restored text preserves mention/emoji structure. That's the recurring blind spot: confirming the surface I focused on while missing an adjacent one. Banking: **when a feature restores/round-trips user content, verify structural fidelity (mentions, emoji, formatting entities), not just that it inserts safely — check which insert primitive it uses (literal-append vs parse-replace) and whether a structured field is being flattened to a plain string.**

### Net
Withdrawing my APPROVE: re-edit uses `insertText` (literal append, no mention/emoji parse) instead of `restoreDraft` (parse + replace), so restoring a recalled message with mentions/emoji flattens them to literal text — byte-verified against the code's own comments. Route through the parse-and-replace primitive, split/​document the MessageRow carry, and add a fidelity test. Credit yujiawei for the content-fidelity catch. The authorization gating (self-revoker + self-sender + text-only) and XSS-safety I verified remain correct.

PR #754, opened 2026-07-14
fix(docs): sync line-spacing dropdown display and tidy custom input

ploy-elison:
## Summary

Fixes the four issues reported from the line-spacing toolbar (schema v17) in the docs editor.

### 1. Line-spacing dropdown display desync (bug — fixed)
Picking a value from the line-spacing dropdown changed the block's line-height in the document, but the dropdown kept showing the previous label ("Default"/old value). Root cause: `setLineHeight` is an attribute-only transaction that leaves the caret put, so the toolbar's selection-keyed re-render (`useEditorTick`, snapshot = `from:to`) never fired. With no re-render, React restored the controlled `<select>` to its stale `value` prop.

Fix: added `useEditorValueTick`, which keys the re-render off the displayed value itself (same shape as the existing `useFindState` fix), and used it for both the line-spacing dropdown and the paragraph space-before/after dropdowns so the display always reflects the current selection. Added a red→green regression test proving the controlled value now follows the document.

### 2. Default option label (fixed)
`默认` → `默认行距` (zh-CN) and `Default` → `Default spacing` (en-US), so the empty option reads clearly as "no explicit line spacing".

### 3. Wide custom input (investigated + narrowed)
The oversized field to the right of the dropdown is the **custom line-height multiplier input** (`.octo-line-height-custom`): a commit-on-blur/Enter field for arbitrary multipliers (e.g. `1.75`) that the presets (`1 / 1.15 / 1.5 / 2`) don't cover. It is **not** redundant with the dropdown and should stay. It had no CSS at all, so it rendered at the browser default text-input width (~20 chars). Capped it to `4.5ch` with padding/border matching the toolbar's visual language.

### 4. Space-before / space-after (investigated — conclusion + recommendation)
- **Semantics confirmed correct:** `spaceBefore` → `margin-top` (space above the paragraph), `spaceAfter` → `margin-bottom` (space below).
- **End-to-end verified via existing coverage:** the attrs are sanitized global attributes on paragraph/heading, default to `null` (old docs round-trip losslessly), persist through the Y.Doc as inline styles, are covered by the paragraph-spacing toolbar tests and the schema round-trip tests, and export to docx (`mapSpacing` converts px/em → twips, `spacing.test.ts`). Editing, persistence, HTML round-trip and docx export are all wired and tested.
- **On the px unit — recommend keeping px, do NOT change here.** px is byte-aligned end-to-end today: the frontend sanitizer, the backend `toDOM`, and the docx twips conversion all agree on it (and `em` is already accepted for pasted values, shown as "Custom"). Switching the stored unit to a more document-native one (pt/磅 or em) would be a **schema/semantics change** that must be coordinated across the frontend sanitizer, the backend schema serialization, and the docx exporter — out of scope for a UI/display fix. Flagging this for a separate ticket (发号) for PM/owner to decide; not changed here. The presets already carry the unit in their labels ("4px", "8px", …), so the control is self-documenting in the meantime.

## Verification
- New regression test (`Toolbar — line-spacing dropdown display sync`) is red without the hook change and green with it; the full `Toolbar.test.tsx` suite passes (52/52).
- `tsc --noEmit` clean for all touched files; stylelint reports 0 errors for `styles.css` (added rules use theme variables only, no new warnings).
- Note: the standalone dev harness build is broken on `main` for unrelated reasons (`VoiceInputButton` / `buildDocLink` missing exports in the dev shim), so verification was done through the component tests, which mount the real `Toolbar` + a real Tiptap `Editor` and reproduce the exact controlled-select desync.

## Scope
Pure frontend — no schema, storage semantics, or backend contract touched.

### DIFF
--- a/packages/docs/src/editor/Toolbar.test.tsx
+++ b/packages/docs/src/editor/Toolbar.test.tsx
@@ -685,6 +685,56 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
   })
 })
 
+describe('Toolbar — line-spacing dropdown display sync (XIN-1039 #1)', () => {
+  // Regression: picking a value from the line-spacing dropdown changed the block's line-height
+  // but the dropdown kept showing the old label. setLineHeight is an attribute-only transaction
+  // that leaves the caret put, so a selection-keyed re-render never fired; React then restored
+  // the controlled <select> to its stale `value` prop. The control must re-render off the
+  // line-height value itself so the display follows the selection.
+  let lhEditor: Editor | null = null
+  let holder: HTMLDivElement | null = null
+
+  function mount(content = '<p>hello</p>') {
+    holder = document.createElement('div')
+    document.body.appendChild(holder)
+    lhEditor = new Editor({
+      element: holder,
+      extensions: [StarterKit.configure({ undoRedo: false }), LineHeight],
+      content,
+    })
+    render(<Toolbar editor={lhEditor} />)
+    return document.querySelector('select.octo-line-height') as HTMLSelectElement
+  }
+
+  afterEach(() => {
+    lhEditor?.destroy()
+    lhEditor = null
+    holder?.remove()
+    holder = null
+  })
+
+  it('reflects the picked preset in the dropdown display, not just the document', () => {
+    const select = mount()
+    expect(select.value).toBe('') // starts on the "Default spacing" option
+
+    fireEvent.change(select, { target: { value: '2' } })
+    // Both the document AND the control must show 2 — the control used to snap back to "".
+    expect(lhEditor!.getAttributes('paragraph').lineHeight).toBe('2')
+    expect(select.value).toBe('2')
+  })
+
+  it('updates the dropdown when a custom multiplier is committed via the input', () => {
+    const select = mount()
+    const input = document.querySelector('input.octo-line-height-custom') as HTMLInputElement
+    input.focus()
+    fireEvent.change(input, { target: { value: '1.75' } })
+    fireEvent.keyDown(input, { key: 'Enter' })
+    // 1.75 is off the preset list, so the dropdown must fall to the "custom" sentinel — not "".
+    expect(lhEditor!.getAttributes('paragraph').lineHeight).toBe('1.75')
+    expect(select.value).toBe('custom')
+  })
+})
+
 describe('Toolbar — paragraph spacing controls (SCHEMA_VERSION 17)', () => {
   // The schema/commands/docx/i18n all carried spaceBefore/spaceAfter, but the toolbar exposed no
   // UI to set them. These are the space-before / space-after dropdowns that close that gap.
--- a/packages/docs/src/editor/Toolbar.test.tsx
+++ b/packages/docs/src/editor/Toolbar.test.tsx
@@ -583,7 +583,9 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
   // on every keystroke: typing bounced the caret back into the editor (only the first char
   // landed) and an in-progress value like "1." was rejected by sanitizeLineHeight and snapped
   // back to empty — so a custom multiplier could not be typed at all. It is now a commit-on-
-  // blur/Enter field with a local draft. These tests pin that behaviour.
+  // blur/Enter field with a local draft. These tests pin that behaviour. The field now renders
+  // only while "custom" is the active option, so each test reveals it first (via revealCustom or
+  // by seeding a non-preset value into the block).
   let lhEditor: Editor | null = null
   let holder: HTMLDivElement | null = null
 
@@ -596,6 +598,12 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
       content,
     })
     render(<Toolbar editor={lhEditor} />)
+    return document.querySelector('select.octo-line-height') as HTMLSelectElement
+  }
+
+  // Pick "custom" from the dropdown to reveal the input, then hand it back.
+  function revealCustom(select: HTMLSelectElement) {
+    fireEvent.change(select, { target: { value: 'custom' } })
     return document.querySelector('input.octo-line-height-custom') as HTMLInputElement
   }
 
@@ -615,7 +623,7 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
   })
 
   it('keeps focus in the field and does not commit while typing a multi-char value', () => {
-    const input = mount()
+    const input = revealCustom(mount())
     input.focus()
     expect(document.activeElement).toBe(input)
 
@@ -638,7 +646,7 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
   })
 
   it('commits the typed value to the editor on Enter', () => {
-    const input = mount()
+    const input = revealCustom(mount())
     input.focus()
     fireEvent.change(input, { target: { value: '1.15' } })
     fireEvent.keyDown(input, { key: 'Enter' })
@@ -646,7 +654,7 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
   })
 
   it('commits the typed value to the editor on blur', () => {
-    const input = mount()
+    const input = revealCustom(mount())
     input.focus()
     fireEvent.change(input, { target: { value: '1.75' } })
     fireEvent.blur(input)
@@ -654,32 +662,33 @@ describe('Toolbar — custom line-height input (SCHEMA_VERSION 17, focus-steal R
   })
 
   it('reverts an invalid/partial value to the last committed value on commit', () => {
-    const input = mount()
+    const input = revealCustom(mount())
     input.focus()
     // Commit a good value first.
-    fireEvent.change(input, { target: { value: '1.5' } })
+    fireEvent.change(input, { target: { value: '1.7' } })
     fireEvent.keyDown(input, { key: 'Enter' })
-    expect(lineHeightOf(lhEditor!)).toBe('1.5')
+    expect(lineHeightOf(lhEditor!)).toBe('1.7')
 
-    // Now type a partial value and commit it — it fails sanitize, so the field restores 1.5
+    // Now type a partial value and commit it — it fails sanitize, so the field restores 1.7
     // and the editor keeps the previous multiplier (no bogus value written).
     fireEvent.change(input, { target: { value: '2.' } })
     expect(input.value).toBe('2.') // stays while typing…
     fireEvent.blur(input) // …but on commit it reverts.
-    expect(input.value).toBe('1.5')
-    expect(lineHeightOf(lhEditor!)).toBe('1.5')
+    expect(input.value).toBe('1.7')
+    expect(lineHeightOf(lhEditor!)).toBe('1.7')
   })
 
   it('seeds the custom field from the block the caret is in (round-trip on mount)', () => {
-    // A block already carrying a custom multiplier shows it in the field when the toolbar mounts.
-    mount('<p style="line-height: 1.15">seed</p>')
+    // A block already carrying a non-preset multiplier shows the field seeded with it on mount —
+    // no need to pick "custom" first, because a non-preset value is inherently "custom".
+    mount('<p style="line-height: 1.3">seed</p>')
     const input = document.querySelector('input.octo-line-height-custom') as HTMLInputElement
-    expect(input.value).toBe('1.15')
+    expect(input).not.toBeNull()
+    expect(input.value).toBe('1.3')
   })
 
   it('picking a preset from the dropdown writes the multiplier to the editor', () => {
-    const input = mount()
-    const select = input.parentElement!.querySelector('select.octo-line-height') as HTMLSelectElement
+    const select = mount()
     fireEvent.change(select, { target: { value: '2' } })
     expect(lineHeightOf(lhEditor!)).toBe('2')
   })
@@ -725,6 +734,8 @@ describe('Toolbar — line-spacing dropdown display sync (XIN-1039 #1)', () => {
 
   it('updates the dropdown when a custom multiplier is committed via the input', () => {
     const select = mount()
+    // Reveal the custom input by picking "custom", then commit an off-preset value through it.
+    fireEvent.change(select, { target: { value: 'custom' } })
     const input = document.querySelector('input.octo-line-height-custom') as HTMLInputElement
     input.focus()
     fireEvent.change(input, { target: { value: '1.75' } })
@@ -732,6 +743,86 @@ describe('Toolbar — line-spacing dropdown display sync (XIN-1039 #1)', () => {
     // 1.75 is off the preset list, so the dropdown must fall to the "custom" sentinel — not "".
     expect(lhEditor!.getAttributes('paragraph').lineHeight).toBe('1.75')
     expect(select.value).toBe('custom')
+  })
+})
+
+describe('Toolbar — custom line-height input conditional visibility (XIN-1055)', () => {
+  // The custom multiplier input used to be permanently mounted next to the dropdown, which read as
+  // a confusing always-there box. It must now appear ONLY while "custom" is the active option —
+  // when the user picks "custom" from the dropdown, or when the caret sits in a block already
+  // carrying a non-preset value — and be truly absent (not just hidden) otherwise, so it occupies
+  // no layout space.
+  let lhEditor: Editor | null = null
+  let holder: HTMLDivElement | null = null
+
+  function mount(content = '<p>hello</p>') {
+    holder = document.createElement('div')
+    document.body.appendChild(holder)
+    lhEditor = new Editor({
+      element: holder,
+      extensions: [StarterKit.configure({ undoRedo: false }), LineHeight],
+      content,
+    })
+    render(<Toolbar editor={lhEditor} />)
+    return document.querySelector('select.octo-line-height') as HTMLSelectElement
+  }
+
+  const input = () => document.querySelector('input.octo-line-height-custom') as HTMLInputElement | null
+
+  afterEach(() => {
+    lhEditor?.destroy()
+    lhEditor = null
+    holder?.remove()
+    holder = null
+  })
+
+  it('does not render the custom input on a default-spacing block', () => {
+    mount()
+    expect(input()).toBeNull()
+  })
+
+  it('does not render the custom input while a preset is selected', () => {
+    const select = mount()
+    fireEvent.change(select, { target: { value: '1.5' } })
+    expect(select.value).toBe('1.5')
+    expect(input()).toBeNull()
+  })
+
+  it('reveals the custom input when "custom" is picked from the dropdown', () => {
+    const select = mount()
+    expect(input()).toBeNull()
+    fireEvent.change(select, { target: { value: 'custom' } })
+    expect(select.value).toBe('custom')
+    expect(input()).not.toBeNull()
+  })
+
+  it('hides the custom input again when switching from custom back to a preset', () => {
+    const select = mount()
+    // Enter custom, commit an off-preset value so the input is genuinely showing a custom value.
+    fireEvent.change(select, { target: { value: 'custom' } })
+    const field = input()!
+    field.focus()
+    fireEvent.change(field, { target: { value: '1.3' } })
+    fireEvent.keyDown(field, { key: 'Enter' })
+    expect(lhEditor!.getAttributes('paragraph').lineHeight).toBe('1.3')
+    expect(input()).not.toBeNull()
+
+    // Switch back to a preset: the input must disappear and the preset must apply.
+    fireEvent.change(select, { target: { value: '2' } })
+    expect(lhEditor!.getAttributes('paragraph').lineHeight).toBe('2')
+    expect(select.value).toBe('2')
+    expect(input()).toBeNull()
+  })
+
+  it('shows the input seeded with the value when the caret is in a block already on a custom multiplier (#1 sync, not regressed)', () => {
+    // Continuation of the XIN-1039 #1 display-sync guarantee: a block whose line-height is off the
+    // preset list must surface as "custom" in the dropdown AND reveal the input carrying that value
+    // — without the user having to pick "custom" first.
+    const select = mount('<p style="line-height: 1.3">seed</p>')
+    expect(select.value).toBe('custom')
+    const field = input()
+    expect(field).not.toBeNull()
+    expect(field!.value).toBe('1.3')
   })
 })
 
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -190,6 +190,31 @@ function useEditorTick(editor: Editor): void {
 }
 
 /**
+ * Like useEditorTick but also re-renders when the value returned by `read(editor)` changes —
+ * not only when the selection moves. A setLineHeight / setSpaceBefore transaction rewrites a
+ * block attribute while leaving the caret put, so the selection-keyed snapshot of useEditorTick
+ * does NOT change and React skips the re-render; a controlled <select> is then restored to its
+ * stale `value` prop, so the dropdown kept showing the old label after a value was picked
+ * (line-spacing display desync, XIN-1039 #1). Keying the snapshot off the displayed value itself
+ * makes the control re-render exactly when that value changes — the same fix shape as useFindState
+ * below. Returns the freshly read value so the caller renders from it directly.
+ */
+function useEditorValueTick(editor: Editor, read: (editor: Editor) => string): string {
+  useSyncExternalStore(
+    (cb) => {
+      editor.on('transaction', cb)
+      editor.on('selectionUpdate', cb)
+      return () => {
+        editor.off('transaction', cb)
+        editor.off('selectionUpdate', cb)
+      }
+    },
+    () => `${editor.state.selection.from}:${editor.state.selection.to}:${read(editor)}`,
+  )
+  return read(editor)
+}
+
+/**
  * Subscribe a component to the find/replace plugin state, returning the live FindReplaceState.
  *
  * The plain useEditorTick snapshot keys only off the selection (from:to), so a setFindQuery
@@ -772,8 +797,7 @@ function LineHeightCustomInput({ editor }: { editor: Editor }) {
  * two block types is active, so the control reflects the caret's block.
  */
 function LineHeightSelect({ editor }: { editor: Editor }) {
-  useEditorTick(editor)
-  const current = currentLineHeight(editor)
+  const current = useEditorValueTick(editor, currentLineHeight)
   const isPreset = (LINE_HEIGHTS as readonly string[]).includes(current)
   return (
     <span className="octo-line-height-control">
@@ -814,12 +838,14 @@ const PARAGRAPH_SPACINGS = ['0px', '4px', '8px', '12px', '16px', '24px'] as cons
  * (e.g. an em length pasted in) still shows as "Custom" so it isn't silently reset.
  */
 function ParagraphSpacingSelect({ editor, edge }: { editor: Editor; edge: 'before' | 'after' }) {
-  useEditorTick(editor)
   const attr = edge === 'before' ? 'spaceBefore' : 'spaceAfter'
-  const current =
-    (editor.getAttributes('paragraph')[attr] as string | undefined) ??
-    (editor.getAttributes('heading')[attr] as string | undefined) ??
-    ''
+  const current = useEditorValueTick(
+    editor,
+    (e) =>
+      (e.getAttributes('paragraph')[attr] as string | undefined) ??
+      (e.getAttributes('heading')[attr] as string | undefined) ??
+      '',
+  )
   const isPreset = (PARAGRAPH_SPACINGS as readonly string[]).includes(current)
   const titleKey = edge === 'before' ? 'docs.toolbar.spaceBefore' : 'docs.toolbar.spaceAfter'
   return (
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -731,9 +731,19 @@ function currentLineHeight(editor: Editor): string {
  * and the value is pushed to the editor once, on blur or Enter. An invalid draft on commit is
  * reverted to the last committed value rather than written.
  */
-function LineHeightCustomInput({ editor }: { editor: Editor }) {
+function LineHeightCustomInput({ editor, autoFocus = false }: { editor: Editor; autoFocus?: boolean }) {
   const committed = currentLineHeight(editor)
   const [draft, setDraft] = useState(committed)
+  const inputRef = useRef<HTMLInputElement>(null)
+
+  // Focus the field the moment it is revealed by picking "custom" from the dropdown, so the user
+  // can type a multiplier straight away. Only on mount, and only when the reveal was an explicit
+  // pick — never when the field appears because the caret entered a block that already carries a
+  // custom value (stealing focus there would yank the caret out of the document mid-navigation).
+  useEffect(() => {
+    if (autoFocus) inputRef.current?.focus()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [])
 
   // Adopt the committed value into the draft when it changes underneath us (caret moved to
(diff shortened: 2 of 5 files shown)


Jerry-Xin (dismissed):
**APPROVE** — clean, well-targeted line-spacing dropdown UX fix. Root cause is correctly diagnosed and the fix is minimal and correct. Reviewed at `4c065039`.

### ✅ State-sync correctness (the headline)
The desync is real and precisely explained: `setLineHeight` is an attribute-only transaction that leaves the caret put, so `useEditorTick`'s selection-keyed snapshot (`from:to`) never changes, React skips the re-render, and the controlled `<select>` is restored to its stale `value` prop. `useEditorValueTick` (`packages/docs/src/editor/Toolbar.tsx:200`) fixes it by folding the displayed value into the snapshot key (`` `${from}:${to}:${read(editor)}` ``) and subscribing to both `transaction` and `selectionUpdate` — the control now re-renders exactly when the value changes. Display binding traces cleanly to the live value: `currentLineHeight(editor)` → `isPreset ? current : current ? 'custom' : ''` (`Toolbar.tsx:800`). No stale-state / wrong-source-of-truth path remains. `ParagraphSpacingSelect` gets the same fix shape consistently.

### ✅ Custom input validation
Bounds are enforced at the extension layer, not the UI: `sanitizeLineHeight` (`packages/docs/src/editor/LineHeight.ts:55`) rejects non-strings, regex-gates `^\d+(\.\d+)?$` (so no negative sign and no NaN can pass), then rejects `n <= 0 || n > 10` and non-finite values. `setLineHeight` (`LineHeight.ts:145`) returns `false` on invalid input, and `LineHeightCustomInput.commit()` (`Toolbar.tsx:748`) reverts the draft to the last committed value on that `false`. No crash-render, no silently-broken spacing, no unbounded value. Commit-on-blur/Enter (not per-keystroke) also avoids the caret-steal regression called out in the comment.

### ✅ i18n
Only keyed strings changed (`lineHeightDefault`), and both `en-US.json` and `zh-CN.json` are updated symmetrically ("Default spacing" / "默认行距"). No hardcoded user-facing CJK in the changed `.tsx`.

### ✅ Scope & regression
Five files, all line-spacing scoped. No stray feature-flag flips, no dependency changes, no `console`/`debugger` leaks. CSS is confined to `.octo-line-height-control` / `.octo-line-height-custom`.

### ✅ Tests
The two new tests exercise the sync behavior directly: preset pick reflects in the dropdown display (not just the document), and a custom multiplier (`1.75`) falls to the `custom` sentinel rather than snapping to `""`. Regression-anchored.

No blocking items. Nice, tightly-scoped fix.

OctoBoooot (dismissed):
## Review: #754 — sync line-spacing dropdown display + tidy custom input

**Verdict:** Approve. A tidy, correct display-sync fix for the line-spacing toolbar (schema v17), well-tested and cleanly scoped. No findings. Risk tier: normal. CI green (Build, osv-scan, secret-scan, dependency-review, pr-title-lint).

### Fix — dropdown display desync (`useEditorValueTick`)
Correct root-cause fix. `setLineHeight`/`setSpaceBefore` are attribute-only transactions that leave the caret put, so `useEditorTick`'s selection-keyed snapshot (`from:to`) never changed → no re-render → React restored the controlled `<select>` to its stale `value`. The new `useEditorValueTick` keys the `useSyncExternalStore` snapshot off `${from}:${to}:${read(editor)}` — including the displayed value — so the control re-renders exactly when the value changes even with the selection unmoved, and returns the fresh value for direct render. Subscribes to `transaction` + `selectionUpdate`, mirrors the existing `useFindState` pattern. Applied consistently to `LineHeightSelect` and both `ParagraphSpacingSelect` edges. The new test mounts a real Toolbar + Tiptap Editor and asserts `select.value` follows the document after a pick (red without the hook, green with) — a genuine behavior assertion.

### Custom input (fix #3) — CSS-only, validation preserved
The width cap (`.octo-line-height-custom { width: 4.5ch }`, `:focus`, theme-variable-scoped under `.octo-line-height-*`, no global bleed) is presentational only. `sanitizeLineHeight` is unchanged and still enforces bounds: non-string → null, regex-gated, finite `0 < n ≤ 10` else null. No crash/bad-value/silent-coercion path introduced.

### i18n + scope
`lineHeightDefault` updated symmetrically in both locales ("Default spacing" / "默认行距"). All 5 files are line-spacing-scoped — no bundled flag/dependency changes, no console leaks.

### Praise
The desync diagnosis is precise (attribute-only transaction → selection-key unchanged → controlled-select reverts) and the fix reuses the established `useSyncExternalStore` value-key pattern rather than inventing a new mechanism — and the regression test proves it red→green against a real editor, not a mock.

### Note (informational, not blocking — matches the author's own §4)
The author correctly scoped out the px-vs-pt/em stored-unit question as a schema/semantics change requiring frontend-sanitizer + backend-serialization + docx-exporter coordination — a separate ticket, not this UI fix. Agreed: px is byte-aligned end-to-end today; changing it here would be out of scope.

### Net
Correct display-sync fix (value-keyed re-render), preserved custom-input validation, symmetric i18n, scoped CSS, real red→green regression test. Nothing to change. Approving. Credit ploy-elison.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

**PR:** Mininglamp-OSS/octo-web #754
**Head SHA:** `4c065039d5dbf4391f67a0a54c5513f566411528`

---

# Code Review — PR #754 (octo-web)

## Summary

This PR fixes the line-spacing dropdown display desync (XIN-1039 #1): after picking a preset or committing a custom multiplier, the dropdown kept showing the old label because `setLineHeight` is an attribute-only transaction that leaves the caret in place, so the selection-keyed snapshot of `useEditorTick` didn't change and React restored the stale `value` prop. The fix introduces `useEditorValueTick`, a variant that includes the displayed value in its subscription snapshot, ensuring the controlled `<select>` re-renders when the attribute changes. The same pattern is applied to `ParagraphSpacingSelect` preventively. CSS styling for the line-spacing control group is added (XIN-1039 #3), and the default-option label is updated for clarity in both locales.

## Verification

- ✅ **Hook correctness** — `useEditorValueTick` (`packages/docs/src/editor/Toolbar.tsx:202`) follows the same `useSyncExternalStore` pattern as the existing `useEditorTick` and `useFindState`: subscribes to `transaction` + `selectionUpdate`, provides stable subscribe/unsubscribe, and includes the read value in the snapshot key (`from:to:value`). Cleanup is symmetric.
- ✅ **Data flow: `LineHeightSelect`** (`packages/docs/src/editor/Toolbar.tsx:800`) — `currentLineHeight(editor)` reads `paragraph.lineHeight ?? heading.lineHeight ?? ''`. The select maps this through `isPreset ? current : current ? 'custom' : ''`, which correctly handles all three states (preset, custom value, default).
- ✅ **Data flow: `ParagraphSpacingSelect`** (`packages/docs/src/editor/Toolbar.tsx:842`) — same pattern via inline reader for `spaceBefore`/`spaceAfter`. The `useEditorTick` removal is safe because `useEditorValueTick` subscribes to the same events.
- ✅ **`LineHeightCustomInput` still syncs** (`packages/docs/src/editor/Toolbar.tsx:731`) — reads `currentLineHeight(editor)` via its parent `LineHeightSelect`, which re-renders through `useEditorValueTick`. The render-phase draft sync at `:740` correctly adopts the new committed value.
- ✅ **CSS variables** — `--octo-border`, `--octo-bg`, `--octo-fg`, `--octo-accent` are all defined at root scope (`packages/docs/src/editor/styles.css:3-7`) with dark-mode overrides (`:15-19`).
- ✅ **Test coverage** — two regression tests cover preset selection (`Toolbar.test.tsx:717`) and custom multiplier commit (`Toolbar.test.tsx:726`), both asserting on the select element's displayed value. Cleanup in `afterEach` is correct.
- ✅ **i18n parity** — both `en-US.json` and `zh-CN.json` update the same key (`lineHeightDefault`) consistently.
- ✅ **No side effects on other toolbar components** — five other `useEditorTick` call sites (`:273`, `:601`, `:632`, `:659`, `:983`) are untouched and don't use controlled `<select>` elements.

## Data Flow Traceback

**`LineHeightSelect` (line 800):**
- Consumer: `<select value={isPreset ? current : current ? 'custom' : ''}>` ← `current` ← `useEditorValueTick(editor, currentLineHeight)` ← `editor.getAttributes('paragraph').lineHeight ?? editor.getAttributes('heading').lineHeight ?? ''`
- Trigger: `onChange` → `editor.chain().focus().setLineHeight(v).run()` → dispatches `transaction` event → `useEditorValueTick` subscriber fires → `getSnapshot` returns new value → React re-renders → select `value` prop updates.
- ✅ Data flows correctly from editor state through the hook to the controlled select.

**`ParagraphSpacingSelect` (line 842):**
- Consumer: `<select value={...}>` ← `current` ← `useEditorValueTick(editor, inlineReader)` ← `editor.getAttributes('paragraph')[attr] ?? editor.getAttributes('heading')[attr] ?? ''`
- Same trigger pattern via `editor.chain().focus().setSpaceBefore/After(v).run()`.
- ✅ Data flows correctly.

**`LineHeightCustomInput` (line 731):**
- Consumer: `committed` ← `currentLineHeight(editor)` (called during render) ← parent `LineHeightSelect` re-renders via `useEditorValueTick` → child re-renders → fresh `committed` value.
- Render-phase sync: `if (committed !== seenCommitted.current) { setDraft(committed) }` correctly adopts the new value.
- ✅ Data flows correctly through parent re-render cascade.

## Blind-spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or subscribe/unsubscribe paths changed.
- **C2 (control-flow ordering/nested reuse):** Clear — `useEditorValueTick` is called at exactly two sites (`:800`, `:842`), no nested or double-call risk.
- **C3 (authorization boundary):** N/A — no auth/permission changes.
- **C5 (build/note ≠ runtime):** Clear — CSS uses defined variables, hook follows established pattern, tests verify runtime behavior.
- **C6 (governance/docs self-consistency):** N/A — no governance or policy docs changed.

N/A — automated review review, no prior review rounds.

## Findings

No P0/P1 issues. No correctness, security, build, or functional concerns. The hook is a minimal, well-targeted fix that follows the established `useFindState` pattern already in the codebase.

## Verdict: APPROVED

Clean, well-scoped fix that correctly addresses the display desync root cause (attribute-only transactions not triggering selection-keyed re-renders). The regression tests lock both the preset and custom-value paths, the CSS is consistent with existing styles, and the i18n change is cosmetic. No concerns.

Static analysis only at head `4c065039`; build and tests not executed in this environment.

[Octo-Q] verdict: APPROVE — well-scoped bug fix with correct root-cause analysis, proper hook abstraction following established codebase patterns, adequate regression test coverage, and no side effects on other toolbar components.

yujiawei (dismissed):
# Code Review — PR #754 (octo-web)

**Head SHA reviewed:** `4c065039d5dbf4391f67a0a54c5513f566411528`
**Scope:** frontend-only — docs editor toolbar (`packages/docs`). No schema, storage, or backend contract touched. Verified: the 5 changed files are `Toolbar.tsx`, `Toolbar.test.tsx`, `styles.css`, `en-US.json`, `zh-CN.json`.

## 1. Spec compliance

The PR closes the four reported line-spacing toolbar issues and does exactly what the description claims — nothing more, nothing less.

- **Missing (漏建): none.** All four items addressed: (1) dropdown display desync fixed via the new `useEditorValueTick` hook; (2) default label `默认`→`默认行距` / `Default`→`Default spacing`; (3) custom input width capped in CSS; (4) space-before/after investigated and **correctly scoped OUT** with a recommendation to file a separate ticket for the px→pt/em unit question — no code change, which is the right call for a UI/display fix.
- **Over-build (超建): none.** The one change slightly beyond the literal "line-spacing dropdown" is applying the same hook to the paragraph space-before/after dropdowns (`ParagraphSpacingSelect`). It is the same class of display-desync bug, and the PR body discloses it explicitly. In scope.
- **Divergence (偏离): none.** `spaceBefore`→`margin-top`, `spaceAfter`→`margin-bottom` semantics confirmed. px unit kept, matching the existing frontend sanitizer / backend `toDOM` / docx twips agreement.

**Spec: ✅**

## 2. Code quality

The fix is well-targeted. `useEditorValueTick` keys the `useSyncExternalStore` snapshot off `${from}:${to}:${read(editor)}` so an attribute-only transaction (which leaves the caret put) now forces the re-render that a purely selection-keyed snapshot missed, and the controlled `<select>` follows the document instead of snapping back to its stale `value` prop. This is the same shape as the existing `useFindState` fix in the same file.

Verification I reproduced locally at the head SHA:
- `Toolbar.test.tsx` — **52/52 pass**. Confirmed the regression is real: reverting `LineHeightSelect` to the old `useEditorTick` turns both new assertions red (`select.value` stays `''`/`''` instead of `'2'`/`'custom'`), and the new hook turns them green. Not a tautological test.
- `tsc` — no errors in any of the 5 touched files. (Two pre-existing errors exist in unrelated files on the base — `dmworkbase/src/i18n/format.ts`, `members/sort.ts` — consistent with the PR note that `main` has unrelated breakage.)
- `stylelint` on `styles.css` — 0 errors. The new `.octo-line-height-control` / `.octo-line-height-custom` rules use theme variables (`--octo-bg/fg/border/accent`), all of which exist and carry dark-mode overrides. No new warnings.

Findings (all non-blocking):

- **P2 — `ParagraphSpacingSelect` display-sync has no direct regression test.** The `useEditorTick`→`useEditorValueTick` swap for the space-before/after dropdowns fixes the same desync as #1, but only `LineHeightSelect` got a red→green pick-then-display test. The spacing suite covers set/clear/round-trip-on-mount, not the "pick a value, assert the dropdown display updates" path. Consider adding one symmetric assertion so the spacing half of the fix is locked in too. `Toolbar.tsx:840`, `Toolbar.test.tsx:738`.

- **P2 (nit) — `subscribe` is an unmemoized inline arrow, so listeners are torn down/re-added on every render.** True of `useEditorValueTick` (`Toolbar.tsx:203`), but it is the pre-existing idiom already shipped in `useEditorTick` (`:180`) and `useFindState` (`:229`), and the parent `Toolbar` re-renders on every transaction anyway (`:1197`), so this PR introduces no new churn — it's `editor.on/off` of two handlers per render. If this ever matters, wrap all three subscribers in `useCallback((cb)=>..., [editor])` as one cleanup, not scoped to this PR.

- **P2 (nit) — the hook returns `read(editor)` rather than the `useSyncExternalStore` snapshot.** Slightly less pure under concurrent rendering, but `read` returns a primitive string read synchronously against a single mutable source, and `useFindState` does the identical thing. No observable tearing here.

- Note on SSR: no `getServerSnapshot` — correct and consistent, the Tiptap editor is client-only and the sibling hooks omit it too.

**Quality: Approved**

## 3. Overall verdict

**APPROVED** — Spec ✅ and no P0/P1 findings. The three P2s are optional follow-ups, not merge blockers.

## 4. Suggestions

- Add one `ParagraphSpacingSelect` display-sync test mirroring the line-height one, then the shared-hook fix is fully covered on both controls.
- If the subscriber-memoization nit is ever worth acting on, do it once across `useEditorTick` / `useFindState` / `useEditorValueTick`.
- The px→pt/磅/em unit decision for paragraph spacing is correctly deferred; the follow-up ticket the author flagged is the right home for it.

## 5. Review provenance

Primary review by a human reviewer, cross-checked with an independent long-context advisory pass. That advisory pass raised the subscribe-churn and snapshot points captured above as P2; on inspection each describes the file's pre-existing, already-shipped hook idiom, so none blocks this PR. A second independent line-level advisory pass was attempted but did not complete due to a tooling/gateway fault on this run — it is recorded as absent, not as tacit approval, and its absence does not change the verdict (no P0/P1 exist in the diff).

boris-clark (approved):
Delta re-review of head 659fca43 (previous head 4c065039 was force-pushed, dismissing the four earlier approvals). Scoped to the new commit only — "show custom line-spacing input only when custom is selected" — not a full re-review.

**Verified against the delta (`git diff 4c065039 659fca43`), Toolbar.tsx + Toolbar.test.tsx:**

- Custom multiplier input is now conditionally rendered (`{showCustom && <LineHeightCustomInput …/>}`), so it is truly unmounted — not merely hidden — on default/preset selections and reserves no layout space. Confirmed by the new tests asserting `input()` is null on a default block and on a preset.
- Reveal paths are correct: picking "custom" sets the sticky `customPicked` flag (no write to the block until a value is committed), and a caret landing on a block already carrying a non-preset value sets `hasCustomValue`. `showCustom = hasCustomValue || customPicked` drives both the dropdown value and the input mount.
- `autoFocus` is passed only from `customPicked`, so the field auto-focuses on an explicit "custom" pick but does not steal focus (and does not yank the caret out of the document) when it appears from navigating into an existing custom block. Matches the stated requirement.
- Switching back to a preset resets `customPicked`, applies the preset, and unmounts the input — verified by the round-trip test.
- Sticky "custom" mode is dropped on any `selection.from` change via a render-phase ref/setState guard (`seenFrom`); the ref is updated before `setDraft`/`setState`, so no render loop.
- The XIN-1039 #1 display-sync guarantee is preserved: a non-preset value still surfaces as "custom" with the value seeded. The `value={showCustom ? 'custom' : isPreset ? current : ''}` expression is equivalent to the prior logic for the non-picked case. Dedicated regression test passes.
- New "conditional visibility (XIN-1055)" suite is genuinely red→green: reverting only Toolbar.tsx (keeping the new tests) fails the four conditional-visibility cases; at head all 57 Toolbar tests pass.

**Checks:** Toolbar.test.tsx 57/57 green at head. The 4 EditorShell.test.tsx failures, the 2 `tsc` errors (dmworkbase/i18n/format.ts, members/sort.ts), and the stylelint hex/rgba warnings are all present on base `main` and untouched by this PR — no regression introduced by the delta. This commit touches no CSS.

No blocking issues on the conditional-display increment. Approving. Given the head moved after the previous sign-offs, a human re-review of this delta before merge is welcome.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #754 (octo-web)

## Summary

This PR fixes two related UX issues in the line-spacing toolbar control: (1) the dropdown display now syncs with the actual block line-height value after a pick — previously the controlled `<select>` snapped back to its stale `value` prop because `setLineHeight` is an attribute-only transaction that doesn't move the caret, so the selection-keyed `useEditorTick` never triggered a re-render; (2) the custom multiplier input now renders conditionally — only while "custom" is the active option — instead of being permanently visible next to the dropdown.

The fix introduces `useEditorValueTick`, a variant of `useEditorTick` that includes the read value in the `useSyncExternalStore` snapshot key, making the component re-render when the displayed value changes even if the selection stays put. A `customPicked` sticky state with selection-based reset handles the "reveal custom input on pick, hide on navigation" workflow. The PR also adds CSS for the line-height controls and updates the "Default" label to "Default spacing" / "默认行距".

Overall this is a well-structured, well-tested fix. No P0/P1 issues found.

## Verification

- ✅ **`useEditorValueTick` snapshot correctness** — the snapshot includes `from:to:read(editor)`, so re-renders fire on both selection changes and value changes. The `read` callback (`currentLineHeight`) is cheap and deterministic (`getAttributes` x2), so double invocation per render is safe. (`Toolbar.tsx:202-216`)
- ✅ **`customPicked` render-phase reset** — `seenFrom.current` tracks `selection.from`; when it changes, `setCustomPicked(false)` fires as a render-phase setState on the same component, which React handles correctly (immediate re-render, no infinite loop because the ref is already updated). (`Toolbar.tsx:826-830`)
- ✅ **`autoFocus` scoped to explicit pick** — `autoFocus={customPicked}` is true only when the user picked "custom" from the dropdown, not when navigating to a block with a pre-existing custom value. The `useEffect` runs once on mount with an empty deps array, which is correct for conditional-render semantics. (`Toolbar.tsx:741-747, 861`)
- ✅ **Conditional input unmount** — `showCustom = hasCustomValue || customPicked` correctly handles all paths: explicit "custom" pick, block with existing non-preset value, preset selection (hides), and navigation away (selection-based reset). (`Toolbar.tsx:832, 862`)
- ✅ **Paragraph spacing parity (C1)** — `ParagraphSpacingSelect` already uses `useEditorValueTick` for its own display sync. The paragraph spacing dropdown doesn't offer a custom input (by design — no `<option value="custom">` in the list), so the conditional-reveal feature is line-height-specific. No parity gap. (`Toolbar.tsx:879-916`)
- ✅ **`setLineHeight` command return value** — `LineHeight.ts:148-151` returns `false` when `sanitizeLineHeight` rejects, which the `commit()` function at `Toolbar.tsx:770` correctly handles by reverting the draft.
- ✅ **Test coverage** — three new test suites cover: (a) dropdown display sync regression, (b) conditional input visibility, (c) updated existing tests to use `revealCustom()` helper. Test values changed from `1.5` (preset) to `1.7` (non-preset) to ensure the input stays mounted during the revert test.
- ✅ **i18n consistency** — `lineHeightDefault` updated in both `en-US.json` ("Default" → "Default spacing") and `zh-CN.json` ("默认" → "默认行距").
- ✅ **CSS additions** — `.octo-line-height-control` (inline-flex wrapper) and `.octo-line-height-custom` (4.5ch width, themed borders/focus) are clean and use existing CSS custom properties.

Static analysis only at head `659fca43abda5269a684ea4c56690e6b6ebf65f1`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One minor observation below.

### Nit — Double `read(editor)` invocation in `useEditorValueTick` (`packages/docs/src/editor/Toolbar.tsx:213-215`)

The snapshot function calls `read(editor)` once, and the return statement calls it again. For the current callers (`currentLineHeight` and paragraph spacing's `getAttributes`), this is 2–4 cheap property lookups per render — negligible. If a future caller passes an expensive `read` callback, this could be optimized by caching the snapshot value in a ref. Not actionable today.

## Data Flow Traceback

- **`current` in `LineHeightSelect`** — sourced from `useEditorValueTick(editor, currentLineHeight)`, which reads `editor.getAttributes('paragraph').lineHeight ?? editor.getAttributes('heading').lineHeight ?? ''`. The `useSyncExternalStore` snapshot keys off `from:to:<value>`, so the component re-renders when the line-height attribute changes (via `setLineHeight` transaction). Verified: data flows correctly from the TipTap transaction → snapshot change → re-render → updated `<select> value` prop.
- **`showCustom` flag** — derived from `hasCustomValue` (non-preset, non-empty `current`) OR `customPicked` (sticky UI state). Both are grounded in verifiable state: `current` from editor attributes, `customPicked` from user interaction with reset on selection change. No path leaves the input visible when it shouldn't be.
- **`draft` in `LineHeightCustomInput`** — initialized from `committed` (which is `currentLineHeight(editor)` at render time). Synced to `committed` via render-phase ref comparison (`seenCommitted`). Committed to editor only on blur/Enter via `editor.chain().focus().setLineHeight(v).run()`. Invalid values revert to `committed`. Data flow is clean.

## Blind-Point Checklist

- **C1 (dual-path parity)** — Checked. `ParagraphSpacingSelect` uses `useEditorValueTick` (display sync applied). The custom-input reveal is line-height-specific by design — paragraph spacing has no custom input workflow.
- **C2 (control-flow ordering / nested reuse)** — N/A. No shared logic is reused across different ordering contexts.
- **C3 (authorization boundary)** — N/A. Client-side UI control, no authorization surface.
- **C4 (authorization lifecycle)** — N/A.
- **C5 (build/note through ≠ runtime correct)** — N/A. Pure client-side React component; no build/packaging concerns.
- **C6 (governance/docs self-consistency)** — N/A. No governance or policy changes.

## Extra Observations

- The `mount()` helper in the test now returns the `<select>` element instead of the `<input>`, which required updating all existing tests to use `revealCustom(mount())`. This is a clean refactor that aligns the test surface with the new conditional-render behavior.
- The test "seeds the custom field from the block the caret is in" now uses `1.3` (non-preset) instead of `1.15` (preset in LINE_HEIGHTS). This correctly exercises the `hasCustomValue` path — a non-preset value inherently means "custom" without the user picking it first.

[Octo-Q] verdict: APPROVE — no P0/P1 issues; well-implemented display-sync fix with comprehensive regression tests and clean conditional-render logic.

yujiawei (approved):
# Code Review — PR #754 (octo-web)

Reviewed at head SHA `659fca43`. Merge base `8073e2be`. 5 files, +261/-32, scoped entirely to the docs-editor line-spacing toolbar controls.

## Summary

This is a tight, well-tested fix for three real UX/correctness issues in the line-spacing dropdown:

1. **Display desync** — picking a value changed the block's `lineHeight` but the `<select>` snapped back to its stale label. `setLineHeight` is an attribute-only transaction that leaves the caret put, so the selection-keyed `useEditorTick` snapshot never changed and React restored the controlled select to its old `value` prop. The new `useEditorValueTick` keys the snapshot off the displayed value itself, so the control re-renders exactly when the value changes.
2. **Always-mounted custom input** — the custom-multiplier field is now conditionally rendered (`showCustom`) only while "custom" is the active option, so it takes no layout space otherwise.
3. **Styling + i18n** — the custom input is sized to `4.5ch` instead of the browser default, and the default option label reads "Default spacing" / "默认行距".

## Verification

- `packages/docs` Toolbar suite: **57/57 pass**, including the new XIN-1039 #1 (display sync), XIN-1055 (conditional visibility), and the focus-steal RC suites.
- `tsc` on the changed code is clean. (The two `tsc` errors surfaced project-wide — `dmworkbase/src/i18n/format.ts`, `src/members/sort.ts` — are in files this PR does not touch and are pre-existing.)

## Correctness review

I traced the `showCustom` derivation across the tricky transitions and found no bugs:

- **Pick "custom" then navigate away without typing** → `selection.from` changes, `customPicked` resets, `hasCustomValue` is false, input disappears. No lingering empty box.
- **Pick "custom" over a preset block** → block value is untouched until a multiplier is committed; input seeds from the current value. Correct.
- **Typing in the input does not move the editor caret** (the whole point of the commit-on-blur/Enter design), so `customPicked` is not reset mid-type.
- **Commit an off-preset value** → `setLineHeight` is attribute-only, `selection.from` unchanged, `hasCustomValue` flips true, input stays visible seeded with the committed value. Matches the "stays until switching to a preset" test.

The `autoFocus`-only-on-explicit-pick logic is a nice touch: it focuses the field when the user picks "custom" but not when the field appears because the caret entered a block that already carries a custom value, avoiding a focus yank mid-navigation.

## Notes (non-blocking, advisory)

These are consistency/maintainability observations, not merge blockers:

- **`Toolbar.tsx:826-829` — render-phase state adjustment via a ref.** `setCustomPicked(false)` is called during render, guarded by a `seenFrom` ref comparison. This is the accepted React "adjust state during render" idiom, it is idempotent, and it matches the pre-existing `seenCommitted` pattern already used in `LineHeightCustomInput` (line 754). React's docs suggest the fully concurrent-safe variant stores the previous value in *state* rather than a ref; worth considering if this file ever runs under concurrent features, but it is not a defect here (SPA, no SSR, tests green).
- **`useEditorValueTick` (`Toolbar.tsx:202-212`) subscribes to both `transaction` and `selectionUpdate`.** In ProseMirror a `selectionUpdate` rides inside a transaction, so the callback can fire twice for one interaction. This exactly mirrors the existing `useEditorTick`, and `useSyncExternalStore` skips the re-render when the snapshot string is unchanged, so the extra fire is harmless. Consistent with established convention in this file.
- **No `getServerSnapshot` on the `useSyncExternalStore` calls.** octo-web renders as a client-only Vite SPA (no `renderToString`/SSR path), and the pre-existing `useEditorTick` / `useFindState` hooks already omit it. Only relevant if an SSR path is ever added.

## Verdict

The change does exactly what the title and commits describe — nothing more, nothing less — with comprehensive tests pinning the new behavior. No correctness, security, or data-integrity issues. Approving.

OctoBoooot (approved):
## Re-review (DELTA) @ `659fca43abda` — my prior APPROVE was auto-dismissed by this commit; re-anchoring APPROVE

**Verdict: Approve.** The new commit (single, clean fast-forward — `ahead=1 behind=0`, no rebase drift) refines the line-spacing UI and, notably, **strengthens** the original fix rather than regressing it. Byte-verified the delta and confirmed no regression to the XIN-1039 dropdown-sync fix I approved before. Gate is CLEAN, no `needs-human-review` — so this is genuinely mergeable once re-anchored.

### The delta (1 commit, 2 files): "show custom line-spacing input only when custom is selected" (XIN-1055)
- **Conditional custom input** — `LineHeightCustomInput` was permanently mounted (a confusing always-there box); it now renders **only** while "custom" is the active option: either `hasCustomValue` (the block already carries a non-preset value — e.g. a pasted `1.3`) or `customPicked` (the user just picked "custom" from the dropdown). Conditionally rendered, so it takes no layout space when absent.
- **`customPicked` is a sticky UI mode dropped on caret move** — a render-phase state reset guarded by a `seenFrom` ref (`if (selection.from !== seenFrom.current) { …; setCustomPicked(false) }`). This is the legitimate "derive state during render" idiom (ref-guarded so it can't loop), not a side-effect-in-render bug — the input never lingers over a preset block after navigation.
- **`autoFocus` is correctly scoped** — passed as `autoFocus={customPicked}`, so the field focuses **only** when revealed by an explicit "custom" pick. In the display-sync case (`hasCustomValue` true, `customPicked` false) `autoFocus` is false, so focus is NOT stolen when the caret merely enters a block that already has a custom value — which would otherwise yank the caret out of the document mid-navigation. Verified via the wiring; the comment's claim matches the code.

### The original fix is preserved AND extended (no regression)
- `LineHeightSelect` still uses `useEditorValueTick(editor, currentLineHeight)` — the value-keyed `useSyncExternalStore` snapshot that was the whole point of #754 (an attribute-only `setLineHeight` transaction leaves the caret put, so a selection-keyed snapshot never re-renders and the controlled `<select>` shows a stale label). Intact.
- **`ParagraphSpacingSelect` was upgraded from `useEditorTick` → `useEditorValueTick`** in this delta — so the same attribute-only-transaction desync class is now fixed for the before/after spacing dropdowns too. Consistent hardening, not just the line-height control.

### Tests — strong, cover the exact risk surface (praise)
- Original desync still pinned (XIN-1039 #1: picked preset reflects in the dropdown display; a committed custom multiplier shows the "custom" sentinel, not "").
- New conditional visibility (XIN-1055): no input on a default block, none while a preset is selected, revealed when "custom" is picked, hidden again when switching back to a preset.
- Custom-commit-through-input path (blur/Enter reveal-first flow).

### CI + gate
All checks green at this head. `mergeStateStatus=CLEAN`, `reviewDecision=APPROVED`, **no `needs-human-review`** (size/L, not flagged) — so with this re-anchor the PR is mergeable from the review side (no governance gate, no live CR).

### Net
Clean fast-forward that improves the UX (custom input only when relevant, auto-focus on explicit pick without stealing the caret on display-sync) while preserving and extending the original desync fix (now applied to spacing dropdowns too). No regression, strong tests, CI green. Re-anchoring **APPROVE** @ `659fca43abda`. Credit the author — the `autoFocus={customPicked}` distinction (focus on pick, never on caret-entry) is exactly the subtlety that keeps this from stealing the caret.

Jerry-Xin (approved):
Re-anchoring my prior APPROVE at the new head (my earlier approval auto-dismissed when the branch advanced). The delta since then is one commit touching two files: the line-spacing toolbar component and its test.

## Delta review (since prior approval)
The change makes the custom line-height multiplier input conditionally rendered instead of permanently mounted next to the dropdown. It now appears only while "custom" is the active option — either because the caret sits in a block already carrying a non-preset value, or because the user just picked "custom" from the dropdown — and is truly absent (not merely hidden) otherwise, so it occupies no layout space.

Verified the state logic end-to-end:

- The non-preset-value reveal path preserves the earlier display-sync guarantee: a block on an off-preset multiplier still surfaces as "custom" in the dropdown and seeds the input, with no need to pick "custom" first.
- Picking "custom" from the dropdown reveals the input as a sticky UI mode without writing anything to the block until a multiplier is committed; selecting a preset or default clears that mode and applies/unsets as expected.
- The mode is dropped when the caret moves to another block via a render-phase selection comparison, so the input never lingers over a preset/default block after navigation.
- The auto-focus on reveal is mount-only and gated on the explicit pick, so it does not steal focus when the field appears because the caret entered a block that already carries a custom value.
- No remaining code assumes the input is always mounted.

## Tests
The prior focus-steal tests now reveal the input first, and a new describe block pins the conditional-visibility behavior across all the relevant states, including an explicit guard that the earlier display-sync case is not regressed.

## Gate
CI is green (build, dependency review, OSV scan, secret scan all pass); state is CLEAN and mergeable. Pure frontend, no schema/storage/backend contract or injection/round-trip surface touched.

Delta is clean. APPROVE (re-anchored at the current head).

PR #756, opened 2026-07-14
feat(docs): paragraph/heading indent (increase/decrease) — SCHEMA_VERSION v18

ploy-elison:
## Paragraph / heading indent (SCHEMA_VERSION v18)

Adds a global `indent` attribute to the `paragraph` and `heading` nodes, following the
same block-attr pattern as v5 `textAlign` and v17 line-spacing. Indent is an integer level
(default `null` = no indent, `1..8`), rendered via `margin-left` and round-tripped as a
`data-indent` attribute. Byte-aligned with the backend contract (octo-docs-backend#73) and
`docs/schema/SCHEMA-SPEC.md` §16.

This supersedes the earlier fork-internal draft (ploy-elison/octo-web#18, base `docs-v1`),
which could not merge to the deployable upstream `main`. The implementation is rebased onto
current `main` so it stacks cleanly on the already-landed v16 (font-family) and v17
(line-spacing) features — the diff here is indent-only (11 files).

### What changed
- **`ParagraphIndent` extension** — `increaseIndent` / `decreaseIndent` / `setIndent` /
  `unsetIndent` commands, clamp `[0, 8]`, no-op at level 0. `null` (not `0`) is the empty
  sentinel so plain paragraphs stay attr-free through the Y.Doc (no migration). Bound to
  `Mod-]` / `Mod-[` — list `Tab` / `Shift-Tab` sink/lift is untouched.
- **Registration** — added to both the live and the read-only preview extension sets,
  after `LineHeight`, so the `margin-left` declaration is appended after the v17
  declarations (matching the backend `toDOM` style order).
- **Toolbar** — increase / decrease buttons for the active paragraph or heading; decrease
  is disabled at level 0. i18n keys for en-US and zh-CN.
- **Markdown export** — an indented block is wrapped with an inline `margin-left` style
  alongside `align`.
- **Schema** — `SCHEMA_VERSION` 17 → 18; `indent` is an attribute, so no new node/mark
  entry.

### Verification
- `packages/docs` vitest: the indent-specific suites (schema, extensions, ParagraphIndent,
  markdown export) are all green, including a Yjs collaboration round-trip proving the attr
  survives the collab boundary.
- `tsc --noEmit` (docs typecheck): no new errors introduced by this change.
- Backward compatible: a document with no `indent` attr renders with no indent and no
  migration.

Draft — pairs with octo-docs-backend#73. Do not merge until both sides are reviewed together.

### DIFF
--- a/packages/docs/src/editor/IndentControls.test.tsx
+++ b/packages/docs/src/editor/IndentControls.test.tsx
@@ -1,0 +1,82 @@
+import { describe, it, expect, afterEach, beforeEach } from 'vitest'
+import { render, fireEvent, cleanup, act } from '@testing-library/react'
+import { Editor } from '@tiptap/core'
+import StarterKit from '@tiptap/starter-kit'
+import { Toolbar } from './Toolbar.tsx'
+import { ParagraphIndent, INDENT_MAX_LEVEL } from './ParagraphIndent.ts'
+
+// SCHEMA_VERSION 18 toolbar wiring for the indent group. Beyond the command-boundary unit
+// tests in ParagraphIndent.test.ts, this guards the two things the command tests can't see:
+// (1) the decrease button is disabled at level 0 (the "greyed by default" behaviour, which is
+// intentional — nothing to un-indent), and (2) it RE-ENABLES after an increase. (2) is the
+// real regression risk: the toolbar re-renders via useEditorTick, whose snapshot keys off the
+// selection (from:to). increaseIndent only rewrites a node attribute and leaves the caret put,
+// so a naive selection-only subscription could leave the button stale — the same class of bug
+// the find-counter (useFindState) had to work around.
+
+let editor: Editor | null = null
+
+beforeEach(() => {
+  editor = new Editor({
+    extensions: [
+      StarterKit.configure({ undoRedo: false }),
+      ParagraphIndent.configure({ types: ['paragraph', 'heading'] }),
+    ],
+    content: '<p>hello</p>',
+  })
+  // A real caret in the paragraph, as the boss has when clicking the toolbar.
+  editor.commands.setTextSelection(3)
+})
+
+afterEach(() => {
+  cleanup()
+  editor?.destroy()
+  editor = null
+})
+
+function btn(title: string): HTMLButtonElement {
+  const el = document.querySelector<HTMLButtonElement>(`button[title="${title}"]`)
+  if (!el) throw new Error(`no toolbar button with title="${title}"`)
+  return el
+}
+
+const DECREASE = 'docs.toolbar.indentDecrease'
+const INCREASE = 'docs.toolbar.indentIncrease'
+
+describe('Toolbar — indent decrease button disabled state (SCHEMA_VERSION 18)', () => {
+  it('decrease is disabled at level 0 by default, increase is enabled', () => {
+    render(<Toolbar editor={editor!} />)
+    expect(btn(DECREASE).disabled).toBe(true)
+    expect(btn(INCREASE).disabled).toBe(false)
+  })
+
+  it('decrease re-enables immediately after an increase, and disables again once back at 0', () => {
+    render(<Toolbar editor={editor!} />)
+    expect(btn(DECREASE).disabled).toBe(true)
+
+    // Increase → level 1: decrease must light up on the same render tick (reactivity).
+    act(() => {
+      fireEvent.click(btn(INCREASE))
+    })
+    expect(editor!.getAttributes('paragraph').indent).toBe(1)
+    expect(btn(DECREASE).disabled).toBe(false)
+
+    // Decrease back to 0 → button greys out again.
+    act(() => {
+      fireEvent.click(btn(DECREASE))
+    })
+    expect(editor!.getAttributes('paragraph').indent ?? 0).toBe(0)
+    expect(btn(DECREASE).disabled).toBe(true)
+  })
+
+  it('increase applies level by level and the buttons track the current level', () => {
+    render(<Toolbar editor={editor!} />)
+    for (let i = 1; i <= INDENT_MAX_LEVEL; i++) {
+      act(() => {
+        fireEvent.click(btn(INCREASE))
+      })
+      expect(editor!.getAttributes('paragraph').indent).toBe(i)
+      expect(btn(DECREASE).disabled).toBe(false)
+    }
+  })
+})
--- a/packages/docs/src/editor/IndentControls.test.tsx
+++ b/packages/docs/src/editor/IndentControls.test.tsx
@@ -6,13 +6,17 @@ import { Toolbar } from './Toolbar.tsx'
 import { ParagraphIndent, INDENT_MAX_LEVEL } from './ParagraphIndent.ts'
 
 // SCHEMA_VERSION 18 toolbar wiring for the indent group. Beyond the command-boundary unit
-// tests in ParagraphIndent.test.ts, this guards the two things the command tests can't see:
-// (1) the decrease button is disabled at level 0 (the "greyed by default" behaviour, which is
-// intentional — nothing to un-indent), and (2) it RE-ENABLES after an increase. (2) is the
-// real regression risk: the toolbar re-renders via useEditorTick, whose snapshot keys off the
-// selection (from:to). increaseIndent only rewrites a node attribute and leaves the caret put,
-// so a naive selection-only subscription could leave the button stale — the same class of bug
-// the find-counter (useFindState) had to work around.
+// tests in ParagraphIndent.test.ts, this guards the toolbar reactivity across the FULL indent
+// matrix — the things the command tests can't see:
+// (1) at level 0, decrease is disabled (nothing to un-indent) while increase is enabled;
+// (2) decrease RE-ENABLES after an increase (the reported bug: the button stayed greyed after an
+//     increase because the toolbar only re-rendered on selection changes, and increaseIndent
+//     rewrites a node attribute while leaving the caret put — same class as the find-counter bug);
+// (3) at INDENT_MAX_LEVEL, increase is disabled (the clamp ceiling) and re-enables after a
+//     decrease — symmetric with (1);
+// (4) at every level in between BOTH buttons are clickable and each click steps one level.
+// Both boundaries key off useIndentLevel, so a naive selection-only subscription regresses all of
+// (2)–(4) at once.
 
 let editor: Editor | null = null
 
@@ -78,5 +82,58 @@ describe('Toolbar — indent decrease button disabled state (SCHEMA_VERSION 18)'
       expect(editor!.getAttributes('paragraph').indent).toBe(i)
       expect(btn(DECREASE).disabled).toBe(false)
     }
+  })
+
+  it('increase disables at INDENT_MAX_LEVEL and re-enables after a decrease (symmetric with decrease at 0)', () => {
+    render(<Toolbar editor={editor!} />)
+    // At level 0 increase is enabled (decrease disabled — the other boundary).
+    expect(btn(INCREASE).disabled).toBe(false)
+
+    // Climb to the ceiling; increase must stay enabled until we actually reach INDENT_MAX_LEVEL.
+    for (let i = 1; i <= INDENT_MAX_LEVEL; i++) {
+      act(() => {
+        fireEvent.click(btn(INCREASE))
+      })
+      expect(editor!.getAttributes('paragraph').indent).toBe(i)
+      // Enabled for every level BELOW the ceiling; disabled exactly at the ceiling.
+      expect(btn(INCREASE).disabled).toBe(i >= INDENT_MAX_LEVEL)
+    }
+    expect(editor!.getAttributes('paragraph').indent).toBe(INDENT_MAX_LEVEL)
+    expect(btn(INCREASE).disabled).toBe(true)
+
+    // Clicking the disabled-at-max increase is a no-op — the level does not exceed the clamp.
+    act(() => {
+      fireEvent.click(btn(INCREASE))
+    })
+    expect(editor!.getAttributes('paragraph').indent).toBe(INDENT_MAX_LEVEL)
+
+    // One decrease pulls us off the ceiling → increase re-enables on the same render tick.
+    act(() => {
+      fireEvent.click(btn(DECREASE))
+    })
+    expect(editor!.getAttributes('paragraph').indent).toBe(INDENT_MAX_LEVEL - 1)
+    expect(btn(INCREASE).disabled).toBe(false)
+  })
+
+  it('mid-range levels keep BOTH buttons clickable (full matrix: 0 → max, both enabled in between)', () => {
+    render(<Toolbar editor={editor!} />)
+    // Move to a middle level (2) where neither boundary applies.
+    act(() => {
+      fireEvent.click(btn(INCREASE))
+    })
+    act(() => {
+      fireEvent.click(btn(INCREASE))
+    })
+    expect(editor!.getAttributes('paragraph').indent).toBe(2)
+    expect(btn(INCREASE).disabled).toBe(false)
+    expect(btn(DECREASE).disabled).toBe(false)
+
+    // Decrease actually steps down one level (not a no-op) and both stay enabled at level 1.
+    act(() => {
+      fireEvent.click(btn(DECREASE))
+    })
+    expect(editor!.getAttributes('paragraph').indent).toBe(1)
+    expect(btn(INCREASE).disabled).toBe(false)
+    expect(btn(DECREASE).disabled).toBe(false)
   })
 })
--- a/packages/docs/src/editor/ParagraphIndent.test.ts
+++ b/packages/docs/src/editor/ParagraphIndent.test.ts
@@ -1,0 +1,155 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import { Editor } from '@tiptap/core'
+import StarterKit from '@tiptap/starter-kit'
+import Collaboration from '@tiptap/extension-collaboration'
+import * as Y from 'yjs'
+import { ParagraphIndent, clampIndent, INDENT_MAX_LEVEL, INDENT_STEP_EM } from './ParagraphIndent.ts'
+
+// SCHEMA_VERSION 18: indent inc/dec on paragraph + heading. These assertions guard the
+// command boundaries (no-op at 0, clamp at max), the attr round-trip via data-indent, and
+// that list items never gain the attr (list Tab/Shift-Tab sink/lift stays untouched).
+
+function makeEditor(html: string): Editor {
+  return new Editor({
+    extensions: [
+      StarterKit.configure({ undoRedo: false }),
+      ParagraphIndent.configure({ types: ['paragraph', 'heading'] }),
+    ],
+    content: html,
+  })
+}
+
+let editor: Editor | null = null
+afterEach(() => {
+  editor?.destroy()
+  editor = null
+})
+
+/** Indent attr of the document's first top-level block. */
+function firstIndent(e: Editor): number {
+  return clampIndent(e.state.doc.firstChild?.attrs.indent)
+}
+
+describe('clampIndent', () => {
+  it('floors at 0, ceils at INDENT_MAX_LEVEL, rounds, and coerces non-numbers', () => {
+    expect(clampIndent(-3)).toBe(0)
+    expect(clampIndent(0)).toBe(0)
+    expect(clampIndent(2)).toBe(2)
+    expect(clampIndent(INDENT_MAX_LEVEL + 5)).toBe(INDENT_MAX_LEVEL)
+    expect(clampIndent('4')).toBe(4)
+    expect(clampIndent(2.6)).toBe(3)
+    expect(clampIndent(null)).toBe(0)
+    expect(clampIndent(undefined)).toBe(0)
+    expect(clampIndent(NaN)).toBe(0)
+  })
+})
+
+describe('increaseIndent / decreaseIndent on a paragraph', () => {
+  it('increaseIndent raises the level and decreaseIndent lowers it back', () => {
+    editor = makeEditor('<p>hello</p>')
+    editor.commands.selectAll()
+
+    expect(firstIndent(editor)).toBe(0)
+    expect(editor.commands.increaseIndent()).toBe(true)
+    expect(firstIndent(editor)).toBe(1)
+    expect(editor.commands.increaseIndent()).toBe(true)
+    expect(firstIndent(editor)).toBe(2)
+    expect(editor.commands.decreaseIndent()).toBe(true)
+    expect(firstIndent(editor)).toBe(1)
+  })
+
+  it('decreaseIndent at level 0 is a no-op (returns false, stays 0)', () => {
+    editor = makeEditor('<p>hello</p>')
+    editor.commands.selectAll()
+
+    expect(firstIndent(editor)).toBe(0)
+    expect(editor.commands.decreaseIndent()).toBe(false)
+    expect(firstIndent(editor)).toBe(0)
+  })
+
+  it('increaseIndent clamps at INDENT_MAX_LEVEL (further increase is a no-op)', () => {
+    editor = makeEditor('<p>hello</p>')
+    editor.commands.selectAll()
+
+    for (let i = 0; i < INDENT_MAX_LEVEL; i++) editor.commands.increaseIndent()
+    expect(firstIndent(editor)).toBe(INDENT_MAX_LEVEL)
+    expect(editor.commands.increaseIndent()).toBe(false)
+    expect(firstIndent(editor)).toBe(INDENT_MAX_LEVEL)
+  })
+
+  it('unsetIndent resets an indented block to 0', () => {
+    editor = makeEditor('<p>hello</p>')
+    editor.commands.selectAll()
+    editor.commands.increaseIndent()
+    editor.commands.increaseIndent()
+    expect(firstIndent(editor)).toBe(2)
+    expect(editor.commands.unsetIndent()).toBe(true)
+    expect(firstIndent(editor)).toBe(0)
+  })
+})
+
+describe('increaseIndent on a heading', () => {
+  it('applies to headings too (configured type)', () => {
+    editor = makeEditor('<h2>title</h2>')
+    editor.commands.selectAll()
+    expect(editor.commands.increaseIndent()).toBe(true)
+    expect(firstIndent(editor)).toBe(1)
+    expect(editor.state.doc.firstChild?.type.name).toBe('heading')
+  })
+})
+
+describe('indent HTML round-trip (data-indent <-> indent)', () => {
+  it('parses data-indent from HTML into the indent attr', () => {
+    editor = makeEditor('<p data-indent="3">x</p>')
+    expect(firstIndent(editor)).toBe(3)
+  })
+
+  it('renders an indented block with data-indent and a margin-left style', () => {
+    editor = makeEditor('<p>x</p>')
+    editor.commands.selectAll()
+    editor.commands.increaseIndent()
+    editor.commands.increaseIndent()
+    const html = editor.getHTML()
+    expect(html).toContain('data-indent="2"')
+    expect(html).toContain(`margin-left: ${2 * INDENT_STEP_EM}em`)
+  })
+
+  it('renders no indent attr/style at level 0 (backward-compatible with old docs)', () => {
+    editor = makeEditor('<p>x</p>')
+    const html = editor.getHTML()
+    expect(html).not.toContain('data-indent')
+    expect(html).not.toContain('margin-left')
+  })
+})
+
+// The collab boundary strips attrs the schema does not know. Two editors bound to the SAME
+// Y.Doc — both registering ParagraphIndent — must preserve the indent attr across the sync,
+// proving the attr rides through the Yjs XmlFragment intact (normalized structural check on
+// the decoded attr, not a raw-byte compare).
+describe('Yjs collaboration round-trip', () => {
+  it('preserves the indent attr from one peer to another via the shared Y.Doc', () => {
+    const ydoc = new Y.Doc()
+    const mkPeer = () =>
+      new Editor({
+        extensions: [
+          StarterKit.configure({ undoRedo: false }),
+          ParagraphIndent.configure({ types: ['paragraph', 'heading'] }),
+          Collaboration.configure({ document: ydoc }),
+        ],
+      })
+    const peerA = mkPeer()
+    const peerB = mkPeer()
+    try {
+      peerA.commands.selectAll()
+      peerA.commands.increaseIndent()
+      peerA.commands.increaseIndent()
+      // Both peers share one Y.Doc; the ySync observers apply A's change to B synchronously.
+      expect(clampIndent(peerA.state.doc.firstChild?.attrs.indent)).toBe(2)
+      expect(clampIndent(peerB.state.doc.firstChild?.attrs.indent)).toBe(2)
+    } finally {
+      peerA.destroy()
+      peerB.destroy()
+      ydoc.destroy()
+    }
+  })
+})
--- a/packages/docs/src/editor/ParagraphIndent.ts
+++ b/packages/docs/src/editor/ParagraphIndent.ts
@@ -1,0 +1,137 @@
+// Paragraph / heading indent (SCHEMA-SPEC §16, SCHEMA_VERSION 18).
+//
+// No official Tiptap extension covers plain-paragraph indentation (the built-in list
+// sink/lift only nests listItem nodes), so this is a self-built Extension modelled on
+// @tiptap/extension-text-align (SCHEMA_VERSION 5): it adds a global `indent` ATTRIBUTE to
+// the `paragraph` + `heading` nodes — not a new node/mark — that rides in the Y.Doc node
+// attrs and re-parses faithfully in the read-only preview.
+//
+// The value is an integer indent LEVEL (0 = no indent). It round-trips via a `data-indent`
+// attribute (like Callout's `data-variant`); the visual step is rendered as an inline
+// `margin-left` so the live editor and the static preview render identically without a
+// dedicated stylesheet rule. A missing/0 attr renders no style at all, so old documents
+// (no `indent` attr) are unchanged — backward-compatible by construction, no migration.
+//
+// LOCKSTEP: the attr name (`indent`), the `data-indent` round-trip, the null default and the
+// clamp bounds MUST stay byte-aligned with the backend stub + SCHEMA-SPEC.md at v18, or the
+// collab boundary can strip the unknown attr.
+
+import { Extension } from '@tiptap/core'
+
+/** Em per indent level applied as margin-left (presentational only; not persisted). */
+export const INDENT_STEP_EM = 2
+/** Maximum indent level; increaseIndent clamps here, decreaseIndent clamps at 0. */
+export const INDENT_MAX_LEVEL = 8
+
+/** Coerce any stored value to a valid integer indent level in [0, INDENT_MAX_LEVEL] (0 = none). */
+export function clampIndent(value: unknown): number {
+  const n = typeof value === 'number' ? value : Number(value)
+  if (!Number.isFinite(n)) return 0
+  return Math.min(INDENT_MAX_LEVEL, Math.max(0, Math.round(n)))
+}
+
+/** Storage form of a level: a positive level 1..MAX, or `null` for "no indent" (the default).
+ * Using `null` (not 0) as the empty sentinel keeps a plain paragraph attr-free through the Y.Doc
+ * (y-prosemirror stores every non-null attr), so old docs stay byte-identical and no migration is
+ * needed — exactly how textAlign / lineHeight default to null. */
+export function normalizeIndent(value: unknown): number | null {
+  const level = clampIndent(value)
+  return level > 0 ? level : null
+}
+
+export interface ParagraphIndentOptions {
+  /** Node types the `indent` attr is added to (paragraph + heading). */
+  types: string[]
+}
+
+declare module '@tiptap/core' {
+  interface Commands<ReturnType> {
+    paragraphIndent: {
+      /** Set the indent level of the selected block(s) to an explicit clamped level. */
+      setIndent: (level: number) => ReturnType
+      /** Increase the indent level of the selected block(s) by one (clamped at max). */
+      increaseIndent: () => ReturnType
+      /** Decrease the indent level of the selected block(s) by one (no-op at 0). */
+      decreaseIndent: () => ReturnType
+      /** Reset the indent level of the selected block(s) to 0. */
+      unsetIndent: () => ReturnType
+    }
+  }
+}
+
+export const ParagraphIndent = Extension.create<ParagraphIndentOptions>({
+  name: 'paragraphIndent',
+
+  addOptions() {
+    return {
+      types: ['paragraph', 'heading'],
+    }
+  },
+
+  addGlobalAttributes() {
+    return [
+      {
+        types: this.options.types,
+        attributes: {
+          indent: {
+            // null = no indent (default). y-prosemirror stores every non-null attr, so the null
+            // default keeps plain paragraphs attr-free through the Y.Doc (backward-compatible).
+            default: null,
+            // data-indent <-> indent round-trip (byte-aligned with the backend stub).
+            parseHTML: (element) => normalizeIndent(element.getAttribute('data-indent')),
+            renderHTML: (attributes) => {
+              const level = clampIndent(attributes.indent)
+              if (level <= 0) return {}
+              return {
+                'data-indent': String(level),
+                style: `margin-left: ${level * INDENT_STEP_EM}em`,
+              }
+            },
+          },
+        },
+      },
+    ]
+  },
+
+  addCommands() {
+    // Walk every block of a configured type in the selection and rewrite its indent level.
+    // Returns false when nothing changed (e.g. decreaseIndent at 0), so the boundary is a
+    // clean no-op and the toolbar button reflects it.
+    const applyIndent =
+      (compute: (current: number) => number) =>
+      ({ state, dispatch }: { state: import('@tiptap/pm/state').EditorState; dispatch?: (tr: import('@tiptap/pm/state').Transaction) => void }) => {
+        const { from, to } = state.selection
+        const tr = state.tr
+        let changed = false
+        state.doc.nodesBetween(from, to, (node, pos) => {
+          if (!this.options.types.includes(node.type.name)) return
+          const current = clampIndent(node.attrs.indent)
+          const next = clampIndent(compute(current))
+          if (next !== current) {
+            // Store null (not 0) for "no indent" so the attr is stripped from the Y.Doc.
+            tr.setNodeAttribute(pos, 'indent', next > 0 ? next : null)
+            changed = true
+          }
+        })
+        if (changed && dispatch) dispatch(tr)
+        return changed
+      }
+
+    return {
+      setIndent: (level) => applyIndent(() => level),
+      increaseIndent: () => applyIndent((c) => c + 1),
+      decreaseIndent: () => applyIndent((c) => c - 1),
+      unsetIndent: () => applyIndent(() => 0),
+    }
+  },
+
+  addKeyboardShortcuts() {
+    return {
+      // Tab/Shift-Tab keep their list sink/lift behavior — those keys are owned by the list
+      // extensions and only fire inside a listItem. We bind Mod-] / Mod-[ for paragraph indent
+      // so we never shadow the list keymap.
+      'Mod-]': () => this.editor.commands.increaseIndent(),
+      'Mod-[': () => this.editor.commands.decreaseIndent(),
+    }
+  },
+})
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -50,6 +50,17 @@ const IconAlignJustify = () => (
     <path d="M3 5h18v2H3V5zm0 4h18v2H3V9zm0 4h18v2H3v-2zm0 4h18v2H3v-2z" />
   </svg>
 )
+// Indent buttons (SCHEMA_VERSION 18): lines with a right/left chevron marking the indent direction.
+const IconIndentIncrease = () => (
+  <svg className="octo-tb-icon" viewBox="0 0 24 24" aria-hidden="true">
+    <path d="M3 5h18v2H3V5zm8 4h10v2H11V9zm0 4h10v2H11v-2zm-8 4h18v2H3v-2zm.4-8.8L7 12l-3.6 2.8V6.2z" />
+  </svg>
+)
+const IconIndentDecrease = () => (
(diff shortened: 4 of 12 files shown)


Jerry-Xin (dismissed):
### Review — paragraph/heading indent v18 (FE + BE lockstep)

Reviewed the rebased indent-only diff (this PR) together with the backend counterpart
octo-docs-backend#73 and `docs/schema/SCHEMA-SPEC.md` §16. All lockstep checks pass.

**Byte-alignment (FE `ParagraphIndent.ts` ↔ BE `src/schema/index.ts` ↔ SPEC §16)**
- Attr name `indent`, `data-indent` round-trip, level `1..8`, `null` = no indent (0-normalized-to-null sentinel), step `2em`, max `8` — identical on both sides.
- Serialization matches exactly: BE `toDOM('paragraph', {textAlign:'right', indent:2})` → `style="text-align: right; margin-left: 4em"` + `data-indent="2"`; FE renders `margin-left: 4em` + `data-indent="2"` with `ParagraphIndent` registered after `LineHeight` so `margin-left` is appended after the v17 declarations.
- Parse is `data-indent`-authoritative on both; a bare `margin-left` with no `data-indent` → `null` (not treated as an indent level).
- Hostile/out-of-range coercion matches: `99 → 8`, `0 / -3 / "evil" → null` on both FE (`clampIndent`/`normalizeIndent`) and BE (`sanitizeIndent`), with BE also sanitizing on the JSON write path (`sanitizeBlockAttrValues`).

**Behavioral checks**
1. `indent` is an ATTRIBUTE on `paragraph` + `heading`, not a new node/mark — confirmed in schema node/mark lists (excluded) and `spec.attrs.indent` (present, default `null`).
2. Registered in both the live and read-only preview extension sets.
3. Indent binds `Mod-]` / `Mod-[`; list `Tab` / `Shift-Tab` sink/lift is untouched, and `listItem` carries no `indent` attr (asserted).
4. Markdown export emits the `margin-left` style for an indented block.
5. Yjs collaboration round-trip test proves the attr survives the collab boundary.
6. Backward compatible — old docs (no `indent`) render with no indent, no migration.

**Test / gate status**
- `packages/docs` indent suites (schema / extensions / ParagraphIndent / markdown export): **81/81 green**.
- `tsc --noEmit` (docs): no new errors. The two remaining errors (`dmworkbase/i18n/format.ts`, `members/sort.ts`) are pre-existing on `main`, unrelated to this change.
- Full docs vitest run shows 4 failures in `EditorShell.test.tsx` (export / ≡-menu i18n) — confirmed identical on base `main` (pre-existing, unrelated to indent).
- The docs package has no wired eslint config (root `.eslintrc.js` is commented out; turbo skips it), so no eslint gate was run.

**Notes for human re-review**
- Minor, non-blocking: `IndentControls` does not call `useEditorTick`, so the decrease button's `disabled` state can lag a selection/indent change until the toolbar re-renders. The commands are safe no-ops at the boundary, so behavior is correct; only the disabled visual may be stale.
- During the rebase I preserved `escapeHtmlAttr(align)` in the markdown export wrap (the earlier fork-internal branch had dropped the escaping on the `align` value).

Conclusion: **APPROVED** (internal) — implementation is correct and fully byte-aligned with the backend + SPEC. Flagging for upstream human re-review on the merge path this PR now takes (base `main`).

Jerry-Xin (dismissed):
## Review — indent (increase/decrease) + SCHEMA_VERSION v18

Reviewed at head f15a1fce20c3. This is a clean, well-scoped, self-contained feature. Auditing the two highest-risk axes for a schema bump — backward-compat/migration and round-trip fidelity — both hold up. Approving with one non-blocking parity note.

### ✅ Schema migration / backward-compat (no migration needed, correctly)
The new `indent` attribute defaults to `null` (not 0). y-prosemirror persists only non-null node attrs, so existing v17-and-earlier docs stay attr-free and byte-identical through the Y.Doc — they load unchanged with no migration. Missing/absent attr coerces safely to level 0 via the clamp helper. Verified in the extension (`addGlobalAttributes` default null) and covered by tests asserting a level-0 block renders no attr/style (`renders no indent attr/style at level 0 (backward-compatible with old docs)`). This matches the established v5 textAlign / v17 line-spacing pattern.

### ✅ Round-trip fidelity
- Persistence (Y.Doc collab): preserved and directly tested via a two-peer shared-doc sync assertion.
- HTML: `data-indent` parse ↔ render is symmetric and tested both directions.
- Markdown export: updated to emit the block wrapper with a margin-left style; align + indent combine into a single wrapper tag; level 0 stays plain. Tested.
- No Markdown importer exists in the package, so Markdown export is intentionally one-way — no asymmetry there.

### ✅ Input bounds
Indent level is clamped to [0, max] with rounding and non-number coercion in a single helper; increase no-ops at max, decrease no-ops at 0 (command returns false so the boundary is a clean no-op and the decrease button is disabled at 0). Directly tested.

### ✅ i18n
Both en-US and zh-CN gain the two toolbar keys; the toolbar titles are wired through the translation helper. No hardcoded user-facing strings.

### ✅ Scope-clean
11 files scoped to indent + the version bump. No feature-flag flips, no dependency changes, no debug-logging leaks. The toolbar control renders unconditionally (feature fully enabled), consistent with the PR intent.

### 🟡 Non-blocking: DOCX export drops the indent
The DOCX paragraph/heading exporter reads textAlign and spacing but not the new indent attribute (export/docx/nodes.ts), so a DOCX export will silently omit paragraph/heading indent even though Markdown export was updated for parity. This does not corrupt saved state (the Y.Doc keeps the attribute), so it is not a blocker — but it is an export-format parity gap worth a follow-up: map the level to a left indentation in the DOCX paragraph options, mirroring how textAlign/spacing already flow.

### Notes
- CI: meaningful checks pass; Build shows as skipping (draft-gated). PR is currently marked draft.

Nice work — the null-default backward-compat reasoning and the collab/HTML round-trip tests are exactly the right guards for a schema bump.

OctoBoooot:
## #756 — advisory note (skipping the formal review verdict: PR is a draft)

This PR is a **draft**, so I'm not posting a gating APPROVE/REQUEST_CHANGES review event — a formal verdict on a draft would signal false finality while you're still iterating. Flagging the skip explicitly (not silently) per convention, plus an advisory read since this is a `SCHEMA_VERSION v18` migration (high-risk tier → will require human review once Ready).

### Looks migration-sound (advisory, from a read at `f15a1fce`)
- **v17→v18 backward-compat is correct by construction.** The new `indent` attr defaults to `null` (not 0), so old docs with no `indent` attr stay byte-identical through the Y.Doc (`y-prosemirror` stores every non-null attr) — no migration needed, exactly the v5 `textAlign` / v17 `lineHeight` null-default pattern. Good.
- **Bounds enforced** — `clampIndent` coerces to integer `[0, INDENT_MAX_LEVEL=8]` (round + min/max), `increaseIndent` clamps at max, `decreaseIndent` at 0, `normalizeIndent` uses the null sentinel for level 0. No negative/NaN/unbounded path.
- **Attribute, not a new node/mark** — added to paragraph+heading only (list Tab/Shift-Tab sink/lift untouched), `data-indent` round-trip; `SCHEMA_NODES`/`SCHEMA_MARKS` correctly unchanged. Backend-lockstep is called out in-comment.

### Before marking Ready, worth confirming (the usual v-bump checklist)
- **Backend/SCHEMA-SPEC lockstep** — the comment says the attr name, `data-indent` round-trip, null default, and clamp bounds MUST stay byte-aligned with the backend stub + SCHEMA-SPEC.md at v18. Confirm the backend side landed/agrees (the frontend can't verify that half).
- **Round-trip + export tests** — you have `ParagraphIndent.test.ts` + a schema test; make sure the markdown export change (`markdown.ts +16/-6`) round-trips the indent (and that a >MAX or negative stored value clamps rather than leaks).
- CI isn't fully run yet (draft) — the i18n/Build/schema checks will need to be green.

I'll do the full gating review (and the mandatory `needs-human-review` for the schema change) once you flip it out of draft. Flagging now so the migration-critical items are visible early. Credit ploy-elison — the null-default backward-compat reasoning is exactly right for a v-bump.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #756 (octo-web)

## Summary

This PR adds paragraph/heading indentation to the TipTap editor: a new `ParagraphIndent` extension (global `indent` attribute on paragraph + heading nodes), toolbar buttons (increase/decrease), keyboard shortcuts (Mod-]/Mod-[), markdown export support, and a SCHEMA_VERSION bump from 17 to 18. The implementation follows the established pattern of TextAlign and LineHeight (global attributes with null defaults), includes comprehensive unit tests, and maintains backward compatibility (no migration needed for existing documents).

## Verification

- ✅ **Backward compatibility** — `indent` defaults to `null` (not 0), so plain paragraphs carry no attr through Y.Doc; old documents remain byte-identical. `clampIndent` handles any input type safely.
- ✅ **Clamp boundaries** — `clampIndent` enforces `[0, INDENT_MAX_LEVEL=8]`; commands correctly no-op at boundaries (decrease at 0, increase at max), verified by tests.
- ✅ **Extension registration order** — ParagraphIndent registered AFTER LineHeight in `extensions.ts:212` (live) and `:330` (preview), so canonical style serialization order is preserved (text-align → line-height → margin-top/bottom → margin-left).
- ✅ **List isolation** — Keyboard shortcuts use Mod-]/Mod-[ instead of Tab/Shift-Tab, so list sink/lift behavior is not shadowed. Comment at `ParagraphIndent.ts:131-133` explains the design.
- ✅ **Collab sync** — Test at `ParagraphIndent.test.ts:120-154` verifies indent attr survives Y.Doc round-trip between two synced editors.
- ✅ **Schema version** — Bumped 17→18 at `schema/index.ts:62-68`, matching the SCHEMA-SPEC contract.

Static analysis only at head `f15a1fce20c322e1db2f1dc191ae61336daa69fa`; build and tests not executed in this environment.

## Findings

### P2 — Markdown export omits `data-indent`, so re-import loses indent (`packages/docs/src/export/markdown.ts:215`)

The `wrapAlign` function emits `style="margin-left:Xem"` for indented blocks but does NOT emit `data-indent`. The ParagraphIndent extension's `parseHTML` reads ONLY `element.getAttribute('data-indent')` (not the inline style), so a round-trip through markdown export → re-import drops the indent attribute entirely.

Data-flow trace:
- `wrapAlign` at `markdown.ts:207-217` constructs the HTML wrapper with `style="margin-left:${indent * INDENT_STEP_EM}em"` but no `data-indent` attr.
- `ParagraphIndent.parseHTML` at `ParagraphIndent.ts:82` calls `element.getAttribute('data-indent')` only.
- Result: indent level is present in the exported markdown (as visual margin), but lost on re-import.

Fix: emit `data-indent="${indent}"` alongside the style in `wrapAlign`, matching the `renderHTML` pattern at `ParagraphIndent.ts:85-89`.

Diff-scope: NEW — this is new code in this PR (the indent export path did not exist before).

### P2 — Increase-indent toolbar button not disabled at max level (`packages/docs/src/editor/Toolbar.tsx:882`)

The `IndentControls` component disables the decrease button when `current <= 0` but does NOT disable the increase button when `current >= INDENT_MAX_LEVEL`. The command correctly no-ops at max (returns false), so no functional bug, but the UI asymmetry is confusing — the user can click a button that does nothing.

Fix: add `disabled={current >= INDENT_MAX_LEVEL}` to the increase button at `Toolbar.tsx:882`.

Diff-scope: NEW — this is new UI code in this PR.

## Data-flow Tracing

- **`indent` attr in Y.Doc** — defaults to `null` (ParagraphIndent.ts:79), so y-prosemirror strips it from plain paragraphs. Commands store `null` (not 0) when unsetting (ParagraphIndent.ts:113). Correct.
- **`clampIndent` at command boundary** — `applyIndent` at ParagraphIndent.ts:99-115 clamps the computed level before comparing to current, so out-of-range inputs (e.g. `setIndent(999)`) are safely clamped to 8.
- **Toolbar `current` value** — `Math.max(Number(paragraph.indent) || 0, Number(heading.indent) || 0)` at Toolbar.tsx:868-871. The `|| 0` handles null/undefined. Correct.
- **Markdown export `indent` value** — `clampIndent(node.attrs?.indent)` at markdown.ts:209 ensures the exported level is in `[0, 8]`. Correct.
- **i18n keys** — `indentIncrease` / `indentDecrease` added to both en-US and zh-CN. Correct.

## Blind-spot Checklist

- **C1 (dual-path parity)** — Paragraph + heading are symmetric: both receive the `indent` attr, both are clamped, both are rendered. The only asymmetry is the markdown export tag (`<p>` vs `<div>`), which matches the existing pattern for alignment. Clear.
- **C2 (control-flow ordering)** — `applyIndent` walks `state.doc.nodesBetween` once, applies `setNodeAttribute` per block. No double-application risk. Clear.
- **C3 (auth boundary)** — N/A; no auth/permission changes.
- **C4 (container-member state)** — N/A.
- **C5 (build/note)** — N/A; no build/CI changes.
- **C6 (governance/docs)** — N/A; no governance/docs changes.

## Verdict

Clean, well-tested implementation following established patterns. No P0/P1 issues. The two P2 items are non-blocking: the markdown round-trip data loss is a correctness gap (indent survives collab sync but not markdown export/import), and the missing disabled state is a minor UX asymmetry.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; P2 items are non-blocking correctness and UX improvements.

yujiawei (changes requested):
# Code Review — PR #756 (octo-web)

Reviewed at head SHA `f15a1fce20c322e1db2f1dc191ae61336daa69fa`, diffed against `main` (merge-base `8073e2be`, so the branch stacks cleanly on the landed v16/v17 work).

The implementation of the `indent` global attribute itself is clean and faithfully mirrors the established v5 `textAlign` / v17 `lineHeight` pattern: a self-built extension adding a global attr (not a new node/mark), `null` default so old docs stay attr-free through the Y.Doc (no migration), integer clamp `[0, 8]`, and a `data-indent` ⇄ `indent` round-trip. The `packages/docs` unit suites for this change (ParagraphIndent, extensions, markdown, schema — 81 tests) pass locally, and `tsc --noEmit` introduces no new errors (the two pre-existing errors are in `dmworkbase` and `members/sort.ts`, untouched here).

There is, however, one blocking gap.

## Blocking

### P1 — DOCX export silently drops the new `indent` attribute (lossy export)

Markdown export was updated in this diff to carry indent (`export/markdown.ts` `wrapAlign`, emitting `style="margin-left:…em"`), but the DOCX exporter was not. `convertParagraph` and `convertHeading` read `textAlign` (v5, via `mapTextAlign`) and line-spacing (v17, via `mapSpacing`) but never read `node.attrs.indent`:

```ts
// packages/docs/src/export/docx/nodes.ts:171
function convertParagraph(node: MdNode, ctx: DocxContext): Paragraph {
  const runs = convertInlineContent(node.content ?? [], ctx.emojiGlyph)
  const align = mapTextAlign(node.attrs?.textAlign)
  const spacing = mapSpacing(node.attrs)
  return new Paragraph({
    children: runs,
    ...(align ? { alignment: align } : {}),
    ...(spacing ? { spacing } : {}),
    // no indent handling
  })
}
```

`convertHeading` (`nodes.ts:183`) has the identical omission, and there is no `mapIndent` in `styles.ts` alongside `mapSpacing` (`styles.ts:246`). DOCX export is a live, user-reachable path (`EditorShell.tsx:449` → `exportDocToDocx`), so a paragraph indented in the editor exports to Word with **zero indent and no error** — silent data loss.

This is the crux of the concern: the PR's stated design is that indent "follows the same block-attr pattern as v5 `textAlign` and v17 line-spacing," and **both of those are converted in DOCX export**. Updating one exporter (markdown) but not the other is an inconsistency that produces silent lossy output. The `docx` library already supports paragraph indent (`indent: { left: <twips> }` is used for blockquotes at `nodes.ts:304` and callouts at `nodes.ts:515`), so the fix is small and follows the existing `mapSpacing` shape:

- add a `mapIndent(attrs)` in `styles.ts` that reads `clampIndent(attrs.indent)` and returns `{ left: level * INDENT_STEP_EM * PX_PER_EM * TWIPS_PER_PX }` (= `level * 480` twips with the existing `PX_PER_EM=16`, `TWIPS_PER_PX=15` constants), or `undefined` at level 0;
- spread it into the `Paragraph` in both `convertParagraph` and `convertHeading`;
- add a docx export test mirroring the v17 `spacing.test.ts` precedent.

## Non-blocking (suggestions)

- **P2 — No feature flag for the indent toolbar UI.** v16 font-family (`FONT_FAMILY_ENABLED`) and v17 line-spacing (`LINE_SPACING_ENABLED`, `config.ts:143`) gate their toolbar controls behind a `VITE_DOCS_*` flag for staged rollout/rollback, while keeping the extension registered for round-trip compat. `IndentControls` is rendered unconditionally (`Toolbar.tsx:1376`). This is an ops/rollback convenience, not a correctness issue (v5 `textAlign` also ships flag-free), so it's a suggestion rather than a blocker — but matching the two most recent siblings would be consistent.

- **P2 — Use `clampIndent` in the toolbar level read.** `IndentControls` computes the active level with `Number(editor.getAttributes('paragraph').indent) || 0` (`Toolbar.tsx:870`). It's functionally correct, but the module already exports `clampIndent` for exactly this coercion; using it would be clearer and consistent with `markdown.ts` and the extension itself.

- **P2 — Add an adversarial/injection test for the indent style path.** The value is provably safe by construction — `clampIndent` coerces to an integer in `[0, 8]` before it ever reaches the `margin-left` string in both `ParagraphIndent.renderHTML` and markdown `wrapAlign`, so no CSS/HTML injection is possible via `indent` or `data-indent`. There's just no explicit test asserting this, unlike the existing XSS-hardening cases for align/color/variant in `markdown.test.ts`. A one-line adversarial case would lock in the guarantee.

## Notes / not defects

- **Toolbar "disabled" state is not stale-on-command in any way that differs from existing controls.** `IndentControls` intentionally does not call `useEditorTick` itself; it relies on the parent `Toolbar`'s subscription, exactly like its true peer `AlignControls` (`Toolbar.tsx:681`). Note that `useEditorTick`'s snapshot is keyed only on selection (`from:to`), and the toolbar deliberately does not re-render on attribute-only transactions (see the comment at `Toolbar.tsx:748-751`); adding `useEditorTick` to `IndentControls` would therefore not change the decrease-button behavior. The button correctly no-ops at level 0 (`decreaseIndent` returns `false`), and the disabled state refreshes on the next selection move — identical to the shipped align/spacing controls. No change needed here.

## Coverage / things a human should confirm

This PR is security-sensitive and is explicitly a draft "byte-aligned with the backend contract (octo-docs-backend#73) and SCHEMA-SPEC.md §16." Neither the backend stub nor SCHEMA-SPEC.md is present in this checkout, so I could **not** verify the byte-alignment claim (attr name, `data-indent` round-trip, null default, clamp bounds, and the style-declaration order after the v17 line-spacing declarations). Please confirm the client↔backend `toDOM`/serialization contract at v18 before landing the two sides together, as the PR intends.

OctoBoooot:
@/tmp/pr756-calibration-f15a1fce.md

boris-clark (commented):
Delta review @ 1f861c9a (fix commit over f15a1fce) — APPROVED (internal; a human re-review of just this reactivity delta is recommended before merge).

Scope: 2 files, +110/-4 — Toolbar.tsx (new useIndentLevel + swap of the inline compute) and the new IndentControls.test.tsx.

- Snapshot: useIndentLevel subscribes to transaction + selectionUpdate and returns Math.max(paragraph.indent, heading.indent) as a primitive number. increase/decreaseIndent rewrite a node attr (caret stays put) → they fire transaction → the buttons re-render exactly when the level changes; the primitive return keeps Object.is stable, so no getSnapshot infinite-loop. Cleanup removes both listeners. Same pattern as useFindState.
- Symmetry: decrease disabled={current <= 0} is now reactive — level 0 disabled, enables on the same tick after increase, disables again back at 0, stays enabled up to INDENT_MAX_LEVEL.
- red→green: reverting to the pre-fix inline compute makes the two new reactivity cases fail (decrease stays disabled after increase); the level-0 default still passes. Test captures the real bug.
- No regression: only Toolbar.tsx + test touched, so list Tab/Shift-Tab, the v18 schema byte-alignment, and indent attr / markdown / Yjs round-trip are untouched. Delta-scope suites green (IndentControls + ParagraphIndent + extensions + markdown + schema = 84 passed).
- Typecheck: the two tsc errors in the tree exist identically at base f15a1fce — no new errors from this delta.

Not merging — leaving the final merge decision to a human reviewer.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #756 (octo-web)

Reviewed at head `1f861c9a`. Scope: new `ParagraphIndent` Tiptap extension (indent global attr on paragraph/heading), toolbar indent controls, SCHEMA_VERSION v17→v18, markdown export indent rendering, i18n keys, and tests.

## Summary

Well-structured feature PR that adds paragraph/heading indent support to the collaborative document editor. The extension design is sound — using a global attribute with `null` default for Yjs backward compatibility, `data-indent` HTML round-trip, and `useSyncExternalStore` for toolbar state. Markdown export correctly renders indent via inline `margin-left` styles. One P2 gap exists in the DOCX export path, and a minor nit in the toolbar UI.

## Verification

- ✅ **Extension registration** — `ParagraphIndent` registered after `LineHeight` in both `buildExtensions` and `buildPreviewExtensions` at `extensions.ts`, matching `TextAlign`'s type configuration pattern exactly.
- ✅ **Yjs compatibility** — `normalizeIndent` returns `null` for level 0, so existing documents with no `indent` attr remain attr-free in the Yjs CRDT layer. SCHEMA_VERSION bumped to 18 with correct migration comment.
- ✅ **Command logic** — `applyIndent` walks selection blocks correctly, `clampIndent` handles null/undefined/NaN → 0 and clamps to `[0, INDENT_MAX_LEVEL]`. Keyboard shortcuts use `Mod-]`/`Mod-[` to avoid shadowing list `Tab`/`Shift-Tab`.
- ✅ **Toolbar state** — `useIndentLevel` uses `useSyncExternalStore` with `Math.max(paragraph.indent, heading.indent)` snapshot, subscribing to both `transaction` and `selectionUpdate` events. Matches the established `useFindState` pattern.
- ✅ **Markdown export** — `wrapAlign` at `markdown.ts` correctly imports `clampIndent` and emits `<p style="margin-left:Nem">` / `<div style="margin-left:Nem">` for headings with indent > 0. Indent is computed arithmetically (no user string interpolation → no XSS vector).
- ✅ **Tests** — `ParagraphIndent.test.ts` covers clampIndent boundaries, command increase/decrease, no-op at 0, clamp at max, HTML round-trip, and Yjs collaboration round-trip. `IndentControls.test.tsx` covers toolbar disabled-state toggling.
- ✅ **i18n** — Both `en-US.json` and `zh-CN.json` have `indentIncrease` and `indentDecrease` keys.

## Findings

### P2 — DOCX export silently drops indent attribute (`packages/docs/src/export/docx/nodes.ts:171`)

`convertParagraph` (line 171) and `convertHeading` (line 183) read `textAlign` and `spacing` attributes but do not read the new `indent` attribute. When a user exports an indented document to DOCX, all paragraph/heading indentation is silently lost — the exported `.docx` renders every paragraph flush-left regardless of the indent level set in the editor.

The markdown export path handles indent correctly (verified above), so this is an asymmetric export gap: the same document exports with indent in Markdown but without it in DOCX.

```ts
// Fix sketch — add to convertParagraph / convertHeading:
const indent = clampIndent(node.attrs?.indent);
if (indent > 0) {
  // Apply indent as left margin in twips (1 em ≈ 240 twips at 12pt)
  paragraphProperties.push({
    'w:ind': { _attr: { 'w:left': indent * INDENT_STEP_EM * 240 } }
  });
}
```

Not a blocker — the feature is additive and the DOCX gap does not corrupt data — but worth addressing in a follow-up to avoid user confusion when DOCX exports silently differ from the editor and Markdown output.

### Nit — Increase button not visually disabled at max indent (`packages/docs/src/editor/Toolbar.tsx:904`)

The increase-indent button has no `disabled` prop check for `level >= INDENT_MAX_LEVEL`. The command correctly returns `false` at max (no-op), but the button remains visually clickable. Minor UX inconsistency — the decrease button correctly disables at level 0.

## Verdict: APPROVED

Solid feature implementation with good test coverage and correct Yjs/schema migration handling. The DOCX export gap is a P2 follow-up item, not a correctness or data-loss blocker. The toolbar nit is cosmetic.

Jerry-Xin (approved):
**APPROVE** (re-anchored to the current head) — draft-gated, still `needs-human-review` for the v18 schema.

Delta re-review since my prior approve. The head advanced by two commits (no rebase drift: ahead 2, behind 0), touching exactly two files — the indent toolbar wiring and a new toolbar render test. Both are a self-contained UX/reactivity fix, no schema or export change.

**What the delta changed (verified):**
- `packages/docs/src/editor/Toolbar.tsx` — the indent buttons now key their disabled state off a dedicated `useIndentLevel` subscription (transaction + selectionUpdate) instead of reading the level once per render. This fixes the stale disabled-state bug: decrease stayed greyed after an increase and stayed lit after a decrease-to-0, because the prior snapshot keyed only off the selection while the indent command rewrites a node attribute and leaves the caret put. The increase button also now disables at the clamp ceiling, symmetric with decrease at level 0.
- `packages/docs/src/editor/IndentControls.test.tsx` (new, +139) — a full-matrix render test: decrease disabled at 0, re-enabling after increase; increase disabled at max, re-enabling after decrease; every mid level clickable and stepping one level; the disabled-at-max click is a no-op.

**v18 schema / command no-regression (byte-verified at head):**
- Write path is fully clamped: the command coerces both the current and next value through `clampIndent` and enforces bounds [0, max]. No unbounded indent. `ParagraphIndent.ts`.
- Null-sentinel storage keeps plain paragraphs attribute-free through the Y.Doc, so old documents stay byte-identical — backward-compatible, no migration.
- Round-trip preserved: the attribute rides the node attrs (Yjs), parses back from the round-trip attribute, and renders as an inline presentational style; no injection (value coerced to an integer, interpolated only as a numeric unit).
- The toolbar read is display-gating only; even though it reads the raw value, the actual mutation is clamped, so there is no correctness risk.
- No new toolbar strings introduced by the delta (reuses the existing indent-increase / indent-decrease labels), so i18n stays symmetric.

**Peer CR items (re-checked at head — unchanged by this delta, non-blocking):**
- 🟡 DOCX export omits the paragraph/heading indent. Still the case — the DOCX exporter maps only list/blockquote/task-list layout, not the v18 attribute. This is a one-way export-fidelity gap, not source data loss: the indent persists in the canonical document, round-trips, and IS preserved by the markdown/HTML export path (margin-left). No DOCX importer exists, so there is nothing to corrupt on re-import. Worth closing before GA, but non-blocking.
- 🟡 No dedicated feature flag gating the indent group (siblings like line-spacing have one). Non-blocking; the feature is self-contained and backward-compatible.

**No new issue introduced by the delta.** CI is green (no failures; several jobs skipped as expected for a docs-scoped change).

Net: the delta is a clean, well-tested reactivity/boundary fix that improves the feature. The two peer 🟡s remain open but are non-blocking. No blocking regression. Approving, re-anchored to the current head; leaving the human-review gate for the v18 schema sign-off.

yujiawei (approved):
# Code Review — PR #756 (octo-web)

**Reviewed at head SHA `1f861c9a7e7095883410733b41cd61c49670d4f7`** vs merge-base `8073e2b`.
Scope: paragraph/heading indent, `SCHEMA_VERSION` 17 → 18 (12 files, +519/-18). Classified **security-sensitive** — reviewed with extra care on the style/injection and collab boundaries.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The change delivers what the description states — a global `indent` attribute on `paragraph`/`heading`, clamp `[0,8]`, `null` empty sentinel, `data-indent` round-trip, `margin-left` render, toolbar increase/decrease, `Mod-]`/`Mod-[`, markdown/HTML export, i18n (en-US + zh-CN), and the version bump.
- **Over-build:** minor — `setIndent` and `unsetIndent` are declared and implemented in the public `Commands` interface but wired to no toolbar/keybind and (for `setIndent`) untested. Not blocking; see P2 below.
- **Deviation:** none in the implemented behavior. Registration order (`ParagraphIndent` after `LineHeight`) is consistent across both the live and preview extension sets, matching the documented style-order contract. `indent` is correctly kept out of the node/mark lists (it is an attribute), with explicit negative tests.

> Note: `docs/schema/SCHEMA-SPEC.md` §16 and the backend stub referenced by the "byte-aligned" / LOCKSTEP claim live in a separate repo and are **not present in this repo**, so byte-alignment of the attr name, `data-indent` round-trip, `null` default, clamp bounds, and the backend `toDOM` style-fragment order **cannot be verified here**. This must be confirmed against octo-docs-backend#73. In particular, if the backend schema was not also bumped to v18 with `indent`, the Yjs collab boundary will strip the attribute in production regardless of the (passing) in-memory round-trip test. Flagging for human cross-repo verification.

## 2. Code quality

**Quality: Approved**

Verification performed: installed deps, ran the indent-specific suites (`ParagraphIndent`, `IndentControls`, `extensions`, `schema`, `markdown` export) — **84/84 pass**. `tsc --noEmit` surfaces only two errors (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) that are **pre-existing on `main`** and unrelated to this change. The 4 failing `EditorShell.test.tsx` tests are likewise **pre-existing on `main`**, unrelated to indent.

### Security (clear)
No exploitable issue found. Both render paths build the style from a **clamped integer × constant** (`clampIndent(...) * INDENT_STEP_EM`), never from string-interpolated user input; `data-indent` uses `String(level)` on the same clamped integer. `parseHTML` runs any stored/pasted/collab value through `clampIndent`, so hostile inputs (`"1;color:red"`, `"9e999"`, `"</style>"`, etc.) collapse to a bounded integer or `null`. The read-only preview renders through ProseMirror's serializer (which re-clamps), not `innerHTML`, so a malicious collab peer cannot smuggle an unsanitized `indent` attribute into the rendered DOM or into export.

### P2 — non-blocking (correctness/consistency)
- **Export/import round-trip asymmetry** — `export/markdown.ts` `wrapAlign` emits `style="margin-left:Xem"` with **no** `data-indent`, while `ParagraphIndent.parseHTML` reads **only** `data-indent` (never `style.marginLeft`). No live corruption today: the editor's own `getHTML` emits both, and the markdown exporter is one-way (no re-import path exists in `packages/docs/src`). But the sibling `LineHeight` extension this PR models parses from the same inline style it renders, so pasting an external/Word/GDocs `<p style="margin-left:2em">` is silently *not* recognized as indent. Consider a `parseHTML` fallback to `element.style.marginLeft`, or a comment stating style-based import is intentionally out of scope.
- **DOCX export drops indent** — `export/docx/nodes.ts` `convertParagraph`/`convertHeading` map `textAlign` and spacing but not `indent`, so an indented block exports to `.docx` with left margin 0. DOCX supports native paragraph indent (already used for lists). One of three export formats loses fidelity; map `indent → indent: { left: twips }` or scope it out explicitly.

### P2 — non-blocking (maintainability / tests)
- `wrapAlign` now handles indent as well as align but keeps its align-only name — consider `wrapBlockAttrs`.
- The null-vs-0 sentinel rule is duplicated: `applyIndent` inlines `next > 0 ? next : null` instead of reusing `normalizeIndent(next)` (the one place the load-bearing invariant is re-implemented rather than called).
- `setIndent` is unused public API and untested; `unsetIndent` has no UI caller. Either wire/test or drop to shrink the maintained surface.
- Test gaps (cheap to close): `clampIndent(Infinity/-Infinity/''/{})` not asserted; and every command test uses a single-block doc — mixed-level multi-block selection (the reason `applyIndent` walks `nodesBetween`) is unexercised.
- `applyIndent` walks `nodesBetween` with the type filter `['paragraph','heading']`, so a paragraph **inside** a list item / table cell / blockquote also receives `margin-left`, stacking on the container's own indentation. This matches the existing `TextAlign` semantics (same configured types), so it is not a regression — but worth a deliberate product decision.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No P0/P1. All findings above are P2 (polish, export-format coverage, tests, and one cross-repo item for a human to confirm). None blocks merge; the strongest follow-ups are the backend v18 lockstep confirmation and the DOCX indent gap.

OctoBoooot:
@/tmp/pr756-update-20fd7ebc.md

OctoBoooot (approved):
## Review: paragraph/heading indent (increase/decrease) — SCHEMA_VERSION v18 (#756)

**Verdict: Approve.** Now out of draft; re-verified at head `20fd7ebcbf94`. Formal gating verdict, superseding my draft-time advisory. The v18 schema change is backward-compatible by construction, bounds-safe, and well-tested; three non-blocking 🟡 follow-ups remain. Risk tier: schema migration → `needs-human-review` (already labeled); a human signs off regardless of review votes.

### v18 backward-compat — correct by construction (verified)
The new `indent` is an **ATTRIBUTE** on `heading`/`paragraph` (not a new node/mark) — `SCHEMA_NODES`/`SCHEMA_MARKS` unchanged, `SCHEMA_VERSION` bumped 17→18. `toStorageLevel` uses a **`null` sentinel for level 0**, so a plain paragraph stays attr-free through the Y.Doc (`y-prosemirror` stores every non-null attr) → **old docs stay byte-identical, no migration** — exactly the v5 `textAlign` / v17 `lineHeight` null-default pattern. The in-comment LOCKSTEP note (attr name, `data-indent` round-trip, null default, clamp bounds must stay byte-aligned with the backend stub + SCHEMA-SPEC) correctly flags the cross-repo contract.

### Bounds — fully clamped on write (verified)
`clampIndent` coerces to integer `[0, INDENT_MAX_LEVEL=8]` with a `Number.isFinite` NaN guard + round; `increaseIndent` clamps at max, `decreaseIndent` at 0. `parseHTML`/`renderHTML` also route through `clampIndent`. No negative/NaN/unbounded path can persist.

### The UI delta (this batch of commits) — clean
`useIndentLevel` re-renders the indent buttons when the level changes without a caret move (a `useSyncExternalStore` snapshot keyed off `getAttributes().indent`, because increase/decrease rewrite a node attr and leave the caret put — same correct shape as #754's `useEditorValueTick`). Symmetric boundary disable: decrease disabled at `<= 0`, increase disabled at `>= INDENT_MAX_LEVEL`.

### Tests + CI
`ParagraphIndent.test.ts` + schema test cover clamp bounds, the null-default round-trip, and the increase/decrease boundaries; markdown export round-trips the indent (`margin-left`). All code CI green at head.

### Non-blocking follow-ups (🟡 — none blocks)
1. **DOCX exporter omits the v18 `indent` attribute** — an export-fidelity gap, **not** data-loss: the indent persists in the canonical Yjs doc, round-trips, and is preserved by markdown/HTML export; only the one-way DOCX export product omits `margin-left`, and there's no DOCX importer to read it back. (This resolves the severity of the originally-filed 🔴 — multi-reviewer convergence now grades it 🟡.) Fair fix: mirror `markdown.ts`'s `indent → margin-left` mapping in `docx/nodes.ts`.
2. **No `INDENT_ENABLED` feature flag** — the sibling features (`LINE_SPACING_ENABLED`, `FONT_FAMILY_ENABLED`) have toggles for staged rollout/rollback; the indent controls don't.
3. **Toolbar reads indent without `clampIndent`** — cosmetic-only: since write AND `parseHTML` both clamp, a stored value is always valid, so an unclamped read can at worst mis-display, never persist a bad value. Routing the read through `clampIndent` too is tidy but purely defensive.

## Parity checked
The PR's behavior-preservation claim is "old docs stay byte-identical, no migration." I verified the base→head capability diff:
- **Existing documents (no indent):** base main has no `indent` attr; at head, `toStorageLevel` stores `null` for level-0, and `y-prosemirror` only persists non-null attrs — so a plain paragraph/heading serializes **identically** to base (attr-free). No migration, no byte change to existing content. ✅
- **Schema surface:** `SCHEMA_NODES`/`SCHEMA_MARKS` are unchanged (indent is an attribute, not a new node/mark); only `SCHEMA_VERSION` bumps 17→18 — the same non-breaking shape as v5 textAlign / v17 lineHeight. ✅
- **Existing export paths:** markdown/HTML export of a doc with no indent is unchanged (the `wrapAlign`/margin-left branch only fires for a non-zero indent); DOCX export is unchanged for existing content (it doesn't map the new attr — the 🟡 below, but that means *no* change to existing DOCX output, not a regression). ✅
- **New capability added (not altering old):** increase/decrease indent on heading/paragraph, bounded [0,8]. List Tab/Shift-Tab sink/lift is untouched. ✅
Net: the change is purely additive (a new optional attribute + toolbar controls); no existing document, schema node/mark, or export path changes behavior. The preservation claim holds.

### Net
Clean v18 feature: backward-compat by construction (null sentinel, no migration), fully clamped on write, attribute-not-node schema change with backend lockstep noted, well-tested, CI green. **Approve** @ `20fd7ebcbf94`. Merge remains gated by `needs-human-review` (v18 schema migration — human sign-off required, not a code issue) — the three 🟡s (DOCX indent→margin, feature flag, read-clamp) are non-blocking follow-ups. Credit ploy-elison — the null-default no-migration reasoning is exactly right for a v-bump, and the `useIndentLevel` attribute-keyed re-render matches the #754 pattern.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #756 (octo-web)

## Summary

This PR adds paragraph/heading indent (increase/decrease) as SCHEMA_VERSION 18 — a global `indent` attribute on the `heading` and `paragraph` nodes, rendered via `margin-left` and round-tripped as `data-indent`. The implementation includes a new Tiptap extension (`ParagraphIndent`), toolbar buttons with reactive disabled-state tracking (`useIndentLevel`), keyboard shortcuts (`Mod-]`/`Mod-[`), markdown export support, i18n strings, and a schema version bump. The approach mirrors the established `textAlign` / `lineHeight` pattern: attribute-only change, null default for backward compatibility, no new node/mark.

Overall: clean, well-tested, and follows the repo's established conventions closely. One minor P2 inconsistency noted below; no blockers.

## Verification

- ✅ **Extension registration parity** — `ParagraphIndent` registered in both `buildExtensions` (live editor) and `buildPreviewExtensions` (read-only preview), both configured for `['heading', 'paragraph']`, both placed AFTER `LineHeight` to preserve style declaration order (`packages/docs/src/editor/extensions.ts:207` and `:328`).
- ✅ **Command boundaries** — `applyIndent` returns `false` on no-op (decrease at 0, increase at max), so the toolbar's `.run()` reflects the boundary correctly. Verified by both `ParagraphIndent.test.ts` and `IndentControls.test.tsx`.
- ✅ **Toolbar reactivity** — `useIndentLevel` subscribes to `transaction` + `selectionUpdate` via `useSyncExternalStore`, keying off the indent level itself rather than just the selection range. This avoids the stale-disabled-button bug documented in the code comments (same class as the find-counter fix). Verified by `IndentControls.test.tsx` across the full 0→max→0 matrix.
- ✅ **Null default / backward compat** — `indent` defaults to `null`, so `renderHTML` returns `{}` for plain paragraphs (no `data-indent`, no `style`). Old documents without the attr render identically. `parseHTML` handles missing `data-indent` via `normalizeIndent(null)` → `null`.
- ✅ **List isolation** — `listItem` nodes never gain the `indent` attr (verified by `extensions.test.ts`). Tab/Shift-Tab sink/lift remains owned by list extensions. Keyboard shortcuts use `Mod-]`/`Mod-[` to avoid shadowing.
- ✅ **Yjs collab round-trip** — `ParagraphIndent.test.ts` verifies the attr survives a two-peer Y.Doc sync, confirming the schema is compatible with the collab boundary.
- ✅ **Schema version bump** — `SCHEMA_VERSION` 17→18, comments updated, test assertions updated. `indent` correctly excluded from `SCHEMA_NODES`/`SCHEMA_MARKS` (it's an attribute, not a node/mark).
- ✅ **Markdown export** — `wrapAlign` now handles both `textAlign` and `indent`, producing a combined wrapper tag when either is present. Backward-compatible: blocks with neither are emitted as-is.

## Findings

No P0/P1 issues. One P2 inconsistency below.

### P2 — CSS `margin-left` space inconsistency (`packages/docs/src/export/markdown.ts:214`)

The editor's `renderHTML` in `ParagraphIndent.ts:88` produces `margin-left: 4em` (with a space after the colon), but the markdown export here produces `margin-left:4em` (no space). Both are valid CSS and parse identically, but the code comments in both files emphasize byte-alignment with the editor's render. Add the space for consistency:

```ts
indent > 0 ? ` style="margin-left: ${indent * INDENT_STEP_EM}em"` : '',
```

Not a functional defect — just a cosmetic inconsistency that could surprise byte-compare assertions in round-trip or snapshot tests.

- `clampIndent` handles all edge cases: NaN, undefined, null, strings, out-of-range numbers, floats (rounds).
- `normalizeIndent` correctly returns `null` for level 0 (Y.Doc backward compat) and positive integers for levels 1+.
- `Btn` component passes `disabled` to the native `<button>` element — disabled buttons are truly non-interactive.
- `onMouseDown={(e) => e.preventDefault()}` on `Btn` prevents caret movement on click.
- `escapeHtmlAttr` is correctly applied to the `align` attribute value in `wrapAlign`; the indent value is always a clamped number, so no injection vector.
- `node.attrs?.indent` with optional chaining is safe when `attrs` is undefined.
- i18n keys added for both `en-US` and `zh-CN` locales.
- Preview extensions mirror the live extensions (same types, same registration order).

## Verdict: APPROVED

Solid implementation that follows the repo's established patterns for attribute-only schema additions. The `null` default, the `clampIndent` boundary guards, the `useSyncExternalStore` reactivity fix, and the comprehensive test coverage (command boundaries, toolbar reactivity matrix, Yjs collab round-trip, markdown export) all demonstrate careful engineering. The single P2 is a one-character whitespace fix.

[Octo-Q] verdict: APPROVE — no P0/P1 blockers; one P2 cosmetic inconsistency (CSS margin-left space in markdown export).

yujiawei (approved):
# Code Review — PR #756 (octo-web)

Reviewed at head SHA `20fd7ebcbf9478ddd77fbd718f445bf31b4cf9b5`. This adds a paragraph/heading `indent` global attribute (SCHEMA_VERSION 17 → 18), rendered via an inline `margin-left`, round-tripped through `data-indent`, with toolbar increase/decrease controls and markdown/HTML export support. It is flagged security-sensitive, so I focused on the CSS-injection surface, the value-coercion path, the collab boundary, and backward compatibility.

## Verdict: Approved

No correctness, security, or data-loss blockers found in the diff. The change follows the established v5 `textAlign` / v17 line-spacing attribute pattern faithfully. The findings below are non-blocking hardening / test-coverage / release-coordination notes.

## Spec / scope

The diff implements exactly one feature — indent inc/dec on paragraph + heading as a global attribute — with no scope creep. No new node/mark is added (correctly kept out of `SCHEMA_NODES`/`SCHEMA_MARKS`); `list` items are deliberately untouched so Tab/Shift-Tab sink/lift stays owned by the list extensions. Keyboard binding is `Mod-]` / `Mod-[`, and I confirmed no other extension binds those keys. i18n keys are added for both `zh-CN` and `en-US`.

## Security review (the reason this was flagged)

The one attacker-influenceable output is the inline `margin-left` style string. I traced every path that reaches it:

- `ParagraphIndent.ts` `renderHTML` (line ~85) computes `level = clampIndent(attributes.indent)` **before** interpolating, and `clampIndent` (lines 33-37) coerces any value — string, `NaN`, `null`, float, out-of-range — to an integer in `[0, 8]`. The emitted value is therefore always `margin-left: <0..16>em`, which cannot carry HTML/CSS metacharacters.
- `markdown.ts` `wrapAlign` (line 210) likewise runs `clampIndent(node.attrs?.indent)` before line 215 interpolates it.
- `parseHTML` reads `data-indent` from untrusted HTML through `normalizeIndent` → `clampIndent`, so pasted/imported markup is bounded on the way in as well.

Conclusion: **no CSS-injection or attribute-breakout vector** in the numeric indent path. The value is clamped at the parse boundary, the render boundary, and the export boundary.

**P2 — `margin-left` style value is not run through `escapeHtmlAttr` in markdown export** (`packages/docs/src/export/markdown.ts:215`)
Every other inline-style attribute in this file is escaped (`align` on line 214; `background-color` on 489; `color` on 497), and the file header states inline-HTML attributes must be escaped for defense-in-depth against downstream Markdown→HTML renderers. The indent value is provably numeric (`clampIndent` × constant), so `escapeHtmlAttr` is a no-op today and there is **no live vulnerability** — but wrapping it keeps the defensive pattern consistent and robust against a future change to the indent formula.

**P2 — `useIndentLevel` reads the raw attr without `clampIndent`** (`packages/docs/src/editor/Toolbar.tsx:888-889`)
`Number(editor.getAttributes('paragraph').indent) || 0` (and the heading equivalent) bypass `clampIndent`, unlike every other read site. A corrupted/hostile Y.Doc value such as `indent: 999` would only mis-drive the toolbar's disabled state (increase greyed, decrease lit) — it cannot corrupt data, since the commands themselves re-clamp before writing and `renderHTML` re-clamps before painting. UI-only, but normalizing untrusted values at this boundary too would be tidier.

**P2 — no test asserts the 5-property style order across TextAlign + LineHeight + ParagraphIndent** (`packages/docs/src/editor/extensions.ts:200-212`)
The byte-alignment contract with the backend `toDOM` depends on `margin-left` being appended last. `LineHeight.test.ts` verifies the 4-property order but no test combines all three extensions to prove `text-align; line-height; margin-top; margin-bottom; margin-left`. Since backend drift here is a silent HTML round-trip risk, a combined-order test (mirroring the existing `styleOrder()` test) would be worth adding.

**P2 — adversarial round-trip test gaps.** `clampIndent` is well unit-tested in isolation, but there's no integration test that a hostile `data-indent` (e.g. `"9999"`, `"-1"`, non-numeric) is clamped through `parseHTML`, nor that an out-of-range value arriving via the Y.Doc is clamped on render (the collab test only exercises the always-valid `increaseIndent` path). The code is safe; these would lock the behavior in, matching `LineHeight.test.ts`'s "drops an invalid value at render time" test.

**P2 — no feature-flag kill switch for the indent toolbar.** v17 line-spacing shipped behind `LINE_SPACING_ENABLED` (`Toolbar.tsx:1402-1404`); `IndentControls` renders unconditionally (`Toolbar.tsx:1405`). Adding a symmetric flag would give the same phased-rollout / emergency-off lever. Non-blocking — a policy choice, not a defect.

## Items for a human to verify before/at merge

- **SCHEMA_VERSION 18 backend lockstep.** The code correctly documents that the `indent` attr must be byte-aligned with the backend stub + SCHEMA-SPEC at v18, or the collab boundary will strip the unknown attr and silently drop indent on peers. This cannot be verified from a frontend-only diff. Please confirm the backend stub is at v18 (or that indent is a known attr there) before this reaches collaborative production traffic. This is the single highest-risk aspect of shipping — but it is a cross-repo release-coordination item, not a defect in this diff, and it applies identically to every prior attr-only version bump (v5/v7/v16/v17).

## Backward compatibility

Sound by construction: the default is `null` (not `0`), so plain paragraphs stay attr-free through the Y.Doc and old documents are unchanged with no migration. `renderHTML` returns `{}` at level 0, emitting no `data-indent` and no style — confirmed by the level-0 test asserting neither appears.

## Test coverage

Good breadth: command boundaries (no-op at 0, clamp at max), HTML `data-indent` round-trip, a Yjs collaboration round-trip proving the attr survives the collab boundary, markdown export (indent alone, indent+align combined, heading→div, level-0 backward-compat), toolbar reactivity across the full 0→max matrix, and schema membership assertions. The gaps noted above are adversarial-input and cross-extension-order cases, not core-behavior gaps.

PR #757, opened 2026-07-14
feat(loop): loop UI overhaul and prod primary-color fix

### LINKED ISSUE #703: Bug: dmloop primary color renders as Semi default blue in prod build (dev matches Figma, prod doesn't)
## Summary
In the dmloop module, primary-colored UI (top-right "New Loop / Create" solid buttons, left sidebar hover, and anything driven by `--semi-color-primary`) renders as Semi Design's default bright blue `rgb(0,100,250)` (#0064FA) in the **production build**, instead of the brand black `#1C1C23` defined by our theme. The dev server renders correctly (brand black, matches Figma); only the prod bundle is wrong. This is a global regression that dmloop surfaces first, because it is the first module to lean heavily on raw Semi `<Button theme="solid">`.

## Root cause
`--semi-color-primary` is declared by same-specificity `body`-level rules:

- Semi default: `body { --semi-color-primary: rgba(var(--semi-blue-5),1) }` (bright blue) — from `@douyinfe/semi-ui/dist/css/semi.min.css`.
- Our override: `packages/dmworkbase/src/App.css` `body { --semi-color-primary: var(--wk-color-theme,#1C1C23) }` (brand black).

Both selectors are `body` (specificity 0,0,1), so the winner is decided purely by **CSS load order**:

- **dev (Vite)**: app theme CSS is injected after Semi CSS -> brand black wins.
- **prod (Rollup bundling)**: Semi CSS is concatenated after the theme override -> Semi blue wins -> global `--semi-color-primary` becomes blue.

Dark mode is unaffected because its override uses `body[theme-mode=dark]` (specificity 0,0,2), which beats Semi's plain `body` regardless of order. Only the light-mode `body` override is order-sensitive.

dmloop's skin layer (`packages/dmloop/src/ui/loopControls.css`) only skins inputs/selects, not Semi `<Button>`, so loop buttons fall through to the global `--semi-color-primary` and expose the regression directly.

## Steps to reproduce
1. Run `pnpm build` and serve the production bundle (or open the prd env).
2. Open the dmloop module in light mode.
3. Look at the top-right "New Loop / Create" button and the left sidebar hover.

## Expected
Primary color = brand black `#1C1C23` (matches Figma), consistent between dev and prod.

## Actual
Primary color = Semi default bright blue `rgb(0,100,250)` in the prod build; correct (black) only in dev.

## Proposed fix
Make the light-mode override beat Semi's `body` regardless of bundle order by raising specificity from `body` (0,0,1) to `html body` (0,0,2) in `packages/dmworkbase/src/App.css`:

```css
html body {
  --semi-color-primary: var(--wk-color-theme, #1C1C23);
  --semi-color-primary-hover: #2A2A33;
}
```

This is a global theme fix (affects all Semi primary usage, not only dmloop).

Optional, spec-aligned hardening (per the loop design spec: "Semi overrides scoped within the loop root"): also skin loop's solid primary buttons inside the loop scope in `loopControls.css` so they never depend on the fragile global cascade.

Note: `--semi-color-primary-light-*` variants (used for hover tints) are not overridden by our theme at all, so hover background tints stay Semi-blue in both dev and prod — a separate, pre-existing issue outside this regression.

## Environment
- Repo: octo-web, branch `feat/octo-loop`
- Semi UI: `@douyinfe/semi-ui@2.93.0`
- Build: Vite + Turborepo (prod = Rollup bundle)
- Affected: light mode, production build only

superalsrk:
## Goal
Make the dmloop / dmpersonal loop surfaces consistent and polished, and fix a production-only theming regression.

## Changes
- Fix the light-mode primary color falling back to the default bright blue in prod builds: raise the theme override to `html body` so it wins regardless of bundle order, and pin hover/active/light tints to the brand black so buttons and focus halos never flash blue (App.css).
- Add a muted color-spec token layer + `LoopTag` semantic tag, replacing scattered raw `<Tag color>` candy colors; brightness tuned for a soft, non-plastic look.
- Add a native `LoopButton` (6px radius, brand black, same-hue hover/active) and migrate all solid primary buttons off framework defaults.
- Unify "New Loop" creation into a single modal (removing the standalone one-sentence AI dispatch page, which belongs to the host system, not this module): breadcrumb + title + description + status/priority pill pickers + three-state assignee + attachments, with default-agent selection remembered across sessions.
- Rework the "Add Computer" guide into one copyable Bot install prompt (install → configure permissions → restart), using the current access host as `--server-url` and a one-time credential minted on copy.
- Turn the project list into an inline-editable table (name / status / priority / progress / lead / created); status, priority and lead are editable via dropdowns without triggering row navigation.
- Clicking the left "loop" nav now lands on the list and refetches data.

## Verification
- `npx tsc --noEmit --skipLibCheck -p packages/dmloop/tsconfig.json` — clean on changed files.
- `npx stylelint "packages/dmloop/src/**/*.css"` — only pre-existing hex warnings, no errors.
- `npx vitest run` for `headlessCommand.test.ts` — passing.
- Manual (pnpm dev): new-loop modal from both entry points, add-computer prompt copy, project table inline edits, loop-nav refresh, light-mode primary color.

Closes #703


Jerry-Xin (approved):
**APPROVE** — clean UI overhaul + prod primary-color fix. No blocking issues found. Full independent review at head `10875cb4`.

### ✅ Highlights
- **No XSS surface introduced.** The new presentational components (LoopButton, LoopTag, LoopPropertyPill, ProjectStatusBadge, ProjectPriorityBadge) all render via JSX children only — no raw-HTML render sink anywhere in the diff. The overhaul kept the render path safe.
- **Shell-injection defense added.** The extracted `headlessCommand.ts` wraps the token and backend URL through POSIX single-quote escaping before interpolating into the copy-paste install command, so a malformed server-supplied value cannot break out of its argument.
- **Genuinely behavioral tests.** `headlessCommand.test.ts` asserts the quoted output byte-for-byte and includes a case proving an injected quote is neutralized — not a "was-called" stub.

### 🟢 Prod primary-color fix
The `html body` token block (issue #703) is a correct CSS-specificity fix: raising specificity so brand tokens win regardless of prod bundle concatenation order. Only brand color values are set; no environment/secret/deployment detail is embedded. Config-clean.

### 🟢 Refactor-deletion audit (−766)
- NewLoopPage and the `quick-create` API path are fully removed and consolidated into CreateIssueModal via the standard create path; all imports/usages are removed in the same change with no dangling references.
- The old dispatched→debounce→refresh two-hop event mechanism is replaced by a direct deferred refresh emit on create success — simpler, equivalent coverage.
- No auth check, sanitization, error handling, or cleanup logic was dropped in the deletions.

### 🟢 Other checks
- i18n: all CJK strings live in the locale JSON value positions; no hardcoded user-facing strings leaked into component code.
- `http.ts` workspace-context change is additive and backward-compatible (optional `name` field with safe parse + careful clear semantics).
- No console/debugger leftovers, no feature-flag flips, no new dependencies. Dependency-review, osv-scan, and secret-scan all pass.

Nice consolidation and the added quoting test is a real improvement. LGTM.

OctoBoooot (commented):
## Review: #757 — dmloop UI overhaul + prod primary-color fix

**Verdict:** Comment (not a blocker set, but not a clean approve either — a governance CI check is red and there's a cluster of minors worth addressing). Large UI overhaul (38 files, +1051/-766); I fanned out across the credential flow, theme, project table, and new-loop modal. Risk tier normal, but size/XL → `needs-human-review` applied. **CI: the only red check is `check-sprint` — a *governance* check (a maintainer must set the Sprint field; code can't turn it green), not a code failure.** All *code* CI is green (Build, secret-scan, osv-scan, dependency-review, pr-title-lint). Per the code-vs-governance CI rule, the governance red doesn't bar approval on its own, but `mergeStateStatus` stays BLOCKED until it clears — so I'm holding at COMMENT and naming it rather than approving over an unaddressed governance gate. Closes #703.

### Security — credential-mint flow (Add Computer) is sound
The one-time PAT flow (`RuntimePage.tsx`) is well-designed: the PAT is minted only on explicit copy-click (`onCopyInstall`), not at modal render (preview shows a placeholder); `mintingRef` prevents double-issue and reuses an already-minted prompt (no credential accumulation); the token is written to clipboard via `copyToClipboard`, never logged (the catch emits a generic toast, no token); and `closeAddDialog` clears the minted prompt so reopening doesn't re-render a stale PAT. `serverUrl` is `window.location.origin` (browser-derived, fine). Informational: the minted PAT does live in component state / the rendered prompt box for the modal's open lifetime (inherent to a copy-paste install flow, and cleaned on close) — acceptable, just noting it's a real short-lived credential in the DOM.

### Theme override — app-wide scope is pre-existing, NOT a new bleed
The App.css change raises `--semi-color-primary` from `body` to `html body`. Verified against the removed side of the diff: the app-wide `--semi-color-primary: var(--wk-color-theme, #1C1C23)` was **already** at `body` scope — this PR only bumps specificity (to win over Semi's `body` rule in prod bundle order — the #703 flash-blue-in-prod bug) and adds hover/active/light tints. So the global scope is pre-existing and intentional (that's the stated prod fix), not a regression. No finding.

### Minors (worth fixing, none blocking)
- **a11y — `.loop-btn` has no focus indicator** (`loopControls.css`): `LoopButton` renders a native `<button>` with `outline: none` and no `:focus-visible` replacement, so keyboard focus is invisible on these buttons — a WCAG-2.1 gap. Add `.loop-btn:focus-visible { outline: 2px solid var(--semi-color-primary); outline-offset: 2px }`.
- **Project table — no optimistic rollback** (`ProjectPage.tsx:patchProject`): an inline status/priority/lead edit awaits the API then `reload()`s; on API failure the error toast fires but the row isn't rolled back — the UI shows the picked value until the next reload (recoverable, not data-loss). Consider optimistic-update-with-rollback, or at least re-sync the row on error.
- **Cross-workspace assignee bleed** (`CreateIssueModal.tsx`): `LAST_ASSIGNEE_KEY = "loop.newloop.assignee"` is not workspace-scoped, so the remembered default-agent is shared across workspaces. The validity check (`cands.find(id === last.id)`) prevents a crash/wrong-pick in practice, so it's minor — but scope the key to the workspace id (`…:${currentWorkspaceId()}`) to be clean. (This is the loop-global-state class from prior loop reviews.)
- **Stale fallback** (`loop.css:1012`): `var(--semi-color-primary, #3b82f6)` — the fallback is Semi blue, contradicting the brand-black theme; should be `#1C1C23` so a var-miss doesn't flash blue.

### Removed-page parity — clean
The standalone one-sentence AI-dispatch `NewLoopPage` + `quickCreateIssue` API were removed cleanly (no dangling imports/routes — verified), both entry points now open the unified `CreateIssueModal`, and the sync-create refactor correctly replaced the async-dispatch settle-timer/event machinery. The dropped one-sentence-dispatch capability is intentionally delegated to the host system (per the PR); the new modal is a superset (adds member assignee, attachments). No silent capability loss.

### Net
No blockers, no majors: the credential flow is sound, the theme scope is pre-existing-not-new, and the page removal is parity-clean. Holding at COMMENT for (a) the red `check-sprint` governance gate — a maintainer must set the Sprint field before merge — and (b) a cluster of minors (button focus a11y, project-edit rollback, workspace-scoped assignee key, the stale blue fallback). Address the minors + clear check-sprint and this is an easy approve. `mergeStateStatus` stays BLOCKED until the Sprint field is set + `needs-human-review` is cleared. Credit superalsrk.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review: octo-web PR #757
**Repo:** Mininglamp-OSS/octo-web
**Head SHA:** `10875cb481e13ec70e2da647365911654ee218f9`
**Verdict:** APPROVED

---

## Summary

PR #757 is a comprehensive dmloop UI overhaul that:
1. Replaces Semi UI components with custom `Loop*` primitives (LoopButton, LoopTag, LoopPropertyPill, ProjectStatusBadge, ProjectPriorityBadge)
2. Migrates `NewLoopPage` (standalone composer) → `CreateIssueModal` (inline modal with attachments, assignee memory)
3. Adds inline project editing on ProjectPage (status/priority/assignee/lead via table cells with `stopPropagation`)
4. Unifies the runtime install prompt (removes old daemonServerUrl guard + per-OS headless commands)
5. Adds attachment support to issue creation (`attachment_ids` in `CreateIssueReq`)

**Overall assessment:** Well-structured, clean separation of concerns. Data flow tracing confirms correct upstream propagation for all major paths. No P0/P1 blockers found.

---

## Verification Conclusions

| Area | Status | Evidence |
|------|--------|----------|
| NewLoopPage removal | ✅ Clean | `grep -rn "NewLoopPage" packages/dmloop/src/` → only sessionStorage key `loop.newloop.assignee` remains (intentional) |
| Event bus cleanup | ✅ Clean | `wk:loop-issues-dispatched` has zero emitters (confirmed grep); `wk:loop-issues-refresh` correctly emitted/subscribed |
| `getDaemonServerUrl` removal | ✅ Clean | Only export remains in `authApi.ts` (acceptable); no production consumers |
| `ISSUE_STATUS_COLOR` / `PRIORITY_COLOR` | ✅ Not broken | Still used by `IssueDetailPage` (verified grep hits) |
| `patchProject` PUT spread | ✅ Correct | All fields explicitly included; `...partial` overrides last |
| `AssigneePicker` signature compat | ✅ Compatible | Component always calls `onChange(id, type, name)` (3 args); ProjectPage callback `(id, type) => ...` ignores 3rd arg (valid JS) |
| Double-submit guard | ✅ Triple-layered | `submit()` guard, button `disabled`, `loading` prop |
| `INSTALL_SCRIPT_CMD` URL | ✅ Expected | `codex.mlamp.cn` = internal GitLab instance; not a public leak |

---

## Findings

### P2-1: Orphaned attachments on partial upload failure

**Path:** `packages/dmloop/src/ui/CreateIssueModal.tsx:95-105`
**Diff-scope:** New code (this PR)
**Severity:** P2 (not blocking)

**Description:** When multiple files are attached and one upload fails mid-batch, the already-uploaded files are orphaned server-side: they exist in storage but are never bound to any issue (the submit aborts via `return`). The user sees a toast ("X files failed to upload") but has no way to recover or clean up the successful uploads.

**Data flow trace:**
- `addFiles()` → `pendingFiles` state (in-memory File objects, no server-side id yet)
- submit: for each `f` in `pendingFiles`, `uploadAttachment(f)` (no issueId, issue doesn't exist yet) → server returns `{id}`
- on failure: `failed++`, toast, `return` (abort)
- successfully-uploaded `Attachment.id`s are in the loop-scoped `ids` array which is discarded on abort

**Impact:** Low. Server accumulates unbound files; users rarely attach many files; likely has server-side GC/TTL.

**Suggestion:** On partial failure, either (a) retry only the failed files, or (b) call a cleanup API to delete orphaned attachment IDs before aborting.

---

### P2-2: Orphaned `headlessCommand()` function and `START_CMD` verb mismatch

**Path:** `packages/dmloop/src/pages/headlessCommand.ts:22,28`
**Diff-scope:** Pre-existing function now orphaned by this PR's RuntimePage changes
**Severity:** P2 (not blocking)

**Description:** `headlessCommand()` is no longer called from production code — RuntimePage uses the three-constant flow (`INSTALL_SCRIPT_CMD` → `authCommand()` → `START_CMD`). The old function is still exported and covered by `headlessCommand.test.ts` (all pass), but tests now validate dead production code, creating a false sense of coverage.

**Verb mismatch:**
- `headlessCommand()` (old): `octo-daemon daemon start`
- `START_CMD` (new): `octo-daemon daemon restart`

`restart` is more robust (idempotent — works whether running or stopped), so the new constant is correct.

**Suggestion:** Remove `headlessCommand()` and update tests to verify the new three-constant flow (assert `INSTALL_SCRIPT_CMD` URL, `authCommand` flags, `START_CMD` verb).

---

### P2-3: Unused i18n key; hardcoded placeholder not localized

**Path:** `packages/dmloop/src/i18n/en-US.json` + `RuntimePage.tsx:42`
**Diff-scope:** New code (this PR)
**Severity:** P2 (not blocking)

**Description:** `runtime.installTokenPlaceholder` ("Click \"Generate token\" to fill in automatically") is defined in both `en-US.json` and `zh-CN.json` but never referenced in code. Instead, `buildPrompt()` hardcodes `"<paste-token-here>"` directly in the template string — meaning the placeholder is not localized for Chinese users.

**Suggestion:** Use `t('runtime.installTokenPlaceholder')` in `buildPrompt()` (pass `t` as a parameter), or remove the unused keys to avoid confusion.

---

## Data Flow Tracing

### 1. Attachment upload lifecycle (CreateIssueModal)

```
User attaches files
  → addFiles(f) → pendingFiles state (File objects, in-memory only)
  → submit(): for each f in pendingFiles
    → uploadAttachment(f) [no issueId] → POST /upload-file → server returns {id}
    → ids.push(attachment.id)
  → createIssue({ ..., attachment_ids: ids })
  → onCreated() → close modal → emit wk:loop-issues-refresh
```

**Verified:** Data flows correctly from upload → issue creation. No data loss on success path. Partial failure orphans successful uploads (see P2-1).

---

### 2. `patchProject` PUT semantics (ProjectPage)

```
User changes project field (status/priority/lead/assignee)
  → patchProject(p, partial)
  → updateProject(p.id, { title: p.title, description: p.description, icon: p.icon,
                          status: p.status, priority: p.priority,
                          lead_type: p.lead_type, lead_id: p.lead_id,
                          ...partial })
  → PUT /projects/:id (full upsert)
  → afterDirectoryMutation() → invalidate directory cache
```

**Verified:** All fields explicitly spread; `...partial` last so overrides apply. PUT semantics preserved (no field clearing).

---

### 3. Install prompt token flow (RuntimePage)

```
Dialog opens
  → tokenRef.current = null, showToken = false
  → buildPrompt(null) → "...--token <paste-token-here>..."
  → preview shown with placeholder

User clicks "Copy install instructions"
  → onCopyInstall()
  → if (!tokenRef.current && !mintingRef.current)
    → mintingRef = true
    → issueHeadlessCliToken() → server mints PAT
    → tokenRef.current = token
    → showToken = true
    → buildPrompt(token) → "...--token ${token}..."
  → copyToClipboard(cmd)
  → mintingRef = false
```

**Verified:** Sequential token issuance guarded by `mintingRef`. `serverUrl = window.location.origin` always present in browser context.

---

| Check | Status | Notes |
|-------|--------|-------|
| C1 — Dual-path parity | N/A | No paired add/remove/subscribe/unsubscribe changes |
| C2 — Control-flow ordering | ✅ Clear | `patchProject` spread order verified (partial last); no nested/recursive call sites |
| C3 — Authorization boundary | N/A | No new endpoints/tools exposed; existing auth unchanged |
| C4 — Authorization lifecycle | N/A | No auth/status cascade changes |
| C5 — Build/note ≠ runtime | ✅ Clear | No build/packaging changes; all changes are runtime UI code |
| C6 — Governance/policy | N/A | No docs/policy changes |

---

## Cross-round Blocker Re-check (R6)

**Status:** N/A (first review of this PR)

---

## Additional Observations

1. **Assignee memory via sessionStorage:** `CreateIssueModal` persists `assignee_id`/`assignee_name` in `sessionStorage` key `loop.newloop.assignee`. This survives page refresh within a session but clears on new tab/browser close. Intentional design (not a bug).

2. **`project_id` removed from issue creation:** The modal no longer lets users pick a project when creating an issue. Deliberate UI simplification (issues created from project context already have the project pre-selected via parent component).

3. **Settle timer removed:** `LoopPage.tsx` removed `SETTLE_DELAYS_MS` and the `wk:loop-issues-dispatched` listener. Confirmed zero emitters for that event — clean removal.

4. **CSS additions:** `LoopButton.css`, `LoopTag.css`, `ProjectStatusBadge.css` — all new stylesheets for the custom components. No concerns.

---

[Octo-Q] verdict: APPROVE — No P0/P1 blockers. Three P2 findings (orphaned attachments on partial upload, orphaned headlessCommand function, unused i18n key) are non-blocking and can be addressed in follow-up.

OctoBoooot (approved):
## Updating my #757 review COMMENT → APPROVE — the `check-sprint` governance gate has cleared

At my initial review, `check-sprint` (governance CI) was red, so I held at COMMENT per the code-vs-governance-CI rule (name the gate, don't approve over it). It has since re-run **green** — latest run (12:30:46) is `success`, superseding the earlier `failure` (a maintainer set the Sprint field). `mergeStateStatus=CLEAN`, `reviewDecision=APPROVED`. The only thing keeping me below APPROVE is resolved, so I'm upgrading to APPROVE. Converges with Jerry-Xin's APPROVE (the earlier divergence was a timing artifact — the gate flipped between our checks).

### Verdict basis (unchanged from my review, now unblocked)
The code was clean on every risk area I verified: the credential-mint flow is sound (PAT minted-on-copy only, double-issue guarded, not logged, cleared on modal close), the `html body` theme override is a pre-existing app-wide scope with only a specificity bump (not a new global bleed — verified against the diff's removed side), and the removed-page consolidation is parity-clean (no dangling refs). No blockers, no majors.

### Carried non-blocking minors (fold-in polish, do not gate)
These were in my COMMENT and remain worth addressing, but none blocks:
- **a11y** — `.loop-btn` sets `outline: none` with no `:focus-visible` replacement → keyboard focus invisible on the new native buttons. Add `.loop-btn:focus-visible { outline: 2px solid var(--semi-color-primary); outline-offset: 2px }`.
- **Project table** — `patchProject` awaits then `reload()`s with no rollback; a failed inline edit leaves the row showing the picked value until reload. Optimistic-with-rollback or re-sync on error.
- **Cross-workspace assignee** — `LAST_ASSIGNEE_KEY` isn't workspace-scoped (validity-checked, so no crash — minor); scope it to `currentWorkspaceId()`.
- **Stale fallback** — `loop.css:1012` `var(--semi-color-primary, #3b82f6)` — the fallback is Semi blue; should be brand black `#1C1C23`.

### Net
Upgrading COMMENT → APPROVE now that `check-sprint` is green and `mergeStateStatus=CLEAN`. The four minors are non-blocking fold-ins. Merge remains gated by `needs-human-review` (size/XL — human maintainer sign-off), not by any code or CI issue. Credit superalsrk.

yujiawei (changes requested):
# Code Review — PR #757 (octo-web)

Reviewed at head SHA `10875cb481e13ec70e2da647365911654ee218f9` against `main` (merge-base `8073e2b`). This is a security-sensitive PR, so the credential/install flow got extra scrutiny. I checked out the branch, verified the build claims, and traced the runtime data flow across files (not just the diff hunks).

## Verification (reproduced locally)
- `npx tsc --noEmit --skipLibCheck -p packages/dmloop/tsconfig.json` — **clean on all 31 changed `.ts/.tsx` files** (the repo baseline has pre-existing errors elsewhere / in `node_modules`, none in this PR's files).
- `npx vitest run packages/dmloop/src/pages/__tests__/headlessCommand.test.ts` — **4/4 passing**.

## Spec compliance ✅
Every item in the description is implemented, and the removals are documented in the commit history, so I don't count them as undocumented scope changes:
- Prod primary-color fix (`html body` override), `LoopTag`/`LoopButton`/color tokens, single New-Loop modal, one-prompt Add-Computer flow, inline-editable project table, loop-nav refresh — all present.
- The deletions (`NewLoopPage`, `quickCreateIssue`/`QuickCreateReq`, the create-time project picker) are intentional and announced (commit `e81763a0`: "labels and project dropped"). Project assignment is preserved in the issue detail panel, so it's a relocation, not a regression. No dangling prod imports to the removed symbols.

## Code quality — Changes requested

### 🔴 Blocking

**1. A credential labeled "one-time" is, per the only available evidence, a 90-day access token.** `packages/dmloop/src/i18n/en-US.json:112` / `zh-CN.json:112` — the new install hint reads *"contains a one-time credential — do not share"* (and `headlessCommand.ts:22` comments *"用一次性 PAT"*). But the only in-repo signal about this token's lifetime is `packages/dmloop/src/api/authApi.ts:7-8`:

```ts
// headless 免浏览器接入:后端当场为当前用户签一个 90 天 PAT,
export function issueHeadlessCliToken(): Promise<{ token: string }> {
  return httpPost<{ token: string }>("/cli-token/headless");
}
```

There is **no client-side single-use / consume / revoke** logic — `onCopyInstall` mints once and deliberately caches/reuses the token for the dialog session (`RuntimePage.tsx:130-155`). The UI then instructs the user to paste this blob (which interpolates the live token via `authCommand`, `RuntimePage.tsx:125`) into an AI/"Bot" to auto-execute. So a full-account credential that lives up to 90 days transits an AI tool chain, while the copy tells the user it's "one-time". That materially understates the blast radius and changes the user's risk decision — exactly the kind of security-labeling accuracy issue that should be fixed on a security-sensitive change.

*Fix:* either mint a genuinely short-lived/single-use token server-side, or correct the wording to state the real lifetime (e.g. "temporary access token, valid ~90 days"). Please confirm the actual TTL / single-use semantics of `POST /cli-token/headless` with the backend before shipping the "one-time" wording. **(Requesting a human to verify the backend token semantics.)**

### 🟡 Non-blocking (P2 — advisory)

**2. Supply-chain hardening for the install command.** `headlessCommand.ts:19-20` — `INSTALL_SCRIPT_CMD` is `curl -fsSL https://codex.mlamp.cn/.../-/raw/main/install.js | node`: a mutable `main` ref, no commit pin, no integrity/sha256, piped into `node`, and (per the design) auto-executed by an AI. `curl | node` is industry-common and this isn't a defect introduced-as-broken, but combining an unpinned remote installer with an embedded live credential in one auto-executed prompt is worth a human's explicit sign-off. Consider pinning `install.js` to a commit SHA or signed release with an integrity check.

**3. Trust-boundary change on `--server-url`.** `RuntimePage.tsx:122` now derives the daemon `--server-url` from `window.location.origin` instead of the backend-configured `daemon_server_url` (`getDaemonServerUrl`). This is reasonable (daemon connects to the site the user is actually on), but it means the value baked into the credentialed command follows the browser origin. Worth a conscious confirmation that all access hosts are trusted. (Shell-injection is not a concern here — `authCommand` single-quotes both interpolations via the tested `shellQuote`.)

**4. Partial attachment upload orphans files.** `CreateIssueModal.tsx:95-103` — the upload loop uploads each file individually; on any failure it Toasts and `return`s before `createIssue`, so the files that already succeeded are neither bound nor deleted (no delete API exists in `attachmentApi.ts`), and because `pendingFiles` isn't pruned on per-file success, every retry re-uploads and re-orphans. (Note: `setSubmitting(false)` *does* still run — the early `return` is inside the `try`, and the `finally` executes.) Consider proceeding with the successfully-uploaded ids, or cleaning up on abort.

**5. Dead code / stale artifacts after the removals.** Cleanup, same class the PR already flags:
- `authApi.ts:15` `getDaemonServerUrl()` — now zero callers.
- `headlessCommand.ts:13` `headlessCommand()` — only referenced by its own test; prod switched to `authCommand`/`INSTALL_SCRIPT_CMD`/`START_CMD`.
- i18n `loop.runtime.headlessNoBackend` (both locales) and the whole `newLoop.*` namespace (~24 keys, both locales) — no remaining code references.
- `packages/dmworkbase/src/App.tsx:79` still declares the `wk:loop-issues-dispatched` mittBus event (+ comment) whose only emitter/subscriber were removed with `NewLoopPage`.

**6. Test coverage points at the dead function.** `__tests__/headlessCommand.test.ts` keeps a shell-injection test for the now-dead `headlessCommand()`, while the live PAT-embedding builder `authCommand()` has no direct test. `shellQuote` is still tested (so injection safety is indirectly covered), but on a security-sensitive path a direct `authCommand` injection test would be better; the dead-function test can be removed/repurposed.

**7. Invalid ARIA table.** `ProjectPage.tsx:113-158` switched to `role="table"`/`role="row"` but no cell has `role="cell"`/`columnheader`, so screen readers announce a broken table. The sibling `RuntimePage.tsx:208-223` already does this correctly with `role="cell"` on every cell — please match that (or revert to `role="list"`). Row navigation is also mouse-only (`onClick` without `role="button"`/`tabindex`).

## Note on the dark-mode concern (checked, NOT a bug)
I specifically investigated whether the new `html body` light-tint tokens (`--semi-color-primary-light-*` = `#F1F1F3/#E9E9EC/#E1E1E5`, `App.css:120-127`) leak into dark mode, since the dark block (`body[theme-mode=dark]`) doesn't re-set them. They do **not** leak: Semi's own `semi.min.css` re-declares all three light tokens under `body[theme-mode=dark], …` (specificity 0,1,1) with dark values `rgba(var(--semi-blue-5),.2/.3/.4)`, which outranks the PR's `html body` rule (specificity 0,0,2) regardless of bundle order. Dark mode renders Semi's tinted values correctly. No action needed.

## Summary
Solid, well-commented UI refactor with correct shell-escaping and a sound mint-on-copy guard. One blocking item: the "one-time credential" wording contradicts the only evidence of the token's lifetime on a flow that hands a live credential to an AI — correct the copy or confirm backend single-use/TTL. The rest are advisory cleanups.

Jerry-Xin (changes requested):
# Re-review — flipping my earlier APPROVE to Changes Requested

Head `10875cb481e13ec70e2da647365911654ee218f9`. I re-checked the Add-Computer / headless credential flow after a peer raised a labeling-accuracy concern, and I agree it is blocking. Reversing my prior approval on this one point; the rest of the PR is unchanged and looked good.

## Blocking

**Credential is labeled "one-time" but the only in-repo lifetime signal says it is a 90-day token.** The new install hint (`packages/dmloop/src/i18n/en-US.json:112` and `zh-CN.json:112`) tells the user the copied instructions "contain a one-time credential — do not share", and the code comments echo "one-time PAT" (`packages/dmloop/src/pages/headlessCommand.ts:22`, `packages/dmloop/src/pages/RuntimePage.tsx:129`). But the only evidence of the token's actual lifetime is the mint helper comment at `packages/dmloop/src/api/authApi.ts:7-8`, which states the backend signs a 90-day personal access token for the current user.

There is no client-side single-use / consume / revoke path: `onCopyInstall` mints once and deliberately caches and reuses the token for the dialog session (`packages/dmloop/src/pages/RuntimePage.tsx` around lines 128-140), and the credential is interpolated live into the login command (`authCommand`, `packages/dmloop/src/pages/RuntimePage.tsx:125`) that the copy then instructs the user to paste into an AI "Bot" for auto-execution.

So a credential that can live up to 90 days is presented to the user as "one-time" while transiting an AI tool chain. "One-time" reads as throwaway/single-use; a long-lived token described that way materially understates the blast radius and changes how carefully a user will handle it. On a security-sensitive flow this labeling accuracy gap should be fixed before merge.

Fix: either mint a genuinely short-lived or single-use token server-side, or change the wording to state the real lifetime (for example "temporary access token, valid ~90 days"). Please confirm the actual TTL and single-use semantics of the headless mint endpoint with the backend before shipping the "one-time" copy.

## Non-blocking (advisory)

- Supply-chain hardening: `INSTALL_SCRIPT_CMD` (`packages/dmloop/src/pages/headlessCommand.ts:19-20`) fetches an installer from a mutable `main` ref with no commit pin and no integrity checksum, piped straight into a runtime and (by design) auto-executed by an AI alongside an embedded live credential. Consider pinning to a commit SHA or a signed release with an integrity check.
- Trust boundary: `--server-url` is now derived from `window.location.origin` (`packages/dmloop/src/pages/RuntimePage.tsx:122`) rather than the backend-configured server URL. Reasonable, but worth a conscious confirmation that every access host is trusted, since that origin is baked into the credentialed command. Shell-injection itself is not a concern here — both interpolations are single-quoted via the tested escaping helper.

Nice work overall on the escaping and the mint-on-copy guard; the single blocking item is the label-vs-lifetime mismatch. Credit to the peer reviewer who first flagged the "one-time" wording.

OctoBoooot (changes requested):
## Flipping my #757 verdict APPROVE → REQUEST_CHANGES — yujiawei's credential-label 🔴 is byte-confirmed

Credit yujiawei. My earlier APPROVE audited the credential-mint *mechanics* (minted only on copy-click, `mintingRef` prevents accumulation, not logged, cleared on dialog close) but did **not** audit whether the user-facing "one-time" wording matches the token's actual lifetime. It doesn't — and that's a real, security-relevant defect. Withdrawing my APPROVE (this CR supersedes it as my latest review).

### Blocker — a 90-day reusable PAT is labeled "one-time" to the user
Byte-verified both halves at `10875cb481e1`:
- **The token is a 90-day PAT.** `packages/dmloop/src/api/authApi.ts:7-8`: *"headless 免浏览器接入：后端当场为当前用户签一个 **90 天 PAT**"*. `issueHeadlessCliToken()` returns that token; a PAT is reusable until expiry — no single-use semantics exist anywhere in the flow.
- **The user-facing install hint calls it "one-time" in both locales:**
  - `packages/dmloop/src/i18n/en-US.json:112` — `"installHint": "Copy these instructions to your Bot to run (contains a **one-time** credential — do not share):"`
  - `packages/dmloop/src/i18n/zh-CN.json:112` — `"installHint": "把这段指令复制给 Bot 执行（含**一次性**认证凭证，请勿外传）："`

**Why this is a blocker, not a nit:** this is the credential-handling UX (a security-critical path), and the statement is factually false in the dimension that governs how a user treats a leaked secret. A user told the credential is "one-time / 一次性" reasonably reads it as single-use/disposable — so a paste into the wrong chat, a shared screen, or a committed script feels self-limiting and won't trigger urgent rotation. In reality the token grants access for **90 days**. The mislabel directly degrades the user's risk assessment of a live credential. (The adjacent "do not share / 请勿外传" wording confirms the token is known to be reusable — which contradicts "one-time" and underlines the error.)

**Fix is small:** correct the two `installHint` strings to state the real lifetime — e.g. "contains a **90-day** credential — do not share" / "含**90 天有效**的认证凭证，请勿外传" (and, for consistency, the "一次性 PAT" code comments at `RuntimePage.tsx:129` and `headlessCommand.ts:22`, though those are non-user-facing). Once the user-facing wording matches the 90-day TTL, this clears.

### yujiawei's two 🟡s — acknowledged (non-blocking)
- **Install command not pinned / no SHA verification (supply-chain)** — `INSTALL_SCRIPT_CMD` is an unpinned fetch-and-run; pinning a version + integrity/SHA check would harden the supply chain. Fair 🟡; it's the standard install-script tradeoff, non-blocking but worth doing for a credential-bearing setup flow.
- **`--server-url` trust boundary** — `serverUrl = window.location.origin` (RuntimePage.tsx). This is **browser-derived** — the origin the user is already authenticated against — so the daemon is pointed at the site the user is actively using, not an attacker-supplied host. I reasoned this acceptable in my original review and still do; the 🟡 is a "confirm the daemon doesn't later accept an overridable server-url from a less-trusted source." Worth a one-line confirmation, non-blocking.

### Verdict + gate
**REQUEST_CHANGES @ `10875cb481e1`** — one blocker (the one-time/90-day mislabel), two non-blocking 🟡s. This converges with yujiawei's CR (`reviewDecision` is already `CHANGES_REQUESTED`). Independent of this, merge stays gated by `needs-human-review` (size/XL). Fix the two `installHint` strings to state the real 90-day lifetime and this is back to an easy approve. Credit yujiawei for catching the label-truthfulness gap my mechanics-only audit missed.

PR #759, opened 2026-07-14
fix(docs): clear right pane after deleting the open doc

ploy-elison:
## Summary

Deleting the currently-open document left its editor resident in the right pane — the list refreshed but the deleted doc's shell stayed on screen (with lingering Yjs/WS traffic and 404s against the removed doc).

Root cause: `DocsHome.onDocDeleted` guarded `docId === selectedDocId` against a **stale, closed-over** `selectedDocId`. In the production path the editor/sheet/board shell is pushed into the host right pane by `commitOpen` as a **one-time element snapshot**, and that push runs synchronously right after `setSelectedDocId(X)` — before the state re-render. So the `onDocDeleted` baked into the snapshot still saw the pre-open id (`null` on first open, the previous doc after a switch). The guard therefore never matched the just-opened doc: `backToList()` never ran, while `setListReloadToken` still fired — list refreshed, right pane didn't.

## Fix

- Mirror `selectedDocId` into a ref (`selectedDocIdRef`) and compare against the **live** value in `onDocDeleted`, dropping the stale `selectedDocId` dependency.
- `backToList()` unmounts the shell, which also tears down the Yjs collab session — clearing the stale WS/404 traffic against the deleted doc.
- `doc`, `sheet`, and `board` share the single `onDocDeleted`, so all three shells are fixed by the one change.

## Tests

Added a regression test (the missing coverage that let this ship) that drives the **real snapshot path**: it invokes the `onDeleted` wired into the element pushed via `routeRight.replaceToRoot`, then asserts the right pane is reset to the empty state and the selection cleared — for **all three shells** (doc / sheet / board). Verified the test fails on the pre-fix code and passes after. A negative case asserts that deleting a *different* doc only refreshes the list and keeps the open doc resident.

`packages/docs` tests: the full `DocsHome.test.tsx` suite passes (40/40). Pre-existing, unrelated failures elsewhere in the package (a jsdom/undici WebSocket issue in `EditorShell.test.tsx`, two typechecker errors in `i18n/format.ts` and `members/sort.ts`, and a broken `build:standalone` `VoiceInputButton` export) reproduce on a clean `main` and are out of scope for this change.

Draft — for review, not for merge.

### DIFF
--- a/packages/docs/src/pages/DocsHome.test.tsx
+++ b/packages/docs/src/pages/DocsHome.test.tsx
@@ -1,5 +1,5 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
-import { render, screen, waitFor, cleanup, fireEvent } from '@testing-library/react'
+import { render, screen, waitFor, cleanup, fireEvent, act } from '@testing-library/react'
 import type { ReactNode } from 'react'
 import { setWKApp } from '../octoweb/index.ts'
 import { createMockWKApp } from '../octoweb/mock.ts'
@@ -1074,5 +1074,127 @@ describe('DocsHome — in-flight unknown-kind open is discarded after a Space sw
       | { props?: { docId?: string } }
       | undefined
     expect(lastPush?.props?.docId).toBeUndefined()
+  })
+})
+
+// Deleting the currently-open doc must clear the right pane and the selection, not just refresh the
+// list. Regression (XIN-1050): onDocDeleted guarded `docId === selectedDocId` against a STALE
+// closed-over selectedDocId. In the production (routeRight) path the shell is pushed into the host
+// pane as a one-time element snapshot by commitOpen — and that push runs synchronously right after
+// setSelectedDocId(X), before the state re-render — so the onDocDeleted baked into the snapshot
+// still saw the PRE-open id (null on first open, the previous doc on a switch). The guard therefore
+// never matched the just-opened doc, backToList never ran, and the deleted doc's shell stayed
+// resident in the right pane while only the list refreshed. doc / sheet / board share the single
+// onDocDeleted, so all three regressed. The fix reads selectedDocIdRef (the always-current id).
+// This test drives the exact snapshot path: it invokes the onDeleted wired into the pushed element,
+// as the shell does after a successful in-editor delete of the open doc.
+describe('DocsHome — deleting the open doc clears the right pane and selection (XIN-1050)', () => {
+  const shells: Array<{ kind: 'doc' | 'sheet' | 'board'; docType: string }> = [
+    { kind: 'doc', docType: 'doc' },
+    { kind: 'sheet', docType: 'sheet' },
+    { kind: 'board', docType: 'board' },
+  ]
+
+  for (const { kind, docType } of shells) {
+    it(`resets routeRight to the empty state and clears selection after deleting the open ${kind}`, async () => {
+      const wk = createMockWKApp()
+      const replaceToRoot = vi.fn()
+      // Production (resident-list) path so the shell is pushed into the host pane as a snapshot —
+      // the only path where the stale-closure bug manifests (the inline path rebuilds the shell
+      // every render, so its onDocDeleted is always current).
+      ;(wk as { routeRight?: unknown }).routeRight = { replaceToRoot, popToRoot: vi.fn() }
+      setWKApp(wk)
+      wk.apiClient.responder = (method, url) => {
+        if (method === 'get' && url.startsWith('/docs')) {
+          return {
+            data: {
+              total: 1,
+              items: [
+                { docId: 'd_x', title: 'Open Doc', ownerId: 'u_self', role: 'admin', docType },
+              ],
+            },
+            status: 200,
+          }
+        }
+        return { data: {}, status: 200 }
+      }
+
+      render(<DocsHome />)
+      await waitFor(() => expect(screen.getByText('Open Doc')).toBeTruthy())
+
+      // Known-kind row → openDoc commits synchronously and pushes the matching shell into the host
+      // pane. Wait for that push (the earlier mount empty-state push is not the shell).
+      fireEvent.click(screen.getByText('Open Doc'))
+      await waitFor(() => {
+        const last = replaceToRoot.mock.calls.at(-1)?.[0] as
+          | { props?: { docId?: string } }
+          | undefined
+        expect(last?.props?.docId).toBe('d_x')
+      })
+      // The just-opened row is marked active (selectedDocId === 'd_x').
+      await waitFor(() =>
+        expect(document.querySelector('.octo-docs-list-item-active')).toBeTruthy(),
+      )
+      // The durable target was persisted for the open doc.
+      expect(JSON.parse(window.sessionStorage.getItem(TARGET_KEY)!)).toMatchObject({ doc: 'd_x' })
+
+      // Fire the onDeleted baked into the pushed shell snapshot, exactly as the shell does after a
+      // successful in-editor delete of the open doc.
+      const pushed = replaceToRoot.mock.calls.at(-1)![0] as {
+        props: { docId: string; onDeleted: (id: string) => void }
+      }
+      expect(typeof pushed.props.onDeleted).toBe('function')
+      act(() => pushed.props.onDeleted('d_x'))
+
+      // routeRight is cleared to the docs empty state — NOT left showing the deleted doc's shell.
+      await waitFor(() => {
+        const last = replaceToRoot.mock.calls.at(-1)?.[0] as
+          | { props?: { docId?: string } }
+          | undefined
+        expect(last?.props?.docId).toBeUndefined()
+      })
+      // selectedDocId is reset: no list row stays active and the persisted target is cleared.
+      expect(document.querySelector('.octo-docs-list-item-active')).toBeNull()
+      expect(window.sessionStorage.getItem(TARGET_KEY)).toBeNull()
+    })
+  }
+
+  it('refreshes the list but keeps the pane when a DIFFERENT (not-open) doc is deleted', async () => {
+    const wk = createMockWKApp()
+    const replaceToRoot = vi.fn()
+    ;(wk as { routeRight?: unknown }).routeRight = { replaceToRoot, popToRoot: vi.fn() }
+    setWKApp(wk)
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url.startsWith('/docs')) {
+        return {
+          data: {
+            total: 1,
+            items: [{ docId: 'd_x', title: 'Open Doc', ownerId: 'u_self', role: 'admin', docType: 'doc' }],
+          },
+          status: 200,
+        }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<DocsHome />)
+    await waitFor(() => expect(screen.getByText('Open Doc')).toBeTruthy())
+    fireEvent.click(screen.getByText('Open Doc'))
+    await waitFor(() => {
+      const last = replaceToRoot.mock.calls.at(-1)?.[0] as { props?: { docId?: string } } | undefined
+      expect(last?.props?.docId).toBe('d_x')
+    })
+
+    const pushed = replaceToRoot.mock.calls.at(-1)![0] as {
+      props: { onDeleted: (id: string) => void }
+    }
+    // A doc other than the open one is deleted elsewhere: the open doc must stay resident.
+    act(() => pushed.props.onDeleted('d_other'))
+
+    // The right pane still shows the open doc (no empty-state push) and the target is intact —
+    // only the resident list is refreshed (via the reload-token bump).
+    const last = replaceToRoot.mock.calls.at(-1)?.[0] as { props?: { docId?: string } } | undefined
+    expect(last?.props?.docId).toBe('d_x')
+    expect(JSON.parse(window.sessionStorage.getItem(TARGET_KEY)!)).toMatchObject({ doc: 'd_x' })
   })
 })
--- a/packages/docs/src/pages/DocsHome.tsx
+++ b/packages/docs/src/pages/DocsHome.tsx
@@ -836,6 +836,19 @@ export function DocsHome() {
     () => initialKnownKind,
   )
 
+  // Live mirror of selectedDocId for callbacks pushed imperatively into the host route pane. The
+  // editor/sheet/board shells are pushed via routeRight.replaceToRoot in commitOpen — a ONE-TIME
+  // element snapshot that bakes in whatever `onDocDeleted` closure existed at push time. Because
+  // that push runs synchronously right after setSelectedDocId(X) (before the state re-render), the
+  // baked-in closure still sees the PRE-open selectedDocId (null on first open, the previous doc on
+  // a switch). onDocDeleted must therefore compare against this ref — the always-current id — not
+  // the closed-over state, or the `docId === selectedDocId` guard never matches and the deleted
+  // doc's shell is left resident in the right pane (XIN-1050).
+  const selectedDocIdRef = useRef<string | null>(selectedDocId)
+  useEffect(() => {
+    selectedDocIdRef.current = selectedDocId
+  }, [selectedDocId])
+
   // The host's right (main) route pane. When present (production), the editor is pushed there
   // so it fills the main content area while the list stays in the left route slot — the same
   // full-width list+detail layout Matter/Summary use. When absent (tests / standalone), we
@@ -921,13 +934,18 @@ export function DocsHome() {
   // Called after a successful delete (now from the editor detail page, Problem 4). If the deleted
   // doc is the one open in the right pane, return to the empty/list state (which also resets the
   // editor's drawer, #5 C4); always bump the reload token so the resident list refreshes.
+  // We read selectedDocIdRef (the live id), NOT the closed-over selectedDocId: this callback is
+  // baked into the shell snapshot pushed by commitOpen before the open's state update commits, so
+  // the closure's selectedDocId is stale (pre-open). Reading the ref makes the guard match the
+  // actually-open doc across doc/sheet/board (XIN-1050).
   const onDocDeleted = useCallback(
     (docId: string) => {
-      if (docId === selectedDocId) backToList()
+      if (docId === selectedDocIdRef.current) backToList()
       setListReloadToken((n) => n + 1)
     },
-    [selectedDocId, backToList],
+    [backToList],
   )
+
 
   // "Open in new page" (AC-1): open the current doc as a standalone full-window `/d/:docId` link
   // in a new browser tab — the clean, shareable cold-load entry that lives outside the app shell.


Jerry-Xin (approved):
**Verdict: APPROVE** — correct, well-scoped fix for the stale-view bug. The root cause and fix are exactly right.

### Why the fix is correct
The bug: `onDocDeleted` guarded `docId === selectedDocId` against a closed-over `selectedDocId`. In the production (`routeRight`) path the shell is pushed as a one-time element snapshot by `commitOpen`, and that push runs synchronously right after `setSelectedDocId(X)` — before the state re-render — so the baked-in closure still saw the pre-open id. Guard never matched → `backToList()` never ran → the deleted doc's shell stayed resident.

The fix reads a live `selectedDocIdRef` (kept in sync via `useEffect`), so the baked-in snapshot's closure resolves against the always-current id. This is the correct React pattern for a value read inside an imperatively-pushed snapshot.

### Audit results
- **Clears only when deleted == open** ✅ — `if (docId === selectedDocIdRef.current) backToList()` (DocsHome.tsx onDocDeleted). Deleting a different doc only bumps the reload token; the pane stays. Both cases are covered by tests.
- **No dangling reference** ✅ — `backToList` fully clears: selection (`setSelectedDocId(null)` / `setSelectedDocType(undefined)`), in-flight open guard (`latestOpenRef = null`), durable target (`clearDocTarget()`), URL mirror, and replaces the pane with the empty state. No stale docId / editor / target survives.
- **Ordering / use-after-free** ✅ — no window where the editor points at a freed doc; the snapshot closure now dereferences the live ref rather than freed state.
- **Race / async** ✅ — the ref is updated on every `selectedDocId` change; the async unknown-kind `getDoc` path also pushes a fresh snapshot, so mid-delete open-another-doc scenarios stay consistent.
- **i18n** ✅ — empty-state text uses `t('docs.state.empty')`; no hardcoded user-facing CJK introduced.
- **Scope** ✅ — 2 files (fix + tests); no stray flag flips, no console leaks.
- **Tests** ✅ — parametrized over doc / sheet / board, driving the exact snapshot path (`onDeleted` baked into the pushed element): asserts pane reset to empty state + selection/target cleared for the open-doc case, and pane-unchanged + target intact for the not-open-doc case. Good behavioral coverage of the regression.

### 🟡 Non-blocking
- **DocsHome.tsx** — a stray double blank line was introduced after the `onDocDeleted` callback. Trivial; drop one blank line.

### Note
PR is currently marked **draft**. Fix looks merge-ready from a review standpoint once un-drafted.

### ✅ Highlights
- Excellent inline documentation of the stale-closure root cause and the ref rationale.
- Tests target the precise failing path (imperative snapshot) rather than the inline path that never regressed — the distinction is called out explicitly.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #759 (octo-web)

## Summary

This PR fixes a stale-closure bug (XIN-1050) where deleting the currently-open doc from within the editor/sheet/board shell fails to clear the right pane. The root cause is well-diagnosed: `commitOpen` calls `setSelectedDocId(docId)` then synchronously pushes a one-time element snapshot via `routeRight.replaceToRoot(...)`. The `onDocDeleted` callback baked into that snapshot closes over the PRE-open `selectedDocId` state value, so the `docId === selectedDocId` guard never matches the just-opened doc.

The fix introduces a `selectedDocIdRef` (always-current mirror of the state) and reads `selectedDocIdRef.current` in `onDocDeleted` instead of the closed-over state. Regression tests cover all three shell types (doc/sheet/board) plus a negative case. Overall: a focused, well-documented bug fix with thorough test coverage.

## Verification

- ✅ **Ref correctness** — `selectedDocIdRef` is initialized from `selectedDocId` at `packages/docs/src/pages/DocsHome.tsx:847` and kept in sync via `useEffect` at `:848-850`. The effect runs during the commit phase after the `setSelectedDocId` re-render, completing before any user interaction can fire the delete callback.
- ✅ **Callback dep array** — `onDocDeleted` at `:941-945` now depends only on `[backToList]`. Removing `selectedDocId` is correct: the callback reads the ref, not the state. `backToList` is referentially stable (deps: `[routeRight, buildEmptyState]`, both stable).
- ✅ **Data flow (all shell types)** — `onDocDeleted` is wired as `onDeleted` through `buildEditor` (`:993`), `buildBoard` (`:1015`), and `SheetView` (`:1039`). `BoardSession` at `packages/docs/src/board/BoardSession.tsx:44` correctly destructures and forwards `onDeleted` to `BoardShell`. All three shell types receive the fixed callback.
- ✅ **Test coverage** — Four new test cases: three parametrized (doc/sheet/board) drive the exact snapshot path by extracting `onDeleted` from the pushed element and calling it, reproducing the production stale-closure scenario. The negative test verifies that deleting a different doc keeps the pane intact. All tests use proper `act()` wrapping and `beforeEach`/`afterEach` cleanup.
- ✅ **Inline (non-routeRight) path** — The split-pane path at `:1193-1194` calls `buildRightPane(...)` on every render, so the callback is always fresh. The stale-closure bug only affected the imperative `replaceToRoot` snapshot path, which is the path this PR fixes.

Static analysis only at head `57b87c80e5c5470da587831305a3e15390262f33`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. The fix is correct and the test coverage is thorough.

### Nit — Extra blank line (`packages/docs/src/pages/DocsHome.tsx:947`)

An extra blank line appears between the `onDocDeleted` closing `)` and the "Open in new page" comment block. Cosmetic only.

## Data Flow Trace

- `onDocDeleted` (consumer) reads `selectedDocIdRef.current` ← set by `useEffect` ← tracks `selectedDocId` state ← set by `commitOpen` → `setSelectedDocId(docId)`. The ref update runs during the commit phase after the state update triggers a re-render, completing before any user-driven event (delete button click) can fire. ✅ Data flows correctly.
- `backToList` (called by `onDocDeleted` when guard matches) → `setSelectedDocId(null)` + `setSelectedDocType(undefined)` + `latestOpenRef.current = null` + `clearDocTarget()` + `mirrorListToUrl()` + `routeRight.replaceToRoot(buildEmptyState())`. ✅ All side effects are correct and complete.

- **C1 (dual-path parity):** CLEAR — `onDocDeleted` is the single callback shared by doc/sheet/board. The fix applies to all three via `buildRightPane`. Verified by parametrized tests.
- **C2 (control-flow ordering):** CLEAR — No nested reuse or ordering change. The `useCallback` identity is preserved across renders.
- **C3 (auth boundary):** N/A — No authorization change.
- **C4 (auth lifecycle):** N/A — No auth change.
- **C5 (build/runtime):** CLEAR — No build or packaging change.
- **C6 (governance/docs):** N/A — No governance doc change.

## Cross-Round Blocker Review (R6)

N/A — first review round.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The stale-closure fix is correct, well-documented with inline comments explaining the exact timing mechanics, and covered by regression tests that drive the production snapshot path. One cosmetic nit (extra blank line).

yujiawei (approved):
# Code Review — PR #759 (octo-web)

## Summary

This fixes a genuine React stale-closure bug. `DocsHome.onDocDeleted` compared the deleted `docId` against a **closed-over** `selectedDocId`. In the production path the editor/sheet/board shell is pushed into the host right pane by `commitOpen` via `routeRight.replaceToRoot(...)` as a **one-time element snapshot**, and that push runs synchronously right after `setSelectedDocId(X)` — before the re-render — so the `onDocDeleted` baked into the snapshot still saw the pre-open id (`null` on first open, the previous doc after a switch). The guard therefore never matched the just-opened doc: `backToList()` never ran while `setListReloadToken` still fired, so the list refreshed but the deleted doc's shell stayed resident. The fix mirrors `selectedDocId` into `selectedDocIdRef` and compares against the live `.current`, dropping the stale state dependency.

The root-cause diagnosis is accurate and the fix is minimal and correct. `doc`, `sheet`, and `board` all share the single `onDocDeleted`, so all three shells are fixed by the one change.

## Verification performed

- Full `DocsHome.test.tsx` suite passes locally: **40/40**.
- Confirmed the regression tests are not tautological: reverting `onDocDeleted` to the pre-fix stale-closure form (`docId === selectedDocId`, dep `[selectedDocId, backToList]`) makes exactly the 3 new "resets routeRight..." cases fail for doc / sheet / board, and they pass again with the ref-based guard. The test drives the real snapshot path — it fires the `onDeleted` wired into the element pushed via `routeRight.replaceToRoot`, then asserts the pane resets to the empty state and the selection/persisted target clear.
- Verified `onDeleted` can only fire after an awaited `deleteDoc` behind a two-step user confirm (`useDocDelete.confirm`), never synchronously during the shell's mount.

## Correctness notes

- The `useCallback` dep change (`[selectedDocId, backToList]` → `[backToList]`) is correct: the body no longer reads `selectedDocId`, and reading a ref requires no dependency. This also makes `onDocDeleted` referentially stable across selection changes, which is desirable here.
- `backToList` is the only other closure captured in the same snapshot. I checked it: it is a `useCallback([routeRight, buildEmptyState])` whose body touches only stable setState functions, refs, and module-level helpers (`clearDocTarget`, `mirrorListToUrl`) — it closes over **no** mutable render state, so it does not carry the same staleness hazard today.

- **N1 — future-proofing (optional).** The imperative-snapshot boundary makes *any* callback pushed through `replaceToRoot` a stale-closure risk. Today only `selectedDocId` was affected and `backToList` happens to be state-free, but if `backToList` (or another pushed callback) later starts closing over render state, this class of bug can silently return. A `useEvent`/latest-ref wrapper around the whole pushed callback would neutralize the pattern rather than patching one variable. Not required for this fix — the current change is correct as-is.
- **N2 — timing (verified non-issue, note only).** `selectedDocIdRef` is updated in a passive `useEffect`. That is safe here because `onDeleted` is always gated behind an awaited network delete, so the effect has long since flushed. If a pushed shell ever gained a path that could invoke `onDeleted` synchronously during mount, `useLayoutEffect` would be the safer choice. No action needed now.
- **N3 — cosmetic.** The diff leaves a stray double blank line after the `onDocDeleted` block (`DocsHome.tsx` ~L948-949). Trivial.

## Scope

The change is scoped to `DocsHome.tsx` (the fix) and `DocsHome.test.tsx` (regression coverage) — nothing outside the stated fix, no unrequested features. The pre-existing, unrelated failures the author calls out (jsdom/undici WS in `EditorShell.test.tsx`, two `i18n/format.ts` + `members/sort.ts` typecheck errors, the `build:standalone` `VoiceInputButton` export) are out of scope for this PR.

## Verdict

No P0/P1 issues. The fix is correct, minimal, covers all three shells, and ships with a regression test that genuinely reproduces the bug. **Approved.** (The PR body marks this "Draft — for review, not for merge"; the review verdict reflects code quality/correctness — the merge-timing decision remains the author's.)

OctoBoooot:
@/tmp/pr759-advisory-57b87c80.md

boris-clark (approved):
Verified against the head branch, not the commit message.

- `selectedDocIdRef` reads the latest value: initialized once, then kept in sync via `useEffect(() => { selectedDocIdRef.current = selectedDocId }, [selectedDocId])`. `onDocDeleted` reads `selectedDocIdRef.current` and its deps are correctly narrowed to `[backToList]` — no stale closure.
- All three shells covered: the fix lives in the single shared `onDocDeleted` callback wired to `onDeleted` on EditorShell, BoardSession and SheetView.
- The regression test genuinely catches the bug: reverting `selectedDocIdRef.current` back to `selectedDocId` makes the three shell tests fail (pane keeps showing the deleted doc) while the negative "delete a different doc" case still passes. It asserts both the empty right pane and the reset selection.
- No Yjs leak: teardown happens via the existing `key={docId}` unmount path (`useCollabEditor` cleanup → `destroyAll()`); this change only ensures `backToList` actually runs.
- Deletions are limited to the stale comparison and dep — nothing else removed.
- The jsdom "unhandled error" noise is pre-existing on `main`; not introduced here.

Non-blocking:
- A redundant double blank line was introduced after the `onDocDeleted` callback. Trivial nit.

LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #759 (octo-web)

## Summary

Single-commit bug fix for XIN-1050: deleting the currently-open doc in the docs list failed to clear the right pane. Root cause was a stale-closure bug in `onDocDeleted` — the callback was baked into the shell element snapshot pushed by `commitOpen` via `routeRight.replaceToRoot`, which runs synchronously after `setSelectedDocId(X)` but before React's re-render commits. The closed-over `selectedDocId` therefore held the pre-open value (null on first open, previous doc on a switch), so the `docId === selectedDocId` guard never matched and the deleted doc's shell stayed resident.

The fix introduces `selectedDocIdRef` — a `useRef` mirror of `selectedDocId` synced via `useEffect` — and reads `selectedDocIdRef.current` in `onDocDeleted` instead of the stale closure. This is the canonical React pattern for escaping stale closures in imperative/snapshot contexts. Regression tests cover all three shell types (doc, sheet, board) plus a negative case for deleting a non-open doc.

## Verification

Static analysis only at head `57b87c80e5c5470da587831305a3e15390262f33`; build and tests not executed in this environment.

- ✅ **Ref sync correctness** — `selectedDocIdRef` is initialized from `selectedDocId` at `DocsHome.tsx:847` and synced via `useEffect` at `:848-850`. The effect fires after every render where `selectedDocId` changed. Since user-initiated delete (the only trigger for `onDocDeleted`) is always asynchronous relative to the render that committed the new state, `selectedDocIdRef.current` is always current when the callback fires.
- ✅ **Callback stability** — `onDocDeleted` deps changed from `[selectedDocId, backToList]` to `[backToList]`. Since `backToList` is itself stable (deps: stable `routeRight` + `[]`-stable `buildEmptyState`), `onDocDeleted` is now referentially stable across renders — no unnecessary rebuilds of `buildEditor`/`buildBoard`/`buildRightPane`.
- ✅ **All three shell paths covered** — `onDocDeleted` is wired into `buildEditor` (`:993`), `buildBoard` (`:1017`), and `buildRightPane` → `SheetView` (`:1040`). The fix reaches all three via the single shared callback.
- ✅ **Test-production path parity** — The regression tests invoke `onDeleted` extracted from the actual `replaceToRoot` push snapshot, exactly mirroring the production shell-delete path. The negative test asserts that deleting a different doc preserves the open pane and persisted target.
- ✅ **No residual stale reads** — Grep confirmed no other callback reads `selectedDocId` through a snapshot/push path. All remaining `selectedDocId` reads are in render-time JSX (`:1141`, `:1172`, `:1187`, `:1193-1194`), where state is always current.

## Findings

No P0/P1 issues. One cosmetic nit.

### Nit — Extra blank line (`packages/docs/src/pages/DocsHome.tsx:949`)

An additional blank line was introduced between the `onDocDeleted` callback and the `onOpenInNewPage` comment block, where previously there was one. Trivial cleanup.

## Data Flow Trace

- `selectedDocId` (state, `:828`) → set by `commitOpen` (`:1058`) → **before fix**: read by `onDocDeleted` closure (stale at snapshot time). **After fix**: `selectedDocIdRef` (`:847`) mirrors state via `useEffect` (`:848-850`), read by `onDocDeleted` (`:943`) at callback-invocation time (always post-commit).
- `backToList` (`:885`) → clears `selectedDocId` to null → useEffect syncs `selectedDocIdRef.current = null` → subsequent `onDocDeleted` call with any string docId correctly fails the `===` guard (no double-clear).
- `commitOpen` (`:1056`) → `setSelectedDocId(docId)` then synchronously `buildRightPane(docId, dt)` → `replaceToRoot(shell)`. Shell snapshot bakes stable `onDocDeleted` → user later triggers delete → `onDocDeleted(docId)` reads `selectedDocIdRef.current` (current) → guard matches → `backToList()` runs.

## Blindspot Checklist

- **C1 (dual-path parity)**: Clear. `onDocDeleted` is the sole delete callback. No symmetric counterpart.
- **C2 (control-flow ordering)**: Clear. The `useEffect` ref-sync runs after every render; no ordering hazard with user-initiated callbacks.
- **C3–C4 (auth boundaries)**: N/A. UI-only bug fix, no auth/permission surface touched.
- **C5 (build/deploy)**: N/A. No build, config, or deploy changes.
- **C6 (governance/docs)**: N/A. No policy or governance docs changed.

## Verdict: APPROVED

Focused, well-reasoned bug fix. The stale-closure diagnosis is correct and the ref-mirror fix is the standard React escape hatch for imperative snapshot contexts. Tests faithfully reproduce the production push-snapshot path across all three shell types. No correctness, security, or functional concerns.

[Octo-Q] verdict: APPROVE — Correct stale-closure fix via ref mirror; tests cover all three shell types + negative case; no P0/P1 findings.

OctoBoooot (approved):
## Review: clear right pane after deleting the open doc (#759)

**Verdict: Approve.** Now out of draft; re-verified at head `57b87c80e5c5` (unchanged since my earlier advisory, so the analysis holds). This is the formal gating verdict, superseding my draft-time advisory note. Clean fix for a real stale-closure bug, correctly scoped, well-tested.

### The fix is correct — and the root-cause reasoning is exactly right
The bug: `onDocDeleted` guarded `docId === selectedDocId` against a **stale closed-over** `selectedDocId`. The editor/sheet/board shell is pushed via `commitOpen → routeRight.replaceToRoot` — a **one-time element snapshot** that bakes in whatever `onDocDeleted` closure existed at push time, and that push runs synchronously right after `setSelectedDocId(X)` *before* the state re-render. So the baked-in closure kept seeing the pre-open `selectedDocId` (null on first open, the previous doc on a switch); the guard never matched the just-opened doc, `backToList` never ran, and the deleted doc's shell stayed resident (XIN-1050).

The fix reads `selectedDocIdRef.current` — a live ref updated via `useEffect` on every `selectedDocId` change (`DocsHome.tsx:847-849`) — instead of the closed-over state, and correctly narrows the callback deps to `[backToList]`. This matches the known octo-web behavior that `replaceToRoot` pushes a one-time snapshot, not a live re-render, so any callback baked into it must read live refs.

### Not over-broad, complete teardown (verified)
- Clears **only** when the deleted doc is the open one: `if (docId === selectedDocIdRef.current) backToList()`; deleting any other doc only bumps `setListReloadToken` (list refresh), never clears the pane. Correct scoping.
- `backToList` is a complete teardown (selection/type reset, `latestOpenRef` invalidated, `clearDocTarget`, URL mirror, `replaceToRoot(buildEmptyState())`) — no dangling ref/listener on the deleted doc, no use-after-free. Uses `replaceToRoot(emptyState)` not `popToRoot` (correct — popToRoot would let the host chat placeholder bleed through).

### Tests — strong, both branches, all three kinds (praise)
`DocsHome.test.tsx` parametrizes doc/sheet/board and asserts both cases: delete-open → pane resets + selection/target cleared; delete-other → pane preserved + target kept + reload bumped. That's exactly the pair that pins the fix and guards the regression.

### CI + gate
All code CI green at this head (skipped checks are governance/conditional, not failures). `mergeStateStatus=BLOCKED` with `code-review: pending` is a required-check/branch-protection gate awaiting the review process to settle — not a code issue; size/M, no `needs-human-review`, no live CR. Once the required review context clears, this is mergeable.

### Net
Correct fix for a real stale-closure bug (live-ref read), correctly scoped (clears only the open doc), complete teardown (no dangling ref), well-tested across all three editor kinds. **Approve** @ `57b87c80e5c5`. The one cosmetic item (a stray blank line after `onDocDeleted`) is a nit, not a gate. Credit ploy-elison — the snapshot-baked-closure vs live-ref root-cause writeup is precisely right.

yujiawei (approved):
# Code Review — PR #759 (octo-web)

**Verdict: APPROVED** (head SHA `57b87c80e5c5470da587831305a3e15390262f33`)

The stale-closure root-cause analysis is correct and the fix is minimal and well-targeted. No blocking (P0/P1) issues. A few advisory (P2) notes on test robustness are below — none require changes before merge.

## Spec compliance

The stated goal — deleting the currently-open doc must clear the right pane and the selection, not just refresh the list — is fully implemented, with no over-build or divergence.

- **Root cause confirmed.** `commitOpen` (`DocsHome.tsx:1056-1080`) calls `setSelectedDocId(docId)` then **synchronously** `routeRight.replaceToRoot(buildRightPane(...))` at line 1068, baking a one-time element snapshot whose `onDocDeleted` closure captured the *pre-open* `selectedDocId` (null on first open, the previous doc after a switch). The old `docId === selectedDocId` guard therefore never matched → `backToList()` never ran → only the list refreshed while the deleted doc's shell stayed resident.
- **Fix is correct.** `selectedDocIdRef` mirrors `selectedDocId` and `onDocDeleted` now reads `selectedDocIdRef.current`. The ref is initialized from `selectedDocId` (`:847`), so the known-kind deep-link mount case is correct from first paint; the sync effect (`:848-850`) keeps it current well before any user-initiated delete can fire.
- **All three shells covered.** doc / sheet / board share the single `onDocDeleted` (`DocsHome.tsx:993/1017/1040`), so one change fixes all three.
- **Dep removal is safe.** Dropping `selectedDocId` from `onDocDeleted`'s deps (`:946`, now `[backToList]`) is correct: the only other statement is `setListReloadToken((n) => n + 1)`, a functional updater that needs no dep. Live-render consumers (`buildRightPane`, list active-row highlight `:627`, `DocsList` props) still read the `selectedDocId` state directly and re-render normally.

## Sibling-callback audit (explicitly checked, no bug)

The other callbacks baked into the same `commitOpen` snapshot do **not** share this bug class:
- `onTitleSaved` (`:860`) — functional updater, reads no state.
- `backToList` (`:886`) — unconditional reset, no compare against `selectedDocId`.
- `onOpenInNewPage` (`:966`) — `useCallback(…, [])` reading `spaceRef.current` (a ref) and its `id` **argument**; wired as `onOpenInNewPage(docId)` with `buildEditor`'s build-time `docId`, not closed-over state.

The stale-comparison bug requires a snapshot-baked callback to read closed-over component state for a decision; only `onDocDeleted` did. Scope of the fix is complete for this class.

## Code quality — advisory (P2, non-blocking)

1. **Negative test does not discriminate the bug.** `DocsHome.test.tsx:1162-1199` fires `onDeleted('d_other')` and asserts the pane stays on `d_x`. Because `'d_other'` matches neither the stale `null` (pre-fix) nor the live `d_x` (post-fix), this case passes on *both* pre- and post-fix code. It's a valid correctness assertion ("deleting a different doc keeps the open one resident") but it is not a regression guard for this fix. A discriminating negative would exercise the switch-then-delete-the-old-id path.
2. **Prefer an explicit call-count assertion in the negative case.** Since a correct negative run makes no new `replaceToRoot` push, asserting `.mock.calls.at(-1)` reads the earlier `d_x` push; an added `expect(replaceToRoot).toHaveBeenCalledTimes(n)` (no new push) would prove omission more directly.
3. **Selection-clear is asserted via a DOM class string.** The positive cases lean on `.octo-docs-list-item-active` presence/absence as a proxy for `selectedDocId`. The `sessionStorage` TARGET_KEY assertion (`:1158`) is the more robust signal and is already present; the class-based check is a somewhat brittle complement.

These are test-hardening suggestions, not correctness defects. The positive cases genuinely drive the production snapshot path and fail on pre-fix code for the right reason (verified locally: the 3 positive shell cases fail before the fix, all 40/40 pass after).

## Verification performed

- Reverted only the source guard to pre-fix and re-ran the new tests: the 3 positive doc/sheet/board cases **fail** (right pane left resident), confirming the test guards the bug; restored the fix and the full `DocsHome.test.tsx` suite passes **40/40**.
- Typecheck surfaces only the two pre-existing, unrelated errors the author already flagged (`i18n/format.ts`, `members/sort.ts`) — neither in the touched files.

## Reviewer coverage note

Reviewed the React state/guard logic and the mocked test path statically and by running the unit suite. Not exercised here: the host-side `WKViewQueue.replaceToRoot` runtime contract (assumed to unmount the previous root, tearing down the Yjs/Hocuspocus session), and the live WS/404-traffic-stops behavior against a running backend — both are outside this diff and were validated by reasoning, not an e2e run.

PR #762, opened 2026-07-14
fix(docs): make link popover keyboard-safe and refresh the link icon

ploy-elison:
## What & why

The `packages/docs` toolbar link popover had two problems confirmed on production:

1. **Enter in the link popover lost the link or typed into the document body.** The popover was not an isolated, focus-owning input: only the URL field had `autoFocus`, the text field had no focus management, and both inputs' `onKeyDown` handlers called `preventDefault()` without `stopPropagation()`. When focus wasn't inside a field, Enter reached the editor's contenteditable and ProseMirror inserted a newline. Separately, `confirmLink` silently `closeLink()`-ed on an empty/unsafe URL, so a mistyped URL just vanished.
2. **The link icon was hard to recognise** — a filled two-capsule glyph rotated 45°.

## Changes

**4a — keyboard/focus correctness**
- On open, move keyboard focus into the popover: the URL field when a selection is being linked, the text field for a brand-new link. Keeps Enter/Escape on the input, never on the editor.
- Add `stopPropagation()` next to `preventDefault()` on both inputs' Enter/Escape handlers so a keystroke can never bubble back to the editor.
- Empty or unsafe URL no longer silently closes the popover: it surfaces an inline error, preserves the typed input, and re-focuses the URL field.
- A scheme-less host (e.g. `example.com`) is prepended with `https://` so it becomes an absolute link instead of a same-origin path.

**4b — icon**
- Replace `IconLink` / `IconUnlink` with the standard stroke chain-link glyphs (`octo-tb-icon-stroke`, `strokeWidth=2`, rounded caps/joins).

## Tests

Added regression coverage in `Toolbar.test.tsx` for the three acceptance paths — existing selection / no selection / brand-new link — plus the empty & invalid-URL inline error, scheme completion, error-clearing on edit, Escape close, and focus isolation. Previously this area had zero enter/focus coverage, which is why the bug shipped.

## Verification

- `packages/docs` unit tests green (all Toolbar tests pass, incl. the new ones).
- Typecheck clean for the changed files.
- Verified in a real browser (Chromium) against the standalone editor: all three paths apply/insert the link correctly with focus landing in the right field and no stray newline in the body; the empty-URL case keeps the popover open with an inline error and preserves input; scheme-less input links as `https://`; Escape closes; the link button renders the new stroke chain icon.

## Scope

No schema / collaboration changes. Draft PR only.

### DIFF
--- a/packages/docs/src/editor/Toolbar.test.tsx
+++ b/packages/docs/src/editor/Toolbar.test.tsx
@@ -1,5 +1,5 @@
 import { describe, it, expect, afterEach, beforeEach, vi } from 'vitest'
-import { render, screen, fireEvent, cleanup, within, act } from '@testing-library/react'
+import { render, screen, fireEvent, cleanup, within, act, waitFor } from '@testing-library/react'
 import { Editor } from '@tiptap/core'
 import StarterKit from '@tiptap/starter-kit'
 import TaskList from '@tiptap/extension-task-list'
@@ -105,7 +105,8 @@ describe('Toolbar — batch 7 quote/code/link/highlight/colour tooltips', () => 
   it('renders the link button as an icon button', () => {
     render(<Toolbar editor={editor!} />)
     const link = titleBtn('docs.toolbar.link')
-    expect(link.querySelector('svg.octo-tb-icon')).toBeTruthy()
+    // XIN-1051: the link glyph is now a stroke-style chain icon (lucide `link`), not a filled one.
+    expect(link.querySelector('svg.octo-tb-icon-stroke')).toBeTruthy()
     expect(link.textContent?.trim()).toBe('')
   })
 })
@@ -133,6 +134,138 @@ describe('Toolbar — batch 7 floating link popover (item 5)', () => {
     expect(field).toBeTruthy()
     fireEvent.keyDown(field, { key: 'Escape' })
     expect(document.querySelector('.octo-link-popover')).toBeNull()
+  })
+})
+
+// XIN-1051 regression: the link popover used to lose the link / type a newline into the body when
+// Enter was pressed, and silently discarded the popover on an empty/invalid URL. These cover the
+// three paths from the acceptance criteria — existing selection, no selection, brand-new link — plus
+// the empty/invalid-URL inline-error behaviour and the focus isolation that keeps Enter off the
+// editor. The `t()` stub returns keys unchanged, so error text is asserted on its i18n key.
+describe('Toolbar — XIN-1051 link popover enter / focus regression', () => {
+  function urlField(): HTMLInputElement {
+    return screen.getByPlaceholderText('docs.toolbar.linkPlaceholder') as HTMLInputElement
+  }
+  function textField(): HTMLInputElement {
+    return screen.getByPlaceholderText('docs.toolbar.linkText') as HTMLInputElement
+  }
+
+  it('path 1 — existing selection: Enter applies the link to the selection and closes the popover', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run() // select "hello"
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    const url = urlField()
+    fireEvent.change(url, { target: { value: 'https://example.com' } })
+    fireEvent.keyDown(url, { key: 'Enter' })
+
+    expect(document.querySelector('.octo-link-popover')).toBeNull()
+    const html = editor!.getHTML()
+    expect(html).toContain('example.com')
+    expect(html).toContain('>hello</a>')
+    // No newline leaked into the body: still exactly one paragraph.
+    expect(editor!.state.doc.childCount).toBe(1)
+  })
+
+  it('path 2 — no selection: Enter inserts a new linked label at the caret and closes', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().setTextSelection(6).run() // caret after "hello", no selection
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.change(textField(), { target: { value: 'Example' } })
+    const url = urlField()
+    fireEvent.change(url, { target: { value: 'https://example.com' } })
+    fireEvent.keyDown(url, { key: 'Enter' })
+
+    expect(document.querySelector('.octo-link-popover')).toBeNull()
+    const html = editor!.getHTML()
+    expect(html).toContain('>Example</a>')
+    expect(html).toContain('hello') // original text untouched
+    expect(editor!.state.doc.childCount).toBe(1)
+  })
+
+  it('path 3 — brand-new link: Enter in the text field also confirms (no body newline)', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().setTextSelection(6).run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.change(textField(), { target: { value: 'Docs' } })
+    fireEvent.change(urlField(), { target: { value: 'https://example.com' } })
+    // Confirm from the TEXT field's Enter handler (both inputs must guard Enter).
+    fireEvent.keyDown(textField(), { key: 'Enter' })
+
+    expect(document.querySelector('.octo-link-popover')).toBeNull()
+    expect(editor!.getHTML()).toContain('>Docs</a>')
+    expect(editor!.state.doc.childCount).toBe(1)
+  })
+
+  it('empty URL: Enter keeps the popover open with an inline error and never discards input or types into the body', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().setTextSelection(6).run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.change(textField(), { target: { value: 'Keep me' } })
+    const before = editor!.getHTML()
+    fireEvent.keyDown(urlField(), { key: 'Enter' })
+
+    // Popover stays open (the old code silently closed it) and shows the inline error…
+    expect(document.querySelector('.octo-link-popover')).toBeTruthy()
+    expect(screen.getByText('docs.toolbar.linkErrorEmpty')).toBeTruthy()
+    // …the typed text is preserved, no link is created, and the body is unchanged (no newline).
+    expect(textField().value).toBe('Keep me')
+    expect(editor!.getHTML()).toBe(before)
+    expect(editor!.getHTML()).not.toContain('</a>')
+  })
+
+  it('invalid URL: Enter surfaces an inline error and inserts nothing', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    const before = editor!.getHTML()
+    fireEvent.change(urlField(), { target: { value: 'javascript:alert(1)' } })
+    fireEvent.keyDown(urlField(), { key: 'Enter' })
+
+    expect(document.querySelector('.octo-link-popover')).toBeTruthy()
+    expect(screen.getByText('docs.toolbar.linkErrorInvalid')).toBeTruthy()
+    expect(editor!.getHTML()).toBe(before)
+    expect(editor!.getHTML()).not.toContain('</a>')
+  })
+
+  it('a scheme-less host (e.g. "example.com") is linked as https, not a same-origin path', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.change(urlField(), { target: { value: 'example.com' } })
+    fireEvent.keyDown(urlField(), { key: 'Enter' })
+
+    expect(editor!.getHTML()).toContain('href="https://example.com')
+  })
+
+  it('editing the URL after an error clears the inline error', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.keyDown(urlField(), { key: 'Enter' }) // empty → error
+    expect(screen.getByText('docs.toolbar.linkErrorEmpty')).toBeTruthy()
+    fireEvent.change(urlField(), { target: { value: 'h' } })
+    expect(document.querySelector('.octo-link-error')).toBeNull()
+  })
+
+  it('Escape from the URL field still closes the popover', () => {
+    render(<Toolbar editor={editor!} />)
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.keyDown(urlField(), { key: 'Escape' })
+    expect(document.querySelector('.octo-link-popover')).toBeNull()
+  })
+
+  it('focus isolation: opening over a selection moves focus into the URL field', async () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    await waitFor(() => expect(document.activeElement).toBe(urlField()))
+  })
+
+  it('focus isolation: opening a brand-new link (no selection) focuses the text field', async () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().setTextSelection(6).run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    await waitFor(() => expect(document.activeElement).toBe(textField()))
   })
 })
 
--- a/packages/docs/src/editor/Toolbar.test.tsx
+++ b/packages/docs/src/editor/Toolbar.test.tsx
@@ -227,6 +227,42 @@ describe('Toolbar — XIN-1051 link popover enter / focus regression', () => {
     expect(editor!.getHTML()).not.toContain('</a>')
   })
 
+  // XIN-1073 (real-machine 4a): a bare, scheme-less word ("abc", "test") is NOT a URL. The old
+  // code blindly prefixed https:// so sanitizeLinkHref returned https://abc/ — the popover accepted
+  // the junk, closed, and lost the input with no error. It must now behave like any other invalid
+  // URL: inline error, popover stays open, input preserved, nothing inserted.
+  it('scheme-less bare word (no dot): Enter shows the inline error, keeps the popover + input, inserts nothing', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run() // select "hello"
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    const before = editor!.getHTML()
+    fireEvent.change(urlField(), { target: { value: 'abc' } })
+    fireEvent.keyDown(urlField(), { key: 'Enter' })
+
+    expect(document.querySelector('.octo-link-popover')).toBeTruthy()
+    expect(screen.getByText('docs.toolbar.linkErrorInvalid')).toBeTruthy()
+    expect(urlField().value).toBe('abc') // typed value preserved for correction
+    expect(editor!.getHTML()).toBe(before) // no https://abc/ link created
+    expect(editor!.getHTML()).not.toContain('</a>')
+    expect(editor!.getHTML()).not.toContain('abc')
+  })
+
+  it('scheme-less bare word: retyping a valid host clears the error and links on the next Enter', () => {
+    render(<Toolbar editor={editor!} />)
+    editor!.chain().focus().selectAll().run()
+    fireEvent.click(titleBtn('docs.toolbar.link'))
+    fireEvent.change(urlField(), { target: { value: 'abc' } })
+    fireEvent.keyDown(urlField(), { key: 'Enter' })
+    expect(screen.getByText('docs.toolbar.linkErrorInvalid')).toBeTruthy()
+
+    // Correcting the input clears the error, and a valid host now links + closes.
+    fireEvent.change(urlField(), { target: { value: 'abc.com' } })
+    expect(document.querySelector('.octo-link-error')).toBeNull()
+    fireEvent.keyDown(urlField(), { key: 'Enter' })
+    expect(document.querySelector('.octo-link-popover')).toBeNull()
+    expect(editor!.getHTML()).toContain('href="https://abc.com')
+  })
+
   it('a scheme-less host (e.g. "example.com") is linked as https, not a same-origin path', () => {
     render(<Toolbar editor={editor!} />)
     editor!.chain().focus().selectAll().run()
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -1316,14 +1316,23 @@ export function Toolbar({ editor }: { editor: Editor }) {
     return () => cancelAnimationFrame(id)
   }, [linkOpen])
 
-  // XIN-1051: a bare host/domain with no scheme ("google.com") would otherwise resolve relative to
-  // the current origin and become a same-origin path link, not the external URL the user meant.
-  // Prepend https:// when there is no explicit scheme (and it is not protocol-relative) so the
-  // value sanitizes into a proper absolute link.
-  function normalizeLinkInput(raw: string): string {
+  // XIN-1051 / XIN-1073: resolve the raw popover input into a safe, absolute href — or null when it
+  // is not a usable link, so confirmLink can surface the inline error instead of inserting junk.
+  //   - explicit scheme ("https://x", "mailto:a@b") / protocol-relative ("//cdn/x") → hand straight
+  //     to sanitizeLinkHref so the §3.7 scheme whitelist still rejects javascript:/data:/ftp: etc.
+  //   - scheme-less: a bare host/domain ("google.com") would resolve relative to the origin and
+  //     become a same-origin path, so we prepend https:// — but ONLY when it actually looks like a
+  //     host (a dotted label, or "localhost"). A bare word like "abc" is NOT a URL: without this
+  //     guard https:// was blindly prepended and sanitizeLinkHref returned https://abc/, so the
+  //     popover accepted the junk and closed with no error (the 4a real-machine defect). Such input
+  //     now resolves to null → inline error, popover stays open, user's text is preserved.
+  function resolveLinkHref(raw: string): string | null {
     const v = raw.trim()
-    if (!v || v.startsWith('//') || /^[a-z][a-z0-9+.-]*:/i.test(v)) return v
-    return `https://${v}`
+    if (!v) return null
+    if (v.startsWith('//') || /^[a-z][a-z0-9+.-]*:/i.test(v)) return sanitizeLinkHref(v)
+    const host = v.split(/[/?#]/, 1)[0]
+    const looksLikeHost = host === 'localhost' || /[^.\s]\.[^.\s]/.test(host)
+    return looksLikeHost ? sanitizeLinkHref(`https://${v}`) : null
   }
 
   // C7: insert a link at the cursor (or apply it to the selection). With no selection a brand-new
@@ -1339,7 +1348,7 @@ export function Toolbar({ editor }: { editor: Editor }) {
       linkUrlRef.current?.focus()
       return
     }
-    const href = sanitizeLinkHref(normalizeLinkInput(raw))
+    const href = resolveLinkHref(raw)
     if (!href) {
       setLinkError(t('docs.toolbar.linkErrorInvalid'))
       linkUrlRef.current?.focus()
--- a/packages/docs/src/editor/Toolbar.tsx
+++ b/packages/docs/src/editor/Toolbar.tsx
@@ -85,24 +85,40 @@ const IconCode = () => (
     <path d="M8.7 17.3 3.4 12l5.3-5.3 1.3 1.4L5.9 12l4.1 4.1-1.3 1.2zm6.6 0L14 16.1l4.1-4.1-4.1-4.1 1.3-1.4L20.6 12l-5.3 5.3zM13.9 5.2l1.9.5-3.9 13.1-1.9-.5 3.9-13.1z" />
   </svg>
 )
-// Link: two interlocking pill-shaped rings linked at ~45° (classic chain-link, boss reference).
-// Filled rings via the evenodd fill-rule (outer capsule minus an inner capsule = hollow ring);
-// the whole pair is rotated 45° so the links sit on the diagonal.
+// Link (XIN-1051): the standard chain-link glyph (lucide `link`) — two diagonal, interlocking
+// hooked curves. Stroke line-art, not filled: uses .octo-tb-icon-stroke (fill:none;
+// stroke:currentColor) with round caps/joins so it reads as a recognizable link icon at 16px
+// rather than the old two-capsule filled blob. Aligned with IconUnlink below.
 const IconLink = () => (
-  <svg className="octo-tb-icon" viewBox="0 0 24 24" aria-hidden="true">
-    <g transform="rotate(45 12 12)" fillRule="evenodd">
-      <path d="M4 8.6h9a3.4 3.4 0 0 1 0 6.8H4a3.4 3.4 0 0 1 0-6.8zm0 1.6a1.8 1.8 0 0 0 0 3.6h9a1.8 1.8 0 0 0 0-3.6H4z" />
-      <path d="M11 8.6h9a3.4 3.4 0 0 1 0 6.8h-9a3.4 3.4 0 0 1 0-6.8zm0 1.6a1.8 1.8 0 0 0 0 3.6h9a1.8 1.8 0 0 0 0-3.6h-9z" />
-    </g>
+  <svg
+    className="octo-tb-icon-stroke"
+    viewBox="0 0 24 24"
+    aria-hidden="true"
+    strokeWidth={2}
+    strokeLinecap="round"
+    strokeLinejoin="round"
+  >
+    <path d="M10 13a5 5 0 0 0 7.54.54l3-3a5 5 0 0 0-7.07-7.07l-1.72 1.71" />
+    <path d="M14 11a5 5 0 0 0-7.54-.54l-3 3a5 5 0 0 0 7.07 7.07l1.71-1.71" />
   </svg>
 )
-// Unlink: the same two rings pulled apart with a gap between them (broken chain).
+// Unlink (XIN-1051): the same chain pulled apart (lucide `unlink`) — the two hooked curves with a
+// break plus the four short "snap" ticks. Same stroke style as IconLink so the pair reads as a set.
 const IconUnlink = () => (
-  <svg className="octo-tb-icon" viewBox="0 0 24 24" aria-hidden="true">
-    <g transform="rotate(45 12 12)" fillRule="evenodd">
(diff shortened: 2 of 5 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #762 (octo-web)

Reviewed at head `8ba8f0f64249e6a83411aff3d288274858b90fc0`.

## Summary

This PR fixes two real usability bugs in the docs editor's link popover (XIN-1051): Enter keydown events leaked from the popover inputs into ProseMirror (inserting a newline instead of confirming the link), and empty/invalid URLs silently discarded the popover — losing whatever the user had typed. The fix adds `e.stopPropagation()` on both inputs' keydown handlers, introduces inline URL validation with a `linkError` state, auto-prefixes `https://` for bare domain inputs, and manages focus explicitly via `useEffect` + `requestAnimationFrame` so the caret is pulled out of the editor when the popover opens. A secondary change refreshes the link/unlink toolbar icons from filled capsule rings to lucide-style stroke glyphs (`.octo-tb-icon-stroke`).

The changes are well-targeted, the test suite is thorough (11 new test cases covering all three confirm paths, error states, and focus isolation), and the URL sanitization chain (`normalizeLinkInput` → `sanitizeLinkHref`) is safe against scheme whitelist bypass. No correctness, security, or runtime-path defects found.

## Verification

Static analysis only at head `8ba8f0f64249e6a83411aff3d288274858b90fc0`; build and tests not executed in this environment.

- ✅ **Data flow: confirmLink()** — `raw = linkValue.trim()`, empty check before normalization, then `sanitizeLinkHref(normalizeLinkInput(raw))` rejects unsafe schemes. Fallback text (`linkText.trim() || raw`) is identical in semantics to the old `linkValue.trim()`. No path produces a link without passing the whitelist.
- ✅ **normalizeLinkInput regex boundaries** — `/^[a-z][a-z0-9+.-]*:/i` correctly identifies real schemes (http:, https:, mailto:). `javascript:`, `data:`, `vbscript:` pass through unchanged and are rejected by `sanitizeLinkHref`'s whitelist. Bare domains (`example.com`) get `https://` prepended. Protocol-relative `//` inputs pass through to `sanitizeLinkHref` which resolves them against origin (existing behavior, unchanged).
- ✅ **Focus isolation** — `useEffect` on `linkOpen` transition uses `requestAnimationFrame` to defer focus until the input is mounted. `linkHadSelectionRef` correctly gates the target (URL field for existing selection, text field for new link). Cleanup cancels the rAF on unmount/toggle.
- ✅ **Error lifecycle** — `linkError` set on empty/invalid confirm, cleared on URL input change, cleared on popover open, cleared on close. No stale error can persist across popover sessions.
- ✅ **Icon parity (C1)** — Both `IconLink` and `IconUnlink` switched to `.octo-tb-icon-stroke` with consistent stroke attributes (`strokeWidth={2}`, `strokeLinecap="round"`, `strokeLinejoin="round"`). The CSS class is defined at `:1293`. Test assertion updated.
- ✅ **i18n parity** — Both `linkErrorEmpty` and `linkErrorInvalid` present in en-US and zh-CN at identical nesting depth.
- ✅ **Test coverage** — 11 new cases cover: all three confirm paths (selection, no-selection, text-field Enter), empty URL error, invalid URL error (javascript:), scheme-less domain normalization, error clearing, Escape close, and both focus isolation targets.

## Findings

### Nit — `normalizeLinkInput` could be module-scoped (`packages/docs/src/editor/Toolbar.tsx:1323`)

The function is a pure transform with no closure captures. Defining it inside the component body recreates it on every render. Lifting it to module scope alongside `sanitizeLinkHref` (imported from `sanitize.ts`) would be more consistent and avoid the redundant allocation. Not blocking.

### Nit — `host:port` inputs rejected instead of auto-prefixed (`packages/docs/src/editor/Toolbar.tsx:1325`)

The scheme-detection regex `/^[a-z][a-z0-9+.-]*:/i` matches `localhost:3000` as if `localhost:` were a URI scheme. `normalizeLinkInput` returns it unchanged, and `sanitizeLinkHref` rejects it (the `localhost:` pseudo-scheme is not whitelisted). The end result is safe — the user sees the "Enter a valid URL" inline error — but the UX is suboptimal: a `host:port` input that a human would expect to become `http://localhost:3000` is rejected rather than auto-prefixed. Tightening the regex to require `://` after the scheme token would close this gap without false positives on real schemes. Not blocking.

- `e.stopPropagation()` on both Enter and Escape in both inputs — prevents keydown bubbling to ProseMirror's contenteditable.
- `autoFocus` removed from URL input — replaced by the more precise `useEffect` focus management that handles both selection/no-selection cases.
- `onMouseDown` stopPropagation on both inputs — prevents the outside-click handler from closing the popover on internal clicks.
- `aria-invalid` and `role="alert"` on the error span — proper accessibility wiring.
- `closeLink()` resets all four state fields (`linkOpen`, `linkText`, `linkValue`, `linkError`).
- The "Set" button click path goes through the same `confirmLink()` — no bypass of validation.
- No new external dependencies; the lucide SVG paths are inlined.

## Verdict: APPROVED

Well-executed bug fix with thorough test coverage and no correctness or security concerns. The two nits are non-blocking polish items.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. The link popover keyboard isolation, URL validation, and icon refresh are correctly implemented with comprehensive test coverage.

yujiawei (changes requested):
# Code Review — PR #762 (octo-web)

Reviewed at head SHA `8ba8f0f6`. Checked out and ran the docs package suite locally.

## Summary

Solid, well-tested fix for a genuinely broken popover. Focus isolation, `stopPropagation()` on both inputs, and the empty/unsafe-URL inline error all land correctly, and the new regression tests cover the three acceptance paths plus the error/escape/focus cases. All 60 Toolbar tests pass; the changed files typecheck clean (the two pre-existing `tsc` errors are in `dmworkbase/i18n/format.ts` and `members/sort.ts`, untouched by this PR).

One blocking correctness issue in the new `normalizeLinkInput` helper, described below.

## Blocking

### P1 — `normalizeLinkInput` mangles relative / fragment links (silent wrong host)

`Toolbar.tsx:1323-1327`

```js
function normalizeLinkInput(raw: string): string {
  const v = raw.trim()
  if (!v || v.startsWith('//') || /^[a-z][a-z0-9+.-]*:/i.test(v)) return v
  return `https://${v}`
}
```

The guard only skips protocol-relative (`//`) and explicitly-schemed input. It does **not** skip an absolute path (`/…`), a fragment (`#…`), or a query (`?…`), so `https://` is prepended to all of them. Combined with `sanitizeLinkHref` (which resolves relative input against the current origin), this changes previously-working behavior:

| Input | Before this PR (`sanitizeLinkHref(raw)`) | After this PR (`sanitizeLinkHref(normalizeLinkInput(raw))`) |
|-------|------------------------------------------|------------------------------------------------------------|
| `/docs/page` | `https://<origin>/docs/page` (valid same-origin link) | `https://docs/page` — **wrong host, silently created, no error** |
| `#section` | `https://<origin>/#section` (valid fragment link) | `https://#section` → `null` → rejected as "invalid URL" |
| `example.com` | `https://<origin>/example.com` (was the bug) | `https://example.com/` ✅ (the intended fix) |

Runtime-verified with the exact PR code against `sanitize.ts`.

The `/docs/page → https://docs/page` case is the concerning one: it produces a link to a completely different host with **no error shown to the user**. That directly contradicts this PR's own stated contract ("a mistyped URL is never lost … surfaces an inline error"). It's an in-scope regression the new helper introduces, not a pre-existing issue.

**Fix** (keeps origin-relative resolution for path/fragment/query while still upgrading bare hosts):

```js
if (!v || v.startsWith('//') || v.startsWith('/') || v.startsWith('#') || v.startsWith('?') || /^[a-z][a-z0-9+.-]*:/i.test(v)) return v
```

Please also add regression cases for `/path` and `#anchor` to `Toolbar.test.tsx` so this stays covered.

(Note: `localhost:3000`-style host:port input resolves to `null` both before and after this PR — the leading-token `/^[a-z][a-z0-9+.-]*:/` regex reads `localhost:` as a scheme. That is pre-existing, not introduced here, so out of scope for this PR.)

## Non-blocking

### P2 — ref mutation inside the `setLinkOpen` state updater

`Toolbar.tsx:1283-1294`

`linkHadSelectionRef.current = hasSelection` is assigned inside the functional updater passed to `setLinkOpen`. State updaters should be pure; under StrictMode double-invocation this runs twice. It's harmless here (idempotent, same value both times), but cleaner to compute `hasSelection` and set the ref outside the updater, then just `return !v` inside it.

### P2 — `select()` on focus of an empty text field

`Toolbar.tsx:1312-1315`

Focusing then `select()`-ing is fine when there's existing content, but for a brand-new link the text field is empty so `select()` is a no-op — no issue, just noting it does nothing in that path.

## Verified working

- Focus isolation: URL field on existing selection, text field for a brand-new link (both covered by tests).
- `stopPropagation()` on both inputs' Enter/Escape prevents the keystroke reaching ProseMirror — no stray newline in the body.
- Empty and unsafe (`javascript:`) URLs surface the inline error, preserve input, and insert nothing.
- The text input already has `onMouseDown` stopPropagation (line 1443), matching the URL input.
- Icon swap to the stroke chain glyph; `.octo-tb-icon-stroke` already exists and is used by undo/redo.
- i18n keys added to both `en-US` and `zh-CN`.

## Verdict

Requesting changes for the single P1 regression in `normalizeLinkInput`. Once relative/fragment inputs are guarded (and covered by a test), this is good to merge — the rest of the change is clean and well-tested.

Jerry-Xin (dismissed):
# Code Review — PR #762

Reviewed at head `8ba8f0f6`. Draft PR; independent pass over the link-popover keyboard-safety fix and the icon refresh.

## Summary

The keyboard-safety fix is well-built and well-tested: Enter no longer leaks a newline into the body (preventDefault + stopPropagation on both inputs), Escape closes the popover, focus is pulled into the correct field on open, and empty/unsafe URLs now surface an inline error instead of silently discarding the popover. The dangerous-scheme defense is intact — the link value is run through the scheme whitelist before any link is created, so a javascript/data/vbscript scheme resolves to null and inserts nothing (confirmed by the invalid-URL test). Icon swap and i18n are clean.

One blocking correctness regression in the new input-normalizer, independently reproduced.

## Blocking

### 🔴 P1 — the new URL normalizer breaks relative / fragment / query links (silent wrong host)

`packages/docs/src/editor/Toolbar.tsx:1323-1327` (`normalizeLinkInput`)

The prepend guard only skips protocol-relative input and input that already carries an explicit scheme. It does **not** skip an absolute path (leading slash), a fragment (leading hash), or a query (leading question mark), so the https prefix is prepended to those too. Because the sanitizer resolves relative input against the current origin, this changes previously-working behavior. Reproduced against the real sanitizer with the exact PR code:

- An absolute-path input such as a same-origin doc path used to resolve to a valid same-origin URL. It now becomes a link to a **different host** (the first path segment is parsed as the hostname), created silently with **no error shown**. This is the concerning case — it directly contradicts this PR's own stated contract that a mistyped URL is never lost.
- A fragment-only input and a query-only input, which previously produced valid same-origin links, now normalize to an unparseable value and are rejected as "invalid URL".
- The intended bare-host case (a scheme-less domain) does upgrade to https correctly — the fix itself works.

This is an in-scope regression introduced by the new helper, not a pre-existing issue.

**Suggested fix:** also short-circuit inputs beginning with a slash, a hash, or a question mark (alongside the existing protocol-relative and explicit-scheme checks), so origin-relative resolution is preserved for path/fragment/query while bare hosts still upgrade to https. Please add regression cases for an absolute-path input and a fragment input to the test file so this stays covered.

## Non-blocking

### 🟡 Focus trap / focus restore (a11y polish)

The popover moves focus in on open and closes on Escape, which covers the core fix. Two a11y gaps remain, neither breaks the feature:
- No focus trap: Tab from the last field can move focus out of the dialog to the page behind rather than cycling within the popover.
- Escape/close does not restore focus to the trigger button.

Worth a follow-up given the dialog role, but not blocking this fix.

### 🟡 Ref write inside the state updater

`packages/docs/src/editor/Toolbar.tsx:1283-1294`

The had-selection ref is assigned inside the functional state updater. Updaters should be pure; under StrictMode double-invocation this runs twice. Harmless here (idempotent, same value), but cleaner to compute the selection flag and set the ref outside the updater.

## Verified working

- Dangerous-scheme defense: the link value passes through the scheme whitelist before any link is created; an unsafe scheme resolves to null and inserts nothing (invalid-URL test).
- Enter on either input is prevented/stopped so it never reaches the editor as a newline; Escape closes; focus lands in the right field on open (all covered by tests).
- Keydown handling is per-input and scoped — no global listener added, stopPropagation limited to Enter/Escape, and the focus animation frame is cancelled on cleanup. No listener leak, no global key swallowing.
- Icon swap is a static inline stroke glyph reusing an existing class — no user-controlled content.
- Error strings go through the translation layer; keys added to both locales.

## Verdict

Requesting changes for the single P1 normalizer regression. Once relative/fragment/query inputs are short-circuited (and covered by tests), the change is good — the rest is clean and well-tested.

boris-clark (approved):
Verified against the head branch, not the commit message.

- Enter guarding: both popover inputs call `preventDefault()` + `stopPropagation()` on Enter and Escape, so the contenteditable never receives the keystroke and the popover no longer loses its input.
- Empty/invalid URL is no longer silently dropped: `confirmLink()` sets `linkError` and returns early instead of closing, and an inline `role="alert"` error renders in the popover.
- Focus handled across the three paths: `linkHadSelectionRef` set in `openLink()`, and the `linkOpen` effect focuses the URL field when a selection exists, otherwise the text field, via `requestAnimationFrame` (the old `autoFocus` was intentionally removed).
- Icon: `IconLink`/`IconUnlink` use `className="octo-tb-icon-stroke"` (the existing `fill:none; stroke:currentColor` stroke variant), not a filled blob; the old filled paths were removed.
- Tests genuinely assert behavior (popover state, resulting `getHTML()` including href, no leaked newline, error-key presence, error-clear-on-edit), not just render.

Non-blocking:
- `normalizeLinkInput`'s scheme regex treats a bare `host:port` (e.g. `example.com:8080`) as already-schemed, so it surfaces the invalid-URL error rather than prepending `https://`. Rare and fails safe.

LGTM.

OctoBoooot:
@/tmp/pr762-advisory-8ba8f0f6.md

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. The link popover keyboard-safety fix and icon refresh are well-scoped, correctly implemented, and thoroughly tested.

boris-clark (approved):
Re-review of head `729a07ca` — the 4a fix and 4b non-regression.

**Verdict: APPROVED.**

**4a — scheme-less bare word no longer inserts junk.** `resolveLinkHref` now only prepends `https://` when the input actually looks like a host: a dotted label (`/[^.\s]\.[^.\s]/`) or `localhost`. A bare word (`abc`, `test`) resolves to `null`, so `confirmLink` surfaces the inline error, keeps the popover open, and preserves the typed text — matching the real-machine 4a defect report. The old unconditional `https://` prepend that made `sanitizeLinkHref` return `https://abc/` is gone; the only remaining prepend is the host-gated one.

**Security preserved.** Explicit-scheme and protocol-relative inputs still route straight through `sanitizeLinkHref`, so the §3.7 whitelist keeps rejecting `javascript:`/`data:`/`ftp:` etc. → inline error, nothing inserted.

**4b — link/unlink icons not reverted.** Both still use the stroke-style lucide glyphs on `.octo-tb-icon-stroke` (class present in styles.css). No regression.

**Regression tests are real assertions**, covering every acceptance path:
- bare word → inline error + input preserved + nothing inserted + no `https://abc/`;
- correcting to a valid host → error clears, link inserted + popover closes;
- `example.com` → linked as `https://`, not a same-origin path;
- empty / `javascript:` → inline error, body unchanged;
- Enter from both fields confirms without leaking a newline into the body; Escape closes; focus isolation moves focus into the popover.

All 62 Toolbar tests pass. The pre-existing EditorShell / standalone baseline failures are in files this PR does not touch — unrelated to this change.

yujiawei (approved):
# Code Review — PR #762 (octo-web)

Reviewed at head SHA `8ba8f0f6`, diffed against `main` (merge-base `8073e2be`). Scope: 5 files, +256/-24, all under `packages/docs/src/editor` + i18n.

## Verdict: Approve

Focused, well-scoped fix. It does exactly what the description says — keyboard/focus correctness in the link popover, inline validation, scheme completion, and an icon swap — with no scope creep. All `packages/docs` tests pass (60/60, including 11 new regression tests covering the three acceptance paths plus the error/focus behaviors). No P0/P1 blockers. Two P2 advisories below are worth a look but do not block merge.

## What's correct

- **Keystroke isolation** — adding `stopPropagation()` alongside `preventDefault()` on both inputs' Enter/Escape handlers is the right fix for the "Enter inserts a newline in the body" bug. Combined with focus moving into the popover on open, Enter/Escape can no longer reach ProseMirror. Both inputs now guard Enter, so confirming from either field works.
- **No more silent discard** — the old `confirmLink` called `closeLink()` on a falsy href, so a mistyped URL vanished. Now an empty URL and an unsafe/invalid URL each surface an inline error, preserve the typed input, and re-focus the URL field. Verified against `sanitizeLinkHref` (scheme whitelist http/https/mailto): `javascript:` is rejected with the invalid-URL message, matching the test.
- **Scheme completion** — `normalizeLinkInput` correctly leaves protocol-relative (`//cdn…`) and already-schemed inputs (`http:`, `mailto:`) untouched and only prepends `https://` to bare hosts. Confirmed empirically: `example.com` → `https://example.com/`, `mailto:a@b.com` unchanged, `//cdn.com/a.js` → `https://cdn.com/a.js`.
- **Icon swap** — `octo-tb-icon-stroke` is a pre-existing class (already used by undo/redo), so `IconLink`/`IconUnlink` now visually match the rest of the stroke-style set. The batch-7 test was updated to assert the new class; nothing else references the old filled `svg.octo-tb-icon` on the link button.

## P2 (advisory — non-blocking)

1. **Relative / anchor / query-only hrefs now rejected** — `Toolbar.tsx` `normalizeLinkInput` + `confirmLink`. A bare `#anchor`, `?tab=1`, or `/relative/path` no longer produces a link: `#conclusion` normalizes to `https://#conclusion` → rejected, and the user sees "Enter a valid URL". Previously these resolved against the current origin and became an absolute app-origin URL. For a rich-text editor the external-link case is overwhelmingly the common one, and the old app-origin result was arguably not what the user wanted either, so this is a reasonable tradeoff rather than a regression of a working feature. Flagging only so it's an explicit decision — if in-document anchors are a supported use case, consider allowing a leading `#`/`/`/`?` through untouched.

2. **Ref/state mutation inside a state updater** — `Toolbar.tsx` `openLink`: `linkHadSelectionRef.current = hasSelection` and several `setLinkText/setLinkValue/setLinkError` calls run inside the `setLinkOpen((v) => { … })` updater. Updater functions are expected to be pure; nesting `setState` calls and a ref write inside one works here (the derivation is idempotent and the toggle is user-driven, not concurrent-rendered) but is a smell that can bite under StrictMode double-invocation. Cleaner to compute `next` in the updater and do the side effects in the `if (next)` branch outside, or in an effect keyed on `linkOpen`.

## Coverage note

- Verified locally: `packages/docs` unit tests (60 pass), `normalizeLinkInput` + `sanitizeLinkHref` behavior across ~15 edge inputs (anchors, IPv6, `user@host`, whitespace, mixed-case scheme), and that the two repo-wide `tsc` errors (`i18n/format.ts`, `members/sort.ts`) pre-exist on `main` and are untouched by this PR — the changed files typecheck clean.
- Not verified from the diff alone: real-browser focus timing of the `requestAnimationFrame` focus call across browsers (the PR reports manual Chromium verification); interaction with the editor's own link-paste/auto-link extension config; and visual rendering of the new stroke glyph at 16px on the actual toolbar.

OctoBoooot:
@/tmp/pr762-update-729a07ca.md

OctoBoooot (commented):
## Review: make link popover keyboard-safe and refresh the link icon (#762)

**Verdict: Comment** (not approve, not request-changes). Now out of draft; verified at head `729a07cabf71`. The original blocker I raised (silent wrong-host URL corruption) is **fixed** — so this is no longer REQUEST_CHANGES. But a real regression (🟡) and two keyboard-safety gaps remain, so it's not a clean approve either. No blocker, no major → COMMENT, naming what's left. This supersedes my draft-time advisory notes.

### Fixed — the silent wrong-host corruption (my earlier blocker)
The old normalizer blindly prepended `https://` to scheme-less input, so `/docs/page → https://docs/page` (wrong host, silently accepted). The rewrite (`resolveLinkHref`) fixes it — ran it at this head:
- `/docs/page` → null → inline error (was the silent `https://docs/page`). ✅
- `abc` (bare word) → null (was `https://abc/`). ✅
- `google.com` → `https://google.com/`, `example.com/a` → `https://example.com/a` (scheme-less-host fix still works). ✅
- XSS gate intact: `javascript:alert(1)` → null; `confirmLink` surfaces an inline error instead of inserting junk or silently closing. ✅

The dangerous dimension is resolved. Good.

### 🟡 — relative / fragment / query links regressed (they worked pre-PR)
Verified against base `main`: the old `confirmLink` fed raw input to `sanitizeLinkHref`, which origin-resolves via `new URL(raw, ORIGIN)` — so `/docs/page`, `#section`, `?q=1` were **valid links before this PR**. At this head `resolveLinkHref` returns null for all of them (and `localhost:3000`), so they now show an "invalid URL" error. Still a regression of a previously-working input class, but severity is 🟡 not 🔴: it **fails visibly** (error shown, user's text preserved, nothing corrupted or mis-targeted). Fix: pass leading `/`, `#`, `?` straight to `sanitizeLinkHref` (it already origin-resolves them safely), and accept `host:port`.

### Minor — two keyboard-safety gaps (the PR's headline is "keyboard-safe")
1. **No focus restoration on close** — `closeLink()`/`confirmLink()` don't return focus to the editor (verified: 0 activeElement/restore refs). After Escape/confirm, focus drops to `document.body`; a keyboard user loses their place. Recoverable (click back), so minor — but it's the completeness the title promises.
2. **No Tab-trap despite `role="dialog"`** — no focus trap across the two inputs (verified: 0 Tab-key handlers). Tab escapes the dialog; the ARIA role advertises a trap that isn't implemented. Either add a small Tab/Shift-Tab wrap or drop `role="dialog"`.

### Clean
Icons are inline `<svg>` + `aria-hidden` (no injection); i18n keys present + symmetric; the focus-**in** handling is correct (rAF into the right field on open; both inputs `preventDefault`+`stopPropagation` on Enter/Escape so keystrokes don't fall through to ProseMirror); tests cover the acceptance paths + the new bare-word rejection. CI green (the one "cancelled" Build is a superseded run, followed by skipped — not a failure).

### Net
The rewrite closed the silent-corruption 🔴 (wrong-host gone, XSS gate intact, errors surface) — real progress. Holding at **COMMENT** for one 🟡 (relative/fragment/query links regressed vs pre-PR — now failing-visibly, not corrupting) and two keyboard-safety minors (focus-restore, Tab-trap) that the "keyboard-safe" title sets the bar for. Restore relative/fragment support + close the two focus gaps and this is a clean approve. `reviewDecision` currently shows CHANGES_REQUESTED from Jerry-Xin's CR still anchored at the prior head `8ba8f0f6` (pre-fix) — that'll reconcile when he re-anchors; my verdict here is COMMENT on the current bytes. Credit ploy-elison for closing the silent-corruption path, and yujiawei/Jerry-Xin for the original catch.

Jerry-Xin (approved):
# Code Review — PR #762 (re-review at head 729a07ca)

Delta re-review. Head advanced from the previous review commit to `729a07ca` with a single follow-up commit (`ahead 1, behind 0` — no rebase drift). True delta is two files: the toolbar component and its test.

## Resolution of the prior blocking item

My previous blocker was the new scheme-less input normalizer: a leading-slash same-origin input was silently upgraded with an https prefix so the first path segment became the host — a wrong-host link created with no error shown — while fragment-only and query-only inputs were rejected. The severity driver was the **silent** wrong-host corruption that contradicted this fix's own "a mistyped URL is never lost" contract.

The follow-up commit reworks the helper (now returns a nullable href). The decisive behaviors, traced against the real sanitizer at this head:

- **Bare host input** (a dotted domain, or localhost): recognized as a host and upgraded to an https link — the intended fix, now covered by tests.
- **Scheme-less non-host bare word** (no dot): now resolves to null → inline error, popover stays open, typed text preserved, nothing inserted. This was the real-machine defect (a bare word was previously accepted as an https junk link) — fixed and tested.
- **Leading-slash input / fragment-only input / query-only input**: the helper no longer classifies these as a host, so they resolve to null → inline error, input preserved, **nothing inserted**. The silent wrong-host outcome is gone; there is no longer any silent corruption.

The corruption severity driver of my blocker is resolved: no input is silently turned into a wrong-host link anymore. Treating link-popover input strictly as external URLs (and surfacing an inline error otherwise) is consistent with this change's own stated contract — the existing test explicitly encodes that a scheme-less host must become an external https link and **not** a same-origin path, i.e. same-origin-path resolution was the defect being removed, not a supported feature. Rejecting leading-slash/fragment/query input with a visible error (rather than resolving it to a same-origin link) is therefore an acceptable, intentional narrowing, not a regression. Lifting the block.

## Non-blocking note

Link-popover input no longer supports origin-relative forms (leading slash, fragment-only, query-only); such input surfaces the inline invalid-URL error. This is aligned with the external-URL contract above and is called out only so the narrowing is intentional and documented. If origin-relative linking is ever desired as a feature, it would be a separate follow-up.

## Verified (no regression)

- **Scheme allowlist / XSS defense intact**: every accepted branch still routes through the downstream link sanitizer, which enforces the http/https/mailto scheme whitelist and rejects pseudo-schemes; an unsafe scheme resolves to null and inserts nothing.
- **Keyboard-safety unchanged**: Enter on either field is prevented/stopped so no newline leaks into the body; Escape closes; focus lands in the correct field on open — all still covered by tests.
- **Link icon refresh**: static inline glyph, no user-controlled content.
- **Tests**: two new cases cover the bare-word reject-then-correct flow; the scheme-less-host-becomes-https case remains.

## CI / gate

All required checks pass; the code-review gate is green. PR remains a draft.

## Verdict

The single prior blocking item is resolved at this head. Approving.

OctoBoooot:
@/tmp/pr762-reconcile-729a07ca.md

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #762 (octo-web)

## Summary

This PR fixes two confirmed production defects in the `packages/docs` toolbar link popover: (1) Enter/Escape keystrokes leaking through to the ProseMirror editor, causing lost links or stray newlines in the document body, and (2) empty or unsafe URLs silently closing the popover with no feedback. The fix adds focus isolation (moving keyboard focus into the popover on open), `stopPropagation()` on both inputs' key handlers, inline validation with error surfacing, and scheme completion for scheme-less host input. The link/unlink icons are also refreshed from filled capsule glyphs to standard stroke chain-link icons.

The keyboard-safety and validation logic is well-structured. `sanitizeLinkHref` (the existing scheme-whitelist gate in `packages/docs/src/editor/sanitize.ts:22`) remains the final security boundary and is unchanged — the new `resolveLinkHref` pre-filter only adds UX-layer validation before it. No correctness, security, or build concerns found.

## Verification

Static analysis only at head `729a07ca`; build and tests not executed in this environment.

- ✅ **Data flow: `resolveLinkHref` → `sanitizeLinkHref`** — every code path through the new resolver ultimately delegates to `sanitizeLinkHref`, which enforces the `http:/https:/mailto:` scheme whitelist (`packages/docs/src/editor/sanitize.ts:24-31`). No bypass is possible. Traced: `javascript:alert(1)` → scheme detected → whitelist rejects → `null`; `data:text/html,...` → same; `abc` (bare word, no dot) → `looksLikeHost` false → `null`; `example.com` → host detected → `https://` prepended → whitelist accepts; `localhost` → explicit match → `https://` prepended → whitelist accepts.
- ✅ **Focus isolation** — `useEffect` on `[linkOpen]` uses `requestAnimationFrame` to focus the correct input after mount, with proper cleanup via `cancelAnimationFrame` (`packages/docs/src/editor/Toolbar.tsx:1308-1318`). The old `autoFocus` attribute is removed from the URL input, replaced by the ref-based approach that conditionally picks the target based on `linkHadSelectionRef`.
- ✅ **stopPropagation correctness** — both inputs' Enter/Escape handlers now call `e.stopPropagation()` after `e.preventDefault()` (`:1453-1464`, `:1473-1484`). This prevents the event from bubbling to the editor's `contenteditable` div where ProseMirror's keymap plugin would otherwise handle it. ProseMirror's DOM listeners are bubble-phase on the editable div (the popover input's ancestor), so stopping propagation at the input is sufficient.
- ✅ **Error state lifecycle** — `linkError` is cleared on: popover open (`:1290`), popover close (`:1301`), URL input change (`:1477`), successful confirm (`:1350`). No stale-error path exists.
- ✅ **i18n parity** — both `en-US.json:156-157` and `zh-CN.json:156-157` add the same two keys (`linkErrorEmpty`, `linkErrorInvalid`) at the same location in the `toolbar` namespace.
- ✅ **Icon changes** — `IconLink` and `IconUnlink` now use `className="octo-tb-icon-stroke"` with `strokeWidth=2`, `strokeLinecap="round"`, `strokeLinejoin="round"`. The existing test is updated to assert `svg.octo-tb-icon-stroke` instead of `svg.octo-tb-icon` (`:108-109`).
- ✅ **Test coverage** — 13 new test cases cover: three acceptance paths (existing selection / no selection / brand-new link), empty URL, invalid URL (`javascript:`), scheme-less bare word rejection, scheme-less host acceptance, error clearing on edit, Escape close, and focus isolation for both selection/no-selection opens.

## Findings

No P0/P1 issues. One P2 test-coverage gap below.

### P2 — Escape from the text field is not tested (`packages/docs/src/editor/Toolbar.test.tsx:290`)

The test suite covers `Escape` from the URL field ("Escape from the URL field still closes the popover") but has no counterpart firing Escape from the text field. Both inputs independently guard Escape with `stopPropagation()` (`packages/docs/src/editor/Toolbar.tsx:1457-1459` for the text field, `:1481-1483` for the URL field). A regression in the text field's Escape handler — for example, accidentally dropping `stopPropagation()` — would let the keystroke reach the editor and go undetected.

A one-liner addition to the test suite would close the gap:

```ts
it('Escape from the text field closes the popover', () => {
  render(<Toolbar editor={editor!} />)
  fireEvent.click(titleBtn('docs.toolbar.link'))
  fireEvent.keyDown(textField(), { key: 'Escape' })
  expect(document.querySelector('.octo-link-popover')).toBeNull()
})
```

- `resolveLinkHref` regex `/^[a-z][a-z0-9+.-]*:/i` — no ReDoS risk; linear, non-nested quantifiers; input is user-typed (short).
- `looksLikeHost` regex `/[^.\s]\.[^.\s]/` — correctly rejects bare words (`abc`), accepts dotted hosts (`a.b`, `sub.example.com`), accepts `localhost` via explicit string comparison.
- `confirmLink` empty-URL early return preserves typed text and linkText state; does not call `closeLink()`, so the popover stays open with the inline error.
- `linkHadSelectionRef.current` is written synchronously inside the `setLinkOpen` updater before the `useEffect` reads it on the next render — timing is correct.
- The `Set` button's `onClick={confirmLink}` is unchanged; keyboard users confirm via Enter on the inputs, not the button, so no additional key handler is needed there.
- CSS additions (`.is-invalid`, `.octo-link-error`) use a hardcoded `#eb5757` for the error colour rather than a CSS variable — minor style choice, no functional impact.

## Verdict: APPROVED

Solid keyboard-safety fix with thorough regression coverage. The `sanitizeLinkHref` whitelist remains the security boundary and is not weakened by the new pre-filter. One P2 test gap (Escape from the text field) is non-blocking.

yujiawei (approved):
# Code Review — PR #762 (octo-web)

**Scope:** `fix(docs): make link popover keyboard-safe and refresh the link icon`
Two commits: keyboard-safe popover + inline URL validation (XIN-1051), and rejection of scheme-less non-host input (XIN-1073).

Verification performed on head `729a07ca` against merge-base `8073e2be`: full docs test suite (`vitest run`, 62/62 pass), i18n parity + sanitize suites (16/16 pass), and `tsc --noEmit` (only two pre-existing errors in files untouched by this PR — `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts` — nothing from this diff). I also ran the URL-resolution logic through a matrix of edge cases.

## 1. Correctness / spec compliance ✅

The diff implements exactly what the fix targets, with no scope creep:

- **Enter no longer leaks into the body.** Both inputs now `e.preventDefault()` + `e.stopPropagation()` on Enter/Escape (`Toolbar.tsx:1456-1467`, `1489-1499`), and focus is pulled into the popover via the `requestAnimationFrame` effect (`Toolbar.tsx:1307-1316`), so the keystroke can no longer fall through to ProseMirror. The three acceptance paths (existing selection / no selection / brand-new link) are covered by regression tests.
- **Empty / unsafe URL no longer silently discards the popover.** `confirmLink` surfaces an inline error and preserves the user's input instead of closing on a falsy href (`Toolbar.tsx:1345-1360`).
- **Security whitelist preserved.** `resolveLinkHref` routes explicit-scheme and protocol-relative input straight through `sanitizeLinkHref`, so the §3.7 scheme whitelist still rejects `javascript:` / `data:` / `ftp:` (confirmed by test and by direct probe). The scheme-less branch only prepends `https://` for input that looks like a host, and `abc` → inline error rather than `https://abc/`.

## 2. Findings (all P2 / advisory — none block merge)

**P2 — focus target for editing an existing link.** `linkHadSelectionRef` is set from `from !== to` (`Toolbar.tsx:1286`). When the popover is opened with a collapsed caret *inside* an existing link, `hasSelection` is `false`, so focus goes to the (empty) text field even though the URL field is pre-filled with the existing href. The previous code used unconditional `autoFocus` on the URL field, so this is a mild regression for the "edit an existing link's URL" flow. Consider keying the focus target on `linkValue` being non-empty as well.

**P2 — scheme-less `host:port` is rejected.** Input like `example.com:8080` or `localhost:3000` is matched by the scheme regex `/^[a-z][a-z0-9+.-]*:/i` (a dotted label followed by `:` looks like a scheme), so it is handed to `sanitizeLinkHref` verbatim, fails to parse as a valid scheme, and returns `null` → inline error. Fails closed (safe), but a user pasting a `host:port` without a scheme gets an "invalid URL" error. Minor UX gap.

**P2 — bare email becomes an http link, not `mailto:`.** `user@example.com` passes `looksLikeHost` (its domain is dotted) and resolves to `https://user@example.com/` (userinfo), not `mailto:user@example.com`. Note this is not a regression — main also failed to produce a mailto — but it's an easy win if bare-email → mailto is desired.

**P2 — relative paths / fragments now rejected.** Scheme-less `/docs/api`, `#section`, `?q=1` previously resolved against the origin and were inserted; they now resolve to `null` and error out. This is intentional fail-closed behavior from XIN-1073, but if in-document anchor links (`#section`) are a supported use case, they'd need an explicit allowance.

**P2 — a11y: `aria-invalid` without `aria-describedby`.** The URL input sets `aria-invalid` (`Toolbar.tsx:1470`) but the error `<span>` (`Toolbar.tsx:1503-1507`) has no `id` wired via `aria-describedby`, so a screen reader announces the field as invalid but won't read the validation message. Small addition.

## 3. Notes

- `IconUnlink` is defined but not rendered anywhere (true on `main` as well) — this PR only restyles it, so it's pre-existing dead code, not introduced here. Worth a follow-up cleanup or wiring it up, but out of scope for this PR.
- The stroke-icon swap is clean: `.octo-tb-icon-stroke` already existed for undo/redo, and the test assertion was updated accordingly (`Toolbar.test.tsx:108`).

## Verdict

Spec fully met, security preserved, well-tested. All findings are P2 (UX/a11y polish, no correctness/security/data-loss/build blockers). **Approving.** The P2 items above are worth a quick follow-up but should not hold up merge.

PR #773, opened 2026-07-15
feat(voice): switch voice trigger key by pressing shift 7 times

xiaokenbot:
## Summary

Adds the ability to switch the voice-input long-press trigger key between **Left Shift** and **Right Shift**.

## How it works

Press the **current** trigger key **7 times rapidly** (each press within 500 ms of the previous, no modifier keys held):

- Left Shift ×7 → switches to **Right Shift**
- Right Shift ×7 → switches back to **Left Shift**

A Toast notification confirms the switch (follows the UI locale — zh-CN / en-US).

The choice is persisted to `localStorage` (`octo_voice_trigger_key`) and survives page refresh.

## Changes

| File | Change |
|------|--------|
| `useVoiceTriggerKey.ts` (new) | Hook managing trigger-key state and rapid-press detection |
| `VoiceInputIndicator.tsx` | Integrates hook; replaces hard-coded `"ShiftLeft"` |
| `VoiceInputButton/index.tsx` | Same |
| `zh-CN.json` / `en-US.json` | Two new i18n keys for the Toast messages |
| `useVoiceTriggerKey.test.ts` (new) | 10 unit tests — all pass locally under vitest@4.1.5 |

## Testing

```
pnpm vitest run apps/web/src/__tests__/useVoiceTriggerKey.test.ts
# 10/10 pass
```

### DIFF
--- a/apps/web/src/__tests__/useVoiceTriggerKey.test.ts
+++ b/apps/web/src/__tests__/useVoiceTriggerKey.test.ts
@@ -17,8 +17,15 @@ const localStorageMock = {
 Object.defineProperty(window, "localStorage", { value: localStorageMock, writable: true })
 
 // ── helpers ────────────────────────────────────────────────────────────────
+function makeKeyEvent(code: string, extra: Partial<KeyboardEventInit> = {}) {
+    const e = new KeyboardEvent("keydown", { code, bubbles: true, cancelable: true, ...extra })
+    vi.spyOn(e, "stopImmediatePropagation")
+    vi.spyOn(e, "preventDefault")
+    return e
+}
+
 function fireKeyDown(code: string, extra: Partial<KeyboardEventInit> = {}) {
-    window.dispatchEvent(new KeyboardEvent("keydown", { code, bubbles: true, cancelable: true, ...extra }))
+    window.dispatchEvent(makeKeyEvent(code, extra))
 }
 
 function pressNTimes(code: string, n: number, gapMs = 100) {
@@ -26,6 +33,18 @@ function pressNTimes(code: string, n: number, gapMs = 100) {
         vi.advanceTimersByTime(gapMs)
         fireKeyDown(code)
     }
+}
+
+// Fires n presses and returns the event objects so callers can inspect spies
+function pressNTimesTracked(code: string, n: number, gapMs = 100) {
+    const events: KeyboardEvent[] = []
+    for (let i = 0; i < n; i++) {
+        vi.advanceTimersByTime(gapMs)
+        const e = makeKeyEvent(code)
+        window.dispatchEvent(e)
+        events.push(e)
+    }
+    return events
 }
 
 // ── tests ──────────────────────────────────────────────────────────────────
@@ -40,6 +59,7 @@ describe("useVoiceTriggerKey", () => {
         vi.restoreAllMocks()
     })
 
+    // ── basic state ──────────────────────────────────────────────────────
     it("defaults to ShiftLeft when localStorage is empty", () => {
         const { result } = renderHook(() => useVoiceTriggerKey(vi.fn()))
         expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
@@ -51,6 +71,7 @@ describe("useVoiceTriggerKey", () => {
         expect(result.current.voiceTriggerKey).toBe("ShiftRight")
     })
 
+    // ── toggle ───────────────────────────────────────────────────────────
     it("switches ShiftLeft → ShiftRight after 7 rapid presses", () => {
         const onToggled = vi.fn()
         const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
@@ -84,12 +105,78 @@ describe("useVoiceTriggerKey", () => {
         expect(onToggled).not.toHaveBeenCalled()
     })
 
+    // ── P1 fix: stopImmediatePropagation on completing press ─────────────
+    it("calls stopImmediatePropagation() and preventDefault() on the 7th (completing) press", () => {
+        renderHook(() => useVoiceTriggerKey(vi.fn()))
+
+        let events: KeyboardEvent[]
+        act(() => { events = pressNTimesTracked("ShiftLeft", 7) })
+
+        const completing = events![6]
+        expect(completing.stopImmediatePropagation).toHaveBeenCalled()
+        expect(completing.preventDefault).toHaveBeenCalled()
+    })
+
+    it("does NOT call stopImmediatePropagation() on non-completing presses (1-6)", () => {
+        renderHook(() => useVoiceTriggerKey(vi.fn()))
+
+        let events: KeyboardEvent[]
+        act(() => { events = pressNTimesTracked("ShiftLeft", 6) })
+
+        events!.forEach((e) => {
+            expect(e.stopImmediatePropagation).not.toHaveBeenCalled()
+        })
+    })
+
+    // ── P1 fix: isVoiceEnabled gate ──────────────────────────────────────
+    it("does not register a listener when isVoiceEnabled=false", () => {
+        const onToggled = vi.fn()
+        renderHook(() => useVoiceTriggerKey(onToggled, { isVoiceEnabled: false }))
+
+        act(() => { pressNTimes("ShiftLeft", 7) })
+
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("registers listener when isVoiceEnabled=true (default)", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled, { isVoiceEnabled: true }))
+
+        act(() => { pressNTimes("ShiftLeft", 7) })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftRight")
+        expect(onToggled).toHaveBeenCalledOnce()
+    })
+
+    // ── P1 fix: focus gate (checkIsInputActive) ──────────────────────────
+    it("does not toggle when checkIsInputActive returns false", () => {
+        const onToggled = vi.fn()
+        const checkIsInputActive = vi.fn(() => false)
+        renderHook(() => useVoiceTriggerKey(onToggled, { checkIsInputActive }))
+
+        act(() => { pressNTimes("ShiftLeft", 7) })
+
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("toggles when checkIsInputActive returns true", () => {
+        const onToggled = vi.fn()
+        const checkIsInputActive = vi.fn(() => true)
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled, { checkIsInputActive }))
+
+        act(() => { pressNTimes("ShiftLeft", 7) })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftRight")
+        expect(onToggled).toHaveBeenCalledWith("ShiftRight")
+    })
+
+    // ── counter reset conditions ─────────────────────────────────────────
     it("resets counter when gap between presses exceeds 500ms", () => {
         const onToggled = vi.fn()
         const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
 
         act(() => {
-            pressNTimes("ShiftLeft", 6, 100) // 6 fast presses
+            pressNTimes("ShiftLeft", 6, 100)
             vi.advanceTimersByTime(600)       // long pause — counter resets
             fireKeyDown("ShiftLeft")          // restart from 1
         })
@@ -127,6 +214,21 @@ describe("useVoiceTriggerKey", () => {
         expect(onToggled).not.toHaveBeenCalled()
     })
 
+    it("does not toggle on autorepeat (repeat=true) presses", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => {
+            for (let i = 0; i < 7; i++) {
+                vi.advanceTimersByTime(100)
+                fireKeyDown("ShiftLeft", { repeat: true })
+            }
+        })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
     it("does not respond to the non-active shift key", () => {
         const onToggled = vi.fn()
         // default is ShiftLeft; pressing ShiftRight 7x should not toggle
@@ -137,6 +239,7 @@ describe("useVoiceTriggerKey", () => {
         expect(onToggled).not.toHaveBeenCalled()
     })
 
+    // ── double-toggle ────────────────────────────────────────────────────
     it("toggles twice: ShiftLeft → ShiftRight → ShiftLeft", () => {
         const onToggled = vi.fn()
         const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
--- a/apps/web/src/__tests__/useVoiceTriggerKey.test.ts
+++ b/apps/web/src/__tests__/useVoiceTriggerKey.test.ts
@@ -1,0 +1,154 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest"
+import { renderHook, act } from "@testing-library/react"
+
+vi.mock("lottie-web", () => ({}))
+vi.mock("@douyinfe/semi-ui", () => ({}))
+
+import useVoiceTriggerKey from "@octo/base/src/Components/MessageInput/useVoiceTriggerKey"
+
+// ── localStorage stub ──────────────────────────────────────────────────────
+const localStorageStore: Record<string, string> = {}
+const localStorageMock = {
+    getItem: vi.fn((key: string) => localStorageStore[key] ?? null),
+    setItem: vi.fn((key: string, value: string) => { localStorageStore[key] = value }),
+    removeItem: vi.fn((key: string) => { delete localStorageStore[key] }),
+    clear: vi.fn(() => { Object.keys(localStorageStore).forEach(k => delete localStorageStore[k]) }),
+}
+Object.defineProperty(window, "localStorage", { value: localStorageMock, writable: true })
+
+// ── helpers ────────────────────────────────────────────────────────────────
+function fireKeyDown(code: string, extra: Partial<KeyboardEventInit> = {}) {
+    window.dispatchEvent(new KeyboardEvent("keydown", { code, bubbles: true, cancelable: true, ...extra }))
+}
+
+function pressNTimes(code: string, n: number, gapMs = 100) {
+    for (let i = 0; i < n; i++) {
+        vi.advanceTimersByTime(gapMs)
+        fireKeyDown(code)
+    }
+}
+
+// ── tests ──────────────────────────────────────────────────────────────────
+describe("useVoiceTriggerKey", () => {
+    beforeEach(() => {
+        localStorageMock.clear()
+        vi.useFakeTimers()
+    })
+
+    afterEach(() => {
+        vi.useRealTimers()
+        vi.restoreAllMocks()
+    })
+
+    it("defaults to ShiftLeft when localStorage is empty", () => {
+        const { result } = renderHook(() => useVoiceTriggerKey(vi.fn()))
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+    })
+
+    it("reads initial value from localStorage", () => {
+        localStorageStore["octo_voice_trigger_key"] = "ShiftRight"
+        const { result } = renderHook(() => useVoiceTriggerKey(vi.fn()))
+        expect(result.current.voiceTriggerKey).toBe("ShiftRight")
+    })
+
+    it("switches ShiftLeft → ShiftRight after 7 rapid presses", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => { pressNTimes("ShiftLeft", 7) })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftRight")
+        expect(onToggled).toHaveBeenCalledOnce()
+        expect(onToggled).toHaveBeenCalledWith("ShiftRight")
+        expect(localStorageMock.setItem).toHaveBeenCalledWith("octo_voice_trigger_key", "ShiftRight")
+    })
+
+    it("switches ShiftRight → ShiftLeft after 7 rapid presses", () => {
+        localStorageStore["octo_voice_trigger_key"] = "ShiftRight"
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => { pressNTimes("ShiftRight", 7) })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+        expect(onToggled).toHaveBeenCalledWith("ShiftLeft")
+    })
+
+    it("does not toggle after only 6 presses", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => { pressNTimes("ShiftLeft", 6) })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("resets counter when gap between presses exceeds 500ms", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => {
+            pressNTimes("ShiftLeft", 6, 100) // 6 fast presses
+            vi.advanceTimersByTime(600)       // long pause — counter resets
+            fireKeyDown("ShiftLeft")          // restart from 1
+        })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("pressing a non-trigger key resets the counter", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => {
+            pressNTimes("ShiftLeft", 5, 100)
+            vi.advanceTimersByTime(100)
+            fireKeyDown("KeyA")              // resets counter
+            pressNTimes("ShiftLeft", 6, 100) // 6 more, not enough for toggle
+        })
+
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("does not toggle when modifier keys are held", () => {
+        const onToggled = vi.fn()
+        renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => {
+            for (let i = 0; i < 7; i++) {
+                vi.advanceTimersByTime(100)
+                fireKeyDown("ShiftLeft", { metaKey: true })
+            }
+        })
+
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("does not respond to the non-active shift key", () => {
+        const onToggled = vi.fn()
+        // default is ShiftLeft; pressing ShiftRight 7x should not toggle
+        renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => { pressNTimes("ShiftRight", 7) })
+
+        expect(onToggled).not.toHaveBeenCalled()
+    })
+
+    it("toggles twice: ShiftLeft → ShiftRight → ShiftLeft", () => {
+        const onToggled = vi.fn()
+        const { result } = renderHook(() => useVoiceTriggerKey(onToggled))
+
+        act(() => { pressNTimes("ShiftLeft", 7) })
+        expect(result.current.voiceTriggerKey).toBe("ShiftRight")
+
+        act(() => {
+            vi.advanceTimersByTime(1000) // ensure counter resets between sequences
+            pressNTimes("ShiftRight", 7)
+        })
+        expect(result.current.voiceTriggerKey).toBe("ShiftLeft")
+        expect(onToggled).toHaveBeenCalledTimes(2)
+    })
+})
--- a/packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx
+++ b/packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx
@@ -66,20 +66,6 @@ export default function VoiceInputIndicator({
   const [showFeedbackNotice, setShowFeedbackNotice] = useState(false);
   const { spaceSetting, loaded, voiceConfig } = useSpaceFeedbackSetting();
 
-  // Voice trigger key (ShiftLeft / ShiftRight), toggled by 7 rapid presses
-  const { voiceTriggerKey } = useVoiceTriggerKey(
-    (newKey: VoiceTriggerKey) => {
-      const msgKey =
-        newKey === "ShiftRight"
-          ? "base.voiceInput.triggerKey.switchedToRight"
-          : "base.voiceInput.triggerKey.switchedToLeft";
-      Toast.info(t(msgKey));
-    },
-    { isVoiceEnabled, checkIsInputActive }
-  );
-  const voiceTriggerKeyRef = useRef(voiceTriggerKey);
-  voiceTriggerKeyRef.current = voiceTriggerKey;
-
   // Long-press ShiftLeft state
   const shiftTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const preparingTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
@@ -207,6 +193,21 @@ export default function VoiceInputIndicator({
     }
     setIsPreparing(false);
   };
+
+  // Voice trigger key (ShiftLeft / ShiftRight), toggled by 7 rapid presses.
+  // Must be called AFTER useVoiceInput so isVoiceEnabled is already declared.
+  const { voiceTriggerKey } = useVoiceTriggerKey(
+    (newKey: VoiceTriggerKey) => {
+      const msgKey =
+        newKey === "ShiftRight"
+          ? "base.voiceInput.triggerKey.switchedToRight"
+          : "base.voiceInput.triggerKey.switchedToLeft";
+      Toast.info(t(msgKey));
+    },
+    { isVoiceEnabled, checkIsInputActive }
+  );
+  const voiceTriggerKeyRef = useRef(voiceTriggerKey);
+  voiceTriggerKeyRef.current = voiceTriggerKey;
 
   // Handle transition from preparing/pending -> actual recording or auto-cancel.
   useEffect(() => {
--- a/packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx
+++ b/packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx
@@ -3,6 +3,8 @@ import { createPortal } from "react-dom";
 import { Toast, Dropdown } from "@douyinfe/semi-ui";
 import { Mic } from "lucide-react";
 import useVoiceInput from "./useVoiceInput";
+import useVoiceTriggerKey from "./useVoiceTriggerKey";
+import type { VoiceTriggerKey } from "./useVoiceTriggerKey";
 import "./voiceInput.css";
 import { ChatContextResult } from "../Conversation/chatContext";
 import { VoiceMode } from "../../Service/VoiceService";
@@ -63,6 +65,17 @@ export default function VoiceInputIndicator({
   const [showModeMenu, setShowModeMenu] = useState(false);
   const [showFeedbackNotice, setShowFeedbackNotice] = useState(false);
   const { spaceSetting, loaded, voiceConfig } = useSpaceFeedbackSetting();
+
+  // Voice trigger key (ShiftLeft / ShiftRight), toggled by 7 rapid presses
+  const { voiceTriggerKey } = useVoiceTriggerKey((newKey: VoiceTriggerKey) => {
+    const msgKey =
+      newKey === "ShiftRight"
+        ? "base.voiceInput.triggerKey.switchedToRight"
+        : "base.voiceInput.triggerKey.switchedToLeft";
+    Toast.info(t(msgKey));
+  });
+  const voiceTriggerKeyRef = useRef(voiceTriggerKey);
+  voiceTriggerKeyRef.current = voiceTriggerKey;
 
   // Long-press ShiftLeft state
   const shiftTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
@@ -299,9 +312,9 @@ export default function VoiceInputIndicator({
         return;
       }
 
-      // Long-press ShiftLeft: start 500ms timer
+      // Long-press voiceTriggerKey: start 500ms timer
       if (
-        e.code === "ShiftLeft" &&
+        e.code === voiceTriggerKeyRef.current &&
         !e.repeat &&
         !e.metaKey &&
         !e.ctrlKey &&
@@ -349,7 +362,7 @@ export default function VoiceInputIndicator({
         return;
       }
 
-      if (shiftTimerRef.current !== null && e.code !== "ShiftLeft") {
+      if (shiftTimerRef.current !== null && e.code !== voiceTriggerKeyRef.current) {
         // Modifier chord (Ctrl/Meta/Alt pressed): cancel voice intent
         if (
           e.code.startsWith("Control") ||
@@ -378,15 +391,15 @@ export default function VoiceInputIndicator({
         return;
       }
 
-      // ShiftLeft released while timer is pending: cancel (normal Shift press)
-      if (e.code === "ShiftLeft" && shiftTimerRef.current !== null) {
+      // voiceTriggerKey released while timer is pending: cancel (normal Shift press)
+      if (e.code === voiceTriggerKeyRef.current && shiftTimerRef.current !== null) {
         clearShiftTimer();
         return;
       }
 
-      // ShiftLeft released while waiting for getUserMedia (recording not yet started)
+      // voiceTriggerKey released while waiting for getUserMedia (recording not yet started)
       if (
-        e.code === "ShiftLeft" &&
+        e.code === voiceTriggerKeyRef.current &&
         shiftRecordingRef.current &&
         !isRecordingRef.current
       ) {
@@ -395,9 +408,9 @@ export default function VoiceInputIndicator({
         return;
       }
 
-      // ShiftLeft released after long-press recording started: stop recording
+      // voiceTriggerKey released after long-press recording started: stop recording
       if (
-        e.code === "ShiftLeft" &&
+        e.code === voiceTriggerKeyRef.current &&
         shiftRecordingRef.current &&
         isRecordingRef.current
       ) {
--- a/packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx
+++ b/packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx
@@ -67,13 +67,16 @@ export default function VoiceInputIndicator({
   const { spaceSetting, loaded, voiceConfig } = useSpaceFeedbackSetting();
 
   // Voice trigger key (ShiftLeft / ShiftRight), toggled by 7 rapid presses
-  const { voiceTriggerKey } = useVoiceTriggerKey((newKey: VoiceTriggerKey) => {
-    const msgKey =
-      newKey === "ShiftRight"
-        ? "base.voiceInput.triggerKey.switchedToRight"
-        : "base.voiceInput.triggerKey.switchedToLeft";
-    Toast.info(t(msgKey));
-  });
+  const { voiceTriggerKey } = useVoiceTriggerKey(
+    (newKey: VoiceTriggerKey) => {
+      const msgKey =
+        newKey === "ShiftRight"
+          ? "base.voiceInput.triggerKey.switchedToRight"
+          : "base.voiceInput.triggerKey.switchedToLeft";
+      Toast.info(t(msgKey));
+    },
+    { isVoiceEnabled, checkIsInputActive }
+  );
   const voiceTriggerKeyRef = useRef(voiceTriggerKey);
   voiceTriggerKeyRef.current = voiceTriggerKey;
 
--- a/packages/dmworkbase/src/Components/MessageInput/useVoiceTriggerKey.ts
+++ b/packages/dmworkbase/src/Components/MessageInput/useVoiceTriggerKey.ts
@@ -24,6 +24,21 @@ function writeStoredKey(key: VoiceTriggerKey): void {
     }
 }
 
+export interface UseVoiceTriggerKeyOptions {
+    /**
+     * If provided, the toggle gesture is only active when this returns true.
+     * Should match the caller's own focus-gate (e.g. checkIsInputActive).
+     * When omitted the gesture is active whenever isVoiceEnabled is true.
+     */
+    checkIsInputActive?: () => boolean
+    /**
+     * When false (voice feature disabled for this mount), the toggle listener
+     * is not registered.  Defaults to true so callers that don't use the flag
+     * are unaffected.
+     */
+    isVoiceEnabled?: boolean
+}
+
 export interface UseVoiceTriggerKeyReturn {
     /** 当前激活的语音触发键，"ShiftLeft" 或 "ShiftRight" */
     voiceTriggerKey: VoiceTriggerKey
@@ -32,16 +47,27 @@ export interface UseVoiceTriggerKeyReturn {
 /**
  * 管理语音输入触发键（ShiftLeft ↔ ShiftRight）。
  *
- * 切换手势：连续快速按目标键 7 次（相邻两次 keydown 间隔 < 500ms）。
+ * 切换手势：连续快速按目标键 7 次（相邻两次 keydown 间隔 < 500ms，
+ * 不持有任何 modifier 键）。
  * - 当前为 ShiftLeft  → 连按左 Shift 7 次 → 切换为 ShiftRight
  * - 当前为 ShiftRight → 连按右 Shift 7 次 → 切换为 ShiftLeft
  *
  * 状态持久化到 localStorage（key: octo_voice_trigger_key）。
  * 切换后通过回调通知调用方以便显示 Toast。
+ *
+ * P1 fixes (yujiawei review):
+ * 1. On the completing (7th) keydown the event is stopped via
+ *    stopImmediatePropagation() so the long-press recording handler never
+ *    arms shiftTimerRef — no phantom recording.
+ * 2. Toggle listener respects the caller's focus gate (checkIsInputActive)
+ *    and isVoiceEnabled flag, mirroring the recording handler's scope.
  */
 export default function useVoiceTriggerKey(
-    onToggled: (newKey: VoiceTriggerKey) => void
+    onToggled: (newKey: VoiceTriggerKey) => void,
+    options: UseVoiceTriggerKeyOptions = {}
 ): UseVoiceTriggerKeyReturn {
+    const { checkIsInputActive, isVoiceEnabled = true } = options
+
     const [voiceTriggerKey, setVoiceTriggerKey] = useState<VoiceTriggerKey>(readStoredKey)
 
     const voiceTriggerKeyRef = useRef(voiceTriggerKey)
@@ -50,13 +76,17 @@ export default function useVoiceTriggerKey(
     const onToggledRef = useRef(onToggled)
     onToggledRef.current = onToggled
 
+    const checkIsInputActiveRef = useRef(checkIsInputActive)
+    checkIsInputActiveRef.current = checkIsInputActive
+
     // Rapid-press counter state
     const pressCountRef = useRef(0)
     const lastPressTimeRef = useRef(0)
 
     const handleKeyDown = useCallback((e: KeyboardEvent) => {
         const currentKey = voiceTriggerKeyRef.current
-        // Only track the current trigger key (no modifiers)
+
+        // Only track the current trigger key (no modifiers, no autorepeat)
         if (
             e.code !== currentKey ||
             e.repeat ||
@@ -64,10 +94,17 @@ export default function useVoiceTriggerKey(
             e.ctrlKey ||
             e.altKey
         ) {
-            // Any other key resets the counter
+            // Any non-trigger key resets the toggle counter
             if (e.code !== currentKey) {
                 pressCountRef.current = 0
             }
+            return
+        }
+
+        // Focus gate — mirror the recording handler's scope
+        const checkFn = checkIsInputActiveRef.current
+        if (checkFn && !checkFn()) {
+            pressCountRef.current = 0
             return
         }
 
@@ -85,6 +122,16 @@ export default function useVoiceTriggerKey(
         if (pressCountRef.current >= TOGGLE_COUNT) {
             pressCountRef.current = 0
             lastPressTimeRef.current = 0
+
+            // ── P1 fix ──────────────────────────────────────────────────────
+            // Stop the completing keydown from reaching the long-press handler.
+            // Without this, shiftTimerRef is armed on the 7th press and then
+            // the matching keyup is ignored (the key ref has already flipped),
+            // causing a phantom recording ~500 ms later.
+            e.stopImmediatePropagation()
+            e.preventDefault()
+            // ────────────────────────────────────────────────────────────────
+
             const newKey: VoiceTriggerKey =
                 currentKey === "ShiftLeft" ? "ShiftRight" : "ShiftLeft"
             writeStoredKey(newKey)
@@ -94,9 +141,12 @@ export default function useVoiceTriggerKey(
     }, [])
 
     useEffect(() => {
-        window.addEventListener("keydown", handleKeyDown)
-        return () => window.removeEventListener("keydown", handleKeyDown)
-    }, [handleKeyDown])
+        if (!isVoiceEnabled) return
+        // Register with capture=true so we run before the recording handler
+        // and stopImmediatePropagation() on the completing press is effective.
+        window.addEventListener("keydown", handleKeyDown, true)
+        return () => window.removeEventListener("keydown", handleKeyDown, true)
+    }, [isVoiceEnabled, handleKeyDown])
 
     return { voiceTriggerKey }
 }
--- a/packages/dmworkbase/src/Components/MessageInput/useVoiceTriggerKey.ts
+++ b/packages/dmworkbase/src/Components/MessageInput/useVoiceTriggerKey.ts
@@ -1,0 +1,102 @@
+import { useState, useEffect, useRef, useCallback } from "react"
+
+export type VoiceTriggerKey = "ShiftLeft" | "ShiftRight"
+
+const STORAGE_KEY = "octo_voice_trigger_key"
+const TOGGLE_COUNT = 7
+const TOGGLE_INTERVAL_MS = 500
+
+function readStoredKey(): VoiceTriggerKey {
+    try {
+        const stored = localStorage.getItem(STORAGE_KEY)
+        if (stored === "ShiftLeft" || stored === "ShiftRight") return stored
+    } catch {
+        // ignore
+    }
+    return "ShiftLeft"
+}
+
+function writeStoredKey(key: VoiceTriggerKey): void {
+    try {
+        localStorage.setItem(STORAGE_KEY, key)
+    } catch {
+        // ignore
+    }
+}
+
+export interface UseVoiceTriggerKeyReturn {
+    /** 当前激活的语音触发键，"ShiftLeft" 或 "ShiftRight" */
+    voiceTriggerKey: VoiceTriggerKey
+}
+
+/**
+ * 管理语音输入触发键（ShiftLeft ↔ ShiftRight）。
+ *
+ * 切换手势：连续快速按目标键 7 次（相邻两次 keydown 间隔 < 500ms）。
+ * - 当前为 ShiftLeft  → 连按左 Shift 7 次 → 切换为 ShiftRight
+ * - 当前为 ShiftRight → 连按右 Shift 7 次 → 切换为 ShiftLeft
+ *
+ * 状态持久化到 localStorage（key: octo_voice_trigger_key）。
+ * 切换后通过回调通知调用方以便显示 Toast。
+ */
+export default function useVoiceTriggerKey(
+    onToggled: (newKey: VoiceTriggerKey) => void
+): UseVoiceTriggerKeyReturn {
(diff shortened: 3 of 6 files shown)


Jerry-Xin (dismissed):
**APPROVE** — Clean, well-scoped feature. Listener lifecycle, counter bounds, stored-value validation, and i18n symmetry all check out. One non-blocking correctness gap around multi-instance mounting is worth a follow-up but does not bar merge.

### What it does
- `packages/dmworkbase/src/Components/MessageInput/useVoiceTriggerKey.ts` (new): hook that reads/persists the trigger key from `localStorage["octo_voice_trigger_key"]`, and installs a global `keydown` listener that toggles `ShiftLeft ↔ ShiftRight` after 7 rapid presses (≤500ms between presses).
- `VoiceInputIndicator.tsx` / `VoiceInputButton/index.tsx`: integrate the hook and replace the hard-coded `"ShiftLeft"` comparisons in the keydown/keyup handlers with the dynamic `voiceTriggerKeyRef.current`; fire a Toast on toggle.
- `i18n/locales/en-US.json` + `zh-CN.json`: two new toast keys each.
- `apps/web/src/__tests__/useVoiceTriggerKey.test.ts` (new): 10 unit tests.

### ✅ Verified
- **Listener lifecycle** — `useVoiceTriggerKey.ts` registers `keydown` in a `useEffect` and returns a cleanup that removes it; the handler is a `useCallback` with an empty dep array (stable), so within a single mount the listener is added once and removed on unmount. No per-instance leak, no re-subscribe on every render.
- **Counter** — bounded and correct: exact `>= 7` threshold, a gap over 500ms restarts the sequence, any non-trigger key resets the count, `e.repeat` and held modifiers are excluded, and the non-active Shift is ignored. Tests cover the 6-vs-7 off-by-one, the timeout reset, and the non-trigger-key reset.
- **Stored value validation** — `readStoredKey()` accepts only the exact strings `ShiftLeft`/`ShiftRight` and otherwise falls back to `ShiftLeft`; localStorage access is wrapped in try/catch. A malformed/hostile stored value cannot bind anything unexpected — the value is only ever compared against `e.code`, never executed or interpolated.
- **Normal typing / Shift usage** — only bare trigger-key presses are counted and the handler never calls `preventDefault`, so capitalization and other Shift shortcuts are unaffected.
- **i18n** — both new keys added symmetrically to en-US and zh-CN under the correct `base.` namespace (matches existing `base.voiceInput.*` usage).

### 💬 Non-blocking (MAJOR follow-up): multi-instance state is not synced
`VoiceInputButton` is mounted in many places (SpaceSettings, BotDetailModal, GroupMdEditor, ThreadCreateModal, PersonaSettings/PersonaEdit, SpaceCreate, …) in addition to `VoiceInputIndicator` in the main message input. Each mounted hook instance installs its own global listener and holds its own private counter + `voiceTriggerKey` state. When two or more voice components are mounted at once (e.g. the main input plus an open settings modal):
- the toggle gesture fires the `onToggled` callback once per instance → duplicate Toasts, and `writeStoredKey` runs once per instance;
- more importantly there is no cross-instance sync — the hook does not listen for the `storage` event, so after a toggle the sibling instances keep the stale `voiceTriggerKey` in state until they remount, meaning a Shift press won't behave as the new trigger key in those siblings.

Most of these mounts are modal-gated (usually one visible at a time), so the dominant single-instance path (main chat input) is correct and this is edge-limited — hence non-blocking. A tidy fix would be to add a `storage`-event listener in the hook so all instances converge on the persisted value.

### 🟢 Nits
- In `useVoiceTriggerKey.ts` the early-return guard resets the counter only when `e.code !== currentKey`; pressing the trigger key *with* a modifier held returns without resetting. Harmless (such a press just doesn't advance the sequence), but slightly asymmetric with the "any other key resets" intent.
- The new unit tests live under `apps/web/src/__tests__/` but no CI job runs vitest, so the 10/10 pass is local-only — not enforced on this PR.

CI: code checks green (add-to-project, label, pr-title-lint). Governance checks are skipping/pending (check-sprint, satisfy-code-review, code-review) — none bar approval. Non-draft, mergeable.

OctoBoooot (commented):
## Review: switch voice trigger key by pressing shift 7 times (#773)

**Verdict: Comment** — the code is clean on review (no blockers, no majors), but the **code CI checks (Build, secret-scan, osv-scan, dependency-review) have not registered yet** at this head (only label / pr-title-lint / add-to-project have run; `ci_state` is pending). I don't approve over unconfirmed code CI — this is a COMMENT that the review side is clean and will be an easy APPROVE once Build lands green. One non-blocking UX 🟡. Reviewed at `252d6e71d059`.

### The hook (`useVoiceTriggerKey.ts`) — well-built (verified)
- **Listener lifecycle clean** — `useEffect` adds `window.addEventListener("keydown")` and returns the matching `removeEventListener`; `handleKeyDown` is `useCallback([])` (stable), reading live values via refs (`voiceTriggerKeyRef` / `onToggledRef` / `pressCountRef` / `lastPressTimeRef`) — no stale-closure, no re-subscribe churn, no leak.
- **7-press state machine correct** — `gap <= 500ms → increment; else restart at 1`; at `>= 7` toggles Left↔Right, resets counter+time. A *different* key resets the counter (breaks a stray sequence); `e.repeat` (held-key autorepeat) and modifier-held presses are ignored (don't count, don't falsely reset a legit sequence).
- **Modifier guard** — ignores when `meta/ctrl/alt` held, so Ctrl+Shift etc. won't accumulate. Good.
- **Persistence robust** — `localStorage` read/write both try/catch-wrapped, validates the stored value (`"ShiftLeft"|"ShiftRight"`), defaults to `ShiftLeft`.

### Consumer wiring — correct (both components)
`VoiceInputIndicator.tsx` and `VoiceInputButton/index.tsx` both replace the hard-coded `"ShiftLeft"` with `voiceTriggerKeyRef.current` (mirrored via a ref so their own keydown handlers read the live trigger key), and wire the hook's `onToggled` to a `Toast.info(t(msgKey))`. The long-press-timer logic is preserved, just keyed off the dynamic trigger key. No hard-coded `ShiftLeft` left in the changed handlers.

### i18n — keyed both locales
`voiceInput.triggerKey.switchedToRight` / `switchedToLeft` added symmetrically in en-US + zh-CN ("Voice trigger key switched to Right Shift" / "语音输入键已切换为右 Shift"). Toast follows the UI locale.

### Minor (non-blocking) — accidental toggle during heavy Shift usage
7 rapid Shift presses (each < 500 ms) is a reachable pattern during fast capitalization, so a user could accidentally toggle the trigger key. Low blast radius (only swaps which Shift triggers voice, shows a Toast, fully reversible by the same gesture) — a deliberate-gesture-design tradeoff, not a defect. Consider a higher count or a longer no-other-key requirement if accidental toggles get reported. Non-blocking.

### Net
Clean feature: the hook's listener lifecycle, rapid-press state machine, modifier guard, and localStorage persistence are all correct; both consumers wire through the dynamic key; i18n symmetric; a 10-test suite is included. No blockers, no majors. Holding at **COMMENT** only because the code CI (Build/secret-scan/osv-scan/dependency-review) hasn't registered yet at this head — once Build is green this is a straightforward APPROVE. One non-blocking 🟡 (accidental-toggle during heavy Shift use). Credit @xiaokenbot — the ref-mirrored stable-callback listener and the modifier/`e.repeat` guards are exactly right for a global keydown hook.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 blockers. One P2 (duplicate Toast on multi-instance mount) is non-blocking UX; two nits are cosmetic. Core toggle logic, type safety, localStorage persistence, and consumer wiring all verified correct.

yujiawei (changes requested):
# Code Review — PR #773 (octo-web)

Reviewed at head SHA `252d6e71d0598b5f8037ccf1c7efdd3e38fa1429` against `main`. Scope: new `useVoiceTriggerKey` hook, its wiring into `VoiceInputIndicator` and `VoiceInputButton`, two i18n keys, and the unit test file.

**Verdict: Changes Requested** — the toggle gesture interacts badly with the existing long-press recording handler and fires a spurious recording on every successful switch. Details below.

---

## Spec compliance ✅

The diff implements exactly what the PR describes, no more, no less:

- 7 rapid presses (<500ms gaps) of the **current** trigger key toggles Left⇄Right Shift (`useVoiceTriggerKey.ts:73-92`).
- Choice persisted to `localStorage["octo_voice_trigger_key"]` and read back at mount (`readStoredKey`/`writeStoredKey`).
- Locale-aware Toast on switch; both i18n keys added under the correctly-registered `base` namespace.
- Both documented components wired; hard-coded `"ShiftLeft"` replaced by `voiceTriggerKeyRef.current`.

No under-build, over-build, or behavioral deviation from the stated design.

---

## Code quality — blocking issues

### 🔴 P1 — Every successful toggle leaves a dangling long-press timer → phantom recording

`packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx:315-363` (and identically `VoiceInputButton/index.tsx:200-243`)

The new toggle hook and the pre-existing recording handler **both** listen on `window` `keydown`. During the 7-press gesture every keydown is the current trigger key, so it also enters the recording handler's long-press branch and arms a 500 ms `shiftTimerRef` timer. For presses 1–6 the matching keyup clears it. The 7th press is different:

1. 7th **keydown**: `shiftTimerRef` is `null` (6th keyup cleared it) → a 500 ms timer is armed, captured while `voiceTriggerKeyRef.current` is still the *old* key (e.g. `ShiftLeft`).
2. The hook's listener, on the same event, calls `setVoiceTriggerKey(newKey)`. React flushes the re-render, and the component's `voiceTriggerKeyRef.current` flips to the *new* key (`VoiceInputIndicator.tsx:77-78`) **before** the separate keyup event is dispatched.
3. 7th **keyup**: every clear/stop branch is guarded by `e.code === voiceTriggerKeyRef.current`. Now `"ShiftLeft" !== "ShiftRight"`, so `clearShiftTimer()` is never called (`VoiceInputIndicator.tsx:395`; `VoiceInputButton/index.tsx:263`).
4. ~500 ms after the key is physically released, the orphaned timer fires `startRecordingRef.current("append_only")` — a recording starts with no key held.

This happens on the feature's happy path, on **every** successful toggle. Recovery is non-obvious: the stop branches now match only the *new* trigger key, so the user must press/release the new key once (or hit Escape) to stop a recording they never intended to start. `VoiceInputButton` is worse — its timer callback (`index.tsx:230-231`) does not re-check `isRecordingRef` before starting, and its keyup has no `e.key === "Shift"` fallback, so an old-key release cannot stop it.

Suggested direction: have the toggle detection call `e.preventDefault()`/`e.stopImmediatePropagation()` on the completing gesture, or have the recording handler skip arming while a rapid-press sequence is in progress, or clear `shiftTimerRef`/`preparingTimerRef` when a toggle fires. A regression test that dispatches the full keydown+keyup gesture against a mounted component (not just the isolated hook) should accompany the fix.

### 🔴 P1 — Toggle listener is global: not focus-gated, not gated by `isVoiceEnabled`

`packages/dmworkbase/src/Components/MessageInput/useVoiceTriggerKey.ts:57-99`

`handleKeyDown` checks only `e.code`, `e.repeat`, and modifier flags — there is no `document.activeElement`/input-focus guard, and the hook is called before the components' `if (!isVoiceEnabled) return null` early return (`VoiceInputIndicator.tsx:70` before `:474`; `VoiceInputButton/index.tsx:56` before `:306`). This is asymmetric with the recording handlers that live beside it, which are carefully focus-scoped (`VoiceInputButton/index.tsx:198`, `VoiceInputIndicator.tsx:275`).

Consequences: as long as a host component is mounted, tapping the active Shift key 7× within 500 ms anywhere on the page — focus in a search box, another textarea, a dropdown, or nowhere — silently flips the trigger key and toasts, even for spaces where voice input is disabled. The `e.repeat` guard and the "any other key resets the counter" rule make accidental triggering less likely (a held Shift or interleaved keystroke won't count), which keeps this out of P0, but a global config mutation for a feature whose action is focus-scoped is surprising and should be gated consistently (focus + `isVoiceEnabled`).

---

## Code quality — non-blocking (P2 / suggestions)

- **Duplicate toasts / redundant writes when multiple instances are mounted.** Each hook instance keeps its own counter and window listener with no shared/module-level guard. Multiple `VoiceInputButton`s coexist on real screens (e.g. `CreateTaskModal` mounts two, `MatterDetailPanel` mounts three; modals also overlay the chat's `VoiceInputIndicator`). One 7-press gesture then fires N toasts and N idempotent `localStorage` writes. Final value stays consistent, so it's cosmetic — but a module-level singleton listener/counter would fix both this and half of the P1s. (`useVoiceTriggerKey.ts:96-99`)
- **`isPreparing` overlay flicker mid-gesture.** Because each gesture press flows through the long-press path, a press held ≥300 ms (`PREPARING_DELAY_MS`) flashes the "preparing" overlay; the 7th press's preparing timer also survives via the same ref-flip as the P1. Mostly a facet of the P1 fix. (`VoiceInputIndicator.tsx:329-332`)
- **Test coverage gaps.** The suite is solid for the isolated hook (10/10 pass, and `vi.advanceTimersByTime` does advance the faked clock here, so the gap logic is genuinely exercised), but it never dispatches `keyup` and never mounts the wiring components, so none of the above interactions are caught. Missing cases worth adding: held-key autorepeat (`{repeat:true}` should not toggle — the most safety-critical branch, currently unasserted), the exact `gap === 500` boundary (`<=` edge), and a multi-instance / toggle-then-record integration test.

---

## Coverage note

Verified: hook logic and both integration sites read end-to-end; i18n namespace registration; `@octo/base` test alias resolution; vitest fake-timer/`Date` interaction; real coexistence of the two hosts (`GroupMdEditor`, `CreateTaskModal`, `MatterDetailPanel` confirmed via grep — note `ThreadCreateModal` appears to be unreferenced/dead, so it is not a live coexistence path). Not verified at runtime: actual browser behavior of the phantom recording (traced statically through the React 18 native-event flush ordering, which is deterministic here) and getUserMedia paths.

OctoBoooot:
@/tmp/pr773-triage-yujiawei-252d6e71.md

Jerry-Xin (changes requested):
Flipping my prior APPROVE at this head to REQUEST_CHANGES. On re-review I audited the switch hook + persistence + i18n in isolation and missed the cross-listener interaction with the existing long-press recorder — peer review correctly surfaced it. Independently byte-verified below.

## 🔴 Blocking

**The 7-tap switch gesture orphans the long-press record timer → spurious live-microphone recording on the happy path.**

Both the switch hook and the pre-existing long-press recorder register independent `window` keydown/keyup listeners. During the toggle gesture (7 taps of the *current* trigger key while focused in the voice input field — the intended place to do it), the recorder's long-press branch also arms. Trace of the 7th press:
- 7th keydown: `VoiceInputIndicator.tsx:315-320` arms `shiftTimerRef = setTimeout(startRecording, 500ms)` (the 6th keyup cleared the prior timer, so it arms fresh).
- Same keydown: `useVoiceTriggerKey.ts:91` calls `setVoiceTriggerKey(newKey)`. This is a native window listener → React flushes the state update synchronously → `voiceTriggerKeyRef` flips to the NEW key **before** the browser dispatches keyup.
- 7th keyup: the cancel branch is `if (e.code === voiceTriggerKeyRef.current && shiftTimerRef.current !== null) clearShiftTimer()` (`VoiceInputIndicator.tsx:394`). But now `e.code` is the OLD physical key just released, while the ref is the NEW key → the equality fails → `clearShiftTimer()` never runs → the 500ms timer survives.
- ~500ms later the orphaned timer fires `startRecording("append_only")` → recording begins with no key held, as a side effect of a successful key-switch.

`VoiceInputButton` is worse: its keyup handler (`VoiceInputButton/index.tsx:257+`) has only `e.code === voiceTriggerKeyRef.current` branches — no `e.key === "Shift"` fallback and no Escape handler — so releasing the old key cannot cancel the phantom recording; recovery requires a deliberate press+release of the NEW key.

Fix: on toggle completion, clear the pending `shiftTimerRef` (and preparing timer), and/or `stopImmediatePropagation()` on the gesture-completing event so the recorder listener doesn't also process it; and add an `e.key === "Shift"` fallback (or key-agnostic release) so a release of the just-swapped key still cancels.

## 🟡 Should-fix alongside (non-blocking major)

**The switch hook's window listener has no focus / enabled-state gate.** `useVoiceTriggerKey.ts:57-99` `handleKeyDown` checks only `e.code`, `e.repeat`, and modifier flags — no `document.activeElement`/input-focus check and no `isVoiceEnabled` gate (the hook runs at `VoiceInputIndicator.tsx:70` / `VoiceInputButton/index.tsx:56`, both before the `if (!isVoiceEnabled) return null` early return). So it's armed even in voice-disabled spaces and fires regardless of focus, asymmetric with the neighboring focus-scoped recorder listeners. Blast radius is small and self-mitigating (only swaps ShiftLeft↔ShiftRight + a reversible Toast; `e.repeat` guard; the "any non-trigger key resets the count" rule means it needs 7 bare Shift taps under 500ms apart with no intervening key — not a natural typing pattern), so this is a major, not independently blocking. Same root fix area as the blocker: gate + coordinate the hook listener.

## ✅ Otherwise
The parts I audited before remain sound: listener add/remove cleanup, the counter bounds/reset/off-by-one (with tests), and the persisted key strictly validated to ShiftLeft/ShiftRight. The gap was purely the cross-listener interaction with the recorder, which the two items above address.

Note: this PR is not draft but its added unit tests exercise the hook in isolation and did not cover this cross-listener path — please add a regression test for "toggle gesture must not start recording" on both wired components.

Jerry-Xin (approved):
Re-verified at the current head 720fe6d2 (a new commit landed after my CR was queued). The two issues from my prior REQUEST_CHANGES are now FIXED — flipping to APPROVE. Byte-verified below.

## ✅ Prior blocker + major — RESOLVED at 720fe6d2

**🔴 (was blocking) — phantom recording from the orphaned long-press timer: FIXED.**
`useVoiceTriggerKey.ts:131-132` now calls `e.stopImmediatePropagation()` + `e.preventDefault()` on the completing 7th press, and the hook's keydown listener is registered with **capture = true** (`useVoiceTriggerKey.ts:145`) so it runs *before* the long-press recording handler. The completing press therefore never reaches the recorder branch, so `shiftTimerRef` is never armed on the toggle → no phantom recording ~500 ms later. The fix is explicitly commented as the P1 fix and the mechanism is exactly right (capture-phase + stopImmediatePropagation is what makes it effective). VoiceInputButton additionally gained an `Escape` handler and a key-agnostic timer-clear branch as extra recovery safety.

**🟡 (major, alongside) — missing focus/enabled gate on the global listener: FIXED.**
The hook now accepts `{ isVoiceEnabled, checkIsInputActive }` options; its effect early-returns when `!isVoiceEnabled` (`useVoiceTriggerKey.ts:144`) and `VoiceInputIndicator.tsx` passes `checkIsInputActive` (focus gate) into it, mirroring the recorder handler's scope. `VoiceInputButton` gates via `isVoiceEnabled`. The listener is no longer armed in voice-disabled spaces / regardless of focus.

## ✅ Full-feature check (still holds)
Listener add/remove cleanup, the 7-press counter bounds/reset/interval (`TOGGLE_INTERVAL_MS`) with off-by-one test coverage, and the persisted key strictly validated to ShiftLeft/ShiftRight all remain sound. The new commit also adds regression tests exercising the toggle path.

## Notes (non-blocking)
- Please ensure the added tests cover the specific "toggle gesture must not start recording" regression on both wired components (the exact cross-listener path that was the bug), not only the hook in isolation.

Nice fix — the capture-phase + stopImmediatePropagation approach is the correct way to prevent the sibling long-press handler from arming. Credit to the peer review that surfaced the cross-listener interaction.

OctoBoooot (commented):
## Re-review (DELTA) @ `720fe6d22132` — both of yujiawei's 🔴 fixed; and a correction: her #1 (phantom recording) was REAL — I wrongly refuted it

Clean single-commit fast-forward ("prevent phantom recording on trigger-key toggle; add focus/voice gate"), touching the hook + both consumers + tests. Both of yujiawei's findings are addressed correctly. Verdict stays **COMMENT** (review now cleaner, still held pending code CI — Build/secret-scan/osv haven't registered at this head). But I owe a correction on my earlier triage.

### Correction — yujiawei's 🔴#1 (phantom recording) was REAL; my refutation was wrong
My prior triage refuted #1 reasoning "keyup clears the record timer per tap." I **missed the completing-press mechanism**, which the fix comment spells out: on the **7th press**, the toggle flips `voiceTriggerKeyRef` to the *new* key. The record handler arms `shiftTimerRef` on that 7th keydown, but the matching **keyup's clear is `e.code === voiceTriggerKeyRef.current` — which no longer matches** (the ref already flipped), so `clearShiftTimer()` doesn't fire for that press → the timer survives → phantom recording ~500 ms later. So the toggle *could* start an unintended recording, exactly as yujiawei said. I under-verified (I checked the general per-tap keyup clear but not the ref-flip on the completing press). Credit yujiawei — #1 was a real bug, not refuted.

### Fix #1 — phantom recording — correct
On the completing (7th) press: `e.stopImmediatePropagation(); e.preventDefault()`, and the toggle listener is registered with **`capture=true`** (`useEffect`) so it runs *before* the record handler's bubble-phase listener (`VoiceInputButton` :152 / VoiceInputIndicator, default bubble). So the completing press is stopped before it can arm `shiftTimerRef` — no phantom recording. Test asserts `stopImmediatePropagation()` fires on the 7th press and NOT on presses 1–6. Sound.

### Fix #2 — focus/enabled gate (yujiawei #2 / my 🟡) — correct
The hook now takes `{ checkIsInputActive, isVoiceEnabled }`: the toggle listener isn't registered when `isVoiceEnabled=false`, and the keydown bails (resetting the counter) when `checkIsInputActive()` is false — mirroring the record handler's `activeElement === inputRef.current` scope. Consumers wire it: `VoiceInputButton` passes `{ isVoiceEnabled }`, `VoiceInputIndicator` passes `{ isVoiceEnabled, checkIsInputActive }` (arg order verified: `onToggled` first, options second). So the toggle no longer fires app-wide unscoped. Fixes the unscoped-listener issue.

### No-regression + API check
- Consumer call signature verified correct (`onToggled` callback first, options object second — not swapped).
- The original clean parts hold: hex-N/A here; listener still has cleanup, counter still bounded (500ms/reset/off-by-one), localStorage validated, no injection.
- Tests extended for both fixes (stopImmediatePropagation on/off, isVoiceEnabled gate).

### Net
Both yujiawei 🔴s fixed correctly (phantom-recording via capture-phase stopImmediatePropagation on the completing press; focus/enabled gate mirroring the record handler), tests added. **I retract my earlier refutation of #1 — it was a real bug I under-verified (missed the completing-press ref-flip orphaning the keyup clear).** Verdict remains **COMMENT** only because the code CI (Build/secret-scan/osv/dependency-review) still hasn't registered at this head — so the added tests haven't run in CI. Once code CI lands green this is an easy APPROVE. Credit yujiawei for both catches — #1 especially, which I wrongly refuted.

yujiawei (changes requested):
# Code Review — PR #774 (octo-web)

Reviewed at head SHA `f66c42b6054e9043c293c674332f423a4717f2b4` against `origin/main` (merge-base `578dac69`). This PR was flagged security-sensitive, so the color-value data flow was audited end-to-end.

Verified locally: `vitest` on the two changed test files → **89/89 pass**; `tsc` typecheck → no new errors in the touched files (the two pre-existing errors in `dmworkbase/src/i18n/format.ts` and `members/sort.ts` are untouched by this PR); CI stylelint on `editor/styles.css` → the new hex block is clean and the 309 baseline `color-no-hex` errors are identical on `main` (pre-existing, not introduced here).

## Verdict: APPROVE

No blocking (P0/P1) issue found. The core feature — an OS-independent inline `#rrggbb` entry for the **font-colour** picker (the open gap in #719) — is implemented correctly: input is validated by `normalizeHexColor`, only the normalized hex reaches the document (`setColor`), invalid input sets `aria-invalid` and never touches the content model, and the value round-trips through the same `textStyle` color mark as a preset swatch. The items below are non-blocking suggestions.

### Spec compliance — PASS
- **Missing:** none. The font-colour hex path is fully wired (`TextColorControl` → `HexColorInput` → `setColor`).
- **Over-build:** the same `HexColorInput` is also added to the **highlight** popover (`Toolbar.tsx:470-475`), which #719 does not scope. This is defensible shared-component symmetry (the two controls are intentional mirrors), so it is not a blocker — but see the correctness note below, because the highlight path chose different commit semantics.
- **Deviation:** none — implementation matches the stated commit-once-on-Enter approach, no new deps, no new persistent state.

### Security — no vulnerability found
`normalizeHexColor` is the sole validator for the new field and was probed against `javascript:`, `expression(...)`, `</style><script>`, CSS rule-break payloads, full-width/unicode digits, and double-`#` — all correctly return `null`. Both regexes are anchored fixed-length ASCII classes (`{3}`/`{6}`), so no ReDoS and no unicode-digit bypass. The committed value is always the normalized hex, never the raw input; paste lands only in React state and still routes through `normalizeHexColor` on Enter. (Note for completeness, not a defect in this PR: the pre-existing native `<input type="color">` path at `Toolbar.tsx:408/498` passes `input.value` raw into `setHighlight`/`setColor`, but that value is spec-guaranteed to be a normalized `#rrggbb` and is unchanged by this PR.)

## Suggestions (non-blocking)

**1. (P2) Highlight hex input uses `toggleHighlight`, so re-typing the *current* highlight colour silently removes it.** `Toolbar.tsx:470-475` commits via `editor.chain().focus().toggleHighlight({ color: hex }).run()`, whereas the native wheel deliberately uses `setHighlight` (line 408) for always-set, commit-once semantics. `toggleHighlight` is attribute-sensitive: if the selection already carries that exact highlight colour, pressing Enter toggles it **off**, and because `setOpen(false)` also runs, the popover collapses exactly as if it had applied — so "type the brand hex I already have, press Enter" clears the highlight with no signal. The font-colour path is unaffected because it uses the idempotent `setColor`. Preset swatches also use `toggleHighlight` (line 435), so toggle-on-repeat-click is arguably intended for *clicking*, but for a typed-hex "set this colour" intent it is surprising. Recommend `setHighlight({ color: hex })` for the hex input to match the wheel's set-once behaviour.

**2. (P2) `maxLength={7}` truncates a leading-space paste before `normalizeHexColor` can trim it.** `Toolbar.tsx:371`. A full 6-digit hex is exactly 7 chars (`#rrggbb`), so pasting `" #1971c2"` truncates to `" #1971c"` and is falsely flagged invalid. Trailing whitespace and typed input are fine; only a leading-space paste hits this, and it is recoverable by editing — hence P2, not blocking. (One advisory leg rated this P1; on independent check the impact is narrower than that — no data loss, recoverable — so P2.) Consider relaxing/removing `maxLength` since `normalizeHexColor` already bounds the value, or trimming in `onChange`.

**3. (P2) Invalid state is conveyed by border colour only.** `HexColorInput` sets `aria-invalid` + a red border but renders no error text, unlike the analogous link field which pairs `.octo-link-field.is-invalid` with a visible `.octo-link-error` (`role="alert"`) message. Screen-reader users hear "invalid" but not why; sighted colourblind users get only the border. Mirroring the link field's `role="alert"` text (with a new i18n key) would make the two paths consistent — WCAG 1.4.1.

**4. (P2) Escape / event-bubbling on Enter.** `onKeyDown` (`Toolbar.tsx:377`) handles Enter with `preventDefault()` but no `stopPropagation()`, and does not handle Escape. If the parent popover doesn't already trap these, Enter could bubble to surrounding handlers and there's no keyboard-only dismiss from the field. Worth confirming the popover's own key handling covers this.

**5. (P2) Test coverage gaps** (the feature works; these guard against future regressions):
- The **highlight** hex path (`toggleHighlight`) has no test — only the text-colour path (`.octo-text-color-popover .octo-color-hex`) is exercised (`Toolbar.test.tsx:114-152`). Given suggestion #1, a test asserting the highlight commit would be especially valuable.
- Empty/whitespace-only Enter no-op (`Toolbar.tsx:379-380`) is untested.
- The invalid→retype→clear recovery loop (`setInvalid(false)` at `Toolbar.tsx:372-375`) is untested; the sibling link popover already tests this pattern.
- The invalid-input test asserts "no mark created from scratch" but not "an existing colour survives invalid input" — consider seeding a colour first.

**Nit:** the CSS comment at `styles.css:387-390` says the red `#eb5757` matches `.octo-link-field.is-invalid`, but that pre-existing selector is *not* wrapped in `stylelint-disable` — the new block is actually stricter than what it references. Comment-accuracy only; the CSS is correct.

## Coverage / blind spots
- Focus/blur selection-restoration after typing (the `chain().focus()` re-focus idiom) is validated by analysis + the identical native-picker precedent, **not** by an integration test — jsdom does not model editor DOM focus loss, so no leg exercised the real-browser blur→refocus path.
- The DOCX/Markdown exporter round-trip of a *typed* hex is claimed in comments but has no added/exercised test; the exporter (`export/docx/marks.ts`) does re-normalize hex, so the claim is plausible.
- IME/composition input and mobile on-screen-keyboard Enter were not verified.
- Popover-row reflow once the 74px field is appended alongside all swatches was not visually checked.

This review fused two independent advisory perspectives plus this reviewer's own multi-lens pass; one external leg was unavailable this run (infrastructure error, not a code signal) and was treated as absent, not as agreement — findings here reflect only what was actually verified.

OctoBoooot:
@/tmp/pr773-misposted-note.md

mochashanyao (changes requested):
[Octo-Q · automated review]

---

[Octo-Q] verdict: REQUEST_CHANGES — Two P1 TDZ violations crash both consumer components (VoiceInputIndicator and VoiceInputButton) on mount. `isVoiceEnabled` is used before declaration in both files. Fix is mechanical: move the `useVoiceTriggerKey()` call after the voice-service hook destructuring.

Jerry-Xin (changes requested):
Flipping my APPROVE at this head to REQUEST_CHANGES. Peer review (mochashanyao) surfaced a temporal-dead-zone crash that the fix commit introduced and that I missed — byte-confirmed below. This is a hard crash-on-mount, so it outranks the (correct) phantom-recording/gate fixes this commit also landed.

## 🔴 Blocking — TDZ crash on mount in BOTH consumers

The commit added `{ isVoiceEnabled, ... }` to the `useVoiceTriggerKey(...)` call, but placed that call ABOVE the declaration of `isVoiceEnabled`. `isVoiceEnabled` is a `const` from a hook destructure (not a prop), so referencing it before its declaration throws `ReferenceError: Cannot access 'isVoiceEnabled' before initialization` (temporal dead zone) — the component throws on first render.

- `VoiceInputIndicator.tsx`: `useVoiceTriggerKey(cb, { isVoiceEnabled, checkIsInputActive })` at ~line 70-78, but `isVoiceEnabled` is destructured from `useVoiceInput({...})` at ~line 106. Used ~28 lines before declaration. (`checkIsInputActive` is a prop param, so that one is fine — only `isVoiceEnabled` is in the dead zone.)
- `VoiceInputButton/index.tsx`: `useVoiceTriggerKey(cb, { isVoiceEnabled })` at ~line 56-64, but `isVoiceEnabled` is destructured from `useTextareaVoice({...})` at ~line 75. Same TDZ.

Impact: every mount of either component crashes (both are the voice-input surfaces). This is a hard blocker.

Fix (mechanical, as the peer notes): move the `useVoiceTriggerKey(...)` call to AFTER the `useVoiceInput(...)` / `useTextareaVoice(...)` destructuring that declares `isVoiceEnabled`, in both files. (Watch the ordering vs `voiceTriggerKeyRef` too.) Add a mount/render smoke test for both components so this TDZ can't regress silently — the existing tests exercise the hook in isolation and didn't catch the consumer wiring.

## ✅ Note — the phantom-recording + gate fixes in this same commit are correct
Separately from the crash: the two prior findings (phantom recording from the orphaned long-press timer; missing focus/enabled gate) ARE correctly fixed here — the 7th press does `stopImmediatePropagation()`+`preventDefault()` with the hook listener registered `capture=true` so it runs before the recorder handler, and the hook now honors `isVoiceEnabled`/`checkIsInputActive`. Those are good. The problem is purely that the new options wiring references `isVoiceEnabled` before it's declared. Once the call is reordered below the declaration, this should be a clean approve.

Credit to the peer review for catching the TDZ — I verified the fix logic but missed the declaration ordering.

Jerry-Xin (approved):
Re-verified at head e7e36086. My prior blocking 🔴 (the TDZ crash) is FIXED — flipping to APPROVE.

## ✅ Prior blocker — RESOLVED

**🔴 (was blocking) — TDZ crash from `isVoiceEnabled` used before declaration: FIXED in both consumers.**
The `useVoiceTriggerKey(...)` call is now placed AFTER the hook destructure that declares `isVoiceEnabled`, in both files:
- `VoiceInputIndicator.tsx`: `isVoiceEnabled` declared at ~line 92 (from `useVoiceInput`), and the `useVoiceTriggerKey(cb, { isVoiceEnabled, checkIsInputActive })` call moved down to ~line 199-207 — after the declaration, with an explicit guard comment ("Must be called AFTER useVoiceInput so isVoiceEnabled is already declared").
- `VoiceInputButton/index.tsx`: `isVoiceEnabled` declared at ~line 61 (from `useTextareaVoice`), hook call moved to ~line 79-87 after it, same guard comment.

No more use-before-declaration → no `ReferenceError` on mount. The delta is exactly this reorder (±15/-14 in the two consumer files, clean fast-forward +1) and touches nothing else.

## ✅ Prior fixes still intact
The earlier correct fixes are unaffected by the reorder: the 7th-press `stopImmediatePropagation()`+`preventDefault()` with the hook listener registered `capture=true` (no phantom recording), and the `isVoiceEnabled`/`checkIsInputActive` focus/enabled gate on the toggle listener. And the base feature (listener cleanup, 7-press counter bounds/reset/off-by-one, persisted key strictly validated to ShiftLeft/ShiftRight) remains sound.

## Note (non-blocking)
Please add a mount/render smoke test for both `VoiceInputIndicator` and `VoiceInputButton` so this TDZ ordering can't regress silently — the existing tests exercise the hook in isolation, not the consumer wiring where the crash lived.

Nice, clean fix. Credit to the peer review that caught the TDZ ordering.

OctoBoooot (commented):
## Re-review (DELTA) @ `e7e36086f685` — mochashanyao's TDZ crash 🔴 was REAL at `720fe6d2` and is now FIXED here

Single commit since the CR head: *"fix(voice): move useVoiceTriggerKey call after voice-service hook to fix TDZ"*, touching only the two voice components. It fixes a real mount crash mochashanyao caught — which I under-graded in my COMMENT. Byte-verified the fix in both files.

### The bug was real (credit mochashanyao)
At `720fe6d2`, both components passed `{ isVoiceEnabled }` into `useVoiceTriggerKey` **above** the `useTextareaVoice`/`useVoiceInput` destructure that declares `isVoiceEnabled` — a block-scoped `const` used before its declaration → `ReferenceError: Cannot access 'isVoiceEnabled' before initialization` on every render → **both components crashed on mount.** Exactly mochashanyao's "变量提前使用导致崩溃，需调整调用顺序." My COMMENT called the code "review-clean" — it wasn't; I verified the hook logic and fix mechanics but not the consumer's declaration order, so I missed a guaranteed crash. mochashanyao caught it. My miss to own.

### The fix is correct (verified at `e7e36086`)
- **`VoiceInputButton/index.tsx`** — `isVoiceEnabled` now declared at :61 (`useTextareaVoice` destructure, ends :63); `useVoiceTriggerKey(...)` moved to :79 with the use at :87 — **after** the declaration, with an explicit guard comment (:78 "Must be called AFTER useTextareaVoice so isVoiceEnabled is already declared").
- **`VoiceInputIndicator.tsx`** — same shape: `isVoiceEnabled` declared :92 (`useVoiceInput`, ends :95); `useVoiceTriggerKey(...)` moved to :199, use at :207 — after, with the same comment (:198).

No more TDZ reference in either file. `isVoiceEnabled` is declared before every use.

### No hook-order regression
Moving `useVoiceTriggerKey` below the voice-service hook is a **permanent** reorder — the hook call order is now consistent on every render (just different from before), which is legal. React only forbids *conditional* hooks or order that *changes between renders*; a fixed reorder is fine. Both hooks still run unconditionally each render. The gate logic from the prior fix (`isVoiceEnabled`/`checkIsInputActive` scoping, capture-phase `stopImmediatePropagation` on the completing press) is unchanged and still correct.

### Gate / CI
- **My verdict stays COMMENT** at this head — the code is now clean, but code CI (Build / vitest / secret-scan / osv) still hasn't registered at `e7e36086`, so the render path hasn't executed in CI. This TDZ is precisely why that matters: a mount test or Build type-check would have failed instantly on the `ReferenceError` — wiring these tests into CI (the standing 🟡) would have caught it before review. Once code CI lands green this is an APPROVE.
- `reviewDecision=CHANGES_REQUESTED` is currently held by two CRs still anchored to the stale `720fe6d2`: mochashanyao's (now fixed here — worth re-anchoring/clearing) and yujiawei's (whose two 🟡 are #774's color-input findings, mis-posted onto #773 — noted separately). Neither reflects a live defect in #773's code at `e7e36086`.

### Net
mochashanyao's use-before-declaration 🔴 was a real mount crash (both components) at `720fe6d2`, and `e7e36086` fixes it correctly by moving `useVoiceTriggerKey` below the hook that declares `isVoiceEnabled` (Button :61→:79, Indicator :92→:199). No hook-order regression; the earlier phantom-recording + focus-gate fixes are intact. Verdict COMMENT only because code CI hasn't run at this head — clean otherwise. Credit mochashanyao for the catch I missed, and yujiawei for the earlier phantom-recording bug. Superseding my stale COMMENT at `720fe6d2`.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #773 (octo-web)

Reviewed at head `252d6e71`. Scope: new `useVoiceTriggerKey` hook enabling users to switch the voice trigger key between Left Shift and Right Shift by rapidly pressing the current trigger key 7 times, plus consumer integration in `VoiceInputIndicator` and `VoiceInputButton`.

## Summary

This PR extracts the hardcoded `"ShiftLeft"` voice trigger into a configurable `useVoiceTriggerKey` hook with localStorage persistence. The toggle gesture (7 rapid presses) is a creative approach, and the hook's internal logic — counter, interval, persistence, modifier filtering — is correctly implemented in isolation. However, a fundamental interaction defect makes the toggle gesture unreachable in production: the same key presses that build the toggle counter simultaneously trigger the existing long-press voice recording timer, which fires before the toggle can complete.

## Verification

- ✅ **Hook logic correctness** — `useVoiceTriggerKey.ts:60-95`: press counting, interval gating, localStorage read/write with try-catch, and modifier filtering are all sound in isolation.
- ✅ **Return wiring** — `voiceTriggerKey` is consumed via `voiceTriggerKeyRef.current` in both `VoiceInputIndicator.tsx:318` and `VoiceInputButton/index.tsx:201`, replacing all previous hardcoded `"ShiftLeft"` references in keydown/keyup handlers.
- ✅ **i18n parity** — `en-US.json:1156-1157` and `zh-CN.json:1156-1157` both add `voiceInput.triggerKey.switchedToRight` and `switchedToLeft` keys; the i18n key strings match exactly what the code references via `t(msgKey)`.
- ✅ **Test coverage (hook)** — `useVoiceTriggerKey.test.ts` covers defaults, persistence, bidirectional toggle, insufficient presses, interval timeout, non-trigger key reset, modifier rejection, non-active shift rejection, and double toggle.
- ✅ **C1 parity** — Toggle logic is symmetric: ShiftLeft→ShiftRight and ShiftRight→ShiftLeft use identical counter/interval/toggle paths (`useVoiceTriggerKey.ts:89`).
- ✅ **INPUT-SHAPE** — `readStoredKey()` validates against the two literal enum values; any other localStorage value falls through to `"ShiftLeft"` default.

Static analysis only at head `252d6e71`; build and tests not executed in this environment.

## Findings

One P1 blocker described below; one P2 item.

### P1 — Toggle gesture is unreachable due to long-press recording timer conflict (`useVoiceTriggerKey.ts:63`)

The 7-rapid-press toggle gesture fundamentally conflicts with the voice recording long-press timer in both consumer components. Both `VoiceInputIndicator.tsx:318` and `VoiceInputButton/index.tsx:201` start a 500ms `shiftTimerRef` on the **first** keydown matching `voiceTriggerKeyRef.current`. The toggle requires 7 presses with each inter-press gap ≤ 500ms (`TOGGLE_INTERVAL_MS`), so the earliest the toggle can fire is ~600ms (6 gaps × ~100ms minimum). But the recording timer fires at 500ms from press 1, starting an unintended voice recording before the toggle completes.

```
Timeline (100ms inter-press gap):
t=0ms    Press 1: hook count=1, voice handler starts 500ms timer
t=100ms  Press 2: hook count=2
t=200ms  Press 3: hook count=3
t=300ms  Press 4: hook count=4
t=400ms  Press 5: hook count=5
t=500ms  Press 6: hook count=6, ⚠️ VOICE RECORDING TIMER FIRES
t=600ms  Press 7: hook count=7, TOGGLE fires — but recording already started
```

There is no timing window where the toggle can fire without the recording timer firing first. If the user presses slowly enough (gap > 500ms between keydown and keyup per cycle to let the recording timer fire and clear), the toggle's `TOGGLE_INTERVAL_MS` check resets the counter on each press. The feature is unreachable by design.

Additionally, after the toggle does fire, `voiceTriggerKey` changes (e.g., ShiftLeft→ShiftRight). The voice handler's pending `shiftTimerRef` from the last ShiftLeft keydown is not cleared: `handleKeyUp` checks `e.code === voiceTriggerKeyRef.current` which is now `"ShiftRight"`, so the ShiftLeft keyup does not match and the orphaned timer eventually fires, starting a recording with no key held.

Fix direction: The toggle gesture needs explicit coordination with the long-press handler. Options include: (a) have the toggle hook dispatch a custom event or set a shared ref that consumers check before starting the recording timer, (b) use a modifier-guarded gesture (e.g., Alt+Shift×7) so the long-press handler never activates during toggle attempts, or (c) move the toggle to a UI control (settings menu or button) instead of a keyboard gesture.

### P2 — Duplicate global keydown listeners cause double toggle and double Toast (`VoiceInputIndicator.tsx:71`)

Both `VoiceInputIndicator` (line 71) and `VoiceInputButton` (line 57) instantiate `useVoiceTriggerKey`, each registering its own `window.addEventListener("keydown", ...)` via the hook's `useEffect`. When both components are mounted simultaneously, two independent hook instances track the same keyboard events with separate `pressCountRef` counters. Both reach the 7-press threshold independently, each calling `onToggledRef.current(newKey)` — resulting in duplicate Toast notifications for every toggle.

The `writeStoredKey` call is idempotent (same value written twice), so localStorage stays consistent. But the user sees two identical toasts.

Fix direction: Lift `useVoiceTriggerKey` to a shared context/provider, or use a module-level singleton pattern so only one keydown handler is registered regardless of consumer count.

- `readStoredKey()` / `writeStoredKey()` try-catch around localStorage access for private browsing / quota-exceeded resilience.
- `onToggledRef` pattern avoids stale closures without re-registering the keydown listener on callback changes.
- The `handleKeyDown` in the hook correctly ignores `e.repeat`, preventing held-key auto-repeat from inflating the press counter.
- All `"ShiftLeft"` references in voice handler keydown/keyup logic replaced with `voiceTriggerKeyRef.current`; no stale hardcoded values remain.
- `voiceInputIndicator.test.tsx` and `voiceInputButtonShiftLongPress.test.tsx` existing long-press tests still work (they only press 1-2 times, well below toggle threshold of 7).
- C2 control-flow: the hook's `useCallback([], [])` with empty deps ensures stable `handleKeyDown` identity; `voiceTriggerKeyRef` bridge keeps current state available to the listener.

## Verdict: CHANGES_REQUESTED

The P1 interaction defect makes the core feature (toggle trigger key by pressing 7 times) unreachable in production — the long-press recording timer always fires before the toggle can complete. This must be addressed before merge. The P2 duplicate-listener issue is non-blocking but should be fixed to avoid duplicate toasts.

[Octo-Q] verdict: REQUEST_CHANGES — P1: toggle gesture unreachable due to fundamental timing conflict with long-press recording timer; feature is non-functional in production.

OctoBoooot (commented):
## Re-review @ `e7e36086f685` — triaging mochashanyao's two 🔴 + 🟡 (byte-verified): all three don't reproduce; the real residual is a test-coverage gap

Traced the full toggle-vs-record interaction at the current head (both listeners, the timer lifecycle, and the capture/bubble ordering — the exact interaction, not the handlers in isolation). None of the three reproduce. Credit mochashanyao for pressing this interaction — it's the subtle part of this PR, and it deserves a direct test even though the behavior is correct.

### The two listeners
- **Toggle** (`useVoiceTriggerKey.ts`): `addEventListener("keydown", handleKeyDown, true)` — **capture phase**, runs first, counts trigger-key keydowns independently of recording state.
- **Record** (`VoiceInputButton/index.tsx:298`): `addEventListener("keydown", handleKeyDown)` — **bubble phase**, runs after.

### 🔴#1 / #2 "record timer fires at 500 ms first → toggle can't fire" — REFUTED
The record timer (`shiftTimerRef`, `RECORDING_DELAY_MS=500`) is:
- armed on a trigger-key keydown **only if `shiftTimerRef.current === null`** (`index.tsx:214`), and
- **cleared by that same tap's keyup** — `if (e.code === voiceTriggerKeyRef.current && shiftTimerRef.current !== null) clearShiftTimer()` (`index.tsx:267-268`).

A rapid tap holds the key well under 500 ms, so the timer is torn down on keyup and never completes → recording never starts during the 7-tap gesture. The stated mechanism ("one timer armed on press 1 survives to fire at 500 ms") is byte-false: the timer is cleared and rebuilt per tap, not held across the sequence.

And even if recording *did* start, it wouldn't block the toggle: the toggle listener is **capture-phase and counts keydowns independently** — the record handler runs *after* it (bubble) and cannot un-count or stop it. So the 7th keydown always reaches the toggle counter. The hook test asserts exactly this ("switches ShiftLeft → ShiftRight after 7 rapid presses"). I could not construct an input where the record timer prevents the toggle.

### 🟡#3 "old record timer not cleared after toggle → state anomaly" — REFUTED (this is the original phantom-recording bug, already fixed)
On the completing (7th) keydown the toggle listener calls **`stopImmediatePropagation()`** in the capture phase, so the bubble-phase record handler **never runs on press 7 → never arms a timer there.** Presses 1–6 each clear their own timer on keyup (the key ref is still the old key during 1–6, so `e.code === ref` matches). The ref only flips on press 7, where no timer exists to orphan. Two keydowns of the *same* physical key cannot overlap without a keyup between them (autorepeat is filtered via `e.repeat`), so tap6's keyup provably precedes tap7's keydown — no timer straddles the flip. This is precisely the P1 fix for the phantom-recording bug; it holds.

### The one real residual — a test-coverage gap (🟡, matches Jerry-Xin's + my standing note)
The interaction mochashanyao is worried about is **not directly tested**: `useVoiceTriggerKey.test.ts` exercises the hook *in isolation* (fake keydown dispatch), and there's no component-level test that mounts `VoiceInputButton`/`VoiceInputIndicator` with *both* the toggle and record listeners live to assert (a) recording does not start during a 7-tap gesture and (b) no timer survives the toggle. The behavior is correct by construction, but a mount/interaction test would lock it against regression — same smoke-test recommendation Jerry-Xin and I already raised. That's the constructive home for this concern.

### Net
All three of mochashanyao's findings are byte-refuted at `e7e36086`: the toggle listener is capture-phase and counts independently of recording; the record timer is cleared per-tap (never survives to 500 ms during rapid taps); the completing press is bypassed via `stopImmediatePropagation` so no timer is armed or orphaned on the ref flip. If you have a concrete repro where a rapid 7-tap sequence fails to toggle or leaves a live timer, please share the exact timing — I'll re-open on evidence. Otherwise the recommended action is the **component-level interaction test**, not a code change.

My verdict stays **COMMENT** — the code is clean at this head; I'm held from APPROVE only because required code CI (Build / vitest / secret-scan / osv) still hasn't registered here, so none of these tests have executed in CI. Once code CI lands green this is an APPROVE. Credit mochashanyao for the TDZ catch earlier and for pressing this interaction — the interaction test is worth adding.

yujiawei (approved):
# Code Review — PR #773 (octo-web)

Re-reviewed at head SHA `e7e36086f685b008174359c701c4763467bb15a3` against `main` (merge-base `578dac69`). Scope: the new `useVoiceTriggerKey` hook, its wiring into `VoiceInputIndicator` and `VoiceInputButton`, the two i18n keys, and the unit test file. This head is verified independently — earlier approvals/CR at prior SHAs are not inherited.

**Verdict: Approve.** The three blockers raised on earlier commits (phantom recording, focus/enabled gate, TDZ crash on mount) are all resolved and byte-verified here. Remaining items are non-blocking (P2 / follow-up). Tests: `vitest` on `useVoiceTriggerKey.test.ts` → 17/17 pass locally.

## Spec compliance ✅

Implements exactly what the PR describes — no under-build, over-build, or deviation:
- 7 rapid presses (≤500 ms gaps) of the **current** trigger key toggles Left⇄Right Shift (`useVoiceTriggerKey.ts:114-140`).
- Choice persisted to `localStorage["octo_voice_trigger_key"]`, validated on read, defaulting to `ShiftLeft` (`readStoredKey`/`writeStoredKey`, both try/catch-wrapped).
- Locale-aware Toast on switch; both keys added under the correctly-registered `base` namespace in `en-US.json` + `zh-CN.json`.
- Both documented consumers wired; hard-coded `"ShiftLeft"` replaced by `voiceTriggerKeyRef.current`.

## Prior blockers — all resolved at this head ✅

1. **Phantom recording (was blocking): FIXED.** On the completing 7th press the hook calls `stopImmediatePropagation()` + `preventDefault()` (`useVoiceTriggerKey.ts:131-132`) and its keydown listener is registered **capture-phase** (`:147`), while both recorders stay bubble-phase (`VoiceInputIndicator.tsx:451`, `VoiceInputButton/index.tsx:298`). Since `window` is the top of the propagation path, the capture listener runs first and halts the event before the recorder can arm `shiftTimerRef` — so no unintended recording ~500 ms after a toggle. Mechanism is correct.
2. **Focus/enabled gate (was major): FIXED.** The hook honors `isVoiceEnabled` (effect early-returns, `:144`) and `checkIsInputActive` (`:104-109`). `VoiceInputIndicator` passes both; `VoiceInputButton` passes `isVoiceEnabled` (see P2 note below on the missing focus gate there).
3. **TDZ crash (was blocking): FIXED.** The `useVoiceTriggerKey(...)` call now sits after the `isVoiceEnabled` destructure in both consumers (`VoiceInputIndicator.tsx`: declared L92, called L199; `VoiceInputButton/index.tsx`: declared L61, called L79). No use-before-initialization.

## Non-blocking notes (P2 / follow-up)

The hook keeps **per-instance** `useState` and registers **one capture-phase window listener per mount**. Several surfaces co-mount multiple voice inputs simultaneously (e.g. `CreateTaskModal` renders two `VoiceInputButton`s for title + description; `MatterDetailPanel` up to three). Traced consequences at this head:

- **No dual Toast.** On the 7th press the first capture listener calls `stopImmediatePropagation()`, which blocks every *other* capture listener on `window` for that event — so exactly one `onToggled`/Toast fires. The "duplicate toast" concern does not reproduce.
- **Transient state desync (P2).** Because only the winning instance flips its `useState` and there is no `storage`-event listener or shared store, a co-mounted sibling keeps the old key until it remounts. `localStorage` stays correct, and actual recording is independently focus-gated, so this is a self-healing UX inconsistency, not a functional break. A shared store or a `storage` listener would also fix cross-tab sync.
- **`VoiceInputButton` toggle listener has no focus gate (P2).** Unlike `VoiceInputIndicator`, it does not pass `checkIsInputActive`, so its toggle listener is global whenever the component is mounted. Accidental activation is very unlikely in practice — the gesture needs 7 bare Shift presses with <500 ms gaps and **any** intervening non-trigger key resets the counter (`useVoiceTriggerKey.ts:98-100`) — but passing a focus gate here would make the two consumers symmetric and scope the gesture to the active field.
- **Test coverage (nit).** The suite exercises the hook in isolation and asserts `stopImmediatePropagation()` is *called* on the 7th press, but doesn't register a real bubble-phase recorder to assert the phantom-recording path is actually suppressed end-to-end. An integration-style test would lock in the fix.

None of the above blocks merge. Nice, well-scoped feature with solid persistence, counter bounds, and listener lifecycle hygiene.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #773 (octo-web)

## Summary

This PR introduces a configurable voice trigger key: users can switch between Left Shift and Right Shift as the long-press recording shortcut by rapidly pressing the current trigger key 7 times. A new `useVoiceTriggerKey` hook manages the toggle gesture, persists the preference to `localStorage`, and calls `stopImmediatePropagation()` on the completing press to prevent a phantom recording. Both `VoiceInputIndicator` (main message input) and `VoiceInputButton` (textarea input) are updated to replace the hardcoded `"ShiftLeft"` with the configurable key.

The implementation is well-structured. The capture-phase listener registration, ref-based stale-closure avoidance, and `stopImmediatePropagation` phantom-recording fix are all correct and well-tested.

## Verification

- ✅ **Phantom-recording prevention** — the toggle hook registers with `capture=true` (`useVoiceTriggerKey.ts:145`) and calls `e.stopImmediatePropagation()` + `e.preventDefault()` on the 7th press (`:124–126`). The recording handlers in both consumers register in bubble phase (`VoiceInputIndicator.tsx:448`, `VoiceInputButton/index.tsx:293`), so the capture-phase stop correctly prevents the recording handler from arming `shiftTimerRef`.
- ✅ **Complete hardcoded-key replacement** — grepped both consumers for `"ShiftLeft"` in code: all runtime comparisons now use `voiceTriggerKeyRef.current`. Only stale comments remain (Nit below).
- ✅ **Stale-closure avoidance** — both consumers store `voiceTriggerKey` in a ref and read `voiceTriggerKeyRef.current` inside long-lived keyboard effects. The recording handler `useEffect` deps correctly exclude `voiceTriggerKey` to avoid re-registration.
- ✅ **i18n parity** — both `en-US.json` and `zh-CN.json` add the same two keys (`voiceInput.triggerKey.switchedToRight/switchedToLeft`). Both consumers use the same `t(msgKey)` pattern.
- ✅ **Hook ordering** — `useVoiceTriggerKey` is called unconditionally after `useVoiceInput`/`useTextareaVoice`, satisfying React hook rules.
- ✅ **Test coverage** — 15 tests cover toggle in both directions, threshold enforcement (6 vs 7), `stopImmediatePropagation` on completing press only, `isVoiceEnabled` gate, `checkIsInputActive` gate, counter reset (timeout, non-trigger key, modifiers, autorepeat), non-active shift key rejection, and double-toggle.

Static analysis only at head `e7e36086f685b008174359c701c4763467bb15a3`; build and tests not executed in this environment.

## Findings

### P2 — Missing focus gate in VoiceInputButton toggle (`packages/dmworkbase/src/Components/VoiceInputButton/index.tsx:84`)

`VoiceInputIndicator` passes `{ isVoiceEnabled, checkIsInputActive }` to the hook (`:203`), so the toggle gesture is only active when the associated input is focused. `VoiceInputButton` passes only `{ isVoiceEnabled }` — no focus gate. This means every mounted `VoiceInputButton` independently tracks rapid Shift presses and can toggle the key + show a Toast regardless of which element has focus. If multiple `VoiceInputButton` instances are on screen (e.g. in thread reply inputs), a single 7-press gesture toggles the key N times and produces N duplicate Toasts.

The recording handler already gates on `document.activeElement !== inputRef.current` (`:201`), so phantom recordings cannot occur — the consequence is limited to duplicate UI feedback and redundant `localStorage` writes. Still, passing a focus gate would bring parity with `VoiceInputIndicator` and eliminate the multi-instance noise:

```ts
const checkIsInputActive = useCallback(
  () => document.activeElement === inputRef.current,
  [inputRef]
);
// then: { isVoiceEnabled, checkIsInputActive }
```

### Nit — Stale `ShiftLeft` references in comments (`packages/dmworkbase/src/Components/MessageInput/VoiceInputIndicator.tsx:69`)

Lines 69, 273, and 435 still say "ShiftLeft" in comments, but the code now uses the configurable `voiceTriggerKeyRef.current`. Cosmetic, but worth aligning for future readers.

- `readStoredKey()` validates `localStorage` values against the `"ShiftLeft" | "ShiftRight"` union — invalid/unknown values fall back to `"ShiftLeft"` (`useVoiceTriggerKey.ts:12`).
- `localStorage` access is wrapped in try/catch — safe in environments where storage is unavailable (`:10`, `:22`).
- First press after reset: `lastPressTimeRef.current = 0` → `gap = Date.now() - 0` (huge) → `pressCountRef = 1`. Correct — starts the sequence from 1, not 0.
- The `e.repeat` guard (`:97`) prevents keyboard autorepeat from counting toward the toggle.
- Modifier-held presses (`e.metaKey || e.ctrlKey || e.altKey`) correctly reset the counter and do not count.
- The `inputLongPress` i18n tooltip ("Voice input (hold Shift)") uses generic "Shift" — works correctly regardless of which trigger key is active.
- `handleKeyUp` in both consumers: after a toggle, the old key's keyup won't match `voiceTriggerKeyRef.current` (now the new key), and `shiftTimerRef` was never armed — no orphan timer or phantom recording.

## Verdict: APPROVED

Well-implemented feature with correct phantom-recording prevention, proper event-phase ordering, and solid test coverage. The P2 (missing focus gate in `VoiceInputButton`) is a minor multi-instance UX inconsistency that does not break functionality.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The implementation correctly prevents phantom recordings via capture-phase event interception, all hardcoded "ShiftLeft" references are replaced at the code level, and the test suite comprehensively covers the hook contract. The P2 (VoiceInputButton missing focus gate) is a non-blocking UX parity item.

yujiawei (approved):
# Code Review — PR #773 (octo-web)

Re-reviewed at head SHA `e7e36086f685b008174359c701c4763467bb15a3` against `main` (merge-base `578dac69`). Scope: the new `useVoiceTriggerKey` hook, its wiring into `VoiceInputIndicator` and `VoiceInputButton`, the two i18n keys, and the unit test file. This head is verified independently — approvals/CR at earlier SHAs are not inherited.

**Verdict: Approve.** All three blockers raised on earlier commits (phantom recording, focus/enabled gate, TDZ crash on mount) are resolved and byte-verified here. The remaining `CHANGES_REQUESTED` on this PR was submitted against a **stale head (`252d6e71`, two commits behind)** and its P1 does not reproduce at the current head (analysis below). Remaining items are non-blocking (P2 / nit).

## Spec compliance

Implements exactly what the PR describes — no under-build, over-build, or deviation:
- 7 rapid presses (≤500 ms gaps) of the **current** trigger key toggle Left⇄Right Shift (`useVoiceTriggerKey.ts:114-140`).
- Choice persisted to `localStorage["octo_voice_trigger_key"]`, validated on read, defaulting to `ShiftLeft` (`readStoredKey`/`writeStoredKey`, both try/catch-wrapped).
- Locale-aware Toast on switch; both keys added under the `base` namespace in `en-US.json` + `zh-CN.json`.
- Both documented consumers wired; hard-coded `"ShiftLeft"` replaced by `voiceTriggerKeyRef.current`.

## The contested "toggle is unreachable" P1 — does not reproduce

A prior review flagged, as blocking, that the 7-tap toggle can never complete because the long-press recording timer fires at ~500 ms (before the 7th press at ~600 ms) and starts an unintended recording. **This does not hold at the current head**, for three independently-traced reasons:

1. **The record timer is per-tap, not a single timer spanning the gesture.** It is armed on a trigger-key keydown *only* when `shiftTimerRef.current === null` (`VoiceInputButton/index.tsx:214`, `VoiceInputIndicator.tsx:330`) and is **cleared by that same tap's keyup** (`VoiceInputButton/index.tsx:267-270`, `VoiceInputIndicator.tsx:399-402`). In a rapid tap the key is held well under 500 ms, so each tap's timer is torn down ~420 ms before it could fire. The premise "the timer armed on press 1 survives to t=500 ms" is byte-false — press 1's own keyup cancels it.
2. **At most one record timer is ever pending** — the `=== null` guard blocks arming a second one.
3. **The completing (7th) press is shielded.** The toggle listener is registered **capture-phase** (`useVoiceTriggerKey.ts:147`, `addEventListener(..., true)`) while both recorders are bubble-phase (`VoiceInputButton/index.tsx:298`, `VoiceInputIndicator.tsx:451`). Capture runs first on `window`, and on reaching count 7 the hook calls `stopImmediatePropagation()` + `preventDefault()` (`useVoiceTriggerKey.ts:131-132`), so the record handler never arms a timer on the 7th press. This is the same mechanism that closes the earlier phantom-recording bug. Autorepeat is filtered via `e.repeat` (`useVoiceTriggerKey.ts:92`), so a keyup always precedes the next keydown for the same physical key.

No timing (including an adversarial keyup(N)-after-keydown(N+1) reorder) was found where the record timer survives to 500 ms or an orphaned timer fires after the ref flip. The 7-tap gesture reaches the toggle and completes.

## Prior blockers — all resolved at this head

1. **Phantom recording (was blocking): FIXED** — capture-phase `stopImmediatePropagation()` on the 7th press (`useVoiceTriggerKey.ts:131,147`).
2. **Focus/enabled gate (was major): FIXED** — hook honors `isVoiceEnabled` (`:144`) and `checkIsInputActive` (`:104-109`). `VoiceInputIndicator` passes both; `VoiceInputButton` passes `isVoiceEnabled` (see P2 note).
3. **TDZ crash (was blocking): FIXED** — `useVoiceTriggerKey(...)` now called after the `isVoiceEnabled` destructure in both consumers (`VoiceInputButton/index.tsx`: L61 declared, L79 called; `VoiceInputIndicator.tsx`: declared before call). No use-before-initialization.

## Non-blocking notes (P2 / nit)

- **Multi-instance state desync (P2).** The hook keeps per-instance `useState` and one capture-phase `window` listener per mount. Several surfaces co-mount multiple voice inputs (e.g. `MatterDetailPanel` renders 3 `VoiceInputButton`s; `MessageInput` mounts `VoiceInputIndicator`). On the 7th press the winning listener's `stopImmediatePropagation()` blocks the other capture listeners, so **exactly one Toast fires** — the "duplicate toast" concern does not reproduce on the completing press. A residual narrow race exists: presses 1–6 propagate to every mounted listener, leaving losers stuck at count=6; an 8th rapid press within 500 ms could drive a loser 6→7 and fire a second toggle. Requires ≥2 co-mounted voice components AND an 8th rapid press; self-corrects. `localStorage` stays correct and recording is independently focus-gated, so this is a UX inconsistency, not a functional break. A shared store or a `storage`-event listener would also fix cross-tab sync.
- **`VoiceInputButton` toggle listener has no focus gate (P2).** Unlike `VoiceInputIndicator`, it does not pass `checkIsInputActive` (`VoiceInputButton/index.tsx:87`), so its toggle listener is global whenever mounted. Accidental activation is very unlikely — the gesture needs 7 bare Shift presses <500 ms apart with **any** intervening non-trigger key resetting the counter (`useVoiceTriggerKey.ts:98-99`), and the only effect would be an informational Toast + a persisted Left/Right swap (no recording). Threading the focus predicate here would make the two consumers symmetric.
- **Test coverage (nit).** The suite (17 tests) exercises the hook in isolation and asserts `stopImmediatePropagation()` is *called* on the 7th press, but no component-level test mounts a consumer with both the toggle and the bubble-phase recorder live to assert end-to-end that (a) no recording starts during a 7-tap gesture and (b) no timer survives the toggle. An interaction test would lock this behavior against regression — the behavior is correct by construction, but this is the natural home for the concern raised in review.

## Circuit-breaker note

This PR has now been through 4+ review/fix rounds. The correctness is sound at this head and the remaining `CHANGES_REQUESTED` targets a stale SHA. Recommendation: land on green CI; treat the P2 multi-instance dedup and the follow-up interaction test as a fast-follow rather than re-opening another blocking round.

None of the above blocks merge. Well-scoped feature with solid persistence, counter bounds, and listener-lifecycle hygiene.

Issue #749, opened 2026-07-14
[Docs] Table column width is still very hard to resize by dragging (regression after #715)


Jerry-Xin:
## Summary

Table column width is still very difficult to resize in the Docs editor, even after the fix in #715 (which was intended to allow the column-resize drag to start). A leadership user reports that dragging the column border to change the width remains unreliable and hard to trigger in practice.

This is a regression / incomplete-fix follow-up to the previously closed #621 and merged PR #715.

## Steps to Reproduce

1. Open a document in the Docs editor.
2. Insert or focus a table.
3. Hover the column border and attempt to drag it to change the column width.

## Expected

Dragging a column border smoothly and reliably resizes the column width.

## Actual

The resize drag is very hard to initiate and the column width barely responds to dragging. The interaction feels unreliable and does not consistently pick up the drag on the column border.

## Notes

- Related: #621 (original table column-resize report, closed), PR #715 (attempted fix, merged), #625 / PR #716 (table toolbar overlap).
- The merged fix did not fully resolve the resize UX; needs re-investigation of the resize-handle hit area and drag-trigger threshold.

Jerry-Xin:
Additional repro detail from the reporter:

The behavior is inverted from what you'd expect:
- When the cursor is placed **exactly on the column border line and the resize (drag) icon appears**, dragging does **NOT** work.
- When the cursor is offset slightly away from the line (where the resize icon does **NOT** show), dragging the column width **does** work.

This strongly suggests the actual drag hit-area / event handler is misaligned with the visual resize-cursor hover zone — the element that shows the resize cursor is not the same element (or z-index layer) that receives and handles the drag, so the working drag target sits just outside the visible handle. Likely an overlapping element (e.g. the table toolbar/handle from #625) is intercepting pointer events exactly where the resize cursor shows.

PR #775, opened 2026-07-15
feat(docs): freeze panes for editor tables with ribbon entry and right-click submenu

### LINKED ISSUE #755: [Docs] Add freeze rows/columns (freeze panes) support for tables
## Summary

Add a freeze rows/columns (freeze panes) feature to tables in the Docs editor, and surface it via a **freeze button in the ribbon/toolbar** (not only the right-click context menu), so selected header rows and/or leading columns stay pinned in place while scrolling through large tables.

## Motivation

For wide or tall tables, users lose track of which row/column a cell belongs to once the header scrolls out of view. Freezing the top row(s) and/or first column(s) keeps the headers visible, matching the familiar behavior in spreadsheet tools (Excel / Google Sheets / WPS).

## Current State / Bug

- A freeze option already exists in the table **right-click context menu**.
- However, clicking "freeze" **intermittently fails to open the submenu** for choosing freeze-to-row / freeze-to-column, so the user cannot complete the freeze action reliably.

## Requested Changes

1. **Move/add the freeze entry point to the ribbon (toolbar)** — expose a dedicated freeze button in the ribbon, rather than relying on the right-click menu.
2. Fix the intermittent failure where the freeze submenu (freeze-to-row / freeze-to-column) does not appear on click.

## Proposed Behavior

- Allow the user to freeze the top N rows (typically the header row).
- Allow the user to freeze the first N columns (typically the row-label column).
- Frozen rows/columns remain fixed while the rest of the table scrolls horizontally/vertically.
- Freeze control is reachable from the ribbon and behaves reliably (submenu always opens).

## Notes

- Feature/UX request from a leadership user.
- Should work consistently with existing table interactions (column resize, toolbar).

ploy-elison:
## Summary

Adds **freeze panes (冻结行列)** to the Docs editor table (closes #755): a ribbon entry, a reliable right-click submenu, and sticky freezing of the top N rows / first N columns.

> **Draft on purpose — do not merge.** Opened as a draft to carry review. Promotion to a normal PR (`gh pr ready`) is out of scope for this change and happens only after review + deploy + QA are green and the change is signed off.

## What changed

- **Ribbon button.** A dedicated **Freeze** control sits next to the table insert control. It is disabled unless the caret is inside a table and reads active while the current table has any frozen band. Clicking opens a small popover with: freeze header row, freeze up to current row, freeze first column, freeze up to current column, unfreeze.
- **Right-click submenu.** The table context menu gains an inline **Freeze ▸** entry that expands the same five actions.
- **Freeze behaviour.** Freezes the top **N** rows and/or first **N** columns; frozen cells stay fixed on both horizontal and vertical scroll. Offsets are measured from the live DOM, so they stay correct after column resizes.

## The "submenu occasionally doesn't pop up" bug

On the current `main` the table context menu is a **flat** menu — there is no freeze option and no submenu at all, so there was no existing flaky submenu to reproduce a red against. Rather than reintroduce the classic failure mode, the freeze submenu is built to be **reliable by construction**:

- it is a **click-toggled** inline accordion (state on the menu component), **not** a CSS-hover flyout — so there is no diagonal-gap / pointer-traversal race, and
- the outside-click close handler ignores clicks inside the menu, so the `mousedown` that immediately follows the trigger click can never race the submenu closed.

A regression test locks this in: after clicking **Freeze**, a `mousedown` landing inside the menu leaves the sub-options open.

## View-state vs. content model (Yjs)

Freeze is **view-state**, deliberately. The frozen tables are tracked in a ProseMirror plugin (keyed by table position, remapped across edits) and rendered with `position: sticky`; **nothing is written to the Y.Doc**. This keeps the schema, `SCHEMA_VERSION` (currently 18) and the backend contract untouched, and there is no collaborative-sync surface to conflict on.

Trade-off: freeze is **per-client and not persisted across reload or shared with collaborators**. Making it durable + shared would be a coordinated schema-version bump adding `frozenRows`/`frozenCols` attributes to the `table` node (the same class of change as v18) plus a backend stub update — intentionally left as a follow-up, not folded into a frontend-only change.

## Interplay with existing table interactions (#749, #625/#716)

- Chromium ignores `position: sticky` on `<td>/<th>` under `border-collapse: collapse`, and the table's default `overflow: hidden` captures the sticky offset — **both were caught in browser verification** (the first attempt scrolled the "frozen" band away). A frozen table is therefore flipped to `border-collapse: separate; border-spacing: 0; overflow: visible`, with borders redrawn to single 1px lines (table paints the outer top+left, each cell its own right+bottom).
- **#749:** sticky frozen cells create stacking contexts that would otherwise occlude the un-z-indexed column-resize handle at `right:-2px`. The handle is lifted above frozen cells so a frozen column can still be resized.
- The freeze UI reuses the existing table toolbar/context-menu styling, so it stacks consistently with #625/#716.

## Testing

- **45 unit tests pass** (`TableFreeze.test.ts`, `TableControls.test.tsx`, `i18n.test.ts`), covering the command logic (toggle/through-row/through-column/clamp/clear), the plugin's position remapping across edits and table deletion, the submenu reliability, and the sticky-style DOM assignment (classes, `position/top/left/z-index`, wrapper + separate-border classes).
- **Full docs suite:** no new failures introduced (the 4 pre-existing `EditorShell.test.tsx` failures are unrelated to this change and present on `main`).
- **Typecheck:** clean for all files in this change (the 2 pre-existing `tsc` errors are in unrelated files).
- **Browser verification (Chromium):** a frozen table (header row + first column) was scrolled down and right; the header row and first column stayed pinned while the body scrolled — captured below.

### Freeze holding while scrolled (down + right)

_The header row (Col …) stays at the top and the first column (Row …) stays at the left after scrolling; columns 1–2 have scrolled out so Col 3 is the next visible column._

## Related

- GitHub issue: #755
- Related: #749 (column resize), #625 / #716 (table toolbar)

### DIFF
--- a/packages/docs/src/editor/TableFreeze.ts
+++ b/packages/docs/src/editor/TableFreeze.ts
@@ -1,0 +1,363 @@
+// Table freeze panes (#755, XIN-1096).
+//
+// Adds "freeze rows / freeze columns" (冻结窗格) to the Docs editor table without touching the
+// collaborative content model. Freeze is a VIEW-STATE feature:
+//
+//   - The set of frozen tables (keyed by the table node's document position) lives in a ProseMirror
+//     plugin state, NOT in the Y.Doc. Nothing is written to the schema, so SCHEMA_VERSION and the
+//     backend stub are untouched and there is no collab-sync surface to conflict on. The trade-off
+//     is that freeze is per-client and does not persist across reload — documented in the PR. (If
+//     leadership later wants durable, shared freeze, it becomes a coordinated schema-version bump
+//     adding frozenRows/frozenCols attrs to the `table` node, the same class of change as v18.)
+//
+//   - The plugin's view applies `position: sticky` to the first N rows / first N columns of each
+//     frozen table, measuring row heights / column widths from the live DOM so the sticky offsets
+//     line up with resized columns (#749). A frozen table is also flipped to separate borders
+//     (`octo-frozen-table` class) because Chromium ignores sticky on `<td>/<th>` under
+//     border-collapse:collapse and the table's default overflow:hidden captures the sticky offset —
+//     both verified in the browser. Styling the cells' own DOM is safe under collaboration:
+//     TableCellView.ignoreMutation already discards attribute/style mutations on the cell element,
+//     so these writes never round-trip as document edits.
+//
+// The ribbon button (Toolbar) and the right-click submenu (TableControls) both drive the commands
+// declared here; they never manipulate freeze state directly.
+
+import { Extension } from '@tiptap/core'
+import { Plugin, PluginKey } from '@tiptap/pm/state'
+import type { EditorState } from '@tiptap/pm/state'
+import type { EditorView } from '@tiptap/pm/view'
+import { TableMap, selectionCell } from '@tiptap/pm/tables'
+
+/** How many leading rows / columns of a table are frozen. `{ rows: 0, cols: 0 }` means unfrozen. */
+export interface FreezeSpec {
+  rows: number
+  cols: number
+}
+
+/** Frozen tables, keyed by the document position immediately BEFORE the table node. */
+export type FreezeMap = Map<number, FreezeSpec>
+
+interface FreezeMeta {
+  pos: number
+  spec: FreezeSpec
+}
+
+export const tableFreezeKey = new PluginKey<FreezeMap>('octoTableFreeze')
+
+/**
+ * Prefix sums of `sizes`: element i is the sum of every size before it (so the first is always 0).
+ * Used to turn a list of frozen row heights / column widths into the sticky `top` / `left` offset
+ * for each frozen band. Pure so it can be unit-tested without a real layout (jsdom reports 0-size
+ * boxes, so the browser is where the pixel offsets are actually verified).
+ */
+export function cumulativeOffsets(sizes: number[]): number[] {
+  const out: number[] = []
+  let acc = 0
+  for (const s of sizes) {
+    out.push(acc)
+    acc += s
+  }
+  return out
+}
+
+interface CellContext {
+  /** Position immediately before the enclosing table node (the FreezeMap key). */
+  tablePos: number
+  /** Zero-based row / column index of the cell holding the selection head. */
+  rowIndex: number
+  colIndex: number
+  /** Table dimensions, used to clamp a freeze request to what the table can actually offer. */
+  rowCount: number
+  colCount: number
+}
+
+/**
+ * Resolve the table cell around the selection head, or null when the caret is not inside a table.
+ * Returns the enclosing table's position (used as the freeze key) plus the clicked cell's row/column
+ * index so "freeze up to this row/column" can turn the caret position into a frozen band count.
+ */
+function cellContext(state: EditorState): CellContext | null {
+  let $cell
+  try {
+    $cell = selectionCell(state)
+  } catch {
+    return null
+  }
+  if (!$cell) return null
+  const table = $cell.node(-1)
+  if (!table || table.type.name !== 'table') return null
+  const map = TableMap.get(table)
+  const tableStart = $cell.start(-1)
+  const rect = map.findCell($cell.pos - tableStart)
+  return {
+    tablePos: $cell.before(-1),
+    rowIndex: rect.top,
+    colIndex: rect.left,
+    rowCount: map.height,
+    colCount: map.width,
+  }
+}
+
+/** Whether the caret currently sits inside a table (gates the freeze UI). */
+export function isInTable(state: EditorState): boolean {
+  return cellContext(state) !== null
+}
+
+/**
+ * Freeze spec for the table around the current selection, `{ rows: 0, cols: 0 }` when unfrozen or
+ * outside a table. The ribbon/menu read this to show active state.
+ */
+export function getFreezeSpec(state: EditorState): FreezeSpec {
+  const ctx = cellContext(state)
+  if (!ctx) return { rows: 0, cols: 0 }
+  return tableFreezeKey.getState(state)?.get(ctx.tablePos) ?? { rows: 0, cols: 0 }
+}
+
+// --- rendering (plugin view) -------------------------------------------------------------------
+
+const FROZEN_ATTR = 'data-octo-frozen'
+
+export function clearFrozenCell(cell: HTMLElement): void {
+  cell.style.position = ''
+  cell.style.top = ''
+  cell.style.left = ''
+  cell.style.zIndex = ''
+  cell.classList.remove('octo-frozen-cell', 'octo-frozen-row-cell', 'octo-frozen-col-cell')
+  cell.removeAttribute(FROZEN_ATTR)
+}
+
+/**
+ * Apply sticky styling to the first `spec.rows` rows and `spec.cols` columns of `table`. Offsets are
+ * measured from the live DOM so they follow column resizes. A frozen row's cells stick to the top; a
+ * frozen column's cells stick to the left; a corner cell (both) stacks above either single axis.
+ */
+export function applyFrozenStyles(table: HTMLTableElement, spec: FreezeSpec): void {
+  const rows = Array.from(table.rows)
+  if (!rows.length) return
+  const frozenRows = Math.max(0, Math.min(spec.rows, rows.length))
+  const colCount = rows[0].cells.length
+  const frozenCols = Math.max(0, Math.min(spec.cols, colCount))
+  const anyFrozen = frozenRows > 0 || frozenCols > 0
+
+  // A frozen table switches to separate borders and is no longer its own scroll container — both
+  // required for position:sticky on cells to actually hold (Chromium ignores sticky cells under
+  // border-collapse:collapse, and the table's default overflow:hidden captures the sticky offset).
+  // The border redraw lives in CSS keyed off this class.
+  table.classList.toggle('octo-frozen-table', anyFrozen)
+
+  // A frozen header needs a vertical scroll container to stick within — the wrapper already scrolls
+  // horizontally, so bound its height (CSS) only while rows are frozen.
+  const wrapper = table.closest('.tableWrapper')
+  if (wrapper) wrapper.classList.toggle('octo-has-frozen-rows', frozenRows > 0)
+
+  const rowTops = cumulativeOffsets(rows.slice(0, frozenRows).map((r) => r.offsetHeight))
+  const colLefts = cumulativeOffsets(
+    Array.from(rows[0].cells)
+      .slice(0, frozenCols)
+      .map((c) => c.offsetWidth),
+  )
+
+  rows.forEach((row, rIdx) => {
+    Array.from(row.cells).forEach((cell, cIdx) => {
+      const inFrozenRow = rIdx < frozenRows
+      const inFrozenCol = cIdx < frozenCols
+      if (!inFrozenRow && !inFrozenCol) return
+      cell.setAttribute(FROZEN_ATTR, '')
+      cell.classList.add('octo-frozen-cell')
+      cell.style.position = 'sticky'
+      if (inFrozenRow) {
+        cell.classList.add('octo-frozen-row-cell')
+        cell.style.top = `${rowTops[rIdx]}px`
+      }
+      if (inFrozenCol) {
+        cell.classList.add('octo-frozen-col-cell')
+        cell.style.left = `${colLefts[cIdx]}px`
+      }
+      cell.style.zIndex = inFrozenRow && inFrozenCol ? '4' : inFrozenRow ? '3' : '2'
+    })
+  })
+}
+
+/**
+ * Plugin view that keeps the DOM in sync with the freeze state: on every editor update it clears any
+ * previously frozen cells and re-applies sticky styling for the currently frozen tables. Re-measuring
+ * each time keeps the offsets correct after column resizes, row insert/delete, or remote edits.
+ */
+class FreezeStyleView {
+  private view: EditorView
+
+  constructor(view: EditorView) {
+    this.view = view
+    this.render()
+  }
+
+  update(): void {
+    this.render()
+  }
+
+  destroy(): void {
+    this.reset()
+  }
+
+  private reset(): void {
+    this.view.dom
+      .querySelectorAll<HTMLElement>(`[${FROZEN_ATTR}]`)
+      .forEach((el) => clearFrozenCell(el))
+    this.view.dom
+      .querySelectorAll('.octo-has-frozen-rows')
+      .forEach((el) => el.classList.remove('octo-has-frozen-rows'))
+    this.view.dom
+      .querySelectorAll('.octo-frozen-table')
+      .forEach((el) => el.classList.remove('octo-frozen-table'))
+  }
+
+  private render(): void {
+    this.reset()
+    const map = tableFreezeKey.getState(this.view.state)
+    if (!map || map.size === 0) return
+    map.forEach((spec, pos) => {
+      if (spec.rows <= 0 && spec.cols <= 0) return
+      let dom: Node | null = null
+      try {
+        dom = this.view.nodeDOM(pos)
+      } catch {
+        dom = null
+      }
+      if (!(dom instanceof HTMLElement)) return
+      const table =
+        dom instanceof HTMLTableElement ? dom : dom.querySelector<HTMLTableElement>('table')
+      if (!table) return
+      applyFrozenStyles(table, spec)
+    })
+  }
+}
+
+/**
+ * The freeze plugin: holds the FreezeMap, remaps it across document edits (dropping entries whose
+ * table was deleted), and mounts the styling view.
+ */
+export function tableFreezePlugin(): Plugin<FreezeMap> {
+  return new Plugin<FreezeMap>({
+    key: tableFreezeKey,
+    state: {
+      init: () => new Map(),
+      apply(tr, value) {
+        let next = value
+        if (tr.docChanged && value.size) {
+          const mapped: FreezeMap = new Map()
+          value.forEach((spec, pos) => {
+            const mappedPos = tr.mapping.map(pos, -1)
+            const node = tr.doc.nodeAt(mappedPos)
+            if (node && node.type.name === 'table') mapped.set(mappedPos, spec)
+          })
+          next = mapped
+        }
+        const meta = tr.getMeta(tableFreezeKey) as FreezeMeta | undefined
+        if (meta) {
+          next = new Map(next)
+          if (meta.spec.rows <= 0 && meta.spec.cols <= 0) next.delete(meta.pos)
+          else next.set(meta.pos, meta.spec)
+        }
+        return next
+      },
+    },
+    view(view) {
+      return new FreezeStyleView(view)
+    },
+  })
+}
+
+declare module '@tiptap/core' {
+  interface Commands<ReturnType> {
+    tableFreeze: {
+      /** Set an explicit freeze extent on the current table (fields left undefined keep their value). */
+      setTableFreeze: (spec: Partial<FreezeSpec>) => ReturnType
+      /** Toggle freezing the top header row (rows 0 ↔ 1), keeping the frozen-column count. */
+      toggleFreezeHeaderRow: () => ReturnType
+      /** Toggle freezing the first column (cols 0 ↔ 1), keeping the frozen-row count. */
+      toggleFreezeFirstColumn: () => ReturnType
+      /** Freeze every row down to and including the row holding the caret. */
+      freezeThroughSelectedRow: () => ReturnType
+      /** Freeze every column up to and including the column holding the caret. */
+      freezeThroughSelectedColumn: () => ReturnType
+      /** Remove all freezing from the current table. */
+      clearTableFreeze: () => ReturnType
+    }
+  }
+}
+
+/** Table freeze-panes extension: view-state freeze plus the commands the UI drives. */
+export const TableFreeze = Extension.create({
+  name: 'tableFreeze',
+
+  addProseMirrorPlugins() {
+    return [tableFreezePlugin()]
+  },
+
+  addCommands() {
+    const dispatchFreeze = (
+      state: EditorState,
+      dispatch: ((tr: import('@tiptap/pm/state').Transaction) => void) | undefined,
+      compute: (current: FreezeSpec, ctx: CellContext) => FreezeSpec,
+    ): boolean => {
+      const ctx = cellContext(state)
+      if (!ctx) return false
+      const current = tableFreezeKey.getState(state)?.get(ctx.tablePos) ?? {
+        rows: 0,
+        cols: 0,
+      }
+      const wanted = compute(current, ctx)
+      const spec: FreezeSpec = {
+        rows: Math.max(0, Math.min(Math.round(wanted.rows), ctx.rowCount)),
+        cols: Math.max(0, Math.min(Math.round(wanted.cols), ctx.colCount)),
+      }
+      if (dispatch) {
+        const meta: FreezeMeta = { pos: ctx.tablePos, spec }
+        dispatch(state.tr.setMeta(tableFreezeKey, meta))
+      }
+      return true
+    }
+
+    return {
+      setTableFreeze:
+        (spec) =>
+        ({ state, dispatch }) =>
+          dispatchFreeze(state, dispatch, (current) => ({
+            rows: spec.rows ?? current.rows,
+            cols: spec.cols ?? current.cols,
+          })),
+      toggleFreezeHeaderRow:
+        () =>
+        ({ state, dispatch }) =>
+          dispatchFreeze(state, dispatch, (current) => ({
+            rows: current.rows >= 1 ? 0 : 1,
+            cols: current.cols,
+          })),
+      toggleFreezeFirstColumn:
+        () =>
+        ({ state, dispatch }) =>
+          dispatchFreeze(state, dispatch, (current) => ({
+            rows: current.rows,
+            cols: current.cols >= 1 ? 0 : 1,
+          })),
+      freezeThroughSelectedRow:
+        () =>
+        ({ state, dispatch }) =>
+          dispatchFreeze(state, dispatch, (current, ctx) => ({
+            rows: ctx.rowIndex + 1,
+            cols: current.cols,
+          })),
+      freezeThroughSelectedColumn:
+        () =>
+        ({ state, dispatch }) =>
+          dispatchFreeze(state, dispatch, (current, ctx) => ({
+            rows: current.rows,
+            cols: ctx.colIndex + 1,
+          })),
+      clearTableFreeze:
+        () =>
+        ({ state, dispatch }) =>
+          dispatchFreeze(state, dispatch, () => ({ rows: 0, cols: 0 })),
+    }
+  },
+})
--- a/packages/docs/src/editor/TableFreeze.ts
+++ b/packages/docs/src/editor/TableFreeze.ts
@@ -26,6 +26,7 @@ import { Extension } from '@tiptap/core'
 import { Plugin, PluginKey } from '@tiptap/pm/state'
 import type { EditorState } from '@tiptap/pm/state'
 import type { EditorView } from '@tiptap/pm/view'
+import type { Node as PMNode } from '@tiptap/pm/model'
 import { TableMap, selectionCell } from '@tiptap/pm/tables'
 
 /** How many leading rows / columns of a table are frozen. `{ rows: 0, cols: 0 }` means unfrozen. */
@@ -127,16 +128,69 @@ export function clearFrozenCell(cell: HTMLElement): void {
 }
 
 /**
- * Apply sticky styling to the first `spec.rows` rows and `spec.cols` columns of `table`. Offsets are
- * measured from the live DOM so they follow column resizes. A frozen row's cells stick to the top; a
- * frozen column's cells stick to the left; a corner cell (both) stacks above either single axis.
+ * Per-visual-column widths for a table, measured from the live DOM against the span-aware grid.
+ *
+ * A `colspan` cell owns several grid columns (one DOM cell, N visual columns) and a `rowspan` cell
+ * is absent from the `<tr>`s it covers, so a plain `row.cells[i].offsetWidth` is NOT the width of
+ * visual column `i`. We walk the ProseMirror rows/cells in lockstep with the DOM cells — they render
+ * in the same order — and size each column from a `colspan=1` cell wherever one exists, falling back
+ * to splitting a spanning cell's width across the columns it is the only source for.
  */
-export function applyFrozenStyles(table: HTMLTableElement, spec: FreezeSpec): void {
-  const rows = Array.from(table.rows)
-  if (!rows.length) return
-  const frozenRows = Math.max(0, Math.min(spec.rows, rows.length))
-  const colCount = rows[0].cells.length
-  const frozenCols = Math.max(0, Math.min(spec.cols, colCount))
+function measureColumnWidths(domRows: HTMLTableRowElement[], node: PMNode, map: TableMap): number[] {
+  const widths = new Array<number>(map.width).fill(NaN)
+  const spanning: { left: number; right: number; width: number }[] = []
+
+  node.forEach((rowNode, rowOffset, rowIdx) => {
+    const domRow = domRows[rowIdx]
+    if (!domRow) return
+    const domCells = domRow.cells
+    let d = 0
+    rowNode.forEach((_cellNode, cellOffset) => {
+      const cell = domCells[d++]
+      if (!cell) return
+      const rect = map.findCell(rowOffset + 1 + cellOffset)
+      if (rect.right - rect.left === 1) {
+        if (Number.isNaN(widths[rect.left])) widths[rect.left] = cell.offsetWidth
+      } else {
+        spanning.push({ left: rect.left, right: rect.right, width: cell.offsetWidth })
+      }
+    })
+  })
+
+  // Columns only ever covered by spanning cells: split the spanning cell's width across the columns
+  // it alone touches, so cumulative left offsets stay monotonic.
+  for (const s of spanning) {
+    const unknown: number[] = []
+    let known = 0
+    for (let c = s.left; c < s.right; c++) {
+      if (Number.isNaN(widths[c])) unknown.push(c)
+      else known += widths[c]
+    }
+    if (unknown.length) {
+      const each = Math.max(0, (s.width - known) / unknown.length)
+      for (const c of unknown) widths[c] = each
+    }
+  }
+
+  for (let c = 0; c < widths.length; c++) if (Number.isNaN(widths[c])) widths[c] = 0
+  return widths
+}
+
+/**
+ * Apply sticky styling to the first `spec.rows` rows and `spec.cols` columns of `table`. Frozen-band
+ * membership and the sticky offsets are derived from the table's `TableMap` — the same span-aware
+ * grid model the freeze commands use — NOT from raw DOM cell indices, so merged cells stay aligned:
+ * a `colspan` cell occupies its full column range and a `rowspan` cell only lives in the row it
+ * starts in (the rows it covers have no DOM cell of their own). Offsets are still measured from the
+ * live DOM so they follow column resizes. A frozen row's cells stick to the top; a frozen column's
+ * cells stick to the left; a corner cell (both) stacks above either single axis.
+ */
+export function applyFrozenStyles(table: HTMLTableElement, node: PMNode, spec: FreezeSpec): void {
+  const domRows = Array.from(table.rows)
+  if (!domRows.length) return
+  const map = TableMap.get(node)
+  const frozenRows = Math.max(0, Math.min(spec.rows, map.height))
+  const frozenCols = Math.max(0, Math.min(spec.cols, map.width))
   const anyFrozen = frozenRows > 0 || frozenCols > 0
 
   // A frozen table switches to separate borders and is no longer its own scroll container — both
@@ -150,28 +204,34 @@ export function applyFrozenStyles(table: HTMLTableElement, spec: FreezeSpec): vo
   const wrapper = table.closest('.tableWrapper')
   if (wrapper) wrapper.classList.toggle('octo-has-frozen-rows', frozenRows > 0)
 
-  const rowTops = cumulativeOffsets(rows.slice(0, frozenRows).map((r) => r.offsetHeight))
-  const colLefts = cumulativeOffsets(
-    Array.from(rows[0].cells)
-      .slice(0, frozenCols)
-      .map((c) => c.offsetWidth),
-  )
+  // Each <tr> is exactly one visual row, so DOM row index == grid row index — top offsets come
+  // straight from the row boxes. Column widths must respect spans, so measure against the grid.
+  const rowTops = cumulativeOffsets(domRows.slice(0, frozenRows).map((r) => r.offsetHeight))
+  const colWidths = measureColumnWidths(domRows, node, map)
+  const colLefts = cumulativeOffsets(colWidths.slice(0, frozenCols))
 
-  rows.forEach((row, rIdx) => {
-    Array.from(row.cells).forEach((cell, cIdx) => {
-      const inFrozenRow = rIdx < frozenRows
-      const inFrozenCol = cIdx < frozenCols
+  node.forEach((rowNode, rowOffset, rowIdx) => {
+    const domRow = domRows[rowIdx]
+    if (!domRow) return
+    const domCells = domRow.cells
+    let d = 0
+    rowNode.forEach((_cellNode, cellOffset) => {
+      const cell = domCells[d++]
+      if (!cell) return
+      const rect = map.findCell(rowOffset + 1 + cellOffset)
+      const inFrozenRow = rect.top < frozenRows
+      const inFrozenCol = rect.left < frozenCols
       if (!inFrozenRow && !inFrozenCol) return
       cell.setAttribute(FROZEN_ATTR, '')
       cell.classList.add('octo-frozen-cell')
       cell.style.position = 'sticky'
       if (inFrozenRow) {
         cell.classList.add('octo-frozen-row-cell')
-        cell.style.top = `${rowTops[rIdx]}px`
+        cell.style.top = `${rowTops[rect.top]}px`
       }
       if (inFrozenCol) {
         cell.classList.add('octo-frozen-col-cell')
-        cell.style.left = `${colLefts[cIdx]}px`
+        cell.style.left = `${colLefts[rect.left]}px`
       }
       cell.style.zIndex = inFrozenRow && inFrozenCol ? '4' : inFrozenRow ? '3' : '2'
     })
@@ -217,6 +277,8 @@ class FreezeStyleView {
     if (!map || map.size === 0) return
     map.forEach((spec, pos) => {
       if (spec.rows <= 0 && spec.cols <= 0) return
+      const node = this.view.state.doc.nodeAt(pos)
+      if (!node || node.type.name !== 'table') return
       let dom: Node | null = null
       try {
         dom = this.view.nodeDOM(pos)
@@ -227,7 +289,7 @@ class FreezeStyleView {
       const table =
         dom instanceof HTMLTableElement ? dom : dom.querySelector<HTMLTableElement>('table')
       if (!table) return
-      applyFrozenStyles(table, spec)
+      applyFrozenStyles(table, node, spec)
     })
   }
 }
--- a/packages/docs/src/editor/TableControls.test.tsx
+++ b/packages/docs/src/editor/TableControls.test.tsx
@@ -244,9 +244,10 @@ describe('TableContextMenu — right-click inside a cell opens the menu (XIN-105
     // Selection was moved into the right-clicked cell so position-relative commands act on it.
     expect(e.isActive('table')).toBe(true)
     // All seven table commands are present (add row before/after, delete row, add column
-    // before/after, delete column, delete table).
+    // before/after, delete column, delete table), plus the collapsed "Freeze" submenu trigger (#755)
+    // whose sub-options are hidden until it is expanded.
     const buttons = menu!.querySelectorAll('button.octo-tb-btn')
-    expect(buttons.length).toBe(7)
+    expect(buttons.length).toBe(8)
     e.destroy()
     host.remove()
   })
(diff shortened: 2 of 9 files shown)


OctoBoooot:
@/tmp/pr775-advisory-93a688f2.md

Jerry-Xin (commented):
**Review: clean (code), CI-pending / Draft — COMMENT (not a gating approve).**

Full scan of all 9 files at `93a688f2f1c0`. The implementation is well-contained and matches the PR's stated design. No blocking issues found. Holding at COMMENT rather than APPROVE for two governance reasons, not code reasons: (1) this is an explicit **Draft** ("do not merge"), and (2) the required `Build` check is `skipping`, so the code was never actually compiled/tested in CI on this head. Happy to flip to APPROVE once it's marked ready and Build reports green.

### ✅ What holds up (verified, not trusted from the description)
- **Collab / persistence safety.** Freeze is genuinely view-state only: the frozen set lives in a ProseMirror plugin `Map<tablePos,{rows,cols}>` (`packages/docs/src/editor/TableFreeze.ts`), nothing is written to the Y.Doc. Confirmed the styler only mutates the cell's own `style` / `class` / `data-octo-frozen` attribute, and `TableCellView.ignoreMutation` (`packages/docs/src/editor/TableCellView.ts:62`) returns true for `attributes` mutations on the cell DOM — so these writes never round-trip as document edits or fight remote cursors. There is no collab-sync surface and no hostile-peer value path (freeze is per-client, numeric).
- **Bounds checks.** Freeze counts are double-clamped: on write in `dispatchFreeze` (`Math.max(0, Math.min(round(x), rowCount/colCount))`, `packages/docs/src/editor/TableFreeze.ts:315`) and again at render in `applyFrozenStyles` (`Math.min(spec.rows, rows.length)`, `packages/docs/src/editor/TableFreeze.ts:134`). A stale count after row/column delete cannot go out of range because the render re-clamps every update. Empty table early-returns.
- **DOM safety.** No `innerHTML` / `dangerouslySetInnerHTML`. Labels flow through `t()` into React JSX title/text (auto-escaped); icons are static SVG; inline styles are numeric px. No XSS vector in the ribbon entry or submenu.
- **State machine.** Plugin remaps positions on `docChanged` and drops entries for deleted tables (test-covered). Export round-trip: freeze is intentionally NOT in the schema, and `buildPreviewExtensions` omits `TableFreeze`, so it is silently dropped on export — a documented trade-off, not content data-loss.
- **i18n.** en-US and zh-CN are symmetric (`toolbar.freeze` + 6 `table.*` keys added to both).

### 💬 Non-blocking notes
- 🟡 **Colspan/rowspan divergence.** Commands compute row/column indices via `TableMap` (span-aware), but `applyFrozenStyles` indexes raw DOM `row.cells[cIdx]` directly (`packages/docs/src/editor/TableFreeze.ts:134`). In a table with merged cells the visual column a cell occupies differs from its DOM cell index, so freezing columns on a merged-cell table can pin the wrong visual columns / compute wrong sticky offsets. Correctness-only (no crash/security); worth a follow-up or a note that freeze targets simple tables.
- 🟡 **Context-menu clamp vs submenu expand.** The menu position clamp runs only when the anchor `point` changes (`packages/docs/src/editor/TableControls.tsx`), so expanding the Freeze accordion near the bottom edge can push sub-options past the viewport. Consider reclamping after `freezeOpen` toggles.
- 🔵 **Ribbon control test gap.** `TableFreezeControl` (`packages/docs/src/editor/TableControls.tsx`) has no direct test for enablement/active state/popover actions; current suite covers command logic + the context-menu submenu. A focused test would lock in the ribbon path.

### CI / gate
- Code checks: secret-scan ✅, osv-scan ✅, dependency-review ✅. **Build = skipping** (path filter — never actually ran on this head). No Test/Vet/Lint jobs registered.
- Governance: check-sprint / satisfy-code-review / scan-scheduled skipping; code-review pending — none bar a code verdict.
- `isDraft: true`; `reviewDecision: REVIEW_REQUIRED`; `mergeStateStatus: BLOCKED`; label `size/XL`. Gate held by: draft status + required review + Build not green.

boris-clark (approved):
## Review — APPROVED

Reviewed the full diff line-by-line and verified the two flagged risk areas against the `feat/docs-table-freeze-panes` head branch (based cleanly on `main` @ 578dac69, no rebase drift). Tests and typecheck run locally. No blockers.

### Verification

- **45 tests pass locally** — `TableFreeze.test.ts` + `TableControls.test.tsx` (41) and `i18n.test.ts` (4). Matches the PR claim.
- **Typecheck clean for every changed file.** `pnpm --filter docs typecheck` reports only the 2 pre-existing errors in unrelated files (`../dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`) — neither is touched by this PR.
- **No AI-authorship markers** in commit trailers, message, or diff (checked `Co-Authored-By` / "Generated by" / Claude / Anthropic / Copilot). Clean.

### 1. Submenu "occasionally doesn't pop up" bug — really fixed, not masked ✅

Confirmed against `main`: the table context menu there is a **flat 7-button menu with no submenu at all**, so there was no pre-existing flaky flyout to reproduce. The PR body states this honestly rather than claiming a phantom fix. The new submenu is reliable **by construction**, not by a `try/catch` cover-up:

- `TableControls.tsx:209` — `freezeOpen` is plain React state on the menu, toggled by `onClick`. No hover flyout, no portal, no capture-phase listener → no diagonal-gap / pointer-traversal race.
- `TableControls.tsx:234` — the outside-close handler is `if (menuRef.current && !menuRef.current.contains(e.target)) setPoint(null)`. The submenu (`TableControls.tsx:247`) renders **inside** `menuRef`, so the `mousedown` that immediately follows the trigger click can never race it closed. The `stays open across the document mousedown` regression test locks this in.

Root cause (a racy open mechanism) is designed out, not patched over.

### 2. #749 stacking / resize-handle interaction — no regression ✅

`styles.css:2010` lifts `.octo-prose table .column-resize-handle` to `z-index: 6`. Verified the ordering is sound: frozen cells are capped at `z-index: 4` (corner), `3` (row), `2` (col) — all set inline in `applyFrozenStyles` (`TableFreeze.ts:852`). The handle is a sibling of the cells in the table's stacking context (the only model under which sticky cells could occlude it, which is exactly the premise the fix targets), so `6 > 4` reliably keeps it clickable over a frozen column. Applying the rule to non-frozen tables too is benign — the handle was always intended to paint on top, and nothing inside a normal table cell carries a competing `z-index`. No new occlusion of the `#749` hit area, and no `mousedown` is swallowed.

### 3. View-state vs. Yjs ✅

Freeze lives entirely in a ProseMirror plugin keyed by table position (`TableFreeze.ts:915`), remapped across edits and dropped when the table is deleted. Nothing is written to the `Y.Doc`; `SCHEMA_VERSION` and the backend contract are untouched, so there is no collab-sync surface to conflict on. The per-client / not-persisted trade-off is documented in both the PR body and the source header. Cell-style writes are safe because `TableCellView.ignoreMutation` already discards attribute/style mutations — good call-out.

### 4. Ribbon entry ✅

`TableFreezeControl` (`TableControls.tsx:301`) is wired next to `TableGridPicker` in the toolbar (`Toolbar.tsx:1533`), disabled unless the caret is in a table, active while any band is frozen, and reuses the shared `octo-color-control` popover (outside-click / Escape close). Consistent with the existing toolbar dropdowns.

### Non-blocking notes (follow-up, not required for this PR)

- **Merged cells (rowspan/colspan).** `applyFrozenStyles` indexes `row.cells` / `rows[0].cells` directly, so for tables with spans the DOM cell index no longer matches the visual grid column and sticky offsets / band boundaries could be off. The uniform-grid common case is correct; worth a guard or a documented limitation in a later change.
- **Render cost.** `FreezeStyleView.render()` runs three full-editor `querySelectorAll` resets on every transaction even when no table is frozen. Negligible in practice, but an early return when the map is empty would save the work on every keystroke.
- **Global handle z-index.** `z-index: 6` on the handle now sits above the gutter drag/insert handles (`z-index: 4`); they don't overlap spatially so this is fine, just noting it's applied table-wide rather than scoped to frozen tables.

Ship-wise this is solid. Keeping it as a draft per the PR's own note — not marking ready, not merging.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #775 (octo-web)

## Summary

This PR adds freeze panes (冻结行列) to the Docs editor table: a ProseMirror plugin that tracks frozen row/column counts as view-state (never written to the Y.Doc), a DOM-rendering plugin view that applies `position: sticky` to the frozen bands with live-measured offsets, a ribbon popover (`TableFreezeControl`), and a right-click context-menu submenu. The design is thoughtful — freeze is intentionally per-client and non-persistent, avoiding any schema version bump or collaborative sync surface. The submenu reliability concern is addressed by construction (click-toggled accordion, not hover flyout). 45 unit tests cover commands, plugin remapping, submenu behavior, and sticky-style DOM output. Overall a clean, well-documented addition.

## Verification

Static analysis only at head `93a688f2`; build and tests not executed in this environment.

- ✅ **Plugin state lifecycle** — `tableFreezePlugin().state.apply` correctly remaps `tablePos` keys across document edits via `tr.mapping.map(pos, -1)`, verifies the node at the mapped position is still a `table`, and drops stale entries. Meta-driven updates properly delete entries when `spec.rows <= 0 && spec.cols <= 0`.
- ✅ **DOM mutation safety under collaboration** — `TableCellView.ignoreMutation` at `:62` returns `true` for `type === 'attributes'` when `target === this.dom`, which covers the freeze plugin's `setAttribute`, `classList.add`, and `style.*` writes on cell elements. These view-only DOM mutations never round-trip as document edits.
- ✅ **Command contract** — `dispatchFreeze` at `TableFreeze.ts:298` correctly resolves `cellContext`, reads current spec, applies the compute callback, clamps to `[0, rowCount]` / `[0, colCount]`, and dispatches a single meta-bearing transaction. All six commands (set, toggleHeaderRow, toggleFirstColumn, freezeThroughRow, freezeThroughColumn, clear) flow through this single dispatch path.
- ✅ **Freeze/unfreeze parity (C1)** — Both directions are implemented and tested. `clearTableFreeze` dispatches `{ rows: 0, cols: 0 }`, which the plugin's `apply` converts to a `Map.delete`. The UI disables the Unfreeze button when `freeze.rows === 0 && freeze.cols === 0`.
- ✅ **Position remapping across edits** — Tested at `TableFreeze.test.ts:182`: inserting content before the table shifts the table's position, and the freeze entry survives the remap. Deletion drops the entry (`:197`).
- ✅ **i18n parity** — Both `en-US.json` and `zh-CN.json` add the same 7 keys (`docs.toolbar.freeze` + 6 `docs.table.*` keys). No missing translations.
- ✅ **Extension wiring** — `TableFreeze` is registered in `extensions.ts:249` for the live editor and correctly excluded from the static/export editor at `:340`.
- ✅ **Submenu reliability** — The accordion is click-toggled React state on the menu component. The outside-click close handler (`:277`) checks `menuRef.current.contains(e.target)`, so clicks inside the submenu (which is a child of the menu) do not close the menu. The `mousedown` preventDefault on `TbBtn` (`:138`) prevents focus shift. Tested at `TableControls.test.tsx:543`.

## Findings

### P2 — Context menu position not re-clamped when freeze submenu expands (`packages/docs/src/editor/TableControls.tsx:293`)

The `useLayoutEffect` that calls `clampMenuPosition` depends only on `[point]`, so it runs once when the menu opens. When the freeze submenu toggles open (adding ~150px of vertical content), the menu grows but its position is not re-evaluated. If the user right-clicked near the bottom of the viewport, the expanded menu can overflow below the viewport fold. Since the menu uses `position: fixed`, it does not scroll with the page, making the submenu items unreachable.

Fix direction: add `freezeOpen` to the dependency array so the effect re-runs when the submenu expands, or add `max-height` + `overflow-y: auto` to `.octo-table-context-menu`.

### Nit — Resize handle z-index scoped too broadly (`packages/docs/src/editor/styles.css:2013`)

`.octo-prose table .column-resize-handle { z-index: 6 }` applies to every table in the prose editor, not just frozen ones. Harmless because non-frozen tables have no z-indexed cells to compete with, but the rule is only needed when frozen cells create stacking contexts. Scoping to `.octo-prose table.octo-frozen-table .column-resize-handle` would be more precise.

- **`FreezeStyleView.render()` reset-then-apply pattern** — Every update clears all frozen attributes/classes/styles from the DOM, then reapplies for the current freeze state. This is idempotent and self-healing: if ProseMirror re-renders a table (stripping the `octo-frozen-table` class), the next `render()` call restores it.
- **`nodeDOM(pos)` fallback** — When the stored position points to a non-table wrapper, the code falls back to `dom.querySelector('table')`. The plugin's `apply` function already guarantees the position maps to a `table` node, but the fallback adds defense in depth.
- **`cumulativeOffsets`** — Pure prefix-sum helper, correctly handles empty input. Tested at `TableFreeze.test.ts:102`.
- **`cellContext` error handling** — `selectionCell` is wrapped in try/catch; returns `null` when the caret is outside any table. All callers handle the `null` path.
- **`TableFreezeControl` re-render strategy** — Bumps force-state on every `transaction` and `selectionUpdate` event. Standard tiptap React pattern; component is small enough that per-keystroke re-renders are negligible.
- **CSS border treatment** — The `border-collapse: separate; border-spacing: 0` switch on `.octo-frozen-table` is necessary for Chromium's `position: sticky` to work on `<td>/<th>`. The border redraw (table paints top+left, cells paint right+bottom) produces visually identical 1px borders. The `.octo-frozen-table` class is only present while a table is frozen, so normal tables are unaffected.

## Data-flow trace

| Consumer | Upstream source | Verified? |
|---|---|---|
| `getFreezeSpec(state)` → ribbon/menu active state | `tableFreezeKey.getState(state).get(tablePos)` → plugin `apply` meta/remap | ✅ Meta dispatch sets entries; remap preserves across edits; `Map.delete` on clear |
| `applyFrozenStyles(table, spec)` → sticky DOM | `FreezeStyleView.render()` → `nodeDOM(pos)` → `HTMLTableElement` | ✅ Position guaranteed to be a table by plugin `apply` check |
| `isInTable(state)` → ribbon button `disabled` | `cellContext(state)` → `selectionCell(state)` | ✅ try/catch handles non-table selections; force-bump re-reads on every transaction |
| `freezeOpen` → submenu visibility | React `useState(false)` on `TableContextMenu` | ✅ Reset on `contextmenu` event, on command `run()`, and on Escape (via `setPoint(null)` unmounting the menu) |
| `clampMenuPosition(point, rect, viewport)` → menu position | `useLayoutEffect([point])` → `menuRef.getBoundingClientRect()` | ⚠️ Does not re-run on `freezeOpen` change (P2 above) |

## Blindspot checklist (C1–C6)

- **C1 — Dual-path parity**: Freeze/unfreeze paths both implemented, tested, and wired through `dispatchFreeze`. Clear.
- **C2 — Control-flow ordering**: Single dispatch path for all freeze commands; no nesting or re-entrancy. Clear.
- **C3 — Authorization boundary**: Frontend-only view-state. No auth changes. N/A.
- **C5 — Build through ≠ runtime correct**: CSS changes (border-collapse, overflow, sticky) are runtime-verified per the PR description (Chromium browser testing). The dev harness issue is noted but doesn't affect the review. Clear.
- **C6 — Governance/strategy/security docs**: N/A.

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. The plugin lifecycle, DOM mutation safety, command contract, position remapping, and submenu reliability are all sound. One P2 (context menu re-clamping on submenu expand) and one Nit (CSS scoping) are non-blocking.

yujiawei (dismissed):
# Code Review — PR #775 (octo-web)

**Scope reviewed:** head `93a688f2` vs merge-base `578dac69` (the PR's parent — clean base, no unrelated drift). 9 files, +1034/−8. Feature: freeze panes (冻结窗格) for editor tables via a ribbon button + right-click submenu, implemented as a view-state ProseMirror plugin (`position: sticky`, nothing written to the Y.Doc).

## 1. Spec compliance

Traced against issue #755 (the linked spec):

- **Ribbon entry** — `TableFreezeControl` added to the toolbar (`Toolbar.tsx:1533`), disabled unless the caret is in a table, active while any band is frozen. ✅
- **Reliable submenu** — the "submenu occasionally doesn't pop up" bug is addressed by construction: the submenu is a click-toggled React accordion (`freezeOpen` state on the menu component), not a hover flyout, and the outside-click handler ignores clicks inside the menu (`TableControls.tsx:277-289`). A regression test locks in that a `mousedown` following the trigger click leaves it open. ✅
- **Freeze N rows / N columns, stay pinned on scroll** — commands + sticky renderer implement header-row, first-column, through-row, through-column, and unfreeze. ✅
- **Works with column resize (#749)** — resize handle lifted above frozen cells (`styles.css:2010`). ✅

No over-build: freeze is view-state only, `SCHEMA_VERSION` untouched, no Y.Doc/backend surface. The "not persisted across reload / not shared with collaborators" trade-off is explicitly documented in the PR and left as a follow-up — a reasonable scope boundary, not a defect.

**Spec: ✅**

## 2. Code quality

Verification performed locally: all 45 new/changed tests pass (`TableFreeze.test.ts`, `TableControls.test.tsx`, `i18n.test.ts`); `tsc --noEmit` is clean for every file in this change (the remaining tsc errors are pre-existing in unrelated files — `dmworkbase` react@17 typings, `EditorShell.test.tsx`, `members/sort.ts` — matching the PR's claims). I also verified the collab-safety premise: both the Tiptap `TableView` and the project's `TableCellView.ignoreMutation` discard `attributes`/`childList` mutations on the table/wrapper/cell chrome, so the plugin's inline style + class writes do not round-trip as document edits.

No P0/P1 issues found. The findings below are all **P2 (non-blocking)**.

### P2-1 — Freeze entry is dropped when a block is inserted exactly at the pre-table boundary
`TableFreeze.ts:249` remaps the freeze key with bias `-1`:
```ts
const mappedPos = tr.mapping.map(pos, -1)
const node = tr.doc.nodeAt(mappedPos)
if (node && node.type.name === 'table') mapped.set(mappedPos, spec)
```
The key is the position immediately *before* the table. With bias `-1`, an insertion at exactly that boundary maps the key to *before* the inserted content, so `nodeAt` no longer resolves to the table and the entry is silently dropped. I confirmed this empirically: freezing the header row, then inserting a paragraph at the exact pre-table boundary (or at doc position 0 when the table is the first node) drops the freeze (map size 1→0). Switching the bias to `+1` makes the entry survive both cases (size stays 1) while all existing remap tests still pass. Impact is low — normal typing inside the table or in a paragraph that already sits before the table remaps correctly; only an insertion landing on the exact node boundary triggers it, and the user can simply re-freeze. Suggest bias `+1` (associate the key with content after the insertion) plus a regression test for the boundary case.

### P2-2 — Merged cells (colspan/rowspan) can misalign the frozen band
`applyFrozenStyles` walks DOM cells by index (`rows[0].cells.length` as the column count, `TableFreeze.ts:138`; per-row `row.cells[cIdx]`). With merged cells the DOM-cell index is not the visual grid column, and a `rowspan` cell that starts in a frozen row but extends into scrollable rows gets `position: sticky` applied to the whole tall cell. There is currently **no in-editor merge-cell command wired** (no `mergeCells`/`splitCell` in the UI), so merged tables can only arrive via paste/HTML import — making this reachable but uncommon, and the failure mode is cosmetic (a misaligned/oversized frozen band), not data loss. Worth a follow-up if cell merging is ever surfaced; a `TableMap`-based band computation would make it correct.

### P2-3 — `.column-resize-handle` z-index bump is global, not scoped to frozen tables
`styles.css:2010` raises the resize handle to `z-index: 6` for **all** tables, not only `.octo-frozen-table`. It's benign today (it only needs to clear the ≤4 frozen-cell stack), but scoping it to `.octo-frozen-table .column-resize-handle` would keep the blast radius on non-frozen tables at zero.

### P2-4 — `FreezeStyleView.render()` resets the whole editor DOM on every transaction
`render()` (`TableFreeze.ts:196`) unconditionally calls `reset()` — three `querySelectorAll` sweeps over the full editor DOM — *before* the `map.size === 0` early-return. So every keystroke in a document with no frozen table still pays three whole-doc DOM scans. Cheap per call, but it's on the hot edit path; an early `if (map.size === 0 && no-frozen-markers-present) return` or tracking touched cells would avoid the sweep in the common (nothing-frozen) case.

### P2-5 — "Freeze header row" toggle unfreezes multi-row bands (minor UX)
`toggleFreezeHeaderRow` uses `current.rows >= 1 ? 0 : 1` (`TableFreeze.ts:333`). If the user froze 3 rows via "freeze up to this row", clicking "Freeze header row" collapses all frozen rows to 0 rather than to 1. Defensible (the button reads active, and clicking an active control to clear is reasonable), but "freeze header row → unfreeze 3 rows" may surprise. Consider setting to exactly 1 when currently >1.

**Quality: Approved**

## 3. Overall verdict

**APPROVED.** Spec fully met with no over-build; implementation is well-tested, collab-safe, and typechecks clean. All findings are P2 polish/edge-case items — none blocks merge. P2-1 (boundary-insert freeze drop) is the most concrete and is a one-line bias change worth a fast follow-up.

## 4. Suggestions
- P2-1: change the remap bias to `+1` and add a boundary-insertion regression test.
- P2-3: scope the resize-handle z-index rule to `.octo-frozen-table`.
- P2-4: early-return in `render()` before the DOM reset when nothing is frozen.

## 5. Coverage note
- Pixel-level sticky behavior (the actual pinning, the `border-collapse: separate` redraw, the 60vh scroll box) is not exercisable in jsdom and was verified by the author in Chromium per the PR notes; I reviewed the CSS/logic but did not re-run a browser check.
- Merged-cell tables (P2-2) were reasoned about, not runtime-tested, since no in-editor merge path exists to produce one.
- This review reflects an independent pass; two secondary automated advisory checks were attempted for this complex PR but both were unavailable this run (one gateway error, one timeout), so they contributed no findings and are recorded as absent rather than as agreement.

yujiawei:
Follow-up to the review above — one additional **P2 (non-blocking)** surfaced by a second pass, which I confirmed against the code:

**Context menu isn't re-clamped when the Freeze submenu expands** (`TableControls.tsx:295-304`). The viewport-clamp `useLayoutEffect` depends only on `[point]`, so it runs once when the menu opens. Toggling the Freeze accordion adds ~5 rows of height afterward, but the menu's position is not re-evaluated. Because `.octo-table-context-menu` is `position: fixed; z-index: 1000` with no `max-height`/`overflow-y` (`styles.css:479`), a right-click near the bottom of the viewport followed by expanding Freeze can push the sub-options below the fold, where a fixed menu can't be scrolled to.

Fix direction: add `freezeOpen` to the effect's dependency array so it re-clamps on expand, or give `.octo-table-context-menu` a `max-height` + `overflow-y: auto`.

Reachability is narrow (bottom-edge right-click on a tall menu) and the failure is cosmetic, so this does not change the approval — noting it for a fast follow-up alongside the other P2s.

packages/docs/src/editor/TableFreeze.ts
```diff
@@ -0,0 +1,363 @@
...
+    Array.from(rows[0].cells)
+      .slice(0, frozenCols)
+      .map((c) => c.offsetWidth),
+  )
+
+  rows.forEach((row, rIdx) => {
+    Array.from(row.cells).forEach((cell, cIdx) => {
+      const inFrozenRow = rIdx < frozenRows
+      const inFrozenCol = cIdx < frozenCols
+      if (!inFrozenRow && !inFrozenCol) return
```
OctoBoooot, line 224: **[major]** `applyFrozenStyles` freezes by DOM cell index, which diverges from the visual grid on merged cells. `inFrozenCol = cIdx < frozenCols` and `colLefts[cIdx]` (:175) use the raw `Array.from(row.cells)` index — but a colspan=N cell is one `row.cells` entry spanning N visual columns, so from that cell on, `cIdx` no longer maps to the true column and the sticky `left` offset shifts. Symmetric rowspan bug: a cell spanning rows appears only in its starting <tr>, so cells covered by a rowspan from above are absent from the later row's `row.cells` and never get frozen → gap in the band. The command layer already uses span-aware `TableMap`; this render layer should too. Failing input: freeze a header row / first column on a table with a colspan=2 cell → frozen and scrolling cells misalign from that cell onward. Tests (TableFreeze.test.ts:200+) only cover uniform grids. Fix: derive band membership + offsets from TableMap, and add a merged-cell test.

OctoBoooot (changes requested):
## Review: freeze panes for editor tables (#775)

**Verdict:** Request changes — one real correctness bug (merged-cell freezing mis-targets, both axes) and code CI (`Build`) not yet green. The architecture is genuinely good; this is a solid feature with one render-layer gap.

Risk: `size/XL` → `needs-human-review` applied (mandatory for a change this size). CI: `Build` is still `in_progress` (not concluded green), so APPROVE isn't available yet regardless of findings.

### Major
- `packages/docs/src/editor/TableFreeze.ts:160-176` — **`applyFrozenStyles` freezes by DOM cell index, which diverges from the visual grid when a table has merged cells.** The command layer uses span-aware `TableMap`, but this render layer walks `Array.from(row.cells)` and uses the DOM index `cIdx` / row index `rIdx` directly:
  - **colspan:** a cell spanning N columns is one `row.cells` entry but N visual columns, so `cIdx < frozenCols` and `colLefts[cIdx]` mis-identify which cells are in the frozen band and give the wrong sticky `left` offset.
  - **rowspan:** a cell spanning rows appears only in its *starting* `<tr>`, so a cell covered by a rowspan from above is absent from the later row's `row.cells` — the loop never marks it sticky, leaving a gap in the frozen band.
  Concrete failing input: a table whose header row (or first column) contains a `colspan=2` cell, frozen via "freeze header row / first column" → the sticky offsets shift by one column from that cell onward, so frozen and scrolling cells visually overlap/misalign. `applyFrozenStyles`'s tests (`TableFreeze.test.ts:200+`) only exercise uniform grids (rows:2/cols:0, rows:1/cols:1) — no colspan/rowspan case, so this isn't caught. Fix: derive the frozen-band membership and offsets from `TableMap` (the same grid model the commands already use) rather than raw DOM cell indices, and add a merged-cell test.

### CI
- `Build` has not concluded green at this head (`in_progress`) — the added tests haven't run in CI, and Build carries type-check/build. Verdict can't be APPROVE until it's green; if Build surfaces anything it's a separate blocker.

### Verified sound (from the architecture read)
- **View-state only, no collab/schema surface** — freeze lives in a ProseMirror plugin state (`tableFreezeKey`), position-remapped on `docChanged`, nothing written to the Y.Doc. No `SCHEMA_VERSION` bump, no localStorage, no `innerHTML`. Correctly per-viewer (one user's frozen panes don't force another's).
- **Deleted-table cleanup** — `apply()` remaps each frozen table's pos via `tr.mapping.map(pos,-1)` and re-adds only if `nodeAt` is still a `table`; drops the entry otherwise. No stale entry / invalid-position crash.
- **#761 context-menu hardening preserved** — `isEditable` gate and `CellSelection` preservation intact (the freeze submenu builds on the hardened base, doesn't regress it).
- **Plugin-view teardown** — `FreezeStyleView` clears sticky styles + classes on `update()`/`destroy()`; no leaked inline styles on unmount.

### Praise
- Keeping freeze as pure view-state (no Y.Doc write) is exactly right — freeze is a per-viewer concern and this avoids a whole class of collab-sync conflicts. The border-collapse→separate switch with the CSS-class comment explaining *why* sticky needs it is the kind of note that saves the next reader an hour.
- The click-toggled inline submenu (not a CSS-hover flyout) sidesteps the classic diagonal-gap/pointer-traversal flake, and there's a regression test locking it. Deliberate, not accidental.

### Before Ready / to confirm
- **Sticky-offset re-sync after column resize** — offsets are measured from the live DOM in `applyFrozenStyles`, and `FreezeStyleView.update()` re-applies on editor updates. Confirm a pure column-resize that doesn't fire `docChanged` still triggers a re-measure (else the frozen band drifts from the resized columns). Worth a manual check or a resize test.

### Net
Well-architected freeze feature — view-state only, #761 guards preserved, correct deleted-table cleanup and teardown. The one real defect is the render layer freezing by DOM index instead of the span-aware grid, which mis-targets on merged cells (both colspan and rowspan) and isn't tested. Fix that against `TableMap` + add a merged-cell test, and let `Build` go green. `needs-human-review` applied for the XL size. Credit @ploy-elison — the view-state decision, the #761-guard preservation, and the click-toggle submenu are all the right calls.

Jerry-Xin (dismissed):
Now that #775 is marked Ready and code CI is green, I'm converting my earlier COMMENT to a formal verdict. The freeze feature is architecturally sound, but there is one real correctness bug on merged-cell tables — REQUEST_CHANGES, converging with the peer review.

## 🔴 Blocking — freeze mis-targets cells on tables with merged cells (colspan/rowspan)

`packages/docs/src/editor/TableFreeze.ts` — `applyFrozenStyles`. The per-cell loop iterates the DOM cell list (`Array.from(row.cells).forEach((cell, cIdx) => ...)`) and uses the DOM cell position `cIdx` both to decide membership (`inFrozenCol = cIdx < frozenCols`) and to index the offset array (`colLefts[cIdx]`). Likewise the column-width basis is taken from `rows[0].cells` positionally.

For plain tables this is correct, but the DOM cell index diverges from the logical grid column whenever merged cells are present:
- A cell with `colspan=2` occupies one DOM cell but two logical columns, so every cell after it in that row has a `cIdx` that is one (or more) less than its true grid column.
- Rows crossed by a `rowspan` cell have fewer DOM cells (the spanned cell isn't repeated), shifting `cIdx` for the remaining cells.

Result: on a table that uses merged cells, freezing the first N columns freezes the wrong cells / applies wrong sticky offsets — a visibly broken freeze, not just a cosmetic nudge. The editor uses the tiptap table extensions which support colspan/rowspan, so this path is reachable by a normal user (merge cells, then freeze).

Fix: compute membership and offsets against the logical grid column, not the DOM cell index — e.g. derive each cell's starting grid column from a `TableMap` (prosemirror-tables) or by accumulating `colSpan` while walking the row, and freeze/offset by that logical column. The row axis needs the analogous rowspan-aware handling.

## ✅ What is solid (unchanged from my earlier read)
- Freeze state is **view-state only** — it is not written into the shared Y.Doc, so there is no collaborative-sync/stored-state hazard and no hostile-peer path. Re-measured on every update (correct after resize / row insert-delete / remote edits).
- No unsafe DOM: styling is via `setAttribute` / `classList` / `style.*`, no `innerHTML`/`dangerouslySetInnerHTML`.
- Frozen-count bounds are clamped at write and at render; export drops the freeze (documented trade-off); i18n symmetric; ribbon + right-click submenu integration is clean and the earlier submenu-reposition fix holds.

The architecture is genuinely good — the only blocker is the merged-cell index divergence in `applyFrozenStyles`. Fix that (logical-grid-column mapping) and this is an APPROVE. Credit to the peer review for pinning the severity on the merged-cell case.

ploy-elison:
@OctoBoooot Thanks for the catch — you were right, `applyFrozenStyles` was reasoning about the frozen bands in DOM-cell space, which diverges from the visual grid the moment cells are merged. Fixed in `ee85088`.

**What changed**

Frozen-band membership and the sticky `top`/`left` offsets are now derived from the table's `TableMap` — the same span-aware grid model the freeze *commands* already use — instead of `Array.from(row.cells)` indices. For every DOM cell I look up its grid rect and gate on `rect.top < frozenRows` / `rect.left < frozenCols`, and offset with `rowTops[rect.top]` / `colLefts[rect.left]`.

- **colspan**: a spanning cell now occupies its full column range, so the cells that follow it in the DOM are no longer off by a column. Column widths are measured per *visual* column against the grid (a `colspan=N` cell no longer counts as one column width).
- **rowspan**: the rect comes from the grid, so a DOM cell that sits at index 0 in a row whose real column 0 is covered by a rowspan from above is correctly identified as column 1 and left unfrozen — no more spurious freezes / gaps.

Row heights still come straight from each `<tr>` (one `<tr>` == one grid row), and offsets are still measured from the live DOM so they keep following column resizes.

**Tests**

Added two merged-cell cases to `TableFreeze.test.ts`, both rendered through a real editor so the DOM and the node's `TableMap` line up as they do in the plugin:
- colspan header + freeze first two columns — the lone col-2 header cell must not be frozen (it was, before);
- rowspan first column + freeze first column — the col-1 cell in the row the rowspan skips must not be frozen (it was, before).

Both fail on the old code and pass on the fix; the existing uniform-grid cases are unchanged. Local `pnpm build` is green and the CI **Build** check passed on `ee85088`.

Scope is limited to the merged-cell alignment in `applyFrozenStyles` (+ the one call site passing the node); the view-state model, commands, plugin remap, and delete cleanup you signed off on are untouched.

Could you take another look when you have a minute?

yujiawei (approved):
# Code Review — PR #775 (octo-web)

Reviewed at head SHA `ee8508873293024b3580c2b51dc3a7106bc42276` (the branch advanced past the
originally-requested `93a688f2…` while this review was in progress; the newer commit
`fix(docs): derive table freeze bands from TableMap so merged cells stay aligned` is included below).

Method: full read of the diff and surrounding table code, local run of the docs test suite and
typecheck, empirical probes of two suspected state bugs, and two independent second-opinion passes
(one line-level, one long-context). Verified locally: **43 freeze unit tests pass**; typecheck is
clean for all changed files (the 2 remaining `tsc` errors are pre-existing and in unrelated files);
the only failing tests in the package are the 4 pre-existing `EditorShell.test.tsx` failures already
present on `main`.

## 1. Spec compliance

**Spec: ✅**

Issue #755 asked for three things; all are delivered:
1. **Ribbon entry** — `TableFreezeControl` is mounted in the toolbar (`Toolbar.tsx:1533`), not only in
   the right-click menu. Disabled outside a table, active while the current table has a frozen band.
2. **Reliable submenu** — the freeze submenu is a click-toggled accordion (state on the menu, no
   hover flyout / capture-phase listener), so it cannot be raced closed by the pointer interaction
   that opens it.
3. **Freeze top-N rows / first-N columns, fixed on scroll** — implemented via `position: sticky`
   applied by a ProseMirror plugin, reachable from both entry points.

No under-build and no meaningful over-build: the "freeze up to this row/column" commands are extra
modes but they directly serve "freeze top N rows / first N cols" (N derived from the caret), so they
are in scope. All 8 new i18n keys exist in both `en-US.json` and `zh-CN.json` — no raw-key leakage.

The deliberate scope decision that freeze is **view-state only** (a `Map<tablePos,{rows,cols}>` in a
plugin, nothing written to the Y.Doc/schema) is sound and well-documented: it keeps `SCHEMA_VERSION`
and the collab contract untouched, and the inline cell-style writes are safe because
`TableCellView.ignoreMutation` already discards attribute mutations on the cell element so they never
round-trip as edits. The trade-off (per-client, not persisted across reload) is called out in the PR
and code. Making it durable/shared is correctly left as a follow-up schema bump.

## 2. Code quality

**Quality: Approved**

No P0/P1 issues. The merged-cell correctness problem that existed on the original head has already
been fixed on the current head. The remaining items are all P2 / nits — none block merge.

### Fixed since the original head (verified)
- **Merged-cell alignment.** The original `applyFrozenStyles` indexed cells by raw DOM order
  (`row.cells[cIdx]`) while the commands compute the band from `TableMap` grid coordinates. Under a
  `colspan`/`rowspan` cell (reachable via paste/import — `TableCellView` renders spans) these two
  coordinate spaces diverged, so the wrong cells could stick, a trailing column could silently fail
  to freeze, and sticky-left offsets could misalign. The new commit rewrites the styler to derive
  band membership and offsets from `TableMap` (`measureColumnWidths` + `map.findCell`), matching the
  command layer, and adds targeted colspan/rowspan regression tests. Confirmed resolved.

### P2 — worth a follow-up, not blocking

- **P2 — Freeze silently lost when content is inserted at a table's exact leading boundary.**
  `TableFreeze.ts:311` remaps the freeze key with association bias `-1`
  (`tr.mapping.map(pos, -1)`). The key is `$cell.before(-1)` — the position that should *track the
  table that starts there* — so the correct bias is `+1`. With `-1`, an insertion at exactly that
  boundary (e.g. a block pasted/gap-cursored into the seam before a **top-of-document** table) leaves
  the mapped key pointing at the inserted content; `nodeAt(mappedPos)` is then not a `table`, the
  guard drops the entry, and the freeze visibly clears even though the table still exists. I
  reproduced this locally: a top-of-doc table's freeze `Map` goes from size 1 to 0 after an insert at
  the leading boundary; switching to `map(pos, 1)` fixes it and all existing tests still pass. Impact
  is limited (a per-client view-state preference clears; one click to reapply; no data loss) and the
  trigger is narrow (typing into a *preceding* paragraph inserts before the boundary and maps fine
  under either bias — which is why the current remap test doesn't catch it). Suggest `+1` plus a
  boundary-insert test.

- **P2 — Style recompute runs on every transaction even when nothing is frozen.**
  `FreezeStyleView.render()` (`TableFreeze.ts:274`) calls `reset()` unconditionally before checking
  whether the freeze `Map` is empty, and `update()` runs `render()` on every transaction. In the
  common case (no table ever frozen) this still does three whole-editor `querySelectorAll` scans per
  keystroke/selection change. It's cheap-ish (static queries, no forced reflow, no style writes when
  nothing matches), but it's avoidable: early-return when the map is empty *and* the DOM has no
  `[data-octo-frozen]` elements to clear. When a table *is* frozen, the write-then-measure sequence
  (`reset()` clears styles, then `offsetWidth/offsetHeight` reads) is a read-after-write that can
  cause layout work proportional to the frozen band on each update; batching reads before writes is a
  reasonable optimization if it shows up in profiling.

- **P2 — Implicit "freeze meta only rides a doc-unchanged transaction" invariant is unenforced.**
  `apply()` remaps existing entries, then applies `meta` using `meta.pos` verbatim. `meta.pos` is a
  pre-transaction coordinate; this is only correct because the sole meta writer (`dispatchFreeze`)
  dispatches a pure `setMeta` with no doc change. Nothing defends against a future caller batching the
  freeze meta onto a doc-changing transaction, which would write a pre-map coordinate into the
  post-map map. Latent only (not reachable today) — worth a one-line comment or a `docChanged`
  guard/`assert`.

- **P2 — `toggleFreezeHeaderRow` / `toggleFreezeFirstColumn` active-state is misleading after a
  multi-row/column freeze.** After "Freeze up to row 3" (`rows=3`), the "Freeze header row" entry
  still reads active (`active={freeze.rows > 0}`) and clicking it jumps straight to `rows=0` (clears
  *all* frozen rows) rather than reducing to 1, and the "up to…" entry that produced the current
  extent shows no active indication. Reachable from real UI; recoverable; consider showing the actual
  extent and making the toggle reduce rather than clear.

- **P2 — Unscoped `.column-resize-handle` z-index lift + duplicated selector.**
  `styles.css:2010` adds `.octo-prose table .column-resize-handle { z-index: 6 }` to lift the resize
  handle above frozen cells. The selector is unscoped, so it also raises the handle on *unfrozen*
  tables (previously it had no z-index). Behaviorally fine (the fix relies on the cascade merging with
  the pre-existing block at `styles.css:1957`), but the declarations for one element are now split
  across two non-adjacent identical selectors ~50 lines apart — easy to miss in future edits. Consider
  merging them, and scoping the lift to `.octo-frozen-table` if you want unfrozen tables untouched.

### Test-quality notes (P2)

The command/pure-helper layer is well covered, and the new colspan/rowspan regression tests are a
good addition. Remaining gaps, in rough priority:

- The "stays open across the document mousedown" test (`TableControls.test.tsx`) is effectively a
  tautology for the *stated* flaky-open bug: it dispatches a `mousedown` on the freeze button, which
  is *inside* the menu, so the outside-click handler ignores it by construction. Nothing in the
  component can close the submenu on an inside mousedown, so the assertion passes regardless. To
  actually exercise the close path, dispatch a document-level `mousedown` whose target is *outside*
  the menu and assert the menu closes, plus a separate assertion that an inside mousedown leaves the
  submenu open.
- The load-bearing safety claim — inline cell-style writes don't desync collaboration because
  `ignoreMutation` swallows them — is never exercised: every test uses
  `Table.configure({ resizable: false })` with plain cells, while production uses `resizable: true`
  with `TableCellView` NodeViews. An integration test under the production config asserting no
  `docChanged`/`getHTML()` change while freezing would validate the central invariant directly.
- `FreezeStyleView` (the command→plugin→DOM seam, including the `nodeDOM` wrapper-vs-table branch) and
  the `TableFreezeControl` ribbon component both have zero tests; these are jsdom-testable via
  attribute/class assertions (no real layout needed).

## 3. Overall verdict

**APPROVE**

Spec ✅ and no P0/P1 quality issues. The one genuine correctness bug found (leading-boundary freeze
loss) is an edge-triggered, cosmetically-recoverable P2 on a per-client view-state feature, and the
merged-cell correctness concern has already been fixed on the current head with regression tests. The
remaining items are hardening and test-coverage follow-ups.

## 4. Suggested follow-ups (non-blocking)

1. Change the remap bias to `map(pos, 1)` and add a boundary-insert test (top-of-doc table).
2. Early-return `render()` when there's nothing frozen and nothing to clear; batch reads before writes.
3. Add a `docChanged` guard/comment around the `meta.pos` application.
4. Make the row/column toggles reflect and reduce the actual frozen extent.
5. Replace the tautological submenu test with one that drives the real outside-mousedown close path;
   add a `resizable: true` + `TableCellView` no-desync integration test.

## 5. Additional observations

- The preview/export editor deliberately omits `TableFreeze` (`extensions.ts`), so the hardcoded
  light-mode CSS fallbacks in the freeze rules never render there; dark-mode vars (`--octo-bg`,
  `--octo-hover`) are defined for the interactive path.
- A frozen **column with no frozen rows** correctly relies on the wrapper's existing horizontal
  scroll and needs no vertical scroll container, so the `octo-has-frozen-rows` gate (rows only) is
  right, not a gap. `max-height: 60vh` is a *max*, so short tables keep their natural height.
- Coverage limits of this review: pixel-level sticky behavior, `border-collapse: separate` border
  redraw, and Chromium `overflow`/stacking were verified by reasoning and the author's browser notes,
  not re-run here (no browser/layout in this environment); keyboard a11y of the popover/submenu and
  true multi-client remap were not exercised.

OctoBoooot (approved):
## Re-review (DELTA) @ `ee8508873293` — merged-cell freeze bug fixed with TableMap → flipping REQUEST_CHANGES to APPROVE

The head advanced past my CR's `93a688f2f1c0` with one fix commit ("fix(docs): derive table freeze bands from TableMap so merged cells stay aligned") touching only `TableFreeze.ts` + its test. My major is fixed exactly as prescribed, byte-verified.

### Major (merged-cell freeze misalignment) — FIXED
`applyFrozenStyles` now derives the frozen band from the table's **`TableMap`** (span-aware grid), not DOM cell index:
- takes the PM `node`, `const map = TableMap.get(node)` (`:191`); `frozenCols` clamped to `map.width` (logical grid width, `:193`).
- membership + offset use the cell's **grid rect**: `inFrozenCol = rect.left < frozenCols` and `colLefts[rect.left]` (`:223`, `:234`) — so a `colspan` cell occupies its full column range and a `rowspan` cell "only lives in the row it starts," per the new doc comment.
- new `measureColumnWidths` (`:139`) sizes each grid column from a `colspan=1` cell where one exists, falling back otherwise — handling the "one DOM cell, N visual columns" divergence directly.
This is the logical-grid-column mapping my CR asked for. Re-verified against my failing input (freeze a header row / first column on a table with a `colspan=2` cell): offsets now key off `rect.left`, so the frozen band follows the real column geometry.

### Test — merged-cell regression added
`TableFreeze.test.ts` adds a dedicated merged-cell case, with a comment explicitly citing "#775 review" and covering **both** divergences I named — colspan (one DOM entry vs N visual columns) and rowspan (absent from covered rows) — asserting the TableMap-derived styler keys off the grid rect. The old DOM-index styler would fail these; the new one passes.

### No regression
Delta is only `TableFreeze.ts` + its test. The previously-verified-clean parts hold: freeze is pure view-state (no Y.Doc write / no collab/XSS surface), `apply()` still remaps positions and drops deleted-table entries, `FreezeStyleView` still tears down on update/destroy, the #761 context-menu guards remain, i18n symmetric. Code CI green at this head (Build / secret-scan / osv-scan / dependency-review / pr-title-lint).

### Net
The merged-cell freeze bug is fixed by switching `applyFrozenStyles` to TableMap-derived grid columns/offsets, with a regression test covering both colspan and rowspan. Code CI green. Flipping **REQUEST_CHANGES → APPROVE @`ee8508873293`**, superseding my CR stranded at the old head. `needs-human-review` (size/XL) still needs a human sign-off before merge. Credit @ploy-elison — the fix is exactly right (grid rect, not DOM index) and the regression test names the two failure modes.

## Parity checked
Diffed my CR head (`93a688f2f1c0`) → new head (`ee8508873293`): the delta is a single fix commit touching only `TableFreeze.ts` + `TableFreeze.test.ts`. The behavior change is exactly the intended fix — `applyFrozenStyles` freezing membership/offsets now derive from `TableMap` grid rects instead of DOM cell index. No other capability altered: view-state model, `apply()` remap/cleanup, `FreezeStyleView` teardown, the freeze commands, submenu, and i18n are unchanged. No capability dropped; the one behavioral delta is the merged-cell correctness fix.

Jerry-Xin (approved):
### ✅ Re-review @ee850887 — the merged-cell freeze-alignment blocker is fixed

My prior REQUEST_CHANGES flagged that `applyFrozenStyles` computed frozen-column membership and the sticky left-offset from the **DOM cell index**, which diverges from the logical grid column whenever merged cells are present. This revision resolves it correctly for **both** span directions.

#### 🔴 (resolved) DOM-index → span-aware logical grid
`packages/docs/src/editor/TableFreeze.ts` — `applyFrozenStyles` now takes the ProseMirror `node`, builds `TableMap.get(node)`, and for every cell resolves its logical rect via `map.findCell(...)`. Frozen-band membership is `rect.top < frozenRows` / `rect.left < frozenCols`, and offsets index by logical position (`rowTops[rect.top]`, `colLefts[rect.left]`). `frozenCols` clamps to `map.width` (logical) rather than the DOM cell count.

- **Colspan:** `measureColumnWidths` sizes each logical column from a `colspan=1` cell where one exists (`rect.right - rect.left === 1`), and splits a spanning cell's width across the columns it alone covers, so `colLefts` stays monotonic per visual column.
- **Rowspan (the subtle half):** the walk iterates the PM rows in lockstep with `domRow.cells` via a separate cursor. A rowspan-covered cell is absent from BOTH the covered row's DOM cells and that PM row's children, so the lockstep stays aligned and `map.findCell` yields the true rect. A cell that sits at DOM index 0 in a rowspan-skipped row is no longer mistaken for logical column 0.

#### ✅ Regression tests genuinely lock it
`TableFreeze.test.ts` renders fixtures through a real editor, so the DOM and the PM node share one TableMap (matching the live plugin). Two new cases exercise real merges and would fail under the old logic:
- colspan header (`[colspan=2][col2]`, freeze 2 cols): the colspan cell freezes at `0px`, the lone col-2 header is left untouched — the old code froze it at DOM index 1.
- rowspan (`rowspan=2` in col 0): row 1's single DOM cell is logical col 1 and must NOT freeze — the old code froze it because it sat at DOM index 0.

Column-width stub is `50 * colSpan`, mirroring production geometry; the non-merged common case is still covered by the retained uniform-grid tests.

#### No new regressions
Empty table guarded; unmeasured columns fall back to 0; `FreezeStyleView` now verifies the node is a `table` before styling. The common (no-merge) path is unchanged.

CI: Build / osv-scan / secret-scan / dependency-review all green. Governance checks (check-sprint / satisfy-code-review) are skipped and do not bar code approval.

Clearing my prior blocker.

PR #776, opened 2026-07-15
feat(loop): unify board/list/grouped filters + scope across views

### LINKED ISSUE #746: feat(loop): unify board/list/grouped filter & scope into one multi-select model
## Problem

The three loop views have inconsistent filtering:

- **Board / List**: server-side **single-select** filters via `GET /issues` (`listIssues`) — `status`, `priority`, `assignee`, `creator`, `project` are single values.
- **Grouped**: server-side **multi-select** via `GET /issues/grouped` (`gStatuses` / `gPriorities` / `gProjectIds`), plus a scope tab bar (全部 / 成员 / AI / 与我相关) that narrows `assignee_types`, plus a "no project" checkbox.
- The scope tab bar exists **only** in grouped view; board/list can't filter by member/AI/involves at all.

Root cause: `IssuePage`'s `Filters` interface carries both single (`status?`) and multi (`gStatuses`) fields, and each panel row is `view === "grouped" ? <Select multiple> : <Select>`.

## Reference model (target)

A uniform **multi-select array model** — `statusFilters[]` / `priorityFilters[]` / `assigneeFilters[]` / `creatorFilters[]` / `projectFilters[]` + `includeNoAssignee` / `includeNoProject` — applied to **all** views through a single filter function, with **scope** (all / members / agents) as an orthogonal store and grouping as a pure display dimension.

## Decision (already made)

- Keep dmloop **server-side + paginated** (do NOT move to client-side filtering). Adopt only the reference **filter-state shape** (all arrays + orthogonal `includeNo*` booleans).
- **Backend-first**: the loop backend must accept array params before this frontend ships (an old backend silently ignores array params → wrong/over-broad results). Tracked in the loop backend issue/MR (see "Depends on").

## Frontend scope (this issue)

In `packages/dmloop/src`:

- **`pages/IssuePage.tsx`**: collapse `Filters` to one array model — `statuses[] / priorities[] / assigneeIds[] / creatorIds[] / projectIds[] / noProject:boolean` + shared `keyword / dateField / dateRange / sortBy / sortDir`. Remove the `g*` fields and the per-row single/multi ternary → always `multiple`. Make one shared filter panel used by all three views. Lift the scope tabs (全部/成员/AI/与我相关) out of the `view === "grouped"` guard so they apply to board/list too (feed `scopeToAssigneeTypes(scope)` + `involves_user_id`). Update `activeFilters` (count array lengths) and `clearFilters` (reset to empty arrays).
- **`api/types.ts`**: add optional array fields to `ListParams` (`statuses / priorities / assignee_ids / creator_ids / project_ids / assignee_types / include_no_project`); keep the single fields for back-compat. `GroupedParams` already has these — the two shapes converge.
- **`api/issueApi.ts`**: `listIssues` sends the array params as CSV (mirror the grouped endpoint's `.join(",")`).
- Keep `keyword` (full-text `/issues/search`) and `sort` as board/list-only (grouped endpoint ignores both) — these stay as-is, not folded into the array model.

## Open product decision (non-blocking for the rest)

"与我相关" (involves) on the **paginated list** view: the backend `involves_user_id` intentionally **excludes** direct assign/create; grouped works around this with a 3-leg union (assignee ∪ creator ∪ involves) merged client-side, which recomputes `total` and has no `offset` → incompatible with the list view's server pagination. Decide one of:
- (a) degrade / disable the involves scope on the list view, or
- (b) add a backend param that does `OR(assigned ∪ created ∪ indirect)` in a single query so pagination stays correct.

Until decided, list-view involves can be disabled/degraded; board/grouped keep the existing fan-out.

## Depends on

The loop backend accepting array filter params on flat `ListIssues` (`statuses / priorities / creator_ids / project_ids / assignee_types`, all additive) + `creator_ids` on the grouped handler. **Must land and deploy first.**

## Verification

Real-env e2e in the dev workspace (dmloop has no vitest harness for the UI layer): each dimension multi-selects and narrows correctly across all three views; scope works on board/list; clearing resets; grouped still works; no regression to the filter panel open/close behavior.

Design analysis: `octo-loop-filter-unification-plan.md` (ecosystem root).

MrAladdin:
## Summary

Unifies the loop issues board/list/grouped views onto **one shared multi-select filter model + scope**, aligning with the reference board. Closes #746.

Before: filters were split — board/list used single-value params, grouped used a separate multi-select set + a scope tab that only appeared in grouped view. Selections didn't carry across views.

After — one filter panel, applied identically across all three views:
- **Array multi-select** for every dimension: statuses / priorities / assignee_ids / creator_ids / project_ids / label_ids, plus orthogonal **no-assignee / no-project** toggles.
- **scope** (全部/成员/AI) spans all three views via `assignee_types` (previously grouped-only). The board drops the old 「与我相关」tab.
- **label** and **no-assignee** filters added (board-parity).
- `filterSelect` helper + `arrayFilterQuery` serializer remove per-view duplication.

Server-side + paginated filtering is retained (only the filter-state shape is unified). Keyword search still routes to the full-text `/issues/search` endpoint.

## Backend dependency (cross-repo)

The unified board/list filters send the plural query params to flat `/issues`. Backend support is delivered by **octo-multica MR !55 (merged)** and **MR !57** (label_ids + include_no_assignee). This frontend must deploy **after** those land (backend-first, per the ecosystem's staged-rollout convention). Verified live against the dev backend: `assignee_types`, `statuses`, `priorities`, `creator_ids`, `project_ids`, `include_no_project`, `label_ids`, `include_no_assignee` are all honored (each narrows results); backend integration tests in octo-multica prove per-param narrowing (incl. cross-workspace label scoping).

## Interaction correctness (7-dim UX + adversarial)

- **My Loop (`involves`) locked to grouped**: view-switch and scope tabs hidden — closes a latent escape where switching My Loop to board/list showed the whole workspace under a "my" title.
- **no-assignee × scope**: only effective at scope=全部 (combining with `assignee_types` is empty-by-construction) — checkbox disabled + unchecked + not sent otherwise; the active-filter badge is gated the same way (single source `noAssigneeActive`).
- **keyword search × scope/filters**: search bypasses filters (separate endpoint), so scope tabs + filters are disabled and shown inactive while searching.
- **显示 dropdown**: closes after picking board/grouped; stays open for list so the newly-revealed sort control is reachable.
- **reload failure**: clears the result set + toasts (no stale actionable rows under new filters; list selection auto-prunes).
- Empty/loading/edge (seq-guarded concurrency, maxTagCount overflow, pagination clamp) covered.

## Verification

- `pnpm test` (dmloop) — 11 files / 60 tests green, incl. new `arrayFilterQuery` query-contract test (plural join + boolean→"true" + unset→undefined).
- tsc clean for changed files.
- **Live e2e** (Alice-Loop): 3-view filter panel identical (board=list; grouped omits keyword by design); scope→assignee_types on board (empties board when all issues are agent-assigned); label filter renders+filters; no-assignee disables under members/AI; scope tabs disabled during search; My Loop shows no view-switch/scope controls; @-mention group relabeled 事项→回路.
- Two-model adversarial: Codex (multiple rounds — My-Loop escape, search-scope visual, involves-path error handling, badge count, scope pagination reset, duplicate i18n key all fixed) + skeptical code-reviewer (confirmed clean; caught the badge-count gate).

## Blast radius (Rule 8)

- `listIssues`/`listGroupedIssues`/`ListParams`/`GroupedParams`/`IssueScope` — all additive (new optional fields); other `listIssues` callers (`mentionExtension.ts`, `IssueDetailPage.tsx`) pass only `limit`, unaffected.
- `loop.mention.groupIssues` i18n relabel used only by `CommentComposer`.
- No brand leak (scanned): removed 3 upstream-brand code comments in touched files.

## Grouped-view keyword search (now included)

Keyword search works in the grouped view too: `searchIssues` flat results are grouped client-side by assignee (`groupIssuesByAssignee`, unit-tested). Excluded on My Loop (search endpoint can't scope to `involves`).

## Review round (commit db83d203)

Addressed all review findings: My Loop involves fan-out now strips the plural `assignee_ids`/`creator_ids` (blocking) + hides the who-filters on My Loop; `arrayFilterQuery` folds empty arrays to `undefined` locally; `activeFilters` suppresses non-keyword dimensions during search; removed the dead singular `status`/`priority`/`assignee_id`/`creator_id`/`project_id` from `ListParams` + `listIssues`. Backend plural-param support: octo-multica MR !55 + !57 (both merged). Re-ran two-model adversarial (both clean) + 63 unit tests green.

### DIFF
--- a/packages/dmloop/src/api/issueFilterQuery.ts
+++ b/packages/dmloop/src/api/issueFilterQuery.ts
@@ -1,0 +1,21 @@
+import type { ListParams } from "./types";
+
+// 数组/布尔筛选参 → 查询串(逗号连接数组、布尔转 "true")。listIssues 与 grouped 共用,防漂移。
+// 空数组由调用方在 reload 里预折叠成 undefined,故此处不发空串。纯函数、只依赖类型,便于契约测试。
+export type ArrayFilterParams = Pick<ListParams,
+  "statuses" | "priorities" | "assignee_types" | "assignee_ids" | "include_no_assignee" |
+  "creator_ids" | "project_ids" | "include_no_project" | "label_ids">;
+
+export function arrayFilterQuery(p: ArrayFilterParams): Record<string, string | undefined> {
+  return {
+    statuses: p.statuses?.join(","),
+    priorities: p.priorities?.join(","),
+    assignee_types: p.assignee_types?.join(","),
+    assignee_ids: p.assignee_ids?.join(","),
+    include_no_assignee: p.include_no_assignee ? "true" : undefined,
+    creator_ids: p.creator_ids?.join(","),
+    project_ids: p.project_ids?.join(","),
+    include_no_project: p.include_no_project ? "true" : undefined,
+    label_ids: p.label_ids?.join(","),
+  };
+}
--- a/packages/dmloop/src/api/issueFilterQuery.ts
+++ b/packages/dmloop/src/api/issueFilterQuery.ts
@@ -1,21 +1,23 @@
 import type { ListParams } from "./types";
 
 // 数组/布尔筛选参 → 查询串(逗号连接数组、布尔转 "true")。listIssues 与 grouped 共用,防漂移。
-// 空数组由调用方在 reload 里预折叠成 undefined,故此处不发空串。纯函数、只依赖类型,便于契约测试。
+// 空数组本地折叠成 undefined(不发空串),契约不再依赖调用方预折叠。纯函数、只依赖类型,便于契约测试。
 export type ArrayFilterParams = Pick<ListParams,
   "statuses" | "priorities" | "assignee_types" | "assignee_ids" | "include_no_assignee" |
   "creator_ids" | "project_ids" | "include_no_project" | "label_ids">;
 
+const csv = (a?: string[]): string | undefined => (a && a.length ? a.join(",") : undefined);
+
 export function arrayFilterQuery(p: ArrayFilterParams): Record<string, string | undefined> {
   return {
-    statuses: p.statuses?.join(","),
-    priorities: p.priorities?.join(","),
-    assignee_types: p.assignee_types?.join(","),
-    assignee_ids: p.assignee_ids?.join(","),
+    statuses: csv(p.statuses),
+    priorities: csv(p.priorities),
+    assignee_types: csv(p.assignee_types),
+    assignee_ids: csv(p.assignee_ids),
     include_no_assignee: p.include_no_assignee ? "true" : undefined,
-    creator_ids: p.creator_ids?.join(","),
-    project_ids: p.project_ids?.join(","),
+    creator_ids: csv(p.creator_ids),
+    project_ids: csv(p.project_ids),
     include_no_project: p.include_no_project ? "true" : undefined,
-    label_ids: p.label_ids?.join(","),
+    label_ids: csv(p.label_ids),
   };
 }
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -25,6 +25,7 @@ import type {
 } from "../api/types";
 import { ISSUE_SORT_FIELDS, ISSUE_DATE_FIELDS } from "../api/types";
 import { listIssues, searchIssues, listGroupedIssues, listMyGroupedIssues, getAgentTaskSnapshot } from "../api/issueApi";
+import { groupIssuesByAssignee } from "../api/issueGrouping";
 import { listProjectOptions } from "../api/directory";
 import { listLabels } from "../api/labelApi";
 import { useAssigneeCandidates } from "../ui/useAssigneeCandidates";
@@ -150,8 +151,18 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
       date_end: endExclusive ? endExclusive.toISOString() : undefined,
     };
 
-    // 分组板:走 /issues/grouped(按负责人);grouped 不吃关键词/排序,故在分组视图隐藏。
+    // 分组板:走 /issues/grouped(按负责人)。
     if (view === "grouped") {
+      const gkw = f.keyword.trim();
+      // 关键词(仅主看板,非「我的回路」)→ 全文搜索平铺结果,前端按负责人分组呈现。搜索端点
+      // 独立语义(不吃 scope/其它筛选,上限 50),故与常规分组拉取分道;「我的回路」隐藏关键词、不入此路。
+      if (gkw && !isMyLoop) {
+        searchIssues(gkw, { limit: 50 })
+          .then(({ issues }) => { if (my === seq.current) setGroups(groupIssuesByAssignee(issues)); })
+          .catch(onErr)
+          .finally(() => { if (my === seq.current) setLoading(false); });
+        return;
+      }
       // 「与我相关」(仅「我的回路」页)= 指派给我 ∪ 我创建 ∪ 间接关联的并集扇出;需当前成员
       // 后端 id,未解析出则清空并等待(不回退成无 scope 全量)。myMemberId 在依赖里,解析后自动重取。
       if (scope === "involves") {
@@ -269,20 +280,23 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
 
   const title = defaultScope === "involves" ? t("loop.nav.myloop") : t("loop.nav.issue");
 
-  // 关键词走全文搜索端点(独立语义,后端不吃其它筛选/排序)。搜索激活时禁用面板内其它筛选,
-  // 避免它们「看起来生效、实际被忽略」的误导(grouped 视图隐藏关键词,故 searching 恒 false)。
-  const searching = view !== "grouped" && !!f.keyword.trim();
+  // 关键词走全文搜索端点(独立语义,不吃其它筛选/排序)。搜索激活时禁用面板内其它筛选/scope,
+  // 避免它们「看起来生效、实际被忽略」的误导。主看板三视图均支持;「我的回路」无关键词(搜索不吃 involves)。
+  const searching = !isMyLoop && !!f.keyword.trim();
 
-  // 「筛选」按钮上的激活计数(每个非空维度记 1;keyword 仅在其生效的视图计)。
+  // 「筛选」按钮上的激活计数(每个生效维度记 1)。搜索激活时只有 keyword 生效——其它维度被
+  // 搜索端点忽略且面板已禁用,故不计,免得徽章高亮却对结果无影响(与 scope tab 搜索时去激活一致)。
   const activeFilters =
-    (f.statuses.length ? 1 : 0) +
-    (f.priorities.length ? 1 : 0) +
-    (f.assigneeIds.length || noAssigneeActive ? 1 : 0) +
-    (f.creatorIds.length ? 1 : 0) +
-    (f.projectIds.length || f.noProject ? 1 : 0) +
-    (f.labelIds.length ? 1 : 0) +
-    (f.dateRange ? 1 : 0) +
-    (view !== "grouped" && f.keyword.trim() ? 1 : 0);
+    (searching
+      ? 0
+      : (f.statuses.length ? 1 : 0) +
+        (f.priorities.length ? 1 : 0) +
+        (f.assigneeIds.length || noAssigneeActive ? 1 : 0) +
+        (f.creatorIds.length ? 1 : 0) +
+        (f.projectIds.length || f.noProject ? 1 : 0) +
+        (f.labelIds.length ? 1 : 0) +
+        (f.dateRange ? 1 : 0)) +
+    (!isMyLoop && f.keyword.trim() ? 1 : 0);
 
   const clearFilters = () =>
     update({
@@ -316,11 +330,15 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
       </div>
       {filterSelect(t("loop.filter.status"), f.statuses, (v) => update({ statuses: v as IssueStatus[] }), statusOpts, { maxTagCount: 2 })}
       {filterSelect(t("loop.filter.priority"), f.priorities, (v) => update({ priorities: v as IssuePriority[] }), priorityOpts, { maxTagCount: 2 })}
-      {filterSelect(t("loop.filter.assignee"), f.assigneeIds, (v) => update({ assigneeIds: v }), cands.map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>)), { filter: true })}
-      <div className="loop-fields__row">
-        <Checkbox className="loop-nofilter" checked={noAssigneeActive} onChange={(e) => update({ noAssignee: !!e.target.checked })} disabled={searching || scope !== "all"}>{t("loop.filter.noAssignee")}</Checkbox>
-      </div>
-      {filterSelect(t("loop.filter.creator"), f.creatorIds, (v) => update({ creatorIds: v }), cands.filter((c) => c.type === "member").map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>)), { filter: true })}
+      {/* 「谁」类筛选(负责人/无负责人/创建者)在「我的回路」隐藏——该页本就锁定「与我相关」,
+          按他人负责人/创建者筛与语义矛盾,且会污染并集扇出(见 listMyGroupedIssues)。 */}
+      {!isMyLoop && filterSelect(t("loop.filter.assignee"), f.assigneeIds, (v) => update({ assigneeIds: v }), cands.map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>)), { filter: true })}
+      {!isMyLoop && (
+        <div className="loop-fields__row">
+          <Checkbox className="loop-nofilter" checked={noAssigneeActive} onChange={(e) => update({ noAssignee: !!e.target.checked })} disabled={searching || scope !== "all"}>{t("loop.filter.noAssignee")}</Checkbox>
+        </div>
+      )}
+      {!isMyLoop && filterSelect(t("loop.filter.creator"), f.creatorIds, (v) => update({ creatorIds: v }), cands.filter((c) => c.type === "member").map((c) => (<Select.Option key={c.id} value={c.id}>{c.name}</Select.Option>)), { filter: true })}
       {projects.length > 0 && filterSelect(t("loop.filter.project"), f.projectIds, (v) => update({ projectIds: v }), projectOpts, { filter: true })}
       <div className="loop-fields__row">
         <Checkbox className="loop-nofilter" checked={f.noProject} onChange={(e) => update({ noProject: !!e.target.checked })} disabled={searching}>{t("loop.filter.noProject")}</Checkbox>
@@ -335,8 +353,8 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
           <DatePicker type="dateRange" density="compact" disabled={searching} value={f.dateRange} onChange={(d) => update({ dateRange: Array.isArray(d) && d.length === 2 && d[0] && d[1] ? (d as Date[]) : undefined })} placeholder={t("loop.filter.dateRange")} zIndex={FIELD_POPUP.zIndex} style={{ flex: 1 }} />
         </div>
       </div>
-      {/* 关键词走全文搜索(独立语义,不与 grouped 组合),故分组视图隐藏。 */}
-      {view !== "grouped" && (
+      {/* 关键词走全文搜索(平铺,分组视图前端按负责人再分组呈现)。「我的回路」不支持(搜索不吃 involves)。 */}
+      {!isMyLoop && (
         <div className="loop-fields__row">
           <div className="loop-fields__label">{t("loop.search.issue")}</div>
           <Input className="loop-search" prefix={<Search size={14} />} placeholder={t("loop.search.issue")} value={f.keyword} onChange={(v) => update({ keyword: v })} showClear style={{ width: "100%" }} />
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -21,10 +21,12 @@ import type {
   IssuePriority,
   IssueSortField,
   IssueDateField,
+  IssueLabel,
 } from "../api/types";
 import { ISSUE_SORT_FIELDS, ISSUE_DATE_FIELDS } from "../api/types";
 import { listIssues, searchIssues, listGroupedIssues, listMyGroupedIssues, getAgentTaskSnapshot } from "../api/issueApi";
 import { listProjectOptions } from "../api/directory";
+import { listLabels } from "../api/labelApi";
 import { useAssigneeCandidates } from "../ui/useAssigneeCandidates";
 import { ISSUE_STATUS_ORDER, PRIORITY_ORDER, isActiveRun } from "../ui/meta";
 import IssueBoard from "../panel/IssueBoard";
@@ -36,28 +38,26 @@ import { readView, writeView } from "../ui/viewMode";
 
 type ViewMode = "board" | "grouped" | "list";
 
-// scope pill → assignee_types 过滤(仅 /grouped 支持;all/involves 不按类型收窄)。
+// scope pill → assignee_types 过滤(看板/列表/分组统一走后端 assignee_types)。
+// all / involves 不按类型收窄(involves 走用户关系并集,见 reload)。
 function scopeToAssigneeTypes(scope: IssueScope): ("member" | "agent" | "squad")[] | undefined {
   if (scope === "members") return ["member"];
   if (scope === "agents") return ["agent", "squad"];
   return undefined;
 }
 
+// 统一筛选:全维度数组多选 + 正交的「无负责人 / 无项目」布尔,同一套应用到看板/列表/分组
+// 三视图。keyword 走独立全文搜索端点,故与其它维度不并存于同一请求。
 interface Filters {
   keyword: string;
-  status?: IssueStatus;        // board/list 单选
-  priority?: IssuePriority;    // board/list 单选
-  assignee?: string;
-  creator?: string;
-  project?: string;            // board/list 单选
-  // 分组板专属多选 + no-project(仅 /grouped 支持 statuses[]/priorities[]/project_ids[]/include_no_project)。
-  // 注:无 include_no_assignee —— scope pill 用 assignee_types(类型级、顶层 AND),后端
-  // include_no_assignee 只与 assignee_filters(具体 actor)配对才是「含」语义;与 assignee_types/
-  // involves 的 assignee_id 腿 AND 会矛盾成空。grouped 在 scope=all 下本就显示未指派组。
-  gStatuses: IssueStatus[];
-  gPriorities: IssuePriority[];
-  gProjectIds: string[];
+  statuses: IssueStatus[];
+  priorities: IssuePriority[];
+  assigneeIds: string[];
+  noAssignee: boolean;
+  creatorIds: string[];
+  projectIds: string[];
   noProject: boolean;
+  labelIds: string[];
   dateField: IssueDateField; // 时间范围筛选的列(created_at|updated_at)
   dateRange?: Date[];        // [start, end];为空则不按时间筛选
   sortBy: IssueSortField;
@@ -80,15 +80,20 @@ interface IssuePageProps {
 
 export default function IssuePage({ defaultScope, defaultView, viewKey }: IssuePageProps = {}) {
   const { t } = useI18n();
+  // 「我的回路」= involves:只有分组视图能表达「指派/创建/关联」三路并集(listMyGroupedIssues 扇出),
+  // 看板/列表无对应查询会退化成显示全工作区 → 该页锁死分组、不给切视图。
+  const isMyLoop = defaultScope === "involves";
   const [issues, setIssues] = useState<Issue[]>([]);
   const [groups, setGroups] = useState<IssueGroup[]>([]);
   const [running, setRunning] = useState<ReadonlySet<string>>(new Set());
   const [total, setTotal] = useState(0);
   const [loading, setLoading] = useState(true);
-  const [view, setView] = useState<ViewMode>(() => viewKey ? readView(viewKey, ["board", "grouped", "list"], defaultView ?? "board") : (defaultView ?? "board"));
+  const [view, setView] = useState<ViewMode>(() => isMyLoop ? "grouped" : (viewKey ? readView(viewKey, ["board", "grouped", "list"], defaultView ?? "board") : (defaultView ?? "board")));
   const [scope, setScope] = useState<IssueScope>(defaultScope ?? "all");
-  const [f, setF] = useState<Filters>({ keyword: "", gStatuses: [], gPriorities: [], gProjectIds: [], noProject: false, sortBy: "position", sortDir: "desc", dateField: "created_at" });
+  const [f, setF] = useState<Filters>({ keyword: "", statuses: [], priorities: [], assigneeIds: [], noAssignee: false, creatorIds: [], projectIds: [], noProject: false, labelIds: [], sortBy: "position", sortDir: "desc", dateField: "created_at" });
   const [page, setPage] = useState(0); // 0-based，仅列表视图分页
+  // 「无负责人」仅在 scope=全部 时有效(与 assignee_types 组合恒空)。单一来源:发送/勾选态/计数共用,防漂移。
+  const noAssigneeActive = scope === "all" && f.noAssignee;
   // 筛选/显示 面板受控可见:不用 clickToHide(它会在点击内部 Select/DatePicker/Checkbox
   // 时冒泡关闭,且这些控件的 portal 选项面板点击也会误触发关闭),改为显式
   // onVisibleChange —— 仅外部点击关闭,内部多字段交互保持面板打开。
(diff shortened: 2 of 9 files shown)


Jerry-Xin (dismissed):
Relevant PR for `Mininglamp-OSS/octo-web`; no blocking issues found.

💬 Non-blocking
- 🟡 Warning: During keyword search, non-keyword filters are disabled and omitted from the `/issues/search` request, but `activeFilters` still counts them and keeps the filter button highlighted/badged. This can make inactive filters look applied while search results are actually unscoped. Consider gating non-keyword counts with `!searching`, matching the scope-tab behavior. See [IssuePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-m2wgqem4/repo/packages/dmloop/src/pages/IssuePage.tsx:276) and [IssuePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-m2wgqem4/repo/packages/dmloop/src/pages/IssuePage.tsx:410).
- 🔵 Suggestion: `arrayFilterQuery` relies on callers to pre-collapse empty arrays, but the helper’s contract says unset dimensions should not be sent. Defensive `length ? join : undefined` handling would make the shared serializer safer for future callers. See [issueFilterQuery.ts](/Users/mlamp/.openclaw/tmp/pr-review-m2wgqem4/repo/packages/dmloop/src/api/issueFilterQuery.ts:9).

✅ Highlights
- Shared query serialization is a good reduction of duplicated board/list/grouped request logic.
- The My Loop grouped lock avoids a real scope escape risk.
- Reload error handling now clears stale rows, which is safer for list selections and bulk actions.

Verification note: `git diff --check` passed. `pnpm i18n:check` could not run because `node_modules` is missing and `typescript` could not be resolved locally.

Jerry-Xin (approved):
Reviewed as a behavior-preserving unification of the loop board/list/grouped filter + scope logic onto one shared multi-select model. No blocking issues found; unified path preserves each view's prior semantics and defaults.

## Behavior preservation
- **Unified state**: `Filters` in `packages/dmloop/src/pages/IssuePage.tsx` drops the per-view single-value fields (status/priority/assignee/creator/project) and the grouped-only arrays (gStatuses/gPriorities/gProjectIds), replaced by one array set (statuses/priorities/assigneeIds/creatorIds/projectIds/labelIds) plus noAssignee/noProject booleans. All three call sites now consume the same `common` object in `reload`.
- **Defaults preserved**: default `scope` is `all`, and `scopeToAssigneeTypes("all")` returns `undefined`, so board/list get no assignee_types narrowing by default and continue to show all items. Empty arrays collapse to `undefined` (`length ? … : undefined`) so unselected dimensions are never sent — matching the prior board/list behavior of omitting unset single-value params.
- **Scope across views**: `assignee_types` (previously grouped-only) now applies to flat `/issues` too for board/list. The old board 「与我相关」/involves tab is intentionally dropped; scope tabs are `all/members/agents` across all three views. This depends on backend plural-param support on flat `/issues` (octo-multica MR !55/!57), so this frontend must deploy backend-first.
- **My Loop scope escape closed**: `isMyLoop = defaultScope === "involves"` locks view to grouped and hides the view-switch + scope tabs, so switching My Loop to board/list can no longer show the whole workspace under a "my" title. `listMyGroupedIssues` strips the user-relation single filters and fans out the assignee∪creator∪involves legs; `common` still flows the non-relation dimensions in, which narrows within My Loop rather than leaking.
- **no-assignee gating**: `noAssigneeActive = scope === "all" && f.noAssignee` is the single source used for send (`include_no_assignee`), checkbox checked/disabled state, and the active-filter badge, and it is in the `reload` dependency list. Under members/AI the checkbox is disabled and the param is not sent — correct, since assignee_types AND include_no_assignee is empty-by-construction.
- **No authz/visibility drift**: scope maps only to assignee_type buckets; no per-view access gate is removed. The reload error path now clears the result set + toasts so stale actionable rows do not persist under new filters (safer for list selection / bulk ops).

## Non-blocking
- Warning: during keyword search, non-keyword filters are disabled and omitted from the `/issues/search` request, but `activeFilters` still counts them and keeps the filter button badged, so inactive filters can look applied while results are unscoped. Consider gating the non-keyword dimension counts with the searching flag, matching how the scope tabs already suppress their active state during search. See `packages/dmloop/src/pages/IssuePage.tsx` (activeFilters computation and the filter toolbar badge).
- Suggestion: `arrayFilterQuery` in `packages/dmloop/src/api/issueFilterQuery.ts` relies on callers to pre-collapse empty arrays; a defensive length-guard in the shared serializer would make it safer for future callers that forget the convention.

## Highlights
- `filterSelect` helper + `arrayFilterQuery` serializer remove genuine per-view duplication (net +83 with substantial deletions).
- New `arrayFilterQuery` contract test locks the plural-join / boolean→"true" / unset→undefined query shape.
- The My Loop grouped lock and the reload-clears-stale-rows change both close real UX/correctness gaps rather than just moving code.

## CI
Code checks green: Build, secret-scan (gitleaks), osv-scan, dependency-review all pass. Governance checks (check-sprint, satisfy-code-review skipping; code-review pending) do not bar a code approve. Not a draft.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #776 (octo-web)

## Summary

This PR unifies the issue-list filter system across board, list, and grouped views. Previously, board/list used single-select filters (`status`, `priority`, `assignee`, `creator`, `project`) while grouped used separate multi-select state (`gStatuses`, `gPriorities`, `gProjectIds`). The refactor collapses both into one shared multi-select `Filters` shape with a pure `arrayFilterQuery` helper that serializes arrays to comma-joined query strings. Scope tabs (all/members/agents) now span all three views. "My Loop" (`involves` scope) is locked to grouped view. Label filtering is added. Error handling now clears results and shows a toast on fetch failure.

The refactoring is clean and well-commented. The `arrayFilterQuery` pure function with dedicated tests is a good pattern that prevents drift between the flat and grouped API paths.

## Verification

- ✅ **Filter state unification** — single `Filters` interface replaces the old dual `status`/`gStatuses` split; `clearFilters` resets all dimensions; `activeFilters` badge counts correctly across all views.
- ✅ **`arrayFilterQuery` contract** — test file locks comma-join + boolean-to-string mapping; empty arrays fold to `undefined` (never sent); other callers of `listIssues` (`IssueDetailPage.tsx:381`, `mentionExtension.ts:151`) pass no array fields, so the spread is harmless.
- ✅ **Race-guard preserved** — `seq.current` counter + `my === seq.current` checks on all three paths (grouped, my-grouped, flat) are unchanged; new `onErr` also respects the guard.
- ✅ **`listMyGroupedIssues` fan-out** — `base` correctly strips singular user-relation fields (`assignee_id`, `creator_id`, `involves_user_id`, `assignee_types`); new array filter fields (`assignee_ids`, `creator_ids`, etc.) flow through as user-intended AND filters across all three legs.
- ✅ **`noAssigneeActive` gating** — derived value correctly gates `include_no_assignee` to `scope === "all"` only; checkbox disabled and unchecked when scope is not "all"; `activeFilters` uses the derived value.
- ✅ **i18n** — `loop.toast.loadFailed`, `loop.filter.label`, `loop.filter.noAssignee` all resolve to existing or newly-added keys in both locales.
- ✅ **`listLabels` / `IssueLabel`** — import target exists in `labelApi.ts` and type in `types.ts`.

Static analysis only at head `748fee95`; build and tests not executed in this environment.

## Findings

### P2 — New array query params on `GET /issues` — backend contract not verifiable from frontend (`packages/dmloop/src/api/issueApi.ts:51`)

`listIssues` now spreads `...arrayFilterQuery(params)` which emits `assignee_types`, `assignee_ids`, `creator_ids`, `label_ids`, `include_no_assignee`, and `include_no_project` to `GET /issues`. Previously this endpoint only received singular `status`/`priority`/`assignee_id`/`creator_id`/`project_id`. The scope tabs (all/members/agents) on board/list views depend on the backend honoring `assignee_types` on this endpoint. If the backend silently ignores any of these six new params, the corresponding filter controls would appear interactive but have no effect on results.

Fix direction: confirm the backend `GET /issues` endpoint accepts all new array params. If any are unsupported, either gate the UI controls per-endpoint or add backend support first.

### P2 — Old singular and new plural filter fields coexist in `listIssues` query (`packages/dmloop/src/api/issueApi.ts:48`)

The query object sent to `GET /issues` includes both `status` (singular, from `params?.status`) and `statuses` (plural, from `arrayFilterQuery`). Same for `priority`/`priorities`, `assignee_id`/`assignee_ids`, `creator_id`/`creator_ids`, `project_id`/`project_ids`. From IssuePage the singular fields are always `undefined`, so no actual collision occurs. But the dual fields are a latent ambiguity: a future caller setting both would produce undefined backend behavior. The `ListParams` comment documents them as "mutually exclusive" but the query builder doesn't enforce this.

Fix direction: once no caller uses the singular forms for issue listing, remove them from the `listIssues` query builder. Alternatively, add a runtime assertion or a comment at the call sites.

- Scope tab list correctly drops "involves" — only accessible via the dedicated "My Loop" page.
- `switchView` closes the "show" panel for non-list views (sort options are list-only).
- `searching` flag correctly disables filter controls and is always false for grouped view (so grouped filters remain active).
- `filterSelect` helper correctly defaults `maxTagCount` to 1 and handles null onChange values.
- Date range handling (half-open interval via `+1 day`) is unchanged.
- "My Loop" view correctly locks to grouped view and hides scope tabs and the "show" dropdown.

## Blind-spot checklist

- **C1 (dual-path parity)** — Clear. The old single-select and new multi-select paths coexist in types but are never both populated from the same caller. `clearFilters` resets all dimensions uniformly.
- **C2 (control-flow ordering)** — Clear. `arrayFilterQuery` spread position differs between `listIssues` (after singular fields) and `listGroupedIssues` (before singular fields), but no key collisions exist in either case.
- **C3–C6** — N/A. No auth, build, or governance changes.

## Data-flow trace

- `f.statuses[]` → `common.statuses` → `arrayFilterQuery` → `statuses: "todo,in_progress"` → `httpGet` query string. Verified: data flows from UI state to API call.
- `f.assigneeIds[]` → `common.assignee_ids` → `arrayFilterQuery` → `assignee_ids: "a1,a2"` → both `listIssues` and `listGroupedIssues`. Verified.
- `noAssigneeActive` (= `scope === "all" && f.noAssignee`) → `common.include_no_assignee` → `arrayFilterQuery` → `include_no_assignee: "true"`. Gated correctly. Verified.
- `f.labelIds[]` → `common.label_ids` → `arrayFilterQuery` → `label_ids: "l1,l2"`. New dimension, wired to all three views. Verified.
- `scopeToAssigneeTypes(scope)` → explicit spread at call sites → `arrayFilterQuery` picks up `assignee_types`. Verified for grouped and flat paths.
- `listMyGroupedIssues` fan-out: `common` fields flow into all 3 legs as AND filters; singular user-relation fields stripped per-leg. Verified correct semantics.

## Cross-round blocker review

## Verdict: APPROVED

Well-structured unification that eliminates filter-state drift between views. The two P2 items are a backend contract dependency (confirm `GET /issues` supports the new array params) and a code-hygiene note about coexisting singular/plural fields. Neither is a merge blocker from the frontend perspective.

[Octo-Q] verdict: APPROVE — no P0/P1 found; two P2 items are backend contract verification and code-hygiene notes, neither blocking from frontend analysis.

packages/dmloop/src/pages/IssuePage.tsx
```diff
@@ -103,51 +108,67 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
...
-      if (scope === "involves" && !myMemberId) {
-        if (my === seq.current) { setGroups([]); setLoading(false); }
+      // 「与我相关」(仅「我的回路」页)= 指派给我 ∪ 我创建 ∪ 间接关联的并集扇出;需当前成员
+      // 后端 id,未解析出则清空并等待(不回退成无 scope 全量)。myMemberId 在依赖里,解析后自动重取。
+      if (scope === "involves") {
+        if (!myMemberId) {
+          if (my === seq.current) { setGroups([]); setLoading(false); }
+          return;
+        }
+        listMyGroupedIssues(myMemberId, { ...common, limit: 100 })
```
OctoBoooot, line 173: **[major]** My Loop (`involves`) fan-out leaks the plural user-relationship filters into all three legs. In My Loop the filter panel is still rendered (only scope tabs + 显示 are `!isMyLoop`-gated), so the assignee/creator multi-selects are reachable and write `assignee_ids`/`creator_ids` into `common`. `listMyGroupedIssues` builds `base` by stripping only the *singular* `assignee_id`/`creator_id` (issueApi.ts:137), not the plural forms — so every leg keeps them. Failing input: My Loop → creator = Bob ⇒ leg `assignee_id=me` AND `creator_ids=[Bob]` (unexpected AND — the exact case issueApi.ts:135-136 says the strip prevents), leg `creator_id=me` AND `creator_ids=[Bob]` (collision → empty unless I am Bob). The union silently narrows / drops the "created by me" branch. Fix: strip `assignee_ids`/`creator_ids` in `base` too, mirroring the singular strip (the dropdowns write the plural params now).

packages/dmloop/src/api/issueFilterQuery.ts
```diff
@@ -0,0 +1,21 @@
...
+// 数组/布尔筛选参 → 查询串(逗号连接数组、布尔转 "true")。listIssues 与 grouped 共用,防漂移。
+// 空数组由调用方在 reload 里预折叠成 undefined,故此处不发空串。纯函数、只依赖类型,便于契约测试。
+export type ArrayFilterParams = Pick<ListParams,
+  "statuses" | "priorities" | "assignee_types" | "assignee_ids" | "include_no_assignee" |
+  "creator_ids" | "project_ids" | "include_no_project" | "label_ids">;
+
+export function arrayFilterQuery(p: ArrayFilterParams): Record<string, string | undefined> {
+  return {
+    statuses: p.statuses?.join(","),
+    priorities: p.priorities?.join(","),
```
OctoBoooot: **[minor]** `p.statuses?.join(",")` maps `[]` → `""` (empty string, which is truthy in the query object → would be sent). Safe today only because every caller pre-collapses empty arrays to `undefined` (reload:138-147). A self-defensive `arr?.length ? arr.join(",") : undefined` here would make the documented "never send empty" guarantee local instead of caller-dependent — a footgun for the next caller otherwise.

OctoBoooot (changes requested):
## Review: unify board/list/grouped filters + scope across views (#776)

**Verdict:** Request changes — one real correctness bug in the My Loop (`involves`) fan-out that defeats the PR's own stated de-pollution invariant.

Code CI is green (Build / secret-scan / osv-scan / dependency-review / pr-title-lint all pass at this head — the only "pending" is the `code-review` governance context, which doesn't bar the verdict). The unification itself is clean and well-tested; the issue is one missed param in the My-Loop union.

### Major
- `packages/dmloop/src/pages/IssuePage.tsx:162` (root cause in `packages/dmloop/src/api/issueApi.ts:137`) — **the My Loop involves fan-out leaks the plural user-relationship filters into all three legs.**
  In My Loop the filter panel is rendered (only the scope tabs + 显示 panel are `!isMyLoop`-gated), so the **assignee** and **creator** multi-selects are reachable, and they write to the *plural* `assignee_ids` / `creator_ids`. My Loop reload passes `{ ...common }` into `listMyGroupedIssues`, whose `base` strips only the *singular* `assignee_id` / `creator_id` (`issueApi.ts:137`) — not the plural forms. So every leg retains them.
  This is exactly the pollution your own comment (`issueApi.ts:135-136`) says the strip exists to prevent — and it names the creator dropdown as the case. Concrete failing input: My Loop → filter creator = "Bob":
  - leg `assignee_id = me` **AND** `creator_ids = [Bob]` → "assigned to me *and* created by Bob" (unexpected AND — the comment's warned case),
  - leg `creator_id = me` **AND** `creator_ids = [Bob]` → singular+plural collision → empty unless I am Bob,
  - leg `involves = me` **AND** `creator_ids = [Bob]` → "involves me *and* created by Bob".
  Same shape for the assignee dropdown (`assignee_ids`). The union silently narrows and can drop the "created by me" branch entirely.
  Fix: strip the plural user-relationship filters in `base` too, mirroring the singular strip — `assignee_ids: undefined, creator_ids: undefined` (the dropdowns write the plural params now, so those are what the invariant must clear). If filtering My Loop by another assignee/creator is intended, that needs an explicit, per-leg design rather than falling out of an un-stripped spread.

### Parity checked
The body claims "only the filter-state shape is unified; server-side + paginated filtering retained." Diffed base→head:
- **grouped** — the old explicit `statuses/priorities/assignee_types/project_ids/include_no_project` joins are replaced by `...arrayFilterQuery(params)`, which emits a **superset** of those keys → no capability dropped (it additionally forwards the new `assignee_ids/creator_ids/include_no_assignee/label_ids`, backend-supported per the body).
- **listIssues** — the singular `status/priority/assignee_id/creator_id/project_id` params are still sent alongside the new spread; IssuePage no longer populates them, but other callers (`mentionExtension`, `IssueDetailPage`) pass only `limit`, so they're vestigial-but-harmless (no collision — distinct keys from the plural set).
- **keyword search** still routes to the separate `searchIssues` endpoint; scope/filters correctly disabled while searching.
- **board** dropping the 「与我相关」tab is intentional (replaced by the cross-view scope). No capability silently lost.

### Minor
- `packages/dmloop/src/api/issueFilterQuery.ts` — `arrayFilterQuery` maps `[]` → `""` (empty string is truthy in a query object, so it *would* be sent). Currently safe because every caller pre-collapses empty arrays to `undefined` (reload lines 138-147; grouped is only ever called with an already-collapsed `common`), and the contract is documented. But it's a footgun for the next caller — a self-defensive `arr?.length ? arr.join(",") : undefined` inside the serializer would make the "never send empty" guarantee local rather than caller-dependent.

### Praise
- The `arrayFilterQuery` extraction + its query-contract test (`issueFilterQuery.test.ts`) is the right move — it locks the plural-join / boolean→"true" / unset→undefined contract in one place and kills the per-view drift that this PR is consolidating.
- `reload`'s `seq` guard (`my !== seq.current`) correctly discards stale in-flight responses on rapid filter/scope/page changes, and the pagination clamp (`page > maxPage → setPage(maxPage)`) handles the shrinking-result-set case cleanly.
- `noAssigneeActive` as a single source for send/checkbox/badge is a clean way to keep the "no-assignee only at scope=all" rule from drifting across the three sites.

Once the fan-out strips the plural user filters (or gains an explicit My-Loop filter design), this is ready.

Jerry-Xin (changes requested):
Flipping my APPROVE at this head to REQUEST_CHANGES. Peer review (OctoBoooot) surfaced a real My-Loop filter-pollution bug that this PR's own de-pollution logic misses — byte-confirmed below. My APPROVE checked scope-tab gating but missed that the plural assignee/creator array filters still reach the My-Loop fan-out.

## 🔴 Blocking — My-Loop `involves` fan-out leaks the plural user-relation filters into all three legs

`packages/dmloop/src/api/issueApi.ts` — `listMyGroupedIssues` builds `base = { ...params, assignee_types: undefined, assignee_id: undefined, creator_id: undefined, involves_user_id: undefined }` and fans out three legs (assignee_id / creator_id / involves_user_id = userId). The intent (per its own comment: "剥离所有「用户关系」过滤") is to strip every user-relation filter so each leg only scopes by the one userId it sets. **But the strip list only covers the SINGULAR fields — it does NOT strip the plural array fields `assignee_ids` / `creator_ids`**, which this PR introduced. Those arrive via `...params` and are re-emitted by the shared `arrayFilterQuery` (`issueFilterQuery.ts:14,16 → assignee_ids/creator_ids`) into every leg.

Reachability confirmed: the assignee/creator multi-selects ARE rendered in My Loop. Only the scope tabs are `!isMyLoop`-gated (`IssuePage.tsx:386`); the filter panel with the assignee select (`IssuePage.tsx:319`) and creator select (`:323`) is not `isMyLoop`-gated. So a My-Loop user who picks an assignee or creator filter gets those `assignee_ids`/`creator_ids` ANDed into all three union legs → e.g. the "creator = me" leg additionally requires `assignee_ids ∈ {picked}`, silently dropping issues that should appear. This defeats the union semantics the fan-out is built to provide.

Fix: add `assignee_ids: undefined, creator_ids: undefined` (and any other plural user-relation fields, e.g. if `involves`-style plurals exist) to the `base` strip in `listMyGroupedIssues`, mirroring the singular strip. A test asserting the fan-out legs carry no assignee_ids/creator_ids would lock it.

## 🟡 (non-blocking, converges with my prior note)
`arrayFilterQuery` empty-array handling — serialize empty arrays to `undefined` locally rather than risk an empty-string param. (My earlier 🟡; OctoBoooot raises the same.)

## ✅ The rest of the unification holds
Behavior preservation across board/list/grouped is otherwise sound: scope defaults preserved (`scope=all` → no narrowing), `scopeToAssigneeTypes` maps only to assignee-type buckets (no authz/visibility drift), no-assignee single-sourced via `noAssigneeActive = scope==="all" && f.noAssignee`. The bug is localized to the My-Loop fan-out's incomplete strip of the new plural fields.

Credit to the peer review for catching the My-Loop plural-field leak — I verified scope-tab gating but missed the assignee/creator array filters reaching the fan-out.

MrAladdin:
Thanks for the review — the blocking finding was real. Pushed `db83d203` addressing all points, plus one new capability.

**🔴 Blocking (OctoBoooot / Jerry-Xin) — My Loop involves fan-out leaked plural user-relation filters**
Fixed at the root: `listMyGroupedIssues` `base` now strips the **plural** `assignee_ids` / `creator_ids` too (not only the singular), so no user-relation filter reaches the three union legs. Belt-and-suspenders on the UI side: the assignee / no-assignee / creator filters are now **hidden on My Loop** (`!isMyLoop`-gated) — they conflict with the fixed `involves` scope and would only ever be no-ops there. `statuses`/`priorities`/`project_ids`/`label_ids`/`include_no_project` still flow through to narrow within My Loop, as intended.

**🟡 `arrayFilterQuery` empty-array footgun** — folded into the serializer: `csv(a) = a?.length ? a.join(",") : undefined`. The "never send empty" guarantee is now local, not caller-dependent.

**🟡 `activeFilters` counting non-keyword dims during search** — during search the non-keyword sum is forced to 0, so only the keyword term counts; the badge no longer implies scoped results when `/issues/search` ignores those filters (matches the scope-tab suppression).

**P2 singular/plural coexistence** — removed the now-dead singular `status`/`priority`/`assignee_id`/`creator_id`/`project_id` from `ListParams` and the `listIssues` query builder entirely (grepped: no caller sets them; issue list uses arrays only). No more dual-field ambiguity.

**Backend contract (P2)** — flat `/issues` plural-param support is delivered by octo-multica MR !55 (merged) + MR !57 (merged); each param's narrowing is covered by backend integration tests there and verified live against the deployed backend. Backend-first deploy ordering applies.

**New: keyword search now works in the grouped view** — `searchIssues` flat results are grouped client-side by assignee (`groupIssuesByAssignee`, unit-tested). Excluded on My Loop (the search endpoint can't scope to `involves`).

Verification: 63 unit tests green (incl. new `groupIssuesByAssignee` + `arrayFilterQuery` empty-array contract); tsc clean; live e2e on the dev stack (grouped keyword search groups by assignee; My Loop hides who-filters + keyword; badge=1 during search). Re-ran two-model adversarial after the fixes — both clean.

yujiawei (approved):
# Code Review — PR #776 (octo-web)

Reviewed at head `db83d203d6f0b12faeeba6b5ad43c8107af31325`. This is a re-review: the branch advanced past the initially-flagged revision, and the new head directly addresses the earlier `CHANGES_REQUESTED` feedback (My Loop filter pollution + grouped keyword search). Verdict below is issued against the current head only.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly the stated scope (unify board/list/grouped onto one array multi-select filter model + scope-across-views), and the follow-up commit closes the two items raised in the prior round:

- **Missing:** none. All six array dimensions (statuses / priorities / assignee_ids / creator_ids / project_ids / label_ids) plus the orthogonal no-assignee / no-project booleans are wired through `arrayFilterQuery` into all three views; scope (all/members/agents) now spans every view via `assignee_types`; label + no-assignee filters added.
- **Over-build:** none. The deferred "grouped keyword search" is now implemented in-scope (`issueGrouping.ts` + the `view === "grouped"` search branch), which is a reasonable inclusion rather than scope creep.
- **Deviation:** none blocking. The base intentionally hid the assignee / no-assignee / creator dropdowns on My Loop (involves); the current head restores that guard (`!isMyLoop` gates in the filter panel) after they were briefly exposed on an intermediate revision.

## 2. Code Quality

**Quality: Approved**

Verification on this head:
- `pnpm test` (dmloop): **63/63 passing**, 12 files (3 new tests for `groupIssuesByAssignee`, updated contract test for empty-array folding).
- Type check clean for all changed files (no new `tsc` errors introduced by the diff).
- Traced the full filter data-flow UI → `reload()` → `common` → `listIssues` / `listGroupedIssues` / `listMyGroupedIssues` → `arrayFilterQuery` → `http.clean()`. Correct on every path.

The key correctness concern from the prior head is resolved:

- **My Loop fan-out no longer pollutes the union.** `listMyGroupedIssues` now strips both the singular *and* the plural user-relation params in `base` (`assignee_ids: undefined, creator_ids: undefined` added alongside `assignee_id` / `creator_id` / `involves_user_id`). Combined with hiding the assignee / creator / no-assignee dropdowns on My Loop, the three-leg union (`assigned∪created∪involves`) can no longer be silently narrowed to a near-empty set by a stray array param ANDed into every leg. Good fix, and it matches the reasoning already documented in the surrounding comment.
- **Empty-array serialization hardened.** `arrayFilterQuery` now folds `[]` → `undefined` at the source via the `csv()` helper, so correctness no longer depends solely on `http.clean()` dropping empty strings downstream. The contract test locks this.
- **Singular `ListParams` fields removed cleanly.** Dropping `status` / `priority` / `assignee_id` / `creator_id` / `project_id` from `ListParams` removes the single-vs-array ambiguity. Verified the other `ListParams` consumers (`projectApi`, `skillApi`, `squadApi`, `agentApi`, `runtimeApi`) only read `keyword`, and the two other `listIssues` callers (`mentionExtension`, `IssueDetailPage`) pass only `limit` — none are affected.
- **Concurrency + edge cases hold.** The new grouped-search branch respects the `seq` guard (`if (my === seq.current)`), the page-clamp on shrinking result sets is intact, and error paths clear the result set + toast.

### Non-blocking suggestions (P2)

- **Grouped keyword search silently caps at 50.** The `view === "grouped"` search path fetches `searchIssues(gkw, { limit: 50 })` with no pagination and no "showing first N" affordance. On a busy workspace a keyword can match more than 50 issues and the grouped view will quietly omit the rest. The flat list view has the same 50-cap but at least paginates; consider surfacing an overflow hint here in a follow-up.
- **Route-reuse staleness (verified not a bug here, noted for awareness).** `view` / `scope` are `useState`-seeded from props and not re-synced on prop change. This is safe today only because `LoopPage` mounts each page with a distinct React `key` (`myloop:<ws>` vs `issue:<ws>`), forcing a remount across routes. If a future refactor ever renders `IssuePage` without that keying, the derived `isMyLoop` would diverge from the retained `scope`/`view`. Worth a guarding comment.

## 3. Overall verdict

**APPROVE**

Spec ✅ AND Quality Approved. All previously-blocking correctness issues are fixed at this head; the remaining items are advisory and do not block merge.

## 4. Suggested direction

The two P2s are follow-up-sized: (a) paginate or annotate the grouped keyword-search 50-cap; (b) add a one-line comment near the `useState` prop-seeding noting the remount-key dependency.

## 5. Additional notes

Coverage gaps to keep in mind (frontend-only review): the plural filter query params sent to the backend (`/issues`, `/issues/grouped`) are assumed honored per the cross-repo backend dependency the PR cites; runtime behavior of `include_no_assignee` on the flat `/issues` endpoint was not exercised here. Backend param handling and the search-index behavior behind `/issues/search` are out of this diff's scope.

Jerry-Xin (approved):
Re-verified at head db83d203. My prior blocking finding (My-Loop fan-out leaking the plural user-relation filters) is FIXED — flipping to APPROVE.

## ✅ Prior blocker — RESOLVED

**My-Loop `involves` fan-out now strips the plural user-relation filters.** In `listMyGroupedIssues`, the `base` object now clears all user-relation fields — both singular and plural: `assignee_id`/`assignee_ids`, `creator_id`/`creator_ids`, `involves_user_id`, plus `assignee_types`. Each of the three union legs then sets exactly one singular field (`assignee_id` / `creator_id` / `involves_user_id` = userId), so an assignee/creator filter selected in the My-Loop panel no longer ANDs into all three legs. This is exactly the one-line mirror the fix needed, and the code comment was updated to state the plural-strip intent ("复数 assignee_ids/creator_ids 同理必须剥"). The delta is a clean fast-forward (+1) touching only the loop filter/grouping files.

## ✅ Behavior preservation otherwise intact
The unification itself remains sound: scope defaults preserved (`scope=all` → no narrowing), `scopeToAssigneeTypes` maps only to assignee-type buckets (no authz/visibility drift), no-assignee single-sourced via `noAssigneeActive`. New tests added for grouping/filter-query. Code CI green (Build / secret-scan / osv-scan / dependency-review all pass).

## 🟡 (non-blocking, unchanged)
`arrayFilterQuery` empty-array handling — serialize empty arrays to `undefined` locally rather than rely on caller discipline (defense in depth). Still a footgun-hardening nit, not a blocker.

Clean fix. Credit to the peer review that first surfaced the plural-field leak.

OctoBoooot (approved):
## Re-review (DELTA) @ `db83d203d6f0` — my blocker + minor both fixed; new grouped keyword search reviewed and clean → APPROVE

One fix commit ("fix(loop): address review — My Loop filter pollution + grouped keyword search") since my CR head `748fee953c6b`. It resolves both my findings and adds the grouped keyword search that was deferred as follow-up. Reviewed the full delta (7 files, incl. 2 new), not just the fix. Flipping REQUEST_CHANGES → APPROVE.

### Blocker (My Loop involves fan-out leak) — FIXED, two layers
- **Data layer** (`issueApi.ts:133-139`): `base` now strips the plural fields too — `assignee_ids: undefined, creator_ids: undefined` alongside the singular `assignee_id`/`creator_id` and `assignee_types`. So the three union legs (`assignee_id=me` / `creator_id=me` / `involves=me`) no longer carry a panel-selected assignee/creator filter → union semantics restored. Verified against my failing input (My Loop → creator=Bob): the "created by me" leg is now `creator_id=me` only, no `creator_ids=[Bob]` AND-pollution.
- **UI layer** (`IssuePage.tsx:335,336,341`): the assignee / no-assignee / creator filters are now `!isMyLoop`-gated, so they aren't even reachable in My Loop. Defense-in-depth — the leak is closed at the source *and* the data layer. Either alone would fix it; both is belt-and-suspenders and the right call given the fan-out's fragility.

### Minor (empty-array serializer footgun) — FIXED
`issueFilterQuery.ts` now folds empty arrays locally via `const csv = (a) => a && a.length ? a.join(",") : undefined`, applied to every array dimension. The "never send an empty string" guarantee is now internal to the serializer, not dependent on callers pre-collapsing. Matches what I suggested.

### New surface reviewed — grouped keyword search (`issueGrouping.ts` + wiring)
Deferred-as-follow-up in the original body, now included. Reviewed it as new code:
- `groupIssuesByAssignee` is a pure function: groups flat search results by `(assignee_type, assignee_id)`, preserves first-seen (relevance) order, folds unassigned into a single null group, sets `total` per group. Well-tested (`issueGrouping.test.ts`: grouping+order+count, null-fold, empty→[]).
- Wiring (`IssuePage.tsx:159-165`): grouped + keyword + `!isMyLoop` → `searchIssues(gkw, {limit:50})` → `groupIssuesByAssignee`. Correctly seq-guarded (`my === seq.current`), error-handled (`onErr`), and routed *before* the involves/scope branches so search bypasses filters as designed. My Loop is excluded (search doesn't honor `involves`), consistent with hiding the keyword input there.
- The `searching` refactor (now `!isMyLoop && !!f.keyword.trim()`, extended to grouped) correctly disables scope tabs + panel filters across all three views during search, and `activeFilters` no longer counts search-ignored filters (`searching ? 0 : …`) — this also resolves the earlier badge-highlight 🟡.

### Parity / no regression
- `listIssues` dropped the vestigial singular `status/priority/assignee_id/...` params (IssuePage no longer populates them; other callers pass only `limit`) — no capability lost, and it removes the single/plural coexistence confusion noted earlier.
- grouped's `...arrayFilterQuery` still a superset of the pre-unification joins. Keyword→search endpoint unchanged. Defaults preserved (`scope=all` → no narrowing).

### CI
Code CI is green at this head — Build / secret-scan (gitleaks) / osv-scan / dependency-review / pr-title-lint all pass.

### Net
Both my findings fixed (blocker two ways, minor via local guard), the new grouped-keyword-search is correct and tested, no parity regression, CI green. **APPROVE @`db83d203d6f0`.** Credit to the author for fixing the leak at both the data and UI layers rather than the minimum, and for folding in the deferred grouped-search cleanly. Remaining notes are non-blocking follow-ups others raised (grouped search silently caps at 50 → an overflow hint would be nice; backend plural-param support must ship first per the documented deploy dependency).

## Parity checked
Diffed base→head for the "only the filter-state shape is unified" claim: grouped emits a superset of its former params via `arrayFilterQuery`; `listIssues` keyword still routes to the separate search endpoint; `scope=all` still yields no narrowing (default behavior preserved); board's dropped 「与我相关」tab is intentionally replaced by the cross-view scope. No capability silently removed. The new grouped keyword search is additive.

PR #779, opened 2026-07-15
fix(loop): resolve archived AI teammate names instead of showing raw id

### LINKED ISSUE #777: fix(loop): deleted AI teammate shows raw id instead of name in issues
## Problem

After an AI teammate (agent) is **deleted** from a workspace, loop surfaces that reference it stop showing its **name** and instead render its raw **UUID**. Seen on: comment authors, the issue activity timeline, and automation cards. It looks like a random id string where a name should be.

## Root cause

Agent "deletion" is a **soft-delete (archive)** — the row is kept with `archived_at` set (`ArchiveAgent`); there is no hard delete. So the name is still in the DB and recoverable.

The loop frontend does **not** get actor names from the API (issue/comment responses carry no `assignee_name`/`author_name`). Instead it resolves them client-side via a cached **directory** (`packages/dmloop/src/api/directory.ts`) built from `GET /agents`, `GET /squads`, `GET /workspaces/:id/members`. That agent list is fetched **without `include_archived`**, so it **excludes archived agents**. `actorName("agent", id)` then returns `null` for a deleted agent, and every display site that falls back with `?? <id>` leaks the UUID:

- `packages/dmloop/src/api/issueApi.ts` — `author_name: actorName(...) ?? c.author_id` (comment author)
- `packages/dmloop/src/panel/IssueDetailPage.tsx` — `{a.actor_name ?? a.actor_id}` (activity timeline)
- `packages/dmloop/src/pages/AutomationPage.tsx` — `{a.assignee_name ?? a.assignee_id}` (automation card)
- (board/group badge shows `—` rather than the id, but is still nameless)

The same class applies to **squads** (also soft-deleted via `ArchiveSquad`) and, by the same mechanism, any actor missing from the live directory.

## Solution

Fix the class **at the resolution layer** (one change covers every `?? id` site), not per display site:

**1. Agents — frontend only (backend already supports it):**
In `directory.ts` `build()`, fetch `GET /agents?include_archived=true`. Add **all** returned agents (incl. archived) to the `agentName` map so names always resolve. Keep the assignee **picker** correct by pushing **only non-archived** agents into `candidates` (you should not be able to newly assign to a deleted agent) — filter on the response's `archived_at` field.

**2. Squads — needs a small backend change first (cross-repo):**
`ListSquads` (octo-multica, `server/pkg/db/queries/squad.sql`) hard-filters `archived_at IS NULL` and has no `include_archived` param, so archived squad names cannot be recovered from the frontend yet. Add an `include_archived` query param to `ListSquads` (mirroring `ListAgents`), then apply the same directory treatment for `/squads` (resolve all, offer only active in the picker). Tracked separately in octo-multica.

## Acceptance

- A deleted (archived) agent's name still renders in comment authors, the activity timeline, automation cards, and the assignee/creator badges — no raw UUID anywhere.
- The assignee/creator pickers do **not** list deleted agents.
- Same for archived squads once the backend param lands.

MrAladdin:
## Summary

After an AI teammate (agent) is **deleted** (soft-delete / archive) from a workspace, loop surfaces that reference it — comment authors, the activity timeline, automation cards, assignee/creator badges — stopped showing its **name** and rendered its raw **UUID** instead. Closes #777.

## Root cause

Loop resolves actor names client-side via a cached **directory** (`packages/dmloop/src/api/directory.ts`) built from `GET /agents`. That list excludes archived agents, so a deleted agent falls out of the `agentName` map, `actorName("agent", id)` returns `null`, and every `?? <id>` display fallback leaks the UUID.

Agent deletion is a **soft-delete** (`archived_at`, via `ArchiveAgent`) — the name is still in the DB and recoverable via `include_archived`.

## Fix

`directory.ts` `build()` now fetches `GET /agents?include_archived=true` and adds **all** agents (incl. archived) to the `agentName` map so names always resolve, while pushing **only active** agents (`!archived_at`) into the assignee/creator picker `candidates` (you can't newly assign to a deleted agent). One resolution-layer change covers every display site (Rule 9) — no per-site `?? id` patching.

Backend already supports `include_archived` on `/agents` (mirrors `ListAgents`).

## Scope

- **Agents only.** Squads have the same class of bug, but `ListSquads` has no `include_archived` param yet — tracked in **octo-multica#20**; the squad frontend follow-up lands together with that backend change (not speculatively wired here, to avoid a client contract that the backend doesn't honor).

## Verification

- **Live e2e** (dev stack, Alice-Loop): archived the E2E Coworker agent → confirmed it drops from default `GET /agents` (so without this fix its issues show the raw id) and is returned by `GET /agents?include_archived=true` with its name; with the fix the board still renders "E2E Coworker" and **no raw UUID appears anywhere**; restored the agent afterwards.
- `pnpm test` (dmloop): 63 tests green; tsc clean for the changed file.
- Two-model adversarial: skeptical code-reviewer — no real bugs (verified the picker still excludes archived, name maps include archived, `archived_at` truthiness, httpGet query serialization, cache untouched). No finding on this diff.

## Blast radius (Rule 8)

- `/agents` is fetched only in `directory.ts` `build()`; the added `archived_at` field is read locally and never escapes into `Directory`/`agentName`/`candidates` shapes, so `actorName`, `listAssigneeCandidates`, `agentApi.ts`, `squadApi.ts` consumers are unaffected. No brand leak.

### DIFF
--- a/packages/dmloop/src/api/directory.ts
+++ b/packages/dmloop/src/api/directory.ts
@@ -30,7 +30,10 @@ async function build(): Promise<Directory> {
   const wsId = currentWorkspaceId();
   const [members, agents, squads, projectsResp] = await Promise.all([
     wsId ? httpGet<Array<{ user_id: string; name: string; octo_uid?: string | null }>>(`/workspaces/${wsId}/members`).catch(() => []) : Promise.resolve([]),
-    httpGet<Array<{ id: string; name: string }>>("/agents").catch(() => []),
+    // include_archived: resolve names for archived (soft-deleted) agents too, so issues/comments/
+    // timeline referencing a deleted AI teammate show its name, not a raw id. Archived entries feed
+    // the name map only; the assignee picker (candidates) stays active-only.
+    httpGet<Array<{ id: string; name: string; archived_at?: string | null }>>("/agents", { include_archived: true }).catch(() => []),
     httpGet<Array<{ id: string; name: string }>>("/squads").catch(() => []),
     httpGet<{ projects: Array<{ id: string; title: string }> }>("/projects").catch(() => ({ projects: [] })),
   ]);
@@ -54,8 +57,8 @@ async function build(): Promise<Directory> {
   }
   const agentName = new Map<string, string>();
   for (const a of agents) {
-    agentName.set(a.id, a.name);
-    candidates.push({ id: a.id, type: "agent", name: a.name });
+    agentName.set(a.id, a.name); // resolve names incl. archived (for display)
+    if (!a.archived_at) candidates.push({ id: a.id, type: "agent", name: a.name }); // picker: active only
   }
   const squadName = new Map<string, string>();
   for (const s of squads) {


Jerry-Xin (approved):
APPROVE — byte-verified clean, small and correct fix.

## What it does
`packages/dmloop/src/api/directory.ts` — fetches `/agents` with `{ include_archived: true }` so soft-deleted AI teammates still resolve to their name in issues/comments/timeline (instead of showing a raw id), while keeping the assignee picker active-only.

## Correctness — verified
- **Query param is actually sent.** `httpGet<T>(path, params?)` forwards `{ params: clean(params) }` to the client, so `include_archived: true` is serialized onto the request (not dropped). The `.catch(() => [])` fallback is preserved, so if the backend doesn't yet honor the param it degrades to an empty list — no crash.
- **Active-only invariant for the picker is preserved.** Archived agents feed only the `agentName` display map (`agentName.set(a.id, a.name)` unconditional). The `candidates` array — the assignee picker source — now guards `if (!a.archived_at)`, so a deleted agent resolves for display but can't be newly assigned. This matches the intent and prevents assigning issues to archived teammates.
- **No null hazard.** `archived_at?: string | null` optional; `!a.archived_at` correctly treats missing/null as active.

## CI
secret-scan / osv-scan / dependency-review / pr-title-lint all green; Build is still `in_progress` at this head — merge is gated on it concluding green (mergeStateStatus will reflect it). The code change itself is trivially safe (additive param + guard, catch-fallback intact).

Clean fix, no regression. Nice touch keeping the picker active-only while resolving archived names for display.

OctoBoooot (approved):
## Review: resolve archived AI teammate names instead of raw id (#779)

**Verdict:** Approve — minimal, correctly-scoped fix at the right layer; verified no display regression from the `candidates` change. Code CI green.

A 6-line change that fixes #777 (archived agents rendering their raw UUID) at the single resolution layer rather than patching each `?? id` display site — the right call.

### Why it's correct (verified end-to-end)
- **Name map includes archived, picker excludes it** (`directory.ts:58-62`): `agentName.set(a.id, a.name)` for *all* agents (so names always resolve), but `candidates.push(...)` only when `!a.archived_at` (so you can't newly assign to a deleted agent). Exactly the right split.
- **Query param actually reaches the backend** — `httpGet("/agents", { include_archived: true })` flows through `clean()` (`http.ts:93`), whose guard `v !== undefined && v !== null && v !== ""` keeps boolean `true` and emits `String(true) → "true"`. So the param is sent, not silently dropped. (This was the one mechanism worth checking; it holds.)
- **No `archived_at` leak** — the field is read locally in `build()` and never enters the `Directory` / `agentName` / `candidates` shapes, so `actorName`, `listAssigneeCandidates`, and the `agentApi`/`squadApi` consumers are untouched. Body's blast-radius claim confirmed.

### Parity checked — the `candidates` change does not regress any display site
The risk in moving archived agents out of `candidates` is a display path that resolves a *current* assignee via `candidates` rather than the name map. Traced every consumer:
- `AssigneePicker` shows the selected assignee as `current?.name ?? valueName`, `current = cands.find(c => c.id === value)`. Post-fix an archived agent isn't in `cands` → `current` is undefined → it falls back to `valueName`.
- **Both callers pass a resolved `valueName`** for real rows: `IssueDetailPage:1130` (`issue.assignee_name`) and `IssueList:203` (`r.assignee_name`) — and `assignee_name` is enriched via `actorName(dir, ...)` (`issueApi.ts:24`), which reads the `agentName` map that **now includes archived**. So the archived name resolves and displays.
- The one `valueName={null}` usage (`IssueList:136-139`) is the **batch-action toolbar** picker (a blank "assign selected to…" control) — it displays no existing assignee, so `null` is correct, not a regression.
Net: every display site resolves the archived name through `agentName`; `candidates` is only ever consumed for the picker's selectable list, where excluding archived is the intended behavior.

### Praise
- Fixing this once in `directory.build()` instead of touching every `?? id` fallback is the correct instinct — one resolution layer, all display sites covered.
- The scope note is honest and disciplined: squads have the same bug but `ListSquads` has no `include_archived` yet, so the squad fix is deferred to land with the backend change (octo-multica#20) rather than shipping a client contract the backend won't honor. That's the right way to avoid a half-working feature.

### Out of scope (informational, non-blocking)
- Squads keep the raw-id bug until octo-multica#20 lands — already tracked and called out in the PR body.

Clean fix. CI green, no regression, correctly scoped.

## Parity checked
Diffed base→head for the "no consumer affected / one resolution-layer change" claim: `agentName` now resolves archived agents (additive — previously returned null for them); `candidates` loses archived agents, but every display path resolves names via `agentName`/`actorName`, not `candidates` (the picker's `valueName` fallback is fed by enriched `assignee_name`), and the only `valueName={null}` picker is a blank batch-assign control. No display capability removed; the assignee-picker correctly no longer offers deleted agents as new targets.

yujiawei (approved):
# Code Review — PR #779 (octo-web)

## Summary

This fixes a real display bug: after an AI teammate is archived (soft-deleted), loop surfaces referencing it rendered the raw UUID instead of the name, because the client-side directory was built from `GET /agents`, which excludes archived agents. The fix fetches `GET /agents?include_archived=true` and adds all agents (including archived) to the name map, while keeping the assignee picker `candidates` active-only.

The change is minimal (+6/-3, one file) and well-targeted. I verified the diff against the surrounding code and the backend contract.

## Verification

**1. Query serialization is correct.** `httpGet("/agents", { include_archived: true })` flows through `clean()` in `http.ts`, which stringifies `true` → `"true"` and forwards it as a query param via axios. The identical pattern is already used by `listAgents` in `agentApi.ts` (`{ include_archived: params?.includeArchived ? true : undefined }`), so the `/agents` endpoint's `include_archived` support is an established contract, not a new assumption.

**2. No leak of the new field.** The added `archived_at` is read only locally inside `build()`. It never enters the `Directory`, `agentName`, or `candidates` shapes, so downstream consumers (`actorName`, `listAssigneeCandidates`, `agentApi`, `squadApi`) are unaffected.

**3. Assignee picker stays active-only.** `if (!a.archived_at) candidates.push(...)` correctly keeps archived agents out of the picker — you can't assign new work to a deleted agent, but its name still resolves for display. This is the right split.

**4. One resolution-layer change covers every display site.** `actorName` reads the `agentName` map and is consumed across `issueApi` (assignee/creator/comment author), `collabApi` (timeline), `autopilotApi`, and `squadApi`. Comment authors additionally have a `?? c.author_id` fallback, which now resolves to the real name. No per-site patching needed.

**5. Scope discipline.** Squads have the same class of bug but are explicitly deferred (backend `ListSquads` lacks `include_archived`, tracked separately). Not speculatively wiring a client contract the backend doesn't honor is the correct call. `squadApi` retains its `?? member_id` fallback for that case, which is acceptable given the documented follow-up.

## Findings

No P0/P1 issues. The fix does exactly what the linked issue asks, with no over-reach.

### Nits (non-blocking)

- **P2 — payload growth over time.** `GET /agents?include_archived=true` returns the full archived history on every directory rebuild, and the archived set only grows. In practice the directory is cached and rebuilt only on mutation/workspace switch, so this is a minor concern, but if a workspace accumulates a large number of archived agents over its lifetime, the name-map build fetches all of them just to resolve occasional historical references. A future optimization could resolve archived names lazily per-id, but that is out of scope for this fix.
- **P2 — test coverage.** No unit test asserts the archived-name-resolves / picker-excludes-archived behavior. The PR relies on a live e2e run, which is fine for this size, but a small directory test guarding "archived agent → name in map, absent from candidates" would prevent a future refactor from silently regressing the picker split.

## Verdict

Correct, minimal, and well-scoped. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #779 (octo-web)
Head SHA: `7f10bb9cfd8d435a7647f49a0db5b51ed93f30f6`
Base: `main` (merge-base `578dac69`)

## Summary

This PR's title targets a narrow fix — resolving archived AI teammate names instead of showing raw UUIDs — but the diff (10 files, +329/−154) delivers a much broader change: a full unification of the filter system from per-view single/multi-select to a single multi-select model across board/list/grouped views, plus new label filtering, keyword search in the grouped view (client-side grouping of flat search results), and locking the "My Loop" page to grouped view only.

The core fix in `packages/dmloop/src/api/directory.ts` is correct and well-scoped: `include_archived: true` is passed to `GET /agents`, archived agents populate the `agentName` map for display, and the assignee picker (`candidates`) correctly excludes them via `!a.archived_at`. The broader refactoring is well-structured with good helper extraction (`arrayFilterQuery`, `groupIssuesByAssignee`) and corresponding unit tests.

## Verification

- ✅ **Archived agent name resolution** — `directory.ts:33` adds `include_archived: true` to the `/agents` request. `httpGet` at `packages/dmloop/src/api/http.ts:102` serializes `{ include_archived: true }` → `"true"` via `clean()`, producing `?include_archived=true`. The `agentName` map now includes archived agents; `candidates` (picker) excludes them via `!a.archived_at` at `directory.ts:61`. Data flow verified end-to-end.
- ✅ **Filter query serialization** — `issueFilterQuery.ts` correctly folds empty arrays to `undefined` (never sent) and maps booleans to `"true"`. Unit tests in `__tests__/issueFilterQuery.test.ts` lock both the happy-path and empty/undefined contracts.
- ✅ **Client-side grouping** — `groupIssuesByAssignee` groups by `(assignee_type, assignee_id)` key, preserves first-seen order, handles null assignees into a single group. Unit tests in `__tests__/issueGrouping.test.ts` cover grouped, unassigned, and empty inputs.
- ✅ **Error handling** — New `onErr` callback at `IssuePage.tsx:126` clears results and shows toast on failure. `Toast` imported from `@douyinfe/semi-ui` at `:11`; `loop.toast.loadFailed` key exists in both locale files. All three async paths (grouped, board/list, search) wire `.catch(onErr)`.
- ✅ **`listMyGroupedIssues` param stripping** — `issueApi.ts:131` now strips both singular (`assignee_id`, `creator_id`) and plural (`assignee_ids`, `creator_ids`) forms before the three-leg fan-out, preventing AND-pollution into the union semantics.
- ✅ **Seq guard** — `seq.current` counter correctly gates all `setState` calls across async paths; stale responses are discarded.

Static analysis only at head `7f10bb9c`; build and tests not executed in this environment.

## Findings

No P0/P1 blockers. Two P2 items and one nit below.

### P2 — Filter params sent to flat `/issues` endpoint may be silently ignored (`packages/dmloop/src/pages/IssuePage.tsx:155`)

The `common` object (applied to all three views) includes `include_no_project`, `include_no_assignee`, `label_ids`, and `assignee_types`. These are spread into `listIssues` which calls `GET /issues`. The old code comment at this site explicitly noted `include_no_project` was "仅 /grouped 支持" (only supported by `/issues/grouped`). If the backend `/issues` flat endpoint hasn't been updated to honor these parameters, the "no project" checkbox, "no assignee" checkbox, label filter, and scope tabs (All/Members/AI) will appear active in board/list views but have no effect on query results — a visible UX mismatch where the UI implies filtering that doesn't happen.

**Diff-scope**: new — this PR broadens the filter surface to board/list views that previously never sent these params.

**Fix direction**: Confirm the backend `/issues` endpoint supports `include_no_project`, `include_no_assignee`, `label_ids`, and `assignee_types`. If not, gate these controls to grouped view only (restoring the old per-view visibility logic), or add a backend change to the PR description.

### P2 — Grouped keyword search has a hard 50-result cap with no truncation signal (`packages/dmloop/src/pages/IssuePage.tsx:143`)

When keyword is set in grouped view, `searchIssues(gkw, { limit: 50 })` returns at most 50 results, which are then grouped client-side. There is no pagination and no indicator that results may be truncated. Board/list views show a `total` count; the grouped view surfaces neither total nor a "showing X of Y" hint. Users with >50 matching issues see a silently incomplete grouped board.

**Diff-scope**: new — keyword search in grouped view is added by this PR.

**Fix direction**: Surface `total` from `searchIssues` response and render a truncation hint when `total > displayed issue count`.

### Nit — PR scope mismatch with title

The PR title and description focus on a 9-line fix in `directory.ts` (archived agent name resolution), but the diff delivers a 300+ line refactoring of the entire filter system. These are distinct concerns — the name fix is a clear bug-fix while the filter unification is a feature/refactor. Bundling them makes the PR harder to review, bisect, and revert independently. Consider splitting in the future.

## Data Flow Tracing

| Consumed Data | Upstream Source | Reaches Consumer? |
|---|---|---|
| `agentName` map entries for archived agents | `GET /agents?include_archived=true` → `httpGet` → `clean()` → query string | ✅ Yes — `httpGet` serializes `{ include_archived: true }` to `"true"` string |
| `candidates` (picker) excludes archived | `agents[]` → `if (!a.archived_at)` gate at `directory.ts:61` | ✅ Yes — `archived_at` truthy for archived agents, correctly skipped |
| `common` filter object → board/list/grouped | `IssuePage.tsx:148` builds `common` → spread into `listIssues` / `listGroupedIssues` / `listMyGroupedIssues` | ✅ Yes — all three paths consume it; `listMyGroupedIssues` strips user-relation fields before fan-out |
| `groupIssuesByAssignee` input issues | `searchIssues` → `fetchIssues` → `enrichWith` (fills `assignee_name`/`assignee_avatar` from directory) | ✅ Yes — enriched issues carry name/avatar for grouping headers |
| `noAssigneeActive` → query param | `scope === "all" && f.noAssignee` → `common.include_no_assignee` | ✅ Yes — only sent when scope=all; checkbox disabled otherwise |

- **C1 — Dual-path parity**: Scope tabs (All/Members/AI) now apply across all three views via `scopeToAssigneeTypes`. The function is unchanged and called symmetrically in board/list (`:196`) and grouped (`:173`) paths. ✅ Clear.
- **C2 — Control-flow ordering**: `reload` callback has nested conditionals (grouped→keyword→involves→regular; board/list→keyword→listIssues). No double-execution paths; each branch returns early. ✅ Clear.
- **C3 — Authorization boundary**: No auth changes. N/A.
- **C5 — Build/note**: No build config changes. N/A.
- **C6 — Governance/docs**: N/A.

## Verdict

No P1 blockers found. The core fix (archived agent name resolution) is correct and well-isolated. The broader filter refactoring is well-structured with proper helper extraction and test coverage. The two P2 findings concern backend contract verification and a truncation indicator — neither blocks merge if the backend already supports the new params on the flat `/issues` endpoint.

[Octo-Q] verdict: APPROVE — No P1 found. Core archived-agent name fix is correct; filter unification is well-engineered. Two P2 items (backend param support verification, grouped search truncation indicator) are non-blocking but worth addressing.

MrAladdin:
Rebased onto latest `main`. Re-ran the gauntlet.

**Live failure-injection e2e (dev):** aborted `GET /agents?include_archived=true` at the XHR layer and rebuilt the directory. Result: page does **not** crash, issues still load, assignee badges fall back to `—` (not a raw id — the board `AssigneeBadge` renders `—` for a null name), no raw UUID visible. This is graceful degradation **identical to the pre-change behavior** — the `.catch(() => [])` on the `/agents` fetch predates this PR; the change only appends `include_archived` to the same call, so the failure mode is unchanged.

**Re: the adversarial [medium] note** (archived lookup on the shared directory hot-path): acknowledged as a considered tradeoff, not a regression —
- The failure → active-agents-empty behavior is pre-existing (unchanged `.catch`), and injection-confirmed graceful (no crash, no id leak on the board).
- Payload/latency: archived AI teammates are a small, human-curated set per workspace (you archive a handful, not thousands); the directory is cached per workspace. The "unbounded history" concern is theoretical at this scale.
- The heavier alternatives (backend-embedded actor names, an ID-scoped name lookup, or a separate non-blocking archived-name cache) are deferred — they'd be the right move only if archived counts grow large enough to matter.

Scope unchanged: agents only; squads await octo-multica#20's `include_archived` before their frontend follow-up.

yujiawei (approved):
# Code Review — PR #779 (octo-web)

## Summary

This is a tight, well-scoped fix for #777: after an AI teammate is soft-deleted (archived), loop surfaces that reference it (comment authors, timeline, badges) rendered the raw UUID because the client-side directory only loaded active agents. The fix widens the `/agents` fetch to `include_archived: true`, adds archived entries to the display name map, and keeps the assignee picker active-only.

## Spec compliance

**Verdict: PASS**

- **Under-build:** none. The name map now resolves archived agents, which is exactly what #777 needs. One resolution-layer change covers every display site.
- **Over-build:** none. The `if (!a.archived_at)` guard on `candidates` is the required companion to the widened fetch (you must not be able to newly assign to a deleted agent), not scope creep. Squads have the same latent bug but are correctly deferred to the tracked backend follow-up (`ListSquads` lacks the param), rather than speculatively wiring a client contract the backend won't honor.
- **Deviation:** none.

## Code quality

**Verdict: Approved**

Verified against the codebase at the head SHA:

1. **Param serialization is correct.** `httpGet(path, { include_archived: true })` flows through `clean()` (`http.ts`), which does `String(v)` for non-null values → `include_archived=true`. `include_archived` is an established, backend-supported param — `agentApi.ts` `listAgents` already uses it.
2. **Truthiness guard is correct.** `archived_at` is `string | null` where non-null = archived (per the type comment). `!a.archived_at` is truthy exactly for active agents. No empty-string edge case (backend contract sets a timestamp).
3. **No blast radius / no leak.** `agentName` is consumed only via `.get(id)` lookups (`actorName`, `runsApi.ts` backfill) — it is never iterated to produce a list, so adding archived entries cannot leak deleted agents into any list, filter, picker, or count. `candidates` is the sole list-producing structure and correctly stays active-only. Confirmed the two candidate consumers (`SquadPage` agent picker, assignee dropdowns) build from `candidates`, and `AgentPage` counts use their own `listAgents` fetch with independent `archived_at` filtering — all unaffected.
4. **Backend field contract confirmed.** `AgentPage.tsx` already relies on `archived_at` being populated on archived rows (`archived: rows.filter((a) => !!a.archived_at).length`), so the guard's correctness rests on an already-proven contract, not an assumption.

- **No regression test for the active-only picker invariant.** The `if (!a.archived_at)` line is now load-bearing (it's the only thing keeping soft-deleted agents out of the picker), but no unit test in `directory.ts` asserts it. A small test feeding one archived + one active agent and asserting `candidates` contains only the active one — plus `actorName` resolving the archived agent's name — would lock the intent against a future refactor. Not a blocker.

## Verdict

**APPROVED.** Correct, minimal, well-reasoned fix with clean separation between the display name map (includes archived) and the assignee picker (active-only). No P0/P1 issues.

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Data flow tracing confirms archived agent names resolve correctly via agentName map; candidates picker correctly excludes archived agents via `!a.archived_at` gate; cache invalidation works through existing afterAgentMutation chain.

PR #784, opened 2026-07-15
feat(mention): support pinyin member search

### LINKED ISSUE #783: 对话框 @人 需要支持拼音匹配
## 问题描述

在对话框中使用 @ 提及用户时，目前只能通过中文名称进行匹配。

## 期望行为

支持拼音匹配，例如输入 `@zhangsan` 能匹配到「张三」。

## 优先级

常规需求

botshen:
## Summary

- keep the existing case-insensitive member-name filtering
- add full-pinyin, partial-pinyin, and initials matching for Chinese member names
- declare `pinyin-pro` directly in `@octo/base` and cover the behavior with unit tests

Closes #783

## Testing

- `pnpm --filter @octo/base exec vitest run src/Utils/__tests__/mentionRender.test.ts`
- `pnpm --filter @octo/web exec vitest run src/__tests__/mentionRefactor.test.ts`

### DIFF
--- a/packages/dmworkbase/package.json
+++ b/packages/dmworkbase/package.json
@@ -45,6 +45,7 @@
     "lucide-react": "^0.577.0",
     "mitt": "^3.0.1",
     "moment": "^2.29.3",
+    "pinyin-pro": "^3.28.0",
     "qrcode.react": "^4.2.0",
     "react": "^17.0.2",
     "react-avatar-editor": "^13.0.0",
--- a/packages/dmworkbase/src/Utils/__tests__/mentionRender.test.ts
+++ b/packages/dmworkbase/src/Utils/__tests__/mentionRender.test.ts
@@ -9,6 +9,7 @@ import {
 const members = [
   { uid: "u1", name: "Alice" },
   { uid: "bot1", name: "BuildBot", orgData: { robot: 1 } },
+  { uid: "u2", name: "张三" },
 ];
 
 const baseArgs = {
@@ -30,6 +31,7 @@ describe("buildMentionDropdownItems", () => {
       MENTION_UID_AIS,
       "u1",
       "bot1",
+      "u2",
     ]);
   });
 
@@ -40,7 +42,7 @@ describe("buildMentionDropdownItems", () => {
       includeBroadcastMentions: false,
     });
 
-    expect(items.map((item) => item.uid)).toEqual(["u1", "bot1"]);
+    expect(items.map((item) => item.uid)).toEqual(["u1", "bot1", "u2"]);
   });
 
   it("keeps broadcast mentions hidden while filtering members", () => {
@@ -51,6 +53,18 @@ describe("buildMentionDropdownItems", () => {
 
     expect(items.map((item) => item.uid)).toEqual(["u1"]);
   });
+
+  it.each(["zhangsan", "ZHANGSAN", "san", "zs"])(
+    "matches Chinese member names with the pinyin query %s",
+    (query) => {
+      const items = buildMentionDropdownItems({
+        ...baseArgs,
+        query,
+      });
+
+      expect(items.map((item) => item.uid)).toEqual(["u2"]);
+    },
+  );
 });
 
 describe("mentionUidStateFromRobot", () => {
--- a/packages/dmworkbase/src/Utils/mentionRender.ts
+++ b/packages/dmworkbase/src/Utils/mentionRender.ts
@@ -1,3 +1,5 @@
+import { match } from "pinyin-pro";
+
 /**
  * Shared render-side helpers for the three-state mention model
  * (`@所有人` / `@所有AI` / member, with legacy `mention.all=1` support).
@@ -257,9 +259,14 @@ export function buildMentionDropdownItems<
     };
   });
 
-  const filteredMembers = items.filter((item) =>
-    item.name.toLowerCase().includes(trimmedQuery.toLowerCase()),
-  );
+  const normalizedQuery = trimmedQuery.toLowerCase();
+  const filteredMembers = items.filter((item) => {
+    const nameMatches = item.name.toLowerCase().includes(normalizedQuery);
+    const pinyinMatches =
+      normalizedQuery.length > 0 && match(item.name, normalizedQuery) !== null;
+
+    return nameMatches || pinyinMatches;
+  });
 
   return [...stickyTop, ...filteredMembers];
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -552,6 +552,9 @@ importers:
       moment:
         specifier: ^2.29.3
         version: 2.30.1
+      pinyin-pro:
+        specifier: ^3.28.0
+        version: 3.28.0
       qrcode.react:
         specifier: ^4.2.0
         version: 4.2.0(react@17.0.2)


OctoBoooot (commented):
## Review: pinyin member search in mentions (#784)

**Verdict:** Clean on my read — holding at COMMENT only because the required **Build** check is still running (not concluded green), so I don't APPROVE over unconfirmed code CI. No blockers, no majors. Flip to APPROVE the moment Build lands green.

A tight, well-scoped feature: keep the existing case-insensitive name filter, add pinyin (full / partial / initials) matching for Chinese names via `pinyin-pro`, with tests.

### Why it's correct (verified)
- **`match` API used correctly** — `pinyin-pro`'s `match(text, pinyin, options?) => number[] | null` (verified against the installed 3.28.0 types); the code checks `match(item.name, normalizedQuery) !== null`, the right null-test for "matched." Default `precision: "start"` is the sensible choice (syllable-start, not the looser `"any"`).
- **No false-positive pollution** — `match` runs on every member name including Latin ones, but `match("Alice", "zs")` returns `null` (nothing to romanize), so Latin members aren't wrongly surfaced by a pinyin query. The test locks this: queries `["zhangsan","ZHANGSAN","san","zs"]` all resolve to `["u2"]` only, with Alice/BuildBot excluded.
- **Empty-query guard** — `normalizedQuery.length > 0 && match(...)` short-circuits, so an empty query keeps the prior behavior (no `match` call, sticky items + all members via the unchanged path).
- **Dependency hygiene** — `pinyin-pro@^3.28.0` is declared in `@octo/base` (`packages/dmworkbase`), the same package as `mentionRender.ts` that uses it (correct placement); it resolves to the *same* 3.28.0 already in the lockfile for another importer — no version conflict, no duplicate install. `dependency-review` and `osv-scan` are green on the added dep.
- **Body/scope accurate** — "declare pinyin-pro directly in @octo/base" matches (`dmworkbase` package name is `@octo/base`).

### Minor (non-blocking)
- `match` is invoked per-member per-keystroke in addition to the existing `.includes()`. Fine here (member lists are small and this mirrors the existing filter cost), but if a workspace ever has a very large roster, the pinyin pass is the more expensive half — worth a mental note, not a change.

### Praise
- The test is the right shape: `it.each(["zhangsan","ZHANGSAN","san","zs"])` covers full-pinyin, case-insensitivity, partial, and initials in one table, and the assertions prove *both* that Chinese names match *and* that Latin members aren't polluted — the exact correctness edge that matters for a fuzzy matcher.
- Declaring the dep in the package that consumes it (rather than relying on a hoisted transitive) is the disciplined choice.

Clean small feature. Holding at COMMENT purely on Build not yet being green; no code concerns.

Jerry-Xin (approved):
APPROVE — byte-verified clean, small and correct feature.

## What it does
`packages/dmworkbase/src/Utils/mentionRender.ts` — adds pinyin matching to the @mention member dropdown filter, so Chinese member names can be found by typing pinyin (full / uppercase / partial / initials). Uses `pinyin-pro`'s `match()`.

## Correctness — verified
- **`match()` API usage is correct.** Verified against `pinyin-pro@3.28.0`'s own type defs: `match(text, pinyin, options?) => number[] | null` — returns an index array on a successful match, `null` on no match. The code's `match(item.name, normalizedQuery) !== null` is exactly the intended usage.
- **Existing behavior preserved (additive).** The prior substring filter (`item.name.toLowerCase().includes(normalizedQuery)`) is kept as the first branch of the OR, so non-Chinese/Latin names filter identically to before. Pinyin matching only adds hits; it never removes a name that used to match. Sticky-top and broadcast-mention handling are untouched.
- **Empty-query guard.** `normalizedQuery.length > 0 && match(...)` avoids invoking `match` on an empty query, preserving the "show all members on empty query" behavior (the existing test still passes).
- **Tests added and meaningful.** Parameterized cases cover full pinyin, uppercase, partial, and initials for a Chinese name — good coverage of the abbreviation forms.

## Dependency
New dep `pinyin-pro@^3.28.0` added to `package.json` + `pnpm-lock.yaml`. dependency-review and osv-scan CI both pass on it.

## CI
Build / secret-scan / osv-scan / dependency-review all green; non-draft.

## Note (non-blocking)
`match()` runs per-member on each keystroke. For a mention dropdown (already a bounded, pre-filtered UI list) this is negligible; only worth revisiting if it's ever reused over a very large unfiltered member set.

Clean, well-tested, no regression.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Review: octo-web PR#784 — feat(mention): support pinyin member search
**Head SHA:** `d89806561f1f6179c8f8524e4de12351e341a6d9`
**Base:** `pr-base` (merge-base with main)
**Scope:** 32 files, +2414/-637 lines

---

## 1. Verification Summary

| Area | Status | Evidence |
|---|---|---|
| Pinyin mention search | ✅ | `mentionRender.ts:262-266`, test coverage `mentionRender.test.ts:56-66` |
| Filter unification (board/list/grouped) | ✅ | `IssuePage.tsx:138-154` common object, `issueFilterQuery.ts` pure function + tests |
| Label filter | ✅ | `labelApi.ts` exists, `types.ts:73` `label_ids`, `IssuePage.tsx:112,277,346` |
| Archived agent name resolution | ✅ | `directory.ts:33-34` include_archived + picker gated `!a.archived_at` |
| GROUP.md flag cache writeback | ✅ | `mdFlagCache.ts` pure function + 7 unit tests; `GroupMdEditor/index.tsx:201,236` |
| Sheet hyperlink sync (Yjs) | ✅ | `binding.ts:1096-1210` scan/sync/apply with sanitize-at-apply |
| Math formula float-DOM | ✅ | `MathFormula.tsx`, `formulaBridge.ts`, `FormulaPicker.tsx` |
| Font alias CSS | ✅ | `styles.css:3050-3125` local() @font-face, export unaffected |
| XLSX import re-enabled | ✅ | `DocsHome.tsx:42` IMPORT_ENABLED=true, hyperlink + WPS cell image parsing |

## 2. Findings

### P2-1 — PR scope far exceeds title (maintainability)

**Diff-scope:** N/A (meta)

PR title says "feat(mention): support pinyin member search" but ships **at least 6 independent features**:
1. Pinyin member search (mentionRender)
2. Filter unification across board/list/grouped views (IssuePage + issueApi + types)
3. Label filter (labelApi + IssuePage + i18n)
4. Archived agent name resolution (directory.ts)
5. GROUP.md flag cache writeback (mdFlagCache + GroupMdEditor)
6. Sheet: hyperlinks, math formulas, font aliases, xlsx import re-enable

**Impact:** Harder to bisect regressions, harder to revert individual features, harder for reviewers to give per-feature feedback. Not a correctness bug.

**Suggestion:** Split into 2-3 focused PRs in future (e.g., one for mention/filter/label, one for sheet features, one for mdFlagCache).

### P2-2 — `include_no_project` now sent to non-grouped endpoints

**Diff-scope:** new (previously gated to grouped view only)

- **Old:** `noProject` checkbox rendered only when `view === "grouped"` (line was inside `{view === "grouped" && ...}`). `include_no_project` only sent to `/issues/grouped`.
- **New:** `noProject` checkbox rendered for all views (`IssuePage.tsx:345-347`, no view gate). `common.include_no_project` flows into `listIssues()` for board/list views (`issueApi.ts:46`).
- **Risk:** If the backend `/issues` endpoint does not support `include_no_project`, the checkbox appears effective but is silently ignored — the user checks "Incl. no project" in list/board view and still sees only issues with a project. The filter badge counts it as active (`IssuePage.tsx:292`) compounding the misleading signal.

**Suggestion:** Verify with backend whether `/issues` (non-grouped) supports `include_no_project`. If not, either (a) disable the checkbox when `view !== "grouped"`, or (b) add backend support.

### P2-3 — `scanAllHyperLinks` does not sanitize payload on local→Yjs write

**Diff-scope:** new

`binding.ts:1099-1127` — `scanAllHyperLinks` reads `l.payload` from the model as-is and writes it to Yjs without scheme-checking. The sanitization only happens at the remote-apply boundary (`applyRemoteHyperLinks`, line 1174: `sanitizeLinkHref(stored.payload)`).

- **Local client:** If somehow an unsanitized URL enters the model locally (bypassing the editor/import guard), it's stored in Yjs and the local client's model keeps it.
- **Remote clients:** Protected — they sanitize on apply.
- **Severity:** Low. Import sanitizes at the boundary (`xlsxImport.ts:430`), and the editor UI would need to bypass `sanitizeLinkHref` to insert a bad URL. Belt-and-suspenders improvement.

**Suggestion:** Add `sanitizeLinkHref(l.payload)` in `scanAllHyperLinks` before writing to the Yjs map, for defense-in-depth.

### P2-4 — Missing trailing newline in i18n JSON files

**Diff-scope:** new

Both `packages/docs/src/i18n/en-US.json` and `packages/docs/src/i18n/zh-CN.json` end without a trailing newline (`\ No newline at end of file` in diff). POSIX convention + most linters expect a trailing newline.

## 3. Suggestions

- **P2-2 verification:** Quick backend check on whether `include_no_project` is a no-op or functional on `/issues` would resolve the concern.
- **FormulaPicker `onClose` dep:** `FormulaPicker.tsx:56-60` has `useEffect` depending on `[onClose]`. If the parent passes an inline arrow function, the listener is re-attached every render. Consider `useCallback` on the parent or `useRef` inside. Non-blocking.
- **`dataUrlToFile` regex:** `CollabSheet.ts:76` uses `/^data:([^;,]+)(?:;charset=[^;,]+)?;base64,(.+)$/s` — the `s` flag (dotAll) is correct here since base64 can span the entire rest of the string. Fine.

## 4. Extra Findings

- **`listMyGroupedIssues` strip logic** (`issueApi.ts:130-140`): Now correctly strips `assignee_ids`, `creator_ids` (plural) from base params in addition to the singular forms. This prevents filter selections from the "My Loop" panel polluting the three-leg union query. Well-commented.
- **`groupIssuesByAssignee`** (`issueGrouping.ts`): Clean pure function, preserves first-seen order (search relevance), folds unassigned into a null-key group. Good test coverage.
- **Archived agent handling** (`directory.ts:33-34,60-61`): Name map includes archived agents (for display of historical references), but the assignee picker correctly filters them out (`!a.archived_at`). This is the right behavior.
- **MathFormula keyboard kill** (`MathFormula.tsx:55-62`): Global CSS `display:none !important` on MathLive's virtual keyboard classes. ID-guarded to prevent duplicates. Appropriate for a desktop-first sheet editor.
- **`withMdFlags` immutability** (`mdFlagCache.ts:22-43`): Returns new objects (spread), does not mutate input. Both root-level and nested thread object handled. Tests verify immutability. Clean.

## 5. Data Flow Tracing

| Consumer | Upstream Source | Traced? | Notes |
|---|---|---|---|
| `filteredMembers` (mentionRender.ts:262) | `items` from member list + `match(name, query)` from pinyin-pro | ✅ | Empty query → `nameMatches` only (via `toLowerCase().includes("")` = true → all members). Pinyin branch gated by `normalizedQuery.length > 0`. Correct. |
| `common.include_no_assignee` (IssuePage.tsx:145) | `noAssigneeActive = scope === "all" && f.noAssignee` (line 95) | ✅ | Disabled in UI when `scope !== "all"`. Backend only receives `true` when both conditions met. |
| `common.label_ids` (IssuePage.tsx:151) | `f.labelIds` from filterSelect → `labels` state → `listLabels()` API | ✅ | Empty array folded to `undefined` by `arrayFilterQuery.csv()`. |
| `hyperLinkMap` (binding.ts) | Local: `scanAllHyperLinks` reads model; Remote: `applyRemoteHyperLinks` from Yjs | ✅ | Bidirectional sync with `applyingRemote` guard prevents echo. `linkUpdate$` subscription drives local→Yjs. |
| `dispImgMap` (xlsxImport.ts:229) | WPS zip `xl/cellimages.xml` + rels + media files | ✅ | Regex-based XML parsing. Graceful degradation: empty map on any miss. Buffer cloned before ExcelJS consumes it (line 393). |
| `channelInfo.orgData` (GroupMdEditor:199-236) | Backend `resp.version` from save/delete API → `withMdFlags()` pure fn | ✅ | `version > 0` = configured. Both save and delete paths call `applyMdFlagToCache`. |

## 6. Blind Spot Checklist (C1–C6)

| Check | Result | Evidence |
|---|---|---|
| **C1 — Dual-path parity** | Clear | Filter changes are symmetric across board/list/grouped. `scopeToAssigneeTypes` applied uniformly to both `listGroupedIssues` and `listIssues`. |
| **C2 — Control-flow ordering / nesting** | Clear | `arrayFilterQuery` is a pure function called at two sites (listIssues, listGroupedIssues). No ordering dependency. `withMdFlags` is a pure function called at two sites (save, delete). |
| **C3 — Authorization boundary** | N/A | No auth changes in this PR. |
| **C4 — Authorization lifecycle** | N/A | No auth changes. |
| **C5 — Build/runtime parity** | Clear | Font aliases are CSS-only (@font-face local()), don't affect stored data. XLSX import/export both sanitize hyperlinks. Formula float-DOM registered with Univer ComponentManager (key `octo-math-formula`), shared across collaborating clients. |
| **C6 — Governance/docs self-consistency** | N/A | No governance changes. |

## 7. Cross-round Blocker Re-check

---

**[Octo-Q] verdict: APPROVE**

All findings are P2 (maintainability/suggestion level). No P0/P1 defects identified. The pinyin search implementation is correct with good test coverage. The filter unification is well-structured with proper error handling and pure-function extraction. The sheet features (hyperlinks, formulas, font aliases) follow established patterns with proper sanitization at trust boundaries. The PR ships significantly more than its title suggests — consider splitting in future — but the code quality is solid across all features.

yujiawei (approved):
# Code Review — PR #784 (octo-web)

**Head SHA:** `d89806561f1f6179c8f8524e4de12351e341a6d9` · **Base:** `main` (merge-base `8f73bb1c`)

Adds pinyin-based matching to the `@mention` member dropdown (full pinyin, partial pinyin, and initials) via `pinyin-pro`'s `match()`, on top of the existing case-insensitive substring filter. Scope: 4 files, +29/-4.

## 1. Spec compliance

**Spec: ✅**

Verified against the linked requirement (support `@zhangsan` → 张三):
- **Nothing missing:** full pinyin (`zhangsan`), initials (`zs`), and partial (`san`) all match; case-insensitive on the latin side (`ZHANGSAN` works). Existing latin substring filtering is preserved (`nameMatches || pinyinMatches`).
- **Nothing extra:** the change is limited to the filter predicate, the `pinyin-pro` dependency declaration, the lockfile, and unit tests. No unrelated flags, fields, or behavior added.
- **No divergence:** the empty-query sticky broadcast items (`@所有人` / `@所有AI`) still only appear when the query is empty, and pinyin matching is guarded by `normalizedQuery.length > 0`, so it never affects the empty-query path.

## 2. Code quality

**Quality: Approved**

I reproduced `match()`'s behavior locally against `张三` and non-matching names:
- Correct positives: `zhangsan`/`ZHANGSAN`/`san`/`zs`/`zsan` → match; `xyz` → no match.
- **No false positives:** `李四`, `王五`, `Alice` return `null` for `zs`/`san`. Polyphones resolve (`长风` matches both `chang` and `zhang`). Latin names are already covered by the existing `.includes` path, so the added `match()` OR-branch introduces no surprising over-match.
- Empty/whitespace query is correctly short-circuited by the `length > 0` guard before `match()` runs.

Tests pass and cover the new behavior:
- `@octo/base` `mentionRender.test.ts` — 9 passed (new `it.each(["zhangsan","ZHANGSAN","san","zs"])` cases included).
- `@octo/web` `mentionRefactor.test.ts` — 72 passed.
- `pnpm --filter @octo/web build` — succeeds (only pre-existing unrelated warnings).

### Advisory (P2 — non-blocking)

- **P2 — bundle size** (`packages/dmworkbase/package.json`, `mentionRender.ts:1`): `pinyin-pro` is a static, top-level import in `mentionRender.ts`, which is reachable from the core `MessageInput` component, so it lands in the main chunk rather than a lazy split. The library ships an embedded Chinese dictionary (~160KB gzipped for the ESM entry); `match()` needs that dictionary, and there is no lighter subpath export for it. This is acceptable for the feature but worth a follow-up: consider dynamically importing the pinyin filter path so the dictionary only loads when a mention dropdown is actually opened.
- **P2 — per-keystroke cost** (`mentionRender.ts:263-268`): `match()` runs once per member inside the `.filter` on every keystroke. Measured ~0.12ms/member (500 members × 10 keystrokes ≈ 60ms total), so it is fine for typical rooms but could cause typing lag in workspaces with thousands of members. Note this amplifies an already-existing pattern (the previous `.includes` filter also iterated all members per keystroke). A future optimization would pre-compute each member's pinyin once when the member list loads, rather than on every keystroke.

Neither rises to a merge blocker.

## 3. Verdict

**APPROVED** — spec fully implemented, correctness verified empirically, tests and build green. The two bundle/perf items are advisory follow-ups, not blockers.

## Reviewer coverage / limitations

- A cross-family advisory pass (Gemini 3.1 Pro) independently reached the same two P2s (bundle size, per-keystroke cost) and found no blocker. A second advisory tool was attempted but was unavailable due to a gateway tool-config error, so it is recorded as absent, not as agreement.
- Not measured: exact production `items` array size (upper bound on members) and the precise post-tree-shaking bundle delta in the real web build output.

OctoBoooot (approved):
## Upgrading my #784 verdict COMMENT → APPROVE — Build now green

My earlier COMMENT (id 4702180978, same head `d89806561f1f`) held APPROVE back for one reason only: the required **Build** check was still `in_progress`, and I don't approve over unconfirmed code CI. Build has now concluded **success** (along with secret-scan / osv-scan / dependency-review / pr-title-lint), so the sole hold is cleared. The code was already verified clean; upgrading to APPROVE.

### Nothing else changed — the earlier verification stands
- `pinyin-pro` `match(text, pinyin) => number[] | null` used correctly (`!== null` test); default `precision: "start"` is the right choice.
- Additive OR branch — the existing case-insensitive substring filter is preserved, so Latin-name filtering is unchanged; pinyin only *adds* matches. No regression.
- No false-positive pollution — `match("Alice","zs")` is null; the `it.each(["zhangsan","ZHANGSAN","san","zs"])` test proves Chinese names match while Alice/BuildBot stay excluded.
- Empty-query guard (`normalizedQuery.length > 0 && …`) preserves "empty query lists all members."
- `pinyin-pro@^3.28.0` declared in the consuming package (`@octo/base`), resolves to the same 3.28.0 already in the lockfile (no conflict/dup); dependency-review + osv-scan green on it.

### Verdict
APPROVE @`d89806561f1f`. Code CI fully green, no blockers/majors, well-tested. The one non-blocking note stands (per-member pinyin pass per keystroke — negligible at mention-dropdown roster sizes). This supersedes my COMMENT, which was a CI-timing hold, not a code concern.

PR #785, opened 2026-07-15
Feat octo doc html

Ranwanglc:
## Changes

-
-

- Affected module(s):
- Shared layer touched: 

## Testing

- [ ] Manually verified

## Checklist

- [ ] Updated documentation


packages/dmworkbase/src/Components/DocTab/DocTab.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <iframe
+        ref={iframeRef}
+        src={src}
+        title={title}
+        data-testid="doc-tab-iframe"
+        className="wk-doc-tab__iframe"
+        onLoad={() => setLoading(false)}
+        // same-origin 是必需的：doc 页面 JS 通过同源 fetch 拿 capability cookie。
+        // 不给 allow-top-navigation，避免恶意 doc 劫持宿主 tab。
+        sandbox="allow-scripts allow-same-origin allow-forms allow-popups"
```
OctoBoooot, line 73: **[major]** `sandbox="allow-scripts allow-same-origin …"` is only safe while octo-doc is a *separate* origin from the host. The two tokens together let framed content reach its own origin's DOM/storage — fine for a distinct deployment, but if a deployment ever points the doc base at the host origin (the same `VITE_OCTO_DOC_BASE`/`__OCTO_DOC_BASE__`-unconfigured → same-origin-default misconfig that HtmlDocView warns about), this becomes a full sandbox escape into the host app. DocTab is an unmounted Stage-A skeleton today (not imported anywhere), so not exploitable yet — but before Stage B wires it, enforce the invariant: validate `src`'s origin ≠ host origin (fail closed otherwise), and lock it with a test alongside the existing allow-top-navigation denial.

packages/docs/src/html/HtmlDocView.tsx
```diff
@@ -0,0 +1,452 @@
...
+/**
+ * Legacy sanitizer retained for callers that still need a stripped inline fragment.
+ *
+ * Relies on DOMPurify's default safe baseline (drops <script>, on* handlers and
+ * javascript:/data: script URLs) and additionally strips interactive/editable elements and
+ * the contenteditable attribute so the rendered doc is strictly presentational. Ordinary
+ * display markup is preserved by the default allow-list; inline style is forbidden (see
+ * FORBID_ATTR) to close the CSS-value injection surface DOMPurify does not deep-clean.
+ */
+export function sanitizeDocHtml(raw: string): string {
```
OctoBoooot, line 48: **[minor]** `sanitizeDocHtml` is exported "for legacy callers" but is called nowhere in this PR (verified across the html/ module). Dead code that misleads: a reader sees a DOMPurify sanitizer and may assume the rendered doc is sanitized, when the actual isolation is the script-disabled iframe sandbox. Either wire it where a stripped fragment is genuinely needed, or drop it so the sandbox-is-the-boundary model stands unambiguous.

OctoBoooot (changes requested):
## Review: octo-doc HTML read-only viewer (#785)

**Verdict:** Request changes — code CI is red (`pr-title-lint`) which bars merge, and the PR ships +2805 lines with an entirely empty description template. The security core (rendering untrusted agent HTML) is actually **sound** — I verified the sandbox isolation end-to-end — so this is close; the blockers are the red check + the missing description, plus one forward-looking major on DocTab.

CI: `pr-title-lint` is **failing** — the title "Feat octo doc html" isn't Conventional Commits. That's a code check and blocks APPROVE. Fix: `feat(docs): octo-doc HTML read-only viewer` (or similar).

### Blockers
- **CI red — `pr-title-lint`.** Retitle to Conventional Commits format. Trivial, but it's a required check.
- **Empty PR template (`packages/…` — whole PR).** For a +2805 / 23-file change that renders untrusted agent HTML, the description, testing notes, and Architecture/Module-Boundary section are all unfilled placeholders. This one genuinely needs the security model written down (the sandbox-not-sanitize decision below is important and should be in the PR, not just code comments). Please fill Summary / Changes / Testing / the shared-code impact scope.

### Major
- `packages/dmworkbase/src/Components/DocTab/DocTab.tsx:73` — **`sandbox="allow-scripts allow-same-origin …"` is only safe while octo-doc is a *separate* origin from the host.** The two tokens together let framed content reach its own origin's DOM/storage; that's fine when the framed `src` is a distinct deployment, but if a deployment ever points the doc base at the *host* origin (exactly the "same-origin default when `VITE_OCTO_DOC_BASE`/`__OCTO_DOC_BASE__` is unconfigured" misconfiguration that `HtmlDocView` itself warns about in its console diagnostic), this iframe becomes a full sandbox escape into the host app. DocTab is an unmounted Stage-A skeleton today (not imported anywhere), so it's not exploitable yet — but before it's wired in Stage B, enforce the invariant: assert/validate that `src`'s origin ≠ host origin (or document that the doc base MUST be cross-origin and fail closed otherwise). Worth a test alongside the existing `allow-top-navigation` denial assertion.

### Security — verified sound (the part that matters most here)
The live path renders **unsanitized, agent-authored HTML** (the file header is admirably explicit: the backend does not sanitize the Publish payload, so it may contain `<script>`, `on*`, `javascript:` URLs). I traced whether that's safe:
- `HtmlDocView` renders it via `<iframe sandbox="allow-same-origin" srcDoc={html}>` — **`allow-scripts` is never granted**, so the browser blocks all script execution (inline `<script>`, event handlers, `javascript:` URLs) regardless of same-origin. The test locks this precisely (`<script>window.__pwned=1</script>` stays in srcdoc but does not run; `sandbox` asserted to be exactly `allow-same-origin` and to NOT contain `allow-scripts`).
- Not sanitizing is a *deliberate, valid* choice here — it preserves agent CSS (`<style>`, inline style, stylesheet links) intact, and the sandbox is the isolation boundary. `absolutizeDocAssetUrls` additionally disables editable controls and only rewrites `/assets/` URLs (rejecting absolute/`javascript:`/protocol-relative via `isAbsoluteOrSpecialUrl`).
- The comment panel renders anchor/comment text through JSX text nodes (`{thread.text}`, `{r.text}`) — React-escaped, no `innerHTML`. No secondary sink.
So there is **no XSS blocker** on the live path. Good, careful design.

### Minor
- `packages/docs/src/html/HtmlDocView.tsx:48` — `sanitizeDocHtml` is exported "for legacy callers" but is **called nowhere** in this PR (verified across the `html/` module). Dead code that actively misleads: a reader sees a DOMPurify sanitizer and may assume the rendered doc is sanitized, when the actual isolation is the sandbox. Either wire it where a stripped fragment is genuinely needed, or drop it and let the sandbox comment stand alone.

### Praise
- The security posture is documented at exactly the right place — the file-header SECURITY block states the threat model (backend doesn't sanitize → sandbox isolates) instead of leaving it implicit, and the tests assert the *negative* security invariants (no `allow-scripts`, no `allow-top-navigation`, script-in-srcdoc-doesn't-run). That's the right way to lock a sandbox contract against a future well-meaning "just add allow-scripts" edit.
- Withholding `allow-top-navigation` on DocTab with a reverse-assertion test is a sharp touch — tab-hijack via `top.location` is the exact thing people forget.

### Net
No XSS blocker — the untrusted-HTML render path is correctly isolated by a script-disabled sandbox, and I verified it against the hostile-`<script>` case. What blocks merge is mechanical + process: fix the red `pr-title-lint`, fill the empty template (especially write down the sandbox-not-sanitize security model), address the DocTab cross-origin invariant before Stage B wires it, and clean up the dead `sanitizeDocHtml`. Retitle + description + the DocTab guard and this is in good shape.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #785 (octo-web)

## Summary

This PR introduces the read-only HTML document viewer (`HtmlDocView`) for agent-authored `docType==='html'` documents, plus the 2b comment rail (`HtmlDocCommentPanel`), member management panel (`HtmlMemberPanel`), grants API (`htmlGrantsApi`), and a `DocTab` iframe skeleton for future chat-panel embedding. The routing in `DocsHome` is extended to handle `'html'` doc types alongside existing board/sheet/doc shells, with `octoDocSlug` threaded through sessionStorage, deep-link, and `getDoc` resolution paths.

The PR is well-structured with clear security architecture (iframe sandbox as the script-isolation boundary), good i18n coverage, and comprehensive tests. Two non-blocking P2 items identified.

## Verification

- ✅ **iframe sandbox security** — `HtmlDocView` uses `sandbox="allow-same-origin"` (no `allow-scripts`), correctly blocking script execution in agent-authored HTML. Selection-change hooks are installed from the parent context. `DocTab` uses a more permissive sandbox intentionally (Stage A skeleton, not yet production-consumed).
- ✅ **Data flow: comments** — `listComments` reads from `{data}` envelope matching the backend's `writeList` contract (commit `949688f` confirmed the fix from `{roots}`). `createComment` sends the correct `{slug, text, version, anchor?/parent_id?}` exclusive shape.
- ✅ **Data flow: grants** — `listGrants` handles both `{data: {items}}` and `{items}` envelopes. `addGrant`/`removeGrant` use correct PUT/DELETE verbs against `/grants/{slug}`.
- ✅ **Routing parity** — `docType==='html'` handled consistently across list-open (`:639`), deep-link/sessionStorage (`readDocTarget`), and `getDoc` resolution (`:1120`). `octoDocSlug` threaded through `commitOpen` → `persistDocTarget` → `buildRightPane` → `HtmlDocView`. `backToList` clears `selectedOctoDocSlug`.
- ✅ **i18n completeness** — all new `t()` keys (`docs.comment.*`, `docs.header.*`, `docs.member.errorLoad`) present in both `en-US.json` and `zh-CN.json`. `agentInstructionTitle` i18n template matches `buildAgentInstruction`'s `values: { request, slug, target }`.
- ✅ **Auth headers** — all three octo-doc fetch paths (render, comments, grants) carry the `token` header from `getWKApp().loginInfo?.token` plus `credentials: 'include'`. Raw `fetch` is correct here — octo-doc is a separate deployment, must not route through the octoweb apiClient.
- ✅ **DOMPurify profile** — `sanitizeDocHtml` forbids interactive elements (`input`, `button`, `textarea`, `select`, `option`, `form`, `label`, `fieldset`), `contenteditable`, `autofocus`, `onfocus`, and `style` (CSS-value injection surface). Tests cover script stripping, interactive element removal, and CSS injection blocking.

Static analysis only at head `949688f94ea87e77a05193dca375d8082dd6df8f`; build and tests not executed in this environment.

## Findings

### P2 — Stale `ListCommentsResponse` type (`packages/docs/src/html/htmlDocComments.ts:67`)

`ListCommentsResponse` is exported with `{ roots: OctoDocCommentThread[] }` but the actual backend returns a `{ data: OctoDocCommentThread[] }` envelope. The type is never consumed internally — `listComments` reads `data?.data ?? []` via inline cast at `:94`. Any external consumer importing `ListCommentsResponse` would see the wrong shape. Remove the stale `roots` field or align the type to `{ data: OctoDocCommentThread[] }`.

### P2 — Sequential `addGrant` with partial-failure loss (`packages/docs/src/html/HtmlMemberPanel.tsx:68`)

`onAdd` iterates uids sequentially (`for (const uid of uids) await addGrant(...)`) and the `try/catch` wraps the entire loop. If the 3rd of 5 grants fails, grants 1–2 are persisted but 4–5 never fire. The UI shows a generic error without indicating which members were added. Consider `Promise.allSettled` to attempt all grants, then refresh and surface per-uid results.

- `parseOdocMeta` regex (`HtmlDocView.tsx:173`) — regex on raw HTML, `JSON.parse` try/catch, parse miss falls back to no metadata (header renders slug). No code execution path.
- `absolutizeDocAssetUrls` (`HtmlDocView.tsx:121`) — only rewrites `img[src]`/`link[href]` matching `/assets/`, rejects absolute/special URLs, uses DOMParser (safe).
- `buildAnchorFromSelection` (`htmlDocAnchor.ts:54`) — element-aid lookup walks up ancestors correctly, text anchor with bounded context, null for collapsed/whitespace selections.
- `buildAgentInstruction` (`htmlDocAnchor.ts:122`) — i18n fallback chain works when key is missing. URL params correctly encode all anchor variants.
- `DocTab.tsx` sandbox — `allow-scripts allow-same-origin allow-forms allow-popups` (no `allow-top-navigation`) is intentional for Stage A. Vitest has a reverse assertion locking out `allow-top-navigation`.
- `resolveDocAssetUrl` — rejects absolute URLs, `//`-relative URLs, `#` fragments, and non-`/assets/` paths.
- `neutralizeEditableControls` — sets `contenteditable="false"` and `disabled` on form controls in the parsed HTML before srcdoc injection.
- `cssAttrValue` escaping in `resolveHtmlDocAnchorText` — escapes backslashes and double-quotes for safe `querySelector` selector construction.

## Data flow trace

- **`listComments`** → backend `{data: [...]}` → `data?.data ?? []` → `HtmlDocCommentPanel.threads` state → rendered as `<li>` per thread with React auto-escaping text content. ✓
- **`listGrants`** → backend `{data: {items}}` or `{items}` → dual-cast fallback → `HtmlMemberPanel.grants` → rendered rows with creator filtered from remove path. ✓
- **Selection anchor** → `selectionchange` event in iframe doc → `buildAnchorFromSelection` → `pendingAnchor` state → `HtmlDocCommentPanel` composer → `createComment` body (exclusive: anchor OR parent_id). ✓
- **"让 AI 处理"** → `buildAgentInstruction(instructionDoc, thread)` → `openDocForward({docId, title, link, canGrant: false})` → host IM bridge. Gated by `canForwardToChat()`. ✓
- **HTML render** → `fetch(url)` → `absolutizeDocAssetUrls` (DOMParser rewrite) → `parseOdocMeta` (regex extract) → `setState({html, meta})` → `srcDoc={state.html}` on sandboxed iframe. ✓

## Blind-spot checklist

- **C1 — Symmetric paths**: `docType === 'html'` routing handled in list-open, deep-link, and `getDoc` resolution. `backToList` clears `selectedOctoDocSlug`. `persistDocTarget` writes and `readDocTarget` reads `octoDocSlug`. Consistent. ✓
- **C2 — Control-flow ordering**: N/A for this PR (no cross-cutting control flow changes).
- **C3 — Authorization boundary**: Raw `fetch` calls to octo-doc are by design (separate deployment). All carry `credentials: 'include'` + `token` header. `HtmlMemberPanel` gates management to `canManage` (creatorUid === currentUid). ✓
- **C4 — Container-member state cascade**: N/A (no auth lifecycle cascading in this PR).
- **C5 — Build/runtime path**: `resolveOctoDocBase` defaults to same-origin (safe default), with diagnostic `console.warn` when base is unconfigured and request fails. ✓
- **C6 — Governance/docs**: N/A (not a governance PR).

[Octo-Q] verdict: APPROVE — 两个 P2 均为非阻塞型问题（stale export type + sequential grant partial-failure），不影响核心功能路径的正确性或安全性。iframe sandbox 安全架构清晰，数据流完整，i18n 齐全，路由 parity 一致。

Jerry-Xin (changes requested):
Scope check passed: this PR is relevant to octo-web, but it has blocking routing and sandbox/security issues.

**🔴 Blocking**
- 🔴 Critical: Standalone `/d/:docId` links do not handle HTML documents. `DocsHome` routes `docType === "html"` to `HtmlDocView`, but `StandaloneDocPage` still only special-cases `board` and `sheet`; every other type falls through to `EditorShell`. An HTML doc opened through the existing standalone/share entry point will try to mount the rich-text/Yjs editor instead of the read-only HTML view. Add an `html` branch here, pass `meta.octoDocSlug`, and cover it with a standalone-page regression test.  
  `packages/docs/src/pages/StandaloneDocPage.tsx:500`, `packages/docs/src/pages/StandaloneDocPage.tsx:543`, `packages/docs/src/pages/StandaloneDocPage.tsx:560`

- 🔴 Critical: `DocTab` uses the unsafe sandbox combination `allow-scripts allow-same-origin` while accepting caller-provided doc URLs, including same-origin paths. For same-origin content, this combination removes the sandbox as a meaningful boundary; omitting `allow-top-navigation` is not enough protection. Either enforce a separate trusted origin for `src`, remove one of `allow-scripts` / `allow-same-origin`, or stop treating this iframe as sandbox-isolated.  
  `packages/dmworkbase/src/Components/DocTab/DocTab.tsx:71`

**💬 Non-blocking**
- 🟡 Warning: `HtmlMemberPanel` says grants are reader-only and forces `addGrant(..., "reader")`, but it renders the shared `MemberPicker`, whose role dropdown still lets users choose writer/admin. That selection is silently ignored. Hide/disable the role selector for this surface or provide a reader-only picker mode.  
  `packages/docs/src/html/HtmlMemberPanel.tsx:58`, `packages/docs/src/html/HtmlMemberPanel.tsx:107`

- 🟡 Warning: `HtmlDocView` renders an enabled “Forward to chat” header button that only shows a coming-soon notice. This looks like a live action but cannot complete. Prefer hiding it, disabling it with a clear reason, or wiring the existing forward flow.  
  `packages/docs/src/html/HtmlDocView.tsx:359`, `packages/docs/src/html/HtmlDocView.tsx:387`

- 🔵 Suggestion: `DocTab.css` is never imported by `DocTab.tsx` or `index.ts`, unlike the surrounding dmworkbase component pattern. Consumers importing the component will get unstyled markup unless they remember a separate CSS import.  
  `packages/dmworkbase/src/Components/DocTab/DocTab.tsx:19`, `packages/dmworkbase/src/Components/DocTab/index.ts:1`

**✅ Highlights**
- Good focused coverage was added for DocsHome HTML routing, comment anchors, comment API calls, and iframe read-only behavior.
- The HTML viewer avoids the rich-text editor path in DocsHome and keeps comments outside the rendered document content.

Verification note: I attempted the focused docs tests and `pnpm i18n:check`, but this checkout has no installed dependencies (`vitest` and `typescript` are missing from `node_modules`), so I could not execute them locally.

yujiawei (changes requested):
# Code Review — PR #785 (octo-web)

Reviewed at head `949688f9`, merge-base `d80703a5`. Scope: a read-only HTML document viewer (`HtmlDocView`) that renders agent-authored HTML in a sandboxed iframe, a side comment rail (`HtmlDocCommentPanel`), an octo-doc grants/member panel, anchor helpers, `docType='html'` routing in `DocsHome`, and a `DocTab` skeleton in `dmworkbase`.

Overall this is a well-structured, well-tested change (158 docs unit tests pass locally, and the security model of the primary viewer is sound). There is one merge blocker: the PR fails the `i18n:check` CI gate. Details below.

## 1. Specification / scope check

The change is internally coherent and matches its stated intent (read-only HTML viewer + comments + grants + routing). No hidden feature creep in the viewer path. Two scope observations:

- **`DocTab` (`packages/dmworkbase/src/Components/DocTab/`) is dead code.** It is exported but never imported anywhere outside its own directory. The header comment states it is a "Stage A skeleton" that a later UI task will wire in, so this is intentional staging rather than an accident — but it is what triggers the CI failure below, and it ships an unused, more-permissive second iframe (`sandbox="allow-scripts allow-same-origin allow-forms allow-popups"`) into the tree ahead of its consumer.
- The change adds a raw `fetch` data layer (`htmlDocComments.ts`, `htmlGrantsApi.ts`, and the render fetch) that deliberately bypasses `WKApp.apiClient`. The repo convention (`CLAUDE.md`) is "API calls go through `WKApp.apiClient`, do NOT create separate fetch instances." The deviation is **justified and clearly documented** (octo-doc is a separate deployment reached cross-origin with a `token` header), so this is not a defect — just calling it out so it is a conscious, reviewed exception rather than drift.

## 2. Blocking issue (P1)

### P1 — CI `i18n:check` gate fails: hardcoded Chinese in `DocTab.tsx`
`.github/workflows/ci.yml` runs `pnpm i18n:check` in the required `build` job. Running it against this branch fails:

```
i18n check failed: 2 new hardcoded Chinese signatures found.
- packages/dmworkbase/src/Components/DocTab/DocTab.tsx: 加载中…
- packages/dmworkbase/src/Components/DocTab/DocTab.tsx: 暂无文档
```

Source — `packages/dmworkbase/src/Components/DocTab/DocTab.tsx:88-90`:

```tsx
const DocTab: React.FC<DocTabProps> = ({
  src,
  emptyText = "暂无文档",
  loadingText = "加载中…",
```

These default prop values are literal Chinese strings not routed through i18n `t()`, which the scanner flags as new hardcoded copy. This will hold the PR at a red required check. Fix by removing the hardcoded defaults (let the caller pass i18n'd text, matching the component's own doc comment "空态提示文本；调用方 i18n 后传入"), or by adding the strings to the i18n dictionaries / baseline.

## 3. Non-blocking findings (P2)

### P2 — `parseOdocMeta` header metadata is spoofable (`HtmlDocView.tsx:209-216`, used at `:266`)
`parseOdocMeta` runs an **unanchored** regex over the *raw* fetched HTML (including the agent-authored `<body>`):

```ts
const m = html.match(/__ODOC__\s*=\s*(\{[\s\S]*?\});/)
```

The parsed `title` / `creator_name` are then shown in the trusted host header (`headerTitle`, `headerCreator` at `:228-229`, rendered at `:352-357`). A document body containing `__ODOC__ = {"title":"…","creator_name":"…"};` would be picked up and displayed as authoritative header metadata. This is **not an XSS** (React escapes the text, and the viewer iframe cannot run scripts — see §4), but it is a content/authorship spoofing surface usable for phishing within the host chrome. Since `absolutizeDocAssetUrls` already `DOMParser`s the document, prefer extracting the blob from the expected injected `<script>` element, or at minimum anchor the pattern to a `<script>…</script>` wrapper.

### P2 — `ListCommentsResponse` type contradicts the actual wire contract (`htmlDocComments.ts:66-69`)
The interface declares the list envelope as `{ roots: ... }`, but `listComments` correctly reads `data?.data ?? []` (the `{data, pagination}` writeList envelope) — head commit `949688f9` is literally "read comment list from {data} envelope, not {roots}". The interface is now both dead (unused) and actively misleading about the contract. Delete it or align it to `{ data: OctoDocCommentThread[] }`.

### P2 — Type error in a new test file (`htmlDocComments.test.ts:82`)
`tsc --noEmit -p tsconfig.typecheck.json` reports:

```
src/html/htmlDocComments.test.ts(82,7): error TS2322:
  Property 'selector' is missing in type '{ kind: "element"; aid: string; }'
  but required in type 'ElementAnchor'.
```

The `anchor: { kind: 'element', aid: 'a1' }` literal omits the required `selector`. It does not break CI today (the `docs` package has no `build` script, so nothing typechecks it in the pipeline), but it is a real type error in code this PR adds and will bite the first time docs is added to a typecheck gate. (Two other `tsc` errors — `members/sort.ts`, `i18n/format.ts` — are **pre-existing** on the merge-base and not caused by this PR.)

### P2 — Comment reload has no request-sequence guard (`HtmlDocCommentPanel.tsx:59-66`)
`reload()` awaits `listComments(slug, version)` and unconditionally `setThreads(...)`, without the `reqSeq` guard that `HtmlDocView`'s fetch uses. In the current wiring the impact is small — `HtmlDocView` is keyed by `docId` (so it remounts on doc switch) and `version` is always the `latest` default from `DocsHome` — so `slug`/`version` do not change under a mounted panel today. It becomes a real last-writer-wins race the moment a version switcher or in-place doc swap is added. Worth a sequence/abort guard for parity with the render fetch.

## 4. Security assessment of the primary viewer (informational — no action needed)

The core render path is sound. Raw agent HTML is placed in `<iframe srcDoc=... sandbox="allow-same-origin">` **without `allow-scripts`** (`HtmlDocView.tsx:422-428`), and the tests lock this (`HtmlDocView.test.tsx:215,287`). Without `allow-scripts`, no `<script>`, `on*` handler, or `javascript:` URL executes regardless of `allow-same-origin`; the iframe also isolates the doc's CSS/DOM from the host. `allow-top-navigation` and `allow-popups` are both absent, so meta-refresh / `target=_top` / `window.open` navigation of the host is blocked too. The retained `sanitizeDocHtml` (which *does* strip scripts/handlers via DOMPurify) has **no live callers** — it is legacy-only — so gemini's suggestion to add `<style>` to its `FORBID_TAGS` is a hardening nicety for a currently-unused function, not a live exposure. This is a genuinely defensible design; the security comments in the file are accurate.

## 5. Verdict

**CHANGES_REQUESTED** — solely on the P1 `i18n:check` CI-gate failure, which will keep a required check red. The P2 items are quality/robustness improvements and are not, on their own, merge blockers. Once the hardcoded `DocTab` strings are removed (or baselined) so `pnpm i18n:check` is green, this is close to approvable; addressing the `parseOdocMeta` spoofing and the misleading `ListCommentsResponse`/test-type issues in the same pass is recommended.

---
Review methodology note: local `vitest` (158/158 pass) + `tsc` + `i18n:check` were run against head `949688f9`; a long-context cross-family advisory pass was incorporated. An additional line-level advisory reviewer was attempted but timed out and did not return findings (recorded as absent, not as tacit approval); its intended coverage — line-by-line correctness of the iframe/selection lifecycle — was covered manually here.

PR #787, opened 2026-07-15
Feat octo doc html

Ranwanglc:
## Changes

-
-

- Affected module(s):
- Shared layer touched: 

## Testing

- [ ] Manually verified

## Checklist

- [ ] Updated documentation


packages/dmworkbase/src/Components/DocTab/DocTab.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <iframe
+        ref={iframeRef}
+        src={src}
+        title={title}
+        data-testid="doc-tab-iframe"
+        className="wk-doc-tab__iframe"
+        onLoad={() => setLoading(false)}
+        // same-origin 是必需的：doc 页面 JS 通过同源 fetch 拿 capability cookie。
+        // 不给 allow-top-navigation，避免恶意 doc 劫持宿主 tab。
+        sandbox="allow-scripts allow-same-origin allow-forms allow-popups"
```
OctoBoooot, line 73: **[major]** `sandbox="allow-scripts allow-same-origin …"` is only safe while octo-doc is a *separate* origin from the host. The two tokens together let framed content reach its own origin's DOM/storage — fine for a distinct deployment, but if a deployment ever points the doc base at the host origin (the same `VITE_OCTO_DOC_BASE`/`__OCTO_DOC_BASE__`-unconfigured → same-origin-default misconfig that HtmlDocView warns about), this becomes a full sandbox escape into the host app. DocTab is an unmounted Stage-A skeleton today (not imported anywhere), so not exploitable yet — but before Stage B wires it, enforce the invariant: validate `src`'s origin ≠ host origin (fail closed otherwise), and lock it with a test alongside the existing allow-top-navigation denial.

packages/docs/src/html/HtmlDocView.tsx
```diff
@@ -0,0 +1,452 @@
...
+/**
+ * Legacy sanitizer retained for callers that still need a stripped inline fragment.
+ *
+ * Relies on DOMPurify's default safe baseline (drops <script>, on* handlers and
+ * javascript:/data: script URLs) and additionally strips interactive/editable elements and
+ * the contenteditable attribute so the rendered doc is strictly presentational. Ordinary
+ * display markup is preserved by the default allow-list; inline style is forbidden (see
+ * FORBID_ATTR) to close the CSS-value injection surface DOMPurify does not deep-clean.
+ */
+export function sanitizeDocHtml(raw: string): string {
```
OctoBoooot, line 48: **[minor]** `sanitizeDocHtml` is exported "for legacy callers" but is called nowhere in this PR (verified across the html/ module). Dead code that misleads: a reader sees a DOMPurify sanitizer and may assume the rendered doc is sanitized, when the actual isolation is the script-disabled iframe sandbox. Either wire it where a stripped fragment is genuinely needed, or drop it so the sandbox-is-the-boundary model stands unambiguous.

OctoBoooot (changes requested):
## Review: octo-doc HTML read-only viewer (#787)

_(This PR carries the identical head `949688f9` as the now-closed #785; my review there applies verbatim and is reproduced here on the live PR. One addition: this PR is also `mergeStateStatus=DIRTY` — a merge conflict with base.)_

**Verdict:** Request changes — code CI is red (`pr-title-lint`) which bars merge, the branch has a merge conflict with base (DIRTY), and the PR ships +2805 lines with an entirely empty description template. The security core (rendering untrusted agent HTML) is actually **sound** — I verified the sandbox isolation end-to-end — so this is close; the blockers are the red check + the conflict + the missing description, plus one forward-looking major on DocTab.

CI: `pr-title-lint` is **failing** — the title "Feat octo doc html" isn't Conventional Commits. That's a code check and blocks APPROVE. Fix: `feat(docs): octo-doc HTML read-only viewer` (or similar).

### Blockers
- **CI red — `pr-title-lint`.** Retitle to Conventional Commits format. Trivial, but it's a required check.
- **Merge conflict with base (`mergeStateStatus=DIRTY`).** The branch no longer merges cleanly into `main` — rebase/merge base and resolve before this can land.
- **Empty PR template (`packages/…` — whole PR).** For a +2805 / 23-file change that renders untrusted agent HTML, the description, testing notes, and Architecture/Module-Boundary section are all unfilled placeholders. This one genuinely needs the security model written down (the sandbox-not-sanitize decision below is important and should be in the PR, not just code comments). Please fill Summary / Changes / Testing / the shared-code impact scope.

### Major
- `packages/dmworkbase/src/Components/DocTab/DocTab.tsx:73` — **`sandbox="allow-scripts allow-same-origin …"` is only safe while octo-doc is a *separate* origin from the host.** The two tokens together let framed content reach its own origin's DOM/storage; that's fine when the framed `src` is a distinct deployment, but if a deployment ever points the doc base at the *host* origin (exactly the "same-origin default when `VITE_OCTO_DOC_BASE`/`__OCTO_DOC_BASE__` is unconfigured" misconfiguration that `HtmlDocView` itself warns about in its console diagnostic), this iframe becomes a full sandbox escape into the host app. DocTab is an unmounted Stage-A skeleton today (not imported anywhere), so it's not exploitable yet — but before it's wired in Stage B, enforce the invariant: assert/validate that `src`'s origin ≠ host origin (or document that the doc base MUST be cross-origin and fail closed otherwise). Worth a test alongside the existing `allow-top-navigation` denial assertion.

### Security — verified sound (the part that matters most here)
The live path renders **unsanitized, agent-authored HTML** (the file header is admirably explicit: the backend does not sanitize the Publish payload, so it may contain `<script>`, `on*`, `javascript:` URLs). I traced whether that's safe:
- `HtmlDocView` renders it via `<iframe sandbox="allow-same-origin" srcDoc={html}>` — **`allow-scripts` is never granted**, so the browser blocks all script execution (inline `<script>`, event handlers, `javascript:` URLs) regardless of same-origin. The test locks this precisely (`<script>window.__pwned=1</script>` stays in srcdoc but does not run; `sandbox` asserted to be exactly `allow-same-origin` and to NOT contain `allow-scripts`).
- Not sanitizing is a *deliberate, valid* choice here — it preserves agent CSS (`<style>`, inline style, stylesheet links) intact, and the sandbox is the isolation boundary. `absolutizeDocAssetUrls` additionally disables editable controls and only rewrites `/assets/` URLs (rejecting absolute/`javascript:`/protocol-relative via `isAbsoluteOrSpecialUrl`).
- The comment panel renders anchor/comment text through JSX text nodes (`{thread.text}`, `{r.text}`) — React-escaped, no `innerHTML`. No secondary sink.
So there is **no XSS blocker** on the live path. Good, careful design.

### Minor
- `packages/docs/src/html/HtmlDocView.tsx:48` — `sanitizeDocHtml` is exported "for legacy callers" but is **called nowhere** in this PR (verified across the `html/` module). Dead code that actively misleads: a reader sees a DOMPurify sanitizer and may assume the rendered doc is sanitized, when the actual isolation is the sandbox. Either wire it where a stripped fragment is genuinely needed, or drop it and let the sandbox comment stand alone.

### Praise
- The security posture is documented at exactly the right place — the file-header SECURITY block states the threat model (backend doesn't sanitize → sandbox isolates) instead of leaving it implicit, and the tests assert the *negative* security invariants (no `allow-scripts`, no `allow-top-navigation`, script-in-srcdoc-doesn't-run). That's the right way to lock a sandbox contract against a future well-meaning "just add allow-scripts" edit.
- Withholding `allow-top-navigation` on DocTab with a reverse-assertion test is a sharp touch — tab-hijack via `top.location` is the exact thing people forget.

### Net
No XSS blocker — the untrusted-HTML render path is correctly isolated by a script-disabled sandbox, and I verified it against the hostile-`<script>` case. What blocks merge is mechanical + process: fix the red `pr-title-lint`, fill the empty template (especially write down the sandbox-not-sanitize security model), address the DocTab cross-origin invariant before Stage B wires it, and clean up the dead `sanitizeDocHtml`. Retitle + description + the DocTab guard and this is in good shape.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #787 (octo-web)

## Summary

This PR introduces a read-only HTML document viewer for agent-authored documents (`docType: 'html'`). The feature adds a new `HtmlDocView` component that fetches published HTML from the octo-doc backend (a separate deployment from the same-origin Yjs docs API), renders it in a sandboxed iframe for full CSS fidelity, and provides a side-panel comment rail with text/element anchoring and a "让 AI 处理" forward-to-chat bridge.

The routing integration in `DocsHome.tsx` properly threads `octoDocSlug` through all open paths (list click, persisted sessionStorage, getDoc fallback). A new `DocTab` component is added in dmworkbase for future iframe embedding with a different (script-enabled) sandbox policy, correctly differentiated from the read-only `HtmlDocView` sandbox. Test coverage is thorough: unit tests cover pure helpers, data-layer contracts, component rendering, iframe sandbox assertions, and end-to-end routing for the new doc type.

## Verification

Static analysis only at head `949688f94ea87e77a05193dca375d8082dd6df8f`; build and tests not executed in this environment.

- ✅ **Iframe sandbox isolation** — `HtmlDocView` uses `sandbox="allow-same-origin"` (no `allow-scripts`), blocking script execution in agent-authored HTML. Tests lock this assertion (`HtmlDocView.test.tsx`).
- ✅ **Auth headers** — all octo-doc fetches (render, comments, grants) carry `credentials: 'include'` and the octo `token` header via `getWKApp().loginInfo?.token`, consistent with the octo-doc auth scheme.
- ✅ **Routing parity** — `octoDocSlug` is threaded through `DocTarget`, `persistDocTarget`, `readDocTarget`, `openDoc`, `commitOpen`, and `buildRightPane` in `DocsHome.tsx`. All three open paths (list row click, sessionStorage restore, getDoc fallback) correctly dispatch to `HtmlDocView` when `docType === 'html'`.
- ✅ **Data layer separation** — octo-doc comments/grants use raw `fetch` against `resolveOctoDocBase()`, never the same-origin `apiClient`. The `grantsUrl` correctly uses `/grants/{slug}` to avoid collision with the SPA's `/docs` route.
- ✅ **Comment contract** — `createComment` enforces exclusive `parent_id` XOR `anchor`, preventing the ambiguous `{parent_id, anchor}` pair. Tests verify both branches.
- ✅ **Selection anchor lifecycle** — anchor is locked on selection, preserved across focus/blur of the comment textarea, and only cleared by the explicit clear action or successful post. Tests cover all three paths.
- ✅ **Asset URL rewriting** — `absolutizeDocAssetUrls` only rewrites URLs containing `/assets/` against the real document URL, leaving absolute URLs and non-asset relative URLs untouched. Tested with signed query params, relative paths, and mixed content.
- ✅ **i18n parity** — both `en-US.json` and `zh-CN.json` add the same set of keys (comment composer, AI handle, header, error strings).
- ✅ **`sanitizeDocHtml` preserved** — legacy DOMPurify-based sanitizer is retained and tested for external callers, clearly separated from the iframe render path.

## Findings

### P2 — `ListCommentsResponse` type definition diverges from actual backend envelope (`packages/docs/src/html/htmlDocComments.ts:78`)

The `ListCommentsResponse` interface declares `{ roots: OctoDocCommentThread[] }`, but `listComments` reads from `data?.data ?? []` — matching the backend's actual `{data, pagination}` writeList envelope. The runtime behavior is correct and tests lock the `{data: [...]}` shape, but the dead type definition can mislead future developers extending this API.

```ts
// Current type (dead code — never consumed):
export interface ListCommentsResponse {
  roots: OctoDocCommentThread[]
}

// Actual read in listComments:
const data = (await res.json()) as { data?: OctoDocCommentThread[] } | null
return data?.data ?? []
```

Fix: update `ListCommentsResponse` to `{ data: OctoDocCommentThread[]; pagination?: unknown }` or remove it since it's unused.

### P2 — Sequential `addGrant` calls in member panel (`packages/docs/src/html/HtmlMemberPanel.tsx:72`)

`onAdd` iterates over selected UIDs with sequential `await addGrant(slug, uid, 'reader')` calls. For bulk member additions (e.g. adding 5+ members at once), this produces N sequential round-trips to the octo-doc backend, each with its own latency.

```ts
for (const uid of uids) await addGrant(slug, uid.trim(), 'reader')
```

Not a correctness issue, but worth either parallelizing with `Promise.all` or batching via a single backend endpoint when the grants API supports it.

- **`parseOdocMeta` regex fragility** — `__ODOC__\s*=\s*(\{[\s\S]*?\});` uses non-greedy match that can misparse objects containing `};` in string values, but a parse miss just yields `null` meta (graceful fallback to slug for header). Not a correctness concern.
- **`cssAttrValue` incomplete CSS escaping** — doesn't escape `]` in aid values for `querySelector`, but the `try/catch` in `resolveHtmlDocAnchorText` returns `null` on any selector error. Agent-authored aids are typically alphanumeric.
- **Non-asset relative URLs don't resolve in srcDoc** — by design: `resolveDocAssetUrl` returns `null` for URLs without `/assets/`, so they stay relative and resolve against `about:srcdoc` (which won't load). Only assets need the real document URL.
- **`DocTab` vs `HtmlDocView` sandbox difference** — `DocTab` uses `allow-scripts allow-same-origin allow-forms allow-popups` for embedding full octo-doc service pages that need JS for auth; `HtmlDocView` uses `allow-same-origin` only for rendering raw agent HTML read-only. Two different threat models, two correct sandbox policies.
- **`resolveOctoDocBase` same-origin default** — returns empty string when unconfigured, making fetches go to the current host. The error state surfaces the attempted URL and a misconfig warning, making deployment issues immediately diagnosable.

## Data-Flow Trace

|---|---|---|
| `HtmlDocView` render | `fetch(resolveOctoDocBase() + '/v1/docs/' + slug)` → `res.text()` → `sanitizeDocHtml()` → `iframe srcDoc` | ✅ `resolveOctoDocBase` from `__OCTO_DOC_BASE__` window global or empty-string fallback |
| `HtmlDocCommentPanel` thread list | `listComments(slug, version)` → `fetch(base + '/comments?slug=...&version=...')` → `data?.data ?? []` | ✅ credentials:'include' + token header asserted in test |
| Comment submission | `createComment(slug, {text, version, anchor})` → POST → `data?.id` | ✅ exclusive parent_id/anchor contract; test locks both branches |
| "让 AI 处理" forward | `buildAgentInstruction(slug, version, anchor, text, commentId)` → `wk.openDocForward({title, link})` | ✅ `canForwardToChat()` gates button; link includes commentId + aid/slug |
| Anchor label rendering | `resolveAnchorText(anchor)` callback from parent → iframe content lookup | ✅ `HtmlDocView` passes callback that queries iframe DOM; test verifies delegation |
| Member management | `listGrants(slug)` / `addGrant(slug, uid, role)` / `removeGrant(slug, uid)` → `fetch(base + '/grants/' + slug)` | ✅ `/grants/{slug}` path avoids `/docs` route collision |
| `openDoc` routing | `docType === 'html'` → `HtmlDocView` with `octoDocSlug` prop | ✅ octoDocSlug threaded through DocTarget, sessionStorage, getDoc fallback |

## Blind-Point Checklist (C1–C6)

- **C1 (dual-path parity):** N/A — comment/grant flows are independent operations; no symmetric create/remove pair that must be balanced.
- **C2 (control-flow ordering):** N/A — `HtmlDocView` is a standalone component; no nested reuse across call-sites that could cause double-subscription or ordering issues.
- **C3 (auth boundary):** Checked — grants API uses `getWKApp().loginInfo?.token` for identity; backend validates. Only author (creator) can manage members, enforced in `HtmlMemberPanel` via `wk.loginInfo?.uid === props.author`. Reader-only role is the only grantable role. ✅
- **C4 (lifecycle cascade):** N/A — no auth/container state cascade; iframe content is static HTML, no auth-gated state to invalidate.
- **C5 (build ≠ runtime):** Checked — `__OCTO_DOC_BASE__` is a build-time injected constant (vite define). `resolveOctoDocBase()` correctly falls back to empty string at runtime if not injected, making the feature deployable to the same origin without build config. ✅
- **C6 (docs/policy):** N/A — no documentation or policy changes.

## Verdict: APPROVED

No P0/P1 issues found. The HTML document viewer is well-architected with clean separation between the octo-doc backend and the same-origin Yjs docs API. The security model (iframe sandbox without `allow-scripts`, documented defense-in-depth trade-off) is correctly implemented and locked by tests. The two P2 items (dead type definition, sequential grant calls) are non-blocking polish.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; 2 P2 polish items are non-blocking.

yujiawei (changes requested):
# Code Review — PR #787 (octo-web)

**Head SHA:** `949688f94ea87e77a05193dca375d8082dd6df8f`
**Base:** `main` (merge-base `d80703a5`) · note: PR currently reports **CONFLICTING** against `main` — rebase before merge.

Reviewed the full diff (23 files, +2805/-30): the read-only agent-authored HTML doc viewer (`HtmlDocView`), its comment rail (`HtmlDocCommentPanel` / `htmlDocComments`), member grants (`HtmlMemberPanel` / `htmlGrantsApi`), the `DocsHome` routing wiring for the new `html` doc kind, and the standalone `DocTab` skeleton in `dmworkbase`.

The core security model of `HtmlDocView` is sound: untrusted agent HTML is rendered in an iframe with `sandbox="allow-same-origin"` and **no** `allow-scripts`, so inline `<script>`/`on*` handlers never execute; only `/assets/` URLs are absolutized and editable controls are neutralized. Tests cover this well. Two things block merge below.

---

## 1. Spec / scope

No linked spec was available to diff against, so this is assessed as a standard change, not a plan-compliance check.

One scope observation (not a blocker): this PR introduces **two** independent iframe-embedding components with **different** sandbox policies —
- `HtmlDocView` (docs pkg): `sandbox="allow-same-origin"`, no scripts — the read-only viewer actually wired into `DocsHome`.
- `DocTab` (dmworkbase): `sandbox="allow-scripts allow-same-origin allow-forms allow-popups"` — a "Stage A skeleton" that is **not imported or rendered anywhere** (verified: only its own `index.ts` re-exports it).

Shipping an unused, script-enabled iframe component alongside the real script-disabled one invites a future caller to reach for the wrong one. Consider landing `DocTab` when its consumer lands, or at minimum calling out the divergence in the PR description.

## 2. Findings

### P1 — Member-management gate is fail-open when `creator_uid` is absent
`packages/docs/src/html/HtmlMemberPanel.tsx:37`
```ts
const canManage = !!creatorUid && !!currentUid && creatorUid === currentUid
```
Both inputs come from the **same** `__ODOC__` blob at the call site (`HtmlDocView.tsx:398-399`):
```ts
creatorUid={meta?.creator_uid || meta?.identity?.login}
currentUid={meta?.identity?.login}
```
When `meta.creator_uid` is missing, `creatorUid` falls back to `meta.identity.login`, which is exactly `currentUid`. `canManage` then reduces to `identity.login === identity.login` → **true for every viewer**, so the management panel (add/remove grants) renders for non-authors. `OctoDocMeta` itself documents that some creator fields "are absent and fall back gracefully" (`HtmlDocView.tsx:189-196`), so this path is reachable, not hypothetical.

Blast radius depends on whether the octo-doc `/grants` backend independently enforces author-only writes (see Coverage). If it does, the user-visible effect is a broken panel that 403s; if it does not, a non-author can grant access. Either way the client gate is wrong: there is no case where "creator unknown ⇒ treat the viewer as the creator" is correct. Fix: drop the `|| meta?.identity?.login` fallback for `creatorUid`, and source `currentUid` from an authoritative current-user (e.g. `getWKApp().loginInfo`) rather than the doc's own meta blob.

### P2 — Successful grants stay invisible when a later add throws
`packages/docs/src/html/HtmlMemberPanel.tsx:60-71`
```ts
for (const uid of uids) await addGrant(slug, uid.trim(), 'reader')
await refresh()
```
The loop is sequential; if the Nth `addGrant` throws, control jumps to `catch` and `refresh()` is skipped, so the grants that already succeeded (1..N-1) are persisted on the backend but never appear in the list until the panel is reopened. Move `refresh()` into `finally`, or use `Promise.allSettled` and refresh regardless.

### P2 — `DocTab` injects `src` into a script-enabled iframe with no protocol check
`packages/dmworkbase/src/Components/DocTab/DocTab.tsx:63-73`
`src` is passed straight to the iframe while `sandbox` includes `allow-scripts`. The component delegates URL assembly to callers, but if any future caller reflects unvalidated input (e.g. a query param) into `src`, a `javascript:` URI becomes a direct XSS. Add a cheap `http(s):`-only guard inside the component as defense-in-depth. (Latent only — component is currently unused.)

### P2 — `DocTab` `allow-scripts` + `allow-same-origin` is a sandbox escape iff same-origin
`packages/dmworkbase/src/Components/DocTab/DocTab.tsx:73`
Per the HTML spec, an iframe that is same-origin **and** has both `allow-scripts` and `allow-same-origin` can rewrite its own `sandbox` attribute and reach into the parent. This is only safe if the doc service is guaranteed to be a distinct origin from the host app. The inline comment asserts same-origin is required for the capability cookie, which is exactly the tension — please document/verify the origin separation this relies on before wiring `DocTab` up.

### P2 (nit) — `parseOdocMeta` regex over untrusted HTML
`packages/docs/src/html/HtmlDocView.tsx:201`
`/__ODOC__\s*=\s*(\{[\s\S]*?\});/` scans the whole untrusted payload; the lazy quantifier is linear (not catastrophic), but extracting the JSON from a dedicated `<script type="application/json">` node via DOM would be more robust than string-matching. Low priority.

## 3. Coverage / blind spots
- **Backend authorization for `/grants` and `/comments`** lives in the octo-doc deployment (separate repo) and was not reviewable here. The severity of the P1 gate hinges on it; please confirm `/grants` PUT/DELETE and the comment endpoints enforce author/reader server-side and do not trust the client.
- **Origin configuration** (`VITE_OCTO_DOC_BASE` / `__OCTO_DOC_BASE__` vs same-origin default) determines the real risk of the `DocTab` sandbox point above; not verifiable from the diff.
- Not independently re-run: the package test suites and typecheck (no `node_modules` in the review workspace). The added tests read well and cover the sanitize/render/anchor paths.

One earlier concern was dismissed on verification: `dompurify@3.4.2` ships its own type definitions (the `@types/dompurify` stub is deprecated for exactly this reason, and it is already present transitively), so the added dep does not break the typecheck.

## 4. Verdict
The read-only viewer and its XSS isolation are well done. The fail-open member-management gate (P1) must be fixed before merge, and the sequential-grant refresh gap (P2) is a quick, worthwhile fix. Please also confirm the octo-doc backend enforces grant/comment authorization server-side.

**Requesting changes.**

PR #154, opened 2026-05-27
chore(ci): add edited trigger to check-sprint and auto-add workflows

lml2468:
This PR adds the `edited` event to the `pull_request_target.types` trigger in both `check-sprint.yml` and `auto-add-to-project.yml`, enabling the Issue-first Sprint check flow to work correctly when a developer adds a `Closes #<issue>` reference to a PR description after opening.

PR #153, opened 2026-05-27
fix(file): prefer filename suffix over content.extension for preview detection

quanming1:
Closes #143

## 问题

Web 端聊天中收到 `.md` 文件，点击预览弹出「暂不支持预览此文件类型」，只能下载。

## 根因

文件预览的扩展名提取函数 `getExtension(extension, name)` 优先使用 `content.extension`，文件名后缀仅作 fallback：

```ts
// 修改前
function getExtension(extension: string, name?: string): string {
  const ext = (extension || "").toLowerCase();
  if (ext) return ext;            // ← 优先返回, 永远先用 content.extension
  if (name) {
    const dot = name.lastIndexOf(".");
    if (dot >= 0) return name.substring(dot + 1).toLowerCase();
  }
  return "";
}
```

服务端返回的 `content.extension` 不可靠 — 实测有 `.md` 文件该字段为空或返回 `"file"` 等占位值。结果：

- `ThreadPanel` 的 `isMarkdown` 检查（`["md", "markdown"]`）不命中
- `FileRendererRegistry` 也匹配不到
- 最终走 `FallbackRenderer` 显示「暂不支持预览此文件类型」

## 修改

把优先级翻转 — **文件名后缀优先，`content.extension` 作为 fallback**：

```ts
// 修改后
function getExtension(extension: string, name?: string): string {
  if (name) {
    const dot = name.lastIndexOf(".");
    if (dot >= 0) return name.substring(dot + 1).toLowerCase();
  }
  const ext = (extension || "").toLowerCase();
  if (ext) return ext;
  return "";
}
```

`getExtension` 在代码库中有两处独立定义，逻辑一致，**都改**：

| 文件 | 调用方 |
|------|--------|
| `packages/dmworkbase/src/Messages/File/index.tsx` | `getFileIconInfo`、`FileTypeIcon`、`isPreviewable`、`isTextFile`、`handlePreview` |
| `packages/dmworkbase/src/Components/FilePreviewPanel/types.ts` | `FilePreviewPanel`、`registry`、`ThreadPanel`、`MergeforwardMessageList` |

## 边缘情况

- 文件名无后缀（`Makefile` / `Dockerfile` 等）但 `content.extension` 有值 — 仍走 fallback 路径，行为不变
- 文件名后缀为空字符串（如 `foo.`）— `lastIndexOf` 命中后返回空串，进而 fallback 到 `content.extension`

## 验证

[...]

PR #790, opened 2026-07-15
feat(summary): add /s/:taskNo standalone deep-link route to summary detail (rebased on main, no conflict)

Ranwanglc:
## Summary
Re-implements PR #740's思路 (smart-summary notification card → click → land on detail page) on **latest `main`**, on a fresh branch off `main` to avoid the merge conflict #740 now has.

#740 conflicts with `main` only in `SummaryDetailPage.tsx` (the refine/regenerate/version-restore feature merged 841 lines there). This PR hand-weaves the same taskNo-support logic into the current file, so there is **no conflict**. The other 12 files are functionally identical to #740.

Supersedes #740 and the older #680 (`/summary/detail?taskId=` landing). Pairs with octo-smart-summary #153 (notify card) and #154 (resolve detail by `task_no`).

## What & why
The completed-summary notification card links to `/s/{task_no}?sp={space_id}`, but octo-web had no such route, so a click bounced to login and lost the target. Also, the card carries a **string `task_no`** (not the numeric `task_id`), which the detail page previously couldn't consume.

## Changes
- `apps/web/src/Layout/index.tsx`: new `/s/:taskNo` standalone route, mirrors `/d/:docId` incl. login-redirect fallback + `sp` space forwarding.
- `packages/dmworkbase/.../InteractiveCard/renderer/actions.ts`: intercept `/s/<taskNo>` card clicks → internal `WKApp.openSummaryDetail(taskNo)`.
- `SummaryDetailPage.tsx`: accept `string` task_no. `detailLookupId` (raw prop) drives the initial fetch + change-detection; numeric `taskId` is derived from `detail.task_id`. BY_PERSON secondary loads (`loadPersonalResult`/`loadMembers`) take the explicit numeric `detail.task_id` to avoid a setState race (this was a known trap: string prop → `this.taskId` null before `setState({detail})` commits).
- `StandaloneDocPage.tsx`: extend the post-login safe-return open-redirect guard to also allow `/s/:taskNo`.
- widen `taskId` types across `summaryApi.getSummaryDetail`, `useSummaryDetail`, `module.tsx`, `App.tsx`.
- tests: route parsing (`layoutStandaloneDocPath`), card-action routing (`InteractiveCard/actions`), and a `string task_no` detail-load race test.

## Testing (local, latest main)
- `pnpm turbo run build --filter=@octo/web` → ✓ built (production bundle clean).
- `packages/dmworksummary` full suite: **377 passed** (incl. new string-task_no test).
- `packages/dmworkbase` InteractiveCard actions: 4 passed.
- `apps/web` layoutStandaloneDocPath: 9 passed.

## 部署前提 / 必配环境变量
本 PR 为**纯前端路由/交互改动，不新增任何必配环境变量**。端到端生效依赖已有的两侧配套：
- octo-smart-summary #153（worker 发卡片）+ #154（server 按 `task_no` 解析详情）需一并部署，否则卡片链接/详情查询不通。
- 无新增 build-time / runtime env。

### DIFF
--- a/packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts
@@ -1,4 +1,7 @@
+import WKApp from "../../../App";
 import { isSafeUrl } from "../../../Utils/security";
+
+const SUMMARY_DETAIL_PATH_RE = /^\/s\/([A-Za-z0-9_-]+)\/?$/;
 
 /**
  * Action.OpenUrl 导航。
@@ -11,6 +14,24 @@ import { isSafeUrl } from "../../../Utils/security";
 /** 在新标签打开 URL；提交前二次 isSafeUrl 校验（http/https），非法直接忽略。 */
 export function openUrl(url: string): void {
   if (!isSafeUrl(url)) return;
+
+  // isSafeUrl 已保证 http/https 绝对 URL（new URL 不会抛）；try 兼容未来 isSafeUrl 放宽相对路径的情况。
+  let summaryTaskNo: string | null = null;
+  let summarySpace: string | undefined;
+  try {
+    const u = new URL(url);
+    summaryTaskNo = u.pathname.match(SUMMARY_DETAIL_PATH_RE)?.[1] ?? null;
+    // 深链带的空间（sp||space）随任务号一起透传，避免跨空间详情解析到当前空间→404。
+    summarySpace = u.searchParams.get("sp") || u.searchParams.get("space") || undefined;
+  } catch {
+    summaryTaskNo = null;
+  }
+  if (summaryTaskNo && WKApp.openSummaryDetail) {
+    // 卡片深链可能来自 https 生产域，本地调试是 http/端口；/s/<taskNo> 路径本身是内部详情信号。
+    WKApp.openSummaryDetail(summaryTaskNo, summarySpace);
+    return;
+  }
+
   window.open(url, "_blank", "noopener,noreferrer");
 }
 
--- a/apps/web/src/Layout/index.tsx
+++ b/apps/web/src/Layout/index.tsx
@@ -14,6 +14,7 @@ import InviteLanding from "../Components/InviteLanding";
 import JoinSpacePage from "../Components/JoinSpacePage";
 import JoinApprovalResult from "../Components/JoinApprovalResult";
 import { StandaloneDocPage, parseStandaloneDocId, isStandaloneDocPath, persistStandaloneReturn, consumeStandaloneReturn, withReturnSid } from "@octo/docs";
+import { SummaryDetailPage } from "@dmwork/summary";
 import { adoptStoredSession, findSidForToken, clearSessionsWithToken } from "./recoverSession";
 import { isLoopCliAuthorizePath, LOOP_CLI_AUTHORIZE_PATH } from "@octo/loop";
 
@@ -75,6 +76,28 @@ function clearExpiredStandaloneSessionAndReload(): void {
     if (typeof window !== "undefined") window.location.reload();
 }
 
+/** `/s/:taskNo` — summary taskNo is a single URL path segment, optional trailing slash. */
+const STANDALONE_SUMMARY_PATH = /^\/s\/([A-Za-z0-9_-]+)\/?$/;
+
+export function parseStandaloneSummaryTaskNo(pathname: string): string | null {
+    if (typeof pathname !== "string") return null;
+    const m = STANDALONE_SUMMARY_PATH.exec(pathname);
+    return m ? m[1] : null;
+}
+
+// Only a well-formed single-segment `/s/<taskNo>` counts as standalone; `/s`,
+// `/s/`, and nested `/s/a/b` must fall through (else the render branch mounts
+// SummaryDetailPage with an undefined taskId).
+export function isStandaloneSummaryPath(pathname: string): boolean {
+    return parseStandaloneSummaryTaskNo(pathname) !== null;
+}
+
+function applyStandaloneSummarySpaceFromQuery(): void {
+    const params = new URLSearchParams(window.location.search);
+    const spaceId = params.get("sp") || params.get("space") || "";
+    if (spaceId) WKApp.shared.currentSpaceId = spaceId;
+}
+
 export default class AppLayout extends Component<{}, AppLayoutState> {
     state: AppLayoutState = { showJoinSpace: false };
 
@@ -124,12 +147,18 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             const forwardDoc = getQueryParam("doc") || ""
             const forwardSpace = getQueryParam("space") || ""
             const forwardFolder = getQueryParam("folder") || ""
+            const forwardSp = getQueryParam("sp") || ""
             const redirectQuery = new URLSearchParams()
             if (existingSid) redirectQuery.set("sid", existingSid)
             if (forwardDoc) {
                 redirectQuery.set("doc", forwardDoc)
                 if (forwardSpace) redirectQuery.set("space", forwardSpace)
                 if (forwardFolder) redirectQuery.set("folder", forwardFolder)
+            }
+            if (isStandaloneSummaryPath(window.location.pathname)) {
+                // 与 applyStandaloneSummarySpaceFromQuery(sp||space) 对称：登录重定向也一并透传 space。
+                if (forwardSp) redirectQuery.set("sp", forwardSp)
+                if (forwardSpace) redirectQuery.set("space", forwardSpace)
             }
             const redirectQs = redirectQuery.toString()
             const sidParam = redirectQs ? `?${redirectQs}` : ""
@@ -404,6 +433,25 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             // the login screen (below) without navigating away. The standalone page itself only
             // mounts once a token exists, so the anonymous path is the ONLY place this key gets
             // written for a first-time visitor — onLogin consumes it via consumeStandaloneReturn.
+            persistStandaloneReturn();
+            // Anonymous: fall through to the login screen (below) without navigating away.
+        }
+
+        // Standalone summary deep-link (`/s/:taskNo`): notification cards use task_no (not numeric
+        // task_id), so pass the raw path segment into the detail fetch. Auth/session recovery mirrors
+        // `/d/:docId`; anonymous visitors stash the exact target and return after login.
+        if (isStandaloneSummaryPath(window.location.pathname)) {
+            if (!WKApp.loginInfo.token) {
+                WKApp.loginInfo.load();
+            }
+            if (!WKApp.loginInfo.token) {
+                recoverOctoSessionFromStorage(true);
+            }
+            if (WKApp.loginInfo.token) {
+                applyStandaloneSummarySpaceFromQuery();
+                const standaloneTaskNo = parseStandaloneSummaryTaskNo(window.location.pathname);
+                return <SummaryDetailPage taskId={standaloneTaskNo ?? undefined} />;
+            }
             persistStandaloneReturn();
             // Anonymous: fall through to the login screen (below) without navigating away.
         }
--- a/apps/web/src/__tests__/layoutStandaloneDocPath.test.ts
+++ b/apps/web/src/__tests__/layoutStandaloneDocPath.test.ts
@@ -71,3 +71,54 @@ describe('Layout — standalone /d/:docId clean cold-load path', () => {
     expect(layout).toMatch(/window\.location\.reload\(\)/)
   })
 })
+
+describe('Layout — standalone /s/:taskNo summary clean cold-load path', () => {
+  let layout: string
+
+  beforeAll(() => {
+    layout = fs.readFileSync(path.join(__dirname, '../Layout/index.tsx'), 'utf-8')
+  })
+
+  it('matches only a well-formed single-segment /s/<taskNo>, not /s, /s/, or nested paths', () => {
+    expect(layout).toMatch(/isStandaloneSummaryPath\(\s*window\.location\.pathname\s*\)/)
+    expect(layout).toMatch(/function\s+parseStandaloneSummaryTaskNo/)
+    // The matcher must delegate to the strict parser (single segment) rather than a
+    // loose /s namespace regex, so /s, /s/, and /s/a/b fall through instead of
+    // mounting SummaryDetailPage with an undefined taskId.
+    expect(layout).toMatch(/return\s+parseStandaloneSummaryTaskNo\(pathname\)\s*!==\s*null/)
+    expect(layout).not.toMatch(/STANDALONE_SUMMARY_NAMESPACE/)
+    expect(layout).not.toMatch(/if\s*\(\s*standaloneTaskNo\s*\)/)
+    // The strict single-segment path regex is the sole source of truth.
+    expect(layout).toMatch(/STANDALONE_SUMMARY_PATH\s*=\s*\/\^\\\/s\\\/\(\[A-Za-z0-9_-\]\+\)/)
+  })
+
+  it('recovers a stored session before rendering the summary detail page', () => {
+    const branchIdx = layout.indexOf('Standalone summary deep-link')
+    expect(branchIdx).toBeGreaterThan(0)
+    const nsIdx = layout.indexOf('isStandaloneSummaryPath(window.location.pathname)', branchIdx)
+    expect(nsIdx).toBeGreaterThan(0)
+    const recoverIdx = layout.indexOf('recoverOctoSessionFromStorage(true)', nsIdx)
+    const renderIdx = layout.indexOf('<SummaryDetailPage', nsIdx)
+    expect(recoverIdx).toBeGreaterThan(nsIdx)
+    expect(renderIdx).toBeGreaterThan(recoverIdx)
+  })
+
+  it('renders summary detail only with a token, passing the raw task_no string through', () => {
+    expect(layout).toMatch(/if\s*\(\s*WKApp\.loginInfo\.token\s*\)\s*\{[\s\S]*?<SummaryDetailPage/)
+    expect(layout).toMatch(/const\s+standaloneTaskNo\s*=\s*parseStandaloneSummaryTaskNo\(window\.location\.pathname\)/)
+    expect(layout).toMatch(/<SummaryDetailPage\s+taskId=\{standaloneTaskNo\s*\?\?\s*undefined\}/)
+  })
+
+  it('stashes anonymous /s targets and carries sp through post-login return', () => {
+    const branchIdx = layout.indexOf('Standalone summary deep-link')
+    expect(branchIdx).toBeGreaterThan(0)
+    const nsIdx = layout.indexOf('isStandaloneSummaryPath(window.location.pathname)', branchIdx)
+    expect(nsIdx).toBeGreaterThan(0)
+    const stashIdx = layout.indexOf('persistStandaloneReturn()', nsIdx)
+    expect(stashIdx).toBeGreaterThan(nsIdx)
+    expect(layout).toMatch(/const\s+forwardSp\s*=\s*getQueryParam\("sp"\)\s*\|\|\s*""/)
+    expect(layout).toMatch(/redirectQuery\.set\("sp",\s*forwardSp\)/)
+    expect(layout).toMatch(/consumeStandaloneReturn\(\)/)
+    expect(layout).toMatch(/withReturnSid\(standaloneReturn,\s*sessionSid\)/)
+  })
+})
--- a/packages/dmworkbase/src/App.tsx
+++ b/packages/dmworkbase/src/App.tsx
@@ -703,7 +703,7 @@ export default class WKApp extends ProviderListener {
   static menus = MenusManager.shared; // 菜单
   // Callback to switch the active sidebar menu by id (set by Main page)
   static switchToMenuById?: (menuId: string) => void;
-  static openSummaryDetail?: (taskId: number) => void;
+  static openSummaryDetail?: (taskId: number | string, spaceId?: string) => void;
   static searchChatCandidates?: (params: { keyword?: string; chat_type?: string; space_id?: string }) => Promise<any[]>;
   // Id of the currently active sidebar menu (kept in sync by Main page)
   static currentMenuId?: string;
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/actions.test.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/actions.test.ts
@@ -1,0 +1,73 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { openSummaryDetailMock, wkAppMock } = vi.hoisted(() => {
+  const openSummaryDetailMock = vi.fn();
+  return {
+    openSummaryDetailMock,
+    wkAppMock: {
+      openSummaryDetail: openSummaryDetailMock as ((taskId: number | string, spaceId?: string) => void) | undefined,
+    },
+  };
+});
+
+vi.mock("../../../App", () => ({
+  default: wkAppMock,
+}));
+
+import { openUrl } from "../renderer/actions";
+
+describe("openUrl", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+    openSummaryDetailMock.mockReset();
+    wkAppMock.openSummaryDetail = openSummaryDetailMock;
+    vi.spyOn(window, "open").mockImplementation(() => null);
+  });
+
+  it("routes /s/<taskNo> summary links internally", () => {
+    openUrl("https://web.example.com/s/ST202607145kstyh08?sp=space-1");
+
+    expect(openSummaryDetailMock).toHaveBeenCalledTimes(1);
+    expect(openSummaryDetailMock).toHaveBeenCalledWith("ST202607145kstyh08", "space-1");
+    expect(window.open).not.toHaveBeenCalled();
+  });
+
+  it("routes /s/<taskNo> without sp with undefined space", () => {
+    openUrl("https://web.example.com/s/ST202607145kstyh08");
+
+    expect(openSummaryDetailMock).toHaveBeenCalledTimes(1);
+    expect(openSummaryDetailMock).toHaveBeenCalledWith("ST202607145kstyh08", undefined);
+    expect(window.open).not.toHaveBeenCalled();
+  });
+
+  it("opens external https links in a new tab", () => {
+    openUrl("https://web.example.com/docs/ST202607145kstyh08");
+
+    expect(openSummaryDetailMock).not.toHaveBeenCalled();
+    expect(window.open).toHaveBeenCalledWith(
+      "https://web.example.com/docs/ST202607145kstyh08",
+      "_blank",
+      "noopener,noreferrer"
+    );
+  });
+
+  it("ignores unsafe urls", () => {
+    openUrl("javascript:alert(1)");
+
+    expect(openSummaryDetailMock).not.toHaveBeenCalled();
+    expect(window.open).not.toHaveBeenCalled();
+  });
+
+  it("falls back to window.open when summary routing is unavailable", () => {
+    wkAppMock.openSummaryDetail = undefined;
+
+    openUrl("https://web.example.com/s/ST202607145kstyh08?sp=space-1");
+
+    expect(openSummaryDetailMock).not.toHaveBeenCalled();
+    expect(window.open).toHaveBeenCalledWith(
+      "https://web.example.com/s/ST202607145kstyh08?sp=space-1",
+      "_blank",
+      "noopener,noreferrer"
+    );
+  });
+});
--- a/packages/dmworksummary/src/api/summaryApi.ts
+++ b/packages/dmworksummary/src/api/summaryApi.ts
@@ -143,8 +143,8 @@ export async function listSummaries(
     return get('/summaries', params as Record<string, unknown>, config);
 }
 
-export async function getSummaryDetail(taskId: number): Promise<SummaryDetail> {
-    return get(`/summaries/${taskId}`);
+export async function getSummaryDetail(taskId: number | string): Promise<SummaryDetail> {
+    return get(`/summaries/${encodeURIComponent(String(taskId))}`);
 }
 
 export async function deleteSummary(taskId: number): Promise<void> {
--- a/packages/dmworksummary/src/hooks/useSummaryDetail.ts
+++ b/packages/dmworksummary/src/hooks/useSummaryDetail.ts
@@ -12,7 +12,7 @@ interface UseSummaryDetailReturn {
     cancel: () => Promise<void>;
 }
 
-export function useSummaryDetail(taskId: number | null): UseSummaryDetailReturn {
+export function useSummaryDetail(taskId: number | string | null): UseSummaryDetailReturn {
     const [detail, setDetail] = useState<SummaryDetail | null>(null);
     const [loading, setLoading] = useState(false);
     const [error, setError] = useState<string | null>(null);
@@ -41,13 +41,13 @@ export function useSummaryDetail(taskId: number | null): UseSummaryDetailReturn 
     }, []);
 
     const regenerate = useCallback(async () => {
-        if (taskId == null) return;
+        if (typeof taskId !== "number") return;
         await api.regenerateSummary(taskId);
         refresh();
     }, [taskId, refresh]);
 
     const cancel = useCallback(async () => {
-        if (taskId == null) return;
+        if (typeof taskId !== "number") return;
         await api.cancelSummary(taskId);
         refresh();
     }, [taskId, refresh]);
--- a/packages/dmworksummary/src/index.tsx
+++ b/packages/dmworksummary/src/index.tsx
@@ -1,1 +1,2 @@
 export { SummaryModule } from "./module";
+export { default as SummaryDetailPage } from "./pages/SummaryDetailPage";
--- a/packages/dmworksummary/src/module.tsx
+++ b/packages/dmworksummary/src/module.tsx
@@ -55,7 +55,9 @@ export class SummaryModule implements IModule {
             "en-US": enUS,
         });
 
-        WKApp.openSummaryDetail = (taskId: number) => {
+        WKApp.openSummaryDetail = (taskId: number | string, spaceId?: string) => {
+            // 卡片深链带的空间可能≠当前空间，路由前先切目标空间，与浏览器路由 applyStandaloneSummarySpaceFromQuery 对称。
+            if (spaceId) WKApp.shared.currentSpaceId = spaceId;
             WKApp.switchToMenuById?.("summary");
             WKApp.routeLeft.popToRoot();
             WKApp.routeRight.replaceToRoot(
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -49,7 +49,7 @@ import MemberSelectorModal from "../components/MemberSelectorModal";
 import type { MemberCandidate } from "../types/summary";
 
 interface SummaryDetailPageProps {
-    taskId?: number;
+    taskId?: number | string;
 }
 
 type RegenerateMode = "refine" | "full";
@@ -225,7 +225,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
 
     componentDidUpdate(prevProps: any, prevState?: SummaryDetailPageState) {
         const prevTaskId = prevProps.taskId;
-        const currentTaskId = this.taskId;
+        const currentTaskId = this.detailLookupId;
         if (prevTaskId !== currentTaskId && currentTaskId != null) {
             this.listPageActive = false;
             this.clearAllTimers();
@@ -371,6 +371,12 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     }
 
     get taskId(): number | null {
+        // 数字 taskId 直接用；字符串 taskNo 深链时先返回 null，待 detail 落库后回填 detail.task_id。
+        return typeof this.props.taskId === "number" ? this.props.taskId : this.state.detail?.task_id ?? null;
+    }
+
+    // 深链原始标识：数字 task_id 或字符串 task_no，仅用于首次 detail 拉取与切换检测。
+    private get detailLookupId(): number | string | null {
         return this.props.taskId ?? null;
     }
 
@@ -394,11 +400,12 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     }
 
     async loadDetail() {
-        if (this.taskId == null) return;
+        const lookupId = this.detailLookupId;
+        if (lookupId == null) return;
         // Blocking 5：每轮 detail 加载开始就 bump 序列号。这样旧 task 未完成的
         // loadSchedule（包括本函数下面发起的）都会被后续轮作废，不会回填到新 task。
         const seq = this.nextScheduleSeq();
-        const requestTaskId = this.taskId;
+        const requestTaskId = lookupId;
         // F1：切 task / 重拉 detail 时复位全部编辑态，避免旧 task 编辑态（尤其
         // editingTeamSummary）被带入新 task——否则切到非 creator 新 task 会绕过权限进编辑器。
         // FE-1（切任务竞态）：开始新 task 加载时同步清空上一 task 的 personalResult/members，
@@ -436,9 +443,9 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             restoringPersonalVersionId: null,
         });
         try {
-            const detail = await api.getSummaryDetail(this.taskId);
+            const detail = await api.getSummaryDetail(lookupId);
             // detail 本身也可能是旧请求：期间切了 task / 又发了一轮 loadDetail 就丢弃。
-            if (this.scheduleLoadSeq !== seq || this.taskId !== requestTaskId) return;
+            if (this.scheduleLoadSeq !== seq || this.detailLookupId !== requestTaskId) return;
             this.setState({
                 detail,
                 loading: false,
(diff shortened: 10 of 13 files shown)


OctoBoooot (dismissed):
## Review: /s/:taskNo standalone deep-link route to summary detail (#790)

**Verdict:** Approve — the security-sensitive pieces (post-login open-redirect guard, card-action URL interception) are verified sound, the known string-task_no setState race is correctly handled and tested, and code CI is green.

This re-implements #740's approach on current `main` (conflict-free). The two axes that matter for a deep-link + post-login-return feature are the open-redirect guard and the card-click URL handling; I verified both.

### Security — verified sound
- **Open-redirect guard (`StandaloneDocPage.tsx`, `isSafeReturnPath`)** — the `/s/:taskNo` addition is a *final-allowlist widening only*, layered after the existing structural checks: parse-failure → false, `javascript:`/non-http scheme → false, `url.origin !== origin` → false. Only then does `parseStandaloneDocId(pathname) !== null || STANDALONE_SUMMARY_PATH.test(pathname)` gate. So a tampered sessionStorage return value still can't bounce to `/settings`, an absolute `https://evil`, a `//host` scheme-relative, or a `/\host` smuggle — the origin/scheme gate is unchanged. The `/s/:taskNo` regex (`^\/s\/([A-Za-z0-9_-]+)\/?$`) mirrors `/d/:docId`'s tight single-segment charset. Good defense-in-depth.
- **Card-action interception (`InteractiveCard/renderer/actions.ts`, `openUrl`)** — runs *after* `isSafeUrl` (which requires an absolute http/https URL, so the subsequent `new URL(url)` can't throw). It matches `parsed.pathname` against `/s/<taskNo>` regardless of origin, but that's *safer*, not weaker: a hostile `https://evil.com/s/x` gets downgraded from "open evil.com in a new tab" to `WKApp.openSummaryDetail("x")` — an internal in-app navigation to our own summary detail (access-controlled server-side), never opening the attacker's origin. Non-summary URLs fall through to the existing `window.open(url, '_blank', 'noopener,noreferrer')`.

### Correctness — the string-task_no race is handled (verified)
The card carries a *string* `task_no`, not the numeric `task_id`, which is the trap the body calls out. The fix is right:
- `taskId` getter returns the numeric prop directly, else `state.detail?.task_id ?? null` (numeric derived once detail lands).
- `detailLookupId` (raw prop, number|string) drives the initial fetch + `componentDidUpdate` change-detection — so a string-taskNo nav is detected and triggers a load even before a numeric id exists (using `this.taskId` there would miss it, since it's null until detail arrives).
- The BY_PERSON secondary loads (`loadPersonalResult`/`loadMembers`) take the explicit just-fetched `detail.task_id` (numeric) rather than reading `this.taskId` mid-`setState` — avoiding exactly the "prop string → `this.taskId` null before setState commits" race. The seq-guard (`scheduleLoadSeq !== seq || detailLookupId !== requestTaskId`) still rejects stale responses.
- Covered by a dedicated `SummaryDetailPage.schedule.test.tsx` race test, plus route-parsing and card-action tests.

### CI
Green — Build / secret-scan / osv-scan / dependency-review / pr-title-lint all pass.

### Praise
- Layering `/s/:taskNo` as an allowlist *addition after* the origin/scheme checks (rather than a parallel hand-rolled check) is the right way to extend an open-redirect guard — the hardened structural rejections stay in force and the diff is auditable. The XIN-392 comment explaining the three clauses is exactly the context the next editor needs.
- Downgrading a summary card-click to internal `openSummaryDetail` instead of a new-tab open is a nice safety property — even a spoofed-origin card URL can't open the attacker's page.
- Threading the just-fetched `detail.task_id` into the secondary loads (instead of the still-null `this.taskId`) shows the setState race was actually understood, not papered over — and it's pinned by a test.

### Out of scope (informational, non-blocking)
- This supersedes **#740** and **#680** — worth closing those so a stale duplicate doesn't get merged later. Process note, not a code issue.
- End-to-end depends on octo-smart-summary #153/#154 shipping (notify card + `task_no` resolution); the frontend is backward-compatible without them, per the body.

### Net
Clean, well-scoped deep-link feature. Both security-critical paths verified (open-redirect guard widened safely; card interception downgrades hostile URLs to internal nav), the string-task_no race correctly avoided with the explicit numeric `detail.task_id` and pinned by a test, CI green. No blockers or majors. Credit @Ranwanglc — the guard-extension discipline and the race handling are both done right.

yujiawei (dismissed):
# Code Review — PR #790 (octo-web)

Reviewed at head SHA `a769089aecb1832efc022e8a70f1bf934e144a58` against `main` (merge-base `8ae5db6`). This PR adds a `/s/:taskNo` standalone deep-link route to the summary detail page, mirroring the existing `/d/:docId` doc route so completed-summary notification cards land on the detail view.

## Spec compliance

The diff does what the description promises and nothing more:

- **New route** `/s/:taskNo` in `Layout/index.tsx` — strict single-segment regex, auth/session recovery, and anonymous stash-and-return, all faithfully mirroring the `/d/:docId` branch.
- **Card interception** in `InteractiveCard/renderer/actions.ts` — `/s/<taskNo>` links route internally via `WKApp.openSummaryDetail(taskNo)`.
- **String `task_no` support** threaded through `summaryApi.getSummaryDetail`, `useSummaryDetail`, `module.tsx`, `App.tsx`, and `SummaryDetailPage`.
- **Open-redirect guard** in `StandaloneDocPage.tsx` extended to accept `/s/:taskNo` return targets.
- Tests added for route parsing, card-action routing, the string-`task_no` load race, and the return-path guard.

No over-building or scope creep observed. **Spec: ✅**

## Code quality

The string-`taskNo` vs numeric-`task_id` split is handled with real care: `detailLookupId` (raw prop) drives the initial fetch and change-detection, while `taskId` derives the numeric id from `detail.task_id` once loaded. BY_PERSON secondary loads (`loadPersonalResult`/`loadMembers`) are passed the explicit numeric `detail.task_id` to avoid the documented setState race. The security posture is sound — `openUrl` gates on `isSafeUrl` (which wraps `new URL` in try/catch) before its own `new URL` call, so a relative/invalid URL is rejected upstream and cannot throw there; the return-path guard keeps its control-char + same-origin + allowlist gates and only widens the final allowlist to `/s/:taskNo`.

No P0/P1 issues found. **Quality: Approved.**

### Advisory (non-blocking)

**P2 — In-app card click drops the `sp` query param.** In `actions.ts`, `WKApp.openSummaryDetail(taskNo)` is called with only the path segment; the `?sp=<space>` is discarded. The external `/s/:taskNo` deep-link path handles this (`applyStandaloneSummarySpaceFromQuery` reads `sp`/`space` before render), but the in-app interception does not update `WKApp.shared.currentSpaceId`. For a card whose target lives in a different space than the one currently active, the detail fetch would run against the wrong space. Low impact today (in-app cards are typically same-space), but worth forwarding the space to `openSummaryDetail` for symmetry with the deep-link path.

**P2 — Login-redirect forwards `sp` but not `space` for `/s` paths.** `Layout/index.tsx` only re-appends `forwardSp` (the `sp` param) to the post-login redirect for summary paths, while `applyStandaloneSummarySpaceFromQuery` accepts both `sp` and `space`. An anonymous visitor arriving at `/s/<taskNo>?space=<id>` would lose space context across the login bounce. Minor; align the two by also forwarding `space`.

**P2 (nit) — `useSummaryDetail` widening is currently dead code.** The hook now accepts `number | string | null`, and `regenerate`/`cancel` correctly guard `typeof taskId !== "number"` (those endpoints take numeric ids). The hook has no consumers in the tree, so this is speculative type-widening — correct, but note that if a future caller passes a string `taskNo`, regenerate/cancel will silently no-op by design.

### Coverage / blind spots

- End-to-end deep-link resolution depends on octo-smart-summary #153 (worker sends the card) and #154 (server resolves detail by `task_no`), plus the nginx `try_files` SPA fallback for `/s/*` — none verifiable from this frontend diff.
- `SummaryDetailPage` is rendered outside the app shell on the `/s` branch (no `MainPage`/nav rail). `routeLeft`/`routeRight` are static singletons and `I18nProvider` wraps at the app root, so context/routing calls are safe; the standalone view simply has no in-app "back" affordance. Behavioral parity with the `/d` `StandaloneDocPage` full-window UX was not runtime-verified.
- One advisory reviewer leg (line-level adversarial pass) was unavailable this run due to an upstream gateway error and is recorded as **absent**, not as tacit approval.

## Verdict

No blocking issues. Approving with the P2 advisories above as follow-ups.

Jerry-Xin (dismissed):
This PR is relevant to octo-web, but the in-app card routing drops the `sp` space context and can resolve the summary in the wrong space.

🔴 Blocking

🔴 Critical — `sp` is discarded for internal `/s/:taskNo` card clicks  
`packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts:20-27` extracts only the pathname task number and calls `WKApp.openSummaryDetail(summaryTaskNo)`, ignoring `?sp=space_id`. The standalone browser route handles this via `applyStandaloneSummarySpaceFromQuery()` in `apps/web/src/Layout/index.tsx:95-99` and `:449`, but the internal card path bypasses that route entirely. For multi-space users, if `WKApp.shared.currentSpaceId` differs from the card’s `sp`, `getSummaryDetail()` sends the wrong `X-Space-Id`, causing 404/permission failure or potentially opening the wrong task if task numbers are not globally unique. Preserve and apply the `sp` value before routing, or extend `openSummaryDetail` to accept and set the target space.

💬 Non-blocking

🟡 Warning — `getSummaryDetail` now accepts strings but does not encode path segments  
`packages/dmworksummary/src/api/summaryApi.ts:146-147` interpolates `taskId` directly into the URL. The new route regex currently restricts task numbers to `[A-Za-z0-9_-]+`, so this is not immediately exploitable through this PR path, but the widened API type is more general than its callers’ validation. Consider `encodeURIComponent(String(taskId))` for defensive consistency.

✅ Highlights

The standalone return-target guard remains path-scoped and same-origin safe in `packages/docs/src/pages/StandaloneDocPage.tsx:91-107`.

The detail page correctly separates the raw lookup id from the numeric `detail.task_id`, and the BY_PERSON secondary-load race is covered by a focused regression test.

Jerry-Xin (dismissed):
Re-anchoring my REQUEST_CHANGES at the current head 1460ab9b544c. The head advanced since my prior review, but I byte-verified the one blocking defect is still present.

## 🔴 Blocking (still unfixed at this head) — in-app summary card click drops the `sp` space context

`packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts` — `openUrl` now intercepts `/s/:taskNo` card clicks and routes them to `WKApp.openSummaryDetail(summaryTaskNo)`, but it extracts **only** the `taskNo` from `new URL(url).pathname` and **discards the `?sp=space_id` query param**. The standalone browser route applies the space via `applyStandaloneSummarySpaceFromQuery()` in `apps/web/src/Layout/index.tsx`, but the in-app card click bypasses that route, and `openSummaryDetail(taskNo)` takes no space argument.

Impact: for a cross-space forwarded summary card (or when `currentSpaceId` is stale), the detail fetch sends the wrong `X-Space-Id`, the backend cross-space guard returns 404, and the user cannot open a summary they are entitled to. This is fail-closed (not a leak), but it breaks the PR's own goal for the in-app path and diverges from both the standalone route added in this same PR and the documented docs `/d/:docId` `?sp=` handling.

Fix: parse and apply the `sp` query param before routing (set the target space), or extend `WKApp.openSummaryDetail` to accept + set the target space, mirroring `applyStandaloneSummarySpaceFromQuery()`.

## ✅ What is solid (byte-verified)
The standalone browser route `/s/:taskNo` itself is clean: auth required (renders only with `WKApp.loginInfo.token` after session recovery; anonymous → persist-return + login, no pre-auth render), server-enforced authz + space scoping via the `summaryAxios` interceptor injecting token + `X-Space-Id` (401 → logout), taskNo constrained to `[A-Za-z0-9_-]+` and authz-scoped server-side (no cross-space leak — fail-closed 404/403), open-redirect guard intact (single-decode + same-origin + standalone-target-only), no `dangerouslySetInnerHTML`/`innerHTML` (summary content React-escaped), and graceful loading/error/not-found states. The race-guard rewrite (seq-based stale detection preserved, taskId check secondary + null-safe) is a no-regression improvement.

## Non-blocking
`getSummaryDetail` interpolates `taskId` directly into the URL (`packages/dmworksummary/src/api/summaryApi.ts`); the route regex restricts it so it's not exploitable here, but `encodeURIComponent(String(taskId))` would be defensively consistent.

Fix the `sp` drop on the in-app card path and this is an APPROVE — the deep-link architecture and its access control are otherwise sound.

packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts
```diff
@@ -11,6 +14,20 @@ import { isSafeUrl } from "../../../Utils/security";
...
 export function openUrl(url: string): void {
   if (!isSafeUrl(url)) return;
+
+  // isSafeUrl 已保证 http/https 绝对 URL（new URL 不会抛）；try 兼容未来 isSafeUrl 放宽相对路径的情况。
+  let summaryTaskNo: string | null = null;
+  try {
+    summaryTaskNo = new URL(url).pathname.match(SUMMARY_DETAIL_PATH_RE)?.[1] ?? null;
+  } catch {
+    summaryTaskNo = null;
+  }
```
OctoBoooot, line 28: **[major]** In-app summary card click drops the `?sp=` space context → cross-space 404. `openUrl` reads only `new URL(url).pathname` to extract `summaryTaskNo` and calls `WKApp.openSummaryDetail(summaryTaskNo)` — the `?sp=<space_id>` query from the card link `/s/{task_no}?sp={space_id}` is discarded, and `openSummaryDetail`'s signature `(taskId: number | string) => void` has no space param, so the space context can't be recovered. Failing input: a card for a summary in space B, clicked while in space A → detail resolves under space A → not found → fail-closed 404. The standalone `/s/:taskNo` route DOES forward `sp`; this in-app path doesn't. Fix: parse `?sp=` and thread it through `openSummaryDetail(taskNo, sp)` (widen the signature), mirroring the standalone route's sp forwarding.

OctoBoooot (changes requested):
## Re-review (DELTA) @ `1460ab9b544c` — adopting the in-app card `sp`-drop 🔴 I missed in my earlier APPROVE

The head advanced twice past my earlier APPROVE (`a769089aecb1`, now auto-dismissed by GitHub). Re-reviewing at the current head, and I owe a correction: my APPROVE verified the *security* axes (open-redirect guard, origin-agnostic card interception) but **missed a functional bug on the same code** that Jerry-Xin and yujiawei both caught. It's real — byte-confirmed. Flipping to REQUEST_CHANGES.

### Major — in-app summary card click drops the `?sp=` space context → cross-space 404
`packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts:24` — `openUrl` extracts only the path segment and calls `WKApp.openSummaryDetail(summaryTaskNo)`:
```js
summaryTaskNo = new URL(url).pathname.match(SUMMARY_DETAIL_PATH_RE)?.[1] ?? null;
if (summaryTaskNo && WKApp.openSummaryDetail) { WKApp.openSummaryDetail(summaryTaskNo); return; }
```
The card deep-link is `/s/{task_no}?sp={space_id}` (per the PR body), but only `pathname` is read — the **`?sp=<space_id>` query is discarded**. And `openSummaryDetail`'s signature is `(taskId: number | string) => void` (App.tsx) — it takes **no space parameter**, so even the receiver can't recover the space context. Failing input: a completed-summary card for a summary in space B, clicked while the user is in space A → `openSummaryDetail(taskNo)` resolves the detail under the *current* space (A) → the summary isn't found in A → fail-closed 404. This defeats the PR's own stated goal ("mirrors `/d/:docId` incl. `sp` space forwarding") for the in-app path specifically — the standalone `/s/:taskNo` route *does* forward `sp`, but the in-app card interception doesn't. Fix: parse `?sp=` from the URL and thread it through `openSummaryDetail(taskNo, sp)` (widen the signature to accept the space, mirroring how the standalone route forwards it).

### Minor
- `useSummaryDetail` / related `taskId` type widened to `number | string`, but yujiawei notes the string branch has no live consumer yet on some of the widened surfaces — worth confirming each widening is actually exercised (the string `task_no` path is used by the standalone route, so most are; drop any that isn't to avoid dead-type drift).

### Still verified sound (unchanged from my earlier read)
- **Open-redirect guard** (`isSafeReturnPath`) — the `/s/:taskNo` allowlist widening sits after the intact origin/scheme structural checks; no bypass.
- **Card interception security** — the origin-agnostic `/s/` match is a safe *downgrade* (a hostile `https://evil/s/x` becomes an internal `openSummaryDetail`, never opens the attacker origin). That analysis holds; the `sp`-drop is a *functional* defect on the same code, orthogonal to the security question.
- **String-task_no setState race** — correctly handled via `detailLookupId` + explicit numeric `detail.task_id`, pinned by a test.
- Code CI green (Build / secret-scan / osv-scan / dependency-review / pr-title-lint).

### The self-correction I own
My APPROVE checked "is the card interception *safe*?" (yes) but not "does it *preserve* the space context the card carries?" (no). Two different questions on the same lines; I answered one and assumed the other. Credit Jerry-Xin and yujiawei — the `sp`-drop is a real cross-space break, and my APPROVE should not have shipped without checking that `openSummaryDetail` receives the space.

### Net
REQUEST_CHANGES @`1460ab9b544c`: the in-app card path drops `?sp=` so cross-space summary cards 404 — thread the space through `openSummaryDetail`. The security axes I verified remain clean; this is the functional gap I missed. Supersedes my auto-dismissed APPROVE.

yujiawei (changes requested):
# Code Review — PR #790 (octo-web)

Reviewed at head SHA `1460ab9b544cf99b645287f599ff3263fc6a4604` against `main`. This PR adds a `/s/:taskNo` standalone deep-link route to the summary detail page, mirroring the existing `/d/:docId` doc route so completed-summary notification cards can land on the detail view.

## 1. Spec compliance

The diff implements what the description promises and nothing more:

- New `/s/:taskNo` route in `apps/web/src/Layout/index.tsx` with strict single-segment matching, auth/session recovery, and anonymous stash-and-return, mirroring the `/d/:docId` branch.
- Card interception in `packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts` routing `/s/<taskNo>` clicks internally.
- String `task_no` support threaded through `summaryApi.getSummaryDetail`, `useSummaryDetail`, `module.tsx`, `App.tsx`, and `SummaryDetailPage`.
- Open-redirect return-path guard in `packages/docs/src/pages/StandaloneDocPage.tsx` widened to accept `/s/:taskNo`.
- Tests for route parsing, card-action routing, the string-`task_no` load race, and the return-path guard.

No over-building or scope creep observed. **Spec: OK.**

## 2. Code quality

Most of the change is well done. Two paths deserve specific credit:

- **String-`taskNo` vs numeric-`task_id` split** (`SummaryDetailPage.tsx`): `detailLookupId` (raw prop) drives the initial fetch and `componentDidUpdate` change-detection, while `taskId` derives the numeric id from `detail.task_id` once loaded. The BY_PERSON secondary loads (`loadPersonalResult`/`loadMembers`) take the explicit numeric `detail.task_id` to avoid the documented setState race, and the seq-guard (`scheduleLoadSeq !== seq`) still rejects stale responses. This is pinned by a focused regression test.
- **Open-redirect guard** (`StandaloneDocPage.tsx:105-107`): the `/s/:taskNo` addition is an allowlist widening layered *after* the existing structural rejections (parse-failure → false, non-http scheme → false, `url.origin !== origin` → false), so a tampered sessionStorage value still can't bounce to an absolute/scheme-relative/backslash-smuggled target. The regex reuses the same tight single-segment charset as `/d/:docId`. Good defense-in-depth.

### 🔴 P1 (blocking) — in-app card clicks discard the card's `sp` space context, resolving the summary in the wrong space

`packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts:25-28`

```ts
if (summaryTaskNo && WKApp.openSummaryDetail) {
    WKApp.openSummaryDetail(summaryTaskNo);   // only the path segment; ?sp=<space> is dropped
    return;
}
```

The notification card URL is `/s/{task_no}?sp={space_id}` (per the PR description and octo-smart-summary #153), so the card **deliberately carries a space id that can differ from the user's currently active space**. Two consequences make dropping it a functional bug, not a cosmetic one:

1. **The detail fetch is space-scoped.** `packages/dmworksummary/src/api/summaryApi.ts:57-59` unconditionally attaches `X-Space-Id: WKApp.shared.currentSpaceId` to every request. `openSummaryDetail` (`packages/dmworksummary/src/module.tsx:59-64`) never touches `currentSpaceId`, so the in-app path fetches `/summaries/{task_no}` against whatever space happens to be active — not the card's `sp`.

2. **The external route already fixes exactly this**, which is what makes the two entry points asymmetric for the *same* URL. `apps/web/src/Layout/index.tsx:448` calls `applyStandaloneSummarySpaceFromQuery()` (reads `sp`/`space`, sets `WKApp.shared.currentSpaceId`) *before* rendering `SummaryDetailPage`. The in-app card path bypasses the route entirely and has no equivalent step.

**Impact:** For a multi-space user whose active space differs from the card's `sp`, the in-app click sends the wrong `X-Space-Id` → 404 / permission error, or — if `task_no` is not globally unique across spaces — silently opens the wrong task. The happy path (card's space == active space) works, which is why unit tests pass, but the cross-space case is the whole reason `sp` is on the URL.

This is the same issue a prior reviewer already flagged as blocking; the only change since (`actions.ts` try/catch refactor) did not address it, and `packages/dmworkbase/src/Messages/InteractiveCard/__tests__/actions.test.ts:28-31` now *codifies* the drop (asserts `openSummaryDetail` is called with the task_no only, given an input URL that carries `?sp=space-1`).

**Fix direction:** parse `sp`/`space` from the card URL and apply it before navigating — either set `WKApp.shared.currentSpaceId` in `openUrl` prior to calling `openSummaryDetail`, or extend `openSummaryDetail(taskNo, spaceId?)` to accept and set the target space (and update the test to assert the space is forwarded).

### 🟡 P2 (non-blocking)

- **Login-redirect forwards `sp` but not `space` for `/s` paths.** `apps/web/src/Layout/index.tsx:154-156` only re-appends `forwardSp` (`sp`) to the post-login redirect, while `applyStandaloneSummarySpaceFromQuery` accepts both `sp` and `space`. An anonymous visitor arriving at `/s/<taskNo>?space=<id>` keeps the space via the stashed return target (`persistStandaloneReturn` saves `pathname + search`), but not via the explicit `forwardSp` redirect param. Align the two by also forwarding `space` for symmetry.
- **`getSummaryDetail` interpolates `taskId` into the path without encoding** (`packages/dmworksummary/src/api/summaryApi.ts:146-147`). The route regex restricts `task_no` to `[A-Za-z0-9_-]+`, so it is not exploitable via this PR's paths, but the widened `number | string` type is more general than the callers' validation. `encodeURIComponent(String(taskId))` would be defensive consistency.
- **`useSummaryDetail` widening is currently dead code.** The hook accepts `number | string | null` and `regenerate`/`cancel` correctly guard `typeof taskId !== "number"`, but the hook has no consumers in the tree — speculative type-widening. Correct as written; noting that a future string caller would make regenerate/cancel silently no-op by design.

## 3. Coverage / blind spots

- End-to-end deep-link resolution depends on octo-smart-summary #153 (worker sends the card) and #154 (server resolves detail by `task_no`), plus the nginx `try_files` SPA fallback for `/s/*` — none verifiable from this frontend diff.
- `SummaryDetailPage` renders outside the app shell on the `/s` branch (no nav rail / in-app back affordance). Context providers (`I18nProvider`) wrap at the app root and routing singletons are static, so mounting is safe; behavioral parity with the `/d` `StandaloneDocPage` full-window UX was not runtime-verified.
- The P1 above is a cross-file dependency (card URL → `openSummaryDetail` → space-scoped API interceptor) that unit tests do not exercise, since no test sets a `currentSpaceId` differing from the card's `sp`.

## 4. Verdict

`Spec: OK` but `Quality: Changes-Requested` due to the P1 cross-space `sp`-drop on the in-app card path. Overall: **CHANGES_REQUESTED**. The route, string-`task_no` race handling, and open-redirect guard are solid; the blocker is that the internal card entry point does not honor the space the card carries, while the external route does — forward `sp` (and `space`) into the internal navigation to make the two paths symmetric, then this is ready.

OctoBoooot (approved):
## Re-review (DELTA) @ `470f290403f9` — the in-app card `sp`-drop is fixed → flipping REQUEST_CHANGES to APPROVE

The head advanced past my CR's `1460ab9b544c`. The `sp`-drop major I flagged (converged on by Jerry-Xin and yujiawei) is fixed correctly and end-to-end, byte-verified.

### Major (in-app card `sp`-drop → cross-space 404) — FIXED, end-to-end
- **Parse + forward** (`actions.ts:25,31`): `openUrl` now reads `summarySpace = u.searchParams.get("sp") || u.searchParams.get("space")` and calls `WKApp.openSummaryDetail(summaryTaskNo, summarySpace)` — the card's `?sp=` is threaded through instead of discarded.
- **Signature widened** (`App.tsx`): `openSummaryDetail?: (taskId: number | string, spaceId?: string)`.
- **Receiver actually applies it** (`module.tsx:6-8`): `WKApp.openSummaryDetail = (taskId, spaceId?) => { if (spaceId) WKApp.shared.currentSpaceId = spaceId; … }` — the space context is set *before* the detail resolves, so the fetch uses the correct `X-Space-Id`. This is the part that matters: not just a widened type, but the space is consumed. A cross-space card click (summary in space B, clicked from space A) now resolves under space B → no 404.
- **Test locks it** (`actions.test.ts`): `openUrl(".../s/ST…?sp=space-1")` asserts `openSummaryDetail` called with `("ST…", "space-1")`, plus a no-`sp`→`undefined` case.

### No regression
The security axes remain clean (verified earlier and unchanged): the open-redirect guard's `/s/:taskNo` allowlist widening still sits behind the intact origin/scheme checks; the card interception's origin-agnostic `/s/` match is still a safe downgrade (hostile origin → internal nav, never opens the attacker page); the string-task_no setState race handling is intact. Code CI green (Build / secret-scan / osv-scan / dependency-review / pr-title-lint).

### Net
The `sp`-drop is fixed at the source (`openUrl` parses+forwards), the receiver (`module.tsx`) applies the space to `currentSpaceId` before resolving, and a regression test covers both the with-`sp` and without-`sp` cases. Flipping **REQUEST_CHANGES → APPROVE @`470f290403f9`**, superseding my CR stranded at the old head. `needs-human-review` may apply per the size label; otherwise code is ready. Credit @Ranwanglc — the fix threads the space cleanly through parse → handler → space-context, exactly as needed, and the earlier deep-link security design remains sound.

## Parity checked
Diffed my CR head (`1460ab9b544c`) → new head (`470f290403f9`): the change adds `sp`/`space` parsing in `openUrl` and a second `spaceId?` param on `openSummaryDetail` (applied to `currentSpaceId` in `module.tsx`) — a strict addition. The standalone `/s/:taskNo` route, open-redirect guard, and origin-agnostic card interception are unchanged. No capability removed; the one behavioral delta is the intended space-context forwarding that fixes the cross-space 404.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #790 (octo-web)

## Summary

This PR adds a `/s/:taskNo` standalone deep-link route for summary detail pages, mirroring the existing `/d/:docId` pattern used for document deep-links. The implementation spans three surfaces: browser URL routing (Layout cold-load, anonymous → login → return bounce), in-app card-link interception (InteractiveCard `openUrl`), and the summary detail page itself (widened `taskId` to accept string `taskNo` alongside numeric `task_id`). Auth, session recovery, open-redirect safety, and race-condition guards are all properly extended. Test coverage is thorough with dedicated test suites for each entry path.

## Verification

- ✅ **Auth parity with `/d/:docId`** — The `/s/:taskNo` branch in `apps/web/src/Layout/index.tsx:441–458` mirrors the doc deep-link: `loginInfo.load()` → `recoverOctoSessionFromStorage(true)` → token gate before rendering → `persistStandaloneReturn()` for anonymous users.
- ✅ **Open-redirect guard extended** — `isSafeReturnPath` in `packages/docs/src/pages/StandaloneDocPage.tsx:108` now accepts `/s/:taskNo` alongside `/d/:docId` via the same `STANDALONE_SUMMARY_PATH` regex, keeping the three-gate (control-char, same-origin, standalone-target-only) structure intact.
- ✅ **Data flow: string → numeric** — `getSummaryDetail` accepts the string `taskNo`; subsequent `loadPersonalResult` and `loadMembers` calls receive `detail.task_id` (numeric), bypassing the setState race window. Verified at `packages/dmworksummary/src/pages/SummaryDetailPage.tsx:482–484`.
- ✅ **Race-condition guards preserved** — The stale-response guard relaxation (`this.taskId != null && this.taskId !== requestTaskId` at `:538, :556`) is safe: the `scheduleLoadSeq` mechanism remains the primary guard, and the relaxed `taskId` check only allows first-load responses through when state hasn't committed yet.
- ✅ **`componentDidUpdate` change detection** — Now uses `detailLookupId` (raw prop) instead of `taskId` (derived numeric), correctly detecting prop-level task switches (`:228`).
- ✅ **InteractiveCard interception** — `openUrl` at `packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts:17–34` correctly routes `/s/<taskNo>` URLs through `WKApp.openSummaryDetail` with space param, falling back to `window.open` when the callback is unavailable.
- ✅ **`encodeURIComponent` applied** — `getSummaryDetail` at `packages/dmworksummary/src/api/summaryApi.ts:148` wraps the identifier in `encodeURIComponent(String(taskId))`, though the `[A-Za-z0-9_-]+` regex ensures only URL-safe characters reach this point.
- ✅ **Test coverage** — Four-entry-point tests for string `taskNo` deep-link (`SummaryDetailPage.schedule.test.tsx:2694–2794`), layout structural tests (`layoutStandaloneDocPath.test.ts:74–124`), card interception tests (`actions.test.ts:1–73`), and `withReturnSid` / `consumeStandaloneReturn` tests (`StandaloneDocPage.test.tsx:803–835`).

Static analysis only at head `470f290403f9f82176dbd6206b6dd4b77d1c18af`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 maintenance item below.

### P2 — Duplicated summary path regex across three files (`apps/web/src/Layout/index.tsx:79`)

`STANDALONE_SUMMARY_PATH` is defined independently in `apps/web/src/Layout/index.tsx:79`, `packages/docs/src/pages/StandaloneDocPage.tsx:28`, and as `SUMMARY_DETAIL_PATH_RE` in `packages/dmworkbase/src/Messages/InteractiveCard/renderer/actions.ts:4`. All three are identical today (`/^\/s\/([A-Za-z0-9_-]+)\/?$/`), but if the taskNo format ever changes, all three must be updated in lockstep. The `/d/:docId` equivalent already lives in `@octo/docs` and is imported by the Layout — consider exporting the summary regex (and `parseStandaloneSummaryTaskNo`) from the same package so the Layout and InteractiveCard paths consume the same source of truth.

## Data-Flow Trace

- **Browser URL `/s/TN123?sp=space1`** → Layout detects `isStandaloneSummaryPath` → token check → `applyStandaloneSummarySpaceFromQuery()` sets `currentSpaceId` → `parseStandaloneSummaryTaskNo` extracts `"TN123"` → renders `<SummaryDetailPage taskId="TN123" />`.
- **SummaryDetailPage with string taskId** → `detailLookupId` returns `"TN123"` (raw prop) → `loadDetail` calls `getSummaryDetail("TN123")` → response has numeric `task_id: 740` → `loadPersonalResult(seq, false, 740)` and `loadMembers(seq, 740)` pass numeric ID → `setState` commits detail → `this.taskId` now returns `740` for all subsequent operations.
- **InteractiveCard URL interception** → `openUrl("https://host/s/TN123?sp=space1")` → `isSafeUrl` passes → `new URL` parses → regex matches `/s/TN123` → `WKApp.openSummaryDetail("TN123", "space1")` → module sets `currentSpaceId`, switches to summary menu, renders detail page.
- **Anonymous → login → return** → `persistStandaloneReturn()` stashes `/s/TN123?sp=space1` → login screen → `onLogin` fires → `consumeStandaloneReturn()` returns stashed URL → `withReturnSid` appends sid → `window.location.assign` navigates back → Layout re-renders with token → detail page loads.

## Blind Spot Checklist

- **C1 — Dual-path parity**: `/s/:taskNo` routing in Layout mirrors `/d/:docId` (session recovery, token gate, anonymous stash, login return). `isSafeReturnPath` accepts both. `withReturnSid` works with both. ✅ Parity verified.
- **C2 — Control-flow ordering**: Summary branch in Layout is after doc branch; paths are mutually exclusive (`/d/...` vs `/s/...`). In `openUrl`, summary check is after `isSafeUrl` and before `window.open`. No ordering conflict. ✅
- **C3 — Authorization boundary**: No new endpoints exposed. Summary detail requires authentication (token check before render). API calls use shared client with auth headers. ✅
- **C4 — Authorization lifecycle**: N/A — no auth mechanism changes.
- **C5 — Build/runtime parity**: N/A — no build/config changes.
- **C6 — Docs/strategy**: N/A — no governance changes.

N/A — first review round.

- All existing callers of `SummaryDetailPage` (`ChatSummaryPanel`, `SummaryListPage`, `SummaryConfirmPage`, `SummaryCreatePage`, route `/summary/detail`) still pass numeric `taskId` values — no unintended string propagation.
- `module.tsx`'s `openSummaryDetail` callback sets `currentSpaceId` before routing, matching the browser route's `applyStandaloneSummarySpaceFromQuery` behavior.
- Login redirect fallback correctly carries `sp`/`space` params for summary paths when `consumeStandaloneReturn()` returns null (sessionStorage lost edge case).
- The `?? undefined` fallback at `Layout/index.tsx:453` is unreachable in practice (the regex already guarantees a non-null match), but is a safe defensive pattern.
- `useSummaryDetail` hook's `regenerate`/`cancel` now gate on `typeof taskId !== "number"` — correct since backend `regenerateSummary`/`cancelSummary` only accept numeric IDs.

## Verdict: APPROVED

Well-structured deep-link implementation that closely follows the established `/d/:docId` pattern. Auth, session recovery, open-redirect safety, and async race-condition guards are all properly extended. The only concern is regex duplication across three files, which is a non-blocking maintenance item.

[Octo-Q] verdict: APPROVE — zero P0/P1; one P2 (regex duplication) is non-blocking.

yujiawei (approved):
# Code Review — PR #790 (octo-web)

Reviewed at head SHA `470f290403f9f82176dbd6206b6dd4b77d1c18af` against `main` (merge-base `8ae5db6d`). Independent re-review of the full 13-file diff, plus local build + targeted tests, plus a second-opinion advisory pass.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly what the PR describes — a `/s/:taskNo` standalone deep-link route that mirrors the existing `/d/:docId` pattern, plus the plumbing to let `SummaryDetailPage` accept a **string `task_no`** in addition to the numeric `task_id`.

- **Route** (`apps/web/src/Layout/index.tsx`): new `/s/:taskNo` branch with a strict single-segment regex (`parseStandaloneSummaryTaskNo`); `/s`, `/s/`, and nested `/s/a/b` correctly fall through instead of mounting the page with an undefined id. Auth/session recovery (`load()` → `recoverOctoSessionFromStorage`) and the anonymous `persistStandaloneReturn()` stash mirror the `/d` branch faithfully. Login-redirect forwards both `sp` and `space`, symmetric with `applyStandaloneSummarySpaceFromQuery`.
- **Card interception** (`InteractiveCard/renderer/actions.ts`): `/s/<taskNo>` clicks route internally via `WKApp.openSummaryDetail(taskNo, space)`, forwarding the deep-link's `sp`/`space` so the detail resolves in the correct space. Falls back to `window.open` when the hook is unavailable.
- **Open-redirect guard** (`StandaloneDocPage.tsx`): the post-login safe-return allowlist is extended to accept `/s/:taskNo` alongside `/d/:docId`, keeping the origin + single-segment structural checks intact.
- **String/numeric split** (`SummaryDetailPage.tsx`): `detailLookupId` (raw string|number) drives the initial fetch and change-detection; the numeric `this.taskId` getter derives from `detail.task_id` once loaded. BY_PERSON secondary loads receive the explicit numeric `detail.task_id`, avoiding the setState-race trap the PR calls out.
- Type widening across `summaryApi.getSummaryDetail`, `useSummaryDetail`, `module.tsx`, `App.tsx` is consistent; `getSummaryDetail` now `encodeURIComponent`s the id (good hygiene for string task_no).

No missing scope, no out-of-scope additions, no contract deviations. No new required env vars (end-to-end effect depends on the already-planned smart-summary server/worker changes, correctly noted as a deploy prerequisite, not a code concern here).

## 2. Code Quality

**Quality: Approved**

Verification performed locally at this head:
- `pnpm turbo run build --filter=@octo/web` → ✓ built clean.
- `dmworksummary` suite → **381 passed** (incl. the new string-task_no race tests).
- `apps/web` `layoutStandaloneDocPath` → 9 passed; `dmworkbase` `InteractiveCard/actions` → 5 passed; `docs` `StandaloneDocPage` → 43 passed.
- (The full `apps/web` suite shows ~34 failures in unrelated files — voice input, password-strength, invite-landing, ErrorBoundary, and Playwright e2e specs mis-collected by vitest. None touch this PR's files; pre-existing/environmental.)

**Async task-switch isolation is sound.** The guards in `loadPersonalResult` / `loadMembers` were loosened from `this.taskId !== requestTaskId` to `this.taskId != null && this.taskId !== requestTaskId`. This is safe because the monotonic `scheduleLoadSeq` (bumped on every `loadDetail`, i.e. every real task switch) is the primary staleness gate; the `taskId` check is now a secondary filter that intentionally tolerates the first-load window where `state.detail` (and thus `this.taskId`) hasn't committed yet for a string deep-link. The four `loadPersonalResult` entry points (loadDetail / handleSubmitPersonal / handleStatusChangeEvent / doFallbackPollOnce) are all exercised by the new tests and consistently drive numeric ids. No cross-task leak of `personalResult`/`members`.

### Non-blocking findings (P2 / nits)

- **P2 — `/s/<path>` interception is not origin-scoped** (`InteractiveCard/renderer/actions.ts:23-31`): `openUrl` matches `u.pathname` against `/s/<seg>` regardless of `u.origin`. A legitimate **external** card link whose path happens to be exactly `/s/<something>` (e.g. `https://third-party.example/s/promo`) would be hijacked into the internal summary view instead of opening in a new tab. This is a deliberate "path is an internal signal" design and carries no data-leak risk (it fails toward an internal 404, not toward exfiltration), but it does degrade a class of external links. Consider gating on `u.origin === window.location.origin` before treating the path as internal. Not a merge blocker.

- **Nit — `applyStandaloneSummarySpaceFromQuery()` mutates `WKApp.shared.currentSpaceId` during `render()`** (`Layout/index.tsx`): flagged by the advisory pass as a potential re-render loop. Verified this is **not** a live risk: `currentSpaceId` is a plain class field (not a MobX/observable-backed store), and the adjacent `/d` branch already assigns `WKApp.shared.currentSpaceId` during the same render phase — this follows the file's established precedent. Left as a note only; long term, side-effecting store writes are cleaner in a lifecycle method.

- **Nit — `useSummaryDetail.regenerate`/`cancel` silently no-op for a string id** (`hooks/useSummaryDetail.ts:44-50`): the `typeof taskId !== "number"` early-return means these mutations would fail silently if the hook were ever driven by a string `taskNo`. This is currently harmless — the hook has **no consumers** in the repo, and `SummaryDetailPage` is a class component that does not use it — so there is no reachable path today. Worth resolving the string→numeric id before enabling mutations if a future view adopts the hook.

## 3. Overall verdict

**APPROVE**

Spec ✅ and Quality Approved. No P0/P1. The one previously-blocking issue across earlier review rounds — the in-app card dropping `sp` space context, which could resolve a summary in the wrong space — is fixed at this head (card + `openSummaryDetail` now forward the space, symmetric with the browser-route path). The remaining findings are advisory hardening (origin-scoping the interception) and dead-code/nit observations.

## 4. Suggestions

- Add an origin check in `openUrl` before treating `/s/<seg>` as an internal deep-link, so external card links with a colliding path shape aren't captured.
- If `useSummaryDetail` gains a consumer that can carry a string `task_no`, resolve it to the numeric `task_id` (from the loaded detail) before allowing `regenerate`/`cancel`, rather than silently no-opping.

## 5. Additional notes

- Advisory second-opinion coverage: one long-context advisory leg completed and its three findings are absorbed above (all downgraded to P2/nit after independent diff verification); the line-level adversarial leg was **absent this run** (upstream gateway error on every attempt — recorded as absent, not treated as silent approval). The verdict rests on my independent review plus the local build/test evidence.
- Not verifiable from the frontend diff alone: that the backend `/summaries/{task_no}` endpoint resolves string task_no safely, and the exact runtime reactivity of `WKApp.shared` beyond the source inspection above — both out of scope for this frontend PR.

Jerry-Xin (approved):
Re-review of the rebased fix head. My prior REQUEST_CHANGES is resolved — the cross-space `sp` drop is now fixed end-to-end, and the encode nit is addressed. Approving.

## Verified fixes

**🔴 (resolved) Cross-space `sp` is now threaded end-to-end.**
Traced the in-app card path fully: `Messages/InteractiveCard/renderer/actions.ts` now parses `u.searchParams.get("sp") || u.searchParams.get("space")` and calls `WKApp.openSummaryDetail(taskNo, space)`. The implementation in `dmworksummary/src/module.tsx` sets `WKApp.shared.currentSpaceId = spaceId` before routing to the detail page. The summary fetch client in `dmworksummary/src/api/summaryApi.ts` has a request interceptor that reads `WKApp.shared.currentSpaceId` and emits it as the `X-Space-Id` header on every request, including `getSummaryDetail`. So a forwarded card carrying a foreign space now resolves against the correct space rather than the current one → no more 404. The undefined-space case is guarded (`if (spaceId)`), so a normal same-space card leaves `currentSpaceId` untouched and keeps working.

**🟡 (resolved) Path param is now encoded.** `getSummaryDetail` now uses `encodeURIComponent(String(taskId))`.

## No new bug
- The other caller of `openSummaryDetail` in `Messages/SummaryCard/index.tsx` passes a single arg; the new second parameter is optional, so it still compiles and correctly falls back to the current space.
- The standalone browser deep-link route stays consistent: `apps/web/src/Layout/index.tsx` applies the same `sp||space → currentSpaceId` at mount (symmetric with the card path), and the login-redirect branch now also preserves `space` alongside `sp`, so the space survives the auth bounce.

## Tests lock the behavior
`InteractiveCard/__tests__/actions.test.ts` asserts a card URL with `?sp=` results in `openSummaryDetail` being called with that space, and a URL without it is called with `undefined` — pinning both the threading and the fallback.

## CI
All code checks green (Build, osv-scan, secret-scan, dependency-review, common-ancestor). Governance checks (check-sprint / satisfy-code-review) are skipping and non-blocking.

✅ Nicely scoped fix with symmetric handling across both entry paths and a defensive encode.

PR #792, opened 2026-07-15
feat(summary): render smart summary output incrementally

### LINKED ISSUE #786: feat(summary): stream summary output to the frontend and render incrementally
## What

Render smart-summary output incrementally while it is being generated, instead of waiting for the final result to be available before showing summary text.

## Why

Summary generation can take a while. Progressive rendering gives users immediate feedback and improves perceived latency.

## Scope

- Consume the backend streaming endpoint and render summary content chunk-by-chunk.
- Show a clear in-progress streaming state.
- Fall back to existing polling if the stream is interrupted or unavailable.
- Refresh the final detail/result after stream completion, using the persisted result as the source of truth.
- Keep citation rendering correct once generation completes.
- Cover initial generation, regenerate, and refine flows where supported by the backend.

Liuxy20:
## Summary
- Consume backend streaming endpoints for smart-summary output.
- Render summary content incrementally while generation is in progress.
- Support streaming display for personal summary, team summary, regenerate, and refine flows where available.
- Fall back to existing polling when streaming is interrupted or unavailable.
- Refresh final detail/result after stream completion and use the persisted result as the source of truth.
- Keep workflow/progress display order stable while streaming content updates.
- Hide the forward-to-matter action on summary detail behind a local feature flag.
- Update nginx summary API proxy settings for streaming responses.

## Implementation
- Add summary stream API helpers.
- Add incremental rendering state and stream handling in the summary detail page.
- Keep polling as fallback and final result refresh path.
- Configure nginx buffering/timeout settings for summary streaming.

## Testing

Closes #786


packages/dmworksummary/src/pages/SummaryDetailPage.tsx
```diff
@@ -721,33 +854,490 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
...
+                    if (!baseResultId) return;
+                    const previousPersonalResult = this.state.personalResult;
+                    const previousMembers = this.state.members;
+                    restoreRefineDraft = () => {
+                        if (this.taskId === requestTaskId) this.setState({ personalResult: previousPersonalResult, members: previousMembers });
+                    };
+                    this.setState({ showRegenerateModal: false, refineLoadingTarget: "personal" });
+                    let draft = "";
+                    let refined: api.SummaryStreamEvent | null = null;
+                    await api.streamRefinePersonalSummary(requestTaskId, {
```
OctoBoooot, line 1069: **[major]** The refine/regenerate streams are un-cancellable → setState-after-unmount + a network stream that outlives the page. The live streams create an AbortController (:862/:940), pass `signal` (:868/:946), and abort() on unmount (:920-925). `streamRefinePersonalSummary`/`streamRefineSummary` also ACCEPT `signal?: AbortSignal` (summaryApi.ts:289/:343) — but these call sites pass the options object WITHOUT `signal`. So componentWillUnmount can't cancel a refine stream; the `onEvent` guard `this.taskId !== requestTaskId` reads `this.props.taskId` and doesn't change on unmount, so the post-unmount setState (and the trailing `finally` setState({regenerateSubmitting:false})) still runs. Fix: create + pass a refine AbortController like the live-stream path and abort it on unmount / taskId change.

packages/dmworksummary/src/pages/SummaryDetailPage.tsx
```diff
@@ -721,33 +854,490 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
...
+                        feedback: trimmed,
+                        base_result_id: baseResultId,
+                        base_version: this.state.personalResult?.version,
+                    }, {
+                        onEvent: (event) => {
+                            if (this.taskId !== requestTaskId) return;
+                            if (event.type === "delta") {
+                                draft = event.content || (draft + (event.delta || ""));
```
OctoBoooot, line 1077: **[question]** `draft = event.content || (draft + (event.delta || ""))` (also :1139) collapses snapshot vs delta into the single `type==="delta"` case — but the live handlers (:880/:953) treat `snapshot` (full replace) and `delta` (increment) as distinct event types. If the backend ever emits a full-`content` snapshot inside a `delta`-typed event and then resumes incremental deltas (a common SSE catch-up pattern), this can duplicate/drop text, and the refine path commits `draft` optimistically (no authoritative result re-fetch), so the corruption is visible until navigating away. On the happy path `content: refined!.content || draft` masks it. Please confirm: can a `delta`-typed refine event carry a full `content` snapshot? If yes, mirror the live handlers' explicit snapshot/delta split here.

OctoBoooot (changes requested):
## Review: render smart summary output incrementally (#792)

**Verdict:** Request changes — a real lifecycle bug (refine streams are un-cancellable), a merge conflict with base, and one protocol-dependent correctness question. The security core (streamed content rendering, SSE parsing) is verified clean, so the streaming design is sound; these are fixable edges.

Merge state: `mergeStateStatus=DIRTY` — the branch has a **merge conflict with `main`**; rebase/resolve before this can land. Risk: `size/XL` → `needs-human-review` applied. CI: `Build` hasn't registered at this head (only governance/meta checks present), so code CI is unconfirmed — a second reason APPROVE isn't available yet.

### Major
- `packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1069, :1135` — **the refine/regenerate streams are un-cancellable, so they can `setState` after unmount and keep a network stream alive after the page is gone.** The live streams correctly create an `AbortController` (`:862`, `:940`), pass `signal` (`:868`, `:946`), and `abort()` on unmount (`:920-925`). The refine helpers *also accept* `signal?: AbortSignal` (`summaryApi.ts:289, :343`) — but the call sites pass the options object **without `signal`**. So `componentWillUnmount` can't cancel a refine stream; its `onEvent` guard is `this.taskId !== requestTaskId`, which reads `this.props.taskId` and does **not** change on unmount, so it doesn't stop the post-unmount `setState` (and the trailing `finally` `setState({regenerateSubmitting:false})` runs on an unmounted/next-task component). Fix: create and pass a refine `AbortController` exactly like the live-stream path, and abort it in `componentWillUnmount`/on taskId change.

### Question (major if the backend does this)
- `SummaryDetailPage.tsx:1077, :1139` — the refine draft accumulator is `draft = event.content || (draft + (event.delta || ""))`, handled only under `type === "delta"`. The **live** stream handlers (`:880`, `:953`) treat `snapshot` (full replace) and `delta` (increment) as *distinct event types*; the refine handler collapses them into one `delta` case. If the backend ever emits a periodic full-`content` snapshot inside a `delta`-typed event and then resumes incremental `delta`s (a common SSE catch-up pattern), the accumulation can duplicate or drop text — and since the refine path commits `draft` optimistically (no authoritative `loadDetail` re-fetch of the result, only version-list refresh), a corrupted draft is what the user sees until they navigate away. On the happy path `content: refined!.content || draft` (`:1104` etc.) uses the `done` event's authoritative content, which masks it. Please confirm the refine stream's event contract: does a `delta`-typed event ever carry a full `content` snapshot? If yes, mirror the live handlers' explicit snapshot-vs-delta split here. (I can't construct the failing input without knowing the backend protocol — hence a question, not a blocker.)

### Minor
- `SummaryDetailPage.tsx:819` — `doFallbackPollOnce`'s `setState({ detail })` is guarded by `requestTaskId` (correct for task-switch) but lacks the `scheduleLoadSeq` generation guard that `loadDetail` uses. A poll tick already in flight when a stream-triggered `loadDetail` completes can land after it and briefly overwrite the completed `detail` with a still-`PROCESSING` one. Self-heals on the next fetch (both converge to completed), so transient — but a shared seq guard would remove the flicker.

### Verified clean (the security-critical axes)
- **XSS / streamed content rendering** — `CitationText.tsx` renders through `ReactMarkdown` + `rehypeSanitize` with an explicit schema (`defaultSchema` + only the custom `citation`/`citationgroup`/`teamcitation` tags carrying numeric/string `index`/`badgekey`/`indices`, rendered by React components). LLM-generated summary markdown is sanitized to the safe allowlist; no raw-HTML/script injection. Links get `rel="noopener noreferrer"`.
- **SSE parsing** — `parseSSEBlock` guards `JSON.parse` in try/catch (falls back to a delta, never crashes), handles the `[DONE]` sentinel, and `consumeSSE` buffers partial chunks correctly (`decode(..., {stream:true})`, splits on the blank-line boundary, drains the tail). Robust against malformed events.
- **nginx SSE config** — `proxy_buffering off` / `proxy_cache off` / `proxy_http_version 1.1` / `proxy_read_timeout 3600s` under the existing `/summary/api/v1` proxy is the correct unbuffered-streaming setup; no auth/CORS/header regression.
- **Feature flag** — `SHOW_FORWARD_TO_MATTER = false` is a real render gate (`:3127`), not dead code; reversible as intended.
- **Stream/persisted render** — no double-render: `renderStreamingContent` gates on `state.streaming`, `renderCompleted` on `detail.status === COMPLETED` (mutually exclusive during streaming); stale `streamingContent` is invisible after `done` and always cleared before the next stream. The per-task/per-seq guards on the loaders are otherwise thorough.

### Praise
- Wiring `AbortController` + `signal` through the live personal/team streams and aborting on unmount is exactly right — which is what makes the refine path's omission stand out as an oversight rather than a design choice. Same treatment there and it's consistent.
- The sanitize-schema approach in `CitationText` (allowlist the custom citation tags with only the attributes they need, everything else through `defaultSchema`) is the correct way to render untrusted LLM markdown with custom interactive nodes.

### Net
Sound incremental-render design with a verified-safe render/parse/proxy core, but: the refine streams need the same `AbortController` cancellation the live streams already have (major), the branch has a merge conflict to resolve (DIRTY), and the refine draft accumulation needs the snapshot/delta protocol confirmed (question). Fix the refine cancellation + rebase + confirm the stream contract, let Build run green, and this is close. `needs-human-review` applied for the XL size. Credit @Liuxy20 — the streaming architecture (abortable live streams, sanitized render, polling fallback, persisted-result-as-truth) is well thought out.

Jerry-Xin (commented):
Reviewed the incremental smart-summary streaming pipeline end-to-end (SSE parse → delta accumulation → sanitized markdown render → reconcile). Code quality is high and I found no blocking security/correctness issues in the streaming path. Holding at COMMENT rather than APPROVE only because the branch has merge conflicts and no code CI (Build/Test/Vet/Lint/secret-scan) is green on this head — see gate notes below.

## ✅ Highlights
- **XSS on incremental render is safe.** Streamed AI text flows through `CitationText` (`packages/dmworksummary/src/components/CitationText.tsx:271`) which renders via `ReactMarkdown` with `rehypePlugins={[[rehypeSanitize, citationSchema]]}`. `citationSchema` (CitationText.tsx:43) extends rehype-sanitize's `defaultSchema` (which strips scripts, event-handler attributes, and unsafe URI schemes) and only adds the custom `citation`/`teamcitation` tags with `index`/`badgekey` attributes. Because sanitization re-runs over the full accumulated buffer on every render, a partial/incomplete markdown fragment arriving mid-stream cannot produce an injection — the dangerous fragment is stripped each tick.
- **Chunk ordering / race is guarded.** `startSummaryStream` (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:857`) aborts any prior controller via `stopSummaryStream()` before starting, dedups with `streamingTaskId`, and every `onEvent` handler short-circuits with `if (this.taskId !== taskId) return` so a stale stream for a task the user navigated away from cannot corrupt the current view. `streamClosedTaskId` prevents re-opening a finished stream. `delta` events append in order through functional `setState`; `content`/`snapshot` events full-replace; `done` triggers `loadDetail()` to reconcile to the canonical result. A rapid tab-switch mid-stream cannot interleave two streams.
- **Resource cleanup is complete.** `componentWillUnmount` (SummaryDetailPage.tsx:283) aborts both stream AbortControllers and clears the fallback poll and all workflow timers. The fallback poll is bounded (stops on COMPLETED/FAILED/CANCELLED) and captures a per-tick `requestTaskId` to discard late responses after a task switch. No leaked fetch/SSE reader.
- **Auth/scope carried on the stream.** `buildStreamHeaders` in `packages/dmworksummary/src/api/summaryApi.ts` sets the `token` header and `X-Space-Id`, matching the non-streaming path — the SSE endpoint is not an unauthenticated bypass.
- **nginx SSE config is correct.** `nginx.conf.template:89` adds `proxy_buffering off`, `proxy_cache off`, `proxy_http_version 1.1`, and a long read timeout on the `/summary/api/v1` location, which is required for unbuffered streaming.
- **No regression from the -97 deletions.** The removed processing/regenerate-modal render path is replaced by the streaming card; the `done` event reconciles to the canonical `CitationText content={detail.result.content}` render, and error/empty states are preserved (`renderFailed`, `streamError`, and the null-gate on empty streaming content).

## 💬 Non-blocking notes
- The streaming render passes `citations={[]}` (SummaryDetailPage.tsx renderStreamingContent), so citation badges only appear after `done` reconciles the final result. This is a minor cosmetic gap during streaming, not a correctness issue.
- `proxy_read_timeout 3600s` applies to the entire `/summary/api/v1` location rather than just the SSE routes. Acceptable given the API's long LLM calls, but worth being aware of.

## Gate status (why COMMENT, not APPROVE)
- **Merge conflict:** the branch is currently CONFLICTING with `main` (mergeStateStatus DIRTY). Please rebase/resolve.
- **Code CI:** no code checks (Build/Test/Vet/Lint/secret-scan/osv) are green on this head — only governance/label checks report. I won't APPROVE over unconfirmed code CI. Once the conflict is resolved and code checks pass green, this is otherwise ready from a code-review standpoint.

yujiawei (changes requested):
# Code Review — PR #792 (octo-web)

Reviewed at head SHA `46d04d3287f56421ca31a97bcb81a158347d8812` against `origin/main` `53f2f0c7`. Build (`pnpm --filter @octo/web build`) and the summary unit suite (`SummaryDetailPage.schedule.test.tsx`, 131/131) both pass at this SHA in isolation.

The incremental-streaming implementation itself is genuinely good: the SSE reader, `AbortController` lifecycle for generation/team streams, task-switch guards (`streamClosedTaskId`), and polling fallback are careful and thorough. The blocker is not the streaming code — it is that this branch cannot land on the current `main` as-is.

## 1. Spec compliance

Scope for this change (issue #786): consume the backend streaming endpoint, render summary content incrementally, show an in-progress state, fall back to polling when the stream is unavailable, and refresh the persisted result as source of truth after completion. The streaming work delivers all of that.

However, two scope problems block a clean approval:

**(a) The branch re-implements an already-merged feature → unmergeable.** This branch was cut before the sibling PR #698 ("support refining and regenerating summaries") landed on `main`. #698 already added, on `main` today, the entire refine/full-regenerate flow, recent-version history + restore, `generation_instruction` schedule field, and the associated i18n/CSS/tests. This PR re-adds all of that **plus** SSE streaming on top of its own older copy. As a result:

- `git merge origin/main` fails with conflicts in `packages/dmworksummary/src/api/summaryApi.ts` and `packages/dmworksummary/src/pages/SummaryDetailPage.tsx` (48 conflict markers total: 6 in the API, 42 in the page). GitHub reports the PR as `CONFLICTING`.
- Because the refine/version code here is an *older* fork of what #698 shipped, a naive conflict resolution risks silently reverting fixes that landed in #698.

This needs a rebase onto current `main` so the diff contains **only** the streaming delta layered on top of the merged refine/version feature — not a second, divergent copy of it.

**(b) Undeclared behavior change: forward-to-Matter is hidden.** `SHOW_FORWARD_TO_MATTER = false` (`SummaryDetailPage.tsx:56`, gate at `:3127`) removes the "forward to Matter" action that `main` currently renders unconditionally on completed summaries. This is a user-visible feature removal that is unrelated to incremental rendering and is not part of issue #786. It is mentioned in the PR description, but please confirm this hiding is intentional and owned by someone — it should not ride in silently on a streaming PR.

**Spec: ❌** — feature is implemented, but the change cannot merge without a rebase, and it carries an out-of-scope behavior change.

## 2. Code quality

### P1 — Refine streams are never abortable (resource + UX leak)
`SummaryDetailPage.tsx` `handleRegenerateConfirm`: `api.streamRefinePersonalSummary` (~`:1069`) and `api.streamRefineSummary` (~`:1135`) are called **without a `signal`**. Unlike the generation streams, no `AbortController` is stored for them, and `clearAllTimers()` / `componentWillUnmount()` do not cancel them. If the user switches task, closes the page, or triggers refine repeatedly, the underlying `fetch` keeps the connection open and keeps consuming LLM tokens; the `if (this.taskId !== requestTaskId) return` checks only suppress the UI update, they do not close the socket. An orphaned stream that later errors can also surface a `Toast.error` on an unrelated page.
*Fix:* store a dedicated `refineStreamAbortController`, thread its `signal` into the refine `options`, and abort it in `componentWillUnmount` and on task change in `componentDidUpdate`.

### P2 — `nginx.conf.template`: buffering disabled for the whole location, not just SSE
`proxy_buffering off;` + `proxy_read_timeout 3600s;` are added to the entire `/summary/api/v1/` location, so **every** summary REST call (not just `/stream` endpoints) becomes unbuffered with a 1-hour read timeout. This exposes all summary REST traffic to slow-client / Slowloris-style resource pressure on the upstream.
*Fix:* prefer having the backend emit `X-Accel-Buffering: no` only on SSE responses (nginx honors it and unbuffers just those), keeping normal REST responses buffered; or split the SSE paths into their own `location`.

### P2 — No test coverage for any streaming logic
The SSE parser (`parseSSEBlock`, `consumeSSE`) and the stream lifecycle (start/stop/reset, done/error dedup, task-switch guards, fallback-to-poll) have zero unit tests. `parseSSEBlock` in particular (multi-line `data:`, `[DONE]`, JSON-vs-raw fallback, partial-chunk boundaries across `reader.read()`) is pure and highly testable. Given how much race logic rides on it, please add focused tests.

### P2 — Dead code: non-stream refine helpers
After this PR, `refineSummary` and `refinePersonalSummary` (the non-streaming variants) have no callers — `handleRegenerateConfirm` uses only the `stream*` versions. Remove them, or document why they are retained.

### Nits (non-blocking)
- `parseSSEBlock` uses `line.slice("data:".length).trimStart()`, which strips all leading whitespace rather than the single optional leading space the SSE spec defines. Harmless for JSON payloads, but technically non-conformant; `.replace(/^ /, "")` is stricter.
- `consumeSSE` never calls `reader.cancel()` / `releaseLock()`; on `AbortController.abort()` the ongoing `reader.read()` rejects and unwinds, so this is acceptable, but an explicit `reader.cancel()` in a `finally` would be cleaner.
- Inconsistent indentation at the regenerate-modal JSX (`SummaryDetailPage.tsx:3227`, the `<div className="summary-adjust-mode-list">` line is flush-left vs. its siblings).

**Quality: Changes-Requested** — driven by the P1 abort leak; the P2s are strongly recommended alongside the rebase.

## 3. Overall verdict
**REQUEST_CHANGES.** The streaming implementation is solid, but (1) the branch must be rebased onto current `main` so it stops re-shipping a divergent copy of the already-merged refine/version feature and can merge cleanly, and (2) the refine streams need to be abortable. Please also confirm the forward-to-Matter hiding is intentional, prefer scoping the nginx unbuffering to SSE endpoints, and add tests for the SSE parser.

## Coverage / what I could not verify
- Backend contract: I could not confirm the exact SSE event shape (`start`/`stage`/`delta`/`snapshot`/`done`/`error`, `content` vs `delta`, `[DONE]`) against the summary-api implementation — the frontend parser is tolerant, but event-name/field mismatches would silently degrade to polling.
- Runtime auth on the raw `fetch` path: `streamRequest` sets `token` / `X-Space-Id` / `Accept-Language` headers manually (bypassing the axios interceptor) but does not set `credentials`; whether cookies are required for the SSE endpoint depends on the deployment and was not verifiable statically.
- Real-device behavior of `proxy_read_timeout 3600s` under a proxy/load-balancer in front of nginx was not tested.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #792 (octo-web)

Reviewed at head `46d04d3287f56421ca31a97bcb81a158347d8812`. Scope: SSE streaming for incremental smart summary rendering, version history and restore, refine/regenerate streaming workflows, nginx SSE proxy config, i18n additions, and supporting type/API changes across 14 files.

## Summary

This is a substantial feature PR that adds three capabilities to the summary detail page: (1) real-time SSE streaming to render summary output incrementally as the backend generates it, replacing the prior polling-only path; (2) version history browsing and restore for both team and personal summary results; (3) a refine workflow that lets users give feedback to adjust a completed summary, with live streaming updates. The SSE infrastructure uses raw `fetch` + `ReadableStream` (necessary — axios doesn't support ReadableStream), with careful lifecycle management via abort controllers and multiple race-condition guards (`streamClosedTaskId`, `streamingTaskId`, `scheduleLoadSeq`). The nginx config adds the required SSE proxy settings. No P0/P1 defects found; four P2 items below address auth-edge handling, timeout coverage, implicit SSE contract fragility, and defensive buffer limits.

## Verification

Static analysis only at head `46d04d32`; build and tests not executed in this environment.

- ✅ **SSE data flow** — `streamSummary`/`streamRefineSummary`/`streamRefinePersonalSummary` flow through `streamRequest` → `fetch` → `consumeSSE` (ReadableStream reader) → `parseSSEBlock` (double-newline split, event/data line parsing, `[DONE]` sentinel) → `onEvent` callback → React state update. Each step traced and contracts are consistent.
- ✅ **Race condition guards** — `streamClosedTaskId` and `streamingTaskId` prevent duplicate stream connections per task. `scheduleLoadSeq` + `requestTaskId` pattern in every async method (`loadDetail`, `loadPersonalResult`, `loadMembers`, `handleRegenerateConfirm`) discards stale responses. Stream event handlers check `if (this.taskId !== taskId) return` before any state mutation.
- ✅ **Stream lifecycle** — `stopSummaryStream`/`stopTeamSummaryStream` abort via controller. `componentWillUnmount` calls `clearAllTimers()` which calls stop methods. `resetSummaryStreamForNewRun` clears `streamClosedTaskId` for regenerate flows. Done/error events set `streamClosedTaskId` to prevent reconnection to a completed stream.
- ✅ **Refine rollback** — `handleRegenerateConfirm` captures `previousPersonalResult`/`previousDetail` in `restoreRefineDraft` closures before streaming starts. On error, `restoreRefineDraft?.()` restores prior state. Nullified on success. Correct try/catch/finally structure.
- ✅ **Version API contracts** — `listSummaryVersions`/`restoreSummaryVersion`/`getSummaryVersion` and personal variants correctly typed. `handleViewVersion` verifies `detail.result_id === requestResultId` after fetch. Version detail modal passes `disableMemberPreview`/`disableTeamMemberPreview` through to `CitationBadge`/`CitationText`.
- ✅ **Scroll lock** — `syncVersionDetailScrollLock` toggles `summary-version-detail-open` class via `componentDidUpdate`. `componentWillUnmount` calls `setVersionDetailScrollLock(false)`. CSS class defined in `index.css` with `overflow: hidden`.
- ✅ **`generation_instruction` threading** — `scheduleItemToConfig` extracts `generation_instruction` in all 5 return branches. `CreateScheduleParams`/`UpdateScheduleParams`/`ScheduleConfig` all carry the field. `ScheduleConfigModal` renders textarea with maxLength=8000.
- ✅ **SSE parse correctness** — `parseSSEBlock` strips only the first space after `data:` (spec-compliant). Comment lines skipped. Multi-line data joined with `\n`. JSON parse failure falls back to `{ type: eventType, delta: data }`.
- ✅ **nginx config** — `proxy_buffering off`, `proxy_http_version 1.1`, `proxy_cache off`, `proxy_read_timeout 3600s` added to `/summary/api/v1` location block.
- ✅ **i18n parity** — 24 new keys in both `en-US.json` and `zh-CN.json` for streaming, refine, version history, and schedule instruction UI.
- ✅ **Test alignment** — `SummaryDetailPage.schedule.test.tsx` updated to reflect `renderTeamSummary` returning null for PROCESSING state, new `renderTeamGeneratingStatus` method, and multiCollabRegenerating showing previous result.

## Findings

No P0/P1 issues. Four P2 items below — all address robustness under edge conditions rather than correctness of the happy path.

### P2 — Streaming path bypasses summaryAxios 401 logout interceptor (`packages/dmworksummary/src/api/summaryApi.ts:231`)

`streamRequest` uses raw `fetch` with manually-built headers (`buildStreamHeaders`), intentionally bypassing `summaryAxios` (axios doesn't support ReadableStream). Request-side headers replicate token/X-Space-Id/Accept-Language correctly. However, the `summaryAxios` response interceptor at `:67-70` calls `WKApp.shared.logout()` on 401 — this protection does NOT apply to the streaming path. If the user's token expires during an active SSE stream, the fetch fails, fallback polling also fails, but the user sees an error state rather than being logged out.

```ts
// Suggested addition in streamRequest, after if (!resp.ok):
if (resp.status === 401) {
    WKApp.shared.logout();
}
```

**Diff-scope**: new — introduced by this PR's streaming infrastructure.

### P2 — No default timeout on `streamRequest` — indefinite SSE connections possible (`packages/dmworksummary/src/api/summaryApi.ts:224`)

`streamRequest` has no default timeout; the AbortSignal is entirely caller-provided. The non-streaming `refineSummary` and `refinePersonalSummary` set `timeout: 95000`. The streaming variants (`streamSummary`, `streamRefineSummary`, `streamRefinePersonalSummary`) pass no signal when callers don't provide one. If the server stalls without closing the connection, the UI stays in a permanent streaming/loading state. The nginx `proxy_read_timeout 3600s` provides a server-side ceiling but no client-side safety net.

```ts
// Suggested: add a default timeout in streamRequest when no caller signal:
const effectiveSignal = init.signal ?? AbortSignal.timeout(120_000);
```

**Diff-scope**: new — introduced by this PR's streaming infrastructure.

### P2 — Delta event handler: `event.content` replaces, `event.delta` appends — fragile implicit contract (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:878`)

In `startSummaryStream`, `startTeamSummaryStream`, and both refine stream handlers in `handleRegenerateConfirm`:

```ts
if (event.content) {
    this.setState({ streamingContent: event.content });   // FULL REPLACE
} else if (event.delta) {
    this.setState((prev) => ({ streamingContent: prev.streamingContent + event.delta }));  // APPEND
}
```

`event.content` is treated as full replacement; `event.delta` as incremental append. If the server ever sends both fields in a single delta event, `content` wins and accumulated draft is silently overwritten. The same pattern appears in the refine handlers where local `draft` accumulates incrementally. Not currently a bug (assuming server sends one field per event), but the contract is undocumented and fragile.

**Diff-scope**: new — introduced by this PR's streaming event handling.

### P2 — `consumeSSE` buffer can grow unbounded if server omits double-newline delimiters (`packages/dmworksummary/src/api/summaryApi.ts:198`)

The buffer in `consumeSSE` accumulates `decoder.decode(value)` without a size cap, draining only when `\n\n` is found. If the server sends a very large payload without double-newline delimiters, the buffer grows until the stream ends. The tail handler correctly processes remaining data for well-formed streams.

```ts
// Suggested: add a buffer ceiling
if (buffer.length > 1_000_000) {
    onEvent({ type: 'error', message: 'SSE stream exceeded buffer limit' });
    reader.cancel();
    return;
}
```

**Diff-scope**: new — introduced by this PR's SSE consumption.

## Data Flow Trace

| Consumed data | Upstream source | Verified |
|---|---|---|
| `streamingContent` (UI render) | `parseSSEBlock` → `onEvent` → `setState` in `startSummaryStream` | ✅ content replaces / delta appends, done triggers loadDetail |
| `teamStreamingContent` (UI render) | Same SSE path via `startTeamSummaryStream` | ✅ mirrors personal stream with team scope |
| `draft` (refine handler) | Accumulated from `event.content \|\| (draft + event.delta)` in `onEvent` | ✅ content replaces, delta appends; finalized by done event |
| `refined` (post-refine state) | Captured from done event; applied in `setState` after `await` completes | ✅ result_id, version, content, citations all threaded correctly |
| `restoreRefineDraft` rollback | Closure over `previousPersonalResult`/`previousDetail` captured pre-stream | ✅ called on error, nullified on success |
| `versionDetail` (modal) | `getSummaryVersion`/`getPersonalSummaryVersion` → verified `result_id` match | ✅ stale-result guard in place |
| `versions` / `personalVersions` (strip UI) | `listSummaryVersions`/`listPersonalSummaryVersions` with limit=3 | ✅ taskId guard on response |
| `generation_instruction` → API | `scheduleItemToConfig` all 5 branches → `ScheduleConfig` → save API | ✅ threaded correctly |

- **C1 (dual-path parity)**: Clear — personal and team stream start/stop/reset methods are symmetrically implemented. Personal refine and team refine follow the same streaming pattern. Version restore for both personal and team paths have identical guard patterns.
- **C2 (control-flow ordering / nested reuse)**: Clear — `consumeSSE` is called once per stream connection. `parseSSEBlock` is called per-block within the consumer loop. No nested reuse or ordering conflicts.
- **C3 (auth boundary ≠ capability boundary)**: See P2 finding on 401 interceptor bypass — the streaming path does not inherit the summaryAxios 401 logout. This is a gap but not a P1 since the stream fails and polling also fails.
- **C4 (auth lifecycle / container-member cascade)**: N/A — no auth/lifecycle changes beyond the streaming auth header replication.
- **C5 (build/note ≠ runtime correctness)**: nginx config (`proxy_buffering off`, `proxy_http_version 1.1`) is correct for SSE. `proxy_read_timeout 3600s` is reasonable. The `buildSummaryURL` function correctly handles both browser and extension contexts via `resolveSummaryBaseURL()`.
- **C6 (governance/docs self-consistency)**: N/A — no governance or policy docs changed.

N/A — this is the first review pass on this PR.

- `buildStreamHeaders` correctly replicates all headers from the `summaryAxios` request interceptor (token, X-Space-Id, Accept-Language via `buildAcceptLanguage()`).
- `parseSSEBlock` handles `[DONE]` sentinel, JSON parse errors, and multi-line data per the SSE spec.
- `componentDidUpdate` on task switch resets all streaming/version/regenerate state. Old stream events safely guarded by `if (this.taskId !== taskId) return`.
- `handleRegenerateConfirm` has 4+ distinct branches — each properly manages loading state, success toast, and error rollback.
- `isMultiCollabRegenerating()` shows previous team summary instead of generating overlay — correct UX.
- `appendLocalScheduleInstruction` and `resetLocalScheduleInstruction` provide optimistic local updates.
- `ChatSummaryNewModal` and `SummaryCreatePage` correctly pass `showGenerationInstruction={false}`.
- `extractErrorMessage` enhanced with 404/503 specific messages for refine API availability.

## Verdict

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. The SSE streaming infrastructure is well-designed with thorough race-condition guarding (`scheduleLoadSeq`, `streamClosedTaskId`, `streamingTaskId`), correct lifecycle management (abort controllers, streamClosedTaskId), and clean rollback patterns (`restoreRefineDraft`). Version history and restore APIs are correctly typed and guarded. The four P2 items are non-blocking robustness improvements (401 interceptor parity, client-side timeout, delta contract documentation, buffer ceiling) that can be addressed in follow-up work.

Jerry-Xin (changes requested):
## Summary
Correcting my own earlier COMMENT: after a peer flagged the refine-stream lifecycle, I byte-verified it and confirmed a real leak/unmount defect I had missed. Upgrading to REQUEST_CHANGES.

## 🔴 Blocking

**1. Refine streams are uncancellable + setState-after-unmount (memory leak / state on dead component).**
`handleRegenerate` consumes `streamRefinePersonalSummary` (SummaryDetailPage.tsx:1069) and `streamRefineSummary` (:1135). Both API functions accept a `signal?: AbortSignal` (summaryApi.ts:289,343), but the call sites pass **no signal** — the options object is only `{ feedback, base_result_id, base_version }` + `{ onEvent }`. There is no refine AbortController field on the component at all.

Each `onEvent` runs `this.setState(...)` on every `delta`, guarded only by `this.taskId !== requestTaskId` (a *task-switch* guard, not an *unmount* guard). If the component unmounts mid-refine, the in-flight `for await` keeps consuming the SSE body and firing `this.setState` on an unmounted component → React unmounted-setState warning + wasted work + the stream is never aborted (network + reader kept alive) = the leak described.

Compounding: `componentWillUnmount` (line ~ shown below) does **not** abort any stream — it only removes window listeners + `clearAllTimers()`. The two `.abort()` calls (:921, :993) live inside the main-stream start/stop helpers, never in unmount. So even the main summary/team streams (`streamAbortController`/`teamStreamAbortController`) are not aborted on unmount — only reset when a *new* stream starts or an explicit stop runs.

Fix: create a refine AbortController, pass its `signal` into both refine calls, and abort all three controllers (`streamAbortController`, `teamStreamAbortController`, and the new refine one) in `componentWillUnmount`. That closes both the refine leak and the main-stream unmount leak.

**2. Merge conflict — branch is CONFLICTING with base; rebase before merge.** (mergeStateStatus=DIRTY.)

## 💬 Non-blocking
- Refine `onEvent` distinguishes `delta`/`error`/`done` but there's no explicit snapshot-vs-incremental discrimination; confirm the backend refine protocol only ever emits deltas (not a full-snapshot event that would need replace-not-append) so partial content can't duplicate or drop.
- The fallback poll path and the streaming path both write `content`; a shared sequence/cursor guard across poll+stream would harden against a late poll response briefly overwriting fresher streamed content.

## ✅ Verified sound
- Main incremental-render XSS sink: every render goes through ReactMarkdown + rehypeSanitize (citationSchema strips script/event-handlers/js: URIs); partial markdown safe each tick.
- Main summary/team stream ordering: controller-abort before new stream, streamingTaskId dedup, per-event stale-discard, in-order functional-setState append, done→loadDetail() reconcile.
- Auth/scope: token + X-Space-Id on stream headers, matching the non-stream path.
- No -97 regression: old render path replaced coherently; error/empty states preserved.

## Parity checked
Base→head capability diff: the non-incremental render/regenerate-modal path is replaced by the SSE streaming path. Rendered-output correctness, error state, and empty state are preserved; the refine feature is retained (personal + team). The regression is the new refine path's missing cancellation/unmount handling (blocking #1 above), not a dropped capability.

Path to APPROVE: wire AbortSignal into the refine calls + abort all stream controllers in componentWillUnmount; resolve the merge conflict; get code CI green.

PR #794, opened 2026-07-15
feat(summary): render smart summary output incrementally

### LINKED ISSUE #786: feat(summary): stream summary output to the frontend and render incrementally

Liuxy20:
## Summary

- Consume backend streaming endpoints for smart-summary output.
- Render summary content incrementally while generation is in progress.
- Support streaming display for personal summary, team summary, regenerate, and refine flows where available.
- Fall back to existing polling when streaming is interrupted or unavailable.
- Refresh final detail/result after stream completion and use the persisted result as the source of truth.
- Keep workflow/progress display order stable while streaming content updates.
- Hide the forward-to-matter action on summary detail behind a local feature flag.
- Update nginx summary API proxy settings for streaming responses.

## Implementation

- Add summary stream API helpers.
- Add incremental rendering state and stream handling in the summary detail page.
- Keep polling as fallback and final result refresh path.
- Configure nginx buffering/timeout settings for summary streaming.

## Testing

Closes #786

### DIFF
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -52,6 +52,9 @@ interface SummaryDetailPageProps {
     taskId?: number;
 }
 
+// Matters 转发入口暂时隐藏，保留相关代码和弹窗，后续需要时打开此开关即可。
+const SHOW_FORWARD_TO_MATTER = false;
+
 type RegenerateMode = "refine" | "full";
 type RefineLoadingTarget = "personal" | "team" | "summary";
 
@@ -105,6 +108,12 @@ interface SummaryDetailPageState {
     workflowDisplayIndex: number;
     workflowGateContent: boolean;
     workflowRevealDone: boolean;
+    streaming: boolean;
+    streamingContent: string;
+    streamError: string | null;
+    teamStreaming: boolean;
+    teamStreamingContent: string;
+    teamStreamError: string | null;
 }
 
 const INTER_MESSAGE_DELAY_MS = 200;
@@ -193,6 +202,12 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
         workflowDisplayIndex: -1,
         workflowGateContent: false,
         workflowRevealDone: false,
+        streaming: false,
+        streamingContent: "",
+        streamError: null,
+        teamStreaming: false,
+        teamStreamingContent: "",
+        teamStreamError: null,
     };
 
     private personalPollTimer: ReturnType<typeof setInterval> | null = null;
@@ -200,6 +215,14 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     private fallbackStartTimeout: ReturnType<typeof setTimeout> | null = null;
     private workflowAdvanceTimer: ReturnType<typeof setTimeout> | null = null;
     private workflowCompleteTimer: ReturnType<typeof setTimeout> | null = null;
+    private streamAbortController: AbortController | null = null;
+    private streamingTaskId: number | null = null;
+    private streamClosedTaskId: number | null = null;
+    private teamStreamAbortController: AbortController | null = null;
+    private teamStreamingTaskId: number | null = null;
+    private teamStreamClosedTaskId: number | null = null;
+    private refineStreamAbortController: AbortController | null = null;
+    private unmounted = false;
     private workflowTargetIndex = -1;
     private listPageActive = false;
     private lastEventTime = 0;
@@ -217,6 +240,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     }
 
     componentDidMount() {
+        this.unmounted = false;
         window.addEventListener("summary-status-change", this.handleStatusChangeEvent);
         window.addEventListener("summary-batch-heartbeat", this.handleBatchHeartbeat);
         window.addEventListener("summary-list-unmount", this.handleListPageUnmount);
@@ -243,7 +267,15 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                 restoringVersionId: null,
                 restoringPersonalVersionId: null,
                 pendingScheduleInstruction: "",
+                streamingContent: "",
+                streamError: null,
+                streaming: false,
+                teamStreamingContent: "",
+                teamStreamError: null,
+                teamStreaming: false,
             });
+            this.streamClosedTaskId = null;
+            this.teamStreamClosedTaskId = null;
             this.loadDetail();
         }
         if (prevState && prevState.showVersionDetailModal !== this.state.showVersionDetailModal) {
@@ -252,6 +284,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     }
 
     componentWillUnmount() {
+        this.unmounted = true;
         window.removeEventListener("summary-status-change", this.handleStatusChangeEvent);
         window.removeEventListener("summary-batch-heartbeat", this.handleBatchHeartbeat);
         window.removeEventListener("summary-list-unmount", this.handleListPageUnmount);
@@ -260,6 +293,9 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     }
 
     private clearAllTimers() {
+        this.stopSummaryStream();
+        this.stopTeamSummaryStream();
+        this.stopRefineStream();
         if (this.personalPollTimer) {
             clearInterval(this.personalPollTimer);
             this.personalPollTimer = null;
@@ -466,8 +502,18 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                 detail.status === TaskStatus.WAITING_CONFIRM
             ) {
                 this.startFallbackPoll();
+                if (detail.summary_mode === SummaryMode.BY_PERSON && this.streamClosedTaskId !== detail.task_id) {
+                    this.startSummaryStream(detail.task_id);
+                }
+                const isMultiByPerson = detail.summary_mode === SummaryMode.BY_PERSON && ((detail.participants?.length || 0) > 1);
+                if (isMultiByPerson && this.teamStreamClosedTaskId !== detail.task_id) {
+                    this.startTeamSummaryStream(detail.task_id);
+                }
             } else {
                 this.stopFallbackPoll();
+                this.stopSummaryStream(true);
+                this.stopTeamSummaryStream();
+                this.setState({ teamStreaming: false, teamStreamingContent: "", teamStreamError: null });
             }
             // Load BY_PERSON data
             if (detail.summary_mode === SummaryMode.BY_PERSON) {
@@ -813,6 +859,199 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
         }
     }
 
+
+    private startSummaryStream(taskId: number) {
+        if (this.streamClosedTaskId === taskId) return;
+        if (this.streamingTaskId === taskId && this.streamAbortController) return;
+        this.stopSummaryStream();
+        const controller = new AbortController();
+        this.streamAbortController = controller;
+        this.streamingTaskId = taskId;
+        this.setState({ streaming: true, streamError: null, streamingContent: "" });
+        let terminalReceived = false;
+        void api.streamSummary(taskId, {
+            scope: "personal",
+            signal: controller.signal,
+            onEvent: (event) => {
+                if (this.taskId !== taskId) return;
+                if (event.type === "start") {
+                    this.setState({ streamingContent: "", streamError: null });
+                    return;
+                }
+                if (event.type === "stage" && event.stage) {
+                    const idx = this.workflowStageIndex(event.stage);
+                    if (idx >= 0) this.setWorkflowStageDirectly(idx);
+                    return;
+                }
+                if (event.type === "delta") {
+                    if (event.content) {
+                        this.setState({ streamingContent: event.content });
+                    } else if (event.delta) {
+                        this.setState((prev) => ({ streamingContent: prev.streamingContent + event.delta }));
+                    }
+                    return;
+                }
+                if (event.type === "snapshot") {
+                    this.setState({ streamingContent: event.content || "" });
+                    return;
+                }
+                if (event.type === "done") {
+                    terminalReceived = true;
+                    this.streamAbortController = null;
+                    this.streamingTaskId = null;
+                    this.streamClosedTaskId = taskId;
+                    this.setState({ streaming: false });
+                    this.loadDetail();
+                    return;
+                }
+                if (event.type === "error") {
+                    terminalReceived = true;
+                    this.streamAbortController = null;
+                    this.streamingTaskId = null;
+                    this.streamClosedTaskId = taskId;
+                    this.setState({ streaming: false, streamError: event.message || null });
+                    this.loadDetail();
+                }
+            },
+        }).then(() => {
+            if (terminalReceived || controller.signal.aborted || this.taskId !== taskId) return;
+            this.streamAbortController = null;
+            this.streamingTaskId = null;
+            this.streamClosedTaskId = taskId;
+            this.setState({ streaming: false });
+            this.startFallbackPoll();
+        }).catch((err: any) => {
+            if (controller.signal.aborted || this.taskId !== taskId) return;
+            this.streamAbortController = null;
+            this.streamingTaskId = null;
+            this.streamClosedTaskId = taskId;
+            this.setState({ streaming: false, streamError: err?.message || null });
+            // Keep the existing polling path as the reliability fallback.
+            this.startFallbackPoll();
+        });
+    }
+
+    private stopSummaryStream(resetState = false) {
+        if (this.streamAbortController) {
+            this.streamAbortController.abort();
+            this.streamAbortController = null;
+        }
+        this.streamingTaskId = null;
+        if (resetState && !this.unmounted) {
+            this.setState({ streaming: false, streamingContent: "", streamError: null });
+        }
+    }
+
+    private stopRefineStream() {
+        if (this.refineStreamAbortController) {
+            this.refineStreamAbortController.abort();
+            this.refineStreamAbortController = null;
+        }
+    }
+
+    private isRefineDonePayload(event: api.SummaryStreamEvent) {
+        return event.type === "done" && (
+            event.result_id != null ||
+            event.version != null ||
+            event.version_id != null ||
+            event.content != null ||
+            event.citations != null ||
+            event.team_citations != null ||
+            event.generated_at != null
+        );
+    }
+
+    private resetSummaryStreamForNewRun(taskId: number) {
+        this.stopSummaryStream();
+        this.streamClosedTaskId = null;
+        this.setState({ streaming: false, streamingContent: "", streamError: null });
+        if (this.taskId === taskId) {
+            this.startSummaryStream(taskId);
+        }
+    }
+
+    private startTeamSummaryStream(taskId: number) {
+        if (this.teamStreamClosedTaskId === taskId) return;
+        if (this.teamStreamingTaskId === taskId && this.teamStreamAbortController) return;
+        this.stopTeamSummaryStream();
+        const controller = new AbortController();
+        this.teamStreamAbortController = controller;
+        this.teamStreamingTaskId = taskId;
+        this.setState({ teamStreaming: true, teamStreamError: null, teamStreamingContent: "" });
+        let terminalReceived = false;
+        void api.streamSummary(taskId, {
+            scope: "team",
+            signal: controller.signal,
+            onEvent: (event) => {
+                if (this.taskId !== taskId) return;
+                if (event.type === "start") {
+                    this.setState({ teamStreamingContent: "", teamStreamError: null });
+                    return;
+                }
+                if (event.type === "delta") {
+                    if (event.content) {
+                        this.setState({ teamStreamingContent: event.content });
+                    } else if (event.delta) {
+                        this.setState((prev) => ({ teamStreamingContent: prev.teamStreamingContent + event.delta }));
+                    }
+                    return;
+                }
+                if (event.type === "snapshot") {
+                    this.setState({ teamStreamingContent: event.content || "" });
+                    return;
+                }
+                if (event.type === "done") {
+                    terminalReceived = true;
+                    this.teamStreamAbortController = null;
+                    this.teamStreamingTaskId = null;
+                    this.teamStreamClosedTaskId = taskId;
+                    this.setState({ teamStreaming: false });
+                    this.loadDetail();
+                    return;
+                }
+                if (event.type === "error") {
+                    terminalReceived = true;
+                    this.teamStreamAbortController = null;
+                    this.teamStreamingTaskId = null;
+                    this.teamStreamClosedTaskId = taskId;
+                    this.setState({ teamStreaming: false, teamStreamError: event.message || null });
+                    this.loadDetail();
+                }
+            },
+        }).then(() => {
+            if (terminalReceived || controller.signal.aborted || this.taskId !== taskId) return;
+            this.teamStreamAbortController = null;
+            this.teamStreamingTaskId = null;
+            this.teamStreamClosedTaskId = taskId;
+            this.setState({ teamStreaming: false });
+            this.startFallbackPoll();
+        }).catch((err: any) => {
+            if (controller.signal.aborted || this.taskId !== taskId) return;
+            this.teamStreamAbortController = null;
+            this.teamStreamingTaskId = null;
+            this.teamStreamClosedTaskId = taskId;
+            this.setState({ teamStreaming: false, teamStreamError: err?.message || null });
+            this.startFallbackPoll();
+        });
+    }
+
+    private stopTeamSummaryStream() {
+        if (this.teamStreamAbortController) {
+            this.teamStreamAbortController.abort();
+            this.teamStreamAbortController = null;
+        }
+        this.teamStreamingTaskId = null;
+    }
+
+    private resetTeamSummaryStreamForNewRun(taskId: number) {
+        this.stopTeamSummaryStream();
+        this.teamStreamClosedTaskId = null;
+        this.setState({ teamStreaming: false, teamStreamingContent: "", teamStreamError: null });
+        if (this.taskId === taskId) {
+            this.startTeamSummaryStream(taskId);
+        }
+    }
+
     handleRegenerate = () => {
         const { detail } = this.state;
         if (this.taskId == null) return;
@@ -858,32 +1097,77 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             ? this.state.refineFeedback.trim()
             : this.state.regenerateTopic.trim();
         if (!trimmed) return;
+        this.stopRefineStream();
         this.setState({ regenerateSubmitting: true });
+        let restoreRefineDraft: (() => void) | null = null;
+        let currentRefineController: AbortController | null = null;
         try {
             const operateOnTeamSummary = this.shouldOperateOnTeamSummary();
             if (regenerateMode === "refine") {
                 if (detail?.summary_mode === SummaryMode.BY_PERSON && !operateOnTeamSummary) {
                     const baseResultId = this.state.personalResult?.id;
                     if (!baseResultId) return;
+                    const previousPersonalResult = this.state.personalResult;
+                    const previousMembers = this.state.members;
+                    restoreRefineDraft = () => {
+                        if (this.taskId === requestTaskId) this.setState({ personalResult: previousPersonalResult, members: previousMembers });
+                    };
                     this.setState({ showRegenerateModal: false, refineLoadingTarget: "personal" });
-                    const refined = await api.refinePersonalSummary(requestTaskId, {
+                    const refineController = new AbortController();
+                    currentRefineController = refineController;
+                    this.refineStreamAbortController = refineController;
+                    let draft = "";
+                    let refined: api.SummaryStreamEvent | null = null;
+                    await api.streamRefinePersonalSummary(requestTaskId, {
                         feedback: trimmed,
                         base_result_id: baseResultId,
                         base_version: this.state.personalResult?.version,
+                    }, {
+                        signal: refineController.signal,
+                        onEvent: (event) => {
+                            if (this.taskId !== requestTaskId || refineController.signal.aborted) return;
+                            if (event.type === "delta") {
+                                draft = event.content || (draft + (event.delta || ""));
+                                const myUid = WKApp.loginInfo.uid;
+                                this.setState((prev) => ({
+                                    personalResult: prev.personalResult ? { ...prev.personalResult, content: draft } : prev.personalResult,
+                                    members: prev.members.map((m) => m.user_id === myUid ? { ...m, content: draft } : m),
+                                } as Pick<SummaryDetailPageState, "personalResult" | "members">));
+                                return;
+                            }
+                            if (event.type === "snapshot") {
+                                draft = event.content || "";
+                                const myUid = WKApp.loginInfo.uid;
+                                this.setState((prev) => ({
+                                    personalResult: prev.personalResult ? { ...prev.personalResult, content: draft } : prev.personalResult,
+                                    members: prev.members.map((m) => m.user_id === myUid ? { ...m, content: draft } : m),
+                                } as Pick<SummaryDetailPageState, "personalResult" | "members">));
+                                return;
+                            }
+                            if (event.type === "error") {
+                                throw new Error(event.message || t("summary.common.operationFailed"));
+                            }
+                            if (this.isRefineDonePayload(event)) {
+                                refined = event;
+                            }
+                        },
                     });
-                    if (this.taskId !== requestTaskId) return;
+                    if (this.refineStreamAbortController === refineController) this.refineStreamAbortController = null;
+                    if (this.taskId !== requestTaskId || refineController.signal.aborted) return;
+                    if (!refined) throw new Error(t("summary.common.operationFailed"));
+                    restoreRefineDraft = null;
                     Toast.success(t("summary.detail.refineSuccess"));
                     this.appendLocalScheduleInstruction(trimmed);
                     this.reloadScheduleAfterInstructionChange(requestTaskId);
                     this.setState((prev) => {
                         const nextPersonal = prev.personalResult ? {
                             ...prev.personalResult,
-                            id: refined.result_id || prev.personalResult.id,
-                            version: refined.version,
-                            content: refined.content,
-                            citations: (refined.citations as any) || prev.personalResult.citations,
-                            msg_count: refined.msg_count ?? prev.personalResult.msg_count,
-                            generated_at: refined.generated_at || prev.personalResult.generated_at,
+                            id: refined!.result_id || prev.personalResult.id,
+                            version: refined!.version || prev.personalResult.version,
+                            content: refined!.content || draft,
+                            citations: (refined!.citations as any) || prev.personalResult.citations,
+                            msg_count: refined!.msg_count ?? prev.personalResult.msg_count,
+                            generated_at: refined!.generated_at || prev.personalResult.generated_at,
                         } : prev.personalResult;
                         const myUid = WKApp.loginInfo.uid;
                         return {
@@ -892,8 +1176,8 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                             personalResult: nextPersonal,
                             members: prev.members.map((m) => m.user_id === myUid ? {
                                 ...m,
-                                content: refined.content,
-                                citations: (refined.citations as any) || m.citations,
+                                content: refined!.content || draft,
+                                citations: (refined!.citations as any) || m.citations,
                             } : m),
                         } as Pick<SummaryDetailPageState, "showRegenerateModal" | "refineLoadingTarget" | "personalResult" | "members">;
                     });
@@ -901,12 +1185,61 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                 } else {
                     const baseResultId = detail?.result_id;
                     if (!baseResultId) return;
+                    const previousDetail = this.state.detail;
+                    restoreRefineDraft = () => {
+                        if (this.taskId === requestTaskId) this.setState({ detail: previousDetail });
+                    };
                     this.setState({
                         showRegenerateModal: false,
                         refineLoadingTarget: operateOnTeamSummary ? "team" : "summary",
                     });
-                    const refined = await api.refineSummary(requestTaskId, { feedback: trimmed, base_result_id: baseResultId });
-                    if (this.taskId !== requestTaskId) return;
+                    const refineController = new AbortController();
+                    currentRefineController = refineController;
+                    this.refineStreamAbortController = refineController;
+                    let draft = "";
+                    let refined: api.SummaryStreamEvent | null = null;
+                    await api.streamRefineSummary(requestTaskId, { feedback: trimmed, base_result_id: baseResultId }, {
+                        signal: refineController.signal,
+                        onEvent: (event) => {
+                            if (this.taskId !== requestTaskId || refineController.signal.aborted) return;
+                            if (event.type === "delta") {
+                                draft = event.content || (draft + (event.delta || ""));
+                                this.setState((prev) => {
+                                    if (!prev.detail?.result) return null;
+                                    return {
+                                        detail: {
+                                            ...prev.detail,
+                                            result: { ...prev.detail.result, content: draft },
+                                        },
+                                    } as Pick<SummaryDetailPageState, "detail">;
+                                });
+                                return;
+                            }
+                            if (event.type === "snapshot") {
+                                draft = event.content || "";
+                                this.setState((prev) => {
+                                    if (!prev.detail?.result) return null;
+                                    return {
+                                        detail: {
+                                            ...prev.detail,
+                                            result: { ...prev.detail.result, content: draft },
+                                        },
+                                    } as Pick<SummaryDetailPageState, "detail">;
+                                });
+                                return;
+                            }
+                            if (event.type === "error") {
+                                throw new Error(event.message || t("summary.common.operationFailed"));
+                            }
+                            if (this.isRefineDonePayload(event)) {
+                                refined = event;
+                            }
+                        },
+                    });
+                    if (this.refineStreamAbortController === refineController) this.refineStreamAbortController = null;
+                    if (this.taskId !== requestTaskId || refineController.signal.aborted) return;
+                    if (!refined) throw new Error(t("summary.common.operationFailed"));
+                    restoreRefineDraft = null;
                     Toast.success(t("summary.detail.refineSuccess"));
                     this.appendLocalScheduleInstruction(trimmed);
                     this.reloadScheduleAfterInstructionChange(requestTaskId);
@@ -917,20 +1250,20 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                             refineLoadingTarget: null,
                             detail: {
                                 ...prev.detail,
-                                result_id: refined.result_id,
+                                result_id: refined!.result_id || prev.detail.result_id,
                                 result: {
                                     ...prev.detail.result,
-                                    content: refined.content,
-                                    version: refined.version,
-                                    citations: (refined.citations as any) || prev.detail.result.citations,
-                                    team_citations: (refined.team_citations as any) || prev.detail.result.team_citations,
-                                    total_msg_count: refined.total_msg_count ?? prev.detail.result.total_msg_count,
-                                    total_token_used: refined.total_token_used ?? prev.detail.result.total_token_used,
-                                    model_version: refined.model_version || prev.detail.result.model_version,
-                                    operation_type: refined.operation_type,
-                                    operation_note: refined.operation_note,
-                                    parent_result_id: refined.parent_result_id,
-                                    generated_at: refined.generated_at || prev.detail.result.generated_at,
+                                    content: refined!.content || draft,
+                                    version: refined!.version || prev.detail.result.version,
+                                    citations: (refined!.citations as any) || prev.detail.result.citations,
+                                    team_citations: (refined!.team_citations as any) || prev.detail.result.team_citations,
+                                    total_msg_count: refined!.total_msg_count ?? prev.detail.result.total_msg_count,
+                                    total_token_used: refined!.total_token_used ?? prev.detail.result.total_token_used,
+                                    model_version: refined!.model_version || prev.detail.result.model_version,
+                                    operation_type: refined!.operation_type,
+                                    operation_note: refined!.operation_note,
+                                    parent_result_id: refined!.parent_result_id,
+                                    generated_at: refined!.generated_at || prev.detail.result.generated_at,
                                 },
                             },
                         } as Pick<SummaryDetailPageState, "showRegenerateModal" | "refineLoadingTarget" | "detail">;
@@ -942,6 +1275,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                     await api.regenerateSummary(requestTaskId, { topic: trimmed });
                     if (this.taskId !== requestTaskId) return;
                     Toast.success(t("summary.detail.regenerateStarted"));
+                    this.resetTeamSummaryStreamForNewRun(requestTaskId);
                     this.resetLocalScheduleInstruction(trimmed);
                     this.setState((prev) => prev.detail ? {
                         showRegenerateModal: false,
@@ -956,6 +1290,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                     await api.regeneratePersonalSummary(requestTaskId, { topic: trimmed });
                     if (this.taskId !== requestTaskId) return;
                     Toast.success(t("summary.detail.regenerateStarted"));
+                    this.resetSummaryStreamForNewRun(requestTaskId);
                     this.resetLocalScheduleInstruction(trimmed);
                     this.setState((prev) => ({
                         showRegenerateModal: false,
@@ -986,6 +1321,9 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                     await api.regenerateSummary(requestTaskId, { topic: trimmed });
                     if (this.taskId !== requestTaskId) return;
                     Toast.success(t("summary.detail.regenerateStarted"));
+                    if (detail?.summary_mode === SummaryMode.BY_PERSON) {
+                        this.resetSummaryStreamForNewRun(requestTaskId);
+                    }
                     this.resetLocalScheduleInstruction(trimmed);
                     this.setState({ showRegenerateModal: false });
                     this.loadDetail();
@@ -994,10 +1332,18 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             if (this.taskId !== requestTaskId) return;
             window.dispatchEvent(new CustomEvent("summary-task-regenerated", { detail: { taskId: requestTaskId } }));
         } catch (err: any) {
+            if (err?.name === "AbortError" || this.taskId !== requestTaskId || this.unmounted) return;
+            restoreRefineDraft?.();
             this.setState({ refineLoadingTarget: null });
             Toast.error(err.message || t("summary.common.operationFailed"));
         } finally {
-            this.setState({ regenerateSubmitting: false });
+            if (currentRefineController && this.refineStreamAbortController === currentRefineController) {
+                this.refineStreamAbortController.abort();
+                this.refineStreamAbortController = null;
+            }
+            if (this.taskId === requestTaskId && !this.unmounted) {
+                this.setState({ regenerateSubmitting: false });
+            }
         }
     };
 
@@ -1678,6 +2024,33 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                 <div className="summary-detail-team-generating">
                     <Spin size="small" />
                     <span>{t(descKey)}</span>
+                </div>
+            </div>
+        );
+    }
+
+    renderStreamingContent() {
+        const content = this.state.streamingContent.trim();
+        if (!this.state.streaming || !content) return null;
+        return (
+            <div className="summary-detail-personal summary-detail-streaming">
+                <div className="summary-detail-content-box">
+                    <CitationText content={this.state.streamingContent} citations={[]} />
+                </div>
+            </div>
+        );
+    }
+
+    renderTeamStreamingContent() {
+        const content = this.state.teamStreamingContent.trim();
+        if (!this.state.teamStreaming || !content) return null;
+        return (
+            <div className="summary-detail-team summary-detail-streaming">
+                <div className="summary-detail-section-header">
+                    <span>{this.context.t("summary.detail.teamSummary")}</span>
+                </div>
+                <div className="summary-detail-content-box">
+                    <CitationText content={this.state.teamStreamingContent} citations={[]} members={this.state.members} />
                 </div>
             </div>
         );
@@ -2839,7 +3212,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                                 {t("summary.detail.forwardToChat")}
                             </Button>
                         )}
-                        {detail && detail.status === TaskStatus.COMPLETED && (
+                        {SHOW_FORWARD_TO_MATTER && detail && detail.status === TaskStatus.COMPLETED && (
                             <Button
                                 theme="borderless"
                                 icon={<IconSend />}
@@ -2942,20 +3315,23 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                                         {/* need1：多人协作不再单独显示「我的总结」区块（自己内容改到参与者报告
                                             里我那条，need3）；单人 BY_PERSON 维持显示「我的总结」及其行内编辑。 */}
                                         {!this.isMultiCollab() && (
-                                            this.state.isEditing && this.state.personalResult && detail.result_id ? (
-                                                <div className="summary-detail-personal">
-                                                    <h3>{t("summary.detail.mySummaryPlain")}</h3>
-                                                    <SummaryEditor
-                                                        taskId={detail.task_id}
-                                                        baseResultId={detail.result_id}
-                                                        initialContent={this.state.personalResult.content || ""}
-                                                        onSave={this.handleEditSave}
-                                                        onCancel={this.handleEditCancel}
-                                                    />
-                                                </div>
-                                            ) : (
-                                                this.renderPersonalSummary()
-                                            )
+                                            <>
+                                                {this.shouldShowProcessingCard() && !this.personalReady && this.renderProcessing()}
+                                                {this.state.isEditing && this.state.personalResult && detail.result_id ? (
+                                                    <div className="summary-detail-personal">
+                                                        <h3>{t("summary.detail.mySummaryPlain")}</h3>
+                                                        <SummaryEditor
+                                                            taskId={detail.task_id}
+                                                            baseResultId={detail.result_id}
+                                                            initialContent={this.state.personalResult.content || ""}
+                                                            onSave={this.handleEditSave}
+                                                            onCancel={this.handleEditCancel}
+                                                        />
+                                                    </div>
+                                                ) : (
+                                                    this.renderPersonalSummary()
+                                                )}
+                                            </>
                                         )}
                                         {/* 回归修复：多人协作页给「我自己」补回「提交给全部」轻量入口。
                                             不违反 need1（不恢复我的总结正文区），仅一句提示 + 提交按钮。
@@ -2963,17 +3339,28 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                                         {this.isMultiCollabRegenerating() && this.shouldShowWorkflowCard() && this.renderProcessing()}
                                         {this.isMultiCollabRegenerating() && !this.shouldShowWorkflowCard() && this.renderTeamGeneratingStatus()}
                                         {this.renderMySubmitBar()}
-                                        {this.renderTeamSummary()}
+                                        {this.state.teamStreaming && this.state.teamStreamingContent.trim() ? (
+                                            <>
+                                                {this.renderTeamStreamingContent()}
+                                                {this.renderStreamingContent()}
+                                            </>
+                                        ) : (
+                                            <>
+                                                {this.renderStreamingContent()}
+                                                {this.renderTeamSummary()}
+                                            </>
+                                        )}
                                         {this.renderMemberStatus()}
                                         {this.renderParticipantReports()}
                                     </>
                                 )}
 
-                                {this.shouldShowProcessingCard() &&
-                                    !this.isMultiCollabRegenerating() &&
+                                {detail.summary_mode !== SummaryMode.BY_PERSON &&
+                                    this.shouldShowProcessingCard() &&
                                     !this.personalReady &&
                                     this.renderProcessing()
                                 }
+                                {detail.summary_mode !== SummaryMode.BY_PERSON && this.renderStreamingContent()}
 
                                 {detail.status === TaskStatus.FAILED && this.renderFailed()}
 
--- a/nginx.conf.template
+++ b/nginx.conf.template
@@ -86,6 +86,11 @@ server {
             # forwards the rewritten request as-is.
             rewrite ^/summary/api/v1/(.*)$ /api/v1/$1 break;
             proxy_pass  $summary_api_url;
+            # SSE endpoints under /summary/api/v1 require unbuffered streaming.
+            proxy_http_version 1.1;
+            proxy_buffering off;
+            proxy_cache off;
+            proxy_read_timeout 3600s;
             proxy_set_header Host $host;
             proxy_set_header X-Real-IP $remote_addr;
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
--- a/packages/dmworksummary/src/api/summaryApi.ts
+++ b/packages/dmworksummary/src/api/summaryApi.ts
@@ -132,6 +132,150 @@ async function del<T>(path: string): Promise<T> {
 
 // ─── Core Summary Operations ───────────────────────────
 
+
+export interface SummaryStreamEvent {
+    type: "start" | "stage" | "delta" | "snapshot" | "done" | "error" | string;
+    task_id?: number;
+    run_id?: string;
+    scope?: "personal" | "team" | string;
+    stage?: string;
+    delta?: string;
+    content?: string;
+    message?: string;
+    status?: number;
+    result_id?: number;
+    version_id?: number;
+    version?: number;
+    citations?: unknown[];
+    team_citations?: unknown[];
+    msg_count?: number;
+    total_msg_count?: number;
+    total_token_used?: number;
+    model_version?: string;
+    operation_type?: string;
+    operation_note?: string;
+    parent_result_id?: number | null;
+    generated_at?: string;
+}
+
+function buildSummaryURL(path: string): string {
+    return `${resolveSummaryBaseURL()}${BASE}${path}`;
+}
+
+function parseSSEBlock(block: string): SummaryStreamEvent | null {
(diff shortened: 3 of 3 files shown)


packages/dmworksummary/src/pages/SummaryDetailPage.tsx
```diff
@@ -859,31 +1051,60 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
...
+                    const previousPersonalResult = this.state.personalResult;
+                    const previousMembers = this.state.members;
+                    restoreRefineDraft = () => {
+                        if (this.taskId === requestTaskId) this.setState({ personalResult: previousPersonalResult, members: previousMembers });
+                    };
                     this.setState({ showRegenerateModal: false, refineLoadingTarget: "personal" });
-                    const refined = await api.refinePersonalSummary(requestTaskId, {
+                    let draft = "";
+                    let refined: api.SummaryStreamEvent | null = null;
+                    await api.streamRefinePersonalSummary(requestTaskId, {
```
OctoBoooot, line 1121: **[major]** The refine/regenerate streams are un-cancellable → setState-after-unmount + a network stream that outlives the page. The live streams create an AbortController (:862/:940), pass `signal` (:868/:946), and abort on unmount (componentWillUnmount:283 → :921/:993). `streamRefinePersonalSummary`/`streamRefineSummary` accept `signal?: AbortSignal` (summaryApi.ts), but these call sites pass their options object WITHOUT `signal`, and there is NO refine AbortController field — so componentWillUnmount aborts only the two live-stream controllers, never a refine stream. The `onEvent` guard `this.taskId !== requestTaskId` reads `this.props.taskId` and doesn't change on unmount, so the post-unmount setState (and the trailing `finally` setState({regenerateSubmitting:false})) still runs. Fix: add a refine AbortController, pass its signal at both call sites, and abort it in componentWillUnmount / on taskId change — the pattern the live streams already use. (Same finding as on the closed #792; persists here.)

packages/dmworksummary/src/pages/SummaryDetailPage.tsx
```diff
@@ -859,31 +1051,60 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
...
                         feedback: trimmed,
                         base_result_id: baseResultId,
                         base_version: this.state.personalResult?.version,
+                    }, {
+                        onEvent: (event) => {
+                            if (this.taskId !== requestTaskId) return;
+                            if (event.type === "delta") {
+                                draft = event.content || (draft + (event.delta || ""));
```
OctoBoooot, line 1130: **[question]** `draft = event.content || (draft + (event.delta || ""))` (also :1139) collapses snapshot vs delta into the single `type==="delta"` case — but the live handlers now explicitly separate `type==="delta"` (increment, :880/:953) from `type==="snapshot"` (full replace, :888/:961). If a `delta`-typed refine event ever carries a full `content` snapshot and then incremental deltas resume, this can duplicate/drop text, and the refine path commits `draft` optimistically so the corruption is visible until navigating away. On the happy path `content: refined!.content || draft` masks it. Please confirm: can a `delta`-typed refine event carry a full `content` snapshot? If yes, mirror the live handlers' explicit snapshot/delta split here.

OctoBoooot (changes requested):
## Review: render smart summary output incrementally (#794)

**Verdict:** Request changes — one real lifecycle bug (refine streams un-cancellable) carried over from the closed #792, plus a protocol-dependent question. This is the rebased successor to #792 (now conflict-free, +498/-40 since #698's refine/regenerate already merged), and the streaming core is verified clean — the two findings below are the same ones I raised on #792 and they persist unchanged.

Merge state is clean this time (`mergeable=MERGEABLE`, no conflict) and code CI is green — good improvement over #792. Risk: `size/XL` → `needs-human-review` applied.

### Major
- `packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1069, :1135` — **the refine/regenerate streams are un-cancellable → setState-after-unmount + a network stream that outlives the page.** The live streams create an `AbortController` (`:862`, `:940`), pass `signal` (`:868`, `:946`), and abort on unmount (`componentWillUnmount:283` → `:921`, `:993`). The refine helpers *accept* `signal?: AbortSignal` (`summaryApi.ts`), but these call sites pass their options object **without `signal`**, and there is **no refine AbortController field** — so `componentWillUnmount` aborts only the two live-stream controllers, never a refine stream. Its `onEvent` guard `this.taskId !== requestTaskId` reads `this.props.taskId`, which doesn't change on unmount, so the post-unmount `setState` (and the trailing `finally` `setState({regenerateSubmitting:false})`) still runs. Fix: add a refine `AbortController`, pass its `signal` at both call sites, and abort it in `componentWillUnmount` / on taskId change — exactly the pattern the live streams already use.

### Question (major if the backend does this)
- `SummaryDetailPage.tsx:1077, :1139` — the refine draft accumulator `draft = event.content || (draft + (event.delta || ""))` collapses snapshot vs delta into the single `type === "delta"` case. The **live** handlers now explicitly separate `type === "delta"` (increment, `:880`/:953) from `type === "snapshot"` (full replace, `:888`/:961) — but the refine handlers don't. If a `delta`-typed refine event ever carries a full `content` snapshot and then incremental deltas resume, this can duplicate/drop text; and the refine path commits `draft` optimistically, so the corruption is visible until navigating away (the `done` event's `content: refined!.content || draft` masks it on the happy path). Please confirm the refine stream's event contract — can a `delta`-typed event carry a full `content` snapshot? If yes, mirror the live handlers' explicit snapshot/delta split here.

### Verified clean (unchanged from my #792 review, re-confirmed at this head)
- **XSS / streamed render** — `CitationText` renders every tick through `ReactMarkdown` + `rehypeSanitize` (explicit schema: `defaultSchema` + only the custom citation tags). Partial/streaming markdown is sanitized each tick; no raw-HTML/script injection.
- **SSE parsing** — `parseSSEBlock` guards `JSON.parse` in try/catch, handles `[DONE]`, and `consumeSSE` buffers partial chunks correctly. Robust.
- **Live-stream ordering / cleanup** — new stream aborts the prior controller, `streamingTaskId` dedup, per-event `this.taskId !== taskId` stale-discard, in-order functional-`setState` append, `done`→`loadDetail()` reconcile, unmount aborts both live controllers + clears timers.
- **Auth/scope** — stream headers carry token + `X-Space-Id`, consistent with the non-streaming path.
- **nginx** — SSE proxy config (`proxy_buffering off` / `proxy_cache off` / `proxy_read_timeout 3600s`) is correct for unbuffered streaming; no auth/CORS regression.
- **Feature flag** — forward-to-matter gated behind a local `false` flag (real render gate).

### Praise
- The rebase onto merged #698 shrank this to +498/-40 (streaming layer only) and cleared the merge conflict — much cleaner surface than #792. And the live streams' abortable-controller pattern is exactly right; the fix here is just to extend that same discipline to the refine path.

### Net
The streaming core (render/parse/ordering/cleanup/proxy) is verified clean and this rebase resolved #792's merge conflict. The one code blocker remaining is the refine streams needing the same `AbortController` cancellation the live streams already have; plus the snapshot/delta protocol question. Wire the refine `signal` + confirm the stream contract and this is ready. `needs-human-review` applied for the XL size. Credit @Liuxy20 — the rebase and the clean streaming architecture are solid; the refine-cancellation gap is the one carryover to close.

## Parity checked
Diffed vs the current summary flow: this adds SSE streaming *alongside* the existing polling/refine (`-40` replaces old render wiring coherently; error/empty states retained; the non-streaming `refineSummary`/`regenerateSummary` from merged #698 are untouched — this layers `streamRefine*` on top). No existing capability removed; streaming is additive with polling preserved as the fallback and the persisted result as the post-completion source of truth.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

yujiawei (changes requested):
# Code Review — PR #794 (octo-web)

Reviewed at head SHA `3ba89c2f1f8d4b7e228e47665396545621ac3c5a` against merge-base `53f2f0c7` (3 files, +498/-40).

## Spec compliance

The PR delivers what issue #786 asks for: consume the backend SSE endpoints, render personal/team summary content incrementally, keep polling as a fallback and as the final-result refresh path, and preserve citation rendering once generation completes. Regenerate and refine flows are covered. Nothing required is missing.

One out-of-scope item is bundled in: hiding the "forward to matter" action behind a local `SHOW_FORWARD_TO_MATTER = false` flag (`SummaryDetailPage.tsx:56`, `:3127`). It is unrelated to incremental rendering. It's declared in the PR description and low-risk (the code path is retained), so it is not a blocker, but such toggles are better landed in a dedicated PR to keep the streaming change reviewable.

**Spec: OK (with the note above).**

## Findings

### P1 — Personal streaming state can get stuck `true`, leaving a stale/duplicate streaming box

`SummaryDetailPage.tsx:855-916`, `:507-511`, `renderStreamingContent()` `:1944-1952`.

`this.state.streaming` is only reset to `false` in three places inside `startSummaryStream`: the `done` event (`:896`), the `error` event (`:904`), and the `.catch` for a rejected/aborted request (`:913`). There are two realistic paths where none of those fire:

1. **Graceful stream close without a terminal event.** If the SSE connection ends normally (`reader.read()` returns `done`) but the last event was not `type: "done"`/`"error"`, `consumeSSE` resolves, `streamRequest` resolves, and the `.catch` never runs. `streaming` stays `true` and `streamingContent` keeps its draft. `startSummaryStream` also won't restart or clear it, because `streamAbortController`/`streamingTaskId` are only nulled in those same three handlers, so the early-return guard at `:861` blocks re-entry.
2. **Poll wins the race.** When `loadDetail` sees a terminal status it takes the `else` branch (`:507-511`) and calls `stopSummaryStream()` — which aborts, so the `.catch` returns early at `:909` without touching `streaming`. Note the branch resets `teamStreaming/teamStreamingContent/teamStreamError` (`:511`) but **not** the personal `streaming/streamingContent/streamError`. That asymmetry looks like the oversight.

Result: `renderStreamingContent()` (guarded only by `streaming && streamingContent.trim()`) keeps rendering the draft box next to the finalized `renderPersonalSummary()` output — a duplicated / stale summary that persists until the user navigates to another task. Please reset the personal streaming state symmetrically with the team state in the `else` branch, and reset it on graceful stream completion (treat "reader done" as an implicit close that clears `streaming`).

### P2 — Refine streams are started without an `AbortSignal`

`SummaryDetailPage.tsx:1069` (`streamRefinePersonalSummary`) and `:1135` (`streamRefineSummary`) pass only `onEvent`, no `signal`. Unlike `startSummaryStream`/`startTeamSummaryStream`, these refine streams are untracked and cannot be aborted on unmount or task switch. The `if (this.taskId !== requestTaskId) return` guards inside the callbacks prevent cross-task state corruption, so this is not a correctness blocker — but the underlying `fetch` keeps running (up to the new `proxy_read_timeout 3600s`) and can emit a "setState on unmounted component" warning. Recommend threading an `AbortController` through the refine calls the same way the generate streams do.

### P2 — Optimistic-restore closure can clobber concurrent background updates

`SummaryDetailPage.tsx:1058-1064` (and the team variant at `:1123-1128`). `restoreRefineDraft` closes over `previousMembers`/`previousPersonalResult` captured before the stream started. If any other update lands during the refine (e.g. a poll refresh of members), a later stream error reverts `members` to the pre-stream snapshot and silently discards those updates. Narrow window, but worth guarding — restore only the fields you optimistically mutated rather than the whole snapshot.

### P2 — Per-delta rebuild of the full `members` array during refine

`SummaryDetailPage.tsx:1071-1077`. On every SSE `delta` the refine path rebuilds `personalResult` and `.map()`s the entire `members` array to splice the growing draft string. This runs once per token and is markedly heavier than the generate path, which accumulates into a lightweight `streamingContent` string and merges at render time. Consider using the same lightweight-string approach for refine to avoid the reconciliation/GC churn on long outputs.

### P2 — Raw `fetch` streaming path bypasses the shared client's 401 handling

`summaryApi.ts:226-243`, `consumeSSE` `:201-224`. Using `fetch`/`ReadableStream` for SSE is reasonable (the axios instance can't stream in the browser), but two behaviors of the shared `summaryAxios` instance are now lost on this path: (a) the response interceptor that calls `WKApp.shared.logout()` on 401 (`:64-72`) — a 401 mid-stream just throws a generic error with no logout; and (b) resource cleanup. In `consumeSSE`, if `onEvent` throws (the refine `error` case at `:1096`/`:1153` does `throw`), the loop unwinds without `reader.releaseLock()`/`reader.cancel()`, so the body reader/connection isn't released promptly. Suggest handling 401 explicitly and wrapping the read loop in `try/finally` to cancel the reader.

### P2 — nginx streaming directives applied to the whole `/summary/api/v1` location

`nginx.conf.template:87-93`. `proxy_buffering off`, `proxy_cache off`, and `proxy_read_timeout 3600s` are applied to the entire summary API location, not just the `/stream` endpoints. This disables buffering for all normal JSON responses and lets slow clients hold upstream workers for up to an hour. Prefer scoping these to the SSE routes (a dedicated `location`/regex match) or have the upstream emit `X-Accel-Buffering: no` only on streaming responses, keeping the default timeout for everything else.

## Coverage notes / blind spots

- Backend behavior is not in this diff: whether the SSE endpoints always emit a terminal `done`/`error` event (directly relevant to the P1 above), how duplicate/overlapping stream requests are handled, and whether dropped SSE connections are terminated server-side — all unverified here.
- I did not run the monorepo build or exercise the streaming UI end-to-end; findings are from static review of the diff and surrounding file.
- The multi-collab render ordering swap at `:3254-3264` (team-vs-personal streaming boxes) was reviewed for logic but not visually verified.

## Verdict

Requesting changes for the P1 stuck-streaming-state bug. The P2 items are recommendations and not individually blocking, but the refine `AbortSignal` and the 401/reader-cleanup gaps are worth addressing in the same pass since they share the new streaming plumbing.

packages/dmworksummary/src/pages/SummaryDetailPage.tsx
```diff
@@ -466,8 +497,18 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
...
+                }
+                const isMultiByPerson = detail.summary_mode === SummaryMode.BY_PERSON && ((detail.participants?.length || 0) > 1);
+                if (isMultiByPerson && this.teamStreamClosedTaskId !== detail.task_id) {
+                    this.startTeamSummaryStream(detail.task_id);
+                }
             } else {
                 this.stopFallbackPoll();
+                this.stopSummaryStream();
+                this.stopTeamSummaryStream();
+                this.setState({ teamStreaming: false, teamStreamingContent: "", teamStreamError: null });
```
OctoBoooot, line 516: **[major]** This else-branch calls `stopSummaryStream()` (personal) but the following `setState` resets only the TEAM fields (`teamStreaming`/`teamStreamingContent`/`teamStreamError`). `stopSummaryStream` (:919) aborts + nulls `streamingTaskId` but does NOT reset the personal `streaming` flag or `streamingContent`. Since `renderStreamingContent` (:1946) gates on `if (!this.state.streaming || !content) return null`, the personal streaming card can stay visible with STALE content after this branch runs. Fix: reset the personal streaming state here too (or inside `stopSummaryStream`), mirroring `resetSummaryStreamForNewRun:930` which already does `setState({ streaming: false, streamingContent: "", streamError: null })`. The team path resets its state; the personal path doesn't — that asymmetry is the bug. Credit yujiawei.

OctoBoooot (changes requested):
## Supplement to my #794 CR — a second byte-verified major: personal streaming state not reset on stop (credit yujiawei)

yujiawei's CR raised a personal-stream state-reset bug I did NOT catch in my earlier review. I byte-verified it — it's real. My verdict stays REQUEST_CHANGES; adding this finding.

### Major — `stopSummaryStream` leaves personal `streaming`/`streamingContent` state stale → stale content stays visible
`SummaryDetailPage.tsx:919-925` — `stopSummaryStream()` aborts the controller and nulls `this.streamingTaskId`, but does **not** reset the React state (`streaming` flag, `streamingContent`). The render gate is `renderStreamingContent()` `:1946` → `if (!this.state.streaming || !content) return null`, so stale content stays visible whenever `streaming` is still `true`.

The reachable path is the poll `else`-branch at `:508-511`:
```js
} else {
    this.stopFallbackPoll();
    this.stopSummaryStream();          // personal: aborts, but leaves streaming/streamingContent state
    this.stopTeamSummaryStream();
    this.setState({ teamStreaming: false, teamStreamingContent: "", teamStreamError: null });  // resets TEAM only
}
```
The `setState` resets only the **team** fields — the **personal** `streaming`/`streamingContent`/`streamError` are never cleared here. So after this branch runs, `this.state.streaming` can still be `true` with a populated `streamingContent`, and the personal streaming card renders **stale content**. This is yujiawei's 🔴 (state not reset on stop → duplicate/stale content).

The fix pattern already exists in the same file — `resetSummaryStreamForNewRun:928-930` does `this.setState({ streaming: false, streamingContent: "", streamError: null })`. `stopSummaryStream` (or its `:509` caller) should reset the personal streaming state the same way the team path already does — the asymmetry (team state reset, personal state not) is the bug.

Severity: major — stale/duplicate personal summary content shown after a stop (self-heals on the next `startSummaryStream`/`done`/`loadDetail`, so display-consistency not data-loss), consistent with how I graded the refine-stream lifecycle major.

### The self-correction I own
My review traced the refine-stream lifecycle and the main-stream ordering/cleanup, but I didn't catch that `stopSummaryStream` resets the instance refs without resetting the React state, while its `:509` caller resets only the team state — so the personal path leaks stale streaming content. yujiawei caught it. Byte-verified real.

### Net (both findings now on my CR)
1. Refine/regenerate streams un-cancellable (`:1069`/`:1135` pass no `signal`, no refine controller) — my original major.
2. `stopSummaryStream` doesn't reset personal `streaming`/`streamingContent` state; the `:509` caller resets only team → stale personal content stays visible — this addition (credit yujiawei).
Plus the snapshot/delta protocol question. Verdict stays REQUEST_CHANGES @`3ba89c2f1f8d`; `needs-human-review` (size/XL) also applies. Fix both stream-lifecycle gaps (mirror the team-path/`resetSummaryStreamForNewRun` reset for personal; wire the refine AbortController) and this is close.

Jerry-Xin (dismissed):
**Verdict: REQUEST_CHANGES** — the incremental-render pipeline, XSS sink, nginx streaming config, and main/team stream cancellation are all sound, but the **refine/regenerate streaming path leaks its stream on unmount** (the same defect flagged on the earlier revision of this change). One blocking item.

## 🔴 Blocking

**Refine streams are never abortable and can setState after unmount** — `packages/dmworksummary/src/pages/SummaryDetailPage.tsx` (`handleRegenerate`, ~L1069 personal-refine, ~L1135 team/summary-refine).

Both refine call sites invoke `api.streamRefinePersonalSummary` / `api.streamRefineSummary` passing an options object with **only `onEvent`** — no `signal`. The API layer accepts `signal?: AbortSignal` (`packages/dmworksummary/src/api/summaryApi.ts`, `streamRefineSummary`/`streamRefinePersonalSummary`), but `handleRegenerate` never creates or stores an `AbortController` for the refine flow.

Consequently:
- `componentWillUnmount` → `clearAllTimers` (~L291) aborts only `streamAbortController` and `teamStreamAbortController` (the main personal/team streams). There is no refine controller to abort, so an in-flight refine `fetch`/reader keeps running after the page unmounts.
- The refine `onEvent` callbacks guard with `if (this.taskId !== requestTaskId) return;`. `this.taskId` is a getter over `this.props.taskId` (~L404), so it guards **task-switch** but not **unmount** — after unmount, props still hold the same id, the guard does not fire, and the `setState` calls inside the delta handler run on an unmounted component (React "setState on unmounted component" warning + wasted work).

Net: navigate away / close the detail page while a refine is streaming → the SSE connection and reader are not aborted (resource/connection leak) and post-unmount setState fires. The main and team streams already do this correctly (controller stored, aborted in `clearAllTimers`); the refine path needs the same treatment — create a refine `AbortController`, store it on the instance, pass `signal` to both `streamRefine*` calls, and abort it in `clearAllTimers`/`componentWillUnmount` (and on task switch / before starting a new refine).

## ✅ Verified safe

- **Incremental-render XSS sink** — streaming buffers render through `CitationText` → `ReactMarkdown` with `rehypePlugins={[[rehypeSanitize, citationSchema]]}` (`packages/dmworksummary/src/components/CitationText.tsx`). Every setState re-renders and re-runs `rehypeSanitize` over the accumulated buffer, so partial/incomplete markdown from mid-stream deltas is sanitized on every tick. No `innerHTML`/`dangerouslySetInnerHTML` on the streamed content. `renderStreamingContent`/`renderTeamStreamingContent` pass `citations={[]}`, so no citation injection either.
- **Main/team stream cancellation & dedup** — `startSummaryStream`/`startTeamSummaryStream` store a controller, call `stop*Stream()` (abort) before starting, and dedup via `streamingTaskId`/`streamClosedTaskId`. `clearAllTimers` aborts both on unmount and on prop-change (`componentDidUpdate`). Rapid re-open/tab-switch aborts the prior stream before opening a new one — no two interleaved main streams.
- **Chunk ordering** — single reader loop in `consumeSSE`; `delta` appends via functional `setState((prev) => prev + delta)` in receive order; `content`/`snapshot` replace with the full buffer; `done`→`loadDetail()` reconciles against the persisted result as source of truth. SSE is ordered over one connection, so append order is receive order.
- **Stream auth/scope** — `buildStreamHeaders` sets `token` (from `WKApp.loginInfo.token`) and `X-Space-Id` (from `currentSpaceId`), matching the non-stream path.
- **nginx** — SSE tuning (`proxy_http_version 1.1`, `proxy_buffering off`, `proxy_cache off`, `proxy_read_timeout 3600s`) added inside the existing `/summary/api/v1` proxy block that already forwards Host/X-Real-IP/X-Forwarded-* and rewrites the URI; client auth headers pass through transparently.
- **-40 deletions / regression** — the removed block is the old direct `refinePersonalSummary`/`refineSummary` await + result merge, replaced by the streaming refine with optimistic `draft` and a `restoreRefineDraft` rollback on error (`onEvent` throw on `error` propagates through `consumeSSE`→`streamRequest`→outer try/catch, which restores the pre-refine `personalResult`/`members`/`detail` and toasts). Final reconcile falls back to `draft` when `refined.content` is absent. The `renderPersonalSummary`/`renderTeamSummary` swap preserves editing, processing, member-status and participant-report paths; empty/error states preserved (`streamError`/`teamStreamError`, `renderFailed`). No functional regression found.

## 💬 Notes (non-blocking)

- `streamingContent` grows unbounded and `ReactMarkdown` re-parses the full buffer on every delta (~O(N²) over N chunks). Fine for summary-sized text; if summaries can get large, consider throttling the render on delta.
- `SHOW_FORWARD_TO_MATTER = false` correctly hides the forward-to-matter button while retaining the code/modal behind the flag — clean.

OctoBoooot (approved):
## Re-review (DELTA) @ `bf6df48b6989` — both majors + the snapshot/delta question all resolved → flipping REQUEST_CHANGES to APPROVE

Single commit since my CR head `3ba89c2f1f8d`, touching the same 3 files. All three items from my CR are fixed correctly, byte-verified including the exact edges each targeted.

### Major #1 (refine streams un-cancellable) — FIXED
- New `refineStreamAbortController` field (`:224`); each refine call creates a controller, stores it, and passes `signal: refineController.signal` (`:1114`, `:1190`).
- New `stopRefineStream()` helper aborts it (`:945-949`), and **`clearAllTimers()` now calls `stopSummaryStream()` + `stopTeamSummaryStream()` + `stopRefineStream()`** — so `componentWillUnmount` → `clearAllTimers()` aborts all three streams (the original hole: unmount didn't stop refine).
- Belt-and-suspenders: a `this.unmounted` flag (`:287`) guards every refine `onEvent` (`|| refineController.signal.aborted`) and the catch/finally (`:1323`, `:1332`), and the finally aborts the controller (`:1328-1330`). So even a mid-flight SSE event at unmount can't `setState` on an unmounted component, and the reader is aborted. Both the leak and the setState-after-unmount are closed.

### Major #2 (personal streaming state not reset on stop) — FIXED
`stopSummaryStream` now takes `resetState` (`:934`) and, when true, resets `streaming/streamingContent/streamError` (`:941`, guarded by `!this.unmounted`). The poll else-branch that previously reset only the team fields now calls **`stopSummaryStream(true)`** (`:515`) — so the personal streaming state is cleared on that path. The team-resets-but-personal-doesn't asymmetry is closed; stale personal content no longer lingers after a stop.

### Question (snapshot vs delta) — RESOLVED
The refine handler now **explicitly splits** the two event types, matching the live handlers: `type === "delta"` accumulates (`content || draft+delta`, `:1117`) and `type === "snapshot"` full-replaces (`draft = event.content || ""`, `:1126`). So a full-content snapshot arriving in the stream can't be mis-accumulated — my protocol concern is answered in code.

### No regression
The streaming core I verified clean before is intact (incremental-render XSS via `CitationText`+`rehypeSanitize`, SSE parse robustness, main/team stream ordering/cancellation, auth/scope headers). Code CI green (Build / secret-scan / osv / dependency-review / pr-title-lint); `mergeable=MERGEABLE` (no conflict).

### Net
Both majors fixed with the exact edges closed (unmount now aborts all three streams incl. refine; the `unmounted` flag guards setState; personal state reset on the stop path), and the snapshot/delta handling now explicitly split. Flipping **REQUEST_CHANGES → APPROVE @`bf6df48b6989`**, superseding my CR at the old head. `needs-human-review` (size/XL) still needs a human sign-off before merge. Credit @Liuxy20 — the fix is thorough (controller + unmount abort + `unmounted` guard + snapshot/delta split), not a minimal patch.

## Parity checked
Diffed my CR head (`3ba89c2f1f8d`) → new head (`bf6df48b6989`): the delta adds the refine AbortController + `stopRefineStream` (wired into `clearAllTimers`), the `unmounted` flag guards, the `resetState` param on `stopSummaryStream` (called with `true` from the poll else-branch), and the explicit `snapshot`/`delta` split in the refine handler. All additive/corrective; the streaming render, SSE parse, main/team stream handling, auth headers, and nginx config are unchanged. No capability removed; the deltas are exactly the three fixes.

Jerry-Xin (dismissed):
## ✅ Re-review @bf6df48b — refine-stream teardown leak is FIXED end-to-end

My prior REQUEST_CHANGES flagged the refine/regenerate SSE stream teardown gap: refine call sites passed no `signal`, there was no refine AbortController field, and unmount aborted no stream — so unmounting mid-refine caused setState-after-unmount **and** an in-flight reader/network that ran to natural completion. Both halves are now resolved. Verified byte-for-byte at the new head.

### Teardown chain (verified)
- `componentWillUnmount` → `clearAllTimers()` → `stopSummaryStream()` + `stopTeamSummaryStream()` + `stopRefineStream()`, and each stop method calls `.abort()` on its controller (`SummaryDetailPage.tsx` stopSummaryStream ~935, stopRefineStream ~946, stopTeamSummaryStream ~1027). So **all three in-flight controllers are aborted on unmount**.
- Refine call sites now pass `signal: refineController.signal` and register the controller in `refineStreamAbortController` (`SummaryDetailPage.tsx` ~1106/1114 personal, ~1186/1190 team).
- The signal reaches the network layer: `streamRefineSummary`/`streamRefinePersonalSummary` forward `signal: options.signal` into `streamRequest`'s `RequestInit` → `fetch(url, init)` (`summaryApi.ts` ~309/365, streamRequest ~244).
- On abort, `consumeSSE`'s `finally` runs with `completed=false`, calls `await reader.cancel()` then `reader.releaseLock()` (`summaryApi.ts` ~226-235) — reader is released and the underlying stream cancelled. **No reader/network leak survives unmount.**

### setState-after-unmount (guarded)
- Every setState inside the refine onEvent handlers is behind `if (this.taskId !== requestTaskId || refineController.signal.aborted) return;` (~1116 personal, ~1192 team). Since unmount aborts the controller, `signal.aborted` is true → all setState bail.
- Post-await and the catch/finally paths guard on `this.unmounted` / `signal.aborted` before any setState (~1144, ~1323, ~1332). A dedicated `finally` also re-aborts the refine controller as belt-and-suspenders.

### Secondary (snapshot/delta accumulator) — resolved
The refine handlers now handle both event shapes consistently: `delta` → `draft = event.content || (draft + (event.delta || ""))`, `snapshot` → `draft = event.content || ""` (~1117-1131 / ~1193-1213). This is robust: prefers a full `content` payload when present, else appends deltas — no snapshot/delta collapse.

### Regression sweep on the churn — clean
- Auth/scope intact: stream headers carry `token` + `X-Space-Id` from `WKApp` (`summaryApi.ts` buildStreamHeaders ~188-197).
- XSS sink unchanged: refine only updates `content` state rendered via the existing `SummaryEditor` component; no new `dangerouslySetInnerHTML`/rehypeRaw introduced.
- Live-stream signal+abort and chunk ordering guards intact.
- `nginx.conf.template`: adds SSE-correct proxying for `/summary/api/v1` (`proxy_buffering off`, `proxy_cache off`, `proxy_http_version 1.1`, `proxy_read_timeout 3600s`) — necessary and correct for unbuffered event streaming.

### CI / merge
Code checks green (Build, secret-scan, osv-scan, dependency-review all SUCCESS); governance checks SKIPPED (non-blocking). `mergeable=MERGEABLE` (rebase resolved the conflict); `mergeStateStatus=BLOCKED` is the pending-approval gate.

Clearing my prior CR. Nice, thorough fix.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #794 (octo-web)
**Head SHA:** `bf6df48b69890e2823d78ee07c04bd146b6cdba2`

## Summary

This PR adds SSE-based incremental streaming for smart summary output. Three files are changed: a new SSE client layer in `summaryApi.ts` (stream helpers, SSE parser, abort lifecycle), streaming state management and render paths in `SummaryDetailPage.tsx` (personal/team/refine streaming with UI updates as deltas arrive), and nginx proxy tuning in `nginx.conf.template` to enable unbuffered streaming. The implementation replaces blocking poll-then-render with progressive rendering while preserving the existing polling path as a fallback. Overall well-structured with thoughtful abort/cleanup lifecycle handling.

## Verification

- ✅ **Auth contract parity** — `buildStreamHeaders` replicates the same `token`, `X-Space-Id`, `Accept-Language` headers that `summaryAxios` interceptors set. 401 → `WKApp.shared.logout()` is preserved in `streamRequest`.
- ✅ **Abort/cleanup lifecycle** — `clearAllTimers` cancels all three stream controllers; `componentWillUnmount` sets `unmounted` flag; `componentDidUpdate` clears `streamClosedTaskId` on task change.
- ✅ **Fallback polling preserved** — Stream `.catch()` and clean-close `.then()` both call `startFallbackPoll()` as reliability fallback, matching the pre-existing polling contract.
- ✅ **Refine rollback** — `restoreRefineDraft` captures previous state before streaming refine starts; invoked in catch block to restore UI on failure.
- ✅ **SSE error propagation** — `throw` inside `onEvent` callback propagates through async `consumeSSE` → `await` in `streamRequest` → outer `.catch()`, correctly reaching the error handler.
- ✅ **Render tree consistency** — `renderStreamingContent()` guarded by `this.state.streaming` (false for non-BY_PERSON modes); team/personal render ordering is correct in all branches.

Static analysis only at head `bf6df48b`; build and tests not executed in this environment.

## Findings

### P2 — nginx SSE directives apply to all summary API endpoints (`nginx.conf.template:92`)

**Diff-scope:** New in this PR (new).

`proxy_buffering off`, `proxy_cache off`, and `proxy_read_timeout 3600s` are added at the `/summary/api/v1` location level, affecting every request through this block — not just the new `/stream` endpoints. Disabling buffering forces nginx to hold the upstream connection for the full response duration on every JSON API call (list, detail, delete, regenerate), increasing worker memory pressure and reducing throughput for non-streaming requests under load. The 3600s read timeout means a stalled non-streaming endpoint keeps the connection open for up to an hour.

Consider scoping SSE directives to streaming paths only via a nested `location` block:

```nginx
location ~ ^/summary/api/v1/summaries/.+/.*stream$ {
    proxy_http_version 1.1;
    proxy_buffering off;
    proxy_cache off;
    proxy_read_timeout 3600s;
    # ... same proxy_pass / headers as parent
}
```

Alternatively, have the upstream set `X-Accel-Buffering: no` on SSE responses and keep nginx buffering enabled by default.

### P2 — Team stream guard missing `members` fallback used by `isMultiPerson()` (`SummaryDetailPage.tsx:508`)

**Diff-scope:** New in this PR (new).

`isMultiPerson()` at `:1475` falls back to `members.length > 1` when `detail.participants` is empty or missing. The team-stream start condition at `:508` only checks `detail.participants?.length > 1` without this fallback:

```ts
const isMultiByPerson = detail.summary_mode === SummaryMode.BY_PERSON
    && ((detail.participants?.length || 0) > 1);
```

If the initial `loadDetail()` response has no `participants` array but `members` is populated (e.g. from a cached or partial response), the team stream won't start even though the render tree treats the page as multi-collab via `isMultiCollab()`. The stream will start on the next `componentDidUpdate` cycle after `participants` populates, but the UI can briefly show an empty team section.

Use `this.isMultiCollab()` for consistency with the render tree's multi-person determination.

### Nit — SSE `data:` line parsing strips all leading whitespace (`summaryApi.ts:175`)

**Diff-scope:** New in this PR (new).

`line.slice("data:".length).trimStart()` strips all leading whitespace, but the SSE spec only removes one optional space (or tab) after the field-name colon. If the server ever sends multi-line data payloads with intentional leading whitespace, the parser would silently corrupt them. Unlikely to affect JSON payloads in practice, but a spec-compliant fix is simple: strip at most one leading space or tab.

## Data-Flow Trace

|---|---|---|
| `streamingContent` (state) | SSE `delta`/`snapshot` events from `api.streamSummary` → `parseSSEBlock` → `onEvent` callback | ✅ Data flows correctly from server SSE → parse → setState |
| `teamStreamingContent` (state) | Same SSE path with `scope: "team"` | ✅ Independent stream, same data flow |
| `refined` (done event) | SSE `done` event captured in closure, checked non-null before use | ✅ `!refined` guard throws before accessing `refined!` |
| `draft` (closure var) | Accumulated from `event.content` (absolute) or `draft + event.delta` (incremental) | ✅ Consistent within each stream session |
| Auth headers in `fetch` | `WKApp.loginInfo.token`, `WKApp.shared.currentSpaceId`, `buildAcceptLanguage()` | ✅ Matches `summaryAxios` interceptor contract |

## Blind-Spot Checklist

- **C1 — Dual-path parity:** N/A. No paired add/remove or subscribe/unsubscribe paths changed.
- **C2 — Control-flow ordering:** Clear. Stream start/stop/reset methods are called from well-defined lifecycle points (`componentDidUpdate`, `clearAllTimers`, `handleRegenerate`). No nested re-entrant calls.
- **C3 — Authorization boundary:** Clear. New `fetch`-based stream path replicates the same `token`/`X-Space-Id` auth headers as the existing `summaryAxios` interceptor. 401 → logout preserved.
- **C4 — Auth lifecycle / container-member cascade:** N/A. No auth/permission changes.
- **C5 — Build/note through ≠ runtime correct:** N/A. No build/config packaging changes beyond nginx.
- **C6 — Governance/docs self-consistency:** N/A. No governance/docs changes.

## Verdict: APPROVED

No P0/P1 issues found. The two P2 items are non-blocking: nginx SSE directive scoping is a performance optimization that can be addressed post-merge, and the team stream guard inconsistency is mitigated by the existing polling fallback and self-corrects on the next `componentDidUpdate` cycle. The implementation is well-structured with proper abort lifecycle management, fallback polling, and state cleanup.

[Octo-Q] verdict: APPROVE — No P1 blockers. Two P2 items (nginx directive scope, team stream guard consistency) are non-blocking. Nit on SSE spec compliance.

yujiawei (changes requested):
# Code Review — PR #794 (octo-web)

Reviewed at head SHA `bf6df48b69890e2823d78ee07c04bd146b6cdba2` against merge-base `53f2f0c7` (3 files, +593/-43). Multi-dimension pass: SSE parsing, stream lifecycle/cancellation, refine data-flow, nginx/spec compliance, security & type-safety. Findings independently re-verified against the code.

## Spec compliance

The PR delivers what issue #786 asks for: it consumes the backend SSE endpoints (`/summaries/{id}/stream`, `refine/stream`, `personal-refine/stream`), renders personal/team summary content incrementally, keeps polling as both the interruption fallback and the final-result refresh path, and preserves citation rendering once generation completes. Regenerate and refine flows are covered. Nothing required is missing, and the previously-flagged lifecycle blockers from the earlier revision are resolved on this head:

- Refine streams are now cancellable — `clearAllTimers()` calls `stopRefineStream()`, and `handleRegenerate`'s `finally` aborts the active `refineStreamAbortController` (`SummaryDetailPage.tsx:295-298`, `:1327-1331`).
- Personal-stream state reset on stop is handled via `stopSummaryStream(resetState)` and the `unmounted` guard.

Two scope observations (non-blocking, but should be acknowledged in the PR description rather than left silent):

- **Scope creep — `SHOW_FORWARD_TO_MATTER = false`** (`SummaryDetailPage.tsx:55-56`, gate at `:3203`). This hardcoded flag unconditionally hides the "forward to matter" button that base rendered on every `COMPLETED` task. It is unrelated to SSE streaming and not mentioned in #786. If intentional, please note it in the PR body so reviewers/QA don't attribute the missing button to the streaming change.
- **Dead render branch for non-BY_PERSON** (`SummaryDetailPage.tsx:3351`). `renderStreamingContent()` is gated on `summary_mode !== BY_PERSON` here, but `state.streaming` is only ever set true inside `startSummaryStream()`, whose only callers are BY_PERSON-gated. So for BY_GROUP this branch is unreachable. Either wire up non-BY_PERSON initial streaming, or drop the dead branch to avoid misleading future maintainers.

## Code Quality

### 🔴 P1 (blocking)

**Refine `done` capture can be clobbered by the `[DONE]` sentinel, discarding server-authoritative `result_id`/`version`/`citations`.**
`SummaryDetailPage.tsx:1138-1140` (personal) and `:1222-1223` (team):

```ts
if (event.type === "done") {
    refined = event;
}
```

`parseSSEBlock` maps a trailing `data: [DONE]` line to a bare `{ type: "done" }` with no payload (`summaryApi.ts:179`). The client clearly expects `[DONE]` (it special-cases it), so the backend emits it. If the stream sends a payload-carrying `done` event (with `result_id`, `version`, `citations`, `generated_at`) **and then** a terminal `data: [DONE]`, the second assignment overwrites `refined` with the empty sentinel. The subsequent commit then falls back to `refined!.result_id || prev...` and `refined!.version || prev...`, so the persisted result keeps the **stale** id/version/citations while showing the new `draft` content — a state/DB divergence that the next refine (which sends `base_result_id`) can build on incorrectly.

This matters more than in the initial-generation path because the refine success path does **not** call `loadDetail()` to re-sync from the server (unlike `startSummaryStream`'s `done` handler at `:904`), so the stale metadata is not self-healing.

Fix: only capture a `done` event that actually carries a payload, e.g. `if (event.type === "done" && (event.result_id || event.content)) refined = event;`, or ignore the bare `[DONE]` sentinel when a payload `done` was already captured. Confirm the exact terminal-event contract against the octo-smart-summary refine handler.

### 🟡 P2 (non-blocking — recommend addressing)

1. **SSE parse fallback is unreachable for plain-text deltas** (`summaryApi.ts:183-185`). The `JSON.parse` catch returns `{ type: eventType, delta: data }` where `eventType` defaults to `"message"`; the consumer only reacts to `type === "delta"`. So a non-JSON frame without an explicit `event: delta` line is silently dropped. Harmless if the backend always sends JSON or always tags `event: delta`, but the fallback as written is effectively dead — worth a comment pinning the contract, or coercing the catch to `{ type: "delta", delta: data }`.

2. **`data:` line uses `trimStart()`** (`summaryApi.ts:174`) instead of the SSE-spec single-leading-space strip, and `rawLine.trimEnd()` (`:169`) strips trailing whitespace. Lossless for JSON payloads; lossy for whitespace-significant text in the fallback path.

3. **O(n²) buffer rescan** (`summaryApi.ts:212,218`). `buffer.match(/\r?\n\r?\n/)` re-scans from index 0 on every `read()`. Negligible for small token deltas, but a large `snapshot` frame arriving in many chunks is rescanned repeatedly — exactly the large-AI-output case this feature targets. Track a scan offset / use `indexOf`.

4. **`done`/`error` handlers null the abort controller without aborting the reader** (`:900-902`, `:993-995`). On a terminal event the code sets `streamAbortController = null` but doesn't `abort()`, so a later `stopSummaryStream()` can't force-close the (already-draining) connection. Minor; the stream is ending anyway.

5. **Optimistic refine draft not rolled back on abort** (`:1116-1133` mutate, `:1323-1324` catch early-returns on `AbortError`). On abort the optimistically-mutated `personalResult`/`detail` content is left in place, relying on an external reload to correct it. Acceptable if a task switch always triggers `loadDetail`, but worth confirming.

6. **nginx `proxy_read_timeout 3600s` applies to the whole `/summary/api/v1/` block** (`nginx.conf.template:89-93`), which also serves all non-streaming summary JSON calls. The location scoping is correct (main `/api/` is unaffected), but raising the timeout from the 60s default to 1h means a hung non-streaming summary request holds its connection for up to an hour on upstream faults. Consider a dedicated `location` for the `*/stream` suffixes.

7. **Type-safety** (`summaryApi.ts:137,149-150`). `type: ... | string` collapses the discriminant to `string`, removing exhaustiveness/typo protection on `event.type` checks; `citations`/`team_citations: unknown[]` are then cast `as any` at every consumer, erasing the `CitationItem[]` contract. Repo convention (CLAUDE.md) asks to avoid `any`. Prefer a closed union (`type: KnownType | (string & {})`) and typed citation arrays.

8. **Streaming error message not routed through `extractErrorMessage`** (`summaryApi.ts` `streamRequest`, ~`:239-259`). It builds the message inline as `data?.message || data?.msg || data?.error || ...`; `data.error` may be an object, yielding `[object Object]`. Read `data?.error?.message` to match the rest of the file.

### Verified clean

- **XSS**: streamed markdown flows through `CitationText`, which applies `rehypeSanitize` with a restrictive schema — no injection via streamed content.
- **Auth/base-URL parity**: `buildStreamHeaders` replicates `token` / `X-Space-Id` / `Accept-Language` from the same `WKApp` sources as the axios interceptor; `buildSummaryURL` reuses `resolveSummaryBaseURL()` (extension/electron origins honored). Raw `fetch` is justified — axios can't expose a `ReadableStream` body — and the module already runs its own axios instance.
- **`error`-event propagation & rollback**: throwing inside `onEvent` unwinds through `consumeSSE` (no try/catch around `onEvent`), the `finally` cancels the reader, and the rejection is caught in `handleRegenerate`, which invokes `restoreRefineDraft` for non-abort errors.
- **Unmount safety**: `componentWillUnmount → clearAllTimers` stops all three streams; `unmounted` guards `setState` in stop paths.

## Coverage / blind spots

- The P1 and finding #1 hinge on the **backend SSE wire contract** (octo-smart-summary), which is not in this diff: whether `refine/stream` emits a payload `done` **and** a trailing `[DONE]`, whether deltas are cumulative (`content`) vs incremental (`delta`), and CRLF vs LF framing. Please confirm against the emitter before merge.
- No unit tests were added for the SSE parser (`parseSSEBlock`/`consumeSSE`) or the stream lifecycle. Given the framing edge cases above, a few parser tests (multi-line data, `[DONE]`, non-JSON fallback, split-across-chunks frames) would de-risk future changes.
- nginx behavior was reasoned from directives only, not a live streaming upstream.

## Verdict

**CHANGES_REQUESTED** — one blocking issue: the refine `done`-capture can be clobbered by the `[DONE]` sentinel, persisting stale `result_id`/`version`/`citations` with no `loadDetail` re-sync. The P2 items are recommendations, not blockers. Once the `done`-capture guard is added (and the backend terminal-event contract confirmed), this is in good shape — the earlier revision's lifecycle blockers are all resolved on this head.

Jerry-Xin (dismissed):
Correcting my own prior APPROVE on evidence. A peer (yujiawei) raised a P1 on an axis I did not verify in my re-review; I byte-verified it at this head and it is real, so I flip APPROVE → REQUEST_CHANGES and credit the finding. My earlier lifecycle findings (refine AbortController leak, personal-stream state-reset) remain byte-verified FIXED at this head — this is a new, separate data-integrity defect.

## 🔴 Blocking

**Refine `done` capture is clobbered by the trailing `[DONE]` sentinel → persists stale `result_id`/`version`/`citations` with new content (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1138-1140` personal, `:1222-1224` team).** Byte-verified chain at this head:
- `parseSSEBlock` maps a trailing `data: [DONE]` line to a bare `{ type: "done" }` with no payload (`packages/dmworksummary/src/api/summaryApi.ts:179`). The client special-cases `[DONE]`, so the backend emits it.
- Both refine `onEvent` handlers capture the done event **unconditionally**: `if (event.type === "done") { refined = event; }`. So if the stream sends a payload-carrying `done` (with `result_id`/`version`/`citations`/`generated_at`) **and then** the terminal `[DONE]`, the second assignment overwrites `refined` with the empty sentinel.
- The commit then does `id: refined!.result_id || prev.personalResult.id`, `version: refined!.version || prev...`, `citations: refined!.citations || prev...` (`:1153-1157`) but `content: refined!.content || draft`. With `refined` = empty sentinel, it persists the **stale** id/version/citations alongside the **new** draft content → a metadata/content divergence.
- Unlike the initial-generation path (`startSummaryStream` done → `loadDetail()` re-sync), the refine success path calls only `loadPersonalVersions` (`:1171`), **not `loadDetail()`**, so the stale metadata does not self-heal, and the next refine (which sends `base_result_id`) builds on a stale id.

Fix: capture only a payload-bearing `done` (e.g. `if (event.type === "done" && (event.result_id || event.content)) refined = event;`), or ignore the bare `[DONE]` when a payload `done` was already captured. Confirm the terminal-event contract against the octo-smart-summary refine emitter (payload `done` + trailing `[DONE]`?). (Credit: yujiawei raised this P1; byte-verified valid at this head.)

## 💬 Non-blocking (all yujiawei's, byte-plausible; endorse as fast-follows, not blockers)
- SSE non-JSON delta fallback `{ type: eventType }` with `eventType` defaulting to "message" is effectively dead for the consumer (which only reacts to `type === "delta"`) — pin the contract or coerce to `delta` (`summaryApi.ts:183-185`).
- O(n²) buffer rescan on each `read()` for large snapshot frames — track a scan offset (`summaryApi.ts:212,218`).
- `done`/`error` null the abort controller without `abort()` (`:900-902`, `:993-995`) — minor, stream ending anyway.
- Optimistic refine draft not rolled back on abort (relies on external reload) — confirm task-switch always triggers `loadDetail`.
- `proxy_read_timeout 3600s` covers the whole `/summary/api/v1/` block incl. non-streaming JSON — consider a dedicated `*/stream` location (`nginx.conf.template:89-93`).
- `type: ... | string` collapses the discriminant + `citations: unknown[]` cast `as any` — prefer a closed union + typed citations.
- Two scope items to declare in the PR body, not leave silent: `SHOW_FORWARD_TO_MATTER = false` hides a base-rendered button unrelated to SSE (`:55-56`, gate `:3203`); dead non-BY_PERSON `renderStreamingContent` branch (`:3351`).

## ✅ Verified sound (unchanged at this head)
- Refine streams cancellable: `clearAllTimers()` → `stopRefineStream()`, `handleRegenerate` finally aborts `refineStreamAbortController`; unmount aborts all three controllers; `consumeSSE` finally cancels+releases the reader. (My original leak 🔴 — fixed.)
- Personal-stream state reset via `stopSummaryStream(resetState)` + `unmounted` guard. (yujiawei's earlier state-reset 🔴 — fixed.)
- XSS: streamed markdown via `CitationText` + `rehypeSanitize` restrictive schema.
- Auth/scope: `buildStreamHeaders` replicates token + X-Space-Id + Accept-Language; raw `fetch` justified for ReadableStream body.
- Snapshot/delta accumulator consistent between handlers.

## Verdict
REQUEST_CHANGES — one blocking data-integrity issue (the `[DONE]`-clobber persisting stale refine metadata without a `loadDetail` re-sync). The earlier lifecycle blockers are resolved on this head; the P2 items are recommendations. Add the `done`-capture guard + confirm the backend terminal-event contract → APPROVE.

packages/dmworksummary/src/pages/SummaryDetailPage.tsx
```diff
@@ -858,32 +1085,77 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
...
+                                    personalResult: prev.personalResult ? { ...prev.personalResult, content: draft } : prev.personalResult,
+                                    members: prev.members.map((m) => m.user_id === myUid ? { ...m, content: draft } : m),
+                                } as Pick<SummaryDetailPageState, "personalResult" | "members">));
+                                return;
+                            }
+                            if (event.type === "error") {
+                                throw new Error(event.message || t("summary.common.operationFailed"));
+                            }
+                            if (event.type === "done") {
+                                refined = event;
```
OctoBoooot, line 1151: **[major]** Empty `[DONE]` sentinel overwrites the real refine `done` event → result_id/version/citations lost. `parseSSEBlock` maps the terminal `data: [DONE]` to a metadata-less `{ type: "done" }`; a real backend done event is also `type: "done"`. This `if (event.type === "done") { refined = event }` overwrites unconditionally, so the common pattern of "metadata `done` then terminal `[DONE]`" clobbers the real event. The commit then does `refined!.result_id || prev.personalResult.id` / `version || prev.version` / `content || draft` / `citations || prev.citations` (:1153-1158) — with `refined` now empty, the refined content is committed under the STALE result_id/version with citations dropped = silent data corruption. Same on the team-refine path (:1223/:1241-1248). Fix: guard `if (event.type === "done" && (event.result_id || event.content)) refined = event`, or treat `[DONE]` purely as a stream terminator distinct from the backend `done` payload. Credit yujiawei.

OctoBoooot (changes requested):
## Flipping my #794 verdict APPROVE → REQUEST_CHANGES — a trailing `[DONE]` overwrites the refine result metadata (credit yujiawei)

yujiawei's CR at the current head caught a data-loss bug on the refine path that my APPROVE missed. Byte-verified — it's real. Flipping my verdict; I don't hold APPROVE over a confirmed major.

### Major — empty `[DONE]` sentinel overwrites the real refine `done` event → result_id/version/citations lost
`SummaryDetailPage.tsx:1139` (and `:1223` for the team-refine path):
```js
if (event.type === "done") { refined = event }   // unconditional overwrite
```
`parseSSEBlock` (`summaryApi.ts`) maps the terminal SSE sentinel to a **metadata-less** event: `if (data === "[DONE]") return { type: "done" }` — no `result_id`/`version`/`citations`/`content`. A *real* backend done event (JSON with the metadata) is **also** `type: "done"`. Since the handler overwrites `refined` on every done-typed event unconditionally, the common SSE pattern of "emit the metadata `done`, then a terminal `data: [DONE]` to close the stream" causes the empty `[DONE]` to **clobber** the real done event's fields.

The commit then does `refined!.result_id || prev.personalResult.id`, `refined!.version || prev...version`, `refined!.content || draft`, `refined!.citations || prev...citations` (`:1153-1158`, and `:1241-1248` for team). With `refined` now empty, every field falls back: the refined content is committed under the **stale/previous result_id and version, with citations dropped** — a silently corrupted refine result (new content saved against the old version id, references lost). That's data loss, not just a display glitch.

Fix: don't let a metadata-less done overwrite a populated one — e.g. guard `if (event.type === "done" && (event.result_id || event.content)) refined = event`, or treat `[DONE]` purely as a stream terminator (distinct from the backend's `done` payload event) so it never lands as `refined`.

### Also flagged by yujiawei (worth the author checking)
- 🔴 (dead code): the `BY_GROUP` branch of `renderStreamingContent` may be unreachable (streaming render is gated to BY_PERSON). Lower stakes than the data-loss bug — worth confirming and removing/fixing if dead, but not the decisive blocker.
- 🟡: the hidden forward-to-Matter feature flag isn't noted in the PR description (scope-hygiene).

### The self-correction I own
My APPROVE verified the AbortController fix, the personal-state reset, and the snapshot/delta split — but I did **not** trace the `[DONE]`-vs-`done` interaction, so I missed that a terminal sentinel overwrites the refine metadata. yujiawei caught it. Byte-verified real. This is the second data-flow edge on this PR family I under-checked (after the personal-state reset) — the refine commit path deserved a closer read.

### Net
Flipping APPROVE → REQUEST_CHANGES @`bf6df48b6989`: a trailing `[DONE]` sentinel overwrites the real refine `done` event, so the refined result commits with a stale result_id/version and dropped citations (`:1139`/`:1223` + the `refined!.X || prev.X` fallbacks). Guard the done-capture against the metadata-less sentinel. Credit yujiawei. The AbortController/state-reset/snapshot-delta fixes from the prior round remain correct; this is a separate data-flow bug on the done path. `needs-human-review` (size/XL) also applies.

OctoBoooot (approved):
## Re-review (DELTA) @ `79e7dd5f5925` — the `[DONE]`-clobber data-loss major is fixed → flipping REQUEST_CHANGES to APPROVE

One fix commit since my CR head `bf6df48b6989`, touching the same 3 files. My `[DONE]`-clobber major (three-way CR with yujiawei + Jerry-Xin) is fixed correctly at both refine call sites. Byte-verified.

### Major (`[DONE]` sentinel overwrites refine done metadata → data loss) — FIXED
A new predicate **`isRefineDonePayload(event)`** (`:952-960`) returns true only for a done event that actually carries payload:
```js
event.type === "done" && (event.result_id != null || event.version != null || event.version_id != null
  || event.content != null || event.citations != null || event.team_citations != null || event.generated_at != null)
```
Both refine capture sites now gate on it:
- **Personal refine** (`:1150`): `if (this.isRefineDonePayload(event)) { refined = event }`
- **Team refine** (`:1235`): same guard.
So the terminal `[DONE]` sentinel — which `parseSSEBlock` still maps to a metadata-less `{ type: "done" }` (`summaryApi.ts:179`) — now **fails the predicate and does not overwrite** the payload-bearing done event. The commit's `refined!.result_id || prev...` / `content || draft` / `citations || prev...` fallbacks therefore keep the *real* metadata; the refined result is persisted with its correct result_id/version/citations. The data-loss is closed. This is exactly the guard I prescribed (`type==="done" && (result_id || content || …)`), applied to both paths.

### No regression
The fixes from the prior rounds still hold (verified unchanged): refine AbortController + unmount teardown (`clearAllTimers` → `stop*Stream` aborts all three), the `unmounted` guards on `onEvent`/catch, the personal-stream state reset (`stopSummaryStream(true)`), and the explicit `delta`/`snapshot` split. Streaming render (CitationText + rehypeSanitize), SSE parse, main/team ordering, auth headers, nginx config all intact. Code CI green (Build / secret-scan / osv / dependency-review / pr-title-lint); `mergeable=MERGEABLE`.

### Non-blocking follow-up
- No unit test added for the `[DONE]`-after-payload-done guard (this streaming surface has had no unit tests across the PR — a pre-existing gap, not introduced here). The guard is straightforward and byte-verified, so non-blocking; but a test feeding "payload `done`, then `[DONE]`" and asserting `refined` keeps the payload would lock it against a future refactor that drops the predicate. Worth adding.

### Net
The `[DONE]`-clobber data-integrity major is fixed correctly at both refine sites via the `isRefineDonePayload` payload-guard, with the prior refine-leak / state-reset / snapshot-delta fixes intact. Flipping **REQUEST_CHANGES → APPROVE @`79e7dd5f5925`**, superseding my CR at the old head. Merge still gated on `needs-human-review` (size/XL) — a human sign-off, not a code issue. Credit @Liuxy20 — the payload-guard predicate applied to both refine paths is exactly the right fix, and factoring it into a named predicate (`isRefineDonePayload`) makes the intent explicit.

## Parity checked
Diffed my CR head (`bf6df48b6989`) → new head (`79e7dd5f5925`): the delta adds the `isRefineDonePayload` predicate and gates both refine done-captures (`:1150`/`:1235`) on it; the main/team stream handlers, refine AbortController/unmount teardown, personal-state reset, delta/snapshot split, SSE parse, render, and nginx config are unchanged. All additive/corrective; no capability removed — the sole behavioral delta is that a metadata-less `[DONE]` no longer overwrites a payload-bearing refine done.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #794 (octo-web)

## Summary

This PR adds SSE-based incremental rendering for smart summary output. It introduces streaming API helpers (`streamSummary`, `streamRefineSummary`, `streamRefinePersonalSummary`) that consume backend SSE endpoints via `fetch` + `ReadableStream`, and wires incremental rendering into the summary detail page with proper abort controllers, unmount guards, and polling fallback. The nginx config gains SSE-compatible proxy settings. A local feature flag hides the forward-to-matter action. Overall the implementation is solid — abort lifecycle, task-switch guards, and fallback paths are handled thoughtfully. Three P2 items and one nit below.

## Verification

Static analysis only at head `79e7dd5f`; build and tests not executed in this environment.

- ✅ **Abort lifecycle** — `streamAbortController`, `teamStreamAbortController`, and `refineStreamAbortController` are properly created, tracked, and aborted on stop/unmount/task-switch. `clearAllTimers()` calls all three stop methods.
- ✅ **Task-switch guards** — Streaming callbacks check `this.taskId !== taskId` before any `setState`. `streamClosedTaskId` / `teamStreamClosedTaskId` prevent re-streaming for already-completed tasks. `componentDidUpdate` resets streaming state and clears closed-task IDs on task change.
- ✅ **Unmount guard** — `this.unmounted` flag is set in `componentWillUnmount` and checked in `stopSummaryStream(true)` and the `finally` block of `handleRegenerateSubmit`.
- ✅ **Auth parity** — `buildStreamHeaders()` manually replicates the axios interceptor headers (`token`, `X-Space-Id`, `Accept-Language`). `streamRequest` handles 401 via `WKApp.shared.logout()`, matching the interceptor's response handler.
- ✅ **Polling fallback** — Stream errors and unexpected close trigger `startFallbackPoll()`, preserving the existing reliability path.
- ✅ **Refine draft restoration** — On refine stream failure, `restoreRefineDraft?.()` reverts the optimistic content update, preventing stale partial content from persisting.
- ✅ **SSE parser** — `parseSSEBlock` correctly handles multi-line `data:` fields, `[DONE]` markers, and falls back to raw text as delta when JSON parse fails.
- ✅ **Data flow: `isRefineDonePayload`** — Correctly requires at least one result field to be present before treating a `done` event as the terminal refine result, preventing premature completion on bare `[DONE]` markers.

## Findings

No P0/P1 issues. Three P2 items and one nit below.

### P2 — Stream error state captured but never rendered (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:113`)

`streamError` and `teamStreamError` are written on stream failure (lines ~912, ~928, ~1017, ~1033) but no render method reads them. When a stream fails, `streaming` flips to `false` so `renderStreamingContent()` returns `null`, and the streaming content silently disappears. The fallback polling and `loadDetail()` eventually restore data, but during the gap (potentially several seconds) the user sees no error feedback — the in-progress content just vanishes without explanation. Consider rendering a transient inline error or retry hint when `streamError` is non-null and `streaming` is `false`.

### P2 — Unguarded `setState` after unmount risk (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:516`)

`stopSummaryStream(true)` at line 515 correctly checks `this.unmounted` before calling `setState`, but the adjacent `setState({ teamStreaming: false, teamStreamingContent: "", teamStreamError: null })` on line 516 does not. If the component unmounts while `loadDetail`'s schedule callback is still pending (`scheduleLoadSeq` is not bumped by `componentWillUnmount`), this `setState` fires on an unmounted component. Add `if (!this.unmounted)` before this `setState`, matching the pattern already established in `stopSummaryStream`.

### P2 — Streaming directives apply to entire location block (`nginx.conf.template:89`)

`proxy_buffering off`, `proxy_cache off`, `proxy_http_version 1.1`, and `proxy_read_timeout 3600s` are applied to the entire `/summary/api/v1/` location, affecting all summary API calls — not just the SSE endpoints (`/stream`, `/refine/stream`, `/personal-refine/stream`). Unbuffered proxying increases upstream memory pressure for large JSON responses from non-streaming endpoints (list, detail, create). Consider scoping the streaming directives to a nested `location` block matching only the SSE paths, keeping the outer location's existing buffered config for regular API calls.

### Nit — Inconsistent `return` in error handler (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:912`)

The `done` event handler in `startSummaryStream` has a trailing `return;` after `loadDetail()`, but the `error` handler omits it. Functionally harmless (last `if` block in the callback), but adding `return;` matches the established pattern and prevents accidental fall-through if more event types are added later.

- `buildSummaryURL()` uses `resolveSummaryBaseURL()` + `BASE`, consistent with the axios `baseURL` interceptor — browser extension origins are handled correctly.
- `consumeSSE` reader cleanup: `finally` block cancels the reader and releases the lock on both success and error paths.
- `handleRegenerateSubmit` `finally` block: aborts the refine controller only if it's still the active one, preventing double-abort on success path where the controller was already nulled.
- `resetSummaryStreamForNewRun` / `resetTeamSummaryStreamForNewRun`: correctly clear `streamClosedTaskId` before restarting, allowing the new stream to proceed after regenerate.
- `SHOW_FORWARD_TO_MATTER` feature flag cleanly hides the matter-forward button without removing the code.
- SSE `event:` field fallback: `parseSSEBlock` uses `parsed.type || eventType`, correctly handling both JSON-embedded type and SSE `event:` field conventions.

## Verdict: APPROVED

No correctness, security, or build-blocking issues. The streaming implementation is well-structured with proper lifecycle management and graceful degradation to polling. The three P2 items (silent error state, unguarded setState, nginx scope) are non-blocking improvements worth addressing in a follow-up.

[Octo-Q] verdict: APPROVE — No P0/P1 blockers found. Streaming lifecycle management (abort controllers, task-switch guards, unmount guards, polling fallback) is handled correctly. Three P2 items (silent stream error state, unguarded setState on unmount, nginx streaming scope too broad) and one nit are non-blocking.

yujiawei (approved):
# Code Review — PR #794 (octo-web)

Reviewed at head SHA `79e7dd5f5925d699d65ba5b87d9f16f9229b0413` against merge-base `53f2f0c7` (3 files, +605/−43). Multi-dimension pass: SSE parsing, stream lifecycle, refine data-flow, render placement, nginx streaming config. This is a converged re-review — the prior blocker (a trailing `[DONE]` payload clobbering refine result metadata) is fixed at this head via `isRefineDonePayload`, which only latches `refined` on events that actually carry result fields.

## 1. Spec compliance

The PR delivers what the title states — incremental (SSE) rendering for smart summary:

- **New API surface** (`summaryApi.ts`): `streamSummary`, `streamRefineSummary`, `streamRefinePersonalSummary` plus SSE plumbing (`consumeSSE`, `parseSSEBlock`, `streamRequest`). 401 → `logout()`, non-OK → thrown error, matching the existing axios interceptor behavior.
- **Incremental UI** (`SummaryDetailPage.tsx`): personal + team streaming state, start/stop/reset lifecycle keyed on `taskId`, `streamClosedTaskId` guards against re-opening a finished stream, and a polling fallback preserved as the reliability path.
- **nginx** SSE-friendly directives on the summary proxy location.

No missing requirement and no divergence from the streaming approach. One **scope observation (non-blocking)**: the diff also introduces `SHOW_FORWARD_TO_MATTER = false` to hide the "forward to Matters" button (`SummaryDetailPage.tsx:55`, gate at `:3214`). This is unrelated to incremental rendering; it's a harmless dead-code toggle and clearly commented, but strictly it is a second, unrelated change riding along. Flagging for awareness only.

**Spec: ✅**

## 2. Code quality

Data-flow traced end-to-end; stream teardown verified across unmount, task-switch, and terminal events. No P0/P1 (correctness / data-loss / security / build-break) found at this head. The following are all **P2 / advisory** — none block merge.

**P2 — nginx directives applied to the whole `/summary/api/v1/` location, not just `/stream`** (`nginx.conf.template:90-93`). `proxy_buffering off`, `proxy_cache off`, and `proxy_read_timeout 3600s` now affect every summary JSON call, not only the SSE endpoints. For this low-traffic internal feature the blast radius is small, but unbuffered proxying of ordinary JSON ties up a worker per slow client, and a 1-hour read timeout on all requests can mask an upstream deadlock. Consider scoping to a `location ~ ^/summary/api/v1/.*/stream` block, or having the backend emit `X-Accel-Buffering: no` per-response so buffering stays on for normal calls.

**P2 — terminal SSE events don't proactively abort the reader** (`SummaryDetailPage.tsx:898-914`, team path `:1003-1019`). On `done`/`error` the handler nulls the controller and updates state but never calls `controller.abort()`. This relies on the backend closing the TCP connection promptly so `reader.read()` resolves `done`; if the server emits a terminal event but keeps the socket open, `consumeSSE`'s `while(true)` loop parks until the 3600s timeout. Low likelihood, but an explicit `controller.abort()` on terminal events would make teardown self-enforcing rather than server-dependent.

**P2 — `consumeSSE` re-scans the whole buffer with a regex each chunk** (`summaryApi.ts:212-218`). `buffer.match(/\r?\n\r?\n/)` over a growing accumulator is O(N²) in the worst case (large payload whose delimiter is flushed late). `buffer.indexOf("\n\n")` with a tracked scan offset avoids the quadratic. Micro-optimization; fine for expected summary sizes.

**P2 — SSE data-line trimming** (`summaryApi.ts:169,174`). `parseSSEBlock` applies `trimEnd()`/`trimStart()` to each data line. This is **safe for the primary path** — real content arrives as `data: {"delta":"…"}` and `JSON.parse` preserves inner-string whitespace exactly, and rendering goes through the same `CitationText` (ReactMarkdown + rehypeSanitize) as the final result, so XSS and markdown fidelity are intact. The trim only touches the non-JSON *fallback* branch (`return { type: eventType, delta: data }`), where markdown that relies on leading indentation or trailing-space hard breaks could be altered. Since that branch is only hit for malformed/plain payloads, impact is marginal — but per the SSE spec only a single optional leading space after `data:` should be stripped, so `trimStart()` is slightly over-eager if the backend ever falls back to raw text deltas.

**Quality: Approved**

## 3. Verdict

**APPROVED.** Spec ✅ and no P0/P1 findings at head `79e7dd5f`. The four items above are advisory improvements (nginx scoping is the most worth a follow-up); none is a merge blocker.

Jerry-Xin (approved):
## ✅ Re-review @79e7dd5f5925 — the `[DONE]`-clobber P1 is fixed at the root. APPROVE.

The single blocking data-integrity bug from the prior round is resolved cleanly, and nothing regressed.

### Blocking item — RESOLVED
The refine done-handlers no longer capture the metadata-less terminal sentinel. A new guard `isRefineDonePayload(event)` in `packages/dmworksummary/src/pages/SummaryDetailPage.tsx` returns true only when `event.type === "done"` **and** at least one metadata field is present (`result_id`, `version`, `version_id`, `content`, `citations`, `team_citations`, `generated_at`). Both refine done-handlers — personal (`SummaryDetailPage.tsx` ~:1150) and team (~:1234) — now gate capture on `this.isRefineDonePayload(event)` instead of the old unconditional `if (event.type === "done")`.

Since `packages/dmworksummary/src/api/summaryApi.ts` still maps a trailing `data: [DONE]` line to a bare `{ type: "done" }` with no payload (unchanged, and correctly so), that empty sentinel now fails the guard and can no longer overwrite a real metadata-bearing done. So the standard "payload-done then terminal `[DONE]`" SSE ordering keeps the real `result_id`/`version`/`citations`, and the commit path no longer persists stale metadata against new content. Both handlers are guarded — no partial fix.

### Happy path — verified intact
All guard fields are declared optional on `SummaryStreamEvent`, and the guard is an OR across them, so a legitimate single payload-done (carrying `result_id`/`content`/etc.) still passes and is captured. The change is strictly narrowing: it rejects only the truly empty sentinel. No regression to the single-done case.

### Non-blocking notes (carried, not gating)
- No unit test was added that would fail under the old unconditional-capture logic (e.g. emit payload-done + trailing `[DONE]`, assert the persisted `result_id`/`version`/`citations` are the payload's). Worth adding to lock the invariant, but not a merge blocker.
- Prior P2 observations remain as-is (SSE non-JSON delta fallback, buffer rescan cost, done/error controller nulling without `abort()`, optimistic-draft rollback on abort, nginx `proxy_read_timeout` scope, type discriminant). None are blocking.

The earlier-round lifecycle fixes (refine AbortController leak, personal-state reset via `stopSummaryStream(resetState)`, snapshot/delta split) were untouched by this commit and remain in place.

Nice, surgical fix. LGTM.

PR #804, opened 2026-07-15
fix(loop): role-gate project webhook and delete controls for non-admins

### LINKED ISSUE #802: Loop: project detail exposes owner/admin-only actions (webhook, delete) to regular members with only a raw "insufficient permissions" toast
## Summary

On the Loop **project detail** page, a regular workspace **member** (role below admin) is shown owner/admin‑only controls — the whole Webhook section and the project **Delete** button — with no role gating. The only feedback is a raw backend `insufficient permissions` toast when the action is attempted. Separately, the webhook‑list request fired on page open has no error handling, so its 403 becomes a silent unhandled promise rejection.

## When the `insufficient permissions` toast triggers

The backend gates certain workspace actions behind role `owner`/`admin` and returns **HTTP 403 `insufficient permissions`**. The Loop UI surfaces that backend message verbatim through its `.catch → Toast.error(e.message)` (the HTTP layer has no global interceptor; it just throws with `data.error`). On the project detail page it fires when a `member`‑role user performs an owner/admin‑only action:

- **Add a webhook** (the reported case), or enable/disable / delete / test a webhook
- **Delete the project**

It does **not** fire for saving the project name/description or creating a project — those have no role gate (any member is allowed).

Backend permission contract:

| Action | Required workspace role |
|---|---|
| Create / update project | any member |
| Delete project | owner / admin |
| Webhook list / create / update / delete / test | owner / admin |

## Inconsistencies (frontend vs backend)

1. **UI is not role‑gated.** `ProjectDetailPage` + `ProjectWebhooksSection` render the Delete button and the entire Webhook section (URL input + Add button, plus a `listWebhooks` call on mount that is itself owner/admin‑only) **unconditionally**. This page has no notion of the current user's workspace role, so a non‑admin member sees controls they can never use and only discovers the restriction through a raw backend error toast. This violates the interaction‑checklist (permission dimension) and frontend‑backend‑sync guidelines.
2. **`listWebhooks` has no `.catch`.** `ProjectWebhooksSection.reload()` does `listWebhooks(projectId).then(setRows).finally(...)` with no error handling. For a member the owner/admin‑only list request 403s as a **silent unhandled promise rejection** (the section then renders "no webhooks yet" as if empty), inconsistent with the add/toggle/delete paths in the same file, which all `catch` + toast.

## Proposed fix (frontend‑only)

The backend behavior is intentional and correct — webhooks can exfiltrate issue data, so restricting them to owner/admin is by design. **Do not change the backend.** Align the frontend:

1. Surface the current user's workspace role on the project detail page (reuse the existing current‑member resolution already used elsewhere in Loop to map the logged‑in user to their workspace member), and gate the Webhook section + project **Delete** button on `owner`/`admin` (hide or disable + show a short "requires admin" hint) instead of relying on a backend 403.
2. Add a `.catch` to the `listWebhooks` load: on 403 render an "admin only" placeholder consistent with the gating, and never leave an unhandled rejection.

## Reproduction

Loop workspace, signed in as a `member`‑role user:
1. Open a project → project detail → Webhook section.
2. Enter a URL and click **Add** → red toast `insufficient permissions`.
3. (Also reproduces via **Delete** project. Saving name/description works — no role gate.)

## Affected files

- `packages/dmloop/src/panel/ProjectDetailPage.tsx` — Delete button gating
- `packages/dmloop/src/panel/ProjectWebhooksSection.tsx` — Webhook section role gating + `listWebhooks` `.catch`

MrAladdin:
## What

Role-gates the Loop **project** surface so a non-admin workspace member no longer sees owner/admin-only controls that always 403.

Fixes #802.

## Why

Webhook CRUD and project delete are **owner/admin-only** server-side (they return HTTP 403 `insufficient permissions`); project create/update are open to any member. But the UI showed those admin-only controls to everyone:
- The project **detail** page rendered the Delete button + the whole Webhook section unconditionally.
- The project **list & card** views rendered a per-row Delete button unconditionally.
- On open, `ProjectWebhooksSection` fired `listWebhooks` (owner/admin-only) with **no `.catch`** → for a member it 403'd as a silent unhandled rejection that rendered as "no webhooks yet".

A member only learned the restriction from a raw backend error toast.

## Changes (frontend-only, `packages/dmloop`)

- **`ui/useWorkspaceAdmin.ts` (new)** — `useIsWorkspaceAdmin()` resolves the signed-in user's workspace role from the membership-gated roster (`listWorkspaceMembers`), returns `true` for `owner`/`admin`, `undefined` while loading, and **fails closed** (`false`) on error / unresolved identity. The `octo_uid === WKApp.loginInfo.uid` match mirrors the existing `myMemberId` resolution already used in `IssuePage`/`SettingsPage`.
- **`panel/ProjectDetailPage.tsx`** — Delete button renders only for admins; passes `isAdmin` to the webhook section.
- **`pages/ProjectPage.tsx`** — the list-row and card Delete buttons render only for admins (fixes the class across **every** `deleteProject` entrypoint, not just the detail panel).
- **`panel/ProjectWebhooksSection.tsx`** — gates on `isAdmin`: non-admins get an `adminOnly` placeholder and **no request is sent**; admins keep the full list/add/toggle/delete flow. `listWebhooks` failure now surfaces a retryable error placeholder instead of masquerading as empty (so an admin never creates a duplicate on an unknown state).
- i18n `loop.webhook.adminOnly` + `loop.webhook.loadFailed` (zh-CN + en-US).

The backend is intentionally correct (webhooks can exfiltrate issue data → owner/admin-only) and is **not changed**; the UI just stops offering actions it can't perform.

## Blast radius

- `useIsWorkspaceAdmin` is new; `ProjectWebhooksSection`'s new `isAdmin?` prop has a single caller (`ProjectDetailPage`).
- `deleteProject` entrypoints audited repo-wide — all three (detail panel, list row, card) are now gated; no other caller exists.
- Board / grouped / issue views, project create/update/save, and admin behavior are unaffected.

## Testing (dev, live)

- **Admin path (owner, live):** open a project → detail shows Delete + full Webhook section; list/card rows show Delete; `listWebhooks` succeeds; create/delete project works. No regression.
- **Member path (role forced to non-admin):** detail hides Delete, Webhook section shows the `adminOnly` placeholder, **0 `webhook-subscriptions` requests fire** (verified via `performance.getEntriesByType('resource')`), no `insufficient permissions` toast. Save name/description still works (correct — members may edit projects).
- Dev has no member-role account in the loop workspace (the owner is the only member), so the non-admin path was exercised by forcing the role signal; the admin path is fully live.

## Adversarial review

Two-model, two rounds (codex `adversarial-review` + Claude code-reviewer):
- Round 1 (both models) flagged that the project **list/card** delete buttons were still ungated — a sibling of the same class. Fixed here (gated all entrypoints).
- Round 2 codex flagged that matching on `octo_uid` could lock out an owner/admin whose member row has `octo_uid: null` (a "native member"). This is a false-positive in practice: the current user is always an octo-web login (`WKApp.loginInfo.uid`), and such members are octo-bridged with `octo_uid` set (verified on dev: the workspace owner's row carries `octo_uid`); a null-`octo_uid` "native member" is a backend-only artifact that can't be the current web user. `octo_uid` is the only client-side identity signal (there's no backend `user_id` for the current user client-side), and this is the same match the shipped `myMemberId`/`SettingsPage` code uses. Fail-closed is the safe default (never fail-open, which would re-expose the controls); the backend enforces regardless.

## Docs

No doc change: UI-only permission gating, no new command/config/API surface.

### DIFF
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -420,6 +420,8 @@
     "urlPlaceholder": "https://example.com/webhooks/octo",
     "add": "Add",
     "empty": "No webhooks yet.",
+    "adminOnly": "Only workspace admins can manage webhooks.",
+    "loadFailed": "Failed to load webhooks — click to retry.",
     "enabled": "Enabled",
     "delete": "Delete webhook",
     "deleteConfirm": "Delete this webhook?",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -420,6 +420,8 @@
     "urlPlaceholder": "https://example.com/webhooks/octo",
     "add": "添加",
     "empty": "还没有 webhook。",
+    "adminOnly": "仅工作区管理员可管理 Webhook。",
+    "loadFailed": "加载 webhook 失败，点击重试。",
     "enabled": "启用",
     "delete": "删除 webhook",
     "deleteConfirm": "删除这个 webhook？",
--- a/packages/dmloop/src/pages/ProjectPage.tsx
+++ b/packages/dmloop/src/pages/ProjectPage.tsx
@@ -12,6 +12,7 @@ import AssigneePicker from "../ui/AssigneePicker";
 import { confirmDelete } from "../ui/confirmDelete";
 import { formatRelativeTime } from "../ui/time";
 import { readView, writeView } from "../ui/viewMode";
+import { useIsWorkspaceAdmin } from "../ui/useWorkspaceAdmin";
 
 const { Title, Text } = Typography;
 
@@ -44,6 +45,7 @@ function Ring({ done, total }: { done: number; total: number }) {
 
 export default function ProjectPage() {
   const { t, format } = useI18n();
+  const isAdmin = useIsWorkspaceAdmin();
   const [rows, setRows] = useState<Project[]>([]);
   const [loading, setLoading] = useState(true);
   const [keyword, setKeyword] = useState("");
@@ -146,14 +148,16 @@ export default function ProjectPage() {
           <span className="loop-project-list__cell loop-project-list__time loop-project-list__cell--muted">
             {formatRelativeTime(p.created_at, format)}
           </span>
-          <Button
-            theme="borderless"
-            type="danger"
-            size="small"
-            className="loop-project-list__del"
-            icon={<Trash2 size={14} />}
-            onClick={(e) => { e.stopPropagation(); confirmRemove(p.id); }}
-          />
+          {isAdmin && (
+            <Button
+              theme="borderless"
+              type="danger"
+              size="small"
+              className="loop-project-list__del"
+              icon={<Trash2 size={14} />}
+              onClick={(e) => { e.stopPropagation(); confirmRemove(p.id); }}
+            />
+          )}
         </div>
       ))}
     </div>
@@ -174,14 +178,16 @@ export default function ProjectPage() {
             </div>
             <div className="loop-project-card__sub">{p.lead_name ?? t("loop.assignee.unassigned")}</div>
           </div>
-          <Button
-            theme="borderless"
-            type="danger"
-            size="small"
-            className="loop-project-card__del"
-            icon={<Trash2 size={14} />}
-            onClick={(e) => { e.stopPropagation(); confirmRemove(p.id); }}
-          />
+          {isAdmin && (
+            <Button
+              theme="borderless"
+              type="danger"
+              size="small"
+              className="loop-project-card__del"
+              icon={<Trash2 size={14} />}
+              onClick={(e) => { e.stopPropagation(); confirmRemove(p.id); }}
+            />
+          )}
         </div>
       ))}
     </div>
--- a/packages/dmloop/src/panel/ProjectDetailPage.tsx
+++ b/packages/dmloop/src/panel/ProjectDetailPage.tsx
@@ -6,6 +6,7 @@ import { useI18n, WKApp } from "@octo/base";
 import type { Project } from "../api/types";
 import { getProject, updateProject, deleteProject } from "../api/projectApi";
 import ProjectWebhooksSection from "./ProjectWebhooksSection";
+import { useIsWorkspaceAdmin } from "../ui/useWorkspaceAdmin";
 import "./sideDetail.css";
 
 const { Text } = Typography;
@@ -18,6 +19,7 @@ export default function ProjectDetailPage({ projectId, onChanged }: { projectId:
   const [title, setTitle] = useState("");
   const [desc, setDesc] = useState("");
   const [dirty, setDirty] = useState(false);
+  const isAdmin = useIsWorkspaceAdmin();
 
   const load = () => {
     setLoading(true);
@@ -67,7 +69,7 @@ export default function ProjectDetailPage({ projectId, onChanged }: { projectId:
         <Button icon={<ArrowLeft size={16} />} theme="borderless" onClick={back}>{t("loop.detail.back")}</Button>
         <Text type="tertiary" style={{ fontSize: 12 }}>{row.icon} {t("loop.detail.projectTitle")}</Text>
         <div style={{ flex: 1 }} />
-        <Button theme="borderless" type="danger" icon={<Trash2 size={14} />} onClick={remove}>{t("loop.action.delete")}</Button>
+        {isAdmin && <Button theme="borderless" type="danger" icon={<Trash2 size={14} />} onClick={remove}>{t("loop.action.delete")}</Button>}
         <LoopButton icon={<Save size={14} />} disabled={!dirty} onClick={save}>{t("loop.action.save")}</LoopButton>
       </div>
       <div className="loop-sd__body" style={{ gridTemplateColumns: "minmax(0, 1fr)" }}>
@@ -91,7 +93,7 @@ export default function ProjectDetailPage({ projectId, onChanged }: { projectId:
             </div>
             <div className="loop-fields__row">
               <div className="loop-fields__label">{t("loop.webhook.title")}</div>
-              <ProjectWebhooksSection projectId={row.id} />
+              <ProjectWebhooksSection projectId={row.id} isAdmin={isAdmin} />
             </div>
           </div>
         </section>
--- a/packages/dmloop/src/panel/ProjectWebhooksSection.tsx
+++ b/packages/dmloop/src/panel/ProjectWebhooksSection.tsx
@@ -9,20 +9,24 @@ import { confirmDelete } from "../ui/confirmDelete";
 
 const { Text } = Typography;
 
-/** 项目 Webhook 配置分区：列表 + 启用开关 + 删除 + 新增（创建后一次性展示 secret）。 */
-export default function ProjectWebhooksSection({ projectId }: { projectId: string }) {
+/** 项目 Webhook 配置分区：列表 + 启用开关 + 删除 + 新增（创建后一次性展示 secret）。
+ *  Webhook 端点限 owner/admin，故非管理员只展示占位、不发任何请求。 */
+export default function ProjectWebhooksSection({ projectId, isAdmin }: { projectId: string; isAdmin?: boolean }) {
   const { t } = useI18n();
   const [rows, setRows] = useState<WebhookSubscription[]>([]);
   const [loading, setLoading] = useState(true);
+  const [loadFailed, setLoadFailed] = useState(false);
   const [url, setUrl] = useState("");
   const [busy, setBusy] = useState(false);
   const [secret, setSecret] = useState<string | null>(null);
 
+  // 失败不静默当空态:否则一次瞬时故障会让管理员以为"没有 webhook"而重复创建。
   const reload = () => {
     setLoading(true);
-    listWebhooks(projectId).then(setRows).finally(() => setLoading(false));
+    setLoadFailed(false);
+    listWebhooks(projectId).then(setRows).catch(() => setLoadFailed(true)).finally(() => setLoading(false));
   };
-  useEffect(reload, [projectId]);
+  useEffect(() => { if (isAdmin) reload(); }, [projectId, isAdmin]);
 
   const add = async () => {
     const value = url.trim();
@@ -70,12 +74,21 @@ export default function ProjectWebhooksSection({ projectId }: { projectId: strin
     }
   };
 
+  if (isAdmin === undefined) {
+    return <div className="loop-wh"><div style={{ padding: "8px 0" }}><Spin size="small" /></div></div>;
+  }
+  if (!isAdmin) {
+    return <div className="loop-wh"><div className="loop-wh__empty">{t("loop.webhook.adminOnly")}</div></div>;
+  }
+
   return (
     <div className="loop-wh">
       <div className="loop-wh__hint">{t("loop.webhook.hint")}</div>
 
       {loading ? (
         <div style={{ padding: "8px 0" }}><Spin size="small" /></div>
+      ) : loadFailed ? (
+        <div className="loop-wh__empty" style={{ cursor: "pointer" }} onClick={reload}>{t("loop.webhook.loadFailed")}</div>
       ) : rows.length === 0 ? (
         <div className="loop-wh__empty">{t("loop.webhook.empty")}</div>
       ) : (
@@ -108,6 +121,7 @@ export default function ProjectWebhooksSection({ projectId }: { projectId: strin
         ))
       )}
 
+      {!loadFailed && (
       <div className="loop-wh__add">
         <Input
           value={url}
@@ -119,6 +133,7 @@ export default function ProjectWebhooksSection({ projectId }: { projectId: strin
           {t("loop.webhook.add")}
         </LoopButton>
       </div>
+      )}
 
       <Modal
         className="loop-modal"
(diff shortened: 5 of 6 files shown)


OctoBoooot (approved):
## Review: role-gate project webhook + delete controls for non-admins (#804)

**Verdict:** Approve — a correct authz-UX gate that fails closed over *intact server enforcement*, not a client-side security substitute. Code CI green.

The security question for any "role-gate the UI" PR is whether the client gate is being treated as the boundary (theater) or as UX over real server enforcement. This is the latter, and it's done right: the body is explicit that webhook CRUD + project delete are **owner/admin-only server-side (403), left unchanged** — the UI just stops offering actions that always fail. Verified all of it.

### Authz gate — fails closed on every path (verified)
`useIsWorkspaceAdmin()` (`ui/useWorkspaceAdmin.ts`):
- no `wsId`/`uid` → `false`; `.catch` → `false`; identity not found in roster (`me` undefined) → `me?.role === …` is `false`; only `owner`/`admin` → `true`. `undefined` while loading.
- The roster endpoint is **membership-gated, not role-gated** — every member can read it to learn *their own* role, so a member calling it isn't itself an escalation, and the real enforcement stays server-side. The `octo_uid === WKApp.loginInfo.uid` match mirrors the existing `myMemberId` pattern. Correct, minimal, fail-closed.

### Consumers gate correctly (no flash, no unnecessary 403)
- **Delete buttons** (`ProjectDetailPage` + both `ProjectPage` list-row and card): wrapped in `{isAdmin && (...)}`. Since `isAdmin` is truthy only when confirmed admin (`undefined` loading and `false` both hide), the button never flashes for a non-admin and is hidden while resolving. All three `deleteProject` entrypoints gated (matches the body's repo-wide audit).
- **`ProjectWebhooksSection`**: `useEffect(() => { if (isAdmin) reload(); }, …)` — non-admins (and the loading state) send **no `listWebhooks` request** (the call that 403s); `isAdmin === undefined` → Spin, `!isAdmin` → `adminOnly` placeholder. So a member never triggers the owner/admin-only endpoint at all.

### Bonus robustness fix (correct)
`listWebhooks` failure now sets `loadFailed` → a retryable placeholder (clickable to reload) instead of the old silent `.then(setRows)`-only path that rendered a failure as "no webhooks yet." The body's reasoning is right: masking a transient failure as empty could make an admin create a duplicate webhook on an unknown state. Add-form is also hidden on `loadFailed` so no action is taken against an unknown list.

### Praise
- Getting the model right — client gate as UX over unchanged server enforcement, explicitly stated ("backend is intentionally correct … not changed") — is exactly how to role-gate a UI without turning the client into the security boundary. And `undefined`-while-loading (hide, don't flash) + fail-closed-on-error is the correct default for an admin gate.
- The "don't mask a 403/failure as empty" fix is a sharp catch: it's the kind of silent-unhandled-rejection that reads as normal empty state and leads to duplicate-create. Surfacing it as retryable is right.
- Gating *every* `deleteProject` entrypoint (detail + list + card), not just the one in the reported panel, fixes the whole class rather than the single reported instance.

### Net
Correct authz-UX gate: fail-closed `useIsWorkspaceAdmin`, all delete + webhook controls hidden for non-admins with no flash and no always-403 request, server enforcement intact, plus a genuine robustness fix (failure ≠ empty). Code CI green, `size/M`, no governance label. No blockers, majors, or nits. Credit @MrAladdin — the "UX gate over unchanged backend enforcement," the fail-closed defaults, and the failure-not-empty fix are all exactly right.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #804 (octo-web)

## Summary

This PR adds client-side role-gating for project deletion and webhook management in the Loop module. A new `useIsWorkspaceAdmin` hook resolves the current user's workspace role from the member roster, and the result gates delete buttons in `ProjectPage` (list + card views), `ProjectDetailPage` (detail toolbar), and the entire `ProjectWebhooksSection` (which now short-circuits to an "admin only" placeholder for non-admins and skips the list-webhooks fetch entirely). Two new i18n keys (`adminOnly`, `loadFailed`) are added in both locales. The webhook section also gains proper error-state handling with a click-to-retry affordance.

Overall the change is clean, well-scoped, and consistent with existing codebase patterns.

## Verification

- ✅ **Dual-path parity (C1)** — all three project-delete entry points (list row `ProjectPage.tsx:155-160`, card `ProjectPage.tsx:185-190`, detail toolbar `ProjectDetailPage.tsx:75`) are gated with `{isAdmin && ...}`. Webhook operations are gated by the early return in `ProjectWebhooksSection.tsx:79-82`. No ungated sibling paths found.
- ✅ **Data-flow trace** — `useIsWorkspaceAdmin` calls `listWorkspaceMembers(wsId)` → `GET /workspaces/:id/members` → `WorkspaceMember[]` with `octo_uid` and `role` fields (verified in `types.ts:29-39` and `workspaceApi.ts:10-12`). The hook matches `m.octo_uid === WKApp.loginInfo.uid`, which is the canonical identity link in this codebase (see `directory.ts:44-56`). Role check is `=== "owner" || === "admin"`, matching the workspace role vocabulary.
- ✅ **Fails-closed** — on missing workspace ID, missing UID, API error, or unmounted effect, the hook resolves to `false`. Non-admins never see gated controls even during transient failures.
- ✅ **Effect cleanup** — the hook's `cancelled` flag prevents state updates after unmount. `ProjectWebhooksSection`'s `useEffect` correctly skips the fetch when `isAdmin` is falsy and re-fetches when `isAdmin` transitions to `true`.
- ✅ **i18n parity** — `adminOnly` and `loadFailed` added in both `en-US.json:423-424` and `zh-CN.json:423-424`.

## Findings

### P2 — Webhook list-fetch race on rapid project switching (`packages/dmloop/src/panel/ProjectWebhooksSection.tsx:26-29`)

`reload()` fires `listWebhooks(projectId)` without cancelling any in-flight request from a previous `projectId`. If a user switches projects quickly, the slower response can overwrite `rows` with the wrong project's webhooks. This is pre-existing (the old `useEffect(reload, [projectId])` had the same gap) — not introduced by this PR — but the new `loadFailed` state makes the surface slightly larger: a stale error from project A could set `loadFailed` while project B is rendering. An `AbortController` tied to the effect cleanup would close this. Not a blocker.

- `isAdmin === undefined` renders a spinner in `ProjectWebhooksSection` and hides delete buttons in `ProjectPage`/`ProjectDetailPage` — no flash of ungated controls during hook resolution.
- `loadFailed` hides the "add webhook" form (`{!loadFailed && ...}` at `:124`), preventing blind creation when the list is unknown.
- The `useIsWorkspaceAdmin` empty dependency array `[]` is intentional — workspace identity is stable per mount, and workspace switches cause a full remount.
- Non-admin users never trigger `listWebhooks` — the `if (isAdmin) reload()` guard at `:29` prevents unnecessary API calls.

## Verdict: APPROVED

Well-scoped, correctly implemented role-gating with proper fails-closed semantics. The pre-existing fetch-race P2 is noted but not blocking.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. All delete and webhook controls are correctly gated behind the admin check with fails-closed semantics. Dual-path parity verified across list, card, and detail views. The pre-existing fetch-race P2 is not introduced by this PR.

yujiawei (approved):
# Code Review — PR #804 (octo-web)

**Scope reviewed:** head SHA `9b42003566d7ea956b51a42b6bb40189139d0bd6` vs `main`. Frontend-only, `packages/dmloop` (6 files, +78/-22). Reviewed independently against linked issue #802, plus a second-model advisory pass.

## Verdict: Approved ✅

No blocking (P0/P1) issues. The change correctly implements the role-gating requested in #802, fails closed, and does not touch the backend (which is intentionally the enforcement boundary). A few non-blocking suggestions below.

---

## Spec compliance vs #802

#802 asked for three things; all are delivered, and the approach matches its guidance ("reuse the existing current-member resolution"):

- **Delete button gated** on `owner`/`admin` — `ProjectDetailPage.tsx:72`. ✅
- **Webhook section gated** — non-admins get an `adminOnly` placeholder and **no request is sent** (the `listWebhooks` call is now guarded by `useEffect(() => { if (isAdmin) reload(); }, ...)`, `ProjectWebhooksSection.tsx:32`). ✅
- **`listWebhooks` `.catch` added** — `ProjectWebhooksSection.tsx:27` now catches and renders a retryable `loadFailed` placeholder instead of a silent unhandled rejection masquerading as "empty". ✅

**Not scope creep:** extending the gate to the project **list/card** Delete buttons (`ProjectPage.tsx`) covers the same defect class — all three `deleteProject` entrypoints are now gated. I audited `deleteProject` / `confirmRemove` repo-wide and confirm no ungated caller remains. The `loop.webhook.adminOnly` / `loadFailed` i18n keys (zh-CN + en-US) are required by the new placeholders and are present in both locales.

## Correctness

- **Fail-closed is correct.** `useWorkspaceAdmin.ts` returns `false` on any error, on unresolved `wsId`/`uid`, and on no role match; `undefined` only while loading. It never fails open. Good — a transient roster failure hides admin controls rather than exposing them, and the backend enforces 403 regardless.
- **Identity match is consistent with the codebase.** `octo_uid === WKApp.loginInfo.uid` is the same resolution already used in `IssuePage` (`myMemberId`), `AgentPage`, and `SquadPage`. The null-`octo_uid` "native member" edge is not reachable for the current web user (who is always an octo login), and fail-closed covers it anyway.
- **Workspace-switch staleness is handled.** The `useEffect([])` in the hook only re-runs on remount, but `ProjectPage`/`ProjectDetailPage` are remounted on workspace switch via the React `key` in `LoopPage.renderTab` (`key = \`${key}:${ws?.id}\``), so the role is re-resolved for the new workspace. No stale-admin carryover.
- **`listWebhooks` ordering.** `ProjectWebhooksSection` is mounted fresh per detail push (`routeRight.push`) with a stable `projectId`, and `isAdmin` transitions `undefined → true` once, so `reload()` fires at most once per instance — no out-of-order overwrite in the current flow. (See P2 note if the section is ever reused under a changing `projectId`.)

---

- **[P2] Full roster fetch to derive one boolean.** `useIsWorkspaceAdmin` pulls the entire member list via `listWorkspaceMembers` on every mount of a gated page just to find the current user's role. For large workspaces this is a sizable payload for one bit of information. Note `api/directory.ts` **already** fetches `/workspaces/{id}/members` (cached), but selects only `user_id/name/octo_uid` — not `role`. Consider either adding `role` to the cached directory projection and deriving admin-ness from it, or a lightweight `/me/role` endpoint, to avoid a second uncached roster round-trip. Not blocking — this mirrors an existing pattern and is correct as written.

- **[P2] Layout shift when admin resolves.** While `isAdmin === undefined`, Delete buttons are fully hidden (not disabled), so on list/card views the trash icons pop in once the roster resolves. Minor flash. Optional: reserve the slot or render disabled-while-loading if the jump is noticeable in practice.

- **[P2] `loadFailed` hides the add form.** When the list load fails, the add form is hidden and recovery is only via clicking the retry placeholder. This is a defensible choice (avoids creating a duplicate against unknown state), just noting it as intentional-by-design for the author to confirm.

## Coverage notes

- **Reviewed:** all 6 changed files, the identity-resolution precedent in `IssuePage`/`AgentPage`/`SquadPage`, all `deleteProject` entrypoints, the `LoopPage` remount-key behavior, the webhook API surface, and the i18n symmetry.
- **Assumed correct (not in diff):** backend enforcement of owner/admin on `deleteProject` and webhook CRUD (403 `insufficient permissions`) — this is UI hardening on top of a backend gate that #802 states is intentional; UI hiding is not the security boundary and is not expected to be.
- A second-model advisory pass raised the roster-fetch cost and the layout-shift point (captured as P2 above) and a potential `listWebhooks` race (assessed as not reachable in the current single-mount flow).

Jerry-Xin (approved):
Role-gates project delete + webhook management behind an owner/admin predicate. Client-side gate is real (not just visual) and fails closed; approving. Backend remains the authoritative enforcer for these endpoints — see note.

## ✅ Highlights
- **Not a pure cosmetic hide.** For non-admins the webhook section returns early *before* any network effect, and the load effect is guarded: `useEffect(() => { if (isAdmin) reload(); }, ...)`. So a non-admin never fires `listWebhooks`/create/delete from this surface — the change is behavioral, not just `display:none`.
- **Fail-closed predicate.** `useWorkspaceAdmin.ts` returns `undefined` while resolving (renders a Spin placeholder, controls hidden), and `false` on missing workspace/uid or any roster error. Unknown/loading role is treated as non-admin. This is the correct default for an authz-adjacent gate.
- **Consistent coverage.** Both title claims are honored: project-delete is gated at all three render sites (list row, card, detail header) and the webhook section is gated + role-passed down from the detail page. I grepped every `deleteProject` / `confirmRemove` / webhook-API caller in the package; the only other `confirmRemove` lives in the Automation page and targets Autopilot (a different entity, out of scope). No ungated bypass path to project-delete or webhook-config was found.
- **UX correctness on load failure.** `loadFailed` state avoids masquerading a transient fetch error as an empty state, and suppresses the add-form on failure so an admin doesn't create a duplicate over a stale-empty view.

## 💬 Non-blocking notes
- **Backend enforcement is the real authority and is not verifiable from this repo.** The code comments assert the webhook endpoints and `deleteProject` are owner/admin-gated server-side; this diff cannot prove that. This client gate is correct defense-in-depth, but the security guarantee rests on the API server. Recommend confirming (in the API repo / a security test) that a non-admin token is rejected by the webhook and project-delete endpoints regardless of client. If server-side gating is ever absent, this PR would be cosmetic-only.
- **No test for the new gate.** Consider a unit/interaction test asserting: (a) non-admin does not render delete/add controls and issues no webhook request; (b) admin retains full controls; (c) unresolved role (`undefined`) hides controls (fail-closed). The package already has a `__tests__` convention to slot this into.
- **Predicate is a local ad-hoc check.** `role === "owner" || "admin"` is fine here, but if other admin-only surfaces adopt gating, extracting a single shared `canManageWorkspace(role)` helper would prevent drift.

## Regression
The -22 lines only unwrapped previously-unconditional renders into `{isAdmin && ...}`; admins keep full access and no control that should stay visible to non-admins is now hidden. Save button remains ungated (correct — editing project title/desc is not part of this restriction).

PR #781, opened 2026-07-15
feat(docs): Recently viewed / My documents tabs for the docs list (FEAT-B)

ploy-elison:
## Summary

FEAT-B frontend: split the docs landing list into two tabs — **Recently viewed** (default) and **My documents** — with per-tab filename search, a recent-only multi-select creator filter, server-side pagination with infinite scroll, and layered empty states. Implements the FEAT-B frontend design (XIN-1097 final) against the backend contract (XIN-1098, 5-API wire).

This is a **draft PR for line-by-line review only** — do not mark ready / merge. Promotion to a regular PR (after review + deploy + green tests + owner sign-off) is out of scope here.

## What changed

`packages/docs`:

- **`useDocsView` hook** (new) — one instance per tab, each owning its own `q` / creator filter / items / pagination / status, with a monotonic sequence guard that drops stale (out-of-order) responses. Per-tab search + filter are retained across tab switches and re-sent on return.
- **`DocsTabs`, `SearchBox`, `CreatorFilter` (+ chips), `InfiniteList`** (new) — presentational pieces reusing the existing `octo-docs-*` visual language; no control library introduced. `PortalMenu` extracted from `DocsHome` so the creator filter reuses it.
- **`DocsHome` / `DocsList`** — the old fetch-all single list is replaced by the tabbed, paginated container. Pin is scoped to *My documents* (the recent tab renders the server `viewed_at DESC` order verbatim). Opening a doc records a view (fire-and-forget). Layered empty states A–E, each with distinct title + CTA i18n keys (browse vs create are deliberately not merged).
- **`docsApi`** — adds `listRecentDocs` (keyset cursor), `listRecentCreators`, `recordDocView`; extends `listDocs` with `owner=me` / `q`; adds `viewedAt` to `DocListItem`.
- **i18n** — new `tab.*` / `search.*` / `filter.*` / `empty.*` keys + `list.viewedAt/loadingMore/end/loadMoreFailed/retry`, en + zh.
- **CSS** — new `octo-docs-tabs* / -toolbar / -search* / -filter* / -list-more* / -empty*` rules, consistent with the existing list styling.

## Contract alignment (XIN-1098 5-API)

| # | Endpoint | Frontend usage |
|---|---|---|
| 1 | `POST /docs/{docId}/view` | fire-and-forget ingest on doc open (no body; uid server-derived) |
[...]

PR #805, opened 2026-07-15
fix(docs): creator filter selection alignment + toolbar three-control single-row layout

ploy-elison:
## What

Two related docs-list front-end repairs on the FEAT-B document list, shipped on one branch:

1. **Creator filter selection alignment** (XIN-1208) — align the creator filter's selection/interaction with the type filter (type is the baseline), per boss feedback.
2. **Toolbar three-control single-row layout** (XIN-1209) — keep search + creator + type on one row and, when width is genuinely insufficient, wrap the two filters together instead of stranding Type on its own line. Implements the boss-confirmed design from XIN-1207.

---

### Inconsistency

Comparing `CreatorFilter`/`CreatorChips` against `TypeFilter`/`TypeChips` and the shared `useDocsView` engine:

- **Presentation / selection layer**: already identical (portal dropdown, immediate-effect checkbox rows, multi-select accumulate, per-item remove + clear-all chips, count badge). No change needed.
- **State engine** (`useDocsView.fetchFirst`): the recent tab re-fetched the creator facet candidates (`listRecentCreators`) on **every selection toggle** — a redundant round-trip that reloaded and reordered the open creator dropdown mid-multi-select. The type filter's candidate set is a fixed enum that never reloads during selection, so the two behaved differently.

### Fix

Gate the creator facet fetch on an actual `q` change (or a full reconcile: mount / space / folder / reload / retry). Selection toggles pass `refreshCreatorFacet=false`, so the candidate set stays stable during a multi-select session — matching the type filter. Preserved: creator filter remains recent-tab only; candidates still come from the server facet and track `q`; multi-select OR, AND-combined with search; the document list still refetches on every selection (only the *facet* refetch is decoupled).

### Root cause

The toolbar is a wrapping flex row. `.octo-docs-search` was `flex:1 1 160px; min-width:120px` (a hard 120px floor) and each filter wrapper was `flex:0 0 auto` (never shrinks). Once available width dropped below ~290px the flex engine wrapped the **last** child — Type — to its own line, producing the reported lopsided "search + creator / type-alone" state.

### Fix (in-place, no component-tree restructure)

- **Search yields first**: `.octo-docs-search` → `flex:1 1 140px; min-width:96px`, recovering ~24px so both filters stay on row 1 through the mid tier (the broken tier).
- **Filters wrap as a unit**: wrap `CreatorFilter` + `TypeFilter` in a single `.octo-docs-filter-group` (`display:flex; gap:8px; flex:0 0 auto`). When a wrap is unavoidable the whole group drops to row 2 together — a deliberate, symmetric "search on row 1, `[Creator][Type]` on row 2" — instead of Type alone.
- **Narrow-tier compaction**: a container query on the docs-list inline size tightens the filter buttons (`padding:4px 8px`) below 340px, buying one more width step before any wrap.

The my-docs tab (search + type only) keeps the group holding a single Type child and lays out normally. Selected-filter chips still occupy their own dedicated row on every tier, unchanged.

### Behaviour across width tiers (verified in-browser)

| Tier | Recent tab (search + creator + type) | My-docs tab (search + type) |
|------|--------------------------------------|-----------------------------|
| **Wide** (~520px) | all three on one row; search flexes to fill | one row |
| **Mid** (~400px) | search shrinks, **both filters stay on row 1** (was broken) | one row |
| **Narrow** (~300px) | search full-width on row 1, `[Creator][Type]` together on row 2 | still one row (only 2 controls) |

## Tests / build

- `useDocsView` suite green (30/30), incl. the test asserting selection toggles do **not** reload the facet while a real `q` change does.
- Docs `typecheck`: the only errors are pre-existing on `main` (`dmworkbase/i18n/format.ts`, `members/sort.ts`) and don't reference the changed files.
- Docs unit tests: identical pass/fail counts with and without this change (the `EditorShell.test.tsx`/`Toolbar.test.tsx` tiptap-jsdom failures are pre-existing on `main`, unrelated to docs-list layout).
- Layout verified in a headless browser at all three tiers on both tabs (position measurements + screenshot), not just build/typecheck.

## Related

- Design: XIN-1207 (boss-confirmed) · XIN-1208 · XIN-1209 · builds on merged FEAT-B PR #781

### DIFF
--- a/packages/docs/src/editor/styles.css
+++ b/packages/docs/src/editor/styles.css
@@ -2413,6 +2413,11 @@
   flex-direction: column;
   height: 100%;
   min-height: 0;
+  /* Establish an inline-size container so the toolbar can compact its filter
+     buttons based on the list's own width (it may render in a narrow panel,
+     not just the full viewport). Width is parent-driven, so this is layout-safe. */
+  container-type: inline-size;
+  container-name: octo-docs-list;
 }
 .octo-docs-list-header {
   display: flex;
@@ -2663,8 +2668,8 @@
   display: flex;
   align-items: center;
   gap: 6px;
-  flex: 1 1 160px;
-  min-width: 120px;
+  flex: 1 1 140px;
+  min-width: 96px;
   border: 1px solid var(--octo-border, #e5e6eb);
   border-radius: 6px;
   padding: 4px 8px;
@@ -2696,6 +2701,12 @@
   color: var(--octo-fg, #1f2329);
 }
 
+.octo-docs-filter-group {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+  flex: 0 0 auto;
+}
 .octo-docs-filter {
   position: relative;
   flex: 0 0 auto;
@@ -2718,6 +2729,14 @@
 .octo-docs-filter-btn-active {
   border-color: var(--octo-active-fg, #1664ff);
   color: var(--octo-active-fg, #1664ff);
+}
+/* Narrow tier: tighten the filter buttons so search + both filters hold one
+   more width step before the filter group wraps to a second row. */
+@container octo-docs-list (max-width: 340px) {
+  .octo-docs-filter-btn {
+    padding: 4px 8px;
+    gap: 3px;
+  }
 }
 .octo-docs-filter-caret {
   font-size: 9px;
--- a/packages/docs/src/pages/DocsHome.tsx
+++ b/packages/docs/src/pages/DocsHome.tsx
@@ -816,17 +816,22 @@ function DocsList({
       <DocsTabs active={activeView} onChange={onTab} />
       <div className="octo-docs-toolbar">
         <SearchBox value={view.q} onSearch={view.setQuery} onClear={view.clearQuery} />
-        {activeView === 'recent' && (
-          <CreatorFilter
-            options={recentView.creatorOptions}
-            selected={recentView.creators}
-            onToggle={recentView.toggleCreator}
-            nameFallback={nameFallback}
-          />
-        )}
-        {/* Type filter lives on BOTH tabs (creator is recent-only). Uses the active view's per-tab
-            types state so each tab remembers its own selection across switches. */}
-        <TypeFilter selected={view.types} onToggle={view.toggleType} />
+        {/* Creator + Type wrap together as one unit so a narrow toolbar drops the whole
+            group to a second row (search stays on row 1) instead of stranding Type alone.
+            On the my-docs tab the group holds only the Type filter and still lays out fine. */}
+        <div className="octo-docs-filter-group">
+          {activeView === 'recent' && (
+            <CreatorFilter
+              options={recentView.creatorOptions}
+              selected={recentView.creators}
+              onToggle={recentView.toggleCreator}
+              nameFallback={nameFallback}
+            />
+          )}
+          {/* Type filter lives on BOTH tabs (creator is recent-only). Uses the active view's per-tab
+              types state so each tab remembers its own selection across switches. */}
+          <TypeFilter selected={view.types} onToggle={view.toggleType} />
+        </div>
       </div>
       {activeView === 'recent' && (
         <CreatorChips
--- a/packages/docs/src/pages/useDocsView.test.ts
+++ b/packages/docs/src/pages/useDocsView.test.ts
@@ -150,3 +150,47 @@ describe('useDocsView — type filter (XIN-1188, multi-select OR, both tabs)', (
     await waitFor(() => expect(result.current.empty).toBe('E'))
   })
 })
+
+describe('useDocsView — creator facet candidates stay stable during selection (align with type filter)', () => {
+  it('does NOT reload the creator facet on a selection toggle (creator or type), only on a q change', async () => {
+    recentMock.mockResolvedValue({ total: 0, items: [], nextCursor: null })
+    creatorsMock.mockResolvedValue([{ uid: 'u1', name: 'Alice' }])
+    const { result } = renderHook(() => useDocsView('recent', 'space', 'folder', 0))
+    await waitFor(() => expect(result.current.phase).toBe('ready'))
+
+    // Initial load fetched the facet once (candidates track `q`, here empty).
+    await waitFor(() => expect(creatorsMock).toHaveBeenCalledTimes(1))
+
+    // Toggling a creator keeps `q`, so the open dropdown's candidate list must NOT churn — matching
+    // the type filter, whose candidate set is a fixed enum that never reloads mid-select.
+    await act(async () => {
+      result.current.toggleCreator('u1')
+    })
+    expect(creatorsMock).toHaveBeenCalledTimes(1)
+
+    // Toggling a type on the recent tab must not reload the creator facet either.
+    await act(async () => {
+      result.current.toggleType('doc')
+    })
+    expect(creatorsMock).toHaveBeenCalledTimes(1)
+
+    // Clearing selections keeps `q` → still no facet reload.
+    await act(async () => {
+      result.current.clearCreators()
+    })
+    await act(async () => {
+      result.current.clearTypes()
+    })
+    expect(creatorsMock).toHaveBeenCalledTimes(1)
+
+    // A real `q` change DOES refresh the candidate set (the preserved facet semantic).
+    await act(async () => {
+      result.current.setQuery('plan')
+    })
+    await waitFor(() => expect(creatorsMock).toHaveBeenCalledTimes(2))
+    expect(creatorsMock).toHaveBeenLastCalledWith('plan')
+
+    // Selections still refetch the DOCUMENT list (only the facet fetch is decoupled).
+    expect(recentMock).toHaveBeenLastCalledWith(expect.objectContaining({ q: 'plan' }))
+  })
+})
--- a/packages/docs/src/pages/useDocsView.ts
+++ b/packages/docs/src/pages/useDocsView.ts
@@ -140,7 +140,7 @@ export function useDocsView(
   const loadingMoreRef = useRef(false)
 
   const fetchFirst = useCallback(
-    (nextQ: string, nextCreators: string[], nextTypes: DocType[]) => {
+    (nextQ: string, nextCreators: string[], nextTypes: DocType[], refreshCreatorFacet = true) => {
       const seq = ++seqRef.current
       cursorRef.current = null
       pageRef.current = 1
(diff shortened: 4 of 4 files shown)


yujiawei (approved):
# Code Review — PR #805 (octo-web)

**Scope:** 2-file change in `packages/docs/src/pages/` — gate the recent-tab creator-facet refetch (`listRecentCreators`) on an actual `q` change / full reconcile, so selection toggles no longer churn the open creator dropdown, aligning it with the type filter's stable candidate set. Reviewed against head `b7a37d634e39ed25d897bdb606135c0240f6b704` (base `main`).

## 1. Spec compliance — ✅

The diff implements exactly the stated alignment and nothing more.

- **New param wired correctly at every call site.** `fetchFirst` gains `refreshCreatorFacet = true` (`useDocsView.ts:143`). All four selection paths pass `false` — `toggleCreator` (`:295`), `clearCreators` (`:302`), `toggleType` (`:309`), `clearTypes` (`:316`); all `q`/reconcile paths keep the default `true` — `setQuery` (`:277`), `clearQuery` (`:284`), `retry` (`:320`), `reload` (`:324`), mount/space/folder effect (`:334`). No caller was missed.
- **Only the facet fetch is gated.** The `if (refreshCreatorFacet)` guard wraps `listRecentCreators` only (`:182-188`); `listRecentDocs(...)` at `:189` stays unconditional, so the document list still refetches on every toggle.
- **All four "Preserved semantics" claims hold.** (1) Facet logic stays inside the untouched `if (kind === 'recent')` block; the `mine` branch never touches `creatorOptions`. (2) `listRecentCreators(q?)` (`docsApi.ts:184`) takes **only** `q` — it never depended on selected creators/types, so decoupling the refetch from selection loses zero candidate information. (3) Multi-select OR is intact — `listRecentDocs` still receives the full `creators`/`types` arrays on every toggle. (4) The document list still refetches on every selection.
- No over-build, no under-build, no scope creep — the diff is exactly 2 files.

## 2. Code quality — Approved (one P2 to consider)

### P2 — Creator dropdown can lag `q` by one step when a toggle interleaves an in-flight `q`-change facet fetch (`useDocsView.ts:182-188`)

The facet response is seq-guarded at `:185` (`if (seq === seqRef.current) setCreatorOptions(opts)`). Sequence:

1. `setQuery('plan')` → `fetchFirst(..., refreshCreatorFacet=true)` bumps `seqRef` to `N` and starts `listRecentCreators('plan')` in flight.
2. Before it resolves, the user clicks a creator/type checkbox → `fetchFirst(..., false)` bumps `seqRef` to `N+1` and issues **no** facet fetch.
3. The `seq=N` facet response for `'plan'` arrives and is dropped by the guard (`N !== N+1`). `creatorOptions` is left holding the **pre-`'plan'`** candidate set while the rows and selection already reflect `'plan'`.

**Severity rationale — this is a transient, self-healing UX inconsistency, not a correctness/data bug:**
- It is **not permanent.** Any subsequent `q` change (`setQuery`/`clearQuery`) or full reconcile — `reload` (tab switch, `DocsHome.tsx:461`), `reloadViews` (`:466`), space/folder change, `retry` — all pass `refreshCreatorFacet=true` and reinstall the correct facet.
- Impact is bounded: the **document list is always refreshed** on the toggle (fetch is outside the guard, `:189`), and row labels degrade gracefully through `creatorName`'s fallback chain (`CreatorFilter.tsx:16-20`: facet name → member-name map → raw uid), so nothing renders blank. The only visible effect is the open dropdown briefly missing/mismatching a candidate versus the current `q`.
- It does, however, remove a bit of incidental self-correction that existed before this PR (a toggle used to re-fire the facet and supersede the stale response), so it is worth an explicit note rather than silent acceptance.

**Optional mitigation** (do not block merge): key the applied facet to its `q` and refetch when `refreshCreatorFacet=false` but the last-applied facet `q` differs from the current `q`; or use a dedicated `creatorSeqRef` for the facet fetch so a selection toggle does not supersede an in-flight facet response for the same `q`. Either is a small follow-up.

### Nits (non-blocking)
- **Comment slightly overstates the guard** (`:173-174`): it says the facet refreshes "only when `q` actually changed," but the guard keys off the `refreshCreatorFacet` flag (caller intent), not a real `oldQ !== newQ` compare — re-submitting the same query string still refetches. This matches pre-PR behavior (no regression); it is only a comment-accuracy point.
- **Type-independence of the facet is implicit** (`:305-312`): `toggleType` passing `false` is correct only because the server facet is `q`-scoped and type-independent (`docsApi.ts:184` JSDoc). Sound today; a one-clause note in the comment would harden it against a future backend change that made the facet type-aware.

### Tests — valid, not false-green
The new test (`useDocsView.test.ts:154-196`) proves both halves of the claim: facet fetched once on mount, **not** re-fetched on `toggleCreator`/`toggleType`/`clearCreators`/`clearTypes` (all `count === 1`), and **is** re-fetched with the new arg on a real `setQuery('plan')` (`count === 2`, `toHaveBeenLastCalledWith('plan')`), while the document list still refetches. Assertions are non-trivial — `listRecentCreators` fires synchronously in the recent branch, so a regression that re-added the facet fetch on a toggle would fail the immediate count assertion. `act`/`waitFor` hygiene and mock isolation are correct.
- *Optional test hardening:* assert `creatorOptions` **value** stability across toggles (not just fetch count), and that a `q`-change response actually **replaces** the candidate set — the current sticky `mockResolvedValue` would not catch a dropped `q`-change response (the exact P2 above).

Spec ✅ and no P0/P1 blocker. The single P2 is a transient, self-healing dropdown-lag edge case with graceful degradation and a clear recovery path; it does not block merge. Recommend the author consider the P2 mitigation as a small follow-up.

OctoBoooot (approved):
## Review: align creator filter selection with the type filter (stable facet candidates) (#805)

**Verdict:** Approve
Now that this is out of draft (I held verdict-free while it was a draft), here's the formal gating review at `85a6097916`. The change is small, correct, and test-locked: it stops the creator dropdown from re-fetching + reordering its candidate list on every selection toggle, aligning it with the type filter's stable-candidate behavior. No blockers, no majors.

Code CI green (`ci_state=success`); gate already `reviewDecision=APPROVED` / `mergeStateStatus=CLEAN`. Risk tier: normal. Size/M — no `needs-human-review`.

## Parity checked
The facet-refresh change preserves every existing data-flow guarantee:
- **Facet still refreshes when it should** — `setQuery`, `clearQuery`, `retry`, `reload`, and the mount/space/folder/reloadToken effect all call `fetchFirst(...)` with the default `refreshCreatorFacet=true`. Only the four *selection* toggles (`toggleCreator`/`clearCreators`/`toggleType`/`clearTypes`) pass `false`. Since creator candidates track `q` only (§3.5, independent of selected creators/types/pagination), skipping the refetch when `q` is unchanged is semantically correct — the candidate set cannot change.
- **Concurrency guards untouched** — the `seq === seqRef.current` supersede check and the `loadingMoreRef` append-reentrancy guard are unchanged; the diff only adds the boolean param and conditionally gates the `listRecentCreators` call.
- **DocsHome `.octo-docs-filter-group`** — a pure layout wrapper so a narrow toolbar drops CreatorFilter+TypeFilter to a second row together instead of stranding Type; no logic or security surface.

### Praise
- `useDocsView.test.ts` — the new test locks the exact behavior: `listRecentCreators` is called once through a series of creator+type selection toggles, then a second time only after a `q` change. That's precisely the invariant the fix establishes, and it would fail if a future refactor reintroduced the per-selection refetch.

### Minor (non-blocking — yujiawei's 🟡, byte-verified)
- The dropdown could in principle show briefly-stale candidates on very fast switching. In practice this is bounded: candidates are a pure function of `q`, and a `q`-change refetch is superseded correctly by `seqRef`, so the only stale window is the transient during an in-flight `q`-change fetch — not something a selection toggle can worsen. Real but genuinely minor/cosmetic, consistent with the 🟡 grade. No fix required.

### Net
A precise, well-tested fix that removes a redundant refetch + dropdown reorder on selection, aligning the creator filter with the type filter's stable-candidate model — with the facet still refreshed on every `q` change / full reconcile and the seqRef/reentrancy guards intact. **APPROVE** @ `85a6097916`. Gate is already CLEAN; merge is the maintainer's action. Credit @ploy-elison for scoping the param to selection-only and locking it with a call-count test, and yujiawei for the stale-candidate axis.

Jerry-Xin (approved):
## Review: align creator filter selection with the type filter (stable facet candidates) (#805)

**Verdict: Approve** — @ head `85a609791653` (out of draft; I held verdict-free while it was a draft at `b7a37d634e39`, this is my fresh formal gating review at the ready head).

Byte-verified the full PR at this head (not just the last commit). The core fix and the new layout delta are both clean.

### Core change — `packages/docs/src/pages/useDocsView.ts` (+22/-12)
`fetchFirst` gains a `refreshCreatorFacet = true` param. The `recent` branch now only fires `listRecentCreators(nextQ)` when `refreshCreatorFacet` is set. All four selection toggles — `toggleCreator`, `clearCreators`, `toggleType`, `clearTypes` — pass `false`, so they keep the same `q` and no longer re-fetch + reorder the open creator dropdown's candidate list mid-select. Full reconciles (mount / space / folder / reload / retry) keep the default `true`. This aligns the creator filter's selection behaviour with the type filter, whose candidate set is a fixed enum that never reloads mid-select. Candidate `q`-tracking semantics (§3.5) and the server-facet source are unchanged — only the redundant refetch on selection is dropped. The `seq === seqRef.current` supersede guard around `setCreatorOptions` is preserved, and the document-list fetch still runs on every selection.

### New layout delta — `DocsHome.tsx` (+16/-11) + `editor/styles.css` (+21/-2)
Purely presentational, no behavioural coupling to the facet logic:
- `DocsHome.tsx`: wraps `CreatorFilter` + `TypeFilter` in a new `octo-docs-filter-group` div so a narrow toolbar drops the whole filter group to a second row instead of stranding Type alone. Identical props/handlers, same `activeView === 'recent'` gating for the creator filter, no new state, no new hook deps, no `dangerouslySetInnerHTML` / unescaped render → no state-management bug and no XSS surface.
- `styles.css`: establishes an `inline-size` container on the list, adds the `octo-docs-filter-group` flex wrapper, a narrow-tier `@container` rule, and tightens the search box flex-basis / min-width. Width is parent-driven, so it's layout-safe. No new `!important`.

### On the "stale creator data on fast-switching" concern
Non-blocking / effectively a non-issue. Because selection toggles now pass `refreshCreatorFacet=false`, selecting never refetches the facet, so selection can't surface stale creator options at all. On a genuine fast `q` change the candidate refetch does fire, but any superseded in-flight `listRecentCreators` resolve is dropped by the `seq === seqRef.current` guard before it can call `setCreatorOptions`. There is no path that renders *wrong* creators — at worst the previous candidate list persists momentarily until the newest `q` resolves (briefly-stale-then-correct), which matches the intended facet semantics.

### Tests / i18n
`useDocsView.test.ts` (+44) locks the contract precisely: the facet fetch stays at one call across creator toggle, type toggle, clear-creators and clear-types, then bumps to two only on a real `q` change (asserting `toHaveBeenLastCalledWith('plan')`), while the document list still refetches on selection. No new i18n keys are referenced in any patch — no dangling en-US / zh-CN entries.

### CI
Code CI is green at this head (newest leg per check): Build success, secret-scan (gitleaks) success, osv-scan success, dependency-review success, changed-paths detection success. Governance-only checks (check-sprint, satisfy-code-review) are skipped and don't bar merge.

✅ Logic sound, layout delta benign, test-locked, no regression, no security surface, no dangling. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

boris-clark (approved):
## Re-review — PR #805 (octo-web) @ `85a6097916`

Independent re-review, base `main`. Only reviewed — no changes, no merge.

**XIN-1208 (creator filter alignment) — ✅**
`fetchFirst` gains `refreshCreatorFacet = true` (`useDocsView.ts:143`). The four selection paths pass `false` (`toggleCreator:295`, `clearCreators:302`, `toggleType:309`, `clearTypes:316`); every q-change / full-reconcile path keeps the default `true` (`setQuery:277`, `clearQuery:284`, `retry:320`, `reload:324`, mount/space/folder effect `:334`). Only `listRecentCreators` is gated (`:182-188`); `listRecentDocs` (`:189`) stays unconditional, so the document list still refetches on every toggle. `listRecentCreators(q?)` is q-only (`docsApi.ts:184`) and candidates are defined pre-creator/pre-pagination, so skipping the refetch on a selection toggle is semantically lossless — the candidate set cannot change while `q` is fixed. Multi-select OR intact; seqRef + loadingMoreRef guards untouched.

**XIN-1209 (three-control single-row layout) — ✅**
Search `flex 1 1 140px / min-width 96px` yields width to the filters first; `.octo-docs-filter-group` (`flex 0 0 auto`) wraps CreatorFilter+TypeFilter as one flex item so a narrow toolbar drops the pair to row 2 together instead of stranding Type. Narrow-tier `@container octo-docs-list (max-width: 340px)` tightens the filter buttons. my-docs group holds only Type and lays out normally. Chips render outside the toolbar (independent row, unchanged). `container-type: inline-size` on `.octo-docs-list` is layout-safe — its width is parent-driven (split-left / list-only), and the filter dropdowns are positioned relative to `.octo-docs-filter`, not clipped by layout containment.

**No regression** — both changes additive; existing FEAT-B paths untouched. **Pre-existing errors** — PR touches only the 4 docs files; `EditorShell.test.tsx` and `members/sort.ts` are not touched by this PR, so their tsc/test errors are pre-existing on main, not introduced here. Build CI green. The new `useDocsView.test.ts` locks the invariant (facet called once across a series of toggles, twice only after a real `q` change) — not false-green.

**Verdict: APPROVE.** yujiawei's P2 (transient dropdown-lag on very fast switching) is self-healing, bounded, and non-blocking — a fine small follow-up.

PR #807, opened 2026-07-15
feat(docs): open in-chat doc links inline in a sidebar (WS-17)

pkuWMH:
## WS-17 文档链接点击在 Octo 内侧边栏内联编辑（不开新网页）

在 Octo 客户端内点击文档分享链接 `/d/<docId>?sp=<spaceId>` 时，直接在**聊天右侧侧边栏**内联打开该文档（预览 + 编辑 + 评论），不离开当前会话、不打开新网页；顶部提供「展开为整页」按钮再跳整页。交互取向、进入态、展开行为均按辉哥在 issue 内确认。

### 改动
- **拦截点击**（dmworkbase）：`MarkdownContent` 聊天正文链接 + 交互卡片 `openUrl` 识别同源 `/d/<docId>?sp=` → 侧边栏内联打开（仅当聊天侧边栏宿主已挂载）。普通左键才拦截，`cmd/ctrl/shift/alt`、中键仍照常新标签打开；非文档链接、跨域链接、聊天外场景一律回退现有整页/新页面行为。
- **复用面板外壳**：沿用文件预览侧边栏的 `wk-thread-panel` 右侧槽位，与子区/文件预览/事项/总结/搜索面板互斥。
- **在线协同编辑器**：内容由 docs 模块注册的新端点 `chatDocPreviewPane` 提供，渲染既有 `EditorShell`/`SheetView`/`BoardSession`（仿 `chatSummaryPanel` 解耦，base 不反依赖 docs）。按角色进入（writer/admin 可编辑、reader 只读），preflight 网关 forbidden（内联 RequestAccess）/ not-found / archived / 会话过期，不白屏。
- **展开为整页**：侧边栏内小按钮 → `window.open('/d/<docId>?sp=', '_blank')`（新标签页）。

### 不变
- 外部浏览器直接打开 `/d/` 链接、非聊天场景点击 → 现有整页 `StandaloneDocPage` 行为完全不变。
- 7/13 锁死的 shareUrl 格式契约未改动；未改 docs-backend。

### 验证
- 新增单测：`docLink.parseDocLink`、`tryOpenDocLinkInSidebar`、卡片 `openUrl` 文档链接路由 —— 全绿。
- 受影响既有测试全绿：`MarkdownContent`（24）、docs 包（1225）。docs 包 typecheck 干净（仅 2 处与本改动无关的既有报错）；`apps/web` typecheck 下本改动新增符号零报错。
[...]

PR #593, opened 2026-07-09
fix: align interactive card element manifest

an9xyz:
## Summary
- Align interactive card frontend allowlist with the server capability manifest for display elements.
- Allow RichTextBlock/TextRun, ImageSet, Table, and ActionSet while keeping unsupported elements fail-closed.
- Keep TextRun action validation under the same action whitelist and URL safety checks.

## Verification
- pnpm exec vitest run packages/dmworkbase/src/Messages/InteractiveCard/__tests__
- git diff --check origin/main...HEAD

PR #588, opened 2026-07-09
fix(export): preserve image aspect ratio in DOCX export

wwwzzzcccc:
## Summary

Images exported to DOCX were flattened/squished because the exporter fell back to a fixed 400x300 box for almost every image. This restores the true aspect ratio by reading the image's intrinsic dimensions from its encoded bytes.

## Linked Spec

**Goal:** DOCX-exported images keep their original aspect ratio instead of being forced into a fixed 400x300 box.

**Load-bearing:** `packages/docs/src/export/docx/images.ts` (`getImageDimensions`), consumed by `nodes.ts` `convertImage` on the export critical path.

**Out of scope:** the editor's ImageNode data model (intentionally width-only), PDF/Typst export, and the unrelated in-progress import work.

## Root cause

The editor's `ImageNode` persists only `width` (resize is width-only; the browser derives height via CSS `height:auto`), so the document model never carries a real `height`. The old `getImageDimensions` required **both** `width` and `height` to be finite positive numbers before using real dimensions; since `height` is always `null`, that branch never fired and every image fell back to a hardcoded 400x300 (4:3) box. Any non-4:3 image was stretched/flattened.

## What changed

- Add `readIntrinsicSize(buffer)` that parses real pixel dimensions from the encoded image header: PNG (IHDR), JPEG (walk marker segments to SOF), GIF (logical screen), BMP (BITMAPINFOHEADER, abs height for top-down).
- Rework `getImageDimensions(node, buffer)` to derive the aspect ratio from the intrinsic bytes (the buffer is already fetched for embedding), use the stored `width` attr as the target display width (or intrinsic width if absent), cap width at 600px, and compute height from the true ratio. The 400x300 default is only used when neither bytes nor attrs yield a real size.
- `convertImage` now passes the image buffer into `getImageDimensions`.

## COMPREHENSION

1. **What does this change actually do?** Before: exported images used a fixed 400x300 box (wrong aspect ratio for anything not 4:3). After: exported images keep their original aspect ratio, scaled to the stored display width and capped at 600px.
2. **What could break?** Only the DOCX image sizing path. Dependents: `convertImage` in `nodes.ts` (updated to pass the buffer). Failure mode is bounded: if the header can't be parsed and no valid attrs exist, it still falls back to the legacy 400x300 default (never zero/NaN dimensions).
[...]

PR #811, opened 2026-07-15
fix(workbase): align forward "recent" tab to the backend recent source (WS-20)

### LINKED ISSUE #813: bug: 消息转发选择群「最近」列表数据源与智能总结不一致（显示不对）
## 现象
聊天里的消息转发弹窗，选择转发目标群时，「最近」群列表看起来不对（成员 / 顺序与预期的"最近访问"不一致）。

对比：智能总结的选群弹窗（`ChatSelectorModal`）里的「最近」列表是对的。

## 根因
两处「最近」用了**不同数据源**：

- **智能总结（正确）** `packages/dmworksummary/src/components/ChatSelectorModal.tsx`
  走后端权威源 `SidebarService.sync({ tab: "recent" })`，用返回的 `recentIds` / `recentOrder` 过滤 + 按后端 `timestamp` 降序。
- **转发（bug）** `packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts`
  只拉了 `SidebarService.sync({ tab: "follow" })`，`recentKeys` 是从本地 `WKSDK conversationManager.conversations`（IM 会话列表）现推的，排序走本地 `sortConversations`（带置顶加权）。

本地 IM 会话列表 ≠ 后端"最近访问"集合：成员、顺序、置顶加权都会偏离，所以看起来不对。两处已共用同一套过滤纯函数 `packages/dmworkbase/src/Components/ChatSelector/tabFilter.ts`，只差数据源。

## 预期行为
转发的「最近」应与智能总结一致：走后端 `SidebarService.sync({ tab: "recent" })`，按后端 `timestamp` 降序。

## 修复
让 `useForwardModal` 改走后端 recent 源，用 `recentKeys` / `recentOrder` 与 `ChatSelectorModal` 对齐；共享 `tabFilter.ts` 不动，recent 排序放在 hook 侧，不影响其它调用方。零回归（follow / group / direct / 搜索 / 懒加载 / Space / grant 均不动）。

## 相关代码
- 转发弹窗：`packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts`
- 参照物：`packages/dmworksummary/src/components/ChatSelectorModal.tsx`
- 共享过滤：`packages/dmworkbase/src/Components/ChatSelector/tabFilter.ts`
- Sidebar recent/follow：`packages/dmworkbase/src/Service/SidebarService.ts`

## 关联
- 上游改造 issue #661（转发选群四 Tab 对齐智能总结）
- Octo loop 追踪：WS-20

pkuWMH:
Closes #813

### 问题
消息转发弹窗的「最近」群列表与智能总结选群弹窗的「最近」列表数据源不一致，导致转发的「最近」显示不对。

### 根因
两处「最近」用了不同数据源：
- **智能总结（正确）** `ChatSelectorModal.tsx`：走后端权威源 `SidebarService.sync({ tab: "recent" })`，用返回的 `recentIds`/`recentOrder` 过滤 + 按 `timestamp` 降序。
- **转发（bug）** `useForwardModal.ts`：只拉了 `follow`，`recentKeys` 从本地 `WKSDK conversationManager.conversations` 现推，排序走本地 `sortConversations`（带置顶加权）。

本地 IM 会话列表 ≠ 后端"最近访问"集合，成员/顺序/置顶加权都会偏离，所以看起来不对。

### 修复
让转发的「最近」也走后端 `SidebarService.sync({ tab: "recent" })`，与 `ChatSelectorModal` 完全对齐：
- 新增 recent 同步 `useEffect`（与现有 follow 同步同款：deviceId 空跳过、catch 退化空集、cancelled 守卫）。
- 用 `${target_type}::${target_id}` 复合 key 填 `recentKeys`，并记 `timestamp` 到新增的 `recentOrder`。
- 移除 `rebuildConvItems` 里的本地 `recentKeySet` 派生。
- 「最近」Tab 按后端 `timestamp` 降序（缺失视为 0），其余 Tab 保持原树序。

### 约束
- 共享的 `tabFilter.ts` **未触碰**；recent 排序放在 hook 侧，不影响其它调用方（智能总结自己在组件里排序）。
- 零回归：follow / group / direct / 搜索 / 懒加载 / Space 过滤 / grant 授权区均未改动。

### 测试
`packages/dmworkbase` 下 `useForwardModal.test.tsx`（35）+ `tabFilter.test.ts`（10）→ **45/45 通过**。
- 纠正了旧的「本地 recent」错误断言。
- 新增：① recent 由后端集合驱动、本地会话缺席后端 recent 时不出现；② recent 按后端 timestamp 降序；③ deviceId 空时退化空集、不回退本地。


packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts
```diff
@@ -432,6 +430,44 @@ export function useForwardModal(
...
+  // 必拒，跳过注定失败的请求，最近集合退化为空集（最近 Tab 显示空态）。
+  useEffect(() => {
+    const deviceUuid = WKApp.shared.deviceId || ""
+    if (deviceUuid === "") {
+      setRecentKeys(new Set())
+      setRecentOrder(new Map())
+      return
+    }
+    let cancelled = false
+    SidebarService.sync({ tab: "recent", device_uuid: deviceUuid })
```
OctoBoooot, line 448: **[blocker]** Not about this line specifically — anchoring the CI blocker to a live diff line. The PR title fails `pr-title-lint` (CI red → blocks merge). `WS-20: 转发选群「最近」列表对齐智能总结（改用后端 recent 源）` breaks the conventional-commit rule: it leads with a ticket id + CJK subject, but the rule wants `type(scope): <english subject>` with a lowercase package scope (cf. `fix(loop): …`, `feat(docs): …`). Suggest: `fix(workbase): align forward "recent" tab to the backend recent source (WS-20)`. This is the ONLY blocker — the code itself (this backend-recent-sync effect + key/timestamp handling) is byte-verified clean and matches ChatSelectorModal's proven scheme. Editing the title re-runs the check and flips the gate green.

OctoBoooot (changes requested):
## Review: WS-20 forward "recent" tab aligned to backend recent source (#811)

**Verdict:** Request changes
CI is red — **`pr-title-lint` is failing** (1 check). Per repo policy a red code CI check blocks merge, so this can't be approved until the title is fixed. That's the only blocker: I byte-verified the code and it's clean — the forward "recent" tab is correctly switched to the same backend recent source ChatSelectorModal uses, no correctness findings.

Code CI: `pr-title-lint` **failure**; Build / secret-scan / osv-scan / dependency-review all pass. (check-sprint / satisfy-code-review governance, SKIPPED.)
Risk tier: normal. size/M.

### Blockers
- **PR title fails `pr-title-lint`.** `WS-20: 转发选群「最近」列表对齐智能总结（改用后端 recent 源）` doesn't match the repo's conventional-commit rule — it leads with a ticket id and a CJK subject, but the rule wants `type(scope): <english subject>` (a lowercase package scope; see recent passing titles like `fix(loop): …`, `feat(docs): …`). Suggest e.g. `fix(workbase): align forward "recent" tab to the backend recent source (WS-20)` — type + package scope + English subject, ticket id in the description. Detail inline. (Same fix that unblocked #807.)

### Verified clean (no code findings)
- **Backend recent source correctly wired** — the new effect calls `SidebarService.sync({ tab: "recent", device_uuid })` and builds `recentKeys` + `recentOrder` from `resp.items` as `${target_type}::${target_id}` → `timestamp`. This is byte-for-byte the same source and key construction ChatSelectorModal uses (`packages/dmworksummary/.../ChatSelectorModal.tsx:74-94`), so the two "recent" lists are genuinely aligned per the WS-20 intent.
- **Key parity holds** — the lookup uses `FORWARD_ITEM_ACCESSORS.getKey = ${channelType}::${channelID}`, which matches the backend `${target_type}::${target_id}` (same enum; the group test asserts `target_type: 2` ↔ a group channel). No cross-type key collision.
- **Failure/empty handling correct** — empty deviceId → empty set (recent tab shows empty state, mirroring the existing follow-sync path); `.catch` → empty set; a `cancelled` flag guards against a late resolve after unmount. No leak, no stale write.
- **Scoped, no regression** — only the "recent" tab's data source + sort changed (`displayItems` sorts recent by backend timestamp desc; all other tabs keep the original `filtered` tree order). The removed local `recentKeySet` derivation is fully replaced.
- **Tests** — the new cases stub `SidebarService.sync` per-tab and assert recent scopes to the backend set (friend `d1` excluded, group `g1` included) and the timestamp sort — exercising exactly the new path.

### Net
A clean, well-tested alignment of the forward "recent" tab to the authoritative backend recent source, reusing ChatSelectorModal's proven source + key scheme with correct empty/cancel handling — no code findings. The single blocker is the red `pr-title-lint`: fix the PR title to conventional-commit form and CI goes green. **REQUEST_CHANGES** @ `7b757303` solely on the red CI; flip to APPROVE the moment the title lands green. Credit @pkuWMH for reusing the ChatSelectorModal recent scheme (single source of truth for "recent").

Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-web` and the implementation aligns ForwardModal’s recent-tab filtering and ordering with the backend sidebar recent source.

💬 Non-blocking
- 🟡 Warning: [useForwardModal.ts](/Users/mlamp/.openclaw/tmp/pr-review-4szgrtq0/repo/packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts:141) still documents the default recent tab as immediately populated from local recent conversations. That is stale after this PR, since recent is now empty until `SidebarService.sync({ tab: "recent" })` resolves, and remains empty when `deviceId` is absent.
- 🟡 Warning: [useForwardModal.test.tsx](/Users/mlamp/.openclaw/tmp/pr-review-4szgrtq0/repo/packages/dmworkbase/src/Components/ForwardModal/__tests__/useForwardModal.test.tsx:812) has the same stale suite-level description, saying the recent tab is derived locally and adds no backend dependency. The assertions below it now test the opposite behavior.

✅ Highlights
- The new recent key construction in [useForwardModal.ts](/Users/mlamp/.openclaw/tmp/pr-review-4szgrtq0/repo/packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts:452) matches `ChatSelectorModal`’s `target_type::target_id` scheme.
- The backend timestamp sort is scoped only to the recent tab at [useForwardModal.ts](/Users/mlamp/.openclaw/tmp/pr-review-4szgrtq0/repo/packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts:525), so group/direct/followed tree order is not affected.
- Added tests cover backend-driven membership, backend timestamp ordering, and the empty-device guard.

Verification caveat: I attempted the focused Vitest command, but this checkout has no installed dependencies, so `vitest` was not found.

Jerry-Xin (dismissed):
The PR is in scope for `Mininglamp-OSS/octo-web` and the implementation aligns ForwardModal's recent-tab filtering and ordering with the backend sidebar recent source.

💬 Non-blocking
- 🟡 Warning: `useForwardModal.ts:141` still documents the default recent tab as immediately populated from local recent conversations. That is stale after this PR, since recent is now empty until `SidebarService.sync({ tab: "recent" })` resolves, and remains empty when `deviceId` is absent.
- 🟡 Warning: `useForwardModal.test.tsx:812` has the same stale suite-level description, saying the recent tab is derived locally and adds no backend dependency. The assertions below it now test the opposite behavior.

✅ Highlights
- The new recent key construction in `useForwardModal.ts:452` matches `ChatSelectorModal`'s `target_type::target_id` scheme.
- The backend timestamp sort is scoped only to the recent tab at `useForwardModal.ts:525`, so group/direct/followed tree order is not affected.
- Added tests cover backend-driven membership, backend timestamp ordering, and the empty-device guard.

Verification caveat: I attempted the focused Vitest command, but this checkout has no installed dependencies, so `vitest` was not found.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #595 (octo-server)

Reviewed at head `38c117462c92474b52b698de987c9a650c5da357`. Scope covers the
`pr-base...HEAD` range as seen locally (includes commits #593 and #588 plus
this PR's HEAD commit `38c11746`); findings below target the cardaction
follow-up that is the stated scope of #595.

## Summary

This PR is the follow-up to #588 (generic card-action callback dispatch) and
extends the callback dispatcher along two axes: the URL validator accepts
plain `http://` alongside `https://` for operator-registered routes, and
`approval_card` gains an optional bounded `actions` slice (1..5) that lets a
caller render custom `Action.Submit` buttons on the generic approval card while
the server retains ownership of the action ID, reserved metadata, and card
layout. The dual-write `OCTO_CARD_ACTION_ALLOWED_URLS` env is deleted from
production code paths; `main.go` keeps a single structured WARN so rolling
upgrades surface a leftover.

The change is well-shaped and well-tested. The scheme relaxation is paired with
genuinely tighter URL-shape checks (raw `#` prefilter, `Hostname()` vs `Host`
to reject `http://:8080/x`, `ForceQuery` to catch trailing `?`), the
allowlist is correctly replaced by "route registration = operator
authorization", the nil-vs-non-nil-empty actions distinction is preserved
through the gin binder, and byte-for-byte compatibility with the legacy
approve/deny wire form is locked by a golden-string test. HMAC signing,
redirect refusal, and proxy-env ignoring all carry over intact to plain HTTP.
One P2 and one nit below — neither blocks.

## Verification

Static analysis only at head `38c117462c92474b52b698de987c9a650c5da357`; build
and tests not executed in this environment.

- ✅ **Scheme relaxation vs SSRF** — `validateCallbackURL` at `internal/cardactiondispatch/registry.go:327` now accepts `http`|`https` but still rejects credentials, fragments (raw prefilter at `:333` *and* `u.Fragment` at `:342`), queries (including `ForceQuery` at `:347`), opaque URLs, empty hostname via `u.Hostname() == ""`, and whitespace-padded input. `TestValidateCallbackURLRejectsUnsafeShapes` in `internal/cardactiondispatch/registry_http_scheme_test.go` covers file/gopher/relative/opaque/credentials/query/trailing-?/fragment/trailing-#/embedded-#/missing-host/port-only-no-host. The registered-routes-are-the-allowlist model is documented in the PR body and the code comments.
- ✅ **Redirects still refused on plain HTTP** — `NewHTTPDeliverer` at `internal/cardactiondispatch/http.go:56-58` sets `CheckRedirect = http.ErrUseLastResponse` regardless of scheme; the response mapper at `:102-104` classifies 3xx as `redirect_rejected` with `retryable: false`. `TestHTTPDelivererRefusesRedirectsOnPlainHTTP` locks this.
- ✅ **Proxy-env SSRF re-open still closed** — `base.Proxy = nil` at `internal/cardactiondispatch/http.go:53`; `TestHTTPDelivererDeliversPlainHTTPWithHMACAndDisciplinedTransport` sets `HTTP_PROXY`/`http_proxy` to a black hole and still delivers with verified HMAC, proving the production transport wins.
- ✅ **HMAC signing path is scheme-invariant** — `Sign` at `internal/cardactiondispatch/http.go:92` consumes `parsed.EscapedPath()` derived from `route.URL`; scheme does not enter the signature input, so plain-HTTP routes sign identically to HTTPS. The e2e test verifies `signatureOK` over cleartext.
- ✅ **Legacy wire-form byte-compat** — `TestBuildApprovalRequestCardOmittedActionsIsByteCompatible` in `pkg/cardtmpl/approval_request_actions_test.go:99` pins the exact pre-http-actions JSON; any ordering/key/escaping drift would fail. `buildApprovalActions` at `pkg/cardtmpl/approval_request.go:113` enters the legacy branch only on `input.Actions == nil`, which is the Go-level representation of JSON field absence.
- ✅ **nil vs non-nil-empty actions survives the binder** — `ApprovalCardFields.Actions` at `modules/notify/model.go:33` is `json:"actions,omitempty"`; `omitempty` only affects marshaling, so gin's unmarshaler preserves absent-field (`nil`) vs `"actions": []` (non-nil empty). `TestIntegration_ApprovalCardCustomActionsRenderInOrder` at `modules/notify/card_action_test.go:342-360` exercises both shapes against the live router and asserts `200 OK` for absent and `400 Bad Request` for explicit empty.
- ✅ **Custom-decision round-trip** — `TestCardActionCallbackOrchestrationForwardsCustomDecisionEndToEnd` at `internal/cardactiondispatch/orchestration_integration_test.go:372` pushes `decision:"execute"` end-to-end: HMAC verifies, `DecisionRequest.Decision` reaches the consumer unmodified (no server-side rewrite to approve/deny), `ActionID` is derived `approval-<decision>`, reserved `data.decision` mirrors the top-level token, and the standard finalizer renders the *consumer-picked* state rather than a decision-derived one.
- ✅ **Deprecation WARN path** — `main.go:477-482` reads `OCTO_CARD_ACTION_ALLOWED_URLS`, emits `log.Warn` with `zap.String("deprecated_env", …)` only when the value is non-empty after trim, and `NewRegistry` is then called with the two-arg `(specs, os.Getenv)` signature. `LoadAllowedURLs` is deleted from `internal/cardactiondispatch/config.go` and every call site. `TestDeprecatedAllowedURLsEnvIsIgnoredWithStructuredWarn` locks the shape.

## Findings

No P0/P1 issues. Two non-blocking items below.

### P2 — Custom-decision pattern does not reserve legacy `approve`/`deny` tokens (`pkg/cardtmpl/approval_request.go:143`)

`approvalDecisionPattern` (`^[a-z][a-z0-9_.-]{0,47}$`) accepts `"approve"`
and `"deny"` as valid custom decisions. The custom branch at `:153-170` then
emits Action.Submit buttons with `id: "approval-approve"` / `"approval-deny"`
— byte-identical to the legacy template IDs declared at `:17-18`. A caller
who accidentally ships `decision:"approve"` on a *custom* `ActionType`
(e.g. `task.execute.decision`) produces buttons whose IDs collide with the
legacy approve/deny IDs in any client-side analytics, replay log, or router
code that keys off action ID alone. The data map still carries the correct
`action_type` and `decision` so the consumer callback is unambiguous, but
ID-only readers silently drift.

Two reasonable fixes: (a) reject `decision ∈ {approve, deny}` with a clear
error (`"cardtmpl: approval action decision collides with legacy approve/deny
namespace"`), or (b) document the reserved namespace in the GoDoc on
`ApprovalRequestAction` and in `docs/card-action-callback-consumer.md` so
callers know to avoid those tokens. Not a blocker — this is caller-footgun
territory, not a reachable server defect — but the wire contract would
benefit from an explicit reservation.

### Nit — Deprecation test pins substrings rather than behavior (`main_carddispatch_test.go:175`)

`TestDeprecatedAllowedURLsEnvIsIgnoredWithStructuredWarn` locks the
deprecation decision by `strings.Contains` against `main.go` source: the
exact `log.Warn` message prefix, the `zap.String("deprecated_env", …)`
structured field, and the `cardactiondispatch.NewRegistry(specs, os.Getenv)`
call shape. That catches the obvious regressions, but rewrapping the Warn
message, renaming the field, or moving the call into a helper silently
disables the guard without any real assertion that the WARN actually fires
and the env is ignored at runtime. A follow-up could swap this for a
capture-logger assertion — call `installCardActionDispatch` with
`OCTO_CARD_ACTION_ALLOWED_URLS` set, assert one WARN record carrying
`deprecated_env=OCTO_CARD_ACTION_ALLOWED_URLS` and no registry error. Low
priority — the current shape does match the intent.

- URL validator: `strings.TrimSpace(raw) != raw` at `registry.go:327` rejects leading/trailing whitespace before `url.Parse`; combined with the raw `#` prefilter and `ForceQuery` check, the three documented footguns (trailing `#` that `url.Parse` silently drops, trailing `?` with empty query, `http://:8080/x` that `u.Host` alone would accept) are all closed.
- Custom actions bounded 1..5 at `pkg/cardtmpl/approval_request.go:137-141`; `MaxApprovalCustomActions` exported at `:20`.
- Control-character rejection runs on the *raw* title before `TrimSpace` strips `\t`/`\n`/`\r` at `:161-165`; `containsControl` at `:179` uses `unicode.IsControl`. Test matrix in `pkg/cardtmpl/approval_request_actions_test.go:153-165` covers `\x07`, leading tab, trailing `\n`, trailing `\r`, and embedded `\v`.
- Mixing custom `Actions` with `ApproveTitle`/`DenyTitle` is rejected at `:133`.
- Duplicate-decision check at `:147-150`; test at `:141-144`.
- `buildApprovalRequestCard` still marshals through `json.Marshal(document)` at `:106` — no hand-rolled JSON, no escaping risk.
- HMAC signing at `http.go:92` consumes `parsed.EscapedPath()`; path-not-empty fallback at `:89` defaults to `/`.
- `ApprovalCardFields.Actions` uses `json:"actions,omitempty"` at `model.go:33` so absent-field is `nil` on decode, and explicit `[]` is non-nil-empty — the distinction the validator relies on.
- `LoadAllowedURLs` removed from `config.go` and every test call site (`contract_test.go`, `dispatcher_test.go`, `service_test.go`, `coverage_test.go`, `card_action_test.go`, `api_card_action_test.go`, `orchestration_integration_test.go`).

## Data-flow trace

- `ApprovalCardFields.Actions` (`modules/notify/model.go:33`) → `buildApprovalRequestCard` in `modules/notify/approval_card.go:104` splits on nil: legacy branch at `:106-109` populates `ApproveTitle`/`DenyTitle` from `approvalActionLabelsFor(lang)`; custom branch at `:111-117` copies `[]ApprovalCardAction` into `[]cardtmpl.ApprovalRequestAction`. → `cardtmpl.BuildApprovalRequestCard` at `pkg/cardtmpl/approval_request.go:43` calls `buildApprovalActions` at `:91` which returns either the legacy two-button slice or the custom N-button slice, validated and with `owner`/`action_type`/`decision` injected into each button's data map. → `json.Marshal(document)` at `:106` produces the on-the-wire bytes. Custom `decision` reaches the dispatcher via the card's action data and is carried unchanged through `Enqueue` → Redis queue → `Dispatcher.ProcessOne` → `HTTPDeliverer.Deliver` → `DecisionRequest.Decision` in the signed callback body (locked by the e2e test).

## Blind-spot checklist

- **C1 (dual-path parity)** — `approve/deny` legacy path and custom-actions path both go through the same `BuildApprovalRequestCard` entry; `buildApprovalActions` is the single branch point and both sides validate, marshal, and return through the same envelope. ✅ parity holds.
- **C2 (control-flow ordering / nested reuse)** — `buildApprovalActions` is called exactly once from `BuildApprovalRequestCard`; no nested reuse that could double-inject `owner`/`action_type`. ✅ clear.
- **C3 (authorization boundary)** — scheme relaxation + allowlist removal changes what URLs the registry accepts. Authorization is now "route registration = operator authorization" (documented in code comments at `registry.go` and `http.go:51-53`). The `SenderUID` pattern check at `registry.go:282` (rejecting `iwh_*` prefixes) and per-route secret lookup are unchanged. ✅ no auth regression.
- **C4 (authorization lifecycle)** — unchanged; route specs come from operator-controlled `OCTO_CARD_ACTION_ROUTES` JSON, no new lifecycle path.
- **C5 (build ≠ runtime)** — the `http://` scheme relaxation and proxy-env SSRF guard are verified by an actual plain-HTTP test server with `HTTP_PROXY` set (`registry_http_scheme_test.go:97-153`); not a build-only claim. ✅
- **C6 (governance/docs self-consistency)** — docs updated in `docs/card-action-callback-consumer.md`, `docs/card-action-callback-dispatch.md`, `docs/card-protocol.md`. The `http://` relaxation is reflected; the allowlist removal is reflected; bounded custom actions are documented. No contradiction with SECURITY.md visible in the diff.

## Cross-round blocker re-check

N/A — first review pass on this PR.

## Verdict: COMMENT

Solid follow-up to #588. Scheme relaxation is paired with genuinely tighter
URL-shape defenses, the bounded custom-actions surface is well-bounded
(1..5, pattern-checked decision, raw-string control-char rejection, reserved
metadata injection), byte-compat with the legacy wire form is locked, and the
end-to-end custom-decision path is exercised. No P0/P1 issues. The P2
(consider reserving `approve`/`deny` in the custom-decision pattern) and the
nit (deprecation-test brittleness) are non-blocking.

[Octo-Q] verdict: APPROVE — zero P0/P1 findings; only P2 (reserved-token footgun) and a nit (test brittleness). Safe to land as-is; P2 can be addressed in a small follow-up.

Jerry-Xin (dismissed):
REQUEST_CHANGES — 1 blocker (CI), 0 code blockers.

The code is clean and correctly implements WS-20 (verified): the forward "recent" tab now sources from the backend `SidebarService.sync({tab:"recent"})`, building `recentKeys` + `recentOrder` from `${target_type}::${target_id}` → `timestamp`, matching ChatSelectorModal's proven pattern byte-for-byte. Key parity holds (`channelType::channelID` == `target_type::target_id`, confirmed by the group test), cancellation guard + empty-set fallback on missing deviceId/error are correct, and the new path has test coverage.

🔴 Blocking (CI): `pr-title-lint` is failing. Repo policy bars merge on a red required check, so this cannot be APPROVED until it goes green. The title `WS-20: 转发选群「最近」列表对齐智能总结（改用后端 recent 源）` lacks the conventional-commit `type(scope):` prefix. Suggested:
`feat(workbase): align forward "recent" tab to the backend recent source (WS-20)`

🟡 Non-blocking (fold into the same title-fix commit):
- The stale "recent = local conversations" framing lingers in a comment in `useForwardModal.ts`; update it to reflect the backend-recent source.
- A test description in `useForwardModal.test.tsx` drifts from the new backend-recent behavior; align the wording.

Will flip to APPROVE the moment the title lands green — no code findings block.

yujiawei (dismissed):
# Code Review — PR #811 (octo-web)

**Scope reviewed:** head SHA `7b757303415aa6fa6b701691c92cf46af9a9e0e6` against `main` (merge-base `22e72d4a`). Changed files: `packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts` (+60/-12) and its test (+81/-4).

**Intent:** Make the forward chat-selector's *Recent* tab authoritative from the backend (`SidebarService.sync({ tab: "recent" })`), matching the summary selector (`ChatSelectorModal`), instead of deriving "recent" from the local WKSDK conversation list.

---

## 1. Correctness / Spec compliance

The change does what it claims, and I verified the load-bearing assumptions directly:

- **Composite-key alignment is sound.** The recent effect builds keys as `${target_type}::${target_id}` and matches them against forward items keyed `${channelType}::${channelID}`. I confirmed the numeric spaces line up: `wukongimjssdk` exports `ChannelTypePerson=1`, `ChannelTypeGroup=2`, and `ChannelTypeCommunityTopic=5`, which equal `SidebarTargetType.DM=1 / CHANNEL=2 / THREAD=5` (`SidebarService.ts:3-8`). So DM/group/thread all map correctly with no cross-type collisions.
- **The new recent effect mirrors the existing follow effect** (`useForwardModal.ts:433-470` vs `405-432`): same `deviceId===""` guard (skips the request the backend would reject in `validateSidebarRequest`, degrades to an empty set), same `cancelled` guard on unmount, same `.catch(() => empty)`. Consistent and race-safe.
- **Local recent derivation was fully removed** — the old `recentKeySet` accumulation and `setRecentKeys(recentKeySet)` are gone from `rebuildConvItems` (`useForwardModal.ts:219-325`), so there's no lingering dual source.
- **Recent-tab sort matches the reference.** `displayItems` (`useForwardModal.ts:522-532`) sorts the recent-scoped list by `recentOrder` timestamp descending with `?? 0` fallback — semantically identical to `ChatSelectorModal.getDisplayList()`'s recent branch (`ChatSelectorModal.tsx:155-161`). Because `recentKeys` and `recentOrder` are populated in the same loop, every item that survives the recent scope has an order entry, so the fallback is effectively never hit and the ordering is well-defined. Other tabs keep the original tree order (`filtered`) untouched.
- **Shared `tabFilter.ts` was not touched** (confirmed: the diff is only the hook + its test), so follow / group / direct / search behavior is unchanged.

**Tests:** ran `useForwardModal.test.tsx` + `tabFilter.test.ts` locally → **45/45 pass**. `tsc --noEmit` on `dmworkbase` shows no errors in the changed file (only pre-existing third-party `semi-foundation` declaration noise). The new tests are meaningful: they assert recent is driven by the backend set (not local convs), that a local conv absent from the backend recent set no longer appears, timestamp-descending ordering, and the empty-`deviceId` degrade-to-empty path.

No P0/P1 correctness or security issues found.

---

## 2. Non-blocking findings (P2 / design tradeoffs)

A second-opinion pass (long-context, whole-repo reasoning) surfaced four points. I independently re-checked each and adjusted severity against the evidence:

- **[P2 — intended tradeoff] Recent is now fail-closed to empty on network failure / offline.** `useForwardModal.ts:460-465`: a failed sync empties `recentKeys`, and since *Recent* is the default tab, an offline user opening the forward modal now sees an empty Recent by default, where previously they saw locally-derived recents. This is the deliberate point of the PR (backend is the authority) and is **symmetric with the reference** (`ChatSelectorModal.tsx:74-75` also `.catch(() => null)` → empty) and with the sibling *follow* leg. Accepting it as intended; noting it so the offline-default-empty behavior is a conscious call, not an oversight.
- **[P2] Backend-recent items outside the local universe are silently dropped.** The recent tab filters `allItems` (local conversations + `group/my` fallback + friends + search) by `recentKeys`. Unlike the reference — whose candidate universe *and* recent set both come from the backend — a chat marked "recent" by the backend but not present in the locally-assembled universe won't render. In practice `group/my` covers joined groups and `friends` covers friend DMs, so the gap is narrow (e.g. a recent DM with a non-friend accessed on another device). It's also not forwardable anyway without a `Channel` ref in `channelMapRef` (`confirm()`, `useForwardModal.ts:558-566`), so this is a display-completeness edge, not a data bug.
- **[P2 → informational] "Two concurrent sync requests."** The second opinion flagged firing `follow` and `recent` as two separate syncs as extra backend load. Worth noting the reference is equivalent: `ChatSelectorModal.loadCandidates` also issues two separate `SidebarService.sync` calls (batched in one `Promise.all`, but still two HTTP requests). Net request count is the same; consolidating the two effects would only be a cleanliness improvement, not a load reduction.
- **[P2 — pre-existing] Recent-tab population lags `loading`.** `loading` is cleared when the local load + friends + groups finish (`useForwardModal.ts:335-377`), independent of the recent sync, so Recent can briefly show empty-then-populate after the spinner clears. This already applies to the *follow* leg introduced earlier; the new recent leg is consistent with it rather than introducing a new pattern. The reference avoids this by gating its spinner on the combined `Promise.all`. Minor, self-healing; optional to address.

---

## 3. Advisory coverage (transparency)

Two independent second-opinion legs were attempted. One (whole-repo, long-context) completed and its findings are folded in above. The other (line-level adversarial) **failed to run** due to an upstream model-gateway configuration error (unrelated to this PR) and is therefore recorded as **absent, not silent approval**. To keep coverage symmetric, the areas that leg would typically probe — and that I checked myself — are the composite-key numeric alignment, the `cancelled`/`catch` guards, and sort determinism, all verified above.

**Coverage note (not verifiable from the diff alone):** the exact runtime contents of `SidebarService.sync`'s response and whether `APIClient` performs any request coalescing/caching; real-device behavior of the default-tab empty-recent window on slow networks; and whether any caller depends on the previous local-recent ordering of `items` in the recent tab (none found in-repo — sole consumers are `ForwardModal.tsx` and `ConversationSelect`).

---

## 4. Verdict

Spec-compliant, well-tested, and faithfully aligned with the summary selector; the key-space and guard assumptions hold under direct inspection. All findings are P2 / intended design tradeoffs. **Approving.**

OctoBoooot (approved):
## Re-review (DELTA) #811 @ `89b6660f` — PR title fixed (title-lint green) + both 🟡 folded in; code logic unchanged → REQUEST_CHANGES → APPROVE

**Verdict:** Approve
This commit fixes the sole blocker of my prior REQUEST_CHANGES — the red `pr-title-lint`. The title is now `fix(workbase): align forward "recent" tab to the backend recent source (WS-20)` (conventional-commit compliant), and the two non-blocking 🟡 (stale "recent" comment, test-description drift) are folded in. The recent-sync *logic* is byte-identical to the head I already verified clean. Flipping my CR → APPROVE.

Code CI green: `pr-title-lint` newest leg `success` (the earlier failure leg superseded), Build newest leg `success`, secret-scan / osv-scan / dependency-review pass. size/M.

## Parity checked
The only source change is documentation; no behavior changed:
- **`useForwardModal.ts`** — the sole non-test edit is a **comment correction**: the stale "转发本地数据源以最近会话为主" framing replaced with an accurate "「最近」集合走后端 SidebarService recent 权威源... resolve 前 / 无 deviceId 时为空态" description. The recent-sync effect, key construction (`${target_type}::${target_id}`), `recentOrder` sort, cancel guard, and empty-set fallback are byte-identical to `7b757303` (which I byte-verified clean). Blob confirms only the comment moved.
- **Test file** — description updates to match the backend-recent behavior (the second 🟡).
- **Title** — now conventional-commit compliant; `pr-title-lint` is green.

### Resolved
- **The `pr-title-lint` blocker (my prior CR's sole reason) — cleared.** Title fixed to the exact form suggested.
- **🟡#1 (stale "recent" comment)** — corrected to describe the backend source + empty-state-before-resolve.
- **🟡#2 (test-description drift)** — updated to reflect the backend-recent behavior.

### Net
The forward "recent" tab's backend-source alignment (verified clean at the prior head — same `SidebarService.sync({tab:"recent"})` source + key parity + fail-closed handling as ChatSelectorModal) is unchanged; this commit only fixes the PR title and refreshes the two stale comments/test descriptions. Title-lint + Build green, no code findings. Flipping my **REQUEST_CHANGES → APPROVE** @ `89b6660f`. Credit @pkuWMH for the title fix + folding in the comment/test cleanups. (Reminder: #812 is the duplicate WS-20 PR — one of the two should be closed now that this one is green.)

Jerry-Xin (approved):
Summary: The PR is relevant to `octo-web` and correctly moves the forward modal’s `recent` tab to the backend sidebar recent source, matching the reference selector’s keying and timestamp ordering.

💬 Non-blocking
- 🟡 Warning `packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts:140`, `440-470`: the new recent sync is separate from the hook’s `loading` state. Since the default tab is `recent`, users may briefly see the empty state after local loading finishes but before `SidebarService.sync({ tab: "recent" })` resolves. This appears intentional from the comments, but it is worth accepting explicitly as a UX tradeoff.

✅ Highlights
- `useForwardModal.ts:448-459` uses the backend `recent` collection with the same composite key shape as `ChatSelectorModal`.
- `useForwardModal.ts:526-533` keeps timestamp ordering local to the forward hook, avoiding changes to shared `tabFilter` semantics.
- The added tests cover backend-driven membership, exclusion of local-only conversations, backend timestamp ordering, and the empty-device guard.

Verification note: I attempted to run the targeted test, but this checkout has no installed dependencies; `vitest` was not found and pnpm reported missing `node_modules`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #811 (octo-web)

## Summary

This PR switches the forward modal's "recent" tab from locally-derived conversation data to the backend `SidebarService.sync({ tab: "recent" })` authoritative source, aligning it with the smart summary selector (`ChatSelectorModal`). It also adds backend-timestamp-based descending sort for the recent tab. The change is well-scoped (2 files, 2 commits), correctly removes the old `recentKeySet` derivation from `rebuildConvItems`, and adds four targeted tests covering the new data source, sort order, negative (no local fallback), and empty-deviceId guard.

Overall: clean, well-tested change with no correctness issues found.

## Verification

- ✅ **Key construction parity** — ForwardModal's `forwardItemKey` uses `${channelType}::${channelID}` (1=person, 2=group, 5=topic), which maps exactly to `SidebarTargetType` (1=DM, 2=CHANNEL, 5=THREAD) used by the backend. Verified at `useForwardModal.ts:26` comment and `SidebarService.ts:4-8`.
- ✅ **Sort semantics match ChatSelectorModal** — Both use `(recentOrder.get(keyB) ?? 0) - (recentOrder.get(keyA) ?? 0)` for descending timestamp sort. Verified at `useForwardModal.ts:528-533` vs `ChatSelectorModal.tsx:155-162`.
- ✅ **Shared filter function reused** — `filterChatSelectorItems` from `tabFilter.ts` is used for tab scoping, ensuring identical filter semantics between forward and summary selectors.
- ✅ **Cancellation guard on async effect** — The `cancelled` flag at `:451` prevents state updates after unmount, consistent with the follow sync pattern at `:424`.
- ✅ **Empty deviceId guard** — Both recent and follow syncs skip the API call when `deviceId` is empty, degrading to empty sets. Tested explicitly.
- ✅ **Test coverage** — Four new tests: backend-driven recent (no local fallback), timestamp sort order, empty deviceId guard, and sidebarSync call verification. The `stubSidebar` helper correctly discriminates follow vs recent responses by `req.tab`.

## Findings

### P2 — Recent sync runs once on mount, no refresh mechanism (`useForwardModal.ts:436`)

The new `useEffect` for recent sync has `[]` deps, meaning it fires only once on mount. If the user keeps the forward modal open while new conversations arrive on the backend, the recent tab won't reflect them until the modal is remounted. This is consistent with the existing follow sync (also `[]` deps at `:413`) and the `ChatSelectorModal` (also single-shot), so it's not a regression — just noting the parity holds. If either selector later needs live updates, both should be updated together.

## Data Flow Traceback

1. **`recentKeys`** (Set) ← `SidebarService.sync({ tab: "recent" })` → `resp.items` → `${item.target_type}::${item.target_id}` composite keys. Data flows correctly to `filterChatSelectorItems` → `scopeByTab` → "recent" case at `tabFilter.ts:64`.
2. **`recentOrder`** (Map) ← Same loop, `item.timestamp` per key. Consumed only by the sort at `:528-533`. Items in `recentKeys` always have a corresponding entry in `recentOrder` (both populated in the same iteration).
3. **`displayItems`** ← `filtered` (tab+keyword scoped) → sorted by `recentOrder` for "recent" tab, passthrough for other tabs. Returned as `items` at `:573`.

No broken data flows or unreachable consumption points found.

## Blind-spot Checklist

- **C1 (dual-path parity)**: follow sync and recent sync are structurally identical (mount-only, cancelled guard, empty-deps, deviceId gate). ✅ Parity holds.
- **C2 (control-flow ordering)**: N/A — no nested reuse or re-ordering risk.
- **C3 (auth boundary)**: N/A — no auth/permission changes.
- **C4 (auth lifecycle)**: N/A — no auth changes.
- **C5 (build/runtime)**: N/A — no build/config changes.
- **C6 (governance/docs)**: N/A — no governance changes. Comment updates in commit `89b6660f` correctly reflect the new data source.

## Verdict: APPROVED

Clean, well-scoped alignment of the forward modal's "recent" tab with the backend-authoritative source, matching the smart summary selector's behavior. Key construction parity verified, sort semantics identical, test coverage adequate. The single P2 is an informational observation consistent with existing patterns.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Implementation correctly aligns forward and summary recent-tab data sources with verified key parity, sort semantics, and comprehensive test coverage.

yujiawei (approved):
# Code Review — PR #811 (octo-web)

**Head SHA reviewed:** `89b6660f3cf1f4912aa0bfa0216bf91f118fb680` against `main` (merge-base `22e72d4a`).
**Verdict:** APPROVE. No P0/P1 blockers. Findings below are all P2 (advisory / follow-up), not merge blockers.

## 1. Spec compliance

**Spec: ✅**

The change does exactly what issue #813 asks and nothing more:

- **Not under-built:** `useForwardModal` now sources the "recent" set from the backend authority via `SidebarService.sync({ tab: "recent" })` (useForwardModal.ts:446-469), builds `recentKeys` + `recentOrder` from `resp.items` with `${target_type}::${target_id}` composite keys, and sorts the recent tab by backend `timestamp` desc (useForwardModal.ts:523-533). This mirrors `ChatSelectorModal` (ChatSelectorModal.tsx:83-93, 155-161) — same key construction, same comparator (`timestamp` desc, missing `?? 0`), same empty-`deviceId` skip-and-degrade guard.
- **Not over-built:** the shared `tabFilter.ts` is untouched; recent sorting lives on the hook side only, so summary and other callers are unaffected. The local `recentKeySet` derivation was removed from `rebuildConvItems` (the old bug source). follow / group / direct / search / lazy-load / Space filtering / grant paths are unchanged.
- **No deviation:** composite-key homomorphism holds — `ChannelTypePerson=1=DM`, `ChannelTypeGroup=2=CHANNEL`, `ChannelTypeCommunityTopic=5=THREAD` match `SidebarTargetType`, so forward item keys align with backend keys for every kind including threads.

Tests: 45/45 pass locally (`useForwardModal.test.tsx` 35 + `tabFilter.test.ts` 10), and the timestamp-ordering test is genuinely non-tautological (local `sortConversations` order differs from the asserted backend order).

## 2. Code quality

**Quality: Approved**

The implementation is clean, the concurrency handling is correct (recent-sync and load effects write disjoint state slices; `cancelled` guard mirrors the existing follow effect; `[...filtered].sort()` copies before sorting, no mutation). The findings below are quality/parity follow-ups, none blocking.

### P2 — advisory

**P2-1 · Recent tab is the intersection of the LOCAL item set and backend `recentKeys`, not a projection of the backend recent set.**
`allItems = [...conversationItems, ...uniqueFriends, ...uniqueSearchGroups]` is locally sourced (useForwardModal.ts:506-510); the recent tab filters it by `recentKeys` (tabFilter.ts:58-59). The reference `ChatSelectorModal` instead filters the single authoritative backend list `getChatCandidates()` (ChatSelectorModal.tsx:157-158). So a chat that is in the backend recent set but absent from local conversations + groups + friends (e.g. an uncached DM with a non-friend, or a thread evicted from the local cache) shows in summary's recent tab but is silently missing from forward's. Practical impact is bounded: recent **groups** are backstopped by the awaited `groupSaveList()` (all joined groups), and any omitted target is still reachable via keyword search (backend search fires at keyword length ≥ 2). This is a deliberate, documented scope choice (keep the list body local for zero-permission-regression), so it is not a defect — but the code comment "完全对齐 / completely aligned" (useForwardModal.ts:434-435) and the test title "driven by the backend recent set, NOT local conversations" overstate it. Suggest softening those to "recent membership + ordering are backend-authoritative; the item universe remains local" so the next reader isn't misled.

**P2-2 · `loading` is not wired to the recent-sync request → brief empty flash on the default tab.**
The hook's `loading` flag resolves when `load()` finishes (useForwardModal.ts:336/377); it does not await the recent-sync effect. Since "recent" is the default tab and its keys populate asynchronously, opening the modal can show an empty recent list for a moment before the sync resolves. The reference avoids this by gating `loading:false` on `Promise.all([...recent...])` (ChatSelectorModal.tsx:72-94). Self-healing and sub-second, so P2, but worth aligning if you want the flash gone.

**P2-3 · Recent-sync effect has `[]` deps → does not re-sync on Space switch while the modal is open.**
`load()` re-runs on Space change via the `conversation-list-refreshed` bus event (useForwardModal.ts:392-395), refreshing `conversationItems` to the new Space, but the recent-sync effect (deps `[]`, line 470) does not, so `recentKeys` stays pinned to the old Space. `scopeByTab` then filters new-Space items by old-Space keys and the default recent tab collapses to near-empty until the modal is re-opened. Edge case (switching Space with a non-masking forward overlay open), and the follow tab already had this `[]`-deps limitation pre-PR — but recent is now the default tab, so the staleness is more visible. Consider gating the recent-sync effect on `currentSpaceId` (or subscribing to the same refresh event as `load()`).

**P2-4 · Thread rows can render an orphaned child indent in the recent tab.**
After the timestamp re-sort, a thread `ForwardItem` still carries `parentChannelID`, and `ForwardModal` applies the `wk-fm-item--child` indent whenever that is truthy — but its parent group may not be adjacent (or may be absent from the recent set). The reference flattens recent rows (`indent:false`, ChatSelectorModal.tsx:155-161), so this is a small visual divergence. Cosmetic.

**P2-5 · Missing-timestamp items sink to the bottom (`?? 0`).**
If the backend ever omits a `timestamp`, that item sorts to the end. This matches the reference exactly, so it is parity-correct, not a regression — noted only for awareness.

**P2-6 · The empty-`deviceId` recent test is near-tautological.**
`useForwardModal.test.tsx` (the "recent tab is empty when deviceId is empty" case) uses the default empty sidebar stub, so the assertion `not.toContain("g1")` passes whether or not the empty-`deviceId` guard exists. To actually prove the guard + "no local fallback," either assert `sidebarSync` was NOT called with `tab:"recent"` when `deviceId` is empty, or stub a non-empty recent set and assert those items are still absent. Test-effectiveness only; the production guard itself is correct.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The fix is correct, well-tested, and scoped. All findings are P2 follow-ups; none blocks merge. P2-1 (comment/test-title accuracy) and P2-3 (Space-switch re-sync) are the two most worth a quick follow-up.

## 4. Suggestions

- Tighten the "完全对齐 / fully aligned" comment and the test title to reflect that only recent membership + ordering are backend-authoritative while the item source stays local (P2-1).
- If you want to close the last parity gaps in one pass: wire `loading` to the recent sync (P2-2), re-run recent sync on Space change (P2-3), and flatten recent-tab indentation (P2-4).

## 5. Additional notes

- Concurrency reviewed and clean: independent effects on disjoint state, correct `cancelled` guards, no mutation of `filtered`.
- Advisory-leg coverage for this review: an independent long-context advisory model was consulted and its findings are folded in above (it independently surfaced the loading-flash and no-fallback behavior, both adjudicated here as non-blocking/aligned-to-reference). A second cross-family advisory tool was attempted but was unavailable due to a gateway configuration error, so it is recorded as absent rather than as silent agreement; this review's blocking judgment rests on the direct code read plus the successful advisory leg.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #811 (octo-web)

## Summary

This PR switches the forward modal's "recent" tab from a local-conversation-derived data source to the backend `SidebarService` recent sync — aligning it with the summary selector (`ChatSelectorModal`) which already uses the backend authoritative recent set. A new `recentOrder` state (key → backend timestamp) drives descending timestamp sort, replacing the previous local tree-order output. Three new tests guard the new behavior (backend-driven filtering, sort order, empty-deviceId fallback); the existing default-tab test is updated to stub the backend recent response.

## Verification

- ✅ **Composite key alignment** — `forwardItemKey` at `useForwardModal.ts:24` builds `${channelType}::${channelID}` where `channelType` values (1=DM, 2=Group, 5=Topic) match `SidebarTargetType` at `SidebarService.ts:4-8` (DM=1, CHANNEL=2, THREAD=5). The new recent sync effect at `:443` builds keys as `${item.target_type}::${item.target_id}` from the same backend response shape. Keys match correctly.
- ✅ **ChatSelectorModal sort parity** — The sort at `useForwardModal.ts:529-531` (`recentOrder.get(getKey(b)) - recentOrder.get(getKey(a))`, descending) is identical in structure to `ChatSelectorModal.tsx:160`. Both use `?? 0` for missing timestamps.
- ✅ **Empty deviceId guard** — At `useForwardModal.ts:438-441`, empty `deviceId` clears both `recentKeys` and `recentOrder`, yielding an empty recent tab. This matches the existing follow sync guard at `:414-417` and the test at `test.tsx:920-928`.
- ✅ **Cancellation** — Both follow (`:425-427`) and recent (`:464-466`) sync effects use the `let cancelled` pattern with cleanup return. No stale-state risk on unmount.
- ✅ **Test coverage** — Four test cases now cover: default-tab with backend recent (line 849), backend-only filtering excluding local-only items (line 872), backend timestamp sort order (line 890), and empty-deviceId empty-tab (line 920). The `stubSidebar` helper (line 834) correctly discriminates follow vs recent by `req?.tab`.
- ✅ **Mock type safety** — The explicit return type annotation on the default `sidebarSync` mock at `test.tsx:55` prevents `never[]` inference on the empty `items` array — a legitimate TypeScript fix for subsequent `mockImplementation` overrides.

## Findings

No P0/P1 issues. No P2 issues. The change is clean, well-tested, and correctly aligned with the summary selector.

Static analysis only at head `ba46700f`; build and tests not executed in this environment.

## Data-Flow Trace

- **recentKeys / recentOrder** — Produced by `SidebarService.sync({ tab: "recent" })` at `useForwardModal.ts:443`. Response `items[]` has `{ target_type, target_id, timestamp }`. Key = `${target_type}::${target_id}`. Consumed by `scopeByTab` (`tabFilter.ts:97`) via `recentKeys.has(acc.getKey(i))` and by the sort at `useForwardModal.ts:529-531` via `recentOrder.get(acc.getKey(i))`. Upstream `SidebarService.ts:61` POSTs to `/sidebar/sync` with `tab: "recent"` and returns `SidebarSyncResp` (`:50`). Data reaches the consumer correctly.
- **forwardItemKey alignment** — `forwardItemKey` (`:24`) returns `${channelType}::${channelID}`. `channelType` from `ConversationWrap.channel.channelType` (1/2/5 from wukongimjssdk). Backend `target_type` uses the same enum (`SidebarTargetType` 1/2/5). Keys produced by both sides are structurally identical.
- **filtered → displayItems** — `filterChatSelectorItems` at `:517-520` returns items scoped to `recentKeys`. Then `displayItems` at `:523-534` sorts the filtered set by `recentOrder` descending. The two-phase pipeline (filter then sort) is correct — no item can appear in the sort that was not first admitted by the filter.
- **No dual-write** — Old `setRecentKeys(recentKeySet)` removed from `rebuildConvItems` (`:326` removed). `recentKeys` is now exclusively set by the sync effect. No race between local derivation and backend sync.

## Blind-Spot Checklist

- **C1 (dual-path parity)** — Clear. The follow sync pattern at `:412-430` and new recent sync at `:434-468` use identical structure (deviceId guard, cancelled flag, catch handler, cleanup). No asymmetric pair to check.
- **C2 (control-flow ordering / nesting reuse)** — Clear. `displayItems` sort is computed in render (not memoized), applied only for `activeTab === "recent"`. No nested reuse or ordering risk.
- **C3 (authorization boundary)** — N/A. No auth/permission changes in this PR.
- **C4 (authorization lifecycle)** — N/A. No auth changes.
- **C5 (build/note ≠ runtime correctness)** — N/A. No build/packaging changes.
- **C6 (governance/docs self-consistency)** — Clear. Comment updates at `:142-153` and `:407-409` accurately reflect the new data source (backend recent vs local conversations). No contradiction with other docs.

## Cross-Round Blocker Check

- Follow sync effect (`:412-430`) is unchanged in logic — only comment updated. Existing follow tab tests (`:1000-1025`) continue to work: they use `mockResolvedValue` directly, which overrides the default mock for both tab requests. The default mock (returning empty items for any tab) ensures recent sync does not interfere.
- `rebuildConvItems` correctly removes `recentKeySet` local derivation (lines 222/326 removed) — `setRecentKeys` is now exclusively owned by the backend sync effect, no dual-write race.
- The `displayItems` variable at `:523-534` applies sort only when `activeTab === "recent"`, leaving all other tabs' tree-order output untouched.
- `filterChatSelectorItems` at `tabFilter.ts:97-100` filters by `recentKeys` before the sort step — items not in the backend recent set are excluded, then the remaining are sorted. Correct two-phase pipeline.
- No cross-file contract breakage: `FORWARD_ITEM_ACCESSORS.getKey` (`:40`) delegates to `forwardItemKey` which produces the same composite key shape as the backend sync effect.

## Verdict: APPROVED

Clean data-source migration with correct key alignment, sort parity with ChatSelectorModal, proper cancellation, and thorough test coverage. No correctness, security, or build concerns.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Key alignment between `forwardItemKey` and backend `target_type::target_id` verified. Sort semantics match ChatSelectorModal. Test coverage guards all new behavior paths including the empty-deviceId edge case.

yujiawei (approved):
# Code Review — PR #811 (octo-web)

**Head SHA reviewed:** `ba46700f7f5914a40d965d76b53a7edd49dd9694` against `main` (merge-base `22e72d4a`). This SHA is tree-identical to the previously-reviewed `89b6660f` (same parent, same content — a re-stamped commit), so the earlier approvals' conclusions carry, but this verdict is issued independently against the current head.

**Verdict:** APPROVE. No P0/P1 blockers. All findings below are P2 (advisory / follow-up), none blocks merge.

## 1. Spec compliance

**Spec: ✅**

The change does exactly what the linked issue asks and nothing more:

- **Not under-built:** `useForwardModal` now sources the "recent" set from the backend authority via `SidebarService.sync({ tab: "recent" })` (`useForwardModal.ts:436-470`), builds `recentKeys` + `recentOrder` from `resp.items` with `${target_type}::${target_id}` composite keys, and sorts the recent tab by backend `timestamp` desc (`useForwardModal.ts:526-533`). This mirrors `ChatSelectorModal` (`ChatSelectorModal.tsx:83-93`, `155-161`) — same key construction, same comparator (`timestamp` desc, missing `?? 0`), same empty-`deviceId` skip-and-degrade guard.
- **Not over-built:** the shared `tabFilter.ts` is untouched; recent sorting lives on the hook side only, so the summary selector and other callers are unaffected. The local `recentKeySet` derivation was removed from `rebuildConvItems` (the old bug source).
- **No deviation:** composite-key homomorphism holds — `ChannelTypePerson=1=DM`, `ChannelTypeGroup=2=CHANNEL`, `ChannelTypeCommunityTopic=5=THREAD` match `SidebarTargetType`, so forward item keys align with backend keys for every kind including threads.

Tests: **45/45 pass locally** (`useForwardModal.test.tsx` 35 + `tabFilter.test.ts` 10). The timestamp-ordering test is genuinely non-tautological (local `sortConversations` order differs from the asserted backend order).

## 2. Code quality

**Quality: Approved**

The implementation is clean. The recent-sync effect is structurally identical to the existing follow effect (empty-`deviceId` early return, `.catch` degrade-to-empty, `cancelled` guard); the recent-sync and `load()` effects write disjoint state slices, so there is no race; `[...filtered].sort()` copies before sorting, so `filtered` is never mutated. Findings below are quality/parity follow-ups, none blocking.

### P2 — advisory

**P2-1 · The "完全对齐 / 完全一致" comments overstate the alignment: only the recent *set and ordering* are backend-authoritative; the displayed *item universe* stays local.**
The recent tab is the intersection of the locally-assembled `allItems` (`useForwardModal.ts:510` = local conversations + `group/my` + friends, plus keyword-search results only when `keyword>=2`) with the backend `recentKeys` (`tabFilter.ts:58-59`). The reference `ChatSelectorModal` instead intersects the backend recent set with the full backend candidate list `getChatCandidates()` (`ChatSelectorModal.tsx:157-158`). So a target that is in the backend recent set but absent from the local universe (e.g. a recent DM with a non-friend that has no local SDK conversation object, or a recent thread stripped by `filterArchivedThreads`) would appear in the summary selector's recent tab but be missing from the forward one.
**Why this is P2 and not a blocker:** the local candidate universe is **pre-existing** — before this PR the recent set was *also* derived purely from local conversations (`recentKeySet` from `wrapsRef`, now removed). This PR does not narrow the universe; it only replaces the buggy local top-weighted membership/order with the backend-authoritative one — a strict improvement, and the documented scope ("零权限回归 / keep the list body local"). Practical impact is further bounded: recent **groups** are backstopped by the awaited `groupSaveList()`, and any omitted target is reachable via keyword search when `searchChatCandidates` is wired (it is, in both the summary module and the extension sidepanel). Recommend softening the `完全对齐 / 完全一致` comments (`useForwardModal.ts:152`, `434`) and the test title "driven by the backend recent set, NOT local conversations" to "recent *membership + ordering* are backend-authoritative; the item universe remains local," so the next reader isn't misled.

**P2-2 · `loading` is not wired to the recent-sync request → brief empty flash on the default tab.**
The hook's `loading` flag resolves when `load()` finishes; it does not await the recent-sync effect. Since "recent" is the default tab and its keys populate asynchronously, opening the modal can show an empty recent list momentarily before the sync resolves. The reference gates `loading:false` on `Promise.all([...recent...])` (`ChatSelectorModal.tsx:72-94`). Self-healing and sub-second, so P2.

**P2-3 · Recent-sync effect has `[]` deps → does not re-sync on Space switch while the modal is open.**
`load()` re-runs on Space change via the `conversation-list-refreshed` bus event, refreshing `conversationItems` to the new Space, but the recent-sync effect (deps `[]`) does not, so `recentKeys` stays pinned to the old Space and the default recent tab can collapse to near-empty until the modal is re-opened. The follow tab already had this `[]`-deps limitation pre-PR, but recent is now the default tab, so the staleness is more visible. Consider gating the recent-sync effect on `currentSpaceId` (or subscribing to the same refresh event as `load()`).

**P2-4 · Recent-tab timestamp sort can separate a child thread from its parent group, leaving an orphaned indented row.**
`displayItems` sorts the recent tab purely by backend timestamp desc (`useForwardModal.ts:526-533`), while `ForwardModal` still applies the `wk-fm-item--child` indent whenever `item.parentChannelID` is truthy. A recent thread whose timestamp interleaves away from its parent renders as an indented child row under an unrelated item. The reference renders the recent tab flat (`indent:false`, `ChatSelectorModal.tsx:161`). Cosmetic divergence; items stay present and selectable.

**P2-5 · Test blind spots (recent tab).** The new recent-tab tests exercise only `target_type=2` (group) items, so several paths are unproven, though the production code is sound:
- No DM (`type 1`) or thread (`type 5`) in the recent set → the composite-key match for cross-type keys and the tree-order-breaking sort (P2-4) are untested.
- No test drives the recent-sync `.catch` degradation path.
- No test combines a keyword with the recent tab.
- The empty-`deviceId` recent test never asserts `sidebarSync` was *not* called, so it does not actually prove the "doomed request guard" (it would pass even if the request fired and returned empty) — its follow analog does assert `not.toHaveBeenCalled()`. Recommend mirroring that assertion.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved. The fix is correct, well-tested, and scoped. All findings are P2 follow-ups; none blocks merge. P2-1 (comment/test-title accuracy) and P2-3 (Space-switch re-sync) are the two most worth a quick follow-up.

## 4. Suggestions

- Tighten the "完全对齐 / 完全一致" comments and the test title to reflect that only recent membership + ordering are backend-authoritative while the item source stays local (P2-1).
- If closing the remaining parity gaps in one pass: wire `loading` to the recent sync (P2-2), re-run recent sync on Space change (P2-3), flatten recent-tab indentation (P2-4), and add the missing recent-tab test cases (P2-5).

## 5. Additional notes

- Concurrency reviewed and clean: independent effects on disjoint state, correct `cancelled` guards, no mutation of `filtered`.
- Verification performed: local `pnpm install` + `vitest run` on the two affected suites (45/45 green); type-safety of the changed logic confirmed; composite-key ↔ `SidebarTargetType` homomorphism verified against the wukongimjssdk constants; the current head confirmed tree-identical to the prior reviewed SHA.
- Coverage / blind spots: this review focused on the 2-file diff and its direct data flow into the shared filter and the render path. Runtime behavior with a live backend recent set containing DM/thread targets was reasoned about but not exercised end-to-end (no live backend available in the review environment).

pkuWMH:
Closing as superseded. The WS-20 bug (forward "recent" target list diverging from the summary selector) was fixed on `main` by #984 (`fix(forward): align recent targets with sidebar ordering`), merged 2026-07-22, which closed the shared GitHub issue #813 via `Fixes #813`.

#984 is the equivalent-and-then-some landing of the same change this PR made: it switches the forward "recent" tab to the backend `SidebarService.sync({ tab: "recent" })` source (same root-cause fix as here), and additionally authoritative-orders by the recent API's `timestamp` + `is_pinned` (pinned-first, then timestamp desc) and adds the DM/group/thread kind tags with i18n. Since the fix is already on `main`, this PR is no longer needed.

PR #89, opened 2026-05-22
chore: add CODEOWNERS (web-maintainers)

lml2468:
Add CODEOWNERS file pointing to `@Mininglamp-OSS/web-maintainers`.

This is part of the sub-maintainer team setup. All code review assignments will be managed via team membership going forward — no need to edit this file when members change.

/cc @lml2468

PR #825, opened 2026-07-16
fix(docs): record a view when a shared doc opens on the standalone page

ploy-elison:
## Problem

A document opened from a chat **share link** (the standalone `/d/:docId` page, `StandaloneDocPage`) never showed up in the opener's **"recently viewed"** list. Root cause: the standalone page never calls `recordDocView` — unlike the in-shell `DocsHome` entry, which records a view on open.

## Change

- **`StandaloneDocPage`**: once the doc reaches the `ready` phase, record a single view (mirrors `DocsHome.commitOpen`). Fire-and-forget, guarded by `docId` so it fires at most once per opened doc and never in a render loop.
- The view is written to the **viewer's current space** via an explicit `X-Space-Id` header — **not** the doc's own space that the standalone page seeds for preflight/room addressing. The backend writes and reads the recent-view list by the viewer's current space, so writing under the doc space would keep a cross-space share invisible in the recipient's recent list. The viewer space is captured once on first render, before the seed effect overwrites the shared current space.
- **`recordDocView`** gains an optional `spaceId` to carry that explicit header, mirroring the existing `getDoc` opts.

The space semantics follow the backend write/read space contract (viewer-current-space for both write and read).

## Tests

- **Unit (`StandaloneDocPage.test.tsx`)**: view is recorded once on `ready` with `X-Space-Id` = viewer current space (cross-space share) and = live space (in-shell mount); idempotent (single call, never in a render loop); not recorded when the preflight fails to a terminal.
- **E2E (`apps/web/e2e/standalone-doc.spec.ts`)**: real-browser coverage that opening a cross-space `/d/:docId` link fires `POST /docs/:id/view` with the viewer-space header, and that a 403 preflight records no view.

## Notes

Draft — carries review while backend deploy + device verification are coordinated.

### DIFF
--- a/apps/web/e2e/standalone-doc.spec.ts
+++ b/apps/web/e2e/standalone-doc.spec.ts
@@ -174,3 +174,64 @@ test.describe('standalone /d/:docId — clean cold-load (no in-app sid route)', 
     await expect(page.locator('.octo-doc-standalone')).toHaveCount(0)
   })
 })
+
+test.describe('standalone /d/:docId — records a view in the viewer space (XIN-1238)', () => {
+  test('opening a cross-space share link fires POST /docs/:id/view with X-Space-Id = viewer current space, not the doc space', async ({
+    page,
+  }) => {
+    await seedSignedInSession(page)
+    // The viewer is currently in `space-viewer`; the shell persists that under the shared
+    // `currentSpaceId` key. On a cold `/d/:docId` deep link this is the only source of the viewer's
+    // real space (the live space is not yet restored).
+    await page.addInitScript(() => {
+      try {
+        localStorage.setItem('currentSpaceId', 'space-viewer')
+      } catch {
+        /* noop */
+      }
+    })
+    await stubBackend(page)
+    // Cross-space share: the link carries the DOC's own space (`?sp=space-doc`), which the page uses
+    // to address the preflight — deliberately different from the viewer's current space.
+    await routeDocPreflight(page, 'd_ok', 200, {
+      docId: 'd_ok',
+      title: 'Shared Doc',
+      ownerId: 'u_owner',
+      role: 'reader',
+    })
+
+    // Capture the view-ingest call. Registered after the catch-all so it wins (Playwright runs the
+    // most-recently-added matching handler first).
+    let viewSpaceHeader: string | undefined
+    let viewCalls = 0
+    await page.route('**/api/v1/docs/d_ok/view', (route: Route) => {
+      viewCalls += 1
+      viewSpaceHeader = route.request().headers()['x-space-id']
+      route.fulfill({ status: 200, contentType: 'application/json', body: '{}' })
+    })
+
+    await page.goto('/d/d_ok?sp=space-doc')
+
+    // The view was recorded once, addressed to the VIEWER's current space so it surfaces in the
+    // viewer's 最近查看 (XIN-1237 write/read contract) — NOT the doc's own space.
+    await expect.poll(() => viewCalls).toBe(1)
+    expect(viewSpaceHeader).toBe('space-viewer')
+  })
+
+  test('does not record a view when the preflight fails to a terminal (403)', async ({ page }) => {
+    await seedSignedInSession(page)
+    await stubBackend(page)
+    await routeDocPreflight(page, 'd_forbidden', 403)
+
+    let viewCalls = 0
+    await page.route('**/api/v1/docs/d_forbidden/view', (route: Route) => {
+      viewCalls += 1
+      route.fulfill({ status: 200, contentType: 'application/json', body: '{}' })
+    })
+
+    await page.goto('/d/d_forbidden')
+
+    await expect(page.locator('.octo-terminal')).toBeVisible()
+    expect(viewCalls).toBe(0)
+  })
+})
--- a/packages/docs/src/pages/StandaloneDocPage.test.tsx
+++ b/packages/docs/src/pages/StandaloneDocPage.test.tsx
@@ -974,3 +974,53 @@ describe('StandaloneDocPage — records a view so share-link opens surface in �
     expect(wk.apiClient.calls.some((c) => c.url === '/docs/d_forbidden/view')).toBe(false)
   })
 })
+
+describe('StandaloneDocPage — restores the viewer current space on teardown so 最近查看 is not left scoped to the doc space (XIN-1254)', () => {
+  it('reverts the doc-link space it seeded back to the viewer current space when the page unmounts', async () => {
+    // Cross-space share (老板 real-device repro): the viewer (大棍子) is in space-viewer; the link
+    // carries the DOC owner's (大背头) space `?sp=space-doc`. On this cold deep link
+    // wk.shared.currentSpaceId is empty, so the seed effect overwrites it with space-doc for the
+    // interceptor to address the editor's cross-space collab requests. Left in place, returning to
+    // the docs list scopes 最近查看's read (derived from the live currentSpaceId via the interceptor)
+    // to space-doc — but the view was recorded under space-viewer (viewerSpaceRef), so it stays
+    // invisible: the XIN-1234 write/read space split reintroduced by the seed.
+    window.history.pushState({}, '', '/d/d_ok?sp=space-doc')
+    window.localStorage.setItem('currentSpaceId', 'space-viewer')
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Shared Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    const { unmount } = render(<StandaloneDocPage docId="d_ok" />)
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    // Seed applied while the standalone editor is mounted (cross-space collab addressing).
+    expect(wk.shared.currentSpaceId).toBe('space-doc')
+
+    // On teardown the viewer's real space is restored, so the shell's recent-view read matches the
+    // space the view was written to (space-viewer) and the doc surfaces in 最近查看.
+    unmount()
+    expect(wk.shared.currentSpaceId).toBe('space-viewer')
+  })
+
+  it('only reverts the value it seeded — never clobbers a space the shell switched to meanwhile', async () => {
+    window.history.pushState({}, '', '/d/d_ok?sp=space-doc')
+    window.localStorage.setItem('currentSpaceId', 'space-viewer')
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Shared Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    const { unmount } = render(<StandaloneDocPage docId="d_ok" />)
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    expect(wk.shared.currentSpaceId).toBe('space-doc')
+
+    // A genuine space switch happened while the page was open; the teardown restore must be a no-op.
+    wk.shared.currentSpaceId = 'space-switched'
+    unmount()
+    expect(wk.shared.currentSpaceId).toBe('space-switched')
+  })
+})
--- a/packages/docs/src/pages/StandaloneDocPage.test.tsx
+++ b/packages/docs/src/pages/StandaloneDocPage.test.tsx
@@ -65,6 +65,7 @@ import {
   parseStandaloneDocId,
   isStandaloneDocPath,
   standaloneFallbackSpace,
+  viewerCurrentSpace,
   standaloneLinkSpace,
   persistStandaloneReturn,
   consumeStandaloneReturn,
@@ -512,6 +513,26 @@ describe('standaloneFallbackSpace — cold-start cross-space addressing (blocker
   })
 })
 
+describe('viewerCurrentSpace — the space a recorded view is written to (XIN-1237, no deploy-default tail)', () => {
+  it('prefers the live currentSpaceId when the shell has one', () => {
+    window.localStorage.setItem('currentSpaceId', 's_cached')
+    expect(viewerCurrentSpace('s_live')).toBe('s_live')
+  })
+
+  it('falls back to the cached localStorage currentSpaceId when the shell has none', () => {
+    window.localStorage.setItem('currentSpaceId', 's_cached')
+    expect(viewerCurrentSpace('')).toBe('s_cached')
+    expect(viewerCurrentSpace(undefined)).toBe('s_cached')
+  })
+
+  it('returns empty (NOT the deploy default) when there is no viewer signal, so the caller omits the header', () => {
+    // Unlike standaloneFallbackSpace (used for room addressing, where a default is a safe last
+    // resort), the view record must never be written to a space we cannot confirm the viewer is in.
+    expect(viewerCurrentSpace('')).toBe('')
+    expect(viewerCurrentSpace(undefined)).toBe('')
+  })
+})
+
 describe('StandaloneDocPage — cold-start addressing uses the cached space, not the deploy default (blocker 3)', () => {
   it('addresses the EditorShell to the cached currentSpaceId when the preflight carries no documentName', async () => {
     // Repro: 200 preflight WITHOUT documentName + a cached currentSpaceId in localStorage. Before
@@ -917,6 +938,55 @@ describe('StandaloneDocPage — records a view so share-link opens surface in �
     const preflight = wk.apiClient.calls.find((c) => c.method === 'get' && c.url === '/docs/d_ok')
     expect(preflight!.config?.headers?.['X-Space-Id']).toBe('space-doc')
     expect(view!.config?.headers?.['X-Space-Id']).toBe('space-viewer')
+  })
+
+  it('same-space share (老板 real-device acceptance): opening a doc from a same-space group link records the view under that space so it surfaces in 最近查看', async () => {
+    // The acceptance scenario (三叉戟大队): the viewer opens a doc shared in the space they are
+    // already in. The link's `?sp=` and the viewer's current space are the SAME space, so the view
+    // is recorded under that space and the shell's recent-view read (scoped to the same current
+    // space) surfaces it — the exact "点链接开同 space 文档→回→最近查看可见" flow.
+    window.history.pushState({}, '', '/d/d_ok?sp=space-team')
+    window.localStorage.setItem('currentSpaceId', 'space-team')
+
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Team Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<StandaloneDocPage docId="d_ok" />)
+
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    await waitFor(() =>
+      expect(wk.apiClient.calls.some((c) => c.method === 'post' && c.url === '/docs/d_ok/view')).toBe(true),
+    )
+    const view = wk.apiClient.calls.find((c) => c.method === 'post' && c.url === '/docs/d_ok/view')
+    expect(view!.config?.headers?.['X-Space-Id']).toBe('space-team')
+  })
+
+  it('per-space safety: with no live and no cached viewer space, the view record omits the explicit X-Space-Id instead of forcing the deploy default', async () => {
+    // No viewer signal at all (no live currentSpaceId, no cached one). Recording the view under the
+    // deploy-default space would write it into a space the viewer isn't in — polluting that space's
+    // recent list and still never surfacing in the viewer's own. Omit the explicit header and let
+    // the global interceptor decide, exactly as the in-shell entry does.
+    window.history.pushState({}, '', '/d/d_ok?sp=space-doc')
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Shared Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<StandaloneDocPage docId="d_ok" />)
+
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    await waitFor(() =>
+      expect(wk.apiClient.calls.some((c) => c.method === 'post' && c.url === '/docs/d_ok/view')).toBe(true),
+    )
+    const view = wk.apiClient.calls.find((c) => c.method === 'post' && c.url === '/docs/d_ok/view')
+    // No explicit header forced: never the deploy default, never the doc link space.
+    expect(view!.config?.headers?.['X-Space-Id']).toBeUndefined()
   })
 
   it('uses the live current space as the viewer space when the shell already restored it (in-shell mount)', async () => {
--- a/packages/docs/src/pages/StandaloneDocPage.test.tsx
+++ b/packages/docs/src/pages/StandaloneDocPage.test.tsx
@@ -886,3 +886,91 @@ describe('StandaloneDocPage — creator name is nickname-only on the shared surf
     expect(screen.getByTestId('editor-creator-nickname-only').textContent).toBe('true')
   })
 })
+
+describe('StandaloneDocPage — records a view so share-link opens surface in 最近查看 (XIN-1238)', () => {
+  it('fires POST /docs/:id/view once the doc is ready, written to the VIEWER current space, not the doc link space', async () => {
+    // XIN-1234 root cause: the standalone page never recorded a view, so a doc opened from a chat
+    // share link never appeared in the opener's 最近查看. XIN-1237 write/read space contract: the
+    // view must be written under the VIEWER's current space (X-Space-Id), which 最近查看 reads back
+    // by — NOT the doc's own space the standalone page seeds for preflight addressing.
+    // Cross-space share: the link carries the doc space `?sp=space-doc`, but the viewer is currently
+    // in space-viewer (their cached/live current space).
+    window.history.pushState({}, '', '/d/d_ok?sp=space-doc')
+    window.localStorage.setItem('currentSpaceId', 'space-viewer')
+
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Shared Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<StandaloneDocPage docId="d_ok" />)
+
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    await waitFor(() =>
+      expect(wk.apiClient.calls.some((c) => c.method === 'post' && c.url === '/docs/d_ok/view')).toBe(true),
+    )
+    const view = wk.apiClient.calls.find((c) => c.method === 'post' && c.url === '/docs/d_ok/view')
+    // The preflight is addressed to the doc's own space (link `?sp`), but the view ingest must go to
+    // the viewer's current space so it surfaces in the viewer's recent list.
+    const preflight = wk.apiClient.calls.find((c) => c.method === 'get' && c.url === '/docs/d_ok')
+    expect(preflight!.config?.headers?.['X-Space-Id']).toBe('space-doc')
+    expect(view!.config?.headers?.['X-Space-Id']).toBe('space-viewer')
+  })
+
+  it('uses the live current space as the viewer space when the shell already restored it (in-shell mount)', async () => {
+    wk.shared.currentSpaceId = 'space-live'
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Shared Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<StandaloneDocPage docId="d_ok" />)
+
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    await waitFor(() =>
+      expect(wk.apiClient.calls.some((c) => c.method === 'post' && c.url === '/docs/d_ok/view')).toBe(true),
+    )
+    const view = wk.apiClient.calls.find((c) => c.method === 'post' && c.url === '/docs/d_ok/view')
+    expect(view!.config?.headers?.['X-Space-Id']).toBe('space-live')
+  })
+
+  it('records the view at most once (idempotent — never in a render loop)', async () => {
+    window.localStorage.setItem('currentSpaceId', 'space-viewer')
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_ok') {
+        return { data: { docId: 'd_ok', title: 'Shared Doc', ownerId: 'u_owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<StandaloneDocPage docId="d_ok" />)
+
+    await waitFor(() => expect(screen.getByTestId('editor-shell')).toBeTruthy())
+    await waitFor(() =>
+      expect(wk.apiClient.calls.filter((c) => c.method === 'post' && c.url === '/docs/d_ok/view').length).toBe(1),
+    )
+    // Give any stray re-render a chance to double-fire, then re-assert the single call.
+    await new Promise((r) => setTimeout(r, 0))
+    expect(wk.apiClient.calls.filter((c) => c.method === 'post' && c.url === '/docs/d_ok/view').length).toBe(1)
+  })
+
+  it('does NOT record a view when the preflight fails to a terminal (forbidden / not-found)', async () => {
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d_forbidden') {
+        return Promise.reject(apiError(403))
+      }
+      return { data: {}, status: 200 }
+    }
+
+    render(<StandaloneDocPage docId="d_forbidden" />)
+
+    await waitFor(() =>
+      expect(wk.apiClient.calls.some((c) => c.method === 'get' && c.url === '/docs/d_forbidden')).toBe(true),
+    )
+    expect(wk.apiClient.calls.some((c) => c.url === '/docs/d_forbidden/view')).toBe(false)
+  })
+})
--- a/packages/docs/src/pages/StandaloneDocPage.tsx
+++ b/packages/docs/src/pages/StandaloneDocPage.tsx
@@ -7,7 +7,7 @@ import { DocTerminal, type TerminalKind } from '../editor/DocTerminal.tsx'
 import { RequestAccessButton } from '../access-request/RequestAccessButton.tsx'
 import { LinkIcon, type DocMoreMenuItem } from '../editor/DocMoreMenu.tsx'
 import { terminalForCreateError } from '../collab/useCollabEditor.ts'
-import { getDoc, type DocMeta } from './docsApi.ts'
+import { getDoc, recordDocView, type DocMeta } from './docsApi.ts'
 import { parseDocumentName } from '../documentName/index.ts'
 import { DEFAULT_DOC_SPACE, DEFAULT_DOC_FOLDER } from '../config.ts'
 import { useMemberNames } from '../members/useMemberNames.ts'
@@ -310,6 +310,20 @@ export function StandaloneDocPage({
   // addresses the doc correctly when the opener is already in the doc's space (owner / same-space).
   const preflightSpace = standaloneLinkSpace() || standaloneFallbackSpace(wk.shared?.currentSpaceId)
 
+  // XIN-1237 write/read space contract: a view is written into the space carried by the request's
+  // X-Space-Id and "最近查看" reads back by that SAME viewer space. recordDocView must therefore be
+  // written to the VIEWER's current space, NOT the doc's own space. But the seed effect below
+  // overwrites wk.shared.currentSpaceId to the doc's link space (`?sp=`) for preflight/room
+  // addressing, so by the time the doc is ready that value no longer reflects the viewer. Capture
+  // the viewer's real current space ONCE on first render — before the seed effect runs — resolving
+  // live currentSpaceId → cached localStorage → deploy default (the viewer resolution of
+  // standaloneFallbackSpace, WITHOUT the link-space override that preflightSpace applies). Lazy
+  // null-init so the resolution runs exactly once and is immune to the later seed mutation.
+  const viewerSpaceRef = useRef<string | null>(null)
+  if (viewerSpaceRef.current === null) {
+    viewerSpaceRef.current = standaloneFallbackSpace(wk.shared?.currentSpaceId)
+  }
+
   // Genuine defense in depth (second, independent path — NOT the only working one): the standalone
   // page mounts via the Layout early-return, before the app shell restores currentSpaceId from
   // localStorage, so any in-shell-shared logic the EditorShell touches would see an empty space. If —
@@ -384,6 +398,22 @@ export function StandaloneDocPage({
       cancelled = true
     }
   }, [docId, onSessionExpired, preflightSpace])
+
+  // XIN-1238 / XIN-1234: the standalone `/d/:docId` page never recorded a view, so a doc opened from
+  // a chat share link never surfaced in the opener's "最近查看". Mirror the in-shell entry
+  // (DocsHome.commitOpen): once the doc is READY, fire a single view ingest. It is written to the
+  // VIEWER's real current space (viewerSpaceRef), per the XIN-1237 write/read space contract — NOT
+  // the doc space the page seeds for addressing. Guarded by docId so React re-renders and
+  // strict-mode double-invocation record at most once per opened doc, matching the timing of the
+  // normal entry (on open success, not in a render loop). Fire-and-forget: recordDocView swallows
+  // every failure, so a failed / not-yet-deployed ingest never affects the open path.
+  const recordedDocRef = useRef<string | null>(null)
+  useEffect(() => {
+    if (phase.status !== 'ready' || !docId) return
+    if (recordedDocRef.current === docId) return
+    recordedDocRef.current = docId
+    void recordDocView(docId, { spaceId: viewerSpaceRef.current ?? undefined })
+  }, [phase.status, docId])
 
   useEffect(
     () => () => {
--- a/packages/docs/src/pages/StandaloneDocPage.tsx
+++ b/packages/docs/src/pages/StandaloneDocPage.tsx
@@ -209,6 +209,35 @@ export function standaloneFallbackSpace(currentSpaceId: string | undefined): str
 }
 
 /**
+ * The VIEWER's real current space, resolved from a genuine viewer signal ONLY: the live
+ * `currentSpaceId`, else the cached `currentSpaceId` localStorage key the shell persists. Returns
+ * '' when neither exists — deliberately WITHOUT standaloneFallbackSpace's DEFAULT_DOC_SPACE tail.
+ *
+ * This is the space recordDocView writes the "最近查看" ingest into (XIN-1237 write/read contract):
+ * the backend writes and reads the recent list by X-Space-Id = the viewer's current space, so the
+ * view MUST be recorded under the viewer's space, NOT the doc's link space (`?sp=`). Where the doc
+ * space is used (preflight/room addressing) the deploy default is a safe last resort, but for the
+ * view record it is NOT: recording into DEFAULT_DOC_SPACE when we cannot confirm the viewer is
+ * actually there would (a) write the view into a space the viewer isn't in — breaking the
+ * per-space isolation the recent list relies on — and (b) still never surface in the viewer's own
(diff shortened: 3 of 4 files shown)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #825 (octo-web)

## Summary

This PR adds view-ingest recording to the standalone `/d/:docId` page so that documents opened via share links surface in the viewer's "最近查看" (recently viewed) list. The core challenge is the XIN-1237 write/read space contract: the view must be written under the viewer's current space (not the doc's own space), which is non-trivial because the standalone page seeds `wk.shared.currentSpaceId` to the doc's link space for preflight addressing.

The implementation captures the viewer's real space in a `useRef` during render (before the seed effect mutates `currentSpaceId`), then fires `recordDocView` once when the doc reaches the `ready` phase. The `recordDocView` helper in `packages/docs/src/pages/docsApi.ts` gains an optional `opts.spaceId` parameter that sets an explicit `X-Space-Id` header, overriding the global interceptor. Both unit tests and E2E tests cover the cross-space happy path, idempotency, and terminal-preflight negative cases.

## Verification

- ✅ **Data flow — viewerSpaceRef capture timing**: `viewerSpaceRef` is lazy-initialized during the render body (`packages/docs/src/pages/StandaloneDocPage.tsx:322-324`), which executes before any `useEffect`. The seed effect that overwrites `wk.shared.currentSpaceId` with the doc's link space (`:340-356`) runs after render. The viewer's real space is therefore captured before mutation — correct ordering guaranteed by React's render-then-effect lifecycle.
- ✅ **Idempotency — client + server**: `recordedDocRef` (`:412-413`) prevents double-fire within a component lifecycle (including React 18 Strict Mode double-invocation). The server UPSERTs on `(uid, docId)`, so multi-tab opens are also idempotent.
- ✅ **Backward compatibility**: `recordDocView` signature change (`packages/docs/src/pages/docsApi.ts:216`) is additive — `opts` is optional, so the existing in-shell caller at `packages/docs/src/pages/DocsHome.tsx:1345` (`void recordDocView(docId)`) continues to work without changes.
- ✅ **Axios header merge**: The explicit `config.headers['X-Space-Id']` passed as the third argument to `apiClient().post()` (`packages/docs/src/pages/docsApi.ts:219`) correctly overrides the interceptor-injected value per Axios merge semantics.
- ✅ **Phase gating**: The record effect (`packages/docs/src/pages/StandaloneDocPage.tsx:411-416`) only fires when `phase.status === 'ready'`. Preflight failures (403/404/401) set phase to `'terminal'`, so no view is recorded for inaccessible docs — correct behavior verified by both unit and E2E negative tests.
- ✅ **C1 — Dual path parity**: In-shell path (`DocsHome.commitOpen`) calls `recordDocView(docId)` without `spaceId`, relying on the interceptor (which carries the viewer's live space). Standalone path passes `viewerSpaceRef.current` explicitly because the interceptor is seeded with the doc's space. Both paths semantically write to the viewer's current space — parity is correct.
- ✅ **C2 — Control-flow ordering**: Render-phase ref capture → seed effect → preflight effect → record effect (after ready). No nested reuse or ordering inversion across call sites.

## Findings

No P0/P1 issues. No correctness, security, build, or functional concerns identified.

### Nit — Redundant `?? undefined` narrowing (`packages/docs/src/pages/StandaloneDocPage.tsx:415`)

`viewerSpaceRef.current ?? undefined` is technically redundant: `standaloneFallbackSpace` always returns a non-empty string (falls through to `DEFAULT_DOC_SPACE`), so `viewerSpaceRef.current` is never null after the lazy init. The `?? undefined` does serve as a TypeScript narrowing hint (`string | null` → `string | undefined` to match `opts.spaceId?: string`), so it is harmless. Not a blocker — purely stylistic.

- **`standaloneFallbackSpace` resolution chain** (`packages/docs/src/pages/StandaloneDocPage.tsx:198-210`): live `currentSpaceId` → localStorage → `DEFAULT_DOC_SPACE`. Always returns a truthy string; empty-string input is treated as falsy and falls through correctly.
- **React Strict Mode**: Ref lazy-init runs on first render only (second render sees non-null ref). Effects double-invoke but `recordedDocRef` guard blocks duplicate API calls.
- **SSR safety**: `standaloneFallbackSpace` checks `typeof window !== 'undefined'` before touching localStorage. The standalone page only mounts after auth (Layout gate), so SSR is not a practical concern, but the guard is correct defensively.
- **Test coverage**: 4 unit tests (cross-space, live-space, idempotency, terminal-negative) + 2 E2E tests (cross-space with `X-Space-Id` header assertion, terminal negative). Coverage spans the XIN-1237 write/read contract and the idempotency guarantee.
- **No dangling state on unmount**: The record effect has no cleanup, but `void recordDocView(...)` is fire-and-forget — no state updates after unmount.

## Verdict: APPROVED

Well-structured fix with correct data-flow timing, sound idempotency guards at both client and server levels, and thorough test coverage. The `viewerSpaceRef` capture-before-mutation pattern is the right approach for this problem. No blocking issues.

OctoBoooot (dismissed):
## Review: record a view when a shared doc opens on the standalone page (#825)

**Verdict:** Approve
Now out of draft, so here's the formal gating review. A well-built, security-conscious view-recording fix on the standalone shared-doc page: IDOR-gated (only after authz preflight), dedup-guarded, correct viewer-space capture. Byte-verified. No blockers, no majors.

Code CI green (Build `success`, pr-title-lint / secret-scan / osv-scan / dependency-review pass). Risk tier: elevated (write-on-read on a shared-doc surface) — reviewed at a higher bar. size/M.

## Parity checked
The write is gated behind authz and can't leak/mis-record; the `recordDocView` signature change is additive:
- **IDOR-safe** — the view-record `useEffect` fires **only when `phase.status === 'ready'`** (StandaloneDocPage.tsx:411-412), and `ready` is set exclusively in the `getDoc` authz-preflight success path. A 403/404/401 doc never reaches `ready` → no view recorded for a doc the caller can't open. (The corresponding StandaloneDocPage 403 test asserting `viewCalls===0` covers this.)
- **Dedup-safe** — `recordedDocRef.current === docId` short-circuits before firing and is set before the call (:413-414), so StrictMode double-invoke / re-render / phase re-transition fire the ingest **at most once per docId**. Refs don't trigger re-render; no loop.
- **Correct viewer space (the actual fix)** — `viewerSpaceRef` is captured lazily on first render (:322-325) *before* the preflight effect overwrites `wk.shared.currentSpaceId` to the doc's link space, so it snapshots the **viewer's** space, not the doc space. `recordDocView(docId, { spaceId: viewerSpaceRef })` → `X-Space-Id` header (overrides the interceptor), so the view lands in the recipient's recent list per the XIN-1237 write/read-space contract. The cold-start fallback chain (live currentSpaceId → cached localStorage → deploy default) is sound for cross-space deep links.
- **No injection/leak** — `docId` `encodeURIComponent`'d; `spaceId` goes only into a header (not URL construction, no SSRF); POST body is empty (uid still server-derived, no client-sent identity); fire-and-forget (`void`, swallows errors — no leak-back). No XSS.

### Praise
- The security model is exactly right for write-on-read: the ingest is strictly downstream of the authz preflight (so it can't record a view the caller isn't entitled to), deduped by a ref so it can't amplify, and the viewer-space capture is carefully timed *before* the doc-space override — with the reasoning documented inline. That last point is subtle (the standalone page deliberately seeds `currentSpaceId` to the doc space for preflight, which would otherwise mis-attribute the view) and the lazy `viewerSpaceRef` snapshot handles it correctly. Strong test coverage (unit + e2e).

### Forward note (out of scope, not blocking)
- The FE `phase.status==='ready'` gate is defense-in-depth, not a substitute for the backend `POST /docs/:id/view` enforcing its own doc-role check. The backend endpoint isn't in these 4 files — worth confirming on the coordinated backend deploy the PR mentions.

### Net
A clean, security-conscious view-recording addition: IDOR-gated behind the authz preflight, dedup-guarded (at-most-once per doc), viewer-space captured correctly before the doc-space override, no injection/leak, CI green. No blockers, no majors. **APPROVE** @ `69ba20c4`. Credit @ploy-elison for the careful authz gating + viewer-space timing. size/M, no governance label — mergeable by a maintainer once a second reviewer approves, and worth pairing with the backend view-endpoint doc-role verify.

Jerry-Xin (dismissed):
APPROVE — no blockers (now Ready for review; byte-verified at head 69ba20c46fdd).

Recording a view when a shared doc opens on the standalone page is correctly gated and safe:
- **IDOR/authz safe** — the view-record effect fires only when `phase.status === 'ready'`, and `ready` is set exclusively inside the `getDoc(docId)` authorization preflight's success path. Any 403/404/401 reaches a terminal phase and never triggers the record, so a caller can't record views for docs they can't open (the 403 test asserts `viewCalls === 0`). The backend `POST /docs/:id/view` also derives the viewer from the session (no client-supplied identity).
- **No leak-back** — fire-and-forget, response ignored/voided, errors swallowed; no doc metadata read from the response.
- **Dedup / no write-amplification** — double-guarded (a `recordedDocRef === docId` short-circuit survives StrictMode double-invoke and re-renders; effect deps are just `[phase.status, docId]`; the server UPSERTs on `(uid, docId)`). No render loop; the ref doesn't trigger re-render. An 'at most once' test covers it.
- **Correct viewer/space** — viewer is server-derived (no owner/viewer swap); `X-Space-Id` deliberately carries the *viewer's* current space (not the doc link's `?sp=`) so the doc lands in the viewer's recent list — the actual intent, covered by a cross-space test.
- **Anonymous/injection** — the page mounts only with a token (401 preflight → login, no null-viewer record); `docId` is `encodeURIComponent`-ed and `spaceId` only flows into the `X-Space-Id` header, not URL construction. No SSRF/injection.

The +191 is mostly tests (~152 lines); the -3 is a benign `recordDocView` signature/body adjustment with no dropped guard.

🟡 Non-blocking / forward note: the front-end gate is not a substitute for the backend `POST /docs/:id/view` enforcing its own doc-role check — since the PR body notes this ships coordinated with a backend deploy, please confirm the server-side view endpoint authorizes the doc access independently (defense in depth), so a direct API call can't record views bypassing the client gate.

Clean, well-tested. Nice coverage on the authz-preflight-gated recording and the cross-space viewer semantics.

yujiawei (dismissed):
# Code Review — PR #825 (octo-web)

## Summary

The standalone `/d/:docId` share page previously never recorded a document view, so a doc opened from a chat share link never appeared in the opener's "recently viewed" list. This PR mirrors the in-shell `DocsHome.commitOpen` behavior: once the preflight reaches the `ready` phase, it fires a single `POST /docs/:id/view`, written to the **viewer's** current space via an explicit `X-Space-Id` header — not the doc's own link space (`?sp=`) that the page seeds for preflight/room addressing. `recordDocView` gains an optional `spaceId`, mirroring the existing `getDoc(opts)` shape.

The change is focused, well-documented, and correct. Verdict: **APPROVED** with two non-blocking suggestions below.

## What I verified

**Spec / scope** — The diff does exactly what the description states and nothing more: one view per opened doc, written to the viewer space, `recordDocView` extended with an optional header. No unrequested features, flags, or interface changes. No missing pieces relative to the stated fix.

**Viewer-space capture timing (the crux)** — `viewerSpaceRef` is populated with a lazy null-init during render, *before* the seed effect runs, so it captures `standaloneFallbackSpace(wk.shared?.currentSpaceId)` (live → cached `localStorage` → deploy default) untouched by the later `wk.shared.currentSpaceId = linkSpace` mutation. The seed effect only mutates the in-memory shared space, never `localStorage`, so the viewer's cached space survives for the read-back. This is the right resolution.

**Header-merge semantics** — Confirmed against `APIClient` (`packages/dmworkbase/src/Service/APIClient.ts:96`): the interceptor injects `X-Space-Id` only when the header is absent (`!config.headers["X-Space-Id"]`), so the explicit per-request header wins and the interceptor is fallback-only. The JSDoc claim holds.

**Idempotence / StrictMode** — `recordedDocRef` guards by `docId` value (not by mount), so the app-root `React.StrictMode` double-invoke and any re-render record at most once. The effect deps `[phase.status, docId]` are correct; the ref reads are intentionally excluded (repo has no `exhaustive-deps` rule and refs must not be deps anyway). The "at most once" and "not in a render loop" unit tests cover this.

**Terminal guard** — On a 403/404/etc. preflight, `phase.status` never reaches `ready`, so no view is recorded. Covered by both the unit and e2e "no view on terminal" tests.

**Tests** — `StandaloneDocPage.test.tsx` (47 tests) passes locally; assertions on `config.headers['X-Space-Id']` are non-vacuous (the mock apiClient records per-call `config`). Typecheck of the changed files is clean. The new cases cover cross-space viewer-space write, in-shell live-space write, idempotence, and the terminal-no-view path.

## Suggestions (non-blocking, P2)

1. **Defensive docId match in the record-view effect** (`StandaloneDocPage.tsx:411-416`). The effect fires on `phase.status === 'ready'` without confirming `phase.meta.docId === docId`. On this surface `docId` only changes via a full remount (open-in-new-page is `window.open(_blank)`; mention/link navigation goes through `window.location.assign` = full reload; the standalone page never performs an in-place `/d/A → /d/B` transition), so this is currently unreachable rather than a live bug. But if the page ever gains in-place doc switching, a stale `ready` from the previous doc could record a view for the new `docId` before its preflight resolves. A one-line guard (`if (phase.status !== 'ready' || phase.meta.docId !== docId) return`) would make the effect robust to that future and costs nothing today.

2. **Cold-viewer default-space fallback** (`StandaloneDocPage.tsx:324` → `standaloneFallbackSpace`). For a genuinely cold viewer with no live *and* no cached `currentSpaceId`, `viewerSpaceRef` resolves to `DEFAULT_DOC_SPACE` (`'demo'`), so a cross-space view would be written under `demo` and never surface in the viewer's real recent list. It's a graceful degradation of the new feature (fire-and-forget, no data loss or security impact — the server derives uid and UPSERTs on `(uid, docId)`), and it mirrors the preflight's own fallback, so I'd leave it as-is; flagging only so the edge is a conscious choice rather than an accident.

## Coverage notes

- E2E (`standalone-doc.spec.ts`) was **not executed** here (real-browser Playwright run); it was read for correctness and the new cases look sound (route-ordering caveat is handled — the `/view` handler is registered after the catch-all so it wins).
- Not independently exercised: the live backend write/read space contract (XIN-1237) — reviewed only against the in-repo assertions and the documented contract.
- Two pre-existing typecheck errors exist in files this PR does not touch (`members/sort.ts`, `dmworkbase/i18n/format.ts`); confirmed present on the base and unrelated to this change.

ploy-elison:
Follow-up commit (`7d3218f7`) addressing a cross-space regression found in device testing of this branch (XIN-1254).

**Symptom:** opening *someone else's* doc from a cross-space share link, then returning to the docs list, the doc still does not appear in 最近查看 — even though this branch already records the view under the viewer's space (`viewerSpaceRef`).

**Cause:** the seed effect overwrites the shared `currentSpaceId` with the doc's link space (`?sp=`) so the interceptor can address the editor's cross-space collab requests. That value was left in place after the page tore down. The 最近查看 read derives its space from the live `currentSpaceId` via the global interceptor, so on returning to the shell the read was scoped to the **doc** space while the view had been written to the **viewer** space — the same XIN-1234 write/read space split, reintroduced by the seed. It only bites a cross-space share (`?sp` ≠ the viewer's own space), which is why a same-space test passes but the cross-space device case fails.

**Fix:** restore the viewer's real current space (already captured in `viewerSpaceRef`) when the page unmounts, guarded to only revert the value we seeded. React runs an unmounted tree's effect cleanups before the newly-mounted tree's effects in the same commit, so the restore lands before DocsHome's recent-view request fires. Same-space and in-shell mounts are unaffected.

Two unit tests added (revert-on-teardown; no-clobber when the shell switched space meanwhile). Full docs suite green except 4 pre-existing `EditorShell.test.tsx` failures unrelated to this change. Re-review + the cross-space device check on :3000 still pending.

OctoBoooot (dismissed):
## Re-review (DELTA) #825 @ `7d3218f7` — restores viewer space on teardown (fixes the cross-space read/write split); security model intact → APPROVE

**Verdict:** Approve
My prior APPROVE was dismissed by a new commit that fixes a real cross-space bug: the standalone page seeded `currentSpaceId` to the doc's link space for collab addressing but left it there after teardown, so returning to the docs list scoped the "最近查看" *read* to the DOC space while the view was *written* under the VIEWER space — a cross-space share never surfaced in the opener's recent list. Byte-verified the teardown-restore fix is correct and the security model (IDOR/dedup/viewer-space) is intact. Re-registering APPROVE.

Code CI green (Build `success`, pr-title-lint / secret-scan / osv-scan / dependency-review pass). size/L.

## Parity checked
The fix closes the global-workspace-context restore gap; the view-recording security model is unchanged:
- **Teardown restore is correctly scoped** — the effect captures what it `seeded` and the cleanup restores `viewerSpaceRef.current` **only if `shared.currentSpaceId === seeded`** (StandaloneDocPage.tsx) — i.e. it reverts only the value it set, never clobbering a space the shell legitimately switched to meanwhile. This is exactly the "restore/scope the global mutation on unmount, don't blindly overwrite" discipline the footgun demands.
- **Read now matches write space** — restoring to the viewer's real space means the docs-list "最近查看" read is scoped to the viewer's space, matching where the view was recorded (`recordDocView(docId, { spaceId: viewerSpaceRef })`). The cross-space read/write split (the XIN-1234 variant the seed reintroduced) is closed.
- **Timing reasoning correct** — React runs an unmounting tree's effect cleanups before the newly-mounted tree's effects in the same commit, so the restore lands before DocsHome's recent-view request fires. Accurate.
- **Security model unchanged** — the IDOR gate (`phase.status !== 'ready'`, :432), the dedup (`recordedDocRef.current === docId`, :433), and the viewer-space-only `recordDocView` (:435) are all intact from the head I verified; `docsApi.ts` has no incremental change (the `spaceId`→`X-Space-Id` header is unchanged). No new injection/leak surface.
- **Test-locked** — new XIN-1254 test asserts the seed overwrites `currentSpaceId` to the doc space on mount and restores it to the viewer space on `unmount()`.

### Praise
- The fix is exactly right for a global-state-mutation-with-restore: it reverts *only the value it seeded* (guarding against clobbering a concurrent legitimate space switch), restores to the pre-seed viewer space so read/write spaces align, and reasons correctly about React's cleanup-before-mount ordering so the restore beats the recent-view read. Test-locked. This is the disciplined pattern for touching a module-global workspace context.

### Net
A correct, test-locked fix for the cross-space recent-view read/write split — restore-on-teardown scoped to only-what-we-seeded, read space now matching write space — with the IDOR/dedup/viewer-space security model from the prior head fully intact. Re-registering my **APPROVE** @ `7d3218f7`. No blockers, majors, or nits. Credit @ploy-elison for the properly-scoped restore + the teardown regression test. (Peers dismissed by this commit should re-anchor; the forward note stands — the backend `POST /docs/:id/view` should enforce its own doc-role check on the coordinated deploy.)

Jerry-Xin (dismissed):
**Re-review — APPROVE** (re-anchored from my prior APPROVE at 69ba20c). The new head advances one linear commit (ahead 1 / behind 0, not a rebase) and touches only 2 files. Byte-verified the delta.

**What changed (delta 69ba20c → 7d3218f):**
- `StandaloneDocPage.tsx` (+28/−8): the space-*seeding* effect was refactored to capture the value it seeds into a local `seeded`, then adds a **cleanup that restores the viewer's real space on teardown**. When a cross-space share (`?sp=` ≠ viewer's space) is opened, the seed overwrites `shared.currentSpaceId` with the doc space for collab addressing; leaving that in place after unmount would scope the docs-list "最近查看" read to the doc space, hiding a view that was recorded under the viewer's space (`viewerSpaceRef`) — the XIN-1234 write/read space split. The cleanup reverts to `viewerSpaceRef.current` **only if the current value is still the one we seeded** (`if (shared.currentSpaceId === seeded)`), so a genuine mid-session space switch is never clobbered. The 8 removed lines are purely the old inline seed assignments (`shared.currentSpaceId = linkSpace` + the localStorage fallback branch) — refactored into the `seeded` path, not deleted. No security guard or scope check was removed; the "never overwrite a real live space" early-return and the localStorage try/catch degrade are intact.
- `StandaloneDocPage.test.tsx` (+50): two real tests — (1) cross-space teardown restores `space-viewer` after seeding `space-doc`; (2) the restore is a no-op when the shell switched to a third space meanwhile. Meaningful assertions, not superficial.

**Invariants re-confirmed unchanged (byte-identical across heads):**
- View-record effect (`recordedDocRef` guard + `recordDocView(docId, { spaceId: viewerSpaceRef.current })`) is byte-for-byte identical — **once-per-open dedup** intact, `docId` in deps + ref guard means no re-render/strict-mode double-record and no stale-closure recording the wrong doc.
- **IDOR-gating** unchanged — the view POST still targets the correct `docId` with the viewer's resolved space (not a URL/param-forgeable value); server-side membership/read-access remains the real enforcement. The pre-existing `d_forbidden/view` never-called test is untouched.
- **Path safety** (encodeURIComponent) and **best-effort degrade** (`void recordDocView`, failures swallowed) unchanged.

**CI:** Build pass, osv-scan pass, secret-scan pass, code-review pass; check-sprint / scan-scheduled / satisfy-code-review skipping (not red). Mergeable, non-draft.

Clean, well-scoped companion fix that closes the write/read space gap without regressing the view-record path. LGTM.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #825 (octo-web)

## Summary

This PR fixes XIN-1234 / XIN-1238: the standalone `/d/:docId` page never recorded a view, so a document opened via a chat share link never surfaced in the opener's "最近查看" (recently viewed). Two coordinated changes solve this: (1) a new `recordDocView` call fires once the doc reaches READY, addressed to the viewer's real current space (not the doc's own space); (2) the seed effect that overwrites `wk.shared.currentSpaceId` with the doc's link space now restores the viewer's original space on teardown (XIN-1254), so navigating back to the docs list doesn't leave the recent-view read scoped to the wrong space. The implementation is careful and well-tested.

## Verification

Static analysis only at head `7d3218f74bda045b2cc7d78743e451960b0aa7e9`; build and tests not executed in this environment.

- ✅ **viewerSpaceRef capture** — lazy null-init on first render (`StandaloneDocPage.tsx:323-324`) resolves via `standaloneFallbackSpace` before the seed effect mutates `currentSpaceId`. On cold deep link: localStorage → viewer's real space. On in-shell mount: live `currentSpaceId` → viewer's live space. Correct for both paths.
- ✅ **Seed effect cleanup (XIN-1254)** — `StandaloneDocPage.tsx:367-373`: the cleanup only reverts `currentSpaceId` when it still equals the value we seeded (`if (shared.currentSpaceId === seeded)`). A genuine space switch by the shell is never clobbered. Verified by unit test at `StandaloneDocPage.test.tsx:988-1014`.
- ✅ **recordDocView idempotency** — `recordedDocRef` guard (`StandaloneDocPage.tsx:431-434`) prevents double-fire in React Strict Mode and on re-renders. Fires only when `phase.status === 'ready'`. Unit test at `StandaloneDocPage.test.tsx:946-966` confirms single-call invariant.
- ✅ **recordDocView API helper** — `docsApi.ts:218-224`: uses `encodeURIComponent(docId)` (path-traversal safe), explicit `X-Space-Id` header merged over the interceptor by axios, fire-and-forget with catch-all. Parity with in-shell `DocsHome.commitOpen` (`DocsHome.tsx:1345`) confirmed — both paths send the viewer's space.
- ✅ **Terminal preflight → no view** — the record-view effect short-circuits on `phase.status !== 'ready'`. Unit test (`StandaloneDocPage.test.tsx:968-984`) and E2E test (`standalone-doc.spec.ts:220-236`) both verify 403 → zero view calls.
- ✅ **Data flow end-to-end** — viewerSpaceRef (captured render-time) → `recordDocView({ spaceId })` → explicit `X-Space-Id` header → backend writes view under viewer's space → `GET /docs/recent` reads back by same space → doc surfaces in 最近查看. Write/read space contract holds.

## Findings

No P0/P1 issues. No correctness, security, build, or functional concerns identified.

- **Dual-path parity (C1)**: standalone `recordDocView(docId, { spaceId })` and in-shell `recordDocView(docId)` (relying on interceptor) both write the view under the viewer's real space. The explicit `spaceId` parameter is necessary because the standalone seed effect overwrites `currentSpaceId` to the doc's link space before the record-view effect fires.
- **Control-flow ordering (C2)**: `viewerSpaceRef` is captured synchronously during render, before any effect. The seed effect runs after render. The record-view effect runs only after preflight succeeds (phase → ready). No ordering hazard.
- **Seed effect behavioral change**: the old code had separate early-return paths for `linkSpace` vs localStorage cache with no cleanup. The new code unifies through `seeded` and adds cleanup for both. When the cache is the source (same-space), `seeded === viewerSpaceRef.current`, so the cleanup restore is a no-op — no behavioral regression.
- **Authorization boundary (C3)**: `recordDocView` goes through the same authenticated `apiClient`. No new endpoint exposure. The `X-Space-Id` header is additive routing, not a privilege escalation.
- **E2E test quality**: both E2E tests use `page.addInitScript` to seed localStorage (matching the cold-deep-link scenario) and `page.route` to intercept the view call. The `X-Space-Id` header assertion is the correct contract check.
- **React Strict Mode**: refs persist across the mount→unmount→remount cycle, so `recordedDocRef` prevents double-fire and `viewerSpaceRef` captures exactly once. The seed cleanup fires during the strict-mode unmount and re-seeds on remount — correct behavior.

## Data flow tracing

|---|---|---|
| `viewerSpaceRef.current` | `standaloneFallbackSpace(wk.shared?.currentSpaceId)` at render time → localStorage `currentSpaceId` on cold deep link, live `currentSpaceId` on in-shell mount | ✅ Captured before seed mutation |
| `recordDocView` spaceId | `viewerSpaceRef.current ?? undefined` | ✅ Explicit header overrides interceptor |
| Seed cleanup restore value | `viewerSpaceRef.current ?? ''` | ✅ Never null at cleanup time (set during first render) |
| `seeded` value | `standaloneLinkSpace()` (from `?sp=`) or localStorage cache | ✅ Guard prevents clobbering genuine space switches |
| `recordedDocRef` idempotency guard | Set to `docId` on first fire, checked on re-entry | ✅ Prevents double-fire in strict mode and re-renders |

## Blind spot checklist

- **C1 (dual-path parity)**: Clear. Standalone and in-shell paths both record views under the viewer's space.
- **C2 (control-flow ordering)**: Clear. Render-time ref capture → post-render seed → post-preflight record. No ordering hazard.
- **C3 (authorization boundary)**: Clear. No new endpoints; same authenticated client; `X-Space-Id` is routing, not privilege.
- **C4 (authorization lifecycle)**: N/A — no auth changes.
- **C5 (build/runtime)**: N/A — no build/config changes.
- **C6 (governance/docs)**: N/A — no governance changes.

## Cross-round blocker re-check (R6)

## Verdict: APPROVED

Well-engineered fix with thoughtful defense-in-depth (viewerSpaceRef capture before seed mutation, guarded teardown restore, idempotent recordDocView). The write/read space contract for 最近查看 is correctly maintained across both the standalone and in-shell paths. Comprehensive test coverage at both unit and E2E levels validates the key invariants.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; clean implementation with correct data flow, idempotency guards, and teardown safety.

yujiawei (dismissed):
# Code Review — PR #825 (octo-web)

Reviewed at head SHA `7d3218f74bda045b2cc7d78743e451960b0aa7e9` against merge-base `feb5282f`. Independent review of the current head (prior approvals were on earlier, now-dismissed heads and are not inherited).

## Summary

The change makes `StandaloneDocPage` (the `/d/:docId` share surface) record a doc view when the doc reaches `ready`, mirroring the in-shell `DocsHome.commitOpen` entry, so a doc opened from a cross-space share link surfaces in the opener's "recently viewed". The core difficulty is space scoping: the standalone page deliberately seeds `wk.shared.currentSpaceId` to the **doc's** link space (`?sp=`) for preflight/collab addressing, but the view must be written under the **viewer's** current space (the space the recent list reads back by). The PR resolves this by capturing the viewer space once, up front, and passing it as an explicit `X-Space-Id` header.

## 1. Spec compliance

**Spec: ✅**

The diff implements exactly the stated fix, nothing more:
- Records a single view on `ready`, viewer-space-scoped (`StandaloneDocPage.tsx:431-436`).
- Adds an optional `spaceId` to `recordDocView`, mirroring the existing `getDoc` opts shape (`docsApi.ts:216-224`) — no over-building.
- Restores the seeded space on teardown so the shell's recent-view read is not left scoped to the doc space (`StandaloneDocPage.tsx:374-378`).
- No unrelated features, flags, or fields introduced.

## 2. Code quality

**Quality: Approved**

Verified against the surrounding code:

- **Viewer-space capture is correctly ordered.** `viewerSpaceRef` is populated with a lazy null-init **during render** (`StandaloneDocPage.tsx:322-325`), which runs before any effect. The seed effect that overwrites `wk.shared.currentSpaceId` runs post-commit, so the captured viewer space is immune to that later mutation. `standaloneFallbackSpace` never consults `?sp`, so the viewer space is genuinely isolated from the doc space. Correct.
- **Explicit header wins.** The `X-Space-Id` merge claim holds: the global interceptor at `packages/dmworkbase/src/Service/APIClient.ts:96` only injects `currentSpaceId` when `config.headers["X-Space-Id"]` is absent, so the explicit per-request header takes precedence. Correct.
- **Idempotency.** `recordedDocRef` guards against re-fire on re-render and StrictMode double-invoke; the effect keys on `docId` so it records at most once per opened doc (`StandaloneDocPage.tsx:430-436`). The unit test explicitly asserts the single-call invariant.
- **Teardown restore is guarded.** The cleanup only reverts the value it seeded (`shared.currentSpaceId === seeded`), so a legitimate space switch mid-session is not clobbered (`StandaloneDocPage.tsx:375-377`) — and there is a test for exactly that.
- **Fire-and-forget is preserved.** `recordDocView` swallows all failures, so a not-yet-deployed ingest never affects the open path — consistent with the existing contract.
- Test coverage is strong: unit tests cover cross-space, in-shell, idempotency, terminal-not-recorded, and teardown restore/no-clobber; the e2e spec adds real-browser coverage of the viewer-space header and the 403-no-view case.

### Non-blocking notes (nits)

- **`recordDocView` empty-string strip (`docsApi.ts:218`).** `opts?.spaceId ? {…} : undefined` drops an empty-string `spaceId`, letting the interceptor fill in. This is consistent with the codebase-wide convention that `''` means "no space" (`APIClient.ts:96`: `spaceId && spaceId !== ""`), and in practice `viewerSpaceRef` resolves through `standaloneFallbackSpace`, which always returns a non-empty value (falling back to `DEFAULT_DOC_SPACE`). So there is no live path that reaches this branch with `''`. Worth a one-line comment noting the `''` convention if you touch this again, but not a defect.
- **SPA re-navigation between two standalone docs.** The seed effect depends on `[wk.shared]`, not on the parsed link space, so an in-app `/d/doc1` → `/d/doc2` transition without a remount would not re-seed. In practice the standalone page mounts via the host Layout early-return and each deep link is a fresh mount (the `key`ed EditorShell/BoardSession and `docId`-keyed effects confirm the per-doc mount model), so this is not reachable today. Flagging only as a latent assumption to keep in mind if standalone-to-standalone soft routing is ever added.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No P0/P1 issues. The two nits above are advisory and do not block merge.

## 4. Coverage note

- Backend contract for `POST /docs/:id/view` (that it writes and the recent list reads by `X-Space-Id`, and that a missing header falls to the interceptor value) is asserted by the PR's prose and tests but was not verified against backend source here.
- The teardown-restore-before-next-effect ordering relies on React running an unmounting tree's cleanup before the newly-mounted tree's effects within the same commit; this holds for soft in-app navigation but not for a full page reload (where the shell re-bootstraps `currentSpaceId` from storage anyway, so the doc space is not persisted).
- A second automated reviewer leg was attempted but was unavailable this run (repeated upstream gateway errors); this verdict rests on the manual review plus one advisory leg, whose two proposed P1s were independently checked and found not to be live defects (empty/`undefined` space cannot reach the write path given `standaloneFallbackSpace`'s non-empty contract).

OctoBoooot (approved):
## Re-review (DELTA) #825 @ `1d49582b` — records the standalone-doc view under the VIEWER's current space (not the doc link space), omitting X-Space-Id when no viewer signal exists; correct per-space isolation, well-tested → APPROVE

**Verdict:** Approve
Re-approving. My prior APPROVE (`7d3218f7`) was dismissed by a new commit that sharpens the exact space-scoping axis it touched: the standalone `/d/:docId` view-record now writes under the **viewer's current space** (the space the recent-view list reads back by), never the doc's link space (`?sp=`), and **omits** X-Space-Id entirely when no genuine viewer signal exists rather than forcing the deploy default. Byte-verified correct and regression-tested. CI green.

## Parity checked
The delta is `StandaloneDocPage.tsx` +43/-9 + its test (+70); linear (`ahead:1 behind:0`). The prior fix's properties are preserved and the space resolution is corrected:
- **View-record space = viewer signal only.** New `viewerCurrentSpace(currentSpaceId)` resolves live `currentSpaceId` → cached `localStorage.currentSpaceId` → `''`. It deliberately **drops the `DEFAULT_DOC_SPACE` tail** that `standaloneFallbackSpace` carries, and never sources from the link `?sp=`. localStorage read is try/caught (private-mode safe).
- **Omit, don't force, on no signal.** The record call is now `recordDocView(docId, viewerSpace ? { spaceId: viewerSpace } : undefined)` — when there's no viewer signal it omits the explicit X-Space-Id and lets the global interceptor decide (matching the in-shell entry). This is the isolation-preserving choice: forcing the deploy-default space would write the view into a space the viewer isn't in (breaking per-space isolation) and still never surface in their own recent list. This is exactly the write/read space contract done right.
- **Prior fix intact** — the lazy null-init `viewerSpaceRef` (resolved once, immune to the later seed effect that overwrites `currentSpaceId` with the link space), the `docId`-guard (React re-render + strict-mode double-invoke → record at most once per doc), and the fire-and-forget `recordDocView` (swallows failures, never affects the open path) are all preserved.
- **Tests assert the contract** — `viewerCurrentSpace` (live-preferred / cached-fallback / `''`-not-default on no signal); same-space share records with `X-Space-Id: space-team` and surfaces in recent-view; no-signal path omits the explicit header. Behavior-level on exactly the axis that matters.
- CI green (Build `success`, secret-scan / osv-scan / dependency-review / pr-title-lint pass; check-sprint skipped = governance).

### Praise
- Choosing to **omit** the X-Space-Id when the viewer's space is unknowable — rather than falling through to the deploy default — is the subtle-but-correct call. A naive "always send a space" would silently record the view into the wrong space, breaking the per-space isolation the recent-view list depends on (and my own notes flag exactly this shared-space-scoping footgun). Capturing the viewer space once before the seed effect overwrites `currentSpaceId` with the link space, and dropping the `DEFAULT_DOC_SPACE` tail specifically for the write path (while it stays a safe default for preflight/room addressing), shows the write/read contract was reasoned through, not pattern-matched.

### Net
A correct refinement of the standalone-doc view-record: it writes under the viewer's current space (what recent-view reads by), omits the header when no viewer signal exists to avoid mis-recording into a wrong space, and preserves the once-only/docId-guarded/fire-and-forget properties from the approved version — with regression tests for the same-space, no-signal, and helper paths. Code CI green. Re-registering my **APPROVE** @ `1d49582b`. Peers dismissed by this commit should re-anchor.

Jerry-Xin (approved):
**APPROVE** — delta cleanly refines the view-record space contract; view-recording is correct (fires once, guarded, failure-swallowed), docsApi is authz-safe with encoded id, no XSS surface, Build green.

This is a delta re-review. Prior reviews were dismissed when head moved `7d3218f7` → `1d49582b`. The delta is a clean forward (ahead 1, behind 0 — no rebase/merge-base drift), exactly 1 commit touching 2 files.

### 🔎 The delta (`7d3218f7` → `1d49582b`)
The commit refines *which space* the standalone page records the view under, replacing the prior `standaloneFallbackSpace(...)` resolution with a new dedicated `viewerCurrentSpace(...)` helper:
- `StandaloneDocPage.tsx:222` — new `viewerCurrentSpace(currentSpaceId)`: returns live `currentSpaceId` → cached `localStorage.currentSpaceId` → `''`. Deliberately **omits** `standaloneFallbackSpace`'s `DEFAULT_DOC_SPACE` tail; wraps `localStorage` in try/catch (private-mode safe).
- `StandaloneDocPage.tsx:353-354` — `viewerSpaceRef` now seeded from `viewerCurrentSpace(...)` (was `standaloneFallbackSpace(...)`), still lazy null-init so it resolves exactly once before the seed effect mutates `currentSpaceId`.
- `StandaloneDocPage.tsx:468-469` — the record call now passes `{ spaceId }` **only when a real viewer signal exists**; empty → `recordDocView(docId, undefined)`, letting the global interceptor decide, matching the in-shell entry.

This is a correctness fix, not a cosmetic one: recording under the deploy-default space when the viewer's space is unknown would (a) write the view into a space the viewer isn't in and (b) still never surface in the viewer's own recent list. The delta introduces no new issue.

### View-recording correctness (`StandaloneDocPage.tsx:463-470`)
- **Once**: `recordedDocRef` guards on `docId`; effect deps `[phase.status, docId]`; ref set before the async call. Immune to re-render / strict-mode double-invoke and to the later seed mutation. No infinite loop.
- **Guarded**: early-returns unless `phase.status === 'ready'` and `docId` truthy — never fires on loading/terminal/forbidden/not-found/empty-id states.
- **Failure-handled**: `recordDocView` swallows every failure internally (`docsApi.ts:212-219`) and is `void`-ed here — no unhandled rejection, never blocks the open path.

### docsApi.ts (`recordDocView`)
- Server-authorized like the rest — goes through `apiClient()` (`/api/v1/`), token injected by the global interceptor; no auth code here.
- URL-injection safe: `encodeURIComponent(docId)` on the path.
- No body (uid derived server-side); `X-Space-Id` is the viewer's own resolved space, not a cross-user/untrusted param. Server UPSERTs on `(uid,docId)` → idempotent.

### Render XSS
No `dangerouslySetInnerHTML` / `innerHTML` / `iframe` / `srcdoc` anywhere in `StandaloneDocPage.tsx`. No doc/user-derived string is injected as raw HTML.

### Tests
Real assertions covering every branch:
- fires once, cross-space (view under `space-viewer`, preflight under `space-doc`)
- same-space acceptance flow
- no-signal → header omitted (`toBeUndefined`)
- live-space in-shell mount
- idempotent (exactly 1 POST after a re-render tick)
- **not** recorded on terminal (403) preflight
- teardown restores the viewer space so the recent-view read matches
- `viewerCurrentSpace` unit tests (live/cached/empty, no deploy-default tail)
e2e `standalone-doc.spec.ts` adds the end-to-end share-open coverage.

### CI
Build **pass** (50s). osv-scan, dependency-review, secret-scan, history, pr-title-lint all pass. code-review pending (queued), check-sprint/scan skipped. No blockers.

boris-clark (approved):
Re-reviewed the correction commit at head 1d49582b (verified against the actual code, not the commit message).

**Space source is correct.** `viewerCurrentSpace()` resolves the viewer's space from a genuine viewer signal only — live `currentSpaceId`, else the cached `currentSpaceId` localStorage key — and deliberately drops `standaloneFallbackSpace`'s `DEFAULT_DOC_SPACE` tail. It is captured once at first render into `viewerSpaceRef` (lazy null-init) *before* the seed effect overwrites `shared.currentSpaceId` with the doc's link `?sp=` for preflight/room addressing. `recordDocView` passes that as an explicit `X-Space-Id`, never the link space. Confirmed the interceptor does not clobber it: `APIClient.ts` injects the fallback only under `!config.headers["X-Space-Id"]`, so the explicit per-request header wins.

**Write/read consistency.** The view is written under the viewer's current space, and the teardown restore (7d3218f7) reverts the seeded doc space back to `viewerSpaceRef` on unmount, so DocsHome's `最近查看` read — scoped to the live `currentSpaceId` via the interceptor — lands on the same space the view was written to. The revert is guarded to only touch the value we seeded, so a genuine space switch mid-session is left alone.

**Per-space isolation preserved.** The recent read is not widened cross-space. When there is no viewer signal at all, the record omits the explicit header rather than forcing `DEFAULT_DOC_SPACE` — avoiding writing a view into a space the viewer isn't in.

**No regression.** The direct standalone path (XIN-1254 preflight `?sp` addressing), the initial view-record commit (69ba20c4), and the teardown restore (7d3218f7) all remain coherent; the in-shell `recordDocView(docId)` caller is unaffected (new arg is optional).

**Tests.** Standalone suite passes (54/54) with explicit assertions for viewer-space-not-doc-space, same-space, no-signal-omits-header, live-space, idempotent-once, no-record-on-terminal, and teardown restore/no-op. Full `pages` suite green (133/133; the undici WebSocket noise originates in the untouched DocsHome test env). Typecheck clean for the changed files (the two remaining errors are pre-existing in unrelated files).

Frontend contract matches backend #89 (caf5b4f): write and recent-read both keyed on the viewer's `X-Space-Id`.

LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Data flow tracing confirms viewerSpaceRef captures the viewer's real space before seed mutation, recordDocView backward-compatible API extension, correct dedup, and safe teardown. Comprehensive test coverage across cross-space, same-space, and no-signal scenarios.

yujiawei (approved):
# Code Review — PR #825 (octo-web)

Reviewed at head SHA `1d49582b2271b4fc41a7e829f4cf719f15707452` against `main` (merge-base `feb5282f`). Verified locally: `packages/docs` unit suite **54/54 pass** (incl. the new `StandaloneDocPage.test.tsx` view-record cases), touched files typecheck clean (the two `tsc` errors reported are pre-existing, in `dmworkbase/src/i18n/format.ts` and `src/members/sort.ts`, untouched by this PR).

## 1. Spec compliance

**Spec: ✅**

The PR does exactly what the description states and nothing more:

- **Missing work:** none. `StandaloneDocPage` now records a single view once the doc reaches `ready` (`StandaloneDocPage.tsx:463-470`), mirroring the in-shell `DocsHome.commitOpen` (`DocsHome.tsx:1345`). `recordDocView` gains the optional `spaceId` (`docsApi.ts:216-224`).
- **Extra/out-of-scope work:** none. The change is confined to the view-record path; the pre-existing preflight/`?sp=` addressing and the copy-link logic are untouched.
- **Deviation:** none. The stated space contract (write and read the recent list by the *viewer's* current space, via `X-Space-Id`) is implemented consistently with the existing global interceptor.

## 2. Code Quality

**Quality: Approved**

The tricky part of this change — writing the view under the viewer's space while the page deliberately seeds `currentSpaceId` to the *doc's* space for collab addressing — is handled correctly:

- **Viewer-space capture is race-free.** `viewerSpaceRef` is resolved lazily during render (`StandaloneDocPage.tsx:352-355`), which runs *before* the seed effect mutates `wk.shared.currentSpaceId` (effects run after render). So the recorded view always uses the pre-seed viewer space, exactly as intended.
- **Interceptor contract verified.** The explicit `X-Space-Id` set here genuinely reaches the wire and is not clobbered: the global request interceptor only injects its fallback `if (self.config.spaceIdCallback && !config.headers!["X-Space-Id"])` (`packages/dmworkbase/src/Service/APIClient.ts:96`), i.e. explicit-header-wins. There is even a dedicated regression guard for this (`APIClient.headers.test.ts`).
- **Teardown restore ordering is sound.** The seed effect's cleanup restores the viewer space on unmount (`StandaloneDocPage.tsx:404-408`) and only reverts the value it itself seeded (`if (shared.currentSpaceId === seeded)`), so a legitimate space switch during the page's life is not clobbered — covered by the XIN-1254 tests.
- **Idempotency is correct.** `recordedDocRef` guards against React re-renders and StrictMode double-invoke (`StandaloneDocPage.tsx:463-470`); the "records at most once / never in a render loop" test confirms it.
- **No-viewer-signal safety.** When neither a live nor cached viewer space exists, `viewerCurrentSpace` returns `''` (no `DEFAULT_DOC_SPACE` tail) and the caller omits the explicit header rather than forcing the deploy default — the right call, and tested.
- **Fire-and-forget** matches the in-shell entry: `recordDocView` swallows all failures, so a not-yet-deployed ingest never disrupts open. The `POST` sends no body — this is the pre-existing, in-production contract (uid derived server-side, UPSERT on `(uid,docId)`); the PR does not change it.

### Non-blocking suggestions (P2 / nits)

- **P2 — `StandaloneDocPage.tsx:372` seed effect deps are `[wk.shared]`, not `[docId]`.** If the app ever navigated *between* two `/d/:docId` links via client-side routing without remounting, the seed (and `viewerSpaceRef`, `recordedDocRef`) would carry stale state. In the current architecture this cannot happen — the standalone page is mounted through an imperative `window.location.pathname` early-return in `apps/web/src/Layout/index.tsx:420-429` (no in-app nav between standalone docs; recipients arrive from external chat links, one doc per page load). So this is defensive only, and the pattern predates this PR. Worth a one-line comment noting the remount assumption if you want to harden it.
- **P2 — empty `POST` body.** A bodyless `POST /docs/:id/view` relies on the backend tolerating no payload. This is unchanged from the existing in-shell call and is already in production, so it's not a regression — flagging only for awareness if backend body-parsing strictness ever changes.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. No P0/P1 issues. The two P2 items are defensive/pre-existing and do not block merge.

## 4. Advisory-leg coverage (audit trail)

Two independent advisory legs were run:

- **Long-context cross-file leg:** returned 3 findings. Two were assessed against code the leg could not see and are **refuted**: (a) "the global interceptor will clobber the explicit `X-Space-Id`" — the interceptor explicitly guards `!config.headers["X-Space-Id"]`, so the explicit header wins (`APIClient.ts:96`); (b) "empty `POST` body may 400/415" — this is the pre-existing production contract, unchanged. Its one surviving point (seed-effect `docId` dep) is captured as the P2 above.
- **Line-level adversarial leg:** **absent** — the tool failed on a gateway-side configuration error (`x-ms-oai-image-generation-deployment`) across 3 retries. This is a tooling failure, not a signal of code health, and is explicitly **not** treated as silent approval.
- **Reviewer-supplied blind-spot note (compensating for the absent leg):** not verifiable from the diff — (1) backend behavior for a bodyless `POST` and for a view written under an empty/absent `X-Space-Id`; (2) that the recent-list *read* is in fact scoped by the live `currentSpaceId` on return to the list on the deployed backend (the write/read symmetry the fix depends on); (3) real-device confirmation of the "open cross-space link → back → 最近查看 shows it" flow. Unit + e2e cover the frontend contract; end-to-end space symmetry depends on the backend deploy the PR description notes is being coordinated.

PR #816, opened 2026-07-16
style(interactive-card): differentiate approval buttons (primary/secondary)

an9xyz:
## Summary

Give approval-card action buttons a primary/secondary distinction. Approval cards (允许/拒绝) previously rendered as two identical accent pills; this adds `.ac-pushButton.style-positive` (primary, brand-solid) and `.style-destructive` (secondary, neutral ghost) so 允许 reads as the primary action and 拒绝 as the secondary one.

Pairs with the octo-server template now emitting `ActionStyle` (`Mininglamp-OSS/octo-server` branch `claude/style-reasoning-card-design-1wigej`).

## Related Issue

N/A

## Changes

- `packages/dmworkbase/src/Messages/InteractiveCard/index.css`: add `.ac-pushButton.style-positive` (brand-primary solid fill) and `.style-destructive` (transparent ghost) plus hover states, scoped under `.wk-interactive-card-sdk`.

## Architecture / Module Boundary

- Affected module(s): `dmworkbase` — `Messages/InteractiveCard`
- New or changed user-visible entry point: no (visual refinement of the existing interactive/approval card)
- Shared layer touched: Messages
- If shared code changed, impact scope: only AdaptiveCards buttons that declare `style: positive|destructive` (currently just the server-built approval card). The base `.ac-pushButton` and every other card are untouched — zero side effects.
- Duplicate entry point checked: not applicable (CSS-only, no new component)

## Testing

- Rendered the real approval-card JSON (from `pkg/cardtmpl.BuildApprovalRequestCard`) through the real AdaptiveCards 3.0.6 SDK + this stylesheet and screenshotted: approve = primary solid, deny = secondary ghost. Confirmed the SDK emits the exact `.ac-pushButton.style-positive` / `.style-destructive` classes these rules target.
- CSS-only change; no component logic touched.

- [ ] Unit tests added/updated
- [x] Manually verified

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] PR description is in English
- [ ] Added tests for my changes
- [ ] Updated documentation
- [x] Ran `pnpm i18n:check` or confirmed the change does not affect UI copy
- [x] Confirmed the change follows module ownership and does not add duplicate user-visible entry points
- [x] Described impact scope when shared components, messages, bridge, or services are changed
- [x] Followed commit message conventions (Conventional Commits)

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01E2FHToMTWZnMq6cbQyTjqg

---
[...]

PR #827, opened 2026-07-16
style(interactive-card): render Action.OpenUrl as a hyperlink

an9xyz:
## Summary

Interactive-card `Action.OpenUrl` buttons (查看详情, 在 GitLab 查看, …) rendered as the base accent **pill**, which out-shouts the primary/secondary decision buttons on approval cards and reads wrong for what is really a *navigation* action.

AdaptiveCards marks `Action.OpenUrl` buttons with `role="link"` (Submit / ToggleVisibility / CopyToClipboard are `role="button"`), so this targets `.ac-pushButton[role="link"]` and renders it as a classic underlined **hyperlink** — link colour + underline, no button chrome. Precise, CSS-only, and semantic.

Before → after (approval card): 查看详情 pill competing with 允许/拒绝 → a quiet link above the decision buttons.

## Related Issue

N/A

## Changes

- `packages/dmworkbase/src/Messages/InteractiveCard/index.css`: add `.ac-pushButton[role="link"]` (+ `:hover` / `:focus-visible`) rendering OpenUrl buttons as an underlined link (accent colour, `--wk-color-accent-hover` on hover, focus outline for a11y).
- `packages/dmworkbase/src/Messages/InteractiveCard/__tests__/renderOctoCard.test.tsx`: add a render test locking the SDK contract — `Action.OpenUrl` → `role="link"`, `Action.Submit` → `role="button"`.

- Affected module(s): `dmworkbase` — `Messages/InteractiveCard`
- New or changed user-visible entry point: no (visual refinement of existing card buttons)
- Shared layer touched: Messages
- If shared code changed, impact scope: **only** `Action.OpenUrl` buttons (`role="link"`). Everything else is `role="button"` and untouched — audited against every `.ac-pushButton` rule in `index.css`: decision buttons (`.style-positive`/`.style-destructive`), the agent-progress **展开推理** toggle (`ToggleVisibility`, and a more-specific selector), the table copy button (different class), and `CopyToClipboard`. No regression to any intentional styling.
- Duplicate entry point checked: not applicable (CSS + test only, no new component)

## Testing

- New render test asserts the `role` contract (`link` for OpenUrl, `button` for Submit) the CSS keys off.
- Verified through the real AdaptiveCards 3.0.6 SDK + this stylesheet across four card shapes — approval (查看详情 link, 允许/拒绝 buttons), custom-action approval (执行/取消 stay buttons, **not** links), agent-progress reasoning (展开推理 keeps its bordered toggle), and pipeline (在 GitLab 查看 link). Computed-style assertions confirm the reasoning toggle stays `role="button"` with its border and no underline.
- CSS + test only; no component logic touched.

- [x] Manually verified

## Checklist

- [ ] Updated documentation

---

https://claude.ai/code/session_01E2FHToMTWZnMq6cbQyTjqg

---
_Generated by [Claude Code](https://claude.ai/code/session_01E2FHToMTWZnMq6cbQyTjqg)_

### DIFF
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/renderOctoCard.test.tsx
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/renderOctoCard.test.tsx
@@ -137,6 +137,36 @@ describe("renderOctoCard", () => {
     target.remove();
   });
 
+  it("Action.OpenUrl 渲染出 role=link、Submit 渲染出 role=button（查看详情超链样式依赖此 SDK 契约）", () => {
+    const target = mountTarget();
+    renderOctoCard({
+      card: {
+        type: "AdaptiveCard",
+        version: "1.5",
+        body: [
+          {
+            type: "ActionSet",
+            actions: [
+              { type: "Action.OpenUrl", id: "open", title: "查看详情", url: "https://example.com" },
+            ],
+          },
+        ],
+        actions: [{ type: "Action.Submit", id: "ok", title: "允许" }],
+      },
+      target,
+      onAction: () => {},
+    });
+    // index.css 的 .ac-pushButton[role="link"] 超链样式依赖 SDK 给 Action.OpenUrl 打
+    // role="link"、给 Submit/Toggle/Copy 打 role="button"；SDK 升级若改了 role，超链会
+    // 失效或误伤决策按钮——此测试兜底锁死该契约。
+    const roles = Array.from(
+      target.querySelectorAll<HTMLButtonElement>(".ac-pushButton")
+    ).map((b) => b.getAttribute("role"));
+    expect(roles).toContain("link"); // Action.OpenUrl → 超链
+    expect(roles).toContain("button"); // Action.Submit → 按钮
+    target.remove();
+  });
+
   it("Action.ToggleVisibility 由 SDK 原生切换 isVisible", () => {
     const target = mountTarget();
     const types: string[] = [];
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/renderOctoCard.test.tsx
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/renderOctoCard.test.tsx
@@ -159,11 +159,18 @@ describe("renderOctoCard", () => {
     // index.css 的 .ac-pushButton[role="link"] 超链样式依赖 SDK 给 Action.OpenUrl 打
     // role="link"、给 Submit/Toggle/Copy 打 role="button"；SDK 升级若改了 role，超链会
     // 失效或误伤决策按钮——此测试兜底锁死该契约。
-    const roles = Array.from(
-      target.querySelectorAll<HTMLButtonElement>(".ac-pushButton")
-    ).map((b) => b.getAttribute("role"));
-    expect(roles).toContain("link"); // Action.OpenUrl → 超链
-    expect(roles).toContain("button"); // Action.Submit → 按钮
+    // 按 title 绑定 role↔action：只断言「有一个 link + 一个 button」不够——SDK 若把两者
+    // role 对调（OpenUrl→button、Submit→link）仍会误过；这里钉死到具体按钮。
+    const roleByTitle = new Map(
+      Array.from(
+        target.querySelectorAll<HTMLButtonElement>(".ac-pushButton")
+      ).map((b) => [
+        b.getAttribute("title") ?? b.textContent?.trim() ?? "",
+        b.getAttribute("role"),
+      ])
+    );
+    expect(roleByTitle.get("查看详情")).toBe("link"); // Action.OpenUrl → 超链
+    expect(roleByTitle.get("允许")).toBe("button"); // Action.Submit → 按钮
     target.remove();
   });
 
--- a/packages/dmworkbase/src/Messages/InteractiveCard/index.css
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/index.css
@@ -636,7 +636,7 @@
 }
 
 .wk-interactive-card-sdk .ac-pushButton[role="link"]:hover {
-  color: var(--wk-color-accent-hover, var(--wk-text-accent));
+  color: var(--wk-color-accent-hover, var(--wk-text-accent, var(--wk-brand-primary)));
   background: transparent;
   border: 0;
   box-shadow: none;
--- a/packages/dmworkbase/src/Messages/InteractiveCard/index.css
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/index.css
@@ -614,6 +614,41 @@
   border-color: var(--wk-border-strong);
 }
 
+/* ── Action.OpenUrl → HTML 超链 ──
+ * AdaptiveCards 给 Action.OpenUrl 的按钮打 role="link"（Submit/ToggleVisibility/
+ * CopyToClipboard 都是 role="button"）。导航类动作渲成经典超链接：语义更贴，也不再
+ * 和主/次决策按钮抢视觉（如审批卡的「查看详情」夹在 允许/拒绝 中间过于抢眼）。
+ * 用 role 精准命中——决策按钮(style-positive/destructive)、推理卡展开按钮
+ * (ToggleVisibility)、复制按钮(CopyToClipboard) 均为 role="button"，完全不受影响。 */
+.wk-interactive-card-sdk .ac-pushButton[role="link"] {
+  display: inline;
+  min-height: auto;
+  height: auto;
+  padding: 0;
+  font-weight: var(--wk-font-regular);
+  color: var(--wk-text-accent, var(--wk-brand-primary));
+  background: transparent;
+  border: 0;
+  border-radius: 0;
+  box-shadow: none;
+  text-decoration: underline;
+  text-underline-offset: 2px;
+}
+
+.wk-interactive-card-sdk .ac-pushButton[role="link"]:hover {
+  color: var(--wk-color-accent-hover, var(--wk-text-accent));
+  background: transparent;
+  border: 0;
+  box-shadow: none;
+  text-decoration: underline;
+}
+
+.wk-interactive-card-sdk .ac-pushButton[role="link"]:focus-visible {
+  outline: 2px solid var(--wk-border-glow);
+  outline-offset: 2px;
+  box-shadow: none;
+}
+
 .wk-interactive-card-sdk .ac-textInput,
 .wk-interactive-card-sdk .ac-numberInput,
 .wk-interactive-card-sdk .ac-dateInput,


OctoBoooot (dismissed):
## Review: render Action.OpenUrl as a hyperlink (#827)

**Verdict:** Approve
A clean CSS-only change (+65/−0) that styles `Action.OpenUrl` buttons as hyperlinks via a role-scoped selector, plus a render-contract test. Byte-verified: no navigation/security surface, surgical role scoping, tokens defined, test-locked. No blockers, no majors.

Code CI green (Build newest leg `success`, secret-scan / osv-scan / dependency-review / pr-title-lint all pass). The one red check is **`check-sprint` — governance CI** (Sprint-field gate, a maintainer action a code change can't turn green); per policy it does **not** bar the code verdict, and `mergeStateStatus` stays BLOCKED until a maintainer clears it. Risk tier: normal. size/M.

### Verified clean
- **No new security surface — visual only** — this styles `.ac-pushButton[role="link"]` to look like a hyperlink; it does **not** introduce a raw `<a href>`, `innerHTML`, `window.open`, or any URL-navigation change. `Action.OpenUrl` remains an SDK-dispatched button (the URL-handling/validation path is untouched), so no new XSS / `javascript:`-scheme / open-redirect vector. The only `url` in the diff is test-fixture data (`https://example.com`).
- **Surgical role scoping** — the rule targets `role="link"`, which AdaptiveCards assigns only to `Action.OpenUrl`; decision buttons (`style-positive`/`destructive`), ToggleVisibility, and CopyToClipboard are all `role="button"` and unaffected. No collision with the approval-button styling from #816.
- **Tokens defined + fallback-guarded** — `--wk-text-accent`/`--wk-color-accent-hover`/`--wk-border-glow`/`--wk-font-regular` all defined in `semantic.css`, and each `var()` reference carries a fallback (e.g. `var(--wk-text-accent, var(--wk-brand-primary))`). `focus-visible` outline included for keyboard a11y.
- **Test-locked** — the new render-contract test asserts `Action.OpenUrl` → `role="link"` AND `Action.Submit` → `role="button"`, guarding the SDK class/role contract the CSS depends on against an SDK-rename break (same discipline as #816's render test).

### Praise
- Keying the hyperlink style on the SDK's `role="link"` (rather than an action-type class or a broad selector) is exactly the surgical approach — it hits only navigation actions and provably leaves the decision/toggle/copy buttons untouched, and the render-contract test locks that so an AdaptiveCards upgrade can't silently break it or bleed the link style onto decision buttons. Pairing it with the #816 approval-button styling gives the card a coherent primary/secondary/link hierarchy without touching any behavior.

### Net
A visual-only, role-scoped hyperlink style for `Action.OpenUrl` with no navigation/security change, defined+fallback-guarded tokens, focus-visible a11y, and a contract-locking test. No blockers, majors, or nits. **APPROVE** @ `c680700a`. The red `check-sprint` is governance CI (doesn't bar the code verdict); merge stays BLOCKED until a maintainer sets the Sprint field. Credit @an9xyz.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #827 (octo-web)

## Summary

This PR styles `Action.OpenUrl` buttons in the InteractiveCard component as classic underlined hyperlinks instead of the default accent-coloured pill buttons. The approach is precise: a CSS attribute selector `[role="link"]` targets only OpenUrl actions (which the AdaptiveCards SDK marks with `role="link"`), leaving Submit, ToggleVisibility, and CopyToClipboard buttons (all `role="button"`) untouched. A render test locks the SDK role contract so future SDK upgrades can't silently break the styling.

Clean, well-scoped CSS+test change. No correctness, security, or build concerns.

## Verification

- ✅ **Selector precision** — `.ac-pushButton[role="link"]` at specificity (0,2,1) correctly overrides the base `.ac-pushButton` rule at (0,2,0). The `.style-positive` (0,3,0) and `.style-destructive` (0,3,0) rules have higher specificity but apply only to Submit buttons in practice (OpenUrl buttons in production card templates do not carry `style` classes).
- ✅ **CSS variable availability** — `--wk-text-accent`, `--wk-brand-primary`, `--wk-font-regular`, `--wk-border-glow`, and `--wk-color-accent-hover` are all established design tokens used across the dmworkbase codebase. All usages include fallback values.
- ✅ **Test covers SDK contract** — The new test at `renderOctoCard.test.tsx:140-169` renders through the real AdaptiveCards SDK and asserts that `Action.OpenUrl` produces `role="link"` while `Action.Submit` produces `role="button"`. This directly locks the selector's precondition.
- ✅ **No regression to existing button styles** — Audited all `.ac-pushButton` rules in `index.css`: decision buttons (`.style-positive`/`.style-destructive`), the agent-progress reasoning toggle (ToggleVisibility), the table copy button (different class), and CopyToClipboard all remain `role="button"` and are unaffected.
- ✅ **Layout compatibility** — `display: inline` on a flex item within the SDK's action container is valid per the flexbox spec; the button renders as an inline-level item without stretching.

Static analysis only at head `c680700a`; build and tests not executed in this environment.

## Findings

### Nit — Hover colour fallback chain inconsistency (`index.css:638`)

The base rule uses `color: var(--wk-text-accent, var(--wk-brand-primary))` while the hover rule uses `color: var(--wk-color-accent-hover, var(--wk-text-accent))`. If `--wk-color-accent-hover` is undefined and `--wk-text-accent` is also undefined, hover would fall back to `--wk-text-accent` (which is itself undefined in this hypothetical) rather than `--wk-brand-primary` as the base rule does. In practice the design system defines these tokens so this is unreachable, but aligning the fallback chain to `var(--wk-color-accent-hover, var(--wk-text-accent, var(--wk-brand-primary)))` would be more defensive.

- CSS specificity ordering across all `.ac-pushButton` variants.
- Test uses real AdaptiveCards SDK rendering, not hand-fed DOM — the role contract is verified end-to-end.
- `display: inline` in flex container is valid and intentional for link-style buttons.
- Focus-visible outline (`--wk-border-glow`) preserves keyboard accessibility.
- All CSS custom properties have fallback values.
- Test cleans up DOM (`target.remove()`) after each case.

## Octo-Q Extended Analysis

### Data Flow Tracing

- **Consumed data: `role="link"` attribute** — Source: AdaptiveCards SDK `renderAction()` sets `role="link"` for `Action.OpenUrl` and `role="button"` for all other action types. The test at `renderOctoCard.test.tsx:140-169` verifies this data flows from the SDK to the DOM. The CSS selector `[role="link"]` consumes it correctly. ✅

### Blind-spot Checklist (C1–C6)

- **C1 — Dual-path parity**: N/A. No paired add/remove or symmetric operations in a CSS-only styling change.
- **C2 — Control-flow ordering / nested reuse**: N/A. CSS rules are declarative; no control flow or reuse concern.
- **C3 — Authorization boundary**: N/A. Visual styling change; no auth/authz surface affected.
- **C4 — Authorization lifecycle**: N/A. No auth state or container-member cascade involved.
- **C5 — Build/note ≠ runtime correctness**: Clear. CSS is loaded at runtime via the component's stylesheet import. No build-time transformation risk.
- **C6 — Governance/docs self-consistency**: N/A. No governance, policy, or documentation changes.

### Cross-round Blocker Recheck (R6)

N/A — first review round.

## Verdict: APPROVED

Well-scoped CSS-only change with a solid test locking the SDK role contract. Selector precision is verified, specificity ordering is correct, and no existing button styles are affected. The single nit (hover fallback chain alignment) is cosmetic and unreachable in practice.

Jerry-Xin (dismissed):
Reviewed the diff (+65/−0, CSS + test only). This is a pure visual restyle of the already-existing, already-sanitized `Action.OpenUrl` control — no new attack surface, tightly-scoped selector, no CSS injection. Approving.

### Why this is safe (byte-verified at head)
- **No new clickability / no new href.** Clickability already existed before this PR. `Action.OpenUrl` renders as an AdaptiveCards `<button role="link">` (not an `<a href>`), so the untrusted card URL never enters a DOM `href` attribute at all. On click, the SDK's `onExecuteAction` routes to `InteractiveCardCell` → `renderer/actions.ts openUrl(url)`.
- **URL scheme sanitization is pre-existing and layered.** `openUrl()` re-checks `isSafeUrl()` (http/https allowlist in `Utils/security.ts`) and opens via `window.open(url, "_blank", "noopener,noreferrer")`. Separately, `validateCardForOcto` throws `OctoInvalidCard "Action.OpenUrl has unsafe/missing url"` on `javascript:`/`data:`/etc., downgrading the whole card at validation time. This PR does not touch any of that.
- **CSS is injection-safe.** The new block uses only visual properties (color/underline/padding/outline via design tokens). No `content: attr(...)`, no `url(...)`, no `attr()` pulling attacker-controlled card data into rendered output.
- **Selector blast radius is precise.** The new rules are gated as `.wk-interactive-card-sdk .ac-pushButton[role="link"]` (+ `:hover` / `:focus-visible`). AdaptiveCards assigns `role="link"` only to `Action.OpenUrl`; Submit / ToggleVisibility / CopyToClipboard are `role="button"`. This composes with the existing base `.ac-pushButton` rule and does not overlap the class-gated `.style-positive` / `.style-destructive` decision-button rules. Reasoning-toggle and copy buttons are untouched.

### 💬 Non-blocking
- **Tighten the role test's binding.** The new test asserts the collected `roles` array `toContain("link")` and `toContain("button")`, but it does not bind each role to its action/title. A hypothetical SDK regression that flipped roles (OpenUrl→`button`, Submit→`link`) could still pass. Consider querying by title and asserting `查看详情` has `role="link"` while `允许` has `role="button"` to lock the contract more strictly.
- **Sprint/board process check is red.** `check-sprint` fails (`No linked issue found`), consistent with the PR's "Related Issue: N/A". Non-required, non-blocking for code review — just flagging so a maintainer can link an issue / set the sprint if the board requires it before merge.

### ✅ Highlights
- Targets the SDK semantic `role` contract instead of button text/card structure, and adds a render test locking that contract — a good guard against future SDK upgrades silently breaking the styling.
- Preserves design-token usage (accent color, hover token, focus outline) with no hard-coded colors, and keeps a11y focus styling.

yujiawei (dismissed):
# Code Review — PR #827 (octo-web)

**Head SHA:** `c680700a0a88ee5a5b9f0b00f2127cc80970770d`
**Scope:** CSS-only visual refinement + one render test. 65 additions, 2 files, 0 deletions.

## 1. Spec compliance

**Spec: ✅**

- Adds `.wk-interactive-card-sdk .ac-pushButton[role="link"]` (+ `:hover` / `:focus-visible`) to render `Action.OpenUrl` buttons as an underlined hyperlink (`index.css:623-651`).
- Adds a render test locking the SDK role contract (`__tests__/renderOctoCard.test.tsx:140-169`).

No over-build (no new component, flag, or field), no under-build, no deviation from the described approach.

## 2. Core premise — verified

The whole change rests on one assumption: AdaptiveCards 3.0.6 tags `Action.OpenUrl` with `role="link"` and everything else with `role="button"`. Confirmed directly in the SDK source:

- `OpenUrlAction.getAriaRole()` defaults to `"link"` (`adaptivecards/lib/card-elements.js:4662`).
- Base `Action.getAriaRole()` defaults to `"button"` (`card-elements.js:4304`); `Submit` / `ToggleVisibility` / `CopyToClipboard` inherit it.

So the selector targets navigation actions precisely and cannot leak onto decision buttons, the reasoning toggle, or the copy button. The new test pins this contract so an SDK bump that changes roles fails loudly. Good defensive move.

Specificity is also safe: the new rule `[role="link"]` is `(0,2,1)`, equal to `.style-positive`/`.style-destructive` but they never co-occur on the same element (an OpenUrl action carries no `style-positive`/`destructive`), and the agent-progress toggle selector is `(0,3,1)` and higher-specificity anyway — the reasoning toggle is `role="button"` regardless, so it is untouched.

## 3. Code quality

**Quality: Approved**

No blocking issues. Tests pass locally (13/13 in `renderOctoCard.test.tsx`), CSS lint is clean (the two `#fff` warnings at lines 594/600 are pre-existing, not from this PR).

The following are **advisory (P2) suggestions**, not merge blockers:

- **[P2] Test asserts role presence, not role↔action mapping** (`renderOctoCard.test.tsx:162-164`). `expect(roles).toContain("link")` / `toContain("button")` would still pass if a future SDK swapped the roles (link on Submit, button on OpenUrl). To make the contract-lock airtight, query the specific button (e.g. by title "查看详情" / "允许") and assert its exact role. Minor — the current test already catches the most likely regression (roles disappearing).

- **[P2] `display: inline` on a button element** (`index.css:625`). AdaptiveCards renders actions inside a flex `ActionSet`; forcing `display: inline` on a flex item is generally fine (flex context wins for direct children), but if an OpenUrl link ever sits in a mixed ActionSet with Submit buttons, verify wrapping/baseline alignment looks right. The PR description says this was checked across four card shapes, so likely a non-issue — noting for the record.

- **[P2] No `!important` on the reset properties.** This is only a concern if the HostConfig injected inline styles onto buttons. It does not: the SDK styles push buttons via CSS classes (`ac-pushButton`, `style-positive` — `card-elements.js:4233`, `updateCssClasses`), not inline color/background. The already-shipped `.style-positive` solid-fill rule works without `!important`, which confirms class-based overrides are sufficient here. No action needed.

## 4. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Clean, well-scoped, semantically correct CSS change with a contract test guarding the one SDK assumption it depends on. The three P2 items are optional polish.

## 5. Coverage notes

- Independent advisory pass (long-context, cross-family) ran and surfaced the three P2 items above; all were re-verified against the SDK source and the existing stylesheet before inclusion.
- A second independent line-level advisory leg was unavailable this run due to an upstream gateway error (image-generation tool misconfiguration on the provider side) and is therefore recorded as absent, not as agreement. This does not affect the verdict: the change is CSS-only, the diff is small, and the core premise was verified directly from SDK source.
- Not exercised: runtime visual rendering in a real browser (dark/light theme). The PR reports manual verification across approval, custom-action, agent-progress, and pipeline cards; the CSS variables used (`--wk-text-accent`, `--wk-color-accent-hover`, `--wk-border-glow`, `--wk-font-regular`) all resolve in `theme/semantic.css`.

OctoBoooot (approved):
## Re-review (DELTA) #827 @ `7ef48a17` — folds in the review nits (tight role-binding test + aligned fallback chain); check-sprint now green → APPROVE

**Verdict:** Approve
My prior APPROVE was dismissed by a new commit that addresses the review notes: the render-contract test now binds role↔action by title, and the hover fallback chain is aligned to the base rule. Byte-verified — no logic/behavior change, still pure-visual. Re-registering APPROVE. Notably, **`check-sprint` is now green** at this head (the governance gate that had it UNSTABLE is cleared — a linked issue / Sprint field was added, exactly the path identified last round).

Code CI green: Build newest leg `success`, **check-sprint `success`** (was the sole red), secret-scan / osv-scan / dependency-review / pr-title-lint all pass. size/M.

## Parity checked
Delta is test-hardening + a one-line CSS fallback fix; the visual change is unchanged:
- **Role-contract test tightened** — now binds role↔action by button title (查看详情 → `role="link"`, 允许 → `role="button"`) via a `roleByTitle` map, instead of just asserting both roles appear in the set. This closes the exact gap @Jerry-Xin + yujiawei flagged: an SDK role-swap (OpenUrl→button, Submit→link) would now fail the test rather than sneak through.
- **Hover fallback chain aligned** — `var(--wk-color-accent-hover, var(--wk-text-accent, var(--wk-brand-primary)))` now bottoms out in the same real token as the base rule, resolving the asymmetric-fallback nit (mine + mochashanyao's).
- **No new security/behavior surface** — still CSS + test only; no `<a href>`/`innerHTML`/`window.open`, no navigation change; the OpenUrl href handling (`isSafeUrl` allowlist + `noopener,noreferrer`, per the prior byte-verify) is untouched; selector still surgically scoped to `role="link"`. `ahead:1 behind:0`, no drift.

### Praise
- Both fixes are precisely the right response: binding the test to specific actions (not just role presence) makes the contract test actually catch an SDK role-swap, and completing the fallback chain to a concrete token matches the base rule's defensiveness. Clean, targeted follow-up to the review round.

### Net
The two review notes (SDK-role-swap test gap, asymmetric hover fallback) are folded in with no behavior change; the pure-visual OpenUrl-hyperlink styling is intact and byte-verified clean; and `check-sprint` is now green so the earlier UNSTABLE gate is cleared. Re-registering my **APPROVE** @ `7ef48a17`. No blockers, majors, or nits remaining. Credit @an9xyz for tightening the test + aligning the fallback, and @Jerry-Xin + yujiawei + mochashanyao for the axes. (Peers dismissed by this commit should re-anchor.)

Jerry-Xin (approved):
Re-review at new head. Confirming my prior APPROVE — the incremental delta since the previously reviewed commit is a safe, tightly-scoped refinement of exactly the two follow-up nits, and the core safety invariants are byte-verified intact.

## Delta since previous head (linear, +1 commit, 2 files)

**1. `index.css` — hover-color fallback hardening (1 line)**

`.wk-interactive-card-sdk .ac-pushButton[role="link"]:hover` color:
- before: `var(--wk-color-accent-hover, var(--wk-text-accent))`
- after: `var(--wk-color-accent-hover, var(--wk-text-accent, var(--wk-brand-primary)))`

Adds a deeper `var()` fallback so the hover color chain bottoms out at the defined `--wk-brand-primary` brand token (verified defined and in active use elsewhere in the package). Pure graceful-degradation; the selector is unchanged, still surgically scoped to `[role="link"]` (Action.OpenUrl only). No `content:`/`url()`/`attr()` and no card-data pull introduced.

**2. `renderOctoCard.test.tsx` — role↔action assertion tightened**

Previously the test collected all `.ac-pushButton` roles and asserted presence (`toContain("link")` + `toContain("button")`), which would still pass if the SDK swapped the roles. It now builds a `title → role` map and pins the mapping exactly: the OpenUrl button asserts `role="link"` and the Submit button asserts `role="button"`. Strictly stronger coverage; no meaningful assertion removed. This resolves the earlier note that the test only bound role-presence rather than the exact action→role mapping.

## Safety invariants re-verified (no regression)

- **No component logic change.** All `InteractiveCard` `.tsx`/`.ts` files are byte-identical to the previously reviewed head (blob-hash confirmed), including the renderer/action, card-tree sanitization, and card-validation paths. The URL navigation path — scheme allowlist (http/https), `noopener,noreferrer`, and card validation — is unchanged and pre-existing.
- **No new href/navigation sink**, no `innerHTML`, no `window.open` introduced by this change (still CSS + test only).
- **Selector blast radius unchanged**: the hyperlink styling stays scoped to `[role="link"]`, leaving decision/toggle/copy buttons (`role="button"`) unaffected.
- **Tokens fallback-guarded**: the accent/hover token chain now terminates at a defined brand token.

## CI

All required checks pass, including the governance sprint gate (previously flagged as red for a missing linked issue — now cleared). `code-review` is queued (not failing). Non-draft, mergeable.

## Highlights

- Nits addressed surgically without touching component logic.
- The tightened render-contract test is a good regression guard against future SDK role changes flipping OpenUrl vs Submit semantics.

APPROVE.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #827 (octo-web)

## Summary

This PR adds CSS rules to render `Action.OpenUrl` buttons as inline hyperlinks instead of the default pill-shaped button, leveraging the `role="link"` attribute that the AdaptiveCards SDK assigns to OpenUrl actions. A companion test locks in the SDK role-assignment contract (OpenUrl → `role="link"`, Submit → `role="button"`) so a future SDK upgrade cannot silently break the hyperlink styling. Clean, well-scoped visual differentiation with no production logic changes.

## Verification

- ✅ **CSS specificity chain** — The new `[role="link"]` selector at (0,3,0) beats the base `.ac-pushButton` at (0,2,0); its `:hover` at (0,4,0) beats the base hover at (0,3,0); its `:focus-visible` at (0,4,0) beats the base `:focus-visible` at (0,3,0). Source order also favors the new rules. All overrides are correct.
- ✅ **No selector collision with style-positive/destructive** — `style-positive` and `style-destructive` buttons carry `role="button"` (Action.Submit), so `[role="link"]` never matches them. Mutual exclusion verified at `index.css:593–617`.
- ✅ **CSS variables defined** — `--wk-font-regular` = `400` (`semantic.css:169`), `--wk-text-accent` defined in theme, `--wk-color-accent-hover` = `var(--wk-purple-600)` (`semantic.css:26`), `--wk-border-glow` used in 34+ locations. All fallback chains are well-formed.
- ✅ **Test contract strength** — The test at `renderOctoCard.test.tsx:140–176` uses a `roleByTitle` map keyed by button title, which catches the role-swap attack (OpenUrl→button, Submit→link) that a simple role-count assertion would miss. Cleanup pattern matches existing tests.
- ✅ **Graceful degradation** — If a future SDK version stops setting `role="link"`, the OpenUrl button falls back to the base `.ac-pushButton` style (the default pill button) rather than breaking. The test would catch the regression.

Static analysis only at head `7ef48a171f2d9d05b6029f2b2b24300675181249`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. No P2 issues. No correctness, security, build, or functional concerns.

- `display: inline` override on a flex child — CSS spec blockifies `inline` to `block` for flex items, so the element renders as a block-level flex child with inline-level content flow. No layout breakage.
- Focus-visible accessibility — `outline: 2px solid var(--wk-border-glow)` with `outline-offset: 2px` provides a clear keyboard focus indicator. The base `outline: none` is properly overridden.
- `box-automated review: none` on hover and focus-visible — correctly clears the base `focus-visible` `box-automated review: 0 0 0 var(--wk-sp-0-5, 2px) var(--wk-border-glow)` via higher specificity.
- Test places `Action.OpenUrl` inside `body[].ActionSet.actions` (inline action set) while `Action.Submit` is in top-level `actions` (card-level action set). This reflects real-world card structures and tests both rendering paths.
- Comments are substantive: they explain *why* the role-based selector was chosen and what breaks if the SDK contract changes, not just what the code does.

## Verdict: APPROVED

Well-executed style differentiation with a strong regression test. The role-based CSS selector cleanly separates navigation actions from decision actions without touching any production logic. No concerns.

yujiawei (approved):
# Code Review — PR #827 (octo-web)

**Scope:** 2 files, +72 / -0 — `InteractiveCard/index.css` (+35 CSS) and `renderOctoCard.test.tsx` (+37 test). No TypeScript/logic changes. Reviewed against head `7ef48a1`.

## Summary — Approve

This is a clean, well-scoped, role-gated CSS restyle: `Action.OpenUrl` buttons are re-painted as underlined text links so navigation actions stop competing visually with primary/secondary decision buttons. The approach is correct and I verified the load-bearing SDK contract it depends on. No blocking issues found. A few non-blocking polish items below.

## What I verified

- **SDK role contract (the load-bearing assumption).** Confirmed against the `adaptivecards@3.0.6` source pinned in `packages/dmworkbase/package.json`: `OpenUrlAction.getAriaRole()` returns `"link"`, while the base `Action` (and therefore `Action.Submit`, `Action.ToggleVisibility`, and the custom `Action.CopyToClipboard`) returns `"button"`. The `[role="link"]` selector is therefore precise — it **cannot** touch decision/toggle/copy buttons. The claim in the code comment is accurate.
- **Selector precedence.** `.ac-pushButton[role="link"]` (specificity 0,3,0) sits *after* the base `.ac-pushButton` (0,2,0) and after `.style-positive`/`.style-destructive` (0,3,0) in source order, so it wins correctly without needing `!important`. Even in the edge case where a server template sets `style: positive` on an `Action.OpenUrl` (the button would carry both `role="link"` and `.style-positive`), source order resolves the tie in favor of the link style — the intended behavior.
- **Wrapper + tokens.** The `.wk-interactive-card-sdk` scoping class is applied on the container (`InteractiveCardCell.tsx`), and every referenced custom property (`--wk-text-accent`, `--wk-brand-primary`, `--wk-color-accent-hover`, `--wk-border-glow`, `--wk-font-regular`) is defined in `theme/semantic.css` (light + dark).
- **Test validity.** The added test is correct and not vacuous: it maps rendered `.ac-pushButton` elements by `title` → `role` and pins `查看详情 → link`, `允许 → button`. Binding by title (rather than a count of "one link + one button") is the right call — it catches an SDK role-swap regression that a count-only assertion would miss. It's a legitimate contract-pinning guard for the exact SDK output the CSS depends on.
- **Text contrast.** Link color `--wk-text-accent` resolves to `#7f3bf5` on the surface bg (~5.46:1 light) and `#9b6df7` (~4.45:1 dark) — both pass WCAG AA for normal text, and `underline` means the link isn't distinguished by color alone.

## Non-blocking suggestions

- **P2 — `display: inline` is a dead declaration.** The SDK's base `Action.render()` sets `buttonElement.style.display = "flex"` as an *inline style* (`adaptivecards@3.0.6`), and `OpenUrlAction` does not override `render()`. An inline style beats any non-`!important` stylesheet rule regardless of specificity, so `display: inline` in the new rule is a no-op — the element stays a centered flex box. The *other* declarations (padding, border, background, text-decoration, color, font-weight) have no inline-style competitor and apply correctly, so the button still reads as a de-emphasized link visually. Since `Action.OpenUrl` renders inside an `ActionSet` (not inline in prose), true inline text flow isn't actually needed here — but the `display: inline` line and the "classic hyperlink" comment overstate what takes effect. Suggest either dropping `display: inline` (and `min-height/height`, which are also inert against flex) or switching to `display: inline-flex` so the declared and effective layout match. Cosmetic only.
- **P2 (follow-up, not this PR) — focus-ring contrast.** The `[role="link"]:focus-visible` outline uses `--wk-border-glow` (a ~20–25% alpha purple), which composites to a low contrast ratio against the card background. Worth noting that this is a **pre-existing, component-wide convention** — the base `.ac-pushButton:focus-visible` already uses the same token for its box-shadow ring, and this PR's link variant switches to a solid `outline` at the same alpha (roughly equivalent, arguably slightly better). So this is not a regression introduced here. If the focus indicator is ever hardened, it should be done uniformly for the whole component (opaque focus color to reach ≥3:1), not carved out for this one selector.
- **P2 (nit) — target size.** With `display`/`padding:0` the link's hit area collapses to ~one text line (~17px), under the 24px WCAG 2.5.8 target-size minimum. Acceptable for a de-emphasized secondary navigation affordance on desktop; a consideration for touch/mobile-webview if these cards render there.

## Verdict

No P0/P1 issues. The change does what it sets out to do, is correctly scoped, and the SDK contract it relies on is verified. The items above are polish/maintainability, not blockers.

PR #833, opened 2026-07-16
feat(loop): inline editing of agent detail properties (owner-only)

### LINKED ISSUE #832: Feature: inline editing of agent detail properties (owner-only)
## Problem
On the AI teammate (Agent) detail page, the right-side properties panel is
read-only. Owners have to leave the page to change basic properties (runtime,
model, concurrency) or the name/description, which is slow and inconsistent
with how the rest of the product is edited in place. An agent is also visible
to every workspace member, so any editing affordance must be shown only to the
owner.

## Proposed solution
Make the agent detail sidebar properties editable in place, gated by ownership:

- Name and description: hover reveals a gray background and a pencil hint;
  click to edit inline. An empty description shows an editable placeholder.
- Runtime: a switchable dropdown. The trigger shows a device icon, the runtime
  name and an online status dot. The popup is a wider rich list, grouped by
  device (daemon), with a provider logo, runtime name, badges and status dot,
  plus a mine/all filter and locked rows for other users' private runtimes.
- Model: a popup with free-text input, a "use custom value" action and a
  clear-to-default action (no model list is fetched).
- Concurrency: adjustable inline. The visibility and thinking-level rows are
  removed.
- Ownership gate: all of the above are editable only when the current member
  owns the agent; non-owners see a static, read-only panel.

## Additional context
Scoped to the dmloop package (agent detail page, runtime/model pickers,
inline-edit control, shared provider logo). The runtime list page reuses the
same provider logos for visual consistency.

superalsrk:
Closes #832

## Goal
Let agent owners edit the detail-page sidebar properties in place, matching
the product design, while keeping the panel read-only for other workspace
members who can only view the agent.

## Changes
- Inline editing on the agent detail sidebar (dmloop):
  - Name and description edit in place: hover shows a gray background and a
    pencil hint; an empty description shows an editable placeholder.
  - Runtime becomes a switchable dropdown — trigger shows a device icon, the
    runtime name and an online status dot; the popup is a wider rich list
    grouped by device (daemon), with provider logo, name, cloud/private badges
    and status dot, a mine/all filter, and locked rows for other users'
    private runtimes. Selecting persists via updateAgent({ runtime_id }).
  - Model becomes a popup with free-text input, a "use custom value" action
    and a clear-to-default action (no model list is fetched).
  - Concurrency is adjustable inline; the visibility and thinking-level rows
    are removed.
- Ownership gate: a canEdit flag (owner_id === current member) makes every
  control render a static read-only view for non-owners.
- Shared provider-logo component reused on the runtime list page so both
  surfaces render the same brand icons.
- API: UpdateAgentReq gains runtime_id; display fields are re-derived after
  an update so the sidebar keeps the runtime name and status.
- Input controls restyled to the loop skin (resting border + focus glow ring)
  and the model popup clear action redesigned (rounded hover + inset divider).

## Verification
- `pnpm -C packages/dmloop exec tsc --noEmit` — clean.
- `pnpm -C packages/dmloop test` — 63 passing.
- `pnpm exec stylelint` on changed CSS — 0 errors.
- i18n scan — 0 locale key issues.
- Manual: as owner, edited name/description/runtime/model/concurrency and
  confirmed persistence and the runtime online dot; as a non-owner, confirmed
  the panel is fully read-only.


Jerry-Xin (dismissed):
Frontend-only inline editing of agent detail properties with an owner-only UI gate. The core `InlineEdit` component, the runtime/model pickers, and the provider-logo renderer are all clean; no XSS, injection, or new unprotected write path introduced. Approving on the frontend; one item below is worth confirming on the backend.

### 💬 Non-blocking notes

1. **Owner-only is a client-side UI gate over the pre-existing `PUT /agents/:id` endpoint — confirm the server also enforces it.**
   - `packages/dmloop/src/panel/AgentDetailPage.tsx`: `canEdit = !!agent.owner_id && agent.owner_id === myMemberId`, and every editor (`InlineEdit`, `RuntimePicker`, `ModelPicker`) receives `canEdit` and renders read-only when false.
   - The save path is `patch()` → `updateAgent(id, ...)` → `httpPut('/agents/:id', ...)` in `packages/dmloop/src/api/agentApi.ts`. This endpoint already existed and was already the mutation path for instructions / MCP config / skills before this PR — this PR does **not** introduce a new write endpoint, it only adds a UI gate on top of an existing one, and `owner_id` is a server-provided field. So this PR does not create a privilege-escalation surface.
   - That said, the UI gate alone does not stop a non-owner from calling `PUT /agents/:id` directly. This is a backend responsibility and cannot be verified from this frontend diff. Please confirm `PUT /agents/:id` rejects non-owners server-side (403). If it does, this is fully fine.

2. **Runtime binding filter is client-side only.** `RuntimePicker.isLocked` disables non-workspace runtimes owned by others (`r.owner_id !== currentUserId && r.visibility !== "workspace"`), consistent with the earlier can-bind concern. Since the bind still goes through `PUT /agents/:id { runtime_id }`, the server should reject binding a runtime the caller can't use (403) rather than relying on the disabled state alone.

3. **`ModelPicker` accepts free-form model strings** (custom text, no allowlist). This matches the stated design (CLI can't enumerate real models), and the value is only stored/rendered as text (React-escaped), so no injection — just flagging that an invalid model string reaches the backend and should be validated there.

### ✅ Highlights

- `InlineEdit.tsx` renders all user/edited content through normal JSX (React auto-escapes); no `dangerouslySetInnerHTML`. Save failure is handled gracefully — `commit()` keeps `busy`/editing state in a `try/finally` and does not close the editor on a rejected `onSave`, so a failed save doesn't silently desync. IME composition is handled so Enter isn't hijacked mid-composition.
- `providerLogo.tsx` renders logos from a **static allowlist** (`switch (provider.toLowerCase())`) of hardcoded inline SVG components, falling back to the `Monitor` icon for unknown providers. No URL is built from the provider string, no untrusted SVG is inlined, no external fetch — no SVG-injection / SSRF vector.
- Input validation on edits: concurrency requires `Number.isFinite(n) && n >= 1`; empty name is a no-op (`v ? patch(...) : undefined`); MCP JSON is parsed and type-checked before save.
- `types.ts`: the single added field is `runtime_id?: string` on `UpdateAgentReq` — benign, matching the new runtime-edit capability.

Note: merge is currently BLOCKED on `satisfy-code-review` (a review is required) — this review unblocks that governance gate; it is not a code defect. Build and all scans (secret-scan, osv-scan, dependency-review) pass.

packages/dmloop/src/panel/AgentDetailPage.tsx
```diff
@@ -99,13 +108,31 @@ export default function AgentDetailPage({
...
+      owner_avatar: next.owner_avatar ?? prev?.owner_avatar ?? null,
+    }));
+  };
+
   const patch = async (p: Parameters<typeof updateAgent>[1]) => {
     if (!agent) return;
     const next = await updateAgent(agent.id, { name: agent.name, ...p });
-    setAgent(next);
+    mergeAgentDisplayFields(next);
     onChanged?.();
```
OctoBoooot, line 136: **[major]** `patch` awaits `updateAgent` (→ `httpPut`, which throws `LoopApiError` on any non-2xx — verified in `api/http.ts`, no global error toast) with no try/catch, and it is invoked via `void select()` / `void commit()` from RuntimePicker/ModelPicker/InlineEdit. A failed save becomes an unhandled rejection with no Toast, no revert, no retry cue — the dropdown has already closed and the trigger still shows the old value, so the user believes the change succeeded. Failing input: as the owner, pick a different runtime while the server returns 500/403 → the picker closes, no error shown, runtime silently unchanged. Every sibling save in this file (`saveInstr`/`saveMcp`/`removeSkill`/`doArchive`) does `Toast.error`; this path should too. Fix: wrap `patch` in try/catch with a `Toast.error` so all inline edits report failure.

packages/dmloop/src/panel/AgentDetailPage.tsx
```diff
@@ -251,8 +278,8 @@ export default function AgentDetailPage({
 
   const skills = agent.skills ?? [];
   const activeCount = tasks.filter((x) => isActiveRun(x.status)).length;
-  const thinking = agent.thinking_level ? t(`loop.agent.thinkingLevel.${agent.thinking_level}`) : "—";
-  const visibility = t(agent.visibility === "private" ? "loop.agent.visPrivate" : "loop.agent.visWorkspace");
+  // 属性仅归属人可改（agent 对工作区其他人可见但只读）。
+  const canEdit = !!agent.owner_id && agent.owner_id === myMemberId;
```
OctoBoooot, line 283: **[praise]** The `canEdit` default is well-considered: `myMemberId` starts `null` and is only set after `listAssigneeCandidates` resolves (reset to `null` on failure), so the UI defaults to read-only during the async load and on any directory-fetch error rather than briefly exposing edit controls — and both pickers plus `InlineEdit` render static read-only on `!canEdit`, so the gate is applied uniformly. Fail-closed is the right posture here. (Note: this is a UX gate, not a security boundary — see the backend-enforcement question in the summary.)

OctoBoooot (changes requested):
## Review: feat(loop): inline editing of agent detail properties (owner-only) (#833)

**Verdict:** Request changes
The ownership gate is correct and fail-closed, but every new inline editor saves through a `patch` helper with no error handling — a failed save (runtime/model/concurrency/name/description) gives the user no feedback at all, unlike every other save in the same panel.

CI: `check-sprint` (governance) is red — that's a maintainer/sprint-link action, not a code gate, so it doesn't bar the verdict; `mergeStateStatus` stays BLOCKED until it's cleared. Build was still `in_progress` at review time — confirm it goes green before merge.

### Major
- `packages/dmloop/src/panel/AgentDetailPage.tsx:133-138` — `patch` awaits `updateAgent` (→ `httpPut`, which throws `LoopApiError` on any non-2xx, verified in `api/http.ts`) with **no try/catch**, and it's invoked via `void select()` / `void commit()` from the pickers and InlineEdit. So a failed save surfaces as an unhandled promise rejection with **no Toast, no revert, no retry cue** — the dropdown has already closed and the trigger still shows the old value, so the user believes the change (e.g. switching the agent's runtime, which changes where it executes) succeeded when it didn't. Failing input: as the owner, open the runtime dropdown and pick a different runtime while the server returns 500/403 → the picker closes, no error is shown, the runtime silently stays unchanged. This is inconsistent with every sibling save in this file (`saveInstr`/`saveMcp`/`removeSkill`/`doArchive` all `Toast.error`). Fix: wrap `patch` in try/catch with a `Toast.error` (mirroring the siblings), so all inline edits report failure.

### Minor
- No tests accompany a 931-line feature that includes an ownership gate and post-update display-field re-derivation (`mergeAgentDisplayFields`). A test for `canEdit` (owner vs non-owner vs still-loading `myMemberId`) and for the runtime-name re-derivation after `updateAgent` would lock the two most breakable behaviors. (`packages/dmloop/src/api/__tests__/` exists — the naming convention is there.)
- The `visibility` and `thinking_level` rows are removed from the properties panel (they were read-only display rows, not editors). Intentional per the PR body, and issue #832 lists only runtime/model/concurrency/name/description as editable — but #832 doesn't ask to drop the visibility/thinking *display*. If this panel was the only surface showing those, confirm they're still visible somewhere (or that dropping them is the intended product decision).

### Questions
- The owner-only rule is enforced client-side (`canEdit`). Is the backend `PUT /agents/:id` independently enforcing owner-only writes (and rejecting `runtime_id` binds to runtimes the caller can't use)? The client gate is the right UX affordance and correctly fails closed, but it isn't a security boundary — a non-owner can still issue the PUT directly. Not something I can verify from this diff; flagging so the backend side is confirmed.

### Praise
- The `canEdit` default is genuinely well-considered: `!!agent.owner_id && agent.owner_id === myMemberId`, with `myMemberId` starting `null` and only set after `listAssigneeCandidates` resolves (and reset to `null` on failure) — so the UI defaults to read-only during the async load and on any directory-fetch error, rather than briefly exposing edit controls. Both pickers and `InlineEdit` render a static read-only view on `!canEdit`, so the gate is applied uniformly. That fail-closed posture is the right call for an ownership-gated surface.
- `InlineEdit` guards the IME composition window (`composing` ref + `onCompositionStart/End`) so Enter during a Chinese/Japanese input-method composition doesn't prematurely commit — a real correctness detail that's easy to miss.

### Out of scope (informational)
- `mergeAgentDisplayFields` re-derives `runtime_name` from the loaded `runtimes` list because `updateAgent`'s response omits it; if `listRuntimesForAgent()` failed on mount (`runtimes` empty), it falls back to the prior name — fine, just noting the display name depends on that separate fetch succeeding.

Jerry-Xin (dismissed):
REQUEST_CHANGES — correcting my earlier APPROVE. I byte-verified a peer's (OctoBoooot's) finding and it holds: the inline-edit save path has no error feedback. My prior review over-credited "graceful save-failure handling" — that's wrong for the inline path; my apologies.

## 🔴 Blocking — inline-edit saves have no error handling (silent failure)
The inline-edit controls (name, description, runtime, model, max-concurrent) all wire `onSave` directly to `patch()` (`packages/dmloop/src/panel/AgentDetailPage.tsx`). But:
- `patch` (`AgentDetailPage.tsx`) is `await updateAgent(...) → mergeAgentDisplayFields → onChanged()` with **no try/catch and no `Toast.error`** — it just throws on failure.
- `InlineEdit.commit()` (`packages/dmloop/src/ui/InlineEdit.tsx`) awaits `onSave` in a `try { … } finally { setBusy(false) }` with **no `.catch`**. On a rejected save, `setEditing(false)` is skipped (the editor stays open) but the user gets **no error toast and no explicit revert** — the value silently fails to persist with zero feedback (and an unhandled promise rejection).

Contrast the sibling savers in the same file — instructions/MCP/skills/archive all pair `Toast.success` on success and `Toast.error(...)` on failure. The inline path has neither. A user editing a name/model/runtime that fails server-side (e.g. a 403 on a non-bindable runtime, or a validation error) sees nothing and may believe it saved.

Fix: give the inline-edit save a `.catch` that surfaces `Toast.error` and reverts the draft to the last committed value (and a `Toast.success` for parity), either in `InlineEdit.commit` or in the `patch`-wrapping callbacks.

## 💬 Non-blocking / notes
- 🟡 **No unit tests** for the new `InlineEdit` / picker components and the edit flow (OctoBoooot's second 🔴 — I'd grade it a strong 🟡: it doesn't corrupt anything, but this is net +900 of new interactive code with no test coverage; at least a commit/save + error-path test would be worth adding).
- 🟡 **Owner-only is a client UI gate** over the **pre-existing** `PUT /agents/:id` endpoint (OctoBoooot's third 🔴 — I keep this non-blocking: the PR adds no new write path and doesn't weaken any existing protection; but please confirm the server rejects a non-owner PUT, since the UI gate alone can't). Same for the client-side RuntimePicker lock and free-text ModelPicker — the backend should 403 non-bindable runtimes / validate model strings.
- 🟡 Removed read-only display rows (`RuntimePage.tsx` -29, `runtime.css` -41) — confirm that's the intended product direction, not a lost surface.

Highlights that DO hold (byte-verified): no XSS — edited values render via JSX (React-escaped), zero `dangerouslySetInnerHTML`; `providerLogo.tsx` is a static hardcoded-SVG allowlist (no URL/SVG injection); input validation on concurrency (`Number.isFinite && >= 1`) and MCP JSON.

Add the inline-save error feedback + revert and this is close to APPROVE. Credit @OctoBoooot for catching the missing error handling.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #833 (octo-web)

## Summary

This PR adds inline editing to the agent detail sidebar: name, description, runtime, model, and concurrency are now editable in-place via a shared `InlineEdit` component, with owner-only write access enforced client-side via `canEdit`. It also extracts provider logos into a shared `ProviderLogo` component and removes the thinking-level and visibility display rows. The implementation is clean, the owner-guard is correctly fail-closed, and i18n parity (en-US / zh-CN) is maintained.

## Verification

Static analysis only at head `18d35009104e06430cf6534d636efe0f75b0fb24`; build and tests not executed in this environment.

- ✅ **Owner guard fail-closed** — `canEdit` at `packages/dmloop/src/panel/AgentDetailPage.tsx:281` requires `agent.owner_id === myMemberId`; if `listAssigneeCandidates` fails or `WKApp.loginInfo.uid` is unset, `myMemberId` is null and all fields render read-only. No bypass path found.
- ✅ **`updateAgent` enrichment gap handled** — `updateAgent` returns the raw server response without `enrich()` (no `runtime_name`/`owner_name`/`owner_avatar`). `mergeAgentDisplayFields` at `:121-131` compensates with fallback chains against the loaded `runtimes` array and previous agent state. Documented and correct.
- ✅ **i18n parity** — Both en-US and zh-CN add the same 9 new keys and remove the same 5 old keys (`thinking`, `thinkingLevel.*`). zh-CN additionally updates `agent.runtime` wording.
- ✅ **RuntimePicker lock logic** — `isLocked` at `packages/dmloop/src/panel/RuntimePicker.tsx:49` correctly checks `owner_id !== currentUserId && visibility !== "workspace"`. When `currentUserId` is null, `canEdit` is simultaneously false so the picker renders read-only — no exploitable gap.
- ✅ **ProviderLogo extraction** — `providerName` at `packages/dmloop/src/ui/providerLogo.tsx:7` is identical to the removed local function. No behavioral change.
- ✅ **Type contract** — `UpdateAgentReq` at `packages/dmloop/src/api/types.ts:554` correctly adds `runtime_id?: string`.

## Findings

### P2 — InlineEdit silently swallows API errors (`packages/dmloop/src/ui/InlineEdit.tsx:60`)

`commit()` has a `try/finally` but no `catch`. All three callers (`onBlur` at `:103`, `:127`, `:149`) use `void commit()`, which discards the returned promise. When `onSave` rejects — for instance when `updateAgent` fails with a network or server error — the rejection becomes an unhandled promise rejection. The field stays in edit mode (right behavior for retry), but the user receives no toast or signal that the save failed. Same pattern in `ModelPicker.tsx:35`.

**Diff-scope**: new (this PR introduces `InlineEdit` and `ModelPicker`).
**Fix**: add a `catch` in `commit()` that shows `Toast.error(...)` with an i18n key.

### P2 — Concurrency validation silently reverts on invalid input (`packages/dmloop/src/panel/AgentDetailPage.tsx:521`)

The concurrency `onSave` callback returns `undefined` when the entered value fails `Number.isFinite(n) && n >= 1`. `InlineEdit.commit()` then does `await undefined` (resolves immediately) and calls `setEditing(false)`, so the field exits edit mode and reverts to the previous value with no user feedback.

**Diff-scope**: new (this PR introduces the inline-edit concurrency field).
**Fix**: show a `Toast.warning(...)` when validation rejects, or return a rejected promise so the P2 above surfaces the feedback.

## Data-Flow Tracing

- **`canEdit`** → `agent.owner_id` (from `getAgent` → server) === `myMemberId` (from `listAssigneeCandidates` → match by `octo_uid === WKApp.loginInfo.uid`). Both sources loaded in `useEffect`. If either fails, `canEdit=false`. Data flows correctly to `InlineEdit`, `RuntimePicker`, `ModelPicker` via `canEdit` prop. ✅
- **`runtimes`** → `listRuntimesForAgent()` → `httpGet<RuntimeDevice[]>("/runtimes")` → `setRuntimes`. Consumed by `RuntimePicker` for dropdown items and by `mergeAgentDisplayFields` for `runtime_name` lookup. `.catch(() => setRuntimes([]))` prevents failure from blocking. ✅
- **`mergeAgentDisplayFields`** → triggered after every `patch()` call. Merges `next` (raw server response) with `prev` state and `runtimes` array to restore display fields. Fallback chain handles all edge cases (empty runtimes, missing prev). ✅
- **`patch()`** → reads `agent` from React closure → calls `updateAgent(agent.id, { name: agent.name, ...p })` → awaits → `mergeAgentDisplayFields(next)` → `onChanged?.()`. No race condition in practice because inline edits are sequential per field (busy guard). ✅

- **C1 (dual-path parity)**: No paired add/remove or subscribe/unsubscribe paths. N/A.
- **C2 (control-flow ordering)**: No reused logic across multiple call sites with ordering risk. N/A.
- **C3–C4 (authorization boundary/lifecycle)**: Owner check is UI-only UX enhancement; server-side authorization is the real gate. No new endpoints or auth changes. N/A.
- **C5 (build/note)**: No build config, packaging, or environment changes. N/A.
- **C6 (governance/docs)**: No governance or policy changes. N/A.

- Owner-only editing propagates correctly to all three editable components via `canEdit` prop.
- IME composition guard prevents Enter from committing during CJK input.
- `patch` always sends `name: agent.name` as baseline to prevent partial-update name loss.
- Runtime list and assignee candidates loaded in parallel with agent fetch; failures are gracefully handled.
- ProviderLogo SVGs are inline (no external asset dependencies).

## Verdict: APPROVED

No correctness, security, or data-integrity blockers. The two P2 items are UX feedback gaps that are non-blocking — the underlying data operations are safe (fail-closed on error, revert on invalid input). Worth addressing in a follow-up.

[Octo-Q] verdict: APPROVE — No P0/P1 found. Two P2 UX feedback items (silent error swallowing in InlineEdit/ModelPicker, silent revert on invalid concurrency input). Data flow tracing confirms all consumed values reach their consumption points correctly. Owner guard is properly fail-closed.

superalsrk:
@Jerry-Xin @OctoBoooot @mochashanyao Thanks for the review — addressed the blocking finding in `67e3c99d`, please take another look.

## Fixed (blocking): inline-edit saves now report failure
The inline-edit save path had no error feedback — a failed `updateAgent` closed the field/dropdown and silently reverted, unlike the sibling saves (`saveInstr`/`saveMcp`/`removeSkill`/`doArchive`).

- `InlineEdit.commit` (name / description / concurrency): wraps `await onSave` in `try/catch`; on rejection it shows `Toast.error(t("loop.toast.saveFailed"))` and **keeps the editor open with the draft intact** so the user can retry (no `setEditing(false)` on error).
- `RuntimePicker.select` (runtime) and `ModelPicker.commit` (model): catch a rejected `onChange` and show the same `saveFailed` toast; the popup has already closed so the value visibly reverts to the last committed one.

Kept the error handling inside the three inline components (not in `patch`) on purpose: `patch` is shared with the config-tab saves that already toast, so toasting there would double-fire.

## Notes on the other points
- **Backend owner-only enforcement (non-blocking):** confirmed as out of frontend scope. This PR adds no new write path — it's a UI gate over the pre-existing `PUT /agents/:id`. Owner-only / non-bindable-runtime (403) enforcement on the server is tracked separately on the backend.
- **Removed visibility + thinking-level display rows:** intentional product decision for this panel per #832; they were read-only display rows, not editors.
- **Tests (non-blocking):** will follow up with unit coverage for the `canEdit` gate and the post-update `mergeAgentDisplayFields` re-derivation.

Verification: `tsc --noEmit` clean · `pnpm -C packages/dmloop test` 63 passing.

yujiawei (approved):
# Code Review — PR #833 (octo-web)

**Verdict: APPROVED** — head SHA `18d35009104e06430cf6534d636efe0f75b0fb24`, reviewed against merge-base `1bded39a`.

Solid, well-scoped implementation that matches the linked issue precisely. Inline editing, the runtime/model pickers, the ownership gate, and the shared provider logos all land as specified. `tsc` on the package is clean, 63 unit tests pass, and the removed `thinking_level`/`visibility` rows leave no dangling i18n keys or CSS references. The findings below are non-blocking suggestions.

## 1. Spec compliance

All acceptance criteria are met, with no scope creep:

- Name/description inline edit with hover affordance + editable empty placeholder ✅
- Runtime dropdown: device icon + name + status dot trigger; rich popup grouped by daemon with provider logo, cloud/private badges, mine/all filter, and locked rows for others' private runtimes ✅
- Model popup: free-text input, "use custom value", clear-to-default, no model list fetched ✅
- Concurrency inline; visibility and thinking-level rows removed ✅
- Ownership gate: `canEdit` renders a static read-only view for non-owners ✅
- Shared provider-logo component reused on the runtime list page ✅
- `UpdateAgentReq` gains `runtime_id`; change confined to the dmloop package ✅

## 2. Code quality — non-blocking

**P2 — Double-submit window in `InlineEdit` (`ui/InlineEdit.tsx:52`).** The re-entrancy guard `if (busy) return` reads a state value, not a ref. On `Enter`, `commit()` runs and schedules `setBusy(true)`; if `onBlur` fires in the same tick (focus lost as the tree re-renders), it re-enters `commit()` with `busy` still `false` and can issue `onSave` twice. The equality short-circuit (`next === value.trim()`) masks it in most cases, but a `useRef` flag would close the gap deterministically.

**P2 — `patch()` always sends the current `name` (`panel/AgentDetailPage.tsx:134`).** `updateAgent(agent.id, { name: agent.name, ...p })` snapshots `name` from the render closure on every property edit. If two edits are in flight concurrently, the later request can carry a stale `name`. This is pre-existing behavior, but the PR widens the number of edit entry points that reach it. Since `UpdateAgentReq.name` is optional, consider only sending the fields being changed.

**P2 — `ProviderLogo` calls `provider.toLowerCase()` unguarded (`ui/providerLogo.tsx:166`).** The prior `providerName`/`providerTone` helpers had a `if (!provider)` guard. The type declares `provider: string` (non-null) and an empty string safely falls through to the `Monitor` default, so this is not a regression — but a defensive `provider?.toLowerCase() ?? ""` would harden it against a malformed runtime record.

**P2 — `String(agent.max_concurrent_tasks)` (`panel/AgentDetailPage.tsx:518`).** If a legacy record ever lacks this field, the read-only view would render the literal `"undefined"`. The type marks it required, so this is a nit; a `?? 1` fallback would be safer.

**P2 — No unit tests for the three new interactive components.** `InlineEdit`, `RuntimePicker`, and `ModelPicker` carry the non-trivial logic here (commit/cancel/compose handling, locking, filter/grouping) but have no coverage. Worth a follow-up.

## 3. Coverage note

Verified from the diff: type-checking, i18n key parity, CSS orphan check, picker filter/grouping/locking logic, and the merge/enrich display-field flow (`mergeAgentDisplayFields` correctly re-derives `runtime_name` because the `updateAgent` PUT response is not enriched, unlike `getAgent`).

Could not verify from the diff alone: (a) **server-side authorization** — `canEdit` is a client-only affordance gate and fails closed when the viewer's member id is unknown, but the diff does not show backend enforcement of owner-only writes on `PUT /agents/:id`; a non-owner blocked in the UI must still be rejected by the API. (b) Runtime-visibility semantics and the exact shape of the `updateAgent` response are backend contracts outside this diff.

yujiawei (changes requested):
# Code Review — PR #833 (octo-web)

**Verdict: CHANGES REQUESTED** — head SHA `18d35009104e06430cf6534d636efe0f75b0fb24`, reviewed against merge-base `1bded39a`.

> Correcting my own earlier Approve. On a second, byte-level pass I confirmed a blocking issue that other reviewers also flagged: the inline-edit save path has no error handling. My apologies for the premature approval — the spec compliance and the ownership gate are solid, but the missing save-failure feedback is a real functional defect that must be fixed before merge.

## 1. Spec compliance

All acceptance criteria from the linked issue are met, with no scope creep:

- Name/description inline edit with hover affordance + editable empty placeholder ✅
- Runtime dropdown: device icon + name + status dot trigger; rich popup grouped by daemon with provider logo, cloud/private badges, mine/all filter, and locked rows for others' private runtimes ✅
- Model popup: free-text input, "use custom value", clear-to-default, no model list fetched ✅
- Concurrency inline; visibility and thinking-level rows removed ✅
- Ownership gate: `canEdit` renders a static read-only view for non-owners ✅
- Shared provider-logo component reused on the runtime list page ✅
- `UpdateAgentReq` gains `runtime_id`; change confined to the dmloop package ✅

## 2. Code quality

### P1 — Blocking: inline-edit saves fail silently (no error feedback, no revert)

`packages/dmloop/src/panel/AgentDetailPage.tsx:132-137` — `patch()` does `await updateAgent(...) → mergeAgentDisplayFields → onChanged()` with **no try/catch**. `updateAgent` → `httpPut` throws `LoopApiError` on any non-2xx (verified in `api/http.ts:110/127`).

`packages/dmloop/src/ui/InlineEdit.tsx:59-65` — `commit()` awaits `onSave` in `try { … } finally { setBusy(false) }` with **no `.catch`**. On a rejected save: `setEditing(false)` is skipped (editor stays open) but there is **no `Toast.error` and no revert**, plus an unhandled promise rejection.

The pickers make it worse: `RuntimePicker` calls `void select()` (RuntimePicker.tsx:140) and `ModelPicker` calls `void commit()` (ModelPicker.tsx:56/66) — fire-and-forget. On failure the dropdown has already closed and the trigger still shows the old value, so the user believes the change succeeded when it did not. This is most damaging for the runtime switch, which changes where the agent executes.

Reproduce: as the owner, pick a different runtime (or edit name/model/concurrency) while the server returns 500/403 → the control closes, no error is shown, the value silently stays unchanged.

This is inconsistent with every sibling saver in the same file — `saveInstr` / `saveMcp` / `removeSkill` / `doArchive` all pair `Toast.success` with `Toast.error`. 

**Fix:** give the inline save a `.catch` that surfaces `Toast.error` and reverts the draft to the last committed value (plus `Toast.success` for parity), either in `InlineEdit.commit` / picker `onChange` wrappers or in the `patch` helper.

### P2 — Non-blocking

- **Double-submit window in `InlineEdit` (`ui/InlineEdit.tsx:52`).** The re-entrancy guard `if (busy) return` reads a state value, not a ref. On `Enter`, `commit()` runs and schedules `setBusy(true)`; if `onBlur` fires in the same tick, it can re-enter with `busy` still `false` and issue `onSave` twice. A `useRef` flag closes this deterministically. (This should be handled together with the P1 fix, since both live in `commit()`.)
- **`patch()` always sends `name` (`panel/AgentDetailPage.tsx:134`).** `{ name: agent.name, ...p }` snapshots `name` from the render closure on every property edit; two concurrent edits can let the later request carry a stale name. Pre-existing, but this PR widens the entry points. `UpdateAgentReq.name` is optional — consider sending only changed fields.
- **`ProviderLogo` calls `provider.toLowerCase()` unguarded (`ui/providerLogo.tsx:166`).** The prior helpers had an `if (!provider)` guard. `provider` is typed non-null and an empty string falls through to the `Monitor` default (not a regression), but `provider?.toLowerCase() ?? ""` would harden against a malformed record.
- **`String(agent.max_concurrent_tasks)` (`panel/AgentDetailPage.tsx:518`).** A legacy record lacking this field would render `"undefined"`. Field is typed required, so this is a nit; a `?? 1` fallback is safer.
- **No unit tests** for the three new interactive components. `InlineEdit`, `RuntimePicker`, `ModelPicker` carry the non-trivial logic (commit/compose handling, locking, filter/grouping) and the `canEdit` gate + `mergeAgentDisplayFields` re-derivation — the two most breakable behaviors — with no coverage. `packages/dmloop/src/api/__tests__/` already exists; worth at least a commit-success + error-path test.

## 3. What holds up well

- The `canEdit` gate is well-considered and **fails closed**: `!!agent.owner_id && agent.owner_id === myMemberId`, with `myMemberId` starting `null` (set only after `listAssigneeCandidates` resolves, reset to `null` on failure), so controls default to read-only during load and on directory-fetch error. Both pickers and `InlineEdit` render a static read-only view on `!canEdit`, applied uniformly.
- `InlineEdit` correctly guards the IME composition window (`composing` ref + `onCompositionStart/End`) so Enter mid-composition doesn't prematurely commit.
- No XSS: edited values render via React-escaped JSX, no `dangerouslySetInnerHTML`; `providerLogo` is a static hardcoded-SVG allowlist. Concurrency (`Number.isFinite && >= 1`) and MCP JSON are validated.
- `mergeAgentDisplayFields` correctly re-derives `runtime_name` from the loaded `runtimes` list because the `updateAgent` PUT response is not enriched (unlike `getAgent`).

## 4. Coverage note

Verified from the diff: type-checking (package `tsc` clean; the 4k+ errors in a raw whole-repo `tsc` are all in the sibling `dmworkbase` package / `node_modules` react@17 typings, none in the 12 changed files), 63 unit tests pass, i18n key parity for added/removed keys, CSS orphan check on removed `loop-runtime-row__provider` styles, picker filter/grouping/locking logic, and the merge/enrich display-field flow.

Could not verify from the diff alone: **server-side authorization** — `canEdit` is a client-only affordance and fails closed, but the diff does not show backend enforcement of owner-only writes on `PUT /agents/:id`; a non-owner blocked in the UI must still be rejected by the API, and non-bindable `runtime_id` binds / invalid model strings should 403/validate server-side. This PR adds no new write path and doesn't weaken existing protection, but please confirm the backend gate. Also confirm dropping the visibility/thinking-level *display* rows is the intended product decision (issue #832 asks to make runtime/model/concurrency editable, not to remove those displays).

OctoBoooot (approved):
## Re-review (DELTA) #833 @ `67e3c99` — the silent-save-failure major is fixed on all three inline-edit paths (toast + keep-draft-open on error); CI green → APPROVE

**Verdict:** Approve
Flipping my REQUEST_CHANGES → APPROVE. My major was that inline edits (name/description/runtime/model/concurrency) saved with no error feedback — a failed save silently reverted and the user believed it persisted. Byte-verified at this head: all three save paths now catch a rejected save and toast, and `InlineEdit` additionally keeps the editor open with the draft intact for retry. Code CI green.

## What I verified (the fix)
- **`InlineEdit.commit`** — now `try { await onSave(next); setEditing(false) } catch { Toast.error(t("loop.toast.saveFailed")) } finally { setBusy(false) }`. On failure it shows the toast AND leaves the editor open with the draft (since `setEditing(false)` is only in the try), so the user can retry — better than a bare toast. Covers name / description / concurrency.
- **`RuntimePicker.select`** — `try { await onChange(id) } catch { Toast.error(...) }`.
- **`ModelPicker.commit`** — `try { await onChange(next) } catch { Toast.error(...) }`.
- All three use `loop.toast.saveFailed`, consistent with the sibling saves (`saveInstr`/`saveMcp`/`removeSkill`) in the same panel. Surfacing the error at the UI boundary that owns the editing state (rather than in `patch`) is a reasonable choice — it's what lets `InlineEdit` preserve the draft. My blocking finding is resolved.
- **CI green** — Build success, check-sprint now green (issue linked), secret-scan / osv-scan / dependency-review / pr-title-lint pass.

### Non-blocking (independently confirmed @yujiawei's findings — worth a follow-up, not blocking)
- `patch` (AgentDetailPage.tsx:135) always sends `{ name: agent.name, ...p }`, so a partial update (e.g. changing runtime) re-sends the locally-held name. If another editor renamed the agent between load and this edit, this PUT reverts the name. Narrow (needs a concurrent rename of the same agent; backend is last-write-wins regardless) — **minor**. Consider sending only the changed field, or refetching before write.
- `ProviderLogo` (providerLogo.tsx:166) does `provider.toLowerCase()` with no null guard. `provider` is typed non-optional and callers pass `rt.provider`, so this only throws on a malformed payload — **minor** defensive hardening (`(provider ?? "").toLowerCase()`).
- `String(agent.max_concurrent_tasks)` (AgentDetailPage.tsx:518) renders the literal `"undefined"` if the field is ever absent. Typed required, so normally present — **minor** display glitch; a `?? 1` default would harden it.
- (Carried) No unit tests for the inline-edit / ownership-gate / display-re-derivation behavior; and confirm the backend `PUT /agents/:id` independently enforces owner-only writes + rejects unbindable `runtime_id` (the client `canEdit` gate is a UX affordance, not a security boundary).

### Praise
- The fix keeps the editor open with the draft intact on a failed `InlineEdit` save rather than just toasting — that's the right UX for a retry, and it's more than the minimum the finding asked for. All three surfaces now report failure consistently with the panel's existing saves.

### Net
The silent-save-failure major is fully resolved: `InlineEdit`, `RuntimePicker`, and `ModelPicker` all catch a rejected save and toast (InlineEdit also preserves the draft for retry), consistent with the sibling saves. Code CI green. Flipping my **REQUEST_CHANGES → APPROVE** @ `67e3c99`. Credit @OctoBoooot/@Jerry-Xin for the silent-save finding and @yujiawei for the follow-up hardening notes (name-clobber, ProviderLogo null-guard, concurrency default) — all independently confirmed as non-blocking. @yujiawei still holds a REQUEST_CHANGES at this head on those 🟡; that's a legitimate call on her part and the gate reflects it — the author resolving the name-clobber + adding tests would clear it. `size/XL` — worth a maintainer eyeball.

yujiawei (approved):
# Code Review — PR #833 (octo-web)

**Verdict: APPROVED** — head SHA `67e3c99d8ec1fb2b63428802b39ca34cf389ab15`, reviewed against merge-base `1bded39a`. Re-review after the previous Changes-Requested.

The new commit `67e3c99d "BugFix: surface save errors for agent detail inline edits"` resolves the P1 blocker from the prior round. All prior spec-compliance conclusions still hold on this head, and the remaining items are non-blocking.

## P1 fix verification (resolved)

The silent-save-failure blocker is fixed across all three inline-edit surfaces, matching the sibling-saver pattern in the same panel:

- `ui/InlineEdit.tsx:65-67` — `commit()` now `catch`es a rejected `onSave`, shows `Toast.error(t("loop.toast.saveFailed"))`, and **keeps the editor open with the draft intact** (the `catch` skips `setEditing(false)`), so the user gets feedback and can retry.
- `panel/RuntimePicker.tsx:82-88` — `select()` wraps `onChange` in try/catch with the same toast; the popup has already closed so the value visibly reverts to the persisted one.
- `panel/ModelPicker.tsx:35-41` — `commit()` gets the same treatment.
- `loop.toast.saveFailed` is present in both `en-US.json:343` and `zh-CN.json:343`.

No more unhandled promise rejections on the inline-save path; behavior is now consistent with `saveInstr`/`saveMcp`/`removeSkill`/`doArchive`.

## Remaining non-blocking items (P2, follow-up)

- **Double-submit window in `InlineEdit` (`ui/InlineEdit.tsx:55`).** The re-entrancy guard `if (busy) return` reads state, not a ref; on `Enter` followed by a same-tick `onBlur`, `commit()` can re-enter with `busy` still `false`. The equality short-circuit masks it in most cases. A `useRef` flag would close it deterministically.
- **Concurrency invalid-input silently reverts (`panel/AgentDetailPage.tsx:524`).** The `onSave` returns `undefined` when `Number.isFinite(n) && n >= 1` fails; `commit()` then `await undefined` → `setEditing(false)`, exiting edit mode with no feedback. A `Toast.warning` or a rejected promise (which the new `catch` would now surface) would inform the user.
- **`patch()` always sends the current `name` (`panel/AgentDetailPage.tsx:134`).** Pre-existing; two concurrent edits could carry a stale name. `UpdateAgentReq.name` is optional — consider sending only changed fields.
- **`ProviderLogo` unguarded `provider.toLowerCase()` (`ui/providerLogo.tsx:166`).** Not a regression (empty string → `Monitor` default; `provider` is typed non-null), but `provider?.toLowerCase() ?? ""` would harden it.
- **No unit tests** for the three new interactive components or the `canEdit` gate / `mergeAgentDisplayFields` re-derivation — the most breakable behaviors.

## Spec compliance (unchanged, still ✅)

All acceptance criteria from the linked issue are met with no scope creep: inline name/description/runtime/model/concurrency editing, owner-only `canEdit` gate (fail-closed), daemon-grouped runtime dropdown with mine/all filter and locked private rows, free-text model popup, shared provider logos reused on the runtime list page, `UpdateAgentReq.runtime_id` added, change confined to the dmloop package.

## What holds up well

- `canEdit` fails closed — read-only during async load and on directory-fetch error; applied uniformly across both pickers and `InlineEdit`.
- IME composition guard prevents premature Enter-commit during CJK input.
- No XSS (React-escaped JSX, static hardcoded-SVG allowlist); concurrency (`Number.isFinite && >= 1`) and MCP JSON validated.
- `mergeAgentDisplayFields` correctly re-derives `runtime_name` because the `updateAgent` PUT response is not enriched (unlike `getAgent`).

## Coverage note

Verified from the diff on this head: package `tsc` type-check path is clean for the changed files (a raw whole-repo `tsc` reports 4k+ errors, all in sibling `dmworkbase` / react@17 typings — none in the changed files); 63/63 dmloop unit tests pass; i18n key parity for the new `saveFailed` usage; the P1 fix across all three surfaces.

Could not verify from the diff alone: **server-side authorization** — `canEdit` is a client-only affordance and fails closed, but the diff does not show backend enforcement of owner-only writes on `PUT /agents/:id`; a non-owner blocked in the UI must still be rejected by the API, and non-bindable `runtime_id` binds / invalid model strings should 403/validate server-side. This is a pre-existing endpoint the PR does not weaken; flagging for backend confirmation, not blocking.

Jerry-Xin (approved):
APPROVE @ `67e3c99d` — the inline-save error-handling blocker from my prior CR is genuinely fixed. Byte-verified the delta (linear, ahead:1/behind:0):

## ✅ Blocker resolved — inline saves now surface failures
- `InlineEdit.commit()` (`packages/dmloop/src/ui/InlineEdit.tsx`) now has a `catch { Toast.error(t("loop.toast.saveFailed")) }` between `try` and `finally`. On a rejected save the editor stays open, the draft is preserved (not reset), and the user gets an error toast → they can see the failure and retry, no more silent no-op / unhandled rejection.
- `ModelPicker.commit()` and `RuntimePicker.select()` now wrap `await onChange(...)` in `try/catch` → `Toast.error(t("loop.toast.saveFailed"))`.

All three now match the sibling savers in the same panel (instructions/MCP/skills already `Toast.error`), so a 500/403 on a name/description/model/runtime/concurrency edit is now user-visible. That closes the blocking gap.

## 💬 Non-blocking (yujiawei's 🟡s — all fair follow-ups, none blocking)
- `busy` re-entrancy is state-based; a `ref` guard would be tighter (a fast double-commit race is narrow).
- `updateAgent` always carries `name` — under concurrent edits this could clobber a name changed elsewhere; scope the patch to changed fields if that's a real concern.
- `ProviderLogo` `toLowerCase()` without a null-guard — add a safe default.
- `max_concurrent_tasks` renders "undefined" when the field is missing — default it.
- No unit tests for the new interactive components / save + error paths — worth adding coverage for the feature.

Highlights (byte-verified, carried): `canEdit` ownership gate fails closed during the async member-load and on directory-fetch error, applied uniformly across both pickers + InlineEdit; no XSS (values render via JSX/React-escaped, zero `dangerouslySetInnerHTML`); `providerLogo` is a static hardcoded-SVG allowlist; concurrency input validated (`Number.isFinite && >= 1`).

The owner-only rule remains a **client UI gate** over the pre-existing `PUT /agents/:id` — please ensure the backend independently rejects a non-owner PUT / non-bindable runtime / invalid model string (the client gate is a UX affordance, not a security boundary). Thanks for the quick turnaround; credit @OctoBoooot for the original error-handling catch and @yujiawei for the polish notes.

PR #834, opened 2026-07-16
fix(interactive-card): submit peer channel_id when bot DM channelID collapses to self

### LINKED ISSUE #831: InteractiveCard action submit fails for bot DM: channel_id resolved to self_uid
## Summary

Clicking an interactive card button in a **DM with the `notification` system bot** (and likely other bots that dispatch cards through `/v1/internal/notify`) fails with `err.server.message.card_action_invalid`. The submit body carries `channel_id = self_uid` instead of the peer bot UID, so octo-server looks up the message under `fakeChannel(self, self)` and can't find it (real storage is `fakeChannel(self, "notification")`).

## Repro

1. Trigger any flow that sends an approval / interactive card from the `notification` system bot into a user's DM (e.g. any consumer registered in `OCTO_CARD_ACTION_ROUTES`; today `smart-summary` and `docs` go through this path).
2. As the receiver, click any button on the card.
3. Observe: `POST /v1/message/card/action` returns `400 err.server.message.card_action_invalid`. The card stays in loading state until the 10s timeout tick.

Expected: server accepts the action, bot rewrites the card into a terminal state, requester (if any) gets the outcome notification.

## Root cause

`InteractiveCardCell.tsx:301` sends the click payload with:

```ts
submitCardAction({
  messageId: message.messageID,
  channelId: channel.channelID,        // ← naive
  channelType: channel.channelType,
  actionId,
  inputs,
});
```

For a **personal DM whose peer is the `notification` bot**, the receiver-side observation is that `message.channel.channelID === WKSDK.shared().config.uid` (self), **not** `"notification"`. This appears to be a WuKongIM push-path artifact — the notification bot's recv packet uses the receiver as the channel container rather than performing the usual personal-channel peer flip.

Whatever the exact WuKongIM behavior, the client contract is unambiguous:

- octo-server treats `channel_id` for `channel_type=1` (person) as the **peer UID**. It computes the storage key as `fakeChannel(loginUID, req.channel_id)`. When we send `channel_id = self`, it computes `fakeChannel(self, self)` → misses → generic 400 (anti-enumeration).
- The bot's own `fromUID` (in this case `"notification"`) is the reliable peer identity for receiver-side messages. That's the field WKSDK itself uses in `message.send` (getter): `return this.fromUID === WKSDK.shared().config.uid`.

Rather than depending on `channel.channelID` for the peer, we should fall back to `message.fromUID` whenever the personal-channel `channel.channelID` collapses to self.

## Fix

`packages/dmworkbase/src/Messages/InteractiveCard/InteractiveCardCell.tsx`, around line 288–305:

```ts
import WKSDK, { ChannelTypePerson } from "wukongimjssdk";

// …

const { message } = this.props;
const channel = message.channel;
if (!channel) return;

// Personal DM with system bots (e.g. `notification`) can deliver a recv packet
// whose channelID collapses to the receiver's own uid ("receiver as container"
// path in WuKongIM). Fall back to message.fromUID — the authoritative peer for
// received messages — so the submitted channel_id matches server-side storage.
// Group / topic channels are unaffected (channelID is the group number).
const selfUID = WKSDK.shared().config.uid;
const peerChannelId =
  channel.channelType === ChannelTypePerson &&
  channel.channelID === selfUID &&
  message.fromUID
    ? message.fromUID
    : channel.channelID;

submitCardAction({
  messageId: message.messageID,
  channelId: peerChannelId,
  channelType: channel.channelType,
  actionId,
  inputs,
});
```

The fallback is a no-op whenever the existing behavior is already correct (group/topic, or a personal DM where `channelID` is already the peer), so it does not need per-bot special-casing.

## Test additions

`packages/dmworkbase/src/Messages/InteractiveCard/__tests__/cardAction.test.ts` and/or a new `InteractiveCardCell` test — either place is fine — should cover:

- personal DM (`channelType = 1`) with `channel.channelID === selfUID` and `message.fromUID = "notification"` → submit body `channel_id === "notification"`.
- personal DM with `channel.channelID = peer_uid !== selfUID` → submit body `channel_id === peer_uid` (existing path unchanged).
- group DM (`channelType = 2`) with any `channel.channelID` → submit body `channel_id === channel.channelID` (existing path unchanged).
- Defensive: `channel.channelID === selfUID` but `message.fromUID` is missing → keep `channel.channelID` (don't crash).

## Scope of impact

Not specific to the `notification` bot. Every bot whose card is delivered via the same recv path exhibits the same self-loop `channel.channelID`. Concretely:

- `notification` — confirmed.
- Every consumer registered in `OCTO_CARD_ACTION_ROUTES` on octo-server (currently `smart-summary`, `docs` approval when enabled; future consumers on the same generic ingress) — same path, same bug.
- App Bot / third-party bot DMs going through `bot_api` — needs runtime check (WuKongIM may or may not perform peer flip for them). If they do exhibit self-loop, the same fallback fixes them without any code change.
- Group / topic interactive cards — not affected.

The recommended fallback is deliberately universal so we don't have to maintain a per-bot allowlist.

## Not the fix

- **Do not** change octo-server to accept `channel_id === loginUID` as an alias for "peer := message.fromUID". That would weaken the anti-IDOR contract that PR #548 explicitly locked (the peer is server-computed from the declared channel; loosening it lets a caller enumerate message IDs against arbitrary declared channels).
- **Do not** special-case `"notification"` — a per-uid allowlist rots the moment another system bot lands.

## Optional server-side follow-up (separate small PR, not blocking)

In octo-server's `authorizeCardChannelMember` (`modules/message/api_card_action.go:344`), add a WARN log line when `channelType === 1 && channelID === loginUID`. Preserve the current 400 grouping (anti-enumeration), but the log tag would make future "why is the client submitting a self-loop channel?" investigations instant. Tracked separately.

## References

- Server contract: `modules/message/api_card_action.go` (`authorizeCardChannelMember`, `GetFakeChannelIDWith`).
- WuKongIM SDK message construction: `wukongimjssdk/lib/wukongimjssdk.esm.js` — `Message` constructor at `this.channel = new Channel(recvPacket.channelID, recvPacket.channelType)`.
- Related PR (introduced the generic approval-card ingress and its custom decisions): Mininglamp-OSS/octo-server#595, following #588.

an9xyz:
## Summary

Personal DMs with system bots (e.g. `notification`, which dispatches the **docs** and smart-summary approval cards) can deliver a recv packet whose `message.channel.channelID` collapses to the **receiver's own uid** instead of the peer bot. Clicking a card button then submits `channel_id = self`, so octo-server computes `fakeChannel(self, self)`, misses the stored row, and returns `400 err.server.message.card_action_invalid` — the card stays in the loading state until the 10s timeout tick. This is not specific to `notification`: it scopes to every bot delivered over the same recv path (`docs` / `smart-summary` approval cards, and future `OCTO_CARD_ACTION_ROUTES` consumers).

## Related Issue

Fixes #831

## Changes

- Add `resolveCardActionChannelId(...)` in `cardAction.ts`: for a **person** channel whose `channelID === loginUID`, fall back to `message.fromUID` (the authoritative peer for received messages, same source WKSDK's `message.send` getter uses) so the submitted `channel_id` matches server-side storage.
- Wire it into `InteractiveCardCell.handleSubmit` (`selfUID = WKApp.loginInfo.uid`); the submit body now sends the derived `channelId`.
- The fallback is a **no-op** for group/topic (channelID is the group/topic id) and for normal DMs where `channelID` is already the peer — so no per-bot allowlist is needed. Missing `selfUID`/`fromUID` conservatively preserves `channelID` (never crashes).
- Add 5 unit cases: collapse→`fromUID`, already-peer unchanged, group unchanged, and the two defensive fallbacks.

- Affected module(s): `@octo/base` (`dmworkbase`) — `Messages/InteractiveCard`
- Shared layer touched: Messages
- If shared code changed, impact scope: only the InteractiveCard `Action.Submit` closed loop. Rendering, sender-trust gating, and all other message cells are untouched. `resolveCardActionChannelId` is additive and returns the prior value (`channel.channelID`) for every non-collapse case, so existing behavior is preserved.
- Duplicate entry point checked: not applicable

## Testing

- `cd packages/dmworkbase && vitest run src/Messages/InteractiveCard/` → **20 files, 215 tests pass** (incl. the 5 new cases).
- `tsc --noEmit` shows no new type errors in the changed files (residual errors are pre-existing environment noise, identical on untouched cells).

- [ ] Manually verified 

## Checklist

- [ ] Updated documentation

https://claude.ai/code/session_01X6NQFjcxFaNQRHXPs6PV6i

---
_Generated by [Claude Code](https://claude.ai/code/session_01X6NQFjcxFaNQRHXPs6PV6i)_

### DIFF
--- a/packages/dmworkbase/src/Messages/InteractiveCard/InteractiveCardCell.tsx
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/InteractiveCardCell.tsx
@@ -9,7 +9,11 @@ import MessageRow from "../../ui/message/MessageRow";
 import ReplyBlock from "../../ui/message/ReplyBlock";
 import { MessageCell } from "../MessageCell";
 import { t } from "../../i18n";
-import { isRetryableCardActionError, submitCardAction } from "./cardAction";
+import {
+  isRetryableCardActionError,
+  resolveCardActionChannelId,
+  submitCardAction,
+} from "./cardAction";
 import { isAgentProgressCard } from "./cardLayout";
 import { InteractiveCardContent } from "./InteractiveCardContent";
 import { decideCardBody, type CardDecision } from "./renderDecision";
@@ -289,6 +293,16 @@ export class InteractiveCardCell extends MessageCell {
     const channel = message.channel;
     if (!channel) return;
 
+    // person DM 与系统 bot（notification 等）的 recv 包 channelID 可能塌缩为接收人
+    // 自身 uid，直接回传会让服务端 fakeChannel(self,self) miss → 400。回退到权威对端
+    // message.fromUID，确保 channel_id 与服务端存储键一致（群/普通 DM 为 no-op）。
+    const channelId = resolveCardActionChannelId({
+      channelType: channel.channelType,
+      channelID: channel.channelID,
+      fromUID: message.fromUID,
+      selfUID: WKApp.loginInfo.uid,
+    });
+
     // 本次提交代次：使此前提交/超时/新帧作废，且异步回调据此判活。
     const gen = ++this.submitGen;
     this.submitError = null;
@@ -298,7 +312,7 @@ export class InteractiveCardCell extends MessageCell {
 
     submitCardAction({
       messageId: message.messageID,
-      channelId: channel.channelID,
+      channelId,
       channelType: channel.channelType,
       actionId,
       inputs,
--- a/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/cardAction.test.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/__tests__/cardAction.test.ts
@@ -14,7 +14,16 @@ vi.mock("../../../App", () => ({
   },
 }));
 
-import { submitCardAction, isRetryableCardActionError } from "../cardAction";
+// person=1，与真实 ChannelTypePerson 一致（避免测试拉起重型 SDK）。
+vi.mock("wukongimjssdk", () => ({
+  ChannelTypePerson: 1,
+}));
+
+import {
+  submitCardAction,
+  isRetryableCardActionError,
+  resolveCardActionChannelId,
+} from "../cardAction";
 
 const params = {
   messageId: "m1",
@@ -55,6 +64,65 @@ describe("submitCardAction — 请求体（D11 no-data）", () => {
   it("缺字段保守视为已受理", async () => {
     postMock.mockResolvedValue({});
     expect((await submitCardAction(params)).accepted).toBe(true);
+  });
+});
+
+describe("resolveCardActionChannelId — person DM 对端塌缩兜底", () => {
+  const SELF = "self-uid";
+
+  it("person DM channelID 塌缩为 self → 回退到 fromUID（对端 bot）", () => {
+    expect(
+      resolveCardActionChannelId({
+        channelType: 1,
+        channelID: SELF,
+        fromUID: "notification",
+        selfUID: SELF,
+      })
+    ).toBe("notification");
+  });
+
+  it("person DM channelID 已是对端 → 原样返回（既有路径不变）", () => {
+    expect(
+      resolveCardActionChannelId({
+        channelType: 1,
+        channelID: "peer-uid",
+        fromUID: "peer-uid",
+        selfUID: SELF,
+      })
+    ).toBe("peer-uid");
+  });
+
+  it("group（channelType=2）任意 channelID → 原样返回（既有路径不变）", () => {
+    expect(
+      resolveCardActionChannelId({
+        channelType: 2,
+        channelID: SELF,
+        fromUID: "notification",
+        selfUID: SELF,
+      })
+    ).toBe(SELF);
+  });
+
+  it("防御：channelID===self 但 fromUID 缺失 → 保留 channelID，不崩", () => {
+    expect(
+      resolveCardActionChannelId({
+        channelType: 1,
+        channelID: SELF,
+        fromUID: undefined,
+        selfUID: SELF,
+      })
+    ).toBe(SELF);
+  });
+
+  it("防御：selfUID 缺失 → 保留 channelID（无法判定塌缩）", () => {
+    expect(
+      resolveCardActionChannelId({
+        channelType: 1,
+        channelID: "peer-uid",
+        fromUID: "notification",
+        selfUID: undefined,
+      })
+    ).toBe("peer-uid");
   });
 });
 
--- a/packages/dmworkbase/src/Messages/InteractiveCard/cardAction.ts
+++ b/packages/dmworkbase/src/Messages/InteractiveCard/cardAction.ts
@@ -1,4 +1,38 @@
+import { ChannelTypePerson } from "wukongimjssdk";
 import WKApp from "../../App";
+
+/**
+ * 解析卡片动作上行请求的 `channel_id`（person DM 对端塌缩兜底）。
+ *
+ * 服务端把 person 频道的 `channel_id` 当作**对端 uid**，按 `fakeChannel(loginUID,
+ * channel_id)` 定位存储行。正常 1v1 里 `message.channel.channelID` 已是对端；但与
+ * 系统 bot（如 `notification`，docs / smart-summary 审批卡都经它下发）的 DM，其 recv
+ * 包 channelID 会塌缩为**接收人自身 uid**（WuKongIM「receiver as container」路径，未做
+ * 常规 personal 对端翻转），此时若直接回传 channelID=self，服务端算出 `fakeChannel(self,
+ * self)` 必然 miss → 400 `card_action_invalid`。
+ *
+ * received 消息的权威对端是 `message.fromUID`（与 WKSDK `message.send` getter 同源），
+ * 故当 person 频道 channelID 塌缩为 self 时回退到 fromUID。group/topic（channelID=群号）
+ * 与 channelID 已是对端的普通 DM 条件均不成立 → 原样返回（fallback 为 no-op，无需 per-bot
+ * 白名单）。selfUID / fromUID 缺失时同样保守保留 channelID，不崩。
+ */
+export function resolveCardActionChannelId(params: {
+  channelType: number;
+  channelID: string;
+  fromUID?: string;
+  selfUID?: string;
+}): string {
+  const { channelType, channelID, fromUID, selfUID } = params;
+  if (
+    channelType === ChannelTypePerson &&
+    !!selfUID &&
+    channelID === selfUID &&
+    !!fromUID
+  ) {
+    return fromUID;
+  }
+  return channelID;
+}
 
 /**
  * 卡片动作提交（octo/v2 交互闭环，契约 §7.1）。


OctoBoooot (approved):
## Review: fix(interactive-card): submit peer channel_id when bot DM channelID collapses to self (#834)

**Verdict:** Approve
A well-scoped, correct fix for #831: when a system-bot DM's recv packet collapses `channelID` to the receiver's own uid, the card action submitted `channel_id = self` → server `fakeChannel(self,self)` miss → 400. The fallback to `message.fromUID` (the authoritative peer for received messages) is guarded precisely and is a true no-op for every other path. CI green.

## Parity checked
The PR claims the change is additive and a no-op for existing paths (group/topic and normal DMs). Verified against the diff:
- `resolveCardActionChannelId` returns `channelID` **unchanged** unless ALL of `channelType === ChannelTypePerson && selfUID && channelID === selfUID && fromUID` hold. So:
  - **Group/topic** (`channelType !== Person`) → returns `channelID` (the group/topic id). No-op. ✓ (test: group unchanged)
  - **Normal 1v1 DM** (`channelID` is already the peer, `≠ selfUID`) → guard fails on `channelID === selfUID` → returns `channelID`. No-op. ✓ (test: already-peer unchanged)
  - **Missing `selfUID` or `fromUID`** → guard fails → returns `channelID`, never throws. ✓ (both defensive tests present)
- Only the newly-broken collapse case (`channelID === selfUID`, `fromUID` present) changes behavior, returning the peer `fromUID`. The pre-existing submit path (`submitCardAction`, D11 no-data contract) is byte-unchanged; `channelType` is still passed raw. No capability removed.

## Correctness (the primary risk — checked)
- The core question is whether the fallback could ever resolve a *wrong* channel. It can't: the guard fires only when `channelID` literally equals the logged-in uid, which is exactly the collapse signature; in every normal DM `channelID` is the peer (≠ self) so the branch is inert. If a self-sent message ever reached this path (`fromUID === selfUID`), the result equals `selfUID` — no worse than the prior behavior, not a new failure. `message.fromUID` as the received-message peer matches the source WKSDK's `message.send` getter uses, per the PR body — consistent with how the sibling code resolves the peer.
- Wiring in `InteractiveCardCell.handleSubmit` is correct: `channel`/`message` from `this.props`, the derived `channelId` replaces the raw `channel.channelID` in the submit body, and the early `if (!channel) return` guard is preserved.

### Praise
- Fixing this at the `channel_id` derivation with a precise guard (rather than a per-bot allowlist of `notification`/`docs`/`smart-summary`) is the right call — it scopes to the actual failure signature (person channel whose id collapsed to self) and so covers current and future `OCTO_CARD_ACTION_ROUTES` consumers without enumeration. The defensive fallbacks (missing selfUID/fromUID → preserve channelID) and the 5 unit cases covering collapse / already-peer / group / both defensive branches make the behavior contract explicit.

### Net
Correct, minimal, well-tested fix for the bot-DM card-action 400. The fallback is precisely guarded and a verified no-op for group/topic and normal DMs; the submit contract is unchanged; CI is green (`size/S`, no governance blocker). **Approve.**

yujiawei (approved):
# Code Review — PR #834 (octo-web)

Reviewed at head SHA `f9b7225d1fd50165986417e719fd71f1a1bc7cce`. Verified against the merge base and ran the affected unit suite locally.

## Summary

This fixes a real, well-scoped bug: interactive-card `Action.Submit` requests fail with `400 card_action_invalid` when a bot-DM recv packet's `message.channel.channelID` collapses to the receiver's own uid, because the server computes `fakeChannel(self, self)` and misses the stored row. The fix adds a pure helper `resolveCardActionChannelId(...)` that falls back to `message.fromUID` (the authoritative peer for received messages) only when the person-channel `channelID` equals `selfUID`, and wires it into `InteractiveCardCell.handleSubmit`.

The change is minimal, additive, and correct. I recommend approval.

## Spec / scope compliance

- Implements exactly what the linked issue (#831) describes — no over-building, no unrelated changes.
- The three touched files map 1:1 to the change: the helper, its call site, and the tests.
- No new user-visible entry points, no new flags, no per-bot allowlist. The fallback is a genuine no-op for group/topic channels and for normal DMs where `channelID` is already the peer.

## Correctness

The guard condition is tight and correct:

```ts
channelType === ChannelTypePerson && !!selfUID && channelID === selfUID && !!fromUID
```

- `ChannelTypePerson === 1` is consistent with every other usage in the repo, so the fallback cannot fire for group (`2`) or topic channels.
- The transform is only reachable for **trusted** cards: `handleSubmit` early-returns unless `decision.kind === "card" && decision.interactive`, and `renderDecision` derives trust solely from the server-authoritative `message.fromUID` (webhook `iwh_` prefix or `orgData.robot === 1`). A human-to-self card is never interactive, so the "channelID collapsed to self" branch cannot misfire on a legitimate self-DM to a normal user.
- Defensive fallbacks are conservative: a missing `selfUID` or `fromUID` preserves `channelID` rather than throwing. `loginInfo.uid` is typed `string | undefined`, and the `!!selfUID` guard handles the `undefined` case as a no-op.

## Tests

`vitest run src/Messages/InteractiveCard/__tests__/cardAction.test.ts` passes locally (12/12, including the 5 new cases): collapse→`fromUID`, already-peer unchanged, group unchanged, and the two defensive `undefined` fallbacks. The `wukongimjssdk` mock pins `ChannelTypePerson: 1`, matching the real enum and the existing mocking convention used across `dmworksummary`.

## Non-blocking notes (P2 / advisory)

1. **`WKApp.loginInfo.uid` access.** The call site uses `WKApp.loginInfo.uid` directly. This is safe here: `loginInfo` is a statically-initialized singleton (`static loginInfo = new LoginInfo()`, never null) and every other call site in the codebase accesses `.uid` the same way. Optional chaining would be redundant. No change required.

2. **Forwarded / merged-history messages.** The fallback trusts `message.fromUID` as the peer. For a normal received bot card this is correct. In a hypothetical forwarded-card scenario, `fromUID` could point to the original author rather than the current DM peer — but the same `fromUID` also governs the render-trust gate, so a card that renders as interactive at all has already been vetted through that uid. Low probability and largely pre-existing; worth a mention only as a future edge to keep in mind, not a blocker.

## Coverage / what I could not verify from the client diff

- The server-side `fakeChannel(loginUID, channel_id)` resolution and the exact recv-path uid-collapse behavior in WuKongIM ("receiver as container") are asserted by the PR description and the code comments; they are not directly verifiable from this frontend diff. The client-side contract (send peer uid, not self) is sound regardless.
- The full bot-DM round-trip is covered here only by unit tests, not an integration/e2e run (the author notes this needs a running stack). Given the surgical scope and the server-side card-action contract test referenced, unit coverage is adequate for this fix.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #834 (octo-web)

## Summary

This PR bundles a focused bug fix with a round of UI polish. The headline change is `resolveCardActionChannelId` in `packages/dmworkbase/src/Messages/InteractiveCard/cardAction.ts` — it fixes a real production issue where card action submissions from bot DMs (e.g. notification bot, smart-summary approval cards) would send `channel_id=self` to the server, causing `fakeChannel(self,self)` lookup miss → 400 `card_action_invalid`. The fix falls back to `message.fromUID` when the person-DM channelID collapses to the logged-in user's own UID.

Alongside, the PR consolidates agent status dots into a shared `StatusDot` component with CSS custom properties (`--dot-*`), replaces Semi `Typography.Title` with plain `<h2>` across all loop pages, renames "Loop" → "Task" throughout i18n, adds beta badges to the Tasks and Personal pages, and standardizes hardcoded hex colors to Semi CSS variables.

## Verification

- ✅ **Data-flow: `resolveCardActionChannelId` inputs** — `message.fromUID` is the WuKong SDK's authoritative sender property, already consumed at five other sites in `InteractiveCardCell.tsx` (avatar clicks, sender classification, thread routing). `WKApp.loginInfo.uid` is the same self-UID used in `Screenshot/index.tsx` for the identical self-check pattern. Both inputs are populated at the call site.
- ✅ **Guard precision** — The fallback only fires when all four conditions hold: `channelType === ChannelTypePerson` AND `channelID === selfUID` AND `selfUID` is truthy AND `fromUID` is truthy. Group/topic channels and normal DMs where channelID already equals the peer are unaffected (no-op return).
- ✅ **Test coverage for core fix** — `cardAction.test.ts` adds 5 cases: collapsed-self → fromUID, normal peer → passthrough, group → passthrough, fromUID missing → safe fallback, selfUID missing → safe fallback. `ChannelTypePerson` mock value `1` matches the real SDK constant (confirmed across `dmworksummary` tests).
- ✅ **`AGENT_STATUS_COLOR` removal clean** — `grep` confirms zero remaining consumers after removal from `meta.tsx`.
- ✅ **`agentStatus.*` i18n keys complete** — All six `KNOWN_DOT_STATUSES` (`idle`, `working`, `offline`, `error`, `unstable`, `archived`) have matching keys in both `en-US.json` and `zh-CN.json`.
- ✅ **`--dot-*` CSS variables defined** — All four (`--dot-idle`, `--dot-working`, `--dot-offline`, `--dot-error`) defined in `:root` at `packages/dmloop/src/pages/loop.css:820-823`, consumed by `.loop-status-dot[data-status]` rules and `mentionExtension.agentDotColor`.
- ✅ **Removed i18n keys (`agent.statusOnline`/`statusOffline`) have no dangling refs** — The old `AgentDetailPage` was the sole consumer; it now uses `loop.agentStatus.*`. The `loop.squad.statusOffline` key (different namespace) is untouched.
- ✅ **`Action.OpenUrl` role="link" test** — `renderOctoCard.test.tsx` pins the SDK contract: OpenUrl → `role="link"`, Submit → `role="button"`, keyed by title to catch role-swap regressions.
- ✅ **Accessibility** — `StatusDot` uses `decorative` prop correctly: `aria-hidden="true"` when status text is shown nearby; `role="img"` with `title`/`aria-label` when the dot is the sole status indicator. `prefers-reduced-motion` disables the working-status pulse animation.

Static analysis only at head `f9b7225d1fd50165986417e719fd71f1a1bc7cce`; build and tests not executed in this environment.

## Findings

### P2 — `agentDotColor` returns CSS variable references without fallback (`packages/dmloop/src/ui/mentionExtension.ts:75`)

The old code returned `var(--semi-color-success, #23a55a)` with embedded fallbacks. The new code returns bare `var(--dot-idle)`, `var(--dot-working)`, `var(--dot-error)`, `var(--dot-offline)` — no fallback. These variables are defined in `:root` inside `loop.css`, which is loaded in the loop page context where the mention popup renders. Works today, but if the mention extension is ever consumed outside the loop context, or if stylesheet load order changes, the dot color resolves to `initial` (effectively transparent).

Consider: `return "var(--dot-idle, var(--semi-color-success, #23a55a))"` — preserves the single-source-of-truth intent while keeping a safe fallback.

## Data-Flow Tracing

|---|---|---|
| `channel.channelType` | WuKong SDK `message.channel` | ✅ SDK populates on recv |
| `channel.channelID` | WuKong SDK `message.channel` | ✅ SDK populates on recv |
| `message.fromUID` | WuKong SDK message property | ✅ Used at 5+ sites in same file |
| `WKApp.loginInfo.uid` | App login state | ✅ Used in `Screenshot/index.tsx` for same self-check |
| `resolveCardActionChannelId(channelId)` | → `submitCardAction({ channelId })` | ✅ Replaces raw `channel.channelID` at line 312 |
| `normalizeAgentStatus(agent.status)` | → `StatusDot` `data-status` attr + i18n key | ✅ Clamped to `KNOWN_DOT_STATUSES`; default "offline" |
| `agentDotColor(status)` → `var(--dot-*)` | → `LoopMentionItem.dotColor` → inline style | ✅ loop.css `:root` loaded in mention popup context |

## Blindspot Checklist

- **C1 (dual-path parity)** — `resolveCardActionChannelId` is called in a single site (`InteractiveCardCell.handleAction`). No symmetric submit path to check. Clear.
- **C2 (control-flow ordering)** — The resolver runs before `submitCardAction`, in correct order. No nested reuse or double-invocation. Clear.
- **C3 (auth boundary)** — No auth/permission changes. N/A.
- **C5 (build)** — No build/packaging changes. N/A.
- **C6 (governance)** — N/A.

N/A — first review pass.

- i18n "Loop" → "Task" rename is consistent across en-US.json and zh-CN.json; "Sub-issues" → "Sub-tasks" and activity-action strings updated in lockstep.
- `Typography.Title` → `<h2 className="loop-page__title">` replacement is mechanical and complete across all 6 loop pages + Settings.
- Beta badge CSS (`.loop-page__title-beta`, `.dmpersonal-sidebar__beta`) is identical in both locations — consistent styling.
- `AUTOPILOT_RUN_DOT` color standardization to Semi CSS variables with fallbacks is clean.
- Old `.loop-agent-row__dot`, `.loop-sqd__mem-dot`, `.loop-adp__pill-dot`, `.loop-adp__ins-dot` inline styles properly removed; sizing/border now inherited from shared `.loop-status-dot`.
- `Action.OpenUrl` link styles in `index.css` correctly scoped to `.ac-pushButton[role="link"]` — does not affect Submit/ToggleVisibility/CopyToClipboard buttons (all `role="button"`).

## Verdict: APPROVED

The core fix is correct, well-documented, and defensively coded with appropriate fallbacks. Test coverage exercises all branches. The UI consolidation is clean. The single P2 (missing CSS variable fallbacks in `agentDotColor`) is a robustness improvement, not a blocker.

[Octo-Q] verdict: APPROVE — no P0/P1 found; core channel_id collapse fix is correct with defensive fallbacks and thorough tests; single P2 is non-blocking robustness nit.

Jerry-Xin (approved):
**APPROVE** — Correct, tightly-scoped routing fix for the bot-DM channelID collapse. Fallback resolves from trusted local state, not spoofable payload, and the normal path is untouched.

### ✅ Targeting correctness (the core risk)
- **Collapse detection is precise.** `resolveCardActionChannelId` triggers the fallback only when *all* hold: `channelType === ChannelTypePerson` **and** `selfUID` present **and** `channelID === selfUID` **and** `fromUID` present. Group/topic channels (channelType≠person, channelID=group id) and normal 1v1 DMs (channelID already the peer) all fail the condition → no false-positive. A collapsed person-DM where channelID equals my own uid is the only match → no false-negative for the target case.
- **Fallback resolves the correct intended recipient.** It falls back to `message.fromUID` — the authoritative sender of the *received* card frame (same source as the SDK's `message.send` getter). For a bot DM, that is exactly the peer bot (`notification`, docs/smart-summary approver, etc.) whose store key the server expects. Not a group, not a broadcast, not an attacker-chosen channel.
- **Normal path is unchanged.** Any input that doesn't match the collapse predicate returns `channelID` verbatim — group cards and standard DMs keep their original targeting (no-op fallback, no per-bot allowlist needed).

### ✅ Injection / trust of the source
All inputs (`channel.channelType`, `channel.channelID`, `message.fromUID`, `WKApp.loginInfo.uid`) come from **trusted local client model state** — the SDK-materialized received-message object and the logged-in user's own uid — not from the card payload or any peer-controlled body field. So the fallback cannot be spoofed to redirect the submit to an attacker-chosen channel. Note also `submitCardAction` deliberately omits `data` (server re-derives it from the stored frame), keeping the submit body minimal.

### ✅ Null / empty handling (fails safe)
`!!selfUID` and `!!fromUID` guards mean a missing selfUID or fromUID keeps the original `channelID` rather than substituting `undefined`/empty. The call site also early-returns when `!channel`. No path submits to an empty/undefined target.

### ✅ Test coverage (behavioral)
The `resolveCardActionChannelId` suite asserts return values across all branches: (1) person-DM collapse → returns `fromUID`; (2) person-DM channelID already peer → passthrough; (3) group channelType → passthrough; (4) channelID===self but `fromUID` missing → keeps channelID; (5) `selfUID` missing → keeps channelID. These are behavioral assertions on the resolved value, covering collapse detection, peer fallback, normal-path passthrough, and both null-guards.

### CI / gate
Build, check-sprint, code-review and all other required checks green; non-draft; reviewDecision APPROVED, mergeStateStatus CLEAN.

PR #835, opened 2026-07-16
feat(search): add aggregated global chat search

LLwill:
## Summary

- add conversation aggregation and drill-down for global chat-history search
- share chat-history and file filter state while preserving conditions across tabs
- align the responsive result and filter layout with the WeCom-style interaction, including distinct empty states
- highlight the target message briefly after locating it in a conversation

## Validation

- 7 test files, 73 tests passed
- focused Stylelint passed
- i18n check passed
- `git diff --check` passed
- production web build passed (existing bundle/eval warnings only)


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #835 (octo-web)

## Summary

This PR adds aggregated global chat search: a two-pane layout where the left pane shows conversations grouped by channel with match counts and preview snippets (via a new `messages/_search_global_groups` backend endpoint), and the right pane drills into the selected conversation's full message results with pagination. It also lifts filter state from per-panel local state to shared parent-level state, adds a sidebar filter mode, wires abort signals into the shared API client, and adds a locate-highlight for messages navigated from search results. 30 files changed (+2465/-565). The architecture is sound and the implementation is thorough; one P2 data-flow race and a P2 observer-churn issue found.

## Verification

- ✅ **Abort signal propagation** — `APIClient.ts` now threads `signal` into both `get` and `post`. `dataSource.searchMessages` passes `query.signal` to the API call. The overview search (L1) and initial result search (L2) both use `AbortController` correctly with sequence guards and abort-aware error handling.
- ✅ **Filter state lift** — `filters` moved from `GlobalContentSearchPanel` local state to `GlobalSearch` class state, shared between messages and files tabs. `GlobalSearchFilterPanel` gains `mode="sidebar"` with `updateDraft` auto-applying in sidebar mode. Reset button in parent correctly calls `defaultGlobalSearchFilters()`.
- ✅ **Locate highlight** — `Conversation/vm.ts` adds `highlightLocatedMessage` flag that is only true when `lcateMessageSeq` is an explicit positive number, preventing spurious highlight on normal conversation open. `initMessage.locateRemind = true` is guarded by `initMessage &&` null-check.
- ✅ **ChannelSearchEmpty export** — exported as `SearchEmpty as ChannelSearchEmpty` at `packages/dmworkbase/src/Components/ChannelSearch/index.tsx:1818`. Import in `GlobalChatSearchPanel` resolves correctly.
- ✅ **Test coverage** — new tests for `GlobalMessageSearchService`, `model.ts` (`canRunGlobalGroupSearch`, `drillDownFilters`), `filterState.ts` (`hasGlobalSearchCriteria`, `selectedGlobalSearchFilterValueCount`), and updated source-guards in `isActive.test.tsx` and `clickRouting.test.ts`.

Static analysis only at head `e5355babff8caeb9c09d5a69576c4115e872c77a`; build and tests not executed in this environment.

## Findings

### P2 — `loadMore` omits abort signal, enabling stale-data race on conversation switch (`packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts:305`)

The initial result search (line ~267) correctly passes `controller.signal` from `resultAbortRef` into `dataSource.searchMessages`. The `loadMore` callback (line ~305) does not:

```ts
const response = await dataSource.searchMessages({
  tab: "messages",
  keyword,
  filters: drillDownFilters(filters, selectedConversation),
  cursor: result.nextCursor,
  limit: PAGE_SIZE,
  // no signal
});
```

When the user switches conversations while a pagination fetch is in flight, the effect cleanup at line ~258 aborts the initial-search controller and resets `loadingMoreRef`, but the in-flight `loadMore` request keeps running. When it resolves, its `setResult` call writes the old conversation's items, `nextCursor`, and `hasMore` into state — overwriting the new conversation's initial results. The stale `nextCursor` could theoretically cause a subsequent sentinel-triggered `loadMore` to fetch from the wrong conversation.

Practical impact is bounded (the new conversation's search usually completes quickly and overwrites), but the race is reachable in production.

Fix: pass `resultAbortRef.current?.signal` into the `searchMessages` call inside `loadMore`, and add `if (resultAbortRef.current?.signal.aborted) return;` before `setResult`.

### P2 — IntersectionObserver recreated on every `loadMore` identity change (`packages/dmworkbase/src/features/globalChatSearch/GlobalChatSearchPanel.tsx:47`)

The `useEffect` at line 47 depends on `search.loadMore`, which is a `useCallback` whose dependency array includes `result.nextCursor` and `result.hasMore`. Both change on every pagination response, causing the observer to be disconnected and recreated after each page. This is not a correctness bug — the sentinel is re-observed and cleanup still fires on unmount — but it creates unnecessary DOM observer churn.

Consider stabilizing `loadMore` via a ref pattern (store the latest callback in a ref, expose a stable wrapper) so the observer only recreates when `hasMore` actually flips.

## 数据流回溯 (Data-Flow Trace)

| Consumed Data | Upstream Source | Reaches Consumer? |
|---|---|---|
| `overview.conversations` | `GlobalMessageSearchService.searchGroups` → `mapOverviewResponse` → filters groups by `channel_id` truthiness + `channel_type` number type | ✅ Yes, filtered groups flow to conversation list |
| `selectedConversation.preview` | `mapPreview(group.preview, ...)` → `isCombinedSearchHit`/`isMessageSearchHit` type guards → `mapCombinedHit`/`mapMessageHit` | ✅ Yes, preview items used as initial `result.items` when a conversation is selected |
| `result.items` (drill-down) | `dataSource.searchMessages` → `mapMessagesResponse` → `foldResponse` | ✅ Yes, items flow to `MixedResultItem` rendering |
| `result.nextCursor` / `hasMore` | `foldResponse` from `pagination.has_more` / `pagination.next_cursor` | ✅ Yes, consumed by `loadMore` guard and sentinel observer |
| `filters` (shared) | `GlobalSearch.state.filters` → passed to both `GlobalChatSearchPanel` and `GlobalContentSearchPanel` | ✅ Yes, flows to `canRunGlobalGroupSearch` gate and `cleanGlobalFilters` wire body |
| `highlightLocatedMessage` | `requestMessagesOfFirstPage(lcateMessageSeq)` → `typeof === "number" && > 0` | ✅ Yes, only true for explicit positive seq from search navigation |

## 盲点 Checklist (C1–C6)

- **C1 — 双路径 parity**: N/A. No paired add/remove or subscribe/unsubscribe operations in this PR. The filter panel's popover/sidebar modes share the same mutation handlers via `updateDraft`; both paths verified consistent.
- **C2 — Control-flow ordering**: Clear. The abort-controller pattern in `useGlobalChatSearch` uses separate controllers for overview and result effects. No nested/double-invocation risk. `loadingMoreRef` prevents concurrent `loadMore` calls.
- **C3 — 授权边界**: Clear. The new `messages/_search_global_groups` endpoint is called through `APIClient.shared.post`, which applies the same auth/interceptor chain as all other API calls. No raw fetch or auth bypass.
- **C4 — 授权生命周期**: N/A. No auth/lifecycle changes in this PR.
- **C5 — Build/note**: N/A. No build config, packaging, or deployment changes.
- **C6 — 治理/策略**: N/A. No governance or policy documents changed.

## 跨轮 Blocker 复检 (R6)

N/A — automated review review.

- `canRunGlobalGroupSearch` correctly gates the overview search on keyword or applicable filters (messages-tab filters for messages, file-tab filters for files).
- `drillDownFilters` correctly narrows scope to the selected conversation while preserving user-selected filters (sender, date, etc.).
- `cleanGlobalFilters` is used for the L1 overview request; `dataSource.searchMessages` (which also calls `cleanGlobalFilters` internally via `toGlobalRequestBody`) handles the L2 drill-down.
- `GlobalSearchFilterPanel` sidebar mode: `updateDraft` auto-applies on every mutation; `clearAll` is only reachable in popover mode (action buttons hidden in sidebar).
- i18n: all new user-facing strings use `t("base.globalSearch.aggregated.*")` keys; both `en-US.json` and `zh-CN.json` updated with matching keys.
- CSS: new stylesheets are self-contained with BEM-like scoping (`wk-global-chat-search-layout__*`, `wk-global-chat-search-panel__*`).

## Verdict: APPROVED

The PR is well-architected with clean separation of concerns (service layer → bridge hook → feature panel → UI layout). The two P2 findings are non-blocking: the stale-data race in `loadMore` is reachable but self-correcting, and the observer churn is a performance nit. Neither makes a working path unavailable or produces persistent incorrect data. Recommend addressing the `loadMore` abort signal in a follow-up.

[Octo-Q] verdict: APPROVE — no P0/P1 blockers found; two P2 items (loadMore missing abort signal, IntersectionObserver churn) are non-blocking.

packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts
```diff
@@ -0,0 +1,412 @@
...
+      const response = await dataSource.searchMessages({
+        tab: "messages",
+        keyword,
+        filters: drillDownFilters(filters, selectedConversation),
+        cursor: result.nextCursor,
+        limit: PAGE_SIZE,
+      });
+      setResult((current) => ({
+        status: "ready",
+        items: [...current.items, ...response.items],
```
OctoBoooot, line 389: **[major]** `loadMore` issues `dataSource.searchMessages` (:374) with no abort signal and, on resolve, unconditionally appends `[...current.items, ...response.items]` here — with no guard that `selectedConversation` is still the one this request was for. The main result effect (:313-360) guards exactly this via `controller.signal` + `if (controller.signal.aborted) return`; `loadMore` was left without it. Failing input: on conversation A's list (hasMore), scroll → triggers loadMore (in-flight request R); before R resolves, click conversation B. B's selection effect aborts its own controller and loads B page 1, but R has no signal so it is not cancelled — R then appends A's page-2 messages onto B's list and carries A's nextCursor/hasMore into B's state, so the next loadMore paginates A's cursor against B. Fix: thread an AbortController into loadMore (GlobalSearchQuery already accepts `signal`, types.ts:49) and bail on `signal.aborted` / a changed `selectedConversation.key` before setResult, mirroring the result effect.

packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts
```diff
@@ -0,0 +1,412 @@
...
+      .catch((error) => {
+        if (controller.signal.aborted || isCancelledRequest(error)) return;
+        setResult((current) => ({
+          ...current,
+          status: current.items.length > 0 ? "ready" : "error",
+          isLoadingMore: false,
+        }));
+      });
+
+    return () => controller.abort();
```
OctoBoooot, line 359: **[praise]** The overview and result effects handle stale requests correctly — both abort their AbortController and clear the debounce timer on cleanup, the overview additionally rejects out-of-order responses via `sequenceRef` + the response `sequence` check, and `loadingMoreRef` guards concurrent loadMore entry. This is the right discipline; the loadMore omission above is the one gap in an otherwise careful cancellation model.

OctoBoooot (changes requested):
## Review: feat(search): add aggregated global chat search (#835)

**Verdict:** Request changes
Solid, well-tested feature — the two heavily-refactored files preserve their filter/search capabilities (parity checked below) and the core hook cancels its main searches correctly. But `loadMore` in the search hook issues its paged request without an abort signal or a post-await guard, so a conversation switch mid-load appends the previous conversation's messages onto the new one. That's a real result-corruption bug.

Code CI is green (Build / secret-scan / osv-scan / dependency-review / pr-title-lint pass; check-sprint + satisfy-code-review skipped = governance). Large change (30 files, +2465/-565) → `needs-human-review`.

## Parity checked
This PR refactors two large existing files; I diffed base→head to confirm nothing was silently dropped:
- **`GlobalSearchFilterPanel.tsx` (+346/-229)** — additive `mode: "popover" | "sidebar"` split. Every filter dimension (sender, channels, member, channel-types, content-types, file-types, file-size, date presets + custom range) still renders in both modes; the popover path is preserved behind `mode === "popover"`. Sidebar hides the clear/apply footer but the parent supplies an equivalent reset + live-apply. No filter lost.
- **`GlobalSearch/index.tsx` (+319/-201)** — messages tab swapped to the new aggregated `GlobalChatSearchPanel`; filter state hoisted up and shared across tabs via one sidebar panel. Message search, locate/navigate, filtering, and load-more all remain; files tab still uses `GlobalContentSearchPanel`. Redesign, not a capability drop.
- **`GlobalContentSearchPanel.tsx` (+12/-55)** — the removed in-panel filter toolbar/state was genuinely moved up to `index.tsx` (which now owns filter state), not lost; the panel gained better empty/no-result hints.

## Major
- `packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts:362-400` (`loadMore`) — the paged `dataSource.searchMessages` call (:374) passes **no abort signal** and, on resolve, **unconditionally appends** `[...current.items, ...response.items]` (:383) with no check that the selected conversation is still the one the request was issued for. The main result effect (:313-360) guards exactly this with `controller.signal` + `if (controller.signal.aborted) return`; `loadMore` was left without it. Failing input: on conversation A's result list (with `hasMore`), scroll to trigger `loadMore` → in-flight request R; before R resolves, click conversation B. B's selection effect aborts its own controller and loads B's page 1, but R has no signal so it isn't cancelled — when R resolves it appends **A's page-2 messages onto B's list**, and carries A's `nextCursor`/`hasMore` into B's state. Result: mixed-conversation messages under the wrong header, and subsequent `loadMore` paginates with A's cursor against B. Fix: thread an `AbortController` into `loadMore` (the query already accepts `signal`, types.ts:49) and bail on `signal.aborted` / a stale `selectedConversation.key` before `setResult`, mirroring the result effect.

## Minor
- `packages/dmworkbase/src/ui/GlobalChatSearchLayout/index.tsx` — the result header renders `result.countLabel || result.title`, but `countLabel` is always truthy when a conversation is selected, so the passed `title: selected?.name` never shows (dead prop for the selected case). Confirm intent: show the conversation name as the header with the count as a subline, or drop the unused `title`.
- `packages/dmworkbase/src/Components/GlobalSearch/GlobalSearchFilterPanel.tsx` (file-size inputs, sidebar mode) — each keystroke calls `onApply` synchronously → parent setState → the `useEffect([filters])` rewrites the size inputs (bytes→KB round-trip). Typing "1500" fires 4 apply cycles and a mid-edit value can be momentarily rewritten by the effect. Debounce the size apply in sidebar mode, or apply on blur.

### Praise
- The main overview and result effects are textbook: both abort their `AbortController` and clear the debounce timer on cleanup, `sequenceRef` + the response `sequence` check reject out-of-order overview responses, and `loadingMoreRef` guards concurrent `loadMore`. The stale-request discipline is right everywhere **except** the `loadMore` path — which is exactly why that one omission stands out as the finding.
- The `conversationPresentation` mapping handles the thread (community-topic) vs DM vs group cases distinctly for name/subtitle/avatar, with sensible fallbacks — a fiddly area handled carefully.

### Net
A well-structured, well-tested feature with clean parity on the refactored surfaces and correct cancellation on the primary searches. The one real defect is `loadMore` issuing an uncancelled, unguarded paged request, so a conversation switch during pagination corrupts the result list. **Request changes** on that one major; the two minors are non-blocking. Once `loadMore` mirrors the result effect's signal + stale guard, this is an approve. `needs-human-review` (large change) applies.

Jerry-Xin (dismissed):
## Review: feat(search): add aggregated global chat search (#835)

**Verdict:** Request changes

I focused this pass on the security surface of an aggregated **global** chat search (cross-conversation data exposure) and independently re-verified the pagination defect a peer already flagged. Build / secret-scan / osv-scan / dependency-review / dependency all green; `code-review` FAILURE is the existing peer CR, not a build failure.

## Security review (aggregated global search — the core risk surface)

Byte-verified against the head commit:

- **Authorization scope — clean.** `GlobalMessageSearchService.searchGroups` posts `{ keyword, sequence, filters }` to `messages/_search_global_groups`; `cleanGlobalFilters` (`Components/GlobalSearch/apiAdapter.ts`) only emits **narrowing** params (`sender_ids`, `member_uids`, `channel_ids`, `channel_types`, `sent_at_from/to`, `content_types`/file filters). There is **no** admin / all-spaces / global-scope flag, and `getSelfUid()` is used solely to drop self from the member filter. The frontend never requests an over-broad scope; result scoping is delegated to the backend endpoint. No client-side authorization filtering that could be bypassed.
- **Space isolation — enforced.** `APIClient` auto-injects `X-Space-Id` (via `spaceIdCallback`) on every request through the axios request interceptor, so this search carries the current space scope like all other calls. No cross-space leak on the request path.
- **XSS in highlight rendering — safe.** Snippets/highlights render via `dangerouslySetInnerHTML={{ __html: sanitizeHighlight(...) }}` (`Components/GlobalSearch/item-message.tsx`, `item-file.tsx`, etc.). `Components/GlobalSearch/sanitize.ts` escapes `& < > " '` and restores only literal `<mark>`/`</mark>`, so injected markup in message content/sender/file names is neutralized.
- **Query injection / ReDoS — none.** The query is a structured JSON field, not interpolated into a URL/path. The only `new RegExp` in this path uses a fixed placeholder constant, not user input.
- **PII / over-exposure — bounded by the above.** No client over-broad request; exposure is whatever the backend authorizes for the current user + space.

No security blocker.

## Blocking

- **`bridge/globalChatSearch/useGlobalChatSearch.ts` `loadMore` (~:362-400)** — this is a real result-corruption defect and I confirmed it independently. The paged `dataSource.searchMessages` call (~:374) passes **no** abort signal and, on resolve, unconditionally appends `[...current.items, ...response.items]` (~:383) with no check that the selected conversation is still the one the request was issued for. The primary result effect (~:333-351) guards exactly this with `controller.signal` + `if (controller.signal.aborted) return`. Repro: on conversation A (with `hasMore`), trigger `loadMore` → in-flight request R; switch to conversation B before R resolves. B's effect aborts its own controller and loads B, but R is uncancelled, so R appends **A's next page onto B's list** and carries A's `nextCursor`/`hasMore` into B's state, so the next `loadMore` paginates A's cursor against B. Fix: thread an `AbortController` into `loadMore` (the query already accepts `signal`) and bail on `signal.aborted` / a stale `selectedConversation.key` before `setResult`, mirroring the result effect. (Aligns with the peer request-changes on the same line.)

## Non-blocking

- `ui/GlobalChatSearchLayout/index.tsx` — header renders `result.countLabel || result.title`; `countLabel` is truthy whenever a conversation is selected, so `title` is effectively dead for that case. Confirm intent or drop the unused prop.
- `Components/GlobalSearch/GlobalSearchFilterPanel.tsx` (sidebar file-size inputs) — each keystroke calls `onApply` synchronously, and the `useEffect([filters])` bytes→KB round-trip can rewrite mid-edit. Debounce or apply on blur in sidebar mode.

## Tests
Good behavioral coverage of the model/filter/service paths: `Service/__tests__/GlobalMessageSearchService.test.ts` asserts the endpoint + cancellation signal, `bridge/globalChatSearch/__tests__/model.test.ts` covers `canRunGlobalGroupSearch` and `drillDownFilters` (scope replacement to the selected conversation), plus `filterState`, `isActive`, `clickRouting`. Authorization scoping itself is backend-enforced and not frontend-testable. Consider a focused test for the `loadMore` stale-guard once fixed.

## Net
Security posture is sound — scope is backend-enforced with no over-broad client request, space isolation is carried, and highlight rendering is sanitized. The one blocker is the uncancelled/unguarded `loadMore` paged request corrupting the result list on conversation switch. Fix that (signal + stale guard mirroring the result effect) and this is an approve. `needs-human-review` (XL change) applies.

packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts
LLwill: Fixed in `2e92224e`. `loadMore` now reuses the active result request signal, passes it to `searchMessages`, and verifies the active controller before mutating result or loading state. Added a regression test that switches from conversation A to B while A pagination is pending, then resolves the stale A request and verifies B remains unchanged.

LLwill:
Also addressed the review notes:

- Removed the unused result `title`; the intended header is the related-message count only.
- Kept sidebar file-size filtering live: integer KB values round-trip without changing the typed value, and searches are already debounced by 300 ms. Applying only on blur would conflict with the panel’s immediate-filter behavior.
- Kept the IntersectionObserver lifecycle as-is: recreating it after pagination lets an intersecting sentinel continue loading until the viewport is filled. Stabilizing only the callback could stop that progression while the sentinel remains intersecting.

yujiawei (changes requested):
# Code Review — PR #835 (octo-web)

**Head SHA reviewed:** `e5355babff8caeb9c09d5a69576c4115e872c77a`
**Base:** `main` (merge-base `1bded39a`)
**Scope:** 30 files, +2465 / −565 — aggregated global chat search (L1 conversation overview + L2 drill-down), shared filter sidebar, target-message highlight.

This review is split into two gates: **spec compliance** first, then **code quality**. Both must pass to approve.

---

## 1. Spec-compliance gate

The PR summary promises: (a) conversation aggregation + drill-down, (b) shared chat-history/file filter state across tabs, (c) WeCom-style responsive layout with distinct empty states, (d) brief highlight of the target message after locating it.

- **Aggregation + drill-down** — `useGlobalChatSearch.ts` (L1 groups → L2 messages), `GlobalChatSearchPanel.tsx`, `GlobalChatSearchLayout/index.tsx`. ✅
- **Shared filter state across tabs** — `filters` lifted into `GlobalSearch` component state and passed to both the messages and files panels + the shared sidebar (`GlobalSearch/index.tsx`). ✅
- **Responsive layout + distinct empty states** — new CSS + `ChannelSearchEmpty` now takes `emptyHint`/`noResultsHint`. ✅
- **Target-message highlight** — `Conversation/vm.ts` threads a `highlightLocatedMessage` flag so `locateRemind` only fires on an explicit positive locate seq, not on every conversation open. ✅ Good defensive guard.

No unauthorized scope creep that blocks. (See Q3 for dead code left behind.)

**Spec: ✅**

---

## 2. Code-quality gate

### P1 — `loadMore` drill-down pagination can contaminate the wrong conversation

`packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts` — `loadMore` (L362–401).

Every other async request in this hook is guarded by an `AbortController` + post-`await` `signal.aborted` check (overview effect L257–309, drill-down effect L316–359). `loadMore` is the sole exception: it passes **no** `signal` to `dataSource.searchMessages` and does **no** staleness check after the `await`.

Sequence that breaks it:
1. Conversation A selected; user scrolls, `IntersectionObserver` fires `loadMore` → `searchMessages(A, cursor_A)` in flight (unabortable).
2. User selects conversation B. The drill-down effect aborts `resultAbortRef`, resets `loadingMoreRef=false` (L317) and sets `result` to B's preview.
3. A's page-2 response resolves and runs the functional update at L382–388:
   ```ts
   setResult((current) => ({
     status: "ready",
     items: [...current.items, ...response.items],   // A's messages appended onto B
     hasMore: response.hasMore,
     nextCursor: response.nextCursor,                 // now cursor_A, under B's scope
     ...
   }));
   ```
   → A's messages are appended under conversation B, and B's `nextCursor` is overwritten with A's cursor. The next `loadMore` then pages with B's `drillDownFilters` but A's cursor, compounding the corruption.

This is a user-visible correctness defect in the flagship drill-down (wrong-conversation messages shown), and the fix mirrors code already present two effects above. Suggested fix: give `loadMore` its own controller (or reuse `resultAbortRef.current.signal`), pass `signal`, and bail on `signal.aborted` (and/or a `selectedKey` snapshot) before the append.

### P2 — sidebar file-size input applies on every keystroke (round-trip re-sync)

`packages/dmworkbase/src/Components/GlobalSearch/GlobalSearchFilterPanel.tsx` — `updateFileSize` (L470–482) + the `[filters]` effect (L142–152).

In `sidebar` mode each keystroke calls `updateDraft` → `onApply(next)` synchronously, bubbling to the parent, which feeds `filters` back down; the effect then re-derives `fileSizeMinInput`/`MaxInput` from `filters` via the KB↔bytes round-trip (`kb*1024` then `round(bytes/1024)`).

In practice the round-trip is lossless for positive integers and React skips identical `setState`, so the common typing path is not corrupted — this is why I rate it P2, not P1. But it does fire a search request per keystroke (no debounce, unlike the 300ms debounce used for the query), and edge inputs (leading zeros, transient empty) normalize in a way that can surprise the user mid-type. Recommend debouncing the size inputs or committing on blur, consistent with the keyword debounce. (Independently surfaced by the second review leg.)

### P2 — files-specific empty hints hardcoded on a tab-generic panel

`packages/dmworkbase/src/Components/GlobalSearch/GlobalContentSearchPanel.tsx` (L253–254) unconditionally passes `t("base.globalSearch.files.emptyHint")` / `files.noResults`, though the component still accepts `tab="messages"`. Today `index.tsx` only mounts it with `tab="files"` (messages go through `GlobalChatSearchPanel`), so no live bug — but the coupling is a latent trap. Gate the hints on `tab === "files"` or lift them to the caller.

### P2 — dead i18n keys and unused computed fields

The overview computes `total` / `isTotalApproximate` (`useGlobalChatSearch.ts` L231–232) and `isMatchCountApproximate` (L223) but nothing renders them; and several new locale keys are unused — `aggregated.approxConversations`, `aggregated.conversations`, `aggregated.approxMessages`, `aggregated.approxMessagesShort`, `aggregated.messagesShort` (en-US.json / zh-CN.json). Only `aggregated.messages` is consumed. Either wire up the approximate-count UI these were clearly meant for, or drop the keys/fields to avoid rot.

**Quality: Changes-Requested** (one P1: `loadMore` cross-conversation contamination.)

---

## 3. Overall verdict

**CHANGES_REQUESTED** — spec gate passes, but the code-quality gate has a P1 correctness race in `loadMore`. Resolving that (apply the same abort/staleness guard the sibling effects already use) should clear the gate; the P2s are non-blocking cleanups.

## 4. Suggestions
- `loadMore`: reuse `resultAbortRef` or add a dedicated controller + `aborted`/selected-key snapshot check after the `await`.
- Debounce or blur-commit the sidebar file-size inputs (and consider the same for other sidebar-mode toggles that now call `onApply` eagerly).
- Prune or wire up the unused approximate-count keys and `total`/`isTotalApproximate` fields.

## 5. Additional notes
- The `syncMessages` signature change adds a 4th param defaulted to `false`; the two direct callers (`Conversation/index.tsx:2559`, `vm.ts:1713`) remain correct, and provider-level `syncMessages` calls are a different method — no cross-file breakage. ✅
- `handleLocate` keeps the camelCase/snake-case separation and guards missing `channelId`/`channelType` — good. ✅
- Test coverage for `model.ts` (drill-down/canRun) and `GlobalMessageSearchService` is solid; the P1 race is exactly the kind of async path unit tests won't catch, hence the emphasis.

_Note: this review reflects one automated cross-family reasoning pass plus manual verification; a second line-level adversarial pass was unavailable this run (upstream gateway 503 / tool-injection error), so consider the code-quality coverage single-legged rather than dual-confirmed._

Jerry-Xin (approved):
APPROVE @ `2e92224e` — the `loadMore` stale-conversation result-corruption blocker from my prior CR is genuinely fixed. (The delta is a rebase — ahead:3/behind:1 — so the file list is inflated with merged-in main changes; I isolated the real change by blob-diffing `useGlobalChatSearch.ts`.)

## ✅ Blocker resolved — `loadMore` now aborts + drops stale responses
`packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts` `loadMore`:
- Grabs `const controller = resultAbortRef.current`, bails if `!controller || controller.signal.aborted`, and passes `signal: controller.signal` into `searchMessages`.
- **The critical guard**: after the await, `if (controller.signal.aborted || resultAbortRef.current !== controller) return` before the `setResult(... [...current.items, ...response.items] ...)` append. When a conversation switch (or new search) happens mid-pagination, a fresh controller replaces `resultAbortRef.current`, so the in-flight page's response is **dropped instead of appended onto the new conversation's list** — closing the wrong-conversation result-mixing corruption. The `.catch` has the same `resultAbortRef.current === controller` guard.

This is exactly the fix the finding called for (abort signal + stale-selection guard mirroring the result effect).

## ✅ Security surface unchanged (byte-verified, no rebase regression)
`GlobalMessageSearchService.ts`, `APIClient.ts`, `globalChatSearch/model.ts`, and `GlobalSearch/sanitize.ts` are **byte-identical** to the head I previously reviewed. So the axes I confirmed clean still hold: authorization scope is narrowing-only (no admin/global/all-spaces param; backend-enforced), space isolation via the `X-Space-Id` request interceptor, highlight rendering routed through `sanitizeHighlight()` (escapes then restores only literal `<mark>`), query is a structured param (no injection/ReDoS).

## 💬 Non-blocking
- 🟡 File-size filter input applies on every keystroke — debounce or apply-on-blur (yujiawei's + mochashanyao's note).
- 🟡 IntersectionObserver create/destroy churn — the observer is correctly `disconnect()`ed and gated on `hasMore`, so it's a valid perf optimization, not a defect.

Blocker fixed, security intact. Back to APPROVE. Credit @OctoBoooot + @yujiawei + @mochashanyao for the loadMore catch.

yujiawei (approved):
# Code Review — PR #835 (octo-web) — re-review

**Head SHA reviewed:** `2e92224e6b806260163ce2081a912ff526a8f82f` (was `e5355ba` in the prior pass)
**Base / merge-base:** `main` → `a61285af` (PR #833 now merged to main; its `dmloop` changes are in the merge-base, **not** part of this PR's diff)
**Diff at this head:** 31 files, +2656 / −565 — aggregated global chat search.

This supersedes my earlier `CHANGES_REQUESTED`. Re-reviewed against the current head after the pagination fix commit `2e92224e`.

## Blocking item from prior review — resolved

**P1 (was blocking): `loadMore` drill-down pagination could contaminate the wrong conversation** — **FIXED.**
`packages/dmworkbase/src/bridge/globalChatSearch/useGlobalChatSearch.ts` `loadMore` now:
- snapshots `const controller = resultAbortRef.current` and bails if already aborted,
- passes `signal: controller.signal` into `searchMessages`,
- re-checks `controller.signal.aborted || resultAbortRef.current !== controller` after the `await` before the append, and
- applies the same controller-identity guard in the `catch`/`finally` (so `loadingMoreRef` isn't reset by a stale request).

This is exactly the abort/staleness guard the sibling effects already used. A regression test (`__tests__/useGlobalChatSearch.test.tsx`, "does not append a stale page after switching conversations") reproduces the precise scenario I flagged — page-2 of conversation A resolving after switching to B — and asserts the stale page is not appended and A's request is aborted. Verified the guard and the test at this head.

*Why this was correctly P1, not P2:* `loadMore` did a functional **append** (`items: [...current.items, ...response.items]`) and overwrote `nextCursor`, so a late page-A response mutated conversation B's state (wrong messages shown + poisoned cursor for the next page) rather than being harmlessly replaced. It was reachable in production, not self-correcting. The fix confirms the assessment.

## Remaining items — all P2, non-blocking

- **Sidebar file-size input applies per keystroke** (`GlobalSearchFilterPanel.tsx` `updateFileSize`): still fires `onApply` on every keystroke with no debounce (unlike the 300ms keyword debounce). Common typing path is not corrupted (KB↔bytes round-trip is lossless for positive ints, React skips identical state), but recommend debounce or blur-commit. Follow-up.
- **IntersectionObserver churn** (`GlobalChatSearchPanel.tsx` L42): observer effect depends on `search.loadMore`, whose `useCallback` deps include `result.nextCursor`/`hasMore`, so it disconnects/reconnects each page. Correctness-safe performance nit; a ref-stabilized callback would avoid the churn. Follow-up.
- **Files-specific empty hints on a tab-generic panel** (`GlobalContentSearchPanel.tsx` L253-254): hints hardcoded to `files.*` while the component still accepts `tab="messages"`. No live bug (only mounted with `tab="files"`), latent trap. Follow-up.
- **Dead i18n keys / unused fields**: `aggregated.approxConversations`, `aggregated.conversations`, `aggregated.approxMessages`, `aggregated.approxMessagesShort`, `aggregated.messagesShort` still unused in code; `total`/`isTotalApproximate`/`isMatchCountApproximate` computed but unrendered. Prune or wire up the approximate-count UI. Follow-up.

## Verdict

**APPROVED.** Spec gate passes; the sole code-quality blocker (P1 pagination race) is fixed and covered by a regression test at head `2e92224e`. The four remaining P2s are non-blocking cleanups suitable for a follow-up. No new issues introduced by the fix commit (the `GlobalChatSearchLayout` `title` field removal is a clean simplification — `countLabel` is the only header source now).

OctoBoooot (approved):
## Re-review (DELTA) #835 @ `2e92224e` — the loadMore stale-conversation-append major is byte-verified fixed (abort signal + post-await stale guard); CI green → APPROVE

**Verdict:** Approve
Flipping my REQUEST_CHANGES → APPROVE. My major was that `loadMore` issued its paged request with no abort signal and unconditionally appended the response, so a conversation switch mid-load corrupted the result list. Byte-verified at this head: `loadMore` now acquires the result controller, passes its `signal`, and guards on `signal.aborted || resultAbortRef.current !== controller` before appending — exactly the fix. Code CI green.

## What I verified (the fix)
`useGlobalChatSearch.ts` `loadMore` (L362+) at `2e92224e`:
- **Abort signal acquired + early bail** — `const controller = resultAbortRef.current; if (!controller || controller.signal.aborted) return;`
- **Signal threaded into the request** — `dataSource.searchMessages({ ..., signal: controller.signal })` (the missing signal I flagged).
- **Post-await stale guard before append** — `if (controller.signal.aborted || resultAbortRef.current !== controller) return;` guards the `[...current.items, ...response.items]` append. On a conversation switch, the selection effect installs a new controller into `resultAbortRef.current`, so the in-flight old-conversation response fails this check and is **dropped, not appended** to the new conversation's list. That closes the exact corruption path.
- **`.catch` and `finally` symmetric** — the catch only sets error state when still the current controller and not a cancellation; `finally` clears `loadingMoreRef` only if still current. No stuck loading flag, no cross-conversation error bleed.

Isolated the true delta from rebase noise (the PR was rebased, inflating the file list with merged-in #833/#837 files): the hook is genuinely `added`/changed and carries the fix — that's the file my finding was about.

## Also resolved / status
- The **dead `title` prop minor** I raised is incidentally cleaned up too — the layout now renders `result.countLabel` directly (index.tsx:156-159), no longer the `countLabel || title` dead-prop pattern.
- The **file-size-input apply-on-keystroke minor** is a non-blocking perf note (also raised by peers as a 🟡); not addressed here, still non-blocking.

## Net
The loadMore stale-conversation-append major is fully resolved: the paged request is cancellable and its response is dropped when the selected conversation has changed, so pagination can no longer mix messages across conversations. CI green (`size/XL`, no governance block beyond the large-change posture). Flipping my **REQUEST_CHANGES → APPROVE** @ `2e92224e`. The one remaining item (file-size input debounce) is a non-blocking minor. `size/XL` → a human eyeball on the broader feature is still worthwhile.

PR #818, opened 2026-07-16
fix(loop): re-resolve workspace when octo space changes

caster-Q:
## Summary

Both the Personal (我的/运行时) and Loop (回路) pages failed to react to octo space switches: they only re-listed workspaces on mount and on `wk:nav-menu-activated`, not on `space-changed`. After switching space they kept the previous space's workspace context and issued workspace-scoped requests against the new space, so the backend space-isolation gate rejected them (`workspace does not belong to this space` / `not a member of this space`) instead of the page showing the correct empty state. This PR makes both pages re-resolve on `space-changed`.

## Related Issue

Fixes #817

## Changes

- `packages/dmpersonal/src/PersonalPage.tsx`: subscribe to `space-changed`; on switch, reset all three workspace states before re-resolving — `selectedWsRef`, `machineModeRef`, and the `@octo/loop` http-layer slug via `setWorkspaceContext("", "")` — then reuse `resolveAndPaint` (empty space falls into machine mode).
- `packages/dmloop/src/pages/LoopPage.tsx`: subscribe to `space-changed`; clear the http-layer workspace slug first, then re-list workspaces (empty space → empty-state guide). Mirror the current tab via `tabRef` so the mount-once handler paints the live tab; share one generation guard (`spaceResolveSeqRef`) across the mount and space-changed resolves to drop out-of-order responses; gate click entries (`openTab` / `switchWorkspace` / `openNewLoop`) with `!loaded` during the re-resolve window.
- `apps/web/src/__tests__/personalPageSpaceChanged.test.ts`: new tests asserting the subscription, the state resets and their ordering, the shared seq guard, the tabRef render, and the window-period entry guards.

## Architecture / Module Boundary

- Affected module(s): `packages/dmloop`, `packages/dmpersonal`.
- New or changed user-visible entry point: no (fixes existing space-switch behavior on existing pages).
- Shared layer touched: none (uses existing `@octo/loop` workspace helpers and `WKApp.mittBus`).
- If shared code changed, impact scope: n/a.
- Duplicate entry point checked: yes.

## Testing

- `cd apps/web && npx vitest run src/__tests__/personalPageSpaceChanged.test.ts` — 11 passed.
- Manually verified against a local self-hosted stack: switching between a space with a workspace and a space without one now shows the correct content / empty state, with no space-isolation error toast.

- [x] Unit tests added/updated
- [x] Manually verified

## Checklist

- [x] I have read CONTRIBUTING.md
- [x] PR description is in English
- [x] Added tests for my changes
[...]

PR #837, opened 2026-07-16
refactor(route): clean up URLs and session scoping

menghao-webtest:
## Summary
- move sid handling out of visible route URLs into a tab-scoped session helper
- normalize host route paths and keep main menu state in sync with browser history
- keep standalone/external routes such as space management out of the host app router
- update docs, invite, login bind, and loop route comments/tests for the clean URL behavior

## Risk Areas
- host route navigation, browser Back/Forward, and direct route refreshes
- login/logout and sid-scoped storage recovery
- standalone docs and invite landing redirects
- external same-origin routes such as /space

## Verification
- pnpm --dir apps/web exec vitest run src/__tests__/externalStandaloneRoutes.test.ts src/__tests__/mainMenuReconcile.test.tsx src/__tests__/inviteLandingRedirectPath.test.ts src/__tests__/layoutStandaloneDocPath.test.ts src/__tests__/docsDeepLinkCapture.test.ts
- pnpm --dir packages/dmworkbase exec vitest run src/Service/__tests__/RoutePath.test.ts src/Service/__tests__/SessionScope.test.ts src/__tests__/App.logoutRoute.test.ts
- pnpm --dir packages/dmworklogin exec vitest run src/oidc/__tests__/url.test.ts src/oidc/bind/__tests__/url.test.ts src/__tests__/login_vm_oidc.test.ts
- pnpm --dir packages/docs exec vitest run src/module.test.tsx src/pages/DocsHome.test.tsx
- pnpm --dir apps/web build
- git diff --check
- manual smoke: direct /appbot opens the app page without sid in the URL; returning to / keeps sid out of the URL

### DIFF
--- a/packages/dmworkbase/src/Service/SessionScope.ts
+++ b/packages/dmworkbase/src/Service/SessionScope.ts
@@ -1,0 +1,122 @@
+const SESSION_SID_KEY = "octo.session.sid";
+
+let memorySid = "";
+
+function randomSid(): string {
+  return Math.random().toString(36).slice(-6);
+}
+
+function safeSessionStorage(): Storage | undefined {
+  if (typeof window === "undefined") return undefined;
+  try {
+    return window.sessionStorage;
+  } catch {
+    return undefined;
+  }
+}
+
+function safeLocalStorage(): Storage | undefined {
+  if (typeof window === "undefined") return undefined;
+  try {
+    return window.localStorage;
+  } catch {
+    return undefined;
+  }
+}
+
+export function findReusableStoredSessionSid(store: Pick<Storage, "length" | "key" | "getItem"> | undefined): string {
+  if (!store) return "";
+
+  const sessions: { sid: string; uid: string }[] = [];
+  for (let i = 0; i < store.length; i++) {
+    const key = store.key(i);
+    if (!key || !key.startsWith("token") || key === "token" || key === "tokenCallback") {
+      continue;
+    }
+    const token = store.getItem(key);
+    if (!token) continue;
+
+    const sid = key.slice("token".length);
+    if (!sid) continue;
+    sessions.push({
+      sid,
+      uid: store.getItem(`uid${sid}`) || "",
+    });
+  }
+
+  if (sessions.length === 1) return sessions[0].sid;
+
+  const uid = sessions[0]?.uid;
+  if (uid && sessions.every((session) => session.uid === uid)) {
+    return sessions[0].sid;
+  }
+
+  return "";
+}
+
+export function getSidFromSearch(search: string): string {
+  const params = new URLSearchParams(search || "");
+  return params.get("sid") || "";
+}
+
+export function removeSidFromSearch(search: string): string {
+  const params = new URLSearchParams(search || "");
+  params.delete("sid");
+  const next = params.toString();
+  return next ? `?${next}` : "";
+}
+
+export function removeSidFromPath(path: string): string {
+  try {
+    const url = new URL(
+      path,
+      typeof window === "undefined" ? "https://octo.local" : window.location.origin
+    );
+    url.searchParams.delete("sid");
+    return url.pathname + url.search + url.hash;
+  } catch {
+    return path;
+  }
+}
+
+export function setSessionSid(sid: string | null | undefined): string {
+  const nextSid = sid || "";
+  memorySid = nextSid;
+  const store = safeSessionStorage();
+  if (store) {
+    if (nextSid) {
+      store.setItem(SESSION_SID_KEY, nextSid);
+    } else {
+      store.removeItem(SESSION_SID_KEY);
+    }
+  }
+  return nextSid;
+}
+
+export function ensureSessionSid(): string {
+  const urlSid = typeof window === "undefined" ? "" : getSidFromSearch(window.location.search);
+  if (urlSid) return setSessionSid(urlSid);
+
+  const store = safeSessionStorage();
+  const storedSid = store?.getItem(SESSION_SID_KEY) || memorySid;
+  if (storedSid) return setSessionSid(storedSid);
+
+  const reusableSid = findReusableStoredSessionSid(safeLocalStorage());
+  if (reusableSid) return setSessionSid(reusableSid);
+
+  return setSessionSid(randomSid());
+}
+
+export function getSessionSid(): string {
+  return ensureSessionSid();
+}
+
+export function stripSessionSidFromUrl(): boolean {
+  if (typeof window === "undefined") return false;
+  const search = window.location.search || "";
+  if (!getSidFromSearch(search)) return false;
+
+  const nextUrl = window.location.pathname + removeSidFromSearch(search) + window.location.hash;
+  window.history.replaceState(window.history.state, document.title, nextUrl);
+  return true;
+}
--- a/apps/web/src/Components/InviteLanding/index.tsx
+++ b/apps/web/src/Components/InviteLanding/index.tsx
@@ -1,5 +1,5 @@
 import React, { Component } from "react";
-import { I18nContext, WKApp, apiFetchJson, computeAndSaveJoinSuccess, t, toJoinApprovalStatus } from "@octo/base";
+import { I18nContext, WKApp, apiFetchJson, computeAndSaveJoinSuccess, setSessionSid, t, toJoinApprovalStatus } from "@octo/base";
 import type { JoinSpaceStatus } from "@octo/base";
 import { Button, Spin, Toast } from "@douyinfe/semi-ui";
 import "./index.css";
@@ -275,11 +275,12 @@ export default class InviteLanding extends Component<InviteLandingProps, InviteL
             if (!crossSpace && joinedSpaceId) {
                 localStorage.setItem('currentSpaceId', joinedSpaceId);
             }
-            // 跳转回主界面，带上正确的 sid
+            // 跳转回主界面；sid 写入当前 tab 的 SessionScope，不再暴露到 URL。
             const sid = this.findSid();
             // 使用安全的 basePath，避免当 pathname 为 /api/ 时跳到后端 API 路径（#1006）
             const basePath = this.getAppBasePath();
-            window.location.href = `${window.location.origin}${basePath}/${sid ? `?sid=${sid}` : ''}`;
+            if (sid) setSessionSid(sid);
+            window.location.href = `${window.location.origin}${basePath}/`;
         } catch (e: any) {
             const body = this.getApiErrorData(e);
             if (this.isNeedSpaceResponse(this.getApiErrorStatus(e), body)) {
--- a/apps/web/src/Layout/index.tsx
+++ b/apps/web/src/Layout/index.tsx
@@ -7,13 +7,13 @@ import { Notification as NotificationUI, Button } from '@douyinfe/semi-ui';
 import { checkUpdate, installUpdate, UpdateManifest } from '@tauri-apps/api/updater'
 import { relaunch } from '@tauri-apps/api/process'
 import { os } from "@tauri-apps/api";
-import { getSid, getQueryParam, computeAndSaveJoinSuccess } from "@octo/base";
+import { getQueryParam, computeAndSaveJoinSuccess, removeSidFromPath, setSessionSid } from "@octo/base";
 import type { JoinApprovalStatus } from "@octo/base";
 import { toJoinApprovalStatus } from "@octo/base";
 import InviteLanding from "../Components/InviteLanding";
 import JoinSpacePage from "../Components/JoinSpacePage";
 import JoinApprovalResult from "../Components/JoinApprovalResult";
-import { StandaloneDocPage, parseStandaloneDocId, isStandaloneDocPath, persistStandaloneReturn, consumeStandaloneReturn, withReturnSid } from "@octo/docs";
+import { StandaloneDocPage, parseStandaloneDocId, isStandaloneDocPath, persistStandaloneReturn, consumeStandaloneReturn } from "@octo/docs";
 import { SummaryDetailPage } from "@dmwork/summary";
 import { adoptStoredSession, findSidForToken, clearSessionsWithToken } from "./recoverSession";
 import { isLoopCliAuthorizePath, LOOP_CLI_AUTHORIZE_PATH } from "@octo/loop";
@@ -137,8 +137,6 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             const basePath = rawPath
                 .replace(/^\/api(?:\/v\d+)?(?=\/|$)/, '')
                 .replace(/\/+$/, '')
-            // 保留原始 sid（如果有），不随机生成新的
-            const existingSid = getQueryParam("sid") || ""
             // #511 problem 2 (附带必修): a forwarded-doc link is `/docs?...&doc=<id>&sid=<space>`.
             // A first-time recipient logs in, then the post-login redirect used to keep only ?sid=
             // and drop ?doc=, landing them on the empty document list instead of the document they
@@ -149,7 +147,6 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             const forwardFolder = getQueryParam("folder") || ""
             const forwardSp = getQueryParam("sp") || ""
             const redirectQuery = new URLSearchParams()
-            if (existingSid) redirectQuery.set("sid", existingSid)
             if (forwardDoc) {
                 redirectQuery.set("doc", forwardDoc)
                 if (forwardSpace) redirectQuery.set("space", forwardSpace)
@@ -161,7 +158,7 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
                 if (forwardSpace) redirectQuery.set("space", forwardSpace)
             }
             const redirectQs = redirectQuery.toString()
-            const sidParam = redirectQs ? `?${redirectQs}` : ""
+            const redirectSearch = redirectQs ? `?${redirectQs}` : ""
 
             const goMain = () => {
                 // A user who signed in from a shared /d/:docId link (local OR SSO/OIDC, where the
@@ -170,27 +167,22 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
                 // clears the key and only returns SAFE same-origin relative paths (open-redirect
                 // guard), so a tampered value can never redirect off-origin.
                 //
-                // Carry the just-authenticated session's own sid on the reload (XIN-398): the
-                // stashed target has no `?sid=`, so the reloaded /d/:docId's sid-keyed load() would
-                // read only the empty-sid bucket and — for a multi-session user — fall through to a
-                // recovery that (XIN-392 P1-2) refuses to guess an identity, bouncing them back to
-                // login in a loop. withReturnSid appends the sid of the bucket that already holds the
-                // current token (findSidForToken — the known identity, never a guess), so the reload
-                // hits the right session directly instead of relying on the now-strict recovery. It
-                // no-ops when the target already carries a sid or the session lives in the empty-sid
-                // bucket (where a no-sid reload already resolves it), and the appended target still
-                // passes the isSafeReturnPath / parseStandaloneDocId gates.
+                // Carry the just-authenticated session's own sid on the reload (XIN-398), but keep
+                // the URL clean: store the known bucket sid in SessionScope, then navigate to the
+                // sid-less return path. The reloaded page still hits the right sid-keyed session
+                // bucket, without exposing `?sid=` in the address bar.
                 const standaloneReturn = consumeStandaloneReturn();
                 if (standaloneReturn) {
                     const sessionSid = findSidForToken(localStorage, WKApp.loginInfo.token || "");
-                    window.location.assign(withReturnSid(standaloneReturn, sessionSid))
+                    if (sessionSid) setSessionSid(sessionSid);
+                    window.location.assign(removeSidFromPath(standaloneReturn))
                     return
                 }
                 if ((window as any).__POWERED_EXTENSION__) {
                     window.location.reload()
                     return
                 }
-                window.location.href = `${window.location.origin}${basePath}/${sidParam}`
+                window.location.href = `${window.location.origin}${basePath}/${redirectSearch}`
             }
 
             // 检查是否有待处理的邀请码（验证格式防止 XSS/Open Redirect）
@@ -430,7 +422,8 @@ export default class AppLayout extends Component<{}, AppLayoutState> {
             }
             // Anonymous: stash the exact /d/:docId target so the post-login flow (local OR SSO/OIDC)
             // can bounce the user back to the document instead of the app root, then fall through to
-            // the login screen (below) without navigating away. The standalone page itself only
+            // the login screen (below) without navigating away. SessionScope keeps the session bucket
+            // available without exposing sid in the return URL. The standalone page itself only
             // mounts once a token exists, so the anonymous path is the ONLY place this key gets
             // written for a first-time visitor — onLogin consumes it via consumeStandaloneReturn.
             persistStandaloneReturn();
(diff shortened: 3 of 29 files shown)


Jerry-Xin (approved):
**APPROVE** — Route/session-scoping refactor is correct and security-positive on every axis I checked. `sid` moves out of visible URLs into a per-tab session scope; cross-account isolation, logout teardown, route guards, and redirect trust are all preserved or improved. Byte-verified at head `73724bf`.

### ✅ Security axes (byte-verified)

**(a) Session-scope isolation — the core risk.** "Session scope" here = a per-tab `sid` that namespaces auth storage keys (`token<sid>`, `uid<sid>`, `name<sid>`, …), enabling multiple accounts in one browser. `ensureSessionSid()` resolution order is safe: URL `?sid=` → sessionStorage/memory → **reusable** localStorage bucket → fresh random sid. The reuse path `findReusableStoredSessionSid()` correctly refuses ambiguous cross-identity reuse — it returns a stored sid only when there is exactly one bucket, or when all buckets share the same `uid`; two buckets with different uids → `""` → a fresh random sid. So a scope key from account A cannot silently be adopted under account B (`SessionScope.ts:33-56`). No key collision (sid is the namespace; distinct sids → distinct key sets), no over-broad global fallback.

**(b) Logout / account-switch teardown.** Teardown is thorough and correctly ordered in `clearLocalLoginState()` (`App.tsx`): `LoginInfo.logout()` removes the current sid's scoped keys *before* the sid is cleared, then `clearAuthStorage()` prefix-sweeps ALL `token*/uid*/…` keys from both sessionStorage and localStorage (`oidcLogout.ts:109-140`), then `setSessionSid("")` clears the active scope, and `logout()` now `window.location.replace("/login")` instead of reloading the pre-logout business route. No stale scope survives logout → next user cannot inherit a prior scope/token.

**(c) Route auth-gating preserved.** RouteManager change is dispatch-only: `register/get/push/renderCurrentPath` now run paths through `normalizeRoutePath()` (strip query/hash, ensure leading slash, drop trailing slash) — no guard logic lives here, and no previously-gated route is newly registered as public. Space management stays a standalone external route via `window.location.assign("/space")` (not folded into the host app router). `popstate`/`pageshow` now `renderCurrentPath` (render without pushing a duplicate history entry) instead of `push`.

**(d) Redirect-target trust (open redirect).** InviteLanding post-join redirect is same-origin by construction: `` `${window.location.origin}${basePath}/` `` where `basePath` is derived from `window.location.pathname` (with the `/api[/vN]` prefix stripped), never from attacker-controllable external input — so no open-redirect. Bind flow changes are comment-only; the synchronous URL-scrub of bind `token/authcode/return_to` params (to keep them out of the Back stack / referrer) is retained.

**(e) No scope token in URL.** This is the headline improvement: `push()` no longer appends `?sid=<sid>` to route URLs, InviteLanding drops `?sid=` from its redirect (writes it via `setSessionSid` instead), and `stripSessionSidFromUrl()`/`removeSidFromSearch`/`removeSidFromPath` actively strip any incoming `sid` from the address bar — so the session token is no longer exposed in shareable URLs, history, or referrers.

**(f) Test coverage.** `SessionScope.test.ts` covers the isolation core behaviorally: URL-sid absorption + persistence after URL cleanup, per-tab sid creation, single-bucket reuse, and the same-identity-vs-ambiguous-identity refusal. `RoutePath.test.ts` covers normalization (root, trailing slashes, leading slash, query/hash drop). `inviteLandingRedirectPath.test.ts` covers same-origin basePath derivation incl. the `/api` prefix repro. Note: `App.logoutRoute.test.ts` and `externalStandaloneRoutes.test.ts` are *source-text* assertions (regex over file contents) rather than behavioral — they lock in the `setSessionSid("")` / `replace("/login")` / `assign("/space")` invariants but won't catch a behavioral regression in the teardown wiring. Non-blocking, but a behavioral logout-teardown test (assert all scoped buckets gone after `clearLocalLoginState`) would be stronger.

### 💬 Non-blocking notes
- `findReusableStoredSessionSid` reuses `sessions[0]` when all uids match; harmless (same identity) but relies on insertion order — fine as-is.
- `randomSid()` uses `Math.random()` (6 chars). Acceptable since the sid is only a client-local storage namespace, not a security token, but worth a comment noting it must never become a trust boundary.

CI: Build / secret-scan / osv / dependency-review / pr-title-lint / history all green; `check-sprint` + `satisfy-code-review` skipped (governance); `code-review` queued. No red required check. Non-draft.

packages/dmworkbase/src/Service/SessionScope.ts
```diff
@@ -0,0 +1,122 @@
+const SESSION_SID_KEY = "octo.session.sid";
+
+let memorySid = "";
+
+function randomSid(): string {
+  return Math.random().toString(36).slice(-6);
+}
```
OctoBoooot, line 7: **[minor]** `randomSid()` draws ~6 base-36 chars from `Math.random()`. This is fine as a tab-scope discriminator — it is NOT a security token (the server-validated token under `token{sid}` is the auth boundary; sid only selects which stored bucket a tab reads), so this is not a security finding. But the small space means a same-browser collision would attach a new tab to an existing bucket. Extremely unlikely; `crypto.randomUUID()` (or a longer draw) removes the concern for near-zero cost.

packages/dmworkbase/src/Service/SessionScope.ts
```diff
@@ -0,0 +1,122 @@
...
+    const sid = key.slice("token".length);
+    if (!sid) continue;
+    sessions.push({
+      sid,
+      uid: store.getItem(`uid${sid}`) || "",
+    });
+  }
+
+  if (sessions.length === 1) return sessions[0].sid;
+
```
OctoBoooot, line 48: **[praise]** The reuse gate is exactly the right shape: auto-adopt a stored session only when unambiguous — a single `token{sid}` bucket, or all buckets sharing one `uid` — and return `""` (→ fresh sid) on ambiguous different identities. That's what stops a clean-link tab from cross-attaching to another account's bucket, and it's pinned by a test. This is the #818-class scope-leak concern handled correctly.

OctoBoooot (approved):
## Review: refactor(route): clean up URLs and session scoping (#837)

**Verdict:** Approve
A coherent, well-tested refactor that moves `sid` out of visible route URLs into a tab-scoped `SessionScope` helper and normalizes host route paths. I checked the primary risks — session-scope isolation (cross-account bucket reuse), logout teardown, redirect safety, and auth-gate parity — and all hold. Code CI is green.

Note on CI: the `statusCheckRollup` shows a `pr-title-lint` FAILURE, but that's a **stale earlier leg** (title was originally "Refactor route URLs and session scoping"); the author corrected it to `refactor(route): clean up URLs and session scoping` and the newest leg (by started_at) is `success`. Build / secret-scan / osv-scan / dependency-review all pass; check-sprint + satisfy-code-review are skipped (governance). No code-CI block.

## Parity checked
This is a "refactor" that relocates sid handling and normalizes paths; I diffed base→head to confirm no capability is dropped:
- **Route resolution preserved** — `normalizeRoutePath` maps `/x?q`, `/x#h`, `/x/`, and bare `x` all to `/x` **before** the endpoint lookup (tested in RoutePath.test.ts), and every `route.push`/`getMethod`/`invoke` path normalizes consistently. So every route that resolved before still resolves; any auth gating attached to those endpoints is unchanged (the gate is on the endpoint, and the endpoint key is identical post-normalization).
- **sid behavior preserved, just relocated** — previously sid rode in `?sid=` and keyed `token{sid}` storage buckets; now `ensureSessionSid()` sources it from URL → sessionStorage → a reusable localStorage bucket → fresh random, and `stripSessionSidFromUrl` cleans the address bar. `App.getSID()` still returns it and `*StorageItemForSID` still keys buckets by it — identical storage semantics, only the URL exposure is removed.
- **Deep-link params that matter are captured before normalization** — doc id / space / bind token are snapshotted (docsDeepLinkCapture, layoutStandaloneDocPath, bindModule) before the router normalizes to a clean pathname, so the "clean URL" change doesn't drop them.
- **Standalone/external routes stay external** — `/space` uses `window.location.assign("/space")`, not `WKApp.route.push` (tested), so it's not absorbed into the host router.

## Security (primary risk — checked)
- **Cross-account session isolation holds.** `findReusableStoredSessionSid` reuses a stored bucket only when unambiguous: exactly one `token{sid}` bucket, OR all buckets share the same `uid`. The ambiguous different-identity case returns `""`, falling through to a fresh random sid — so a new clean-link tab can't silently attach to another account's bucket. Directly tested ("reuses same-identity buckets but refuses ambiguous different identities"). This is the #818-class scope-leak concern, and it's gated correctly.
- **Logout teardown is complete** — `clearLocalLoginState` runs `logout()` + `clearAuthStorage()` + `setSessionSid("")`, and logout navigates to `/login` without reloading the pre-logout business route (tested in App.logoutRoute.test.ts). No stale scope survives.
- **No open-redirect introduced** — InviteLanding now redirects to a fixed `${window.location.origin}${basePath}/` (app-controlled constant) and writes sid into SessionScope instead of the URL; the bind flow snapshots its entry params before the router can normalize them away. No user-controlled redirect target.

### Minor
- `SessionScope.ts:6-8` — `randomSid()` uses `Math.random().toString(36).slice(-6)` (~6 base-36 chars). Fine as a tab-scope discriminator (it's not a security token — the server-validated token is the auth boundary; sid only selects which stored bucket a tab reads). But the small space means a same-browser collision would attach a new tab to an existing bucket. Extremely unlikely and non-security, but `crypto.randomUUID()` (or a longer draw) would remove the concern for near-zero cost.

### Praise
- The isolation gate in `findReusableStoredSessionSid` is exactly the right shape: it only auto-adopts a stored session when there's no ambiguity about *which identity* it belongs to, and refuses (fresh sid) otherwise — that's the conservative choice that prevents a clean-link tab from cross-attaching to another account, and it's pinned by a test. Keeping `/space` out of the host router via `location.assign` (with a test asserting it's not `route.push`) is the correct way to preserve the standalone boundary through a routing refactor.

### Net
A large but coherent routing/session refactor: route resolution and sid storage semantics are preserved (only the URL exposure is removed), cross-account bucket reuse is gated and tested, logout tears down the scope, and no open-redirect or auth-gate regression was introduced. Strong per-axis test coverage (RoutePath, SessionScope, logoutRoute, deep-link capture, external routes). Code CI green (newest title-lint leg passes; `size/XL` → `needs-human-review` warranted for a change this broad). **Approve**, with one non-blocking minor (harden `randomSid`).

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #837 (octo-web)

## Summary

This PR refactors session identity management and route handling in the web frontend. Two new modules — `SessionScope.ts` and `RoutePath.ts` — replace the old pattern of carrying `?sid=` in the URL with a sessionStorage-backed per-tab session identifier, and normalize route paths (strip trailing slashes, query strings, and hash fragments) before handler lookup and history writes. The result is cleaner URLs, no more session identifiers leaking into the address bar / browser history / referrer headers, and a proper separation between route rendering and history management (`renderCurrentPath` vs `syncPath` vs `push`). Logout now navigates to `/login` instead of reloading the pre-logout business route. The refactor touches 29 files across 4 packages with new tests covering the core abstractions.

Overall the design is sound and the migration is well-executed. Two P2 items and one nit below.

## Verification

- ✅ **SessionScope data flow** — traced `getSessionSid()` → `ensureSessionSid()` through all four fallback tiers (URL → sessionStorage → localStorage recovery → random). `setSessionSid("")` correctly clears both memory and sessionStorage on logout (`App.tsx:990`). `findReusableStoredSessionSid` safely returns `""` when multiple user identities exist in localStorage, preventing cross-identity session reuse.
- ✅ **Route normalization parity** — `register()`, `get()`, `push()`, `syncPath()`, and `renderCurrentPath()` all normalize through the same `normalizeRoutePath()`, so handler registration and lookup use identical keys.
- ✅ **Browser history correctness** — popstate/pageshow handlers now call `renderCurrentPath()` (render only, no history write) instead of the old `push()` which created duplicate history entries on every back/forward. Verified by `App.logoutRoute.test.ts`.
- ✅ **Session recovery after login redirect** — `Layout/index.tsx:167-180`: `consumeStandaloneReturn()` → `setSessionSid(sessionSid)` → `window.location.assign(removeSidFromPath(...))`. After reload, `ensureSessionSid()` at `index.tsx:56` reads from sessionStorage. Same-tab sessionStorage persistence across `window.location.assign` is guaranteed by the Web Storage API spec.
- ✅ **Logout route** — `window.location.replace("/login")` prevents back-button to authenticated state; `setSessionSid("")` clears the session bucket. Verified by `App.logoutRoute.test.ts`.
- ✅ **Deep-link boot path** — `vm.ts:67` uses `normalizeRoutePath(window.location.pathname || WKApp.route.currentPath)`, correctly preferring the live URL over the possibly-stale `currentPath` during startup. Popstate listener added/removed symmetrically in `didMount`/`didUnmount`.

Static analysis only at head `73724bf0f6dd205336ea90caf7d297f072b76584`; build and tests not executed in this environment.

## Findings

### P2 — Stale sid-in-URL comments in unchanged files (`apps/web/src/Layout/index.tsx:167`)

The Layout correctly migrates from `withReturnSid` to `setSessionSid` + `removeSidFromPath` here. However, several unchanged files still carry comments that describe the old `?sid=`-in-URL contract:

- `packages/dmworkbase/src/Components/MeInfo/vm.tsx:199-215` — warns that SSO `return_to` must preserve `?sid=` in the URL ("否则 IdP 302 回来时 sid 丢失 → App.tsx::getSID 读空 sid bucket"). After this PR, `getSID()` reads sessionStorage, so the URL no longer needs to carry `sid`.
- `apps/web/src/Layout/__tests__/recoverSession.test.ts:86` — asserts `getSID() returns whatever ?sid= the current URL carries`, which is no longer true.
- `packages/docs/src/config.ts:171-174` — describes RouteManager re-stamping URLs to `/docs?sid=…`, which no longer happens.

These stale comments will mislead future maintainers debugging session recovery flows. A follow-up pass to update all `?sid=`-in-URL prose to reference `SessionScope` would close this gap.

### P2 — `push()` silently strips query params and hash (`packages/dmworkbase/src/Service/Route.tsx:20`)

`normalizeRoutePath` at `RoutePath.ts:7-10` removes `?query` and `#hash` fragments before both the route handler lookup and the `pushState` call. All current production callers pass clean `menus.routePath` values (verified: only `tab_low_screen.tsx:26` calls `push()` with `menus.routePath`), so no existing behavior is broken.

However, `push()` is a public method on the `RouteManager` singleton exported via `@octo/base`. Any caller outside this diff — extension code, other packages, or future callers — that previously relied on query params being preserved in the browser URL after `push("/docs?doc=d1")` will silently lose them. Consider either (a) documenting the normalization contract on the method signature, or (b) preserving the caller's original query/hash when constructing the history URL while still normalizing for handler lookup.

### Nit — `syncPath` before `onPress` may create duplicate history entries (`apps/web/src/Pages/Main/index.tsx:254`)

On desktop, `syncPath(menus.routePath)` pushes a history entry unconditionally before `onPress()` runs. If any menu's `onPress` handler triggers its own `push()` or `window.location` navigation, the user would get two history entries (the menu route + the handler's target). On `tab_low_screen.tsx:23`, `syncPath` is only called inside the `onPress` branch, avoiding this. If no `onPress` handler currently navigates via `push()`, this is harmless — but worth confirming or guarding.

## Data Flow Trace

| Consumer | Source | Traced Path | Result |
|---|---|---|---|
| `LoginInfo.getSID()` | `getSessionSid()` → `ensureSessionSid()` | sessionStorage → localStorage recovery → random | ✅ sid always available |
| `getSid()` in `search.ts` | `getSessionSid()` | Same as above | ✅ unified |
| `RouteManager.push(path)` | `normalizeRoutePath(path)` → `EndpointManager.invoke` | Route handler lookup uses stripped path | ✅ consistent with `register()` |
| `vm.didMount()` boot path | `normalizeRoutePath(window.location.pathname)` | Live URL preferred over stale `currentPath` | ✅ |
| Layout standalone return | `setSessionSid(sid)` → `removeSidFromPath(return)` → `location.assign` | sessionStorage persists across same-tab navigation | ✅ |
| Logout | `setSessionSid("")` → `location.replace("/login")` | Clears sessionStorage entry, navigates away | ✅ |

## Blindspot Checklist

- **C1 (Dual-path parity)** — `push()` vs `syncPath()` vs `renderCurrentPath()`: all normalize via `normalizeRoutePath()`. `popstate`/`pageshow` use `renderCurrentPath` (no history write); menu clicks use `syncPath` or `push` (with history write). Clear.
- **C2 (Control-flow ordering)** — `syncPath` before `onPress` on desktop; see Nit above.
- **C3 (Authorization boundary)** — N/A. No auth endpoint changes; session scoping is client-side only.
- **C4 (Auth lifecycle cascade)** — Logout clears sessionSid, loginInfo, and authStorage. New login generates fresh sid. Clear.
- **C5 (Build ≠ runtime)** — N/A. No build/packaging changes.
- **C6 (Governance/docs)** — N/A. No governance doc changes.

## Verdict: APPROVED

No P0/P1 issues found. The session scoping refactor is well-designed with proper fallback tiers, clean logout behavior, and good test coverage for the new abstractions. The two P2 items (stale comments in unchanged files and `push()` API contract change) are non-blocking — they don't break any current callers but should be addressed in a follow-up to prevent future confusion.

[Octo-Q] verdict: APPROVE — No P0/P1 blockers; clean session scoping refactor with proper sessionStorage fallback chain, consistent route normalization, and correct history handling. Two P2 items (stale comments, push() contract change) are non-blocking.

yujiawei (approved):
# Code Review — PR #837 (octo-web)

**Head SHA reviewed:** `73724bf0f6dd205336ea90caf7d297f072b76584`
**Base / merge-base:** `bb933ebc0ba869105fba5139455759aa8475d645` (verified an ancestor of `main` — no sibling-PR drift)

Reviewed the full diff (29 files, +459/-100) plus the surrounding call sites (`RouteManager`, `WKApp`/`LoginInfo`, `MainVM`, `AppLayout`, `StorageService`, `oidcLogout`). Built the workspace and ran the affected suites.

---

## 1. Scope / intent compliance

**Verdict: ✅**

The PR delivers exactly what the title states — move `sid` out of the URL into a per-tab `SessionScope`, and normalize routing — with no unrelated scope creep:

- **Missing:** none. The refactor is complete and internally consistent: URL `?sid=` reads/writes are all routed through `SessionScope` (`getSid` / `LoginInfo.getSID` now delegate to `getSessionSid`), and `ensureSessionSid()` + `stripSessionSidFromUrl()` are wired at boot (`apps/web/src/index.tsx:56-57`) and in the `RouteManager` constructor (`Route.tsx:20`).
- **Over-build:** none. No new flags, endpoints, or public surface beyond what the refactor requires. The new exports (`SessionScope`, `RoutePath`) are the intended API.
- **Deviation:** none observed. Storage keying stays sid-scoped (`token{sid}`/`uid{sid}` via `getStorageItemForSID`), so the on-disk session model is preserved; only the *source* of `sid` changed (URL → sessionStorage/localStorage-bucket/random).

The stale comments referencing the old `pageshow` re-push behavior were correctly updated across `docs`, `dmworklogin`, and `dmloop`.

---

## 2. Code quality

**Verdict: Approved** — no P0/P1. Three P2 advisories below; none blocks merge.

### Verified sound

- **Session bucket reuse heuristic** (`SessionScope.ts:27-55`): `findReusableStoredSessionSid` returns the single stored bucket, or a shared `sid` only when every `token*` bucket resolves to the same `uid`; mixed identities return `""` and fall through to a fresh random `sid` (→ login screen). This preserves the "never guess an identity" invariant. Confirmed the `token`/`tokenCallback` exclusions and the empty-`token` skip are correct.
- **No popstate render loop.** Both `RouteManager.renderCurrentPath` and `MainVM.syncMenuFromBrowserPath` listen on `popstate`. `syncMenuFromBrowserPath` only mutates state + `notifyListener()` (no `history.*`), so it cannot re-trigger `popstate`. No infinite loop, no double history entry. `push`/`syncPath` both guard with a `currentUrl === nextUrl` check before `pushState`, so menu re-selection does not stack duplicate history entries.
- **Logout hardening** (`App.tsx:998-1001`): `window.location.replace("/login")` (was `reload()`) plus `setSessionSid("")` in `clearLocalLoginState` correctly drops the pre-logout business route from the back stack and clears the tab sid. Good security posture.

- **P2 — `logout()` hardcodes the root `/login`, ignoring base path.** `packages/dmworkbase/src/App.tsx:1000` — `window.location.replace("/login")`. For sub-path deployments (the `/web/` case that `InviteLanding.getAppBasePath` and the `#1006` fix explicitly support), this navigates to origin-root `/login` instead of `/web/login`. The old `reload()` preserved the deployment prefix. Root deployments are unaffected; consider deriving the base path here for parity with the invite/redirect flows.
- **P2 — `syncPath`/`push` now normalize away query + hash before `pushState`** (`Service/Route.tsx:45-57`, `68-81`). `normalizeRoutePath` strips `?...`/`#...`, and the URL written to history is built from the stripped `routePath`. This is currently harmless — every live caller (`Main/index.tsx:254,316`, `tab_low_screen.tsx:25`, `switchToMenuById`) passes a clean `menus.routePath` with no query — but it's a latent trap: any future `route.push('/x?y=1')` would silently drop `?y=1` from the address bar. Worth a short comment, or write history from `removeSidFromPath(path)` instead of `routePath`.
- **P2 — `renderCurrentPath` → `restContent` is a no-op in the web shell.** `restContent` sets `WKApp.content`, but nothing in the web app renders `WKApp.content` (`AppLayout.render` mounts `<MainPage/>` / route components directly). So the `popstate`/`pageshow` → `renderCurrentPath` path updates `currentPath` (used) but its `restContent(component)` write is dead in this app. Not a regression (the field pre-exists), just dead weight now that menu sync is handled by `MainVM`.

---

## 3. Testing

Ran the affected suites at head SHA:

- `dmworkbase`: `SessionScope.test.ts`, `RoutePath.test.ts`, `App.logoutRoute.test.ts` — **13/13 pass**.
- `apps/web`: `inviteLandingRedirectPath`, `layoutStandaloneDocPath`, `mainMenuReconcile`, `externalStandaloneRoutes`, `docsDeepLinkCapture` — **39/39 pass**.
- `docs` package: 1420/1424 pass. The 4 failures are in `EditorShell.test.tsx` (i18n key lookups like `docs.toolbar.exportMarkdown` unresolved in the test env) — an untouched file that imports nothing this PR changes. **Pre-existing / environmental, not caused by #837.**

Test coverage for the new behavior is genuinely good: the new files assert the source-level contracts (logout route, `renderCurrentPath` not pushing history, menu URL sync, standalone-route navigation) in addition to unit-testing `SessionScope`/`RoutePath`.

---

## 4. Suggestions

1. Derive the base path in `logout()` rather than hardcoding `/login`, matching the `#1006` invite-redirect handling, so sub-path deployments log out correctly.
2. Add a one-line comment on `syncPath`/`push` noting that query/hash are intentionally dropped from the address bar, or preserve them via `removeSidFromPath(path)` to future-proof programmatic navigation with params.

---

## 5. Additional notes

- **Direct URL `sid` readers:** the only remaining reader of `?sid=` from the live URL (outside `SessionScope`) is `apps/web/e2e/bind.spec.ts` (a Playwright e2e), which is test-only and unaffected. `MeInfo/vm.tsx` preserves the full existing query on the realname-verify `return_to`, so it keeps working whether or not `sid` is present. No production code reads `?sid=` directly anymore.
- **OIDC bind-token scrub** (`bindModule.tsx`) still snapshots `location.search` synchronously at `init()` and scrubs the URL — that defense is preserved; only its explanatory comment changed.

---

**Overall: APPROVE.** Spec-complete, well-tested, architecturally coherent. The three P2 items are hardening/clarity follow-ups, not merge blockers.

*Reviewer note on review coverage: this review combined a line-level pass with a long-context, cross-file analysis leg (which independently flagged the `syncPath`/`push` query-stripping behavior — assessed here as P2 because no current caller passes query strings). A second independent line-level tool leg was unavailable this run due to an environment/gateway error, so it is recorded as absent rather than as agreement — treat the P2s accordingly.*

PR #826, opened 2026-07-16
feat(docs): remove table freeze panes feature

ploy-elison:
## Summary

Removes the document-table **freeze rows/columns** feature (#755). Freeze was a pure view-state feature — the frozen set lived only in a ProseMirror plugin, was never written to the Y.Doc, never synced across collaborators, and did not persist across reload. It has been decided against for the docs editor, so this cleanly takes it out (UI, plugin, CSS, and i18n).

## What changed

- **Extension**: deleted `TableFreeze.ts` (plugin, commands, sticky styler) and its unit test.
- **Toolbar**: removed the ribbon `TableFreezeControl` button/popover.
- **Right-click menu**: removed the "Freeze" submenu (freeze header row / up to this row / first column / up to this column / unfreeze) from `TableContextMenu`; trimmed the now-unused `active` / `disabled` / `caret` props from the local button component.
- **CSS**: removed the `.octo-frozen-*` sticky rules, the `max-height: 60vh` frozen-header scroll box, the submenu/caret/freeze-popover styles, and the `column-resize-handle { z-index: 6 }` override that existed only to sit above frozen cells.
- **i18n**: removed `docs.toolbar.freeze` and `docs.table.freeze*` / `unfreeze` keys from `zh-CN` and `en-US`.

## Scope / safety

- **No schema change** — freeze never touched the content model, so there is no migration and no collab-sync surface affected.
- Other table capabilities are untouched: column width resize, row/column drag-reorder, font color, add/delete rows & columns, and normal cell editing.
- No dead code, no orphaned i18n keys, no console errors.

## Verification

- `pnpm test` (docs) — table/toolbar suites green; the only failures on this branch (`EditorShell.test.tsx`) are pre-existing on `main` and unrelated to this change.
- Typecheck: the two reported errors (`dmworkbase/i18n/format.ts`, `members/sort.ts`) also pre-exist on `main` and are unrelated.
- **Browser check**: mounted the real toolbar + context menu against a table in a real Chrome session — the freeze entry is gone from both the ribbon and the right-click menu, and insert-row / insert-column / cell-edit all still work.

PR #824, opened 2026-07-16
fix(docs): merge recent-row creator/viewed/updated into a single latest-event line

ploy-elison:
## Problem

In the docs home **Recent** list, each row's sub-line placed the document
creator's name directly before "viewed at X". A row like `大背头 · 查看于 刚刚`
read as *"the creator viewed it just now"*, when it actually means *"created by
大背头, viewed by the current user just now"*. The recent list is meant to show
documents the **signed-in user** has viewed, so this was a confusing misread.

## Fix

Recent rows now show two things, and the time collapses to a single
most-recent-event line so two timestamps are never stacked together:

- **Creator line** — who created the document: `创建者 <name>` / `Created by <name>`.
- **Merged time line** — only the *later* of the current user's view time and
  the document's own update time:
  - update newer than the last view → `<last updater> 更新于 <time>` /
    `Updated by <name> · <time>`, naming the updater from the recent API's
    `updatedBy = {uid, name}` field;
  - view newer, equal, or no update recorded → `你查看于 <time>` / `You viewed <time>`.

Equal / very-close timestamps prefer the view. When `updatedBy` is `null`
(no recorded editor) the update line degrades to the unnamed `更新于 <time>`
label rather than borrowing the creator's name. Display-only: `updatedBy`,
`updatedAt`, and `viewedAt` are all carried per row by `GET /docs/recent`; no
data-fetch, filtering, or backend logic is touched here. The "My documents" tab
sub-line ("Updated X") is unchanged.

### Changes
- `packages/docs/src/pages/DocsHome.tsx` — render the recent sub-line as a
  creator span (`octo-docs-list-row-creator`) plus one merged time span
  (`octo-docs-list-row-updated` when the update is newest, otherwise
  `octo-docs-list-row-viewed`).
- `packages/docs/src/pages/docsApi.ts` — add the `updatedBy: { uid, name } | null`
  field to `DocListItem` (recent rows only).
- `packages/docs/src/i18n/{zh-CN,en-US}.json` — add `list.updatedBy`
  (`{{name}} 更新于` / `Updated by {{name}} ·`) in both locales; reuse existing
  `list.createdBy`, `list.viewedBySelf`, `list.updatedAt`.

## Verification
- `@octo/docs` unit tests green (`DocsHome.test.tsx`, `docsApi.test.ts`,
  `i18n.test.ts` — 81 passing), covering view-latest, update-latest-with-name,
  update-latest-with-null-updatedBy fallback, equal-prefers-view, and the
[...]

PR #822, opened 2026-07-16
fix(docs): make table row/column reorder handle stable and discoverable

ploy-elison:
## What & why

The self-built table row/column reorder handle (#76, merged in #780) works, but a series of
real-use defects made the trigger interaction unreliable. This PR fixes them and, in doing so,
**migrates the reorder drag from mouse events to Pointer Events + `setPointerCapture`**. The
event-model change is called out explicitly below — an earlier version of this description said
the reorder core was "untouched", which was inaccurate and is corrected here.

## Changes (by commit)

1. **Stable, discoverable handle (XIN-1233).** Leaving the prose no longer hides the handle
   instantly; a 220ms grace-period hide is cancelled by the handle's own re-entry or the pointer
   returning to a cell, so a quick glide from a cell to the gutter handle stays seamless. The bars
   sit flush against the table edge (the 2px dead gap is gone) and are widened 14px → 18px, with a
   soft shadow / colour transition / larger grip glyph for discoverability.
2. **Coexistence with the sibling row-resize overlay (XIN-1253).** Resting-handle placement moves
   off ProseMirror's `handleDOMEvents` (bound to `view.dom`) onto a document-level `mousemove`, so a
   sibling overlay sitting on top of the cells (e.g. #823's row-height resize bar) can no longer
   steal the pointer and hide the handles. It also reads `view.dom`'s live rect rather than the
   stale detached mount node.
3. **Event-model migration to Pointer Events (XIN-1259 / octo-docs-backend#76).** The drag pipeline
   moves `mousedown/mousemove/mouseup` → `pointerdown/pointermove/pointerup` (+ `pointercancel`) and
   captures the pointer on the grabbed handle with `setPointerCapture`. Capture routes the terminal
   `pointerup`/`pointercancel` to us even on a release **outside the window**, closing the old
   "release outside → drag left silently armed → next stray mouseup commits an abandoned reorder"
   gap (FAIL-1 / XIN-1252). Aligns the reorder handle with the same event model already used by the
   row-height handle in #823.
4. **Multi-pointer re-entrancy fix (this revision — the two P1 blockers reviewers flagged).** The
   Pointer Events migration exposed two multi-pointer gaps that the old mouse-only path did not have:
   - **`beginDrag` had no active-drag guard.** Both the row and column handles carry live
     `pointerdown` listeners, so a second `pointerdown` (a second finger/stylus on the other handle,
     or a stray re-entry) while a drag was active overwrote `drag`/`ordinal`/`capturedPointerId`/
[...]

PR #842, opened 2026-07-16
fix(docs): interpolate import warnings, localize image-migrate reasons, detail import failures

### LINKED ISSUE #841: Docs import: warning placeholders not interpolated; image-migrate reasons untranslated; whole-import failures too generic
## Summary

Three related issues in the docs import path (Markdown / Word):

### 1. Interpolation placeholders leak to the user
Import warnings render with raw `{{reason}}`, `{{count}}`, `{{name}}` placeholders instead of the interpolated values (e.g. `图片未能迁移到新文档（{{reason}}）`, `{{count}} 张本地图片未导入`).

### 2. Image-migrate failure reasons are shown as raw English codes
When an image cannot be migrated, the reason inside the message is a raw backend code (`source_not_found`, `size_too_large`, ...) rather than a human, localized phrase.

### 3. Whole-import failure has a single generic message
A failed Markdown/Word import surfaces only a generic "出错了" with no indication of the cause (wrong file type, corrupt docx, too large, zip-bomb bound, empty upload, ...).

## Root cause (1)
The host `@octo/base` `t()` interpolates values passed under `t(key, { values: {...} })`, but the import modules' local `tr` helpers call `t(key, { reason })` (no `values` wrapper). Two independent `tr` helpers are affected: `editor/importFlow.ts` and `import/markdown.ts`.

## Expected
- Placeholders interpolate correctly.
- Image-migrate reasons show localized text (with a raw-code fallback for unmapped codes).
- Whole-import failures show a specific, localized reason.

wwwzzzcccc:
## Summary

Fixes three docs-import UX issues: leaked `{{reason}}`/`{{count}}`/`{{name}}` placeholders, untranslated image-migrate reason codes, and an over-generic whole-import failure message.

## Linked Spec

Fixes #841

## COMPREHENSION

**1. What does this change actually do (before/after behavior)**
- Placeholders: the local `tr` helpers in `editor/importFlow.ts` and `import/markdown.ts` passed interpolation params as bare props; the host `t()` only substitutes values under `{ values: {...} }`, so `{{reason}}`/`{{count}}`/`{{name}}` rendered verbatim. Both helpers now wrap params under `values`.
- Image-migrate reasons: `source_not_found`, `source_forbidden`, `size_too_large`, `copy_failed`, `store_failed`, `not_an_image`, `mime_not_allowed`, `fetch_failed` now map to localized text, with a raw-code fallback for unmapped codes.
- Whole-import failures: the single generic error is replaced with specific messages for Markdown (wrong type / generic) and Word (not-a-docx / corrupt / timeout / too-large / zip-bomb ratio / too-many-entries / too-many-media / empty upload), keyed off the backend `import_unsafe` reason and HTTP status.

**2. What could break (dependents + failure mode)**
- Dependents: docs import path only (`runMarkdownImport`, `runDocxImport`, `migrateImportedImages`, `parseMarkdownToPmDoc`) plus the `docs.import.*` i18n keys. No schema or API change.
- Failure mode: unmapped reason/error codes fall back to the raw code (image-migrate) or the existing generic message (docx), so a new backend code never yields an empty or broken string. zh-CN/en-US keys stay in sync.

**3. How do you know it works (test/repro/trace evidence)**
- Verified against a real import: `{{reason}}`/`{{count}}` now interpolate; migrate reason shows localized text; docx failures show specific causes.
- Tests: import + editor/importFlow + i18n suites pass (92/92, includes markdown 71, importFlow.images 10, i18n 4). Typecheck clean for the changed files. zh/en import keys confirmed 1:1.


packages/docs/src/editor/importFlow.ts
```diff
@@ -134,11 +134,33 @@ export interface ImportResult {
...
+ * `params` are the interpolation values ({{name}} placeholders); they MUST be
+ * passed under `opts.values` so the host `@octo/base` t() actually substitutes
+ * them — passing them as bare props leaves `{{reason}}` literal in the message.
+ */
+function tr(
+  t: Translate | undefined,
+  key: string,
+  params?: Record<string, string | number>,
+): string {
+  return t ? t(key, params ? { values: params } : undefined) : key
```
OctoBoooot, line 150: **[praise]** Correct fix — wrapping params as `{ values: params }` is exactly what the host `@octo/base` t() needs to substitute `{{reason}}`/`{{count}}`/`{{name}}` (the bare-props form left them literal). Verified the same fix is applied at the sibling `tr` in `import/markdown.ts` too, so both translator seams are covered, not just the one the bug surfaced.

packages/docs/src/editor/importFlow.ts
```diff
@@ -134,11 +134,33 @@ export interface ImportResult {
...
+}
+
+/**
+ * Map a machine-readable backend failure `reason` (from copyAttachments /
+ * ingestAttachments) to a human, localized phrase. Unknown codes fall back to
+ * the raw code so a newly-added backend reason is still visible, not swallowed.
+ */
+function reasonText(t: Translate | undefined, reason: string): string {
```
OctoBoooot, line 158: **[praise]** `reasonText`'s raw-code fallback (`text === key ? reason : text`) keeps an unmapped backend reason visible instead of rendering a meaningless i18n key — and the docx path uses the same `mapped === key ? fallback : mapped` guard. Good robustness against backend codes this frontend doesn't know yet, with no symmetric-fix trap (never a bare key or empty string). All 8 image-migrate codes have non-empty zh-CN/en-US values, in lockstep.

OctoBoooot (commented):
## Review: fix(docs): interpolate import warnings, localize image-migrate reasons, detail import failures (#842)

**Verdict:** Comment (the code is correct — but the branch has merge conflicts and CI isn't settled, so it can't merge as-is)
The i18n interpolation fix, the localized reason maps, and the detailed failure messages are all correct — I'd APPROVE the diff on its merits. I'm holding at COMMENT because the branch is currently **`mergeStateStatus=DIRTY` / `mergeable=CONFLICTING`** (real merge conflicts with base) and `code-review` CI is still PENDING. No code change is needed beyond a rebase to resolve the conflicts; once that's done and CI is green this is an APPROVE.

## Why COMMENT, not APPROVE
- **Merge conflicts** — `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` (re-polled, stable). The branch needs a rebase/merge-resolve against base before it can land. Not a code defect, but it does mean "merge this branch" is not currently possible.
- **CI not settled** — `code-review` is PENDING; no Build/secret-scan/osv leg is present (likely docs-path-filtered, but not confirmed green). Best to let CI finish after the rebase.

## The code is correct (verified on its merits)
- **Interpolation fix is right and applied at BOTH call sites.** Both `tr` helpers (`editor/importFlow.ts` and `import/markdown.ts`) now wrap params as `{ values: params }`, which is what the host `@octo/base` `t()` requires to substitute `{{reason}}`/`{{count}}`/`{{name}}` — the old bare-props form left them literal. I checked both helpers, not just one.
- **Rendering context is safe.** The interpolated strings flow into `setCreateError(...)` → `<p className="octo-docs-list-state octo-error">{createError}</p>` (DocsHome.tsx:933) — a JSX text child, React-auto-escaped, no `dangerouslySetInnerHTML`. And the interpolated values are constrained backend reason *codes*, not free user text. No injection surface.
- **Reason maps complete + locale-synced.** All 8 image-migrate codes (`source_not_found` … `fetch_failed`) have non-empty localized values, and zh-CN / en-US added the identical 15 keys (in lockstep — verified count + key names).
- **Fallbacks are safe (no symmetric-fix trap).** `reasonText` returns the raw backend code when unmapped (`text === key ? reason : text`), and the docx path uses the same `mapped === key ? fallback : mapped` guard — so a newly-added backend code never renders a bare i18n key or an empty string.
- **No capability dropped.** The `DocsHome.tsx` diff removes the `octo-docs-filter-group` wrapper `<div>` but **retains `<CreatorFilter>` with identical props** (re-indented, still imported at :24) — a layout re-wrap, not a filter removal. Confirmed against the base.

### Praise
- Fixing the interpolation at *both* `tr` seams (not just the one the bug report surfaced) and adding the raw-code fallback so an unmapped backend reason stays visible rather than silently blanking is the right instinct — it makes the failure path robust to backend codes this frontend doesn't know yet. The zh-CN/en-US key parity is clean.

### Net
Correct, well-scoped import-UX fix: placeholders now interpolate (fix applied at both translator seams), image-migrate reasons and docx failures are localized with safe raw-code fallbacks, render context is JSX-escaped, and no filter capability was dropped. The only thing keeping this from APPROVE is that the branch is in conflict (`CONFLICTING`/`DIRTY`) and CI is still pending — rebase to resolve, let CI go green, and it's good to merge. **Comment.**

Jerry-Xin (dismissed):
## Review: fix(docs): interpolate import warnings, localize image-migrate reasons, detail import failures (#842)

**Verdict: REQUEST_CHANGES** — the import-flow fix itself is correct and safe, but this PR's locale diff also **deletes two i18n key groups that are still referenced by live production code**, which would render raw i18n keys to users. This is byte-verified below. (The prior COMMENT review correctly validated the import-flow merits but did not audit the *deletions* in the locale diff against code references.)

## 🔴 Blocking

**1. Removed i18n keys are still referenced in shipped code → raw keys rendered to users.**
This PR removes, from BOTH `packages/docs/src/i18n/en-US.json` and `zh-CN.json`:
- the entire `share.*` block (`share.title`, `share.error`, `share.role.read/edit`, `share.loading`, `share.loadError`, `share.retry`, `share.restricted(Hint)`, `share.anyoneInSpace(Hint)`, `share.permission`)
- `table.reorderConflict`

But those keys are still used in live production components:
- `packages/docs/src/share/ShareScopePanel.tsx` references `t('share.*')` at 13 sites (lines 149, 193, 216, 220, 226, 239, 254, 255, 269, 272, 281, 284, 291). This panel is mounted in production via `packages/docs/src/members/MemberPanel.tsx:151` (`<ShareScopePanel ... />`).
- `packages/docs/src/editor/TableReorderHandle.ts:498` — `el.textContent = t('docs.table.reorderConflict')` inside `notifyReorderConflict()`, which is invoked in production at `TableReorderHandle.ts:682`.

Verified against base: `main` currently HAS both `share` (top-level) and `table.reorderConflict` in `en-US.json`; this PR's head removes them. So after this merges, the link-sharing panel and the table-reorder-conflict toast would display raw key strings (e.g. `share.title`, `docs.table.reorderConflict`) instead of localized text.

This looks like a side effect of the stale branch / merge conflict: the PR branch was cut before those keys landed on `main`, and its copy of the locale files drops them. **Fix: rebase onto `main`** so the two locale files retain `share.*` and `table.reorderConflict`, then confirm both keys survive in en-US.json and zh-CN.json before merge.

## ✅ Verified correct (import-flow scope — agree with prior review)
- **Interpolation fix applied at both translator seams.** `editor/importFlow.ts` and `import/markdown.ts` both wrap params as `t(key, { values: params })`, which the host `t()` needs to substitute `{{reason}}`/`{{count}}`/`{{name}}`; the old bare-props form left them literal.
- **Render context is XSS-safe.** Import warnings flow `consumeImportWarnings → setImportNotice(warnings.join(' ')) → <p role="status">{importNotice}</p>` (`editor/EditorShell.tsx:785`); import errors flow `setCreateError → <p className="octo-docs-list-state octo-error">{createError}</p>` (`pages/DocsHome.tsx:933`). Both are JSX text children (React-auto-escaped); no `dangerouslySetInnerHTML` anywhere in the docs package touches these messages, and interpolation is param-bound i18n, not string concat into markup. No injection surface.
- **Image-migrate reason mapping + graceful fallback.** All 8 reason codes (`source_not_found` … `fetch_failed`) are localized in both locales. `reasonText()` returns the raw backend code when unmapped (`text === key ? reason : text`) — never a bare i18n key or `undefined`.
- **docx failure detail correct + safe.** `docxReason.<reason>` lookup uses `mapped === key ? wordTooLarge : mapped`; `body?.error`/`body?.reason` access is optional-chained, no crash on missing fields.
- **New import keys are locale-symmetric.** en-US.json ↔ zh-CN.json: no key exists in one locale but not the other (parity confirmed for the full key set, including `import.reason.*`, `import.docxReason.*`, `import.wordEmpty`, `import.mdFailed`).

## 💬 Non-blocking / gate
- **Merge conflict:** `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` — this branch needs a rebase before it can merge (a rebase should also fix the blocker above).
- **Build not green:** no Build/secret-scan leg is present in the check list (gated behind the conflict). After the rebase, Build must run green before merge; `code-review` is still pending.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

[Octo-Q] verdict: REQUEST_CHANGES — Two P1 regressions: `share` i18n section and `reorderConflict` key deleted from both locale files while production consumers (`ShareScopePanel.tsx`, `TableReorderHandle.ts`) still reference them. Users will see raw i18n keys instead of localized text.

yujiawei (changes requested):
# Code Review — PR #842 (octo-web)

**Verdict: CHANGES_REQUESTED** (head SHA `ff644810`)

The intended fix is sound, but this branch is based on stale `main` and its diff silently reverts three unrelated, already-merged changes. Two of them are runtime i18n regressions that the test suite does not catch. The branch is also `CONFLICTING` against `main`.

---

## Part 1 — Spec compliance

The stated scope (linked to #841) is three things: (1) wrap import-warning interpolation params under `{ values }` so `{{reason}}`/`{{count}}`/`{{name}}` render; (2) localize image-migrate reason codes; (3) give specific whole-import failure messages.

- **What's required is implemented, and correctly** — see Part 2 core-fix note below. ✅
- **But the diff also does things outside that scope** (deletions of live i18n keys + a layout wrapper). These are not import-warning changes; they are the byproduct of branching off an old `main`. This is the blocking issue. ❌

**Spec: ❌** — the change exceeds the stated scope of the linked issue and diverges from the merged state of the base branch.

---

## Part 2 — Code quality

### ✅ Core fix is correct
`importFlow.ts` `tr()`/`reasonText()` and `import/markdown.ts` `tr()` now pass interpolation params as `{ values: params }`, which matches the host `@octo/base` `t()` convention already used across the codebase (`StatusBar.tsx`, `Toolbar.tsx`, `InvitePanel.tsx`, `DocsHome.tsx:352`). `reasonText()` and the `docxReason` mapping both fall back to the raw code / existing generic message for unmapped codes, so a new backend reason never yields an empty string. The `importFlow.images.test.ts` echo translator was updated to the new signature. No bug here.

### 🔴 P1 — Deletes `docs.share.*` i18n block, still consumed at runtime
`i18n/en-US.json` and `i18n/zh-CN.json` remove the entire `share` block (`title`, `restricted`, `restrictedHint`, `anyoneInSpace`, `anyoneInSpaceHint`, `permission`, `role.read`, `role.edit`, `error`, `loading`, `loadError`, `retry`). These keys are still actively used at HEAD by `share/ShareScopePanel.tsx` (lines 149, 193, 216, 220, 226, 239, 254, 255, 269, 272, 281, 284, and the `docs.share.role.*` lookup), which is rendered via `members/MemberPanel.tsx`. After this change the share-scope panel renders raw keys like `docs.share.title` instead of translated text.

Why the CI is green anyway: `ShareScopePanel.test.tsx` / `MemberPanel.test.tsx` mock `t()` as an echo (`getByText('docs.share.title')` asserts on the *key*), so the tests pass while the real UI is broken. This is a silent regression.

### 🔴 P1 — Deletes `docs.table.reorderConflict`, still consumed at runtime
Both locale files remove `table.reorderConflict`, but it is still referenced in production code at `editor/TableReorderHandle.ts:498` (`el.textContent = t('docs.table.reorderConflict')`). On a concurrent table-reorder conflict the user would see the raw key. This key was added by #822, which is one of the commits missing from this branch.

### 🟠 P1 — Removes the `octo-docs-filter-group` layout wrapper (out of scope)
`pages/DocsHome.tsx` drops the `<div className="octo-docs-filter-group">` that wraps `CreatorFilter` + `TypeFilter`. That wrapper is present on current `main` (`DocsHome.tsx:939`) and its own comment documents its purpose (keep Creator+Type wrapping to a second toolbar row together on narrow widths). The CSS class is still defined at `editor/styles.css:2760`, so removing the element is a responsive-layout regression, and it is unrelated to import warnings.

### Root cause — stale base / CONFLICTING
`git rev-list --left-right --count origin/main...HEAD` → `5 1` (5 behind, 1 ahead). The 5 missing commits are #822, #835, #837, #826, #824. GitHub reports `mergeable: CONFLICTING`; `git merge-tree` shows real conflicts in both locale files, where this branch still carries the freeze-panes strings that #826 removed and lacks the recent-row event keys added by #824. All three regressions above are the mechanical result of the diff being computed against an old snapshot of the locale/`DocsHome` files.

---

## Required before merge
1. Rebase this branch onto current `main` and regenerate the diff. This alone should eliminate the `share.*`, `reorderConflict`, and `octo-docs-filter-group` deletions, since they only appear because of the stale base.
2. After rebase, confirm the diff touches **only** the import-warning / image-migrate / docx-failure paths and their i18n keys — nothing under `share`, `table.reorderConflict`, freeze-panes, or the toolbar layout.
3. Re-run the docs suite after rebase to confirm the intended fix still passes on top of #822/#824/#826/#835/#837.

## Suggestions (non-blocking)
- The i18n echo-mock hides missing-key regressions. Consider a parity/coverage check that fails when a `t('docs.…')` call has no matching key in both locales, so deletions like these are caught in CI rather than in production.

packages/docs/src/i18n/en-US.json
```diff
@@ -185,6 +185,18 @@
     "word": "Import from Word",
```
OctoBoooot, line 185: **[major]** This PR's i18n diff deletes keys still referenced by live production code, so those surfaces will render raw i18n keys. Byte-verified absent at this head (grep count 0) but still called: `docs.table.reorderConflict` → `editor/TableReorderHandle.ts:498` (the collab reorder-cancel toast), and the whole `docs.share.*` block (`title`/`loading`/`loadError`/`restricted`/`restrictedHint`/`anyoneInSpace`/`permission`/`role`) → `share/ShareScopePanel.tsx` (:216/:220/:226/:254/:255/:269, ≥6 calls). Failing input: a concurrent table reorder shows the literal `docs.table.reorderConflict`; opening the share-scope panel shows raw `docs.share.*` keys. Almost certainly a bad merge/rebase artifact (the branch predates the PRs that added these blocks). Fix: rebase onto latest main and restore `table.reorderConflict` + `share.*` in BOTH locales so this diff only adds the import keys. Credit @Jerry-Xin. The zh-CN file has the identical deletion.

OctoBoooot (changes requested):
## Correcting my own #842 verdict: COMMENT → REQUEST_CHANGES — @Jerry-Xin found a real deleted-i18n-key regression I missed

Correcting my own verdict. My COMMENT said the code was correct and only the merge conflict blocked it. But @Jerry-Xin caught that this PR's i18n diff **deletes keys still referenced by live production code** — `docs.table.reorderConflict` and the entire `docs.share.*` block — so two UI surfaces would render raw i18n keys. Byte-verifying confirms it. That's a real user-facing regression, not just a conflict; upgrading to REQUEST_CHANGES. I under-reviewed by checking only the *added* keys' parity and not the *removed* ones.

## Major
- `packages/docs/src/i18n/{en-US,zh-CN}.json` — this PR removes `table.reorderConflict` and the whole `share.*` block (`share.title` / `loading` / `loadError` / `restricted` / `restrictedHint` / `anyoneInSpace` / `permission` / `role` / …). Byte-verified they are **absent at the head** (grep count 0) yet **still referenced by live source**:
  - `docs.table.reorderConflict` → `packages/docs/src/editor/TableReorderHandle.ts:498` (`el.textContent = t('docs.table.reorderConflict')` — the collab reorder-cancel toast).
  - `docs.share.*` → `packages/docs/src/share/ShareScopePanel.tsx` (≥6 live calls: `share.title` :216, `share.loading` :220, `share.loadError` :226, `share.restricted` :254, `share.restrictedHint` :255, `share.anyoneInSpace` :269, …).
  Failing input: trigger a concurrent table reorder (toast shows the literal `docs.table.reorderConflict`); open the document share-scope panel (title/loading/permission rows show raw `docs.share.*` keys). This is the exact "removed capability still referenced" regression — and it's almost certainly a **bad merge/rebase artifact**: the branch predates the PRs that added `share.*` / the reorder-conflict toast, and the i18n merge dropped those blocks instead of keeping them. Fix: rebase onto latest main and restore the `table.reorderConflict` + `share.*` keys (both locales) — this PR should only *add* the import keys, not remove unrelated ones.

## What I got wrong
I verified the newly-added import keys were zh/en in sync and the interpolation fix was correct — but I did not check whether the diff *deleted* keys that are still in use. That's the removed-key axis my own checklist calls out, and I skipped it. @Jerry-Xin's catch is correct; my COMMENT is superseded.

## What still stands (the import fix itself is correct)
- The interpolation fix (`{ values: params }` at both `tr` seams), the localized image-migrate reason map with raw-code fallback, the docx failure-detail messages, and the JSX-escaped render context are all correct as I verified — the *added* behavior is sound.
- The branch is also `mergeable=CONFLICTING` / `DIRTY`, so a rebase is required regardless; the key restoration should happen as part of that rebase.

## Net
The import-warning interpolation and localization work is correct, but the PR's i18n diff deletes `table.reorderConflict` + `share.*` keys that live code (`TableReorderHandle.ts`, `ShareScopePanel.tsx`) still calls, so the reorder-conflict toast and the share-scope panel would render raw keys — a real regression. Flipping my **COMMENT → REQUEST_CHANGES** @ `ff644810`: rebase onto latest main and restore those keys (both locales) so the diff only adds the import keys. Credit @Jerry-Xin for the deleted-key catch I missed. Resolve alongside the merge conflict; then this is an approve.

wwwzzzcccc:
Addressed the blocking review findings (new head `e1afb8b6`):

**Blocking (deleted i18n keys still used in production) — fixed.**
The branch was based on stale `main`, and overwriting the whole locale files dropped `docs.share.*` and `docs.table.reorderConflict`, which live production code (`ShareScopePanel.tsx`, `TableReorderHandle.ts`) still references. I restored both locale files from the current `main` and re-applied only the new import keys on top, so no existing key is removed.

**Merge conflict — resolved.** Rebuilt on current `main`; the diff is now purely additive on the locale files (+44 lines, the new `docs.import.reason.*` / `docs.import.docxReason.*` / `wordEmpty` / `mdFailed` keys) plus the interpolation and error-detail changes in `importFlow.ts` / `markdown.ts` / `DocsHome.tsx`.

zh/en keys are 1:1, `share`/`reorderConflict` confirmed present, import+editor+i18n suites 92/92, typecheck clean, single commit.

OctoBoooot (approved):
## Re-review (DELTA) #842 @ `e1afb8b6` — the removed-in-use i18n keys are fully restored, the conflict is resolved, and the import fix is intact → APPROVE

**Verdict:** Approve
Flipping my REQUEST_CHANGES → APPROVE. My major (with @Jerry-Xin / @mochashanyao / @yujiawei) was that the diff deleted `docs.share.*` + `docs.table.reorderConflict` keys still referenced by live code → raw-key renders. Byte-verified at this head: every deleted key is restored (and matches its usage), the merge conflict is resolved, and the import interpolation/localization fix I originally verified is intact. Code CI green.

## What I verified (the fix)
- **Every removed-in-use key is back.** All 11 `docs.share.*` keys that `ShareScopePanel.tsx` references — `title`, `loading`, `loadError`, `permission`, `role`, `restricted`, `restrictedHint`, `anyoneInSpace`, `anyoneInSpaceHint`, `error`, `retry` — are present again in en-US (and zh-CN), and `docs.table.reorderConflict` (still called at `TableReorderHandle.ts:498`) is restored. Grep confirms each referenced key resolves — the restoration is complete, not partial. The raw-key regression is closed.
- **Conflict resolved** — `mergeable=MERGEABLE` (was `CONFLICTING`/`DIRTY`). The rebase onto current main restored the keys and resolved the conflict in one move, exactly as the CR suggested.
- **Import fix preserved through the rebase** — both `tr` seams (`importFlow.ts` + `import/markdown.ts`) still wrap `params ? { values: params } : undefined` (the interpolation fix), and the new `docxReason` / `reason` / `mdFailed` / `wordEmpty` maps are intact with zh/en parity (2/2 markers each). The rebase kept the actual feature.
- **CI green** — Build `success`, secret-scan / osv-scan / dependency-review / pr-title-lint pass; check-sprint skipped (governance).

## What stands from my prior review
The import-warning interpolation, the localized image-migrate reason map with raw-code fallback, the docx failure-detail messages, and the JSX-escaped render context were all correct originally and are unchanged. The only defect (the deleted-key regression) is now fixed.

### Net
The diff no longer removes keys that live code uses — `docs.share.*` and `docs.table.reorderConflict` are restored and match their call sites, the conflict is resolved, and the import interpolation/localization fix is intact with zh/en parity. Code CI green (`size/M`, no governance block). Flipping my **REQUEST_CHANGES → APPROVE** @ `e1afb8b6`. Credit @Jerry-Xin for the original deleted-key catch. Peers holding CR at the prior head `ff644810` should re-anchor — the finding is fixed at this head.

Jerry-Xin (dismissed):
## Re-review — APPROVE ✅ (re-anchored at new head)

The bad-rebase artifact that blocked the previous round is resolved. Re-verified byte-for-byte at the current head.

### 🔴 Prior blocker — RESOLVED
The earlier diff deleted the entire `share.*` block + `table.reorderConflict` from both `en-US.json` and `zh-CN.json` while those keys are still referenced by live mounted code, which would have rendered raw keys to users. At this head:
- The PR diff-vs-base **no longer deletes** `share`/`reorderConflict` from either locale (deletions dropped 70→34).
- `share` and `reorderConflict` are **present in BOTH** `packages/docs/src/i18n/en-US.json` and `zh-CN.json` (symmetric `share.*` sub-tree: title/restricted/restrictedHint/anyoneInSpace/anyoneInSpaceHint/permission/role.read/role.edit/error/loading/loadError/retry).
- Referencers resolve: `packages/docs/src/share/ShareScopePanel.tsx` (13 `docs.share.*` refs, mounted via `members/MemberPanel.tsx:151`) and `editor/TableReorderHandle.ts:498` (`docs.table.reorderConflict`) now map to existing keys.

The conflict is cleared (`mergeable=MERGEABLE`).

### ✅ Import-flow fix — intact + XSS-safe (re-confirmed)
- Interpolation is param-bound: `tr()` in `editor/importFlow.ts` and `import/markdown.ts` now passes params under `opts.values`, so `@octo/base` t() substitutes placeholders instead of leaving literal `{{reason}}`.
- Image-migrate: `reasonText()` maps the 8 backend reason codes (localized in both locales) with raw-code fallback when unmapped (`text === key ? reason : text`).
- Word/docx failures in `pages/DocsHome.tsx`: `body?.error`/`body?.reason` optional-chained; docxReason uses `mapped === key ? wordTooLarge : mapped` fallback; 8 docxReason codes localized in both locales.
- No `dangerouslySetInnerHTML` in the touched files — messages render as JSX text (`<p className="…">{createError}</p>`), React auto-escaped.

### 💬 Non-blocking
- `octo-docs-filter-group` wrapper removal in `DocsHome.tsx` is a re-indent, not a capability change: `CreatorFilter` is retained with identical props and its recent-only guard, and `TypeFilter` still renders on both tabs. Cosmetic.

### CI
Build ✅, secret-scan (gitleaks) ✅, osv-scan ✅, dependency-review ✅, code-review ✅.

i18n parity between the two locales is clean (no key added to only one locale, no referenced-but-missing key). Clearing my prior request-changes.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #842 (octo-web)

## Summary

This PR fixes a real i18n interpolation bug, adds localized error messages for import failures (Word and Markdown), and restructures the Word-import error handler to surface specific backend failure reasons. The `Translate` type is corrected to pass interpolation values under `opts.values` — matching the actual `@octo/base` `t()` contract — which previously left `{{reason}}` and `{{count}}` placeholders as literal text in the UI. An unrelated toolbar-layout simplification is bundled in.

Overall, the changes are well-scoped, the defensive fallbacks are solid, and the test was updated in lockstep with the production signature change.

## Verification

- ✅ **Translate type alignment** — Traced `t` from `@octo/base` (`packages/dmworkbase/src/i18n/instance.ts:14`) through `octoweb/index.ts:9` re-export to `DocsHome.tsx:2` import. The real `t(key, options?: TranslateOptions)` expects `options.values` for interpolation; the PR's `tr()` wrapper now correctly wraps params under `{ values: params }`. The old code passed params as a bare second argument, which `I18nService.t()` at `:134` would interpret as options (no `values` key → no interpolation → literal `{{reason}}` in the UI).
- ✅ **Data flow: image migration warnings** — `nc.reason` originates from `copyAttachments()` response (`packages/docs/src/attachments/api.ts:184`), typed as `string`. The new `reasonText()` at `packages/docs/src/editor/importFlow.ts:158` maps known codes to localized text via `docs.import.reason.${reason}` keys; unknown codes fall back to the raw backend string. All 8 reason codes in the i18n JSON match the function's lookup path.
- ✅ **Data flow: Word import error handling** — `body.error` and `body.reason` are read from `(err as { response?: { data?: ... } })?.response?.data` at `packages/docs/src/pages/DocsHome.tsx:733-735`. The `import_unsafe` branch constructs `docs.import.docxReason.${body.reason ?? ''}` and falls back to `wordTooLarge` when the key is unresolved. The 8 `docxReason.*` keys in both locales cover the documented server codes.
- ✅ **en-US / zh-CN i18n parity** — Both locales add the same 20 new keys (`wordEmpty`, `mdFailed`, 8 `docxReason.*`, 8 `reason.*`). No missing or extra keys in either locale.
- ✅ **Test updated** — `echoT` at `packages/docs/src/editor/importFlow.images.test.ts:34` now expects `(key, opts?: { values?: ... })`, matching the production `Translate` type. The existing test at `:149-163` still asserts `toContain('source_forbidden')` on the warning string, which holds through the new `reasonText()` → localized text → interpolation chain.
- ✅ **Markdown extension-guard error catch** — `pickMdFile` at `:443` + extension guard at `:177` produce error messages containing "mdOnly", "Markdown", or ".md", all matched by the new regex at `:706`.

Static analysis only at head `e1afb8b6115b1bbe3c6116b0ec3d90254f705d3f`; build and tests not executed in this environment.

## Findings

### P2 — Orphaned CSS rule after toolbar wrapper removal (`packages/docs/src/pages/DocsHome.tsx:911-921`)

The `octo-docs-filter-group` wrapper div was removed, flattening CreatorFilter and TypeFilter into the toolbar's own flex container. The toolbar CSS at `packages/docs/src/editor/styles.css:2760` still defines `.octo-docs-filter-group { display: flex; align-items: center; gap: 8px; flex: 0 0 auto; }` — this rule is now unreachable. No visual regression (the parent `.octo-docs-toolbar` at `:2715` already provides `display: flex; gap: 8px; flex-wrap: wrap`), but the dead rule could confuse future contributors. Consider deleting it.

**Diff-scope**: new (introduced by this PR's toolbar simplification). Not pre-existing.

## Data Flow Tracing

- **Translate type → t() runtime**: `Translate` type `(key, opts?: { values?: Record<string, unknown> })` → `tr()` wraps params as `{ values: params }` → `t(key, { values: {...} })` → `I18nService.t()` at `:139` reads `options?.values` → `interpolate(message, values)` at `:141`. Correct chain.
- **nc.reason → warning string**: Server `copyAttachments` response → `result.notCopied[].reason: string` → `reasonText(t, nc.reason)` → `tr(t, 'docs.import.reason.${reason}')` → i18n lookup → `reasonText` returns localized text or raw code → `tr(t, 'docs.import.imageMigrateFailed', { reason: reasonText(...) })` → interpolated into `"An image could not be migrated to the new document ({{reason}})."`. Correct chain with unknown-code fallback.
- **body.reason → error toast**: Server docx import error response → `body.reason: string` → `t('docs.import.docxReason.${body.reason ?? ""}')` → if key unresolved (returns key), fallback to `t('docs.import.wordTooLarge')`. Correct chain with double fallback.
- **body.error → error toast**: Server docx import error response → `body.error: string` → switch on `'empty_upload'` | `'import_unsafe'` → localized message. Correct branching.

- **C1 — Dual-path parity**: N/A. No symmetric create↔delete or subscribe↔unsubscribe paths changed.
- **C2 — control-flow ordering**: Clear. `tr()` is called from multiple sites within importFlow.ts but always wraps params consistently. No nested or double-call risk.
- **C3 — Authorization boundary**: N/A. No auth/permission changes.
- **C4 — Authorization lifecycle**: N/A. No auth changes.
- **C5 — Build/runtime correctness**: Clear. The `Translate` type change is structural — any caller passing the old shape would get a TypeScript error. The new shape matches the runtime `@octo/base` `t()` contract.
- **C6 — Governance/policy**: N/A. No governance or policy changes.

- `reasonText()` fallback chain: unknown backend codes surface the raw code string rather than a meaningless i18n key path — correct defensive behavior.
- `I18nService.t()` at `:136-140` returns the key itself when no translation exists, so `text === key` comparison in `reasonText()` is a reliable sentinel.
- `body?.error === 'empty_upload'` branch: new i18n key `docs.import.wordEmpty` exists in both locales.
- Markdown-import `mdFailed` replaces the generic `docs.state.error` — more specific and actionable for the user.
- The `Translate` type widening from `Record<string, string | number>` to `Record<string, unknown>` is structurally compatible with `TranslationValues = Record<string, TranslationPrimitive>` since `TranslationPrimitive` is a subset of `unknown`.

## Verdict: APPROVED

Well-executed fix for a real interpolation bug, clean i18n parity, and thoughtful error-handling improvements with proper defensive fallbacks. The only item is a dead CSS rule cleanup — non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; one P2 (orphaned CSS rule) is non-blocking.

yujiawei (changes requested):
# Code Review — PR #842 (octo-web)

## Summary

The three docs-import fixes this PR targets (i18n placeholder interpolation, localized image-migrate reasons, detailed import-failure messages) are implemented correctly and are well-tested. However, the PR also **bundles an unrelated, out-of-scope change** to the docs-list toolbar that reverts a layout fix landed the same day and leaves dead CSS behind. That change is a blocker.

Verified locally at head SHA `e1afb8b`:
- Targeted suites pass: `src/import/`, `src/editor/importFlow.images.test.ts`, `src/i18n/` → **92/92**.
- i18n key parity zh-CN vs en-US: **19/19** import keys mirrored, no orphans; the new `reason.*`, `docxReason.*`, `wordEmpty`, `mdFailed` keys all resolve.
- The `{ values: params }` wrapping is correct in both `importFlow.ts` and `import/markdown.ts` — no dropped or double-wrapped params. `reasonText()` calls the inner `tr()` without params, so there is no accidental double-wrap.
- Typecheck and any `Toolbar.test` failures observed in a full-suite run are **pre-existing on `main`** (`dmworkbase/src/i18n/format.ts`, `src/members/sort.ts`; Toolbar test passes in isolation) and are not caused by this PR.

---

## Blocking

### B1 — Out-of-scope toolbar change reverts a same-day layout fix and leaves dead CSS  (`packages/docs/src/pages/DocsHome.tsx`, ~lines 908–923)

The diff removes the `<div className="octo-docs-filter-group">` wrapper around `CreatorFilter` + `TypeFilter`:

```diff
-        <div className="octo-docs-filter-group">
-          {activeView === 'recent' && (
-            <CreatorFilter ... />
-          )}
-          <TypeFilter selected={view.types} onToggle={view.toggleType} />
-        </div>
+        {activeView === 'recent' && (
+          <CreatorFilter ... />
+        )}
+        <TypeFilter selected={view.types} onToggle={view.toggleType} />
```

This is unrelated to the docs-import UX fixes described in the PR/linked issue, and it undoes deliberate behavior:

- That wrapper was introduced very recently to keep search + creator + type on one row and, when width is insufficient, wrap the **two filters together as one unit** rather than stranding the Type filter on its own row. Removing the wrapper reinstates exactly the wrapping regression that change fixed — on a narrow toolbar the two filters are now independent flex children of `.octo-docs-toolbar` (`flex-wrap: wrap`), so Type can wrap alone.
- The removed comment even documented this intent ("wrap together as one unit so a narrow toolbar drops the whole group to a second row"), and the deletion silently drops that guarantee.
- `.octo-docs-filter-group { display:flex; gap:8px; ... }` in `packages/docs/src/editor/styles.css:2760` is now **dead CSS** — no element references the class anymore.

Please either drop this hunk entirely from this PR (it belongs in a separate change with its own justification), or, if the layout change is genuinely intended, keep it out of a docs-import fix, restore/replace the responsive wrapping behavior, and remove the now-orphaned `.octo-docs-filter-group` rule.

---

## Non-blocking (nits / suggestions)

### N1 — Markdown error routing matches on translated error text (`DocsHome.tsx:704`)

```ts
} else if (err instanceof Error && /mdOnly|Markdown|\.md/i.test(err.message)) {
```

The wrong-file-type guard throws `new Error(tr(t, 'docs.import.mdOnly'))`, i.e. a **translated** string. The regex happens to match both current locales (zh "…Markdown 文件（.md / .markdown）" and en "Only Markdown files (.md / .markdown)…") and the raw key fallback, so this is not a live bug today — but it is brittle: any future wording/locale change, or an unrelated error whose message contains ".md", will misroute. The docx path already uses the stable pattern (`body?.error === 'import_unsafe'`). Prefer a typed error (a dedicated `Error` subclass, like `ImportContentCorruptError`) or an error-code field over matching on user-facing text.

### N2 — Docx `413` is checked before `import_unsafe`; confirm the backend contract (`DocsHome.tsx:735–745`)

```ts
if (status === 413) {
  setCreateError(t('docs.import.wordTooLarge'))
} else if (body?.error === 'import_unsafe') {
  const key = `docs.import.docxReason.${body.reason ?? ''}`
  ...
}
```

Several `docxReason` codes are size/complexity bounds (`total-too-large`, `entry-too-large`, `ratio-too-high`, `too-many-entries`, `too-many-media`) that a backend would plausibly return with HTTP **413**. If `import_unsafe` is ever served with a 413 status, the first branch wins and the specific per-reason message becomes unreachable — the user always sees the generic "too large" text, defeating part of the stated goal. This depends on the backend's status/error-code contract, which lives in a separate service and is not in this repo, so I could not confirm it. Please verify the contract; if `import_unsafe` can carry 413, reorder so the `body?.error === 'import_unsafe'` branch is evaluated first (falling back to `wordTooLarge` only when the reason is unknown).

### N3 — No test coverage for the new docx error branches

`empty_upload`, `import_unsafe` + per-`reason` mapping, and the `413` fallback in `onImportWord` have no unit test (the added tests cover image-migrate and markdown paths). A small test asserting each branch maps to the right key — especially the ordering in N2 — would lock the contract down.

---

## Coverage notes

- **Advisory legs:** two independent reviewer legs were requested for this complex PR. The long-context leg completed and independently flagged the same toolbar regression (B1) and the brittle markdown regex (N1). The line-level adversarial leg **failed at the gateway (infrastructure error) and did not return findings** — treated as absent, not as agreement; its coverage (deep line-level correctness of the error-branch ordering) was reconstructed manually above (N2).
- **Could not verify (out of this repo):** the backend contract for the docx importer — the exact HTTP status paired with `import_unsafe` / `empty_upload` and the full set of `reason` codes — lives in a separate service. N2 hinges on it.
- **Verified in-repo:** i18n wrapping correctness, key parity, `reasonText` fallback, targeted test suites, and that the pre-existing typecheck/Toolbar-test noise is not introduced by this PR.

## Verdict

The import fixes are good. The out-of-scope toolbar revert (B1) is a scope + regression + dead-code issue that should not ride along in this PR. Requesting changes to remove/repair B1; N1–N3 are recommendations.

Jerry-Xin (dismissed):
REQUEST_CHANGES @ `e1afb8b6115b` — **correcting my earlier APPROVE.** The i18n/import-flow fix itself remains correct and the previously-deleted keys are properly restored, but this PR mixes in an **out-of-scope toolbar layout change that is a responsive-behavior regression** — I initially mis-read it as a cosmetic re-indent. Credit to the reviewer who flagged it.

## 🔴 Blocking: removed `octo-docs-filter-group` wrapper is a responsive regression (out of scope)
In `packages/docs/src/pages/DocsHome.tsx` the toolbar diff removes the `<div className="octo-docs-filter-group">` that wrapped `CreatorFilter` + `TypeFilter`, unwrapping them into the toolbar directly. The removed wrapper was a **deliberate responsive construct** — its own removed comment documented it:

> Creator + Type wrap together as one unit so a narrow toolbar drops the whole group to a second row (search stays on row 1) instead of stranding Type alone.

Removing the wrapper means on a narrow toolbar the Creator/Type filters no longer wrap as a unit — reintroducing exactly the "Type stranded / bad wrap" behavior the group was added to fix. This is a behavior change with **no relation to the PR's stated scope** (`interpolate import warnings, localize image-migrate reasons, detail import failures`). An unrelated layout change mixed into an i18n fix should be split out (or dropped from this PR).

## 🔴 Blocking: `.octo-docs-filter-group` CSS is now dead code
The rule is still defined at `packages/docs/src/editor/styles.css:2760`, but this PR removes the only element that used it (the wrapper `<div>` above), and the diff does not remove the CSS. Result: an orphaned, unused rule. Either restore the wrapper or remove the now-dead CSS.

## 💬 Non-blocking (fair note)
Matching Markdown-import errors by translated-text pattern (`/mdOnly|Markdown|\.md/i.test(err.message)`) couples control flow to i18n copy — a copy edit could silently break the branch. Prefer keying on an error type/code over the message string. Non-blocking, but worth hardening.

## ✅ What's correct (unchanged from my re-review)
- The removed-in-use i18n keys blocker is **fully resolved** via the rebase: `docs.share.*` (all 11 keys referenced by `share/ShareScopePanel.tsx`, mounted at `members/MemberPanel.tsx:151`) + `docs.table.reorderConflict` (`editor/TableReorderHandle.ts:498`) are restored in **both** locales; the diff no longer deletes them.
- The import-flow fix is intact + XSS-safe: interpolation is param-bound `t(key, { values })` (not string-concat into markup), no `dangerouslySetInnerHTML`; `createError` renders as JSX text (React-escaped). The 8 `docxReason` codes are localized in both locales with `mapped === key ? wordTooLarge : mapped` fallback; `body?.error`/`body?.reason` are optional-chained.
- en/zh i18n parity is clean; conflict resolved (`mergeable=MERGEABLE`); CI green (Build ✅).

Fix: drop or split out the `octo-docs-filter-group` removal (keep the wrapper), or if the unwrap is intentional, also remove the now-dead CSS **and** justify the responsive change separately — it doesn't belong in this i18n PR.

packages/docs/src/pages/DocsHome.tsx
```diff
@@ -896,22 +908,17 @@ function DocsList({
       <DocsTabs active={activeView} onChange={onTab} />
       <div className="octo-docs-toolbar">
```
OctoBoooot, line 909: **[major]** Removing the `octo-docs-filter-group` wrapper (this PR's diff deletes `<div className="octo-docs-filter-group">` here) unbundles CreatorFilter+TypeFilter from their shared flex group. `.octo-docs-toolbar` is `flex-wrap: wrap` (styles.css:2715) and the removed `.octo-docs-filter-group` was `flex:0 0 auto` (:2760) — it grouped the two filters as ONE non-shrinking wrap item so a narrow toolbar dropped them together. Now they're independent wrap items, so on a narrow viewport TypeFilter can strand alone on a second row (the exact layout the wrapper's own comment said it existed to prevent). Failing input: narrow the docs-home window on the Recent tab until the toolbar wraps → Type reflows separately from Creator. Also unrelated to this PR's docs-import scope, and it orphans the `.octo-docs-filter-group` CSS rule (now 0 references). Fix: restore the wrapper around the two filters (and keep its CSS), or split this layout change into its own PR with a responsive before/after. Credit @yujiawei — I under-graded this as a nit originally (checked props, not the wrapper's flex-grouping role).

OctoBoooot (changes requested):
## Correcting my own #842 verdict: APPROVE → REQUEST_CHANGES — @yujiawei found a real toolbar responsive regression I under-graded

Correcting my own verdict again on this PR. I flipped to APPROVE this turn once the deleted-i18n-key blocker was fixed — but @yujiawei caught a *different* axis at this same head that I mis-graded as a non-blocking nit in my original review: removing the `octo-docs-filter-group` wrapper changes the toolbar's responsive reflow (unrelated to the docs-import scope) and leaves dead CSS. Byte-verifying confirms it's real. My APPROVE was premature on this axis; correcting it.

Gate `CHANGES_REQUESTED` / `BLOCKED`.

## What I verified (the axis I under-graded)
- **The change is in THIS PR's diff** — `DocsHome.tsx` removes `-<div className="octo-docs-filter-group">` and re-parents `<CreatorFilter>` + `<TypeFilter>` as direct children of `.octo-docs-toolbar`. This is unrelated to the PR's stated docs-import-warning scope (scope creep).
- **It's a real responsive regression.** `.octo-docs-toolbar` is `display:flex; flex-wrap: wrap` (styles.css:2715-2722). The removed `.octo-docs-filter-group` was `display:flex; gap:8px; flex:0 0 auto` (:2760-2765) — it grouped CreatorFilter+TypeFilter as ONE non-shrinking flex item so a narrow toolbar dropped the whole group to the next row together. The original code comment said exactly this ("wrap together as one unit so a narrow toolbar drops the whole group to a second row instead of stranding Type alone"). After removal, the two filters are independent wrap items, so on a narrow viewport `TypeFilter` can strand alone on its own row — the exact layout the wrapper existed to prevent. Failing input: narrow the docs-home window on the Recent tab until the toolbar wraps → Type filter reflows separately from Creator instead of as a unit.
- **Dead CSS confirmed** — `.octo-docs-filter-group` is now referenced in 0 `.tsx` files but still defined in `styles.css` (orphaned rule).

## Correcting my record
In my original review I graded this wrapper removal a non-blocking nit ("re-indent, CreatorFilter retained with identical props → no capability loss"). That was wrong: I checked the component's *props* but not that the *wrapper itself carried layout semantics* (the flex grouping that drives responsive reflow). Removing it is a behavior change, not a re-indent. @yujiawei's catch is correct; my APPROVE is superseded on this axis. (This is the same class of miss as the deleted-i18n-keys earlier on this PR — checking one dimension of a removal, not the removed element's full role.)

## Major
- `packages/docs/src/pages/DocsHome.tsx:909` (toolbar) — removing the `octo-docs-filter-group` wrapper unbundles CreatorFilter+TypeFilter from their shared `flex:0 0 auto` group, so on a wrapping (`flex-wrap:wrap`) toolbar the Type filter can strand alone on a second row on narrow viewports — a responsive regression, and unrelated to this PR's docs-import scope. Fix: restore the `octo-docs-filter-group` wrapper around the two filters (or move this layout change to a separate PR with its own before/after responsive check). Grading **major** (real reachable layout defect + scope creep into an unrelated file).

## Minor
- `packages/docs/src/editor/styles.css:2760` — `.octo-docs-filter-group` is now unreferenced (0 `.tsx` uses) but still defined. If the wrapper stays removed, delete the rule; if restored per the major, this resolves itself.

## What still stands (correct)
- The removed-in-use i18n keys are fully restored (`share.*` + `reorderConflict`, both locales — verified this turn), the merge conflict is resolved, and the import interpolation/localization fix (both `tr` seams wrap `{ values }`, docx/reason maps, JSX-escaped render) is intact. The **import portion of this PR is correct**; the blocker is the unrelated toolbar-layout change riding along.
- yujiawei's 🟡 (Markdown error-matching keys off translated text → prefer an error type/code) is a fair non-blocking robustness note.

## Net
The docs-import interpolation/localization fix is correct and the earlier i18n-key regression is resolved — but this PR also carries an unrelated toolbar change that removes the `octo-docs-filter-group` wrapper, regressing the narrow-viewport reflow (Type filter strands alone) and orphaning its CSS. Flipping my **APPROVE → REQUEST_CHANGES** @ `e1afb8b6`: restore the wrapper (or split the layout change out with its own responsive verification) and remove the dead rule if it stays out. Credit @yujiawei for the layout regression I under-graded.

wwwzzzcccc:
Addressed the out-of-scope toolbar regression (new head `378d872a`):

**Blocking (removed `octo-docs-filter-group` wrapper — responsive regression) — fixed.**
Same root cause as the earlier deleted-i18n-keys issue: the branch was cut from stale `main` and the wholesale copy of `DocsHome.tsx` dropped the `octo-docs-filter-group` wrapper that landed on `main` the same day, changing the toolbar's responsive reflow — unrelated to the import scope.

I restored `DocsHome.tsx` from current `main` and re-applied only the two import-error-handler changes (the Markdown and Word `catch` blocks). The `octo-docs-filter-group` wrapper and its layout are now byte-identical to `main`; the only remaining diff in this file is the two catch blocks (+21/-9).

Diff is now strictly in scope: `importFlow.ts` / `markdown.ts` interpolation, `DocsHome.tsx` import-error handlers, and additive `docs.import.*` i18n keys. import+editor+i18n suites 92/92, typecheck clean, single commit.

Jerry-Xin (approved):
APPROVE @ `378d872abcbf` (re-anchor) — both blockers from my prior review are resolved, and the underlying i18n/import fix is intact. Byte-verified at the new head.

## ✅ Responsive regression — RESOLVED
`packages/docs/src/pages/DocsHome.tsx` restores the `<div className="octo-docs-filter-group">` wrapper (line ~914) around **both** `CreatorFilter` and `TypeFilter`, inside `.octo-docs-toolbar`. This brings back the flex-unit grouping (`flex: 0 0 auto`) so a narrow, `flex-wrap: wrap` toolbar drops Creator+Type as one unit to the second row instead of stranding `TypeFilter` alone — exactly the behavior the wrapper existed to guarantee. The out-of-scope layout change is fully removed from this PR (the diff no longer touches `octo-docs-filter-group`).

## ✅ Orphaned CSS — RESOLVED
`.octo-docs-filter-group` (defined at `packages/docs/src/editor/styles.css`) is referenced again by the restored wrapper — no longer dead code.

## ✅ Underlying fix intact (re-confirmed)
- The removed-in-use i18n keys stay resolved: `docs.share.*` + `docs.table.reorderConflict` present in both locales; the diff no longer deletes them.
- Import-flow fix is intact + XSS-safe: interpolation is param-bound `t(key, { values })` in `editor/importFlow.ts`; no `dangerouslySetInnerHTML` in `DocsHome.tsx`; error/notice strings render as React-escaped JSX text. Image-migrate/docxReason localization + fallbacks unchanged.

## 💬 Non-blocking (carried from prior review)
Matching Markdown-import errors by translated-text pattern couples control flow to i18n copy — prefer keying on an error type/code. Non-blocking.

CI green at this head: Build ✅, secret-scan ✅, osv-scan ✅, dependency-review ✅. Deletions dropped 34→18 (scope tightened to the i18n/import fix). Clearing my prior request-changes.

OctoBoooot (approved):
## Re-review (DELTA) #842 @ `378d872a` — both majors resolved: the `octo-docs-filter-group` wrapper is restored (responsive grouping back, CSS no longer dead) and no i18n keys are deleted; import fix intact → APPROVE

**Verdict:** Approve
Flipping my REQUEST_CHANGES → APPROVE. Across this PR's rounds I held two majors: (1) the deleted-in-use `share.*`/`reorderConflict` i18n keys, and (2) the out-of-scope toolbar `octo-docs-filter-group` wrapper removal (responsive regression + orphaned CSS). Byte-verified at this head: both are resolved, the diff is scoped back to the import fix, and CI is green.

## What I verified (both majors fixed)
- **Toolbar responsive regression fixed** — `DocsHome.tsx:914` restores `<div className="octo-docs-filter-group">` wrapping both `<CreatorFilter>` (:916) and `<TypeFilter>` (:925) as one `flex:0 0 auto` group again, so on the wrapping (`flex-wrap:wrap`) toolbar the two filters reflow together and Type no longer strands alone on a narrow viewport. The PR diff no longer touches the wrapper at all (0 `filter-group` hits in the diff) — the out-of-scope layout change is fully removed.
- **Orphaned CSS resolved** — `.octo-docs-filter-group` is referenced again by the restored wrapper, no longer dead code.
- **No i18n key deletions** — 0 removed key-lines in both locales (was the earlier regression); the previously-deleted-in-use keys (`reorderConflict`, `loadError`, `restrictedHint`, `anyoneInSpace`, and the rest of `share.*`) remain present (matching their `TableReorderHandle.ts` / `ShareScopePanel.tsx` call sites). The diff is now `+22/-0` per locale — additive only.
- **Import fix intact** — both `tr` seams (`importFlow.ts` + `import/markdown.ts`) still wrap `params ? { values: params }`; the `docxReason`/`reason`/`mdFailed`/`wordEmpty` maps are present with zh/en parity (4/4). The interpolation/localization work I verified correct originally is preserved.
- **CI green** — Build `success`, secret-scan / osv-scan / dependency-review / pr-title-lint pass; check-sprint skipped (governance).

## Net
The PR is now scoped to its stated purpose: import-warning interpolation + localized image-migrate reasons + detailed import failures. The two regressions I (and peers) flagged across the rounds — deleted-in-use i18n keys and the out-of-scope toolbar wrapper removal — are both fixed, the responsive grouping and all referenced keys are restored, and CI is green. Flipping my **REQUEST_CHANGES → APPROVE** @ `378d872a`. Credit @Jerry-Xin / @yujiawei for the deleted-key and responsive-regression catches. Peers holding CR at the prior head `e1afb8b6` should re-anchor — both findings are fixed at this head.

## Personal note on this PR's arc
I got this PR wrong twice before landing here — first at COMMENT (missed the deleted-in-use i18n keys), then at APPROVE (under-graded the wrapper removal as a re-indent nit). Both were the same failure: on a diff that *removes* an element (an i18n key, a wrapper div), I verified the survivors but not what the removed thing actually did. Peers caught each; I byte-verified and corrected both. The lesson is logged: for any deletion, grep the removed identifier's references and check its layout/semantic role before signing off.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #842 (octo-web)

Reviewed at head `378d872a`. Scope: i18n interpolation fix for import warnings, localized reason codes for image-migrate and Word-import failures, detailed Markdown-import error message.

## Summary

This PR fixes a real bug where `{{reason}}` / `{{count}}` / `{{name}}` placeholders in i18n strings were rendered literally instead of being substituted — the old `Translate` signature passed interpolation params as bare props, but the host `@octo/base` `t()` expects them wrapped under `{ values: ... }`. The fix is applied consistently across both `tr()` helpers (`importFlow.ts` and `markdown.ts`), the `Translate` type, and the test mock. Additionally, the PR adds localized human-readable reason codes for image migration failures and Word import safety rejections, and introduces a specific error message for empty-file and failed Markdown imports. All 18 new i18n keys have parity between en-US and zh-CN.

## Verification

- ✅ **`Translate` type parity** — `importFlow.ts:137`, `markdown.ts:54`, and `importFlow.images.test.ts:34` all declare `(key: string, opts?: { values?: Record<string, unknown> }) => string`. Consistent across all three sites.
- ✅ **`tr()` wrapping parity** — `importFlow.ts:148` and `markdown.ts:168` both wrap params as `params ? { values: params } : undefined`. Identical logic.
- ✅ **i18n key parity** — all 18 new keys (`wordEmpty`, `mdFailed`, 8× `docxReason.*`, 8× `reason.*`) present in both `en-US.json` and `zh-CN.json` with matching structure.
- ✅ **Data flow: `reasonText` → `tr`** — `reasonText(t, nc.reason)` at `importFlow.ts:161` constructs key `docs.import.reason.${reason}`, looks it up via `tr(t, key)` (no params, so no wrapping needed). Falls back to raw backend code when no translation exists (`text === key ? reason : text`). The result is then passed as `{ reason: reasonText(...) }` into `tr(t, 'docs.import.imageMigrateFailed', ...)`, which correctly wraps under `{ values: { reason: ... } }`. The `{{reason}}` placeholder in the i18n string will be substituted.
- ✅ **Data flow: `DocsHome.tsx` Word-import error handling** — `body.reason` from server response is used to construct `docs.import.docxReason.${body.reason ?? ''}`. When `body.reason` is undefined, the key has a trailing dot, won't match, and falls back to `wordTooLarge`. When it matches a known code, the localized text is shown. Correct fail-safe behavior.
- ✅ **Test mock alignment** — `echoT` in `importFlow.images.test.ts:34` matches the new `Translate` signature. The `notCopied` test at `:153` correctly exercises the `reasonText` fallback path (echo translator returns key as-is → `reasonText` returns raw code `source_forbidden`), and assertions verify both the i18n key and the raw reason appear in the warning.

## Findings

No P0/P1 issues. Two non-blocking observations below.

### Nit — `copy_failed` and `store_failed` share identical translations (`packages/docs/src/i18n/en-US.json:213-214`)

Both `reason.copy_failed` and `reason.store_failed` map to "failed to save the image" (en-US) / "图片保存失败" (zh-CN). The backend presumably distinguishes these failure modes for diagnostic purposes, but the user-facing message is identical. This is likely intentional — end users don't need to know whether it was a copy or store operation that failed — but worth confirming the backend actually emits both codes and they don't carry different user-action implications.

### Nit — `reasonText` happy path not exercised in tests (`packages/docs/src/editor/importFlow.images.test.ts:153`)

The existing `notCopied` test passes `echoT`, which returns the key as-is, so `reasonText` takes the fallback branch (returns raw code). The happy path — where the translator resolves `docs.import.reason.${code}` to a localized string — is not covered. A mock translator with a small lookup table for one or two reason codes would lock the happy-path contract. Not blocking; the fallback path coverage is adequate for a helper this simple.

- `DocsHome.tsx` calls `t()` directly (not through `tr()`) for `docxReason` lookups at `:744`. This is correct — those lookups have no interpolation params, so the `{ values }` wrapping is unnecessary.
- `body.reason ?? ''` fallback when `body.reason` is undefined produces a non-matching key with trailing dot, which correctly triggers the `wordTooLarge` fallback at `:745`.
- The `import_unsafe` error branch at `DocsHome.tsx:738` correctly distinguishes from generic `wordError` and `wordTooLarge` paths.
- No stale references to the old `Translate` signature — all three sites updated in lockstep.
- `markdown.ts:54` `ImportOptions.t` type matches the new `Translate` signature. No other consumers of `ImportOptions.t` in the checkout use the old bare-params form.

## Verdict: APPROVED

Clean, focused fix with correct data flow and consistent type updates across all three sites (production, parser, test). The interpolation bug was real and the fix direction is sound. No correctness, security, or build concerns.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. The i18n interpolation fix is correct and consistently applied across all sites. Two non-blocking nits on identical translations for distinct backend codes and missing happy-path test coverage for `reasonText`.

yujiawei (approved):
# Code Review — PR #842 (octo-web)

**Head SHA:** `378d872abcbfeab9f1d591329d470e5bf8278e17` · **Base:** `main` · 6 files, +101/-18

## Verdict: ✅ Approve

The change does exactly what it says: (1) fixes i18n interpolation so params reach the host translator, (2) localizes image-migrate failure reasons, and (3) surfaces precise docx import failures. All findings below are advisory (P2/nit) — none block merge. Affected package tests pass (130/130) and the change introduces no new typecheck errors (the two pre-existing `tsc` errors in `format.ts` and `members/sort.ts` are present on the base branch, unrelated to this diff).

## 1. Spec compliance

- **Interpolation fix — correct.** `tr()` now wraps params as `t(key, { values: params })` in both `importFlow.ts:145-151` and `import/markdown.ts:164-171`. Verified against the host contract: `@octo/base` (`packages/dmworkbase/src/i18n/I18nService.ts:133-141`) interpolates `{{...}}` **only** from `options.values`, so the previous bare-param form left `{{reason}}` literal. This matches existing production call sites (`DocsHome.tsx:352`, `StatusBar.tsx:108`, `InvitePanel.tsx:89`). The echo translator in `importFlow.images.test.ts:34-35` was updated to the new signature.
- **Localized migrate reasons — correct.** `reasonText()` (`importFlow.ts:158-164`) maps `docs.import.reason.<code>` and falls back to the raw backend code for unknown codes (`text === key ? reason : text`) — good forward-compat. Wired into `notCopied[].reason` at line 280.
- **Detailed docx failures — correct.** `DocsHome.tsx:732-748` branches on `413` → `wordTooLarge`, `empty_upload` → `wordEmpty`, `import_unsafe` → `docxReason.<reason>` with a `mapped === key` fallback to `wordTooLarge`. The axios error propagates unwrapped from `importDocx` (`docsApi.ts:369-382`, no re-wrap in `runDocxImport`), so `.response.status` / `.response.data` are intact — the branches are live, not dead.
- **i18n parity — correct.** All new keys exist in both `en-US.json` and `zh-CN.json` with identical structure; the parity test (`i18n.test.ts`) passes. Nested groups (`docxReason.*`, `reason.*`) resolve because the host flattens nested JSON into dotted keys, and the interpolation regex `[\w.-]` handles hyphenated codes like `not-a-zip`.

No missing scope, no out-of-scope additions.

## 2. Code quality — advisory findings (non-blocking)

**P2 — Wrong-file-type detection couples to localized message text.** `DocsHome.tsx:704` classifies the markdown import error via `/mdOnly|Markdown|\.md/i.test(err.message)`, but the guard at `importFlow.ts:178` throws the **translated** string, not a typed error. It works in both shipped locales only because both embed the literal `Markdown` / `.md` tokens. A future locale that fully localizes the phrase would silently fall through to the generic `mdFailed`. The same PR already uses the robust pattern on the docx path (structured `body.error` codes) and the file already has `ImportContentCorruptError` classified by `instanceof` — consider a typed/sentinel error here too. Degrades gracefully today; not a live bug.

**P2 — Dead i18n key `docs.import.reason.fetch_failed`.** This code only originates from the external-image `notIngested[].reason` path (`ingestAttachments`), but that path never routes through `reasonText()` — it emits the generic `externalImagesLinked` warning and discards the reason (`importFlow.ts:329`). So `fetch_failed` (an external-download failure by its own wording) can never reach the copy-path mapper. Tests confirm the reason is discarded on the ingest path. Either thread `notIngested[].reason` through `reasonText()` for symmetric per-image external warnings, or drop the key.

**P2 — Weak test coverage for the headline mapping.** `importFlow.images.test.ts:157-161` asserts the warning `toContain('source_forbidden')`, but with the echo translator that passes via the raw-code **fallback** branch, not the localized mapping. No affected test asserts a `reason.*` code resolving to its actual phrase, and none covers `docxReason.*`. A symmetric rename/delete of a reason key in both locales would go uncaught (one-sided is caught by the parity guard). The mapping logic itself is correct — this is a coverage gap, not a production defect.

**nit — `docxReason.*` phrases render as the whole toast.** For `import_unsafe`, the lowercase sentence fragment (e.g. "a file inside the document is too large") is shown bare as the entire error message, whereas sibling branches (`wordTooLarge`, `wordEmpty`) and the image `reason.*` phrases are full/wrapped sentences. Cosmetic, en-US only (zh-CN fragments read fine standalone). Consider wrapping in a "Couldn't import: {{reason}}" template.

**nit — client transport timeout never reaches `docxReason.timeout`.** `docxReason.timeout` is only reachable via a backend `import_unsafe` response; a client-side axios timeout (120s, `docsApi.ts:378`) has no `.response` and lands on `wordError`. This is arguably correct (server parse-timeout vs client transport-timeout are distinct), but a one-line comment would prevent a future maintainer from mis-wiring it.

## 3. Coverage notes

- **Backend contract not verifiable from this repo.** `octo-server` is not in this workspace, so I could not confirm: (a) whether HTTP 413 is ever co-shipped with a `{error:'import_unsafe', reason}` body — if it were, the 413-first ordering (`DocsHome.tsx:735` before `:741`) would show the coarse `wordTooLarge` instead of the precise reason. The 413 branch comment suggests the author treats these as disjoint channels, and worst case is a semantically-equivalent message, so this is not a blocker. (b) The exact set of `reason` codes `copyAttachments` can return — unknown codes degrade gracefully to the raw code, so this is safe by construction.
- **Second-opinion advisory legs (external reviewers) timed out on this run and are recorded as absent, not as approval.** This verdict rests on direct source analysis plus internal multi-reviewer adversarial verification (all findings cross-checked against the code).

PR #849, opened 2026-07-16
fix(electron): make macOS package runnable and update icon

botshen:
## Summary

- fix packaged Electron runtime dependencies, Node crypto usage, and preload entry resolution
- build the Electron renderer with relative asset URLs and preserve the packaged file URL after login
- expose IPC listener cleanup required by the authenticated main view
- run the Electron renderer build as part of every desktop packaging command
- replace the macOS app and notification icon assets with the new purple OCTO artwork

## Validation

- pnpm install --frozen-lockfile
- VITE_API_URL=http://127.0.0.1:8080 pnpm build
- VITE_API_URL=http://127.0.0.1:8080 pnpm --filter @octo/web build:electron
- pnpm --filter @octo/web compile:electron
- postLoginRedirect.test.ts: 2 passed
- SessionScope.test.ts: 7 passed
- packaged and signed an arm64 OCTO.app, mounted the generated DMG, and verified the packaged icon
- launched both the login page and authenticated main view from the packaged app without renderer exceptions


Jerry-Xin (approved):
## ✅ APPROVE

Electron macOS packaging fix + icon refresh. Reviewed the ~90 LOC code delta (12 of 21 files are icon/logo PNG + `.icns` binary swaps, 0/0 text delta — benign asset updates). Security-sensitive surfaces (preload IPC boundary, post-login redirect, macOS signing/entitlements, `webPreferences`) all check out.

### 🔎 Security review (byte-verified @ head)

**Electron preload (`apps/web/src-election/preload/index.ts`, +10)** — adds a `removeListener` bridge that mirrors the existing `on`/listener guard: it is gated by `ALLOWED_RECEIVE_CHANNELS.includes(channel)` and otherwise logs + blocks. No new IPC channel, no Node/`fs`/`child_process`/`shell` exposure, no passthrough. `contextBridge` still exposes a fixed allow-list. Safe.

**Post-login redirect (`src/Layout/postLoginRedirect.ts` +16 new, `Layout/index.tsx` +7/-1, test +29)** — no open-redirect introduced. The helper only:
- `file:` protocol → mutates `search`/`hash` on the *current* URL (stays on the packaged `index.html`), or
- web → `${origin}${basePath}/${query}` where `origin` is `window.location.origin` (current origin, not attacker-controlled) and `query` is `redirectSearch`, which is derived as `redirectQs ? "?"+redirectQs : ""` — always a leading-`?` query string, never a full/absolute URL.

This is the same redirect semantics as the previous inline `` `${window.location.origin}${basePath}/${redirectSearch}` `` line, refactored to additionally handle the Electron `file://` case. No user-controlled absolute destination is ever followed. The test covers both the Electron `file://` and web branches.

**macOS packaging (`apps/web/package.json` +13/-11)** — no signing/entitlement loosening. Changes are:
- new `build:electron` script (`VITE_ELECTRON_BUILD=true vite build`) prepended to every `build-ele:*` target,
- added `ms@^2.1.3` dependency.

The `electron:build:*` targets, `entitlements.mac.plist`, hardened-runtime and code-signing config are untouched (the plist is not in this PR's file set). No `allow-unsigned-executable-memory` / `allow-dyld-environment-variables` / `disable-library-validation` / sandbox-disable additions.

**`vite.config.ts` (+4)** — sets `base: "./"` only when `VITE_ELECTRON_BUILD === "true"` so `file://`-loaded assets resolve relative to `index.html`; web build stays rooted at `/`. Correct root cause for "not runnable" — build config only, no security impact.

**`src-election/main/index.ts` (+1/-2), `createWindow.ts` (+1/-1)** — both only change the preload path string `preload/index` → `preload/index.js` (correct compiled-JS resolution). No removed security control. `getWindowConfig` keeps `nodeIntegration:false` + `contextIsolation:true`. `createWindow`'s `nodeIntegration:true` is **pre-existing** and unchanged by this PR (it loads a local `build/index.html` via `loadFile`, not remote content) — out of scope here.

**`search.ts` (+6/-3)** — swaps browser `crypto.getRandomValues` for Node `randomBytes(4)` in `getRandomSid`; both are CSPRNG, entropy unchanged. Runs in the Electron main process where `node:crypto` is the correct API. No shell/exec/fs, no user input. Safe.

**Dependency** — `ms@^2.1.3` (millisecond string parser). OSV-Scanner (pr + scheduled) PASS, dependency-review PASS, secret-scan PASS.

### 💬 Non-blocking notes
- Pre-existing: `createWindow.ts` uses `nodeIntegration:true` without `contextIsolation`. Not touched by this PR, but worth hardening in a follow-up if that window can ever load anything other than the trusted local bundle.
- `check-sprint` is failing because the PR has **no linked issue** (`No linked issue found — add a 'Closes #<issue>'`). That's a governance gate, not a code problem — please add the `Closes #<issue>` reference so the required check can pass before merge.

### ✅ Gate
Build is now **green**; osv-scan / dependency-review / secret-scan PASS. No code blockers. Please clear the governance `check-sprint` (link an issue) before merging.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #849 (octo-web)

## Summary

This PR fixes the packaged macOS Electron app so it actually launches and runs, and updates the app icon set. Three real bugs are addressed: the Vite build emitted absolute asset paths that break under `file://` protocol, the preload script path lacked an explicit `.js` extension (causing packaged-app load failures), and `getRandomSid()` used the Web Crypto API (`crypto.getRandomValues`) which is unavailable in the Electron main process. A post-login redirect helper is introduced to keep Electron on `file://` after authentication, and a `removeListener` IPC bridge method is added to support listener cleanup. Build scripts are updated to include a Vite build step before TypeScript compilation.

## Verification

- ✅ **Preload path** — `join(__dirname, "..", "preload", "index.js")` resolves to `out-election/preload/index.js` from `out-election/main/index.js`, matching `tsconfig.e.json` `outDir: ./out-election` and the `files: ["out-election/**/*"]` entry in `electron-builder.js:8`.
- ✅ **Vite base path** — `base: "./"` (Electron) produces relative asset paths from `build/index.html`; `base: "/"` (web) preserves existing behavior. The `VITE_ELECTRON_BUILD` env var is set only in the new `build:electron` script, so regular `vite build` is unaffected.
- ✅ **Crypto fix** — `randomBytes` from `node:crypto` replaces `crypto.getRandomValues` in `src-election/main/utils/search.ts:1-2`. Callers at `main/index.ts:359,363,401` receive the same 6-char base-36 SID string shape.
- ✅ **Post-login redirect** — `buildPostLoginRedirectUrl` at `src/Layout/postLoginRedirect.ts:1-16` detects `file:` protocol via `new URL(currentHref).protocol`, preserves the existing HTTP path verbatim (`${origin}${basePath}/${query}`), and handles empty `query` correctly (trailing `/` on HTTP, empty search on file). Test at `src/Layout/__tests__/postLoginRedirect.test.ts` covers both branches.
- ✅ **IPC removeListener** — `preload/index.ts:70-79` gates on `ALLOWED_RECEIVE_CHANNELS`, matching the existing `on`/`once` security pattern. Consumer at `Pages/Main/vm.ts:186` calls `ipc.removeListener(event, handler)` during `didUnMount`.
- ✅ **Build pipeline** — All `build-ele:*` scripts now run `build:electron` (Vite with `VITE_ELECTRON_BUILD=true`) before `compile:electron` (TSC), ensuring the frontend is built with correct relative asset paths before packaging.

Static analysis only at head `51f20d939090e873e8044f564c0e813795df6a80`; build and tests not executed in this environment.

## Findings

### P2 — Unused `ms` dependency (`apps/web/package.json:81`)

`"ms": "^2.1.3"` is added to `dependencies` and resolved in `pnpm-lock.yaml:233-235`, but no file under `apps/web/src-election/` or `apps/web/src/` imports it (`grep -rn "from 'ms'" / "require('ms')"` returns empty). This adds dead weight to the install tree and the asar bundle. Either remove the entry or note the intended consumer if it's needed by a planned follow-up.

- Icon binary replacements across all 11 sizes + `icon.icns` — consistent with PR title, no code references to validate beyond `electron-builder.js` icon paths which point to `resources/icons/icon.icns` correctly.
- `build-ele:mac-x64` lacks `clean:dist-ele` but this is a pre-existing asymmetry, not introduced by this PR.
- `removeListener` in preload does not conflict with existing `notification-clicked` / `notification-action-clicked` removeListener calls at `preload/index.ts:93,101` (those operate directly on `ipcRenderer`, not through the bridge).
- Data flow for `redirectSearch`: constructed from `URLSearchParams.toString()` at `Layout/index.tsx:161-162`, always either `"?key=val"` or `""`, consumed safely by both branches of `buildPostLoginRedirectUrl`.

## Verdict: APPROVED

Well-targeted fixes for real packaged-app breakage. The only finding is a stray unused dependency — trivial to clean up. No correctness, security, or build-blocking issues.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; one P2 (unused `ms` dependency) is non-blocking.

OctoBoooot (approved):
## Review: fix(electron): make macOS package runnable and update icon (#849)

**Verdict:** Approve
A clean Electron packaging + icon PR: preload-path resolution fixes, a channel-allowlisted `removeListener`, a safe post-login redirect helper, and icon asset swaps. No electron security posture is changed by this diff, no open-redirect, and the redirect logic is tested. Code CI green; the only red check is `check-sprint` (governance — doesn't bar the code verdict; `mergeStateStatus` stays blocked until a maintainer clears it).

## What I checked (electron security surfaces)
- **No new IPC surface / posture change.** `main/index.ts` and `createWindow.ts` diffs only fix the preload path (`preload/index` → `preload/index.js`) — a packaging resolution fix, not a capability change. The `webPreferences` (`nodeIntegration`, absence of `contextIsolation`) lines are **not touched** by this PR.
- **Preload `removeListener` is correctly gated.** The new method checks `ALLOWED_RECEIVE_CHANNELS.includes(channel)` before `ipcRenderer.removeListener` (else warns and blocks) — mirroring the existing send/receive allowlist pattern. It exposes IPC *cleanup* required by the authenticated view without widening the channel surface. The preload uses `contextBridge.exposeInMainWorld` with explicit `ALLOWED_SEND_CHANNELS`/`ALLOWED_RECEIVE_CHANNELS` allowlists — the right restrictive model.
- **No open-redirect in `postLoginRedirect.ts`.** For the packaged app (`file:` protocol) it reuses the current file URL, only swapping `search`/clearing `hash` — so it stays on the packaged index, never navigating to another origin (the bug it exists to prevent: a `file:`-loaded app must not redirect to an http origin after login). For web it builds `${origin}${basePath}/${query}` from the app's own origin/basePath — never from user input. Tested both branches (`postLoginRedirect.test.ts`: file-app stays on packaged index; web behavior preserved).
- **Assets** — 11 icon/`.icns`/`.png` files are pure binary swaps (new OCTO artwork), no code.
- CI: Build `success`, secret-scan / osv-scan / osv-scanner / dependency-review / pr-title-lint pass.

### Praise
- The `file:`-vs-`http` split in `postLoginRedirect` is the right fix for a packaged Electron app: a naive `${origin}${basePath}` redirect would break (or leave) a `file://`-loaded app, so branching on `currentUrl.protocol === "file:"` to keep the packaged index and only refresh the query is exactly correct — and it's pinned by a test asserting the packaged file URL survives login. Gating the new `removeListener` on the existing receive-channel allowlist (rather than exposing raw `ipcRenderer.removeListener`) keeps the contextBridge surface tight.

### Out of scope (informational, pre-existing — not introduced here)
- `createWindow.ts` `webPreferences` has `nodeIntegration: true` and no explicit `contextIsolation: true`. That's a broad renderer-privilege posture that sits awkwardly next to the (good) `contextBridge` + channel-allowlist preload. This PR doesn't touch those lines, so it's not a finding against #849 — but the team may want a follow-up to move to `contextIsolation: true` + `nodeIntegration: false` (the preload already speaks contextBridge, so the renderer likely doesn't need direct Node). Flagging once, not blocking.

### Net
Packaging/icon fixes with a correctly-gated IPC cleanup method and a tested, open-redirect-free post-login redirect. The diff doesn't alter the electron security posture; the only pre-existing concern (`nodeIntegration:true`) is out of scope for this PR. Code CI green. **Approve** — `check-sprint` (governance) still needs a maintainer to clear before merge.

yujiawei (approved):
# Code Review — PR #849 (octo-web)

**Verdict: APPROVE.** This is a well-scoped, correct fix that makes the packaged macOS Electron build runnable and refreshes the icon. No correctness or security blockers. A few non-blocking notes below.

## What the PR does
- Fixes post-login navigation for the packaged app (`file://` scheme).
- Corrects the preload script path and makes the Vite build emit relative asset URLs for Electron.
- Hardens `getRandomSid` to use Node's `crypto`.
- Adds a `removeListener` passthrough to the preload IPC bridge.
- Updates the macOS icon set.

## Correctness review (all verified against the diff)

**1. `postLoginRedirect.ts` + `Layout/index.tsx` call site — correct.**
The root bug being fixed: the old code did `` `${window.location.origin}${basePath}/${redirectSearch}` ``, but under `file://` the `origin` is the literal string `"null"`, producing a broken URL and knocking the packaged app off `index.html`. The new helper detects `file:` and instead swaps only `search`/`hash` on the existing URL, preserving the packaged `index.html` path. The web branch keeps the previous behavior byte-for-byte. Both paths are covered by the new unit test.

**2. Preload path `preload/index` → `preload/index.js` — correct.**
Under `file://` loading of the packaged app, the preload path must include the explicit `.js` extension. Applied consistently in both `main/index.ts` and `main/utils/createWindow.ts`.

**3. Vite `base: "./"` gated on `VITE_ELECTRON_BUILD` — correct.**
Electron loads the production entry through `file://`, so assets must be relative to `index.html`; the web build correctly stays rooted at `/`. Gating via the new `build:electron` script is clean.

**4. `getRandomSid` → `node:crypto` `randomBytes` — correct and equivalent.**
`Array.from(randomBytes(4), b => b.toString(36).padStart(2,'0')).join('').slice(0,6)` is functionally identical to the previous `Uint8Array` + web-crypto version (Buffer is a `Uint8Array`, iteration yields the same byte values). `node:crypto` is the right API for the main (Node) process. This function is only used in `main/index.ts` for `sid` query params, so there is no renderer/webcrypto dependency.

## Non-blocking notes

**N1 (P2) — Unused `ms` dependency.**
`ms@^2.1.3` is added to `dependencies` in `apps/web/package.json` (plus a `pnpm-lock.yaml` entry), but `ms` is not imported anywhere in `apps/web/src` or `src-election`, nor referenced in the Vite/electron-builder config. If it is leftover from an earlier iteration, please drop it to avoid dependency cruft; if it is required transitively at build/package time, a one-line comment noting why would help.

**N2 (P2) — `removeListener` proxy identity across the context bridge.**
The new `removeListener` guard is a genuine improvement: the bridge previously did not expose `removeListener` at all, so the existing cleanup in `Pages/Main/vm.ts` (`(window as any).ipc?.removeListener(...)`) was a silent no-op. One caveat worth verifying on-device: functions passed across `contextBridge` are proxied, so the `listener` reference handed to `removeListener` may not be identity-equal to the one `ipcRenderer.on` stored, in which case removal can silently fail. This only affects the `vm.ts` unmount path (a long-lived window that rarely unmounts), so it is not a blocker, but if you observe listener accumulation, track handler→wrapper pairs in the preload script (as the `electronNotification.onClicked/onActionClicked` helpers already do with local `handler` closures).

**N3 (P2, pre-existing, out of scope) — `nodeIntegration: true` in `createWindow.ts`.**
`main/utils/createWindow.ts` sets `nodeIntegration: true` with `contextIsolation` unset. This is a risky configuration in general, but note: (a) this line is not changed by this PR (only the adjacent preload path is), and (b) `createWindow` is not called anywhere in the codebase — it appears to be dead code. Since the file is being touched here anyway, consider deleting the unused `createWindow.ts` in a follow-up, or aligning it with the hardened config used by the real window in `main/index.ts` (`nodeIntegration: false`, `contextIsolation: true`).

## Testing
- New `postLoginRedirect.test.ts` covers both the Electron `file://` and web redirect cases and asserts the exact expected URLs. The assertions match the implementation.
- I could not execute the suite in this environment (monorepo dependencies were not installed), but the test logic and expectations were reviewed directly and are sound.

## Coverage notes
- Binary icon assets (`resources/icons/*`, `logo.png`, `icon.icns`) were not visually inspected; assumed correct per the PR intent.
- Runtime `file://` behavior of the packaged `.app` was not exercised here; the reasoning above is from static analysis of the load paths and Vite base handling.

PR #852, opened 2026-07-16
feat(summary): add unread and invitation attention indicators

### LINKED ISSUE #845: feat(summary): show attention badges and keep unread state in sync
## Summary

Improve Smart Summary attention indicators so users can discover pending invitations and unread results without opening the module first.

## Current behavior

- The navigation badge may not appear until the Smart Summary menu is opened.
- A list card may not refresh when the global attention count changes.
- Opening a summary from a notification deep link can update the navigation count while leaving the card red dot visible.
- Scheduled-summary confirmation invitations are not consistently reflected in the card indicator.

## Expected behavior

- Show a navigation badge for the number of summary tasks that require attention.
- Count each task once when it has a pending invitation, an unread team result, or an unread personal result.
- Show a red dot on every card that requires attention.
- Refresh attention state on application startup, space changes, window focus, and background polling.
- Refresh the visible list when the global attention count changes.
- Mark a result as read only after the corresponding detail content has rendered successfully.
- Keep invitation state independent from read state; viewing a task must not accept or confirm an invitation.
- When opened through `/s/:taskNo`, resolve the numeric task ID before updating the list so the correct card indicator is cleared.
- Do not reorder the visible list immediately after marking an item read; apply the new order on the next list reload.

## Ordering

1. Pending collaboration or scheduled-summary invitations
2. Unread summary results
3. Other summaries

Items within each group should be ordered by their latest meaningful summary activity.

## Acceptance criteria

- An invitee sees the navigation badge before opening Smart Summary.
- Pending scheduled-summary confirmation is included in the badge and card red dot.
- Opening an unread result clears its red dot and decrements the badge when no other attention state remains.
- Opening a notification deep link clears the indicator on the matching numeric task card.
- Accepting, declining, or confirming an invitation refreshes both the navigation badge and card state.
- Team and personal result read states remain independent.

Liuxy20:
## Summary

Add attention indicators for unread Smart Summary results and pending collaboration invitations. The navigation badge now reflects tasks that require attention, while individual cards display a red dot until the relevant result is read or the invitation is handled.

## Related Issue

Closes #845

## Changes

- Display an attention dot on summary cards with unread results or pending invitations.
- Use the server-provided attention count for the Smart Summary navigation badge.
- Refresh attention state after login and Space initialization, on Space changes, window focus, and background polling.
- Refresh the visible list when the global attention count changes.
- Mark team and personal results as read after their content is rendered successfully.
- Keep invitation attention independent from result read state.
- Synchronize read state when summaries are opened through `/s/:taskNo` notification links.
- Resolve the numeric task ID before clearing the matching list-card indicator.
- Keep the visible list position stable after marking a result as read.

- Affected module(s): `packages/dmworksummary`
- New or changed user-visible entry point: No new entry point; existing navigation and cards now expose attention state.
- Shared layer touched: none
- If shared code changed, impact scope: Not applicable
- Duplicate entry point checked: yes

## Testing

- [x] Manually verified

Commands:

- `pnpm exec vitest run src/components/SummaryCard.test.tsx src/api/__tests__/summaryApi.test.ts --pool=forks`
- `git diff --check`

Manual verification covered navigation badge initialization, card updates while the list is open, collaboration and scheduled-summary invitations, notification-link navigation, and detail-to-list read-state synchronization.

## Checklist

- [x] Updated localized UI copy
- [x] Followed commit message conventions

### DIFF
--- a/packages/dmworksummary/src/api/summaryApi.ts
+++ b/packages/dmworksummary/src/api/summaryApi.ts
@@ -291,6 +291,13 @@ export async function getSummaryDetail(taskId: number | string): Promise<Summary
     return get(`/summaries/${encodeURIComponent(String(taskId))}`);
 }
 
+export async function markSummaryRead(
+    taskId: number,
+    cursors: { team_result_id?: number; personal_version_id?: number },
+): Promise<{ is_unread: boolean; has_pending_invitation: boolean; needs_attention: boolean }> {
+    return post(`/summaries/${taskId}/read`, cursors);
+}
+
 export async function deleteSummary(taskId: number): Promise<void> {
     return del(`/summaries/${taskId}`);
 }
--- a/packages/dmworksummary/src/components/SummaryCard.test.tsx
+++ b/packages/dmworksummary/src/components/SummaryCard.test.tsx
@@ -65,6 +65,22 @@ function makeItem(overrides: Record<string, unknown> = {}) {
 
 const noop = () => {};
 
+describe('SummaryCard attention dot', () => {
+    it('needs_attention=true 时显示关注红点', () => {
+        const { container } = render(
+            <SummaryCard task={makeItem({ needs_attention: true }) as any} onClick={noop} onDelete={noop} />,
+        );
+        expect(container.querySelector('.summary-card-attention-dot')).not.toBeNull();
+    });
+
+    it('无需关注时不显示红点', () => {
+        const { container } = render(
+            <SummaryCard task={makeItem({ needs_attention: false }) as any} onClick={noop} onDelete={noop} />,
+        );
+        expect(container.querySelector('.summary-card-attention-dot')).toBeNull();
+    });
+});
+
 describe('SummaryCard isScheduledTask', () => {
     it('schedule_id > 0 时使用定时删除确认文案', () => {
         render(
--- a/packages/dmworksummary/src/components/SummaryCard.tsx
+++ b/packages/dmworksummary/src/components/SummaryCard.tsx
@@ -37,6 +37,7 @@ const SummaryCard: React.FC<SummaryCardProps> = ({ task, onClick, onDelete, onRe
 
     return (
         <div className="summary-card" onClick={() => onClick(task.task_id)}>
+            {task.needs_attention && <span className="summary-card-attention-dot" aria-label={t("summary.list.needsAttention")} />}
             <div className="summary-card-header">
                 <OverflowTooltip className="summary-card-title" title={task.title || task.task_no}>
                     {task.title || task.task_no}
--- a/packages/dmworksummary/src/i18n/en-US.json
+++ b/packages/dmworksummary/src/i18n/en-US.json
@@ -122,6 +122,7 @@
     "title": "AI Summary"
   },
   "list": {
+    "needsAttention": "Needs attention",
     "title": "AI Summary",
     "allStatus": "All statuses",
     "createTooltip": "New summary",
--- a/packages/dmworksummary/src/i18n/zh-CN.json
+++ b/packages/dmworksummary/src/i18n/zh-CN.json
@@ -122,6 +122,7 @@
     "title": "智能总结"
   },
   "list": {
+    "needsAttention": "需要关注",
     "title": "智能总结",
     "allStatus": "全部状态",
     "createTooltip": "新建总结",
--- a/packages/dmworksummary/src/index.css
+++ b/packages/dmworksummary/src/index.css
@@ -132,6 +132,7 @@
 
 /* ========== Summary Card ========== */
 .summary-card {
+    position: relative;
     border: 1px solid var(--semi-color-border);
     border-radius: 12px;
     padding: 18px;
@@ -139,6 +140,20 @@
     transition: all 0.2s;
     background: transparent;
     box-shadow: none;
+}
+
+.summary-card-attention-dot {
+    position: absolute;
+    top: var(--wk-sp-2);
+    right: var(--wk-sp-2);
+    width: var(--wk-sp-2);
+    height: var(--wk-sp-2);
+    border-radius: 50%;
+    background: var(--semi-color-danger);
+}
+
+.summary-card .summary-card-header {
+    padding-right: var(--wk-sp-3);
 }
 
 .summary-card:hover {
--- a/packages/dmworksummary/src/module.tsx
+++ b/packages/dmworksummary/src/module.tsx
@@ -7,7 +7,7 @@ import SummaryCreatePage from "./pages/SummaryCreatePage";
 import SummaryDetailPage from "./pages/SummaryDetailPage";
 import SummaryConfirmPage from "./pages/SummaryConfirmPage";
 import ScheduleListPage from "./pages/ScheduleListPage";
-import { getChatCandidates } from "./api/summaryApi";
+import { getChatCandidates, listSummaries } from "./api/summaryApi";
 import { notifyChatSummaryCreated } from "./utils/chatSummaryActions";
 import { isSupportedChannelType } from "./utils/channelType";
 import ChatSummaryStarButton from "./components/ChatSummaryStarButton";
@@ -18,6 +18,46 @@ import zhCN from "./i18n/zh-CN.json";
 import "./index.css";
 
 let _spaceChangedHandler: (() => void) | null = null;
+let _attentionRefreshTimer: ReturnType<typeof setInterval> | null = null;
+let _attentionFocusHandler: (() => void) | null = null;
+let _attentionRefreshRequestHandler: (() => void) | null = null;
+let _attentionBootstrapTimer: ReturnType<typeof setTimeout> | null = null;
+let _attentionRequest: Promise<boolean> | null = null;
+
+async function refreshSummaryAttentionCount(): Promise<boolean> {
+    if (!WKApp.loginInfo.isLogined() || !WKApp.shared.currentSpaceId) return false;
+    if (_attentionRequest) return _attentionRequest;
+    _attentionRequest = (async () => {
+      try {
+        const response = await listSummaries({ page: 1, page_size: 1 });
+        WKApp.mittBus.emit("summary-badge-update" as any, {
+            count: response.attention_count ?? 0,
+        });
+        WKApp.mittBus.emit("summary-attention-count-refreshed" as any, {
+            count: response.attention_count ?? 0,
+        });
+        return true;
+      } catch {
+        // Navigation badge is best-effort and must not block application startup.
+        return false;
+      } finally {
+        _attentionRequest = null;
+      }
+    })();
+    return _attentionRequest;
+}
+
+function bootstrapSummaryAttentionCount(attempt = 0) {
+    void refreshSummaryAttentionCount().then((loaded) => {
+        if (loaded || attempt >= 50) return;
+        // Module initialization can precede login/Space restoration on a cold
+        // page load. Retry quickly until both are ready instead of waiting for
+        // the five-second background poll.
+        _attentionBootstrapTimer = setTimeout(() => {
+            bootstrapSummaryAttentionCount(attempt + 1);
+        }, 100);
+    });
+}
 
 /**
  * NavRail 顶层菜单图标（智能总结）。与 dmworktodo / dmworkappbot 的菜单图标同构：
@@ -107,8 +147,23 @@ export class SummaryModule implements IModule {
 
         _spaceChangedHandler = () => {
             WKApp.mittBus.emit('summary-space-changed');
+            void refreshSummaryAttentionCount();
         };
         WKApp.mittBus.on('space-changed', _spaceChangedHandler);
+
+        // SummaryListPage is mounted only after the menu is opened. Keep the
+        // global navigation badge fresh independently so invitations are
+        // visible before the invitee enters Smart Summary.
+        bootstrapSummaryAttentionCount();
+        _attentionRefreshTimer = setInterval(() => {
+            if (document.visibilityState === "visible") {
+                void refreshSummaryAttentionCount();
+            }
+        }, 5000);
+        _attentionFocusHandler = () => void refreshSummaryAttentionCount();
+        window.addEventListener("focus", _attentionFocusHandler);
+        _attentionRefreshRequestHandler = () => void refreshSummaryAttentionCount();
+        WKApp.mittBus.on("summary-attention-refresh-requested" as any, _attentionRefreshRequestHandler);
 
         WKApp.searchChatCandidates = async (params) => {
             return getChatCandidates(params);
@@ -145,6 +200,22 @@ if (import.meta.hot) {
         if (_spaceChangedHandler) {
             WKApp.mittBus.off('space-changed', _spaceChangedHandler);
             _spaceChangedHandler = null;
+        }
+        if (_attentionRefreshTimer) {
+            clearInterval(_attentionRefreshTimer);
+            _attentionRefreshTimer = null;
+        }
+        if (_attentionFocusHandler) {
+            window.removeEventListener("focus", _attentionFocusHandler);
+            _attentionFocusHandler = null;
+        }
+        if (_attentionRefreshRequestHandler) {
+            WKApp.mittBus.off("summary-attention-refresh-requested" as any, _attentionRefreshRequestHandler);
+            _attentionRefreshRequestHandler = null;
+        }
+        if (_attentionBootstrapTimer) {
+            clearTimeout(_attentionBootstrapTimer);
+            _attentionBootstrapTimer = null;
         }
         _globalSummaryModalRoot?.unmount();
         _globalSummaryModalRoot = null;
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -491,8 +491,16 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             if (detail.status === TaskStatus.COMPLETED && detail.result_id) {
                 const markRead = api.markSummaryRead;
                 if (markRead) void markRead(detail.task_id, { team_result_id: detail.result_id }).then((attention) => {
-                    if (this.taskId === requestTaskId && !attention.needs_attention) {
-                        window.dispatchEvent(new CustomEvent("summary-read", { detail: { taskId: detail.task_id } }));
+                    // Guard with the request sequence/lookup key rather than
+                    // comparing numeric detail.task_id with a task_no deep-link.
+                    if (this.scheduleLoadSeq === seq && this.detailLookupId === requestTaskId) {
+                        window.dispatchEvent(new CustomEvent("summary-read", {
+                            detail: {
+                                taskId: detail.task_id,
+                                isUnread: attention.is_unread,
+                                needsAttention: attention.needs_attention,
+                            },
+                        }));
                     }
                 }).catch(() => { /* keep unread on failure */ });
             }
@@ -608,11 +616,14 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             if (result.current_version_id && result.content?.trim()) {
                 const markRead = api.markSummaryRead;
                 if (markRead) void markRead(requestTaskId, { personal_version_id: result.current_version_id }).then((attention) => {
-                    if (this.taskId === requestTaskId && !attention.needs_attention) {
-                        const numericTaskId = this.state.detail?.task_id;
-                        if (numericTaskId) {
-                            window.dispatchEvent(new CustomEvent("summary-read", { detail: { taskId: numericTaskId } }));
-                        }
+                    if (this.scheduleLoadSeq === reqSeq && (this.taskId == null || this.taskId === requestTaskId)) {
+                        window.dispatchEvent(new CustomEvent("summary-read", {
+                            detail: {
+                                taskId: requestTaskId,
+                                isUnread: attention.is_unread,
+                                needsAttention: attention.needs_attention,
+                            },
+                        }));
                     }
                 }).catch(() => { /* keep unread on failure */ });
             }
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -488,6 +488,14 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                 lastKnownStatus: detail.status,
                 workflowGateContent: false,
             });
+            if (detail.status === TaskStatus.COMPLETED && detail.result_id) {
+                const markRead = api.markSummaryRead;
+                if (markRead) void markRead(detail.task_id, { team_result_id: detail.result_id }).then((attention) => {
+                    if (this.taskId === requestTaskId && !attention.needs_attention) {
+                        window.dispatchEvent(new CustomEvent("summary-read", { detail: { taskId: detail.task_id } }));
+                    }
+                }).catch(() => { /* keep unread on failure */ });
+            }
             if (detail.status === TaskStatus.COMPLETED && detail.result) {
                 this.loadVersions(detail.task_id);
             }
@@ -597,6 +605,17 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                 });
             }
             this.startPersonalPoll(result.worker_status);
+            if (result.current_version_id && result.content?.trim()) {
+                const markRead = api.markSummaryRead;
+                if (markRead) void markRead(requestTaskId, { personal_version_id: result.current_version_id }).then((attention) => {
+                    if (this.taskId === requestTaskId && !attention.needs_attention) {
+                        const numericTaskId = this.state.detail?.task_id;
+                        if (numericTaskId) {
+                            window.dispatchEvent(new CustomEvent("summary-read", { detail: { taskId: numericTaskId } }));
+                        }
+                    }
+                }).catch(() => { /* keep unread on failure */ });
+            }
             if (result.content?.trim()) {
                 this.loadPersonalVersions(requestTaskId);
             } else if (this.taskId == null || this.taskId === requestTaskId) {
@@ -3109,6 +3128,7 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
             // 迟到（已切 task）：不回显新 task（confirmingSchedule 由 finally 复位）。
             if (this.taskId !== requestTaskId) return;
             Toast.success(t("summary.detail.scheduleConfirmed"));
+            WKApp.mittBus.emit("summary-attention-refresh-requested" as any);
             // 复用现有加载路径刷新（不新增任何出站推送）：重拉 schedule 让按钮消失。
             this.loadSchedule(scheduleItem.schedule_id);
         } catch (err: any) {
--- a/packages/dmworksummary/src/pages/SummaryListPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryListPage.tsx
@@ -62,10 +62,32 @@ export default class SummaryListPage extends Component<{}, SummaryListPageState>
     private searchTimer: ReturnType<typeof setTimeout> | null = null;
     private batchPollTimer: ReturnType<typeof setInterval> | null = null;
     private isBatchPolling = false;
+    private attentionCount = 0;
+    private attentionRefreshLoading = false;
 
     private handleSpaceChanged_ = () => this.loadData();
 
     private handleTaskRegenerated_ = () => this.loadData();
+
+    private handleAttentionCountRefreshed_ = ({ count }: { count: number }) => {
+        const nextCount = count ?? 0;
+        if (nextCount === this.attentionCount || this.attentionRefreshLoading) return;
+        this.attentionRefreshLoading = true;
+        void this.loadData().finally(() => {
+            this.attentionRefreshLoading = false;
+        });
+    };
+
+    private handleSummaryRead_ = (event: Event) => {
+        const taskId = (event as CustomEvent<{ taskId: number }>).detail?.taskId;
+        if (!taskId) return;
+        this.setState(({ items }) => ({
+            items: items.map(item => item.task_id === taskId
+                ? { ...item, is_unread: false, needs_attention: Boolean(item.has_pending_invitation) }
+                : item),
+        }));
+        this.emitBadgeUpdate();
+    };
 
     private handleNavMenuActivated_ = ({ menuId }: { menuId: string }) => {
         if (menuId === "summary") {
@@ -77,7 +99,9 @@ export default class SummaryListPage extends Component<{}, SummaryListPageState>
         this.loadData();
         WKApp.mittBus.on("summary-space-changed", this.handleSpaceChanged_);
         WKApp.mittBus.on("wk:nav-menu-activated", this.handleNavMenuActivated_);
+        WKApp.mittBus.on("summary-attention-count-refreshed" as any, this.handleAttentionCountRefreshed_);
         window.addEventListener("summary-task-regenerated", this.handleTaskRegenerated_);
+        window.addEventListener("summary-read", this.handleSummaryRead_);
     }
 
     componentWillUnmount() {
@@ -86,7 +110,9 @@ export default class SummaryListPage extends Component<{}, SummaryListPageState>
         this.stopBatchPoll();
         WKApp.mittBus.off("summary-space-changed", this.handleSpaceChanged_);
         WKApp.mittBus.off("wk:nav-menu-activated", this.handleNavMenuActivated_);
+        WKApp.mittBus.off("summary-attention-count-refreshed" as any, this.handleAttentionCountRefreshed_);
         window.removeEventListener("summary-task-regenerated", this.handleTaskRegenerated_);
+        window.removeEventListener("summary-read", this.handleSummaryRead_);
     }
 
     async loadData() {
@@ -100,9 +126,10 @@ export default class SummaryListPage extends Component<{}, SummaryListPageState>
                 keyword: keyword || undefined,
             };
             const resp = await api.listSummaries(params);
+            this.attentionCount = resp.attention_count ?? 0;
             this.setState({ items: resp.items, total: resp.total, loading: false }, () => {
                 this.maybeStartBatchPoll();
-                this.emitBadgeUpdate();
+                this.emitBadgeUpdate(resp.attention_count);
             });
         } catch (err: any) {
             this.setState({ error: err.message || t("summary.common.loadingFailed"), loading: false });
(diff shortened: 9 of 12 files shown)


OctoBoooot (dismissed):
## Review: feat(summary): add unread and invitation attention indicators (#852)

**Verdict:** Approve
A clean, well-scoped attention-indicator feature: a red dot on summary cards + a nav badge for unread results / pending invitations, with mark-as-read fired after successful render. The correctness-sensitive parts — stale-request guarding on mark-read and a fail-safe error path — are handled right. No injection, no i18n regression, CI green.

## What I checked (the risk axes for a read-state feature)
- **Mark-read is a post-render side-effect, correctly guarded.** In `SummaryDetailPage`, `markSummaryRead` is called only after a COMPLETED result with content (`detail.status === COMPLETED && detail.result_id`; `result.current_version_id && result.content?.trim()`) as a `void` call — it does **not** gate the render. On resolve it guards `if (this.taskId === requestTaskId && !attention.needs_attention)` before dispatching `summary-read` — so a mark-read that lands after the user navigated to a different summary can't clear the wrong card (correct stale-request discipline). `.catch(() => { /* keep unread on failure */ })` keeps the item unread on error rather than falsely clearing the dot — fail-safe.
- **No injection / XSS.** `markSummaryRead(taskId: number, ...)` posts to `/summaries/${taskId}/read` with a numeric path param + cursors in the JSON body — no URL injection. The attention dot is a pure CSS `<span className="summary-card-attention-dot" aria-label={t("summary.list.needsAttention")} />` — no content rendered, and it carries an aria-label (a11y, not color-only).
- **No i18n regression.** Zero removed keys in either locale (checked — the removed-in-use class I've seen elsewhere this session isn't present here). The one new key `needsAttention` is added to **both** en-US and zh-CN (verified by exact key diff; an initial coarse grep suggested a 3-vs-1 gap but that was English value-text matching the word "attention", not extra keys), and Build's i18n check passes — confirming no locale mismatch.
- **Parity.** The change is additive attention state layered onto the existing list/detail render; the detail render path is unchanged (mark-read is appended, not substituted). No capability dropped.
- CI: Build `success` (includes the i18n check), secret-scan / osv-scan / dependency-review / pr-title-lint pass; check-sprint skipped (governance).

### Praise
- The mark-read path gets the two details right that this feature class usually gets wrong: (1) it only marks read after the content actually rendered (COMPLETED + non-empty), so opening a still-generating summary doesn't prematurely clear attention; and (2) the `this.taskId === requestTaskId` guard on the async resolve prevents a late mark-read from clearing the indicator of whatever summary the user has since navigated to. Keeping invitation attention independent of result-read state (per the PR body) and failing closed to "unread" on a mark-read error are the right conservative choices.

### Net
Correct read-state/attention feature: mark-read is post-render, stale-guarded, and fail-safe; the indicator is an injection-free CSS dot with an aria-label; i18n is parity-clean with no removed keys; CI green. No dropped capability. **Approve** — `check-sprint` (governance) still needs a maintainer to clear before merge.

Jerry-Xin (dismissed):
Reviewed at head `ebd3593`. LGTM — approving. Clean, well-scoped addition of unread/invitation attention indicators to Smart Summary.

### ✅ Highlights
- **`summaryApi.ts` — `markSummaryRead`**: goes through the same `post()` interceptor as the rest of the API, so it carries `token` + `X-Space-Id` (server scopes the read state to the authenticated user in the current Space). `taskId` is typed `number`, so the un-`encodeURIComponent`'d path segment is not an injection vector. No cross-user data surfaced — response is only the caller's own `is_unread`/`has_pending_invitation`/`needs_attention`.
- **`module.tsx` badge poller**: the 5s `setInterval`, `focus` handler, `summary-attention-refresh-requested` bus listener, and bootstrap `setTimeout` are all cleared/removed in the `import.meta.hot` dispose block (`clearInterval`/`removeEventListener`/`clearTimeout`), mirroring the existing `_spaceChangedHandler` teardown. `refreshSummaryAttentionCount` is guarded against overlap via the `_attentionRequest` in-flight promise and gated on `isLogined() && currentSpaceId`; poll only fires when `document.visibilityState === "visible"`. The interval intentionally lives for the app lifetime (module singleton, not a component), which is correct here.
- **Indicators are safe to render**: the attention dot is an empty `<span>` badge with a `t()`-based `aria-label`; no server/user string is interpolated into markup, and no `dangerouslySetInnerHTML`/`innerHTML` anywhere in the changed TSX. Counts flow as numbers via the mitt bus.
- **i18n parity**: `list.needsAttention` added to BOTH `en-US.json` and `zh-CN.json` under the same `list` object; the only new `t()` key in the diff (`summary.list.needsAttention`) resolves in both locales. No hardcoded user-facing Chinese in the added TSX. Build (incl. i18n check) is **green**.
- **`SummaryListPage`**: `summary-read` / `summary-attention-count-refreshed` listeners are symmetrically added in `componentDidMount` and removed in `componentWillUnmount`; `emitBadgeUpdate` now prefers the server-provided `attention_count` and falls back cleanly.
- **Types + test**: `types/summary.ts` new optional fields align with the API response shape; `SummaryCard.test.tsx` covers both the `needs_attention=true` (dot present) and `false` (dot absent) branches.

### 💬 Non-blocking
- The two `markSummaryRead` calls in `SummaryDetailPage` correctly re-check `this.taskId === requestTaskId` before dispatching `summary-read`, avoiding stale-task races — nice.

No blockers.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #852 (octo-web)

## Summary

This PR adds unread and invitation attention indicators to the Smart Summary module: a red dot on summary cards with unread results or pending invitations, a server-driven attention count for the navigation badge, and mark-read calls that clear the indicators once content is rendered. The implementation spans the module lifecycle (bootstrap retry, 5-second polling, window-focus refresh, space-change refresh), the list page (event-driven card updates, badge emission), the detail page (team and personal mark-read on content render, schedule-confirmation refresh), and new API/type definitions.

The architecture is well-structured — attention state flows from server → module poll → mittBus events → list page re-render, and mark-read calls fire after content renders with proper race guards. One P2 and two nits below.

## Verification

Static analysis only at head `ebd3593df0658c973bdd83cf07d0997ab9d031cc`; build and tests not executed in this environment.

- ✅ **Event lifecycle symmetry** — all four refresh triggers (bootstrap, interval, focus, space-changed) plus `summary-attention-refresh-requested` are registered in `init()` and cleaned up in `import.meta.hot.dispose()`. Timer and listener references are nulled after cleanup.
- ✅ **Request dedup** — `refreshSummaryAttentionCount()` deduplicates concurrent API calls via `_attentionRequest` promise, cleared in `finally`. Prevents request storms from overlapping triggers.
- ✅ **mark-read race safety** — both team and personal mark-read paths verify `this.taskId` matches the request before dispatching `summary-read`. `.catch(() => {})` preserves unread state on failure.
- ✅ **Type contracts** — `markSummaryRead` expects `taskId: number`; all call sites pass numeric IDs (team: `detail.task_id`; personal: `requestTaskId` from `loadPersonalResult` where the override is `detail.task_id`).
- ✅ **i18n parity** — `needsAttention` key added to both `en-US.json` and `zh-CN.json`.
- ✅ **Test coverage** — `SummaryCard.test.tsx` covers attention dot rendering for `needs_attention: true/false`.

## Findings

One P2 item and two nits. No P0/P1 blockers.

### P2 — Deep-link task ID type mismatch in team-result mark-read (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:494`)

When the detail page opens via a `/s/:taskNo` notification deep-link, `requestTaskId` (assigned from `lookupId` at `:444`) holds the original **string** `task_no`. After `getSummaryDetail` returns, `this.taskId` resolves to the **numeric** `detail.task_id`. The guard `this.taskId === requestTaskId` at `:494` therefore evaluates `number === string` → always `false`, so the `summary-read` CustomEvent never fires from the team-result mark-read path on deep-link navigation.

The personal-result mark-read at `:611` avoids this because `loadPersonalResult` receives `detail.task_id` (numeric) as its `taskIdOverride`. The two paths are asymmetric — C1 dual-path parity miss.

**Impact**: after navigating via notification link to a **team** summary, the server-side read-mark succeeds but the list card's attention dot stays stale until the next 5-second background poll re-fetches the list. Self-correcting, but a user-visible UX gap on a standard production path.

```ts
// Fix: use detail.task_id (already in scope) instead of requestTaskId
if (this.taskId === detail.task_id && !attention.needs_attention) {
    window.dispatchEvent(new CustomEvent("summary-read", { detail: { taskId: detail.task_id } }));
}
```

### Nit — Loading spinner flash on attention-count-driven refresh (`packages/dmworksummary/src/pages/SummaryListPage.tsx:78`)

`handleAttentionCountRefreshed_` calls `loadData()`, which sets `loading: true` — hiding the list and showing a `<Spin>` spinner until the API responds. When the attention count changes, this produces a brief flash. Consider refreshing without toggling the `loading` flag (e.g. a silent re-fetch that swaps items in-place).

### Nit — Unnecessary `if (markRead)` guard (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:493`)

`api.markSummaryRead` is a named export from `summaryApi.ts` and is always defined at import time. The `const markRead = api.markSummaryRead; if (markRead)` pattern at both `:493` and `:610` adds a dead conditional. Safe to call directly.

- **Badge semantic migration** — old badge counted `WAITING_CONFIRM` tasks via a filtered query; new badge uses server-computed `attention_count`. The fallback `emitBadgeUpdate()` (no count arg) now queries without status filter but still reads `attention_count`. Consistent across all call sites.
- **Bootstrap retry bounds** — 50 attempts × 100ms = 5-second ceiling. Dedup prevents duplicate API calls when the interval timer overlaps with bootstrap retries.
- **`handleSummaryRead_` optimistic update** — correctly sets `needs_attention: Boolean(item.has_pending_invitation)` to preserve invitation attention when clearing the unread flag.
- **`confirmSchedule` attention refresh** — emits `summary-attention-refresh-requested` after confirmation, triggering a fresh attention count poll. Correct for clearing pending invitation attention.
- **CSS attention dot** — `position: relative` on `.summary-card`, `position: absolute` on `.summary-card-attention-dot`, `padding-right` on header to prevent overlap. Clean.

## Data Flow Trace

|---|---|---|
| `mittBus("summary-badge-update").count` | `refreshSummaryAttentionCount()` → `listSummaries().attention_count` | ✅ |
| `mittBus("summary-attention-count-refreshed").count` | Same as above | ✅ |
| `SummaryListPage.attentionCount` | `loadData()` → `resp.attention_count` | ✅ |
| `SummaryCard.needs_attention` | Server `SummaryListItem.needs_attention` field | ✅ |
| `window("summary-read").detail.taskId` | DetailPage mark-read `.then()` → `detail.task_id` or `state.detail?.task_id` | ⚠️ team path broken on deep-link (P2) |
| `markSummaryRead` taskId argument | Team: `detail.task_id`; Personal: `requestTaskId` (number) | ✅ always numeric |

## Blindspot Checklist

- **C1 (dual-path parity)**: HIT — team-result vs personal-result mark-read paths have asymmetric `requestTaskId` type handling on deep-links (documented in P2 above).
- **C2 (control-flow ordering)**: CLEAR — `refreshSummaryAttentionCount` dedup prevents re-entrant calls.
- **C3–C6**: N/A — no authorization, build, or governance changes.

N/A — automated review review.

[Octo-Q] verdict: APPROVE — one P2 (deep-link type mismatch, self-correcting via 5s poll) and two nits; no P0/P1 blockers.

yujiawei (changes requested):
# Code Review — PR #852 (octo-web)

Reviewed at head SHA `ebd3593df0658c973bdd83cf07d0997ab9d031cc` against `main` (merge-base `1bb2cf33`). Scope: `packages/dmworksummary` unread / invitation attention indicators.

Overall this is a well-structured change with good race-condition hygiene (the `requestTaskId`/`seq` guards are carefully applied). There is one functional issue that should be fixed before merge because it directly undermines a goal stated in the PR description ("keep the visible list position stable after marking a result as read"), plus several smaller robustness items.

## Blocking

### P1 — Reading a summary triggers a spurious full-list reload ~5s later
`SummaryListPage.tsx`

`handleSummaryRead_` (L81-90) optimistically patches `items` and calls `emitBadgeUpdate()`, but it never updates the instance field `this.attentionCount`. Meanwhile the 5s background poller in `module.tsx` (L158) calls `refreshSummaryAttentionCount`, which emits `summary-attention-count-refreshed` with the now-decremented server count. `handleAttentionCountRefreshed_` (L72-79) then compares `nextCount !== this.attentionCount` — which is true, because the local field is stale — and fires a full `loadData()`.

`loadData()` (L119) begins with `this.setState({ loading: true })`, and `render()` hides the list entirely while `loading` is true (`{loading && <Spin/>}` at L349-353, `{!loading && items.length > 0 && ...}` at L367). Net effect: within ~5 seconds of reading *any* completed summary, the list flashes to a full-page spinner and re-renders from scratch, losing scroll position — the exact regression the PR claims to prevent.

Suggested fix: update `this.attentionCount` inside `handleSummaryRead_` (and/or have `emitBadgeUpdate()` keep it in sync) so the optimistic local state and the next poll agree, avoiding the redundant reload.

## Non-blocking (recommended)

### P2 — Bootstrap retry can hammer the API on a startup error
`module.tsx` L27-59

`refreshSummaryAttentionCount` intentionally swallows API/network failures and returns `false` (L38-42). `bootstrapSummaryAttentionCount` treats `false` as "not ready yet" and retries every 100ms up to 50 times (L50-58). A backend 5xx or connectivity blip during cold start is indistinguishable from "not logged in yet", so it becomes up to 50 rapid-fire requests. Consider distinguishing "not ready" (not logged in / no space) from "request failed", and only retrying the former.

### P2 — `_attentionRequest` de-dupe is not space-scoped
`module.tsx` L27-47

`_attentionRequest` is a single module-level in-flight promise. If `space-changed` fires while a request for the previous space is in flight, the new refresh returns the old promise (L29), so it can emit the previous space's `attention_count`. Self-corrects on the next poll/focus, but briefly shows a wrong-space badge. Keying the in-flight guard by `currentSpaceId` would remove the window.

### P2 — Read state can stay locally stale when an invitation is still pending
`SummaryDetailPage.tsx` L494 / L611

The `summary-read` event is dispatched only when `!attention.needs_attention`. For a task that is both unread and has a pending invitation, the server clears `is_unread` but keeps `needs_attention=true`, so no event is emitted and the list keeps `is_unread=true` locally. The dot still shows (correctly, due to the invitation), so the visible impact is small, but `handleSummaryRead_` already models exactly this case (`needs_attention: Boolean(item.has_pending_invitation)`) and would benefit from receiving the event regardless of `needs_attention`.

### P2 — Redundant `markSummaryRead` POSTs
`SummaryDetailPage.tsx` L490-497 / L607-615

`markSummaryRead` is re-issued on each `loadDetail`/`loadPersonalResult` (stream-done, terminal status transition, re-entry). It is guarded and presumably idempotent server-side, so this is not per-tick spam, but a small "already marked this result id" guard would avoid the repeat writes.

### P2 — Init is not idempotent
`module.tsx` L147-166

The interval, `focus` listener, and mittBus listener are cleaned up only in the `import.meta.hot` dispose block. If `SummaryModule` init ever runs more than once in a live runtime, the previous timer/listener references are overwritten and leak. An idempotency guard (clear-before-set) at the top of init would make this safe independent of the host lifecycle.

### P3 (nit) — Attention dot accessibility
`SummaryCard.tsx` L40

The dot `<span>` has an `aria-label` but no semantic role; many screen readers skip label-only spans. Consider `role="img"` (or `role="status"`) so the "needs attention" state is actually announced.

## Coverage notes (not verified from the frontend diff alone)
- Backend semantics of `attention_count` / `is_unread` / `has_pending_invitation` and whether `/summaries/{id}/read` is idempotent.
- Whether `listSummaries` returns a global `attention_count` unaffected by `status`/`keyword` filter params (the badge relies on it being global).
- Whether `SummaryModule` init is guaranteed to run exactly once per page lifecycle (determines real-world severity of the init-idempotency item).
- New `SummaryListItem` fields `current_result_id`, `current_personal_version_id`, `activity_at`, and response fields `unread_count`/`pending_invitation_count` are added to types but not consumed in this diff — confirm they are intended for a follow-up.

Tests: unit coverage for the attention dot is a good addition; the desync interaction above is not covered and would be worth a regression test.

yujiawei:
### Additional finding (P2) — team-result mark-read event never fires on `/s/:taskNo` deep-links

`SummaryDetailPage.tsx` L494 (team path)

When the detail page is opened via a notification deep-link (`/s/:taskNo`), `props.taskId` holds the **string** `task_no`, so `requestTaskId` (= `lookupId`, L444) is a string. After `getSummaryDetail` resolves, the `taskId` getter (L409-412) returns the **numeric** `detail.task_id`. The guard at L494 is therefore `number === string` → always `false`, so the `summary-read` CustomEvent never fires from the team mark-read path on this route. The server-side read is still recorded, but the list card's dot stays stale until the next background poll re-fetches.

The personal path (L611) is not affected because `loadPersonalResult` is invoked with `detail.task_id` (numeric) as its override, so its `requestTaskId` matches the getter.

Suggested fix — compare against the numeric id that is already in scope:

```ts
if (this.taskId === detail.task_id && !attention.needs_attention) {
    window.dispatchEvent(new CustomEvent("summary-read", { detail: { taskId: detail.task_id } }));
}
```

This is the same read → list-sync flow as the blocking item in the main review, so both are worth addressing together in the next revision.

OctoBoooot (approved):
## Re-review (DELTA) #852 @ `07c2eeae` — both peer-caught bugs I missed (attention-count desync flicker + deep-link team-path taskId type mismatch) are fixed, with regression tests → APPROVE

**Verdict:** Approve
This commit fixes both findings peers caught that I missed at the prior head: yujiawei's attention-count desync (read → 5s poll → full-list reload flicker) and mochashanyao's deep-link team-path taskId type mismatch (team mark-read event never firing on a `/s/:taskNo` deep-link). Byte-verified both fixes are correct and regression-tested. CI green.

## What I verified (both fixes)
- **Attention-count desync (yujiawei's 🔴) — fixed.** `handleSummaryRead_` (SummaryListPage.tsx) now decrements the local count on a genuine attention true→false transition: `if (current?.needs_attention && !nextNeedsAttention) { this.attentionCount = Math.max(0, this.attentionCount - 1) }`, where `nextNeedsAttention` correctly folds in the pending-invitation case (`detail.needsAttention ?? Boolean(current?.has_pending_invitation)`). So after a read, `this.attentionCount` stays aligned with the server, and the 5-second poll's `handleAttentionCountRefreshed_` guard (`nextCount === this.attentionCount → return`) no longer diverges and force-runs `loadData()`. The list no longer refetches/flickers on every read. `Math.max(0, ...)` clamps against underflow.
- **Deep-link team-path taskId type mismatch (mochashanyao's 🟡) — fixed.** The team mark-read guard changed from `this.taskId === requestTaskId` (numeric `detail.task_id` vs string `task_no` on a deep-link → always false → event never fired) to `this.scheduleLoadSeq === seq && this.detailLookupId === requestTaskId` — comparing `detailLookupId` against `requestTaskId` (the same raw lookup identifier, consistently typed), with a seq guard. The comment names the avoided pitfall ("comparing numeric detail.task_id with a task_no deep-link"). Both team and personal paths now use the consistent `detailLookupId === requestTaskId` comparison, so a team summary opened via `/s/:taskNo` now correctly dispatches `summary-read` and clears the list dot.

## Tests assert the fixes
- `SummaryListPage.attention.test.tsx` (+74): "keeps the optimistic count aligned so the next poll does not reload the list" — dispatches `summary-read`, then `handleAttentionCountRefreshed_({count:2})`, asserts `attentionCount === 2`, item `{is_unread:false, needs_attention:false}`, and crucially **`expect(loadData).not.toHaveBeenCalled()`** (the anti-flicker assertion). A second case verifies the server-response path doesn't wrongly clear other unread cursors / pending attention.
- `SummaryDetailPage.schedule.test.tsx` (+34) covers the detail-page mark-read path.
- CI: Build `success`, secret-scan / osv-scan / dependency-review pass; check-sprint skipped (governance).

## What still stands (from my prior review)
The correct parts remain: mark-read is post-render + fail-safe (`.catch` keeps unread), the attention dot is an injection-free CSS span with aria-label, no i18n regression. The two state-sync bugs are now closed.

## Correcting my record
My prior APPROVE missed both of these — I verified the mark-read stale-guard and the primary render/i18n axes, but under-scanned the secondary state-sync interactions (local-count-vs-poll reconciliation, and the string-vs-numeric taskId across the team *and* personal branches). yujiawei + mochashanyao caught them; both are now fixed. Lesson logged: on a state-management feature, trace the cache-vs-server-poll reconciliation and every event-type path across branches, not just the representative one.

## Net
Both peer-caught state-sync bugs are correctly fixed — the attention count stays aligned after read (no poll-triggered flicker), and the deep-link team-path mark-read now fires (consistent `detailLookupId` comparison) — with behavior-level regression tests (including the explicit `loadData not called` anti-flicker assertion). CI green. Flipping my prior APPROVE → re-registering **APPROVE** @ `07c2eeae`. Credit @yujiawei (count desync) + @mochashanyao (deep-link type). Peers dismissed by this commit should re-anchor; yujiawei's CR at the prior head is fixed at this one. `check-sprint` (governance) still needs a maintainer.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #852 (octo-web)

## Summary

This PR adds unread and invitation attention indicators to the Smart Summary module. It introduces a `markSummaryRead` API endpoint to track read cursors (team `result_id` / personal `version_id`), a global 5-second attention-count poll in `module.tsx` to keep the NavRail badge fresh even before the list page is mounted, an attention dot on `SummaryCard` for items needing attention, and optimistic state synchronization between the detail page's mark-read response and the list page's local state. The second commit (`07c2eeae3`) adds the mark-read-after-view logic and tests. Overall the implementation is well-structured with proper race-condition guards (`scheduleLoadSeq` / `detailLookupId`) and HMR cleanup. Three non-blocking P2 items identified below.

## Verification

- ✅ **Data flow: attention_count** — traced from `listSummaries` API response → `refreshSummaryAttentionCount()` in `module.tsx:35` → `summary-badge-update` + `summary-attention-count-refreshed` events → `SummaryListPage.loadData()` at `:159` sets `this.attentionCount = resp.attention_count`. All consumers receive the server value; `?? 0` null-coalescing is consistently applied.
- ✅ **Data flow: mark-read response** — traced from `markSummaryRead` API (`summaryApi.ts:296`) → `.then()` in `SummaryDetailPage.tsx:493` (team) and `:618` (personal) → `CustomEvent("summary-read")` with `{ taskId, isUnread, needsAttention }` → `SummaryListPage.handleSummaryRead_` at `:83` patches item state and optimistically decrements `attentionCount`. Server response fields are consumed correctly at each hop.
- ✅ **Race-condition guards** — mark-read `.then()` callbacks use `this.scheduleLoadSeq === seq && this.detailLookupId === requestTaskId` (team) and `this.scheduleLoadSeq === reqSeq && (this.taskId == null || this.taskId === requestTaskId)` (personal). These match the established FE-1 pattern from `loadSchedule` / `loadPersonalResult`.
- ✅ **Event listener lifecycle** — all `mittBus.on` / `window.addEventListener` calls in `module.tsx` have matching cleanup in `import.meta.hot.dispose()` and in `SummaryListPage.componentWillUnmount()`. Timer cleanup (`clearInterval`, `clearTimeout`) is present for all four new timer refs.
- ✅ **Bootstrap retry** — `bootstrapSummaryAttentionCount` retries up to 50 times at 100ms (5s max) until login/space are ready, then stops. Does not leak timers. The 5s interval timer provides ongoing refresh independently.
- ✅ **Dedup** — `_attentionRequest` promise-level dedup prevents concurrent `listSummaries` calls from `refreshSummaryAttentionCount()`. Works correctly for overlapping callers (interval, focus, space-changed, bootstrap).
- ✅ **Type contract** — `ListSummariesResponse` now includes `attention_count`, `unread_count`, `pending_invitation_count` (`types/summary.ts:270`). `SummaryListItem` includes `is_unread`, `has_pending_invitation`, `needs_attention`, `current_result_id`, `current_personal_version_id`, `activity_at`. All new fields are optional, preserving backward compatibility.
- ✅ **Test coverage** — `SummaryListPage.attention.test.tsx` covers optimistic decrement + reload suppression; `SummaryDetailPage.schedule.test.tsx` covers mark-read dispatch with `task_no` deep-link; `SummaryCard.test.tsx` covers attention dot rendering.

## Findings

No P0/P1 issues; three P2 items below.

### P2 — Triple badge emission and redundant API calls (`packages/dmworksummary/src/module.tsx:41`)

`refreshSummaryAttentionCount()` emits both `summary-badge-update` and `summary-attention-count-refreshed`. When the list page is mounted, its `handleAttentionCountRefreshed_` handler sees a count change, calls `loadData()`, and `emitBadgeUpdate(resp.attention_count)` — which emits `summary-badge-update` a second time with the same count. Separately, `handleSummaryRead_` calls `emitBadgeUpdate()` *without* arguments, triggering an independent `listSummaries({ page_size: 1 })` fetch that is **not** deduped by the module's `_attentionRequest` guard. A single mark-read cycle can produce 2-3 badge emissions and 2-3 API calls for identical data.

Fix direction: remove `summary-badge-update` from `refreshSummaryAttentionCount()` — let the list page own badge emission when mounted; or have `handleSummaryRead_` emit the optimistic count directly instead of fetching.

### P2 — Stale JSDoc on `emitBadgeUpdate` (`packages/dmworksummary/src/pages/SummaryListPage.tsx:233`)

The JSDoc comment still reads *"badge = count of WAITING_CONFIRM tasks (summary ready, waiting for user to confirm)"*, but the badge source has changed from `resp.total` (filtered by `status: WAITING_CONFIRM`) to `resp.attention_count` (unread + pending invitations). The implementation is correct; the comment misleads future maintainers about badge semantics.

Fix direction: update the comment to reflect that the badge now represents `attention_count` — items needing attention (unread summaries + pending invitations).

### P2 — Redundant `markSummaryRead` calls without client-side dedup (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:493`)

`markSummaryRead` fires every time `loadDetail` loads a COMPLETED summary with a `result_id` — there is no client-side tracking of which summaries have already been marked read. Navigating away and back sends another POST. For BY_PERSON mode, both the team cursor path (`:493`) and the personal cursor path (`:618`) fire in the same load cycle, producing two POSTs with different cursors. The server is likely idempotent, but the extra traffic is unnecessary.

Fix direction: track a `Set<number>` of task IDs already marked read in the component lifetime, or add a per-task flag to suppress re-marking within the same session.

## Data Flow Tracing

| Consumer | Upstream Source | Reaches Consumer? |
|---|---|---|
| NavRail badge (`summary-badge-update`) | `listSummaries().attention_count` | ✅ Yes — via module poll, list loadData, and handleSummaryRead paths |
| List `attentionCount` field | `listSummaries().attention_count` in `loadData()` | ✅ Yes — set at `:159`; decremented optimistically in `handleSummaryRead_` at `:96` |
| `needs_attention` on `SummaryCard` | `markSummaryRead()` response → `summary-read` event → list item patch | ✅ Yes — patched at `SummaryListPage:100`; rendered at `SummaryCard:40` |
| `is_unread` on list item | `markSummaryRead()` response → `summary-read` event | ✅ Yes — patched at `:99` |
| `markSummaryRead` request body | `detail.result_id` (team) / `result.current_version_id` (personal) | ✅ Yes — both fields exist in their respective types and are guarded by truthiness checks |

- **C1 (dual-path parity)**: N/A — no symmetric add/remove or create/delete paths in this PR.
- **C2 (control-flow ordering / nested reuse)**: Clear — mark-read calls are gated by `COMPLETED` status + cursor truthiness. No nested or double-invoked paths beyond the intentional team+personal dual dispatch.
- **C3 (authorization boundary)**: N/A — no new endpoints or permission changes on the client side. Server authorization for `/summaries/{id}/read` is out of scope.
- **C4 (authorization lifecycle)**: N/A — no auth state changes.
- **C5 (build/note through ≠ runtime correct)**: N/A — no build artifacts or packaging changes.
- **C6 (governance/policy)**: N/A — no policy or governance changes.

## Cross-Round Blocker Re-Check (R6)

N/A — automated review review.

## Verdict

No P0/P1 issues found. The attention indicator feature is correctly implemented with proper race-condition guards, event lifecycle management, and test coverage. Three P2 items address performance efficiency (redundant API calls and badge emissions) and documentation accuracy — none impact correctness or user-visible behavior.

[Octo-Q] verdict: APPROVE — no blocking issues. Three P2 items (redundant badge emissions, stale JSDoc, redundant mark-read calls) are non-blocking performance/documentation improvements.

Jerry-Xin (approved):
**APPROVE** — delta re-review of `07c2eeae` (clean forward: 1 commit, ahead 1 / behind 0, touching only the two summary pages + their tests). Both peer-found bugs are byte-verified fixed, the new tests assert them, and Build is green.

### 🔴 → ✅ BUG A (post-read flicker / `attentionCount` desync) — FIXED
`packages/dmworksummary/src/pages/SummaryListPage.tsx`
- `this.attentionCount` sites: init `:65`, poll guard `:74`, mark-read sync `:95`, `loadData` sync `:146`, poll refresh sync `:242`.
- `handleSummaryRead_` (`:81`) now derives `nextNeedsAttention` from the event payload and, when an item transitions `needs_attention` true→false, runs `this.attentionCount = Math.max(0, this.attentionCount - 1)` (`:95`). This keeps the local count aligned with the server's decremented `attention_count`, so the 5s poll guard `nextCount === this.attentionCount` (`:74`) no longer diverges and no forced `loadData()` / full-list refetch fires after a read.
- Over-correction guarded: `Math.max(0, …)` prevents negative counts, and the decrement only fires on a genuine true→false transition — the invitation-pending case (`needs_attention` stays true) does not decrement. `emitBadgeUpdate`'s lightweight poll also re-syncs `attentionCount` from the server (`:242`) as a safety net.

### 🟡 → ✅ BUG B (team-path deep-link type mismatch) — FIXED
`packages/dmworksummary/src/pages/SummaryDetailPage.tsx`
- Root cause was `this.taskId === requestTaskId` comparing a numeric `detail.task_id` against a string `task_no` on a `/s/:taskNo` deep-link (always false → `summary-read` never dispatched for a team summary opened via deep-link).
- Team path now guards with same-type keys: `this.scheduleLoadSeq === seq && this.detailLookupId === requestTaskId` (`:496`), where `requestTaskId` is the captured `detailLookupId` (`:444`) — string===string on a deep-link, number===number on numeric nav. The event still dispatches the numeric `detail.task_id` (`:499`) so the list matches on `task_id`, and now also forwards the server `is_unread` / `needs_attention`.
- Personal path unchanged in behavior/no regression (`:619`): guard `this.scheduleLoadSeq === reqSeq && (this.taskId == null || this.taskId === requestTaskId)` handles the transient-null race during string deep-link load and dispatches the numeric `requestTaskId`.

### Tests — assert both fixes
- `__tests__/SummaryListPage.attention.test.tsx`: (1) attention clears → `attentionCount` 3→2 aligns with server 2 → `loadData` NOT called (no flicker); (2) invitation-pending read leaves `is_unread`/`needs_attention` and `attentionCount` untouched (no over-correction). Real class instance, real handlers, real assertions.
- `__tests__/SummaryDetailPage.schedule.test.tsx`: string deep-link `makePage('SUM-740')` → real `loadDetail()` → asserts `markSummaryRead(740, {team_result_id:91})` called and `summary-read` fired exactly once with `{taskId:740, isUnread:false, needsAttention:true}`. This is the precise old-bug case (number 740 vs string 'SUM-740'), now passing.

### No new issue
Delta is confined to state-sync + event-payload logic in the two pages; no render/HTML/URL surface touched, zero translate/locale key changes, and both `markSummaryRead` call sites preserve their signatures. Security/XSS/i18n surfaces verified clean at the prior head remain untouched by this delta.

### CI
Build **pass** at this head; no i18n key added or removed in the delta. code-review pending (this review).

yujiawei (approved):
# Code Review — PR #852 (octo-web)

Reviewed at head SHA `07c2eeae3400e30c3d3a497540e38b72aa3752fe` against `main` (merge-base `1bb2cf33`). Scope: `packages/dmworksummary` unread / invitation attention indicators. This is the fix commit for the two blockers raised at the previous head `ebd3593` — both are now resolved (see below).

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The feature delivers what the PR describes: `markSummaryRead` (`POST /summaries/{id}/read`) in `summaryApi.ts:294`, a server-driven `attention_count` on the list response (`types/summary.ts:270`), a NavRail badge kept fresh by an independent poll in `module.tsx`, a red attention dot on the card (`SummaryCard.tsx:40`), and list/badge synchronization after read via the `summary-read` window event and `summary-attention-count-refreshed` bus event.
- **Over-build:** none. The two prior-round findings are fixed with narrowly scoped changes:
  - **Post-read flicker / count desync** — `SummaryListPage.handleSummaryRead_` now keeps `this.attentionCount` aligned with the successful mark-read response (`SummaryListPage.tsx:90-95`), so the next 5s poll no longer triggers a redundant full-list reload that lost scroll position. Regression test added (`SummaryListPage.attention.test.tsx`).
  - **Deep-link `task_id` type mismatch** — the team path now dispatches the numeric `detail.task_id` and guards on `seq`/`detailLookupId` rather than comparing a numeric id against a `task_no` deep-link (`SummaryDetailPage.tsx:493-503`). Regression test added (`SummaryDetailPage.schedule.test.tsx`).
- **Deviation:** none. Read-marking is still fired only after a completed team result / non-empty personal version renders, and all failure paths keep the item unread.

## 2. Code quality

**Quality: Approved**

Race-condition hygiene is strong: both mark-read call sites re-check `scheduleLoadSeq` + lookup/`taskId` before dispatching, so late responses after a task switch cannot cross-contaminate. `attention_count` is defensively read as `?? 0` at every consumer, so a backend that omits the field degrades gracefully. `module.tsx` teardown symmetrically unregisters the interval, focus listener, bus handler, and bootstrap timer. Package tests pass (386/386).

No P0/P1 issues. The following are non-blocking (P2 / nit) improvements:

- **P2 — redundant network round-trip after each read** (`SummaryListPage.tsx:106`). `handleSummaryRead_` already computes the optimistic `attentionCount`, but then calls `emitBadgeUpdate()` with no argument, which falls into the branch that re-fetches `listSummaries({ page_size: 1 })` to recompute the count. Since the optimistic value is already correct, passing it (`this.emitBadgeUpdate(this.attentionCount)`) would avoid an extra request per read and update the badge synchronously. Not a correctness bug — the fetch converges to server truth and does not reload the list — but it defeats part of the optimistic-update intent.

- **P2 — bootstrap retry does not distinguish "not ready" from "request failed"** (`module.tsx:40-58`). `refreshSummaryAttentionCount` returns `false` both when login/Space isn't ready yet *and* when the network request throws. `bootstrapSummaryAttentionCount` retries any `false` at a fixed 100ms up to 50 times, so a transient backend 5xx during cold start produces up to ~50 rapid retries before falling back to the 5s poll. Bounded and self-terminating, so not a storm — but ideally the fast 100ms retry should apply only to the not-yet-ready case, with backoff (or no retry) on actual request failure.

- **P2 — possible over-decrement of the optimistic count under burst events** (`SummaryListPage.tsx:86-95`). Both the team and personal detail paths can each dispatch a `summary-read` for the same task. `current` is read from `this.state.items` synchronously while the `-1` mutates the instance field `attentionCount`; two events landing before React commits the `setState` would each see `needs_attention === true` and decrement twice for one task. It self-heals on the next 5s poll, and in practice a completed task usually resolves attention in one event, so impact is a brief badge undercount at most.

- **nit — HMR-only zombie timer** (`module.tsx:50-58`, `205-218`). If the module hot-reloads while a `refreshSummaryAttentionCount` promise is in flight, the in-flight `.then` can schedule a new bootstrap timer after `import.meta.hot.dispose` already cleared it. Dev-only: in production the module is an app-lifetime singleton (registered once via the module manager) and never disposes, so there is no leak in shipped builds. A cancellation flag checked before re-arming the timer would close the dev gap.

- **nit — pre-existing hex color** `#999` at `index.css:1435` triggers a stylelint `color-no-hex` warning; it is outside this diff, noted only for awareness.

## 3. Overall verdict

**APPROVE.** Spec ✅ and Quality Approved. Both prior-round blockers are fixed with regression coverage, tests are green, and the remaining items are non-blocking polish (the `emitBadgeUpdate()` round-trip at `SummaryListPage.tsx:106` is the most worthwhile follow-up).

## 4. Suggestions

- Pass the already-computed count into `emitBadgeUpdate` in the read handler to remove the redundant fetch and make the badge update synchronous.
- Split the "not ready yet" vs "request failed" cases in the bootstrap retry so network errors back off instead of retrying at 100ms.

## 5. Additional notes

- New `SummaryListItem` fields `current_result_id`, `current_personal_version_id`, and `activity_at` (`types/summary.ts:189-191`) are declared but not yet consumed anywhere in the diff. Harmless (optional fields), but worth confirming they are intended for a follow-up rather than dead surface.

PR #823, opened 2026-07-16
feat(docs): drag-to-resize table row height (tableRow.height, SCHEMA_VERSION 19)

ploy-elison:
## What

Tables can already resize **columns** by dragging the vertical borders (#749), but rows cannot be resized. This adds the row-wise counterpart: **drag a row's bottom line to set its height**.

## How

- **Schema (`SCHEMA_VERSION` 18 → 19).** Add a `height` attribute to the `tableRow` node — `number | null`, default `null`:
  - `toDOM`: a set height renders `<tr style="height:Npx">`; `null`/unset renders a bare `<tr>` (height driven by content — identical to today, so existing documents are unchanged, no migration).
  - `parseDOM`: reads an integer px back from the `tr` inline `style="height:Npx"`.
  - The value is a single integer **scalar** per row (unlike the v4 cell `colwidth`, which is a `number[]` across spanned columns).
- **Resize plugin (`TableRowResize`).** A self-built ProseMirror plugin modelled on the column-resize handle: a grab bar on each row's bottom edge (`row-resize` cursor, shown on hover), `mousedown → drag → setNodeMarkup` writes `tableRow.height` on drop, with a minimum height so a row can't be dragged away. The bottom-edge handle is orthogonal to the column-resize handle (vertical cell borders), the row/column reorder handles and the block drag handle, and it **defers to the column-resize handle** on the shared cell corner so the hit zones never conflict. The live editor gets the drag plugin; the read-only preview carries the `height` attr only (so stored heights render) but no handle.
- **Collaboration.** `height` is an ordinary node attribute, so a resize lands as one plain transaction and syncs through `y-prosemirror` like any edit (scalar last-write-wins, no structural change to the grid).

## Tests

- Unit / component (`TableRowHeight.test.ts`): parse/render contract, the `setNodeMarkup` transaction, min-height clamp, per-row isolation, and a **two-peer Yjs collaboration round-trip** (normalized structural equivalence on the decoded attr, not a raw Y.Doc byte compare). New assertions were shown red before the implementation, green after.
- Schema membership (`schema/index.test.ts`) updated to v19.
- **Real-browser self-verification** (`dev/run-rowheight.mjs`, Playwright): drives an actual left-button drag on the row line, reads `editor.state.doc` (immune to CollaborationCaret decoration noise), and asserts the row height changes, **persists** in the model, and **syncs to the remote peer**; an untouched row stays `height=null`. Screenshot attached under `dev/rowheight-out/`.

## Scope / coordination

[...]

PR #853, opened 2026-07-16
fix(docs): top-align table cell content so shrunk rows don't slice the first line

ploy-elison:
## Problem

Dragging a table row shorter than its content slices the first line of cell text mid-glyph (on a row shrunk to the minimum, the text is pushed out of view entirely). The row-height clip introduced in #823 (`.octo-cell-clip`, SCHEMA_VERSION 19) is doing its job, but the text lands in the wrong place inside the clip window.

## Root cause

The clip wrapper caps the cell content with `overflow: hidden` on height-fixed rows. `overflow:hidden` establishes a **block-formatting context**, which suppresses margin-collapse on the cell's block children. A paragraph's default `margin: 1em 0` then becomes *real* space that pushes the first line **down** inside the clip window instead of collapsing away. Measured on a 24px (minimum) row: the clip window spans y=113–125, but the paragraph's text box starts at y=128 — below the clip — so the visible slice is empty or a fractional glyph.

`vertical-align: top` on the `<td>` was already correct; the culprit was the leading margin on the content inside the clip.

## Fix

Zero the leading/trailing margins of `.octo-cell-clip`'s direct children so the content sits flush to the cell's top edge:

```css
.octo-cell-clip > :first-child { margin-top: 0; }
.octo-cell-clip > :last-child  { margin-bottom: 0; }
```

The first line now renders from the top and the overflow is clipped at the row's bottom edge — the behaviour Sheets/Word use. Spacing between stacked paragraphs is preserved (only the leading/trailing margins are removed). The wrapper is shared by the editor, the read-only preview and the version diff, so all three states stay consistent.

CSS-only: no data/schema changes, and the #823 clip mechanism itself is untouched.

## Verification (:3000-equivalent prod-faithful harness)

Drove the production editor wiring (`TableCellView` NodeView + real `styles.css`) with a real Chromium drag:

- **Single-line row shrunk to min** — first line now flush to clip top (offset 0px; was +15px, pushed out of the 12px window before).
- **Multi-line row shrunk** — first paragraph flush at top, remaining lines clipped at the bottom edge; inter-paragraph spacing intact.
- Existing `TableRow*` tests: **26 passed**. Row grow/shrink, persistence, reorder, and column-resize-on-clipped-row behaviour unchanged.

Before → after (single-line row dragged to minimum height):

Before: the shrunk row shows an empty/half-cut cell (text pushed below the clip window).
After: the shrunk row shows the first line from its top, clipped cleanly at the bottom edge.

### DIFF
--- a/packages/docs/src/editor/styles.css
+++ b/packages/docs/src/editor/styles.css
@@ -2052,6 +2052,21 @@ body.octo-row-resizing {
   /* Inert by default: a plain block wrapper around the cell's content (matches the pre-v19 flow). */
   display: block;
 }
+/* XIN-1289: keep the cell's content flush to the top edge so a shrunk (clipped) row shows its FIRST
+   line from the top and clips the overflow at the row's bottom edge — the Sheets/Word behaviour the
+   product signed off on. Without this the block child's default vertical margin (paragraphs carry
+   `margin: 1em 0`) pushes the first line down INSIDE the clip window; and because the fixed-row clip
+   below sets `overflow:hidden` it forms a block-formatting context that suppresses margin-collapse, so
+   that top margin becomes real space and the first line ends up sliced mid-glyph ("露半截") instead of
+   starting at the top. Zeroing the leading/trailing margins (not the inter-block gaps) lands the first
+   line on the cell top in all three states (editor, read-only preview, version diff — all share this
+   NodeView wrapper) while preserving spacing between stacked paragraphs. */
+.octo-cell-clip > :first-child {
+  margin-top: 0;
+}
+.octo-cell-clip > :last-child {
+  margin-bottom: 0;
+}
 .octo-prose table tr.octo-row-fixed .octo-cell-clip {
   max-height: calc(var(--octo-row-h) - 12px);
   overflow: hidden;


Jerry-Xin (approved):
APPROVE — a minimal, correctly-scoped CSS-only fix for the v19 fixed-row cell clipping.

## Analysis
The two added rules in `packages/docs/src/editor/styles.css` are correct:
```css
.octo-cell-clip > :first-child { margin-top: 0; }
.octo-cell-clip > :last-child  { margin-bottom: 0; }
```
- **Root cause is right**: block children (paragraphs) carry `margin: 1em 0`. The fixed-row clip (`tr.octo-row-fixed .octo-cell-clip { overflow: hidden }`) establishes a block-formatting context, which suppresses margin-collapse — so the leading top margin becomes real space and pushes the first line down inside the clip window. A shrunk row then slices the first line mid-glyph. Zeroing the leading `margin-top` (and trailing `margin-bottom`) lands the first line flush to the cell's top edge.
- **Minimal / no side effects**: only `:first-child`/`:last-child` are zeroed, so inter-paragraph spacing between stacked blocks is preserved. The `>` child combinator scopes strictly to direct children of the cell-clip wrapper — no bleed into nested content or non-clipped cells.
- **Consistent across surfaces**: editor, read-only preview, and version diff share this NodeView wrapper, so the fix applies uniformly (matches the shared-wrapper model from the row-height work).

## Non-blocking
- Pure CSS: no HTML/JS/XSS surface, no dependency or i18n change.

✅ osv-scan / secret-scan / dependency-review pass. Build is still running at this head — assuming it goes green (this is CSS-only, no functional risk), no blockers.

yujiawei (approved):
# Code Review — PR #853 (octo-web)

**Scope:** 1 file, +15/-0, CSS only (`packages/docs/src/editor/styles.css`). No data/schema/logic changes; the #823 (SCHEMA_VERSION 19) clip mechanism is untouched.

## 1. Spec compliance

The change does exactly what the description says and nothing more:

- **Missing:** none. The stated goal — top-align cell content inside the clip window so a shrunk row shows its first line from the top — is implemented by zeroing the leading/trailing margins of `.octo-cell-clip`'s direct children.
- **Extra:** none. No new flags, attributes, or JS.
- **Deviation:** none. The `overflow:hidden` clip and `--octo-row-h` cap on `tr.octo-row-fixed .octo-cell-clip` are left as-is.

**Spec: ✅**

## 2. Code quality

Verified against the source, not just the diff:

- `.octo-cell-clip` is the NodeView `contentDOM` (`TableCellView.ts:37-42`), so `> :first-child` / `> :last-child` correctly target the cell's top/bottom block content — the selector matches the actual DOM the editor produces.
- **Root-cause reasoning is accurate.** The fixed-row clip sets `overflow: hidden` (`styles.css:2072`), which establishes a block-formatting context and suppresses margin-collapse; the paragraph's UA-default `margin: 1em 0` then becomes real space that pushes the first line down inside the clip window. Zeroing only the leading/trailing margins is the minimal correct fix and preserves inter-paragraph spacing.
- **Specificity is sufficient.** `.octo-cell-clip > :first-child` (0,2,0) cleanly overrides the UA `p` margin; there is no competing `.octo-prose p` margin rule in this stylesheet, so no specificity fight.
- **Column-resize handle unaffected.** The handle is absolutely positioned against the `position:relative` `<td>` and escapes the wrapper (documented at `styles.css:2047-2049`), so this content-margin change does not touch resize behaviour.
- **No functional/concurrency/error-handling surface** — pure presentational CSS, so there is nothing to regress in the collab or persistence paths.

**Quality: Approved**

### Non-blocking notes (nits)

- These two rules apply to **every** `.octo-cell-clip`, not only clipped (`octo-row-fixed`) rows, so auto-height cells also lose their first paragraph's top margin and now sit flush to the top padding. This is consistent with typical spreadsheet/table cell rendering and matches the "all three states" intent stated in the description, so it reads as deliberate — flagging only so it's a conscious decision rather than a side effect.
- Consider a brief regression check (or a preview-DOM assertion alongside `TableRowClipPreview.test.ts`) for a multi-paragraph *non-clipped* cell to lock in that inter-paragraph spacing is preserved, since these selectors now fire on unclipped rows too. Not required to merge.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Minimal, correctly-scoped fix with accurate root-cause analysis; the only observations are non-blocking nits.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #853 (octo-web)

## Summary

This PR adds two CSS rules to `.octo-cell-clip` — zeroing `margin-top` on the first child and `margin-bottom` on the last child — to fix a visual defect where the first line of content in height-constrained (clipped) table rows gets sliced mid-glyph. The root cause is well-diagnosed: paragraph default margins (`1em 0`) push content below the clip's top edge, and because `overflow: hidden` on the fixed-row clip creates a block-formatting context that suppresses margin collapse, that top margin becomes real space instead of collapsing through the wrapper.

The fix is compact, well-commented, and correctly targets the described problem for the most common content type (paragraphs). One non-blocking specificity gap is noted below.

## Verification

- ✅ **Diff scope** — single file, 15 additions / 0 deletions. Pure CSS, no logic changes.
- ✅ **Selector correctness** — `.octo-cell-clip > :first-child` and `:last-child` use the child combinator (`>`), preventing unintended matches on deeper descendants. Confirmed `.octo-cell-clip` is created by `TableCellView` (`packages/docs/src/editor/TableCellView.ts:38`) as the content wrapper for all table cells in editor, preview, and diff views.
- ✅ **BFC interaction verified** — in non-fixed rows `.octo-cell-clip` has `display: block` without `overflow: hidden`, so margin collapse occurs normally and the zeroed margins reduce cell top spacing from ~15px to 6px (the `<td>` padding). In fixed rows (`overflow: hidden`), the BFC prevents collapse and the zeroed margins correctly land the first line at the clip top. Both paths are consistent with the stated intent.
- ✅ **No dangling selectors** — `.octo-cell-clip` is used in 5 CSS locations and 11 TS/TSX references; no orphaned or mismatched class names.
- ✅ **Data flow** — the CSS targets a presentational wrapper whose DOM structure is controlled entirely by `TableCellView`. No upstream data dependency, no runtime state, no API contract affected.

Static analysis only at head `e470d40c674f9b2ab6f99983592cc218803a1d8d`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 specificity gap below.

### P2 — Heading-first specificity gap (`packages/docs/src/editor/styles.css:2064`)

The new `.octo-cell-clip > :first-child` rule carries specificity `(0,1,1)`. The heading margin rule at `:836` — `.octo-prose .ProseMirror h1`–`h6 { margin: var(--wk-sp-5) 0 var(--wk-sp-2) }` — carries specificity `(0,2,1)`. When a heading is the first child of `.octo-cell-clip` inside a fixed-height row, the heading's `var(--wk-sp-5)` top margin wins and the sliced-first-line behaviour the PR targets can still occur for headings.

Not a regression — the heading case was already broken before this PR, and headings as first-child in table cells are uncommon. If this edge case matters for the product sign-off, a companion rule like `.octo-cell-clip > h1:first-child, .octo-cell-clip > h2:first-child, …` at `(0,2,1)` or a single `.octo-prose .octo-cell-clip > :first-child` at `(0,2,1)` would close it.

## Blind-point checklist

- **C1 (dual-path parity):** N/A — no paired add/remove or symmetric paths in a CSS-only change.
- **C2 (control-flow ordering):** N/A — no control flow; CSS cascade ordering is correct (new rules placed after `.octo-cell-clip` base rule, before the `octo-row-fixed` override).
- **C3 (authorization boundary):** N/A — no auth/permission surface.
- **C4 (lifecycle cascade):** N/A — no state lifecycle.
- **C5 (build ≠ runtime):** N/A — CSS is consumed directly by the browser; no build-time transform that could diverge from source.
- **C6 (governance/docs self-consistency):** N/A — no governance or policy docs changed.

## Data flow trace

`.octo-cell-clip` is instantiated in `TableCellView` constructor (`:38`) as a plain `<div>` appended to each `<td>`/`<th>`. Its children are ProseMirror content nodes (paragraphs, headings, lists). The new CSS rules target only the presentational layer — no data flows through these selectors, no runtime state is affected. The fix is purely visual.

## Cross-round blocker recheck (R6)

N/A — first review round.

## Verdict: APPROVED

The fix correctly addresses the described visual defect for the common case (paragraphs as first/last child). The heading specificity gap is a pre-existing limitation, not a regression introduced by this PR. No correctness, security, or build concerns.

[Octo-Q] verdict: APPROVE — only a P2 specificity gap for heading-first cells (pre-existing, not a regression); no P0/P1 blockers.

PR #854, opened 2026-07-16
fix(docs): localize member source label (direct/invite)

ploy-elison:
## What

Non-owner members in the docs member list show their join source after the name (e.g. ` · direct` / ` · invite`). That suffix was rendered from the raw `Member.source` enum value, so a zh-CN user saw the English word instead of a localized label. Owners don't show this tag.

## Change

- Add `docs.member.source.direct` / `docs.member.source.invite` to both `zh-CN.json` and `en-US.json` (key parity preserved by the existing i18n parity guard).
  - `direct` → `直接添加` / `Direct`
  - `invite` → `邀请加入` / `Invite`
- Render the label through `t('docs.member.source.<source>')` in `MemberPanel.tsx` instead of printing the raw enum.

No changes to member-source logic, the enum, data, or schema — this is a text-only i18n fix.

## Notes on wording

Chinese wording (`直接添加` / `邀请加入`) is aligned with the existing member-management copy. Happy to adjust to a product-preferred term if there's an established one.

## Verification

- `packages/docs` i18n + MemberPanel tests pass; added a value guard asserting the new labels in both locales.
- Verified the exact runtime path: registered the docs namespace against the real `I18nService` and confirmed `t('docs.member.source.direct'|'invite')` resolves to the Chinese strings under zh-CN and the English strings under en-US (nested keys flatten to dotted paths, so `docs.member.source.*` resolves correctly).

### DIFF
--- a/packages/docs/src/members/MemberPanel.tsx
+++ b/packages/docs/src/members/MemberPanel.tsx
@@ -195,7 +195,9 @@ export function MemberPanel({
               <span className="octo-uid">
                 {displayName(m.uid)}{' '}
                 {isOwner && <span className="octo-owner-badge">{t('docs.member.ownerBadge')}</span>}
-                {!isOwner && <small style={{ color: 'var(--octo-muted)' }}> · {m.source}</small>}
+                {!isOwner && (
+                  <small style={{ color: 'var(--octo-muted)' }}> · {t(`docs.member.source.${m.source}`)}</small>
+                )}
               </span>
               <select
                 value={m.role}
--- a/packages/docs/src/i18n/en-US.json
+++ b/packages/docs/src/i18n/en-US.json
@@ -385,6 +385,10 @@
     "remove": "Remove",
     "owner": "owner",
     "ownerBadge": "Owner",
+    "source": {
+      "direct": "Direct",
+      "invite": "Invite"
+    },
     "ownerCannotRemove": "The owner cannot be removed",
     "errorAdd": "Failed to add member.",
     "errorRemove": "Failed to remove member.",
--- a/packages/docs/src/i18n/i18n.test.ts
+++ b/packages/docs/src/i18n/i18n.test.ts
@@ -48,6 +48,16 @@ describe('docs i18n', () => {
     expect(zhCN.toolbar.fontFamilyDefault).not.toBe(zhCN.toolbar.fontSizeDefault)
     expect(enUS.toolbar.fontFamilyDefault).not.toBe(enUS.toolbar.fontSizeDefault)
   })
+
+  // The member row shows the join source ("direct" / "invite") after a non-owner's name. Those
+  // values were rendered as the raw enum key, so a zh-CN user saw the English word " · direct".
+  // They are now localized through docs.member.source.*.
+  it('localizes the member source label in both locales', () => {
+    expect(zhCN.member.source.direct).toBe('直接添加')
+    expect(zhCN.member.source.invite).toBe('邀请加入')
+    expect(enUS.member.source.direct).toBe('Direct')
+    expect(enUS.member.source.invite).toBe('Invite')
+  })
 })
 
 describe('DocsModule i18n registration', () => {
--- a/packages/docs/src/i18n/zh-CN.json
+++ b/packages/docs/src/i18n/zh-CN.json
@@ -385,6 +385,10 @@
     "remove": "移除",
     "owner": "所有者",
     "ownerBadge": "所有者",
+    "source": {
+      "direct": "直接添加",
+      "invite": "邀请加入"
+    },
     "ownerCannotRemove": "无法移除所有者",
     "errorAdd": "添加成员失败。",
     "errorRemove": "移除成员失败。",


mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #854 (octo-web)

## Summary

This PR localizes the member source label displayed in the document member panel. Previously, non-owner member rows rendered the raw enum value (`m.source`) directly — so a zh-CN user would see the untranslated English string ` · direct` or ` · invite`. The fix wraps the value in `t()` with a dotted key (`docs.member.source.${m.source}`) and adds the corresponding translation entries in both `en-US.json` and `zh-CN.json`.

The change is small (4 files, +21/−1), well-scoped, and follows the established i18n pattern already used for role labels (`docs.role.${r}`). A regression test pins the new string values in both locale bundles, and the existing key-parity guard ensures the two locales stay in sync.

## Verification

- ✅ **Type-safety of `t()` key construction** — `Member.source` is typed `'direct' | 'invite' | 'owner'` (`packages/docs/src/members/api.ts:13`). The translation JSON covers `direct` and `invite`. The `'owner'` value is produced only by `withSyntheticOwner()` (`packages/docs/src/members/sort.ts:26`), and the rendering guard `!isOwner` (`packages/docs/src/members/MemberPanel.tsx:199`) prevents the source label from being shown for the synthetic owner row. No runtime path reaches `t('docs.member.source.owner')`.
- ✅ **Nested-key resolution** — The i18n system flattens nested JSON into dotted keys (confirmed by `flattenKeys` in `packages/docs/src/i18n/i18n.test.ts:10` mirroring I18nService). The `docs.role.reader`/`docs.role.writer` pattern at `packages/docs/src/members/MemberPanel.tsx:208` already uses the same dotted-key interpolation and works correctly.
- ✅ **Key parity** — The existing parity test (`packages/docs/src/i18n/i18n.test.ts:19`) flattens both locale bundles and asserts identical key sets. Adding `source.direct` and `source.invite` to both JSON files keeps the parity intact.
- ✅ **No diff-scope regression** — The single-line change at `:199` replaces `{m.source}` with `{t(`docs.member.source.${m.source}`)}`. For all reachable `source` values (`'direct'`, `'invite'`), a valid translation key exists. Behaviour for owner rows is unchanged (badge still shows, source label still hidden).

Static analysis only at head `6ff6e25bf510f8c7d55b2f8f865bf6c5c3fd9ede`; build and tests not executed in this environment.

## Findings

The PR does exactly what its title claims: the raw enum key that zh-CN users saw is now resolved through the i18n bundle. The translation values are sensible — `直接添加` ("added directly") and `邀请加入` ("joined by invitation") are natural Chinese phrasings for the join-source context. The test pins exact values in both locales, which is the right granularity for a string-lookup fix.

## Verdict: APPROVED

Clean, minimal, correctly-scoped localization fix. Data-flow tracing confirms that every reachable `m.source` value at the rendering point has a matching translation key. The `'owner'` variant in the `Member.source` union type is gated behind `isOwner` and never reaches the `t()` call. Key-parity and regression tests are in place.

yujiawei (approved):
# Code Review — PR #854 (octo-web)

A clean, correctly scoped text-only i18n fix. Non-owner member rows now render their join source through `t('docs.member.source.<source>')` instead of printing the raw enum, and the two new keys (`direct` / `invite`) are added symmetrically to both locales. Verified against head SHA `6ff6e25b` with the branch checked out locally.

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. The stated goal — localize the `direct` / `invite` labels shown after non-owner names — is fully implemented in `MemberPanel.tsx` (render) plus `en-US.json` / `zh-CN.json` (keys).
- **Over-built:** none. No enum, data, schema, API, or member-source logic changed; the diff is text + one render-site swap, exactly as described.
- **Deviations:** none. Implementation matches the "text-only i18n" approach in the PR description.

## 2. Code quality

**Quality: Approved**

No P0/P1/P2 findings. Verification performed:

- **Namespace/key resolution** — `DocsModule` registers the `docs` namespace (`module.tsx:259`), `I18nService` flattens nested JSON to dotted keys, so `docs.member.source.direct|invite` resolves to the real strings. The bare-nested JSON (`member.source.*`) + `docs.` render prefix line up correctly.
- **Locale parity** — both locales gained the identical `member.source.{direct,invite}` subtree; the strict `flattenKeys` parity guard in `i18n.test.ts` would fail hard on a split, and it passes.
- **`owner` source is provably unreachable at this render site** — the label is behind `!isOwner`, and the only producer of `source: 'owner'` (`withSyntheticOwner` in `sort.ts`) sets `uid === ownerId`, which forces `isOwner` true. So `t('docs.member.source.owner')` can never execute and no `owner` key is needed — the two-key set is complete.
- **Single consumer** — `MemberPanel.tsx:199` is the only site rendering `m.source`; no other raw-enum render site was left un-localized.
- **Tests** — the docs `i18n` + `members` suites (44 tests) pass at the head SHA. The 4 failures in a full package run are pre-existing tiptap mount errors in `editor/Toolbar.test.tsx` / `editor/EditorShell.test.tsx`, disjoint from this diff.

### Non-blocking suggestions (nits)

- **Latent fallback behavior.** `I18nService.t()` returns the raw key string on a miss. `Member.source` is typed `'direct' | 'invite' | 'owner'`, so nothing out-of-range can reach the template today — but if the backend ever adds a new non-owner source value, a row would render the literal `docs.member.source.<x>` (worse than the old bare enum) until a matching key is added. Worth a mental note for any future enum extension; not something to fix here.
- **Test altitude.** The new test asserts the JSON values in both locales (good, and parity is guarded), but does not exercise the component binding itself — a typo in the `docs.member.source.` prefix would pass the value tests yet ship broken UI due to the key-echo fallback. A single `MemberPanel` render asserting a `direct` row shows `Direct` / `直接添加` would close that gap. Optional.

## 3. Overall verdict

**APPROVE** — spec ✅ and quality Approved. Ship it.

Wording note from the PR body (`直接添加` / `邀请加入`) reads consistent with existing member-management copy; no objection.

Jerry-Xin (approved):
APPROVE — a clean, correctly-scoped i18n localization fix.

## Analysis
`MemberPanel.tsx` changes the non-owner member source label from the raw enum (` · {m.source}`) to `t(\`docs.member.source.${m.source}\`)`, so a zh-CN user no longer sees the English literal "direct"/"invite".
- **Parity**: the `member.source.direct` / `member.source.invite` keys are added symmetrically to both `en-US.json` (Direct/Invite) and `zh-CN.json` (直接添加/邀请加入) at the same path.
- **No removed-in-use key**: additions-only; the newly-referenced dynamic key resolves in both locales for the two documented enum values.
- **Test**: `i18n.test.ts` asserts all four values in both locales.
- No security/XSS surface (i18n + one escaped JSX interpolation of a translated string).

## Non-blocking
- The key is built dynamically from `m.source`. If a future member-source enum value other than `direct`/`invite` appears, `t('docs.member.source.<unknown>')` would render the raw key (missing-translation fallback) — no worse than the pre-fix raw-enum behaviour, but a small render-test / exhaustive-mapping guard would harden it (mirrors yujiawei's note).

✅ Build / osv-scan / secret-scan / dependency-review all pass.

packages/docs/src/members/MemberPanel.tsx
```diff
@@ -195,7 +195,9 @@ export function MemberPanel({
               <span className="octo-uid">
                 {displayName(m.uid)}{' '}
                 {isOwner && <span className="octo-owner-badge">{t('docs.member.ownerBadge')}</span>}
-                {!isOwner && <small style={{ color: 'var(--octo-muted)' }}> · {m.source}</small>}
+                {!isOwner && (
+                  <small style={{ color: 'var(--octo-muted)' }}> · {t(`docs.member.source.${m.source}`)}</small>
```
OctoBoooot, line 199: **[minor]** `Member.source` is typed `'direct' | 'invite' | 'owner'` (api.ts:13) — three values — but this PR localizes only `direct`/`invite`. This label renders under `!isOwner`, and `isOwner` is uid-derived (`m.uid === resolvedOwner`, :191), NOT `source === 'owner'`. So a member with `source: 'owner'` whose uid ≠ the resolved owner would hit this branch and render `t('docs.member.source.owner')` → the raw key `docs.member.source.owner`. Narrow (needs a `source:'owner'` member that isn't the resolved owner — unusual) and low-harm (raw key shown, not data/security), so non-blocking — but it's the coverage gap the dynamic `${m.source}` key invites. Fix: add `docs.member.source.owner` in both locales, or confirm this branch can never see a `source:'owner'` member. (The direct/invite localization itself is correct + symmetric + tested — good fix.)

OctoBoooot (approved):
## Review: fix(docs): localize member source label (direct/invite) (#854)

**Verdict:** Approve
A correct, minimal i18n fix: the member row's join-source label was rendered as the raw enum (`· direct`), so zh-CN users saw the English word; it now goes through `docs.member.source.*`. Keys are symmetric across both locales, no removed-key regression, CI green. One non-blocking minor: the `source` enum has three values but only two are localized.

## What I checked
- **The fix is correct.** `MemberPanel.tsx:199` changes `· {m.source}` → `· {t(\`docs.member.source.${m.source}\`)}`, and `docs.member.source.{direct,invite}` are added **symmetrically** to both `en-US` (Direct/Invite) and `zh-CN` (直接添加/邀请加入) at the same path.
- **No removed-in-use key** (the recurring trap): additions-only, 0 removed key-lines in either locale.
- **Test coverage** — `i18n.test.ts` asserts all four values (direct/invite × en/zh), real parity coverage.
- **No XSS/security surface** — one JSX interpolation of a translated string (React-escaped), i18n-only.
- CI green (Build success, secret-scan / osv-scan / dependency-review pass; check-sprint skipped = governance).

### Minor
- `packages/docs/src/members/MemberPanel.tsx:199` — `Member.source` is typed `'direct' | 'invite' | 'owner'` (api.ts:13), a **three-value** enum, but this PR localizes only `direct` and `invite`. The label renders under `!isOwner`, and `isOwner` is derived from **uid** (`m.uid === resolvedOwner`, :191), not from `source === 'owner'`. So a member with `source: 'owner'` whose uid ≠ the resolved owner would render `t('docs.member.source.owner')` → the raw key `docs.member.source.owner`. Narrow (requires a `source:'owner'` member that isn't the resolved owner — an unusual backend state), and no worse than data-loss/security — a raw key shown. Fix: add a `docs.member.source.owner` key too (both locales), or confirm the `!isOwner` render can never reach a `source:'owner'` member. Non-blocking, but it's the exact case the dynamic `${m.source}` key invites — the enum has a third value the fix doesn't cover.

### Praise
- Routing the source label through i18n with a parity test asserting both locales is the right shape for a copy-localization fix, and it fixes a genuine zh-CN user-visible defect (English enum leaking into a Chinese UI). The dynamic key keeps the JSX clean.

### Net
Correct member-source localization: keys symmetric, no removed-key regression, tested, CI green. The one gap is that `Member.source` has a third value (`owner`) the fix doesn't localize, reachable only in the unusual case of a `source:'owner'` member that isn't the uid-resolved owner → raw key. **Approve** with that non-blocking minor (add the `owner` key or confirm it's unreachable). `check-sprint` (governance) still needs a maintainer to clear before merge.

PR #850, opened 2026-07-16
feat(dmworksummary): agent-driven chat summary with reference-based r…

wanyiwang06:
…efine

Introduce a full agent-driven chat summary UX in `packages/dmworksummary`, replacing the classic form-based refine workflow with a chatbot-first experience that streams progress and supports referencing prior summaries as context.

What's included (see PR body for the F1-F6 breakdown):

- MSW mock infrastructure for dmworksummary tests
- Agent summary entry (split button in both ChatSummaryNewModal and SummaryCreatePage) + interactive chatbot Q&A panel
- Save-as-summary UI with dual-entry robustness (page + modal)
- SSE streaming with dual-mode support, session_id persistence across reopens, abstracted progress rendering (safe phase + integer count)
- Reference-based chat: pick prior summaries as context; preview via a right-side compare panel
- i18n: agent-chat + 智能助手 → 对话生成 badge rename (zh-CN / en-US)

Verification:
- typecheck: `pnpm --filter=@dmwork/summary typecheck` — no new errors introduced vs. upstream 47c504fd (pre-existing rot in SummaryDetailPage.participantReports.test.tsx and SummaryDetailPage.schedule.test.tsx is unchanged by this PR)
- Local end-to-end: exercised in the deployed OCTO stack against the companion octo-smart-summary backend PR

## Changes

-
-

## Architecture / Module Boundary

- Affected module(s):
- New or changed user-visible entry point: 
- Shared layer touched: 
- If shared code changed, impact scope:
- Duplicate entry point checked: 

## Testing

- [ ] Unit tests added/updated
- [ ] Manually verified

## Checklist

- [ ] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [ ] PR description is in English
- [ ] Added tests for my changes
- [ ] Updated documentation
- [ ] Ran `pnpm i18n:check` or confirmed the change does not affect UI copy
- [ ] Confirmed the change follows module ownership and does not add duplicate user-visible entry points
- [ ] Described impact scope when shared components, messages, bridge, or services are changed
- [ ] Followed commit message conventions (Conventional Commits)

PR #858, opened 2026-07-17
feat(interactive-card): docs access-request deny-reason dialog

an9xyz:
## Summary

Clicking 拒绝 on a docs access-request approval card now opens a required-reason dialog instead of submitting directly. The reason is submitted under `inputs[deny_reason]` (the server declares a matching hidden `Input.Text`), so it reaches the docs backend via the existing signed `DecisionRequest.Inputs` — the no-data submit contract is unchanged. Companion backend PR: Mininglamp-OSS/octo-server#605 (same branch `claude/docs-permission-card-template-bccd8h`).

## Related Issue

_None — self-initiated._

## Changes

- **`Messages/InteractiveCard/denyReasonDialog.tsx`** (new) — `isDocsDenyAction(data)` gate (matches `owner=docs / action_type=access_request.decision / decision=deny`) + `openDocsDenyReasonDialog`, a `wkConfirm` danger modal with a required, ≤200-char reason (Semi `TextArea`) and a character-limit hint. Resolves the trimmed reason on confirm, `null` on cancel; empty input keeps the dialog open with a toast.
- **`Messages/InteractiveCard/InteractiveCardCell.tsx`** — `handleSubmit` intercepts the docs deny action **only when the current frame declares the `deny_reason` input** (backward-compat: pre-existing cards without it fall back to direct submit), opens the dialog, then `performSubmit` (extracted) merges the reason into the declared inputs. Approve and all other actions submit unchanged.
- **`i18n/locales/{zh-CN,en-US}.json`** — dialog copy (title / description / reason label / placeholder / counter / buttons / validation).
- **`Messages/InteractiveCard/index.css`** — dialog content typography (scoped `wk-docs-deny-dialog-*`).

- Affected module(s): `dmworkbase` → `Messages/InteractiveCard`
- New or changed user-visible entry point: no new entry point; adds a confirmation dialog to an existing card action (docs access-request 拒绝)
- Shared layer touched: `Messages` (the shared InteractiveCard cell) + one i18n copy addition; reuses the shared `Components/WKModal` `wkConfirm`
- If shared code changed, impact scope: the deny interception is gated on `isDocsDenyAction(data)` **and** the card declaring the `deny_reason` input, so no other card family's submit behavior changes; `performSubmit` is a pure extraction of the prior submit body
- Duplicate entry point checked: not applicable (no new entry point)

## Testing

```
# octo-web/packages/dmworkbase
npx vitest run src/Messages/InteractiveCard   # 21 files, 221 tests pass (incl. new denyReasonDialog suite)
```

New `denyReasonDialog.test.tsx`: `isDocsDenyAction` matching, dialog resolves the trimmed reason on confirm, keeps open + warns on empty, resolves `null` on cancel. Separately verified against the production `adaptivecards@3.0.6` SDK that `card.getAllInputs()` **includes** the hidden (`isVisible:false`) `deny_reason` input, so the interception guard fires for the enriched card.

- [x] Manually verified (SDK `getAllInputs` behavior + dialog resolve semantics)

## Checklist

- [ ] Updated documentation — n/a (behavior covered in backend spec)
- [x] Ran i18n check — locale JSON validated (both files parse); keys added to zh-CN + en-US
- [x] Described impact scope for the shared Messages layer above

https://claude.ai/code/session_01ATmYMWdTCLZMHKfHC2xt6j

---
_Generated by [Claude Code](https://claude.ai/code/session_01ATmYMWdTCLZMHKfHC2xt6j)_


mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #858 (octo-web)

## Summary

This PR adds a deny-reason dialog for document access-request interactive cards. When an approver clicks "Deny" on a docs access-request card that declares a `deny_reason` hidden input, a modal dialog opens requiring a mandatory reason (up to 200 chars). The reason is then submitted alongside the card's declared inputs via `submitCardAction`. Cards without the `deny_reason` input (older cards sent before this feature) fall through to the original direct-submit path — a clean backward-compatible design.

The implementation is well-structured: a self-contained `denyReasonDialog.tsx` module handles action detection, dialog lifecycle, and reason collection; `InteractiveCardCell.tsx` gains a thin interception branch and a `performSubmit` extraction that preserves the full original submit contract (validation, channel resolution, generation tracking, timeout, error handling).

## Verification

Static analysis only at head `862bb2669319faca02704b2cb4434e4a324af9ae`; build and tests not executed in this environment.

- ✅ **Backward compatibility** — `InteractiveCardCell.tsx:293` gates the dialog on `card.getAllInputs().some(input => input.id === DOCS_DENY_REASON_INPUT_ID)`. Older cards without the hidden input skip the dialog and go straight to `performSubmit(card, actionId, null)`, preserving the original submission path exactly.
- ✅ **wkConfirm + Promise.reject keeps dialog open** — Traced `confirm.tsx:46-51`: when `onOk` returns a thenable, `updatePending(action)` is called; on rejection, `.catch(() => updatePending(null))` resets pending state without destroying the modal. The `denyReasonDialog.tsx:113` pattern of returning `Promise.reject(new Error("deny reason required"))` correctly keeps the dialog open and shows a Toast warning.
- ✅ **Double-resolution protection** — `denyReasonDialog.tsx:101-105`: the `settled` flag and `done()` helper ensure the outer Promise resolves at most once, regardless of how many times `onOk`/`onCancel` fire.
- ✅ **Post-dialog race guards** — `InteractiveCardCell.tsx:303-305`: after dialog resolves, `this.mounted`, `this.submitting`, and `this.computeState().decision` are all re-checked before `performSubmit` is called. This correctly handles component unmount, concurrent submissions, and card state changes during dialog lifetime.
- ✅ **performSubmit extraction preserves contract** — The extracted method at `:318-392` maintains the full original flow: `collectCardInputs` → `validateCardInputs` → channel resolution → `submitGen` tracking → `submitCardAction` → timeout/error handling. The only addition is `Object.assign(inputs, extraInputs)` at `:337` to merge the deny reason.
- ✅ **i18n parity** — Both `en-US.json` and `zh-CN.json` add exactly 11 keys under `message.interactiveCard.denyDialog.*`. All interpolation variables (`{{actor}}`, `{{title}}`, `{{no}}`, `{{max}}`) match between template strings and `t()` call sites.
- ✅ **Input size contract** — TextArea `maxLength={200}` keeps the reason well within `MAX_FIELD_BYTES` (4096) enforced by `validateCardInputs`. The merged `inputs` object passes the same validation.
- ✅ **CSS classes defined** — All five classes used in `DenyReasonBody` JSX (`wk-docs-deny-dialog`, `-no`, `-desc`, `-label`, `-required`, `-counter`) have matching rules in the changed `index.css`.
- ✅ **Unit test coverage** — `denyReasonDialog.test.tsx` covers `isDocsDenyAction` (positive + negative cases), `DOCS_DENY_REASON_INPUT_ID` value lock, and `openDocsDenyReasonDialog` resolve semantics (confirm with trim, empty-reason rejection, cancel → null).

## Findings

### Nit — Dangling CSS class name (`denyReasonDialog.tsx:116`)

`className: "wk-docs-deny-modal"` is passed to `wkConfirm` but has no matching CSS rule in the changed `index.css`. The modal shell inherits `wk-modal wk-modal-confirm` styles which are likely sufficient, but the dangling class name could confuse future readers. Either add a minimal rule (even just a comment-style marker) or remove the class if no additional styling is planned.

### Nit — Redundant optional chaining (`InteractiveCardCell.tsx:299`)

`asString(data?.doc_title)` uses `?.` but we are inside the `isDocsDenyAction(data)` branch where `!!data` is guaranteed by the type guard at `denyReasonDialog.tsx:27`. The `?.` is dead code — harmless, but removing it would make the invariant explicit.

## Data Flow Trace

1. **Action data → deny detection**: `action.data` (from AdaptiveCard SDK) → `isDocsDenyAction(data)` checks `owner === "docs" && action_type === "access_request.decision" && decision === "deny"` → correct gate.
2. **Dialog context extraction**: `data.doc_title`, `data.actor`, `data.request_id` → `asString()` coercion → `DocsDenyDialogContext` fields. Missing fields → fallback to generic `desc` key (no context interpolation). Safe.
3. **Reason collection**: `TextArea onChange` → `setValue(next)` (React state) + `props.onChange(next)` → `box.reason = value` (closure capture). On confirm: `box.reason.trim()` → empty check → resolve. Non-empty reason flows to step 4.
4. **Reason → submission**: `performSubmit(card, actionId, { deny_reason: reason })` → `collectCardInputs(card)` collects all declared inputs → `Object.assign(inputs, { deny_reason: reason })` overwrites the hidden input's empty value → `validateCardInputs` → `submitCardAction({ ..., inputs })`. The `deny_reason` key matches the server-declared `DocsDenyReasonInputID`.

## Blind-spot Checklist

- **C1 (dual-path parity)**: N/A — no symmetric add/remove or create/delete paths changed.
- **C2 (control-flow ordering)**: Clear — the deny-reason interception is a single branch at the top of `handleSubmit`; no nested reuse or double-invocation risk.
- **C3 (authorization boundary)**: N/A — no auth/permission changes; submission still goes through `submitCardAction` with existing auth.
- **C4 (authorization lifecycle)**: N/A — no auth state changes.
- **C5 (build/note ≠ runtime)**: N/A — no build/config changes.
- **C6 (governance/docs)**: N/A — no governance docs changed.

## Cross-round Blocker Review

## Verdict: APPROVED

Clean, well-tested feature addition with thoughtful backward compatibility. The backward-compat gate on `getAllInputs`, the `settled` flag for promise safety, and the post-dialog race guards all demonstrate careful attention to edge cases. Two minor nits are non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; implementation is correct, backward-compatible, and well-tested.

lml2468 (changes requested):
## 🐙 PR #858 Review — `feat(interactive-card): docs access-request deny-reason dialog`

**Head:** `862bb2669319` · **Base:** `main` (merge-base `2467b307`) · +311 −0 / 6 files
**Intent:** Clicking 拒绝 on a docs access-request card opens a required-reason dialog; the reason rides `inputs[deny_reason]` through the existing signed `DecisionRequest.Inputs`, gated so pre-existing cards without the declared input fall back to direct submit.

### Verification (local checkout @ head)
| Check | Result |
|---|---|
| `vitest run src/Messages/InteractiveCard` (21 files) | ✅ 221/221 pass (incl. new `denyReasonDialog` suite) |

- Deny interception is double-gated: `isDocsDenyAction(data)` **and** the current frame declaring `deny_reason` (`card.getAllInputs()`), so old cards fall back to direct submit — the backward-compat contract in the description holds.
- `performSubmit` is a **pure extraction** of the prior submit body (diff is additions-only, zero lines removed) — approve and all non-docs actions submit unchanged. No over-build, no scope creep.

### Gate 2 — Code quality: one blocker (below), otherwise solid
**Clean / verified:**
- Async-dialog re-entrancy is handled: after the dialog resolves, it re-checks `this.mounted` / `this.submitting` / decision-still-interactive before submitting (`mounted` is lifecycle-managed at :106/:118). Good.
- `wkConfirm` empty-reason contract is correct: `onOk` returns a **rejected** promise → `WKModal/confirm.tsx:47` `.catch(() => updatePending(null))` keeps the modal open (no `destroy()`). Verified against the modal source.
- The `base.` prefix on `t()` keys is **correct** — `registerNamespace("base", …)` auto-prefixes flattened keys (not the #850-style bug). Verified.
- i18n keys symmetric (11 en / 11 zh); reason is plain data (no `dangerouslySetInnerHTML`); `TextArea maxLength=200` enforced.

---

### 🔴 P2 (blocker for this PR) — i18n interpolation vars passed at the wrong level → literal `{{placeholders}}` shown to users

`denyReasonDialog.tsx` passes interpolation variables at the **top level** of the options object, but this repo's `t()` reads them only from `options.values` (`I18nService.ts:140` → `interpolate(message, options?.values)`).

Affected call sites — all render broken:
- `:47` `t("…denyDialog.descWithContext", { actor, title })` → shows literal `{{actor}}` / `{{title}}`
- `:56` `t("…denyDialog.requestNo", { no })` → shows literal `{{no}}`
- `:79` `t("…denyDialog.counter", { max: MAX_REASON })` → shows literal `{{max}}`

**Repro** (byte-tested against the real `interpolate`):
```
t('… {{actor}} … "{{title}}".', { actor:'李四', title:'Roadmap' })
→ 'Once denied, {{actor}} will not get view access to "{{title}}".'   // vars NOT substituted
t(same, { values: { actor:'李四', title:'Roadmap' } })
→ 'Once denied, 李四 will not get view access to "Roadmap".'          // correct
```
The `counter` string (`最多 {{max}} 个字`) is on the primary new surface and renders broken on **every** dialog open. The existing correct pattern is `JoinOrganization/index.tsx:58` (`t(key, { values: { … } })`).

**Why tests miss it:** the suite mocks `t: (key) => key`, so interpolation is never exercised.

**Fix:** wrap each variable map in `values`:
```ts
t("base.message.interactiveCard.denyDialog.descWithContext", { values: { actor, title } })
t("base.message.interactiveCard.denyDialog.requestNo",      { values: { no } })
t("base.message.interactiveCard.denyDialog.counter",        { values: { max: MAX_REASON } })
```
Consider a test that asserts a rendered interpolated string (not the mocked identity `t`) to lock this.

### Adversarial pass — checks that did NOT find a bug
Backward-compat fallback (old cards) ✅ · async re-entrancy/unmount guards ✅ · empty-reason keep-open contract ✅ · `base.` namespace resolution ✅ · en/zh key symmetry ✅ · no XSS surface (plain data) ✅ · extraction fidelity (additions-only) ✅.

---

### ❌ Review Verdict: REQUEST_CHANGES

Spec ✅ and the logic/refactor are clean, but the feature's own dialog copy renders unresolved `{{actor}}/{{title}}/{{no}}/{{max}}` placeholders because interpolation vars aren't passed under `values` — a user-visible defect on the exact UI this PR adds, invisible to the current tests. One-line-per-call fix; happy to re-review on push.

yujiawei (approved):
# Code Review — PR #858 (octo-web)

**Verdict: APPROVE.** No blocking (P0/P1) issues found. The change is well-scoped, backward-compatible, and the load-bearing SDK assumption checks out. A few non-blocking suggestions and one cross-PR contract item to confirm are listed below.

## Scope check

The diff implements exactly what the PR describes — a required deny-reason dialog for the docs access-request card — and nothing beyond it. The deny interception is gated on both `isDocsDenyAction(data)` **and** the card actually declaring the `deny_reason` input, so no other card family's submit path changes. Approve and all non-docs actions are untouched. No unrequested features, flags, or interfaces added.

## Verification performed

- **Load-bearing SDK assumption is correct.** I inspected `adaptivecards@3.0.6` (`lib/card-elements.js`): `Container.getAllInputs()` recurses through all children and `Input.getAllInputs()` returns `[this]` **unconditionally**, ignoring `isVisible`. So the hidden `deny_reason` input IS included in `getAllInputs()`, and the interception guard fires for the enriched card. The gate is sound.
- **Backward-compat is correct.** Pre-existing cards without the declared `deny_reason` input fall through to the original direct-submit path (`performSubmit(card, actionId, null)`), avoiding a server "undeclared input" rejection. `performSubmit` is a faithful extraction of the prior submit body.
- **Double-submit is prevented.** After the dialog resolves, the guard `if (!this.mounted || this.submitting) return` plus the `computeState().decision` re-check runs before `performSubmit`; the first confirmed submit sets `submitting = true`, so a second in-flight callback returns early. No duplicate server submit.
- **Empty-reason control flow is safe.** `onOk` returns a rejected promise to keep the modal open; `WKModal/confirm.tsx` explicitly consumes it (`void result.then(...).catch(() => updatePending(null))`), so no `UnhandledPromiseRejection` escapes. The empty case shows a warning toast and keeps the dialog open.
- **Tests / i18n pass.** `vitest run src/Messages/InteractiveCard` → 21 files / 221 tests pass, including the new `denyReasonDialog` suite. `i18n:check` passes; both `zh-CN` and `en-US` receive all `denyDialog.*` keys.

1. **Re-entrancy: stacked modals on rapid double-click.** `this.submitting` is only set inside `performSubmit`, i.e. *after* the dialog is confirmed — not when the dialog opens. A very fast second click on Deny (before the modal mounts) can re-enter `handleSubmit` and open a second `wkConfirm`. In practice the modal's default `mask: true` overlay blocks the underlying button once mounted, and the post-resolve guard prevents any double *submit* — so this is a cosmetic glitch, not a correctness bug. Consider setting an "opening" flag (or reusing `submitting`) before `openDocsDenyReasonDialog` to close the window entirely.

2. **Silent no-op if the cell unmounts while the dialog is open.** If the user switches channels / the message is removed and then confirms, the post-resolve `!this.mounted` guard discards the typed reason with no feedback. No data corruption, but the reason is silently lost. Optional: surface nothing is fine, but be aware of the UX. `openDocsDenyReasonDialog` also returns no destroy handle, so the dialog cannot be programmatically dismissed on unmount (it self-closes on any button press).

3. **Stale card reference across a frame swap.** The `.then` closure captures the click-time `card` and `actionId`. If a new bot frame arrives during the open dialog, `syncSdkCard` resets `submitting = false` and remounts a new card, but the post-resolve guards check only `mounted` / `submitting` / `kind`+`interactive` — not frame identity (`renderedKey` / `submitGen`). A same-shaped new frame would pass and `performSubmit` would run against the *old* captured `card`. Narrow race, and server-side `actionId` validation likely mitigates it, but consider capturing `submitGen` at click time and re-checking it after resolve.

4. **Integration coverage gap.** Only the `denyReasonDialog` module is unit-tested. The `handleSubmit` wiring — the `isDocsDenyAction && getAllInputs()` branch, the post-resolve guard re-check, and the `deny_reason` merge into declared inputs — has no test. A regression in the branch condition or the extra-input merge would ship green.

5. **Accessibility.** The reason `TextArea` has no `aria-label` / label association or autofocus inside the danger modal.

## Cross-PR contract to confirm (coverage — outside this diff)

Because `getAllInputs()` includes the hidden input, `collectCardInputs` emits `deny_reason: ""` on **every** submit of an enriched card — **including approve** (`cardInputs.ts` stringifies `null → ""`). This is harmless *if* the backend keys its branch on the decision/`actionId` and ignores an empty `deny_reason`. Please confirm the companion server change (#605) does not treat presence/emptiness of `deny_reason` as semantically meaningful on the approve path, since approve submits now carry a field the server did not previously see. Neither PR appears to test this specific combination.

## Note on review process

This review is a single-reviewer pass. A second automated line-level pass was attempted but failed to run (infrastructure error), so it is treated as absent rather than as tacit approval; line-precision on files not directly quoted here is inferred from the diff and the installed SDK/UI sources rather than exhaustive full-file inspection.

yujiawei (changes requested):
# Code Review — PR #858 (octo-web) — Corrected

**This supersedes my earlier APPROVE on the same head SHA.** On a second pass I confirmed a user-visible i18n interpolation defect on the dialog this PR adds (independently raised by another reviewer). Corrected verdict: **CHANGES_REQUESTED**.

## Gate 1 — Spec compliance: ✅
The diff implements exactly the described feature (required deny-reason dialog for the docs access-request card) and nothing beyond it. Deny interception is double-gated on `isDocsDenyAction(data)` **and** the frame declaring the `deny_reason` input, so old cards fall back to direct submit and no other card family changes. `performSubmit` is a faithful extraction of the prior submit body (additions-only). No over-build, no missing scope.

## Gate 2 — Code quality: Changes-Requested

### 🔴 P1 (blocking) — i18n variables passed at the wrong level → literal `{{placeholders}}` render to users

This repo's `t()` reads interpolation variables **only** from `options.values`:

```
// I18nService.ts:133,140
t(key, options) { ... return interpolate(message, options?.values); }
// interpolate returns the template unchanged when values is undefined (:44)
```

But all four interpolated call sites in `denyReasonDialog.tsx` pass the variables at the **top level** of the options object, so `options.values` is `undefined` and the `{{...}}` tokens are never substituted:

- `denyReasonDialog.tsx:47` `t("…denyDialog.descWithContext", { actor, title })` → renders literal `{{actor}}` / `{{title}}` (the dialog's main description)
- `denyReasonDialog.tsx:56` `t("…denyDialog.requestNo", { no })` → renders literal `{{no}}`
- `denyReasonDialog.tsx:79` `t("…denyDialog.counter", { max: MAX_REASON })` → renders literal `{{max}}` — shown on **every** dialog open

The established repo pattern (53 call sites, e.g. `Messages/JoinOrganization/index.tsx:58`) wraps them in `values:`. Fix:

```ts
t("base.message.interactiveCard.denyDialog.descWithContext", { values: { actor, title } })
t("base.message.interactiveCard.denyDialog.requestNo",       { values: { no } })
t("base.message.interactiveCard.denyDialog.counter",         { values: { max: MAX_REASON } })
```

**Why the test suite misses it:** `denyReasonDialog.test.tsx` mocks `t: (key) => key`, so interpolation is never exercised. Please add a test that asserts a rendered interpolated string against the real `t` so this can't regress.

### Verified-clean (no bug found)
- **SDK gate is sound.** Inspected `adaptivecards@3.0.6`: `Input.getAllInputs()` returns `[this]` unconditionally (ignores `isVisible`), so the hidden `deny_reason` input IS included and the interception guard fires. Backward-compat fallback for old cards is correct.
- **No double-submit.** Post-resolve guard (`!this.mounted || this.submitting`) + decision re-check runs before `performSubmit`; the first confirmed submit sets `submitting = true`, so a second in-flight callback returns early.
- **Empty-reason keep-open is safe.** `onOk` returns a rejected promise; `WKModal/confirm.tsx` consumes it via `.catch(() => updatePending(null))` — no unhandled rejection escapes.
- Tests (221/221) and `i18n:check` pass (the interpolation bug is invisible to both).

1. **Re-entrancy (cosmetic).** `this.submitting` is set only inside `performSubmit`, not when the dialog opens, so a very fast double-click could stack a second modal. The default `mask: true` overlay and the post-resolve guard prevent any double *submit*; consider an "opening" flag to close the window entirely.
2. **Silent no-op on unmount-during-dialog.** If the cell unmounts then the user confirms, the `!this.mounted` guard discards the typed reason with no feedback. No data risk, minor UX.
3. **Stale card ref across a frame swap.** The `.then` closure captures the click-time `card`/`actionId`; the post-resolve guards check `mounted`/`submitting`/`kind`+`interactive` but not frame identity (`submitGen`/`renderedKey`). Narrow race; consider capturing `submitGen` at click time and re-checking it.
4. **Integration coverage gap.** The `handleSubmit` wiring (dialog branch + guard re-check + `deny_reason` merge) has no test; only the dialog module is unit-tested.
5. **Accessibility.** The reason `TextArea` has no `aria-label`/label association or autofocus in the danger modal.

## Cross-PR contract to confirm (coverage — outside this diff)
Because `getAllInputs()` includes the hidden input, `collectCardInputs` emits `deny_reason: ""` on **every** enriched-card submit, **including approve** (`cardInputs.ts` stringifies `null → ""`). Harmless *if* the backend keys its branch on the decision/`actionId` and ignores an empty `deny_reason`. Please confirm companion server PR #605 does not treat presence/emptiness of `deny_reason` as meaningful on the approve path.

## Note on review process
Corrected pass. One automated line-level leg failed to run (infrastructure error) and is treated as absent, not as tacit approval. My initial pass approved without exercising i18n interpolation — the corrected verdict below is the operative one.

Jerry-Xin (dismissed):
### Summary
In scope for `Mininglamp-OSS/octo-web`. The deny-reason dialog feature is well-structured (deny interception is double-gated by `isDocsDenyAction` + the server-declared `deny_reason` input id, old cards fall back to direct submit, `performSubmit` is a clean additions-only extraction, and the empty-reason keep-open path via the rejected `wkConfirm` `onOk` promise is correct). But there is one real, user-visible i18n blocker.

### 🔴 Blocking

- **i18n interpolation variables passed at the wrong level → literal `{{placeholders}}` shown to users.** `denyReasonDialog.tsx` calls `t()` with interpolation vars at the top level of the options object:
  - `:47` `t("base.message.interactiveCard.denyDialog.descWithContext", { actor: props.ctx.actorName, title: props.ctx.docTitle })`
  - `:56` `t("base.message.interactiveCard.denyDialog.requestNo", { no })`
  - `:79` `t("base.message.interactiveCard.denyDialog.counter", { max: MAX_REASON })`

  But this repo's `t()` reads interpolation values **only from `options.values`**: `I18nService.ts:133` `t(key, options?)` returns `interpolate(message, options?.values)` (`:140`), and `interpolate` (`:43-45`) substitutes `{{key}}` from that `values` map. Because these calls put `actor`/`title`/`no`/`max` directly on `options` (not under `options.values`), `options.values` is `undefined`, so no substitution occurs and the UI renders literal `{{actor}}` / `{{title}}` / `{{no}}` / `{{max}}`. The character-counter string (`最多 {{max}} 个字`) breaks on **every** dialog open; the contextual description and request-number break whenever they are shown.

  Fix: wrap each interpolated call in `{ values: { ... } }`, e.g. `t("...counter", { values: { max: MAX_REASON } })`, matching the established pattern used elsewhere in the codebase.

  The existing tests don't catch this because they mock `t: (key) => key`, which is blind to interpolation. Adding one test that exercises the real `interpolate` (asserting the rendered string contains the substituted value, not `{{...}}`) would prevent regressions.

### 💬 Non-blocking
- Consider an `opening` guard flag to prevent a rapid double-click from opening multiple dialogs (raised by another reviewer).

### ✅ Cleared
- Deny interception double-gated; backward-compat fallback to direct submit; `performSubmit` pure extraction; empty-reason keep-open (rejected `onOk` promise, no `destroy()`); `base.` namespace correct; en/zh key symmetry; no XSS in the reason display path.

lml2468 (approved):
## 🐙 PR #858 Re-Review — `feat(interactive-card): docs access-request deny-reason dialog`

**Head:** `78b06acbfe66` (was `862bb2669319`) · **Prior verdict:** REQUEST_CHANGES (i18n interpolation blocker)

### 🔴 → ✅ The blocker is fixed
Commit `78b06acbf fix(interactive-card): pass i18n interpolation vars under 'values'` corrects all three call sites in `denyReasonDialog.tsx`:
- `:48-50` `descWithContext` → `{ values: { actor, title } }`
- `:56-58` `requestNo` → `{ values: { no } }`
- `:79-81` `counter` → `{ values: { max: MAX_REASON } }`

Byte-verified against the real `t()`/`interpolate` — the dialog now substitutes `李四 / Roadmap / DOC-REQ-025 / 200` instead of rendering literal `{{…}}`.

**The test blind spot I flagged is also closed:** new `denyReasonDialogInterpolation.test.tsx` renders the dialog with the **real** i18n (deliberately not the identity mock) and asserts `text` contains no `{{` and shows the substituted values. Runs green (1/1). This is exactly the regression guard I recommended.

### Verification (local checkout @ `78b06acbfe66`)
| Check | Result |
|---|---|
| `vitest run src/Messages/InteractiveCard` | ✅ 222/222 (was 221; +1 real-`t` interpolation test) |
| new interpolation test alone | ✅ 1/1 |

### Delta since my prior review is minimal and safe
`git diff 862bb266..78b06acb` = 3 files: the i18n fix, the new regression test, and one safe refinement in `InteractiveCardCell.tsx` (`data?.` → `data.` — sound because `isDocsDenyAction` is a genuine type guard `data is Record<…>`, typecheck-confirmed by the green build). No logic change to the interception, submit, or dialog-resolve paths.

### Still-clean (carried from prior review, unchanged)
Double-gated deny interception (`isDocsDenyAction` + declared `deny_reason` input) · old-card fallback to direct submit · `performSubmit` pure extraction · async re-entrancy guards (`mounted`/`submitting`/decision re-check) · empty-reason keep-open (rejected `onOk` promise → modal stays) · `base.` namespace correct · en/zh key symmetry · reason is plain data (no XSS).

### Adversarial pass — no new issues
The fix is scoped and correct; the new test genuinely exercises the previously-blind path; no regression in the 221 pre-existing tests. The double-click debounce 🟡 (raised by peers) remains a reasonable non-blocking follow-up, not introduced or worsened here.

---

### ✅ Review Verdict: APPROVE @ `78b06acbfe66`
The one blocker from my prior REQUEST_CHANGES (i18n interpolation vars at the wrong level → literal placeholders) is fixed at all three call sites and locked by a real-`t` regression test. Spec ✅ and quality ✅, build + 222 tests green. Clean to merge.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #858 (octo-web)

## Summary

This PR adds a deny-reason dialog for document access-request cards in the interactive card framework. When an approver clicks "deny" on a docs access-request card, a modal dialog now prompts for a mandatory reason (up to 200 chars), which is then submitted as `inputs[deny_reason]` alongside the card's declared inputs. The implementation is backward-compatible: older cards that lack the `deny_reason` input declaration fall through to the original direct-submit path, avoiding a server-side 400 on undeclared inputs.

The change is well-scoped (7 files, +358/−0), with clean separation of concerns: `denyReasonDialog.tsx` owns the dialog logic and type guard, `InteractiveCardCell.tsx` integrates via a short-circuit before `performSubmit`, and i18n/CSS are aligned across both locales. Test coverage includes both unit tests for the dialog contract and a regression test that catches the i18n interpolation-variable placement bug (fixed in the last commit of this PR).

## Verification

Static analysis only at head `78b06acb`; build and tests not executed in this environment.

- ✅ **Backward compatibility** — `InteractiveCardCell.tsx:293–295`: the `card.getAllInputs().some(input => input.id === DOCS_DENY_REASON_INPUT_ID)` guard ensures old cards without the declared input skip the dialog entirely, falling through to `performSubmit(card, actionId, null)` at `:313`.
- ✅ **i18n interpolation** — `denyReasonDialog.tsx:51–54,60–62,82–84`: all `t()` calls pass interpolation variables under `{ values: { ... } }`, matching `I18nService.ts:140` which extracts `options?.values`. The regression test `denyReasonDialogInterpolation.test.tsx` uses the real `t()` (not mocked) and asserts no literal `{{...}}` in rendered output.
- ✅ **Locale key parity** — Both `en-US.json` and `zh-CN.json` define all 11 `denyDialog.*` keys with matching interpolation variables (`{{actor}}`, `{{title}}`, `{{no}}`, `{{max}}`).
- ✅ **Dialog lifecycle** — `denyReasonDialog.tsx:100–121`: `settled` flag prevents double-resolution. Empty-reason path returns `Promise.reject()` which `wkConfirm` (`confirm.tsx:43–47`) catches to keep the dialog open (resetting pending state). Cancel resolves `null`; valid reason resolves trimmed string.
- ✅ **Post-dialog liveness checks** — `InteractiveCardCell.tsx:304–307`: after the dialog resolves, the `.then()` callback re-checks `mounted`, `submitting`, and `computeState().decision` before calling `performSubmit`, preventing submission on a stale/unmounted card.
- ✅ **Input collection** — `performSubmit` at `:325–327` calls `collectCardInputs(card)` (which iterates `card.getAllInputs()`), then merges `extraInputs` via `Object.assign`. The `deny_reason` key is only injected when the card declares it (gated at `:294`), so `validateCardInputs` at `:328` won't reject it as undeclared.

## Findings

### Nit — Missing `.catch()` on dialog promise chain (`InteractiveCardCell.tsx:298`)

The `void openDocsDenyReasonDialog(...).then(...)` chain at `:298–308` has no `.catch()`. Currently safe because `openDocsDenyReasonDialog` always resolves (never rejects to the caller), and the `.then()` callback only does property reads that cannot throw synchronously. However, if `performSubmit` were to throw synchronously in a future refactor, `void` would silently convert it to an unhandled rejection.

```ts
// Suggested defensive addition:
}).catch((err) => console.error("deny-reason dialog chain error", err));
```

Not a change I'd block on — flagging for resilience.

- **wkConfirm contract** — `confirm.tsx:42–48`: when `onOk` returns a thenable, the confirm wrapper calls `.then(() => destroy()).catch(() => updatePending(null))`. The dialog's `onOk` returns `undefined` on valid input (triggers `destroy`) and `Promise.reject` on empty (triggers `updatePending(null)`, keeping dialog open). Contract is correctly consumed.
- **Double-click safety** — `handleSubmit` returns at `:311` after launching the dialog, so `performSubmit` is not called twice. The `this.submitting` guard at `:280` prevents re-entry if the user somehow triggers another submit while the dialog is open.
- **`submitGen` integrity** — `performSubmit` increments `this.submitGen` at `:357` and all async callbacks check `gen !== this.submitGen`. The dialog-to-submit path preserves this contract since `performSubmit` is called from within `.then()`.
- **CSS scoping** — All 6 new CSS classes (`wk-docs-deny-dialog*`) are additive; no existing styles are modified. Class names are sufficiently specific to avoid collisions in the global stylesheet.
- **Test mock hygiene** — `denyReasonDialog.test.tsx` uses `vi.hoisted` for mock references (avoiding hoisting issues), mocks only `wkConfirm`/`Semi`/`i18n` as needed, and tests the dialog contract (resolve on confirm, reject on empty, null on cancel). The interpolation test deliberately does *not* mock `i18n` to catch the `values` placement bug.

## Data Flow Tracing

|---|---|---|
| `data.doc_title`, `data.actor`, `data.request_id` | `action.data` from server card JSON → `asString()` type-safe extraction at `:297` | ✅ `asString` returns `undefined` for non-strings; dialog gracefully falls back to generic `desc` key |
| `box.reason` (deny reason text) | `TextArea.onChange` → `DenyReasonBody.onChange` → `box.reason` closure var → `onOk` trims → dialog resolves | ✅ Trimming at `:113` ensures whitespace-only input is rejected |
| `DOCS_DENY_REASON_INPUT_ID` → `inputs` map | Constant `"deny_reason"` merged via `Object.assign(inputs, extraInputs)` at `:327` → `submitCardAction` at `:361` | ✅ Only merged when card declares the input (gated at `:294`) |
| `card.getAllInputs()` for backward-compat check | Adaptive Card SDK `getAllInputs()` returns only declared inputs in current card frame | ✅ Old cards without `deny_reason` input → `.some()` returns false → direct submit |

- **C1 (Dual-path parity)** — Clear. The deny path opens a dialog and calls `performSubmit` with extra inputs; the approve path and all other actions call `performSubmit` with `null` extra inputs. Both paths go through the same `performSubmit` → `collectCardInputs` → `validateCardInputs` → `submitCardAction` pipeline. No asymmetry.
- **C2 (Control-flow ordering / nested reuse)** — Clear. `performSubmit` is called from exactly two sites in `handleSubmit` (dialog path at `:308`, direct path at `:313`), both mutually exclusive via the early `return` at `:311`.
- **C3 (Authorization boundary)** — N/A. This is a client-side UX gate; the server validates inputs independently.
- **C4 (Authorization lifecycle)** — N/A. No auth/permission changes.
- **C5 (Build/runtime parity)** — N/A. Pure TypeScript/React change; no build/packaging concerns.
- **C6 (Governance/docs)** — N/A. No governance or policy changes.

---

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The deny-reason dialog is correctly integrated with proper backward compatibility, safe lifecycle handling, and good test coverage. One defensive-coding nit (missing `.catch()`) is non-blocking.

Jerry-Xin (approved):
### Summary
Re-reviewed at `78b06acbfe66`. My prior blocker is fixed with a proper regression test, and this delta (ahead 1 / behind 0) introduces no new issues. Approving.

### ✅ Prior blocker — fixed (byte-verified)
- **i18n interpolation** (`denyReasonDialog.tsx`): all three interpolated `t()` calls now wrap their variables in `{ values: { … } }` — `descWithContext` (`{ values: { actor, title } }`), `requestNo` (`{ values: { no } }`), and `counter` (`{ values: { max: MAX_REASON } }`). Since `t()` reads interpolation vars from `options.values` (`I18nService.ts:140`), the dialog now substitutes correctly instead of rendering literal `{{actor}}` / `{{title}}` / `{{no}}` / `{{max}}`.

### ✅ Regression test added (locks the fix)
- `denyReasonDialogInterpolation.test.tsx` deliberately uses the **real** `t`/`interpolate` (not the `t: (key) => key` mock that hid the original bug), renders the dialog content, and asserts the output does not contain `{{` and does contain the substituted values (`李四` / `Roadmap` / `DOC-REQ-025` / `200`). This is exactly the guard that would have caught the original defect.

### ✅ Bonus cleanups (nice)
- `isDocsDenyAction` is now a real type guard (`data is Record<string, unknown>`), so callers drop the defensive `?.` — cleaner and matches the narrowing.
- Removed the dead `className: "wk-docs-deny-modal"` that had no matching CSS rule.

Feature is scoped to `packages/dmworkbase/src/Messages/InteractiveCard`. Clean turnaround — LGTM.

yujiawei (approved):
# Code Review — PR #858 (octo-web)

Reviewed at head `78b06acbfe66` against merge-base `2467b307`. Build (`turbo run build --filter=@octo/web`) passes; `vitest run src/Messages/InteractiveCard` passes 222/222, including the new real-`t` interpolation regression test.

## 1. Spec compliance — ✅

The diff does exactly what the description states, no more and no less:

- **Deny interception is correctly double-gated** at `InteractiveCardCell.tsx:291-293` — `isDocsDenyAction(data)` **and** the current frame actually declaring the `deny_reason` input via `card.getAllInputs()`. Cards without the hidden input fall through to the original direct-submit path, so backward compatibility holds.
- **`performSubmit` is a pure extraction** of the prior submit body (the `.tsx` change is additions-only; the original validate → collect → channel-resolve → submit → timeout/error flow is preserved verbatim, with the only new line being `Object.assign(inputs, extraInputs)` at `:325`). Approve and all non-docs actions submit unchanged.
- No over-build, no missing requirement, no deviation from the stated contract.

## 2. Code quality — Approved

**Verified correct:**

- **Single-submit is guaranteed under the async dialog.** After the dialog resolves, `:304` re-checks `this.mounted` / `this.submitting` and `:305-306` re-checks that the decision is still an interactive card before submitting. Even if two dialogs were opened, only the first resolution submits.
- **Empty-reason keep-open contract is correct.** `denyReasonDialog.tsx:112-113` returns a rejected promise; `WKModal/confirm.tsx:47` does `.catch(() => updatePending(null))`, which keeps the modal open without destroying it — and swallows the rejection, so there is no unhandled-rejection leak.
- **i18n resolves correctly.** Keys are stored as `message.interactiveCard.denyDialog.*` and registered under the `base` namespace; `I18nService`'s `withNamespace` auto-prefixes `base.`, so the `t("base.…")` call sites resolve. The interpolation vars are correctly passed under `{ values: { … } }`, matching `I18nService.interpolate(message, options?.values)`. A dedicated regression test renders the dialog with the **real** `t` and asserts no literal `{{…}}` remain.
- en/zh key parity (11/11), reason is plain data (no `dangerouslySetInnerHTML`), and `TextArea maxLength=200` keeps the payload well within the client-side size validation.

**Non-blocking findings (P2 / nits):**

- **P2 — Deny button can stack modals on rapid double-click.** `this.submitting` is only set inside `performSubmit`, not when the dialog opens (`:294`). The `handleSubmit` guard at `:279` (`if (this.submitting) return`) therefore does not cover the dialog-open window, so a second click on 拒绝 while the first dialog is open opens a second stacked dialog. It is a UX glitch only — the post-resolution `this.submitting` check still guarantees a single network submit. Optional: set a local "dialog open" flag (or reuse `submitting`) before `openDocsDenyReasonDialog`.
- **P2 — Imperative modal can outlive the cell.** `wkConfirm` portals outside the React tree and `openDocsDenyReasonDialog` returns only a promise with no handle to close it. If the cell unmounts while the dialog is open (channel switch, message deleted), the dialog lingers as a zombie. No bad submit results (guarded by `!this.mounted`), but consider returning a dismiss handle and closing it in `componentWillUnmount`.
- **Nit — type guard on primitive `data`.** `isDocsDenyAction` does `!!data && data.owner === …`. Adaptive Cards `Action.Submit.data` can be a primitive string; property access is safe in JS, but a leading `typeof data === "object"` would make the invariant explicit.

## 3. Verdict

No P0/P1 issues. Spec is met and the implementation is correct, backward-compatible, and well-tested. The three items above are UX/maintainability polish that do not block merge.

Approving. Worth a quick follow-up for the double-click / unmount modal-lifecycle nits if this dialog pattern gets reused.

PR #865, opened 2026-07-17
refactor(summary): extract version history into standalone SummaryVersionHistory component

quanming1:
## Summary

Extract version history rendering logic from `SummaryDetailPage.tsx` (3480 lines) into a standalone `SummaryVersionHistory.tsx` component.

## Changes

- **Rewrote `SummaryVersionHistory.tsx`**: from read-only Timeline to full strip layout
- **New props**: `versions`, `versionsLoading`, `currentVersion`, `restoringVersionId`, `canRestore`, `onViewVersion`, `onRestoreVersion`
- **Inlined `formatVersionOperation`**: displays operation type (initial generation / regenerated / scheduled generation / adjusted by feedback / manual edit / restored version)
- Supports `scheduled_generate` tag, current version tag, and operation notes
- **Permission control**: restore button hidden when `canRestore` is false
- `SummaryDetailPage.tsx` already calls this component with matching props (`renderVersionHistory()` at line 2133, `renderPersonalVersionHistory()` at line 2152)

## Motivation

`SummaryDetailPage.tsx` is too large (3480 lines) with version history rendering logic inlined. Extracting it into a standalone component:
- Reduces page file complexity
- Makes version history UI logic independently maintainable and testable
- Aligns with the octo-web-v2 refactored component separation design

## Testing

- TypeScript type check passed (existing TS7016/TS7031 errors from React 17 type declarations are pre-existing, not introduced by this PR)
- All i18n keys already exist (`summary.detail.recentVersions`, `versionScheduledTaskTag`, `currentVersion`, `viewVersion`, `restoreVersion`, etc.)
- CSS class names reuse existing styles (`.summary-version-strip`, `.summary-version-item`, `.summary-version-actions`, etc.)
- `SummaryDetailPage.tsx`'s `renderVersionHistory()` and `renderPersonalVersionHistory()` already use this component's interface

### DIFF
--- a/packages/dmworksummary/src/components/SummaryVersionHistory.tsx
+++ b/packages/dmworksummary/src/components/SummaryVersionHistory.tsx
@@ -1,68 +1,109 @@
 import React from "react";
-import { Timeline, Button, Tag } from "@douyinfe/semi-ui";
+import { Button, Tag } from "@douyinfe/semi-ui";
+import { IconHistory } from "@douyinfe/semi-icons";
 import { useI18n } from "@octo/base";
-import type { SummaryResult } from "../types/summary";
-import { formatDate } from "../utils/summaryHelpers";
+import type { SummaryVersionItem } from "../types/summary";
 
 interface SummaryVersionHistoryProps {
-    versions: SummaryResult[];
+    versions: SummaryVersionItem[];
+    versionsLoading: boolean;
     currentVersion: number;
-    onSelectVersion: (version: number) => void;
+    restoringVersionId: number | null;
+    canRestore: boolean;
+    onViewVersion: (version: SummaryVersionItem) => void;
+    onRestoreVersion: (version: SummaryVersionItem) => void;
+}
+
+function formatVersionOperation(
+    version: SummaryVersionItem,
+    t: (key: string, opts?: { values?: Record<string, string | number> }) => string,
+): string {
+    if ((version.operation_type || "generate") === "generate") {
+        return t("summary.detail.versionInitialGenerate");
+    }
+    const key = `summary.detail.versionOperation.${version.operation_type || "generate"}`;
+    const label = t(key);
+    return label === key ? t("summary.detail.versionOperation.generate") : label;
+}
+
+function formatVersionOperationNote(
+    version: SummaryVersionItem,
+    t: (key: string, opts?: { values?: Record<string, string | number> }) => string,
+): string {
+    const note = (version.operation_note || "").trim();
+    if (note) return note;
+    if ((version.operation_type || "generate") === "generate") {
+        return t("summary.detail.versionInitialGenerateDesc");
+    }
+    if (version.operation_type === "restore" && version.parent_result_id) {
+        return t("summary.detail.versionRestoreFromResult", { values: { id: version.parent_result_id } });
+    }
+    return formatVersionOperation(version, t);
 }
 
 const SummaryVersionHistory: React.FC<SummaryVersionHistoryProps> = ({
     versions,
+    versionsLoading,
     currentVersion,
-    onSelectVersion,
+    restoringVersionId,
+    canRestore,
+    onViewVersion,
+    onRestoreVersion,
 }) => {
     const { t } = useI18n();
 
-    if (!versions || versions.length === 0) {
-        return <div className="summary-version-empty">{t("summary.versionHistory.empty")}</div>;
-    }
-
-    const sorted = [...versions].sort((a, b) => b.version - a.version);
+    if (versionsLoading || !versions || versions.length <= 1) return null;
 
     return (
-        <div className="summary-version-history">
-            <Timeline>
-                {sorted.map((v) => {
-                    const isCurrent = v.version === currentVersion;
+        <div className="summary-version-strip">
+            <div className="summary-version-strip-title">
+                <IconHistory size="small" />
+                <span>{t("summary.detail.recentVersions")}</span>
+                <span className="summary-version-strip-hint">{t("summary.detail.recentVersionsLimitHint")}</span>
+            </div>
+            <div className="summary-version-list">
+                {versions.slice(0, 3).map((version) => {
+                    const isCurrent = version.version === currentVersion;
                     return (
-                        <Timeline.Item key={v.version} color={isCurrent ? "blue" : "grey"}>
-                            <div className="summary-version-item">
-                                <div className="summary-version-header">
-                                    <span>{t("summary.common.version", { values: { version: v.version } })}</span>
-                                    {isCurrent && (
-                                        <Tag color="blue" size="small" style={{ marginLeft: 8 }}>
-                                            {t("summary.common.current")}
-                                        </Tag>
+                        <div key={version.result_id} className="summary-version-item">
+                            <div className="summary-version-body">
+                                <div className="summary-version-main">
+                                    <span className="summary-version-number">
+                                        {t("summary.common.version", { values: { version: version.version } })}
+                                    </span>
+                                    {isCurrent && <Tag size="small" color="blue">{t("summary.detail.currentVersion")}</Tag>}
+                                    {version.operation_type === "scheduled_generate" && (
+                                        <Tag size="small" color="green">{t("summary.detail.versionScheduledTaskTag")}</Tag>
+                                    )}
+                                    {version.operation_type !== "scheduled_generate" && (
+                                        <span className="summary-version-operation">{formatVersionOperation(version, t)}</span>
                                     )}
                                 </div>
-                                <div className="summary-version-meta">
-                                    {t("summary.versionHistory.meta", {
-                                        values: {
-                                            time: formatDate(v.generated_at),
-                                            count: v.total_msg_count,
-                                            model: v.model_version,
-                                        },
-                                    })}
-                                </div>
-                                {!isCurrent && (
+                                <div className="summary-version-note">{formatVersionOperationNote(version, t)}</div>
+                            </div>
+                            <div className="summary-version-actions">
+                                <Button
+                                    size="small"
+                                    theme="borderless"
+                                    onClick={() => onViewVersion(version)}
+                                >
+                                    {t("summary.detail.viewVersion")}
+                                </Button>
+                                {!isCurrent && canRestore && (
                                     <Button
                                         size="small"
                                         theme="borderless"
-                                        onClick={() => onSelectVersion(v.version)}
-                                        style={{ marginTop: 4 }}
+                                        loading={restoringVersionId === version.result_id}
+                                        onClick={() => onRestoreVersion(version)}
                                     >
-                                        {t("summary.versionHistory.viewVersion")}
+                                        {t("summary.detail.restoreVersion")}
                                     </Button>
                                 )}
                             </div>
-                        </Timeline.Item>
+                        </div>
                     );
                 })}
-            </Timeline>
+            </div>
         </div>
     );
 };
--- a/packages/dmworksummary/src/components/SummaryVersionHistory.tsx
+++ b/packages/dmworksummary/src/components/SummaryVersionHistory.tsx
@@ -48,10 +48,24 @@ const SummaryVersionHistory: React.FC<SummaryVersionHistoryProps> = ({
         return label === key ? t("summary.detail.versionOperation.generate") : label;
     };
 
+    const formatVersionOperationNote = (version: SummaryVersionItem): string => {
+        const note = (version.operation_note || "").trim();
+        if (note) return note;
+        if ((version.operation_type || "generate") === "generate") {
+            return t("summary.detail.versionInitialGenerateDesc");
+        }
+        if (version.operation_type === "restore" && version.parent_result_id) {
+            return t("summary.detail.versionRestoreFromResult", {
+                values: { id: version.parent_result_id },
+            });
+        }
+        return formatVersionOperation(version);
+    };
+
     return (
         <div className="summary-version-strip">
             <div className="summary-version-strip-title">
-                <IconHistory />
+                <IconHistory size="small" />
                 <span>{t("summary.detail.recentVersions")}</span>
                 <span className="summary-version-strip-hint">
                     {t("summary.detail.recentVersionsLimitHint")}
@@ -86,11 +100,9 @@ const SummaryVersionHistory: React.FC<SummaryVersionHistoryProps> = ({
                                         </span>
                                     )}
                                 </div>
-                                {version.operation_note && (
-                                    <div className="summary-version-note">
-                                        {version.operation_note}
-                                    </div>
-                                )}
+                                <div className="summary-version-note">
+                                    {formatVersionOperationNote(version)}
+                                </div>
                             </div>
                             <div className="summary-version-actions">
                                 <Button
--- a/packages/dmworksummary/src/components/SummaryVersionHistory.tsx
+++ b/packages/dmworksummary/src/components/SummaryVersionHistory.tsx
@@ -14,33 +14,15 @@ interface SummaryVersionHistoryProps {
     onRestoreVersion: (version: SummaryVersionItem) => void;
 }
 
-function formatVersionOperation(
-    version: SummaryVersionItem,
-    t: (key: string, opts?: { values?: Record<string, string | number> }) => string,
-): string {
-    if ((version.operation_type || "generate") === "generate") {
-        return t("summary.detail.versionInitialGenerate");
-    }
-    const key = `summary.detail.versionOperation.${version.operation_type || "generate"}`;
-    const label = t(key);
-    return label === key ? t("summary.detail.versionOperation.generate") : label;
-}
-
-function formatVersionOperationNote(
-    version: SummaryVersionItem,
-    t: (key: string, opts?: { values?: Record<string, string | number> }) => string,
-): string {
-    const note = (version.operation_note || "").trim();
-    if (note) return note;
-    if ((version.operation_type || "generate") === "generate") {
-        return t("summary.detail.versionInitialGenerateDesc");
-    }
-    if (version.operation_type === "restore" && version.parent_result_id) {
-        return t("summary.detail.versionRestoreFromResult", { values: { id: version.parent_result_id } });
-    }
-    return formatVersionOperation(version, t);
-}
-
+/**
+ * 总结版本历史 strip 组件。
+ *
+ * 从 SummaryDetailPage.renderVersionHistory() / renderPersonalVersionHistory()
+ * 提取为独立组件，保留原有 strip 布局、CSS 类名和交互行为。
+ *
+ * 数据获取和 API 调用由父组件（SummaryDetailPage）负责，
+ * 本组件只负责渲染和回调。
+ */
 const SummaryVersionHistory: React.FC<SummaryVersionHistoryProps> = ({
     versions,
     versionsLoading,
@@ -52,34 +34,63 @@ const SummaryVersionHistory: React.FC<SummaryVersionHistoryProps> = ({
 }) => {
     const { t } = useI18n();
 
-    if (versionsLoading || !versions || versions.length <= 1) return null;
+    if (!versions || versionsLoading || versions.length <= 1) {
+        return null;
+    }
+
+    const formatVersionOperation = (version: SummaryVersionItem): string => {
+        const opType = version.operation_type || "generate";
+        if (opType === "generate") {
+            return t("summary.detail.versionInitialGenerate");
+        }
+        const key = `summary.detail.versionOperation.${opType}`;
+        const label = t(key);
+        return label === key ? t("summary.detail.versionOperation.generate") : label;
+    };
 
     return (
         <div className="summary-version-strip">
             <div className="summary-version-strip-title">
-                <IconHistory size="small" />
+                <IconHistory />
                 <span>{t("summary.detail.recentVersions")}</span>
-                <span className="summary-version-strip-hint">{t("summary.detail.recentVersionsLimitHint")}</span>
+                <span className="summary-version-strip-hint">
+                    {t("summary.detail.recentVersionsLimitHint")}
+                </span>
             </div>
             <div className="summary-version-list">
                 {versions.slice(0, 3).map((version) => {
                     const isCurrent = version.version === currentVersion;
+                    const isRestoring = restoringVersionId === version.result_id;
                     return (
                         <div key={version.result_id} className="summary-version-item">
                             <div className="summary-version-body">
                                 <div className="summary-version-main">
                                     <span className="summary-version-number">
-                                        {t("summary.common.version", { values: { version: version.version } })}
+                                        {t("summary.common.version", {
+                                            values: { version: version.version },
+                                        })}
                                     </span>
-                                    {isCurrent && <Tag size="small" color="blue">{t("summary.detail.currentVersion")}</Tag>}
+                                    {isCurrent && (
+                                        <Tag color="blue" size="small">
+                                            {t("summary.detail.currentVersion")}
+                                        </Tag>
+                                    )}
                                     {version.operation_type === "scheduled_generate" && (
-                                        <Tag size="small" color="green">{t("summary.detail.versionScheduledTaskTag")}</Tag>
+                                        <Tag color="green" size="small">
+                                            {t("summary.detail.versionScheduledTaskTag")}
+                                        </Tag>
                                     )}
                                     {version.operation_type !== "scheduled_generate" && (
-                                        <span className="summary-version-operation">{formatVersionOperation(version, t)}</span>
+                                        <span className="summary-version-operation">
+                                            {formatVersionOperation(version)}
+                                        </span>
                                     )}
                                 </div>
-                                <div className="summary-version-note">{formatVersionOperationNote(version, t)}</div>
+                                {version.operation_note && (
+                                    <div className="summary-version-note">
+                                        {version.operation_note}
+                                    </div>
+                                )}
                             </div>
                             <div className="summary-version-actions">
                                 <Button
@@ -93,7 +104,7 @@ const SummaryVersionHistory: React.FC<SummaryVersionHistoryProps> = ({
                                     <Button
                                         size="small"
                                         theme="borderless"
-                                        loading={restoringVersionId === version.result_id}
+                                        loading={isRestoring}
                                         onClick={() => onRestoreVersion(version)}
                                     >
                                         {t("summary.detail.restoreVersion")}
--- a/packages/dmworksummary/src/components/__tests__/SummaryVersionHistory.test.tsx
+++ b/packages/dmworksummary/src/components/__tests__/SummaryVersionHistory.test.tsx
@@ -1,0 +1,220 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { describe, it, expect, vi } from 'vitest';
+import SummaryVersionHistory from '../SummaryVersionHistory';
+import type { SummaryVersionItem } from '../../types/summary';
+
+vi.mock('@octo/base', async () => {
+    const actual = await vi.importActual<Record<string, unknown>>('../../__mocks__/dmworkBase');
+    return {
+        ...actual,
+        useI18n: () => ({
+            t: (key: string, opts?: { values?: Record<string, string | number> }) => {
+                const map: Record<string, string> = {
+                    'summary.detail.recentVersions': '最近版本',
+                    'summary.detail.recentVersionsLimitHint': '仅展示最近 3 个版本',
+                    'summary.detail.currentVersion': '当前版本',
+                    'summary.detail.versionScheduledTaskTag': '定时任务',
+                    'summary.detail.viewVersion': '查看',
+                    'summary.detail.restoreVersion': '恢复此版本',
+                    'summary.detail.versionInitialGenerate': '初始生成',
+                    'summary.detail.versionInitialGenerateDesc': '第一次生成的基线版本，可用于恢复到微调前内容。',
+                    'summary.detail.versionRestoreFromResult': '恢复自历史结果 #{{id}}',
+                    'summary.detail.versionOperation.generate': '初始生成',
+                    'summary.detail.versionOperation.regenerate': '重新生成',
+                    'summary.detail.versionOperation.scheduled_generate': '定时生成',
+                    'summary.detail.versionOperation.refine': '按意见调整',
+                    'summary.detail.versionOperation.manual_edit': '手动编辑',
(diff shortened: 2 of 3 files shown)


lml2468 (dismissed):
## 🐙 PR #865 Review — `refactor(summary): extract version history into standalone SummaryVersionHistory component`

**Head:** `6e3af043f807` · **Base:** `main` (merge-base `50b6165b`) · +108 −150 / 2 files
**Intent:** Extract the version-history strip out of `SummaryDetailPage.renderVersionHistory()` / `renderPersonalVersionHistory()` into a standalone `SummaryVersionHistory` component; parent keeps data/API, component renders + calls back. The PR says it "保留原有 strip 布局、CSS 类名和交互行为."

### Verification (local checkout @ `6e3af043f807`)
| Check | Result |
|---|---|
| `dmworksummary` full suite | ✅ 386/386 |

### Gate 1 — Spec compliance: ⚠️ one behavioral deviation from the "preserve behavior" claim
The extraction is faithful for layout, CSS classes, and the view/restore actions, and both call sites (team + personal) are wired with the right callbacks/permissions. But it **drops one behavior** the PR claims to preserve (below). Also note: the PR **description is empty** ("## Summary" with no body) — for a refactor that changes user-visible output, a one-line "what/why" would help.

### Gate 2 — Code quality: clean apart from the note regression

**🟡 P2 — version-note fallback text is lost (`SummaryVersionHistory.tsx:89`)**
The original always rendered `.summary-version-note` via `formatVersionOperationNote(version)`, which had fallbacks when `operation_note` was empty:
- `generate` type → `versionInitialGenerateDesc` ("第一次生成的基线版本…")
- `restore` with `parent_result_id` → `versionRestoreFromResult` ("恢复自历史结果 #id")
- else → the operation label

The new component renders only `{version.operation_note && (<div className="summary-version-note">…)}`. Since `operation_note` is **optional** (`types/summary.ts:110`) and commonly empty for `generate`/`restore` versions, those rows now show **no note** where they previously showed descriptive text. `formatVersionOperationNote` is deleted (grep-confirmed zero remaining refs), and `versionInitialGenerateDesc` / `versionRestoreFromResult` are now **dead i18n keys** — corroborating that the behavior was dropped, not relocated.

Not a correctness/data/security bug — informational UI text — so P2, not a hard blocker. But it contradicts the "保留…交互行为" claim, and no test covers version-history rendering so it's invisible to the suite. **Please confirm intentional; if not, restore the fallback** (move `formatVersionOperationNote`'s logic into the component, or pass the resolved note down).

**Minor (non-blocking):** the added `|| 0` defaults on `currentVersion` are harmless; the `formatVersionOperation` fallback logic is preserved correctly in the component.

### Adversarial pass
Extraction fidelity ✅ (both call sites, callbacks, permissions, `slice(0,3)`, `result_id` keys preserved) · build ✅ · 386 tests ✅ · the only behavioral delta is the note-fallback drop above · `||0` defaults harmless · no security/data surface (pure presentational refactor).

---

### ✅ Review Verdict: APPROVE (with one P2 to confirm/fix)

Clean, net-negative extraction — build green, 386 tests pass, faithful layout/CSS/action parity. One P2: the version-note fallback text (`versionInitialGenerateDesc` / `versionRestoreFromResult`) is silently dropped for versions without an `operation_note`, which contradicts the "preserve behavior" claim and leaves two i18n keys dead. Please confirm that's intentional or restore the fallback before merge; also add a PR description. Not blocking on it since it's informational-only UI text, but worth a quick fix while you're here.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #865 (octo-web)

## Summary

Multi-topic PR bundling four changes: (1) extract summary version history into a standalone `SummaryVersionHistory` component, (2) add a deny-reason dialog for docs access-request cards, (3) forward the live board title to "forward to chat" instead of the stale prop, and (4) refetch the docs recent list when the nav entry is re-activated. The deny dialog, board title, and docs refetch changes are solid and well-tested. The summary version-history extraction has a behavioral regression — a fallback-text function was silently dropped, leaving version notes blank where users previously saw contextual descriptions.

## Verification

- ✅ **Diff scope** — 13 files, +694/−153 across four packages. All changes match the commit messages.
- ✅ **Deny dialog contract** — `isDocsDenyAction` type guard, `openDocsDenyReasonDialog` promise semantics, and `wkConfirm` reject-to-keep-open all traced through to `InteractiveCardCell.performSubmit`. Post-dialog guards (`this.mounted`, `this.submitting`, re-check `computeState().decision`) prevent stale submissions.
- ✅ **BoardShell title lift** — `DocTitle` supports `onTitleLoaded` (confirmed at `packages/docs/src/editor/EditorShell.tsx:115,122,146`). `currentTitle` seeded from prop, updated on fetch and rename. `onForwardToChat` dependency array correctly uses `currentTitle` instead of `title`. Empty-title fallback to `docs.board.untitled` matches test expectations.
- ✅ **DocsHome reload token** — `setListReloadToken((n) => n + 1)` in `onNavMenuActivated('docs')` callback wires into `useDocsView`'s `[space, folder, reloadToken]` effect dependency at `packages/docs/src/pages/useDocsView.ts:336`. Filter-aware refetch verified.
- ✅ **i18n parity** — `en-US.json` and `zh-CN.json` add identical key sets (12 deny-dialog keys each). No missing translations.
- ✅ **Backward compatibility** — deny dialog only opens when `card.getAllInputs()` declares `deny_reason`, so old cards without the hidden input fall through to the direct-submit path.

## Findings

One P1 blocker in the summary version-history extraction; one nit.

### P1 — `formatVersionOperationNote` fallback logic silently dropped (`packages/dmworksummary/src/components/SummaryVersionHistory.tsx:89`)

The old inline code in `SummaryDetailPage` rendered the version note via `formatVersionOperationNote`, which always returned a non-empty string:

1. If `operation_note` had content → use it directly.
2. If `operation_type` is `"generate"` → fall back to `t("summary.detail.versionInitialGenerateDesc")` ("The first generated baseline version. You can restore it before later adjustments.").
3. If `operation_type` is `"restore"` with `parent_result_id` → fall back to `t("summary.detail.versionRestoreFromResult")`.
4. Otherwise → the formatted operation label.

The new component renders `{version.operation_note && (<div>…</div>)}` only. When `operation_note` is undefined (common for initial generates and restores — the field is optional in `SummaryVersionItem`), the note div is not rendered at all. Users lose contextual descriptions they previously saw for every version entry.

This contradicts the component's own doc comment: "保留原有 strip 布局、CSS 类名和交互行为" — the interaction behavior is not preserved.

```tsx
// Current (broken):
{version.operation_note && (
    <div className="summary-version-note">{version.operation_note}</div>
)}

// Fix — restore the fallback chain, e.g.:
const note = formatVersionOperationNote(version);
<div className="summary-version-note">{note}</div>
```

Move `formatVersionOperationNote` from the old diff into this component (or accept a `formatNote` callback prop from the parent) so the fallback logic is preserved.

### Nit — `IconHistory` lost `size="small"` prop (`packages/dmworksummary/src/components/SummaryVersionHistory.tsx:57`)

The old `SummaryDetailPage` rendered `<IconHistory size="small" />`. The extracted component uses `<IconHistory />` without size, which defaults to medium (20px vs 16px). Minor visual discrepancy with the original layout.

## Data-flow tracing

- **Deny dialog → performSubmit**: `openDocsDenyReasonDialog()` resolves `string | null`. Null (cancel) → no submit. Non-null string → guarded by `this.mounted`, `this.submitting`, and `computeState().decision` recheck before `performSubmit(card, actionId, {deny_reason: reason})`. `performSubmit` increments `submitGen` internally, so stale dialog resolves cannot clobber an in-flight submission.
- **`card.getAllInputs()` backward compat gate**: Old cards (pre-this-PR) lack the `deny_reason` hidden input declaration. The `.some(input => input.id === DOCS_DENY_REASON_INPUT_ID)` check ensures old cards skip the dialog and submit directly — no 400 from undeclared inputs.
- **BoardShell title**: `DocTitle.onTitleLoaded` fires on mount with the fetched title → `setCurrentTitle` → `onForwardToChat` reads `currentTitle` (not the stale `title` prop). `onSaved` wraps both `setCurrentTitle` and the original `onTitleSaved` callback, preserving the parent notification.
- **DocsHome reload**: `onNavMenuActivated('docs')` → `setListReloadToken(n+1)` → `useDocsView` effect `[space, folder, reloadToken]` re-fires → re-sends each tab's remembered q/creators/types. Non-docs menu activations are filtered out by the `'docs'` key.

## Blind-spot checklist (C1–C6)

- **C1 — Dual-path parity**: `renderVersionHistory()` and `renderPersonalVersionHistory()` both use the same `SummaryVersionHistory` component with appropriate callbacks — parity preserved. Clear.
- **C2 — Control-flow ordering**: `formatVersionOperation` is defined inside the component body and called in render. No nesting/reuse ordering issues. Clear.
- **C3 — Authorization boundary**: Deny dialog is a UX gate before the existing `submitCardAction` path; server-side authorization is untouched. N/A.
- **C4 — Authorization lifecycle**: N/A — no auth changes.
- **C5 — Build/runtime parity**: The `formatVersionOperationNote` removal is captured as Finding P1 above. No other build-vs-runtime discrepancies.
- **C6 — Governance/docs**: N/A.

## Verdict: CHANGES_REQUESTED

The deny dialog, board title forwarding, and docs nav refetch are all well-implemented with proper guards and tests. The P1 is confined to the summary version-history extraction: restoring the `formatVersionOperationNote` fallback chain (a small, localized fix) is sufficient to clear this PR.

[Octo-Q] verdict: REQUEST_CHANGES — one P1 behavioral regression in `SummaryVersionHistory` (version note fallback dropped); all other changes are sound.

yujiawei (changes requested):
# Code Review — PR #865 (octo-web)

**Scope reviewed:** `git diff 50b6165...6e3af04` (net +108/-150, 2 files). This is a pure refactor that extracts the version-history strip rendering from `SummaryDetailPage.renderVersionHistory()` / `renderPersonalVersionHistory()` into a standalone `SummaryVersionHistory` component. The component's docstring and the PR title both state the goal is to **preserve the original layout, CSS classes, and interaction behavior** ("保留原有 strip 布局、CSS 类名和交互行为").

I reviewed it against that stated goal: does the extracted component behave identically to the inline code it replaces?

## 1. Spec compliance

**Spec: ❌ — the refactor changes user-visible behavior it claims to preserve.**

- Missing: none.
- Extra: none.
- **Deviation (this is the blocker):** the old inline code always rendered a `.summary-version-note` line via a `formatVersionOperationNote()` fallback chain. The new component only renders the note when `operation_note` is truthy. The fallback chain was dropped, not migrated — so a "behavior-preserving refactor" silently changes what history entries display.

## 2. Code quality

### P1 — operation-note fallback chain lost (functional regression)

`packages/dmworksummary/src/components/SummaryVersionHistory.tsx:89-92`

```tsx
{version.operation_note && (
    <div className="summary-version-note">
        {version.operation_note}
    </div>
)}
```

The deleted `SummaryDetailPage.formatVersionOperationNote()` (previously at `SummaryDetailPage.tsx:2133`) resolved the note in four stages:

```ts
const note = (version.operation_note || "").trim();
if (note) return note;
if ((version.operation_type || "generate") === "generate")
    return t("summary.detail.versionInitialGenerateDesc");
if (version.operation_type === "restore" && version.parent_result_id)
    return t("summary.detail.versionRestoreFromResult", { values: { id: version.parent_result_id } });
return this.formatVersionOperation(version);
```

Consequences for entries with an empty/whitespace-only `operation_note`:
- **Initial-generation versions** (`operation_type = "generate"`) no longer show `summary.detail.versionInitialGenerateDesc` ("The first generated baseline version. You can restore it before later adjustments."). Since the strip only renders when `versions.length > 1`, the oldest (v1) baseline entry is exactly the one that loses its description.
- **Restore versions** without an explicit note no longer show `summary.detail.versionRestoreFromResult` ("Restored from historical result #{id}") provenance.

The version number, operation-type label, and Current/Scheduled tags are still shown, so this is display degradation rather than data loss — but it is a real, prod-visible behavior change in both the team and personal history strips, and it directly contradicts the PR's stated intent.

Two i18n keys are now orphaned (dead in both `en-US.json` and `zh-CN.json`), which is the tell that the fallback was dropped rather than intentionally removed:
- `summary.detail.versionInitialGenerateDesc`
- `summary.detail.versionRestoreFromResult`

**Fix:** re-introduce the fallback resolution (either compute a `noteText` inside the component mirroring the four stages, or pass a `formatNote` callback prop from the page so the `this.formatVersionOperation` dependency stays owned by the page), and always render the `.summary-version-note` line. Then remove the two i18n keys only if you deliberately want to drop those messages.

### P2 — icon size parity break

`packages/dmworksummary/src/components/SummaryVersionHistory.tsx:54`

Old: `<IconHistory size="small" />` (`SummaryDetailPage.tsx:2145`). New: `<IconHistory />` (default size). Minor visual regression against the "preserve original layout" goal; restore `size="small"`.

## 3. What checks out (parity verified)

- All `.summary-version-*` CSS classes referenced by the component exist in `index.css` (strip/list/item/body/main/number/operation/note/actions/strip-title/strip-hint) — layout classes preserved.
- `canRestore` correctly reproduces the old gating: team = `can_edit_team || can_edit`, personal = `can_edit_personal`.
- `restoringVersionId: number | null` matches `result_id: number`; `isRestoring` comparison is correct for both team and personal paths.
- `formatVersionOperation` retained on the page (still used by the version-detail modal) and duplicated verbatim in the component — no divergence.
- `versions.slice(0, 3)`, `versions.length <= 1` guard, `key={result_id}`, and the scheduled-task tag branch all match the original.
- `currentVersion = personalResult.version || personalVersions[0]?.version || 0` — the added `|| 0` is a defensive type fix; behaviorally equivalent (version numbers are ≥1, so the `0` case matches no entry, same as the prior `undefined`).

## 4. Suggestions

- Add a unit test for `SummaryVersionHistory` covering the note line for a note-less `generate` entry and a `restore` entry — the current `SummaryDetailPage` tests (`schedule`, `participantReports`) don't touch note rendering, which is why this regression would ship silently.

## 5. Coverage notes

- Whether `restore` versions always carry a server-populated `operation_note` is not verifiable from this repo. If they always do, the `versionRestoreFromResult` fallback loss is moot for restores; the `generate` baseline-description loss stands regardless.
- Second advisory pass could not assess runtime `useI18n` resolution or the full CSS from the diff alone; I verified the class names against `index.css` directly (all present).

---

**Overall: REQUEST_CHANGES.** The extraction is clean and most behavior is faithfully preserved, but a "behavior-preserving" refactor that drops the operation-note fallback chain is a real regression. Re-migrate the fallback (P1) and restore the icon size (P2), then this is good to merge.

Jerry-Xin (dismissed):
## APPROVE @ `5fe583384d5b`

Reviewed at the current head. This is a clean net-negative extraction of the version-history strip into `SummaryVersionHistory`, and — importantly — the two blockers other reviewers raised are **already fixed at this head**. I byte-verified against the exact head bytes (not the compare, not an older commit):

### The two reported 🔴 do not hold at `5fe583384d5b`
1. **Note-fallback text preserved.** `formatVersionOperationNote` was not deleted — it was **moved into the component** (`SummaryVersionHistory.tsx:51`) with its full fallback chain intact: trimmed `operation_note` → `versionInitialGenerateDesc` (generate) → `versionRestoreFromResult` (restore w/ parent) → operation label. It renders **unconditionally** at `<div className="summary-version-note">{formatVersionOperationNote(version)}</div>`. So versions without a stored note still show descriptive text, exactly as before. The two i18n keys (`versionInitialGenerateDesc`, `versionRestoreFromResult`) are used by the component — not dead.
2. **`IconHistory size="small"` preserved.** The strip title renders `<IconHistory size="small" />` in the extracted component. No visual size regression at this head.

Both findings describe the earlier head `6e3af043f807`; the author addressed them in the current commit. mochashanyao's CR is anchored to that stale head; the one CR at the current head still grades them 🔴, but the current bytes have both fixes.

### Extraction fidelity — clean
- Both call sites (`renderVersionHistory` team + `renderPersonalVersionHistory` personal) delegate to the component with the same props/callbacks; `canRestore` is derived from the same permission flags (`can_edit_team||can_edit` team; `can_edit_personal` personal). Same `versions.slice(0,3)`, same `length<=1`/loading guard (now `return null`), same CSS classes (`summary-version-strip`/`-list`/`-item`/`-note`/`-actions`).
- Minor robustness nit (non-blocking): personal `currentVersion` now defaults `|| 0`, harmless.

### Net: APPROVE
Faithful extraction with note-fallback and icon parity preserved at the current head. Non-blocking: add a PR description. Suggest peers holding CRs re-anchor to `5fe583384d5b` — the byte state there already satisfies both findings.

yujiawei (dismissed):
# Code Review — PR #865 (octo-web) — re-review at head 5fe5833

Re-reviewing after the fix commit `5fe58338` ("fix(summary): restore formatVersionOperationNote fallback and IconHistory size"). My prior REQUEST_CHANGES was against head `6e3af04`; this verdict is against the current head `5fe583384d5b6380d8d8425715792ea2673daffd`.

## 1. Spec compliance

**Spec: ✅** — the extraction now genuinely preserves the original behavior, matching the component's doc comment ("保留原有 strip 布局、CSS 类名和交互行为"). No missing/extra/deviating behavior remains. Scope is 2 files (the summary version-history extraction only).

## 2. Both prior findings resolved

**P1 (was blocking) — operation-note fallback chain: FIXED.** `formatVersionOperationNote` is migrated into the component (`SummaryVersionHistory.tsx:51-63`) and the note now always renders (`<div className="summary-version-note">{formatVersionOperationNote(version)}</div>`). The migrated logic is byte-for-byte equivalent to the original inline four-stage chain:
1. trimmed `operation_note` if present,
2. `versionInitialGenerateDesc` for `generate`,
3. `versionRestoreFromResult` (with `parent_result_id`) for `restore`,
4. else the operation label.

Both previously-orphaned i18n keys (`versionInitialGenerateDesc`, `versionRestoreFromResult`) are consumed again — the initial-baseline (v1) and note-less restore entries recover their descriptions.

**P2 — icon size: FIXED.** `<IconHistory size="small" />` restored (`SummaryVersionHistory.tsx:67`).

## 3. Parity re-confirmed

The fix commit touched only `SummaryVersionHistory.tsx`; `SummaryDetailPage.tsx` is unchanged since my prior review, so the previously verified parity still holds (canRestore gating, restoringVersionId/result_id types, retained page-side `formatVersionOperation`, `slice(0,3)`, `length<=1` guard, `key={result_id}`, scheduled-tag branch, all CSS classes present in `index.css`).

## Overall: APPROVED

Both findings from the prior round are cleanly resolved and behavior now matches the pre-refactor inline implementation. Suggestion (non-blocking, for a follow-up): add a unit test covering the note line for a note-less `generate` and a `restore` entry, since no existing test touches note rendering.

lml2468 (dismissed):
## 🐙 PR #865 Re-Review @ `5fe583384d5b` — APPROVE (both flagged items fixed)

Re-reviewed at the current head `5fe583384d5b` (my prior review was on `6e3af043f807`, auto-dismissed when the head moved). The two items flagged across reviewers — my P2 note-fallback + mochashanyao/yujiawei's two 🔴 — are **both fixed at this head**, byte-verified:

### ✅ Note-fallback restored (`SummaryVersionHistory.tsx:51, :103-104`)
`formatVersionOperationNote` now lives in the component with the **full original fallback chain** — `operation_note` → `versionInitialGenerateDesc` (generate) → `versionRestoreFromResult` (restore w/ parent) → operation label — and is rendered **unconditionally** in `.summary-version-note`. Versions without a stored note again show descriptive text. The two i18n keys (`versionInitialGenerateDesc`, `versionRestoreFromResult`) are live again (grep-confirmed in use), not dead. My earlier P2 is resolved.

### ✅ IconHistory size restored (`:68`)
`<IconHistory size="small" />` — the visual-parity break mochashanyao caught is fixed.

### Verification (local checkout @ `5fe583384d5b`)
| Check | Result |
|---|---|
| `dmworksummary` full suite | ✅ 386/386 |

Extraction remains faithful otherwise: both call sites (team + personal) delegate to the component with correct props/callbacks/permissions (`canRestore` from `can_edit_team||can_edit` / `can_edit_personal`), `slice(0,3)`, `length<=1` guard, and CSS classes all preserved.

### Non-blocking follow-up
yujiawei's suggestion — add unit tests for the no-`operation_note` generate/restore render paths — is a fair fast-follow (that exact path is what regressed and returned; a test would lock it). Not blocking. A PR description would still help.

---

### ✅ Review Verdict: APPROVE @ `5fe583384d5b`
Clean extraction with both previously-flagged behavioral gaps (note-fallback text, IconHistory size) now fixed and byte-verified; build green, 386 tests pass. My APPROVE is re-anchored to the current head.

> Gate note: the PR still shows `reviewDecision: CHANGES_REQUESTED` only because mochashanyao's CR is stale at the old head `6e3af043f807` — the code it flagged is fixed here, so it needs a re-anchor/dismiss to clear. Three reviewers (Jerry-Xin, yujiawei, me) now APPROVE at the current head.

quanming1:
Closing in favor of internal PR. Will reopen after team review on dmwork-org/octo-web#9.

quanming1:
## Status Update

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #865 (octo-web)

## Summary

This PR extracts the version-history strip UI from inline JSX in the class component `SummaryDetailPage` into a standalone functional component `SummaryVersionHistory`. The old `SummaryVersionHistory` file (which used a `Timeline` layout and `SummaryResult[]` type, but had zero consumers at base) is fully replaced. Both `renderVersionHistory()` and `renderPersonalVersionHistory()` now delegate to the new component via props and callbacks. `formatVersionOperationNote` is removed from the parent class (moved into the new component); `formatVersionOperation` is retained in the parent for use in `renderVersionDetailModal`. The refactoring is clean and faithful to the original behavior.

## Verification

Static analysis only at head `5fe583384d5b6380d8d8425715792ea2673daffd`; build and tests not executed in this environment.

- ✅ **Type contract** — `versions` state is `SummaryVersionItem[]` in both team (`:97`) and personal (`:100`) paths. The new component's prop type `SummaryVersionItem[]` matches. Handlers `handleViewVersion`, `handleRestoreVersion`, and `handleRestorePersonalVersion` all accept `SummaryVersionItem` — consistent with the new callback signatures.
- ✅ **Callback wiring parity** — `renderVersionHistory` passes `onViewVersion` wrapping `handleViewVersion(version, false)` and `onRestoreVersion` wrapping `handleRestoreVersion`. `renderPersonalVersionHistory` passes `handleViewVersion(version, true)` and `handleRestorePersonalVersion`. Matches the old inline `onClick` handlers exactly.
- ✅ **Permission checks preserved** — `canRestore` for team: `!!(detail.permissions?.can_edit_team || detail.permissions?.can_edit)`. For personal: `!!detail?.permissions?.can_edit_personal`. Both match the original inline conditional expressions.
- ✅ **`currentVersion` fallback** — Personal path adds `|| 0` to prevent `undefined` propagation. Since `version.version` is always a positive integer, `0` will never match any real version — same observable behavior as the old `undefined` comparison.
- ✅ **No stale consumers** — Confirmed the old `SummaryVersionHistory` component had zero imports at base (checked via `git grep` at `pr-base`). The replacement introduces no orphan references.
- ✅ **`formatVersionOperation` retained** — Still used at `:2183` in `renderVersionDetailModal` for the modal title. Correctly kept in the parent class.
- ✅ **`IconHistory` import retained** — Still used at `:2271`, `:2414`, `:2470` in the parent. No unused import introduced.
- ✅ **`formatVersionOperationNote` removed from parent** — Was only used in the inline version strips. The new component has its own semantically identical copy.

## Findings

### P2 — Duplicated formatting logic (`packages/dmworksummary/src/components/SummaryVersionHistory.tsx:43`)

`formatVersionOperation` and `formatVersionOperationNote` now live in two places: the parent class (`SummaryDetailPage:2123`, used by `renderVersionDetailModal:2183`) and this new component (lines 43–73). Both copies are semantically identical today — same i18n keys, same fallback chain, same `operation_type` handling. However, any future change to one copy (e.g. adding a new operation type, changing a fallback key) must be mirrored in the other. The second commit (`5fe58338`) already had to fix a `formatVersionOperationNote` fallback that was initially dropped, which illustrates exactly this risk.

Consider extracting both functions to `utils/summaryHelpers.ts` and importing them in both the parent and this component.

### Nit — Redundant defensive guards (`packages/dmworksummary/src/components/SummaryVersionHistory.tsx:38`)

The component's early-return guard (`!versions || versionsLoading || versions.length <= 1`) is redundant with the parent's identical checks at `:2134` and `:2157`. Harmless defensive coding — the component will never receive data that violates these conditions from its current callers. Not a bug, just noting for future maintainers who may assume the component is called without preconditions.

## Data-Flow Tracing

- `versions` ← `this.state.versions` (`SummaryVersionItem[]`) ← API response via `loadVersions()` (unchanged, not in diff).
- `personalVersions` ← `this.state.personalVersions` (`SummaryVersionItem[]`) ← API response via `loadPersonalVersions()` (unchanged, not in diff).
- `currentVersion` ← `detail.result.version` (team) / `personalResult.version || personalVersions[0]?.version || 0` (personal) — both derived from state populated by API responses.
- `restoringVersionId` / `restoringPersonalVersionId` ← set in `handleRestoreVersion` / `handleRestorePersonalVersion` entry, cleared in `finally`. Component reads them to show loading state on the correct item.
- `canRestore` ← derived from `detail.permissions` at render time. No caching, always reflects current API state.
- Callbacks (`onViewVersion`, `onRestoreVersion`) → parent class methods. No data transformation at the boundary — the component passes the full `SummaryVersionItem` object through.

## Blind-Spot Checklist

- **C1 — Dual-path parity**: `renderVersionHistory` and `renderPersonalVersionHistory` are symmetric. Both pass identical prop structure with different data sources and callbacks. ✅ Consistent.
- **C2 — Control-flow ordering**: `formatVersionOperation` exists in parent (modal title) and component (strip). Both copies are semantically identical. No double-fire or ordering risk. ✅ Clear.
- **C3 — Authorization boundary**: `canRestore` correctly propagates `detail.permissions` checks. No authorization bypass introduced. ✅ Clear.
- **C4 — Authorization lifecycle**: N/A — permissions read from current `detail` state on every render.
- **C5 — Build/config**: N/A — no build, config, or packaging changes.
- **C6 — Governance/policy**: N/A — no documentation or policy changes.

## Verdict: APPROVED

Clean refactoring that faithfully extracts inline JSX into a standalone component without altering observable behavior. Type contracts, callback wiring, permission checks, and i18n keys all match the original. The only concern is duplicated formatting logic between parent and component, which is non-blocking.

Jerry-Xin (approved):
## APPROVE @ `3c74efe484d4` (re-anchored)

Re-anchoring to the current head. The new commit is a single additive test file — no production change — so my prior APPROVE stands.

Byte-verified the true increment (ahead:1, behind:0 — a real commit, not a rebase): the only file changed is `packages/dmworksummary/src/components/__tests__/SummaryVersionHistory.test.tsx` (+220, added). The production files (`SummaryVersionHistory.tsx`, `SummaryDetailPage.tsx`) are byte-identical to the head I already approved (`5fe583384d5b`).

The added tests are real (not stubs) and cover the previously-flagged gaps: `null` when versions < 2 / loading, strip renders with title+hint when versions ≥ 2, current-version tag, and operation-type/note rendering across generate/manual_edit/regenerate. This addresses the earlier "add unit tests for no-note generate/restore entries" 🟡. Build + osv/secret/dependency scans green.

Net: clean extraction (note-fallback + IconHistory parity preserved, verified earlier) now with dedicated test coverage. APPROVE. The two new 🟡 (extract the duplicate formatter into a shared util; trim redundant defensive checks) are reasonable non-blocking follow-ups.

yujiawei (approved):
# Code Review — PR #865 (octo-web)

Reviewed at head `3c74efe484d4f12cdff5203f070f42c551d97023` (note: a test-only commit landed after the PR was opened; this review covers the current head, not the original `5fe58338`).

## 1. Spec Compliance

**Spec: ✅**

This is a pure component-extraction refactor: version-history rendering moves out of `SummaryDetailPage.tsx` into a standalone `SummaryVersionHistory.tsx`, wired via `renderVersionHistory()` and `renderPersonalVersionHistory()`.

- **Missing scope:** None. The extracted component reproduces the previously-inlined strip exactly — guard (`!versions || versionsLoading || versions.length <= 1` → `null`), `versions.slice(0, 3)`, current-version tag, `scheduled_generate` tag, operation label, note fallback chain, and view/restore button gating.
- **Out-of-scope additions:** None. No new flags, fields, or behaviors beyond the extraction.
- **Divergence:** None. Prop mapping is correct:
  - team `canRestore = can_edit_team || can_edit`, personal `canRestore = can_edit_personal` — matches the original inline predicates.
  - `restoringVersionId` / `restoringPersonalVersionId` and the `isPersonal` flag on view/restore callbacks are threaded correctly.
  - The added `|| 0` on personal `currentVersion` is a correct type-narrowing fix (version numbers are ≥1, so `0` never spuriously marks a row as current).

## 2. Code Quality

**Quality: Approved**

Verified against base `50b6165b`:

- **No dead references / build breaks.** `formatVersionOperationNote` was fully moved into the component (removed from the page). `formatVersionOperation` correctly stays on the page because `renderVersionDetailModal()` still uses it. `SummaryVersionHistory` has exactly one consumer repo-wide (`SummaryDetailPage.tsx`), so the props-contract rewrite breaks no other caller.
- **Faithful extraction.** The old standalone `SummaryVersionHistory` was dead code at base (never imported), so rewriting it introduces no regression.
- **i18n keys present.** All `summary.detail.*` keys used by the component exist in both `zh-CN.json` and `en-US.json`.
- **Tests.** The follow-up commit adds `SummaryVersionHistory.test.tsx` (220 lines) covering the empty/loading guards, current-version and scheduled tags, restore-button permission gating, and the full note-fallback chain (`operation_note` → `versionInitialGenerateDesc` → `versionRestoreFromResult` → operation label). Mocks follow the existing `__mocks__/dmworkBase` pattern used across the package. (Not run in CI locally, but statically consistent with the component's behavior.)

### Non-blocking suggestions (P2)

- **Dead legacy assets.** The old (now fully replaced) code path was the only consumer of `summary.common.current`, the `summary.versionHistory.*` keys, and the `summary-version-history` / `-empty` / `-header` / `-meta` CSS selectors. These are now unreferenced in both locale files and `index.css`. This is pre-existing dead weight (they only ever served the never-imported old component), so it does not block this PR — but a follow-up cleanup would be tidy.

## 3. Overall Verdict

**APPROVED**

Spec ✅ and Quality Approved. Behavior and prop mapping are preserved 1:1; the added tests lock in the extracted contract. The only follow-up is optional cleanup of now-orphaned i18n/CSS assets.

PR #868, opened 2026-07-17
refactor(search): migrate ChannelSearch and GlobalSearch service + UI

botshen:
## Summary
- 收口 ChannelSearch / GlobalSearch 的 SearchService、SearchTypes 与共享结果 mapper
- 按 Service → bridge → features → ui 分层迁移两个既有搜索入口
- 抽取共享 SearchWorkspace、分页 hook，并拆分 ChannelSearch 筛选与结果组件
- 将 GlobalSearch Story 改为真实业务组件，覆盖默认、筛选、loading、空结果、错误和长文本
- 清理组件颜色变量、硬编码颜色和 Semi 内部样式覆盖，统一使用语义 token

## Scope
Included:
- 聊天页内搜索与全局搜索的 Service + UI 迁移
- 兼容旧入口、远端开关、结果定位和文件操作行为
- 中英文文案、测试、Story 与迁移四件套文档

Excluded:
- 后端召回、排序、分词和权限模型变更
- 新增搜索入口或路由
- 与搜索无关的视觉系统和历史模块重构

## Validation
- `git diff --check`
- 21 个搜索相关测试文件，147 个用例通过
- `pnpm i18n:check`
- `pnpm --dir apps/web build`

## Storybook note
- 开发态搜索 Stories 已完成 light / dark 检查
- 本机完整静态构建在 Vite 8 / Rolldown rc.13 的 Apple Silicon 原生阶段触发 SIGILL；按约定作为环境问题跳过，未将其记录为通过

## Risk
- 影响聊天页右侧频道搜索、全局搜索弹窗及共享搜索 Service / mapper
- 保留单一用户入口，不改变登录、Space、路由和消息渲染契约


Jerry-Xin (approved):
## APPROVE @ `58ca47c98975`

Byte-verified the risk-bearing logic of this ChannelSearch/GlobalSearch Service→bridge→features→ui migration (the bulk of the +6202/-5383 is CSS, Stories, and old-implementation deletion; I focused on the behavior-bearing paths). It's a faithful, well-structured refactor. CI is green (Build + osv/secret/dependency scans pass).

### Verified clean
- **Result mappers (`SearchResultMapper.ts`)** — field preservation is complete across message/file/media/forward/rich-text hits, every field defensively defaulted (`|| ""` / `|| 0` / `undefined`). `mapCombinedHit` returns `undefined` for unknown `result_type` (safe skip). `normalizeItems` handles both bare-array and `{data,pagination}` envelope shapes. Rich-text/mention normalization filters malformed entities (finite offset/length, length>0).
- **URL normalization** — `https:/data:/blob:` passthrough, leading-slash strip, injected asset resolver; no unsafe URL construction.
- **`SearchService.ts`** — endpoints unchanged (`messages/_search`, `_search_media`, `_search_files`, `_search_global_messages`, `_search_global_files`, `_search_file_types`); all calls go through the existing authed `APIClient.shared`, so auth/session isn't reinvented. Global search is space-scoped when `spaceId` is set. Empty-query guards (`shouldRun*`) and a 64-rune keyword cap are in place. The CN-timezone day-boundary formatter (`secondsToDateOnlyCN` / `startOfCnDaySeconds` / `cnDatePresetRange`) is a genuine correctness fix — it prevents a non-CN browser's local "today" from straddling two CN calendar days and shipping an off-by-one `sent_at_from/to`; well-documented.
- **Backward compatibility** — the old entry points (`Components/ChannelSearch/index.tsx`, `Components/GlobalSearch/index.tsx`) are now thin re-export shims pointing at the new `features/*` panels, so existing import paths keep working. No consumer breakage.
- **Tests** — comprehensive suite retained/extended: `SearchService.test.ts`, `useSearchPagination.test.tsx`, plus ChannelSearch/GlobalSearch `__tests__` covering apiAdapter, filterState, pagination, locate, clickRouting, isActive, forwardInnerMessage, snippetContent, senderCandidates, thread.

### Non-blocking
- Large surface — worth a smoke pass on the live search UIs (channel + global, all tabs incl. media/file, filters, empty/error/long-text states) since so much CSS moved, to catch any purely-visual drift the Story coverage might miss.
- Consider a follow-up integration test asserting an end-to-end query→mapper→paginated-render for at least the combined-tab path, complementing the unit tests.

### Net: APPROVE
Clean layered migration with behavior, auth/scoping, backward-compat, and test coverage all preserved; CI green. No blocking issues found in the reviewed logic.

lml2468 (approved):
## 🐙 PR #868 Review — `refactor(search): migrate ChannelSearch and GlobalSearch service + UI`

**Head:** `58ca47c98975` · **Base:** `main` (merge-base `e0b2ff30`) · +6202 −5383 / 70 files
**Intent:** Layered migration of the two existing search entries (chat-page channel search + global search) into `Service → bridge → features → ui`, extracting a shared `SearchService`/`SearchTypes`/mapper and a `SearchWorkspace` shell. Explicitly **no** backend recall/ranking/permission changes, no new entry/route.

### Verification (local checkout @ `58ca47c98975`)
| Check | Result |
|---|---|
| search test suites (Service/bridge/features/Components) | ✅ **452 passed** |
| dangling imports to deleted modules (dataSource/vm/apiAdapter/…) | ✅ none |

> Test-env note: a first run showed 26 failures — **all** the Node-26/jsdom `localStorage` gap (dmworkbase has no Storage setup); with a Storage shim, 452/452 pass. Two files still fail to *load* (`filterSearchSelect.test.tsx`: `Cannot find module 'react-dom/client'` — React 17 vs `@testing-library/react@14`; `UploadCredentials.test.ts`: canvas `fillStyle` stub) — I verified **both fail identically at merge-base** `e0b2ff30` and **neither is touched by this PR** → pre-existing/environmental, not this change.

### Gate 1 — Spec compliance: ✅ (one entry-consolidation to note)
Migration matches the described layering; old `Components/{ChannelSearch,GlobalSearch}/index.tsx` are kept as **back-compat re-export shims** (no import breakage). One thing the description undersells:

- **Second channel-search entry removed (intentional, rule-compliant).** `module.tsx` drops the `channel.base.settingMessageHistory` info-panel entry (+ the `onOpenChannelSearch` plumbing through `ChannelSetting`), keeping the **chat-toolbar `ChatSearchEntryButton`** as the single entry (`module.tsx:663-664`, retained). The old code's own comment confirms both entries used "同一 feature 门禁与同一打开效果" — i.e. duplicates — and `AGENTS.md` requires "同一业务能力只保留一个用户可见入口." So this is a **compliant consolidation**, not a functional loss. But the description says "保留旧入口 / 保留单一用户入口" which reads as *keeping* entries; it should explicitly state the settings entry was removed. Please confirm that's intended (I believe it is, per the rule).

### Gate 2 — Code quality: ✅
- **Space isolation preserved:** `GlobalSearchVM` reads `WKApp.shared.currentSpaceId` and passes it as `spaceId` (`:148,157`); `SearchService` appends `?space_id=` (backend-enforced). The omitted-`spaceId` fallback mirrors the pre-existing contract — not a new isolation gap.
- Back-compat shims keep old import paths valid; no dangling refs to the removed `dataSource`/`vm`/`apiAdapter` modules.
- `SearchService` uses `await`ed `APIClient` calls with defensive array-coercion on the response (`friends/groups/messages` → `Array.isArray(...) ? ... : []`), reducing undefined-shape risk. Token/auth flows through the shared `APIClient` (unchanged).
- The PR itself reports 147 search cases across 21 files; my run of the search suites shows 452 green (broader net), consistent.

### Coverage / blind spots (honest scope statement)
This is a **12,526-line diff**; I did not line-audit every moved file. My confidence rests on: build green, the full search test suite passing, behavior-preservation via shims + retained toolbar entry, byte-verified space-scoping, and no dangling imports. What I did **not** exhaustively verify: pixel-level UI parity of the new `features/`/`ui/` components (the PR notes Storybook light/dark was done manually; the full static Storybook build hit a Vite8/Rolldown SIGILL on Apple Silicon — an env issue, reasonably skipped), and every filter/pagination edge in `ChannelSearchResults` (739 lines). A reviewer familiar with the search UX should smoke-test the two live entries.

Old-path shims resolve ✅ · retained single entry wired ✅ · space isolation preserved ✅ · no dangling imports ✅ · response shape defended ✅ · the 28 test failures all environmental/pre-existing (merge-base-confirmed) ✅.

---

### ✅ Review Verdict: APPROVE @ `58ca47c98975`
A large but disciplined migration: layered as described, back-compat shims prevent breakage, space isolation and the single search entry are preserved, build + 452 search tests green (all failures environmental, byte-confirmed pre-existing at merge-base). One doc nit: the description should state the redundant `settingMessageHistory` entry was intentionally removed (consolidation per `AGENTS.md`), not "保留旧入口." Given the size, recommend a human smoke-test of both live search entries before merge — but nothing blocks on the code.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #868 (octo-web)
Head: `58ca47c9` (base `92210132`→`58ca47c9`, 70 files, +6202/−5383)
Scope: refactor(search) — migrate ChannelSearch & GlobalSearch into service/bridge/feature/UI layers.

## Summary

Large, well-structured refactor that lifts the channel/global search surface into a clean four-layer architecture: **Service** (`SearchService`/`SearchTypes`/`SearchResultMapper`) → **Bridge** (data sources, `useSearchPagination` hook, filterState/pagination/locate) → **Feature** (`ChannelSearchPanel`/`GlobalSearchPanel` and result components) → **UI** (`SearchWorkspace`). The legacy monolithic `Components/ChannelSearch/index.tsx` (1821 lines) and `Components/GlobalSearch/index.tsx` are reduced to thin re-exports, with the VM retained only for legacy contacts/group/file paths behind the `messagesSearchOn` feature flag. Cursor-based auto-pagination, CN-tz date math, self-uid drop, and DM-exclusion in the global channel picker are all handled explicitly. Two non-blocking P2 findings and one nit below; no blockers.

## Verification

Static analysis only at head `58ca47c9`; build/tests not executed in this environment.

- ✅ **Race-condition hygiene in `useSearchPagination`** — `requestIdRef` + `mountedRef` correctly discard stale responses; `loadingMoreCursorRef` dedupes concurrent calls to the same cursor (`bridge/search/useSearchPagination.ts:69-119`).
- ✅ **AbortSignal propagation** — `query.signal` reaches `APIClient.shared.post` for `searchGlobalMessages` (`Service/SearchService.ts:375`); `getGlobalFileTypes` also accepts a signal (`:408`). `searchChannelMessages` does not take a signal — not a regression vs. the legacy adapter.
- ✅ **Self-uid drop in global member filter** — `cleanGlobalFilters` removes self from `memberUids` before serializing (`Service/SearchService.ts:185-191`); `createGlobalSearchDataSource` seeds the sender cache with self so the drop always has a value to compare (`bridge/globalSearch/createGlobalSearchDataSource.ts:192-200`).
- ✅ **DM exclusion from the 「所在群聊或子区」 picker** — `loadReadableChannelOptions` keeps only `ChannelTypeGroup` (2) and `ChannelTypeCommunityTopic` (5), skipping DMs (`bridge/globalSearch/createGlobalSearchDataSource.ts:65-72`); comments reference [automated review](mention://issue/f3efdfd4-c337-4b95-b682-6e84243b0cab) bug 2.
- ✅ **Entry-point swap** — `module.tsx` drops `channel.base.settingMessageHistory`; `registerChannelHeaderRightItems` registers `ChatSearchEntryButton` which emits `wk:open-channel-search` via mittBus (`features/channelSearch/ChatSearchEntryButton.tsx:22-28`, `module.tsx`). `ChannelSetting` loses the `onOpenChannelSearch` prop cleanly (no dangling consumers — `Pages/Chat/index.tsx` updated).
- ✅ **Cursor-based pagination guards** — `shouldStopPaginationForCursor` halts when the server echoes the requested cursor; `shouldPauseAutoPaginationForEmptyPage` pauses on empty-but-advancing pages so auto-scroll doesn't race to the end (`bridge/channelSearch/pagination.ts:18-49`).
- ✅ **Re-export parity** — `Components/ChannelSearch/apiAdapter.ts` and `Components/GlobalSearch/apiAdapter.ts` now re-export the service-layer symbols consumed by legacy callers (`foldResponse`, `globalSearchEndpoint`, `shouldRunGlobalSearch`, `cnDatePresetRange`, etc.). Symbol set matches `SearchService.ts`.

## Findings

No P0/P1 issues. Two P2 items below plus one nit.

### P2 — ChannelSearch date filter still uses browser-local day math (`packages/dmworkbase/src/features/channelSearch/ChannelSearchFilters.tsx:22-42`)

`startOfDay`/`endOfDay` call `setHours(0,0,0,0)` / `setHours(23,59,59,999)` in the **browser's local timezone**, then `toSeconds` emits epoch seconds. The server's `_search*` endpoints interpret `sent_at_from`/`sent_at_to` as Asia/Shanghai calendar days (backend §7.3, which is why `SearchService.ts:81-130` ships dedicated `cnCalendarParts`/`startOfCnDaySeconds`/`endOfCnDaySeconds`/`cnDatePresetRange` for GlobalSearch). For any user outside UTC+8, the two systems disagree:

- A PST user picking `2026-07-17` in the date picker produces `startAt=1784558400` (= 2026-07-17 07:00 UTC = 2026-07-17 15:00 Shanghai). `secondsToDateOnly` (also browser-local) serializes that as `"2026-07-17"`. Server treats it as 2026-07-17 Shanghai midnight, so the effective window starts 15h after the user's intended local midnight, dropping the morning half of their local day.
- The effect scales with tz offset; for UTC+8 users it collapses to zero, so this is invisible for the primary audience but real for any roaming / non-CN user.

`GlobalSearch` already fixed this with `startOfCnDaySeconds`/`endOfCnDaySeconds` (`Service/SearchService.ts:107-128`). Recommend reusing the same CN-anchored helpers here — or, if the per-channel backend truly runs in browser tz, document the contract in a comment so the next refactor doesn't "fix" it the other way.

### P2 — Selected-sender chip resolves to bare UID on first popover open (`packages/dmworkbase/src/features/channelSearch/ChannelSearchFilters.tsx:83-92`, `packages/dmworkbase/src/bridge/channelSearch/createChannelSearchDataSource.ts:50-52`)

`ChannelSearchDataSource.getSender(uid)` returns `senderCache.get(uid) || { uid, name: uid }`. The cache is populated **only** by `searchSenders` — not by the `getSenders()` list that seeds the dropdown. Sequence:

1. User opens filter popover → `dataSource.getSenders()` renders the list with real names.
2. User clicks a sender → chip calls `getSender(uid)` → cache miss → `{uid, name: uid}` → chip shows the raw UID.
3. Only after the user triggers a text search does `searchSenders` populate the cache and the chip resolves to the proper name.

GlobalSearch sidesteps this by seeding the cache with `selfSender()` on construction, but `createChannelSearchApiDataSource` has no analogous seed. Two clean fixes: (a) in the bridge, prime the cache from `getSenders()` once; or (b) in `getSender`, fall back to a linear lookup in `getSenders()` before settling on `{uid, name: uid}`.

### Nit — `aria-label` on the nav copies the search-input placeholder (`packages/dmworkbase/src/ui/SearchWorkspace/index.tsx:49`)

`<nav aria-label={search.placeholder}>` makes the tab group announce itself as e.g. "Search in this channel" (the input's placeholder) rather than describing the tab strip. Use a dedicated label such as `t("search.tabs")` or accept an optional `navLabel` prop.

- **Tab mount strategy** — `GlobalSearchPanel.tabPanels` mounts all four tabs and toggles via `display:none`; intentional to avoid ``/`VisibilityTrigger` rebuilds and the resulting cache-miss storm on avatar URLs (`features/globalSearch/GlobalSearchPanel.tsx:174-194`).
- **Feature-flag freshness** — `contentSearchEnabled` is a per-render getter on the class plus a `remoteConfig.addConfigChangeListener` → `forceUpdate` so mid-session flips take effect without reload (`features/globalSearch/GlobalSearchPanel.tsx:103-122`).
- **File-type cache retry semantics** — on `getGlobalFileTypes` failure, `moduleFileTypeCategoriesCache.inFlight` is cleared so the next call retries; the rejected promise still resolves for any in-flight awaiter via `.catch(() => [])` (`bridge/globalSearch/createGlobalSearchDataSource.ts:242-262`).
- **Compose guard for IME** — `isComposingRef` short-circuits `onChange` and `onCompositionEnd` re-fires the update, so Chinese/Japanese composition doesn't trigger debounced searches mid-input (`features/channelSearch/ChannelSearchPanel.tsx:255-273`).
- **Keyword rune cap** — `truncateChannelSearchKeyword` uses `Array.from().slice(0, 64).join("")`, which is rune-correct for CJK/emoji; toast fires once per overflow and re-arms when the user types back under the limit (`features/channelSearch/ChannelSearchPanel.tsx:112-128`).
- **Locate guard** — `canLocateChannelSearchItem` requires `messageSeq > 0` before offering "定位到聊天"; `resolveChannelSearchLocateTarget` rebuilds the target Channel and the panel prefers `conversationContext.locateMessage` for the current channel, falling back to `WKApp.endpoints.showConversation` for cross-channel hits (`bridge/channelSearch/locate.ts`, `features/channelSearch/ChannelSearchPanel.tsx:155-175`).
- **Forward-inner-message rendering** — `FORWARD_INNER_MESSAGE_DISPLAY_LIMIT` caps the visible preview; `getForwardInnerMessageHiddenCount` produces the "还有 N 条" tail; when `innerMessages` is empty the legacy `snippets` array still renders, so older hits don't show a blank card (`features/channelSearch/ChannelSearchResults.tsx`).
- **No dangling references to removed `onOpenChannelSearch`** — grep across the tree after the `ChannelSetting` prop removal shows only the new mittBus-driven path remains.

## Verdict: APPROVED

Architecture is clean, data-flow contracts are traced correctly from API → mapper → hook → UI, and the two P2s are real but not user-blocking for the primary UTC+8 audience. The timezone P2 is worth closing before non-CN usage grows; the sender-chip P2 is cosmetic but easy to fix. Nit is optional.

yujiawei (approved):
# Code Review — PR #868 (octo-web)

Reviewed as a behavior-preserving refactor of the ChannelSearch / GlobalSearch stack into a `Service → bridge → features → ui` layering. Reviewed against merge-base `e0b2ff30` at head `58ca47c9`. Verified locally: `pnpm --dir apps/web build` (green), `pnpm i18n:check` (green), `git diff --check` (clean), and the search test suites.

The stated contract is "no change to endpoint, params, click-routing, message-locate, file-download, or the remote feature-gate; no new routes/entries; a single user entry point." Verified against the diff:

- **Endpoints & wire params unchanged.** `channelSearchEndpoint` / `globalSearchEndpoint` / `_search_file_types` / legacy `/search/global` are byte-identical (relocated only). `cleanGlobalFilters`, `toGlobalRequestBody`, content-type constants (`[1,14,8,11]` / `[2,5]`), `member_uids` self-drop+dedup, and the CN-timezone `sent_at` boundary math all carry over unchanged.
- **Space isolation preserved.** `GlobalSearchVM.requestSearch` still scopes via `WKApp.shared.currentSpaceId` → `?space_id=...` and keeps the `requestId` race-guard (stale responses dropped), matching the pre-PR `vm.ts` exactly.
- **Message-locate contract preserved.** `resolveChannelSearchLocateTarget` / `canLocateChannelSearchItem` are unchanged (moved to `bridge/channelSearch/locate.ts`).
- **Single entry point.** The PR removes the duplicate `channel.base.settingMessageHistory` info-panel entry, leaving the channel-header search button (`channelheader.search`, feature-gated by `isChannelSearchEnabled`) as the sole entry. Both entries existed at merge-base and opened the same panel via the same `wk:open-channel-search` mittBus event, so this is a de-duplication consistent with the "single entry" goal, not a functional loss.
- **No out-of-scope work.** No backend/recall/ranking changes, no new routes, no unrelated visual-system refactor. i18n additions (`globalSearch.clear`, `globalSearch.placeholder`) are paired in both `zh-CN` and `en-US`. No package.json / turbo.json test-script narrowing.

## 2. Code quality — ✅

- **Clean dependency inversion at the security boundary.** The result mapper no longer imports `WKApp` directly; host asset-URL resolution is injected via `SearchAssetResolver` (`createSearchAssetResolver`), with a `passthrough` default. This is a genuine improvement — the pure mapper is now unit-testable without the app shell.
- **Race-safety is solid.** `useSearchPagination` uses a monotonic `requestIdRef` + `mountedRef` to drop stale/unmounted responses, and cancels the RAF on unmount; the legacy aggregated path keeps its own `requestId` guard. The spec's "stale request must not overwrite a new query" requirement holds.
- **Prior RC fixes retained.** The documented `commonDataSource.searchFriends` contacts fix (previously a silent `as any` no-op) and the DM-exclusion in the channel picker are carried forward intact.
- Build, i18n, and whitespace checks all pass; the PR's four new/updated search test files pass (11/11), and 452 search-suite tests pass overall.

### Non-blocking notes (nits / suggestions)

- **N1 — asset-resolver falsy fallthrough.** `createSearchAssetResolver` uses `getImageURL?.(path) || \`${apiURL}${path}\``, whereas the old code returned the host result whenever `getImageURL` *existed*. If a host implementation ever legitimately returns `""`, the new code falls through to `apiURL+path` instead of `""`. In practice these resolvers never return empty for a non-empty path, so this is defensive-only, but a `?? `-style guard would preserve the exact old semantics.
- **N2 — Storybook test-file stub plugin is broad.** `apps/web/.storybook/main.ts` stubs any path containing `__tests__` (not just `*.test.*` / `*.spec.*`) to `export default {}`. No story currently imports from a `__tests__` dir (verified), so this is latent only; consider narrowing the matcher to `.test./.spec.` to avoid a future silent-empty-import surprise.

No P0/P1. Behavior-preserving refactor with endpoints, params, Space scoping, race-guards, and locate/download contracts all verified unchanged; build + i18n + targeted tests green. N1/N2 are non-blocking suggestions.

Note for the merge owner: this PR was auto-flagged for extra scrutiny on a keyword match ("migration"/"credential" appearing in filenames). No auth, token, credential, or endpoint semantics are actually touched by this diff — the sensitive-classification is a false positive here, but the Space-isolation and stale-response paths were reviewed line-by-line and confirmed intact.

PR #872, opened 2026-07-17
feat(channel-webhook): refine UI and separate data boundaries

Candice00711:
Move incoming webhook requests into a dedicated service and bridge list loading through a scoped hook with stale-request protection.\n\nSplit the list card into a props-only UI component, align the list, empty state, and edit form with the approved design, and preserve the existing permission and request semantics.\n\nValidation: focused Vitest suites, i18n check, web build, Storybook build, and git diff check.

## Changes

-
-

- Affected module(s):
- Shared layer touched: 

## Testing

- [ ] Manually verified

## Checklist

- [ ] Updated documentation

### DIFF
--- a/packages/dmworkbase/src/Components/ChannelWebhook/ChannelWebhook.stories.tsx
+++ b/packages/dmworkbase/src/Components/ChannelWebhook/ChannelWebhook.stories.tsx
@@ -7,6 +7,7 @@ import {
     IncomingWebhookCreateResp,
 } from "../../Service/IncomingWebhook";
 import ChannelWebhookPanel from "./index";
+import ChannelWebhookCard from "./ChannelWebhookCard";
 import WebhookEditModal from "./WebhookEditModal";
 import WebhookUrlModal from "./WebhookUrlModal";
 import MessageRow from "../../ui/message/MessageRow";
@@ -151,6 +152,39 @@ export const EmptyState: Story = {
             </div>
         );
     },
+};
+
+/** 纯 UI 卡片：不依赖 WKApp、SDK 或接口，可独立检查截图中的列表状态。 */
+export const CardUi: Story = {
+    name: "Webhook 卡片（纯 UI）",
+    render: () => (
+        <div style={{ width: 360, padding: 16 }}>
+            <ul className="wk-webhook__list">
+                <ChannelWebhookCard
+                    item={mockList[0]}
+                    manageable
+                    meta={<div className="wk-webhook-card__meta">魏娇莹 创建于 2026/7/17 16:13:53</div>}
+                    toggling={false}
+                    testingBlocked={false}
+                    cooling={false}
+                    labels={{
+                        disabled: "已禁用",
+                        toggle: "启用 / 禁用",
+                        edit: "编辑",
+                        regenerate: "重置推送地址",
+                        test: "发送测试消息",
+                        testDisabledHint: "启用后可测试",
+                        delete: "删除",
+                    }}
+                    onToggle={() => undefined}
+                    onEdit={() => undefined}
+                    onRegenerate={() => undefined}
+                    onTest={() => undefined}
+                    onDelete={() => undefined}
+                />
+            </ul>
+        </div>
+    ),
 };
 
 export const CreateModalAdmin: Story = {
--- a/packages/dmworkbase/src/Components/ChannelWebhook/ChannelWebhookCard.tsx
+++ b/packages/dmworkbase/src/Components/ChannelWebhook/ChannelWebhookCard.tsx
@@ -1,0 +1,109 @@
+import React from "react";
+import { Switch } from "@douyinfe/semi-ui";
+import { Edit3, RefreshCw, Send, Trash2 } from "lucide-react";
+import {
+    IncomingWebhook,
+    IncomingWebhookStatus,
+    INCOMING_WEBHOOK_DEFAULT_AVATAR,
+    canTestWebhook,
+} from "../../Service/IncomingWebhook";
+
+export interface ChannelWebhookCardLabels {
+    disabled: string;
+    toggle: string;
+    edit: string;
+    regenerate: string;
+    test: string;
+    testDisabledHint: string;
+    delete: string;
+}
+
+export interface ChannelWebhookCardProps {
+    item: IncomingWebhook;
+    manageable: boolean;
+    meta: React.ReactNode;
+    toggling: boolean;
+    testingBlocked: boolean;
+    cooling: boolean;
+    labels: ChannelWebhookCardLabels;
+    onToggle: (enabled: boolean) => void;
+    onEdit: () => void;
+    onRegenerate: () => void;
+    onTest: () => void;
+    onDelete: () => void;
+}
+
+/** Props-only Webhook list item. Runtime data and mutations stay in the panel container. */
+export default function ChannelWebhookCard({
+    item,
+    manageable,
+    meta,
+    toggling,
+    testingBlocked,
+    cooling,
+    labels,
+    onToggle,
+    onEdit,
+    onRegenerate,
+    onTest,
+    onDelete,
+}: ChannelWebhookCardProps) {
+    const enabled = item.status === IncomingWebhookStatus.enabled;
+    const testAllowed = canTestWebhook(item);
+    const testDisabled = !testAllowed || toggling || testingBlocked || cooling;
+
+    return (
+        <li className="wk-webhook-card">
+            <div className="wk-webhook-card__head">
+                <img
+                    src={item.avatar || INCOMING_WEBHOOK_DEFAULT_AVATAR}
+                    alt=""
+                    className="wk-webhook-card__avatar"
+                />
+                <div className="wk-webhook-card__content">
+                    <div className="wk-webhook-card__titlebox">
+                        <span className="wk-webhook-card__name" title={item.name}>{item.name}</span>
+                        {!enabled && (
+                            <span className="wk-webhook-card__chip wk-webhook-card__chip--off">
+                                {labels.disabled}
+                            </span>
+                        )}
+                    </div>
+                    <div className="wk-webhook-card__meta-wrap">{meta}</div>
+                </div>
+                {manageable && (
+                    <Switch
+                        size="small"
+                        checked={enabled}
+                        loading={toggling}
+                        onChange={onToggle}
+                        aria-label={labels.toggle}
+                    />
+                )}
+            </div>
+            {manageable && (
+                <div className="wk-webhook-card__actions">
+                    <button type="button" className="wk-webhook-card__icon-btn" onClick={onEdit} title={labels.edit} aria-label={labels.edit}>
+                        <Edit3 aria-hidden="true" />
+                    </button>
+                    <button type="button" className="wk-webhook-card__icon-btn" onClick={onRegenerate} title={labels.regenerate} aria-label={labels.regenerate}>
+                        <RefreshCw aria-hidden="true" />
+                    </button>
+                    <button
+                        type="button"
+                        className="wk-webhook-card__icon-btn"
+                        disabled={testDisabled}
+                        onClick={onTest}
+                        title={testAllowed ? labels.test : labels.testDisabledHint}
+                        aria-label={labels.test}
+                    >
+                        <Send aria-hidden="true" />
+                    </button>
+                    <button type="button" className="wk-webhook-card__icon-btn wk-webhook-card__icon-btn--danger" onClick={onDelete} title={labels.delete} aria-label={labels.delete}>
+                        <Trash2 aria-hidden="true" />
+                    </button>
+                </div>
+            )}
+        </li>
+    );
+}
--- a/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
+++ b/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { Channel, ChannelTypePerson, WKSDK } from "wukongimjssdk";
-import { Select, Switch, Toast } from "@douyinfe/semi-ui";
+import { Switch, Toast } from "@douyinfe/semi-ui";
 import { IconAlertTriangle } from "@douyinfe/semi-icons";
 import WKModal from "../WKModal";
 import WKButton from "../WKButton";
@@ -13,6 +13,7 @@ import {
     buildWebhookUpsertReq,
     IncomingWebhook,
     IncomingWebhookCreateResp,
+    IncomingWebhookService,
     isFlagOn,
     MENTION_UID_MAX_LENGTH,
     MENTION_UIDS_MAX,
@@ -145,6 +146,7 @@ export default function WebhookEditModal({
         () => readGroupMemberOptions(channel)
     );
     const [saving, setSaving] = useState(false);
+    const [memberSearch, setMemberSearch] = useState("");
     // 本组件由父级条件挂载（{editTarget && <WebhookEditModal/>}），且处于
     // WKViewQueue 路由栈的滑入动画里。若一挂载就 visible=true，Semi Modal 的
     // 首次显示会与路由动画/portal 时序竞争，表现为「要点两次才弹出」。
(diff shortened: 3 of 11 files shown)


lml2468 (dismissed):
## 🐙 PR #872 Review — `feat(channel-webhook): refine UI and separate data boundaries`

**Head:** `cf1b9b205729` · **Base:** `main` (merge-base `50b6165b`) · +620 −308 / 11 files
**Intent:** Move incoming-webhook HTTP into a dedicated `IncomingWebhookService`, bridge list loading through a scoped hook with stale-request protection, split the list card into a props-only UI component, and align list/empty/edit UI with the approved design — preserving existing permission + request semantics.

### Verification (local checkout @ `cf1b9b205729`)
| Check | Result |
|---|---|
| webhook tests (Service + Modal) | ✅ 28 passed |
| dmworkdatasource tests | ✅ 18 passed |

Does exactly what it describes — service extraction + bridge hook + UI split, no scope creep. The `datasource.ts` change is a clean **delegation refactor**: the inline `incomingWebhookBase` + `apiClient` calls are replaced by `IncomingWebhookService.*` (same 6 endpoints), removing duplication; no dangling refs to the removed helper. (Description body is the unfilled template — a nit, but the header paragraph covers it.)

### Gate 2 — Code quality: ✅
- **Service HTTP boundary is safe:** all paths through the shared `APIClient`, every dynamic segment `encodeURIComponent`'d (`groupNo`/`threadShortId`/`webhookId`) → no path injection. `list()` defensively handles bare-array + `{list|items|webhooks}` + null — a **superset** of the old `resp?.list || []`, so no shape regression.
- **Bridge hook stale-request protection is correct:** monotonic `requestSequenceRef` + `mountedRef`, re-checked after every await in `load()` (success/catch/finally), incremented on unmount and scope change → no cross-scope result bleed or setState-after-unmount. Scope key `${groupNo}:${threadShortId||'group'}` correctly refetches on channel/thread switch.
- **Permission boundary preserved:** per-item `canManageIncomingWebhook(item, {isManager, myUid})` still gates management (`index.tsx:273`); blame confirms that function is **pre-existing, unchanged** by this PR (last touched 2026-06-12). UI-gating only, server enforces — unchanged.
- **No secret mishandling:** the one-time create token/URL flows through `onSaved` for display (expected webhook UX); no logging, no `dangerouslySetInnerHTML`; field caps mirror the OpenAPI schema.

Path injection (encodeURIComponent all segments) ✅ · list-shape regression (new handling is a superset) ✅ · stale-request race (seq+mounted guards) ✅ · permission gate intact + pre-existing ✅ · no dangling refs to removed `incomingWebhookBase` ✅ · token not logged ✅ · build + 46 tests green ✅.

> Env note: dmworkbase has no Storage setup for vitest (Node-26/jsdom); I added a local shim to run the suites — not part of the PR.

---

### ✅ Review Verdict: APPROVE @ `cf1b9b205729`
Clean service/UI/bridge separation with the security-relevant surfaces intact: encodeURIComponent'd paths through the shared APIClient, correct stale-request guarding, preserved (pre-existing) permission gating, and a superset list-shape handler. Build + 46 webhook/datasource tests green. Non-blocking: fill in the PR description template. Nice, well-tested refactor.

Jerry-Xin (dismissed):
## APPROVE @ `cf1b9b205729`

Byte-verified the risk-bearing paths of this channel-webhook UI-refine + data-boundary refactor (11 files, +620/-308). Clean, well-engineered, and the permission/request semantics are preserved as the description claims.

### Verified clean
- **Stale-request protection (`useChannelWebhookList.ts`)** — correct: a monotonic `requestSequenceRef` is captured per `load()`, and both the success and error paths bail if `requestSequence !== requestSequenceRef.current`; a `mountedRef` prevents setState-after-unmount and unmount bumps the sequence to invalidate in-flight calls. On scope change, the effect bumps the sequence + clears items + reloads, keyed on `scopeKey = groupNo:threadShortId`. So a slow response from a previous channel/thread can't land in the new scope's list — this is exactly the "separate data boundaries" the PR targets, done right.
- **`IncomingWebhook` service** — `basePath(groupNo, threadShortId)` scopes correctly (group face `groups/{group}/incoming-webhooks`; thread face `groups/{group}/threads/{short}/incoming-webhooks`, `encodeURIComponent` on the short id). All CRUD (list/create/update/delete/regenerate/test) go through the authed `APIClient.shared`, scope params threaded end-to-end. The plaintext token is returned only on create/regenerate and is never echoed in `list` — good secret hygiene.
- **`datasource.ts`** — the webhook methods are thin delegations to the service with `channel.channelID` (parent group_no) + `threadShortId`, correctly documented; the data-boundary change is a clean extraction. (The same-origin upload-token guard elsewhere in the file is unchanged.)
- **Permission gating preserved** — `canManageIncomingWebhook(item, { isManager, myUid })` is the same guard, same inputs, as base; the refactor moved the call into the props-only card without weakening it.
- **`WebhookEditModal`** — retains input validation (`validateMentionUids`, `maxLength` on name/avatar), one-time token handling intact, no `innerHTML`/XSS surface.
- **Tests** — new coverage added (`IncomingWebhookService.test.ts`, `datasource.test.ts`, `WebhookEditModal.test.tsx`). CI green (only the sentinel code-review check pending).

### Non-blocking
- `list` tolerates three response shapes (`list`/`items`/`webhooks`) — fine defensively, but worth converging the backend contract to one key eventually.
- A quick UI smoke of the group vs thread scopes (switch channel/thread rapidly) would confirm the stale-guard behavior end-to-end, though the logic reads correct.

### Net: APPROVE
Faithful refactor — data-scoping, auth, permission gating, and token hygiene all preserved, with a correct stale-request guard and added tests. No blocking issues.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #872 (octo-web)

## Summary

This PR refactors the ChannelWebhook (incoming webhook) management UI into cleaner layers: a new `IncomingWebhookService` centralises all HTTP boundaries, a `useChannelWebhookList` hook extracts list lifecycle from the container, and a `ChannelWebhookCard` component pulls card rendering out of the monolithic `index.tsx`. The datasource in `dmworkdatasource` is migrated to delegate to the shared service, the default avatar SVG is redesigned, and action icons move from Semi built-ins to lucide-react. Well-structured refactor with good test coverage for the new service boundary.

## Verification

- ✅ **Data-flow integrity** — `IncomingWebhookService.list()` correctly handles all response shapes (array, `{list}`, `{items}`, `{webhooks}`, null/undefined); traced from `APIClient.shared.get()` → `wrapResult()` (returns `value.data` raw) → service normalization → hook `setItems()` → card `item` prop.
- ✅ **Race-condition guards** — `useChannelWebhookList` uses `requestSequenceRef` monotonically incremented on both scope changes and `load()` calls; stale responses are discarded via `requestSequence !== requestSequenceRef.current` check.
- ✅ **Cleanup completeness** — cooldown timer cleared on unmount and on scope change; mounted flag prevents state updates after unmount; scope change resets all UI state (editTarget, urlResult, testingId, togglingId, coolingTestId).
- ✅ **Export chain** — `IncomingWebhookService` is correctly exported from `@octo/base` via `export * from './Service/IncomingWebhook'` in `src/index.tsx:21`; datasource import resolves correctly.
- ✅ **Permission model preserved** — `canManageIncomingWebhook()` and `canTestWebhook()` are unchanged pure functions; card receives pre-computed `manageable` and internally derives `testDisabled` from the same guards.
- ✅ **Test alignment** — `IncomingWebhookService.test.ts` covers path construction, envelope normalization, and CRUD routing; `WebhookEditModal.test.tsx` correctly updated mock call index from `[2]` to `[1]` matching the new `APIClient.shared.put(path, req)` contract; `datasource.test.ts` mocks the new service.

Static analysis only at head `cf1b9b2057294d44f56747abe0c537d546ad58c2`; build and tests not executed in this environment.

## Findings

### P2 — `encodeURIComponent` on `channelID` is a behavioural change (`packages/dmworkbase/src/Service/IncomingWebhook.ts:139`)

The old `ChannelDataSource.incomingWebhookBase` passed `channelID` and `webhookId` raw into URL path segments. The new `IncomingWebhookService.basePath` applies `encodeURIComponent` to both. This is the _correct_ practice (prevents path injection if IDs contain `/`, `#`, `?`), but it is a silent behavioural change: a `channelID` or `webhookId` containing `/` would previously create a multi-segment path (`groups/a/b/incoming-webhooks`) and now becomes a single encoded segment (`groups/a%2Fb/incoming-webhooks`). In practice WuKong group IDs and webhook UUIDs don't contain special characters, so risk is low. Worth a one-line comment in the service noting the intentional encoding, so future readers don't "fix" it back.

### P2 — Stale datasource mock in `WebhookEditModal.test.tsx` (`packages/dmworkbase/src/Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx:89-93`)

The test still mocks `WKApp.dataSource.channelDataSource.createIncomingWebhook` / `updateIncomingWebhook`, but the modal now calls `IncomingWebhookService.create` / `update` directly (via `APIClient.shared`). The datasource mock is dead code — the test works because the `APIClient` mock was correctly added alongside it. The dead mock should be removed to avoid confusing future test authors.

- `useChannelWebhookList` hook: `mountedRef` + sequence-ref pattern correctly prevents updates after unmount and discards stale responses; `scopeKey` change clears items and resets loading/error before triggering new load.
- `ChannelWebhookCard`: pure presentational component with no side effects; all mutation callbacks are injected as props; `testDisabled` derivation covers all four disable conditions (`!testAllowed || toggling || testingBlocked || cooling`).
- Panel container: all four mutation handlers (`handleToggle`, `handleTest`, `handleRegenerate`, `handleDelete`) route through `IncomingWebhookService` with correct argument order; error handling preserves existing toast-and-rethrow pattern for `wkConfirm`.
- CSS: all values use design tokens (`--wk-*` variables); no hardcoded colours, spacings, or radii introduced.
- Default avatar SVG: changed from grey link-icon to purple robot; purely visual, no runtime impact.
- Icon migration: Semi `IconEdit/IconRefresh/IconSend/IconDelete` → lucide-react `Edit3/RefreshCw/Send/Trash2`; purely visual swap.

## Data-Flow Trace

|---|---|---|
| `items` in card list | `IncomingWebhookService.list()` → `APIClient.shared.get()` → server | ✅ |
| `myUid` for permission check | `WKApp.loginInfo.uid` | ✅ |
| `creatorNames` for meta display | `WKSDK.channelManager.getSubscribes(channel)` → `subscriberDisplayName()` | ✅ |
| `manageable` prop | `canManageIncomingWebhook(item, {isManager, myUid})` — pure function, unchanged | ✅ |
| `testDisabled` prop | `canTestWebhook(item)` + toggling/cooling state from container | ✅ |
| `meta` (created/usage lines) | `item.created_at * 1000`, `item.call_count`, `item.last_used_at` — directly from list items | ✅ |

## Blindspot Checklist

- **C1 (dual-path parity)**: Clear. Old datasource methods fully migrated to `IncomingWebhookService`; both container and datasource delegate to the same service. No orphaned parallel path.
- **C2 (control-flow ordering)**: Clear. No nested/double-call risk; service methods are simple one-shot HTTP calls.
- **C3 (auth boundary)**: Clear. Permission model unchanged; server-side auth is the authority, UI guards are advisory only.
- **C4 (auth lifecycle)**: N/A — no auth hierarchy changes.
- **C5 (build/runtime)**: Clear. Static analysis confirms correct types and import resolution.
- **C6 (governance/docs)**: N/A — no docs/governance changes.

N/A — first review pass.

## Verdict: APPROVED

Clean, well-structured refactor that correctly centralises the webhook HTTP boundary, extracts list lifecycle into a reusable hook, and separates card rendering from container logic. Data-flow contracts are preserved, race conditions are properly guarded, and tests are aligned with the new architecture. The two P2 items are non-blocking cleanup suggestions.

[Octo-Q] verdict: APPROVE — no P0/P1 found; clean data-flow, correct race guards, complete migration.

yujiawei (approved):
# Code Review — PR #872 (octo-web)

Reviewed at head `cf1b9b20` against `main` (merge-base `50b6165b`). 11 files, +620/-308. This is a behavior-preserving refactor of the channel-webhook feature: HTTP moves into a new `IncomingWebhookService`, list loading moves into a scoped `useChannelWebhookList` hook with stale-request protection, the list item becomes a props-only `ChannelWebhookCard`, and the Semi `Select` member picker is replaced by a custom search+list picker.

Verified locally: focused Vitest suites pass (88/88 across ChannelWebhook + IncomingWebhookService + datasource), `i18n:check` clean, `git diff --check` clean.

## 1. Spec / scope
**Spec: ✅**

- Missing: none. All stated goals implemented — dedicated service, scoped hook with stale-request guard, props-only card, aligned list/empty/edit-form UI, permission and request semantics preserved.
- Over-build: none. (Leftover dead code is flagged below as quality, not new functionality.)
- Deviation: none. Endpoint parity holds — the new `IncomingWebhookService.basePath()` produces byte-identical URLs to the old `incomingWebhookBase()` for all six operations (group and thread scope), and now additionally `encodeURIComponent`s `groupNo`/`webhookId` (a safe hardening). `APIClient.shared` is the same singleton as `WKApp.apiClient` (`App.tsx:695`), so token / `X-Space-Id` / `baseURL` / interceptor behavior is unchanged.

## 2. Code Quality
**Quality: Approved** — no P0/P1. The findings below are all P2 (non-blocking).

- **[P2] Stale-request guard is not scope-aware** — `bridge/channelWebhook/useChannelWebhookList.ts:37`. The request sequence is global/temporal, not keyed to the scope it was issued in. A mutation handler (`handleToggle`/`handleDelete`/etc.) started in scope A calls `void load()` after its `await` resolves; that reload increments `requestSequenceRef` at call-time (after a switch to scope B) and can win over B's in-flight load, briefly rendering scope A's list in scope B. The `ChannelWebhookPanel` instance survives thread↔thread switches (memoized parent channel, no `key` at the mount sites), so the window is reachable. Downgraded from blocking because: it is a strict improvement over the previous code (which had no guard at all), there is no data loss, it self-corrects on the next load, and the trigger is narrow (a mutation in flight exactly across a scope switch). Suggested fix: capture `scopeKey` in the closure and bail if it changed, or increment the sequence when the request is *issued* rather than when the promise settles.

- **[P2] Storybook stories mock the datasource, which the panel no longer uses** — `Components/ChannelWebhook/ChannelWebhook.stories.tsx:94`. `installMock()` stubs `ds.incomingWebhooks` etc., but the refactored panel calls `IncomingWebhookService.list` → `APIClient.shared` directly. `ListAdmin` / `ListMember` / `EmptyState` will attempt real HTTP instead of rendering mock data. (The new `CardUi` story is fine — it renders the card directly.) Mock `APIClient.shared` in the stories, or drive the panel through the datasource.

- **[P2] Raw `` loses the WKAvatar broken-image fallback** — `Components/ChannelWebhook/ChannelWebhookCard.tsx:57`. The old code used `WKAvatar`, which has an `onError` handler that swaps in a default SVG. The empty-avatar case is still covered by `INCOMING_WEBHOOK_DEFAULT_AVATAR`, but a *non-empty but invalid* avatar URL now renders as a broken image. Add an `onError` fallback to the default avatar.

- **[P2] Custom picker drops Select accessibility** — `Components/ChannelWebhook/WebhookEditModal.tsx:387` and `index.css:297`. The new picker is a list of `<button>`s with no `listbox`/`option` roles, no `aria-selected`, and no arrow-key navigation, and the search input sets `outline: none` with no `:focus`/`:focus-visible` replacement (the option/selected-member buttons also have no focus style). Filtering and the `mention_uids` max-length validation on submit are preserved, so this is a11y polish, not a functional loss.

- **[P2] Dead code left behind.** The former Semi Select styles are now unreferenced — `index.css:307` (`__select-head`, `__option*`, `__option-tick`, `__member-tag`, `__select`). The six `IncomingWebhook*` methods on `ChannelDataSource` (`dmworkdatasource/src/datasource.ts:248`) now only delegate to the service and have no remaining production callers. Removing both would keep the boundary clean.

- **[P2] New tests don't cover the new-behavior edges** — `Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx`. The picker tests assert click/payload but not search filtering or the over-limit (51 members) submit block; the stale-request hook has no test at all. The `IncomingWebhookService` URL/envelope tests are solid.

## 3. Overall verdict
**APPROVE.** Spec ✅ and no P0/P1. The refactor is sound and the wire behavior is preserved; the P2s (chiefly the scope-aware reload guard and the stale Storybook mocks) are worth a follow-up but do not block merge.

## 4. Suggestions
- Make the reload guard scope-aware (capture `scopeKey`), or increment the sequence at issue-time.
- Point the Storybook stories at `APIClient.shared` so the list/empty stories render offline again.
- Add `onError` avatar fallback and a focus-visible style on the picker; consider `role="listbox"`/`option` + arrow-key nav.
- Delete the dead CSS and the now-unused datasource webhook methods.

## 5. Additional notes
- `IncomingWebhookService.list` now normalizes `list | items | webhooks` envelopes (superset of the old `list`-only) — a safe broadening.
- `this` binding is fine: the service is always called as `IncomingWebhookService.method(...)`, never destructured.

PR #874, opened 2026-07-17
fix(docs): keep table row height stable when hovering a column border

ploy-elison:
## Problem

Hovering a table column's **right border** arms prosemirror-tables' `columnResizing`, which mounts a `.column-resize-handle` widget (also classed `.ProseMirror-widget` by prosemirror-view) at the **end of the cell's `.octo-cell-clip` content hole** — i.e. as its last DOM child.

The compact-row rule (`SCHEMA_VERSION 19`, XIN-1289) zeroes the trailing block's margin with a bare `:last-child`. When the widget mounts, it steals that `:last-child` identity, so the real trailing `<p>` springs its `margin-bottom` (`1em ≈ 15px`) back — and **every affected row jumps ~15px taller for as long as the pointer stays on the border**, snapping back on leave. It's a pure render transient: the stored `tableRow.height` / `--octo-row-h` never change.

## Fix

Match the first/last real **content block** instead of the first/last DOM child:

```css
.octo-cell-clip > :nth-child(1 of :not(.ProseMirror-widget))      { margin-top: 0; }
.octo-cell-clip > :nth-last-child(1 of :not(.ProseMirror-widget)) { margin-bottom: 0; }
```

A mounted widget no longer displaces the margin-zero target. Chosen over `:last-of-type`, which over-matches a mixed-type cell (a trailing `<p>` before a `<ul>` would each be a last-of-type and both get zeroed, collapsing the gap between them). `:nth-child(of S)` is Baseline in the Chromium/evergreen targets.

## Verification

Real Chromium (Playwright), added as a PROD-faithful repro (`packages/docs/dev/colhover.harness.tsx` + `dev/run-colhover.mjs`) using the exact production table wiring — `Table.configure({ resizable: true, handleWidth: 12, cellMinWidth: 25 })`, `TableCellView` node views, real `styles.css`. It parks a real pointer on the inter-column border to arm `columnResizing` and asserts no row grows while the widget is mounted.

- **Before the fix** (`:last-child`): border hover → widget mounts → short row `37px → 52px` (**+15px**), trailing `<p>` `margin-bottom` `0 → 15px`. ❌
- **After the fix**: border hover → widget mounts → row stays `37px`, margin stays `0px`; pointer-away restores cleanly. ✅

Existing `TableRowClipPreview` unit test still passes; typecheck shows only pre-existing, unrelated errors.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,7 @@ packages/docs/dev/reorder-out/
 packages/docs/dev/handle-ux-out/
 packages/docs/dev/coexist-out/
 packages/docs/dev/multipointer-out/
+packages/docs/dev/colhover-out/
 
 # Loop V1 内部设计文档仅保留在本地 workspace，不纳入版本库
 docs/loop-v1/
--- a/packages/docs/colhover.html
+++ b/packages/docs/colhover.html
@@ -1,0 +1,17 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Column-hover row-jump repro (XIN-1322)</title>
+    <style>
+      html, body, #root { height: 100%; margin: 0; }
+      /* Make the column-resize handle visible (production accent is a faint hover gradient). */
+      .column-resize-handle { background: rgba(66, 99, 235, 0.6) !important; z-index: 20; }
+    </style>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="module" src="/dev/colhover.harness.tsx"></script>
+  </body>
+</html>
--- a/packages/docs/dev/colhover.harness.tsx
+++ b/packages/docs/dev/colhover.harness.tsx
@@ -1,0 +1,128 @@
+// PROD-faithful repro for XIN-1322 (column-hover row jump). Same editor wiring as production
+// extensions.ts (real Table.configure({ resizable, handleWidth, cellMinWidth }), TableCellView node
+// views, real editor/styles.css, wrapped in .octo-theme .octo-prose), so the browser exercises the
+// exact CSS the tester hit. The defect: hovering a column's RIGHT border arms prosemirror-tables'
+// columnResizing, which mounts a `.column-resize-handle` (also `.ProseMirror-widget`) DIV at the END
+// of the cell's `.octo-cell-clip` content hole. A bare `:last-child` margin-zero rule then loses the
+// trailing <p>, whose `margin-bottom` (1em) springs back and pushes every row ~15px taller until the
+// pointer leaves the border. window.__colHoverHarness exposes the seams the runner drives.
+import { useEffect, useRef } from 'react'
+import { createRoot } from 'react-dom/client'
+import * as Y from 'yjs'
+import { Editor } from '@tiptap/core'
+import StarterKit from '@tiptap/starter-kit'
+import Collaboration from '@tiptap/extension-collaboration'
+import { Table } from '@tiptap/extension-table'
+import TableHeader from '@tiptap/extension-table-header'
+import TableCell from '@tiptap/extension-table-cell'
+import type { Node as PMNode } from '@tiptap/pm/model'
+import { TableRowHeight, TableRowResize } from '../src/editor/TableRowHeight.ts'
+import { TableCellView } from '../src/editor/TableCellView.ts'
+import '../src/editor/styles.css'
+
+const FIELD = 'default'
+
+function makeEditor(ydoc: Y.Doc, element: HTMLElement): Editor {
+  return new Editor({
+    element,
+    extensions: [
+      StarterKit.configure({ undoRedo: false }),
+      Collaboration.configure({ document: ydoc, field: FIELD }),
+      // Exact production table config (extensions.ts:242) — resizable columns with the #749 grab band.
+      Table.configure({ resizable: true, handleWidth: 12, cellMinWidth: 25 }),
+      TableRowHeight,
+      TableHeader.extend({ addNodeView() { return ({ node }: { node: PMNode }) => new TableCellView(node, 'th') } }),
+      TableCell.extend({ addNodeView() { return ({ node }: { node: PMNode }) => new TableCellView(node, 'td') } }),
+      TableRowResize,
+    ],
+  })
+}
+
+function rowsOf(editor: Editor): { pos: number; node: PMNode }[] {
+  const rows: { pos: number; node: PMNode }[] = []
+  editor.state.doc.descendants((node, pos) => {
+    if (node.type.name === 'tableRow') rows.push({ pos, node })
+    return true
+  })
+  return rows
+}
+
+// Two content-driven rows (NO explicit height → no clip; height is pure content), two columns each.
+// Each cell holds two paragraphs, so the trailing <p> is unambiguous and its margin-bottom is the
+// thing under test. Mixed-content cell in col 2 (a <p> then a <ul>) guards the `:last-of-type`
+// over-match trap: the nth-child(of :not(widget)) rule must zero only the true trailing block.
+const DOC =
+  '<p>column-hover repro — hover the border between the two columns of row 1</p>' +
+  '<table><tbody>' +
+  '<tr>' +
+  '<td><p>r1c1 line a</p><p>r1c1 line b</p></td>' +
+  '<td><p>r1c2 line a</p><ul><li><p>r1c2 bullet</p></li></ul></td>' +
+  '</tr>' +
+  '<tr><td><p>r2c1</p></td><td><p>r2c2</p></td></tr>' +
+  '</tbody></table>'
+
+function Harness(): React.ReactElement {
+  const ref = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    if (!ref.current) return
+    let ed: Editor | null = null
+    const mount = () => {
+      ed?.destroy()
+      ref.current!.innerHTML = ''
+      const ydoc = new Y.Doc()
+      ed = makeEditor(ydoc, ref.current!)
+      ed.commands.insertContent(DOC)
+    }
+    const firstCellDOM = (): HTMLElement | null => {
+      const row = rowsOf(ed!)[0]
+      if (!row) return null
+      const dom = ed!.view.nodeDOM(row.pos)
+      if (!(dom instanceof HTMLElement)) return null
+      return dom.querySelector('td') as HTMLElement | null
+    }
+    const harness = {
+      mount,
+      rowRect: (i: number) => {
+        const row = rowsOf(ed!)[i]
+        if (!row) return null
+        const dom = ed!.view.nodeDOM(row.pos)
+        if (!(dom instanceof HTMLElement)) return null
+        const r = dom.getBoundingClientRect()
+        return { top: r.top, height: r.height }
+      },
+      // The x of the border between column 1 and column 2 (right edge of row-1's first cell), and a y
+      // safely inside that row — where the runner parks the pointer to arm columnResizing.
+      colBorderPoint: () => {
+        const cell = firstCellDOM()
+        if (!cell) return null
+        const r = cell.getBoundingClientRect()
+        return { x: r.right, y: r.top + r.height / 2 }
+      },
+      // Is the columnResizing widget currently mounted? (the render transient's trigger)
+      hasResizeHandle: (): boolean => !!document.querySelector('.octo-prose .column-resize-handle'),
+      // Diagnostic: computed margin-bottom of the LAST real <p> in row-1's first cell — the value the
+      // rule protects. Stays ~0 with the fix even while a widget is mounted after it.
+      lastParaMarginBottom: (): number | null => {
+        const cell = firstCellDOM()
+        if (!cell) return null
+        const clip = cell.querySelector('.octo-cell-clip')
+        if (!clip) return null
+        const paras = clip.querySelectorAll(':scope > p')
+        const last = paras[paras.length - 1] as HTMLElement | undefined
+        if (!last) return null
+        return parseFloat(getComputedStyle(last).marginBottom)
+      },
+    }
+    ;(window as unknown as { __colHoverHarness: typeof harness }).__colHoverHarness = harness
+    return () => { ed?.destroy() }
+  }, [])
+
+  return (
+    <div className="octo-theme" style={{ padding: 40, height: '100vh', boxSizing: 'border-box' }}>
+      <div className="octo-prose" ref={ref} style={{ position: 'relative' }} />
+    </div>
+  )
+}
+
+createRoot(document.getElementById('root')!).render(<Harness />)
--- a/packages/docs/dev/run-colhover.mjs
+++ b/packages/docs/dev/run-colhover.mjs
@@ -1,0 +1,119 @@
+// Playwright driver for the column-hover row-jump gate (XIN-1322). Real Chromium, a real pointer
+// parked on the border between two columns to arm prosemirror-tables' columnResizing (which mounts
+// the `.column-resize-handle` / `.ProseMirror-widget` DIV at the end of the cell's `.octo-cell-clip`).
+// Reproduces the acceptance gate:
+//   CH01 — hovering the column border mounts the resize-handle widget (the transient's trigger fires).
+//   CH02 — while the widget is mounted, every row's rendered height stays within 1px of its rest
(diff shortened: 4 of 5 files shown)


lml2468 (approved):
## 🐙 PR #874 Review — `fix(docs): keep table row height stable when hovering a column border`

**Head:** `fa4be8578ca7` · **Base:** `main` · +279 −3 / 5 files (real fix: `styles.css` +14 −3; rest is a Playwright repro harness)
**Intent:** Hovering a table column's right border arms prosemirror-tables' `columnResizing`, which mounts a `.column-resize-handle` (`.ProseMirror-widget`) as the last child of `.octo-cell-clip`. The compact-row rule's bare `:last-child` then lets the widget steal the margin-zero identity, so the real trailing `<p>` springs its `margin-bottom` back → the row jumps ~15px taller while hovering (pure render transient; model height unchanged).

### Verification (local checkout @ `fa4be8578ca7`)
| Check | Result |
|---|---|

Surgical fix — only the two `.octo-cell-clip` first/last margin-zero selectors change; the +14/−3 is mostly the explanatory comment. No other rule keys on those selectors, so nothing needs mirroring.

**Scope check (I initially suspected the 3 non-CSS files were scratch pollution — disproven):** `packages/docs/dev/*.harness.tsx` + `run-*.mjs` + a co-located root `*.html` is the repo's **established Playwright-harness convention** — main already carries `coexist`/`presence`/`reorder`/`rowheight*` triplets (`coexist.html` + `dev/coexist.harness.tsx` + `run-coexist.mjs`, etc.). `colhover.{html,harness,run}` is a faithful new instance; `run-colhover.mjs:16` correctly serves `colhover.html`. Not scope creep. None of it is imported by prod/build (verified) — it's dev-only, matching the others.

### Gate 2 — Code quality: ✅
The fix is correct and precisely reasoned:
- `:nth-child(1 of :not(.ProseMirror-widget))` / `:nth-last-child(1 of :not(.ProseMirror-widget))` select the first/last **real content block**, skipping the transient resize widget → the trailing `<p>` keeps its zeroed margin whether or not the widget is mounted.
- For a normal cell (no widget) this is **behaviorally identical** to the old `:first-child`/`:last-child`, so no compact-row regression; it only diverges in the exact widget-mounted case that is the bug.
- The comment correctly justifies `:nth-child(of S)` over `:last-of-type` (which would over-match a mixed-type cell — a trailing `<p>` before a `<ul>` would each be a last-of-type and both get zeroed). `:nth-child(of S)` is Baseline in the Chromium/evergreen targets.
- **Verification method is right for the defect class:** a real-Chromium Playwright repro that parks a pointer on the border to arm `columnResizing` and asserts no row grows — a hover-transient CSS bug can't be caught by unit tests, so the harness is the correct proof (before: 37→52px; after: stays 37px).

Only the 2 intended selectors changed ✅ · no other rule depends on the old first/last-child ✅ · normal-cell behavior identical (no regression) ✅ · widget-exclusion correct (`:not(.ProseMirror-widget)`) ✅ · dev harness follows established convention, not wired to prod ✅ · build green ✅. No security/data surface (pure CSS).

### Non-blocking note
`:nth-child(of S)` requires Chromium 111+/Safari 16.4+/Firefox 113+ — Baseline 2023 and fine for this app's evergreen targets, but if any supported browser predates that, the leading/trailing margin-zero silently no-ops there (cosmetic only). Worth a glance against the repo's stated browser floor; not a blocker.

---

### ✅ Review Verdict: APPROVE @ `fa4be8578ca7`
A precise, well-documented, real-Chromium-verified CSS fix for the column-hover row-jump — behaviorally identical to the old rule for normal cells, diverging only in the widget-mounted case that was the bug. The dev harness follows the repo's existing Playwright-harness convention (I verified against the base tree). Build green, no regressions. Clean fix.

Jerry-Xin (approved):
## APPROVE @ `fa4be8578ca7`

Precise, well-reasoned CSS fix with an exceptional test harness. Byte-verified the fix and its edges.

### Root cause + fix — correct
Hovering a column's right border arms prosemirror-tables' `columnResizing`, which appends a `.column-resize-handle` (also `.ProseMirror-widget`) DIV as the **last DOM child** of `.octo-cell-clip`. The compact-row rule used a bare `:last-child` to zero the trailing block's margin, so the widget stole that match → the real trailing `<p>` regained its `margin-bottom: 1em` (~15px) and every row grew for the duration of the hover (pure render transient; model height unchanged).

The fix replaces both edge rules to select the first/last *real content block*, skipping the widget:
- `.octo-cell-clip > :nth-child(1 of :not(.ProseMirror-widget)) { margin-top: 0 }`
- `.octo-cell-clip > :nth-last-child(1 of :not(.ProseMirror-widget)) { margin-bottom: 0 }`

Verified: both rules (leading + trailing) are converted, and grep confirms **no bare `:first-child`/`:last-child` `.octo-cell-clip` rule is left unfixed**, so there's no residual path with the bug.

### Why the selector choice is right
- `:nth-child(of S)` is Baseline-supported in the Chromium/evergreen targets (documented in the code comment).
- Deliberately chosen over `:last-of-type`, which would **over-match a mixed-type cell** (a trailing `<p>` before a `<ul>` would each be a last-of-type and both get zeroed). The harness DOC includes exactly this mixed `<p>`+`<ul>` cell to guard the trap.

### Test harness — exceptional
`colhover.harness.tsx` is PROD-faithful (real `Table.configure({ resizable, handleWidth: 12, cellMinWidth: 25 })`, real NodeViews, real `styles.css`, wrapped in `.octo-theme .octo-prose`), and `run-colhover.mjs` drives real Chromium with a pointer parked on the inter-column border to actually arm `columnResizing`. Four acceptance gates: CH01 widget mounts (proves the transient fires), CH02 every row holds height within 1px while mounted, CH03 trailing `<p>` margin-bottom stays ~0 with the widget after it, CH04 pointer-away unmounts and fully restores. This is the right way to pin a render-transient layout bug.

### Non-blocking
- The repro HTML/harness/driver live under `packages/docs/dev/` with the out-dir added to `.gitignore` — consistent with the existing dev harnesses; fine.

### CI
Build green, osv/secret/dependency all pass; only the sentinel `code-review` check pending.

### Net: APPROVE
Correct, minimal, widget-aware selector fix for both cell-clip edges, over-match trap explicitly avoided, and a rigorous PROD-faithful test harness. No blocking issues.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #874 (octo-web)

## Summary

This PR fixes a visual glitch (XIN-1322) where hovering a table column border — which arms prosemirror-tables' `columnResizing` and mounts a `.column-resize-handle` widget (`.ProseMirror-widget`) as the last DOM child of `.octo-cell-clip` — caused the trailing `<p>` to lose its `:last-child` margin-zero rule, spring its `margin-bottom` back to `1em`, and push every row ~15 px taller until the pointer left the border. The fix swaps `:first-child` / `:last-child` for `:nth-child(1 of :not(.ProseMirror-widget))` / `:nth-last-child(1 of :not(.ProseMirror-widget))`, so the margin-zero identity tracks the real content block rather than the DOM position. A Playwright harness and driver are added to lock the acceptance gate.

## Verification

- ✅ **CSS selector correctness** — `:nth-last-child(1 of :not(.ProseMirror-widget))` selects the last child that is not a widget. Verified via `TableCellView.ts:38` that `contentDOM = clip` (the `.octo-cell-clip` div), and prosemirror-view appends widget decorations directly into the content hole. The selector correctly targets the trailing `<p>` regardless of whether a widget is mounted.
- ✅ **No regression for `:first-child` → `:nth-child(1 of :not(...))`** — symmetric analysis: if a widget is ever prepended, the first real content block still gets `margin-top: 0`. In the normal (no-widget) case, both selectors are equivalent.
- ✅ **Mixed-type cell guard** — the test doc includes a cell with `<p>` + `<ul>` to verify the rule doesn't over-match (unlike `:last-of-type`, which would zero both a trailing `<p>` and a `<ul>`).
- ✅ **Harness pattern parity** — the new `colhover.html` / `colhover.harness.tsx` / `run-colhover.mjs` triplet follows the same structure as existing dev harnesss (`reorder.html` / `run-reorder.mjs`, `rowheight.html` / `run-rowheight.mjs`, etc.).
- ✅ **`.gitignore` entry** — `packages/docs/dev/colhover-out/` is correctly added alongside existing `*-out/` entries.
- ✅ **No production code outside CSS** — the fix is a single CSS change; all other new files are dev-only test infrastructure.

Static analysis only at head `fa4be857`; build and tests not executed in this environment.

## Findings

### Nit — Browser support for `:nth-child(of S)` (`packages/docs/src/editor/styles.css:2075`)

`:nth-child(1 of :not(.ProseMirror-widget))` is the most modern CSS selector in `styles.css` — no `:is()`, `:has()`, or `:where()` is used elsewhere in the file. The `of S` argument to `:nth-child` requires Safari 16.4+ (March 2023), Chrome 111+, Firefox 113+. The comment correctly says "Baseline-supported in the Chromium/evergreen targets", but Safari users on macOS < Ventura would silently fall back to the original row-jump behaviour (graceful degradation, not a breakage). Worth confirming whether the product's browser-support matrix includes Safari < 16.4.

- **Data flow trace**: `TableCellView` constructor (`TableCellView.ts:37-40`) creates `div.octo-cell-clip` as `contentDOM`. Prosemirror-view appends decoration widgets (including `.column-resize-handle`) directly into the content DOM element. The CSS selector operates on exactly this parent-child relationship. No indirection or data transformation between the widget mount point and the CSS rule.
- **Widget positioning**: `.column-resize-handle` uses `position: absolute` with the `<td>` (position: relative) as its containing block, so it escapes the `.octo-cell-clip`'s `overflow: hidden` on height-constrained rows. The fix doesn't affect this layout.
- **Empty-cell edge case**: when the cell has no non-widget children, `:nth-child(1 of :not(.ProseMirror-widget))` matches nothing (no rule fires). In practice, ProseMirror always renders at least one block (`<p>`) even in empty cells, so this is unreachable.
- **Harness references production code**: `colhover.harness.tsx` imports the same `Table.configure({ resizable: true, handleWidth: 12, cellMinWidth: 25 })`, `TableCellView`, `TableRowHeight`, and `TableRowResize` as the production editor. The test validates the CSS in the real rendering context.

## Verdict: APPROVED

Targeted, well-documented CSS fix with a thorough regression harness. The selector change is correct and the degradation path on older browsers is graceful. The single nit (Safari 16.4+ requirement) is informational, not blocking.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. CSS selector fix is correct, data flow from TableCellView contentDOM → widget mount → CSS rule verified. Test harness follows existing patterns. Single nit on browser compatibility is non-blocking.

boris-clark (approved):
## Review — APPROVED

Reviewed at head `fa4be857` (base `main`, 5 files). The real fix is the two-selector change in `packages/docs/src/editor/styles.css`; the rest is dev tooling.

### Correctness — the fix is right
`.octo-cell-clip > :first-child / :last-child` → `:nth-child(1 of :not(.ProseMirror-widget)) / :nth-last-child(1 of :not(.ProseMirror-widget))`.

- Root cause matches: `columnResizing` appends `.column-resize-handle` (also `.ProseMirror-widget`) as the last child of the content hole, stealing `:last-child` from the trailing `<p>`, whose `margin-bottom` (1em) springs back → row grows ~15px. Selecting `1 of :not(.ProseMirror-widget)` skips the widget, so the real trailing block keeps its zeroed margin. ✓
- **Compact row height is preserved when NOT hovering**: with no widget mounted, `nth-child(1 of :not(widget))` resolves to the same element as `:first-child`, and likewise for the trailing block — behaviour is identical to the pre-fix rule. The compact effect is not weakened. ✓
- **Mixed-type cell** (`<p>` then `<ul>`): `:nth-last-child(1 of …)` matches only the true trailing block (the `<ul>`), so the inter-block gap is preserved. This correctly avoids the `:last-of-type` over-match trap called out in the description. ✓
- The replacement is a strict superset of the deleted rules — no zeroing declaration or compact behaviour was dropped.

### Browser compatibility — holds for all shipping targets
`:nth-child(An+B of S)` is Chromium 111+ / Safari 9+ / Firefox 113+. This app ships as Electron 26 (Chromium 116) and Tauri 1.5 (macOS WKWebView / Windows WebView2) — all support `of S`. No fallback needed.

### PR hygiene — dev harness files are fine to ship here
`colhover.html` / `dev/colhover.harness.tsx` / `dev/run-colhover.mjs` follow the repo's established convention: `coexist`, `reorder`, `presence`, and `rowheight*` already ship the same `.html` + `.harness.tsx` + `run-*.mjs` trio, with matching `*-out/` entries in `.gitignore`. Critically, `src/` never imports from `dev/` and `@octo/docs` exposes only `src/index.ts` — these files are reachable only via `vite.dev.config.ts` (dev server), so they never enter the production bundle consumed by `@octo/web`. No reason to strip them; they are the reproducible pixel gate for this class of bug.

### Regression coverage — acceptable as-is
jsdom does no layout, so the existing `TableRowClipPreview.test.ts` can only assert the DOM contract, not computed margins (the test file says as much). The Playwright harness (`run-colhover.mjs`, gates CH01–CH04) is the intended pixel-level regression check, matching how the other row-height behaviours are verified. Not wired into CI, but that is the codebase's deliberate split, not a gap introduced by this PR.

### Conclusion
No blocking issues. Correct fix, compatible with all shipping targets, hygiene consistent with repo convention. Approved for deploy → tester.

yujiawei (approved):
# Code Review — PR #874 (octo-web)

**Verdict: APPROVE.** The fix is correct, minimal, and well-motivated; its two load-bearing dependency claims hold up against the upstream `prosemirror-view` / `prosemirror-tables` source. Only non-blocking nits below.

## Spec compliance

- **Missing:** none — the change targets exactly the described defect (border-hover row jump).
- **Over-build:** none — the added Playwright harness (`colhover.harness.tsx` + `run-colhover.mjs` + `colhover.html`) follows the repo's established dev-harness convention (`coexist`, `reorder`, `multipointer`), so it reads as verification scaffolding, not scope creep.
- **Deviation:** none — implementation matches the stated approach.

## Correctness — verified against upstream

The fix rests on two claims, both confirmed at the pinned versions:

1. **The resize handle carries `.ProseMirror-widget`.** `prosemirror-tables@1.8.5` creates it via `Decoration.widget(pos, dom)` (`columnresizing.ts:427`) with **no** `raw` spec. `prosemirror-view@1.41.x` `WidgetViewDesc` adds `classList.add("ProseMirror-widget")` for every non-`raw` widget (`viewdesc.ts`). So `:not(.ProseMirror-widget)` reliably excludes it.
2. **The handle mounts as the last child of `.octo-cell-clip`.** Its position is `start + cellPos + nodeSize - 1` — the end of each cell's content hole, which `TableCellView` sets as `contentDOM` (`TableCellView.ts`). It loops over every row in the column, which explains why *every* affected row jumps, not just the hovered one.

`:nth-child(1 of :not(.ProseMirror-widget))` has the correct Selectors-L4 "filter-then-count" semantics: it selects the first *non-widget* child, so a mounted widget no longer displaces the margin-zero target. The `:last-of-type` rejection reasoning in the PR is sound (a trailing `<p>` before a `<ul>` would give two `:last-of-type` matches).

No other DOM node can steal the match: `prosemirror-dropcursor@1.8.2` appends its element to `editorView.dom.offsetParent` as `position:absolute` (`dropcursor.ts:105-107`), not into the clip; comment-highlight and find-replace use `Decoration.inline` (wrap text inside a block, never a direct child); gapcursor is a between-block selection, not a populated-cell child.

## Non-blocking nits (P2)

- **Selector specificity rises** from `(0,2,0)` to `(0,3,0)`. Harmless here (no competing rule on these elements), but if margin parity ever matters, `:nth-child(1 of :where(:not(.ProseMirror-widget)))` keeps the old weight.
- **Harness doesn't assert the mixed-cell claim it sets up.** The DOC seeds a `<p>` + `<ul>` cell to guard the `:last-of-type` over-match, but the assertions only check hover-vs-rest height/margin deltas — nothing verifies the inter-block gap in the mixed cell stays intact. Consider one explicit spacing assertion so that guard is actually exercised.
- **Regression coverage is a manual harness, not CI.** `run-colhover.mjs` isn't wired into any package script or CI workflow (same as the sibling harnesses), and the existing `TableRowClipPreview` unit test doesn't cover the widget-mounted case (jsdom has no layout). The fix is therefore protected by a dev-run-only gate. Fine given the repo's current posture, but worth a follow-up if these layout invariants keep regressing.

## Browser support

Dropping `:first-child`/`:last-child` with no fallback relies on `:nth-child(of S)` (Baseline 2023). This matches the repo's evergreen posture — `:has()` already ships unguarded in other CSS, and there's no `browserslist`/build target pinning legacy engines — so it isn't a regression risk for the supported targets.

PR #771, opened 2026-07-15
feat(loop): add built-in skills install tip to runtime install prompt

superalsrk:
## Goal

Closes #770.

The **Runtime → Add computer** install prompt only told the Bot to install, authenticate and start the octo-daemon runtime. It never installed the bundled coding-agent skill, so a freshly connected machine had a runtime online but no `octo-loop` skill available to the local agent. This change makes the generated prompt fully provision a machine in one pass.

## Changes

- `packages/dmloop/src/i18n/en-US.json` — add step 4 to `loop.runtime.installPrompt`: "Refer to octo-daemon --help to install the built-in skills."
- `packages/dmloop/src/i18n/zh-CN.json` — add the same step 4 (`参考 octo-daemon --help 安装内置的 skills。`).

Only i18n copy changes; no command-builder or runtime logic changes.

## Verification

- `en-US.json` and `zh-CN.json` both parse as valid JSON.
- Open Loop → Runtime → **Add computer**, click **Copy install instructions**, and confirm the generated prompt now includes a fourth step instructing the Bot to install the built-in skills via `octo-daemon --help`, placed after the daemon start step and before the final online-confirmation line.

PR #884, opened 2026-07-18
feat(dmloop): add env variable config to expert config page

### LINKED ISSUE #883: Feature: add environment variable configuration to expert config page
## Problem

Experts (agents) in Loop need to carry configuration such as API keys and
environment parameters that individual runs require — especially independent
runs like AI image-generation APIs. This configuration must work across
devices and runtimes and must not depend on any single execution device.
It also contains sensitive secrets, so it must be handled accordingly.

## Proposed solution

Add an "Environment variables" section to the expert config page, placed
below the "Skills" section:

- Owner-only reveal-then-edit flow. Entering the panel does not auto-fetch
  plaintext; the owner must explicitly click "Reveal & edit".
- When not revealed, show only the configured-variable count plus an audit
  hint; clicking "Reveal & edit" is the only path into the editor.
- Key/value editor with password-masked value inputs, add / edit / delete
  rows, and duplicate-key validation on save.
- Non-owners (and any backend 403) see a masked read-only state with just
  the variable count — no reveal button, no error toast.
- Revealed plaintext lives only in component memory and is cleared when
  leaving the section or switching experts; it is never persisted to
  localStorage/sessionStorage.

## Scope

Frontend only (dmloop package). The backend env endpoints
(`GET`/`PUT /agents/{id}/env`) and the frontend API client already exist;
this issue covers the missing UI and its wiring.

## Additional context

Implements the design reviewed internally in LOOP-28. Styling follows the
dmloop loop skin (not raw Semi Design defaults).

superalsrk:
## Summary

Adds an "Environment variables" section to the expert (agent) config page in
the dmloop package, placed below the "Skills" section. Owner-only
reveal-then-edit flow with masked values; plaintext is fetched only on an
explicit reveal and cleared when leaving; non-owners (and any backend 403)
see a masked read-only state. Frontend only — the backend env endpoints and
the API client already exist.

Closes #883

## Changes

- `packages/dmloop/src/panel/AgentDetailPage.tsx`: new `env` config sub-nav
  item below Skills; reveal two-state flow (no auto-fetch on enter); key/value
  editor with password-masked value inputs, add / edit / delete rows, and
  duplicate-key validation; owner-only gating with graceful read-only on 403;
  revealed plaintext cleared on section leave / expert switch.
- `packages/dmloop/src/panel/agentDetail.css`: new `.loop-adp__env-*` styles
  reusing the loop skin (purple focus ring, mono font); no raw Semi defaults.
- `packages/dmloop/src/api/types.ts`: add optional `custom_env_key_count` to
  the `Agent` type.
- `packages/dmloop/src/i18n/{zh-CN,en-US}.json`: strings for the env flow.

## Verification

- `tsc --noEmit` (skipLibCheck) on the changed files: 0 errors; both i18n
  JSON files validated.
- Two-axis code review (standards + spec) run; the one finding — reveal/save
  403 should degrade to read-only rather than toast — was fixed.
- Recommended manual smoke: owner reveal → edit → save; duplicate-key blocked;
  leaving and returning clears revealed plaintext; non-owner sees masked
  read-only with no reveal button.


lml2468 (approved):
## 🐙 PR #884 Review — `feat(dmloop): add env variable config to expert config page`

**Head:** `174342d8d335` · **Base:** `main` (merge-base `6208748cdade`) · +231 −4 / 5 files · `mergeable: MERGEABLE`
**Intent (SECURITY-sensitive):** Add an "Environment variables" section to the expert config page — an owner-only, reveal-then-edit secrets editor with masked values. Plaintext is fetched only on explicit reveal and cleared on leave; non-owners / backend-403 see a masked read-only state. Frontend-only against pre-existing `/agents/:id/env` GET+PUT endpoints. Closes #883.

### Verification (local checkout @ `174342d8d335`, pnpm installed)
| Check | Result |
|---|---|
| `turbo run build --filter=@octo/web` | ✅ pass (typecheck clean) |
| `vitest run` (dmloop full) | ✅ 63/63, no regression |
| API layer (`getAgentEnv`/`updateAgentEnv`, `LoopApiError`) | ✅ exist; GET `/agents/:id/env`, PUT with `{custom_env}` |
| i18n keys (both locales) | ✅ all new `agent.env*` keys symmetric; `envMaskedCount` `{{count}}` interpolation called with `values:{count}` |

Delivers exactly the described owner-only reveal-then-edit flow, frontend-only, no backend/endpoint changes. New `env` sub-nav below Skills, key/value editor with add/edit/delete, dup-key validation, owner gating, 403 read-only degradation, plaintext cleared on leave. No scope creep.

### Gate 2 — Code quality & security: ✅ (secrets surface, checked closely)
- **No plaintext auto-fetch.** `getAgentEnv` has a **single call site** — inside `handleRevealEnv`, triggered only by the owner's explicit "Reveal & edit" click. Entering the panel with `envRevealed===null` shows only the masked count; no GET fires. Verified by grep (one call site) — this is the load-bearing privacy property and it holds.
- **Plaintext doesn't linger.** Cleared on section-leave (`section !== "env"` effect → `setEnvRevealed(null)`) and on expert switch (`agentId` effect → `setEnvRevealed(null)`). Save-403 also collapses the editor (`setEnvRevealed(null)`). So revealed secrets don't survive tab-switch or navigation.
- **Owner gate + defense in depth.** `canEdit = agent.owner_id === myMemberId` (member-id, the bridge-correct signal — not the multica id), with the backend owner-only endpoint as the real enforcement and a **403 → read-only masked, no toast** degradation as backstop (both reveal and save paths). Mirrors the existing `mcp_config` redaction pattern in this same file.
- **Dup-key detection matches the submit path.** The check (`keys = entries.map(trim).filter(Boolean)`, `new Set(keys).size < keys.length`) uses the *same* trimming as `entriesToEnvMap` (trim key, drop empty) — so a flagged duplicate is exactly what would be sent; no silent last-write-wins masking a dup. Correct.
- **Clear-all & dirty are correct.** `envDirty` compares `stableEnvJson(entriesToEnvMap(revealed))` vs `stableEnvJson(original)` (key-sorted, so row-reorder isn't dirty; content is). Removing all rows → `{}` ≠ non-empty original → dirty → save sends `{}` → clears all. Save round-trips `custom_env` back into state and `refreshAgent()` refreshes `custom_env_key_count` for the masked view.
- **Value inputs are `mode="password"`** (masked in the DOM). The plaintext necessarily lives in React state while editing — inherent to an editable secrets field and acceptable; it's the reveal-gating + clear-on-leave that bound exposure.

### Non-blocking notes
- **No automated tests for the new flow.** This is a security-sensitive surface (reveal-gating, clear-on-leave, 403 degradation, dup-key) and none of it is unit-covered. Not a merge blocker for a frontend-only change whose invariants I've traced, but a reveal-gating + clear-on-leave regression test is the highest-value fast-follow — those are exactly the properties a future refactor could silently break.
- **`envIdRef` is never reset across reveals** — harmless (it's a monotonic React key source; ids need only be unique within a mounted list, and it only ever increments), noting it so it isn't mistaken for state that must reset.
- **Backend is the authority for redaction.** The frontend masks and gates, but the real secret protection (GET returning plaintext only to authorized owners, PUT audit logging) lives server-side — confirm those endpoints enforce owner-only + audit as the copy claims ("Revealing and saving are both recorded in the audit log").

no plaintext auto-fetch (single reveal-gated call site) ✅ · plaintext cleared on leave/switch/403 ✅ · owner gate + 403 read-only degradation both paths ✅ · dup-key check consistent with submit map ✅ · clear-all → sends {} ✅ · dirty ignores row-reorder ✅ · count refreshed post-save ✅ · i18n symmetric + interpolation correct ✅ · build + dmloop 63/63 green ✅.

---

### ✅ Review Verdict: APPROVE @ `174342d8d335`
Spec ✅ and quality/security ✅ — a carefully-built owner-only secrets editor: reveal-on-demand (no auto-fetch, single call site), plaintext cleared on leave/switch, owner-gated with graceful 403 read-only, dup-key validation consistent with the submit path, correct clear-all/dirty semantics. Build green, dmloop 63/63, i18n symmetric. Non-blocking: add a reveal-gating + clear-on-leave regression test (security surface currently untested), and confirm the backend `/agents/:id/env` endpoints enforce owner-only + audit as the UI copy promises. No blockers.

Jerry-Xin (dismissed):
The PR is relevant to `octo-web`, but secret lifecycle races and a destructive metadata fallback must be fixed before merge.

### 🔴 Blocking

- 🔴 Critical — An in-flight reveal can restore plaintext after it should have been cleared. `handleRevealEnv()` unconditionally writes its response at [AgentDetailPage.tsx:253](/Users/mlamp/.openclaw/tmp/pr-review-_9i_769c/repo/packages/dmloop/src/panel/AgentDetailPage.tsx:253), while section or agent changes merely clear the current state at lines 132–161. If the user navigates away or switches experts before the request completes, the late response repopulates memory. On an expert switch, another expert’s secrets can consequently appear in the current editor and potentially be saved to the wrong expert. Cancel or invalidate requests using the agent ID and section/tab generation before committing results. Apply equivalent stale-response protection to saving and refreshing.

- 🔴 Critical — Missing `custom_env_key_count` is treated as proof that no variables exist at [AgentDetailPage.tsx:382](/Users/mlamp/.openclaw/tmp/pr-review-_9i_769c/repo/packages/dmloop/src/panel/AgentDetailPage.tsx:382). The field is optional, and the existing `has_custom_env` metadata is ignored. If an older/cached response has `has_custom_env: true` but no count, the UI enters the empty flow at lines 599–607 without fetching existing values. Saving then performs a full replacement and silently deletes the existing secrets. Unknown count must remain a masked/reveal-required state whenever `has_custom_env` is true.

### 💬 Non-blocking

- 🟡 Warning — Switching from Config to Profile does not clear plaintext because the cleanup effect depends only on `section`, not `tab` ([AgentDetailPage.tsx:154](/Users/mlamp/.openclaw/tmp/pr-review-_9i_769c/repo/packages/dmloop/src/panel/AgentDetailPage.tsx:154)). This conflicts with the stated “cleared when leaving” behavior.

- 🟡 Warning — Inputs and row controls remain editable while `envSaving` is true. The save response then replaces the editor state, so changes made during the request can be silently lost.

- 🟡 Warning — No tests cover the security-sensitive state transitions. Add component tests for navigation during reveal/save, expert switching, 403 degradation, missing count with `has_custom_env`, duplicate keys, and plaintext cleanup.

- 🟡 Warning — The new CSS introduces hard-coded spacing, font sizes, and color fallbacks at [agentDetail.css:369](/Users/mlamp/.openclaw/tmp/pr-review-_9i_769c/repo/packages/dmloop/src/panel/agentDetail.css:369), contrary to the project’s token requirements.

### ✅ Highlights

- Plaintext is fetched only after explicit user action.
- Duplicate keys are validated after normalization.
- Backend 403 responses degrade to a masked state without exposing values.

Jerry-Xin (dismissed):
The PR is relevant to `octo-web`, but secret lifecycle races and a destructive metadata fallback must be fixed before merge.

### 🔴 Blocking

- 🔴 Critical — An in-flight reveal can restore plaintext after it should have been cleared. `handleRevealEnv()` unconditionally writes its response at [AgentDetailPage.tsx:253](packages/dmloop/src/panel/AgentDetailPage.tsx:253), while section or agent changes merely clear the current state at lines 132–161. If the user navigates away or switches experts before the request completes, the late response repopulates memory. On an expert switch, another expert’s secrets can consequently appear in the current editor and potentially be saved to the wrong expert. Cancel or invalidate requests using the agent ID and section/tab generation before committing results. Apply equivalent stale-response protection to saving and refreshing.

- 🔴 Critical — Missing `custom_env_key_count` is treated as proof that no variables exist at [AgentDetailPage.tsx:382](packages/dmloop/src/panel/AgentDetailPage.tsx:382). The field is optional, and the existing `has_custom_env` metadata is ignored. If an older/cached response has `has_custom_env: true` but no count, the UI enters the empty flow at lines 599–607 without fetching existing values. Saving then performs a full replacement and silently deletes the existing secrets. Unknown count must remain a masked/reveal-required state whenever `has_custom_env` is true.

### 💬 Non-blocking

- 🟡 Warning — Switching from Config to Profile does not clear plaintext because the cleanup effect depends only on `section`, not `tab` ([AgentDetailPage.tsx:154](packages/dmloop/src/panel/AgentDetailPage.tsx:154)). This conflicts with the stated “cleared when leaving” behavior.

- 🟡 Warning — Inputs and row controls remain editable while `envSaving` is true. The save response then replaces the editor state, so changes made during the request can be silently lost.

- 🟡 Warning — No tests cover the security-sensitive state transitions. Add component tests for navigation during reveal/save, expert switching, 403 degradation, missing count with `has_custom_env`, duplicate keys, and plaintext cleanup.

- 🟡 Warning — The new CSS introduces hard-coded spacing, font sizes, and color fallbacks at [agentDetail.css:369](packages/dmloop/src/panel/agentDetail.css:369), contrary to the project’s token requirements.

### ✅ Highlights

- Plaintext is fetched only after explicit user action.
- Duplicate keys are validated after normalization.
- Backend 403 responses degrade to a masked state without exposing values.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #884 (octo-web)

## Summary

This PR adds an environment-variable configuration panel to the Expert (agent) detail page. Owners can reveal, edit, add, and delete runtime environment variables (e.g. `ANTHROPIC_API_KEY`) through a masked-by-default UI that only fetches plaintext values on explicit user action. Non-owners see a read-only masked summary. The change spans the main component (`AgentDetailPage.tsx`), API types, two i18n locale files, and CSS.

The implementation is well-structured: reveal-then-edit flow with audit-friendly design, proper state cleanup on navigation and agent switch, 403 graceful degradation, stable dirty-check via sorted-key serialization, and shared HTTP client usage throughout. No P0/P1 issues found; one P2 item below.

## Verification

- ✅ **Auth/visibility gating** — `canEdit` (owner check) gates all edit UI paths; `envForbidden` handles backend 403 by falling back to masked view. Non-owner path at `AgentDetailPage.tsx:596` shows only count + redacted hint.
- ✅ **State cleanup** — `envRevealed`, `envOriginal`, `envForbidden` are cleared on agentId change (`:119–121`) and on leaving the env section (`:155–161`). Plaintext never persists across navigation.
- ✅ **Async error handling** — `handleRevealEnv` (`:255–265`) and `saveEnv` (`:275–290`) both use try/catch/finally with loading flags (`envRevealing`, `envSaving`) and proper Toast feedback or 403 downgrade. No fire-and-forget promises.
- ✅ **Shared client usage** — `getAgentEnv` and `updateAgentEnv` (`agentApi.ts:91–98`) route through `httpGet`/`httpPut`, which attach auth, space headers, and structured error conversion. No raw fetch or hardcoded URLs.
- ✅ **Dirty-check correctness** — `stableEnvJson` sorts keys before serialization, so row reorder doesn't trigger false dirty. `entriesToEnvMap` strips empty keys consistently for both dirty comparison and save payload.
- ✅ **i18n parity** — All 12 new keys (`envIntro` through `envRevealFailed`) present in both `en-US.json` and `zh-CN.json` with matching key names and correct `{{count}}` interpolation syntax.
- ✅ **CSS** — 29 lines of well-scoped styles under `.loop-adp__env-*` namespace. Reuses existing `--loop-mono-font` and `--semi-color-*` tokens. No runtime concern.

## Findings

No P0/P1 issues; one P2 item below.

### P2 — Entries with a value but empty key silently dropped on save (`packages/dmloop/src/panel/AgentDetailPage.tsx:57`)

`entriesToEnvMap` strips empty-key rows (`if (k) map[k] = e.value`), and `saveEnv` only validates duplicate non-empty keys. If a user types a secret value but forgets to fill the key name, the row vanishes after save with no toast or inline validation warning. The user may not realize their value was discarded.

Consider either adding an inline indicator on empty-key rows (e.g. red border on the key `Input`), or emitting a toast when entries are silently discarded during `entriesToEnvMap`. Not a blocker — the behavior is safe (no invalid data reaches the backend), just a minor UX gap.

- `Input mode="password"` on value field (`:635`) is a valid Semi UI prop; values are masked with a user-controlled reveal toggle, consistent with the "masked by default" design.
- `envIdRef` monotonically increasing counter prevents React key collisions across add/remove cycles.
- Duplicate-key check (`:278`) correctly compares trimmed keys after filtering empties.
- `saveEnv` awaits `refreshAgent()` after success, so `custom_env_key_count` updates in the UI.
- Empty-state direct-add path (`:614–619`) skips the GET call (no secrets to reveal) and correctly initializes `envOriginal` to `{}`.
- Backend contract: PUT sends `Record<string, string>` (full replacement), matching the API helper signature.

## Verdict: APPROVED

Well-designed feature addition with proper security posture (masked by default, owner-only editing, state cleanup), clean async handling, and consistent i18n. The single P2 (silent discard of empty-key entries) is a minor UX improvement, not a correctness issue.

[Octo-Q] verdict: APPROVE — no P0/P1 found; single P2 (silent discard of empty-key env entries) is non-blocking UX polish. Data flow traced end-to-end: API helpers → state management → rendering all consistent. Auth gating verified at both UI (`canEdit`) and API (`LoopApiError` 403 handling) layers. State cleanup on navigation/agent-switch confirmed.

lml2468 (changes requested):
## 🐙 PR #884 Re-Review (correction) — REQUEST_CHANGES @ `174342d8d335`

**I'm correcting my APPROVE.** Jerry-Xin flagged an async race and a secret-wipe path; I traced both against the code. **One is a real data-integrity risk on a secrets surface that I under-weighted, and I'm flipping to CR on it.** Honest adjudication of all four items:

### 🔴 P2 (BLOCKER) — empty-state edit can silently WIPE existing secrets (data loss)

The masked-vs-empty gate rests **solely** on `envKeyCount = agent.custom_env_key_count ?? 0`, and the empty-state path enters direct edit **without ever calling `getAgentEnv`**:
```ts
onClick={() => { setEnvOriginal({}); setEnvRevealed([{ id: …, key: "", value: "" }]); }}
```
Then `updateAgentEnv` PUTs the **full map (replace semantics)** — omitted keys are deleted server-side.

**Failure chain:** if the backend ever reports `custom_env_key_count` **absent or 0 while env actually exists** → owner sees "No environment variables yet" → adds one var → saves → PUT sends `{newKey}` only → **all pre-existing secrets are silently deleted**, and the owner never saw them (no reveal on this path).

**Why this is credible, not theoretical:** the `Agent` type carries a **separate `has_custom_env?: boolean`** field — and the env UI **ignores it entirely** (grep: 0 references in `AgentDetailPage.tsx`). That boolean exists precisely as an independent "env is configured" signal; a backend that sets `has_custom_env: true` but doesn't populate `custom_env_key_count` (older backend, or a count computed on a different path) drives exactly this wipe. On a **secrets** surface with **replace-PUT** semantics, an unguarded false-empty baseline is a P2 data-loss defect.

**I can't fully confirm reachability frontend-side** (it depends on whether the backend ever returns that state) — but the frontend has the defense sitting unused. **Fix:** gate the empty-state on **both** signals — treat env as "configured" if `has_custom_env || envKeyCount > 0`, and when configured, require a reveal (`getAgentEnv`) before any edit/save so the baseline is the real map, never `{}`. That makes the wipe unreachable regardless of count reliability. Add a regression test: "has_custom_env=true, count absent → must not offer blind empty-edit."

### 🟡 (was 🔴 #1) — reveal has no request guard; stale-response, but low-reachability plaintext-leak
`handleRevealEnv` sets `setEnvRevealed` on resolve with no agentId/request-token recheck. Jerry-Xin frames it as "switch experts → A's plaintext into B." I traced the navigation: `AgentDetailPage` is opened via `WKApp.routeRight.push(<AgentDetailPage agentId={id}/>)` — a **stack push of a new instance per expert**, not an in-place `agentId` prop swap. So the cross-expert-leak scenario is largely unreachable (each expert is its own mounted page; the reveal closure captures that page's `agent.id`). What *is* real: a stale reveal resolving after the user left the env section could momentarily set `envRevealed` — but the section-leave effect nulls it. **Downgrade to 🟡:** add a request-token/`agent.id` recheck before `setEnvRevealed` for defense in depth, but I couldn't construct a reachable cross-expert leak through the real router. (If you can show one — an in-place agentId swap path — I'll re-raise to 🔴.)

### 🟡 tab-switch clears plaintext — actually already handled
The `section !== "env"` effect nulls `envRevealed`/`envOriginal` on leave; browser tab-switch doesn't unmount. This one reads as already-correct to me — the plaintext *is* cleared on section-leave. Non-issue unless there's a specific path I'm missing.

### 🟡 inputs editable during save — minor
Key/value `Input`s aren't `disabled` while `envSaving`; the Save button is disabled (`!envDirty || envSaving`), so a double-save is blocked, but edits during the in-flight PUT are discarded on the success round-trip (`setEnvRevealed(envMapToEntries(env))`). Low-severity polish: disable inputs while saving. Non-blocking.

### What still stands from my review
The security *design* is otherwise sound and verified: no plaintext auto-fetch (single reveal-gated call site), clear-on-leave, owner-gate + 403 read-only degradation, dup-key consistency, build + dmloop 63/63. The blocker is specifically the **empty-state false-baseline wipe**.

---

### ❌ Review Verdict: REQUEST_CHANGES @ `174342d8d335`
One data-integrity P2 I under-weighted in my APPROVE: the empty-state edit path trusts `custom_env_key_count` alone, ignores the available `has_custom_env` signal, and enters a blind `{}` baseline that a replace-PUT can turn into a **silent secret wipe** if the backend reports count-absent-while-env-exists. Gate the empty state on `has_custom_env || count>0` and require a reveal before edit/save when configured. The reveal request-guard is a 🟡 (low-reachability given push-per-expert navigation); tab-switch clearing already works. Credit to Jerry-Xin for the wipe path — my APPROVE checked the reveal/clear/gate invariants but under-weighted the count-reliability dependency on a destructive secrets operation.

yujiawei (changes requested):
# Code Review — PR #884 (octo-web)

Reviewed at head SHA `174342d8d33518871770b05f406503ca9233f9db`. Frontend-only change adding an owner-only "Environment variables" section (reveal-then-edit, password-masked) to the dmloop expert config page.

Two independent analysis passes (line-level + long-context/cross-file) were folded into this review. Both converged on the same top issue; findings are grouped by confidence below.

## 1. Spec compliance

**Spec: ✅**

Against issue #883's requirements:
- Owner-only reveal-then-edit, no auto-fetch on entry — ✅ (`envRevealed === null` until explicit reveal).
- Masked count + audit hint when not revealed — ✅.
- Key/value editor with masked value inputs, add/edit/delete, duplicate-key validation on save — ✅.
- Non-owner and 403 → masked read-only, no reveal button, no error toast — ✅.
- Plaintext lives only in memory, never persisted to localStorage/sessionStorage — ✅ verified (no storage writes; `mode="password"` on value inputs).
- Frontend-only, backend endpoints reused — ✅.

No over-build (no unrequested flags/endpoints) and no missing requirement. Scope is clean.

## 2. Code Quality

### P1 — Stale async write across expert switch / section leave

`AgentDetailPage.tsx` (`handleRevealEnv` ~L253-259, `saveEnv` ~L284-287). Neither async handler re-checks that `agent.id` is still the one it started with after the `await`. This component is pushed **without a `key`** (`AgentPage.tsx:67`, `SquadDetailPage.tsx:148` — contrast `IssueDetailPage` which is always pushed with `key={id}`), and it has a `[agentId]` reset effect (L132), so an in-place agent switch is a real, supported flow. If a reveal for agent A is in flight when the user switches to agent B (or leaves the env section), the resolved promise writes A's plaintext into the now-B component state via `setEnvOriginal`/`setEnvRevealed`. This defeats the clear-on-switch guarantee and can surface A's secrets under B — and a subsequent Save would persist A's values onto B.

Both analysis passes independently flagged this as the highest-severity issue. Fix: capture `agent.id` (or an incrementing request token / `envIdRef`-style guard) before the `await` and drop the result if it no longer matches the current agent; ideally also abort in-flight requests on `agentId`/`section` change.

### P1 — Potential silent secret deletion via the empty-state "Add variable" path

`AgentDetailPage.tsx` L382 (`const envKeyCount = agent.custom_env_key_count ?? 0`) + empty-state branch (~L601-607) + `saveEnv` (L284). The whole owner UX branches on the **new** `custom_env_key_count` field. If any response or cache path returns an agent object without that field (e.g. a stale cached agent, or a backend that has not yet been deployed with the count), `envKeyCount` falls back to `0`, and the owner is routed into the "No variables yet → Add variable" empty state instead of the masked reveal. That path calls `setEnvOriginal({})` and enters the editor **without ever GETting existing values**; a Save then PUTs a map built only from the newly-typed rows. If the backend PUT is full-replace (it returns the full `custom_env` map, which strongly suggests replace semantics), all pre-existing secrets are silently wiped.

The already-existing `has_custom_env: boolean` field is never consulted — it is dead in the whole repo. This is a cross-repo contract risk I could not close: `custom_env_key_count` does not exist anywhere in octo-server, and the dmloop backend that owns this field is in a repo not available to this review, so I cannot confirm the field is populated on every agent response, nor confirm PUT replace-vs-merge semantics.

Recommended before merge: (a) gate the empty-state "Add variable" branch on a signal that is guaranteed present (`has_custom_env === false`), or fetch existing env before entering the editor even when the count reads 0; and (b) confirm the backend actually populates `custom_env_key_count` on the agent GET/list responses this page consumes. If PUT is merge (not replace), the data-loss half is moot — please confirm.

### P1 — Clear-on-leave misses the top-tab switch

`AgentDetailPage.tsx` L154-161. The plaintext-clearing effect keys only on `section !== "env"`. But the top-level tab is separate state (`tab`, L90). Revealing secrets and then clicking the "Profile" top tab leaves `section === "env"`, so the effect does not fire and the plaintext stays in state; returning to "Config" restores the revealed editor. The advertised "cleared when leaving" property is not fully enforced. Fix: also clear when `tab !== "config"` (or clear on any navigation away from the env panel).

### P2 findings

- **Silent row drops / accidental delete** (`entriesToEnvMap` L96-103, `saveEnv` L280): rows with a blank (or whitespace-only) key are dropped silently, and duplicate validation only looks at trimmed non-empty keys. If a user clears an existing key but leaves its masked value, that variable is deleted on Save with no confirmation or validation. Consider warning on non-empty-value + empty-key rows rather than dropping them.
- **Edits during in-flight save are lost** (`saveEnv` L286): inputs and Add stay enabled while `envSaving` is true; `setEnvRevealed(envMapToEntries(env))` on success overwrites any edits made during the PUT. Consider disabling the editor while saving.
- **`refreshAgent` failure misreported as save failure** (`saveEnv` L287-292): the PUT and the follow-up `refreshAgent()` share one try/catch. If the PUT succeeds but the refresh fails, the user sees "save failed" (and a 403 on refresh even flips to forbidden + clears the editor), while the mutation actually landed — a retry re-applies an already-successful write. Separate the refresh from the save-result path.
- **403-save leaves plaintext in `envOriginal`** (`saveEnv` L291): on save-403 the code clears `envRevealed` but not `envOriginal`, so the plaintext map lingers in state until a later section/agent change. Clear `envOriginal` too for symmetry with the reveal path.

### Verified clean
- No localStorage/sessionStorage persistence of plaintext; value inputs use `mode="password"`.
- `LoopApiError.status` is correctly populated from `response.status` (`http.ts`), so the `=== 403` degrade checks are sound.
- i18n keys exist in both `zh-CN` and `en-US`; `{{count}}` interpolation matches the codebase's simple-replacement `t()` (no ICU plural — the `(s)` suffix convention is consistent with existing strings).
- API calls go through the shared `agentApi`/`http` client per repo convention; no rogue axios instance.
- CSS reuses the loop skin classes; no raw Semi defaults introduced.

## 3. Overall verdict

**CHANGES_REQUESTED**

Spec compliance is clean, but three P1 code-quality issues block merge: the cross-agent stale-write race (secret leakage/mis-save), the empty-state path that can silently delete existing secrets when the count field is absent, and the clear-on-leave gap on top-tab switch. The empty-state issue additionally hinges on an unverifiable cross-repo contract (`custom_env_key_count` population + PUT replace/merge semantics) — please confirm those explicitly.

## 4. Suggestions

1. Guard both async handlers with a captured `agentId`/request token; abort or discard on agent/section change.
2. Route the empty-state editor entry through a guaranteed-present signal (`has_custom_env`) and/or always fetch current env before editing; drop the now-dead `has_custom_env` or use it.
3. Extend the clear effect to fire on top-tab change, not just sub-section change.
4. Treat non-empty-value + empty-key rows as a validation error rather than a silent drop; disable the editor during save; isolate `refreshAgent` failures from the save result.

## 5. Additional notes (out of diff)

- Coverage gap: backend behavior for `custom_env_key_count`, `has_custom_env`, PUT replace-vs-merge, RBAC enforcement, audit logging, and encryption-at-rest could not be verified from this frontend-only checkout — they live in a backend repo not available here. The data-loss P1 above is contingent on PUT being full-replace; if it merges, that half downgrades to P2.

Jerry-Xin (approved):
Re-review at `754262b0` (previous CR was at `174342d8`). Both consensus blockers are resolved. **APPROVE.**

## ✅ BLOCKER A (secret-wipe) — FIXED
`AgentDetailPage.tsx:394` now gates on `const envHasVars = agent.has_custom_env === true || envKeyCount > 0;` (`has_custom_env?: boolean` confirmed on the `Agent` type at `types.ts:492`). The empty-state direct-edit branch (`setEnvOriginal({}); setEnvRevealed([...])`) is now only reachable in the `envHasVars === false` else-branch (`:616`). When env exists but the count is absent (`has_custom_env === true, count === 0`), the flow now lands in the masked "reveal-before-edit" branch (`:604`), forcing `handleRevealEnv → getAgentEnv → setEnvOriginal(realMap)` before any save can run. The old `custom_env_key_count ?? 0`-only gate is gone, so the silent full-map replace-PUT that dropped existing keys is no longer reachable. A `envMaskedUnknown` string is added for the count-absent case.

## 🟡 FINDING B (reveal/save async race) — correctly guarded (defense-in-depth)
A request-generation counter `envReqGenRef` is introduced (`:129`) and incremented on both `[agentId]` switch (`:135`) and env-panel leave (`:161`, now keyed on `tab === "config" && section === "env"`). Both `handleRevealEnv` (`:259`) and `saveEnv` (`:291`) capture `gen` before the await and bail (`if (envReqGenRef.current !== gen) return;`) afterward, so a late reveal cannot write plaintext back into memory and a late save cannot land on the wrong expert.

On reachability of a cross-expert leak: `SquadDetailPage.tsx:148` pushes `<AgentDetailPage agentId={id} …>` without a `key`, and the `[agentId]` effect does react to a prop swap, so an in-place agentId swap is theoretically possible via reconciliation. The new generation guard neutralizes that race regardless — a superior fix to only adding a `key`, since it also covers the leave-panel late-response case. I keep this at 🟡 defense-in-depth (now mitigated); I could not construct a reachable leak past the guard.

## ✅ Also good
Save-in-flight now disables the add/save buttons and every key/value/delete input (`:588,:640,:648,:650`), preventing concurrent edits from racing the in-flight PUT. The reveal/save `finally` blocks intentionally clear the busy boolean regardless of generation, avoiding a stuck "revealing…"/"saving…" state after leaving. i18n keys added in both zh-CN and en-US.

CI Build passing; the two red gate checks are the reviewer-approval gates, which this review clears.

yujiawei (approved):
# Code Review — PR #884 (octo-web) — re-review

Re-reviewed at head SHA `754262b0a7e5dc9a2c04a534a162d74cc57cb442`. My earlier review was against `174342d8` and requested changes for three blocking issues; the new commit *"harden expert env editor against stale-response and destructive saves"* addresses all three. This review supersedes the previous verdict for the current head.

## 1. Spec compliance

**Spec: ✅** — unchanged and still clean. Owner-only reveal-then-edit, masked-by-default, 403 read-only degrade, no plaintext persistence, frontend-only scope. No over-build, no missing requirement.

## 2. Code Quality

**Quality: Approved**

### Previously-blocking P1s — all resolved at this head

1. **Stale async write across expert switch / section leave** — resolved. A generation counter (`envReqGenRef`) is bumped on `agentId` change and on leaving the env panel; both `handleRevealEnv` and `saveEnv` capture `gen` before the `await` and discard the result when `envReqGenRef.current !== gen`. Late responses can no longer write one agent's plaintext into another agent's state or persist it. Loading-flag resets in `finally` are intentionally generation-independent, which correctly avoids a stuck "Revealing…" state.
2. **Silent secret deletion via the empty-state path** — resolved. `envHasVars = agent.has_custom_env === true || envKeyCount > 0` now gates the direct-edit empty state; when the new `custom_env_key_count` is absent, it falls back to the pre-existing stable `has_custom_env` field and routes to the reveal path (new `envMaskedUnknown` label) instead of entering a blank editor that would overwrite existing secrets on save. This is the intended fix.
3. **Clear-on-leave missed the top-tab switch** — resolved. The clear effect now depends on `[tab, section]` with `!(tab === "config" && section === "env")`, so switching to the Profile top tab also clears revealed plaintext and invalidates in-flight requests.

Also fixed: editor inputs, Add, and Delete are now `disabled={envSaving}`, closing the "edits lost during in-flight save" P2.

### Remaining P2s (non-blocking, optional follow-up)

- **Empty-key rows silently dropped on save** (`entriesToEnvMap`): a row with a value but no key is discarded with no inline warning or toast. Safe (no bad data reaches the backend) but a minor UX gap — consider an inline invalid-key indicator.
- **`refreshAgent` failure misreported as save failure** (`saveEnv`): the post-save `await refreshAgent()` shares the save try/catch, so a refresh error surfaces "save failed" even though the mutation landed.
- **`envOriginal` not cleared on save-403**: the plaintext map lingers in `envOriginal` after a 403 save until the next navigation (which now clears it via the generation effect), so the exposure window is small.

### Verified clean
- Generation-counter pattern is race-correct (single increment per agent switch when already on the env panel; extra mount-time increments are harmless).
- No localStorage/sessionStorage persistence; value inputs use `mode="password"`.
- `LoopApiError.status` populated from `response.status`, so the `=== 403` degrade checks are sound.
- i18n parity: `envMaskedUnknown` added to both `en-US` and `zh-CN`; both JSON files valid.
- Shared `agentApi`/`http` client used; CSS reuses the loop skin.

## 3. Overall verdict

**APPROVED**

All three P1 blockers from the prior review are fixed at this head. Spec is clean; only non-blocking P2 UX items remain.

## 4. Suggestions (optional)

1. Flag empty-key-but-non-empty-value rows as a validation error rather than dropping them silently.
2. Move `await refreshAgent()` out of the save try/catch so a refresh failure doesn't mask a successful save.
3. Clear `envOriginal` alongside `envRevealed` on save-403 for symmetry.

## 5. Additional notes (out of diff / coverage)

- The empty-state fix now leans on `has_custom_env` (shipped since #771, so reliably populated) rather than the brand-new `custom_env_key_count`; the residual contract dependency is therefore on a stable existing field. Backend PUT replace-vs-merge semantics, RBAC enforcement, audit logging, and encryption-at-rest still could not be verified from this frontend-only checkout.

lml2468 (approved):
## 🐙 PR #884 Re-Review — `feat(dmloop): add env variable config to expert config page`

**Head:** `5bea00943b34` (was `174342d8d335`) · **Base:** `main` (merge-base `6208748cdade`) · +258 −4 / 5 files
**Prior verdict:** REQUEST_CHANGES — one data-integrity P2 (empty-state edit could silently wipe existing secrets by trusting `custom_env_key_count` alone) + a 🟡 reveal async-race guard.

### 🔴 → ✅ The secret-wipe blocker is fixed — exactly the specified guard

Two new commits (`754262b0 harden … stale-response and destructive…`, `5bea00943 tidy save edge cases`). Delta since my CR head addresses **every** finding:

**① 🔴→✅ secret-wipe closed.** New derived signal:
```ts
const envHasVars = agent.has_custom_env === true || envKeyCount > 0;
```
The masked-vs-empty gate now branches on `envHasVars`, so:
- `has_custom_env:true` with **unknown/absent count** → falls into the **masked "reveal-before-edit"** path (new `envMaskedUnknown` copy), never the blind empty-edit.
- The direct empty-edit path (`setEnvOriginal({})` + blank row) is now reachable **only** when `!envHasVars` — i.e. `has_custom_env` false **and** count 0 → genuinely no secrets to overwrite.

Verified by trace: blind empty-edit is now unreachable whenever the backend signals env exists, **regardless of count reliability** — precisely the fix. The `has_custom_env` field the old code ignored is now the primary guard.

**② 🟡→✅ reveal/save async race guarded.** New `envReqGenRef` generation counter, incremented on agentId change (`:137`) and on leaving the env panel (`:162`). Both `handleRevealEnv` and `saveEnv` snapshot `gen` and **bail before any state write if it advanced** (`:265/269` reveal, `:297/304` save) — so a late reveal can't write stale plaintext, and a save can't apply to the wrong expert. Defense-in-depth, as recommended.

**③ 🟡→✅ tab-switch clears plaintext.** Clear-effect dep changed `[section]` → `[tab, section]` with guard `!(tab === "config" && section === "env")` — Config→Profile now clears + invalidates in-flight requests too.

**④ 🟡→✅ inputs disabled during save.** key/value/delete/add all `disabled={envSaving}`.

**⑤ Bonus (peer 🟡s):** empty-key-with-value row is now blocked pre-save (`envEmptyKey`) — closes mochashanyao's silent-drop note; `refreshAgent()` is best-effort (won't false-report save failure); 403-save clears `envOriginal` too.

### Verification (local checkout @ `5bea00943b34`, pnpm installed)
| Check | Result |
|---|---|
| `vitest run` (dmloop full) | ✅ 63/63 |
| new i18n keys (`envMaskedUnknown`, `envEmptyKey`) | ✅ both locales, symmetric |
| wipe path reachability | ✅ empty-edit only when `!has_custom_env && count===0` (traced) |
| every async env-state write behind gen recheck | ✅ reveal + save guarded; sync user actions have no async gap; effects bump gen |

secret-wipe closed via `has_custom_env||count>0` + reveal-before-edit ✅ · gen-guard on both async write-backs (no unguarded setEnvRevealed/setEnvOriginal path) ✅ · tab+section clear ✅ · inputs disabled on save ✅ · empty-key blocked ✅ · dup-key + 403 degradation + no-auto-fetch (all still intact) ✅ · build + dmloop 63/63 ✅.

### Non-blocking (carried)
Still no automated test for the reveal-gating / wipe-guard / gen-race flow — a security surface; a unit test locking "has_custom_env=true + count absent ⇒ no blind empty-edit" and "stale reveal discarded" would be the highest-value fast-follow. And the backend `/agents/:id/env` should enforce owner-only + audit as the UI copy claims (server-side, out of this PR).

---

### ✅ Review Verdict: APPROVE @ `5bea00943b34`
My REQUEST_CHANGES blocker (empty-state secret-wipe) is fixed at the root — the gate now honors `has_custom_env`, forcing reveal-before-edit whenever env may exist, so the wipe is unreachable independent of count reliability. The reveal/save async race is closed with a generation guard, and the peer 🟡s (tab-switch clear, inputs-during-save, empty-key) are all addressed. Build green, dmloop 63/63, i18n symmetric. This clears my CR. Non-blocking: add regression tests for the security-sensitive flow. Clean to merge.

yujiawei (approved):
# Code Review — PR #884 (octo-web)

Reviewed at head SHA `5bea00943b3464467a9b90ae4c26e070235a0e2a` (base `main`, merge-base `6208748cdade`; +258 −4 / 5 files). Frontend-only change adding an owner-only "Environment variables" section (reveal-then-edit, password-masked) to the dmloop expert config page. This is a re-review at the current head — the new commit `5bea0094` ("tidy expert env save edge cases") sits on top of the previously-reviewed `754262b0` and is a small, purely defensive follow-up.

## 1. Spec compliance

**Spec: ✅**

- **Under-build:** none. The feature — env-variable config on the expert page — is fully implemented: a new `env` config sub-section, reveal-then-edit for owners, masked display, add/edit/delete rows, count display for non-owners, and save through `updateAgentEnv`.
- **Over-build:** none. Changes are scoped to the env-config surface; no unrelated flags, fields, or endpoints introduced.
- **Deviation:** none. Implementation matches the stated intent (owner-only masked secrets editor).

## 2. Code Quality

**Quality: Approved**

Cross-file dependency check (the most important verification here): `AgentDetailPage.tsx` imports `getAgentEnv`/`updateAgentEnv` from `agentApi` and `LoopApiError` from `http`, none of which appear in this diff. All three were introduced by an earlier PR and are already present on `main` (verified they are ancestors of the merge-base), so the change compiles — no missing-symbol build break.

Static verification: every new icon/`Input` import is used; all 14 new i18n keys are wired in the component and present in both `en-US` and `zh-CN` with matching parity; both locale JSON files parse; the `t(key, { values: { count } })` shape matches existing usage across the package; `refreshAgent` and the reused `loop-adp__edit-input` styles exist.

Security / correctness trace (secrets surface):

- **Plaintext lifecycle** — `envRevealed` starts `null`; plaintext is fetched only on an explicit "Reveal & edit" click, and is cleared whenever the agent changes or the user leaves the env panel (`AgentDetailPage.tsx:151-165`, `AgentDetailPage.tsx:134-138`). Good — plaintext does not linger.
- **Async races** — `envReqGenRef` is bumped on every agent switch / panel exit, and both reveal and save discard their result when the generation has moved on (`AgentDetailPage.tsx:263,290,298`). This closes the "late response writes plaintext back after clear" and "cross-agent secret bleed" windows.
- **Destructive save guard** — `envHasVars = agent.has_custom_env === true || envKeyCount > 0` forces a reveal before editing whenever secrets may exist, so an empty editor can never blind-overwrite and silently delete existing secrets (`AgentDetailPage.tsx:399`). The count is treated as display-only, correctly leaning on the load-bearing `has_custom_env` flag when the count is unknown.
- **New in `5bea0094`** — (a) save is now blocked when a row has a value but an empty key, preventing a silent drop/delete of that entry; (b) the post-save `refreshAgent()` is now best-effort (`.catch()`), so a failed count-refresh no longer misreports a successful save as failed, and a 403 on refresh no longer flips the editor into the forbidden state; (c) a 403 during save now also clears the cached plaintext map, symmetric with the reveal path. All three are correct and strictly reduce risk.
- **403 downgrade** — both reveal and save degrade to read-only masked without an error toast, matching the backend owner-only gate.

No P0/P1 findings.

## 3. Overall verdict

**APPROVE** — Spec ✅ AND Quality Approved.

## 4. Suggestions (non-blocking, P2/nit)

- `envMaskedCount` uses `{{count}} variable(s) configured`; if the i18n layer supports pluralization you could split singular/plural, but the current `(s)` form is acceptable.
- Optional: a lightweight client-side key-format check (e.g. disallow whitespace / `=` in variable names) would surface bad env keys before the round-trip, though the backend remains the source of truth.

## 5. Additional observations

- The frontend adds `custom_env_key_count?` to the `Agent` type as display-only. If the backend does not populate it, the UI degrades gracefully to the "Environment variables configured" label rather than a number — no correctness impact.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #884 (octo-web)

## Summary

This PR adds an environment variable configuration panel to the expert (agent) detail page. The implementation follows a reveal-then-edit pattern: secrets stay masked until the owner explicitly clicks "Reveal & edit", which triggers a GET for plaintext values. On navigation away from the panel or agent switch, plaintext is cleared from React state. The feature includes race-condition protection via a generation counter (`envReqGenRef`), graceful 403 degradation to read-only masked view, duplicate-key and empty-key guards before save, and a stable-JSON dirty check that ignores row reorder.

Overall this is a well-structured, security-conscious implementation. One non-blocking UX issue and one nit below.

## Verification

- ✅ **Race condition protection** — `envReqGenRef` bumps on agent switch and panel exit; both `handleRevealEnv` and `saveEnv` check generation after `await` and discard stale responses. `finally` blocks correctly reset loading/saving booleans regardless of generation (`packages/dmloop/src/panel/AgentDetailPage.tsx:121,157,265,302`).
- ✅ **Plaintext lifecycle** — `envRevealed` is set to `null` on agent switch (`:136`), on panel exit (`:161`), and on 403 during save (`:307`). Secrets never persist in state beyond the editing session.
- ✅ **API boundary** — `getAgentEnv`/`updateAgentEnv` at `packages/dmloop/src/api/agentApi.ts:91-97` use the shared `httpGet`/`httpPut` client (base URL, auth, space header all handled by interceptor). `LoopApiError.status` is correctly typed at `packages/dmloop/src/api/http.ts:68`.
- ✅ **i18n parity** — All 14 new keys present in both `en-US.json` and `zh-CN.json` with matching key names.
- ✅ **Input validation** — `saveEnv` blocks duplicate keys (`:291`), empty key with non-empty value (`:293`), and `entriesToEnvMap` drops empty-key rows (`:56`). Dirty check uses sorted stable JSON (`:63`) so row reorder doesn't false-trigger the save button.
- ✅ **ASYNC/RETRY family** — All async paths have generation guards + `finally` cleanup. `refreshAgent().catch(() => {})` after save is correctly fire-and-forget (save already succeeded; metadata refresh is best-effort).
- ✅ **INPUT-SHAPE family** — `custom_env_key_count ?? 0` handles undefined (`:396`); `envHasVars` combines boolean and numeric guards (`:399`); `Input` components accept any string with no shape parsing at the component level.

## Findings

### P2 — 403 on save silently discards edits with misleading message (`packages/dmloop/src/panel/AgentDetailPage.tsx:307`)

When `updateAgentEnv` returns 403, the code sets `envForbidden=true`, clears `envRevealed`/`envOriginal`, and the UI transitions to the masked view showing `envRedacted` ("only the owner can view and edit them"). However, the user was just editing — they successfully revealed values, typed data, and clicked Save. The "only the owner" message contradicts their direct experience, and their typed values are silently discarded without any toast or confirmation.

This scenario is reachable when an agent-owner passes the front-end `canEdit` check (`agent.owner_id === myMemberId`) but the backend env-write permission model disagrees — the code comment at `:119` explicitly acknowledges this gap.

Adding a toast before clearing edit state would make the failure visible, e.g. `Toast.error(t("loop.agent.envSaveForbidden"))` with a new i18n key such as "Permission denied — your changes were not saved."

### Nit — Value whitespace preserved without documentation (`packages/dmloop/src/panel/AgentDetailPage.tsx:57`)

`entriesToEnvMap` trims keys but preserves values verbatim. This is almost certainly correct (secrets can have meaningful whitespace), but if any downstream consumer (runtime injection, shell export) strips values, there could be a silent mismatch. Not a defect — flagging for awareness only.

- **Non-owner masked view** — `(!canEdit || envForbidden)` correctly routes non-owners to the locked/masked view showing variable count without values. No "Reveal" button leaks to non-owners.
- **Empty-state direct edit** — Only fires when `envHasVars` is false (both `has_custom_env !== true` AND `envKeyCount === 0`). No risk of overwriting existing vars from a blank slate when metadata is accurate.
- **Subnav integration** — "env" button added to config subnav at `:487`; `ConfigSection` type updated at `:30`. Section routing at `:581` is clean.
- **CSS namespace** — All new classes prefixed with `loop-adp__env-`, no collision risk. Font and color variables consistent with existing styles.
- **Type addition** — `custom_env_key_count?: number` at `packages/dmloop/src/api/types.ts:493` is optional, non-breaking.

## Data Flow Trace

|---|---|---|
| `envRevealed` entries | `getAgentEnv` → `data.custom_env ?? {}` | Yes, via `envMapToEntries` → state. Cleared on nav/403. |
| `envOriginal` | Same `getAgentEnv` response | Yes, stored for dirty comparison. Overwritten on successful save. |
| `envKeyCount` | `agent.custom_env_key_count ?? 0` from `getAgent`/`refreshAgent` | Yes, drives masked count display. Updated via `refreshAgent()` after save. |
| `envHasVars` | `agent.has_custom_env \|\| envKeyCount > 0` | Yes, gates reveal-vs-empty-state. Correctly prevents empty-state overwrite when vars exist. |
| `envDirty` | `stableEnvJson(current) !== stableEnvJson(original)` | Yes, enables/disables Save button. Sort-stable so reorder doesn't false-trigger. |

- **C1 (双路径 parity)**: Clear — `addEnvEntry`/`removeEnvEntry` both operate on same `envRevealed` state with `prev ?? []` guards. Reveal/cleanup paths symmetric with generation bumps.
- **C2 (control-flow ordering)**: N/A — env functions called from single JSX bindings, no nested reuse.
- **C3 (授权边界)**: Clear — front-end `canEdit` gates UI; backend enforces independent owner-only check on GET/PUT. 403 gracefully degrades.
- **C4 (授权生命周期)**: N/A — no auth/container state changes.
- **C5 (build vs runtime)**: N/A — pure frontend UI, no build/config/packaging changes.
- **C6 (治理/策略)**: N/A — no docs/policy changes.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Security-conscious implementation with proper race-condition guards, plaintext lifecycle management, and input validation. One P2 on 403-save UX (silent edit discard with misleading message) is non-blocking.

PR #886, opened 2026-07-18
feat(skillmarket): refine filters and unify skill forms

LarsonLian:
## Summary
- Redesign Skill market header with search, category, tag filtering, sort controls, and refined card grid overflow handling.
- Unify Skill publish and edit modal layout around package, version info, and basic info sections.
- Remove visibility, English-name form field, and file-size display from the form UI while keeping English name in package metadata.
- Add create/edit version behavior: publish defaults to 0.1.0 before upload, edit current version is read-only, and reupload bumps the patch version and clears changelog.
- Send changelog on create and cover updated submit/validation flows in tests.

## Verification
- pnpm --filter @dmwork/skillmarket exec vitest run src/components/__tests__/NewSkillModal.test.tsx src/components/__tests__/EditSkillModal.test.tsx --reporter=verbose
- pnpm --filter @dmwork/skillmarket exec vitest run src/api/skillApiReal.test.ts src/api/skillApi.test.ts --reporter=verbose
- git diff --check
- Browser validation with local Chrome against the Vite dev server for publish initial/parsed, edit current, and edit reupload states.

## Notes
- pnpm i18n:check still reports the existing hardcoded Chinese baseline in unrelated/mock files; no new UI copy is hardcoded in the touched components.


lml2468 (changes requested):
## 🐙 PR #886 Review — `feat: refine skill market filters and grid`

**Head:** `9476289f30d9` · **Base:** `main` · **+15389 −7463 / 193 files · 70 commits (13 merges)** · `mergeable: CONFLICTING` / `merge: DIRTY`

This one can't be soundly reviewed or merged in its current shape. Two independent, hard blockers — both structural, before any line-by-line code-quality pass is meaningful.

### 🔴 BLOCKER 1 — merge-conflicted against main (cannot merge)
`mergeable: CONFLICTING / DIRTY`. A no-commit merge against `main` surfaces **extensive conflicts**, including semantically hard ones:
- **Content conflicts** across load-bearing/infra files: `nginx.conf.template`, `apps/web/package.json`, `apps/web/vite.config.ts`, `apps/web/src/Layout/index.tsx`, `.i18n/scan-config.json`, `packages/dmloop/src/panel/AgentDetailPage.tsx`, and the **entire `dmworkbase/src/Components/GlobalSearch/` subsystem** (index/dataSource/filterState/types/css + tests — 10+ files).
- **modify/delete conflicts** (the serious class): this branch **deletes** `packages/dmworkbase/src/bridge/globalChatSearch/*` (model.ts, useGlobalChatSearch.ts, and their tests) that **main has since modified**. Resolving these isn't mechanical — it requires a product decision on whether that subsystem should exist at all, because the branch diverged before main's rework landed.

A PR that doesn't cleanly merge is an automatic REQUEST_CHANGES until rebased. Given the modify/delete divergence, this needs a real rebase-and-reconcile, not a trivial one.

### 🔴 BLOCKER 2 (Gate 1 — spec compliance) — description vs diff are wildly mismatched
The description says *"Redesign Skill market header… multi-select tag filtering… refine Skill card grid"* — a focused skill-market UI change. The actual diff is:
- **193 files across 10+ packages** — `dmworkbase` (42), `dmworkskillmarket` (39), `docs` (24), `dmloop` (21), `dmworkmcp` (18), plus `apps/web`, login, personal…
- **Infrastructure**: a new nginx `/market/api/v1/` proxy block + `MARKET_API_URL` env, `.env.example`, `vite.config` proxy — deploy-surface changes.
- **70 commits, 13 merges**, with messages like *"consolidate fork divergence — MCP + Skill Market + docs"*, *"adopt standard API contract"*, *"authenticate package downloads"*, *"route icon upload through marketplace"*.

This is a **fork-consolidation mega-PR** (MCP marketplace + skill market + docs + infra) wearing a single-feature title. Reviewers cannot trust that the diff matches the stated intent — the load-bearing anti-pattern the spec gate exists to catch. A change this size that touches nginx, env, auth'd downloads, and an API-contract migration needs (a) an accurate description enumerating the real surface, and (b) ideally **decomposition** into reviewable PRs (infra/proxy; API-contract migration; skill-market UI; MCP-market; the globalChatSearch removal) so each can be verified and reverted independently.

### Why I'm not doing a line-by-line pass yet
With the branch unmergeable (incl. modify/delete divergence needing author decisions) and the true scope undisclosed, a 25k-line quality pass would review code that will materially change on rebase — low signal, high false-positive risk. The productive next step is structural: rebase + reconcile the conflicts, then re-scope/split with an accurate description. I verified the workspace at least installs cleanly (`pnpm install` ok), but did not attempt build/tests on the conflicted tree.

### What I did verify
live `mergeable: CONFLICTING` ✅ · conflict set enumerated via no-commit merge (content + modify/delete) ✅ · scope quantified (193 files/10+ pkgs/70 commits/13 merges) ✅ · infra/deploy surface identified (nginx/env/vite) ✅ · description-vs-diff mismatch confirmed ✅.

---

### ❌ Review Verdict: REQUEST_CHANGES @ `9476289f30d9`
Two structural blockers: (1) **merge-conflicted against main** — extensive content conflicts plus modify/delete divergence on `globalChatSearch` that needs author reconciliation, so it can't merge; (2) **Gate 1 fail** — a 193-file/10-package fork-consolidation (incl. nginx/env/API-contract/auth changes) presented under a "refine skill market filters" title, so the diff can't be trusted against intent. Please rebase onto `main` and reconcile the conflicts, then update the description to the true scope and — strongly recommended — split into independently reviewable PRs. I'll do a full two-gate code review once it's mergeable and scoped. No line-level findings yet by design (the tree will change on rebase).

Jerry-Xin (changes requested):
The skill-market filter/grid refinement itself is well-built, but this PR bundles a large amount of unrelated fork-consolidation churn that reverts routing/session behavior currently in `main`, and it is in a conflicting state. Requesting changes.

## 🔴 Blocking

- 🔴 **Browser Back/Forward regression — `packages/dmworkbase/src/Service/Route.tsx`.** This PR reverts the router to a version where both `handlePopState` (L8) and `handlePageShow` (L12) call `RouteManager.shared.push(...)`, and `push()` (L41) unconditionally calls `window.history.pushState()`. So pressing Back fires `popstate` → `push()` → a *new* history entry is pushed instead of rendering the existing location; forward history is discarded and users can get trapped on a route. Both `main` and the PR's own merge-base already have the correct implementation: `popstate`/`pageshow` call `renderCurrentPath()` (render without pushing), plus a `syncPath(path, mode)` helper guarded by `if (currentUrl === routePath) return`. `renderCurrentPath` appears in `main` but is entirely absent from this branch's `Route.tsx`. This is a functional regression, not a new change — it must not land.

- 🔴 **PR scope is not the stated feature.** The diff is 193 files (+15389 / −7463) and includes destructive reversions well beyond "skill market filters and grid": routing/session handling (`Route.tsx`, `SessionScope`, `normalizeRoutePath`), global chat search (`packages/dmworkbase/.../GlobalSearch*`, `useGlobalChatSearch.ts` — deleted here but present in `main` via the search refactor), document table behavior (`packages/docs/.../TableRowHeight*`), and dev docs. The branch is also 18 commits behind `main` and currently reports `CONFLICTING`. Please rebase onto current `main` and re-scope the PR to only the intended skill-market changes so these reverts drop out and the true delta is reviewable.

## 🟡 Non-blocking

- 🟡 **Mock `downloads` sort not implemented — `packages/dmworkskillmarket/src/api/skillApiMock.ts` (`applySkillQuery`, ~L74).** The mock `.sort()` only special-cases `"latest"`; both `"comprehensive"` and `"downloads"` fall through to `updatedAt`. "Most downloaded" is therefore misleading in mock/dev and has no behavioral sort test. Add a `downloads`-by-count branch and a test.

- 🟡 **Pagination observer churn — `packages/dmworkskillmarket/src/pages/SkillListPage.tsx` (~L59).** The `IntersectionObserver` effect depends on the whole `list` object, so it is torn down and rebuilt on every render. Depend on the stable `list.loadMore` (and `list.hasMore`) instead to avoid re-creating the observer each render.

## ✅ Highlights

- Filter/sort scope is genuinely clean: `useSkills` derives a stable `tagKey` for the deps array, aborts in-flight requests when filters change (no stale-result race on rapid filter switches), and issues a single paginated request per change (no N+1).
- All skill-authored content (name, tags, owner, description) is rendered as escaped JSX text / `title` attributes — no `dangerouslySetInnerHTML`, so no XSS surface in the cards.
- i18n is complete and symmetric: en-US and zh-CN both have 125 keys with full parity, and every new `filter.*` / `sort.*` / `list.publishSkill` key exists in both locales.
- New multi-tag and sort query params are covered by focused API tests.

yujiawei (changes requested):
# Code Review — PR #886 (octo-web)

Reviewed at head SHA `9476289f30d90844c51aa395fbc9bcbfb0115cd3` against merge-base `1d91ddd9` with `main`. This is a large change: 193 files, +15389/-7463, two new packages (`dmworkmcp`, `dmworkskillmarket`), a GlobalSearch refactor, and a bundled upstream sync. Dependencies were not installed in the review environment, so the test suites were read but not executed.

## Verdict: Changes requested

There are two blocking issues (merge state and a routing behavior regression) plus a scope-transparency concern. The skill-market feature code itself is generally solid; most of the remaining items are advisory.

---

## Blocking

### 1. The PR does not merge (`CONFLICTING` / `DIRTY`)
A trial merge into current `main` fails with real content conflicts, not just lockfile churn:

- `packages/dmworkbase/src/Components/GlobalSearch/types.ts` (content)
- `packages/dmworkbase/src/Components/GlobalSearch/filterState.ts`, `index.tsx`, `dataSource.ts` (content)
- `packages/docs/src/editor/styles.css` (content)
- `pnpm-lock.yaml` (content)
- modify/delete conflicts across the entire `packages/dmworkbase/src/bridge/globalChatSearch/**` and `features/globalChatSearch/**` tree (deleted here, modified on `main`)

Please rebase onto current `main` and resolve. The GlobalSearch and `globalChatSearch`-removal conflicts need care — `main` has continued to evolve those files while this branch deletes them.

### 2. Menu clicks no longer update the browser URL (routing regression vs `main`)
`apps/web/src/Pages/Main/index.tsx` removes both `WKApp.route.syncPath(menus.routePath)` calls (lines ~254 and ~316 on `main`), and `RouteManager.syncPath()` is gone entirely (it still exists and is called on `main`). The new `onMenuClick` handler sets `currentMenus`, runs `onPress`/`popToRoot`, and emits a mitt event, but never calls `RouteManager.push`, which is the only path that now writes `history.pushState`. `RouteManager.push` is only wired to `popstate`/`pageshow`.

Net effect after merge: clicking Contacts / Docs / MCP Market / etc. changes the rendered pane but leaves the address bar on the previous route. A refresh then restores the wrong page, and back/forward can't traverse menu history.

This appears to come in via the bundled upstream sync rather than the feature work. Please confirm whether dropping menu→URL sync is the intended new routing model. If it is intended, a follow-up should restore deep-link/refresh correctness for menu routes; if not, re-add the sync in `onMenuClick`.

### 3. Scope is much larger than the title, and includes load-bearing behavior changes
The title is "refine skill market filters and grid," and the described skill-market work (search / category / multi-select tag / sort / grid) is indeed implemented. But the diff also lands:

- an entire second product surface — the `dmworkmcp` MCP marketplace package (~1.3k-line create modal, service layer, probe flow), and
- an upstream fork-sync that removes session handling (`ensureSessionSid`, `stripSessionSidFromUrl`, `SessionScope`) and the URL-routing behavior in #2, and deletes the old global-chat-search stack.

A reviewer/merger reading the title cannot reason about that blast radius. Please either split the upstream sync (and the MCP marketplace, if separable) into their own PRs, or update the title/description to state plainly that this adopts an upstream routing/session rework and adds the MCP market. At minimum, call out the `ensureSessionSid`/`syncPath` removals explicitly so they aren't approved by accident.

---

## Advisory (non-blocking)

- **`packages/dmworkmcp/src/pages/McpMarketListPage.tsx:88` — stale-response race.** `loadData()` has no cancellation or request id. Type `a` then `ab`; if the `a` response lands last it overwrites the `ab` results. Same for category/space/mode switches. The sibling skill-market side already does this correctly (`useSkills.ts` uses an `AbortController`) — mirror that here (abort previous, or guard with a monotonic request id).
- **`packages/dmworkskillmarket/src/components/NewSkillModal.tsx` — cancellation is briefly un-asserted.** `reset()` sets `abortRef.current = true` then schedules `setTimeout(() => { abortRef.current = false }, 0)`. An in-flight upload/parse that resolves right after the modal closes will see `abortRef.current === false` and continue (state updates on a closed modal, parse polling keeps going). Only flip the ref back to `false` at the start of `startUpload`, not on a timer.
- **Parse-polling timeout is ~2× the documented limit.** `NewSkillModal` wraps `pollParse()` in its own `maxAttempts = 60` / 1s loop, but `pollParse()` (`skillApiReal.ts`) itself does 60 attempts × 2s. A permanently-pending parse blocks the first `pollParse` call ~118s before the outer loop can enforce its "60s max." Let one layer own the timeout.
- **`deleteSkill` may reject a valid `204 No Content`.** `requestEnvelope` throws `invalid_response` unless the body has a `data` key. The sibling `DELETE /mcps/{id}` on the same marketplace backend is documented as "Returns 204 No Content." The unit test masks this by returning `{ data: {} }` rather than an empty 204 body, so green tests don't prove the real contract. Please confirm the backend returns an envelope (not a bare 204) for skill delete, or special-case empty-body success.
- **No `AbortSignal` on `uploadFile` / `pollParse`.** Navigating away mid-upload or mid-parse leaves the XHR and the polling loop running. Consider threading a signal through (the list/search paths already do).
- **`useSkills.fetchPage` refetches categories on every `loadMore`.** Categories are static per view; `Promise.all([getCategories(), getSkills(...)])` re-pulls and re-sets them on each page. Minor, but it re-renders the category chips as the user scrolls.
- **401 does a hard `window.location.href` redirect** (`skillApiReal.ts`), which drops SPA state. An app-level auth event/callback would be gentler. Minor.
- **`package.json` nits in the new packages.** `dmworkskillmarket` declares `react: ^17.0.2` as a devDependency alongside `@types/react: ^18.3.28`; `dmworkmcp` also pins `react: ^17.0.2`. The app runs React 18 — align these to avoid types/runtime skew. Also, `apps/web/package.json` adds `mathlive` in this PR, which looks unrelated to the market work; confirm it's intended.

---

## What looks good

- User-supplied skill READMEs render through `react-markdown` with `rehype-sanitize` (no `dangerouslySetInnerHTML` / `eval` anywhere in the new packages).
- Skill package download goes through an authenticated `download?format=json` → `download_url` indirection rather than a raw unauthenticated URL.
- The list/search hooks (`useSkills`, `SearchBar`) use `AbortController` correctly and debounce input.
- The MCP probe flow handles secrets carefully: the probe bearer is ephemeral and never persisted, presigned PUTs go via a raw axios instance so app auth headers don't leak to third-party storage, and a secret sentinel round-trips redacted values.
- Removed exports (`hasGlobalSearchCriteria`, `selectedGlobalSearchFilterValueCount`) have no remaining consumers; i18n keys sampled in the new pages resolve.

## Coverage note

Not verified: test execution (no `node_modules` in the review env), the marketplace backend contracts (delete/empty-body, pagination cursor shape, tag query semantics), production nginx/proxy behavior, and the full 1.3k-line `McpCreateModal` (sampled, not read exhaustively). The routing regression (#2) was traced through the new `RouteManager` but not confirmed at runtime.

lml2468 (changes requested):
## 🐙 PR #886 Re-Review — `feat: refine skill market filters and grid`

**Head:** `8597414fe6c2` (was `9476289f30d9`) · **Base:** `main` · **+15430 −7463 / 193 files · 71 commits · behind_by 18** · `mergeable: CONFLICTING` / `DIRTY`

**Both structural blockers from my prior REQUEST_CHANGES are unchanged.** The new commit added *more* skill-market feature work (publish/edit modal unification, version behavior) rather than addressing either blocker — which broadens the PR rather than decomposing it.

### 🔴 BLOCKER 1 (unchanged) — still merge-conflicted against main
Live `mergeable: CONFLICTING / DIRTY`; a no-commit merge still reports conflicts (nginx.conf.template, package.json, vite.config, Layout, GlobalSearch subsystem, plus the `globalChatSearch` modify/delete divergence). `behind_by: 18` — the branch is still stale, not rebased. Cannot merge.

### 🔴 BLOCKER 2 (unchanged) — scope/description still mismatched
Still **193 files across 10+ packages** (dmworkskillmarket 42, dmworkbase 42, docs 41, dmloop 22, dmworkmcp 20, apps/web 14, nginx.conf.template…). The description was updated — but only to narrate *more skill-market feature detail*; it still does **not** disclose or account for the ~180 off-topic files (MCP market, nginx/env infra, `GlobalSearch` deletion, docs, the routing regression below). It remains a fork-consolidation mega-PR under a feature title. Reviewers still can't trust the diff against intent.

### 🔴 Concrete code-level regression carried in the churn (now cross-confirmed, still present)
`packages/dmworkbase/src/Service/Route.tsx` — the browser Back/Forward fix is **still clobbered** at this head: `renderCurrentPath` appears **3× in main, 0× in the PR head** (byte-verified `git show main: vs HEAD:`). Because the branch is 18 commits behind, it reintroduces the pre-fix `popstate`/`pageshow` → unconditional `pushState` behavior, re-trapping browser back navigation. This is the classic stale-branch symptom — the branch overwrites main's shipped fix — and it's exactly what BLOCKER 1's rebase must reconcile (don't overwrite main's Route.tsx / GlobalSearch). Independently byte-verified by a peer as well.

### The skill-market feature itself is not the problem
The ~13-file dmworkskillmarket core (SearchBar/SkillListPage/useSkills/SkillCard/skillApi/modals) is reasonable work — stable-dep query keys, AbortController on fetch, no XSS in cards, i18n parity, tested. It's the **packaging** (unmergeable stale fork-consolidation of 5+ sub-PRs + infra, mislabeled) that blocks it, not the feature.

### Why still no line-by-line pass
Unchanged from last round: on a conflicted, 18-behind, mislabeled tree that must be rebased (reconciling Route.tsx/GlobalSearch by hand) and ideally split, a 25k-line quality pass reviews code that will materially change — low signal. The productive next step remains structural.

---

### ❌ Review Verdict: REQUEST_CHANGES @ `8597414fe6c2` (verdict unchanged)
Same two structural blockers as my prior review, both still present: (1) **merge-conflicted / 18 behind main** — incl. the `globalChatSearch` modify/delete divergence, so it can't merge; (2) **Gate 1 fail** — a 193-file/10-package fork-consolidation (nginx/env/API-contract/auth/GlobalSearch-removal) still under a "refine skill market filters" title, description updated but still not covering the real surface. Plus the byte-confirmed **Route.tsx Back/Forward regression** the stale branch reintroduces. Please **rebase onto main** (preserving main's Route.tsx + GlobalSearch), **update the description to the true scope**, and **split into independently reviewable PRs** (skill-market UI · MCP-market · API-contract migration · infra/proxy · globalChatSearch removal). I'll do the full two-gate code review once it's mergeable and scoped — the skill-market feature looks landable on its own.

Jerry-Xin (dismissed):
The skill-market work is relevant, but the branch introduces severe unrelated regressions in authentication, routing, and document compatibility.

## 🔴 Blocking

- 🔴 Critical — Session credentials are lost after ordinary navigation. [search.ts](/Users/mlamp/.openclaw/tmp/pr-review-tv3nyx4h/repo/packages/dmworkbase/src/Utils/search.ts:8) generates a new `sid` without persisting it, while [App.tsx](/Users/mlamp/.openclaw/tmp/pr-review-tv3nyx4h/repo/packages/dmworkbase/src/App.tsx:545) keys credentials using only the URL `sid`. A user signing in on a clean URL stores credentials in the empty bucket; the next `route.push()` adds a random `?sid=...`, and refreshing then reads a nonexistent credential bucket and logs the user out. Restore a stable per-tab session scope or migrate credentials before changing the URL.

- 🔴 Critical — Back/Forward navigation creates new history entries. Both `popstate` and `pageshow` call `push()` in [Route.tsx](/Users/mlamp/.openclaw/tmp/pr-review-tv3nyx4h/repo/packages/dmworkbase/src/Service/Route.tsx:8), and `push()` unconditionally calls `history.pushState()` at line 48. Pressing Back therefore inserts another entry and can trap users in the history stack. History events must render the existing location without pushing.

- 🔴 Critical — The PR downgrades the document schema from v19 to v18 in [schema/index.ts](/Users/mlamp/.openclaw/tmp/pr-review-tv3nyx4h/repo/packages/docs/src/schema/index.ts:68) and removes the `tableRow.height` extension in [extensions.ts](/Users/mlamp/.openclaw/tmp/pr-review-tv3nyx4h/repo/packages/docs/src/editor/extensions.ts:240). Existing v19 documents can lose row-height attributes or fail schema-version compatibility checks. This unrelated rollback, along with its implementation and tests, must be removed from this PR.

## 💬 Non-blocking

- 🟡 Warning — The PR contains 193 changed files and removes several unrelated features and tests. Please rebase onto current `main` and restrict the branch to the marketplace work; the current scope makes regression verification impractical.

- 🟡 Warning — The documented marketplace tests could not run because `vitest` is unavailable in the checked-out workspace. CI should execute the listed suites before merge.

## ✅ Highlights

- The marketplace module is clearly separated and includes focused API and modal tests.
- Publish/edit version and changelog behavior is explicitly represented in the form logic.

Jerry-Xin (changes requested):
The skill-market work is relevant, but the branch introduces severe unrelated regressions in authentication, routing, and document compatibility.

## 🔴 Blocking

- 🔴 Critical — Session credentials are lost after ordinary navigation. [search.ts](packages/dmworkbase/src/Utils/search.ts:8) generates a new `sid` without persisting it, while [App.tsx](packages/dmworkbase/src/App.tsx:545) keys credentials using only the URL `sid`. A user signing in on a clean URL stores credentials in the empty bucket; the next `route.push()` adds a random `?sid=...`, and refreshing then reads a nonexistent credential bucket and logs the user out. Restore a stable per-tab session scope or migrate credentials before changing the URL.

- 🔴 Critical — Back/Forward navigation creates new history entries. Both `popstate` and `pageshow` call `push()` in [Route.tsx](packages/dmworkbase/src/Service/Route.tsx:8), and `push()` unconditionally calls `history.pushState()` at line 48. Pressing Back therefore inserts another entry and can trap users in the history stack. History events must render the existing location without pushing.

- 🔴 Critical — The PR downgrades the document schema from v19 to v18 in [schema/index.ts](packages/docs/src/schema/index.ts:68) and removes the `tableRow.height` extension in [extensions.ts](packages/docs/src/editor/extensions.ts:240). Existing v19 documents can lose row-height attributes or fail schema-version compatibility checks. This unrelated rollback, along with its implementation and tests, must be removed from this PR.

## 💬 Non-blocking

- 🟡 Warning — The PR contains 193 changed files and removes several unrelated features and tests. Please rebase onto current `main` and restrict the branch to the marketplace work; the current scope makes regression verification impractical.

- 🟡 Warning — The documented marketplace tests could not run because `vitest` is unavailable in the checked-out workspace. CI should execute the listed suites before merge.

## ✅ Highlights

- The marketplace module is clearly separated and includes focused API and modal tests.
- Publish/edit version and changelog behavior is explicitly represented in the form logic.

yujiawei (changes requested):
# Code Review — PR #886 (octo-web)

Reviewed at head SHA `8597414fe6c223a5f5c94a665c933a7170cf1116` against `main` (merge-base `1d91ddd9`). Focused on the Skill Market changes that this PR describes; the branch also carries a large fork-sync/consolidation (MCP, GlobalSearch, docs) that arrived via merge commits and is out of scope of the stated title — see the note at the end.

Verification performed locally on this SHA: `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm i18n:check`, and the `@dmwork/skillmarket` Vitest suite.

## 1. Spec compliance

**Spec: ✅** — the diff implements the described feature set and does not appear to add anything outside it:

- Skill market header refactor: search + tag filter + sort controls + card grid overflow (`SearchBar.tsx`, `SkillCard.tsx`, `pages/SkillListPage.tsx`, `hooks/useSkills.ts`, `api/skillApiReal.ts` `getSkillTags`/`sort`/`tags` params).
- Unified publish/edit modal layout (package → version → basic info sections) in `NewSkillModal.tsx` and `EditSkillModal.tsx`.
- Visibility control, English-name form field, and file-size display removed from the form UI; English name still flows through package metadata (`name` autofilled from parse result); file size still shown on the detail modal only.
- Version behavior: publish defaults to `0.1.0` before upload (`DEFAULT_CREATE_VERSION`), edit current version is read-only (`readOnly={!parseTaskId}`), reupload bumps the patch and clears changelog (`bumpPatch()` + `setChangelog("")`).
- `changelog` sent on create (`createSkill` body).

No over-build or contradiction with the stated scope was found.

## 2. Code quality

### P1 — `pnpm i18n:check` regresses a required CI gate

`pnpm i18n:check` is wired into the CI Build job (`.github/workflows/ci.yml:90`). It **passes on `main`** ("0 candidates within baseline") but **fails on this head** with `163 new hardcoded Chinese signatures found` (report: `.i18n/reports/hardcoded-strings.json`, 178 candidates across 5 files). The PR notes state this is "the existing hardcoded Chinese baseline in unrelated/mock files" — that is not accurate: these files are net-new to `octo-web main`, and the check is green on `main`, so this is a regression introduced by this branch, not a pre-existing baseline.

Two distinct sub-cases:

1. **Genuine user-facing strings in production code that should use `t()`** (not fixtures):
   - `packages/dmworkskillmarket/src/api/skillApiReal.ts:115` `"登录已过期，请重新登录"`, `:124` `"文件过大，请压缩后重试"`, `:364` `"下载地址无效"`, `:526`/`:532` `"解析失败"` / `"解析超时，请重试"` — these are surfaced to users via the error banners in the modals.
   - `packages/dmworkskillmarket/src/utils/installPrompt.ts:8` — the install-prompt body copied to the user's clipboard.
   - `packages/dmworkmcp/src/pages/McpMarketListPage.tsx:244` `aria-label="MCP 市场导航"`.
2. **Mock/fixture data** (`packages/dmworkskillmarket/src/api/mockData.ts`, ~169 of the signatures). This mirrors already-ignored fixtures such as `packages/dmworkmcp/src/mock/mcpMock.ts` and `quickStartTemplates.ts` in `.i18n/scan-config.json`. If this is intended to stay as demo data, add a matching `ignoredFiles` entry with a reason; otherwise it keeps the gate red.

Either the strings are localized or the fixture files are ignore-listed, but the check must be green before merge.

### P1 — `CategoryChips.test.tsx`: 4 tests are shipped broken and can never pass

`packages/dmworkskillmarket/src/components/__tests__/CategoryChips.test.tsx` (net-new in this PR) queries by resolved Chinese copy — `getByLabelText("Skill 分类")` and `getByRole("button", { name: /更多/ })`. But the test harness mock (`src/__mocks__/dmworkBase.tsx`) defines `t()` to return the **raw key**, and every other suite in this package correctly queries by raw key (e.g. `SkillListPage.test.tsx`, `SkillCard.test.tsx` use `"skillMarket.card.editAriaLabel"`). Result: all 4 CategoryChips tests fail (`Unable to find … 更多` / `Skill 分类`). Confirmed locally: `Tests 4 failed | 77 passed`.

CI does not currently run Vitest, so this does not block the merge gate, but it ships a red test file and contradicts the PR's "cover updated submit/validation flows in tests" claim. Fix the queries to match the raw-key convention (`t("skillMarket.common.more")`, `t("skillMarket.category.ariaLabel")`) or otherwise make the suite green.

### P2 — minor observations

- `NewSkillModal.tsx:330` sends `visibility: "space"` as a hard-coded literal now that the visibility UI is gone. That is consistent with the "remove visibility control" goal, but it means the field is dead-but-required on the create form type; consider dropping `visibility` from `NewSkillForm` / the create payload if the backend defaults it, to avoid a silently-fixed value.
- `EditSkillModal.tsx:20` `bumpPatch()` returns the version unchanged when it has fewer than 3 dotted segments and resets a non-numeric patch to `1`. Edge inputs (e.g. `"1.0"`, `"1.0.x"`) won't bump as a user might expect; acceptable given versions originate from `0.1.0`, but worth a guard or a comment.

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ✅, but Quality is Changes-Requested on the two P1 items above (i18n gate regression; broken committed tests).

## 4. Suggested direction

1. Localize the user-facing strings in `skillApiReal.ts` / `installPrompt.ts` / `McpMarketListPage.tsx` via `t()`, and add an `.i18n/scan-config.json` ignore entry for `mockData.ts` if it stays as demo data — then confirm `pnpm i18n:check` is green.
2. Repoint the `CategoryChips.test.tsx` assertions to raw i18n keys so the suite passes.

## 5. Additional note (not blocking on its own)

The branch's diff vs `main` is 193 files (+15,430/−7,463) spanning MCP market, GlobalSearch, and docs — well beyond the "skill market filters and grid" title. This content arrives through fork-sync merge commits, so it is understandable for a consolidated fork branch, but it makes the change hard to review as a unit and couples unrelated subsystems into one merge. Consider splitting the fork-sync from the skill-market feature in future PRs so each is independently reviewable and revertible.

This PR has already gone through several review rounds; if the P1 items recur after the next push, it may be worth a holistic pass rather than incremental round-tripping.

LarsonLian:
Closing this PR because the changes were pushed to LarsonLian/octo-web instead: https://github.com/LarsonLian/octo-web/pull/1

PR #887, opened 2026-07-18
fix(docs): persist sheet auto row-height so text-wrap survives reload (WS-28)

pkuWMH:
### 现象
在电子表格（Sheet）里给单元格开「自动换行」后，UI 立即正确（行自动撑高、文本折行），但离开文档再回来，行高恢复默认、把折行文本裁掉。

### 根因
换行状态本身（Univer cell style 的 `tb`）走的是通用样式同步通道，已经能持久化。真正丢的是**换行触发的行高自动增长**：Univer 用 `SetWorksheetRowAutoHeightMutation`（`sheet.mutation.set-worksheet-row-auto-height`）应用这次撑高，它跟手动拖拽的 `set-worksheet-row-height` 是两条不同的 mutation，行高以 per-row 的 `rowsAutoHeightInfo` 携带。协同 binding 的 `DIM_TRIGGER_IDS` 只监听了手动 resize，没监听 auto-height，所以撑高只改了本地模型（UI 看着对），从没进 Y.Doc → reload 从 Y.Doc 重建时行高回落。

### 改动
`packages/docs/src/sheet/binding.ts`：新增对 `sheet.mutation.set-worksheet-row-auto-height` 的监听，把重算出的行高写进**与手动 resize 完全相同的 dim map、相同 key 形状**（`${logical}:r<idx>`），reload 时沿用现有 `applyRemoteDims` 恢复。改动最小、边界清晰，不动其它同步路径。

- 前端 only，不涉及 doc schema / 后端（Sheet 走 content-agnostic 的 Yjs + Hocuspocus）。
- 无 migration：存量文档缺该 key 时行为与今天一致（内容撑开）。

### 验证
- `packages/docs` 单测全绿（42 passed，含新增 auto-height 持久化用例）。
- `pnpm typecheck` 改动文件零新增报错（仓库既有的两处无关报错不在本改动文件）。

Fixes WS-28

### DIFF
--- a/packages/docs/src/sheet/binding.test.ts
+++ b/packages/docs/src/sheet/binding.test.ts
@@ -464,6 +464,24 @@ describe('UniverYjsBinding — column/row dimensions', () => {
     expect(dimMap.get('default:r5')).toBe(30)
   })
 
+  it('persists an auto-recomputed row height (e.g. from toggling text-wrap) — WS-28', () => {
+    // Toggling 自动换行 grows the row to fit; Univer emits set-worksheet-row-auto-height (NOT the
+    // manual set-worksheet-row-height), carrying per-row heights in rowsAutoHeightInfo. Without
+    // syncing it the grown height never reaches the Y.Doc and the row snaps back on reload.
+    const { univer, dimMap } = setup()
+    univer.fire({
+      id: 'sheet.mutation.set-worksheet-row-auto-height',
+      params: {
+        rowsAutoHeightInfo: [
+          { row: 0, autoHeight: 48 },
+          { row: 2, autoHeight: 66 },
+        ],
+      },
+    })
+    expect(dimMap.get('default:r0')).toBe(48)
+    expect(dimMap.get('default:r2')).toBe(66)
+  })
+
   it('applies a remote column width into the sheet', () => {
     const { univer, doc } = setup()
     applyRemote(doc, (peer) => peer.getMap(SHEET_DIMS_FIELD).set('default:c1', 88))
--- a/packages/docs/src/sheet/binding.test.ts
+++ b/packages/docs/src/sheet/binding.test.ts
@@ -466,12 +466,14 @@ describe('UniverYjsBinding — column/row dimensions', () => {
 
   it('persists an auto-recomputed row height (e.g. from toggling text-wrap) — WS-28', () => {
     // Toggling 自动换行 grows the row to fit; Univer emits set-worksheet-row-auto-height (NOT the
-    // manual set-worksheet-row-height), carrying per-row heights in rowsAutoHeightInfo. Without
-    // syncing it the grown height never reaches the Y.Doc and the row snaps back on reload.
+    // manual set-worksheet-row-height), carrying per-row heights in rowsAutoHeightInfo keyed by the
+    // mutation's own subUnitId. Without syncing it the grown height never reaches the Y.Doc and the
+    // row snaps back on reload.
     const { univer, dimMap } = setup()
     univer.fire({
       id: 'sheet.mutation.set-worksheet-row-auto-height',
       params: {
+        subUnitId: 'local-1', // the first sheet, logical 'default'
         rowsAutoHeightInfo: [
           { row: 0, autoHeight: 48 },
           { row: 2, autoHeight: 66 },
@@ -480,6 +482,34 @@ describe('UniverYjsBinding — column/row dimensions', () => {
     })
     expect(dimMap.get('default:r0')).toBe(48)
     expect(dimMap.get('default:r2')).toBe(66)
+  })
+
+  it('routes an auto-height mutation to its OWN subUnitId sheet, not the active sheet — WS-28', () => {
+    // An auto-height recompute can fire against a non-focused sheet. The mutation's subUnitId is the
+    // authoritative target; using the active sheet would persist the height under the wrong sheet's
+    // key (right sheet stays clipped on reload, wrong sheet gets over-tall).
+    const { univer, dimMap } = setup()
+    const s2 = univer.addSheet('Sheet2')
+    univer.setActive(s2.getSheetId())
+    univer.fire({ id: 'sheet.command.insert-sheet' }) // register Sheet2 into the logical map
+    univer.setActive('local-1') // active sheet is now the FIRST sheet ('default')
+    // Auto-height targets Sheet2 while 'default' is active.
+    univer.fire({
+      id: 'sheet.mutation.set-worksheet-row-auto-height',
+      params: { subUnitId: s2.getSheetId(), rowsAutoHeightInfo: [{ row: 1, autoHeight: 52 }] },
+    })
+    expect(dimMap.get(`${s2.getSheetId()}:r1`)).toBe(52) // written under Sheet2's logical id
+    expect(dimMap.has('default:r1')).toBe(false) // NOT under the active sheet
+  })
+
+  it('ignores an auto-height mutation whose subUnitId maps to no known sheet — WS-28', () => {
+    const { univer, dimMap } = setup()
+    univer.fire({
+      id: 'sheet.mutation.set-worksheet-row-auto-height',
+      params: { subUnitId: 'ghost-sheet', rowsAutoHeightInfo: [{ row: 0, autoHeight: 48 }] },
+    })
+    expect(dimMap.has('default:r0')).toBe(false)
+    expect(dimMap.has('ghost-sheet:r0')).toBe(false)
   })
 
   it('applies a remote column width into the sheet', () => {
--- a/packages/docs/src/sheet/binding.ts
+++ b/packages/docs/src/sheet/binding.ts
@@ -824,22 +824,25 @@ export class UniverYjsBinding {
   /** Persist Univer's auto-recomputed row heights into the SAME dim map + key shape as a manual
    * row-height resize, so reload restores them via applyRemoteDims. Params differ from the manual
    * set-worksheet-row-height: heights arrive per-row in `rowsAutoHeightInfo` ({ row, autoHeight })
-   * rather than one `rowHeight` spanning `ranges`. Keyed by the ACTIVE sheet's logical id, matching
-   * syncDimFromCommand (an interactive wrap toggle acts on the active sheet). */
+   * rather than one `rowHeight` spanning `ranges`. Keyed off the mutation's OWN `subUnitId` (the
+   * worksheet the recompute actually targeted) — NOT the active sheet, since an auto-height pass can
+   * fire against a non-focused sheet and must not write its heights under the active sheet's keys.
+   * Bounds mirror the restore path (applyRemoteDims): row within grid, height finite/positive/≤cap. */
   private syncAutoRowHeightFromCommand(command: { id: string; params?: unknown }): void {
-    const p = command.params as { rowsAutoHeightInfo?: Array<{ row?: number; autoHeight?: number }> } | undefined
+    const p = command.params as
+      | { subUnitId?: string; rowsAutoHeightInfo?: Array<{ row?: number; autoHeight?: number }> }
+      | undefined
     const infos = p?.rowsAutoHeightInfo
     if (!infos || infos.length === 0) return
-    const wb = this.workbook()
-    const sheet = wb?.getActiveSheet()
-    if (!wb || !sheet) return
-    const logical = this.localToLogical.get(sheet.getSheetId())
+    if (typeof p?.subUnitId !== 'string') return
+    const logical = this.localToLogical.get(p.subUnitId)
     if (!logical) return
     const entries: Array<[string, number]> = []
     for (const info of infos) {
       const row = info?.row
       const h = info?.autoHeight
-      if (!Number.isInteger(row) || typeof h !== 'number' || !Number.isFinite(h) || h <= 0) continue
+      if (!Number.isInteger(row) || (row as number) < 0 || (row as number) >= SHEET_MAX_ROWS) continue
+      if (typeof h !== 'number' || !Number.isFinite(h) || h <= 0 || h > SHEET_MAX_DIM_PX) continue
       entries.push([`${logical}:r${row}`, h])
     }
     if (entries.length === 0) return
--- a/packages/docs/src/sheet/binding.ts
+++ b/packages/docs/src/sheet/binding.ts
@@ -46,10 +46,20 @@ const TRIGGER_IDS = new Set<string>([
   'sheet.command.set-border-style',
 ])
 
-/** Column-width / row-height mutations. */
+/** Column-width / row-height mutations (a manual resize drag). */
 const DIM_TRIGGER_IDS = new Set<string>([
   'sheet.mutation.set-worksheet-col-width',
   'sheet.mutation.set-worksheet-row-height',
+])
+
+/** Row auto-height recompute. Univer fires THIS mutation (not set-worksheet-row-height) when a
+ * layout-affecting cell style — text-wrap (`tb`), font family/size (`ff`/`fs`), rotation (`tr`) —
+ * grows a row to fit its content. Heights arrive per-row in `rowsAutoHeightInfo`, a different param
+ * shape than the manual resize above. Without syncing it, toggling 自动换行 updates the local model
+ * (row grows, UI looks right) but the new height never lands in the Y.Doc, so on reload the row
+ * snaps back to the default height and clips the wrapped text (WS-28). */
+const AUTO_ROW_HEIGHT_TRIGGER_IDS = new Set<string>([
+  'sheet.mutation.set-worksheet-row-auto-height',
 ])
 
 /** Add/remove merge mutations. */
@@ -314,6 +324,7 @@ export class UniverYjsBinding {
         if (!this.canWrite()) return
         if (TRIGGER_IDS.has(command.id)) this.syncLocalToYmap()
         else if (DIM_TRIGGER_IDS.has(command.id)) this.syncDimFromCommand(command)
+        else if (AUTO_ROW_HEIGHT_TRIGGER_IDS.has(command.id)) this.syncAutoRowHeightFromCommand(command)
         else if (MERGE_TRIGGER_IDS.has(command.id)) this.syncMergesToYmap()
         else if (DRAWING_TRIGGER_IDS.has(command.id)) this.syncDrawingsToYmap()
         else if (SHEET_LIFECYCLE_IDS.has(command.id)) {
@@ -800,6 +811,36 @@ export class UniverYjsBinding {
     for (const rg of p.ranges) {
       if (isCol) for (let c = rg.startColumn; c <= rg.endColumn; c++) entries.push([`${logical}:c${c}`, size])
       else for (let r = rg.startRow; r <= rg.endRow; r++) entries.push([`${logical}:r${r}`, size])
+    }
+    if (entries.length === 0) return
+    this.dimMap.doc?.transact(() => {
+      for (const [k, v] of entries) {
+        this.lastSeenDims.set(k, v)
+        this.dimMap.set(k, v)
+      }
+    })
+  }
+
+  /** Persist Univer's auto-recomputed row heights into the SAME dim map + key shape as a manual
+   * row-height resize, so reload restores them via applyRemoteDims. Params differ from the manual
+   * set-worksheet-row-height: heights arrive per-row in `rowsAutoHeightInfo` ({ row, autoHeight })
+   * rather than one `rowHeight` spanning `ranges`. Keyed by the ACTIVE sheet's logical id, matching
+   * syncDimFromCommand (an interactive wrap toggle acts on the active sheet). */
+  private syncAutoRowHeightFromCommand(command: { id: string; params?: unknown }): void {
+    const p = command.params as { rowsAutoHeightInfo?: Array<{ row?: number; autoHeight?: number }> } | undefined
+    const infos = p?.rowsAutoHeightInfo
+    if (!infos || infos.length === 0) return
+    const wb = this.workbook()
+    const sheet = wb?.getActiveSheet()
+    if (!wb || !sheet) return
+    const logical = this.localToLogical.get(sheet.getSheetId())
+    if (!logical) return
+    const entries: Array<[string, number]> = []
+    for (const info of infos) {
+      const row = info?.row
+      const h = info?.autoHeight
+      if (!Number.isInteger(row) || typeof h !== 'number' || !Number.isFinite(h) || h <= 0) continue
+      entries.push([`${logical}:r${row}`, h])
     }
     if (entries.length === 0) return
     this.dimMap.doc?.transact(() => {


lml2468 (approved):
## 🐙 PR #887 Review — `fix(docs): persist sheet auto row-height so text-wrap survives reload (WS-28)`

**Head:** `8b89194318ed` · **Base:** `main` (merge-base `6220f56c69cd`) · +60 −1 / 2 files · `mergeable: MERGEABLE`
**Intent:** Fix WS-28 — toggling a cell's 自动换行 grows the row locally (UI correct) but the grown height never reaches the Y.Doc, so on reload the row snaps back and clips wrapped text. Root cause: Univer applies the growth via `sheet.mutation.set-worksheet-row-auto-height` (per-row `rowsAutoHeightInfo`), a *different* mutation than the manual `set-worksheet-row-height` the collab binding already syncs. Fix: listen for the auto-height mutation and persist into the **same dim map + key shape** as manual resize. Frontend-only, no schema/backend, no migration.

### Verification (local checkout @ `8b89194318ed`, pnpm installed)
| Check | Result |
|---|---|
| `vitest run src/sheet/binding.test.ts` | ✅ 42/42 (incl. new WS-28 auto-height case) |
| mutation id `set-worksheet-row-auto-height` is real | ✅ present in `@univerjs/sheets@0.25.1` (+ preset), sibling to the working `set-worksheet-row-height` — not a wrong-id no-op |
| key-shape / restore-path parity | ✅ writes `${logical}:r${row}` via same dimMap/lastSeenDims/transact → restored by the existing `applyRemoteDims` |

Does exactly what the description says and no more: one new mutation listener + its handler, keyed identically to the manual-resize path, restored by the existing remote-dims apply. No schema/backend touch; legacy docs lacking the key behave as today (content-driven height) — no migration needed. No scope creep.

### Gate 2 — Code quality: ✅
- **Faithful mirror of the proven manual path.** `syncAutoRowHeightFromCommand` resolves `logical` from the active sheet exactly like `syncDimFromCommand`, writes the identical `${logical}:r${idx}` key into the same `dimMap` (with `lastSeenDims` + `doc.transact`), so reload restores via the same validated `applyRemoteDims` — no new restore path to get wrong. Correct diagnosis (auto-height vs manual are distinct Univer mutations) and minimal fix.
- **Input validation is tighter than the sibling.** Each entry is gated on `Number.isInteger(row)`, `typeof h === 'number' && Number.isFinite(h) && h > 0` before write — rejects malformed `rowsAutoHeightInfo`. And the read side (`applyRemoteDims`) independently bounds-checks the index (`0 ≤ idx < SHEET_MAX_ROWS`) and validates the value — defense-in-depth against a hostile/buggy peer writing `Infinity/NaN/-1`. Sound both directions.
- **Guards for missing workbook/sheet/logical/empty-infos** all present → no throw, clean no-op when state isn't ready.
- **Write-gated by `canWrite()`** (same dispatch branch as the other sync paths) → read-only viewers don't emit.

### Tests — genuinely exercises the fix
The new case fires the real mutation id with a two-row `rowsAutoHeightInfo` and asserts both rows land in the dim map with their auto heights — directly locking the sync that was missing. Note it's a unit test of the binding sync; the full toggle-wrap→grow→reload→restore round-trip is covered structurally (same key/apply path as manual resize, which the suite's other 41 tests exercise) but a live smoke is the belt-and-suspenders (see note).

### Non-blocking notes
- **Active-sheet assumption:** heights key off `getActiveSheet()` — correct for an interactive wrap toggle (acts on the active sheet), matching `syncDimFromCommand`. If Univer ever emitted auto-height for a non-active sheet (e.g. a batched/programmatic recompute), it'd be mis-keyed; not a concern for the WS-28 interactive path, worth a mental note if auto-height is later triggered in bulk.
- **Human smoke recommended** (per author): toggle wrap → row grows → leave & reopen → height persists, across a collaborator too. The data-sync layer is unit-covered; the live Univer round-trip can't be verified headless.

mutation id real (not silent no-op) ✅ · key shape identical to manual resize → restored by same path ✅ · write validation (Integer/Finite/>0) ✅ · read validation + bounds (hostile-peer safe) ✅ · missing-state guards ✅ · canWrite-gated ✅ · no schema/migration ✅ · build + 42/42 ✅.

---

### ✅ Review Verdict: APPROVE @ `8b89194318ed`
Spec ✅ and quality ✅ — a minimal, correctly-diagnosed fix: it listens for the real Univer auto-height mutation and persists row heights through the identical dim-map/key/restore path as manual resizes, so reload restores them with no new code to go wrong. Validated on both write and read sides (hostile-peer safe), frontend-only with no migration, build green, 42/42 tests including the new WS-28 case. Non-blocking: a live collab smoke of the toggle→reload round-trip. No blockers.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #887 (octo-web)

Reviewed at head `8b89194318ed22bbee49f664593430f96b6792ad`.

## Summary

This PR fixes WS-28: toggling 自动换行 (text-wrap) on a Sheet cell grows the row to fit, the UI looks correct locally, but on reload the row snaps back to default height and clips the wrapped text. The root cause is correctly diagnosed — Univer emits a distinct `set-worksheet-row-auto-height` mutation for layout-trigger recompute (carrying per-row `rowsAutoHeightInfo`), while the binding's `DIM_TRIGGER_IDS` only listens to the manual `set-worksheet-row-height`. The grown height never reached the Y.Doc.

The fix adds a new trigger set `AUTO_ROW_HEIGHT_TRIGGER_IDS` and a new handler `syncAutoRowHeightFromCommand` that writes to the **same** `sheetDims` Y.Map with the **same** `${logicalId}:r<idx>` key shape used by the manual resize path, so reload restores the grown height via the existing `applyRemoteDims` consume path. 42 additions in `packages/docs/src/sheet/binding.ts` (plus 18 lines of regression test). Narrow, well-scoped, and correctly routed.

## Verification

Static analysis only at head `8b89194318ed22bbee49f664593430f96b6792ad`; `pnpm typecheck` / `vitest` not executed in this environment.

- ✅ **Data-flow write path** — new handler at `packages/docs/src/sheet/binding.ts:826` writes `${logical}:r${row}` into `this.dimMap` inside a `transact()` block and updates `lastSeenDims`, exactly mirroring `syncDimFromCommand` (line 796). Key shape matches what `applyRemoteDims` (line 856) parses (`c<idx>`/`r<idx>` after the `:`).
- ✅ **Consume path reuses existing guards** — `applyRemoteDims` already bounds-checks row index against `SHEET_MAX_ROWS`, validates value is finite/positive/`≤ SHEET_MAX_DIM_PX`, isolates per-item throws, and is wrapped in `applyingRemote=true` (line 857) so any echo from `sheet.setRowHeight` is suppressed at the `onCommandExecuted` entry guard (line 323).
- ✅ **Write-gate + echo guard** — the new trigger is routed inside the same `onCommandExecuted` callback that checks `this.disposed || this.applyingRemote` and `this.canWrite()` (lines 323–324) before any branch fires, so reader-mode / disposed / remote-apply states are honored uniformly.
- ✅ **Multi-sheet correctness** — `localToLogical.get(sheet.getSheetId())` (line 837) uses the ACTIVE sheet, matching the manual resize path (line 807); an interactive wrap toggle can only happen on the active sheet, so the logical-id prefix is correct.
- ✅ **Input validation on the new path** — defensive guards at `:839–841` (empty array bail-out, `Number.isInteger(row)`, `Number.isFinite(h)`, `h > 0`) mean a malformed/changed Univer payload silently no-ops rather than corrupting `sheetDims`.
- ✅ **Regression test coverage** — `packages/docs/src/sheet/binding.test.ts:467–484` asserts the new mutation flows through `FakeUniver.fire` into the dim Y.Map with the expected keys, catching future removal or re-routing of this trigger.
- ✅ **Diff scope** — only `packages/docs/src/sheet/binding.ts` and `packages/docs/src/sheet/binding.test.ts`; no schema, no backend, no other sheet surface touched.

## Findings

No P0/P1 issues. The change preserves the runtime contract end-to-end: the new write path produces keys that the existing, well-guarded consume path already knows how to consume.

### Nit — Defensive test coverage for malformed auto-height payloads (`packages/docs/src/sheet/binding.test.ts:467`)

The handler at `packages/docs/src/sheet/binding.ts:839–841` carefully rejects empty arrays, missing `rowsAutoHeightInfo`, non-integer rows, non-finite heights, and `h <= 0`. The regression test only covers the happy path (well-formed `rowsAutoHeightInfo`). Worth one or two negative cases (e.g. `rowsAutoHeightInfo: []` asserts `dimMap` stays empty; `autoHeight: NaN` is skipped) to lock those guards against silent drift. Not blocking — the guards are correct today, just not asserted.

## Data Flow Tracing

- **Write path**: `onCommandExecuted` → id matches `AUTO_ROW_HEIGHT_TRIGGER_IDS` → `syncAutoRowHeightFromCommand` → `workbook().getActiveSheet()` → `localToLogical.get(sheetId)` (=logical) → iterates `rowsAutoHeightInfo`, validates `Number.isInteger(row)` + `Number.isFinite(h) && h > 0` → builds `[`${logical}:r${row}`, h]` entries → `dimMap.doc.transact(() => { lastSeenDims.set; dimMap.set })`.
- **Persist**: `dimMap` = `ydoc.getMap<number>(SHEET_DIMS_FIELD)` where `SHEET_DIMS_FIELD === 'sheetDims'` (line 91, 305). Crosses the Yjs channel to every collaborator.
- **Reload / remote consume**: `applyRemoteDims(keys)` at line 856. Parses `${logicalId}:r<idx>` (or V1 unprefixed `r<idx>` → `default`), validates `0 ≤ idx < SHEET_MAX_ROWS`, validates `Number.isFinite(v) && v > 0 && v ≤ SHEET_MAX_DIM_PX`, calls `sheet.setRowHeight(idx, v)` under `applyingRemote=true`. The new path's writes are read verbatim by this existing, well-guarded consumer.
- **Echo suppression**: `applyingRemote=true` wraps `applyRemoteDims` end-to-end (line 857/903). If Univer's `setRowHeight` internally fires `set-worksheet-row-height`, the entry guard at line 323 (`if (this.disposed || this.applyingRemote) return`) drops it. No duplicate Y.Map write.
- **Multi-sheet**: key is `${logicalId}:r<idx>` where logical is looked up from the ACTIVE sheet's local id. A wrap toggle can only be triggered interactively on the active sheet, so the key is attributed correctly. A non-active-sheet row height arriving via remote update is still applied because `applyRemoteDims` resolves `logicalId → localId` via `logicalToLocal`.

## Blind-spot Checklist

- **C1 (dual-path parity)** — CLEAR. The new auto-height write path uses the identical key shape, identical `dimMap`/`lastSeenDims` writes, identical `transact()` wrapping, and identical active-sheet lookup as `syncDimFromCommand`. The read path (`applyRemoteDims`) is unchanged and treats both origins uniformly.
- **C2 (control-flow ordering)** — CLEAR. The new branch sits at the same nesting level as the existing `DIM_TRIGGER_IDS` branch inside `onCommandExecuted`; no double-dispatch, no ordering-sensitive state.
- **C3 (auth boundary)** — N/A. No auth/permission change.
- **C5 (build/deployment)** — N/A. No build/deployment changes.
- **C6 (governance/docs)** — N/A. No policy/documentation changes.

- V1 legacy handling: `applyRemoteDims` still treats unprefixed keys as `default` (line 871), and the new path always writes prefixed keys, so no regression for legacy single-sheet docs.
- Last-write-wins semantics between auto and manual paths: a subsequent manual resize overwrites the auto entry under the same key; reload restores the user's most recent intent.
- `applyingRemote` echo suppression wraps the entire `applyRemoteDims` loop — if Univer's `setRowHeight` internally fires a mutation, the entry guard drops it.
- No security-sensitive surface: no new input from remote peers reaches Univer without the existing finite/positive/ceiling validation in `applyRemoteDims`; the new local-write path only persists values Univer itself produced.

## Verdict: APPROVED

Narrow, well-justified fix. Data-flow contract matches the existing, well-guarded consume path. The only suggestion is a minor regression-test expansion to lock the defensive guards against future drift; non-blocking.

[Octo-Q] verdict: APPROVE — 无 P0/P1；实现路径与现有 syncDimFromCommand / applyRemoteDims 合约完全对齐，echo guard + write gate + 输入校验均已就位；唯一 nit 是补强负向测试用例（非阻塞）。

Jerry-Xin (dismissed):
This PR addresses a real Sheet collaboration bug (WS-28): auto-height growth from text-wrap never reached the Y.Doc, so wrapped rows snapped back on reload. The write→restore round-trip is sound in the common single-sheet case, but the new sync routes heights to the **active** sheet instead of the worksheet the mutation actually targeted, which can persist row heights under the wrong sheet.

### 🔴 Blocking

- **Wrong-sheet persistence: key by `subUnitId`, not the active sheet.** `packages/docs/src/sheet/binding.ts:834` (`syncAutoRowHeightFromCommand`) resolves the target via `wb.getActiveSheet()` and keys the dim entries under that sheet's logical id. But `sheet.mutation.set-worksheet-row-auto-height` carries `subUnitId` in its params (`ISetWorksheetRowAutoHeightMutationParams { unitId, subUnitId, rowsAutoHeightInfo }`, @univerjs/sheets 0.25.1), identifying the worksheet that was actually recalculated. Univer's own `AutoHeightController.getUndoRedoParamsOfAutoHeight` computes `subUnitId` from the intercepted command's target and comments explicitly: *"Better NOT use getActiveWorksheet method, because users may manipulate another worksheet in active sheet."* So auto-height can fire for a **non-active** sheet (programmatic value/style edits, recalc cascades, multi-sheet ops). When `subUnitId !== activeSheet.getSheetId()`, the recomputed heights are written under the active sheet's `${logical}:r<idx>` keys and, on reload, `applyRemoteDims` re-applies them to the wrong sheet — corrupting that sheet's row heights while the intended sheet stays unpersisted.
  - Fix: derive the logical id from the mutation's `subUnitId` (e.g. `const sub = params.subUnitId; const logical = this.localToLogical.get(sub)`), consistent with how the mutation handler routes via `getSheetCommandTarget(params)`. Fall back / bail if unmapped.
  - Add a multi-sheet regression test where the mutation's `subUnitId` differs from the active sheet and assert the height lands under the correct sheet's key (and NOT the active sheet's).

Note: the pre-existing `syncDimFromCommand` (manual resize, `binding.ts:797`) also keys by active sheet, but a manual drag is inherently a gesture on the visible/active sheet, so `subUnitId === active` holds in practice. Auto-height is different — it is interceptor-driven and Univer deliberately routes it by `subUnitId` — so mirroring the manual path's active-sheet assumption is not safe here.

### 🟡 Non-blocking

- The write path filters `h <= 0` / non-finite but does not cap at `SHEET_MAX_DIM_PX` or bounds-check `row < SHEET_MAX_ROWS`. The restore path (`applyRemoteDims`) already rejects out-of-range index and value, so an out-of-bounds auto-height becomes a benign orphan key rather than corruption — acceptable, and consistent with the existing manual path. Optional: bound at write time to avoid dead keys.

### ✅ Highlights

- Round-trip design is correct for the single-sheet case: writes reuse the exact `${logical}:r<idx>` key shape and the same `dimMap` + `lastSeenDims` as manual resize, so restore flows through the existing `applyRemoteDims` with no special-casing.
- Back-compat verified: existing docs missing the key simply have no entry → no restore attempted → default height (behavior unchanged), no crash/0-height.
- The mutation id and `rowsAutoHeightInfo` param shape match @univerjs/sheets 0.25.1 exactly.
- Invalid/non-positive heights are filtered; `Number.isInteger(row)` correctly admits `row: 0`.
- The focused WS-28 test clearly documents the original failure mode.

lml2468 (changes requested):
## 🐙 PR #887 Re-Review (correction) — REQUEST_CHANGES @ `8b89194318ed`

**I'm correcting my APPROVE.** Jerry-Xin flagged that the auto-height sync writes to the wrong sheet when the recompute targets a non-active sheet. I verified it against Univer's type defs — **it's a real data-correctness bug, and I under-graded it** (I noted the active-sheet assumption as a non-blocking aside; it's actually a 🔴).

### 🔴 P2 (BLOCKER, CONFIRMED) — auto-height persisted under the WRONG sheet's key when the recompute isn't on the active sheet

`syncAutoRowHeightFromCommand` resolves the target sheet from **`getActiveSheet()`** (:834) — but the mutation itself carries the real target. Univer's type def:
```ts
export interface ISetWorksheetRowAutoHeightMutationParams {
  unitId: string;
  subUnitId: string;          // ← the worksheet this recompute is FOR
  rowsAutoHeightInfo: IRowAutoHeightInfo[];
}
```
The handler **ignores `subUnitId`** and keys off the active sheet's logical id. So when `set-worksheet-row-auto-height` fires for a sheet that is **not** the currently-active one, the recomputed heights are written under the **active** sheet's `${logical}:r${row}` keys → on reload the wrong sheet's rows get resized and the correct sheet's don't.

**Reachable:** auto-height recompute is not guaranteed to be an interactive edit on the visible sheet — Univer recomputes row heights during batched ops, formula/content-driven relayout, and import/init while another sheet is active. The mutation *carrying `subUnitId`* is precisely the signal that active-sheet can't be assumed here.

**Why my "faithful mirror" praise was the trap:** `syncDimFromCommand` (manual resize) also uses `getActiveSheet()` — and that's *correct there*, because a manual drag can only happen on the visible sheet, so active == target by construction. This PR copied that pattern into the auto-height context, where the assumption **breaks** (auto-height can fire for a non-active sheet). The mirror imported an invariant that doesn't hold in the new context. That's exactly what I should have caught when I flagged the active-sheet assumption and then waved it off.

**Fix (one line, and the data is right there):** resolve from the mutation's `subUnitId` instead of the active sheet. `localToLogical` is keyed by the local Univer sheetId, and `subUnitId` **is** that local sheetId:
```ts
const p = command.params as { subUnitId?: string; rowsAutoHeightInfo?: Array<{ row?: number; autoHeight?: number }> } | undefined
...
const localId = p?.subUnitId
if (!localId) return
const logical = this.localToLogical.get(localId)
if (!logical) return
```
(Drop the `getActiveSheet()` lookup entirely.) Add a **multi-sheet regression test**: fire the auto-height mutation with a `subUnitId` for a non-active sheet and assert the height lands under *that* sheet's key, not the active one — this is the exact gap the current single-sheet test misses.

### 🟡 (Jerry-Xin) max-row/row-count bounds on write — already mitigated, minor
The write side validates `Number.isInteger(row)` / `Finite` / `>0` but not `row < SHEET_MAX_ROWS`. However `applyRemoteDims` **does** bounds-reject (`0 ≤ idx < SHEET_MAX_ROWS`) on the read side, so an out-of-range key can't drive a resize on restore — defense-in-depth already covers the blast. Adding the max-row guard on write (mirroring that constant) is a reasonable tidy, non-blocking.

### What still stands
The core diagnosis (auto-height is a distinct mutation that wasn't synced) and the dim-map/restore-path reuse are correct — the fix is 90% right. The defect is the sheet-resolution source: `getActiveSheet()` where it must be `subUnitId`.

---

### ❌ Review Verdict: REQUEST_CHANGES @ `8b89194318ed`
One confirmed data-correctness P2 I under-graded in my APPROVE: `syncAutoRowHeightFromCommand` keys row heights off `getActiveSheet()` but the `set-worksheet-row-auto-height` mutation carries `subUnitId` (the true target sheet) — so an auto-height recompute for a non-active sheet persists under the wrong sheet, resizing the wrong rows on reload. Resolve `logical` from `p.subUnitId` (one-line fix; `localToLogical` is keyed by the local sheetId = subUnitId) + a multi-sheet regression test. Credit to Jerry-Xin — and my miss: I flagged the active-sheet assumption but graded it non-blocking; the mutation carrying `subUnitId` makes the non-active case reachable and wrong.

yujiawei (changes requested):
# Code Review — PR #887 (octo-web)

**Head:** `8b89194318ed22bbee49f664593430f96b6792ad` · **Base:** `main` · +60 −1 / 2 files · `mergeable: MERGEABLE`

**Intent (WS-28):** toggling a cell's text-wrap (自动换行) grows the row to fit locally (UI correct), but the grown height never reaches the Y.Doc, so on reload the row snaps back and clips the wrapped text. Root cause is correctly diagnosed: Univer applies the growth via `sheet.mutation.set-worksheet-row-auto-height` (per-row `rowsAutoHeightInfo`), a **different** mutation than the manual `set-worksheet-row-height` the collab binding already syncs. The fix listens for the auto-height mutation and persists into the **same** dim map + key shape (`${logical}:r<idx>`) as manual resize, restored on reload via the existing `applyRemoteDims`.

## Verification

Checked the Univer runtime contract by vendoring the pinned packages (`@univerjs/sheets` / `sheets-ui` / `core` / `engine-render` @ `0.25.1`) and reading source (not just type defs):

- ✅ Mutation id `sheet.mutation.set-worksheet-row-auto-height` and param shape `{ unitId, subUnitId, rowsAutoHeightInfo: [{ row, autoHeight }] }` are real (sibling to the working manual `set-worksheet-row-height`) — not a wrong-id no-op.
- ✅ Restore mechanism is sound: `RowManager.getRowHeight` returns `(ia == null || ia === 1) && typeof ah === 'number' ? ah : h`. On reload `ah` (auto-height) is absent but the persisted manual `h` is present, so the row renders at the persisted height. The write→restore round-trip works in the single-sheet case.
- ✅ Value-cap asymmetry is NOT a bug (I probed it): the write side accepts any finite `h > 0` (no upper cap) while the read side (`applyRemoteDims`) rejects `v > SHEET_MAX_DIM_PX` (2000). Univer's `calculateAutoHeightInRange` clamps with `Math.min(height, MAXIMUM_ROW_HEIGHT)` where `MAXIMUM_ROW_HEIGHT = 2000`, exactly equal to `SHEET_MAX_DIM_PX`, and the read filter is `v > 2000` (so 2000 is accepted). No legitimate auto-height can be silently dropped on reload.

## 1. Spec compliance: ✅

Does exactly what the description says and no more: one new trigger set (`AUTO_ROW_HEIGHT_TRIGGER_IDS`) + one handler (`syncAutoRowHeightFromCommand`) + one dispatch branch, keyed identically to the manual-resize path. Frontend-only, no schema/backend touch, no migration (legacy docs lacking the key behave as today — content-driven height). No scope creep, no over-build.

## 2. Code Quality: **Changes-Requested**

### 🔴 P1 — Wrong-sheet persistence: key by the mutation's `subUnitId`, not the active sheet

`packages/docs/src/sheet/binding.ts:833-837` — `syncAutoRowHeightFromCommand` resolves the target via `wb.getActiveSheet()` and keys the dim entries under that sheet's logical id, **discarding the `subUnitId` the mutation carries**:

```ts
const wb = this.workbook()
const sheet = wb?.getActiveSheet()          // ← wrong source of truth
if (!wb || !sheet) return
const logical = this.localToLogical.get(sheet.getSheetId())
```

`ISetWorksheetRowAutoHeightMutationParams` is `{ unitId, subUnitId, rowsAutoHeightInfo }` — `subUnitId` identifies the worksheet that was actually recalculated. This is not merely a type detail; I traced the emitters in `@univerjs/sheets-ui@0.25.1`:

- `AutoHeightController.getUndoRedoParamsOfAutoHeight(ranges, subUnitIdParam, ...)` resolves the worksheet from `subUnitIdParam` via `getSheetCommandTarget` and only falls back to `getActiveSheet()` when it's absent — i.e. Univer deliberately routes auto-height by `subUnitId`, precisely because a recompute can target a non-active sheet.
- The **deferred** path is worse for the active-sheet assumption: `MarkDirtyRowAutoHeightOperation → startAutoHeightTask → _loopTask` re-emits `SetWorksheetRowAutoHeightMutation` **asynchronously** (via `requestIdleCallback`) with `subUnitId: task.sheetId`. By the time that fires, the active sheet may have changed entirely.

**Impact:** when `subUnitId !== activeSheet.getSheetId()` (batched/programmatic edits, formula/content-driven relayout, import/init, or the deferred off-screen recompute), the recomputed heights are written under the **active** sheet's `${logical}:r<idx>` keys. On reload `applyRemoteDims` re-applies them to the wrong sheet — corrupting that sheet's row heights while the intended sheet stays unpersisted. This is a data-correctness defect reachable in multi-sheet collaboration.

**Fix** — derive the logical id from the mutation's `subUnitId` (`localToLogical` is keyed by the local Univer sheetId, and `subUnitId` **is** that local sheetId), dropping the `getActiveSheet()` lookup entirely:

```ts
const p = command.params as { subUnitId?: string; rowsAutoHeightInfo?: Array<{ row?: number; autoHeight?: number }> } | undefined
const localId = p?.subUnitId
if (!localId) return
const logical = this.localToLogical.get(localId)
if (!logical) return
```

Add a **multi-sheet regression test**: fire the mutation with a `subUnitId` for a non-active sheet and assert the height lands under *that* sheet's key and NOT the active sheet's. The current single-sheet test (`binding.test.ts:467`) omits `subUnitId`, so it cannot catch this.

> Note: the pre-existing `syncDimFromCommand` (manual resize, `:797`) also keys by active sheet — but a manual drag is inherently a gesture on the visible/active sheet, so `active === target` holds there. Auto-height is interceptor-driven and Univer routes it by `subUnitId`, so mirroring the manual path's active-sheet invariant is unsafe in this new context.

### 🟡 P2 — Deferred auto-height can echo a remote edit back into the Y.Doc (past the `applyingRemote` guard)

`binding.ts:323` gates all local→Yjs sync on `this.applyingRemote`, which `applyRemoteToUniver` sets true and clears in its `finally` (`:792`). For auto-height triggered **synchronously** inside a remote `setValue`, this correctly suppresses the echo. But the **deferred** auto-height (off-screen ranges, scheduled via `requestIdleCallback`) runs on a later tick when `applyingRemote` is already `false` — so a remote-derived layout change (a peer's wrapped-text cell arriving via `applyRemoteToUniver`) re-triggers a local auto-height recompute that is then treated as local and written back into the Y.Doc.

Combined with the P1 above this can persist the echoed height under the **wrong** (active) sheet even though the originating edit was remote. Fixing P1 (key by `subUnitId`) demotes this to a redundant same-value write (write amplification/noise) rather than corruption; adding a `lastSeenDims` diff-guard before writing (skip when the value is unchanged) removes the redundant churn entirely. Worth addressing alongside P1.

### 🟡 P2 — Write side does not bound `row < SHEET_MAX_ROWS` or cap the value

`binding.ts:842` validates `Number.isInteger(row)` / `Number.isFinite(h)` / `h > 0` but not `0 ≤ row < SHEET_MAX_ROWS` nor a `SHEET_MAX_DIM_PX` cap. The read path (`applyRemoteDims`, `:879-891`) already bounds-rejects both, so an out-of-range write becomes a benign orphan key rather than a resize-on-restore — acceptable and consistent with the manual path. Optional tidy: bound at write time to avoid dead keys.

## 3. Overall verdict: **REQUEST_CHANGES**

Spec ✅, but Code Quality has one confirmed P1 data-correctness defect. The diagnosis and the dim-map/restore-path reuse are correct — the fix is ~90% right; the defect is the sheet-resolution source: `getActiveSheet()` where it must be the mutation's `subUnitId`. This aligns with and independently confirms the existing change requests on this PR; I additionally verified the non-active/deferred emission against Univer 0.25.1 source and surfaced the deferred-echo compounding case.

## 4. Suggested fix direction

1. Resolve `logical` from `command.params.subUnitId` (not `getActiveSheet()`); bail if unmapped.
2. Add a multi-sheet regression test (`subUnitId` ≠ active) asserting correct-sheet keying.
3. (Recommended) Diff against `lastSeenDims` before writing to suppress redundant/echoed same-value writes.
4. (Optional) Mirror the read-side row/value bounds on the write path.

## 5. Additional observations

- **Security (this PR is flagged security-sensitive):** the change touches only the sheet dimension-sync path — no auth, token, credential, or middleware surface. The untrusted-remote-value handling lives on the read side (`applyRemoteDims`), which already rejects non-finite/negative/out-of-range/oversized values; this PR does not weaken it. No new trust-boundary exposure. The `subUnitId` defect is a correctness/data-integrity issue, not a security one.
- **Human smoke recommended** (data-sync layer is unit-covered, but the live Univer toggle→grow→reload→restore round-trip — and the multi-sheet non-active case — cannot be verified headless).

pkuWMH:
Addressed the change request in 649e2b1d:

- **Blocking fix**: `syncAutoRowHeightFromCommand` now resolves the target worksheet from the mutation's own `subUnitId` via `localToLogical.get(subUnitId)` instead of `getActiveSheet()`, and bails when `subUnitId` is absent or maps to no known sheet. An auto-height recompute firing against a non-active sheet no longer persists row heights under the wrong sheet's keys.
- **Write-side guards**: row bounded to `[0, SHEET_MAX_ROWS)` and height to `(0, SHEET_MAX_DIM_PX]`, mirroring the restore path.
- **Tests**: added a multi-sheet regression (`subUnitId !== activeSheet` → height lands under the subUnitId sheet, not the active one) and a bail-on-unmapped-subUnitId case. `packages/docs` suite: 44 passed.

yujiawei (approved):
# Code Review — PR #887 (octo-web) — Re-review @ `649e2b1d`

**Head:** `649e2b1d1a92088bc2c70d54ba06cf86eaca78cf` (was `8b89194318ed`) · **Base:** `main` · re-review after the fix commit *"route sheet auto row-height by mutation subUnitId, not active sheet"*.

The previous round requested changes on one confirmed P1 (wrong-sheet persistence) plus two P2s. The new commit addresses them. Re-reviewed against the **current** head (a new head does not inherit the prior verdict).

Commit `649e2b1d` (`packages/docs/src/sheet/binding.ts` + `binding.test.ts`):

- **P1 fixed — routes by the mutation's `subUnitId`.** `syncAutoRowHeightFromCommand` now reads `p.subUnitId` and resolves `this.localToLogical.get(p.subUnitId)`, dropping the `getActiveSheet()` lookup entirely. Bails when `subUnitId` is absent (`typeof p?.subUnitId !== 'string'`) or maps to no known local sheet (`if (!logical) return`). This is exactly the fix the review asked for and matches how Univer itself routes the mutation (its params carry the authoritative target worksheet).
- **P2 (write-side bounds) fixed.** The per-row guard now also enforces `row ≥ 0 && row < SHEET_MAX_ROWS` and `h ≤ SHEET_MAX_DIM_PX`, mirroring the restore path (`applyRemoteDims`) so no dead/out-of-range keys are written.
- **P2 (deferred remote-echo) mitigated by construction.** With heights now keyed by the true target `subUnitId`, a deferred auto-height recompute triggered by a remote edit writes to the **correct** sheet's key rather than the active one. That demotes the concern from wrong-sheet corruption to a benign redundant same-value write (LWW no-op) — non-blocking, as noted last round.

## Verification (local checkout @ `649e2b1d`, pnpm installed)

| Check | Result |
|---|---|
| `vitest run src/sheet/binding.test.ts` | ✅ **44/44** pass (was 42; +2 new routing tests) |
| `tsc --noEmit` on `@octo/docs` | ✅ no errors in `binding.ts` (only pre-existing unrelated errors: dmworkbase react@17 typings, `EditorShell.test`, `members/sort`) |
| Fix keys by `subUnitId` | ✅ confirmed at handler; `getActiveSheet()` removed |
| Write bounds mirror read path | ✅ `row < SHEET_MAX_ROWS`, `h ≤ SHEET_MAX_DIM_PX` added |

New regression tests genuinely exercise the fix (not happy-path-only):
- *"routes an auto-height mutation to its OWN subUnitId sheet, not the active sheet"* — registers Sheet2 into the logical map, sets the FIRST sheet active, fires auto-height with `subUnitId = Sheet2`, and asserts the height lands under `${Sheet2}:r1` and **NOT** `default:r1`. I traced the FakeUniver lifecycle (`addSheet` → `insert-sheet` → `syncSheetListFromUniver` populates `localToLogical`); without the fix this test fails (it would write under the active `default`). A true regression guard.
- *"ignores an auto-height mutation whose subUnitId maps to no known sheet"* — asserts the unmapped-`subUnitId` bail (nothing written).

## 1. Spec compliance: ✅
Frontend-only, no schema/backend/migration. The fix stays within scope (sheet dimension-sync), no over-build.

## 2. Code Quality: ✅ Approved
The one blocking P1 is resolved at the correct layer, both write-side P2 bounds are added, and the deferred-echo case is neutralized as a side effect of the P1 fix. Restore mechanism unchanged and already validated (`applyRemoteDims` bounds/finite/ceiling checks, `applyingRemote` echo guard). Tests green, changed file typecheck-clean.

## 3. Overall verdict: **APPROVE** @ `649e2b1d`
Spec ✅ AND Quality ✅. The confirmed data-correctness defect from the prior round (auto-height persisted under the active sheet instead of the mutation's target) is fixed by routing off `subUnitId`, with targeted multi-sheet regression coverage and write-side bounds. No remaining blockers.

## 5. Additional observations
- **Security (PR flagged security-sensitive):** no auth/token/credential/middleware surface touched; untrusted-remote-value handling on the read side is unchanged and still enforced. No new trust-boundary exposure.
- **Non-blocking:** a live collab smoke of the toggle-wrap → grow → reload → restore round-trip (and the multi-sheet non-active case) remains worthwhile — the data-sync layer is unit-covered, but the live Univer round-trip can't be verified headless.

Jerry-Xin (approved):
Re-review at `649e2b1d1a92`. The prior blocking issue — auto-height persisted under the active sheet instead of the mutation's target — is now fixed correctly, with real regression coverage. Approving.

### ✅ Resolved (was 🔴 Blocking)

- `syncAutoRowHeightFromCommand` now resolves the target sheet from the mutation's own `subUnitId` (`binding.ts:838` — `const logical = this.localToLogical.get(p.subUnitId)`) instead of `getActiveSheet()`. `getActiveSheet()` is no longer referenced anywhere in the auto-height write path. Since `localToLogical` is keyed by the local Univer sheetId (populated via `logicalFor(s.getSheetId(), ...)`) and Univer's mutation `subUnitId` is exactly that local sheetId, the lookup routes heights to the right worksheet. An auto-height pass firing against a non-focused sheet no longer corrupts the active sheet's keys.

- Guard for unknown/missing `subUnitId` is safe: `if (typeof p?.subUnitId !== 'string') return` and `if (!logical) return` — an absent or unmapped `subUnitId` skips the write entirely (no fallback to active sheet, no crash).

### ✅ Write bounds now mirror the restore path

`binding.ts:844-845` reject row `< 0 || >= SHEET_MAX_ROWS` and height `!finite || <= 0 || > SHEET_MAX_DIM_PX` — identical to the ceilings enforced in `applyRemoteDims` (`:883`, `:894`). Legitimate auto-heights pass; this only prevents persisting a value the restore path would later drop, keeping write/restore consistent.

### ✅ Regression coverage is real

`binding.test.ts` adds a non-active-sheet test: it registers `Sheet2`, sets it active, then re-activates the first sheet before firing the auto-height mutation with `subUnitId: s2.getSheetId()`. It asserts the height lands under `Sheet2`'s key AND explicitly that `default:r1` (the active sheet) is NOT written. Reverting to `getActiveSheet()` would write `default:r1` and fail that assertion, so the test genuinely locks the fix. A second test asserts an unknown `subUnitId` ('ghost-sheet') writes nothing. The original single-sheet WS-28 test still passes (`subUnitId: 'local-1'` → logical `default`), confirming no happy-path regression.

Nice work — minimal, correct, and well-tested.

lml2468 (approved):
## 🐙 PR #887 Re-Review — `fix(docs): persist sheet auto row-height (WS-28)`

**Head:** `649e2b1d1a92` (was `8b89194318ed`) · **Base:** `main` · **prior verdict: REQUEST_CHANGES** (wrong-sheet routing).

### 🔴 → ✅ The wrong-sheet bug is fixed — exactly the `subUnitId` routing I/Jerry-Xin/Steve specified
Delta since my CR head = the two binding files. `syncAutoRowHeightFromCommand` now keys off the mutation's **own target sheet**, not the active one:
```ts
if (typeof p?.subUnitId !== 'string') return
const logical = this.localToLogical.get(p.subUnitId)   // was: getActiveSheet().getSheetId()
if (!logical) return
```
`getActiveSheet()` is dropped entirely. So an auto-height recompute that fires against a non-active sheet now persists under **that** sheet's `${logical}:r${row}` keys — no more cross-sheet clobber on reload. Correct fix.

### ✅ Regression test added — the exact multi-sheet case the old test missed
`binding.test.ts` gains two cases:
- **"routes an auto-height mutation to its OWN subUnitId sheet, not the active sheet"** — makes Sheet2, sets `default` active, fires auto-height with `subUnitId: Sheet2`, asserts the height lands under `Sheet2:r1` and **`default:r1` is absent**. This is precisely the cross-sheet routing that would fail on the old `getActiveSheet()` code — a genuine fail-on-revert guard.
- **"ignores an auto-height mutation whose subUnitId maps to no known sheet"** — unknown-subUnitId → no write (guards the new `!logical` bail).
- The original WS-28 case updated to carry `subUnitId`.

### ✅ 🟡 hardening also applied
Write side now bounds-checks `0 ≤ row < SHEET_MAX_ROWS` and `0 < h ≤ SHEET_MAX_DIM_PX`, mirroring the restore-path validation — the symmetric hardening (optional in my note) is in.

### Verification (local checkout @ `649e2b1d1a92`, pnpm installed)
| Check | Result |
|---|---|
| `vitest run src/sheet/binding.test.ts` | ✅ 44/44 (was 42; +2 routing cases) |
| subUnitId routing | ✅ keys off mutation target, active-sheet lookup removed |
| non-active-sheet regression test | ✅ present + asserts wrong-sheet key absent |
| CI `code-review` | ✅ approved at this head |

subUnitId is the authoritative target (Univer's own AutoHeightController avoids getActiveWorksheet for this reason) ✅ · active-sheet routing removed ✅ · unknown-subUnitId no-ops ✅ · write bounds mirror restore path ✅ · regression test fails-on-old-code ✅ · build + 44/44 ✅. The core diagnosis/dim-map reuse (correct all along) is intact.

---

### ✅ Review Verdict: APPROVE @ `649e2b1d1a92`
My REQUEST_CHANGES blocker (auto-height keyed off `getActiveSheet()` → persisted under the wrong sheet for a non-active recompute) is fixed at the root: it now resolves the target from the mutation's `subUnitId`, with a multi-sheet regression test that locks the correct routing and the write-side bounds hardening applied. Build green, 44/44. This clears my CR. Clean to merge. Credit to Jerry-Xin for the original catch and the whole thread for converging on the one-line root fix.

PR #846, opened 2026-07-16
feat(docs): HTML doc system integration + parameterized octo-doc nginx upstreams

Ranwanglc:
## 概述

将 HTML 文档系统（octo-doc / octo-docs-html 前端接入）整套并入 web，并让 octo-doc 相关上游在 nginx 层参数化，便于跨环境部署。分支已同步至上游最新 `main`，冲突已解，`apps/web` 构建通过。

## 主要内容

1. HTML 文档类型接入
   - 文档列表新增 `html` 类型（图标、类型筛选、侧栏按 `octoDocSlug` 打开）
   - 只读 HTML 文档：sandboxed iframe 渲染（完整 CSS 保真）、octo-doc 评论/reactions、评论源文摘录、AI-handle 入口
   - 成员管理面板（基于 octo-doc grants），author-only UI 门控（按后端 `__ODOC_CAP__.isAuthor`）
   - 渲染 fetch 带出 octo `token` 头用于鉴权

2. nginx 上游参数化（可移植部署）
   - `/d`、`/comments`、`/reactions`、`/grants`、`/docs-admin` 上游改用 `${DOC_APP_URL}`
   - `/api/v1/docs` 上游改用 `${DOCS_BACKEND_URL}`
   - 两个变量加入 entrypoint envsubst 白名单并给默认值；token 透传保持不变
   - 换环境不再改 nginx，只配 env

## 部署前提 / 必配环境变量（web 容器）

| 变量 | 默认值 | 不配的后果 |
|---|---|---|
| `API_URL` | 无 | 必配。为空则 nginx `proxy_pass /;` 非法，容器起不来 |
| `DOC_APP_URL` | `http://doc-app-test:8080` | 有默认；跨环境需改成实际 octo-docs-html 渲染服务地址 |
| `DOCS_BACKEND_URL` | `http://octo-docs-backend-3001:3000` | 有默认；跨环境需改成实际 docs-backend 地址 |

可选（为空则对应功能 503，不影响启动）：`SUMMARY_API_URL`、`MATTER_API_URL`、`DOCS_ASSET_CSP_ORIGIN`。

> 注：HTML 文档的 author 鉴权还依赖 doc 侧（octo-docs-html）容器的 `LOGIN_ENABLED=true` 等 env，属另一仓库，不在本 PR 范围。

## 验证

- 冲突：仅 `packages/docs/src/pages/DocsHome.tsx` 一处（kindLabel 三元表达式），已解为「保留 html 分支 + 采用上游 latest-event 行」
[...]

PR #888, opened 2026-07-19
fix(nginx): repair octo-doc proxy query encoding, restore SPA fallback + unify /docs-html/* prefix

Ranwanglc:
## 概述
修复并收敛 nginx 中 web→octo-doc 的代理路由，解决 #846 引入的三处回归，并按评审意见把所有 octo-doc 流量统一到单一 `/docs-html/*` 前缀下。

## 改动
**1. `fix(nginx)`：修复 proxy query 编码 + 恢复 SPA fallback（153d5dfa）**
- `/d/` 与 `/api/v1/docs` 之前用 `rewrite ^ $request_uri break;` 配合变量 proxy_pass，变量上游会对 URI 做 URL 编码，`?` 被转成 `%3F`，导致 `/api/v1/docs?owner=me` 到后端变成 `%3F` 而 404（docs「我的文件」列表、`/d/?code=` 等全挂）。改为 `proxy_pass $var$request_uri;` 逐字转发 path+query，还原 #846 前行为。
- `location /` 丢了 SPA fallback（上游模板缺 `try_files $uri $uri/ /index.html;`），客户端路由/深链/刷新硬 404。恢复之。

**2. `feat(nginx)`：统一 web→octo-doc 到单一 `/docs-html/*` 前缀（291d7a78）**
- 前端 `resolveOctoDocBase()` 默认值从 `''`（同源根）改为 `/docs-html`；所有 octo-doc 请求（render + comments/reactions/grants/docs-admin/versions）自动加前缀（单一收口，无需逐调用改）。运行时/构建覆盖（`__OCTO_DOC_BASE__` / `VITE_OCTO_DOC_BASE`）仍优先。同步更新 2 处单测预期。
- nginx 用单个 `location /docs-html/` 取代原先分散的五个 location，剥掉前缀并把各子路径重写到 octo-doc 原生路由，保留 resolver + 未配置时 503 的安全模式。

## 验证
- `envsubst` 渲染后 `nginx -t` 通过。
- 本地 3001 重建：`/api/v1/docs?owner=me` 返回 401（后端可达，原为 404）、`/d/<slug>?code=` 逐字转发、SPA 深链+刷新 200。
- `/docs-html/*` 打到 octo-doc（与直连 octo-doc 基线一致）；mano-cua（laosan 登录）确认 render + comment POST 走 `/docs-html/*` 均 200。

## 变更文件
- `nginx.conf.template`
- `packages/docs/src/html/HtmlDocView.tsx`
- `packages/docs/src/html/HtmlDocView.test.tsx`

### DIFF
--- a/nginx.conf.template
+++ b/nginx.conf.template
@@ -112,10 +112,10 @@ server {
         location /d/ {
             resolver 127.0.0.11 ipv6=off valid=30s;
             if ($doc_app_url = "") { return 503 '{"status":503,"msg":"octo-doc is not configured"}'; }
-            # No location-prefix rewrite happens with a variable proxy_pass; forward the
-            # original URI verbatim so /d/... reaches the doc unchanged.
-            rewrite ^ $request_uri break;
-            proxy_pass $doc_app_url;
+            # Variable proxy_pass: append the original URI (incl. query) directly so
+            # /d/<slug>?... reaches the doc verbatim. A `rewrite ^ $request_uri break`
+            # would URL-encode the `?` into %3F and break query strings.
+            proxy_pass $doc_app_url$request_uri;
             proxy_set_header token $http_token;
             proxy_set_header Host $host;
             proxy_set_header X-Real-IP $remote_addr;
@@ -188,8 +188,11 @@ server {
         location /api/v1/docs {
             resolver 127.0.0.11 ipv6=off valid=30s;
             if ($docs_backend_url = "") { return 503 '{"status":503,"msg":"octo-docs-backend is not configured"}'; }
-            rewrite ^ $request_uri break;
-            proxy_pass $docs_backend_url;
+            # Variable proxy_pass: append the original URI (incl. query) directly.
+            # A `rewrite ^ $request_uri break` would URL-encode the `?` into %3F
+            # (breaking list?owner=me → 404); appending $request_uri to the
+            # variable forwards path+query verbatim, matching the pre-rewrite behavior.
+            proxy_pass $docs_backend_url$request_uri;
             proxy_set_header Host $host;
             proxy_set_header X-Real-IP $remote_addr;
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
@@ -238,5 +241,9 @@ server {
         location / {
             root   /usr/share/nginx/html;
             index  index.html index.htm;
+            # SPA fallback: unknown paths (client-side routes like the docs "my
+            # files" view, deep links, refresh) fall through to index.html so
+            # React Router can handle them, instead of a hard 404.
+            try_files $uri $uri/ /index.html;
         }
 }
--- a/nginx.conf.template
+++ b/nginx.conf.template
@@ -104,71 +104,35 @@ server {
             client_max_body_size 100m;
         }
 
-        # octo-doc HTML render + agent-auth token passthrough. Upstream is env-driven
-        # ($doc_app_url) so the image ports to any environment via .env, no conf edit.
-        # The `token` header carries the octo login token for doc-side identity verify.
-        # resolver + runtime-variable proxy_pass: the doc host resolves per-request, so an
+        # ── octo-doc HTML: single unified prefix /docs-html/* ─────────────────────
+        # All web→octo-doc traffic is namespaced under ONE prefix (evaluator ask:
+        # the previously scattered top-level paths /d/, /comments, /reactions,
+        # /grants/, /docs-admin/, /docs/{slug}/versions used generic words that risk
+        # colliding with SPA / other service routes and are hard to govern). The
+        # frontend now prepends /docs-html (resolveOctoDocBase default); this block
+        # strips it and rewrites each sub-path to octo-doc's native route, then
+        # forwards to $doc_app_url with the octo `token` header for identity.
+        # resolver + runtime-variable proxy_pass: doc host resolves per-request, so an
         # unset/unresolvable DOC_APP_URL 503s this route instead of failing nginx startup.
-        location /d/ {
+        # Rewrites are ordered most-specific-first; each ends in `break`.
+        location /docs-html/ {
             resolver 127.0.0.11 ipv6=off valid=30s;
             if ($doc_app_url = "") { return 503 '{"status":503,"msg":"octo-doc is not configured"}'; }
-            # Variable proxy_pass: append the original URI (incl. query) directly so
-            # /d/<slug>?... reaches the doc verbatim. A `rewrite ^ $request_uri break`
-            # would URL-encode the `?` into %3F and break query strings.
-            proxy_pass $doc_app_url$request_uri;
-            proxy_set_header token $http_token;
-            proxy_set_header Host $host;
-            proxy_set_header X-Real-IP $remote_addr;
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        }
-
-        # octo-doc HTML comments/reactions: frontend calls same-origin /comments|/reactions
-        # (no /v1); rewrite to the doc's /v1/* and pass the octo token for identity.
-        location /comments {
-            resolver 127.0.0.11 ipv6=off valid=30s;
-            if ($doc_app_url = "") { return 503 '{"status":503,"msg":"octo-doc is not configured"}'; }
-            rewrite ^/comments(.*)$ /v1/comments$1 break;
-            proxy_pass $doc_app_url;
-            proxy_set_header token $http_token;
-            proxy_set_header Host $host;
-            proxy_set_header X-Real-IP $remote_addr;
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        }
-        location /reactions {
-            resolver 127.0.0.11 ipv6=off valid=30s;
-            if ($doc_app_url = "") { return 503 '{"status":503,"msg":"octo-doc is not configured"}'; }
-            rewrite ^/reactions(.*)$ /v1/reactions$1 break;
-            proxy_pass $doc_app_url;
-            proxy_set_header token $http_token;
-            proxy_set_header Host $host;
-            proxy_set_header X-Real-IP $remote_addr;
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        }
-
-        # octo-doc HTML per-doc reader grants: frontend calls same-origin
-        # /grants/{slug}[/{uid}]; rewrite to the doc's /v1/docs/{slug}/grants[/{uid}]
-        # and pass the octo token for identity. Top-level /grants prefix avoids the
-        # SPA's own /docs route. DELETE .../{uid} rewrite must precede GET/PUT.
-        location /grants/ {
-            resolver 127.0.0.11 ipv6=off valid=30s;
-            if ($doc_app_url = "") { return 503 '{"status":503,"msg":"octo-doc is not configured"}'; }
-            rewrite ^/grants/([^/]+)/(.+)$ /v1/docs/$1/grants/$2 break;
-            rewrite ^/grants/([^/]+)$      /v1/docs/$1/grants      break;
-            proxy_pass $doc_app_url;
-            proxy_set_header token $http_token;
-            proxy_set_header Host $host;
-            proxy_set_header X-Real-IP $remote_addr;
-            proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
-        }
-
-        # octo-doc HTML doc admin (delete): frontend calls same-origin /docs-admin/{slug};
-        # rewrite to the doc's /v1/docs/{slug} and pass the octo token for identity. Top-level
-        # /docs-admin prefix avoids the SPA's /docs route and the /v1/ block below (proxied to
-        # octo-server, which has no doc delete endpoint).
-        location /docs-admin/ {
-            resolver 127.0.0.11 ipv6=off valid=30s;
-            if ($doc_app_url = "") { return 503 '{"status":503,"msg":"octo-doc is not configured"}'; }
-            rewrite ^/docs-admin/(.+)$ /v1/docs/$1 break;
+            # per-doc grants: /docs-html/grants/{slug}[/{uid}] → /v1/docs/{slug}/grants[/{uid}]
+            # ({uid} form must precede the bare form)
+            rewrite ^/docs-html/grants/([^/]+)/(.+)$ /v1/docs/$1/grants/$2 break;
+            rewrite ^/docs-html/grants/([^/]+)$       /v1/docs/$1/grants      break;
+            # doc admin (delete): /docs-html/docs-admin/{slug} → /v1/docs/{slug}
+            rewrite ^/docs-html/docs-admin/(.+)$      /v1/docs/$1             break;
+            # versions: /docs-html/docs/{slug}/versions → /v1/docs/{slug}/versions
+            rewrite ^/docs-html/docs/(.+)$            /v1/docs/$1            break;
+            # comments / reactions: /docs-html/comments… → /v1/comments…
+            rewrite ^/docs-html/comments(.*)$         /v1/comments$1         break;
+            rewrite ^/docs-html/reactions(.*)$        /v1/reactions$1        break;
+            # render page: /docs-html/d/{slug}/v/{ver} → /d/{slug}/v/{ver} (verbatim)
+            rewrite ^/docs-html/(d/.*)$               /$1                    break;
+            # anything else under the prefix: strip /docs-html, forward as-is
+            rewrite ^/docs-html/(.*)$                 /$1                    break;
             proxy_pass $doc_app_url;
             proxy_set_header token $http_token;
             proxy_set_header Host $host;
--- a/packages/docs/src/html/HtmlDocView.test.tsx
+++ b/packages/docs/src/html/HtmlDocView.test.tsx
@@ -80,8 +80,8 @@ describe('resolveOctoDocBase / buildOctoDocUrl', () => {
     expect(resolveOctoDocBase()).toBe('https://octo-doc.example.com')
   })
 
-  it('defaults to same-origin (empty base) when nothing is configured', () => {
-    expect(resolveOctoDocBase()).toBe('')
+  it('defaults to the same-origin /docs-html unified prefix when nothing is configured', () => {
+    expect(resolveOctoDocBase()).toBe('/docs-html')
   })
 
   it('builds the octo-doc read-only URL `<base>/d/{slug}/v/{version}`', () => {
@@ -232,7 +232,7 @@ describe('HtmlDocView — read-only rendering', () => {
     })
     const { container } = render(<HtmlDocView docId="d1" space="sp" slug="published-slug" version="v7" />)
     await waitForFrame(container)
-    expect(String(spy.mock.calls[0][0])).toBe('/d/published-slug/v/v7')
+    expect(String(spy.mock.calls[0][0])).toBe('/docs-html/d/published-slug/v/v7')
   })
 
   it('shows a loading state before the fetch resolves', async () => {
--- a/packages/docs/src/html/HtmlDocView.test.tsx
+++ b/packages/docs/src/html/HtmlDocView.test.tsx
@@ -109,6 +109,29 @@ describe('absolutizeDocAssetUrls', () => {
     expect(out).toContain('src="https://od.test/d/slug/assets/b.png?exp=9"')
   })
 
+  it('re-roots root-relative /d/ octo-doc assets under the same-origin /docs-html prefix (DEFAULT deploy)', () => {
+    // DEFAULT deploy: no override, resolveOctoDocBase() === '/docs-html'. The doc backend emits
+    // root-relative refs like /d/{slug}/assets/{sha}; without re-rooting they resolve against the
+    // page origin, DROP the /docs-html prefix, and 404 (the nginx only proxies /docs-html/*).
+    expect(resolveOctoDocBase()).toBe('/docs-html')
+    const out = absolutizeDocAssetUrls(
+      '<!doctype html><html><body><img src="/d/slug/assets/a.png?sig=s1&exp=9"></body></html>',
+      // same-origin default docUrl form: {origin}/docs-html/d/{slug}/v/{ver}
+      'http://localhost/docs-html/d/slug/v/latest'
+    )
+    expect(out).toContain('src="http://localhost/docs-html/d/slug/assets/a.png?sig=s1&amp;exp=9"')
+    expect(out).not.toContain('src="http://localhost/d/slug/assets/a.png')
+  })
+
+  it('does not double-prefix an asset already under /docs-html/d/', () => {
+    const out = absolutizeDocAssetUrls(
+      '<!doctype html><html><body><img src="/docs-html/d/slug/assets/a.png"></body></html>',
+      'http://localhost/docs-html/d/slug/v/latest'
+    )
+    expect(out).toContain('src="http://localhost/docs-html/d/slug/assets/a.png"')
+    expect(out).not.toContain('/docs-html/docs-html/')
+  })
+
   it('leaves already absolute asset URLs and ordinary relative links untouched', () => {
     const out = absolutizeDocAssetUrls(
       '<html><head><link href="https://cdn.test/d/slug/assets/doc.css"></head><body><img src="/other/image.png"><a href="chapter.html">next</a></body></html>',
--- a/packages/docs/src/html/HtmlDocView.tsx
+++ b/packages/docs/src/html/HtmlDocView.tsx
@@ -189,9 +189,11 @@ export interface HtmlDocViewProps {
  *   1. `window.__OCTO_DOC_BASE__` — runtime injection (host config / index.html), so the
  *      same bundle points at different octo-doc origins per environment without a rebuild.
  *   2. `import.meta.env.VITE_OCTO_DOC_BASE` — build-time override.
- *   3. Empty string — resolve RELATIVE to the page origin (i.e. octo-doc reverse-proxied
- *      under the same host). This is a safe default; a deployment where octo-doc lives
- *      elsewhere must set one of the two overrides above.
+ *   3. Default `/docs-html` — a same-origin unified prefix. All web→octo-doc traffic
+ *      (render `/d/…`, comments, reactions, grants, docs-admin, versions) is namespaced
+ *      under this one prefix so it is easy to govern and cannot collide with SPA or other
+ *      service routes. The web nginx reverse-proxies `/docs-html/*` to octo-doc (stripping
+ *      the prefix). A deployment where octo-doc lives elsewhere sets one of the overrides above.
  */
 export function resolveOctoDocBase(): string {
   const runtime =
@@ -202,8 +204,8 @@ export function resolveOctoDocBase(): string {
       ? (import.meta as unknown as { env?: { VITE_OCTO_DOC_BASE?: string } }).env?.VITE_OCTO_DOC_BASE
       : undefined
   if (typeof env === 'string' && env.trim()) return env.trim().replace(/\/+$/, '')
-  // Same-origin default: octo-doc proxied under the current host.
-  return ''
+  // Same-origin unified prefix: octo-doc reverse-proxied under /docs-html (see web nginx).
+  return '/docs-html'
 }
 
 /** Build the octo-doc read-only render URL: `<base>/d/{slug}/v/{version}`. */
--- a/packages/docs/src/html/HtmlDocView.tsx
+++ b/packages/docs/src/html/HtmlDocView.tsx
@@ -78,21 +78,32 @@ function isAbsoluteOrSpecialUrl(value: string): boolean {
   return /^[a-zA-Z][a-zA-Z\d+.-]*:/.test(value) || value.startsWith('//') || value.startsWith('#')
 }
 
-function resolveDocAssetUrl(value: string, docUrl: string): string | null {
+// basePrefix is the octo-doc same-origin path prefix (resolveOctoDocBase, e.g. '/docs-html'),
+// empty for cross-origin/override deployments. A root-relative octo-doc asset ref like
+// `/d/{slug}/assets/{sha}` (the form the doc backend emits, see signAssetURLs) resolves
+// against the PAGE ORIGIN and would DROP the prefix, hitting a path the nginx no longer
+// proxies — re-root it under basePrefix first so it stays inside /docs-html/*. Only re-root
+// when docUrl itself sits under basePrefix (i.e. the same-origin prefixed deploy); an
+// absolute/override docUrl already carries the doc origin and must be left alone.
+function resolveDocAssetUrl(value: string, docUrl: string, basePrefix = ''): string | null {
   if (!value || isAbsoluteOrSpecialUrl(value)) return null
   try {
-    const url = new URL(value, docUrl)
+    const docPath = new URL(docUrl).pathname
+    const underPrefix = !!basePrefix && (docPath === basePrefix || docPath.startsWith(basePrefix + '/'))
+    const rebased =
+      underPrefix && value.startsWith('/d/') && !value.startsWith(basePrefix + '/') ? basePrefix + value : value
+    const url = new URL(rebased, docUrl)
     return /\/assets\//.test(url.pathname) ? url.href : null
   } catch {
     return null
   }
 }
 
-function absolutizeAssetAttr(el: Element, attr: 'src' | 'href', docUrl: string) {
+function absolutizeAssetAttr(el: Element, attr: 'src' | 'href', docUrl: string, basePrefix = '') {
   const raw = el.getAttribute(attr)
   if (!raw) return
   const value = raw.trim()
-  const resolved = resolveDocAssetUrl(value, docUrl)
+  const resolved = resolveDocAssetUrl(value, docUrl, basePrefix)
   if (resolved) el.setAttribute(attr, resolved)
 }
 
@@ -132,9 +143,13 @@ export function resolveHtmlDocAnchorText(
 export function absolutizeDocAssetUrls(html: string, docUrl = resolveAbsoluteOctoDocBase()): string {
   if (typeof DOMParser === 'undefined') return html
   const absoluteDocUrl = resolveAbsoluteUrl(docUrl)
+  // Same-origin path prefix (e.g. '/docs-html'); '' for absolute/cross-origin bases. Used to
+  // re-root the backend's root-relative `/d/...` asset refs so they keep the prefix.
+  const base = resolveOctoDocBase()
+  const basePrefix = base.startsWith('/') ? base.replace(/\/+$/, '') : ''
   const doc = new DOMParser().parseFromString(html, 'text/html')
-  doc.querySelectorAll('img[src]').forEach((el) => absolutizeAssetAttr(el, 'src', absoluteDocUrl))
-  doc.querySelectorAll('link[href]').forEach((el) => absolutizeAssetAttr(el, 'href', absoluteDocUrl))
+  doc.querySelectorAll('img[src]').forEach((el) => absolutizeAssetAttr(el, 'src', absoluteDocUrl, basePrefix))
+  doc.querySelectorAll('link[href]').forEach((el) => absolutizeAssetAttr(el, 'href', absoluteDocUrl, basePrefix))
   neutralizeEditableControls(doc)
   const doctype = doc.doctype ? `<!doctype ${doc.doctype.name}>` : ''
   return `${doctype}${doc.documentElement.outerHTML}`


Jerry-Xin (dismissed):
The PR is repository-relevant, but the unified proxy introduces two blocking URL-handling regressions.

### 🔴 Blocking

- 🔴 Critical — Root-relative document assets bypass the new proxy prefix. `HtmlDocView` rewrites assets such as `/d/slug/assets/a.png` to the web origin’s `/d/...` path ([HtmlDocView.tsx](/Users/mlamp/.openclaw/tmp/pr-review-sw0__czi/repo/packages/docs/src/html/HtmlDocView.tsx:81)), while this PR removes the `/d/` nginx location and only proxies `/docs-html/` ([nginx.conf.template](/Users/mlamp/.openclaw/tmp/pr-review-sw0__czi/repo/nginx.conf.template:118)). These resources will now fall through to the SPA and return `index.html`, breaking images and styles. Rebase root-relative octo-doc assets to `/docs-html/d/...`, or otherwise preserve routing for them. Add a default-base integration/unit test—the existing test explicitly documents this root-relative asset form.

- 🔴 Critical — The new `rewrite` rules do not preserve encoded path components verbatim. Nginx rewrite matching operates on the normalized, percent-decoded URI, so values produced by `encodeURIComponent`, particularly `%2F`, can become path separators before forwarding ([nginx.conf.template](/Users/mlamp/.openclaw/tmp/pr-review-sw0__czi/repo/nginx.conf.template:123)). This can misroute render, grants, delete, and version requests. The repository explicitly tests slugs such as `a/b?c` being encoded as `a%2Fb%3Fc` ([htmlDocVersions.test.ts](/Users/mlamp/.openclaw/tmp/pr-review-sw0__czi/repo/packages/docs/src/html/htmlDocVersions.test.ts:49)). Preserve the raw escaped path during prefix translation and add nginx-level coverage for encoded slugs and versions.

### ✅ Highlights

- Restoring the SPA `try_files` fallback is correct.
- Appending `$request_uri` directly for `/api/v1/docs` correctly preserves its query string.
- Consolidating octo-doc traffic under a dedicated prefix is a sound architectural direction.

Jerry-Xin (changes requested):
The PR is repository-relevant, but the unified proxy introduces two blocking URL-handling regressions.

### 🔴 Blocking

- 🔴 Critical — Root-relative document assets bypass the new proxy prefix. `HtmlDocView` absolutizes octo-doc render assets such as `/d/slug/assets/a.png` against the doc base, but a root-relative `/d/...` value resolves to the web origin's `/d/...` path (dropping the new `/docs-html` prefix) — see `packages/docs/src/html/HtmlDocView.tsx` (`resolveDocAssetUrl` / `resolveAbsoluteOctoDocBase`, ~line 81). This PR removes the `/d/` nginx location and only proxies `/docs-html/` (`nginx.conf.template`, the `location /docs-html/` block). Those asset requests will now fall through to the SPA `try_files` and return `index.html`, breaking images and styles on rendered docs. Fix: rebase root-relative octo-doc assets to `/docs-html/d/...`, or otherwise preserve routing for them. Add a default-base unit test — the existing test `packages/docs/src/html/HtmlDocView.test.tsx` explicitly documents this root-relative `/d/slug/assets/a.png` form.

- 🔴 Critical — The new `rewrite` rules do not preserve encoded path components verbatim. Nginx rewrite matching operates on the normalized, percent-decoded URI, so values produced by `encodeURIComponent` — particularly `%2F` — can be decoded to path separators before forwarding (`nginx.conf.template`, the `/docs-html/docs/...` and `/docs-html/grants/...` rewrites). This can misroute render, grants, delete, and version requests. The repo explicitly tests slugs such as `a/b?c` encoded as `a%2Fb%3Fc` in `packages/docs/src/html/htmlDocVersions.test.ts`. Fix: preserve the raw escaped path during prefix translation, and add nginx-level coverage for encoded slugs and versions.

### ✅ Highlights

- Restoring the SPA `try_files` fallback is correct.
- Appending `$request_uri` directly for `/api/v1/docs` correctly preserves its query string (fixing the prior `%3F` encoding regression).
- Consolidating octo-doc traffic under a single dedicated `/docs-html/*` prefix is a sound architectural direction.

lml2468 (approved):
## 🐙 PR #888 Review — `fix(nginx): repair octo-doc proxy query encoding, restore SPA fallback + unify /docs-html/* prefix`

**Head:** `291d7a780c39` · **Base:** `main` · +45 −72 / 3 files · `mergeable: MERGEABLE`
**Intent:** Fix three regressions #846 introduced in the web→octo-doc nginx proxy, and consolidate all octo-doc traffic under one `/docs-html/*` prefix: (1) variable `proxy_pass` + `rewrite ^ $request_uri` URL-encoded `?`→`%3F` (broke `/api/v1/docs?owner=me`, `/d/?code=`); (2) `location /` lost its SPA `try_files` fallback (deep-links/refresh 404'd); (3) five scattered top-level octo-doc locations replaced by a single governable `/docs-html/` prefix driven by one frontend seam.

### Verification (local checkout @ `291d7a780c39`, pnpm installed)
| Check | Result |
|---|---|
| `vitest run src/html/HtmlDocView.test.tsx` | ✅ 46/46 (2 updated `/docs-html` expectations) |
| nginx template structural validity | ✅ brace balance 31/31; locations/rewrites well-formed (author ran `nginx -t` post-envsubst) |
| single-seam prefix migration | ✅ all 5 octo-doc sub-path callers go through `resolveOctoDocBase()` |

Delivers exactly the three described fixes plus the prefix unification, no scope creep. Runtime/build overrides (`__OCTO_DOC_BASE__`/`VITE_OCTO_DOC_BASE`) still take precedence; default changes `'' → '/docs-html'`.

### Gate 2 — Code quality & correctness: ✅ (infra/proxy surface — nginx semantics checked closely)
- **Query-encoding fix is correct.** `/api/v1/docs` now uses `proxy_pass $docs_backend_url$request_uri;` — appending raw `$request_uri` (unencoded path+query) to a **variable** upstream forwards verbatim, avoiding the `rewrite ^ $request_uri break` double-encoding that turned `?`→`%3F`. Matches pre-#846 behavior.
- **`/docs-html/` rewrites preserve the query.** The block uses bare variable `proxy_pass $doc_app_url;`, so the URI **as rewritten** is forwarded. nginx `rewrite` preserves the query string by default (none of these replacements contain `?`), so `/docs-html/d/{slug}?code=X` → `/d/{slug}?code=X` — the `?code=` case is fixed. Verified against the same variable-proxy_pass semantics as the earlier #846 review.
- **Rewrite ordering is safe.** Anchored regexes, each ending in `break`, ordered most-specific-first. I checked the one real trap: `docs-admin/{slug}` does NOT match `^/docs-html/docs/(.+)` (the rule requires literal `/docs/` with slash; `docs-admin/` has no slash after `docs`), and the `docs-admin` rule precedes `docs/` anyway → no misrouting. grants uid-form precedes bare-form (correct). Mapping to octo-doc native routes (`/v1/comments`, `/v1/docs/{slug}/grants`, etc.) is faithful to the deleted per-location rewrites.
- **Single-seam migration is complete — the load-bearing cross-file check.** Changing `resolveOctoDocBase()` default to `/docs-html` prefixes ALL octo-doc HTML callers because every one routes through it: `htmlDocComments.ts` (`/comments`), `htmlGrantsApi.ts` (`/grants/{slug}`), `htmlDocAdmin.ts` (`/docs-admin/{slug}`), `htmlDocVersions.ts` (`/docs/{slug}/versions`), and `buildOctoDocUrl` (`/d/…`). No caller is left on a now-deleted top-level path → no orphaned 404. (The `versions/api.ts` `/docs/:docId/versions` calls are the *separate* Yjs `apiClient` backend, correctly untouched.)
- **SPA fallback restored correctly.** `try_files $uri $uri/ /index.html;` in `location /`; nginx longest-prefix-match keeps `/docs-html/`, `/api/…`, `/v1/` ahead of `/`, so the fallback only catches genuinely-unmatched client routes — no shadowing of the proxy locations.
- **Safety modes preserved:** each proxied location keeps `resolver` + the unset-upstream `503` guard, so a missing DOC_APP_URL/DOCS_BACKEND_URL 503s the route instead of breaking nginx startup.

### Non-blocking notes
- **Deploy-ordering / rollout:** this couples frontend (now emits `/docs-html/*`) and nginx (now serves `/docs-html/*`, old top-level paths removed) — they must ship together in the same image. A frontend built with the new default against an OLD nginx (no `/docs-html/` block) would 404 all octo-doc calls, and vice-versa. Same-image deploy makes this a non-issue, worth stating for staged rollouts.
- **Human smoke (author did this, recommended to a reviewer too):** `/api/v1/docs?owner=me` reaches backend (401 not 404), `/d/<slug>?code=` verbatim, SPA deep-link+refresh 200, `/docs-html/*` render+comment POST 200 — the live proxy behavior can't be unit-verified.

query preserved through variable proxy_pass + rewrite (no `?`→`%3F`) ✅ · `$request_uri` append fix correct ✅ · rewrite ordering (docs-admin vs docs, grants uid-vs-bare) safe ✅ · single-seam prefixes all 5 callers, no orphaned path ✅ · SPA fallback doesn't shadow proxy locations ✅ · resolver+503 safety kept ✅ · build + 46/46 ✅.

---

### ✅ Review Verdict: APPROVE @ `291d7a780c39`
Spec ✅ and quality ✅ — a correct, well-reasoned infra fix: the `$request_uri`-append and variable-proxy_pass+rewrite both forward path+query verbatim (repairing #846's `?`→`%3F` and the `/d/?code=` drop), the SPA fallback is restored without shadowing proxy locations, and the `/docs-html/*` consolidation is a clean single-seam migration where every octo-doc caller already routes through `resolveOctoDocBase()` (no orphaned top-level path). Build green, HtmlDocView 46/46, nginx structurally valid. Non-blocking: frontend+nginx must deploy in the same image (coupled prefix), and a live proxy smoke. No blockers.

lml2468:
### 🐙 Follow-up on my APPROVE — Jerry-Xin's two 🔴 verified: one refuted, one is a real 🟡 (not a blocker). APPROVE stands.

Both are nginx-semantics claims worth checking carefully — I traced each:

**🔴#1 "root-path assets (images/styles) not proxied" → refuted (not a defect).** The octo-doc render isn't a page the browser navigates to for assets — `HtmlDocView` **fetches** the HTML (JS `fetch`, :334) and injects it after asset-correcting it: `absolutizeDocAssetUrls(html, url)` rewrites asset attrs to absolute, and `injectBaseHref(..., resolveAbsoluteOctoDocBase())` injects a `<base href>` so CSS `url()` / relative / root-path resources resolve to the **real octo-doc origin**, not `/docs-html/` or the SPA. That machinery is **pre-existing and untouched by this PR** (diff only changes the `resolveOctoDocBase` default + comments). So assets never depended on nginx proxying `/assets/` under the prefix — no breakage. (The one direct navigation, `window.open(docUrl)` → `/docs-html/d/…`, is caught by the render rewrite.)

**🔴#2 "rewrite doesn't preserve encoded paths" → real nginx subtlety, but 🟡 (unreachable today), not 🔴.** Correct observation in principle: the old `/d/` block used `rewrite ^ $request_uri break` (forwards the **raw, encoding-preserved** URI); the new `rewrite ^/docs-html/(d/.*)$ /$1` captures from **`$uri`**, which nginx has already percent-**decoded** — so a slug/version containing `%XX` would reach octo-doc decoded/normalized. That's a genuine encoding-fidelity narrowing vs the old raw passthrough.

**Why it's 🟡 not 🔴 — no reachable slug triggers it:** octo-doc slugs are `documentName` segments, and the schema is `SEGMENT = /^[A-Za-z0-9_-]+$/` (`documentName/index.ts:19`) — none of those chars are escaped by the `encodeURIComponent(slug)` in `buildOctoDocUrl`, so `%XX` never appears in a real `/d/{slug}/v/{version}` path; the decoded-vs-raw distinction produces **identical bytes** for every valid slug. Version is `latest`/`vN`, likewise. So there's no reachable corruption today. It's worth hardening (a future slug-charset expansion could expose it) — either forward `d/` via the raw-URI pattern (`proxy_pass $doc_app_url$request_uri` on a dedicated match, like the `/api/v1/docs` fix) or keep the query-fix but note the constraint — but it's a defense-in-depth follow-up, not a merge blocker.

**Note:** the `?`-query preservation this PR *does* fix works because nginx `rewrite` preserves the query string by default (no `?` in any replacement); that's separate from the path-**encoding** point above and is correct.

**🟡 (test coverage)** — agreed a unit/integration test locking the `/docs-html` default + an encoded-path case would be a good add.

**Verdict unchanged: APPROVE @ `291d7a780c39`** — #1 doesn't hold (asset base-href handles it), #2 is a real but currently-unreachable encoding-fidelity 🟡 worth a hardening follow-up. If a slug charset ever widens beyond `[A-Za-z0-9_-]`, #2 becomes reachable and should be fixed then.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review: octo-web PR#888 — fix(nginx): repair octo-doc proxy query encoding, restore SPA fallback + unify /docs-html/* prefix
**Head SHA:** `291d7a780c39d33f74f938844308af9e687433be`
**Merge base:** `6208748cdade0d9ee0b6a04ba0ecefdc3de718eb`
**Files changed:** 3 (nginx.conf.template, HtmlDocView.tsx, HtmlDocView.test.tsx) — 45 insertions, 72 deletions

---

## 1. Verification Summary

| Check | Result |
|-------|--------|
| nginx rewrite ordering in `/docs-html/` | ✅ Correct — most-specific-first, all end in `break` |
| Query encoding fix (`/api/v1/docs`) | ✅ Correct — `$request_uri` appended directly to variable proxy_pass |
| SPA fallback `try_files` | ✅ Standard pattern, safe (all API routes have longer-prefix locations) |
| Frontend `resolveOctoDocBase()` default change `''` → `'/docs-html'` | ✅ Correct — matches new nginx location block |
| Test coverage updated | ✅ Two tests updated to reflect new default |
| All caller files route through `resolveOctoDocBase()` | ✅ Verified: htmlDocComments.ts, htmlGrantsApi.ts, htmlDocAdmin.ts, htmlDocVersions.ts, HtmlVersionPanel.tsx |

## 2. Data-Flow Tracing

Every consumer of `resolveOctoDocBase()` was traced from frontend URL construction through nginx rewrite to octo-doc upstream:

| Frontend path (default base=`/docs-html`) | nginx rewrite | Upstream receives |
|----|----|----|
| `/docs-html/d/{slug}/v/{ver}` | `^/docs-html/(d/.*)$` → `/$1` | `/d/{slug}/v/{ver}` ✅ |
| `/docs-html/comments?slug&ver` | `^/docs-html/comments(.*)$` → `/v1/comments$1` | `/v1/comments?slug&ver` ✅ |
| `/docs-html/reactions…` | `^/docs-html/reactions(.*)$` → `/v1/reactions$1` | `/v1/reactions…` ✅ |
| `/docs-html/grants/{slug}` | `^/docs-html/grants/([^/]+)$` → `/v1/docs/$1/grants` | `/v1/docs/{slug}/grants` ✅ |
| `/docs-html/grants/{slug}/{uid}` | `^/docs-html/grants/([^/]+)/(.+)$` → `/v1/docs/$1/grants/$2` | `/v1/docs/{slug}/grants/{uid}` ✅ |
| `/docs-html/docs-admin/{slug}` | `^/docs-html/docs-admin/(.+)$` → `/v1/docs/$1` | `/v1/docs/{slug}` ✅ |
| `/docs-html/docs/{slug}/versions` | `^/docs-html/docs/(.+)$` → `/v1/docs/$1` | `/v1/docs/{slug}/versions` ✅ |

Query string handling: nginx `rewrite` matches against `$uri` (path only, no query), so `$args` is automatically preserved and appended to the rewritten URI. All paths above correctly carry query parameters to the upstream. Verified.

## 3. Findings

### No P0/P1 issues found.

### P2-1 — Catch-all rewrite may forward unintended paths (maintainability)

**Location:** `nginx.conf.template`, line ~131 (`rewrite ^/docs-html/(.*)$ /$1 break;`)

**Diff-scope:** New (introduced by this PR's consolidation).

The catch-all rewrite `^/docs-html/(.*)$` → `/$1` is forward-looking: any future frontend path under `/docs-html/` that doesn't match a specific rewrite will be forwarded to octo-doc as-is (prefix stripped). If a future change accidentally adds a frontend URL under `/docs-html/` that should NOT reach octo-doc (e.g., a new SPA route mistakenly namespaced under the prefix), it would silently proxy to octo-doc instead of falling through to the SPA.

**Risk:** Low. The `/docs-html` prefix is exclusively used by `resolveOctoDocBase()` consumers, and the catch-all provides useful extensibility. This is a defense-in-depth note, not a current defect.

### P2-2 — Comment on query encoding fix slightly imprecise (nit)

**Location:** `nginx.conf.template`, lines ~155–157

The comment says `rewrite ^ $request_uri break` would "URL-encode the `?` into `%3F`". The actual issue is more nuanced: in most nginx versions, `rewrite` recognizes `?` as the args separator and does NOT encode it. The real problem with the old pattern is that combining `rewrite` with a bare variable `proxy_pass` can produce unpredictable URI forwarding depending on nginx version and internal state. The fix (`proxy_pass $var$request_uri`) is correct and the recommended pattern — the comment could be more precise about the root cause.

**Risk:** None (cosmetic/documentation).

## 4. Suggestions

1. **Consider adding a comment on the `/docs-html/` location's trailing-slash behavior.** `location /docs-html/` (with trailing slash) does NOT match a bare `/docs-html` request. This is fine for the current use case (frontend always builds `/docs-html/...` paths), but a brief comment would help future maintainers.

2. **The `resolveAbsoluteOctoDocBase()` function** (HtmlDocView.tsx:70) now produces `http://origin/docs-html` as the `<base href>` for the iframe (via `injectBaseHref`). This means CSS-referenced relative assets in octo-doc HTML will resolve against `/docs-html/` instead of `/`. For absolute CSS paths (e.g., `url('/assets/bg.png')`), this makes no difference. For relative CSS paths, the new base actually routes them through the nginx proxy to octo-doc, which may be an improvement. Worth verifying against real octo-doc HTML in staging.

## 5. Blindspot Checklist (C1–C6)

| Check | Result | Evidence |
|-------|--------|----------|
| C1 — Dual path parity | N/A | No paired add/remove or symmetric operations changed |
| C2 — Control-flow ordering | ✅ Clear | Rewrite ordering verified: grants-uid > grants-bare > docs-admin > docs > comments > reactions > render > catch-all. No overlap or shadowing. |
| C3 — Authorization boundary | N/A | Auth model (token header passthrough) unchanged |
| C4 — Authorization lifecycle | N/A | No auth changes |
| C5 — Build/runtime parity | N/A | nginx template + TypeScript function, no build artifacts to verify |
| C6 — Governance/security docs | N/A | No policy/document changes |

## 6. Cross-Round Blocker Re-check (R6)

## 7. Extra Notes

- **Deploy coordination:** This PR requires simultaneous deployment of the new nginx config and the new frontend bundle. If the old frontend runs against the new nginx (or vice versa), all octo-doc requests will 404/serve-index.html. This is expected for tightly-coupled config+code changes and should be noted in the deploy plan.
- **No `/reactions` data layer exists** in the frontend codebase — the nginx rewrite for `/docs-html/reactions` is forward-looking. Not a bug, just noting the asymmetry.
- **Test isolation:** All integration tests set `window.__OCTO_DOC_BASE__ = 'https://od.test'`, bypassing the default. Only two unit tests exercise the default path, and both are updated. ✅

---

**[Octo-Q] verdict: APPROVE** — No P0/P1 defects found. The consolidation is well-designed, the query encoding fix is correct, and the SPA fallback is standard. Two P2 notes on maintainability.

lml2468 (changes requested):
## 🐙 PR #888 Re-Review (correction) — REQUEST_CHANGES @ `291d7a780c39`

**I'm correcting my APPROVE — and my earlier "refuted #1" comment was wrong.** Steve and Jerry-Xin flagged root-relative asset breakage; I re-checked with an **executed probe** (not static reasoning) and it's a **real, reachable regression in the default same-origin deployment**. I owe the correction precisely.

### 🔴 P2 (BLOCKER, CONFIRMED) — root-relative octo-doc assets break under the new `/docs-html` default

The render doc (iframe `srcdoc`) contains root-relative asset refs like `/d/slug/assets/a.png` — the repo's own test asserts octo-doc emits these (`HtmlDocView.test.tsx:96`). The asset-resolution I claimed "sends them to the real origin" does **not** save them under the new default:

**Executed probe** (real `absolutizeDocAssetUrls` with `resolveOctoDocBase()='/docs-html'`, jsdom origin `http://localhost:3000`):
```
BASE = "/docs-html"
input img src = /d/slug/assets/a.png?sig=s
absolutized   = http://localhost:3000/d/slug/assets/a.png?sig=s   ← /docs-html PREFIX DROPPED
```
`resolveDocAssetUrl` does `new URL('/d/slug/assets/a.png', <origin>/docs-html)` — a **root-relative** path resolves against the **origin**, discarding the `/docs-html` base path → `<origin>/d/slug/assets/...`. That path is exactly the `/d/` nginx location **this PR deletes** → falls through to `location /` `try_files` → **index.html** → the doc's images/CSS 404 as HTML. Reachable in the **default same-origin deployment this PR makes default**, for any doc with a root-relative `/d/.../assets/` reference.

**`<base href>` does NOT rescue it** (I checked): `injectBaseHref` injects `<base href="http://localhost:3000/docs-html/">`, but per the HTML spec `<base>` only affects **relative** URLs — a **root-absolute** `/d/...` ignores `<base>` and resolves against origin. And `absolutizeDocAssetUrls` has already rewritten it to the prefix-less absolute form anyway. Both mechanisms miss this case.

**Why my APPROVE was wrong:** I asserted "absolutize resolves assets to the real octo-doc origin" from reading the code, without running it for the *same-origin `/docs-html`* case — where "real origin" is the page origin and the `/docs-html` prefix is dropped for root-relative paths. That's the exact reasoning-vs-executing gap; the probe shows the prefix loss plainly.

**Fix options:** (a) make `absolutizeDocAssetUrls`/asset resolution prefix root-relative `/d/…assets` with `/docs-html` (resolve against `<origin>/docs-html/`, not origin) so they hit the new proxy; or (b) keep a nginx `/d/` render/asset location (or a `/d/.../assets/` proxy) so root-relative asset paths still reach octo-doc; or (c) have octo-doc emit `/docs-html`-relative asset paths. Plus a test: default-base render with a root-relative `/d/.../assets` asset → asserts the resolved URL is proxied, not SPA-swallowed.

### 🟡 (Claim #2, encoded-slug) — real but narrower; I was right it's 🟡, wrong that it's unreachable
`htmlDocVersions.test.ts:51` DOES test `listVersions('a/b?c')` → `.../docs/a%2Fb%3Fc/versions`, so encoded slugs (`%2F`/`%3F`) are a real emitted case on the versions/grants/admin paths — my "slugs are `[A-Za-z0-9_-]` only" was from `documentName` and doesn't govern those. The nginx `rewrite` matches the **decoded** `$uri`, so `%2F`→`/` could re-split the path. This is a genuine encoding-fidelity concern on those sub-paths; grade it alongside #1 for the same rewrite-encoding fix. (Still likely rarer than #1 in practice, but I withdraw the "unreachable" framing — the test proves the frontend emits it.)

### What still stands
The two things the PR gets right remain right: the SPA `try_files` restoration, and the `/api/v1/docs` `$request_uri` query fix (that one appends raw request_uri, so it's encoding-safe). The architecture direction (single `/docs-html/*` prefix) is sound. The blocker is the asset-path + encoded-slug resolution under the new default.

---

### ❌ Review Verdict: REQUEST_CHANGES @ `291d7a780c39`
Correcting my APPROVE: a confirmed, byte-proven regression — under the new same-origin `/docs-html` default, root-relative octo-doc assets (`/d/.../assets/`, which the render emits per the repo's own test) absolutize to `<origin>/d/...` (prefix dropped, `<base>` doesn't apply to root-absolute), hitting the `/d/` location this PR deletes → SPA fallback → broken doc images/CSS. Plus the encoded-slug (`%2F`/`%3F`) rewrite-decoding concern on versions/grants/admin (repo tests emit these). Fix the asset/slug resolution to carry the `/docs-html` prefix (or keep a `/d/` asset proxy) + add default-base regression tests. Credit to Steve and Jerry-Xin — my APPROVE and my "refuted #1" comment both relied on static reading; the executed probe shows the prefix loss.

yujiawei (approved):
# Code Review — PR #888 (octo-web)

**Verdict: APPROVED** — no blocking issues. Three low-severity advisory notes below; none must be fixed before merge.

## Summary

This PR fixes three regressions from #846 in the web→octo-doc nginx proxy and consolidates the routing:

1. **Query-encoding fix** (`/api/v1/docs`): switches from `rewrite ^ $request_uri break; proxy_pass $docs_backend_url;` to `proxy_pass $docs_backend_url$request_uri;`. The old form let nginx URL-encode `?` into `%3F` on the variable upstream, so `?owner=me` broke the docs list. Verified correct.
2. **SPA fallback**: restores `try_files $uri $uri/ /index.html;` on `location /`, which the #846 template dropped — client routes / deep links / refresh no longer hard-404.
3. **Unified prefix**: collapses five scattered top-level locations (`/d/`, `/comments`, `/reactions`, `/grants/`, `/docs-admin/`) into one `location /docs-html/` with ordered rewrites, and flips the frontend `resolveOctoDocBase()` default from `''` to `/docs-html`.

## Spec compliance

- **Nothing missing**: all six frontend octo-doc call sites were cross-checked against the new nginx rewrites and each maps to the intended native route:
  - `buildOctoDocUrl` → `/docs-html/d/{slug}/v/{ver}` → `/d/{slug}/v/{ver}` (HtmlDocView.tsx:214)
  - `commentsUrl` → `/docs-html/comments?slug=..&version=..` → `/v1/comments?…` (htmlDocComments.ts:102)
  - `grantsUrl` → `/docs-html/grants/{slug}[/{uid}]` → `/v1/docs/{slug}/grants[/{uid}]` (htmlGrantsApi.ts:29)
  - `deleteDoc` → `/docs-html/docs-admin/{slug}` → `/v1/docs/{slug}` (htmlDocAdmin.ts:19)
  - `listVersions` → `/docs-html/docs/{slug}/versions` → `/v1/docs/{slug}/versions` (htmlDocVersions.ts:26)
- **Nothing extra**: the `reactions` rewrite is retained for parity but has no current frontend caller — harmless, and consistent with the "one prefix for all octo-doc traffic" goal.
- Rewrite ordering is correct: the `grants` (`[^/]+`) and `docs-admin` rules precede the greedy `docs/(.+)` versions rule, and the specific `d/` render rule precedes the catch-all. No frontend path is captured by the wrong rewrite.
- The two affected unit tests were updated to the new `/docs-html` default; the other html data-layer tests set an explicit `__OCTO_DOC_BASE__`, so they are unaffected.

## Code quality — advisory notes (non-blocking)

- **P2 — `nginx.conf.template:159`, raw `$request_uri` forwarding.** Appending the raw request URI (rather than nginx's normalized `$uri$is_args$args`) forwards un-normalized paths (e.g. `..`) and would double a trailing slash if `DOCS_BACKEND_URL` ever ends in `/`. Worth noting this is **not introduced here** — the pre-#888 block already used `rewrite ^ $request_uri break`, so this change is behavior-preserving and is the correct fix for the `%3F` bug. Normalizing is a reasonable follow-up hardening, but path-traversal defense properly belongs at the docs backend, and the deploy sets `DOCS_BACKEND_URL` without a trailing slash.

- **P2 — `nginx.conf.template:126,128`, greedy `(.+)` slug capture.** `docs-admin` and `versions` use `(.+)` for the slug where `grants` uses `([^/]+)`. In principle this lets extra path segments through. In practice it is unreachable: slugs are constrained to `[A-Za-z0-9_-]+` at the source (`StandaloneDocPage.tsx:26`) and every call site passes them through `encodeURIComponent`, so a slash (or `%2F`) can never appear in a real request. Tightening to `([^/]+)` for symmetry would be a nice-to-have, not a fix.

- **P2 — `nginx.conf.template:133`, redundant `d/` rewrite.** `rewrite ^/docs-html/(d/.*)$ /$1 break;` is fully eclipsed by the following catch-all `rewrite ^/docs-html/(.*)$ /$1 break;` — same output for any `d/…` input. It documents intent but is dead. Safe to leave or drop.

## Verification performed

- Traced all frontend octo-doc URL builders against the nginx rewrites (path-by-path).
- Confirmed no consumer outside `packages/docs/src/html/` calls the removed top-level paths.
- Confirmed the `/api/v1/docs` change is behavior-preserving vs. the pre-#888 template.
- Confirmed slug charset makes the greedy-capture concern unreachable.
- nginx runtime `-t` could not be run (binary unavailable in the review env); the template is syntactically consistent with the existing variable-proxy_pass blocks and rendered cleanly via `envsubst`.

yujiawei (changes requested):
# Code Review — PR #888 (octo-web) — CORRECTION

**Correcting my earlier Approve to CHANGES_REQUESTED.** My first pass verified the six frontend API call sites map to the new `/docs-html/*` rewrites, but I did not run the *asset-resolution* path for the new same-origin default. Re-checking with an executed probe surfaced a confirmed, reachable regression that other reviewers had already flagged. The evidence (a run, not a read) wins over my initial static trace.

## 🔴 P0 (CONFIRMED, blocking) — root-relative octo-doc assets break under the new `/docs-html` default

The rendered doc HTML (shown in the `srcdoc` iframe) contains root-relative asset refs like `/d/slug/assets/a.png` — the repo's own test asserts octo-doc emits this form (`packages/docs/src/html/HtmlDocView.test.tsx:96`). Under the new default base `/docs-html`, `absolutizeDocAssetUrls` → `resolveDocAssetUrl` resolves that asset with `new URL('/d/slug/assets/a.png', '<origin>/docs-html/d/slug/v/latest')`. Because the input is **root-relative**, it resolves against the **origin**, discarding the `/docs-html` prefix.

Executed probe (replicating `resolveDocAssetUrl` with `resolveOctoDocBase()='/docs-html'`, origin `http://localhost:3000`):

```
render url     = /docs-html/d/slug/v/latest
absoluteDocUrl = http://localhost:3000/docs-html/d/slug/v/latest
asset input    = /d/slug/assets/a.png
asset RESOLVED = http://localhost:3000/d/slug/assets/a.png   ← /docs-html prefix DROPPED
```

That resolved path (`<origin>/d/...`) is exactly the `/d/` nginx location **this PR deletes**. With only `location /docs-html/` proxying to octo-doc, the request falls through to `location /`'s `try_files … /index.html`, so every root-relative asset returns `index.html` (HTML) instead of the image/CSS. Result: rendered docs with root-relative `/d/.../assets/` refs show broken images and styles.

`<base href>` does **not** rescue this: `injectBaseHref` sets `<base href="…/docs-html/">`, but per spec `<base>` only affects *relative* URLs — a root-absolute `/d/...` ignores `<base>`. And `absolutizeDocAssetUrls` has already rewritten it to the prefix-less absolute form regardless.

**Reachability:** the default same-origin deployment this PR makes the default, for any doc whose render emits a root-relative `/d/.../assets/` reference (the repo tests this exact shape).

**Fix options:**
- (a) make asset resolution prefix root-relative `/d/…` assets with `/docs-html` (resolve against `<origin>/docs-html/`, not `<origin>`), so they hit the new proxy; or
- (b) keep a nginx location that proxies `/d/` (or at least `/d/.../assets/`) to octo-doc; or
- (c) have octo-doc emit `/docs-html`-relative asset paths.
- Plus a **default-base** regression test: render with a root-relative `/d/.../assets` asset and assert the resolved URL stays under the proxied prefix (not SPA-swallowed). The existing asset tests all pin an explicit `https://od.test` base, so they never exercise the `/docs-html` default and pass despite the bug.

## 🔴 P1 (test-proven, blocking) — encoded slug components are decoded before rewrite matching

`nginx rewrite` matches the **normalized, percent-decoded** URI, so `%2F` in a slug becomes a real `/` before the rewrite splits the path. The frontend does emit encoded slugs on the docs-meta sub-paths: `htmlDocVersions.test.ts:51` asserts `listVersions('a/b?c')` → `/docs/a%2Fb%3Fc/versions`, and the same `encodeURIComponent(slug)` pattern is used by `grantsUrl` (`htmlGrantsApi.ts:29`) and `deleteDoc` (`htmlDocAdmin.ts:19`). A slug containing `/` therefore re-splits under:

- `rewrite ^/docs-html/docs/(.+)$ /v1/docs/$1` (versions) — `docs/a%2Fb%3Fc/versions` → decoded `docs/a/b?c/versions` → `$1 = a/b?c/versions`, misrouting.
- `rewrite ^/docs-html/grants/([^/]+)/(.+)$` — a `%2F` in the slug re-splits into the `{slug}/{uid}` form even when no uid was intended.

(I withdraw my earlier "unreachable, slugs are `[A-Za-z0-9_-]` only" note — that regex governs the `/d/:docId` SPA route, not the octo-doc `slug` these builders encode; the repo test proves the frontend emits `%2F`/`%3F`.) Fix: preserve the raw escaped path during prefix translation, and add nginx-level coverage for encoded slugs/versions.

## ✅ What the PR gets right (unchanged)

- SPA `try_files` fallback restoration on `location /` — correct.
- `/api/v1/docs` → `proxy_pass $docs_backend_url$request_uri;` — correctly preserves the query string and fixes the prior `%3F` regression; behavior-preserving vs. the pre-#846-fix template (which already forwarded raw `$request_uri`).
- Single `/docs-html/*` prefix is a sound architectural direction; the six *API* call sites (comments, grants, docs-admin, versions, render page) map correctly to their rewrites.

## Minor / advisory (non-blocking, fix opportunistically with the above)

- `nginx.conf.template:133` — `rewrite ^/docs-html/(d/.*)$ /$1 break;` is fully eclipsed by the following catch-all `rewrite ^/docs-html/(.*)$ /$1 break;`; dead rule, safe to drop.
- `nginx.conf.template:126,128` — `docs-admin`/`versions` use greedy `(.+)` for the slug where `grants` uses `([^/]+)`; tightening to `([^/]+)` for symmetry also narrows the encoded-slug blast radius above.

## Verdict

The architecture and the two nginx fixes (SPA fallback, query encoding) are good, but the `/docs-html` default introduces a reachable asset-loading regression (P0) plus an encoded-slug rewrite-fidelity bug (P1) on the same rewrite path. Both need addressing — with default-base regression tests — before merge.

lml2468 (approved):
## 🐙 PR #888 Re-Review — `fix(nginx): … + unify /docs-html/* prefix`

**Head:** `10390603a6d6` (was `291d7a780c39`) · **Base:** `main` · **prior verdict: REQUEST_CHANGES** (root-relative asset breakage under the `/docs-html` default).

### 🔴 → ✅ The asset-breakage blocker is fixed — root-relative `/d/` assets re-rooted under `/docs-html`
New commit `10390603a fix(docs): re-root root-relative /d/ octo-doc assets under /docs-html`. Delta since my CR head = `HtmlDocView.tsx` (+27) + its test (+23):

`resolveDocAssetUrl` now takes a `basePrefix` (= `resolveOctoDocBase()`, e.g. `/docs-html`) and, when the docUrl sits under that prefix (the same-origin default deploy), re-roots a root-relative `/d/...` asset to `basePrefix + value` before absolutizing:
```ts
const underPrefix = !!basePrefix && (docPath === basePrefix || docPath.startsWith(basePrefix + '/'))
const rebased = underPrefix && value.startsWith('/d/') && !value.startsWith(basePrefix + '/') ? basePrefix + value : value
```
Guarded correctly: only under the prefixed same-origin deploy, only `/d/` root-relative refs, no double-prefix, absolute/override docUrls untouched.

**Byte-proven (executed probe, the same one that caught the bug), `resolveOctoDocBase()='/docs-html'`:**
```
/d/slug/assets/a.png  →  http://localhost:3000/docs-html/d/slug/assets/a.png   ← prefix now KEPT
/d/slug/assets/x.css  →  http://localhost:3000/docs-html/d/slug/assets/x.css
```
No longer drops to `<origin>/d/...` → no longer hits the deleted `/d/` location → no SPA-swallow. The exact regression I flagged is closed.

**Covers the real emit surface.** The repo's own contract test enumerates the forms octo-doc emits: root-relative `/d/{slug}/assets/` (now re-rooted ✅) and doc-relative `assets/`/`./assets/`/`../assets/` (resolve against the full render docUrl → `/d/slug/v/assets/…` ✅). Both handled. (A bare `/assets/*` root form isn't in the emit contract — not a real case.)

**New test is load-bearing (fail-on-revert verified).** "re-roots root-relative /d/ octo-doc assets under the same-origin /docs-html prefix (DEFAULT deploy)" asserts the re-rooted URL AND `.not.toContain` the prefix-less form; plus a "does not double-prefix" case. I reverted the `basePrefix` logic → the test **fails**; with the fix → passes. Genuine guard.

### Verification (local checkout @ `10390603a6d6`, pnpm installed)
| Check | Result |
|---|---|
| `vitest run src/html/HtmlDocView.test.tsx` | ✅ 48/48 (was 46; +2 re-root cases) |
| asset re-root (executed probe) | ✅ `/d/` assets keep `/docs-html`, no prefix-drop |
| re-root test fail-on-revert | ✅ fails on old code (guards the fix) |

### 🟡 encoded-slug (still open, non-blocking fast-follow)
The nginx render rewrite is unchanged this commit, so the `%2F`/`%3F`-in-slug decode-then-match concern on `/docs-html/docs|grants|docs-admin` remains. Low blast radius (read-only GET; worst case an exotic-slug 404, not data loss/corruption) and the repo does emit encoded slugs on those sub-paths (`htmlDocVersions.test.ts:51`) — so worth hardening (raw-passthrough on those rewrites like the `/api/v1/docs` block already does), but it doesn't block the asset fix. Recommend as a follow-up.

### What still stands (correct all along)
SPA `try_files` restoration ✅, `/api/v1/docs` `$request_uri` query fix (encoding-safe) ✅, single `/docs-html/*` prefix consolidation ✅.

`/d/` root-relative re-rooted under prefix (probe-verified) ✅ · doc-relative assets resolve correctly ✅ · no double-prefix ✅ · absolute/override docUrl untouched ✅ · covers real emit forms ✅ · fail-on-revert test ✅ · build + 48/48 ✅. Only the encoded-slug 🟡 remains, non-blocking.

---

### ✅ Review Verdict: APPROVE @ `10390603a6d6`
My REQUEST_CHANGES blocker (root-relative `/d/` octo-doc assets dropping the `/docs-html` prefix under the same-origin default → SPA-swallowed → broken images/CSS) is fixed at the root: assets are re-rooted under `basePrefix` before absolutizing, byte-proven via probe, covering the real emit forms, and locked by a fail-on-revert regression test. Build green, 48/48. This clears my CR. The encoded-slug rewrite-fidelity item remains a non-blocking fast-follow (raw-passthrough on the render/sub-path rewrites). The SPA-fallback + query fixes and prefix consolidation remain sound. Clean to merge. Credit to Steve for the render-site precision on the asset finding.

Jerry-Xin (dismissed):
The PR is relevant to octo-web, but the unified prefix breaks some published-document assets.

### 🔴 Blocking

- 🔴 Critical — Root-relative assets inside CSS and `srcset` still bypass `/docs-html`. `HtmlDocView.tsx` rewrites only `img[src]` and `link[href]`. The `<base>` fallback at `line 168` cannot fix root-relative URLs: `/d/...` always resolves from the origin root, regardless of `<base href="/docs-html/">`. Consequently, URLs in `srcset`, inline/style-block CSS, or fetched stylesheets such as `url(/d/slug/assets/font.woff2)` hit `/d/...`, which nginx no longer proxies and the SPA fallback serves as HTML. Rewrite every supported root-relative asset reference, make octo-doc emit prefixed/relative URLs, or provide another prefix-preserving solution. Add coverage for CSS `url(/d/.../assets/...)` and `srcset`.

- 🔴 Critical — The `/docs-html/*` render/grants/versions rewrites do not preserve percent-encoding. They capture from the decoded request path and re-emit it through `rewrite ... break`, so nginx re-encodes the captured groups; a slug carrying encoded reserved characters (`%2F`, `%3F`) collapses/double-encodes before reaching octo-doc. The sibling `/api/v1/docs` block already documents and uses the raw-passthrough workaround (appending the raw request URI to the upstream target to avoid the `?`→`%3F` breakage), but the render prefix rewrites were not synced to that pattern. Apply the same encoding-preserving passthrough to the `/docs-html/` render (and grants/versions) rewrites, and add coverage for an encoded-slug render URL.

### 💬 Non-blocking

- 🟡 Warning — The targeted test suite could not be executed because dependencies are absent (`vitest: command not found`).

### ✅ Highlights

- The `/api/v1/docs` query-forwarding correction is well explained.
- Consolidating octo-doc traffic under a dedicated namespace reduces route collisions.
- Restoring the SPA fallback addresses deep-link refresh failures.

lml2468:
### 🐙 On Jerry-Xin's two 🔴 — correcting myself: the CSS-`url()`/`srcset` gap is mechanically real; severity hinges on octo-doc's emit contract

I need to correct an error in my own reasoning before adjudicating, because I almost dismissed #1 on a false premise.

**🔴#1 (CSS `url()` / `srcset` root-relative `/d/` assets not re-rooted) — mechanically valid.** I first thought inline `style` was stripped (making CSS `url()` moot), but that's **wrong**: the render path is `injectBaseHref(absolutizeDocAssetUrls(html, url))` → iframe `srcDoc` (HtmlDocView.tsx:373/588) — it does **NOT** call `sanitizeDocHtml` (that's the "legacy fragment" sanitizer with `FORBID_ATTR: style`). So inline `style` reaches the srcdoc unstripped. And:
- `absolutizeDocAssetUrls` only rewrites `img[src]` + `link[href]` — **not** `srcset`, not CSS `url()`/`background-image`.
- `<base href="/docs-html/">` fixes **relative** URLs, but per spec does **NOT** re-root **root-absolute** `/d/...` paths.
So a root-relative `/d/{slug}/assets/…` inside CSS `url()` or an `srcset` would still resolve to `<origin>/d/...`, drop the prefix, hit the deleted location → broken. The re-root fix (targets `/d/` in `img/link` only) and `<base>` both miss this surface. Jerry-Xin is mechanically right, and I was wrong to lean on "style is stripped."

**The open question is reachability — does octo-doc emit root-relative `/d/` in CSS `url()`/`srcset`?** The repo's asset-contract tests only cover `img src` / `link href` root-relative `/d/` forms (HtmlDocView.test.tsx:93-123); there is **no `srcset` or CSS-`url()` fixture** with a `/d/` root-relative ref. So this is mechanically-possible but **unproven in the emit contract**. If octo-doc's published render never emits `/d/` root-relative refs in CSS/srcset (only in `img/link`, which the fix now handles), it's a non-issue; if it can, it's a real broken-asset gap on the same class as the one just fixed.

**🔴#2 (encoded-slug rewrite decode) — I already hold this as a non-blocking 🟡** (nginx render rewrite unchanged this commit; low blast radius — read-only GET, exotic-slug 404 not data loss; repo does emit encoded slugs on versions/grants/admin). Severity split: Jerry-Xin 🔴, me 🟡.

**Net / verdict:** the `img/link` root-relative regression I blocked on **is** fixed and verified (fail-on-revert test). Jerry-Xin surfaces an **adjacent** surface (CSS `url()`/`srcset`) that the fix doesn't cover and that I incorrectly reasoned was stripped — real mechanism, reachability unconfirmed. Given I just corrected my own error, I'm **not** going to over-claim it's safe: **if the author or Jerry-Xin can point to octo-doc emitting a root-relative `/d/` ref in CSS `url()` or `srcset`, this is a 🔴 and I'll flip** — the cleanest fix is to extend `absolutizeDocAssetUrls` to also re-root `srcset` + a CSS-`url()` pass (or have octo-doc emit `/docs-html`-relative assets). Absent that emit evidence, it + the encoded-slug are the two things to close before merge; both are the same "re-root/raw-passthrough for every asset+path form, not just `img/link`" root cause.

I'll defer to the peers' byte-check on the emit contract here rather than assert — my last two calls on this PR's asset surface were each corrected, so the honest move is: the fix closed the proven case, these two adjacent surfaces need the emit-reachability confirmed. Leaving my APPROVE standing would be over-confident given #1's mechanism is real; I'd put this back to **needs the CSS/srcset + encoded-slug forms confirmed-or-covered**.

lml2468 (changes requested):
## 🐙 PR #888 Re-Review (correcting my APPROVE) — REQUEST_CHANGES @ `10390603a6d6`

I'm superseding my APPROVE from earlier this turn. My follow-up comment acknowledged Jerry-Xin's CSS-`url()`/`srcset` gap is mechanically real but left my verdict hedged — that's an inconsistent state (live review APPROVE vs. my own analysis saying the mechanism holds). Resolving it to a definite verdict, honestly bounded.

### ✅ What the new commit DID fix (verified, stands)
Root-relative `/d/…` assets in **`img[src]` / `link[href]`** are now re-rooted under `/docs-html` (byte-proven via probe; fail-on-revert regression test). That's the case I originally blocked on, and it's genuinely closed. Also correct and unchanged: SPA `try_files`, the `/api/v1/docs` `$request_uri` query fix, the `/docs-html/*` consolidation.

### 🔴 Remaining — adjacent asset surfaces the re-root fix does NOT cover (Jerry-Xin #1, mechanically confirmed)
`absolutizeDocAssetUrls` rewrites only `img[src]` + `link[href]`. It does **not** touch:
- **CSS `url()` / `background-image`** in inline `style` — and I must correct my own earlier claim: the render path is `injectBaseHref(absolutizeDocAssetUrls(html))` → iframe `srcDoc` (:373/588); it does **NOT** run `sanitizeDocHtml` (the `FORBID_ATTR:'style'` stripper is the *legacy fragment* sanitizer). So inline `style` reaches the srcdoc unstripped.
- **`srcset`** — not in the querySelector set.

`<base href="/docs-html/">` does **not** rescue these: per spec `<base>` re-roots **relative** URLs, not **root-absolute** `/d/...`. So a root-relative `/d/{slug}/assets/…` inside CSS `url()` or `srcset` still resolves to `<origin>/d/...` → drops the prefix → hits the deleted location → broken, exactly like the `img/link` case that was just fixed.

**Reachability caveat (honest):** octo-doc is a **separate backend/repo**, so whether its published render actually emits root-relative `/d/` refs in CSS `url()`/`srcset` is **not verifiable from this repo** — the in-repo asset-contract tests only cover `img/link`. So this is mechanically-real, emit-unconfirmed. I'm grading it a blocker rather than a 🟡 because (a) it's the *same* prefix-drop failure mode as the one this PR exists to fix, just on sibling surfaces, and (b) I can't prove it's unreachable — and on an asset-integrity fix, "we fixed img/link but not the other ref forms" leaves the stated goal (assets survive under the new default) only partially met. If you can confirm octo-doc never emits `/d/` root-relative in CSS/srcset, I'll drop this to 🟡 and re-approve immediately.

### 🟡 encoded-slug (unchanged)
nginx render rewrite still matches decoded `$uri` (`%2F`/`%3F` slugs on versions/grants/admin, emitted per `htmlDocVersions.test.ts:51`). Low blast radius, fast-follow — same root as above: apply raw-passthrough on those rewrites like `/api/v1/docs` already does.

### Fix
Extend `absolutizeDocAssetUrls` to also re-root `srcset` and do a CSS-`url()` pass (or have octo-doc emit `/docs-html`-relative asset refs), + the encoded-slug raw-passthrough. Both are the one root cause: **re-root/raw-passthrough for every asset+path form, not just `img/link`.**

### Honest note
My reviews on this PR's asset surface have needed correction twice (missed the same-origin prefix-drop initially; then wrongly reasoned inline style was stripped). I'm converging with Jerry-Xin here on the byte-mechanism; the only genuinely open piece is the cross-repo emit contract, which neither of us can byte-verify in-repo — so I'm erring toward blocking on the unproven-but-plausible broken surface rather than approving over it.

---

### ❌ Review Verdict: REQUEST_CHANGES @ `10390603a6d6`
The `img/link` root-relative regression is fixed (verified). But the re-root fix doesn't cover CSS `url()` (inline style is NOT stripped on the render path — correcting my earlier claim) or `srcset`, and `<base>` can't re-root root-absolute `/d/` — so those asset forms would still drop the `/docs-html` prefix and break, if octo-doc emits them (unverifiable in-repo, same failure mode as the fixed case). Plus the open encoded-slug 🟡. Extend the re-root to srcset + CSS-url (or emit prefixed refs) + raw-passthrough the encoded-slug rewrites. Credit to Jerry-Xin for the CSS/srcset surface — and my correction: I was wrong that inline style is stripped on the render path.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #888 (octo-web)

## Summary

This PR consolidates the scattered octo-doc proxy routes (`/d/`, `/comments`, `/reactions`, `/grants/`, `/docs-admin/`) into a single `/docs-html/*` prefix, fixes a query-string encoding regression in the `/api/v1/docs` proxy that broke list endpoints with query parameters, and restores the missing SPA `try_files` fallback for client-side routing. The frontend `resolveOctoDocBase()` default changes from empty string to `/docs-html`, and `resolveDocAssetUrl()` is updated to re-root the backend's root-relative `/d/` asset references under the new prefix.

The change is well-scoped, well-tested, and the three commits tell a clear story. All production API callers (`htmlDocComments.ts`, `htmlGrantsApi.ts`, `htmlDocAdmin.ts`, `htmlDocVersions.ts`, `HtmlDocView.tsx`) use `resolveOctoDocBase()` and automatically pick up the new prefix.

## Verification

- ✅ **Query encoding fix** (`nginx.conf.template:155-159`) — `proxy_pass $docs_backend_url$request_uri` passes path+query verbatim to the upstream. The old `rewrite ^ $request_uri break` + bare variable `proxy_pass` caused nginx to URL-encode the `?` into `%3F`, turning `/api/v1/docs?owner=me` into a path-only request that 404'd. The new form restores pre-rewrite behavior. Verified the nginx variable-proxy_pass semantics: with a variable in `proxy_pass` and no URI suffix, the rewritten URI is forwarded as-is.

- ✅ **SPA fallback** (`nginx.conf.template:208-211`) — `try_files $uri $uri/ /index.html` in `location /` restores the standard React Router catch-all. Static assets and directory indexes are served normally; only unmatched paths fall through to `index.html`.

- ✅ **Unified `/docs-html/` rewrite ordering** (`nginx.conf.template:121-135`) — Rewrites are most-specific-first with `break` guards. Verified each path: `/docs-html/grants/{slug}/{uid}` → `/v1/docs/{slug}/grants/{uid}`, `/docs-html/docs-admin/{slug}` → `/v1/docs/{slug}`, `/docs-html/docs/{slug}/versions` → `/v1/docs/{slug}/versions`, `/docs-html/comments` → `/v1/comments`, `/docs-html/reactions` → `/v1/reactions`, `/docs-html/d/{slug}/v/{ver}` → `/d/{slug}/v/{ver}`, catch-all strips prefix. No regex ambiguity between `docs/` and `docs-admin/`.

- ✅ **Cross-file wiring (C1 — symmetric paths)** — Grepped all octo-doc API callers in `packages/docs/src/html/`: `htmlDocComments.ts:100` (`commentsUrl`), `htmlGrantsApi.ts:29` (`grantsUrl`), `htmlDocAdmin.ts:23` (`deleteDoc`), `htmlDocVersions.ts:26` (versions GET), and `HtmlDocView.tsx:228` (`buildOctoDocUrl`) all route through `resolveOctoDocBase()`. No hardcoded paths remain.

- ✅ **Asset re-rooting** (`HtmlDocView.tsx:88-95`) — `resolveDocAssetUrl` correctly re-roots `/d/...` asset refs when the doc URL sits under the same-origin prefix, skips re-rooting for cross-origin/override deploys. Tests cover re-root, double-prefix prevention, and cross-origin passthrough.

- ✅ **Auth headers preserved** — The unified `location /docs-html/` block retains `proxy_set_header token $http_token`, `Host`, `X-Real-IP`, `X-Forwarded-For` — identical to the five blocks it replaces. No auth regression.

- ✅ **Tests updated** — `HtmlDocView.test.tsx` default-base assertion updated to `/docs-html`, fetch-URL assertion updated to `/docs-html/d/published-slug/v/v7`. Two new tests cover asset re-rooting and double-prefix prevention.

## Findings

Static analysis only at head `10390603`; build and tests not executed in this environment.

- All 8 nginx rewrite regexes in the unified block produce the correct upstream URIs (verified each path individually).
- `resolver 127.0.0.11 ipv6=off valid=30s` and 503 guard for empty `$doc_app_url` are preserved in the consolidated block.
- The `$request_uri` appended to `$docs_backend_url` is already decoded by nginx's internal representation — no double-encoding risk.
- `location /docs-html/` (trailing slash) correctly takes priority over `location /` via nginx longest-prefix match. The frontend never constructs a bare `/docs-html` request (all URLs have sub-paths).
- Cross-origin deploy path (`__OCTO_DOC_BASE__` = absolute URL): `basePrefix` correctly becomes `''`, no rebasing attempted — the absolute base URL already carries the origin.
- The `/docs/{slug}/versions` rewrite (`^/docs-html/docs/(.+)$` → `/v1/docs/$1`) does not collide with `/docs-admin/` — the regex requires `docs/` (slash), which `docs-admin` cannot match.

## Verdict: APPROVED

Clean, well-structured change that fixes real production regressions (query encoding, SPA fallback) and consolidates route governance. All production callers verified to route through the unified prefix. No blockers.

Jerry-Xin (approved):
## 🐙 PR #888 Re-Review (superseding my REQUEST_CHANGES) — APPROVE @ `10390603a6d6`

I'm flipping my earlier REQUEST_CHANGES to APPROVE. My two 🔴 blockers both fall under cross-repo byte-verification; the only genuinely open piece (whether octo-doc emits root-relative `/d/` refs inside CSS `url()`/`srcset`) — which both I and lml2468 flagged as "mechanically-real but emit-unverifiable in-repo" — I have now verified across the octo-doc repos. It does not.

### ✅ What the new commit fixed (verified, stands)
Root-relative `/d/{slug}/assets/{sha}` refs in `img[src]` / `link[href]` are re-rooted under the same-origin `/docs-html` prefix when the doc URL sits under that prefix, with a fail-on-revert regression test and double-prefix guard. Also correct: SPA `try_files` fallback, the `/api/v1/docs` verbatim path+query passthrough, and the unified `/docs-html/*` consolidation.

### 🔴→REFUTED (A): CSS `url()` / `srcset` root-relative assets
My blocker assumed octo-doc might emit root-relative `/d/...` asset refs inside CSS `url()` or `srcset` that the `img/link`-only re-root would miss. I traced the actual emit contract across the octo-doc repos and it does not produce those forms:
- **Editor intake** normalizes every asset URL (both `src` and any `srcset` candidates) through a sanitizer that resolves to a fully-absolute href against origin and enforces a storage-host allowlist. Stored refs are absolute, not root-relative — so they are left untouched by the client and load directly.
- **Image node `toDOM`** emits only `` (plus `data-attach-id`); it never emits `srcset`. `srcset` exists only as a defensive parse/sanitize helper for pasted HTML, never as a stored render form.
- **Backend render/sign path**: the asset signer is a text-level regex over `/d/{slug}/assets/{64-hex}` and is form-agnostic, but the stored HTML it operates on carries the ref only in `` (the backend's own asset contract test uses exactly ``). No template/schema emits `/d/` inside CSS `url()`/`background-image` or `srcset`.
- A grep for `url(/d`, `url(/assets`, and `srcset`-with-`/d/`|`/assets` across all three octo-doc repos returns zero hits.
- Any *relative* CSS `url()`/`srcset` an authored doc might contain is already rescued by the injected `<base href="{origin}/docs-html/">` (relative resolution keeps the prefix). Only *root-relative* `/d/` would need the JS re-root — and that form is only emitted into `img[src]`/`link[href]`, which the fix now covers.

So 🔴 A is not reachable: every real emitted form is handled. Withdrawn.

### 🔴→🟡 (B): encoded-slug on the render/versions rewrites
The observation is correct — the old `/d/` block forwarded the raw request URI verbatim (preserving `%XX`), whereas the new `/docs-html/*` block uses capture-group rewrites that operate on the decoded URI, so an escapable char in a path segment would not be preserved the way the author's `/api/v1/docs` verbatim-passthrough does. But it is effectively unreachable for real docs: the octo-doc slug charset is restricted to `[a-zA-Z0-9_-]{1,64}` and the version is a numeric integer, so `encodeURIComponent` of any valid slug is a no-op — there is no `%2F`/`%3F` to drop. The `a/b?c → a%2Fb%3Fc` case in the frontend test is defensive encoding for a slug the backend would reject as `invalid_slug` regardless of encoding preservation. Read-only GET, no data loss. Downgraded to a non-blocking hardening follow-up: for consistency, the render/versions/grants/admin rewrites could adopt the same verbatim path+query passthrough the author already applied to `/api/v1/docs`.

### Verdict: APPROVED @ `10390603a6d6`
The `img/link` prefix-drop regression this PR targets is genuinely fixed and tested. The CSS-`url()`/`srcset` concern is not reachable in the real emit contract (verified across the octo-doc repos), and the encoded-slug gap is a low-impact consistency nit gated out by the slug charset, not a functional regression. Credit to lml2468 and the earlier reviewers for converging on the byte-mechanism; the cross-repo emit contract is what resolves it.

lml2468 (approved):
## 🐙 PR #888 Re-Review (verdict update) — APPROVE @ `10390603a6d6`

Updating my REQUEST_CHANGES to APPROVE — the **one condition I explicitly pre-committed to** is now met.

### What my CR was blocking on, and why it lifts now
My CR graded the CSS-`url()`/`srcset` root-relative-`/d/` gap as a blocker **solely** because its mechanism is real AND I **could not verify from this repo** whether octo-doc emits those ref forms (octo-doc is a separate backend/repo). I wrote verbatim: *"if you can confirm octo-doc never emits `/d/` root-relative in CSS/srcset, I'll drop this to 🟡 and re-approve immediately."*

**That confirmation is now on record from the finding's originator.** Jerry-Xin — who raised the CSS/srcset 🔴 — flipped to APPROVE after his own byte-check, concluding **"CSS url()/srcset 不存在根路径引用风险"** (octo-doc emits no root-relative `/d/` refs on those surfaces). That resolves the exact emit-contract unknown my blocker rested on. Since the emit-reachability was the only thing keeping it a 🔴 rather than a 🟡 (the `img/link` case IS fixed + tested), the blocker's basis is gone.

I can't independently byte-verify a separate repo's render template from here — that's precisely why my CR was conditioned on confirmation rather than asserting reachability. The originator's byte-verified flip is the decisive available evidence on that fact, and it converges the whole review.

### State of the findings
- ✅ **`img/link` root-relative re-root** — fixed & verified (probe + fail-on-revert test). This was my original P2.
- ✅ **CSS `url()`/`srcset`** — mechanism real, but **emit-unreachable** per Jerry-Xin's byte-confirmation → drops from 🔴 to a **defense-in-depth 🟡** (worth extending `absolutizeDocAssetUrls` to srcset + a CSS-url pass as hardening, since it costs little and future octo-doc changes could emit such refs — but not a current defect).
- 🟡 **encoded-slug** rewrite-decode on versions/grants/admin — unchanged, non-blocking fast-follow (raw-passthrough like `/api/v1/docs`).
- ✅ SPA `try_files`, `/api/v1/docs` `$request_uri` query fix, `/docs-html/*` consolidation — all correct.

### Verdict
### ✅ Review Verdict: APPROVE @ `10390603a6d6`
The blocker I held (CSS/srcset asset re-root gap) was conditioned on the octo-doc emit contract, which I couldn't verify in-repo; Jerry-Xin (the finding's originator) has now byte-confirmed those surfaces carry no root-relative `/d/` refs, so it drops to a non-blocking defense-in-depth 🟡. The proven `img/link` regression is fixed and tested; SPA-fallback, query-encoding, and prefix consolidation are sound. Two non-blocking follow-ups remain (harden absolutize for srcset/CSS-url; raw-passthrough the encoded-slug rewrites). No standing blockers. Clean to merge.

Honest close: I flipped APPROVE→CR→APPROVE on this PR's asset surface as the evidence resolved — the CR was the right call while the emit contract was unverifiable, and lifting it on the originator's byte-confirmation (my pre-stated condition) is the right call now. Thanks to Jerry-Xin for closing the emit-contract question I couldn't.

yujiawei (approved):
# Code Review — PR #888 (octo-web)

Reviewed at head SHA `10390603a6d63e28cf8640458609d6db5ef5fb82` (the third commit, which adds root-relative asset re-rooting on top of the earlier `291d7a78`). This review re-audits the current head independently; earlier review signals on `291d7a78` are treated as stale.

## Summary

The PR fixes three concrete regressions from the prior nginx refactor and consolidates all web→octo-doc traffic under a single `/docs-html/*` prefix. I verified the runtime behavior of the nginx rewrites empirically (nginx `1.29.8` in a container, exercising each location with real requests). The two live 404 bugs (docs query encoding, SPA deep-link fallback) are genuinely fixed. The one previously-flagged blocker (root-relative assets) is addressed at this head with tests. The remaining encoding concern is narrower than it first appears and is not reachable with today's slug charset.

## What's correct (verified)

- **`/api/v1/docs` query preservation** (`nginx.conf.template:152`). `proxy_pass $docs_backend_url$request_uri` forwards path+query byte-for-byte. Verified: `GET /api/v1/docs?owner=me` reaches the upstream as `/api/v1/docs?owner=me` (the prior `rewrite ^ $request_uri break` on a variable upstream encoded `?`→`%3F`, causing the 404). Correct fix.
- **SPA fallback restored** (`nginx.conf.template:211`). `try_files $uri $uri/ /index.html;` — deep links / refresh now resolve to the SPA instead of a hard 404. Correct.
- **Root-relative asset re-rooting** (`HtmlDocView.tsx:88-95`, `143-152`). Root-relative `/d/{slug}/assets/…` refs on `img[src]`/`link[href]` are rebased to `/docs-html/d/…` when the doc URL sits under the prefix, and left untouched for absolute/override bases. The `underPrefix` + `!value.startsWith(basePrefix + '/')` guard correctly avoids double-prefixing. Unit tests lock both the default-base rebase and the no-double-prefix case; all 63 tests in the touched html suites pass.
- **Rewrite ordering** (`nginx.conf.template:122-135`). Verified most-specific-first ordering resolves without collisions: `grants/{slug}/{uid}` precedes bare `grants/{slug}`; `docs-admin/` does not collide with `docs/` (the `-` vs `/` boundary); `comments`/`reactions` and the `d/…` render path all route to their intended upstream targets. Query strings are preserved through the rewrites (verified `?code=` on the render path and `?doc=&v=` on comments).

## Findings

### P2 — render `/d/` path loses raw-encoding fidelity (defense-in-depth, unreachable today)

`rewrite ^/docs-html/(d/.*)$ /$1 break;` (`nginx.conf.template:133`) rebuilds the forwarded URI from `$uri`, which nginx has already percent-decoded. Verified: `GET /docs-html/docs/a%2Fb%3Fc/versions` forwards as `/v1/docs/a/b?c/versions` — the `%2F` is decoded into a real path separator. This is a genuine narrowing versus the *old* standalone `/d/` block, which used `rewrite ^ $request_uri break` (raw, encoding-preserving passthrough).

Two things bound the severity to P2 rather than a blocker:

1. **Scope is only the render path.** The other four sub-paths this PR folds in — `comments`, `reactions`, `grants`, `docs-admin`, `versions` — were **already** rewrite-based (and therefore already decode-normalizing) *before* this PR. This PR introduces no new regression for them; it only changes `/d/` render from raw→decoded.
2. **Not reachable with today's slug charset.** `buildOctoDocUrl` `encodeURIComponent`s the slug, and octo-doc slugs in practice stay within `[A-Za-z0-9_-]` (none of which `encodeURIComponent` escapes), so decoded and raw bytes are identical for every real slug. Decisive corroboration: if slugs containing `/` or `?` were actually produced, the pre-existing decode-based `grants`/`versions`/`docs-admin` rewrites would *already* be broken today — they aren't, so special-char slugs aren't in play.

The frontend's own `htmlDocVersions.test.ts:51-52` does lock `listVersions('a/b?c')` → `…/a%2Fb%3Fc/…`, i.e. it documents the *intent* that slugs may carry special characters. If the slug charset is ever widened beyond `[A-Za-z0-9_-]`, this rewrite becomes reachable and would misroute. Suggested hardening (follow-up, not required for merge): forward the `d/` render path via raw `$request_uri` (the same technique already used for `/api/v1/docs`) rather than a `$uri`-based rewrite, and add nginx-level coverage for an encoded slug.

### P2 — root-relative `/d/` refs inside CSS `url()` still bypass the prefix (narrow residual)

`absolutizeDocAssetUrls` re-roots only `img[src]` and `link[href]` (`HtmlDocView.tsx:151-152`); CSS-referenced assets rely on the injected `<base href>`. In the default deploy that base is `http://{origin}/docs-html/`, and a root-relative CSS `url(/d/slug/assets/x.png)` resolves against the base *origin* only (the path segment is discarded by URL resolution), yielding `http://{origin}/d/…` — a path the nginx no longer proxies, so it falls through to `try_files` → `index.html`. This only bites if octo-doc emits root-relative `/d/…` refs inside CSS `url()`/`background-image` (the primary carriers — stylesheet `link[href]` and `img[src]` — are already handled). Low likelihood; worth a follow-up only if such CSS refs are confirmed to exist in published output.

## Spec compliance

- **Missing:** none. All three stated goals (query-encoding fix, SPA fallback, unified `/docs-html/*` prefix) are implemented.
- **Over-build:** none. No new feature flags, endpoints, or fields beyond the routing consolidation.
- **Deviation (minor, narrative-only):** the commit-1 description states `/d/` would forward verbatim via `proxy_pass $var$request_uri`, but the shipped head routes `/d/` through the unified `$uri`-based rewrite instead (superseded by the commit-2 unification). Net behavior matches the governing unification intent; the practical consequence is the P2 encoding item above.

## Verdict

The PR fixes two real, user-visible 404 regressions and addresses the prior blocker with test coverage. The residual items are P2 hardening/defense-in-depth against a slug charset that is not reachable today. No P0/P1 blocker at this head.

Approving. Please consider the two P2 items as follow-ups — the render-path raw-passthrough in particular is a cheap, targeted hardening that would eliminate the encoding-fidelity gap entirely if the slug charset ever widens.

PR #161, opened 2026-05-28
fix(i18n): sync Semi locale with app language

LLwill:
## Summary

- Sync Semi UI locale/direction with the app i18n provider for component chrome such as DatePicker.
- Localize remaining task deadline labels in create task modals.
- Stabilize task sidebar tab sizing in English so all segments share the same selected-state footprint.
- Document the Semi locale bridge for future i18n work.

## Validation

- pnpm i18n:check
- pnpm exec vitest run packages/dmworkbase/src/i18n/__tests__/detectLocale.test.ts packages/dmworkbase/src/i18n/__tests__/I18nService.test.ts packages/dmworkbase/src/i18n/__tests__/semiLocale.test.ts
- pnpm --filter @octo/web build
- Browser verified DatePicker chrome switches between English and Chinese.

Closes #147

PR #158, opened 2026-05-28
feat: use reset endpoint instead of delete for restoring defaults

ploy-elison:
Change the 'Restore Defaults' button behavior from DELETE to POST /reset.

- Calls `VoiceService.resetLocalConfig({enabled})` instead of `deleteLocalConfig()`
- Preserves current enabled state when resetting other fields
- Clearer semantic: reset to defaults ≠ delete config entirely

Closes #159

PR #902, opened 2026-07-20
refactor(thread): unify thread create flow

katydids:
## Summary
- Extract a shared ThreadCreateDialog / ThreadCreateForm for the existing create-subspace and create-topic entry points.
- Add ThreadService methods for thread creation while keeping service code limited to HTTP and response normalization.
- Add a bridge action for name-based thread creation so wk:thread-created is emitted consistently after successful creates.
- Rewire ThreadCreate, ThreadCreateModal, and ThreadPanel to reuse the shared create flow while preserving their existing loading, error, close, and success behavior.

## Validation
- pnpm exec vitest run packages/dmworkbase/src/ui/ThreadCreateDialog/__tests__/ThreadCreateDialog.test.tsx packages/dmworkbase/src/Service/__tests__/ThreadService.test.ts packages/dmworkbase/src/bridge/thread/__tests__/createThread.test.ts --environment jsdom
- pnpm i18n:check
- pnpm --dir apps/web build
- git diff --check

## Notes
- Shared layer touched: yes, packages/dmworkbase service and bridge code are updated.
- Legacy ThreadCreate remains as a delegating entry point; it is not removed.

PR #907, opened 2026-07-20
fix(dmworksummary): agent-summary UX fast-follow — persist referencedTask + fix forward-to-chat + hide unsupported buttons

wanyiwang06:
## Summary

Three agent-summary UX gaps discovered while testing the merged smart-summary #158 / #161 in local octo-alex stack:

1. **Referenced-task lost after refresh** → save fails with opaque 400
   ("Request failed with status code 400").
   Root cause: `referencedTask` lived only in React state while
   `session_id` was persisted in localStorage; on resume, session +
   history came back but referencedTask didn't, so
   `referenced_task_ids` was empty in the save request, backend
   fallback for borrowing origin from the referenced task had nothing
   to borrow → 40001 `origin_channel_id 未传且无法从 session 反查`.

2. **Forward-to-chat button silently does nothing on agent summaries**.
   Root cause: guard `!detail?.result?.content` returned early because
   `detail.result` is null for agent summaries (backend
   `agent_summary.go` writes to `personal_result` only, not
   `summary_result`).

3. **Regenerate / Set schedule / Edit buttons render but backend
   doesn't support them for agent tasks** — clicking either sends the
   task to a pipeline path with no participants/sources (worker either
   loops empty → task stuck at Pending or crashes → task Failed), or
   for edit, returns 40008 "总结结果不存在".

## Linked Spec

Follow-up to CHAT-REFERENCE-BASED-DESIGN-v1 and smart-summary #158 / #161.
No separate brief — three focused fixes traceable to specific reproducer scenarios.

## How verified

Local test in octo-alex compose stack (rebuild octo-web image, hard refresh, all 4 scenarios):

- **Reference persistence**: open "继续优化" from a summary detail page, chat,
  refresh, referenced badge auto-restored, save works.
- **Storage fallback**: `localStorage.clear()` mid-session, save now shows friendly
  toast instead of raw 400.
- **Forward-to-chat on agent summary**: opens conversation picker, sends to
  selected group successfully (previously no-op / no console output).
- **Regenerate / Schedule / Edit buttons**: hidden on agent summary
  (`trigger_type === TriggerType.AGENT`), preserved on traditional
  (trigger_type=MANUAL) summary — reverse-verified against a MANUAL task.

Automated:
- `pnpm test` — 457/457 tests pass across 33 files, including 9 new
  storage-helper tests + 1 new save-time 40001 handler test.
- `pnpm build` — turbo build 2/2 pass, no rolldown warnings introduced.

## COMPREHENSION

1. **What does this change actually do**: makes agent-summary UX
   feature-parity with what the frontend already renders — either the
   backend path works (persist + forward), or the button is hidden so
   users can't step on the broken paths (regenerate / schedule / edit).
2. **What could break**: `referencedTask` snapshot in localStorage may
   drift (referenced summary renamed / deleted) → title shown may be
   stale but save still works because backend re-validates by `task_id`
   via `canAccessTaskDB`; malformed storage payloads are defended
   against with typed shape checks in the read helper.
3. **How do you know it works**: three-scenario local walkthrough + full
   pnpm test suite green + build green + reverse-tested that traditional
   summaries retain the three gated buttons.

### DIFF
--- a/packages/dmworksummary/src/i18n/en-US.json
+++ b/packages/dmworksummary/src/i18n/en-US.json
@@ -195,7 +195,8 @@
         "titleRequired": "Title is required",
         "saving": "Saving...",
         "noOutputToSave": "No summary content to save yet. Please chat with AI first to generate content.",
-        "agentSummaryCreated": "AI summary saved"
+        "agentSummaryCreated": "AI summary saved",
+        "savedReferenceLostRetry": "Save failed: please re-select the referenced summary, or start a new session"
     },
     "chatSummary": {
         "panelTitle": "Summaries in this chat",
--- a/packages/dmworksummary/src/i18n/zh-CN.json
+++ b/packages/dmworksummary/src/i18n/zh-CN.json
@@ -195,7 +195,8 @@
         "titleRequired": "标题不能为空",
         "saving": "保存中...",
         "noOutputToSave": "当前对话还没有可保存的总结，请先与 AI 对话产出内容",
-        "agentSummaryCreated": "AI 总结已保存"
+        "agentSummaryCreated": "AI 总结已保存",
+        "savedReferenceLostRetry": "保存失败：请重新选择引用总结，或点「新会话」重来"
     },
     "chatSummary": {
         "panelTitle": "聊天内的智能总结",
--- a/packages/dmworksummary/src/pages/SummaryCreatePage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryCreatePage.tsx
@@ -39,7 +39,7 @@ import type {
     CreateAgentSummaryParams,
 } from "../types/summary";
 import { SummaryMode, SourceType } from "../types/summary";
-import { describeSchedule, scheduleToParams, genSessionId, readAgentChatSession, writeAgentChatSession, clearAgentChatSession } from "../utils/summaryHelpers";
+import { describeSchedule, scheduleToParams, genSessionId, readAgentChatSession, writeAgentChatSession, clearAgentChatSession, readAgentChatReferenced, writeAgentChatReferenced, clearAgentChatReferenced } from "../utils/summaryHelpers";
 import { resolveTemplate, computeTemplateSelection, getTemplateEditableFields, deriveSummaryTitle, type ResolvableTemplate } from "../utils/templateResolver";
 
 const { Text } = Typography;
@@ -161,6 +161,12 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
             this.setState({
                 mode: 'agent',
                 referencedTask: this.props.derivedFromTask,
+            });
+            // 与 session_id 同生命周期持久化引用总结，避免 refresh/重进后
+            // referencedTask 只活在 React state 里而丢失 → 保存时 400。
+            writeAgentChatReferenced(this.agentChannelId(), {
+                task_id: this.props.derivedFromTask.task_id,
+                title: this.props.derivedFromTask.title ?? '',
             });
         }
     }
@@ -575,9 +581,15 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
      */
     private enterAgentMode() {
         const stored = readAgentChatSession(this.agentChannelId());
+        // 恢复引用总结与 session 同生命周期：storage 里有 → 自动回填。
+        // 无 → 保持 state 现值（可能是 mount 时 derivedFromTask 塞进来的）。
+        const storedRef = readAgentChatReferenced(this.agentChannelId());
         this.setState((prev) => ({
             mode: 'agent',
             sessionId: stored || prev.sessionId,
+            referencedTask: storedRef
+                ? { task_id: storedRef.task_id, title: storedRef.title } as SummaryListItem
+                : prev.referencedTask,
         }));
         if (stored) void this.loadAgentHistory(stored);
     }
@@ -603,6 +615,8 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
     /** 「新会话」：清 localStorage 的 session_id、清空消息，下次发送重新生成新 session_id。 */
     handleNewSession = () => {
         clearAgentChatSession(this.agentChannelId());
+        // 引用总结跟 session 同生命周期 → 一起清。
+        clearAgentChatReferenced(this.agentChannelId());
         // 作废在途历史拉取，避免旧会话历史回灌到新会话。
         this.historyLoadToken++;
         this.setState({
@@ -646,6 +660,8 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
                             e.stopPropagation();
                             // 移除引用同时强制关闭 SidePanel(引用没了没意义再显示)
                             this.setState({ referencedTask: null, sidePanelOpen: false });
+                            // 引用同步清持久化，避免 refresh 后又回填。
+                            clearAgentChatReferenced(this.agentChannelId());
                         }}
                         title={translate('summary.chatReference.remove')}
                     >
@@ -718,6 +734,8 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
             //   2. 重置组件内 state(messages/sessionId/referencedTask)
             //   3. 后端会在保存事务里 DELETE agent_message 表对应行
             clearAgentChatSession(this.agentChannelId());
+            // 引用总结跟 session 同生命周期 → 一起清。
+            clearAgentChatReferenced(this.agentChannelId());
             this.historyLoadToken++;
             this.setState({
                 messages: [],
@@ -747,6 +765,14 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
                 // 40004: session 无产出
                 if (code === 40004) {
                     Toast.error(t('summary.create.noOutputToSave'));
+                    return false;
+                }
+                // 40001: origin_channel_id 反查失败(通常是引用总结退出重进后
+                // referencedTask 丢失,前端没发 referenced_task_ids,后端 fallback
+                // 借 origin 无路可走)。给友好文案指导用户下一步动作。
+                // 见 SUM-161 fast-follow · CHAT-REFERENCE-BASED-DESIGN-v1。
+                if (code === 40001) {
+                    Toast.error(t('summary.create.savedReferenceLostRetry'));
                     return false;
                 }
             }
@@ -844,10 +870,17 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
                             <SummaryReferencePicker
                                 visible={this.state.showReferencePicker}
                                 onCancel={() => this.setState({ showReferencePicker: false })}
-                                onSelect={(task) => this.setState({
-                                    referencedTask: task,
-                                    showReferencePicker: false,
-                                })}
+                                onSelect={(task) => {
+                                    this.setState({
+                                        referencedTask: task,
+                                        showReferencePicker: false,
+                                    });
+                                    // 用户选择新引用 → 同步持久化 → refresh 后仍在。
+                                    writeAgentChatReferenced(this.agentChannelId(), {
+                                        task_id: task.task_id,
+                                        title: task.title ?? '',
+                                    });
+                                }}
                                 selectedTaskId={this.state.referencedTask?.task_id}
                             />
                             {/* Modal 保留:未来其他触发点(比如详情页快照预览)可复用;主 UI 已改用 SidePanel */}
--- a/packages/dmworksummary/src/pages/SummaryCreatePage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryCreatePage.tsx
@@ -158,9 +158,20 @@ export default class SummaryCreatePage extends Component<SummaryCreatePageProps,
         // 从详情页「继续优化」打开时:自动切 agent 模式 + 预填引用。
         // 见 CHAT-REFERENCE-BASED-DESIGN-v1 决策 1B(详情页显眼按钮入口)。
         if (this.props.derivedFromTask) {
+            // #907 review (Jerry-Xin) P1 cross-session contamination:
+            // 走「继续优化」= 用户明确要针对当前 task 开新一轮 · 复用工作台
+            // 上一次残留的 session_id 语义完全不匹配(旧 chat 讨论的是别的
+            // 总结 · 现在换了 reference)。如果只 overwrite referenced 不清
+            // session · refresh-before-send 会 restore「旧 session_id + 新
+            // reference」的错配组合 · loadAgentHistory 灌回旧 messages ·
+            // 保存时血统被污染。所以进入时先原子清一遍 session · 再 write
+            // 新 reference · 保证 storage 里的两条永远一致。
+            clearAgentChatSession(this.agentChannelId());
             this.setState({
                 mode: 'agent',
                 referencedTask: this.props.derivedFromTask,
+                sessionId: '',
+                messages: [],
             });
             // 与 session_id 同生命周期持久化引用总结，避免 refresh/重进后
             // referencedTask 只活在 React state 里而丢失 → 保存时 400。
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -1887,10 +1887,17 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     };
 
     handleForwardToChat = () => {
-        const { detail } = this.state;
-        if (!detail?.result?.content?.trim()) return;
+        const { detail, personalResult } = this.state;
+        // #158/#161 agent 总结 fallback:agent workflow 只写 personal_result 表,
+        // 不写 summary_result 表(agent_summary.go: creatorPR 是 deliverable,没有
+        // 单独的 SummaryResult 行)。所以 GET /summaries/:id 返 detail.result=null,
+        // 但 detail.personal_result 有 content(前端渲染正文也是走这个 fallback)。
+        // 传统 workflow 优先走 detail.result;agent workflow 走 personalResult。
+        // 两者的 content 语义都是"给用户看的最终交付文本",转发到聊天的姿势一致。
+        const sourceContent = detail?.result?.content ?? personalResult?.content ?? '';
+        if (!sourceContent.trim()) return;
         WKApp.shared.baseContext.showConversationSelect(async (channels: Channel[]) => {
-            const cleanContent = (detail?.result?.content ?? '').replace(/\[\d+\]/g, '').replace(/  +/g, ' ').trim();
+            const cleanContent = sourceContent.replace(/\[\d+\]/g, '').replace(/  +/g, ' ').trim();
             const chunks = splitSummaryText(cleanContent);
             const errors: string[] = [];
 
@@ -2453,7 +2460,14 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                         {/* need5：单人 BY_PERSON 定时按钮放「编辑」按钮左边。多人协作不在此区渲染
                             （need1 不显示我的总结区；need5 多人定时按钮在团队框）。 */}
                         {!this.isMultiCollab() && this.renderScheduleButton()}
-                        {detail && detail.status === TaskStatus.COMPLETED && detail.permissions?.can_edit && !this.state.isEditing && (
+                        {/* #158/#161 fast-follow：agent 总结不支持 edit —— 后端 PUT /edit
+                            走的是 SummaryResult 表更新，但 agent 保存路径只写
+                            personal_result 表（agent_summary.go 里 creatorPR 是唯一
+                            deliverable，不建 summary_result 行）。用户点击编辑保存后
+                            会 404 "总结结果不存在"。前端直接不渲染。
+                            如果未来后端为 agent 建 SummaryResult 行，只需删除下面
+                            这行 trigger_type 判断即可。 */}
+                        {detail && detail.status === TaskStatus.COMPLETED && detail.permissions?.can_edit && !this.state.isEditing && detail.trigger_type !== TriggerType.AGENT && (
                             <Button
                                 size="small"
                                 theme="borderless"
@@ -3087,6 +3101,11 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
         const { t } = this.context;
         // OCT-21 / GPT-S1：草稿编辑态也隐藏 schedule 按钮，与其它编辑态保持一致约束。
         if (!detail?.permissions?.can_schedule || isEditing || editingTeamSummary || editingPersonalReport || editingMyDraft) return null;
+        // #158/#161 fast-follow：agent 总结不支持 schedule —— schedule 到点会 trigger
+        // 后端 pipeline 走传统 map-reduce 重跑，但 agent 总结的产出是 chat 交互生成，
+        // 没有可 replay 的 sources/participants。触发后任务会卡在 Pending 或 fail，
+        // 用户体验很差。前端直接不渲染这个按钮，避免用户误点。
+        if (detail?.trigger_type === TriggerType.AGENT) return null;
 
         // 任务3：hasSchedule 仅在存在且 is_active 时为 true。
         // 停用后文案回到「设置定时更新」。
@@ -3267,7 +3286,12 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
 
         // Build "..." menu items
         const menuItems: { node: string; key: string; onClick: () => void; danger?: boolean }[] = [];
-        if (detail && canRegenerate(detail.status)) {
+        // #158/#161 fast-follow：agent 总结不支持 regenerate —— 传统 regenerate 走
+        // POST /summaries/:id/regenerate → triggerWorker("personal_summary")，会尝试重跑
+        // pipeline 但 agent 任务无 participants / sources 可 replay，任务会卡死或 fail。
+        // 语义上 "重生成" 对 agent 总结应该是 "重开一次 chat"，那是主人的 continueRefine
+        // 已经覆盖的入口。这里直接不给 menu 项，避免用户走错路径。
+        if (detail && canRegenerate(detail.status) && detail.trigger_type !== TriggerType.AGENT) {
             menuItems.push({ node: t("summary.detail.regenerate"), key: "regenerate", onClick: this.handleRegenerate });
         }
         if (detail && canCancel(detail.status)) {
--- a/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
+++ b/packages/dmworksummary/src/pages/SummaryDetailPage.tsx
@@ -1941,10 +1941,14 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     };
 
     handleForwardToMatter = () => {
-        const { detail } = this.state;
+        const { detail, personalResult } = this.state;
         if (!detail || detail.status !== TaskStatus.COMPLETED) return;
 
-        const content = detail.result?.content;
+        // #907 review (yujiawei P2): mirror handleForwardToChat's agent
+        // summary fallback so this path won't ship broken when
+        // SHOW_FORWARD_TO_MATTER is flipped back on. See handleForwardToChat
+        // for the full rationale (agent workflow only writes personal_result).
+        const content = detail.result?.content ?? personalResult?.content;
         if (!content?.trim()) {
             Toast.warning(t("summary.detail.noForwardContent"));
             return;
@@ -1954,10 +1958,11 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
     };
 
     handleMatterSelected = async (matterId: string, matterTitle: string) => {
-        const { detail } = this.state;
+        const { detail, personalResult } = this.state;
         if (!detail) return;
 
-        const content = detail.result?.content;
+        // Same fallback as handleForwardToMatter (they must stay in sync).
+        const content = detail.result?.content ?? personalResult?.content;
         if (!content?.trim()) return;
 
         this.setState({ forwardingToMatter: true, showMatterPicker: false });
@@ -3314,15 +3319,29 @@ export default class SummaryDetailPage extends Component<SummaryDetailPageProps,
                                 {t("summary.detail.continueRefine")}
                             </Button>
                         )}
-                        {detail && detail.status === TaskStatus.COMPLETED && (
-                            <Button
-                                theme="borderless"
-                                icon={<IconSend />}
-                                onClick={this.handleForwardToChat}
-                            >
-                                {t("summary.detail.forwardToChat")}
-                            </Button>
-                        )}
+                        {detail && detail.status === TaskStatus.COMPLETED && (() => {
+                            // #907 review (yujiawei P2-1): agent summary forward
+                            // sources fall through to personalResult.content, which
+                            // is fetched async by loadPersonalResult. During that
+                            // load window a click would silently no-op. Disable +
+                            // loading state until the fallback content is in.
+                            // Traditional workflow has detail.result inline, so it
+                            // is never gated here.
+                            const isAgent = detail.trigger_type === TriggerType.AGENT;
+                            const agentContentReady = !!this.state.personalResult?.content?.trim();
+                            const waitingForFallback = isAgent && !detail.result?.content?.trim() && !agentContentReady;
+                            return (
+                                <Button
+                                    theme="borderless"
+                                    icon={<IconSend />}
+                                    onClick={this.handleForwardToChat}
+                                    loading={waitingForFallback && this.state.personalLoading}
+                                    disabled={waitingForFallback}
+                                >
+                                    {t("summary.detail.forwardToChat")}
+                                </Button>
+                            );
+                        })()}
                         {SHOW_FORWARD_TO_MATTER && detail && detail.status === TaskStatus.COMPLETED && (
                             <Button
                                 theme="borderless"
--- a/packages/dmworksummary/src/pages/__tests__/SummaryCreatePage.test.tsx
+++ b/packages/dmworksummary/src/pages/__tests__/SummaryCreatePage.test.tsx
@@ -509,3 +509,57 @@ describe('SummaryCreatePage agent SSE session_id sync', () => {
         expect(lastMessage.content).toBe('Server response');
     });
 });
+
+describe('SummaryCreatePage handleSubmit error handling', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+    });
+
+    it('shows friendly toast for 40001 (referenced summary lost after refresh)', async () => {
+        const { Toast } = await import('@douyinfe/semi-ui');
+
+        // Mock createAgentSummary to reject with axios-style 40001 error.
+        // This is the exact shape the backend returns when
+        //   session_id has no fetch_channel tool trace AND
+        //   referenced_task_ids is empty AND
+        //   origin_channel_id was not supplied by the front-end.
+        // See internal/api/handler/agent_summary.go: "origin_channel_id 未传且无法从 session 反查".
+        const err = {
+            response: { data: { code: 40001, message: 'origin_channel_id 未传且无法从 session 反查' } },
+        };
+        (api.createAgentSummary as any).mockRejectedValueOnce(err);
+
+        const ref = React.createRef<any>();
+        await act(async () => {
+            render(<SummaryCreatePage ref={ref} />);
+            await flushPromises();
+        });
+
+        const instance = ref.current as any;
+
+        // Simulate a completed agent chat session ready to save.
+        await act(async () => {
+            instance.setState({
+                sessionId: 'session-abc',
+                mode: 'agent',
+                messages: [{ role: 'user', content: 'hi' }, { role: 'assistant', content: 'summary' }],
+            });
+        });
+
+        (Toast.error as any).mockClear();
+
+        // Trigger the save flow directly on the instance.
+        let result: boolean | undefined;
+        await act(async () => {
+            result = await instance.handleSaveAsSummary('a title');
+            await flushPromises();
+        });
+
+        // Whichever method the component exposes, the 40001 branch should
+        // surface the friendly, actionable copy — NOT the raw backend message.
+        expect(Toast.error).toHaveBeenCalled();
+        const shown = (Toast.error as any).mock.calls[0]?.[0] ?? '';
+        expect(shown).toBe('保存失败：请重新选择引用总结，或点「新会话」重来');
+        expect(result).toBe(false);
+    });
+});
--- a/packages/dmworksummary/src/pages/__tests__/SummaryCreatePage.test.tsx
+++ b/packages/dmworksummary/src/pages/__tests__/SummaryCreatePage.test.tsx
@@ -563,3 +563,80 @@ describe('SummaryCreatePage handleSubmit error handling', () => {
         expect(result).toBe(false);
     });
 });
+
+describe('SummaryCreatePage derivedFromTask cross-session isolation (#907 P1 Jerry-Xin)', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+        localStorage.clear();
+    });
+
+    it('clears leftover workbench session_id + messages when mounted with derivedFromTask', async () => {
+        // Pre-seed leftover state from a previous chat about a DIFFERENT summary:
+        //   - session_id survived from earlier workbench chat
+        //   - referencedTask points to the old summary (task_id=99)
+        // This mirrors the exact bug scenario Jerry-Xin flagged:
+        //   user chats about A, closes without saving, opens "continue refine"
+        //   from summary B's detail page → new mount receives derivedFromTask=B
+        //   → old session_id (about A) must be discarded, else refresh-before-send
+        //   restores A's history alongside B's reference and save corrupts derivation.
+        localStorage.setItem('agent-chat-session:__workbench__', 'old-session-about-A');
+        localStorage.setItem(
+            'agent-chat-referenced:__workbench__',
+            JSON.stringify({ task_id: 99, title: 'Old Summary A' }),
+        );
+
+        const derivedFromTaskB = {
+            task_id: 42,
+            task_no: 'ST-B',
+            title: 'New Summary B',
+            summary_mode: 1 as const,
+            status: 3 as const,
+            trigger_type: 3,
+        };
+
+        const ref = React.createRef<any>();
+        await act(async () => {
+            render(<SummaryCreatePage ref={ref} derivedFromTask={derivedFromTaskB as any} />);
+            await flushPromises();
+        });
+
+        const instance = ref.current as any;
+
+        // 1. Old session_id storage key MUST be gone — no restore of the stale A chat.
+        expect(localStorage.getItem('agent-chat-session:__workbench__')).toBeNull();
+
+        // 2. Referenced storage key MUST point to B (the new derivation target), not A.
+        const storedRef = JSON.parse(localStorage.getItem('agent-chat-referenced:__workbench__') || 'null');
+        expect(storedRef).toEqual({ task_id: 42, title: 'New Summary B' });
+
+        // 3. React state must reflect a fresh session for B, not resumed A.
+        expect(instance.state.sessionId).toBe('');
+        expect(instance.state.messages).toEqual([]);
+        expect(instance.state.referencedTask?.task_id).toBe(42);
+        expect(instance.state.mode).toBe('agent');
+    });
+
+    it('when derivedFromTask is absent, does NOT touch existing workbench session (bare workbench entry unchanged)', async () => {
+        // Reverse guard: bare full-page workbench entry (no derivedFromTask prop)
+        // must NOT clear the user's in-progress session. This test protects the
+        // #158/#161 resume-after-refresh scenario from being broken by the
+        // #907 P1 fix — the two behaviours are orthogonal.
+        localStorage.setItem('agent-chat-session:__workbench__', 'in-progress-session');
+        localStorage.setItem(
+            'agent-chat-referenced:__workbench__',
+            JSON.stringify({ task_id: 7, title: 'Reference C' }),
+        );
+
+        await act(async () => {
+            render(<SummaryCreatePage />);
+            await flushPromises();
+        });
+
+        // Both storage keys survive the mount — enterAgentMode will restore them
+        // when the user actually switches into agent mode.
+        expect(localStorage.getItem('agent-chat-session:__workbench__')).toBe('in-progress-session');
+        expect(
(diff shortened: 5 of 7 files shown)


Jerry-Xin (dismissed):
The PR is relevant, but the new persistence logic can associate an old chat session with the wrong referenced task.

## 🔴 Blocking

- 🔴 Critical — Opening “Continue refine” overwrites the stored reference without clearing the previous stored session. In [SummaryCreatePage.tsx:160](/Users/mlamp/.openclaw/tmp/pr-review-b_j0prfv/repo/packages/dmworksummary/src/pages/SummaryCreatePage.tsx:160), `derivedFromTask` writes a new reference but leaves the workbench session key intact. If a previous session exists and the page refreshes before the first new message, `enterAgentMode()` restores that old session together with the new reference at [SummaryCreatePage.tsx:583](/Users/mlamp/.openclaw/tmp/pr-review-b_j0prfv/repo/packages/dmworksummary/src/pages/SummaryCreatePage.tsx:583). Saving then submits the old session with the unrelated task ID at line 724, corrupting the derivation relationship and potentially borrowing origin metadata from the wrong task.

  Clear the stored session when entering through `derivedFromTask`, or persist session and reference atomically and verify they belong together. Add a regression test covering: existing session → continue refine another task → refresh before sending → restore.

## 💬 Non-blocking

- 🟡 Warning — Mapping every backend `40001` response to a lost-reference message at [SummaryCreatePage.tsx:767](/Users/mlamp/.openclaw/tmp/pr-review-b_j0prfv/repo/packages/dmworksummary/src/pages/SummaryCreatePage.tsx:767) may misdiagnose other invalid-request cases sharing that generic code. A dedicated backend error code would be safer.

- 🔵 Suggestion — Add component tests for forwarding agent `personalResult` content and hiding edit, schedule, and regenerate controls. Current additions do not exercise these UI changes.

## ✅ Highlights

- Defensive storage parsing and failure handling are solid.
- Forwarding now follows the agent summary’s actual content source.
- Unsupported agent actions are consistently gated by `TriggerType.AGENT`.

Focused tests could not be executed because the checkout lacks the `vitest` executable/dependencies.

Jerry-Xin (dismissed):
The PR is relevant and the three fixes are on the right track, but the new persistence logic can associate an old chat session with the wrong referenced task.

## 🔴 Blocking

- 🔴 Critical — Opening "Continue refine" overwrites the stored reference without clearing the previous stored session. In `packages/dmworksummary/src/pages/SummaryCreatePage.tsx:160` (`componentDidMount`), a `derivedFromTask` mount writes the new `referencedTask` to localStorage but leaves the workbench `session_id` key intact. If a previous agent session exists under the same channel key and the page refreshes before the first new message is sent, `enterAgentMode()` (`SummaryCreatePage.tsx:583`) restores that **old** `session_id` together with the **new** reference, and `loadAgentHistory(stored)` reloads the old session's messages. Saving then submits the old session with the unrelated `referenced_task_ids` (`handleSaveAsSummary`, ~line 724), corrupting the derivation relationship and potentially borrowing origin metadata from the wrong task. The workbench fallback key (`__workbench__`, used for the full-page entry where channelId is undefined) makes this collision reachable.

  Fix: clear the stored session when entering through `derivedFromTask` (so refresh starts a clean session for the new reference), or persist session and reference atomically and verify they belong together on restore. Add a regression test: existing session → continue-refine a different task → refresh before sending → save, asserting the reference and session are consistent.

## 💬 Non-blocking

- 🟡 Warning — Mapping every backend `40001` at `SummaryCreatePage.tsx:767` to the lost-reference message may misdiagnose other invalid-request cases sharing that generic code. A dedicated backend error code would be safer and more precise.

- 🔵 Suggestion — Add component tests covering forwarding agent `personalResult` content and hiding the edit / schedule / regenerate controls for `TriggerType.AGENT`. The new unit tests cover the storage helpers and the 40001 toast, but not these UI-gating changes.

## ✅ Highlights

- Defensive storage parsing/failure handling in the new referenced-task helpers is solid (typed shape checks, silent degrade on private-mode/quota, per-channel scoping, minimal snapshot).
- Forward-to-chat now correctly sources content from `detail.result.content ?? personalResult.content`, matching the agent-summary render fallback; it reuses the existing `splitSummaryText` + `MessageText` send pipeline and does not touch the doc-forward path, so no regression to the unified forward work.
- Unsupported agent actions (regenerate / schedule / edit) are consistently gated by `trigger_type === TriggerType.AGENT` at the render/menu-build sites, so the actions are truly unreachable, not just visually hidden.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #907 (octo-web)

## Summary

This PR is a fast-follow UX polish pass for agent-generated summaries. It adds localStorage persistence for the `referencedTask` (so the referenced summary survives page refresh/resume), wires a friendly 40001 error toast when the backend can't resolve the origin channel, fixes the "Forward to Chat" action to fall back to `personalResult.content` for agent summaries (whose `detail.result` is null by design), and hides the Edit / Schedule / Regenerate buttons for agent summaries since those actions target the traditional `summary_result` pipeline that agent summaries don't use.

The persistence helpers in `summaryHelpers.ts` are clean and well-tested. The lifecycle management (write on mount + picker select, read on agent-mode enter, clear on new-session / manual-remove / save-success) is thorough. Two P2 items below — neither blocks correctness today but both leave rough edges for the next person.

## Verification

- ✅ **Persistence lifecycle coverage** — write paths (mount with `derivedFromTask` at `SummaryCreatePage.tsx:167`, picker select at `:877`), read path (`enterAgentMode` at `:586`), and all three clear paths (new-session at `:619`, manual-remove at `:662`, save-success at `:738`) are wired. No orphaned persistence entry.
- ✅ **Forward-to-chat fallback** — `handleForwardToChat` at `SummaryDetailPage.tsx:1897` correctly chains `detail?.result?.content ?? personalResult?.content ?? ''`. Data flow verified: agent summaries have `summary_mode === BY_PERSON`, so `loadPersonalResult` is called at `:548` and `personalResult.content` is populated from `api.getPersonalResult(taskId)`.
- ✅ **Button guard parity** — edit (`:2470`), schedule (`:3108`), and regenerate (`:3294`) all consistently guard with `detail.trigger_type !== TriggerType.AGENT`. The continue-refine button at `:3308` correctly uses `=== TriggerType.AGENT`.
- ✅ **40001 error handler** — `axiosErr.response?.data?.code` extraction at `SummaryCreatePage.tsx:763` uses correct optional chaining; placed after 40004 check with no fall-through; i18n keys present in both `en-US.json:198` and `zh-CN.json:198`.
- ✅ **localStorage helper robustness** — `readAgentChatReferenced` at `summaryHelpers.ts:96` validates `typeof task_id === 'number'` and `typeof title === 'string'`, handles malformed JSON, and swallows exceptions for privacy-mode localStorage. Unit tests in `agentChatReferenced.test.ts` cover key scoping, round-trip, malformed JSON, wrong shape, null-clear shortcut, channel isolation, and extra-field stripping.
- ✅ **`agentChatReferencedKey` collision-free** — uses distinct prefix `agent-chat-referenced:` vs the session key prefix `agent-chat-session:`, sharing the fallback segment `__workbench__` without namespace collision.

Static analysis only at head `b455678029c6e525834b2e763a6615ac48546d2b`; build and tests not executed in this environment.

## Findings

### P2 — `handleForwardToMatter` misses the `personalResult` fallback (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1944`)

The PR fixes `handleForwardToChat` to fall back to `personalResult?.content` for agent summaries, but leaves the sibling `handleForwardToMatter` at `:1944` reading only `detail.result?.content`. Since agent summaries have `detail.result === null` (confirmed by the forward-to-chat comment at `:1895`), clicking "Forward to Matter" on an agent summary always hits the `noForwardContent` toast — even though the content exists in `personalResult`.

The forward-to-matter button at `:3326` is rendered for all `COMPLETED` details with no `trigger_type` guard, so the button is visible and clickable but non-functional for agent summaries. This is a C1 sibling-path parity miss.

Fix direction: either apply the same `detail?.result?.content ?? personalResult?.content ?? ''` fallback, or guard the button with `trigger_type !== TriggerType.AGENT` consistent with the edit/schedule/regenerate guards.

### P2 — Unsafe `as SummaryListItem` cast from minimal persisted shape (`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:590`)

The persisted `{ task_id, title }` object is cast `as SummaryListItem`, which requires ~15 additional fields (`task_no`, `summary_mode`, `status`, `trigger_type`, `sources`, `time_range_start`, etc.). Current consumers only access `task_id` and `title`, so there is no runtime breakage today. However, the cast defeats the TypeScript compiler's ability to flag any future code that reads other fields from `this.state.referencedTask` after it was restored from localStorage — accessing `referencedTask.status` would silently return `undefined` rather than producing a type error.

Fix direction: narrow the `referencedTask` state type to `Pick<SummaryListItem, 'task_id' | 'title'> | null`, or introduce a dedicated `ReferencedTaskRef` interface. The full `SummaryListItem` is not needed for the consumed fields.

- The `splitSummaryText` import and citation-stripping regex in `handleForwardToChat` are unchanged and work correctly with the new `sourceContent` variable.
- The `writeAgentChatReferenced(null)` convenience shortcut correctly delegates to `clearAgentChatReferenced`, maintaining a single clear path.
- The i18n keys are added to both `en-US.json` and `zh-CN.json` in the same `summary.create` namespace, and the `t('summary.create.savedReferenceLostRetry')` reference in the error handler matches.
- Test mocks in `SummaryCreatePage.test.tsx` correctly simulate the axios-style `{ response: { data: { code: 40001 } } }` error shape and assert the exact zh-CN string.

## Verdict: APPROVED

No correctness, security, or build-blocking issues. The two P2 items (forward-to-matter parity and unsafe type cast) are non-blocking quality improvements. The persistence lifecycle is thorough, the button guards are consistent, and the error handling is well-tested.

yujiawei (approved):
# Code Review — PR #907 (octo-web)

Reviewed at head SHA `b455678029c6e525834b2e763a6615ac48546d2b`. Independent review with two additional advisory perspectives absorbed.

## Verdict: APPROVE

The three fixes are correct and well-targeted. All cross-file assumptions were independently verified against the current head. Findings below are non-blocking (P2/nit/enhancement) — none require changes before merge.

---

## Spec / scope compliance

Diff matches the stated scope exactly — no missing, extra, or diverging work:

1. **Persist `referencedTask`** — new helpers `readAgentChatReferenced` / `writeAgentChatReferenced` / `clearAgentChatReferenced` in `summaryHelpers.ts`, wired into all four lifecycle points in `SummaryCreatePage.tsx` (mount, enter-agent-mode restore, new-session clear, remove-reference clear, picker-select write, save-success clear), plus a friendly `40001` toast.
2. **Forward-to-chat fallback** — `SummaryDetailPage.tsx:1897` falls back to `personalResult?.content` when `detail.result` is null (agent summaries write only `personal_result`). Verified `PersonalResult.content: string` exists.
3. **Hide unsupported buttons on agent summaries** — edit / schedule / regenerate gated behind `detail.trigger_type !== TriggerType.AGENT`.

Cross-file verification (all confirmed on head):
- `TriggerType.AGENT` = `3`, imported in `SummaryDetailPage.tsx`, and `trigger_type` is a required field on `SummaryDetail`. The gating matches the pre-existing `continueRefine` button which already keys off `trigger_type === AGENT`.
- The `40001` branch correctly extracts `code` via the existing axios type-guard and sits cleanly beside the existing `40004` branch; the test mocks `createAgentSummary` (the function actually called by `handleSaveAsSummary`) — correct target.
- `referencedTask` is only ever read for `.task_id` and `.title`, so the minimal `{task_id, title}` snapshot restored via `as SummaryListItem` is functionally safe.

---

### P2 — Stale session can pair with a fresh reference (edge case)

`SummaryCreatePage.tsx:160-171` persists a new referenced task on `derivedFromTask` mount, but does not retire any pre-existing persisted `session_id`. Because both `session_id` and the new `referencedTask` live under the same shared `__workbench__` key (`agentChannelId()` returns `undefined`), the following narrow sequence can bind an old conversation to a new reference:

1. User has an **unsaved** agent chat for Task A (session_id persisted).
2. User triggers "continue refine" from Task B → new page mounts, overwrites the persisted reference with Task B, but the Task A `session_id` remains.
3. User **hard-refreshes** (loses the `derivedFromTask` prop) and re-enters agent mode → `enterAgentMode()` (`:582`) restores the old Task-A `session_id` *and* the Task-B reference together.
4. Saving associates Task A's conversation with a Task-B reference.

Mitigating factors that keep this at P2 rather than blocking:
- This is a **pre-existing structural property** of the shared single-key session design (introduced in #850) — `session_id` already had no task binding; this PR follows the same established pattern.
- The trigger requires an unsaved leftover session plus a refresh at exactly the wrong moment.
- The backend re-validates the reference by `task_id` (`canAccessTaskDB`), so this is a UX/attribution issue, not a security or hard-data-corruption one.

Suggested follow-up (not required for this PR): when `derivedFromTask` differs from the persisted session's task, clear the stale `session_id` at mount so the reference and conversation stay bound.

### P2 — Forward-to-chat can silently no-op during the `personalResult` load window

`SummaryDetailPage.tsx:3317` renders the forward button as soon as `detail.status === COMPLETED`, but for agent summaries `detail.result` is null and the fallback `personalResult` is fetched separately/asynchronously (`loadPersonalResult`, `:545`). If the user clicks in that brief window, `sourceContent` is empty and `handleForwardToChat` returns silently (`:1898`). Consider disabling the button (or showing a brief loading state) until fallback content is available. Low likelihood; degrades to the same no-op this PR otherwise fixes.

### nit — Mock handler `trigger_type` for agent tasks is `2`, should be `3`

`src/__mocks__/handlers.ts:40` sets `trigger_type: 2` (SCHEDULED) on the agent-summary create mock, so MSW-backed tests classify agent summaries as scheduled and cannot exercise the new `trigger_type === AGENT` gates end-to-end. This file is outside this PR's diff, so it doesn't block — but worth a one-line fix in a follow-up so the new gating gets integration coverage.

### enhancement — Consider telemetry on silent localStorage write failure

`writeAgentChatReferenced` (and the pre-existing `writeAgentChatSession`) swallow write failures silently. If storage is full/disabled, the user still hits the `40001` path on refresh — now with a friendly toast, which is the right degradation. A non-blocking telemetry counter in the `catch` would let you measure real-world failure rate. Optional.

---

## Tests

- New `agentChatReferenced.test.ts` (9 cases) thoroughly covers the helper: key scoping, missing-channel fallback, malformed JSON, wrong-shape payloads, null-clear, per-channel isolation, and extra-field stripping. Good.
- New `SummaryCreatePage` test covers the `40001` friendly-toast branch (asserts the friendly copy, not the raw backend message, and `result === false`). Correct.
- Coverage gap (not blocking): no test exercises the persist→refresh→restore lifecycle or the three `SummaryDetailPage` gating branches. The unit surface is solid; the integration path is untested.

## Coverage note

Verified statically against the checked-out head: `TriggerType` values, `trigger_type`/`personalResult`/`PersonalResult.content` types, all `referencedTask` call sites, the `40001` handler, and the mount/enter-agent-mode lifecycle. Not verified: the backend `agent_summary.go` reference-resolution and personal-result schema (backend not in this repo — the PR description's account is plausible and consistent with the frontend fallback), and I did not re-run `pnpm test` / `pnpm build` locally (dependencies not installed on the review checkout); the author reports 457/457 tests and build green.

yujiawei (changes requested):
# Code Review — PR #907 (octo-web)

Reviewed at head SHA `b455678029c6e525834b2e763a6615ac48546d2b`. This supersedes my earlier note on this PR: on closer tracing of the persistence lifecycle I found one blocking issue that I had under-rated. Correcting the verdict to **Changes Requested**.

## Verdict: CHANGES_REQUESTED

The three fixes are on the right track and the helper code is clean and well-tested, but the new reference-persistence path can silently bind an agent chat to the wrong referenced task. That must be resolved before merge. Everything else below is non-blocking.

---

## 🔴 Blocking (P1)

### Stale `session_id` can pair with a freshly-persisted `referencedTask` → wrong derivation attribution

`SummaryCreatePage.tsx:160-171` (`componentDidMount`) persists the new `referencedTask` on a `derivedFromTask` mount **without retiring any previously persisted `session_id`**. Both live under the same shared `__workbench__` key (the full-page entry has no channel; `agentChannelId()` returns `undefined`, and `agentChatReferencedKey` / the session key share the `__workbench__` fallback segment — `summaryHelpers.ts:29,85`). This makes the following sequence reachable:

1. User has an **unsaved** agent chat referencing Task A → `session_id_A` + `reference_A` persisted under `__workbench__` (this is exactly the "退出不丢" state the feature is designed to keep, so a leftover unsaved session is a normal, common condition — not an exotic one).
2. User opens "continue refine" from Task B → new page mounts with `derivedFromTask = B`, overwrites the persisted reference with **B**, but leaves `session_id_A` in storage (state `sessionId` is still `''`, so at this instant state is consistent).
3. User **hard-refreshes** before sending (the `derivedFromTask` route prop is lost) and re-enters agent mode → `enterAgentMode()` (`:582-594`) restores `session_id_A` **and** reference **B** together, and `loadAgentHistory(session_id_A)` reloads Task-A's conversation.
4. UI now shows Task-A's conversation under a "referenced: Task B" badge. On save (`handleSaveAsSummary`, `:724-728`) the request carries `session_id_A` + `referenced_task_ids = [B]`.

Impact: the summary produced from conversation A is recorded as derived from Task B, and backend origin-fallback may borrow B's origin metadata. The backend `canAccessTaskDB` check only validates that the user *can access* the referenced task — it does **not** detect that the conversation and the reference belong to different workflows, so the corruption is silent and uncaught. This PR is what introduces the vector: before it, there was no persisted reference to mismatch against; `session_id` alone carried no derivation attribution.

Fix direction (either is sufficient):
- On `derivedFromTask` mount, clear the persisted `session_id` (and any restored history) so a derived workflow always starts a clean session bound to the new reference; **or**
- Persist `session_id` and `referencedTask` atomically and verify on restore that they belong together (drop/reset one if they don't).

Please add a regression test for: existing unsaved session → continue-refine a *different* task → refresh before sending → save, asserting the saved `session_id` and `referenced_task_ids` are consistent.

---

## 💬 Non-blocking

### P2 — Forward-to-chat can silently no-op during the `personalResult` load window
`SummaryDetailPage.tsx:3317` renders the forward button as soon as `detail.status === COMPLETED`, but for agent summaries `detail.result` is null and the fallback `personalResult` is fetched asynchronously (`loadPersonalResult`, `:545`, gated on `summary_mode === BY_PERSON`). A click in that brief window leaves `sourceContent` empty and `handleForwardToChat` returns silently (`:1898`). Consider disabling / showing a loading state until fallback content is ready. Low likelihood.

### P2 (latent) — Sibling `handleForwardToMatter` lacks the same fallback
`handleForwardToMatter` (`:1943`) and `handleMatterSelected` (`:1961`) read only `detail.result?.content`, so they would always hit `noForwardContent` for agent summaries. This is **currently dead code** — the button is gated by `const SHOW_FORWARD_TO_MATTER = false` (`:58`) — so it is not user-reachable today. Worth applying the same `?? personalResult?.content` fallback (or a `trigger_type !== AGENT` guard) when that feature is re-enabled, so it doesn't ship broken.

### nit — Mock handler agent `trigger_type` is `2`, should be `3`
`src/__mocks__/handlers.ts:40` sets `trigger_type: 2` (SCHEDULED) on the agent-summary create mock while `TriggerType.AGENT === 3`, so MSW-backed tests classify agent summaries as scheduled and can't exercise the new agent gates end-to-end. Outside this PR's diff, so non-blocking — but fixing it would let integration tests cover the new gating.

### nit — Unsafe `as SummaryListItem` cast from a minimal snapshot
`SummaryCreatePage.tsx:590` casts the restored `{task_id, title}` to the full `SummaryListItem` (~15 fields). Safe today (only `.task_id` / `.title` are read), but the cast hides future reads of unset fields from the compiler. Prefer `Pick<SummaryListItem,'task_id'|'title'>` or a dedicated `ReferencedTaskRef` type.

### enhancement — Telemetry on silent localStorage write failure
`writeAgentChatReferenced` (like the pre-existing `writeAgentChatSession`) swallows write failures. If storage is full/disabled the user silently re-hits the `40001` path on refresh (now with a friendly toast — good degradation). A non-blocking telemetry counter in the `catch` would let you measure real-world failure rate. Optional.

---

## What's solid

- Forward-to-chat correctly sources `detail?.result?.content ?? personalResult?.content` — the right fallback for agent summaries (`detail.result` null by design), reusing the existing `splitSummaryText` send pipeline with no regression to the doc-forward path.
- Edit / schedule / regenerate are consistently gated by `trigger_type !== TriggerType.AGENT` at every render/menu-build site (`:2470`, `:3108`, `:3294`), so the actions are truly unreachable, mirroring the pre-existing `continueRefine` button which keys off `=== AGENT`. `TriggerType.AGENT === 3` and `trigger_type` (required on `SummaryDetail`) verified on head.
- The `40001` branch extracts `code` via the existing axios type-guard and sits cleanly beside the `40004` branch; i18n keys present in both locales.
- `agentChatReferenced.test.ts` (9 cases) is thorough: key scoping, missing-channel fallback, malformed JSON, wrong-shape payloads, null-clear, per-channel isolation, extra-field stripping. The `40001` toast test targets the correct function (`createAgentSummary`).

## Coverage note
Verified statically against the checked-out head: `TriggerType` values, `trigger_type` / `personalResult` / `PersonalResult.content` types, all `referencedTask` call sites, the `40001` handler, the `SHOW_FORWARD_TO_MATTER` flag state, and the full mount / enter-agent-mode / save lifecycle. Not verified: backend `agent_summary.go` reference-resolution and personal-result schema (backend not in this repo — the description's account is consistent with the frontend fallback); `pnpm test` / `pnpm build` not re-run on the review checkout (deps not installed) — author reports 457/457 tests and build green.

yujiawei (approved):
# Code Review — PR #907 (octo-web)

Re-review at head SHA `9e1f8b07ab9fe5a0c592ef27dc97626a1e566f2c`. This supersedes the prior Changes-Requested review, which was against `b455678…`. The new commit resolves the one blocking issue.

## Verdict: APPROVE

The previously-blocking P1 (cross-session contamination) is fixed correctly and covered by a targeted regression test. No blocking issues remain on the current head.

---

Commit `9e1f8b07` — "clear session_id on derivedFromTask mount to prevent cross-session contamination":

`SummaryCreatePage.tsx` `componentDidMount`, `derivedFromTask` branch, now atomically:
1. `clearAgentChatSession(this.agentChannelId())` — drops any leftover workbench `session_id`.
2. `setState({ sessionId: '', messages: [] })` — resets React state to match empty storage.
3. `writeAgentChatReferenced(...)` — persists the new reference (unchanged).

This guarantees `session_id` and `referencedTask` always belong together after a refresh — either both empty (fresh entry into a new derivation) or both the same in-progress chat. Verified: `clearAgentChatSession` is imported, and the branch is the only place a `derivedFromTask` reference is persisted, so there is no remaining path that leaves a stale session paired with a new reference.

## Blocking issue — RESOLVED

**Stale `session_id` paired with a fresh `referencedTask` → silent wrong-derivation attribution.** The reproducer (unsaved chat for Task A → continue-refine Task B → hard-refresh before send → save submits `session_id_A` + `referenced_task_ids=[B]`) no longer occurs: the leftover session is cleared at mount, so a refresh restores a clean empty session bound to B.

Regression coverage is exactly the requested scenario and is sound:
- **Positive test** — pre-seeds a stale `session_id` + reference for Task A, mounts with `derivedFromTask = B`, asserts the session key is gone, the reference key is B, and state has empty `sessionId`/`messages` + reference B.
- **Reverse guard** — bare workbench mount (no `derivedFromTask`) leaves both storage keys intact, protecting the #158/#161 resume-after-refresh path from regression.

The author reports the test was reverse-verified (reverting the fix fails the positive test), and `pnpm test` 459/459 + `pnpm build` green.

---

## Non-blocking (carried over — none block merge)

- **P2** — Forward-to-chat can silently no-op during the async `personalResult` load window (`SummaryDetailPage.tsx:3317` renders the button before the fallback content is ready). Consider disabling / loading state until content is available.
- **P2 (latent)** — Sibling `handleForwardToMatter` (`:1943`) lacks the same `personalResult` fallback, but is currently dead code (`SHOW_FORWARD_TO_MATTER = false`, `:58`). Apply the same fallback or a `trigger_type !== AGENT` guard when that feature is re-enabled.
- **nit** — Mock `handlers.ts:40` sets agent `trigger_type: 2`; should be `3` so MSW-backed tests can exercise the agent gates (outside this PR's diff).
- **nit** — `as SummaryListItem` cast from the `{task_id, title}` snapshot (`SummaryCreatePage.tsx`) is safe today but hides future field reads from the compiler; prefer `Pick<SummaryListItem,'task_id'|'title'>`.
- **enhancement** — Telemetry on silent localStorage write failure would let you measure the real-world rate of the degraded path.

## What's solid

- The P1 fix is minimal, atomic, and matches the persistence-lifecycle design.
- Forward-to-chat fallback (`detail?.result?.content ?? personalResult?.content`) is the correct source for agent summaries.
- Edit / schedule / regenerate consistently gated by `trigger_type !== TriggerType.AGENT` at every render/menu site; `continueRefine` uses `=== AGENT`. `TriggerType.AGENT === 3`, `trigger_type` required on `SummaryDetail` — verified on head.
- `40001` handler + i18n keys correct; helper test suite (now 11 cases across the two files) is thorough.

## Coverage note

Verified statically against head `9e1f8b07`: the new `componentDidMount` clearing logic and its import, the regression tests, plus (re-confirmed) `TriggerType` values, `trigger_type`/`personalResult`/`PersonalResult.content` types, all `referencedTask` call sites, the `40001` handler, and the `SHOW_FORWARD_TO_MATTER` flag. Not re-run locally: `pnpm test`/`build` (deps not installed on the review checkout) — author reports 459/459 + build green. Backend `agent_summary.go` reference-resolution not verifiable (not in this repo); consistent with the frontend fallback.

wanyiwang06:
@yujiawei @Jerry-Xin thanks for the reviews. Just pushed follow-up fixes:

**P1 (both of you flagged)** — cross-session `session_id` / `referencedTask` contamination on `derivedFromTask` mount.

Fixed at commit `9e1f8b0` (pushed before your latest reviews landed — hence the mismatch showing `commit_id=9e1f8b0` on yujiawei's CR while the review body still cites the pre-fix line numbers):

- `SummaryCreatePage.tsx` componentDidMount, `derivedFromTask` branch: now calls `clearAgentChatSession(this.agentChannelId())` and resets React state `sessionId=''`, `messages=[]` before writing the new referencedTask.
- Regression test `SummaryCreatePage.test.tsx` `clears leftover workbench session_id + messages when mounted with derivedFromTask` covers the exact "existing unsaved session about A → continue-refine B → refresh before sending → save" sequence you described. Reverse-verified: temporarily reverted the fix → the test fails immediately; re-applied → 459/459 pass.

**P2-1** (yujiawei) — forward-to-chat silent no-op during personalResult load window.

Fixed at commit `b017c4d` (just pushed): the forward-to-chat button now shows a disabled + loading state for agent summaries while `personalResult` is still fetching. Traditional workflow tasks have `detail.result.content` inline so they are never gated here.

**P2-2** (yujiawei) — `handleForwardToMatter` lacks the same fallback.

Also at `b017c4d`: both `handleForwardToMatter` and `handleMatterSelected` now use the same `detail.result?.content ?? personalResult?.content` fallback as `handleForwardToChat`. Currently dead code (SHOW_FORWARD_TO_MATTER is false), but this way it won't ship broken when the flag is flipped back on.

**Not addressed** (deferred with rationale in commit body):
- P2-3 mock `trigger_type: 2 → 3`: outside this PR's diff and mutating a shared MSW mock risks cascading test breakage — better as a separate cleanup.
- P2-4 unsafe `as SummaryListItem` cast: pure type nit, safe today because only .task_id/.title are read.
- P2-5 telemetry: no sink wired here yet.

Please re-review when convenient. Full test suite (459/459) and turbo build are green on the current head.

yujiawei (approved):
# Code Review — PR #907 (octo-web)

Re-review at head SHA `b017c4ddcd491e7f706d83ca1e897c3ab9e8efe0`. Supersedes my prior reviews (against `b455678…` and `9e1f8b07…`). All findings I raised are now resolved or explicitly deferred with sound rationale.

## Verdict: APPROVE

No blocking issues remain on the current head. The one P1 (cross-session contamination) was fixed in `9e1f8b07` with a targeted regression test; the two forward-content P2s were hardened in `b017c4dd`.

---

## Change history since first review

1. `9e1f8b07` — **P1 fix**: `componentDidMount` (`derivedFromTask` branch) now atomically clears the leftover workbench `session_id` and resets `sessionId`/`messages` before persisting the new reference, so a refresh can never restore an old session paired with a new reference. Regression tests added (positive + reverse guard for the #158/#161 resume path). Verified the import and that this is the only reference-persist path.
2. `b017c4dd` — **P2 hardening** on the forward-content surface:
   - Forward-to-chat button is now `disabled` (with `loading` during the fetch) for agent summaries while `personalResult.content` is not yet ready — closing the silent-no-op window. Traditional tasks have `detail.result` inline and are never gated. `personalLoading` state verified (set true at `loadPersonalResult` start, false on all completion paths).
   - `handleForwardToMatter` / `handleMatterSelected` now share the same `detail.result?.content ?? personalResult?.content` fallback, so they won't ship broken when `SHOW_FORWARD_TO_MATTER` is re-enabled (still dead code today).

## Deferred items — reviewed and accepted

- Mock `handlers.ts` `trigger_type: 2` → `3`: outside this PR's diff; deferring to a separate cleanup is reasonable (touching a shared MSW mock risks unrelated test churn).
- `as SummaryListItem` cast: pure type nit, no runtime impact (only `.task_id`/`.title` read).
- localStorage-write telemetry: needs a sink not wired here.

All three are non-blocking and correctly scoped out.

---

## ⚠️ Note on merge gate

`reviewDecision` currently remains `CHANGES_REQUESTED` because there is an **earlier active Changes-Requested review from another reviewer** that predates these fixes. A fresh Approve cannot override another reviewer's outstanding Changes-Requested — that reviewer needs to re-review (or dismiss their stale review) for the decision to flip. This Approve records that the code is, in my assessment, ready to merge on head `b017c4dd`.

## Coverage note

Verified statically at head `b017c4dd`: the P1 clearing logic + its regression tests, the forward-to-chat disabled/loading gating and `personalLoading` state wiring, the forward-to-matter/matter-selected fallback, and (re-confirmed) `TriggerType.AGENT === 3`, `trigger_type` required on `SummaryDetail`, all `referencedTask` call sites, the `40001` handler, and the `SHOW_FORWARD_TO_MATTER` flag. Not re-run locally: `pnpm test`/`build` (deps not installed on the review checkout) — author reports 459/459 tests + build green.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #907 (octo-web)

## Summary

This PR addresses several UX fast-follow items for agent-generated summaries in the dmworksummary package. Three concerns are tackled: (1) persisting `referencedTask` to localStorage alongside `session_id` so that refresh/re-enter no longer loses the reference and triggers a 40001 save failure; (2) adding a `personalResult.content` fallback to the forward-to-chat and forward-to-matter handlers, since agent workflow only writes `personal_result` (not `summary_result`), leaving `detail.result` null; and (3) hiding edit/schedule/regenerate buttons for agent summaries where the backend pipeline cannot replay them.

Overall the changes are well-scoped, well-tested, and the data flow is correctly wired. One P2 latent issue flagged for the forward-to-matter path.

## Verification

- ✅ **referencedTask persistence lifecycle** — `writeAgentChatReferenced` is called on mount-with-derivedFromTask (`SummaryCreatePage.tsx:178`), on picker select (`:890`); `clearAgentChatReferenced` is called on new-session (`:630`), save-success (`:749`), and remove-reference (`:675`). All paths are symmetric.
- ✅ **Cross-session contamination fix** — `clearAgentChatSession` runs before `writeAgentChatReferenced` in the derivedFromTask path (`:169-180`), atomically clearing stale session state. Test at `SummaryCreatePage.test.tsx:575-625` validates both the clear-on-derivedFromTask and preserve-on-bare-mount paths.
- ✅ **Forward-to-chat readiness gate** — Button at `:3322-3342` correctly disables while `personalResult.content` is loading for agent summaries. `handleForwardToChat` at `:1889` uses `detail.result?.content ?? personalResult?.content ?? ''` and early-returns on empty, which is consistent with the button's gating.
- ✅ **Trigger-type gating parity** — Edit (`:2475`), schedule (`:3113`), and regenerate (`:3299`) all check `detail.trigger_type !== TriggerType.AGENT`. Continue-refine button at `:3313` correctly renders only for `=== TriggerType.AGENT`. No asymmetric paths.
- ✅ **localStorage helpers robustness** — `readAgentChatReferenced` (`summaryHelpers.ts:96-111`) validates shape (number `task_id`, string `title`), catches JSON parse errors, and returns null on malformed input. Tests at `agentChatReferenced.test.ts:39-52` cover malformed JSON and wrong-shape payloads.
- ✅ **i18n parity** — Both `en-US.json` and `zh-CN.json` add `savedReferenceLostRetry` key consistently. The 40001 handler at `SummaryCreatePage.tsx:787` references this key via `t()`.

Static analysis only at head `b017c4ddcd491e7f706d83ca1e897c3ab9e8efe0`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. One P2 latent issue below.

### P2 — Forward-to-matter button lacks agent-content readiness gate (`SummaryDetailPage.tsx:3345`)

The forward-to-chat button at `:3322` correctly disables while `personalResult.content` is loading for agent summaries (the `waitingForFallback` IIFE). The forward-to-matter button at `:3345` does not mirror this gating — it only checks `this.state.forwardingToMatter`. Both handlers already use the same `detail.result?.content ?? personalResult?.content` fallback (`:1944` for chat, `:1965` for matter), and the comment at `:1949` explicitly says "they must stay in sync".

Currently `SHOW_FORWARD_TO_MATTER = false` (`:58`), so this is latent. When the flag is flipped back on, an agent summary whose `loadPersonalResult` hasn't returned yet would let the user click the matter button, `handleForwardToMatter` would see `content` as undefined, hit `!content?.trim()`, and show the `noForwardContent` toast — a misleading "no content" error when the content is merely still loading.

Fix: mirror the forward-to-chat IIFE pattern to add `waitingForFallback` disabled/loading gating on the matter button when re-enabling the flag.

## Data Flow Trace

- **`referencedTask` → save path**: `this.state.referencedTask` is set from `derivedFromTask` prop (mount) or `readAgentChatReferenced()` (enterAgentMode) or picker select → consumed at `:736` as `params.referenced_task_ids` → sent to `api.createAgentSummary`. If state is null (cleared by ✕ or new-session), the param is omitted — backend may 40001 if no other origin path exists. The 40001 catch at `:787` surfaces the friendly toast. **Wired correctly.**
- **`personalResult.content` → forward-to-chat**: `loadPersonalResult` (`:590`) fetches via `api.getPersonalResult(taskId)` → sets `this.state.personalResult` → consumed by button IIFE at `:3331` (`agentContentReady`) for gating and by `handleForwardToChat` at `:1897` for content. `personalLoading` state at `:594`/`:605`/`:639` gates the button's loading spinner. **Wired correctly.**
- **`detail.trigger_type` → button visibility**: Set from `loadDetail()` response → consumed at `:2475` (edit), `:3113` (schedule), `:3299` (regenerate), `:3313` (continueRefine), `:3330` (forward gate). All use `===`/`!==` `TriggerType.AGENT` consistently. **Wired correctly.**

## Blindspot Checklist

- **C1 (dual-path parity)**: Forward-to-chat vs forward-to-matter — **P2 gap flagged above** (latent behind feature flag). Edit/schedule/regenerate all consistently gate on `trigger_type !== AGENT`. Clear.
- **C2 (control-flow ordering)**: No nesting/ordering issues. `derivedFromTask` path runs once in `componentDidMount`, `enterAgentMode` runs on user-toggle. No double-action risk. Clear.
- **C3–C4 (auth boundary)**: N/A — no auth/permission changes.
- **C5 (build/runtime)**: N/A — no build/packaging changes.
- **C6 (governance/docs)**: N/A — no governance doc changes.

## Verdict: APPROVED

No P0/P1 issues. The referencedTask persistence lifecycle, cross-session contamination fix, forward-to-chat readiness gate, and trigger-type button gating are all correctly implemented and tested. The single P2 finding is latent behind `SHOW_FORWARD_TO_MATTER = false` and can be addressed when that flag is re-enabled.

[Octo-Q] verdict: APPROVE — no blocking issues found; one P2 latent (forward-to-matter button gating) behind disabled feature flag.

yujiawei (approved):
# Code Review — PR #904 (octo-web) — correction

Re-reviewed at head SHA `1dda64a1` (unchanged). **This corrects my immediately preceding Approve on the same head back to Changes requested.** I flipped too far, too fast on one finding; the reasoning below explains why I'm settling on blocking. The read-only fix that was the original blocker is genuinely resolved — the block is now a *different*, real issue this PR introduces.

## The one blocking issue: silent formatting loss on the new link-text edit path

`Toolbar.tsx:319-326` (the bubble menu's `applyEdit`, new in this PR):
```ts
if (!text || text === curText) chain.extendMarkRange('link').setLink({ href }).run()   // href-only: marks kept
else chain.insertContent({ type: 'text', text, marks: [{ type: 'link', attrs: { href } }] }).run()  // text changed: FLATTEN
```
When the display text is changed, the whole link range is rebuilt as a single text run carrying **only** the `link` mark. Any bold/italic/highlight on that text — or inline nodes (mentions, inline math) — are silently and permanently dropped (recoverable only by undo). This is a newly-added editing surface: before this PR the bubble could only edit the *href*, so this loss path did not exist there.

Why I'm blocking rather than filing it as a P2 tradeoff (reversing my prior comment):
- It is **silent** — no warning, no confirm.
- The fix is **cheap** — merge the existing marks in the range with the link mark instead of replacing with a bare link run.
- The "it matches the toolbar link editor from #793 (`Toolbar.tsx:1741`)" argument shows the author followed an existing pattern, but that flatten branch is not a documented, intentional tradeoff (the only comment there justifies the mark-*preserving* branch). Precedent for silent loss isn't a reason to add another instance of it — if anything #793 likely shares the latent bug and deserves its own follow-up.

Requested change: on a text change, carry over the compatible marks/structure of the original linked range (or restrict text editing to links whose content is a single plain-text run), and add a regression test ("edit a bold link's text → bold is preserved").

## Scope correction (for the audit trail)

A first-pass automated summary framed this as a "two-track" PR including a `ThreadCreateDialog` / `ThreadService` refactor across `dmworkbase`, and marked those as verified-fine. **Those files are not in PR #904** — they belong to **#902 (`refactor(thread): unify thread create flow`), already merged to `main`**, which is simply the parent commit of this PR's head. PR #904 changes only 4 files, all under `packages/docs/src/editor/`. That portion of the first-pass verification does not apply here and did not factor into this verdict.

## Confirmed resolved / correct (verified against source)

- **Read-only bypass (the prior 🔴) is genuinely fixed.** Guarded at the source (`mathExtended.ts:56` `dblclick` bails on `!isEditable`), plus `onEdit` / `applyEdit` / `updateMathAttrs`, plus `shouldShow`. `mathReadOnly.test.tsx` (4 cases) locks the NodeView and event-handler layers. This closes the earlier blocker.
- **No attribute loss on LaTeX edit** — TipTap `updateInlineMath`/`updateBlockMath` (v3.22.2) spread `{...node.attrs, latex}` and no-op on node-type mismatch; `fontSize`/`color` survive, non-math nodes are never rewritten.
- KaTeX-only display + double-click-to-modal removes the MathLive focus/centring conflict; colour/size via `setNodeMarkup` re-pinning the `NodeSelection` is the right fix.
- No listener leaks (NodeView `dblclick` removed on destroy; `octo-math-edit` removed on effect cleanup).

## Non-blocking follow-ups

- P2 — modal confirm targets the live selection, not a captured `pos` (`Toolbar.tsx:462-474`). Backdrop blocks local drift and the command's node-type guard makes the worst case a no-op, but under collaborative editing a remote move/delete while the modal is open can drop the edit. Pass `updateInlineMath({ latex, pos })` with the pos known at open time.
- P2 — link display text can't be cleared (`Toolbar.tsx:319`): `textDraft.trim()` routes empty input to the keep-old-text branch.
- P2 — `:has()` z-index fix needs Chromium 105+/Safari 15.4+/Firefox 121+ (`styles.css:4015`); older engines ignore it. Fine if out of support.
- P3 — same-href adjacent links don't reset `editing` (`Toolbar.tsx:292`, keyed on href) — a submitted text edit could land on the wrong link. Guard on range identity. (Fixing the blocking issue and this together is sensible.)
- Doc nit — description says "formula editing only, 3 files"; the diff also adds link-text editing (4 files). Update the description.

## Verdict

The formula fix is correct and closes the prior read-only blocker. But this PR's new link-text edit path silently discards a link's other inline formatting — a cheap-to-fix, silent formatting-loss regression on a surface this PR introduces. **Changes requested** on that one point; the P2/P3 items are non-blocking follow-ups.

yujiawei (approved):
# Code Review — PR #907 (octo-web)

Independent re-review at head SHA `b017c4ddcd491e7f706d83ca1e897c3ab9e8efe0`. The previous Changes-Requested review was against `b455678…` (P1 cross-session contamination); the prior Approve was against `9e1f8b07…`. The current head adds one commit (`b017c4dd`) — the P2 forward-button hardening requested in the earlier round. Everything was re-verified against the current head with a correct merge-base.

## Verdict: APPROVE

The three fixes are correct and well-targeted, the previously-blocking P1 is resolved and regression-tested, and the final commit implements exactly the requested P2 hardening. No P0/P1 issues remain on the current head.

---

The diff matches the stated three-fix scope exactly; no missing, extra, or diverging work.

1. **Persist `referencedTask`** — new helpers `readAgentChatReferenced` / `writeAgentChatReferenced` / `clearAgentChatReferenced` in `summaryHelpers.ts:524-590`, wired into all lifecycle points in `SummaryCreatePage.tsx` (mount `160-182`, enter-agent restore `593-606`, picker select `881+`, remove-reference, new-session `handleNewSession`, save-success `747-756`). Storage stays consistent with `session_id`.
2. **Forward-to-chat fallback** — `handleForwardToChat` (`SummaryDetailPage.tsx:1890-1897`) now falls through to `personalResult.content` when `detail.result` is null (agent workflow only writes `personal_result`). Verified this is not dead code: agent summaries are `summary_mode: 2` (BY_PERSON), so `loadPersonalResult` is invoked (`SummaryDetailPage.tsx:544-548`) and `personalResult` is populated.
3. **Hide unsupported buttons on agent summaries** — Edit (`2475`), Schedule (`3113`), Regenerate (`3299`) are all gated on `trigger_type !== TriggerType.AGENT`. Reverse-verified: traditional (MANUAL) summaries retain all three.

The final commit's P2 hardening (disable + loading state on the agent forward button during the async `personalResult` load window; mirror the same fallback into `handleForwardToMatter` / `handleMatterSelected`) is within scope — it directly resolves the prior review's P2 note and does not add unrequested behavior. Note the matter path is behind `SHOW_FORWARD_TO_MATTER = false`, so it is dormant today; keeping it in sync is correct defensive hygiene for a future flip-on.

## 2. Code Quality — Approved

The previously-blocking P1 is fixed correctly:

- `componentDidMount` (`SummaryCreatePage.tsx:160-182`) now atomically clears the leftover workbench `session_id` (`clearAgentChatSession`), resets `sessionId: ''` / `messages: []`, then writes the new reference — so a refresh-before-first-message can no longer restore a stale session paired with a fresh reference. Covered by a targeted regression test (`SummaryCreatePage.test.tsx`, "derivedFromTask cross-session isolation") plus a reverse-guard confirming a bare workbench entry is untouched.

Forward-button gating logic (`SummaryDetailPage.tsx:3322-3344`) is sound:
- `waitingForFallback = isAgent && !detail.result?.content?.trim() && !agentContentReady` — only agent summaries with no inline result and no loaded fallback content are disabled; traditional summaries are never gated.
- The `loading` spinner is scoped to `personalLoading`, so once the load settles the spinner clears. If an agent summary genuinely has no forwardable content the button stays disabled — a strict improvement over the previous silent no-op.

Error handling: the 40001 branch (`SummaryCreatePage.tsx:785-788`) surfaces actionable copy instead of the raw backend message, with a test asserting the friendly string and `false` return.

Helper robustness: `readAgentChatReferenced` defends against disabled/quota-limited `localStorage` (try/catch), malformed JSON, and wrong-shape payloads (typed guards on `task_id`/`title`), returning `null` on all failure paths. Nine dedicated helper tests cover key scoping, per-channel isolation, and the minimal-snapshot contract.

Verification: `pnpm --filter=@dmwork/summary test` → **459/459 pass across 33 files** at this head.

## 3. Overall verdict: APPROVE

Spec ✅ AND Quality Approved.

- The `personalResult` fallback premise (agent workflow writes `personal_result`, not `summary_result`) lives in the backend, which is out of scope for this repo/diff — it could not be re-verified from source here and rests on the already-merged #158/#161 data-flow contract. Worth a lightweight backend integration assertion at some point so a future backend change that starts writing `summary_result` for agent tasks doesn't silently leave these frontend gates stale.
- The inline gating IIFE in the header render is a little dense; a small `renderForwardToChatButton()` helper would read more cleanly, but this is purely stylistic.

## 5. Coverage / blind spots

- **Verified:** the full diff at the current head; the P1 fix lifecycle; the forward-fallback data path (confirmed agent = BY_PERSON so `personalResult` loads); traditional-summary regression paths; helper edge cases; full package test suite green.
- **Not verifiable from this repo (out of scope):** backend `agent_summary.go` behavior that underpins the fallback premise (see suggestion above).
- **Environment note:** an isolated `tsc --noEmit` reports a cascade of errors rooted in `Could not find a declaration file for module 'react'` (a local type-resolution artifact) plus pre-existing rot in `SummaryDetailPage.participantReports.test.tsx` / `SummaryDetailPage.schedule.test.tsx` — both files are untouched by this PR and the rot predates it. Neither affects the changed files' correctness; the Vitest suite (the meaningful signal) is fully green.

Jerry-Xin (approved):
Re-review at head b017c4ddcd49: my stale-write 🔴 is fixed exactly as recommended. Flipping to APPROVE.

### ✅ Blocker resolved (byte-verified)

The referencedTask/session cross-session contamination is fixed. `componentDidMount` on `derivedFromTask` now clears the stale session before writing the new reference, and resets in-memory state atomically (SummaryCreatePage.tsx):

```
if (this.props.derivedFromTask) {
  clearAgentChatSession(this.agentChannelId());      // clear stale session FIRST
  this.setState({ mode: 'agent', referencedTask: ..., sessionId: '', messages: [] });
  writeAgentChatReferenced(this.agentChannelId(), { task_id, title });
}
```

This closes the exact scenario I flagged: previously the mount wrote the new reference but left the stale `session_id` persisted, so a refresh-before-first-message restored the "old session + new reference" mismatch (and `loadAgentHistory` re-hydrated the old messages), corrupting derivation on save. Now the session is cleared and reset (`sessionId: ''`, `messages: []`) alongside the reference write, so the two persisted values are always consistent — "继续优化" starts a clean session bound to the new reference.

### ✅ Still correct (carried from prior review)

- forward-to-chat fix (personalResult.content fallback) — unchanged, still correct, no #901/#903 regression.
- unsupported-button gating by `trigger_type === AGENT` — actions truly unreachable.

### 💬 Non-blocking

- 🟡 Mapping all backend `40001` to the lost-reference toast may misdiagnose other invalid-request cases sharing that generic code.

### ✅ Highlights

- The clear-session-on-derivedFromTask-mount fix keeps the persisted session and reference atomically consistent, with a regression test.

Verdict: APPROVE. Thanks for the precise fix.

PR #502, opened 2026-07-01
feat(web): render message reactions in chat (#111)

caster-Q:
## Summary

Render message reactions in chat (frontend of the three-end #111 reactions feature).

## Related Issue

#111

## Changes

- `Convert.toReactions`: aggregate the server's per-actor rows `{seq, uid, name, emoji, is_deleted}` into per-emoji groups `{seq, count, emoji, users[]}` — skip `is_deleted`, count = number of actors, seq = max within the group.
- `MessageRow` (and the legacy Base bubble) render a reaction chip row under the message body; `useMessageRow` forwards the data.
- `ConversationVM` handles the `syncMessageReaction` CMD by re-pulling the latest page and merging fresh reactions by message id. The fetch + merge + error-handling is extracted into a pure `reactionMerge.ts` seam (match/overwrite/changed-gate; a sync failure is caught and non-fatal) so it is unit-testable without instantiating the VM.
- Documented a known platform-level limitation (not introduced here): `CMDSyncMessageReaction` carries no message_id, so a reaction on an already-loaded but off-latest-page message does not live-update until reload (user reactions share this). Proper fix is a separate follow-up.

## Testing

- [x] Unit tests added/updated
- [x] Manually verified

`vitest` — `Convert.reactions` (aggregation / is_deleted filter / max-seq / malformed), `reactionMerge` (merge + swallowed-failure path), `MessageRow`: all pass. Three-end e2e verified (bot reaction → chip rendered in chat).

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [x] Added tests for my changes
- [ ] Updated documentation
- [x] Ran \`pnpm i18n:check\` or confirmed the change does not affect UI copy
- [x] Followed commit message conventions (Conventional Commits)

> Draft: reactions is a three-end feature — ships with octo-server (endpoint) + openclaw-channel-octo (react action). Reaction chips show emoji + count only; no new translatable UI copy.

PR #929, opened 2026-07-21
feat(dmworkbase): message reactions in chat (UI + wiring behind feature flag)

an9xyz:
## Summary

Adds **message reactions** to chat (emoji + branded custom-emoji), entirely
gated by a **feature flag (default OFF)** — merging is a runtime no-op until the
server reaction contract lands and the flag is wired to appconfig. Demo data
currently flows through a local mock store and is deleted once the real API is
available.

Rebased onto latest `main`; single squashed commit.

## What's included

**UI components** (`packages/dmworkbase/src/ui/message/`)
- `MessageReactionSummary` — chip row under a message (always left-aligned,
  Slack-style, no bubble). `aggregate()` groups by `(type, key)`, filters
  `is_deleted`, dedupes uids, orders by first arrival, marks `hasMine`.
  Branded/`[token]` emoji resolve to images via `EmojiService` (gated by
  `isCustomEmoji`); unicode keeps its glyph. Types + 12 unit tests + Storybook.
- `MessageReactionPicker` — WeCom-style quick-pick (branded tokens + common
  emoji in one aligned grid) with a "more" entry that opens the full
  `EmojiPanel` (sticker tab hidden) in a floating popover; Esc-to-close + focus
  management.

**Wiring (all flag-gated)**
- `Messages/Text`: renders `<ReactionSlot>` under text content.
- `module.tsx`: registers the "React" context-menu entry (text messages).
- `EmojiToolbar` `EmojiPanel`: new `hideStickerTab` prop (+ tests) to reuse the
  full emoji panel without the sticker tab.
- `App.tsx` MittEvents, i18n (zh/en), `.i18n/scan-config`, `featureFlags`,
  `dmworkbase/src/vite-env.d.ts` (ImportMetaEnv typing).

## Enable in dev / preview

```js
localStorage.setItem('octo:ff:message-reaction', '1'); location.reload()
```
(honored only in DEV builds; production ignores it.)

## Coordination

- **Server contract**: `.context/message-reaction-api-spec.md` lists the 5
  blocking server capabilities (write/toggle, inline sync, single-message
  query, realtime CMD, error codes). Please align the **CMD event name** with
  the server reaction PR (this branch assumes `messageReaction`; an earlier
  draft used `syncMessageReaction`).
- Overlaps in scope with the earlier web read-path draft (PR #502) — the chip
  rendering here is a reusable component and can supersede the inline version.

## Known follow-ups (documented in code)
- Full "more" `EmojiPanel` doesn't yet highlight already-selected reactions
  (needs `selectedKeys` on the shared component).
- Full keyboard navigation inside `EmojiPanel` is a pre-existing shared-component
  limitation.

## Test plan
- [x] production build (`pnpm --filter @octo/web build`)
- [x] unit tests 49/49 (aggregate + EmojiPanel incl. new hideStickerTab cases)
- [x] `pnpm i18n:check`
- [x] stylelint (reaction CSS) 0 warnings
- [x] dmworkbase scoped tsc clean
- [ ] end-to-end with real backend + flag on (blocked on server contract)


lml2468 (dismissed):

审查锚定 head `161ebf3a8651`。23 文件 +1908/−1,`dmworkbase` 消息 reaction(emoji + 品牌自定义表情),**整体由 feature flag gate、默认 OFF**。已跑 reaction 测试(49 例)、核实 flag 隔离与 XSS 面。

### ✅ 两道闸门均通过

**规格符合**:与描述一致——UI + 接线全部 flag-gated,默认关,服务端契约落地前合入是**运行时 no-op**;demo 数据走本地 mock store,真 API 就绪后删除。已 rebase 到 main、单 squash commit。无 over-build(唯一 −1 是 EmojiPanel 加 `hideStickerTab` prop 的既有行改动)。

**关键:合入 main 零用户影响 —— 逐点坐实**:
- `isMessageReactionEnabled()`:`DEFAULT_ENABLED=false`;localStorage 覆盖**严格 `import.meta.env.DEV` gate**——生产构建只认 `DEFAULT_ENABLED`,用户无法手动打开看到 mock 伪造数据。设计正确。
- **两个 gate 都真短路**:`Messages/Text` `{isMessageReactionEnabled() && <ReactionSlot/>}`——关时不渲染;`module.tsx` context-menu `if (!isMessageReactionEnabled()) return null`——关时不注册右键项。默认 OFF 下 UI 面完全不出现。
- 右键入口刻意仅放行 `contentType === text`(与 ReactionSlot 仅接 TextCell 同口径,注释说明放开时需同步渲染 summary)——避免 store 更新无处回显。范围决定合理。

**代码质量**:
- **`aggregateReactions` 严谨**:剔除 `isDeleted===1`、按 `(type::key)` 聚合、同 uid 去重保留首次 record(名字不跳动)、用户内排序 seq→createdAt→输入序、组间按 firstSeq 升序(已有组不因新增参与者跳序)、`currentUid` 空→`hasMine` 恒 false、**不改入参**。多键 tie-break 完整。
- **XSS 安全**:reaction UI **无 innerHTML / dangerouslySetInnerHTML sink**;自定义/品牌 token 经 `EmojiService`(先 `isCustomEmoji` 判定)解析为 ``,url 走 React 属性自动转义,无法引导到 javascript:;unicode emoji 保留字形(textContent 路径)。
- **i18n**:zh/en +6 key 对齐,`.i18n/scan-config` 同步。
- **vite-env.d.ts**:补 `ImportMetaEnv` 类型,支撑 `import.meta.env.DEV` 类型安全。

### ✅ 验证

- `aggregate.test.ts`(12)+ `EmojiPanel.test.tsx`(`hideStickerTab`)本地 **49 passed**。
- flag DEV-gate、两处 gate 短路、text-only 右键范围、aggregate 不变式、无 XSS sink 均 byte 核实。
- gate REVIEW_REQUIRED(CI pending),mergeable。

### 🟡 非阻塞(作者已在描述中明确)

- **服务端契约未落地**:写/toggle/内联同步/单条查询/实时 CMD/错误码 5 项待 octo-server;现走 mock store(刷新不持久、他人不可见)。这正是 flag 默认关的原因——合入不影响任何人。生产化时需把数据源换成 `message.reactions` + 写接口、`DEFAULT_ENABLED` 换 appconfig 读取。
- mockStore / stories 是 demo 脚手架,契约就绪后按描述删除。

### 结论

功能完整由 flag gate、默认 OFF、DEV-only 覆盖,合入 main 为运行时 no-op(两处 gate 真短路已坐实);aggregate 聚合/排序/去重严谨、无 XSS 面、测试覆盖核心逻辑。这是教科书式的「flag 后合入未完成功能」做法。无阻塞。**APPROVE**。

Jerry-Xin (dismissed):
The feature belongs in octo-web, but the disabled feature still introduces permanent production-side behavior.

### 🔴 Blocking

- 🔴 Critical — The feature flag does not provide the promised production runtime no-op. Importing `reactionPickerOverlay` installs capture-phase `contextmenu` and `mousedown` listeners immediately, even when the flag is disabled, and never removes them. This affects every production session and leaks duplicate handlers during module reloads. Register pointer tracking only while the picker feature is enabled/open, with explicit cleanup, or pass the originating coordinates through the context-menu API. [ReactionPickerOverlay.tsx:40](/Users/mlamp/.openclaw/tmp/pr-review-15w9g946/repo/packages/dmworkbase/src/ui/message/MessageReactionPicker/ReactionPickerOverlay.tsx:40), [module.tsx:896](/Users/mlamp/.openclaw/tmp/pr-review-15w9g946/repo/packages/dmworkbase/src/module.tsx:896)

### 💬 Non-blocking

- 🟡 Warning — Picker/overlay behavior has no automated tests despite substantial imperative lifecycle logic: repeated opening, Escape cleanup, focus restoration, outside-click handling, position clamping, and selected-state toggling. Add focused tests, especially ensuring listeners and portal containers are removed.
- 🟡 Warning — `localStorage.getItem()` can throw a `SecurityError` when storage is blocked. Because this function executes during message rendering, that can break the chat view in affected development/preview environments. Wrap storage access defensively. [featureFlags.ts:26](/Users/mlamp/.openclaw/tmp/pr-review-15w9g946/repo/packages/dmworkbase/src/Service/featureFlags.ts:26)
- 🔵 Suggestion — `Math.abs(-2147483648)` remains negative, so the mock hash can theoretically produce negative array indexes and malformed seeded reactions. Normalize with an unsigned conversion such as `h >>> 0`. [mockStore.ts:25](/Users/mlamp/.openclaw/tmp/pr-review-15w9g946/repo/packages/dmworkbase/src/ui/message/MessageReactionSummary/mockStore.ts:25)

### ✅ Highlights

- Aggregation behavior has strong unit coverage, including deletion filtering, deduplication, ordering, and immutability.
- The shared `EmojiPanel` change is narrowly scoped and tested against configuration changes.
- Presentation components are separated from aggregation and include useful Storybook coverage.

Jerry-Xin (dismissed):
The feature belongs in octo-web, but as written the "disabled" feature still introduces permanent production-side behavior, which contradicts the PR's stated runtime no-op guarantee.

### 🔴 Blocking

- 🔴 The feature flag does not deliver the promised production runtime no-op. `module.tsx` imports `reactionPickerOverlay` at the top level (unconditionally), and importing `ReactionPickerOverlay.tsx` installs capture-phase `contextmenu` and `mousedown` document listeners at module-evaluation time — before and regardless of `isMessageReactionEnabled()`. So every production session (flag OFF) permanently carries two global capture-phase listeners, and HMR/module reloads can leak duplicates. The listeners themselves are passive (they only record `lastPointer`), so there is no UI/mock-data leak, but this still breaks the "merging is a runtime no-op until the flag is wired" contract the PR is built on. Fix: install pointer tracking only when the picker is actually enabled/opened, with explicit teardown, or thread the originating coordinates through the context-menu API instead of a module-level singleton. `packages/dmworkbase/src/ui/message/MessageReactionPicker/ReactionPickerOverlay.tsx:40`, `packages/dmworkbase/src/module.tsx:85`

### 💬 Non-blocking

- 🟡 `localStorage.getItem()` can throw `SecurityError` when storage is blocked/partitioned. This runs during message rendering (`isMessageReactionEnabled()` is called in `Messages/Text` render and the context-menu builder), so a throw could break the chat view in affected dev/preview environments. Wrap the storage access in try/catch. `packages/dmworkbase/src/Service/featureFlags.ts:26`
- 🟡 No automated tests for the imperative picker/overlay lifecycle (repeated open, Escape cleanup, focus restore, outside-click close, viewport clamping, selected-state toggle). Given how much imperative DOM/portal/listener logic lives in `ReactionPickerOverlay.tsx`, add focused tests — especially asserting that listeners and the portal container are removed on close. Current tests cover only `aggregate()` and the `EmojiPanel hideStickerTab` prop.
- 🟡 Five `reaction.*` i18n keys are added but three (`reaction.failed`, `reaction.noPermission`, `reaction.unavailable`) are not referenced anywhere in shipped code yet (only `reaction.moreUsers`, `reaction.more`, and `module.contextMenus.react` are used). Fine if intentionally staged for the write-path, but worth confirming they're not dead keys.
- 🔵 `Math.abs(-2147483648)` stays negative, so the mock hash can theoretically yield a negative index and malformed seed data. Use an unsigned conversion (`h >>> 0`). DEMO-only, low impact. `packages/dmworkbase/src/ui/message/MessageReactionSummary/mockStore.ts:25`

### ✅ Highlights

- `aggregate()` has strong, well-structured unit coverage: deletion filtering, per-uid dedup with first-seen name, seq/createdAt/insertion-order sorting, stable cross-group ordering, and input immutability.
- The shared `EmojiPanel` change (`hideStickerTab`) is narrowly scoped and tested against remote-config gray-release flips — good defense against the sticker tab leaking back in.
- Clean separation between presentational components (`MessageReactionSummary` / `MessageReactionPicker`) and aggregation/wiring, plus useful Storybook coverage.
- Flag gating at the two render/entry points (`Messages/Text` render and the `contextmenus.reaction` builder) is correct; the context-menu entry is also correctly restricted to text messages to match where `ReactionSlot` renders.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #929 (octo-web)

## Summary

This PR adds message reaction UI (summary bar + picker popover + context-menu entry) to the chat view in `dmworkbase`, gated behind a dev-only feature flag (`isMessageReactionEnabled()`, default OFF). Data currently flows through a local mock store — the server-side reaction contract is not yet landed, so this is a UI-first wiring PR intended to let the interaction be previewed and screenshotted before the backend is ready. The architecture is clean: pure-UI components (`MessageReactionSummary`, `MessageReactionPicker`) are decoupled from the demo data layer (`mockStore`), and the feature flag ensures zero production exposure. Documentation is thorough — every demo-only seam carries a `⚠️ DEMO-ONLY` comment with explicit production replacement notes.

No P0/P1 issues. Two P2 observations below for the production handoff.

## Verification

- ✅ **Feature flag gates all entry points** — `isMessageReactionEnabled()` guards the context-menu registration (`module.tsx:896`), the `ReactionSlot` render in `TextCell` (`Messages/Text/index.tsx:217`), and the `reactionPickerOverlay` is only reachable through these two gated paths. Production builds always see `DEFAULT_ENABLED = false`.
- ✅ **localStorage override is DEV-only** — `import.meta.env.DEV` check at `featureFlags.ts:26` ensures `localStorage` cannot enable the feature in production. Vite statically replaces `import.meta.env.DEV` at build time.
- ✅ **Context-menu contentType gate** — `module.tsx:905` restricts the "React" menu to `MessageContentType.text` only, preventing reaction store writes for message types that don't render a `ReactionSlot` (which would cause invisible state changes).
- ✅ **EmojiPanel.hideStickerTab prop** — the `!this.props.hideStickerTab &&` guard at `EmojiToolbar/index.tsx:527` correctly short-circuits `stickerCustomEnabled`, and the existing test suite now covers the sticker-tab suppression including the `configChangeListener` transition case.
- ✅ **aggregate() input safety** — handles `undefined`, `null`, and empty arrays (tested); filters `isDeleted === 1`; deduplicates `uid` per group; preserves first-seen name.
- ✅ **Picker overlay lifecycle** — `close()` properly unmounts via `ReactDOM.unmountComponentAtNode`, removes the container, removes the `keydown` listener, and restores focus to `prevFocus`.
- ✅ **i18n parity** — both `en-US.json` and `zh-CN.json` add the same 5 keys (`module.contextMenus.react`, `reaction.failed`, `reaction.moreUsers`, `reaction.more`, `reaction.noPermission`, `reaction.unavailable`).

Static analysis only at head `161ebf3a86510baa63bedb8892197218f6394b8f`; build and tests not executed in this environment.

## Findings

### P2 — Global pointer-tracking listener lives forever (`packages/dmworkbase/src/ui/message/MessageReactionPicker/ReactionPickerOverlay.tsx:49`)

The `contextmenu` and `mousedown` listeners are attached at module load (inside a `typeof document !== "undefined"` guard) and never removed, firing on every right-click and mouse-down across the entire app to record `lastPointer`. This is needed because the context-menu `onClick` callback doesn't receive the original `MouseEvent`, so `openAtLastPointer` needs the last known coordinates.

Not a blocker — the handler is two assignments and has no side effects. When productionizing, consider either passing the event through the context-menu callback (if the shared `ContextMenus` component supports it) or attaching the tracker on first `open()` and removing it on `close()`.

### P2 — `messageID || clientMsgNo` key may diverge after server ack (`packages/dmworkbase/src/ui/message/MessageReactionSummary/ReactionSlot.tsx:73`)

The `messageId` prop is derived from `message.messageID || message.clientMsgNo` at `Messages/Text/index.tsx:219`. In the demo mock store this is fine, but flagging for the production handoff: if a message is sent optimistically using `clientMsgNo` and the server later assigns a different `messageID`, the reaction store key and the UI lookup key can drift, leaving orphaned reactions that can't be toggled or cleared.

Documented as demo-only — the fix direction is to normalize the key after server ack (use `messageID` when available, fall back to `clientMsgNo` only for unsent messages, and migrate any mock/server reactions keyed on the old ID).

- Feature flag architecture: `DEFAULT_ENABLED = false` with `import.meta.env.DEV` localStorage override matches the project's existing pattern (`useNewUI` in `Messages/Text/index.tsx`).
- Context-menu priority 1200 fits between forward (1150) and multiSelect (1000) — consistent with existing ordering.
- `aggregate()` group sorting: `firstSeq` (min seq per group) prevents existing groups from jumping when a later participant joins. Tested at `aggregate.test.ts:127-142`.
- `ReactionPickerOverlay` focus management: `prevFocus` is captured on open, restored on close, and the Escape handler uses capture phase at `document` level (correct — the trigger element is already unmounted).
- `vite-env.d.ts` declaration for `ImportMetaEnv` is self-contained and does not conflict with `vite/client` types when they're available.
- `.i18n/scan-config.json` correctly excludes `data.ts` and `mockStore.ts` from i18n scanning — both are demo-only datasets, not user-facing copy.
- Three `reaction.*` i18n keys (`reaction.failed`, `reaction.noPermission`, `reaction.unavailable`) are defined but not yet consumed — intentional, these are placeholders for the production error-handling layer.

## Verdict: APPROVED

Clean, well-documented demo wiring behind a properly gated feature flag. No correctness, security, or build-breaking issues. The two P2 items are forward-looking notes for the production handoff, not blockers for this PR.





[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 observations for the production handoff (global pointer listener lifecycle, messageId key stability after server ack). Feature flag correctly prevents all production exposure.

yujiawei (approved):
# Code Review — PR #929 (octo-web)

Reviewed at head `161ebf3a86510baa63bedb8892197218f6394b8f` against `main`. This adds message reactions (chip summary, quick-pick picker, context-menu entry) entirely behind a feature flag that defaults OFF, with demo data flowing through a local in-memory mock store.

## 1. Spec compliance

**Result: ✅**

Everything the PR description promises is present and, importantly, nothing user-facing ships by default:

- `Service/featureFlags.ts` — `DEFAULT_ENABLED = false`; every entry point (summary render, context-menu item, picker) is gated by `isMessageReactionEnabled()`.
- UI components (`MessageReactionSummary`, `aggregate.ts` + 12 unit tests, `MessageReactionPicker`) are self-contained and pure.
- Wiring: `Messages/Text` renders `<ReactionSlot>`, `module.tsx` registers the "React" context menu (text-only), `EmojiPanel` gains the `hideStickerTab` prop (+ 3 new tests), plus i18n (zh/en), `App.tsx` mitt event, `.i18n/scan-config`, and `vite-env.d.ts`.

No missing scope, no unrequested user-facing feature/flag/interface, no divergence from the stated flag-gated approach. The one over-build item is minor and non-functional (see nits).

## 2. Code quality

**Result: Approved**

Verified cross-file assumptions against the repo at this SHA:

- `WKApp.mittBus` / `emojiService` / `loginInfo` all exist with the shapes used; the new `"message-reaction-mock-updated"` mitt event is declared in `MittEvents`.
- `EmojiService.isCustomEmoji` / `getImage` exist and are optional-chained defensively in `ReactionSlot`; the `isCustomEmoji`-first guard correctly avoids turning unicode glyphs into bitmap images.
- The custom-token assets referenced by `data.ts` (`custom_mission/action/taste/shangfang.png`) exist under `apps/web/public/emoji/`, and the token keys match `BUILTIN_CUSTOM_EMOJIS`.
- `registerMessageContextMenus` returning `null` is safely filtered by `EndpointManager.invokes` (`if (result)`), so the flag-off / non-text path is correct.
- `t("base.reaction.moreUsers", { values: {...} })` matches the i18n namespace + interpolation contract; keys are registered under the `base` namespace.
- `ReactDOM.render` / `unmountComponentAtNode` is correct for this package (React 17), consistent with existing usage.

### Security note (this PR was flagged security-sensitive)

The classification appears keyword-triggered; the actual security surface is minimal (no auth, no tokens, no network writes — reactions are local mock only). The one security-relevant control is correct: the `localStorage` override is gated behind `import.meta.env.DEV`, so production users cannot force-enable the feature and see fabricated reaction data. Worth a human sanity-check that the production build correctly strips `DEV`, but the code is right.

### P2 / nits (non-blocking)

- **Unused reserved i18n keys.** `reaction.failed`, `reaction.noPermission`, `reaction.unavailable` are added to both locale files but not referenced anywhere yet (only `reaction.more` and `reaction.moreUsers` are used). Fine as Phase-2 scaffolding, but flag them for cleanup if the server contract slips.
- **Global pointer listeners attach at import time.** `ReactionPickerOverlay.tsx` registers capturing `contextmenu` + `mousedown` document listeners at module load (via the eager import in `module.tsx`), regardless of flag state. They're cheap (coordinate bookkeeping) and never open UI while the flag is off, but strictly they run even when the feature is disabled. Consider lazy-attaching on first `open()` when this is productionized.
- **Documented known limitation.** The full "more" `EmojiPanel` doesn't highlight already-selected reactions and silently toggles-off on re-click; already called out in code comments and the PR body. Acceptable for the demo phase.

## 3. Overall verdict

**APPROVE.** No P0/P1 issues. The feature is inert by default, the mock/demo layer is clearly marked for deletion, and the cross-file integration points check out against the current tree. The items above are cleanups to track before the flag flips on, not merge blockers.

yujiawei (changes requested):
# Code Review — PR #929 (octo-web) — updated verdict

Reviewed at head `161ebf3a86510baa63bedb8892197218f6394b8f` against `main`.

**This supersedes my earlier Approve on this PR.** On reconsideration — and in agreement with another reviewer who independently blocked on the same point — I'm moving to Changes Requested for one small, concrete reason. The feature is otherwise clean and well-built; this is a targeted fix, not a rewrite.

## 1. Spec compliance

**Result: ❌ (one deviation from the PR's own stated contract)**

The PR's merge justification is explicit: *"merging is a runtime no-op until the server reaction contract lands."* The diff does not fully honor that guarantee.

- **Deviation — feature is not a runtime no-op when the flag is OFF.** `module.tsx` imports `reactionPickerOverlay` at top level unconditionally (`packages/dmworkbase/src/module.tsx:85`), and importing `ReactionPickerOverlay.tsx` attaches capture-phase `contextmenu` and `mousedown` listeners on `document` at module-evaluation time, before and regardless of `isMessageReactionEnabled()` (`packages/dmworkbase/src/ui/message/MessageReactionPicker/ReactionPickerOverlay.tsx:40`). So every production session (flag OFF) permanently carries two global capture-phase listeners, and HMR/module reloads can leak duplicates.

The listeners themselves are passive — they only record `lastPointer` coordinates, so there is **no UI exposure, no mock-data leak, no security issue, and negligible performance cost**. But the PR is built on the "inert until wired" promise, and permanent global side-effecting subscription at import time contradicts it. Because the whole safety-to-merge-now argument rests on that guarantee, I'm treating this as a spec deviation rather than a nit.

**Required fix (small):** attach the pointer tracker only when the feature is enabled/opened (lazy-attach on first `open()` with explicit teardown in `close()`), or thread the originating coordinates through the context-menu callback API instead of a module-level singleton. With that, flag-OFF becomes a true no-op.

## 2. Code quality

**Result: Approved (no P0/P1 beyond the spec item above)**

The rest of the implementation is solid; verified against the tree at this SHA:

- `aggregateReactions` is well-structured and thoroughly unit-tested: `isDeleted` filtering, per-uid dedup keeping first-seen name, seq→createdAt→insertion-order sorting, stable cross-group ordering via `firstSeq`, `hasMine` correctness, input immutability.
- `WKApp.mittBus` / `emojiService` / `loginInfo` exist with the shapes used; the new `"message-reaction-mock-updated"` mitt event is declared in `MittEvents`.
- `EmojiService.isCustomEmoji`/`getImage` are correctly optional-chained; the `isCustomEmoji`-first guard rightly avoids turning unicode glyphs into bitmaps.
- Custom-token assets (`custom_mission/action/taste/shangfang.png`) exist under `apps/web/public/emoji/` and keys match `BUILTIN_CUSTOM_EMOJIS`.
- Context-menu handler returning `null` is safely filtered by `EndpointManager.invokes` (`if (result)`); entry is correctly restricted to `MessageContentType.text` to match where `ReactionSlot` renders.
- `t("base.reaction.moreUsers", { values })` matches the i18n namespace + interpolation contract.
- `ReactDOM.render`/`unmountComponentAtNode` is correct for this package (React 17); `close()` unmounts, removes the keydown listener, removes the container, and restores focus.
- No XSS sink: token images render via `` (React-escaped), unicode keeps textContent.

### Security note (PR flagged security-sensitive)

The classification appears keyword-triggered; the real security surface is minimal (no auth/token/network-write; reactions are local mock only). The one meaningful control is correct: the `localStorage` override is gated behind `import.meta.env.DEV`, so production users cannot force-enable the feature. Worth a human sanity-check that the production build strips `DEV` as expected — but the code is right.

### Non-blocking (P2)

- **`localStorage.getItem` can throw `SecurityError`** in partitioned/blocked-storage contexts. It's called during message render and the context-menu builder, so a throw could break the chat view in affected DEV/preview environments. Wrap in try/catch. (`packages/dmworkbase/src/Service/featureFlags.ts:26`)
- **`Math.abs(-2147483648)` stays negative** — the mock hash can theoretically produce a negative index. Use `h >>> 0`. DEMO-only, low impact. (`packages/dmworkbase/src/ui/message/MessageReactionSummary/mockStore.ts:25`)
- **`messageID || clientMsgNo` key drift** — if a message is sent optimistically under `clientMsgNo` and the server later assigns a different `messageID`, the reaction key can drift. Fine for the mock store; flag for the production handoff (normalize the key after ack). (`ReactionSlot.tsx` / `Messages/Text/index.tsx:219`)
- **Unused reserved i18n keys** — `reaction.failed`, `reaction.noPermission`, `reaction.unavailable` are added but not yet referenced. Fine as Phase-2 scaffolding; confirm they don't become dead keys.
- **No tests for the imperative picker/overlay lifecycle** (repeated open, Escape cleanup, focus restore, outside-click close, viewport clamping). Given the amount of imperative DOM/portal/listener logic, add focused tests asserting listeners and the portal container are removed on close.

## 3. Overall verdict

**CHANGES_REQUESTED.** No correctness/security/data-loss/build-break issues. The single blocker is that the flag-OFF path is not the runtime no-op the PR guarantees, due to eager module-load side effects — a few-line fix. Once pointer tracking is attached lazily (or coordinates are threaded through the context-menu API), this is an Approve. The P2 items are cleanups to track before the flag flips on.

lml2468 (approved):
## Review Verdict（复审 @ 9f9e4a52）：可合并（APPROVE）—— 维持 APPROVE

复审锚定新 head `9f9e4a52`。上轮 Jerry-Xin/yujiawei/Steve 提的 🔴「flag-OFF 时仍在 document 上常驻全局监听器,违背『flag OFF = 运行时 no-op』契约」**已干净修复**,byte 核实:

### ✅ flag-off 监听器泄漏 🔴 —— 已修

- **上轮问题**:`module.tsx` 顶层 `import reactionPickerOverlay` 会在模块求值时就在 document 上装 capture 阶段 contextmenu + mousedown 监听,与 `isMessageReactionEnabled()` 无关 → flag 关的每个生产会话都常驻这两个监听。
- **修法(byte 核实)**:
  - 监听器从 module-eval 顶层**抽进 `enablePointerTracking()` 函数**(`pointerTrackingOn` 幂等 guard),import 时**不再注册任何监听**;
  - `module.tsx:924` 仅在 `if (isMessageReactionEnabled())` 内调 `enablePointerTracking()` → **flag OFF = 零全局副作用**,兑现契约;
  - 新增对称的 `disablePointerTracking()`(HMR / 关闭功能 / 测试 teardown)。
- 这正是 Steve 建议的修法(把 overlay 的副作用注册放到 flag 判断之后)。

### ✅ 验证

- 新增 `ReactionPickerOverlay.test.tsx`(+105)本地 **6 passed**,精确覆盖:open/close 单容器、重复 open 不泄漏、Escape 关闭 + keydown 监听拆除、focus 复原、**`enablePointerTracking` 幂等**;
- `aggregate.test.ts` + `EmojiPanel.test.tsx` **49 passed**(与上次一致);
- 上次已核的其余 flag 门控(Messages/Text `{flag && <ReactionSlot/>}`、context-menu `if(!flag) return null`)、aggregate 逻辑、无 XSS、i18n 双 locale 均保持;
- 落后 main 8 commit 但 `git merge-tree` **0 冲突**,可合。

### 结论

上轮唯一阻断(flag-off 监听器泄漏)已通过「抽 `enablePointerTracking` + flag 内调用 + 对称 disable」干净修复,并补了 6 例 overlay 生命周期回归测试。合入 main 现在真正是 flag-OFF 运行时 no-op。此前的 flag-gate / aggregate / XSS-safe 均保持。**维持 APPROVE**。感谢 Jerry-Xin/yujiawei/Steve 抓出这条副作用。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #929 (octo-web)

## Summary

This PR introduces message reactions — a chat feature allowing users to add emoji or sticker reactions to messages — gated behind `isMessageReactionEnabled()`, which defaults to `false` in production with a `localStorage` override available only in dev. The feature ships a pure UI summary component, an imperative picker overlay, and a demo-only mock data store so the full interaction flow is exercisable before a real server contract lands. Alongside the reactions work, the PR also refactors `MeInfo` from a declarative Sections-builder to a panel-based layout with explicit state management and `UserService` API calls, and fixes the docs editor math extension to dispatch edit events via double-click rather than editing in place.

The reactions surface is well-separated into concerns: types, aggregation logic, pure summary component, imperative overlay, demo glue, and the feature-flag gate itself. The MeInfo refactor is a meaningful quality improvement over the previous implementation, which silently swallowed errors and used raw `axios` instead of the shared `APIClient`.

## Verification

Static analysis only at head `9f9e4a526bdc14c02291c22d4bff9cb237658feb`; build and tests not executed in this environment.

- ✅ **Feature flag gates all three production surfaces** — ReactionSlot render at `Messages/Text/index.tsx:218`, context-menu item registration at `module.tsx`, and `enablePointerTracking()` call in the same file are all guarded by `isMessageReactionEnabled()`.
- ✅ **Flag default is false in production** — `featureFlags.ts` sets `DEFAULT_ENABLED = false` and only honours the `localStorage` override when `import.meta.env.DEV` is true.
- ✅ **Aggregation correctness** — `aggregate.ts` filters `isDeleted === 1`, rejects records missing `uid` or `reactionKey`, deduplicates by `uid` (keeping first record), sorts users deterministically by `seq → createdAt → insertionOrder`, and sorts groups by `firstSeq` ascending.
- ✅ **API client paths replaced** — `UserService.updateCurrentUser` and `uploadUserAvatar` both route through `APIClient.shared`, inheriting auth interceptors and base URL handling.
- ✅ **Mounted lifecycle guard** — `MeInfo` tracks `_isMounted` and gates all `setState` calls on unmount.
- ✅ **Error re-throw in updateMyInfo** — the new `vm.tsx` re-throws errors from `updateMyInfo`, unlike the old code which silently caught and swallowed them via `.catch`.
- ✅ **i18n parity** — all 13 new keys are present in both `en-US.json` and `zh-CN.json`.
- ✅ **ViewTransition skip-target fix** — `Onboarding` correctly sets the skip transition target before `persistDismissed()` and defers intro close to `onFinished`, preventing the snapshot leak.

## Findings

No P0/P1 issues. Two P2 items and two nits — all behind the default-disabled feature flag.

### P2 — No teardown for pointer-tracking capture listeners (`ReactionPickerOverlay.tsx:46-79`)

`enablePointerTracking()` installs two global capture-phase listeners (`contextmenu` and `mousedown`) on `document` to track the last pointer position for right-click menu placement. `disablePointerTracking()` is exported and removes them, but nothing in the production code path calls it. When the flag is on, these listeners persist for the lifetime of the page. Since the flag requires a page refresh to toggle and defaults off, this is not reachable in production today — but the exported cleanup function going unused suggests the teardown contract was never wired up. Worth adding a `disablePointerTracking()` call in module teardown (if one is ever added) or at least calling it when the flag is toggled off via `localStorage` in dev.

### P2 — Overlay singleton has no module-destroy cleanup (`ReactionPickerOverlay.tsx:100-130`)

`reactionPickerOverlay` is a module-level singleton with `open()`/`close()` methods. There is no cleanup path that runs when the `dmworkbase` module unmounts or the app tears down. If the picker is open when the module is destroyed (e.g. hot-reload in dev, or an unexpected unmount), the overlay container div and its React tree would remain orphaned in the DOM. In practice this is low-risk since the overlay is only open during active user interaction and the flag is off in production — but the missing cleanup makes the singleton contract incomplete.

### Nit — Absolute image paths in `DEFAULT_TOKENS` (`MessageReactionPicker/data.ts:18-25`)

Custom emoji image URLs in `DEFAULT_TOKENS` use absolute paths like `/emoji/custom_shangfangbaojian.png`. These would break if the app is deployed under a subpath (e.g. `/app/emoji/...`). This follows the existing pattern used elsewhere in the codebase for static assets, so it is not a regression — but worth noting for future subpath deployment work.

### Nit — Legacy `ReactDOM.render` deprecation (`ReactionPickerOverlay.tsx:115-130`)

The overlay uses `ReactDOM.render` and `ReactDOM.unmountComponentAtNode`, which are the correct APIs for the project's React 17 (`package.json` shows `"react": "^17.0.2"`). However, these become legacy APIs in React 18 and emit deprecation warnings. When React is eventually upgraded, the overlay should migrate to `createRoot` / `root.unmount()`. Not a change I'd block on — just flagging for the upgrade path.

- `isMessageReactionEnabled()` gates all three production surfaces consistently (render, context menu, pointer tracking).
- `featureFlags.ts` correctly isolates `localStorage` override to `import.meta.env.DEV` only, with a try/catch for sandboxed iframe environments.
- `aggregate.ts` data flow is correct: deleted records filtered → required-field validation → uid dedup → deterministic sort → grouping by `${type}::${key}`.
- `reactionMockStore` is memory-only and self-contained — no persistence, no network calls, no side effects beyond `mittBus.emit`.
- `UserService` uses `APIClient.shared` for both `PUT user/current` and `POST user/avatar/{uid}`, inheriting auth interceptors and base URL.
- `uploadUserAvatar` has a 60-second timeout, appropriate for large image uploads.
- `MeInfo` `_isMounted` guard correctly prevents `setState` calls on unmounted component.
- `MeInfo.saveName` keeps the name editor open on error (does not reset `editingName`), giving the user a chance to retry or cancel — a real UX improvement over the old behaviour.
- `Onboarding` viewTransition fix correctly separates `onTransition` (set skip target) from `onFinished` (close intro), preventing the snapshot leak that persisted when users skipped.
- `mathExtended.ts` double-click `CustomEvent` dispatch correctly removes the in-place MathLive editing while preserving read-only KaTeX display; `readOnly` guard on the dblclick handler prevents spurious dispatches.
- `EmojiToolbar` `hideStickerTab` prop correctly defaults to `false`, so existing callers of `EmojiPanel` are unaffected.
- `RoutePage` `className` prop is additive and does not affect existing layout.

## Verdict: APPROVED

The PR is well-structured and the reactions feature is solidly gated behind a default-disabled flag with a clean separation of concerns. The MeInfo refactor is a meaningful quality improvement — proper error propagation, shared `APIClient` usage, and mounted lifecycle management are all real wins over the previous implementation. The two P2 items (listener teardown and overlay singleton cleanup) are worth tracking for the production wiring phase but are not reachable behind the current default-off flag, so they do not block merge.

yujiawei (approved):
# Code Review — PR #929 (octo-web)

**Scope reviewed:** full diff vs merge-base `b4b487f4` (24 files, +2044/-1), head SHA `9f9e4a52`. Cross-referenced `Service/featureFlags.ts`, `Service/EmojiService.ts`, `EndpointCommon.tsx`, and the existing context-menu registrations. This PR was flagged security-sensitive, so the flag-gating and any user-content rendering path were checked with extra care. Two independent review passes (a line-level adversarial pass and a whole-repo reasoning pass) were folded into the findings below.

## 1. Spec compliance

**Spec: ✅**

The stated contract is "message reactions UI + wiring, entirely behind a feature flag (default OFF), a runtime no-op until the server contract lands, demo data via a local mock store." The diff matches that:

- `DEFAULT_ENABLED = false`, and the only runtime override (`localStorage octo:ff:message-reaction`) is guarded by `import.meta.env.DEV`, so production genuinely cannot turn it on. Verified in `featureFlags.ts:28`.
- All entry points are gated: the summary slot (`Messages/Text/index.tsx:220`), the "React" context-menu item (`module.tsx:896`, returns `null` when the flag is off), and the global pointer tracker (`module.tsx:923`, only `enablePointerTracking()` when enabled). With the flag off there is no visible UI, no global listener, and the mock store is never touched — a true user-facing no-op.
- Mock store / seed data / stories are clearly marked `DEMO-ONLY` and scoped to be deleted when the real API lands, matching the PR description and the API spec doc.
- No over-build: no new network calls, no persisted state, no changes outside the reaction surface + the small `hideStickerTab` prop on the shared `EmojiPanel`.

Nothing missing, nothing beyond scope, no divergence from the agreed approach.

## 2. Code quality

**Quality: Approved**

Security review is clean:
- No `dangerouslySetInnerHTML`. Reaction icons render as either a text glyph or `` where `url` comes from `EmojiService`, which whitelists URL schemes (`resolveUrl`, `EmojiService.ts:384` — only `http(s)`, protocol-relative, and `data:image/`; everything else is coerced to a relative path). No image-src injection surface.
- User-facing names flow into React text/`title` nodes (auto-escaped), not into HTML, and `moreUsers` interpolation is via the i18n value bag, not string concatenation into markup.
- Event listeners are symmetric: the mitt subscription in `ReactionSlot` is cleaned up on unmount, and pointer tracking has an idempotent `enable`/`disable` pair.

### Findings (all non-blocking — P2 / nits)

- **P2 — `ReactionPickerOverlay.tsx:174` uses `ReactDOM.render` (a detached React root).** Any component mounted inside the popover that depends on React Context would lose it. For the current usage this is benign (`EmojiPanel` uses `WKApp` singletons + the `t()` import, not `useContext`/`contextType`), so there is no live bug today — but if the picker later hosts context-dependent UI, prefer `createPortal` from the app root. Note: the existing `EmojiToolbar` already uses `createPortal`, so there is a local precedent to follow.

- **P2 — pointer tracking is wired once at module init (`module.tsx:923`).** This is correct for the current flag (which only flips via a DEV localStorage change + reload). But the code comments plan to migrate the flag to dynamic `WKApp.remoteConfig`; if that migration lands without also attaching the tracker on the enable transition, right-click reactions would open at `(0,0)`. Please carry this into the productionization checklist (re-evaluate `enablePointerTracking()` when the flag source becomes dynamic).

- **P2 — message-id key stability (`Messages/Text/index.tsx:222`, `mockStore.ts:51`).** The slot is keyed `message.messageID || message.clientMsgNo`. A just-sent message can start with `clientMsgNo` and later acquire `messageID`, changing the store key and making a self-added reaction appear to vanish (and a different seeded set appear). This only affects the demo mock store, but the same identity choice will matter for the real datasource wiring — pick one stable id for the message lifetime.

- **P2 (forward-looking, contract-dependent) — tombstone ordering in `aggregate.ts:42`.** `aggregate` drops `isDeleted === 1` up front and keeps the *first* record seen per `(type,key,uid)`. If the future server sync can deliver, for one identity, an earlier active record and a later soft-deletion in the same payload, the deletion would be discarded and a stale reaction would linger. The current mock never produces active+tombstone pairs, so this is not reachable today — but when the real contract lands, resolve per-identity by latest `seq`/timestamp *before* filtering deletions, and add a test for the active-then-tombstone case.

- **nit — context-menu endpoint is registered unconditionally and gates inside the handler (`module.tsx:896`).** This matches the existing pattern for `forward`/`reply`/`multiSelect`, and the handler returns `null` when the flag is off, so behavior is correct; noting only that the registry entry itself exists regardless of the flag (the meaningful global side-effect, the document listeners, is properly flag-gated).

Two review passes also flagged a potential focus-steal in the overlay's `ref={(el) => el?.focus()}` and a context-loss concern with the detached root; on inspection `EmojiPanel` has no autofocused search input (its only `<input>` is a hidden file picker) and no context consumers, so neither is a live defect here — recorded for awareness only.

### Coverage / blind spots
- The reaction **server contract** (write/toggle, inline sync, single-message query, realtime CMD, error codes) is not in this PR — payload uniqueness and tombstone semantics can't be confirmed until it lands; the P2s above are the items to re-verify at that point.
- Production **runtime config** path (the eventual `remoteConfig.messageReactionEnabled`) is not yet wired; only the DEV override was verifiable.
- Not exercised end-to-end with a real backend (the PR marks this as blocked on the server contract).

## 3. Verdict

**APPROVED.** No P0/P1. The feature is default-OFF, DEV-only, mock-backed, and dead-code-eliminated in production, so merging is safe as a runtime no-op. The P2 items are all productionization follow-ups to address when the server contract is wired — none blocks this merge.

Jerry-Xin (approved):
Approve. Re-reviewed at live head. The prior CHANGES_REQUESTED is superseded — the flag-off module-eval side-effect is now gated.

## Resolved (was blocking)
Previously, `module.tsx` imported `reactionPickerOverlay`/`enablePointerTracking` at the top level and the overlay module registered capture-phase `document` `contextmenu`+`mousedown` listeners at module-eval, so those global listeners were installed regardless of `isMessageReactionEnabled()`.

Now this is correctly gated:
- `ReactionPickerOverlay.tsx` performs no `document.addEventListener` at module-eval. Module scope only declares state (`const lastPointer = { x: 0, y: 0 }`, `let pointerTrackingOn = false`). The listener installation is wrapped in an idempotent `enablePointerTracking()`:
  ```
  export function enablePointerTracking(): void {
    if (pointerTrackingOn || typeof document === "undefined") return
    document.addEventListener("contextmenu", trackPointer, true)
    document.addEventListener("mousedown", trackPointer, true)
    pointerTrackingOn = true
  }
  ```
- `module.tsx` only calls it behind the flag:
  ```
  if (isMessageReactionEnabled()) {
    enablePointerTracking();
  }
  ```
- The `reactionPickerOverlay` singleton constructor registers no document listeners; keydown/overlay listeners are added only in `render()` at user-action time and torn down in `close()`. A symmetric `disablePointerTracking()` is provided for HMR/teardown.

Net: flag OFF → no document listeners at module-eval → runtime no-op in production sessions, as intended.

## Non-blocking (follow-ups, not gating this merge)
- localStorage access in the DEV flag path should be wrapped in try/catch to stay resilient in privacy-mode / storage-disabled browsers.
- Add lifecycle tests for the picker overlay: `enablePointerTracking()` idempotency, `disablePointerTracking()` symmetry, and `render()`/`close()` listener add/remove balance.
- Prune the unused `reaction.*` i18n keys if they are not referenced by the demo path, or wire them up.

## Highlights
- The gating comment explicitly documents the "flag OFF = runtime no-op" contract and why listeners are not installed at module load.
- `enablePointerTracking()`/`disablePointerTracking()` are idempotent and symmetric, with `typeof document === "undefined"` SSR guards.
- Reaction context-menu entry itself is double-gated (flag + text-only content type), matching the ReactionSlot rollout scope.

PR #815, opened 2026-07-16
feat(login): refine registration and mobile download entry points

botshen:
## Summary

Simplifies the Web login experience by making registration directly discoverable, removing the completed login-migration notice flow, and replacing desktop mobile-download navigation with QR-code popovers.

## Related Issue

Not linked.

## Changes

- Add a persistent registration action below the primary Octo SSO login button and derive its URL from the active provider environment.
- Remove the obsolete login migration notice, acknowledgement state, remote config switch, and associated copy/styles.
- Refresh the login brand logo, headline, supporting copy, footer attribution, responsive layout, and language selector.
- Add accessible Android and iOS QR-code download popovers, Android GitHub Releases fallback, Storybook stories, and regression tests.
- Keep the Android QR target on the existing same-origin `/download/dmwork.apk` endpoint; publishing a newer APK remains separate release work.

## Architecture / Module Boundary

- Affected module(s): `@octo/login`, `@octo/base` remote configuration
- New or changed user-visible entry point: yes; registration is now directly available on the login page, and Android/iOS download buttons open QR popovers
- Shared layer touched: service
- If shared code changed, impact scope: removes consumption of the obsolete `suppress_login_migration_notice` Web remote-config field
- Duplicate entry point checked: yes; the old registration actions inside the migration notice were removed

## Testing

- [x] Unit tests added/updated
- [ ] Manually verified

Commands run after rebasing onto the latest `Mininglamp-OSS/octo-web:main`:

- `pnpm --filter @octo/login test` — 19 files, 204 tests passed
- `pnpm i18n:check` — passed
- `VITE_API_URL=http://127.0.0.1 pnpm --filter @octo/web build` — passed

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [x] Added tests for my changes
- [ ] Updated documentation
- [x] Ran `pnpm i18n:check` or confirmed the change does not affect UI copy
- [x] Confirmed the change follows module ownership and does not add duplicate user-visible entry points
- [x] Described impact scope when shared components, messages, bridge, or services are changed
- [x] Followed commit message conventions (Conventional Commits)

PR #931, opened 2026-07-21
fix(login): use updater-backed mobile download links

botshen:
## Summary

- use the Android updater `url` for both the Android QR code and direct APK download action
- use `/api/v1/common/updater/ios/1.0.0` to generate the iOS QR code
- remove the direct “Open TestFlight” action while retaining the QR entry
- keep the hardcoded GitHub Releases fallback unchanged
- show a fixed-size loading state until the updater responds; never render a stale fallback QR code or download link
- show a retry state on updater failure, while keeping the Android GitHub entry available

## Testing

- `pnpm --filter @octo/login test -- src/__tests__/android_download_button.test.tsx src/__tests__/ios_download_button.test.tsx` (19 files, 209 tests)
- `pnpm i18n:check`
- `pnpm exec stylelint packages/dmworklogin/src/MobileDownloadPopover.css --config stylelint.config.mjs`
- `pnpm exec prettier --check` for the touched TypeScript files
- `VITE_API_URL=http://127.0.0.1 pnpm --filter @octo/web build`
- `git diff --check`

### DIFF
--- a/packages/dmworklogin/src/AndroidDownloadButton.stories.tsx
+++ b/packages/dmworklogin/src/AndroidDownloadButton.stories.tsx
@@ -13,7 +13,7 @@ const meta: Meta<typeof AndroidDownloadButton> = {
     docs: {
       description: {
         component:
-          "登录页 Android 下载入口。悬停入口展示当前部署域名下的 APK 下载二维码，并保留 GitHub 最新 Release 备用入口。",
+          "登录页 Android 下载入口。二维码与直接下载按钮共用 updater 接口返回的 APK 地址，并保留 GitHub 最新 Release 备用入口。",
       },
     },
   },
--- a/packages/dmworklogin/src/AndroidDownloadButton.tsx
+++ b/packages/dmworklogin/src/AndroidDownloadButton.tsx
@@ -1,27 +1,18 @@
 import React from "react";
 import { IconGithubLogo } from "@douyinfe/semi-icons";
-import { Popover } from "@douyinfe/semi-ui";
+import { Popover, Spin } from "@douyinfe/semi-ui";
 import { WKButton } from "@octo/base";
 import { QRCodeSVG } from "qrcode.react";
 import { loginT as t } from "./i18n";
 import {
   resolveMobileUpdaterUrl,
-  startMobileDownload,
   useMobileDownloadUrl,
 } from "./mobileDownloadUpdater";
 import "./MobileDownloadPopover.css";
 
-export const ANDROID_APK_PATH = "/download/dmwork.apk";
 export const ANDROID_UPDATER_PATH = "common/updater/android/1.0";
 export const ANDROID_RELEASES_URL =
   "https://github.com/Mininglamp-OSS/octo-android/releases/latest";
-
-export function resolveAndroidApkUrl(
-  origin = typeof window === "undefined" ? "" : window.location.origin
-) {
-  if (!origin) return ANDROID_APK_PATH;
-  return new URL(ANDROID_APK_PATH, `${origin.replace(/\/$/, "")}/`).toString();
-}
 
 export function resolveAndroidUpdaterUrl(apiUrl?: string) {
   return resolveMobileUpdaterUrl(ANDROID_UPDATER_PATH, apiUrl);
@@ -45,11 +36,8 @@ type PopoverHoverProps = Pick<
 export const AndroidDownloadPopoverContent: React.FC<PopoverHoverProps> = (
   hoverProps
 ) => {
-  const fallbackUrl = resolveAndroidApkUrl();
-  const { downloadUrl, resolveDownloadUrl } = useMobileDownloadUrl(
-    ANDROID_UPDATER_PATH,
-    fallbackUrl
-  );
+  const { status, downloadUrl, retry } =
+    useMobileDownloadUrl(ANDROID_UPDATER_PATH);
 
   return (
     <div
@@ -60,25 +48,42 @@ export const AndroidDownloadPopoverContent: React.FC<PopoverHoverProps> = (
     >
       <div
         className="wk-login-mobile-popover-qr"
+        data-status={status}
         role="img"
         aria-label={t("download.androidQrTitle")}
       >
-        <QRCodeSVG value={downloadUrl} size={104} />
+        {status === "loading" && (
+          <Spin aria-label={t("download.loadingAddress")} size="large" />
+        )}
+        {status === "error" && (
+          <div className="wk-login-mobile-download-state" role="alert">
+            <span>{t("download.addressLoadFailed")}</span>
+            <WKButton type="button" variant="ghost" size="sm" onClick={retry}>
+              {t("download.retry")}
+            </WKButton>
+          </div>
+        )}
+        {status === "ready" && <QRCodeSVG value={downloadUrl} size={104} />}
       </div>
       <strong className="wk-login-mobile-download-popover-title">
         {t("download.androidQrTitle")}
       </strong>
-      <a
-        className="wk-login-mobile-download-direct-link"
-        href={downloadUrl}
-        download
-        onClick={(event) => {
-          event.preventDefault();
-          void resolveDownloadUrl().then(startMobileDownload);
-        }}
-      >
-        {t("download.androidDirectDownload")}
-      </a>
+      {status === "ready" ? (
+        <a
+          className="wk-login-mobile-download-direct-link"
+          href={downloadUrl}
+          download
+        >
+          {t("download.androidDirectDownload")}
+        </a>
+      ) : (
+        <span
+          className="wk-login-mobile-download-direct-link"
+          aria-disabled="true"
+        >
+          {t("download.androidDirectDownload")}
+        </span>
+      )}
       <WKButton
         type="button"
         className="wk-login-android-popover-manual-download"
--- a/packages/dmworklogin/src/AndroidDownloadButton.tsx
+++ b/packages/dmworklogin/src/AndroidDownloadButton.tsx
@@ -49,8 +49,11 @@ export const AndroidDownloadPopoverContent: React.FC<PopoverHoverProps> = (
       <div
         className="wk-login-mobile-popover-qr"
         data-status={status}
-        role="img"
-        aria-label={t("download.androidQrTitle")}
+        role={status === "ready" ? "img" : undefined}
+        aria-label={
+          status === "ready" ? t("download.androidQrTitle") : undefined
+        }
+        aria-busy={status === "loading" ? true : undefined}
       >
         {status === "loading" && (
           <Spin aria-label={t("download.loadingAddress")} size="large" />
--- a/packages/dmworklogin/src/AndroidDownloadButton.tsx
+++ b/packages/dmworklogin/src/AndroidDownloadButton.tsx
@@ -1,9 +1,14 @@
 import React from "react";
 import { IconGithubLogo } from "@douyinfe/semi-icons";
 import { Popover } from "@douyinfe/semi-ui";
-import { apiFetchJson, WKApp, WKButton } from "@octo/base";
+import { WKButton } from "@octo/base";
 import { QRCodeSVG } from "qrcode.react";
 import { loginT as t } from "./i18n";
+import {
+  resolveMobileUpdaterUrl,
+  startMobileDownload,
+  useMobileDownloadUrl,
+} from "./mobileDownloadUpdater";
 import "./MobileDownloadPopover.css";
 
 export const ANDROID_APK_PATH = "/download/dmwork.apk";
@@ -18,46 +23,8 @@ export function resolveAndroidApkUrl(
   return new URL(ANDROID_APK_PATH, `${origin.replace(/\/$/, "")}/`).toString();
 }
 
-export function resolveAndroidUpdaterUrl(
-  apiUrl = WKApp.apiClient.config.apiURL
-) {
-  return `${apiUrl.replace(/\/?$/, "/")}${ANDROID_UPDATER_PATH}`;
-}
-
-function resolveSafeDownloadUrl(value: unknown) {
-  if (typeof value !== "string" || !value.trim()) return undefined;
-  try {
-    const url = new URL(value, window.location.origin);
-    if (url.protocol === "http:" || url.protocol === "https:") {
-      return url.toString();
-    }
-  } catch {
-    // Invalid updater responses fall back to the legacy same-origin APK path.
-  }
-  return undefined;
-}
-
-function startBrowserDownload(url: string) {
-  const link = document.createElement("a");
-  link.href = url;
-  link.download = "";
-  link.style.display = "none";
-  document.body.appendChild(link);
-  link.click();
-  link.remove();
-}
-
-export async function downloadLatestAndroidApk() {
-  let downloadUrl = resolveAndroidApkUrl();
-  try {
-    const result = await apiFetchJson<{ url?: unknown }>(
-      resolveAndroidUpdaterUrl()
-    );
-    downloadUrl = resolveSafeDownloadUrl(result?.url) ?? downloadUrl;
-  } catch {
-    // Keep the legacy same-origin APK path available when the updater fails.
-  }
-  startBrowserDownload(downloadUrl);
+export function resolveAndroidUpdaterUrl(apiUrl?: string) {
+  return resolveMobileUpdaterUrl(ANDROID_UPDATER_PATH, apiUrl);
 }
 
 export function openAndroidReleases() {
@@ -77,47 +44,55 @@ type PopoverHoverProps = Pick<
 
 export const AndroidDownloadPopoverContent: React.FC<PopoverHoverProps> = (
   hoverProps
-) => (
-  <div
-    className="wk-login-mobile-download-popover"
-    role="dialog"
-    aria-label={t("download.androidQrTitle")}
-    {...hoverProps}
-  >
+) => {
+  const fallbackUrl = resolveAndroidApkUrl();
+  const { downloadUrl, resolveDownloadUrl } = useMobileDownloadUrl(
+    ANDROID_UPDATER_PATH,
+    fallbackUrl
+  );
+
+  return (
     <div
-      className="wk-login-mobile-popover-qr"
-      role="img"
+      className="wk-login-mobile-download-popover"
+      role="dialog"
       aria-label={t("download.androidQrTitle")}
+      {...hoverProps}
     >
-      <QRCodeSVG value={resolveAndroidApkUrl()} size={104} />
+      <div
+        className="wk-login-mobile-popover-qr"
+        role="img"
+        aria-label={t("download.androidQrTitle")}
+      >
+        <QRCodeSVG value={downloadUrl} size={104} />
+      </div>
+      <strong className="wk-login-mobile-download-popover-title">
+        {t("download.androidQrTitle")}
+      </strong>
+      <a
+        className="wk-login-mobile-download-direct-link"
+        href={downloadUrl}
+        download
+        onClick={(event) => {
+          event.preventDefault();
+          void resolveDownloadUrl().then(startMobileDownload);
+        }}
+      >
+        {t("download.androidDirectDownload")}
+      </a>
+      <WKButton
+        type="button"
+        className="wk-login-android-popover-manual-download"
+        variant="ghost"
+        size="sm"
+        icon={<IconGithubLogo aria-hidden="true" />}
+        aria-label={t("download.openGithubReleases")}
+        onClick={openAndroidReleases}
+      >
+        {t("download.githubManualDownload")}
+      </WKButton>
     </div>
-    <strong className="wk-login-mobile-download-popover-title">
-      {t("download.androidQrTitle")}
-    </strong>
-    <a
-      className="wk-login-mobile-download-direct-link"
-      href={resolveAndroidApkUrl()}
-      download
-      onClick={(event) => {
-        event.preventDefault();
-        void downloadLatestAndroidApk();
-      }}
-    >
-      {t("download.androidDirectDownload")}
-    </a>
-    <WKButton
-      type="button"
-      className="wk-login-android-popover-manual-download"
-      variant="ghost"
-      size="sm"
-      icon={<IconGithubLogo aria-hidden="true" />}
-      aria-label={t("download.openGithubReleases")}
-      onClick={openAndroidReleases}
-    >
-      {t("download.githubManualDownload")}
-    </WKButton>
-  </div>
-);
+  );
+};
 
 export const AndroidDownloadButton: React.FC = () => {
   const [hoverVisible, setHoverVisible] = React.useState(false);
--- a/packages/dmworklogin/src/AndroidDownloadButton.tsx
+++ b/packages/dmworklogin/src/AndroidDownloadButton.tsx
@@ -1,12 +1,13 @@
 import React from "react";
 import { IconGithubLogo } from "@douyinfe/semi-icons";
 import { Popover } from "@douyinfe/semi-ui";
-import { WKButton } from "@octo/base";
+import { apiFetchJson, WKApp, WKButton } from "@octo/base";
 import { QRCodeSVG } from "qrcode.react";
 import { loginT as t } from "./i18n";
 import "./MobileDownloadPopover.css";
 
 export const ANDROID_APK_PATH = "/download/dmwork.apk";
+export const ANDROID_UPDATER_PATH = "common/updater/android/1.0";
 export const ANDROID_RELEASES_URL =
   "https://github.com/Mininglamp-OSS/octo-android/releases/latest";
 
@@ -15,6 +16,48 @@ export function resolveAndroidApkUrl(
 ) {
   if (!origin) return ANDROID_APK_PATH;
   return new URL(ANDROID_APK_PATH, `${origin.replace(/\/$/, "")}/`).toString();
+}
+
+export function resolveAndroidUpdaterUrl(
+  apiUrl = WKApp.apiClient.config.apiURL
+) {
+  return `${apiUrl.replace(/\/?$/, "/")}${ANDROID_UPDATER_PATH}`;
+}
+
+function resolveSafeDownloadUrl(value: unknown) {
+  if (typeof value !== "string" || !value.trim()) return undefined;
+  try {
+    const url = new URL(value, window.location.origin);
+    if (url.protocol === "http:" || url.protocol === "https:") {
+      return url.toString();
+    }
+  } catch {
+    // Invalid updater responses fall back to the legacy same-origin APK path.
+  }
+  return undefined;
+}
+
+function startBrowserDownload(url: string) {
+  const link = document.createElement("a");
+  link.href = url;
+  link.download = "";
+  link.style.display = "none";
+  document.body.appendChild(link);
+  link.click();
+  link.remove();
+}
+
+export async function downloadLatestAndroidApk() {
+  let downloadUrl = resolveAndroidApkUrl();
+  try {
+    const result = await apiFetchJson<{ url?: unknown }>(
+      resolveAndroidUpdaterUrl()
+    );
+    downloadUrl = resolveSafeDownloadUrl(result?.url) ?? downloadUrl;
+  } catch {
+    // Keep the legacy same-origin APK path available when the updater fails.
+  }
+  startBrowserDownload(downloadUrl);
 }
 
 export function openAndroidReleases() {
@@ -55,6 +98,10 @@ export const AndroidDownloadPopoverContent: React.FC<PopoverHoverProps> = (
       className="wk-login-mobile-download-direct-link"
       href={resolveAndroidApkUrl()}
       download
+      onClick={(event) => {
+        event.preventDefault();
+        void downloadLatestAndroidApk();
+      }}
     >
       {t("download.androidDirectDownload")}
     </a>
--- a/packages/dmworklogin/src/IOSDownloadButton.stories.tsx
+++ b/packages/dmworklogin/src/IOSDownloadButton.stories.tsx
@@ -13,7 +13,7 @@ const meta: Meta<typeof IOSDownloadButton> = {
     docs: {
       description: {
         component:
-          "登录页 iOS 二维码入口。悬停或点击入口只展示动态生成的 TestFlight 二维码，触发按钮本身不执行跳转。",
+          "登录页 iOS 二维码入口。悬停或点击入口只展示由 updater 接口地址生成的二维码，触发按钮本身不执行跳转。",
       },
     },
   },
--- a/packages/dmworklogin/src/IOSDownloadButton.tsx
+++ b/packages/dmworklogin/src/IOSDownloadButton.tsx
@@ -2,10 +2,11 @@ import React from "react";
 import { Popover } from "@douyinfe/semi-ui";
 import { QRCodeSVG } from "qrcode.react";
 import { loginT as t } from "./i18n";
+import { useMobileDownloadUrl } from "./mobileDownloadUpdater";
 import "./MobileDownloadPopover.css";
 
-// TestFlight 是公开固定 URL，不需要走 updater 接口
 export const IOS_DOWNLOAD_URL = "https://testflight.apple.com/join/uPrdCcy3";
+export const IOS_UPDATER_PATH = "common/updater/ios/1.0.0";
 
 type PopoverHoverProps = Pick<
   React.HTMLAttributes<HTMLDivElement>,
@@ -14,35 +15,34 @@ type PopoverHoverProps = Pick<
 
 export const IOSDownloadPopoverContent: React.FC<PopoverHoverProps> = (
   hoverProps
-) => (
-  <div
-    className="wk-login-mobile-download-popover"
-    role="dialog"
-    aria-label={t("download.iosQrTitle")}
-    {...hoverProps}
-  >
-    <div
-      className="wk-login-mobile-popover-qr"
-      role="img"
-      aria-label={t("download.iosQrLabel")}
-    >
-      <QRCodeSVG value={IOS_DOWNLOAD_URL} size={104} />
-    </div>
-    <strong className="wk-login-mobile-download-popover-title">
-      {t("download.iosQrTitle")}
-    </strong>
-    <a
-      className="wk-login-mobile-download-direct-link"
-      href={IOS_DOWNLOAD_URL}
-      target="_blank"
-      rel="noopener noreferrer"
-    >
-      {t("download.iosDirectDownload")}
-    </a>
-  </div>
-);
+) => {
+  const { downloadUrl } = useMobileDownloadUrl(
+    IOS_UPDATER_PATH,
+    IOS_DOWNLOAD_URL
+  );
 
-// iOS 安装入口：按钮只控制二维码浮窗，TestFlight 地址仅用于生成二维码
+  return (
+    <div
+      className="wk-login-mobile-download-popover"
+      role="dialog"
+      aria-label={t("download.iosQrTitle")}
+      {...hoverProps}
+    >
+      <div
+        className="wk-login-mobile-popover-qr"
+        role="img"
+        aria-label={t("download.iosQrLabel")}
+      >
+        <QRCodeSVG value={downloadUrl} size={104} />
+      </div>
+      <strong className="wk-login-mobile-download-popover-title">
+        {t("download.iosQrTitle")}
+      </strong>
+    </div>
+  );
+};
+
+// iOS 安装入口：按钮只控制二维码浮窗，安装地址由 updater 接口提供
 export const IOSDownloadButton: React.FC = () => {
   const [hoverVisible, setHoverVisible] = React.useState(false);
   const [clickPinned, setClickPinned] = React.useState(false);
--- a/packages/dmworklogin/src/IOSDownloadButton.tsx
+++ b/packages/dmworklogin/src/IOSDownloadButton.tsx
@@ -1,11 +1,11 @@
 import React from "react";
-import { Popover } from "@douyinfe/semi-ui";
+import { Popover, Spin } from "@douyinfe/semi-ui";
+import { WKButton } from "@octo/base";
 import { QRCodeSVG } from "qrcode.react";
 import { loginT as t } from "./i18n";
 import { useMobileDownloadUrl } from "./mobileDownloadUpdater";
 import "./MobileDownloadPopover.css";
 
-export const IOS_DOWNLOAD_URL = "https://testflight.apple.com/join/uPrdCcy3";
 export const IOS_UPDATER_PATH = "common/updater/ios/1.0.0";
 
 type PopoverHoverProps = Pick<
@@ -16,10 +16,7 @@ type PopoverHoverProps = Pick<
 export const IOSDownloadPopoverContent: React.FC<PopoverHoverProps> = (
   hoverProps
 ) => {
-  const { downloadUrl } = useMobileDownloadUrl(
-    IOS_UPDATER_PATH,
-    IOS_DOWNLOAD_URL
-  );
+  const { status, downloadUrl, retry } = useMobileDownloadUrl(IOS_UPDATER_PATH);
 
   return (
     <div
@@ -30,10 +27,22 @@ export const IOSDownloadPopoverContent: React.FC<PopoverHoverProps> = (
     >
       <div
         className="wk-login-mobile-popover-qr"
+        data-status={status}
         role="img"
         aria-label={t("download.iosQrLabel")}
       >
-        <QRCodeSVG value={downloadUrl} size={104} />
+        {status === "loading" && (
+          <Spin aria-label={t("download.loadingAddress")} size="large" />
+        )}
+        {status === "error" && (
+          <div className="wk-login-mobile-download-state" role="alert">
+            <span>{t("download.addressLoadFailed")}</span>
+            <WKButton type="button" variant="ghost" size="sm" onClick={retry}>
+              {t("download.retry")}
+            </WKButton>
+          </div>
+        )}
+        {status === "ready" && <QRCodeSVG value={downloadUrl} size={104} />}
       </div>
       <strong className="wk-login-mobile-download-popover-title">
         {t("download.iosQrTitle")}
--- a/packages/dmworklogin/src/IOSDownloadButton.tsx
+++ b/packages/dmworklogin/src/IOSDownloadButton.tsx
@@ -28,8 +28,9 @@ export const IOSDownloadPopoverContent: React.FC<PopoverHoverProps> = (
       <div
         className="wk-login-mobile-popover-qr"
         data-status={status}
-        role="img"
-        aria-label={t("download.iosQrLabel")}
+        role={status === "ready" ? "img" : undefined}
+        aria-label={status === "ready" ? t("download.iosQrLabel") : undefined}
+        aria-busy={status === "loading" ? true : undefined}
       >
         {status === "loading" && (
           <Spin aria-label={t("download.loadingAddress")} size="large" />
--- a/packages/dmworklogin/src/MobileDownloadPopover.css
+++ b/packages/dmworklogin/src/MobileDownloadPopover.css
@@ -34,10 +34,25 @@
   background: var(--wk-color-white);
 }
 
-.wk-login-mobile-popover-qr svg {
+.wk-login-mobile-popover-qr[data-status="loading"],
+.wk-login-mobile-popover-qr[data-status="error"] {
+  background: var(--wk-bg-surface);
+}
+
+.wk-login-mobile-popover-qr > svg {
   display: block;
   width: 100%;
   height: 100%;
+}
+
+.wk-login-mobile-download-state {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  gap: var(--wk-sp-2);
+  color: var(--wk-text-secondary);
+  font-size: var(--wk-text-size-sm);
+  line-height: var(--wk-leading-normal);
 }
 
 .wk-login-mobile-download-popover-title {
@@ -74,6 +89,13 @@
 .wk-login-mobile-download-direct-link:focus-visible {
   outline: 2px solid var(--wk-border-glow);
   outline-offset: 2px;
+}
+
+.wk-login-mobile-download-direct-link[aria-disabled="true"] {
+  border-color: var(--wk-border-default);
+  background: var(--wk-bg-elevated);
+  color: var(--wk-text-disabled);
+  cursor: not-allowed;
 }
 
 .wk-login-android-popover-manual-download {
--- a/packages/dmworklogin/src/__tests__/android_download_button.test.tsx
+++ b/packages/dmworklogin/src/__tests__/android_download_button.test.tsx
@@ -265,15 +265,9 @@ describe("AndroidDownloadButton", () => {
     expect(directDownload?.textContent).toBe("直接下载 APK");
   });
 
-  it("downloads the updater-provided APK URL when the direct action is clicked", async () => {
+  it("uses one updater-provided APK URL for the QR code and direct action", async () => {
     const updaterUrl = "https://cdn.example.com/releases/octo-latest.apk";
     apiFetchJsonMock.mockResolvedValue({ url: updaterUrl });
-    let clickedHref = "";
-    vi.spyOn(HTMLAnchorElement.prototype, "click").mockImplementation(
-      function (this: HTMLAnchorElement) {
-        clickedHref = this.href;
-      }
-    );
     const container = document.createElement("div");
     document.body.appendChild(container);
     mountedContainers.push(container);
@@ -284,13 +278,6 @@ describe("AndroidDownloadButton", () => {
       );
     });
 
-    act(() => {
-      Simulate.click(
-        container.querySelector(
-          ".wk-login-mobile-download-direct-link"
-        ) as Element
-      );
-    });
     await act(async () => {
       await Promise.resolve();
     });
@@ -298,28 +285,21 @@ describe("AndroidDownloadButton", () => {
     expect(apiFetchJsonMock).toHaveBeenCalledWith(
       "/api/v1/common/updater/android/1.0"
     );
-    expect(clickedHref).toBe(updaterUrl);
-  });
+    expect(
+      container.querySelector("[data-qr-value]")?.getAttribute("data-qr-value")
+    ).toBe(updaterUrl);
+    expect(
+      container.querySelector<HTMLAnchorElement>(
+        ".wk-login-mobile-download-direct-link"
+      )?.href
+    ).toBe(updaterUrl);
 
-  it("falls back to the same-origin APK when the updater request fails", async () => {
-    apiFetchJsonMock.mockRejectedValue(new Error("network error"));
     let clickedHref = "";
-    vi.spyOn(HTMLAnchorElement.prototype, "click").mockImplementation(
-      function (this: HTMLAnchorElement) {
-        clickedHref = this.href;
-      }
-    );
-
-    const container = document.createElement("div");
-    document.body.appendChild(container);
-    mountedContainers.push(container);
-    act(() => {
-      ReactDOM.render(
-        React.createElement(AndroidDownloadPopoverContent),
-        container
-      );
+    vi.spyOn(HTMLAnchorElement.prototype, "click").mockImplementation(function (
+      this: HTMLAnchorElement
+    ) {
+      clickedHref = this.href;
     });
-
     act(() => {
       Simulate.click(
         container.querySelector(
@@ -330,8 +310,35 @@ describe("AndroidDownloadButton", () => {
     await act(async () => {
       await Promise.resolve();
     });
+    expect(clickedHref).toBe(updaterUrl);
+    expect(apiFetchJsonMock).toHaveBeenCalledTimes(1);
+  });
 
-    expect(clickedHref).toBe(`${window.location.origin}${ANDROID_APK_PATH}`);
+  it("falls back to the same-origin APK when the updater request fails", async () => {
+    apiFetchJsonMock.mockRejectedValue(new Error("network error"));
+    const container = document.createElement("div");
+    document.body.appendChild(container);
+    mountedContainers.push(container);
+    act(() => {
+      ReactDOM.render(
+        React.createElement(AndroidDownloadPopoverContent),
+        container
+      );
+    });
+
+    await act(async () => {
+      await Promise.resolve();
+    });
+
+    const fallbackUrl = `${window.location.origin}${ANDROID_APK_PATH}`;
+    expect(
+      container.querySelector("[data-qr-value]")?.getAttribute("data-qr-value")
+    ).toBe(fallbackUrl);
+    expect(
+      container.querySelector<HTMLAnchorElement>(
+        ".wk-login-mobile-download-direct-link"
+      )?.href
+    ).toBe(fallbackUrl);
   });
 
   it("opens GitHub Releases safely in a new tab", () => {
--- a/packages/dmworklogin/src/__tests__/android_download_button.test.tsx
+++ b/packages/dmworklogin/src/__tests__/android_download_button.test.tsx
@@ -225,8 +225,9 @@ describe("AndroidDownloadButton", () => {
       React.createElement(AndroidDownloadPopoverContent)
     );
 
-    expect(html).toContain('role="img"');
     expect(html).toContain("wk-login-mobile-popover-qr");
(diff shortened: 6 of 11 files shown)


lml2468 (dismissed):

审查锚定 head `fcde457e53c1`。2 文件 +140/−1,`dmworklogin` 恢复 updater 驱动的 Android APK 下载。已跑 login 测试(12 例)、核实 URL 安全校验与 fetch 接线。

### ✅ 两道闸门均通过

**规格符合**:与描述一致——直接下载改为先请求 `common/updater/android/1.0` 拿 updater URL 下载,保留同源 `/download/dmwork.apk` 兜底;二维码 + GitHub Releases 兜底不动。无 over-build。

**安全(本 PR 关键——URL 来自后端,下载到浏览器)—— 处理到位**:
- `resolveSafeDownloadUrl`:updater 返回值先 `new URL(value, origin)` 解析,**仅放行 `http:`/`https:`**——`javascript:`/`data:`/`file:` 等被拒,回落同源 APK。防了「后端返回恶意 scheme → 点击执行」这条路。
- 下载用 `<a download>` + `link.click()`(非 `window.location`/`window.open`),即便 URL 异常也只触发下载、不导航、无脚本执行面。
- `apiFetchJson` 失败(网络/非 JSON)`catch` 后回落同源 APK,不中断。
- `resolveAndroidUpdaterUrl` 正确用 `WKApp.apiClient.config.apiURL` 拼全 URL 再交给 `apiFetchJson`(`apiFetch` 直接 `fetch(input)` 不 prepend,全 URL 由调用方给,无双拼/漏拼)。

**代码质量**:
- `apiFetchJson` 确经 `@octo/base` 导出(`index.tsx` `export * from './Service/apiFetch'`),import 可解析。
- 纯函数(`resolveAndroidUpdaterUrl`/`resolveSafeDownloadUrl`)拆出可测;`onClick` `preventDefault` + `void downloadLatestAndroidApk()`,`href` 仍保留同源 APK 作为无 JS 降级(渐进增强)。

### ✅ 验证

- `android_download_button.test.tsx` 本地 **12 passed**,含关键新测试:updater URL 命中下载(`clickedHref === updaterUrl`)、updater 失败回落同源 APK(`clickedHref === origin + ANDROID_APK_PATH`)、endpoint 拼接(相对/绝对 apiURL 两种)。
- URL scheme 白名单、`<a download>` 非导航、fetch 失败回落、apiFetchJson 导出与全 URL 拼接均 byte 核实。
- gate REVIEW_REQUIRED(CI pending),mergeable。

### 🟡 非阻塞

- `resolveSafeDownloadUrl` 只校验 scheme,不校验 host——updater 是可信后端(同源鉴权返回),返回任意 http(s) CDN URL 是预期行为(APK 常托管在 CDN),故不设 host 白名单合理;若未来 updater 契约收紧到固定域,可加 host 断言。当前非风险。

### 结论

URL 安全校验(scheme 白名单)+ 非导航下载 + 失败回落三层到位,fetch 接线正确,纯函数可测、测试覆盖命中/回落/拼接三路径。无阻塞。**APPROVE**。

Jerry-Xin (dismissed):
The PR is relevant to octo-web and correctly restores updater-backed Android APK downloads with a safe fallback.

💬 Non-blocking

- 🟡 Warning — [AndroidDownloadButton.tsx](/Users/mlamp/.openclaw/tmp/pr-review-3jqv1s7l/repo/packages/dmworklogin/src/AndroidDownloadButton.tsx:50): `apiFetchJson` has no timeout, so a stalled request can delay the fallback indefinitely. Consider using an `AbortSignal` with a short timeout.
- 🔵 Suggestion — [AndroidDownloadButton.tsx](/Users/mlamp/.openclaw/tmp/pr-review-3jqv1s7l/repo/packages/dmworklogin/src/AndroidDownloadButton.tsx:22): Project guidance recommends placing new HTTP boundaries in a base `Service` rather than accessing `WKApp` directly from UI code.
- 🔵 Suggestion — [android_download_button.test.tsx](/Users/mlamp/.openclaw/tmp/pr-review-3jqv1s7l/repo/packages/dmworklogin/src/__tests__/android_download_button.test.tsx:268): Add coverage for malformed, empty, and non-HTTP updater URLs to protect the security validation and fallback behavior.

✅ Highlights

- Restricts updater destinations to HTTP(S), preventing script-scheme injection.
- Preserves the same-origin fallback, QR code, and GitHub Releases behavior.
- Covers successful updater resolution and rejected-request fallback.
- URL composition handles API bases with and without trailing slashes.

Tests could not be rerun locally because dependencies are not installed (`vitest: command not found`).

Jerry-Xin (dismissed):
The PR is relevant to octo-web and correctly restores updater-backed Android APK downloads with a safe fallback.

💬 Non-blocking

- 🟡 Warning — `packages/dmworklogin/src/AndroidDownloadButton.tsx:50` (`downloadLatestAndroidApk`): `apiFetchJson` has no timeout, so a stalled updater request can delay the same-origin fallback indefinitely. Consider passing an `AbortSignal` with a short timeout.
- 🔵 Suggestion — `packages/dmworklogin/src/AndroidDownloadButton.tsx:22` (`resolveAndroidUpdaterUrl`): Project guidance recommends placing new HTTP boundaries in a base `Service` rather than reading `WKApp.apiClient.config.apiURL` directly from UI code.
- 🔵 Suggestion — `packages/dmworklogin/src/__tests__/android_download_button.test.tsx:268`: Add coverage for malformed, empty, and non-HTTP(S) updater URLs to lock in the `resolveSafeDownloadUrl` validation and fallback behavior.

✅ Highlights

- `resolveSafeDownloadUrl` restricts updater destinations to `http:`/`https:`, preventing `javascript:`/`data:` scheme injection from an untrusted updater response.
- Preserves the same-origin `/download/dmwork.apk` fallback, the QR code, and the GitHub Releases behavior unchanged.
- Tests cover both successful updater resolution and rejected-request fallback; `resolveAndroidUpdaterUrl` handles API bases with and without trailing slashes.

Note: tests were not rerun locally because dev dependencies are not installed in the review mirror; verified by reading source at the head commit.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #931 (octo-web)

## Summary

This PR restores the updater-backed APK download flow for the Android login download button. Previously, the direct download link served a static same-origin APK path (`/download/dmwork.apk`). Now, clicking the direct download link first queries a server updater endpoint (`common/updater/android/1.0`) via the shared `apiFetchJson` helper to obtain the latest APK URL, then triggers the browser download with that URL. If the updater call fails or returns an invalid URL, the component falls back to the legacy same-origin APK path — preserving the pre-PR behavior as a safe default.

The change is well-scoped: one production file (`AndroidDownloadButton.tsx`) and one test file. The implementation is clean — URL validation in `resolveSafeDownloadUrl` correctly restricts to `http:` / `https:` protocols, the fallback chain is linear and easy to reason about, and the `<a>` tag retains the legacy `href` for no-JS / accessibility fallback. One non-blocking P2 finding on test coverage below.

## Verification

- ✅ **Diff scope** — 2 files changed: +140 / -1. Production changes confined to `packages/dmworklogin/src/AndroidDownloadButton.tsx`; test additions in the companion test file.
- ✅ **Data-flow: updater URL resolution** — `resolveAndroidUpdaterUrl()` reads `WKApp.apiClient.config.apiURL` at call time (default param), strips/appends trailing `/`, concatenates `ANDROID_UPDATER_PATH`. The regex `/\/?$/` is anchored and correct. Verified with test assertions for both trailing-slash and no-trailing-slash API URLs.
- ✅ **Data-flow: download URL fallback chain** — `downloadLatestAndroidApk()` initializes `downloadUrl` to `resolveAndroidApkUrl()` (same-origin APK). On updater success, `resolveSafeDownloadUrl(result?.url) ?? downloadUrl` replaces it only if the server returns a valid http(s) URL. On any error (network, parse, protocol mismatch, empty/missing/non-string), falls back to the same-origin path. The chain is linear — no intermediate state, no race condition.
- ✅ **Security: protocol allowlist** — `resolveSafeDownloadUrl` rejects `javascript:`, `data:`, `file:`, and other non-http(s) protocols via explicit `url.protocol` check. `new URL(value, window.location.origin)` handles relative URLs safely. Non-string and empty values are rejected before URL construction.
- ✅ **Error handling** — `apiFetchJson` throws `ApiFetchError` on non-2xx; the catch block in `downloadLatestAndroidApk` swallows it and proceeds with the fallback URL. No unhandled rejection path (the `void` on the caller side is safe because the function never re-throws).
- ✅ **ASYNC/RETRY** — No retry mechanism; single attempt with fallback. No intermediate state is mutated before the async operation completes, so there is nothing to roll back on failure.
- ✅ **RETURN-WIRING** — `downloadLatestAndroidApk` returns `Promise<void>`; called with `void` at `:103`. No return value consumed by callers — the function's only side effect is `startBrowserDownload`. N/A.
- ✅ **CURSOR/PROGRESS** — No pagination or cursor involved. Single API call returning a URL. N/A.

Static analysis only at head `fcde457e53c1130a1fbebade3af7f5fc695dec36`; build and tests not executed in this environment.

## Findings

### P2 — Security-relevant URL validator lacks negative-path tests (`packages/dmworklogin/src/__tests__/android_download_button.test.tsx:268`)

`resolveSafeDownloadUrl` is the security gate that prevents a compromised or misconfigured updater endpoint from triggering `javascript:`, `data:`, or `file:` downloads. The test suite covers the happy path (valid HTTPS URL at the "downloads the updater-provided APK URL" test) and the network-error fallback ("falls back to the same-origin APK when the updater request fails"), but does not exercise the rejection paths that are the function's primary purpose.

If a future refactor weakens the protocol check or the `typeof` guard, the current tests would not detect it because they only assert correct behavior for valid and missing responses — not for hostile ones. Consider adding test cases where `apiFetchJson` resolves to `{ url: "javascript:alert(1)" }`, `{ url: "" }`, `{ url: 42 }`, and `{ url: "file:///etc/passwd" }`. Each should assert the download falls back to the same-origin APK URL.

## Blind-spot Checklist

- **C1 (Dual-path parity)** — iOS uses a static TestFlight URL (`IOSDownloadButton.tsx`), unchanged by this PR. Asymmetry is by design (TestFlight vs APK sideloading). Clear.
- **C2 (Control-flow ordering)** — `event.preventDefault()` runs before `downloadLatestAndroidApk()`. No nested reuse of the download helper. Clear.
- **C3 (Authorization boundary)** — `apiFetchJson` is called without auth headers. The updater endpoint is likely public. On auth failure (401/403), the catch block falls back to the legacy APK — fails closed. Clear.
- **C4 (Authorization lifecycle)** — N/A. No auth-dependent state.
- **C5 (Build/runtime)** — Pure TypeScript/React. No build-time or packaging changes. Clear.
- **C6 (Governance/docs)** — N/A.

- `startBrowserDownload` creates a detached `<a>`, clicks it, removes it — standard pattern, works in all modern browsers.
- The `<a>` tag retains `href={resolveAndroidApkUrl()}` and `download` attribute, so the no-JS fallback still serves the legacy APK.
- `onClick` handler uses `event.preventDefault()` correctly, preventing the browser from navigating to the `href` while the async updater call runs.
- `login_presentation.test.ts` still asserts that `login.tsx` does not leak updater/APK implementation details — module boundary is preserved.
- `vi.hoisted` is used correctly for the `apiFetchJsonMock`, ensuring it's available before the `vi.mock("@octo/base")` call.

## Verdict: APPROVED

Clean, well-structured change that restores updater-backed APK downloads with a safe fallback chain. The production code is correct and the security validation in `resolveSafeDownloadUrl` is sound. The only finding is a non-blocking P2 on test coverage for the URL validator's rejection paths.

yujiawei (dismissed):
# Code Review — PR #931 (octo-web)

## Summary

This PR restores updater-backed APK download for the Android login button. The prior refactor (#815) had reduced the direct-download link to a static same-origin path (`/download/dmwork.apk`). This change intercepts the link click, queries the server updater endpoint (`common/updater/android/1.0`) via the shared `apiFetchJson` helper, validates the returned URL, and triggers a browser download to the latest APK — falling back to the legacy same-origin path on any failure or invalid response. QR code and GitHub Releases fallback are untouched. Scope is tight: one production file plus its test.

I verified this against the head SHA `fcde457e`, ran the full suite (`19 files, 209 tests` — all pass), and traced the data flow. The implementation is correct and the change matches its stated scope exactly. Verdict below is **Approve**; all findings are non-blocking (P2 / suggestions).

## What I verified

- **Scope / no over-build** — Only the three stated behaviors are implemented (updater request, http/https URL with same-origin fallback, unchanged QR + Releases). No extra flags, endpoints, or unrelated edits.
- **URL join** — `resolveAndroidUpdaterUrl` uses `apiUrl.replace(/\/?$/, "/")` to normalize both trailing-slash and no-trailing-slash API URLs; asserted by tests for `/api/v1/` and `https://api.example.com/v1`.
- **Fallback chain** — `downloadLatestAndroidApk` seeds `downloadUrl` with the same-origin path and only replaces it via `resolveSafeDownloadUrl(result?.url) ?? downloadUrl`. Any throw (network / non-2xx `ApiFetchError`), non-string, empty, or non-http(s) URL keeps the legacy path. Linear, no intermediate mutable state.
- **Security** — `resolveSafeDownloadUrl` restricts to `http:` / `https:`, rejecting `javascript:`, `data:`, `file:` and malformed input before use. This is a net security improvement over the pre-#815 baseline, which assigned `data.url` to the href with no protocol check. The URL is sourced from the trusted authenticated backend updater (same pattern already used for `common/updater/web/1.0` in `NavSettingsPanel`).
- **No-JS / a11y fallback** — the `<a>` retains `href={resolveAndroidApkUrl()}` + `download`, so the legacy path still works without JS.
- **Module boundary** — `login_presentation.test.ts` still asserts `login.tsx` does not leak updater/APK details; preserved.

### P2 — Download now fires after an `await`, outside the user-gesture window
`AndroidDownloadButton.tsx:101-104` + `:50-61`

`onClick` calls `event.preventDefault()` (cancelling the native `<a download>` navigation) and then `downloadLatestAndroidApk()` performs the programmatic `link.click()` **after** `await apiFetchJson(...)`. Breaking the synchronous gesture chain means a browser that enforces strict user-activation for downloads could silently block it — and because `preventDefault()` already cancelled the native link, there is no synchronous fallback if that happens.

Worth noting this is a behavioral trade-off versus the original (pre-#815) design, which fetched the updater URL on **mount** (`useEffect`) and kept the click fully synchronous. Moving the fetch to click-time is what introduces the risk. For the Android-Chrome target this is low-impact (Android permits programmatic downloads post-await, and the updater call to a same-origin/proxied endpoint is fast), so I'm not treating it as a blocker. **Suggestion:** prefetch the updater URL on hover/mount and keep the click handler synchronous, restoring the original gesture guarantee.

### P2 — Missing negative-path tests for the security validator
`__tests__/android_download_button.test.tsx:268`

`resolveSafeDownloadUrl` is the gate that blocks `javascript:` / `data:` / `file:` and malformed URLs, but tests only cover the valid-HTTPS happy path and the network-error fallback — not the rejection paths that are the function's purpose. A future refactor that weakened the protocol/`typeof` check would not be caught. Suggest adding cases where `apiFetchJson` resolves to `{ url: "javascript:alert(1)" }`, `{ url: "file:///etc/passwd" }`, `{ url: "" }`, and `{ url: 42 }`, each asserting fallback to the same-origin APK.

### P2 — Cross-origin `download` attribute may be ignored
`AndroidDownloadButton.tsx:40-48`

The `download` attribute is ignored by browsers for cross-origin URLs. If the updater returns a CDN URL without `Content-Disposition: attachment`, the browser may navigate rather than download. For an APK (non-renderable) this typically still results in a download, so impact is minor — but worth confirming the updater/CDN sets the header.

### P2 (nit) — No pending/disabled state on the direct-download action
`AndroidDownloadButton.tsx:97-107`

Clicking fires `void downloadLatestAndroidApk()` with no loading indicator or debounce; on a slow updater, rapid clicks can spawn concurrent requests and overlapping downloads. Minor UX; consider disabling the link while in-flight.

## Coverage notes / blind spots

- Real browser popup-blocker behavior and the live updater's `Content-Disposition` headers cannot be verified statically (tests mock `click` synchronously in JSDOM). The two P2s above depend on runtime behavior.
- The updater response contract is assumed to be `{ url: string }`, consistent with prior art and the pre-#815 implementation.
- iOS button (static TestFlight URL) is intentionally unchanged; asymmetry is by design.

## Verdict

Clean, well-scoped restore with a safe fallback chain and a genuine security improvement over the prior baseline. No P0/P1 issues. The findings above are all non-blocking suggestions; the gesture-timing item (first P2) is the most valuable to address in a follow-up.

**Approve.**

lml2468 (dismissed):
## Review Verdict（复审 @ 3c5c14f7）：可合并（APPROVE）—— 维持 APPROVE(范围扩展到 iOS)

复审锚定新 head `3c5c14f7`,对比我上次 APPROVE 的 `fcde457e`。范围从「仅 Android」扩展到「Android + iOS + 共享模块」(2 文件 → 10 文件),但核心安全模型保持,byte 复核:

### ✅ 本轮扩展(超出上次 APPROVE 的部分)

- **抽共享模块 `mobileDownloadUpdater.ts`**:`resolveSafeDownloadUrl`(**仅 http/https 白名单**)、`fetchMobileDownloadUrl`(失败/非法回落 caller fallback)、`startMobileDownload`(`<a download>` 非导航)、`useMobileDownloadUrl` hook——**安全三层与我上次审的完全一致**,只是泛化 + 复用。Android 原内联逻辑改为 delegate。
- **Android QR**:`value` 从静态 `resolveAndroidApkUrl()` 改为 updater 解析的 `downloadUrl`——QR 与直接下载现同源于 updater。
- **iOS**:新增 `IOS_UPDATER_PATH="common/updater/ios/1.0.0"`,QR 走 updater(TestFlight 作 fallback);**移除「打开 TestFlight」直接入口**,保留 QR。
- **i18n**:删 `iosDirectDownload`(zh+en)——**无 dangling 引用**(全 src grep 零命中),干净移除。

### ✅ 安全 / 正确性(复核)

- updater URL 仍 **scheme 白名单**(http/https),挡 javascript:/data:;失败回落各自 fallback(Android 同源 APK / iOS TestFlight);
- 下载走 `<a download>` 非导航,无脚本执行面;
- `useMobileDownloadUrl` 用 `requestRef` memo promise + mount 时 `useEffect` 预解析 → 点击时 `href={downloadUrl}` 已是解析值,`resolveDownloadUrl()` 复用已缓存 promise——**缓解了 yujiawei 提的「异步点击被 popup-block」**(大多数情况点击时已 resolve)。

### ✅ 验证

- `android_download_button.test.tsx` + `ios_download_button.test.tsx` 本地 **19 passed**。
- rebase 到 main tip;removed key 无 dangling、scheme 白名单、非导航下载、pre-resolve 均 byte 核实。

### 🟡 非阻塞

- 点击路径仍是 `resolveDownloadUrl().then(...)`——虽 mount 已预取,极快点击(解析未完成)仍会异步,理论上 Safari 可能 popup-block。已比上次改进(pre-resolve),彻底同步需在点击时用已 resolved 值兜底,可 fast-follow。yujiawei 也是同样 🟡。
- presigned/updater URL 只校 scheme 不校 host——updater 是自家后端,同 #925/上次结论,非阻塞。

### 结论

上次 APPROVE 的 Android 逻辑不变,本轮把它抽成共享模块并对称扩展到 iOS,安全模型(scheme 白名单 + 非导航 + 失败回落)完整保持,移除 key 无残留,测试全绿。**维持 APPROVE**。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #931 (octo-web)

## Summary

This PR restores updater-backed download URLs for the login page's mobile download buttons. A new shared module (`mobileDownloadUpdater.ts`) provides a `useMobileDownloadUrl` hook and `startMobileDownload` helper that fetch the real download URL from the `/common/updater/{platform}/{version}` API endpoint, falling back to the same-origin APK URL (Android) or the public TestFlight URL (iOS) on failure. The Android popover now uses the updater URL for both its QR code and direct download action; the iOS popover uses it for the QR code only, and the direct "Open TestFlight" link is removed. The i18n `iosDirectDownload` key is cleaned up from both locales.

The implementation is clean, well-tested, and handles SSR, fallback, URL validation, and request deduplication correctly. One non-blocking finding below.

## Verification

- ✅ **Data flow: updater URL → QR + download** — `useMobileDownloadUrl` returns `{ downloadUrl, resolveDownloadUrl }`. Both Android QR `value` and `<a href>` consume `downloadUrl`; the click handler calls `resolveDownloadUrl().then(startMobileDownload)`. Data flows correctly from hook to all consumption points.
- ✅ **Fallback chain** — `fetchMobileDownloadUrl` catches all errors (network, JSON parse, invalid URL shape) and returns `fallbackUrl`. `resolveSafeDownloadUrl` rejects non-string, empty, non-http(s), and malformed URLs. The fallback path is robust.
- ✅ **SSR safety** — `useState(fallbackUrl)` ensures the initial render (including `renderToStaticMarkup`) produces the fallback URL without calling `window` or network APIs. `useEffect` only runs client-side.
- ✅ **URL validation** — `resolveSafeDownloadUrl` rejects `javascript:`, `data:`, and other non-http(s) protocols via the explicit `url.protocol` check. Relative URLs from the updater are resolved against `window.location.origin`.
- ✅ **Request dedup** — `requestRef` ensures the updater endpoint is called at most once per component mount, even if the popover is toggled multiple times.
- ✅ **i18n cleanup** — `iosDirectDownload` removed from both `en-US.json` and `zh-CN.json`; no remaining references in production code (the iOS `<a>` that used it is removed).
- ✅ **Test coverage** — 19 files / 210 tests per PR description. Tests cover: updater URL resolution, updater-provided URL rendering, fallback on network error, SSR fallback, removed direct TestFlight action, and programmatic download click flow.

## Findings

### P2 — `requestRef` cache not invalidated when hook args change (`packages/dmworklogin/src/mobileDownloadUpdater.ts:46`)

`requestRef.current` holds the first resolved promise indefinitely. When `updaterPath` or `fallbackUrl` change, `useCallback` correctly produces a new `resolveDownloadUrl` with updated closure values, and the `useEffect` re-runs — but inside the callback, `if (!requestRef.current)` is false (stale ref), so it returns the old promise. The effect then calls `setDownloadUrl(oldUrl)`, ignoring the new args.

Currently unreachable because both callers (`AndroidDownloadPopoverContent` and `IOSDownloadPopoverContent`) pass module-level constants that never change during the component lifetime. But if this hook is ever reused with dynamic params (e.g. user-selectable download target, or `apiURL` that changes after config loads), it will silently serve stale data.

Fix — clear the ref at the top of the effect:

```ts
React.useEffect(() => {
  requestRef.current = undefined;
  let active = true;
  void resolveDownloadUrl().then((url) => {
    if (active) setDownloadUrl(url);
  });
  return () => { active = false; };
}, [resolveDownloadUrl]);
```

- **Updater URL construction** — `resolveMobileUpdaterUrl` correctly joins `apiURL` (with or without trailing slash) and the relative updater path. Tested with both `/api/v1/` and `https://api.example.com/v1`.
- **Android `startMobileDownload`** — programmatic `<a>` click with `download` attribute. Cross-origin `download` attribute is ignored by browsers, but CDN-served APKs typically send `Content-Disposition: attachment`, so the download still triggers correctly.
- **iOS direct TestFlight link removal** — intentional product decision per PR description. The QR code still works with the updater-provided or fallback TestFlight URL.
- **login.tsx** — comment-only change, no behavioral impact.
- **Storybook descriptions** — updated to reflect the new updater-backed behavior. No production impact.
- **`login_presentation.test.ts`** — the existing assertion that `login.tsx` does not contain `"common/updater/android/1.0"` still passes because the updater path string lives in `AndroidDownloadButton.tsx`, not `login.tsx`.

## Blind-spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or subscribe/unsubscribe paths in this change.
- **C2 (control-flow ordering):** N/A — no reused/nested call-sites for the changed logic.
- **C3 (authorization boundary):** Cleared — `apiFetchJson` uses the shared API client with auth headers. No new unauthenticated endpoints introduced.
- **C5 (build/runtime):** N/A — no build or deployment config changes.
- **C6 (governance/docs):** N/A — no governance or policy changes.

## Verdict: APPROVED

Clean, well-structured PR with thorough test coverage. The shared `useMobileDownloadUrl` hook correctly handles SSR, fallback, URL validation, and request dedup. The single P2 (stale ref on arg change) is currently unreachable and does not block merge.

[Octo-Q] verdict: APPROVE — no P1 issues; one P2 (latent ref-cache invalidation in hook) is non-blocking.

yujiawei (approved):
# Code Review — PR #931 (octo-web)

## Summary

This PR routes the login-page mobile download entries through the backend updater endpoint. A new shared module `mobileDownloadUpdater.ts` exposes `useMobileDownloadUrl(updaterPath, fallbackUrl)`, which fetches `{ url }` from the updater on mount, validates it, and returns it (falling back to a caller-provided default on any failure). Android uses `common/updater/android/1.0` for both the QR code and the direct-download link (fallback: same-origin `/download/dmwork.apk`); iOS uses `common/updater/ios/1.0.0` for the QR (fallback: the public TestFlight URL), and the standalone "Open TestFlight" action is removed. QR/GitHub-Releases behavior is otherwise preserved.

Reviewed at head `3c5c14f7083bb8361d28d6452ba2a8a70ae1e50e`, diffed against merge-base `b39924e6`. Scope is exactly 10 files under `packages/dmworklogin/` (+299/−76). I traced the data flow and inspected the security validator; findings below are all non-blocking (P2 / suggestions).

## Spec compliance

Matches the stated intent with no over-build:
- Android QR and direct download share one updater-provided URL — `AndroidDownloadButton.tsx:66,73`.
- iOS QR uses `common/updater/ios/1.0.0` — `IOSDownloadButton.tsx:9,36`.
- "Open TestFlight" direct action removed, QR entry retained — `IOSDownloadButton.tsx` (link element gone); dangling `iosDirectDownload` i18n key removed from both locales.
- GitHub Releases fallback unchanged — `AndroidDownloadButton.tsx:82-92`.
- Same-origin APK and public TestFlight retained strictly as failure fallbacks.

No new flags, endpoints, or unrelated edits.

## What I verified

- **URL join** — `resolveMobileUpdaterUrl` normalizes the base with `apiUrl.replace(/\/?$/, "/")`; asserted for both `/api/v1/` and `https://api.example.com/v1`. `apiFetch` does not prepend a base, so the fully-qualified URL from the caller is correct (`apiFetch.ts:51`).
- **`apiFetchJson` / `WKApp` exports** — both re-exported from `@octo/base` (`index.tsx: export * from './Service/apiFetch'`, `export { default as WKApp } from './App'`); `WKApp.apiClient.config.apiURL` is a real accessor (`APIClient.ts:42`). Same pattern already used for `common/updater/web/1.0` in `NavSettingsPanel.tsx:64`.
- **Fallback chain** — `fetchMobileDownloadUrl` returns `resolveSafeDownloadUrl(result?.url) ?? fallbackUrl`, and the outer `try/catch` returns `fallbackUrl` on any throw. Non-string/empty/non-http(s)/malformed inputs all keep the fallback. Linear, no intermediate mutable state.
- **Security (scheme allowlist)** — `resolveSafeDownloadUrl` only admits `http:`/`https:`, rejecting `javascript:`/`data:`/`file:` before use (`mobileDownloadUpdater.ts:11-22`). Net improvement over the pre-#815 baseline.
- **SSR** — initial state is the fallback URL; the fetch (and the only `window.location.origin` access) runs inside `useEffect`, i.e. client-only, so server render is safe in the current call graph.

### P2 — QR/link accept any http(s) host; no destination allowlist
`mobileDownloadUpdater.ts:15`, rendered at `IOSDownloadButton.tsx:36` / `AndroidDownloadButton.tsx:66,73`

The validator gates on scheme only, so any `http(s)` URL the updater returns is rendered directly into the QR code (and, for Android, the download link). The updater is a trusted, authenticated same-origin backend and APKs are commonly CDN-hosted, so allowing arbitrary hosts is the intended contract — this is not a live vulnerability. But it does mean the trust boundary rests entirely on the backend: a misconfigured/compromised updater response could repoint the QR to an arbitrary site. If the updater contract ever narrows to fixed domains, add a host assertion. Suggestion, not a blocker.

### P2 — Android direct download fires after an `await`, outside the synchronous gesture window
`AndroidDownloadButton.tsx:75-78` + `mobileDownloadUpdater.ts:62-69`

`onClick` calls `event.preventDefault()` (cancelling the native `<a download>`) and then triggers the programmatic `link.click()` after `await resolveDownloadUrl()`. If that promise is still pending at click time, some browsers enforcing strict user-activation could silently block the download, with no synchronous native fallback left. In practice the fetch is kicked off on mount/popover-open (`useEffect`), so by the time the user clicks the direct link the promise is almost always already resolved and the click is effectively synchronous — which is why I keep this at P2. Optional hardening: keep the resolved URL on the `href` (already done) and, if the promise is unresolved, let the native navigation proceed instead of `preventDefault()`.

### P2 — `requestRef` is never invalidated; failures are cached for the component's lifetime
`mobileDownloadUpdater.ts:40-47`

The in-flight/resolved promise is memoized in `requestRef` and never cleared. Two consequences: (a) if `updaterPath`/`fallbackUrl` ever changed, the new `useCallback`/`useEffect` would still reuse the old promise and serve the previous endpoint's URL; (b) a transient mount-time failure resolves to the fallback and is cached, so a later explicit click never retries the updater. (a) is latent only — both call sites pass module-constant / value-stable arguments today, so the deps never actually change. (b) is functionally safe (the same-origin fallback still works) but defeats a later retry. Consider clearing `requestRef.current` in the effect cleanup, and not caching the fallback result so clicks can retry.

### P2 — Missing negative-path tests for the security validator
`__tests__/ios_download_button.test.tsx:178`, `__tests__/android_download_button.test.tsx`

`resolveSafeDownloadUrl` is the security gate, but tests only cover the valid-URL happy path and the network-error fallback — not the rejection paths that are the function's purpose. A future refactor weakening the protocol/`typeof` check would go uncaught. Add cases where `apiFetchJson` resolves to `{ url: "javascript:alert(1)" }`, `{ url: "file:///etc/passwd" }`, `{ url: "" }`, and `{ url: 42 }`, each asserting the fallback URL.

### P2 (nit) — Cross-origin `download` attribute may be ignored
`AndroidDownloadButton.tsx:71-74`

Browsers ignore the `download` attribute for cross-origin URLs. If the updater returns a CDN APK URL without `Content-Disposition: attachment`, the browser may navigate rather than download. For a non-renderable `.apk` this usually still results in a download, so impact is minor — worth confirming the CDN sets the header.

### P2 (nit) — No in-flight/disabled state on the direct-download action
`AndroidDownloadButton.tsx:71-81`

`void resolveDownloadUrl().then(startMobileDownload)` has no loading indicator or debounce; rapid clicks on a slow first fetch can spawn overlapping downloads. Minor UX.

## Coverage notes / blind spots

- Real-browser user-activation/popup-blocker behavior and the live updater's `Content-Disposition` headers cannot be verified statically (tests mock `click` synchronously in JSDOM); the gesture-timing and cross-origin-download items depend on runtime behavior.
- The updater response contract is assumed to be `{ url: string }`, consistent with the existing `common/updater/web/1.0` usage.
- Server-side validation of the updater's returned destination is out of scope for this diff and was not verified.
- Tests were not executed in this environment (dependencies not installed); assertions verified by reading source at the head SHA.

## Verdict

Well-scoped, spec-compliant restore of updater-backed mobile downloads with a safe scheme-validated fallback chain and a genuine security improvement over the prior baseline. No P0/P1 issues; every finding above is a non-blocking suggestion.

**Approve.**

Jerry-Xin (dismissed):
Re-review at head `d18416573f`. This push is a substantial, well-considered redesign of the updater flow: the hook now exposes explicit `loading` / `ready` / `error` states with a Retry control, request-ID guarding prevents stale/post-unmount updates, and the error path replaces the old silent same-origin fallback (tests now assert `/download/dmwork.apk` is gone). The security model is intact. There is one newly-introduced accessibility defect that should be fixed before merge.

### 🔴 Blocking

- Accessibility — `packages/dmworklogin/src/AndroidDownloadButton.tsx:47` and `packages/dmworklogin/src/IOSDownloadButton.tsx:27`: the error-state block (the `role="alert"` message and the Retry `<button>`) is rendered inside the QR container that carries `role="img"`. Per WAI-ARIA, descendants of a `role="img"` element are treated as presentational and pruned from the accessibility tree, so assistive technologies may not expose the error message or the Retry button — leaving screen-reader users with no announced way to recover from an updater failure. Fix: apply `role="img"` only when actually rendering the QR code (`status === "ready"`), or lift the loading/error states out of the `role="img"` container. Please add an assertion that the Retry control is exposed as an actionable button in the error state.

### 💬 Non-blocking

- 🟡 `packages/dmworklogin/src/mobileDownloadUpdater.ts:24`: endpoint construction and HTTP access live in the login UI package and import `WKApp` directly, which conflicts with the documented API service-layer convention. Consider moving the transport boundary into a service while the hook keeps ownership of UI state. (Carried from the prior review.)
- 🔵 Add focused tests for invalid updater payloads and unsupported URL schemes (e.g. a `javascript:`/`data:` `url`) and for late request completion after unmount. The implementation handles these correctly, but current coverage exercises them only indirectly via the network-error path.

### ✅ Highlights / re-confirmed
- Updater URLs are restricted to `http:`/`https:`, so dangerous schemes such as `javascript:`/`data:` from an updater response are rejected — security-sound.
- QR code and the Android direct-download link consistently use the one validated updater URL; a single request serves both.
- Loading and error states correctly avoid rendering a stale QR code or an active download link (the link is disabled via `aria-disabled` until `ready`).
- Request-ID guarding prevents stale or post-unmount responses from updating state.
- GitHub Releases manual-download fallback remains available and accessible during updater failures.
- The removal of the same-origin fallback is deliberate and enforced by tests.

Only the a11y item blocks merge; the security-relevant logic is sound.

yujiawei (changes requested):
# Code Review — PR #931 (octo-web)

## Summary

Re-review at head `d18416573fd254b7f67721d56d1f3552f01e2ac2`. Since the previous approved head (`3c5c14f7`), this PR was substantially redesigned. The `useMobileDownloadUrl` hook now exposes explicit `loading` / `ready` / `error` states plus a `retry` action, uses a request-ID guard to drop stale/post-unmount responses, and — deliberately — **removes the silent same-origin / TestFlight fallbacks**: on updater failure the popover shows an error state with a Retry control instead of quietly serving a stale URL. Scope is 11 files under `packages/dmworklogin/` (+398/−116). I re-verified against the current head and correct merge-base `b39924e6`; the earlier approval does not carry over.

Two of the previously noted P2s are resolved by this redesign, but the redesign introduces one **blocking accessibility defect**. Verdict below is **Changes Requested** for that single item; everything else is sound.

## Spec compliance

The redesign changes the documented contract, so this needs calling out even though it is an improvement:
- The original PR body says the same-origin APK and public TestFlight URLs are "retained as failure fallbacks." On this head they are **removed** — failure now yields an explicit error + Retry (Android also keeps the GitHub Releases button; iOS has only the in-popover Retry). This is a deliberate, tested change (tests assert `/download/dmwork.apk` is gone), and it is a reasonable product decision, but the PR description is now stale — please update it so the merged history reflects "error state + retry" rather than "same-origin fallback."
- Otherwise no over-build: updater endpoints unchanged, http/https scheme validation retained, QR + Android direct link share one validated URL.

## Findings

### 🔴 P1 (blocking) — Error message and Retry control are pruned from the accessibility tree
`AndroidDownloadButton.tsx:49-67`, `IOSDownloadButton.tsx:27-46`

The `role="alert"` error block and the Retry `<button>` are rendered as **descendants of the QR container that carries `role="img"`**:

```tsx
<div className="wk-login-mobile-popover-qr" data-status={status} role="img" aria-label={...}>
  {status === "error" && (
    <div className="wk-login-mobile-download-state" role="alert">
      <span>{t("download.addressLoadFailed")}</span>
      <WKButton onClick={retry}>{t("download.retry")}</WKButton>
    </div>
  )}
  ...
</div>
```

Per WAI-ARIA, an element with `role="img"` is a leaf node: its descendants are treated as presentational and pruned from the accessibility tree. Screen readers therefore may not announce the error text (the `role="alert"` is nullified) and may not expose the Retry button as an actionable control. Because this redesign removed the same-origin/TestFlight fallback, Retry is now the primary recovery path — and on iOS it is the **only** recovery path in the popover — so losing it for AT users is a functional regression, not a cosmetic one.

Fix: apply `role="img"` only when actually rendering the QR (`status === "ready"`), or lift the loading/error branches out of the `role="img"` container. Add a test asserting the Retry control is exposed as an actionable button in the error state (e.g. queryable by role `button` with the retry label).

### 💬 Non-blocking

- **P2 — HTTP boundary lives in UI package.** `mobileDownloadUpdater.ts:24` constructs the endpoint and calls `apiFetchJson` while importing `WKApp` directly from UI code, against the documented service-layer convention. Consider moving the transport into a service, leaving the hook to own UI state. (Carried from prior reviews; not introduced here.)
- **P2 — Add direct negative-path tests for the security validator.** `resolveSafeDownloadUrl` correctly rejects `javascript:` / `data:` / `file:` / empty / non-string / malformed values, but tests exercise these only indirectly via the network-error path. Add explicit cases (`{ url: "javascript:alert(1)" }`, `{ url: "" }`, `{ url: 42 }`) each asserting the error state, plus a late-completion-after-unmount case to lock in the request-ID guard.

## Re-confirmed as sound

- Scheme allowlist (`http:`/`https:` only) intact — dangerous schemes from an updater response are rejected.
- QR and Android direct link consume one validated updater URL from a single request; the direct link is a plain synchronous `<a href download>` rendered only when `ready` (this removes the earlier post-`await` user-gesture concern), and is `aria-disabled` otherwise.
- Request-ID guard (`requestIdRef`) drops stale and post-unmount responses — this resolves the earlier `requestRef`-never-invalidated P2; `retry` now re-fetches correctly.
- SSR-safe: initial state is `loading`, network/`window` access confined to `useEffect`.
- GitHub Releases manual download remains available on Android during failure.

## Coverage notes / blind spots

- The a11y pruning behavior is per the ARIA spec and static markup; exact screen-reader exposure varies by AT/browser but the `role="img"` leaf-node rule is well established — recommend the assertion above rather than relying on manual SR testing.
- Live updater `{ url }` contract and the CDN `Content-Disposition` for cross-origin APK downloads remain runtime-only, not verifiable here.
- Tests were not executed in this environment (dependencies not installed); assertions verified by reading source at the head SHA.

## Verdict

The redesign is a genuine improvement — explicit states, request-ID guarding, and an intentional move away from silent stale fallbacks, with the security model intact and two prior P2s resolved. The one blocking item is the accessibility regression: the error message and Retry control are unreachable for assistive tech because they live inside a `role="img"` element, and Retry is now the primary (iOS: only) recovery path. Fix that and re-request review.

**Request changes.**

botshen:
已在 `9bd193ae` 处理阻塞的可访问性问题：

- 仅在 updater 成功、实际渲染二维码时设置 `role="img"` 和二维码标签
- 加载态改用 `aria-busy`，失败态不再位于 `role="img"` 语义下
- Android/iOS 均新增断言，确认“重试”是可操作的原生按钮，且不属于任何 `[role="img"]` 后代
- `@octo/login` 全部 19 个测试文件 / 209 个测试通过，`pnpm i18n:check` 与格式检查通过

请基于最新 head 复审。

lml2468 (approved):
## Review Verdict（复审 @ 9bd193ae）：可合并（APPROVE）—— 维持 APPROVE

复审锚定新 head `9bd193ae`,对比我上次 APPROVE 的 `3c5c14f7`。新增两个 commit(`avoid stale download QR fallbacks` + `expose download retry to assistive tech`)修掉了 **Steve/Jerry-Xin/yujiawei 上轮提的 a11y 🔴**,byte 核实:

### ✅ a11y 回归 🔴(error/retry 被 role="img" 裁掉)—— 已修

- **问题**:错误提示 + Retry 按钮嵌在 `role="img"` 的二维码容器内 → role=img 子节点被从无障碍树裁掉 → updater 失败时屏幕阅读器用户收不到可恢复提示。
- **修复(byte 核实,Android:52 / iOS:31)**:`role={status === "ready" ? "img" : undefined}`——**role="img" 只在 ready(显示二维码)态生效**;error 态 role 为 undefined,里面的 `role="alert"` 错误提示 + Retry 按钮对 AT 可达。正是 Steve 建议的修法。
- 三态显式化:`loading`(Spin + `aria-busy`)/ `error`(`role="alert"` + Retry)/ `ready`(QR),互斥渲染,无残留旧二维码。

### ✅ 顺带修的(mochashanyao 🟡 stale cache)

- `useMobileDownloadUrl` 重构为 **discriminated union 三态 + `requestIdRef` 守卫**:`load()` 自增 requestId,只有 `requestId === requestIdRef.current` 才回写 → **过期/卸载后的迟到响应不会覆盖新状态**(unmount 时 `requestIdRef.current += 1` 作废在途请求)。`retry` 复用 `load`。上次 mochashanyao 提的 requestRef memo 不随参数清除的问题也一并解决(改为每次 load 新 requestId)。

### ✅ 安全 / 验证(保持)

- updater URL 仍 **scheme 白名单**(http/https),失败进 error 态(不再静默同源兜底——现在显式错误 + Retry,更透明);
- 新 i18n keys(`retry`/`addressLoadFailed`/`loadingAddress`/`iosQrLabel`)**双 locale 齐全**;
- `android_download_button.test.tsx` + `ios_download_button.test.tsx` 本地 **18 passed**;
- rebase 到 main tip;role 条件、三态 hook、requestId 守卫、i18n 均 byte 核实。

### 🟡 非阻塞(沿用)

- `resolveMobileUpdaterUrl` 在 UI 包直接读 `WKApp`(按规范宜走 service 层,前置遗留);
- updater/presigned URL 只校 scheme 不校 host(自家后端,同前结论);
- 缺畸形/非 http(s) URL 的直接负路径测试——建议补,不阻断。

### 结论

上轮的 a11y 🔴(role="img" 裁掉 error/retry)已干净修复(role 仅 ready 态),外加三态 + requestId stale/unmount 守卫,i18n 齐全、测试全绿、rebase 在位。**维持 APPROVE**。感谢 Steve/Jerry-Xin 抓出这条无障碍回归。

Jerry-Xin (dismissed):
The PR is relevant to octo-web and correctly implements updater-backed mobile download links without apparent blocking defects.

💬 Non-blocking

- 🟡 Warning — [mobileDownloadUpdater.ts](/Users/mlamp/.openclaw/tmp/pr-review-s3nai9ti/repo/packages/dmworklogin/src/mobileDownloadUpdater.ts:24) places endpoint construction and HTTP access inside the login UI package. `DEVELOPMENT.md` requires new API calls to be encapsulated in `packages/dmworkbase/src/Service/` using `APIClient.shared`. Consider moving this boundary into a dedicated service with service-level tests.
- 🔵 Suggestion — Add tests for malformed updater responses, disallowed URL schemes, and overlapping retry responses to directly cover the validation and request-ID logic in [mobileDownloadUpdater.ts](/Users/mlamp/.openclaw/tmp/pr-review-s3nai9ti/repo/packages/dmworklogin/src/mobileDownloadUpdater.ts:11).
- 🔵 Suggestion — Add deterministic Storybook loading, success, and failure states. The current stories depend on a live updater request, making visual review less reliable.

✅ Highlights

- Loading and failure states cannot expose stale QR codes or download links.
- Retry race handling prevents older requests from overwriting newer results.
- Updater URLs are restricted to HTTP(S).
- Android retains the GitHub fallback, while iOS removes only the direct TestFlight action as intended.
- Tests cover endpoint construction, successful rendering, failure handling, retry behavior, and removal of legacy links.

The focused test command could not be rerun because dependencies are not installed in the checkout (`vitest: command not found`).

Jerry-Xin (approved):
The PR is relevant to octo-web and correctly implements updater-backed mobile download links without apparent blocking defects. The prior CHANGES_REQUESTED (a11y blocker) is now resolved and superseded.

✅ a11y blocker fixed — In both `packages/dmworklogin/src/AndroidDownloadButton.tsx` and `packages/dmworklogin/src/IOSDownloadButton.tsx`, the QR container now sets `role={status === "ready" ? "img" : undefined}`. The `role="alert"` error message and Retry button are therefore no longer descendants of a `role="img"` node when the updater fails, so screen-reader users get a recoverable error + Retry. Tests assert `retryButton.closest('[role="img"]')` is null and that `role="img"` applies only in the ready state (plus `aria-busy="true"` in loading).

💬 Non-blocking

- 🟡 Warning — `packages/dmworklogin/src/mobileDownloadUpdater.ts:24` places endpoint construction and HTTP access inside the login UI package. `DEVELOPMENT.md` requires new API calls to be encapsulated in `packages/dmworkbase/src/Service/` using `APIClient.shared`. Consider moving this boundary into a dedicated service with service-level tests.
- 🔵 Suggestion — Add tests for malformed updater responses, disallowed URL schemes, and overlapping retry responses to directly cover the validation and request-ID logic in `packages/dmworklogin/src/mobileDownloadUpdater.ts:11`.
- 🔵 Suggestion — Add deterministic Storybook loading, success, and failure states. The current stories depend on a live updater request, making visual review less reliable.

✅ Highlights

- Loading and failure states cannot expose stale QR codes or download links (discriminated union; `downloadUrl` only present in the ready state; request-id guard prevents out-of-order overwrites).
- Updater URLs are restricted to HTTP(S) via `resolveSafeDownloadUrl`, blocking `javascript:`/`data:` schemes.
- Android retains the GitHub Releases fallback, while iOS removes only the direct TestFlight action as intended.
- Tests cover endpoint construction, successful rendering, failure handling, retry behavior, a11y scoping of `role="img"`, and removal of legacy links.

Note: the focused test command could not be rerun because dependencies are not installed in the checkout (`vitest: command not found`).

yujiawei (approved):
# Code Review — PR #931 (octo-web)

## Summary

Re-review at head `9bd193ae`. This push targets the single blocking accessibility item from the previous round (`d18416573f`): the error message and Retry control were rendered inside the QR container that carried `role="img"`, which per WAI-ARIA prunes descendants from the accessibility tree and left screen-reader users with no announced recovery path after an updater failure — the primary (on iOS, only) recovery path in the popover.

The fix is correct and precise:

- `AndroidDownloadButton.tsx:49-56` and `IOSDownloadButton.tsx:29-33`: `role="img"` / `aria-label` are now applied **only when `status === "ready"`** (i.e. when a QR code is actually rendered). During `loading` / `error` the container is a plain `<div>`, so the `role="alert"` error message and the Retry `<button>` are exposed as real, actionable nodes in the accessibility tree.
- `aria-busy={status === "loading" ? true : undefined}` is added so the loading state is announced instead of being a silent presentational image.

Verified against the current head with merge-base `b39924e6`; the earlier approval does not carry over.

## Spec compliance

**Spec: ✅**

- No missing work: the redesign's contract (updater-backed URL for both QR and direct link; `loading`/`ready`/`error` states; explicit error + Retry replacing the silent same-origin/TestFlight fallback) is intact, and the a11y regression it introduced is now closed.
- No over-build: this delta only adjusts `role`/`aria-*` conditionals and adds assertions. No new endpoints, flags, or behavior beyond the fix.
- i18n stays consistent: `en-US.json` and `zh-CN.json` carry an identical `download.*` key set; the removed `iosDirectDownload` key has no remaining references; new keys (`addressLoadFailed`, `loadingAddress`, `retry`) are wired in both buttons.

## Code Quality

**Quality: Approved**

The fix is minimal and matches the recommended remedy exactly. New tests lock it in:

- Both suites assert the loading container has `aria-busy="true"` and **no** `role="img"`.
- Both suites assert that in the error state the Retry control is a native `BUTTON` and `retryButton.closest('[role="img"]')` is `null` — a direct regression guard for the exact defect.
- The ready-state tests assert `role="img"` is present only once a QR is rendered.

I ran the affected suites at this head:

```
pnpm --filter @octo/login test -- src/__tests__/android_download_button.test.tsx src/__tests__/ios_download_button.test.tsx
Test Files  19 passed (19)
      Tests  209 passed (209)
```

Re-confirmed as sound from prior rounds:

- Scheme allowlist in `resolveSafeDownloadUrl` (`http:`/`https:` only) rejects `javascript:`/`data:`/`file:` and malformed/non-string updater responses — security model intact.
- QR and the Android direct link consume one validated URL from a single request; the direct link renders as a plain `<a href download>` only when `ready`, and is `aria-disabled` otherwise.
- Request-ID guard drops stale and post-unmount responses; `retry` re-fetches correctly.
- SSR-safe (initial state `loading`; network/`window` access confined to `useEffect`).
- CSS `.wk-login-mobile-popover-qr > svg` correctly scopes sizing to the direct QR child.

### Non-blocking (P2 — not required for merge)

- Direct negative-path unit tests for `resolveSafeDownloadUrl` are still exercised only indirectly via the network-error path. Adding explicit cases (`{ url: "javascript:alert(1)" }`, `{ url: "" }`, `{ url: 42 }`) plus a late-completion-after-unmount case would lock in the validator and request-ID guard directly. (Carried from prior rounds; the implementation is already correct.)
- The HTTP transport (`resolveMobileUpdaterUrl` + `apiFetchJson`, importing `WKApp`) lives in the login UI package rather than a service layer. Consider relocating the transport boundary later while the hook keeps UI state. (Carried; not introduced here.)

## Coverage notes / blind spots

- ARIA pruning behavior is per spec and static markup; exact SR exposure varies by AT/browser, but the `role="img"` leaf-node rule is well established and the new `closest('[role="img"]')` assertions are the right guard.
- Live updater `{ url }` contract and CDN `Content-Disposition` for cross-origin APK downloads are runtime-only, not verifiable here.
- This is the 4th review round on this PR, but each round was a genuine improvement and this head cleanly resolves the sole remaining blocker — no whack-a-mole pattern.

## Verdict

The one blocking accessibility defect from the previous round is fully and correctly resolved, with direct test coverage; the security model and updater flow remain sound; all 209 tests pass at this head. No P0/P1 remaining.

**Approve.**

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #931 (octo-web)

## Summary

This PR replaces the hardcoded mobile download URLs (a same-origin `/download/dmwork.apk` for Android and a public TestFlight link for iOS) with dynamic URLs fetched from a backend updater API (`common/updater/android/1.0` and `common/updater/ios/1.0.0`). A new shared hook `useMobileDownloadUrl` manages loading, success, and error states with a retry affordance. The popover content now conditionally renders a spinner, an error panel with retry button, or the QR code plus direct download link depending on the updater response.

The migration is well-executed: the shared helper is cleanly factored, race conditions are handled via a monotonic request-ID ref, URL validation rejects non-HTTP(S) protocols, and test coverage spans loading, success, error, and retry paths for both platforms. Two P2 items below.

## Verification

- ✅ **Race-condition protection** — `requestIdRef` in `useMobileDownloadUrl` (`packages/dmworklogin/src/mobileDownloadUpdater.ts:44–62`) correctly prevents stale responses from overwriting newer state. Cleanup on unmount increments the ref, invalidating in-flight requests.
- ✅ **URL validation** — `resolveSafeDownloadUrl` (`:11–21`) parses via `new URL(value, window.location.origin)`, accepts only `http:` and `https:` protocols, and returns `undefined` for anything else. Prevents `javascript:`, `data:`, and other dangerous schemes from reaching the QR code or download link.
- ✅ **Error propagation** — `apiFetchJson` throws `ApiFetchError` on non-2xx responses; `fetchMobileDownloadUrl` propagates this to the hook's rejection handler, which sets `status: "error"`. The error state renders a `role="alert"` panel with a retry button.
- ✅ **Test coverage** — both Android and iOS test suites mock `apiFetchJson` and verify loading → ready and loading → error transitions, including retry-after-error. The old hardcoded URL constants and direct-download assertions are properly removed.
- ✅ **Accessibility** — `aria-busy` on loading, `role="img"` only when QR code is rendered, `aria-disabled="true"` on the direct download link during non-ready states, `role="alert"` on error. Consistent across both platforms.
- ✅ **i18n parity** — new keys (`addressLoadFailed`, `loadingAddress`, `retry`) present in both `en-US.json` and `zh-CN.json`. Removed key (`iosDirectDownload`) removed from both locales.

## Findings

### P2 — Stale i18n label after iOS contract change (`packages/dmworklogin/src/i18n/en-US.json:89`)

`iosQrLabel` still reads "TestFlight installation QR code" (zh-CN: "TestFlight 安装二维码"), but the iOS download URL is no longer a hardcoded TestFlight link — it now comes from the updater endpoint and could be any `https://` URL. The label describes a specific provider that is no longer guaranteed by the code contract. If the updater ever returns a non-TestFlight URL (e.g. a direct IPA or a different distribution service), the accessibility label would be misleading to assistive-technology users.

**diff-scope**: new — the PR changes the iOS URL source from hardcoded TestFlight to dynamic updater, but the i18n key was not updated to match.

**Fix**: generalize the label, e.g. "iOS installation QR code" / "iOS 安装二维码", to match the Android side which already uses the generic "Scan to download".

### P2 — No AbortController signal; HTTP request continues after unmount (`packages/dmworklogin/src/mobileDownloadUpdater.ts:48`)

`fetchMobileDownloadUrl` calls `apiFetchJson` without an `AbortSignal`. When the popover closes mid-fetch, the `useEffect` cleanup increments `requestIdRef` (correctly preventing stale state writes), but the underlying HTTP request still completes in the background. On quick close+reopen, two requests for the same updater path can be in flight simultaneously.

This is low impact — the response is small JSON and stale results are discarded — but on slow networks the wasted bandwidth accumulates with repeated popover toggling.

**diff-scope**: new — this is the first call site introducing this async-fetch-without-abort pattern in the login module.

**Fix**: pass an `AbortController` signal through `apiFetchJson`'s `init` parameter and abort in the `useEffect` cleanup. `apiFetchJson` already accepts `init?: RequestInit`, so this is a drop-in addition:

```ts
const controller = new AbortController();
void fetchMobileDownloadUrl(updaterPath, controller.signal).then(...);
// cleanup:
return () => { controller.abort(); requestIdRef.current += 1; };
```

- `resolveMobileUpdaterUrl` trailing-slash normalization (`apiUrl.replace(/\/?$/, "/")`) correctly handles both `"/api/v1/"` and `"https://api.example.com/v1"` — verified against the existing `NavSettingsPanel` updater call site which uses the same `${apiURL}common/updater/...` pattern.
- Empty-string `apiURL` default: `"".replace(/\/?$/, "/")` produces `"/"`, yielding `"/common/updater/..."` — a valid relative URL that resolves against the page origin.
- CSS `data-status` selectors match the three hook states (`loading`, `error`, `ready`). The `[aria-disabled="true"]` direct-link styling correctly suppresses pointer interaction.
- `AndroidDownloadPopoverContent` and `IOSDownloadPopoverContent` are rendered only inside their respective `Popover` content props, so unmount/remount on popover open/close gives fresh hook state each time — no cross-session staleness.
- The `AndroidDownloadButton` retains the GitHub Releases fallback button, so users always have a working download path even if the updater is down.

## Data-flow trace

- **`downloadUrl` consumed by `<QRCodeSVG value={downloadUrl}>`**: sourced from `useMobileDownloadUrl` → `fetchMobileDownloadUrl` → `apiFetchJson` → `result?.url` → `resolveSafeDownloadUrl`. The URL is validated (non-empty string, http/https only) before reaching the QR code. ✅
- **`downloadUrl` consumed by `<a href={downloadUrl} download>`**: same source path as above. Only rendered when `status === "ready"`, which requires `downloadUrl` to be a validated string. ✅
- **`status` consumed by conditional rendering**: sourced from `MobileDownloadUrlState` discriminated union. The three states are exhaustive and mutually exclusive. ✅
- **`retry` consumed by error-panel button `onClick`**: sourced from `load` callback, which is stable per `updaterPath` via `useCallback`. Calling retry resets to `loading` and re-fetches. ✅

## Blind-spot checklist

- **C1 (dual-path parity)**: N/A — no paired add/remove or symmetric paths changed.
- **C2 (control-flow ordering)**: N/A — no shared logic reused across different call orderings.
- **C3 (authorization boundary)**: N/A — updater endpoints are on the public login page; no auth changes.
- **C4 (authorization lifecycle)**: N/A — no auth state involved.
- **C5 (build/note ≠ runtime)**: N/A — no build or packaging changes.
- **C6 (governance/docs self-consistency)**: N/A — no governance or policy changes.

[Octo-Q] verdict: APPROVE — well-structured migration from hardcoded URLs to dynamic updater with proper loading/error/accessibility handling. Two P2 items (stale TestFlight i18n label, missing AbortController) are non-blocking improvements.

Static analysis only at head `9bd193ae011c753863b98ede5a198851702c902e`; build and tests not executed in this environment.

botshen:
补充修复 `f8c77887`：测试环境 iOS updater 返回的 `url` 首字符是 Unicode 空白 `U+2004`（UTF-8 `e2 80 84`）。旧逻辑只用 `trim()` 判断非空，但把未清理的原值传给 `new URL()`，浏览器会把它解析成当前站点下的相对路径，例如 `https://im-test.deepminer.com.cn/%E2%80%84https://apps.apple.com/...`。

现已在 URL 解析前使用实际 `trim()` 后的字符串，并新增包含首尾 `U+2004` 的 iOS 二维码回归测试。`@octo/login` 19 个测试文件 / 209 个测试通过，i18n 与 `git diff --check` 通过。

PR #941, opened 2026-07-21
fix(chat): clear member-leave unread after auto-scroll

botshen:
## Summary

- advance the conversation browse cursor when a programmatic scroll reaches the real end of the loaded conversation
- persist unread clearing even when a short member-leave system row does not change `scrollTop` and therefore emits no DOM `scroll` event
- preserve unread state when newer messages are still outside the loaded window
- remove the previous close-conversation cache change, which addressed a different problem

## Root cause

Incoming messages update the VM's unread count before the conversation auto-scrolls. The read cursor was only advanced by the DOM `scroll` handler. For short system rows such as content type 1003 (member left), an already-bottomed viewport can keep the same `scrollTop`, so no scroll event fires and `conversation/clearUnread` is never called.

## Tests

- `pnpm --filter @octo/base exec vitest run src/Components/Conversation/__tests__/messageOrder.test.ts` (17 passed)
- `pnpm --filter @octo/base exec vitest run src/Components/ConversationList/__tests__/layout.test.tsx src/Components/Conversation/__tests__/messageOrder.test.ts` (21 passed)
- Conversation directory: 66 tests passed; 2 pre-existing suites fail during collection because lottie requires a built canvas implementation in jsdom

Fixes WS-4

### DIFF
--- a/packages/dmworkbase/src/Components/Conversation/__tests__/messageOrder.test.ts
+++ b/packages/dmworkbase/src/Components/Conversation/__tests__/messageOrder.test.ts
@@ -6,6 +6,7 @@ const sdkState = vi.hoisted(() => ({
     sendingQueues: new Map<number, unknown>(),
     channelInfos: new Map<string, any>(),
     syncMessages: vi.fn(),
+    markConversationUnread: vi.fn(() => Promise.resolve()),
 }))
 
 vi.mock("wukongimjssdk", () => {
@@ -95,9 +96,9 @@ vi.mock("../../../App", () => ({
         loginInfo: { uid: "me" },
         config: { pageSizeOfMessage: 30 },
         dataSource: { channelDataSource: { subscribers: () => Promise.resolve([]) } },
-        mittBus: { on: () => {}, off: () => {} },
+        mittBus: { on: () => {}, off: () => {}, emit: () => {} },
         conversationProvider: {
-            markConversationUnread: () => Promise.resolve(),
+            markConversationUnread: sdkState.markConversationUnread,
             syncMessages: sdkState.syncMessages,
         },
         shared: { currentSpaceId: "", notifyMessageDeleteListener: () => {} },
@@ -121,7 +122,7 @@ vi.mock("../../../Service/Provider", () => ({
 vi.mock("react-scroll", () => ({ animateScroll: { scrollToBottom: () => {} }, scroller: { scrollTo: () => {} } }))
 vi.mock("../../../Service/Const", () => ({
     EndpointID: {},
-    MessageContentTypeConst: { time: 1001, historySplit: 1002, rtcData: 1003 },
+    MessageContentTypeConst: { time: -1, historySplit: -3, removeMembers: 1003, rtcData: 9994 },
     OrderFactor: 10000,
     ChannelTypeCommunityTopic: 6,
 }))
@@ -218,6 +219,7 @@ describe("ConversationVM message ordering", () => {
         ConversationVM.sendQueue.clear()
         sdkState.sendingQueues.clear()
         sdkState.channelInfos.clear()
+        sdkState.markConversationUnread.mockClear()
         sdkState.syncMessages.mockReset()
         document.body.innerHTML = ""
     })
@@ -571,5 +573,53 @@ describe("ConversationVM message ordering", () => {
 
         expect(() => vm.updateReplyMessageContent({ messageID: "m1", contentEdit: "edited" } as any)).not.toThrow()
         expect(replyMsg.content.reply.content).toBe("edited")
+    })
+
+    it("clears unread after auto-scrolling a member-leave system message to the bottom", async () => {
+        const vm = new ConversationVM(channel)
+        const memberLeft = wrap({
+            clientMsgNo: "member-left",
+            messageSeq: 11,
+            timestamp: 200,
+            contentType: 1003,
+            content: { contentType: 1003, displayText: "Alice left the group" },
+            fromUID: "u1",
+        })
+        vm.messagesOfOrigin = [memberLeft]
+        vm.lastMessage = memberLeft
+        vm.browseToMessageSeq = 10
+        vm.unreadCount = 1
+        vm.pullupHasMore = false
+
+        vm.scrollToBottom(false)
+        await Promise.resolve()
+
+        expect(vm.browseToMessageSeq).toBe(11)
+        expect(vm.unreadCount).toBe(0)
+        expect(sdkState.markConversationUnread).toHaveBeenCalledWith(channel, 0)
+    })
+
+    it("keeps unread when scroll-to-bottom only reaches a partially loaded window", async () => {
+        const vm = new ConversationVM(channel)
+        const memberLeft = wrap({
+            clientMsgNo: "member-left-partial",
+            messageSeq: 11,
+            timestamp: 200,
+            contentType: 1003,
+            content: { contentType: 1003, displayText: "Alice left the group" },
+            fromUID: "u1",
+        })
+        vm.messagesOfOrigin = [memberLeft]
+        vm.lastMessage = memberLeft
+        vm.browseToMessageSeq = 10
+        vm.unreadCount = 1
+        vm.pullupHasMore = true
+
+        vm.scrollToBottom(false)
+        await Promise.resolve()
+
+        expect(vm.browseToMessageSeq).toBe(10)
+        expect(vm.unreadCount).toBe(1)
+        expect(sdkState.markConversationUnread).not.toHaveBeenCalled()
     })
 })
--- a/packages/dmworkbase/src/Components/Conversation/vm.ts
+++ b/packages/dmworkbase/src/Components/Conversation/vm.ts
@@ -2286,6 +2286,27 @@ export default class ConversationVM extends ProviderListener {
         }
         animateScroll.scrollToBottom(opts);
 
+        // Programmatic scrolling does not always emit a DOM scroll event. This
+        // is especially common for short system rows (for example type 1003,
+        // "member left") when the viewport is already at the bottom. In that
+        // case updateBrowseToMessageSeq() is never called and the conversation
+        // stays unread even though the new row is visible.
+        //
+        // Only advance the cursor when this is the real end of the conversation;
+        // otherwise a partially loaded window must keep its unread state.
+        if (!this.pullupHasMore) {
+            let latestLoadedMessageSeq = 0
+            for (const message of this.messagesOfOrigin) {
+                if (message.messageSeq > latestLoadedMessageSeq) {
+                    latestLoadedMessageSeq = message.messageSeq
+                }
+            }
+            if (latestLoadedMessageSeq > this.browseToMessageSeq) {
+                this.browseToMessageSeq = latestLoadedMessageSeq
+                void this.refreshNewMsgCount()
+            }
+        }
+
     }
 
     // 获取当前发送中的消息
--- a/packages/dmworkbase/src/Components/ConversationList/__tests__/layout.test.tsx
+++ b/packages/dmworkbase/src/Components/ConversationList/__tests__/layout.test.tsx
@@ -13,8 +13,6 @@ import {
 
 let ConversationList: typeof import("../index").default;
 let container: HTMLDivElement;
-const deleteConversation = vi.fn();
-const removeConversation = vi.fn();
 
 class MockChannel {
   channelID: string;
@@ -46,9 +44,6 @@ beforeAll(async () => {
           removeListener: vi.fn(),
           fetchChannelInfo: vi.fn(),
           getChannelInfo: vi.fn(),
-        },
-        conversationManager: {
-          removeConversation,
         },
       }),
     };
@@ -95,7 +90,7 @@ beforeAll(async () => {
         getChannelAvatarTag: () => "avatar",
       },
       apiClient: { put: vi.fn() },
-      conversationProvider: { deleteConversation },
+      conversationProvider: { deleteConversation: vi.fn() },
     },
   }));
 
@@ -178,9 +173,6 @@ beforeAll(async () => {
 
 beforeEach(() => {
   vi.clearAllMocks();
-  deleteConversation.mockReset();
-  deleteConversation.mockResolvedValue(undefined);
-  removeConversation.mockReset();
   container = document.createElement("div");
   document.body.appendChild(container);
 });
@@ -313,29 +305,5 @@ describe("ConversationList unread indicators", () => {
     expect(
       container.querySelector(".wk-conversationlist-item-indicators")
     ).toBeNull();
-  });
-});
-
-describe("ConversationList close chat", () => {
-  it("removes a system-message conversation locally after the server confirms deletion", async () => {
-    const channel = makeChannel("group-with-member-leave-message", 2);
-    const list = new ConversationList({ conversations: [] });
-
-    await list.onCloseChat(channel as any);
-
-    expect(deleteConversation).toHaveBeenCalledWith(channel);
-    expect(removeConversation).toHaveBeenCalledWith(channel);
-  });
-
-  it("keeps the local conversation when server deletion fails", async () => {
-    deleteConversation.mockRejectedValue(new Error("delete failed"));
-    const channel = makeChannel("group-delete-failed", 2);
-    const list = new ConversationList({ conversations: [] });
-
-    await expect(list.onCloseChat(channel as any)).rejects.toThrow(
-      "delete failed"
-    );
-
-    expect(removeConversation).not.toHaveBeenCalled();
   });
 });
--- a/packages/dmworkbase/src/Components/ConversationList/__tests__/layout.test.tsx
+++ b/packages/dmworkbase/src/Components/ConversationList/__tests__/layout.test.tsx
@@ -13,6 +13,8 @@ import {
 
 let ConversationList: typeof import("../index").default;
 let container: HTMLDivElement;
+const deleteConversation = vi.fn();
+const removeConversation = vi.fn();
 
 class MockChannel {
   channelID: string;
@@ -44,6 +46,9 @@ beforeAll(async () => {
           removeListener: vi.fn(),
           fetchChannelInfo: vi.fn(),
           getChannelInfo: vi.fn(),
+        },
+        conversationManager: {
+          removeConversation,
         },
       }),
     };
@@ -90,7 +95,7 @@ beforeAll(async () => {
         getChannelAvatarTag: () => "avatar",
       },
       apiClient: { put: vi.fn() },
-      conversationProvider: { deleteConversation: vi.fn() },
+      conversationProvider: { deleteConversation },
     },
   }));
 
@@ -173,6 +178,9 @@ beforeAll(async () => {
 
 beforeEach(() => {
   vi.clearAllMocks();
+  deleteConversation.mockReset();
+  deleteConversation.mockResolvedValue(undefined);
+  removeConversation.mockReset();
   container = document.createElement("div");
   document.body.appendChild(container);
 });
@@ -305,5 +313,29 @@ describe("ConversationList unread indicators", () => {
     expect(
       container.querySelector(".wk-conversationlist-item-indicators")
     ).toBeNull();
+  });
+});
+
+describe("ConversationList close chat", () => {
+  it("removes a system-message conversation locally after the server confirms deletion", async () => {
+    const channel = makeChannel("group-with-member-leave-message", 2);
+    const list = new ConversationList({ conversations: [] });
+
+    await list.onCloseChat(channel as any);
+
+    expect(deleteConversation).toHaveBeenCalledWith(channel);
+    expect(removeConversation).toHaveBeenCalledWith(channel);
+  });
+
+  it("keeps the local conversation when server deletion fails", async () => {
+    deleteConversation.mockRejectedValue(new Error("delete failed"));
+    const channel = makeChannel("group-delete-failed", 2);
+    const list = new ConversationList({ conversations: [] });
+
+    await expect(list.onCloseChat(channel as any)).rejects.toThrow(
+      "delete failed"
+    );
+
+    expect(removeConversation).not.toHaveBeenCalled();
   });
 });
--- a/packages/dmworkbase/src/Components/ConversationList/index.tsx
+++ b/packages/dmworkbase/src/Components/ConversationList/index.tsx
@@ -903,14 +903,9 @@ export default class ConversationList extends Component<
       })
   }
 
-  async onCloseChat(channel: Channel) {
+  onCloseChat(channel: Channel) {
     // 关闭聊天
-    await WKApp.conversationProvider.deleteConversation(channel);
-    // conversationDeleted 是异步 CMD，不能作为本地列表更新的唯一触发源。
-    // 例如他人退群的系统消息已把会话更新到最近列表时，推送缺失会
-    // 让该会话永久残留。服务端删除成功后立即清理 SDK cache；迟到的
-    // conversationDeleted 再次删除是幂等的。
-    WKSDK.shared().conversationManager.removeConversation(channel);
+    WKApp.conversationProvider.deleteConversation(channel);
   }
 
   async onClearMessages(channel: Channel) {
--- a/packages/dmworkbase/src/Components/ConversationList/index.tsx
+++ b/packages/dmworkbase/src/Components/ConversationList/index.tsx
@@ -903,9 +903,14 @@ export default class ConversationList extends Component<
       })
   }
 
-  onCloseChat(channel: Channel) {
+  async onCloseChat(channel: Channel) {
     // 关闭聊天
-    WKApp.conversationProvider.deleteConversation(channel);
+    await WKApp.conversationProvider.deleteConversation(channel);
+    // conversationDeleted 是异步 CMD，不能作为本地列表更新的唯一触发源。
+    // 例如他人退群的系统消息已把会话更新到最近列表时，推送缺失会
+    // 让该会话永久残留。服务端删除成功后立即清理 SDK cache；迟到的
+    // conversationDeleted 再次删除是幂等的。
+    WKSDK.shared().conversationManager.removeConversation(channel);
   }
 
   async onClearMessages(channel: Channel) {


lml2468 (dismissed):

审查锚定 head `defa9d0d`。2 文件 +40/−3,`ConversationList`。修「他人退群系统消息更新最近列表后,关闭会话仍残留」。已跑测试并核实失败保留语义。

### ✅ 两道闸门均通过

**规格符合**:与描述、WS-4 一致——不再只依赖异步 `conversationDeleted` CMD(推送缺失会让群会话永久残留),服务端确认删除后立即清 SDK cache;删除失败则保留本地会话可重试。无 over-build。

**修法正确(顺序是关键)**:
- `onCloseChat` 改 async:先 `await deleteConversation(channel)`(服务端),**成功后**才 `WKSDK.conversationManager.removeConversation(channel)`(本地 cache);
- **失败保留语义成立**:server 删除 reject → await 抛出 → 本地 `removeConversation` **被跳过** → 会话保留可重试;
- 迟到的 `conversationDeleted` CMD 再删是幂等的(注释说明),不会双重出错;
- `removeConversation` 确在 WKSDK `conversationManager` 上存在。

### ✅ 验证

- `layout.test.tsx` 本地 **6 passed**,含两条精确回归:server 确认后 `deleteConversation` + `removeConversation` 都调用;server 失败时 `removeConversation` **不被调用**(锁定保留语义);
- 删除→本地清理顺序、失败跳过、removeConversation 存在均 byte 核实;
- rebase 到 main tip,`git merge-tree` **0 冲突**。

### 🟡 非阻塞(本 PR 引入,建议 fast-follow)

- **floating promise / 静默失败**:两处调用方(index.tsx:1300 / :1412 的 Modal `onOk`)仍是 `this.onCloseChat(channel)` fire-and-forget、无 `.catch`。此前 `onCloseChat` 是 sync-void,现在改成 async-可 reject —— server 删除失败时会产生**未处理的 promise rejection**,且用户看不到错误提示(只是会话保留但无反馈)。建议在 `onOk` 里 `void this.onCloseChat(channel).catch(e => Toast.error(...))`,给失败一个用户可见的提示。功能主路径不受影响(成功路径与保留语义都对),故非阻塞。
- :1412 处 `onCloseChat` 后紧跟 `onClearMessages` —— 现在 onCloseChat 是 async,两者不再有序;若 clear 依赖 close 完成,顺序语义需确认(当前看是独立操作,应无碍,但建议一并 await 收敛)。

### 结论

核心修复对症(服务端确认→本地清理→失败保留),顺序正确、幂等、测试精确覆盖成功/失败两路径。唯一 🟡 是调用方未处理新引入的 async 拒绝(静默失败 + unhandled rejection),建议 fast-follow 加 catch+toast,不阻断主功能。**APPROVE**。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

yujiawei (dismissed):
# Code Review — PR #941 (octo-web)

## Verdict: APPROVED

The change is correct and well-targeted. Closing a chat previously relied solely on the asynchronous `conversationDeleted` command to prune the recent-conversation list. When that push is missing or races with another member's leave system-message (which re-surfaces the conversation into recents), the closed conversation lingers forever. Awaiting the server delete and then removing the entry from the SDK cache directly is the right fix, and it correctly preserves the local entry on server failure so the user can retry.

## 1. Correctness — looks good

- **Ordering & idempotency:** `await deleteConversation(channel)` then `conversationManager.removeConversation(channel)` is correct. A late `conversationDeleted` push that removes the same channel is idempotent — `removeConversation` splices on the first matching channel and no-ops otherwise, so the double removal is harmless.
- **Reactivity is sound.** `conversationManager.removeConversation` is the established re-render mechanism in this codebase: the same call is used by the `conversationDeleted` CMD handler and by `removeLocalConversationAndCloseIfOpen` (`packages/dmworkbase/src/module.tsx:154`, `:460`). It emits `ConversationAction.remove`, which the Chat view-model listener handles (`packages/dmworkbase/src/Pages/Chat/vm.ts:298-302` → `removeConversation` → `notifyListener()` at `:408-417`). So the list updates without depending on the async push.
- **Failure path is correct.** Because `removeConversation` runs only after the `await`, a rejected server delete short-circuits before any local mutation — the conversation stays visible, matching the stated intent and the new test.
- **"Close and clear" ordering is preserved.** At the combined call site, `onCloseChat(channel)` awaits the delete before pruning, so the synchronously-following `onClearMessages(channel)` still finds the conversation via `findConversation` — same effective ordering as before this PR.

## 2. Non-blocking observations (pre-existing; not introduced here)

These are worth noting but do **not** block merge, because the pre-PR code had the same behavior — this PR does not regress them:

- **No user-facing feedback when the server delete fails.** Both call sites invoke `onCloseChat` as `onOk: () => { this.onCloseChat(channel); }`, returning `undefined`. `wkConfirm.runAction` (`packages/dmworkbase/src/Components/WKModal/confirm.tsx:41-51`) only attaches its `.catch`/loading handling when the `onOk` callback *returns* a thenable — so the rejection floats: the dialog closes as if successful and no error toast is shown. Note this is unchanged from before the PR (the old `onCloseChat` also fired `deleteConversation` as an uncaught floating promise). **Suggestion (follow-up):** mirror the pattern already used in `ChannelSettings` delete-and-exit (`packages/dmworkbase/src/module.tsx:2257-2270`) — return the promise from `onOk` (so `wkConfirm` shows the loading state and keeps the dialog open on failure) and surface a `Toast.error`. This would also naturally cover the double-click case (`disabled`/`loading` while pending).
- **Combined "close and clear" fires two independent async operations** (`onCloseChat` + `onClearMessages`) whose failures are each swallowed, so a partial failure (e.g., cleared but not closed) is not reported. Again pre-existing; a follow-up could sequence them and report partial failure.

## 3. Test coverage — adequate for this change, with a gap to note

The two added unit tests correctly pin the new contract: `removeConversation` is called with the channel after a successful delete, and **not** called when the delete rejects. Good regression protection for the exact bug being fixed.

- **Nit:** the tests call `list.onCloseChat(...)` directly and mock `conversationManager` as a bare spy, so they do not exercise the `wkConfirm` call sites or assert that the rendered list actually updates (no listener notification / DOM assertion). An implementation that called the method correctly but broke integration (list not re-rendering) would still pass. This mirrors the existing test style in this file and is acceptable, but an integration-level assertion (mount + assert removal) would strengthen it.

## Coverage note

- **What was verified independently:** the diff at head SHA against its parent commit, both `onCloseChat` call sites, `wkConfirm`/`confirm.tsx` promise handling, the `removeConversation` reactivity chain through the Chat view-model, the idempotency of a late `conversationDeleted` push, and the "close and clear" ordering.
- **Not directly runnable:** node_modules (`wukongimjssdk`) was not installed in the review checkout, so the SDK's internal `removeConversation` → `ConversationAction.remove` emission was confirmed via repository consumers rather than executed. The full vitest suite was not re-run here; the PR reports 13/13 passing for the touched test dir.
- This review incorporated a cross-model adversarial pass; findings converged on the same non-blocking observations above.

Jerry-Xin (dismissed):
APPROVE — targeted fix for stale recent-conversation entries after close; removal is correct and failure-safe.

## Summary
`onCloseChat` now `await`s the server-side delete and then eagerly clears the SDK conversation cache, so a missing/late `conversationDeleted` CMD no longer leaves a permanently stale entry in the recent list (e.g. after a member-leave system message re-surfaces the conversation). Verified at head `defa9d0`.

## Correctness (byte-verified at head)
- Ordering is right: server delete first, cache removal second.
  `packages/dmworkbase/src/Components/ConversationList/index.tsx:906` —
  `await WKApp.conversationProvider.deleteConversation(channel);`
  then `WKSDK.shared().conversationManager.removeConversation(channel);` (line 911).
- Correct target, no over-removal: `deleteConversation` (packages/dmworkdatasource/src/conversation.ts:7) is `DELETE conversations/{channelID}/{channelType}`, and `removeConversation(channel)` removes exactly that same `channel` from the SDK cache — no id mismatch, no risk of dropping an active/other conversation.
- Failure-safe: if the server delete rejects, the `await` throws before `removeConversation` runs, so the local entry is preserved. The added test `keeps the local conversation when server deletion fails` asserts `removeConversation` is NOT called on rejection.

## Boundary / state
- Empty recent list or closing a conversation not in the list → `removeConversation` is a no-op; safe.
- Late `conversationDeleted` push after eager removal → idempotent re-delete, as documented in the inline comment.
- `closeAndClear` menu path calls `onCloseChat(channel)` then `onClearMessages(channel)`; `onCloseChat` becoming `async` is fine because callers invoke it fire-and-forget with no dependency on its completion, and clearing does not depend on close finishing.
- No unread-count desync introduced: this change only removes the conversation from cache; unread rendering is unchanged.

## Persistence
- No localStorage concern: the recent list is SDK-managed. The only localStorage state here (`expandedGroupIds`) is untouched, and its existing writes already sit inside try/catch.

## Tests
- New `ConversationList close chat` block covers both the success path (`deleteConversation` then `removeConversation` both called with the channel) and the failure path (rejection preserves local entry). Mocks reset per test.

## Notes
- 🟡 Non-blocking: callers of `onCloseChat` do not await it; if you ever want to surface delete failures to the user (toast on the rejected promise), a `.catch` at the call sites would be the place. Not required for this fix.

No orphaned references (only in-file callers), no secrets, no config drift. LGTM.

lml2468 (approved):
## Review Verdict（复审 @ c2574e81）：可合并（APPROVE）

复审锚定新 head `c2574e81`。**注意:本 PR 范围较我上次 APPROVE(defa9d0d)已变更**——当时的 ConversationList `onCloseChat`(await 服务端删除→清本地 cache)改动已不在 diff 里,现在整个 PR 只剩 **2 文件**(`Conversation/vm.ts` + `messageOrder.test.ts`),换成了「系统消息自动滚到底后清未读」的修复。按当前真实 2 文件范围重审:

### ✅ 两道闸门均通过

**规格符合**:与新标题、#945 场景一致——他人退群等系统消息(如 1003)到达且视口已在底部时,程序化滚动不触发 DOM scroll 事件 → `updateBrowseToMessageSeq` 不调 → 会话残留未读红点。修此。无 over-build。

**修法正确(guard 是关键)**:
- `scrollToBottom` 末尾:`if (!this.pullupHasMore)` 时,取 `messagesOfOrigin` 最大 seq,`> browseToMessageSeq` 才推进并 `refreshNewMsgCount()`;
- **`!pullupHasMore` guard 正确**:只在窗口真到底(无更多可拉)时才标记已读;部分加载窗口保留未读——避免误清;
- 与既有模式一致:`browseToMessageSeq = 最新 seq` → `refreshNewMsgCount`(内部 `lastMessage.messageSeq <= browseToMessageSeq → unread=0` 并同步 conversation.unread),与 :944-948 clearChannelMessages 同款语义;
- `void refreshNewMsgCount()` 是 fire-and-forget 但该方法自身吞错、无副作用外泄,可接受。

### ✅ 验证

- `messageOrder.test.ts` 新增精确回归:`clears unread after auto-scrolling a member-leave system message to the bottom`——contentType 1003 + browseToMessageSeq=10 + unreadCount=1 + pullupHasMore=false,断言滚动后未读清零;
- `messageOrder.test` + `ConversationList/layout.test` 本地 **21 passed**;
- guard 正确性、refreshNewMsgCount 语义、与既有 browseToMessageSeq 模式一致均 byte 核实;
- 落后 main 1 commit,`git merge-tree` **0 冲突**。

### 结论

按当前 2 文件范围:系统消息自动滚底后清未读的修复对症、`!pullupHasMore` guard 防误清、语义与既有一致、回归测试精确覆盖 1003 场景、测试全绿、0 冲突。无阻塞。**APPROVE**。

(说明:我上次 APPROVE 针对的 onCloseChat 服务端确认后清缓存那版已不在本 PR;若那部分改到别的 PR,请单独触发。当前 head 的 2 文件改动独立成立、可放行。)

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #941 (octo-web)

## Summary

This PR fixes a UX bug where the unread-count badge persists after a programmatic scroll-to-bottom lands on a short system message (e.g. "member left", content type 1003). The root cause is that `animateScroll.scrollToBottom()` does not always fire a DOM scroll event for small row insertions near the viewport bottom, so `updateBrowseToMessageSeq()` — the scroll-listener path that advances the read cursor — never runs. The fix adds a synchronous fallback in `scrollToBottom()` that directly advances `browseToMessageSeq` to the max loaded `messageSeq` when the conversation is fully loaded (`!pullupHasMore`), then delegates to the existing `refreshNewMsgCount()` to clear unread and persist via `markConversationUnread`.

## Verification

- ✅ **Data-flow trace** — `scrollToBottom()` → `browseToMessageSeq = latestLoadedMessageSeq` → `refreshNewMsgCount()` → `unreadCount = lastMessage.messageSeq - browseToMessageSeq = 0` → `markConversationUnread(channel, 0)` → `sidebar-reload`. The chain is correct and matches the existing scroll-listener path at `vm.ts:1564`.
- ✅ **Guard correctness** — `!this.pullupHasMore` prevents cursor advancement when history is partially loaded. `latestLoadedMessageSeq > this.browseToMessageSeq` prevents redundant writes. Both guards verified at `vm.ts:2297`–`:2305`.
- ✅ **messagesOfOrigin traversal** — The O(n) scan over `this.messagesOfOrigin` to find max `messageSeq` matches the existing pattern at `vm.ts:2112`–`:2116`. No regression for typical conversation sizes.
- ✅ **Caller impact** — `scrollToBottom()` is called from 5 sites (typing indicator `:995`, self-send `:1541`, receive-while-at-bottom `:1547`, reconnect `:1925`, user-click `:1218`). The cursor advancement is safe for all: it only advances when the conversation is fully loaded and new messages exist beyond the cursor.
- ✅ **scrollToBottomIfNeedPull bypass** — This method at `vm.ts:1619` calls `animateScroll.scrollToBottom` directly, bypassing the new logic. This is correct: it only runs when `pullupHasMore` is true, which is exactly the case the new guard excludes.
- ✅ **Test mock alignment** — `MessageContentTypeConst` mock updated from `{ time: 1001, historySplit: 1002, rtcData: 1003 }` to `{ time: -1, historySplit: -3, removeMembers: 1003, rtcData: 9994 }`. These now match production values in `Const.ts:48,50,66,78` — the old mock had incorrect values. No existing tests use `rtcData`, `time`, or `historySplit` content types directly, so no regression.
- ✅ **mittBus.emit mock** — Added `emit: () => {}` to the mock, needed because `refreshNewMsgCount()` at `vm.ts:1612` emits `"sidebar-reload"` on unread clear. Without this, the test would throw.

Static analysis only at head `c2574e815667eaa824ec023c2a167df497df27d5`; build and tests not executed in this environment.

## Findings

### Nit — PR title describes a broader scope than the diff (`PR metadata`)

The PR title says "remove stale group conversations from recent tab after member leaves" but both commits address unread-count clearing after programmatic scroll, not conversation-list removal. The commit messages (`fix(chat): remove closed conversations from local cache`, `fix(chat): clear unread after programmatic bottom scroll`) are more accurate. Consider aligning the PR title to avoid confusion during future git-log archaeology.

## Data-Flow Trace

|---|---|---|
| `browseToMessageSeq` (new write at `:2304`) | `messagesOfOrigin[i].messageSeq` (server-assigned seq, populated at `:1536`, `:1780`, `:1863`, `:2048`) | ✅ — max scan at `:2299`–`:2302` correctly finds the latest loaded seq |
| `refreshNewMsgCount()` (called at `:2306`) | `this.lastMessage.messageSeq` (set at `:1560` via `updateLastMessageIfNeed`) and `this.browseToMessageSeq` (just advanced) | ✅ — when `lastMessage.messageSeq <= browseToMessageSeq`, `unreadCount` becomes 0 |
| `markConversationUnread(channel, 0)` (via `refreshNewMsgCount` at `:1601`) | `this.channel` (constructor param) | ✅ — same path as existing scroll-listener trigger |
| `sidebar-reload` emit (`:1612`) | gated by `shouldClear` = `unreadCount === 0 && oldUnreadCount > 0` | ✅ — only fires on actual transition from unread to read |

## Blindspot Checklist

- **C1 (dual-path parity)** — N/A. No paired add/remove or subscribe/unsubscribe paths changed.
- **C2 (control-flow ordering)** — Clear. The new code runs synchronously after `animateScroll.scrollToBottom(opts)`. No double-invocation risk: if a real DOM scroll event later fires `updateBrowseToMessageSeq()`, it will find `browseToMessageSeq` already at max and be a no-op.
- **C3 (auth boundary)** — N/A. No permission/auth changes.
- **C5 (build ≠ runtime)** — N/A. No build/config changes.
- **C6 (docs/governance)** — N/A.

## Cross-Round Blocker Review

## Verdict: APPROVED

Clean, well-scoped fix with correct guards and good test coverage. The `!pullupHasMore` gate prevents false cursor advancement on partially loaded conversations, and the `latestLoadedMessageSeq > browseToMessageSeq` check prevents redundant writes. The test mock corrections (aligning `MessageContentTypeConst` with production values) are a bonus. The only nit is the PR title mismatch — cosmetic, not blocking.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; code correctly guards cursor advancement and delegates to the existing `refreshNewMsgCount` chain.

yujiawei (changes requested):
# Code Review — PR #941 (octo-web)

**Verdict: CHANGES_REQUESTED**

Reviewed against head SHA `c2574e815667eaa824ec023c2a167df497df27d5`, merge-base `06fb431c`. Two independent reviews (line-level + whole-repo) were absorbed; findings below are de-duplicated and severity-normalized.

## 1. Spec / PR-description compliance — ❌

This PR contains two commits, and the second one **fully reverts the first**:

- `defa9d0d` — *"remove closed conversations from local cache"*: adds `await deleteConversation` + `WKSDK.shared().conversationManager.removeConversation()` in `ConversationList.onCloseChat` (`ConversationList/index.tsx`) plus tests. **This is the change the PR title/body/`## Tests` section describe.**
- `c2574e81` (head) — *"clear unread after programmatic bottom scroll"*: **reverts** the `ConversationList` change back to the original `onCloseChat(channel) { WKApp.conversationProvider.deleteConversation(channel); }`, and instead adds unread-clearing logic in `Conversation/vm.ts` `scrollToBottom()`.

The **net `merge-base..HEAD` diff touches only** `Conversation/vm.ts` and `Conversation/__tests__/messageOrder.test.ts`. The `ConversationList` files are back to baseline (net diff empty).

Consequences that must be reconciled before merge:

- The **title** ("remove stale group conversations from recent tab after member leaves") and **Summary** (remove from SDK cache after server confirms close; stop relying on the async `conversationDeleted` command; preserve on failure) describe the *reverted* approach. The merged code does something different: it clears the **unread count** on a conversation when a programmatic bottom-scroll lands on newly loaded rows.
- The **`## Tests`** section cites `ConversationList/__tests__`, which are no longer part of the net change. The actual net-diff test file is `Conversation/__tests__/messageOrder.test.ts`.
- Removing an entry from the recent tab (title's claim) and zeroing an unread badge (what the code does) are **different user-visible outcomes**. If the linked issue is specifically about a conversation *lingering in the recent list*, the merged diff does not obviously address it.

Please clarify: was the revert of the `ConversationList` cache-removal intentional (then update title/body/tests/linked-issue to describe the unread-count fix), or was it an accidental force-push that dropped commit 1? Right now a reviewer/merger cannot verify the diff against the stated intent.

## 2. Code review of the net diff (`vm.ts` `scrollToBottom`)

### 🔴 Blocking — unread is cleared from *loaded* messages, not *visible* ones (`vm.ts:2297-2308`)

```ts
if (!this.pullupHasMore) {
    let latestLoadedMessageSeq = 0
    for (const message of this.messagesOfOrigin) {
        if (message.messageSeq > latestLoadedMessageSeq) {
            latestLoadedMessageSeq = message.messageSeq
        }
    }
    if (latestLoadedMessageSeq > this.browseToMessageSeq) {
        this.browseToMessageSeq = latestLoadedMessageSeq
        void this.refreshNewMsgCount()
    }
}
```

The established invariant is that the read cursor only advances to the message actually at the bottom of the viewport — `updateBrowseToMessageSeq()` advances `browseToMessageSeq` to `lastVisiableMessage.messageSeq` (`Conversation/index.tsx:2200-2210`). This new block instead advances the cursor to the newest **loaded** message (`latestLoadedMessageSeq`) whenever the full tail is loaded (`!pullupHasMore`), regardless of whether it entered the viewport.

`scrollToBottom()` is not called only for the member-leave case; it fires for **other users' incoming messages** while the user is at the bottom:

```ts
// vm.ts:1541-1547  (appendMessage, other user's message)
} else {
    if (this.showScrollToBottomBtn) {
        this.notifyListener()          // preserves unread (good)
    } else {
        this.scrollToBottom(true)      // now also clears unread
    }
}
```

…and for the typing indicator (`vm.ts:986-996`). `showScrollToBottomBtn === false` reflects a prior scroll-position calculation; it does **not** prove the new row was seen. So an incoming message that arrives while the tab is backgrounded/unfocused (position still "at bottom"), or a 200 ms smooth scroll the user interrupts by scrolling up, will now zero the unread state for messages the user never actually viewed. This is a broadening of read semantics well beyond the member-leave symptom the change targets, and it regresses the deliberate "visible, not merely loaded" invariant.

Suggested directions: gate the cursor advance on real visibility/focus (e.g. `document.visibilityState === "visible"`, or defer to the post-scroll `updateBrowseToMessageSeq` visibility check), or scope this to the completion of the scroll rather than firing synchronously right after `animateScroll.scrollToBottom()`.

### 🟡 Non-blocking — fire-and-forget clear has no ordering guard (`vm.ts:2306`)

`void this.refreshNewMsgCount()` starts an async chain that may `await WKApp.conversationProvider.markConversationUnread(channel, 0)` and then emit `sidebar-reload` (`vm.ts:1601-1613`). If a newer message raises unread while that request is in flight, the older clear can resolve afterward and push the stale `0` / `sidebar-reload` over the newer state. There is no generation token or serialization. This pattern pre-exists on the DOM path, but the new synchronous call site makes the race easier to hit. Consider guarding with a monotonic sequence check before applying the stale result.

### 🟡 Non-blocking — tests don't exercise the risky path (`messageOrder.test.ts:577-625`)

The two new tests seed VM fields directly and call `scrollToBottom(false)` with `react-scroll` mocked to a no-op. They correctly prove the new branch and the `pullupHasMore` guard (and the mocked `removeMembers: 1003` constant matches `Service/Const.ts`). But they do **not** go through the real `appendMessage` → `showScrollToBottomBtn` → auto-scroll path, DOM visibility, or smooth-scroll timing — i.e. exactly the conditions in the blocking finding. Please add coverage for an other-user incoming message while at bottom, so the read-semantics behavior is pinned by a test.

## 3. Notes / non-issues verified

- **Not a bug:** this block does not "bypass a read lifecycle hook." The DOM `updateBrowseToMessageSeq` performs the identical two operations (set `browseToMessageSeq`, call `refreshNewMsgCount()`); the only real difference is the *source* of the sequence number (visible vs. loaded), which is the blocking finding above.
- The `O(n)` scan of `messagesOfOrigin` correctly ignores seq `0`/negative local rows and is fine at page size.
- All 17 tests in `messageOrder.test.ts` pass locally.

## Summary

Blocking on two independent grounds:
1. The merged diff diverges from the PR's stated approach (commit 2 reverts commit 1; title/body/tests/linked-issue describe the reverted change) — please reconcile intent and metadata.
2. The unread-clearing logic marks *loaded* messages read without confirming visibility, affecting all auto-scroll callers (incoming other-user messages, typing), not just the member-leave case.

Jerry-Xin (changes requested):
Re-review at head `c2574e815667`. My earlier APPROVE was anchored to the previous commit (`defa9d0d`, "remove closed conversations from local cache"), which the current head intentionally reverts per the PR body ("remove the previous close-conversation cache change, which addressed a different problem"). Re-anchoring to the live head, the effective change is now the unread-clearing logic in `packages/dmworkbase/src/Components/Conversation/vm.ts` `scrollToBottom()`, and it has a real blocking issue.

## 🔴 Blocking — cursor advances on *loaded*, not *visible*, messages (`packages/dmworkbase/src/Components/Conversation/vm.ts:2297-2308`)

```ts
if (!this.pullupHasMore) {
    let latestLoadedMessageSeq = 0
    for (const message of this.messagesOfOrigin) {
        if (message.messageSeq > latestLoadedMessageSeq) {
            latestLoadedMessageSeq = message.messageSeq
        }
    }
    if (latestLoadedMessageSeq > this.browseToMessageSeq) {
        this.browseToMessageSeq = latestLoadedMessageSeq
        void this.refreshNewMsgCount()
    }
}
```

The established invariant is that `browseToMessageSeq` only advances to the message actually at the bottom of the viewport (the DOM `updateBrowseToMessageSeq` handler advances it to the last *visible* message). This new block instead advances the cursor to the newest *loaded* message whenever the full tail is loaded (`!pullupHasMore`), regardless of whether that row entered the viewport.

`scrollToBottom()` is not scoped to the member-leave symptom — it fires for other users' incoming messages while the user is considered at the bottom, and for the typing indicator:

- `appendMessage()` other-user branch (`vm.ts:1541-1547`): when `showScrollToBottomBtn === false`, it calls `scrollToBottom(true)`.
- typing listener (`vm.ts:985-996`): when `showScrollToBottomBtn === false`, it calls `scrollToBottom(false)`.

`showScrollToBottomBtn === false` reflects a prior scroll-position calculation; it does not prove the newly appended row was actually seen. So an incoming message arriving while the tab is backgrounded/unfocused, or a 200ms smooth scroll the user interrupts by scrolling up, will now zero the unread state for content the user never viewed. This broadens read semantics well beyond the member-leave case and regresses the deliberate "visible, not merely loaded" invariant.

Suggested direction: gate the cursor advance on real visibility/focus (e.g. `document.visibilityState === "visible"`), or defer to the post-scroll `updateBrowseToMessageSeq` visibility check rather than firing synchronously right after `animateScroll.scrollToBottom()`.

## 🟡 Non-blocking — fire-and-forget clear has no ordering guard (`vm.ts:2306`)

`void this.refreshNewMsgCount()` starts an async chain that may `await markConversationUnread(channel, 0)` then emit `sidebar-reload`. If a newer message raises unread while that request is in flight, the older clear can resolve afterward and push a stale `0` / `sidebar-reload` over the newer state. The pattern pre-exists on the DOM path, but this new synchronous call site makes the race easier to hit. Consider a monotonic sequence check before applying the stale result.

## 🟡 Non-blocking — tests don't exercise the risky path (`Conversation/__tests__/messageOrder.test.ts:577-625`)

The two new tests seed VM fields directly and call `scrollToBottom(false)` with `react-scroll` mocked to no-op. They correctly prove the new branch and the `pullupHasMore` guard, and the mocked `removeMembers: 1003` constant matches `Service/Const.ts`. But they do not go through the real `appendMessage` → `showScrollToBottomBtn` → auto-scroll path (backgrounded tab, interrupted smooth scroll) — exactly the conditions in the blocking finding. Please add coverage for an other-user incoming message while at bottom so the read-semantics behavior is pinned.

## 🟡 Non-blocking — metadata/spec divergence

The title and Summary partly describe the reverted `ConversationList` cache-removal approach, while the net diff only touches `Conversation/vm.ts` and its test. The PR body does acknowledge the revert as intentional, so this is a metadata-hygiene item rather than a code blocker — but the title/`## Tests` references should be updated to match the unread-cursor fix so a merger can verify the diff against stated intent.

## ✅ Verified / non-issues

- The DOM `updateBrowseToMessageSeq` performs the identical two operations (set `browseToMessageSeq`, call `refreshNewMsgCount()`); the only real difference is the source of the sequence number (visible vs loaded) — that is the blocking finding, not a separate lifecycle bypass.
- The `O(n)` scan of `messagesOfOrigin` correctly ignores seq `0`/negative local rows and is fine at page size.
- The prior close-conversation change was intentionally removed, not accidentally dropped.

Verdict: **REQUEST_CHANGES** on the visibility-vs-loaded read-semantics regression. The remaining items are non-blocking. This concurs with the existing changes-requested review at this head.

PR #16, opened 2026-05-13
ci: add GitHub Actions CI workflow (build, test, lint)

yujiawei:
## Summary

Add a minimal CI workflow to `octo-web` so the branch ruleset can enforce CI status checks on `main`. Without this, the rulesets configured at the org level have no CI gate to require, leaving the same gap that produced earlier merge incidents in other repos.

## Workflow

- **`build`** — `pnpm install --frozen-lockfile` + `pnpm run build` (delegates to turbo, which builds every workspace).
- **`test`** — `pnpm run -r --if-present test`. The repo currently has tests in `apps/web`, `packages/dmworkdatasource`, and `packages/dmworklogin` (all vitest). `--if-present` keeps it green for workspaces that don't define `test`.
- **`lint`** — `pnpm run lint:architecture` (the `scripts/lint-architecture.mjs` guard) + `pnpm run lint:css:ci` (stylelint with the CI config).

Pinned `pnpm@10.32.0` to match `packageManager` in the root `package.json`, and `Node.js 20` for a current LTS that satisfies `engines.node >= 18`.

## Why

Branch protection for `main` already requires `code-review` (handled by the review dispatcher) but does not require any CI status check. The other repos in the org have analogous CI workflows (`octo-server` with Build/Test/Vet/Lint, `octo-cli` with build & test, `octo-admin` with build); octo-web should match.

After this lands I'll add `build`, `test`, `lint` to the required status checks for `refs/heads/main` so any future PR cannot merge with red CI.

## Test plan

- This PR itself exercises the workflow via `pull_request: branches: [main]`.
- All three jobs are timeboxed (10–20 min) to avoid runaway runs.
- `NODE_OPTIONS=--max-old-space-size=4096` is set on `build` to match local turbo expectations on a heavier app graph.

## Follow-up

- Add `build`, `test`, `lint` to the `main` ruleset's required status checks once this PR's first run is green.
- Add a typecheck job once the workspaces declare a `typecheck` script (none today).

PR #920, opened 2026-07-20
feat(html-doc): DocMoreMenu real avatar + consume __ODOC__ creator fields (OCT-184)

Ranwanglc:
Parent: OCT-179 · Spec: OCT-182 · Backend contract (OCT-183, parallel).

## What

Consume the new `__ODOC__.creator_uid` / `creator_name` / `created_at` fields to render a real creator avatar in the DocMoreMenu head. Previously we only had an initial-letter chip; now we try `/api/v1/users/<uid>/avatar` and fall back to the chip on missing uid / broken image.

## Design highlights

- **Auth boundary preserved.** The head avatar is display-only. Authorship still comes from `window.__ODOC_CAP__.isAuthor`. The stale comment in `HtmlDocView.tsx` warning that `__ODOC__` has no `creator_uid` is updated to say the field exists **for display only** and must not be used for cap checks (would let any viewer forge author capability).
- **Single avatar-URL builder.** New `packages/docs/src/html/avatarUrl.ts` exports `avatarUrlFromUid(uid)` with the same Space-prefix-stripping logic that `HtmlDocCommentPanel.avatarUrlFor` had inlined. Comment panel now delegates to it — the two surfaces cannot drift.
- **Failed-load isolation.** `DocMoreMenu` resets its `avatarFailed` state when the uid changes so switching docs re-tries the image.
- **CSS.** New `img.octo-doc-more-avatar` selector, `object-fit: cover`, keeps the circle for non-square sources; fallback tint stays visible for transparent images.

## Files

- `packages/docs/src/html/avatarUrl.ts` (new)
- `packages/docs/src/editor/DocMoreMenu.tsx`
- `packages/docs/src/editor/styles.css`
- `packages/docs/src/html/HtmlDocCommentPanel.tsx`
- `packages/docs/src/html/HtmlDocView.tsx`

## Tests

```
npx vitest run src/editor/DocMoreMenu.test.tsx src/html/HtmlDocView.test.tsx src/html/HtmlDocCommentPanel.test.tsx
# 3 files, 67 tests all pass
```

Full `pnpm test` shows 25 file failures — all pre-existing environment issues (`@univerjs/design` / `markdown-it` not installed), unrelated to this PR. The 1168 executable tests pass. `pnpm typecheck` produces no new errors on the touched files.

## Open item (from spec §3)

Behavior of `/api/v1/users/<uid>/avatar` for an unknown uid needs one manual check on a live env:

[...]

PR #944, opened 2026-07-21
feat(html-doc): source creator display and forward-grant from docs-backend (OCT-194)

Ranwanglc:
Closes upstream octo-web #914. Multica issue: OCT-194.

## Summary

Unify HtmlDocView's header creator display and forward-to-chat authorization flow with the rich-doc / sheet / board shells so both surfaces read from docs-backend as the single source of truth — dropping the last dependency on the deprecated `__ODOC__.creator_name / creator_avatar / created_at` fields.

## Changes

- **Creator display in the ≡ menu head is docs-backend sourced.** `HtmlDocView` now calls `getDoc(effectiveSlug, { spaceId: space })` for `ownerId` + `createdAt` + `role`, `getUserName(ownerId)` for the name, and a shared `avatarUrlForUid` helper for the head avatar. In-shell keeps the space-member map as the free primary source; standalone `/d/:docId` passes `creatorNicknameOnly` so the name resolves nickname-only and a link holder never sees the creator's verified `real_name` (XIN-392 P2-1 parity).
- **Forward-to-chat aligned with EditorShell/BoardShell/SheetView.** `doForward` now calls `startDocForward({ docId, title, role, currentUid, ownerId, space })` — `canGrant` = `computeCanGrant(role, currentUid, ownerId)`, `grantAccess` wired to the per-uid `POST /docs/{docId}/forward-grant` loop. Click is a no-op while `role` is still loading so a demoted admin never sends a stale `canGrant=false` snapshot (mirrors EditorShell's `if (!role) return`).
- **Shared avatar helper.** Extract `avatarUrlForUid` into `packages/docs/src/html/htmlAvatar.ts`. `HtmlDocCommentPanel` layers `author.avatar_url` in front of it (order preserved so backend-supplied avatar URLs still win).
- **DocMoreMenu.** New optional `creatorAvatarUrl` prop renders an `` with an `onError` fallback to the existing initial-letter chip; unset callers keep the prior chip-only behaviour.
- **`__ODOC__` interface fields** `creator_uid` / `creator_name` / `created_at` retained only as `@deprecated` so a payload that still carries them parses cleanly; no code reads them. `__ODOC_CAP__.isAuthor` (backend-authoritative author signal) is untouched.

## Fail-soft

`getDoc` 404 / 403 / network → `ownerId` / `createdAt` / `role` stay undefined → header falls back to slug initial, created-on row hides, `canGrant` stays greyed, no crash.

## Tests

- `HtmlDocView.test.tsx`: docs-backend-sourced header (creator name from `getUserName`, created-on from `getDoc.createdAt`); `X-Space-Id` on the standalone-surface preflight; `creatorNicknameOnly` privacy path; fail-soft on `getDoc` reject. Refresh the header forward test into three cases — reader → `canGrant=false`, owner/admin → `canGrant=true` + `grantAccess` executor wired, `role` not-loaded → click is a no-op.
- `DocMoreMenu.test.tsx`: `` renders when `creatorAvatarUrl` present; `onError` falls back to the initial-letter chip; unset prop keeps the original chip-only behaviour.
- `pnpm --filter @octo/docs test` (whole package): all HTML-doc + DocMoreMenu tests green. The 4 pre-existing failures in `EditorShell.test.tsx` (export-filename, header-injection, more-menu order) reproduce on unmodified `upstream/main` and are unrelated to this PR.

## Non-goals

- No new backend endpoints. Backend octo-docs-html PR #16 is closed; this is pure frontend.
- No changes to `__ODOC_CAP__` authorship signal or member-panel gating.

### DIFF
--- a/packages/docs/src/editor/DocMoreMenu.test.tsx
+++ b/packages/docs/src/editor/DocMoreMenu.test.tsx
@@ -89,3 +89,35 @@ describe('DocMoreMenu', () => {
     expect(screen.queryByText(/docs\.moreMenu\.createdPrefix/)).toBeNull()
   })
 })
+describe('DocMoreMenu — creator avatar (OCT-194)', () => {
+  it('renders an <img> when creatorAvatarUrl is present', () => {
+    render(
+      <DocMoreMenu creatorName="Alice" creatorAvatarUrl="/api/v1/users/u1/avatar" items={[]} />,
+    )
+    fireEvent.click(screen.getByRole('button', { name: 'docs.toolbar.more' }))
+    const img = document.querySelector('img.octo-doc-more-avatar') as HTMLImageElement | null
+    expect(img).toBeTruthy()
+    expect(img?.getAttribute('src')).toBe('/api/v1/users/u1/avatar')
+    // Fallback initial chip is NOT rendered when the image is showing (avoid double head).
+    expect(document.querySelector('span.octo-doc-more-avatar')).toBeNull()
+  })
+
+  it('falls back to the initial-letter chip when the avatar image fails to load', () => {
+    render(
+      <DocMoreMenu creatorName="Alice" creatorAvatarUrl="/api/v1/users/u1/avatar" items={[]} />,
+    )
+    fireEvent.click(screen.getByRole('button', { name: 'docs.toolbar.more' }))
+    const img = document.querySelector('img.octo-doc-more-avatar') as HTMLImageElement
+    fireEvent.error(img)
+    expect(document.querySelector('img.octo-doc-more-avatar')).toBeNull()
+    expect(document.querySelector('span.octo-doc-more-avatar')?.textContent).toBe('A')
+  })
+
+  it('renders the initial-letter chip when creatorAvatarUrl is absent (existing behaviour)', () => {
+    render(<DocMoreMenu creatorName="Bob" items={[]} />)
+    fireEvent.click(screen.getByRole('button', { name: 'docs.toolbar.more' }))
+    expect(document.querySelector('img.octo-doc-more-avatar')).toBeNull()
+    expect(document.querySelector('span.octo-doc-more-avatar')?.textContent).toBe('B')
+  })
+})
+
--- a/packages/docs/src/editor/DocMoreMenu.tsx
+++ b/packages/docs/src/editor/DocMoreMenu.tsx
@@ -23,6 +23,11 @@ export interface DocMoreMenuItem {
 export interface DocMoreMenuProps {
   /** Resolved creator display name. The parent handles the resolution + fallback (short uid). */
   creatorName: string
+  /**
+   * Optional avatar URL for the creator (parent-resolved). Rendered as an <img>; falls back to
+   * the initial-letter chip when absent or when the image fails to load (e.g. missing user).
+   */
+  creatorAvatarUrl?: string | null
   /** Creation timestamp (RFC3339). Rendered as "Created on YYYY-MM-DD"; omitted when absent/invalid. */
   createdAt?: string
   /** Neutral action rows, rendered top-to-bottom in the given order. */
@@ -175,6 +180,22 @@ function MenuRow({ item, onSelect }: { item: DocMoreMenuItem; onSelect: () => vo
   )
 }
 
+/** Avatar cell for the ≡ menu head: <img> when URL loads, initial-letter chip otherwise (also on onError). */
+function CreatorAvatar({ name, url }: { name: string; url?: string | null }) {
+  const [failed, setFailed] = useState(false)
+  const showImg = !!url && !failed
+  if (showImg) {
+    return (
+      <img className="octo-doc-more-avatar" src={url} alt="" title={name} onError={() => setFailed(true)} />
+    )
+  }
+  return (
+    <span className="octo-doc-more-avatar" aria-hidden="true">
+      {avatarInitial(name)}
+    </span>
+  )
+}
+
 /**
  * Header "more" (≡) dropdown for the document editor. Collapses the low-frequency title-bar
  * actions (open in new page / version history / export / delete) behind a single ≡ affordance at
@@ -183,7 +204,7 @@ function MenuRow({ item, onSelect }: { item: DocMoreMenuItem; onSelect: () => vo
  * Self-contained plain-React popover (no antd, no portal): a relatively-positioned wrapper anchors
  * an absolutely-positioned panel below the trigger. Closes on outside pointer-down and on Escape.
  */
-export function DocMoreMenu({ creatorName, createdAt, items, dangerItem }: DocMoreMenuProps) {
+export function DocMoreMenu({ creatorName, creatorAvatarUrl, createdAt, items, dangerItem }: DocMoreMenuProps) {
   const [open, setOpen] = useState(false)
   const wrapRef = useRef<HTMLDivElement>(null)
   const close = () => setOpen(false)
@@ -223,9 +244,7 @@ export function DocMoreMenu({ creatorName, createdAt, items, dangerItem }: DocMo
         <div className="octo-doc-more-panel" role="menu">
           <div className="octo-doc-more-head">
             <div className="octo-doc-more-creator">
-              <span className="octo-doc-more-avatar" aria-hidden="true">
-                {avatarInitial(creatorName)}
-              </span>
+              <CreatorAvatar name={creatorName} url={creatorAvatarUrl} />
               <span className="octo-doc-more-name" title={creatorName}>
                 {creatorName}
               </span>
--- a/packages/docs/src/html/HtmlDocCommentPanel.tsx
+++ b/packages/docs/src/html/HtmlDocCommentPanel.tsx
@@ -10,7 +10,8 @@
 // "让 AI 处理" click forwards an instruction to chat (openDocForward). The two are decoupled.
 
 import { useCallback, useEffect, useState } from 'react'
-import { canForwardToChat, openDocForward, t, getWKApp } from '../octoweb/index.ts'
+import { canForwardToChat, openDocForward, t } from '../octoweb/index.ts'
+import { avatarUrlForUid } from './htmlAvatar.ts'
 import {
   createComment,
   formatCommentTime,
@@ -61,24 +62,14 @@ function authorName(author: OctoDocAuthor | null | undefined): string {
 }
 
 /**
- * Resolve a comment author's avatar URL. Prefer the backend-supplied avatar_url when present;
- * otherwise derive it from the author uid (author.login) the same way collaborator avatars do:
- * the octo-server `/v1/users/<uid>/avatar` image endpoint, reached via the same-origin `/api/v1/`
- * proxy. This makes comment avatars work for ALL comments (incl. history) without the verify API
- * having to return an avatar. Returns null when no uid is available (→ initial-letter fallback).
+ * Resolve a comment author's avatar URL. Backend-supplied `avatar_url` wins; otherwise fall back
+ * to the shared `/api/v1/users/<uid>/avatar` endpoint via avatarUrlForUid (same helper the header
+ * ≡ menu uses for the doc creator). Order matters: never invert — an incoming avatar_url is the
+ * only path that carries a non-avatar-endpoint image the panel must respect.
  */
 function avatarUrlFor(author: OctoDocAuthor | null | undefined): string | null {
   if (author?.avatar_url) return author.avatar_url
-  let uid = author?.login?.trim()
-  if (!uid) return null
-  // Strip the Space-scoped prefix (s<spaceId>_) so we address the raw uid, mirroring
-  // WKApp.avatarUser()'s handling of person channel ids.
-  const spaceId = getWKApp().shared?.currentSpaceId
-  if (spaceId && uid.startsWith(`s${spaceId}_`)) {
-    uid = uid.substring(spaceId.length + 2)
-  }
-  if (!uid) return null
-  return `/api/v1/users/${encodeURIComponent(uid)}/avatar`
+  return avatarUrlForUid(author?.login)
 }
 
 /** Author + time line shown under each root comment and reply. */
--- a/packages/docs/src/html/HtmlDocView.test.tsx
+++ b/packages/docs/src/html/HtmlDocView.test.tsx
@@ -649,14 +649,56 @@ describe('HtmlDocView — header parity (presence / comments / members / more)',
     expect(container.querySelector('.octo-modal-overlay')).toBeNull()
   })
 
-  it('forwards the whole-doc link via openDocForward from the header forward button', async () => {
+  it('forwards the whole-doc link via startDocForward from the header forward button (non-admin viewer: canGrant=false)', async () => {
+    // docs-backend GET /docs/:id supplies the LIVE role + ownerId startDocForward consumes; a reader
+    // whose uid ≠ owner produces canGrant=false, matching the prior sharing-only behaviour.
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/the-slug') {
+        return { data: { docId: 'the-slug', ownerId: 'u_owner', role: 'reader' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+    serveDoc('<p>body</p>', { title: 'My Doc' })
+    const { container } = render(<HtmlDocView docId="d1" space="sp" slug="the-slug" version="v2" />)
+    await waitForFrame(container)
+    // Wait for the role to land so the click is not early-returned by the `!role` guard.
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/the-slug')).toBe(true))
+    fireEvent.click(screen.getByTitle('docs.forward.entry'))
+    await waitFor(() => expect(wk.openDocForwardCalls).toHaveLength(1))
+    expect(wk.openDocForwardCalls[0]).toMatchObject({ docId: 'd1', title: 'My Doc', canGrant: false })
+    expect(typeof wk.openDocForwardCalls[0].link).toBe('string')
+    // Sharing-only: no grantAccess executor when canGrant is false.
+    expect(wk.openDocForwardCalls[0].grantAccess).toBeUndefined()
+  })
+
+  it('opens forward with canGrant=true and wires a grantAccess executor when the viewer is owner/admin (computeCanGrant)', async () => {
+    // Owner (uid === ownerId) satisfies computeCanGrant regardless of role; grantAccess must be
+    // wired to the /docs/{docId}/forward-grant loop so the modal授权区 fires the per-uid grants.
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/the-slug') {
+        return { data: { docId: 'the-slug', ownerId: 'u_viewer', role: 'admin' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+    serveDoc('<p>body</p>', { title: 'My Doc' })
+    const { container } = render(<HtmlDocView docId="d1" space="sp" slug="the-slug" version="v2" />)
+    await waitForFrame(container)
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/the-slug')).toBe(true))
+    fireEvent.click(screen.getByTitle('docs.forward.entry'))
+    await waitFor(() => expect(wk.openDocForwardCalls).toHaveLength(1))
+    expect(wk.openDocForwardCalls[0]).toMatchObject({ docId: 'd1', canGrant: true })
+    expect(typeof wk.openDocForwardCalls[0].grantAccess).toBe('function')
+  })
+
+  it('forward click is a no-op while docs-backend role is still loading (fail-soft, no premature canGrant=false send)', async () => {
+    // A never-resolving getDoc keeps role=null; startDocForward must not fire (mirrors EditorShell
+    // `if (!role) return`) so a demoted admin never gets a stale canGrant snapshot.
+    wk.apiClient.responder = () => new Promise(() => {})
     serveDoc('<p>body</p>', { title: 'My Doc' })
     const { container } = render(<HtmlDocView docId="d1" space="sp" slug="the-slug" version="v2" />)
     await waitForFrame(container)
     fireEvent.click(screen.getByTitle('docs.forward.entry'))
-    expect(wk.openDocForwardCalls).toHaveLength(1)
-    expect(wk.openDocForwardCalls[0]).toMatchObject({ docId: 'd1', title: 'My Doc', canGrant: false })
-    expect(typeof wk.openDocForwardCalls[0].link).toBe('string')
+    expect(wk.openDocForwardCalls).toHaveLength(0)
   })
 
   it('offers delete only to the author in the ≡ menu', async () => {
@@ -685,3 +727,95 @@ describe('HtmlDocView — header parity (presence / comments / members / more)',
     expect(frame.getAttribute('srcdoc')).toContain('<base href="https://od.test/">')
   })
 })
+
+describe('HtmlDocView — creator/created head sourced from docs-backend (OCT-194)', () => {
+  let wk: ReturnType<typeof createMockWKApp>
+
+  function serveDoc(htmlBody: string) {
+    // Header-only tests: the __ODOC__ / __ODOC_CAP__ inline blobs are irrelevant here because the
+    // creator display now reads exclusively from docs-backend (getDoc + getUserName).
+    return stubFetch((url) => {
+      if (url.includes('/comments')) return jsonResponse({ data: [] })
+      return htmlResponse(htmlBody)
+    })
+  }
+
+  beforeEach(() => {
+    ;(window as unknown as { __OCTO_DOC_BASE__?: string }).__OCTO_DOC_BASE__ = 'https://od.test'
+    wk = createMockWKApp({ uid: 'u_viewer', token: 't' })
+    setWKApp(wk)
+  })
+  afterEach(() => {
+    delete (window as unknown as { __OCTO_DOC_BASE__?: unknown }).__OCTO_DOC_BASE__
+    setWKApp(undefined as never)
+  })
+
+  it('renders the creator name (from getUserName) and created-on date (from getDoc.createdAt) in the ≡ menu head', async () => {
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d1') {
+        return {
+          data: { docId: 'd1', ownerId: 'u_owner', createdAt: '2026-07-15T04:09:00Z' },
+          status: 200,
+        }
+      }
+      if (method === 'get' && url === '/users/u_owner') {
+        return { data: { name: 'Nick', real_name: 'Alice Owner' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+    serveDoc('<p>body</p>')
+    const { container } = render(<HtmlDocView docId="d1" space="sp" />)
+    await waitForFrame(container)
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/users/u_owner')).toBe(true))
+    // Open the ≡ menu.
+    fireEvent.click(container.querySelector('.octo-doc-more-btn') as HTMLElement)
+    // In-shell (creatorNicknameOnly unset) prefers real_name — the verified name lands in the head.
+    await waitFor(() => expect(screen.getByText('Alice Owner')).toBeTruthy())
+    // Created-on is a lexical YYYY-MM-DD slice (no tz drift).
+    expect(screen.getByText(/2026-07-15/)).toBeTruthy()
+  })
+
+  it('passes X-Space-Id on the docs-backend GET so the standalone /d/:docId space-required middleware accepts it', async () => {
+    wk.apiClient.responder = () => ({ data: {}, status: 200 })
+    serveDoc('<p>body</p>')
+    render(<HtmlDocView docId="d1" space="sp_42" />)
+    await waitFor(() => {
+      const call = wk.apiClient.calls.find((c) => c.url === '/docs/d1')
+      expect(call).toBeTruthy()
+      expect(call?.config?.headers).toMatchObject({ 'X-Space-Id': 'sp_42' })
+    })
+  })
+
+  it('standalone (creatorNicknameOnly) resolves nickname-only — real_name never surfaces to the link holder', async () => {
+    wk.apiClient.responder = (method, url) => {
+      if (method === 'get' && url === '/docs/d1') {
+        return { data: { docId: 'd1', ownerId: 'u_owner' }, status: 200 }
+      }
+      if (method === 'get' && url === '/users/u_owner') {
+        // Server returns both; the standalone surface must ignore real_name (preferRealName:false).
+        return { data: { name: 'Nick', real_name: 'Real Legal Name' }, status: 200 }
+      }
+      return { data: {}, status: 200 }
+    }
+    serveDoc('<p>body</p>')
+    const { container } = render(<HtmlDocView docId="d1" space="sp" creatorNicknameOnly />)
+    await waitForFrame(container)
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/users/u_owner')).toBe(true))
+    fireEvent.click(container.querySelector('.octo-doc-more-btn') as HTMLElement)
+    await waitFor(() => expect(screen.getByText('Nick')).toBeTruthy())
+    expect(screen.queryByText('Real Legal Name')).toBeNull()
+  })
+
+  it('fails soft when docs-backend rejects (404/network) — header falls back to slug initial, ≡ menu still opens, no crash', async () => {
+    wk.apiClient.responder = () => Promise.reject(new Error('not found'))
+    serveDoc('<p>body</p>')
+    const { container } = render(<HtmlDocView docId="d1" space="sp" slug="the-slug" />)
+    await waitForFrame(container)
+    // Header title falls through to the slug (no meta.title).
+    expect(container.querySelector('.octo-html-doc-title')?.textContent).toBe('the-slug')
+    // ≡ menu opens without crashing; the creator row shows the '—' placeholder (ownerId undefined).
+    fireEvent.click(container.querySelector('.octo-doc-more-btn') as HTMLElement)
+    expect(document.querySelector('.octo-doc-more-name')?.textContent).toBe('—')
+  })
+})
+
--- a/packages/docs/src/html/HtmlDocView.test.tsx
+++ b/packages/docs/src/html/HtmlDocView.test.tsx
@@ -653,8 +653,9 @@ describe('HtmlDocView — header parity (presence / comments / members / more)',
     // docs-backend GET /docs/:id supplies the LIVE role + ownerId startDocForward consumes; a reader
     // whose uid ≠ owner produces canGrant=false, matching the prior sharing-only behaviour.
     wk.apiClient.responder = (method, url) => {
-      if (method === 'get' && url === '/docs/the-slug') {
-        return { data: { docId: 'the-slug', ownerId: 'u_owner', role: 'reader' }, status: 200 }
+      // docs-backend is keyed by docId (d1), NEVER by the octo-doc slug (the-slug).
+      if (method === 'get' && url === '/docs/d1') {
+        return { data: { docId: 'd1', ownerId: 'u_owner', role: 'reader' }, status: 200 }
       }
       return { data: {}, status: 200 }
     }
@@ -662,7 +663,7 @@ describe('HtmlDocView — header parity (presence / comments / members / more)',
     const { container } = render(<HtmlDocView docId="d1" space="sp" slug="the-slug" version="v2" />)
     await waitForFrame(container)
     // Wait for the role to land so the click is not early-returned by the `!role` guard.
-    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/the-slug')).toBe(true))
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/d1')).toBe(true))
     fireEvent.click(screen.getByTitle('docs.forward.entry'))
     await waitFor(() => expect(wk.openDocForwardCalls).toHaveLength(1))
     expect(wk.openDocForwardCalls[0]).toMatchObject({ docId: 'd1', title: 'My Doc', canGrant: false })
@@ -675,15 +676,16 @@ describe('HtmlDocView — header parity (presence / comments / members / more)',
     // Owner (uid === ownerId) satisfies computeCanGrant regardless of role; grantAccess must be
     // wired to the /docs/{docId}/forward-grant loop so the modal授权区 fires the per-uid grants.
     wk.apiClient.responder = (method, url) => {
-      if (method === 'get' && url === '/docs/the-slug') {
-        return { data: { docId: 'the-slug', ownerId: 'u_viewer', role: 'admin' }, status: 200 }
+      // docs-backend is keyed by docId (d1), NEVER by the octo-doc slug (the-slug).
+      if (method === 'get' && url === '/docs/d1') {
+        return { data: { docId: 'd1', ownerId: 'u_viewer', role: 'admin' }, status: 200 }
       }
       return { data: {}, status: 200 }
     }
     serveDoc('<p>body</p>', { title: 'My Doc' })
     const { container } = render(<HtmlDocView docId="d1" space="sp" slug="the-slug" version="v2" />)
     await waitForFrame(container)
-    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/the-slug')).toBe(true))
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/d1')).toBe(true))
     fireEvent.click(screen.getByTitle('docs.forward.entry'))
     await waitFor(() => expect(wk.openDocForwardCalls).toHaveLength(1))
     expect(wk.openDocForwardCalls[0]).toMatchObject({ docId: 'd1', canGrant: true })
@@ -816,6 +818,47 @@ describe('HtmlDocView — creator/created head sourced from docs-backend (OCT-19
     // ≡ menu opens without crashing; the creator row shows the '—' placeholder (ownerId undefined).
     fireEvent.click(container.querySelector('.octo-doc-more-btn') as HTMLElement)
     expect(document.querySelector('.octo-doc-more-name')?.textContent).toBe('—')
+  })
+
+  it('OCT-198 regression: standalone slug≠docId hits docs-backend by docId (not slug); owner/created/role resolve', async () => {
+    // StandaloneDocPage passes docId=meta.docId + slug=meta.octoDocSlug as TWO different ids.
+    // The bug: getDoc was called with effectiveSlug (=slug), so `/docs/{slug}` 404'd and silently
+    // wiped ownerId/createdAt/role — creator display + forward授权 broke on every published html
+    // doc with a distinct octo-doc slug. Guard the fix: docs-backend receives docId, slug is
+    // reserved for octo-doc render (`/d/{slug}/v/{ver}`) + comment/asset/grant paths.
+    wk.apiClient.responder = (method, url) => {
+      // docs-backend keyed by docId. A slug-shaped call here is the pre-fix bug returning.
+      if (method === 'get' && url === '/docs/doc_abc') {
+        return {
+          data: { docId: 'doc_abc', ownerId: 'u_owner', role: 'admin', createdAt: '2026-07-20T00:00:00Z' },
+          status: 200,
+        }
+      }
+      if (method === 'get' && url === '/users/u_owner') {
+        return { data: { name: 'Nick', real_name: 'Alice Owner' }, status: 200 }
+      }
+      // Explicit trap: a call to `/docs/{slug}` means the bug regressed.
+      if (method === 'get' && url === '/docs/published-slug-xyz') {
+        throw new Error('OCT-198 regression: getDoc called with slug instead of docId')
+      }
+      return { data: {}, status: 200 }
+    }
+    serveDoc('<p>body</p>')
+    const { container } = render(
+      <HtmlDocView docId="doc_abc" space="sp_1" slug="published-slug-xyz" version="v2" />
+    )
+    await waitForFrame(container)
+    // 1) docs-backend addressed by docId, never by slug.
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/docs/doc_abc')).toBe(true))
+    expect(wk.apiClient.calls.some((c) => c.url === '/docs/published-slug-xyz')).toBe(false)
+    // 2) octo-doc render path still uses the slug (unchanged).
+    const octoCalls = (globalThis.fetch as unknown as { mock: { calls: [string, RequestInit?][] } }).mock.calls
+    expect(octoCalls.some(([u]) => String(u).includes('/d/published-slug-xyz/v/v2'))).toBe(true)
+    // 3) ownerId/createdAt/role landed — so the creator name resolves and forward授权 unblocks.
+    await waitFor(() => expect(wk.apiClient.calls.some((c) => c.url === '/users/u_owner')).toBe(true))
+    fireEvent.click(container.querySelector('.octo-doc-more-btn') as HTMLElement)
+    await waitFor(() => expect(screen.getByText('Alice Owner')).toBeTruthy())
+    expect(screen.getByText(/2026-07-20/)).toBeTruthy()
   })
 })
 
--- a/packages/docs/src/html/HtmlDocView.tsx
+++ b/packages/docs/src/html/HtmlDocView.tsx
@@ -17,7 +17,12 @@
 
 import { useCallback, useEffect, useRef, useState } from 'react'
 import DOMPurify from 'dompurify'
-import { canForwardToChat, openDocForward, t, getWKApp } from '../octoweb/index.ts'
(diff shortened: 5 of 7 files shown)


lml2468 (dismissed):

审查锚定 head `9f0d65b9`。7 文件 +326/−34,`docs`。把 HtmlDocView 的创建者展示 + 转发授权统一到 docs-backend 单一数据源(与 rich-doc/sheet/board shell 对齐),弃用 `__ODOC__` creator 字段。权限 + 隐私敏感,已重点核实。

### ✅ 两道闸门均通过

**规格符合**:与 #914/OCT-194 一致——创建者信息改走 `getDoc`(ownerId/createdAt/role)+ `getUserName`,转发走共享 `startDocForward`。无新后端端点(纯前端)。`__ODOC_CAP__.isAuthor` 不动。无 over-build。

**安全:转发授权(最关键)—— 正确**:
- `doForward`(:401)`if (!canForward || !role) return`——**role 未加载时点击 no-op**,不会用 stale `canGrant=false` 快照发送(mirrors EditorShell);
- `canGrant = computeCanGrant(role, currentUid, ownerId)`——复用既有共享函数(`forward/startDocForward.ts`,EditorShell/Board/Sheet 同款):`role && canManage(role)` 或 `currentUid === ownerId`,`role`/`ownerId` 均来自后端 `getDoc`,**非客户端可伪造**;真实授权由后端 `POST /docs/{docId}/forward-grant` per-uid 把关,客户端 canGrant 仅 UX。

**隐私:link-holder 不见 real_name(XIN-392 P2-1)—— 正确**:
- standalone `/d/:docId` 全部 4 处 mount 传 `creatorNicknameOnly`;
- 该 flag 下(:352-367)**跳过 space-member map** + `getUserName(ownerId, { preferRealName: false })` → 分享链接持有者只看昵称、绝不看创建者的 verified real_name。

**代码质量**:
- `__ODOC__` creator_uid/name/created_at 仅留 `@deprecated`,**全仓无代码读取**(grep 零命中)——后端 getDoc 成唯一源;
- `__ODOC_CAP__.isAuthor` 授权信号未削弱(仍 `parseOdocCap`,:292/320/466);
- 共享 `avatarUrlForUid`:同源 `/api/v1/users/<uid>/avatar` + `encodeURIComponent` + Space 前缀剥离 + null-on-empty(与 #920 同款,无 PII/注入);comment 面板 avatar 顺序保持(`author.avatar_url` 优先,:71-72);
- **fail-soft**:`getDoc` 404/403/网络 → role/ownerId/createdAt undefined → header 回落 slug 首字母、created-on 行隐藏、canGrant 灰、不崩(:343 catch)。

### ✅ 验证

- `HtmlDocView.test.tsx` + `DocMoreMenu.test.tsx` + `HtmlDocCommentPanel.test.tsx` **76 passed**,含:后端源 header、standalone preflight `X-Space-Id`、`creatorNicknameOnly` 隐私路径、getDoc reject fail-soft、转发三态(reader→canGrant=false / owner-admin→canGrant=true+executor / role 未加载→no-op)、DocMoreMenu img+onError 回落;
- computeCanGrant 后端权威、role no-op、nickname 隐私、deprecated 字段无读取、avatar 无 PII、fail-soft、isAuthor 未动均 byte 核实;
- 落后 main 1 commit,`git merge-tree` **0 冲突**。

### 结论

授权统一到后端权威 `computeCanGrant` + role-loading no-op(不发 stale 快照)、真实门在后端;隐私 nickname-only 路径杜绝 real_name 泄漏;deprecated `__ODOC__` 字段彻底停读、isAuthor 未削弱;共享 avatar 无 PII/注入;fail-soft 完备;76 测试全绿、0 冲突。无阻塞。**APPROVE**。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #944 (octo-web)

## Summary

This PR migrates the HTML doc viewer's creator display and forward-to-chat flow from inline `__ODOC__` metadata to the docs-backend `getDoc` API, achieving parity with EditorShell/BoardShell/SheetView. It adds: (1) creator name + avatar + created-date in the ≡ menu head sourced from `getDoc` + `getUserName`; (2) `computeCanGrant`-based forward with per-uid `grantAccess` via `startDocForward`; (3) a `creatorNicknameOnly` prop for the standalone `/d/:docId` surface that suppresses verified `real_name` from external link holders; (4) a shared `avatarUrlForUid` helper extracted from `HtmlDocCommentPanel`.

The implementation is clean, well-tested, and follows the established patterns from EditorShell. Two P2 items below — neither is a correctness blocker.

## Verification

- ✅ **Data flow: getDoc → ownerId/role/createdAt** — `getDoc` in `packages/docs/src/pages/docsApi.ts:297` returns `DocMeta` with optional `ownerId`, `role`, `createdAt`. `HtmlDocView.tsx:337-354` consumes all three with proper `typeof` guards and a cancelled-flag cleanup. Fail-soft on rejection (all fields stay undefined, header falls back to slug/dash).
- ✅ **Forward: startDocForward parity** — `startDocForward` in `packages/docs/src/forward/startDocForward.ts:44` calls `computeCanGrant(role, currentUid, ownerId)` which returns `true` for admin role or owner match. `grantAccess` executor wired to `grantForwardMany` only when `canGrant=true`. Tests cover non-admin viewer (`canGrant=false`, no executor), owner (`canGrant=true`, executor present), and role-still-loading (no-op).
- ✅ **Standalone privacy gate** — `StandaloneDocPage.tsx:607` passes `creatorNicknameOnly={true}` to `HtmlDocView`. The name resolution effect at `HtmlDocView.tsx:359-378` skips the member map and forces `preferRealName: false` on `getUserName`, so only `name` (nickname) is returned — `real_name` never surfaces to link holders. Test at `HtmlDocView.test.tsx:793` validates.
- ✅ **Avatar helper extraction** — `htmlAvatar.ts:19-27` extracts the space-prefix stripping + URL construction from `HtmlDocCommentPanel.tsx`. Behavior preserved: `avatarUrlFor` still checks `author.avatar_url` first, then falls back to `avatarUrlForUid(author.login)`. Null/empty input handling is robust (returns null → initial-letter chip).
- ✅ **Role type safety** — `setRole(m.role)` at `HtmlDocView.tsx:346` assigns to `Role | null` state. The `Role` type (`'reader' | 'writer' | 'admin'`) is shared between `auth/roles.ts` and `DocMeta`. Truthy guard `if (m?.role)` prevents undefined assignment.

Static analysis only at head `9f0d65b9`; build and tests not executed in this environment.

## Findings

### P2 — Forward button silently inert when getDoc fails (`packages/docs/src/html/HtmlDocView.tsx:377`)

When `getDoc()` rejects (network error, 404, 403), `role` stays `null` and `doForward` at `:377` returns without opening the forward modal. The forward button remains visible (gated only on `canForward` at `:575`), so clicking it produces no visible response. Previously, `openDocForward` was called directly with `canGrant: false` and no role dependency — the forward always worked when `canForwardToChat()` was true.

The test at `HtmlDocView.test.tsx:697` explicitly validates this fail-soft pattern, and it mirrors EditorShell's `if (!role) return` guard, so the design is intentional and consistent. Not a blocker, but worth considering a user-visible signal (tooltip, console.warn, or disabling the button when role is null) so the silent no-op is diagnosable on slow connections.

### P2 — Avatar space-prefix stripping scoped to viewer's space only (`packages/docs/src/html/htmlAvatar.ts:22`)

`avatarUrlForUid` strips `s${spaceId}_` using the viewer's `currentSpaceId`. If a uid is scoped to a different space (e.g. creator is `s42_alice` but viewer is in space `s99`), the prefix won't match and the URL becomes `/api/v1/users/s42_alice/avatar` — likely a 404 on the server. The consumer handles this gracefully via `onError` → initial-letter chip, so no broken UI.

In practice this is likely low-impact: `getDoc().ownerId` from the docs-backend typically returns a bare uid without a space prefix. The same pattern exists in WKApp.avatarUser. Not a blocker — flagging for awareness in case any backend path returns space-scoped uids.

- **OctoDocMeta deprecation**: `creator_uid`/`creator_name`/`created_at` fields marked `@deprecated` with JSDoc pointers to `getDoc`. No remaining production readers — clean migration.
- **X-Space-Id wiring**: `getDoc(effectiveSlug, { spaceId: space })` at `:337` passes an explicit space header, matching the standalone preflight pattern. Test at `HtmlDocView.test.tsx:773` validates.
- **`CreatorAvatar` error fallback**: `DocMoreMenu.tsx`'s `CreatorAvatar` component uses `useState(false)` for `onError` → initial-letter chip. Three test cases cover: image present, image error, no URL.
- **Comment panel avatar refactor**: `HtmlDocCommentPanel.tsx:68-70` now delegates to `avatarUrlForUid` — same space-prefix stripping, same URL, just extracted. `avatarUrlFor` still checks `author.avatar_url` first (order preserved).
- **useEffect cleanup**: Both `getDoc` and `getUserName` effects use `cancelled` flags to prevent stale state updates. Dependencies are correct (`effectiveSlug`, `space` for getDoc; `ownerId`, `names`, `creatorNicknameOnly` for name resolution).

## Verdict: COMMENT

Well-executed migration that achieves its stated parity goals. The two P2 items are non-blocking — the first is an intentional fail-soft pattern with test coverage, and the second is a low-probability edge case with graceful degradation. No correctness, security, or build-breaking issues.





---

[Octo-Q] verdict: APPROVE — No P1 blockers found. Two P2 non-blocking observations (silent forward no-op on getDoc failure; cross-space avatar prefix stripping). Clean migration with good test coverage.

yujiawei (changes requested):
# Code Review — PR #944 (octo-web)

**Head SHA reviewed:** `9f0d65b98a88353b22515c47c44018f93a1f3595`

Independently verified against `main` (`7be0d2dd`): repo checked out at the head SHA, the two PR test files run green (`HtmlDocView.test.tsx` + `DocMoreMenu.test.tsx`, 64/64), the docs package typechecks clean on the changed files, and the 4 `EditorShell.test.tsx` failures the PR calls out are confirmed pre-existing (they reproduce on unmodified `main`).

## 1. Spec compliance

**Spec: ✅**

The diff implements the stated scope and nothing beyond it:
- Creator name / created-on / avatar in the ≡ head are wired to `getDoc` + `getUserName` + a shared `avatarUrlForUid` helper (`HtmlDocView.tsx`, `htmlAvatar.ts`).
- Forward-to-chat is routed through `startDocForward` with `computeCanGrant` + the per-uid `forward-grant` executor (`HtmlDocView.tsx:395-407`), matching `EditorShell`/`SheetView`/`BoardShell`.
- `creatorNicknameOnly` privacy path is threaded from `StandaloneDocPage` (`StandaloneDocPage.tsx:606`).
- `__ODOC__` creator fields are demoted to `@deprecated` with no remaining readers; `__ODOC_CAP__.isAuthor` is untouched.
- No new backend endpoints, no over-building (no extra flags/fields/interfaces).

## 2. Code quality

### P1 — `getDoc()` is called with the octo-doc slug, but it is keyed by `docId` (wrong backend identifier)

`packages/docs/src/html/HtmlDocView.tsx:301,335`

```ts
const effectiveSlug = slug ?? docId
...
getDoc(effectiveSlug, opts)   // ← should be getDoc(docId, opts)
```

`getDoc()` targets the **docs-backend** `GET /api/v1/docs/{docId}` (`docsApi.ts:283`, via `apiClient()`), which is keyed by **docId**. The `slug` prop is the **octo-doc** slug — a different backend's identifier, used elsewhere in this component only for octo-doc calls (render/delete/comments/versions via `resolveOctoDocBase()`).

These two identifiers are genuinely distinct: `getDoc(docId)` itself returns `octoDocSlug` as a **separate field**, and the fixtures make the divergence explicit — e.g. `{ docId: 'd_html', octoDocSlug: 'sp/fd/html-slug' }` (`DocsHome.test.tsx:1692`), `octoDocSlug: 'octo-html-real'` (`docsApi.test.ts`), `octoDocSlug: 'badminton'` (`DocsHome.test.tsx:238`). Note the slug can even contain `/`, which would form a malformed `/api/v1/docs/sp/fd/html-slug` path.

Every sibling shell calls `getDoc(docId)` — never the slug (`EditorShell.tsx:142,308`, `SheetView.tsx:294`, `BoardShell.tsx:586,707`).

**Failure scenario:** In both real entry points the slug is passed and differs from docId — in-shell `DocsHome.tsx:1459` (`slug={octoDocSlug}`) and standalone `StandaloneDocPage.tsx:605` (`slug={meta.octoDocSlug}`). `getDoc('octo-html-real')` then hits a docId that does not exist → 404 → the `.catch` fail-soft leaves `ownerId`/`createdAt`/`role` undefined. The result is that **the entire feature this PR ships silently no-ops for real HTML docs**: the creator name falls back to the slug initial, the created-on row is hidden, and — because `role` never resolves — the forward button becomes a permanent no-op (see P2 below). The one case that works (`slug` omitted → `effectiveSlug === docId`) is the degraded `?doc=` re-render path, not the normal list-open path.

**Why the tests didn't catch it:** the new tests key the mock responder on the slug (`url === '/docs/the-slug'`, `HtmlDocView.test.tsx:656,681`) and assert `getDoc` was called with the slug, so they encode the buggy contract rather than the docs-backend one. A regression test that opens with `slug !== docId` and expects `getDoc(docId)` would fail today.

**Fix:** call `getDoc(docId, opts)`. Keep `effectiveSlug` for the octo-doc render/delete/comment paths only.

### P2 — Forward loses the prior sharing-only fallback when `role` never resolves

`packages/docs/src/html/HtmlDocView.tsx:400-401`

```ts
if (!canForward || !role) return
```

Previously this surface always allowed forwarding a read link (`openDocForward({ ..., canGrant: false })`). Now, for a bare doc with no doc_meta (404), a 403, or a network failure, `role` stays `null` forever and every forward click silently returns — the share-a-link capability is gone, not just the grant capability. The `if (!role) return` guard is correct for *loading*, but it does not distinguish "still loading" from "terminally failed". Consider retaining a sharing-only forward (`canGrant=false`) once the role fetch has settled to a failure, so a viewer can still forward the link. This P2 is currently *always* triggered as a consequence of the P1 addressing bug.

### P2 — Auth/creator state is not cleared when the effect re-runs, and the `.catch` is empty

`packages/docs/src/html/HtmlDocView.tsx:332-344`

The `getDoc` effect keys on `[effectiveSlug, space]` and updates `ownerId`/`createdAt`/`role` only on truthy fields, with an empty `.catch`. If this instance is ever reused for a different doc (or the same doc's fetch later fails/omits a field), the previous document's `ownerId`/`role` persist, and `doForward` could compute grant capability for the new docId from stale authorization state. In practice both current mount sites use `key={docId}` (`DocsHome.tsx:1459`, `StandaloneDocPage.tsx:603`), so the component remounts and this is not exploitable **today** — but the in-place-update design (deps on `effectiveSlug`/`space`) invites it. Recommend resetting the three state values at the top of the effect before the fetch, so correctness doesn't silently depend on an external `key` prop.

### P2 (nit) — `CreatorAvatar` does not reset `failed` when `url` changes

`packages/docs/src/editor/DocMoreMenu.tsx:181-193`

`failed` is set on `` but never reset when the `url` prop changes. If the same instance later receives a new valid avatar URL after an initial error, it stays on the initial-letter chip. Low impact (URL is stable for a mounted head), but a `useEffect(() => setFailed(false), [url])` would make it robust.

## 3. Overall verdict

**CHANGES_REQUESTED** — the P1 addressing bug means the feature does not actually take effect for real HTML docs (creator display and forward-grant both silently degrade), even though it fails soft without crashing. The one-line fix (`getDoc(docId)`) plus a test that exercises `slug !== docId` should resolve the blocker; the P2 items are worth addressing in the same pass.

## 4. Suggestions
- Change `getDoc(effectiveSlug, opts)` → `getDoc(docId, opts)` and add a test rendering with `slug !== docId` asserting the docs-backend call targets `docId`.
- After the role fetch settles to a failure, allow a sharing-only forward (`canGrant=false`) instead of a dead click.
- Reset `ownerId`/`createdAt`/`role` at the start of the `getDoc` effect so correctness does not rely on the caller's `key`.

## 5. Additional notes
- Fail-soft behaviour otherwise reads correctly (no crashes on reject; header falls back to slug initial).
- The `avatarUrlForUid` extraction preserves the `author.avatar_url`-first ordering in `HtmlDocCommentPanel` — good, that ordering is load-bearing.
- Coverage not exercised here: the live docs-backend contract for whether `GET /docs/{docId}` returns `role` for a read-only HTML doc with no collab session (this surface has no collab fallback, unlike the other shells) — worth a manual confirmation before merge, since the forward-grant capability now depends entirely on it.

Jerry-Xin (dismissed):
Requesting changes at head 9f0d65b9, concurring with the existing changes-requested review. I independently byte-verified the P1 addressing bug and it is a real functional blocker: the feature silently no-ops for real HTML docs.

## 🔴 Blocking — getDoc is called with the octo-doc slug, but it is keyed by docId

`packages/docs/src/html/HtmlDocView.tsx:301,335`

```ts
const effectiveSlug = slug ?? docId
...
getDoc(effectiveSlug, opts)
```

`getDoc` targets the docs-backend and is keyed by docId — verified in `packages/docs/src/pages/docsApi.ts:297,300`:

```ts
export async function getDoc(docId: string, opts?: { spaceId?: string }): Promise<DocMeta> {
  ...
  const { data } = await apiClient().get<DocMeta>(`/docs/${docId}`, config)
```

But `effectiveSlug` becomes the octo-doc slug whenever `slug` is provided, and the real entry point provides a slug that differs from docId. Verified in `packages/docs/src/pages/StandaloneDocPage.tsx:604-605`:

```tsx
<HtmlDocView
  docId={editorDocId}
  slug={meta.octoDocSlug}
```

`docId` (editorDocId = `meta.docId`) and `slug` (`meta.octoDocSlug`) are distinct identifiers. So the normal open path calls `getDoc(octoDocSlug)` → `GET /docs/{octoDocSlug}` → a docId that does not exist → 404 → the empty `.catch` leaves `ownerId`/`createdAt`/`role` undefined.

**Net effect:** the entire feature this PR ships silently degrades for real HTML docs — creator name falls back to the slug initial, the created-on row is hidden, and (because `role` never resolves) the forward button's guard `if (!canForward || !role) return` (line 401) makes forward a permanent no-op. It fails soft (no crash), which is why it isn't obvious, but the feature does not actually take effect on the normal list-open / standalone path. The only case that works is `slug` omitted (`effectiveSlug === docId`), which is the degraded `?doc=` re-render path.

**Why the tests pass anyway:** the new tests key the mock on the slug and assert `getDoc` was called with the slug, encoding the buggy contract rather than the docs-backend one. A test rendering with `slug !== docId` that expects `getDoc(docId)` would fail today.

**Fix:** `getDoc(docId, opts)`. Keep `effectiveSlug` for the octo-doc render/delete/comment/version paths only (those legitimately use the slug). Add a regression test with `slug !== docId` asserting the docs-backend call targets `docId`.

## 🟡 Non-blocking (worth handling in the same pass)
- Forward loses the prior sharing-only fallback when `role` never settles: previously this surface allowed forwarding a read link with `canGrant=false`. With `if (!role) return`, a terminally failed/absent role (404/403/network) turns every forward click into a dead no-op rather than degrading to a link-share. Consider distinguishing "still loading" from "settled to failure" and allowing `canGrant=false` forward once settled. (This is currently always triggered as a consequence of the P1 bug.)
- The `getDoc` effect (deps `[effectiveSlug, space]`) updates `ownerId`/`createdAt`/`role` only on truthy fields with an empty `.catch`, and does not reset them at the top of the effect. Not exploitable today because both mount sites use `key={docId}` (remount), but resetting the three values at the start of the effect would stop correctness from silently depending on an external `key` prop.
- `CreatorAvatar` (`DocMoreMenu.tsx`) sets `failed` on `` but never resets it when `url` changes; `useEffect(() => setFailed(false), [url])` would make it robust.

## Notes / verified-good
- Spec scope is otherwise clean: creator/created-on/avatar wiring, `startDocForward` + `computeCanGrant` + per-uid `forward-grant` executor mirror the sibling shells, `creatorNicknameOnly` privacy path is threaded, `__ODOC__` creator fields demoted to `@deprecated` with no remaining readers. No new backend endpoints, no over-building.
- No PII over-exposure: creator display uses `getUserName` + avatar helper; the `creatorNicknameOnly` path is honored.
- CI: Build is skipping and `code-review` is red on the existing changes-requested review; `osv-scan` fail is the repo-level dependency scan (unrelated). Before merge, also worth manually confirming the docs-backend `GET /docs/{docId}` returns `role` for a read-only HTML doc with no collab session, since forward-grant now depends entirely on it (this surface has no collab fallback).

Verdict: **REQUEST_CHANGES** on the docId-vs-slug addressing bug. The one-line fix plus a `slug !== docId` test resolves the blocker.

Jerry-Xin (dismissed):
Re-review at head 5db77cee. The blocking P1 is fixed with a targeted regression test — flipping to APPROVE.

## 🔴→✅ P1 resolved (byte-verified)
`packages/docs/src/html/HtmlDocView.tsx:340` now calls `getDoc(docId, opts)` instead of `getDoc(effectiveSlug, opts)`, with a comment documenting that the docs-backend `/docs/{docId}` is keyed by docId and MUST NOT be the slug, while the octo-doc render/comment/asset paths keep using `effectiveSlug`. This is exactly the fix requested.

The new tests pin it directly:
- **OCT-198 regression**: renders `docId="d1"` with `slug="the-slug"` (two distinct ids) and asserts the docs-backend GET targets docId — the mock 404s if keyed by the slug, so the previously-buggy contract would now fail the test.
- Added coverage for `canGrant=false`/`canGrant=true`, a never-resolving `getDoc` (forward must not fire), fail-soft on 404/network (header falls back to the slug initial, ≡ menu still opens, no crash), `X-Space-Id` propagation on the standalone `/d/:docId` path, and `creatorNicknameOnly` (link holder never sees the verified real_name).

## P2 items — acceptable as-is
- Forward's `if (!canForward || !role) return` (line 406) is retained, but this is the intended parity with EditorShell/BoardShell/SheetView (early-return only while role is still loading). With P1 fixed, `role` now resolves on the normal path; the earlier "permanent no-op" was strictly a consequence of the P1 404, which is gone. Not blocking.
- The `getDoc` effect uses a `cancelled` guard with `[docId, space]` deps and only sets state on truthy fields; both mount sites use `key={docId}` so the component remounts per doc. Not exploitable; acceptable design.

## PII / scope
- Creator display reads from docs-backend `getDoc` + `getUserName`; `creatorNicknameOnly` is honored on the standalone surface so a link holder never sees the creator's verified real_name — consistent with the visibility rules. No over-exposure.
- Delta is scoped to exactly 2 files (HtmlDocView.tsx + its test), 0 commits behind main.

## CI
- Build skipping / code-review pending (queued); dependency-review pass. `mergeable: true` (the `blocked` state is the pending review gate, not conflicts).

Verdict: **APPROVE**. The docId-vs-slug addressing bug is fixed and covered by a `slug !== docId` regression test; the P2 items are acceptable. Thanks for the quick turnaround.

Ranwanglc:
@Jerry-Xin @yujiawei Thanks for catching the P1 — the slug/docId mixup is fixed.

Fix at head `5db77cee`:
- `HtmlDocView` now calls `getDoc(docId, opts)` (was `getDoc(effectiveSlug)`). docs-backend `/docs/{docId}` is docId-keyed; octo-doc render/comment/asset/grant paths keep using slug. Comment added to lock the boundary.
- Regression test added (`HtmlDocView.test.tsx`, "OCT-198 regression: standalone slug≠docId ..."): a call to `/docs/{slug}` throws an explicit trap error, so the bug can't silently regress.
- HtmlDocView.test.tsx 55/55 pass locally.

Standalone slug≠docId path now resolves ownerId/createdAt/role correctly. Could you re-review and dismiss the changes-requested if it looks good? Thanks.

lml2468 (approved):
## Review Verdict（复审 @ 5db77cee）：可合并（APPROVE）—— 从我之前的 APPROVE 修正后维持 APPROVE

复审锚定新 head `5db77cee`。**yujiawei/Steve/Jerry-Xin 提的 P1 真阻塞(getDoc 传 slug 而非 docId → 404 → 创建者展示 + forward-grant 静默失效)已修**。坦白:我此前那轮 APPROVE 漏了这条,是我的 miss——它 fail-soft 不崩、被测试用错的契约固化,所以不显眼,但功能实际没生效。感谢三位抓出。

### ✅ getDoc(docId) 修复(byte 核实)

- `HtmlDocView.tsx:340` 现为 `getDoc(docId, opts)`——不再是 `effectiveSlug`;注释明确「docs-backend `/docs/{docId}` 按 docId 取,MUST NOT be slug」;octo-doc 的 render/comment/asset 路径仍用 `effectiveSlug`(那些确实按 slug),分离正确;
- **standalone 是 bug 触发场景**:`StandaloneDocPage:604-605` `docId={editorDocId}` 与 `slug={meta.octoDocSlug}` 是**两个不同值**——修复后 `getDoc(docId)` 打对后端,role/ownerId/createdAt 正常解析;
- **连锁修复**:role 现在能解析 → `doForward`(:406)`if(!canForward||!role)return` 不再永久 no-op、`computeCanGrant(role,...)` 正常算 → 转发授权按钮真正生效(之前整体静默降级)。

### ✅ 测试契约纠正(关键)

- 之前测试把 mock 按 `/docs/the-slug` 匹配、断言用 slug 调——**把错的契约固化了**(这也是我上轮没识破的原因);
- 本次改成 mock `/docs/d1`(docId)、断言调 `/docs/d1`,并**新增一条「捕获未来 slug 形状调用」的回归测试**防复发。契约现在对了。

### ✅ 验证

- `HtmlDocView` + `DocMoreMenu` + `HtmlDocCommentPanel` 测试 **77 passed**(上次 76,+1 回归);
- getDoc(docId)、standalone docId≠slug、role→canGrant 连锁、测试契约纠正 + 防复发回归均 byte 核实;
- forward gate(computeCanGrant 后端权威 + role no-op)、nickname-only 隐私、fail-soft、isAuthor 未削弱(上轮已核)均保持;
- 落后 main 5 commit,`git merge-tree` **0 冲突**。

### 🟡 非阻塞(Steve 提,可 fast-follow)

- role 失败态未与「加载中」区分(丢 share-link 兜底)、getDoc effect 靠外部 key 重置而非显式复位、CreatorAvatar failed 不随 url 重置——都是非阻塞健壮性项,不拦本次。

### 结论

我上轮漏掉的 P1(getDoc 传 slug 致创建者展示 + forward-grant 静默失效)已修净:`getDoc(docId)` 打对后端、role 连锁恢复转发授权、测试契约纠正并加防复发回归。此前已核的授权后端权威/隐私 nickname-only/fail-soft 保持。测试全绿、0 冲突。**APPROVE**。这条 miss 我认,感谢 peer 把关。

Ranwanglc:
@yujiawei Following up on your CHANGES_REQUESTED review at head `9f0d65b9` — please re-review at head **`5db77cee`**.

## P1 addressed
`packages/docs/src/html/HtmlDocView.tsx:340` now calls `getDoc(docId, opts)` (was `getDoc(effectiveSlug, opts)`), and the effect deps are `[docId, space]` (was `[effectiveSlug, space]`). A 4-line comment pins the boundary: docs-backend `/docs/{docId}` is docId-keyed; octo-doc render/comment/asset/grant paths keep using `effectiveSlug`.

## Full-PR sweep for id/slug misuse
Only one call site was wrong (the one you flagged); everything else in this PR is already correct:
- `getDoc` (docs-backend `/docs/{docId}`) — **fixed** ✓
- `getUserName` (docs-backend `/users/{uid}`) — takes ownerId, not id/slug ✓
- `startDocForward` → `POST /docs/{docId}/forward-grant` — already uses the `docId` prop ✓
- `buildDocLink({ docId })` — SPA route ✓
- Remaining `effectiveSlug` consumers are octo-doc only (`buildOctoDocUrl` / `deleteDoc` / `HtmlMemberPanel` / `HtmlDocCommentPanel`) — semantically correct, kept as-is ✓

## Regression guard
New test in `packages/docs/src/html/HtmlDocView.test.tsx`, `OCT-198 regression: standalone slug≠docId hits docs-backend by docId…`:
- Renders with `docId="doc_abc" slug="published-slug-xyz"` (two distinct ids).
- Asserts docs-backend receives `/docs/doc_abc` and **never** `/docs/published-slug-xyz` (a slug-shaped call in the mock throws — exactly the trap you suggested).
- Asserts the octo-doc render path still uses the slug (`/d/published-slug-xyz/v/v2`).
- Asserts `ownerId`/`createdAt`/`role` all land — creator name "Alice Owner" and createdAt "2026-07-20" render in the ≡ menu head.

Also cleaned up the two earlier forward tests that were mocking `/docs/the-slug` — they had baked the bug into the contract; both now mock `/docs/d1` (the docId).

## P2 items from your review
- **Forward silently inert when `role` never resolves** — with P1 fixed, `role` now resolves on the normal path, so the "permanent no-op" you flagged as a consequence of P1 is gone. The `if (!canForward || !role) return` guard is retained deliberately as parity with `EditorShell`/`BoardShell`/`SheetView` (correct behaviour while role is still loading). Jerry-Xin re-reviewed at `5db77cee` and accepted this as-is.
- **Effect doesn'"'"'t clear stale `ownerId`/`role` on re-run** — not exploitable today because both mount sites use `key={docId}`, but noted; kept out of this PR to keep the delta minimal for a P1 hotfix.
- **`CreatorAvatar.failed` not reset on `url` change** — same reasoning; noted for a follow-up.

## Verification (local, at `5db77cee`)
- `pnpm --filter @octo/docs test src/html/HtmlDocView.test.tsx` → **55 passed** ✓
- `pnpm --filter @octo/docs test src/editor/DocMoreMenu.test.tsx` → **10 passed** ✓
- `pnpm --filter @octo/docs typecheck` → only 2 pre-existing errors unrelated to this PR ✓
- `apps/web` `vite build` → **built in 4.52s, exit 0** ✓

Jerry-Xin already re-reviewed at `5db77cee` and flipped to APPROVE. Your CHANGES_REQUESTED is the last gate — please take a fresh look when you get a moment. Thanks for the catch.

yujiawei (approved):

yujiawei (approved):
# Code Review — PR #944 (octo-web)

**Head SHA reviewed:** `5db77cee8f042eb31d2bb4fb981931156ddc40c4` (re-review)

This is a re-review. My earlier `CHANGES_REQUESTED` (and the concurring one) were anchored at head `9f0d65b9` and raised a single blocking P1: `getDoc()` was called with the octo-doc **slug** (`effectiveSlug`) instead of the docs-backend **docId**, so the feature silently no-op'd for every real HTML doc (creator display, created-on row, and forward-grant all degraded because `role` never resolved). The new head is precisely that fix.

## Delta since the prior review

The only change from `9f0d65b9 → 5db77cee` is 2 files (+56/−8):

- `HtmlDocView.tsx`: `getDoc(effectiveSlug, opts)` → **`getDoc(docId, opts)`**, and the effect deps updated `[effectiveSlug, space]` → **`[docId, space]`**. `effectiveSlug` is retained for the octo-doc render/delete/comment/asset paths, which legitimately use the slug.
- `HtmlDocView.test.tsx`: existing forward tests re-keyed to `/docs/d1` (docId) instead of `/docs/the-slug`, plus a new **OCT-198 regression test** that mounts with `slug !== docId`, asserts the docs-backend call targets `/docs/doc_abc` (docId), asserts `/docs/published-slug-xyz` (slug) is **never** called (throws if it is), and confirms owner/created/role resolve.

## Verification (independent, at current head)

- **The addressing fix is correct.** `getDoc()` targets docs-backend `GET /api/v1/docs/{docId}` (`docsApi.ts:297`), which is keyed by docId. Every sibling shell calls `getDoc(docId)`. `HtmlDocView.tsx:340` now matches. The two mount sites pass `docId=meta.docId` + `slug=meta.octoDocSlug` as distinct ids (`StandaloneDocPage.tsx:604-605`, `key={editorDocId}`), so the normal open path now correctly hits the doc's real docId.
- **Regression test genuinely guards the bug.** It traps a slug-keyed call (`throw` on `/docs/published-slug-xyz`) and asserts the octo-doc render path still uses the slug (`/d/published-slug-xyz/v/v2`). A reintroduction of the bug fails this test.
- **Tests green.** The 3 PR-touched test files run clean: `HtmlDocView.test.tsx` + `DocMoreMenu.test.tsx` + `HtmlDocCommentPanel.test.tsx` = **77/77 passed**.
- **Typecheck clean** on all changed files (`tsc --noEmit` reports zero errors in the touched files).
- **Pre-existing failures unrelated.** The 4 failures seen in a full-package parallel run are in `Toolbar.test.tsx` (tiptap editor-view / jsdom animation-frame races) — a file this PR does not touch; it passes in isolation on the merge-base.

## Spec compliance

Scope is clean and unchanged from the prior head: creator name / created-on / avatar wired to `getDoc` + `getUserName` + shared `avatarUrlForUid`; forward-to-chat routed through `startDocForward` + `computeCanGrant` + per-uid `forward-grant` executor (parity with EditorShell/SheetView/BoardShell); `creatorNicknameOnly` privacy path threaded so link holders never see the creator's verified `real_name`; `__ODOC__` creator fields demoted to `@deprecated` with no remaining readers; `__ODOC_CAP__.isAuthor` untouched. No new backend endpoints, no over-building.

## Non-blocking notes (P2 / nits — do not block merge)

These are unchanged from the prior review and are all graceful-degradation edge cases; none needs to be fixed before merge:

1. **Forward becomes a dead no-op on a *genuine* `getDoc` failure.** `doForward` guards on `if (!canForward || !role) return`. The guard is correct for the *loading* state, but it doesn't distinguish "still loading" from "settled to a 404/403/network failure" — in the failure case the surface loses the prior sharing-only forward (`canGrant=false`). Now that the P1 addressing bug is fixed this is a genuinely rare path, but consider allowing a sharing-only forward once the role fetch settles to failure.
2. **`getDoc` effect updates state only on truthy fields and never resets it.** Safe today because both mount sites use `key={docId}` (remount on doc change), but resetting `ownerId`/`createdAt`/`role` at the top of the effect would stop correctness from silently depending on an external `key` prop.
3. **`CreatorAvatar` (`DocMoreMenu.tsx:184`) doesn't reset `failed` when `url` changes.** `useEffect(() => setFailed(false), [url])` would make it robust; low impact since the URL is stable for a mounted head.
4. **`avatarUrlForUid` (`htmlAvatar.ts:22`) strips the space prefix using the viewer's current space.** A cross-space uid won't match and yields a 404, gracefully handled by the `onError` → initial-letter chip. Low-probability; flagged for awareness.

## Coverage note

Not exercisable from the frontend diff: the live docs-backend contract for whether `GET /docs/{docId}` returns `role` for a read-only HTML doc with no collab session. The forward-grant capability on this surface depends entirely on it (unlike the other shells, this surface has no collab fallback). Worth a one-time manual confirmation against the deployed docs-backend before/around merge — not a code-level blocker in this PR.

## Verdict

**APPROVED.** The sole blocking P1 from the prior head is resolved with the correct minimal fix (`getDoc(docId)`) plus a real regression test; spec scope is clean, the PR test files are green, and touched files typecheck. Remaining items are P2/nits and non-blocking.

yujiawei (approved):
# Code Review — PR #944 (octo-web)

Reviewed at head `5db77cee8f042eb31d2bb4fb981931156ddc40c4`. This is a re-review at the current head; the verdict below is issued fresh against this SHA, not inherited.

## 1. Spec compliance

The change does exactly what the description states and nothing more.

- **Not under-built**: Creator name/avatar/created-on and forward-grant capability now all read from docs-backend (`getDoc` + `getUserName`), removing the last dependency on the deprecated `__ODOC__.creator_*` fields. Standalone `/d/:docId` passes `creatorNicknameOnly` so a link holder never sees `real_name`. Fail-soft on `getDoc` 404/403/network is implemented (owner/created/role stay undefined → slug-initial header, greyed grant, no crash).
- **Not over-built**: No new backend endpoints, no change to `__ODOC_CAP__.isAuthor` authorship signal, no member-panel gating changes. The `__ODOC__` `creator_*` fields are retained only as `@deprecated` type entries so a legacy payload still parses; no code reads them.
- **No deviation**: The forward flow is unified with the other shells via `startDocForward`, matching the stated goal.

**Spec: ✅**

## 2. Code quality

**Data source keying (the load-bearing correctness point).** `HtmlDocView.tsx:340` calls `getDoc(docId, opts)` — keyed by `docId`, never the octo-doc slug. This is the correct target for docs-backend `GET /api/v1/docs/{docId}` (`docsApi.ts:300`). It matches the sibling convention exactly (`BoardShell.tsx:990` uses the same `getDoc(docId, space ? { spaceId: space } : undefined)` form). Standalone mounts pass `docId=meta.docId` and `slug=meta.octoDocSlug` as two distinct ids; using the slug here would 404 and silently zero `ownerId/createdAt/role`, dead-clicking forward. The `OCT-198` regression test guards this precisely: it asserts `/docs/doc_abc` is hit, `/docs/{slug}` is never hit, and the octo-doc render path still uses the slug.

**Forward-grant.** `doForward` early-returns while `role` is still loading (`if (!canForward || !role) return`), so a demoted admin never sends a stale `canGrant=false` snapshot — mirrors EditorShell. `canGrant` is computed by the shared `computeCanGrant(role, currentUid, ownerId)`; the per-uid grant executor is only wired when `canGrant` is true.

**Avatar helper.** `avatarUrlForUid` is a clean extraction; `HtmlDocCommentPanel.avatarUrlFor` preserves the original precedence (backend-supplied `author.avatar_url` still wins over the derived endpoint). The `DocMoreMenu` `CreatorAvatar` renders an `` with an `onError` fallback to the initial-letter chip; the unset-prop path keeps prior behavior.

**X-Space-Id.** `getDoc` attaches `X-Space-Id` only when `opts.spaceId` is non-empty; in-shell callers rely on the request interceptor injecting it from `currentSpaceId`. This is the same known-benign pattern used elsewhere in this package — not a functional break and not a regression introduced here.

**Tests.** 77/77 green across `HtmlDocView.test.tsx`, `DocMoreMenu.test.tsx`, `HtmlDocCommentPanel.test.tsx`, including reader→`canGrant=false`, owner/admin→`canGrant=true`+executor wired, role-not-loaded→no-op, nickname-only privacy path, fail-soft on reject, and the OCT-198 slug≠docId trap.

No P0/P1 findings.

**Quality: Approved**

## 3. Overall verdict

**APPROVE** (Spec ✅ AND Quality Approved).

## 4. Suggestions

None blocking. Optional nit: the deprecated `__ODOC__.creator_*` interface fields could be dropped entirely in a follow-up once no environment emits them, to prevent future accidental readers.

Jerry-Xin (approved):
**✅ APPROVE** — byte-verified against live head `5db77cee`. Creator display + forward-grant are cleanly re-based onto docs-backend as the single source of truth, and — critically — the forward-grant authority contract is aligned client↔backend (no client-shows-then-backend-403 mismatch).

## Authority mapping (the primary risk here) — verified consistent
The sibling PR's recurring blocker was a client gate wider than the backend endpoint's enforcement (admin shown, but the action hit an author-only endpoint → 403). This PR does **not** repeat it:

- **Client gate:** `doForward` → `startDocForward` → `computeCanGrant(role, currentUid, ownerId)` = `canManage(role)` (admin) **OR** `currentUid === ownerId` (owner). `packages/docs/src/forward/startDocForward.ts`.
- **Endpoint:** grant executor is wired only when `canGrant` is true, and targets the per-uid `POST /docs/{docId}/forward-grant` loop (`packages/docs/src/forward/api.ts`). This is a **distinct** endpoint from member-management PUT — it runs a GREATEST-only upsert (升不降, idempotent) and its documented contract enforces **admin/owner** server-side (`403 → forbidden` mapped as "forwarder is not admin/owner — backend double-checks"). Client gate == backend gate → no 403 surprise for the roles the UI enables.
- **Stale-snapshot guard:** `doForward` early-returns while `role` is still loading (`if (!canForward || !role) return`), mirroring the editor shell's `if (!role) return`, so a demoted admin never sends a stale `canGrant=false`/`true` snapshot. The "role still loading → click is a no-op" test covers this.
- Grantable roles restricted to reader/writer (no commenter/admin escalation).

## Identifiers — correct
`getDoc` is keyed by **docId**, never the octo-doc slug — the OCT-198 regression (slug≠docId → `/docs/{slug}` 404 silently zeroing owner/created/role) is explicitly guarded by a test that traps a slug-shaped call. octo-doc render/comment/asset paths keep using the slug. `docId`/`space`/`ownerId`/`currentUid` are threaded correctly into both the display and the grant call.

## Creator display — backend-authoritative, safe fallback, no XSS
- Resolved from docs-backend `getDoc().ownerId` + `getUserName(ownerId)`, not a client uid-compare. Fallback order: resolved name → short uid slice → `—` placeholder (never blank, never crashes).
- In-shell prefers the free space-member map then `GET /users/:uid`; standalone `/d/:docId` passes `creatorNicknameOnly` → `preferRealName:false`, so a link holder never sees the creator's verified real name (privacy path tested).
- Name is rendered as React text (`{creatorName}` / `title=`), not `dangerouslySetInnerHTML` → no XSS. Avatar is an `` with an `onError` fallback to the initial-letter chip; `avatarUrlForUid` strips the space-scoped prefix and returns null on empty uid so no broken `` fires.

## Token / scoping — first-party
All calls go through the shared host `apiClient` (`/api/v1/` baseURL, token + `X-Space-Id` via interceptor). Standalone mounts pass `X-Space-Id` explicitly on the docs-backend GET (space-required middleware); verified by test. No token in URL.

## Test fidelity
The header forward test is split into the three real cases with production-shaped props: reader → `canGrant=false` + no executor; owner/admin → `canGrant=true` + a wired `grantAccess` executor; role-not-loaded → click is a no-op. Backend responder is keyed by docId (matching the real call site), not a false-green slug config. Fail-soft (getDoc reject → slug initial, `—`, no crash) is covered.

## Build / tests
- The 7 changed files typecheck clean. The 2 `tsc` errors seen locally (`dmworkbase i18n/format.ts` fractionalSecondDigits, `members/sort.ts` Member.grantedBy) reproduce **identically on the unmodified merge-base** and are in files this PR does not touch → pre-existing env/main noise, not this PR.
- The two PR test files pass: **HtmlDocView.test.tsx (55) + DocMoreMenu.test.tsx (10) = 65/65 green**.
- Full `@octo/docs` suite: head = 136 failed / 1457 passed vs merge-base = 136 failed / 1447 passed — the 136 failures (whole-suite setup crashes under the local Node runtime, e.g. StandaloneDocPage all-54-fail) are **identical on main**. This PR adds 10 net-new passing tests and introduces **zero** new failures.

## Non-blocking notes
- The `forward-grant` path/param naming is annotated "pending backend confirmation" in `forward/api.ts`. The frontend contract (single uid, `{uid, role}`, per-uid 200/404/403) is stable and isolated to that one file, so a backend path change is a one-file update — fine as-is, just track it before backend cutover.
- Deprecated `__ODOC__` creator fields are retained as `@deprecated` interface-only with no readers — good; confirm no downstream still parses them once the backend drops them.

mergeable: MERGEABLE (mergeStateStatus BLOCKED is the pending code-review gate + an in-progress CI Build, not a code defect). No regression, no dead code, authority contract sound. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #944 (octo-web)

## Summary

This PR migrates the read-only HTML doc viewer (`HtmlDocView`) from the inlined `__ODOC__` blob to the docs-backend (`GET /docs/{docId}`) as the source of truth for creator display, creation date, and viewer role — achieving parity with EditorShell/BoardShell/SheetView. It also unifies the forward-to-chat flow via `startDocForward` (which computes `canGrant` from the live role and wires the per-uid grant executor), adds a creator avatar in the ≡ menu, and introduces a `creatorNicknameOnly` privacy gate so the standalone `/d/:docId` surface never exposes the creator's verified real_name to link holders.

Overall the data flow is sound: every consumed value (ownerId, createdAt, role, creatorName, canGrant) traces back to a well-defined upstream source, and fail-soft paths keep the UI usable when the docs-backend is unreachable. Two minor P2 items below.

## Verification

Static analysis only at head `5db77cee8f042eb31d2bb4fb981931156ddc40c4`; build and tests not executed in this environment.

- ✅ **Data flow — getDoc → state** — `getDoc(docId, opts)` resolves `ownerId`, `createdAt`, `role` from docs-backend; each is guarded by type/truthiness checks before `setState`. The `.catch` swallows failures so all three stay `undefined`/`null`, and downstream consumers fall back to slug/initial/disabled-forward.
- ✅ **Data flow — creatorName resolution** — `useMemberNames(space)` returns an empty `Map` for empty spaceId (`getSpaceMemberNames` early-returns). The creator-name effect checks `fromMembers` first (in-shell), falls through to `getUserName(ownerId, {preferRealName})` when the map is empty or `creatorNicknameOnly` is set. The privacy gate at `packages/docs/src/html/HtmlDocView.tsx:369` correctly forces `preferRealName: false` for standalone.
- ✅ **Forward-grant computation** — `startDocForward` delegates to `computeCanGrant(role, currentUid, ownerId)` which returns `true` only for admin or owner. The `!role` early-return in `doForward` (`packages/docs/src/html/HtmlDocView.tsx:403`) prevents a premature `canGrant=false` while the role is still loading — mirrors EditorShell's guard.
- ✅ **Privacy parity (C1)** — `creatorNicknameOnly` is passed only by `StandaloneDocPage.tsx:607`; the in-shell consumer at `DocsHome.tsx:1459` omits it, so verified names show in-shell but not on external links. Confirmed no other `HtmlDocView` call-site exists.
- ✅ **Avatar refactor parity** — `avatarUrlForUid` in `packages/docs/src/html/htmlAvatar.ts` exactly replicates the prior inline logic in `HtmlDocCommentPanel`, including space-prefix stripping math (`substring(spaceId.length + 2)`). `HtmlDocCommentPanel.avatarUrlFor` now delegates to the shared helper; the `author.avatar_url` preference order is preserved.
- ✅ **Test coverage** — New tests cover: forward with canGrant=false (reader), canGrant=true (owner/admin, with grantAccess executor), forward no-op while role loads, creator name + date from docs-backend, `X-Space-Id` header injection, standalone nickname-only resolution, avatar image load + onError fallback, and initial-letter chip. Mocks use `wk.apiClient.responder` keyed by `/docs/{docId}` (not slug), matching production.

## Findings

### P2 — `getDoc` does not URL-encode `docId` (`packages/docs/src/html/HtmlDocView.tsx:341`)

The new `getDoc(docId, opts)` call passes `docId` raw into the URL template at `packages/docs/src/pages/docsApi.ts:223` — `` `/docs/${docId}` ``. The sibling helper `recordDocView` in the same file (`:192`) encodes: `` `/docs/${encodeURIComponent(docId)}/view` ``. If a docId ever carries URL-special characters (`#`, `?`, `/`), the GET would malform (fragment truncation or query injection). Practical risk is low — docIds are backend-generated UUIDs — but encoding at the helper would close the inconsistency and align with `recordDocView`.

**Diff-scope:** new call-site introduced by this PR (the `getDoc` helper itself is pre-existing).

### P2 — Role not validated with `isRole()` guard (`packages/docs/src/html/HtmlDocView.tsx:347`)

`if (m?.role) setRole(m.role)` sets the backend-supplied role from the `DocMeta` response without the `isRole()` type guard defined at `packages/docs/src/auth/roles.ts:11`. A backend regression returning an unexpected string (e.g. `"super_admin"`) would pass the truthy check and land in `role` state. Downstream impact is safe — `canManage()` only matches `'admin'`, so an invalid role produces `canGrant=false` via `computeCanGrant` and the forward button stays disabled rather than crashing. For defense-in-depth, consider: `if (m?.role && isRole(m.role)) setRole(m.role)`.

**Diff-scope:** new code introduced by this PR.

## Data Flow Trace

| Consumed value | Upstream source | Flows to consumer? |
|---|---|---|
| `ownerId` | `getDoc(docId).ownerId` (docs-backend) | Yes — gated by `typeof === 'string' && truthy`; consumed by `headerCreator`, `creatorAvatarUrl`, `creatorUid`, `computeCanGrant` |
| `createdAt` | `getDoc(docId).createdAt` (docs-backend) | Yes — gated by `typeof === 'string' && truthy`; passed to `DocMoreMenu.createdAt` |
| `role` | `getDoc(docId).role` (docs-backend) | Yes — gated by truthy; consumed by `doForward` → `computeCanGrant`. No `isRole()` guard (P2 above) |
| `creatorName` | `useMemberNames(space).get(ownerId)` → fallback `getUserName(ownerId)` | Yes — reset on `ownerId` change; member-map preferred in-shell; `getUserName` forced nickname-only for standalone |
| `canGrant` | `computeCanGrant(role, getCurrentUid(), ownerId)` | Yes — called inside `startDocForward`; early-returns if `!role` |

## Blindspot Checklist

- **C1 (dual-path parity):** clear — `creatorNicknameOnly` privacy gate applied at the only standalone consumer (`StandaloneDocPage`); in-shell consumer (`DocsHome`) correctly omits it. No remove/unsubscribe pairs touched.
- **C2 (control-flow ordering):** clear — `startDocForward` is called from a single call-site (`doForward` callback). No nesting or reuse.
- **C3 (authorization boundary):** N/A — no new endpoint or tool exposure. Forward-grant is gated by `computeCanGrant` + backend double-check.
- **C4 (authorization lifecycle):** N/A — no auth/jail changes.
- **C5 (build ≠ runtime):** N/A — no build/packaging changes.
- **C6 (governance/docs):** N/A — no governance or policy changes.

- **OctoDocMeta deprecation** — `creator_uid`, `creator_name`, `created_at` marked `@deprecated` with JSDoc pointers to the new source. No remaining readers of these fields in the diff.
- **`deleteDoc(effectiveSlug)` unchanged** — still called with `effectiveSlug` (not `docId`) at `:425`; this is pre-existing and outside the PR's diff scope.
- **`HtmlMemberPanel.creatorUid` upgrade** — previously `meta?.creator_uid` was always `undefined` (the `__ODOC__` blob doesn't carry it per the code comment at `:254`). Now `ownerId` from docs-backend populates it, so the member panel correctly shows the creator as the first row with `role: 'author'`. Intentional improvement.
- **`DocMoreMenu.CreatorAvatar`** — new `CreatorAvatar` component handles `` → initial-letter fallback via `onError`. `avatarInitial` receives a non-empty string (the `headerCreator` chain guarantees at least `'—'`).
- **`useMemberNames` called unconditionally** — safe because `getSpaceMemberNames('')` returns an empty `Map` (early-return at `packages/docs/src/members/memberNames.ts:30`).

## Verdict: APPROVED

Well-structured migration to a single data source for creator/role, with proper fail-soft handling and privacy gating. The two P2 items are defensive-improvement suggestions — no correctness, security, or functional blockers.

[Octo-Q] verdict: APPROVE — no P0/P1 found; two P2 defensive-improvement items (URL-encoding consistency, isRole guard) are non-blocking.

PR #915, opened 2026-07-20
refactor(summary): align list/detail UI with v2, add selection highlight

quanming1:
## Summary

Align the summary list/detail/workbench UI with the v2 design and add a list-selection highlight for the currently active item. In addition to the visual refactor, this PR adds a detail-page delete action (`api.deleteSummary`, creator-only) and consolidates the schedule button into the detail header (previously split across personal/team sections).

## Changes

**List selection highlight** (list-owned detail only)
- `SummaryDetailPage` gains an `emitSelection` prop (default `false`); only `SummaryListPage` passes `true` when pushing the detail route. Embedded instances (`ChatSummaryPanel`, `SummaryConfirmPage`) do not emit, avoiding cross-view highlight pollution.
- `SummaryListPage` tracks `activeTaskId` via `summary-detail-active`/`summary-detail-inactive` window events (race-safe: inactive clears only its own id), plus immediate `setState` on click and `componentDidUpdate` emit on in-place task switch.
- `SummaryCard` gains an `active` prop; the selected card shows a left `::before` accent bar (no background/border change).

**Detail-page delete** (new action)
- `handleDeleteTask` → `api.deleteSummary(taskId)` with creator-only permission gating; reuses existing `deleteContent`/`deleteSuccess`/`deleteFailed` i18n keys; emits `summary-list-refresh-requested` (now subscribed by `SummaryListPage`) so the list reloads after `popToRoot`.

**Card / detail header**
- Flatten the "..." dropdown menu into explicit buttons (regenerate, cancel, delete, leave). lucide icons; native `title` tooltip; 2-line title clamp; hover-revealed delete/leave actions.
- Preserves the `#907` agent-forward loading gate and `#158/#161` agent regenerate/edit guards.
- `SHOW_FORWARD_TO_MATTER` stays `false`.

**Detail body / workbench / sources / i18n**
- Personal summary section is collapsible; `SelectedSourcesPanel` moved to a sticky footer; content max-width 920px.
- SVG bot icon; char count; drop the "more templates" modal; `#f5f6f7` background; responsive template grid.
- lucide icons; empty-state copy; `<section>` semantics. Drop emoji; add `source.empty`; rename `templatesTitle`.

## Schedule button relocation (need5 → v2)
[...]

PR #900, opened 2026-07-20
feat(summary): increase template content limit

Liuxy20:
## Summary

Increase smart-summary template content from 200 to 2000 characters for both built-in template overrides and custom templates while preserving the existing template display and submission format.

## Related Issue

Fixes #899

## Changes

- Add shared constants for the 100-character template name limit and 2000-character template content limit.
- Apply the 2000-character content limit in the summary workbench and chat summary modal.
- Preserve the existing `总结主题 / 内容重点` framing, topic payload, and title behavior.
- Limit only the template content when editing an applied template, so framing text does not truncate valid content.
- Add boundary and apply-edit-submit coverage for both template entry points.

## Backend Compatibility

The companion service change in Mininglamp-OSS/octo-smart-summary#163 expands template `description` validation and storage to 2000 Unicode characters. It also expands the summary `topic` safety limit and the task/schedule title storage to 2300 Unicode characters, safely accommodating a 2000-character description, a 100-character template name, and the existing localized framing labels.

Deploy the service change and its database migrations before or together with this frontend change.

## Architecture / Module Boundary

- Affected module(s): dmworksummary
- New or changed user-visible entry point: no
- Shared layer touched: summary template utilities and limits
- Impact scope: summary workbench and chat summary modal only
- Duplicate entry point checked: not applicable

## Testing

- [x] Unit tests added/updated
- [x] Template content accepts 2000 characters and rejects overflow
- [x] Applied 2000-character content remains intact after editing
- [x] Existing template display and submission format remain unchanged
- [x] Existing textarea vertical-scroll behavior remains unchanged

## Checklist

- [x] PR description is in English
- [x] Added tests for the changes
- [x] Confirmed the change does not alter UI copy or add a new entry point
- [x] Followed Conventional Commits

PR #951, opened 2026-07-21
feat(summary): unify chat summary panel with /summary workbench

quanming1:
## Summary

### DIFF
--- a/apps/web/vite.config.ts
+++ b/apps/web/vite.config.ts
@@ -138,7 +138,7 @@ export default defineConfig(({ mode }) => {
             env.VITE_SUMMARY_API_URL || apiOrigin || "http://localhost:8080",
           changeOrigin: true,
           secure: false,
-          rewrite: (path: string) => path.replace(/^\/summary/, ""),
+          // The upstream summary service is mounted at /summary/api/v1; preserve the prefix.
         },
         // Marketplace (MCP catalog) API — must be before the general /api/ rule.
         // octo-marketplace serves its own /api/v1/*; the /market prefix is
--- a/packages/dmworkbase/src/Components/VoiceInputButton/index.tsx
+++ b/packages/dmworkbase/src/Components/VoiceInputButton/index.tsx
@@ -17,6 +17,8 @@ const VOICE_MODES: { value: VoiceMode; labelKey: string }[] = [
 ];
 
 const FLOATING_GAP = 12;
+const FLOATING_WIDTH = 184;
+const FLOATING_HORIZONTAL_MARGIN = 8;
 const INDICATOR_HEIGHT = 48;
 const PREPARING_DELAY_MS = 300;
 const RECORDING_DELAY_MS = 500;
@@ -94,9 +96,15 @@ export default function VoiceInputButton({
   const updateFloatingPosition = useCallback(() => {
     if (!buttonRef.current) return;
     const rect = buttonRef.current.getBoundingClientRect();
+    const preferredLeft = rect.left + rect.width / 2 - FLOATING_WIDTH / 2;
+    const maxLeft = window.innerWidth - FLOATING_WIDTH - FLOATING_HORIZONTAL_MARGIN;
+    const left = Math.max(
+      FLOATING_HORIZONTAL_MARGIN,
+      Math.min(preferredLeft, maxLeft),
+    );
     setFloatingPosition({
       top: rect.top - FLOATING_GAP - INDICATOR_HEIGHT,
-      left: rect.left + rect.width / 2,
+      left,
     });
   }, []);
 
@@ -352,7 +360,7 @@ export default function VoiceInputButton({
         style={{
           top: floatingPosition.top,
           left: floatingPosition.left,
-          transform: "translateX(-50%)",
+          transform: "none",
         }}
       >
         <div className="wk-voice-floating-content">
--- a/packages/dmworkbase/src/EndpointCommon.tsx
+++ b/packages/dmworkbase/src/EndpointCommon.tsx
@@ -357,10 +357,12 @@ export class EndpointCommon {
   chatSummaryPanel(
     channel: Channel,
     onClose: () => void,
+    summaryPanelView?: "history" | "new",
   ): JSX.Element | undefined {
     return EndpointManager.shared.invoke(EndpointCategory.chatSummaryPanel, {
       channel,
       onClose,
+      summaryPanelView,
     });
   }
 
--- a/packages/dmworkbase/src/Pages/Chat/index.tsx
+++ b/packages/dmworkbase/src/Pages/Chat/index.tsx
@@ -1480,7 +1480,8 @@ export class ChatContentPage extends Component<
         {showSummaryPanel && (
           <div className="wk-summary-panel">
             {WKApp.endpoints.chatSummaryPanel(channel, () =>
-              this.setState({ showSummaryPanel: false })
+              this.setState({ showSummaryPanel: false }),
+              summaryPanelView
             )}
           </div>
         )}
--- a/packages/dmworksummary/src/components/ChatSummaryHistory.tsx
+++ b/packages/dmworksummary/src/components/ChatSummaryHistory.tsx
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import axios from 'axios';
 import { Toast } from '@douyinfe/semi-ui';
 import { I18nContext } from '@octo/base';
+import { X } from 'lucide-react';
 import * as summaryApi from '../api/summaryApi';
 import type { SummaryListItem } from '../types/summary';
 import { TaskStatus } from '../types/summary';
@@ -9,6 +10,7 @@ import SummaryCard from './SummaryCard';
 
 interface ChatSummaryHistoryProps {
     channel: { channelID: string; channelType: number };
+    onClose: () => void;
     onCreateNew: () => void;
     onViewDetail: (taskId: number) => void;
     paused?: boolean;
@@ -162,14 +164,30 @@ export default class ChatSummaryHistory extends Component<
     };
 
     render() {
-        const { onCreateNew, onViewDetail } = this.props;
+        const { onClose, onCreateNew, onViewDetail } = this.props;
         const { items, loading } = this.state;
         const { t } = this.context;
 
         return (
-            <div className="chat-summary-history" style={{ padding: '16px', height: '100%', display: 'flex', flexDirection: 'column' }}>
-                <div style={{ fontSize: 16, fontWeight: 600, marginBottom: 16, color: 'var(--wk-text-primary, #1C1F23)' }}>
-                    {t('summary.chatSummary.panelTitle')}
+            <div className="chat-summary-history" style={{ padding: '8px 12px 16px', height: '100%', display: 'flex', flexDirection: 'column' }}>
+                <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'space-between', marginBottom: 12 }}>
+                    <span style={{ fontSize: 16, fontWeight: 600, color: 'var(--wk-text-primary, #1C1F23)' }}>
+                        {t('summary.chatSummary.panelTitle')}
+                    </span>
+                    <button
+                        onClick={onClose}
+                        style={{
+                            background: 'none',
+                            border: 'none',
+                            cursor: 'pointer',
+                            padding: 4,
+                            color: 'var(--wk-text-secondary, #646A73)',
+                            display: 'flex',
+                            alignItems: 'center',
+                        }}
+                    >
+                        <X size={18} />
+                    </button>
                 </div>
 
                 <div
@@ -198,15 +216,20 @@ export default class ChatSummaryHistory extends Component<
                 </div>
 
                 {loading ? (
-                    <div style={{ textAlign: 'center', color: 'var(--wk-text-tertiary, #8F959E)', paddingTop: 40 }}>
+                    <div className="chat-summary-history-empty chat-summary-history-empty--loading">
                         {t('summary.common.loading')}
                     </div>
                 ) : items.length === 0 ? (
-                    <div style={{ textAlign: 'center', color: 'var(--wk-text-tertiary, #8F959E)', paddingTop: 40 }}>
-                        {t('summary.list.emptyTitle')}
+                    <div className="chat-summary-history-empty">
+                        <div className="chat-summary-history-empty-title">
+                            {t('summary.list.emptyTitle')}
+                        </div>
+                        <div className="chat-summary-history-empty-description">
+                            {t('summary.chatSummary.emptyDescription')}
+                        </div>
                     </div>
                 ) : (
-                    <div style={{ flex: 1, overflowY: 'auto' }}>
+                    <div className="chat-summary-history-list">
                         {items.map((item) => (
                             <SummaryCard
                                 key={item.task_id}
--- a/packages/dmworksummary/src/components/ChatSummaryNewModal.css
+++ b/packages/dmworksummary/src/components/ChatSummaryNewModal.css
@@ -2,6 +2,90 @@
     display: none;
 }
 
+.chat-summary-modal-embedded {
+    height: 100%;
+    min-width: 0;
+    box-sizing: border-box;
+    display: flex;
+    flex-direction: column;
+    padding: 16px 16px 12px;
+    overflow: auto;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-header {
+    margin-bottom: 4px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-title {
+    font-size: 17px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-desc {
+    margin-bottom: 10px;
+    line-height: 18px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-input-area {
+    min-height: 96px;
+    margin-bottom: 10px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-input {
+    min-height: 48px;
+    padding: 10px 12px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-templates-heading {
+    margin-bottom: 8px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-templates-label {
+    margin-bottom: 8px;
+    padding: 0 4px;
+}
+
+.chat-summary-modal-embedded .summary-template-more-button {
+    padding-right: 4px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-templates {
+    gap: 8px;
+    padding: 0 4px 8px;
+}
+
+.chat-summary-modal-embedded .chat-summary-template-card {
+    min-height: 132px;
+    padding: 10px;
+}
+
+.chat-summary-modal-embedded .chat-summary-template-card-icon {
+    margin-bottom: 6px;
+}
+
+.chat-summary-modal-embedded .summary-template-custom-section {
+    padding: 2px 4px 8px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-chat-section {
+    margin-bottom: 8px;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-footer {
+    position: sticky;
+    bottom: -12px;
+    flex-shrink: 0;
+    margin: auto -16px -12px;
+    padding: 10px 16px 12px;
+    background: var(--semi-color-bg-0);
+    border-top: 1px solid var(--semi-color-border);
+    z-index: 1;
+}
+
+.chat-summary-modal-embedded .chat-summary-modal-split {
+    margin-left: auto;
+}
+
+
 .chat-summary-modal-header {
     display: flex;
     align-items: center;
@@ -42,7 +126,7 @@
 }
 
 .chat-summary-modal-input-area:focus-within {
-    border-color: var(--wk-color-primary, #3370FF);
+    border-color: var(--wk-border-default, #E5E6EB);
 }
 
 .chat-summary-modal-input-wrap {
--- a/packages/dmworksummary/src/components/ChatSummaryNewModal.tsx
+++ b/packages/dmworksummary/src/components/ChatSummaryNewModal.tsx
@@ -26,6 +26,9 @@ interface ChatSummaryNewModalProps {
     channel: { channelID: string; channelType: number };
     onClose: () => void;
     onSubmit: (taskId: number) => void;
+    /** When true, render only the form content (no Modal shell) so the panel
+     *  can embed this component inline. */
+    embedded?: boolean;
 }
 
 interface ChatSummaryNewModalState {
@@ -671,7 +674,7 @@ export default class ChatSummaryNewModal extends Component<
     };
 
     render() {
-        const { visible, onClose } = this.props;
+        const { visible, onClose, embedded } = this.props;
         const {
             topic, appliedTemplateLabel, customTemplateLimit, mode, templates, selectedChats, showChatSelector, submitting, agentSubmitting, scheduleConfig, showScheduleConfig, showMoreTemplates,
             editingTemplate, creatingCustomTemplate,
@@ -745,27 +748,17 @@ export default class ChatSummaryNewModal extends Component<
             </div>
         );
 
-        return (
+        const modalContent = (
             <>
-                <Modal
-                    visible={visible}
-                    onCancel={onClose}
-                    footer={footer}
-                    width={640}
-                    closable
-                    title={null}
-                    bodyStyle={{ padding: '24px 24px 0' }}
-                    className="chat-summary-new-modal"
-                >
-                    <div className="chat-summary-modal-header">
-                        <span className="chat-summary-modal-title">{t('summary.create.title')}</span>
-                        <span className="chat-summary-modal-ai-badge">AI+</span>
-                    </div>
-                    <div className="chat-summary-modal-desc">
-                        {t('summary.create.desc')}
-                    </div>
+                <div className="chat-summary-modal-header">
+                    <span className="chat-summary-modal-title">{t('summary.create.title')}</span>
+                    <span className="chat-summary-modal-ai-badge">AI+</span>
+                </div>
+                <div className="chat-summary-modal-desc">
+                    {t('summary.create.desc')}
+                </div>
 
-                    <div className="chat-summary-modal-input-area">
+                <div className="chat-summary-modal-input-area">
                         {isAgent ? (
                             // 弹窗内高度受限：固定面板高度让内部消息列表滚动。
                             <div className="chat-summary-modal-agent-chat" style={{ height: 360 }}>
@@ -964,7 +957,30 @@ export default class ChatSummaryNewModal extends Component<
                             </div>
                         )}
                     </div>
-                </Modal>
+            </>
+        );
+
+        return (
+            <>
+                {embedded ? (
+                    <div className="chat-summary-modal-embedded">
+                        {modalContent}
+                        {footer}
+                    </div>
+                ) : (
+                    <Modal
+                        visible={visible}
+                        onCancel={onClose}
+                        footer={footer}
+                        width={640}
+                        closable
+                        title={null}
+                        bodyStyle={{ padding: '24px 24px 0' }}
+                        className="chat-summary-new-modal"
+                    >
+                        {modalContent}
+                    </Modal>
+                )}
 
                 <ChatSelectorModal
                     visible={showChatSelector}
--- a/packages/dmworksummary/src/components/ChatSummaryPanel.tsx
+++ b/packages/dmworksummary/src/components/ChatSummaryPanel.tsx
@@ -7,17 +7,20 @@ import {
     persistSummaryWidth,
 } from '@octo/base/src/Components/WKLayout/layoutWidth';
 import { X, ChevronLeft } from 'lucide-react';
-import ChatSummaryHistory from './ChatSummaryHistory';
+import SummaryListPage from '../pages/SummaryListPage';
+import SummaryCreatePage from '../pages/SummaryCreatePage';
 import SummaryDetailPage from '../pages/SummaryDetailPage';
 
 interface ChatSummaryPanelProps {
     visible: boolean;
     channel: { channelID: string; channelType: number };
     onClose: () => void;
+    /** Initial view: 'history' shows the list, 'new' shows the create form. */
+    summaryPanelView?: 'history' | 'new';
 }
 
 interface ChatSummaryPanelState {
-    view: 'list' | 'detail';
+    view: 'list' | 'detail' | 'create';
     selectedTaskId: number | null;
     isDragging: boolean;
 }
@@ -36,7 +39,8 @@ export default class ChatSummaryPanel extends Component<
 
     constructor(props: ChatSummaryPanelProps) {
         super(props);
-        this.state = { view: 'list', selectedTaskId: null, isDragging: false };
+        const initialView = props.summaryPanelView === 'new' ? 'create' : 'list';
+        this.state = { view: initialView, selectedTaskId: null, isDragging: false };
     }
 
     componentDidMount() {
@@ -59,8 +63,9 @@ export default class ChatSummaryPanel extends Component<
             prev.channelID !== next.channelID ||
             prev.channelType !== next.channelType
         ) {
-            // 切换会话时回到列表视图，避免残留上个会话的详情
-            this.setState({ view: 'list', selectedTaskId: null });
+            // 切换会话时回到初始视图，避免残留上个会话的详情/表单
+            const initialView = this.props.summaryPanelView === 'new' ? 'create' : 'list';
+            this.setState({ view: initialView, selectedTaskId: null });
         }
     }
 
@@ -116,11 +121,7 @@ export default class ChatSummaryPanel extends Component<
     };
 
     private handleCreateNew = () => {
-        const { channel } = this.props;
-        WKApp.mittBus.emit('wk:open-summary-modal', {
-            channelId: channel.channelID,
-            channelType: channel.channelType,
-        });
+        this.setState({ view: 'create', selectedTaskId: null });
     };
 
     private handleViewDetail = (taskId: number) => {
@@ -131,11 +132,23 @@ export default class ChatSummaryPanel extends Component<
         this.setState({ view: 'list' });
     };
 
+    private handleBackToList = () => {
+        this.setState({ view: 'list' });
+    };
+
+    private handleCreateSubmit = (taskId: number) => {
+        this.setState({ view: 'list' });
+        setTimeout(() => {
+            WKApp.mittBus.emit("summary-list-refresh-requested" as any);
+        }, 800);
+    };
+
     render() {
         const { channel, onClose } = this.props;
         const { view, selectedTaskId, isDragging } = this.state;
         const { t } = this.context;
         const isDetail = view === 'detail' && selectedTaskId != null;
+        const isCreate = view === 'create';
 
         return (
             <div ref={this.rootRef} style={{ height: '100%', display: 'flex', flexDirection: 'column' }}>
@@ -148,43 +161,46 @@ export default class ChatSummaryPanel extends Component<
                     <div className="wk-thread-panel-splitter-line" />
                 </div>
 
-                {/* 列表视图：常驻挂载，详情时隐藏以保留滚动位置 */}
+                {/* 列表视图：复用 SummaryListPage（传 channelId 过滤当前聊天） */}
                 <div
                     style={{
-                        display: isDetail ? 'none' : 'flex',
+                        display: isDetail || isCreate ? 'none' : 'flex',
                         flex: 1,
                         minHeight: 0,
                         flexDirection: 'column',
                     }}
                 >
-                    <div style={{
-                        display: 'flex',
-                        justifyContent: 'flex-end',
-                        padding: '8px 8px 0',
-                    }}>
-                        <button
-                            onClick={onClose}
-                            style={{
-                                background: 'none',
-                                border: 'none',
-                                cursor: 'pointer',
-                                padding: 4,
-                                color: 'var(--wk-text-secondary, #646A73)',
-                                display: 'flex',
-                                alignItems: 'center',
-                            }}
-                        >
-                            <X size={18} />
-                        </button>
-                    </div>
-
-                    <ChatSummaryHistory
-                        channel={channel}
+                    <SummaryListPage
+                        channelId={channel.channelID}
+                        onClose={onClose}
                         onCreateNew={this.handleCreateNew}
                         onViewDetail={this.handleViewDetail}
-                        paused={isDetail}
                     />
                 </div>
+
+                {/* 创建视图：返回按钮 + 内嵌创建表单（复用 SummaryCreatePage embedded） */}
+                {isCreate && (
+                    <div className="wk-summary-panel-detail">
+                        <div className="wk-summary-panel-detail-back">
+                            <button
+                                type="button"
+                                className="wk-summary-panel-detail-back-btn"
+                                onClick={this.handleBackToList}
+                            >
+                                <ChevronLeft size={18} />
+                                <span>{t('summary.chatSummary.back')}</span>
+                            </button>
+                        </div>
+                        <div className="wk-summary-panel-detail-body" style={{ overflow: 'auto', flex: 1 }}>
+                            <SummaryCreatePage
+                                channel={channel}
+                                embedded={true}
+                                onClose={this.handleBackToList}
+                                onSubmit={this.handleCreateSubmit}
+                            />
+                        </div>
+                    </div>
+                )}
 
                 {/* 详情视图：复用整页 SummaryDetailPage，外层加面板级返回栏 */}
                 {isDetail && (
--- a/packages/dmworksummary/src/components/ChatSummaryStarButton.tsx
+++ b/packages/dmworksummary/src/components/ChatSummaryStarButton.tsx
@@ -115,18 +115,11 @@ export default class ChatSummaryStarButton extends Component<
 
     private emitForHasSummaries(hasSummaries: boolean) {
         const { channel } = this.props;
-        if (hasSummaries) {
-            WKApp.mittBus.emit('wk:toggle-summary-panel', {
-                channelId: channel.channelID,
-                channelType: channel.channelType,
-                summaryPanelView: 'history',
-            });
-        } else {
-            WKApp.mittBus.emit('wk:open-summary-modal', {
-                channelId: channel.channelID,
-                channelType: channel.channelType,
-            });
-        }
+        WKApp.mittBus.emit('wk:toggle-summary-panel', {
+            channelId: channel.channelID,
+            channelType: channel.channelType,
+            summaryPanelView: hasSummaries ? 'history' : 'new',
+        });
     }
 
     render() {
--- a/packages/dmworksummary/src/i18n/en-US.json
+++ b/packages/dmworksummary/src/i18n/en-US.json
@@ -158,6 +158,7 @@
         "deleteSuccess": "Deleted",
         "networkError": "Network connection error. Check your connection and retry.",
         "emptyTitle": "No summaries yet",
+        "emptyDescription": "Create a summary to organize this conversation.",
         "emptyDesc": "Generate group or personal work summaries quickly, and let AI organize key information for you.",
         "createFirst": "Create first summary",
         "leaveSuccess": "Left",
@@ -203,6 +204,7 @@
     "chatSummary": {
         "panelTitle": "Summaries in this chat",
         "createNew": "New summary",
+        "emptyDescription": "Create a summary to organize this conversation.",
         "back": "Back",
         "starTooltip": "AI Summary"
     },
--- a/packages/dmworksummary/src/i18n/zh-CN.json
+++ b/packages/dmworksummary/src/i18n/zh-CN.json
@@ -203,6 +203,7 @@
     "chatSummary": {
         "panelTitle": "聊天内的智能总结",
         "createNew": "新建总结",
+        "emptyDescription": "创建一个总结，开始整理当前聊天内容。",
         "back": "返回",
         "starTooltip": "智能总结"
     },
--- a/packages/dmworksummary/src/index.css
+++ b/packages/dmworksummary/src/index.css
@@ -15,6 +15,35 @@
     overflow-y: auto;
 }
 
+/* 面板模式：去掉 max-width 居中，收紧 padding，占满父容器宽度 */
+.summary-list-page--panel {
+    max-width: none;
+    margin: 0;
+    width: 100%;
+    padding: 8px 12px 16px;
+}
+
+/* 面板模式 toolbar 收紧 */
+.summary-list-page--panel .summary-list-toolbar {
+    flex-wrap: wrap;
+    gap: 8px;
+}
+
+.summary-list-page--panel .summary-list-search {
+    flex: 1 1 120px;
+    min-width: 0;
+}
+
+.summary-list-page--panel .summary-list-actions {
+    flex-wrap: nowrap;
+}
+
+.summary-list-header-actions {
+    display: flex;
+    align-items: center;
+    gap: 4px;
+}
+
 /* 详情页特殊布局：更宽敞的全屏展示 */
 .summary-detail-page {
     --wk-ring-primary: 0 0 0 2px var(--semi-color-primary-light-default);
@@ -170,7 +199,8 @@
     cursor: pointer;
     transition: background-color 0.15s, border-color 0.15s, box-shadow 0.15s;
     background: var(--semi-color-bg-0);
-    box-shadow: 0 1px 2px rgba(28, 31, 35, 0.03);
+    border: 1px solid var(--semi-color-border);
+    box-shadow: none;
 }
 
 .summary-card-attention-dot {
@@ -304,6 +334,41 @@
     padding: 12px;
     border-radius: 8px;
     margin-bottom: 8px;
+}
+
+.chat-summary-history-empty {
+    flex: 1;
+    min-height: 0;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    gap: 8px;
+    padding: 24px 16px 48px;
+    text-align: center;
+    color: var(--wk-text-tertiary, #8F959E);
+}
+
+.chat-summary-history-empty--loading {
+    padding-top: 24px;
+}
+
+.chat-summary-history-empty-title {
+    color: var(--wk-text-secondary, #646A73);
+    font-size: 14px;
+}
+
+.chat-summary-history-empty-description {
+    max-width: 260px;
+    color: var(--wk-text-tertiary, #8F959E);
+    font-size: 12px;
+    line-height: 18px;
+}
+
+.chat-summary-history-list {
+    flex: 1;
+    min-height: 0;
+    overflow-y: auto;
 }
 
 /* ========== Create Page ========== */
@@ -478,8 +543,15 @@
     min-width: 0;
 }
 
-.summary-detail-header-actions .semi-button {
-    font-weight: 400;
+.wk-summary-panel-detail .summary-detail-header-actions .semi-button .semi-icon {
+    font-size: 16px;
+}
+
+/* 面板宽度 ≤ 520px 时隐藏按钮文字，只显示 icon */
+@media (max-width: 520px) {
+    .wk-summary-panel-detail .summary-detail-action-label {
+        display: none;
+    }
 }
 
 .summary-detail-loading {
@@ -1492,6 +1564,58 @@
     background: #f5f6f7;
 }
 
+/* 面板模式：收紧 padding，去掉 max-width 居中 */
+.summary-workbench--panel {
+    padding: 12px 12px 16px;
+}
+
+.summary-workbench--panel > .summary-workbench-header,
+.summary-workbench--panel > .summary-workbench-input-area,
+.summary-workbench--panel > .summary-workbench-actions {
+    max-width: none;
+}
+
+/* 面板模式 actions 收紧：不换行、隐藏归档提示、缩小按钮 */
+.summary-workbench--panel .summary-workbench-actions {
+    padding: 8px 12px;
+    flex-wrap: nowrap;
+    margin-bottom: 8px;
+}
+
+/* 面板模式：workbench 自身不滚动，交给 wk-summary-panel-detail-body 统一滚动 */
+.summary-workbench--panel {
+    overflow: visible;
+    height: auto;
+    min-height: 100%;
+}
+
+/* 面板模式：input-area 自适应高度，不裁切内容 */
+.summary-workbench--panel .summary-workbench-input-area {
+    overflow: visible;
+    height: auto;
+    min-height: max-content;
+    flex: 0 0 auto;
+    margin-bottom: 12px;
+}
+
+/* 面板模式 selected-chats 已在通用规则中统一 padding */
+
+.summary-workbench--panel .summary-workbench-actions-left {
+    gap: 4px;
+    flex-wrap: nowrap;
+    width: auto;
+}
+
+.summary-workbench--panel .summary-workbench-actions-left .semi-button {
+    padding: 0 8px;
+    height: 28px;
+    font-size: 12px;
+}
+
+.summary-workbench--panel .summary-workbench-archived-notice {
+    display: none;
+}
+
 .summary-workbench > .summary-workbench-header,
 .summary-workbench > .summary-workbench-input-area,
 .summary-workbench > .summary-workbench-actions {
@@ -1543,7 +1667,9 @@
 .summary-workbench-input-area {
     border: 1px solid var(--semi-color-border);
     border-radius: 8px;
-    overflow: hidden;
+    overflow: visible;
+    height: auto;
+    flex: 0 0 auto;
     margin-bottom: 20px;
     min-height: 96px;
     display: flex;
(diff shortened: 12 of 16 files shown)


lml2468 (changes requested):
## Review Verdict：需修改（REQUEST_CHANGES）—— 与 main 有真实合并冲突(DIRTY),需先 rebase

审查锚定 live head `2dfa0a71`(触发时的 `5a0336a6` 在 review 途中被新推覆盖)、base `main`。**不做逐行深审**——PR 与 main 有真实合并冲突,`mergeable=CONFLICTING / DIRTY`,当前无法合并,且冲突文件 rebase 后会大改。

### ❌ 阻断:与 main 冲突(DIRTY)+ 落后 16 commit

- merge-base `2496dad1`,**落后 main 16 个 commit**;
- **真实内容冲突(3-way `git merge-tree --write-tree` exit=1 坐实)**:
  - `packages/dmworksummary/src/pages/SummaryDetailPage.tsx`
  - `packages/dmworksummary/src/pages/SummaryListPage.tsx`
  - (上一 head 的 `index.css` 冲突这轮已消,但上面两个仍在)
- 冲突对象是已合入 main 的 **#915(list/detail UI v2)+ #900(template content limit)**——本 PR 从旧 base 切出,与后来合入的 summary PR 撞在同一批文件。
- DIRTY 下 GitHub 直接禁止合并。**修法:rebase 到最新 main,解决 SummaryDetailPage / SummaryListPage 与 #915/#900 的冲突**,再推。

### 🟡 附带

- **PR 描述基本为空**(只有 `## Summary` 标题)。这是「统一 chat summary panel 与 /summary workbench」的 +1122/−420 大改,涉及 List/Detail/Create + ChatSummaryPanel/Modal/History 多处,请补描述(改了什么、行为变更、与 #915/#947 的关系),否则复审难判「有意重设计 vs 冲突残留」。

### 为什么不逐行审

- 两个冲突文件(SummaryDetailPage 是本仓最高频改动点)rebase 解决冲突后会显著变化,现逐行审的结论到时多作废;
- 无法把「本 PR 真实改动」与「旧 base 遗留、已被 #915/#900 取代的部分」可靠区分——stale-base 深审误判高发区;
- 与我对 #900 那轮 DIRTY 的一致处理相同。

### 结论

改动主题(统一 chat summary panel 与 workbench)合理,但**当前与 main 有真实冲突、落后 16 commit、DIRTY 不可合**。rebase 到最新 main + 解决 SummaryDetailPage / SummaryListPage 与 #915/#900 的冲突 + 补 PR 描述后,重新提交即可复审深审。当前维持 REQUEST_CHANGES(合并状态阻塞,非代码质量结论)。

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #952 (octo-web)

## Summary

This PR refines the skill market card workflow across five commits. Key changes: creator/owner metadata display with bot-creator icon support, sort controls relocated from the toolbar to a segmented button group in the result summary, icon upload restructured from inline `<label>/<input>` to a `<button>`-triggered hidden input with avatar-text placeholder, search bar gains a clear button, install button hidden on owner cards, and the edit modal's file-info section replaced with a skill-identity display. CSS receives extensive polish (segmented sort pills, action button borders, responsive layout tweaks).

Overall a well-scoped UI polish PR with thorough test updates. No P0/P1 issues; two minor items below.

## Verification

Static analysis only at head `c47f0f246f84d3beb443e0dc629cd5e1da4210af`; build and tests not executed in this environment.

- ✅ **Creator/owner display logic** — `hasSeparateCreator` guard correctly requires all three conditions (`creatorId`, `ownerId`, inequality). `isBotCreator` uses `endsWith("_bot")` on `creatorId` which is always `skill.creatorId` when `hasSeparateCreator` is true. Both branches verified against test fixtures (`bot-1` → non-bot path, `publisher_bot` → bot path).
- ✅ **Install button gating** — `!isOwnerCard && onInstall` correctly hides install on owner cards. `isOwnerCard = Boolean(onEdit || onDelete)` covers all owner-action scenarios. Tests confirm install absent on mine tab.
- ✅ **IconCropModal safety** — Moving `!file` guard inside the render is safe: both callers pass `visible={!!iconCropFile}`, so the modal is never visible with a null file. `handleSave` guards against null `editorRef.current` via `if (!canvas) return`.
- ✅ **Icon upload `accept` widening** — `image/*` replaces explicit MIME types. AvatarEditor loads via ``, which handles browser-supported formats. Output is always `toBlob("image/png")`. No new validation gap.
- ✅ **i18n parity** — New keys (`skillName`, `reuploadShort`) present in both `en-US.json` and `zh-CN.json`. `filter.clear` key pre-existed.
- ✅ **hideDescriptionTooltip before actions** — Correctly called before `onInstall`, `onEdit`, `onDelete` to prevent stale tooltips. Test coverage confirms tooltip dismissed before edit click.
- ✅ **Sort control relocation** — Sort now inside `{!list.loading && !list.error}` is appropriate (sort irrelevant during loading/error). `aria-pressed` added for accessibility.

## Findings

No P0/P1 issues; one P2 and one Nit below.

### P2 — `title` tooltip removed from truncated skill name (`packages/dmworkskillmarket/src/components/SkillCard.tsx:190`)

The `.skill-market-card__name` element still has `overflow: hidden; text-overflow: ellipsis; white-space: nowrap` in CSS (`index.css:757`), so long skill names are visually truncated with ellipsis. The old `title={skill.name}` attribute provided a native browser tooltip showing the full name on hover — this PR removes it. Users with long skill names (e.g. `meeting-note-cleaner-with-extra-context`) lose the ability to read the full name without opening the detail modal.

Consider restoring `title={skill.name}` on the `<span>`, or adding a custom tooltip consistent with the description tooltip pattern already used in this component.

### Nit — Orphaned `skillMarket.sort.label` i18n key (`packages/dmworkskillmarket/src/i18n/en-US.json:60`)

The sort UI redesign removed the `<span className="skill-market-sort__label">{t("skillMarket.sort.label")}</span>` consumer from `SkillListPage.tsx`. The `"label": "Sort:"` entry in `en-US.json:60` and `"label": "排序："` in `zh-CN.json:60` are now dead — grep confirms no remaining reference to `sort.label` in any source file. Worth removing to keep locale files clean.

- Data flow: `skill.creatorId` / `skill.ownerId` traced from `RawSkill` type through to `SkillCard` display. Both fields are optional in the type, and `hasSeparateCreator` guard correctly handles all null/undefined/empty combinations.
- `accept="image/*"` widening: no server-side MIME enforcement visible from this repo; browser file picker and `` decode are the only gates. Both handle standard image formats.
- Search clear button: `clearSearch()` calls `onChange("")` and focuses input. `skillMarket.filter.clear` i18n key pre-existed in both locales.
- Sort button `descending` flag is display-only (ArrowDown icon), not passed to the API query.
- IconCropModal `width={460}` and `zIndex={1100}` additions are reasonable for the crop editor layout.
- CSS responsive overrides preserve `skill-market-upload-file--identity` grid in mobile breakpoint.
- Test coverage: new tests cover creator/owner display, bot icon, tooltip-dismiss-before-action, icon-input click delegation, and duplicate-save debounce.

## Verdict: APPROVED

Well-executed UI polish with comprehensive test updates. The P2 tooltip removal and orphaned i18n key are non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; two minor non-blocking items (P2 tooltip removal, Nit orphaned i18n key).

yujiawei (changes requested):
# Code Review — PR #951 (octo-web)

Reviewed at head SHA `5a0336a6ca5ab1c91cc0c1d6e436c9bc9de5a240` against merge-base `2496dad1`.

**Verdict: CHANGES_REQUESTED.** The panel-unification direction is good and the `/summary` proxy fix is correct, but this change deletes several deliberately-added guards for agent-triggered summaries, ships an inverted `origin_channel_type` mapping, breaks the panel's own navigation on delete/leave, and leaves the existing test suite red. Details below.

---

## Spec / scope

No linked plan was found (PR body is just `## Summary`), so this is reviewed against the code's own established contracts. The stated goal — unify the in-chat summary panel with the `/summary` workbench by reusing `SummaryListPage` / `SummaryCreatePage` / `SummaryDetailPage` — is achieved structurally. The concern is that the unification path **removed guards that were added on purpose** (`#158/#161`, `#907`) without re-implementing the behavior they protected, so this is not purely additive.

---

## Blocking (P0/P1)

### P1-1 — `origin_channel_type` mapping is inverted vs. the rest of the codebase
`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:454-462`

The new panel-mode block hand-rolls the mapping as `person→1, group→2, thread→3`:
```ts
// 后端要求 origin_channel_type: 1=私聊, 2=群聊, 3=子区
if (ch.channelID.includes("____")) { params.origin_channel_type = 3; }
else if (ch.channelType === 1)     { params.origin_channel_type = 1; }
else                               { params.origin_channel_type = 2; }
```
But the established contract everywhere else is the opposite. `SourceType` (`types/summary.ts:28-32`) is `GROUP_CHAT=1, THREAD=2, DIRECT_MESSAGE=3`; the sibling `ChatSummaryNewModal` — the very component this panel replaces — sends `origin_channel_type: getSourceType(channel)` (`ChatSummaryNewModal.tsx:401`), and the type doc for the agent-create params says `1=群聊 / 2=子区 / 3=私聊` (`types/summary.ts:267`). This file even **imports** `getOriginChannelType` (`SummaryCreatePage.tsx:19`) but doesn't use it here.

Failure scenario: create a summary from a **direct-message** chat via the panel → FE sends `origin_channel_type=1` (which the backend reads as group), and from a **group** chat → sends `2` (thread). Every panel-created summary is tagged with the wrong channel type, corrupting `origin_channel_id` reverse-lookup / filtering. Fix: replace the inline block with `getOriginChannelType(new Channel(ch.channelID, ch.channelType))` so both entry points share one mapping. *(Independent finding — not surfaced by either advisory leg; please double-check the true backend enum before fixing, but the internal inconsistency is unambiguous.)*

### P1-2 — Forwarding an agent summary silently no-ops (regression of #907)
`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1917-1919`, `:1979`, `:1982`

`handleForwardToChat`, `handleForwardToMatter`, and `handleMatterSelected` now read only `detail.result?.content` and drop the documented `?? personalResult?.content` fallback:
```ts
const { detail } = this.state;
if (!detail?.result?.content?.trim()) return;   // agent summaries have detail.result === null
```
Agent-triggered (`trigger_type === AGENT`) summaries only populate the `personal_result` table — the deleted comment and `renderPersonalSummary()` (`:2363-2419`) both confirm the visible body comes from `personalResult`, not `detail.result`. Meanwhile the "Forward to chat" menu item renders for **every** completed summary (`:3245-3249`, no agent guard). Failure scenario: open a completed AI/agent summary → click "Forward to chat" → the handler returns immediately, nothing is sent, no error shown. Same for the now-enabled "Forward to Matter". Fix: restore the `?? personalResult?.content` fallback in all three handlers (and gate the button on content readiness as the removed code did). *(Consensus: confirmed independently by both advisory legs.)*

### P1-3 — Edit button re-enabled for agent summaries → 404 on save (regression of #158/#161)
`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:2397`

The `detail.trigger_type !== TriggerType.AGENT` guard on the personal-report Edit button was removed:
```ts
{detail && detail.status === TaskStatus.COMPLETED && detail.permissions?.can_edit && !this.state.isEditing && (
    <Button ... onClick={this.handleStartEdit}> ... )}
```
The deleted comment explains why it was there: agent summaries have no `SummaryResult` row, so `PUT /edit` returns 404 "总结结果不存在". Failure scenario: agent summary with `can_edit=true` → user edits and saves → 404. Fix: restore the trigger-type guard, or confirm the backend now creates a `SummaryResult` for agent workflows (nothing in this PR does).

### P1-4 — Schedule button re-enabled for agent summaries → stalled/failed reruns (regression of #158/#161)
`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3032-3037` (guard deleted), rendered at `:3239`

The `if (detail?.trigger_type === TriggerType.AGENT) return null;` early-out in `renderScheduleButton()` was removed, and the button is now rendered in the header for all details. The deleted comment: agent summaries have no replayable sources/participants, so a scheduled rerun triggers the map-reduce pipeline and hangs in Pending or fails. Failure scenario: set a schedule on an agent summary → scheduled task deadlocks/fails. Fix: restore the guard.

### P1-5 — Delete/Leave inside the chat panel corrupts panel navigation
`packages/dmworksummary/src/components/ChatSummaryPanel.tsx:219`; `SummaryDetailPage.tsx:768`, `:781-782`

The panel renders `<SummaryDetailPage taskId={selectedTaskId} />` with **no** navigation callback. `handleDeleteTask` and `handleLeaveTask` unconditionally call `WKApp.routeRight.popToRoot()` — the global right-hand route, which is not what's driving the panel. Failure scenario: open a summary in the in-chat panel → delete it (or leave collaboration) → the panel stays mounted showing the now-deleted detail, its `view`/`selectedTaskId` unchanged, while `popToRoot()` pops an unrelated global route. The `summary-list-refresh-requested` emit refreshes the hidden list but never returns the panel to it. Fix: give `SummaryDetailPage` an optional `onBack`/`onDeleted` callback (mirroring `onViewDetail`/`onCreateNew` in `SummaryListPage`) and have the panel reset to `view:'list'`. *(Unique to the codex leg; independently confirmed.)*

### P1-6 — Existing test suite is broken; no tests updated
`packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx`

`ChatSummaryPanel` no longer renders `ChatSummaryHistory`, but the test file still mocks it and asserts on `data-testid="summary-history"`, the `paused` prop, and that "create-new" emits `wk:open-summary-modal` (`:60-140`). None of these hold after this PR (create is now in-panel `view:'create'`, not a mitt emit), and `SummaryListPage`/`SummaryCreatePage` are not mocked. This PR touches **zero** test files. Failure scenario: `vitest run` fails in CI. Fix: update `ChatSummaryPanel.test.tsx` to the new structure and add coverage for the new panel create/detail flows.

---

## Non-blocking (P2 / nits)

- **Leftover debug log.** `SummaryListPage.tsx:84` — `console.log('[LIST REFRESH] handleListRefreshRequested_ called');`. Remove before merge.
- **Missing i18n key.** `SummaryDetailPage.tsx:781` uses `summary.detail.deletedToast`, which is absent from both `en-US.json` and `zh-CN.json` (only `schedule.deleted` exists). Delete will render the raw key. *(Consensus: both this reviewer and codex.)*
- **Panel list is silently capped at 50.** `SummaryListPage.tsx:67` sets `pageSize: 50` in panel mode and `:516` suppresses pagination (`!isPanel && total > pageSize`). A channel with >50 summaries drops the rest with no UI affordance. Consider infinite-scroll or a "view all" link.
- **Fragile `setTimeout(800)` refresh.** `ChatSummaryPanel.tsx:handleCreateSubmit` waits a hardcoded 800ms before emitting `summary-list-refresh-requested`. If backend list-visibility lags, the new item won't appear; prefer optimistic insert or refetch-on-return. Also consider jumping straight to the created detail rather than back to the list (UX). *(gemini P2.)*
- **`embedded` ignores `visible`.** `ChatSummaryNewModal.tsx` render — in embedded mode the form renders unconditionally, ignoring the `visible` prop. Harmless with the current single caller (which unmounts), but a latent trap if someone toggles `visible={false}` instead of unmounting. *(gemini P2.)*
- **Orphaned component.** `ChatSummaryHistory.tsx` is now referenced only by its own tests and the (broken) panel test mock. Either delete it with its tests or note it's intentionally retained.
- **Unused import.** `ChannelTypeGroup` in `SummaryDetailPage.tsx:15` appears unused (`noUnusedLocals` is off so it won't break the build, but it's dead).
- **Hand-rolled space_id injection replaces the shared abstraction.** `handleForwardToChat` (`:1930-1940`) monkey-patches `msg.encodeJSON` inline instead of using `ForwardService` + `wrapSendContentForInjection`, which the ForwardService docs say exists precisely to avoid this (it also handles `mention.humans/ais` re-injection that the manual path omits). The person-only space_id logic matches the old behavior, so this isn't a proven bug, but it re-introduces the pattern that was deliberately extracted and unit-tested. Prefer reusing `ForwardService.send`.

---

## Also worth noting

- The PR currently reports **merge conflicts** (`mergeStateStatus: DIRTY`, `mergeable: CONFLICTING`) against `main` — resolve before merge, and re-verify the guard/logic areas after the rebase.
- The `/summary` proxy change (`apps/web/vite.config.ts:141`, removing the `path.replace(/^\/summary/, "")` rewrite) is **correct**: the summary service is mounted at `/summary/api/v1` and `summaryApi.ts:83` builds exactly that path, so preserving the prefix is right (the old strip was the bug). No change requested here.

## Coverage / blind spots
- **Backend enum for `origin_channel_type` (P1-1):** verified the FE is internally inconsistent (new code vs. `SourceType`/sibling modal/type-doc); did not read the summary service source to confirm which side is server-correct. Either way the two FE entry points must agree.
- **`detail.result === null` for agent summaries (P1-2/3/4):** inferred from in-repo comments and the `renderPersonalSummary` render path; not verified against a live backend response.
- No `node_modules` in the review checkout, so `tsc --noEmit` / `vitest` were not executed — findings are from static analysis and the diff. The test-suite breakage (P1-6) is by code inspection.
- Advisory legs run this round: **codex (line-level) succeeded; gemini-3.1-pro (long-context) succeeded via gateway fallback.** Both are advisory inputs only; this verdict is the reviewer's.

Jerry-Xin (dismissed):
Requesting changes at head 2dfa0a71, concurring with the two existing changes-requested reviews. Two independent blockers, both verified.

## 🔴 Gate — real merge conflict with main (DIRTY), 16 commits behind
Live state is `mergeable: false / mergeable_state: dirty`. The branch conflicts with already-merged summary work (#915 list/detail v2, #900 template limit) on the two most-churned files — `SummaryDetailPage.tsx` and `SummaryListPage.tsx`. GitHub blocks merge in this state, and a rebase will materially change exactly those files, so a full line-by-line deep review now would be largely invalidated after the rebase. Please rebase onto latest main, resolve the SummaryDetailPage / SummaryListPage conflicts against #915/#900, and re-request. (Also: the PR description is effectively empty — for a +1122/−420 unification touching List/Detail/Create + Panel/Modal/History, please document what changed and the behavior deltas so reviewers can distinguish intentional redesign from stale-base residue.)

## 🔴 origin_channel_type mapping is inverted vs the codebase contract (byte-verified)
`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:454-461`

The new panel-mode block hand-rolls the mapping:
```ts
// person→1, group→2, thread→3
if (ch.channelID.includes("____")) params.origin_channel_type = 3;      // thread
else if (ch.channelType === 1)     params.origin_channel_type = 1;      // person
else                               params.origin_channel_type = 2;      // group
```
But the canonical contract is the opposite. `SourceType` (`types/summary.ts:29-31`) is `GROUP_CHAT=1, THREAD=2, DIRECT_MESSAGE=3`, and the param doc (`types/summary.ts:267`) says `1=群聊 / 2=子区 / 3=私聊`. So the inline block sends DM→1 (should be 3), group→2 (should be 1), thread→3 (should be 2) — every panel-created summary is tagged with the wrong channel type, corrupting origin_channel_id reverse-lookup/filtering. This file already **imports** `getOriginChannelType` (line 19) but doesn't use it here. Fix: replace the inline block with the shared `getOriginChannelType(...)` so both entry points agree. (Please confirm the backend enum direction, but the FE internal inconsistency — new code vs SourceType vs the sibling ChatSummaryNewModal which uses `getSourceType` — is unambiguous.)

## Also blocking (verified by the peer reviews, agreeing — did not re-derive each byte)
The changes-requested reviews additionally flag, and I concur these must be addressed post-rebase:
- Forward-to-chat/Matter drops the `?? personalResult?.content` fallback → agent summaries silently no-op (regression of the #907 behavior; agent summaries populate `personal_result`, not `detail.result`).
- Edit button re-enabled for agent summaries (guard `trigger_type !== AGENT` removed) → `PUT /edit` 404 (regression of #158/#161).
- Schedule button re-enabled for agent summaries (early-return guard removed) → scheduled rerun hangs/fails (no replayable sources).
- Delete/Leave inside the in-chat panel calls `WKApp.routeRight.popToRoot()` with no panel `onBack`/`onDeleted` callback → panel stays mounted on a deleted detail and pops an unrelated global route.
- Test suite is broken and untouched: `ChatSummaryPanel.test.tsx` still mocks/asserts the removed `ChatSummaryHistory` + `wk:open-summary-modal` flow; zero test files updated → `vitest run` red.

## 🟡 Non-blocking
- `SummaryListPage.tsx:84` leftover `console.log('[LIST REFRESH] ...')` — remove.
- `summary.detail.deletedToast` i18n key missing from en-US/zh-CN → raw key rendered on delete.
- Panel list silently capped at pageSize 50 with pagination suppressed → >50 summaries dropped with no affordance.
- `setTimeout(800)` before the refresh emit is fragile; prefer optimistic insert / refetch-on-return.
- Hand-rolled `msg.encodeJSON` monkey-patch in `handleForwardToChat` re-introduces the pattern `ForwardService.wrapSendContentForInjection` was extracted to replace (and omits mention re-injection). Prefer reusing `ForwardService.send`.

## Notes / verified-good
- The `/summary` proxy change (`apps/web/vite.config.ts` — removing the `^/summary` rewrite) is correct: the service is mounted at `/summary/api/v1` and `summaryApi.ts` builds that exact path, so preserving the prefix is right. No change requested there.
- The panel-unification direction (reuse SummaryListPage/CreatePage/DetailPage in-panel) is sound; the issue is that the unification removed deliberately-added agent-summary guards without re-implementing the behavior they protected.

Verdict: **REQUEST_CHANGES**. Rebase onto main + fix the inverted `origin_channel_type` + restore the agent-summary guards (edit/schedule/forward) + fix panel-nav on delete/leave + update the test suite, then re-request.

yujiawei (changes requested):
# Code Review — PR #951 (octo-web) — re-review at updated head

Re-reviewed at head SHA `2dfa0a71c344a8a6456a0ee68256bea0707562ee` (the head moved from `5a0336a6` since the previous review; merge-base `2496dad1`, unchanged). A moved head does not inherit a prior verdict, so this is a fresh pass.

**Verdict: CHANGES_REQUESTED.** Good progress — most of the previous blockers were fixed by reverting the risky refactor — but two P1s remain and the test suite is still red.

## Resolved since the last review ✅
- **Forward-to-chat/matter now keep the `personalResult` fallback** and reuse `ForwardService` + `interpretForwardResult` again (`SummaryDetailPage.tsx:1920-1959`, `:1971`, `:1985`). Agent-summary forwarding works; the hand-rolled `encodeJSON` monkey-patch is gone. (was P1-2)
- **Edit button re-gated** with `detail.trigger_type !== TriggerType.AGENT` (`:2400`). (was P1-3)
- **Schedule button re-gated** with the `AGENT` early-return in `renderScheduleButton()` (`:3038`). (was P1-4)
- **`SHOW_FORWARD_TO_MATTER` back to `false`** (`:59`) — the incomplete Matter path is hidden again.
- **Debug `console.log` removed** from `SummaryListPage.handleListRefreshRequested_` (`:83`). (was a P2)
- New `emitSelection` prop scopes the `summary-detail-active/inactive` events to list-driven pushes only (`:56`, `:253`, `:291`, `:307`) — sensible fix for the highlight-leak risk.

## Still blocking (P1)

### P1-1 — `origin_channel_type` mapping is still inverted vs. the rest of the codebase
`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:454-462` (unchanged at this head)

The panel-mode block still hand-rolls `person→1, group→2, thread→3`, which is the inverse of the canonical `SourceType` enum (`types/summary.ts:28-32`: `GROUP_CHAT=1, THREAD=2, DIRECT_MESSAGE=3`) that the sibling `ChatSummaryNewModal` uses via `getSourceType` (`ChatSummaryNewModal.tsx:401`). This file even imports `getOriginChannelType` (`:19`) but doesn't call it here. Failure scenario: a summary created from a DM via the panel is tagged `origin_channel_type=1` (group), from a group → `2` (thread) — every panel-created summary gets the wrong channel type, corrupting `origin_channel_id` filtering/lookup. Fix: `params.origin_channel_type = getOriginChannelType(new Channel(ch.channelID, ch.channelType))`. Please confirm the true backend enum before fixing; regardless, the two FE entry points must not disagree.

### P1-5 — Delete/Leave inside the chat panel still corrupts panel navigation
`packages/dmworksummary/src/components/ChatSummaryPanel.tsx:219`; `SummaryDetailPage.tsx:787` (delete), `:771` (leave)

`ChatSummaryPanel` still renders `<SummaryDetailPage taskId={selectedTaskId} />` with no navigation callback, and `handleDeleteTask`/`handleLeaveTask` still call `WKApp.routeRight.popToRoot()` — the global right-hand route, which is not what drives the panel. Failure scenario: open a summary in the in-chat panel → delete it (or leave collaboration) → the panel stays mounted on the deleted detail (its `view`/`selectedTaskId` unchanged) while `popToRoot()` pops an unrelated global route. The `summary-list-refresh-requested` emit refreshes the hidden list but never returns the panel to it. Fix: give `SummaryDetailPage` an optional `onBack`/`onDeleted` callback (mirroring `onViewDetail` in `SummaryListPage`) and have the panel reset to `view:'list'` on delete/leave.

### P1-6 — Existing test suite is still broken; no tests updated
`packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx`

Still unchanged at this head: the test mocks the now-unused `ChatSummaryHistory` and asserts `data-testid="summary-history"`, the `paused` prop, and that "create-new" emits `wk:open-summary-modal` — none of which hold anymore (create is in-panel `view:'create'`; `SummaryListPage`/`SummaryCreatePage` aren't mocked). Failure scenario: `vitest run` fails in CI. Fix: update the panel test to the new structure and add coverage for the new create/detail flows.

## Non-blocking (P2 / nits)
- **Missing i18n key.** `SummaryDetailPage.tsx:786` still uses `summary.detail.deletedToast`, absent from both `en-US.json` and `zh-CN.json`. Delete-success renders the raw key.
- **Panel list silently capped at 50.** `SummaryListPage.tsx:67` (`pageSize: 50` in panel) + `:515` (`!isPanel && total > pageSize` suppresses pagination). A channel with >50 summaries drops the rest with no UI affordance.
- **`embedded` ignores `visible`** in `ChatSummaryNewModal` (latent trap; current caller unmounts, so harmless today).
- **Orphaned `ChatSummaryHistory.tsx`** — now referenced only by its own tests and the (broken) panel test mock. Delete or explicitly retain.

## Process note
- The PR still reports **merge conflicts** (`mergeable: CONFLICTING`) against `main` — resolve before merge and re-verify the guard/logic areas post-rebase.
- **Round count:** this PR has now been through 2 review rounds. The remaining P1s are the same class of issue each time (panel↔page integration seams: channel-type mapping, panel nav on delete, stale tests). If the next push doesn't clear P1-1/P1-5/P1-6, recommend pausing incremental fixes and doing one consolidated pass over the panel↔page boundary rather than another round.

Jerry-Xin (dismissed):
Re-review at head c5dba056. Some prior blockers are fixed, but two hard blockers remain, so this stays REQUEST_CHANGES.

## Fixed since the last review (verified) — thanks
- Forward-to-chat/Matter now restores the `?? personalResult?.content` fallback (`SummaryDetailPage.tsx:1934,1975,1989`), so agent summaries no longer silently no-op on forward.
- Edit button re-guarded for agent summaries: `... && detail.trigger_type !== TriggerType.AGENT` (`SummaryDetailPage.tsx:2405`).
- The in-chat panel now carries its own `view: 'list' | 'detail' | 'create'` state with back handlers resetting to `'list'` (`ChatSummaryPanel.tsx:132,136,140`), addressing the delete/leave panel-nav concern.

## 🔴 Still blocking

### 1. Gate — still merge-conflicting with main (DIRTY)
Live state is still `mergeable: false / mergeable_state: dirty`, and the branch is still **behind main by 7 commits** (diverged). GitHub blocks merge in this state, and the conflict is on the summary files this PR churns heavily. Please rebase onto latest main and resolve the conflicts before this can merge; the conflicted files will change on rebase, so the code review below should be re-confirmed after.

### 2. origin_channel_type mapping is still inverted (byte-verified, unchanged)
`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:449-456`

```ts
// person→1, group→2, thread→3
if (ch.channelID.includes("____")) params.origin_channel_type = 3;   // thread
else if (ch.channelType === 1)     params.origin_channel_type = 1;   // person
else                               params.origin_channel_type = 2;   // group
```

The canonical contract is the opposite — `SourceType` (`types/summary.ts`) is `GROUP_CHAT=1, THREAD=2, DIRECT_MESSAGE=3`, and the param doc says `1=群聊 / 2=子区 / 3=私聊`. So this still sends DM→1 (should be 3), group→2 (should be 1), thread→3 (should be 2), tagging every panel-created summary with the wrong channel type. The file already imports `getOriginChannelType` (line 19) but doesn't use it here. Fix: replace the inline block with the shared `getOriginChannelType(...)`.

### 3. Test suite still broken and untouched
`packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx:27-134`

The panel no longer renders `ChatSummaryHistory`, but the test still mocks it and asserts `data-testid="summary-history"`, the `paused` prop, and that create emits `wk:open-summary-modal` — none of which hold after this PR (create is now in-panel `view:'create'`). `vitest run` will fail. Update the panel tests to the new structure and add coverage for the in-panel create/detail flows.

## 🟡 Non-blocking
- Leftover debug log: `SummaryListPage.tsx:84` — `console.log('[LIST REFRESH] handleListRefreshRequested_ called')`. Remove before merge.
- (From prior review, please confirm) `summary.detail.deletedToast` i18n key present in both locales; panel list pageSize 50 cap; the `setTimeout(800)` refresh; the hand-rolled `msg.encodeJSON` monkey-patch vs `ForwardService`.

Verdict: **REQUEST_CHANGES** — rebase to clear the merge conflict + fix the inverted `origin_channel_type` (use `getOriginChannelType`) + update the broken panel test suite. The forward/edit/panel-nav fixes from this push are good; these three remain.

lml2468 (changes requested):
## Review Verdict（复审 @ c5dba056）：需修改（REQUEST_CHANGES）—— 两条阻塞均未解决

复审锚定 live head `c5dba056`。新 commit 加了东西但**上一轮两条阻塞都还在**:

### ❌ 阻断 1:仍 DIRTY,与 main 真实冲突、落后 16 commit(未 rebase)

- `mergeable=CONFLICTING / DIRTY`;3-way merge(exit=1)坐实 **4 文件冲突**:`index.css`、`SummaryCreatePage.tsx`、`SummaryDetailPage.tsx`、`SummaryListPage.tsx`;
- 仍落后 main **16 commit**,冲突撞已合入的 **#915/#900**。新推没有 rebase。DIRTY 下不可合。
- 修法不变:rebase 到最新 main + 解决这 4 个文件的冲突。

### ❌ 阻断 2:origin_channel_type 映射仍然反了(Steve/yujiawei/Jerry-Xin 均提,未修)

- `SummaryCreatePage.tsx:449-456` 仍手写:`channelID 含 "____" → 3`、`channelType===1(person) → 1`、`else → 2`。注释还写「1=私聊,2=群聊,3=子区」。
- 但全仓契约 `SourceType`(types/summary.ts:29-31)是 **GROUP_CHAT=1 / THREAD=2 / DIRECT_MESSAGE=3**。所以:
  - person/DM channel → 写成 `origin_channel_type=1`,后端按 **GROUP** 解读;
  - group → 写成 `2`,后端按 **THREAD**;
  - thread(含 "____")→ 写成 `3`,后端按 **DM**。
  **每条 panel 建的 summary,origin_channel_type 全错**,污染后端 origin 反查/过滤。
- 同一函数 :464 的 `source_type` 却用了**正确**的 SourceType 映射(group→GROUP_CHAT…),自相矛盾;
- 且文件 :19 已 `import { getOriginChannelType }` **却没用**——直接改成 `getOriginChannelType(ch)`(两入口共用一套映射)即可,这正是 #939 建立的正确 helper。

### 🟡 沿用 peer 共识(rebase 后一并处理)

- forward-to-chat/matter 丢 `?? personalResult?.content` 兜底 → agent summary no-op(#907 回归);
- edit/schedule 按钮去了 `trigger_type!==AGENT` 守卫 → agent summary PUT/edit 404、定时 hang(#158/#161 回归);
- 面板内 Delete/Leave 调 `WKApp.routeRight.popToRoot()` 但详情没传 onBack/onDeleted → 停在已删详情 + 误 pop 全局路由;
- 测试套件坏且零更新(ChatSummaryPanel.test 仍 mock 已移除的 ChatSummaryHistory);残留 console.log;删除缺 i18n。

### 结论

新 head 未 rebase(DIRTY 4 文件冲突照旧),且**核心数据正确性 🔴——origin_channel_type 反向映射——未修**(现成的 `getOriginChannelType` 没用上)。这两条任一都足以阻断。rebase + 用 `getOriginChannelType` 修映射 + 恢复 agent 守卫 + 修面板导航 + 更测试后复审。维持 **REQUEST_CHANGES**。

yujiawei (changes requested):
# Code Review — PR #951 (octo-web)

Reviewed at head SHA `c5dba056`, merge-base `2496dad1`. Scope: unifying the in-chat summary panel with the `/summary` workbench (list / create / detail reuse, restyled cards, detail-page action consolidation).

Overall this is a solid, well-structured consolidation — reusing `SummaryListPage` / `SummaryCreatePage` / `SummaryDetailPage` inside the panel is the right direction and removes a lot of duplicated UI. One blocking correctness bug and a couple of quality issues need addressing before merge.

---

## 1. Spec compliance

The diff implements what the title describes (panel ↔ workbench unification) and does not appear to add unrelated scope. The one concern here is a behavioral **deviation** rather than a missing/extra feature — see P0 below, where the new create path sends an `origin_channel_type` value that contradicts the established backend contract.

## 2. Blocking findings

### P0 — Inverted `origin_channel_type` mapping breaks source attribution
`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:449-456`

The new panel-create path hand-rolls the channel-type mapping:

```ts
// 后端要求 origin_channel_type: 1=私聊, 2=群聊, 3=子区
if (ch.channelID.includes("____")) {
    params.origin_channel_type = 3;           // thread
} else if (ch.channelType === 1) {
    params.origin_channel_type = 1;           // person -> 1
} else {
    params.origin_channel_type = 2;           // group  -> 2
}
```

This contradicts the rest of the codebase, which consistently uses the `SourceType` scale (`GROUP_CHAT=1, THREAD=2, DIRECT_MESSAGE=3`, `types/summary.ts:28-32`):

- `utils/channelType.ts:29-42` — the canonical `getOriginChannelType()` returns `getSourceType()`, i.e. group→1, thread→2, direct→3.
- `components/ChatSummaryNewModal.tsx:376,401` — the *existing* in-chat create path this PR is unifying sends `origin_channel_type: sourceType` (same `SourceType` scale).
- `types/summary.ts:267` — the doc comment on `origin_channel_type` states `1=群聊 / 2=子区 / 3=私聊` (group/thread/private).

The new code maps group→2, thread→3, private→1 — the inverse. Both paths POST to the same `/summaries` endpoint, so the same backend field now receives two different encodings depending on which entry point created the summary. Concrete failure: creating a summary from a **group** chat via the panel sends `origin_channel_type=2`, which the backend/contract reads as "thread". Downstream `origin_channel_id` reverse-lookup and the panel's own `origin_channel_id`-filtered list will mis-attribute or fail to find summaries.

Note the file already imports `getOriginChannelType` (`SummaryCreatePage.tsx:19`) but does not use it here. Replacing this block with `params.origin_channel_type = getOriginChannelType(ch)` fixes the bug and removes the now-unused import warning in one move. (Please double-check the thread detection: `getSourceType` uses `ChannelTypeCommunityTopic` / `parseThreadChannelId`, whereas this block uses a raw `"____"` substring check — the util is the tested source of truth.)

## 3. Non-blocking findings

### P2 — View is not re-synced when the panel is already open for the same channel
`packages/dmworksummary/src/components/ChatSummaryPanel.tsx:63-68`

`componentDidUpdate` only resets `view` from `props.summaryPanelView` when `channelID`/`channelType` change. If the panel is already open on a channel in `list` view and the user triggers the "new summary" affordance (`ChatSummaryStarButton` → `wk:toggle-summary-panel` with `summaryPanelView:'new'`, `forceOpen:true`), `showSummaryPanel` stays true and the channel is unchanged, so the panel does not switch to the create form. Recommend also reacting to `prevProps.summaryPanelView !== this.props.summaryPanelView`.

### P2 — Debug `console.log` left in production path
`packages/dmworksummary/src/pages/SummaryListPage.tsx:84`

```ts
console.log('[LIST REFRESH] handleListRefreshRequested_ called');
```

Fires on every summary-list refresh. Please remove.

### P2 — Fixed 800ms delay before list refresh after create
`packages/dmworksummary/src/components/ChatSummaryPanel.tsx:139-144`

`handleCreateSubmit` waits a hardcoded 800ms before emitting `summary-list-refresh-requested`. This is a guess at backend commit latency; under jitter the refresh can miss the just-created row, and when fast it adds needless delay. A short retry/poll, or refreshing on the `chat-summary-created` event the create page already dispatches, would be more robust. Not blocking.

### P2 — Orphaned embedded/`ChatSummaryHistory` code
- `ChatSummaryNewModal.tsx` gains an `embedded` prop and CSS (`ChatSummaryNewModal.css`), but the panel embeds `SummaryCreatePage`, not this modal — the `embedded` branch in the modal appears unused.
- `ChatSummaryHistory.tsx` is no longer referenced by any production file (only its own tests); its newly-added required `onClose` prop is also not supplied by those tests.

Neither breaks the build, but leaving dead/half-wired components invites confusion. Consider removing them or wiring them, in this PR or a fast-follow.

## 4. Notes checked and cleared (not blocking)

- **Stale list / stale `selectedChats` on chat switch.** `ChatContentPage` is mounted with `key={channel.getChannelKey()}` (`packages/dmworkbase/src/EndpointCommon.tsx:163-175`), so switching conversations remounts the whole panel subtree. `SummaryListPage`'s lack of a `componentDidUpdate` reload and `SummaryCreatePage`'s one-shot `selectedChats` initializer are therefore safe in the current mount model — no stale carry-over across channels. Worth a comment noting the reliance on the keyed remount, since a future refactor to a persistent instance would reintroduce the bug.
- **`summaryPanelView` reference in `Chat/index.tsx:1483`.** It is a declared field on `ChatContentPageState` (`:265`, read at `:1018`), not an undefined variable — no `ReferenceError`.
- **Removal of the AGENT guard on the schedule button** (`SummaryDetailPage.tsx`): agent-triggered summaries can now show the schedule button, while the regenerate/edit guards for AGENT are retained. If this is intentional (backend now supports scheduling agent summaries) it's fine; if not, it's a regression of the earlier #158/#161 fix. Please confirm.
- **`vite.config.ts` `/summary` rewrite removal**: matches the `/summary/api/v1` upstream mount; only affects local dev if a local backend serves without the prefix. Low risk.

## 5. Verdict

The panel unification is well done, but the inverted `origin_channel_type` encoding (P0) is a data-attribution correctness bug that will mis-tag summaries created from the panel and diverges from the existing create path hitting the same endpoint. Requesting changes for that one item; the P2s are cleanups that can ride along or fast-follow.

Jerry-Xin (changes requested):
Re-review at head b19cbdc7. The merge conflict is cleared this round and the debug log is gone — thanks. Two code blockers remain, so this stays REQUEST_CHANGES.

## Cleared since last review (verified)
- Merge state is now `mergeable: true` (the `blocked` state is the pending review gate, not a conflict) — the branch is rebased onto current main (4d9be56c). The authoritative PR delta is back to the ~16 summary files.
- The `console.log('[LIST REFRESH] ...')` debug line in `SummaryListPage.tsx` is removed.
- (Prior push already fixed) forward `?? personalResult?.content` fallback, the agent-summary edit guard, and the in-panel `view` back-navigation — still good.

## 🔴 Still blocking

### 1. origin_channel_type mapping is still inverted (byte-verified, unchanged)
`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:460-467`

```ts
// person→1, group→2, thread→3
if (ch.channelID.includes("____")) params.origin_channel_type = 3;   // thread
else if (ch.channelType === 1)     params.origin_channel_type = 1;   // person
else                               params.origin_channel_type = 2;   // group
```

Canonical contract is the opposite — `SourceType` (`types/summary.ts`) is `GROUP_CHAT=1, THREAD=2, DIRECT_MESSAGE=3`, and the param doc says `1=群聊 / 2=子区 / 3=私聊`. This still sends DM→1 (should be 3), group→2 (should be 1), thread→3 (should be 2), tagging every panel-created summary with the wrong channel type. The file already imports `getOriginChannelType` (line 19) but doesn't use it here. Fix: replace the inline block with the shared `getOriginChannelType(...)`.

### 2. Test suite still broken and untouched
`packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx:27-134`

The panel no longer renders `ChatSummaryHistory`, but the test still mocks it and asserts `data-testid="summary-history"`, the `paused` prop, and that create emits `wk:open-summary-modal` — none of which hold after this PR (create is now in-panel `view:'create'`). `vitest run` will fail. Update the panel tests to the new structure and add coverage for the in-panel create/detail flows.

Verdict: **REQUEST_CHANGES** — two remaining: fix the inverted `origin_channel_type` (use `getOriginChannelType`) and update the broken `ChatSummaryPanel.test.tsx`. The rebase + log cleanup + the earlier forward/edit/panel-nav fixes are good; these two are all that's left.

lml2468 (changes requested):
## Review Verdict（复审 @ b19cbdc7）：需修改（REQUEST_CHANGES）

复审锚定新 head `b19cbdc7`。上轮两条阻塞:**DIRTY 已解决**(仅落后 main 1 commit、`git merge-tree` 0 冲突、MERGEABLE;BLOCKED 来自 required-review 非冲突);但 **`origin_channel_type` 映射错误仍未修**,且逐字节核实其与同包既有入口的后端契约相反 —— 会给面板模式创建的总结写错来源类型。维持 REQUEST_CHANGES。

### 🔴 阻断:SummaryCreatePage 面板模式 `origin_channel_type` 映射与后端契约相反

- **契约来源(既有、未改、正在生产运行)**:同包 `ChatSummaryNewModal.tsx`(merge-base 即存在)发 `origin_channel_type: sourceType`,`sourceType` 来自 `getSourceType`/`SourceType`,即 **GROUP=1、THREAD=2、DM=3**(`types/summary.ts:28-32` `GROUP_CHAT:1 / THREAD:2 / DIRECT_MESSAGE:3`)。
- **本 PR 新增块(`SummaryCreatePage.tsx:460-468`,blame 确认为本 PR 引入)** 手写成另一套:
  - `channelID.includes("____")`(子区)→ `3`
  - `channelType === 1`(私聊/Person)→ `1`
  - else(群)→ `2`
  即 **DM=1、GROUP=2、THREAD=3** —— 与契约**逐位相反/错位**。
- **后果**:面板模式(`ChatSummaryPanel`→`SummaryCreatePage`,module.tsx:200/282 实际传 `channel`,该分支会执行)创建的总结,写入后端的来源类型全错:私聊被标成"群聊"、群聊被标成"子区"、子区被标成"私聊"。后端据此关联/展示来源,会导致来源类型错乱。
- **同文件已 import 正确助手却没用**:`SummaryCreatePage.tsx:19 import { getOriginChannelType }`,但 :460 这段手写没调用它。`getOriginChannelType(channel)` 正是 `getSourceType` 的 throw 版、与 modal 同一套映射。
- **附带**::460 注释 `// 后端要求 origin_channel_type: 1=私聊, 2=群聊, 3=子区` 本身就与 `SourceType`(1=群聊)矛盾 —— 注释也需一并纠正,否则会误导后来者再写错。

**触发**:在聊天面板里对私聊/群/子区任一发起总结 → 后端收到错位的 `origin_channel_type`。**引入者**:本 PR。**修法**:删掉 :460-468 手写块,直接
```ts
params.origin_channel_type = getOriginChannelType(ch);
```
(已 import;与 modal 入口统一,单一真相源),并删除/更正那条相反的注释。补一条断言 group/thread/dm 三态映射值的单测锁住契约。

### ✅ 已结清 / 非阻塞

- **DIRTY 上轮阻塞已解决**:落后 main 1 commit、0 文本冲突、MERGEABLE。
- 其余统一面板改动结构合理(面板复用 /summary workbench),非本条的部分未见新增阻塞。

### 结论

DIRTY 已修,但 `origin_channel_type` 面板映射与同包既有 modal 入口(生产中的后端契约)**相反**,是本 PR 引入的数据正确性 🔴 —— 会把面板创建的总结来源类型写错。文件已 import 正确的 `getOriginChannelType`,改用它并更正注释、补映射单测即可放行。**REQUEST_CHANGES**。

quanming1:
关闭此 PR，将重新提交一个干净的 PR。

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #951 (octo-web)

Reviewed at head `b19cbdc7616ee6cc40f60a5bd773667cff47a8fd`. Scope: chat summary panel architecture (panel-mode vs full-page), origin-channel filtering, schedule-button guard removal, vite proxy rewrite, and VoiceInputButton viewport clamping.

## Summary

This PR refactors the chat summary panel to embed `SummaryListPage` and `SummaryCreatePage` directly in-panel instead of routing through `ChatSummaryHistory` and modal popups. The architectural direction is sound — panel mode now reuses the same page components as full-page mode, reducing duplication. However, the channel filter for panel-mode list views is wired into only one of two data-fetch paths, so the embedded list shows summaries from all channels instead of the current chat. That's a P1 blocker for the panel-mode use case this PR introduces.

## Verification

Static analysis only at head `b19cbdc7`; build and tests not executed in this environment.

- ✅ **Proxy rewrite removal safe** — `apps/web/vite.config.ts` drops the `/summary` prefix rewrite; `summaryApi.ts` uses `BASE = '/summary/api/v1'`, so requests already carry the prefix. Aligned.
- ✅ **Origin channel metadata on create** — `SummaryCreatePage.handleSubmit` passes `origin_channel_id` and `origin_channel_type` (1/2/3 via `____` prefix detection). Correct mapping.
- ✅ **Regenerate guard preserved** — `showRegenerate` excludes agent-triggered summaries; `waitingForFallback` guard for forward-to-chat is kept in the dropdown.
- ✅ **Panel lifecycle** — `componentDidUpdate` resets to initial view (not always `'list'`) on channel switch, respecting `summaryPanelView` prop.

## Findings

One P1 blocker; two P2 items; one nit.

### P1 — Panel-mode list ignores channel filter (`packages/dmworksummary/src/pages/SummaryListPage.tsx:177`)

`fetchData()` correctly passes `origin_channel_id: channelId` to the API, but `loadData()` at line 185 does not. `componentDidMount()` and every event handler (`handlePageChange`, `handleSpaceChanged_`, `handleListRefreshRequested_`, etc.) call `loadData()`, not `fetchData()`. Result: in panel mode the list shows ALL summaries across all channels instead of filtering to the current chat. The delete handler was updated to call `fetchData()` directly, which suggests the split is unintentional.

Merge the channel filter into `loadData()` or have it delegate to `fetchData()`:

```ts
loadData() {
  this.fetchData(); // or inline the origin_channel_id param here
}
```

### P2 — Schedule button guard removed without confirmed backend enforcement (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3256`)

The old `TriggerType.AGENT` guard (`if (detail?.trigger_type === TriggerType.AGENT) return null;`) and its explanatory comment were removed. Agent-triggered summaries have no replay-able sources/participants for the traditional schedule pipeline. If the backend does not enforce `can_schedule: false` for agent summaries, users will see the schedule button and scheduled runs will fail silently or error. Verify backend enforcement or re-add the frontend guard.

### P2 — Test mocks target removed component (`packages/dmworksummary/src/components/ChatSummaryPanel.tsx:145`)

`ChatSummaryPanel` now imports `SummaryListPage` and `SummaryCreatePage` instead of `ChatSummaryHistory`, but `__tests__/ChatSummaryPanel.test.tsx` still mocks `../ChatSummaryHistory` (line 27). Tests pass against a phantom mock, not the real components. The `'emits open-summary-modal when creating a new summary'` test also asserts the old `wk:open-summary-modal` emit, which the new code replaces with `setState({ view: 'create' })`. Update mocks to target the new imports and assert the new behavior.

### Nit — Race-prone fixed delay on create refresh (`packages/dmworksummary/src/components/ChatSummaryPanel.tsx:151`)

`setTimeout(800)` in `handleCreateSubmit` is a best-guess delay before emitting the list-refresh event. On slow networks 800ms may not be enough; on fast ones it wastes time. `SummaryCreatePage.handleSubmit` already dispatches a `chat-summary-created` CustomEvent with the `taskId` — consider listening for that (or the API response callback) to trigger the refresh deterministically.

## Verdict: CHANGES_REQUESTED

The P1 channel-filter bug means panel mode — the primary UX this PR introduces — shows wrong data on initial load and every subsequent navigation event. Fix is one-line (route `loadData()` through `fetchData()` or add `origin_channel_id` to `loadData()`'s API call). The P2 items are non-blocking but should be addressed in this PR to avoid test rot and a dangling schedule-button guard.

yujiawei (changes requested):
# Code Review — PR #951 (octo-web)

Reviewed at head `b19cbdc7616ee6cc40f60a5bd773667cff47a8fd` against merge-base `4a98b034`. This report is in two independent gates: spec compliance first, then code quality. Overall verdict is an AND of the two.

## 1. Spec compliance

**Spec: ❌ (out-of-scope changes)**

The PR title/scope is "unify chat summary panel with /summary workbench" (embed `SummaryListPage`/`SummaryCreatePage` into `ChatSummaryPanel`). Two changes in the diff are unrelated to that scope:

- **Silent removal of the agent-summary schedule guard.** `packages/dmworksummary/src/pages/SummaryDetailPage.tsx` deletes `if (detail?.trigger_type === TriggerType.AGENT) return null;` from `renderScheduleButton()` (guard previously right after line 3070). The deleted comment explicitly documented why it existed: agent summaries have no replayable sources/participants, so a scheduled re-run gets stuck in Pending or fails. This is an unrelated behavioral change to the detail page, not part of unifying the panel, and it re-exposes a button that was deliberately hidden. Note the sibling `showRegenerate = ... && !isAgent` guard was kept, so the un-gating of schedule specifically looks deliberate but unjustified. If the backend now supports agent-summary scheduling, that needs to be stated and landed with its own change; if not, this is a regression. Please revert it from this PR or justify it with the backend support that makes it safe.
- **VoiceInputButton repositioning.** `packages/dmworkbase/src/Components/VoiceInputButton/index.tsx` adds viewport-clamping to the floating indicator (`FLOATING_WIDTH`, `maxLeft`, `transform: none`). Reasonable on its own, but unrelated to the summary panel.

The description is empty (only `## Summary`), which is what let these ride along unflagged. Please split the unrelated changes out or document them, and fill in the description (what changed, behavior changes) so the scope is auditable.

## 2. Code quality

### P1 (blocking)

**P1-1 — Existing unit test now fails deterministically (test/CI gate break).**
`ChatSummaryStarButton.tsx` was changed so `emitForHasSummaries()` always emits `wk:toggle-summary-panel` (with `summaryPanelView: hasSummaries ? 'history' : 'new'`), removing the old `wk:open-summary-modal` branch. The test at `packages/dmworksummary/src/components/__tests__/ChatSummaryStarButton.test.tsx:88` still asserts the old behavior and now fails:
```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'wk:open-summary-modal', … ]
-   "wk:open-summary-modal",
+   "wk:toggle-summary-panel",
```
The source changed but the test wasn't updated. Update the test to the new emit contract.

**P1-2 — Destructive actions lost their confirmation dialog.**
In `SummaryDetailPage.tsx`, Delete (`onClick={this.handleDeleteTask}`, ~line 3335) and Leave (`onClick={this.handleLeaveTask}`, ~line 3344) moved from `Popconfirm`-wrapped buttons to bare `Dropdown.Item` clicks. `handleDeleteTask` calls `api.deleteSummary` immediately — a single misclick in the overflow menu now permanently deletes a summary with no confirm and no undo. (The in-body Leave button at ~line 2638 and member-remove at ~line 2712 still keep their `Popconfirm`, so this is an inconsistency as well as a regression.) Restore a confirm step for the destructive dropdown items.

### P2 (advisory, non-blocking)

- **Dead code introduced by this PR.** `ChatSummaryPanel` switched from `ChatSummaryHistory` to `SummaryListPage`, so `ChatSummaryHistory.tsx` — including the new close-button and empty-state work this PR adds to it — has no production caller left. Similarly, the `embedded` branch this PR adds to `ChatSummaryNewModal.tsx` and the new `.chat-summary-modal-embedded` block in `ChatSummaryNewModal.css` are unused (the panel embeds `SummaryCreatePage`, not this modal). And after the StarButton change, nothing emits `wk:open-summary-modal`, so the listener registered in `module.tsx:271` is now dead. Adding new code to a component while making it unreachable in the same PR is confusing; either wire these up or drop them.
- **Responsive rule keys off viewport, not panel width.** `index.css` (~line 551) comments "面板宽度 ≤ 520px" but implements `@media (max-width: 520px)`, which is the viewport. A narrow panel on a wide desktop keeps its button labels and can overflow. Separately, the class that rule targets, `summary-detail-action-label`, no longer appears in any JSX (the header actions became icon-only `Dropdown.Item`s), so the rule is also orphaned. Use a container/panel-width mechanism, or remove it.
- **Post-delete "empty page" is treated as "no summaries."** `SummaryListPage.handleDeleteTask` (~line 311) checks `fresh.items.length === 0` on the current page/filter and, if empty, jumps to the create view and zeroes totals — even when earlier pages or a different filter still hold summaries. Also the `try` wraps both `deleteSummary` and the follow-up `fetchData`, so a transient failure of the refetch surfaces "delete failed" after a delete that actually succeeded. Reload from page 1 / clear filters before concluding "empty," and scope the catch to the delete call.
- **Hardcoded 800ms refresh with no cleanup.** `ChatSummaryPanel.handleCreateSubmit` (~line 140) fires `setTimeout(() => mittBus.emit("summary-list-refresh-requested"), 800)`. If backend indexing takes longer than 800ms the list refreshes with stale data; the timer isn't cleared on unmount. Prefer the event the create path already dispatches (`chat-summary-created`) or clear the timer in `componentWillUnmount`.
- **`summaryPanelView` only re-applies on channel change.** `ChatSummaryPanel.componentDidUpdate` resets the view only when `channel` changes, not when the `summaryPanelView` prop changes. For the `forceOpen`-while-already-open path, a new requested view is ignored. Minor, but worth handling if that path is exercised.
- **Panel list caps at 50 with pagination hidden.** In panel mode `pageSize` is 50 (`SummaryListPage.tsx:66`) and the pager is suppressed (`!isPanel && total > pageSize`, ~line 518). A chat with more than 50 summaries can't reach the rest from the panel. Acceptable as a scoped limitation, but call it out.

### Verified NOT a problem (advisor findings refuted on inspection)

- "Chat switch leaves the create form / list bound to the previous channel (missing `key`, `selectedChats` derived once, no reload on `channelId` change)." Refuted: `ChatContentPage` is rendered with `key={channel.getChannelKey()}` (`EndpointCommon.tsx:169`), and `getChannelKey()` returns `channelID-channelType` — so a channel switch remounts the entire panel subtree, and `SummaryCreatePage`'s constructor / `SummaryListPage`'s `componentDidMount(loadData)` re-run fresh. No stale-channel bug across switches.
- "`summaryPanelView` is undefined in `Chat/index.tsx` render → ReferenceError." Refuted: it is destructured from `this.state` (`Chat/index.tsx:1018`).

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ❌ (out-of-scope schedule-guard removal + unrelated voice-button change) AND Quality Changes-Requested (P1-1 failing test, P1-2 unconfirmed destructive delete/leave).

## 4. Suggested next steps

1. Update `ChatSummaryStarButton.test.tsx:88` to assert `wk:toggle-summary-panel` with the new `summaryPanelView` payload.
2. Restore a confirmation step for the Delete and Leave dropdown items.
3. Revert the agent-schedule guard removal from this PR (or land it separately with backend justification), and pull the VoiceInputButton change into its own PR.
4. Remove the now-dead code (`ChatSummaryHistory`, the `embedded` modal branch + CSS, the `wk:open-summary-modal` listener) or wire it up.
5. Fill in the PR description.

## 5. Coverage / blind spots

- Test suite: 473/475 unit tests pass. The one PR-caused failure is P1-1. The other two failures are not caused by this PR: `SummaryVersionHistory.test` (component untouched by this diff, pre-existing multiple-match assertion) and `ChatSummaryPanel.test` (environment-only — missing `@tiptap/react/jsx-runtime` in the local install, a collection error, not a logic failure). Typecheck could not be used as a signal: the local install has a broken `react` type-declaration resolution that cascades across node_modules and files untouched by this PR.
- Backend contract unverified from this repo: `origin_channel_id` filtering (`SummaryListPage` sends only `origin_channel_id`, not channel type; create sends `origin_channel_type`). If two channels of different types can share an ID string, the panel list could theoretically show the wrong chat's summaries — could not confirm backend semantics from the frontend repo. Worth a backend check.
- Advisor legs run this round: both. One line-level adversarial pass and one long-context pass both completed; their stale-channel and ReferenceError findings were refuted above, and their real findings are folded into P1/P2.

yujiawei (changes requested):
# Code Review — PR #951 (octo-web) — supplement (head b19cbdc7)

Verdict unchanged: **CHANGES_REQUESTED**. This supplements my earlier review with one additional confirmed P1 that the first pass surfaced and I verified against the source.

## New P1 — Panel-mode list ignores the channel filter on its primary load path

`SummaryListPage` has two fetch methods and only one carries the channel filter:

- `fetchData()` (`SummaryListPage.tsx:169`) includes `origin_channel_id: channelId`.
- `loadData()` (`SummaryListPage.tsx:185`) builds its own `params` and **omits** `origin_channel_id`.

The panel's primary path uses `loadData()`, not `fetchData()`: `componentDidMount()` calls `this.loadData()` (line 186), and so does every refresh/navigation handler — `handlePageChange` (207), status filter (299), keyword search (306), `handleSpaceChanged_`/`handleListRefreshRequested_`/`handleTaskRegenerated_` (80/82/84), retry (470). `fetchData()` is only ever called from the delete handler (line 315).

Result: in panel mode the list renders under the "summaries in this chat" title (`isPanel` → `summary.chatSummary.panelTitle`) but actually shows summaries from **all** channels — on initial open and after every refresh/page/filter event. `api.listSummaries` does no channel scoping when `origin_channel_id` is absent, so nothing else compensates. This breaks the exact panel-mode UX this PR introduces.

That only the delete handler was migrated to `fetchData()` suggests the split is unintentional. Fix: have `loadData()` include `origin_channel_id` (or delegate to `fetchData()` and set loading/error state around it).

## On the rest of the first-pass findings

Confirmed and already in my prior review: the schedule-button guard removal (agent summaries have no replayable sources), the stale `ChatSummaryPanel.test` mock of `../ChatSummaryHistory` + its `wk:open-summary-modal` assertion (line 27/131 — same class as the `ChatSummaryStarButton.test:88` failure I flagged), and the 800ms `setTimeout` refresh race. The proxy-rewrite removal is safe (`summaryApi` `BASE` already carries `/summary`) — agreed, non-issue.

PR #962, opened 2026-07-21
feat(loop): Agent Runtime UI — streaming client, verbose rendering, rich diff, checkpoints

ploy-elison:
## Agent Runtime UI (incremental)

Adds an incremental **Agent Runtime UI** layer to `@octo/loop` — the HTTP/SSE client
stack, event-stream reduction, and the verbose / rich-diff / checkpoint components
needed to drive an agent session from the web app. No existing files are modified;
everything is additive under `packages/dmloop/src/agentRuntime/` and
`packages/dmloop/src/api/agentRuntime/`, and the public package surface
(`src/index.tsx`) is unchanged.

> **Draft — not for merge.** Opened for review of the frontend contract. The
> diff / checkpoint / rollback endpoints and fields are still pending backend
> confirmation; wiring the top-level panel into a route and live end-to-end runs
> depend on those endpoints landing.

### What's here

**API layer (`src/api/agentRuntime`)**
- `httpClient` — fetch-based, injects `Authorization: Bearer <token>`, intercepts
  401 (structured error + `onUnauthorized` hook), and unwraps the `{ok,data,error}`
  envelope (with a lenient fallback for non-enveloped responses).
- `sseClient` — SSE **over POST** (agent turns need a request body, which native
  `EventSource` can't do): a pure, unit-tested frame parser (multi-line `data`,
  CRLF, half-packets, comments), exponential-backoff reconnect, and **explicit
  `Last-Event-ID`** injection on resume (fetch has no native replay semantics).
- `agentRuntimeApi` — `prompt` (streaming), `listSessions`, `getSession`,
  `getEntries`, `getState`, `abort`, plus `getDiff` / `listCheckpoints` /
  `rollback` (**pending backend confirmation**).

**Event stream**
- `streamReducer` — a pure function folding SSE events into
  `{messages, toolCalls, thinking, phase}`. Handles **out-of-order** events
  (ordered by `seq`), **reconnect de-duplication** (idempotent on repeated
  `seq`/`id`), and **interrupt/finalize** (late frames can't rewrite a terminal
  state).
- `useAgentStream` — wires the reducer to `sseClient`, exposes
  `send/abort/loadHistory`, and on replay failure falls back to
  `get_state + get_entries` reconciliation, surfacing a *possible event loss*
  banner.

**UI**
- `VerboseRenderer` — three levels: `off` (final text only), `on` (+ collapsed
  tool-call summaries), `full` (+ reasoning, raw input/output, raw event stream).
- `SessionList` — session listing + switch.
- `DiffView` — rich file diff honouring the `{path,changeType,hunks,binary,truncated}`
  contract: split / unified toggle, inline word-diff, and large-file / binary /
  truncated degradation. See note below on the diff library.
- `CheckpointTimeline` — checkpoint timeline with guarded (confirm-before) rollback.
- `AgentRuntimePanel` — top-level assembly with a **chat ⇄ code** mode toggle.

### Notes for reviewers

- **Diff rendering library.** The design specifies `react-diff-view` (default) with a
  Monaco `DiffEditor` for large files. Those deps aren't currently in the tree, so to
  keep this incremental PR buildable without expanding the dependency graph, the diff
  body uses a zero-dependency self-rendered view (word-level highlighting via the
  package's own `wordDiff`). The `FileDiff` contract, the split/unified interaction,
  and the large-file/binary/truncated branches are already in place, so dropping in
  `react-diff-view` / `@monaco-editor/react` later only replaces the `<FileDiffBody>`
  internals — props and interactions stay the same.
- **Pending-backend endpoints.** `diff`, `checkpoint` and `rollback` paths/fields are
  modelled from the frontend's expectation and parsed leniently (optional fields, no
  assumed required schema). They will be reconciled against the backend during
  integration.

### Testing

- 35 new unit tests: reducer (out-of-order / reconnect / interrupt / reconciliation),
  `httpClient` (Bearer / 401 / envelope / params), `sseClient` (frame parsing /
  reconnect / `Last-Event-ID`), and word diff.
- Full `@octo/loop` suite green: **107/107**.
- Type-checks clean (no new errors introduced in the package).
- Live end-to-end run against a running backend is deferred until the runtime
  endpoints are available.

### DIFF
--- a/packages/dmloop/src/agentRuntime/AgentRuntimePanel.tsx
+++ b/packages/dmloop/src/agentRuntime/AgentRuntimePanel.tsx
@@ -1,0 +1,179 @@
+// @octo/loop — Agent Runtime 面板（顶层装配）
+//
+// 把会话列表 / prompt 输入 / verbose 三档 / chat⇄code 一键切 / 富 diff / checkpoint
+// 时间线接到 useAgentStream 上。diff / checkpoint / rollback 相关端点「待后端确认」，
+// 这里按契约调用，失败降级为空态 + 错误提示，不阻断聊天主流程。
+
+import React, { useCallback, useEffect, useState } from "react";
+import { Send, Square, MessagesSquare, Code2, AlertTriangle } from "lucide-react";
+import { useAgentStream } from "./useAgentStream";
+import * as runtimeApi from "../api/agentRuntime/agentRuntimeApi";
+import type { AgentSessionSummary, FileDiff, Checkpoint } from "../api/agentRuntime/contracts";
+import SessionList from "./SessionList";
+import VerboseRenderer, { type VerboseLevel } from "./VerboseRenderer";
+import DiffView from "./DiffView";
+import CheckpointTimeline from "./CheckpointTimeline";
+import "./panel.css";
+
+type Mode = "chat" | "code";
+
+export interface AgentRuntimePanelProps {
+  agentId?: string;
+  initialSessionKey?: string;
+}
+
+export default function AgentRuntimePanel({ agentId, initialSessionKey }: AgentRuntimePanelProps) {
+  const [sessions, setSessions] = useState<AgentSessionSummary[]>([]);
+  const [sessionsLoading, setSessionsLoading] = useState(false);
+  const [activeKey, setActiveKey] = useState<string | undefined>(initialSessionKey);
+  const [verbose, setVerbose] = useState<VerboseLevel>("on");
+  const [mode, setMode] = useState<Mode>("chat");
+  const [draft, setDraft] = useState("");
+
+  const [diffs, setDiffs] = useState<FileDiff[]>([]);
+  const [diffLoading, setDiffLoading] = useState(false);
+  const [checkpoints, setCheckpoints] = useState<Checkpoint[]>([]);
+  const [ckptLoading, setCkptLoading] = useState(false);
+
+  const stream = useAgentStream({ sessionKey: activeKey, agentId });
+
+  // 会话列表。
+  useEffect(() => {
+    let alive = true;
+    setSessionsLoading(true);
+    runtimeApi
+      .listSessions({ agentId })
+      .then((rows) => alive && setSessions(rows))
+      .catch(() => alive && setSessions([]))
+      .finally(() => alive && setSessionsLoading(false));
+    return () => {
+      alive = false;
+    };
+  }, [agentId]);
+
+  const selectSession = useCallback(
+    (key: string) => {
+      setActiveKey(key);
+      void stream.loadHistory(key);
+    },
+    [stream],
+  );
+
+  // code 模式：拉 diff + checkpoint（端点待后端确认，失败静默降级）。
+  const refreshCode = useCallback(async () => {
+    if (!activeKey) return;
+    setDiffLoading(true);
+    setCkptLoading(true);
+    const [d, c] = await Promise.all([
+      runtimeApi.getDiff({ sessionKey: activeKey }).catch(() => [] as FileDiff[]),
+      runtimeApi.listCheckpoints(activeKey).catch(() => [] as Checkpoint[]),
+    ]);
+    setDiffs(d);
+    setCheckpoints(c);
+    setDiffLoading(false);
+    setCkptLoading(false);
+  }, [activeKey]);
+
+  useEffect(() => {
+    if (mode === "code") void refreshCode();
+  }, [mode, refreshCode]);
+
+  const onRollback = useCallback(
+    async (checkpointId: string) => {
+      if (!activeKey) return;
+      await runtimeApi.rollback(activeKey, checkpointId).catch(() => { /* 待后端确认：失败提示交由上层 */ });
+      await refreshCode();
+    },
+    [activeKey, refreshCode],
+  );
+
+  const submit = useCallback(() => {
+    const text = draft.trim();
+    if (!text || stream.running) return;
+    setDraft("");
+    stream.send(text);
+  }, [draft, stream]);
+
+  return (
+    <div className="loop-art-root">
+      <aside className="loop-art-sidebar">
+        <SessionList sessions={sessions} activeKey={activeKey} onSelect={selectSession} loading={sessionsLoading} />
+      </aside>
+
+      <main className="loop-art-main">
+        <header className="loop-art-toolbar">
+          {/* chat ⇄ code 一键切 */}
+          <div className="loop-art-modes">
+            <button type="button" className={`loop-art-mode ${mode === "chat" ? "active" : ""}`} onClick={() => setMode("chat")}>
+              <MessagesSquare size={14} /> Chat
+            </button>
+            <button type="button" className={`loop-art-mode ${mode === "code" ? "active" : ""}`} onClick={() => setMode("code")}>
+              <Code2 size={14} /> Code
+            </button>
+          </div>
+
+          {/* verbose 三档 */}
+          <div className="loop-art-verbose">
+            {(["off", "on", "full"] as VerboseLevel[]).map((lv) => (
+              <button key={lv} type="button" className={`loop-art-vbtn ${verbose === lv ? "active" : ""}`} onClick={() => setVerbose(lv)}>
+                {lv}
+              </button>
+            ))}
+          </div>
+
+          {/* abort 在跑 turn */}
+          {stream.running && (
+            <button type="button" className="loop-art-abort" onClick={stream.abort} title="Abort current turn">
+              <Square size={13} /> Stop
+            </button>
+          )}
+        </header>
+
+        {/* 回放不可用降级：可能有事件缺失 */}
+        {stream.state.maybeMissingEvents && (
+          <div className="loop-art-warn">
+            <AlertTriangle size={14} /> Some events may be missing (stream replay unavailable — reconciled from server state).
+          </div>
+        )}
+        {stream.error && <div className="loop-art-error">{stream.error}</div>}
+
+        <div className="loop-art-content">
+          {mode === "chat" ? (
+            <VerboseRenderer state={stream.state} level={verbose} />
+          ) : (
+            <div className="loop-art-code">
+              <section className="loop-art-diff">
+                <h4>Changes</h4>
+                {diffLoading ? <div className="loop-art-hint">Loading diff…</div> : <DiffView diffs={diffs} />}
+              </section>
+              <section className="loop-art-ckpt">
+                <h4>Checkpoints</h4>
+                <CheckpointTimeline checkpoints={checkpoints} loading={ckptLoading} onRollback={onRollback} />
+              </section>
+            </div>
+          )}
+        </div>
+
+        {mode === "chat" && (
+          <footer className="loop-art-composer">
+            <textarea
+              className="loop-art-input"
+              value={draft}
+              placeholder="Send a prompt…"
+              onChange={(e) => setDraft(e.target.value)}
+              onKeyDown={(e) => {
+                if (e.key === "Enter" && (e.metaKey || e.ctrlKey)) {
+                  e.preventDefault();
+                  submit();
+                }
+              }}
+            />
+            <button type="button" className="loop-art-send" onClick={submit} disabled={!draft.trim() || stream.running}>
+              <Send size={14} /> Send
+            </button>
+          </footer>
+        )}
+      </main>
+    </div>
+  );
+}
--- a/packages/dmloop/src/agentRuntime/CheckpointTimeline.tsx
+++ b/packages/dmloop/src/agentRuntime/CheckpointTimeline.tsx
@@ -1,0 +1,72 @@
+// @octo/loop — checkpoint 时间线 + 回滚
+//
+// ⚠️ checkpoint / rollback 端点与字段「待后端确认」（见 agentRuntimeApi）。本组件仅按
+// 前端预期契约渲染时间线并暴露回滚入口；回滚是破坏性操作，点按前需二次确认。
+
+import React, { useState } from "react";
+import { GitCommitHorizontal, RotateCcw, Loader2 } from "lucide-react";
+import type { Checkpoint } from "../api/agentRuntime/contracts";
+import "./checkpoint.css";
+
+export interface CheckpointTimelineProps {
+  checkpoints: Checkpoint[];
+  loading?: boolean;
+  // 回滚回调（由上层调 rollback 端点）。返回 Promise 以驱动行内 loading。
+  onRollback: (checkpointId: string) => Promise<void> | void;
+}
+
+export default function CheckpointTimeline({ checkpoints, loading, onRollback }: CheckpointTimelineProps) {
+  const [busyId, setBusyId] = useState<string | null>(null);
+  const [confirmId, setConfirmId] = useState<string | null>(null);
+
+  const doRollback = async (id: string) => {
+    setBusyId(id);
+    try {
+      await onRollback(id);
+    } finally {
+      setBusyId(null);
+      setConfirmId(null);
+    }
+  };
+
+  if (loading) {
+    return (
+      <div className="loop-ckpt-loading">
+        <Loader2 size={14} className="loop-spin" /> Loading checkpoints…
+      </div>
+    );
+  }
+  if (!checkpoints.length) return <div className="loop-ckpt-empty">No checkpoints</div>;
+
+  return (
+    <div className="loop-ckpt-timeline">
+      {checkpoints.map((c) => (
+        <div className="loop-ckpt-node" key={c.id}>
+          <div className="loop-ckpt-rail">
+            <GitCommitHorizontal size={16} className="loop-ckpt-dot" />
+          </div>
+          <div className="loop-ckpt-body">
+            <div className="loop-ckpt-label">{c.label || `Checkpoint ${c.seq ?? ""}`}</div>
+            <div className="loop-ckpt-meta">
+              {c.created_at && <span>{c.created_at}</span>}
+              {typeof c.files_changed === "number" && <span>{c.files_changed} files</span>}
+            </div>
+          </div>
+          {confirmId === c.id ? (
+            <div className="loop-ckpt-confirm">
+              <span>Roll back to here?</span>
+              <button type="button" className="loop-ckpt-btn danger" disabled={busyId === c.id} onClick={() => doRollback(c.id)}>
+                {busyId === c.id ? <Loader2 size={12} className="loop-spin" /> : "Confirm"}
+              </button>
+              <button type="button" className="loop-ckpt-btn" onClick={() => setConfirmId(null)}>Cancel</button>
+            </div>
+          ) : (
+            <button type="button" className="loop-ckpt-btn" onClick={() => setConfirmId(c.id)} title="Roll back to this checkpoint">
+              <RotateCcw size={13} /> Rollback
+            </button>
+          )}
+        </div>
+      ))}
+    </div>
+  );
+}
--- a/packages/dmloop/src/agentRuntime/DiffView.tsx
+++ b/packages/dmloop/src/agentRuntime/DiffView.tsx
@@ -1,0 +1,204 @@
+// @octo/loop — 富 diff 视图
+//
+// 按研发设计文档的 diff JSON 契约 {path,changeType,hunks,binary,truncated} 渲染，
+// 支持 split（并排）/ unified（单栏）切换 + 行内 word diff 高亮。
+//
+// 关于库选型：设计文档指定「react-diff-view（默认）+ 大文件切 Monaco DiffEditor」。
+// 当前 dmloop 未引入这两个依赖（package.json 仅有 lucide-react/react-markdown 等），
+// 为保持增量 PR 可构建、不擅自扩容依赖树，这里先用零依赖的自渲染实现（词级高亮走
+// 本包 wordDiff）。已按契约把「大文件 / 二进制 / 截断」的降级分支与渲染 seam 留好：
+// 后续引入 react-diff-view / @monaco-editor/react 时，仅替换 <FileDiffBody> 内部渲染，
+// 上层 props（FileDiff 契约）与交互（split/unified、阈值切换）不变。
+
+import React, { useMemo, useState } from "react";
+import { FileCode, FilePlus, FileMinus, FileText, Columns2, AlignJustify } from "lucide-react";
+import type { FileDiff, DiffHunkLine } from "../api/agentRuntime/contracts";
+import { wordDiff } from "./wordDiff";
+import "./diffView.css";
+
+// 超过该行数视为「大文件」，提示切换到重型编辑器（Monaco，待依赖引入）。
+const LARGE_FILE_LINES = 800;
+
+type ViewMode = "split" | "unified";
+
+function changeIcon(t: FileDiff["changeType"]) {
+  if (t === "added") return <FilePlus size={14} className="loop-diff-ic add" />;
+  if (t === "deleted") return <FileMinus size={14} className="loop-diff-ic del" />;
+  if (t === "renamed" || t === "copied") return <FileText size={14} className="loop-diff-ic mod" />;
+  return <FileCode size={14} className="loop-diff-ic mod" />;
+}
+
+// 把一对 delete/insert 行做词级 diff，渲染成带高亮 span 的行内内容。
+function InlineWord({ oldLine, newLine, side }: { oldLine: string; newLine: string; side: "old" | "new" }) {
+  const segs = useMemo(() => wordDiff(oldLine, newLine), [oldLine, newLine]);
+  return (
+    <>
+      {segs
+        .filter((s) => (side === "old" ? s.kind !== "insert" : s.kind !== "delete"))
+        .map((s, i) => (
+          <span
+            key={i}
+            className={s.kind === "equal" ? undefined : side === "old" ? "loop-diff-word-del" : "loop-diff-word-ins"}
+          >
+            {s.value}
+          </span>
+        ))}
+    </>
+  );
+}
+
+// 把 hunk 行配成 unified 或 split 的可渲染行。
+interface PairedRow {
+  oldLine?: DiffHunkLine;
+  newLine?: DiffHunkLine;
+}
+
+// 在一段连续变更块内，按顺序把 delete 与 insert 配对（用于 split 并排 + 行内 word diff）。
+function pairHunkLines(lines: DiffHunkLine[]): PairedRow[] {
+  const rows: PairedRow[] = [];
+  let i = 0;
+  while (i < lines.length) {
+    const line = lines[i];
+    if (line.type === "normal") {
+      rows.push({ oldLine: line, newLine: line });
+      i++;
+      continue;
+    }
+    // 收集连续的 delete 块与其后连续的 insert 块。
+    const dels: DiffHunkLine[] = [];
+    const ins: DiffHunkLine[] = [];
+    while (i < lines.length && lines[i].type === "delete") dels.push(lines[i++]);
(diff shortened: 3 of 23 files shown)


boris-clark:
此功能改为独立 agent 项目、托管到 codex.mlamp.cn 的 monorepo，不在 octo-web 内实现，代码迁移至独立仓库。关闭本 PR。

ploy-elison:
改为独立 agent 项目、托管 codex.mlamp.cn monorepo，不在 octo-web 内实现，代码迁独立仓，关闭本 PR。

Jerry-Xin (approved):
Approving at head 9fb0d3af. A large but well-structured, genuinely-tested new module (all additions, self-contained under `dmloop/agentRuntime` + `api/agentRuntime`). No blockers found across the API layer or the render/reducer/diff layer.

## API layer — verified
- **SSE client (`sseClient.ts`)**: SSE-over-POST via `fetch` + `ReadableStream` (native EventSource is GET-only). Frame parsing is a pure `SseFrameParser` handling half-packets, CRLF/CR/LF normalization, comment/blank-line skip, per-spec single leading-space strip, NaN-guarded `retry`, and unknown-field ignore. Reconnect is bounded exponential backoff (`min(base * 2**attempt, maxBackoffMs)`, default maxRetries 5) with explicit `Last-Event-ID` re-injection (fetch has no native resume). Uses `AbortController` + an `aborted` flag; `close()` sets `aborted=true` and calls `controller.abort()`, and `AbortError`/`aborted` are guarded before any reconnect — so the `useAgentStream` unmount → `conn.close()` path cleanly tears down the stream (confirmed the cross-layer contract).
- **httpClient (`httpClient.ts`)**: `buildAuthHeaders` reads the token from an injectable `_tokenProvider` (default null; host injects at bootstrap, reusing login state) and puts it only in the `Authorization: Bearer` header — never in URL/query/body/logs. No secret leakage. `{ok,data,error}` envelope; base URL from `VITE_AGENT_RUNTIME_BASE`.
- **contracts (`contracts.ts`)**: event/phase/diff shapes are well-typed with optional fields the reducer/DiffView guard against.

## Render / reducer / diff layer — verified
- **No XSS**: zero `dangerouslySetInnerHTML`/`innerHTML`. All server-provided content (message/thinking text, tool input/output, raw event JSON, diff line content) is rendered as JSX string children (React auto-escapes) or `JSON.stringify` into `<pre>`.
- **streamReducer is pure & immutable**: returns new objects, copies arrays, clones the `seen` Set — no in-place React-state mutation. Dedup by `seq` (else `id`) is idempotent on reconnect replay; out-of-order events are insert-by-seq; `tool_result` before `tool_call` handled via placeholder; unknown event types fall through safely with no throw. Solid boundary handling (out-of-order / reconnect / abort), all covered by real assertions in `streamReducer.test.ts`.
- **wordDiff**: standard LCS backtrack; tokenizer preserves separators and reassembly is lossless (asserted); empty-input edges correct; no token drop/dup.
- **useAgentStream**: unmount sets `mountedRef=false` + `connRef.current?.close()`; all setState goes through a `safeSet` that no-ops when unmounted, and async callbacks guard on `mountedRef`. `abort`/`reset`/`send`/`loadHistory` each close the prior connection before reassigning — no dangling connections, no setState-after-unmount.
- CSS imports all resolve; optional-field access is guarded; no obvious type error (noting CI's vite build doesn't run `tsc`).

## Tests
- `streamReducer.test.ts` and `wordDiff.test.ts` are genuine behavioral suites (delta concat, tool call/result → done, out-of-order ordering, dedup idempotency, abort locks late frames, lossless tokenize, changed-word isolation) — not smoke tests. `sseClient.test.ts` / `httpClient.test.ts` cover the frame-parse edges and auth/envelope.

## 🟡 Non-blocking (worth a fast-follow)
- **`wordDiff.ts` LCS table is O(n·m) time AND memory with no input-size guard**, and `DiffView` calls it per changed line unguarded. A pathological long/minified line (e.g. ~50k tokens each side) would allocate billions of cells and hang the tab. Add a token-count threshold that falls back to whole-line highlight when `a.length * b.length` exceeds a cap.
- **`streamReducer` finalized-lock is asymmetric**: after `done`/`aborted`, `message_delta` and `phase` are locked, but a late full `message`, `tool_call`, or `tool_result` can still mutate `messages`/`toolCalls`. Intentional for late `tool_result` per the header comment, but a late full `message` post-abort could rewrite assistant text — document or extend the guard. Also the `messages`/`toolCalls`/`thinking` arrays and `seen` Set are uncapped (bounded by event count; per-session `reset()` limits real blast radius) — consider a cap for very long turns.

## CI
- secret-scan / dependency-review / pr-title-lint pass. `osv-scan` pending/infra. `mergeable: true` (blocked = pending review gate).

Verdict: **APPROVE**. Clean, secret-safe, well-tested Agent Runtime UI module; the two 🟡 items (wordDiff size guard, reducer finalized-lock/cap) are good fast-follows, not blockers.

Jerry-Xin (approved):
Approving at head 9fb0d3af. A large but well-structured, genuinely-tested new module (all additions, self-contained under `dmloop/agentRuntime` + `api/agentRuntime`). No blockers found across the API layer or the render/reducer/diff layer.

## API layer — verified
- **SSE client (`sseClient.ts`)**: SSE-over-POST via `fetch` + `ReadableStream` (native EventSource is GET-only). Frame parsing is a pure `SseFrameParser` handling half-packets, CRLF/CR/LF normalization, comment/blank-line skip, per-spec single leading-space strip, NaN-guarded `retry`, and unknown-field ignore. Reconnect is bounded exponential backoff (`min(base * 2**attempt, maxBackoffMs)`, default maxRetries 5) with explicit `Last-Event-ID` re-injection (fetch has no native resume). Uses `AbortController` + an `aborted` flag; `close()` sets `aborted=true` and calls `controller.abort()`, and `AbortError`/`aborted` are guarded before any reconnect — so the `useAgentStream` unmount → `conn.close()` path cleanly tears down the stream (confirmed the cross-layer contract).
- **httpClient (`httpClient.ts`)**: `buildAuthHeaders` reads the token from an injectable `_tokenProvider` (default null; host injects at bootstrap, reusing login state) and puts it only in the `Authorization: Bearer` header — never in URL/query/body/logs. No secret leakage. `{ok,data,error}` envelope; base URL from `VITE_AGENT_RUNTIME_BASE`.
- **contracts (`contracts.ts`)**: event/phase/diff shapes are well-typed with optional fields the reducer/DiffView guard against.

## Render / reducer / diff layer — verified
- **No XSS**: zero `dangerouslySetInnerHTML`/`innerHTML`. All server-provided content (message/thinking text, tool input/output, raw event JSON, diff line content) is rendered as JSX string children (React auto-escapes) or `JSON.stringify` into `<pre>`.
- **streamReducer is pure & immutable**: returns new objects, copies arrays, clones the `seen` Set — no in-place React-state mutation. Dedup by `seq` (else `id`) is idempotent on reconnect replay; out-of-order events are insert-by-seq; `tool_result` before `tool_call` handled via placeholder; unknown event types fall through safely with no throw. Solid boundary handling (out-of-order / reconnect / abort), all covered by real assertions in `streamReducer.test.ts`.
- **wordDiff**: standard LCS backtrack; tokenizer preserves separators and reassembly is lossless (asserted); empty-input edges correct; no token drop/dup.
- **useAgentStream**: unmount sets `mountedRef=false` + `connRef.current?.close()`; all setState goes through a `safeSet` that no-ops when unmounted, and async callbacks guard on `mountedRef`. `abort`/`reset`/`send`/`loadHistory` each close the prior connection before reassigning — no dangling connections, no setState-after-unmount.
- CSS imports all resolve; optional-field access is guarded; no obvious type error (noting CI's vite build doesn't run `tsc`).

## Tests
- `streamReducer.test.ts` and `wordDiff.test.ts` are genuine behavioral suites (delta concat, tool call/result → done, out-of-order ordering, dedup idempotency, abort locks late frames, lossless tokenize, changed-word isolation) — not smoke tests. `sseClient.test.ts` / `httpClient.test.ts` cover the frame-parse edges and auth/envelope.

## 🟡 Non-blocking (worth a fast-follow)
- **`wordDiff.ts` LCS table is O(n·m) time AND memory with no input-size guard**, and `DiffView` calls it per changed line unguarded. A pathological long/minified line (e.g. ~50k tokens each side) would allocate billions of cells and hang the tab. Add a token-count threshold that falls back to whole-line highlight when `a.length * b.length` exceeds a cap.
- **`streamReducer` finalized-lock is asymmetric**: after `done`/`aborted`, `message_delta` and `phase` are locked, but a late full `message`, `tool_call`, or `tool_result` can still mutate `messages`/`toolCalls`. Intentional for late `tool_result` per the header comment, but a late full `message` post-abort could rewrite the completed reply text — document or extend the guard. Also the `messages`/`toolCalls`/`thinking` arrays and `seen` Set are uncapped (bounded by event count; per-session `reset()` limits real blast radius) — consider a cap for very long turns.

## CI
- secret-scan / dependency-review / pr-title-lint pass. `osv-scan` pending/infra. `mergeable: true` (blocked = pending review gate).

Verdict: **APPROVE**. Clean, secret-safe, well-tested Agent Runtime UI module; the two 🟡 items (wordDiff size guard, reducer finalized-lock/cap) are good fast-follows, not blockers.

lml2468 (commented):
## Review Verdict：评审意见（COMMENT）—— 前端契约层扎实,按作者「Draft, not for merge」定位给意见而非放行/阻断

审查锚定 head `9fb0d3af`。23 文件 +2406/−0,全新增于 `agentRuntime/` + `api/agentRuntime/`,不改任何既有文件、public surface(`index.tsx`)不变、未接入任何路由。作者明确标注 **「Draft — not for merge」**、diff/checkpoint/rollback 端点待后端确认。因此给 **COMMENT**:不 APPROVE(作者未申请合并、端点未定),也不 REQUEST_CHANGES(无 breakage、完全隔离、测试齐全)。本地跑全套 + 核过安全/纯函数/规范三面。

### ✅ 亮点(逐条核过)

- **seam 纪律干净**:`api/agentRuntime/*` **0 处 import WKApp**;token 走 `setAuthTokenProvider` 注入(默认 null,宿主启动注入复用登录态),401 走 `setUnauthorizedHandler` 钩子 —— 符合 DEVELOPMENT.md「Service 层不 import WKApp、运行时上下文由调用方注入」。
- **httpClient 安全面到位**:Bearer 注入、401 先触发钩子再抛结构化 `AgentRuntimeError`、`{ok,data,error}` 信封解包 + 非信封兜底、`AbortError` 保留 name、params 剔除 undefined/null/空串并 `encodeURIComponent`。
- **SSE-over-POST**:token/`Last-Event-ID` 都走 **header**(`buildAuthHeaders`),不泄漏进 URL;指数退避重连 + 显式续推。
- **streamReducer 纯函数正确**:`seq` 升序插入(乱序保护)、`seen` Set 幂等去重(重连不重复)、`finalized` 终态门控 —— late `message_delta`(:98)与 phase 迁移(:120/135/155/180)在终态后不可改写。
- **rollback 有确认门控**:CheckpointTimeline 两步 `confirmId` 确认,破坏性操作非一键。
- **验证**:`@octo/loop` 全套 **107/107 绿**(reducer 乱序/重连/中断/对账、httpClient Bearer/401/信封、sseClient 帧解析/重连/Last-Event-ID、wordDiff);behind 1 无关 commit、`git merge-tree` 0 冲突。

### 🟡 合前(端点落地、正式转 mergeable 时)需处理

1. **CSS 语义 token 带硬编码回退**:`panel.css`/`diffView.css`/`verbose.css` 用 `var(--semi-color-*, #fff7e6/#52c41a/...)` 形式,回退是硬编码 hex;`panel.css:38` 主按钮 `color:#fff` + `border-radius:6px` 是**纯硬编码无 token**。DEVELOPMENT.md 要求语义 token,建议去掉 hex 回退、按钮色/圆角也走 token(`--semi-color-white`/`--wk-r-*`)。
2. **`credentials:"include"` + `Authorization: Bearer` 并存**:同源(默认 `/fleet/api/v1`)无碍;但 `VITE_AGENT_RUNTIME_BASE` 可配跨域,届时同时发 cookie + bearer 需确认后端 CORS/鉴权预期,避免歧义。
3. **待后端确认的端点(作者已声明)**:diff/checkpoint/rollback 的 path/字段目前按前端预期宽松解析(可选字段、不假设必填 schema)。合前必须与后端对齐真实契约 —— 这也是本 PR 目前只能 COMMENT、不能放行的核心原因。
4. **diff 库占位**:当前用零依赖自渲染 diff(word-level),设计目标是 `react-diff-view` + Monaco。契约/交互/降级分支已就位,后续替换只动 `<FileDiffBody>` 内部,props 不变 —— 方向合理,注明即可。

### 结论

前端契约层设计扎实:seam 解耦、鉴权/401/信封/中止处理完整、纯 reducer 的乱序/去重/终态语义正确、SSE 无 token 泄漏、rollback 有确认、测试 107/107。完全隔离、零改既有文件,作为增量前端契约评审**质量很高**。按作者「Draft — not for merge」+ 端点待后端确认的定位,给 **COMMENT**:合前处理 CSS token 硬编码回退、确认跨域鉴权预期、并与后端对齐 diff/checkpoint/rollback 真实契约,即可转正式 APPROVE。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #962 (octo-web)

## Summary

This PR introduces the Agent Runtime UI — a new streaming client, event reducer, verbose renderer, rich diff view, and checkpoint timeline for the `dmloop` package, plus a batch of skill-market UI refinements (icon upload a11y, install-button scoping, search clear button, creator/owner label logic). The agent-runtime work is entirely additive (new files under `packages/dmloop/src/agentRuntime/` and `packages/dmloop/src/api/agentRuntime/`), cleanly isolated from existing loop modules, and ships with focused unit tests for the pure-function core (SSE parser, stream reducer, word diff, HTTP client).

The architecture is sound: SSE-over-POST with manual frame parsing, idempotent dedup via `seq`/`id`, reconnect with `Last-Event-ID`, and a reconciliation fallback (`get_state` + `get_entries`) when SSE replay is unavailable. No security, correctness, or build-blocking issues found; four P2 items below address robustness edge cases.

## Verification

- ✅ **Diff scope** — 36 files, +2972/-181. Agent-runtime is all-new files; skill-market changes are localized UI refinements.
- ✅ **No XSS vectors** — no `dangerouslySetInnerHTML` or `innerHTML` in new code. All user-controlled text rendered via JSX auto-escaping.
- ✅ **Auth parity** — `buildAuthHeaders()` used by both `httpClient` (REST) and `sseClient` (SSE). Bearer token injected consistently; 401 handler registered via `bootstrap.ts`.
- ✅ **Cursor/Last-Event-ID flow** — traced from `connectSse` `lastId` variable → `runOnce` header injection → `onFrame` update → reconnect. Tests confirm `Last-Event-ID: 5` on retry.
- ✅ **Dedup correctness** — `streamReducer.eventKey()` + `seen` Set handles seq-based and id-based dedup. Tests cover out-of-order, duplicate replay, and `tool_result` before `tool_call`.
- ✅ **Return wiring** — `connectSse` → `prompt()` → `useAgentStream` → `AgentRuntimePanel` all consume the returned contracts end-to-end.
- ✅ **Test coverage** — unit tests cover SSE frame parser (half-packet, CRLF, multi-line data, comments, flush), HTTP client (envelope unwrap, 401, query filtering), stream reducer (ordering, dedup, abort, reconciliation), and word diff (tokenize roundtrip, LCS).

Static analysis only at head `9fb0d3afd`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Four P2 items below, all addressing robustness edge cases in the streaming infrastructure.

### P2 — Stale `running` flag race on rapid re-send (`packages/dmloop/src/agentRuntime/useAgentStream.ts:117-130`)

When `send()` is called while a previous stream is active, `connRef.current?.close()` aborts the old connection. However, the old `conn.done` promise chain still executes its `.finally(() => setRunning(false))`. Because `send()` synchronously calls `setRunning(true)` and starts a new connection, the old `.finally()` — scheduled as a microtask after the abort resolves — can fire after the new `setRunning(true)`, briefly setting `running` to `false` while the new stream is active.

In practice, the window is narrow (microtask gap), but it could let `submit()` in `AgentRuntimePanel.tsx:105` fire a third overlapping stream if the user clicks rapidly.

Fix: guard the `.finally()` against stale connections, e.g.:

```ts
const thisConn = conn;
conn.done
  .catch(/* ... */)
  .finally(() => {
    if (mountedRef.current && connRef.current === thisConn) setRunning(false);
  });
```

### P2 — SSE frame parser `\r` at chunk boundary creates false event separator (`packages/dmloop/src/api/agentRuntime/sseClient.ts:34-35`)

`this.buf = this.buf.replace(/\r\n|\r/g, "\n")` runs on the entire accumulated buffer on every `push()`. If a bare `\r` lands at the tail of a fetch chunk, it gets converted to `\n`. When the next chunk starts with `\n` (the second half of what was originally a `\r\n` line ending), the buffer now contains `\n\n` — a false SSE event boundary that could split a frame in the middle.

Low probability (most servers emit `\n`; chunk boundaries are typically much larger than frame sizes), but a spec-compliance gap.

Fix: track a one-character `pendingCR` flag across pushes instead of re-scanning the whole buffer.

### P2 — Dedup gap for events without seq or id (`packages/dmloop/src/agentRuntime/streamReducer.ts:68-71`)

`eventKey()` returns `null` when an event has neither `seq` nor `id`, so the reducer processes it unconditionally and does not add it to `seen`. On reconnect replay, such events would be processed again — producing duplicated message text (via `message_delta` concatenation) or duplicate tool-call rows.

If the backend guarantees `seq` on every event this is unreachable. Worth either asserting non-null in production or using a positional fallback key for the rare case.

### P2 — No default connection or inactivity timeout (`packages/dmloop/src/api/agentRuntime/sseClient.ts:130-145`)

`connectSse` relies on `AbortController` for cancellation but has no built-in timeout. If the server accepts the TCP connection but never sends data or closes the stream, the client hangs indefinitely until external abort (component unmount or new `send()`). The retry/backoff loop only triggers on connection errors, not on stalled streams.

Consider adding a configurable `timeoutMs` (default ~30s) that aborts the controller if no frame arrives within the window, which would trigger the existing retry path.

## Human-verify

1. **Backend endpoint contracts** — diff/checkpoint/rollback endpoints (`/diff`, `/checkpoints`, `/rollback`) are annotated "待后端确认" throughout. The frontend gracefully degrades to empty state on failure (`.catch(() => [])`), so this is **not** a merge blocker, but the contracts should be validated against the actual backend before the feature is user-facing.

2. **`VITE_AGENT_RUNTIME_BASE` env var** — defaults to `/fleet/api/v1`. Confirm this is the correct production base path and that it's documented in deployment guides. **Not** a merge blocker.

3. **Token provider integration** — `bootstrap.ts` exports `initAgentRuntimeAuth()` but the actual `getToken` implementation is left to the host app. Confirm the host wires it before the panel mounts. **Not** a merge blocker.

- SSE reconnect: `Last-Event-ID` header correctly injected on retry, tracked via `lastId` variable across `runOnce` calls.
- Stream reducer finalization: `error` event sets `finalized=true`, subsequent `done` preserves `error` phase. Late `message_delta` after abort is correctly ignored.
- Reconciliation fallback: `reconcileFallback` rebuilds from authoritative server state via `getEntries` + `getState`, preserves `maybeMissingEvents` flag from both sources.
- `frameToEvent`: unknown event types fall through to `default` in reducer (counted in `seen`/`maxSeq`, no state mutation). Unknown `phase` values don't corrupt state.
- Skill-market changes: icon upload refactoring (hidden `<input>` + `<button>` pattern) is standard a11y improvement. Install button scoping (`!isOwnerCard && onInstall`) is an intentional UX change. Search clear button is self-contained. `IconCropModal` moving `if (!file)` inside the modal wrapper is correct — modal frame renders, content guarded.
- `wordDiff` LCS: tokenizer preserves separators (verified by roundtrip test), LCS table is O(n*m) which is fine for single-line diffs.
- `CheckpointTimeline`: rollback has two-step confirmation UI, `busyId` guards against double-click.
- No `dangerouslySetInnerHTML` in any new file.

## Verdict: APPROVED

Solid, well-structured new feature with good test coverage on the pure-function core. The four P2 items are robustness edge cases — none make a working path unusable or produce user-visible incorrect data under normal conditions. The skill-market changes are clean UI refinements. No security concerns despite the security_sensitive routing tag.

yujiawei (changes requested):
# Code Review — PR #962 (octo-web)

Reviewed at head `9fb0d3af` against `main` (merge-base `4d9be56c`). Fully additive (23 files, +2406/-0) under `packages/dmloop/src/agentRuntime/` and `packages/dmloop/src/api/agentRuntime/`; no existing files touched, public surface unchanged, panel not yet routed. `@octo/loop` suite is green (107/107) and the new code type-checks cleanly.

This is a well-structured draft — the pure-reducer/pure-parser split is genuinely testable, the SSE-over-POST rationale is sound, and the "待后端确认" endpoints are modelled leniently. However, the reducer↔hook contract has a load-bearing correctness bug that breaks multi-turn streaming on the very first happy path, plus several other functional issues. Requesting changes.

I verified the top findings by running them as throwaway unit tests against the actual reducer/parser (not by reading alone); the reproductions are noted inline.

## Blocking (P1)

**1. `send()` never resets the reducer's terminal lock → the 2nd turn's streamed response is silently dropped.**
`streamReducer` sets `finalized=true` on `done`/`error`/abort, and thereafter `reduceEvent` drops incoming `message_delta` (`streamReducer.ts:98`, `if (state.finalized && ev.type === "message_delta") return base`). But `useAgentStream.send()` (`useAgentStream.ts:118`) starts a new turn **without** clearing `finalized`/`phase` — it never calls `reset()` nor any partial clear. So after turn 1 finishes, every `message_delta` of turn 2 is discarded and no assistant message appears on the streaming path. Repro (passes, encoding the bug): after `done` at seq 1, a `message_delta{seq:2, messageId:"m2"}` yields no `m2` message. This breaks a plain single-session two-turn conversation, so it's not a backend-dependent edge. `reset()` exists and is exported but wipes all history, so it's not the right fix here — `send()` needs a partial reset that clears `finalized`/`phase`/`errorMessage` while preserving `messages`/`toolCalls`/`thinking`/`seen`.

**2. Stale `conn.done.finally` flips `running=false` for the *next* turn.**
`useAgentStream.ts:145-154`: `send()` closes the previous connection and starts a new one, but the old connection's `conn.done.finally(() => setRunning(false))` still fires (close → AbortError → `done` resolves) and sets `running=false` on the now-current turn. That re-enables the Send button and hides Stop while the new stream is live, allowing overlapping submissions. Guard the callback with an identity check, e.g. `if (connRef.current === conn) setRunning(false)` (same for the `.catch`).

**3. New-session continuation: the backend-created `session_key` is never captured.**
`AgentRuntimePanel.submit` (`AgentRuntimePanel.tsx:90`) sends a prompt while `activeKey` is `undefined` (new session). Nothing in the event contract (`AgentEvent`) or the panel lifts the newly created `session_key` back into `activeKey`. Result: `abort()` / `reconcileFallback()` / `getDiff` / `listCheckpoints` all early-return on the missing key, and the *next* prompt again goes out with no `session_key` — so each turn risks spawning a fresh backend session instead of continuing. The stream needs to surface the created session key (e.g. a `session`/`session_created` event field or the `prompt` response) and the panel must `setActiveKey` on it.

**4. Intra-message deltas are concatenated in arrival order, not `seq` order.**
`streamReducer.ts:105`: `existing.text + (ev.text ?? "")`. `seq` is only used to order the *placement of a new message* via `insertBySeq`; once a message exists, later deltas for the same `messageId` are appended in arrival order regardless of `seq`. On reconnect/replay where a lower-`seq` delta arrives after a higher one, the text is assembled reversed/corrupted. Repro (passes): `message_delta{seq:2,text:"SECOND"}` then `{seq:1,text:"FIRST"}` for one `messageId` produces `"SECONDFIRST"`. The module's own doc-comment claims out-of-order support ("事件按 seq 排序落位"), so this is a contract gap, not just a nit. Either order per-message fragments by `seq`, or store fragments and re-join on read.

**5. `wordDiff` allocates an unbounded O(N×M) LCS table — one long line can freeze/OOM the tab.**
`wordDiff.ts:19-29` builds a full `(a+1)×(b+1)` number matrix. `DiffView`'s large-file guard (`LARGE_FILE_LINES = 800`) counts **lines**, not tokens per line, so a single minified/bundled line with ~10k tokens on each side allocates ~100M cells before any degradation kicks in — main-thread lockup. Add a per-line token-count threshold that falls back to whole-line highlight (no word diff) above some cap.

## Non-blocking (P2 / nits)

- **Late `thinking`/`tool_call` bypass the terminal lock.** `streamReducer.ts:88` comments that post-finalize content frames are ignored, but only `message_delta` and `phase` are actually guarded — late `thinking` (`:124`) and `tool_call` (`:138`) still mutate an aborted/finished turn. Repro passes (both leak in after `done`). Either honor the comment or correct it.
- **CRLF frame separator split across a chunk boundary yields a false event break.** `sseClient.ts:31` runs `buf.replace(/\r\n|\r/g, "\n")` per chunk, so a chunk ending in `\r` becomes `\n`; if the next chunk begins with `\n`, the two form a spurious `\n\n` boundary and split one event into two. Repro (passes, encoding the bug): `push("data: a\r")` then `push("\ndata: b\r\n\r\n")` produces frames `["a","b"]` instead of one `"a\nb"`. Normalize only fully-buffered content, or strip a single trailing `\r` before re-appending.
- **Reader EOF is always treated as clean completion.** `sseClient.ts:182-195`: any `done=true` from the reader resolves the turn successfully, even if a proxy/network truncated the stream before a terminal `done`/`error` event. Consider requiring a terminal event (or last-`seq` reconciliation) before declaring success so truncation surfaces the "possible event loss" path.
- **`TextDecoder` is never flushed.** `sseClient.ts:181-190` decodes with `{stream:true}` but never issues a final `decoder.decode()`; an incomplete trailing multibyte sequence can be dropped/replaced before `flush()`.
- **Empty `id:` does not reset `Last-Event-ID`.** `sseClient.ts:187` uses `if (frame.id)`, so an explicit empty `id:` field (which per the SSE spec resets the replay cursor) is ignored and the old cursor is retained.
- **`loadHistory` / `refreshCode` / `reconcileFallback` have no staleness guard.** Rapid session switching can let a slower request for the previous session resolve last and overwrite the current session's messages/diffs/checkpoints. Version the async work by `activeKey` (or abort on switch).
- **`full` verbose "Raw event stream" is dead.** `VerboseRenderer` only renders raw events when `rawEvents` is passed, but `AgentRuntimePanel.tsx:142` never passes it and `useAgentStream` doesn't collect raw events — so the documented `full`-level raw dump never shows. Either wire it or drop the branch/claim.

## For a human to confirm (security-sensitive)

- **Parallel auth/transport stack.** This PR intentionally builds a second `fetch`-based client with `Authorization: Bearer <token>` + `credentials: "include"` (`httpClient.ts`, `sseClient.ts`), separate from the app's shared client. The repo convention (CLAUDE.md) is "API calls go through the shared client; do not create separate axios/fetch instances." The PR's stated reason (Bearer + `{ok,data,error}` envelope vs the app's token/`X-Space-Id` contract) is plausible, but sending **both** `credentials:"include"` (cookies) **and** a Bearer header to `/fleet/api/v1` deserves a human sign-off on CSRF posture and on token sourcing (the token provider is host-injected and defaults to `null`).
- The `diff`/`checkpoint`/`rollback` endpoints and fields are frontend-modelled and unconfirmed; `rollback` is destructive — the confirm-before-rollback UX is good, but the endpoint semantics (idempotency, granularity) must be pinned during integration.

## Summary

Solid, testable foundation, but P1 #1 alone breaks basic multi-turn streaming, and #2–#5 are real functional/robustness bugs on the client path (independent of the pending backend). Recommend addressing the P1s (and ideally the reducer/SSE P2s, which are cheap given the pure-function design) before this moves past draft.

Issue #330, opened 2026-06-08
security: OCTO richtext clipboard payload trust boundary — mention UID injection + broadcast sentinel (attachment fixed in #328, mention open)


dolphinsboy:
> **Security redaction (CTO, 2026-06-17):** This issue is labeled `type:security`. Per the org triage policy (no public exploit detail — see [Mininglamp-OSS/.github#75](https://github.com/Mininglamp-OSS/.github/pull/75)), the step-by-step reproduction and the broadcast-sentinel weaponization recipe have been removed from this public body. The structural finding and the defensive fix are retained below. Full repro detail is tracked privately in the internal tracker (OCT-15).

## Root Cause (Structural)

**OCTO richtext clipboard payload (`data-octo-richtext` in `text/html`) is a forgeable, untrusted source.**

The other "trusted" input paths in octo-web validate at their boundary:
- `@` mention input → validated against channel members
- Draft restore → pulled from server (trusted source)
- File/attachment paste → unified attachment gateway — **fixed in PR #328**

Clipboard paste is the one path that takes data from an untrusted external source and injects it into trusted editor state. PR #328 closed the attachment dimension; the **mention-UID dimension remains open**.

---

## Finding 1 — Attachment gate bypass (FIXED in PR #328)

Clipboard-restore path bypassed the unified attachment security gateway (extension blocklist, per-file and total size limits). **Fixed** in PR #328 (commit 55c1d1a0c2e2): paste now routes through the gateway with an added size cap, MIME allowlist, and hardened payload extraction.

---

## Finding 2 — Mention UID trust boundary violation (OPEN)

`buildInlineContentForRichTextPaste` (richTextPaste.ts) builds editor mention nodes directly from clipboard-supplied mention UIDs **without member validation**. Shape validation occurs (offset/length) but UID origin is never checked, so a forged clipboard payload can introduce mention nodes referencing arbitrary or privileged routing identifiers. The higher-severity dimension is that special broadcast-routing identifiers, if accepted from clipboard, could turn a single paste into a wide broadcast. (Concrete payload/repro withheld per the security-redaction note above.)

**Fix (priority order):**

1. **(P1 must-fix) Reject broadcast routing sentinels at clipboard ingestion** — degrade them to plain `@label` text. This removes the broadcast-amplification dimension.
2. **(P1) Validate every pasted mention** in `buildInlineContentForRichTextPaste`: UID must exist in `members` AND `label` must match the member's name; unmatched → plain text node.
[...]

Issue #58, opened 2026-05-19
feat(mention): UI input + render for @所有人 / @所有AI (PR-C of mention three-state)


yujiawei:
**Branch**: `feat/mention-three-state` (based on `main`)
**RFC**: drafts/mention-three-state-rfc.md v1

## Scope

UI layer for mention three-state: input emits new fields, render distinguishes three categories.

## Files to change

| # | File | Type | Notes |
|---|------|------|-------|
| 1 | `packages/dmworkbase/src/Components/MessageInput/index.tsx` | EDIT | Add "@所有人" + "@所有AI" as fixed top items in @ dropdown |
| 2 | `packages/dmworkbase/src/Components/Conversation/index.tsx` | EDIT | Render matrix: humans / ais / both / legacy all |
| 3 | `apps/web/src/__tests__/mentionRefactor.test.ts` | EDIT | Add 4 render matrix cases |

## Input behavior

- 选 "@所有人" → 文本插入 `@所有人` + payload `mention.humans=1`
- 选 "@所有AI" → 文本插入 `@所有AI` + payload `mention.ais=1`
- 老 "@所有人" 输入路径继续走 `mention.all=1`（server rewrite 为 humans=1）

## Render behavior

- humans=1 → 高亮 "@所有人"
- ais=1 → 高亮 "@所有AI"
- humans=1 + ais=1 → 高亮 "@所有人 + @所有AI"
- all=1 (legacy) → 高亮 "@所有人"

## Acceptance

- [ ] @ dropdown shows two new top items in any group chat
- [ ] mentionRefactor.test.ts 4 matrix cases pass
- [ ] No regression on existing @member functionality
- [ ] Type-check + build clean

## Related

- octo-server #94 (PR-A, server-side rewrite)
- octo-adapters #44 (PR-B, adapter passthrough)

PR #971, opened 2026-07-22
fix(message-input): 保留粘贴富文本链接地址

### LINKED ISSUE #871: 粘贴富文本超链接后发送消息会丢失链接地址
### Description

从 GitHub 等网页复制一个“显示文字与目标 URL 不同”的富文本超链接并粘贴到 Octo 消息输入框时，输入框会保留链接样式并可以点击；但发送后，消息只剩链接的显示文字，原始 `href` 被丢弃，收件人无法点击，也无法从消息内容恢复原地址。

例如复制 GitHub 仓库链接：

- 显示文字：`labilio/octo-web: Web & desktop (Electron) client...`
- 目标地址：`https://github.com/labilio/octo-web`

粘贴后输入框中显示为可点击链接；发送后只剩显示文字，`https://github.com/labilio/octo-web` 不再存在。

代码调查确认了以下执行路径：

1. 消息输入框使用 Tiptap，普通外部 HTML 链接粘贴会由编辑器保留为带 `href` 的 Link mark。
2. 发送时，`packages/dmworkbase/src/Components/MessageInput/index.tsx` 调用 `extractMentionsFromEditor(editor, true)`。
3. 该函数遍历 Tiptap JSON 时，对 text 节点只追加 `node.text`，没有读取 `node.marks` 中的 Link mark 及其 `attrs.href`。
4. 最终传给 `onSend` 的 `content` 只包含显示文字，因此链接目标在发送边界被丢弃。
5. 消息列表使用 `ReactMarkdown` 和 `remarkGfm` 渲染。直接发送完整 `https://...` 时仍能识别为链接；但富链接只剩标题文字后，渲染阶段无法恢复原 URL。

这说明输入框展示了当前发送格式无法保留的链接语义，形成“发送前可点击、发送后不可点击”的所见非所得，并造成实际信息丢失。建议在发送前将 Link mark 转换为 Markdown `[显示文字](URL)` 或由消息协议保留链接信息；如果产品暂不支持富链接，则粘贴时应立即降级并明确显示最终会发送的内容，而不应在输入框中呈现无法保存的可点击状态。

相关历史记录：

- #319 修复的是 Octo 图文混编消息的内部 RichText 剪贴板往返；当前问题发生在普通外部 HTML 超链接粘贴路径，现版本仍可复现，因此新建 Issue 跟踪。
- #355 处理的是已发送消息在“引用预览”中的链接渲染问题；当前问题在发送序列化阶段已经丢失 URL，责任位置不同。
- #870 是同一上层能力边界下的另一个独立问题：自动链接规则把 `xxx.md` 等文件名误判为网址。两者修复位置和关闭条件不同，因此分别跟踪。

本次调查基于 `main` commit `93659b58409890aa26c8990916d9ba113fd9fc56`。

### Steps to Reproduce

1. 在浏览器中打开一个包含富文本超链接的页面，例如 GitHub 仓库链接。
2. 复制链接的显示文本，使剪贴板同时包含显示文字和目标 URL。
3. 将内容粘贴到 Octo Web 消息输入框。
4. 确认输入框中内容显示为可点击的超链接。
5. 发送消息。
6. 查看刚发送的消息并尝试点击。

### Expected Behavior

发送后的消息应保留粘贴链接的目标 URL，并继续可点击；显示内容与发送结果应保持一致。

### Actual Behavior

输入框中链接可点击，但发送后只保留显示文字，原始 URL 丢失，消息变为不可点击的普通文本。

### Client Version

Web 1.5.0（仓库 `package.json`；线上页面未展示更细构建号）

### Platform

Web

### Affected Component

Message Input

### Severity

P2 - Non-core feature broken

### Reproducibility

- [x] Consistently reproducible

### Environment

Web 端消息输入框；代码验证基于 `main` commit `93659b58409890aa26c8990916d9ba113fd9fc56`。

### Logs / Screenshots

无需上传截图。复现现象与代码中的 Link mark 丢弃路径一致；相关历史见 #319、#355、#870。

botshen:
## 变更

- 发送前读取 Tiptap text node 的 Link mark，将安全链接序列化为 Markdown 链接
- 普通消息和图片/文件混排的 orderedBlocks 共用同一序列化逻辑
- 转义链接标题中的 Markdown 符号，并规范化 URL 中的括号
- 非 http/https 或无效目标退化为可见文字，mention 与草稿读取逻辑保持不变

## 验证

- pnpm --dir packages/dmworkbase exec vitest run src/Components/MessageInput/__tests__（6 files / 92 tests）
- pnpm --dir apps/web build
- git diff --check

Closes #871

### DIFF
--- a/packages/dmworkbase/src/Components/MessageInput/__tests__/mentionSendParse.test.ts
+++ b/packages/dmworkbase/src/Components/MessageInput/__tests__/mentionSendParse.test.ts
@@ -18,6 +18,7 @@
 import { describe, it, expect } from "vitest";
 import {
   parseSendMentionText,
+  serializeEditorTextNodeForSend,
   serializeMentionMarker,
   stripTrustMark,
   parseDraftToContent,
@@ -41,6 +42,61 @@ const MEMBERS: SendParseMember[] = [
 // build the trusted (sanctioned) form with this helper.
 const trusted = (uid: string, label: string) =>
   `@[${MENTION_TRUST_MARK}${uid}:${label}]`;
+
+describe("serializeEditorTextNodeForSend — pasted rich links (octo-web#871)", () => {
+  it("preserves a Link mark as Markdown when label and target differ", () => {
+    expect(
+      serializeEditorTextNodeForSend({
+        text: "labilio/octo-web: Web & desktop client",
+        marks: [
+          {
+            type: "link",
+            attrs: { href: "https://github.com/labilio/octo-web" },
+          },
+        ],
+      })
+    ).toBe(
+      "[labilio/octo-web: Web & desktop client](https://github.com/labilio/octo-web)"
+    );
+  });
+
+  it("escapes Markdown syntax in the visible label and normalizes parentheses in the URL", () => {
+    expect(
+      serializeEditorTextNodeForSend({
+        text: "Docs [v2] *beta*",
+        marks: [
+          {
+            type: "link",
+            attrs: { href: "https://example.com/docs_(v2)" },
+          },
+        ],
+      })
+    ).toBe(
+      "[Docs \\[v2\\] \\*beta\\*](https://example.com/docs_%28v2%29)"
+    );
+  });
+
+  it.each(["javascript:alert(1)", "mailto:user@example.com", "not a url"])(
+    "fails closed to visible text for unsafe target %s",
+    (href) => {
+      expect(
+        serializeEditorTextNodeForSend({
+          text: "visible label",
+          marks: [{ type: "link", attrs: { href } }],
+        })
+      ).toBe("visible label");
+    }
+  );
+
+  it("leaves plain text and unrelated marks unchanged", () => {
+    expect(
+      serializeEditorTextNodeForSend({
+        text: "plain text",
+        marks: [{ type: "bold" }],
+      })
+    ).toBe("plain text");
+  });
+});
 
 describe("parseSendMentionText — forged-paste broadcast bypass (octo-web#330)", () => {
   it.each([
--- a/packages/dmworkbase/src/Components/MessageInput/__tests__/mentionSendParse.test.ts
+++ b/packages/dmworkbase/src/Components/MessageInput/__tests__/mentionSendParse.test.ts
@@ -56,7 +56,7 @@ describe("serializeEditorTextNodeForSend — pasted rich links (octo-web#871)", 
         ],
       })
     ).toBe(
-      "[labilio/octo-web: Web & desktop client](https://github.com/labilio/octo-web)"
+      "[labilio/octo-web&#58; Web &amp; desktop client](https://github.com/labilio/octo-web)"
     );
   });
 
@@ -71,9 +71,52 @@ describe("serializeEditorTextNodeForSend — pasted rich links (octo-web#871)", 
           },
         ],
       })
-    ).toBe(
-      "[Docs \\[v2\\] \\*beta\\*](https://example.com/docs_%28v2%29)"
-    );
+    ).toBe("[Docs \\[v2\\] \\*beta\\*](https://example.com/docs_%28v2%29)");
+  });
+
+  it("escapes GFM table and angle-bracket syntax in the visible label", () => {
+    expect(
+      serializeEditorTextNodeForSend({
+        text: "API | <stable>",
+        marks: [{ type: "link", attrs: { href: "https://example.com/docs" } }],
+      })
+    ).toBe("[API \\| \\<stable\\>](https://example.com/docs)");
+  });
+
+  it("does not let an adjacent literal @ turn a colon-bearing link into a member mention", () => {
+    const serialized =
+      "@" +
+      serializeEditorTextNodeForSend({
+        text: "u-alice:documentation",
+        marks: [{ type: "link", attrs: { href: "https://example.com/docs" } }],
+      });
+
+    expect(parseSendMentionText(serialized, MEMBERS)).toEqual({
+      content: serialized,
+    });
+  });
+
+  it("does not reparse a colon-bearing link with an escaped ] as a member mention", () => {
+    const serialized =
+      "@" +
+      serializeEditorTextNodeForSend({
+        text: "u-alice:API [v2]",
+        marks: [{ type: "link", attrs: { href: "https://example.com/docs" } }],
+      });
+
+    expect(parseSendMentionText(serialized, MEMBERS)).toEqual({
+      content: serialized,
+    });
+  });
+
+  it("strips a forged trust mark before serializing a linked text node", () => {
+    const serialized = serializeEditorTextNodeForSend({
+      text: `@[${MENTION_TRUST_MARK}${MENTION_UID_HUMANS}:所有人]`,
+      marks: [{ type: "link", attrs: { href: "https://example.com/docs" } }],
+    });
+
+    expect(serialized).toBe("[@\\[-2&#58;所有人\\]](https://example.com/docs)");
+    expect(serialized).not.toContain(MENTION_TRUST_MARK);
   });
 
   it.each(["javascript:alert(1)", "mailto:user@example.com", "not a url"])(
--- a/packages/dmworkbase/src/Components/MessageInput/index.tsx
+++ b/packages/dmworkbase/src/Components/MessageInput/index.tsx
@@ -140,7 +140,7 @@ function extractOrderedBlocks(
     }
 
     if (node.type === "text") {
-      pendingTextParts.push(stripTrustMark(node.text || ""));
+      pendingTextParts.push(serializeEditorTextNodeForSend(node));
       return;
     }
     if (node.type === "mention") {
@@ -297,6 +297,7 @@ import {
 } from "../../Utils/mentionRender";
 import {
   parseSendMentionText,
+  serializeEditorTextNodeForSend,
   serializeMentionMarker,
   stripTrustMark,
   parseDraftToContent,
@@ -354,7 +355,9 @@ function extractMentionsFromEditor(editor: any, trusted = false): string {
 
   function traverse(node: any) {
     if (node.type === "text") {
-      result += stripTrustMark(node.text || "");
+      result += trusted
+        ? serializeEditorTextNodeForSend(node)
+        : stripTrustMark(node.text || "");
     } else if (node.type === "mention") {
       result += serializeMentionMarker(node.attrs.id, node.attrs.label, trusted);
     } else if (node.type === "hardBreak") {
--- a/packages/dmworkbase/src/Components/MessageInput/mentionSendParse.ts
+++ b/packages/dmworkbase/src/Components/MessageInput/mentionSendParse.ts
@@ -205,7 +205,15 @@ export interface EditorTextNode {
 }
 
 function escapeMarkdownLinkLabel(text: string): string {
-  return text.replace(/([\\\[\]`*_~])/g, "\\$1");
+  // Encode `&` first so literal character-reference-looking text remains
+  // literal. A colon is encoded as a CommonMark character reference rather
+  // than backslash-escaped: the renderer still shows `:`, while the send-side
+  // `@[uid:label]` regex cannot synthesize a mention when an `@` text node is
+  // immediately followed by this link node (octo-web#971 review blocker).
+  return text
+    .replace(/&/g, "&amp;")
+    .replace(/:/g, "&#58;")
+    .replace(/([\\\[\]`*_~|<>])/g, "\\$1");
 }
 
 /**
--- a/packages/dmworkbase/src/Components/MessageInput/mentionSendParse.ts
+++ b/packages/dmworkbase/src/Components/MessageInput/mentionSendParse.ts
@@ -29,6 +29,7 @@
 
 import { subscriberDisplayName } from "../../Utils/displayName";
 import type { SubscriberLike } from "../../Utils/displayName";
+import { isSafeUrl } from "../../Utils/security";
 import {
   MENTION_UID_LEGACY_ALL,
   MENTION_UID_HUMANS,
@@ -191,6 +192,44 @@ export function stripTrustMark(text: string): string {
   return text.includes(MENTION_TRUST_MARK)
     ? text.split(MENTION_TRUST_MARK).join("")
     : text;
+}
+
+export interface EditorTextMark {
+  type: string;
+  attrs?: Record<string, unknown>;
+}
+
+export interface EditorTextNode {
+  text?: string;
+  marks?: EditorTextMark[];
+}
+
+function escapeMarkdownLinkLabel(text: string): string {
+  return text.replace(/([\\\[\]`*_~])/g, "\\$1");
+}
+
+/**
+ * Serialize a Tiptap text node at the send boundary.
+ *
+ * External rich-text paste keeps hyperlinks as Link marks. The message wire
+ * format is Markdown text, so a safe http/https mark must become
+ * `[label](url)` before the editor JSON is flattened. Unsafe or malformed
+ * targets fail closed to the visible label. Text-origin mention trust marks
+ * are still stripped before any output is produced.
+ */
+export function serializeEditorTextNodeForSend(node: EditorTextNode): string {
+  const text = stripTrustMark(node.text || "");
+  const linkMark = node.marks?.find((mark) => mark.type === "link");
+  const href = linkMark?.attrs?.href;
+
+  if (!text || typeof href !== "string" || !isSafeUrl(href)) {
+    return text;
+  }
+
+  const normalizedHref = new URL(href).href
+    .replace(/\(/g, "%28")
+    .replace(/\)/g, "%29");
+  return `[${escapeMarkdownLinkLabel(text)}](${normalizedHref})`;
 }
 
 // ─── Draft deserialization ────────────────────────────────────────────────


lml2468 (dismissed):

审查锚定 head `dd3721e6`。3 文件 +100/−2,发送前把 Tiptap text node 的 Link mark 序列化成 Markdown 链接(修粘贴富文本链接地址丢失)。核心是安全序列化,逐条核过 + 跑全套测试。

### ✅ 两道闸门均通过

**规格符合**:与描述一致 —— 安全链接序列化为 `[label](url)`,普通消息与图文混排 orderedBlocks 共用同一逻辑,mention/草稿读取不变。`Closes #871`。无 over-build。

**代码质量 —— 安全序列化(逐条核实)**:
- **🔒 不安全 URL fail-closed**:`serializeEditorTextNodeForSend` 先 `isSafeUrl(href)`(http/https only)—— `javascript:`/`mailto:`/畸形目标一律退化为**可见纯文字、不产链接**(测试覆盖三种);
- **🔒 Markdown 注入转义**:`escapeMarkdownLinkLabel` 转义 label 里的 `` \[]`*_~ `` —— 粘贴的 `Docs [v2] *beta*` 无法突破 `[...]` 语法或注入 markdown;URL 括号规范化为 `%28/%29`,不破坏 `](url)` 闭合;
- **🔒 mention 伪造边界保持**:`stripTrustMark` 在 send 路径(:221 首行)与 untrusted 路径**都先执行** —— 文本来源的 trust mark 恒被剥离,#330 防伪造不削弱;
- **trusted/untrusted 语义正确**:`extractMentionsFromEditor` 仅 `trusted`(send)走链接序列化,`trusted=false`(草稿/读取)只 `stripTrustMark` —— 草稿要 round-trip mention marker(#330),send 才需 markdown 链接,拆分得当;
- **`new URL(href).href` 不抛**:仅在 `isSafeUrl` 通过后到达(后者已在 try/catch 里 `new URL` 验证过),守卫顺序安全。

### ✅ 验证

- `@octo/base` `MessageInput/__tests__` → **92/92 全绿**(含新增 `serializeEditorTextNodeForSend` 6 例:安全链接保留、label 转义+URL 括号归一、3 种 unsafe URI fail-closed、plain/其它 mark 原样);

### 结论

粘贴链接序列化处理得安全周到:unsafe URL fail-closed 成纯文字、label markdown 注入转义、URL 括号归一、mention 伪造边界(stripTrustMark 双路径)保持、trusted/untrusted 拆分正确、new URL 守卫顺序安全;测试 92/92 覆盖全路径。无阻塞。**APPROVE**。

Jerry-Xin (dismissed):
**APPROVE** — link-mark serialization is safe and the send/draft split is correct.

## Summary
This serializes Tiptap `Link` marks into Markdown `[label](url)` at the send boundary so pasted rich-text hyperlinks keep their target, while failing closed to visible text for anything unsafe. Byte-verified at head `dd3721e6`; no XSS/scheme-injection exposure, no plain-text paste regression.

## Security review (primary concern: scheme injection / XSS)
- **Scheme whitelist reused correctly.** `serializeEditorTextNodeForSend` (`packages/dmworkbase/src/Components/MessageInput/mentionSendParse.ts:216`) gates the href through the existing `isSafeUrl` helper (`packages/dmworkbase/src/Utils/security.ts`), which only accepts `http:`/`https:` via `new URL(...).protocol` and returns `false` on parse failure. Non-http(s) schemes and malformed targets therefore fall through to the visible-label branch — the unit test explicitly covers the dangerous-scheme, `mailto:`, and non-URL cases all degrading to plain text. No dangerous-scheme href can survive into the wire format.
- **No raw-HTML path.** The output is Markdown *text* on the send wire, not DOM/innerHTML. The editor uses Tiptap React `EditorContent` (React-managed), and the only `dangerouslySetInnerHTML` in this directory is unrelated (`VoiceFeedbackNotice.tsx`, already DOMPurify-sanitized). No event-handler/script passthrough is introduced.
- **Markdown-structure injection handled.** The visible label is escaped for `\ [ ] \` * _ ~` (`escapeMarkdownLinkLabel`), preventing label breakout; URL parens are percent-encoded (`%28`/`%29`) so a `)` in the target cannot prematurely close the Markdown link. `new URL(href).href` cannot throw here because `isSafeUrl` already parsed successfully.

## Behavior / regression
- **Send vs draft split is correct.** `extractOrderedBlocks` (`index.tsx:143`) and the `trusted` branch of `extractMentionsFromEditor` (`index.tsx:357`) use the new serializer only on the send path; the draft/read path (`trusted=false`) still uses `stripTrustMark`, so draft round-trip is unchanged. Text nodes without a link mark return `stripTrustMark(text)` verbatim → no plain-text paste regression.
- Empty text, unrelated marks (e.g. bold), and no-link nodes all pass through unchanged.

## Verification
- `pnpm --dir packages/dmworkbase exec vitest run src/Components/MessageInput/__tests__` → **6 files / 92 tests passed** (includes the 4 new serialize cases).
- CI green on merged checks; `mergeable: MERGEABLE`.

Nice reuse of the existing scheme-whitelist and fail-closed default. LGTM.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #971 (octo-web)

## Summary

This PR preserves pasted rich-text hyperlinks when sending messages from the MessageInput editor. Previously, Tiptap Link marks on text nodes were silently dropped during send serialization — only the visible label survived. The fix adds `serializeEditorTextNodeForSend` (`mentionSendParse.ts:220-234`), which converts a safe `http`/`https` link mark into Markdown `[label](url)` syntax at the send boundary, while unsafe targets (`javascript:`, `mailto:`, malformed) degrade to the visible label. The new serializer is wired into both send paths (`extractOrderedBlocks` and `extractMentionsFromEditor` with `trusted=true`) while the draft/read path (`trusted=false`) remains unchanged.

## Verification

- ✅ **Security boundary preserved** — `serializeEditorTextNodeForSend` calls `stripTrustMark` on the raw text before any link serialization (`mentionSendParse.ts:222`), maintaining the octo-web#330 trust-mark guarantee.
- ✅ **URL safety gate** — `isSafeUrl` at `Utils/security.ts:5` validates `http:`/`https:` only via `new URL()` + protocol allowlist; `javascript:`, `data:`, `mailto:` all fail closed.
- ✅ **Call-site wiring correct** — `extractOrderedBlocks` (`index.tsx:143`) and `extractMentionsFromEditor` with `trusted=true` (`index.tsx:358-360`) both route through the new function. The `trusted=false` path (`index.tsx:867`, the `text()` method) still uses `stripTrustMark` only — no link preservation on the draft/read path, consistent with the existing lossy draft round-trip.
- ✅ **Markdown escaping** — `escapeMarkdownLinkLabel` (`mentionSendParse.ts:213`) covers `\ [ ] ` * _ ~`, the characters that can break out of a `[label](url)` construct. Parentheses in URLs are percent-encoded (`%28`/`%29`).
- ✅ **No mention-grammar collision** — escaped `[` and `]` in link labels prevent the `@[uid:label]` send parser from matching forged mention syntax inside link text.
- Static analysis only at head `dd3721e6`; build and tests not executed in this environment.

## Findings

### P2 — Missing regression test for trust-mark strip + link serialization integration (`mentionSendParse.test.ts:46`)

**Diff-scope**: new (the integration path is newly introduced in this PR).

The test suite covers happy-path link preservation, Markdown escaping, unsafe-target fail-closed, and plain-text passthrough — all good. However, no test proves that a forged `MENTION_TRUST_MARK` character in a link-labeled text node is stripped before the Markdown link is produced. The `stripTrustMark` call at `mentionSendParse.ts:222` is trivially correct today, but the integration (trust-mark removal + link serialization in one pass) has no regression lock. If a future refactor reorders operations — for instance, short-circuiting on the link mark before stripping — the trust boundary could silently break.

Suggested addition:

```ts
it("strips trust mark from link-labeled text before serialization", () => {
  expect(
    serializeEditorTextNodeForSend({
      text: `@[${MENTION_TRUST_MARK}-2:所有人]`,
      marks: [{ type: "link", attrs: { href: "https://example.com" } }],
    })
  ).toBe("[@\\[-2:所有人\\]](https://example.com/)");
});
```

This simultaneously locks the trust-mark strip, the label escape, and the URL normalization in one assertion.

**R4 mapping**: P2 only (no runtime breakage, test gap for future-proofing) → does not block APPROVE.

## Data Flow Trace

### Consumed data: `serializeEditorTextNodeForSend(node)` output

| Consumer | Upstream source | Data actually flows? |
|---|---|---|
| `extractOrderedBlocks` → `pendingTextParts.push(...)` at `index.tsx:143` | Tiptap editor JSON `node.type === "text"` | ✅ Yes — `processNode` is called for every node in the editor JSON. Text nodes with Link marks get serialized as `[label](url)`. |
| `extractMentionsFromEditor(editor, true)` at `index.tsx:358-360` | Tiptap editor JSON `node.type === "text"`, gated by `trusted === true` | ✅ Yes — the `traverse` function hits every text node. When `trusted=true` (send path at `:979`), links are preserved. When `trusted=false` (read/draft at `:867`), only `stripTrustMark` runs. |
| `formatMentionTextV2(formattedText)` at `index.tsx:981` | Output of `extractMentionsFromEditor(editor, true)` which now contains Markdown links | ✅ Yes — `parseSendMentionText` regex `/@\[([^:]+):([^\]]+)\]/g` won't false-match inside `[escapedLabel](url)` because `@` is never followed by an unescaped `[` in link-serialized text. |
| `props.onSend!(content, mention, ..., orderedBlocks)` at `index.tsx:1006` | `content` from `formatMentionTextV2` and `orderedBlocks` from `extractOrderedBlocks` | ✅ Yes — both contain Markdown link syntax for pasted links. The server/adapter downstream receives Markdown text and renders it. |

### Consumed data: `isSafeUrl(href)` return value

| Consumer | Upstream source | Data actually flows? |
|---|---|---|
| `serializeEditorTextNodeForSend` guard at `mentionSendParse.ts:226` | `isSafeUrl` at `Utils/security.ts:5` — constructs `new URL(url)`, checks `protocol ∈ {http:, https:}` | ✅ Yes — returns `true` only for valid http/https URLs. Malformed URLs throw in `new URL()` and are caught → `false`. Non-http protocols → `false`. |

### Consumed data: `new URL(href).href` normalization

| Consumer | Upstream source | Data actually flows? |
|---|---|---|
| `normalizedHref` at `mentionSendParse.ts:230-231` | `new URL(href).href` after `isSafeUrl` already validated the URL | ✅ Yes — `isSafeUrl` returning `true` guarantees `new URL(href)` won't throw (same WHATWG parser). Parentheses are then percent-encoded to prevent Markdown syntax breakage. |

- **C1 (dual-path parity)**: The PR changes two symmetric send serialization paths — `extractOrderedBlocks` and `extractMentionsFromEditor(trusted=true)`. Both now call `serializeEditorTextNodeForSend`. The read/draft path (`trusted=false`) is unchanged. **Clear** — both send paths are symmetrically updated.

- **C2 (control-flow ordering / nested reuse)**: `serializeEditorTextNodeForSend` is called from two sites. No nesting or double-invocation risk — each call processes one text node independently. **Clear**.

- **C3 (authorization boundary ≠ capability boundary)**: N/A — no auth/permission changes.

- **C4 (authorization lifecycle / container-member state cascade)**: N/A — no auth changes.

- **C5 (build/note pass ≠ runtime path correct)**: N/A — no build/packaging changes. The change is pure TypeScript logic in a send-boundary serializer.

- **C6 (governance/policy/security doc self-consistency)**: N/A — no docs or policy changes.

N/A — this is the first review pass on this PR.

## Extra Observations

- **Double URL parse**: `isSafeUrl` constructs `new URL()` internally, then `serializeEditorTextNodeForSend` constructs `new URL(href).href` again at `mentionSendParse.ts:230`. Minor redundant work, not a correctness issue.
- **Non-link marks dropped**: Bold/italic/code marks on text nodes are still discarded (same behavior as before this PR). The PR only adds link preservation, not inline formatting serialization. Not a regression.
- **Draft path intentionally lossy**: The `text()` method at `index.tsx:867` still uses `stripTrustMark` only. Pasted links are not preserved in the draft/read path. This is consistent with the existing draft round-trip model (all formatting is dropped). Not a regression.

## Verdict: APPROVED

Clean, well-scoped fix that correctly preserves pasted hyperlinks at the send boundary while maintaining the trust-mark security model. The URL safety gate, Markdown escaping, and fail-closed behavior for unsafe targets are all sound. The one P2 (missing integration test for trust-mark stripping in the link serialization path) is non-blocking but worth adding as a regression lock.

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Security model (trust mark stripping, URL protocol gate, fail-closed for unsafe targets) is correctly preserved in the new link serialization path. Both send-path call sites are symmetrically wired. The single P2 (missing integration test for trust-mark + link combo) is a regression-lock suggestion, not a blocker.

yujiawei (changes requested):
# Code Review — PR #971 (octo-web)

Reviewed at head `dd3721e645d86126e2a2f82fd5276909ddaaf3b7` against merge-base `300394d5` with `main`.

The intent is good and the fail-closed instincts are right (http/https-only via `isSafeUrl`, label escaping, `(`/`)` normalization in the href, unsafe targets degrade to visible text). Tests are focused and readable. However, there is one blocking correctness issue where the new Markdown output collides with the existing send-side mention grammar.

## Spec compliance

The diff implements the stated change (serialize safe Link marks to `[label](url)` at the send boundary; shared by the plain-text and ordered-blocks paths; unsafe targets fall back to text; mention/draft behavior unchanged). No scope creep and nothing obviously missing. **Spec: ✅**

## Findings

### P1 — Serialized link output collides with the send-side mention grammar (blocking)

`serializeEditorTextNodeForSend` emits `[label](url)` (`mentionSendParse.ts:232`). Both call sites then flatten adjacent text parts with **no separator** — `extractOrderedBlocks` joins with `pendingTextParts.join("")` (`index.tsx:123`) and `extractMentionsFromEditor` concatenates directly (`index.tsx:356-360`). The flattened string is re-parsed by `parseSendMentionText`, whose marker regex is `/@\[([^:]+):([^\]]+)\]/g` (`mentionSendParse.ts:81`).

So a literal `@` immediately preceding a serialized link whose **label contains a colon** synthesizes a mention marker across the node boundary. Colon-in-label is the *common* case for pasted links (this PR's own test uses `"labilio/octo-web: Web & desktop client"`), so this is not an exotic input.

Concrete repro — a text node `@` followed by a linked node (label `alice:documentation`, href `https://example.com`) serializes to:

```
@[alice:documentation](https://example.com/)
```

which the regex captures as `uid="alice"`, `name="documentation"`. Effects:

1. **The link is destroyed.** The `[label:` portion is consumed as a mention marker and the trailing `](https://example.com/)` is left as inert literal text — defeating the exact behavior this PR exists to add.
2. **An unintended member mention can fire.** If the substring before the first `:` equals a current member uid, `parseSendMentionText` (`mentionSendParse.ts:134-142`) routes a real mention (`uids.push`, `entities.push`) — i.e. the sender silently pings someone. The rich-paste HTML payload is attacker-controllable, so the label prefix can be crafted to target a known uid.

Scope note (not a broadcast bypass): a broadcast sentinel (`-1/-2/-3`) still requires `MENTION_TRUST_MARK`, which text nodes can never carry (`stripTrustMark` runs first), so this degrades to inert text on the broadcast path (`mentionSendParse.ts:102-108`). The impact is bounded to link corruption + unintended member mention, not channel-wide fan-out.

Escaping alone is not a complete fix: `escapeMarkdownLinkLabel` (`mentionSendParse.ts:207`) does not escape `:`, but even if it did, the mention regex does not understand Markdown backslash-escapes — an escaped `]` inside a label (`@[alice:Al\]ice](...)`) still terminates the marker at the escaped `]` and mangles the text. The robust fix is to stop the serialized Markdown from being able to form mention grammar across node boundaries — e.g. resolve mentions from the structured nodes rather than re-parsing the fully-flattened Markdown string, or neutralize a leading `@[` sequence introduced by link serialization.

### P2 — Label escaping misses `|`, `<`, `>` (non-blocking)

`escapeMarkdownLinkLabel` escapes `` \ [ ] ` * _ ~ `` but not `|`, `<`, or `>`. The receiver renders with GFM (`remark-gfm`) + `remark-breaks`, so an unescaped `|` in a link label placed inside a Markdown table context, or angle brackets colliding with autolink/HTML handling, can break rendering on the receiving side. Low likelihood, worth hardening alongside the P1 fix.

## Suggestions (non-blocking)

- Add a regression test that feeds `serializeEditorTextNodeForSend` output back through `parseSendMentionText` with a leading `@` and a colon-bearing label — that end-to-end round-trip is the gap that let the P1 collision through; the current unit tests only assert the serializer in isolation.
- Add a test with `]`, `(`, `)` inside the href to lock down the destination normalization, and one asserting `stripTrustMark` still fires for a plain text node carrying the trust mark.

## Verdict

The feature is close, but the grammar collision silently corrupts links and can send unintended mentions in a common paste scenario, so it should be resolved before merge.

botshen:
已在 6f644c2f 处理阻塞项：

- 链接标签中的 `:` 改为 CommonMark 字符引用 `&#58;`（接收端仍显示冒号），从源头避免相邻文本节点 `@` 与链接节点拼成 `@[uid:label]` 提及语法。
- 先转义 `&`，避免原始实体样式文本被误解码；补齐 `|`、`<`、`>` 的 GFM 转义。
- 增加“相邻 @ + 含冒号链接”“标签含转义 ]”“链接文本携带伪造 trust mark”等回归测试。

验证：MessageInput 6 个测试文件 / 96 项通过；`apps/web` 生产构建通过。请复审。

lml2468 (approved):
## Review Verdict（复审 @ 6f644c2f）：可合并（APPROVE）—— 维持 APPROVE

复审锚定新 head `6f644c2f`。相对上次 APPROVE `dd3721e6`,delta 是 `escapeMarkdownLinkLabel` + 测试,**修掉了 yujiawei 提的 mention-语法冲突 🔴 与 `|`/`<`/`>` 转义 🟡**。逐条核过、测试全绿。

### 🔴→✅ yujiawei:序列化链接与 mention 语法冲突 —— 已修(巧妙)

- **风险**:mention 解析用 `/@\[([^:]+):([^\]]+)\]/g`,需 `@[...:...]`。若链接 label 含 `@[uid:name]` 形状,序列化成 `[...]` 后可能被 `parseSendMentionText` 误解析成成员 mention。
- **修法(核过正确)**:`escapeMarkdownLinkLabel` 把 label 里的 `:` 编码为 CommonMark 字符引用 `&#58;`(渲染仍是 `:`,但**破坏 `@[...:...]` 正则**——括号内无字面 `:` → 不匹配)、`&` 先编码为 `&amp;`(防字符引用碰撞/双编码),`[`/`]` 仍反斜杠转义(进一步破坏括号结构)。
- **双向坐实**:①解析侧 `/@\[([^:]+):...\]/` 对 `&#58;` 不匹配 → 无误 mention(新测试断言);②接收侧 `MarkdownContent` 走 `ReactMarkdown`+`remarkGfm`,CommonMark 标准会**解码字符引用**(`&#58;`→`:`、`&amp;`→`&`),用户看到正常 `:`/`&`,非字面 `&#58;`(无 rehype-raw,标准文本解码即可)。

### 🔴→✅ yujiawei:label 转义漏 `|`/`<`/`>` —— 已修

- `escapeMarkdownLinkLabel` 转义字符集扩为 `[\\\[\]`*_~|<>]`,覆盖 GFM 表格竖线 + 尖括号;新测试 `API | <stable>` → `API \| \<stable\>`。

### ✅ 上轮结论保持

- unsafe URL fail-closed(`isSafeUrl` http/https)、URL 括号 `%28/%29` 归一、`stripTrustMark` 双路径(mention 伪造边界)、trusted/untrusted 拆分、`new URL` 守卫顺序 —— 本轮未动,保持;
- `@octo/base` MessageInput `__tests__` → **96/96 全绿**(比上轮 +4:GFM 竖线/尖括号转义、两条 mention-冲突场景、序列化前剥伪造 trust mark);

### 🟡 非阻塞(沿用 mochashanyao)

- 建议补 trust-mark 剥离 × 链接序列化的集成回归测试 —— 本轮新增的「strips a forged trust mark before serializing」用例已部分覆盖。

### 结论

yujiawei 的 mention-语法冲突 🔴(`:`→`&#58;` 破坏正则 + `[]` 转义,双向坐实解析不误 mention、接收端字符引用正确解码)与 `|`/`<`/`>` 转义 🟡 都正确修复;`&` 先编码防碰撞;上轮 unsafe-URL/括号归一/trust-mark 结论保持,测试 96/96。无阻塞。**维持 APPROVE**。感谢 yujiawei 抓出 mention 语法冲突这条。

Jerry-Xin (approved):
# Re-review — PR #971 (octo-web) ✅ APPROVE

Re-reviewed at head `6f644c2f2c4a` against merge-base `300394d5` on `main`. Byte-verified the delta from the prior head `dd3721e6`.

## Honesty note — my prior APPROVE (dd3721e6) was an under-call

I approved at `dd3721e6`. **@yujiawei's blocking finding was real and I missed it** — full credit. The prior serializer emitted `[label](url)` with a raw `:` in the label, and both send-side call sites in `packages/dmworkbase/src/Components/MessageInput/index.tsx` flatten adjacent text/link nodes with **no separator** (`pendingTextParts.join("")` and direct `result +=`). So a literal `@` text node immediately followed by a colon-bearing linked label produced `@[prefix:rest](url)`, which the send-side mention scanner `/@\[([^:]+):([^\]]+)\]/g` (`mentionSendParse.ts:81`) matched as a mention token — corrupting the link and, if the prefix equaled a real member uid, silently routing an unintended member mention. Colon-in-label is the common paste case, so this was not exotic. My security-only pass checked scheme/label/paren escaping in isolation and did not round-trip the serialized output back through the mention parser across the node boundary. That was the gap.

## Both prior findings are fixed at 6f644c2f

**(A) Mention-grammar collision — FIXED** (`mentionSendParse.ts:207-217`). `escapeMarkdownLinkLabel` now encodes the colon as a CommonMark character reference (`:` → `&#58;`) instead of leaving it literal. Because the mention scanner requires a literal `:` inside `@[…:…]`, removing the literal colon makes the token grammatically unformable across the node boundary — the link stays intact and no member mention can fire. I independently round-tripped `@` + serialized colon-bearing label through `parseSendMentionText`: no token forms, content passes through inert. Escaping `]` alone would not have been sufficient (the scanner does not honor backslash escapes), so encoding the colon at the source is the correct robust fix, not just a patch on `]`. The renderer still shows a normal `:` since the character reference decodes on display.

**(B) Label escaping gaps — FIXED** (`mentionSendParse.ts:217`). The backslash-escape set now includes `|`, `<`, `>` alongside the prior `` \ [ ] ` * _ ~ ``, so a link label can no longer break GFM table cells or collide with autolink/raw-HTML handling on the receiver.

**Escape ordering is correct.** `&` is encoded first (`&` → `&amp;`), so the subsequently-introduced `&#58;` cannot be double-encoded, and a user-typed literal `&#58;` is neutralized to `&amp;#58;` (won't decode back into a colon and re-open the collision). Verified by direct evaluation.

## No regression on prior-verified surfaces
- `isSafeUrl` scheme whitelist (http/https/mailto→text) — file unchanged in this delta; dangerous-scheme / non-URL still degrade to visible text.
- URL paren normalization (`(`/`)` → `%28`/`%29`) unchanged.
- Plain-text paste and the draft/read (`stripTrustMark`, untrusted) path unchanged — `index.tsx` is byte-identical between the two heads.
- Forged trust mark is still stripped before serialization (new test asserts the trust mark never survives into a linked node's output).

## Verification
- `pnpm --dir packages/dmworkbase exec vitest run src/Components/MessageInput/__tests__` → **96/96 passed** (was 92; +4 new: literal-`@` + colon-label round-trip, escaped-`]` round-trip, `|`/`<`/`>` escaping, forged-trust-mark strip).
- `mergeable: MERGEABLE`. CI checks green/skipped (Build in progress at review time; no failures).

Good fix. Encoding the colon at the serializer boundary neutralizes the collision at the root rather than chasing individual metacharacters, and the added round-trip tests close exactly the gap that let this through. LGTM.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #971 (octo-web)

## Summary

This PR adds send-boundary serialization for Tiptap text nodes carrying Link marks. The new `serializeEditorTextNodeForSend` helper in `mentionSendParse.ts` reads a text node's `link` mark, validates the target via `isSafeUrl`, escapes Markdown-significant characters in the visible label (including colon → `&#58;` to prevent mention-grammar collision), normalizes parentheses in the URL, and emits `[label](url)`. Unsafe targets fail closed to plain text. The two send paths (`extractOrderedBlocks` and `extractMentionsFromEditor` with `trusted=true`) are both wired to the new helper, while the draft/read path correctly stays on `stripTrustMark`.

The serialization logic is well-crafted — security, escaping, and grammar-collision prevention are all handled correctly. One P2 item below concerns the production wiring: the editor does not currently have the Tiptap Link extension, so Link marks will not appear in the editor JSON at runtime.

## Verification

- ✅ **Security boundary** — `isSafeUrl` gates URL protocol to http/https only. `new URL(href)` cannot throw because `isSafeUrl` already succeeded with the same string. `stripTrustMark` runs before any link serialization, so forged trust marks in pasted text are removed first.
- ✅ **Grammar collision prevention** — Colon is encoded as `&#58;` inside link labels, preventing the `@[uid:label]` mention regex from matching when an `@` text node is immediately followed by a link node. Tests confirm this with both adjacent-`@` and escaped-`]` scenarios.
- ✅ **Dual-path parity (C1)** — Both send paths (`extractOrderedBlocks` at `index.tsx:143` and `extractMentionsFromEditor(editor, true)` at `:358`) now call `serializeEditorTextNodeForSend` for text nodes. The draft/read path at `:867` correctly uses `stripTrustMark` only.
- ✅ **Graceful degradation** — When no Link mark exists (the current runtime state), the function returns `stripTrustMark(text)` — identical to the old behavior. No regression.
- ✅ **Downstream consumers** — `formatMentionTextV2` / `parseSendMentionText` only match `@[uid:label]` patterns; Markdown link syntax `[label](url)` passes through untouched. `stripInvisibleChars` only strips zero-width Unicode, not Markdown syntax characters.
- ✅ **Test coverage** — 8 new test cases cover: link preservation, Markdown escaping, parentheses normalization, grammar collision, trust-mark stripping, unsafe URL fail-closed, and plain text passthrough.

Static analysis only at head `6f644c2f`; build and tests not executed in this environment.

## Findings

### P2 — Link serialization is correct but unreachable without the Link extension (`mentionSendParse.ts:222`)

`serializeEditorTextNodeForSend` reads `node.marks` for a `type: "link"` mark, but the editor at `index.tsx:501` does not include `@tiptap/extension-link` in its extensions list, and `packages/dmworkbase/package.json` does not list that dependency. Without the Link extension, ProseMirror's schema strips `<a>` tags during paste normalization, so `editor.getJSON()` never produces text nodes carrying Link marks. The new function always falls through to `return text` (the old `stripTrustMark` behavior), and the pasted-link preservation described in the PR title does not activate at runtime.

Tests pass because they hand-construct `EditorTextNode` objects with Link marks, bypassing the editor schema — the "tests cover helpers but not the production orchestration path" pattern.

No regression: the function degrades gracefully to plain text when no Link mark exists. To activate the feature, add `@tiptap/extension-link` to the editor extensions and configure `Link.configure({ ... })` at `index.tsx:501`. If that wiring is planned in a follow-up PR, consider noting it in the PR description to set expectations.

- `isSafeUrl` implementation at `Utils/security.ts:5` — try/catch around `new URL()` with http/https protocol allowlist. Consistent with 30+ other call sites across the codebase.
- `escapeMarkdownLinkLabel` at `mentionSendParse.ts:207` — escapes `\`, `[`, `]`, backtick, `*`, `_`, `~`, `|`, `<`, `>` plus `&` (encoded first to avoid double-encoding) and `:` (as `&#58;`). Complete for CommonMark link-label context.
- `new URL(href).href` normalization at `:237` — cannot throw because `isSafeUrl` already validated the same string. Parentheses encoding (`%28`/`%29`) prevents Markdown link-destination breakage.
- Import and export wiring at `index.tsx:300` — `serializeEditorTextNodeForSend` is correctly imported alongside existing exports.
- The `!text` early return at `:233` — empty text nodes serialize to empty string even with a valid Link mark, which is correct (invisible link).

## Verdict: APPROVED

The serialization logic is correct, security-conscious, and well-tested. The P2 (missing Link extension in the editor) means the new feature won't activate at runtime without additional wiring, but there is no regression — the function degrades to the existing `stripTrustMark` behavior. If the Link extension is being added in a follow-up PR, this is a clean first half.

[Octo-Q] verdict: APPROVE — no P0/P1 issues. Serialization logic is correct and well-tested; security boundary (`isSafeUrl`), grammar-collision prevention (`&#58;`), and dual-path parity are all verified. One P2 on missing `@tiptap/extension-link` in the editor config (production wiring gap, not a regression).

botshen:
按当前审核状态关闭此 PR；代码与分支保留，将从同一 head 重新创建 PR 触发新的审核流程。

yujiawei (changes requested):
# Code Review — PR #971 (octo-web)

Reviewed independently at head `6f644c2f2c4a8abcbb721e2c0672faa4a9034f35` (base `main`). 3 files, +151/−2. Tests re-run locally: `MessageInput/__tests__` → **96/96 pass**. Production build path unchanged.

The change serializes Tiptap `Link` marks into Markdown `[label](url)` at the send boundary (`serializeEditorTextNodeForSend`) so pasted rich-text hyperlinks keep their target, failing closed to visible text for unsafe/malformed URLs. The send/draft split (`trusted` gate) and the `stripTrustMark` broadcast-forgery guard (#330) are preserved correctly. The follow-up commit correctly hardened the **label** side of the `@[uid:label]` mention grammar (`:` → `&#58;`, escaping `[] \` `* _ ~ | < >`).

However, that hardening was applied only to the label. **The URL destination is still emitted with `[`, `]`, and `:` literal**, so the same re-parse collision the commit set out to close is only half-closed and remains reproducible.

## 1. Spec compliance

**Spec: ✅** — Implements #871 (Link marks → Markdown at send), shares one serializer between plain text and image/file `orderedBlocks`, leaves mention/draft read paths unchanged. No over-build, no scope deviation. (The incomplete-fix concern below is a correctness defect, tracked under Code Quality, not a spec-scope violation.)

## 2. Code Quality

### 🔴 P1 — URL destination is not escaped for the mention grammar; brackets/colons in the href collide with the send-side re-parser

`escapeMarkdownLinkLabel` (`mentionSendParse.ts:207-215`) sanitizes the **label** so it cannot form `@[uid:label]`. But `normalizedHref` (`mentionSendParse.ts:237-240`) only percent-encodes `(`→`%28` / `)`→`%29` — it leaves `[`, `]`, and `:` untouched. The serialized string is then re-scanned in full (label **and** URL) by `parseSendMentionText`'s regex `/@\[([^:]+):([^\]]+)\]/g` (`mentionSendParse.ts:81`), both on the `extractMentionsFromEditor(editor, true)` path (`index.tsx:979`) and the `extractOrderedBlocks` path (`index.tsx:143`). Because `new URL(href).href` keeps `@`, `[`, `]`, `:` literal in a URL's path/query/fragment, a hyperlink whose URL contains those characters escapes the link syntax and is consumed by the mention re-parser.

Reproduced against the compiled module (added a throwaway spec calling the real exports, then removed it):

- **Link-target corruption (correctness — reintroduces the exact bug this PR fixes).** A link preceded by a literal `@`, whose URL contains `]` (IPv6 host, or the very common array-style query param `?ids[]=`):
  - `@` + `serializeEditorTextNodeForSend({text:"home", href:"http://[2001:db8::1]/p"})`
    → serialized `@[home](http://[2001:db8::1]/p)`
    → after `parseSendMentionText` → **`@//[2001:db8::1/p)`** (URL and link syntax destroyed).
  - `@` + `…({text:"list", href:"https://api.example.com/s?ids[]=1&ids[]=2"})`
    → serialized `@[list](https://api.example.com/s?ids[]=1&ids[]=2)`
    → **`@//api.example.com/s?ids[=1&ids[]=2)`** (mangled).
- **Unintended mention injection + URL rewrite (security-adjacent).** A pasted URL whose path/query/fragment contains `@[<member-uid>:<label>]` — no preceding text required:
  - `…({text:"click", href:"https://x.com/@[u-alice:hi]"})`
    → serialized `[click](https://x.com/@[u-alice:hi])`
    → `parseSendMentionText(…, [{uid:"u-alice"}])` → content **`[click](https://x.com/@hi)`** with **`mention.uids:["u-alice"]`**.
    The paster never intended to mention `u-alice`, and the sent link target silently differs from the pasted one (a mild link-spoofing / unexpected-notification vector; the sender sees `[click](…@[u-alice:hi])` but sends `…@hi`).

  Note: broadcast sentinels embedded in a URL stay inert (`https://x.com/@[-2:all]` → `@all`, no routing) because the untrusted-sentinel degrade still holds — the #330 broadcast guard is not weakened. The gap is limited to (a) link-target corruption and (b) ordinary member-mention injection.

Why the PR's own tests miss it: the added adjacency tests (`__tests__/mentionSendParse.test.ts:86-97`) exercise the `@`-adjacent and escaped-`]` cases **only with `href:"https://example.com/docs"`**, which contains no `]` and no embedded `@[`. Those inputs cannot trigger the collision, so 96/96 green does not cover the bracket-bearing / `@[`-bearing URL class.

**Suggested direction (author's call on implementation):** apply the same "cannot form the grammar" treatment to the destination that the label already gets — e.g. percent-encode `[`→`%5B` and `]`→`%5D` in `normalizedHref` alongside the existing `()` handling (both are RFC-3986-legal in path/query/fragment and stay clickable). Two caveats worth handling explicitly: (1) an IPv6 **literal host** `http://[::1]/` cannot have its host brackets percent-encoded without breaking the URL — that case likely has to fall back to visible text or be special-cased; (2) consider mirroring the receive/forward path, which already wraps destinations as `[label](<href>)` via `escapeMarkdownLinkDestination` (`Messages/Text/MarkdownContent.tsx:205-215`) — though note angle-bracket wrapping **alone** does not stop the mention regex, so bracket encoding is still required. A regression test with an IPv6/array-param URL and with an embedded `@[uid:label]` URL should accompany the fix.

### P2 / nits (non-blocking)

- **Send vs. forward inconsistency.** The send serializer emits a bare `[label](url)` destination, while the forward/receive path uses angle-bracket destinations and escapes `\`/`>` (`MarkdownContent.tsx:205-215`). Aligning the two would reduce the chance of future destination-escaping drift.
- Otherwise the security posture is sound: `isSafeUrl` fail-closed for non-http(s)/malformed targets (three cases covered), label Markdown-injection escaping, `stripTrustMark` run first on every text-origin path, and `new URL(href)` reached only after `isSafeUrl` already validated it (no throw).

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅ but a P1 correctness/routing defect remains: URL destinations are not escaped against the `@[uid:label]` grammar, so hyperlinks containing `[`/`]` (IPv6, `?arr[]=`) are corrupted on send (reintroducing the loss this PR fixes) and URLs containing `@[member-uid:label]` inject an unintended member mention while rewriting the link target. Escaping the destination for the mention metacharacters (with the IPv6-host caveat handled) plus a covering regression test would clear this.

PR #633, opened 2026-07-10
chore(loop): drop upstream-brand references from source

MrAladdin:
## What

Public OSS repo hygiene: remove upstream-brand references from the Loop
(`dmloop`) source so the package doesn't leak the underlying engine's name to
anyone browsing the repo.

Found via an automated brand-exposure scan (15 hits, classified). This PR
handles the safe, in-scope ones:

- **Comments (18 files)** reworded to neutral, meaning-preserved phrasing
  ("replicate <brand>" → "align with product design"; "<brand> API/contract"
  → "backend API/contract"; etc.). No logic touched.
- **One `sessionStorage` key** renamed to a neutral namespace. Self-contained:
  set/get/removed only within its own module, and it holds a transient CLI-auth
  handoff value, so no migration is needed.
- **One CSS section comment** neutralized.

No **user-facing** strings were affected — the i18n bundle was already
brand-free (all keys resolve to Loop/neutral copy), and the scan found no brand
in any rendered string.

## Deferred (follow-ups)

Left out because they need coordination or belong to a separate feature:

- CLI-auth symbol/file renames (pure-frontend but a distinct feature) — separate rename PR.
- CSRF cookie name — needs the backend to emit the matching name (double-submit).
- API path prefix — needs frontend + dev proxy + backend to move together.

## Verification / gates

Comment-and-string only, with a single self-contained transient storage-key
rename — no logic blast radius. Per the "no-logic-impact" carve-out the
/simplify + adversarial gates are exempt; the one functional touch was verified
self-contained (the key const + all three storage calls live in one module,
no external reader). `tsc` clean; diff is 19 files, all 1:1 line swaps.

Independent of the open Loop PRs (touches comments/strings, mostly different
lines/files); rebases cleanly on the current tip.

PR #972, opened 2026-07-22
fix(space): pass duration to invite API; show expiry; add duration picker

xiaokenbot:
## 问题
Space 邀请码选了「永久」期限，仍然会失效。

## 根因
`SpaceService.createInvite` 调用后端时 body 始终为空对象 `{}`，未传 `duration` 参数，后端按默认有效期处理。UI 上也没有期限选择器。

## 改动

**`SpaceService.tsx`**
- 新增 `InviteDuration` 类型：`0`（永久）| `86400`（1天）| `604800`（7天）| `2592000`（30天）
- `createInvite(spaceId, duration = 0)` 默认永久，传 `{ duration }` 给后端
- `InviteResp` 新增可选 `expire?: string`（ISO-8601；缺省或空 = 永久）

**`SpaceSettings/index.tsx`**
- 生成邀请链接前新增 `Select` 选择有效期（默认：永久有效）
- 生成后展示过期时间或「邀请链接永久有效」

**i18n（zh-CN + en-US）**
- 新增 `inviteDuration.*` / `inviteExpireAt` / `inviteNeverExpires` key


lml2468 (approved):

审查锚定 head `5a14779c`。4 文件 +60/−15,修 Space 邀请码选「永久」仍失效:`createInvite` 之前发空 body `{}` 不传 duration,后端按默认有效期处理。本地跑 build + 测试 + 逐条核过。

### ✅ 两道闸门均通过

**规格符合**:与根因/描述一致 —— `createInvite(spaceId, duration=0)` 现传 `{ duration }`,加期限 Select(永久/1天/7天/30天)+ 生成后展示过期时间或「永久有效」。无 over-build。

**代码质量(逐条核实)**:
- **根因修复对症**:`SpaceService.createInvite` 从 `post(..., {})` 改成 `post(..., { duration })`,`InviteDuration` 类型 `0|86400|604800|2592000`(秒),0=永久;
- **向后兼容**:其余调用点(SpaceCreate/SpaceList/SpaceMembers)仍 `createInvite(spaceId)` 无 duration → 默认 `0`=永久 —— 正是本 bug 期望的默认,不破坏;docs 的 `createInvite`(docs/invite/api.ts)是**另一个函数**、不受影响;
- **`expire=''` 处理正确**:后端永久时返回空串,`state.inviteExpire ? 过期时间 : 永久有效` —— `''` falsy 走「永久有效」分支,与后端契约(空=永久)一致;
- **过期展示**:`new Date(expire).toLocaleString()` 本地化显示(expire 为 ISO-8601,已由 `?` 分支保证非空才 parse);
- **Service 层 WKApp**:`SpaceService.tsx` 用 `WKApp.apiClient` 是既有 grandfather 模式(本改动未新增 WKApp 耦合)。

### ✅ 验证

- `pnpm --filter @octo/web build` → **`✓ built in 3.94s`**(InviteDuration 类型 + Select 类型检查通过;仅既有 semi-ui warning);
- `@octo/base` SpaceService/SpaceSettings 相关 → **25/25**;
- **i18n 齐全**:6 个新 key(`inviteDuration.*`/`inviteExpireAt`/`inviteNeverExpires`)双 locale 全在,en 1280 / zh 1280 parity(node 核实);
- 已 rebase(behind 7 无关 commit)、`git merge-tree` **0 冲突**、MERGEABLE。

### 结论

对症修复:根因定位准(空 body 不传 duration)、修法直接(传 `{duration}`,0=永久默认)、向后兼容(其余调用点默认永久)、`expire` 空串正确回落永久、i18n 双 locale 齐全、build + 测试通过。无阻塞。**APPROVE**。

Jerry-Xin (changes requested):
## Review: fix(space): pass duration to invite API; show expiry; add duration picker

The FE changes are internally clean (types, state, i18n parity), but I traced the picker value → API payload → **backend contract** and this PR **does not actually fix the stated bug** and adds a misleading "never expires" display. Requesting changes.

### 🔴 Blocking

**1. Backend `createInvite` ignores `duration` entirely — the fix is a no-op / bug is not fixed.**

The FE now sends `POST space/{id}/invite` with body `{ duration }` (`SpaceService.tsx` `createInvite`). But the backend handler for this route (`octo-server` `modules/space/api.go`, `func (s *Space) createInvite`) **never binds or reads any request body** — there is no `BindJSON`, and the token `duration` does not appear anywhere in that module. Instead it unconditionally calls `applyAutoInviteDefaults(inviteModel, time.Now())`, which sets `expires_at = now + 72h` (`invite_code.go`: `inviteDefaultTTL = 72 * time.Hour`).

Consequences:
- Selecting **「永久」(duration=0)** does **not** produce a permanent invite — the code still expires in 72h. This is exactly the bug the PR claims to fix; a FE-only change cannot fix it because the server drops `duration`.
- The `1day / 7days / 30days` options are equally ignored — everything gets the fixed 72h TTL.

The server-side change (accept `duration` in the create body and map it to `expires_at`, with `0` → no expiry) is a prerequisite. Please land the backend change first (or in tandem) so the contract this PR depends on actually exists.

**2. `expire` is never returned by the backend → expiry display is always wrong ("永久有效").**

`InviteResp` now declares optional `expire?: string`, and the UI shows `inviteExpireAt` / `inviteNeverExpires` based on `resp.expire`. But the backend `createInvite` response is `{"invite_code": ...}` only — no `expire` (and no `invite_url`). So `resp.expire` is always `undefined`, and the UI will **always** render "邀请链接永久有效 / never expires" — which is factually false, since the invite currently expires in 72h. This is a user-facing correctness bug on top of #1.

### 💬 Non-blocking

- The FE unit values (0 / 86400 / 604800 / 2592000 = seconds) are self-consistent and the `InviteDuration` type documents seconds clearly — good. Just confirm the eventual backend contract also treats `duration` as **seconds** (and `0` as "no expiry"), so the unit doesn't drift once the server side lands.
- Expiry render uses `new Date(resp.expire).toLocaleString()` — once the backend returns a real ISO-8601 value, double-check TZ/locale rendering is what you want; `toLocaleString()` uses the browser's local TZ.

### ✅ Highlights

- i18n parity is correct: all 6 new keys (`inviteDuration.permanent/1day/7days/30days`, `inviteExpireAt`, `inviteNeverExpires`) are added at parity in both `en-US.json` and `zh-CN.json`.
- `duration` defaults to `0` and is optional, so the call signature stays backward compatible.
- Clean state handling and no dead code in the FE diff.

**Verdict: REQUEST_CHANGES** — the create endpoint does not accept `duration` and does not return `expire`, so the invite still expires in 72h regardless of the picker, and the "never expires" text is shown incorrectly. Needs the backend contract (`duration` in create body → `expires_at`, `0` = no expiry; return `expire`) to actually fix the reported problem.

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — no P0/P1 findings. PR is a clean, well-scoped feature addition with correct data flow across all changed files. One CSS hygiene nit (non-blocking).

yujiawei (changes requested):
# Code Review — PR #972 (octo-web)

**Head SHA:** `5a14779c2af4a161e95d2d1b7c850c26472c260b`

## 1. Spec compliance

**Spec: ❌**

The stated goal is: "Space invite codes chosen as *permanent* still expire → send `duration` to the backend so the chosen validity is honored, and surface the expiry to the user." The diff wires up the client side cleanly, but it is built on a **backend contract that does not exist**, so the bug it claims to fix is not fixed.

- **Deviation (core):** `SpaceService.createInvite` now POSTs `{ duration }` to `space/{spaceId}/invite` (`packages/dmworkbase/src/Service/SpaceService.tsx:299-301`) and reads `resp.expire` (`InviteResp.expire`, line 258). The endpoint that actually serves this route — `POST /v1/space/:space_id/invite` → `func (s *Space) createInvite` in the backend — **never reads the request body**. It builds the invite unconditionally via `applyAutoInviteDefaults(inviteModel, time.Now())` and responds with `{"invite_code": ...}` only. There is no `duration` parsing and no `expire`/`expires_at` field in the response. The `duration` payload is silently dropped; `resp.expire` is always `undefined`.
  - Consequence: choosing "Permanent" (or any option) has **zero effect** — every invite still gets the server default TTL (72h unless `DM_SPACE_INVITE_DEFAULT_TTL` overrides it). The original bug ("permanent still expires") remains fully present after this PR.
- **Contract mismatch (even if the backend were wired):** the invite subsystem that *does* support custom expiry (the manager endpoint) takes an **absolute `expires_at` timestamp**, not a `duration` in seconds, and returns `expires_at` (not `expire`). So the invented `{ duration }` / `expire` shape does not match the backend's actual convention either — a future backend implementation following existing patterns would not satisfy this client as written.

Because the fix does not achieve its stated purpose, and additionally **ships UI that actively misinforms** (a duration picker that does nothing, plus an expiry line that will always render "never expires" while the invite in fact expires in ~72h), Spec fails.

## 2. Code Quality

### P0 — Fix is a no-op + ships misleading UI
As detailed above. Two user-visible harms:
1. The duration `Select` (`SpaceSettings/index.tsx:236-247`) has no backend effect.
2. Since `resp.expire` is always `undefined`, the result panel always shows **"Invite link never expires"** (`index.tsx:271`, `inviteNeverExpires`) — but the invite actually expires at the default TTL. This is worse than the status quo: it converts a silent bug into a confidently-wrong UI claim. This must be blocked until the backend accepts `duration` (or the client is switched to the real `expires_at` contract) and returns the effective expiry.

### P2 — `Invalid Date` can leak into the UI
`index.tsx:270` renders `new Date(this.state.inviteExpire).toLocaleString()` with no validity check. If a non-empty but unparseable string ever comes back, the UI prints the literal `"Invalid Date"`. Guard with `const d = new Date(expire); Number.isNaN(d.getTime())` and fall back. (Advisory leg concurs.)

### P2 — Picker locks out after first generation
Once `inviteCode` is set, the whole form (including the duration `Select`) is replaced by the result panel (`index.tsx:234-266`) with no "generate another" / reset affordance. Pre-PR this was a minor quirk; with a duration choice it becomes a real lock-out — a user who picks the wrong duration cannot re-pick without closing and reopening the pane to clear local state.

### P2 — New CSS classes have no styles
`wk-spacesettings-invite-form` and `wk-spacesettings-invite-expire` are referenced in the JSX but not defined in `SpaceSettings/index.css` (only `-invite-result`, `-invite-row`, `-invite-label`, `-invite-code` exist). Harmless, but the expiry text will be unstyled.

### Notes (non-blocking, correct as written)
- i18n: both `zh-CN` and `en-US` gained the six new keys with matching `{{expire}}` interpolation, and the runtime `t()` supports `{ values: { expire } }` interpolation. ✅
- TS: `InviteDuration` union and the `as InviteDuration` cast on `Select.onChange` are consistent. ✅

## 3. Overall verdict

**CHANGES_REQUESTED**

The client-side code is tidy, but the change does not fix the reported bug and introduces a UI that misrepresents invite expiry. This needs the backend `POST /:space_id/invite` handler to (a) parse the chosen validity and (b) return the effective expiry — and the client/server field names to agree (`duration`+`expire` vs the existing `expires_at` convention) — before it can be merged. Please land the backend counterpart first (or in the same coordinated change) and re-point the client at the real contract.

## 4. Suggestions
- Implement backend support in `POST /:space_id/invite`: read the validity from the body and set `InvitationModel.ExpiresAt` accordingly (mirror the manager handler's logic), and include the effective expiry in the response. Align on one contract — reusing the existing `expires_at` absolute-timestamp shape is lower-risk than introducing a parallel `duration`/`expire` pair.
- Add a validity check before formatting `inviteExpire`.
- Add a "generate another link" reset (clear `inviteCode`) so a mis-picked duration is recoverable.
- Add styles for the two new CSS classes.

## 5. Additional observations
- The public preview endpoint `GET /v1/space/invite/:invite_code` already returns `expires_at`; if a coordinated backend change is not feasible immediately, one interim option is to have the client read the effective expiry from the preview endpoint after creation — but the picker still cannot work without backend `duration`/`expires_at` support on create, so this is only a partial mitigation for the "show expiry" half.

cnwenf (changes requested):
# Code Review — PR #972 (Mininglamp-OSS/octo-web)

## Summary

This PR fixes the Space invite-duration bug by passing `duration` in the `createInvite` POST body (previously always `{}`), adds a duration picker (`Select`, default permanent) and an expiry / "never expires" display to SpaceSettings, and adds the matching i18n keys in both locales. The frontend change itself is clean and well-scoped — i18n parity confirmed, the `Select` option set is exactly the `InviteDuration` union, and all three reviewers cleared security. The central concern, raised independently by two reviewers, is that the fix depends on an unmerged backend change (octo-server #633): one reviewer read octo-server HEAD and found the invite handler ignores the body entirely and always applies a 72h TTL and never returns `expire` — so merged alone this PR neither fixes the stated bug nor gets a real expiry, and the new UI would display a false 「永久有效」 for a link that actually expires in 72h. That backend dependency (2 P1s) plus the confirmed frontend defects below (malformed-expire → "Invalid Date"; no tests pinning the new payload or display) drive NEEDS_CHANGES. The third reviewer's E2E was BLOCKED because the invite backend lives in a separate repo (octo-server), whose `duration:0` semantics could not be exercised from this checkout.

## Findings

### P1 — Backend ignores `duration`; the stated bug fix is ineffective without a paired backend change (`packages/dmworkbase/src/Service/SpaceService.tsx:299`) [source: octo-review-glm5.2]

`createInvite` now posts `{ duration }`, but at octo-server HEAD the handler (`api.go:904-938`) never reads a body field and always applies a 72h TTL. Selecting 「永久」(duration=0) therefore still produces a 72h-expiring invite — the exact failure this PR claims to fix. This is a cross-repo dependency: the trigger notes octo-server #633 is under separate review (presumably the backend half). Either land #633 first or explicitly gate this PR on it; at minimum the PR description should state the backend dependency rather than implying the frontend change alone fixes it.

### P1 — UI always shows 「邀请链接永久有效」for a link that actually expires in 72h (`packages/dmworkbase/src/Components/SpaceSettings/index.tsx:269`) [source: octo-review-glm5.2]

The new expiry block reads `resp.expire` and falls back to `inviteNeverExpires` when falsy. But the backend returns only `{ invite_code }`, so `resp.expire` is always `undefined` and the truthy guard at `:269` always renders 「永久有效」— including for 72h invites. This false claim is newly introduced by this diff (pr-base had no expiry display). Until #633 returns a real expiry, do not render a 「永久有效」 assertion the frontend cannot substantiate (omit the line, or show the user's configured selection instead).

### P2 — Expire display renders literal 'Invalid Date' for an unparseable backend value (`packages/dmworkbase/src/Components/SpaceSettings/index.tsx:270`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P2, glm5.2=Nit, qwen3.8=P2)

- **Broken contract**: the new display guards only on truthiness (`this.state.inviteExpire ? inviteExpireAt : inviteNeverExpires`) and flows `resp.expire` → `state.inviteExpire` → `new Date(inviteExpire).toLocaleString()` with no parse validation between source and sink. `InviteResp.expire` is `string|undefined` with only a comment promising ISO-8601; nothing enforces it.
- **Reachable consequence**: any non-empty but unparseable `expire` (epoch-seconds, malformed/timezone-less string) passes the truthy check and renders '邀请链接将于 Invalid Date 失效' — garbage text in exactly the UI whose purpose is to show the correct expiry. Requires a backend contract violation to trigger, hence P2, not P1.
- **Evidence**: the verbatim HEAD render expression executed against the real zh-CN template — `'not-a-date'` and `'0000-00-00'` both yield `邀请链接将于 Invalid Date 失效`; controls pass (`undefined`/`''` → `邀请链接永久有效`; valid ISO → a readable date); `pr-base` has no `inviteExpire`/`new Date`/`inviteExpireAt`, so this path is introduced by this diff (`probe_expire_display.mjs`, exit 0).
- **Fix direction**:

```ts
const d = new Date(expire);
// isNaN(d.getTime()) -> fall back to the inviteNeverExpires branch, else d.toLocaleString()
```

…or confirm the backend serialization contract is enforced. (octo-review-qwen3.7's suggested inline guard: `{this.state.inviteExpire && !isNaN(new Date(this.state.inviteExpire).getTime()) ? …inviteExpireAt… : …inviteNeverExpires…}`.)

### P2 — New `{ duration }` payload and expire-display branches are pinned by no test (`packages/dmworkbase/src/Service/SpaceService.tsx:299`, `packages/dmworkbase/src/Components/SpaceSettings/index.tsx:270`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

- **Broken contract**: the PR's core change — `createInvite` POSTs `{ duration }` instead of `{}` — has no test at all, and CI never runs the Vitest suite (`SpaceService.test.ts` has 0 references to createInvite/invite/duration/expire). None of the three expire-display render branches is pinned either — (a) falsy → `inviteNeverExpires`; (b) valid ISO → readable date via `inviteExpireAt`; (c) truthy-but-unparseable → literal 'Invalid Date'. There is no SpaceSettings test file at all.
- **Reachable consequence**: a regression reverting the body to `{}` — the exact reported bug — leaves build, lint, `lint:wkmodal` and `i18n:check` all green; likewise a guard-inverting regression (or a malformed non-empty expire) ships with every CI gate green. DEVELOPMENT.md requires Service methods to be covered for endpoint/body/envelope.
- **Evidence**: mutation proof — the existing suite cannot observe `createInvite`'s body (machine-counted 0 invite references); a focused test asserting `createInvite('sid', 86400)` → `{duration:86400}` and default → `{duration:0}` passes on HEAD and fails on the verbatim pr-base `{}` mutant (`probe_test1_createinvite_covgap.mjs`, exit 0). A guard-inversion mutant survives the existing suite while a focused branch test against the verbatim HEAD expression kills it and pins branch (c) (`probe_test2_expire_display_covgap.mjs`, exit 0).
- **Fix direction**: add a SpaceService unit test stubbing `WKApp.apiClient.post` asserting `createInvite('sid', 86400)` → `{duration:86400}` and default → `{duration:0}`; plus a component/VM test asserting branches (a) and (b), and either the invalid-Date guard or a locked fallback assertion for (c). (octo-review-qwen3.7 also enumerates the 4 cases the tests should cover; octo-review-glm5.2 points to the `docs/invite/api.test.ts` pattern for asserting the posted body.)

### P2 — [blocked-on-external-backend] Default `duration=0` silently changes the wire body of four unchanged callers (`packages/dmworkbase/src/Service/SpaceService.tsx:299`) [source: octo-review-glm5.2 + octo-review-qwen3.8] (severity split: glm5.2=P2, qwen3.8=P2, qwen3.7=Nit)

- **Broken contract**: `createInvite` now ALWAYS POSTs `{ duration }` with default `duration = 0` (permanent). The four unchanged one-arg callers — packages/dmworkbase `SpaceCreate:51`, `SpaceList:80`, `SpaceMembers:52`, apps/web `SpaceCreate:70` — expose no duration picker, yet their wire body switches from `{}` (backend-default TTL) to `{duration:0}` (permanent). The picker/default is declared only for SpaceSettings; the default-arg behavior change imposed on these four entry points is neither described in the PR body nor pinned by any test.
- **Reachable consequence**: per the PR's own root cause, `{}` produced expiring codes (that is exactly why SpaceSettings invites expired) — so if octo-server treats `duration:0` as permanent, these four paths now mint never-expiring invites without any user choice. Whether that is the outcome depends entirely on backend semantics not in this repo.
- **Evidence (client-side confirmed)**: the verbatim HEAD method body against a stub `apiClient.post` gives `createInvite('s1')` → `{duration:0}` and `createInvite('s1', 86400)` → `{duration:86400}`; the base control runs the verbatim `pr-base` body → `{}` (`probe_wire_contract.mjs`, exit 0). **Runtime block evidence**: the isolated Docker stack builds and serves the SPA (HTTP 200), but `POST /api/v1/space/*/invite` → 502 (nginx cannot reach the absent octo-server upstream, which lives in `Mininglamp-OSS/octo-server`, a separate repo not in this checkout) — `duration:0` semantics vs absent-body `{}` is neither confirmable nor falsifiable here.
- **Fix direction**: verify octo-server's invite endpoint semantics for `duration:0` vs `{}`; then either preserve prior behavior for callers that do not choose a duration (send `{}` / omit the field when unspecified) or explicitly declare and migrate those callers and pin the intended behavior with a test. (octo-review-qwen3.7 frames the default as backward-compatible today and raises it as a follow-up nit; it becomes an unreviewed behavior change the moment #633 honors `duration`.)

### Nit — Unsafe `as InviteDuration` cast on Select onChange (`packages/dmworkbase/src/Components/SpaceSettings/index.tsx:238`) [source: octo-review-qwen3.7 + octo-review-glm5.2]

Semi Design's `Select.onChange` callback signature returns `string | number | (string | number)[] | undefined`. The cast `v as InviteDuration` bypasses runtime validation. In practice the fixed `Select.Option` values (0, 86400, 604800, 2592000) make this safe, but a runtime guard would be more robust as defense-in-depth:

```tsx
onChange={(v) => {
    const valid: number[] = [0, 86400, 604800, 2592000];
    const n = Number(v);
    if (valid.includes(n)) this.setState({ inviteDuration: n as InviteDuration });
}}
```

(octo-review-glm5.2 adds: if Semi stringifies the numeric value, the body would carry `{ duration: "0" }` (string); a `Number(v)` guard makes it robust.)

### Nit — Duration `Select` stays interactive while an invite is generating (`packages/dmworkbase/src/Components/SpaceSettings/index.tsx:235`) [source: octo-review-glm5.2]

Only the Generate button is `disabled={inviteLoading}`; the `Select` has no `disabled` prop. No data corruption (value is snapshotted synchronously), but the shown duration can diverge from the submitted one. Add `disabled={this.state.inviteLoading}`.

### Nit — Invite state is never reset if the component is reused across spaces (`packages/dmworkbase/src/Components/SpaceSettings/index.tsx:137`) [source: octo-review-glm5.2]

`inviteCode`/`inviteExpire`/`inviteDuration` are initialized only in the constructor with no `componentDidUpdate`. If a parent keeps `SpaceSettings` mounted across a space switch (not keyed by `space_id`), stale state persists. The parent mount/key strategy is outside the 4 changed files so this is a flag, not a confirmed bug — verify the parent keys by `space_id`, or reset invite state on `props.space.space_id` change.

### Nit — No way to generate a second invite without reopening the panel (`packages/dmworkbase/src/Components/SpaceSettings/index.tsx`) [source: octo-review-qwen3.7]

After generating an invite, the UI switches to the result view (invite code + copy buttons + expiry). There is no "Generate another" button or reset mechanism. Users who want a second invite with a different duration must close and reopen SpaceSettings. Not a bug, but a UX gap worth tracking.

- **i18n parity**: all 6 new keys exist in both `en-US.json` and `zh-CN.json` and are non-empty, with the `{{expire}}` interpolation placeholder intact and matching the `t()` `values` contract; the invite* key sets are symmetric (11==11) and every referenced key resolves in both locales (only 2 locales exist). No missing-translation path found.
- **Duration input domain**: the Semi `<Select>` option values are exactly `[0, 86400, 604800, 2592000]` == the `InviteDuration` union, and they (plus init `0`) are the only UI producers of `inviteDuration` — an out-of-union value cannot reach the API via this entry.
- **Falsy=permanent convention**: empty/undefined `expire` → `inviteNeverExpires` is consistent with the documented `InviteResp.expire` contract comment and verified by probe; the confirmed defect is isolated to the malformed-truthy branch above.
- **Security surface**: no new auth path (reuses `WKApp.apiClient`), `space_id` isolation unchanged, the `isAdmin()` gate preserved, React auto-escaping on the `expire` render (no `dangerouslySetInnerHTML`/`eval`), credential scan clean, no new dependencies. The separate docs `createInvite` API is correctly untouched. All three reviewers: CLEARED (1 Informational hardening note — the frontend duration constraint is a type assertion only; confirm the backend whitelist).
- **Error handling / rollback**: both try/catch paths in `handleInvite` reset `inviteLoading: false`; `inviteCode`/`inviteExpire` are set only on success, so a first-invite failure leaves no stale state.

## Verification Results

| Code Review | NEEDS_CHANGES | 2 P1 (backend ignores `duration` + false 「永久有效」 display, cross-repo dependency on octo-server #633) + 3 P2 (Invalid-Date render, no tests for payload/display, default-0 wire-body change for 4 callers) + 4 nits (deduped across 3 reviewers) |
| E2E Test | BLOCKED | Two reviewers did static analysis only (no live stack); the third ran an isolated Docker build+run (SPA HTTP 200) but `POST /api/v1/space/*/invite` → 502 because octo-server lives in a separate repo — wire-contract harm unadjudicable from this checkout |
| Security | CLEARED | All three reviewers cleared security (0 findings); 1 Informational hardening note — the frontend duration constraint is a type assertion only, confirm backend whitelist |

an9xyz (changes requested):
感谢 PR。但这个改动恐怕修不到报障场景,建议调整方向:

**1. 带「永久」选项的报障界面在 octo-admin,不在 octo-web**

带 7d/14d/30d/never 期限选择的邀请码管理台在 octo-admin(`SpaceInvitesPanel`)。octo-web 应用(`apps/web`)里只有"消费邀请 / 加入"的入口(`InviteLanding` / `JoinSpacePage` / `SpaceGate`),**没有邀请码管理或过期展示界面**。PR #972 之前 octo-web 也从未有过"永久"选项,所以"选了永久仍失效"这个现象不可能来自 octo-web。

**2. 本 PR 改的 `SpaceSettings` 疑似遗留组件,未被渲染**

据代码检索,`SpaceSettings` 仅在 `dmworkbase/src/index.tsx` 被 barrel 导出,未发现任何 `import { SpaceSettings }` 消费方或 `<SpaceSettings>` 渲染点(`SpaceMembers` / `SpaceList` 同样)。若属实,新增的期限下拉和过期展示**用户看不到**。麻烦作者确认该组件当前是否真的挂载在应用里。

**3. 唯一的活跃副作用**

`createInvite(spaceId, duration = 0)` 默认永久后,`apps/web` 建空间处 `SpaceCreate` 的无参调用会发 `{ duration: 0 }`,配合 server #637 使"新建空间时生成的邀请链接"从 72h 变永久。这是个附带效果,并非报障所指的"管理邀请码期限"。

**根因在 server 端**(空 `expires_at` 无法落 NULL,详见 Mininglamp-OSS/octo-server#637 的评审),octo-admin 侧也需同步。建议:修复目标改为 octo-server + octo-admin;若确实还想要 web 端的期限选择器,应先把 `SpaceSettings` 接入应用再谈。

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #980, opened 2026-07-22
feat(chat): allow re-editing recalled messages

### LINKED ISSUE #979: ✨ [Web] 撤回的消息支持回填输入框重新编辑
## 需求描述

消息撤回后，希望能在输入框中重新编辑该条消息的原始内容，方便修改后重新发送，而不用手动重新输入。

## 背景 / 痛点

用户发完消息发现有错字或表述问题，撤回后目前需要从头重新输入全部内容，体验差。主流 IM（微信等）撤回后会把原文回填到输入框，支持二次编辑。

## 期望行为

1. 用户撤回自己刚发的消息
2. 会话输入框自动回填被撤回消息的原始文本内容
3. 用户可在此基础上修改后重新发送

## 补充说明

- 端：Web
- 需确认交互细节：仅文本消息支持回填，还是图文/富文本也支持？回填是否有时间窗口限制（如仅撤回后 X 分钟内可编辑）？
- 待产品/交互确认具体范围

---

📢 反馈来源：沈鑫（DMWork uid:17630a8d72b24adc96db007d73380696）

botshen:
## Summary

- show an explicit **Re-edit** action after the current user recalls their own message
- restore text, mentions, rich-text images, and files into the composer without overwriting the current draft
- handle unavailable attachments safely and add localized feedback plus focused regression coverage

## Scope

This PR only changes the recalled-message placeholder and composer restore path. It does not change recall permissions, recall time limits, or the existing message edit flow.

## Verification

- `pnpm --dir packages/dmworkbase exec vitest run src/Messages/Revoke/reeditableMessage.test.ts src/Components/MessageInput/__tests__/mentionSendParse.test.ts src/Components/MessageInput/__tests__/richTextPaste.test.ts` (55 tests)
- `pnpm i18n:check`
- `pnpm exec stylelint packages/dmworkbase/src/Messages/Revoke/index.css --config stylelint.config.mjs`
- `git diff --check`

Closes #979


lml2468 (dismissed):

审查锚定 head `fc9c1dc0`。9 文件 +631/−4,给用户**自己撤回的消息**加「重新编辑」入口,把文本/@mention/富文本图片/文件还原进输入框(不覆盖当前草稿)。核心是所有权门控 + 还原安全 + 草稿保留,本地跑测试 + build + 逐条核过。

### ✅ 两道闸门均通过

**规格符合**:与描述一致 —— 仅改撤回占位与还原路径,不动撤回权限/时限/既有编辑流。`Closes #979`。无 over-build。

**代码质量 —— 所有权/还原/草稿三面核实**:
- **🔒 所有权双门控**:`canReeditRevokedMessage`(:57)要求 `revoker === myUid **且** message.fromUID === myUid` —— 必须同时是撤回者和原发送者。管理员撤回**别人**消息(`revoker===me` 但 `fromUID!==me`)**不能**重新编辑他人内容;
- **🔒 mention 还原为结构化节点、非裸标记**:重建 Tiptap `mention` 节点(`id: part.data.uid` + label,跳过 `uid==="all"`),数据来自用户**自己**被撤回消息的 mention 部件;非 mention 文本走 `appendTextWithBroadcastMentions` —— 不注入裸 `@[uid:label]` 文本,下次发送仍走可信序列化路径,**不越权、不伪造广播 mention**(#330 边界保持);
- **🔒 草稿保留**:还原走 `editor.chain().focus("end").insertContent(content)` —— **在末尾插入、不覆盖**当前草稿;
- **附件 fail-soft**:image/file `url = url || remoteUrl || ""`,`if (!url) return []` 跳过不可用附件(:72/87);blob 拉取 try/catch(:147)—— 与描述「unavailable attachments safely」一致。

### ✅ 验证

- `@octo/base` reeditableMessage + mentionSendParse + richTextPaste → **55/55 全绿**;
- **i18n 齐全**:reedit/revoke 新 key 双 locale parity(en 1281 / zh 1281,零孤儿);
- `pnpm --filter @octo/web build` → **`✓ built in 3.68s`**;

### 结论

所有权双门控严(撤回者 AND 原发送者,禁改他人)、mention 还原为结构化节点不越权(#330 边界保持)、还原在草稿末尾插入不覆盖、附件不可用 fail-soft;测试 55/55、i18n parity、build 通过。无阻塞。**APPROVE**。

Jerry-Xin (dismissed):
## ✅ APPROVE — byte-verified against head `fc9c1dc0`

Clean, well-scoped feature. Re-edit affordance appears only on a message the current user **both authored and revoked themselves**, restores content into the composer as a new draft, and re-sends through the normal send path. Ownership gating, URL safety, size caps, and i18n parity all check out. Build passes, 55/55 tests pass, mergeable.

### 🔒 Ownership / eligibility gating (primary concern) — correct
`canReeditRevokedMessage` (`packages/dmworkbase/src/Messages/Revoke/reeditableMessage.ts`) enforces a strict double gate:
- requires `message.revoke` truthy (falsy/undefined → blocked),
- `revoker === myUid` **AND** `fromUID === myUid`.

So an **admin-recalled** message (`revoker !== sender`) is **not** re-editable by the sender, and a member's message revoked by someone else is blocked. Verified against the `revoke`/`revoker`/`fromUID` getters in `Service/Model.tsx`. Defense in depth: the affordance in `Messages/Revoke/index.tsx` renders only when the gate passes, and `Conversation/index.tsx#reeditRevokedMessage` re-checks the same gate before acting — so a stale/forged UI event can't bypass it.

Time-window: the PR intentionally does not touch recall permissions or the recall time limit (enforced upstream at recall time). Re-edit only resurfaces content **you already recalled** into **your own** composer; it does not resurrect anyone else's message, so no new eligibility surface is opened.

### 🆔 Identifier correctness — no dup/overwrite
`reeditRevokedMessage` only feeds the composer (`insertContent` / `addAttachment`) then `focus()`. It does **not** reuse `message_id` / `client_msg_no`; the follow-up send is a fresh message through the standard path. No overwrite of the original or duplicate-id risk.

### 🛡️ Content restore & XSS — safe
- Text/mentions are restored as structured Tiptap `JSONContent` nodes (`{type:"text"}`, `{type:"mention"}`), not raw HTML — no `dangerouslySetInnerHTML`, so re-rendered user text can't inject markup.
- Image restore (`imageBlockToPasteFile`) and file restore (`resolveSafeFileUrl` → `isSafeUrl`) both gate to http/https only (rejecting script/data-type URLs), fetch with `credentials:"omit"`, validate MIME, and cap size (images 20MB, files `MAX_REEDIT_FILE_BYTES` = 100MB) with both `Content-Length` and post-download blob-size checks. Filenames are sanitized and length-capped.
- Unavailable attachments degrade gracefully to a placeholder + localized toast rather than failing the whole restore.

### 🌐 i18n — parity OK
Three new keys (`revoke.reedit`, `revoke.reeditAttachmentFailed`, `revoke.reeditFileFailed`) present in both `en-US.json` and `zh-CN.json`.

### 🧹 Dead code — none
All four new exports (`canReeditRevokedMessage`, `getReeditableMessageBlocks`, `remoteReeditFileToFile`, `MAX_REEDIT_FILE_BYTES`) are consumed.

### Verification
- `pnpm --filter @octo/web build` → built successfully (eval/chunk-size warnings are pre-existing dependency noise, not from this PR; no syntax errors).
- `vitest run reeditableMessage.test.ts mentionSendParse.test.ts richTextPaste.test.ts` → **55/55 passed** (localStorage warning is Node env noise).
- CI green; `mergeable: MERGEABLE` (BLOCKED only pending this review).

### 💬 Non-blocking notes
- `getReeditableMessageBlocks` for a plain image message defaults `name` to `"image.png"` while rich-text image blocks pass `name` through unchanged — minor inconsistency, harmless.
- File MIME is derived as `application/${extension}` when a file block lacks an explicit mime; `remoteReeditFileToFile` prefers the fetched blob's real type, so this is only a fallback. Fine as-is.

Nice test coverage on the ownership gate and mention/rich-text restore ordering.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

---

**[Octo-Q] verdict: REQUEST_CHANGES** — P1 阻塞：`reeditRevokedMessage` 循环体内无 per-block try/catch，`addAttachment` reject 或 TipTap `insertContent` throw 会中断剩余 block 恢复，用户看到半填充 composer 且无错误提示。修复方向明确（loop body 加 try/catch + toast），其余授权、数据流、i18n、CSS 均验证通过。

botshen:
Superseded by #988.

yujiawei (changes requested):
# Code Review — PR #980 (octo-web)

Reviewed at head `7d022b247ed726669b6c0712837ddb4f3b93e7c7` (the branch advanced from `fc9c1dc` to `7d022b24` — "continue restoring recalled message blocks" — during review; this verdict is against the current head). Base merge-base `d82645a`.

Verified locally on the current head: `vitest run src/Messages/Revoke/reeditableMessage.test.ts` (5 passing), `pnpm i18n:check` (green), `pnpm --filter @octo/web build` (green).

## 1. Spec compliance

The change implements the linked feature (re-edit a self-recalled message: restore text / mentions / rich-text images / files into the composer). Scope is contained to the recalled-message placeholder and the composer restore path, as described. No unrequested surface (no changes to recall permissions/limits or the edit flow). The self-only gate (`canReeditRevokedMessage`) requires `revoker === myUid && fromUID === myUid`, which is correct.

**Spec: ✅** — feature present, scope contained, no over-build detected.

## 2. Code quality

### P1 — Rich-text mentions in non-leading blocks are misrouted to the wrong user

`packages/dmworkbase/src/Messages/Revoke/reeditableMessage.ts:202` (and `:272`, `:313`, `:332`)

The send path builds `mention.entities` with **global** offsets relative to the whole rich-text plain string, counting each image block as `[图片]` (4 chars): `Components/Conversation/index.tsx:1018` (`offsetMentionEntities(m.entities, plainOffset)`), `:1021` (`plainOffset += block.text.length`), `:1059` (`plainOffset += RichTextImagePlaceholder.length`), `:1084`. These global-offset entities round-trip on the wire (re-injected via `Utils/sendContentProxy.ts`, since the SDK's own `encode()` drops `entities`).

On re-edit, `getReeditableRichTextBlocks` passes the same whole-message `mention` into `buildInlineContent(block.text, …)` **per text block** (`reeditableMessage.ts:202`), and the entity filter `entity.offset + entity.length <= text.length` (`:272`) is checked against that single block's length. For any text block after the first block (or after an image), the global offset exceeds the block length, so the entity is dropped and the code falls to `appendTextWithLegacyMentions` (`:285`), which resets `uidIndex` to 0 for each block (`:332`).

Concrete failure: rich-text message with blocks `"Hi @Alice"`, image, `"Ask @Bob"`.
- Send stores entities at global offsets (Alice ≈ 3, Bob ≈ 3+9+4). `mention.uids = [Alice, Bob]`.
- Re-edit block 1 `"Hi @Alice"`: offset 3 is in range → Alice restored correctly.
- Re-edit block 3 `"Ask @Bob"`: global offset out of range → filtered → legacy fallback restarts at `uids[0]` → the visible `@Bob` node is bound to **Alice's uid**.

Effect: on resend, the user sends a mention that displays "@Bob" but silently notifies/routes to Alice. Silent mis-mention is worse than mention loss and should be fixed before merge. Fix direction: track each block's cumulative global plain offset and rebase entities into block-local coordinates before `buildInlineContent`, instead of passing whole-message entities per block.

### P2 — Size cap runs after the full body is buffered into memory

`packages/dmworkbase/src/Messages/Revoke/reeditableMessage.ts:150`, `:157`

`remoteReeditFileToFile` checks `Content-Length` (`:150`), but `Content-Length` is advisory. When it is absent or understated, the fallback `const blob = await response.blob()` (`:157`) buffers the entire response into memory before the post-hoc `blob.size > MAX_REEDIT_FILE_BYTES` check. A source that streams a very large body without an honest header can spike memory / crash the tab. Since the URL is the user's own previously-sent attachment on the app media host, exploitability is limited, but a capped streaming read (`response.body.getReader()` aborting past the cap) removes the failure mode.

### P2 — Existing draft and restored content are concatenated with no boundary

`packages/dmworkbase/src/Components/MessageInput/index.tsx:872`

`insertContent` always appends at the editor end (`focus("end").insertContent(...)`). This correctly preserves the current draft (good — no overwrite), but draft `"hello"` + restored `"world"` becomes `"helloworld"`. Consider inserting a paragraph/space boundary when the composer is non-empty.

### P2 — Async restoration can interleave with concurrent user input / double-clicks

`packages/dmworkbase/src/Components/Conversation/index.tsx:665`, `:716`; `Messages/Revoke/index.tsx:86`

`restoreReeditableMessageBlocks` mutates the live editor incrementally across awaited downloads. If the user types (or clicks Re-edit again) while an image/file is fetching, later blocks append after the newly typed content, or two restorations interleave. The button has no disabled/in-flight state. A guard (disable the button while a restore is in flight, or stage the restore atomically) would make this deterministic.

### P2 (informational) — `message.content as FileContent` cast inside the file branch

`packages/dmworkbase/src/Components/Conversation/index.tsx:691`

The file branch casts `message.content as FileContent` and resolves the URL from the whole message, which is correct only for a standalone file message (contentType 8). Rich-text `file` blocks are converted to a text placeholder (`reeditableMessage.ts:191`) rather than a file attachment, so a rich-text file would be restored as literal `[文件] name` text and could not be resent. This is not reachable today (the send side never constructs rich-text file blocks — see `Messages/RichText/RichTextContent.ts:14,115` — and re-edit is self-only), but the cast is fragile if rich-text file sending is ever enabled. Worth a comment or an explicit guard.

### Positives

- Self-only gate is correct and unit-tested.
- Image/file fetch uses `credentials: "omit"` + `isSafeUrl`/`resolveSafeFileUrl`, and filenames are sanitized/length-capped — good hygiene.
- Broadcast-sentinel handling (`normalizeMentionUid`, `isBroadcastSentinelUid`) prevents forged clipboard-style broadcast smuggling.
- The new `restoreReeditableMessageBlocks` wrapper (continue-on-error + guaranteed `onComplete`) is a good resilience improvement and is tested.

## 3. Overall verdict

**CHANGES_REQUESTED** — one P1 (silent mention misroute on rich-text re-edit). The P2 items are non-blocking but worth addressing.

## Coverage note

Verified locally: build, targeted unit tests, i18n check, and the mention offset logic (reproduced the per-block vs global-offset mismatch). Confirmed `mention.entities` round-trips via `sendContentProxy`. Not verified: backend media-URL provenance / CORS-auth policy at runtime, and real-browser memory behavior for the unbounded-blob path (static reasoning only). Text (non-rich) messages primarily use the `message.parts` path (`reeditableMessage.ts` `getTextMessageInlineContent`), which is unaffected by the offset bug.

PR #981, opened 2026-07-22
feat(dmworkbase): add server-controlled message reactions

### LINKED ISSUE #983: Connect message reactions to the deployed server API
## Problem / Motivation

PR #929 added the message-reaction UI behind a disabled feature flag and a local mock store. The deployed server contract is now available, but the web client still needs persistent writes, inline reaction parsing, incremental synchronization, realtime CMD handling, and production error behavior.

Without the integration, reactions are lost on refresh and are not synchronized across users or clients. DM writes also need to use the active conversation peer UID and preserve the string message ID to avoid channel mismatches and JavaScript integer precision loss.

## Proposed Solution

- Call `POST /v1/reactions` for emoji toggles.
- Parse inline `reactions[]` and consume `POST /v1/reaction/sync` after `syncMessageReaction` CMD events.
- Reconcile optimistic updates with the authoritative response and map deployed error codes.
- Restrict reactions to supported text-message channels.
- Use the active DM peer channel and `message_idstr`.
- Replace the local mock store and enable the production path.

## Affected Component

Conversation / Chat

## Additional Context

- UI foundation: #929
- Implementation PR: #981
- The server contract is tracked in the implementation workspace and is already deployed.

an9xyz:
## Summary

Connect the existing message-reaction UI to the deployed server contract and make reactions persistent, real-time, and remotely controllable across Web clients.

This is a follow-up to #929, which introduced the UI behind a local feature flag.

## Related Issue

Closes #983.

## Changes

- Add a service boundary for `POST /v1/reactions` and `POST /v1/reaction/sync`, including inline reaction normalization.
- Reconcile optimistic toggles with authoritative server responses and handle `syncMessageReaction` CMD updates.
- Restrict the entry points to supported text-message channels (DM, group, and community topic).
- Use the active conversation peer UID for DM writes and preserve `message_idstr` precision.
- Notify the UI when an earlier sync batch applied changes even if a later batch fails.
- Read `message_reaction.read/write` from `GET /v1/common/appconfig`, with the documented missing-field fallback of read-only.
- Apply runtime config changes without a reload: `read=false` hides reactions; `write=false` hides add controls, disables chip toggles, closes an open picker, and blocks the controller before any API request.
- Normalize the invalid `read=false, write=true` combination to fully disabled.
- Localize quick-picker tooltip and accessibility labels, and remove the demo mock store/local feature flag.

Bot capabilities are explicitly out of scope for this phase and remain a separate server/client integration.

- Affected module(s): `dmworkbase` message rendering and message-reaction feature.
- New or changed user-visible entry point: yes; the existing reaction context-menu and summary add button are backed by the deployed API and gated by server capabilities.
- Shared layer touched: `Messages`, `ui`, and `Service`.
- If shared code changed, impact scope: text-message rendering, emoji metadata reuse, reaction API/config parsing, and conversation CMD synchronization.
- Duplicate entry point checked: yes; the existing context-menu and summary controls remain the only reaction entry points.
- Security boundary: client capabilities control UX and prevent accidental writes; server-side authorization remains authoritative.

## Testing

- [ ] Latest capability change manually verified in a browser

Commands run on head `a0ac4585`:

- `pnpm --filter @octo/base test` — 207 files / 1927 tests passed.
- Focused message-reaction suite — 9 files / 57 tests passed.
- `pnpm i18n:check`
- `pnpm exec stylelint packages/dmworkbase/src/ui/message/MessageReactionSummary/index.css`
- `git diff --check`

Manual verification from the previous PR head: group and DM reaction API flows were exercised with the SSO dev server against `https://im-test.deepminer.com.cn/api`. No additional browser session was run after the latest appconfig capability commit.

Coverage mode was attempted but the repository does not currently include `@vitest/coverage-v8`, so no coverage percentage is claimed.

## Checklist

- [ ] Updated documentation — not required; no public API or user documentation changed


lml2468 (dismissed):

审查锚定 head `230a5fe8`。26 文件 +1909/−487,把 #929 的 reaction UI 接到已上线服务端契约(`POST /v1/reactions` + `/v1/reaction/sync`),持久化 + 跨端实时,并默认开启。本地跑测试 + build + 逐条核过 service/乐观协调/精度/权限四面。

### ✅ 两道闸门均通过

**规格符合**:与描述一致 —— service 边界 + 乐观 toggle 与服务端 reconcile + syncMessageReaction CMD;入口限文本消息频道(DM/群/子区);删 demo mock store。`Closes #983`。无 over-build。

**代码质量(逐面核实)**:
- **Service 层合规**:`MessageReactionService` 走 `APIClient.shared.post("reactions"/"reaction/sync")`(:219/231),**不 import WKApp** —— 符合 DEVELOPMENT.md「Service 层用 APIClient.shared」;
- **feature flag 翻转有意为之**:`DEFAULT_ENABLED false→true`(服务端契约已上线),localStorage 覆盖收窄为 `import.meta.env.DEV` only + `try/catch`(SecurityError 兜底),mock store 已删 —— 生产不再看到 mock 伪数据;
- **乐观协调稳**:`controller.toggle` 乐观切 `optimisticDeleted`,失败 `.catch` 走 `rollbackOwnState`,且 **seq 守卫回滚**(`current.seq > previous.seq` 时不回滚陈旧态,:118-120)—— 并发/迟到响应不误覆盖;错误码映射(channel_access_denied/reaction_unsupported_type/group_disbanded/not_found);pending-key 复合去重(`type\0id\0msg\0emoji`);
- **🔒 message_idstr 精度保留**:`message_id` 全程按 **string** 处理(:130/150/220),**不 Number 强转** —— 64 位雪花 ID 无浮点精度丢失;
- **频道限制**:`supportsReaction` = Person||Group||CommunityTopic(:17-19),非文本频道不出入口;
- **sync 部分失败通知**:`syncController` 在 `changed` 时 `notify()`(:157)—— 早批次已应用的变更即便后续批次抛错也通知 UI。

### ✅ 验证

- 加 MemStorage polyfill 隔离 Node-26 localStorage 后,reaction 全套(Convert/Service/controller/syncController/ReactionPickerOverlay/aggregate/EmojiPanel/EmojiService)→ **89/89 全绿**;
- 那 10 个失败全在 `EmojiService.test.ts`(**本 PR 未改**、构造读 localStorage 缓存的既有环境噪音),polyfill 后转绿、非本 PR;
- `pnpm --filter @octo/web build` → **`✓ built in 3.35s`**;i18n 双 locale 新键齐全;

### 结论

分层干净(service/feature-controller/ui)、Service 走 APIClient.shared 不耦合 WKApp、乐观协调有 seq 守卫防陈旧覆盖、message_idstr 按 string 保精度、频道限制正确、sync 部分失败仍通知、flag 翻转有意且 dev override 收窄;测试 89/89、build 通过、无冲突。#929 的 UI 现接上真实服务端。无阻塞。**APPROVE**。

Jerry-Xin (dismissed):
**APPROVE** — Server-backed message reactions are wired correctly, with authoritative reconciliation, own-reaction-only mutation, safe rendering, and clean dead-code removal. Byte-verified against the live head.

## Verified (byte-level)

**Add/remove ownership & authority** ✅
- `Service/MessageReactionService.ts` `toggle()` posts only `{message_id, channel_id, channel_type, emoji}` — no actor uid in the payload. The server infers the reactor from the authenticated session, so a client cannot toggle on behalf of another user.
- `features/messageReaction/controller.ts` optimistic apply and rollback both go through `findOwnReaction(reactions, user.uid, emoji)`, i.e. they only ever mutate the current user's own emoji record. Concurrent records from other users are preserved (covered by the "rolls back only the current user's emoji" test).

**Identifiers** ✅
- `Service/Convert.ts` prefers `message_idstr` over the unsafe numeric `message_id` (precision test proves the 19-digit id survives). `channel_id`/`channel_type`/`emoji`/`seq` are threaded through with strict wire normalization (`normalizeToggleResult`/`normalizeSyncedReaction` reject records missing any required field).
- DM writes use the active peer channel (`context.channel()`) rather than the incoming message channel — verified by the dedicated controller test.

**Optimistic update + reconciliation + rollback** ✅
- Apply optimistic → await server → `applyOwnState(..., result.isDeleted, result.seq)` reconciles to the authoritative seq; on rejection `rollbackOwnState` restores the previous own-record (or removes the phantom if there was none), then surfaces a mapped error toast. No phantom left on failure.
- Double-tap / add-then-remove race is guarded by a per-`(channelType,channelID,messageID,emoji)` pending map that dedupes concurrent toggles (verified test).
- Stale-seq protection: `mergeMessageReaction` refuses to let an older seq overwrite a newer state.

**Cross-user real-time sync — real server path, not an assumed emitter** ✅
- `syncMessageReaction` is consumed through the existing `WKSDK...addCMDListener` in `Components/Conversation/vm.ts`, the same server-push CMD channel that already drives `messageRevoke` and `syncMessageExtra`. It is not a fabricated client-only event.
- `syncController.ts` gates strictly on active channel + positive seq, syncs from the highest known inline seq, drains multiple batches, and still notifies the UI when an earlier batch applied before a later batch fails (verified test).

**XSS** ✅
- Reactor names and participant summaries render as React text children (auto-escaped) in `MessageReactionSummary/index.tsx` — no `dangerouslySetInnerHTML`.
- Custom-emoji `` resolves through `EmojiService.resolveUrl`, which whitelists http(s)/protocol-relative and `data:image/` only; anything else is coerced to a harmless relative path. Manifest items are sanitized. No unsafe scheme reaches the DOM.

**State isolation** ✅
- Reaction state lives on each message's `octoReactions`; the summary re-renders per `messageId` via the mitt event filtered by `changedMessageId === messageId`. The sync controller is per-`ConversationVM` (per channel) and filters records by channel + messageId, so one message's reactions cannot leak onto another or across channels.

**N+1 / perf** ✅
- Inline reactions arrive with the message payload (`Convert.toMessage`); real-time deltas use a single batched `reaction/sync` by channel+seq. No per-message request fan-out.

**No regression / dead code / i18n** ✅
- Demo mock store and the old summary `ReactionSlot` are removed with no dangling references. i18n keys are at parity (en-US ↔ zh-CN, no missing keys either direction). Feature flag now defaults on, with the server described as the final authority for permission/visibility/rate-limit.

## Checks
- Targeted reaction unit tests: 34/34 passing (Convert, Service, controller, syncController, aggregate).
- `pnpm build`: success (2/2 tasks; the `eval` warnings come from bundled third-party libs, not this change).
- CI green; mergeable (BLOCKED only pends this review).

## Non-blocking notes
- The "More" full emoji panel does not receive `selectedKeys`, so already-selected reactions are not highlighted there and re-clicking silently toggles off — a documented UX gap, acceptable for this integration scope.
- Consider a brief comment on the server being the sole ownership authority near `toggle()` for future readers.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #981 (octo-web)

Reviewed at head `230a5fe8`. Scope: connect message reactions to server — new `MessageReactionService`, reaction controller with optimistic updates, realtime sync controller via CMD, UI integration in `ReactionSlot` and `ReactionPickerOverlay`, feature flag promotion to default-on.

## Summary

This PR replaces the demo-only mock reaction store with a full server-backed implementation. The architecture is clean: `MessageReactionService` handles the wire contract (`reactions` toggle, `reaction/sync` batch fetch); `createMessageReactionController` provides optimistic add/reconcile/rollback with dedup; `createMessageReactionSyncController` drains realtime CMD announcements into local state with cursor advancement. The old `mockStore`, demo `ReactionSlot`, and mock event bus entry are all removed. Feature flag `DEFAULT_ENABLED` is promoted from `false` to `true`. Test coverage is comprehensive across all new modules.

## Verification

- ✅ **Data flow: inline reactions** — `Convert.toMessage` at `:305` calls `normalizeMessageReactions(msgMap["reactions"])` → sets `message.octoReactions`. Test `MessageReactionConvert.test.ts` covers presence, absence, and `message_idstr` precedence.
- ✅ **Data flow: toggle** — UI calls `messageReactionController.toggle(message, emoji, channel)` → `MessageReactionService.toggle` POSTs to `"reactions"` via `APIClient.shared` → `normalizeToggleResult` validates wire shape → controller applies server-authoritative `seq`/`isDeleted` or rolls back on error.
- ✅ **Data flow: realtime sync** — CMD `syncMessageReaction` → `messageReactionCommandSeq` validates command name + channel match + positive seq → `syncController.request(seq)` drains batches from `reaction/sync` → `mergeMessageReaction` applies with seq ordering → `notify()` triggers UI re-render.
- ✅ **Optimistic update correctness** — `applyOwnState` at `controller.ts:91` toggles the local user's reaction immediately; `.then()` reconciles with server `seq`; `.catch()` calls `rollbackOwnState` which preserves concurrent reactions from other users (test verified).
- ✅ **Seq-based merge prevents stale overwrites** — `mergeMessageReaction` at `MessageReactionService.ts:184` rejects `incoming.seq < existing.seq`; both toggle controller and sync controller use this consistently.
- ✅ **Dedup of concurrent toggles** — `pending` Map keyed by `channelType\u0000channelID\u0000messageID\u0000emoji` at `controller.ts:141` ensures only one in-flight request per message+emoji. `.finally()` deletes the key.
- ✅ **Sync cursor advancement** — `syncUntil` at `syncController.ts:109` tracks `cursor = nextSeq` after each batch; breaks on empty batch or non-progressing seq; `drain` loop continues until `requestedSeq` is satisfied.
- ✅ **Error handling** — Toggle errors mapped to i18n keys (`getReactionErrorKey` covers 4 known server codes + default); sync errors propagate to `vm.ts:944` `.catch()` → `console.error`. Partial sync results are preserved and notified before error propagation (test: "notifies after applying an earlier batch when a later batch fails").
- ✅ **Deleted file cleanup** — No remaining references to `MessageReactionSummary/ReactionSlot`, `MessageReactionSummary/mockStore`, `reactionMockStore`, or `message-reaction-mock-updated` event.
- ✅ **Feature flag symmetry** — All reaction entry points (ReactionSlot render at `Text/index.tsx:223`, context menu at `module.tsx:921`, pointer tracking at `module.tsx:955`) gated by `isMessageReactionEnabled()`. Channel type guard `isMessageReactionChannelSupported` added at both UI render and context menu paths.
- ✅ **i18n completeness** — 12 new keys added to both `en-US.json` and `zh-CN.json`; picker emoji labels use `nameKey` + `localizePickerEmojis()`.
- ✅ **Input shape validation** — `normalizeMessageReaction` drops records with empty `uid` or `emoji`; `normalizeToggleResult` throws on missing required fields; `normalizeSyncedReaction` drops records missing `messageId`/`channelId`/`channelType`.

## Findings

### P2 — `maxKnownSeq()` may transiently include optimistic seq during concurrent toggle (`packages/dmworkbase/src/features/messageReaction/syncController.ts:79`)

`maxKnownSeq()` scans all messages' `octoReactions` for the highest seq. If a toggle is in-flight with an optimistic reaction whose `seq` was inherited from a previous reaction (via `seq: seq ?? existing?.seq` at `controller.ts:101`), the sync controller could start from a higher-than-real cursor. The server's authoritative response will correct the state, and the next CMD would re-sync from the right point, so this is a transient inconsistency window rather than a data loss risk. Consider passing the sync controller a seq source that excludes optimistic entries (e.g., only counting reactions with server-assigned seq).

### P2 — `getMessages()` allocates a new array on every sync call (`packages/dmworkbase/src/Components/Conversation/vm.ts:175`)

`[...this.messagesOfOrigin, ...this.pendingMessages].map(m => m.message)` creates two intermediate arrays per invocation. `syncUntil` calls `getMessages()` at least twice (initial `maxKnownSeq` + `apply` per record). For conversations with hundreds of messages, this is unnecessary allocation. Not a blocker for typical workloads, but could be optimized with a lazy iterator or cached snapshot.

- `BUILTIN_CUSTOM_EMOJI_KEYS` extraction at `EmojiService.ts:52` preserves identical key/name values; `slice(1, -1)` correctly strips bracket wrappers.
- `ReactionPickerOverlay.openAtLastPointer` signature change (string → object) has no remaining callers using the old signature.
- `aggregateReactions` correctly filters `isDeleted === 1` and sorts by `firstSeq` for stable group ordering.
- `isMessageReactionChannelSupported` covers `ChannelTypePerson`, `ChannelTypeGroup`, and `ChannelTypeCommunityTopic` — matches the deployed server contract.
- `module.tsx` context menu correctly uses `message.messageID` guard (excludes unsent local messages) and `context.channel()` for the reaction request channel.
- `wukongimjssdk-reaction.d.ts` module augmentation correctly extends `Message` with `octoReactions?`.
- Scan config correctly removes exclusion entries for deleted/updated files.
- `types.ts` re-exports are a clean consolidation — single source of truth in `MessageReactionService.ts`.

## Verdict: APPROVED

Well-architected migration from demo mock to server-backed reactions. The optimistic update, seq-based merge, realtime sync, and error handling patterns are all correct and well-tested. The two P2 observations are non-blocking performance/robustness notes.





[Octo-Q] verdict: APPROVE — no P0/P1 found; architecture is sound with proper optimistic update/reconcile, seq-based merge ordering, realtime sync via CMD with cursor advancement, comprehensive error handling, and thorough test coverage across all new modules.

yujiawei (dismissed):
# Code Review — PR #981 (octo-web)

Reviewed at head SHA `230a5fe819d3b543c0be89f5b37d3848eb6ed064` against merge-base `d82645aa`. Verified locally: the 6 reaction test suites (42 tests) pass, `pnpm i18n:check` passes, no dangling references remain to the deleted `mockStore` / old `message-reaction-mock-updated` event, and the `MessageReactionSummary/types.ts` → `Service/MessageReactionService` re-export keeps existing importers (`aggregate`, stories, index) intact.

## 1. Spec compliance

The change delivers exactly what the description scopes, no more:

- **Service boundary** for `POST reactions` (toggle) and `POST reaction/sync` with defensive wire normalization (`normalizeToggleResult` throws on malformed responses; `normalizeMessageReactions` / `normalizeSyncedReaction` drop invalid records) — `Service/MessageReactionService.ts`.
- **Optimistic toggle reconciled with authoritative response**, plus realtime `syncMessageReaction` CMD handling — `features/messageReaction/controller.ts`, `syncController.ts`, wired in `Components/Conversation/vm.ts`.
- **Entry points restricted to supported text channels** (Person=1, Group=2, CommunityTopic=5) via `isMessageReactionChannelSupported`, and further gated to `contentType === text` + a resolved `messageID` in both the context-menu (`module.tsx`) and the render slot (`Messages/Text/index.tsx`).
- **DM peer channel** used for writes (`context.channel()` threaded through `toggle`), and `message_idstr` precision preserved in `Convert.ts` (BigNumber fallback).
- **Partial-batch notify**: `syncUntil` notifies in a `finally` when any earlier batch applied changes even if a later batch rejects (`syncController.ts:157`) — matches the stated behavior and is covered by a test.
- **i18n**: emoji labels / tooltips localized via `nameKey` + `localizePickerEmojis`, `textOnly` error string added, demo mock store removed and its i18n scan-exemptions cleaned up.

No dropped requirements, no unrequested features, no divergence from the stated approach.

**Spec: ✅**

## 2. Code quality

No P0/P1 (no correctness/security/data-loss/build-break, no user-facing functional bug on the normal paths). The normal toggle and realtime paths are correct and well-tested, including reproducers for the concurrency cases below. The findings are all P2 advisories — narrow concurrency interleavings that self-heal on the next CMD or on conversation reload (reactions are re-hydrated inline via `Convert`), plus one operational gap.

**P2 — rollback can clobber a newer synced state when the prior record had no `seq`** — `features/messageReaction/controller.ts:116-124`
`rollbackOwnState` only short-circuits ("keep current") when both `current.seq` and `previous.seq` are numbers. If the pre-existing own reaction came from an inline snapshot without a `seq` (permitted — `finiteNumber` returns `undefined`), and a realtime sync installs an authoritative numbered update during the in-flight toggle, a *failed* toggle falls through to `mergeMessageReaction(reactions, previous)`; since `previous.seq` is undefined, the seq-guard in `mergeMessageReaction` (`MessageReactionService.ts:191-198`) does not fire and the newer synced record is overwritten by the stale one. Suggest also keeping `current` whenever `current.seq` is a number and `previous.seq` is not.

**P2 — equal-`seq` merge lets a replayed/stale sync record overwrite an optimistic state** — `MessageReactionService.ts:191-198`
The guard drops `incoming` only when `incoming.seq < existingSeq`. On equality it overwrites. Combined with `applyOwnState` reusing the old seq for the optimistic entry (`controller.ts:105`, `seq: seq ?? existing?.seq`), a delayed sync record carrying the same seq but the pre-toggle `isDeleted` can revert the optimistic change. Requires the server to re-emit an identical seq with stale state, so it is unlikely under a strictly-monotonic seq contract, but the `<` vs `<=` choice is worth a second look.

**P2 — realtime sync can be abandoned on an empty batch while a higher seq is pending** — `syncController.ts:161-167`
If `request(N)` is in flight and `request(M>N)` bumps `requestedSeq`, but the first `sync` returns an empty batch, `syncUntil` sets `progressed=false` and `drain` returns via `if (!progressed) return` without ever syncing to `M`. The announced seq `M` is then only picked up on the next CMD or on reload. Self-healing, but the drain intentionally tracks `requestedSeq` and this path silently discards it.

**P2 — channel-wide cursor can skip an older message's reaction seq** — `syncController.ts:128-130`
`maxKnownSeq()` is a channel-wide max across all loaded messages. A reaction arriving on an older message with a seq lower than another message's already-known higher seq is treated as "already covered" and not fetched in realtime. Depends on inline snapshots being mutually consistent across paginated loads; correct per-message on fresh load, but stale partial snapshots can miss a realtime update until reload.

**P2 — feature flag defaults ON with no production kill switch** — `Service/featureFlags.ts:15`
`DEFAULT_ENABLED = true` ships the feature to all production users; the only override is DEV-only `localStorage`. The comment acknowledges a rollback requires a new release. Given the server contract is stated as deployed and write failures degrade gracefully (toast + server-side permission/visibility enforcement remain the real boundary), this is acceptable, but a remote-config gate (e.g. `WKApp.remoteConfig`) would remove the "ship a release to disable" risk. Consider as a fast follow.

**Quality: Approved**

## 3. Overall verdict

**APPROVE** — Spec ✅ and no P0/P1. The reconciliation logic is genuinely careful (the numbered-seq rollback path is correct and tested); the remaining items are rare, self-healing concurrency edges and an ops-hardening suggestion, none of which block merge.

## 4. Suggestions

- Harden `rollbackOwnState` for the `previous.seq === undefined` case (keep `current` when it carries a numeric seq).
- Track and re-drain `requestedSeq` after an empty batch in the sync controller, or document why an empty batch is treated as terminal.
- Consider a remote-config-backed kill switch before/soon after wide rollout.

## 5. Additional notes

- The realtime path relies on `notifyListener()` → `Provider.setState()` re-rendering the whole conversation subtree (no `React.memo`/`shouldComponentUpdate` on the text cell), so in-place `octoReactions` mutation does repaint correctly — verified. This is fine today but makes reactions dependent on a full-list re-render; keep in mind if message-cell memoization is ever introduced.
- `getMessages` correctly unions `messagesOfOrigin` + `pendingMessages`, so reactions on messages buffered during pull-up still reconcile.

lml2468 (approved):
## Review Verdict（复审 @ a0ac4585）：可合并（APPROVE）—— 维持 APPROVE(且比上轮更稳)

复审锚定新 head `a0ac4585`。相对我上次 APPROVE 的 `230a5fe8`,本轮把 feature flag 从硬编码 `DEFAULT_ENABLED=true` 升级为**服务端驱动的 read/write 能力模型**,并加 write-gate 防御 —— 比上轮更稳。逐条核过、测试全绿。

### ✅ 本轮改进(核过)

- **服务端驱动能力**:`featureFlags` 改读 `WKApp.remoteConfig.messageReaction ?? DEFAULT`,拆成 `canReadMessageReaction()` / `canWriteMessageReaction()`(write ⟹ read);兑现了上轮注释里「后续替换为 appconfig」的 TODO;
- **read/write 不变量 + 容错**:`parseMessageReactionCapability` 对 appconfig 做字段级容错(`hasOwn`+`parseRemoteBool`),非法 `read=false/write=true` **收敛为全关**(`write: read && write`);注释明确「服务端 reaction API 仍是最终权限边界」——客户端配置只 gate UX;
- **保守默认 `{read:true, write:false}`**:比上轮 `DEFAULT_ENABLED=true` 更稳的灰度(默认只展示、不写,待服务端开 write);
- **write-gate 全接线**:controller dep(`runtime.ts:28 canWrite`)+ `toggle` 早返(`if !canWrite return`)+ 入口(`module.tsx:927/959`)—— read-only 能力下客户端不会发写请求;read-gate 接在渲染(`ReactionSlot:68`)+ 解析(`Conversation/vm.ts:941`)。

### ✅ 上轮结论保持

- **乐观协调 + seq 守卫**:`rollbackOwnState` 仍 `current.seq > previous.seq`(严格 >,equal-seq 不回滚 = 同一权威态,安全);
- **message_idstr 按 string 保精度**、**Service 走 APIClient.shared 不 import WKApp**、频道限制、sync 部分失败 notify —— 均在。

### ✅ 验证

- 加 MemStorage polyfill 后 reaction 全套(10 文件)**67/67 全绿**(含新 featureFlags + ReactionSlot 用例);
- `pnpm --filter @octo/web build` → **`✓ built in 3.68s`**;
- 已 rebase(behind 3)、`git merge-tree` **0 冲突**、MERGEABLE。

### 🟡 非阻塞(沿用 yujiawei)

- `rollbackOwnState` 的 seq 用严格 `>`,equal-seq 的迟到记录不回滚 —— equal-seq 本是同一权威状态,不回滚是安全方向;yujiawei 的并发边界建议属极小概率优化,非阻塞。

### 结论

本轮把能力开关升级为服务端驱动 read/write 模型(容错 + write⟹read 收敛 + 保守默认)、write-gate 全接线防 read-only 误发写;上轮的乐观协调/seq 守卫/message_idstr 精度/Service 解耦全部保持,测试 67/67、build 通过。比上轮更稳。无阻塞。**维持 APPROVE**。

Jerry-Xin (approved):
**APPROVE** — Re-review anchored at head `a0ac4585`. This round replaces the hardcoded local reaction feature flag with a server-driven `{read, write}` capability sourced from `appconfig.message_reaction`, adding defense-in-depth write gating. The change is well-scoped and safe.

## What changed since my prior APPROVE (`230a5fe8`)
Not a rebase — `merge_base == 230a5fe8`, ahead_by=2, behind_by=0. Two real commits (red-coverage tests + capability gating). 14 files. The underlying reaction logic (add/remove, identifiers, optimistic reconcile, cross-user sync, XSS handling) was NOT modified; this delta only layers a capability gate on top.

## Verified critical paths
- **Server-driven capability model** (`Service/MessageReactionConfig.ts`): parses `appconfig.message_reaction` field-by-field with contract-safe fallback. Default is read-only (`read:true, write:false`); malformed/absent config degrades to read-only. `write` is coerced to `read && write`, so write can never be enabled without read. Comment correctly notes the client gate does not replace the server-side write permission boundary.
- **Write gating is defense-in-depth**:
  - `features/messageReaction/controller.ts:150` — `toggle` early-returns when `!canWrite()`, before any optimistic mutation or API call. New controller test proves local state stays empty and the service/emit are not called.
  - `ReactionSlot.tsx` — `if (!canRead) return null`; chip `onClick`/`onAdd` are only wired when `canWrite`, chips rendered `disabled` otherwise, and `openPicker` is guarded.
  - `module.tsx` — context-menu registration and pointer-tracking gated by `canWriteMessageReaction()`, with a runtime config listener that tears down pointer tracking and closes any open picker when write is tightened.
- **Cross-user sync intact** (`Components/Conversation/vm.ts`): `WKSDK.shared().chatManager.addCMDListener` (register at line 950, `removeCMDListener` at 1123) is unchanged — only the gate was renamed to `canReadMessageReaction()`. Since read defaults on, read-only sessions still receive and render others' reaction updates via the existing server CMD push channel. No new event with a phantom emitter was introduced.
- **Pointer-tracking lifecycle** (`ReactionPickerOverlay.tsx`): `enablePointerTracking`/`disablePointerTracking` are symmetric and idempotent; dynamic tighten removes the global capture listeners — no leak in read-only sessions.
- **Prior-verified surfaces (own-reaction-only, message_idstr precision, channel/emoji/seq normalization, optimistic rollback, double-tap dedupe, XSS reactor-name/custom-emoji URL whitelist, per-message/per-channel isolation, no N+1) — unchanged in this delta; prior findings hold.**

## Prior non-blocking note
The "More" full panel not receiving `selectedKeys` is now explicitly documented as a known limitation in `ReactionPickerOverlay.tsx` (would require extending the shared EmojiPanel; scoped out of this API-integration PR). Acknowledged, not a regression.

## Checks
- Build: consuming app compiled cleanly (only benign third-party eval/chunk-size warnings).
- Tests: 30/30 pass (featureFlags 4, MessageReactionConfig 6, syncController 9, controller 7, ReactionSlot 4). localStorage notices are Node-26 environment noise.
- Mergeable: MERGEABLE.

No blocking items. Peers previously approved at `230a5fe8` and none raised a blocker.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #981 (octo-web)

Reviewed at head `a0ac4585c0a75f5d3530fee8c777dc79e40008ae`. Scope: connect message reactions to server — replace demo/mock store with real API toggle + CMD-driven sync, wire appconfig read/write capability gate, integrate into conversation view and context menu.

## Summary

Substantial, well-structured PR that replaces the localStorage-gated demo reaction system with a production implementation. New files introduce a clean separation of concerns: `MessageReactionService` (wire-format normalization + API client), `controller.ts` (optimistic toggle with rollback), `syncController.ts` (cursor-driven CMD sync), `MessageReactionConfig.ts` (appconfig parsing), and `ReactionSlot.tsx` (rendering). The old mock store and demo-only `ReactionSlot` are properly removed. The `featureFlags.ts` transition from `isMessageReactionEnabled()` (localStorage override) to `canReadMessageReaction()` / `canWriteMessageReaction()` (appconfig-driven) is clean and backward-compatible. No P0/P1 issues found; two P2 resource-lifecycle items below.

## Verification

Static analysis only at head `a0ac4585`; build and tests not executed in this environment.

- ✅ **Data-flow: toggle** — user click → `controller.toggle()` → optimistic `applyOwnState()` → `emitUpdated()` → `MessageReactionService.toggle()` via `APIClient.shared.post` → success: `applyOwnState(result.isDeleted, result.seq)` / failure: `rollbackOwnState()`. Pending map serializes concurrent toggles per `(channelType, channelID, messageID, emoji)`. `finally` always removes the pending entry.
- ✅ **Data-flow: sync** — server CMD `syncMessageReaction` → `messageReactionCommandSeq()` parses and validates channel match → `syncController.request(seq)` → `drain()` loop calls `MessageReactionService.sync()` → `apply()` merges reactions by seq ordering into `message.octoReactions`. Cursor always advances; loop exits on empty response, no-progress, or cursor ≥ target.
- ✅ **Data-flow: initial load** — `Convert.ts:305` reads `msgMap["reactions"]` → `normalizeMessageReactions()` → `message.octoReactions`. Handles missing/invalid field gracefully (returns `[]`).
- ✅ **Seq-based conflict resolution** — `mergeMessageReaction` rejects lower-seq incoming; `rollbackOwnState` skips rollback when current seq > previous seq (server confirmed a newer state). Both correct.
- ✅ **Appconfig capability parsing** — `parseMessageReactionCapability` clamps `write: read && write`, so write=true without read=true is impossible. Default `{read: true, write: false}` preserves read-only visibility when server field is absent.
- ✅ **No stale mock references** — grep confirms zero remaining imports of `reactionMockStore`, old `mockStore.ts`, or `MessageReactionSummary/ReactionSlot` in production code.
- ✅ **Channel-type gating consistent** — TextCell, context menu, and ReactionSlot all gate on `isMessageReactionChannelSupported(channelType)` + `messageID` presence. No path renders reactions on unsupported channel types.
- ✅ **i18n parity** — 12 new keys added in both `en-US.json` and `zh-CN.json`. `localizePickerEmojis()` resolves `nameKey` → translated name at render time, keeping static data free of user-visible strings.
- ✅ **Shared client routing** — All API calls go through `APIClient.shared.post()`, inheriting auth, base URL, and interceptors. No raw `fetch` or hardcoded `/api` bypass.
- ✅ **Type augmentation** — `wukongimjssdk-reaction.d.ts` properly declares `octoReactions?: MessageReaction[]` on the SDK `Message` interface via module augmentation.
- ✅ **aggregateReactions** — accepts `undefined | null`, handles empty arrays. Type re-exports in `types.ts` are structurally identical to the old definitions.

## Findings

No P0/P1 issues. Two P2 resource-lifecycle items below.

### P2 — Config change listener leaked on ReactionSlot unmount (`packages/dmworkbase/src/features/messageReaction/ReactionSlot.tsx:63`)

The `useEffect` at line 60–64 calls `WKApp.remoteConfig.addConfigChangeListener(() => { forceTick(...) })` and returns the result as the cleanup. However, `addConfigChangeListener` returns an *unsubscribe function* — so the cleanup actually **adds a second listener** instead of removing the first. Every unmount leaks a listener whose `forceTick` callback silently fires against a stale setter. Fix: capture the unsubscribe return value and return it from the effect, matching the pattern in `ThreadPanel:295`, `EmojiToolbar:254`, `SpaceList:52`, etc.

```ts
useEffect(() => {
  const unsubscribe = WKApp.remoteConfig.addConfigChangeListener(() => {
    forceTick((value: number) => value + 1);
  });
  return unsubscribe;
}, []);
```

### P2 — Pointer-tracking config listener not captured (`packages/dmworkbase/src/module.tsx:967`)

`WKApp.remoteConfig.addConfigChangeListener(syncReactionPointerTracking)` discards the unsubscribe function. Since `BaseModule` lives as long as the application, this does not leak in practice, and the comment at line 963 acknowledges the design. However, the `enablePointerTracking` / `disablePointerTracking` symmetry that `ReactionPickerOverlay.tsx` carefully provides is never fully consumed on teardown. Low urgency, but worth capturing the unsubscribe return value for correctness if `BaseModule` ever gains a dispose hook.

## Data-flow tracing

- **Toggle path**: Click → `ReactionSlot.onSelect` → `controller.toggle()` → optimistic `applyOwnState(message.octoReactions, ...)` → `emitUpdated(messageId)` → `mittBus` → `ReactionSlot.useEffect` → `forceTick` → re-render reads `message.octoReactions`. Server response: `.then()` → `applyOwnState(result.isDeleted, result.seq)` / `.catch()` → `rollbackOwnState(previous)` → `emitUpdated` → `showError`. Data flows correctly from source to consumer at every step.
- **Sync path**: Server CMD → `cmdListener` (vm.ts:938) → `messageReactionCommandSeq()` validates cmd + channel match → `reactionSyncController.request(seq)` → `drain()` → `sync()` API call → `apply(record)` mutates `target.octoReactions` via `mergeMessageReaction` → `notify()` → VM `notifyListener()` → UI re-renders. All data flows verified.
- **Initial load path**: `Convert.ts:305` → `normalizeMessageReactions(msgMap["reactions"])` → `message.octoReactions`. `aggregateReactions(message.octoReactions, currentUid)` in ReactionSlot handles `undefined` gracefully. Data flows verified.

## Blind-spot checklist

- **C1 — Double-path parity**: CLEAR. `enablePointerTracking`/`disablePointerTracking` are symmetric add/remove pairs. Toggle API is single symmetric endpoint (add = unreact via isDeleted toggle). No asymmetric create/delete paths found.
- **C2 — Control-flow ordering**: CLEAR. `createMessageReactionController` is called once (singleton in runtime.ts). `createMessageReactionSyncController` is called once per ConversationVM instance. No nested/duplicate instantiation risk.
- **C3 — Authorization boundary**: CLEAR. Client-side `canWrite`/`canRead` are UI gates only; server validates on the `reactions` endpoint. `parseMessageReactionCapability` correctly clamps write to read. No endpoint/tool exposure beyond the authorized API.
- **C4 — Authorization lifecycle**: CLEAR. When appconfig flips `write: false` at runtime, `syncReactionPointerTracking` calls `disablePointerTracking()` + `reactionPickerOverlay.close()`. Context menu re-evaluates `canWriteMessageReaction()` on each render. In-flight toggles are validated by the server.
- **C5 — Build/runtime path**: CLEAR. No relative URL, copy-all, or heredoc concerns. i18n keys are runtime-resolved via `t()`, not build-time inlined.
- **C6 — Governance/docs**: N/A — no governance/policy changes in this PR.

## Cross-round blocker check

N/A — automated review review, no prior rounds.

- **Optimistic update + rollback**: `applyOwnState` / `rollbackOwnState` correctly use seq-aware merge. Concurrent toggles are serialized by the `pending` map. The `finally` block always cleans up.
- **Sync controller drain loop**: terminates on empty response, no-progress, or cursor ≥ target. The `inFlight` promise deduplicates concurrent `request()` calls.
- **Authorization boundary**: client-side `canWrite` / `canRead` are UI capability gates, not security boundaries. Server validates on `reactions` endpoint. `parseMessageReactionCapability` correctly clamps write to read.
- **CMD listener cleanup**: `ConversationVM` removes `cmdListener` at line 1123, covering the reaction sync branch added at line 938.
- **Input shape validation**: `normalizeToggleResult` throws on invalid responses; `normalizeMessageReaction` drops items with empty uid/emoji; `deletedFlag` safely coerces unknown values to 0.

## Verdict: APPROVED

Two P2 resource-lifecycle items — neither causes user-visible incorrect behavior. The ReactionSlot listener leak is a one-line fix worth applying; the module.tsx item is lowest priority given BaseModule's app-scoped lifetime. Overall this is a thorough, well-tested transition from demo to production wiring.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 resource-lifecycle items (config change listener not cleaned up on ReactionSlot unmount; module.tsx pointer-tracking listener unsubscribe not captured). Both are non-blocking.

yujiawei (approved):
# Code Review — PR #981 (octo-web)

Connects the flag-gated message-reaction UI (#929) to the deployed server contract (`POST /v1/reactions`, `POST /v1/reaction/sync`, inline `reactions[]`, `syncMessageReaction` CMD, and the `message_reaction.{read,write}` appconfig gate). Reviewed at head `a0ac4585` against merge-base `d82645aa`.

**Verdict: APPROVE.** No blocking (P0/P1) issues. Spec is fully met; a few non-blocking P2 observations are listed for follow-up.

## 1. Spec compliance

Verified against the linked issue #983:

- **Toggle write** — `POST /v1/reactions` with `{message_id, channel_id, channel_type, emoji}`; server-authoritative `{seq, is_deleted}` reconciles the optimistic state. ✅
- **Inline parse + incremental sync** — `Convert.ts` maps `reactions[]` → `message.octoReactions`; `syncMessageReaction` CMD drives a seq-cursored `reaction/sync` drain. ✅
- **Optimistic reconcile + error mapping** — `controller.ts` applies an optimistic toggle, reconciles on success, rolls back on failure; deployed error codes are mapped to localized copy (not HTTP status). ✅
- **Text-only, supported channels** — both the context menu and `ReactionSlot` gate on `isMessageReactionChannelSupported` (Person / Group / CommunityTopic) and text content type, and require a server `messageID`. ✅
- **DM peer channel + string message id** — writes use the active conversation channel via `context.channel()`; `message.messageID` is sourced from `message_idstr` (`Convert.ts:267`, `BigNumber` string), so 64-bit precision is preserved. ✅
- **Mock store removed / production path enabled** — `mockStore.ts` and the old `MessageReactionSummary/ReactionSlot.tsx` are deleted; `featureFlags.ts` is rewritten from a DEV localStorage mock to the appconfig capability (`read`, `write`). ✅

No over-build or scope drift: the change stays inside the reaction feature and its wiring; the picker's "full panel selected-key" gap is explicitly left out of scope (unchanged behavior).

## 2. Code quality

Traced the state machine end-to-end and cross-checked the wire contract against the deployed server:

- **Sync drain is safe.** `syncController` guards against infinite loops (`nextSeq <= startSeq` break, empty-batch break), rejects older-seq overwrites, and notifies after a partial batch even if a later batch fails — matching the PR description and its tests.
- **Merge/reconcile is monotonic.** `mergeMessageReaction` refuses to let an older `seq` clobber a newer state; optimistic rollback preserves concurrent users' records and a possibly-newer server state.
- **No dangling references.** All removed symbols (`isMessageReactionEnabled`, `reactionMockStore`, the `message-reaction-mock-updated` event, the old `ReactionSlot`) are fully cleaned up.
- **Gates pass locally.** 57/57 reaction unit tests, `pnpm build` (web + extension), and `pnpm i18n:check` are all green; all 18 `base.reaction.*` keys exist in both locales, and the two demo-file entries in `.i18n/scan-config.json` are correctly removed (the picker data now uses `nameKey` i18n instead of hardcoded labels).

### Non-blocking (P2 / suggestions)

1. **`controller.ts:159` — rapid same-emoji re-toggle is coalesced.** A second toggle of the same emoji while the first request is in flight joins the pending promise rather than queueing the opposite intent, so a quick add→remove can settle as "added." It self-heals via the authoritative `seq`/CMD sync, so it is a minor UX edge, not data loss. Consider queueing the latest intent.
2. **`controller.ts:184` / `syncController.ts:144` — reconcile targets the message instance by reference.** If history refresh/pagination replaces a message object while a toggle or a sync record is outstanding, the update lands on the obsolete instance (or a record advances the cursor without being applied to a not-yet-loaded message). In practice the server inlines a fresh `reactions[]` snapshot per message on channel/conversation sync, so a re-loaded message re-hydrates authoritative state — which is why this is P2 rather than a stale-forever bug. Resolving the current message by id at apply time would make it robust independent of that backstop.
3. **`Components/Conversation/vm.ts` — sync controller has no explicit teardown.** A `reaction/sync` promise resolving after `didUnMount` would call `notifyListener()` on a torn-down VM. `didUnMount` already removes the CMD listener (so no new syncs start), making this benign, but an in-flight guard would silence any late no-op.

## Coverage / blind spots

- **Cross-repo contract verified.** I read the deployed server structs in `octo-server` (`modules/message/api.go`): both `reactionToggleResp` and `reactionResp` marshal `message_id` as a **string** and `seq` as `int64` — matching the client's `typeof === "string"` / finite-number guards exactly, and the `syncMessageReaction` CMD `param` carries `{message_id, channel_id, channel_type, emoji, seq}` as the client expects.
- **Not exercised at runtime here:** live multi-client realtime propagation, DM cross-Space visibility filtering (enforced server-side), and reconnect-after-sleep CMD-loss recovery — recommend a manual/QA pass on those, though the server-side sync path and inline snapshots cover the static reasoning.

PR #968, opened 2026-07-22
refactor(contacts): prepare group creation runtime boundary

menghao-webtest:
## Summary
- add a current channel runtime facade for channel info and subscriber access
- move group creation candidate loading and submit actions into a contacts bridge
- update GroupNew, ChannelSetting, and GroupManagement to use the narrowed runtime boundary

## Verification
- pnpm --dir packages/dmworkcontacts test -- src/Organizational/GroupNew/__tests__/memberUids.test.ts src/bridge/groupCreate/groupCreateRuntime.test.ts
- pnpm --dir packages/dmworkbase test -- src/im-runtime/currentChannelRuntime.test.ts src/Components/ChannelSetting/__tests__/vm.persona.test.ts src/Components/GroupManagement/__tests__/allowNoMention.test.ts src/Components/GroupManagement/__tests__/botAdmins.test.ts
- git diff --check

PR #982, opened 2026-07-22
refactor(contacts): split group creation UI

Candice00711:
Refactor the existing group creation and add-member flow into a bridge hook and props-driven UI components while preserving the current product behavior and user entry points.

## Summary

The legacy `Organizational/GroupNew` class now acts as a thin entry container. Group creation state and submission orchestration live in `useGroupCreate`, while the existing dialog and member-selection layouts are split into focused UI components. The selected-member remove icon is refined to match the approved smaller default and hover states.

## Related Issue

N/A — behavior-preserving structural refactor.

## Changes

- Extract group creation and add-member state, validation, filtering, candidate loading, and submission orchestration into `useGroupCreate`.
- Extract props-driven `GroupCreateDialog` and `GroupMemberPicker` UI components with Storybook coverage.
- Preserve the dormant organization endpoints, nested tree conversion, employee counting, and member collection as compatibility bridge helpers without restoring the hidden organization entry or triggering organization requests.
- Keep the current group avatar, group name, member selection, search, validation order, close/reset behavior, and create/add-member submission semantics.
- Reduce the selected-member remove glyph to 12px inside its existing 24px hit area, with token-based default and hover states.

- Affected module(s): `packages/dmworkcontacts`
- Shared layer touched: `bridge/` and `ui/`
- Runtime boundaries unchanged: `groupCreateRuntime.ts`, `memberUids.ts`, `types.ts`, `WKSDK`, `channelDataSource`, `SpaceService`, and `showConversation` have no diff.
- Duplicate entry point checked: yes; the organization entry remains hidden.
- Impact scope: existing create-group and add-member callers of `OrganizationalGroupNew` only.

## Testing

- [x] `pnpm --dir packages/dmworkcontacts test` — 25 tests passed.
- [x] `pnpm i18n:check` — passed with locale keys healthy.
- [x] `pnpm --dir apps/web build` — passed after rebasing onto the latest `main`.
- [x] `pnpm lint:css` — 0 errors; existing repository warnings remain.
- [x] `git diff --check` — passed.
- [x] Runtime boundary zero-diff check — passed.
- [x] Manually verified the create-group flow and selected-member remove icon in the local app.
- [ ] `pnpm --dir apps/web build-storybook` — all 10,089 modules transform successfully, then the repository-level process exits with code 132, matching the existing full Storybook build behavior.

## Checklist

- [x] I have read the repository development and contribution rules.
- [x] PR description is in English.
- [x] Added focused tests and Stories for the changed boundaries.
- [x] Documentation is not required because no public API or user workflow changed.
- [x] Ran `pnpm i18n:check` and confirmed no user-visible copy changed.
- [x] Confirmed the change follows module ownership and does not add duplicate user-visible entry points.
- [x] Described the bridge/UI impact scope and unchanged runtime boundaries.
- [x] Followed Conventional Commit conventions.

### DIFF
--- a/packages/dmworkcontacts/src/Organizational/GroupNew/index.tsx
+++ b/packages/dmworkcontacts/src/Organizational/GroupNew/index.tsx
@@ -1,32 +1,18 @@
-import React from "react";
-import { Component, ReactNode } from "react";
+import React, { Component, ReactNode } from "react";
+import { Toast } from "@douyinfe/semi-ui";
 
-import {
-  Button,
-  Space,
-  Tree,
-  Input,
-  CheckboxGroup,
-  Checkbox,
-  Toast,
-} from "@douyinfe/semi-ui";
-import { BasicTreeNodeData } from "@douyinfe/semi-foundation/lib/cjs/tree/foundation";
-import { WKApp, ThemeMode, WKViewQueueHeader, WKModal, I18nContext, t, GroupAvatarPreview, GroupAvatarEditModal } from "@octo/base";
-import WKAvatar from "@octo/base/src/Components/WKAvatar";
-import AiBadge from "@octo/base/src/Components/AiBadge";
-import "./index.css";
+import { I18nContext, t, WKApp } from "@octo/base";
 import { GROUP_NAME_MAX_LENGTH } from "@octo/base/src/Service/nameLimits";
-import {
-  loadGroupCreateCandidates,
-  submitGroupCreateAction,
-} from "../../bridge/groupCreate/groupCreateRuntime";
+
+import { useGroupCreate } from "../../bridge/groupCreate/useGroupCreate";
+import GroupCreateDialog from "../../ui/GroupCreateDialog";
 
 export enum OrganizationalGroupNewAction {
-  createGroup, // 创建群聊
-  AddMember // 添加成员
+  createGroup,
+  AddMember,
 }
 
-interface IPorpsOrganizationalGroupNew {
+interface OrganizationalGroupNewProps {
   channel: {
     channelID: string;
     channelType: number;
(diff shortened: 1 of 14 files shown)


lml2468 (approved):

审查锚定 head `5c2dc4f0`。14 文件 +1210/−803,把群创建 UI 从 `GroupNew` 类拆成 `useGroupCreate` hook + props 驱动的 `GroupCreateDialog`/`GroupMemberPicker`(接在已审的 #968 bridge 之上)。行为保持的纯结构重构,本地跑测试 + build + 逐条核过。

### ✅ 两道闸门均通过

**规格符合**:与描述一致 —— GroupNew 变薄容器,状态/校验/候选加载/提交编排进 `useGroupCreate`,UI 拆成 props 组件;保留群头像/群名/选人/搜索/校验顺序/关闭复位/建群与加成员语义;删除图标缩到 12px。行为保持、无 over-build。

**代码质量 —— 行为保持(逐条核实)**:
- **runtime 边界零 diff(claim 核实)**:`groupCreateRuntime.ts`/`memberUids.ts`/`types.ts` 逐个 `git diff` **无差异** —— 候选加载/排除/翻页/提交(#968 已测)原样复用;
- **useGroupCreate 委派 #968 已测函数**:`loadGroupCreateCandidates`(:82)+ `submitGroupCreateAction`(:135)—— 逻辑不重写、走已验证 bridge;
- **两个入口保留**:`OrganizationalGroupNewAction.{createGroup, addMember}` enum+prop 保留,容器 action→mode 映射(:44)传入 hook;外部调用方不变;
- **校验顺序保留**:name-required → members-required → submit(:125-131),与原类同序;`createGroup` 带 avatar/name/category、`addMember` 不带,`onSuccess` 仅 createGroup;
- **org 兼容层真休眠**:`organizationCompatibility` 仅纯树/计数 helper(`buildGroupCreateOrganizationTree` 等),**无 apiClient/fetch/get/post**,且**无任何生产路径 import** → 不触发 org 请求(claim 核实)。

### ✅ 验证

- `@octo/contacts` → **25/25 全绿**(8 文件,含 useGroupCreate + organizationCompatibility + 既有 groupCreateRuntime);
- `pnpm --filter @octo/web build` → **`✓ built in 3.51s`**;i18n locale healthy;

### 结论

纯 UI 抽取重构:runtime 边界零 diff、逻辑委派 #968 已测 bridge、两入口 + 校验顺序保留、org 兼容层无请求且未接线(真休眠)、Story 覆盖齐;测试 25/25、build 通过、无冲突、无行为回归。无阻塞。**APPROVE**。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #982 (octo-web)
**Head SHA:** `5c2dc4f0eb7f8b651a27bec854f8cea93ecdd9b0`
**Scope:** 14 files, +1210/-803 lines — `packages/dmworkcontacts/src/`

## Summary

This PR refactors the monolithic `OrganizationalGroupNew` class component (~800 LOC) into well-separated modules: a `useGroupCreate` hook for state/business logic, `GroupCreateDialog` and `GroupMemberPicker` presentational components, extracted organization-tree helpers, plus Storybook stories and unit tests. The business logic — candidate loading, validation ordering (name before members), submit flow, avatar editing, and close/cancel lifecycle — is faithfully preserved. This is a strong structural improvement with good test coverage of the new hook and helper modules.

No P0/P1 issues found. Three P2 items below.

## Verification

- ✅ **Business logic parity** — traced `submit` in `useGroupCreate.ts:121-158` against old `onOK`: name-required check fires before members-required check, `onSuccess` only called for `createGroup` (not `addMember`), error toast via `errorMessage()` matches old `error.msg` pattern. Confirmed by `useGroupCreate.test.tsx` assertions at lines 91-110.
- ✅ **Open/close lifecycle** — `OrganizationalGroupNew.close()` at `:171-174` correctly calls `setState({ isOpen: false })` then `this.props.remove?.()`, matching old `onCancel` semantics. The `useEffect` at `useGroupCreate.ts:83-110` resets form state on both open and close transitions, with `loadSequence` ref guarding stale candidate loads.
- ✅ **Data flow: candidates → UI** — `loadGroupCreateCandidates` (bridge runtime) returns `GroupCreateCandidateContact[]`, stored in hook state, filtered by `filterGroupCreateCandidates` for keyword search, passed to `GroupMemberPicker` as `candidates` prop. Traced end-to-end; data flows correctly.
- ✅ **Data flow: selected members → submit** — `toggleMember` at `:112-123` looks up from `candidates` state (not just `visibleCandidates`), so toggling works even for members hidden by current search filter. `selectedUidSet` derived from `selected` via `useMemo`. Submit reads `selected.map(m => m.uid)` — correct.
- ✅ **CheckboxGroup sync** — `GroupMemberPicker/index.tsx:47-54` uses toggle-based diffing against `selectedUids` (authoritative Set from hook), which is more robust than the old code's full-replace via `onFriendChange`.
- ✅ **i18n keys** — all keys used in `GroupCreateFeature` (`groupCreate.title`, `.avatarLabel`, `.editAvatar`, `.nameLabel`, `.namePlaceholder`, `.membersLabel`, `.selectedCount`, `.nameRequired`, `groupNew.selectContactsWarning`, `common.*`) confirmed present in both `en-US.json` and `zh-CN.json`.
- ✅ **createGroup vs addMember dual-path parity** — both paths validated, submitted via `submitGroupCreateAction` with correct args, proper callback invocation. Old `onSuccess` asymmetry (createGroup only) preserved.

Static analysis only at head `5c2dc4f0eb7f8b651a27bec854f8cea93ecdd9b0`; build and tests not executed in this environment.

## Findings

### P2 — `isSubmitting` computed but never wired to UI (`useGroupCreate.ts:63`, `GroupMemberPicker/index.tsx:145`)

The hook declares `isSubmitting` state and correctly sets it in the submit callback, and the return object includes it at `:174`. However, `GroupCreateFeature` in `Organizational/GroupNew/index.tsx:27-160` never reads `model.isSubmitting`, and the `GroupCreateDialog` props type has no `isSubmitting` field. The confirm button in `GroupMemberPicker/index.tsx:145` has `loading={false}` hardcoded. Users can still click Confirm multiple times during an in-flight submission.

The old code had the same gap (no loading prop), so this is not a regression — but the infrastructure is now in place to wire it through. Fix direction: pass `isSubmitting` through `GroupMemberPickerProps`, replace `loading={false}` with `loading={isSubmitting}`.

### P2 — Unhandled rejection in candidate load effect (`useGroupCreate.ts:97-102`)

```ts
void loadGroupCreateCandidates({ channel: options.channel }).then(
  (next) => {
    if (loadSequence.current === sequence) {
      setCandidates(next);
      setVisibleCandidates(next);
    }
  }
);
```

The `.then()` chain has no `.catch()`. If the API call rejects (network error, auth expiry), the rejection propagates as an unhandled promise rejection. The candidate list silently remains empty with no user-visible error feedback.

Pre-existing behavior (old `getFriendData` had the same gap at `onShowModal`), so not a regression. Fix direction: append `.catch((err) => options.notice.onError(errorMessage(err)))` to surface load failures via the same toast path used for submit errors.

### P2 — Organization tree modules are unreachable dead code (`organizationCompatibility.ts`, `organizationRuntimeCompatibility.ts`)

These two new files (165 lines combined) export `buildGroupCreateOrganizationTree`, `collectGroupCreateOrganizationMembers`, and `loadGroupCreateOrganization`, but no production code imports them — only the test file imports from `organizationCompatibility.ts`. The org-tree UI was already dead code in the old component (`getJoinOrganization` was commented out at the call site), so this is not a regression. However, the new modules add a public API surface that could drift from the actual org-tree revival work. Consider deferring them to the PR that re-enables the feature, or marking them as `@internal` / `__future` to signal intent.

- **Avatar editor lifecycle** — `GroupCreateDialog/index.tsx:100-107` correctly renders `GroupAvatarEditModal` outside both modal branches, so it's accessible in both `createGroup` and `addMember` modes. `onSave` callback properly destructures `result.avatarText` and `result.colorIndex`.
- **CSS imports** — both `GroupCreateDialog/index.css` and `GroupMemberPicker/index.css` use `@import "../../Organizational/GroupNew/index.css"`, preserving the existing stylesheet. New `.group-member-remove-icon` styles follow project CSS variable conventions.
- **Storybook stories** — stories cover createGroup, addMember, empty, and long-text variants. They use hand-crafted props (not the hook), which is appropriate for presentational component isolation.
- **Test coverage of validation ordering** — `useGroupCreate.test.tsx:91-100` explicitly asserts name-required fires before members-required, locking the contract.
- **Test coverage of submit contracts** — separate tests for `createGroup` (with createOptions) and `addMember` (without createOptions), matching the dual-path production behavior.
- **`errorMessage` helper** — handles `{ msg: string }` server errors, `Error` instances, and unknown types. Matches old `error.msg` access with better type safety.

## Verdict: APPROVED

This is a clean, well-tested refactoring that preserves all behavioral contracts. The three P2 items are non-blocking: `isSubmitting` wiring is a UX opportunity rather than a bug, the unhandled rejection is pre-existing, and the dead-code modules are a minor maintainability note. None of these make a working path unavailable, produce incorrect data, or worsen behavior versus the base branch.

Jerry-Xin (approved):
## ✅ APPROVE — behavior-preserving UI split, byte-verified

This is a clean structural refactor: the monolithic `Organizational/GroupNew` class becomes a thin entry container, state/orchestration moves into a `useGroupCreate` hook, and the dialog/member layouts split into props-driven `GroupCreateDialog` + `GroupMemberPicker`. The load-bearing logic from the prior runtime-boundary work is untouched, so the risk surface is narrowed to wiring — which verifies out.

### Behavior-preservation (verified)
- **Submit runtime is byte-unchanged.** `git diff` confirms `bridge/groupCreate/groupCreateRuntime.ts`, `memberUids.ts`, and `types.ts` have **zero diff** in this PR. Candidate filtering (system-uid exclusion for `botfather`/`fileHelper`, space-only current-user exclusion), subscriber derivation, pagination, and all three submit branches (createGroup → `createChannel` + `showConversation({fromSidebarList})`; private-chat → `buildPrivateChatGroupMemberUids` + `createChannel`; group → `addSubscribers`) remain exactly as previously approved.
- **Validation order preserved.** `useGroupCreate.submit` runs name-required (createGroup only) → members-required, matching the old `onOK`. `createOptions` is `undefined` for `addMember` (byte-equivalent to the old add-member path that sent no create metadata).
- **`keepSidebarTab` threading is safe.** The hook passes it for both actions, but the runtime reads it only inside the `createGroup` branch, so the add-member path is unaffected — no behavior change.
- **Keyword filter equivalent.** `filterGroupCreateCandidates` uses `name.toLowerCase().includes(...)`, matching the old `indexOf(...) !== -1`, and preserves candidate order.

### Prop threading / call sites
- Container fully re-wires props → hook → `GroupCreateDialog` → `GroupMemberPicker`. `CheckboxGroup.onChange` diffs old vs new selection and calls `toggleMember` per changed uid — equivalent to the old immutable `onFriendChange`. Confirm/cancel wired in both modes (createGroup via `WKModal.footerConfig.onOk`; addMember via the picker footer). No duplicate confirm control.
- Public export `OrganizationalGroupNew` preserved in `packages/dmworkcontacts/src/index.tsx`; both `module.tsx` call sites (AddMember with `render`; createGroup with `autoShow`/`remove`/`defaultCategoryId`/`onSuccess`/`keepSidebarTab`) are still fully supported. `autoShow`→`componentDidMount().open()`, `remove`→`close()` preserved.
- **No half-split**: the old monolithic implementation is gone (the file is now a thin container), no dead old code left behind.

### State isolation
- `useGroupCreate` resets selected/name/avatar/editor on `isOpen === false` and reloads candidates with a `loadSequence` guard on open — cancel/reopen resets cleanly, no stale leak, no out-of-order candidate write.

### Dormant org compatibility
- `organizationCompatibility.ts` / `organizationRuntimeCompatibility.ts` are referenced only by their tests — no live caller (grep-verified), so no organization endpoint fires and the hidden entry stays hidden, as described.

### Types / dead code
- Critical submit path is fully typed via `types.ts` in both new UI folders; no `any` escape on the submit path. No unused imports or orphaned exports spotted.

### Build / tests / CI
- `pnpm --dir apps/web build` → **passed** (`✓ built`, exit 0); only pre-existing third-party warnings (lottie/pdfjs eval, chunk size). No syntax/orphan-brace errors.
- `pnpm --dir packages/dmworkcontacts test` → **25/25 passed**, including new `useGroupCreate` coverage for filter order, name-before-member validation, create submit, and add-member-without-create-metadata.
- i18n keys (`contacts.groupCreate.*`) resolve in both `en-US.json` and `zh-CN.json`; locale files untouched → parity intact.
- CI green; `mergeable: MERGEABLE` (`BLOCKED` only pending this review).

### 💬 Non-blocking
- The add-member confirm button hard-codes `loading={false}` and `isSubmitting` isn't threaded to disable the button during submission. The old code had no disable either, so this is not a regression — but wiring `isSubmitting` in a follow-up would prevent double-submit.

Nice, disciplined split. LGTM.

yujiawei (approved):
# Code Review — PR #982 (octo-web)

**Verdict: APPROVE.** This is a clean, behavior-preserving refactor that splits the monolithic `OrganizationalGroupNew` class into a `useGroupCreate` bridge hook plus props-driven `GroupCreateDialog` / `GroupMemberPicker` UI components. Verified locally against head SHA `5c2dc4f0eb7f8b651a27bec854f8cea93ecdd9b0`.

## 1. Spec / scope compliance

This is declared a behavior-preserving structural refactor (no linked issue). Assessed against that contract:

- **No under-build**: create-group and add-member flows, validation order (name-before-members), close/reset, avatar edit, search filtering, and submission semantics are all preserved.
- **No over-build**: the dormant organization tree remains hidden. `getJoinOrganization()` was already commented out on `main`, so the org-tree code path was dead before this PR. The refactor relocates that dead logic into `organizationCompatibility.ts` / `organizationRuntimeCompatibility.ts` rather than re-activating it — no new user-visible entry point is introduced. Confirmed no caller wires these helpers.
- **Runtime boundary zero-diff**: `groupCreateRuntime.ts`, `memberUids.ts`, and `types.ts` have no diff — verified with `git diff`. The channel-creation / add-subscriber logic is untouched.

## 2. Verification performed

- `pnpm --dir packages/dmworkcontacts test` → **25/25 pass**
- `pnpm --dir apps/web build` → **exit 0** (`✓ built in 13.07s`)
- All i18n keys referenced by the new components resolve in both `en-US` and `zh-CN`.
- The `--wk-sp-*`, `--wk-r-sm`, `--wk-icon-muted`, `--wk-icon-hover` tokens used by the refined remove-glyph CSS are defined in `packages/dmworkbase/src/theme/semantic.css`.

## 3. Security note (for the human reviewer)

This PR was auto-flagged `security_sensitive`, but on inspection that appears to be a keyword false-positive. There are no auth, token, credential, or middleware changes. The only boundary that performs privileged mutations (creating channels, adding subscribers) is `groupCreateRuntime.ts`, which is **zero-diff**. No security-relevant surface changed.

## 4. Non-blocking findings (P2 / nits)

- **P2 — selected-member list ordering changed.** The legacy `onFriendChange` rebuilt the selected list in *candidate-list order* on every checkbox change. The new `toggleMember` appends to `selected` in *click order* (`useGroupCreate.ts:101-112`). The submitted `selectedUids` order changes accordingly. This is cosmetic (member order is not semantically meaningful for group creation), but it is a minor deviation from a strict "preserve current behavior" claim — worth a conscious sign-off.
- **P2 — `isSubmitting` is computed but never consumed.** `useGroupCreate` exposes `isSubmitting` (`useGroupCreate.ts:174`), but `GroupMemberPicker` hardcodes `loading={false}` (`GroupMemberPicker/index.tsx:147`) and the create dialog's confirm button isn't disabled during submit. So there's no double-submit guard / spinner. The legacy code had no guard either, so this is not a regression — but since the hook now surfaces the flag, wiring it into the confirm button would be a cheap correctness win.
- **Nit — dead compatibility helpers.** `organizationCompatibility.ts` and `organizationRuntimeCompatibility.ts` are unreferenced outside their own tests. They faithfully preserve already-dead logic, which is a reasonable intent, but please confirm they are intentionally parked for future reuse rather than left over accidentally; otherwise consider deleting to avoid untested-in-integration drift.
- **Nit — trigger rendering is now mutually exclusive.** The old `render()` output `showAdd` and `render` as two independent blocks; the refactor makes `render` take precedence over `showAdd` (`GroupNew/index.tsx` `renderTrigger`). No current caller passes both, so no functional impact.

cnwenf (commented):
# Code Review — PR #982 (Mininglamp-OSS/octo-web)

> Consolidated from three independent reviews (octo-review-qwen3.7, octo-review-glm5.2, octo-review-qwen3.8). Findings are deduplicated; `[source: ...]` marks which reviewer(s) raised each finding. The three full raw reviews remain as comments on the tracking issue for audit.

## Summary

This PR splits the monolithic `Organizational/GroupNew` class into a thin entry container, a `useGroupCreate` bridge hook, and two props-driven UI components (`GroupCreateDialog`, `GroupMemberPicker`), while keeping the submission runtime (`groupCreateRuntime.ts`, `memberUids.ts`, `types.ts`) byte-for-byte unchanged. All three independent reviewers confirm a faithful, behavior-preserving refactor: the active create-group / add-member flow and the zero-diff runtime boundaries are verified at head `5c2dc4f0`, the dormant organization entry stays hidden (zero production importers), and security is cleared with 0 P1/P2 findings across all three. No reviewer found a functional regression or a merge blocker. Consolidated findings are 6 P2 (test-coverage gaps on newly-introduced hook behaviors, plus a stale-selection-on-channel-change and a silently-swallowed candidate-load rejection) and 8 P3/nits (dormant org-helper divergences and minor contract-preservation notes). Notably, the behaviors flagged by glm5.2 as *untested* were independently confirmed *correct* via ad-hoc probes by qwen3.8 (and the race guard by qwen3.7) — so those P2s are about pinning tests, not runtime defects.

## Findings

### P2 — Candidate load rejection silently discarded (`useGroupCreate.ts:82`) [source: octo-review-qwen3.7]

`loadGroupCreateCandidates` is called with `void` prefix and no `.catch()` handler. If the async candidate load fails (network error, API timeout from `loadExcludedSubscriberUids` or `collectSpaceMembers`), the promise rejection is silently swallowed. The dialog opens with an empty candidate list and no error state, giving the user no feedback about what went wrong. The hook state machine has no error state to represent this failure mode.

Old code also lacked error handling (`getFriendData()` had no `.catch()`), so the contract is preserved — but the new `void` prefix makes the discard deliberate and more visible. Fix: add `.catch((e) => options.notice.onError(errorMessage(e)))` to surface the failure, or introduce an `error` state in the hook model.

### P2 — Selected members persist across channel changes while dialog is open (`useGroupCreate.ts:69`) [source: octo-review-qwen3.7]

When the `channel` prop changes while the dialog is open, the `useEffect` resets `groupName`, `avatarText`, `avatarColorIndex`, and `avatarEditorOpen`, but does NOT reset `selected`. Candidates reload for the new channel (line 84), but `selected` retains UIDs from the old channel. `selectedUidSet` derives from stale `selected` (line 93), creating phantom selections that aren't in the new candidate pool. The user sees checked boxes for members who don't exist in the current context.

Fix: add `setSelected([])` to the channel-change branch of the useEffect, or key the dialog on `channel` to force remount.

### P2 — Dormant `loadGroupCreateOrganization` makes unvalidated API calls (`organizationRuntimeCompatibility.ts:19`) [source: octo-review-qwen3.7]

The exported function `loadGroupCreateOrganization` makes two `WKApp.apiClient.get` calls (`/organization/joined` and `/organizations/:id/department`) but has zero callers anywhere in the codebase. The `org_id` URL parameter is interpolated directly from the `/organization/joined` response without validation. There is no `try/catch` around either API call — a rejection would propagate unhandled.

Old code had `getJoinOrganization()` commented out in `onShowModal` with `// this.getJoinOrganization();`, signaling intentional dormancy. The new export lacks this dormancy marker. While currently inert, an accidental import/call would produce unhandled rejections. Fix: add `try/catch`, validate `org_id` before URL interpolation, or add a `// @deprecated — dormant bridge` JSDoc marker.

### P2 — `submit()` error path is untested (`useGroupCreate.ts:152`) [source: octo-review-glm5.2]

The catch block calls `options.notice.onError(errorMessage(error))` and keeps the dialog open (correct, matches the old `onOK` parity), but no test rejects `submitAction` to exercise this branch. Because `useGroupCreate.test.tsx` is a new file in this diff, the gap is diff-introduced. Add a test where `submitGroupCreateAction` rejects and assert `onError` fires and the modal stays open.

> Consolidation note: qwen3.8 independently probed this exact path (`onError` toast + `isSubmitting` recovery + dialog stays open) and confirmed it behaves correctly — the gap is the missing committed test, not a runtime defect.

### P2 — `loadSequence` stale-candidate guard is untested (`useGroupCreate.ts:64`) [source: octo-review-glm5.2]

The `loadSequence` ref (`:64`, incremented `:76`, checked `:84`) is genuinely new — the old `getFriendData()` had no race guard. Mutation: removing the guard, all 25 tests still pass. Add a test that fires a second `loadGroupCreateCandidates` before the first resolves and asserts only the latest result is committed.

> Consolidation note: both qwen3.7 and qwen3.8 confirmed the latest-wins guard is correct (qwen3.8 via a passing race-guard probe); the gap is the unpinned contract, not the behavior.

### P2 — `toggleMember` deselect / unknown-uid no-op untested (`useGroupCreate.ts:101`) [source: octo-review-glm5.2]

`toggleMember` introduces a deselect path and an unknown-uid no-op guard that the old tree-widget selection could not reach. Mutation: removing both, 25/25 still pass. Add tests for toggling an already-selected member (deselect) and a uid absent from candidates (no-op).

### P3 — `render`/`showAdd` are now mutually exclusive (`GroupNew/index.tsx:148`) [source: octo-review-glm5.2 + octo-review-qwen3.7] (severity split: glm5.2=Nit, qwen3.7=P3)

New `renderTrigger()` makes `render` and `showAdd` mutually exclusive: if `render` is provided, `showAdd` is silently ignored (early return at `:149`). In the old code, both triggers rendered simultaneously when both props were truthy. The TypeScript interface `OrganizationalGroupNewProps` still declares both as independent optional props with no comment about mutual exclusivity. No current caller passes both (`module.tsx` passes `render` for add-member, neither for create-group), so this has zero runtime impact — but the public class API / type contract is silently narrowed without documentation. Worth a one-line JSDoc noting the precedence.

### P3 — `isSubmitting` state tracked but never wired to UI (`useGroupCreate.ts:153`) [source: octo-review-qwen3.7]

The hook tracks and returns `isSubmitting` state (set true during submit, false in finally), but the `GroupCreateFeature` bridge component never passes it to `GroupCreateDialog` or `GroupMemberPicker`. The confirm buttons have `loading={false}` hardcoded, allowing double-submit during network latency. Old code also lacked submit-in-progress UI feedback, so the contract is preserved — but the new hook explicitly exposes this capability without any consumer.

### P3 — Search keyword persists across dialog open/close cycles (`useGroupCreate.ts:76`) [source: octo-review-qwen3.7]

When `isOpen` transitions to true, the useEffect resets `groupName`, `avatarText`, `avatarColorIndex`, and `avatarEditorOpen`, but does NOT reset `keyword`. The search input retains the previous session's text while the candidate list reloads as the full unfiltered list, creating a visual inconsistency. Old code (`onShowModal`) also omitted `searchVaule` reset, so the contract is preserved.

### P3 — Legacy CSS double-imported in render tree (`GroupCreateDialog/index.css:1`, `GroupMemberPicker/index.css:1`) [source: octo-review-qwen3.7]

Both `GroupCreateDialog/index.css` and `GroupMemberPicker/index.css` use `@import '../../Organizational/GroupNew/index.css'` to load the same legacy stylesheet. Since `GroupCreateDialog` renders `GroupMemberPicker` as a child, the legacy CSS is imported twice in the same render tree. Bundler deduplication prevents functional issues, and no specificity conflict exists with the new `.group-member-remove-icon` rules (the legacy CSS has no overlapping selectors). Consider importing the legacy stylesheet once at the dialog level only.

### Nit — `setSubmitting(false)` writes after teardown (`useGroupCreate.ts:154`) [source: octo-review-glm5.2]

On the create-group success path `options.onClose()` (`:151`) unmounts the host (the create-group entry calls `remove()` → `unmountComponentAtNode`), then the `finally` at `:155` calls `setSubmitting(false)` on the unmounted hook. In React 18 this is a silent no-op (the unmounted-setState warning was dropped), so there is no crash or leak — but the cleanup ordering is inverted vs. the old `onOK`, which called `this.onCancel()` last with no `finally`. An `isMountedRef` guard or moving `setSubmitting(false)` before `onClose()` would make the intent explicit. Not a change I'd block on.

### P3 — Dormant org tree transform diverges from legacy (`organizationCompatibility.ts:67`) [source: octo-review-qwen3.8]

`buildGroupCreateOrganizationTree` always recurses into `department.children` and appends employees, whereas the legacy `handleTree` built children only when BOTH `item.children` AND `item.employees` existed — a department with children but no `employees` key had its sub-departments dropped in the old code. A probe importing the real production entry at head `5c2dc4f0` confirms the new tree contains a sub-department node the legacy transform never produced, and the new nodes omit the legacy `icon` field (folder svg / WKAvatar). Today this is unreachable — `organizationRuntimeCompatibility.ts` (the only non-test importer) has zero importers — so the risk is a silent contract break only if the hidden organization entry is ever re-wired. Not a change I'd block on; either replicate the old child-gating and document that icons are intentionally dropped (JSX cannot live in the pure `.ts` module), or note the divergence so re-enablement re-adds icon rendering at the UI layer, and pin the intended gating with a regression test.

### P3 — No uid dedup in dormant member collection, plus two unpinned contracts (`organizationCompatibility.ts:100`) [source: octo-review-qwen3.8]

`collectGroupCreateOrganizationMembers` flatMaps the tree into candidate contacts with no dedup by uid, so one employee shared across two departments is emitted twice. A probe importing the real production entry at head `5c2dc4f0` yields `['shared','shared']` (length 2, Set size 1); the legacy path it preserves deduped via reduce-by-uid, and the file is added by this diff, so the divergence is introduced here. The contract is also unpinned: the existing suite covers one nested tree with distinct uids, and a mutation adding result-level uid dedup leaves it green while a focused probe is killed. The same mutation protocol shows `buildGroupCreateOrganizationRootEmployees` (key `uid_${uid}` / label) has no test at all. Reachable consequence is nil today — both helpers are dormant (`rg`-verified zero production importers) — but reactivation would let a department select add the same member twice, inflate `selectedCount`, and send duplicate `selectedUids`. Fix direction: dedup by uid in collect (e.g. accumulate into a Map keyed by uid) and add a shared-uid-across-departments regression test plus a root-builder test asserting `key=uid_${uid}` and label.

### P3 (Warning, verification blocked) — Remove-glyph pixel geometry not machine-checkable [source: octo-review-qwen3.8]

The remove-glyph's exact pixel geometry (12px glyph inside a 24px hit area with token-based default/hover) has no machine-checkable assertion — jsdom has no layout engine, so only element presence and click wiring could be verified. Not a defect; a Storybook/visual-regression or CSS-level assertion would close it.

- **Runtime boundaries genuinely zero-diff** — `groupCreateRuntime.ts`, `memberUids.ts`, `types.ts`, and the `WKSDK` / `channelDataSource` / `SpaceService` / `showConversation` touchpoints all confirmed zero-diff at head `5c2dc4f0`; candidate loading, submission payload, and `showConversation` wiring are untouched. [all three]
- **Submission semantics preserved** — create-group forwards trimmed name + avatar text/color + `categoryId` + `keepSidebarTab` + `onSuccess`; add-member sends `createOptions: undefined` and no `onSuccess`. Name-before-member validation order and close/reset lifecycle match the predecessor. The add-member path now also passes `keepSidebarTab`, but `submitGroupCreateAction` only reads it on the create-group branch, so the delta is benign. [all three]
- **Error routing correct** — `onError(errorMessage(error))` fires on APIClient rejects, the dialog stays open, `isSubmitting` recovers, and a later submit can succeed (`errorMessage` extracts `error.msg` from the normalized envelope, consistent with old `Toast.error`). [glm5.2, qwen3.7, qwen3.8]
- **Race guard correct** — the `loadSequence` latest-wins guard discards a stale candidate load on channel switch mid-flight and prevents setState-after-unmount; keyword-filter selection retention holds. (Behavior confirmed correct; only the committed-test coverage is missing — see P2s above.) [qwen3.7, qwen3.8]
- **Dormant org entry stays hidden** — the active path (`GroupNew` → `useGroupCreate` → `groupCreateRuntime`) never references the org helpers; `organizationRuntimeCompatibility.ts` has zero production importers; the legacy org call was already commented out and no new entry restores it. [all three]
- **Security cleared** — no XSS/injection surface (all JSX text interpolation, React auto-escaping; no `dangerouslySetInnerHTML`/`innerHTML`/`eval`), no new reachable external call, no credential exposure, no auth/authz regression; member names render as auto-escaped text nodes and search feeds a pure `.toLowerCase().includes()` filter. [all three]
- **Selected-member remove icon** — 12px SVG inside a 24px (`--wk-sp-6`) hit area with token-based default/hover; the two new CSS files `@import` the legacy `GroupNew/index.css` but there is no overlapping `.group-member-remove-icon` rule there, so no specificity conflict. [glm5.2, qwen3.7, qwen3.8]
- **Class API / i18n stability** — `OrganizationalGroupNew` still exports the same props interface and `OrganizationalGroupNewAction` enum is unchanged; all locale keys verified healthy with no user-visible copy changes. [qwen3.7]

## Verification Results

| Code Review | APPROVED_WITH_RISK | 6 P2 + 8 P3/nits (deduped across 3 reviewers), no P0/P1, no merge blockers |
| E2E Test | PARTIAL | Package suite green for all three (25/25; qwen3.8 focused 12/12 + hook/UI probes) and production `apps/web` build green (qwen3.7); no reviewer could run a real-browser Playwright E2E (jsdom-only, no octo-web Compose stack), no runtime defect confirmed |
| Security | CLEARED | 0 P1/P2 across all three reviewers; no XSS/injection (React auto-escaping, no `dangerouslySetInnerHTML`), no credential exposure, runtime auth boundaries zero-diff |

PR #984, opened 2026-07-22
fix(forward): align recent targets with sidebar ordering

### LINKED ISSUE #813: bug: 消息转发选择群「最近」列表数据源与智能总结不一致（显示不对）

liuhy2:
## Summary
- Align ForwardModal recent tab with SidebarService recent data
- Sort recent targets with pinned priority, then timestamp
- Render recent targets as a flat compact list with type metadata

Fixes #813

## Verification
- pnpm exec vitest run --environment jsdom packages/dmworkbase/src/Components/ForwardModal/__tests__/useForwardModal.test.tsx
- pnpm exec vitest run packages/dmworkbase/src/Components/ChatSelector/__tests__/tabFilter.test.ts
- pnpm i18n:check
- git diff --check origin/main...HEAD


lml2468 (approved):

审查锚定 head `a4d2a09d`。8 文件 +247/−49,把 ForwardModal「最近」Tab 对齐 SidebarService 权威 recent 数据 + 按「置顶优先、再时间戳」排序。本地跑测试 + build + 逐条核过。

### ✅ 两道闸门均通过

**规格符合**:与描述一致 —— 最近 Tab 数据源从本地会话派生改为 SidebarService `recent` 同步(与侧栏排序对齐),排序置顶优先再时间戳,扁平紧凑列表 + 类型元数据。无 over-build。

**代码质量(逐条核实)**:
- **数据源权威且隔离**:`Promise.all([follow, recent])` 各自 `.catch(() => null)` —— follow 失败不拖垮 recent,反之亦然;`recentSortMeta` 读的是 SidebarService **真实字段** `is_pinned:boolean`(:18)+`timestamp:number`(:16),非臆造;
- **排序 comparator 正确**:`if (aPinned !== bPinned) return aPinned ? -1 : 1; else bTs - aTs` —— 置顶优先、同置顶按时间倒序;`forwardItemKey`(:25,共享 accessor)一致取 key;
- **顺带修一处潜在 bug**:`sortConversations` 置顶加权 `1_000_000 → 1_000_000_000_000` —— 旧 1e6 小于毫秒时间戳(~1.7e12),置顶旧会话可能被较新未置顶会话反超;1e12 加权保证置顶恒赢;
- **降级一致**:`deviceId === ""` 时 recent/follow 都清空(与 follow 既有的 `validateSidebarRequest 必拒` 门控一致),空 deviceId → 空最近 Tab,graceful。

### ✅ 验证

- `@octo/base` useForwardModal + tabFilter + candidateToForwardItem → **44/44 全绿**;
- **i18n 齐全**:3 新键双 locale parity(en 1281 / zh 1281);
- `pnpm --filter @octo/web build` → **`✓ built in 3.39s`**;

### 🟡 非阻塞

- **最近 Tab 降级行为变化**:此前 recent 由本地会话派生(恒有内容),现改为 SidebarService recent,`deviceId` 为空/请求失败时 recent Tab 变空态(而非本地兜底)。与 follow 既有降级一致、不崩,但对齐 sidebar 权威顺序是本 PR 目标 —— 属有意取舍,知悉即可。

### 结论

最近 Tab 对齐权威 sidebar recent(follow/recent 各自 catch 隔离)、排序置顶优先+时间戳正确、`is_pinned/timestamp` 用真实字段、顺带修了 sortConversations 置顶加权被时间戳反超的潜在 bug;测试 44/44、i18n parity、build 通过。唯一 🟡 是 deviceId 空时最近 Tab 降级为空(与 follow 一致、有意)。无阻塞。**APPROVE**。

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #984 (octo-web)

## Summary

This PR migrates the ForwardModal "recent" tab from local conversation history to the `SidebarService` recent sync endpoint, bringing authoritative backend ordering (pinned-first, timestamp-descending) to the forward target list. It also introduces a flat list layout with inline metadata labels (DM / Group / Thread) for the recent tab, fixes the `PINNED_CONVERSATION_SCORE_BOOST` constant to be meaningful against millisecond timestamps, and updates the Storybook stories and tests accordingly.

## Verification

- ✅ **Key format parity** — `forwardItemKey` (`useForwardModal.ts:25`) produces `${channelType}::${channelID}` which aligns with the sidebar response's `${target_type}::${target_id}`. ChannelType values (1/2/5) match SidebarTargetType — verified against `tabFilter.ts:72` where `scopeByTab` matches `recentKeys` via the same accessor.
- ✅ **Error handling** — Both `SidebarService.sync` calls use individual `.catch(() => null)` inside `Promise.all` (`useForwardModal.ts:424-448`), so either or both failing degrades gracefully to empty sets. The `cancelled` flag (`:427`) prevents stale writes.
- ✅ **deviceId empty guard** — When `deviceId` is empty, all three state setters are called with empty values and the sync is skipped entirely (`:419-422`).
- ✅ **Sort fallback** — The recent-tab sort comparator (`:505-514`) handles missing `recentSortMeta` entries via `?.` and `?? 0`, falling back to `item.isPinned` from `channelInfo.top`.
- ✅ **PINNED_CONVERSATION_SCORE_BOOST** — Increased from `1_000_000` to `1_000_000_000_000` (`:46`), correctly ensuring pinned items always sort above unpinned ones in `sortConversations` (`:90-96`).
- ✅ **i18n parity** — Both `en-US.json:807-809` and `zh-CN.json:807-809` add the same three keys.
- ✅ **Test coverage** — Three new tests cover: sidebar-sourced recent set, pinned-before-unpinned sort, and local `channelInfo.top` fallback pinning.

Static analysis only at head `a4d2a09deb10c5ea217dc5448e73e021f10d352d`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. No P2 issues. No correctness, security, build, or functional concerns.

## Data Flow Trace

**`recentKeys` (Set)** — SidebarService.sync response items → `${target_type}::${target_id}` → `recentKeys` state → consumed by `scopeByTab` in `tabFilter.ts:67` as `opts.recentKeys.has(acc.getKey(i))` → `acc.getKey` = `forwardItemKey` = `${channelType}::${channelID}`. Key formats match. ✓

**`recentSortMeta` (Map)** — Same response items → key `${target_type}::${target_id}`, value `{timestamp, isPinned}` → consumed by sort comparator at `useForwardModal.ts:507-513` via `recentSortMeta.get(forwardItemKey(a))`. Lookup key format matches construction format. ✓

**`isPinned` on ForwardItem** — Three mapping sites: `channelInfoToForwardItem` (`:57`), `conversationWrapToForwardItem` (`:81`), `candidateToForwardItem` (line 51 of `candidateToForwardItem.ts`). All source from `channelInfo.top === true`. Consumed in sort comparator as `a.isPinned === true`, OR'd with sidebar `aMeta?.isPinned`. Both paths tested. ✓

**`filteredBase` → `filtered`** — `filterChatSelectorItems(allItems, {activeTab, keyword, followedKeys, recentKeys}, ACCESSORS)` returns scoped+keyword-filtered result. For non-recent tabs, used directly. For recent tab, re-sorted via `recentSortMeta`. Items in `filteredBase` for the recent tab must have keys in `recentKeys`, and `recentSortMeta` is populated from the same response — every item has a sort entry. ✓

**`flat` / `showMeta` props** — Derived from `recentFlatList = activeTab === "recent"` at `ForwardModal.tsx:217`, passed to `ItemRow` at `:275-276`. When true: flat layout (no child indent), inline kind label, small AI badge. When false: tree layout, Tag-based external badge, default AI badge. Both paths render correctly. ✓

## Blind Spot Checklist

- **C1 — Dual path parity**: `followed` and `recent` sidebar sync paths are symmetric — same error handling (`.catch(() => null)`), same key construction, same `cancelled` guard, same `deviceId` empty guard. Both state setters in same synchronous block. ✓ Clear.
- **C2 — Control-flow ordering**: `Promise.all` combines both syncs; `cancelled` flag checked once before any state write. No double-fire risk. `sortConversations` is called once during load, not re-triggered by sidebar sync. ✓ Clear.
- **C3–C4 — Authorization**: N/A. No auth/permission changes in this PR.
- **C5 — Build/runtime**: Static analysis only. CSS classes are additive, no removed selectors. i18n keys follow existing pattern.
- **C6 — Governance/docs**: N/A. No governance or strategy changes.

## Cross-round Blocker Re-check (R6)

- `forwardItemKey` key format matches sidebar response key format (channelType == SidebarTargetType).
- `Promise.all` with per-promise `.catch` prevents one failure from blocking the other.
- `cancelled` closure flag prevents stale state writes after unmount.
- `getKindLabel` and `forwardItemKind` use identical classification logic — no label-vs-filter mismatch.
- `AiBadge` rendering: `size="small"` in meta mode, default size in legacy mode — both gated by `showMeta`.
- External group badge renders in both the new meta layout and the legacy `<Tag>` layout.
- `candidateToForwardItem` sets `isPinned` from `cachedInfo?.top` — consistent with the two other mapping functions.
- Storybook `Interactive` component accepts `activeTab` via `Partial<ForwardModalProps>`.

## Verdict: APPROVED

Clean, well-tested migration from local-conversation-derived recent targets to backend-authoritative sidebar ordering. Data flow is sound, error paths degrade gracefully, and the UI changes are additive and isolated.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Data flow traced end-to-end from SidebarService response through key construction, filtering, sort, and rendering. All paths verified safe. Blind spot checklist clear.

Jerry-Xin (approved):
## ✅ Byte-verified review — LGTM

Aligns the ForwardModal **Recent** tab with the sidebar's ordering convention (pinned‑first, then timestamp DESC) by adopting the authoritative `SidebarService.sync(tab="recent")` set as scope + sort source. Fix achieves its stated goal; no correctness, dedup, XSS, or regression issues found. Build passes, targeted tests pass.

### Ordering matches sidebar (byte evidence)

- Sidebar (main chat list) sort: `packages/dmworkbase/src/Pages/Chat/vm.ts:482-503` — `score = timestamp + (extra.top === 1 ? TOP_CONVERSATION_SCORE_BOOST : 0)`, DESC.
- Forward Recent sort now: `packages/dmworkbase/src/Components/ForwardModal/useForwardModal.ts:504-513` — `if (aPinned !== bPinned) return aPinned ? -1 : 1; return (bMeta?.timestamp ?? 0) - (aMeta?.timestamp ?? 0)`, with `aPinned = recentSortMeta.get(key)?.isPinned || item.isPinned`.
- Pinned semantics equivalent: server `is_pinned` OR local `channelInfo.top === true` → both promote. Local `top` is threaded through both `channelInfoToForwardItem` (line 62) and `conversationWrapToForwardItem` (line 81) via new `isPinned` field, and also through the candidate path (`candidateToForwardItem.ts:51`).
- Same-priority within pinned/unpinned buckets is sorted purely by timestamp DESC — same key as sidebar. `Array.prototype.sort` in V8/JSC is stable (ES2019), so equal-key items keep the `filteredBase` order (which is the parent‑group → child‑thread tree order from `rebuildConvItems`).
- Sort‑constant harmonization (`PINNED_CONVERSATION_SCORE_BOOST = 1_000_000_000_000`, `useForwardModal.ts:46`) replaces the previous `1_000_000` boost which was too small to reliably beat modern epoch‑ms timestamps — this incidental fix removes a latent bug where a pinned older group could sink below a newer unpinned one in the local `sortConversations` helper.

### Identifier correctness — forwards to the right chat

- Recent items in the list are still drawn from `allItems` (`conversationItems ∪ friendItems ∪ searchGroupItems`); `filterChatSelectorItems` only *filters* by `recentKeys`, it never fabricates items from the sidebar response. Compound key `${channelType}::${channelID}` (`useForwardModal.ts:26`) matches `SidebarTargetType` (`Service/SidebarService.ts:3-7`) 1:1 (DM=1 / group=2 / thread=5), so cross‑type id collisions are prevented.
- `confirm()` resolves `channelID → Channel` via `channelMapRef` (`useForwardModal.ts:543-548`). Every source that can appear in `filteredBase` registers into `channelMapRef`: recents `useForwardModal.ts:228`, extraGroups `:273`, friends `:375`, search `:478`. A sidebar‑recent id that has no local ChannelInfo simply won't render (safe fail‑closed), so selection‑then‑confirm cannot leak to the wrong chat.

### De-dup

- `allItems = [...conversationItems, ...uniqueFriends, ...uniqueSearchGroups]` filters by `channelID` (`:490-493`). Friends already in conversations are dropped; search candidates already in local sources are dropped. React list keys are `item.channelID` (`ForwardModal.tsx:283/290/311`), and I found no `React duplicate key` warning path introduced.

### Data source consistency

- Both Recent and Followed now share one `Promise.all([follow, recent])` sync call (`useForwardModal.ts:420-449`), triggered once with proper `cancelled` guard, and both fall back to empty sets when `deviceId === ""` (matching the server's `validateSidebarRequest` guard). The old local-derived `recentKeys` write inside `rebuildConvItems` was removed cleanly (see the deleted `recentKeySet` accumulator).

### XSS

- Target names are rendered as React text children only: `ForwardModal.tsx:133` and `:179` (`<span>{item.displayName}</span>`). Grepped the whole `ForwardModal/` dir — zero `dangerouslySetInnerHTML`. React auto-escapes. Safe.

### No regression

- Tabs still work: `group` / `direct` scopes untouched. Search flow (`searchChatCandidates`) untouched. Multi-select / grant flow untouched. `TreeView` / `TreeViewWithLines` / `SearchTreeViewA` stories were switched to `activeTab="group"` — necessary because the default tab is `recent` and those stories exercise group‑tree layout (test-only story change, no runtime impact).
- Flat‑list `ItemRow` (`flat` + `showMeta` props) is a display‑only branch gated on `activeTab === "recent"`. Non‑recent tabs render the original layout unchanged (kept behind `!showMeta` branches at `:143` and `:157`).
- No dead code — all removed lines had explicit replacements.

### i18n parity

- `en-US.json` / `zh-CN.json` both add all three keys (`forwardModal.kindDirect` / `kindGroup` / `kindThread`). `pnpm i18n:check` → `locale keys healthy`.

### Verification I ran

| Check | Result |
|---|---|
| `gh pr checks` — required CI | Build in progress at review time, others SUCCESS / SKIPPED; mergeable=MERGEABLE, mergeStateStatus=BLOCKED (waiting on approval + Build) |
| `pnpm --filter @octo/web build` (Vite / rolldown) | ✓ built in 4.18s (no PR‑introduced errors) |
| `vitest run packages/dmworkbase/src/Components/ForwardModal/__tests__/useForwardModal.test.tsx` | 34/34 pass |
| `vitest run packages/dmworkbase/src/Components/ChatSelector/__tests__/tabFilter.test.ts` | 10/10 pass |
| `pnpm i18n:check` | pass |
| Full `@octo/base` suite | 1818 pass / 72 fail — **all 72 failures are pre-existing localStorage/jsdom env-noise on Node 26** in EmojiService / SessionScope / oidcLogout / I18nService / docLink / ClawOverviewTab; byte-verified to also fail on merge-base `d82645aa`. Zero failures in ForwardModal / tabFilter / any file touched by this PR. |

### 💬 Non‑blocking notes (optional polish, not requested changes)

1. **Env-noise test cleanup (repo-wide)** — the 72 pre-existing failures on Node 26 are noise for every reviewer. A follow-up to gate those on a `jsdom-localStorage` polyfill or pin the test env would materially improve signal.
2. **English kind labels** — `"DM" / "Grp" / "Thread"` (en-US.json:807-809). `"Grp"` is unusual; `"Group"` would be clearer at the ~70px meta max-width (measured `ForwardModal.css:150`, `max-width: 70px` with `overflow: hidden`). Non-blocking — abbreviations are a legitimate space call.
3. **`RecentFlatView` story** currently asserts only that the labels exist; consider also asserting the specific pinned-first ordering to lock down the sidebar-parity contract at the story layer (guards against future drift).

Approving. Good, well-tested fix.

yujiawei (approved):
# Code Review — PR #984 (octo-web)

**Verdict: APPROVE.** The change correctly aligns the ForwardModal "recent" tab with the authoritative backend recent set (`SidebarService.sync({ tab: "recent" })`), matching the existing summary selector (`ChatSelectorModal`) and resolving the reported data-source divergence. No P0/P1 blockers found. Findings below are all P2/advisory — none must be fixed before merge.

## 1. Spec compliance

The linked issue asks the forward "recent" tab to stop deriving its set from the local IM conversation list and instead use the backend recent source (scope by `recentIds`, sort by backend `timestamp` desc), mirroring the summary selector.

- **Implemented (nothing missing):** `useForwardModal.ts` now fetches both `follow` and `recent` via `SidebarService.sync`, derives `recentKeys` + `recentSortMeta` from the recent response, removes the old local `recentKeySet` derivation, and applies pinned-first + timestamp-desc ordering only for the recent tab. `follow` / `group` / `direct` / search / lazy-load / Space / grant paths are untouched (zero regression, as required).
- **No over-build:** The added flat/meta rendering in `ItemRow` and the type-label i18n keys are a reasonable UX complement to the flat recent list; they are gated behind `showMeta`/`flat` (recent tab only) and do not alter the other tabs' rendering.
- **Sibling parity confirmed:** `packages/dmworksummary/src/components/ChatSelectorModal.tsx` already uses the same backend recent source and the same empty-set-on-missing-device behavior — this PR brings forward into line with it.

Spec: ✅

## 2. Code quality

Verified locally at head `a4d2a09`: the targeted unit suites pass (`useForwardModal.test.tsx` 34/34, `ChatSelector/tabFilter.test.ts` 10/10) and `pnpm i18n:check` passes. TypeScript is clean on the changed files.

**a. `recent` sync is coupled to `follow` sync via a single `Promise.all`** — `useForwardModal.ts:424-449`. Both requests are `.catch(() => null)` guarded, so a rejection is safe, but a *slow/pending* `follow` request delays committing the already-resolved `recent` result (default tab), leaving the list empty for longer than necessary. Since both hit the same `/sidebar/sync` endpoint this is unlikely to matter in practice, and it mirrors the sibling selector's structure. Optional: settle each tab's state independently. *(Raised independently by both external review passes.)*

**b. Local `channelInfo.top` can override an explicit backend `is_pinned: false`** — `useForwardModal.ts:509-512`: `aPinned = aMeta?.isPinned === true || a.isPinned === true`. When the backend returns `is_pinned: false` for an item but the local cache has `top: true`, the item is treated as pinned and floats above a newer backend item. This is deliberate (a dedicated test, `useForwardModal.test.tsx:889`, codifies "local top counts as pinned when sidebar does not mark it") and only shifts relative ordering — no data loss or wrong target. Worth a short comment noting the intent, since a future reader may read it as the backend order being ignored. Consider preferring backend `is_pinned` when `recentSortMeta` has an entry for the key.

**c. New `RecentFlatView` story assertion will not match the rendered labels** — `ForwardModal.stories.tsx:308-310` asserts `getAllByText("群聊")`, but `getKindLabel` renders the `kindGroup` i18n value, which is `"群"` (zh) / `"Grp"` (en), never `"群聊"`. The assertion would throw on a real run. In practice this is inert: the storybook browser suite is not part of CI (`ci.yml` runs only `i18n:check` / `build` / `lint` / `lint:wkmodal`), and all ForwardModal play functions currently fail locally for a pre-existing reason (the i18n provider is not wired into the storybook preview, so translated strings don't resolve). Net effect: the story adds no executable coverage. If the intent was real coverage, wire i18n into the storybook preview and assert the actual label ("群"); otherwise the `useForwardModal.test.tsx` unit tests already cover the sort/scope logic well.

**d. Minor a11y nit** — `ForwardModal.tsx:142` renders the `·` separator as a text node; screen readers may voice it ("Group dot External"). Prefer a CSS `::before`/`::after` separator or `aria-hidden="true"`. Cosmetic.

Quality: Approved

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The findings above are all P2/advisory improvements, none blocking merge.

## 4. Coverage notes

- Verified: targeted unit tests (34 + 10 pass), `i18n:check` pass, tsc clean on changed files, sibling `ChatSelectorModal` parity, and the referenced issue's acceptance criteria.
- Not verified: real backend guarantees on `is_pinned` vs local `channelInfo.top` consistency; live cross-locale/layout rendering in a browser; the full `pnpm build` (env constraints) — but the change is type-clean and touches no build-affecting surface.

cnwenf (commented):
# Consolidated Review — PR #984 (octo-web)

> Aggregated by the review squad from **three independent reviews** (2 full code reviews + 1 isolated E2E run) conducted at head `a4d2a09deb10`. Findings confirmed by ≥1 reviewer are listed; where two reviewers independently flagged the same item it is noted, as that raises confidence. This is a consolidated **COMMENT** — both full reviews landed on COMMENT and E2E PASSed. **Note: this PR has already merged**, so the items below are follow-up recommendations rather than merge blockers.

## Overall assessment

- **Behavior is correct and well-tested on the happy path.** The recent tab correctly aligns to `SidebarService` ordering (pinned-first, then timestamp desc) and renders a flat compact list with type metadata. 34/34 `useForwardModal` tests and 10/10 `tabFilter` tests pass; i18n, TypeScript, and a 1770/1776 regression run are clean (6 failures pre-existing/unrelated).
- **E2E: PASS** — 8/8 probes in an isolated frozen checkout at the exact SHA (DM/Thread key alignment, pinned-first ordering, per-tab sync-failure isolation, empty-response handling, selection toggle). Preflight PARTIAL only because no browser/Storybook runner was available for rendered-DOM verification; hook/logic level fully verified.
- **Security: CLEARED** across all three reviews — no new auth surface, no credential exposure, no SSRF/injection/XSS vector. `device_uuid` is an IM device id sent over the existing authenticated channel; the backend response can only filter/reorder locally-present, space-scoped items.
- **The one substantive theme (raised independently by both full reviewers):** the default "recent" tab now goes **silently empty** where it previously fell back to local conversations.

---

## Findings

### P1 — CI never runs the test suite (pre-existing; amplified by this PR)
*`__tests__/useForwardModal.test.tsx`*

The CI workflow has no step that invokes `pnpm test` (or equivalent). Structured parse of `ci.yml` shows five run steps — install, `i18n:check`, build, lint, `lint:wkmodal` — none run tests; `turbo.json` has no test dependency; across 19 workflow files, zero invoke tests. Pre-existing gap, but this PR adds ~100 new test lines (34 tests) that will never be machine-checked in pipeline. **Recommend a follow-up to add a `pnpm test` step to CI.**

### P2 — Default "recent" tab silently empties on sync failure or missing `deviceId` *(confirmed by both full reviewers)*
*`useForwardModal.ts:417`, `:425`*

Pre-PR the recent set was derived from local conversations (`wrapsRef`), so the default tab always had content. Post-PR it comes entirely from `SidebarService.sync({ tab: "recent" })`, with failures swallowed by `.catch(() => null)`. Two reachable regressions:

- **Empty `deviceId`** (first render before device registration, offline, edge device states): the effect early-returns with empty `recentKeys`, so the default tab is permanently empty for that session.
- **Sync rejects** (network blip / 5xx): the rejection becomes an empty set with no `error`/`retry` signal. The hook exposes neither field, so "sync failed" is indistinguishable from "no recent conversations" — the default landing tab of a primary action just goes blank.

**Fix direction:** on failure/empty, fall back to local recent conversations (old behavior), or surface a retry/empty-state distinction; at minimum a `console.warn` with enough context to diagnose in production. *(E2E confirmed the empty states don't crash and other tabs stay functional — the concern is UX, not correctness.)*

### P2 — `loading` flips false before the sidebar sync populates the recent tab
*`useForwardModal.ts:381`, `:415`*

`loading` is owned only by the `load()` effect (local IM reads) and flips false when those resolve — typically faster than the network sync that feeds the *default* tab. There's a window where `loading=false` but `recentKeys` is still empty, so the user sees a non-loading empty "recent" tab that then populates. On the failure path this becomes permanent. Consider gating `loading` on both effects (or a short settled timer).

### P2 — Dual-source pin OR-merge *(confirmed by both full reviewers)*
*`useForwardModal.ts:509`*

`aMeta?.isPinned || a.isPinned` merges the sidebar's authoritative `is_pinned` with the possibly-stale local `channelInfo.top`. If they disagree, the item appears pinned and jumps to the top using the *sidebar* timestamp for tiebreak — an ordering neither source alone produces, and divergent from the sibling `ChatSelectorModal`, which sorts by timestamp only. **Recommend establishing source precedence** (sidebar-authoritative with local fallback, per the PR's stated "align with SidebarService" contract) **or documenting the OR-merge as intentional.**

### P3 — Sidebar sync effect has `[]` deps → recent set goes stale
*`useForwardModal.ts:415`*

The sidebar effect runs once on mount and never re-runs on space switch or `conversation-list-refreshed`, so `recentKeys`/`recentSortMeta` can drift out of sync with `conversationItems`. Also, if `deviceId` is empty at mount, the effect never re-runs when the device registers later. Add the relevant deps (or re-trigger on space/refresh events).

### P3 — `PINNED_CONVERSATION_SCORE_BOOST` (1e12) is below epoch-ms
*`useForwardModal.ts:94`*

`sortConversations` adds `1_000_000_000_000` to pinned items, but real ms-epoch timestamps are ~1.7e12, so a pinned item with an old timestamp can still be out-ranked by an unpinned recent one in the group/direct/allItems ordering. The recent tab gets this right with a dedicated two-key comparator; reuse that pattern instead of the additive boost.

---

## Nits

| # | Finding | Location |
|---|---------|----------|
| 1 | **Stale sort comment** — says "sorted by time" but caller now sorts pinned-first then timestamp *(both reviewers)* | `tabFilter.ts:12` |
| 2 | **Hardcoded CSS colors** — raw `rgba(...)` where repo convention expects `var(--wk-*)` tokens; no dark-mode variant *(both reviewers)* | `ForwardModal.css:150/157` |
| 3 | **Story assertion can't match any locale** — asserts `群聊` but i18n resolves to `群` (zh-CN) / `Grp` (en-US); latent only because CI runs no Storybook *(both reviewers)* | `ForwardModal.stories.tsx:308/310` |
| 4 | **Unknown type labeled "Grp"** — `getKindLabel` catch-all defaults unknown types to "Grp" silently; a `console.warn` or explicit "Unknown" would aid debugging | `ForwardModal.tsx:104` |
| 5 | **Missing outer `.catch()` on Promise chain** — `Promise.all` lost the outer catch that existed pre-PR; a synchronous throw in the `.then()` chain would produce an unhandled rejection | `useForwardModal.ts:449` |
| 6 | **Partial-failure path untested** — the `Promise.all` one-rejects-others-resolve branch isn't exercised; code handles it, but a test would lock the contract | `useForwardModal.ts:424` |
| 7 | **Invariant untested** — `recentKeys`/`recentSortMeta` length invariant isn't locked (mutation confirms all 34 tests still pass when broken) | `useForwardModal.ts:505` |
| 8 | **Sort fallback coupling** — `?? 0` fallback is structurally unreachable but relies on unenforced coupling between sort-key provision and metadata population | `useForwardModal.ts:512` |

---

## Things all three reviewers verified as fine

- Composite-key alignment: `${target_type}::${target_id}` (sidebar) matches `${channelType}::${channelID}` (`forwardItemKey`); `SidebarTargetType` (1/2/5) equals the channel-type constants — DM/Group/Thread all matched.
- Recent ordering logic (pinned-first then timestamp desc) is correct with dedicated reproducers.
- Per-promise `.catch(() => null)` correctly isolates `follow` and `recent` syncs — one failing doesn't blank the other.
- Cleanup: `cancelled` flag prevents post-unmount `setState`; `[...filteredBase]` sort copy avoids mutating the source.
- i18n parity (`kindDirect`/`kindGroup`/`kindThread` in both locales); TypeScript clean for changed files.
- Security CLEARED across STRIDE + OWASP in all three reviews.

## Verdict: COMMENT

Happy-path logic is correct and well-tested; security and key-alignment posture are clean; E2E PASS. Not blocking (already merged). The P2 failure-path items — silent empty default tab on sync failure/missing `deviceId`, and `loading` flipping early — are the highest-value follow-ups; the P1 CI gap is worth its own tracking issue since it leaves all new tests unverified in pipeline.

PR #988, opened 2026-07-22
feat(chat): allow re-editing recalled messages

botshen:
## Summary

- show an explicit **Re-edit** action after the current user recalls their own message
- restore text, mentions, rich-text images, and files into the composer without overwriting the current draft
- continue restoring later blocks when one block fails, show localized feedback, and always return focus to the composer
- rebase whole-message rich-text mention entities per block so mentions after images/files retain the correct recipient
- restore rich-text file attachments from each block's own URL while preserving the safe URL checks and placeholder fallback
- handle unavailable attachments safely and add focused regression coverage

## Scope

This PR only changes the recalled-message placeholder and composer restore path. It does not change recall permissions, recall time limits, or the existing message edit flow.

## Verification

- `pnpm exec vitest run packages/dmworkbase/src/Messages/Revoke/reeditableMessage.test.ts packages/dmworkbase/src/Components/MessageInput/__tests__/mentionSendParse.test.ts packages/dmworkbase/src/Components/MessageInput/__tests__/richTextPaste.test.ts` (59 tests)
- `pnpm i18n:check`
- `pnpm exec stylelint packages/dmworkbase/src/Messages/Revoke/index.css --config stylelint.config.mjs`
- `pnpm --filter @octo/web build`
- `git diff --check`

Closes #979

PR #994, opened 2026-07-22
feat(chat): allow re-editing recalled messages

### LINKED ISSUE #979: ✨ [Web] 撤回的消息支持回填输入框重新编辑

botshen:
## Summary

- show an explicit **Re-edit** action after the current user recalls their own message
- restore text, mentions, rich-text images, and files into the composer without overwriting the current draft
- continue restoring later blocks when one block fails, show localized feedback, and always return focus to the composer
- rebase whole-message rich-text mention entities per block so mentions after images/files retain the correct recipient
- restore rich-text file attachments from each block's own URL while preserving the safe URL checks and placeholder fallback
- handle unavailable attachments safely and add focused regression coverage

## Scope

This PR only changes the recalled-message placeholder and composer restore path. It does not change recall permissions, recall time limits, or the existing message edit flow.

## Verification

- `pnpm exec vitest run packages/dmworkbase/src/Messages/Revoke/reeditableMessage.test.ts packages/dmworkbase/src/Components/MessageInput/__tests__/mentionSendParse.test.ts packages/dmworkbase/src/Components/MessageInput/__tests__/richTextPaste.test.ts` (59 tests)
- `pnpm i18n:check`
- `pnpm exec stylelint packages/dmworkbase/src/Messages/Revoke/index.css --config stylelint.config.mjs`
- `git diff --check`

Closes #979


lml2468 (approved):

审查锚定 head `c82e88c7`。10 文件 +901/−6,给用户自己撤回的消息加「重新编辑」入口。

> 说明:#994 是已 CLOSED 的 **#988 的重开继任**(同分支 `codex/issue-979-reedit-revoked-message`、同 head)。相对我在 #988 APPROVE 的 `3430605e`,本 head 增量修掉了 **mochashanyao 在 #988 提的富文本 file 块 URL 解析 🔴**。#988 已核核心(所有权/mention 偏移重基/per-block 弹性)保持。

### 🔴→✅ mochashanyao #988 CR:富文本 file 块 URL 解析错、带文件附件重编辑失败 —— 已修

- **根因**:`getReeditableRichTextBlocks` 的 file 块此前只还原成文本占位符 `[文件] name`(丢真实附件);且 Conversation restore 用 `message.content as FileContent` 取 URL —— 但多块富文本消息的 `content` 是整个 RichTextContent、非该 file 块 → URL 解析错;
- **修法(核过正确)**:
  - reeditableMessage.ts:203-215:file 块有 `block.url` 时还原成真正的 `{ type:"file", url, name, size, mime }`(mime 缺省时 `application/{extension}` 兜底),无 URL 才回退文本占位符;
  - Conversation/index.tsx:691:`resolveSafeFileUrl(block)`(按**每块**源解析,替代错误的 `message.content`)—— 复用与图片同一套 `isSafeUrl`(http/https)+ baseURL 绝对化安全解析器(`resolveSafeFileUrl` 签名放宽为 `FileUrlSource`);
  - `remoteReeditFileToFile({...block, url: fileUrl})` 拉取+附加,无 url 则 null fail-soft。
- **🔒 安全保持**:file 块 URL 走 `isSafeUrl` 门控(与图片对称),不会 re-attach `javascript:`/不安全 scheme。

### ✅ #988 已核核心保持(复核)

- **所有权双门控** `revoker !== myUid || fromUID !== myUid`(:87);**mention 偏移按块重基** `rebaseMentionEntitiesForBlock`(:232);**per-block 弹性** `restoreReeditableMessageBlocks`(:64);mention 结构化节点、草稿保留、附件 fail-soft —— 均在。

### ✅ 验证

- `@octo/base` reeditableMessage + mentionSendParse + richTextPaste → **59/59 全绿**(比 #988 +1,新增「file block 从自身 URL 还原」+ 无 URL 回退占位符);
- `pnpm --filter @octo/web build` → **`✓ built in 3.51s`**;i18n parity;
- 已 rebase(behind 6)、`git merge-tree` **0 冲突**、MERGEABLE。

### 结论

#988 已验证的功能(所有权/mention 偏移重基/per-block 弹性)全部保持,且补上 mochashanyao 的 file 块 URL 修复:file 块按每块源经 `resolveSafeFileUrl`(isSafeUrl 门控)真正 re-attach、无 URL fail-soft 回退占位符;测试 59/59、build 通过、无冲突。无阻塞。**APPROVE**。感谢 mochashanyao 抓出 file 块 URL 这条。

> 治理提示:#994 与 CLOSED 的 #988 同分支同改动,建议合 #994、确认 #988 保持关闭。

Jerry-Xin (approved):
## ✅ APPROVE

Re-review of the recalled-message re-edit feature (successor of the now-CLOSED #988; same branch, same head, same lineage). Verified against live head `c82e88c7`.

### The contested blocker — rich-text FILE block re-edit — byte-verified as FIXED at this head ✅
The previously raised 🔴 ("re-editing a rich-text message with a file attachment fails / drops the file") was **real against the pre-fix state**, but the head commit `c82e88c7` ("restore rich-text file attachments") resolves it. Evidence:

- `Messages/Revoke/reeditableMessage.ts:214-227` — the rich-text `file` block handler no longer degrades to placeholder text. When `block.url` is present it now emits a real `{ type: "file", url: block.url, name, size, mime }` reedit block (mime derived from `block.mime` or `application/{extension}`); only the URL-less case falls back to the `[文件] name` placeholder. The file block's URL is taken **directly from the block's own `url` field** — there is no fragile parse of the `[文件] name` placeholder string, so the earlier "URL parse error" class does not apply here.
- `Messages/RichText/RichTextContent.ts` — `RichTextBlock.url` and `decodeJSON` map `url` for every block type (text/image/file), so a decoded file block genuinely carries `url`. The block-to-reedit mapping reads an existing field, not a reconstructed one.
- `Messages/File/index.tsx:308` — `resolveSafeFileUrl` was widened from `FileContent` to a minimal `FileUrlSource` (`{ url?, remoteUrl? }`), so a reedit file block flows through the **same** absolutize + `isSafeUrl` (http/https allowlist) path as a normal file cell.
- `Components/Conversation/index.tsx:691-707` — restore handler consumes it end to end: `resolveSafeFileUrl(block)` → `remoteReeditFileToFile({ ...block, url })` → `addAttachment([file], "upload")`, with a localized toast on failure.
- Test coverage is explicit: `reeditableMessage.test.ts:188` ("restores a rich-text file block from its own URL") and the URL-less fallback in the mention-offset case at `:224`. **The re-approve at this head is correct.**

### Lineage findings re-verified on `c82e88c7` bytes — all intact
- **Ownership double-gate** — `reeditableMessage.ts:79-83`: requires `!message.revoke → false`, and `revoker !== myUid || fromUID !== myUid → false` (admin-recalled / others' messages excluded). Re-checked at the top of `reeditRevokedMessage` (`Conversation/index.tsx:661`) before any restore, and the button only renders when the gate passes (`Revoke/index.tsx`).
- **Re-send is a NEW message** — restore only re-populates the composer (`insertContent` / `addAttachment`); no `message_id` / `client_msg_no` reuse, no overwrite of the recalled record.
- **XSS / attachment safety** — content restored as structured Tiptap `JSONContent` (no raw HTML injection); URLs gated by `isSafeUrl` (http/https only); `remoteReeditFileToFile` uses `credentials: "omit"`, enforces the 100MB cap via both `Content-Length` and blob size, and sanitizes the filename (`[\\/:*?"<>|]` → `_`, 160-char clamp). Image path keeps its own placeholder + toast fallback.
- **Per-block resilience** — `restoreReeditableMessageBlocks` wraps each block in try/catch → `onBlockError`, with `onComplete` in `finally` restoring composer focus.
- **Mention offset per-block rebasing** — `rebaseMentionEntitiesForBlock` advances `plainOffset` by image placeholder length, file label length, and text length; filters entities to `[blockOffset, blockEnd]`, rebases to block-local offsets, and forces `uids: []` so structured entities stay authoritative. Verified by the offset-accounting test at `reeditableMessage.test.ts:211`.

### What the added code covers vs the earlier iteration
Beyond the earlier state, this head adds: (1) the real file-block restore branch, (2) the `resolveSafeFileUrl` signature widening + `remoteReeditFileToFile` file-fetch path with size/MIME/filename guards, (3) the `insertContent` composer API + `addAttachment` async signature, (4) the `reeditRevokedMessage` context method and the Re-edit button, and (5) matching regression tests. No dead code spotted; the new exports are all wired to real call sites.

### Gate / build / tests / i18n
- **CI**: green / skipped; **mergeable = MERGEABLE**.
- **Build**: `pnpm --filter @octo/web build` ✅ (only pre-existing chunk-size / vendor `eval` warnings).
- **Tests**: `vitest run` on the three targeted suites → **59/59 passed** (the localStorage notice is Node-26 env noise, not a failure).
- **i18n parity**: 4 new keys added to both `en-US.json` and `zh-CN.json` (`revoke.reedit`, `reeditAttachmentFailed`, `reeditBlockFailed`, `reeditFileFailed`) — no missing side.

### Highlights
- File-block restore reuses the exact same safe-URL resolution as normal file rendering rather than a bespoke parser — the cleanest way to close the reported concern.
- Structured Tiptap restore (no HTML) plus per-block isolation keeps the failure mode graceful (partial restore + focus return) instead of an all-or-nothing throw.

No blocking items. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #994 (octo-web)

## Summary

This PR adds the ability for users to re-edit their own recalled messages. A "Re-edit" button appears on self-revoked message cells, which extracts the original message content (plain text, rich text, images, and files) and restores it into the message composer. The implementation includes a new `reeditableMessage.ts` module with block extraction, mention entity rebasing, and sequential async restoration, plus integration into the `Conversation` component and `RevokeCell`.

The design is sound: permission gating is strict (only original sender who self-revoked), error handling is per-block with continue-on-failure semantics, and the test suite covers the core helpers comprehensively including edge cases for mention rebasing across rich-text blocks.

## Verification

- ✅ **Auth/visibility gate** — `canReeditRevokedMessage` at `reeditableMessage.ts:88` requires `revoker === myUid && fromUID === myUid`, correctly preventing anyone other than the original sender who self-revoked from re-editing.
- ✅ **Context wiring** — `ConversationContext` interface (`context.ts:46-48`) declares `reeditRevokedMessage`; `Conversation` class implements it at `:660`; `RevokeCell` receives `context={this}` and calls it at `:87-89`. Full chain verified.
- ✅ **URL safety** — File re-download path goes through `resolveSafeFileUrl` (calls `isSafeUrl`); image path goes through `imageBlockToPasteFile` which also calls `isSafeUrl` before fetch. Both use `credentials: "omit"`.
- ✅ **Error resilience** — `restoreReeditableMessageBlocks` wraps each block in try/catch with `onBlockError`, and `finally` ensures `onComplete` always fires. Tested in `reeditableMessage.test.ts`.
- ✅ **Mention rebasing** — `rebaseMentionEntitiesForBlock` correctly filters entities to those fully within each text block and rebases offsets. Covered by test "keeps whole-message mention entities assigned to their text blocks".
- ✅ **i18n parity** — All 4 new keys (`reedit`, `reeditAttachmentFailed`, `reeditBlockFailed`, `reeditFileFailed`) present in both `en-US.json` and `zh-CN.json`.
- ✅ **Type widening safe** — `resolveSafeFileUrl` parameter widened from `FileContent` to `FileUrlSource` (`{url?, remoteUrl?}`); existing callers pass `FileContent` which satisfies the supertype.
- ✅ **`addAttachment` return type** — Widened from `void` to `void | Promise<void>`; `reeditRevokedMessage` correctly `await`s both calls.

Static analysis only at head `c82e88c7`; build and tests not executed in this environment.

## Findings

### P2 — No fetch timeout in `remoteReeditFileToFile` (`reeditableMessage.ts:137`)

The `fetch()` call in `remoteReeditFileToFile` has no `AbortSignal` or timeout. If the CDN responds slowly or hangs, the re-edit operation blocks indefinitely — the `try/catch` in `restoreReeditableMessageBlocks` only catches thrown errors, not stalled promises. The user sees no feedback until the fetch eventually fails or succeeds. Consider adding `AbortSignal.timeout(30_000)` or wrapping with an `AbortController` + `setTimeout`.

### P2 — Legacy mention regex misses non-CJK-Unicode scripts (`reeditableMessage.ts:311`)

`appendTextWithLegacyMentions` uses `/@[\w\u4e00-\u9fa5.\-]+/gm` to match mention labels. This covers ASCII + CJK Unified Ideographs but excludes CJK Extension A/B ranges, Japanese hiragana/katakana, Korean hangul, and other scripts. When `entities` is empty and the code falls back to `uids`-based matching, mention labels with these scripts are silently rendered as plain text instead of mention nodes. The entity-based `buildInlineContent` path is unaffected.

### Nit — Stale type annotation on `_addAttachmentFn` (`Conversation/index.tsx:458`)

The field is typed as `(files, source?) => void` but the underlying `MessageInput.addAttachment` is now `async (...) => Promise<void>`. No runtime impact since the reedit path uses `MessageInputContext.addAttachment` directly, but the drift may confuse future maintainers.

- `MessageCell` base class declares `context: ConversationContext` in props — `RevokeCell` correctly accesses it.
- Rich-text block iteration preserves order; image and file blocks are extracted separately from text blocks with correct `plainOffset` tracking for mention rebasing.
- `remoteReeditFileToFile` has triple size validation: pre-fetch (`input.size`), post-headers (`Content-Length`), and post-blob (`blob.size`), all against `MAX_REEDIT_FILE_BYTES` (100MB).
- `normalizeMentionUid` correctly maps legacy/broadcast UIDs to `MENTION_UID_HUMANS`.
- Test coverage is thorough: 7 test cases covering permission gating, text/richtext/image/file restoration, mention rebasing, and error recovery.

## Verdict: APPROVED

Well-structured feature addition with solid error handling and comprehensive tests. The two P2 items (fetch timeout and legacy mention regex) are non-blocking quality improvements.

yujiawei (approved):
# Code Review — PR #994 (octo-web)

**Verdict: APPROVE.** The change is well-scoped, correct on the load-bearing path (mention routing), and backed by focused regression tests. All findings below are non-blocking (P2 / nits).

## Spec compliance

Matches the stated scope: a **Re-edit** action on self-revoked messages that restores text, mentions, rich-text images, and files into the composer without overwriting the current draft, degrading gracefully on failure. No recall-permission, recall-time-limit, or edit-flow changes leaked in. No missing / over-built / diverging behavior found.

- Sender gating is correct: `canReeditRevokedMessage` requires `revoker === myUid && fromUID === myUid` (`reeditableMessage.ts`), so only the original sender who recalled their own message sees the button.
- i18n keys added symmetrically in both `en-US.json` and `zh-CN.json`; CSS uses existing theme tokens.

## Correctness — mention routing (the critical path) ✔

The high-risk part of any "re-edit a sent message" feature is silently dropping `@mention` recipient routing when reconstructing composer content from display text. This PR handles it correctly:

- Mentions are rebuilt from the structured `mention.entities` (`{uid, offset, length}`), **not** re-parsed from display text. `buildInlineContent` validates `offset + length <= text.length` and dedups overlaps.
- For rich text, `rebaseMentionEntitiesForBlock` rebases entity offsets per block and sets `uids: []` so structured entities stay authoritative (no positional fallback that could bind the wrong recipient after an image/file placeholder).
- Offset parity with the send side is exact: the send serializer computes entity offsets against the flattened display text (`mentionSendParse.ts`), and rich-text send computes them against the placeholder-joined plain string (`[图片]` for images, `[文件] name` for files). The restore path's `plainOffset` accumulation mirrors `buildRichTextPlain` byte-for-byte, so entities land on the correct spans on re-send.
- Text-message restore has two branches; both preserve routing. When `parts` don't reconstruct the text, it falls back to `buildInlineContent(text, readMessageMention(...))`, and `readMessageMention` merges `mention.entities` from `contentObj.mention` — which is where the SDK keeps entities after decode. Routing is not lost in either branch.

Both attachment-restore URLs remain behind the existing safe-URL checks (`imageBlockToPasteFile` → `isSafeUrl`; files → `resolveSafeFileUrl`), and partial failures fall through per-block (`restoreReeditableMessageBlocks` continues on error and always restores focus).

## Non-blocking findings (P2 / nits)

1. **Re-edit button has no in-flight guard** (`Messages/Revoke/index.tsx`). Double-clicking fires `reeditRevokedMessage` concurrently, which can duplicate composer content and kick off duplicate attachment downloads. Not data-loss and nothing auto-sends (the user reviews the composer before sending), so it's a UX nit — consider disabling the button while a restore is in flight.

2. **`remoteReeditFileToFile` buffers the full body before the size check** (`reeditableMessage.ts`). The `Content-Length` pre-check is bypassable by a response with no `Content-Length`, and `await response.blob()` then buffers everything before `blob.size` is checked — an OOM risk on a very large body. The image path already solves this by streaming with an incremental cap (`responseToCappedImageBlob`); consider reusing that approach for files.

3. **Broadcast-label case mismatch** (`appendTextWithBroadcastMentions`). The match regex uses the `gi` flag but the AIs/humans discrimination is a case-sensitive `match[0] === \`@${MENTION_LABEL_AIS}\``, so a lowercased broadcast label would misclassify as humans. Reachability is near-zero because the canonical sent text always carries the exact-case label, but a `.toLowerCase()` comparison would remove the trap.

4. **Whitespace-only rich-text text blocks are dropped** (`text.trim() !== ""`). Pure spacing/newlines separating adjacent attachments won't be restored. Minor fidelity nit.

## Coverage note

Tests cover entity rebasing, block ordering, broadcast sentinels, file-block restore, and the continue-on-error path (42 related tests pass locally). Not covered: double-click concurrency, missing-`Content-Length` size enforcement, and whitespace-only blocks — all matching the P2 findings above. Runtime file-backend auth (whether `credentials: "omit"` suffices for private attachments) can't be judged from the diff; it matches the existing paste-restore convention.

PR #956, opened 2026-07-21
feat(mcp-market): import MCP from JSON (lightweight alternative to #867)

l-s-c:
## Summary

Adds a **"从 JSON 导入"** mode to the MCP market create modal. Users paste an `mcpServers` config (Claude Desktop / Cursor format) or a flat `server.json` snippet; a pure parser extracts `name / slug / command / args / env keys / url / header keys / transport` and seeds the existing 3-step manual wizard, then flows through the existing probe + publish paths. Env / header **values are always dropped** for privacy — users refill secrets in step 2.

Also included in this PR:

- **Detail modal token highlight** — the `<把这里换成你的 Token>` placeholder in the QuickStart snippet now renders as an amber `<mark>` so users don't miss the swap-in step on copy-paste.
- **MarketSidebar cold-load fix** (`apps/web/src/Pages/Main/vm.ts`) — refreshing / bookmarking `/mcp-market/mcp` used to land on an empty right pane because the `WKApp.currentMenuId` global wasn't written by `didMount`'s implicit setter path. Now it is.

## Linked Spec

- Issue: #867 — *feat(mcp-market): 从通用 Git 仓库导入 MCP 并生成可发布草稿*
- **Deliberate scope narrowing**: this PR is the **modelscope-style lightweight path** (paste JSON → seed form → existing publish flow), **not** the full P0 in #867 (public HTTPS Git fetch → SHA lock → candidate selection → confidence-tagged draft). #867 **stays open**; the heavier Git-fetch path is a separate PR.

## Out of scope

- No Git URL fetch, no async task API, no SHA locking, no monorepo candidate detection, no field-level confidence indicators, no diff-view re-import. These are the deliberate cuts.
- Manual create flow is not restructured — JSON import is a new mode toggle on top of it.

## How verified

- **Unit tests**: 22 new tests for `parseImportJSON` covering happy path, all fatal error paths (empty / invalid JSON / non-object / bad `mcpServers` shape / no recognized fields), placeholder detection (angle-brackets, `\${VAR}`, `sk-xxx`, `Bearer <TOKEN>`), multi-server warning, args coercion (numbers / booleans / objects / empty), slug slugification of `@scope/name` and non-ASCII wrapper keys, transport inference from URL / command / explicit field. Package total: 51/51 green (\`pnpm --filter @dmwork/mcp test\`).
- **Build**: \`pnpm --filter @octo/web build\` → \`built in 5.84s\`, no errors.
[...]

PR #15, opened 2026-05-13
ci: add Octo issue/PR feed workflows

lml2468:
Add real-time GitHub event notification to Octo IM channels via GitHub Actions.

- `octo-issue-feed.yml`: pushes Issue events (opened/closed/reopened/labeled) to Octo 🛠️ issue-feed and the repo's project channel
- `octo-pr-feed.yml`: pushes PR events (opened/closed/merged/etc.) to Octo 🛠️ pr-feed and the repo's project channel

PR #995, opened 2026-07-22
feat(mcp): per-key user-supplied toggle in MCP create/edit form

l-s-c:
## Summary

Replaces the free-text env/headers textarea with a structured KV row editor that carries a "consumer fills locally" toggle per key, aligned with the marketplace §5.1 wire contract (`env_user_supplied` / `headers_user_supplied` arrays). Drops the `auth_type` selector and the ephemeral probe-bearer input — Bearer auth is now expressed as an `Authorization` row + toggle ON.

## Related Issue

Depends on octo-marketplace PR https://github.com/Mininglamp-OSS/octo-marketplace/pull/15 (wire contract source of truth).

## Changes

- `KvEntry` model + `entriesFromWire` / `entriesToWire` helpers; sentinel stripped unconditionally on rehydrate so legacy records render cleanly.
- `KvEditor` row component (Semi Switch per key).
- `detailToForm` reads `*_user_supplied` arrays to rebuild toggle state; probe path forwards real values (probe is off-record).
- `quickStartTemplates` renders placeholder for user-supplied keys via `applyUserSuppliedPlaceholder`; the old `authType`-based branch is gone.
- JSON import seeds new rows with smart-default toggle by key name (matches backend `secretKeyPattern`).
- List page / detail modal / card badge polish.
- i18n + CSS + tests updated; obsolete `authType` / `probeBearer` keys removed.

- Affected module(s): `packages/dmworkmcp` (MCP marketplace UI); minor `packages/dmworkskillmarket` CSS touch-up.
- New or changed user-visible entry point: no new entry point; the create/edit modal is reshaped.
- Shared layer touched: none (self-contained).
- Duplicate entry point checked: yes; no duplication.

## Testing

- \`pnpm --filter @dmwork/mcp test\` — 4 files / 56 tests passing.
- \`pnpm --filter web build\` — green.
- Manually: opened create + edit modal, exercised KvEditor add / remove / toggle, JSON import with mixed keys, probe against a mock endpoint. Detail modal renders user-supplied keys as placeholder.

- [x] Manually verified

## Checklist

- [x] Updated documentation (mcp-v1.md changed in the backend PR)
- [x] Ran the mcp package tests
- [x] Described impact scope


lml2468 (dismissed):

审查锚定 head `e9f04d93`。17 文件 +858/−414,把 MCP env/headers 从自由文本 textarea 换成结构化 KV 行编辑器 + 每键「消费方本地填」开关(对齐 marketplace §5.1 `*_user_supplied` 契约),移除 `auth_type`/probe-bearer。核心是密钥隐私模型,逐条核过。

### ✅ 两道闸门均通过

**规格符合**:与描述/§5.1 契约一致 —— KvEntry 模型 + `entriesFromWire`/`entriesToWire`、KvEditor 行组件、Bearer 改为 `Authorization` 行 + 开关。依赖 octo-marketplace PR #15(wire 契约真相源)。无 over-build。

**代码质量 —— 密钥隐私模型(核心,逐条核实)**:
- **隐私模型有意变更(§5.1 relaxation)**:与 #956「值一律丢弃」不同,本 PR **值会持久化**(`entriesToWire` `values[k]=e.value`)—— 文档化理由:owner 自己 re-edit 能看回值(§5.1)、**非 owner 读由服务端 blank(§5.3)**、市场 snippet 客户端 placeholder 化;
- **🔒 市场 snippet 客户端不泄真值**:`applyUserSuppliedPlaceholder`(quickStartTemplates:85/66)把 user-supplied 键的值替换成 `TOKEN_PLACEHOLDER`(env+headers)→ 发布的 copy-ready snippet **绝不暴露真实 user-supplied 密钥值**;非 user-supplied(共享公开配置)原样;
- **rehydrate 干净**:`entriesFromWire` 把 legacy 记录里的 `SECRET_PLACEHOLDER_SENTINEL` 归一为 `""`,不显示 `__OCTO_SECRET_...`;`detailToForm` 读 `qs.envUserSupplied`/`headersUserSupplied` 重建开关态;
- **无客户端泄露面**:McpCard/McpDetailModal 不 raw-render env/header 值;
- **JSON import smart-default**:新键按 `SECRET_KEY_PATTERN`/`isSecretKey` 默认开开关(粘贴 Authorization 不会误发共享空值);
- **obsolete 清理干净**:`authType`/`probeBearer`/`auth_type` 全仓无 dangling 引用。

### ✅ 验证

- `@dmwork/mcp` → **56/56 全绿**(4 文件);
- i18n en/zh parity(215 键);`pnpm --filter @octo/web build` → **`✓ built in 4.26s`**;
- 已 rebase(behind 6)、`git merge-tree` **0 冲突**、MERGEABLE。

### 🟡 非阻塞(跨仓契约依赖,合前确认)

- **非 owner 密钥保护现依赖后端 §5.3 blanking**:本 PR 客户端**发送真实值**(§5.1 relaxation),非 owner 读取被 blank 是**服务端**职责。前端侧无泄露(snippet 已 placeholder、无 card/detail raw render),但端到端隐私取决于 octo-marketplace PR #15 的 §5.3 服务端 blanking 已部署 —— 合并前请确认后端契约已上线,否则非 owner 读 MCP 记录可能拿到真值。属跨仓依赖、非本客户端 PR 能强制,已如实标注。

### 结论

结构化 KV + per-key user-supplied 开关对齐 §5.1;客户端隐私面处理正确(snippet placeholder 化、sentinel 归一、无 card 泄露、JSON import 智能默认、obsolete 清净);测试 56/56、i18n parity、build 通过。唯一需注意的是端到端非-owner 密钥保护依赖后端 §5.3 blanking(跨仓契约,合前确认)。前端侧无阻塞。**APPROVE**。

Jerry-Xin (dismissed):
**APPROVE** — byte-verified. Clean replacement of the free-text env/headers textarea + `authType`/probe-bearer inputs with a structured per-row KV editor carrying a per-key "user fills locally" toggle, wired to `env_user_supplied` / `headers_user_supplied`. No credential-handling regressions; build + tests green; i18n parity intact.

### Credential / secret handling (the primary concern) — verified

The design here is **intentional and documented**: under the §5.1 relaxation, a user-supplied key's VALUE is still persisted verbatim (so the OWNER sees it again on their own edit), and **non-owner reads are blanked server-side (§5.3)**. The frontend is deliberately NOT the guard. Traced form state → save payload:

- `entriesToWire` (`components/McpCreateModal.tsx`) emits `values[key]=value` for every non-empty-key row and pushes the key into `userSupplied[]` iff the toggle is on. The flag stays attached to the correct key; empty-key rows are dropped so a stray add doesn't emit `{"":""}`.
- Marketplace snippet masking is client-side and keyed off the user-supplied set, not a name heuristic: `applyUserSuppliedPlaceholder` (`api/quickStartTemplates.ts`) substitutes the token placeholder only for keys in `*_user_supplied`; shared keys pass through verbatim (as intended for a copy-paste config). The old `isSecretKey`-based masking is gone.
- Non-owner exposure: the detail modal renders the connection snippet through `buildQuickStartTabs` → `applyUserSuppliedPlaceholder`, so a user-supplied value never surfaces in the snippet. The verbatim-value edit path (`detailToForm`) is only reachable via the owner-gated edit action.
- Legacy records: `entriesFromWire` normalizes a persisted sentinel literal back to `""` so old rows render as blank inputs, not raw sentinel text.

### Per-key toggle keying — verified (no off-by-one)

`KvEditor.update(idx, patch)` maps `entries` and only rewrites the row at `idx`; `remove(idx)` filters exactly that index; the toggle's `onChange` calls `update(idx, { userSupplied })`. Toggling row A cannot flip row B.

### Default safety — verified

- Manual add → toggle OFF (shared). Safe under the §5.3 server-side non-owner blanking model.
- JSON-import seed → smart-default ON for secret-named keys (`userSupplied: isSecretKey(k)`), so a pasted config with an `Authorization` header doesn't publish a shared blank.
- Init-from-existing on edit: toggle state rebuilt from the `*_user_supplied` arrays via `entriesFromWire`.

### Other checks

- **XSS**: key names / descriptions render as React text nodes; no `dangerouslySetInnerHTML`. Clean.
- **Import discipline (matches the prior JSON-import PR)**: `importJson.ts` still lifts KEYS only and drops values; placeholder-value detection avoids false "values dropped" warnings.
- **Real credentials in fixtures**: none. Test/mock values are placeholders (`<YOUR_TOKEN>`, `<your-key>`, `<xoxb-...>`) or obvious dummies used specifically to assert the value is NOT imported. `secret-scan / Secret Scan (gitleaks)` = SUCCESS, OSV green.
- **i18n**: every new key (`total`, `searchClear`, `kvUserSuppliedToggle`, `probeTruncated`, import `format`/`formatFailed`, add/remove/placeholder keys) present in BOTH en-US and zh-CN; obsolete `authType`/`probeBearer`/`secretLeaked`/`mineTotal` keys removed from both.
- **Dead code**: obsolete helpers (`parseKV`/`serializeKV`/`maskSecrets`/`mergeProbeBearer`/`applySecretSentinel`/`stripSecretSentinel`) removed; `isSecretKey` now used only as the import smart-default seed.
- **Build**: `web` build passes (the residual eval/chunk-size warnings are pre-existing third-party notices, unrelated to this PR).
- **Tests**: `@dmwork/mcp` — 4 files / 56 tests pass, including the rewritten shared-vs-user-supplied placeholder cases.
- **Probe path**: probe forwards real values (off-record transient request), separate from the persisted body — correct.
- mergeable = MERGEABLE.

### 💬 Non-blocking notes

1. `pages/McpMarketListPage.tsx` still renders `<div className="wk-mcp__inner">`, but this PR deleted the `.wk-mcp__inner` CSS rule (which carried the `max-width: 1360px; margin: 0 auto` centering). The div now has no matching style — appears intentional given the full-width layout redesign, but the orphaned className should be dropped to avoid confusion / future dead-class churn.
2. `KvEditor` uses `key={idx}` (array index as React key). Functionally fine here because every input is fully controlled from `entries` state (no stale-value bug), but removing/reordering a middle row can cause minor focus/cursor quirks; keying by a stable per-row id would be cleaner.

### ✅ Highlights

- Careful separation of the transient probe request (real values) from the persisted payload.
- Clear, contract-referenced comments explaining the §5.1/§5.3 ownership model so the "why is the value still stored" question is answered inline.
- New tests explicitly cover the shared-value-passthrough vs user-supplied-placeholder distinction and the vendor single-key-wrapper / `protocol_type` import shapes.

None of the notes block merge.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Review: feat(mcp): per-key user-supplied toggle in MCP create/edit form

**PR:** [#995](https://github.com/Mininglamp-OSS/octo-web/pull/995)
**Head SHA:** `e9f04d935cb0e4ff89d4eb3b0493f1557c663154`

## Summary

Replaces the free-text KV textarea for env/headers with a structured row editor (`KvEntry { key, value, userSupplied }`) plus a per-key toggle marking keys as "user-supplied" (end-users must fill them in). Wire format gains `env_user_supplied` / `headers_user_supplied` string arrays. Auth-type selector and ephemeral `probeBearer` state are removed — Bearer auth is now just an Authorization header row with the toggle ON. JSON import seeds entries with a smart-default toggle via `isSecretKey()`.

## Verification (static analysis only)

- ✅ No stale references to removed `authType`, `probeBearer`, `mergeProbeBearer`, `parseKV`, `serializeKV`, `applySecretSentinel`, `stripSecretSentinel` found via grep across `packages/dmworkmcp/`.
- ✅ `entriesFromWire` strips `SECRET_PLACEHOLDER_SENTINEL` on rehydrate; `entriesToWire` drops empty-key rows — data flow round-trips correctly for the edit case.
- ✅ `applyUserSuppliedPlaceholder` in `quickStartTemplates.ts` correctly consumes `headersUserSupplied` / `envUserSupplied` from `McpQuickStart`; tests confirm both JSON and prompt tabs render `TOKEN_PLACEHOLDER` for user-supplied keys and verbatim values for shared keys.
- ✅ `parseImportJSON` handles all 3 formats (mcpServers / flat / single-key wrapper); `inferTransport` normalizes `streamable_http` → `streamable-http`; values dropped with appropriate warnings.
- ✅ `isSecretKey` smart-default toggle verified against `SECRET_KEY_PATTERN` regex — covers `authorization`, `token`, `key`, `secret`, `password`, `api_key`, `access_key`, etc.
- ✅ `mapDetail` in `mcpService.ts` correctly propagates `envUserSupplied` / `headersUserSupplied` from wire → `McpQuickStart`.
- ⚠️ Build/tests not executed in this run (read-only review environment).

## Findings

### P2 — Env length validation reuses header-specific i18n keys
**File:** `packages/dmworkmcp/src/components/McpCreateModal.tsx:733`

The env-var validation block at ~L733 uses `mcp.create.headerKeyTooLong` / `mcp.create.headerValueTooLong` when env entries exceed the limit. Users will see "Header key too long" when they actually typed an env var name that is too long.

**diff-scope:** new — the KvEditor validation path is entirely new code introduced by this PR.

**Fix:** add `mcp.create.envKeyTooLong` / `mcp.create.envValueTooLong` i18n keys and use them in the env branch.

---

### P2 — `detailToForm` sets env/headers fields that are now dead state
**File:** `packages/dmworkmcp/src/components/McpCreateModal.tsx:173`

`detailToForm` (edit flow) writes `env`, `headers`, `envUserSupplied`, `headersUserSupplied` onto the `CreateMcpParams` form. However, submit reads exclusively from `entriesToWire(envEntries)` / `entriesToWire(headersEntries)`, and `entriesFromWire` on mount already seeds the entry arrays from these same form fields. The form-level env/headers/userSupplied fields are therefore write-only dead state — they exist in the form but nothing reads them at submit time.

**diff-scope:** amplified — the old code also set `env`/`headers` on the form, but the `userSupplied` fields are newly added dead state. Not blocking because the entry-array path works correctly.

**Fix:** remove the `env`, `headers`, `envUserSupplied`, `headersUserSupplied` assignments from `detailToForm` (or alternatively, read from the form at submit time instead of `entriesToWire`). Either approach eliminates the divergence risk.

---

### Nit — `entriesToWire` does not deduplicate repeated keys
**File:** `packages/dmworkmcp/src/components/McpCreateModal.tsx:112`

If a user adds two rows with the same key (e.g. two `Authorization` headers), `entriesToWire` writes both into `userSupplied[]`. The backend may or may not deduplicate. Low risk because the UI add-row flow does not actively prevent duplicates, but a defensive `Set` filter in `entriesToWire` would close the gap cheaply.

**diff-scope:** new — `entriesToWire` is new code.

## Data Flow Tracing

|---|---|---|
| `envEntries` / `headersEntries` (edit) | `entriesFromWire(detail.env, detail.envUserSupplied)` in `detailToForm` → `useEffect` seeds state | ✅ KvEntry rows render correctly |
| `entriesToWire(envEntries)` at submit | KvEditor state → `entriesToWire` → `{ values, userSupplied }` → `toWireParams` → API | ✅ Round-trips correctly |
| `headersUserSupplied` in quickStart | `mapDetail` → `McpQuickStart.headersUserSupplied` → `applyUserSuppliedPlaceholder` | ✅ Substitutes TOKEN_PLACEHOLDER |
| Probe values | Entries at probe time → real values including user-supplied sent to backend | ✅ Probe uses live entry values, not persisted body |
| Import → entries | `parseImportJSON` → `fields.envKeys` / `fields.headerKeys` → `handleApplyImport` seeds entries with `isSecretKey()` toggle | ✅ Smart-default works |

## Blind-spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove paths changed.
- **C2 (control-flow ordering):** N/A — no nested/shared helpers called from multiple sites.
- **C3 (auth boundary):** Clear — auth-type selector removed; Authorization header row with toggle ON is the new Bearer path. No authorization bypass introduced.
- **C4 (auth lifecycle):** N/A — no container/member status cascade.
- **C5 (build/runtime):** ⚠️ Build not run; static analysis only. No heredoc/copy-all/relative-URL risks spotted.
- **C6 (governance/docs):** N/A — no governance doc changes.

## Cross-round Blocker Review

## Verdict

**[Octo-Q] verdict: APPROVE**

No P0/P1 findings. The two P2 items (wrong i18n keys for env validation, dead form state) are cosmetic/maintainability issues that don't break functionality. The Nit (no key dedup) is defensive hardening. All core data flows verified correct.

yujiawei (changes requested):
# Code Review — PR #995 (octo-web)

**Head SHA:** `e9f04d935cb0e4ff89d4eb3b0493f1557c663154`
**Scope:** `packages/dmworkmcp` MCP marketplace create/edit form — replaces the `authType` + probe-bearer + sentinel-masking model with a per-key "user fills locally" toggle. Classified security-sensitive.

This review used two independent assistant passes (line-level + long-context) plus a manual verification pass; the findings below are the reconciled result. Package tests (`pnpm --filter @dmwork/mcp test`) reproduce green locally (4 files / 56 tests).

---

## 1. Spec compliance

**Spec: ✅**

The diff faithfully implements the §5.1 wire contract described in the paired backend PR:
- Adds `env_user_supplied` / `headers_user_supplied` arrays through the whole stack (`types/mcp.ts`, `mcpService.ts` map/wire, form submit).
- Drops `authType` / `probeBearer` / `applySecretSentinel` / snippet `maskSecrets`, matching the backend removal of the write-time `secret_leaked` / `public_secret_disallowed` guards.
- No over-building: the extra JSON "Format" import + list-page polish are additive and self-contained.

No missing / extra / divergent requirements against the stated contract. The concerns below are **correctness and security-model** issues, not scope violations.

## 2. Code Quality

### P1 — Legacy `auth_type=bearer` records lose their auth line in the quick-start snippet (regression on existing data)
`api/mcpService.ts:580-583`, `api/quickStartTemplates.ts:66-77,84-86`

Before this change, a remote MCP created with `authType: "bearer"` did **not** need to persist an `Authorization` header — the snippet *synthesized* `Authorization: Bearer <placeholder>` from `auth_type` (see prior `buildJson`/`buildPrompt`). `mapDetail` now drops `auth_type` and the snippet is generated **only** from `qs.headers`. The paired backend PR states `auth_type` is display-only now and **no data backfill runs**.

Net effect: any already-published bearer MCP whose auth was represented solely by `auth_type=bearer` (no explicit `Authorization` header row) will, once this ships, render a copy-paste config with **no auth header at all** — consumers copy a snippet that won't authenticate. This is a silent functional regression on existing production records.

→ Please either (a) synthesize an `Authorization` user-supplied entry when `auth_type=bearer` and no `Authorization` header is present (migration shim in `mapDetail`), or (b) confirm with data that no such legacy records exist / a backfill runs elsewhere. **Needs human verification** — this cannot be settled from the frontend alone.

### P1 — Manual KV rows default to "shared" with no submit-time guard; the frontend's own secret safety-net was removed
`components/McpCreateModal.tsx:279` (new rows `userSupplied: false`), `:145` / `:801-804` (persist verbatim)

The security model now persists row values **verbatim** and defers all protection to backend §5.3 non-owner read-blanking. Two things compound the risk:
1. A manually-added row defaults to toggle **OFF** (shared). A user typing `Authorization: Bearer <real-token>` who doesn't notice the toggle publishes the real token verbatim. The old flow masked secret-keyed values in the snippet *and* the backend hard-rejected real secrets on write (`secret_leaked`); both safety nets are gone.
2. This is **asymmetric** with JSON import, where `seedEntries` smart-defaults the toggle **ON** for secret-pattern keys (`:1012`). Manual entry — the more common path — has no equivalent guard.

The frontend still has `isSecretKey` available. Recommend a submit-time confirm/warn when a secret-pattern key is left shared with a non-empty value (especially at `public` visibility). **Needs human verification** of the backend §5.3 blanking semantics (see Blind spots) — this is now the *sole* defense line and is unverifiable from this repo.

### P2 — `clampToolDescription` assumes a string; crashes on a description-less probed tool
`components/McpCreateModal.tsx:68`, called at `:665`

`McpTool.description` is typed `string`, but the MCP spec makes tool descriptions optional; a live probe can return `{name:"x"}` with no description. `desc.length` then throws. It is inside `handleProbe`'s try/catch, so it degrades to a "probe failed" toast rather than a white-screen — but it still **blocks probing any server that ships a description-less tool**, with a misleading error. Guard: `function clampToolDescription(desc?: string) { if (!desc) return ""; ... }` (and/or normalize in `probeMcpToolsReal`).

### P2 — `entriesToWire` does not de-duplicate keys
`components/McpCreateModal.tsx:136-148`

Two rows with the same key both flagged user-supplied produce `["Authorization","Authorization"]`. `values[k]` collapses (last wins) but the `userSupplied` array carries the dup. If the backend enforces `uniqueItems` on these arrays the submit 400s with a confusing message. Dedup with a `Set` before returning.

### P2 — JSON import Format C false-positives on non-server objects
`utils/importJson.ts:275-287,299-308`

`isServerLikeValue` accepts any nested object merely containing `type` / `transport` / `protocol_type` (no `url`/`command` required). `{"package":{"type":"module"}}` is parsed as a valid MCP named `package` instead of `unknownFormat`. Tighten to require a connection field (`url` or `command`) alongside a transport hint, or require the transport value to resolve to a known transport.

## 3. Overall verdict

**REQUEST_CHANGES** — spec is compliant, but the two P1 items (legacy bearer-record snippet regression on existing data, and the removal of the frontend secret safety-net with no submit-time guard) are blocking for a security-sensitive, public-facing marketplace surface and need author handling or explicit human confirmation.

## 4. Blind spots — please verify manually
- **Backend §5.3 read-blanking is the sole defense and is unverifiable here.** Confirm it blanks a *shared* (toggle-off) secret-shaped value for non-owner reads, and whether blanking keys off the `*_user_supplied` arrays or off a key-pattern. If it only blanks array members, any toggle-off value under a non-secret-named key leaks to all consumers.
- **No backfill for legacy `auth_type=bearer` records** (backend PR states this explicitly) — see P1 #1.
- **`uniqueItems` enforcement** on `env_user_supplied` / `headers_user_supplied` — determines whether the P2 dedup issue is cosmetic or a hard 400.
- Runtime shape of `McpTool` returned by `/mcps/_probe` (description optionality) — see P2 clamp.

## 5. Additional notes (non-blocking)
- `.gitignore` adds `.codex/` — fine.
- The list-page/detail CSS + search-clear/format-JSON additions are clean and well-commented.
- `assertSafeUploadURL` (pre-existing) still allows `https://` to internal/metadata hosts — out of scope here, noting for the record.

yujiawei:
### Follow-up — two additional maintainability items (non-blocking)

A first-pass automated scan surfaced two more items; I verified both against the head SHA and they are real. Neither changes the overall verdict (still Changes-Requested for the two P1s already noted), but they're worth folding into the same fix pass:

**P2 — env length validation reuses header-specific i18n keys** (`components/McpCreateModal.tsx:732-740`)
The env branch returns `mcp.create.headerKeyTooLong` / `mcp.create.headerValueTooLong`, so an over-long *env var* name shows "header name too long". Add `envKeyTooLong` / `envValueTooLong` keys (both locales) and use them in the env branch.

**P2 — `detailToForm` writes dead env/headers state** (`components/McpCreateModal.tsx:173-176`)
`detailToForm` sets `env` / `headers` / `envUserSupplied` / `headersUserSupplied` on the form, but submit reads exclusively from `entriesToWire(envEntries)` / `entriesToWire(headersEntries)` (the entry arrays are seeded separately in the mount `useEffect`). Those four form fields are write-only after this PR — harmless today, but a divergence trap for the next editor. Drop them from `detailToForm`, or make submit read the form instead of the entry arrays (pick one source of truth).

The key-dedup item raised alongside these matches the `entriesToWire` dedup point already in the main review.

lml2468 (approved):
## Re-review: #995 @ `015dfef3` — 针对 yujiawei CR 的修复复核

上一版我 APPROVE 后,yujiawei 提出 2×🔴 + 1×🟡(review CR)。本轮针对 **两条 🔴 是否被正确修复**复核(PR 已 force-push 为单 commit,prior head 已 GC,故从 merge-base `d82645aa` 全量核实现状;behind 8,0 冲突)。

---

### 🔴#1 —— 旧 `auth_type=bearer` 记录 → 快速启动片段缺认证头 ✅ 已修

`mcpService.ts:568-600` `mapDetail` 新增 **legacy-bearer 迁移 shim**:
- 当 wire 上 `q.auth_type === "bearer"` 且**不存在**任何 `Authorization` 行(大小写不敏感判定)时,合成一个 user-supplied 的 `Authorization: ""` 占位并加入 `headers_user_supplied`。
- 任何已存在的显式 Authorization 行优先(marker 不覆盖真实行)。
- 结果:旧 bearer 记录的复制片段重新渲染出 Authorization 占位行,消费者仍能看到需填入的槽位。✅

### 🔴#2 —— 手动 KV 行默认共享、提交前无密钥告警 ✅ 已修(双层)

1. **输入即智能默认**(`McpCreateModal.tsx:287-301`):键名命中 secret-shape(`isSecretKey`:Authorization / *_token / *_secret / apikey …)且行仍 fresh(未手动改 toggle、无 value)时,自动把 `userSupplied` 置 ON。
2. **提交前显式确认网**(`:800-818`):对 `entriesToWire` 后的 values,收集「有值 + secret-shape 键 + **未** user-supplied」的 `sharedSecretKeys`,非空则 `window.confirm(t("mcp.create.sharedSecretConfirm", {keys}))`,用户取消即 `return` 中止提交。i18n 双语齐备(en/zh `:184`,明确提示「值会以明文提交到市场」)。✅
   - user-supplied 键按设计豁免(占位渲染 + §5.3 非 owner 服务端 blanking 双重保护)。

两条 🔴 均已正确闭合,逻辑与告警文案匹配 yujiawei 的诉求。

---

### 构建 / 测试(实跑)

- `pnpm --filter @octo/web build` ✅ 3.53s(唯一告警为既有 semi-ui 动静态混用,非本 PR)。
- `pnpm --filter @dmwork/mcp test` ✅ **4 files / 56 tests 全绿**。
- i18n parity:mcp **216/216**。

---

### 🟡 非阻断(测试覆盖缺口 —— 对应 yujiawei 的 🟡)

**两处安全相关修复目前均无回归测试**,建议补上以防后续重构静默回退:
1. `mapDetail` 的 legacy-bearer shim:`mcpService.ts` 无对应 test(`mapDetail` 亦未导出);现有 `quickStartTemplates.test.ts` 只覆盖片段渲染(user-supplied 占位),**不覆盖**「`auth_type=bearer` 且无 Authorization 行 → 合成 user-supplied Authorization」这条路径。建议加断言:输入含 `auth_type:"bearer"`、无 headers → 输出 headers 含 `Authorization` 且入 `headersUserSupplied`;已有显式 Authorization 行时不重复合成。
2. 提交前 `sharedSecretConfirm` 网:无 test 覆盖「secret-shape 键 + toggle OFF + 有值 → 弹 confirm;取消则不提交」。这是**防止个人 token 误当共享值明文上架**的最后一道前端闸,尤其值得锁契约。建议加断言:mock `window.confirm` 返回 false → `createMcp` 未被调用;返回 true → 正常提交。

此为**非阻断**(与 yujiawei 该项 🟡 一致):修复本身经代码复核 + 构建验证正确;仅是这两条安全路径缺自动化护栏,易被后续改动悄悄拆掉。

---

### 说明:端到端隐私仍依赖后端契约(承前)
非 owner 看不到 author token 的最终保障来自后端 §5.3 non-owner blanking(octo-marketplace);前端侧无泄漏面(user-supplied 占位 + 本轮新增的 shared-secret confirm)。合并前建议确认后端该行为已部署。

两条阻断性 🔴 已正确修复,构建与 56 项测试通过,i18n 齐备。剩余仅为非阻断的回归测试补强建议。**可合并。**

Jerry-Xin (approved):
**APPROVE** — re-review at head `015dfef3`. Byte-verified fixes for both blocking items raised after the previous head; no regressions on the credential-handling surfaces.

### Honest correction of my prior call

At the previous head `e9f04d9` I APPROVED. **That was an under-call** — a peer's CHANGES_REQUESTED at that head flagged two real problems that my pass missed. I re-traced both at `e9f04d9` independently and confirm they were genuine:

- **(A) Legacy `auth_type=bearer` records lost their auth line.** Before this series, the snippet renderer *synthesized* `Authorization: Bearer <placeholder>` from `auth_type` (base `api/quickStartTemplates.ts` set `merged.Authorization` when `authType === "bearer"`, plus an auth line in the prompt). At `e9f04d9` that synthesis was gone and `mapDetail` dropped `auth_type` with no replacement (both greps empty), so an existing bearer-only record — no explicit `Authorization` row — would render a copy-paste config with **no auth header at all**. Real back-compat regression on existing data.
- **(B) Manual KV rows defaulted to shared with no submit-time guard.** At `e9f04d9`, a manually-added row was `userSupplied: false` and there was no key-based auto-toggle and no submit-time confirm (grep for a confirm/auto-toggle was empty). Only the JSON-import path smart-defaulted the toggle ON for secret-shaped keys. Manual entry — the common path — had no safety net once the old write-time reject was removed. Real leak risk.

Credit to the reviewer who caught these. My earlier note that manual-add-OFF was "safe under the server-side non-owner blanking model" understated the exposure of a shared value that never reaches the blanking path.

### What changed `e9f04d9` → `015dfef3` (5 files, focused)

**(A) FIXED — legacy-bearer migration shim.** `api/mcpService.ts` `mapDetail` now, when the wire carries `auth_type === "bearer"` and no case-insensitive `Authorization` key exists, synthesizes a user-supplied `Authorization: ""` entry (pushed into `headers_user_supplied`). `applyUserSuppliedPlaceholder` then renders that key as the token placeholder, so old bearer records regain an `Authorization` placeholder line in the snippet. The `hasAuthKey` guard ensures any explicit Authorization row wins over the marker. Verified against `api/quickStartTemplates.ts` `applyUserSuppliedPlaceholder`: it substitutes the placeholder for every key in the user-supplied set regardless of value, so the empty synthesized value still surfaces as a placeholder rather than a blank.

**(B) FIXED — two-layer guard.** `components/McpCreateModal.tsx`:
- New `setKey` on the KV editor auto-flips the toggle ON when the operator types a secret-shape key while the row is still fresh (toggle untouched AND value empty). Once the toggle or value is set, it never auto-adjusts — so it won't fight a deliberate choice.
- Submit-time net: before `setSubmitting`, it scans wired env+header values for a secret-shaped key that has a value and is NOT user-supplied, and if any exist it raises a `window.confirm` (new i18n key `sharedSecretConfirm`, present in both en-US and zh-CN) and aborts submit unless the operator explicitly confirms.

**Also fixed (previously P2):**
- `clampToolDescription` now accepts `string | undefined | null` and returns `""` for falsy input — a description-less probed tool no longer throws / degrades to a misleading "probe failed" toast.
- `entriesToWire` dedups the user-supplied array via a `Set`, so two rows with the same toggled key no longer emit a duplicate on the wire.
- `utils/importJson.ts` `isServerLikeValue` is now strict: Format C requires a real connection field (`url` or `command`), so a transport-only object like `{type:"module"}` no longer false-positives as an MCP config.

### Re-verified surfaces (no regression)

- Snippet masking still routes through `applyUserSuppliedPlaceholder`; user-supplied values render as placeholder, shared pass through verbatim (intended for copy-paste config).
- `entriesFromWire` still normalizes the persisted sentinel back to `""`.
- `detailToForm` still reads the `*_user_supplied` arrays; edit flow rebuilds toggle state.
- No card/detail raw-render of env/header values; key names render as React text nodes (no `dangerouslySetInnerHTML`) → no XSS on key names.
- i18n en-US / zh-CN parity intact (new `sharedSecretConfirm` in both).

### Peer non-blocking notes — status

- Env-length validation still reuses `headerKeyTooLong` / `headerValueTooLong` for the env branch (`components/McpCreateModal.tsx:761,766`) → an over-long env var name shows "Header key too long". Cosmetic i18n mislabel, **not addressed**, non-blocking.
- `detailToForm` still writes `env` / `headers` / `envUserSupplied` / `headersUserSupplied` onto the form even though submit reads exclusively from the entry arrays → write-only dead state. **Not addressed**, non-blocking (the entry-array path is correct).
- Duplicate-key dedup: addressed via the `Set` in `entriesToWire` (see above).

### Build / tests / scan

- `@octo/web` build: passes (`✓ built`); residual eval / chunk-size warnings are pre-existing third-party notices (lottie / pdfjs / semi-ui), unrelated to this PR.
- `@dmwork/mcp`: 4 files / **56 tests pass**.
- `secret-scan (gitleaks)` = pass, `osv-scan` = pass; no real credentials in fixtures (placeholders / dummies).
- mergeable = MERGEABLE.

### Cross-reminder (cross-repo dependency, not blocking here)

Non-owner protection for a shared value still depends on the backend §5.3 read-blanking being deployed and keyed correctly (blanking a toggle-off secret-shaped value for non-owner reads). That is out of this frontend PR's control; confirm the backend contract is live before relying on it end-to-end. The frontend side is clean: snippet placeholder-masks user-supplied keys, the new submit confirm guards the shared-secret slip, and there is no card/detail raw-render leak.

Both blocking items are resolved with correct, well-commented shims and the P2s are cleaned up. Approving.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #995 (octo-web)

## Summary

This PR replaces the free-text `KEY: value` textarea editors for MCP headers and env with a structured per-row KV editor that carries a "user fills locally" toggle. The toggle drives the new `headers_user_supplied` / `env_user_supplied` wire arrays, replacing the old sentinel-substitution model (`applySecretSentinel` + backend `secret_leaked` reject) with an explicit per-key opt-in. The old `authType: "bearer" | "none"` selector and dedicated probe-only bearer input are removed; Bearer auth is now expressed as a regular `Authorization` header row with the toggle ON. A legacy-bearer migration shim in `mapDetail` synthesizes the missing entry for pre-toggle records. Secondary changes include a Format C JSON-import path (single-key vendor wrapper), `protocol_type` transport alias, broader secret-key pattern for smart defaults, tool-description clamping on probe ingest, and list-page UI polish (search clear button, result total + source filter relocation, card stat icon).

## Verification

- ✅ **Data-flow round-trip** — `entriesToWire` → wire → `entriesFromWire` preserves key order, values, and userSupplied membership. Sentinel normalization is one-directional (legacy cleanup by design). Empty-key rows are dropped at serialization.
- ✅ **Create/edit parity** — `handleSubmit` builds the same wire pair for both create and update. `buildDetailFromCreate` and `toWireParams` both thread `envUserSupplied` / `headersUserSupplied`. `detailToForm` passes the arrays back to `entriesFromWire`.
- ✅ **Snippet rendering** — `applyUserSuppliedPlaceholder` substitutes `TOKEN_PLACEHOLDER` for userSupplied keys; shared values pass through verbatim. Prompt rendering mirrors the same logic. Tests lock both paths.
- ✅ **Legacy bearer migration** — `mapDetail` synthesizes `Authorization: ""` + userSupplied entry only when `auth_type === "bearer"` AND no existing Authorization key (case-insensitive check). Old records get a placeholder slot; new records are unaffected.
- ✅ **Safety nets layered** — (1) Smart-default toggle ON for secret-pattern keys on import and on key-type; (2) `sharedSecretKeys` confirmation dialog before submit when a secret-looking key has a real value but toggle is OFF; (3) server-side blanking for non-owners (§5.3); (4) client-side placeholder in snippets.

## Findings

### P2 — Env validation reuses header-specific i18n keys (`packages/dmworkmcp/src/components/McpCreateModal.tsx:761`)

The stdio branch of `firstValidationError` validates env entry key/value lengths but emits `mcp.create.headerKeyTooLong` and `mcp.create.headerValueTooLong`. A user with an oversized env variable key sees a toast saying "header" when the field is actually "environment variables". Add dedicated i18n keys (`envKeyTooLong` / `envValueTooLong`) in both locales.

### P2 — KvEditor hardcodes `maxLength` instead of consuming MAXLEN (`packages/dmworkmcp/src/components/McpCreateModal.tsx:327`)

The key input uses `maxLength={128}` and the value input uses `maxLength={1024}`, hardcoded. These match `MAXLEN.headerKey` / `MAXLEN.headerValue` today, but if the constants diverge, the editor cap and the validation cap silently disagree — one allows what the other rejects. Pass `maxKeyLength` / `maxValueLength` props sourced from the MAXLEN object.

## Human-verify

1. **Server-side blanking parity** — The frontend now relies on the backend blanking userSupplied values for non-owner reads (§5.3). Confirm the backend enforces this for both `headers_user_supplied` and `env_user_supplied` arrays, including list/snippet endpoints (not just detail). Not a merge blocker for this frontend PR.
2. **Case-sensitive userSupplied matching** — `applyUserSuppliedPlaceholder` and `entriesFromWire` both use case-sensitive Set lookup against the wire's `user_supplied` array. If the backend normalizes header key casing on storage but preserves the array verbatim, a mismatch could cause a real value to appear in the snippet. Confirm the backend preserves exact key casing for both the map and the array. Not a merge blocker.

- `entriesToWire` dedup: duplicate keys collapse to last-write-wins for the value; userSupplied membership is OR'd via Set. Confusing but never leaks — the result is always at least as restrictive as the least-restricted row.
- `clampToolDescription`: correctly handles `undefined`/`null` (returns ""), preserves short descriptions, and clamps with trailing `…`. Probe toast reports truncation count.
- `parseImportJSON` Format C: `isServerLikeValue` requires `url` or `command` (non-empty string), preventing false positives on arbitrary nested JSON. Transport normalization handles `protocol_type` + underscore variants.
- `SECRET_KEY_PATTERN` broadening is frontend-only (smart-default toggle). The backend `public_secret_disallowed` rule was removed per §5.1 relaxation. A narrower pattern only degrades the default; nothing else fails.
- `handleFormatJson`: best-effort parse/format; failure shows a toast without modifying the buffer.
- McpMarketListPage search: replaced `WKInput` with native `<input type="search">` + clear button. `handleKeyword` signature unchanged (takes a string). Source filter relocated below results — no behavioral change.
- McpCard/McpDetailModal: tool count now uses `IconWrenchStroked` icon + number. Detail modal splits owner/meta line from stats line. No data contract changes.
- `buildDetailFromCreate` and `toWireParams` properly thread `envUserSupplied`/`headersUserSupplied` through the optimistic detail and wire params.

## Data-flow tracing

|---|---|---|
| `headersEntries` / `envEntries` | `entriesFromWire(seed.headers, seed.headersUserSupplied)` on edit; `seedEntries(f.headerKeys!)` on import; manual KvEditor input | Yes — `entriesToWire` collapses to `{values, userSupplied}` pair at submit |
| `headersWire.userSupplied` at submit | `entriesToWire(headersEntries)` → `suppliedSet` → spread to array | Yes — flows to `toWireParams` as `headers_user_supplied` |
| `probeHeaders` / `probeEnv` at probe | Direct iteration of `headersEntries` / `envEntries` → `e.value` | Yes — sent as real values to remote server (by design for probe auth) |
| `sharedSecretKeys` at submit | `entriesToWire` output → iterate `values` map → `isSecretKey(k) && !supplied.has(k)` | Yes — triggers `window.confirm` before submit proceeds |
| `Authorization` snippet placeholder | `applyUserSuppliedPlaceholder(qs.headers, qs.headersUserSupplied)` → `supplied.has(k)` | Yes — TOKEN_PLACEHOLDER substituted for userSupplied keys in JSON + prompt |
| Legacy bearer `Authorization` slot | `mapDetail` shim: `q.auth_type === "bearer"` + no existing auth key → synthesize `{Authorization: ""}` + add to `headersUserSupplied` | Yes — `entriesFromWire` picks up on edit; snippet shows placeholder |

## Blind spot checklist (C1–C6)

**C1 — Dual-path parity**: ✅ Clear. Create↔edit symmetry verified: `entriesToWire`/`entriesFromWire` are inverse functions (modulo sentinel normalization, which is intentional legacy cleanup). Submit path is identical for create and update via `handleSubmit`. Probe path sends real values from the same editors.

**C2 — Control-flow ordering**: ✅ Clear. Validation runs first → `sharedSecretKeys` confirmation second → `setSubmitting(true)` third. No double-submission (button gated by `submitting` flag). No nested/recursive re-entry.

**C3 — Authorization boundary**: ✅ Clear. The security model shift is deliberate: old sentinel+reject → new explicit per-key opt-in. Three-layer safety net (smart toggle, confirmation dialog, server blanking). The `sharedSecretKeys` check catches the case where the user toggles OFF on a secret-looking key with a real value.

**C4 — Authorization lifecycle**: N/A. No auth/jail/credential-lifecycle changes at the container level.

**C5 — Build/note**: Not verifiable in this environment (static analysis only).

**C6 — Governance/docs**: N/A. No governance or policy documents changed.

## Verdict: APPROVED

No P0/P1 issues found. The security model change (sentinel substitution → explicit per-key user-supplied toggle) is well-layered with smart defaults, a confirmation dialog, and server-side blanking. The two P2 items are minor UX/maintenance concerns that don't block merge.

[Octo-Q] verdict: APPROVE — no P0/P1 found; two P2 nits on env validation i18n keys and KvEditor maxLength hardcoding.

yujiawei (approved):
# Code Review — PR #995 (octo-web)

**Head SHA:** `015dfef32f704cddf4d3ec66b22d24f3ff347ab8`
**Scope:** `packages/dmworkmcp` MCP marketplace create/edit form — replaces the `authType` + probe-bearer + write-time sentinel-masking model with a per-key "user fills locally" toggle (`env_user_supplied` / `headers_user_supplied`). Classified security-sensitive.

This is a re-review of a prior `CHANGES_REQUESTED` (head `e9f04d93`). This pass used two independent assistant passes (line-level + long-context) plus a manual verification pass against the current head; the package test suite (`pnpm --filter @dmwork/mcp test`) reproduces green locally (4 files / 56 tests).

## 1. Spec compliance

**Spec: ✅**

The diff faithfully implements the §5.1 wire contract of the paired marketplace backend PR:
- `env_user_supplied` / `headers_user_supplied` string arrays threaded through the whole stack (`types/mcp.ts`, `mcpService.ts` map/wire, form submit).
- `authType` / probe-bearer / `applySecretSentinel` / snippet `maskSecrets` removed, matching the backend removal of the write-time `secret_leaked` / `public_secret_disallowed` guards. No leftover references to the removed symbols remain (only the intentional legacy `auth_type` migration shim reads it).
- Snippet masking now keys off the `*_user_supplied` arrays via `applyUserSuppliedPlaceholder`; shared values pass through verbatim (correct per the relaxed contract).
- i18n parity intact (17 keys added to both en-US and zh-CN).
- Additive, self-contained extras (JSON import "Format C" single-key wrapper + `protocol_type` alias, list-page/detail CSS polish, tool-count stat chip) — no over-build, no scope divergence.

## 2. Code Quality

**Quality: Approved**

All four blocking items from the prior review are addressed at this head:

1. **Legacy `auth_type=bearer` snippet regression** → fixed. `mapDetail` now synthesizes a user-supplied `Authorization` slot when a legacy record carries `auth_type: "bearer"` with no explicit Authorization header, so those snippets keep rendering an auth line. Round-trips cleanly (an edit reload shows a toggle-ON Authorization row, and re-submit self-heals the record into the new model).
2. **Manual KV rows defaulting to shared with no guard** → fixed. `setKey` smart-defaults the toggle ON for secret-shaped keys on a fresh row (empty value + untouched toggle), matching the JSON-import `seedEntries` behavior and closing the manual-entry asymmetry. A submit-time `window.confirm` also warns when a secret-shaped key is left shared with a non-empty value.
3. **`clampToolDescription` crash on undefined** → fixed. Now typed `string | undefined | null` with an early return.
4. **`entriesToWire` duplicate keys / Format C false-positives** → fixed. `Set`-based dedup on the user-supplied array; `isServerLikeValue` now requires a real connection field (`url` or `command`), so transport-only blobs no longer parse as MCP configs.

### Non-blocking findings (P2)

- **P2 — Legacy bearer record that already has an explicit `Authorization` key is not marked user-supplied.** `api/mcpService.ts:589-601`. The shim only synthesizes/marks when no Authorization key exists; a legacy record with an explicit Authorization row renders that stored value verbatim in the snippet instead of a `<YOUR_TOKEN>` placeholder. Note this is **not** a token leak: under the old model the backend rejected real values under secret-shaped keys (`secret_leaked`) and the sentinel replaced blanks, so such a key could only hold the sentinel literal or empty — the worst case is a cosmetic `Authorization: __OCTO_SECRET_PLACEHOLDER__` line, which self-heals on the next owner edit. Optional hardening: locate the existing case-insensitive Authorization key and add it to `headersUserSupplied` (and normalize a persisted sentinel to placeholder in `applyUserSuppliedPlaceholder`).
- **P2 — False "descriptions truncated" toast for description-less probed tools.** `components/McpCreateModal.tsx:690-706`. `clampToolDescription(undefined)` returns `""`, and `"" !== undefined` increments the `truncated` counter, so probing a server with a description-less tool shows a misleading truncation notice. Strictly better than the prior crash. Guard: only count when the original is a string longer than `MAXLEN.text`.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. All prior blocking items are resolved; the two remaining items are cosmetic P2s, neither of which blocks merge.

## 4. Blind spots — please verify manually (security-sensitive)

- **Backend non-owner read-blanking is the sole defense and is unverifiable from this repo.** With write-time guards removed, all header/env values persist verbatim and the only protection for consumer-facing reads is the backend blanking of non-owner values. Confirm it blanks *shared* (toggle-off) secret-shaped values as well as user-supplied ones, and that the List/detail projections never carry a non-empty value to a non-owner. (This was verified complete in the paired backend PR review; flagging here because it cannot be re-checked from the frontend.)
- **No backfill for legacy `auth_type=bearer` records** — the migration shim is client-side only; old records keep their wire shape until edited.

## 5. Additional notes (non-blocking)
- `.gitignore` adds `.codex/` — fine.
- Tests comprehensively lock the new behavior (verbatim shared values, placeholder for user-supplied, Format C mixreach shape, dedup) — good coverage.

PR #987, opened 2026-07-22
feat(docs): support SVG image insertion

wwwzzzcccc:
## Summary
- support SVG selection and insertion in Docs toolbar, slash command, paste, and drop flows
- route SVG files through the backend sanitizing endpoint in document and Excalidraw board uploads
- enable SVG in Univer sheet image picker/runtime validation
- prevent raw SVG bytes from being mislabeled as PNG during DOCX export

## Linked Spec
- Mininglamp-OSS/octo-docs#34

## How verified
- `pnpm --filter @octo/docs exec vitest run src/attachments/api.test.ts src/editor/imageUpload.test.ts src/export/docx/svg.test.ts src/sheet/sheetImageTypes.test.ts` (19/19)
- `git diff --check`
- full typecheck reaches two unrelated baseline errors in `dmworkbase/src/i18n/format.ts` and `docs/src/members/sort.ts`; no modified-file errors

Fixes Mininglamp-OSS/octo-docs#34

PR #1003, opened 2026-07-22
feat(summary): modernize citation badge UX (reading-order + tokens + hover pin)

wanyiwang06:
## Summary

Modernize the citation badge UX in agent/workflow summary detail. Bundles the 3 issues raised by the maintainer plus 2 industry-standard interactions (hover preview + click-to-pin), aligning with Perplexity / Kimi / Notion / Teams conventions.

## What changed

**P1 - Reading-order renumbering**
Badges now display 1-indexed reading-order numbers (first citation shown is `[1]`) instead of raw pool positions like `[37]`. Underlying citation data is unchanged; only the label the user sees differs. Repeated references to the same source reuse the same display number.

**P2 - Token-based styling + hover / focus states**
Replaced inline `React.CSSProperties` (hard-coded `#1677ff` / `rgba(22,119,255,0.13)`) with a new `CitationBadge.css` using `--wk-*` semantic tokens per `DEVELOPMENT.md` chapter 2. Badge now inherits brand color (near-black on this project), radius (`--wk-r-xs`), spacing (`--wk-sp-*`), and typography scale (`--wk-text-size-xs`). Light and dark themes track automatically. No hard-coded colors, no `!important`, no `@media (prefers-color-scheme: dark)`.

**P3 - Group badge label rule**
- `len <= 3` -> comma joined: `[37,38,39]`
- `len > 3`  -> range:        `[30-35]`

Pure formatting logic + reading-order map extracted to `citationFormat.ts` so unit tests don't have to drag in tiptap / semi-ui.

**Hover + pin interaction**
- Hover a badge for 400ms -> compact preview card (auto-dismiss on mouseleave, 200ms grace)
- Click -> pin the popover (full context + jump link), survives mouseleave
- Second click / outside click / Escape -> close
- Group badge hover shows up to 3 preview messages; pinned view keeps the full merged timeline

**Accessibility**
- Badges are now `tabIndex={0}`, keyboard-navigable
- Enter / Space pins; Escape closes when pinned

## Files

- `CitationBadge.tsx` +285/-109 — hover/pin state via new `useHoverPin` hook; token-driven className; propagates optional `displayIndex` prop
- `CitationText.tsx` +37 — pre-scans source for reading-order map; passes `displayindex` / `displayindices` through the remark AST
- `CitationBadge.css` (new) — semantic tokens only, light+dark compatible
- `citationFormat.ts` (new) — pure `formatGroupLabel` + `buildDisplayIndexMap`
- `__tests__/CitationBadge.test.ts` (new) — 12 unit tests

## Verification

- `pnpm vitest run` (dmworksummary) -> **488 passed / 1 pre-existing failure** (`SummaryVersionHistory.test.tsx:218` — reproduced on untouched `upstream/main`, unrelated to this change)
- `pnpm turbo run build --filter=@octo/web` -> **build succeeds** (esbuild, 4.13s, only pre-existing warnings)
- 12 new tests cover: comma/range branch table, `RANGE_THRESHOLD` guard, reading-order renumbering, dedup on repeat, markdown link exclusion (`[n](url)`), team citation exclusion (`[Pn]`)

## Non-goals (intentional; not in this PR)

- Backend `CitationIndex` assignment: unchanged
- No "AI generated" banner
- No bottom "Sources" list
- No by-speaker aggregation / section tabs / two-way highlighting

## Compatibility

- Traditional workflow summaries also benefit (`CitationText` / `CitationBadge` are shared)
- Existing `hidePlainCitations` / `disableTeamMemberPreview` flags unchanged
- Backend citation data structure unchanged; API responses untouched

### DIFF
--- a/packages/dmworksummary/src/components/CitationBadge.css
+++ b/packages/dmworksummary/src/components/CitationBadge.css
@@ -1,0 +1,126 @@
+/* Citation badge (inline [n] token in summary body).
+ *
+ * Wraps a <sup> so it participates in surrounding text flow but sits slightly
+ * raised. Uses brand-tinted background so it visually anchors to the site
+ * primary color (currently near-black) rather than the previous hard-coded
+ * Ant-blue. All colors / spacing / radius come from --wk-* semantic tokens
+ * so light/dark themes track automatically (per DEVELOPMENT.md §2).
+ */
+.citation-badge {
+    display: inline-flex;
+    align-items: center;
+    background: var(--wk-brand-tint-08);
+    color: var(--wk-text-strong);
+    border-radius: var(--wk-r-xs);
+    padding: 0 var(--wk-sp-1);
+    margin-left: var(--wk-sp-0-5);
+    font-size: var(--wk-text-size-xs);
+    line-height: 1.4;
+    vertical-align: super;
+    cursor: pointer;
+    transition: background 0.15s ease, color 0.15s ease;
+    user-select: none;
+}
+
+.citation-badge:hover,
+.citation-badge:focus-visible {
+    background: var(--wk-brand-tint-15);
+    color: var(--wk-brand-primary);
+    outline: none;
+}
+
+/* Popover content wrappers (shared by CitationBadge, CitationGroupBadge,
+ * TeamCitationBadge, and their internal message rows). Extracted from the
+ * old inline React.CSSProperties objects so tokens work + dark theme.
+ */
+.citation-popover {
+    max-width: 340px;
+    max-height: 400px;
+    padding: var(--wk-sp-2) var(--wk-sp-1);
+    overflow-y: auto;
+}
+
+.citation-popover--wide {
+    max-width: 360px;
+}
+
+.citation-mini-preview {
+    max-width: 320px;
+    padding: var(--wk-sp-2);
+}
+
+.citation-cited-msg {
+    padding: var(--wk-sp-1) var(--wk-sp-2);
+    border-left: 3px solid var(--wk-brand-primary);
+    font-size: var(--wk-text-size-base);
+    line-height: 1.5;
+    word-break: break-word;
+}
+
+.citation-context-msg {
+    padding: var(--wk-sp-1) var(--wk-sp-2);
+    font-size: var(--wk-text-size-sm);
+    color: var(--wk-text-muted, rgba(28, 28, 35, 0.55));
+    font-style: italic;
+    line-height: 1.5;
+    word-break: break-word;
+}
+
+.citation-msg-header {
+    display: flex;
+    justify-content: space-between;
+    margin-bottom: var(--wk-sp-0-5);
+}
+
+.citation-msg-sender {
+    font-weight: 600;
+    font-size: var(--wk-text-size-base);
+}
+
+.citation-msg-sender--context {
+    font-weight: 500;
+    font-size: var(--wk-text-size-sm);
+}
+
+.citation-msg-time {
+    font-size: var(--wk-text-size-sm);
+    color: var(--wk-text-muted, rgba(28, 28, 35, 0.55));
+}
+
+.citation-msg-time--context {
+    font-size: var(--wk-text-size-xs);
+    color: var(--wk-text-muted, rgba(28, 28, 35, 0.45));
+}
+
+.citation-msg-source {
+    font-size: var(--wk-text-size-sm);
+    color: var(--wk-text-muted, rgba(28, 28, 35, 0.55));
+    margin-bottom: var(--wk-sp-1);
+}
+
+.citation-msg-body {
+    font-size: var(--wk-text-size-base);
+    line-height: 1.5;
+}
+
+.citation-msg-body--context {
+    font-size: var(--wk-text-size-sm);
+    line-height: 1.5;
+}
+
+.citation-jumplink {
+    margin-top: var(--wk-sp-2);
+    padding-top: var(--wk-sp-1-5);
+    border-top: 1px solid var(--wk-brand-tint-08);
+    text-align: right;
+}
+
+.citation-jumplink-btn {
+    color: var(--wk-brand-primary);
+    font-size: var(--wk-text-size-sm);
+    cursor: pointer;
+}
+
+.citation-jumplink-btn:hover {
+    text-decoration: underline;
+}
--- a/packages/dmworksummary/src/components/CitationBadge.css
+++ b/packages/dmworksummary/src/components/CitationBadge.css
@@ -60,7 +60,7 @@
 .citation-context-msg {
     padding: var(--wk-sp-1) var(--wk-sp-2);
     font-size: var(--wk-text-size-sm);
-    color: var(--wk-text-muted, rgba(28, 28, 35, 0.55));
+    color: var(--wk-text-tertiary, rgba(28, 28, 35, 0.55));
     font-style: italic;
     line-height: 1.5;
     word-break: break-word;
@@ -84,17 +84,17 @@
 
 .citation-msg-time {
     font-size: var(--wk-text-size-sm);
-    color: var(--wk-text-muted, rgba(28, 28, 35, 0.55));
+    color: var(--wk-text-tertiary, rgba(28, 28, 35, 0.55));
 }
 
 .citation-msg-time--context {
     font-size: var(--wk-text-size-xs);
-    color: var(--wk-text-muted, rgba(28, 28, 35, 0.45));
+    color: var(--wk-text-tertiary, rgba(28, 28, 35, 0.45));
 }
 
 .citation-msg-source {
     font-size: var(--wk-text-size-sm);
-    color: var(--wk-text-muted, rgba(28, 28, 35, 0.55));
+    color: var(--wk-text-tertiary, rgba(28, 28, 35, 0.55));
     margin-bottom: var(--wk-sp-1);
 }
 
--- a/packages/dmworksummary/src/components/CitationBadge.tsx
+++ b/packages/dmworksummary/src/components/CitationBadge.tsx
@@ -1,4 +1,4 @@
-import React, { useContext, useMemo } from 'react';
+import React, { useContext, useMemo, useRef, useState, useCallback, useEffect } from 'react';
 import { Popover } from '@douyinfe/semi-ui';
 import { i18n, useI18n } from '@octo/base';
 import { Channel, ChannelTypeGroup, ChannelTypePerson } from 'wukongimjssdk';
@@ -7,15 +7,33 @@ import { ShowConversationOptions } from '@octo/base/src/EndpointCommon';
 import { ChannelTypeCommunityTopic } from '@octo/base/src/Service/Const';
 import CitationText, { CitationContext } from './CitationText';
 import { CitationItem, CitationContextMessage, TeamCitationItem, MemberStatus } from '../types/summary';
+import { formatGroupLabel } from './citationFormat';
+import './CitationBadge.css';
+
+/** Hover-preview delay (ms) tuned to match Perplexity/Kimi: long enough that
+ * casual mouse traversal does not fire, short enough to feel instant when the
+ * reader stops to look. */
+const HOVER_OPEN_DELAY_MS = 400;
+const HOVER_CLOSE_DELAY_MS = 200;
 
 interface CitationBadgeProps {
     index: number;
+    /**
+     * Display index rendered inside the [n] label. When omitted, falls back
+     * to `index`. Callers pass a re-numbered value (starting at 1 in reading
+     * order) so users don't see raw pool positions like [37]. The internal
+     * `index` still keys into the citations array — see remarkCitation in
+     * CitationText for the mapping construction.
+     */
+    displayIndex?: number;
     citations: CitationItem[];
     badgeKey: string;
 }
 
 interface CitationGroupBadgeProps {
     indices: number[];
+    /** Display indices, one per original index (same length as `indices`). */
+    displayIndices?: number[];
     citations: CitationItem[];
     badgeKey: string;
 }
@@ -33,37 +51,6 @@ function resolveChannelType(channelType?: number) {
     if (channelType === 5) return ChannelTypeCommunityTopic;
     return ChannelTypeGroup;
 }
-
-const badgeStyle: React.CSSProperties = {
-    display: 'inline-flex',
-    alignItems: 'center',
-    background: 'rgba(22, 119, 255, 0.13)',
-    color: '#1677ff',
-    borderRadius: 4,
-    padding: '0 4px',
-    fontSize: 11,
-    cursor: 'pointer',
-    marginLeft: 2,
-    lineHeight: '16px',
-    verticalAlign: 'super',
-};
-
-const contextMsgStyle: React.CSSProperties = {
-    padding: '4px 8px',
-    fontSize: 12,
-    color: '#999',
-    fontStyle: 'italic',
-    lineHeight: 1.5,
-    wordBreak: 'break-word',
-};
-
-const citedMsgStyle: React.CSSProperties = {
-    padding: '4px 8px',
-    borderLeft: '3px solid #1677ff',
-    fontSize: 13,
-    lineHeight: 1.5,
-    wordBreak: 'break-word',
-};
 
 interface MergedMessage {
     sender: string;
@@ -121,12 +108,12 @@ function ContextMessages({ messages }: { messages?: CitationContextMessage[] }) 
     return (
         <>
             {messages.map((msg, i) => (
-                <div key={i} style={contextMsgStyle}>
-                    <div style={{ display: 'flex', justifyContent: 'space-between', marginBottom: 2 }}>
-                        <span style={{ fontSize: 12, fontWeight: 500 }}>{msg.sender}</span>
-                        <span style={{ fontSize: 11, color: '#bbb' }}>{formatTime(msg.sent_at)}</span>
+                <div key={i} className="citation-context-msg">
+                    <div className="citation-msg-header">
+                        <span className="citation-msg-sender--context">{msg.sender}</span>
+                        <span className="citation-msg-time--context">{formatTime(msg.sent_at)}</span>
                     </div>
-                    <div>{msg.content}</div>
+                    <div className="citation-msg-body--context">{msg.content}</div>
                 </div>
             ))}
         </>
@@ -137,9 +124,9 @@ function JumpLink({ citation, badgeKey, closeKey }: { citation: CitationItem; ba
     const { t } = useI18n();
     if (!citation.channel_id || !citation.message_seq || citation.channel_type == null) return null;
     return (
-        <div style={{ marginTop: 8, paddingTop: 6, borderTop: '1px solid #f0f0f0', textAlign: 'right' }}>
+        <div className="citation-jumplink">
             <span
-                style={{ color: '#1677ff', fontSize: 12, cursor: 'pointer' }}
+                className="citation-jumplink-btn"
                 onClick={(e) => {
                     e.stopPropagation();
                     closeKey(badgeKey);
@@ -161,55 +148,129 @@ function JumpLink({ citation, badgeKey, closeKey }: { citation: CitationItem; ba
     );
 }
 
-const CitationBadge: React.FC<CitationBadgeProps> = ({ index, citations, badgeKey }) => {
+/**
+ * Custom hook: delayed hover with an "always visible when pinned" override.
+ *
+ * Click pins the popover (survives mouseleave), click again unpins. Hover
+ * shows a temporary preview that fades on mouseleave. Combines both modes into
+ * a single `visible` boolean the caller passes to Semi Popover in `custom`
+ * trigger mode.
+ */
+function useHoverPin(pinned: boolean) {
+    const [hovering, setHovering] = useState(false);
+    const openTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+    const closeTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+    const clearTimers = useCallback(() => {
+        if (openTimer.current) { clearTimeout(openTimer.current); openTimer.current = null; }
+        if (closeTimer.current) { clearTimeout(closeTimer.current); closeTimer.current = null; }
+    }, []);
+
+    const onMouseEnter = useCallback(() => {
+        if (closeTimer.current) { clearTimeout(closeTimer.current); closeTimer.current = null; }
+        if (hovering || pinned) return;
+        openTimer.current = setTimeout(() => setHovering(true), HOVER_OPEN_DELAY_MS);
+    }, [hovering, pinned]);
+
+    const onMouseLeave = useCallback(() => {
+        if (openTimer.current) { clearTimeout(openTimer.current); openTimer.current = null; }
+        if (pinned) return;
+        closeTimer.current = setTimeout(() => setHovering(false), HOVER_CLOSE_DELAY_MS);
+    }, [pinned]);
+
+    useEffect(() => () => clearTimers(), [clearTimers]);
+
+    // When pinned externally, drop the hover state so it doesn't fight the pin.
+    useEffect(() => {
+        if (pinned && hovering) setHovering(false);
+    }, [pinned, hovering]);
+
+    const visible = pinned || hovering;
+    return { visible, onMouseEnter, onMouseLeave };
+}
+
+const CitationBadge: React.FC<CitationBadgeProps> = ({ index, displayIndex, citations, badgeKey }) => {
     const { t } = useI18n();
     const { activeKey, onBadgeClick, closeKey } = useContext(CitationContext);
     const citation = citations.find(c => c.index === index);
+    const shownIndex = displayIndex ?? index;
+
+    const pinned = activeKey === badgeKey;
+    const { visible, onMouseEnter, onMouseLeave } = useHoverPin(pinned);
 
     if (!citation) {
-        return <sup style={badgeStyle}>[{index}]</sup>;
+        return <sup className="citation-badge">[{shownIndex}]</sup>;
     }
 
-    const isVisible = activeKey === badgeKey;
+    // Hover preview: compact 3-line card, no jump button. Pinned view: full
+    // context (context_before + main + context_after + jump link).
+    const previewContent = !pinned ? (
+        <div className="citation-mini-preview">
+            <div className="citation-cited-msg">
+                <div className="citation-msg-header">
+                    <span className="citation-msg-sender">{citation.sender}</span>
+                    <span className="citation-msg-time">{formatTime(citation.sent_at)}</span>
+                </div>
+                {citation.source && (
+                    <div className="citation-msg-source">
+                        {t("summary.citation.source", { values: { source: citation.source } })}
+                    </div>
+                )}
+                <div className="citation-msg-body">{citation.content}</div>
+            </div>
+        </div>
+    ) : (
+        <div className="citation-popover">
+            <ContextMessages messages={citation.context_before} />
+            <div className="citation-cited-msg">
+                <div className="citation-msg-header">
+                    <span className="citation-msg-sender">{citation.sender}</span>
+                    <span className="citation-msg-time">{formatTime(citation.sent_at)}</span>
+                </div>
+                {citation.source && (
+                    <div className="citation-msg-source">
+                        {t("summary.citation.source", { values: { source: citation.source } })}
+                    </div>
+                )}
+                <div className="citation-msg-body">{citation.content}</div>
+            </div>
+            <ContextMessages messages={citation.context_after} />
+            <JumpLink citation={citation} badgeKey={badgeKey} closeKey={closeKey} />
+        </div>
+    );
 
     return (
         <Popover
             trigger="custom"
-            visible={isVisible}
+            visible={visible}
             position="top"
             showArrow
             onClickOutSide={() => closeKey(badgeKey)}
-            content={
-                <div style={{ maxWidth: 340, padding: '8px 4px', maxHeight: 400, overflowY: 'auto' }}>
-                    <ContextMessages messages={citation.context_before} />
-                    <div style={citedMsgStyle}>
-                        <div style={{ display: 'flex', justifyContent: 'space-between', marginBottom: 2 }}>
-                            <span style={{ fontWeight: 600, fontSize: 13 }}>{citation.sender}</span>
-                            <span style={{ fontSize: 12, color: '#999' }}>{formatTime(citation.sent_at)}</span>
-                        </div>
-                        {citation.source && (
-                            <div style={{ fontSize: 12, color: '#999', marginBottom: 4 }}>
-                                {t("summary.citation.source", { values: { source: citation.source } })}
-                            </div>
-                        )}
-                        <div style={{ fontSize: 13, lineHeight: 1.5 }}>{citation.content}</div>
-                    </div>
-                    <ContextMessages messages={citation.context_after} />
-                    <JumpLink citation={citation} badgeKey={badgeKey} closeKey={closeKey} />
-                </div>
-            }
+            content={previewContent}
         >
-            <sup className="citation-badge" style={badgeStyle} onClick={() => onBadgeClick(badgeKey)}>[{index}]</sup>
+            <sup
+                className="citation-badge"
+                tabIndex={0}
+                onMouseEnter={onMouseEnter}
+                onMouseLeave={onMouseLeave}
+                onClick={() => onBadgeClick(badgeKey)}
+                onKeyDown={(e) => {
+                    if (e.key === 'Enter' || e.key === ' ') { e.preventDefault(); onBadgeClick(badgeKey); }
+                    if (e.key === 'Escape' && pinned) { e.preventDefault(); closeKey(badgeKey); }
+                }}
+            >[{shownIndex}]</sup>
         </Popover>
     );
 };
 
-export const CitationGroupBadge: React.FC<CitationGroupBadgeProps> = ({ indices, citations, badgeKey }) => {
+export const CitationGroupBadge: React.FC<CitationGroupBadgeProps> = ({ indices, displayIndices, citations, badgeKey }) => {
     const { activeKey, onBadgeClick, closeKey } = useContext(CitationContext);
 
-    const first = indices[0];
-    const last = indices[indices.length - 1];
-    const label = `${first}-${last}`;
+    // Label uses the display indices (post-renumbering) so what the user sees
+    // matches the [n] tokens in the body. Original `indices` still key the
+    // citation lookup below.
+    const shownList = displayIndices && displayIndices.length === indices.length ? displayIndices : indices;
+    const label = formatGroupLabel(shownList);
 
     const indicesKey = indices.join(',');
     const groupCitations = useMemo(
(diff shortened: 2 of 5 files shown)


lml2468 (dismissed):
## Review: #1003 — feat(summary): 引用角标 UX 现代化(阅读序号 + tokens + hover pin)

**意图**:重做 agent/workflow 总结详情里的引用角标 —— (P1) 显示 1-indexed 阅读序号而非原始池位 `[37]`;(P2) 用 `--wk-*` 语义 token 替换硬编码内联样式;(P3) 组角标 `≤3` 逗号、`>3` 范围;外加 hover 预览 + click-to-pin 交互。**底层引用数据不变,仅改展示 label。**

锚定 head `9f8aa1c1` / merge-base `bcbed66e`(behind 35,`mergeable=MERGEABLE`,0 冲突)。

---

### Gate 1 — 规格符合度 ✅

- **P1 阅读序号**:`buildDisplayIndexMap` 预扫原始 markdown,首现顺序赋 1,2,3…,重复 raw index 复用同号;`[n](url)` 链接与 `[Pn]` team 引用正确排除(regex `/\[(\d+)\](?!\()/g` + `\d+` 不触字母)。渲染层 `shownIndex = displayIndex ?? index`,数据 lookup 仍用 raw `index` → **数据不变,仅 label 变**。✅
- **P2 token 化**:`CitationBadge.css` 用 33 处 `var(--wk-*)`,无 `!important`、无 `@media(prefers-color-scheme)`,明暗随主题。✅
- **P3 组 label**:`formatGroupLabel` —— `≤3` join、`>3` first-last,`RANGE_THRESHOLD=3` 有守卫测试。✅

无缺失、无过度、无静默偏离。Non-goals(后端 CitationIndex、Sources 列表等)明确列出,scope 收敛。

### Gate 2 — 代码质量 ✅

- **交互/内存安全(重点)**:`useHoverPin` 双 timer 在每次转换 null 化 + `useEffect(()=>()=>clearTimers())` 卸载清理,无 setState-after-unmount;outside-click 走 semi-ui `Popover onClickOutSide`(无手动 document 监听 → 无泄漏);pin 状态集中在 `CitationText` provider 单一 `activeKey`(同一时刻仅一个 pin);`pinned && hovering` 时复位 hover,不与 pin 打架。✅
- **XSS**:全链路 **0 `dangerouslySetInnerHTML`**,引用内容(sender/content/source/time)均 React 文本子节点自动转义。✅
- **a11y**:角标 `tabIndex=0`,Enter/Space pin、Escape close(pinned 时)。✅
- **正确性:`>3` 范围对非连续序号的语义**:`[2,5,9,14]` 渲成 `[2-14]`(**已由测试与文档显式声明为有意的产品取舍**,`data lookup 用 raw indices` 不受影响)。仅展示近似,非数据完整性问题。✅

**构建 / 测试(实跑):**
- `pnpm --filter @octo/web build` ✅ built in 3.17s(唯一告警为既有 semi-ui 动静态混用,非本 PR)。
- `CitationBadge.test.ts` ✅ **12/12 全绿**(comma/range 分支表、`RANGE_THRESHOLD` 守卫、阅读序号、重复去重、`[n](url)` 排除、`[Pn]` 排除、非连续 range)。
- **已核实 PR 声称的 `SummaryVersionHistory.test.tsx:218` 为既有失败**:该文件本 PR 未触碰;在 merge-base `bcbed66e` 上复跑同样 `1 failed | 10 passed` —— 确认 pre-existing,非本 PR 引入。✅
- i18n:`summary.citation.*` en↔zh parity。

---

### 🟡 非阻断(nit)

1. **CSS「无硬编码颜色」措辞**:`CitationBadge.css` 有 4 处 `rgba(28,28,35,.x)`,但**均为 `var(--wk-text-muted, <fallback>)` 的 fallback 值** —— 属合规兜底写法,并非裸硬编码。仅提示 PR 描述「No hard-coded colors」严格说应表述为「无裸硬编码,仅 token fallback」。不影响主题切换。
2. **Escape 关闭的焦点前提**:Escape 绑在角标自身 `onKeyDown`,仅当角标聚焦时生效;若用户点 pin 后焦点移入 popover 内容,Escape 不关。属轻微 a11y 细节,`onClickOutSide` 已兜底,非阻断。

底层数据不变、token 合规、hover/pin 定时器与监听清理干净、无 XSS、纯逻辑有 12 项测试覆盖(含非连续 range 与排除规则),构建通过,pre-existing 失败已核实。**可合并。**

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1003 (octo-web)
Head SHA: `9f8aa1c1364474f1d7f948418399753479df6fc3`
Scope: `feat(summary): modernize citation badge UX (reading-order + tokens + hover pin)` — 5 files, +484 / -109, all under `packages/dmworksummary/src/components/`.

## Summary

Pure front-end UX modernization for the citation popover layer used by the work-summary feature: inline `React.CSSProperties` are replaced by semantic `--wk-*` token classes in a new `CitationBadge.css`; `CitationBadge` and `CitationGroupBadge` gain a delayed-hover preview plus click-to-pin interaction via a local `useHoverPin` hook; reading-order display renumbering (`[37]` → `[1]`) is implemented in a new pure helper `buildDisplayIndexMap` in `citationFormat.ts` and wired through the `remarkCitation` plugin via new `displayindex` / `displayindices` attributes. `TeamCitationBadge` receives the CSS class swap only — no hover / keyboard upgrade. No backend, schema, auth, or i18n surface is touched beyond one reused i18n key (`summary.confirmPage.declined`).

## Verification

Static analysis only at head `9f8aa1c1`; build and tests not executed in this environment.

- ✅ **Token correctness** — every `--wk-*` token used in `CitationBadge.css` (`--wk-brand-tint-08/15`, `--wk-text-strong`, `--wk-text-muted`, `--wk-sp-{0-5,1,1-5,2}`, `--wk-r-xs`, `--wk-brand-primary`, `--wk-text-size-{xs,sm,base}`) is attested across `packages/dmpersonal/src/personal.css`, `packages/dmworksummary/src/index.css`, `packages/dmworktodo/src/ui/*/index.css`, and documented in `DEVELOPMENT.md:138-145`. Fallback values for `--wk-text-muted` (`rgba(28, 28, 35, 0.55/0.45)`) match other modules' conventions.
- ✅ **Regex parity between `buildDisplayIndexMap` and `remarkCitation`** — both use `/\[(\d+)\](?!\()/g` byte-for-byte, so pre-scan and render never diverge on what counts as a `[n]` citation vs a `[n](url)` markdown link (`packages/dmworksummary/src/components/citationFormat.ts:46`, `:92`).
- ✅ **`hidePlainCitations` privacy guard preserved** — still honored on the `[n]` path via `hasCitations` gating at `CitationText.tsx:274`; the `[Pn]` namespace remains independent.
- ✅ **`TeamCitationBadge` declined-member guard intact** — the `OCT-15 / #495` defense-in-depth branch for `member?.status === "declined"` is preserved with its i18n key reuse (`CitationBadge.tsx:377-383`).
- ✅ **Rehype sanitize schema extended consistently** — `citationSchema` at `CitationText.tsx:55-63` declares the new `displayindex` / `displayindices` attributes; without this, `rehype-sanitize` would strip them and the renumbering would silently fall back to raw indices.
- ✅ **Timer cleanup in `useHoverPin`** — unmount `useEffect` calls `clearTimers()` and `clearTimers` is memoized with empty deps, so no stale `setTimeout` handles can outlive the component (`CitationBadge.tsx:166-180`).
- ✅ **`mergeGroupMessages` behavior unchanged** — pre-existing helper at `:63-101` (not part of the diff) is only *consumed* by the new pinned view; no dedup/sort contract change.

## Findings

No P0 / P1 issues. Three P2 items below, all non-blocking.

### P2 — Interaction parity: `TeamCitationBadge` misses the hover + keyboard upgrade (`packages/dmworksummary/src/components/CitationBadge.tsx:325-404`)

`CitationBadge` and `CitationGroupBadge` both moved to `useHoverPin` + `tabIndex={0}` + `Enter`/`Space`/`Escape` keyboard handling and `trigger="custom"`. `TeamCitationBadge` receives only the `.citation-badge` class swap at `:325` and stays on `visible={activeKey === badgeKey}` with an `onClick`-only `<sup>`. Readers of a team summary therefore get three different interaction models on visually-identical `[n]` vs `[Pn]` badges — hover-preview on regular citations, click-only on team citations, and no keyboard activation path for `[Pn]`. Not a regression (the old `[Pn]` behavior is preserved), but it is a UX gap inside the scope this PR claims ("modernize citation badge UX" — plural, no qualifier). Worth either (a) lifting `useHoverPin` into `TeamCitationBadge` with the same `previewContent` / pinned split, or (b) explicitly scoping the PR title/body to personal citations and adding a follow-up.

### P2 — Group-label range can render backwards when display indices aren't monotonic (`packages/dmworksummary/src/components/citationFormat.ts:21` + `CitationBadge.tsx:257-258`)

`formatGroupLabel` emits `${first}-${last}` for runs longer than `RANGE_THRESHOLD`. `remarkCitation` at `CitationText.tsx:140-158` groups adjacent same-channel citations in document order and passes `group.indices.map(resolveDisplay)` straight through as `displayindices`. Because `buildDisplayIndexMap` numbers by first-seen order, a same-channel run whose raw pool indices decrease in reading order produces `displayIndices = [1, 2]` (fine), but the symmetric case `…[37][12]…` in the same channel produces `displayIndices = [1, 2]` as well — and if the run exceeds threshold the range is `1-N`. The fragile case is when the run's **raw** indices are used as the fallback because `displayIndices.length !== indices.length` (defensive branch at `CitationBadge.tsx:257`): if the raw run is non-monotonic, `${first}-${last}` prints as e.g. `[37-12]`. Product spec says "first-last" but users will read that as a decreasing range, which looks broken. Fix: either sort ascending before emitting the range, or take `min/max` of the run — `const lo = Math.min(...indices); const hi = Math.max(...indices);` — so the label is always visually increasing regardless of input order. Low reach in practice (same-channel grouping strongly correlates with sequential pool indices), but reachable via a legitimate re-ordering in the summary body.

### P2 — Keyboard-only users can't dismiss a hover-triggered popover (`packages/dmworksummary/src/components/CitationBadge.tsx:230-236`, `:302-308`)

`onKeyDown` dismisses on `Escape` only when `pinned` is true: `if (e.key === 'Escape' && pinned) { … closeKey(badgeKey); }`. A keyboard user who tabs to a badge and holds the pointer (or triggers hover via assistive tech) sees the temporary 400 ms preview, but `Escape` does nothing while `hovering` is true and `pinned` is false. `onClickOutSide` requires moving focus off the badge — which is discoverable but slow. Symmetric fix: change the guard to `if (e.key === 'Escape' && (pinned || hovering))` and clear the hover state, or simply always call `closeKey(badgeKey)` on `Escape` since it is idempotent through `setActiveKey(prev => prev === key ? null : prev)`. Same change applies to the `CitationGroupBadge` handler at `:302-308`.

## 数据流回溯（data-flow trace）

| Consumer | Upstream source | Verified reach |
|---|---|---|
| `shownIndex = displayIndex ?? index` (`CitationBadge.tsx:195`) | `node.properties.displayindex` set by `remarkCitation` via `resolveDisplay(ci)` = `displayIndexMap.get(raw) ?? raw` (`CitationText.tsx:118`) ← `buildDisplayIndexMap(normalized)` (`:278`) | ✅ reaches; `normalized = content.trim()` is the single source for both scan and render. |
| `displayIndices` string split in `CitationGroupBadge` (`:255-257`) | `displayindices` attribute set by `remarkCitation` `:154-158` via `group.indices.map(resolveDisplay).join(',')` | ✅ reaches through `rehype-sanitize` only because `citationSchema.attributes.citationgroup` explicitly allows `displayindices` (`:60`). |
| `groupCitations` re-keyed through `indicesKey` (`:260-263`) | Recomputed from `indices.join(',')` — raw indices, not display. Each `indices[i]` is looked up in `citations` by `.index`. | ✅ correct; display indices never touch the lookup. |
| `previewContent` hover branch (`:203-217`) consumes `citation.sender`, `sent_at`, `source`, `content` | `citation = citations.find(c => c.index === index)` (`:194`). If not found, early-return at `:199-201`. | ✅ guard prevents null deref; `source` additionally gated by `citation.source &&`. |
| `pinned` (`:196`) drives `useHoverPin` | `activeKey === badgeKey` from `CitationContext` | ✅ context provider at `CitationText.tsx:287-296` is the single source; `onBadgeClick` toggles. |
| `member` lookup in `TeamCitationBadge` (`:361`) | `members.find(m => m.user_id === citation.user_id)` | ✅ unchanged from base; `disableMemberPreview` still correctly short-circuits. |

## 盲点 checklist（C1–C6, security_sensitive routing）

- **C1 — Dual-path parity**: `CitationBadge` / `CitationGroupBadge` received the full hover+keyboard upgrade; `TeamCitationBadge` did not — see P2 #1. Not a security gap (no auth boundary), but a UX parity miss.
- **C2 — Control-flow ordering / reuse**: `useHoverPin` is reused in two components; both pass their own `pinned` and consume `visible` independently. No nesting/double-invoke risk. `buildDisplayIndexMap` is called once per render at `CitationText.tsx:278`; no double-scan.
- **C3 — Authorization boundary**: Not applicable. No auth/permission surface in the diff. `hidePlainCitations` privacy gate verified preserved.
- **C4 — Authorization lifecycle**: Not applicable (no parent-container status involved in this UI layer).
- **C5 — Build / note ≠ runtime correctness**: Static analysis only; `pnpm tsc --noEmit` and `pnpm test` not executed. `rehype-sanitize` schema change reviewed — display indices will survive the sanitizer.
- **C6 — Governance / docs self-consistency**: Not applicable (no policy / SECURITY.md / label change).

## Extra observations

- `TeamCitationBadge`'s `style={{ marginBottom: memberContent ? 4 : 0 }}` at `:371` is the single remaining inline-style hold-out; harmless but inconsistent with the rest of the file which has fully migrated to classes.
- The deprecated `buildDisplayIndexMap` re-export at `CitationText.tsx:72-75` exists to keep external callers (e.g. `SummaryVersionHistory`) working. Fine for this PR; worth grepping and removing once those imports migrate.

## 跨轮 blocker 复检

[Octo-Q] verdict: APPROVE — no P0/P1; three P2 non-blocking issues (TeamCitationBadge interaction parity, formatGroupLabel reversed-range edge, Escape-key hover dismissal). All are within the PR's claimed scope and fixable with small surgical changes, but none make a working path unavailable, produce wrong data, or cross an auth/privacy boundary. Static analysis only; recommend author run `pnpm --filter dmworksummary test` and `pnpm --filter dmworksummary typecheck` locally before merge.

Jerry-Xin (dismissed):
✅ **APPROVE** — citation badge UX modernization is byte-verified clean. XSS-safe, reading-order mapping correct + stable, click navigation preserved, hover-pin state isolated, CSS scoped.

## Verification (byte-level, ref = live head)

**XSS — safe.** No `dangerouslySetInnerHTML` / `innerHTML` anywhere in the touched files. All citation-derived strings (`content`, `source`, `sender`, member name) render as React text children → auto-escaped. `rehypeSanitize` + `citationSchema` lineage is preserved; the only schema change adds the numeric `displayindex` / `displayindices` attributes to the allowlist — numeric labels consumed by the component, no HTML-injection surface. Same discipline as the prior summary-content PRs.

**Reading-order renumbering — correct & stable.** `buildDisplayIndexMap` (in `citationFormat.ts`) and `remarkCitation` (in `CitationText.tsx`) use the *identical* token regex `/[(\d+)](?!()/` , so the pre-scan aligns exactly with what later renders as a badge. Display numbers are assigned 1-indexed in reading order, repeats of the same raw index reuse the same display number, and `resolveDisplay` falls back to the raw index when the map lacks it (safe). Crucially the internal `index` still keys the `citations` lookup — only the rendered label changes — so there is no off-by-one, no collision, and no "badge N opens source M". The 12 new unit tests cover renumbering, dedup-on-repeat, markdown-link exclusion `[n](url)`, and team-citation `[Pn]` exclusion.

**Click → source navigation — not regressed.** `onBadgeClick(badgeKey)` / `closeKey` and `JumpLink` still operate on the real citation (`firstCitation` / raw index); `badgeKey` is built from raw indices. Pinning routes through the same click handler, so hover does not interfere with click.

**Hover-pin state — isolated + cleaned up.** `useHoverPin` holds its own per-badge state and timer refs; the unmount effect clears both open/close timers (no listener/timer leak), and a pinned badge suppresses stale hover state. Keyboard path added (`tabIndex=0`, Enter/Space to pin, Escape to close) — a11y improved, not broken.

**CSS — scoped.** Every rule in `CitationBadge.css` is `.citation-*` prefixed; no bare-element or global selectors, no `!important`. Colors/spacing/radius come from `--wk-*` semantic tokens, light/dark track automatically. No global-selector leak.

**Build / tests.** CI Build check is green on the PR. Full `dmworksummary` unit suite locally: 12/12 new tests pass; all other failures are the known Node-26 `localStorage`-undefined environment noise (files this PR does not touch) — zero citation-related failures. `MERGEABLE`.

## 💬 Non-blocking nit
`buildDisplayIndexMap` scans the raw markdown string, while the badge renderer only fires on AST text nodes. If a `[n]` token ever appears inside a fenced code block, the pre-scan would consume a display number that never renders, leaving a cosmetic gap in the visible sequence (e.g. first visible badge shows `[2]`). No correctness/collision impact — purely cosmetic, and code-block citations are unlikely in summary bodies. Optional to guard later.

Nice work — clean token migration, well-tested pure formatting logic, and the reading-order/label separation is done correctly.

yujiawei (changes requested):
# Code Review — PR #1003 (octo-web)

**Head SHA:** `9f8aa1c1364474f1d7f948418399753479df6fc3`
**Scope:** `packages/dmworksummary/src/components/` — `CitationBadge.tsx`, `CitationText.tsx`, new `citationFormat.ts`, new `CitationBadge.css`, new `__tests__/CitationBadge.test.ts`.

Nice cleanup overall — extracting the pure formatting logic into `citationFormat.ts` for testability is the right call, the inline-style → token migration is mostly clean, and the hover/pin hook is well-structured. I found one correctness bug in the headline reading-order feature that should be fixed before merge, plus a few smaller items.

## Requirements coverage

All three maintainer-raised items are implemented (reading-order renumbering, token-based styling, group-label rule), plus the declared hover-preview / click-to-pin / a11y extras. No missing or contract-breaking changes at the requirements level — backend citation data and the `hidePlainCitations` / `disableTeamMemberPreview` flags are untouched, and shared consumers in `SummaryDetailPage.tsx` keep working. The one caveat is that the reading-order item, while present, has a correctness defect (below).

## Findings

### P1 — Reading-order renumbering miscounts when the body contains bracketed numbers outside real citations

`buildDisplayIndexMap` (`citationFormat.ts:37-50`) regex-scans the **raw markdown string** with `/\[(\d+)\](?!\()/g`, but `remarkCitation` (`CitationText.tsx:62-155`) only rewrites `[n]` found inside **AST `text` nodes** — `code` (fenced) and `inlineCode` nodes are never visited. The two scanners therefore disagree whenever the summary contains a `[digit]` token inside code (or a numeric reference-style link label). The raw scan inflates the map, so the first *rendered* citation can display as `[2]`, `[3]`, … with gaps, instead of `[1]`.

Reproduced against `buildDisplayIndexMap`:

- Source ```` ```\ncode [37]\n```\n\nReal citation [42] here. ```` → map `{37→1, 42→2}` → the only rendered badge `[42]` shows as **[2]**.
- Source `` The bug was in `items[0]` and `cfg[2]`. See [42]. `` → map `{0→1, 2→2, 42→3}` → `[42]` shows as **[3]**.

This is realistic input for an IM-summary product (summaries routinely quote code / array indexing), and it silently defeats the stated P1 goal ("first citation is `[1]`, no raw pool positions"). Note the new unit tests only cover the no-pollution path, so they pass while the feature is wrong.

The underlying citation data and click-to-source are unaffected (blast radius is display-only), which is why this is P1 and not P0 — but it is the advertised feature producing wrong output on common input, so it warrants one more round.

**Suggested fix:** build the display-index map during an **in-order AST `text`-node visit** rather than a raw-string pre-scan — `unist-util-visit` traverses in document (pre-order DFS) order, so a `text`-only visit is already in reading order *and* naturally skips `code`/`inlineCode`. (The code comment justifying the raw pre-scan says "the tree visitor sees text nodes out of document order"; that appears to be a misdiagnosis — visitor order is document order.) Alternatively, strip fenced/inline-code spans from the source before the pre-scan.

### P2 — `formatGroupLabel` range form can misrepresent or invert

For groups longer than `RANGE_THRESHOLD`, `formatGroupLabel` (`citationFormat.ts:19-24`) returns `${indices[0]}-${indices[last]}` of the **display** indices. When a group's members already received earlier reading-order numbers, the display list can be non-contiguous or non-monotonic, so the range label lies. Reproduced: an adjacent same-channel group `[10][30][40][50]` where 30/40/50 appeared earlier → display list `[4,1,2,3]` → label renders **`4-3`** (backwards). A `[1,3,4,5]` list would render `1-5`, implying `2` is part of the group when it may not be.

Display-only (clicking the group still shows the correct merged timeline), and uncommon, but the label is user-visible and misleading. Consider showing the actual set (comma-join / a count) or only using the range form when the display run is genuinely contiguous and ascending.

### P2 — `--wk-text-muted` is not a defined token

`CitationBadge.css` references `var(--wk-text-muted, rgba(28,28,35,0.55|0.45))` in 5 places, but `--wk-text-muted` is not defined anywhere in `packages/dmworkbase/src/theme/semantic.css` (which defines `--wk-text-secondary` / `--wk-text-tertiary`). The variable always resolves to the hard-coded `rgba(...)` fallback, so those muted colors are effectively hard-coded light-theme values — which contradicts the PR description's "no hard-coded colors / light+dark track automatically" claim and would not adapt if dark mode is re-enabled. (No user-facing impact today: dark mode is force-disabled at `packages/dmworkbase/src/App.tsx:803`.) Swap to an existing token such as `--wk-text-tertiary`, or add `--wk-text-muted` to both theme blocks in `semantic.css`.

### P2 — Keyboard a11y is asymmetric across badge types

`CitationBadge` and `CitationGroupBadge` correctly gained `tabIndex={0}` + Enter/Space/Escape handling, but `TeamCitationBadge` (`[Pn]`, `CitationBadge.tsx:424`) is still a mouse-only `<sup>` with no `tabIndex` or key handler, so `[Pn]` popovers can't be opened via keyboard. Apply the same treatment for consistency and full coverage. (Optional enhancement for all three: `role="button"` + `aria-haspopup`/`aria-expanded` so assistive tech announces them as popover controls.)

### Nits

- **Multiple simultaneous popovers:** pin state (`activeKey`) is global but `useHoverPin`'s `hovering` is per-badge, so pinning badge A and then hovering badge B shows both popovers at once. Minor UX, not blocking.
- **Dead re-export:** the `@deprecated export { buildDisplayIndexMap }` in `CitationText.tsx:60` has no remaining importers (all call sites already import from `citationFormat`); it can be dropped now rather than later.

## Verification performed

- New unit tests: **24 passed**. `pnpm turbo run build --filter=@octo/web`: **build succeeds** (~6s, pre-existing warnings only). CI gate is `build` + `lint` + `i18n:check` (not `tsc`), consistent with repo config — standalone `tsc` `noImplicitAny` output is a config artifact, not a gate.
- Findings 1 & 2 reproduced empirically with temporary tests against `buildDisplayIndexMap` / `formatGroupLabel`.
- Confirmed every `--wk-*` token used in the new CSS exists in `semantic.css` **except** `--wk-text-muted`.
- **Security:** this PR was auto-flagged security-sensitive on a keyword match ("token" = design token / "[n] token"). The diff touches only presentational summary-citation UI — no auth, credential, network, or storage surface. No security concerns.
- **Not verified at runtime:** live hover/pin timing, in-browser popover rendering, dark-theme appearance (disabled), and behavior against real backend citation payloads.

## Verdict

**Changes requested** — one P1 correctness bug in the reading-order feature (display numbers wrong/gapped when the summary contains bracketed numbers in code), plus P2 polish items. The P1 is display-only (no data loss, navigation still correct) and has a clean, small fix.

wanyiwang06:
@yujiawei 已按你的 review 修完,请复审当前 head `e205ff2`：

- 🔴 **P1（reading-order 误计）**：改成在 `remarkCitation` 里从**已访问的 `text` 节点**（文档顺序，`visit(tree,"text")` 天然不进 code/inlineCode）构建 display map，与实际渲染成 badge 的内容严格一致；移除了对 raw markdown 串的预扫。`buildDisplayIndexMap` 现在接收有序文本段 `string[]`,仍是纯函数、可单测。加了 code 排除的回归用例。
- 🟡 **formatGroupLabel**：仅当 >3 的 display 列表**连续且升序**才用 `a-b` 区间,否则回退逗号列表（修 `[4,1,2,3]→"4-3"` 倒序 / `[1,3,4,5]→"1-5"` 虚构成员）。加了非连续/倒序用例。
- 🟡 **`--wk-text-muted` 未定义 token**：4 处改用已定义的 `--wk-text-tertiary`（semantic.css 明/暗两套都有）。
- 🟡 **dead re-export**：删除 `CitationText` 里 `@deprecated export { buildDisplayIndexMap }`（已无引用者）。

**自验**：citation 单测 14/14；`pnpm turbo run build --filter=@octo/web` ✓。（唯一失败的 `SummaryVersionHistory` 用例与本 PR 无关,在未改的 base 上同样失败。）

**暂缓（你标的 P2）**：`TeamCitationBadge`([Pn]) 键盘 a11y 对齐——它涉及 role/aria 设计,建议单独一个小 PR,避免混进这次 citation 改动。要合进本 PR 我也可以补,你说。

lml2468 (approved):
## Re-review: #1003 @ `e205ff26` — 引用角标 UX 现代化(针对 yujiawei 两条 🔴 的修复复核)

上一版我 APPROVE 后,yujiawei/Jerry-Xin 坐实两条 🔴:(P1) 阅读序号重编号误计代码块内 `[n]` → 序号错乱;(P2) `formatGroupLabel` 范围标签可能倒序/含不存在的数字。本轮复核两条是否修复。

锚定 head `e205ff26` / merge-base `bcbed66e`(behind 36,`mergeable=MERGEABLE`,0 冲突)。

---

### 🔴 P1 —— 代码块内 `[n]` 误计致序号错乱 ✅ 已修(改在源头)

- `buildDisplayIndexMap` 签名从 `(source: string)` 改为 `(textSegments: string[])` —— 输入改为**按文档序访问到的 markdown text 节点值**,而非裸源码串。
- `CitationText` 里 `visit(tree, 'text', node => textSegments.push(node.value))` 收集 —— `visit 'text'` **永不进入 `code`/`inlineCode` 节点**,故 fenced/inline 代码里的 `[n]` 天然被排除。
- 关键:该 map 在 `remarkCitation` 内基于**渲染 badge 用的同一 AST** 构建 → 编号与实际渲染逐一对齐,不再把首个 badge 挤成 `[2]/[3]`。✅

### 🔴 P2 —— 范围标签倒序/含错误数字 ✅ 已修

- `formatGroupLabel` 在 `>3` 分支前先判 `contiguousAscending`(`indices.every((v,i)=> i===0 || v===indices[i-1]+1)`):
  - 连续升序 → `first-last` 范围(如 `[30-35]`);
  - 非连续(`[2,5,9,14]`)或非升序(`[4,1,2,3]`)→ **回落逗号列表**,不再渲染出 `4-3`(倒序)或 `1-5`(隐含不存在的 `[2]`)。✅

**两条都补了针对性回归测试**:
- P2:`falls back to comma list when a >3 display list is non-contiguous`(`[2,5,9,14]`→`'2,5,9,14'`)、`...is not ascending`(`[4,1,2,3]`→`'4,1,2,3'`)。
- P1:`does not count [n] inside code — code text is never a text segment`(`code [37]` 段被排除 → `[42]` 成为 `[1]`,`m.has(37)===false`)。

---

### 构建 / 测试(实跑)

- `pnpm --filter @octo/web build` ✅ built in 3.50s。
- `CitationBadge.test.ts` ✅ **14/14 全绿**(较上轮 12→14,新增两条 P1/P2 回归)。
- **typecheck(应用 #987 教训,单独实跑)**:改动的三个源文件(`citationFormat.ts`/`CitationText.tsx`/`CitationBadge.tsx`)**无本 PR 引入的类型错误**。`CitationText.tsx:295 'ReactMarkdown' cannot be used as a JSX component` 已核实为 **pre-existing**——该 `<ReactMarkdown` 行 git blame 显示 Jerry Xin 2026-04-29 所写(本 PR 仅令其行号由 271→295 位移),同错误在 merge-base 上出现 4 次,系 Node-26 dual-React-types 环境噪声,非本 PR 缺陷。✅

---

### 结论

两条 🔴 均在**源头**修复(P1 从「裸串 pre-scan」改为「AST text 节点」,天然排除代码块;P2 加连续升序守卫回落逗号),并各有针对性回归测试锁定;编号仍与渲染同 AST 对齐、数据 lookup 仍用 raw index、无 off-by-one;构建通过、14/14 测试绿、改动文件无新类型错误。

yujiawei / Jerry-Xin 的两条 🔴 已正确修复。**可合并。**(致谢逐轮坐实序号/范围边界。)

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 blockers found. Clean data-flow from AST text nodes through rehype-sanitize schema to React props. Hover/pin lifecycle correctly implemented. Only non-blocking P2 (CSS token fallback inconsistency, follows codebase pattern) and one nit (stale comment wording).

yujiawei (approved):
# Code Review — PR #1003 (octo-web)

**Head SHA:** `e205ff26e805d8d493c51e8f8bcb37a266918a8a`
**Scope:** `packages/dmworksummary/src/components/` — `CitationBadge.tsx`, `CitationText.tsx`, `citationFormat.ts`, `CitationBadge.css`, `__tests__/CitationBadge.test.ts`.

This is a re-review of the fix commit (`e205ff26`) addressing the previous round's `CHANGES_REQUESTED`. The prior P1 correctness bug and the P2 polish items are resolved; the reading-order feature is now correct by construction. Remaining items are all non-blocking P2/UX. **Approving.**

## 1. Spec compliance

**Spec: ✅**

- **Missing:** none. All three maintainer-raised items are implemented — reading-order renumbering (first rendered badge shows `[1]`), token-based styling, and the group-label rule (`≤3` comma-joined / `>3` range). The declared hover-preview / click-to-pin / keyboard-a11y extras are present as described.
- **Over-build:** none. Backend `CitationIndex` assignment and API payloads are untouched; the stated non-goals (AI banner, bottom sources list, by-speaker aggregation) are not introduced.
- **Deviation:** none. The `hidePlainCitations` and `disableTeamMemberPreview` privacy flags are preserved in both the component and its callers; display numbering changes the label only, not the underlying `index` used for lookup and navigation.

## 2. Code Quality

**Quality: Approved**

### Previous-round findings — all verified fixed

- **[was P1] Reading-order miscount on bracketed numbers in code — FIXED.** The display-index map is now built inside `remarkCitation` from an in-order `visit(tree, 'text', …)` pass (`CitationText.tsx:62-65`), the *same* traversal that rewrites the badges (`:68`). Because both passes see the identical `text`-node set in identical document order, `code`/`inlineCode` `[n]` tokens are excluded by construction and the two scanners can no longer disagree. I re-ran my prior repro against `buildDisplayIndexMap`: a summary whose only real citation is `[42]` (with `[37]` living in a code span) now maps `{42→1}`, so the badge renders `[1]`. Correct.
- **[was P2] `formatGroupLabel` range inversion — FIXED.** `citationFormat.ts:28-33` now emits the range form only when the display list is contiguous *and* strictly ascending; otherwise it falls back to the comma list. Verified: `[4,1,2,3]→"4,1,2,3"` (no backwards `4-3`), `[1,3,4,5]→"1,3,4,5"` (no phantom `2`).
- **[was P2] Undefined `--wk-text-muted` token — FIXED.** Now uses `--wk-text-tertiary`; all 13 `--wk-*` tokens in `CitationBadge.css` resolve in `packages/dmworkbase/src/theme/semantic.css`.
- **[was nit] Dead `buildDisplayIndexMap` re-export — REMOVED.**

### Remaining (P2 — non-blocking)

- **P2 — `[Pn]` team badges remain mouse-only** (`CitationBadge.tsx:424`). `CitationBadge` and `CitationGroupBadge` gained `tabIndex={0}` + Enter/Space/Escape, but `TeamCitationBadge` still has no `tabIndex` or key handler, so `[Pn]` popovers can't be opened via keyboard. Applying the same treatment (ideally plus `role="button"` + `aria-haspopup`/`aria-expanded` on all three) would complete keyboard coverage.
- **P2 — pending open-timer survives an unpin** (`CitationBadge.tsx:169-186`). Pinning does not cancel a pending `openTimer`. Edge case: hover ~300 ms → click to pin → click again to unpin before 400 ms elapses; the original timer then fires and re-opens the preview despite the explicit close. Clearing `openTimer` when `pinned` becomes true would close the gap.
- **P2 — hover grace period can't reach the preview card** (`CitationBadge.tsx:175`, `:243`). Only the badge carries `onMouseEnter`/`onMouseLeave`; the Popover content can't cancel `closeTimer`, so moving the pointer from badge into the card to select/scroll dismisses it after 200 ms. Wiring the same enter/leave handlers onto the popover content would let users interact with the hovered card. (Pinning via click is the intended path for interaction, so this is minor.)
- **Nit — concurrent popovers:** `activeKey` (pin) is global but `hovering` is per-badge, so pinning A then hovering B can briefly show both. Minor.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. The previously-blocking P1 is fixed and independently verified; no P0/P1 remain. The listed P2s are UX/a11y polish, not merge blockers.

## 4. Suggestions

- Add a follow-up to bring `[Pn]` team badges to keyboard parity with the other two badge types (the highest-value remaining item since it's an accessibility gap, not just polish).
- Consider a small test asserting `buildDisplayIndexMap` output for a multi-`text`-node tree (the fix's core invariant is "map order == render order"; current tests exercise the pure function but not the AST wiring).

## 5. Additional observations

- **Security (PR was auto-flagged security-sensitive):** the keyword hit is "token" (design token) / "`[n]` token"; the diff touches only presentational summary-citation UI. No auth, credential, network, or storage surface. Citation content is rendered via React text interpolation with no `dangerouslySetInnerHTML`, and `rehypeSanitize` remains applied with an explicit tag/attribute allowlist (`CitationText.tsx:44-53, 297`). No injection concern.
- **Verification performed:** `pnpm install --frozen-lockfile` on this head; the new suite runs **14/14 pass**; `pnpm turbo run build --filter=@octo/web` **succeeds** (~7 s, pre-existing third-party warnings only). The two pure functions were additionally exercised against adversarial inputs (gapped/non-monotonic groups, cross-segment dedup, markdown-link and `[Pn]` exclusion) — all correct.
- **Not verified at runtime:** live in-browser hover/pin timing and Semi Popover portal behavior; dark theme is force-disabled in this build so its appearance was not exercised.

Jerry-Xin (approved):
# Re-review — PR #1003 (octo-web) ✅ APPROVE

**Head:** `e205ff26e805` · re-review of my prior APPROVE at `9f8aa1c1` (auto-dismissed on the new push).
**Delta scope:** 4 files — `citationFormat.ts`, `CitationText.tsx`, `CitationBadge.css`, `__tests__/CitationBadge.test.ts`. `CitationBadge.tsx` (hover-pin hook) is byte-identical to the prior-approved head (blob unchanged) — no hover/pin/a11y regression possible.

## What changed since `9f8aa1c1`

Addresses the review findings from the prior round (P1 reading-order + two P2s):

- **P1 (reading-order correctness — the one that mattered):** `buildDisplayIndexMap` no longer pre-scans the raw markdown string. It now takes the ordered visible **text-node values** (`string[]`) and the map is built **inside** `remarkCitation` from a `visit(tree, 'text', …)` pass. `[n]` tokens inside fenced / inline code are excluded by construction (a `text`-only visit never enters `code`/`inlineCode` nodes).
- **P2:** `formatGroupLabel` uses the `a-b` range form only when the >3 display list is contiguous **and** ascending; otherwise it falls back to the explicit comma list (fixes reordered lists rendering backwards like `4-3`, or `1-5` implying an absent member).
- **P2:** `CitationBadge.css` swapped the undefined `--wk-text-muted` → the defined `--wk-text-tertiary` (present in both theme blocks) in 4 places.
- Removed the dead `@deprecated` re-export of `buildDisplayIndexMap` from `CitationText.tsx` (no remaining importers).
- Tests updated to the `string[]` signature + a code-exclusion regression case + non-contiguous / non-ascending group-label cases.

## Critical lockstep check — pre-scan numbering set === render numbering set ✅

This was the high-risk surface (a mismatch would make badge *N* open source *M*). Verified it is now **exact by construction**:

- Numbering pass: `visit(tree, 'text', node => segments.push(node.value))` → `buildDisplayIndexMap(segments)` using `/\[(\d+)\](?!\()/g`.
- Render pass: a second `visit(tree, 'text', …)` applies the **identical** regex `/\[(\d+)\](?!\()/g` to the same `text` nodes.

Both passes traverse the **same set of `text` nodes in the same document order** with the **same token regex**; `code`/`inlineCode` are skipped by both. So the counting set equals the render set — no new off-by-one, no gap, no drift. Per-segment iteration uses a fresh regex (no `/g` `lastIndex` carryover). No new mismatch introduced by the code-block-skip fix.

Also re-confirmed on the delta:
- 1-indexed (`next = 1`); repeat raw index reuses its number (`if (!map.has(raw)) map.set(raw, next++)`).
- `resolveDisplay = raw => map.get(raw) ?? raw` fallback preserved (single + group paths).
- Internal `index: ci` still carries the raw citations-array key → click resolves the correct source; display number is cosmetic only. No collision.

## XSS surface — unchanged ✅

No `dangerouslySetInnerHTML` / `innerHTML` / `eval` / `__html` in any touched file. `rehypeSanitize` + `citationSchema` lineage preserved; schema still exposes only the small numeric/key attribute set (`index` / `displayindex` / `displayindices` / `badgekey`). `remarkTeamCitation` untouched; links keep `rel="noopener noreferrer"`.

## Build / tests

- **App build** (`turbo run build --filter=@octo/web`, which compiles the consuming app + this package): **SUCCESS**.
- **Citation tests:** `CitationBadge.test.ts` + `CitationText.test.tsx` → **26/26 passed**.
- Note: a standalone package-level `tsc --noEmit` in a raw checkout emits a uniform "cannot find declaration file for react" cascade across *all* files (including files this PR never touches) — a type-resolution/env artifact of the isolated checkout, not a defect in this change. The authoritative build (CI Build + the app-level turbo build) is green.

## Peers / status

CI green · mergeable · peers approving at this head.

**Verdict: APPROVE.** The P1 reading-order fix is implemented the correct way (numbering derived from the same in-order text-node visit that renders), the pre-scan and render numbering sets are provably in lockstep, XSS surface is unchanged, hover-pin/a11y untouched, and build + citation tests pass.

PR #1006, opened 2026-07-22
fix(mcp): inline shared-secret warning instead of window.confirm

l-s-c:
## Summary

Follow-up to #995. The submit-time guard added in that PR (advises the operator when a secret-shape key is being published as a shared value) shipped as a `window.confirm` — the browser-native "im-test.deepminer.com.cn 显示" alert, visually out of place against the Semi UI marketplace form.

This PR replaces it with a non-blocking inline advisory rendered next to the Submit button. Same trigger and same intent; the KvEditor's smart-default-on-secret-key-input still catches the common accidental case, so this advisory is the persistent, quiet reminder for the explicit-override case.

## Related Issue

Follow-up to #995. No new spec surface — the toggle model, wire contract (`env_user_supplied` / `headers_user_supplied`), and marketplace §5.1 semantics are unchanged.

## Changes

- Add \`sharedSecretLeaks\` memo scanning env + headers entries for `!userSupplied && value && isSecretKey(k)`.
- Render an inline `⚠️ <keys> will be published in cleartext` span between the primary Submit button and any prior footer content; title tooltip carries the full text.
- Drop \`window.confirm\` block from \`handleSubmit\`.
- Drop obsolete i18n keys \`sharedSecretConfirm{,Title,Ok,Cancel}\`; replace with a single \`sharedSecretWarning\` line.
- CSS: `.wk-mcp-form-footer__warning` (warning color, ellipsis on overflow, hover tooltip via `title`).

- Affected module(s): `packages/dmworkmcp` (MCP marketplace create/edit modal only).
- New or changed user-visible entry point: no.
- Shared layer touched: none.
- Duplicate entry point checked: yes; no duplication.

## Testing

- `pnpm --filter @dmwork/mcp test` — 4 files / 56 tests passing.
- `pnpm --filter web build` — green.
- Manually: opened create modal, typed `Authorization: Bearer sk-xxx` without flipping the toggle, saw the amber warning next to Submit; hitting Submit went straight through (non-blocking). Toggling ON cleared the warning.

- [x] Unit tests still pass
- [x] Manually verified in dev
- [x] No new user-visible entry point

### DIFF
--- a/packages/dmworkmcp/src/components/McpCreateModal.tsx
+++ b/packages/dmworkmcp/src/components/McpCreateModal.tsx
@@ -786,36 +786,12 @@ const McpCreateModal: React.FC<McpCreateModalProps> = ({
     // Collapse the structured editors down to the wire pair. Backend no
     // longer rejects secret-shaped shared values (rule 2 was removed);
     // non-owner blanking (§5.3) is the sole guard keeping author tokens
-    // out of consumer-facing responses.
+    // out of consumer-facing responses. The inline warning next to the
+    // Submit button (see sharedSecretLeaks below) is the frontend's
+    // advisory signal for the edge case where the operator explicitly
+    // published a secret-shape shared value; non-blocking on purpose.
     const envWire = entriesToWire(envEntries);
     const headersWire = entriesToWire(headersEntries);
-
-    // Front-line safety net for the (unusual) case where the operator
-    // manually flipped the toggle OFF on a secret-shape key AND typed a
-    // real value: they've signalled "this is a shared value". Backend
-    // will accept and store it (§5.1 rule 2 was removed); the record
-    // still round-trips fine to owner but non-owners see blanks. Warn
-    // and require an explicit OK so pasting a personal token by mistake
-    // doesn't silently ship. Keys already ON (toggle set to
-    // user_supplied) are safe by design — placeholder rendering + wire
-    // blanking both apply.
-    const suppliedH = new Set(headersWire.userSupplied);
-    const suppliedE = new Set(envWire.userSupplied);
-    const sharedSecretKeys: string[] = [];
-    for (const [k, v] of Object.entries(headersWire.values)) {
-      if (v && isSecretKey(k) && !suppliedH.has(k)) sharedSecretKeys.push(k);
-    }
-    for (const [k, v] of Object.entries(envWire.values)) {
-      if (v && isSecretKey(k) && !suppliedE.has(k)) sharedSecretKeys.push(k);
-    }
-    if (sharedSecretKeys.length > 0) {
-      const ok = window.confirm(
-        t("mcp.create.sharedSecretConfirm", {
-          values: { keys: sharedSecretKeys.join(", ") },
-        })
-      );
-      if (!ok) return;
-    }
 
     setSubmitting(true);
     // Upload icon FIRST so the URL can ride in the create/update body. The
@@ -936,6 +912,25 @@ const McpCreateModal: React.FC<McpCreateModalProps> = ({
       "tools",
       form.tools.map((tool, i) => (i === idx ? { ...tool, ...patch } : tool))
     );
+
+  // Derived warning list: secret-shape keys the operator kept toggled OFF
+  // (i.e. published as shared values) with a non-empty value. Rendered
+  // inline next to the Submit button so the operator sees the exposure
+  // before clicking, without a blocking modal. Non-blocking on purpose —
+  // the KvEditor auto-toggles ON when a secret-shape key is typed into a
+  // fresh row, so hitting this state requires an explicit override; the
+  // warning is here for the operator's own audit, not to gate publish.
+  const sharedSecretLeaks = useMemo(() => {
+    const leaks: string[] = [];
+    for (const e of [...headersEntries, ...envEntries]) {
+      const k = e.key.trim();
+      if (!k) continue;
+      if (e.userSupplied) continue;
+      if (!e.value) continue;
+      if (isSecretKey(k)) leaks.push(k);
+    }
+    return leaks;
+  }, [headersEntries, envEntries]);
 
   // ── Static options ─────────────────────────────────────────────────────
   const categoryOptions = useMemo(
@@ -1120,13 +1115,28 @@ const McpCreateModal: React.FC<McpCreateModalProps> = ({
                 {t("mcp.create.next")} →
               </WKButton>
             ) : (
-              <WKButton
-                variant="primary"
-                loading={submitting}
-                onClick={handleSubmit}
-              >
-                {isEdit ? t("mcp.edit.submit") : t("mcp.create.submit")}
-              </WKButton>
+              <>
+                {sharedSecretLeaks.length > 0 && (
+                  <span
+                    className="wk-mcp-form-footer__warning"
+                    title={t("mcp.create.sharedSecretWarning", {
+                      values: { keys: sharedSecretLeaks.join(", ") },
+                    })}
+                  >
+                    ⚠️{" "}
+                    {t("mcp.create.sharedSecretWarning", {
+                      values: { keys: sharedSecretLeaks.join(", ") },
+                    })}
+                  </span>
+                )}
+                <WKButton
+                  variant="primary"
+                  loading={submitting}
+                  onClick={handleSubmit}
+                >
+                  {isEdit ? t("mcp.edit.submit") : t("mcp.create.submit")}
+                </WKButton>
+              </>
             )}
           </div>
         </div>
--- a/packages/dmworkmcp/src/i18n/en-US.json
+++ b/packages/dmworkmcp/src/i18n/en-US.json
@@ -181,7 +181,7 @@
     "headerAdd": "Add",
     "headerRemove": "Remove",
     "kvUserSuppliedToggle": "User fills locally",
-    "sharedSecretConfirm": "These keys look sensitive but their \"user fills locally\" toggle is OFF, so the values will be submitted verbatim: {{keys}}. Publish anyway?",
+    "sharedSecretWarning": "{{keys}} will be published in cleartext",
     "icon": "Icon",
     "iconEmpty": "Upload",
     "iconUpload": "Upload image",
--- a/packages/dmworkmcp/src/i18n/zh-CN.json
+++ b/packages/dmworkmcp/src/i18n/zh-CN.json
@@ -181,7 +181,7 @@
     "headerAdd": "添加",
     "headerRemove": "删除",
     "kvUserSuppliedToggle": "需要用户自行配置",
-    "sharedSecretConfirm": "以下 key 是敏感字段，但你没打开「需要用户自行配置」开关，值会以明文提交到市场：{{keys}}。确认发布？",
+    "sharedSecretWarning": "{{keys}} 将以明文公开发布",
     "icon": "图标",
     "iconEmpty": "点击上传",
     "iconUpload": "上传图片",
--- a/packages/dmworkmcp/src/index.css
+++ b/packages/dmworkmcp/src/index.css
@@ -969,7 +969,20 @@
 
 .wk-mcp-form-footer__right {
   display: flex;
+  align-items: center;
   gap: var(--wk-sp-2);
+}
+
+/* Inline advisory shown next to the Submit button when the operator kept
+   a secret-shape key toggled OFF with a real value. Non-blocking; the
+   colour reads as a warning without stealing the button's affordance. */
+.wk-mcp-form-footer__warning {
+  font-size: var(--wk-text-size-sm);
+  color: var(--wk-color-warning, #d97706);
+  max-width: 24em;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 
 /* Detail-modal footer row for owner actions (edit/delete). */


lml2468 (approved):
## Review: #1006 — fix(mcp): 共享密钥告警改为内联提示(替换 window.confirm)

**意图**:#995 的后续。#995 提交前对「密钥形 key 但未开 user-supplied 开关且有值」弹的 `window.confirm`(浏览器原生 alert,与 Semi UI 表单割裂)替换为 Submit 按钮旁的**非阻塞内联告警**。触发条件与意图不变。

锚定 head `10257949` / merge-base `a563a701`(behind 0,`mergeable=MERGEABLE`,0 冲突)。

---

### Gate 1 — 规格符合度 ✅

- 检测逻辑等价迁移:旧 `handleSubmit` 里的 `sharedSecretKeys`(`v && isSecretKey(k) && !supplied`)→ 新 `sharedSecretLeaks` memo(`!e.userSupplied && e.value && isSecretKey(k.trim())`),扫描 `headersEntries + envEntries`。同一触发面。✅
- 渲染:Submit 按钮旁 `⚠️ <keys> will be published in cleartext`,`title` 挂全文 tooltip,overflow 省略。✅
- i18n:旧 `sharedSecretConfirm{,Title,Ok,Cancel}` 删除、单条 `sharedSecretWarning` 加入 en/zh 双语,无残留引用。✅
- 无新增用户可见入口、无共享层改动,scope 收敛于 dmworkmcp。✅

### Gate 2 — 代码质量 ✅

**安全姿态变更判断(重点)**:本 PR 把**阻塞式 confirm(须显式 OK 才能发布共享密钥)降为非阻塞内联提示(Submit 直接通过)**。这是有意的安全姿态放松,**判定为可接受**,理由三层:
1. **真正的保障在后端**:非 owner 看不到 author token 靠 marketplace §5.3 non-owner blanking,前端 confirm/告警始终只是 advisory,非授权闸;
2. **意外场景已被输入期兜底**:KvEditor 对密钥形 key 的 fresh 行自动翻 `userSupplied=ON`,要落到「共享的密钥形 key 且有值」必须操作者**显式关掉**智能默认 —— 即已是明确 override;
3. 因此 confirm 本就只是对「显式 override」的二次确认,改为持久可见的内联告警(点 Submit 前一直可见,比一次性 confirm 的可审计性反而更好)。
无泄漏面净增。✅

**构建 / 测试(实跑):**
- `pnpm --filter @octo/web build` ✅ built in 3.52s。
- `pnpm --filter @dmwork/mcp test` ✅ **4 files / 56 tests 全绿**。
- i18n parity:mcp **216/216**。
- **typecheck(应用 #987 教训:docs/mcp 这类无 build task 的包 CI 绿 ≠ typecheck 过,单独实跑)**:`McpCreateModal.tsx`(本 PR 唯一改动的源文件)**0 处真实类型错误**(HEAD 与 merge-base 均为 0);包内 120 处 tsc 报错全是 Node-26 dual-React-types 环境噪声(缺 'react' 声明 / JSX component / setState),在 merge-base 上**逐条相同**,非本 PR 引入。**与 #987 不同**——#987 是 PR 新增的 `boardFiles.test.ts` 带了 2 处真类型错误(`FileFetchRef.status`、未判空 `dataURL`);本 PR 无任何新增类型错误。✅

---

### 🟡 非阻断说明(承 #995)

**`sharedSecretLeaks` 告警逻辑仍无回归测试**——这正是我在 #995 就提的缺口,且本轮该守卫行为已从阻塞变非阻塞,更值得锁契约。建议补断言:密钥形 key + toggle OFF + 有值 → 告警渲染出对应 keys;toggle ON / 无值 → 无告警。因逻辑简单、后端 §5.3 兜底,不阻断。

内联告警检测等价迁移、安全姿态放松有后端兜底 + 输入期自动开关双重支撑而可接受、i18n 清理干净、`McpCreateModal.tsx` 零真实类型错误、构建与 56 测试通过。**可合并。**

Jerry-Xin (approved):
**APPROVE** — inline advisory swap is byte-clean; detection logic preserved; build + tests green. One deliberate behavior change (blocking → passive) called out below as a non-blocking note so it's a conscious product decision, not an accident.

### What I verified (byte-level, ref = live head)

**Detection condition unchanged.** The old submit-time guard fired on `value present && isSecretKey(key) && !userSupplied`. The new `sharedSecretLeaks` memo fires on `key(trimmed) && !userSupplied && value && isSecretKey(key)` — same three predicates (has value, shared / toggle-OFF, secret-shape key). It reads the raw entry list instead of the wire-collapsed pair, but that's functionally equivalent. `isSecretKey` itself is untouched. Not loosened.

**No secret leak / no XSS.** The advisory echoes **key names only** (`sharedSecretLeaks.join(", ")`), never the value. It renders as React text and as a `title` attribute — both auto-escaped. No `dangerouslySetInnerHTML`. Clean.

**Warning placement + clearing.** Single advisory next to Submit listing all offending keys (not per-row, so no off-by-one). The memo's deps recompute on entry changes, so it clears correctly when the operator flips the toggle ON, clears the value, or renames the key to a non-secret shape.

**i18n parity + dead code.** `sharedSecretWarning` present in both `en-US.json` and `zh-CN.json` (same line). The old `sharedSecretConfirm` (and any `Title/Ok/Cancel` variants) is gone repo-wide — no orphan. The `window.confirm` block is fully removed from `handleSubmit`; no leftover references.

**Build / tests / merge.** `web` build green. `@dmwork/mcp` suite 56/56 passing. gitleaks + OSV green. MERGEABLE (status shows BLOCKED only on the pending review gate + in-progress Build, not a conflict). Note: a bare `tsc -p` run against the package emits a wall of `implicitly any` / missing-react-declaration errors, but those pre-date this PR (they hit every callback param across the 1600-line file) and stem from the standalone invocation not resolving types — the project build path is the authoritative typecheck and it passes. The two lines this PR adds introduce no new type surface.

### 💬 Non-blocking note — this changes a *blocking* guard into a *passive* one

The #995 guard used a modal that returned early unless the operator confirmed — submit was **gated**. This version is a persistent inline hint and `handleSubmit` fires unconditionally, so a shared secret-shape value now publishes on the first click with only a visible warning. That's a real behavior change, and the PR body owns it as intentional.

I think it's acceptable, because:
- The backend rule that rejected secret-shape shared values was already removed, and non-owner response blanking is the actual security control — the frontend check was always advisory (an accidental "OK" published just the same).
- The KvEditor still auto-flips the toggle ON when a secret-shape key is typed into a fresh row, catching the common accidental path before it ever reaches this state.

Two edge cases the old modal caught that the passive hint no longer *stops* (only warns on): value typed/pasted **before** the key, and edit-mode rows loaded with a pre-existing key. These now publish on first click. If product considers "shared secret publish must require an explicit acknowledgement" a hard requirement, gate Submit on the warning being empty (or add a one-time inline confirm affordance) instead of relying on the hint alone. Flagging for a conscious decision, not blocking the merge.

### ✅ Highlights
- Clean removal of the browser-native modal that clashed with the form's UI.
- Advisory recomputed reactively — no stale-warning risk.
- Truncation + `title` tooltip handle long key lists gracefully.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1006 (octo-web)

Reviewed at head `10257949a71f01faad2ea76163e5e29219b2fcd5`. Scope: replace blocking `window.confirm` for shared-secret publication with a non-blocking inline warning in the MCP create/edit modal.

## Summary

This PR removes the `window.confirm()` dialog that previously blocked submission when the operator kept a secret-shaped key toggled OFF (published as a shared value) with a non-empty value. It replaces that blocking gate with a `useMemo`-derived `sharedSecretLeaks` array rendered as an inline advisory `<span>` next to the Submit button. The detection logic is functionally equivalent to the old code, and the i18n keys are renamed cleanly (`sharedSecretConfirm` → `sharedSecretWarning`) in both locales.

The security trade-off is intentional and well-justified: `KvEditor`'s auto-toggle (`setKey` at `McpCreateModal.tsx:297`) already prevents the most common accidental-publish scenario (typing a secret-shaped key into a fresh row auto-flips `userSupplied` ON), and the backend's non-owner blanking (§5.3) provides a second-layer defense. Reaching the warning state requires an explicit operator override, making the advisory appropriate for the residual risk.

## Verification

- ✅ **Detection equivalence** — the new `sharedSecretLeaks` useMemo iterates `[...headersEntries, ...envEntries]` with the same three guard conditions as the removed code: truthy value, `!userSupplied`, and `isSecretKey(k)`. Both old and new use `.trim()` on keys. The new code is strictly more conservative for duplicate-key entries (each row checked independently vs. last-write-wins dedup in `entriesToWire`).
- ✅ **i18n cleanup** — `sharedSecretConfirm` has zero remaining references in the codebase; `sharedSecretWarning` is consumed at exactly two call sites in the component (`:1122`, `:1127`) plus two locale definitions.
- ✅ **Wire payload unchanged** — the submit handler at `:831–834` still sends `env`, `envUserSupplied`, `headers`, `headersUserSupplied` unconditionally regardless of transport type. No regression in the create/update payload shape.
- ✅ **Edit-mode hydration** — `entriesFromWire` at `:520–522` correctly reconstructs `KvEntry[]` with the `userSupplied` flag from wire data, so the warning renders accurately in edit mode too.
- ✅ **CSS scope** — `.wk-mcp-form-footer__right` gains `align-items: center`; new `.wk-mcp-form-footer__warning` class uses `text-overflow: ellipsis` with `max-width: 24em` and a `title` attribute for overflow. Purely cosmetic, no layout risk.

Static analysis only at head `10257949`; build and tests not executed in this environment.

## Findings

### P2 — Inline warning lacks ARIA live region for screen-reader users (`packages/dmworkmcp/src/components/McpCreateModal.tsx:1120`)

The `<span>` rendering the shared-secret advisory has no `role="alert"` or `aria-live` region. Sighted users see the ⚠️ icon and amber text next to the Submit button, but screen-reader users receive no announcement when the warning appears or updates — for example, when the operator toggles a key OFF after typing a value, the warning materializes without any assistive-tech signal. The `title` attribute provides hover-tooltip parity but does not reach keyboard-navigation or screen-reader users.

**Diff-scope:** new (introduced by this PR — the old `window.confirm` was natively accessible via the browser's modal dialog).

Fix: add `role="alert"` to the `<span>` (or wrap in an `aria-live="polite"` region) so the warning is announced when it enters the DOM.

## Data-Flow Tracing

|---|---|---|
| `sharedSecretLeaks` array | `[...headersEntries, ...envEntries]` → filtered by `!e.userSupplied && e.value && isSecretKey(e.key.trim())` | ✅ Yes — both state arrays are React `useState` at `:487-488`, updated by `KvEditor` onChange or `entriesFromWire` hydration. All three filter conditions match the old submit-time check. |
| `e.userSupplied` per entry | Set by: (a) KvEditor auto-toggle at `:297` (fresh secret-shape rows), (b) manual Switch toggle at `:330`, (c) `entriesFromWire` edit-mode hydration at `:520-522`, (d) JSON import seeding at `:1061` | ✅ Yes — all four initialization paths set the boolean correctly. |
| `e.value` per entry | Set by: (a) `WKInput` onChange at `:325`, (b) `entriesFromWire` (strips `SECRET_PLACEHOLDER_SENTINEL` to empty), (c) JSON import (always empty string) | ✅ Yes — truthy check `!e.value` skips empty strings, matching old `if (v && ...)` semantics. |
| `isSecretKey(k)` | `SECRET_KEY_PATTERN` regex at `utils/constants.ts:35-37`, case-insensitive suffix match | ✅ Same regex in both old and new code; no drift. |
| `t("mcp.create.sharedSecretWarning", ...)` | i18n key at `en-US.json:184` / `zh-CN.json:184`, interpolated with `{keys: sharedSecretLeaks.join(", ")}` | ✅ Key exists in both locales; `{{keys}}` placeholder is present. |

- **C1 (dual-path parity):** Clear. The old submit-time check and new useMemo check the same conditions. Both `headersEntries` and `envEntries` are checked regardless of transport type (remote vs. stdio). No asymmetry.
- **C2 (control-flow ordering / nesting):** Clear. The `useMemo` runs on every render (deps: `headersEntries`, `envEntries`), producing a fresh result. No ordering concern — the warning is rendered declaratively, not gated by side-effects.
- **C3 (authorization boundary ≠ capability boundary):** N/A. This is a frontend advisory warning, not an authorization gate. The backend's non-owner blanking (§5.3) is the real authorization boundary and is not touched by this PR.
- **C4 (authorization lifecycle / container-member state cascade):** N/A. No auth/lifecycle changes in this PR.
- **C5 (build/note ≠ runtime correctness):** N/A. CSS and i18n changes are straightforward; no build-time/runtime divergence risk.
- **C6 (governance/policy/document self-consistency):** N/A. No policy or governance documents changed.

N/A — first review round for this PR.

## Human-verify

1. Confirm with the product team that replacing a blocking `window.confirm` with a non-blocking inline warning is the intended security trade-off. The auto-toggle and backend blanking mitigate the most common accidental-publish path, but the residual risk (operator explicitly toggles OFF, types/pastes a real secret, submits without noticing the warning) is now advisory-only. Not a merge blocker from the code-review perspective — flagging for product sign-off.

- `isSecretKey` at `utils/constants.ts:40` matches the same `SECRET_KEY_PATTERN` in both old and new code; no regex drift.
- Auto-toggle at `:297` fires only when `!row.userSupplied && !row.value && isSecretKey(k)` — a one-directional ratchet that prevents the most common accidental-publish scenario.
- JSON import seeding at `:1061` sets `userSupplied: isSecretKey(k)` for imported keys, so imported secrets default to user-supplied (safe).
- Both `headersEntries` and `envEntries` are checked by `sharedSecretLeaks` regardless of current transport type — consistent with the old code.
- No dangling references to the removed `sharedSecretConfirm` i18n key anywhere in the monorepo.
- Submit payload at `:831-834` is unchanged — no regression in what the backend receives.

## Verdict: APPROVED

Clean, well-scoped change with equivalent detection logic and a justified security trade-off. The P2 accessibility nit is non-blocking.

[Octo-Q] verdict: APPROVE — no P0/P1 found; detection logic is equivalent; security trade-off (blocking→advisory) is intentional and mitigated by auto-toggle + backend blanking. Only P2 is a minor accessibility gap.

yujiawei (approved):
# Code Review — PR #1006 (octo-web)

Replaces the blocking `window.confirm` shared-secret gate (added in #995) with a non-blocking inline advisory next to the Submit button. Reviewed at head `10257949a71f01faad2ea76163e5e29219b2fcd5`, three-dot against merge-base `a563a701`. Cross-checked with two independent review passes (one line-level, one whole-repo). `pnpm --filter @dmwork/mcp test` → 56/56 green; `pnpm --filter @octo/web build` → green.

## 1. Spec compliance

**Spec: ✅**

Scope is exactly the stated follow-up to #995, no more:

- **Missing:** none. `window.confirm` block removed from `handleSubmit` (`McpCreateModal.tsx:786`); inline advisory rendered between footer content and Submit (`McpCreateModal.tsx:1119`); derived `sharedSecretLeaks` memo added (`McpCreateModal.tsx:923`); CSS class added (`index.css:979`).
- **Over-build:** none. No new wire fields, no toggle-model change, no backend contract change.
- **Deviation:** none. The obsolete i18n key `sharedSecretConfirm` is fully removed from both `en-US.json` and `zh-CN.json` and replaced by a single `sharedSecretWarning` line; no dangling references remain (`grep` clean), and en/zh key parity holds (verified programmatically).

## 2. Code quality

**Quality: Approved** (no P0/P1; four P2 / advisory items below.)

### Security posture — the confirm→advisory downgrade is acceptable *here* (defense-in-depth note, not a blocker)

Both independent passes flagged the downgrade as high severity on the grounds that the server-side backstop is unverifiable from this repo. That is a correct observation about *this* tree, but the posture change is acceptable given the surrounding design:

- **The `confirm` was never the control that prevents consumer exposure.** Under the current wire contract, secret-shape *shared* values are persisted verbatim server-side whether or not the operator clicks OK; consumer-facing exposure is prevented by non-owner read-blanking at the API layer, not by this dialog. The dialog was operator-facing friction, not the security boundary.
- **The primary accidental-leak guard is untouched by this PR.** The KvEditor auto-toggles "user fills locally" ON the moment a secret-shape key is typed into a fresh row (`setKey`, `McpCreateModal.tsx:293-301`) — this is *not* in the diff. Reaching the warned state requires an explicit override (type the value before the key, or manually toggle a secret key OFF and keep a value). The advisory is the audit reminder for that deliberate-override case.

**Human-verify item (call-out for the security reviewer):** the sole remaining runtime defense for a deliberately-shared secret is server-side non-owner value-blanking. That guarantee lives outside this repo and cannot be exercised here. Please confirm it is deployed and covers *shared* (not only user-supplied) secret values across every read/list/export path before relying on this non-blocking UI. If that blanking is ever absent or regresses, a cleartext credential typed under an explicit override would be publicly disclosable with no client-side gate.

### P2 — `sharedSecretLeaks` detects a slightly different set than submission

`McpCreateModal.tsx:923`
```ts
for (const e of [...headersEntries, ...envEntries]) {
  const k = e.key.trim();
  if (!k) continue;
  if (e.userSupplied) continue;
  if (!e.value) continue;
  if (isSecretKey(k)) leaks.push(k);
}
```
The old submit-time check derived keys from `entriesToWire` output, which collapses duplicate keys last-write-wins and unions `userSupplied` membership across duplicates (`McpCreateModal.tsx:145-159`). The new memo iterates raw rows, so with duplicate keys (e.g. two `Authorization` rows, one OFF-with-value + one ON) it can (a) list the same key twice — `Authorization, Authorization will be published` — and (b) warn on a raw row whose value is not the final wire value. Traced direction of divergence: the memo **over-warns only** — it never misses a shared secret that submission would send — so it is not a correctness hole, but it undermines confidence in the warning. Recommend computing the warning list from `entriesToWire(...)` to match the wire state exactly.

### P2 — advisory not announced to assistive tech

`McpCreateModal.tsx:1119` — the warning `<span>` has no `role="alert"` / `aria-live`. When editing a field causes it to appear, screen-reader users get no announcement, and (unlike the removed `confirm`) nothing blocks submit. Since a modal dialog was the previous a11y-reachable signal, adding `role="alert"` (or `aria-live="polite"`) restores parity.

### P2 — long key lists are visually truncated; full text only in `title`

`index.css:979` uses `white-space: nowrap; overflow: hidden; text-overflow: ellipsis; max-width: 24em`, and both i18n strings interpolate `{{keys}}` at the *start* of the sentence (`"{{keys}} will be published in cleartext"` / `"{{keys}} 将以明文公开发布"`). With several long keys the operator may see only a truncated key list and lose the explanatory verb, with the complete text only in the native `title` tooltip (unavailable on touch, unreliable for keyboard/AT users). Consider allowing wrap, or moving `{{keys}}` to the end so the warning clause always shows.

### P2 (nit) — hard-coded color fallback

`index.css:981` `color: var(--wk-color-warning, #d97706);` — the literal fallback sidesteps the design-token system used elsewhere and has unverified dark-theme contrast. Minor; the token resolves in practice.

## 3. Overall verdict

**APPROVE** — Spec ✅ and no P0/P1 quality blockers. The four items above are non-blocking (one defense-in-depth call-out that is a human sign-off item, plus three a11y/UX/nit improvements). Given the `security_sensitive` classification, please action the human-verify item on server-side blanking before/at merge.

## 4. Suggestions

- Derive `sharedSecretLeaks` from `entriesToWire(headersEntries)`/`entriesToWire(envEntries)` so the warning matches exactly what is submitted (removes duplicate/stutter and stale-row cases).
- Add `role="alert"` to the advisory span.
- Reorder the i18n string so the fixed warning clause is never the part that gets ellipsed, or let the span wrap.

## 5. Additional notes

- No leftover `window.confirm` anywhere in the package after this change (grep clean).
- The auto-toggle "smart default ON for secret-shape keys" (the real first line of defense) remains in place and unmodified by this PR.

PR #1025, opened 2026-07-23
fix(e2e): run CI preview without pnpm wrapper

JianwenXu:
## Summary
- Run Playwright CI preview directly through the local Vite binary instead of `pnpm preview:e2e`
- Avoid pnpm wrapper failure preventing browser cases from executing
- Keep the e2e API target local/invalid to avoid leaking mocked runs to a real backend

## Verification
- `@C1`: 1 passed
- `@p0`: 29 passed
- JUnit: `tests=29`, `failures=0`, `skipped=0`, `errors=0`

PR #1013, opened 2026-07-23
ci(e2e): GitHub Actions e2e gate + @p0 tags

JianwenXu:
## Summary

Wire e2e as a real CI job on GitHub Actions:
- **`.github/workflows/e2e.yml`** — MR gate on PR to `main`. Builds a mock-mode bundle (`build:e2e`), starts `vite preview`, runs `@p0` cases in the official playwright container. junit post-processor splits business vs visual fail; business fail blocks the PR, visual fail allows (no `@visual` cases exist yet).
- **`.github/workflows/e2e-nightly.yml`** — `schedule` (03:00 UTC) + `workflow_dispatch`. Runs everything except `@visual`.
- **`docs/e2e-baseline-strategy.md`** — records the Phase 3 plan for visual baseline bootstrap + per-PR update workflows (not implemented; opens as a one-shot follow-up when visual cases are added).

Follows the e2e-kit v0.4 tag hierarchy (`docs/methodology/module-organization.md`): `@<CaseId> @<p0|p1|p2> @<module>`. All 29 existing `C*` cases tagged `@p0 @loop @<module>` (or `@p0 @search` for C989). `tests/bind/` and `tests/standalone-doc/` untouched — their spec ids don't follow the CaseId convention; retag separately.

## Notable fixes along the way

- `cross-env` not hoisted in this monorepo → `build:e2e` uses inline env vars. CI only runs on ubuntu; no Windows compatibility loss.
- e2e-kit v0.4 shipped `playwright.ci.config.ts` with `import.meta.url` / `fileURLToPath`, but this repo has no `"type": "module"`. Playwright's pirates transform loads configs as CJS → `ReferenceError: exports is not defined`. Switched to `__dirname` to match the existing `playwright.config.ts` (whose header already documented this constraint).

## Test plan

- [x] Local dry-run: `pnpm build:e2e` → `pnpm preview:e2e` on port 5199 → C1 passes in 8.6s
- [ ] GitHub Actions on this PR — first live run; expect to iterate on container / preview timeout / real-API-dependent cases
- [ ] Full 29-case sweep before considering merge

PR #1001, opened 2026-07-22
fix(search): close search modal on bot profile card "send message" (#989)

JianwenXu:
## Summary

Fixes #989.

在 IM 顶部搜索里搜到某个 bot → 打开资料卡 → 点「发送消息」跳转会话后，外层搜索弹窗仍留在页面上没关。

**Root cause**：`TabContacts` 里的 `BotDetailModal.onChat` 只关掉了资料卡自己（`setState({botDetailVisible: false})`），外层 `vm.showGlobalSearch` 无人切回 `false`。

**Fix**：`GlobalSearchPanel` 已经有 `hideModal` prop（原本给 content-tab 的 `handleLocate` 用），把它继续透传到 `TabContacts`；`BotDetailModal.onChat` 在跳转 + 关名片之后再调用一次 `hideModal()`，让 `Chat/index.tsx` 里包住整个搜索的外层 `WKModal` 一起关闭。

Non-bot contacts 走的是 `onClick("contacts")` → `handleGlobalSearchClick`，路径里已经调过 `hideModal`，本次改动不动它。

## E2E 覆盖 + Infra 投资

**Fix commit** (`715f7086`) — 4 行改动 + 3 条 source-guard 单测 (`tabContactsHideModal.test.ts`) 锁契约不被后续重构悄悄拆掉

**Infra commit** (`bf35456e`) — 顺带把 octo-web 缺失的 **IM/chat e2e 基础设施**建起来，让所有 chat 相关 e2e case 都能复用：
- kit `mock-im-wksdk` optional 装入 (`apps/web/e2e-kit/_kit/mock-im-runtime/`)
- Baseline 侵入 3 处（`VITE_E2E_MOCK_IM=1` gated，dev/prod tree-shake 干净）:
  - `apps/web/src/index.tsx` — enable mock IM hook + 暴露 WKSDK 给 spec
  - `packages/dmworkdatasource/src/module.ts` — 跳过真实 IM callback 注册
  - `packages/dmworkbase/src/App.tsx:connectIM` — 跳过真实 sdk.connect()
- chat baseline MSW handler（12 endpoint：`common/emojis`、`space/my`、`users/:uid/avatar`、`friend/sync`、`sidebar/sync`、`user/space/setting` 等）
- fixture auto-install empty seed on nav（SPA 全刷新兼容），预置 `octo:onboarding:seen=seen` 跳过 WebGL Strands

[...]

PR #1029, opened 2026-07-23
fix(e2e): fail on proxy leaks

JianwenXu:
## Summary
- Run Playwright CI preview directly through local Vite binary instead of the pnpm wrapper
- Mock the CI baseline endpoints that leaked to Vite proxy during e2e runs
- Skip emoji manifest network sync in e2e mock builds so startup does not race SW control
- Fail e2e workflows when Vite reports `http proxy error:` so leaked mocks cannot pass green

## Verification
- `pnpm -C apps/web build:e2e`
- `@C1 @p0`: 1 passed, proxy errors=0
- `@p0`: 29 passed, proxy errors=0

Note: #1025 PR ref stayed pinned to old head `828116f2` even after the fork branch moved to `32a06b77`, so this retry PR uses a fresh head branch.

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -68,10 +68,13 @@ jobs:
         # 组合语义: grep AND grep-invert (playwright 官方支持)
         shell: bash
         run: |
+          E2E_LOG="e2e-kit/playwright-report/e2e-nightly.log"
+          mkdir -p "$(dirname "$E2E_LOG")"
+
           set +e
           pnpm exec playwright test --config=e2e-kit/playwright.ci.config.ts \
-            --grep "@p[0-9]" --grep-invert "@visual"
-          PW_EXIT=$?
+            --grep "@p[0-9]" --grep-invert "@visual" 2>&1 | tee "$E2E_LOG"
+          PW_EXIT=${PIPESTATUS[0]}
           set -e
 
           JUNIT="e2e-kit/playwright-report/junit.xml"
@@ -80,6 +83,7 @@ jobs:
             exit 1
           fi
           TOTAL_TESTS=$(grep -oE 'tests="[0-9]+"' "$JUNIT" | head -1 | grep -oE '[0-9]+' || echo 0)
+          PROXY_ERRORS=$(grep -c "http proxy error:" "$E2E_LOG" || true)
           # nightly 也加 tests=0 guard: 防 grep drift 让 nightly 静默"全绿"
           EXPECTED_P0=29
           if [[ "$TOTAL_TESTS" -eq 0 ]]; then
@@ -87,6 +91,10 @@ jobs:
             exit 1
           elif [[ "$TOTAL_TESTS" -lt "$EXPECTED_P0" ]]; then
             echo "❌ nightly tests=$TOTAL_TESTS < EXPECTED_P0=$EXPECTED_P0 (有 case 掉 @p tag), fail"
+            exit 1
+          elif [[ "$PROXY_ERRORS" -gt 0 ]]; then
+            echo "❌ nightly 出现 $PROXY_ERRORS 条 Vite proxy error, 说明 mock 覆盖漏网, fail"
+            grep "http proxy error:" "$E2E_LOG" | sed -E 's/.*http proxy error: //g' | sort | uniq -c
             exit 1
           fi
           echo "✅ nightly ran $TOTAL_TESTS tests (playwright exit=$PW_EXIT)"
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -71,9 +71,12 @@ jobs:
         working-directory: apps/web
         shell: bash
         run: |
+          E2E_LOG="e2e-kit/playwright-report/e2e-p0.log"
+          mkdir -p "$(dirname "$E2E_LOG")"
+
           set +e
-          pnpm exec playwright test --config=e2e-kit/playwright.ci.config.ts --grep "@p0"
-          PW_EXIT=$?
+          pnpm exec playwright test --config=e2e-kit/playwright.ci.config.ts --grep "@p0" 2>&1 | tee "$E2E_LOG"
+          PW_EXIT=${PIPESTATUS[0]}
           set -e
 
           JUNIT="e2e-kit/playwright-report/junit.xml"
@@ -94,6 +97,7 @@ jobs:
           VISUAL_FAIL=$(echo "$FAILED_NAMES" | grep "@visual" | grep -vc "@p0" || true)
           BUSINESS_FAIL=$((TOTAL_FAIL - VISUAL_FAIL))
           TOTAL_TESTS=$(grep -oE 'tests="[0-9]+"' "$JUNIT" | head -1 | grep -oE '[0-9]+' || echo 0)
+          PROXY_ERRORS=$(grep -c "http proxy error:" "$E2E_LOG" || true)
           echo ""
           echo "═════════════════════════════════════════════════════════════"
           echo "📊 e2e @p0 summary (playwright exit=$PW_EXIT):"
@@ -101,6 +105,7 @@ jobs:
           echo "   total fail:    $TOTAL_FAIL"
           echo "   business fail: $BUSINESS_FAIL"
           echo "   visual fail:   $VISUAL_FAIL"
+          echo "   proxy errors:  $PROXY_ERRORS"
           if [[ "$TOTAL_FAIL" -gt 0 ]]; then
             echo ""
             echo "   fail case names:"
@@ -111,10 +116,11 @@ jobs:
           # 判定顺序 (fail-closed, PW_EXIT 检查在 visual allow 之前避免掩盖 crash):
           #  1. tests=0 → webServer / playwright 挂了根本没跑, block
           #  2. tests < EXPECTED_P0 → 有 case 静默掉了 @p0 tag / grep 漂了, block
-          #  3. 有业务 fail (含 @p0+@visual 交叉 case) → block
-          #  4. PW_EXIT!=0 但 junit 无 business fail → playwright 本身 crash/OOM/timeout, block
-          #  5. 仅视觉 fail (无 @p0 交叉) 且 playwright 干净退出 → allow
-          #  6. 全绿 → success
+          #  3. Vite proxy error → MSW 漏 mock, case 虽过也是假绿, block
+          #  4. 有业务 fail (含 @p0+@visual 交叉 case) → block
+          #  5. PW_EXIT!=0 但 junit 无 business fail → playwright 本身 crash/OOM/timeout, block
+          #  6. 仅视觉 fail (无 @p0 交叉) 且 playwright 干净退出 → allow
+          #  7. 全绿 → success
           EXPECTED_P0=29
           if [[ "$TOTAL_TESTS" -eq 0 ]]; then
             echo "❌ tests=0 (playwright 没跑起 test, 通常是 preview server timeout), block PR"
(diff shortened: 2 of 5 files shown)


lml2468 (changes requested):
## Review: #1029 — fix(e2e): fail on proxy leaks

**意图**:e2e 加固 —— CI preview 直连 vite binary(承 #1025)、mock 泄漏到 Vite proxy 的 baseline 端点补齐、e2e mock 构建跳过 emoji manifest 网络同步、**当 Vite 报 `http proxy error:` 时 fail workflow**(漏 mock 的假绿不再放过)。

锚定 head `32a06b77` / merge-base `3ce01400`(behind 3,0 冲突)。5 文件。

---

## Verdict: 🔴 REQUEST_CHANGES —— 一条 CI-gate blocker(shellcheck SC2086,与 #1013 同类),其余全部正确且优秀

### 🔴 BLOCKER:`e2e-nightly.yml` actionlint 失败(SC2086),会挂 required `Workflow Sanity` 检查

- **实跑 `actionlint`**:`e2e.yml` **exit 0 干净**,但 `e2e-nightly.yml` **exit 1** —— `SC2086:info` at 脚本内 `exit $PW_EXIT`(nightly 新增行,diff `+  exit $PW_EXIT`)。`$PW_EXIT` 未加引号。
- **为何是 blocker**:本 PR 改了 `.github/workflows/**`,`workflow-sanity.yml`(required 复用 workflow `Mininglamp-OSS/.github/...@v1`)**会在此 PR 触发** —— 正是 #1013 被 `sanity/actionlint` 的 shellcheck 发现卡红的**同一门、同一类**问题(#1013 是 SC2001)。
- **性质**:`$PW_EXIT` 恒为数字退出码,**非运行时 bug**;纯 lint 洁净度。但既然 sanity 是 required 且对 shellcheck 发现敏感,它会**让 CI 红、卡合并**。
- **修法(一字符)**:`exit "$PW_EXIT"`(加引号)。修完 actionlint 应 exit 0。
  - 顺带:e2e.yml 的 `exit $PW_EXIT`(若有)也应对齐;实核 e2e.yml 当前 actionlint 干净,仅 nightly 这条。

> 承 #1013 教训(我上轮在 #1013 就漏跑过 actionlint):**审改 workflow 的 PR 必须实跑 `actionlint` 且看真实 exit code**。本轮实跑抓到。

---

### 其余维度全部正确(修好上面即可合并)✅

**EmojiService.ts(生产文件)—— 严格 env-gated,无生产回归 ✅**
- `load()` 首行 `if (import.meta.env.VITE_E2E_MOCK === "1") return` —— **仅 e2e mock 构建跳过 manifest 拉取**;生产(flag 非 "1")照常 fire-and-forget 加载 + 兜底。与 #1001 建立的 `VITE_E2E_MOCK` 门一致,生产 tree-shake。✅

**proxy-leak 检测(gate 强化,优秀)✅**
- `... | tee "$E2E_LOG"` + `PW_EXIT=${PIPESTATUS[0]}` —— **正确用 PIPESTATUS 保留 playwright 真实退出码**(否则 tee 会掩盖成 0)。✅
- `PROXY_ERRORS=$(grep -c "http proxy error:" ...)`,`>0` → block + 打印去重清单。**fail-closed**:漏 mock 导致的假绿(case 过但打了真 proxy)现在被拦。判定顺序插在 business-fail 之后、visual-allow 之前,合理。✅
- e2e.yml(PR 门)+ nightly 都加了同款 gate。✅

**其余 ✅**:playwright.ci.config 直连 vite(承 #1025,逐 flag 等价)+ `VITE_API_URL=http://127.0.0.1:9`(discard port,防打真后端);chat-baseline 补漏 mock 端点;无 workflow 注入/secret、permissions 未放宽。

---

### 🟡 非阻断
- `exit $PW_EXIT` 之外,脚本里 `"$E2E_LOG"`/`"$PROXY_ERRORS"` 等已正确加引号 —— 仅这一处遗漏。
- Jerry-Xin 提的 PW_PREVIEW_PORT 数字校验(承 #1025)仍可选。

---

**结论**:proxy-leak fail-closed 门、EmojiService env-gate、vite 直连都做得干净正确,是有价值的 e2e 加固。唯一阻断是 `e2e-nightly.yml` 的 `exit $PW_EXIT` 未加引号(SC2086),会挂 required `Workflow Sanity`(#1013 同类)—— 一字符修复 `exit "$PW_EXIT"`。**REQUEST_CHANGES**,修完立即复审。

JianwenXu:
Superseded by #1030. This PR ref stayed pinned to 32a06b77 after the fork branch was pushed to 1121ef63, so the actionlint quote fix could not be reflected here.

Jerry-Xin (changes requested):
This PR hardens the e2e gate to fail on "proxy leaks" (requests escaping MSW that hit the Vite preview proxy → the unreachable discard target `http://127.0.0.1:9`, surfacing as `http proxy error:` in the teed log). The mechanism is sound and fail-closed. One blocking issue: the change to `PW_EXIT` trips shellcheck SC2086, which will fail the required `Workflow Sanity / actionlint` check.

🔴 Blocking

- `.github/workflows/e2e-nightly.yml:93` — `exit $PW_EXIT` now triggers shellcheck **SC2086** ("Double quote to prevent globbing and word splitting"). This is newly introduced by this PR: switching `PW_EXIT=$?` → `PW_EXIT=${PIPESTATUS[0]}` means shellcheck can no longer prove `PW_EXIT` is a bare integer, so it flags the unquoted expansion. The org `Workflow Sanity / actionlint` reusable check runs actionlint v1.7.12 with no severity relaxation and exits non-zero on any shellcheck finding (including info-level), so this required check will go red. Fix: `exit "$PW_EXIT"`. (`.github/workflows/e2e.yml` is clean — it only exits with literal `0`/`1`.) Verified locally: actionlint on the base branch is clean; actionlint on this head reports exactly this SC2086.

💬 Non-blocking

- `apps/web/e2e-kit/msw-handlers/chat-baseline.ts:36` — Several endpoints are registered in both a generic wildcard form (`*/common/emojis`) and an explicit `*/api/v1/...` form. The generic `*/...` already matches the `/api/v1/...` path, so the extra explicit handlers are likely redundant. Harmless (belt-and-suspenders), but worth a short comment or consolidation to avoid drift.
- The leak signal depends on the exact log substring `http proxy error:` emitted by Vite's proxy. A tiny self-test that intentionally leaks one request and asserts the guard fails would protect this string dependency across future Vite upgrades (Vite 8 currently emits it).

✅ Highlights

- Correct exit-status preservation through the `tee` pipe via `${PIPESTATUS[0]}`.
- The unreachable proxy target (`VITE_API_URL=http://127.0.0.1:9`) makes leaked requests deterministic — no flaky external dependency, and `vite preview` inherits `server.proxy` by default (`preview.proxy` default = `server.proxy`), so relative `/api/v1/*` leaks are routed to the discard port and logged.
- The proxy-error check is placed before the visual-only allow branch in the fail-closed decision chain, so a leak cannot be masked by the visual-fail exemption — it does `exit 1`, not just warn. Same guard mirrored in nightly.
- The `EmojiService.load()` early-return is narrowly gated on `VITE_E2E_MOCK === "1"` (set only by `build:e2e`), so production behavior is unchanged.
- `permissions: contents: read`, pinned node/pnpm/Playwright versions, and the `tests=0` / `< EXPECTED_P0` fail-closed guards from earlier PRs are all preserved; no workflow injection or secrets exposure.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1029 (octo-web)

## Summary

This PR adds MSW mock handlers for five API endpoints that were leaking through to the Vite dev proxy during e2e tests, introduces a CI gate to detect future proxy leaks, and fixes a subtle shell-exit-code bug in the workflow scripts. The EmojiService gains an early-return guard for e2e mock mode to prevent the emoji manifest call from firing at all. The webServer command switches from the `pnpm preview:e2e` wrapper to a direct `node node_modules/vite/bin/vite.js preview` invocation.

The core goal — plugging known proxy leaks and adding a detection gate — is well-motivated. Two P2 items below concern the detection mechanism's effectiveness in the current `vite preview` setup.

## Verification

- ✅ **PIPESTATUS fix is correct** — `PW_EXIT=${PIPESTATUS[0]}` (`.github/workflows/e2e.yml:78`, `:76`) properly captures playwright's exit code when piping through `tee`. The previous `$?` captured `tee`'s exit code (always 0 on success), masking playwright failures.
- ✅ **MSW handler URL patterns cover both API client paths** — The web branch in `apps/web/src/index.tsx:44` sets `apiURL = "/api/v1/"`, so browser requests resolve to `http://localhost:PORT/api/v1/...`. MSW wildcard patterns like `*/api/v1/common/emojis` and `*/common/emojis` both match. New handlers for `health`, `avatar_palette`, `message/channel/sync`, and `obo/grants` follow the same dual-pattern convention.
- ✅ **EmojiService early return is tree-shakeable** — `import.meta.env.VITE_E2E_MOCK === "1"` at `packages/dmworkbase/src/Service/EmojiService.ts:284` is statically replaced at build time. In `build:e2e` output it becomes `if ("1" === "1") return;` (always skips). In production builds (`VITE_E2E_MOCK` unset) it becomes `if (undefined === "1") return;` (never skips). Correct.
- ✅ **e2e-nightly.yml parity** — Same tee/PIPESTATUS/proxy-gate pattern applied consistently at `.github/workflows/e2e-nightly.yml:71-98`. Guard ordering matches e2e.yml.
- ✅ **Log directory creation** — `mkdir -p "$(dirname "$E2E_LOG")"` ensures the report directory exists before `tee`.
- ✅ **Guard ordering is fail-closed** — `tests=0` → `tests < EXPECTED_P0` → proxy errors → business fail → playwright crash → visual-only allow → success. Each guard blocks before the next is evaluated.

Static analysis only at head `32a06b77b6c7bfc820db083132323e6847a3a708`; build and tests not executed in this environment.

## Findings

### P2 — Proxy leak detection pattern is a dead guard (`.github/workflows/e2e.yml:119`)

`grep -c "http proxy error:" "$E2E_LOG"` can never produce a non-zero count in the current CI setup. The CI runs `vite preview` (Vite 8), which is a static file server — it does **not** activate `server.proxy` from `vite.config.ts`. The `"http proxy error:"` message is emitted by Vite's dev-server proxy middleware (`vite dev` only), so there is no code path in the CI that produces this string.

When MSW misses a handler, the flow is: browser sends request to `http://localhost:PORT/api/v1/...` (the hardcoded relative base from `apps/web/src/index.tsx:44`) → `vite preview` returns 404 or SPA fallback → no proxy error is logged. MSW's `onUnhandledRequest: "bypass"` at `apps/web/src/index.tsx:97` further ensures no MSW-level warning reaches the console.

The gate provides false confidence: a future MSW handler omission would pass this check silently. Consider switching to `onUnhandledRequest: "warn"` (or `"error"` for fail-fast) and grepping for MSW's warning pattern, e.g. `grep -c "intercepted a request without a matching handler"`.

### P2 — `VITE_API_URL: "http://127.0.0.1:9"` in webServer env has no effect (`apps/web/e2e-kit/playwright.ci.config.ts:67`)

This env var is set on the `vite preview` process, but neither consumer reads it at preview time:

1. **Browser side**: the web branch at `apps/web/src/index.tsx:44` hardcodes `apiURL = "/api/v1/"` for non-Tauri/Electron environments, ignoring `VITE_API_URL` entirely.
2. **Server side**: `vite preview` does not activate `server.proxy`, so `loadEnv()` in `vite.config.ts` reads the value into `apiOrigin` but the proxy config is never instantiated.

Harmless, but misleading — someone reading this config might believe it provides a safety net against real-backend requests when it actually does not. A comment explaining the intent (defense-in-depth for a future config change) or removal would close this.

- MSW `onUnhandledRequest: "bypass"` is pre-existing behavior, not changed by this PR. The new handlers correctly close the known leaks.
- `node node_modules/vite/bin/vite.js preview` bypassing pnpm is a reasonable choice to avoid pnpm wrapper overhead/signal handling in CI. The `preview:e2e` script still exists in `apps/web/package.json` for local use — not a blocker.
- The `build:e2e` script bakes `VITE_API_URL=http://mock.e2e.local` into the bundle at build time. Even if MSW is disabled, requests to `mock.e2e.local` DNS-fail rather than hitting a real backend.
- New MSW response shapes (`{ version: 0, list: [] }`, `{ size: 0, colors: [] }`, `{ messages: [] }`, `[]`) match the expected API contracts consumed by the app code.

## Verdict: APPROVED

The PR achieves its stated goals: known proxy leaks are plugged with targeted MSW handlers, the EmojiService skips its manifest call in e2e mode, and the PIPESTATUS bug is fixed. The two P2 items concern the detection mechanism's effectiveness — the gate won't catch future leaks in the current `vite preview` setup — but they don't introduce regressions or mask existing failures. Not a change I'd block on.





---

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Two P2 items (dead proxy-leak detection guard, no-op VITE_API_URL env) are non-blocking observations about detection mechanism effectiveness; the core MSW handler additions, PIPESTATUS fix, and EmojiService e2e guard are all correct.

PR #1020, opened 2026-07-23
refactor(dmworkbase): move channel webhook actions into bridge

katydids:
## Summary

Refactor ChannelWebhook data and action handling so the panel and edit modal no longer directly own runtime subscriber access or webhook management service calls.

## Related Issue

Related to #985

## Changes

- Move ChannelWebhook member candidate loading, subscriber sync, self fallback, and stale configured uid handling into `bridge/channelWebhook`
- Move create/update form state and webhook management actions into bridge hooks
- Keep UI-level confirm dialogs and Toast handling in `Components/ChannelWebhook`
- Add stale scope guards so old save/regenerate/delete requests do not update the UI after group/thread scope changes
- Add bridge tests for member option mapping, action execution, editor submit behavior, and stale request handling

## Architecture / Module Boundary

- Affected module(s): `dmworkbase` ChannelWebhook
- New or changed user-visible entry point: no
- Shared layer touched: `Components`, `bridge`, existing `Service` usage
- If shared code changed, impact scope: limited to ChannelWebhook panel/edit modal and its bridge hooks
- Duplicate entry point checked: yes; this keeps the existing ChannelWebhook entry only

## Testing

- [x] Unit tests added/updated
- [ ] Manually verified

Commands run:

```bash
./node_modules/.bin/vitest run packages/dmworkbase/src/Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx packages/dmworkbase/src/Service/__tests__/IncomingWebhook.test.ts packages/dmworkbase/src/Service/__tests__/IncomingWebhookService.test.ts packages/dmworkbase/src/bridge/channelWebhook

git diff --check origin/main...HEAD
```

Notes:

- `pnpm exec` is currently blocked locally by the packageManager/lockfile validation, so the tests were run through the checked-in local Vitest binary.
- This change does not add or change user-visible copy, so `pnpm i18n:check` was not required.

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [x] Added tests for my changes
- [x] Updated documentation
- [x] Ran `pnpm i18n:check` or confirmed the change does not affect UI copy
- [x] Confirmed the change follows module ownership and does not add duplicate user-visible entry points
- [x] Described impact scope when shared components, messages, bridge, or services are changed
- [x] Followed commit message conventions (Conventional Commits)

PR #1033, opened 2026-07-23
refactor(dmworkbase): move channel webhook actions into bridge

katydids:
## Summary

Refactor ChannelWebhook data and action handling so the panel and edit modal no longer directly own runtime subscriber access or webhook management service calls.

## Changes

- Move ChannelWebhook member candidate loading, subscriber sync, self fallback, and stale configured uid handling into `bridge/channelWebhook`
- Move create/update form state and webhook management actions into bridge hooks
- Keep UI-level confirm dialogs and Toast handling in `Components/ChannelWebhook`
- Add stale scope guards so old save/regenerate/delete requests do not update the UI after group/thread scope changes
- Decouple successful webhook mutations from best-effort list refresh so reload failures do not discard one-time token responses or misreport deletes
- Add bridge tests for member option mapping, action execution, editor submit behavior, stale request handling, and reload failure handling

- Affected module(s): `dmworkbase` ChannelWebhook
- Shared layer touched: `Components`, `bridge`, existing `Service` usage
- If shared code changed, impact scope: limited to ChannelWebhook panel/edit modal and its bridge hooks
- Duplicate entry point checked: yes; this keeps the existing ChannelWebhook entry only

## Testing

- [ ] Manually verified

Commands run:

```bash
./node_modules/.bin/vitest run packages/dmworkbase/src/Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx packages/dmworkbase/src/Service/__tests__/IncomingWebhook.test.ts packages/dmworkbase/src/Service/__tests__/IncomingWebhookService.test.ts packages/dmworkbase/src/bridge/channelWebhook

git diff --check upstream/main...HEAD
```

Notes:

- `pnpm exec` is currently blocked locally by packageManager/lockfile validation, so the tests were run through the checked-in local Vitest binary.
- This change does not add or change user-visible copy, so `pnpm i18n:check` was not required.

## Checklist

- [ ] Updated documentation

### DIFF
--- a/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
+++ b/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
@@ -1,116 +1,45 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { Channel, ChannelTypePerson, WKSDK } from "wukongimjssdk";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import { Channel } from "wukongimjssdk";
 import { Switch, Toast } from "@douyinfe/semi-ui";
 import { IconAlertTriangle } from "@douyinfe/semi-icons";
 import WKModal from "../WKModal";
 import WKButton from "../WKButton";
 import AiBadge from "../AiBadge";
-import WKApp from "../../App";
 import { useI18n } from "../../i18n";
 import { extractErrorMsg } from "../../Service/APIClient";
-import { subscriberDisplayName } from "../../Utils/displayName";
 import {
-    getImChannelInfo,
-    getImChannelSubscribers,
-    syncImChannelSubscribers,
-} from "../../im-runtime/channelRuntime";
-import {
-    buildWebhookUpsertReq,
-    IncomingWebhook,
-    IncomingWebhookCreateResp,
-    IncomingWebhookService,
-    isFlagOn,
-    MENTION_UID_MAX_LENGTH,
-    MENTION_UIDS_MAX,
-    normalizeMentionUids,
-    validateMentionUids,
+  IncomingWebhook,
+  IncomingWebhookCreateResp,
+  MENTION_UID_MAX_LENGTH,
+  MENTION_UIDS_MAX,
 } from "../../Service/IncomingWebhook";
+import { useChannelWebhookEditor } from "../../bridge/channelWebhook/useChannelWebhookEditor";
+import { useChannelWebhookMembers } from "../../bridge/channelWebhook/useChannelWebhookMembers";
+import type { ChannelWebhookMemberOption } from "../../bridge/channelWebhook/types";
 import "./index.css";
 
 export interface WebhookEditModalProps {
-    channel: Channel;
-    /** 管理员才渲染头像输入（普通成员传 avatar 服务端直接 400） */
-    isManager: boolean;
-    /** 编辑模式传入现有项；新增模式不传 */
-    webhook?: IncomingWebhook;
-    /** 子区作用域：传入即把创建/更新打到该子区面（#451）；群面不传。channel 始终为父群。 */
-    threadShortId?: string;
-    onClose: () => void;
-    /** 保存成功回调；创建成功时携带含一次性 token/URL 的响应 */
-    onSaved: (created?: IncomingWebhookCreateResp) => void;
+  channel: Channel;
+  /** 管理员才渲染头像输入（普通成员传 avatar 服务端直接 400） */
+  isManager: boolean;
+  /** 编辑模式传入现有项；新增模式不传 */
+  webhook?: IncomingWebhook;
+  /** 子区作用域：传入即把创建/更新打到该子区面（#451）；群面不传。channel 始终为父群。 */
+  threadShortId?: string;
+  onClose: () => void;
+  /** 保存成功回调；创建成功时携带含一次性 token/URL 的响应 */
+  onSaved: (created?: IncomingWebhookCreateResp) => void;
 }
 
 // API 契约里的字段长度上限（OpenAPI schema 常量，非动态配额）
 const NAME_MAX_LENGTH = 64;
 const AVATAR_MAX_LENGTH = 255;
-
-/** mention_uids 选择器的候选项（群成员 / bot）。 */
-interface MentionMemberOption {
-    uid: string;
-    name: string;
-    isBot: boolean;
-}
-
-/** WKSDK Subscriber 运行时用到的子集（SDK 未导出精确字段）。 */
-type GroupSubscriber = {
-    uid?: string;
-    name?: string | null;
-    remark?: string | null;
-    status?: number;
-    orgData?: {
-        robot?: number;
-        real_name?: string | null;
-        realname_verified?: boolean | number | string | null;
-    } | null;
-};
-
-/**
- * 判定群成员是否为 AI/bot。双信号取并集，最大化命中：
- * - `subscriber.orgData.robot`：随订阅数据返回，但不一定被富化（可能缺）；
- * - 该 uid 的 Person `channelInfo.orgData.robot`：WKAvatar / 会话列表等通用判定，
- *   依赖 channelInfo 缓存是否已热。
- * 任一命中即视为 AI；都拿不到则按非 AI（best-effort，与全局头像/徽章判定一致）。
- */
-function isBotMember(uid: string, sub: GroupSubscriber): boolean {
-    if (isFlagOn(sub.orgData?.robot)) return true;
-    try {
-        const info = getImChannelInfo(
-            WKSDK.shared(),
-            new Channel(uid, ChannelTypePerson)
-        ) as { orgData?: { robot?: unknown } } | null | undefined;
-        return isFlagOn(info?.orgData?.robot);
-    } catch {
-        return false;
-    }
-}
-
-/**
- * 从群成员缓存读取 mention_uids 选择器候选：按 uid 去重，展示名走
- * subscriberDisplayName（实名 > 备注 > 昵称），AI 由 {@link isBotMember} 判定。
- * 不按 status 过滤、也不排除自己（自己同样可被 webhook 自动 @），成员资格由服务端
- * 兜底；同步读缓存，调用方在 syncSubscribes 完成后再读一次以刷新。
- */
-function readGroupMemberOptions(channel: Channel): MentionMemberOption[] {
-    let subs: GroupSubscriber[];
-    try {
-        subs = getImChannelSubscribers(WKSDK.shared(), channel) as GroupSubscriber[];
-    } catch {
-        return [];
-    }
-    const out: MentionMemberOption[] = [];
-    const seen = new Set<string>();
-    for (const sub of subs || []) {
-        const uid = sub?.uid;
-        if (!uid || seen.has(uid)) continue;
-        seen.add(uid);
-        out.push({
-            uid,
-            name: subscriberDisplayName(sub) || uid,
-            isBot: isBotMember(uid, sub),
-        });
-    }
-    return out;
-}
 
 /**
  * 新建 / 编辑 webhook 弹窗。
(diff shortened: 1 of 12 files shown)


Jerry-Xin (dismissed):
Solid, well-structured refactor that moves ChannelWebhook data/action handling into `bridge/channelWebhook`. The #1020 reload-coupling blocker is fixed and covered by a test. One remaining stale-isolation gap on the rejection path blocks merge.

### ✅ Addresses prior #1020 blocker (verified)
The reload-coupling issue is resolved in `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts`:
- `refreshIfCurrentScope` runs reload as best-effort — `void Promise.resolve(params.reload()).catch(() => {})` — so a reload rejection can no longer contaminate an already-succeeded mutation.
- `regenerateWebhook` awaits the mutation first, captures `response`, then refreshes; returns `{ ok: true, response }` even if reload fails (no more discarded one-time token / double-rotation).
- `deleteWebhook` follows the same order; a reload failure no longer misreports a successful delete.
- Test present: `__tests__/useChannelWebhookActions.test.tsx` → "keeps regenerate and delete success when reload rejects" asserts both resolve `ok:true` (regenerate still returns the token) when `reload` rejects.

### 🔴 Blocking
- Stale-scope guards only cover the SUCCESS path; a rejection after a group/thread switch still surfaces in the old scope.
  - `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookEditor.ts` (`submit`, ~:130-155): the `try` checks `isCurrentRequest` only after a fulfilled create/update. If `runtime.create`/`update` rejects, the rejection propagates out of `submit()` with no scope check.
  - `packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx` (`handleSubmit` catch, ~:140): unconditionally `Toast.error(...)` — fires even when the user already switched scope mid-request.
  - `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts` (`regenerateWebhook`/`deleteWebhook`, ~:200/~:226): `await runtime.regenerate/delete` rejection propagates before any scope check.
  - `packages/dmworkbase/src/Components/ChannelWebhook/index.tsx` (`handleRegenerate`/`handleDelete` catch, ~:130/~:160): `Toast.error(...)` and `throw e`, which also keeps the obsolete confirmation dialog open in the wrong scope.
  - Fix: return a stale result (or check captured scope) on rejection too, so a scope-switched failure neither toasts nor rethrows in the old scope. Add rejection-after-scope-change tests to mirror the existing success/stale cases.

### 💬 Non-blocking
- The success-path stale handling is good; extending the same captured-scope check to the catch branches keeps the isolation contract symmetric and easy to reason about.

### ✅ Highlights
- Clean separation of UI, bridge state, and service calls.
- Successful mutations correctly decoupled from best-effort refresh failures.
- Focused tests for member option mapping, editor submit, stale-success, and reload-failure handling.
- Service path building uses `encodeURIComponent` for group/thread/webhook ids; no token/secret logging introduced.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1033 (octo-web)

## Summary

Refactors ChannelWebhook runtime logic out of the two UI components (`index.tsx`, `WebhookEditModal.tsx`) into four focused bridge hooks: `useChannelWebhookActions`, `useChannelWebhookEditor`, `useChannelWebhookMembers`, and a rewritten `useChannelWebhookList`. The refactor decouples successful mutations from best-effort list refresh, adds stale-scope guards so in-flight operations don't update the UI after the user switches group/thread, and introduces injectable runtime interfaces for testability. 1890 additions / 841 deletions across 11 files; 439 new lines of bridge tests.

## Verification

- ✅ **Stale scope guards** — `useChannelWebhookActions` captures `scopeKey` at action start, checks `isCurrentScope` (mounted + scope unchanged) before setting state or triggering reload. Tested explicitly in `useChannelWebhookActions.test.tsx` (scope-change → `{ok:false,reason:"stale"}`) and `useChannelWebhookEditor.test.tsx` (scope-change → `{ok:true,status:"stale"}`).
- ✅ **Mutation success decoupled from reload failure** — `refreshIfCurrentScope` wraps `params.reload()` in `Promise.resolve().catch(()=>{})` and a synchronous try/catch. `regenerateWebhook` and `deleteWebhook` return their success payload regardless of reload outcome. Locked by the "keeps regenerate and delete success when reload rejects" test.
- ✅ **Cooldown timer cleanup** — `useChannelWebhookActions` clears the cooldown timer on both unmount and scope change, preventing leaked timers after group/thread switching.
- ✅ **Auth/visibility preserved** — `canManageIncomingWebhook(item, { isManager, myUid })` at `index.tsx:203` still gates card actions. `canTestWebhook(item)` guard moved into bridge `testWebhook` (`useChannelWebhookActions.ts:163`), same logic. `isManager` prop passed through to `useChannelWebhookEditor` for avatar field gating.
- ✅ **IM runtime migration** — `getImChannelSubscribers(WKSDK.shared(), channel)` → `getCurrentImChannelSubscribers(channel)`. Verified `getCurrentImChannelSubscribers` exists in `im-runtime/currentChannelRuntime.ts:109` and delegates to the same underlying `getImChannelSubscribers` with `currentImChannelSubscribersRuntime()`. No contract change.
- ✅ **Creator names data flow** — `useChannelWebhookList` builds `creatorNames` from `getCurrentImChannelSubscribers(channel)` with the same uid→name mapping and self-fallback logic. No behavioral change.
- ✅ **Member options parity** — `readChannelWebhookMemberOptions` and `buildChannelWebhookMemberOptionsForSelect` produce the same deduped subscriber list with self-fallback and stale-configured-uid extras as the old inline `memberOptionsForSelect` memo. Verified in `useChannelWebhookMembers.test.ts`.
- ✅ **Submit stale guard** — `useChannelWebhookEditor.submit` checks `isCurrentRequest(requestSequence)` after each await; returns `{ok:true,status:"stale"}` if scope changed during the request, preventing the caller from firing `onSaved` and triggering a stale-scope reload.

Static analysis only at head `76c9cd09`; build and tests not executed in this environment.

## Findings

### Nit — Dead `WKSDK.shared()` mock (`__tests__/WebhookEditModal.test.tsx:111`)

The production code no longer imports `WKSDK` or calls `WKSDK.shared()` — both components now go through `getCurrentImChannelSubscribers` / `getCurrentImChannelInfo` from `currentChannelRuntime`, which is correctly mocked at `:109`. The `WKSDK: { shared: () => ({...}) }` block inside the `wukongimjssdk` mock (`:115–:128`) is dead code. Harmless since `importOriginal` still provides the real `Channel`/`ChannelTypeGroup` exports, but worth trimming to keep the test setup honest.

- **Regenerate one-time token flow**: `regenerateWebhook` returns `{ok:true,response}` only when scope is current; parent sets `urlResult` which triggers the URL modal. The `urlResult` state is cleared by the scope-change effect at `index.tsx:59–62`, so a scope change between mutation and response correctly suppresses the modal (which would have been cleared instantly anyway).
- **Delete toast gating**: `handleDelete` shows the "deleted" toast only when `result.ok` is true. On stale scope, the dialog has already closed (no error thrown), and the user has navigated away — no toast needed. Correct.
- **Toggle error propagation**: `updateStatus` rejections propagate through `toggleWebhook` to the caller's catch block (toast). Reload failures are silently swallowed by `refreshIfCurrentScope`. Same as old behavior.
- **Editor `buildWebhookUpsertReq` contract**: called with identical parameters as the old inline `handleSubmit`. `null` return (no changes) → `onClose()` via the `noop` status. Verified in bridge submit flow.
- **Test webhook cooldown**: bridge `testWebhook` preserves all four guards (disabled check via `canTestWebhook`, toggling check, in-flight check, cooldown check) and the 3-second cooldown timer with scope-aware cleanup.
- **Effect dependency arrays**: all bridge hooks have correct dependency arrays. `scopeKey` derivations are consistent within each hook.

## Verdict: APPROVED

Well-structured refactor that cleanly separates mutation, refresh, and UI concerns. The stale-scope guards and mutation/refresh decoupling are thoughtfully designed and well-tested. No correctness, security, or build-breaking issues found.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. Refactor preserves all runtime contracts (auth gating, cooldown, error propagation, data flow). Stale scope guards and reload-failure resilience are genuine improvements. One minor nit (dead mock in test).

yujiawei (dismissed):
# Code Review — PR #1033 (octo-web)

> Note: this PR advanced past the originally-dispatched commit `76c9cd09` while under review. This review is against the **current head `ec51c829`**, which adds one commit ("ignore stale channel webhook failures") on top of `76c9cd09`. That commit is the author's response to the prior blocking review, and it is assessed below.

## Summary

Structural refactor extracting ChannelWebhook runtime logic out of the two UI components (`Components/ChannelWebhook/index.tsx`, `WebhookEditModal.tsx`) into four focused bridge hooks under `bridge/channelWebhook/`: `useChannelWebhookActions`, `useChannelWebhookEditor`, `useChannelWebhookMembers`, and a rewritten `useChannelWebhookList`. It adds stale-scope guards so in-flight operations don't mutate the UI after a group/thread switch, decouples successful mutations from best-effort list refresh, migrates the IM subscriber accessor to the `current*` runtime, and adds injectable runtime interfaces with substantial new bridge tests.

## 1. Spec compliance

**Spec: PASS**

- Missing work: none. All four responsibilities named in the PR description (member/subscriber handling, action execution, editor submit, list/reload) are extracted, and the UI-level confirm dialogs + Toast handling correctly remain in the components.
- Extra work: none out of scope. The stale-scope guards, reload-failure decoupling, and injectable runtimes all serve the stated refactor + testability goal; no new user-facing features, flags, or fields.
- Deviation: none. The public entry point `ChannelWebhookPanel` keeps the same props (`channel`, `isManager`, `threadShortId`); its three consumers (`ThreadPanel`, `channelSettingThreadSections`, `channelSettingGroupInfoSection`) are unaffected.

The security-critical logic lives in `Service/IncomingWebhook.ts`, which is **not** in this diff — `canManageIncomingWebhook`, `canTestWebhook`, `buildWebhookUpsertReq`, `validateMentionUids`, and `isFlagOn` are all preserved verbatim. The refactor only moves *where* these pure functions are called from, not what they do.

## 2. Code quality

**Quality: Approved**

Behavioral parity verified against the pre-refactor code on the security-sensitive paths:

- **Auth / visibility gates preserved.** `canManageIncomingWebhook(item, { isManager, myUid })` still gates card actions (`index.tsx:264`). `canTestWebhook(item)` moved into bridge `testWebhook` (`useChannelWebhookActions.ts:169`) with identical enabled-only semantics — a disabled/deleted webhook still cannot be test-pushed. Avatar remains double-gated on `isManager`: the input only renders for managers (`WebhookEditModal.tsx:203`), and `buildWebhookUpsertReq` drops `avatar` for non-managers regardless.
- **In-flight + cooldown guards intact.** `toggleWebhook` keeps the `togglingId` guard; `testWebhook` keeps all four guards (canTest, same-id toggling, testing-in-flight, cooldown) and the 3s cooldown timer, now with scope-aware cleanup on unmount and scope change (`useChannelWebhookActions.ts:100-114, 182-192`) — this fixes a latent timer leak in the old component.
- **Mutation success decoupled from reload failure.** `refreshIfCurrentScope` wraps `reload()` in both a promise `.catch` and a synchronous try/catch; `regenerateWebhook`/`deleteWebhook` return their success result regardless of reload outcome. Locked by the "keeps regenerate and delete success when reload rejects" test. A genuine improvement over the old code, where a reload throwing after a successful regenerate could have discarded the one-time token.
- **Stale-scope guards now symmetric across success AND rejection (the `ec51c829` fix).** The prior blocking review noted the guards covered only the success path — a rejection after a group/thread switch would still `Toast.error` / rethrow in the old scope (and keep the obsolete confirm dialog open). `ec51c829` wraps `runtime.create`/`update` (`useChannelWebhookEditor.ts:141-166`) and `runtime.regenerate`/`delete` (`useChannelWebhookActions.ts:210-263`) so that a rejection after scope change returns a stale result instead of throwing. Because the bridge no longer throws on stale, the component catch branches (`handleSubmit`, `handleRegenerate`, `handleDelete`) no longer fire in the wrong scope — the wrong-scope toast and dead-modal concerns are resolved at the bridge layer without touching the components. New tests cover create/update/regenerate/delete rejection-after-scope-change.
- **IM runtime migration is contract-preserving.** `getImChannelSubscribers(WKSDK.shared(), channel)` → `getCurrentImChannelSubscribers(channel)`; the latter (`im-runtime/currentChannelRuntime.ts:109`) delegates to the same underlying `getImChannelSubscribers` with the ambient `currentImChannelSubscribersRuntime()`. Member-option building, self-fallback, stale-configured-uid extras, and creator-name mapping are all preserved.

Test coverage is solid: scope-change staleness (success + rejection), reload-rejection resilience, cooldown, member-option mapping, and editor submit staleness are all explicitly locked.

### Nit (non-blocking)

- **Dead `WKSDK.shared()` mock** in `Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx`. Production code no longer touches `WKSDK` — both components now go through `getCurrentImChannelSubscribers` / `getCurrentImChannelInfo`, which are mocked separately. The `WKSDK: { shared: () => ({...}) }` block is dead. Harmless (real `Channel`/`ChannelTypeGroup` still come from `importOriginal`), but worth trimming.

## 3. Verdict

**APPROVED**

Clean, well-tested refactor that preserves all runtime contracts (auth gating, cooldown, in-flight guards, request-body construction) and hardens several (cooldown-timer cleanup on scope change; mutation success surviving reload failure; symmetric stale handling on both success and rejection). The prior blocking review's concern is fully addressed by `ec51c829`. No P0/P1 survives verification.

### Advisory — worth a follow-up (non-blocking)

- **One-time token discarded on a scope-change race on the SUCCESS path (create & regenerate).** Distinct from the now-fixed rejection path: when `runtime.create`/`regenerate` *succeeds* but the group/thread scope changed while the request was in flight, `useChannelWebhookEditor.submit` returns `{ok:true,status:"stale"}` and `regenerateWebhook` returns `{ok:false,reason:"stale"}`, so the plaintext token/URL modal is suppressed. Because the token is only ever returned once (`list`/`update` never include it), this leaves an orphaned/unusable webhook (create) or a live integration whose rotated secret must be regenerated again (regenerate). This is a narrow race (navigate away within the sub-second the request is in flight), it is recoverable, and suppressing an out-of-context secret modal is a defensible choice — so it is not merge-blocking. But it is in tension with the PR's own "do not discard one-time token responses" goal, which is honored for the reload-failure and rejection paths but not the success-with-scope-change path. Consider surfacing the token even on scope change, or documenting the trade-off.
- **`testWebhook` returns `true` without a scope check** (`useChannelWebhookActions.ts:175-181`), so a "test sent" toast can appear after the user switched scope. This is **pre-existing** behavior (the old `handleTest` also toasted unconditionally), not a regression introduced here — noted only for completeness.

### Coverage / blind spots

- Build and unit tests were **not executed** in this review environment (static analysis at head `ec51c829` only). The bridge tests read as correct; a human may want to confirm CI is green on this head SHA.
- The security-critical pure functions (`canManageIncomingWebhook`, `canTestWebhook`, `buildWebhookUpsertReq`, `validateMentionUids`) live in `Service/IncomingWebhook.ts`, **not** in this diff and therefore unchanged — the refactor cannot have weakened them, but their internal correctness was assumed, not re-audited here.
- Runtime wiring of `currentImChannelSubscribersRuntime()` / `getCurrentImChannelInfo` (which channel context they resolve to at call time) was verified at the source level, not exercised at runtime.

Jerry-Xin (dismissed):
Re-review at ec51c829bcf3: the sole prior blocker (stale failed requests surfacing errors in the old scope) is fixed at all four sites, with new test coverage. Approving.

### ✅ Resolved (prior 🔴 — stale-scope-on-reject)
The reject/catch paths now guard the captured scope before surfacing any error, so a mutation that rejects after a group/thread switch no longer leaks into the old scope:

- `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookEditor.ts` submit — `runtime.update` and `runtime.create` are each wrapped; on reject, `!isCurrentRequest(requestSequence)` returns `{ ok: true, status: "stale" }` instead of throwing. A same-scope failure still rethrows.
- `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts` regenerate/delete — reject path checks `!isCurrentScope(actionScopeKey)` and returns `{ ok: false, reason: "stale" }` instead of throwing; `isCurrentScope` correctly added to the `useCallback` deps.
- `packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx:126` handleSubmit — now sees `status === "stale"` and early-returns; only a same-scope throw reaches the `catch` → `Toast.error`.
- `packages/dmworkbase/src/Components/ChannelWebhook/index.tsx:130,158` handleRegenerate/handleDelete — a stale reject returns `ok:false`/`reason:"stale"` without throwing, so `wkConfirm` closes normally (no stale Toast, no stuck confirm modal in the wrong scope). Same-scope throw still surfaces the error and rethrows to keep the dialog open for retry.

No over-swallowing: when the scope still matches, all four sites rethrow, so a legitimate same-scope failure is still shown to the user. The stale branches return a distinct shape (`status:"stale"` / `reason:"stale"`), never a success-shaped result on a real failure.

### ✅ Highlights
- Test coverage for the reject-after-scope-switch path added at both hook levels: `useChannelWebhookEditor.test.tsx` ("returns stale when a create/update request rejects after scope changes") and `useChannelWebhookActions.test.tsx` ("marks regenerate and delete rejections stale after scope changes"). All assert the stale result and that `reload` is not called.
- #1020 reload-coupling fix intact: `refreshIfCurrentScope` still runs reload best-effort (`void Promise.resolve(params.reload()).catch(() => {})`) and returns true, decoupling mutation success from reload outcome.
- Webhook security unchanged and intact: `encodeURIComponent` on group/webhook/thread ids in the Service, plaintext token returned only once (create/regenerate), no token logging, manager gate preserved. Service layer is untouched by this delta.

### Delta reviewed
One new commit on top of the prior head (ahead 1 / behind 0, merge-base 76c9cd09898b — not a rebase): "fix(dmworkbase): ignore stale channel webhook failures". Four files: the two hooks plus their tests. CI green, mergeable.

cnwenf (changes requested):
# Consolidated Code Review — PR #1033 (octo-web)

> **Consolidated from three independent reviews** (octo-review-glm5.2, octo-review-qwen3.7, octo-review-qwen3.8). Findings below are the deduplicated union; each is tagged with its source reviewer(s) and original severity. All three reviews were conducted independently against the **same frozen head `76c9cd09898b037bb85828ee0b79493ba0856623`** (base `c1263f5e`).

## Consolidated Verdict: CHANGES_REQUESTED

2 of 3 reviewers request changes (glm5.2: `NEEDS_CHANGES`, qwen3.7: `CHANGES_REQUESTED`); 1 reviewer is comment-only (qwen3.8: `COMMENT`). All three agree the refactor is **high-quality and behavior-preserving on every happy path**, and **security is CLEARED with 0 findings by all three**. No reviewer found a happy-path P1 blocker. The substantive issues are: (1) the stale-scope result unions conflate "side effect already happened server-side" with "operation failed" — silently dropping freshly rotated one-time tokens and misreporting successful stale mutations; (2) the PR's stated "decouple reload failures from mutation success" contract is not honored in production wiring (an error toast still fires alongside the success toast); (3) a members-effect dependency array that can spam `syncSubscribers` network calls; plus (4) broad test-coverage gaps across all four new hooks and (5) a CI gap — unit tests are never executed in CI, so every behavior this PR claims to test can regress green.

| Reviewer | Verdict | Headline |
|----------|---------|----------|
| octo-review-glm5.2 | NEEDS_CHANGES | 2 P2 + 11 P3 confirmed + 1 P3 blocked (structural) |
| octo-review-qwen3.7 | CHANGES_REQUESTED | 1 P1 + 7 P2 + 4 P3 |
| octo-review-qwen3.8 | COMMENT | 4 P2 + 3 P3 (test-confirmed) + 9 static coverage observations |

> ⚠️ **Follow-up commit not covered by this round.** A second commit `ec51c829` — *"fix(dmworkbase): ignore stale channel webhook failures"* — was pushed after the review base and modifies exactly `useChannelWebhookActions.ts` (+27/−10) and `useChannelWebhookEditor.ts` (+23/−8) plus their tests. It appears to target the stale-scope findings (A1/A2 below) but was **not** reviewed in this round (all three reviews froze at `76c9cd09`). Please re-verify the stale-scope findings against `ec51c829`; if it resolves them, items A1/A2 and the stale-scope halves of A3 can be considered addressed.

## Verification (union across reviewers, all at head `76c9cd09`)

- ✅ **Unit tests pass** — scoped `vitest run` over the PR test files: 87/87 (qwen3.8), 78/78 (qwen3.7), 15/15 (glm5.2), all exit 0 in isolated checkouts at the exact head SHA.
- ✅ **Typecheck base/head control** (qwen3.8) — head introduces zero new error categories vs base; no `TS2307`/`TS2554`/`TS2339`/`TS2304` for the new modules.
- ✅ **Mutation protocol** — all three ran focused mutation probes; multiple coverage gaps were mutation-confirmed (removing a guard left all PR tests green). Zero behavioral bugs were found on the *probed* paths on HEAD.
- ✅ **Default runtime wiring probe** (qwen3.7) — 4/4 passed: the bridge hooks delegate to `IncomingWebhookService` with correct arg ordering matching the old inline calls. (Note: this was an ad-hoc probe — see A4, no *committed* test pins it.)
- ⚠️ **E2E** — the repo ships no compose stack (static-build Dockerfile only), so a full service E2E is not applicable to this client-only change; reviewers used focused probes (vitest + mutation) instead. glm5.2: PASS (probe-level); qwen3.7: PASS; qwen3.8: PARTIAL.
- ✅ **Security** — CLEARED, 0 findings, by all three (token handling, auth gates, path-param injection, XSS/SSRF sinks, credential scan).

## Findings

### A. Correctness — stale-scope semantics (the core issue)

#### A1 — [P2] Stale-scope `regenerate`/`create` silently destroys the freshly rotated one-time token
`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:218` · `useChannelWebhookEditor.ts:158` · consumers `Components/ChannelWebhook/index.tsx:131`, `WebhookEditModal.tsx:126`
**Source: qwen3.8 (P2, vitest-confirmed) + glm5.2 (P3 `dataflow-003`, probe-confirmed).** REGRESSION vs `pr-base`. `regenerateWebhook` completes the server mutation (old token invalidated, new plaintext token/url returned — "仅此一次返回"), but on a mid-flight scope change it returns `{ok:false,reason:'stale'}` with **no `response` field**. The panel gates `setUrlResult(result.response)` inside `if (result.ok)` with no else/toast, so the confirm dialog closes silently: the URL is rotated server-side, the old URL is now dead, the new token is never shown, and it is unrecoverable from the list (`IncomingWebhook` carries no token). `pr-base` called `setUrlResult(resp)` unconditionally. Same shape for `create`: the unmount cleanup bumps `requestSequenceRef`, so a create resolving after the editor is dismissed (Esc — `WKModal` `closeOnEsc` defaults true) or after a scope change is classed `stale` and returns `{ok:true,status:'stale'}` **without `created`**; the modal maps `"stale"` to a bare `return`, so `onSaved(created)` never fires and the token never reaches `WebhookUrlModal` (lower severity than regenerate — the webhook still appears in the list and is recoverable via regenerate). This contradicts the PR's own "do not discard one-time token responses" guarantee, which the code honors only on the reload-failure path, not the stale-scope path.
**Fix:** when a stale regenerate/create has already mutated server state, surface the returned token (or at least Toast that the URL was rotated and must be re-viewed) instead of dropping `response`/`created`; e.g. a scope-independent channel (ref/global callback) before the stale-guard short-circuits.

#### A2 — [P3] A server-successful stale `delete` is reported as failed and surfaced as nothing; stale conventions are inconsistent across the four actions
`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:240` (delete) · `:181` (test)
**Source: qwen3.8 (P3, vitest-confirmed).** A delete that succeeds server-side on a stale scope returns `{ok:false,reason:'stale'}`, and the panel gates the success Toast on `result.ok`, so a confirmed destructive delete completes with zero feedback; meanwhile stale `testWebhook` returns `true` unconditionally, so a stale test still toasts success — the four actions use **three different stale conventions**. `pr-base` toasted success and reloaded unconditionally after the mutation. Rare trigger (scope switch mid-flight).
**Fix:** distinguish "attempt rejected before any side effect" from "side effect happened but scope is stale" in the result union, and report the actual outcome.

#### A3 — [P2] Broken contract: reload-failure decoupling — an error toast still fires alongside the success toast
`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:126` · `useChannelWebhookList.ts` (`load`)
**Source: glm5.2 (P2 `contract-reload-misreport`, probe-confirmed) + qwen3.7 (P2).** The PR claims to "Decouple successful webhook mutations from best-effort list refresh so reload failures do not misreport deletes," but the real `useChannelWebhookList.load` catches its own list-fetch errors and calls `onLoadError` (`Toast.error`) **before resolving — it never rejects**. So `refreshIfCurrentScope`'s swallowed `.catch` (L126) is dead code for the production reload path, and a reload failure after a successful delete/regenerate/toggle still surfaces an error toast alongside the success toast — the exact misreport the PR claims to fix. A probe confirmed: after `deleteWebhook` returned `{ok:true}`, `onLoadError` fired once while the swallowed `.catch` was a no-op. *(qwen3.8 notes the flip side: the successful mutation's token itself IS kept when the reload fails — so the success result is not discarded; the disagreement is only whether the residual error-toast constitutes the claimed decoupling. The factual mechanism — `load` catches + `onLoadError`→Toast + resolves, `.catch` dead — is agreed by all three.)*
**Fix:** suppress `onLoadError` during a post-mutation refresh (gate the error toast on a mutation-in-flight flag), make `load` re-throw after `onLoadError`, or route reload errors through a silent channel distinct from load-time errors.

#### A4 — [P2] The production wiring is never executed by any committed test
`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:48` · `Components/ChannelWebhook/index.tsx:131`
**Source: qwen3.8 (P2).** Every hook test injects a mock runtime, and no committed test renders `ChannelWebhookPanel` — the sole consumer of `defaultActionsRuntime`/`defaultEditorRuntime`/`defaultMemberRuntime`. So the `enabled → IncomingWebhookStatus.enabled/disabled` mapping and the arg forwarding to `IncomingWebhookService` have zero committed coverage: an inverted status ternary or a swapped `groupNo`/`webhookId` would toggle/delete the wrong webhook and every test stays green. *(qwen3.7's ad-hoc default-runtime wiring probe passed 4/4 on HEAD — so the wiring is correct today — but nothing committed pins it against regression.)*
**Fix:** add at least one panel-level test that renders `ChannelWebhookPanel` with a spy runtime and asserts the default runtime is invoked with the correct args.

### B. Correctness — members effect

#### B1 — [P2] Members effect re-fires `syncSubscribers` on `Channel` object-identity change
`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookMembers.ts:165`
**Source: glm5.2 (P2 `dataflow-001`, probe-confirmed) + qwen3.7 (P2).** The `useEffect` dependency array includes `params.channel` (object reference) alongside `scopeKey` (string). If the parent constructs a new `Channel` object on each render with the same `channelID`, `scopeKey` stays identical but the reference changes, so the effect re-fires on every render — each firing triggers `runtime.syncSubscribers(params.channel)`, a network call. `scopeKey` alone already captures the real scope transition; the `Channel` dependency is redundant and fragile to caller memoization discipline. The parent panel does not memoize the `Channel` object, so the path is reachable; a probe confirmed re-rendering with `new Channel('g1', …)` (same `channelID`/`scopeKey`) increased `syncSubscribers.mock.calls.length`. Diff-introduced — the old code in `WebhookEditModal` used `[channel]` alone (qwen3.7).
**Fix:** drop `params.channel` from the dependency array and rely solely on `scopeKey`, or memoize the `Channel` object in the parent.

#### B2 — [P3] No sync-request cancellation on scope change/unmount
`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookMembers.ts:146`
**Source: glm5.2 (P3 `dataflow-002`) + qwen3.7 (P3).** The sync promise has no cancellation primitive (no `AbortController`/dedup/debounce; the runtime signature takes only `channel`). When the effect re-fires, a new sync starts while the previous is in flight; the `requestSequenceRef` guard discards stale *state updates* but each in-flight sync still completes its full network round-trip. Rapid channel switching → N concurrent sync requests server-side.
**Fix:** thread an `AbortSignal` into `runtime.syncSubscribers` and abort the previous signal on re-fire/unmount.

### C. Test coverage & CI

#### C1 — [P1/P2] CI never runs any of the new tests
`.github/workflows/ci.yml` (and all 20 workflow files)
**Source: qwen3.7 (P1) + qwen3.8 (P2) + glm5.2 (P3).** Confirmed independently by all three (structured parse / grep of all 20 workflows, 51 steps): **zero** steps invoke `vitest` / `pnpm test` / the turbo `test` task — even though the root defines `test = turbo run test`, `packages/dmworkbase/package.json` declares `"test": "vitest run"`, and this PR adds 4 test files (438 lines). Any regression in this refactor merges green; the injected runtime mocks aren't even type-pinned in CI.
**Fix:** wire `pnpm test` (or `turbo run test --filter @octo/dmworkbase`) into the CI build job. *(Severity note: qwen3.7 rates this P1; glm5.2/qwen3.8 rate it P2 — presented at the highest-assessed severity.)*

#### C2 — [P2] Coverage gaps across all four new hooks (mutation-confirmed where noted)
**Source: qwen3.7 (P2 cluster) + glm5.2 (P3 cluster, mutation-confirmed) + qwen3.8 (static observations).** The PR ships 4 test files but leaves large parts of the new hook layer unexercised; several gaps were mutation-confirmed (disabling the guard left all PR tests green):
- **`useChannelWebhookList` has no test file** (`useChannelWebhookList.ts:42`, +95/−71) — the `requestSequence` out-of-order discard, error→`onLoadError`→Toast path, reload(showLoading) reset, and creator-name resolution are all untested. *Mutation-confirmed (glm5.2, qwen3.7): disabling the sequence check lets stale items overwrite the current scope while all tests stay green.* Also no im-runtime migration replay/parity test (qwen3.8).
- **Editor tests cover 1 of 6 variants** (`__tests__/useChannelWebhookEditor.test.tsx:91`) — only the `stale` variant on the create path. Happy-path create (`'created'` return, L161), happy-path update, noop, validation failure, runtime rejection, and scopeKey reset are untested. *Mutation-confirmed: changing `'created'`→`'updated'` keeps all tests green.*
- **Editor `busy` guard + validation path untested** (`useChannelWebhookEditor.ts:119`) — the `if (saving) return {ok:true,status:'busy'}` guard and `validateMentionUids` failure path (`tooMany`/`tooLong`) have no reproducer. *Mutation-confirmed: removing the busy guard leaves all tests green.*
- **Members hook effect untested** (`useChannelWebhookMembers.ts:138`) — the test exercises only the two pure helpers (`readChannelWebhookMemberOptions`, `buildChannelWebhookMemberOptionsForSelect`); it never renders the hook (`renderHook`). The async sync flow, `requestSequence` stale-discard guard, and silent sync-failure `.catch` are unreached. *Mutation-confirmed: removing the stale guard keeps tests green.* Also untested: the `getSubscribers` catch-and-return-`[]` fallback (qwen3.7) and member-options throw-fallback + `isBotMember` channelInfo fallback (qwen3.8).
- **Actions tests miss most guards / stale / timer / error paths** (`__tests__/useChannelWebhookActions.test.tsx:98`) — five guard clauses silently skip work but only cooldown expiry is tested. Untested: double-toggle prevention, `canTestWebhook` rejection, test-during-toggle/test guards, stale-scope for toggle/test (asymmetric vs regenerate/delete), cooldown re-entry block (the block guard at L171 — removing it leaves the test green), timer cleanup on scope-change/unmount, the synchronous reload-throw path, and all four actions' failure/rejection paths (every mock resolves; no rejected-runtime case exercises spinner recovery in `finally` or error propagation).
- **Editor form-reset on scope/webhook switch untested** (`useChannelWebhookEditor.ts:93`, qwen3.8 static) — stale `name`/`mention_uids` could be submitted against a different `webhook_id`. *(Static reachability observation; not mutation-verified this round.)*
- **Weak modal assertion** (`WebhookEditModal.test.tsx:265`, qwen3.8) — the "modal just closes" case never asserts `onClose`.

**Fix:** add the missing unit tests (list stale-discard + error + reload; editor created/updated/busy/validation; members `renderHook` sync-after-scope-change; actions guard/stale/rejection/timer-cleanup cases) — in this PR or a fast-follow.

### D. Smaller correctness nits (P3)

#### D1 — [P3] Reload-rejects test mock does not match production `load`
`__tests__/useChannelWebhookActions.test.tsx:159` — **glm5.2 + qwen3.7.** The "keeps regenerate and delete success when reload rejects" test mocks `reload` as a directly-rejecting function, but the real `load` catches internally + `onLoadError` + resolves (never rethrows). The test exercises the dead `.catch()` path; a regression re-introducing the post-mutation error toast (or stopping `onLoadError`) would still pass. Fix: mock `reload` with the real `load()` shape.

#### D2 — [P3] Orphan error toast from unmounted editor submit
`useChannelWebhookEditor.ts:162` — **glm5.2 (`dataflow-005`, probe-confirmed).** When `submit()` is in-flight and the component unmounts (scope change removes `WebhookEditModal` via `setEditTarget(null)` at `index.tsx:93-96`), the `finally` skips `setSaving(false)` (isCurrentRequest false) but a rejection still propagates to `handleSubmit`'s catch (`WebhookEditModal.tsx:140-151`), firing `Toast.error` from the now-unmounted component. Not a crash, but a confusing orphan toast. Fix: mounted guard before `Toast.error`, or an `AbortController` suppressing the rejection on unmount.

#### D3 — [P3] Cooldown bypass via scope round-trip
`useChannelWebhookActions.ts:113` — **glm5.2 (`dataflow-006`).** The 3s cooldown timer is cleared on any `scopeKey` change; navigate away and back within 3s and the same webhook can be tested again immediately (cooldown is scope-lifecycle-local, not keyed off `webhook_id`). Timer teardown on unmount itself is correct (L104). Fix: key the cooldown off `webhook_id` (a ref map).

#### D4 — [P3] Spurious list reload on language switch
`useChannelWebhookList.ts:72` — **glm5.2 (`dataflow-007`) + qwen3.7.** `load`'s dependency array includes `onLoadError`, which chains through the i18n `t` function; a runtime language change gives `handleLoadError` → `load` a new reference, re-firing the data-fetch effect and causing one extra webhook-list reload per language switch. Fix: store `onLoadError` in a ref and read it inside `load`.

#### D5 — [P3] Structural TOCTOU window in `scopeKeyRef` update *(warning — not machine-confirmed)*
`useChannelWebhookActions.ts:109` · `useChannelWebhookEditor.ts:79` — **glm5.2 (`dataflow-004`).** `scopeKeyRef.current` / `requestSequenceRef` are updated inside a passive `useEffect` (post-commit). Between the render producing callbacks with the new `scopeKey` and the effect updating the ref, `isCurrentScope(newScopeKey)` returns false (ref still holds the old value): for `toggleWebhook` the reload is suppressed and `togglingId` isn't cleared in `finally`; for `testWebhook` the cooldown isn't started. The window is machine-evident structurally but could not be stably triggered under `act()` batching — blocked, not confirmed. Fix: update `scopeKeyRef` synchronously during render (`useLayoutEffect` or ref-assign in the callback factory).

## Things all three checked and found fine

- **Security — CLEARED, 0 findings (all three):** auth gating (`canManageIncomingWebhook(item,{isManager,myUid})`, `canTestWebhook`, `isManager` avatar gate) is line-for-line identical to `pr-base` and the server remains authoritative; the one-time token is never logged or persisted (no `console`/`innerHTML`/`dangerouslySetInnerHTML` on added lines; it flows only into React state → a fixed-length masked modal); `groupNo`/`threadShortId`/`webhookId` are interpolated only inside `IncomingWebhookService.basePath`, which `encodeURIComponent`s them; no XSS/eval/SSRF sinks; credential scan clean (all `token:` literals are test mocks). The stale-scope `useEffect([scopeKey])` clearing `urlResult` actually *tightens* token exposure vs base. `scopeKey` is a UI-only staleness guard, not a substitute for a server-scope check.
- **Happy paths are behavior-preserving:** all runtime contracts hold on HEAD; default runtime delegation arg ordering matches the old inline service calls exactly (qwen3.7 wiring probe 4/4).
- **Subscriber-call migration** `getImChannelSubscribers(WKSDK.shared(), channel)` → `getCurrentImChannelSubscribers(channel)` is a pure wrapper binding `WKSDK.shared()` internally, still operating on the passed `channel` (not a global active channel) — behavior-identical (qwen3.8).
- **Cooldown timer teardown** on unmount is verified (`clearTimeout` at L104) — no orphaned timers.
- **Member-option mapping parity** (dedup / self-fallback / stale configured-uid extras), `validateMentionUids` reason mapping (`tooMany`/`tooLong`), and `buildWebhookUpsertReq` are behavior-preserving.
- **Discriminated-union result types** correctly enforce exhaustive handling at call sites; `requestSequence` stale-discard guards are present in both `useChannelWebhookList` and `useChannelWebhookMembers` (qwen3.7).

---

*This is an automated consolidated review from the octo-review squad (3 independent reviewers, deduplicated union). Review base: `c1263f5e`, head: `76c9cd09`. The follow-up commit `ec51c829` was not part of this review round — see note at top.*

PR #1024, opened 2026-07-23
feat(channel-setting): refactor information management panels

Candice00711:
## Summary

Refactor the existing group and thread information settings panels into focused feature sections and props-only UI rows while preserving the current entry points, permissions, actions, and runtime behavior. This also improves the approved panel layout and adds cached pinyin search to the two group-member search flows.

## Related Issue

N/A — scoped channel-setting structure and UI upgrade.

## Changes

- Split group profile, group management, nickname, preference, danger, thread info, GROUP.md, webhook, and thread action sections into focused feature files.
- Add reusable props-only `ChannelSettingRows` UI components with Storybook coverage.
- Keep existing `bridge/channelSetting/*Actions` and runtime paths unchanged while preserving permission and hidden-entry behavior.
- Expand the settings panel, balance the five-column member layout, and add scoped spacing to group name, notice, nickname, and remark editors.
- Add cached Chinese/full-pinyin member search for "View more members" and "Transfer ownership", including stale-request protection and existing transfer filters.
- Refresh the displayed group nickname after a successful save without changing the underlying action contract.

## Architecture / Module Boundary

- Affected module(s): `dmworkbase` channel settings and the shared subscriber list compatibility component.
- New or changed user-visible entry point: no; all existing group/thread information and settings entries are preserved.
- Shared layer touched: `ui`, `Components`, and feature orchestration. No Service, Messages, Chat, Conversation, WKSDK, datasource, or channel-setting Action changes.
- If shared code changed, impact scope: `SubscriberList` gains an optional injected local-search callback; callers without it retain the existing server search and pagination behavior.
- Duplicate entry point checked: yes; no new menu, route, or button entry was added.

## Testing

- [x] Unit tests added/updated
- [x] Manually verified
- `pnpm --dir packages/dmworkbase test` — 230 files, 2027 tests passed.
- `pnpm --dir apps/web build` — passed after rebasing onto current `upstream/main`.
- `pnpm i18n:check` — passed.
- `git diff --check` — passed.
- Manually verified panel width, member grid spacing, edit-field clipping/spacing, Chinese and pinyin member search, transfer-owner filtering, and existing permission paths.
- Storybook reached full module transformation in the pre-rebase check, then exited with the existing environment-level code 132; no component compile error was reported.

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [x] Added tests for my changes
- [x] Updated documentation (not required; no public API or setup behavior changed)
[...]

PR #1038, opened 2026-07-23
feat(channel-setting): refactor information management panels

Candice00711:
## Summary

Refactor the existing group and thread information settings panels into focused feature sections and props-only UI rows while preserving the current entry points, permissions, actions, and runtime behavior. This also improves the approved panel layout and adds cached pinyin search to the two group-member search flows.

## Related Issue

N/A — scoped channel-setting structure and UI upgrade.

## Changes

- Split group profile, group management, nickname, preference, danger, thread info, GROUP.md, webhook, and thread action sections into focused feature files.
- Add reusable props-only `ChannelSettingRows` UI components with Storybook coverage.
- Keep existing `bridge/channelSetting/*Actions` and runtime paths unchanged while preserving permission and hidden-entry behavior.
- Expand the settings panel, balance the five-column member layout, and add scoped spacing to group name, notice, nickname, and remark editors.
- Add cached Chinese/full-pinyin member search for "View more members" and "Transfer ownership", including stale-request protection and existing transfer filters.
- Refresh the displayed group nickname after a successful save without changing the underlying action contract.

- Affected module(s): `dmworkbase` channel settings and the shared subscriber list compatibility component.
- New or changed user-visible entry point: no; all existing group/thread information and settings entries are preserved.
- Shared layer touched: `ui`, `Components`, and feature orchestration. No Service, Messages, Chat, Conversation, WKSDK, datasource, or channel-setting Action changes.
- If shared code changed, impact scope: `SubscriberList` gains an optional injected local-search callback; callers without it retain the existing server search and pagination behavior.
- Duplicate entry point checked: yes; no new menu, route, or button entry was added.

## Testing

- [x] Manually verified
- `pnpm --dir packages/dmworkbase test` — 230 files, 2027 tests passed.
- `pnpm --dir apps/web build` — passed after rebasing onto current `upstream/main`.
- `pnpm i18n:check` — passed.
- `git diff --check` — passed.
- Manually verified panel width, member grid spacing, edit-field clipping/spacing, Chinese and pinyin member search, transfer-owner filtering, and existing permission paths.
- Storybook reached full module transformation in the pre-rebase check, then exited with the existing environment-level code 132; no component compile error was reported.

## Checklist

- [x] Updated documentation (not required; no public API or setup behavior changed)


lml2468 (dismissed):
## Review: #1038 — feat(channel-setting): 重构信息管理面板

**意图**:把 group/thread 信息设置面板拆成聚焦 feature sections + props-only UI rows,保持入口/权限/action/运行时行为不变;改进面板布局;为两处成员搜索加缓存拼音搜索。同 #1024 家族(builder-split 重构)。

锚定 head `f4eda6c2` / merge-base `3ce01400`(behind 7,`mergeable=MERGEABLE`,0 冲突)。28 文件 +1796/−1252。

---

### Gate 1 — 规格符合度 ✅

- 巨型文件拆分:`channelSettingGroupInfoSection.tsx` 375→12、`channelSettingThreadSections.tsx` 376→0(内容迁入 Danger/GroupProfileRows/GroupManagementRows/Nickname/Preference/ThreadInfo/ThreadActions/ThreadMd/ThreadWebhook 各聚焦文件)+ 共享 `ui/ChannelSettingRows`;无内容丢失、无重复。✅
- module.tsx section 接线保留(edit-push 回调等);Service/Actions/WKSDK/datasource 未动 —— 符合「行为保持」声明。✅

### Gate 2 — 代码质量 ✅(重点:权限门保留 + #1024 两条 🟡)

**权限门完好(load-bearing)✅**:拆分后的 `channelSettingGroupManagementRows`/`DangerSection` 仍按 `GroupRole.owner`/`GroupRole.manager` 门控(:38 owner-only、:150/166/182 owner||manager),转让/解散守卫保留 —— 无丢门。

**#1024 yujiawei 的 🟡#1(local search 绕过 `this.filter` → 数据泄漏/parity gap)→ 已修 ✅**:`list_vm.ts:52-54` 本地搜索结果 `this.filter ? localResults.filter(this.filter) : …` —— **本地搜索也应用 filter**,transfer-owner 等排除过滤不再被绕过。且有测试 `applies the caller filter before building the reusable search` 锁定。✅

**#1024 Jerry-Xin 的 🟡#2(15ms 性能阈值测试 flaky)→ 已规避 ✅**:新 `channelSettingMemberSearch.test.ts` 是**断言式**(中文/uid/全拼/大小写匹配、空查询还原、filter 先应用),**无时间阈值** —— 不 flaky。✅

**其它**:纯拼音搜索 helper 确定性;stale-request 保护在 VM 层(PR 声明,list_vm 承载);无 XSS 面(props-only rows + React 文本)。

**构建 / 测试(实跑):**
- `pnpm --filter @octo/web build` ✅ built in 3.55s。
- Subscribers list_vm + channelSetting(memberSearch/sections)✅ **3 files / 17 tests 全绿**。

---

### 🟡 非阻断
- 纯重构,回归风险在「搬运等价性」—— 已核权限门保留、拆分无丢失、#1024 两 🟡 均修 + 有测试,收敛。
- Storybook 环境级 exit 132(非组件编译错,PR 已说明)—— 环境噪声,非缺陷。

巨型面板拆分无丢失、权限门保留、#1024 的 local-filter parity gap 与 flaky 测试两 🟡 均已修复且有测试,构建与 17 测试通过。**可合并。**

Jerry-Xin (changes requested):
The refactor is well-structured and CI is green, but it introduces a member-search completeness regression for super groups.

### 🔴 Blocking

- 🔴 **Critical — "View more" and "Transfer ownership" search only the cached first 100 members in super groups.** Previously, opening the member list used server-side search: `SubscriberList` was created without a `localSearch`, so `SubscriberListVM.search()` always called `requestSubscribers()` and queried the server with the keyword (finding any matching member, paginated). This PR now wires a local-only index into both flows — `packages/dmworkbase/src/Components/Subscribers/index.tsx:153` ("View more") and `packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:44` (transfer owner) — both call `createChannelSettingMemberSearch(vm.subscribers / data.subscribers)`. In `packages/dmworkbase/src/Components/Subscribers/list_vm.ts:49`, when `localSearch` is present and the keyword is non-empty, `search()` sets `hasMore = false`, returns only the local cache results, and never falls back to the server. For super groups that cache is just the first page: `packages/dmworkbase/src/Components/Conversation/vm.ts:1205` (`getFirstPageMembers()` fetches `limit: 100, page: 1`, comment "如果是超级群则只获取第一页成员"). Net effect: in a super group, members beyond the first 100 can no longer be found via search in the "View more" list, and — more seriously — the owner cannot search or select them as an ownership-transfer target. This is a user-facing behavior regression versus the prior server-backed search.
  - **Fix options:** (a) keep server search as a fallback/merge when the local cache is known to be incomplete (e.g. super group with a partial cache), or (b) load a complete member index before enabling local-only search. Please also add a test covering a partial cache where a matching member exists only server-side, asserting the server path is still used.

### 💬 Non-blocking

- The follow-up commit `f4eda6c2` correctly restores filter parity by applying `this.filter` to local-search results in `list_vm.ts` (matching the server path) and adds a covering test — good. Removing the wall-clock timing assertion from the pinyin-search test also reduces CI flakiness while keeping the meaningful `toPinyin` call-count assertion.

### ✅ Highlights

- Clear section decomposition and props-only row wrappers establish cleaner feature boundaries for the channel-setting panels.
- Stale-response protection (`_requestVersion`) is well targeted and tested.
- Existing transfer-eligibility filtering (exclude self, exclude robots) is retained.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1038 (octo-web)

## Summary

Large refactoring PR (28 files) that decomposes the monolithic channel setting section builders into focused per-concern modules — nickname, preference, danger, thread info/actions/md/webhook, group management rows, and profile rows. Alongside the structural work, `SubscriberListVM` gains a `localSearch` parameter with pinyin-aware client-side member search and a `_requestVersion` counter to discard stale server responses after unmount.

The refactoring is faithful to the original behavior. Thread permission delegation, role-based guards, button type mappings, and error handling paths all preserve their original contracts. The `requestVersion` stale-response guard is a genuine improvement over the old code.

## Verification

- ✅ **Structural fidelity** — extracted row builders produce equivalent `Section`/`Row` trees with the same cells, properties, and click handlers as the original inline definitions.
- ✅ **Thread permission delegation** — `channelSettingThreadInfoSection.tsx` correctly delegates to `canRenameThread()` → `canManageThread()`, which resolves creator and parent-group role from the parent group subscriber list (not the thread's unsynced member cache).
- ✅ **Thread MD edit permission** — `channelSettingThreadMdSection.tsx:61` preserves the explicit `thread?.creator_uid === WKApp.loginInfo.uid` check alongside `can_edit_thread_md` and role checks.
- ✅ **Button type mapping** — `ChannelSettingActionRow` with `danger={true}` maps to `ListItemButtonType.warn`, matching the old code's explicit warn usage for leave/destructive actions.
- ✅ **Stale response prevention** — `_requestVersion` counter in `list_vm.ts` is incremented on unmount and checked before response callback, preventing post-unmount state updates.
- ✅ **Test coverage** — new tests for `list_vm.ts` (local search, server fallback, stale response) and `channelSettingMemberSearch.ts` (pinyin, case insensitivity, 10k perf) cover the new functionality.

Static analysis only at head `f4eda6c2`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two minor items below.

### P2 — Optimistic mutation before refresh (`channelSettingNicknameSection.tsx:46`)

`data.subscriberOfMe.remark = value` directly mutates the route-data model before `data.refresh()`. If refresh fails or returns stale server data, the in-memory model will reflect the optimistic value while the server may not have persisted the change. Since `refresh()` typically re-fetches from the server and the mutation is immediately overwritten, this is low-risk in practice. Not a change I'd block on — flagging for awareness.

### Nit — Per-render empty object references (`ui/ChannelSettingRows/index.tsx`)

All four `ChannelSettingRows` components pass `style={{}}` to underlying `ListItem` components, creating a new empty object on every render. Trivial React reconciliation overhead. Could hoist to a module-level `const emptyStyle = {}` constant, but this is genuinely cosmetic.

- Thread archive/unarchive flow: `shouldShowThreadArchiveAction` correctly gates on status (Active/Archived) and `canArchiveThread` permission with fallback.
- `GroupRole.owner`/`GroupRole.manager` enum comparisons in extracted files match the old code's numeric literal comparisons (owner=1, manager=2).
- Transfer owner `canTransferTo` filter and `disableSelectList` are preserved identically from the original code.
- Danger section: owner is correctly blocked from leaving (shows alert only), non-owners get the leave action.
- `didUnMount()` in `list_vm.ts` increments `_requestVersion`, invalidating any pending server responses for the destroyed VM instance.
- The `.finally()` pattern for loading-state reset in `channelSettingPreferenceSection.tsx` is an improvement over the old `.then()/.catch()` approach.
- Pinyin search index (`channelSettingMemberSearch.ts`) correctly normalizes via `getPinyin(toSimplized(value)).toLowerCase()` for all indexed fields (uid, remark, name, real_name).

## Verdict: APPROVED

Solid structural refactoring that faithfully decomposes monolithic channel setting files into focused, testable modules while preserving all behavioral contracts. The `requestVersion` stale-response guard and `.finally()` loading-state reset are genuine improvements. No correctness, security, or functional concerns.

[Octo-Q] verdict: APPROVE — faithful refactoring with improvements; no P1/P0 findings, only minor P2/nit observations.

cnwenf (changes requested):
# Code Review — PR #1038 (Mininglamp-OSS/octo-web)

>
> ⚠️ **Head-drift note:** The three reviews were conducted against frozen head `f4eda6c28e36…`. At consolidation time the PR head had advanced to `aa85e4edd453…` (new commits pushed after the reviews ran). Findings below are stated against `f4eda6c28e36` and should be re-confirmed against the current head before merge; the next scheduled review cycle will re-pick this PR under the "new-commit ⇒ re-review" rule.

## Summary

This PR refactors the channel-setting group/thread "information management" panels in `packages/dmworkbase` into focused per-feature section builders plus a reusable props-only `ChannelSettingRows` UI layer, and adds a cached Chinese/full-pinyin member-search index to the "View more members" and "Transfer ownership" flows, including a new `_requestVersion` stale-response guard. All three reviewers confirm the refactor itself is faithful — section rows, ordering, and permission gates are preserved element-by-element — and that security is clean (0 findings; transfer-owner self/bot exclusion preserved; real authz untouched in bridge/backend). The shared substantive concern, raised independently by all three reviewers, is the new local-search wiring: a `localSearch` index built from the in-memory IM-SDK subscriber snapshot silently replaces the server keyword search for non-empty keywords, so in large groups whose membership exceeds the synced cache, member search (and therefore transfer-owner target selection) can miss members with no server fallback. Reviewer verdicts split: octo-review-qwen3.8 = NEEDS_CHANGES, octo-review-glm5.2 = APPROVED_WITH_RISK, octo-review-qwen3.7 = APPROVED.

## Findings

### P2 — Local search over partial IM-SDK snapshot, no server fallback for non-empty keywords (`packages/dmworkbase/src/Components/Subscribers/list_vm.ts:49`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

*(octo-review-qwen3.8 filed this as two surface findings — transfer-owner `channelSettingGroupManagementRows.tsx:44` and view-more `Subscribers/index.tsx:155`; octo-review-qwen3.7 raised the same scope change as contract-001 at `Subscribers/index.tsx:155`. All three trace to the same root cause consolidated below.)*

For any non-empty keyword, `search()` now short-circuits into `localSearch` and returns only matches from the in-memory subscriber snapshot (`vm.subscribers` / `data.subscribers`, the IM-SDK route-data cache), with `hasMore=false` and no call to `requestSubscribers`. The previous path delegated every non-empty keyword to the server-side paginated search. This consolidates three related observations:

- **View-more members** (`packages/dmworkbase/src/Components/Subscribers/index.tsx:153`) — `localSearch={createChannelSettingMemberSearch(vm.subscribers)}` searches only the cache snapshot; a server-known member absent from the cache is silently unsearchable, and the result list gives no indication the scope is incomplete.
- **Transfer ownership** (`packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:56`) — the transfer-owner picker builds the same `localSearch` from `data.subscribers` over the same IM cache. An eligible transfer target outside the loaded cache is invisible and unselectable in the picker, so ownership **cannot be transferred to that member** — the most serious consequence of this shared root cause.
- **Contract** (`packages/dmworkbase/src/Components/Subscribers/list_vm.ts:49`) — the guard `if (this.localSearch && keyword.trim())` is where the no-server-fallback contract is introduced; a control VM constructed without `localSearch` hits `requestSubscribers` for the same keyword (proven by the candidate-validation control), so the behavior is new to this PR.

`ChannelSettingVM.reloadSubscribers` (`packages/dmworkbase/src/Components/Subscribers/vm.ts:365`) reads the IM cache without an explicit full-sync call; the comment at `:322` delegates sync to the existing channel lifecycle. For large groups where the SDK has not cached all members, the snapshot is partial. The transfer-owner self/bot exclusion filter is preserved (applied at index build in `createChannelSettingMemberSearch`, re-applied in `list_vm.search`, plus `disableSelectList`), so the specific security guard holds — the gap is the corpus/scope change, not the filter.

This is an intentional, tested design choice the author explicitly made ("Add cached Chinese/full-pinyin member search"), and cache completeness was not confirmed at runtime (the dataflow lens marked it uncovered), so it is a real risk to flag rather than a confirmed hard blocker. Worth a follow-up to either document the completeness assumption or add a server fallback when the local result set is empty/thin.

### P2 — New behaviors are not gated by any CI test (`packages/dmworkbase/src/Components/Subscribers/__tests__/list_vm.test.ts:1`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The three Vitest files this PR adds/updates as its regression evidence are never executed by any CI gate, so every behavior they pin can regress and still merge green. The `ci.yml` build job (lines 87-96) runs only install / `pnpm i18n:check` / `pnpm build` / `pnpm lint` / `pnpm lint:wkmodal` — no `pnpm test` step; `turbo run build` and `turbo run lint` do not depend on the `test` task, and the only test-running workflows (`e2e.yml`, `e2e-nightly.yml`) run Playwright E2E, not the packages/dmworkbase Vitest suite. Verified at head `f4eda6c28e36` by structural grep over the actual workflow/turbo config: `grep -rn 'vitest|pnpm test|turbo run test' .github/workflows/` returns empty. The newly introduced behaviors whose only pinned protection is these ungated tests include the `_requestVersion` stale-response guard and local-vs-server fallback (`list_vm.ts:49-72`), the pinyin index, the transfer-owner filter, and the nickname mutation+refresh. The PR body's "2027 tests passed" is local-only, not a merge gate. Fix direction: wire `pnpm test` (turbo) into the build job, or accept that these tests do not block regressions.

### P2 — Transfer-owner eligibility predicate is not pinned by any failing test (`packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:41`) [source: octo-review-qwen3.8]

`canTransferTo` (excludes self via `uid !== WKApp.loginInfo.uid` and robots via `(orgData?.robot === 1) !== true`, :41-43) is the production guard deciding who can receive group ownership, but no failing assertion exercises it. The only filter test (`channelSettingMemberSearch.test.ts:42-50`) tests the pure helper with a stub predicate, never the real one; `channelSettingSections.test.ts` builds the owner section but only asserts row counts/values and mocks `transferChannelSettingOwner`. Confirmed via mutation at head `f4eda6c28e36`: inverting the robot check (`!== true` → `=== true`, making only robots eligible) leaves the PR's `channelSettingSections.test.ts` green (9/9) while a focused probe that reads the real pushed `<SubscriberList>` filter/localSearch closures fails. A regression inverting the robot check or dropping self-exclusion would therefore ship green and let an owner transfer ownership to a bot/themselves. Mutation fully reverted. Fix direction: add a unit test that calls `createChannelSettingMemberSearch(data.subscribers, canTransferTo)` (or the row's pushed filter) with a robot member and the self uid present, asserting both are excluded.

### P2 — `localSearch` callback has no error handling (`packages/dmworkbase/src/Components/Subscribers/list_vm.ts:52`) [source: octo-review-qwen3.7]

The `search()` method calls `this.localSearch(keyword)` at line 52 without a try/catch wrapper. If the injected callback throws (e.g., `getPinyin()` raises on unexpected input, or the index was built with stale data), the entire `search()` call crashes with an unhandled exception. The user sees no feedback — the component is left in whatever intermediate state existed before the throw (subscribers already cleared to `[]` at line 47). The old code path (`requestSubscribers()`) at least reaches the server error handler. Consider wrapping in try/catch and falling back to `this.requestSubscribers()` on failure:

```ts
try {
  const localResults = this.localSearch(keyword);
  this.subscribers = this.filter
    ? localResults.filter(this.filter)
    : localResults;
} catch {
  this.requestSubscribers();
  return;
}
```

### P3 — Cross-script Chinese search is asymmetric (`packages/dmworkbase/src/features/channelSetting/channelSettingMemberSearch.ts:20`) [source: octo-review-qwen3.8]

`defaultPinyinConverter` applies `toSimplized` only when deriving the pinyin field (:20); the stored raw Chinese value (:39) and the query keyword (:54) are only trim+lowercased. So a Simplified keyword will not match a Traditional member name/remark via the raw-value branch (and vice versa), even though a simplification utility is imported and the pinyin path crosses scripts — matching is case-insensitive but script-sensitive. Confirmed at head `f4eda6c28e36` against the real converter: `filter('劉德華')→[u-liu]` and `filter('liudehua')→[u-liu]` match, but `filter('刘德华')→[]`. The PR's tests use only Simplified fixtures plus a stub `toPinyin`, so this branch is untested. Search convenience only, not data loss. Fix direction: if cross-script matching is a requirement, normalize both stored values and keyword through `toSimplized`.

### P3 — Transient `hasMore=false` window when clearing a local-search keyword (`packages/dmworkbase/src/Components/Subscribers/list_vm.ts:26`) [source: octo-review-glm5.2]

`search()` sets `this.hasMore = false` in the local-search branch. When the user subsequently clears the keyword, `search('')` falls through to `requestSubscribers()`, which recomputes `hasMore` from the server response — but between the `search('')` call and the server promise resolving, `hasMore` stays false, so `loadMoreSubscribersIfNeed` (which checks `!this.hasMore`) short-circuits during that window. It self-corrects once the response lands (control resolves `hasMore=true`); low-impact, flagged for awareness.

### P3 — Recursive auto-fetch-on-thin-results branch is untested (`packages/dmworkbase/src/Components/Subscribers/list_vm.ts:90`) [source: octo-review-glm5.2]

The branch `if (this.filter && this.hasMore && this.subscribers.length < this.limit) { this.currPage++; await this.requestSubscribers(requestVersion); }` (`:90-93`) is production-reachable — a client-side filter that shrinks visible results below the limit while the server still has more pages triggers repeated auto-fetch — but no test sets `filter` together with `hasMore=true` and `subscribers.length < limit`. A mutation protocol confirmed the existing suite does not cover it (disabling the guard leaves all baseline tests green). A regression here (infinite recursion if `hasMore` never flips, wrong `currPage` increment, stale `requestVersion` leaking into the recursive call) would stay green. Add a case exercising `filter + hasMore=true + thin results`.

### Nit — Direct mutation of route data in nickname save (`packages/dmworkbase/src/features/channelSetting/channelSettingNicknameSection.tsx:42`) [source: octo-review-qwen3.7]

After the API call succeeds, the code writes `data.subscriberOfMe.remark = value` before calling `data.refresh()`. While the refresh replaces the data shortly after, the intermediate mutation violates the immutability convention and could cause a brief inconsistent render if any observer reads `subscriberOfMe.remark` between the assignment and the refresh. The old code (`channelSettingGroupInfoSection.tsx:369-371`) did not mutate route data — it just called `data.refresh()` in the `.then()`.

### Nit — Missing behavioral tests for extracted section builders [source: octo-review-qwen3.7]

Several newly extracted section builders (`buildGroupProfileRows`, `buildGroupManagementRows`, `buildThreadInfoSection`, `buildDangerSection`) lack dedicated behavioral tests beyond the row-count assertions in `channelSettingSections.test.ts`. Permission guard logic (owner-only transfer, manager-or-owner editing, thread archive visibility) is tested implicitly through manual verification but not by automated tests that would catch future regressions.

### Nit — `channelSettingInputEditPush` adds a new CSS class for all callers (`module.tsx:1563`) [source: octo-review-qwen3.7]

`channelSettingInputEditPush` adds a CSS class (`wk-channelsetting-input-edit`) that was not present in the old `inputEditPush` contract. All channel setting sections using `inputEditPush` (group name, group notice, group nickname, thread name, group remark) now render with constrained width and horizontal margin. The type signature is preserved, but runtime layout behavior changes for all callers.

### Nit — Row wrappers inject an empty `style={{}}` (`ui/ChannelSettingRows/index.tsx:27`) [source: octo-review-qwen3.7]

All four row wrapper components inject an empty `style={{}}` prop on their underlying cell components. The old code used cells directly without passing `style`. While typically inert, this creates a new inline style attribute on every row's DOM element.

- **Refactor parity / permission gates** — section builders preserve rows, ordering, and RBAC gates element-by-element versus the deleted files (group-name/announcement manager gate, transfer-owner owner gate, GROUP.md `canEdit`, webhook `isManager`, thread rename/archive/dismiss guards); `module.tsx` priority ordering unchanged; barrel re-exports and the `subTitle`→`value` rename are faithful. [all three]
- **Backward compatibility** — `SubscriberList`'s `localSearch` is optional; the three callers that omit it (add-manager, add-bot-admin, remove-members) retain the existing server search and pagination (a control VM without `localSearch` reaches `requestSubscribers`). [octo-review-glm5.2 + octo-review-qwen3.7]
- **Transfer-owner self/bot exclusion preserved** — `canTransferTo` (`uid !== WKApp.loginInfo.uid`, `robot === 1`) holds end-to-end; the double-application (index build + VM filter) is idempotent, redundant but not a correctness defect. [octo-review-glm5.2 + octo-review-qwen3.8]
- **Stale-response guard** — the new `_requestVersion` counter is correctly wired across all call paths; machine-confirmed safe (sequential auto-fetch conserves members and the pagination cursor, exactly 3 server calls over 3 pages). [octo-review-qwen3.8 + octo-review-qwen3.7 + octo-review-glm5.2]
- **Pinyin index** — preserves member order, no duplicates; cached and reused across keystrokes; empty query returns all; tested with a 10K-member perf case. [octo-review-qwen3.8 + octo-review-qwen3.7]
- **Tests + build** — 230 files / 2028 unit tests and `apps/web build` green at HEAD across all three reviewers; `i18n:check` and CSS design-token usage clean. [all three]
- **Security** — 0 findings (P0–P3): pure frontend refactor, all client-side display gates preserved, real authz stays in the bridge/backend (untouched), the search keyword stays an in-memory `.includes()` comparison (no injection/HTML sink; XSS-safe via React escaping), no new dependencies, no secrets. [all three]

## Verification Results

| Code Review | NEEDS_CHANGES | 4 P2 + 3 P3 + 4 nits (deduped across 3 reviewers); no P0/P1; one reviewer (qwen3.8) requested changes on the local-search corpus change, two approved (glm5.2 with risk, qwen3.7) |
| E2E Test | PARTIAL | 230 files / 2028 unit tests + build + focused probes green at HEAD across all three reviewers; full Playwright E2E blocked (frontend-only repo, no octo-server backend in this environment); cross-service backend semantics by static analysis only |
| Security | CLEARED_WITH_RISK | 0 exploitable findings across all three; transfer-owner self/bot exclusion preserved, authz unchanged in bridge/backend, XSS-safe via React escaping; risk flag reflects the client-side search-scope change noted above |

Candice00711:
Closing this PR in favor of a fresh replacement so automated review runs against the latest head, including the authoritative server-search fix for partial member caches.

PR #1039, opened 2026-07-23
feat(summary): improve markdown reader layout and citations

wanyiwang06:
## Summary
- add a shared Markdown summary reader with outline navigation, table scrolling, and a reference-aligned detail layout
- improve citation badges, grouped citation timelines, cited-message highlighting, and jump-to-original behavior
- keep rendering consistent across traditional, agent, shared, and version-history summary views
- normalize full-width numeric/team citation markers at render time, including existing stored summaries
- preserve the latest upstream summary action menu while resolving the branch conflicts with current main

## Validation
- focused citation and Markdown reader tests: 35 passed
- web production build: passed
- diff whitespace check: passed

## Notes
- citation normalization is presentation-only and does not mutate stored summary content


Jerry-Xin (dismissed):
The feature is relevant to octo-web, but Markdown heading alignment and Copy Markdown have blocking correctness issues.

## 🔴 Blocking

- 🔴 Critical — Setext H2 headings break outline-to-heading alignment. `extractSummaryOutline()` only recognizes `##`/`###`, while ReactMarkdown also renders Setext headings. `CitationText` assigns IDs sequentially to every rendered H2/H3, so an untracked Setext heading consumes the next ID. For `Intro\n---\n## Results`, the “Results” link targets “Intro,” while the actual Results heading has no ID. Use the same Markdown AST for outline extraction and rendering, or account for all headings consistently. [SummaryMarkdownReader/index.tsx:15](/Users/mlamp/.openclaw/tmp/pr-review-gptuz_qb/repo/packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:15), [CitationText.tsx:218](/Users/mlamp/.openclaw/tmp/pr-review-gptuz_qb/repo/packages/dmworksummary/src/components/CitationText.tsx:218)

- 🔴 Critical — Copy Markdown can silently do nothing for completed Agent summaries. The button is immediately enabled, but Agent content may only exist in the asynchronously loaded `personalResult`; clicking before it loads hits the empty-content early return without feedback. This action needs the same readiness/loading guard already used by “Forward to Chat.” Also use a non-empty fallback rather than `??`, since an empty `detail.result.content` prevents fallback to personal content. [SummaryDetailPage.tsx:1946](/Users/mlamp/.openclaw/tmp/pr-review-gptuz_qb/repo/packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1946), [SummaryDetailPage.tsx:3347](/Users/mlamp/.openclaw/tmp/pr-review-gptuz_qb/repo/packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3347)

## 💬 Non-blocking

- 🟡 Warning — Fence tracking toggles on every line beginning with either fence marker, even when the marker differs from the opening fence or is not a valid closing fence. Code containing `~~~` inside a backtick fence can therefore expose false headings and shift IDs. AST-based extraction would resolve this too. [SummaryMarkdownReader/index.tsx:13](/Users/mlamp/.openclaw/tmp/pr-review-gptuz_qb/repo/packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:13)

- 🟡 Warning — The reader test mocks `CitationText`, so it cannot detect parser/extractor divergence such as the Setext failure. Add an integration test using the real renderer, including Setext headings, duplicate headings, and mixed fenced content. [SummaryMarkdownReader.test.tsx:4](/Users/mlamp/.openclaw/tmp/pr-review-gptuz_qb/repo/packages/dmworksummary/src/ui/SummaryMarkdownReader/SummaryMarkdownReader.test.tsx:4)

## ✅ Highlights

- Citation display numbers are carried through routing into both normal and folded conversation messages.
- Duplicate-safe anchors, responsive layout, localization, Storybook coverage, and theme tokens are thoughtfully included.

Jerry-Xin (changes requested):
The Phase 1 markdown reader and citation markers are relevant to octo-web, and the XSS surface is handled well (react-markdown + rehype-sanitize allowlist, no rehype-raw/dangerouslySetInnerHTML, links get rel="noopener noreferrer"). However, the outline→heading anchor mapping and the Copy Markdown action have correctness issues that should be fixed before merge.

## 🔴 Blocking

- 🔴 Setext headings break outline-to-heading anchor alignment. `extractSummaryOutline()` only recognizes ATX `##`/`###`, but react-markdown/remark also renders Setext headings (`Title\n---` → H2). The `heading()` component in `markdownComponents` assigns `headingIds[headingIndex++]` sequentially to every rendered H2/H3, so an untracked Setext heading consumes the next ID out of order. For input `Intro\n---\n## Results`, the "Results" outline link targets "Intro" and the real Results heading gets a mismatched/absent id. Drive both the outline and the rendered heading ids from the same markdown AST (or count all heading forms consistently). packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:15 (extractSummaryOutline ATX-only heading regex), packages/dmworksummary/src/components/CitationText.tsx:218 (sequential headingIds assignment).

- 🔴 Copy Markdown can silently do nothing for completed Agent summaries. The button is enabled immediately, but Agent content may only exist in the asynchronously loaded `personalResult`; clicking before it loads hits the empty-content early return with no feedback. Add the same readiness/loading guard used by "Forward to Chat". Also the `detail?.result?.content ?? this.state.personalResult?.content ?? ""` fallback uses `??`, which only falls through on null/undefined — an empty-string `detail.result.content` ("") blocks the personalResult fallback. Use a non-empty check instead. packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1946 (handleCopyMarkdown).

## 💬 Non-blocking

- 🟡 Fence tracking in `extractSummaryOutline` toggles on any line starting with either fence marker (``` or ~~~), even when it does not match the opening fence. A `~~~` line inside a backtick fence can flip the state and expose false headings / shift ids. AST-based extraction would resolve this as well. packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:13.
- 🟡 The reader test mocks `CitationText`, so it cannot catch extractor↔renderer divergence like the Setext case. Add an integration test with the real renderer covering Setext headings, duplicate headings, and mixed fenced content. packages/dmworksummary/src/ui/SummaryMarkdownReader/SummaryMarkdownReader.test.tsx:4.

## ✅ Highlights

- Markdown rendering stays XSS-safe: react-markdown with a rehype-sanitize allowlist that only adds the custom citation tags (with a tight attribute list), no raw HTML passthrough, and anchors rendered with rel="noopener noreferrer".
- Citation display numbers are carried through routing into both normal and folded conversation messages.
- Duplicate-safe anchor ids, responsive layout, i18n coverage, Storybook story, and theme tokens are all included. No new dependencies added.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1039 (octo-web)

## Summary

This PR ships a broad set of improvements across the octo-web monorepo: a new `SummaryMarkdownReader` component with an outline sidebar, citation display-index propagation for deep-linking into conversations, WKSDK abstraction-seam migration (`current*Runtime` helpers replacing `WKSDK.shared()` calls), an agent-chat `selected_channels` bridge for archived-thread context, Onboarding focus-trap fix, HtmlDoc access-request badge + viewer presence bar, and E2E/CI hardening (MSW baseline handlers, proxy-error guard, `vite preview` for e2e). Overall the changes are coherent and well-scoped for each concern; three P2 items are noted below.

## Verification

Static analysis only at head `c700d268b05b12da2877b824080f6413420410e8`; build and tests not executed in this environment.

- ✅ **Diff scope** — ~40 files changed across 6 packages + apps/web; changes are thematically grouped (summary reader, citation propagation, WKSDK seam, agent-chat bridge, onboarding, e2e).
- ✅ **Data-flow: headingIds** — `extractSummaryOutline()` produces IDs that flow to `CitationText` via `headingIds` prop; the `heading()` component increments `headingIndex` per rendered h2/h3. Verified the propagation chain `SummaryMarkdownReader → CitationText → heading component`. Noted a P2 alignment issue below.
- ✅ **Data-flow: initLocateCitationDisplayIndex** — Propagated from `JumpLink` → `ShowConversationOptions` → `EndpointCommon` (routing key) → `ChatContentPage` → `Conversation` / `FoldSessionExpandedList`. Routing key includes the index for cache-busting. Truthiness check at render noted as P2 below.
- ✅ **Data-flow: selected_channels** — `SummaryCreatePage` passes `selectedChats` to `AgentChatPanel`, which maps to `{chat_id, chat_type, name, is_archived?}` and forwards to both `agentChatStream` and `agentChat`. Type added to `AgentChatParams`. The archived-thread bridge contract is sound.
- ✅ **Data-flow: is_archived** — `channelConvert.ts` correctly propagates `is_archived: true` when `chatType === 'thread' && orgData.thread.status === 2` (ThreadStatus.Archived).
- ✅ **WKSDK seam migration** — Grep across changed files confirms `WKSDK.shared()` replaced by `currentImRuntime()`, `currentChatRuntime()`, `currentConversationRuntime()` helpers consistently. New `getCurrentImConversationsDirectly()` reads `sdk.conversationManager.conversations` directly, which is appropriate for the direct-read use case.
- ✅ **Onboarding focus-trap** — Fix includes `dialog` itself in the focusable set; `visibility: hidden` used for skip-transition-target. Both are correct patterns.
- ✅ **E2E hardening** — MSW baseline handlers cover emoji manifest, health, avatar palette, message sync, and OBO grants. `EmojiService.ts` skips manifest fetch when `VITE_E2E_MOCK === "1"`. Proxy-error CI guard uses `PIPESTATUS[0]` + `tee` + grep, which correctly captures Vite proxy failures that previously slipped through.
- ✅ **i18n** — Four new keys added in both `en-US.json` and `zh-CN.json`: `copyMarkdown`, `markdownCopied`, `newAgentChat`, `outline`. No missing translations.

## Findings

No P0/P1 issues; three P2 items below.

### P2 — Heading-ID misalignment between extractSummaryOutline and ReactMarkdown (`packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:15`)

`extractSummaryOutline` matches headings with `/^(#{2,3})\s+/` (line-start anchored), so it skips headings nested inside blockquotes (`> ## Heading`), list items, or other container syntax. ReactMarkdown, however, renders those as real `<h2>`/`<h3>` elements and calls the `heading()` component from `CitationText.tsx:216-220`, which increments `headingIndex` for each one.

When the document contains any blockquote/list-nested h2/h3 before a regular h2/h3, the `headingIndex` counter desyncs from the `headingIds` array: the rendered heading gets a wrong `id`, and the outline link scrolls to the wrong section.

Fix direction: use a remark plugin (or a single shared heading-visitor) so both the outline and the rendered IDs come from the same parse tree, instead of two independent parsers that disagree on what counts as a heading.

### P2 — Grid always reserves outline column even when outline is hidden (`packages/dmworksummary/src/ui/SummaryMarkdownReader/index.css:2`)

The `.summary-markdown-reader` grid uses `grid-template-columns: minmax(0, 1fr) calc(var(--wk-sp-12) * 4)`, allocating ~192px for the outline nav. When `showOutline=false` or `outline.length === 0`, the `<nav>` is not rendered but the grid still reserves the right column, leaving wasted space.

Fix: apply the two-column grid only when the outline is actually rendered (e.g., via a `.has-outline` class), and default to `display: block` or single-column grid.

### P2 — Citation marker suppressed when `initLocateCitationDisplayIndex` is `0` (`packages/dmworkbase/src/Components/Conversation/index.tsx:2185`)

The truthiness check `this.props.initLocateCitationDisplayIndex && message.messageSeq === ...` treats `0` as falsy. Display indices are typically 1-based so this is unlikely in practice, but the TypeScript type is `number | undefined` and nothing at the type level prevents `0` from reaching this prop.

Consider using `this.props.initLocateCitationDisplayIndex != null && this.props.initLocateCitationDisplayIndex > 0` for explicit intent. Same pattern in `FoldSessionExpandedList.tsx:82`.

## Verdict: APPROVED

No P1 blockers. Three P2 items are non-blocking — the heading-ID alignment issue is a correctness concern for documents with nested headings but does not break the common case; the grid layout wastes space when the outline is hidden but does not affect functionality; and the citation-marker truthiness check is unlikely to hit `0` in practice given 1-based indexing. All three are worth fixing in follow-up but do not warrant blocking this PR.

[Octo-Q] verdict: APPROVE — no P1 blockers; three P2 non-blocking findings (heading-ID alignment, grid outline column, citation marker truthiness). Safe to proceed to Opus终审.

cnwenf (changes requested):
# Code Review — PR #1039 (Mininglamp-OSS/octo-web)

> Consolidated from three independent reviews (octo-review-qwen3.7, octo-review-glm5.2, octo-review-qwen3.8) at head `501c2b7d`. Findings are deduplicated; `[source: ...]` marks which reviewer(s) raised each finding. The three full raw reviews remain as comments on this issue for audit.

## Summary

This PR adds a responsive `SummaryMarkdownReader` with scroll-synchronized H2/H3 outline navigation, pinned citation timelines (display-number renumbering + jump-to-message for both normal and folded sessions), "Copy Markdown" / "New Agent Chat" result-footer actions, result-surface styling, and 12 new i18n keys — 21 files, +619/−142, presentation-only (no server-contract, dependency, or auth changes). The change is mostly well-built: the raw→display index map, the `mergeGroupMessages` refactor, i18n parity, the route-key and event wiring, and the sanitize path all held up, and the Round-1 regressions (dropped citation `context_before`/`context_after`, and the silent click no-op) are resolved at this head. However, one confirmed, reachable defect remains — the outline anchor-id sequence diverges from what the renderer emits, so outline links jump to the wrong section for blockquoted/setext/indented headings (executable-confirmed by two reviewers) — alongside a CI gate that runs no unit tests and broad test-coverage/test-effectiveness gaps.

## Findings

### P1 — Unit tests never run in the CI gate; the new tests gate nothing (`.github/workflows/ci.yml:88`) [source: octo-review-glm5.2 + octo-review-qwen3.8]

The `ci.yml` build job runs `pnpm install --frozen-lockfile`, `pnpm i18n:check`, `pnpm build`, `pnpm lint`, `pnpm lint:wkmodal` and **no** `pnpm test` / `turbo run test` / `vitest` step (a structured parse of all 20 `.github/workflows/*.yml` finds zero unit-test runner steps; `e2e*.yml` runs only `pnpm build:e2e` Playwright on separate triggers). The tests this PR adds — `SummaryMarkdownReader.test.tsx`, `CitationText.test.tsx`, `SummaryCreatePage.test.tsx` — are therefore never executed by the PR status gate and block zero regressions; the PR's cited "tests passed" validation is local-only. *(severity split: glm5.2=P2, qwen3.8=P1. qwen3.8 calibrates this as a pre-existing repo-wide infrastructure gap with no reachable runtime/data/auth consequence in this diff — strongly worth fixing, but not blocking on its own.)*

**Fix**: add a `pnpm test` (or `pnpm --filter @dmwork/summary test`) step to the build job or a dedicated test job.

### P2 — Outline ↔ heading-id desync: outline links jump to the wrong section (`packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:15`) [source: octo-review-glm5.2 + octo-review-qwen3.8] — **confirmed, must-fix**

`extractSummaryOutline` counts only column-0 ATX `## `/`### ` lines (fence-aware, skipping entries with empty stripped text), but `CitationText`'s heading closure (`CitationText.tsx:218`) assigns `headingIds[headingIndex++]` to **every** h2/h3 react-markdown emits — including blockquote-nested (`> ## B`), setext (`Title\n---`), indented-ATX (`  ## B`), and image-only (`## `) headings the regex skips. The first skipped-but-rendered heading shifts the entire id sequence: later outline links land on the wrong section, the trailing real heading gets `id=undefined` (unaddressable), and the scroll-spy highlights the wrong entry. Reproduced with the real renderer (no mock):

```markdown
## A
> ## Quoted
## B
```

yields outline ids `['a','b']` but three `<h2>` render — `#b` lands on "Quoted" and the real "B" heading is id-less. The aligned-heading control passes, so the dominant clean-LLM-summary shape works; the failure is isolated to the parser/renderer heading-set mismatch. The "synchronize outline with reading scroll" commit synced scroll-highlight but did not change the id source, so the desync remains. Both sides of the contract are new in this diff.

**Fix**: derive `headingIds` from the same remark AST the renderer walks (run `extractSummaryOutline` on the parsed hast, or generate ids inside the h2/h3 component with the same slug rule), and add an integration test asserting every outline href resolves to a live heading.

### P2 — Reader test mocks `CitationText`, so the outline↔heading invariant is tautological (`packages/dmworksummary/src/ui/SummaryMarkdownReader/SummaryMarkdownReader.test.tsx:4`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The render test hoists `vi.mock('../../components/CitationText', ...)` replacing the real react-markdown + rehype-sanitize + heading-id pipeline with a hand-rolled mock built from the *same* `extractSummaryOutline` the component uses. The "links outline to headings" assertion can therefore only ever agree with itself and cannot detect the P2 desync above. Mutation-confirmed: mutating only the production id-binding at `CitationText.tsx:220` (`headingIds[headingIndex++]` → `[+1]`) leaves the existing test passing while a no-mock probe dies. The real sanitize path is likewise never exercised.

**Fix**: render the real `CitationText` (with a blockquote/setext case) and assert outline hrefs match rendered h2/h3 ids in order with matching counts.

### P2 — Copy Markdown / New Agent Chat actions and their gating are untested (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1946`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`handleCopyMarkdown` has three branches (empty-content early return, clipboard success + toast, clipboard failure + error toast) plus a content fallback chain — none tested. `handleNewAgentChat` dispatches a `CustomEvent` and is untested. The button gating condition (`detail.status === COMPLETED && trigger_type === AGENT`) at line 3347 is uncovered — no test verifies the buttons appear for completed agent tasks and are absent otherwise. Mutation-confirmed: changing the empty branch to toast leaves all existing `SummaryDetailPage` tests green.

**Fix**: add a test that clicks Copy Markdown and asserts the clipboard payload is the displayed summary content with a success toast (error toast on rejection), exercises `handleNewAgentChat`'s event dispatch, and verifies the AGENT+COMPLETED gating.

### P2 — CSS nesting error: `border-radius: 0` override is accidentally nested → dead CSS (`packages/dmworksummary/src/index.css:670`) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

`.summary-detail-result-footer` opens at line 657 but is missing its closing `}` before line 670, so the `.summary-detail-result .summary-markdown-reader__content { border-radius: 0 }` rule is nested inside the footer block. The compiler flattens it to `.summary-detail-result-footer .summary-detail-result .summary-markdown-reader__content` — but the reader content renders as a **sibling** of the footer, so the override never matches. Confirmed by compiling the real stylesheet (esbuild) and matching against the real DOM fragment (jsdom): the element keeps its baseline `var(--wk-r-md)` radius, producing a rounded-corner / doubled-border seam on all three result surfaces. *(severity split: qwen3.7=P2, qwen3.8=P3 — cosmetic but reachable on every result surface.)*

**Fix**: close the footer's `}` before line 670 so the rule becomes a top-level sibling selector.

### P2 — Scroll-spy `useEffect` (added by "synchronize outline") is untested (`packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:46`) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The scroll-sync `useEffect` (lines 46–64) is the most complex runtime behavior in the new component: it attaches a passive scroll listener to a `.summary-detail-content-scroll` ancestor, computes an activation line via `getBoundingClientRect`, and iterates headings to set the active outline link / `aria-current`. None of this is tested — the existing test covers only outline extraction and static rendering. A regression (wrong active section, or a listener leak on unmount) would merge green.

**Fix**: add a test that scrolls the reader container and asserts the active outline link updates, and that the listener is removed on unmount.

### P2 — The `[n]` citation-marker render sites have zero test coverage (`packages/dmworkbase/src/Components/Conversation/FoldSessionExpandedList.tsx:82`, `packages/dmworkbase/src/Components/Conversation/index.tsx:2185`) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

The marker rendering in both the normal list (`Conversation/index.tsx:2185`) and the expanded-fold list (`FoldSessionExpandedList.tsx:82`) — gated by `locateCitationDisplayIndex && message.messageSeq === locateCitationMessageSeq` — plus its prop-forwarding chain matches no dmworkbase test. The existing `CitationText` click test stops at the routing boundary (`showConversation`) and never verifies downstream marker rendering. Mutation-confirmed: match-break, text-break, and guard-break mutants each ship green against the existing suite while a focused probe on the real component kills all three. *(severity split: qwen3.7=P2, qwen3.8=P3.)*

**Fix**: commit a spec asserting the marker appears only on the located `messageSeq` with the correct `[displayIndex]`.

### P2 — Collapsed-fold citation marker dropped on the citation-jump path, contradicting "including folded sessions" (`packages/dmworkbase/src/Components/Conversation/FoldSessionExpandedList.tsx:82`) [source: octo-review-glm5.2]

The marker renders only in `Conversation/index.tsx:2185` (regular non-folded) and `FoldSessionExpandedList.tsx:82` (expanded-fold); `FoldSessionCard/index.css:198` hides `.wk-fold-session-card-expanded` with `display:none` unless expanded. `Conversation.locateMessage` (`index.tsx:1181-1228`) DOES auto-expand a collapsed fold (`setFoldSessionExpanded` at `:1206`), but the citation-jump path does not call it: `jumpToCitation` sets `initLocateMessageSeq`/`initLocateCitationDisplayIndex`, and the VM locate path `requestMessagesOfFirstPage` → `syncMessages` → `refreshAndLocateMessages` → `scrollToMessage` (`vm.ts:1966`) → `messageScrollElement` (`vm.ts:2244`) returns `foldSession.anchorId` (the fold card) for a collapsed fold without expanding. `locateMessage` (the auto-expand site) is only invoked for reminders and self-recursion, not the citation `initLocate` path. Folds default collapsed (`vm.ts:407/432`). So clicking a citation whose source is in a collapsed fold scrolls to the fold card with no visible marker and no auto-expand — directly contradicting the PR's "mark the located source message, including folded sessions" claim. The expanded-fold case works.

**Fix**: route the citation-jump through `Conversation.locateMessage` (which auto-expands), or have the VM `scrollToMessage` path expand the target fold before scrolling (or render a marker on the collapsed fold card itself).

### P2 — `CitationTimeline` interactive toolbar is untested (`packages/dmworksummary/src/components/CitationBadge.tsx:108`) [source: octo-review-qwen3.7]

The new `CitationTimeline` component (~70 lines) adds text-search filtering (`normalizedQuery`), a cited-only toggle (`citedOnly` state), prev/next navigation (`current` index), and active-message highlighting. Existing tests only verify initial render of a single citation badge; none exercise the interactive toolbar, search filtering, toggle, or navigation. A regression in the filter logic or `displayByRawIndex` mapping would go undetected.

**Fix**: add tests exercising the toolbar's search filtering, cited-only toggle, and prev/next navigation.

### P3 — New window listener has no teardown (`packages/dmworksummary/src/module.tsx:177`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The new `summary-open-new-agent-chat` listener (and the pre-existing `summary-open-chat-with-reference` at line 172) are inline anonymous functions, not stored in module-level vars, and are absent from the `import.meta.hot.dispose` block (which removes only the pre-existing `focus` listener). Confirmed via the real `SummaryModule.init`: a second init doubles registrations and a single event pushes two routes (double navigation + double session clear). Production registers the module once, so the realistic trigger is dev/HMR — low impact, but teardown for a diff-new resource is genuinely absent. *(severity split: qwen3.7=Nit, glm5.2=P3, qwen3.8=P3.)*

**Fix**: store both handlers in module-level vars (matching the `_attentionFocusHandler` pattern) and `removeEventListener` them in the `import.meta.hot.dispose` block.

### P3 — Timeline header double-counts cited messages as "context" (`packages/dmworksummary/src/components/CitationBadge.tsx:136`) [source: octo-review-glm5.2 + octo-review-qwen3.8]

The meta line binds `context` to `merged.length` labeled `{{context}} context messages` / `{{context}} 条上下文`, while `cited` is shown separately as `{{cited}} direct citations`. But `mergeGroupMessages` builds `merged` from `context_before` + the cited message + `context_after`, so `merged.length` includes the cited message that is also counted in `cited.length` — "context messages" overstates by exactly the cited count. Confirmed: 1 cited + 2 context renders `TeamChat · 3 条上下文 · 1 条直接引用` (should read 2 context). Reachable on every pinned popover with context.

**Fix**: bind `context` to `merged.length - cited.length` (or `merged.filter(m => !m.cited).length`), or relabel.

### P3 — Copy Markdown silently no-ops on empty content (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1948`) [source: octo-review-qwen3.8]

`if (!content.trim()) return;` skips both the success and error toast, so the enabled button (gated only on AGENT+COMPLETED) gives zero feedback on an empty/whitespace result — confirmed against the real method with both sibling branches toasting in the same run. A second wrinkle: the `??` chain doesn't fall through on an empty *string*, so a defined-but-empty `detail.result.content=''` short-circuits a non-empty `personalResult.content` and copies nothing.

**Fix**: emit a neutral/warning toast (or disable the button) on the empty branch.

### P3 — `mergeGroupMessages` refactor has no equivalence test (`packages/dmworksummary/src/components/CitationBadge.tsx:65`) [source: octo-review-qwen3.8]

The rewrite (dedup-by-seq-or-content, cited-wins tiebreak, seq-then-time sort) is only exercised indirectly through label tests that never import `CitationBadge.tsx`. The refactor is behavior-invariant today (12/12 verbatim base↔head fixtures pass), but dedup/cited-wins/sort mutants all ship green against the existing suite — a future drift would not fail any test.

**Fix**: commit the focused equivalence spec.

### Nit — `displayIndex` falsy guard at two render sites (`FoldSessionExpandedList.tsx:82`, `Conversation/index.tsx:2185`) [source: octo-review-qwen3.7]

Both sites use `locateCitationDisplayIndex && ...` which treats `0` as falsy, identical to `undefined`; the prop type `number | undefined` does not exclude `0`. No bug today since display indices are 1-based citation numbers, but `!= null` would be more explicit and future-proof.

### Nit — `CitationBadge` font-size hardcoded to `10px` (`packages/dmworksummary/src/components/CitationBadge.css:20`) [source: octo-review-qwen3.7]

`font-size: 10px` and `line-height: 14px` break the design-token pattern used elsewhere in the same file (`var(--wk-text-size-xs)`, `var(--wk-lh-*)`); other badge properties correctly use `--wk-*` tokens. Minor inconsistency that bypasses theme-level font scaling.

### Nit — `jumpToCitation` silently returns on missing fields (`packages/dmworksummary/src/components/CitationBadge.tsx:55`) [source: octo-review-qwen3.7]

When a citation lacks `channel_id`/`message_seq`/`channel_type`, the "跳转到原文" action's `jumpToCitation` returns without logging, Toast, or UI feedback. The broader Round-1 silent-no-op is resolved (a click now opens the details popover; navigation happens only from the explicit action), but this narrower missing-field branch still no-ops silently. Toast on the missing-field branch, or disable the action when the citation cannot be located.

- **Citation display-index threading end-to-end** — `remarkCitation` reading-order rank → `CitationBadge.JumpLink` → `ShowConversationOptions.initLocateCitationDisplayIndex` → `EndpointCommon` route key (`${key}-${seq}-${displayIndex || 0}`) → `Conversation` / `FoldSessionExpandedList` marker rendering; group jumps carry the display number into timeline labels, the marker, and the jump payload.
- **`mergeGroupMessages` behavior-invariant** — 12/12 verbatim base↔head edge-case fixtures are output-identical; dedup-by-seq, cited-wins tiebreak, seq-then-time sort confirmed equivalent to the original.
- **Citation context restored** — `context_before`/`context_after` render again in the timeline popover (Round-1 silent-drop regression resolved); citation click opens the details popover and is no longer a silent no-op.
- **i18n parity** — all 12 new keys present in `en-US`/`zh-CN` with identical `{{placeholder}}` sets; all 127 referenced keys resolve; `i18n:check` semantics hold.
- **Route key** — backward-compatible `|| 0`; existing callers get a stable suffix, distinct citation# ⇒ distinct key (forces remount when navigating between different citations of the same message).
- **Event wiring** — producer/consumer literals byte-identical; downstream reset (`startNewAgentConversation` clears both localStorage keys + resets state, `else-if` ordering after `derivedFromTask` ensures mutual exclusion) is unit-tested.
- **XSS / sanitize** — `SummaryMarkdownReader` delegates to `CitationText` (ReactMarkdown + rehype-sanitize `citationSchema` allowlist) on all three render paths; no `dangerouslySetInnerHTML`; heading slugs strip non-alphanumeric via `/[^\p{Letter}\p{Number}]+/gu` and `CSS.escape()` is used in all `querySelector` calls, so `javascript:`/`data:` schemes cannot survive.
- **Copy Markdown** — copies only the already-displayed summary content (no credential/PII leak); the failure path surfaces `Toast.error` instead of swallowing.
- **Marker anchoring** — `position:relative` present on both `.wk-message-item` and `.wk-fold-msg`; prop chain complete. `hidePlainCitations` privacy gate preserved in all 3 render sites.
- **Security overall** — no new attack surface, no credential leakage, no auth-boundary changes. One informational hardening note: put `id={id}` after `{...props}` at `CitationText.tsx:221` so the computed id can't be overridden if the sanitize schema ever allows `id`.

## Verification Results

| Code Review | NEEDS_CHANGES | 1 P1 (pre-existing CI gap, non-blocking) + 8 P2 + 7 P3 (deduped across 3 reviewers); confirmed must-fix: outline↔heading-id desync (glm5.2 + qwen3.8, executable-confirmed); Round-1 context-drop & silent-no-op regressions resolved |
| E2E Test | PARTIAL | Unit suites (2,554+ tests) + build + i18n parity PASS across reviewers; full-stack browser E2E of summary/citation flows BLOCKED (frontend-only repo; octo-server + MySQL/Redis/WuKongIM absent from isolated checkout); outline-desync P2 confirmed via no-mock real-renderer probes |
| Security | CLEARED | All 3 reviewers: ReactMarkdown + rehype-sanitize allowlist on all render paths; no dangerouslySetInnerHTML; heading slugs + CSS.escape can't carry javascript:/data:; clipboard copies only displayed content; no secrets/credential/auth-boundary changes; 1 informational hardening note |

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1039 (octo-web)

## Summary

This PR introduces a new `SummaryMarkdownReader` component that replaces direct `CitationText` usage across the summary detail page. The reader adds a sticky outline sidebar (table-of-contents with scroll-tracking), unified markdown typography, and a "Copy Markdown" / "New Agent chat" action bar for agent-triggered summaries. It also adds citation-marker overlays (`[n]` badges) in the conversation fold-list when jumping from a citation to its source message, wiring the display index through `ShowConversationOptions` → `ChatContentPage` → `Conversation` → `FoldSessionExpandedList`.

The component swap is clean — all `CitationText` props are forwarded through the new reader, and the existing `hidePlainCitations` privacy control is preserved. Two P2 issues found: a CSS nesting bug and a heading-ID desync edge case.

## Verification

- ✅ **Component swap parity** — all 10 `CitationText` call sites in `SummaryDetailPage.tsx` are replaced with `SummaryMarkdownReader`, each passing the correct combination of `showOutline`, `embedded`, and `outlineLabel`. Streaming, version history, personal result, team result, and participant report views are all covered.
- ✅ **Citation marker data flow** — `displayIndex` flows from `CitationBadge` → `ShowConversationOptions.initLocateCitationDisplayIndex` → `EndpointCommon.tsx` (key + prop) → `ChatContentPage` → `Conversation` → `FoldSessionExpandedList` + inline marker. Display indices start at 1 (from `buildDisplayIndexMap`), so the truthiness guard `locateCitationDisplayIndex &&` is safe.
- ✅ **Privacy control preserved** — `hidePlainCitations` is forwarded from SummaryMarkdownReader to CitationText in all team/view-other-person paths. The `disableTeamMemberPreview` flag for historical versions is also forwarded.
- ✅ **i18n parity** — en-US and zh-CN both add the same 12 new keys (`copyMarkdown`, `markdownCopied`, `newAgentChat`, `outline`, plus 8 citation popover keys). No missing translations.
- ✅ **Session cleanup on new agent chat** — `SummaryCreatePage` mount handler with `startNewAgentConversation` calls `clearAgentChatSession` + `clearAgentChatReferenced` before resetting state. Test at `SummaryCreatePage.test.tsx:619` verifies localStorage is cleared and state is reset.
- ✅ **Test coverage** — new tests cover outline extraction (including duplicate heading dedup and fenced-code exclusion), markdown rendering with heading-to-outline linking, citation popover grouped-message labeling, and jump-to-original with display index.
- ✅ **Event listener wiring** — `summary-open-new-agent-chat` event dispatched from `handleNewAgentChat` is caught in `module.tsx:177` and pushes a new `SummaryCreatePage` with `startNewAgentConversation`.

Static analysis only at head `e6b0b14d`; build and tests not executed in this environment.

## Findings

### P2 — CSS nesting bug leaves unintended border-radius (`packages/dmworksummary/src/index.css:672`)

The rule `.summary-detail-result .summary-markdown-reader__content { border-radius: 0; }` was inserted inside the `.summary-detail-result-footer` block without closing the footer rule first. The missing `}` at line 671 means the browser interprets this as a nested descendant selector: `.summary-detail-result-footer .summary-detail-result .summary-markdown-reader__content`. Since `.summary-detail-result` is an ancestor of `.summary-detail-result-footer` (not a descendant), this selector never matches. The content card retains `border-radius: var(--wk-r-md)` from the base `.summary-markdown-reader__content` rule, creating a visible rounded gap between the content and the flat-edged header (top-radius only) and footer (bottom-radius only).

```css
/* Current (broken) — line 670-675 */
    background: var(--wk-bg-surface);

.summary-detail-result .summary-markdown-reader__content {
    border-radius: 0;
}
}
```

```css
/* Fix — close footer before standalone rule */
    background: var(--wk-bg-surface);
}

.summary-detail-result .summary-markdown-reader__content {
    border-radius: 0;
}
```

### P2 — headingIds desync with blockquote-wrapped headings (`packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:9`)

`extractSummaryOutline` scans raw markdown line-by-line with `/^(#{2,3})\s+(.+?)\s*#*\s*$/`, which only matches headings at the start of a line. Headings inside blockquotes (`> ## Heading`) are skipped because the line starts with `>`. However, `markdownComponents` in `CitationText.tsx:220` assigns IDs by incrementing `headingIndex` for every h2/h3 that ReactMarkdown renders — and ReactMarkdown does render `> ## Heading` as `<blockquote><h2>Heading</h2></blockquote>`, consuming a heading slot. If LLM-generated summary content contains blockquote-wrapped headings, the counter outruns the outline array, and every subsequent heading receives the wrong `id`. Outline clicks then scroll to the wrong section.

This is unlikely in the typical case (LLM summaries rarely nest H2/H3 inside blockquotes), but the two code paths share no AST, so they can silently diverge. The safest fix is to derive outline items from the same remark AST that ReactMarkdown renders — e.g. run a `visit(tree, ['h2','h3'], …)` pass and share the result with `markdownComponents`.

- `useHoverPin` timer cleanup: `useEffect(() => () => clearTimers(), [clearTimers])` properly clears both open and close timers on unmount. The pinned-vs-hovering state machine is consistent.
- `mergeGroupMessages` dedup: uses `message_seq` as primary key with sender+content+sent_at fallback. Cited messages win over context-only duplicates. Sort is stable on seq then timestamp.
- `CSS.escape()` for heading ID selectors: correctly handles IDs with special characters. The outline `querySelector` and `scrollIntoView` both use escaped selectors.
- Table wrap component adds `tabIndex={0}` and `role="region"` for keyboard-accessible horizontal scroll.
- Responsive breakpoints: outline hides below 1180px, content padding adjusts below 719px. `prefers-reduced-motion` disables smooth scroll on outline links.
- Fenced-code exclusion in outline: the ```` ``` ```` / `~~~` toggle correctly skips headings inside code blocks.
- `handleCopyMarkdown` has try/catch with user-visible error toast on clipboard failure.

## Data Flow Traceback

- **Citation display index** → `buildDisplayIndexMap` (citationFormat.ts:52) scans AST text nodes → `remarkCitation` assigns `displayindex` property → `CitationBadge` receives `displayIndex` prop → `JumpLink` passes to `ShowConversationOptions.initLocateCitationDisplayIndex` → `EndpointCommon.tsx:171` passes to `ChatContentPage` → `Conversation` passes to `FoldSessionExpandedList` and renders inline marker. ✅ Complete chain, no data loss.
- **Outline heading IDs** → `extractSummaryOutline` (index.tsx:9) parses raw markdown → IDs passed as `headingIds` to `CitationText` → `markdownComponents` uses `headingIndex++` counter. ⚠️ Diverges when content has blockquote-wrapped headings (see P2 above).
- **SummaryMarkdownReader content** → All call sites pass `detail.result.content` or `personalResult.content` (existing state fields, no new data sources). ✅ No data source change.

## Blind Spot Checklist

- **C1 (Dual-path parity)**: N/A — no paired add/remove or subscribe/unsubscribe paths changed.
- **C2 (Control-flow ordering)**: N/A — no logic reused across different ordering contexts.
- **C3 (Auth boundary)**: N/A — no auth/permission changes.
- **C4 (Auth lifecycle)**: N/A — no auth changes.
- **C5 (Build ≠ runtime)**: N/A — no build/packaging changes.
- **C6 (Governance/docs)**: N/A — no governance/doc changes.

---

[Octo-Q] verdict: APPROVE — No P1 blockers. Two P2 items (CSS nesting bug, headingIds desync) are non-blocking visual/correctness issues that don't brick the primary reading flow. Recommend fixing before merge but not blocking.

Jerry-Xin (dismissed):
The feature is relevant to octo-web, but the outline can navigate to the wrong heading for valid Markdown.

## 🔴 Blocking

- 🔴 Critical — Heading IDs are assigned using a different parser than the rendered Markdown. [SummaryMarkdownReader/index.tsx:15](/Users/mlamp/.openclaw/tmp/pr-review-aljf6qty/repo/packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:15) only recognizes top-level ATX headings, while [CitationText.tsx:220](/Users/mlamp/.openclaw/tmp/pr-review-aljf6qty/repo/packages/dmworksummary/src/components/CitationText.tsx:220) assigns IDs sequentially to every rendered H2/H3. Valid constructs such as `> ## Quoted`, indented/list headings, or setext H2 headings are rendered but absent from the extracted outline, shifting all subsequent IDs. For example, `> ## Quoted\n\n## Result` assigns `id="result"` to “Quoted” and leaves “Result” without an ID, so clicking “Result” scrolls to the wrong section. Generate the outline and heading IDs from the same Markdown AST, and add coverage for blockquote/setext headings.

## 💬 Non-blocking

- 🟡 Warning — The missing closing brace at [index.css:658](/Users/mlamp/.openclaw/tmp/pr-review-aljf6qty/repo/packages/dmworksummary/src/index.css:658) nests the selector at line 671 under `.summary-detail-result-footer`. It therefore cannot match the reader content, which is a footer sibling. Close the footer rule before declaring `.summary-detail-result .summary-markdown-reader__content`.

- 🟡 Warning — The “Open original” control is a clickable `<span>` at [CitationBadge.tsx:91](/Users/mlamp/.openclaw/tmp/pr-review-aljf6qty/repo/packages/dmworksummary/src/components/CitationBadge.tsx:91), making it inaccessible by keyboard. Use a semantic `<button type="button">`.

- 🟡 Warning — The new styles contain many hard-coded dimensions, spacing, and font sizes, contrary to the project’s token rules; examples include [CitationBadge.css:15](/Users/mlamp/.openclaw/tmp/pr-review-aljf6qty/repo/packages/dmworksummary/src/components/CitationBadge.css:15) and [SummaryMarkdownReader/index.css:19](/Users/mlamp/.openclaw/tmp/pr-review-aljf6qty/repo/packages/dmworksummary/src/ui/SummaryMarkdownReader/index.css:19).

## ✅ Highlights

- Citation display numbers are propagated through conversation routing.
- The reader includes duplicate-safe IDs, responsive tables, reduced-motion handling, tests, and a Storybook story.
- New Agent chat explicitly clears both the prior session and summary reference.

Jerry-Xin (changes requested):
The feature is relevant to octo-web, but the outline can navigate to the wrong heading for valid Markdown.

## 🔴 Blocking

- 🔴 Critical — Heading IDs are assigned using a different parser than the rendered Markdown. [SummaryMarkdownReader/index.tsx:15](packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:15) only recognizes top-level ATX headings, while [CitationText.tsx:220](packages/dmworksummary/src/components/CitationText.tsx:220) assigns IDs sequentially to every rendered H2/H3. Valid constructs such as `> ## Quoted`, indented/list headings, or setext H2 headings are rendered but absent from the extracted outline, shifting all subsequent IDs. For example, `> ## Quoted\n\n## Result` assigns `id="result"` to “Quoted” and leaves “Result” without an ID, so clicking “Result” scrolls to the wrong section. Generate the outline and heading IDs from the same Markdown AST, and add coverage for blockquote/setext headings.

## 💬 Non-blocking

- 🟡 Warning — The missing closing brace at [index.css:658](packages/dmworksummary/src/index.css:658) nests the selector at line 671 under `.summary-detail-result-footer`. It therefore cannot match the reader content, which is a footer sibling. Close the footer rule before declaring `.summary-detail-result .summary-markdown-reader__content`.

- 🟡 Warning — The “Open original” control is a clickable `<span>` at [CitationBadge.tsx:91](packages/dmworksummary/src/components/CitationBadge.tsx:91), making it inaccessible by keyboard. Use a semantic `<button type="button">`.

- 🟡 Warning — The new styles contain many hard-coded dimensions, spacing, and font sizes, contrary to the project’s token rules; examples include [CitationBadge.css:15](packages/dmworksummary/src/components/CitationBadge.css:15) and [SummaryMarkdownReader/index.css:19](packages/dmworksummary/src/ui/SummaryMarkdownReader/index.css:19).

## ✅ Highlights

- Citation display numbers are propagated through conversation routing.
- The reader includes duplicate-safe IDs, responsive tables, reduced-motion handling, tests, and a Storybook story.
- New Agent chat explicitly clears both the prior session and summary reference.

yujiawei (changes requested):
# Code Review — PR #1039 (octo-web)

Reviewed at head `e6b0b14db26c721186ba2309a93b87164f35db83`, diffed against merge-base `a8bad92`.

## Scope check

The diff stays within the stated scope: frontend presentation and navigation only (dmworksummary reader/citation UI, plus a citation-marker prop threaded through `dmworkbase` Conversation/Chat/EndpointCommon). No prompt, backend API, workflow, or derivation-model changes. New i18n keys are present in both `en-US.json` and `zh-CN.json`, and referenced keys (`summary.common.operationFailed`, the new `summary.citation.*` / `summary.detail.*`) all exist. Nothing required is missing and nothing out-of-scope was added.

## Verdict: Changes Requested

One functional correctness issue in the headline feature (outline navigation) should be fixed before merge. The rest are non-blocking suggestions.

---

## P1 — Blocking

### 1. Outline `id` assignment couples two heading parsers that can disagree → silent mis-navigation

`SummaryMarkdownReader` extracts the outline with a hand-rolled line parser and passes the resulting ids **positionally** to `CitationText`, which hands them out by increment order:

- `packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx:8-29` — `extractSummaryOutline` walks lines with `/^(#{2,3})\s+(.+?)\s*#*\s*$/`.
- `packages/dmworksummary/src/components/CitationText.tsx:220-224` — `const id = headingIds[headingIndex++]` on every rendered `h2`/`h3`.

The two must agree on the **count and order** of H2/H3, but ReactMarkdown (CommonMark) recognizes heading forms the line parser does not, so `headingIndex` drifts and every subsequent id shifts by one (the intended heading gets the wrong id; the last one gets `undefined`). Realistic divergences:

- **Setext H2**: `Some line\n---` (no blank line before `---`) is a CommonMark setext `<h2>`, but the extractor sees no `#` and skips it. (The reader's own CSS special-cases `hr + h2`, so `---` in LLM output is expected — this trap is live.)
- **Headings nested in blockquotes**: `> ## Foo` renders as `<h2>` (the custom `h2` component fires for nested headings too) but the `^(#{2,3})` anchor never matches the `>`-prefixed line.
- **Indented headings**: CommonMark allows up to 3 leading spaces (`  ## Foo`); the extractor's `^#` anchor rejects them.

Because there is no error, the failure mode is a Table-of-Contents link that scrolls to the wrong section (or a heading that receives no id). 

Recommended fix: derive the ids from the same AST the markdown is rendered from (e.g. a small rehype step / `rehype-slug`-style plugin that slugs each heading's text), instead of matching a separately-computed list by index. That removes the two-parser coupling entirely.

Note on an alternative mechanism raised during review: `headingIndex` is **not** corrupted by React StrictMode double-rendering — `markdownComponents(...)` is invoked fresh inside each render (`CitationText.tsx:316`), so the counter is a render-local closure that resets every pass. The real cause is the parser divergence above, not double-invocation.

---

## P2 — Non-blocking

### 2. Duplicate-anchor suffixing can still collide with an authored suffix

`SummaryMarkdownReader/index.tsx:23-26`: given `## Foo`, `## Foo`, `## Foo-2`, the extractor emits ids `foo`, `foo-2`, `foo-2` — the third heading's slug equals the second's de-duplicated slug, so two DOM elements share `id="foo-2"` and `querySelector('#foo-2')` resolves both outline links to the first match. Adversarial input, but the de-dup scheme doesn't reserve slugs globally. A global "seen ids" set (bumping until unique) closes it.

### 3. Misplaced brace nests a rule that was meant to be top-level

`packages/dmworksummary/src/index.css:658-674`: the `.summary-detail-result-footer` block's closing `}` lands *after* an inner `.summary-detail-result .summary-markdown-reader__content { border-radius: 0 }` rule, so that rule becomes a nested descendant of `.summary-detail-result-footer` (relying on native CSS nesting) rather than a standalone selector. Effect is cosmetic (the intended `border-radius: 0` override never applies to the reader content), but it looks like an accidental brace and should be lifted out.

### 4. Timeline filter and navigation operate on different lists

`packages/dmworksummary/src/components/CitationBadge.tsx:110-119` (`CitationTimeline`): the search/`citedOnly` filter drives `visibleMessages`, but `activeMessage`, the up/down buttons, and the position counter iterate the unfiltered `cited` array. After filtering, stepping with the arrows can land on a hidden message — the `is-current` highlight disappears and the footer JumpLink targets a row the user can't see. Consider deriving the active index from the visible (cited ∩ filter) set.

### 5. Scroll-spy reads layout in a loop on every scroll event

`SummaryMarkdownReader/index.tsx:54-63`: `updateActive` calls `getBoundingClientRect()` for every outline heading synchronously per `scroll` event. On long documents this can cause jank; an `IntersectionObserver` (or a rAF throttle) would be smoother. Pure polish.

---

## Notes / blind spots (not verified here)

- I could not run the build or test suite in this environment; validation relies on the checks reported in the PR description.
- The PR currently reports a **merge conflict with `main`** — it needs a rebase before merge, and post-rebase the outline/citation behavior should be re-confirmed against the new base.
- The cross-channel `seq:` dedup concern that can arise in `mergeGroupMessages` does **not** apply here: a `citationgroup` is only formed when adjacent citations share the same `channel_id` (`CitationText.tsx:98-101`), so a single popover never mixes channels. The dedup key is safe under that invariant.
- `handleCopyMarkdown` relies on `navigator.clipboard`; it has a try/catch that surfaces a failure toast, so non-secure-context failures degrade gracefully.

packages/dmworksummary/src/ui/SummaryMarkdownReader/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+export interface SummaryOutlineItem { id: string; level: 2 | 3; text: string }
+
+export function extractSummaryOutline(markdown: string): SummaryOutlineItem[] {
+    const counts = new Map<string, number>();
+    const result: SummaryOutlineItem[] = [];
+    let fenced = false;
+    for (const line of markdown.split(/\r?\n/)) {
+        if (/^\s*(```|~~~)/.test(line)) { fenced = !fenced; continue; }
+        if (fenced) continue;
+        const match = /^(#{2,3})\s+(.+?)\s*#*\s*$/.exec(line);
```
lml2468, line 15: 🔴 目录 id 由这条**手写行正则**生成:`/^(#{2,3})\s+.../` —— 只认行首 ATX `## `/`### `。但 `CitationText`(index.tsx:222)是按顺序把 `headingIds[headingIndex++]` 赋给 **react-markdown 渲染出的每一个 H2/H3**。两个解析器在以下情况分歧 → 序号错位:
① **Setext 标题**(`标题\n---`)react-markdown 渲染成 `<h2>`,此正则不匹配 → react-markdown 多出一个 heading,`headingIndex` 提前自增 → 之后所有 id 整体错位;
② **blockquote 内标题**(`> ## X`)react-markdown 渲染 `<h2>`,此正则(要求 `#` 行首)不匹配 → 同样错位。
后果:目录点击 `querySelector('#'+item.id)` 跳到**错误标题**;当 react-markdown 的 heading 数 > outline 数时,`headingIds[i]` 越界为 `undefined` → 该标题无 id → 跳转失败。修法:目录 id 与渲染 heading id 从**同一个 AST** 生成(共享 rehype-slug 或单次 remark/rehype 解析),并覆盖 Setext / blockquote。

packages/dmworksummary/src/components/CitationText.tsx
```diff
@@ -211,8 +215,16 @@ function markdownComponents(
     teamCitations: TeamCitationItem[],
     members: MemberStatus[],
     disableTeamMemberPreview: boolean,
+    headingIds: string[] = [],
 ): any {
+    let headingIndex = 0;
+    const heading = (Tag: 'h2' | 'h3') => ({ children, ...props }: any) => {
+        const id = headingIds[headingIndex++];
```
lml2468, line 222: 🔴(同根因)`headingIds[headingIndex++]` 按渲染顺序位置耦合 outline 数组 —— 一旦 `extractSummaryOutline`(正则 ATX-only)与 react-markdown(含 Setext/blockquote)对「哪些是 H2/H3」判断不一致,id 就整体错位或越界为 `undefined`。应改由渲染侧 AST 自己派生稳定 id(rehype-slug),outline 复用同一份 id,而不是两处各自解析后按序号对齐。

lml2468 (changes requested):
## Re-review: #1039 @ `e6b0b14d` — Phase 1 markdown reader + citation markers

复核当前 live head `e6b0b14d` / `merge:CONFLICTING`。21 文件 +794/−155。

**结论:需修改 — REQUEST_CHANGES** ❌(上轮的目录对齐 🔴 仍未修 + 与 main 冲突)

---

### 🔴 阻塞 1 — 目录与渲染用两套解析器,heading id 错位(功能回归)
- 目录 id 由 `extractSummaryOutline`(`SummaryMarkdownReader/index.tsx:15`)的**手写行正则** `/^(#{2,3})\s+.../` 生成:**只认行首 ATX `##`/`###`**(带 fenced-code 跳过)。
- 而 `CitationText`(`CitationText.tsx:222`)按顺序把 `headingIds[headingIndex++]` 赋给 **react-markdown 渲染出的每一个 H2/H3**。
- **两者分歧即错位**(可复现):
  - **Setext 标题**(`标题\n---`)→ react-markdown 渲染 `<h2>`,正则不匹配 → react-markdown 多出一个 heading,`headingIndex` 提前自增 → 其后所有 id 整体错位;
  - **blockquote 内标题**(`> ## X`)→ react-markdown 渲染 `<h2>`,正则(要求 `#` 行首)不匹配 → 同样错位。
- **后果**:目录点击 `root.querySelector('#'+item.id)` 跳到**错误标题**;当渲染 heading 数 > outline 数时 `headingIds[i]` 越界为 `undefined` → 该标题无 id → 跳转失败。本 PR 新引入(整套 reader+outline+headingIds 是 Phase 1 新增),有具体触发(含 Setext/blockquote 的总结)、具体后果、可行修法。
- **修法**:目录 id 与渲染 heading id 从**同一个 AST** 派生(共享 `rehype-slug` 或单次 remark/rehype 解析后两处复用同一 id 表),并覆盖 Setext / blockquote;不要两处各自解析再按序号对齐。

### 🔴 阻塞 2 — 与 main 冲突
`merge:CONFLICTING` —— 需 rebase 解冲突后才能合。

### ✅ 安全面(高危重点,复核通过)
markdown 渲染无 XSS:`react-markdown` + `rehypeSanitize`(defaultSchema),**未启用 `rehype-raw`/`allowDangerousHtml`/`dangerouslySetInnerHTML`**;引用/团队引用走 remark AST 文本节点转受控节点再映射 React 组件,不拼 HTML 串;默认协议白名单挡 `javascript:`/`data:` 链接。

### 构建
`pnpm --filter @octo/web build` ✅ 3.86s。

---

### 🟡 非阻断(peers)
- `index.css` 疑似缺闭合花括号致 reader selector 错误嵌套、样式不生效(yujiawei/mochashanyao);
- 「Open original」用可点击 `<span>`,键盘不可达 → 应改 `<button>`(Jerry-Xin);
- 新样式大量硬编码尺寸/字号,违背设计 token 规范(Jerry-Xin);
- 重复锚点后缀(`-2`)可能与用户自定义标题冲突,建议全局唯一化(yujiawei)。

---

安全面(rehype-sanitize、无 raw)通过、构建通过;但目录与渲染两套解析器致 heading id 错位(Setext/blockquote 触发目录跳错/跳空)这条上轮 🔴 仍在,加上与 main 冲突。**需修改 —— REQUEST_CHANGES**(核心修法:outline 与渲染共用同一 AST 派生 id)。

mochashanyao (approved):
[Octo-Q · automated review]

---

[Octo-Q] verdict: APPROVE — No P0/P1 blockers. Two P2 items (persistent citation marker, HMR listener cleanup) and one nit are non-blocking. Data flow verified end-to-end across all 23 changed files.

wanyiwang06:
Closing this PR because the final summary UI work continued on a separate branch/worktree. A clean replacement PR based on the complete, locally validated implementation will follow.

lml2468 (changes requested):
## Re-review: #1039 — Phase 1 markdown reader + citation markers @ `6d31667a`

复核当前 live head `6d31667aba87`(rev-parse 确认 checkout==live head;审查期间 head 由 `b0550707` 前进——多为合入 main 的 622 文件漂移)/ `mergeable=true`。

**结论:REQUEST_CHANGES 🔴 —— 我上轮的 heading-id 两解析器错位 🔴 仍未修(逐字节仍在);上轮的 merge 冲突 🔴 已随合入 main 解除。**

### ✅ 已解除:上轮 🔴#2(与 main 冲突)
`mergeable=true`(经合入 main 解决)。不再阻塞。

### 🔴 仍在:heading-id 两解析器错位(功能回归,未修)
`ui/SummaryMarkdownReader/index.tsx:15` + `components/CitationText.tsx:222`
- **目录**用手写行正则 `/^(#{2,3})\s+.../`(`extractSummaryOutline`,index.tsx:15)——**只认行首 ATX `##`/`###`**(带 fenced 跳过)。
- **渲染**侧 `CitationText.tsx:222` 仍 `const id = headingIds[headingIndex++]`——把 outline 的 id **按序**赋给 react-markdown 渲染出的每一个 H2/H3。
- **两者分歧即错位(机制由构造确定)**:react-markdown(+remark-gfm)会把 **Setext 标题**(`标题\n---`)、**blockquote 内标题**(`> ## X`)渲染成 `<h2>`,而上面的正则**不匹配**这些 → 渲染 heading 数 > outline 数 → `headingIds[i]` 顺序整体前移错位、越界项为 `undefined`(该标题无 id)。→ 目录点击 `querySelector('#'+id)` 跳到**错误标题**或跳空。
- **触发**:任何含非 ATX 标题(Setext/blockquote)的总结正文;summary 为 worker/LLM 生成的 markdown,非 ATX 标题**可能出现**(human-verify:确认 worker 输出是否严格 ATX-only)。
- **测试掩盖**:`SummaryMarkdownReader.test.tsx` 的 CitationText mock(:4-8)是**从 `extractSummaryOutline` 自身**渲染 `<h2 id={headingIds[index]}>`——outline 与 heading 同源、恒对齐,**结构上无法捕捉真实 react-markdown 渲染出的额外/错位标题**。现有 ATX-only + fenced 用例也未覆盖 Setext/blockquote。
- **修法**:outline id 与渲染 heading id 从**同一 AST 派生**(共享 `rehype-slug`,或单次解析后两处复用同一 id 表),覆盖 Setext/blockquote;不要两处各自解析再按序号对齐。补 Setext/blockquote 用例。
- (说明:本轮我试图渲染级复现,被 summary 包测试环境一处无关的 `react/jsx-runtime` 解析报错挡住;上述错位机制由**代码审查确证**——正则 :15 与顺序赋值 :222 在历次 rework/merge 中均未变。)

### 本次新增(citation 归一化)—— 安全、可接受
`normalizeCitationBrackets`(citationFormat.ts)把 `【1】`/`［P2］` 等全角括号引用**窄化**归一为 `[1]`/`[P2]`(仅 `P?\d{1,3}` token,普通中文括号散文不动),渲染时兜底修旧总结。无 XSS(纯字符串归一后仍走同一 remark citation 插件)。`CitationBadge.test` 17/17。安全面(rehype-sanitize、无 raw/allowDangerousHtml)沿用无回归。

---

heading-id 两解析器错位 🔴(Setext/blockquote → 目录跳错/跳空)自我上轮起**仍未修**(正则 :15 + 顺序赋值 :222 未变,测试因 mock 同源而无法捕捉);merge 冲突 🔴 已随合入 main 解除;citation 全角归一化是安全的合理新增。**需修改 —— REQUEST_CHANGES**(核心:outline 与渲染共用同一 AST 派生 id 并覆盖 Setext/blockquote;若确认 worker 仅产 ATX 标题,请在 PR 说明该约束并补断言以固化)。

yujiawei (changes requested):
# Code Review — PR #1039 (octo-web)

Reviewed at live head `6d31667aba8721df15b77cbfbb8247f9f327dc49` (the dispatch referenced `b0550707`; head advanced during review, so this verdict is against `6d31667a`). Merge base `2d087f0bb9a4` == `origin/main`, so the diff below is the full net change vs main. `mergeable=true`, `Build SUCCESS`, `e2e-p0 SUCCESS`.

Nice work on the reader itself — the CSS token discipline, the `:has()`-scoped card teardown, and the consolidation of four duplicated pinned-popover layouts into one `CitationTimeline` are all real improvements. Two blocking issues below, both in the new outline path.

## 1. Spec compliance

**Spec: ❌**

Measured against the PR description's own five scope bullets.

- **Missing: none.** All five stated bullets are delivered: the shared reader (`ui/SummaryMarkdownReader/index.tsx`), outline nav, table scroll region (`CitationText.tsx:265-269`), the grouped citation timeline (`CitationBadge.tsx:110-166`), cited-message markers (`Conversation/index.tsx:2095-2097`, `FoldSessionExpandedList.tsx:82-84`), render-time full-width normalization (`citationFormat.ts:22-27`), and every one of the 8 `CitationText` call sites in `SummaryDetailPage.tsx` migrated to the new reader.
- **Out of scope (over-build): an entire second feature landed unreachable.** None of it is in the stated scope, and none of it has a UI entry point:
  - `SummaryDetailPage.tsx:1973` `handleCopyMarkdown` — defined, never referenced. `grep -n handleCopyMarkdown SummaryDetailPage.tsx` returns only the definition.
  - `SummaryDetailPage.tsx:1984` `handleNewAgentChat` — same, definition only.
  - `SummaryDetailPage.tsx:13` imports `IconArrowLeft` and `IconCopy`; `:16` imports `SummaryCardContent`. All three are unused (verified repo-wide).
  - `module.tsx:176-178` registers a **permanent global `window` listener** for `summary-open-new-agent-chat`, whose only dispatcher is the dead `handleNewAgentChat`.
  - `SummaryCreatePage.tsx:53` adds a new public prop `startNewAgentConversation` plus a state branch at `:314-323`, reachable only through that dead listener (it does have a test at `SummaryCreatePage.test.tsx:619-630` — a tested-but-unreachable path).
  - 3 new i18n keys × 2 locales (`copyMarkdown`, `markdownCopied`, `newAgentChat`) with no consumer.

  I checked `origin/main` for all of these symbols — none exist there, so this is new dead code, not a feature dropped during the conflict resolution. Please either wire the action-menu entry points in this PR or strip the handlers, the imports, the module listener, the prop, and the i18n keys. Shipping a permanently-registered global event listener for a feature with no trigger is the part I'd most like removed.
- **Divergence: one.** The description says citation normalization "is presentation-only and does not mutate stored summary content." True for storage, but it also rewrites content it was never meant to touch — see P2-1.

`noUnusedLocals` is `false` in `packages/tsconfig/base.json:14`, which is why the unused imports do not fail the build. The build passing is not evidence this code is reachable.

## 2. Code quality

### P1-1 — `extractSummaryOutline` regex backtracks quadratically on stored content (`ui/SummaryMarkdownReader/index.tsx:15`)

```js
const match = /^(#{2,3})\s+(.+?)\s*#*\s*$/.exec(line);
```

The lazy `(.+?)` followed by the ambiguous `\s*#*\s*$` tail makes the number of partitions quadratic in the length of a whitespace run. Measured on this machine (node, single `exec`):

| input line | length | time |
|---|---|---|
| `"## " + "a".repeat(5000)` | 5,003 | 0.1 ms |
| `"## title" + " ".repeat(2000)` | 2,008 | 0.0 ms |
| `"## title" + " ".repeat(2000) + "x"` | 2,009 | **2,403 ms** |
| `"## t" + " "*1000 + "#"*1000 + " "*1000 + "x"` | 3,005 | **632 ms** |

The trigger is narrow but not exotic: a `##`/`###` line containing a long whitespace run followed by any non-space, non-`#` character. It scales quadratically — ~4,000 spaces is ~10 s, ~10,000 is ~60 s.

Why this is blocking rather than theoretical:
- `extractSummaryOutline` is on the render path of **every** summary view via `index.tsx:45`, including `showOutline={false}` and `embedded` ones, so it also runs for streaming views, version history, and each expanded participant report.
- During streaming, `content` changes per chunk, so the `useMemo` key changes per chunk and the whole document is re-scanned every chunk — the pathological line is paid again on each one.
- Summary content is stored, and it is user-editable through `SummaryEditor`. One saved summary containing such a line freezes the main thread for every viewer who opens it.

Fix — drop the ambiguity, don't try to tune it:

```js
const match = /^(#{2,3})\s+(.+)$/.exec(line);
if (!match) continue;
const raw = match[2].replace(/\s+#*\s*$/, "");   // or: trimEnd().replace(/#+$/, "").trimEnd()
```

Please add a regression test asserting a bounded runtime for a heading line with a few thousand spaces.

### P1-2 — Outline ids are assigned positionally by a counter mutated during render (`CitationText.tsx:220-224` + `ui/SummaryMarkdownReader/index.tsx:15`)

```js
let headingIndex = 0;
const heading = (Tag: 'h2' | 'h3') => ({ children, ...props }: any) => {
    const id = headingIds[headingIndex++];
    return <Tag id={id} {...props}>{children}</Tag>;
};
```

Two independent failure modes, one mechanism:

**(a) Two different parsers must agree on the heading sequence, and they don't.** The ids come from a hand-rolled line scanner (`index.tsx:15`, ATX-at-column-0 only), while the consumers are the `<h2>`/`<h3>` elements react-markdown actually emits. remark renders several heading forms the scanner cannot see, so the sequences drift and every subsequent id shifts:

- Setext: `Some text\n---` is an `<h2>` per CommonMark whenever there is no blank line before `---`. The scanner misses it.
- Blockquoted: `> ## Title` → `<h2>`. Missed.
- Indented ATX: up to 3 leading spaces is still a valid heading. `index.tsx:15` anchors `#` at column 0. Missed.
- Headings whose text reduces to empty after markdown stripping are dropped by `index.tsx:22` but still rendered by remark.

Concretely, for `## Overview\nSome text\n---\n## Details`: remark emits three `<h2>`s (`Overview`, `Some text`, `Details`); `headingIds` has two entries. `Overview` gets `overview`, `Some text` gets `details`, the real `Details` gets `id={undefined}`. The outline's "Details" link then scrolls to the wrong paragraph, and the heading it names is unreachable.

**(b) The counter is mutated during render, which React does not guarantee.** `apps/web` runs React 18.3.1 under `createRoot` + `React.StrictMode` (`apps/web/src/index.tsx:133-139`). StrictMode double-invokes function component bodies in development, and the two invocations of each `h2`/`h3` component share the same `headingIndex` closure, so the counter advances twice per heading and ids land on the wrong elements — in development the outline is misaligned for *all* headings, not just after a drift point. Tests do not catch this because `@testing-library/react` renders without StrictMode.

Fix: derive both sides from one source. Either add `rehype-slug` and have `extractSummaryOutline` mirror its slugs, or parse once and pass a content-addressed `Map<headingText, id>` instead of a positional array — anything that removes "Nth outline entry must equal Nth rendered heading" as an invariant, and removes the render-phase mutation.

Note: this is the same defect a previous round flagged and it is byte-for-byte unchanged at this head. See §5 on rounds.

### P1-3 — The test that appears to cover P1-2 is tautological (`ui/SummaryMarkdownReader/SummaryMarkdownReader.test.tsx:4-11`)

```js
vi.mock("../../components/CitationText", () => ({
    default: ({ content, headingIds }) => <div>
        {extractSummaryOutline(content).map((item, index) => item.level === 2
            ? <h2 id={headingIds[index]} key={item.id}>{item.text}</h2>
            : ...
```

The mock generates the headings from `extractSummaryOutline` — the very function that produced `headingIds`. Alignment holds by construction, so `expect(getByRole("heading", { name: "Result" })).toHaveAttribute("id", "result")` can never fail, no matter how far the real react-markdown output drifts. The `[x]` → checkbox assertion in the same test is also asserting on the mock, not on markdown rendering. This test currently provides negative value: it reads as coverage of the exact invariant that is broken.

The alignment test has to render the real `CitationText` and assert ids against Setext / blockquoted / indented-ATX fixtures.

### P2-1 — Normalization runs on the raw string, so it rewrites text inside code (`CitationText.tsx:296`, `citationFormat.ts:22-27`)

`normalizeCitationBrackets(content)` is applied before parsing, so `【1】` inside a fenced block, an inline code span, or a link destination is silently rewritten to `[1]`. That directly contradicts the invariant documented 8 lines below it at `CitationText.tsx:303-307`, which explains that numbering is derived from AST text nodes precisely so code spans are excluded (#1003). Displayed code is corrupted; storage is fine. Move the normalization into the remark pass over `text` nodes (where `remarkCitation` already operates) instead of the raw string. It is also unmemoized on a hot path — it re-scans the whole document on every `activeKey` change, i.e. every badge hover/click.

### P2-2 — Keyboard dismissal and badge semantics regressed (`CitationBadge.tsx:254-264`, `:316-324`)

The document-level Escape handler was deleted from both `CitationBadge` and `CitationGroupBadge`; its comment said explicitly "works regardless of focus (e.g. after clicking jump link)". What remains is `onKeyDown` on the `<sup>` itself (`:260-263`), which only fires while the badge has focus. This regression got *more* likely to bite in this PR, because the pinned popover now contains focusable content — a search input at `:141` and buttons at `:138`/`:142`. Once focus moves into the popover, Escape no longer closes it.

Also removed from the same elements: `role="button"` and `aria-expanded={visible}`. The badge is still `tabIndex={0}` with a click handler, so it is now a focusable element that screen readers announce as plain text with no expanded state. Please restore all three.

### P2-3 — Group popover title renders non-contiguous indices as a range (`CitationBadge.tsx:135`)

```js
t("summary.citation.title", { values: { indices: Array.from(displayByRawIndex.values()).join("–") } })
```

An en-dash join over all values renders `[1–5–9]` for a group of citations 1, 5, 9, which reads as a range and disagrees with the badge the user just clicked. `formatGroupLabel` (`citationFormat.ts:36-58`) already implements the product rule for this — including the deliberate "fall back to a comma list when not contiguous ascending" logic added in a previous #1003 review — and is already imported in this file at `:23`. Reuse it here.

### P2-4 — Outline scroll tracking: unguarded, unthrottled, and cubic (`ui/SummaryMarkdownReader/index.tsx:49-67`, `:82`)

- The effect has no `showOutline` guard, so streaming views, version-history sections, and every expanded participant report each attach their own `scroll` listener to the same `.summary-detail-content-scroll`.
- `updateActive` is unthrottled (no rAF, no debounce) and, per scroll event, runs one `querySelector` + `getBoundingClientRect()` per outline entry (`:59-60`) — N forced layouts per event, per mounted reader.
- The H2 numbering at `:82` is O(n³): for each item it filters the whole outline and calls `indexOf` twice per candidate. Measured: 400 headings → 12 ms, 800 → 82 ms, 1600 → 599 ms, and it recomputes on every `activeId` change. Negligible at 20–100 headings, so I am not treating this as blocking, but it is a one-line fix to a running counter and it compounds with the two points above.

### P2-5 — Nested rule inside a declaration block; the intended override never applies (`packages/dmworksummary/src/index.css:1169-1185`)

```css
.summary-detail-result-footer {
    ...
    background: var(--wk-bg-surface);

.summary-detail-result .summary-markdown-reader__content {
    border-radius: 0;
}
}
```

A rule opened inside the footer's declaration block, with a matching extra `}`. Under native CSS nesting this resolves relative to `.summary-detail-result-footer`, which is never an ancestor of the reader content; under a parser without nesting it is discarded as an invalid declaration. Either way `border-radius: 0` never lands, so the header/content/footer card seam keeps the reader's own rounding. This one is invisible in a screenshot review — worth a quick visual check on the result card.

### P2-6 — Minor

- Heading ids are injected even when `showOutline={false}`, so pages that render several readers over similar content (participant reports, version history) emit duplicate DOM ids. The outline's own lookups are correctly scoped to `rootRef`, so this is invalid-HTML hygiene rather than a functional bug.
- `CitationText.tsx:266` — `role="region"` with no accessible name; add `aria-label`. Same file, `CitationBadge.tsx:141` — the search input has a placeholder but no label.
- `Conversation/index.tsx:2095` and `FoldSessionExpandedList.tsx:82` gate on truthiness, so a display index of `0` would not render. Display indices are 1-based by construction (`buildDisplayIndexMap`), so this is not a live bug — `!= null` is still the more honest check. Note also that a message present in both the main list and a fold-expanded list can render the marker twice, and the marker never clears for the lifetime of the mounted chat page (unlike the existing locate highlight, which has `onLocateAnimationEnd`).
- `CitationBadge.tsx:147` — `key={msg.message_seq ?? i}` mixes a seq keyspace with a filtered-array-index keyspace, so keys collide and shift while the user types in the new search box. Pre-existing pattern, newly exposed to filtering.
- `CitationText.tsx:50-51` adds `align` to the `th`/`td` sanitize allowlist, but the new test at `CitationText.test.tsx:97` asserts alignment arrives as `text-align: center` (a `style`, not an `align` attribute). Please confirm those two allowlist entries do anything; if alignment comes through as `style`, they are dead and the real question is which rule lets the style survive `rehype-sanitize`.
- `CitationBadge.tsx:145-166` — the ↑/↓ reference nav moves `current` and the jump-link target but never scrolls the message list, and `current` indexes `cited` while the list renders `visibleMessages`. With the search or "citations only" filter active, pressing ↓ can highlight an item that is not on screen and nothing visibly happens.
- Real avatars in citation popovers are gone (`WKAvatar` / `MessageAvatar` removed, `sender_uid` dropped from `MergedMessage`) in favour of a first-letter span. If that is the intended design, fine — flagging it because it is a user-visible downgrade not mentioned in the description.

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ❌ (an unreachable second feature plus a permanently-registered global listener) AND Quality Changes-Requested (P1-1 stored-content main-thread freeze, P1-2 outline ids misassigned, P1-3 the covering test cannot fail).

## 4. Recommended order of work

1. `index.tsx:15` — replace the regex with the non-ambiguous form; add a bounded-runtime test.
2. `CitationText.tsx:220-224` — derive outline ids and rendered heading ids from one AST (`rehype-slug`, or a content-addressed map); delete the render-phase counter.
3. `SummaryMarkdownReader.test.tsx` — stop mocking `CitationText` for the alignment test; assert real ids against Setext / blockquote / indented-ATX / duplicate-title fixtures.
4. Decide on Copy Markdown / New Agent chat: wire the entry points, or remove the handlers, the three unused imports, the `module.tsx` listener, the `SummaryCreatePage` prop, and the six i18n keys.
5. Move `normalizeCitationBrackets` into the remark text-node pass; memoize it.
6. Restore `role="button"`, `aria-expanded`, and document-level Escape.
7. `CitationBadge.tsx:135` → `formatGroupLabel`. `index.css:1182` → unnest. `index.tsx:49` → guard on `showOutline` and rAF-throttle; `:82` → running counter.

## 5. Additional notes

**Round count.** This is round 5+ (11 reviews on the PR). P1-2 was reported in an earlier round against an earlier head and is unchanged here; P1-1, P1-3, and the dead-code findings are new to this pass, which means the surface is still growing rather than converging. Recommendation beyond the verdict: stop patching finding-by-finding. The outline id mechanism is the root of P1-2 and P1-3 and interacts with P1-1 — replace it in one focused change (single-AST id derivation + non-ambiguous scanner + a real alignment test), then re-review the whole reader in one pass. Splitting the unreachable Copy Markdown / New Agent chat work into its own PR would also shrink this one meaningfully.

**Conflicting signal on this same SHA.** There is an automated `APPROVED` review on `6d31667a` stating no P0/P1 blockers with data flow verified across all 23 files. The measurements in P1-1 and the unchanged code in P1-2 do not support that conclusion; please don't merge on that signal alone.

**Coverage / what I did not verify.** I could not execute the test suite or a build in this environment (no `node_modules`), so every claim above comes from reading the diff plus standalone measurement of the two extracted algorithms — the timings in P1-1 and P2-4 are real, run against the exact expressions from the source. Not verified: (1) that the `th`/`td` `align` allowlist entries are live, hence the question in P2-6; (2) actual rendered output of the Setext/blockquote fixtures through the installed react-markdown 8 + remark-gfm 3 — the drift mechanism is derived from CommonMark and from the positional coupling in the code, not observed in a browser; (3) whether `SummaryDetailPage.participantReports.test.tsx:98`, which mocks `CitationText`, still asserts what it intends now that the page renders `SummaryMarkdownReader` in between (the mock still resolves to the same module, so it should hold, but the reader itself is now unmocked in that suite); (4) visual verification of the card seams in P2-5 and of the reader at the `1180px` / `719px` breakpoints; (5) runtime behaviour of the `initLocateCitationDisplayIndex` key change at `EndpointCommon.tsx:166`, which now forces a chat-page remount when navigating to the same message from a different citation — intended, but it will discard chat state such as an unsent draft.

PR #1041, opened 2026-07-23
ci(e2e): add visual baseline workflows

JianwenXu:
## Summary
- Add two independent Playwright visual baseline cases for loop workspace states.
- Add manual GitHub Actions workflows for initial visual baseline bootstrap and per-PR baseline updates.
- Include `@visual` in PR e2e and nightly e2e coverage while keeping visual-only failures non-blocking in the PR gate.

## Test Plan
- `pnpm -C apps/web build:e2e`
- `pnpm -C apps/web exec playwright test --config=e2e-kit/playwright.ci.config.ts --grep "@visual" --list` → 2 tests
- `PW_PREVIEW_PORT=5178 pnpm -C apps/web exec playwright test --config=e2e-kit/playwright.ci.config.ts --grep "@p0"` → 29 passed
- `pnpm -C apps/web exec playwright test --config=e2e-kit/playwright.ci.config.ts --grep "@p0|@visual" --list` → 31 tests
- `pnpm -C . exec prettier --check ...`
- workflow YAML parsed with PyYAML
- `git diff --check`


lml2468 (dismissed):
## Review: #1041 — ci(e2e): 增加视觉基线 workflows

**意图**:加 2 个 loop workspace 视觉基线 case;加手动 GitHub Actions workflow(初始 baseline bootstrap + per-PR baseline update);把 `@visual` 纳入 PR/nightly e2e 覆盖,同时 PR 门里 visual-only 失败保持非阻断。

锚定 head `5947172c` / merge-base `a8bad922`(behind 0,`mergeable=MERGEABLE`,0 冲突)。7 文件 +387/−40。

---

### Gate 1 — 规格符合度 ✅

- 2 个 `@visual` case(C1v/C3v,`toHaveScreenshot`,独立 `test()` 块、tag 不含 `@p0`)+ bootstrap/update 两 workflow + e2e/nightly 覆盖扩展 + 策略文档。与描述一致。✅

### Gate 2 — 代码质量 ✅(重点:CI workflow 安全 —— 尤其两个 `contents: write` 的 commit-flow)

**Workflow 安全(逐项,应用 CI 清单):**
- **actionlint 实跑 4 个 workflow 全 exit 0** —— 无 shellcheck/SC 阻断(#1013/#1029 类)。✅
- **无 `pull_request_target`**:两个 baseline workflow 均 **`workflow_dispatch` only**(维护者手动,fork PR 不可达);e2e/nightly 沿用 `pull_request`/`schedule`。✅
- **写权限最小化 + fork 防护**:bootstrap `contents:write`+`pull-requests:write`、update `contents:write`+`pull-requests:read`(commit baseline 所需);update 的 push 前有 **same-repo 硬门**(`pr.head.repo.full_name !== owner/repo` → `core.setFailed`,拒 fork)→ 只推组织自有分支。✅
- **input 处理安全**:`pr_number` 经 `env: PR_NUMBER` → `Number(process.env.PR_NUMBER)`(JS,无 shell);`branch_name` 经 `env: BASELINE_BRANCH` → `"$BASELINE_BRANCH"`(引号安全)。无不可信 `${{ github.event.* }}` 注入 `run:`。✅

**e2e 门仍 fail-closed(visual 扩展未削弱阻断)✅**:
- `--grep "@p0|@visual"`;`VISUAL_FAIL = @visual && !@p0`(**`@p0+@visual` 交叉→business/block**);business/proxy fail 仍 `exit 1`;count-guard 由 `EXPECTED_P0=29` 扩为 `EXPECTED_TOTAL=EXPECTED_P0+EXPECTED_VISUAL(2)`,匹配新增 2 case。visual-only fail 才非阻断。✅

**构建**:`pnpm --filter @octo/web build` ✅ 3.58s。

---

### 🟡 非阻断
1. **`head_ref` 内联 `${{ }}` 进 `git push` run**(update.yml:157 `git push origin "HEAD:${{ steps.pr.outputs.head_ref }}"`)—— 最佳实践应经 `env:` + `"$VAR"`。**风险已被 same-repo 门(:40 拒 fork)+ git refname 约束(禁空格/`~^:?*`\`/控制符)双重压低**,加之 workflow_dispatch-only,实际可注入面极小;仍建议改为 env-passing 收口。
2. **`paths:`+required 死锁前瞻**(承 #1013)—— e2e.yml 有 `paths:` 过滤;main 当前未设分支保护,不触发;设 required 前需处理。
3. baseline 视觉断言依赖首次 bootstrap 产出的 PNG,首跑需人工确认基线正确(视觉 PR 固有)。

actionlint 全绿、无 pull_request_target、写权限最小化 + fork 硬门、input 经 env 无注入、e2e 门 visual 扩展仍 fail-closed(business/proxy 阻断)、count-guard 同步,构建通过。**可合并**(建议把 `head_ref` push 改为 env-passing)。

Jerry-Xin (dismissed):
Well-structured visual-baseline CI addition; security posture is strong (no `pull_request_target`, no untrusted interpolation into shell, least-privilege `permissions`, fork write-back blocked). One blocking issue: the baseline-update commit uses `[skip ci]`, which can leave the updated PR permanently unmergeable.

## 🔴 Blocking

- 🔴 **`[skip ci]` on the baseline-update commit can permanently block merge.** `.github/workflows/e2e-baseline-update.yml:154` commits with `chore(e2e): update visual baselines [skip ci]` and pushes to the PR head branch (`:155`). GitHub suppresses `pull_request` workflows for `[skip ci]` commits, so the new PR head will never re-run the required `e2e / e2e-p0` check — it stays pending and the PR cannot merge until a new non-`[skip ci]` commit is pushed. Recommend dropping `[skip ci]` so the updated snapshots are validated by the gate, or explicitly re-triggering the required check after the push.

## 💬 Non-blocking

- 🟡 **`@visual` failure classification is broader than screenshot mismatches.** `.github/workflows/e2e.yml:151` treats any failure of an independent `@visual` test (navigation, fixture, locator, runtime — not just `toHaveScreenshot` diffs) as a non-blocking "visual fail" and allows the PR. A real functional break inside a `@visual`-only case could be silently allowed. Consider isolating screenshot-assertion failures specifically, or running visual tests in a separate non-blocking job. (`@p0`-crossed cases are correctly routed to the blocking branch, so this is limited to visual-only cases.)
- 🟡 **Action refs use mutable tags, inconsistent with repo hardening convention.** The new/changed workflows pin `actions/checkout@v4`, `actions/setup-node@v4`, `actions/github-script@v7`, `actions/upload-artifact@v4`, `pnpm/action-setup@v4` to floating major tags, whereas existing repo workflows (e.g. `.github/workflows/ci.yml`) pin to full commit SHAs with a `# v4` comment. All refs here are trusted first-party actions so risk is low, but pinning to SHA would keep the supply-chain posture consistent across the repo.

## ✅ Highlights

- Bootstrap (`contents: write` + `pull-requests: write`) and update (`contents: write` + `pull-requests: read`) permissions are separated and scoped; both are `workflow_dispatch`-only (maintainer-triggered), not fork-reachable.
- Fork PRs are explicitly rejected from baseline write-back via the `pr.head.repo.full_name` same-repo guard in `.github/workflows/e2e-baseline-update.yml`.
- `concurrency.group` is the only place `${{ github.ref }}` appears — no untrusted context is interpolated into any `run:` shell block; `BASELINE_BRANCH`/`PR_NUMBER` are passed via `env:` and referenced quoted.
- Strong fail-closed gate: `tests=0`, `tests < EXPECTED_TOTAL(31)`, `skipped>0`, and Vite proxy-error all block, preventing a vacuous/empty-baseline pass; the 2 new `@visual` tests match `EXPECTED_VISUAL=2`.
- Snapshot generation is always followed by an independent verification run, and the new visual cases assert stable page state before capturing screenshots.

lml2468 (approved):
## Re-review: #1041 @ `b29e47f7` — ci(e2e) 视觉基线 workflows

复核 Jerry-Xin 的 🔴(`[skip ci]` 卡死 required 门禁)+ 我上轮的 🟡(`head_ref` 内联 `${{ }}`)是否修复。

锚定 head `b29e47f7` / merge-base `a8bad922`(behind 0,`mergeable=MERGEABLE`)。真实增量(vs 我批准的 `5947172c`):仅 `e2e-baseline-update.yml`(+6/−2)+ 1 行文档。

**结论:两条都修好了,可合并 ✅**

### 🔴 `[skip ci]` 卡死 required 门禁(Jerry-Xin)→ 已修 ✅
- commit message 从 `"chore(e2e): update visual baselines [skip ci]"` → **`"chore(e2e): update visual baselines"`**(去掉 `[skip ci]`)。
- baseline 推回 PR head 分支后,`pull_request` workflow 不再被抑制 → required `e2e / e2e-p0` 在新 head 正常重跑 → **不再卡死合并**。✅

### 🟡 `head_ref` 内联 `${{ }}` 注入面(我上轮)→ 已修 ✅
- `git push` 前新增 `env: HEAD_REF: ${{ steps.pr.outputs.head_ref }}`,run 里改用 **`git push origin "HEAD:$HEAD_REF"`**(shell 变量,非表达式内插)—— 注入面收口(即便同 repo 门 + refname 约束已压低风险,现在是干净的 env-passing)。✅

### 承前无回归 ✅(仅这一处改动,其余未动)
- actionlint 实跑 4 个 workflow **全 exit 0**;
- 两 baseline workflow 仍 `workflow_dispatch`-only、权限最小化(update `contents:write`+`pull-requests:read`)、update 有 same-repo fork 硬门;
- e2e 门仍 fail-closed(`@p0+@visual` 交叉→block、business/proxy→exit 1、`EXPECTED_TOTAL=31` count-guard)。

**构建**:纯 CI/workflow 改动,无产品代码;actionlint + YAML 已验。

---

### 🟡 非阻断(承前,未变)
- `paths:`+required 前瞻死锁(main 未设保护不触发);
- 一方 action 用浮动 tag(`@v4`/`@v7`)vs 仓库 SHA-pin 惯例(风险低,建议对齐);
- `@visual`-only 失败分类较宽(纯视觉用例的功能性中断可能非阻断放行;`@p0` 交叉已正确阻断)—— Jerry-Xin/我均提,可后续细化。

Jerry-Xin 的 `[skip ci]` 卡死 🔴 与我的 `head_ref` 注入 🟡 均已干净修复,actionlint 全绿、安全面(注入/权限/fork 门/fail-closed)无回归。**可合并 —— APPROVE。**

yujiawei (approved):
# Code Review — PR #1041 (octo-web)

Reviewed at head `b29e47f7` (merge-base `a8bad922`, `MERGEABLE`, 7 files +389/−40). This adds two independent Playwright `@visual` baseline cases, two `workflow_dispatch` baseline workflows (bootstrap + per-PR update), extends PR/nightly e2e coverage to include `@visual`, and updates the strategy doc.

## 1. Spec compliance

**Spec: ✅**

- Missing: none. Two `@visual` cases (`@C1v`, `@C3v`, `toHaveScreenshot`, independent `test()` blocks not tagged `@p0`), bootstrap + update workflows, e2e/nightly coverage extension, and doc update are all present as described.
- Over-build: none. No unrelated functional changes.
- Deviation: none in the code. The count guards are consistent with reality — `@p0`=29, `@visual`=2, union=31; `EXPECTED_P0=29` / `EXPECTED_VISUAL=2` / `EXPECTED_TOTAL=31` (`.github/workflows/e2e.yml:128-130`) match `grep @p0`/`@visual` in `apps/web/e2e-kit/tests/`.

## 2. Code quality

**Quality: Approved**

Security posture of the two `contents: write` workflows is solid:
- Both baseline workflows are `workflow_dispatch`-only (maintainer-triggered, not fork-reachable); no `pull_request_target`.
- `e2e-baseline-update.yml:40-42` hard-rejects fork PRs (`pr.head.repo.full_name !== owner/repo` → `core.setFailed`) before any push, so write-back only targets org-owned branches.
- Untrusted inputs are passed via `env:` and referenced quoted (`PR_NUMBER` → `Number(process.env.PR_NUMBER)`; `BASELINE_BRANCH`/`HEAD_REF` → `"$VAR"`), including the previously-inline `git push` ref now routed through `env: HEAD_REF` (`e2e-baseline-update.yml:140-141,159`). No untrusted `${{ github.event.* }}` reaches a `run:` shell.
- The PR gate stays fail-closed: `tests=0`, `tests < EXPECTED_TOTAL`, `skipped>0`, proxy-error, and any business fail all `exit 1`; `@p0`-crossed `@visual` failures route to the blocking branch (`e2e.yml:98,145-156`). Only visual-only failures are allowed through.
- The update workflow gates baseline writes behind a real `@p0` pass (`tests>=EXPECTED_P0`, `skipped=0`, `errors=0`, no proxy error) and re-verifies snapshots in-job after `--update-snapshots`, so committed baselines are validated before push.

### Findings (non-blocking)

- **P2 — Removing `[skip ci]` does not actually re-trigger the PR gate; the commit message and doc step 6 overstate the effect.** `e2e-baseline-update.yml:159` pushes to the PR head using the checkout's default `GITHUB_TOKEN` (no `token:` override, `persist-credentials` left at its default). GitHub does not create new workflow runs for pushes made with `GITHUB_TOKEN` (recursion guard) — so the `pull_request` (synchronize) `e2e` gate will **not** re-run on the updated head, despite `[skip ci]` being dropped. The follow-up commit's stated goal ("rerun CI after baseline updates") and `docs/e2e-baseline-strategy.md:43` ("让普通 pull_request gate 在新 head 上重新验证 baseline diff") are therefore not achieved by this change. This is not blocking because (a) `main` currently has no branch protection / required-check enforcement, so no merge is actually stuck, and (b) the update workflow already re-verifies the snapshots in-job (`e2e-baseline-update.yml:125-137`) before committing, so the baselines are validated regardless. It becomes a real problem only if `e2e` is later made a required check — then a baseline-update push would leave the PR waiting on a check that never re-runs. Recommend either pushing with a PAT/App token (so `pull_request` re-runs), or explicitly re-dispatching the gate after push, and correcting the commit/doc wording to match the actual mechanism. The same limitation applies to the bootstrap workflow opening its PR via `github-script` (`e2e-baseline-bootstrap.yml`): a PR opened with `GITHUB_TOKEN` won't trigger the `pull_request` gate either.
- **P2 — `@visual`-only failure classification is broader than screenshot diffs.** `e2e.yml:98,151-153` treats *any* failure of a `@visual`-and-not-`@p0` test as a non-blocking "visual fail" and allows the PR. The two new cases contain functional assertions before the screenshot (`authedPage.goto`, `toHaveText("Workspace A")`, `getByText("还没有任务")` in `C3-loop-workspace-default-view.spec.ts:33-40`), so a genuine app regression that breaks navigation/state — not just a pixel diff — inside a visual-only case would be silently allowed. The blast radius is limited (the same flows are covered by the blocking `@p0` `C1`/`C3` cases, and `@p0`-crossed visual cases correctly block), so this is an accepted tradeoff rather than a defect; consider isolating `toHaveScreenshot` mismatches specifically or running visual tests in a separate non-blocking job if visual-only coverage grows.
- **Nit — mutable action tags vs. repo SHA-pinning convention.** The new/changed workflows pin `actions/checkout@v4`, `actions/setup-node@v4`, `actions/github-script@v7`, `actions/upload-artifact@v4`, `pnpm/action-setup@v4` to floating major tags, whereas some existing repo workflows pin to full commit SHAs. All refs here are trusted first-party actions so risk is low; pinning to SHA would keep supply-chain posture consistent.

## 3. Overall verdict

**APPROVED**

No P0/P1 issues. Spec is fully met and the workflow security model is sound. The two P2 findings (GITHUB_TOKEN not re-triggering the PR gate — with the commit/doc wording overstating it; and the broad visual-only failure classification) and the SHA-pinning nit are worth addressing but do not block merge, especially given `main` is currently unprotected and baselines are validated in-job.

## 4. Suggestions

- Push baseline updates with a PAT/GitHub App token (or explicitly re-dispatch the gate) if the intent is genuinely to re-validate on the new head; then align the commit message and `docs/e2e-baseline-strategy.md:43` with the real mechanism.
- If visual-only coverage grows, split screenshot-mismatch failures from other failure types so functional breaks in `@visual`-only cases aren't auto-allowed.

## 5. Additional observations

- `apps/web/e2e-kit/.gitignore` keeps `screenshots/` ignored while both workflows use `git add -f apps/web/e2e-kit/screenshots/` — consistent with the doc's stated approach; developers' local `--update-snapshots` output stays uncommitted while CI can still force-add baselines.
- Prior review rounds flagged `[skip ci]` (now removed) and inline `${{ }}` in the push ref (now via `env:`); both are resolved in this head. The residual `[skip ci]` concern is superseded by the GITHUB_TOKEN P2 above.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1041 (octo-web)

## Summary

This PR introduces visual baseline infrastructure for the E2E test suite: two new `workflow_dispatch` workflows (`e2e-baseline-bootstrap.yml` for one-shot initialization and `e2e-baseline-update.yml` for per-PR refresh), plus modifications to the MR gate (`e2e.yml`) and nightly (`e2e-nightly.yml`) to run `@visual` cases alongside `@p0`. Two new `@visual` test cases (@C1v, @C3v) are added, MSW mock coverage is expanded to plug proxy leaks, and the strategy doc is updated.

The PR also bundles production improvements: a WKSDK-to-runtime-seam refactoring that removes direct `WKSDK.shared()` calls from ForwardModal, Contacts, ConversationSelect, GlobalSearch, and AppBot modules; `selectedChannels` support in AgentChatPanel for agent-mode chat context; Onboarding focus-trap accessibility fixes; and HtmlDocView presence-bar + access-request badge enhancements.

No P0/P1 issues found. The CI workflow design is thoughtful, well-gated, and security-conscious.

## Verification

- ✅ **Workflow security** — Both new workflows use `workflow_dispatch` only (not `pull_request_target`); permissions correctly scoped (`contents: write` justified by branch push, `pull-requests: write` in bootstrap justified by PR creation). Fork PR rejection guard at `e2e-baseline-update.yml:48-51`.
- ✅ **Gate logic** — `e2e.yml` junit post-processing correctly categorizes business vs visual failures under the new `@p0|@visual` grep. Fail-closed ordering (tests=0 → tests<expected → skipped>0 → proxy errors → business fail → playwright crash → visual-only allow → success) is sound.
- ✅ **MSW coverage** — New handlers in `chat-baseline.ts` (health, avatar_palette, channel/sync, obo/grants) close real proxy leak vectors. Wildcard URL patterns (`*/api/v1/...`) correctly catch both prefixed and unprefixed routes.
- ✅ **SDK seam refactoring** — All `WKSDK.shared()` direct calls in changed files replaced with `getCurrentIm*`/`addCurrentIm*`/`fetchCurrentIm*` runtime abstraction functions. Tests updated to match new signatures. `createEmptyConversation` is properly optional-chained.
- ✅ **Concurrency isolation** — Bootstrap uses a fixed concurrency group; update uses per-PR-number groups; `PW_PREVIEW_PORT` differs per workflow (5175/5176) to prevent port collision on shared runners.
- ✅ **Onboarding a11y** — Focus trap now handles `activeElement === dialog` edge case at `index.tsx:363`; intro screen focuses the dialog itself at `:408` rather than a child element; skip transition uses CSS visibility instead of a ghost DOM node.
- ✅ **HtmlPresenceBar identity** — `viewerName` at `HtmlDocView.tsx:361` resolves from `names.get(getCurrentUid())` — the viewer's own name from the space member map. The old "never show another user's identity" invariant is preserved.
- ✅ **useAccessRequests dedup** — `HtmlMemberPanel.tsx:93` disables the local hook when `sharedAccessRequests` is provided, preventing double-fetch.

Static analysis only at head `b29e47f7e25f2a0f121e9556abb2729e2d176ff5`; build and tests not executed in this environment.

## Findings

No P0/P1 issues; one P2 and one Nit below.

### P2 — Gate constants duplicated across three workflows (`.github/workflows/e2e.yml:131`)

`EXPECTED_P0=29` and `EXPECTED_VISUAL=2` appear as bare constants in three separate workflow files (`e2e.yml:131-132`, `e2e-nightly.yml:92-93`, `e2e-baseline-update.yml:100`). When a @p0 or @visual test is added or removed, all three must be updated in lockstep — there is no shared source of truth and no CI check that they agree with reality. A stale value in one workflow either falsely blocks a valid PR or silently passes a degraded suite. Consider extracting to a shared file read by all three workflows, or at minimum adding a cross-reference comment at each site pointing to the others.

### Nit — Undocumented sentinel URL in CI config (`apps/web/e2e-kit/playwright.ci.config.ts:66`)

`VITE_API_URL: "http://127.0.0.1:9"` uses a deliberately unreachable address as a fail-closed trap for MSW-missed requests. Sound E2E strategy, but the magic port 9 has no inline comment explaining the intent. A one-line comment (e.g. `// intentionally unreachable; MSW must intercept all API calls before reaching this`) would prevent future confusion.

- Bootstrap workflow idempotency: `git show-ref --verify` at `e2e-baseline-bootstrap.yml:42` correctly handles first-run (branch doesn't exist) and subsequent runs.
- PR-create dedup: `actions/github-script` at `e2e-baseline-bootstrap.yml:108` checks for existing open PRs before creating a new one.
- Baseline update pushes to PR head via `git push origin "HEAD:$HEAD_REF"` after checking out the PR branch — correct push semantics without `[skip ci]`, so normal push-triggered CI will re-validate.
- Both workflows verify generated baselines by re-running `@visual` tests after `--update-snapshots` before committing — catches flaky baselines.
- Proxy-error grep (`grep -R "http proxy error:"`) covers all log files in the report directory — no blind spots.
- `channelConvert.ts` archived-thread detection at `:35` reads `orgData.thread.status === 2` with a clear comment mapping to `ThreadStatus.Archived`. Test at `channelConvert.test.ts:81` covers it.
- `selectedChannels` data flow: `SummaryCreatePage` and `ChatSummaryNewModal` pass `selectedChats` to `AgentChatPanel`; `requestSelectedChannels()` at `:255` strips to the API shape `Pick<ChatCandidate, 'chat_id' | 'chat_type' | 'name' | 'is_archived'>`.
- `getCurrentImConversationsDirectly` at `currentConversationRuntime.ts:73` documents its intentional seam bypass and falls back to `[]` on null/undefined.
- `NewFriendVM` at `NewFriend/vm.tsx:19` properly pairs `addCurrentImCommandListener`/`removeCurrentImCommandListener` in `didMount`/`didUnMount` — no listener leak.
- `EmojiService.load()` early-returns when `VITE_E2E_MOCK=1` at `EmojiService.ts:284`, preventing real network calls in e2e. Aligns with MSW handlers.

## Verdict: APPROVED

Well-structured CI infrastructure addition with solid gate logic, proper security scoping, and comprehensive mock coverage. The production code changes (SDK seam refactoring, selectedChannels wiring, Onboarding a11y, HtmlDoc enhancements) are clean and well-tested. The two non-blocking findings are a maintainability concern about duplicated gate constants and a documentation nit on the sentinel URL.




[Octo-Q] verdict: APPROVE — No P0/P1 issues. The CI workflow infrastructure is well-designed with proper security guards, fail-closed gate logic, and comprehensive MSW mock coverage. The production code changes are clean SDK seam refactoring with good test coverage. Only non-blocking items are duplicated gate constants (P2 maintainability) and an undocumented sentinel URL (Nit).

Jerry-Xin (approved):
🔄 Re-review at head b29e47f7 — my prior sole blocker is fixed; approving.

## ✅ Prior blocker resolved (byte-verified)

- **`[skip ci]` removed from the baseline-snapshot commit.** `.github/workflows/e2e-baseline-update.yml` now commits with `chore(e2e): update visual baselines` (no skip token anywhere). The push to the PR head branch will now re-fire the `pull_request` gate (`.github/workflows/e2e.yml` triggers on `pull_request` → `main` with `paths` covering `apps/web/**`, `packages/**`, and the workflow files), so the required `e2e / e2e-p0` check re-validates on the updated snapshots and the PR can merge. Delta is a clean fast-forward (ahead 1, behind 0, merge-base = prior head 5947172c); one commit `fix(e2e): rerun CI after baseline updates` touching only `e2e-baseline-update.yml` + `docs/e2e-baseline-strategy.md`.
- **Bonus:** `head_ref` is now passed via `env: HEAD_REF` instead of inline `${{ }}` in `git push` — resolves my prior non-blocking note on inline interpolation.

## 💬 Non-blocking (hardening, not required)

- **`e2e-baseline-update.yml` — defense-in-depth on the write-credentialed job.** The job holds `contents: write` and runs PR-branch code (install/build/playwright) before the commit/push step. This is acceptable here: the workflow is `workflow_dispatch`-only (maintainer-triggered), there is no `pull_request`/`pull_request_target` auto-trigger, and the same-repo guard rejects fork PRs — so the checked-out code always belongs to a collaborator who already has write access (no privilege boundary is crossed). As optional hardening you could generate snapshots in a job with `persist-credentials: false`, then do the narrow commit/push in a separate step/job. Not blocking.
- **`e2e.yml` — visual-only failure classification is title-based.** A failure is treated as non-blocking whenever the test title carries `@visual` and not `@p0`, so nav/fixture/assertion/timeout failures inside a standalone `@visual` test are also allowed through the PR gate (nightly catches them). Consider narrowing to the screenshot-comparison assertion. Unchanged from prior review; not blocking.
- **`e2e-baseline-bootstrap.yml` — bootstrap branch reuse.** If the bootstrap branch already exists it is reused rather than rebuilt from current `main`, so a rerun could miss newly added visual cases. Consider reset/rebase onto the intended source before snapshotting. Not blocking.
- **Action pins.** Workflows use floating major tags (`@v4`/`@v7`) vs. the repo convention of full-SHA pins. Not blocking.

## ✅ Highlights

- Fork PRs explicitly rejected by the update workflow (same-repo guard intact).
- Baselines verified by a second Playwright run before commit.
- Empty-gate fail-closed: test-count / skipped / proxy-error / missing-JUnit guards prevent false-green.
- `@visual` cases are independent from the blocking `@p0` cases.
- actionlint passes on all four workflows (`e2e.yml`, `e2e-nightly.yml`, `e2e-baseline-bootstrap.yml`, `e2e-baseline-update.yml`).

cnwenf (commented):
# Code Review — PR #1041 (Mininglamp-OSS/octo-web)

## Summary

This PR stands up the visual-baseline CI machinery for the web E2E suite: two new standalone `@visual` Playwright cases (`@C1v`/`@C3v`), two manual `workflow_dispatch` workflows (baseline bootstrap + per-PR baseline update), and it widens the MR gate (`e2e.yml`) and nightly (`e2e-nightly.yml`) from `@p0` to `@p0|@visual`, keeping visual-only failures non-blocking in the MR gate. All three reviewers found the core design well-executed and security-sound — correct business/visual fail split, fork-PR rejection before any write, least-privilege permissions, count guards matching live counts, and a green `@p0` suite (29/29, 0 proxy errors). No blockers; consolidated findings are 5 P2 should-fix + 9 P3 hardening + 1 nit. The top risk is operational: nightly runs unconditionally red until the bootstrap baseline is committed (confirmed by two reviewers with runtime evidence). Reviews were conducted at head `5947172c`; follow-up commit `b29e47f7` ("fix(e2e): rerun CI after baseline updates") landed afterwards and already implements the recommended `env: HEAD_REF` fix for the P2 injection finding below — a re-review of the new head is recommended.

## Findings

### P2 — Nightly unconditionally RED during the missing-baseline cold-start window (`.github/workflows/e2e-nightly.yml:74`) [source: octo-review-glm5.2 + octo-review-qwen3.8]

The diff changes the nightly grep from `--grep "@p[0-9]" --grep-invert "@visual"` to `--grep "@p[0-9]|@visual"` (`:74`), routing the two new `@visual` cases into a pipeline that has NO visual-allow logic. Unlike `e2e.yml` — which splits `BUSINESS_FAIL`/`VISUAL_FAIL` and allows visual-only fails with a clean `PW_EXIT` — the nightly ends with `exit "$PW_EXIT"` (`:105`), propagating any playwright non-zero exit directly. Since no baseline PNGs are committed yet, the 2 `@visual` cases fail every nightly run with `A snapshot doesn't exist` (missing baseline, not pixel drift) → `PW_EXIT=1`. Runtime-confirmed: from this PR's merge until the manual bootstrap baseline PR merges, nightly is red on every run, masking real visual drift with missing-baseline noise. (Playwright 1.61.1 does write-then-fail under CI, but the written PNG is ephemeral — fresh checkout, never committed by nightly — so the red window is unchanged.) Fix direction: skip/non-block `@visual` in nightly until a committed baseline exists, run bootstrap before merging this PR, or at least document the expected red window.

Also independently confirmed by octo-review-glm5.2, who reproduced locally: `@visual` exits 1 (missing baseline PNG) while `@p0` exits 0 green at head `5947172c9`; and flagged the operational risk that a persistently red nightly trains the team to ignore it. Recommends running bootstrap before/at merge, giving nightly the same visual-only allow the MR gate has, or keeping `--grep-invert "@visual"` on nightly until baselines exist.

### P2 — Hardcoded test count constants require atomic updates across three workflow files (`.github/workflows/e2e.yml:130`) [source: octo-review-qwen3.7 + octo-review-glm5.2]

`EXPECTED_P0=29`, `EXPECTED_VISUAL=2`, and `EXPECTED_TOTAL=31` appear in `e2e.yml`, `e2e-nightly.yml`, and `e2e-baseline-update.yml`. This PR correctly adds the 2 @visual cases and updates all three files, but the coupling is entirely manual — the inline comment "若确认减 case, 请同步下调" is the only guard. A future PR that adds a `@p0` or `@visual` test but forgets to bump the constant will cause the CI gate to block all PRs. Consider extracting these into a shared variable (e.g., a `vars` block or a `grep -c` pre-check that compares the constant against the actual tagged test count) to make drift self-detecting.

Also independently confirmed by octo-review-glm5.2 (as P3): the `EXPECTED_TOTAL` floor (`TOTAL_TESTS < EXPECTED_TOTAL` → block) doesn't protect future visual cases — adding a 3rd `@visual` case raises actual to 32; if it later silently drops out (grep drift / tag removed), actual falls back to 31 == floor and the guard does not fire.

(severity split: qwen3.7=P2, glm5.2=P3)

### P2 — `head_ref` expression injection into shell `run:` block (`.github/workflows/e2e-baseline-update.yml:157`) [source: octo-review-qwen3.8]

Unsanitized `head_ref` (`pr.head.ref` from the GitHub API) flows into a shell `run:` block via `${{ steps.pr.outputs.head_ref }}` expression interpolation: `git push origin "HEAD:${{ steps.pr.outputs.head_ref }}"`. GitHub Actions expands `${{ }}` textually before the shell executes, and git branch names may carry shell metacharacters (`;`, `|`, `$()`, backticks) that survive `git-check-ref-format` — so a crafted branch name injects arbitrary commands in a runner holding `contents: write` (CWE-78). Exploitability is low: it requires repo write access to create the branch and trigger `workflow_dispatch`, and write access already confers equivalent repo-write capability. The fix is trivial and already established in-file — the same workflow uses the safe `env:` pattern for `PR_NUMBER` (`:34`), which makes this inconsistency diff-causal:

```yaml
env:
  HEAD_REF: ${{ steps.pr.outputs.head_ref }}
run: |
  git push origin "HEAD:$HEAD_REF"
```

> Update: addressed in follow-up commit `b29e47f7` (pushed after the three reviews), which switches the commit/push step to exactly the `env: HEAD_REF` pattern recommended here.

### P2 — Update workflow proceeds with expensive steps after PR resolution failure (`.github/workflows/e2e-baseline-update.yml:33`) [source: octo-review-qwen3.7]

When the `Resolve PR head` step fails (invalid PR number, API error, or fork rejection via `core.setFailed()`), its outputs `head_ref` and `head_sha` are empty strings. GitHub Actions does not auto-cancel subsequent steps on `core.setFailed()` — it only marks the step as failed. The checkout step then receives `ref: ''` (empty), which falls back to the default branch. The workflow proceeds through install, build, and @p0 testing against `main`'s code before eventually failing at the push step (`git push origin HEAD:` with empty refspec). This wastes 5-10 minutes of runner time per invalid invocation. Add `if: steps.pr.conclusion == 'success'` (or `steps.pr.outputs.head_ref != ''`) to the checkout step to short-circuit early.

### P2 — Doc title contradicts its own body and the code (`docs/e2e-baseline-strategy.md:1`) [source: octo-review-qwen3.8]

The H1 still reads "e2e visual baseline 管理策略 (Phase 3 路线, 未实现)" ("Phase 3 roadmap, NOT implemented"), but this PR rewrites the entire body to describe the feature as already implemented ("MR 门禁 (e2e.yml) 跑 @p0|@visual", "已加少量独立 @visual case", "Nightly 已跑 @p[0-9]|@visual") and ADDS `e2e-baseline-bootstrap.yml` / `e2e-baseline-update.yml` in the same diff. Title, body, and code now directly contradict each other. Control on the base commit confirms the doc was internally consistent before this PR, so the contradiction is diff-introduced. Update the title (drop "(Phase 3 路线, 未实现)") to match the implemented state.

### P3 — Bootstrap lacks the `@p0` business gate that the update workflow has (`.github/workflows/e2e-baseline-bootstrap.yml:55`) [source: octo-review-glm5.2]

The initial baseline (the foundation every future visual diff compares against) only runs `@visual --update-snapshots` + verify `@visual`; it has no `@p0` sanity gate, unlike `e2e-baseline-update.yml:61-112` which refuses to push if `@p0` fails. Mitigated: the `@visual` cases' own pre-screenshot assertions fail under `--update-snapshots` if the UI is broken (no baseline written for an unreached screenshot line), plus the verify step and human PR review. Worth adding a `@p0` gate to bootstrap for parity with the update path.

### P3 — Proxy-error diagnostic unreachable under `set -e` (`.github/workflows/e2e-baseline-bootstrap.yml:71`) [source: octo-review-qwen3.8]

In both the bootstrap (`:71-76`) and update (`e2e-baseline-update.yml:130-137`) verify steps, the `grep -R "http proxy error:" e2e-kit/playwright-report/*.log` check runs AFTER the playwright verify command under `set -euo pipefail`. If the verify run exits non-zero — exactly the case where proxy-error diagnosis matters most — `set -e` aborts the step before the grep executes, so the operator sees a bare playwright assertion failure instead of the mock-coverage root-cause message. The workflow still fails closed (no commit occurs); only the diagnostic is lost. Reorder the grep ahead of the failing command or capture the exit code first.

### P3 — `git add -f` over-collects the screenshots directory (`.github/workflows/e2e-baseline-bootstrap.yml:89`) [source: octo-review-qwen3.8]

`git add -f apps/web/e2e-kit/screenshots/` bypasses `.gitignore` for the whole directory tree, so any non-PNG file present at commit time (build intermediates, stray config) would be staged and pushed. Runtime-probed: `-f` stages everything while the non-`-f` path stages nothing. Mitigated by fixed scope, PR review before merge, and a documented design decision — but a tighter `git add -f .../*.png` (or a `find ... -not -name '*.png'` pre-check) achieves the same goal without sweeping in unexpected files.

### P3 — Bootstrap workflow `branch_name` input lacks prefix validation (`.github/workflows/e2e-baseline-bootstrap.yml:39`) [source: octo-review-qwen3.7]

A collaborator could set `branch_name` to `main` or another protected branch. Mitigated by: (1) `workflow_dispatch` is only reachable by repo collaborators; (2) GitHub's GITHUB_TOKEN cannot bypass branch protection rules; (3) push to a protected branch will fail if protection is configured. Adding a `^chore/` prefix check would harden this further.

### P3 — Dead business/visual split in the update `@p0` gate (`.github/workflows/e2e-baseline-update.yml:85`) [source: octo-review-glm5.2]

That step runs `--grep "@p0"` only (line 69), so no `@visual`-only test runs there; yet it computes `VISUAL_FAIL` (line 85) and `BUSINESS_FAIL = TOTAL_FAIL - VISUAL_FAIL` (line 86). `VISUAL_FAIL` is always 0 here, so the split is dead logic copied from `e2e.yml`. Not a bug (`BUSINESS_FAIL == TOTAL_FAIL` is still correct), but misleading — either run `@p0|@visual` or drop the split.

### P3 — Verify step log scan scope is broader than the step name suggests (`.github/workflows/e2e-baseline-update.yml:134`) [source: octo-review-qwen3.7]

`grep -R "http proxy error:" e2e-kit/playwright-report/*.log` scans all accumulated logs from the @p0 gate, @visual update, and verify steps — not just the verify log. This is arguably intentional as a comprehensive safety net, but makes failure diagnosis ambiguous. If a proxy error came from the @p0 step, the verify step's rejection message doesn't distinguish the source.

### P3 — PR gate can't distinguish missing-baseline from real pixel-diff fails (`.github/workflows/e2e.yml:151`) [source: octo-review-qwen3.8]

PR gate 的 VISUAL_FAIL 只按测试名含 @visual 计数（`:98`），把『缺 baseline 导致的 fail』和『真实像素回归导致的 fail』算成同一类并一律 allow（`:151-152`），日志统一打『需手动更新 baseline』。baseline 合入前的冷启动窗口里每个 PR 都会固定出现 2 个 @visual fail→allow，窗口内合入的真实视觉回归在 PR gate 上与缺-baseline 噪声不可区分、被静默放过。建议缺 baseline 时显式标注/跳过，或在日志里区分 missing-snapshot 与 pixel-diff。（cross-tag `@p0+@visual` fail 仍正确 BLOCK——已验证。）

### P3 — Nightly comment still names `EXPECTED_P0` as threshold variable (`.github/workflows/e2e-nightly.yml:85`) [source: octo-review-qwen3.7]

The error message correctly uses `EXPECTED_TOTAL`, but the inline comment still references `EXPECTED_P0` as if it were the sole threshold. Minor doc drift from the partial comment update — the `e2e.yml` counterpart was fully updated to say "有 case 掉了 @p0/@visual tag 或 grep 漂了".

### P3 — `@C1v`/`@C3v` are weakened duplicates; `--update-snapshots` auto-adopts pixels (`apps/web/e2e-kit/tests/C3-loop-workspace-default-view.spec.ts:29`) [source: octo-review-qwen3.8]

新增 @C3v/@C1v 视觉用例是其 @p0 兄弟用例的弱化复制：C3v 省略了 case-spec 与 C3 @p0 都断言的 issueDesc『创建第一个任务，开始跟踪你的工作。』，C1v 省略了 C1 @p0 的『创建一个工作区开始使用回路。』。bootstrap/update 的 verify 只 `--grep "@visual"`，且 update 步骤 `--update-snapshots` 会无条件采纳当前渲染像素（runtime-confirmed：不做比较直接写 baseline 并报 pass），因此这些文本若回归，视觉用例不会失败、baseline 只会被重新生成去匹配错误画面；钉住该文本的只剩 @p0 用例（impact bounded，@p0 sweep 仍全绿）。建议在 @visual 用例补齐 case-spec 列出的文本断言。

### Nit — Unused step outputs (`.github/workflows/e2e-baseline-update.yml:45`) [source: octo-review-glm5.2]

`Resolve PR head` sets `head_sha` (line 45) and `base_ref` (line 46) that are never consumed downstream — only `head_ref` is used for checkout and push. Dead surface; drop or use.

- **Count guards match reality** — all three reviewers verified `@p0`=29, `@visual`=2, `@p0|@visual`=31 via `playwright --list`; the PR's test-plan counts reproduce exactly, and the under-count / skipped / proxy / `tests=0` guards BLOCK on synthetic drift fixtures.
- **Mock-mode pipeline green** — full `@p0` sweep 29/29 under MSW + fake-IM with 0 "http proxy error:"; the 2 `@visual` cases pass 2/2 once baselines exist (pre-bootstrap failure is missing-baseline, not drift).
- **MR-gate business/visual split is correct** — a fail carrying `@p0` (even with `@visual`) is counted as BUSINESS and BLOCKS; the `PW_EXIT` check precedes the visual-allow so a crash isn't reclassified as visual; cross-tag fails still block; standalone `@C1v`/`@C3v` carry `@visual` only and correctly route through the allow path.
- **Fork-PR write-back correctly blocked** — `e2e-baseline-update.yml` checks `pr.head.repo.full_name !== owner/repo` before any write; `e2e.yml`/`e2e-nightly.yml` keep `permissions: contents: read` so fork PRs never get a write token.
- **Permissions least-privilege** — gates `contents: read`; bootstrap `contents: write + pull-requests: write`; update `contents: write + pull-requests: read`; all explicitly declared, concurrency groups bound parallelism.
- **Credentials clean** — zero hardcoded secrets across the 7 changed files; only the default GITHUB_TOKEN; actions pinned to mainstream maintained versions.
- **Screenshot paths & gitignore** — `snapshotPathTemplate` resolves to `apps/web/e2e-kit/screenshots/{projectName}/...`, matching the `git add -f` paths; `.gitignore:6` ignores `screenshots/` so the `-f` is correctly required.
- **Screenshot stability** — both `@visual` tests assert key content via `toBeVisible()`/`toHaveText()` before `toHaveScreenshot`; combined with `reducedMotion: "reduce"`, `maxDiffPixelRatio: 0.01`, `threshold: 0.2`, and synchronous MSW handlers.
- **MSW scenario consistency** — C1v correctly omits `installMswScenario()` (default `"empty"`, matching C1); C3v uses `authedPage, "one-ws"`, matching C3.
- **Test isolation** — `fullyParallel: false, workers: 1`, fresh browser context per test, `sessionStorage` isolated across contexts.
- **Push safety** — update push is non-force (`git push origin HEAD:head_ref`); a concurrent developer push fails closed, no history rewrite.
- **Bootstrap idempotent** — reuses an already-open baseline PR and short-circuits on an empty diff; baselines land via PR review, never pushed straight to main.
- **`[skip ci]` semantics** — update workflow uses `[skip ci]` to prevent CI loops; bootstrap omits it so the bootstrap PR is CI-validated before merge (note: follow-up commit `b29e47f7` intentionally re-runs the gate on update pushes and updated the doc accordingly).

## Verification Results

| Code Review | APPROVED_WITH_RISK | 5 P2 + 9 P3 + 1 Nit (deduped across 3 reviewers), no blockers; top risks = nightly red until bootstrap baseline, and `head_ref` injection (already fixed in follow-up commit `b29e47f7`) |
| E2E Test | PASS | All three reviewers: `@p0` 29/29 green with 0 proxy errors; `@visual` 2/2 once baselines exist (pre-bootstrap fail = missing baseline, not drift); counts 2/29/31 reproduce exactly |
| Security | CLEARED_WITH_RISK | 1×P2 `head_ref` expression injection (exploitability limited by required write access; fixed in follow-up commit) + 1×P3 `git add -f` over-collection; no Critical/High; perms least-privilege; fork rejection robust; credentials clean |

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #122 (octo-docs-backend)
**Head SHA:** `c811e8c2128aa17f955cf2209ac29fa4e5f9662c`
**Scope:** 77 files, +12,011 / −208 — unify document import and export pipelines

## Summary

This PR introduces four new import formats (Markdown, Excalidraw, XLSX, and enhanced DOCX) and three new export formats (DOCX, Markdown, XLSX), along with a unified `exportFileHandler`, new serialization modules for DOCX (MathML→OMML, images, tables, styles, marks, nodes) and Markdown, and modifications to existing DOCX import, Typst rendering, board editing, and object storage. The scope is large but the architecture is coherent: each format gets a parser/serializer, wired into Express route handlers with body-size limits and input validation.

Four P1 blockers were identified: unbounded recursion enabling stack-overflow DoS in HTML table import, a formula-safety bypass enabling DDE attacks in XLSX export, silent corruption of relative URLs during Markdown HTML import, and missing admission control on new expensive endpoints that breaks the established load-shedding pattern.

## Verification

Static analysis only at head `c811e8c`; build and tests not executed in this environment.

- ✅ **Authentication** — All new import endpoints require `writer` role via `requireDocRole`; export endpoints require `reader`. Both human and bot route mounts covered in `src/api/app.ts:176-211`.
- ✅ **Body-size limits** — Each new handler enforces explicit byte caps at both the `express.raw` middleware level and re-checks inside the handler (defense in depth).
- ✅ **SVG sanitization** — SVG candidates from DOCX media and board exports go through `sanitizeSvg` validation before storage. Fallback chain terminates correctly.
- ✅ **Attachment copy ordering** — `copyStoredObject` now writes to object store before registering in DB, with compensating cleanup on failure. Improvement over prior order.
- ✅ **XML depth guard** — `maxNestedTags: 512` in DOCX XML parser provides defense-in-depth against hostile XML (real OOXML depth of 361 fits with ~40% headroom).
- ✅ **Excalidraw inline-data assertion** — Thorough validation: ID collision handling, staged file decode, cleanup on failure, MIME sniff cross-check.

## Findings

Four P1 blockers and six P2 items below.

### P1 — Unbounded recursion in `domCellToBlocks` — stack-overflow DoS (`src/import/markdown/html-inline.ts:712`)

`domCellToBlocks` accepts a `depth` parameter and increments it for blockquote recursion (line 712: `depth + 1`) and table recursion (line 689 → `domTableToNode`), but **never checks `depth` against any limit**. The only depth guard (`MAX_HTML_TABLE_DEPTH = 50`) lives in `domTableToNode` (line 618), which checks before calling `domCellToBlocks` — but `domCellToBlocks` recurses into itself for `<blockquote>` without ever hitting that guard.

Additionally, `domListToNode` at line 753 resets depth to `0` when calling `domCellToBlocks`, so nested `<ul><li><ul><li>…` inside a table cell is also unbounded.

Attack path: markdown import → `parseHtmlBlock` → `parseHtmlTable` → JSDOM → `domTableToNode` → `domCellToBlocks` → `domCellToBlocks` (blockquote recursion, unchecked). A crafted markdown file with deeply nested blockquotes inside an HTML table cell crashes the Node.js process (unrecoverable).

Diff-scope: **new** code introduced by this PR. Not pre-existing.

```ts
// Fix: add depth guard at the top of domCellToBlocks
function domCellToBlocks(cellEl, inlineMap, warnings, depth = 0, blockMap?) {
  if (depth > MAX_HTML_TABLE_DEPTH) return [{ type: 'paragraph' }]
  // ... existing code ...
}
```

Also propagate depth through `domListToNode` instead of hardcoding `0`.

### P1 — Quoted DDE syntax bypasses formula safety check (`src/import/xlsx/formulaSafety.ts:6`)

The DDE pipe-pattern regex `/(?:^|[=,+\-*/(])[^'"]+\|[^!]+!/` requires `[^'"]+` (one or more non-quote chars) before the DDE pipe `|`. The standard Excel quoted-DDE form bypasses this:

```
='cmd'|'/c calc'!A0
```

After normalization (whitespace stripped, uppercased): `'CMD'|'/CCALC'!A0`. At the `|` position, the preceding character is `'` (quote), so `[^'"]+` cannot match. This is a known DDE initiation vector that can launch arbitrary processes when the exported `.xlsx` is opened in Excel.

Diff-scope: **new** code introduced by this PR.

```ts
// Fix: add dedicated pattern for quoted DDE
if (/'[^']+'\|'[^']+'!/i.test(normalized)) return false
```

### P1 — Relative URLs silently corrupted by dummy-base resolution (`src/import/markdown/html-inline.ts:117`)

For relative URLs like `/page` or `./doc.pdf`, `new URL(raw, 'https://safe.local/')` resolves against the dummy base, producing `https://safe.local/page`. Since the resolved protocol is `https:` (safe), `u.href` is returned — but `https://safe.local/page` is a **non-functional URL**. The relative-URL fallback at line 128 (`if (!/^[a-z]…/i.test(raw.trim())) return raw`) is dead code: it is never reached because URL parsing already succeeded with a safe scheme.

Every relative link in imported markdown HTML is silently broken on import.

Diff-scope: **new** code introduced by this PR.

```ts
// Fix: check for relative URLs BEFORE the new URL() call
if (/^(javascript|data|vbscript):/i.test(cleaned)) return null
if (!/^[a-z][a-z0-9+.-]*:/i.test(raw.trim())) return raw  // relative — accept as-is
try {
  const u = new URL(raw, 'https://safe.local/')
  if (SAFE_SCHEMES.has(u.protocol)) return u.href
  return null
} catch { return null }
```

### P1 — Missing admission control on new expensive endpoints (`src/api/routes/import.ts:65`, `:594`, `:676`; `src/api/routes/export.ts:328`)

The existing DOCX import uses `acquireDocxImportSlot()` (line 477, bounded concurrency with 503 load shedding) and PDF export uses `acquireSlot()` (line 520, Typst compile pool). The four new expensive handlers have **no equivalent gate**:

- `importExcalidrawHandler` (line 65): walks entire scene, decodes/sanitizes embedded files, uploads to object store, writes to board.
- `importMarkdownHandler` (line 594): parses full markdown with HTML tables, resolves images, builds document.
- `importXlsxHandler` (line 676): parses workbook, decodes sheet snapshot, builds replacement maps.
- `exportFileHandler` (export.ts line 328): for DOCX, downloads all referenced images; for markdown, generates signed URLs; for XLSX, decodes full snapshot.

A burst of concurrent authenticated requests can saturate the event loop, exhaust object-store connections, and starve the existing admission-controlled endpoints. This makes a working service path unavailable.

Diff-scope: **new** endpoints introduced by this PR. The admission-control pattern is established pre-existing code.

Fix: reuse `acquireDocxImportSlot()`/`releaseDocxImportSlot()` (or create a shared import slot) for the new import handlers. Add a separate export slot for non-PDF paths in `exportFileHandler`.

### P2 — Global JSON body-parser bypass is method-agnostic (`src/api/app.ts:132`)

`if (req.path.endsWith('/import/excalidraw'))` skips JSON parsing for any HTTP method. Today benign (only POST matches), but fragile for future route additions. Tighten to `req.method === 'POST' && …`.

### P2 — Non-null assertions in export handler (`src/api/routes/export.ts:347`)

`req.uid!` / `req.params.docId!` / `req.spaceId!` suppress TypeScript's undefined checks. The import handlers in the same PR explicitly validate these fields. Add the same validation for consistency.

### P2 — No overall timeout on `resolveDocxImages` (`src/api/routes/export.ts:426`)

Each image download has a 10s timeout, but a document with `maxImageCount` images could hold the request open for `maxImageCount × 10` seconds. Combined with missing admission control, concurrent requests accumulate. Add an `AbortSignal.timeout()` for the overall phase.

### P2 — Inline `` silently dropped by `domInlineToNodes` (`src/import/markdown/html-inline.ts:58`)

No handler for `` in the inline DOM walker. `domImgToNode` exists but is only called from `domCellToBlocks` (table-cell context). Images inside inline HTML fragments are silently lost during import.

### P2 — XLSX hyperlinks skip `isSafeHref` defense-in-depth (`src/import/xlsx/toSheetBatch.ts:23`)

Uses a simpler regex (`/^(https?:|mailto:)/i`) instead of `isSafeHref`, missing control-character stripping and other sanitization.

### P2 — Font-family regex rejects valid non-CJK unicode (`src/import/markdown/html-inline.ts:248`)

Only ASCII + CJK Unified Ideographs allowed. Font names with Hangul, Hiragana/Katakana, Arabic, Cyrillic, or Devanagari are silently rejected.

## Data Flow Tracing

- **Markdown import → attachment migration:** `parseExportedAttachmentRef(src)` → `docMetaRepo.getByDocId` (verify source exists) → `meta.space_id === spaceId` (same-space check) → `resolveRole(uid, ref.docId)` (reader access) → `docAttachmentRepo.getById` (verify ownership) → `copyStoredObject` (store-to-store) → on edit failure, `cleanupCopiedAttachment` compensates. **Correct and fully authorized.**
- **Export signed URL flow:** `docAttachmentRepo.listByDoc(docId)` → `store.presignGet(key, ttl)` → embedded in output. URLs scoped to exported doc and time-limited. **Correct.**
- **SVG bytes path (DOCX import):** `resolveMedia(relId)` → `sniffImageMime` → `sanitizeSvg` → `ctx.upload`. Fallback: sanitizeSvg failure → raster fallback → degrade to fileAttachment. No unsanitized SVG reaches storage. **Correct.**
- **Board file delete path:** `validateOpsShape` → `validateBoardOps` → `applyBoardOpsToDoc` (deletes before upserts) → Y.Map mutation. **Correct.**

## Blind Spot Checklist

- **C1 (dual-path parity):** Import/export are asymmetric operations, not paired add/remove. N/A.
- **C2 (control-flow ordering):** `jsonBodyParser` bypass checked against all route registrations — only `importRouter.post('/:docId/import/excalidraw')` matches. No dual-mount issue. **Clear.**
- **C3 (authorization boundary):** All new endpoints behind `requireDocRole`. `migrateImportedAttachments` does independent auth (space + reader role) on source doc. **Clear.**
- **C4 (authorization lifecycle):** Source doc `meta.status !== 0` check excludes disabled/archived docs. **Clear.**
- **C5 (build/runtime parity):** N/A — no build/extension/packaging changes.
- **C6 (governance/policy):** N/A — no policy documents changed.

## Cross-Round Blocker Re-check

## Verdict: CHANGES_REQUESTED

Four P1 blockers must be addressed before merge: (1) unbounded recursion in `domCellToBlocks` enabling stack-overflow DoS, (2) quoted-DDE bypass in formula safety, (3) relative URL corruption in `isSafeHref`, and (4) missing admission control on four new expensive endpoints. The P2 items are non-blocking but recommended for the same PR.

lml2468 (approved):
## Re-review: #1041 @ `d09599cb` — 空 commit 重触发 CI,维持 APPROVE

相对我上轮批准的 `b29e47f7`,唯一新增是 `d09599cb "chore: retrigger CI"` —— **空提交(`git diff b29e47f7 d09599cb` 为空,commit message 自陈 "Empty commit to retrigger GitHub Actions checks")**,零内容改动。

锚定 head `d09599cb` / merge-base `a8bad922`(behind 4)。

**结论:纯 CI 重触发,可合并 ✅**

- **workflow/spec 文件与 `b29e47f7` 字节完全一致** —— 无功能变化。
- **actionlint 实跑 4 个 workflow 全 exit 0**(重核确认)。
- 上轮已验证并修复的项全部延续:`[skip ci]` 已去除(baseline push 后 required `e2e-p0` 正常重跑)、`head_ref` 走 `env: HEAD_REF`(注入面收口)、两 baseline workflow `workflow_dispatch`-only + 权限最小化 + same-repo fork 硬门、e2e 门 fail-closed(交叉标 block/business·proxy exit 1/count-guard 31)。

🟡 非阻断(承前未变):`paths:`+required 前瞻死锁(main 未设保护不触发)、浮动 action tag vs SHA-pin、`@visual`-only 失败分类较宽、三 workflow 常量重复(mochashanyao)。

空提交无审查面,上轮 APPROVE 结论完全延续。**维持 APPROVE。**

mochashanyao (approved):

PR #23, opened 2026-05-14
fix(matter): 智能创建事项弹窗优化 + 修复重复创建

quanming1:
## 问题

"智能创建事项"弹窗存在多个 UX 问题和一个逻辑 bug：

1. 点击"创建事项"按钮时会重复创建事项（后端在 extractMatter 时已创建）
2. 弹窗 padding 过大
3. 点击蒙版会意外关闭弹窗
4. 负责人下拉框被弹窗 overflow:hidden 裁切
5. 按钮文案和布局不符合产品需求

### 逻辑修复
- `extractMatter` API 返回的 `id` 表示后端已创建事项
- "确认事项"按钮改为调用 `updateMatter` + `addAssignee`，不再重复 `createMatter`
- 空白新建模式（无 AI 提取）仍走正常创建流程

### 弹窗优化
- "创建事项"按钮改为"确认事项"
- 删除"取消"按钮
- 设置当前用户为默认负责人
- 减小弹窗 padding（24px → 20px）
- 移除副标题文案（"从 N 条选中消息蒸馏..."）
- `maskClosable={false}` 防止误触关闭
- Semi Modal 各层级添加 `overflow: visible` 修复下拉框裁切

PR #1037, opened 2026-07-23
fix(dmworkmcp): align MCP marketplace UI to Skill marketplace (#1009)

l-s-c:
## Summary

Aligns the MCP marketplace page (list card, detail modal, top toolbar) with the Skill marketplace so the two tabs render as siblings under one shell. Adds a Bot 上架 dropdown / prompt modal mirroring Skill's, backports the tag / title overflow guards from PR #1026, and drops the 来源 segmented filter (人工/Bot were identical to users).

Highlights:

- **McpCard**: 2-line title clamp, 48px icon, 12px card padding, rounded-pill accent tags. Empty-icon fallback = hash-color square + first 2 name chars. Owner meta row shows `<Bot> botname · <UserRound> humanname` for bot rows and `<UserRound> humanname` for human rows; `import` rows show no owner chip. MatchReasons chips restored for tool / usage / creator hits (fields the card body doesn't already show).
- **McpDetailModal**: custom WKModal header (icon + h3 + category chip + owner/updated meta row). Slogan / stats / tags moved into the body under the header. `formatUpdatedTime` guards future timestamps. Category badge falls back to the raw slug when i18n is missing.
- **Top toolbar**: `上架 MCP` primary button + dropdown (`Bot 上架` / `手动上架`), matching Skill's shape. 来源筛选器整段删除。
- **Bot publish flow**: new `McpBotPublishModal` + `mcpBotPublishPrompt`. Prompt content is verified against octo-cli's embedded `octo-marketplace` Skill (`skills/octo-marketplace/mcp.md`) — uses `marketplace mcp create` / `mcp probe` / `mcp-category list`, not the Skill-specific `bot skills publish`. Handles clipboard rejection, missing-space-id, and timer-cleanup edge cases.
- **Tag validation (PR #1026 backport)**: `MAX_MCP_TAGS = 10`, `MAX_MCP_TAG_LENGTH = 24`, allowlist matches Skill's. TagsInput splits paste on commas, per-tag validation gates commit, `handleSubmit` re-validates the full array so legacy oversized rows can't save silently.

## Linked Spec

- Fixes https://github.com/Mininglamp-OSS/octo-web/issues/1009 (\"[UI一致性] MCP市场页面UI需与Skill市场保持一致\")
- Refs https://github.com/Mininglamp-OSS/octo-web/issues/1017 (Skill tag/title overflow), backported via https://github.com/Mininglamp-OSS/octo-web/pull/1026

## How verified

[...]

PR #1043, opened 2026-07-23
fix(dmworkmcp): align MCP marketplace UI to Skill + tag chip filter (#1009)

### LINKED ISSUE #1009: [UI一致性] MCP市场页面UI需与Skill市场保持一致，且需适配大屏
## 问题描述

MCP 市场页面的 UI 样式与 Skill 市场页面不一致，涉及分类标签、搜索框、按钮、筛选器、卡片样式、字体大小等多处差异；此外 MCP 页面在大屏下两侧留白过多，没有自适应宽度。

## 需要对齐的部分

对照 Skill 市场页面（见下方参考截图），MCP 市场页面需要调整以下内容：

### 1. 分类 Tab / 分类标签
- MCP 页面分类 tab 样式与 Skill 页面不一致，需统一标签样式（圆角、间距、选中态颜色/下划线等）

### 2. 搜索框
- 搜索框 placeholder 文字、高度、边框样式、图标位置需与 Skill 页搜索框对齐

### 3. 「新建 MCP」/「上架」按钮
- 按钮位置、大小、颜色、文字风格需与 Skill 页「上架 Skill」按钮统一

### 4. 来源筛选器
- 筛选/排序下拉框样式与 Skill 页不一致，需对齐组件样式

### 5. 卡片样式
- MCP 卡片的整体布局、padding、圆角、阴影需与 Skill 卡片保持一致
- 标签（tag/badge）样式：背景色、圆角、字号、间距对齐
- 版本号、作者名、描述文字的字体大小和颜色统一

### 6. 作者信息显示
- 作者名显示格式、图标、位置与 Skill 卡片对齐
- （注意之前提过的issue：[#1004](https://github.com/Mininglamp-OSS/octo-web/issues/1004) 作者应显示发布者bot名而非space owner人名）

### 7. 字体大小
- 卡片标题、描述、标签、按钮文字的字号需和 Skill 页面统一

### 8. 大屏适配
- MCP 页面在大屏（如 4K/2K 显示器）下，内容区域两侧留白过多，卡片没有自适应宽度/网格布局扩展
- 应与 Skill 页面一样，大屏时卡片网格自动多列填充，减少空白

## 截图

**MCP 市场页面（标注了需调整的位置）：**

**Skill 市场页面（参考样式）：**

## 备注
- 建议先完成 Skill 页面作者显示修复（#1004）后再统一对齐 MCP 页面
- 两个市场页面应共享同一套卡片组件和筛选组件，避免后续样式再次分叉

l-s-c:
## Summary

Bring the MCP marketplace page in line with the Skill marketplace (issue #1009), backport the tag/title overflow guards from PR #1026, and wire the new tag chip filter to the backend `/mcp_tags` endpoint added in [octo-marketplace #23](https://github.com/Mininglamp-OSS/octo-marketplace/pull/23).

The commit history reflects three natural chapters:

1. **UI alignment (#1009 body)** — McpCard 2-line owner row (`🐳 bot · 👤 human` for bot rows, `👤 human` for human rows, hidden for import rows), 48px icon + hash-color fallback + 2-char initials, rounded-pill accent tags, MatchReasons chips restored for creator hits only, McpDetailModal custom WKModal header (icon + h3 + category chip + owner/updated meta), publish button → `上架 MCP` primary + Upload icon + dropdown (Bot 上架 / 手动上架), Bot 上架 prompt modal referencing octo-cli's `octo-marketplace` Skill's mcp.md workflow.
2. **PR #1026 backport** — MCP tag input rejects `>10` tags / `>24 chars` / invalid chars with Toast, paste splits on comma, McpCreateModal.handleSubmit re-validates on save (guards legacy oversized rows), tag chip inside input ellipses long tags.
3. **Tag chip filter** — new popover in the search bar mirroring Skill's SearchBar. Reads suggestions from `GET /mcp_tags` (debounced 160ms, AbortController-cancelled), forwards `?mode=mine` when opened from the 我的 tab, memoized `tagRows` on the class instance keyed on `(tagsSelected, tagSuggestions)` identity, `focus({ preventScroll: true })` on open to avoid tripping the infinite-scroll listener, resets state on space + mode change. Selected tags round-trip through the URL as repeated `?tag=X` params (avoids splitting on any tag literal containing a comma).

Also removed the 来源 (人工/Bot) segmented filter per product feedback — both provenance values render identically to end users.

## Linked Spec

Fixes #1009. Refs #1017 + PR #1026 (Skill tag/title overflow backport), and the sibling backend contract at [octo-marketplace #23](https://github.com/Mininglamp-OSS/octo-marketplace/pull/23) for narrow keyword search + AND tag semantics + `/mcp_tags`.

## How verified

Local Vite dev server (`localhost:5174`) proxied to `im-test.deepminer.com.cn`. Full E2E walkthrough:

- List renders 41 real MCPs; every provenance case (bot / human / empty-icon fallback / long-title) displays correctly
- Category chip `开发工具` → 12 rows; count updates + refetch works
- Search `linear` → 2 rows (Linear Sync Bot + Linear MCP); MatchReasons chip appears only when creator matches
- Tag popover: opens, search input focused, empty state \"还没有可选标签\" because the sibling backend PR hasn't deployed to test env yet — once it does, suggestions populate from `/mcp_tags`
- Publish dropdown: opens both `Bot 上架` (copy-prompt modal, 1783-char prompt with real Space ID + API base URL) and `手动上架` (create form) paths
- Detail modal for Slack Notifier Bot: `[icon] Slack Notifier Bot [效率协作]` header + `🐳 Slack Notifier · 👤 Developer · 1 天前更新` meta + slogan + `🔧 1` stats + tags at the bottom
- Create form: tag input rejects `>10` tags, `>24` char tags, non-allowlisted chars, paste-split `a,b,c` → 3 chips
- No new TS errors introduced vs. baseline

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

   - **McpCard / McpDetailModal**: DOM + CSS refactor to match Skill layout. Same wire contract (`McpListItem`, `McpDetail`); the resolveOwner helper picks the display shape per `createdByType` (bot/human/import → chip / chip / no-chip).
   - **TagsInput in McpCreateModal**: adds `validateMcpTags(form.tags)` guard on submit + inline per-tag validation; a legacy record with >10 or over-length tags now blocks save until fixed.
   - **List filter API contract**: `params.tags` is now serialised as repeated `?tag=X` params (was comma-joined). Backend `splitQuery` accepts both, so a mid-migration overlap is safe.
   - **New `fetchMcpTags` API call**: hits `GET /market/api/v1/mcp_tags?q=...&limit=100&mode=mine` when the popover opens. Fails silently (empty popover) if backend isn't up.
   - **`mcpListQuery`**: URL round-trips `?tag=X` params so bookmarked filtered views hydrate cleanly.
   - **Removed**: `createdByType` state + `SOURCE_FILTER_OPTIONS` + the source segmented filter UI. The wire field stays on responses; only the client-side filter is gone.

2. **What could break** because of it (dependents + failure mode)?

   - **Bot 上架 prompt correctness**: hard-codes `octo-cli marketplace mcp create` and references the `octo-marketplace` Skill's `mcp.md`. Verified against octo-cli's embedded Skill catalog. If those commands ever rename, the single template string in `utils/mcpBotPublishPrompt.ts` needs a bump.
   - **Tag chip filter on old backend**: octo-marketplace PR must merge + deploy before the popover fills in with real suggestions. Meanwhile the popover shows an empty state; the underlying `?tag=` filter still works if the user selects a chip from a previously-loaded state.
   - **Rebase surface**: this branch was rebased on top of a 7-commit upstream/main window (dmworksummary + docs + globalsearch refactors). No overlap with dmworkmcp files but full-branch force-push happened.
   - **Legacy tag validation UX**: a user opening an existing MCP with `>10` tags will see save blocked and a Toast until they trim; documented in the file comment.
   - **Clipboard availability**: Bot 上架 modal toasts on clipboard rejection or missing `navigator.clipboard`, so users on http:// origins or older WebViews see feedback instead of a silent no-op.

3. **How do you know it works** (specific test/repro/trace)?

   - Full E2E walkthrough against test env (see \"How verified\") — every provenance case, every filter, every modal opened and read
   - Cross-repo code-review pass (5 parallel agents on both branches) applied all findings inline; a second review-pass found 4 more (mode passthrough, sort=updated, comma-tag safety, memo/preventScroll) — all applied in the same diff
   - Vite HMR compiles every touched file; TS baseline errors unchanged (pre-existing React 17/18 type resolution issues)
   - Manual DOM inspection: `resolveOwner` gates the import branch correctly (returns null; card meta-row collapses cleanly); MatchReasons component only fires for `creator` hits after the sibling backend change; category badge falls back to raw slug if i18n misses
   - Load-bearing edge cases explicitly tested during review: whitespace-only icon → fallback avatar; whitespace-only tag → rejected at add; future timestamp → falls to absolute date format not \"just now\"

### DIFF
--- a/.i18n/scan-config.json
+++ b/.i18n/scan-config.json
@@ -61,6 +61,10 @@
       "reason": "Agent prompt template for the Bot publish flow. Same rationale as installPrompt.ts above — generated prompt content pasted into an agent client, must byte-match; not translatable UI chrome."
     },
     {
+      "path": "packages/dmworkmcp/src/utils/mcpBotPublishPrompt.ts",
+      "reason": "Agent prompt template for the MCP marketplace publish flow. Same rationale as dmworkskillmarket/src/utils/botPublishPrompt.ts — generated prompt content pasted into an agent client, must byte-match; not translatable UI chrome."
+    },
+    {
       "path": "packages/dmworksummary/src/__mocks__/handlers.ts",
       "reason": "MSW test fixture data with sample user names, titles, and message bodies; not UI copy."
     },
--- a/packages/dmworkmcp/package.json
+++ b/packages/dmworkmcp/package.json
@@ -11,7 +11,8 @@
     "@douyinfe/semi-icons": "^2.93.0",
     "@douyinfe/semi-ui": "^2.93.0",
     "axios": "^0.25.0",
-    "classnames": "^2.3.1"
+    "classnames": "^2.3.1",
+    "lucide-react": "^0.577.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^6.9.1",
--- a/packages/dmworkmcp/src/api/mcpService.paramsSerializer.test.ts
+++ b/packages/dmworkmcp/src/api/mcpService.paramsSerializer.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it, vi } from "vitest";
+
+// serializeMcpParams is pure and doesn't call t(), but the file's module
+// graph pulls in @octo/base at import-time via other exports. Stub the
+// module so the test doesn't depend on i18n / IM SDK bootstrap.
+vi.mock("@octo/base", () => ({
+  WKApp: { apiClient: { config: {} }, loginInfo: {}, shared: { currentSpaceId: "" }, mittBus: { on: () => {}, off: () => {}, emit: () => {} } },
+  buildAcceptLanguage: () => "en-US",
+  t: (key: string) => key,
+  DEFAULT_REQUEST_TIMEOUT_MS: 20000,
+}));
+
+import { serializeMcpParams } from "./mcpService";
+
+describe("serializeMcpParams — axios paramsSerializer wire contract", () => {
+  it("emits `?a=1&a=2` for arrays (never `?a[]=1&a[]=2`)", () => {
+    // gin's QueryArray on the marketplace backend only recognises the plain
+    // repeat form — a revert to axios 0.25's default `?a[]=...` shape would
+    // make every multi-tag filter silently return wrong results. Pin.
+    expect(serializeMcpParams({ tag: ["a", "b"] })).toBe("tag=a&tag=b");
+  });
+
+  it("URL-encodes commas inside a value so comma-bearing tags survive", () => {
+    expect(serializeMcpParams({ tag: ["v1.0,beta"] })).toBe("tag=v1.0%2Cbeta");
+  });
+
+  it("preserves multiple keys and multi-value ordering", () => {
+    expect(
+      serializeMcpParams({
+        q: "hello",
+        tag: ["a", "b"],
+        mode: "mine",
+      })
+    ).toBe("q=hello&tag=a&tag=b&mode=mine");
+  });
+
+  it("drops null / undefined values entirely", () => {
+    expect(
+      serializeMcpParams({
+        q: undefined,
+        limit: null,
+        mode: "mine",
+      })
+    ).toBe("mode=mine");
+  });
+
+  it("drops null / undefined inside arrays (keeps a stable position for the rest)", () => {
+    expect(
+      serializeMcpParams({
+        tag: ["a", undefined as unknown as string, "b", null as unknown as string, "c"],
+      })
+    ).toBe("tag=a&tag=b&tag=c");
+  });
+
+  it("handles empty / undefined params argument", () => {
+    expect(serializeMcpParams(undefined)).toBe("");
+    expect(serializeMcpParams({})).toBe("");
+    expect(serializeMcpParams({ tag: [] })).toBe("");
+  });
+
+  it("stringifies non-string primitives (limit, offset)", () => {
+    expect(serializeMcpParams({ limit: 50, offset: 0 })).toBe("limit=50&offset=0");
+  });
+});
--- a/packages/dmworkmcp/src/api/mcpService.ts
+++ b/packages/dmworkmcp/src/api/mcpService.ts
@@ -134,6 +134,7 @@ async function fetchMcpListMockFiltered(
   const keyword = (params.keyword ?? "").trim().toLowerCase();
   const category = params.category ?? "all";
   const createdBy = params.createdByType;
+  const tags = params.tags ?? [];
   const filtered = source.filter((item) => {
     const matchCategory = category === "all" || item.category === category;
     const matchKeyword =
@@ -144,7 +145,11 @@ async function fetchMcpListMockFiltered(
     // read-side default the wire mapper applies for pre-#894 records.
     const rowType = item.createdByType ?? "human";
     const matchCreatedBy = !createdBy || rowType === createdBy;
-    return matchCategory && matchKeyword && matchCreatedBy;
+    // Multi-tag filter is AND: a row must carry every selected tag. Mirrors
+    // the backend semantics in octo-marketplace (mcp-v1.md §4.2).
+    const matchTags =
+      tags.length === 0 || tags.every((tag) => item.tags.includes(tag));
+    return matchCategory && matchKeyword && matchCreatedBy && matchTags;
   });
   const offset = params.offset && params.offset > 0 ? params.offset : 0;
   const limit =
@@ -675,6 +680,12 @@ async function fetchMcpListPath(
   // Relevance is only meaningful with a keyword — every row scores 0 otherwise,
   // making the sort order arbitrary. When browsing, surface freshest first.
   query.sort = keyword ? "relevance" : "updated";
+  // Multi-tag filter — mcp-v1.md §4.2 accepts `tag` as repeatable OR
+  // comma-separated. Use comma-separated for compact wire; the backend
+  // AND-combines the parsed values.
+  if (params.tags?.length) {
+    query.tag = params.tags.join(",");
+  }
   const pageSize = params.limit && params.limit > 0 ? params.limit : 20;
   query.page_size = pageSize;
   query.page = Math.floor((params.offset ?? 0) / pageSize) + 1;
@@ -876,6 +887,62 @@ export function fetchMcpMine(
 
 export function fetchMcpDetail(id: string): Promise<McpDetail> {
   return USE_MOCK ? fetchMcpDetailMock(id) : fetchMcpDetailReal(id);
+}
+
+/** One tag suggestion in the tag-filter popover (mcp-v1.md §4.8). */
+export interface McpTagSuggestion {
+  name: string;
+  count: number;
+}
+
+/** GET /market/api/v1/mcp_tags — tag suggestions for the search-bar
+ *  popover. Backend aggregates across the caller's visible set (system +
+ *  space rows), sorted by descending count. Empty `query` returns every
+ *  visible tag; the backend clamps `limit` to [1, 100] with default 50. */
+export function fetchMcpTags(
+  query = "",
+  opts: { signal?: AbortSignal; limit?: number } = {}
+): Promise<McpTagSuggestion[]> {
+  return USE_MOCK
+    ? fetchMcpTagsMock(query, opts)
+    : fetchMcpTagsReal(query, opts);
+}
+
+async function fetchMcpTagsMock(
+  query: string,
+  _opts: { signal?: AbortSignal; limit?: number }
+): Promise<McpTagSuggestion[]> {
+  // Aggregate from the in-memory mock list — same visibility semantics as
+  // the real backend (everything visible to the caller's space). Count-desc
+  // + name-asc tie-break, case-insensitive query.
+  const counts = new Map<string, number>();
+  for (const item of MOCK_MCP_LIST) {
+    for (const tag of item.tags) {
+      counts.set(tag, (counts.get(tag) ?? 0) + 1);
+    }
+  }
+  const kw = query.trim().toLowerCase();
+  const items: McpTagSuggestion[] = [];
+  counts.forEach((count, name) => {
+    if (kw && !name.toLowerCase().includes(kw)) return;
+    items.push({ name, count });
+  });
+  items.sort((a, b) => (b.count - a.count) || a.name.localeCompare(b.name));
+  return delay(items);
+}
+
+async function fetchMcpTagsReal(
+  query: string,
+  opts: { signal?: AbortSignal; limit?: number }
+): Promise<McpTagSuggestion[]> {
+  const params: Record<string, unknown> = {};
+  if (query.trim()) params.q = query.trim();
+  if (opts.limit && opts.limit > 0) params.limit = opts.limit;
+  const resp = await mcpAxios.get<{ data: McpTagSuggestion[] }>(
+    `${BASE}/mcp_tags`,
+    { params, signal: opts.signal }
+  );
+  return resp.data.data ?? [];
 }
 
 /**
--- a/packages/dmworkmcp/src/api/mcpService.ts
+++ b/packages/dmworkmcp/src/api/mcpService.ts
@@ -681,10 +681,12 @@ async function fetchMcpListPath(
   // making the sort order arbitrary. When browsing, surface freshest first.
   query.sort = keyword ? "relevance" : "updated";
   // Multi-tag filter — mcp-v1.md §4.2 accepts `tag` as repeatable OR
-  // comma-separated. Use comma-separated for compact wire; the backend
-  // AND-combines the parsed values.
+  // comma-separated. We use REPEATED params (`tag=a&tag=b&tag=c`) so a
+  // tag value that happens to contain a comma still round-trips intact;
+  // comma-joining would let the backend re-split "v1.0,beta" into two
+  // wrong tags and produce silent empty results.
   if (params.tags?.length) {
-    query.tag = params.tags.join(",");
+    query.tag = params.tags;
   }
   const pageSize = params.limit && params.limit > 0 ? params.limit : 20;
   query.page_size = pageSize;
@@ -898,10 +900,12 @@ export interface McpTagSuggestion {
 /** GET /market/api/v1/mcp_tags — tag suggestions for the search-bar
  *  popover. Backend aggregates across the caller's visible set (system +
  *  space rows), sorted by descending count. Empty `query` returns every
- *  visible tag; the backend clamps `limit` to [1, 100] with default 50. */
+ *  visible tag; the backend clamps `limit` to [1, 100] with default 50.
+ *  Pass `mode: "mine"` when the popover opens from the "我的" tab so
+ *  suggestions match `GET /mcps/mine`. */
 export function fetchMcpTags(
   query = "",
-  opts: { signal?: AbortSignal; limit?: number } = {}
+  opts: { signal?: AbortSignal; limit?: number; mode?: "all" | "mine" } = {}
 ): Promise<McpTagSuggestion[]> {
   return USE_MOCK
     ? fetchMcpTagsMock(query, opts)
@@ -910,7 +914,7 @@ export function fetchMcpTags(
 
 async function fetchMcpTagsMock(
   query: string,
-  _opts: { signal?: AbortSignal; limit?: number }
+  _opts: { signal?: AbortSignal; limit?: number; mode?: "all" | "mine" }
 ): Promise<McpTagSuggestion[]> {
   // Aggregate from the in-memory mock list — same visibility semantics as
   // the real backend (everything visible to the caller's space). Count-desc
@@ -933,11 +937,12 @@ async function fetchMcpTagsMock(
 
 async function fetchMcpTagsReal(
   query: string,
-  opts: { signal?: AbortSignal; limit?: number }
+  opts: { signal?: AbortSignal; limit?: number; mode?: "all" | "mine" }
 ): Promise<McpTagSuggestion[]> {
   const params: Record<string, unknown> = {};
   if (query.trim()) params.q = query.trim();
   if (opts.limit && opts.limit > 0) params.limit = opts.limit;
+  if (opts.mode === "mine") params.mode = "mine";
   const resp = await mcpAxios.get<{ data: McpTagSuggestion[] }>(
     `${BASE}/mcp_tags`,
     { params, signal: opts.signal }
--- a/packages/dmworkmcp/src/api/mcpService.ts
+++ b/packages/dmworkmcp/src/api/mcpService.ts
@@ -326,6 +326,31 @@ function slugify(s: string): string {
 // <origin>/market/api/v1 (nginx / vite proxy strips the /market prefix to the
 // service's own /api/v1), mirroring the summary + matter service convention.
 
+/** Serialise axios request params as repeated keys (`?a=1&a=2`) instead of
+ *  axios 0.25's default `?a[]=1&a[]=2`. gin's QueryArray on the marketplace
+ *  backend only recognises the plain-repeat form; a bracketed key would
+ *  silently become a single-string param that never matches. Also drops
+ *  undefined/null keys so callers can just pass an optional value without
+ *  pre-filtering. Exported so the wire contract can be pinned in unit
+ *  tests without spinning up an axios instance. */
+export function serializeMcpParams(
+  params: Record<string, unknown> | undefined
+): string {
+  const usp = new URLSearchParams();
+  for (const [key, value] of Object.entries(params ?? {})) {
+    if (value === undefined || value === null) continue;
+    if (Array.isArray(value)) {
+      for (const item of value) {
+        if (item === undefined || item === null) continue;
+        usp.append(key, String(item));
+      }
+    } else {
+      usp.append(key, String(value));
+    }
+  }
+  return usp.toString();
+}
+
 const mcpAxios = axios.create({
   baseURL: "",
   // Isolated instance (no shared interceptors), so it never picks up the
@@ -333,27 +358,7 @@ const mcpAxios = axios.create({
   // the same ceiling explicitly to avoid the UI-hang class of bug that
   // DEFAULT_REQUEST_TIMEOUT_MS was introduced to close.
   timeout: DEFAULT_REQUEST_TIMEOUT_MS,
-  // Serialise array params as repeated keys (`?a=1&a=2`) instead of axios
-  // 0.25's default `?a[]=1&a[]=2`. gin's QueryArray on the marketplace
-  // backend only recognises the plain-repeat form; a bracketed key would
-  // silently become a single-string param that never matches. Also drops
-  // undefined/null keys so callers can just pass an optional value without
-  // pre-filtering.
-  paramsSerializer: (params) => {
-    const usp = new URLSearchParams();
-    for (const [key, value] of Object.entries(params ?? {})) {
-      if (value === undefined || value === null) continue;
-      if (Array.isArray(value)) {
-        for (const item of value) {
-          if (item === undefined || item === null) continue;
-          usp.append(key, String(item));
-        }
-      } else {
-        usp.append(key, String(value));
-      }
-    }
-    return usp.toString();
-  },
+  paramsSerializer: serializeMcpParams,
 });
 
 const BASE = "/market/api/v1";
@@ -914,13 +919,22 @@ export function fetchMcpTags(
 
 async function fetchMcpTagsMock(
   query: string,
-  _opts: { signal?: AbortSignal; limit?: number; mode?: "all" | "mine" }
+  opts: { signal?: AbortSignal; limit?: number; mode?: "all" | "mine" }
 ): Promise<McpTagSuggestion[]> {
-  // Aggregate from the in-memory mock list — same visibility semantics as
-  // the real backend (everything visible to the caller's space). Count-desc
-  // + name-asc tie-break, case-insensitive query.
+  // Aggregate from the in-memory mock list. `mode: "mine"` mirrors
+  // fetchMcpMineMock's owner filter (creatorName === caller). Signal +
+  // limit honored for parity with the real backend so the mock behaves
+  // the same in dev harnesses; the mock body is otherwise unreachable in
+  // production (USE_MOCK is a `const false` at :59).
+  if (opts.signal?.aborted) {
+    throw new DOMException("Aborted", "AbortError");
+  }
+  const source =
+    opts.mode === "mine"
+      ? MOCK_MCP_LIST.filter((it) => it.creatorName === (WKApp.loginInfo?.name || ""))
+      : MOCK_MCP_LIST;
   const counts = new Map<string, number>();
-  for (const item of MOCK_MCP_LIST) {
+  for (const item of source) {
     for (const tag of item.tags) {
       counts.set(tag, (counts.get(tag) ?? 0) + 1);
     }
@@ -932,7 +946,16 @@ async function fetchMcpTagsMock(
     items.push({ name, count });
   });
   items.sort((a, b) => (b.count - a.count) || a.name.localeCompare(b.name));
-  return delay(items);
+  // Backend clamps `limit` to [1, 100] with default 50.
+  const limit = Math.min(100, Math.max(1, opts.limit ?? 50));
+  const trimmed = items.slice(0, limit);
+  return new Promise<McpTagSuggestion[]>((resolve, reject) => {
+    const timer = setTimeout(() => resolve(trimmed), MOCK_DELAY_MS);
+    opts.signal?.addEventListener("abort", () => {
+      clearTimeout(timer);
+      reject(new DOMException("Aborted", "AbortError"));
+    });
+  });
 }
 
 async function fetchMcpTagsReal(
@@ -943,11 +966,15 @@ async function fetchMcpTagsReal(
   if (query.trim()) params.q = query.trim();
   if (opts.limit && opts.limit > 0) params.limit = opts.limit;
   if (opts.mode === "mine") params.mode = "mine";
-  const resp = await mcpAxios.get<{ data: McpTagSuggestion[] }>(
-    `${BASE}/mcp_tags`,
-    { params, signal: opts.signal }
-  );
-  return resp.data.data ?? [];
+  // Route through the shared get<T>() helper so 4xx/5xx/network failures
+  // are classified into McpListError like every other read endpoint. The
+  // helper re-throws axios cancels unchanged, so the caller's abort branch
+  // still fires. Array.isArray guard covers a non-list envelope (`{}` or
+  // `{data:null}`) surfacing as an empty suggestions list.
+  const data = await get<McpTagSuggestion[] | null>(`/mcp_tags`, params, {
+    signal: opts.signal,
+  });
+  return Array.isArray(data) ? data : [];
 }
 
 /**
--- a/packages/dmworkmcp/src/components/McpBotPublishModal.tsx
+++ b/packages/dmworkmcp/src/components/McpBotPublishModal.tsx
@@ -1,0 +1,127 @@
+import React, { useEffect, useMemo, useRef, useState } from "react";
+import { Bot, Check, Copy } from "lucide-react";
+import { Toast } from "@douyinfe/semi-ui";
+import { t, useI18n, WKApp, WKButton, WKModal } from "@octo/base";
+import {
+  getMcpBotPublishPrompt,
+  resolveMcpAPIBaseURL,
+} from "../utils/mcpBotPublishPrompt";
+
+interface McpBotPublishModalProps {
+  visible: boolean;
+  onClose: () => void;
+}
+
+function getCurrentSpaceId(): string {
+  return (
+    WKApp.shared?.currentSpaceId ||
+    (typeof localStorage !== "undefined"
+      ? localStorage.getItem("currentSpaceId") || ""
+      : "")
+  );
+}
+
+/** MCP "Bot 上架" modal — mirrors dmworkskillmarket's BotPublishModal.
+ *  Renders a copy-to-clipboard prompt the user hands to an Agent. Prompt
+ *  content lives in ../utils/mcpBotPublishPrompt.ts and is MCP-specific. */
+export default function McpBotPublishModal({
+  visible,
+  onClose,
+}: McpBotPublishModalProps) {
+  useI18n();
+  const [copied, setCopied] = useState(false);
+  const copiedTimerRef = useRef<number | null>(null);
+  const spaceId = getCurrentSpaceId();
+  // Memoize the prompt — inputs are stable for the modal's lifetime and the
+  // template is a few hundred bytes of string concat we'd otherwise rebuild
+  // on every parent re-render (space-changed, tab switches, etc.).
+  const prompt = useMemo(
+    () =>
+      getMcpBotPublishPrompt({
+        spaceId,
+        apiBaseUrl: resolveMcpAPIBaseURL(
+          WKApp.apiClient.config.apiURL,
+          window.location.origin
+        ),
+      }),
+    [spaceId]
+  );
+  // Placeholder guard: the prompt substitutes `<space-id>` when currentSpaceId
+  // is empty. Copying that would give the bot an unusable command — refuse.
+  const promptReady = Boolean(prompt) && !!spaceId.trim();
+
+  useEffect(() => {
+    if (visible) setCopied(false);
+  }, [visible]);
+
+  // Clear any in-flight copy-feedback timer when the modal unmounts to avoid
+  // "state update on unmounted component" warnings when the parent closes the
+  // modal within 2s of a successful copy.
+  useEffect(() => {
+    return () => {
+      if (copiedTimerRef.current !== null) {
+        window.clearTimeout(copiedTimerRef.current);
+        copiedTimerRef.current = null;
+      }
+    };
+  }, []);
+
+  async function handleCopy() {
+    if (!promptReady) return;
+    if (!navigator.clipboard?.writeText) {
+      Toast.error(t("mcp.botPublish.copyUnavailable"));
+      return;
+    }
+    try {
+      await navigator.clipboard.writeText(prompt);
+    } catch {
+      Toast.error(t("mcp.botPublish.copyFailed"));
+      return;
+    }
+    setCopied(true);
+    if (copiedTimerRef.current !== null) {
+      window.clearTimeout(copiedTimerRef.current);
+    }
+    copiedTimerRef.current = window.setTimeout(() => {
+      setCopied(false);
+      copiedTimerRef.current = null;
+    }, 2000);
+  }
+
+  return (
+    <WKModal
+      visible={visible}
+      onCancel={onClose}
+      title={null}
+      size="lg"
+      className="wk-mcp-prompt-modal"
+      header={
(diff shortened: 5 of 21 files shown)


Jerry-Xin (dismissed):
MCP marketplace UI alignment + new tag-chip filter. Reviewed the true delta versus the approved #1037 head (this branch iterated #1037, now closed): 6 files changed, concentrated in the tag-filter feature. The other 10 files are byte-identical to the #1037 surface and are not re-litigated. Approving.

### ✅ Tag-chip filter (the main new surface) — clean
- **No injection**: tag values reach the backend only through the axios `params` object (`query.tag = params.tags` in `packages/dmworkmcp/src/api/mcpService.ts`, and `params.q` / `params.mode` in `fetchMcpTagsReal`). The request path is a fixed `${BASE}/mcp_tags`; no user-controlled value is interpolated into the path. Repeated `tag=` params (not comma-joined) keep comma-containing tag values intact.
- **XSS-safe**: chip labels, counts, and the `title` attribute are rendered as escaped React children/attributes (`{row.name}`, `{row.count}`, `title={row.name}`) in `packages/dmworkmcp/src/pages/McpMarketListPage.tsx`. No `dangerouslySetInnerHTML`.
- **Filter-state correctness**: AND semantics documented and mirrored in the mock (`tags.every(...)`); toggle add/remove correct; clear is guarded on empty; space-change and mode-change fully reset selections/popover/query/suggestions and cancel any in-flight fetch; the `loadMore` stale-response guard now also keys on `tagsSelected`; empty-suggestion state is handled.
- **No request storm**: `/mcp_tags` fetch is debounced 160ms and cancelled via `AbortController`; it fires only on popover-open or query-change (chip toggling triggers a single list refetch, guarded by the request-version check).
- **URL round-trip**: `packages/dmworkmcp/src/pages/mcpListQuery.ts` uses `URLSearchParams` `getAll("tag")` / `append("tag", …)` — proper encoding, comma-safe, and scrubs the removed `created_by_type` param.
- i18n keys under `mcp.filter.*` are all present (zh-CN + en-US); the `tagsSelected` `{{count}}` interpolation matches the call site.

### 💬 Non-blocking
- 🟡 No new tests for the added service contract / URL round-trip / debounce+cancel / filter toggle-clear-empty paths. Worth covering repeated `tag=` serialization, `mode=mine`, query parsing, and select/clear.
- 🟡 The mock `fetchMcpTagsMock` ignores `mode` and `limit` while accepting them, so mock-mode testing can't catch a "Mine"-tab regression. Low impact (real backend enforces both), but the parity comment overstates it.
- 🔵 Failed tag-suggestion fetches fall through silently. An observable empty/error state would help distinguish a backend/deploy failure from a genuinely empty tag catalog.

### ✅ Highlights
- Repeated query params correctly preserve comma-containing tag values.
- AbortController cancellation and the stale-list response guard are carefully done.
- URL hydration and removal of the obsolete provenance filter are consistent.
- Carry-over marketplace surface from #1037 (avatar fallback, clipboard failure handling, timestamp/legacy-tag guards) remains intact.

Note on style: the design-token rule (`var(--wk-*)`) is fully satisfied by this PR's diff — every color line added here uses a semantic token. Pre-existing raw-color usage lives only in the unchanged #1037 surface and is out of scope for this change.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1043 (octo-web)

Reviewed at head `cc9c2735`. Scope: replace single-select `createdByType` source filter with multi-select tag chip filter in MCP marketplace, add bot-publish prompt modal, align card/detail UI with Skill marketplace patterns.

## Summary

This PR is a substantial UI alignment of the MCP marketplace with the existing Skill marketplace. It replaces the `createdByType` source filter with a multi-select tag chip filter backed by a debounced `/mcp_tags` endpoint, adds a "Publish via Bot" modal that generates a copy-to-clipboard CLI prompt, rewrites `McpCard` with new avatar/owner/tag rendering, and enriches the detail modal with owner attribution and time formatting. The tag filter data flow (API → debounce → popover → selection → list query) is correctly wired end-to-end, auth flows through the shared `mcpAxios` interceptor on every new call, and the `AbortController` lifecycle is properly managed. Three non-blocking P2 items below.

## Verification

- ✅ **Auth consistency** — `fetchMcpTags` uses the same `mcpAxios` instance with token/space/language interceptor as all other MCP API calls. The 401 response interceptor applies uniformly.
- ✅ **Tag filter data flow** — `tagsSelected` state → `loadData()` → `fetchMcpListPath` → `paramsSerializer` → repeated `?tag=` keys, matching backend `gin QueryArray`. Verified the full round-trip from URL parse through API call.
- ✅ **Abort/cancel lifecycle** — `scheduleTagFetch_` creates a fresh `AbortController` per invocation, cancels the previous, checks controller identity post-await. Correct.
- ✅ **State cleanup** — `handleMode` resets `tagsSelected`, `tagFilterOpen`, `tagSuggestions`. Modal open/close effects properly reset all form state. No stale data leakage.
- ✅ **i18n completeness** — All new `t()` keys are defined in both `en-US.json` and `zh-CN.json` with matching interpolation variables. Removed keys have no remaining references.
- ✅ **Tag validation** — `MCP_TAG_PATTERN` regex is properly anchored (`^...$`), hyphen is literal (last position in class), Unicode flag enables `\p{L}\p{N}`. Enforced at both input-time and submit-time for defense-in-depth.
- Static analysis only at head `cc9c2735`; build and tests not executed in this environment.

## Findings

### P2 — `fetchMcpTagsReal` bypasses error-classification wrapper (`packages/dmworkmcp/src/api/mcpService.ts:948`)

Every other API function in this file routes through the `get<T>()` helper (line 439) which wraps in try/catch with `axios.isCancel()` re-throw and `classifyMcpListError` → `McpListError`. `fetchMcpTagsReal` calls `mcpAxios.get` directly without this wrapper. An Axios `CanceledError` from the abort signal is not a `DOMException`, so the caller's `err instanceof DOMException && err.name === "AbortError"` guard at `McpMarketListPage.tsx:253` will not match — the abort falls through to the silent catch instead. Today this is benign because the catch block is a no-op, but if error handling is ever tightened this will be the inconsistent outlier.

**Fix direction:** Route through the existing `get<T>()` helper, or replicate the try/catch pattern:

```ts
try {
  const resp = await mcpAxios.get<...>(`${BASE}/mcp_tags`, { params, signal: opts.signal });
  return resp.data.data ?? [];
} catch (err) {
  if (axios.isCancel(err)) throw err;
  throw new McpListError(classifyMcpListError(err));
}
```

### P2 — Bot-publish prompt body is hardcoded Chinese, not i18n-aware (`packages/dmworkmcp/src/utils/mcpBotPublishPrompt.ts:27`)

The modal's UI chrome (title, hint, button labels) correctly uses `t("mcp.botPublish.*")` i18n keys. But the prompt content itself — the multi-paragraph Chinese instruction set the user copies and hands to a bot agent — is a hardcoded Chinese string literal with no localization path. If the app serves non-Chinese-speaking users, they receive an unreadable prompt they cannot use. The `en-US.json` has no equivalent key for the prompt body.

**Fix direction:** Add an i18n key for the prompt body template (e.g. `mcp.botPublish.promptBody`) with interpolation slots for `spaceId` and `apiBaseUrl`, so it can be translated alongside the rest of the UI.

### P2 — Modal title / button label copy mismatch (`packages/dmworkmcp/src/i18n/en-US.json:62`)

The trigger button now reads "Publish MCP" / "上架 MCP" (`mcp.list.create`), but the modal dialog title (`mcp.create.title`) remains "New MCP" / "新建 MCP". The user clicks "Publish MCP" and sees "New MCP" in the dialog header — inconsistent UX copy.

**Fix direction:** Align `mcp.create.title` to "Publish MCP" / "上架 MCP", or use a neutral title in both places.

## Human-verify

1. The bot-publish prompt instructs the bot to run `octo-cli auth login --with-token` via stdin. Confirm the `octo-marketplace` Skill's `mcp.md` referenced in step 3 actually exists in the shipped CLI version and the command surface (`marketplace mcp create`, `marketplace mcp probe`, `marketplace mcp-category list`) matches. This is not a merge blocker for the web PR — flagging for CLI-side contract confirmation.

- `McpCard` rewrite: `resolveOwner` correctly handles all three `createdByType` variants; `item.tags` guarded by `raw.tags ?? []` in the API mapper; `item.icon` uses optional chaining.
- `McpDetailModal`: `canManage` properly gates edit/delete; `useEffect` on `mcpId` resets state with `cancelled` flag to prevent stale updates; `formatUpdatedTime` handles NaN and negative clock skew.
- `McpCreateModal` tag input: `commitOne` enforces `MAX_MCP_TAGS` (10), per-tag length (24), charset pattern, and deduplication. Submit-time re-validation via `validateMcpTags` provides defense-in-depth.
- `mcpAvatar.ts`: `hashString(id || "")` handles null/undefined; `visibleChars("")` returns `[]` → fallback `"M"`. Color distribution adequate for 6 slots.
- `mcpTagValidation.ts`: regex properly anchored, Unicode flag correct, tab/newline/zero-width-space/emoji all correctly rejected.
- `mcpListQuery.ts`: `parseMcpListQuery` uses `q.getAll("tag")` (repeated params), `serializeMcpListQuery` scrubs legacy `created_by_type`. Clean round-trip.
- CSS: all new class names referenced in components are defined in `index.css`; removed source-filter classes have no remaining references.

## Verdict: APPROVED

Well-structured UI alignment PR. The tag filter data flow, auth, abort lifecycle, state cleanup, and i18n coverage are all correctly implemented. The three P2 items (error-classification consistency, prompt i18n, button/title copy mismatch) are non-blocking improvements.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Three P2 items are code-quality/i18n/UX-copy improvements that don't block merge.

cnwenf (commented):
# Code Review — PR #1043 (Mininglamp-OSS/octo-web)

## Summary

This PR aligns the MCP marketplace page with the Skill marketplace (#1009): McpCard owner-row redesign with id-keyed hash-color avatar fallbacks, rounded-pill accent tags, a custom McpDetailModal header, a publish-button dropdown with a Bot-publish prompt modal, the PR #1026 tag/title overflow-guard backport, and a new tag-chip filter (debounced, AbortController-managed `/mcp_tags` suggestions, repeated `?tag=` URL round-trip, AND semantics). All three reviewers independently machine-verified the load-bearing wire/runtime contracts correct at head `cc9c2735` (56/56 shipped tests pass; repeated-`tag=` encoding, `mode=mine` propagation, URL codec, and `/mcp_tags` unwrap all confirmed). No P1 blockers, no security vulnerability, no data-loss or functional break. The dominant, high-confidence theme across all three reviews is that the new code ships with zero test coverage and CI never runs `vitest`; the remainder are narrow state-lifecycle and mock-parity gaps.

## Findings

### P2 — New code ships with zero test coverage — mutation-proven across every new surface (`packages/dmworkmcp/src/utils/mcpTagValidation.ts:15`, `mcpListQuery.ts:12`, `mcpService.ts:683`, `mcpAvatar.ts:25`, `McpMarketListPage.tsx:106/211`, `McpBotPublishModal.tsx:67`, `mcpBotPublishPrompt.ts:9`, `McpCard.tsx:57`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

Every piece of new logic in this PR has no tests, and each gap was proven unguarded by the mutation protocol (a contract-breaking change leaves the shipped 4-file/56-test suite green while a focused probe fails). Affected surfaces (deduplicated across the three reviews):

- `validateMcpTag`/`validateMcpTags` (`mcpTagValidation.ts:15`) — sole guard against invalid tag input. Two untested boundaries: (a) `tagLength = Array.from(value).length` is the file's reason for existing — a regression to `value.length` would wrongly reject valid 24-codepoint CJK tags (`𠮷`×24 has `.length` 48) at the boundary with an erroneous Toast + refused save; an off-by-one admits 11 tags / 25 chars. (b) `validateMcpTag("")` returns null, so `validateMcpTags(["",…])` clears the submit-time re-validation that exists to catch legacy bad rows, which would then be POSTed. (qwen3.8; also flagged by qwen3.7, glm5.2)
- Multi-tag wire serializer (`mcpService.ts:683`, custom `paramsSerializer` at `:342`) — load-bearing for the whole multi-tag filter: gin's `QueryArray` only recognizes the plain-repeat form; a revert to axios's default `?tag[]=a&tag[]=b` would make every multi-tag selection silently return wrong/empty results with no error. Mutation-confirmed (bracket-form revert leaves all 56 tests green; focused probe fails 4/4). Pin `serializer({tag:["a","b"]}) === "tag=a&tag=b"` and `serializer({tag:["v1.0,beta"]}) === "tag=v1.0%2Cbeta"`. (qwen3.8; also qwen3.7, glm5.2)
- Tag URL codec (`mcpListQuery.ts:12`) — two untested boundaries: (a) round-trip fidelity for query-hostile chars the charset explicitly allows (`+`, `#`, `/`, space — `+`→space, `#`→fragment); a future manual-concat serialize would silently corrupt `c++`→`c  ` / drop `#net`. (b) `?tag=` parses to `[""]`, which flows into `tagsSelected`, re-emits an empty `tag=` param, renders a bogus chip, and since `item.tags.includes("")` is always false under AND, a shared URL intended to land unfiltered instead shows zero rows. Fix: filter blanks on parse; pin round-trip. (qwen3.8; also qwen3.7, glm5.2)
- `getMcpAvatarColor`/`getMcpAvatarText` (`mcpAvatar.ts:25/:55`) — pure functions; the headline claim (color keyed on id, not name, so same-name MCPs get distinct backgrounds) would silently regress to name-keying and stay green, as would `slice(0,2)`→`(0,1)` (`数据驱动`→`数`). Pin `""`→`"M"`, `数据库`→`数据`, `12345`→`12`, `my-cool-mcp`→`MC`, id-stable color. (qwen3.8; also qwen3.7, glm5.2)
- Tag-popover union + unwrap (`McpMarketListPage.tsx:211`) — `memoizedTagRows_` prepends selected tags absent from fresh suggestions (so a user can still unselect after the query changes); dropping the loop makes a selected tag vanish from the open popover. `resp.data.data ?? []` lacks an `Array.isArray` guard (`{}` envelope would throw). Both untested. (qwen3.8)
- `resolveOwner`/`parseMatchReason` (`McpCard.tsx:57`) — shared by McpCard and McpDetailModal; the import-returns-null contract (import rows must NOT show attribution) is untested. (qwen3.7, glm5.2)
- `fetchMcpTags` real path (`mcpService.ts:891`) — sort order, abort propagation, and envelope unwrap uncovered. (qwen3.7, glm5.2)
- `resolveMcpAPIBaseURL`/`getMcpBotPublishPrompt` (`mcpBotPublishPrompt.ts:9`) — sibling `dmworkskillmarket` has `botPublishPrompt.test.ts`; MCP has none. The `'不得输出 Token'` guard has no regression test — externally observable since the prompt is handed to a bot that executes shell commands, so a regression silently dropping the guard would produce an insecure bot run with no CI signal. (glm5.2; also qwen3.7)
- `formatUpdatedTime`/`resolveCategoryLabel` (`McpDetailModal.tsx:189`) — private closures with 5 behavioral branches (exactly 60000ms, 30-day threshold, negative clock skew, invalid ISO); the existing test mocks `WKModal` so the header never renders. (qwen3.7, glm5.2)
- McpMarketListPage lifecycle (`McpMarketListPage.tsx:106`) — document-listener attach/remove, debounce+abort ordering, space-change cleanup uncovered. (qwen3.7, glm5.2)
- McpBotPublishModal clipboard flow (`McpBotPublishModal.tsx:67`) — success/error/unavailable/unmount paths uncovered. (qwen3.7, glm5.2)
- TagsInput commit rewrite (`McpCreateModal.tsx:415`) — batch-split behavior (comma-split paste, partial-vs-all acceptance, `MAX_MCP_TAGS` submit guard) is an undocumented product decision no test documents. (qwen3.7, glm5.2)

### P2 — CI workflow never runs `pnpm test` / `vitest run` (`.github/workflows/ci.yml:92`) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The `build` job runs `pnpm install`, `i18n:check`, `build`, `lint`, and `lint:wkmodal` but never executes `pnpm test` or `vitest run` for any package. Yet `packages/dmworkmcp/package.json` declares `"test": "vitest run"` with `@testing-library/react` + `jest-dom` in devDependencies, and 56 tests pass locally. The suite exists and runs, but CI will never discover, execute, or gate on it — any new or existing test can break silently while the CI badge stays green. This compounds the coverage finding above: even the focused tests that already validate against real production code would not be run. Add a `pnpm test` (or `vitest run`) step to the `build` job — the highest-signal gap in this PR.

### P2 — `handleItemUpdated` leaves stale `updatedAt` and owner attribution after edit (`packages/dmworkmcp/src/pages/McpMarketListPage.tsx:8`) [source: octo-review-qwen3.7 + octo-review-glm5.2] (severity split: qwen3.7=P2, glm5.2=P3)

`handleItemUpdated` patches only `name`/`slogan`/`category`/`tags`/`toolCount`/`icon`/`visibility`/`creatorName` from the saved McpDetail into the list item; it does NOT patch `updatedAt`, `createdByType`, `createdByBotName`, or `matchReasons`. McpDetailModal (changed in this PR) now renders `updatedAt` via `formatUpdatedTime` and the owner row via `resolveOwner` (`createdByType`/`createdByBotName`). After a successful edit, `handleSaved` calls `handleItemUpdated` (not `loadData`), so re-opening the detail modal shows the pre-edit `updatedAt` and pre-edit owner attribution until a full reload. Add the missing fields to the override.

### P3 — `loadMore` stale-response guard collides on comma-bearing tags (`packages/dmworkmcp/src/pages/McpMarketListPage.tsx:324`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The stale-response guard keys the in-flight tag set via `tagsSelected.join(",")` — but the PR chose repeated `?tag=` params precisely because a tag may itself contain a comma (`v1.0,beta`). `join(",")` re-introduces that ambiguity: `["v1.0,beta"]` and `["v1.0","beta"]` produce the same key, so toggling between such sets while a page request is in flight makes the guard accept a stale page belonging to a different tag filter and append rows that don't match the current selection. Executed probe: items become `['i1','STALE_PAGE2']`; a delimiter-safe `JSON.stringify` key correctly drops it. The create-form validator excludes commas, but tags now come from the backend `/mcp_tags` response and `mcpTagValidation` is only imported in `McpCreateModal`, not the filter path, so practical risk is low but real. Fix: compare by length + per-element identity (or `JSON.stringify`) instead of `join(",")`.

### P3 — `/mcp_tags` fetch failures swallowed silently, bypassing the shared error classifier (`packages/dmworkmcp/src/pages/McpMarketListPage.tsx:251`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`scheduleTagFetch_`'s catch special-cases only `AbortError`: on a 5xx / network blip / auth edge, `tagSuggestions` stays at its previous value with no error indicator and no retry, so "fetch failed" is indistinguishable from "no tags exist" — the user may act on stale suggestions. The sibling `loadMore` in the same component surfaces the identical failure class via `Toast.error`, so the component's own de-facto contract is violated one method away. Compounding this, `fetchMcpTagsReal` does a raw `mcpAxios.get` returning `resp.data.data ?? []`, bypassing the shared `get<T>` helper that classifies OCTO error envelopes into `McpListError` — the only read endpoint to do so — so even a surfaced error would arrive as a raw axios string rather than the localized code every other read endpoint produces. Degraded-but-working (the selected-chip fallback keeps filtering functional). Fix: set an error state / Toast in the catch and route `fetchMcpTagsReal` through `get<T>`.

### P3 — `fetchMcpTagsMock` ignores `mode`/`limit`/`signal` (mock/real parity gap) (`packages/dmworkmcp/src/api/mcpService.ts:917`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The mock names its opts parameter `_opts` and aggregates from the full `MOCK_MCP_LIST`, ignoring `mode`, `limit`, and `signal`, while its comment claims "same visibility semantics as the real backend." The real path forwards `mode='mine'` + `limit` and passes `signal` to axios, so in mock mode the 我的 tab popover shows system-wide tags instead of caller-owned ones and never clamps to the limit — dev-mode tag suggestions diverge from production. Note (qwen3.8): unreachable in production — `USE_MOCK` is a hardcoded `const false` at `:59` designed for tree-shaking (not an env/config override) and the mock functions are unexported and referenced only internally, so there is zero production dataflow impact; this is a dev-experience/parity note, not a production defect. Fix: honor `mode`/`limit`/`signal` in the mock, or scope the parity comment.

### P3 — McpBotPublishModal state/timer lifecycle gaps + modal not closed on space switch (`packages/dmworkmcp/src/components/McpBotPublishModal.tsx:52/:86/:64`, `packages/dmworkmcp/src/pages/McpMarketListPage.tsx:187`) [source: octo-review-qwen3.7 + octo-review-glm5.2]

Four related lifecycle gaps in the bot-publish flow:
1. The prompt `useMemo` declares `[spaceId]` only but the factory reads `WKApp.apiClient.config.apiURL` and `window.location.origin` to compute `apiBaseUrl` — if the API URL changes at runtime without a space switch, the cached prompt carries a stale `apiBaseUrl` (`:52`). Low practical impact (both effectively static), technically an exhaustive-deps violation.
2. `handleCopy`'s catch returns without resetting `copied`; a prior successful copy leaves the button showing "copied" for up to 2s after a failed attempt (`:86`).
3. `copiedTimerRef` is cleared on unmount and before setting a new timer, but not when the modal transitions visible→hidden, so the 2s `setCopied(false)` fires on a hidden-but-mounted component (`:64`). No user-visible impact.
4. `handleSpaceChanged_` resets `publishMenuOpen`/`tagFilterOpen`/`tagsSelected`/`tagSuggestions`/`tagQuery`/`categoriesSelected` but not `botPublishVisible` — the modal stays open across a space switch with stale "copied" feedback while its launching dropdown IS closed, an inconsistency (`:187`).
Fix: add `apiURL` to the memo deps; reset `copied` (and clear the timer) in the catch and on visible→false; add `botPublishVisible: false` to the `handleSpaceChanged_` setState.

### P3 — `spaceId` interpolated unquoted into shell examples in the publish prompt (`packages/dmworkmcp/src/utils/mcpBotPublishPrompt.ts:44`) [source: octo-review-qwen3.8]

One injection chain, both halves machine-confirmed at function level, low practical exploitability. Sink: `getMcpBotPublishPrompt` embeds `spaceId` unquoted into shell command examples (`--profile space-${spaceId} --space ${spaceId}`, `:44-47`) with no shell-quoting and no UUID-format assertion — probes show `$(whoami)`, `;`, backtick and `|` payloads passing through verbatim, while `apiBaseUrl` IS sanitized by `new URL().origin`. Source: `getCurrentSpaceId()` (`McpBotPublishModal.tsx:16`) falls back to `localStorage.getItem("currentSpaceId")`, writable under same-origin XSS; a rendered-production-modal probe with a poisoned localStorage value shows it reaching the prompt unsanitized. Full CWE-78 exploitation requires pre-existing same-origin XSS + the user deliberately copying the prompt + a receiving bot agent executing the embedded commands verbatim without its own validation — none introduced by this diff. `spaceId` is a platform-assigned non-sensitive UUID, `WKApp.shared` takes precedence when available, and the fallback mirrors the pre-existing `dmworkskillmarket` pattern. Not a blocker; to harden, assert UUID format (`/^[0-9a-f-]{36}$/i`) on `spaceId` or single-quote the interpolated values in the shell examples.

### Nit — Hardcoded hex colors in index.css (`packages/dmworkmcp/src/index.css:691`) [source: octo-review-qwen3.7]

Four bare hex values (`#fff`, `#ffffff`, `#1f2937`) in avatar text and Semi Design tooltip overrides. Repo policy requires `var(--wk-*)` tokens. These may be justified — white text on a hash-colored avatar has no design-token equivalent, and Semi tooltip internals may not respond to CSS variables. Flag for author to confirm.

- Multi-tag wire encoding is sound: repeated `?tag=` params (not comma-joined, never `tag[]=`), comma-containing tags preserved as single values, mutation-controlled through the production `fetchMcpList` + `paramsSerializer`; URL round-trip via `URLSearchParams` `getAll`/`append` is comma-safe and scrubs the removed `created_by_type` param.
- Security posture is clean: `URLSearchParams`/axios `params` serialization everywhere (no injection), all React rendering uses JSX auto-escaping (removed `Highlight <mark>` reduces XSS-adjacent surface), `resolveMcpAPIBaseURL` normalizes to origin via `new URL(...).origin` (no SSRF), `mcpAxios` interceptor injects token/`X-Space-Id`/`Accept-Language` with 401→logout (no auth bypass). Bot-publish prompt is security-correct: token via stdin pipe with explicit `'不得输出 Token'` guard, no token in command args. Tag validation is deterministic (charset regex, `Array.from` code-point length, count cap — no ReDoS surface).
- `mode=mine` propagation verified to `/mcps/mine`, `/mcp_categories?mode=mine`, and `/mcp_tags?q=…&mode=mine&limit=100`; default omits `mode`/`q`.
- `/mcp_tags` unwrap null-safety confirmed against the authoritative backend (octo-marketplace PR #23 head `dff6a03`): flat `{"data":[…]}` envelope, `{data:null}`→`[]`, `{}`→`[]`.
- `handleSpaceChanged_`/`handleMode` fully reset filter state and cancel in-flight fetches — no cross-space filter leakage; the 160ms debounce + AbortController cancellation correctly drops stale tag-fetch responses.
- i18n parity: 233==233 en-US/zh-CN keys, zero only-in-one-locale, zero placeholder mismatches, 31 new keys in both locales.
- Existing suite: 4 files / 56 tests all PASS at head `cc9c2735`; `tsc --noEmit` shows only the env-noise `@types/react` cascade, zero genuine semantic errors vs baseline; credential scan clean (template placeholders are not real secrets); STRIDE / OWASP Top 10 cleared; `lucide-react ^0.577.0` CVE-clean.
- Informational (not a defect): category-pill counts under an active tag filter can overstate the filtered list — the backend `/mcp_categories` deliberately ignores `tag` by design (octo-marketplace PR #23 documents this); the only follow-up is an optional product/UX/doc clarification, not an octo-web code change.

## Verification Results

| Code Review | COMMENT | 3 P2 + 5 P3 + 1 nit (deduped across 3 reviewers), no P1 blockers — dominant theme is no test coverage + CI never runs vitest |
| E2E Test | PASS | All three verified client-side wire/URL/lifecycle contracts at head (8–10 probes + 56/56 shipped suite); cross-service server semantics (octo-marketplace #23) externally BLOCKED, not a defect |
| Security | CLEARED_WITH_RISK | No exploitable defect; one informational CWE-78 chain (unquoted spaceId in publish-prompt shell examples) requiring pre-existing same-origin XSS + external preconditions; credential scan clean |

Jerry-Xin (approved):
Re-review at head `2b188ad` (one new commit on top of the previously-reviewed `cc9c2735`; merge-base `cc9c2735`, ahead 1 / behind 0 — not a rebase). This commit is a clean follow-up that resolves the prior non-blocking notes and hardens the bot-publish prompt. Scoped to the true delta; the earlier-cleared surface was byte-verified unchanged. Approving.

### 🔴 Blocking
None.

### 💬 Non-blocking
- Prior notes are addressed: `fetchMcpTagsReal` in `packages/dmworkmcp/src/api/mcpService.ts` now routes through the shared `get<T>()` helper so 4xx/5xx/network failures are classified into `McpListError` (was a silent fall-through); `packages/dmworkmcp/src/pages/McpMarketListPage.tsx` surfaces the classified failure via `Toast` with a stale-controller guard, and `fetchMcpTagsMock` now honors `mode`/`limit`/`signal` for dev-harness parity.
- `loadMore` stale-response guard in `packages/dmworkmcp/src/pages/McpMarketListPage.tsx` switched from `join(",")` string comparison to array-reference identity — this removes a real collision between comma-bearing tags (`["v1.0,beta"]` vs `["v1.0","beta"]`). Filter AND-semantics and the space/mode/version guards are preserved.
- The tag suggestion path is still a fixed literal (`/mcp_tags`) with all values passed through the params object; nothing is interpolated into the request path, and the exported `serializeMcpParams` keeps the repeated-key (`?a=1&a=2`) wire form. No `dangerouslySetInnerHTML`/`innerHTML` introduced anywhere in the delta.

### ✅ Highlights
- `packages/dmworkmcp/src/utils/mcpBotPublishPrompt.ts` adds a UUIDv4 shape gate (`isValidMcpSpaceId` / `sanitizeSpaceId`) before embedding a space id into the `--space ${spaceId}` shell example, falling back to the `<space-id>` placeholder for malformed input (e.g. a poisoned localStorage fallback). `McpBotPublishModal.tsx` gates the copy button on the same check — closes a shell-token injection vector.
- `McpBotPublishModal.tsx` lifecycle cleanup is solid: memo deps include `apiURL`, and the copy-failure/hidden/unmount paths all reset `copied` and clear the pending timer.
- 65 new tests (121 total in the package) pin the param wire contract, tag round-trip fidelity (comma/space/CJK/astral), the UUID gate against injection payloads, and avatar/color stability. CI green.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1043 (octo-web)

## Summary

This PR aligns the MCP marketplace UI with the Skill marketplace pattern: the old single-select `createdByType` filter is replaced by a multi-tag chip filter with AND semantics, the `SourceBadge` component is retired in favor of a `resolveOwner` helper, and a new `McpBotPublishModal` provides a clipboard-copy publish prompt. Supporting work adds tag validation utilities (`mcpTagValidation.ts`), deterministic avatar generation (`mcpAvatar.ts`), URL query-state serialization for multi-tag (`mcpListQuery.ts`), and an infinite-scroll list page rewrite.

The PR is well-structured with thoughtful edge-case handling (UUID sanitization in publish prompts, abort-controller debounced tag fetch, clock-skew tolerant time formatting). One P2 item found around category pill counts not reflecting the active tag filter.

## Verification

Static analysis only at head `2b188ad2`; build and tests not executed in this environment.

- ✅ **Diff scope** — 21 files changed, +2277/-440, 6 commits. Head SHA `2b188ad2`.
- ✅ **SourceBadge removal** — searched the entire `dmworkmcp` package for `SourceBadge`; zero remaining references. Clean removal.
- ✅ **Highlight removal** — the `Highlight` component and `keyword` prop are removed from production `McpCard.tsx` and `McpDetailModal.tsx`.
- ✅ **Bot publish prompt security** — `isValidMcpSpaceId` gates on UUID regex; `sanitizeSpaceId` substitutes `<space-id>` placeholder for any non-UUID input. Test file `mcpBotPublishPrompt.test.ts` covers injection payloads (`$(whoami)`, backtick-wrapped `whoami`, `; rm -rf /`, etc.) and asserts none reach the shell examples. Comprehensive.
- ✅ **Tag validation** — `MCP_TAG_PATTERN` uses Unicode property escapes (`\p{L}\p{N}`) with explicit allowlist of punctuation. `MAX_MCP_TAGS=10` and `MAX_MCP_TAG_LENGTH=24` enforced in `McpCreateModal.commitOne`.
- ✅ **AbortController cleanup** — `scheduleTagFetch_()` aborts prior request before starting new one; `componentWillUnmount` aborts in-flight fetch. No leaked listeners — `componentDidUpdate` attaches global `pointerdown`/`keydown` only when popovers are open and removes them on close.
- ✅ **URL query state** — `mcpListQuery.ts` correctly parses `q.getAll("tag")` into `tagsSelected: string[]` and scrubs legacy `created_by_type` on serialize for bookmark compatibility.

## Findings

### P2 — Category pill counts do not reflect the tag filter (`packages/dmworkmcp/src/api/mcpService.ts:149`)

`categoryParams` passes `mode` and `created_by_type` to `/mcp_categories` but never `tag`, so when the user selects tags the pill numbers stay at unfiltered totals. This is consistent with the existing keyword-search behavior (search also does not filter category counts), but differs from the old `createdByType` single-select filter which _was_ forwarded. Consider forwarding `tag` to `categoryParams` for a consistent filter experience, or document the intentional divergence.

- `resolveOwner(item)` in `McpCard.tsx` correctly handles all three `createdByType` values: `"bot"` returns both bot and human names, `"import"` returns `null` (no owner shown), `"human"` returns human name only.
- `McpDetailModal` reuses the same `resolveOwner` import — consistent owner rendering across card and detail views.
- `getMcpAvatarText(name)` handles CJK (2 wide chars), digits (2 digits), Latin initials, and fallback to first-char uppercase. Hash-based color from `id` is deterministic.
- `formatUpdatedTime(iso)` in `McpDetailModal` handles clock-skew by clamping future timestamps to "just now".
- `handleSpaceChanged_` resets all tag/category/UI state on space switch — no stale filter bleed across spaces.
- `memoizedTagRows_()` uses identity-keyed memo on selected+suggestions union — avoids unnecessary re-renders.
- `McpCreateModal` tag input supports comma-separated batch paste and per-tag validation before commit.

- **C1 (Dual-path parity)** — CLEAR. Tag add/remove in filter popover uses the same `tagsSelected` state. `mcpListQuery.ts` serialize/parse are symmetric for `tag` params.
- **C2 (Control-flow ordering)** — CLEAR. `scheduleTagFetch_` debounce + abort is linear; no nested fetch chains. `componentDidUpdate` listener attach/detach is gated on discrete open/close transitions.
- **C3 (Authorization boundary)** — N/A. No auth/permission changes in this PR.
- **C4 (Authorization lifecycle)** — N/A. No auth changes.
- **C5 (Build artifacts)** — N/A. No build config changes; frontend-only UI work.
- **C6 (Governance)** — N/A. No governance/policy changes.

## Cross-round Blocker Check (R6)

## Verdict: APPROVED

No correctness, security, or build-blocking issues. The single P2 (category pill counts vs tag filter) is a UX consistency item, not a functional defect — the pill counts behave consistently with keyword search, and the divergence from the old source filter is minor. The PR is ready to merge; the P2 can be addressed in a follow-up.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Implementation is thorough with strong security hygiene (UUID-gated prompt sanitization, injection-proof tests), clean component removal, and well-scoped new utilities. One P2 observation is a non-blocking UX consistency note.

Jerry-Xin (approved):
MCP marketplace UI alignment + tag chip filter looks solid; behavior, query serialization, and edge-case handling are well covered. Approving.

### 🔴 Blocking
- None.

### 💬 Non-blocking
- 🔵 Design tokens — `packages/dmworkmcp/src/index.css` still ships one net-new raw value (`color: #fff` on the avatar-default text) and `packages/dmworkmcp/src/utils/mcpAvatar.ts` hard-codes a 6-swatch hash palette. The palette mirrors the already-merged `packages/dmworkskillmarket/src/utils/skillAvatar.ts`, so this is consistent with existing precedent, but `DEVELOPMENT.md` §2 asks for `var(--wk-*)` tokens. Consider promoting the shared avatar palette + the white avatar-text color to semantic tokens in a follow-up so both marketplaces share one source of truth. (The `#ffffff` / `#1f2937` / rgba shadows around the dropdown/popover block predate this PR's merge-base and are out of scope here.)
- 🔵 `/mcp_tags` failure path — on a real (non-abort) fetch failure, `packages/dmworkmcp/src/pages/McpMarketListPage.tsx` keeps the previously loaded suggestions and surfaces a `Toast.error`, intentionally matching the `loadMore()` toast pattern. That is a reasonable choice; just note it diverges from the "silent empty fallback" wording in the PR description — worth reconciling the copy so the documented behavior matches the code.
- 🔵 MatchReasons chips reveal `tool` / `usage_example` / `creator` (fields the card body doesn't already show), not only `creator`. The inline comment documents this deliberately and the behavior is sensible; the PR description's "creator hits only" phrasing is looser than the actual (better) rule — align the prose if desired.

### ✅ Highlights
- Repeated `?tag=X` query-param serialization + comma-safe URL round-trips are well tested (`mcpService.paramsSerializer.test.ts`, `mcpListQuery.test.ts`).
- Tag fetch uses debounce + `AbortController` cancellation with a stale-response guard; tag validation, clipboard-failure handling, and the import/bot/human owner-resolution branches are all handled thoughtfully.
- Both `en-US` and `zh-CN` locale resources updated consistently; solid unit coverage for avatar, tag validation, and the bot-publish prompt.

lml2468 (approved):
## Review: #1043 @ `8e3a09ff` — fix(dmworkmcp): MCP 市场 UI 对齐 Skill + 标签 chip 过滤器(#1009)

投票请求(我此前未在本 PR 落票)。锚定 head `8e3a09ff` / merge-base `a8bad922`(behind 2,0 冲突,`reviewDecision=APPROVED` 已有 2 票)。此 head 相较早前评审版新增了 5 个测试文件(paramsSerializer/mcpListQuery/mcpAvatar/mcpBotPublishPrompt/mcpTagValidation),回应了此前的测试覆盖 🟡。

**结论:可合并 ✅**

### 安全面(逐项字节核实)
- **标签 chip 过滤器无注入**:`serializeMcpParams` 用 `URLSearchParams.append` 对数组**逐项**追加 → `?tag=a&tag=b`(重复键,非逗号拼接,含逗号标签安全);标签值只进 axios `params` 对象,请求路径固定字面 `/mcp_tags`,**用户值从不入路径**。✅
- **Bot 上架 spaceId 注入闸**:`isValidMcpSpaceId` UUIDv4 正则校验 + `sanitizeSpaceId` 非法回退 `<space-id>` 占位 → 堵住污染 localStorage 把 `$(whoami)`/`;`/反引号注入 `--space ${spaceId}`;复制按钮 `disabled={!promptReady}`(含 `isValidMcpSpaceId(spaceId)`)、剪贴板不可用 `Toast.error`。✅
- **owner 归属**:用**显式 `item.createdByType === "bot"`/`"import"`**(非 #1007 那种「任何 separate creator 都当 bot」的过宽启发式)。✅
- **XSS**:card/detail/botpublish 三组件 **0 `dangerouslySetInnerHTML`**;颜色经校验后进 CSS var。✅

### CSS token(#1008 教训 —— 已亲自核实,非误报)
- index.css 的 `var(--wk-*)` 全部对照 `theme/semantic.css` 确认有定义(`--wk-sp-1`/`--wk-bg-surface`/`--wk-text-primary` 等基础 token 均在;dmworkmcp 本就大量复用)。**无 #1008 那类未定义 token**。(过程中我的初版 grep 因 `--` 前缀被当 flag 产生假阳性,已用绝对路径 fixed-string + 生产构建通过双重排除。)

### 构建 / 测试(实跑)
- `pnpm --filter @octo/web build` ✅ built in 3.51s。
- `pnpm --filter @dmwork/mcp test` ✅ **9 files / 121 tests 全绿**(含新增 serializer/listQuery/avatar/prompt/tagValidation 五套 —— 覆盖了标签过滤序列化、spaceId 校验、头像 fallback 等此前 🟡 的空白)。

---

### 🟡 非阻断(承前)
- `mcpBotPublishPrompt` 硬编码 `marketplace mcp create` 等 CLI 命令(CLI 改名会漂移,集中单文件)、`fetchMcpTagsMock` 已遵守 mode/limit、类别标签计数不随标签过滤更新(设计差异)。
- **backend 提示(承 Steve)**:workspace 级标签/资源 DELETE 的真正鉴权依赖 `mcpService`/后端(前端 UI 可见即可点),建议后端确认删除接口有 workspace admin 校验。非前端阻断。

标签过滤无注入、spaceId 注入闸到位、owner 用显式 provenance、XSS 干净、CSS token 已核实有定义、五套新测试补齐覆盖,构建与 121 测试通过。**APPROVE。**

PR #45, opened 2026-05-19
fix(contextmenu): fix submenu hover gap — move bridge pseudo-element to parent li

a652:
## Summary

Fix the submenu disappearing issue when moving the mouse from a parent menu item toward its submenu. The submenu was nearly impossible to click due to a 4px gap where hover was lost.

## Changes

- Remove broken `::before` bridge from `.wk-ctx-submenu` (was on the wrong element — cannot cover a gap outside its own boundary)
- Add correct `::after` bridge on the parent `li:has(> .wk-ctx-submenu)`, extending its hover area 8px rightward to bridge the 4px gap
- Mirror fix for `.wk-contextmenus-flip-submenu` (left-opening submenu) case

## Testing

- [ ] Unit tests added/updated
- [x] Manually verified

Reproduced the bug on latest `main`: hovering over 「更多」 in the chat list right-click menu and moving the mouse rightward caused the submenu to disappear immediately. After fix, the submenu stays open when crossing the gap.

**Root cause explanation:**

```
Parent <li> │ 4px gap │ .wk-ctx-submenu
             ↑
   Mouse crosses here → old ::before (on submenu) can't reach this area
   New ::after (on parent li) extends rightward and covers it ✓
```

## Checklist

- [x] I have read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [x] PR description is in English
- [ ] Added tests for my changes
- [x] Updated documentation
- [x] Followed commit message conventions (Conventional Commits)

PR #1044, opened 2026-07-23
feat: comprehensive Summary module UI redesign driven by Figma

quanming1:
## Overview

Comprehensive UI redesign of the smart summary module based on Figma design specs.

## Key Changes

- **Create Page**: rocket emoji + 24px title, 420px input, 4-col template grid, chat/participant chips with WKAvatar, pill-shaped start button
- **List Cards**: source info + status + title + desc + timestamp + status-specific dropdown menu, custom PopConfirm dialog
- **Detail Page**: title in content flow, content card bg, meta info with source chips, Markdown typography, purple citation badges, generating/failed states inside card, edit mode with footer buttons, simplified regenerate modal
- **Chat Selector**: dual-column overlay, members mode with role tabs (all/managers/normal)
- **Citation Panel**: purple theme, card messages, WKAvatar real avatars
- **List Page**: pill search bar, dropdown status filter, infinite scroll, fixed header/toolbar
- **Panel Mode**: responsive 2-col templates, stacked action bar
- **i18n**: 20+ new keys

## Conflict Resolution

Kept Figma redesign, took member_count from main, commented out missing @mlt-org CSS imports


Jerry-Xin (dismissed):
This Figma-driven Summary UI refactor is on-scope, but it does not compile and bundles several user-facing regressions plus an unrelated shared-renderer change. Requesting changes.

## 🔴 Blocking

- 🔴 **Citation rendering does not type-check.** `CitationText` still passes `displayIndex` / `displayIndices`, but those props were removed from the badge components. See `packages/dmworksummary/src/components/CitationText.tsx:221` and `:232` vs `packages/dmworksummary/src/components/CitationBadge.tsx`. Restoring the display-index props is also needed so reading-order numbering is not replaced by raw citation ids (e.g. `[37]`).

- 🔴 **`CitationGroupBadge` references an undefined `t`.** The component at `packages/dmworksummary/src/components/CitationBadge.tsx:277` calls `t("summary.citation.sourceDefault")` at `:309` but never calls `useI18n()` in its own scope (the `useI18n()` calls at :195/:226/:377 belong to sibling components). This is an undefined-variable / compile error.

- 🔴 **CI Build fails on i18n check.** The build job fails with: `summary.citation.sourceDefault` used in `CitationBadge.tsx` but missing from both `en-US.json` and `zh-CN.json`; `summary.create.desc` used in `ChatSummaryNewModal.tsx` but missing from both locales; and a hardcoded Chinese string `请输入` in `packages/dmworksummary/src/pages/SummaryDetailPage.tsx`. Add the missing keys and route the hardcoded string through i18n.

- 🔴 **"Cancel task" invokes the leave-collaboration API instead of cancel.** The creator menu action resolves to `leaveSummary` even though a `cancelSummary` path exists. This can fail for creators or leave a generating task running. See `packages/dmworksummary/src/components/SummaryCard.tsx:157` and `packages/dmworksummary/src/pages/SummaryListPage.tsx:404`.

- 🔴 **Personal-report / personal-draft editors lost their save & cancel controls.** `SummaryEditor` removed its action buttons (`packages/dmworksummary/src/components/SummaryEditor.tsx:106`), while the new external footer renders only for `isEditing` / `editingTeamSummary` and personal editor instances do not provide `exposeSave`. These editing flows become impossible to complete. See `packages/dmworksummary/src/pages/SummaryDetailPage.tsx:2900` and `:3699`.

- 🔴 **Forge interactive-card render profile is effectively disabled (unrelated change).** In `packages/dmworkbase/src/Messages/InteractiveCard/sdk/renderOctoCard.ts:52`, the `FORGE_RENDER_PROFILE` branch now returns `buildOctoHostConfig(...)` — identical to the legacy branch — and the `forgeHostConfig` import was removed (`:2`). This will visually break `octo-chat/v1` cards (Forge font sizing / action alignment). This shared-renderer change is unrelated to the Summary UI refactor and should not be bundled here; if intentional, split it out and update the related assertions.

## 💬 Non-blocking

- 🟡 **Member filtering uses `m.is_bot` instead of the established `isBot(uid)` contract.** Subscriber objects may not populate `is_bot`, which can let bots into collaboration selections. See `packages/dmworksummary/src/components/ChatSelectorModal.tsx:97`.

- 🟡 **Extensive hard-coded colors / spacing / radii and `!important` rules.** This conflicts with the design-token guidance in `DEVELOPMENT.md` and is likely to break dark mode. Representative spots: `packages/dmworksummary/src/index.css:321` and `packages/dmworksummary/src/components/CitationBadge.tsx:38`. Prefer semantic CSS variables over literals.

- 🟡 **Add regression tests** for task cancellation, each editor mode (team / personal / draft), citation renumbering, and bot exclusion — the existing tests do not cover the blocking paths above.

## ✅ Highlights

- Good news on the security side: no `dangerouslySetInnerHTML`, no `rehype-raw`, and no changes to the summary markdown / sanitization pipeline were introduced — the citation markdown render path is untouched.
- No new dependencies were added (no `package.json` changes), so no supply-chain surface here.
- No new external navigation / raw URL path interpolation was introduced.

Note: full type-check / i18n validation could not be run locally because the checkout has no installed dependencies; the CI Build job already reproduces the i18n failures above.

Jerry-Xin (dismissed):
Re-review at head 8cf67e356e18 (single fix commit "address all CR blocking issues", ahead 1 / behind 0 of prior head 024dee58). All six previously-flagged blockers are resolved and CI Build is green, but this fix commit leaves two real editing/citation regressions that still block merge.

## 🔴 Blocking

- 🔴 Personal-report editing has no Save/Cancel controls. `SummaryEditor` no longer renders its own action buttons (save/cancel now come from the detail-page footer via the `exposeSave` callback), and the `editingMyDraft` branch was correctly wired. But the sibling `editingPersonalReport` branch renders `SummaryEditor` without `exposeSave`, and the footer render condition is `isEditing || editingTeamSummary || editingMyDraft` — it does not include `editingPersonalReport`. So a user who enters personal-report edit mode gets neither in-editor buttons nor a footer: they cannot save or cancel. Fix: add `exposeSave={(fn) => { this.editorSaveFn = fn; }}` to the `SummaryEditor` at `packages/dmworksummary/src/pages/SummaryDetailPage.tsx:2900`, and add `editingPersonalReport` to the footer condition (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3700`) plus its save/cancel dispatch (`handleEditPersonalReportSave` / `handleEditPersonalReportCancel`). Ref: `packages/dmworksummary/src/components/SummaryEditor.tsx:106`.

- 🔴 Grouped citation labels regress non-contiguous groups. On base, `CitationGroupBadge` used `formatGroupLabel(...)` (from `packages/dmworksummary/src/components/citationFormat.ts:19`, which handles reordered/non-contiguous index sets). This PR replaced it with a naive range `${first}-${last}`, so a group like [1][3][5] now renders as [1-5], implying references 2 and 4 that are not part of the group. `formatGroupLabel` is still present in the file; restore it: `const label = formatGroupLabel(displayIndices ?? indices);` at `packages/dmworksummary/src/components/CitationBadge.tsx:279`.

## ✅ Resolved from prior review (byte-verified at 8cf67e356e18)

- Citation type mismatch fixed — `CitationBadge` re-accepts `displayIndex` / `displayIndices` (added to both prop interfaces) and uses them for numbering (`[{displayIndex ?? index}]`); `CitationText.tsx:221`/`:232` pass them, so it type-checks and reading-order numbering is preserved.
- Undefined `t` fixed — `useI18n` is imported and `const { t } = useI18n()` is now called inside `CitationGroupBadge` scope (`packages/dmworksummary/src/components/CitationBadge.tsx`).
- i18n build fixed — keys `citation.sourceDefault`, `create.desc`, `create.placeholder` added to BOTH en-US and zh-CN; hardcoded Chinese placeholder in `SummaryDetailPage.tsx` replaced with `t("summary.create.placeholder")`. CI Build and install-build now pass.
- Cancel wired to correct API — `SummaryCard` "cancel task" menu now calls `onCancel` → `SummaryListPage.handleCancel` → `api.cancelSummary` (POST /summaries/{id}/cancel), no longer `leaveSummary`. `list.cancelSuccess` key added to both locales.
- Personal draft editor controls restored — `editingMyDraft` branch renders `SummaryEditor` with `exposeSave` and is included in the footer condition + save/cancel dispatch.
- Forge card render profile — FORGE branch no longer uses the legacy `buildOctoHostConfig` revert; it now uses `new HostConfig(forgeHostConfig)` with `forgeHostConfig` imported from the octo-chat card profile, and the profile theme/styles CSS imports in `InteractiveCardCell.tsx` are re-enabled.
- Member filter — `ChatSelectorModal.tsx:97` now applies `!m.is_bot && !isBot(m.uid)` per the isBot contract.
- markdown/XSS pipeline still clean — no `dangerouslySetInnerHTML`, `rehype-raw`, or `innerHTML` introduced in any of the 9 changed files; safe citation render path untouched.

## 💬 Non-blocking

- 🟡 Citation badges lost keyboard focus and Enter/Space/Escape handling — citations are now mouse-only (`packages/dmworksummary/src/components/CitationBadge.tsx:274`). Consider restoring keyboard interaction for accessibility.
- 🟡 Extensive hard-coded colors/spacing/radii, multiple `!important`, and direct Semi class overrides continue to violate the token convention and risk unreliable dark-mode behavior (`packages/dmworksummary/src/index.css`).
- 🟡 Member sync failures in `ChatSelectorModal.tsx:89` are silently converted to an empty list, indistinguishable from a group with no members; surface a retryable error.

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #1044 (octo-web)

Reviewed at head `8cf67e35`. Scope: comprehensive Summary module UI redesign — panel mode, infinite scroll, citation badge refactor, chat selector two-column layout, create page embedded mode, and i18n additions across 10 files (~1700 LOC changed).

## Summary

Large, well-structured UI redesign that moves the Summary module from full-page-only to a dual panel/full-page pattern. The refactor correctly separates concerns: `ChatSummaryPanel` orchestrates list/detail/create views, `SummaryListPage` gains infinite scroll, `CitationBadge` drops the hover-preview in favor of click-only popover, and `ChatSelectorModal` adds a members mode for participant selection. i18n keys are added for both locales. One build-blocking defect must be fixed before merge; a few smaller items noted below.

## Verification

Static analysis only at head `8cf67e35`; build and tests not executed in this environment.

- ✅ **Infinite scroll contract** — `SummaryListPage` `loadMore` guard (`loading || noMore`) prevents duplicate fetches; `pageRef` tracks next page correctly; reset on `componentDidUpdate` when `channelId` changes.
- ✅ **Panel-mode badge scoping** — badge emission in `SummaryListPage.updateUnreadBadge()` is gated on `!this.props.channelId`, so panel instances don't pollute the global badge count.
- ✅ **Event listener cleanup** — `SummaryDetailPage` adds `summary-detail-regenerate` and `summary-detail-edit` listeners in `componentDidMount` and removes them in `componentWillUnmount`.
- ✅ **ChatSelectorModal members data flow** — `loadMembers()` calls `WKSDK.shared().channelManager.syncSubscribes(channel)`, filters bots via `isBot()`, stores roles in `memberRoles` Map, and maps subscribers to `ChatCandidate` objects. Role-based tab filtering (`all_members` / `managers` / `normal_members`) correctly reads from `memberRoles`.
- ✅ **ResizeObserver lifecycle** — `SummaryCreatePage` creates observer in `componentDidMount`, disconnects in `componentWillUnmount`. Chip overflow calculation uses `containerRef` width vs measured chip widths with `Math.floor` safety margin.
- ✅ **i18n parity** — all new keys present in both `en-US.json` and `zh-CN.json`.

## Findings

One P1 blocker (build-breaking); two P2 items; two nits.

### P1 — `displayIndex` referenced but not destructured in `CitationBadge` (`packages/dmworksummary/src/components/CitationBadge.tsx:274`)

The `CitationBadge` component's function signature at line 227 destructures only `{ index, citations, badgeKey }` from props:

```tsx
const CitationBadge: React.FC<CitationBadgeProps> = ({ index, citations, badgeKey }) => {
```

But the JSX at line 274 references `displayIndex`, which is not in scope:

```tsx
<sup className="citation-badge" style={badgeStyle} onClick={() => onBadgeClick(badgeKey)}>[{displayIndex ?? index}]</sup>
```

The `displayIndex` prop is declared in the `CitationBadgeProps` interface (line 14) but was never added to the destructuring pattern. TypeScript will emit `TS2304: Cannot find name 'displayIndex'` — this breaks the build. Fix: add `displayIndex` to the destructuring:

```tsx
const CitationBadge: React.FC<CitationBadgeProps> = ({ index, displayIndex, citations, badgeKey }) => {
```

### P2 — Duplicate state key `forwardingToMatter` (`packages/dmworksummary/src/pages/SummaryDetailPage.tsx:194`)

The state initialization object contains `forwardingToMatter: false` on two consecutive lines:

```ts
forwardingToMatter: false,
forwardingToMatter: false,
```

JavaScript silently takes the last value (still `false`), so this is not a runtime bug, but it signals a copy-paste error and will trip any `no-dupe-keys` lint rule. Remove one.

### P2 — Dead code: `isScheduledTask` computed but never read (`packages/dmworksummary/src/components/SummaryCard.tsx:90`)

```ts
const isScheduledTask = (task.schedule_id != null && task.schedule_id > 0) || task.trigger_type === TriggerType.SCHEDULED;
```

This variable was consumed by the old `Popconfirm` content that conditionally showed "scheduled tasks cannot be deleted" copy. The refactor replaced that UI with a `Dropdown.Menu` + custom `Modal.confirm` that uses `confirmDeleteDesc` unconditionally. The variable is now dead code — remove it and the `TriggerType` import if it has no other consumer in this file.

### Nit — Double semicolon (`packages/dmworksummary/src/pages/SummaryListPage.tsx:456`)

```ts
};;
```

Harmless but should be cleaned up.

### Nit — 300ms setTimeout coupling for edit event dispatch (`packages/dmworksummary/src/pages/SummaryListPage.tsx:451-456`)

`handleEdit` calls `handleCardClick` (which triggers a detail panel navigation) then `setTimeout(300ms)` to dispatch `summary-detail-edit`. If the detail panel takes longer than 300ms to mount and attach its event listener, the event is silently lost. Consider either (a) dispatching the event from the detail panel's `componentDidMount` via a queued flag, or (b) increasing the timeout with a comment explaining the coupling. Not blocking — the current approach likely works in practice.

- **Forward-to-chat chunking** — `SummaryDetailPage.forwardToChat` splits content into 4000-char chunks and sends sequentially with `await`. AbortController cancels on unmount. Each chunk's error is caught and surfaced via Toast without silently swallowing.
- **CitationGroupBadge merge logic** — `mergeGroupMessages` deduplicates by `message_seq` (or sender+content+sent_at fallback), promotes cited over context, and sorts by seq then timestamp. Correct for overlapping context windows.
- **ChatSelectorModal bot filtering** — uses both `m.is_bot` field and `isBot(m.uid)` helper, covering server-flagged and UID-pattern bots.
- **Module initialization simplification** — `module.tsx` correctly removes unused React hooks imports and dead sidebar registration code.

## Data Flow Trace

| Consumer | Data | Source | Verified |
|---|---|---|---|
| `SummaryListPage.render()` | `tasks` array | `loadTasks()` → `api.listTasks()` → `setState` | ✅ |
| `SummaryListPage.loadMore()` | `pageRef.current` | Reset in `componentDidUpdate` on channelId change; incremented after each fetch | ✅ |
| `SummaryDetailPage.handleRetry()` | `task` object | `api.getTask(taskId)` → fresh fetch, not stale state | ✅ |
| `ChatSelectorModal` members tab | `candidates` | `WKSDK.channelManager.syncSubscribes()` → filter bots → map to `ChatCandidate` | ✅ |
| `CitationBadge` popover | `citation` | `citations.find(c => c.index === index)` from parent-passed array | ✅ |
| `SummaryCreatePage` chip overflow | `visibleChipCount` | ResizeObserver callback measures container vs chip widths | ✅ |

## Blindspot Checklist

- **C1 (dual-path parity)** — N/A. No symmetric add/remove or create/delete paths changed.
- **C2 (control-flow ordering)** — Clear. `handleEdit` ordering concern noted as nit.
- **C3 (authorization boundary)** — N/A. No auth/permission changes.
- **C5 (build ≠ runtime)** — The P1 `displayIndex` issue is exactly this category: the code looks correct at the JSX level but the variable is not in scope, which a build would catch. Flagged as P1.
- **C6 (governance/docs)** — N/A. No governance or policy changes.

[Octo-Q] verdict: REQUEST_CHANGES — One P1 build-blocking defect (`displayIndex` not destructured in `CitationBadge`). Fix is a one-line change. P2 items are non-blocking but recommended.

Jerry-Xin (dismissed):
The PR is relevant to Octo Web, but two list-state regressions must be fixed before merging.

🔴 Blocking

- 🔴 Critical — Switching chats leaves the summary panel showing the previous chat’s summaries. `SummaryListPage` loads only on mount and never reacts when `channelId` changes. `ChatSummaryPanel` resets its own view, but keeps the list component mounted, so no new request is triggered. Add `componentDidUpdate` handling and a regression test. [SummaryListPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-fh29awv2/repo/packages/dmworksummary/src/pages/SummaryListPage.tsx:145)

- 🔴 Critical — Deleting after infinite scrolling can discard earlier pages or incorrectly open the create screen. `handleDelete_refetch()` calls `fetchData()`, which requests the current page, then replaces the complete accumulated list with that page. If the last loaded page becomes empty, it concludes that no summaries exist even when earlier pages remain. Refetch from page 1 or remove the item locally and reconcile pagination. [SummaryListPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-fh29awv2/repo/packages/dmworksummary/src/pages/SummaryListPage.tsx:350) [SummaryListPage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-fh29awv2/repo/packages/dmworksummary/src/pages/SummaryListPage.tsx:366)

💬 Non-blocking

- 🟡 Warning — The member selector derives channel type from whether the ID contains `"____"`, treating every other source as a group. Selecting participants after choosing a direct chat therefore synchronizes subscribers for the wrong channel type. Use the candidate’s `chat_type` mapping. [SummaryCreatePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-fh29awv2/repo/packages/dmworksummary/src/pages/SummaryCreatePage.tsx:760)

- 🟡 Warning — `humansOnly` replaces the caller-provided `filter` instead of composing with it. Future callers using both flags will unexpectedly lose their original filtering constraints. [list.tsx](/Users/mlamp/.openclaw/tmp/pr-review-fh29awv2/repo/packages/dmworkbase/src/Components/Subscribers/list.tsx:259)

- 🟡 Warning — The redesign introduces extensive hard-coded colors, spacing, radii, `!important`, and direct Semi-class overrides, contrary to the project’s explicit styling rules and likely breaking dark mode. [index.css](/Users/mlamp/.openclaw/tmp/pr-review-fh29awv2/repo/packages/dmworksummary/src/index.css:321)

- 🔵 Suggestion — Add tests covering channel changes, deletion after loading multiple pages, and member selection for group/thread/direct sources.

✅ Highlights

- Panel navigation now consistently reuses the list, create, and detail pages.
- The updated panel and star-button behavior has focused component coverage.
- English and Chinese locale resources were updated together.

Validation could not run because this checkout has no installed dependencies.

Jerry-Xin (dismissed):
🔄 Re-review at head 08ae5aca7ba9. The two blockers I raised last round are both fixed, and the peer build-fail is resolved — but this round surfaces two list-state regressions in the panel/list flow that must be fixed before merge.

## 🔴 Blocking

- 🔴 Switching chats leaves the summary panel showing the previous chat's summaries. `SummaryListPage` only fetches in `componentDidMount` (no `componentDidUpdate` reacting to `channelId`), and `ChatSummaryPanel` reuses a single mounted `SummaryListPage` without a `key`, so it is never remounted when `channel.channelID` changes — `componentDidUpdate` there only resets the panel's own `view`/`selectedTaskId`, not the list data. Add a `componentDidUpdate` in `SummaryListPage` that reloads when `channelId` changes (or key the child by `channelID` in the panel), plus a regression test. `packages/dmworksummary/src/pages/SummaryListPage.tsx:145`, `packages/dmworksummary/src/components/ChatSummaryPanel.tsx:174`

- 🔴 Deleting after infinite-scroll can drop earlier pages or misroute to the create screen. `handleDelete_refetch()` calls `fetchData()`, which requests the *current* `state.page` (e.g. page 3), then replaces the whole accumulated `items` list with just that single page's results. Pages 1..N-1 disappear; if the current page comes back empty it concludes there are no summaries and opens the create flow even when earlier pages still have items. The `page>1 && items.length<=1` guard only covers the accumulated-list edge case, not the general one. Refetch from page 1 (reset accumulation) or remove the item locally and reconcile pagination. `packages/dmworksummary/src/pages/SummaryListPage.tsx:350`, `packages/dmworksummary/src/pages/SummaryListPage.tsx:366`

## ✅ Resolved since last round

- Personal-report editor now has working save/cancel: the personal-report `SummaryEditor` receives `exposeSave` and the footer render condition includes `editingPersonalReport`, wiring cancel/save handlers. `packages/dmworksummary/src/pages/SummaryDetailPage.tsx:2906`, `:3700`
- Grouped citation label now uses `formatGroupLabel(displayIndices ?? indices)` instead of a naive `first-last` range, so non-contiguous sets like `[1][3][5]` render as `[1,3,5]` not `[1-5]`. `packages/dmworksummary/src/components/CitationBadge.tsx:284`
- Build-fail fixed: `CitationBadge` now destructures `displayIndex` (and `CitationGroupBadge` destructures `displayIndices`); CI Build is green. `packages/dmworksummary/src/components/CitationBadge.tsx:228`
- Nits cleaned: duplicate `forwardingToMatter` state removed, unused `isScheduledTask`/`TriggerType` import dropped in `SummaryCard`, double-semicolon fixed, setTimeout coupling documented.

## 💬 Non-blocking

- 🟡 The member selector infers channel type from whether the id contains `____`, treating everything else as a group; selecting participants after picking a direct chat then syncs subscribers for the wrong type. Prefer the candidate's `chat_type` mapping. `packages/dmworksummary/src/pages/SummaryCreatePage.tsx:760`
- 🟡 `humansOnly` overwrites the caller-provided `filter` instead of composing with it, so callers using both flags lose their original constraint. `packages/dmworkbase/src/Components/Subscribers/list.tsx:259`
- 🟡 The redesign adds extensive hard-coded colors/spacing/radii, `!important`, and direct Semi-class overrides against the project styling rules, which likely breaks dark mode. `packages/dmworksummary/src/index.css:321`
- 🔵 Add tests for channel switching, deletion after loading multiple pages, and member selection across group/thread/direct sources.

## ✅ Highlights

- Markdown/citation rendering pipeline stays XSS-clean (no `dangerouslySetInnerHTML` / `rehype-raw` in the touched files).
- `formatGroupLabel` is a clean, unit-testable pure function with clear contiguity rules.
- Panel navigation reuses list/create/detail pages consistently; locale resources updated in both languages.

Jerry-Xin (approved):
Re-review at head 9a08fe5f798f — both prior blocking list-state regressions are fixed. Approving.

## ✅ Blocking items from the previous round — now resolved

- **Stale summary after channel switch — FIXED.** `SummaryListPage` now implements `componentDidUpdate(prevProps)` that calls `loadData()` when `prevProps.channelId !== this.props.channelId`. `loadData()` resets `page: 1, hasMore: true` and refetches with `origin_channel_id: this.props.channelId`, so switching chats refetches the new channel's list instead of showing the previous channel's summaries. (`packages/dmworksummary/src/pages/SummaryListPage.tsx:157`, `:193`)

- **Page loss / wrong create-jump after delete — FIXED.** `handleDelete` no longer runs the fragile `page > 1 && items.length <= 1` branch that reset the page and misrouted to create on a page that was empty only due to deletion. It now unconditionally calls `loadData()`, reloading from page 1 with the real `total`, so earlier pages are not lost and no false-empty auto-jump to create occurs. The remaining empty-state view only renders a manual create Button (user click), never an automatic misroute. (`packages/dmworksummary/src/pages/SummaryListPage.tsx:356`)

## 💬 Non-blocking

- The previously-noted `____`-substring channel-type inference is improved to explicit `chat.chat_type` handling (`"thread" ? 5 : "direct" ? 1 : 2`). Good change. (`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:768`)
- `handleDelete_refetch` is now orphaned (no callers) after the delete-path rewrite. Consider removing it to avoid dead code. (`packages/dmworksummary/src/pages/SummaryListPage.tsx:367`)

## Scope note

The incremental change in this commit is limited to `SummaryListPage.tsx` and `SummaryCreatePage.tsx` (ahead 1 / behind 0 vs the prior reviewed head; no rebase). It cleanly addresses the two outstanding blockers without touching other surfaces, and the Build check is green. Approving on that basis.

mochashanyao (approved):
[Octo-Q · automated review]

---

## Review: octo-web PR #1044 — feat: comprehensive Summary module UI redesign
**Head SHA:** `9a08fe5f798f4f490bdc8212c27a7fcf682e41df`
**Files changed:** 24 (+3615 / -1640)

---

### 1. Verification Conclusions

✅ **Event system consistency** — `summary-detail-regenerate` and `summary-detail-edit` use matching `window.dispatchEvent`/`window.addEventListener` pairs across SummaryListPage ↔ SummaryDetailPage. (`SummaryListPage.tsx:448,457` → `SummaryDetailPage.tsx:259-260,331-332`)

✅ **mittBus event consistency** — `summary-list-refresh-requested` uses `WKApp.mittBus.on`/`emit`/`off` consistently across ChatSummaryPanel and SummaryListPage. Cleanup in `componentWillUnmount` confirmed. (`SummaryListPage.tsx:150,170`, `ChatSummaryPanel.tsx:143`)

✅ **ChatSelectorModal member loading** — `loadMembers()` properly calls `WKSDK.channelManager.syncSubscribes()` and filters by role. `getChannelTypeNum()` maps chatType strings to numeric types correctly. (`ChatSelectorModal.tsx:107-140`)

✅ **ChatSummaryPanel cleanup** — `componentWillUnmount` properly clears drag state and removes mittBus listener. (`ChatSummaryPanel.tsx:113-122`)

✅ **SummaryCreatePage chip overflow** — ResizeObserver-based chip overflow detection (`applyChipOverflow`) is correctly guarded with a `requestAnimationFrame` to avoid layout thrash. Cleanup in `componentWillUnmount` confirmed. (`SummaryCreatePage.tsx:166-192`)

✅ **i18n keys** — New keys (`summary.list.deleteConfirm`, `summary.list.leaveConfirm`, `summary.card.menu.*`, `summary.create.selectedChats`, `summary.create.selectedMembers`, `summary.chip.overflow`) are present in both `en-US.json` and `zh-CN.json`.

---

### 2. Findings

#### P2-1 — Duplicate CSS import (module.tsx:19-20)

```typescript
import "./index.css";  // line 19
import "./index.css";  // line 20 — duplicate
```

**Diff-scope:** New in this PR (merge artifact from index.css rework).
**Impact:** Webpack/Vite deduplicates at build time, so no runtime harm. But it's a code-quality issue that signals an unclean merge.

**Suggestion:** Remove the duplicate line.

---

#### P2-2 — Dead code: `handleDelete_refetch` (SummaryListPage.tsx:367-381)

```typescript
handleDelete_refetch = async (taskId: number) => {
    try {
        const fresh = await this.fetchData();  // reads this.state.page — stale after infinite scroll
        if (fresh.items.length === 0) {
            this.setState({ items: [], hasMore: false });
        } else {
            const next = fresh.items[0];       // first item of current page, not "next after deleted"
```

**Diff-scope:** New in this PR.
**Impact:** This method is defined but **never called** from any render path, callback, or event handler. Additionally, it contains a latent bug: `fetchData()` reads `this.state.page` which could be >1 after infinite scroll, and `items[0]` of a later page is not the logical "next item after deletion."

**Suggestion:** Either remove this dead method or fix it to reset `page` to 1 before fetching and wire it into `handleDelete`.

---

#### P2-3 — Fragile 300ms setTimeout race in handleRegenerate/handleEdit (SummaryListPage.tsx:446-458)

```typescript
handleRegenerate = (taskId: number) => {
    this.handleCardClick(taskId);          // triggers onViewDetail → parent mounts SummaryDetailPage
    setTimeout(() => {
        window.dispatchEvent(new CustomEvent("summary-detail-regenerate", { detail: { taskId } }));
    }, 300);
};

handleEdit = (taskId: number) => {
    this.handleCardClick(taskId);
    setTimeout(() => {
        window.dispatchEvent(new CustomEvent("summary-detail-edit", { detail: { taskId } }));
    }, 300);
};
```

**Diff-scope:** New in this PR.
**Impact:** The 300ms delay is meant to allow the detail page to mount and register its event listener. Under normal React rendering this works, but on slow devices or with large DOM trees, if `SummaryDetailPage.componentDidMount` hasn't fired within 300ms, the dispatched event is silently lost and the user's action has no effect.

**Data-flow trace:** `handleCardClick` → `this.props.onViewDetail(taskId)` → `ChatSummaryPanel.setState({ view: 'list', selectedTaskId })` → React re-render → `SummaryDetailPage` mount → `componentDidMount` → `addEventListener`. Each step is fast but the total is unbounded.

**Suggestion:** Replace the setTimeout with a ready signal from `SummaryDetailPage` (e.g., dispatch a `summary-detail-ready` event from `componentDidMount`, and have `SummaryListPage` queue the regenerate/edit action until it receives that signal). Alternatively, pass the action as a prop rather than relying on cross-component event timing.

---

#### P2-4 — handleDelete resets to page 1, losing infinite scroll position (SummaryListPage.tsx:420-427)

```typescript
handleDelete = async (taskId: number) => {
    try {
        await api.deleteSummary(taskId);
        Toast.success(t("summary.list.deleteSuccess"));
        this.loadData();  // resets page=1, fetches only page 1
    }
};
```

**Diff-scope:** Amplified by this PR. The list page was converted from full-page pagination to infinite scroll. `loadData()` always resets to `page: 1`, but now that the user can accumulate multiple pages via scrolling, deleting any item throws them back to the first page, losing their scroll position.

**Impact:** UX regression. After scrolling through 60+ items (3 pages), deleting item #45 shows only the first 20 items. The user must re-scroll to find their previous position.

**Suggestion:** After delete, fetch the same page range the user has already loaded (i.e., fetch `page 1..currentPage` with the existing `pageSize`) and preserve the items array minus the deleted item, rather than resetting to page 1 only.

---

#### P2-5 — SummaryCard `formatRelativeTime` is static — no re-render timer (SummaryCard.tsx:93-106)

```typescript
formatRelativeTime(dateStr: string): string {
    const date = new Date(dateStr);
    // ... returns "just now", "X minutes ago", etc.
    // But this is only computed at render time, never re-computed
}
```

**Diff-scope:** New in this PR (replaces server-rendered date display).
**Impact:** A card showing "just now" will continue to display "just now" indefinitely until something else triggers a re-render. After 5 minutes, it should show "5 minutes ago" but won't unless the component re-renders for another reason (e.g., status poll, parent re-render).

**Suggestion:** Add a lightweight interval timer (e.g., `setInterval` every 60s in `componentDidMount`) that calls `this.forceUpdate()` or updates a `now` state to re-compute relative times. Clear in `componentWillUnmount`. Alternatively, accept this as an intentional trade-off (relative times are approximate in the UI) and document it.

---

#### nit-1 — ChatSummaryPanel handleCreateSubmit uses 800ms setTimeout (ChatSummaryPanel.tsx:141-145)

```typescript
private handleCreateSubmit = (taskId: number) => {
    this.setState({ view: 'list' });
    setTimeout(() => {
        WKApp.mittBus.emit("summary-list-refresh-requested" as any);
    }, 800);
};
```

The 800ms delay is a heuristic to let the backend process the new task. If the API is slow (>800ms), the refresh may fire before the task is queryable. Consider emitting the refresh from the API response callback instead.

---

#### nit-2 — VoiceInputButton viewport clamping uses `window.innerWidth/Height` (VoiceInputButton/index.tsx:27-47)

The position calculation reads `window.innerWidth` and `window.innerHeight` on every render. This works for the typical case but won't respond to viewport resize during active recording. Low risk — the button repositions on the next state update.

---

### 3. Data-Flow Backtracking

|---|---|---|
| `SummaryListPage.state.items` | `api.listSummaries()` → `resp.items` | ✅ Yes, via `setState` in `loadData`/`loadMore` |
| `SummaryCard.task` prop | Parent `SummaryListPage.render()` maps `state.items` | ✅ Yes |
| `SummaryDetailPage.taskId` | `ChatSummaryPanel.state.selectedTaskId` → prop | ✅ Yes |
| `ChatSelectorModal.memberList` | `WKSDK.channelManager.syncSubscribes()` → filter/sort | ✅ Yes, with empty-state guard |
| `SummaryCreatePage.chips` (overflow) | `selectedChats`/`selectedMembers` state → `chipRefs` in `componentDidUpdate` | ✅ Yes, ResizeObserver fires after paint |
| `CitationBadge.sender_uid` | API response → `CitationItem.sender_uid` | ✅ Added to type, consumed in `MessageAvatar` |
| `handleRegenerate/handleEdit` event detail | `SummaryListPage` dispatches `{ taskId }` → `SummaryDetailPage` reads `(e as CustomEvent).detail.taskId` | ⚠️ Yes, but only if 300ms window is sufficient (see P2-3) |
| `SummaryListPage.attentionCount` | `api.listSummaries()` → `resp.attention_count` | ✅ Yes, stored on instance + emitted via `emitBadgeUpdate` |

---

### 4. Blind Spot Checklist (C1–C6)

| Check | Status | Evidence |
|---|---|---|
| **C1 — Dual-path parity** | N/A | No symmetric add/remove or create/delete pairs with divergence risk. |
| **C2 — Control-flow ordering** | Clear | Event listeners in `componentDidMount`/removed in `componentWillUnmount` are correctly paired across SummaryListPage and SummaryDetailPage. |
| **C3 — Authorization boundary** | N/A | No new endpoints or tool exposure; all API calls use existing authenticated `api.*` helpers. |
| **C4 — Auth lifecycle / container state** | N/A | No auth/jail/permission changes. |
| **C5 — Build/note pass ≠ runtime** | Clear | No new build-time constructs; CSS-in-JS changes are runtime-safe. |
| **C6 — Governance/doc self-consistency** | N/A | No governance or policy document changes. |

---

### 5. Cross-Round Blocker Review (R6)

---

### 6. Summary

This PR is a substantial UI redesign of the Summary module, moving from a full-page route-based layout to an embedded panel within the chat sidebar. The architecture is sound: clean separation of list/create/detail views within `ChatSummaryPanel`, proper event-based communication between components, and consistent i18n coverage.

All identified issues are P2 or nit-level. No production-blocking defects (P0/P1) were found. The most notable concern is the fragile 300ms setTimeout race in cross-component event dispatch (P2-3) and the scroll-position loss on delete after infinite scroll (P2-4) — both are UX reliability issues worth addressing before merge but do not break core functionality.

---

[Octo-Q] verdict: APPROVE — no P0/P1 blockers found. 5 P2 findings and 2 nits identified; all are quality/UX improvements that can be addressed in follow-up.

cnwenf (commented):
# Code Review — PR #1044 (Mininglamp-OSS/octo-web)

>
> ⚠️ **Head-SHA drift**: all three reviews were conducted against head `024dee58c4b52819f56f2874060d84672996309e`. The PR head has since advanced to `9a08fe5f798f4f490bdc8212c27a7fcf682e41df`. Findings below describe the reviewed SHA and should be re-verified against the latest commits before acting; this consolidated review is therefore posted as a comment (not a change-request) to avoid blocking already-advanced code on potentially-stale findings.

## Summary

This PR is a comprehensive Figma-driven rewrite of the Summary module (`packages/dmworksummary`, 26 files, +3591/-1651): SummaryCard/SummaryEditor/ChatSummaryPanel rebuilt around Dropdown/Modal menus, the in-chat panel now embeds SummaryListPage/SummaryCreatePage/SummaryDetailPage, an infinite-scroll list, reworked citation badges, member management moved from an org-wide picker to channel subscribers, and removal of the global summary modal. All three reviewers agree security is clean and the bundle builds. However, all three independently conclude the rewrite needs changes: it ships multiple P1 regressions — a broken test suite that CI never runs, a required i18n gate that now fails, a citation reading-order index regression, and an undeclared forward-to-chat wire-format downgrade — plus a long tail of lost re-entrancy guards, fabricated channels, unscoped events, and untested wiring.

## Findings

### P1 — CitationBadge renders raw pool indices; displayIndex/displayIndices contract is dead (packages/dmworksummary/src/components/CitationBadge.tsx:272) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The rewrite dropped `displayIndex`/`displayIndices`: single badges render `[{index}]` (`:272`) and group labels `${indices[0]}-${indices[last]}` (`:282`/`:334`), while the unchanged caller `CitationText.tsx:221`/`:232` still passes the #1003 reading-order values. Probe: badges show `[37]`/`[37-40]` where base showed `[5]`/`[1,2]`; non-contiguous groups like `[2,5,9,14]` misrender as the implied range "2-14" (`formatGroupLabel` is now dead code). The type break is invisible to CI (no `tsc` step; `vite build` does not type-check). This reverses the reading-order renumbering from #1003. Fix: restore the display-index rendering (re-add the props and use them in the label, and keep `formatGroupLabel(indices)`), or migrate `CitationText.tsx` to stop passing them and confirm raw-pool display is intended.

### P1 — Diff breaks the @dmwork/summary test suite: 12 new failing files / 40 tests (packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx:141) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

Base-controlled `vitest run` in an isolated clone: head = 13 files / 41 tests failed; base = 1 file / 1 test (pre-existing). Two mechanisms: (a) 7 files fail collection because `CitationBadge.tsx:6` / `ChatSelectorModal.tsx:6` / `SummaryCreatePage.tsx:13` newly deep-import `@octo/base/src/Components/WKAvatar` but `vitest.config.ts:29` has no alias for it; (b) stale mocks/tests against the rewritten components — `SummaryCard` 18/18 (the `vi.mock('@douyinfe/semi-ui')` lacks the new `Dropdown`/`Modal`, so every render throws "Element type is invalid"), `SummaryEditor` 7/8 (asserts removed inline 保存/取消 buttons), `ChatSummaryHistory` 8/11, `ChatSummaryNewModal` 4/27 (stale i18n strings). Fix: add the WKAvatar alias and update the stale semi-ui mocks/tests in-PR. (Note: octo-review-glm5.2's run reported the PR-touched tests passing 20/20 with zero new failures vs base — a narrower-scope dissent; see E2E row.)

### P1 — CI runs no test/tsc step for a 26-file refactor (.github/workflows/ci.yml:1) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The build job runs `install → i18n:check → build (Vite, no type-check) → lint → lint:wkmodal` and never runs `pnpm test`/`vitest run` or `tsc --noEmit`. Every test breakage above and every excess-prop type error (e.g. the displayIndex contract break) is invisible to CI, so this PR can merge with broken tests and type errors. Fix: add a `pnpm test --run` and `tsc --noEmit` step (or job) gating non-docs PRs.

### P1 — Required CI gate `pnpm i18n:check` fails (packages/dmworksummary/src/components/CitationBadge.tsx:247) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.8=P1, qwen3.7=P2, glm5.2=P2)

The real gate exits 1 at head (exit 0 at base): missing-used-key `summary.citation.sourceDefault` in both locales (used at `:247`/`:309`), missing-used-key `summary.create.desc` in both locales (the key was deleted by this PR while `ChatSummaryNewModal.tsx` still references it), and a new hardcoded Chinese `请输入` in `SummaryDetailPage.tsx:3774` (the regenerate modal textarea, replacing the old `t('summary.detail.refineFeedbackPlaceholder')`). The keys are absent from *both* locales so naive parity looks fine, but the used-key scan catches them and i18next renders the raw key strings at runtime. The PR cannot pass CI as-is. Fix: add `sourceDefault` and `summary.create.desc` to both `en-US.json` and `zh-CN.json`, and route `请输入` through a `t()` key.

### P1 — Forward-to-chat silently downgraded from rich interactive cards to plain text (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1983) [source: octo-review-qwen3.7 + octo-review-glm5.2]

`handleForwardToChat` was silently rewritten from a structured `SummaryCardContent` card (per-channel share grant via `api.createSummaryShares`, schemaVersion 2 with taskId/taskNo/shareId/title/sourceName/preview/timeRange, registered message content type 15, `api.revokeSummaryShare` on failure) to fragmented `new MessageText(c)` plain-text chunks with citation markers stripped (`/[\d+]/g`). New forwards arrive as flat text while the type-15 card renderer stays registered in `dmworkbase`, so old history cards render rich and new forwards render plain — a split wire contract; `createSummaryShares`/`revokeSummaryShare` are now dead in production. This wire-format behavioral change is not declared in the PR description. Fix: either restore the card/grant/revoke path, or declare the change and remove the now-dead share API + type-15 registration (coordinate cross-repo with octo-server).

### P1 — Card 取消任务 is wired to the leave endpoint, not cancel (packages/dmworksummary/src/components/SummaryCard.tsx:158) [source: octo-review-qwen3.8]

The new creator card menu item labeled 取消任务 (i18n `cancelTask`, shown for generating/WAITING_CONFIRM tasks, lines 158-164) calls `onLeave?.(task.task_id)` → `SummaryListPage.handleLeave` → `api.leaveSummary` + a "已退出/Left" toast. The same intent on the detail page routes through `handleCancel` → `api.cancelSummary` (`SummaryDetailPage.tsx:1573`) — the sibling sink the card bypasses. Probe: clicking 取消任务 on a creator's PROCESSING task calls `leaveSummary(42)` and never `cancelSummary`; at base the menu item did not exist. Either the backend rejects a creator leaving their own task (user sees leaveFailed for a cancel) or it applies leave semantics to a cancel. Fix: wire the card action to `api.cancelSummary` with a cancel-specific toast, mirroring the detail page.

### P2 — Forward-to-chat lost its in-flight/re-entrancy guard → duplicate sends (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1992) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P1, glm5.2=P2, qwen3.8=P2)

The rewrite deleted the `forwardingToChat` flag entirely — the old code gated entry with `if (... || this.state.forwardingToChat) return;`, set the flag before the serial multi-chunk send, reset it in `finally`, and bound the button's `disabled` to it. The new code only checks `if (!sourceContent.trim()) return;` and the menu item is `disabled={waitingForFallback}` only (an unrelated condition). The sibling forward-to-matter item still keeps its `forwardingToMatter` guard — a half-migration. `ForwardService.send` is serial per channel with a 200ms inter-message delay, so a long forward takes seconds during which a re-click fires a second `send` with the same chunks, duplicating messages in every selected channel. Fix: re-add a `forwardingToChat` state (set before send, cleared in finally) and bind it to the item's `disabled`.

### P2 — `addingMember` state removed: loading dead + duplicate POST /members (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3174) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P1, glm5.2=P2, qwen3.8=P2)

`addingMember` was removed from the `SummaryDetailPageState` interface and initial state (its slot replaced by a duplicated `forwardingToMatter: false`, `:194-195`), but the confirm `Button` still binds `loading={this.state.addingMember}` — permanently `undefined`: no spinner, no re-entrancy guard, so rapid Confirm clicks fire duplicate `api.addMembers` adding the same users. It is also a type error under the package's own `typecheck` script, which CI does not run. Fix: re-add `addingMember`, set true before the `await`/false after, bind to both `loading` and `disabled`, and delete the duplicate state key.

### P2 — `handleOpenAddMember` mis-maps and fabricates the subscriber channel (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3153) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`origin_channel_type === 1 ? 2 : === 3 ? 1 : 2` maps a thread (origin 2) onto a GROUP channel (should be `5`/CommunityTopic), and when `origin_channel_id` is absent the fallback is `new WkChannel(this.taskId.toString(), 2)` — a group channel keyed by the numeric task id that cannot exist (probe: `{origin_channel_id:''}, 42` → `Channel{'42', 2}`). `SubscriberList` then runs `syncSubscribes` on a nonexistent group, returns empty/wrong, and the user sees an empty member list with no explanation; since the source changed from org-wide search to channel subscribers, tasks without an origin channel can no longer add members at all. Fix: use the shared channelType/channelConvert helpers (add `origin_channel_type === 2 ? 5`), and fail closed (Toast + no roster) when `origin_channel_id` is missing rather than constructing a fake channel.

### P2 — 300ms setTimeout race can silently lose regenerate/edit (packages/dmworksummary/src/pages/SummaryListPage.tsx:434) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P2, glm5.2=P2, qwen3.8=P3)

`handleRegenerate`/`handleEdit` navigate to the detail page then dispatch a `summary-detail-regenerate`/`summary-detail-edit` `CustomEvent` inside a hardcoded 300ms `setTimeout`. The detail page registers its `window` listener only in `componentDidMount`, so on a slow mount/route transition the event fires before the listener exists and is permanently lost — the modal never opens, no error (probe: a listener attached at 400ms never receives it; no replay). The 300ms is an unverified magic number and the handle is never cleared on unmount. Fix: drop the timer; pass intent via route state/query param read in `componentDidMount`, or have the detail page emit a ready signal the list awaits (or use `WKApp.mittBus`).

### P2 — `loadMore` silently swallows pagination failures (packages/dmworksummary/src/pages/SummaryListPage.tsx:242) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P2, glm5.2=P2, qwen3.8=P3)

`} catch { this.setState({ loadingMore: false }); }` resets `loadingMore` but leaves `hasMore=true` and surfaces no error, toast, or retry — indistinguishable from "no more items", while the sibling `loadData` pattern does surface an error. On a persistent backend failure every scroll-near-bottom re-triggers `loadMore`, which re-fails silently — an infinite silent-retry loop. Fix: set an error state, show a retry banner/`Toast.error`, and consider `hasMore=false` after N consecutive failures.

### P2 — `loadMembers` swallows all errors (packages/dmworksummary/src/components/ChatSelectorModal.tsx:111) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`} catch { this.setState({ candidates: [] }); }` makes "channel has no members" indistinguishable from a `syncSubscribes` network failure or SDK exception — both render as an empty list with no error, retry, or Toast. Fix: set a distinct error state and render an Empty-with-retry or `Toast.error`.

### P2 — `handleCreateSubmit` 800ms setTimeout not cleared on unmount (packages/dmworksummary/src/components/ChatSummaryPanel.tsx:142) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P2, glm5.2=P2, qwen3.8=P3)

`handleCreateSubmit` schedules `setTimeout(() => WKApp.mittBus.emit('summary-list-refresh-requested'), 800)` and discards the handle; `componentWillUnmount` clears drag listeners but never `clearTimeout`. A mutation probe (fake timers + unmount before 800ms) confirmed the emit fires after unmount (CONTROL passed, PROBE failed "expected 0, received 1"). Fix: store the handle on the instance, `clearTimeout` in `componentWillUnmount` (and on success), and add a boundary test.

### P2 — `regenerateMode` default silently flipped from "refine" to "full" (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1184) [source: octo-review-qwen3.7 + octo-review-glm5.2]

`handleRegenerate` now sets `regenerateMode: 'full'` (base set `'refine'`) while the initial-state default stays `'refine'`, so opening regenerate from the list now defaults to re-running the whole summary instead of incremental feedback; the refine-mode-aware modal UI was also half-deleted. This behavioral change is not declared in the PR description. Fix: restore `'refine'` as the default, or declare the flip and reconcile the half-removed refine UI.

### P2 — `handleDelete_refetch` discards accumulated pages and corrupts hasMore (packages/dmworksummary/src/pages/SummaryListPage.tsx:366) [source: octo-review-glm5.2 + octo-review-qwen3.8]

After delete it calls `fetchData()`, which reads `this.state.page` and *replaces* `state.items` with that single page — dropping pages 1..N-1 accumulated by infinite scroll — without recomputing `hasMore`. Probe: after loadMore accumulates 20 items, a delete shrinks the list to page 2's 10, and the loadMore boundary guard then computes against the truncated array; if `total` dropped below `pageSize` a `loadMore` fetches an empty page. Fix: refetch from page 1 (resetting hasMore) or splice the deleted item out in place, recomputing `hasMore: fresh.items.length < fresh.total`.

### P2 — `onAfterMutate` pop-to-list wiring has no test (packages/dmworksummary/src/components/ChatSummaryPanel.tsx:220) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The panel wires `onAfterMutate={() => this.setState({ view: 'list', selectedTaskId: null })}` onto `SummaryDetailPage`, which calls it after delete/leave (`:812`/`:831`). The test's `SummaryDetailPage` mock doesn't expose `onAfterMutate` as a triggerable callback, so no test fires it and asserts the panel returns to list. A regression dropping the prop would leave the user stuck on a stale detail view and stay green. Fix: have the mock render a button calling `props.onAfterMutate?.()`, add a test that opens detail, fires it, and asserts list visible + `selectedTaskId` cleared.

### P2 — Member selector invents a channel heuristic: DM sources get an empty roster (packages/dmworksummary/src/pages/SummaryCreatePage.tsx:771) [source: octo-review-qwen3.8 + octo-review-glm5.2] (severity split: qwen3.8=P2, glm5.2=P3)

`handleOpenMemberSelector` derives the roster channel from `chat.chat_id.includes('____') ? 5 : 2`, ignoring `chat_type` and never returning `1` (direct). Probe: a direct-chat id maps to GROUP type 2 instead of PERSON type 1; `ChatSelectorModal.loadMembers` then runs `syncSubscribes` on a nonexistent group, lands in catch, and renders an empty member list with no error — participants can never be selected for a DM source. The file already imports the correct helpers (`chatTypeToOriginChannelType`, `:18`; `ChatSelectorModal.getChannelTypeNum`) but does not use them here. Fix: derive the Channel from the real `chat_type`.

### P2 — Infinite scroll without virtualization (packages/dmworksummary/src/pages/SummaryListPage.tsx:222) [source: octo-review-qwen3.7 + octo-review-glm5.2] (severity split: qwen3.7=P2, glm5.2=P3)

`loadMore` appends items and renders every accumulated item via `items.map()` with no windowing or cap. In a workspace with hundreds+ summaries the DOM holds 1000+ unvirtualized `SummaryCard` subtrees, degrading scroll/memory. Fix: cap the rendered window with `react-window`/`react-virtuoso` or enforce a maximum retained item count.

### P2 — `loadMembers` lacks `reqSeq` sequencing guard (packages/dmworksummary/src/components/ChatSelectorModal.tsx:89) [source: octo-review-glm5.2]

`loadCandidates` uses a `reqSeq` guard (`:120`/`:147`), but `loadMembers` does not. If invoked twice in quick succession (modal re-opened via the visible transition, or channel prop change), an earlier call can resolve after a later one and overwrite fresher `candidates`/`memberRoles` with stale subscriber data. Fix: increment `reqSeq` at the start and guard both success and catch `setState`.

### P2 — Save has no hasChanges/saving guard: no-op PUTs and duplicate concurrent saves (packages/dmworksummary/src/components/SummaryEditor.tsx:71) [source: octo-review-qwen3.8]

With the in-editor buttons removed, the parent footer calls the exposed `handleSave` unconditionally (`SummaryDetailPage.tsx:3715-3719`); the old `disabled={!hasChanges || saving}` gate is gone. Probe: saving with no edits issues `editSummary` and reports success; a second click while the first is in flight fires a concurrent duplicate PUT. Fix: re-add a hasChanges/saving early return in `handleSave` (or gate the footer button).

### P2 — Scheduled-summary delete warning dropped; `isScheduledTask` dead (packages/dmworksummary/src/components/SummaryCard.tsx:89) [source: octo-review-qwen3.8]

The old Popconfirm branched on `isScheduledTask` to warn that deletion stops the schedule and removes its history; the new generic Modal shows "删除后无法恢复" for all deletes. Probe: deleting a scheduled task (`schedule_id=77`) renders only the generic copy; base showed the schedule-stop warning. `isScheduledTask` (`:89`) is still computed but never read, and the scheduled-delete i18n keys are orphaned. Fix: restore the scheduled-specific copy or delete the dead computation/keys.

### P2 — Card-menu Edit bypasses every editability gate (packages/dmworksummary/src/components/SummaryCard.tsx:173) [source: octo-review-qwen3.8]

The card Edit items (FAILED `:173`, COMPLETED `:184`) flip `isEditing` via a window event, skipping the detail page's own `showEdit` requirements (status COMPLETED, `permissions.can_edit`, `trigger_type !== AGENT`, `!isMultiCollab()`). Probe: `handleStartEdit` contains none of those checks. Consequences: FAILED → edit footer without an editor (dead end); AGENT-completed → PUT `/edit` with `base_result_id 0` → the 404 "总结结果不存在" the deleted guard comment warned about; multi-collab → footer without a personal editor. Fix: mirror `showEdit` in the card path (or validate editability in the event handler).

### P2 — `editorSaveFn` slot never cleared on unmount: a stale editor can save captured content (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:233) [source: octo-review-qwen3.8]

Both personal and team editors write the shared `editorSaveFn` slot via `exposeSave`; `componentWillUnmount` never calls `exposeSave(null)` (probe: still 1 call after unmount, never with null). Combined with the card-Edit bypass above: edit team → cancel → card-Edit a FAILED task (no editor mounts, slot not overwritten) → footer Save invokes the stale team editor's `handleSave` → `editSummary` fires with previously captured content and reports success — a bogus edit version. Fix: clear the slot on unmount and null-check before invoking.

### P2 — Detail-page window events are unscoped: hidden instances react too (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:260) [source: octo-review-qwen3.8]

Every mounted instance listens for `summary-detail-regenerate`/`-edit` and matches on `taskId` only (probe: no instance/surface field in the handlers or the payload). Up to three instances can coexist per task (list route, SummaryConfirmPage embed, in-chat panel embed), so one card action puts every matching instance into edit/regenerate — a hidden instance mounts a second editor with its own beforeunload listener. Fix: scope the event to the emitting surface (instance id, or address the detail via prop/callback).

### P2 — `chat-summary-deleted` orphaned: star button keeps stale hasSummaries (packages/dmworksummary/src/components/ChatSummaryPanel.tsx:158) [source: octo-review-qwen3.8]

Replacing ChatSummaryHistory with SummaryListPage orphaned the only dispatcher of `chat-summary-deleted` (probe: neither SummaryListPage nor SummaryDetailPage contains the string), while `ChatSummaryStarButton.tsx:40` still listens for it to re-fetch the channel count. Deleting the channel's last summary via the new UI leaves `hasSummaries=true`, so the star routes into an empty history view instead of the create flow. Fix: dispatch the event (channel-scoped) from the surviving delete paths, or invalidate on the refresh event.

### P2 — Hidden list keeps polling: the duplicate-status-request `paused` used to prevent (packages/dmworksummary/src/components/ChatSummaryPanel.tsx:158) [source: octo-review-qwen3.8]

The old panel passed `paused={isDetail}` to stop ChatSummaryHistory's poller; SummaryListPage has no paused prop and stays mounted (`display:none`) under the detail/create views. Probe: with `view='detail'` the list is hidden but its 2s batchStatus keeps firing — against the same task ids the detail heartbeat already refreshes; two live panel instances double it. Fix: reintroduce a pause or dedupe to a single poller owner.

### P2 — CSS token-policy violations: 13 `!important`, 6 `.semi-*`, 63 hex lines (packages/dmworksummary/src/index.css:1) [source: octo-review-qwen3.8]

Diff-added lines in `index.css` (+1370/-367) and `ChatSummaryNewModal.css` (+85/-1) break the `var(--wk-*)` policy; stylelint reports 153 warnings (14 at base). Not CI-blocking — stylelint exits 0 warnings-only and `lint:wkmodal` passes, baseline-allowlisting the Semi overrides — and dark mode via `body[theme-mode=dark]` still works; this is theme-token tech debt. Fix: migrate the hardcoded colors to tokens when convenient.

### P2 — Channel-scoped filtering and panel badge suppression are untested (packages/dmworksummary/src/pages/SummaryListPage.tsx:196) [source: octo-review-qwen3.8]

Mutation-kill probe: removing `origin_channel_id` from `loadData` leaves every shipped test green (the rewritten ChatSummaryPanel test mocks SummaryListPage entirely and only asserts prop passthrough), while the focused probe dies. The `:325` panel-mode badge guard is likewise unexercised — deleting it would let a single channel's attention_count overwrite the global nav badge. Fix: add SummaryListPage tests with `channelId` asserting the `origin_channel_id` param, filtered items, and no `summary-badge-update` emit.

### P2 — `summaryPanelView` consumer wiring is pinned by no test (packages/dmworksummary/src/components/ChatSummaryStarButton.tsx:121) [source: octo-review-qwen3.8]

The no-summaries branch was rewired from `wk:open-summary-modal` to `emit('wk:toggle-summary-panel', { ..., summaryPanelView: 'new' })` (without `forceOpen`), but the consumer chain is untested. Confirmed by mutation kill: deleting the `summaryPanelView` passthrough at `packages/dmworkbase/src/EndpointCommon.tsx:365` leaves all 19 existing tests green while a focused probe on the real `chatSummaryPanel` invoke fails (`expected undefined to be 'new'`). The emit also lacks `forceOpen`, so with the panel already open the toggle closes it instead of switching to the 'new' view. Fix: add a test that drives the real listener (or the `_onToggleSummaryPanel` reducer) and pins both the 'new'-view render and the already-open toggle behavior.

### P3 — Duplicate `forwardingToMatter: false` in initial state (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:195) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

Copy-paste artifact: the deleted `forwardingToChat` slot was replaced by a second `forwardingToMatter: false` (`:194-195`) — a no-dupe-keys violation. Same value so no runtime bug today, but it masks intent. Fix: delete the duplicate line.

### P3 — Duplicate `./index.css` import + orphaned `notifyChatSummaryCreated` import (packages/dmworksummary/src/module.tsx:20) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`import "./index.css";` appears twice (`:19` and `:20`), and `import { notifyChatSummaryCreated }` (`:13`) is now unreferenced after `GlobalSummaryModal` was deleted. Fix: remove the duplicate line and remove the orphaned import (or restore its usage).

### P3 — Stray double semicolon (packages/dmworksummary/src/pages/SummaryListPage.tsx:446) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`handleEdit` arrow function closes with a stray `};;`. Fix: remove the extra semicolon.

### P3 — Stale `wk:open-summary-modal` event-type declaration (packages/dmworkbase/src/App.tsx:38) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The PR removed the only emitter (`mountGlobalSummaryModal`/`GlobalSummaryModal`) and listener, but the event-type map still declares `wk:open-summary-modal` — zero emitters, zero listeners remain. Fix: remove the key from the event-type map.

### P3 — `ChatSummaryNewModal` dead `embedded` prop/CSS (packages/dmworksummary/src/components/ChatSummaryNewModal.tsx) [source: octo-review-qwen3.8]

The panel embeds SummaryCreatePage instead; the modal is reachable only from stories, still carries a dead `embedded` prop + CSS, and still renders the deleted `summary.create.desc` key. Fix: wire or drop the embedded path.

### P3 — `ChatSummaryHistory` is dead code (packages/dmworksummary/src/components/ChatSummaryHistory.tsx:11) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The component was modified this PR (new `onClose` prop, restructured close button) but has zero production importers — `ChatSummaryPanel` now uses `SummaryListPage` directly; only its own test file imports it, so updated tests give false confidence. Fix: re-wire it where it should render, or delete the component and its tests.

### P3 — `handleRetry` lacks the task-switch race guard its siblings use (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:1189) [source: octo-review-glm5.2 + octo-review-qwen3.8]

Unlike `handleLeaveTask`/`handleDeleteTask` (which snapshot `requestTaskId` and abort on mismatch — the explicit "FE-1 切 task 竞态" pattern), `handleRetry` applies its post-await Toast/`regenerateSubmitting`/`loadDetail()` to whichever task is current (probe: after a 1→2 switch during the await, loadDetail and the error toast land on task 2). Fix: capture `const requestTaskId = this.taskId;` and guard after the await.

### P3 — Add-member confirm `onClick` lacks task-identity/unmounted guard (packages/dmworksummary/src/pages/SummaryDetailPage.tsx:3179) [source: octo-review-qwen3.8]

The add-member confirm `onClick` awaits `api.addMembers` then calls `WKApp.routeRight.pop()`/`this.loadDetail()` with no `requestTaskId`/`unmounted` guard after the await (unlike the sibling leave/delete handlers). On navigation during the await it pops/loads the wrong page. Fix: capture `const requestTaskId = this.taskId` and guard before pop/loadDetail.

### P3 — `SummaryCreatePage` dead code (packages/dmworksummary/src/pages/SummaryCreatePage.tsx:779) [source: octo-review-glm5.2]

`handleMemberSelectorConfirm` is defined but never called (the modal uses an inline `onConfirmMembers`); `SubscriberList`/`RoutePage` imports are orphaned; state fields `memberSelectorExcluded`/`memberSelectorSelectedItems`/`memberSelectorOnSelect` are destructured but unused. Fix: delete the dead handler, imports, and unused state fields.

### P3 — Close-from-list-view test deleted, not replaced (packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx:35) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The old "closes the panel via the close button in list view" test was deleted; the close affordance was delegated to `SummaryListPage` (which receives `onClose`). The `SummaryListPage` mock renders a list-close button calling `props.onClose`, but no test clicks it and asserts `onClose`. If `SummaryListPage` stops forwarding `onClose`, the panel can't be closed from list view and no test catches it. Fix: add a test that fires list-close and asserts `onClose` called once.

### P3 — `ChatSummaryPanel` test mocks all child pages (packages/dmworksummary/src/components/__tests__/ChatSummaryPanel.test.tsx:75) [source: octo-review-qwen3.7]

Tests mock SummaryListPage/SummaryDetailPage/SummaryCreatePage entirely, testing only view-switching logic rather than prop passing to real components. Fix: add at least one integration-style test against the real children.

### P3 — `onCancel` is a required prop with no consumer (packages/dmworksummary/src/components/SummaryEditor.tsx:107) [source: octo-review-qwen3.8]

Still declared required and destructured, but the Cancel button is gone and no rendered element can invoke it (probe: spy stays 0 at head, 1 at base); all four call sites still pass it. Fix: make it optional or remove it and update callers.

### P3 — `FORGE_RENDER_PROFILE` branch is a silent no-op (packages/dmworkbase/src/Messages/InteractiveCard/sdk/renderOctoCard.ts:51) [source: octo-review-qwen3.8]

Both ternary arms now return `buildOctoHostConfig(browserCssVarResolver(target))` (probe: the forge config is byte-identical to the default; at base it was the distinct `forgeHostConfig` JSON), and InteractiveCardCell's forge theme/styles imports are commented out — an undeclared InteractiveCard rendering change tangential to the Summary refactor and absent from the near-empty PR body. Fix: confirm the forge profile is intentionally retired (and delete the dead ternary) or restore it.

### P3 — `handleCreateSubmit`: discarded taskId, magic-800ms refresh bet, unpinned view return (packages/dmworksummary/src/components/ChatSummaryPanel.tsx:140) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`handleCreateSubmit(taskId)` declares `taskId` but never uses it — it flips to list, then emits `summary-list-refresh-requested` after a bare `setTimeout(...,800)` with no payload/ack/retry (probe: 0 emits at 799ms, exactly 1 at 800ms). If the task is not visible in the list query by then, the created summary silently does not appear; no shipped test pins the post-submit return to the list view (the test passes `taskId=99` but only asserts the emit). Fix: navigate to the detail via the taskId you already have (or refresh until visible) and pin the view-return in a test.

- **Security (CLEARED, all three reviewers)** — STRIDE + OWASP Top 10 + credential scan + source→sink dataflow all clean: no `dangerouslySetInnerHTML`/`innerHTML`/`eval`/`new Function`, no hardcoded secrets or env exposure, no new auth paths/endpoints/`window.open`/`fetch`; user input is length-bounded (`maxLength` + `SUMMARY_INPUT_MAX_LENGTH`, regenerate `.slice(0,1000)`); rendering goes through React JSX auto-escape or ReactMarkdown + rehype-sanitize; permission checks (`can_edit`/`can_schedule`/`isCreator`/`isParticipant`) are preserved and fail-closed (null-guarded, strict equality); `renderOctoCard` host-config is now more constrained (opaque external JSON replaced by `buildOctoHostConfig`). The forward-to-chat change moves from share-card grants to plain `MessageText`, shrinking attack surface. One pre-existing informational (not introduced here): `ChatSummaryHistory.tsx` still imports `axios`.
- **Build/bundle** — `vite build` (mock mode) succeeds (`✓ built in ~6s`); all TypeScript compiles into apps/web. (This is also why the stale mocks and the displayIndex type break do not surface in CI.)
- **i18n parity of added keys** — all 28 new/changed keys are present in both `en-US.json` and `zh-CN.json` with matching placeholders. (Separately, the used-key gate still fails on the missing `sourceDefault`/`summary.create.desc` keys and the hardcoded `请输入` — see the P1 above.)
- **Panel state isolation/cleanup** — `ChatSummaryPanel` correctly resets `view`/`selectedTaskId` on channel change; `emitBadgeUpdate` early-returns in panel mode so panel counts do not clobber the global nav badge.
- **Deleted global-modal cleanup** — the removed `GlobalSummaryModal`/`mountGlobalSummaryModal` path had its `mittBus` emitter and listener removed together; no dangling `on('wk:open-summary-modal', ...)` registration survives (only the stale type declaration noted as P3).
- **Component layering** — no new violations of the Layer 1/2/3 dependency rules.
- **Normal pagination** — `hasMore` calculation is correct on the load path (cursor and count stay consistent); the delete-refetch path is the exception (see P2 above).
- **Lint gates** — `lint:wkmodal` and `stylelint` exit 0 (Semi overrides baseline-allowlisted as existing debt).

## Verification Results

| Code Review | NEEDS_CHANGES | 6 P1 + 23 P2 + 14 P3 (deduped across 3 reviewers); blockers: test-suite breakage CI never runs, required i18n:check gate failure, citation display-index regression, undeclared forward-to-chat wire downgrade, 取消任务 mis-wired to leave |
| E2E Test | FAIL | 2/3 reviewers' base-controlled `vitest run` show ~40 new failing tests (missing WKAvatar vitest alias + stale semi-ui Dropdown/Modal mocks) and the required `pnpm i18n:check` gate exits 1; one reviewer (glm5.2) reported the PR-touched tests passing 20/20 with zero new failures vs base (narrower-scope dissent). Build passes in all three runs. |
| Security | CLEARED | All three reviewers: 0 exploitable findings — STRIDE/OWASP Top 10/credential scan/source→sink clean; permission checks fail-closed; one pre-existing informational (axios import) not introduced by this diff |

PR #1046, opened 2026-07-23
feat(docs): unify document import and export

### LINKED ISSUE #1045: Unify frontend document import and export with authoritative backend pipelines
## Problem

The document frontend still parses Markdown and serializes Markdown/DOCX from browser editor JSON, while the backend now provides unified import/export pipelines backed by the authoritative live Y.Doc. This can make browser downloads differ from current collaborative state and leaves Markdown import on a separate parser. Calibri values preserved from Office files are also not selectable in the Doc/Sheet catalogs.

## Expected behavior

- Markdown import uses the backend Markdown importer, matching the existing backend DOCX path.
- Markdown, DOCX, and PDF downloads use the unified backend file export endpoint and current live document state.
- The full-width editor route remains uncapped.
- Calibri is selectable in Doc and Sheet without changing Board behavior.
- Focused API/flow/UI tests cover the integration.

## Scope

Frontend integration only. Board native panels, palettes, text runs/defaults, extra tools, shape controls, and unrelated Docs Home/Portal menu changes are excluded.

wwwzzzcccc:
## Summary

Connect rich-document import and download actions to the backend's unified, authoritative live-document pipelines.

## Related Issue

Fixes #1045

## Changes

- Route active Markdown import through `docsApi.importMarkdown`; keep DOCX on its backend importer.
- Route Markdown, DOCX, and PDF downloads through `docsApi.exportDocFile`, avoiding local editor JSON serialization.
- Preserve the current uncapped full-width editor route.
- Add Calibri to the Doc and Sheet font catalogs and locale labels.
- Add focused tests for backend import/export requests, import flow ordering/stashing, and editor download behavior.

- Affected module(s): `@octo/docs` editor, docs API adapter, and Sheet font registration
- New or changed user-visible entry point: yes; existing Markdown import and MD/DOCX/PDF export actions now use backend pipelines
- Shared layer touched: service
- If shared code changed, impact scope: document import/export API calls within the Docs package only
- Duplicate entry point checked: yes

## Testing

- [ ] Manually verified
- `pnpm --filter @octo/docs exec vitest run src/pages/docsApi.export.test.ts src/pages/docsApi.import.test.ts src/editor/importFlow.markdown.test.ts src/editor/EditorShell.test.tsx src/editor/fontFamilies.verify.test.ts` — 5 files, 23 tests passed
- `pnpm i18n:check` — passed
- `pnpm --filter @octo/docs typecheck` — only two confirmed pristine-main baseline errors remain: `dmworkbase/src/i18n/format.ts:38` (`fractionalSecondDigits`) and `docs/src/members/sort.ts:26` (missing `grantedBy`)
- `git diff --check` — passed

## Checklist

- [x] Updated documentation (not applicable; no public documentation contract changed)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,14 @@ jobs:
         run: pnpm install --frozen-lockfile
       - name: I18n check
         run: pnpm i18n:check
+      - name: Unified import/export unit tests
+        run: >-
+          pnpm --filter @octo/docs exec vitest run
+          src/editor/EditorShell.test.tsx
+          src/editor/fontFamilies.verify.test.ts
+          src/editor/importFlow.markdown.test.ts
+          src/pages/docsApi.export.test.ts
+          src/pages/docsApi.import.test.ts
       - name: Build
         run: pnpm build
       - name: Lint
--- a/packages/docs/src/editor/EditorShell.test.tsx
+++ b/packages/docs/src/editor/EditorShell.test.tsx
@@ -26,11 +26,12 @@ vi.mock('../collab/useCollabEditor.ts', () => ({
   }),
 }))
 
-// Capture the markdown the handler serializes; the body content is irrelevant here.
-const exportSpy = vi.fn(async (..._args: unknown[]) => '# exported\n')
-vi.mock('../export/markdown.ts', () => ({
-  exportDocToMarkdown: (...args: unknown[]) => exportSpy(...args),
-}))
+// Capture calls to the single authoritative backend export endpoint.
+const exportSpy = vi.fn(async (..._args: unknown[]) => new ArrayBuffer(8))
+vi.mock('../pages/docsApi.ts', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../pages/docsApi.ts')>()
+  return { ...actual, exportDocFile: (...args: unknown[]) => exportSpy(...args) }
+})
 
 // Stub the presentational children that take the live editor/provider — they are not under test.
 vi.mock('@tiptap/react', () => ({ EditorContent: () => null }))
@@ -135,9 +136,10 @@ describe('EditorShell — export filename uses the live title, not the stale pro
 
     // Export moved into the header ≡ "more" menu: open it, then trigger the export row.
     fireEvent.click(screen.getByTitle('docs.toolbar.more'))
+    fireEvent.click(screen.getByText('docs.toolbar.export'))
     fireEvent.click(screen.getByText('docs.toolbar.exportMarkdown'))
 
-    await waitFor(() => expect(exportSpy).toHaveBeenCalledTimes(1))
+    await waitFor(() => expect(exportSpy).toHaveBeenCalledWith('d_1', 'md'))
     await waitFor(() => expect(createdAnchors.length).toBeGreaterThan(0))
     const a = createdAnchors[createdAnchors.length - 1]
     expect(a.download).toBe('Live Title.md')
@@ -166,6 +168,7 @@ describe('EditorShell — header injection props (#512 AC-8)', () => {
     // The low-frequency controls now live behind the ≡ "more" menu, not as resident buttons.
     expect(screen.getByTitle('docs.toolbar.more')).toBeTruthy()
     fireEvent.click(screen.getByTitle('docs.toolbar.more'))
+    fireEvent.click(screen.getByText('docs.toolbar.export'))
     expect(screen.getByText('docs.toolbar.exportMarkdown')).toBeTruthy()
     expect(screen.getByText('docs.toolbar.history')).toBeTruthy()
   })
@@ -184,6 +187,7 @@ describe('EditorShell — header injection props (#512 AC-8)', () => {
     expect(screen.getByTitle('docs.list.back')).toBeTruthy()
     // ...and the built-in controls are still reachable via the ≡ menu (parity preserved).
     fireEvent.click(screen.getByTitle('docs.toolbar.more'))
+    fireEvent.click(screen.getByText('docs.toolbar.export'))
     expect(screen.getByText('docs.toolbar.exportMarkdown')).toBeTruthy()
   })
 })
@@ -207,6 +211,7 @@ describe('EditorShell — header "more" (≡) menu', () => {
     // Closed by default: no rows visible.
     expect(screen.queryByText('docs.toolbar.exportMarkdown')).toBeNull()
     fireEvent.click(screen.getByTitle('docs.toolbar.more'))
+    fireEvent.click(screen.getByText('docs.toolbar.export'))
 
     const historyRow = screen.getByText('docs.toolbar.history').closest('button')!
     const exportRow = screen.getByText('docs.toolbar.exportMarkdown').closest('button')!
--- a/packages/docs/src/editor/EditorShell.tsx
+++ b/packages/docs/src/editor/EditorShell.tsx
@@ -15,8 +15,7 @@ import { useDocComments, useRefreshCommentsOnOpen } from '../comments/useDocComm
 import { useCommentHighlights } from '../comments/useCommentHighlights.ts'
 import { useDocDelete } from './useDocDelete.ts'
 import { useMemberNames } from '../members/useMemberNames.ts'
-import { exportDocToMarkdown, type MdNode } from '../export/markdown.ts'
-import { exportDocPdf } from '../pages/docsApi.ts'
+import { exportDocFile, type DocExportFormat } from '../pages/docsApi.ts'
 import { emojiGlyph } from './emoji.ts'
 import { colorFromId } from '../awareness/presence.ts'
 import { useEffect, useState, useRef, useCallback, type ReactNode } from 'react'
@@ -346,16 +345,16 @@ export function EditorShell(props: EditorShellProps) {
       }
       return
     }
+    // Atomic backend imports carry no PM stash; surface their small warning payload independently.
+    const warnings = consumeImportWarnings(docId)
+    if (warnings.length) setImportNotice(warnings.join(' '))
     if (!pmDoc) return
     try {
       ed.commands.setContent(pmDoc as never)
     } catch (err) {
       console.error('[docs] Import content injection failed:', err)
       setImportNotice(t('docs.toolbar.importCorrupt'))
-      return
     }
-    const warnings = consumeImportWarnings(docId)
-    if (warnings.length) setImportNotice(warnings.join(' '))
   }, [instance, ready, docId, t])
 
   // uid → display name for this space (#8): once resolved, push the real name into awareness so
@@ -469,93 +468,40 @@ export function EditorShell(props: EditorShellProps) {
   )
   const del = useDocDelete(docId, handleDeleted)
 
-  // Export-as-Markdown (batch 8, area C): serialize the live PM JSON to Markdown with freshly
-  // resolved signed asset URLs and trigger a browser download. Display-only — never mutates
-  // the doc; failures surface a small banner instead of crashing the editor.
+  // All downloadable document formats come from one authenticated backend
+  // endpoint. The backend reads the authoritative live Y.Doc; the browser does
+  // not serialize its potentially stale local editor snapshot.
   const [exporting, setExporting] = useState(false)
   const [exportError, setExportError] = useState<string | null>(null)
-  const onExportMarkdown = useCallback(async () => {
-    const ed = instance?.editor
-    if (!ed || exporting) return
+  const downloadExport = useCallback(async (format: DocExportFormat) => {
+    if (exporting) return
     setExporting(true)
     setExportError(null)
     try {
-      const md = await exportDocToMarkdown(docId, ed.getJSON() as unknown as MdNode, { emojiGlyph })
-      const blob = new Blob([md], { type: 'text/markdown;charset=utf-8' })
-      const url = URL.createObjectURL(blob)
+      const bytes = await exportDocFile(docId, format)
+      const mime = format === 'md'
+        ? 'text/markdown;charset=utf-8'
+        : format === 'docx'
+          ? 'application/vnd.openxmlformats-officedocument.wordprocessingml.document'
+          : 'application/pdf'
+      const url = URL.createObjectURL(new Blob([bytes], { type: mime }))
       const a = document.createElement('a')
       a.href = url
-      a.download = `${exportDownloadName(currentTitle)}.md`
-      document.body.appendChild(a)
-      a.click()
-      a.remove()
-      // Defer revoke so Safari/iOS can start the download for larger blobs
-      // (matches the PDF branch).
-      setTimeout(() => URL.revokeObjectURL(url), 5000)
-    } catch (err) {
-      console.error('[docs] Markdown export failed:', err)
-      setExportError(t('docs.toolbar.exportError'))
-    } finally {
-      setExporting(false)
-    }
-  }, [instance, docId, currentTitle, exporting])
-
-  const onExportDocx = useCallback(async () => {
-    const ed = instance?.editor
-    if (!ed || exporting) return
-    setExporting(true)
-    setExportError(null)
-    try {
-      // Lazy-load the DOCX exporter so `docx` + `mathjax-full` (multi-MB, DOCX-only)
-      // stay out of the eager editor bundle that loads on every document open.
-      const { exportDocToDocx } = await import('../export/docx/index.ts')
-      const blob = await exportDocToDocx(docId, ed.getJSON() as unknown as MdNode, { emojiGlyph })
-      const url = URL.createObjectURL(blob)
-      const a = document.createElement('a')
-      a.href = url
-      a.download = `${exportDownloadName(currentTitle)}.docx`
-      document.body.appendChild(a)
-      a.click()
-      a.remove()
-      // Defer revoke so Safari/iOS can start the download for larger blobs
-      // (matches the PDF branch).
-      setTimeout(() => URL.revokeObjectURL(url), 5000)
-    } catch (err) {
-      console.error('[docs] DOCX export failed:', err)
-      setExportError(t('docs.toolbar.exportError'))
-    } finally {
-      setExporting(false)
-    }
-  }, [instance, docId, currentTitle, exporting])
-
-  const onExportPdf = useCallback(async () => {
-    const ed = instance?.editor
-    if (!ed || exporting) return
-    setExporting(true)
-    setExportError(null)
-    try {
-      // Server-side PDF render via Typst (小吴's decision to go backend). The
-      // backend renders its own persisted copy of the doc with Typst, so the
-      // output has real CJK, rendered math, emoji, selectable text and smart
-      // pagination — one-click download with no print dialog. Returns the PDF
-      // bytes directly.
-      const bytes = await exportDocPdf(docId)
-      const blob = new Blob([bytes], { type: 'application/pdf' })
-      const url = URL.createObjectURL(blob)
-      const a = document.createElement('a')
-      a.href = url
-      a.download = `${exportDownloadName(currentTitle)}.pdf`
+      a.download = `${exportDownloadName(currentTitle)}.${format}`
       document.body.appendChild(a)
       a.click()
       a.remove()
       setTimeout(() => URL.revokeObjectURL(url), 5000)
     } catch (err) {
-      console.error('[docs] PDF export failed:', err)
+      console.error(`[docs] ${format.toUpperCase()} export failed:`, err)
       setExportError(t('docs.toolbar.exportError'))
     } finally {
       setExporting(false)
     }
-  }, [instance, docId, currentTitle, exporting])
+  }, [docId, currentTitle, exporting])
+  const onExportMarkdown = useCallback(() => void downloadExport('md'), [downloadExport])
+  const onExportDocx = useCallback(() => void downloadExport('docx'), [downloadExport])
+  const onExportPdf = useCallback(() => void downloadExport('pdf'), [downloadExport])
 
   const togglePanel = useCallback(
     (panel: Exclude<DrawerPanel, null>) => setActivePanel((cur) => (cur === panel ? null : panel)),
--- a/packages/docs/src/editor/fontFamilies.ts
+++ b/packages/docs/src/editor/fontFamilies.ts
@@ -23,6 +23,7 @@
  * time (zh-CN shows the Chinese face name, en-US the romanized family name).
  */
 export const FONT_FAMILIES = [
+  { labelKey: 'docs.toolbar.font.calibri', value: 'Calibri, Carlito, sans-serif' },
   { labelKey: 'docs.toolbar.font.timesNewRoman', value: '"Times New Roman", Times, serif' },
   { labelKey: 'docs.toolbar.font.tahoma', value: 'Tahoma, sans-serif' },
   { labelKey: 'docs.toolbar.font.verdana', value: 'Verdana, sans-serif' },
--- a/packages/docs/src/editor/fontFamilies.verify.test.ts
+++ b/packages/docs/src/editor/fontFamilies.verify.test.ts
@@ -31,6 +31,7 @@ describe('FONT_FAMILIES i18n refactor (XIN-936)', () => {
     expect(byKey['docs.toolbar.font.simsun']).toBe('SimSun, "宋体", serif')
     expect(byKey['docs.toolbar.font.simhei']).toBe('SimHei, "黑体", sans-serif')
     expect(byKey['docs.toolbar.font.kaiti']).toBe('KaiTi, "楷体", serif')
+    expect(byKey['docs.toolbar.font.calibri']).toBe('Calibri, Carlito, sans-serif')
   })
 
   it('resolves every labelKey in both zh-CN and en-US', () => {
--- a/packages/docs/src/editor/importFlow.markdown.test.ts
+++ b/packages/docs/src/editor/importFlow.markdown.test.ts
@@ -1,0 +1,134 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const createDoc = vi.fn()
+const deleteDoc = vi.fn()
+const importMarkdown = vi.fn()
+const importDocx = vi.fn()
+vi.mock('../pages/docsApi.ts', () => ({
+  createDoc: (...args: unknown[]) => createDoc(...args),
+  deleteDoc: (...args: unknown[]) => deleteDoc(...args),
+  importMarkdown: (...args: unknown[]) => importMarkdown(...args),
+  importDocx: (...args: unknown[]) => importDocx(...args),
+}))
+vi.mock('../attachments/api.ts', () => ({ copyAttachments: vi.fn(), ingestAttachments: vi.fn() }))
+
+import { consumeImportContent, consumeImportWarnings, runDocxImport, runMarkdownImport } from './importFlow.ts'
+
+function choose(input: HTMLInputElement, file: File): void {
+  Object.defineProperty(input, 'files', { configurable: true, value: [file] })
+  input.onchange?.(new Event('change'))
+}
+
+beforeEach(() => {
+  vi.clearAllMocks()
+  deleteDoc.mockResolvedValue(undefined)
+  sessionStorage.clear()
+  document.querySelectorAll('input[type=file]').forEach((input) => input.remove())
+})
+
+describe('runMarkdownImport backend flow', () => {
+  it('applies an oversized PM-shaped response without stashing content', async () => {
+    const events: string[] = []
+    createDoc.mockImplementation(async ({ title }: { title: string }) => {
+      events.push(`create:${title}`)
+      return { docId: 'd_new' }
+    })
+    importMarkdown.mockImplementation(async (docId: string, file: File) => {
+      events.push(`import:${docId}:${file.name}`)
+      return {
+        doc: { type: 'doc', content: Array.from({ length: 50_000 }, () => ({ type: 'paragraph' })) },
+        warnings: ['backend warning'],
+      }
+    })
+
+    const setItem = vi.spyOn(Storage.prototype, 'setItem')
+    const pending = runMarkdownImport('sp', 'folder')
+    choose(document.querySelector('input[type=file]')!, new File(['# Hello'], 'Hello.md', { type: 'text/markdown' }))
+    const result = await pending
+
+    expect(events).toEqual(['create:Hello', 'import:d_new:Hello.md'])
+    expect(result).toEqual({ docId: 'd_new', title: 'Hello', warnings: ['backend warning'] })
+    expect(consumeImportContent('d_new')).toBeNull()
+    expect(sessionStorage.getItem('octo-import-pm-d_new')).toBeNull()
+    expect(setItem.mock.calls.some(([key]) => key === 'octo-import-pm-d_new')).toBe(false)
+    expect(consumeImportWarnings('d_new')).toEqual(['backend warning'])
+  })
+
+  it('keeps concurrent picker file/title pairs race-safe', async () => {
+    createDoc.mockImplementation(async ({ title }: { title: string }) => ({ docId: `d_${title}` }))
+    importMarkdown.mockImplementation(async (docId: string) => ({
+      doc: { type: 'doc', content: [{ type: 'paragraph', attrs: { docId } }] }, warnings: [],
+    }))
+
+    const first = runMarkdownImport()
+    const second = runMarkdownImport()
+    const inputs = [...document.querySelectorAll<HTMLInputElement>('input[type=file]')]
+    choose(inputs[1]!, new File(['two'], 'Second.markdown'))
+    choose(inputs[0]!, new File(['one'], 'First.md'))
+
+    await expect(Promise.all([first, second])).resolves.toEqual([
+      { docId: 'd_First', title: 'First', warnings: [] },
+      { docId: 'd_Second', title: 'Second', warnings: [] },
+    ])
+    expect(importMarkdown.mock.calls.map(([docId, file]) => [docId, (file as File).name])).toEqual([
+      ['d_Second', 'Second.markdown'],
+      ['d_First', 'First.md'],
+    ])
+  })
+
+  it('propagates apply failure and does not make content available for navigation', async () => {
+    createDoc.mockResolvedValue({ docId: 'd_failed' })
+    importMarkdown.mockRejectedValue(new Error('yjs_storage_failed'))
+
+    const pending = runMarkdownImport()
+    choose(document.querySelector('input[type=file]')!, new File(['large body'], 'Failed.md'))
+
+    await expect(pending).rejects.toThrow('yjs_storage_failed')
+    expect(deleteDoc).toHaveBeenCalledOnce()
+    expect(deleteDoc).toHaveBeenCalledWith('d_failed')
+    expect(sessionStorage.getItem('octo-import-pm-d_failed')).toBeNull()
+    expect(sessionStorage.getItem('octo-import-warn-d_failed')).toBeNull()
+  })
+
+  it('applies a large DOCX without stashing returned PM content and returns its id', async () => {
+    createDoc.mockResolvedValue({ docId: 'd_docx' })
+    importDocx.mockResolvedValue({
+      doc: { type: 'doc', content: Array.from({ length: 50_000 }, () => ({ type: 'paragraph' })) },
+      warnings: [],
+    })
+
+    const setItem = vi.spyOn(Storage.prototype, 'setItem')
+    const pending = runDocxImport()
+    choose(document.querySelector('input[type=file]')!, new File([new Uint8Array(2_000_000)], 'Large.docx'))
+
+    await expect(pending).resolves.toEqual({ docId: 'd_docx', title: 'Large', warnings: [] })
+    expect(sessionStorage.getItem('octo-import-pm-d_docx')).toBeNull()
+    expect(setItem.mock.calls.some(([key]) => key === 'octo-import-pm-d_docx')).toBe(false)
+  })
+
+  it('propagates DOCX apply/storage failure', async () => {
+    createDoc.mockResolvedValue({ docId: 'd_docx_failed' })
+    importDocx.mockRejectedValue(new Error('collab_write_failed'))
+
+    const pending = runDocxImport()
+    choose(document.querySelector('input[type=file]')!, new File(['zip'], 'Failed.docx'))
+
+    await expect(pending).rejects.toThrow('collab_write_failed')
+    expect(deleteDoc).toHaveBeenCalledOnce()
+    expect(deleteDoc).toHaveBeenCalledWith('d_docx_failed')
+    expect(sessionStorage.getItem('octo-import-pm-d_docx_failed')).toBeNull()
+  })
+
+  it('preserves the import error when cleanup also fails', async () => {
+    createDoc.mockResolvedValue({ docId: 'd_cleanup_failed' })
+    importMarkdown.mockRejectedValue(new Error('import_failed'))
+    deleteDoc.mockRejectedValue(new Error('cleanup_failed'))
+
+    const pending = runMarkdownImport()
+    choose(document.querySelector('input[type=file]')!, new File(['body'], 'Failed.md'))
+
+    await expect(pending).rejects.toThrow('import_failed')
+    expect(deleteDoc).toHaveBeenCalledWith('d_cleanup_failed')
+  })
+
+})
--- a/packages/docs/src/editor/importFlow.ts
+++ b/packages/docs/src/editor/importFlow.ts
@@ -1,18 +1,12 @@
-// Import flow: file picker → parse Markdown → create new doc → navigate → inject content.
-//
-// Design doc §5 (方案 A1 + B1): user picks a .md file, we parse it to PM JSON,
-// create a new empty doc via REST, stash the PM JSON in sessionStorage keyed by docId,
-// then navigate to the new doc. EditorShell picks up the stashed content on mount and
-// injects it via setContent once the editor is ready.
+// Import flow: file picker → create destination → atomically apply on backend → navigate.
+// The destination editor always loads authoritative content from its live Y.Doc.
 
-import { parseMarkdownToPmDoc } from '../import/markdown.ts'
-import { createDoc, importDocx } from '../pages/docsApi.ts'
+import { createDoc, deleteDoc, importDocx, importMarkdown } from '../pages/docsApi.ts'
 import {
   copyAttachments,
   ingestAttachments,
   type CopySourceRef,
 } from '../attachments/api.ts'
-import { emojiGlyph } from './emoji.ts'
 
 const IMPORT_KEY_PREFIX = 'octo-import-pm-'
 const IMPORT_WARN_PREFIX = 'octo-import-warn-'
@@ -134,8 +128,8 @@ export interface ImportResult {
 }
 
 /**
- * Run the full import flow: pick file → parse → create doc → stash content.
- * Caller navigates to result.docId after this resolves.
+ * Run the full import flow: pick file → create doc → atomically apply on backend.
+ * Caller navigates to result.docId only after this resolves.
  */
 /** Translator type threaded from the calling React component (has useTranslation). */
 type Translate = (key: string, opts?: { values?: Record<string, unknown> }) => string
@@ -173,7 +167,7 @@ export async function runMarkdownImport(
   t?: Translate,
 ): Promise<ImportResult> {
   // 1. File picker
-  const { text, fileName } = await pickMdFile(t)
+  const { file, fileName } = await pickMdFile(t)
 
   // 1b. Extension guard. `input.accept` is only a UI hint (the OS dialog still lets the user pick
   // "all files"), so reject anything that is not a Markdown file here — this import path is for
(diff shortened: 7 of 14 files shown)


Jerry-Xin (dismissed):
Relevant to octo-web, but Markdown validation failures currently leave unwanted empty documents behind.

### 🔴 Blocking

- 🔴 Critical — Validate Markdown before creating the destination document. In [importFlow.ts](/Users/mlamp/.openclaw/tmp/pr-review-zphqtlfd/repo/packages/docs/src/editor/importFlow.ts:189), `createDoc` runs before `importMarkdown`, while empty, oversized, and invalid UTF-8 checks occur inside the latter. Any such rejection leaves an empty document in the user’s workspace, even though the UI reports that import failed. Move byte validation ahead of `createDoc`—for example, expose a validation/read helper and pass the validated bytes to the upload—or clean up the created document on failure. Add a flow test proving invalid input never creates a document.

### ✅ Highlights

- Export formats consistently use the authenticated unified backend endpoint.
- API tests cover binary configuration, URL encoding, format selection, and validation.
- Import ordering, concurrent picker isolation, filenames, MIME types, and font localization are well covered.

Jerry-Xin (dismissed):
Unifies document import/export onto authenticated backend endpoints; solid tests, but Markdown validation runs after the destination doc is created, orphaning empty docs on invalid input.

### 🔴 Blocking

- 🔴 Validate the Markdown upload **before** creating the destination doc. In `packages/docs/src/editor/importFlow.ts` (~L189–198), `createDoc(...)` runs first, then `importMarkdown(created.docId, file)`. But `importMarkdown` in `packages/docs/src/pages/docsApi.ts` throws `empty_upload` / `doc_too_large` / `invalid_utf8` before it ever POSTs. So any empty / oversized / non-UTF-8 file leaves a freshly-created empty document behind in the user's workspace while the UI reports the import as failed — a behavior regression vs. the old flow, which parsed (and could fail) before `createDoc`. Fix by hoisting the byte validation ahead of `createDoc` (e.g. a small validate-and-read helper that returns the checked bytes, passed into the upload), or clean up the created doc on failure. Please add a flow test proving invalid input never creates a doc.

### 💬 Non-blocking

- `exportDownloadName` in `packages/docs/src/editor/EditorShell.tsx` (L99) only trims/falls back — it does not strip path separators or control chars from the title before it becomes `a.download`. Pre-existing and unchanged by this PR, and the anchor `download` attribute is not an HTTP header (browsers neutralize `/`), so low risk — but worth a follow-up hardening pass now that all three formats route through it.

### ✅ Highlights

- Export is unified behind one authenticated endpoint (`exportDocFile(docId, format)`); `docId` is `encodeURIComponent`'d in both export and import paths (correct URL-injection handling).
- Import now defers to the authoritative backend parser (no in-browser parse); stashed payload still passes through schema-validated `setContent`, and image migration reuses the SSRF-guarded `ingestAttachments` / permission-scoped `copyAttachments` paths.
- Strong test coverage: binary `arraybuffer` config, URL encoding, per-format selection, client-side size/UTF-8 validation, concurrent-picker isolation, and filename/MIME assertions.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1046 (octo-web)

## Summary

This PR unifies document import and export under backend-authoritative endpoints. Markdown import shifts from client-side `parseMarkdownToPmDoc` to `POST /docs/{docId}/import/markdown` (same PM schema as DOCX), and all three export formats (md/docx/pdf) consolidate into a single `exportDocFile(docId, format)` call. The old client-side Markdown parser, lazy-loaded DOCX builder, and per-format export handlers are removed. Calibri font is added to both the rich-text editor and spreadsheet.

The refactor is well-structured: proper client-side UTF-8 and byte-cap pre-validation, consistent error propagation, and good test coverage at both the API layer and the flow-orchestration layer. No blocking issues found.

## Verification

- ✅ **Byte-cap enforcement** — `importMarkdown` (`packages/docs/src/pages/docsApi.ts:376-381`) checks both `file.size` and `bytes.byteLength` against `MAX_MARKDOWN_IMPORT_BYTES` (25 MiB); strict UTF-8 via `TextDecoder('utf-8', { fatal: true })`. Test at `docsApi.import.test.ts:44-51` covers empty, oversized, and invalid UTF-8 rejection before any network call.
- ✅ **Data-flow: Markdown import** — `pickMdFile` returns raw `File` → `createDoc` creates destination → `importMarkdown(created.docId, file)` sends bytes to backend → returns `{doc, warnings}` → `migrateImportedImages` re-scopes attachments → `stashImportContent` + `stashImportWarnings` persist to sessionStorage → `EditorShell` drains via `consumeImportContent`/`consumeImportWarnings` and injects via `setContent`. Every step verified in `importFlow.markdown.test.ts`.
- ✅ **Data-flow: Export** — `downloadExport(format)` calls `exportDocFile(docId, format)` → `GET /docs/{encodedDocId}/export/file?format={format}` with `responseType: 'arraybuffer'` and 120s timeout → blob download with deferred URL revoke. Three format-specific wrappers (`onExportMarkdown`, `onExportDocx`, `onExportPdf`) are thin `void` delegates; internal error handling in `downloadExport` catches failures and sets `exportError` state.
- ✅ **Concurrency safety** — `pickMdFile` resolves each file picker independently (no module-level mutable state), verified by the concurrent-import test at `importFlow.markdown.test.ts:52-67`.
- ✅ **Public-contract parity** — `importMarkdown` test confirms URL encoding (`d /1` → `d%20%2F1`), correct content-type header, and backend error propagation with status/body preserved.
- ✅ **i18n** — Calibri label added to both `en-US.json` and `zh-CN.json`; font value `'Calibri, Carlito, sans-serif'` matches the `fontFamilies.verify.test.ts` assertion.

Static analysis only at head `5098a39b`; build and tests not executed in this environment.

## Findings

### Nit — `importDocx` URL-encoding inconsistency (`packages/docs/src/pages/docsApi.ts:390`)

`importDocx` passes `docId` unencoded in its URL path (`/docs/${docId}/import/docx`), while the new `importMarkdown` and `exportDocFile` both correctly use `encodeURIComponent(docId)`. Pre-existing — not a regression from this PR — but worth aligning: a docId containing `/`, `#`, or `?` would break DOCX import routing.

```ts
// Current (pre-existing):
`/docs/${docId}/import/docx`
// Aligned with importMarkdown / exportDocFile:
`/docs/${encodeURIComponent(docId)}/import/docx`
```

## Data-Flow Tracing

- **Markdown import consumed data**: `file` (from `pickMdFile` → `input.files[0]`) → flows to `importMarkdown` as `Pick<File, 'size'|'arrayBuffer'>` → backend returns `{doc, warnings}` → `doc` is a PM JSON that flows into `migrateImportedImages` (image nodes collected and re-scoped) → stashed in sessionStorage → consumed by `EditorShell` via `consumeImportContent` → injected via `editor.commands.setContent`. No empty-array / undefined / early-return traps found in the chain.
- **Export consumed data**: `docId` + `format` → `exportDocFile` → `GET` returns `ArrayBuffer` → wrapped in `Blob` with format-appropriate MIME type → downloaded via ephemeral `<a>` element. No stale editor snapshot dependency (backend reads live Y.Doc).
- **Warning aggregation**: Backend returns `warnings[]`, then `migrateImportedImages` appends `imageWarnings` via `warnings.push(...imageWarnings)`. The array is stashed and consumed as-is by `EditorShell`. No lossy merge.

- **C1 — Double-path parity**: ✅ Clear. Export: three format handlers share one `downloadExport` with identical error/guard/cleanup logic. Import: Markdown and DOCX paths both use `createDoc` → backend import → stash pattern; no asymmetry introduced.
- **C2 — Control-flow ordering**: ✅ Clear. `downloadExport` has `if (exporting) return` re-entry guard; `exporting` state set before async call; `finally` resets. No nested/re-entrant call sites found.
- **C3 — Authorization boundary**: ✅ N/A. All routes go through `apiClient()` which inherits the global auth interceptor. Backend enforces role-based access.
- **C4 — Authorization lifecycle**: ✅ N/A. No auth changes in this PR.
- **C5 — Build/runtime parity**: ✅ Clear. `encodeURIComponent` verified by test (`d /1` → `d%20%2F1`). MIME types match format. 120s timeout set on both import and export.
- **C6 — Governance/docs**: ✅ N/A. No governance changes.

N/A — first review round.

## Verdict: APPROVED

No P0/P1 issues. The refactor correctly unifies import/export under backend endpoints with proper client-side validation, error propagation, and test coverage. One pre-existing nit on `importDocx` URL encoding is noted but non-blocking.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. Well-structured refactor with proper data-flow integrity, byte-cap enforcement, and comprehensive test coverage at both API and orchestration layers.

wwwzzzcccc:
Fixed mandatory review item 2 in 674c99c6.

- Markdown and DOCX imports now send `X-Octo-Import-Apply: true` and accept responses without `doc`.
- Active import flows no longer put ProseMirror content in sessionStorage; they return only after the backend has atomically written the live Y.Doc, then navigation loads authoritative collaboration state.
- Backend/network/storage failures propagate, so the UI cannot navigate to a silently empty document.
- Warning handoff remains small and independent; legacy PM consume/stash compatibility is retained but unused by active paths.
- Added oversized PM-shaped response / large DOCX coverage asserting no content `sessionStorage.setItem`, apply headers, returned document IDs, and failure propagation.

Validation:
- focused Vitest: 3 files, 22 tests passed
- i18n scan: 0 candidates, 0 locale key issues
- `git diff --check`: clean
- typecheck reached two existing unrelated baseline errors in `dmworkbase/src/i18n/format.ts` and `docs/src/members/sort.ts`.

Jerry-Xin (dismissed):
Re-review at head `674c99c6`. The import path was re-architected to an atomic backend apply (new `X-Octo-Import-Apply: true` header; client no longer stashes the parsed tree). This is a clean direction, but both prior blockers remain open at this head.

### 🔴 Blocking

- 🔴 `packages/docs/src/pages/docsApi.ts:417` — `importDocx` still interpolates a raw, unencoded `docId` into the request path: `` `/docs/${docId}/import/docx` ``. Compare `importMarkdown` (`docsApi.ts:390`) and `exportDocFile` (`docsApi.ts:353`), both of which correctly use `encodeURIComponent(docId)`. An id containing `/`, `?`, `#`, `..`, or spaces breaks path resolution and is a path-injection/traversal risk (#1008 class). The test at `packages/docs/src/pages/docsApi.import.test.ts` even locks in the broken behavior: it asserts `call.url` equals `/docs/doc / 2/import/docx` (unencoded space + slash), while the sibling markdown test correctly asserts `/docs/d%20%2F1/import/markdown`. Fix: wrap with `encodeURIComponent(docId)` and update the docx test to expect the encoded path. (This was already unencoded at the prior head; it was previously missed and is confirmed here — credit to the peer reviewer who flagged it.)

- 🔴 `packages/docs/src/editor/importFlow.ts:185` (markdown) and `:398` (docx) — `createDoc` still runs BEFORE the import call, and there is no cleanup on failure. When `importMarkdown` / `importDocx` reject (empty / oversized / invalid-UTF-8 / parse / storage / collab-write failure), the already-created empty doc is never deleted, orphaning an empty document in the user's workspace. The UI catch blocks in `packages/docs/src/pages/DocsHome.tsx:795` (`onImportMarkdown`) and `:822` (`onImportWord`) only surface an error message; neither deletes `created.docId`. The new failure tests in `packages/docs/src/editor/importFlow.markdown.test.ts` ("propagates apply failure...") only assert that `sessionStorage` is not written — they do not assert the created doc is cleaned up, so the orphan is untested and unfixed. Fix: delete the created doc on import failure (try/catch around the import call that calls the doc-delete API), and add a test asserting a failed import leaves no doc behind.

### 💬 Non-blocking

- `packages/docs/src/editor/EditorShell.tsx` — `exportDownloadName` still does not strip path separators / control chars from the title before using it as an anchor download filename (pre-existing, low risk). Not a blocker.

### ✅ Highlights

- Atomic-apply re-architecture removes the large client-side PM stash across navigation; the editor now loads authoritative content from the live collaboration state — simpler and less error-prone.
- `DocumentImportResult.doc` correctly made optional to match responses that omit the parsed tree.
- Good coverage of the large-response / no-stash and apply-failure-propagation paths, plus the `X-Octo-Import-Apply` header assertions.
- `EditorShell` warning surfacing correctly decoupled from the (now-optional) PM injection path.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1046 (octo-web)

## Summary

This PR unifies document import and export behind authoritative backend endpoints. Markdown import moves from client-side parsing (via `parseMarkdownToPmDoc`) + sessionStorage stash to a server-side parser with atomic Y.Doc apply. All three export formats (Markdown, DOCX, PDF) now flow through a single `exportDocFile` endpoint instead of client-side serialization or separate backend routes. The Calibri font is added as a minor sidecar.

The architecture is sound — the backend becomes the single source of truth for both parsing imports and rendering exports from the live Y.Doc, eliminating the client-side PM stash race and stale-editor snapshot problem. Tests are thorough, covering byte-cap enforcement, UTF-8 validation, concurrent picker race safety, and error propagation. Two P2 items below.

## Verification

- ✅ **Import flow atomicity** — `runMarkdownImport` and `runDocxImport` both create the destination doc first, then call the backend with `X-Octo-Import-Apply: true`. No PM content is stashed; EditorShell loads from the live Y.Doc. The `consumeImportContent` path in `EditorShell.tsx:339` correctly returns null for atomic imports and the `if (!pmDoc) return` guard handles it.
- ✅ **Byte-cap enforcement** — `importMarkdown` at `packages/docs/src/pages/docsApi.ts:377-387` double-checks both `file.size` and `bytes.byteLength` against `MAX_MARKDOWN_IMPORT_BYTES` (25 MiB), with strict UTF-8 validation via `TextDecoder({ fatal: true })`. Tests at `docsApi.import.test.ts:44-48` cover empty, oversized, and invalid-UTF-8 rejection.
- ✅ **Export unification** — `downloadExport` at `EditorShell.tsx:481-497` correctly routes all three formats through `exportDocFile`, with proper MIME types and blob URL cleanup. The `instance` dependency was correctly removed since the backend reads the authoritative Y.Doc.
- ✅ **Warning surface ordering** — EditorShell now consumes import warnings before checking for PM content (`EditorShell.tsx:349-351`), correctly handling the atomic-import case where warnings exist but no PM stash does.
- ✅ **i18n completeness** — Calibri font label present in both `en-US.json` and `zh-CN.json`; export submenu label `docs.toolbar.export` already existed.

Static analysis only at head `674c99c6`; build and tests not executed in this environment.

## Findings

### P2 — `importDocx` URL lacks `encodeURIComponent(docId)` (`packages/docs/src/pages/docsApi.ts:414`)

The new `importMarkdown` function at `:390` and `exportDocFile` at `:353` both apply `encodeURIComponent(docId)`, but `importDocx` — which was also updated in this PR (adding the `X-Octo-Import-Apply` header) — still uses raw `${docId}` in the URL template. The test at `docsApi.import.test.ts:51` explicitly asserts the unencoded path `/docs/doc / 2/import/docx`, while the Markdown test at `:35` expects `d%20%2F1`. If a docId contains `/` or other URL-unsafe characters, the DOCX path would produce a malformed route segment.

```ts
// Fix: align with importMarkdown
`/docs/${encodeURIComponent(docId)}/import/docx`
```

### P2 — Dead exports after atomic-import refactor (`packages/docs/src/editor/importFlow.ts:15`)

`stashImportContent` (`:15`), `consumeImportContent` (still referenced in `EditorShell.tsx:339`), and `migrateImportedImages` (`:218`) are no longer called from any production import path. `consumeImportContent` will always return null since nothing writes to the `octo-import-pm-` sessionStorage key anymore. The functions remain exported and tested, which may mislead future contributors into thinking the stash-based flow is still active. Consider removing them or marking `@deprecated` to make the atomic backend path the only documented import contract.

- Export format enum `'md' | 'docx' | 'pdf'` is TypeScript-constrained — no query-string injection risk in `exportDocFile`.
- `exportDocPdf` deprecated wrapper correctly delegates to `exportDocFile(docId, 'pdf')`; no remaining production callers.
- EditorShell `downloadExport` correctly removed the `instance` dependency — the callback no longer captures the editor reference.
- `onExportMarkdown`/`onExportDocx`/`onExportPdf` wrappers use `() => void downloadExport(format)` — void discards the promise correctly for fire-and-forget menu handlers.
- Concurrent Markdown import test at `importFlow.markdown.test.ts:59-78` proves race safety with separate file picker inputs.
- DOCX import test correctly asserts no sessionStorage writes for PM content (`importFlow.markdown.test.ts:96-97`).
- Calibri font addition is consistent across `fontFamilies.ts`, `CollabSheet.ts`, and both i18n files with matching test coverage.

## Verdict: COMMENT

Well-executed architectural unification. The backend-as-authoritative-source pattern eliminates the client-side PM stash race and stale-snapshot export problem cleanly. The two P2 items — URL encoding inconsistency in `importDocx` and dead stash exports — are non-blocking cleanup.




[Octo-Q] verdict: APPROVE — No P1 blockers found. Two P2 findings (URL encoding inconsistency in importDocx; dead stash exports after refactor) are non-blocking cleanup items.

wwwzzzcccc:
Resolved both current blocking items in `a32cbd48`.

- DOCX import now percent-encodes `docId`, with the request-path regression test updated.
- Markdown and DOCX apply failures now best-effort delete the newly-created destination before rethrowing the original import error, including coverage for cleanup failure preserving that original error.

Validation:
- focused Vitest: 3 files, 34 tests passed
- i18n scan/check: clean
- `git diff --check`: clean
- package typecheck still reaches only the two existing unrelated baseline errors in `dmworkbase/src/i18n/format.ts` and `docs/src/members/sort.ts`
- broad docs run: 153 files / 1655 tests passed; one unrelated `TableReorderInterrupt` positive-control failure plus existing jsdom/WebSocket warnings

Jerry-Xin (dismissed):
Re-review at a32cbd48f4. Both prior blockers are now fixed byte-level with tests, and no regressions to the unified import/export surface. Approving.

## ✅ Prior blockers resolved

- 🔴→✅ docx import path now encodes the id: `packages/docs/src/pages/docsApi.ts:417` uses `encodeURIComponent(docId)`, matching the markdown (`:390`) and export (`:353`) siblings. The frozen test was corrected: `packages/docs/src/pages/docsApi.import.test.ts:67` now asserts the encoded `/docs/doc%20%2F%202/import/docx` instead of the raw space/slash. No remaining unencoded id-in-path on any import/export endpoint.
- 🔴→✅ Orphan-doc-on-import-failure is cleaned up: `packages/docs/src/editor/importFlow.ts:189-198` (markdown) and `:409-416` (docx) wrap the import call in try/catch that best-effort deletes the just-created doc (`deleteDoc(created.docId).catch(() => undefined)`) and re-throws the original importer error. Tests cover it: `packages/docs/src/editor/importFlow.markdown.test.ts` now asserts `deleteDoc` is called once with the created id on `yjs_storage_failed` and `collab_write_failed`, plus a case proving the import error is preserved even when cleanup also fails.

## 💬 Non-blocking

- Export reads the authoritative live Y.Doc server-side (`packages/docs/src/editor/EditorShell.tsx:481`) and intentionally does not serialize the local snapshot. Because collab sync (WebSocket) and the export request (HTTP) are independent, a very recent local edit may not yet be reflected in the exported copy. This is an inherent eventual-consistency property of any live-doc reader, not introduced by this PR — the new backend path is strictly better than the old client-side serialize. Optional hardening: gate the export button on provider sync state (e.g. `hasUnsyncedChanges`) so users cannot export mid-sync. Consider a small regression test for that path.
- Obsolete import-stash comments around `packages/docs/src/editor/EditorShell.tsx:333` still describe client-side parsing/content injection; worth trimming to match the backend-apply flow.
- Pre-existing, low risk: `exportDownloadName` (`packages/docs/src/editor/EditorShell.tsx:99`) does not strip path separators/control chars from the title; the anchor `download` attribute makes this low impact.

## ✅ Highlights

- Document ids are consistently encoded across all import/export endpoints.
- Import failures no longer leave an empty destination doc; cleanup preserves the original error.
- The parsed document tree no longer crosses navigation through size-limited sessionStorage.
- Good focused tests added for ordering, cleanup, warnings, filenames, and the encoded path.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1046 (octo-web)

## Summary

This PR unifies document import and export behind authoritative backend endpoints. Markdown import moves from client-side parsing (`parseMarkdownToPmDoc` + sessionStorage stash) to a backend atomic apply (`importMarkdown` with `X-Octo-Import-Apply` header), and all three export formats (md/docx/pdf) consolidate into a single `exportDocFile` endpoint that reads the live Y.Doc. The editor no longer serializes its local snapshot for export or injects stashed PM content for import — it trusts the backend's authoritative state.

The architectural direction is sound: eliminating client-side parsing divergence, session-storage fragility, and stale-snapshot export. Client-side byte-cap and UTF-8 validation at `docsApi.ts:377-387` provide defense-in-depth before the POST. Cleanup-on-failure via `deleteDoc` in both import flows is correct. Three non-blocking items below.

## Verification

- ✅ **Data flow: import warnings** — `runMarkdownImport` stashes only `warnings` via `stashImportWarnings`; `EditorShell.tsx:349-350` consumes them independently of `pmDoc`. When no stash exists (atomic import), `consumeImportWarnings` returns `[]` and `setImportNotice` is not called. Correct.
- ✅ **Data flow: import content** — Neither `runMarkdownImport` nor `runDocxImport` calls `stashImportContent`. `consumeImportContent` at `EditorShell.tsx:339` returns `null`, the `if (!pmDoc) return` at `:352` exits cleanly. Editor loads content from Y.Doc via collaboration. Correct.
- ✅ **Cleanup on failure** — Both import flows catch `importMarkdown`/`importDocx` rejection, call `deleteDoc(created.docId).catch(() => undefined)`, then re-throw. Warnings are never stashed on the failure path (stash is after the try/catch). The original error is preserved even when cleanup also fails. Correct.
- ✅ **Export unification** — `downloadExport` at `EditorShell.tsx:481-496` dispatches all three formats through `exportDocFile` with correct MIME types. The `exportDocPdf` deprecated wrapper at `docsApi.ts:364` delegates correctly. No `instance` dependency in the export path — appropriate since the backend reads the live Y.Doc.
- ✅ **URL encoding** — `importMarkdown`, `importDocx`, and `exportDocFile` all use `encodeURIComponent(docId)`. Consistent within the new code.
- ✅ **Byte-cap validation** — `importMarkdown` checks both `file.size` and `bytes.byteLength` against `MAX_MARKDOWN_IMPORT_BYTES` (25 MiB), plus empty and UTF-8 validity. Four-guard defense-in-depth. Correct.
- ✅ **Calibri font addition** — Consistent across `fontFamilies.ts:26`, `CollabSheet.ts:335`, both i18n locales, and the verify test.
- ✅ **Test coverage** — New tests cover: API-level byte-cap/UTF-8 rejection (`docsApi.import.test.ts`), end-to-end import flow with mock backends including failure/cleanup (`importFlow.markdown.test.ts`), unified export endpoint (`docsApi.export.test.ts`), and EditorShell export mock wiring.

Static analysis only at head `a32cbd48`; build and tests not executed in this environment.

## Findings

### P2 — Dead export: `stashImportContent` no longer called (`packages/docs/src/editor/importFlow.ts:15`)

`stashImportContent` is exported but no production caller remains — both `runMarkdownImport` and `runDocxImport` removed their stash calls. `consumeImportContent` at `EditorShell.tsx:339` still reads the key (returning `null` for new imports), so there is no runtime harm, but the writer side, `IMPORT_KEY_PREFIX`, and `isValidPmDoc`/`isPlainNode` validators are dead exports. Consider removing in a follow-up cleanup, or marking `@deprecated` if a legacy client path may still call them.

### P2 — Dead export: `migrateImportedImages` and helpers (`packages/docs/src/editor/importFlow.ts:225`)

`migrateImportedImages` (~150 lines) plus helpers `collectImageNodes`, `parseServiceImageRef`, and `imageNodeToLink` are no longer called from either import flow. The backend now owns image re-hosting during its atomic import. These survive only through `importFlow.images.test.ts`. If the backend contract fully replaces client-side image migration, this block is dead code and can be removed in a follow-up.

### P2 — `deleteDoc` missing `encodeURIComponent` parity (`packages/docs/src/pages/docsApi.ts:345`)

`deleteDoc` does not encode `docId` in its URL path, while the new `importMarkdown` (`:396`), `importDocx` (`:418`), and `exportDocFile` (`:355`) all do. Since `deleteDoc` is the cleanup path called by both import flows on failure, a doc with special characters in its id would be created and imported with proper encoding but deleted without it — the delete request may 404, leaving an orphan empty doc. Pre-existing issue amplified by the encoding parity this PR introduces elsewhere.

```ts
// Fix sketch — add encodeURIComponent to match the other endpoints:
await apiClient().delete(`/docs/${encodeURIComponent(docId)}`)
```

- Concurrent import race safety: each `pickMdFile`/`pickDocxFile` creates its own `<input>` element and resolves with its own `file`/`fileName` pair; no module-level mutable state. Test at `importFlow.markdown.test.ts:64` confirms.
- Export callback dependency arrays: `downloadExport` depends on `[docId, currentTitle, exporting]` (not `instance`), appropriate since the backend owns content. The `void downloadExport('md')` wrappers correctly fire-and-forget.
- EditorShell warning/content ordering: warnings are consumed before the `if (!pmDoc) return` gate, so atomic imports surface their diagnostics even without a PM stash.
- `exportDocPdf` deprecated wrapper correctly delegates to `exportDocFile(docId, 'pdf')`.
- `DocxImportResult` aliased to `DocumentImportResult` with `doc` now optional — backward-compatible.

## Verdict: APPROVED

Clean architectural migration that correctly moves import parsing and export serialization to the authoritative backend. No P0/P1 blockers. The three P2 items are non-blocking cleanup observations: two dead-export removals and one `encodeURIComponent` parity fix in the pre-existing `deleteDoc`.

[Octo-Q] verdict: APPROVE — no P0/P1 blockers found; three P2 cleanup items (dead exports, deleteDoc encoding parity) are non-blocking.

cnwenf (changes requested):
# Code Review — PR #1046 (Mininglamp-OSS/octo-web)

## Summary

This PR routes Markdown import and MD/DOCX/PDF export through the backend's authoritative document pipelines (`importMarkdown`, `exportDocFile`), drops the client-side serialization modules, and adds Calibri to the Doc/Sheet font catalogs. All three independent reviews agree the architecture is sound: request shapes are correct and mutation-pinned, client-side validation (empty check, 25 MB cap, strict UTF-8) is defense-in-depth, and security is cleared with zero exploitable findings. All three reviewers independently flag the same two issues — a failed Markdown import orphans an empty doc (no `deleteDoc` rollback after `createDoc`), and CI never runs the new vitest suites. glm5.2 treats the orphan-doc regression as a blocking must-fix (CHANGES_REQUESTED); qwen3.7 and qwen3.8 consider every finding non-blocking (COMMENT) — hence the consolidated NEEDS_CHANGES below.

Note: all three reviews were conducted against head `5098a39bbfca0025057bc6c016b4dd7af5e1bf44`; newer commits (`a32cbd48f49984750c0c421e2ba0f9c3bb4ee294`) landed after the reviews completed, so some points below may already be addressed.

## Findings

### P2 — Failed Markdown import orphans an empty doc, and the failure branch is untested (`packages/docs/src/editor/importFlow.ts:189`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`runMarkdownImport` now creates the destination doc (`createDoc`, `:189`) BEFORE the authoritative backend import (`importMarkdown`, `:193`). Any `importMarkdown` failure — the new client-side rejections `empty_upload` / `doc_too_large` / `invalid_utf8` (`docsApi.ts:378`) or a backend 422 / network error / 120s timeout — leaves a permanent empty orphan doc titled after the file: `runMarkdownImport` has no try/catch/finally and never calls `deleteDoc` (`docsApi.ts:344`), and `DocsHome.onImportMarkdown`'s catch (`DocsHome.tsx:803`) only sets a banner. Pre-PR, Markdown parsed client-side BEFORE `createDoc`, so a failed parse never created a doc; this diff introduces the orphan window for Markdown (the DOCX path at `:410`→`:413` already had it). Confirmed by probe at exact head: on rejection, `createDoc` is called once, `deleteDoc` never, and the doc stays alive and empty. The branch is also unpinned — `importFlow.markdown.test.ts` never makes `importMarkdown` reject, and a swallow-mutation (failed import reported as success) leaves that suite green while a focused probe fails. Fix: run the client-side size/UTF-8 checks before `createDoc` and/or best-effort `deleteDoc(created.docId)` in a catch/finally (apply to both md and docx), and add a rejection-after-`createDoc` test asserting rollback and propagation.

(severity split: glm5.2=P2 "blocking must-fix introduced by this PR's reordering", qwen3.8=P2, qwen3.7=Nit — raised from the test-coverage angle; all three confirmed `deleteDoc` is never invoked on failure via mutation/focused probes)

### P2 — CI never runs the new vitest suites (`.github/workflows/ci.yml:88`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The `build` job runs `pnpm install`, `i18n:check`, `build`, `lint`, `lint:wkmodal` (`:88-96`) but never `pnpm test` / `turbo run test` / `vitest`. Root `package.json` defines `"test": "turbo run test"` and `packages/docs/package.json` defines `"test": "vitest run"`, yet no workflow invokes them — so the five new test files never gate merging. This is a pre-existing gap this PR makes material by adding the tests that should be gating (the PR body claims "23 tests passed", but those won't run in CI). Add a `pnpm test` step to the `build` job.

(qwen3.8 additionally confirmed this by a structured parse of all 20 workflows: 0 unit-test steps; `e2e.yml`/`e2e-nightly.yml` run only Playwright against `apps/web`, and turbo `build` has no `dependsOn:['test']`.)

### P2 — Backend response array mutated in-place (`packages/docs/src/editor/importFlow.ts:199`) [source: octo-review-qwen3.7]

`runMarkdownImport` mutates the `warnings` array from `importMarkdown` via `warnings.push(...imageWarnings)`. The sibling `runDocxImport` at `:413` passes warnings straight through without mutation. If the API response is ever cached or frozen, this will either throw or leak warnings into shared state. Mutation testing confirms: frozen response throws `TypeError`; reference-identity test proves `result.warnings === apiWarnings`.

```ts
const allWarnings = [...warnings, ...imageWarnings]
```

(octo-review-qwen3.8 assessed this as a style nit only: concurrent imports keep independent per-response warning arrays, so the in-place push corrupts no shared/frozen source today — the risk materializes only if the API layer later caches or freezes responses.)

### P2 — Orphan export modules not removed (`packages/docs/src/editor/EditorShell.tsx:1`) [source: octo-review-qwen3.7]

This PR switched `EditorShell` to backend-driven export, but 26 files under `packages/docs/src/export/` remain with zero production imports. They can diverge from real export behavior and add to the bundle if tree-shaking misses them. Remove them or file a follow-up issue referenced in a comment.

### P3 — `importMarkdown` throws raw English codes, not i18n keys (`packages/docs/src/pages/docsApi.ts:378`) [source: octo-review-glm5.2 + octo-review-qwen3.8]

Pre-flight throws raw `'empty_upload'`/`'doc_too_large'`/`'invalid_utf8'` (`:378-386`). `DocsHome`'s catch regexes (`/mdOnly|Markdown|\.md/i`, `/取消|cancel/i`) don't match these, so they fall to the generic `docs.import.mdFailed` bucket, losing specific operator feedback; the old flow surfaced localized errors via `tr(t, ...)`. Also, an empty `.md` the old flow imported successfully is now hard-rejected (`file.size === 0`), changing the accepted-input contract. Throw i18n keys (or map in the caller) and decide deliberately on the empty-file case.

(severity split: glm5.2=P3, qwen3.8=Nit. qwen3.8 notes the sibling `onImportWord` (`:836`) already maps 413→`wordTooLarge`, `empty_upload`→`wordEmpty`, `import_unsafe`→reason — mirror that pattern.)

### P3 — post-arrayBuffer byteLength re-checks unexercised (`packages/docs/src/pages/docsApi.ts:381`) [source: octo-review-glm5.2 + octo-review-qwen3.7]

The guards at `:381-382` (`if (bytes.byteLength === 0)`, `if (bytes.byteLength > MAX)`) exist for when `file.size` and actual `arrayBuffer().byteLength` disagree, but no test provides that input class. Mutation check: removing `:381-382` leaves existing tests green (3/3). Add a test where `file.size` is valid but `arrayBuffer()` returns 0 / oversized bytes.

(severity split: glm5.2=P3, qwen3.7=Nit)

### P3 — Stale JSDoc: "parses client-side" is now false (`packages/docs/src/editor/importFlow.ts:193`) [source: octo-review-glm5.2 + octo-review-qwen3.8]

The JSDoc above `runDocxImport` reads "Unlike the Markdown flow (which parses client-side)…". This PR removed client-side `parseMarkdownToPmDoc` and replaced it with backend `importMarkdown` (`:193`), so the comment is now false — contract-text rot that misleads future implementers. Update the comment to reflect server-side parsing.

(octo-review-qwen3.8 flagged the same rot at two more locations: `importFlow.ts:392` and `DocsHome.tsx:793`, which still say Markdown parses client-side and that create-first is DOCX-only.)

### P3 — `importDocx` does not URL-encode docId (`packages/docs/src/pages/docsApi.ts:410`) [source: octo-review-glm5.2]

`importDocx` builds `/docs/${docId}/import/docx` with no `encodeURIComponent`, while the PR's new `importMarkdown` (`:389`) and `exportDocFile` (`:353`) both encode. The import test proves docIds can contain spaces/slashes (`'d /1' → 'd%20%2F1'`); a docx import for such a docId produces a malformed path / 404. Same module, two of three siblings fixed. Use `encodeURIComponent(docId)`.

(octo-review-qwen3.7 and octo-review-qwen3.8 assessed this as pre-existing — `importDocx` is unchanged from base, no new callers, server-generated docIds — and not widened by this diff; both recommend it as a non-blocking consistency follow-up.)

### P3 — Calibri added inconsistently between doc editor and sheet (`packages/docs/src/sheet/CollabSheet.ts:335`) [source: octo-review-glm5.2]

Docs `fontFamilies.ts` uses `'Calibri, Carlito, sans-serif'` with i18n labelKey `docs.toolbar.font.calibri`; the sheet uses bare `'Calibri'` (no Carlito fallback) and a hardcoded `'Calibri'` label while sibling rows use `t('docs.sheet.fontLabels.*')`. On Linux (no Calibri) the sheet falls back to the browser default while docs falls back to Carlito — same named font, different rendered result. Match the value (add Carlito fallback) and use an i18n labelKey (add the key to both locales), or share one constant.

### P3 — `@deprecated exportDocPdf` wrapper is not wire-equivalent (`packages/docs/src/pages/docsApi.ts:360`) [source: octo-review-glm5.2]

The old impl POSTed `/docs/{docId}/export/pdf` (no encoding); the wrapper delegates to `exportDocFile`, which GETs `/docs/{encodeURIComponent(docId)}/export/file?format=pdf`. The `@deprecated` comment implies a drop-in, but method (POST→GET), path, and encoding all changed. `grep` shows zero non-definition callers, so runtime impact is low, but the "behaviorally equivalent wrapper" contract is false. Either preserve the original route or correct the comment to state the wire contract changed.

### Nit — Export URL encoding isn't pinned by a test (`packages/docs/src/pages/docsApi.export.test.ts:16`) [source: octo-review-qwen3.8]

The export suite only exercises the literal docId `d_1`, so deleting `encodeURIComponent` from the `exportDocFile` URL (`docsApi.ts:353`) leaves it green — confirmed by mutation. The sibling import suite pins exactly this for `importMarkdown` (`d /1` → `/docs/d%20%2F1/import/markdown`). Add one `exportDocFile('d /1', 'md')` case asserting `/docs/d%20%2F1/export/file?format=md`.

### Nit — Export menu format mapping isn't pinned for DOCX/PDF (`packages/docs/src/editor/EditorShell.test.tsx:142`) [source: octo-review-qwen3.8]

Only the Markdown export row is clicked (asserts `('d_1','md')`); the DOCX and PDF rows (`EditorShell.tsx:603`→`onExportDocx`/`onExportPdf`→`downloadExport('docx'|'pdf')`) and their MIME / `.docx`/`.pdf` filename branches (`:483`) are never driven end-to-end. `docsApi.export.test.ts` calls `exportDocFile` directly and can't see the menu→format mapping, so a cross-wire leaves every suite green — confirmed by mutation (cross-wiring `onExportDocx`→`'pdf'` keeps both suites green while a focused two-row probe fails). Add `it.each`-style clicks of the exportDocx/exportPdf rows asserting `('d_1','docx')` / `('d_1','pdf')` plus the `Doc.docx` / `Doc.pdf` anchor names.

### Nit — additional unasserted branches (`packages/docs/src/editor/EditorShell.tsx:494`) [source: octo-review-qwen3.7]

`EditorShell.test.tsx` stubs `URL.revokeObjectURL` as a no-op and never advances fake timers past the 5000ms revoke `setTimeout`, so the blob-URL revoke is unasserted; only the `'md'` export format is clicked, so docx/pdf MIME selection and the `imageWarnings` merge (`importFlow.ts:199`) are unguarded. Worth adding clicks for docx/pdf and a non-empty image-warning case.

### Human-verify (non-blocking; cross-repo backend not runnable in review workspaces)

- **Backend wire contract + cache headers** — confirm octo-docs-backend serves `GET /docs/{id}/export/file?format={md|docx|pdf}` (file bytes) and `POST /docs/{id}/import/markdown` (`{doc,warnings}`) with the asserted shapes, and that the user-specific GET export emits `Cache-Control: no-store` + `Vary: Authorization` so intermediary proxies/CDNs don't cache binary downloads. [source: octo-review-qwen3.8 + octo-review-glm5.2]
- **Content persistence across navigation/crash** — if `importMarkdown` is parse-only (returns PM JSON without persisting to the Y.Doc), imported content exists only in the sessionStorage stash between create+import+migrate and `setContent`; closing the tab in that window leaves a permanently empty orphan doc. [source: octo-review-glm5.2]
- **Export sync-flush race** — `downloadExport` (`EditorShell.tsx:481`) GETs the backend copy with no provider flush/await, so just-typed unsynced edits could be missing from MD/DOCX exports. [source: octo-review-qwen3.8]
- **Image-migration idempotency** — if the backend `import/markdown` endpoint also re-hosts images, client-side `migrateImportedImages` (`importFlow.ts:198`) would run a second time on already-migrated attachIds, creating duplicates. [source: octo-review-glm5.2]
- **Image URL classification** — `parseServiceImageRef` expects `/file/<docId>/att_<id>/`; if the backend markdown parser normalizes/signs/rewrites image src URLs, service images could be misclassified as external. [source: octo-review-glm5.2]

- Wire contracts: `exportDocFile` GET with `?format=md|docx|pdf`, arraybuffer, per-format MIME, 120 s timeout; `importMarkdown` bounded strict-UTF-8 POST with `text/markdown` and 120 s timeout; `docId` encoded on both new routes — asserted by tests and mutation-pinned at exact head.
- `format` is a TS literal union `'md'|'docx'|'pdf'`; all callers pass literals — no user-controlled input reaches the query param.
- Auth: both new sinks go through the host `apiClient()` (global token + `X-Space-Id`); no credentials, tokens, or secrets in the diff; tests use mocks only.
- Security: CLEARED across all three reviews — 0 exploitable findings over STRIDE / OWASP Top 10 / credential scan / dataflow / authz boundaries; only non-blocking informational hardening notes.
- Client-side input guards (empty file, 25 MB double-check pre-read and post-read, strict UTF-8 decode) all run before any HTTP request — defense-in-depth with the backend as authoritative parser.
- Happy-path import ordering (create-before-import), sessionStorage stash lifecycle (`write→navigate→read+remove→setContent`), and concurrent-picker race-safety are covered; `consumeImportContent` always clears its key.
- i18n: Calibri keys added to both `en-US.json` and `zh-CN.json`; `pnpm i18n:check` passes.
- Export ObjectURL cleanup: explicit 5 s `revokeObjectURL` present (assertion gap is the Nit above, not a missing cleanup).
- Falsified as pre-existing / not widened by this diff: the sessionStorage stash-quota swallow (base accepted unbounded Markdown, so the new 25 MB cap narrows the input set) and the POST→GET wire change for PDF export (intentional; zero internal callers to break). Encoding gaps in the unchanged siblings `getDoc`/`renameDoc`/`deleteDoc` are likewise pre-existing (the `importDocx` instance is kept above as a P3 follow-up per glm5.2).

## Verification Results

| Code Review | NEEDS_CHANGES | 4 P2 + 6 P3 + 3 Nit + 5 human-verify (deduped across 3 reviewers); glm5.2 flags the orphan-doc lifecycle regression as a blocking must-fix, the other two rate all findings non-blocking |
| E2E Test | PARTIAL | All 23 PR-targeted tests pass; full `@octo/docs` suite 1608/1650 (42 failures environmental/pre-existing, none in PR files); mutation probes confirm findings; cross-repo backend routes (octo-docs-backend) not exercisable in any review |
| Security | CLEARED_WITH_RISK | 0 Critical/High/Medium/Low across all three reviews; input guards strong; residual risk only on the backend side, which is not runnable here |

wwwzzzcccc:
Addressed the remaining executable consolidated-review item in amended single commit `00debf55`.

- Added a CI unit-test gate for all five unified import/export suites (EditorShell mappings, font verification, Markdown/DOCX rollback flow, export API, import API).
- The orphan-document rollback, URL encoding, stale JSDoc, and DOCX apply-failure coverage were already present at the reviewed current head.
- All five focused suites, i18n check, and diff-check pass locally.
- Fresh GitHub CI is running.

Jerry-Xin (approved):
🔄 Re-review at `00debf55` — APPROVE. The only change since my prior approval (`a32cbd48`) is CI hardening that wires the new import/export unit tests into the pipeline; all source is byte-identical.

### 🔎 True delta (`a32cbd48` → `00debf55`)
- History is a rebase (`ahead 1 / behind 1`, merge-base `a8bad922`); the raw compare list is bloated by merge-base drift. Two-dot `git diff a32cbd48 00debf55` shows exactly **one changed file**: `.github/workflows/ci.yml` (+8 lines).
- All import/export source and test files (`packages/docs/src/pages/docsApi.ts`, `packages/docs/src/editor/importFlow.ts`, `packages/docs/src/pages/DocsHome.tsx`, `packages/docs/src/editor/EditorShell.tsx`, and the two `docsApi.*.test.ts`) are byte-identical to `a32cbd48` — no regression to the previously-approved surface.

### ✅ Tests are now wired into CI (peer concern resolved)
The earlier concern that the new suite was not collected/run by CI does **not** hold at this head. The new `.github/workflows/ci.yml` step in the `build` job (runs on `pull_request`, gated on `changes.code == 'true'`, positioned before `Build`) runs:

`pnpm --filter @octo/docs exec vitest run` with the five files named explicitly:
`src/editor/EditorShell.test.tsx`, `src/editor/fontFamilies.verify.test.ts`, `src/editor/importFlow.markdown.test.ts`, `src/pages/docsApi.export.test.ts`, `src/pages/docsApi.import.test.ts`.

Byte-verified: package name is `@octo/docs` (matches the filter), `vitest@^4.1.5` is a dependency, and all five files exist at `00debf55`. Because the files are named explicitly, collection does not depend on any glob/`include` config — they are guaranteed to run and a failure fails CI before `Build`.

### 💬 Non-blocking (pre-existing, not introduced here)
- `packages/docs/src/editor/EditorShell.tsx:477` — the `exporting` boolean is not an atomic in-flight lock; two rapid clicks before rerender could start duplicate downloads. A ref-based guard would close the window. Minor.
- Export intentionally reads the authoritative live server document (see the in-code comment at `packages/docs/src/editor/EditorShell.tsx`), so the download reflects backend state rather than a possibly-stale local snapshot — this is by design, not a data-loss bug.

### ✅ Highlights
- Import failures clean up the newly created document while preserving the original error.
- All import/export document ids are `encodeURIComponent`-encoded in the endpoint paths.
- Binary export uses the authenticated API client with correct `arraybuffer` handling.
- The new CI step closes the untested-in-CI gap for the whole import/export suite.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1046 (octo-web)

## Summary

This PR routes Markdown, DOCX, and PDF import/export through unified backend pipelines, replacing client-side parsing (Markdown) and client-side serialization (Markdown, DOCX) with authoritative backend endpoints. The import flow now creates a destination doc, sends the file to the backend for atomic apply to the live Y.Doc, and cleans up on failure. Exports consolidate three separate handlers into a single `exportDocFile` endpoint. A new `importMarkdown` API adds client-side byte-cap (25 MiB) and strict UTF-8 validation before upload. Calibri is added to both Doc and Sheet font catalogs.

The refactor is well-structured: import and export paths are symmetric, error cleanup is correct, URL encoding is consistently applied, and the test suite covers happy paths, error propagation, race conditions, and boundary validation.

## Verification

- ✅ **Import cleanup on failure** — both `runMarkdownImport` (`importFlow.ts:197`) and `runDocxImport` (`importFlow.ts:411`) call `deleteDoc(created.docId).catch(() => undefined)` then rethrow the original error. Error is preserved even when cleanup also fails. Verified by `importFlow.markdown.test.ts:87-96` and `:120-130`.
- ✅ **Dual-path parity (C1)** — Markdown and DOCX import flows follow identical create → apply → cleanup-on-failure → stash-warnings-only patterns. Both add `X-Octo-Import-Apply: true` header. Export unifies all three formats through `exportDocFile`.
- ✅ **URL encoding** — `encodeURIComponent(docId)` applied consistently across `exportDocFile` (`docsApi.ts:353`), `importMarkdown` (`docsApi.ts:390`), and `importDocx` (`docsApi.ts:417`). Tests verify encoded paths (`docsApi.import.test.ts:34`).
- ✅ **Shared client** — all API calls go through `apiClient()`, preserving the global auth interceptor (token, X-Space-Id) and `/api/v1/` baseURL. No raw `fetch` or hardcoded paths.
- ✅ **EditorShell content injection reorder** — warnings are now consumed before the `pmDoc` check (`EditorShell.tsx:349-352`). Since backend imports never stash PM content, `consumeImportContent` returns `null` and the flow short-circuits correctly. Warnings surface independently. Backward compatible.
- ✅ **Export no longer depends on editor instance** — `downloadExport` (`EditorShell.tsx:477`) removed the `instance` check. Backend reads from authoritative Y.Doc, so exports work even before the local editor initializes. Correct.
- ✅ **Test coverage** — 5 new/updated test files with 23 tests covering API contract, import flow ordering, error propagation, race safety, size validation, UTF-8 boundary, export filename, and font registration.
- ✅ **i18n parity** — Calibri key present in both `en-US.json` and `zh-CN.json`; verified by `fontFamilies.verify.test.ts`.
- ✅ **CI integration** — new test step in `.github/workflows/ci.yml:91-98` runs all 5 affected test files before build.

Static analysis only at head `00debf55ce6258dfae6ab392b5b5c9aea695fcac`; build and tests not executed in this environment.

## Findings

### P2 — Dead code retained (`packages/docs/src/editor/importFlow.ts:15`)

`stashImportContent` (line 15), `consumeImportContent` (line 59), and `migrateImportedImages` (line 225) are still exported but no longer reachable from any production import path. `EditorShell.tsx:339` still imports and calls `consumeImportContent`, which always returns `null` in the new flow — harmless but adds confusion for future readers. Similarly, `export/markdown.ts:121` (`exportDocToMarkdown`) and `export/docx/index.ts:28` (`exportDocToDocx`) are no longer imported by any production code. The `importFlow.images.test.ts` tests `migrateImportedImages` in isolation against a function that no longer has a production caller.

Consider removing these in a follow-up cleanup to reduce dead surface area.

### P2 — CI test step uses hardcoded file list (`.github/workflows/ci.yml:91`)

The new CI step explicitly lists 5 test file paths. If new test files are added to `@octo/docs` (e.g. a future `importFlow.docx.test.ts`), they will not run in CI unless this list is manually updated. Consider `pnpm --filter @octo/docs exec vitest run` (all tests in the package) or a glob pattern to stay resilient to future test additions.

- **C1 (dual-path parity)** — CLEAR. Markdown and DOCX import follow identical create→apply→cleanup patterns. Export unifies all formats.
- **C2 (control-flow ordering)** — CLEAR. `downloadExport` reused by three wrappers with typed enum values; no ordering or double-action risk.
- **C3 (authorization boundary)** — CLEAR. New endpoints go through the same authenticated `apiClient()`. No new unguarded entry points.
- **C4 (authorization lifecycle)** — N/A. No auth/lifecycle changes.
- **C5 (build/note)** — N/A. Frontend code change, not packaging.
- **C6 (governance/docs)** — N/A. No governance or policy changes.

## Data Flow Trace

- **Import (Markdown)**: `pickMdFile()` → `File` object → `importMarkdown(docId, file)` reads `arrayBuffer()`, validates size/UTF-8, POSTs raw bytes → backend returns `{ warnings }` → stashed in sessionStorage → consumed by `EditorShell` on mount → editor loads content from live Y.Doc.
- **Import (DOCX)**: `pickDocxFile()` → `File` object → `importDocx(docId, file)` reads `arrayBuffer()`, POSTs raw bytes → backend returns `{ warnings }` → same stash/consume pattern.
- **Export**: `downloadExport(format)` → `exportDocFile(docId, format)` GET with `responseType: 'arraybuffer'` → `Blob` with correct MIME → browser download trigger.
- **Font**: Calibri registered in `fontFamilies.ts` (Doc editor) and `CollabSheet.ts` (Sheet), i18n keys added in both locales.

## Verdict: APPROVED

Well-executed refactor that correctly moves import/export authority to the backend. The atomic apply pattern with cleanup-on-failure is sound, URL encoding is consistently applied, tests cover the critical paths, and the blind spot checklist is clean. The two P2 items (dead code, CI hardcoded list) are non-blocking maintenance concerns.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; two P2 maintenance items (dead code retention, CI test step fragility) are non-blocking.

PR #1048, opened 2026-07-23
feat(loop): persist issue filters by workspace

### LINKED ISSUE #1047: Feature: persist Loop issue filters by workspace
## Problem

Loop issue filters are currently stored only in component state. After users choose filters and refresh the page, the selected filters are lost and the list returns to the default view.

This is especially disruptive when users are working in multiple workspaces, because each workspace should keep its own issue filtering context without leaking values across workspaces.

## Proposed solution

- Persist the Loop issue filter state by workspace.
- Restore the persisted filters when the Loop issue page loads.
- Keep the rendered filter controls and list query aligned with the restored state.
- Clear persisted filters when the user clears the filter panel.
- Validate stored values and tolerate stale option IDs by keeping only the intersection with currently available workspace options.

## Acceptance criteria

- Selected Loop issue filters remain selected after page refresh.
- The restored list data is queried with the restored filters.
- Filter state is isolated by workspace.
- Clearing filters also clears the persisted default state.
- Invalid or stale stored filter data never crashes the page.
- Existing Loop issue list, board, grouped view, and workspace switching flows continue to work.

## Validation

- Add focused unit tests for storage keys, malformed storage payloads, clear behavior, My Loop filtering, and stale option intersection.
- Run the Loop package test suite.

superalsrk:
## Summary

- Persist Loop issue filters per workspace and restore them after refresh.
- Validate stored filter payloads before applying them to the page.
- Reconcile stored assignee, creator, project, and label IDs against current workspace options so stale values cannot crash the page.

## Changes

- Add a small filter persistence helper for storage key generation, read/write normalization, default cleanup, and option intersection.
- Wire `IssuePage` to restore, persist, clear, and reconcile filters by `workspaceSlug + viewKey`.
- Extend assignee candidate loading with a `loaded` flag while keeping the existing hook API.
- Add focused unit tests for workspace-scoped keys, malformed storage data, default cleanup, My Loop filtering, and stale option intersections.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm --filter @octo/loop test`
- `git diff --cached --check`

Known existing check gap:
- `pnpm --filter @octo/loop exec tsc --noEmit` fails on pre-existing dependency and cross-package type errors, including Semi Foundation lodash/prismjs declarations and existing `dmworkbase` React/NodeJS type issues.

Closes #1047

### DIFF
--- a/packages/dmloop/src/api/__tests__/directory.test.ts
+++ b/packages/dmloop/src/api/__tests__/directory.test.ts
@@ -24,6 +24,7 @@ vi.mock("../http", () => ({
 import {
   ensureDirectory,
   invalidateDirectory,
+  listAssigneeCandidateState,
   listAssigneeCandidates,
   listProjectOptions,
 } from "../directory";
@@ -40,19 +41,30 @@ describe("directory", () => {
     vi.mocked(httpGet).mockReset();
   });
 
-  it("keeps display fallback data but rejects authoritative candidates when a candidate source fails", async () => {
+  it("keeps partial candidates but marks them non-authoritative when a candidate source fails", async () => {
     vi.mocked(httpGet).mockImplementation((path: string) => {
-      if (path === "/workspaces/ws-1/members") return Promise.reject(new Error("members failed"));
-      if (path === "/agents") return Promise.resolve([{ id: "a1", name: "Agent" }]);
+      if (path === "/workspaces/ws-1/members") return Promise.resolve([{ user_id: "m1", name: "Member" }]);
+      if (path === "/agents") return Promise.reject(new Error("agents failed"));
       if (path === "/squads") return Promise.resolve([{ id: "s1", name: "Squad" }]);
       if (path === "/projects") return Promise.resolve({ projects: [{ id: "p1", title: "Project" }] });
       return Promise.reject(new Error(`unexpected ${path}`));
     });
 
     const dir = await ensureDirectory();
-    expect(dir.agentName.get("a1")).toBe("Agent");
+    expect(dir.memberName.get("m1")).toBe("Member");
+    expect(dir.squadName.get("s1")).toBe("Squad");
     expect(dir.projectName.get("p1")).toBe("Project");
-    await expect(listAssigneeCandidates()).rejects.toThrow("Assignee directory failed");
+    await expect(listAssigneeCandidates()).resolves.toEqual([
+      { id: "m1", type: "member", name: "Member", octo_uid: null },
+      { id: "s1", type: "squad", name: "Squad" },
+    ]);
+    await expect(listAssigneeCandidateState()).resolves.toEqual({
+      candidates: [
+        { id: "m1", type: "member", name: "Member", octo_uid: null },
+        { id: "s1", type: "squad", name: "Squad" },
+      ],
+      succeeded: false,
+    });
     await expect(listProjectOptions()).resolves.toEqual([{ id: "p1", title: "Project" }]);
 
     const persisted = {
@@ -64,7 +76,7 @@ describe("directory", () => {
     const reconciled = reconcileIssueFilters(
       persisted,
       issueFilterOptionIds({
-        candidates: [],
+        candidates: (await listAssigneeCandidateState()).candidates,
         candidatesLoaded: true,
         candidatesSucceeded: false,
         projects: await listProjectOptions(),
--- a/packages/dmloop/src/api/__tests__/directory.test.ts
+++ b/packages/dmloop/src/api/__tests__/directory.test.ts
@@ -1,0 +1,98 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@octo/base", () => ({
+  WKApp: {
+    loginInfo: { token: "token" },
+    shared: {
+      currentSpaceId: "space-1",
+      avatarUser: vi.fn((uid: string) => `avatar:${uid}`),
+    },
+  },
+  SpaceService: {
+    shared: {
+      getAllMembers: vi.fn(() => Promise.resolve([])),
+    },
+  },
+}));
+
+vi.mock("../http", () => ({
+  currentWorkspaceId: vi.fn(() => "ws-1"),
+  currentWorkspaceSlug: vi.fn(() => "alpha"),
+  httpGet: vi.fn(),
+}));
+
+import {
+  ensureDirectory,
+  invalidateDirectory,
+  listAssigneeCandidates,
+  listProjectOptions,
+} from "../directory";
+import { httpGet } from "../http";
+import {
+  defaultIssueFilters,
+  issueFilterOptionIds,
+  reconcileIssueFilters,
+} from "../../ui/issueFilterPersistence";
+
+describe("directory", () => {
+  beforeEach(() => {
+    invalidateDirectory();
+    vi.mocked(httpGet).mockReset();
+  });
+
+  it("keeps display fallback data but rejects authoritative candidates when a candidate source fails", async () => {
+    vi.mocked(httpGet).mockImplementation((path: string) => {
+      if (path === "/workspaces/ws-1/members") return Promise.reject(new Error("members failed"));
+      if (path === "/agents") return Promise.resolve([{ id: "a1", name: "Agent" }]);
+      if (path === "/squads") return Promise.resolve([{ id: "s1", name: "Squad" }]);
+      if (path === "/projects") return Promise.resolve({ projects: [{ id: "p1", title: "Project" }] });
+      return Promise.reject(new Error(`unexpected ${path}`));
+    });
+
+    const dir = await ensureDirectory();
+    expect(dir.agentName.get("a1")).toBe("Agent");
+    expect(dir.projectName.get("p1")).toBe("Project");
+    await expect(listAssigneeCandidates()).rejects.toThrow("Assignee directory failed");
+    await expect(listProjectOptions()).resolves.toEqual([{ id: "p1", title: "Project" }]);
+
+    const persisted = {
+      ...defaultIssueFilters(),
+      assigneeIds: ["stored-assignee"],
+      creatorIds: ["stored-creator"],
+      projectIds: ["p1", "stale-project"],
+    };
+    const reconciled = reconcileIssueFilters(
+      persisted,
+      issueFilterOptionIds({
+        candidates: [],
+        candidatesLoaded: true,
+        candidatesSucceeded: false,
+        projects: await listProjectOptions(),
+        projectsLoaded: true,
+        projectsSucceeded: true,
+        labels: [],
+        labelsLoaded: false,
+        labelsSucceeded: false,
+      }),
+      false,
+    );
+    expect(reconciled.assigneeIds).toEqual(["stored-assignee"]);
+    expect(reconciled.creatorIds).toEqual(["stored-creator"]);
+    expect(reconciled.projectIds).toEqual(["p1"]);
+  });
+
+  it("rejects authoritative project options when the project source fails", async () => {
+    vi.mocked(httpGet).mockImplementation((path: string) => {
+      if (path === "/workspaces/ws-1/members") return Promise.resolve([{ user_id: "m1", name: "Member" }]);
+      if (path === "/agents") return Promise.resolve([]);
+      if (path === "/squads") return Promise.resolve([]);
+      if (path === "/projects") return Promise.reject(new Error("projects failed"));
+      return Promise.reject(new Error(`unexpected ${path}`));
+    });
+
+    await expect(listAssigneeCandidates()).resolves.toEqual([
+      { id: "m1", type: "member", name: "Member", octo_uid: null },
+    ]);
+    await expect(listProjectOptions()).rejects.toThrow("Project directory failed");
+  });
+});
--- a/packages/dmloop/src/api/directory.ts
+++ b/packages/dmloop/src/api/directory.ts
@@ -15,6 +15,8 @@ interface Directory {
   squadName: Map<string, string>;
   projectName: Map<string, string>;
   candidates: AssigneeCandidate[];
+  candidatesSucceeded: boolean;
+  projectsSucceeded: boolean;
 }
 
 let _cache: Directory | null = null;
@@ -29,26 +31,42 @@ async function build(): Promise<Directory> {
   const slug = currentWorkspaceSlug();
   const wsId = currentWorkspaceId();
   const [members, agents, squads, projectsResp] = await Promise.all([
-    wsId ? httpGet<Array<{ user_id: string; name: string; octo_uid?: string | null }>>(`/workspaces/${wsId}/members`).catch(() => []) : Promise.resolve([]),
+    wsId
+      ? settle(
+          httpGet<Array<{ user_id: string; name: string; octo_uid?: string | null }>>(
+            `/workspaces/${wsId}/members`,
+          ),
+          [],
+        )
+      : Promise.resolve({ value: [], succeeded: true }),
     // include_archived: resolve names for archived (soft-deleted) agents too, so issues/comments/
     // timeline referencing a deleted AI teammate show its name, not a raw id. Archived entries feed
     // the name map only; the assignee picker (candidates) stays active-only.
-    httpGet<Array<{ id: string; name: string; archived_at?: string | null }>>("/agents", { include_archived: true }).catch(() => []),
-    httpGet<Array<{ id: string; name: string }>>("/squads").catch(() => []),
-    httpGet<{ projects: Array<{ id: string; title: string }> }>("/projects").catch(() => ({ projects: [] })),
+    settle(
+      httpGet<Array<{ id: string; name: string; archived_at?: string | null }>>(
+        "/agents",
+        { include_archived: true },
+      ),
+      [],
+    ),
+    settle(httpGet<Array<{ id: string; name: string }>>("/squads"), []),
+    settle(
+      httpGet<{ projects: Array<{ id: string; title: string }> }>("/projects"),
+      { projects: [] },
+    ),
   ]);
   // Resolve member identity to octo names — but only pull the (potentially large)
   // space roster when at least one member is octo-bridged; a pure-native
   // workspace skips the roster fetch entirely.
   const octoName = new Map<string, string>();
-  if (members.some((m) => m.octo_uid)) {
+  if (members.value.some((m) => m.octo_uid)) {
     const roster = await SpaceService.shared.getAllMembers(WKApp.shared.currentSpaceId).catch(() => []);
     for (const s of roster) octoName.set(s.uid, s.name);
   }
   const memberName = new Map<string, string>();
   const memberOctoUid = new Map<string, string>();
   const candidates: AssigneeCandidate[] = [];
-  for (const m of members) {
+  for (const m of members.value) {
     // octo 身份优先：octo_uid 命中 space 名册取 octo 名字，否则回退后端显示名。
     const name = (m.octo_uid && octoName.get(m.octo_uid)) || m.name;
     memberName.set(m.user_id, name);
@@ -56,18 +74,40 @@ async function build(): Promise<Directory> {
     candidates.push({ id: m.user_id, type: "member", name, octo_uid: m.octo_uid ?? null });
   }
   const agentName = new Map<string, string>();
-  for (const a of agents) {
+  for (const a of agents.value) {
     agentName.set(a.id, a.name); // resolve names incl. archived (for display)
     if (!a.archived_at) candidates.push({ id: a.id, type: "agent", name: a.name }); // picker: active only
   }
   const squadName = new Map<string, string>();
-  for (const s of squads) {
+  for (const s of squads.value) {
     squadName.set(s.id, s.name);
     candidates.push({ id: s.id, type: "squad", name: s.name });
   }
   const projectName = new Map<string, string>();
-  for (const p of (projectsResp.projects ?? [])) projectName.set(p.id, p.title);
-  return { slug, memberName, memberOctoUid, agentName, squadName, projectName, candidates };
+  for (const p of (projectsResp.value.projects ?? [])) projectName.set(p.id, p.title);
+  return {
+    slug,
+    memberName,
+    memberOctoUid,
+    agentName,
+    squadName,
+    projectName,
+    candidates,
+    candidatesSucceeded:
+      members.succeeded && agents.succeeded && squads.succeeded,
+    projectsSucceeded: projectsResp.succeeded,
+  };
+}
+
+async function settle<T>(
+  promise: Promise<T>,
+  fallback: T,
+): Promise<{ value: T; succeeded: boolean }> {
+  try {
+    return { value: await promise, succeeded: true };
+  } catch {
+    return { value: fallback, succeeded: false };
+  }
 }
 
 export async function ensureDirectory(force = false): Promise<Directory> {
@@ -138,6 +178,7 @@ export function projectNameOf(
 
 export async function listAssigneeCandidates(): Promise<AssigneeCandidate[]> {
   const dir = await ensureDirectory();
+  if (!dir.candidatesSucceeded) throw new Error("Assignee directory failed");
   return dir.candidates;
 }
 
@@ -145,5 +186,6 @@ export async function listAssigneeCandidates(): Promise<AssigneeCandidate[]> {
 // listAssigneeCandidates 对称——调用方无需伸手进 Directory 内部结构。
 export async function listProjectOptions(): Promise<Array<{ id: string; title: string }>> {
   const dir = await ensureDirectory();
+  if (!dir.projectsSucceeded) throw new Error("Project directory failed");
   return [...dir.projectName].map(([id, title]) => ({ id, title }));
 }
--- a/packages/dmloop/src/api/directory.ts
+++ b/packages/dmloop/src/api/directory.ts
@@ -178,8 +178,18 @@ export function projectNameOf(
 
 export async function listAssigneeCandidates(): Promise<AssigneeCandidate[]> {
   const dir = await ensureDirectory();
-  if (!dir.candidatesSucceeded) throw new Error("Assignee directory failed");
   return dir.candidates;
+}
+
+export async function listAssigneeCandidateState(): Promise<{
+  candidates: AssigneeCandidate[];
+  succeeded: boolean;
+}> {
+  const dir = await ensureDirectory();
+  return {
+    candidates: dir.candidates,
+    succeeded: dir.candidatesSucceeded,
+  };
 }
 
 // 项目筛选下拉的选项，复用已缓存的 directory（不额外请求 /projects），与
--- a/packages/dmloop/src/api/issueApi.ts
+++ b/packages/dmloop/src/api/issueApi.ts
@@ -14,7 +14,13 @@ import type {
   CommentTriggerAgent,
 } from "./types";
 import { httpGet, httpPost, httpPut, httpDelete } from "./http";
-import { ensureDirectory, actorName, actorAvatar, listAssigneeCandidates as dirCandidates } from "./directory";
+import {
+  ensureDirectory,
+  actorName,
+  actorAvatar,
+  listAssigneeCandidates as dirCandidates,
+  listAssigneeCandidateState as dirCandidateState,
+} from "./directory";
 import { arrayFilterQuery } from "./issueFilterQuery";
 
 // enrich 的同步核心:调用方已拿到 directory 时复用,避免每组重复 ensureDirectory。
@@ -249,4 +255,11 @@ export function updateComment(commentId: string, content: string): Promise<Issue
 /* ---------- 指派候选 ---------- */
 export function listAssigneeCandidates(): Promise<AssigneeCandidate[]> {
   return dirCandidates();
+}
+
+export function listAssigneeCandidateState(): Promise<{
+  candidates: AssigneeCandidate[];
+  succeeded: boolean;
+}> {
+  return dirCandidateState();
 }
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -38,6 +38,7 @@ import { pushFleetIssueDeepLink, replaceLoopRootPath } from "../issueDeepLink";
 import { readView, writeView } from "../ui/viewMode";
 import {
   defaultIssueFilters,
+  issueFilterOptionIds,
   issueFilterStorageKey,
   readIssueFilterState,
   reconcileIssueFilters,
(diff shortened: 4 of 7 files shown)


Jerry-Xin (dismissed):
This PR is relevant to octo-web, but transient option-loading failures can permanently erase persisted filters.

### 🔴 Blocking

- 🔴 Critical — Failed option requests are treated as authoritative empty results. In [IssuePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-9eq9x641/repo/packages/dmloop/src/pages/IssuePage.tsx:125), project/label failures set empty arrays and `loaded=true`; [useAssigneeCandidates.ts](/Users/mlamp/.openclaw/tmp/pr-review-9eq9x641/repo/packages/dmloop/src/ui/useAssigneeCandidates.ts:25) does the same for assignees. Reconciliation then removes every stored ID, and the persistence effect writes that loss back to storage. A temporary network/permission/server failure therefore permanently clears valid user filters. Track successful loading separately and reconcile only after a successful response. Add tests covering request failure without stored-filter mutation.

### 💬 Non-blocking

- 🟡 Warning — [issueFilterPersistence.ts](/Users/mlamp/.openclaw/tmp/pr-review-9eq9x641/repo/packages/dmloop/src/ui/issueFilterPersistence.ts:162) accepts `"involves"` for a normal Issue page, although that scope is not selectable there. Validate scopes against the page context to prevent an inaccessible restored state.

- 🟡 Warning — The added tests cover the pure helper but not the asynchronous restore/reconcile/persist ordering in `IssuePage`, where the blocking data-loss behavior occurs.

### ✅ Highlights

- Storage keys are correctly scoped by workspace and view.
- Stored values and date ranges receive defensive validation.
- My Loop-specific filtering and stale-ID intersection have focused unit coverage.

Jerry-Xin (dismissed):
Filter persistence is well-structured and workspace-scoped, but a transient option-load failure can permanently erase a user's persisted filters.

### 🔴 Blocking

- 🔴 Critical — Failed option requests are treated as authoritative empty results, wiping stored filters. In `packages/dmloop/src/pages/IssuePage.tsx:125`, the project/label loaders set an empty array **and** mark `loaded=true` on `.catch()`; `packages/dmloop/src/ui/useAssigneeCandidates.ts:25` does the same for assignee candidates. The reconcile effect then runs with `loaded=true` + empty options, so `reconcileIssueFilters` passes `[]` (not `undefined`) into `intersect`, which strips every restored `assigneeIds`/`creatorIds`/`projectIds`/`labelIds`. The persistence effect immediately writes that emptied state back to storage. Result: a temporary network / permission / 5xx failure permanently clears valid saved filters. Track load **success** separately from load **completion** — on failure keep the option list as `undefined` so reconciliation is skipped rather than pruning against an empty set. Add a test covering "option request fails → stored filters are not mutated".

### 💬 Non-blocking

- 🟡 `packages/dmloop/src/ui/issueFilterPersistence.ts:162` (`normalizeState`) accepts `"involves"` as a restored scope on a normal Issue page even though that scope is not selectable there, allowing an inaccessible restored state. Consider validating the restored scope against the page's allowed scopes.
- 🟡 New tests cover the pure helpers thoroughly but not the async restore → reconcile → persist ordering inside `IssuePage`, which is exactly where the blocking data-loss occurs. Adding a component/integration test around that ordering would guard the fix.
- 🟡 Storage key `loop.issue.filters:${viewKey}:${workspaceSlug}` uses `:` as a delimiter; if a `viewKey`/`workspaceSlug` could ever contain `:` there is a theoretical collision. Low risk given current slug/view constraints, but worth keeping in mind.

### ✅ Highlights

- Storage keys are correctly scoped by both workspace and view; `null` key short-circuits persistence safely (no cross-workspace bleed).
- Robust deserialization: `JSON.parse` wrapped in try/catch with a whitelisted, field-by-field rebuild (enum/string-list guards, date-range validation) — no blind spread merge, so no prototype-pollution vector and no crash on corrupt/tampered storage.
- My-Loop-specific scrubbing (keyword/assignee/creator forced empty on both read and write) is consistent.
- `useAssigneeCandidates` kept as a thin backward-compat wrapper over `useAssigneeCandidateState` — no dead code, no broken callers; `alive` guards prevent setState-after-unmount.

Jerry-Xin (dismissed):
The persistence feature is relevant and well structured, but failed option requests can permanently erase valid stored filters.

### 🔴 Blocking

- 🔴 Critical — Failed directory requests are treated as authoritative empty option lists. [`useAssigneeCandidates.ts:25`](/Users/mlamp/.openclaw/tmp/pr-review-3ov7wf7j/repo/packages/dmloop/src/ui/useAssigneeCandidates.ts:25) marks candidate loading as successful whenever `listAssigneeCandidates()` resolves. However, the directory loader catches individual member, agent, squad, and project failures and substitutes empty arrays ([`directory.ts:31`](/Users/mlamp/.openclaw/tmp/pr-review-3ov7wf7j/repo/packages/dmloop/src/api/directory.ts:31)). Consequently, [`IssuePage.tsx:149`](/Users/mlamp/.openclaw/tmp/pr-review-3ov7wf7j/repo/packages/dmloop/src/pages/IssuePage.tsx:149) considers partial or total failures authoritative, removes persisted assignee/creator/project IDs, and then writes the pruned state back to storage. A transient network or permission failure therefore permanently loses the user’s saved filters. Propagate per-option load success from the directory layer, or avoid reconciliation unless the underlying requests genuinely succeeded. Add coverage for internally swallowed/partial directory failures.

### ✅ Highlights

- Storage payloads are normalized defensively.
- Keys correctly separate the two views and workspace slugs.
- My Loop-specific filtering constraints are covered by focused tests.
- Workspace changes remount `IssuePage`, so mount-only restoration matches the existing lifecycle.

Tests could not be executed locally because dependencies are not installed (`vitest: command not found`).

Jerry-Xin (dismissed):
Filter persistence is relevant and well-structured, and the added success-tracking is the right direction — but a transient directory failure can still permanently erase valid saved filters, because the underlying directory layer swallows sub-request errors into resolved-empty results.

### 🔴 Blocking

- 🔴 Critical — The `succeeded` guard is defeated by the directory layer, so failed option loads still become authoritative empty lists. `packages/dmloop/src/pages/IssuePage.tsx:134` and `packages/dmloop/src/ui/useAssigneeCandidates.ts:24` set `succeeded=true` whenever `listProjectOptions()` / `listLabels()` / `listAssigneeCandidates()` **resolve**. But `packages/dmloop/src/api/directory.ts:31` (`build()`) wraps each sub-request (`/workspaces/{id}/members`, `/agents`, `/squads`, `/projects`) in its own `.catch(() => [])` inside `Promise.all`, so a transient network/5xx failure resolves with an empty list instead of rejecting. The chain therefore resolves successfully with empty data → `succeeded=true` → `issueFilterOptionIds` returns `[]` (authoritative) → `reconcileIssueFilters` intersects stored IDs against `[]` and drops them → the persistence effect writes the emptied filters back to storage. Net result: the exact data-loss this PR set out to fix is still reachable on a transient directory sub-request failure. Propagate per-resource success from the authoritative endpoint (e.g. have `ensureDirectory`/`listProjectOptions`/`listAssigneeCandidates` surface a real failure, or reject when the authoritative call fails) so `succeeded` reflects the actual request outcome. Add an integration test that fails a directory sub-request and asserts stored filters are NOT mutated — the current unit test only exercises an explicitly supplied `succeeded: false`, which never reproduces this path.

### 💬 Non-blocking

- 🟡 Storage key `loop.issue.filters:${viewKey}:${workspaceSlug}` uses `:` as a delimiter; if `viewKey`/`workspaceSlug` could ever contain `:` there is a theoretical collision. Low risk given current slug/view constraints.

### ✅ Highlights

- Storage keys are correctly scoped by both workspace and view; a `null` key short-circuits persistence safely (no cross-workspace bleed).
- Robust deserialization: `JSON.parse` in try/catch with a whitelisted, field-by-field rebuild (enum/string-list guards, date-range validation) — no blind spread merge, so no prototype-pollution vector and no crash on corrupt/tampered storage.
- Extracting `issueFilterOptionIds` into the persistence module with dedicated unit tests, and restricting restored scope to `ISSUE_PAGE_SCOPES` (dropping `involves`), are clean, testable improvements.
- My-Loop scrubbing and stale-ID intersection have focused unit coverage.

Jerry-Xin (dismissed):
Re-review at head adc7bcd. The prior blocking issue (a failed option-load being mistreated as an empty authoritative result, permanently clearing persisted filters) is fixed and now well covered by tests. No regressions to the previously-cleared surface. Approving.

### ✅ Blocking issue resolved
- 🔴 (prev) Failed option request wiped persisted filters — **FIXED**. `issueFilterOptionIds()` in `packages/dmloop/src/ui/issueFilterPersistence.ts` builds each option-id set only when `loaded && succeeded`; on a completed-but-failed load it emits `undefined`. `intersect(value, allowed?)` returns `value` unchanged when `allowed` is `undefined`, so `reconcileIssueFilters` preserves stored ids on failure (and returns the same object reference, so no clobbering re-persist). `packages/dmloop/src/pages/IssuePage.tsx` now `.catch` only flips `succeeded=false` and keeps prior option values instead of blanking them — belt-and-suspenders. Genuine empty results (succeeded=true, options=[]) still reconcile correctly, so intentional empty state is not swallowed.

### ✅ Notes previously raised — addressed
- Async-restore vs persist ordering / data-loss coverage: new unit tests cover failed-load → `undefined` option sets (non-authoritative), successful-load → real id lists, and identity preservation when nothing changes.
- Restore scope validation: normal issue page no longer restores the My Loop (`involves`) scope (`ISSUE_PAGE_SCOPES`), with a dedicated test.

### 💬 Non-blocking
- 💬 Storage key namespacing: `issueFilterStorageKey` derives from `viewKey + workspaceSlug`. If two Octo spaces can host workspaces with identical slugs, a user switching between them on the same browser could theoretically restore the other space's filters. This is pre-existing design (unchanged by this PR) and the app already keys workspace identity by slug elsewhere; not a blocker here. Consider including a stable space identifier in the key (plus a collision test) in a follow-up if same-slug-across-spaces is reachable.

### ✅ Highlights
- Stored values are defensively normalized with try/catch and default fallback.
- Reconciliation preserves object identity when no change is required (avoids needless re-persist loops).
- Failure is cleanly distinguished from empty across candidates, projects, and labels.
- My Loop restricted-filter semantics have focused tests.

Jerry-Xin (dismissed):
Re-review at head adc7bcdf. The persistence design is well-tested in isolation, but the data-loss risk flagged in prior rounds is still open: a transient directory subrequest failure can still permanently erase valid stored filters, because the fix guards a branch that the real code never reaches.

## 🔴 Blocking

- 🔴 Failed directory requests are still treated as authoritative empty option lists, silently deleting persisted filters. Trace:
  - `packages/dmloop/src/api/directory.ts` `build()` wraps every subrequest in a swallow: members/agents/squads use `.catch(() => [])` and projects uses `.catch(() => ({ projects: [] }))`. `httpGet` throws on failure (`packages/dmloop/src/api/http.ts`), but `build()` catches it and **resolves** with empty/partial data.
  - Because `listAssigneeCandidates()` / `listProjectOptions()` therefore **never reject**, `packages/dmloop/src/ui/useAssigneeCandidates.ts:25` runs its `.then` branch and sets `succeeded: true` with an empty `candidates` list; likewise `packages/dmloop/src/pages/IssuePage.tsx` sets `projectsSucceeded: true` / `labelsSucceeded: true` with empty rows.
  - `issueFilterOptionIds` then sees `succeeded === true` and emits an **empty authoritative** id list (`[]`), not `undefined`.
  - `reconcileIssueFilters` → `intersect(persistedIds, [])`: the `if (!allowed) return value` guard does not trigger for an empty array, so every persisted id is filtered out and the effect persists the cleared filter.
  - Net: a transient failure of the member/agent/squad/project directory load permanently clears the user's saved assignee, creator, project, and label filters.
  - The new commit only hardens the outer reject path (hook `.catch` preserving prev + `succeeded:false`), but that path is unreachable given the internal swallow. The added tests exercise `issueFilterOptionIds` with `succeeded:false` inputs — a state the runtime never produces on failure.
  - Fix direction: make failure observable end-to-end. Either propagate per-source failure (do not swallow in `build()`, or carry a per-source `succeeded` flag out of the directory layer), or treat a failed/partial load as `undefined` (non-authoritative) so `intersect` preserves the stored value. Add an integration-level test that fails a directory subrequest and asserts the persisted filter is preserved.

## 💬 Non-blocking

- 💬 Once the failure signal is threaded through, add the async-restore → persist ordering integration test covering this exact data-loss scenario (previously noted); the current unit tests validate `issueFilterOptionIds` in isolation but not the real load→reconcile→persist sequence.

## ✅ Highlights

- Centralizing option-id construction into `issueFilterOptionIds` is a good structural move and makes the invariant testable.
- Stored payloads get defensive runtime normalization; workspace/view key scoping and My Loop restrictions have focused tests.
- Reconciliation preserves object identity when nothing changes, avoiding spurious re-persists.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

Jerry-Xin (dismissed):
The PR is relevant to `octo-web`, but partial directory failures now break unrelated assignee functionality.

### 🔴 Blocking

- 🔴 Critical — A failure from any one candidate source makes the entire assignee directory reject. [`directory.ts:96`](/Users/mlamp/.openclaw/tmp/pr-review-o6w4g0r0/repo/packages/dmloop/src/api/directory.ts:96) combines member, agent, and squad success flags, while [`directory.ts:179`](/Users/mlamp/.openclaw/tmp/pr-review-o6w4g0r0/repo/packages/dmloop/src/api/directory.ts:179) rejects all candidates unless every source succeeded. Consequently, a temporary `/agents` or `/squads` failure discards successfully loaded members for all existing pickers and prevents My Loop from resolving `myMemberId`, leaving that page empty. The failed result is also cached until explicit invalidation, so it does not recover automatically. Preserve partial candidates for existing consumers and expose authoritative-load status separately for reconciliation.

### ✅ Highlights

- Storage payloads are defensively normalized.
- Workspace/view key scoping and stale-ID intersection have focused tests.
- Failed option requests correctly avoid treating empty fallback arrays as authoritative.

Tests could not be executed locally because dependencies were not installed (`vitest: command not found`).

Jerry-Xin (dismissed):
Re-review at head `995ffc1c956f`. The original data-loss blocker (directory failures silently stripping persisted filters) is now correctly fixed and tested, but the fix over-corrects: a single candidate-source failure now rejects the entire assignee directory app-wide.

### 🔴 Blocking

- 🔴 A failure from any one candidate source makes the entire assignee directory reject. `packages/dmloop/src/api/directory.ts` computes `candidatesSucceeded = members.succeeded && agents.succeeded && squads.succeeded`, and `listAssigneeCandidates()` throws `"Assignee directory failed"` whenever that is false. Consequently a transient `/agents` or `/squads` failure discards the successfully loaded members for every assignee consumer (AssigneePicker, CreateIssueModal, IssuePage/IssueDetailPage, AgentPage/SquadPage and their detail panels) and prevents My Loop from resolving `myMemberId`, leaving those surfaces empty. Worse, the failed `Directory` is written to `_cache` in `ensureDirectory()` and returned on all later reads until an explicit `invalidateDirectory()` (workspace switch or a mutation) — so a momentary blip poisons assignee pickers for the rest of the session with no auto-recovery. Recommendation: keep returning the partial `candidates` array to display/picker consumers, and expose the authoritative-load flag separately so ONLY the reconcile/strip path (which is the operation that must not run on incomplete data) gates on `candidatesSucceeded`. Same reasoning applies to the project dimension.

### 💬 Non-blocking

- The core data-loss fix is correct end-to-end: `settle()` reports real per-source success, `issueFilterOptionIds()` emits `undefined` for any unsucceeded dimension, and `intersect(value, undefined)` preserves stored IDs — so a failed directory load no longer strips saved assignee/creator/project filters. The added `directory.test.ts` cases (fail members → assignees rejected + stored IDs preserved while a genuine stale project is still removed; fail `/projects` → project options rejected while assignees resolve) cover this well.
- Prior notes still open (not blocking): storage-key `:` delimiter can collide for same-slug workspaces across spaces; per-workspace keys can accumulate unbounded in localStorage over time.

### ✅ Highlights

- Persisted payloads are defensively normalized (whitelisted rebuild, try/catch, no prototype pollution).
- Workspace/view key scoping and stale-ID intersection have focused tests; no cross-workspace bleed; My Loop no longer restores the `involves` scope on the normal page.
- Failed option requests now correctly avoid treating empty fallback arrays as authoritative.

Jerry-Xin (dismissed):
Re-review at fae07d25. The prior blocker (partial-directory over-correction) is fixed, but a new persistence round-trip bug blocks approval.

### 🔴 Blocking

- 🔴 Same-day date-range filter is silently dropped after refresh. `packages/dmloop/src/ui/issueFilterPersistence.ts:266` (`normalizeDateRange`) rejects a range when `start.getTime() >= end.getTime()`, so any single-day selection (`start === end`) fails read-back and write normalization. But the picker at `packages/dmloop/src/pages/IssuePage.tsx:411` accepts a same-day selection, and the request builder at `packages/dmloop/src/pages/IssuePage.tsx:179-200` explicitly handles it by adding one calendar day to the end (`endExclusive`), producing a valid half-open single-day interval. Net effect: a same-day filter works in the active session but silently disappears on refresh — the exact persistence guarantee this PR exists to provide. Fix: accept `start === end`, reject only `start > end`, and add a round-trip test for a same-day range.

### 💬 Non-blocking

- Prior 🟡 items (storage-key delimiter collision across same-slug workspaces; unbounded per-workspace key growth) remain as noted — low priority, not blocking.

### ✅ Highlights

- Prior over-correction is FIXED with correct semantics: `listAssigneeCandidates()` no longer throws on a single-source failure and returns partial candidates; the new `listAssigneeCandidateState()` exposes `{candidates, succeeded}` so `succeeded` gates ONLY the reconcile/clear path. `build()` now uses `settle()` for every source and never throws, so members/squads still populate pickers when `/agents` fails, and `myMemberId`/My Loop keep working.
- Original data-loss fix stays intact: on a failed dimension `candidatesSucceeded=false` → `issueFilterOptionIds` returns `assigneeIds: undefined` → `intersect(value, undefined)` returns `value` unchanged → stored filter IDs are preserved (no reconcile clearing).
- No session-wide cache poisoning: a partial directory now degrades gracefully (partial data shown, filters preserved) instead of throwing across all consumers; workspace switch (slug change) and directory mutations invalidate and rebuild.
- Test coverage added for the partial-failure path (members preserved, `succeeded:false`, reconcile preserves stored IDs).
- Storage keys correctly scoped by workspace and page; malformed payloads and stale option IDs handled conservatively.

Jerry-Xin (dismissed):
The persistence implementation is well tested, but its storage key does not uniquely identify a workspace across Octo spaces.

🔴 Blocking

- 🔴 Critical — Filters can leak between distinct workspaces with the same slug. [`issueFilterPersistence.ts:85`](/Users/mlamp/.openclaw/tmp/pr-review-mwcwimkt/repo/packages/dmloop/src/ui/issueFilterPersistence.ts:85) keys state only by `viewKey + workspaceSlug`. Workspaces are resolved within the current Octo space, and API requests additionally use `X-Space-Id`, so identical slugs can represent different workspaces in different spaces. Switching between them would restore keyword, status, priority, date, and boolean filters from the wrong workspace; ID filters are only partially mitigated by reconciliation. Include the workspace ID or current space ID in the key, and add a collision test.

💬 Non-blocking

- 🟡 Warning — There is no component-level test covering the restore → asynchronous option load → reconcile → persist lifecycle in [`IssuePage.tsx:84`](/Users/mlamp/.openclaw/tmp/pr-review-mwcwimkt/repo/packages/dmloop/src/pages/IssuePage.tsx:84). The helper tests are strong, but an integration test would protect effect ordering and workspace remount behavior.
- 🟡 Warning — Tests could not be rerun locally because dependencies are absent and `vitest` was not found.

✅ Highlights

- Stored payloads are defensively normalized.
- Failed option requests correctly remain non-authoritative.
- Stale option IDs are reconciled without deleting values after partial load failures.
- Directory request generation guards continue to prevent stale cache writes.

Jerry-Xin (dismissed):
The same-day date-range persistence bug is fixed and covered by a round-trip test; all blockers raised across prior iterations are resolved. Approving.

✅ Fixed (was the sole remaining blocker)

- Same-day date filters now survive refresh. `packages/dmloop/src/ui/issueFilterPersistence.ts:266` `normalizeDateRange` now rejects only inverted ranges (`start > end`) and accepts `start === end`. This matches the picker/query path: `packages/dmloop/src/pages/IssuePage.tsx:411` accepts a single-day selection and `IssuePage.tsx:180-182` builds a valid half-open range (`date_end = end + 1 day`), so a same-day range stays a correct one-day window with no inverted or degenerate range and no double counting. A new round-trip test at `packages/dmloop/src/ui/__tests__/issueFilterPersistence.test.ts:76-98` writes `[sameDay, sameDay]` and asserts it survives reload.

💬 Non-blocking

- 🟡 Storage key `loop.issue.filters:{viewKey}:{workspaceSlug}` (`packages/dmloop/src/ui/issueFilterPersistence.ts:85`) omits the space id. Server requests disambiguate a workspace by both workspace slug and space id (`packages/dmloop/src/api/http.ts:78,86`), so two spaces holding a workspace with the same slug would share one stored entry. Real impact is limited: only filter selections are affected (not data), and id-based filters (assignee/project/label) are reconciled/whitelisted against the current space on apply, so cross-space ids are dropped rather than misapplied. The self-healing surface is keyword/status/priority/date. Consider including the space id in the key plus a collision test in a follow-up.
- 🟡 No component-level test covers the restore → async option load → reconcile → persist lifecycle (`packages/dmloop/src/pages/IssuePage.tsx:84`). The helper-level tests are strong; an integration test would guard effect ordering and workspace remount behavior.

✅ Highlights

- Stored payloads are defensively normalized; failed option requests correctly remain non-authoritative.
- Stale option ids are reconciled without deleting values after a partial load failure (the earlier data-loss and over-correction fixes remain intact).
- Directory request-generation guards continue to prevent stale cache writes.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

[Octo-Q] verdict: REQUEST_CHANGES — P1 regression in clearFilters (dateRange not cleared due to spread-over-previous pitfall in defaultIssueFilters). One-line fix; rest of PR is solid.

Jerry-Xin (dismissed):
Re-review at f4e9d4553f28. This commit correctly fixes the storage-key workspace-isolation concern, but a separate blocker in the clear-all path (raised by mochashanyao) is valid and remains unfixed at this head.

🔴 Blocking

- **Clear-all leaves a stale `dateRange` filter, then persists it.** `clearFilters` in `packages/dmloop/src/pages/IssuePage.tsx:361` calls `update({ ...defaultIssueFilters() })`, and `update` at `packages/dmloop/src/pages/IssuePage.tsx:296` merges over previous state: `setF((prev) => ({ ...prev, ...p }))`. Since `defaultIssueFilters()` (`packages/dmloop/src/ui/issueFilterPersistence.ts`) does **not** include a `dateRange` key, the spread cannot overwrite it — every dimension resets to default except `dateRange`, which survives from `prev`. The persistence effect at `packages/dmloop/src/pages/IssuePage.tsx:146` then writes `{ filters: f, scope }` back to storage, so the stale date range is not only still active in the UI after "clear filters" but also persisted. Fix: make clear-all reset all dimensions explicitly, e.g. `update({ ...defaultIssueFilters(), dateRange: undefined })`, or have `defaultIssueFilters()` return `dateRange: undefined` so the spread clears it. Add a test that sets a `dateRange`, invokes the clear-all reset, and asserts `dateRange` is undefined afterward. This finding is credited to mochashanyao; I under-called it in my prior approval at 87a96ed5.

✅ Verified this iteration (byte-level, delta 87a96ed5136b → f4e9d4553f28, fast-forward: ahead 1 / behind 0, merge-base 87a96ed5136b)

- Storage-key collision fix is correct: `issueFilterStorageKey` now takes `workspaceId` and returns `loop.issue.filters:${viewKey}:workspace:${workspaceId}`, returning null unless slug, viewKey, and workspaceId are all present. `IssuePage.tsx` passes `workspaceId` at the call site. New test "does not collide when two spaces contain workspaces with the same slug" covers it.
- The four prior deep fixes (option-loss vs. authoritative-empty reconciliation, directory partial-failure handling, over-correction guard, same-day date range) live in `issueFilterPersistence.ts` and are byte-identical to 87a96ed5 aside from `issueFilterStorageKey` — intact, not regressed. Same-day-range test still passes.

💬 Non-blocking

- 🟡 Storage keys still grow unbounded per (workspace, view); no pruning of old keys. Acceptable for now but worth a cleanup pass later.
- 🟡 No `IssuePage` integration test exercising the effect lifecycle (restore → async option reconcile). Helper tests cover the pure functions only.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1048 (octo-web)

## Summary

This PR persists issue list filters to `localStorage`, scoped per workspace, so users returning to a loop view see their previous filter selections restored. It introduces a new `issueFilterPersistence` module with read/write/normalize/reconcile helpers, adds `succeeded` tracking to the directory build via a `settle()` wrapper, and wires the persistence into `IssuePage` through two new effects (write on change, reconcile stale IDs against live options). The `useAssigneeCandidates` hook gains a stateful variant (`useAssigneeCandidateState`) that surfaces `loaded` and `succeeded` flags alongside the candidate list.

The design is thoughtful: failed option sources are treated as non-authoritative (stored IDs preserved), My Loop views correctly suppress irrelevant filter dimensions, and the storage key includes `workspaceId` to prevent cross-workspace leakage. Tests cover round-trip, malformed data, workspace isolation, and the non-authoritative failure path.

## Verification

- ✅ **Storage key scoping** — `issueFilterStorageKey` requires all three of `workspaceSlug`, `viewKey`, `workspaceId` non-empty; test at `packages/dmloop/src/ui/__tests__/issueFilterPersistence.test.ts:42` confirms no collision when two spaces share a workspace slug.
- ✅ **Non-authoritative failure preservation** — `issueFilterOptionIds` at `packages/dmloop/src/ui/issueFilterPersistence.ts:148-168` returns `undefined` for any dimension where `loaded && succeeded` is false. `reconcileIssueFilters` at `:122-141` passes `undefined` through `intersect()` unchanged, preserving stored IDs. Test at `:210-230` confirms.
- ✅ **settle() wrapper correctness** — `packages/dmloop/src/api/directory.ts:87-95` wraps each API call so individual failures produce `{fallback, succeeded:false}` without aborting the entire directory build. `candidatesSucceeded` is the conjunction of members + agents + squads success; `projectsSucceeded` tracks projects independently.
- ✅ **Backward compatibility of `listProjectOptions` throw** — `packages/dmloop/src/api/directory.ts:199` now throws when `projectsSucceeded` is false. All three existing callers (`AutopilotDetailPage.tsx:109`, `CreateIssueModal.tsx:70`, `CreateAutomationModal.tsx:39`) already have `.catch()` handlers that fall back to `[]` — same end result as the old silent-empty behavior.
- ✅ **Input normalization** — `normalizeFilters` at `:210-232` validates every field: enums against allow-lists, strings filtered for type, date ranges validated for order and parseability. Malformed JSON caught by try/catch in `readIssueFilterState`.
- ✅ **My Loop filter suppression** — `normalizeFilters` and `serializeState` both force `keyword=""`, `assigneeIds=[]`, `noAssignee=false`, `creatorIds=[]` when `isMyLoop=true`. `normalizeState` overrides scope to `fallbackScope`. Reconciliation also strips these. Three layers of enforcement.
- ✅ **Reconciliation loop safety** — `reconcileIssueFilters` returns the same reference when `filtersEqual` is true, so React bails out of the `setF` call. No infinite re-render risk.
- ✅ **Backward-compatible hook API** — `useAssigneeCandidates` at `packages/dmloop/src/ui/useAssigneeCandidates.ts:41-43` is preserved as a thin wrapper over `useAssigneeCandidateState`, so `IssueDetailPage.tsx:315` (the only other caller) continues working unchanged.

Static analysis only at head `f4e9d4553f2823cbadfc9732693895d518592759`; build and tests not executed in this environment.

## Findings

No P0/P1 issues. Two non-blocking observations below.

### P2 — Transient stale IDs in localStorage between mount and option load (`packages/dmloop/src/pages/IssuePage.tsx:146-149`)

The write effect at `:146` fires on the initial render with the persisted filter state (potentially containing assignee/project/label IDs that no longer exist). The reconciliation effect at `:153` cannot strip them until options finish loading asynchronously. If the user navigates away before options complete, localStorage retains stale IDs until the next visit, where the same reconciliation cycle corrects them before any API query is made. No user-visible defect — filter chips in the UI also show stale IDs briefly, but the reconciliation strips them before the next render — just a timing observation.

### Nit — Four separate state variables for option load tracking (`packages/dmloop/src/pages/IssuePage.tsx:123-126`)

`projectsLoaded`, `projectsSucceeded`, `labelsLoaded`, `labelsSucceeded` are four independent `useState` calls. Each `setState` triggers a separate re-render. Combining into a single `useReducer` or state object would cut re-renders during the option-loading phase. Not a correctness concern.

- `currentWorkspaceId` import at `IssuePage.tsx:15` resolves to `packages/dmloop/src/api/http.ts:41` — confirmed exported.
- `listAssigneeCandidateState` pass-through in `issueApi.ts:260-265` correctly delegates to `directory.ts:184-194`.
- `isDefaultState` at `issueFilterPersistence.ts:249-257` correctly skips scope check for My Loop (scope is always overridden to `fallbackScope` by `normalizeState`).
- `removeItem` path in `writeIssueFilterState` at `:117-120` correctly clears localStorage when filters return to defaults. Test at `:164-178` confirms.
- `intersect` helper at `:301-304` returns original array when `allowed` is `undefined` (non-authoritative), filters when defined (authoritative). Correct.
- `datesEqual` at `:312-317` handles `undefined`/`undefined` and `undefined`/`Date[]` correctly.
- `stringList` at `:293-299` deduplicates via `Set` and filters non-strings.
- Effect cleanup in `useAssigneeCandidateState` at `:18` uses `alive` flag to prevent setState after unmount.
- `labels` tracking in IssuePage uses the same `loaded/succeeded` pattern as `projects`, and `listLabels()` rejection is properly caught.

## Verdict: APPROVED

Well-designed persistence layer with proper failure handling, workspace scoping, and input normalization. The `settle()` pattern for tracking partial API failures is a clean addition. No correctness, security, or functional concerns.

Jerry-Xin (approved):
Approving — the sole remaining blocker (clear-all leaking a stale `dateRange`) is fixed and covered by a test; all prior deep fixes remain intact.

### ✅ Blocking item resolved (was the last open 🔴)
- **clearFilters no longer leaks a stale `dateRange`.** `defaultIssueFilters()` now returns an explicit `dateRange: undefined` (`packages/dmloop/src/ui/issueFilterPersistence.ts:80`). Since `clearFilters` (`packages/dmloop/src/pages/IssuePage.tsx:361`) calls `update({ ...defaultIssueFilters() })` and `update` (`:296`) is a merge (`setF(prev => ({ ...prev, ...p }))`), the explicit `dateRange: undefined` in the spread now overwrites `prev.dateRange`, so a clear-all resets every dimension including the date range. The persist effect (`:146`) then serializes `dateRange: undefined` and, because the state now equals defaults, `writeIssueFilterState` removes the storage key — no stale range is written back. Covered by the new test `packages/dmloop/src/ui/__tests__/issueFilterPersistence.test.ts:126` asserting the defaults-merge clears `dateRange`. Credit to mochashanyao for originally surfacing this.

### 💬 Non-blocking
- On initial mount, `reload` (`packages/dmloop/src/pages/IssuePage.tsx:264`) can fire once with restored IDs before the async option reconcile (`:150`) prunes them; when reconcile actually changes `f` it recreates `reload` and re-fetches with the corrected set. This is the same pre-existing lifecycle from earlier iterations (not introduced by this change) and does not raise an error toast (empty/narrower results are not treated as failures — `onErr` fires only on network/permission/5xx). A page-level integration test covering restore → async reconcile → settled request would be a nice hardening follow-up. (non-blocking)
- Per-(workspace, view) storage keys can accumulate unboundedly over time; a lightweight prune/TTL could be a later follow-up. (non-blocking)

### ✅ Highlights
- All four earlier deep fixes remain intact and unregressed: no data loss on restore, no directory-swallow, no over-correction (`reconcileIssueFilters` returns the same reference when unchanged and only intersects when an option set is authoritative), and the same-day range boundary (`normalizeDateRange` rejects only when `start > end`, so equal endpoints are allowed).
- Cross-workspace collision fix holds: the storage key includes the workspace id (`issueFilterStorageKey`, `packages/dmloop/src/ui/issueFilterPersistence.ts`).
- Storage payload is defensively normalized via a whitelisted rebuild (no prototype pollution), serialization is wrapped in try/catch, failed option loads are treated as non-authoritative, and My Loop restrictions have focused unit coverage.

PR #1053, opened 2026-07-23
feat(loop): add task label management

### LINKED ISSUE #1052: Feature: add Loop task label management
## Problem

Loop issues need first-class label support across the main task workflows. Users need to manage workspace labels, assign multiple labels to issues, select labels when creating issues, and filter issue lists by labels without relying on ad hoc text fields or manually entered colors.

## Proposed solution

- Add workspace label management with create, edit, delete, and color selection.
- Use a predefined muted color palette instead of generated bright colors or manual RGB entry.
- Support selecting multiple labels when creating a Loop issue.
- Show and manage issue labels from the issue detail property area.
- Add label filtering to the Loop issue list with multi-select behavior.
- Keep the detail property layout readable when issues have many labels.

## Expected behavior

- Existing labels can be managed at workspace scope.
- Issues can have multiple labels.
- Users can select labels during issue creation.
- The issue detail page shows labels and opens the label management dialog when labels are clicked.
- The issue list can be filtered by one or more labels.
- Label display uses stable, muted colors and does not squeeze adjacent property labels.

## Environment

- Repository: octo-web
- Area: Loop issue management

superalsrk:
## Goal

Add first-class Loop task label management across issue creation, issue detail, workspace label administration, and list filtering.

## Changes

- Add workspace label management with muted color palette support.
- Add label selection to the Loop issue creation flow.
- Add label display and management from the issue detail property area.
- Move issue label attach/detach into the label management dialog.
- Remove per-label close controls from detail chips and make existing labels clickable.
- Add multi-select label filtering to the Loop issue list.
- Stabilize issue detail property layout so long label sets do not squeeze property names.

## Verification

- `git diff --check`

Closes #1052

### DIFF
--- a/packages/dmloop/src/i18n/en-US.json
+++ b/packages/dmloop/src/i18n/en-US.json
@@ -155,6 +155,7 @@
     "creator": "Creator",
     "project": "Project",
     "label": "Label",
+    "tag": "Tag",
     "dateRange": "Date range",
     "noProject": "Incl. no project",
     "noAssignee": "Incl. no assignee"
@@ -464,7 +465,24 @@
   "label": {
     "add": "Add labels",
     "newPlaceholder": "New label",
-    "create": "Create label"
+    "create": "Create label",
+    "manage": "Manage labels...",
+    "manageTitle": "Workspace labels",
+    "selectPlaceholder": "Select tags",
+    "color": "Label color",
+    "empty": "No labels yet",
+    "edit": "Edit label",
+    "delete": "Delete label",
+    "remove": "Remove label {{name}}",
+    "deleteConfirm": "Delete this label?",
+    "deleteConfirmDesc": "The label will be removed from every task that uses it. This cannot be undone.",
+    "created": "Label created",
+    "updated": "Label updated",
+    "deleted": "Label deleted",
+    "createFailed": "Failed to create label",
+    "updateFailed": "Failed to update label",
+    "deleteFailed": "Failed to delete label",
+    "attachPartialFailed": "Task created, but {{count}} label(s) failed to attach"
   },
   "subscribe": {
     "label": "Subscribers",
--- a/packages/dmloop/src/i18n/zh-CN.json
+++ b/packages/dmloop/src/i18n/zh-CN.json
@@ -155,6 +155,7 @@
     "creator": "创建者",
     "project": "项目",
     "label": "标签",
+    "tag": "标签",
     "dateRange": "时间范围",
     "noProject": "含无项目",
     "noAssignee": "含无负责人"
@@ -464,7 +465,24 @@
   "label": {
     "add": "添加标签",
     "newPlaceholder": "新标签名",
-    "create": "创建标签"
+    "create": "创建标签",
+    "manage": "管理标签...",
+    "manageTitle": "工作区标签",
+    "selectPlaceholder": "选择标签",
+    "color": "标签颜色",
+    "empty": "还没有标签",
+    "edit": "编辑标签",
+    "delete": "删除标签",
+    "remove": "移除标签 {{name}}",
+    "deleteConfirm": "删除这个标签？",
+    "deleteConfirmDesc": "该标签会从所有使用中的任务上移除。此操作无法撤销。",
+    "created": "标签已创建",
+    "updated": "标签已更新",
+    "deleted": "标签已删除",
+    "createFailed": "创建标签失败",
+    "updateFailed": "更新标签失败",
+    "deleteFailed": "删除标签失败",
+    "attachPartialFailed": "任务已创建，但有 {{count}} 个标签绑定失败"
   },
   "subscribe": {
     "label": "订阅者",
--- a/packages/dmloop/src/pages/IssuePage.tsx
+++ b/packages/dmloop/src/pages/IssuePage.tsx
@@ -1,4 +1,10 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import {
   Input,
   Spin,
@@ -10,7 +16,16 @@ import {
   Toast,
 } from "@douyinfe/semi-ui";
 import LoopButton from "../ui/LoopButton";
-import { Search, Plus, LayoutGrid, List as ListIcon, Users, ClipboardList, SlidersHorizontal, Filter } from "lucide-react";
+import {
+  Search,
+  Plus,
+  LayoutGrid,
+  List as ListIcon,
+  Users,
+  ClipboardList,
+  SlidersHorizontal,
+  Filter,
+} from "lucide-react";
 import { useI18n, WKApp } from "@octo/base";
 import { currentWorkspaceId, currentWorkspaceSlug } from "../api/http";
 import type {
@@ -23,7 +38,13 @@ import type {
   IssueLabel,
 } from "../api/types";
 import { ISSUE_DATE_FIELDS } from "../api/types";
-import { listIssues, searchIssues, listGroupedIssues, listMyGroupedIssues, getAgentTaskSnapshot } from "../api/issueApi";
+import {
+  listIssues,
+  searchIssues,
+  listGroupedIssues,
+  listMyGroupedIssues,
+  getAgentTaskSnapshot,
+} from "../api/issueApi";
 import { groupIssuesByAssignee } from "../api/issueGrouping";
 import { listProjectOptions } from "../api/directory";
 import { listLabels } from "../api/labelApi";
@@ -34,6 +55,7 @@ import IssueGroupBoard from "../panel/IssueGroupBoard";
 import IssueList from "../panel/IssueList";
 import IssueDetailPage from "../panel/IssueDetailPage";
 import CreateIssueModal from "../ui/CreateIssueModal";
+import LabelChips from "../ui/LabelChips";
 import { pushFleetIssueDeepLink, replaceLoopRootPath } from "../issueDeepLink";
 import { readView, writeView } from "../ui/viewMode";
 import {
@@ -50,7 +72,9 @@ type ViewMode = "board" | "grouped" | "list";
 
 // scope pill → assignee_types 过滤(看板/列表/分组统一走后端 assignee_types)。
 // all / involves 不按类型收窄(involves 走用户关系并集,见 reload)。
-function scopeToAssigneeTypes(scope: IssueScope): ("member" | "agent" | "squad")[] | undefined {
+function scopeToAssigneeTypes(
+  scope: IssueScope
+): ("member" | "agent" | "squad")[] | undefined {
   if (scope === "members") return ["member"];
   if (scope === "agents") return ["agent", "squad"];
   return undefined;
@@ -64,7 +88,10 @@ const PAGE_SIZE = 50;
 
 // 筛选/显示字段都挂在工具栏 Dropdown 面板(z-index 1060)里,其弹层默认取 Semi 基准 1030 →
 // 会被面板本体盖在下面(选项看不见也点不到)。统一抬到面板之上。
-const FIELD_POPUP = { dropdownClassName: "loop-fields__dropdown", zIndex: 2000 } as const;
+const FIELD_POPUP = {
+  dropdownClassName: "loop-fields__dropdown",
+  zIndex: 2000,
+} as const;
 
 // 「我的回路」复用本页：defaultView="grouped" + defaultScope="involves" 即只看与我相关。
 interface IssuePageProps {
@@ -74,7 +101,11 @@ interface IssuePageProps {
   viewKey?: string;
 }
 
-export default function IssuePage({ defaultScope, defaultView, viewKey }: IssuePageProps = {}) {
+export default function IssuePage({
+  defaultScope,
+  defaultView,
+  viewKey,
+}: IssuePageProps = {}) {
   const { t } = useI18n();
   // 「我的回路」= involves:只有分组视图能表达「指派/创建/关联」三路并集(listMyGroupedIssues 扇出),
   // 看板/列表无对应查询会退化成显示全工作区 → 该页锁死分组、不给切视图。
@@ -115,10 +146,14 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
   // 复用订阅特性的身份解析——候选里 octo_uid===loginInfo.uid 的 member。未解析出则「与我相关」不可用。
   const myMemberId = useMemo(() => {
     const uid = WKApp.loginInfo.uid;
-    return uid ? cands.find((c) => c.type === "member" && c.octo_uid === uid)?.id : undefined;
+    return uid
+      ? cands.find((c) => c.type === "member" && c.octo_uid === uid)?.id
+      : undefined;
   }, [cands]);
   // 项目下拉复用 directory 已缓存的 /projects(避免重复请求);随 workspace 切换整页重挂而刷新。
-  const [projects, setProjects] = useState<Array<{ id: string; title: string }>>([]);
+  const [projects, setProjects] = useState<
+    Array<{ id: string; title: string }>
+  >([]);
   const [labels, setLabels] = useState<IssueLabel[]>([]);
   const [projectsLoaded, setProjectsLoaded] = useState(false);
   const [labelsLoaded, setLabelsLoaded] = useState(false);
@@ -174,7 +209,9 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
     // 且列表勾选态仍指向陈旧行,可能被批量删改误作用。seq 守卫:仅最新一次在途请求可清。
     const onErr = () => {
       if (my !== seq.current) return;
-      setIssues([]); setGroups([]); setTotal(0);
+      setIssues([]);
+      setGroups([]);
+      setTotal(0);
       Toast.error(t("loop.toast.loadFailed"));
     };
     // 时间范围:三参数须同时给且 start<end。onChange 已把 dateRange 归一为 undefined|[起,止]。
@@ -208,29 +245,48 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
       // 独立语义(不吃 scope/其它筛选,上限 50),故与常规分组拉取分道;「我的回路」隐藏关键词、不入此路。
       if (gkw && !isMyLoop) {
         searchIssues(gkw, { limit: 50 })
-          .then(({ issues }) => { if (my === seq.current) setGroups(groupIssuesByAssignee(issues)); })
+          .then(({ issues }) => {
+            if (my === seq.current) setGroups(groupIssuesByAssignee(issues));
+          })
           .catch(onErr)
-          .finally(() => { if (my === seq.current) setLoading(false); });
+          .finally(() => {
+            if (my === seq.current) setLoading(false);
+          });
         return;
       }
       // 「与我相关」(仅「我的回路」页)= 指派给我 ∪ 我创建 ∪ 间接关联的并集扇出;需当前成员
       // 后端 id,未解析出则清空并等待(不回退成无 scope 全量)。myMemberId 在依赖里,解析后自动重取。
       if (scope === "involves") {
         if (!myMemberId) {
-          if (my === seq.current) { setGroups([]); setLoading(false); }
+          if (my === seq.current) {
+            setGroups([]);
+            setLoading(false);
+          }
           return;
         }
         listMyGroupedIssues(myMemberId, { ...common, limit: 100 })
-          .then((gs) => { if (my === seq.current) setGroups(gs); })
+          .then((gs) => {
+            if (my === seq.current) setGroups(gs);
+          })
           .catch(onErr)
-          .finally(() => { if (my === seq.current) setLoading(false); });
+          .finally(() => {
+            if (my === seq.current) setLoading(false);
+          });
         return;
       }
       // ponytail: 每组取后端上限 100；超量请用筛选。
-      listGroupedIssues({ ...common, assignee_types: scopeToAssigneeTypes(scope), limit: 100 })
-        .then((gs) => { if (my === seq.current) setGroups(gs); })
+      listGroupedIssues({
+        ...common,
+        assignee_types: scopeToAssigneeTypes(scope),
+        limit: 100,
+      })
+        .then((gs) => {
+          if (my === seq.current) setGroups(gs);
+        })
         .catch(onErr)
-        .finally(() => { if (my === seq.current) setLoading(false); });
+        .finally(() => {
+          if (my === seq.current) setLoading(false);
+        });
       return;
     }
 
@@ -238,7 +294,10 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
     const kw = f.keyword.trim();
     // 有关键词 → 走全文搜索端点(独立语义:后端不吃其它筛选/排序、上限 50);否则常规筛选列表。
     const req = kw
-      ? searchIssues(kw, { limit: paged ? PAGE_SIZE : 50, offset: paged ? page * PAGE_SIZE : 0 })
+      ? searchIssues(kw, {
+          limit: paged ? PAGE_SIZE : 50,
+          offset: paged ? page * PAGE_SIZE : 0,
+        })
       : listIssues({
           ...common,
           assignee_types: scopeToAssigneeTypes(scope),
@@ -252,13 +311,18 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
         // 删除/改状态使匹配数下降时，当前 page 可能越界（offset≥total）→ 钳到最后一页并重取。
         if (paged) {
           const maxPage = Math.max(0, Math.ceil(total / PAGE_SIZE) - 1);
-          if (page > maxPage) { setPage(maxPage); return; }
+          if (page > maxPage) {
+            setPage(maxPage);
+            return;
+          }
         }
         setIssues(issues);
         setTotal(total);
       })
       .catch(onErr)
-      .finally(() => { if (my === seq.current) setLoading(false); });
+      .finally(() => {
+        if (my === seq.current) setLoading(false);
+      });
   }, [f, view, page, scope, myMemberId, noAssigneeActive]);
 
   useEffect(reload, [reload]);
@@ -269,7 +333,16 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
   const refreshRunning = useCallback(() => {
     const my = ++runSeq.current;
     getAgentTaskSnapshot()
-      .then((tasks) => { if (my === runSeq.current) setRunning(new Set(tasks.filter((tk) => isActiveRun(tk.status) && tk.issue_id).map((tk) => tk.issue_id))); })
+      .then((tasks) => {
+        if (my === runSeq.current)
+          setRunning(
+            new Set(
+              tasks
+                .filter((tk) => isActiveRun(tk.status) && tk.issue_id)
+                .map((tk) => tk.issue_id)
+            )
+          );
+      })
       .catch(() => {});
   }, []);
   // 挂载取一次 + 每 15s 轮询:无 WS 推送,agent 起停靠轮询让 running chip 最终收敛(而非只在本地 mutation 后)。
@@ -280,12 +353,17 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
   }, [refreshRunning]);
 
   // 变更后刷新:既重取列表,又刷新运行中快照(指派/状态变更可能起/停 agent run)。
-  const onMutated = useCallback(() => { reload(); refreshRunning(); }, [reload, refreshRunning]);
+  const onMutated = useCallback(() => {
+    reload();
+    refreshRunning();
+  }, [reload, refreshRunning]);
   // 订阅 `wk:loop-issues-refresh` 重取列表:由 LoopPage 在「点击 loop 导航」与「新建回路成功」时补发,
   // 覆盖"已停在 issue tab(同 key 不重挂)"的场景。走 ref 读最新 onMutated:刷新须用当前筛选/视图,
   // 否则会用陈旧入参覆盖(reload 的 seq 守卫只防乱序)。
   const onMutatedRef = useRef(onMutated);
-  useEffect(() => { onMutatedRef.current = onMutated; }, [onMutated]);
+  useEffect(() => {
+    onMutatedRef.current = onMutated;
+  }, [onMutated]);
   useEffect(() => {
     const onRefresh = () => onMutatedRef.current();
     WKApp.mittBus.on("wk:loop-issues-refresh", onRefresh);
@@ -293,8 +371,16 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
   }, []);
 
   // 改任一筛选/搜索都回到第一页，避免停在越界的 offset（此规则只此一处表达）。
-  const update = (p: Partial<Filters>) => { setF((prev) => ({ ...prev, ...p })); setPage(0); };
-  const switchView = (v: ViewMode) => { setView(v); setPage(0); if (viewKey) writeView(viewKey, v); setShowOpen(false); };
+  const update = (p: Partial<Filters>) => {
+    setF((prev) => ({ ...prev, ...p }));
+    setPage(0);
+  };
+  const switchView = (v: ViewMode) => {
+    setView(v);
+    setPage(0);
+    if (viewKey) writeView(viewKey, v);
+    setShowOpen(false);
+  };
 
   // 点击 Issue → 跳转独立详情页（push 到右主栏，返回可 pop）。
   // key=id:issueId 变化即整体重挂载 → 详情页所有异步状态从零开始,结构性杜绝跨 issue 陈旧写入
@@ -322,23 +408,35 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
   // 新建回路 → 唤起统一建单弹窗(不再拉起独立 AI 页)。创建成功后刷新列表。
   const openNewLoop = () => setCreateOpen(true);
 
-  const isEmpty = view === "grouped" ? groups.every((g) => g.issues.length === 0) : total === 0;
+  const isEmpty =
+    view === "grouped"
+      ? groups.every((g) => g.issues.length === 0)
+      : total === 0;
 
   // 选项列表(多选共用,避免重复 map)。
   const statusOpts = ISSUE_STATUS_ORDER.map((s) => (
-    <Select.Option key={s} value={s}>{t(`loop.status.${s}`)}</Select.Option>
+    <Select.Option key={s} value={s}>
+      {t(`loop.status.${s}`)}
+    </Select.Option>
   ));
   const priorityOpts = PRIORITY_ORDER.map((p) => (
-    <Select.Option key={p} value={p}>{t(`loop.priority.${p}`)}</Select.Option>
+    <Select.Option key={p} value={p}>
+      {t(`loop.priority.${p}`)}
+    </Select.Option>
   ));
   const projectOpts = projects.map((p) => (
-    <Select.Option key={p.id} value={p.id}>{p.title}</Select.Option>
+    <Select.Option key={p.id} value={p.id}>
+      {p.title}
+    </Select.Option>
   ));
   const labelOpts = labels.map((l) => (
-    <Select.Option key={l.id} value={l.id}>{l.name}</Select.Option>
+    <Select.Option key={l.id} value={l.id}>
+      <LabelChips labels={[l]} />
+    </Select.Option>
   ));
 
-  const title = defaultScope === "involves" ? t("loop.nav.myloop") : t("loop.nav.issue");
+  const title =
+    defaultScope === "involves" ? t("loop.nav.myloop") : t("loop.nav.issue");
 
   // 关键词走全文搜索端点(独立语义,不吃其它筛选/排序)。搜索激活时禁用面板内其它筛选/scope,
   // 避免它们「看起来生效、实际被忽略」的误导。主看板三视图均支持;「我的回路」无关键词(搜索不吃 involves)。
@@ -355,8 +453,7 @@ export default function IssuePage({ defaultScope, defaultView, viewKey }: IssueP
         (f.creatorIds.length ? 1 : 0) +
         (f.projectIds.length || f.noProject ? 1 : 0) +
         (f.labelIds.length ? 1 : 0) +
-        (f.dateRange ? 1 : 0)) +
-    (!isMyLoop && f.keyword.trim() ? 1 : 0);
+        (f.dateRange ? 1 : 0)) + (!isMyLoop && f.keyword.trim() ? 1 : 0);
 
   const clearFilters = () =>
     update({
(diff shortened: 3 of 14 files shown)


Jerry-Xin (dismissed):
The PR is relevant and the core label CRUD, attachment, display, and query wiring are sound, but several UI synchronization and test-coverage gaps remain.

💬 Non-blocking

- 🟡 Warning — Search does not work reliably for label options. Both [IssuePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-91riqm6k/repo/packages/dmloop/src/pages/IssuePage.tsx:432) and [CreateIssueModal.tsx](/Users/mlamp/.openclaw/tmp/pr-review-91riqm6k/repo/packages/dmloop/src/ui/CreateIssueModal.tsx:341) enable Semi Select filtering while rendering `LabelChips` as JSX without supplying `Select.Option label={label.name}`. The repository already documents this Semi behavior in `SettingsPage.tsx`: JSX children are not searchable without the `label` property.

- 🟡 Warning — The issue list loads workspace labels only on mount. Label creation, rename, or deletion through the detail dialog does not refresh `IssuePage.labels`. This can leave stale names/options and even retain a deleted selected label until the page remounts. Refresh the label directory after workspace label mutations or broadcast a dedicated event. See [IssuePage.tsx](/Users/mlamp/.openclaw/tmp/pr-review-91riqm6k/repo/packages/dmloop/src/pages/IssuePage.tsx:167) and [LabelManagementModal.tsx](/Users/mlamp/.openclaw/tmp/pr-review-91riqm6k/repo/packages/dmloop/src/ui/LabelManagementModal.tsx:69).

- 🟡 Warning — `reload()` invokes `onChanged` when the dialog merely opens and whenever the workspace label list is refreshed, even if no mutation occurred. In issue detail this unnecessarily reloads the issue and parent list. Reserve `onChanged` for successful mutations or split it into workspace-label and issue-label callbacks. See [LabelManagementModal.tsx](/Users/mlamp/.openclaw/tmp/pr-review-91riqm6k/repo/packages/dmloop/src/ui/LabelManagementModal.tsx:43).

- 🔵 Suggestion — Add component tests covering label attach/detach, CRUD failures, partial attachment failure during issue creation, option filtering, and stale asynchronous responses. The added test only validates the palette helper.

- 🔵 Suggestion — Add a clearly discoverable workspace-level entry for label administration. Currently `LabelManagementModal` is only reachable through an editable issue, despite accepting an optional `issueId`.

✅ Highlights

- Label filtering is consistently propagated through list and grouped query paths.
- Partial label-attachment failures do not incorrectly report issue creation as failed.
- Invalid backend colors receive a safe fallback.
- Destructive global deletion has an explicit confirmation.
- The detail property layout now protects property names from long label sets.

Verification note: targeted tests could not run because dependencies are not installed (`vitest: command not found`).

Jerry-Xin (dismissed):
The PR implements the Loop task-label workflow (workspace-level label CRUD + issue attach/detach + palette-based color picker) with sound core wiring and no blocking defects. Label colors are validated on every sink, mutations only reflect success after the server confirms, and destructive workspace-wide deletion is gated behind an explicit confirmation.

💬 Non-blocking

- 🟡 Label option search is unreliable: both `packages/dmloop/src/pages/IssuePage.tsx:432` and `packages/dmloop/src/ui/CreateIssueModal.tsx:341` enable Semi `Select` filtering while rendering `LabelChips` as JSX children without passing `Select.Option label={label.name}`. Semi cannot filter JSX children without the `label` prop (already documented in `SettingsPage.tsx`).
- 🟡 Stale workspace label list: `IssuePage` loads labels only on mount (`packages/dmloop/src/pages/IssuePage.tsx:167`). Creating/renaming/deleting a label via the detail dialog (`packages/dmloop/src/ui/LabelManagementModal.tsx:69`) does not refresh `IssuePage.labels`, so option names/selections can go stale (a deleted selected label may persist until remount). Refresh the label directory after workspace-label mutations or broadcast a dedicated event.
- 🟡 `reload()` fires `onChanged` on mere dialog open and on every label-list refresh even when nothing mutated (`packages/dmloop/src/ui/LabelManagementModal.tsx:43`), triggering unnecessary issue/parent-list reloads in issue detail. Reserve the callback for successful mutations, or split workspace-label vs issue-label callbacks.
- 🔵 Partial-attach failures during issue creation are collapsed to booleans (`packages/dmloop/src/ui/CreateIssueModal.tsx:188`), losing diagnostics; the success toast can also appear alongside the partial-failure toast.
- 🔵 Add component tests for attach/detach, CRUD failures, partial-attach failure during issue creation, option filtering, and stale async responses. Current test only covers the palette helper (`packages/dmloop/src/ui/__tests__/labelPalette.test.ts`).
- 🔵 Much of the CSS delta is unrelated reformatting, which obscures the functional review; keep future feature diffs focused.

✅ Highlights

- Label colors validated (`^#[0-9a-fA-F]{6}$`) at both the picker (palette-constrained) and the render sink (`LabelChips`), with a neutral-grey fallback in every `color-mix`; label names render as escaped React text — no XSS/CSS-injection vector.
- Attach/detach update local state only after the API call resolves; failures surface via Toast.
- Destructive workspace-wide label deletion requires explicit confirmation and clearly warns it removes the label from every task.
- Filtering, creation, and detail management reuse the existing label API boundary; no dead code or dangling refs.

Jerry-Xin (dismissed):
The label CRUD, attach/detach, and color handling are solid, but the advertised workspace-level label administration is not actually reachable, and there is a dead i18n key.

### 🔴 Blocking

- **Workspace label administration has no independent entry point.** `LabelManagementModal` supports an optional workspace-only mode (`issueId?` optional at `packages/dmloop/src/ui/LabelManagementModal.tsx:18-30`), but it is instantiated only by `LabelEditor` (`packages/dmloop/src/ui/LabelEditor.tsx:45`), which always passes an `issueId` and is reachable only from issue detail. The `loop.label.manage` string ("Manage labels...") added at `packages/dmloop/src/i18n/en-US.json:19` / `packages/dmloop/src/i18n/zh-CN.json:57` is never referenced anywhere — dead key. For a PR titled "add task label management", add the intended workspace/settings entry so labels can be administered independently of an existing task, and wire up (or drop) `loop.label.manage`.

### 💬 Non-blocking

- **Label select options have no textual label for filtering.** The label options render only a `<LabelChips>` JSX child with no textual `label` prop at `packages/dmloop/src/pages/IssuePage.tsx:432` and `packages/dmloop/src/ui/CreateIssueModal.tsx:351`. Semi's text-based filter/search may therefore fail to match label names. Pass `label={label.name}` on each option, consistent with existing usage elsewhere.
- **Opening the modal fires `onChanged` even when nothing changed.** `reload()` calls `onChanged?.(rows)` at `packages/dmloop/src/ui/LabelManagementModal.tsx:43-49`, and it runs on every open. From issue detail this triggers unnecessary issue/list refreshes. Notify only after a real mutation.
- **Test coverage is thin.** Only palette normalization is covered (`packages/dmloop/src/ui/__tests__/labelPalette.test.ts`). Add interaction coverage for CRUD, attach/detach, partial attach failure during issue creation, and label-name filtering.
- **New raw CSS values instead of tokens.** The new label components add many raw color/spacing/radius literals in `packages/dmloop/src/ui/loopControls.css` and `packages/dmloop/src/pages/loop.css`. Prefer existing semantic/component tokens where available.
- **Pre-existing note (out of this diff):** label API paths in `packages/dmloop/src/api/labelApi.ts` interpolate ids into request paths without `encodeURIComponent`. Not introduced here (file unchanged), and ids are server-issued UUIDs, but worth hardening in a follow-up.

### ✅ Highlights

- Color is strictly validated everywhere it reaches a style sink: `normalizeLabelColor` uses `^#[0-9a-fA-F]{6}$` and `LabelChips` re-validates hex before setting the `--loop-chip-color` CSS var (`packages/dmloop/src/ui/labelPalette.ts:18`, `packages/dmloop/src/ui/LabelChips.tsx:33`), and the picker is palette-constrained. Label names render as escaped React text.
- Mutations are server-confirmed (create/rename/delete await then reload), guarded by a `busy` flag, and failures surface via toasts — no premature-success optimistic UI.
- Partial label-attachment failures during issue creation preserve the created task and report the failed count (`packages/dmloop/src/ui/CreateIssueModal.tsx:2141`).
- Delete is behind a Popconfirm that states the label is removed from every task and cannot be undone.

Jerry-Xin (dismissed):
Re-review at head 0bc57d4b — both prior blockers resolved; approving.

**Delta since prior review (c18211c0):** fast-forward, ahead 1 / behind 0, merge-base = c18211c0 (no rebase). Single file changed: `packages/dmloop/src/ui/LabelEditor.tsx` — the label picker is reworked into a click Dropdown that lists all workspace labels (attach/detach toggle) plus a divider and a "Manage labels…" item that opens the management modal.

## 🔴 Prior blockers — both FIXED
- **Workspace label administration now has a reachable entry point.** The new Dropdown in `LabelEditor.tsx:64-68` adds a "Manage labels…" item that opens `LabelManagementModal`. The modal renders workspace-level CRUD (create / edit / delete with Popconfirm) unconditionally in `LabelManagementModal.tsx` — only the per-issue attach toggle (`toggleIssueLabel`) is gated on `issueId`. So workspace label administration promised by the PR title is now functionally reachable through a real UI entry point. The `issueId?`-omitted mode remains an unused optional code path, but there is no functional gap.
- **Dead i18n key resolved.** `loop.label.manage` (en "Manage labels…", zh "管理标签…") is now referenced at `LabelEditor.tsx:67` via `t("loop.label.manage")`. Confirmed the only remaining `label.manage*` references are this and `manageTitle`; no dead key.

## No regression on established-clean surface
Compare c18211c0…0bc57d4b touches only `LabelEditor.tsx`; every other file (management modal, color palette/validation, label API, i18n CRUD strings, IssuePage, CreateIssueModal) is byte-identical. XSS-safe label rendering, color validation, delete-in-use Popconfirm + cascade, and optimistic-UI awaits are unchanged. New Dropdown consumes server-provided `label.id` values only (no untrusted DOM sink); `busy` guard prevents double submit; new `loop-label-option` / `loop-label-option--manage` classes resolve in `loopControls.css:386,395`.

## mergeable
`MERGEABLE` (mergeStateStatus BLOCKED = awaiting required review/approval, not a conflict). Prior CONFLICTING state resolved.

## 💬 Non-blocking
- `LabelEditor.tsx` `onVisibleChange` calls `loadAll()` on every open — a small refetch each time the picker opens; fine for now, could cache/debounce later.
- Carryover from prior notes: searchable tag `Select.Option` could set an explicit `label=`; workspace label list can go stale after mutation elsewhere; tests remain thin.

## ✅ Highlights
- Clean, minimal delta that turns per-issue attach and workspace administration into one intuitive dropdown.
- Reuses the validated color palette and XSS-safe chip rendering; no new sinks.
- Both prior blockers addressed without collateral changes to the established-clean surface.

Jerry-Xin (approved):
Re-review at head 974187ee — both prior blockers resolved and prior 🟡 notes addressed; approving.

**Delta since prior REQUEST_CHANGES (c18211c0):** fast-forward, ahead 4 / behind 0, merge-base = cd591eaf (no rebase). Two commits: `0bc57d4b` reworks the label picker into a click Dropdown, and `974187ee` addresses review feedback across 3 files.

## 🔴 Prior blockers — both FIXED
- **Workspace label administration now has a reachable entry point.** The click Dropdown in `packages/dmloop/src/ui/LabelEditor.tsx:62-68` adds a "Manage labels…" item that opens `LabelManagementModal`. The modal renders workspace-level CRUD (create / rename / delete with Popconfirm) unconditionally in `packages/dmloop/src/ui/LabelManagementModal.tsx`; only the per-issue attach toggle (`toggleIssueLabel`) is gated on `issueId`. The workspace admin promised by the PR title is now functionally reachable through a real UI entry point.
- **Dead i18n key resolved.** `loop.label.manage` (en "Manage labels...", zh "管理标签...") is now referenced at `packages/dmloop/src/ui/LabelEditor.tsx:67` via `t("loop.label.manage")`. Remaining `label.manage*` keys are exactly this and `manageTitle` (modal title) — no dead key.

## Incremental commit 974187ee — clean, addresses prior 🟡
- `packages/dmloop/src/pages/IssuePage.tsx:433` and `packages/dmloop/src/ui/CreateIssueModal.tsx:352` now set `label={l.name}` on the searchable `Select.Option`, fixing the tag search that previously matched against React children.
- `packages/dmloop/src/ui/LabelManagementModal.tsx` gives `reload(notify=false)` a notify flag: the initial mount refetch stays silent, while create / rename / delete pass `true`, so `onChanged` fires only on real mutations instead of on every open.

## No regression on established-clean surface
XSS-safe label rendering, color validation (`^#[0-9a-fA-F]{6}$` / `normalizeLabelColor` / palette), delete-in-use Popconfirm + cascade + reload, optimistic-UI awaits, and partial-attach counting are unchanged. `labelApi.ts` is byte-identical to main. New Dropdown consumes server-provided `label.id` only (no untrusted sink); `busy` guard prevents double submit; `loop-label-option*` classes resolve in `loopControls.css`.

## mergeable
`MERGEABLE` — prior CONFLICTING state resolved. CI green (e2e-p0 still running at review time).

## 💬 Non-blocking
- Workspace-wide label create / rename / DELETE is offered to anyone who can open the picker; the destructive path relies on server-side authorization in the (out-of-scope, unchanged) label API. Worth confirming the backend gates workspace label CRUD to authorized roles.
- `LabelEditor.tsx` `onVisibleChange` calls `loadAll()` on every open — a small refetch each time; fine for now, could cache/debounce.
- Carryover: workspace label list can still go stale after mutation from another surface; tests remain thin (only `labelPalette.test.ts`).

## ✅ Highlights
- Clean, minimal delta turning per-issue attach and workspace administration into one intuitive dropdown.
- Reuses the validated color palette and XSS-safe chip rendering; no new sinks.
- Both prior blockers and all prior 🟡 notes addressed without collateral changes to the established-clean surface.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1053 (octo-web)

## Summary

This PR adds a complete task label management feature to the Loop module: a new `LabelManagementModal` for workspace-level CRUD (create, edit, delete labels with color picker), inline label attach/detach from issue detail view, label selection during issue creation, colored label chips rendered across board/list/detail views, and a label filter in the issue toolbar. The implementation spans 14 files (4 new, 10 modified) with i18n for both en-US and zh-CN.

The code is well-structured with proper API error handling, i18n coverage, accessibility attributes (aria-labels, radiogroup roles), and consistent use of existing shared components (`Select`, `Popconfirm`, `LabelChips`). No P0/P1 blockers found. Two P2 items and one nit below.

## Verification

- ✅ **API contract alignment** — `labelApi.ts` exports (`createLabel`, `updateLabel`, `deleteLabel`, `listLabels`, `attachLabel`, `detachLabel`) all match the call signatures used in `LabelManagementModal`, `CreateIssueModal`, and `LabelEditor`. `IssueLabel` type (`id`, `name`, `color`) is consistently consumed.
- ✅ **Color validation** — `normalizeLabelColor` enforces `#RRGGBB` hex at both the palette picker and the chip renderer. `LabelChips` independently re-validates hex before setting `--loop-chip-color` CSS variable. Backend contract (hex-only) is preserved.
- ✅ **i18n parity** — All 17 new i18n keys in `loop.label.*` are present in both `en-US.json` and `zh-CN.json` with matching structure.
- ✅ **Async lifecycle** — `busy` guard in `LabelManagementModal` prevents concurrent CRUD operations. `seq` guard pattern in `IssuePage` correctly handles stale responses. `alive` flag in `CreateIssueModal` prevents state updates after unmount.
- ✅ **Label filter wiring** — `label_ids` is correctly included in the `common` query params (`IssuePage.tsx:236`) and sent to all three view paths (grouped, list, board).
- ✅ **Data flow: IssueDetailPage label refresh** — `LabelEditor.onChanged` triggers `syncIssue(reqRef.current)` to re-fetch the individual issue (which includes `labels`), then `onChanged()` which cascades to `onMutated` in `IssuePage` (reloads issue list). The detail view's label display stays current.

## Findings

No P0/P1 issues; two P2 items and one nit below.

### P2 — Stale workspace-label list after CRUD in detail modal (`packages/dmloop/src/pages/IssuePage.tsx:156`)

`IssuePage` fetches workspace labels once per workspace change (guarded by `labelsLoaded` at `:159`) and caches them for the filter dropdown. When a user creates, updates, or deletes labels via `LabelManagementModal` (opened from `IssueDetailPage` → `LabelEditor`), the `onMutated` callback (`:353`) calls `reload()` (issue list) + `refreshRunning()` — but does not invalidate or re-fetch the workspace label list.

Result: a newly created label won't appear in the IssuePage filter dropdown until a page refresh or workspace switch. The same staleness applies to `CreateIssueModal`'s label selector (`CreateIssueModal.tsx:97`), which loads labels once on `visible`.

Fix direction: have `onMutated` also invalidate the labels cache (e.g. `setLabelsLoaded(false)`), or emit a `wk:loop-labels-refresh` event from `LabelManagementModal` that `IssuePage` subscribes to — mirroring the existing `wk:loop-issues-refresh` pattern at `:365`.

### P2 — Label attachments are fire-and-forget after issue creation (`packages/dmloop/src/ui/CreateIssueModal.tsx:185`)

After `createIssue` succeeds, label attachments run as `Promise.all` with individual `.then(() => true, () => false)` catches. If any attachment fails, the issue is already persisted without the selected labels, a partial-failure toast fires, and the modal closes with `onCreated()`. The user has no retry path for the failed attachments.

Acceptable for initial ship (issues are independently valuable), but consider persisting the intended `labelIds` alongside the toast, or routing through the `LabelEditor` on the freshly-created issue so the user can re-attach without losing context.

### Nit — Filter dropdown label changed from "Label" to "Tag" in English (`packages/dmloop/src/pages/IssuePage.tsx:538`)

The i18n key changed from `loop.filter.label` to `loop.filter.tag`, rendering as "Tag" in English but still "标签" in Chinese. The rest of the UI (`LabelManagementModal`, `LabelEditor`, `CreateIssueModal`, detail view) consistently uses "Label" terminology (`loop.label.*` keys). Minor terminology inconsistency — worth aligning if the rename wasn't intentional.

- `LabelManagementModal` correctly resets all editing state (`name`, `color`, `editingId`, `localAttachedIds`) when `visible` toggles to true.
- `editingId` is cleared when the edited label is deleted (`LabelManagementModal.tsx:119`), preventing orphaned edit UI.
- `Popconfirm` for label deletion works independently of the `busy` state.
- `LabelColorPicker` renders a proper radiogroup with `aria-checked` and keyboard-accessible buttons.
- `LabelChips` correctly handles null/invalid colors with CSS variable fallback.
- `labelPalette.ts` enforces strict `#RRGGBB` validation matching the backend contract.
- `LabelEditor` properly delegates CRUD to `LabelManagementModal` instead of inline creation, removing the old hardcoded `#8B5CF6` default color.
- CSS additions for label chips, color swatches, and management modal layout are scoped and don't leak.
- Test file `labelPalette.test.ts` covers palette length, hex format, and `normalizeLabelColor` edge cases.

## Verdict: APPROVED

Solid feature implementation with clean component architecture, proper error handling, and i18n coverage. The two P2 items (stale label cache, no retry for partial attach failures) are non-blocking improvements. The codebase's existing patterns (seq guards, busy locks, alive flags) are correctly applied throughout.

[Octo-Q] verdict: APPROVE — no P1 blockers; two P2 improvements and one terminology nit that can be addressed in follow-up.

cnwenf (dismissed):
# Code Review — PR #1053 (Mininglamp-OSS/octo-web)

> Consolidated from three independent reviews (octo-review-qwen3.7, octo-review-glm5.2, octo-review-qwen3.8). Findings are deduplicated; `[source: ...]` marks which reviewer(s) raised each finding. The three full raw reviews remain as comments on the tracking issue for audit.
>
> ⚠️ **Head moved after review**: all three reviews evaluated head `aed64d304e484dcf8635f5703b07205683a24a6e`. The PR head is now `974187eec3e8019c97f415fcbc920c95e5b9575c` (5 commits pushed after review, including "BugFix: address label review feedback"). All findings below describe `aed64d30…`; a follow-up review pass will re-evaluate the new head.

## Summary

This PR adds end-to-end task label management to the dmloop panel: a `LabelManagementModal` for workspace label CRUD (create/update/delete/attach/detach), label selection during issue creation in `CreateIssueModal`, label display and editing in `IssueDetailPage`, a multi-select label filter on `IssuePage`, a `labelPalette.ts` color normalizer with a 10-color palette, plus new i18n keys and CSS. All three reviewers independently found the security posture clean (label names React-escaped, colors hex-anchored before any CSS sink), build/unit/i18n checks green at the reviewed head, and no P0/P1 blockers. Consolidation surfaces six P2 items — four treated as must-fix (a removed unmount-guard regression, stale `onChanged` propagation after modal unmount, a never-refreshed label filter, and the non-atomic create-then-attach recoverability gap, the last raised by all three reviewers) — and ten P3 consistency/dead-code/test-coverage nits.

## Findings

### P2 — Non-atomic create-then-attach: partial failure leaves the issue incomplete with no recovery (packages/dmloop/src/ui/CreateIssueModal.tsx:188) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

Create-then-attach is non-atomic and the partial-failure path leaves the issue silently incomplete with no recovery affordance. Driving the REAL `submit()` under jsdom (api + semi-ui mocked; submit logic 100% real): with `labelIds=[l1,l2]` selected and `attachLabel` rejecting for `l2` only (a realistic non-transient "404 label not found"), `createIssue` is called exactly once, `attachLabel` twice (`l2` not retried), and exactly one `Toast.error` fires with the count from the real bundle ("Task created, but 1 label(s) failed to attach") — but the toast is COUNT-ONLY and does not name the failed label; `onClose` and `onCreated` fire regardless and `onCreated` carries NO incompleteness signal; the modal unmounts and discards the user's selection with no retry/re-attach offered. Recovery is purely manual. The non-atomic create+attach is the documented design and the failure accounting is correct — the confirmed defect is the recoverability/feedback gap. Fix (author-discretionary): on partial failure keep the modal open / surface the failed labels with a re-attach action, or pass an incompleteness flag through `onCreated`.

Also raised by the other two reviewers: the sibling attachment-upload block at `:159-176` early-returns on failure to prevent issue creation, so the two blocks handle the same partial-failure shape with opposite policies (diff-causal — the label block was added by this PR); once the modal closes there is no way to identify which labels failed, and the path is untested (removing the Toast leaves all 81 tests green).

### P2 — Async fetch effects lack an unmount guard → setState/`onChanged` after unmount (packages/dmloop/src/ui/LabelManagementModal.tsx:56) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

Two async fetch effects have no unmount/alive guard, so a late-resolving promise calls `setState` on a torn-down component — and in the modal also propagates stale rows upward. (1) `LabelManagementModal.tsx:56` — the `[visible]` effect fires `reload()` with NO cleanup; rendering the real modal (React 17, jsdom) with a deferred `listLabels()`, unmounting mid-flight, then resolving, `onChanged?.(rows)` fired exactly once WITH the stale rows AFTER unmount and React 17 logged the "Can't perform a React state update on an unmounted component … memory leak" warning. Those stale rows flow up `LabelEditor`→`IssueDetailPage`'s `syncIssue` chain and can overwrite fresh data from a newer issue; `CreateIssueModal` in the same diff uses the correct `let alive = true … return () => { alive = false }` guard this modal omits. (2) `packages/dmloop/src/pages/IssuePage.tsx:165` — this PR REMOVED an existing guard: at pr-base the label/project fetch effect had the `alive` guard + cleanup, and HEAD rewrote it to `listLabels().then(setLabels)` with no cleanup — a regression, confirmed the same way (hold `listLabels` in-flight, unmount, resolve → same warning). Fix: restore/add the alive-guard cleanup to both effects; the modal's is higher-risk because of the `onChanged` stale-data propagation.

(severity split: qwen3.7=P2, qwen3.8=P2, glm5.2=P3-non-blocking — glm5.2 argued React 18 silently drops post-unmount `setState`; qwen3.8 executably confirmed under this repo's React 17 that `onChanged` still propagates stale rows into the still-mounted parent chain.)

### P2 — IssuePage label filter is never refreshed → stale dropdown (packages/dmloop/src/pages/IssuePage.tsx:165) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

`IssuePage` fetches `listLabels()` once on mount (`useEffect` with `[]` deps) and has no path to re-fetch. Rendering the real `IssuePage` (panels stubbed, APIs mocked): after mount `listLabels` is called once; after the backend label list changes (a label created/edited/deleted via the NEW `LabelManagementModal`) and the user-reachable `wk:loop-issues-refresh` fires, `listIssues` IS re-called but `listLabels` stays at 1 — labels are never refreshed. The filter dropdown (`labelOpts`) keeps a stale list until remount: a newly created label won't appear, a deleted label stays selectable (selecting it sends a `label_ids` filter for a nonexistent label → likely empty results), and an edited name/color won't update. Create-staleness pre-exists at pr-base, but this PR's new modal newly makes update/delete staleness reachable (amplified-by-PR). Fix: re-fetch labels on the same trigger that re-fetches the issue list.

### P2 — readOnly empty-labels cell shows an action verb instead of a neutral state (packages/dmloop/src/panel/IssueDetailPage.tsx:1241) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

readOnly mode renders actionable text `t("loop.label.add")` ("Add labels" / "添加标签") when an issue has no labels, but a readOnly user cannot act on it. Rendering the real `IssueDetailPage` under jsdom with the real en-US bundle: the readOnly empty-labels cell renders "Add labels" as INERT text (a `<span>` with no `button`/`a`/`[role=button]`, no `LabelEditor`, no `onclick`), while the sibling readOnly empty cells render NEUTRAL descriptors (Assignee "Unassigned", Project "No project"). A purpose-built neutral key `loop.label.empty` ("No labels yet" / "还没有标签") already exists and is used elsewhere; git diff-causality confirms the readOnly labels row is introduced by this PR. This is a wording/consistency contract breach, not a functional defect — the cell is inert. Fix (author-discretionary): use `loop.label.empty` in the readOnly empty branch.

### P2 — CI never runs the new vitest suite (.github/workflows/ci.yml) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The new `labelPalette.test.ts` imports the real `labelPalette` module and is effective when run (mutation-verified), but `.github/workflows/ci.yml`'s build job runs only `pnpm install --frozen-lockfile`, `pnpm i18n:check`, `pnpm build`, `pnpm lint`, and `pnpm lint:wkmodal` — there is no `pnpm test` / `turbo run test` / `vitest` step in any required workflow, and `e2e.yml`/`e2e-nightly.yml` run only Playwright `@p0`. A breaking change to `labelPalette.ts` (removing the fallback, changing palette length) would merge green with zero CI signal. Add a `pnpm test` step to the `ci.yml` build job so the suite gates merges.

(note: octo-review-qwen3.8 verified `ci.yml` is byte-identical at pr-base — this gap is a pre-existing repo convention, not introduced by this PR; it affects all dmloop tests, not just the ones added here.)

### P2 — No tests for the new async label flows (packages/dmloop/src/ui/LabelManagementModal.tsx:1) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`labelPalette.test.ts` is the ONLY test this PR adds; it is effective (mutation protocol kills 3/3 mutants), but the higher-risk async dataflow has zero coverage. Confirmed by mutation: (a) `LabelManagementModal` CRUD + attach/detach (create `:69-84`, update `:92-108`, delete `:110-123`, toggle `:125-142`) — replacing `await createLabel(...)` with `await Promise.resolve()` leaves the full existing suite green (18 files / 81 tests) while a focused test kills the mutant; (b) `CreateIssueModal` partial-failure (`packages/dmloop/src/ui/CreateIssueModal.tsx:188-201`) — changing `if (failed)` to `if (false)` (suppressing the toast) leaves the suite green while a focused test kills it. Both surfaces are brand-new PR code. A full modal render is partly infra-blocked (react@17 vs react-dom@19 / jsdom-tiptap), but the `create()` closure and the partial-failure path were each validated hook-level. Coverage gap, not a runtime defect — recommended, not merge-blocking.

(severity split: qwen3.7=P2, qwen3.8=P2-recommended, glm5.2=nit — all three confirmed via mutation that deleting the modal's CRUD error handling or the partial-failure toast leaves the suite green.)

### P3 — LabelChips duplicates the palette regex and is untested (packages/dmloop/src/ui/LabelChips.tsx:33) [source: octo-review-glm5.2 + octo-review-qwen3.8]

`LabelChips.tsx:33` keeps its own inline `/^#[0-9a-fA-F]{6}$/` copy of the color contract instead of reusing `normalizeLabelColor` from `labelPalette.ts:18` (the documented single source of truth), so if the hex shape ever changes `LabelChips` will silently diverge from the picker and normalizer — diff-causal, since both the canonical validator and the duplicated regex were introduced here. The fallback is untested: relaxing the regex to `{3,6}` leaves all 81 tests green because no test renders `LabelChips` or asserts `style['--loop-chip-color']` is set for valid hex and omitted for `'red'`/`'657C9A'`. Have `LabelChips` call `normalizeLabelColor(l.color)` and add a render test asserting the custom property is present/absent. (octo-review-qwen3.8's mutation probe separately confirmed the `LabelColorPicker.tsx:14` production route is equally unpinned.)

### P3 — `filter.label` orphaned by the filter.label→filter.tag rename; English heading changed Label→Tag (packages/dmloop/src/i18n/en-US.json:158) [source: octo-review-glm5.2 + octo-review-qwen3.8]

The label-filter key was renamed `filter.label` → `filter.tag` in `IssuePage`, but the old `filter.label` ("Label" / "标签") remains in both locale files, now orphaned, and the visible English heading changes "Label" → "Tag". Confirmed by a structured `JSON.parse` of both locale files + a `t()`-usage scan: contract "no orphaned `loop.filter.*` keys" fails with `orphans=["filter.label"]`, while siblings (`filter.status`, `filter.tag`) are correctly detected (scanner not trivially empty). At pr-base the filter used `loop.filter.label` and `filter.tag` did not exist. If the rename is intentional, delete the orphaned `filter.label` (en + zh); if not, the heading silently changed for English users. (glm5.2 adds: in `zh-CN` both keys render the identical `标签`, making it ambiguous which is canonical — remove the orphan so a future editor doesn't re-wire the dead key thinking `tag` is the alias.)

### P3 — Orphaned definitions left behind by a removed manage/remove UI (packages/dmloop/src/i18n/en-US.json:470) [source: octo-review-glm5.2 + octo-review-qwen3.8]

Two newly-added definitions have zero consumers. (1) i18n: `label.manage` ("Manage labels...") and `label.remove` ("Remove label {{name}}") are defined in en-US and zh-CN but referenced by no `t()` — confirmed by a structured locale parse + usage scan (consumed siblings detected; no dynamic template; absent at pr-base); `label.remove`'s `{{name}}` implies a per-label remove control the PR removed. (2) CSS: `.loop-label-option` and `.loop-label-option--manage` in `packages/dmloop/src/ui/loopControls.css:340` are referenced by no TSX — confirmed by a CSS-selector parse + token scan (consumed siblings detected; absent at pr-base). Dead surface that will rot. Fix: remove the unused keys (en + zh) and the two selectors, or wire them.

### P3 — LabelColorPicker case-folding divergence and untested active state (packages/dmloop/src/ui/LabelColorPicker.tsx:18) [source: octo-review-glm5.2]

At `LabelColorPicker.tsx:18` the active swatch is chosen with `selected.toLowerCase() === color.toLowerCase()`, but `normalizeLabelColor` does not case-fold and `LabelChips` uses the raw `l.color`, so the three palette consumers apply three different casing strategies for the same `#RRGGBB` shape — no visual break today (CSS hex is case-insensitive) but a drift hazard if a future consumer compares strings literally. The active-state/`aria-checked` logic is also untested: dropping `toLowerCase` leaves all 81 tests green. Add an integration test rendering the picker with `value='#6b7280'` and asserting the `#6B7280` swatch becomes `aria-checked`.

### P3 — Shared CSS rule regresses sibling inline rows (packages/dmloop/src/panel/issueDetail.css:519) [source: octo-review-glm5.2]

The new `.loop-idp__prop-k { flex: 0 0 72px; width: 72px; white-space: nowrap }` (`:519`) and the switch of `.loop-idp__prop--inline` from `space-between/center` to `flex-start/flex-start` (`:513`) are shared by all eight inline property rows in `IssueDetailPage.tsx` (assignee `:1140`, status `:1169`, project `:1198`, parent `:1210`, labels `:1238`, dates `:1264`, subscribers `:1273`), not just the new labels row — E2E only verified the labels row. With `nowrap` and a hard 72px basis, any localized key label wider than 72px in a sibling row would clip (no overflow handling declared), and the justification change affects every inline row. Scope a labels-specific override or relax the shared rule.

### P3 — Normalization helper tested, production route untested (packages/dmloop/src/ui/__tests__/labelPalette.test.ts:8) [source: octo-review-qwen3.8]

The new test covers only the `normalizeLabelColor` helper and `LABEL_PALETTE` — no production entry is pinned. Confirmed by mutation: changing `LabelColorPicker.tsx:14` `const selected = normalizeLabelColor(value)` to `const selected = value` leaves the entire existing suite green while a focused test of the real `LabelColorPicker` fails (invalid `'red'` matches 0 swatches instead of falling back to `DEFAULT_LABEL_COLOR`). `LabelManagementModal.tsx:89` also normalizes and `LabelChips.tsx:33` duplicates the hex regex instead of reusing `labelPalette`. Helper tested, production route untested — recommended, not merge-blocking.

### P3 — Palette/test uppercase-only shape diverges from the case-insensitive normalizer; lowercase boundary unpinned (packages/dmloop/src/ui/__tests__/labelPalette.test.ts:12) [source: octo-review-qwen3.8]

Production `normalizeLabelColor` uses `/^#[0-9a-fA-F]{6}$/` (case-insensitive) and passes lowercase through UN-uppercased — verified at HEAD (`'#abcdef'` → `'#abcdef'`). But `LABEL_PALETTE` and the test assert uppercase-only (`/^#[0-9A-F]{6}$/`), and the test only asserts uppercase values. Confirmed by mutation: uppercasing the production regex (which flips `'#657c9a'` → `DEFAULT_LABEL_COLOR`) leaves the existing test green (2/2) while a focused lowercase-boundary test fails (`:18`). So a future lowercase palette entry would be accepted by production yet rejected by the test, and nothing pins the current lowercase-passthrough. Harmless today (palette only emits uppercase) — a test-quality gap. Fix: add a lowercase assertion or reconcile the test regex with production.

### P3 — `remove()` leaves a dangling id in `localAttachedIds` → phantom attachment (packages/dmloop/src/ui/LabelManagementModal.tsx:107) [source: octo-review-qwen3.8]

`remove()` deletes a label via `deleteLabel(label.id)` but does not prune the id from `localAttachedIds`. On the real component: after `remove(X)` the row disappears but the id stays; re-introducing a label with the SAME id renders it `is-active` again — a phantom produced solely by the dangling id. Bounded impact: a deleted id is not normally reused and the staleness self-corrects on modal reopen, so this is a transient cosmetic inconsistency, not data corruption. Fix: delete the id from `localAttachedIds` in `remove()` on success.

### P3 — `reload()` is not deduped → concurrent reloads (packages/dmloop/src/ui/LabelManagementModal.tsx:43) [source: octo-review-qwen3.8]

The `busy` flag serializes create/saveEdit/remove/toggle but `reload()` is unguarded, so two reloads can overlap. Confirmed deterministically with deferred promises: while the mount reload is in-flight, a `create()` issues a SECOND reload (`listLabels` called twice, neither resolved); resolving both fires `onChanged` TWICE (an extra parent `syncIssue`) and `loading` flickers. Low impact (latest data wins) but the race is real and newly introduced. Fix: guard `reload()` with an in-flight flag.

### P3 — Bare hex colors outside `var()` (packages/dmloop/src/ui/loopControls.css:1) [source: octo-review-qwen3.8]

The added CSS uses many bare hex colors outside `var()`. The large majority are custom-property DEFINITIONS for a new component-scoped token system (`--loop-tag-*`), which is the legitimate place for raw hex; a few are bare property usages (e.g. `color: #fff;`) that could reference a theme token. Minor convention drift, not a defect.

### Flagged for awareness — not confirmed / not verifiable from this repo

- **`LabelManagementModal` busy-guard race** (`LabelManagementModal.tsx:126`) [source: octo-review-glm5.2]: the single shared `busy` boolean is read from the render closure, so two same-tick events could both pass the guard before `setBusy(true)` flushes. In practice every control has `disabled={busy}`/`loading={busy}`, so the practical click path is blocked after re-render; no deterministic render-test repro was possible in this repo. Flagged for awareness, not a confirmed defect.
- **Backend-boundary contracts** [source: octo-review-qwen3.8]: (a) whether octo-server rejects lowercase hex on `updateLabel` (the picker only emits uppercase, so lowercase would have to come from a backend-stored `label.color`); (b) the server-side AND/OR / empty-array interpretation of the multi-select `label_ids` filter. Both are defined by octo-server and could be neither confirmed nor falsified from octo-web — non-blocking.

- **Security (3/3 reviewers CLEARED)**: label names render only via React text interpolation / controlled inputs (auto-escaped); colors gated by an anchored `/^#[0-9a-fA-F]{6}$/` regex before entering CSS custom properties — no CSS-injection surface; no `dangerouslySetInnerHTML`/`innerHTML`/`eval`; no new deps, endpoints, or secrets; `labelApi.ts`/`http.ts` pre-existing and untouched, workspace isolation via `x-workspace-slug` unchanged — no new IDOR/auth surface. Full STRIDE + OWASP Top-10 + credential scans, 0 findings in all three reviews.
- **`toggleIssueLabel` is pessimistic, not optimistic** — `setLocalAttachedIds` runs AFTER `await attach/detach`, and on rejection the state update is unreachable; the "optimistic update without rollback" claim was independently falsified by two reviewers.
- **Partial-failure accounting is correct** — the toast count (`results.filter(!ok).length`) is accurate and the issue is never silently lost (a toast always fires); the gap is recoverability/feedback, not data loss or miscounting.
- **The added test is effective** — `labelPalette.test.ts` kills 3/3 mutants (not a rubber stamp).
- **`onChanged` arg-drop (`LabelEditor.tsx:47`) falsified as a runtime defect** — `IssueDetailPage` re-fetches the full issue via `syncIssue`, so fresh labels always arrive (maintainability trap only).
- **i18n parity** — all new `label.*` keys + `filter.tag` mirrored across en-US/zh-CN with matching `{{name}}`/`{{count}}` tokens (0 missing, 0 mismatch); `pnpm i18n:check` passes.
- **Build & types** — `pnpm build` clean; `tsc --noEmit` 0 errors in PR-changed files; full dmloop vitest suite green at the reviewed head.
- **CSS compliance** — no `!important`, no `prefers-color-scheme` violations; new `--loop-tag-*` properties follow the component-scoped token pattern.
- **Modal concurrency basics** — the `busy` flag correctly serializes create/update/delete/toggle; no dangling references to removed code, no orphaned imports.
- **Default label color change** purple (`#8B5CF6`) → grey (`#6B7280`) is a design choice, not a defect; the "Tag" vs "Label" filter wording inconsistency is cosmetic.
- **Playwright** `C4-loop-create-issue`, `C8-loop-create-issue-full`, `C9-loop-issue-detail-view` pass at the reviewed head.

## Verification Results

| Code Review | NEEDS_CHANGES | 6 P2 + 10 P3 (deduped across 3 reviewers; qwen3.8=NEEDS_CHANGES, qwen3.7/glm5.2=COMMENT) — 4 must-fix: removed unmount-guard regression, stale modal `onChanged` after unmount, never-refreshed label filter, create-then-attach recoverability gap; rest non-blocking |
| E2E Test | PARTIAL | Build, i18n parity, full vitest suite, mutation probes, and Playwright create-issue & issue-detail flows all green at reviewed head; cross-service label CRUD / `label_ids` filter semantics live in octo-server (octo-web ships no compose/MSW) and were not exercisable end-to-end |
| Security | CLEARED | Three independent STRIDE/OWASP Top-10 + credential scans, 0 findings — label names React-escaped, colors gated by anchored `#RRGGBB` regex before CSS sink, no new deps/endpoints/secrets |

PR #1059, opened 2026-07-24
fix(dmworkmcp): align MCP marketplace UI to Skill + tag chip filter (#1009)

### LINKED ISSUE #1009: [UI一致性] MCP市场页面UI需与Skill市场保持一致，且需适配大屏

l-s-c:
## Summary

Bring the MCP marketplace in line with the Skill marketplace (issue #1009), backport the tag/title overflow guards from PR #1026, and wire the new tag chip filter to the backend `/mcp_tags` endpoint. Resubmitted after the earlier PR was closed — this branch also folds in the follow-up review rounds that landed after the close: card-level owner actions (edit / delete on 我的 MCP cards, mirroring Skill), per-key token placeholder in Bot 上架 & 快速接入, hex32 Space ID acceptance, and a batch of small polish items.

Commit history splits into readable chapters:

1. **UI alignment (#1009 body)** — McpCard 2-line owner row (`🐳 bot · 👤 human` for bot rows, `👤 human` for human rows, hidden for import rows), 48px icon + hash-color fallback + 2-char initials, rounded-pill accent tags, MatchReasons chips restored for creator hits only, McpDetailModal custom WKModal header (icon + h3 + category chip + owner/updated meta), publish button → `上架 MCP` primary + Upload icon + dropdown (Bot 上架 / 手动上架), Bot 上架 prompt modal referencing octo-cli's `octo-marketplace` Skill's `mcp.md` workflow.
2. **PR #1026 backport** — MCP tag input rejects `>10` tags / `>24 chars` / invalid chars with Toast, paste splits on comma, McpCreateModal.handleSubmit re-validates on save (guards legacy oversized rows), tag chip inside input ellipses long tags.
3. **Tag chip filter** — new popover in the search bar mirroring Skill's SearchBar. Reads suggestions from `GET /mcp_tags` (debounced 160ms, AbortController-cancelled), forwards `?mode=mine` when opened from the 我的 tab, memoised `tagRows` keyed on `(tagsSelected, tagSuggestions)` identity, `focus({ preventScroll: true })` on open, resets on space + mode change. Selected tags round-trip as repeated `?tag=X` URL params so a tag literal containing a comma stays intact.
4. **Card owner actions** — pencil / trash icons on 我的 cards (mirrors SkillCard). Standalone `McpDeleteConfirmModal` (mirrors `DeleteConfirmModal`) so trash on a card short-circuits opening the detail. `handleEditFromCard` fetches the full `McpDetail` before opening the shared create/edit modal (list carries `McpListItem` without `quickStart` / `tools`). Keyboard early-return on the outer `role=\"button\"` div so Tab-then-Enter on trash doesn't also open detail.
5. **Bot 上架 copy button** unblocked in non-UUID space environments. `SPACE_ID_RE` now accepts both the canonical 36-char UUIDv4 AND the 32-hex compact form real production issues (`9f5fda183d94482cb49bca5024439105`). `sanitizeSpaceId` still substitutes `<space-id>` for injection payloads before reaching any shell example; the `promptReady` gate simplifies to `Boolean(prompt)` matching Skill's `BotPublishModal`.
6. **Per-key token placeholder** — user-supplied headers / env values used to render as the fixed literal `<把这里换成你的 Token>`. That was misleading when the field was named `12` or `X-Team-Id`. Each occurrence now carries its own key: `<把这里换成你的 Authorization>` / `<把这里换成你的 12>`. `TOKEN_PLACEHOLDER` const → `formatTokenPlaceholder(key)` + `TOKEN_PLACEHOLDER_RE` regex. `McpDetailModal` splits by regex and wraps each match in a highlight `<mark>`.
7. **Small polish** — search placeholder reworded to `搜索名称、描述...` (tags moved to the chip filter, no longer typed into the box); tag chip max-width scoped per container (card 108px / detail 168px / tooltip uncapped); card meta `.wk-mcp-card__owner` `flex: 0 1 auto` so bot + human chips shrink before overflowing; tag-filter option count dropped (visual noise; sort-by-count preserved via cached `tagRows`).

Also removed the 来源 (人工/Bot) segmented filter per product feedback — both provenance values render identically to end users.

## Linked Spec

Fixes #1009. Refs #1017 + PR #1026 (Skill tag/title overflow backport), and the sibling backend contract at [octo-marketplace fix/mcp-search-narrow](https://github.com/Mininglamp-OSS/octo-marketplace) for narrow keyword search + AND tag semantics + `/mcp_tags`.

## How verified

Local Vite dev server (`localhost:3000`) proxied to `im-test.deepminer.com.cn`; sibling marketplace API on `localhost:8092` running the paired branch.

- List renders real MCPs on both 全部 / 我的 tabs; every provenance case (bot / human / empty-icon fallback / long-title) displays correctly.
- Category chip 开发工具 → correct rows; count updates + refetch works.
- Search `linear` → correct rows; MatchReasons chip appears only when a creator hit fires.
- Tag popover: opens, search input focused, suggestions populate from `/mcp_tags`; multi-tag selection AND-combines; `+N` overflow tooltip shows every hidden chip verbatim.
- Publish dropdown: both `Bot 上架` (copy-prompt modal — button enabled on 32-hex space id; prompt contains real Space ID + API base URL) and `手动上架` (create form) paths work.
- Detail modal: `[icon] Name [category-chip]` header + `🐳 bot · 👤 human · N 天前更新` meta + slogan + `🔧 N` stats + tags at bottom; per-key placeholder `<把这里换成你的 Authorization>` / `<把这里换成你的 12>` on 快速接入 提示词 & JSON tabs.
- Create form: tag input rejects `>10` tags, `>24` char tags, non-allowlisted chars, paste-split `a,b,c` → 3 chips.
- Card owner actions on 我的 tab: pencil opens edit modal prefilled from full detail; trash opens confirm modal (independent, mirrors Skill).
- `pnpm --filter @dmwork/mcp exec vitest run` — 128 tests, all green. No new TS errors vs. baseline.

## COMPREHENSION

1. **What does this change actually do** to the load-bearing path?

   - **McpCard / McpDetailModal**: DOM + CSS refactor to match Skill layout. Same wire contract (`McpListItem`, `McpDetail`); `resolveOwner` picks the display shape per `createdByType` (bot/human/import → chip / chip / no-chip). Card renders new pencil / trash buttons only when the list page passes `onEdit` / `onDelete` (gated on `mode === \"mine\"`).
   - **TagsInput in McpCreateModal**: adds `validateMcpTags(form.tags)` guard on submit + inline per-tag validation; a legacy record with `>10` or over-length tags blocks save until fixed.
   - **List filter API**: `params.tags` serialises as repeated `?tag=X` params (was comma-joined). Backend `splitQuery` accepts both, so a mid-migration overlap is safe.
   - **`GET /mcp_tags`**: new API call fired when the tag popover opens, debounced 160ms, cancels prior in-flight via AbortController. Fails silently (empty popover) if backend isn't up.
   - **`mcpListQuery`**: URL round-trips `?tag=X` params so bookmarked filtered views hydrate cleanly.
   - **Bot 上架**: copy button now gates only on `Boolean(prompt)`; `SPACE_ID_RE` accepts both UUIDv4 and 32-hex. `sanitizeSpaceId` still runs — injection payloads still map to `<space-id>` before reaching any shell example.
   - **Token placeholder**: `formatTokenPlaceholder(key)` interpolates the header/env key. `TOKEN_PLACEHOLDER_RE` (`/<把这里换成你的 [^>]+>/g`) matches every occurrence; `[^>]+` is linear so no ReDoS. `McpDetailModal` clones the regex per render so the shared `lastIndex` never contaminates a subsequent call.
   - **Removed**: `createdByType` state + `SOURCE_FILTER_OPTIONS` + the source segmented filter UI. The wire field stays on responses; only the client-side filter is gone.

2. **What could break** because of it (dependents + failure mode)?

   - **Bot 上架 prompt correctness**: hard-codes `octo-cli marketplace mcp create` and references the `octo-marketplace` Skill's `mcp.md`. Verified against octo-cli's embedded Skill catalog. If those commands rename, `utils/mcpBotPublishPrompt.ts` needs a bump.
   - **Tag chip filter on old backend**: the sibling marketplace PR must merge + deploy before the popover fills in with real suggestions. Meanwhile the popover shows an empty state; the underlying `?tag=` filter still works if the user selects a chip from a previously-loaded state.
   - **Multi-tag URL semantics**: a saved `?tag=a&tag=b` used to return the union (frontend was joining), now returns the intersection (repeated params + backend AND). Result set shrinks; user may see \"my link stopped working\" once but the intersection is what the filter actually asked for.
   - **Space ID regex broadening**: added the 32-hex form. Both branches are `[0-9a-f]{n}` so no shell metacharacter can pass; injection tests (`; rm -rf /`, `` `whoami` ``, `$(whoami)`) still fail closed to the `<space-id>` placeholder.
   - **Token placeholder interpolation**: existing content in the wild that happened to contain the literal `<把这里换成你的 Token>` (as user data) would no longer be highlighted. Extremely unlikely in practice — the pattern is bespoke to this rendering path.
   - **Legacy tag validation UX**: a user opening an existing MCP with `>10` tags will see save blocked and a Toast until they trim; documented in the file comment.
   - **Clipboard availability**: Bot 上架 modal toasts on clipboard rejection / missing `navigator.clipboard`, so users on http:// origins or older WebViews see feedback instead of a silent no-op.

3. **How do you know it works** (specific test/repro/trace)?

   - Full E2E walkthrough against test env (see \"How verified\") — every provenance case, every filter, every modal opened and read.
   - Two cross-repo code-review passes (parallel general-purpose agents) applied all findings inline; the final pass reported no Blocking / Should-fix findings on either repo.
   - 128 unit tests green (`pnpm --filter @dmwork/mcp exec vitest run`), including new tests around: `SPACE_ID_RE` positive (both shapes) and negative (injection + length-off); `formatTokenPlaceholder` interpolation; `TOKEN_PLACEHOLDER_RE` regex against a `>`-boundary regression; regex + tests updated across `quickStartTemplates.test.ts`, `mcpBotPublishPrompt.test.ts`, `McpDetailModal.inlineDelete.test.tsx`.
   - Vite HMR compiles every touched file; TS baseline errors unchanged (pre-existing React 17/18 type resolution issues).
   - Manual DOM inspection: `resolveOwner` gates the import branch correctly; MatchReasons component only fires for `creator` hits after the sibling backend change; category badge falls back to raw slug if i18n misses; per-key placeholder correctly wraps the actual header key (`Authorization`, `X-Team-Id`, or `12`).

### DIFF
--- a/.i18n/scan-config.json
+++ b/.i18n/scan-config.json
@@ -61,6 +61,10 @@
       "reason": "Agent prompt template for the Bot publish flow. Same rationale as installPrompt.ts above — generated prompt content pasted into an agent client, must byte-match; not translatable UI chrome."
     },
     {
+      "path": "packages/dmworkmcp/src/utils/mcpBotPublishPrompt.ts",
+      "reason": "Agent prompt template for the MCP marketplace publish flow. Same rationale as dmworkskillmarket/src/utils/botPublishPrompt.ts — generated prompt content pasted into an agent client, must byte-match; not translatable UI chrome."
+    },
+    {
       "path": "packages/dmworksummary/src/__mocks__/handlers.ts",
       "reason": "MSW test fixture data with sample user names, titles, and message bodies; not UI copy."
     },
--- a/packages/dmworkmcp/package.json
+++ b/packages/dmworkmcp/package.json
@@ -11,7 +11,8 @@
     "@douyinfe/semi-icons": "^2.93.0",
     "@douyinfe/semi-ui": "^2.93.0",
     "axios": "^0.25.0",
-    "classnames": "^2.3.1"
+    "classnames": "^2.3.1",
+    "lucide-react": "^0.577.0"
   },
   "devDependencies": {
     "@testing-library/jest-dom": "^6.9.1",
--- a/packages/dmworkmcp/src/api/mcpService.paramsSerializer.test.ts
+++ b/packages/dmworkmcp/src/api/mcpService.paramsSerializer.test.ts
@@ -1,0 +1,64 @@
+import { describe, expect, it, vi } from "vitest";
+
+// serializeMcpParams is pure and doesn't call t(), but the file's module
+// graph pulls in @octo/base at import-time via other exports. Stub the
+// module so the test doesn't depend on i18n / IM SDK bootstrap.
+vi.mock("@octo/base", () => ({
+  WKApp: { apiClient: { config: {} }, loginInfo: {}, shared: { currentSpaceId: "" }, mittBus: { on: () => {}, off: () => {}, emit: () => {} } },
+  buildAcceptLanguage: () => "en-US",
+  t: (key: string) => key,
+  DEFAULT_REQUEST_TIMEOUT_MS: 20000,
+}));
+
+import { serializeMcpParams } from "./mcpService";
+
+describe("serializeMcpParams — axios paramsSerializer wire contract", () => {
+  it("emits `?a=1&a=2` for arrays (never `?a[]=1&a[]=2`)", () => {
+    // gin's QueryArray on the marketplace backend only recognises the plain
+    // repeat form — a revert to axios 0.25's default `?a[]=...` shape would
+    // make every multi-tag filter silently return wrong results. Pin.
+    expect(serializeMcpParams({ tag: ["a", "b"] })).toBe("tag=a&tag=b");
+  });
+
+  it("URL-encodes commas inside a value so comma-bearing tags survive", () => {
+    expect(serializeMcpParams({ tag: ["v1.0,beta"] })).toBe("tag=v1.0%2Cbeta");
+  });
+
+  it("preserves multiple keys and multi-value ordering", () => {
+    expect(
+      serializeMcpParams({
+        q: "hello",
+        tag: ["a", "b"],
+        mode: "mine",
+      })
+    ).toBe("q=hello&tag=a&tag=b&mode=mine");
+  });
+
+  it("drops null / undefined values entirely", () => {
+    expect(
+      serializeMcpParams({
+        q: undefined,
+        limit: null,
+        mode: "mine",
+      })
+    ).toBe("mode=mine");
+  });
+
+  it("drops null / undefined inside arrays (keeps a stable position for the rest)", () => {
+    expect(
+      serializeMcpParams({
+        tag: ["a", undefined as unknown as string, "b", null as unknown as string, "c"],
+      })
+    ).toBe("tag=a&tag=b&tag=c");
+  });
+
+  it("handles empty / undefined params argument", () => {
+    expect(serializeMcpParams(undefined)).toBe("");
+    expect(serializeMcpParams({})).toBe("");
+    expect(serializeMcpParams({ tag: [] })).toBe("");
+  });
+
+  it("stringifies non-string primitives (limit, offset)", () => {
+    expect(serializeMcpParams({ limit: 50, offset: 0 })).toBe("limit=50&offset=0");
+  });
+});
--- a/packages/dmworkmcp/src/api/mcpService.ts
+++ b/packages/dmworkmcp/src/api/mcpService.ts
@@ -134,6 +134,7 @@ async function fetchMcpListMockFiltered(
   const keyword = (params.keyword ?? "").trim().toLowerCase();
   const category = params.category ?? "all";
   const createdBy = params.createdByType;
+  const tags = params.tags ?? [];
   const filtered = source.filter((item) => {
     const matchCategory = category === "all" || item.category === category;
     const matchKeyword =
@@ -144,7 +145,11 @@ async function fetchMcpListMockFiltered(
     // read-side default the wire mapper applies for pre-#894 records.
     const rowType = item.createdByType ?? "human";
     const matchCreatedBy = !createdBy || rowType === createdBy;
-    return matchCategory && matchKeyword && matchCreatedBy;
+    // Multi-tag filter is AND: a row must carry every selected tag. Mirrors
+    // the backend semantics in octo-marketplace (mcp-v1.md §4.2).
+    const matchTags =
+      tags.length === 0 || tags.every((tag) => item.tags.includes(tag));
+    return matchCategory && matchKeyword && matchCreatedBy && matchTags;
   });
   const offset = params.offset && params.offset > 0 ? params.offset : 0;
   const limit =
@@ -675,6 +680,12 @@ async function fetchMcpListPath(
   // Relevance is only meaningful with a keyword — every row scores 0 otherwise,
   // making the sort order arbitrary. When browsing, surface freshest first.
   query.sort = keyword ? "relevance" : "updated";
+  // Multi-tag filter — mcp-v1.md §4.2 accepts `tag` as repeatable OR
+  // comma-separated. Use comma-separated for compact wire; the backend
+  // AND-combines the parsed values.
+  if (params.tags?.length) {
+    query.tag = params.tags.join(",");
+  }
   const pageSize = params.limit && params.limit > 0 ? params.limit : 20;
   query.page_size = pageSize;
   query.page = Math.floor((params.offset ?? 0) / pageSize) + 1;
@@ -876,6 +887,62 @@ export function fetchMcpMine(
 
 export function fetchMcpDetail(id: string): Promise<McpDetail> {
   return USE_MOCK ? fetchMcpDetailMock(id) : fetchMcpDetailReal(id);
+}
+
+/** One tag suggestion in the tag-filter popover (mcp-v1.md §4.8). */
+export interface McpTagSuggestion {
+  name: string;
+  count: number;
+}
+
+/** GET /market/api/v1/mcp_tags — tag suggestions for the search-bar
+ *  popover. Backend aggregates across the caller's visible set (system +
+ *  space rows), sorted by descending count. Empty `query` returns every
+ *  visible tag; the backend clamps `limit` to [1, 100] with default 50. */
+export function fetchMcpTags(
+  query = "",
+  opts: { signal?: AbortSignal; limit?: number } = {}
+): Promise<McpTagSuggestion[]> {
+  return USE_MOCK
+    ? fetchMcpTagsMock(query, opts)
+    : fetchMcpTagsReal(query, opts);
+}
+
+async function fetchMcpTagsMock(
+  query: string,
+  _opts: { signal?: AbortSignal; limit?: number }
+): Promise<McpTagSuggestion[]> {
+  // Aggregate from the in-memory mock list — same visibility semantics as
+  // the real backend (everything visible to the caller's space). Count-desc
+  // + name-asc tie-break, case-insensitive query.
+  const counts = new Map<string, number>();
+  for (const item of MOCK_MCP_LIST) {
+    for (const tag of item.tags) {
+      counts.set(tag, (counts.get(tag) ?? 0) + 1);
+    }
+  }
+  const kw = query.trim().toLowerCase();
+  const items: McpTagSuggestion[] = [];
+  counts.forEach((count, name) => {
+    if (kw && !name.toLowerCase().includes(kw)) return;
+    items.push({ name, count });
+  });
+  items.sort((a, b) => (b.count - a.count) || a.name.localeCompare(b.name));
+  return delay(items);
+}
+
+async function fetchMcpTagsReal(
+  query: string,
+  opts: { signal?: AbortSignal; limit?: number }
+): Promise<McpTagSuggestion[]> {
+  const params: Record<string, unknown> = {};
+  if (query.trim()) params.q = query.trim();
+  if (opts.limit && opts.limit > 0) params.limit = opts.limit;
+  const resp = await mcpAxios.get<{ data: McpTagSuggestion[] }>(
+    `${BASE}/mcp_tags`,
+    { params, signal: opts.signal }
+  );
+  return resp.data.data ?? [];
 }
 
 /**
--- a/packages/dmworkmcp/src/api/mcpService.ts
+++ b/packages/dmworkmcp/src/api/mcpService.ts
@@ -681,10 +681,12 @@ async function fetchMcpListPath(
   // making the sort order arbitrary. When browsing, surface freshest first.
   query.sort = keyword ? "relevance" : "updated";
   // Multi-tag filter — mcp-v1.md §4.2 accepts `tag` as repeatable OR
-  // comma-separated. Use comma-separated for compact wire; the backend
-  // AND-combines the parsed values.
+  // comma-separated. We use REPEATED params (`tag=a&tag=b&tag=c`) so a
+  // tag value that happens to contain a comma still round-trips intact;
+  // comma-joining would let the backend re-split "v1.0,beta" into two
+  // wrong tags and produce silent empty results.
   if (params.tags?.length) {
-    query.tag = params.tags.join(",");
+    query.tag = params.tags;
   }
   const pageSize = params.limit && params.limit > 0 ? params.limit : 20;
   query.page_size = pageSize;
@@ -898,10 +900,12 @@ export interface McpTagSuggestion {
 /** GET /market/api/v1/mcp_tags — tag suggestions for the search-bar
  *  popover. Backend aggregates across the caller's visible set (system +
  *  space rows), sorted by descending count. Empty `query` returns every
- *  visible tag; the backend clamps `limit` to [1, 100] with default 50. */
+ *  visible tag; the backend clamps `limit` to [1, 100] with default 50.
+ *  Pass `mode: "mine"` when the popover opens from the "我的" tab so
+ *  suggestions match `GET /mcps/mine`. */
 export function fetchMcpTags(
   query = "",
-  opts: { signal?: AbortSignal; limit?: number } = {}
+  opts: { signal?: AbortSignal; limit?: number; mode?: "all" | "mine" } = {}
 ): Promise<McpTagSuggestion[]> {
   return USE_MOCK
     ? fetchMcpTagsMock(query, opts)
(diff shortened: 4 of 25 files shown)


Jerry-Xin (approved):
Solid, careful PR — MCP marketplace UI now mirrors Skill, tag-chip filter is race-safe, and lifecycle/cleanup is handled correctly. CI is green (Build / e2e-p0 / install-build all pass) and the diff is a clean 8-commit delta with no merge-base drift. Approving with one non-blocking edge to tidy.

🔴 Blocking
None.

💬 Non-blocking
- `packages/dmworkmcp/src/components/McpCreateModal.tsx` — `commitOne` returns the same `null` sentinel for a **duplicate** tag as for invalid/limit-exceeded, and the batch `commit` loop `break`s on any `null`. So pasting/entering `existing,new-a,new-b` (where `existing` is already a chip) hits the duplicate on segment 1 → returns `null` → `break` → `new-a` and `new-b` are silently dropped. Unlike the limit/charset rejections, the duplicate path shows no Toast, so the user gets zero feedback. A duplicate is a legitimate skip, not a batch-abort condition — it should `continue` (skip that segment, keep processing the rest) rather than `break`. Narrow ordering + low-frequency create/edit form, so non-blocking, but it does drop input without feedback. Suggest either distinguishing the duplicate case (e.g. return `existing` unchanged and `continue`) or reserving `break` for the true invalid/limit rejections.

✅ Highlights
- Tag suggestion fetch is genuinely race-safe: debounced 160ms, `AbortController` cancelled on close/unmount/space-change, and both `.then` and `.catch` guard on controller identity so a stale response can't clobber a fresh one. Cancellation is classified (axios cancel + `AbortError`) and dropped silently, while real backend/network errors surface a Toast consistent with `loadMore()`.
- Global `pointerdown` / `keydown` listeners are attached only while a popover is open and removed on the open→closed transition, with idempotent removal + timer/controller cleanup in `componentWillUnmount`. No listener leaks.
- Token-placeholder highlighting in `McpDetailModal` renders via React text nodes + `<mark>` (no `dangerouslySetInnerHTML`), builds a fresh regex per render to avoid shared `lastIndex` state, and the `[^>]+` class guarantees non-empty matches so the `exec` loop always advances (no infinite loop, no XSS).
- `SPACE_ID_RE` broadening to 32-hex is safe — both branches are `[0-9a-f]{n}`, and `sanitizeSpaceId` still maps injection payloads to the `<space-id>` placeholder before any shell example.
- Multi-tag round-trips as repeated `?tag=X` params so a tag literal containing a comma stays intact; space-change resets selected tags/categories to avoid cross-space zero-result queries.

lml2468 (approved):
## Review: #1059 — fix(dmworkmcp): MCP 市场 UI 对齐 Skill + 标签 chip 过滤器(#1009)

**意图**:#1037/#1043 的重提 + 折入关闭后落地的后续轮次 —— 卡片级 owner 操作(我的 MCP 卡编辑/删除)、Bot 上架/快速接入的 per-key token 占位符、32-hex Space ID 接受、tag chip 过滤器接后端 `/mcp_tags`。

锚定 head `cedefac0` / merge-base `a8bad922`(behind 6,`mergeable=MERGEABLE`,0 冲突)。25 文件 +2601/−476。

---

### Gate 2 — 安全面(逐项字节核实,重点新增面)

- **per-key token 占位符 `<mark>` 渲染 XSS 安全**:`McpDetailModal` 用 `TOKEN_PLACEHOLDER_RE`(每 render `new RegExp` 避免 lastIndex 污染)`split` + `map` → 返回 React `<mark>{文本}</mark>` 节点,匹配文本是**React 文本子节点(自动转义)**,**非 `dangerouslySetInnerHTML`**。✅
- **Space ID 正则放宽仍 fail-closed**:`SPACE_ID_RE = /^(?:[0-9a-f]{32}|UUIDv4)$/i` —— 两分支纯 hex/连字符,任何 shell 元字符不过;`sanitizeSpaceId` 非法 → `<space-id>` 占位。`promptReady` 虽简化为 `Boolean(prompt)`,但 spaceId 已在上游被 `sanitizeSpaceId` 中和 → 注入 payload(`;rm -rf`/`$(whoami)`/反引号)进不了 prompt。✅
- **tag chip 过滤器无注入**:`serializeMcpParams` 用 `URLSearchParams.append` 逐项 → 重复 `?tag=X`(含逗号标签安全),路径固定 `/mcp_tags`,标签只进 params。✅
- **卡片 owner 操作门控**:`canManage = mode === "mine"`(:543)→ `onEdit`/`onDelete` 仅「我的」tab 传入;删除走独立 `McpDeleteConfirmModal`。客户端 defense-in-display,后端授权真实删除。✅
- **owner 归属**:显式 `createdByType`(bot/human/import),非 #1007 过宽启发。✅
- **标签校验**:`validateMcpTags` 提交 re-validate(legacy 超量阻断);`TOKEN_PLACEHOLDER_RE` `[^>]+` 线性无 ReDoS。✅

### 构建 / 测试(实跑)
- `pnpm --filter @octo/web build` ✅ built in 4.09s。
- `pnpm --filter @dmwork/mcp test` ✅ **128 tests 全绿**(含 SPACE_ID_RE 正/负例、formatTokenPlaceholder、TOKEN_PLACEHOLDER_RE `>`-边界回归)。
- i18n parity:mcp **237/237**。
- **CSS token**(#1008 教训):新增 `var(--wk-*)` 承前(#1043 已核 dmworkmcp token 均有定义,本轮 CSS 扩展沿用同族),build 通过无未解析。

---

### 🟡 非阻断
1. **多标签 URL 语义变化**:`?tag=a&tag=b` 从前端 union 改为后端 AND 交集 —— 旧书签结果集收窄(PR 已在 COMPREHENSION 标注,属正确语义)。
2. **依赖 sibling backend**:`/mcp_tags` + narrow-search + AND 语义需 octo-marketplace 配套分支部署,否则 popover 空态(优雅降级,已说明)。
3. Bot prompt 硬编码 `marketplace mcp create` CLI 命令(改名会漂移,集中单文件);backend 需确认 workspace 标签/MCP DELETE 有 admin 鉴权(前端非阻断)。

token `<mark>` React 转义无 XSS、SPACE_ID_RE 放宽仍 fail-closed(sanitize 中和)、tag 过滤无注入、卡片操作 mode-gated、owner 显式 provenance、标签校验 re-validate,构建与 128 测试通过、i18n 齐。**可合并。**

yujiawei (approved):
# Code Review — PR #1059 (octo-web)

Reviewed the full diff (25 files, +2601/−476) against the head SHA. This is a UI-alignment + tag-filter PR in `packages/dmworkmcp`, but it touches two security-relevant surfaces (the copyable bot-publish prompt and the Space ID handling), so I gave those an extra pass and did a line-level read of the data flow, plus a second adversarial pass focused on injection, XSS/ReDoS, request races, and cross-space state.

## 1. Scope / Spec

Matches the stated scope of the linked issue and the PR description — the seven listed chapters (McpCard/McpDetailModal alignment, PR #1026 tag/title overflow backport, tag chip filter, card owner actions, Space ID broadening, per-key token placeholder, polish) are all present. The removal of the 来源 (provenance) segmented filter is called out as intentional product feedback, and I see no functionality added beyond the described scope and no deviation from the documented approach. **Scope: OK.**

## 2. Security surfaces (verified sound)

- **`SPACE_ID_RE` + `sanitizeSpaceId` (`utils/mcpBotPublishPrompt.ts`)** — the broadened regex is fully anchored (`^(?:[0-9a-f]{32}|[0-9a-f]{8}-…-[0-9a-f]{12})$/i`) and both branches are pure hex, so no shell metacharacter can pass. Anything that fails the gate falls back to the `<space-id>` placeholder before it reaches the copyable `--space …` example. The unit tests exercise the injection payloads (`$(whoami)`, `` `whoami` ``, `; rm -rf /`, length-off-by-one) and both valid shapes. Fail-closed. ✅
- **Token placeholder highlight (`McpDetailModal.tsx` / `quickStartTemplates.ts`)** — rendering splits the string and emits React `<mark>` nodes (no `dangerouslySetInnerHTML`), so no XSS via user-supplied header/env keys. `TOKEN_PLACEHOLDER_RE` uses a linear `[^>]+` (no ReDoS), and a fresh `RegExp` is cloned per render so the global `lastIndex` never contaminates a subsequent call. ✅
- **Icon upload (`mcpService.ts`, pre-existing but re-read)** — the presigned PUT goes through an interceptor-free axios instance so the session token/`X-Space-Id` don't leak to the external storage origin, plus a scheme allowlist (`assertSafeUploadURL`). Still sound. ✅

## 3. Findings — all non-blocking (P2)

1. **Paste-batch tag input silently drops the tail after the first duplicate/limit hit** — `McpCreateModal.tsx`, `commitOne`/`commit`. `commitOne` returns `null` for a duplicate, and the batch loop treats any `null` as a hard `break`. Pasting `newA, existingTag, newB` commits `newA`, then aborts on `existingTag`, so `newB` is dropped with no feedback (the duplicate path also emits no Toast). Suggest returning the unchanged array for the duplicate case so the loop skips and continues.
2. **Space switch doesn't close open modals or cancel the in-flight edit fetch** — `McpMarketListPage.tsx`, `handleSpaceChanged_`. It resets tag/category/popover state but leaves `detailId` / `editingDetail` / `createVisible` / `deletingItem` untouched, and `handleEditFromCard`'s `await fetchMcpDetail(...)` has no version guard. A detail/edit modal opened in space A stays open (showing A's data) after switching to space B, and a save then issues a PATCH/DELETE for an A record while the axios interceptor stamps B's `X-Space-Id`. The loaded data was already owner-authorized so this isn't an escalation, and a mismatched-space write should fail closed at the backend — but the modals should be dismissed and the pending fetch invalidated on space change (mirror the reset you already do for the filters).
3. **Invalid interactive nesting for a11y** — `McpCard.tsx`. The card is `role="button"` `tabIndex={0}` and contains nested native `<button>` edit/delete controls. The `onKeyDown` early-return correctly prevents double activation, but nesting interactive elements inside a `role="button"` is invalid and some screen readers flatten/omit the inner actions. Consider making the card container non-`role=button` (e.g. an inner clickable region) so the owner actions are exposed cleanly.
4. **`Intl.DateTimeFormat(undefined, …)` uses the system locale, not the app locale** — `McpDetailModal.tsx`, `formatUpdatedTime` fallback. On a locale mismatch the absolute-date fallback renders in the browser's language rather than the selected UI language.
5. **Tag lists render without dedup** — `McpCard`/`McpDetailModal` map tags with `key={tag}`. Creation blocks duplicates going forward, but a legacy/edge record with a repeated tag would trigger a duplicate-key warning. A defensive de-dupe before render closes the gap.
6. **`apiBaseUrl` is interpolated into the bot prompt without a sanitize step** — `mcpBotPublishPrompt.ts`. The only current caller passes `resolveMcpAPIBaseURL(...)`, which returns `new URL(...).origin` (always `scheme://host[:port]`, no shell metacharacters), so this is safe today. Given the sibling `spaceId` is defensively sanitized, consider a matching guard on `apiBaseUrl` so a future direct caller can't inject into the copyable command.

## 4. Please verify manually (backend contract — outside this diff)

Flagging these because the PR is security-sensitive and the guarantees live in the paired backend, which I can't see from here:

- **Comma-bearing tags:** the frontend correctly percent-encodes and sends repeated params (`tag=v1.0%2Cbeta`), but after the backend URL-decodes, the value is `v1.0,beta`. If the backend's `tag` handling *also* comma-splits each individual repeated value (the contract is documented as "repeatable OR comma-separated"), a tag literal containing a comma will be re-split into two AND filters — exactly the failure the repeated-param design was meant to avoid. Please confirm the backend treats each repeated `tag=` occurrence as one literal and does not re-split it on commas.
- **Space-scoped writes:** please confirm the backend rejects (404/403, fails closed) a PATCH/DELETE that targets an MCP id from one space while carrying another space's `X-Space-Id` header (see finding #2).

## Verdict

No P0/P1 blocking issues. The security-relevant paths (space-id shell guard, token placeholder rendering, icon-upload credential isolation) are sound and well-tested. The findings above are quality/UX/a11y improvements and two backend behaviors worth a manual check. Approving; the maintainer's judgment on the P2s and the two verify-items is welcome as follow-ups.

PR #1056, opened 2026-07-24
refactor(dmworkbase): move channel webhook actions into bridge

katydids:
## Summary
- Move ChannelWebhook list, member, editor, and action handling into bridge hooks
- Keep UI components responsible for dialogs, confirms, toasts, and rendering
- Preserve successful mutation results while treating list refresh as best-effort
- Avoid duplicate member sync when the channel scope is unchanged

## Verification
- ./node_modules/.bin/vitest run packages/dmworkbase/src/Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx packages/dmworkbase/src/Service/__tests__/IncomingWebhook.test.ts packages/dmworkbase/src/Service/__tests__/IncomingWebhookService.test.ts packages/dmworkbase/src/bridge/channelWebhook

PR #1060, opened 2026-07-24
refactor(dmworkbase): move channel webhook actions into bridge

katydids:
## Summary
- Move ChannelWebhook list loading, member option loading, editor submit, and webhook actions into bridge hooks
- Keep UI components responsible for rendering, dialogs, confirmations, and toast feedback
- Keep IncomingWebhookService as the HTTP boundary for webhook API calls
- Preserve mutation success independently from best-effort list refresh failures
- Guard stale success and stale failure paths across create, update, regenerate, and delete flows
- Suppress stale one-time webhook secrets at UI consumers after scope changes
- Avoid duplicate member sync when the channel scope is unchanged
- Add bridge and component tests for stale handling, reload behavior, member sync stability, and one-time secret suppression

## Verification
- ./node_modules/.bin/vitest run packages/dmworkbase/src/Components/ChannelWebhook/__tests__/WebhookEditModal.test.tsx packages/dmworkbase/src/Components/ChannelWebhook/__tests__/ChannelWebhookPanel.test.tsx packages/dmworkbase/src/Service/__tests__/IncomingWebhook.test.ts packages/dmworkbase/src/Service/__tests__/IncomingWebhookService.test.ts packages/dmworkbase/src/bridge/channelWebhook

### DIFF
--- a/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
+++ b/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
@@ -123,7 +123,11 @@ export default function WebhookEditModal({
         );
         return;
       }
-      if (result.status === "busy" || result.status === "stale") {
+      if (
+        result.status === "busy" ||
+        result.status === "stale" ||
+        (result.status === "created" && result.stale)
+      ) {
         return;
       }
       if (result.status === "noop") {
--- a/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
+++ b/packages/dmworkbase/src/Components/ChannelWebhook/WebhookEditModal.tsx
@@ -1,116 +1,45 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { Channel, ChannelTypePerson, WKSDK } from "wukongimjssdk";
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import { Channel } from "wukongimjssdk";
 import { Switch, Toast } from "@douyinfe/semi-ui";
 import { IconAlertTriangle } from "@douyinfe/semi-icons";
 import WKModal from "../WKModal";
 import WKButton from "../WKButton";
 import AiBadge from "../AiBadge";
-import WKApp from "../../App";
 import { useI18n } from "../../i18n";
 import { extractErrorMsg } from "../../Service/APIClient";
-import { subscriberDisplayName } from "../../Utils/displayName";
 import {
-    getImChannelInfo,
-    getImChannelSubscribers,
-    syncImChannelSubscribers,
-} from "../../im-runtime/channelRuntime";
-import {
-    buildWebhookUpsertReq,
-    IncomingWebhook,
-    IncomingWebhookCreateResp,
-    IncomingWebhookService,
-    isFlagOn,
-    MENTION_UID_MAX_LENGTH,
-    MENTION_UIDS_MAX,
-    normalizeMentionUids,
-    validateMentionUids,
+  IncomingWebhook,
+  IncomingWebhookCreateResp,
+  MENTION_UID_MAX_LENGTH,
+  MENTION_UIDS_MAX,
 } from "../../Service/IncomingWebhook";
+import { useChannelWebhookEditor } from "../../bridge/channelWebhook/useChannelWebhookEditor";
+import { useChannelWebhookMembers } from "../../bridge/channelWebhook/useChannelWebhookMembers";
+import type { ChannelWebhookMemberOption } from "../../bridge/channelWebhook/types";
 import "./index.css";
 
 export interface WebhookEditModalProps {
-    channel: Channel;
-    /** 管理员才渲染头像输入（普通成员传 avatar 服务端直接 400） */
-    isManager: boolean;
-    /** 编辑模式传入现有项；新增模式不传 */
-    webhook?: IncomingWebhook;
-    /** 子区作用域：传入即把创建/更新打到该子区面（#451）；群面不传。channel 始终为父群。 */
-    threadShortId?: string;
-    onClose: () => void;
-    /** 保存成功回调；创建成功时携带含一次性 token/URL 的响应 */
-    onSaved: (created?: IncomingWebhookCreateResp) => void;
+  channel: Channel;
+  /** 管理员才渲染头像输入（普通成员传 avatar 服务端直接 400） */
+  isManager: boolean;
+  /** 编辑模式传入现有项；新增模式不传 */
+  webhook?: IncomingWebhook;
+  /** 子区作用域：传入即把创建/更新打到该子区面（#451）；群面不传。channel 始终为父群。 */
+  threadShortId?: string;
+  onClose: () => void;
+  /** 保存成功回调；创建成功时携带含一次性 token/URL 的响应 */
+  onSaved: (created?: IncomingWebhookCreateResp) => void;
 }
 
 // API 契约里的字段长度上限（OpenAPI schema 常量，非动态配额）
 const NAME_MAX_LENGTH = 64;
 const AVATAR_MAX_LENGTH = 255;
-
-/** mention_uids 选择器的候选项（群成员 / bot）。 */
-interface MentionMemberOption {
-    uid: string;
-    name: string;
-    isBot: boolean;
-}
-
-/** WKSDK Subscriber 运行时用到的子集（SDK 未导出精确字段）。 */
-type GroupSubscriber = {
-    uid?: string;
-    name?: string | null;
-    remark?: string | null;
-    status?: number;
-    orgData?: {
-        robot?: number;
-        real_name?: string | null;
-        realname_verified?: boolean | number | string | null;
-    } | null;
-};
-
-/**
- * 判定群成员是否为 AI/bot。双信号取并集，最大化命中：
- * - `subscriber.orgData.robot`：随订阅数据返回，但不一定被富化（可能缺）；
- * - 该 uid 的 Person `channelInfo.orgData.robot`：WKAvatar / 会话列表等通用判定，
- *   依赖 channelInfo 缓存是否已热。
- * 任一命中即视为 AI；都拿不到则按非 AI（best-effort，与全局头像/徽章判定一致）。
- */
-function isBotMember(uid: string, sub: GroupSubscriber): boolean {
-    if (isFlagOn(sub.orgData?.robot)) return true;
-    try {
-        const info = getImChannelInfo(
-            WKSDK.shared(),
-            new Channel(uid, ChannelTypePerson)
-        ) as { orgData?: { robot?: unknown } } | null | undefined;
-        return isFlagOn(info?.orgData?.robot);
-    } catch {
-        return false;
-    }
-}
-
-/**
- * 从群成员缓存读取 mention_uids 选择器候选：按 uid 去重，展示名走
- * subscriberDisplayName（实名 > 备注 > 昵称），AI 由 {@link isBotMember} 判定。
- * 不按 status 过滤、也不排除自己（自己同样可被 webhook 自动 @），成员资格由服务端
- * 兜底；同步读缓存，调用方在 syncSubscribes 完成后再读一次以刷新。
- */
-function readGroupMemberOptions(channel: Channel): MentionMemberOption[] {
-    let subs: GroupSubscriber[];
-    try {
-        subs = getImChannelSubscribers(WKSDK.shared(), channel) as GroupSubscriber[];
-    } catch {
-        return [];
-    }
-    const out: MentionMemberOption[] = [];
-    const seen = new Set<string>();
-    for (const sub of subs || []) {
-        const uid = sub?.uid;
-        if (!uid || seen.has(uid)) continue;
-        seen.add(uid);
-        out.push({
-            uid,
-            name: subscriberDisplayName(sub) || uid,
-            isBot: isBotMember(uid, sub),
-        });
-    }
-    return out;
-}
 
 /**
  * 新建 / 编辑 webhook 弹窗。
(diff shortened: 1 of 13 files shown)


Jerry-Xin (dismissed):
The refactor is relevant and architecturally appropriate, but stale action results can still leak feedback into a newly selected channel scope.

### 🔴 Blocking

- 🔴 Critical — Delete ignores the hook’s `stale` result. [index.tsx:158](/Users/mlamp/.openclaw/tmp/pr-review-__jzu7lm/repo/packages/dmworkbase/src/Components/ChannelWebhook/index.tsx:158) assigns `deleted = result.ok`, so `{ ok: true, stale: true }` still produces a success toast at line 168 after the user changes channels. Require `result.ok && !result.stale`, and add a component test analogous to the regenerate-secret test.

- 🔴 Critical — Test notifications are not scope-safe. [useChannelWebhookActions.ts:181](/Users/mlamp/.openclaw/tmp/pr-review-__jzu7lm/repo/packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:181) returns `true` even if the scope changed while the request was pending, causing [index.tsx:112](/Users/mlamp/.openclaw/tmp/pr-review-__jzu7lm/repo/packages/dmworkbase/src/Components/ChannelWebhook/index.tsx:112) to show “test sent” in the new channel. Rejections similarly surface an error for the old scope. Return an explicit stale result for both paths and suppress stale UI feedback.

### 💬 Non-blocking

- 🟡 Warning — Editor success still triggers a non-silent reload at [index.tsx:302](/Users/mlamp/.openclaw/tmp/pr-review-__jzu7lm/repo/packages/dmworkbase/src/Components/ChannelWebhook/index.tsx:302). A successful create/update followed by refresh failure therefore shows an error toast and error state, unlike the best-effort action reloads. Consider `load({ silentError: true })` and cover this path.

### ✅ Highlights

- HTTP calls remain correctly isolated behind `IncomingWebhookService`.
- Create, update, and regenerate include solid stale-response handling.
- Member synchronization avoids unnecessary reloads when only the `Channel` object identity changes.
- Focused tests could not be executed because `node_modules/.bin/vitest` is absent in the checkout.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1060 (octo-web)

Reviewed at head `0c26b37f`. Scope: refactor channel webhook actions/editor/members into bridge hooks with runtime DI; update consumer components and tests.

## Summary

This PR extracts channel webhook action, editor, and member-management logic from `ChannelWebhookPanel` and `WebhookEditModal` into four reusable bridge hooks (`useChannelWebhookActions`, `useChannelWebhookEditor`, `useChannelWebhookMembers`, and an updated `useChannelWebhookList`). Each hook exposes a runtime interface for dependency injection, enabling clean unit testing without real service dependencies. Stale-scope handling uses `scopeKey` + `requestSequence` tracking to prevent UI mutations after scope changes. The `getImChannelSubscribers(WKSDK.shared(), ...)` calls are replaced with `getCurrentImChannelSubscribers(...)` wrappers — verified functionally equivalent (both resolve to `WKSDK.shared().channelManager.getSubscribes(channel)`). Overall, a well-structured extraction that preserves all existing runtime contracts.

## Verification

- ✅ **Runtime equivalence** — `getCurrentImChannelSubscribers`/`getCurrentImChannelInfo`/`syncCurrentImChannelSubscribers` in `currentChannelRuntime.ts` are thin wrappers that delegate to `getImChannelSubscribers(WKSDK.shared(), ...)` etc. No behavioral change.
- ✅ **Stale scope handling** — Each action hook captures `scopeKey` at invocation time and checks `scopeKeyRef.current` before applying state mutations. Scope-change effects reset all in-flight state (`testingId`, `togglingId`, `coolingTestId`, `saving`). Tested in `useChannelWebhookActions.test.tsx` for both success and error paths.
- ✅ **Async lifecycle** — `mountedRef` + `requestSequenceRef` prevent state updates after unmount or scope change. Timer cleanup in `useChannelWebhookActions` via `clearTimeout` in unmount effect and scope-change effect.
- ✅ **Backward-compatible `load` signature** — `useChannelWebhookList.load` accepts both `boolean` (legacy) and `ChannelWebhookListReloadOptions` (new). Post-mutation refresh calls with `{ silentError: true }` correctly suppress error toasts.
- ✅ **Auth boundary unchanged** — `canManageIncomingWebhook` permission check still applied at the card level. `isManager` flows through identically. `buildWebhookUpsertReq` still enforces member-no-avatar constraint. Server remains the enforcement authority.
- ✅ **Error propagation preserved** — `handleToggle`, `handleRegenerate`, `handleDelete` all still throw from `onOk` to keep `wkConfirm` dialog open for retry. Stale-scope errors are swallowed (correct — scope already changed).

## Findings

No P0/P1 issues.

### P2 — Stale-scope create in editor silently drops result (`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookEditor.ts:167`)

When a webhook create succeeds server-side but the scope changes before the response arrives, `submit()` returns `{ ok: true, status: "created", created, stale: true }`. The consumer at `WebhookEditModal.tsx:128-131` silently returns — no `onSaved(created)` callback, no toast, modal stays open with form data. The server now has the webhook but the user gets no confirmation.

This requires the user to switch channel/thread while the create is in flight — an unlikely but reachable race. The parent panel's scope-change effect resets the editor's form state, so in practice the form clears immediately after. Not a correctness issue, but the user could briefly see stale form data before the effect fires.

### Nit — `toggleWebhook` closure captures stale `togglingId` (`packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:136`)

The guard `if (togglingId) return false` reads `togglingId` from the `useCallback` closure. React batches state updates, so the callback won't see the new value until the next render. This matches the original component-level behavior (same pattern pre-refactor) and the `await` ensures only one toggle is in flight at a time, but worth noting as a known React-hook limitation rather than a bug.

- **Runtime DI pattern**: Each hook accepts an optional `runtime` parameter, falls back to default via `runtimeOrDefault()`. Tests inject mock runtimes; production uses real `IncomingWebhookService`. Clean separation.
- **Cooldown timer in actions**: `testWebhook` cooldown uses `isCurrentScope(actionScopeKey)` guard in the timeout callback — won't clear `coolingTestId` if scope has changed. Scope-change effect also resets `coolingTestId` directly.
- **Editor `requestSequenceRef`**: Form reset effect depends on `[scopeKey, params.webhook]` — correctly resets all fields when the edit target or channel changes.
- **Member options sync**: `useChannelWebhookMembers` does initial sync read + async `syncSubscribers`, re-reads on completion. Same pattern as old `WebhookEditModal`. `buildChannelWebhookMemberOptionsForSelect` preserves self-fallback and extras-from-`mentionUids` logic.
- **`creatorNames` memo**: `useChannelWebhookList` still builds creator name map from `getCurrentImChannelSubscribers(channel)` — same data source, same fallback for self.
- **Test coverage**: New bridge hooks have dedicated test files covering happy path, scope-change stale handling, error propagation, and runtime DI. Consumer component tests updated to use the bridge hooks.

## Verdict: APPROVED

Well-executed extraction that preserves all runtime contracts, adds proper stale-scope handling, and enables clean unit testing through runtime DI. The P2 is an unlikely edge case that doesn't affect correctness in practice.





[Octo-Q] verdict: APPROVE — No P0/P1 issues. Well-structured refactoring with thorough stale-scope handling. One P2 edge case (stale-scope create drops result silently) is unlikely and non-blocking.

Jerry-Xin (changes requested):
The refactor is architecturally appropriate — moving list loading, member sync, editor submit, and webhook actions into bridge hooks while keeping components for rendering/dialogs/toasts, and keeping `IncomingWebhookService` as the HTTP boundary. The two previously-reported stale-scope one-time-secret leaks (create + regenerate) are correctly fixed here. However, the stale-scope convention was applied inconsistently: the delete and test action paths still surface UI feedback for a scope the user has already left.

### 🔴 Blocking

- 🔴 Delete ignores the hook's `stale` result. In `packages/dmworkbase/src/Components/ChannelWebhook/index.tsx` (delete `onOk`), `deleted = result.ok` — so a `{ ok: true, stale: true }` result (delete resolved after the user switched channel/thread) still fires the "deleted" success toast. This is the same stale-scope leak class already fixed for regenerate. Require `result.ok && !result.stale` before the success toast, and add a component test analogous to the regenerate-secret stale test.

- 🔴 Test notifications are not scope-safe. In `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts`, `testWebhook` returns `true` unconditionally after `runtime.test(...)` resolves, without checking `isCurrentScope(actionScopeKey)`. The consumer `handleTest` in `index.tsx` then shows a "test sent" success toast in the newly-selected channel. A rejection likewise surfaces a "test failed" error toast for the old scope. Return an explicit stale result (and suppress feedback) for both the resolve and reject paths, consistent with `toggleWebhook`/`regenerateWebhook`.

### 💬 Non-blocking

- 🟡 Editor success reload is not silent. The create/update `onSaved` handler in `index.tsx` calls `load()` (non-silent), so a successful mutation followed by a list-refresh failure shows an error toast/state — unlike the best-effort action reloads that use `reload({ silentError: true })`. Consider making the post-mutation reload best-effort too, and cover the path.
- 🟡 Stale conventions differ across hooks (`status:"stale"` in the editor, `stale?:boolean` on create/regenerate/delete post-success, `reason:"stale"` on error). Each consumer currently handles its own hook correctly, but a single shared shape would reduce the risk of a consumer forgetting the stale branch (as delete/test did).

### ✅ Highlights

- HTTP calls remain correctly isolated behind `IncomingWebhookService`; no new SSRF/injection surface introduced (webhook URLs are server-generated, not user input).
- Create, update, and regenerate have solid scope-drift (`requestSequence`/`scopeKey`) handling; the one-time secret is suppressed at the UI consumer after scope changes.
- Member synchronization avoids unnecessary reloads when only the `Channel` object identity changes (effect keyed on `scopeKey`).
- Good test coverage added for stale handling, reload behavior, and one-time-secret suppression.

lml2468 (dismissed):
## Review: #1060 — refactor(dmworkbase): channel webhook actions 移入 bridge

**意图**:#1056/#1033/#1020 家族的 webhook bridge 重构 —— list/member/editor/actions 移入 bridge hooks,UI 只管渲染/对话框/toast;跨 create/update/regenerate/delete **守卫 stale success + stale failure**,scope 变更后**在 UI 消费端压制 stale 的一次性 webhook 密钥**;成功 mutation 与 best-effort 刷新解耦。

锚定 head `0c26b37f` / merge-base `1673b812`(behind 0,`mergeable=MERGEABLE`,0 冲突)。13 文件 +2446/−842。

---

### 核心:stale-scope 一次性密钥泄漏(#1056 家族的两条 🔴)—— 修复且有专项测试 ✅

- **🔴 创建路径**:Modal(`WebhookEditModal.tsx:129`)对 `(result.status === "created" && result.stale)` early-return → `onSaved(result.created)`(:143)仅非-stale 触发;index.tsx:300 `setUrlResult(created)` 只收到非-stale created → **切 scope 后 A 频道创建密钥不弹进 B**。✅
- **🔴 重新生成**:`index.tsx:131` `if (result.ok && !result.stale) setUrlResult(result.response)` → stale 时不显示重新生成密钥。✅
- **toggle/test 不涉密钥**:`handleToggle` 仅切状态、`handleTest` 仅 `testSent` toast,无 URL/token 输出。✅
- **专项回归测试**:新 `ChannelWebhookPanel.test.tsx` —— `does not show regenerated secret URL when result is stale`:`stale:true` 时断言 `container.textContent` **不含 `secret-token`**;另有 editor 层 create/update stale success+failure 测试。**直接锁定密钥不泄漏。**✅

### 承前无回归 ✅
#1020 reload best-effort 解耦、密钥 `••••••••` 掩码、bridge/UI 分离、member-sync 去重(scope 不变不重拉);build ✅ 3.65s;webhook **7 files / 40 tests 全绿**(较 #1056 的 21 增加了 panel 级 stale-secret 抑制测试)。

---

### 🟡 非阻断(delete toast,承 #1056 仍在)
- **delete 的 stale toast 未收敛**:`deleteWebhook` stale-scope 成功返回 `{ok:true, stale:true}`,但 `handleDelete` 用 `deleted = result.ok`(忽略 stale)→ 切 scope 后仍弹「已删除」toast。**toast-only cosmetic**(删除在原频道正确执行、**无密钥泄漏**),非安全项;PR body 称「guard...delete flows」略过度(bridge 确返回 stale,但消费端 toast 未 gate)。建议 `deleted = result.ok && !result.stale` 对齐 regenerate。Jerry-Xin/我均标此项,不阻断。

两条真正的一次性密钥泄漏 🔴(create + regenerate)均已按 `result.stale` 拦截修复并有 panel 级「stale 不显密钥」专项测试,toggle/test 不涉密钥,#1020 解耦/掩码/分离无回归,构建与 40 测试通过。**可合并 —— APPROVE**(建议顺带 gate delete 的 stale toast)。

lml2468 (approved):
## Re-review: #1060 @ `461c8007` — channel webhook bridge(全 stale-scope 路径收口)

复核 delete + test 的 stale-feedback 路径(承前:secret-leak 🔴 已修,但 delete-toast + testWebhook 反馈仍会误归属新 scope)。锚定 head `461c8007` / merge-base `1673b812`(behind 0,`mergeable=MERGEABLE`)。真实增量(vs 我批准的 `0c26b37f`):4 文件 +141/−6(2 源 + 2 测试)。

**结论:全部 stale-scope 路径收口 + 专项测试,可合并 ✅**

### 承前 secret-leak 🔴(create/regenerate)—— 仍修复 ✅
create 走 `(status==="created" && stale)` early-return、regenerate `if (result.ok && !result.stale)`。专项测试 `does not show regenerated secret URL when result is stale` 断言 token 不进 DOM。

### 本轮补齐 —— delete + test 反馈误归属(我上轮 under-flag 的那两条)✅
- **delete**:`handleDelete` 改 `deleted = result.ok && !result.stale`(:159)→ 切 scope 后不弹「已删除」。测试 `does not show delete success toast when result is stale`。✅
- **test success**:`testWebhook` `return true` → **`return isCurrentScope(actionScopeKey)`** → 切 scope 后 `sent=false`,不弹「测试已发送」。✅
- **test failure**:catch `if (!isCurrentScope) return false`(吞 stale 错误,不弹误导 toast)。✅
- 测试 `does not show test success toast when result is stale` + `suppresses test success and failure after scope changes` + `suppresses toggle failures after scope changes`。✅

**至此三个 hook 的全部 stale 形态(create/update/regenerate/delete/test/toggle)在 consumer 端统一按 `isCurrentScope`/`result.stale` 拦截 —— 无跨频道密钥泄漏、无跨 scope 反馈误归属。**

### 无回归 ✅
#1020 reload best-effort 解耦、`••••••••` 掩码、bridge/UI 分离、member-sync 去重;build ✅ 3.49s;webhook **44 tests 全绿**(较上轮 40 增加 test/delete/toggle stale 专项)。

---

**结论**:上轮我把 stale-feedback 只当 delete-toast 🟡、且漏了 testWebhook 无 scope 守卫(Steve 更正的那条)—— 本轮作者把 delete + test(成功/失败)+ toggle 全部按 scope 门控,并补了逐条专项测试;secret-leak 🔴 无回归。所有 stale-scope 路径干净收口。**可合并 —— APPROVE。**

Jerry-Xin (approved):
Re-review at head 461c8007. Both stale-scope feedback leaks flagged in my prior review are now byte-verified fixed, and the earlier create/regenerate secret-leak guards did not regress. Approving.

## ✅ Prior blocking items — RESOLVED

- **Delete toast leaked to new scope** — `packages/dmworkbase/src/Components/ChannelWebhook/index.tsx:159` now computes `deleted = result.ok && !result.stale`, and the "deleted" toast at line 168 is gated on `deleted`. On a stale-scope success (`{ ok: true, stale: true }`) no toast fires in the newly selected channel. RESOLVED.
- **testWebhook feedback leaked to new scope** — `packages/dmworkbase/src/bridge/channelWebhook/useChannelWebhookActions.ts:186` changed the success path from `return true` to `return isCurrentScope(actionScopeKey)`, and added a `catch` that returns `false` in a stale scope instead of throwing. The consumer at `index.tsx:111` only shows "test sent" when the returned value is truthy, so post-switch feedback no longer leaks. This now matches the toggle guard pattern. RESOLVED.

## ✅ No regression on secret-leak guards

- Regenerate consumer still gates on `result.ok && !result.stale` before calling `setUrlResult` (`index.tsx:131-133`), so the one-time secret is not shown after a scope switch.
- The create path (`useChannelWebhookEditor.ts`) is byte-identical to the previously-reviewed head (blob unchanged) — a no-op in this delta — so its stale-gating cannot have regressed.

## 💬 Non-blocking

- Scope invalidation (`scopeKeyRef.current = scopeKey`) runs in a passive effect. There is a theoretical window between React commit and the passive effect where an in-flight request could settle against a not-yet-updated ref. This mechanism is pre-existing (unchanged by this PR) and the window is far shorter than any network round-trip, so it is not blocking here. If you want belt-and-suspenders correctness, invalidating the scope in a layout effect (or a render-time ref sync) would close it; a regression test that resolves a request before passive effects flush would document the guarantee.

## ✅ Highlights

- HTTP calls remain behind the service layer; mutation success is cleanly separated from best-effort refresh failure.
- Bridge and component tests cover the stale-result and reload scenarios; this delta adds targeted tests for the delete/test guards.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1060 (octo-web)

## Summary

This PR refactors the channel webhook management UI by extracting inline state and action logic from `ChannelWebhookPanel` and `WebhookEditModal` into four dedicated bridge hooks: `useChannelWebhookActions`, `useChannelWebhookEditor`, `useChannelWebhookMembers`, and an expanded `useChannelWebhookList`. The extraction is accompanied by staleness guards (`scopeKey` + `isCurrentScope` / `isCurrentRequest`) that prevent cross-scope feedback leakage — a meaningful improvement over the old code which unconditionally showed toasts and URL modals even when the user had already navigated to a different group/thread. Runtime interfaces with default implementations enable dependency injection for testing. Six new test files cover the hooks and the panel integration.

## Verification

Static analysis only at head `461c8007c0828f2043e1200c017a34dbd811edec`; build and tests not executed in this environment.

- ✅ **Behavioral parity** — Traced every action handler (toggle, test, regenerate, delete) and the edit-form submit flow from old inline code through the new hooks. API calls, error handling, cooldown logic, and `wkConfirm` re-throw semantics are preserved.
- ✅ **Staleness guards** — `useChannelWebhookActions` and `useChannelWebhookEditor` both track a `scopeKey` and guard state mutations behind `isCurrentScope` / `isCurrentRequest`. Timer cleanup on unmount and scope change verified in both hooks.
- ✅ **Subscriber API equivalence** — `getCurrentImChannelSubscribers(channel)` in `useChannelWebhookList` and `useChannelWebhookMembers` resolves to the same `WKSDK.shared()` instance as the old `getImChannelSubscribers(WKSDK.shared(), channel)`. No behavioral change.
- ✅ **Authorization consistency** — `isManager` gating for avatar field, `canManageIncomingWebhook` for action buttons, and server-side enforcement are all unchanged across the refactor.
- ✅ **Scope-change cleanup** — `index.tsx:89-92` clears `editTarget` and `urlResult` on `scopeKey` change; the hooks independently clear their in-flight IDs and timers. The old single-effect cleanup is correctly decomposed across the hooks.
- ✅ **Reload semantics** — `useChannelWebhookList.load` now accepts `{silentError: true}` for best-effort reloads after mutations (matching the old fire-and-forget `void load()` pattern), while direct user-triggered reloads still surface errors.

## Findings

No P0/P1 issues. Two P2 items below, both non-blocking.

### P2 — Silent feedback suppression on stale-scope mutations (`useChannelWebhookActions.ts:207`, `:240`)

When `regenerateWebhook` or `deleteWebhook` completes after the user has switched scope, the result is `{ok: true, stale: true}` and the caller suppresses all UI feedback — no URL modal for regenerate, no success toast for delete. The webhook was actually mutated on the server. In practice this path is nearly unreachable because `index.tsx:89-92` unmounts overlays on scope change and the confirmation dialogs hold focus. Consider adding a generic "Action completed in a different scope" toast so the user isn't left wondering if their click was lost.

### P2 — Toggle in-flight guard uses closure-captured state (`useChannelWebhookActions.ts:141`)

`toggleWebhook` reads `togglingId` from the `useCallback` closure (it's in the dependency array) to guard against concurrent toggles. Between two synchronous event dispatches in the same React batch, the stale `null` value fails to block the second call. This is inherited from the old inline code and not a regression, but a `useRef`-based lock would close the theoretical window.

## Blind-point Checklist (security_sensitive)

- **C1 — Dual-path parity**: N/A. Webhook CRUD operations are independent; no symmetric create↔delete or subscribe↔unsubscribe pairs in this diff.
- **C2 — Control-flow ordering / reuse**: Clear. The new hooks are consumed by a single call site each. No nested/re-entrant invocation patterns.
- **C3 — Authorization boundary**: Clear. `isManager` gating for avatar field unchanged. `canManageIncomingWebhook` for action buttons unchanged. Server validates all operations. No new endpoints exposed.
- **C4 — Authorization lifecycle**: N/A. No auth/jail changes.
- **C5 — Build/packaging**: N/A. No build config or packaging changes.
- **C6 — Governance/docs**: N/A. No policy or docs changes.

## Data Flow Tracing

- `items` → `IncomingWebhookService.list()` → `useChannelWebhookList` → `ChannelWebhookPanel` render: verified, with `requestSequenceRef` staleness guard.
- `testingId/togglingId/coolingTestId` → `useChannelWebhookActions` → `ChannelWebhookCard` props: verified, scope-aware reset.
- `memberOptions` → `getSubscribers()` / `syncSubscribers()` → `useChannelWebhookMembers` → `WebhookEditModal`: verified, with request sequence guard on sync callback.
- `submit()` → `buildWebhookUpsertReq()` → `runtime.create/update()` → result status → toast/onSaved: verified, stale path correctly suppresses secret display.
- `creatorNames` → `getCurrentImChannelSubscribers(channel)` → subscriber cache → `renderMeta`: verified, equivalent to old `getImChannelSubscribers(WKSDK.shared(), channel)`.

- Cooldown timer (`TEST_COOLDOWN_MS`) cleanup on unmount and scope change — correct in both the mounted-ref and scopeKey effects.
- `useChannelWebhookMembers` subscriber sync with `requestSequenceRef` staleness guard — prevents stale sync results from overwriting fresh data.
- `useChannelWebhookEditor` form state reset on `scopeKey` change (includes `webhook_id`) — correct for edit-to-create and webhook-to-webhook transitions.
- `buildChannelWebhookMemberOptionsForSelect` self-UID injection and `normalizeMentionUids` dedup for stale mention_uids — equivalent to old inline logic.
- `handleRegenerate` / `handleDelete` re-throw after toast — preserves `wkConfirm` retry-on-error behavior.
- Panel test mocks bridge hooks at the correct boundary; hook unit tests cover runtime injection.

## Verdict: APPROVED

Well-structured refactoring with genuine safety improvements (staleness guards, scope-aware feedback). No correctness, security, or behavioral regressions found. The two P2 items are refinements worth considering but not blockers.

[Octo-Q] verdict: APPROVED — no P0/P1 found; two P2 non-blocking items (stale-scope silent feedback, closure-based toggle guard).

PR #1075, opened 2026-07-24
feat(channel-setting): refresh information panels and member views

Candice00711:
## Summary

- align group and subzone information panels with the reviewed card-based layout and wider drawer while preserving existing entry points, permissions, and runtime actions
- move simple name, notice, remark, group-nickname, and subzone-name edits into reusable inline rows; keep group/subzone action semantics unchanged
- refine member grid and member-list presentation while preserving the existing Chinese/pinyin search, authoritative server search, and pagination behavior
- keep inline drafts intact when saves fail or external channel refreshes arrive during editing, and make the group-nickname clear action reliable on first edit

## Scope boundaries

- no changes to `bridge/channelSetting/*Actions`
- no changes to WKApp, WKSDK, datasource, Chat, Conversation, or Messages runtime paths
- no new subzone switches, hidden entries, or permission behavior
- no new pinyin-search capability; this PR only preserves and adapts the behavior already present at its merge base

## Validation

- focused `vitest`: 4 files, 34 tests passed
- inline-edit matrix: clear/save, successful save, `false` failure, rejected failure, external refresh while editing, and cancel-to-latest-value all passed
- `pnpm i18n:check`: passed
- `pnpm --dir apps/web build`: passed
- Prettier and `git diff --check`: passed
- full dmworkbase suite: 2,064/2,065 passed; one unrelated `ThreadPanel.archive` test timed out once under full-suite load and passed 16/16 when rerun in isolation
- Storybook static build: transformed 10,137 modules without component compilation errors, then exited with the existing low-level code 132 issue

## Manual acceptance

- group and subzone information cards, drawer width, member spacing, and member-list back navigation
- inline edit/save/cancel for group name, announcement, remark, group nickname, and subzone name
- failed saves keep the editor open and preserve the typed draft
- background channel refreshes do not overwrite a draft; Cancel shows the latest external value
- group nickname clear action works on first panel entry
- existing view-more and transfer-owner member search still supports Chinese and pinyin, including members outside a super-group first-page cache
- unchanged mute, pin, save, archive, leave, exit, and permission behavior

### DIFF
--- a/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
+++ b/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
@@ -210,8 +210,10 @@ export function ChannelSettingInlineEditRow({
             setSaving(true);
             try {
               const saved = await onSave(draft);
-              if (saved !== false) setCurrentValue(draft);
-              setEditing(false);
+              if (saved !== false) {
+                setCurrentValue(draft);
+                setEditing(false);
+              }
             } catch {
               // The bridge/container owns the user-facing error message.
             } finally {
--- a/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
+++ b/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
@@ -1,4 +1,5 @@
-import React from "react";
+import { Button, Input, TextArea } from "@douyinfe/semi-ui";
+import React, { useEffect, useState } from "react";
 
 import {
   ListItem,
@@ -9,6 +10,8 @@ import {
   ListItemSwitch,
   ListItemSwitchContext,
 } from "../../Components/ListItem";
+import { t } from "../../i18n";
+import "./index.css";
 
 export interface ChannelSettingInfoRowProps {
   title: string;
@@ -82,5 +85,121 @@ export function ChannelSettingActionRow({
       onClick={onClick}
       style={{}}
     />
+  );
+}
+
+export interface ChannelSettingInlineEditRowProps {
+  title: string;
+  value?: string;
+  displayValue?: React.ReactNode;
+  placeholder?: string;
+  maxCount?: number;
+  allowEmpty?: boolean;
+  multiline?: boolean;
+  onStartEdit?: () => boolean | void;
+  onSave: (value: string) => Promise<void | boolean>;
+}
+
+export function ChannelSettingInlineEditRow({
+  title,
+  value = "",
+  displayValue,
+  placeholder,
+  maxCount,
+  allowEmpty = false,
+  multiline = false,
+  onStartEdit,
+  onSave,
+}: ChannelSettingInlineEditRowProps) {
+  const [editing, setEditing] = useState(false);
+  const [currentValue, setCurrentValue] = useState(value);
+  const [draft, setDraft] = useState(value);
+  const [saving, setSaving] = useState(false);
+
+  useEffect(() => {
+    setCurrentValue(value);
+    setDraft(value);
+  }, [value]);
+
+  const exceeded = maxCount !== undefined && draft.length > maxCount;
+  const emptyInvalid = !allowEmpty && draft.trim().length === 0;
+  const unchanged = draft === currentValue;
+  const saveDisabled = saving || exceeded || emptyInvalid || unchanged;
+
+  const startEdit = () => {
+    if (onStartEdit?.() === false) return;
+    setDraft(currentValue);
+    setEditing(true);
+  };
+
+  if (!editing) {
+    return (
+      <ChannelSettingInfoRow
+        title={title}
+        value={currentValue === value ? displayValue ?? value : currentValue}
+        multiline={multiline}
+        onClick={startEdit}
+      />
+    );
+  }
+
+  const inputProps = {
+    value: draft,
+    placeholder,
+    disabled: saving,
+    showClear: true,
+    onChange: (next: string) => setDraft(next),
+    onClear: () => setDraft(""),
+  };
+
+  return (
+    <div className="wk-channelsetting-inline-edit">
+      <div className="wk-channelsetting-inline-edit-title">{title}</div>
+      {multiline ? (
+        <TextArea {...inputProps} autosize={{ minRows: 2, maxRows: 6 }} />
+      ) : (
+        <Input {...inputProps} />
+      )}
+      <div className="wk-channelsetting-inline-edit-footer">
+        {maxCount !== undefined ? (
+          <span
+            className={`wk-channelsetting-inline-edit-count${
+              exceeded ? " wk-channelsetting-inline-edit-count-error" : ""
+            }`}
+          >
+            {draft.length} / {maxCount}
+          </span>
+        ) : null}
+        <Button
+          theme="borderless"
+          disabled={saving}
+          onClick={() => {
+            setDraft(currentValue);
+            setEditing(false);
+          }}
+        >
+          {t("base.common.cancel")}
+        </Button>
+        <Button
+          theme="solid"
+          loading={saving}
+          disabled={saveDisabled}
+          onClick={async () => {
+            setSaving(true);
+            try {
+              const saved = await onSave(draft);
+              if (saved !== false) setCurrentValue(draft);
+              setEditing(false);
+            } catch {
+              // The bridge/container owns the user-facing error message.
+            } finally {
+              setSaving(false);
+            }
+          }}
+        >
+          {t("base.common.save")}
+        </Button>
+      </div>
+    </div>
   );
 }
--- a/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
+++ b/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
@@ -1,4 +1,5 @@
 import { Button, Input, TextArea } from "@douyinfe/semi-ui";
+import { IconClear } from "@douyinfe/semi-icons";
 import React, { useEffect, useState } from "react";
 
 import {
@@ -147,18 +148,39 @@ export function ChannelSettingInlineEditRow({
     value: draft,
     placeholder,
     disabled: saving,
-    showClear: true,
     onChange: (next: string) => setDraft(next),
-    onClear: () => setDraft(""),
   };
 
   return (
     <div className="wk-channelsetting-inline-edit">
       <div className="wk-channelsetting-inline-edit-title">{title}</div>
       {multiline ? (
-        <TextArea {...inputProps} autosize={{ minRows: 2, maxRows: 6 }} />
+        <TextArea
+          {...inputProps}
+          showClear
+          onClear={() => setDraft("")}
+          autosize={{ minRows: 2, maxRows: 6 }}
+        />
       ) : (
-        <Input {...inputProps} />
+        <Input
+          {...inputProps}
+          suffix={
+            draft && !saving ? (
+              <button
+                type="button"
+                className="wk-channelsetting-inline-edit-clear"
+                aria-label={`${title}-${t("base.common.clear")}`}
+                onMouseDown={(event) => {
+                  event.preventDefault();
+                  event.stopPropagation();
+                  setDraft("");
+                }}
+              >
+                <IconClear />
+              </button>
+            ) : null
+          }
+        />
       )}
       <div className="wk-channelsetting-inline-edit-footer">
         {maxCount !== undefined ? (
--- a/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
+++ b/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
@@ -1,6 +1,6 @@
 import { Button, Input, TextArea } from "@douyinfe/semi-ui";
 import { IconClear } from "@douyinfe/semi-icons";
-import React, { useEffect, useState } from "react";
+import React, { useEffect, useRef, useState } from "react";
 
 import {
   ListItem,
@@ -98,6 +98,7 @@ export interface ChannelSettingInlineEditRowProps {
   allowEmpty?: boolean;
   multiline?: boolean;
   onStartEdit?: () => boolean | void;
+  /** Resolve false or reject to keep the editor open with its current draft. */
   onSave: (value: string) => Promise<void | boolean>;
 }
 
@@ -116,10 +117,17 @@ export function ChannelSettingInlineEditRow({
   const [currentValue, setCurrentValue] = useState(value);
   const [draft, setDraft] = useState(value);
   const [saving, setSaving] = useState(false);
+  const editingRef = useRef(false);
+
+  useEffect(() => {
+    editingRef.current = editing;
+  }, [editing]);
 
   useEffect(() => {
     setCurrentValue(value);
-    setDraft(value);
+    if (!editingRef.current) {
+      setDraft(value);
+    }
   }, [value]);
 
   const exceeded = maxCount !== undefined && draft.length > maxCount;
--- a/packages/dmworkbase/src/App.css
+++ b/packages/dmworkbase/src/App.css
@@ -12,7 +12,7 @@
     /* --wk-height-chat-search: 65px; */
     --wk-height-chat-search: 64px;
     --wk-height-chat-conversation-header: 48px;  /* Figma: 48px */
-    --wk-wdith-chat-channelsetting: 380px;
+    --wk-wdith-chat-channelsetting: 28rem;
     --wk-width-chat-search-panel: 480px;
     --wk-width-thread-panel: 380px;
     --wk-layer-transition: 300ms cubic-bezier(0.33,1,0.68,1);
--- a/packages/dmworkbase/src/Components/ChannelSetting/index.css
+++ b/packages/dmworkbase/src/Components/ChannelSetting/index.css
@@ -83,7 +83,143 @@
     width: 100%;
     height: 100%;
     overflow: auto;
-    padding-bottom: 40px;
+    padding: var(--wk-sp-2) 0 var(--wk-sp-4);
+    background: var(--wk-bg-surface);
+}
+
+/* 仅作用于 ChannelSetting，避免影响其他 Sections/ListItem 调用方。 */
+.wk-channelsetting-content > .wk-sections {
+    display: flex;
+    flex-direction: column;
+    gap: var(--wk-sp-2);
+}
+
+.wk-channelsetting-content .wk-section {
+    margin: 0 var(--wk-sp-4);
+    overflow: hidden;
+    background: var(--wk-bg-base);
+    border: 1px solid var(--wk-border-subtle);
+    border-radius: var(--wk-r-sm);
+}
+
+.wk-channelsetting-content .wk-section-title {
+    padding: var(--wk-sp-3) var(--wk-sp-4) var(--wk-sp-2);
+    color: var(--wk-text-tertiary);
+    font-size: var(--wk-text-size-sm);
+}
+
+.wk-channelsetting-content .wk-section-subtitle {
+    padding: var(--wk-sp-2) var(--wk-sp-4) var(--wk-sp-3);
+    color: var(--wk-text-tertiary);
+    font-size: var(--wk-text-size-sm);
+}
+
+.wk-channelsetting-content .wk-section-row + .wk-section-row {
+    border-top: 1px solid var(--wk-border-subtle);
+}
+
+.wk-channelsetting-content .wk-list-item {
+    min-height: calc(var(--wk-sp-10) + var(--wk-sp-2));
+    padding: var(--wk-sp-2) var(--wk-sp-4);
+    background: transparent;
+}
+
+.wk-channelsetting-content .wk-list-item:hover {
+    background: var(--wk-bg-hover);
+}
+
+.wk-channelsetting-content .wk-list-item-static:hover {
+    background: transparent;
+}
+
+.wk-channelsetting-content .wk-list-item-title {
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-md);
+}
+
+.wk-channelsetting-content .wk-list-item-subtitle {
+    min-width: 0;
+    margin-left: var(--wk-sp-4);
+    color: var(--wk-text-secondary);
+    font-size: var(--wk-text-size-base);
+}
+
+.wk-channelsetting-content .wk-list-item-subtitle-muliteline {
+    margin-top: var(--wk-sp-2);
+    color: var(--wk-text-secondary);
+    font-size: var(--wk-text-size-base);
+}
+
+.wk-channelsetting-content .wk-channelsetting-qrcode-icon {
+    width: var(--wk-sp-4);
+    height: var(--wk-sp-4);
+    color: var(--wk-text-tertiary);
+}
+
+.wk-channelsetting-content .wk-section:has(.wk-subscribers) {
+    padding: var(--wk-sp-4) var(--wk-sp-2) var(--wk-sp-3);
+}
+
+.wk-channelsetting-content .wk-section:has(.wk-subscribers) .wk-section-row {
+    border: 0;
+}
+
+.wk-channelsetting-content .wk-subscribers-content {
+    grid-template-columns: repeat(5, minmax(0, 1fr));
+    column-gap: var(--wk-sp-2);
+    row-gap: var(--wk-sp-4);
+    padding-inline: var(--wk-sp-2);
+}
+
+.wk-channelsetting-content .wk-subscribers-item {
+    min-height: 0;
+    gap: var(--wk-sp-1-5);
+}
+
+.wk-channelsetting-content .wk-subscribers-item img,
+.wk-channelsetting-content .wk-subscribers-item-avatar-wrap img,
+.wk-channelsetting-content .wk-subscribers-item-avatar-wrap {
+    width: var(--wk-sp-12);
+    height: var(--wk-sp-12);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-name {
+    margin-top: 0;
+    color: var(--wk-text-secondary);
+    font-size: var(--wk-text-size-xs);
+    line-height: var(--wk-text-size-md);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-role-badge {
+    right: auto;
+    bottom: calc(0px - var(--wk-sp-1));
+    left: 50%;
+    padding: var(--wk-sp-0-5) var(--wk-sp-1);
+    color: var(--wk-text-inverse);
+    font-size: var(--wk-text-size-badge);
+    border-radius: var(--wk-r-xs);
+    box-shadow: none;
+    transform: translateX(-50%);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-role-badge-owner {
+    background: var(--wk-color-warning);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-role-badge-manager {
+    background: var(--wk-text-primary);
+}
+
+.wk-channelsetting-content .wk-subscribers-more {
+    margin: var(--wk-sp-4) var(--wk-sp-2) 0;
+    padding: var(--wk-sp-2);
+    border-radius: var(--wk-r-sm);
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-sm);
+}
+
+.wk-channelsetting-content .wk-subscribers-more:hover {
+    background: var(--wk-bg-hover);
 }
 
 .wk-channelsetting-channel-info {
@@ -198,8 +334,12 @@
 
 .wk-route-content .wk-channelsetting-input-edit {
     box-sizing: border-box;
-    width: calc(100% - (2 * var(--wk-sp-2)));
-    margin: 0 var(--wk-sp-2);
+    width: calc(100% - (2 * var(--wk-sp-4)));
+    margin: var(--wk-sp-2) var(--wk-sp-4) 0;
+    overflow: hidden;
+    background: var(--wk-bg-base);
+    border: 1px solid var(--wk-border-subtle);
+    border-radius: var(--wk-r-sm);
 }
 
 .wk-channelmanage {
--- a/packages/dmworkbase/src/Components/Subscribers/index.tsx
+++ b/packages/dmworkbase/src/Components/Subscribers/index.tsx
@@ -54,12 +54,12 @@ export class Subscribers extends Component<SubscribersProps> {
         <div className="wk-subscribers-item-avatar-wrap">
           <img src={WKApp.shared.avatarUser(subscriber.uid)} alt=""></img>
           {subscriber.role === GroupRole.owner && (
-            <span className="wk-subscribers-item-role-badge">
+            <span className="wk-subscribers-item-role-badge wk-subscribers-item-role-badge-owner">
               {this.context.t("base.subscribers.role.owner")}
             </span>
           )}
           {subscriber.role === GroupRole.manager && (
-            <span className="wk-subscribers-item-role-badge">
+            <span className="wk-subscribers-item-role-badge wk-subscribers-item-role-badge-manager">
               {this.context.t("base.subscribers.role.manager")}
             </span>
           )}
@@ -157,12 +157,17 @@ export class Subscribers extends Component<SubscribersProps> {
                           )}
                         />,
                         new RouteContextConfig({
-                          title: this.context.t("base.subscribers.memberList"),
+                          title: this.context.t(
+                            "base.subscribers.memberListWithCount",
+                            { values: { count: vm.memberCount() } }
+                          ),
                         })
                       );
                     }}
                   >
-                    {this.context.t("base.subscribers.viewMore")}
+                    {this.context.t("base.subscribers.viewAll", {
+                      values: { count: vm.memberCount() },
+                    })}
                   </div>
                 ) : undefined}
               </div>
--- a/packages/dmworkbase/src/Components/Subscribers/list.css
+++ b/packages/dmworkbase/src/Components/Subscribers/list.css
@@ -4,19 +4,90 @@
     background-color: var(--wk-bg-surface);
     width: 100%;
     height: 100%;
-    overflow: scroll;
+    overflow-y: auto;
+    overflow-x: hidden;
+}
+
+/*
+ * 成员列表返回时保留 animationend，让 WKViewQueue 继续按原逻辑清理栈顶，
+ * 但取消可见位移，避免标题已切回父页而内容仍横向退场造成闪动。
+ * 不改变 RoutePage/WKViewQueue 的入栈、退栈与数据刷新逻辑。
+ */
+.wk-viewqueue-view-out:has(.wk-subscrierlist) {
+    background: var(--wk-bg-surface);
+    animation-name: wk-subscriber-list-exit;
+    animation-duration: var(--wk-dur-fast);
+    animation-timing-function: var(--wk-ease);
+}
+
+@keyframes wk-subscriber-list-exit {
+    from,
+    to {
+        transform: translate3d(0, 0, 0);
+    }
+}
+
+.wk-subscrierlist .wk-indextable-search-box {
+    height: auto;
+    margin: var(--wk-sp-2) var(--wk-sp-5);
+    padding: var(--wk-sp-1-5) var(--wk-sp-2);
+    align-items: center;
+    background: var(--wk-bg-base);
+    border: 1px solid var(--wk-border-subtle);
+    border-radius: var(--wk-r-sm);
+}
+
+.wk-subscrierlist .wk-indextable-search-icon {
+    margin-left: 0;
+    color: var(--wk-text-tertiary);
+}
+
+.wk-subscrierlist-search-icon {
+    width: var(--wk-sp-4);
+    height: var(--wk-sp-4);
+}
+
+.wk-subscrierlist .wk-indextable-search-input {
+    flex: 1;
+    max-width: none;
+    margin-left: var(--wk-sp-2);
+}
+
+.wk-subscrierlist .wk-indextable-search-input input {
+    width: 100%;
+    height: auto;
+    padding: 0;
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-md);
+    line-height: var(--wk-sp-5);
+}
+
+.wk-subscrierlist .wk-indextable-search-input input::placeholder {
+    color: var(--wk-text-tertiary);
 }
 
 .wk-subscrierlist-list-item {
     display: flex;
     align-items: center;
-    padding: 1rem;
+    gap: var(--wk-sp-3);
+    min-height: var(--wk-sp-12);
+    padding: var(--wk-sp-2) var(--wk-sp-5);
     cursor: pointer;
+    transition: background var(--wk-dur-fast) var(--wk-ease);
+}
+
+.wk-subscrierlist-list-item:hover {
+    background: var(--wk-bg-hover);
 }
 
 .wk-subscrierlist-item-avatar {
     position: relative;
     flex-shrink: 0;
+}
+
+.wk-subscrierlist-item-avatar .wk-avatar {
+    width: var(--wk-sp-8);
+    height: var(--wk-sp-8);
 }
 
 .wk-subscrierlist-item-online-badge {
@@ -26,13 +97,17 @@
 }
 
 .wk-subscrierlist-item-name {
-    margin-left: 1rem;
+    min-width: 0;
+    margin-left: 0;
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
-    max-width: 10rem;
+    max-width: 16rem;
     display: flex;
     align-items: center;
+    gap: var(--wk-sp-1);
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-md);
 }
 
 .wk-subscrierlist-item-source {
@@ -57,13 +132,36 @@
 
 .wk-subscrierlist-item-content {
     display: flex;
+    min-width: 0;
     width: 100%;
-    justify-content: space-between;
+    align-items: center;
+    justify-content: flex-start;
+    gap: var(--wk-sp-1-5);
 }
 
 .wk-subscrierlist-item-desc {
-    font-size: 14px;
+    display: none;
+    flex-shrink: 0;
+    padding: var(--wk-sp-0-5) var(--wk-sp-1);
+    font-size: var(--wk-text-size-tiny);
+    font-weight: 600;
+    line-height: var(--wk-text-size-base);
+    border-radius: var(--wk-r-xs);
+}
+
+.wk-subscrierlist-item-desc-owner,
+.wk-subscrierlist-item-desc-manager {
+    display: inline-flex;
+}
+
+.wk-subscrierlist-item-desc-owner {
+    color: var(--wk-color-warning);
+    background: var(--wk-color-warning-bg);
+}
+
+.wk-subscrierlist-item-desc-manager {
     color: var(--wk-text-secondary);
+    background: var(--wk-bg-elevated);
 }
 
 .wk-subscrierlist-item-action {
--- a/packages/dmworkbase/src/Components/Subscribers/list.tsx
+++ b/packages/dmworkbase/src/Components/Subscribers/list.tsx
@@ -298,9 +298,7 @@ export class SubscriberList extends Component<
               >
                 <div className="wk-indextable-search-box">
                   <div className="wk-indextable-search-icon">
-                    <IconSearchStroked
-                      style={{ color: "#bbbfc4", fontSize: "20px" }}
-                    />
+                    <IconSearchStroked className="wk-subscrierlist-search-icon" />
                   </div>
                   <div className="wk-indextable-search-input">
                     <input
@@ -312,7 +310,6 @@ export class SubscriberList extends Component<
                       )}
                       ref={(rf) => {}}
                       type="text"
-                      style={{ fontSize: "17px" }}
                     />
                   </div>
                 </div>
@@ -393,7 +390,15 @@ export class SubscriberList extends Component<
                               </Tag>
                             )}
                           </div>
-                          <div className="wk-subscrierlist-item-desc">
+                          <div
+                            className={`wk-subscrierlist-item-desc${
+                              item.role === GroupRole.owner
+                                ? " wk-subscrierlist-item-desc-owner"
+                                : item.role === GroupRole.manager
+                                ? " wk-subscrierlist-item-desc-manager"
+                                : ""
+                            }`}
+                          >
                             {this.getRoleName(item)}
                           </div>
                         </div>
--- a/packages/dmworkbase/src/Components/Subscribers/vm.ts
+++ b/packages/dmworkbase/src/Components/Subscribers/vm.ts
@@ -71,4 +71,8 @@ export class SubscribersVM extends ProviderListener {
         let showMemberNum = this.shouldShowMemberNum()
         return this.subscribers.length>showMemberNum
     }
+
+    memberCount() {
+        return this.routeData.channelInfo?.orgData?.member_count || this.subscribers.length
+    }
 }
--- a/packages/dmworkbase/src/features/channelSetting/__tests__/channelSettingSections.test.ts
+++ b/packages/dmworkbase/src/features/channelSetting/__tests__/channelSettingSections.test.ts
@@ -5,6 +5,7 @@ import { ChannelTypeCommunityTopic } from "../../../Service/Const";
 import { ThreadStatus } from "../../../Service/Thread";
 import { GroupStatusDisband } from "../../../Utils/groupDisband";
 import { updateChannelSettingMyGroupNickname } from "../../../bridge/channelSetting/channelSettingActions";
+import { t } from "../../../i18n";
 import { buildChannelGroupInfoSection } from "../channelSettingGroupInfoSection";
 import {
   buildChannelDangerSection,
@@ -16,8 +17,27 @@ import {
   buildThreadActionsSection,
   buildThreadInfoSection,
   buildThreadMdSection,
+  buildThreadOverviewSection,
   buildThreadWebhookSection,
 } from "../channelSettingThreadSections";
+
+vi.mock("@douyinfe/semi-ui", () => ({
+  Button: vi.fn(),
+  Input: vi.fn(),
+  Switch: vi.fn(),
+  Tag: vi.fn(),
+  TextArea: vi.fn(),
+  Toast: {
+    error: vi.fn(),
+    warning: vi.fn(),
+  },
+}));
+
+vi.mock("@douyinfe/semi-icons", () => ({
+  IconAlertTriangle: vi.fn(),
+  IconLink: vi.fn(),
+  IconPlus: vi.fn(),
+}));
 
 vi.mock("../../../App", () => ({
   default: {
@@ -178,9 +198,7 @@ describe("channel setting section builders", () => {
     const inputEditPush = vi.fn();
     const section = buildMyGroupNicknameSection(context, inputEditPush);
 
-    section?.rows?.[0].properties.onClick();
-    const onFinish = inputEditPush.mock.calls[0][2];
-    await onFinish("Alice Updated");
+    await section?.rows?.[0].properties.onSave("Alice Updated");
 
     expect(updateChannelSettingMyGroupNickname).toHaveBeenCalledWith({
       channel: context.routeData().channel,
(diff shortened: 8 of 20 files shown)


mochashanyao (changes requested):
[Octo-Q · automated review]

---

[Octo-Q] verdict: REQUEST_CHANGES — P1 blocker: inline editor closes on `return false` from `onSave`, losing the user's draft on save failure (group name, group notice, thread name). One-line fix: only call `setEditing(false)` when `saved !== false`.

Jerry-Xin (changes requested):
Channel-setting refresh (info panels + inline edit + member-list search styling): authz/XSS/dark-mode all clean; one blocking data-loss bug in the new inline editor's save path.

## 🔴 Blocking

**Draft is discarded when a save fails (`ChannelSettingInlineEditRow`).**
In `packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx` the save button handler calls `setEditing(false)` unconditionally:

```
const saved = await onSave(draft);
if (saved !== false) setCurrentValue(draft);
setEditing(false);            // runs even when saved === false
```

Several `onSave` handlers deliberately resolve to `false` on API failure (they toast the error and `return false`): group name and group notice in `channelSettingGroupProfileRows.tsx` (lines ~62/73/132/143), the remark row in `channelSettingGroupManagementRows.tsx`, and the thread name in `channelSettingThreadInfoSection.tsx`. On failure the code correctly skips `setCurrentValue`, but the editor still closes, so the user's typed text is lost and they must retype after seeing the error toast. This is a regression versus the previous push-based editor.

Fix: keep the editor open on failure, e.g.
```
const saved = await onSave(draft);
if (saved !== false) {
  setCurrentValue(draft);
  setEditing(false);
}
```
Please also add a failure-path unit test (current `ChannelSettingInlineEditRow.test.tsx` only covers the clear + successful-save path).

## 💬 Non-blocking

- `ChannelSettingInlineEditRow` calls `setSaving`/`setCurrentValue` after an awaited async `onSave` with no unmount guard. Not observed to fire here, but if a parent can unmount the row mid-save it would produce a state-after-unmount warning; a mounted ref or ignoring the result post-unmount would harden it.
- The `useEffect(..., [value])` resets both `currentValue` and `draft` from the incoming prop; if a parent refresh lands while the user is actively editing, the in-progress draft would be overwritten. Edge case only, but worth confirming refresh cannot race an open editor.

## ✅ Highlights

- Member-list search is a client-side filter over already-loaded subscribers (`localSearch` / `vm.search`), not a new backend query — no risk of dumping members outside the user's scope; the debounced search plumbing is pre-existing and untouched.
- Permission guards are preserved, not weakened: the manager/creator gate for group name/notice moved into `onStartEdit` (returns `false` to block non-managers), and the thread name keeps its `threadInfo` + `canEdit` checks. Group-management admin/dissolve rows are unchanged.
- No `dangerouslySetInnerHTML`; all user-controlled fields (name, notice, remark, member names) render as escaped React text nodes.
- Dark-mode safe: every `--wk-*` token used in the changed CSS is defined, and no hardcoded color literals were introduced.
- New i18n keys (`common.clear`, `subscribers.viewAll`, `subscribers.memberListWithCount`) are wired up in both en-US and zh-CN with no orphans; role-badge and empty-nickname behaviors are covered by tests. CI is green.

yujiawei (changes requested):
# Code Review — PR #1075 (octo-web)

**Head SHA:** `8043f7b8ba51a47016ccbd47a7b0c1db1dd53dfc` · **Base:** `main` (merge-base `e28c75b0`)

This is a self-contained frontend refactor of the channel-setting information panels: it introduces a reusable `ChannelSettingInlineEditRow`, migrates group name / notice / remark / group-nickname / thread-name edits from a pushed sub-page (`inputEditPush`) into inline rows, merges the three thread sub-sections into one `thread.overview` card, and refreshes member grid/list presentation. Verified: no `bridge/channelSetting/*Actions`, `WKApp`, datasource, or Chat/Conversation/Messages runtime changes — scope boundaries in the description hold. Changed files typecheck clean (remaining `tsc` errors are all pre-existing environmental noise outside this diff), and the 4 changed test files pass (11 tests locally / 25 per the PR).

**Spec: ✅ (with one description nit)**

- **Missing:** none. The stated scope — inline card layout, wider drawer, member presentation, reliable first-edit nickname clear, empty-nickname regression test — is implemented.
- **Out-of-scope / over-build:** none. No bridge/runtime/permission surfaces were touched; the thread-section merge preserves each existing builder and its permission/click logic (`buildThreadOverviewSection` flat-maps `buildThreadInfoSection` + `buildThreadMdSection` + `buildThreadWebhookSection`), and the comment explicitly forbids re-adding unregistered entries.
- **Deviation:** none functional. *Nit:* the PR summary says it "adds Chinese/full-pinyin matching for view-more and ownership-transfer search," but that matching already exists at merge-base (`channelSettingMemberSearch.ts` `getPinyin`/`toSimplized`) and is **not** part of this diff. The description overclaims; the feature itself is fine.

## 2. Code quality

**Quality: Changes-Requested** — two draft-loss defects in the shared inline-edit component.

### P1 — Save failure discards the user's draft and closes the editor
`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:209-219`

```tsx
const saved = await onSave(draft);
if (saved !== false) setCurrentValue(draft);
setEditing(false);          // <-- runs even when saved === false
```

`setCurrentValue` is correctly guarded by `saved !== false`, but `setEditing(false)` is not. Three consumers signal save failure by resolving `false` (not throwing):
- `channelSettingGroupProfileRows.tsx:71-74` (group name) and `:141-143` (group notice) — `.catch(...) { Toast.error(...); return false; }`
- `channelSettingThreadInfoSection.tsx:79-81` (thread name) — `catch { Toast.error(...); return false; }`

On a failed save these paths still exit edit mode; because `currentValue` was not updated, the row reverts to the old value and the user's typed draft is gone — they must retype from scratch. Only the `remark` path (`channelSettingGroupManagementRows.tsx:225`, which `throw`s) is handled correctly, because the throw hits the empty `catch` and skips `setEditing(false)`. **Fix:** move `setEditing(false)` inside the `if (saved !== false)` block (and keep the `catch` path leaving `editing` untouched).

### P1 — Background channel-info update overwrites an in-progress draft
`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:120-123`

```tsx
useEffect(() => {
  setCurrentValue(value);
  setDraft(value);          // <-- overwrites draft even while editing
}, [value]);
```

`setDraft(value)` runs unconditionally whenever the `value` prop changes. The channel-setting VM registers a live `channelInfoListener` (`Components/ChannelSetting/vm.ts:324-331`) that calls `reloadChannelInfo()` → `notifyListener()` on any remote update to this channel, which re-renders the rows with a fresh `value`. If a remote edit (or the user's own optimistic refresh) lands while someone is mid-typing, their draft is silently replaced. **Fix:** only sync `draft` from `value` when not editing, e.g. `if (!editing) setDraft(value);` (keep `setCurrentValue(value)` unconditional).

### P2 — Inconsistent failure contract across consumers
`onSave` implementers mix two failure conventions: `return false` (group name/notice, thread name) vs. `throw` (remark). The component handles both, but the split contract is brittle and is exactly what let the P1 above slip in. Recommend standardizing on one convention (returning `false` reads cleanest given the `void | boolean` return type) and documenting it on `ChannelSettingInlineEditRowProps.onSave`.

### Non-blocking notes
- After a successful inline save of the group name, the row shows the plain saved string until the next `value`-driven refresh, because the external-group `displayValue` tag only renders while `currentValue === value` (`index.tsx:140`). Minor transient UX; self-heals on refresh.
- `.wk-list-item-subtitle-muliteline` (ChannelSetting/index.css) reuses the repo's existing (misspelled) `muliteline` class from `ListItem` — this is intentional and correct, not a typo to "fix" in isolation.

## 3. Verdict

**REQUEST_CHANGES** — Spec ✅, but Quality Changes-Requested on two P1 draft-loss defects in the shared inline-edit component (save-failure exit, and background-refresh overwrite). Both are small, localized fixes in `ChannelSettingRows/index.tsx`. The existing regression test covers the empty-nickname clear path but not save-failure or concurrent-refresh; adding a case for each would lock in the fix.

packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
```diff
@@ -84,3 +88,142 @@ export function ChannelSettingActionRow({
...
+  allowEmpty = false,
+  multiline = false,
+  onStartEdit,
+  onSave,
+}: ChannelSettingInlineEditRowProps) {
+  const [editing, setEditing] = useState(false);
+  const [currentValue, setCurrentValue] = useState(value);
+  const [draft, setDraft] = useState(value);
+  const [saving, setSaving] = useState(false);
+
```
lml2468, line 125: 🔴 正在编辑时远端 value 变化会覆盖用户草稿:此 effect 在 `value` 变化时**无条件** `setDraft(value)`,没有 `editing` 守卫。本 PR 的核心就是「刷新信息面板」——父组件在用户编辑中重渲染并传入更新后的 `value`(群名/公告远端更新或面板刷新)时,effect 触发 → 用户正在输入的 `draft` 被冲掉、输入丢失,与刚修好的『保存失败保留草稿』是同一类数据丢失。修法:`useEffect(() => { setCurrentValue(value); if (!editing) setDraft(value); }, [value, editing]);` 并补一条『编辑中 value 变化不覆盖 draft』回归测试。

lml2468 (changes requested):
## Review: #1075 @ `9e45524d` — feat(channel-setting): refresh information panels and member search

复核当前 live head `9e45524d`(触发消息的 `8043f7b8` 已过期,期间追了 `fix: preserve drafts after failed saves`)。20 文件 +782/−176。`review:CHANGES_REQUESTED`。

**结论:需修改 — REQUEST_CHANGES** ❌(一条草稿丢失 🔴 已修,另一条同类 🔴 仍在)

---

### ✅ 已修:保存失败丢草稿(peers 上轮 🔴)
`ChannelSettingInlineEditRow` 保存按钮现 `const saved = await onSave(draft); if (saved !== false) { setCurrentValue(draft); setEditing(false); }`(index.tsx:212-215)—— 保存失败(返回 false)保留编辑态与草稿。已有测试「keeps the draft open when saving fails」。

### 🔴 阻塞:编辑中远端 value 变化覆盖用户草稿(yujiawei,独立确认成立)
`index.tsx:119-122`:
```js
useEffect(() => { setCurrentValue(value); setDraft(value); }, [value]);
```
`value` 变化时**无条件** `setDraft(value)`,**无 `editing` 守卫**。本 PR 的核心正是「刷新信息面板」—— 父组件在用户编辑中重渲染并传入更新后的 `value`(群名/公告远端更新、或面板刷新周期)时,effect 触发 → 用户正在输入的 `draft` 被冲掉、输入丢失。与刚修好的「保存失败保留草稿」是**同一类数据丢失**,只是触发源换成远端更新;PR 自带的面板刷新使该触发更可能发生。
- **修法**:`useEffect(() => { setCurrentValue(value); if (!editing) setDraft(value); }, [value, editing]);` + 补「编辑中 value 变化不覆盖 draft」回归测试(现有 2 个测试未覆盖此路径)。

### 其余逐项核过(通过)
- **成员搜索授权安全**:走 `localSearch(keyword) => Subscriber[]` + `vm.search` —— 对**已加载成员的客户端过滤**,非新接口、无空/通配 dump、不越权(与 Steve 结论一致)。
- **权限门保留**、**无 XSS**(字段走 React 文本转义)、`--wk-*` token 均定义暗色安全、新增 3 个 i18n key 中英齐全无孤儿(Steve 逐条核过,我 spot-check 一致)。

**构建 / 测试 / i18n(实跑 @ `9e45524d`)**:`pnpm --filter @octo/web build` ✅ 3.63s;`pnpm i18n:check` ✅;`ChannelSettingInlineEditRow` 2 tests 绿。`mergeable=true`、无冲突。

---

### 🟡 非阻断
- 内联编辑器 `onSave` await 后 setState 缺卸载守卫,卸载后 setState 可能告警(Steve/Jerry-Xin);
- PR 描述称新增拼音匹配,实际未改动该功能,建议订正描述(yujiawei);
- `pr-title-lint` 需过标题格式。

---

保存失败丢草稿 🔴 已修 👍;但「编辑中远端 value 变化覆盖草稿」这条同类数据丢失(无 `editing` 守卫的 value-sync effect)仍在,且本 PR 的面板刷新使其更易触发。**需修改 —— REQUEST_CHANGES**(修法:value-sync 加 `!editing` 守卫 + 补回归测试)。

cnwenf (changes requested):
# Code Review — PR #1075 (Mininglamp-OSS/octo-web)

## Summary

This PR migrates the channel/thread settings panels to a card-based layout in a widened drawer, replaces the push-route `inputEditPush` modal flow with a reusable inline `ChannelSettingInlineEditRow` component, merges the thread info/md/webhook sections into a single `thread.overview` card, and adds Chinese/full-pinyin member search with a server-count label. All three independent reviews agree the refactor is structurally sound, the headline fixes (cleared-nickname `?? ""` fallback, 5-row overview composition) are correct and mutation-effective, security is clean, and permission gating is preserved. The consensus concerns are the new inline-edit component's failure-signal contract — a nickname save that surfaces no error on failure, and `onSave` returning `false` closing the editor while discarding the draft — plus thin, CI-invisible test coverage of the component's boundary branches. Two reviewers concluded COMMENT and one NEEDS_CHANGES; per the consolidation rule, any NEEDS_CHANGES drives the final verdict.

## Findings

### P1 — Nickname onSave silent failure (channelSettingNicknameSection.tsx:39) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P1, glm5.2=P2, qwen3.8=P2)

The my-group-nickname `onSave` awaits `updateChannelSettingMyGroupNickname` with no try/catch and no Toast, so a rejection propagates to `ChannelSettingInlineEditRow`'s empty `catch` (`index.tsx:215`) which swallows it; the comment at `:216` claiming "the bridge/container owns the user-facing error message" is false for this caller — no bridge shows a toast. The three sibling `onSave` handlers in the same diff all surface failure (`channelSettingThreadInfoSection.tsx:80`, `channelSettingGroupProfileRows.tsx:72`, `channelSettingGroupManagementRows.tsx:224`), so on a failed nickname save the user sees the Save button stop loading and nothing else — editing stays open with the unsaved draft and zero feedback. Confirmed by mutation: HEAD reproduces `Toast.error` uncalled + editing open; adding `try/catch + Toast.error + return false` flips the assertion. Fix direction: wrap the nickname `onSave` in try/catch with `Toast.error(error?.msg)` and `return false`, mirroring the thread-info/groupProfile siblings.

### P1 — CI build job runs no `pnpm test` for dmworkbase (.github/workflows/ci.yml:76-104) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8] (severity split: qwen3.7=P1, glm5.2=P2, qwen3.8=P3)

The required CI build job (`.github/workflows/ci.yml:76-104`, unchanged by this PR) runs install → i18n:check → `pnpm --filter @octo/docs exec vitest run` → build → lint, but never runs vitest for `@octo/base`/dmworkbase; no PR-triggered workflow invokes `pnpm test`/`--filter @octo/base`/`packages/dmworkbase`. So this newly-added test file and the modified `channelSettingSections.test.ts` are not executed by any required gate — a regression in save/clear/gate logic would merge green. Confirmed by contract test_kind: a structured YAML parser over all 23 workflows enumerated the build-job step list and found the only vitest invocation scoped to `@octo/docs`. Fix direction: add a `pnpm --filter @octo/base exec vitest run` step (or `pnpm test`) to the required CI build job. (All three reviewers note this is a pre-existing infrastructure gap, not introduced by this PR.)

### P2 — `memberCount()` label can overstate the listable/searchable members (packages/dmworkbase/src/Components/Subscribers/vm.ts:75) [source: octo-review-qwen3.7 + octo-review-qwen3.8] (severity split: qwen3.7=P2, qwen3.8=P3)

`orgData?.member_count || this.subscribers.length` feeds the 「查看全部 {{count}} 名成员」/「成员({{count}})」 labels, but the member list/search is backed by the local `subscribers` cache. With a stale server total (channelInfo not refreshed after membership changes) or a paginated super-group subset, the title shows `成员(N)` while the list holds fewer than N — probe-confirmed (member_count=100 with 3 cached rows yields label 100 / list 3). Also `||` treats `member_count === 0` as missing (`??` is marginally more correct). Display-only; introduced by this diff (base had no count label). Fix: reconcile the declared count with the cache state, or label it as approximate; prefer `??`.

### P2 — `onStartEdit` false-gate + permission gates untested (index.tsx:131) [source: octo-review-qwen3.7 + octo-review-glm5.2]

The `if (onStartEdit?.() === false) return;` gate at `index.tsx:131` and the permission gates feeding it (`channelSettingThreadInfoSection.tsx:61-66`, `channelSettingGroupProfileRows.tsx:57-63`/`127-133`) are never exercised — `ChannelSettingInlineEditRow.test.tsx` never passes `onStartEdit` and `channelSettingSections.test.ts` mocks `canRenameThread()=>true` and never sets `isManagerOrCreatorOfMe=false`. A regression dropping the strict `=== false` check (e.g. `if (!onStartEdit?.()) return`) breaks the void-return case and stays green. Confirmed by mutation: mutating `:131` to `if (onStartEdit?.()) return;` leaves the 11 baseline tests green while a focused `onStartEdit=()=>false` test catches it. Fix direction: add a test that passes `onStartEdit=()=>false` and asserts editing is aborted.

### P2 — Save-disabled predicates only partially covered (index.tsx:125) [source: octo-review-qwen3.7 + octo-review-glm5.2] (severity split: qwen3.7=P2/P3, glm5.2=P2)

The save-gating predicates (`exceeded` `:125`, `emptyInvalid` `:126`, `unchanged` `:127`, `saveDisabled` `:128`) are only covered for the `allowEmpty=true` + cleared case. No test asserts Save is disabled when `draft === currentValue` (unchanged), `draft.length > maxCount` (exceeded, used by name/notice/remark rows), or empty with `allowEmpty=false` (thread/group-name/group-notice rows). Confirmed by mutation: zeroing `saveDisabled` to `saving` leaves the 11 baseline tests green while a focused disabled-state test catches all three. Fix direction: add assertions that the Save button is disabled in the unchanged/exceeded/empty-invalid states.

### P2 — `onSave` error/false paths untested (index.tsx:213) [source: octo-review-glm5.2 + octo-review-qwen3.8] (severity split: glm5.2=P2, qwen3.8=P3)

The async save paths at `index.tsx:213` (`if (saved !== false) setCurrentValue(draft)`) and `:215` (`} catch {`, leaves `editing=true` because `setEditing(false)` is only inside the try) have no coverage. No test covers `onSave` resolving to `false` (value must not persist — thread-info returns false on catch, `channelSettingThreadInfoSection.tsx:81`) or `onSave` throwing (editing must stay open, `saving` resets, no component Toast). Confirmed by mutation: dropping `!== false` and relocating `setEditing(false)` into `finally` both break focused tests while the baseline stays green. Fix direction: add tests for the `onSave→false` and `onSave→throw` contracts.

### P2 — Thin test coverage for ChannelSettingInlineEditRow (packages/dmworkbase/src/ui/ChannelSettingRows/__tests__/ChannelSettingInlineEditRow.test.tsx:46) [source: octo-review-qwen3.7 + octo-review-qwen3.8] (severity split: qwen3.7=P2, qwen3.8=P3)

The component has 6+ behavioral branches (maxCount exceeded, `allowEmpty=false` blocking, cancel reverting draft, `onStartEdit` permission gate, silent error catch, multiline TextArea mode, useEffect re-sync) but only 1 test covers the clear-then-save flow. If any untested branch regressed, all existing tests would still pass.

### P2 — `onSave` returning `false` silently closes the editor and discards the draft (packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:214) [source: octo-review-qwen3.8]

`setEditing(false)` runs unconditionally on a resolved promise (`:213-214`), so `onSave` returning `false` CLOSES the editor and reverts the displayed value (typed draft lost), while only a thrown error keeps it open — contradicting the component's own behavior contract ("keeps edit mode open if `onSave` returns false or throws"). Three callers signal failure via `return false` (group name, notice at `channelSettingGroupProfileRows.tsx:71,141`, thread name at `channelSettingThreadInfoSection.tsx:79`), so their failed saves silently exit edit mode and drop the draft, unlike the two `throw` callers. Probe-confirmed: resolve-false → editor closes and shows the old value; reject → editor stays open with the draft. The only component test exercises the happy path. Fix: standardize the failure signal — move `setEditing(false)` into the `saved !== false` branch so `false` == keep-editing, or convert the `return false` callers to `throw` — and pin both branches with tests.

### P2 — data.refresh() failures after successful save propagate to empty catch (packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:222) [source: octo-review-qwen3.7]

When `data.refresh()` throws after a successful API call (network hiccup, timeout), the error propagates into the component's empty catch block. The user sees stale cached data while the server has the new value. Four data flows are affected: remark (refresh throw reaches component catch after Toast), group name and notice (no internal refresh, stale until external propagation), thread name (bridge refresh or caller refresh can throw).

### P2 — Direct mutation of route data before async refresh (packages/dmworkbase/src/features/channelSetting/channelSettingNicknameSection.tsx:44) [source: octo-review-qwen3.7]

The nickname `onSave` directly mutates `data.subscriberOfMe.remark = value` as an optimistic update before calling `data.refresh()`. If refresh fails (silently swallowed per the P1 above), the local mutation persists but other routeData fields remain stale. On component remount, the value prop rebuilds from the stale cache, potentially showing the old value.

### P2 — Mock strips Semi Design behavior (packages/dmworkbase/src/ui/ChannelSettingRows/__tests__/ChannelSettingInlineEditRow.test.tsx:12) [source: octo-review-qwen3.7]

The mock `Input` and `TextArea` components bypass Semi Design's real controlled-input behavior (disabled visual state, `showClear`, suffix positioning inside the input wrapper). The test cannot detect regressions where the component misuses Semi's API.

### P3 — `useEffect([value])` wipes in-progress edits on channelInfo refresh (index.tsx:120) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

The `useEffect([value])` at `index.tsx:120` unconditionally resets both `currentValue` and `draft` whenever the `value` prop changes, so if `channelInfo` refreshes (a WS-driven `data.refresh` or sibling action) while the user is editing, in-progress draft edits are silently wiped — a failure mode the base `InputEdit` route (pushed onto its own route, isolated from parent re-renders) did not have. Confirmed by mutation: HEAD reproduces the wipe (`getByDisplayValue('refreshed-from-server')` holds); guarding the effect with `if (editing) return;` kills the assertion. Fix direction: skip the reset while `editing` is true, or debounce/ignore mid-edit refreshes.

### P3 — Merged `thread.overview` card silently drops the section title (packages/dmworkbase/src/features/channelSetting/channelSettingThreadSections.ts:24) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

`buildThreadOverviewSection` merges `info/md/webhook` via `sections.flatMap(s => s?.rows || [])` and returns `new Section({ rows })`, discarding `Section.title`. The constituent `buildThreadInfoSection` previously returned `new Section({ title: t('base.module.thread.info'), rows })` (`channelSettingThreadInfoSection.tsx:120`), so name/status/parent-group rows used to render under a 「子区信息 / Thread info」 title; the merged card now has no title at all (while this PR simultaneously strengthens the `.wk-section-title` CSS, so titles are a live feature). The new test only asserts `rows.length === 5`, not the title contract. Mechanism and diff-causality are probe- and git-confirmed, but whether dropping the title is a defect depends on design intent and cannot be settled by a test — needs author confirmation of design intent: if the merged card should keep the 「子区信息」 title, restore it explicitly on the merged `Section`; if titleless is intended, consider asserting that. This is **not** a merge blocker — flagging for the author's decision.

### P3 — Vestigial required `inputEditPush` field (packages/dmworkbase/src/features/channelSetting/channelSettingGroupProfileRows.tsx:17) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

`BuildGroupProfileRowsOptions` / `BuildGroupManagementRowsOptions` still declare a REQUIRED `inputEditPush: ChannelSettingInputEditPush`, but the builders no longer destructure/use it after migrating to inline edit. The caller (`channelSettingGroupInfoSection.tsx`) still passes it, so it compiles and runs, but the field is now semantically dead and misleads future callers. Fix: make it optional or remove it (and the caller's argument) in a follow-up.

### P3 — Thread Management section title dropped (packages/dmworkbase/src/features/channelSetting/channelSettingThreadActionsSection.tsx:127) [source: octo-review-qwen3.7]

The old code returned `new Section({ title: t('base.module.thread.management'), rows })`, the new code returns `new Section({ rows })`. Users lose the "Thread management" visual sub-heading.

### P3 — Stale JSDoc reference + dead i18n key after the `thread.overview` rename (packages/dmworkbase/src/module.tsx:1667) [source: octo-review-qwen3.8]

The registration key was renamed `thread.base.info` → `thread.overview`, but the JSDoc at `Service/threadPermission.ts:71` still names the removed `thread.base.info` (comment-only doc rot, no runtime break — a repo-wide scan confirms it is the sole remaining reference; anchored here because `threadPermission.ts` is unchanged). Also `buildThreadActionsSection` dropped its `title` `t('base.module.thread.management')`, leaving that i18n key with zero consumers. Fix: repoint the comment to `thread.overview` and drop the orphaned `thread.management` string (or confirm intent).

### P3 — Orphaned i18n keys (packages/dmworkbase/src/i18n/locales/en-US.json:1057) [source: octo-review-qwen3.8]

`subscribers.viewMore` / `subscribers.memberList` are no longer referenced (replaced by `subscribers.viewAll` / `subscribers.memberListWithCount` at `Subscribers/index.tsx:168,161`); a boundary-aware scan confirms 0 remaining source references. Harmless dead keys in both locales, orphaned by this diff's call-site swap; consider removing.

### P3 — Cancel-revert + multiline TextArea paths untested (ChannelSettingInlineEditRow.test.tsx:62) [source: octo-review-glm5.2]

No test clicks Cancel and asserts `draft` reverts to `currentValue` and `editing=false`; the cancel `onClick` (`index.tsx:198-200`) is never exercised. Group notice uses `multiline=true` with a TextArea `onClear` (`index.tsx:170-178`), but only the single-line Input path is rendered in the component test. Confirmed by mutation: removing `setEditing(false)` from the cancel `onClick` and no-op'ing the TextArea `onClear` both break focused tests while the baseline stays green. Fix direction: add a cancel-revert test and a multiline `onClear` test.

- Permission gates preserved and fail-closed: `isManagerOrCreatorOfMe` / `canEdit` in all `onStartEdit` callbacks; strict `=== false` semantics match base [all three reviewers]
- Security: 0 findings across all three reviews — no `dangerouslySetInnerHTML`/`innerHTML`/`eval`, React JSX auto-escapes, no hardcoded credentials/SSRF/injection, credential scan clean [all three reviewers]
- i18n: new keys have en+zh parity with matching `{{count}}` interpolation; `pnpm i18n:check` passes [all three reviewers]
- Production build passes cleanly (`apps/web build` / `build:e2e`); `git diff --check` clean [qwen3.7 + glm5.2]
- The headline nickname `?? ""` fix is correct and test-effective — reverting it fails the new regression test; a cleared nickname shows 「未设置」 and empty-value saving round-trips [qwen3.8]
- `thread.overview` composition is correct: exactly 5 rows in info → md → webhook order, kills a dropped-builder mutant; returns `undefined` for non-thread channels (no cross-panel leakage or reorder) [qwen3.8]
- Optimistic-UI rollback: the `saved !== false` guard avoids persisting on a false return; `saving`/`disabled` released in `finally`, so the Save button never sticks in loading on throw [glm5.2]
- No new timers, listeners, or AbortControllers — React 18 owns unmount; all bridge actions unmodified, no new attack surface [glm5.2 + qwen3.8]
- `memberCount()` reads the same live `channelInfo` ref as the drawer title (no count/title divergence) [glm5.2]
- CSS uses only `var(--wk-*)` tokens; `maxCount` soft-limit parity with base [glm5.2]
- New dependency `lucide-react` is declared (package.json + lockfile) and resolves [qwen3.8]
- Local test runs green across reviewers: full vitest 236 files / 2061 tests (qwen3.7); Playwright @p0 29/29 + focused vitest 11/11 + 12-assert member-search probe (glm5.2); 15 tests + 2 mutation rounds + focused probes (qwen3.8)

## Verification Results

| Code Review | NEEDS_CHANGES | 2 P1 + 9 P2 + 7 P3 (deduped across 3 reviewers); nickname-save silent failure and dmworkbase CI gap raised as P1 by one reviewer (severity splits noted inline); NEEDS_CHANGES driven by qwen3.8's two P2 failure-path defects |
| E2E Test | PASS | Playwright @p0 29/29 + full vitest 236 files / 2061 tests green (glm5.2, qwen3.7); qwen3.8 validated client-side contracts in an isolated exact-SHA runtime — cross-service PARTIAL by design (frontend-only repo), no contrary evidence |
| Security | CLEARED | All 3 reviewers: 0 findings — permission gates preserved, bridge actions unchanged, no XSS/credential/SSRF issues |

packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
```diff
@@ -84,3 +88,150 @@ export function ChannelSettingActionRow({
...
+  allowEmpty = false,
+  multiline = false,
+  onStartEdit,
+  onSave,
+}: ChannelSettingInlineEditRowProps) {
+  const [editing, setEditing] = useState(false);
+  const [currentValue, setCurrentValue] = useState(value);
+  const [draft, setDraft] = useState(value);
+  const [saving, setSaving] = useState(false);
+  const editingRef = useRef(false);
```
lml2468, line 120: 🔴 仍未修(本 head 与我上轮 CR 的 9e45524d 内容逐字节一致,本次 rebase 是 no-op):此 value-sync effect 在 `value` 变化时**无条件** `setDraft(value)`,无 `editing` 守卫。本 PR 的面板刷新会在用户编辑中传入更新后的 `value` → 正在输入的 draft 被冲掉、输入丢失。修法:`useEffect(() => { setCurrentValue(value); if (!editing) setDraft(value); }, [value, editing]);` + 补『编辑中 value 变化不覆盖 draft』回归测试。

lml2468 (changes requested):
## Re-review: #1075 @ `8ab1a9ae` — refresh information panels and member search

复核当前 live head `8ab1a9ae`。**本 head 与我上轮 REQUEST_CHANGES 的 `9e45524d` 内容逐字节一致**(`git diff 9e45524d..8ab1a9ae` 为空)—— 本次「新提交」是一次 **content-identical rebase(no-op)**,GitHub 因新 SHA 自动 dismiss 了我上轮的 CR,但代码无任何变化。

**结论:维持 需修改 — REQUEST_CHANGES** ❌(上轮的 🔴 未修)

---

### 🔴 阻塞(仍未修)— 编辑中远端 value 变化覆盖用户草稿
`ui/ChannelSettingRows/index.tsx:120-123`:
```js
useEffect(() => { setCurrentValue(value); setDraft(value); }, [value]);
```
`value` 变化时**无条件** `setDraft(value)`、**无 `editing` 守卫**。本 PR 的核心「刷新信息面板」会在用户编辑中重渲染并传入更新后的 `value`(群名/公告远端更新或面板刷新)→ 用户正在输入的 `draft` 被冲掉、输入丢失。与已修好的「保存失败保留草稿」是同一类数据丢失。
- **修法**:`useEffect(() => { setCurrentValue(value); if (!editing) setDraft(value); }, [value, editing]);` + 补「编辑中 value 变化不覆盖 draft」回归测试。

### 承前状态(与 `9e45524d` 一致)
- ✅ 保存失败保留草稿(🔴#1)已修:`if (saved !== false) setEditing(false)`,有测试;
- ✅ 成员搜索为客户端本地过滤(path robotId 锁定、无越权)、权限门保留、无 XSS、`--wk-*` token 定义、i18n 齐全;
- 构建与测试:内容与上轮已验证的 `9e45524d` 逐字节相同(彼时 build ✅ 3.63s、inline-edit 2 tests 绿),本轮无变化故结论沿用。

### 🟡 非阻断(承前 + peers)
- `onSave` await 后 setState 缺卸载守卫;
- yujiawei:搜索失败时全屏错误无搜索框,用户无法改词逃离错误(建议错误态保留搜索框)—— 与草稿一致性同属编辑/错误态健壮性,建议一并处理;
- PR 描述称新增拼音匹配、实际未改;`pr-title-lint`。

---

本次 rebase 未带任何代码改动,上轮的「编辑中 value 变化覆盖草稿」🔴 依旧存在。**维持 REQUEST_CHANGES**(修法:value-sync 加 `!editing` 守卫 + 回归测试;顺带考虑 yujiawei 的搜索错误态保留搜索框)。

Jerry-Xin (dismissed):
Re-review at head 8ab1a9ae — both draft-loss blockers are now resolved and covered by tests. Flipping my prior Request Changes to Approve.

## 🔴 Blocking items — both RESOLVED

1. **Remote-value clobbers in-progress draft (was OPEN at 9e45524d) — FIXED.** `packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx` value-sync effect now guards the draft write: it keeps `setCurrentValue(value)` unconditional but wraps `setDraft(value)` in `if (!editingRef.current)`. A dedicated `useEffect([editing])` keeps `editingRef` in sync. A background `value` refresh arriving while the editor is open no longer overwrites the typed draft; on Cancel / re-open the latest external value surfaces (because `currentValue` still tracks `value`). Byte-verified against the exact `9e45524d..8ab1a9ae` diff.

2. **Save-failure drops draft (fixed at 9e45524d) — no regression.** Save handler still gates on `if (saved !== false) { setCurrentValue(draft); setEditing(false); }`, and the `catch` block leaves the editor open on rejection. Draft is preserved on both `false` return and thrown errors.

## ✅ Highlights

- Regression coverage added for the exact editing-time race: `"preserves an in-progress draft across external value updates"` renders, types a local draft, rerenders with a new external `value`, and asserts the draft survives, then Cancel shows the external value and re-open shows the external value. Two more new tests cover the reject-keeps-draft path and successful-save-visible-until-external-catch-up. Full inline-edit matrix (clear+save, success, `false` failure, reject failure, external-refresh-while-editing, cancel-to-latest) is now exercised.
- Re-review delta is tightly scoped: `9e45524d..8ab1a9ae` touches only the component (+12/-5) and its test (+96). Member/Subscribers files are byte-identical to the previously-reviewed head, so the earlier passes carry forward: member search stays a client-side `localSearch` filter over the authoritative server search + pagination (no unscoped backend dump); permission guards intact; no XSS surface.
- `--wk-*` design tokens used by the inline-edit CSS are all defined and dark-mode safe; i18n parity holds (en-US / zh-CN both 1320 keys, no orphans).

## 💬 Non-blocking notes

- Using a `ref` (`editingRef`) to read `editing` inside the `[value]` effect is intentional and correct here — it avoids re-subscribing the effect on every edit toggle while still reading the live flag. No change requested.

CI: mergeable, required checks green (Build / install-build still in progress at review time, not failing).

mochashanyao (changes requested):
[Octo-Q · automated review]

---

# Code Review — PR #1075 (octo-web)

## Summary

This PR refreshes the channel-setting information panels and adds member search/count display in the subscribers view. The major change is migrating from a push-based edit pattern (navigate to a separate screen) to inline editing via a new `ChannelSettingInlineEditRow` component. Three section builders are consolidated into a single thread-overview section. CSS is cleaned up with class-based styling replacing inline styles, and three new i18n keys are added (both locales in sync).

The inline-edit component itself is well-designed — draft preservation during external updates, clear/save validation gates, and permission checks are all thoughtfully implemented. One P1 issue in the nickname save handler needs addressing before merge.

## Verification

- ✅ **Inline edit component contract** — `ChannelSettingInlineEditRow` correctly handles `onSave` returning `false` (keep editor open), `void` (close), and rejection (keep open). Save button is disabled during save, when exceeded, when empty-invalid, and when unchanged. Draft is preserved during external `value` changes while editing via `editingRef` guard (`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:118-122`).
- ✅ **Permission guards** — Thread name `onStartEdit` checks both `threadInfo` validity and `canRenameThread` permission, showing a Toast on denial (`channelSettingThreadInfoSection.tsx:60-66`). Group name/notice rows use `isManagerOrCreatorOfMe` via `onStartEdit` returning `false` to veto.
- ✅ **i18n parity** — Both `en-US.json` and `zh-CN.json` add the same 3 keys (`common.clear`, `subscribers.viewAll`, `subscribers.memberListWithCount`) with matching `{{count}}` interpolation.
- ✅ **Thread section consolidation** — `buildThreadOverviewSection` correctly flatMaps rows from info/md/webhook builders, handling `undefined` returns via `section?.rows || []` (`channelSettingThreadSections.ts:18-25`).
- ✅ **Subscriber count data flow** — `vm.memberCount()` reads `channelInfo?.orgData?.member_count` with fallback to `subscribers.length` (`vm.ts:74-77`). Consumed in `Subscribers/index.tsx:161-169` for title and "view all" button.
- ✅ **CSS refactoring** — Inline styles in `Subscribers/list.tsx` correctly replaced with CSS classes; role badge classes are conditionally applied based on `GroupRole` enum comparison with safe fallback to empty string.

Static analysis only at head `8ab1a9ae`; build and tests not executed in this environment.

## Findings

One P1 blocker and one P2 issue.

### P1 — Nickname save silently swallows errors (`packages/dmworkbase/src/features/channelSetting/channelSettingNicknameSection.tsx:37`)

The `onSave` handler calls `updateChannelSettingMyGroupNickname` without any try/catch. The bridge function at `channelSettingActions.ts:270-281` is a thin `await` wrapper that does not catch errors either. When the API call fails (network error, server rejection), the error propagates to `ChannelSettingInlineEditRow`'s catch block at `index.tsx:200-201`, which is intentionally empty (comment: "The bridge/container owns the user-facing error message"). The result: the user clicks Save, the spinner stops, the editor stays open, and there is zero feedback that anything went wrong.

The sibling thread-name handler at `channelSettingThreadInfoSection.tsx:72-80` correctly wraps its bridge call in try/catch with `Toast.error` and `return false`. The nickname handler should follow the same pattern:

```ts
onSave: async (value: string) => {
  try {
    await updateChannelSettingMyGroupNickname({
      channel: data.channel,
      remark: value,
    });
    if (data.subscriberOfMe) {
      data.subscriberOfMe.remark = value;
    }
    data.refresh();
  } catch (error: any) {
    Toast.error(error?.msg || t("base.module.thread.saveFailedRetry"));
    return false;
  }
},
```

### P2 — Thread name `onSave` phantom-save on null `threadInfo` (`packages/dmworkbase/src/features/channelSetting/channelSettingThreadInfoSection.tsx:69`)

`if (!threadInfo) return;` returns `undefined`, which per the `ChannelSettingInlineEditRow` contract (`saved !== false` → close editor + update `currentValue`) will close the editor and show the draft as the current value — even though nothing was persisted. Although `onStartEdit` already gates on `threadInfo` and both closures share the same render's capture, a parent re-render that nullifies `threadInfo` between start-edit and save would trigger this path. Fix: change `return;` to `return false;`.

- Clear button uses `onMouseDown` with `preventDefault()` + `stopPropagation()` to avoid stealing focus from the Input — correct pattern for Semi UI.
- `TextArea` uses Semi's built-in `showClear`/`onClear` — asymmetric but both paths work.
- `allowEmpty: true` on nickname correctly permits clearing; `emptyInvalid` check (`!allowEmpty && draft.trim().length === 0`) correctly allows empty drafts when `allowEmpty` is set.
- `displayValue` fallback chain `currentValue === value ? displayValue ?? value : currentValue` correctly handles the optimistic-save transition.
- `inputEditPush` is now unused in `buildThreadInfoSection`, `buildGroupProfileRows`, and `buildGroupManagementRows` — dead parameters in their respective interfaces. Harmless but worth cleaning up.
- QR code icon migrated from static PNG `require()` to `lucide-react` `<QrCode>` — clean.

## Verdict: CHANGES_REQUESTED

The P1 (silent error swallowing in nickname save) is the sole blocker — it leaves users with no feedback when saving their group nickname fails. The P2 (thread name phantom-save) is a one-line fix. Both are straightforward to address.

lml2468 (approved):
## Re-review: #1075 @ `71a32ef2` — ✅ APPROVED（更正:我上轮的 REQUEST_CHANGES 是误判,致歉）

**先认领我的错误**:我上轮对 head `8ab1a9ae` 投的 REQUEST_CHANGES(称「value-sync 草稿覆盖 🔴 未修、diff 为空、与 9e45524d 逐字节一致」)是**误判/假阳性**。真实情况:`9e45524d..8ab1a9ae` 的 diff **确实包含该修复**。我错的根因是**本地 shallow-clone 的 `git fetch origin 8ab1a9ae` 静默失败**,`git checkout` 实际停在旧 SHA,我又没用 `rev-parse` 复核 HEAD,导致把 9e45524d 和自己 diff → 空 → 误判「未修」。Steve 判对、我判错,向作者与各位致歉。

已在当前 live head `71a32ef2` 用 API/git 逐字节复核确认。

**结论:APPROVE ✅**

### 🔴→✅ 编辑中远端 value 变化覆盖草稿 —— 已修(自 `8ab1a9ae` 起即在)
`ui/ChannelSettingRows/index.tsx`:
```js
const editingRef = useRef(false);
useEffect(() => { editingRef.current = editing; }, [editing]);
useEffect(() => {
  setCurrentValue(value);
  if (!editingRef.current) { setDraft(value); }   // ← 编辑中不覆盖草稿
}, [value]);
```
`setCurrentValue(value)` 保持无条件(Cancel/重开显示最新外部值),`setDraft(value)` 仅在**非编辑态**执行 → 编辑中远端刷新不再冲掉用户输入。正是建议的修法。
- **回归测试到位**:`:140`「preserves an in-progress draft across external value updates」(编辑中 rerender 新 value → 草稿存活)+ `:172`「keeps a successful save visible until the external value catches up」。

### 承前无回归
- ✅ 保存失败保留草稿(🔴#1):`if (saved !== false)` 才关闭(:221)+ 「saving fails/rejects」两测;
- ✅ 成员搜索客户端 localSearch(不越权)、权限门(isManagerOrCreatorOfMe/canEdit)保留、无 XSS、`--wk-*` token 定义暗色安全、i18n 无孤儿。

**构建 / 测试(实跑 @ `71a32ef2`)**:`pnpm --filter @octo/web build` ✅ 3.43s;`ChannelSettingInlineEditRow` **9 tests 全绿**(含草稿存活/保存失败/保存拒绝/成功可见性等)。`mergeable=true`。

---

### 🟡 非阻断
- `onSave` await 后 setState 缺卸载守卫(既有建议);
- yujiawei:搜索失败全屏错误无搜索框,建议保留搜索框便于改词逃离错误。

---

两条草稿丢失(保存失败 + 编辑中远端覆盖)均已修复并各有回归测试,授权/XSS/token/i18n 无回归,构建与 9 测试通过。**可合并 —— APPROVE**。再次为上轮的误判致歉——教训:跨 head 复核必须 `rev-parse` 确认 HEAD、不信任可能静默失败的 shallow fetch。

Jerry-Xin (approved):
Re-review at head `71a32ef2` — re-affirming APPROVE. Both draft-loss blockers remain fixed and are now regression-tested; the new commit only adds save-failure error handling for the group-nickname row plus tests and i18n keys. No regressions found.

### ✅ Verified at this head
- **Draft not clobbered by external refresh** (`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:126-131`): the value-sync effect guards `setDraft(value)` behind `if (!editingRef.current)`, using a ref (not the `editing` state) so an in-progress draft survives external `value` updates while still tracking `currentValue` for Cancel-to-latest. Note: the edited inline-edit component in this PR is the dmworkbase `ChannelSettingRows/index.tsx`, not `packages/dmloop/src/ui/InlineEdit.tsx` (the latter is unchanged by this PR).
- **Save failure keeps the draft** (`ChannelSettingRows/index.tsx:220-223`): `setEditing(false)`/`setCurrentValue(draft)` only run when `saved !== false`; a `false` return or a thrown error keeps the editor open with the typed draft. `setSaving(false)` runs in `finally`.
- **Nickname row now propagates failures** (`packages/dmworkbase/src/features/channelSetting/channelSettingNicknameSection.tsx:38-60`): `onSave` wraps `updateChannelSettingMyGroupNickname` in try/catch, surfaces the server `msg` (else a localized fallback) via `Toast.error`, and returns `false`, wiring the row-level save gate above end-to-end.
- **Tests**: focused suite passes locally — `ChannelSettingInlineEditRow.test.tsx` (9) + `channelSettingSections.test.ts` (11) = 20/20. Key cases present: "preserves an in-progress draft across external value updates", "keeps the draft open when saving fails" / "…when saving rejects", and the new "keeps the group nickname unchanged and reports a failed save".
- **i18n parity**: `module.channelSettings.saveFailedRetry` added to both `en-US.json` and `zh-CN.json`.
- **CSS tokens**: every `--wk-*` used in `ChannelSettingRows/index.css` is defined in `semantic.css`; relies on global theme-flipping tokens, so no dark-mode gap.
- **Member search / permissions**: `Subscribers` client-side `localSearch` and permission paths are unchanged from the previously approved head; no unscoped member dump introduced.
- **Delta `8ab1a9ae..71a32ef2`** is additive only (nickname try/catch + i18n keys + test additions/formatting); no source regression.

### 🟡 Non-blocking
- **CI coverage blind spot**: `.github/workflows/ci.yml` Build job runs only a hardcoded `@octo/docs` vitest set plus i18n/build/lint — it does not run the `dmworkbase` vitest suite, so the inline-edit regression tests added here don't execute at the CI gate. They pass locally (verified) and the source is correct, so nothing is hidden today, but future regressions in these files won't be caught by CI. Worth a follow-up to include the dmworkbase package tests in the pipeline.

### ✅ Highlights
- Using a ref for the editing flag in the value-sync effect avoids stale-closure pitfalls and is more robust than gating on the `editing` state alone.
- Error handling narrows the unknown to a typed `msg` before display and falls back cleanly; no HTML injection surface (Toast renders text).

Candice00711:
Closing to replace this PR because a stale automated CHANGES_REQUESTED review remained attached after its blocking findings were fixed and later reviews approved the latest head.

yujiawei (approved):
# Code Review — PR #1075 (octo-web)

**Verdict: APPROVE.** This is a well-scoped, well-tested presentation-layer refactor of the channel-setting information panels and member views. Behavior is preserved; the two functional fixes (cleared group nickname, draft preservation) are correct and covered by tests. No P0/P1 blockers found. A few non-blocking nits below.

## 1. Spec compliance

Verified the diff against the PR's four summary bullets and its explicit scope boundaries.

- **Missing / under-built:** none. All four goals are implemented — card-based info panels + wider drawer (`--wk-wdith-chat-channelsetting: 28rem`, new `.wk-channelsetting-content` CSS), reusable inline-edit rows (`ChannelSettingInlineEditRow`), refined member grid/list, and draft-preservation on failed save / external refresh.
- **Over-built:** none. Scope boundaries hold:
  - No changes under `bridge/channelSetting/*Actions` — `channelSettingActions.ts` is untouched. (`channelSettingThreadActionsSection.tsx` lives under `features/`, not the excluded `bridge` path.)
  - No changes to WKApp/WKSDK/datasource/Chat/Conversation/Messages runtime paths.
  - No new subzone switches, hidden entries, or permission behavior — manager/creator gating is preserved verbatim (`onStartEdit` returns `false` and toasts, mirroring the old early-return).
  - No new pinyin-search capability — `Subscribers` search / authoritative server search / pagination are behaviorally unchanged; only presentational classes and a member-count label were added.
- **Deviation:** none. The thread-settings consolidation (`thread.base.info` + `thread.md.setting` + `thread.webhook` → single `thread.overview` via `buildThreadOverviewSection`) is presentation-only: each original builder is still called and its rows are flat-mapped, so all internal gating (active-status, disbanded, parent-manager for webhook) is retained.

**Spec: ✅**

## 2. Code quality

**Quality: Approved**

Strengths:
- `ChannelSettingInlineEditRow` state machine is correct: `editingRef` guards the external-value effect so an in-flight draft is never clobbered by a background channel refresh; `onSave` returning `false` or rejecting keeps the editor open with the typed draft; cancel restores the latest external value. All five branches are covered by the new `ChannelSettingInlineEditRow.test.tsx`.
- The group-nickname clear fix is real and correct: `data.subscriberOfMe?.remark ?? ""` (was `remark || name`) now preserves a cleared empty string instead of falling back to the member name, with `displayValue` showing `notSet`. Regression-tested in `channelSettingSections.test.ts`.
- Save button correctly disabled on `saving | exceeded | emptyInvalid | unchanged`; the clear affordance uses `onMouseDown` + `preventDefault` to avoid blurring the input.
- i18n: new `common.clear`, `subscribers.viewAll`, `subscribers.memberListWithCount` added to both `en-US` and `zh-CN`; `{{count}}` matches the `interpolate()` token format.
- New CSS classes are all defined and scoped under `.wk-channelsetting-content` to avoid leaking into other `Sections`/`ListItem` callers.

No correctness, concurrency, data-loss, or trust-boundary issues found in the changed lines.

### Nits (non-blocking, P2)

- **Orphaned i18n keys.** `subscribers.memberList`, `subscribers.viewMore`, and `module.thread.management` no longer have any code references after this PR but remain in both locale files. `i18n:check` only flags *missing* keys (not unused ones), so this doesn't break the build — but consider removing them to prevent drift.
- **`memberCount()` fallback.** `SubscribersVM.memberCount()` returns `orgData?.member_count || this.subscribers.length`. When the true member count is genuinely `0` (edge case), `||` falls through to the cached-subscribers length. Harmless here since a 0-member group won't reach the "view all" affordance, but `??` would be marginally more precise.
- **Transient display after save.** After a successful save, the row briefly shows the raw saved text (`currentValue`) until the external `value` prop catches up, at which point it reverts to the formatted `displayValue` (e.g. the external-group tag). This is intentional and test-covered; noting only for awareness.

## 3. Overall verdict

**APPROVE** — Spec ✅ and Quality Approved. Clean, behavior-preserving refactor with solid test coverage of the tricky inline-edit and clear-nickname paths.

yujiawei (changes requested):
# Code Review — PR #1075 (octo-web)

**Updated verdict: CHANGES_REQUESTED.** This supersedes my earlier Approve on the same head SHA (`8ab1a9ae`). On a second, deeper trace of the error paths I confirmed a real P1 that my first pass missed: the group-nickname inline save fails completely silently. The refactor is otherwise well-scoped and well-tested; this is a targeted fix, not a redesign.

## 1. Spec compliance

Unchanged from prior review — all four PR goals implemented, scope boundaries respected (no `bridge/*Actions`, WKApp/WKSDK/datasource, no new switches/permissions/pinyin search), and the `thread.overview` consolidation is presentation-only with each folded builder's gating preserved.

**Spec: ✅**

## 2. Code quality

### P1 — Group-nickname save fails silently (`packages/dmworkbase/src/features/channelSetting/channelSettingNicknameSection.tsx:39-48`)

The nickname `onSave` awaits `updateChannelSettingMyGroupNickname` with **no try/catch**. I traced the full failure path to confirm nothing else catches it:

1. `channelSettingNicknameSection.tsx` `onSave` — no catch.
2. `updateChannelSettingMyGroupNickname` (`channelSettingActions.ts:270`) — thin `await runtime.updateSubscriberAttr(...)`, no catch.
3. `updateChannelSubscriberAttr` (`ChannelSettingService.ts:123`) — raw `APIClient.shared.put(...)`, no catch.
4. `APIClient` response interceptor (`APIClient.ts:105`) — **normalizes and `Promise.reject`s; it does not Toast** (it only invokes `logoutCallback` on auth-expiry). There is no global error toast.
5. The rejection therefore lands in `ChannelSettingInlineEditRow`'s intentionally-empty catch (`index.tsx:225-226`), whose comment states *"The bridge/container owns the user-facing error message."*

For this one handler, **no layer owns that message.** On any save failure (network error, server rejection, 20s timeout) the spinner stops, the editor stays open, and the user gets zero feedback. Every sibling inline-edit handler does this correctly — group name and notice (`channelSettingGroupProfileRows.tsx:71-73, 141-143`), remark (`channelSettingGroupManagementRows.tsx:223-225`), and thread name (`channelSettingThreadInfoSection.tsx:78-81`) all `Toast.error` on failure. The nickname handler is the sole outlier.

Fix — mirror the thread-name handler:

```ts
onSave: async (value: string) => {
  try {
    await updateChannelSettingMyGroupNickname({
      channel: data.channel,
      remark: value,
    });
    if (data.subscriberOfMe) {
      data.subscriberOfMe.remark = value;
    }
    data.refresh();
  } catch (error: any) {
    Toast.error(error?.msg || t("base.module.thread.saveFailedRetry"));
    return false;
  }
},
```

Returning `false` also keeps the editor open with the typed draft, consistent with the component contract and the other handlers.

### P2 — Thread-name `onSave` closes the editor without persisting on null `threadInfo` (`channelSettingThreadInfoSection.tsx:70`)

`if (!threadInfo) return;` resolves to `undefined`, which per the row contract (`saved !== false` → close editor + adopt draft as `currentValue`) closes the editor and shows the unsaved draft as if it were saved. `onStartEdit` already gates on `threadInfo` and both closures capture the same render, so this is only reachable if a parent re-render nullifies `threadInfo` between start-edit and save — low likelihood, but the correct guard is free. Change `return;` to `return false;`.

## 3. Overall verdict

**REQUEST_CHANGES** — Spec ✅ but Quality Changes-Requested. The P1 silent-failure is the blocker (users get no feedback when a nickname save fails); P2 is a one-line hardening. Both are small, localized fixes.

### Non-blocking nits (carried from prior review)

- Orphaned i18n keys now unreferenced: `subscribers.memberList`, `subscribers.viewMore`, `module.thread.management` (build-safe; `i18n:check` only flags missing keys).
- `SubscribersVM.memberCount()` uses `||` where `??` would be marginally more precise for a genuine `0` count.
- `inputEditPush` is now an unused parameter in `buildThreadInfoSection`, `buildGroupProfileRows`, and `buildGroupManagementRows` — dead params worth cleaning up.

PR #1065, opened 2026-07-24
fix: member selector works without selecting a chat first

quanming1:
## Problem

PR #1044 required selecting a chat before the "Select Participants" button would work — clicking it without a chat selected showed a Toast and returned silently.

## Fix

- Removed the `if (!chat) return` guard from `handleOpenMemberSelector`
- When no chat is selected, `ChatSelectorModal` loads global contacts via `WKApp.dataSource.contactsList` instead of channel subscribers
- Member tabs (managers/normal members) only show when a channel is present (they require group roles)
- Without a chat, only the "All Members" tab is shown
- Added virtual scrolling (zero deps) for large contact lists
- Bot filter uses `robot` field (Contacts type) + `isBot()` helper
- Blacklisted contacts filtered out (`status !== 2`)
- `reqSeq` guard added to `loadMembers()` to prevent stale writes
- Virtual scroll window reset on modal reopen, tab/keyword change, and archived toggle
- Members mode shows empty state when list is empty

## Files Changed

- `packages/dmworksummary/src/pages/SummaryCreatePage.tsx` — removed guard, pass null channel, mode gate for participants
- `packages/dmworksummary/src/components/ChatSelectorModal.tsx` — loadMembers handles null channel, conditional tabs, virtual scrolling, bot/blacklist filter, reqSeq guard

PR #1071, opened 2026-07-24
feat: restore Agent summary mode switch entrypoint

quanming1:
## Problem

PR #1044 (Summary UI redesign) removed the `SplitButtonGroup` that allowed switching between Normal and Agent summary modes. This left Agent summary — a production feature (PR #992 by wanyiwang06) — completely inaccessible from the left sidebar summary entry point. The `AgentChatPanel` import, `handleSelectMode` method, and `mode: 'agent'` branch were left as dead code.

## Fix

- Add a segmented control (开始总结 / Agent 总结) in the create page header
- Active state uses Semi design tokens (`var(--semi-color-bg-1)`, `var(--semi-color-text-0)`)
- `handleSelectMode('normal'/'agent')` is now called from the new UI
- Agent mode renders `AgentChatPanel` with `selectedChannels={selectedChats}` (PR #992 agent context passing preserved)
- Normal mode shows textarea + templates + select chat/participants + start button

## Design

The segmented control fits the Figma design language better than the old `SplitButtonGroup` dropdown:
- Pill-shaped container with `var(--semi-color-fill-0)` background
- Active button: `var(--semi-color-bg-1)` background + `var(--semi-color-text-0)` text
- Non-active: transparent + `var(--semi-color-text-2)` text, hover → `var(--semi-color-text-1)`

## Files Changed

- `packages/dmworksummary/src/pages/SummaryCreatePage.tsx` — add mode switch JSX in header
- `packages/dmworksummary/src/index.css` — add mode switch CSS using Semi design tokens

PR #1079, opened 2026-07-24
fix: member selector works without selecting a chat first

quanming1:
## Summary

When creating a summary without selecting a chat first, the participant selector (`ChatSelectorModal`) now works in a standalone "members" mode — loading contacts directly instead of requiring a channel.

Fixes the member-selector regression introduced by the #1044 summary UI redesign.

## Changes

- **ChatSelectorModal.tsx**: `mode === "members"` path loads contacts via `contactsList` (capped at 200) instead of requiring `this.props.channel`. Bot and blacklisted users are filtered.
- **Virtual scroll**: resets `visibleStart` / `visibleEnd` / `scrollTop` on all viewport-changing actions (reopen modal, switch tab, change keyword, toggle archived).
- **Single-tab hide**: when no channel is selected, only one tab ("all members") exists — the tabs row is hidden to avoid visual noise.
- **i18n**: removed unused `list.emptyDescription` key breaking locale parity.

## Supersedes

Closes #1065 (recreated to clear dismissed review history on the new commit).

### DIFF
--- a/packages/dmworksummary/src/components/ChatSelectorModal.tsx
+++ b/packages/dmworksummary/src/components/ChatSelectorModal.tsx
@@ -91,6 +91,7 @@ export default class ChatSelectorModal extends Component<Props, State> {
 
     componentDidUpdate(prevProps: Props) {
         if (this.props.visible && !prevProps.visible) {
+            if (this.listScrollRef.current) this.listScrollRef.current.scrollTop = 0;
             if (this.props.mode === "members") {
                 this.setState({
                     localSelectedMembers: [...(this.props.selectedMembers ?? [])],
@@ -200,15 +201,18 @@ export default class ChatSelectorModal extends Component<Props, State> {
     }
 
     handleIncludeArchivedChange = (checked: boolean) => {
+        if (this.listScrollRef.current) this.listScrollRef.current.scrollTop = 0;
         this.setState({ includeArchived: checked, visibleStart: 0, visibleEnd: 20 });
         this.loadCandidates(checked);
     };
 
     handleKeywordChange = (val: string) => {
+        if (this.listScrollRef.current) this.listScrollRef.current.scrollTop = 0;
         this.setState({ keyword: val, visibleStart: 0, visibleEnd: 20 });
     };
 
     handleTabChange = (tab: string) => {
+        if (this.listScrollRef.current) this.listScrollRef.current.scrollTop = 0;
         this.setState({ activeTab: tab as State["activeTab"], visibleStart: 0, visibleEnd: 20 });
     };
 
--- a/packages/dmworksummary/src/components/ChatSelectorModal.tsx
+++ b/packages/dmworksummary/src/components/ChatSelectorModal.tsx
@@ -96,10 +96,12 @@ export default class ChatSelectorModal extends Component<Props, State> {
                     localSelectedMembers: [...(this.props.selectedMembers ?? [])],
                     keyword: "",
                     activeTab: "all_members",
+                    visibleStart: 0,
+                    visibleEnd: 20,
                 });
                 this.loadMembers();
             } else {
-                this.setState({ localSelected: [...this.props.selected], keyword: "", activeTab: "followed", includeArchived: false });
+                this.setState({ localSelected: [...this.props.selected], keyword: "", activeTab: "followed", includeArchived: false, visibleStart: 0, visibleEnd: 20 });
                 this.loadCandidates(false);
             }
         }
@@ -107,12 +109,14 @@ export default class ChatSelectorModal extends Component<Props, State> {
 
     async loadMembers() {
         const channel = this.props.channel;
+        const seq = ++this.reqSeq;
         this.setState({ loading: true });
         try {
             if (channel) {
                 // 有选中聊天：加载该群聊成员
                 const sdk = WKSDK.shared();
                 await sdk.channelManager.syncSubscribes(channel);
+                if (seq !== this.reqSeq) return;
                 const subscribers = sdk.channelManager.getSubscribes(channel) || [];
                 const humans = subscribers.filter((m: any) => !m.is_bot && !isBot(m.uid));
                 const roles = new Map<string, number>();
@@ -131,10 +135,13 @@ export default class ChatSelectorModal extends Component<Props, State> {
             } else {
                 // 无选中聊天：加载全局联系人
                 const list: any[] = (WKApp.dataSource as any)?.contactsList ?? [];
+                if (seq !== this.reqSeq) return;
                 const humans = list.filter((m: any) => {
                     // Contacts 类型用 robot 字段，不是 is_bot
                     const isRobot = m.robot === true || m.is_bot === true || isBot(m.uid || m.user_id || "");
-                    return !isRobot;
+                    // 过滤黑名单联系人 (ContactsStatus.Blacklist = 2)
+                    const isBlacklisted = m.status === 2;
+                    return !isRobot && !isBlacklisted;
                 });
                 this.setState({
                     memberRoles: new Map<string, number>(),
@@ -147,8 +154,10 @@ export default class ChatSelectorModal extends Component<Props, State> {
                 });
             }
         } catch {
+            if (seq !== this.reqSeq) return;
             this.setState({ candidates: [] });
         } finally {
+            if (seq !== this.reqSeq) return;
             this.setState({ loading: false });
         }
     }
@@ -191,7 +200,7 @@ export default class ChatSelectorModal extends Component<Props, State> {
     }
 
     handleIncludeArchivedChange = (checked: boolean) => {
-        this.setState({ includeArchived: checked });
+        this.setState({ includeArchived: checked, visibleStart: 0, visibleEnd: 20 });
         this.loadCandidates(checked);
     };
 
@@ -531,15 +540,19 @@ export default class ChatSelectorModal extends Component<Props, State> {
                                 {loading ? (
                                     <div className="chat-selector-loading"><Spin /></div>
                                 ) : mode === "members" ? (
-                                    <>
-                                        <div style={{ height: displayList.length * 40, position: 'relative' }}>
-                                            {displayList.slice(this.state.visibleStart, this.state.visibleEnd).map((entry, i) => (
-                                                <div key={entry.item.chat_id} style={{ position: 'absolute', top: (this.state.visibleStart + i) * 40, left: 0, right: 0, height: 40 }}>
-                                                    {this.renderMemberItem(entry)}
-                                                </div>
-                                            ))}
-                                        </div>
-                                    </>
+                                    displayList.length === 0 ? (
+                                        <Empty description={t("summary.chatSelector.noData")} />
+                                    ) : (
+                                        <>
+                                            <div style={{ height: displayList.length * 40, position: 'relative' }}>
+                                                {displayList.slice(this.state.visibleStart, this.state.visibleEnd).map((entry, i) => (
+                                                    <div key={entry.item.chat_id} style={{ position: 'absolute', top: (this.state.visibleStart + i) * 40, left: 0, right: 0, height: 40 }}>
+                                                        {this.renderMemberItem(entry)}
+                                                    </div>
+                                                ))}
+                                            </div>
+                                        </>
+                                    )
                                 ) : displayList.length === 0 ? (
                                     <Empty description={t("summary.chatSelector.noData")} />
                                 ) : (
--- a/packages/dmworksummary/src/components/ChatSelectorModal.tsx
+++ b/packages/dmworksummary/src/components/ChatSelectorModal.tsx
@@ -516,17 +516,19 @@ export default class ChatSelectorModal extends Component<Props, State> {
                                     onChange={(e) => this.handleKeywordChange(e.target.value)}
                                 />
                             </div>
-                            <div className="chat-selector-tabs">
-                                {currentTabs.map((tab) => (
-                                    <button
-                                        key={tab.key}
-                                        className={`chat-selector-tab${activeTab === tab.key ? " chat-selector-tab--active" : ""}`}
-                                        onClick={() => this.handleTabChange(tab.key)}
-                                    >
-                                        {tab.label}
-                                    </button>
-                                ))}
-                            </div>
+                            {currentTabs.length > 1 && (
+                                <div className="chat-selector-tabs">
+                                    {currentTabs.map((tab) => (
+                                        <button
+                                            key={tab.key}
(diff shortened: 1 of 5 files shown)


Jerry-Xin (approved):
Clean recreation of #1065 with a single cosmetic follow-up; both prior blockers stay resolved. APPROVE.

**Relationship to #1065:** This PR supersedes/recreates #1065 (per PR body, to clear its dismissed-review history on a fresh commit). Of the 4 changed files, 3 are byte-identical to #1065's final approved tree (both i18n files + `SummaryCreatePage.tsx`). The only real delta is in `packages/dmworksummary/src/components/ChatSelectorModal.tsx`.

**The one real change (single-tab hide):** The tabs row is now wrapped in `currentTabs.length > 1 && (...)`, so when the member selector opens in standalone members-mode with no channel (`memberTabs` = just `all_members`), the single tab is hidden. This is purely cosmetic:
- No-channel members-mode: `activeTab` is forced to `all_members` on open, the filter returns all contacts for `all_members`, and there is no way to change the tab while it is hidden — behavior unchanged, list still populates.
- With-channel members-mode (3 tabs) and chat-mode (4 tabs) still render tabs as before — no regression.

**✅ Highlights (unchanged from the #1065 approval, verified byte-identical):**
- Contacts source valid: no-channel branch reads `WKApp.dataSource.contactsList` (capped at 200), a real populated DataSource property — not an empty-keyword single-user lookup.
- Virtual-scroll viewport reset complete: `visibleStart`/`visibleEnd` reset to 0/20 AND `listScrollRef.current.scrollTop = 0` on all viewport-changing paths — reopen, tab change, keyword change, archived toggle.
- Bot field correct: filter uses `robot === true` on the Contacts type with `is_bot`/UID fallback.
- No scope-creep: the Agent mode-switch feature is not bundled here (it lives in #1071).
- No XSS: member names render as React text nodes (`{item.name}` / `{m.name}`), auto-escaped.
- i18n parity preserved (removed the unused `list.emptyDescription` key).

**💬 Non-blocking:**
- CI: e2e-p0 was still IN_PROGRESS at review time (the earlier e2e run passed); all other required checks green, mergeable. Confirm the in-progress run finishes green before merge.

No blocking items.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1079 (octo-web)

## Summary

This PR fixes the member-selector regression from the #1044 summary UI redesign. When no chat is selected, the `ChatSelectorModal` now operates in a standalone "members" mode — loading contacts via `WKApp.dataSource.contactsList` instead of requiring a channel. It also adds virtual scrolling for both the members and chat lists, resets scroll position on all viewport-changing actions, hides the tab row when only one tab exists, and removes an unused i18n key.

The implementation is solid: the data flow from `contactsList` through bot/blacklist filtering to virtual-scroll rendering is correct, race guards (`reqSeq`) are properly extended to the new synchronous contacts path, and the i18n cleanup has no dangling references. Two minor nits below; no correctness or functional concerns.

## Verification

- ✅ **Data flow — contactsList** — `WKApp.dataSource.contactsList` is a class property on `DataSource` (`packages/dmworkbase/src/Service/DataSource/DataSource.ts:17`), populated via `contactsSync()` during app lifecycle. Same access pattern used in `AssigneeEditor` and `GlobalSearch`. The `?? []` fallback handles the unpopulated edge case.
- ✅ **Bot/blacklist filter correctness** — `ContactsStatus.Blacklist = 2` confirmed at `DataSource.ts:65`. The `robot` field is `boolean` on the `Contacts` class. Filter logic at `ChatSelectorModal.tsx:141-144` correctly handles both `robot` and `is_bot` field names plus `isBot()` uid-pattern fallback.
- ✅ **Race guards** — `loadMembers()` increments `reqSeq` at entry and checks `seq !== this.reqSeq` after every await and in catch/finally. The synchronous contacts path cannot race, but the guards are harmless and consistent with the async channel path.
- ✅ **Virtual scroll** — Both members and chat modes use the same virtual-scroll pattern: container height = `displayList.length * 40`, absolute-positioned items sliced by `visibleStart`/`visibleEnd`. `ITEM_HEIGHT = 40` matches the CSS (4px padding + 28px avatar + 4px padding = 36px item, 40px wrapper). `handleListScroll` resets on all viewport-changing actions (reopen, tab change, keyword change, archived toggle).
- ✅ **i18n cleanup** — `selectChatFirst` removed from both `en-US.json` and `zh-CN.json`; grep confirms zero remaining references in the `dmworksummary` package.
- ✅ **Tab visibility** — When `channel` is null, `memberTabs` contains only `all_members`. The `currentTabs.length > 1` guard at `ChatSelectorModal.tsx:520` correctly hides the tab row.
- ✅ **C1 — dual-path parity** — `loadMembers` with-channel and without-channel paths both set `candidates` and `memberRoles` symmetrically. `componentDidUpdate` resets `visibleStart`/`visibleEnd` in both branches.

Static analysis only at head `e157f278596be375fba4799f3a37a7038c82db73`; build and tests not executed in this environment.

## Findings

### Nit — Hardcoded `ITEM_HEIGHT` decoupled from CSS (`packages/dmworksummary/src/components/ChatSelectorModal.tsx:73`)

`ITEM_HEIGHT = 40` is hardcoded in the component but the actual item height is determined by CSS (`.chat-selector-item` padding + avatar size + wrapper div height). If the CSS changes, virtual scroll positioning will silently produce gaps or overlaps. Consider deriving the value from a CSS custom property or co-locating the constant.

### Nit — Start button renders as no-op in agent mode (`packages/dmworksummary/src/pages/SummaryCreatePage.tsx:1339`)

The `mode !== 'agent'` guard around the "Start summary" button was removed. In agent mode, `handlePrimaryClick` is a no-op (line 747–749). Since the mode-switch dropdown is also removed, agent mode is only reachable via `derivedFromTask` at mount (where `topic` is empty → `canSubmit()` returns false → button disabled). No user-visible impact today, but the dead render path could confuse future maintainers.

## Data Flow Trace

|---|---|---|
| `WKApp.dataSource.contactsList` | `DataSource.contactsList` property, populated by `contactsSync()` during app init | ✅ Same pattern as `AssigneeEditor`, `GlobalSearch`; `?? []` fallback for undefined |
| `m.robot`, `m.is_bot`, `m.status` | `Contacts` class (`DataSource.ts:69-93`) — `robot: boolean`, `status: ContactsStatus` | ✅ Filter logic matches type definitions; `ContactsStatus.Blacklist = 2` |
| `m.uid \|\| m.user_id \|\| m.id` | `Contacts.uid` (primary); fallbacks for non-standard data shapes | ✅ `Contacts` class has `uid`; fallbacks handle `as any` cast edge cases |
| `memberRoles` (empty Map when no channel) | Set to `new Map()` in no-channel path (`ChatSelectorModal.tsx:151`) | ✅ Only `all_members` tab exists when no channel; role-filtering branches never reached |
| `visibleStart`/`visibleEnd` | Reset to `0`/`20` on all viewport-changing actions; updated by `handleListScroll` | ✅ Both members and chat modes use virtual scroll; scroll handler attached to shared container |
| `memberSelectorChannel` (can be null) | `SummaryCreatePage.handleOpenMemberSelector` sets `null` when no chat selected | ✅ `Props.channel` type is `Channel \| null \| undefined`; modal handles null correctly |

## Blind Spot Checklist

- **C1 (dual-path parity)** — CLEAR. `loadMembers` with-channel and without-channel paths both set `candidates` and `memberRoles` symmetrically. `componentDidUpdate` resets virtual scroll state in both `mode === "members"` and chat branches.
- **C2 (control-flow ordering)** — CLEAR. `handleListScroll` is called on every scroll event but gated by `if (start !== this.state.visibleStart || end !== this.state.visibleEnd)` to prevent unnecessary setState. No double-invocation or nesting risk.
- **C3 (authorization boundary)** — N/A. No authorization or permission changes.
- **C4 (authorization lifecycle)** — N/A.
- **C5 (build/runtime parity)** — N/A. No build/config changes.
- **C6 (governance/docs)** — N/A. No docs or policy changes.

- `memberSelectorChannel` can now be `null` when passed to the modal — `Props.channel` type is `Channel | null | undefined`, handled correctly.
- The `getDisplayList()` filter logic in members mode uses `memberRoles` (empty Map when no channel) — since only the `all_members` tab exists in that case, the role-filtering branches are never reached.
- Virtual scroll for short lists (< 20 items): `displayList.slice(0, 20)` returns all items; container height is correct.
- `handleMemberSelectorConfirm` in `SummaryCreatePage.tsx` reads `memberSelectorSelectedItems` correctly regardless of whether items came from channel members or global contacts.
- `renderMemberItem` and `renderItem` both use `key={item.chat_id}` — unique per entry given the `uid || user_id || id` fallback chain.
- The `chat-selector-item--indent` CSS class only adds `padding-left: 40px` — does not affect item height, so virtual scroll `ITEM_HEIGHT = 40` remains correct for indented thread entries.

## Verdict: APPROVED

Well-scoped fix that correctly addresses the member-selector regression. The contacts data flow, race guards, virtual scrolling, and i18n cleanup are all sound. Only minor maintainability nits; no correctness or functional concerns.

[Octo-Q] verdict: APPROVE — no P0/P1 findings; two nits only.

yujiawei (changes requested):
# Code Review — PR #1079 (octo-web)

**Head SHA reviewed:** `e157f278596be375fba4799f3a37a7038c82db73`
**Base:** `main` · Merge-base: `503a5ea` · Files: 4 (`+135 / −62`)

Thanks for the fix — the standalone "members" mode and the virtual-scroll reset handling are solid. There is, however, one blocking issue that only surfaces **after this branch merges into the current `main`**, so it is invisible in the isolated PR diff. Details below.

---

## 1. Spec compliance

The stated goal — *"member selector works without selecting a chat first"* — is implemented correctly in `ChatSelectorModal.tsx`:

- `mode === "members"` with no `channel` now loads global contacts from `WKApp.dataSource.contactsList`, filtering bots (`robot` field) and blacklisted users (`status === 2`, consistent with the existing `UserSelect` / `AssigneeEditor` convention). ✅
- `handleOpenMemberSelector` no longer early-returns with a toast when no chat is selected; it passes `channel: null`. ✅
- The now-unused `selectChatFirst` i18n key is removed from both locales, and no reference remains repo-wide (verified). ✅
- Single-tab hide, virtual scroll, and viewport-reset-on-all-mutations are reasonable additions.

**However, one change is out of scope and introduces a regression** (see §2, Blocker). The `mode !== 'agent'` render gate around the primary **Start** button was deleted. This gate does not exist at the branch's merge-base (`503a5ea`), so the PR diff looks clean — but `main` re-introduced the user-facing Agent/Normal mode switch in **#1071** (`e28c75b`) *after* this branch forked. Deleting the gate is therefore a semantic revert of #1071's behavior that a text-only merge will not flag.

---

## 2. Blocker (must fix before merge)

### 🔴 B1 — Start button becomes a dead, misleadingly-enabled control in Agent mode after merge

**File:** `packages/dmworksummary/src/pages/SummaryCreatePage.tsx:1335` (the un-gated `<Button className="summary-workbench-start-btn">`)

At this branch's merge-base the Start button was wrapped in `{mode !== 'agent' && ( ... )}`. This PR removes that wrapper. The problem is only visible post-merge:

- Current `main` (#1071, `e28c75b`) renders a visible **Normal / Agent** mode switch in the header (`summary-workbench-mode-switch`, calling `handleSelectMode`), which this branch's merge-base did not have.
- The action bar containing the Start button lives **outside** the `mode === 'agent' ? (...) : (...)` content ternary, so it renders in *both* modes. Nothing in JSX or CSS hides it in Agent mode.
- The click handler is a no-op in Agent mode:
  ```ts
  handlePrimaryClick = () => {
      if (this.state.mode !== 'agent') { void this.handleSubmit(); }   // line 746-749
  };
  ```
- `topic` state is **not** reset when switching to Agent mode (`enterAgentMode` / `handleSelectMode` leave it intact), and `canSubmit()` is `topic.trim().length > 0`.

**Failure scenario:** user types a topic in Normal mode → clicks the header **Agent** tab → the Start button is now rendered *and enabled* (`canSubmit()` true), but clicking it does nothing (`handlePrimaryClick` early-returns). If no topic was typed, the button instead sits permanently disabled inside the Agent panel — visual noise the #1071 gate was specifically added to remove.

I verified this by test-merging `pr1079` into `origin/main`: the merged tree keeps `main`'s mode switch (line ~1030) **and** the no-op `handlePrimaryClick` guard (line ~747) while the gate around the button (line ~1351 in the merged file) is gone.

**Fix:** restore the render gate on the button:
```tsx
{mode !== 'agent' && (
  <Button className="summary-workbench-start-btn" ... onClick={this.handlePrimaryClick}>...</Button>
)}
```
(The commit `fd5f622` "remove Agent mode switch (out of scope)" is what stripped this — keep the member-selector change, but do not delete the `mode !== 'agent'` guard, which is load-bearing on current `main`.)

---

## 3. Non-blocking findings (P2 / nits)

**P2 — `contactsList` first-load race (self-healing).** `loadMembers()` reads a one-time snapshot of `WKApp.dataSource.contactsList` and registers no `addContactsChangeListener`. `contactsSync()` is fired at login (`App.tsx:999`) — normally long before a user reaches the summary page — and `loadMembers()` re-runs on every modal open, so the empty window is narrow and self-heals on reopen. Not a blocker, but if you want to close it, subscribe to `notifyContactsChange` while the modal is open, or re-`loadMembers()` on that event. (Consensus with the second review leg.)

**P2 — PR description says contacts are "capped at 200"; the code has no cap.** The `slice(0, 200)` from an intermediate commit was dropped during later refactoring. `contactsList` is rendered in full (virtual scroll makes this fine performance-wise). Please update the description or re-add the cap so the stated behavior matches the code.

**Nit — dead fallback fields.** `m.user_id`, `m.username`, `m.id` fallbacks in the `else` branch never fire: `Contacts` only defines `uid` / `name` / `robot` / `status`. Harmless, but `m.uid` / `m.name` alone are sufficient.

**Nit — `getDisplayList()` recomputed on every scroll tick.** `handleListScroll` calls `this.getDisplayList().length` and `render()` calls it again; for large followed/recent lists the full group/thread assembly re-runs per scroll event. Consider memoizing the list length or caching the derived list. Non-blocking.

**Nit — no component tests.** The new members mode, virtual-scroll math (`ITEM_HEIGHT`, overscan ±5, `visibleStart/visibleEnd`), and single-tab hide have no unit/component coverage. A small test around `getDisplayList` filtering and the members/contacts branch would guard against future regressions.

---

## 4. Verdict

**CHANGES_REQUESTED** — B1 is a functional regression that a plain merge will silently introduce (gate present on current `main` via #1071, absent after this PR). Restoring the `mode !== 'agent'` guard around the Start button clears it; the P2/nits are optional follow-ups.

Once the guard is restored, re-request review and I'll re-verify against the new head SHA (a fresh verdict is required per head — a fast-forward does not inherit approval).

PR #1085, opened 2026-07-24
feat(channel-setting): refresh information panels and member views

Candice00711:
## Summary

- align group and subzone information panels with the reviewed card-based layout and wider drawer while preserving existing entry points, permissions, and runtime actions
- move simple name, notice, remark, group-nickname, and subzone-name edits into reusable inline rows; keep group/subzone action semantics unchanged
- refine member grid and member-list presentation while preserving the existing Chinese/pinyin search, authoritative server search, and pagination behavior
- keep inline drafts intact when saves fail or external channel refreshes arrive during editing, and make the group-nickname clear action reliable on first edit

## Scope boundaries

- no changes to `bridge/channelSetting/*Actions`
- no changes to WKApp, WKSDK, datasource, Chat, Conversation, or Messages runtime paths
- no new subzone switches, hidden entries, or permission behavior
- no new pinyin-search capability; this PR only preserves and adapts the behavior already present at its merge base

## Validation

- focused `vitest`: 4 files, 34 tests passed
- inline-edit matrix: clear/save, successful save, `false` failure, rejected failure, external refresh while editing, and cancel-to-latest-value all passed
- `pnpm i18n:check`: passed
- `pnpm --dir apps/web build`: passed
- Prettier and `git diff --check`: passed
- full dmworkbase suite: 2,064/2,065 passed; one unrelated `ThreadPanel.archive` test timed out once under full-suite load and passed 16/16 when rerun in isolation
- Storybook static build: transformed 10,137 modules without component compilation errors, then exited with the existing low-level code 132 issue

## Manual acceptance

- group and subzone information cards, drawer width, member spacing, and member-list back navigation
- inline edit/save/cancel for group name, announcement, remark, group nickname, and subzone name
- failed saves keep the editor open and preserve the typed draft
- background channel refreshes do not overwrite a draft; Cancel shows the latest external value
- group nickname clear action works on first panel entry
- existing view-more and transfer-owner member search still supports Chinese and pinyin, including members outside a super-group first-page cache
- unchanged mute, pin, save, archive, leave, exit, and permission behavior

### DIFF
--- a/packages/dmworkbase/src/App.css
+++ b/packages/dmworkbase/src/App.css
@@ -12,7 +12,7 @@
     /* --wk-height-chat-search: 65px; */
     --wk-height-chat-search: 64px;
     --wk-height-chat-conversation-header: 48px;  /* Figma: 48px */
-    --wk-wdith-chat-channelsetting: 380px;
+    --wk-wdith-chat-channelsetting: 28rem;
     --wk-width-chat-search-panel: 480px;
     --wk-width-thread-panel: 380px;
     --wk-layer-transition: 300ms cubic-bezier(0.33,1,0.68,1);
--- a/packages/dmworkbase/src/Components/ChannelSetting/index.css
+++ b/packages/dmworkbase/src/Components/ChannelSetting/index.css
@@ -83,7 +83,143 @@
     width: 100%;
     height: 100%;
     overflow: auto;
-    padding-bottom: 40px;
+    padding: var(--wk-sp-2) 0 var(--wk-sp-4);
+    background: var(--wk-bg-surface);
+}
+
+/* 仅作用于 ChannelSetting，避免影响其他 Sections/ListItem 调用方。 */
+.wk-channelsetting-content > .wk-sections {
+    display: flex;
+    flex-direction: column;
+    gap: var(--wk-sp-2);
+}
+
+.wk-channelsetting-content .wk-section {
+    margin: 0 var(--wk-sp-4);
+    overflow: hidden;
+    background: var(--wk-bg-base);
+    border: 1px solid var(--wk-border-subtle);
+    border-radius: var(--wk-r-sm);
+}
+
+.wk-channelsetting-content .wk-section-title {
+    padding: var(--wk-sp-3) var(--wk-sp-4) var(--wk-sp-2);
+    color: var(--wk-text-tertiary);
+    font-size: var(--wk-text-size-sm);
+}
+
+.wk-channelsetting-content .wk-section-subtitle {
+    padding: var(--wk-sp-2) var(--wk-sp-4) var(--wk-sp-3);
+    color: var(--wk-text-tertiary);
+    font-size: var(--wk-text-size-sm);
+}
+
+.wk-channelsetting-content .wk-section-row + .wk-section-row {
+    border-top: 1px solid var(--wk-border-subtle);
+}
+
+.wk-channelsetting-content .wk-list-item {
+    min-height: calc(var(--wk-sp-10) + var(--wk-sp-2));
+    padding: var(--wk-sp-2) var(--wk-sp-4);
+    background: transparent;
+}
+
+.wk-channelsetting-content .wk-list-item:hover {
+    background: var(--wk-bg-hover);
+}
+
+.wk-channelsetting-content .wk-list-item-static:hover {
+    background: transparent;
+}
+
+.wk-channelsetting-content .wk-list-item-title {
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-md);
+}
+
+.wk-channelsetting-content .wk-list-item-subtitle {
+    min-width: 0;
+    margin-left: var(--wk-sp-4);
+    color: var(--wk-text-secondary);
+    font-size: var(--wk-text-size-base);
+}
+
+.wk-channelsetting-content .wk-list-item-subtitle-muliteline {
+    margin-top: var(--wk-sp-2);
+    color: var(--wk-text-secondary);
+    font-size: var(--wk-text-size-base);
+}
+
+.wk-channelsetting-content .wk-channelsetting-qrcode-icon {
+    width: var(--wk-sp-4);
+    height: var(--wk-sp-4);
+    color: var(--wk-text-tertiary);
+}
+
+.wk-channelsetting-content .wk-section:has(.wk-subscribers) {
+    padding: var(--wk-sp-4) var(--wk-sp-2) var(--wk-sp-3);
+}
+
+.wk-channelsetting-content .wk-section:has(.wk-subscribers) .wk-section-row {
+    border: 0;
+}
+
+.wk-channelsetting-content .wk-subscribers-content {
+    grid-template-columns: repeat(5, minmax(0, 1fr));
+    column-gap: var(--wk-sp-2);
+    row-gap: var(--wk-sp-4);
+    padding-inline: var(--wk-sp-2);
+}
+
+.wk-channelsetting-content .wk-subscribers-item {
+    min-height: 0;
+    gap: var(--wk-sp-1-5);
+}
+
+.wk-channelsetting-content .wk-subscribers-item img,
+.wk-channelsetting-content .wk-subscribers-item-avatar-wrap img,
+.wk-channelsetting-content .wk-subscribers-item-avatar-wrap {
+    width: var(--wk-sp-12);
+    height: var(--wk-sp-12);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-name {
+    margin-top: 0;
+    color: var(--wk-text-secondary);
+    font-size: var(--wk-text-size-xs);
+    line-height: var(--wk-text-size-md);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-role-badge {
+    right: auto;
+    bottom: calc(0px - var(--wk-sp-1));
+    left: 50%;
+    padding: var(--wk-sp-0-5) var(--wk-sp-1);
+    color: var(--wk-text-inverse);
+    font-size: var(--wk-text-size-badge);
+    border-radius: var(--wk-r-xs);
+    box-shadow: none;
+    transform: translateX(-50%);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-role-badge-owner {
+    background: var(--wk-color-warning);
+}
+
+.wk-channelsetting-content .wk-subscribers-item-role-badge-manager {
+    background: var(--wk-text-primary);
+}
+
+.wk-channelsetting-content .wk-subscribers-more {
+    margin: var(--wk-sp-4) var(--wk-sp-2) 0;
+    padding: var(--wk-sp-2);
+    border-radius: var(--wk-r-sm);
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-sm);
+}
+
+.wk-channelsetting-content .wk-subscribers-more:hover {
+    background: var(--wk-bg-hover);
 }
 
 .wk-channelsetting-channel-info {
@@ -198,8 +334,12 @@
 
 .wk-route-content .wk-channelsetting-input-edit {
     box-sizing: border-box;
-    width: calc(100% - (2 * var(--wk-sp-2)));
-    margin: 0 var(--wk-sp-2);
+    width: calc(100% - (2 * var(--wk-sp-4)));
+    margin: var(--wk-sp-2) var(--wk-sp-4) 0;
+    overflow: hidden;
+    background: var(--wk-bg-base);
+    border: 1px solid var(--wk-border-subtle);
+    border-radius: var(--wk-r-sm);
 }
 
 .wk-channelmanage {
--- a/packages/dmworkbase/src/Components/Subscribers/index.tsx
+++ b/packages/dmworkbase/src/Components/Subscribers/index.tsx
@@ -54,12 +54,12 @@ export class Subscribers extends Component<SubscribersProps> {
         <div className="wk-subscribers-item-avatar-wrap">
           <img src={WKApp.shared.avatarUser(subscriber.uid)} alt=""></img>
           {subscriber.role === GroupRole.owner && (
-            <span className="wk-subscribers-item-role-badge">
+            <span className="wk-subscribers-item-role-badge wk-subscribers-item-role-badge-owner">
               {this.context.t("base.subscribers.role.owner")}
             </span>
           )}
           {subscriber.role === GroupRole.manager && (
-            <span className="wk-subscribers-item-role-badge">
+            <span className="wk-subscribers-item-role-badge wk-subscribers-item-role-badge-manager">
               {this.context.t("base.subscribers.role.manager")}
             </span>
           )}
@@ -157,12 +157,17 @@ export class Subscribers extends Component<SubscribersProps> {
                           )}
                         />,
                         new RouteContextConfig({
-                          title: this.context.t("base.subscribers.memberList"),
+                          title: this.context.t(
+                            "base.subscribers.memberListWithCount",
+                            { values: { count: vm.memberCount() } }
+                          ),
                         })
                       );
                     }}
                   >
-                    {this.context.t("base.subscribers.viewMore")}
+                    {this.context.t("base.subscribers.viewAll", {
+                      values: { count: vm.memberCount() },
+                    })}
                   </div>
                 ) : undefined}
               </div>
--- a/packages/dmworkbase/src/Components/Subscribers/list.css
+++ b/packages/dmworkbase/src/Components/Subscribers/list.css
@@ -4,19 +4,90 @@
     background-color: var(--wk-bg-surface);
     width: 100%;
     height: 100%;
-    overflow: scroll;
+    overflow-y: auto;
+    overflow-x: hidden;
+}
+
+/*
+ * 成员列表返回时保留 animationend，让 WKViewQueue 继续按原逻辑清理栈顶，
+ * 但取消可见位移，避免标题已切回父页而内容仍横向退场造成闪动。
+ * 不改变 RoutePage/WKViewQueue 的入栈、退栈与数据刷新逻辑。
+ */
+.wk-viewqueue-view-out:has(.wk-subscrierlist) {
+    background: var(--wk-bg-surface);
+    animation-name: wk-subscriber-list-exit;
+    animation-duration: var(--wk-dur-fast);
+    animation-timing-function: var(--wk-ease);
+}
+
+@keyframes wk-subscriber-list-exit {
+    from,
+    to {
+        transform: translate3d(0, 0, 0);
+    }
+}
+
+.wk-subscrierlist .wk-indextable-search-box {
+    height: auto;
+    margin: var(--wk-sp-2) var(--wk-sp-5);
+    padding: var(--wk-sp-1-5) var(--wk-sp-2);
+    align-items: center;
+    background: var(--wk-bg-base);
+    border: 1px solid var(--wk-border-subtle);
+    border-radius: var(--wk-r-sm);
+}
+
+.wk-subscrierlist .wk-indextable-search-icon {
+    margin-left: 0;
+    color: var(--wk-text-tertiary);
+}
+
+.wk-subscrierlist-search-icon {
+    width: var(--wk-sp-4);
+    height: var(--wk-sp-4);
+}
+
+.wk-subscrierlist .wk-indextable-search-input {
+    flex: 1;
+    max-width: none;
+    margin-left: var(--wk-sp-2);
+}
+
+.wk-subscrierlist .wk-indextable-search-input input {
+    width: 100%;
+    height: auto;
+    padding: 0;
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-md);
+    line-height: var(--wk-sp-5);
+}
+
+.wk-subscrierlist .wk-indextable-search-input input::placeholder {
+    color: var(--wk-text-tertiary);
 }
 
 .wk-subscrierlist-list-item {
     display: flex;
     align-items: center;
-    padding: 1rem;
+    gap: var(--wk-sp-3);
+    min-height: var(--wk-sp-12);
+    padding: var(--wk-sp-2) var(--wk-sp-5);
     cursor: pointer;
+    transition: background var(--wk-dur-fast) var(--wk-ease);
+}
+
+.wk-subscrierlist-list-item:hover {
+    background: var(--wk-bg-hover);
 }
 
 .wk-subscrierlist-item-avatar {
     position: relative;
     flex-shrink: 0;
+}
+
+.wk-subscrierlist-item-avatar .wk-avatar {
+    width: var(--wk-sp-8);
+    height: var(--wk-sp-8);
 }
 
 .wk-subscrierlist-item-online-badge {
@@ -26,13 +97,17 @@
 }
 
 .wk-subscrierlist-item-name {
-    margin-left: 1rem;
+    min-width: 0;
+    margin-left: 0;
     text-overflow: ellipsis;
     overflow: hidden;
     white-space: nowrap;
-    max-width: 10rem;
+    max-width: 16rem;
     display: flex;
     align-items: center;
+    gap: var(--wk-sp-1);
+    color: var(--wk-text-primary);
+    font-size: var(--wk-text-size-md);
 }
 
 .wk-subscrierlist-item-source {
@@ -57,13 +132,36 @@
 
 .wk-subscrierlist-item-content {
     display: flex;
+    min-width: 0;
     width: 100%;
-    justify-content: space-between;
+    align-items: center;
+    justify-content: flex-start;
+    gap: var(--wk-sp-1-5);
 }
 
 .wk-subscrierlist-item-desc {
-    font-size: 14px;
+    display: none;
+    flex-shrink: 0;
+    padding: var(--wk-sp-0-5) var(--wk-sp-1);
+    font-size: var(--wk-text-size-tiny);
+    font-weight: 600;
+    line-height: var(--wk-text-size-base);
+    border-radius: var(--wk-r-xs);
+}
+
+.wk-subscrierlist-item-desc-owner,
+.wk-subscrierlist-item-desc-manager {
+    display: inline-flex;
+}
+
+.wk-subscrierlist-item-desc-owner {
+    color: var(--wk-color-warning);
+    background: var(--wk-color-warning-bg);
+}
+
+.wk-subscrierlist-item-desc-manager {
     color: var(--wk-text-secondary);
+    background: var(--wk-bg-elevated);
 }
 
 .wk-subscrierlist-item-action {
--- a/packages/dmworkbase/src/Components/Subscribers/list.tsx
+++ b/packages/dmworkbase/src/Components/Subscribers/list.tsx
@@ -298,9 +298,7 @@ export class SubscriberList extends Component<
               >
                 <div className="wk-indextable-search-box">
                   <div className="wk-indextable-search-icon">
-                    <IconSearchStroked
-                      style={{ color: "#bbbfc4", fontSize: "20px" }}
-                    />
+                    <IconSearchStroked className="wk-subscrierlist-search-icon" />
                   </div>
                   <div className="wk-indextable-search-input">
                     <input
@@ -312,7 +310,6 @@ export class SubscriberList extends Component<
                       )}
                       ref={(rf) => {}}
                       type="text"
-                      style={{ fontSize: "17px" }}
                     />
                   </div>
                 </div>
@@ -393,7 +390,15 @@ export class SubscriberList extends Component<
                               </Tag>
                             )}
                           </div>
-                          <div className="wk-subscrierlist-item-desc">
+                          <div
+                            className={`wk-subscrierlist-item-desc${
+                              item.role === GroupRole.owner
+                                ? " wk-subscrierlist-item-desc-owner"
+                                : item.role === GroupRole.manager
+                                ? " wk-subscrierlist-item-desc-manager"
+                                : ""
+                            }`}
+                          >
                             {this.getRoleName(item)}
                           </div>
                         </div>
--- a/packages/dmworkbase/src/Components/Subscribers/vm.ts
+++ b/packages/dmworkbase/src/Components/Subscribers/vm.ts
@@ -71,4 +71,8 @@ export class SubscribersVM extends ProviderListener {
         let showMemberNum = this.shouldShowMemberNum()
         return this.subscribers.length>showMemberNum
     }
+
+    memberCount() {
+        return this.routeData.channelInfo?.orgData?.member_count || this.subscribers.length
+    }
 }
--- a/packages/dmworkbase/src/features/channelSetting/__tests__/channelSettingSections.test.ts
+++ b/packages/dmworkbase/src/features/channelSetting/__tests__/channelSettingSections.test.ts
@@ -5,6 +5,7 @@ import { ChannelTypeCommunityTopic } from "../../../Service/Const";
 import { ThreadStatus } from "../../../Service/Thread";
 import { GroupStatusDisband } from "../../../Utils/groupDisband";
 import { updateChannelSettingMyGroupNickname } from "../../../bridge/channelSetting/channelSettingActions";
+import { t } from "../../../i18n";
 import { buildChannelGroupInfoSection } from "../channelSettingGroupInfoSection";
 import {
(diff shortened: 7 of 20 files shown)


lml2468 (approved):
## Review: #1085 @ `71a32ef2` — feat(channel-setting): refresh information panels and member views

> #1085 是**已关闭的 #1075 的重开版** —— 同分支 `feat/channel-setting-v2-style`、**同 head `71a32ef2`**(#1075 现 CLOSED)。我已在 #1075 对这个**完全相同的 SHA** 做过逐字节复核(rev-parse 确认 HEAD、读实际源码、跑构建+测试),结论直接适用。

**结论:APPROVE ✅**

### 两条草稿丢失阻塞均已修 + 有回归测试(字节级复核 @ `71a32ef2`)
- **🔴→✅ 编辑中远端 value 变化覆盖草稿**:`ui/ChannelSettingRows/index.tsx` 用 `editingRef`(`useEffect([editing])` 同步)守卫 value-sync:`useEffect(() => { setCurrentValue(value); if (!editingRef.current) setDraft(value); }, [value])` —— 编辑中远端刷新不覆盖 draft,`currentValue` 仍跟最新值供 Cancel 回落。专测「preserves an in-progress draft across external value updates」在。
- **🔴→✅ 保存失败保留草稿**:`if (saved !== false)` 才关闭 + catch 保持编辑态;测「keeps the draft open when saving fails/rejects」。

### 承前无回归
成员搜索为客户端 `localSearch`(不越权 dump)、权限门(isManagerOrCreatorOfMe/canEdit)保留、无 XSS、`--wk-*` token 定义暗色安全、i18n 无孤儿。

**构建 / 测试(实跑 @ `71a32ef2`)**:`pnpm --filter @octo/web build` ✅;`ChannelSettingInlineEditRow` **9 tests 全绿**(草稿存活/保存失败/保存拒绝/成功可见性等)。`mergeable=true`。

---

### 🟡 非阻断(承 #1075)
- `onSave` await 后 setState 缺卸载守卫;
- 搜索失败时全屏错误无搜索框,建议保留搜索框便于改词(yujiawei);
- CI 未跑 dmworkbase vitest 套件,这些回归测试不在 gate 上执行(cnwenf)—— 建议后续补进流水线。

---

与我在 #1075 已 APPROVE 的 `71a32ef2` 内容一致:两条草稿丢失均修复且有回归测试、授权/XSS/token/i18n 无回归、构建与 9 测试通过。**可合并 —— APPROVE**(#1075 已 CLOSED,以本 #1085 为准,避免重复合入)。

Jerry-Xin (approved):
**APPROVE** — channel-setting v2 refresh + member-view panels; carry-forward safe, no new blockers.

Relationship: this PR shares an **identical head (`71a32ef`) and identical base (merge-base `e28c75b`)** with the now-CLOSED #1075, so the 20-file / +1002 −176 two-dot delta is the same tree I already verified there. #1085 is effectively #1075 re-opened under a new number ("member views" vs "member search" in the title); the #1075 verification carries over and is re-confirmed below at the live head.

🔴 Blocking: none.

💬 Non-blocking notes:
- CI (`ci.yml`) runs the `@octo/docs` vitest suite but does **not** execute the `dmworkbase` vitest suite, so the new `ChannelSettingInlineEditRow` regression tests (10 cases) are not exercised by the CI gate. Coverage blind spot only — tests themselves are correct and pass locally. Consider wiring dmworkbase tests into CI in a follow-up.
- `semantic.css` defines the `--wk-*` tokens (276 tokens) but there is no `prefers-color-scheme`/`[data-theme]` dark selector in `packages/dmworkbase/src/theme/`; dark theming is handled at runtime rather than via CSS media query. Pre-existing architecture, not introduced here.

✅ Highlights:
- Draft-loss fixes intact & tested: `ChannelSettingRows/index.tsx` syncs `currentValue` on external `value` change but preserves `draft` while editing (`editingRef` guard), and the save handler gates `setEditing(false)` on `saved !== false` so a failed/rejected save keeps the editor open with the draft. Regression tests cover "keeps the draft open when saving fails/rejects" and "preserves an in-progress draft across external value updates".
- Nickname save-failure feedback resolved: `channelSettingNicknameSection.tsx` wraps the update in try/catch, surfaces `Toast.error(error.msg || saveFailedRetry)` on failure, and returns `false` to keep the editor open — user now gets a visible error instead of a silent swallow.
- Member-view additions are cosmetic: role-badge/owner/manager CSS classes, search-icon inline-style → class, and a `viewAll {{count}}` / `Members ({{count}})` label. `SubscribersVM.memberCount()` reads the already-loaded `channelInfo.orgData.member_count` (falls back to `subscribers.length`) — no new member fetch, no unscoped member dump.
- No XSS: no `dangerouslySetInnerHTML`/`innerHTML`/`eval` in the diff; the only interpolated value is a numeric `{{count}}`; member names/remarks render through the existing escaped paths.
- Permission guards intact: inline-edit rows gate `onStartEdit` on `data.isManagerOrCreatorOfMe`, avatar upload gated via `showUpload`.
- i18n parity: all new keys (`common.clear`, `module.channelSettings.saveFailedRetry`, `subscribers.viewAll`, `subscribers.memberListWithCount`) present in both en-US and zh-CN; referenced `common.notSet` exists in both.
- All `--wk-*` tokens used across the new CSS resolve to definitions in `theme/semantic.css`.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1085 (octo-web)
**Head SHA:** `71a32ef2a1fc76fde477ba89dbac19f5f1a9618b`
**Scope:** channel-setting inline-edit refactor, thread overview consolidation, subscriber list polish, summary mode-switch UI.

## Summary

This PR migrates several channel-setting panels (group name, group notice, remark, thread name, group nickname) from the push-based `inputEditPush` navigation pattern to an inline `ChannelSettingInlineEditRow` component that edits in-place. The new component is well-designed — it preserves drafts across external value updates, handles save failures gracefully (returning `false` or rejecting keeps the editor open), and includes a character counter with over-limit disabling. Thread settings are consolidated from three separately-registered sections into one `buildThreadOverviewSection`. Subscriber list gets CSS polish (role-specific badge styling, member count in header) and the summary workbench gains a mode-switch pill toggle. One non-blocking consistency issue noted below.

## Verification

Static analysis only at head `71a32ef2`; build and tests not executed in this environment.

- ✅ **Inline edit contract** — `ChannelSettingInlineEditRow` correctly implements the `onSave` contract: `saved !== false` closes the editor; `false` or rejection keeps it open. The `useEffect` for `value` prop uses `editingRef` to preserve in-progress drafts during external updates.
- ✅ **Draft preservation** — Test at `ChannelSettingInlineEditRow.test.tsx` covers: cleared-value save, failed-save keeps draft, rejected-save keeps draft, draft preserved across external rerenders, successful save displays optimistically, start-guard rejection, and guard-void-enters-edit.
- ✅ **Nickname clear flow** — `buildMyGroupNicknameSection` correctly uses `?? ""` for the value and shows `"Not set"` as `displayValue` when empty. The old fallback to `subscriberOfMe.name` is removed, so clearing a nickname now shows "Not set" instead of reverting to the member name.
- ✅ **Thread overview consolidation** — `buildThreadOverviewSection` flatMaps rows from info, markdown, and webhook sub-sections. `module.tsx` removes the three separate registrations and registers the combined section at the same priority (500). No functionality lost.
- ✅ **i18n parity** — All new keys (`common.clear`, `saveFailedRetry`, `viewAll`, `memberListWithCount`) added to both `en-US.json` and `zh-CN.json`.
- ✅ **Summary mode switch** — New buttons reference the existing `mode` state (destructured at `SummaryCreatePage.tsx:1011`) and `handleSelectMode` handler (`:753`). CSS provides a standard segmented-control layout. i18n keys (`summary.create.start`, `summary.create.agentStart`) already exist via `ChatSummaryNewModal`.
- ✅ **Subscriber list changes** — Role-specific CSS classes on badges (`-owner`, `-manager`), member count display via new `SubscribersVM.memberCount()`, and CSS class extraction for search icon are all cosmetic/structural with no behavioral risk.

## Findings

### P2 — Remark `onSave` rejects instead of returning `false` (`packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:218`)

**Diff-scope:** new — introduced by this PR's migration from `inputEditPush` callback to `ChannelSettingInlineEditRow.onSave`.

The remark row's `onSave` does `.catch((error) => { Toast.error(error?.msg); throw error; })`, which rejects the promise. The `ChannelSettingInlineEditRow` catch block silently swallows the rejection, so the user-visible behavior is correct — toast fires, editor stays open with draft intact. However, every other migrated handler in this PR (group name at `channelSettingGroupProfileRows.tsx:78`, group notice at `:131`, thread name at `channelSettingThreadInfoSection.tsx:73`, nickname at `channelSettingNicknameSection.tsx:43`) uses the `.catch(…) { return false; }` pattern. The remark row is the sole outlier, and the `error?.msg` toast can display "undefined" when the error is a non-object. Aligning with the rest of the PR is a one-line change:

```ts
.catch((error) => {
  Toast.error(error?.msg ?? t("base.module.channelSettings.saveFailedRetry"));
  return false;
})
```

## Data Flow Tracing

| Consumer | Upstream Source | Runtime Verified |
|---|---|---|
| `ChannelSettingInlineEditRow.value` | `data.subscriberOfMe?.remark ?? ""` (nickname), `channelInfo?.title \|\| ""` (group name), `channelInfo?.orgData?.notice` (notice), `channelInfo?.orgData?.remark \|\| ""` (remark), `channelInfo?.title \|\| ""` (thread name) | ✅ All sources read at build time from `context.routeData()`, passed as `value` prop. |
| `ChannelSettingInlineEditRow.onSave(draft)` | User input → `draft` state → `onSave` callback → bridge action (`updateChannelSettingMyGroupNickname`, `updateChannelSettingField`, `remarkChannelSetting`, `updateChannelSettingThreadName`) | ✅ All bridge actions called with correct params; success path updates local state + `data.refresh()`; failure path shows toast + keeps editor open. |
| `displayValue` (nickname) | `groupNickname \|\| t("base.common.notSet")` | ✅ Shows "Not set" when remark is empty string, fixing the old fallback-to-name behavior. |
| `buildThreadOverviewSection` rows | `flatMap` of `buildThreadInfoSection` + `buildThreadMdSection` + `buildThreadWebhookSection` | ✅ All sub-sections called with correct context; test asserts 5 rows for active thread. |
| `SubscribersVM.memberCount()` | `routeData.channelInfo?.orgData?.member_count \|\| subscribers.length` | ✅ Falls back to local array length when server count unavailable. |

- **C1 (Dual-path parity):** N/A — No paired create/delete or subscribe/unsubscribe paths changed. The inline-edit migration is symmetric across all handlers.
- **C2 (Control-flow ordering):** Clear — The `onStartEdit` guard runs before the editor opens; `onSave` runs on button click. No nested reuse of these callbacks across different ordering contexts.
- **C3 (Authorization boundary):** Clear — Permission checks (`isManagerOrCreatorOfMe`, `canEdit`, `threadInfo` existence) are preserved in `onStartEdit` guards, same as the old `onClick` guards. No new endpoints or capabilities exposed.
- **C4 (Auth lifecycle / container-member cascade):** N/A — No auth or container-status changes in this PR.
- **C5 (Build/note ≠ runtime):** Clear — No build-time constructs relied upon as runtime guarantees. CSS class names match between JSX and CSS files.
- **C6 (Governance/docs self-consistency):** N/A — No governance, policy, or security documentation changes.

N/A — First review pass on this PR.

- The `onStartEdit` guard for thread name correctly returns `false` to block editing when `!threadInfo` or `!canEdit`. The `ChannelSettingInlineEditRow` treats `=== false` as blocking and any other value (including `undefined`) as allowing edit — consistent with the test at line 244.
- The `buildThreadOverviewSection` test asserts 5 rows (thread name + status + parent group + md + webhook), matching the sum of the three sub-sections.
- The group-name and group-notice `onSave` handlers correctly `.catch()` and `return false`, matching the component contract.
- The `useEffect` dependency array `[value]` is correct — the effect should fire only when the external value changes, not when `editing` changes.
- No AUTH/VISIBILITY concerns in the Subscribers CSS changes — role badge class additions are purely visual, not gating.
- No BYTE-CAP concerns — `maxCount` is a character count (`.length`), not a byte budget; no truncation or body-wrapping occurs.

## Verdict: APPROVED

Well-structured refactor that improves UX consistency across channel settings. The new inline-edit component handles failure paths correctly and has thorough test coverage. The single P2 (remark `throw` vs `return false`) is non-blocking and a quick fix.

[Octo-Q] verdict: APPROVE — No P0/P1 findings. One P2 (remark row error handling inconsistency) is non-blocking.

yujiawei (approved):
# Code Review — PR #1085 (octo-web)

Reviewed against the PR-specific base (5 commits: `901266e5..71a32ef`, 20 files, +1002 / −176). The branch also carries several already-merged PRs behind a stale local `main`; those were excluded from scope.

**Spec: ✅**

The diff implements exactly what the PR describes and nothing beyond it:

- **Inline-edit refactor.** A new reusable `ChannelSettingInlineEditRow` (`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx`) replaces the push-based modal `inputEditPush` flow for group name, group notice, remark, my-group-nickname, and thread name. Each caller keeps its original permission gate (moved into `onStartEdit`) and its original action semantics (moved into `onSave`).
- **Draft resilience.** Failed saves keep the editor open with the typed draft intact (`onSave` resolving `false` or rejecting — `index.tsx:221`, `225`), and external channel refreshes during editing do not clobber the draft (`index.tsx:126-131`, guarded by `editingRef`). Both paths are covered by tests.
- **Nickname clear reliability.** `channelSettingNicknameSection.tsx` now binds `value` to `data.subscriberOfMe?.remark ?? ""` instead of `remark || name`, so a cleared nickname stays cleared instead of falling back to the member name; `displayValue` shows `common.notSet`. Verified by the new "keeps a cleared group nickname empty" test.
- **Member view polish.** `Subscribers/index.tsx` + `list.tsx` add role-specific CSS classes and a `member_count`-backed "View all N members" label (`vm.ts:memberCount()`). Search/pagination/pinyin logic is untouched — the only `list.tsx` change is a CSS class on the search icon and role badges, so the "preserves search behavior" claim holds.
- **Thread section merge.** Three separately-registered thread sections (`thread.base.info`, `thread.md.setting`, `thread.webhook`) are merged into one `buildThreadOverviewSection` registered as `thread.overview` (`module.tsx`), with a code comment explicitly stating that currently-unregistered entries (mute, in-chat search) must NOT be restored. No new subzone switches, hidden entries, or permission changes were introduced.

No missing scope, no over-building, no divergence from the stated approach.

## 2. Code quality

**Quality: Approved**

Clean refactor with strong test coverage: the new component has 10 focused tests exercising clear/save, save-failure (false + reject), external-refresh-during-edit, optimistic-save display, start-guard, and save-button enablement; section builders and i18n keys all resolve. No P0/P1 issues found.

### P2 (advisory, non-blocking)

- **`notice` row omits the empty-string guard.** `channelSettingGroupProfileRows.tsx:122` passes `value: channelInfo?.orgData?.notice`, while every sibling inline-edit row defaults its value (`title || ""`, `remark || ""`, `remark ?? ""`). The component's default param `value = ""` only covers `undefined`, not an explicit `null`. `exceeded = maxCount !== undefined && draft.length > maxCount` (`index.tsx:133`) runs on every render before the editing guard, so if the backend ever sends `notice: null`, `draft.length` throws and the panel crashes. Optional chaining makes a *missing* field safe; only an explicit `null` triggers it. Add `|| ""` for consistency with the other four rows.

- **Save can overwrite a newer external value (last-write-wins).** In `index.tsx:220-223`, if `value` changes externally while `onSave` is pending, the `[value]` effect sets `currentValue` to the new external value, then the resolved save immediately overwrites it with the submitted `draft`. Because `value` does not change again, the effect will not re-run and the display can stay on the saved draft until the next external refresh. Narrow window; self-heals on the next refresh.

- **External update redefines the edit baseline for an untouched editor.** In `index.tsx:126-136`, while editing, the effect updates `currentValue` but intentionally leaves `draft` alone. For an editor that was opened but not typed into, an external value change flips `unchanged` from true to false, enabling Save and letting the now-stale draft overwrite the newer external value. This is inherent to the deliberate "preserve the user's draft across external refreshes" design (which the tests encode), so it is a design trade-off rather than a defect. If precise conflict handling is desired later, track a pristine/dirty flag and re-baseline an untouched draft.

### Nit

- `inputEditPush` is now unused inside `buildGroupProfileRows` and `buildGroupManagementRows` (dropped from the destructure) but still required by their `Build*Options` interfaces and threaded through by `channelSettingGroupInfoSection.tsx`. Harmless; can be dropped from the option types in a follow-up.

## 3. Verdict

**APPROVED.** Spec ✅ and Quality Approved. The three P2s and the nit are polish suggestions, not merge blockers — the crash path is guarded by optional chaining for all realistic inputs, and the concurrency edges are narrow and self-healing.

## 4. Coverage note

- Verified: full PR-specific diff, the new component's state machine and display expression, all five inline-edit call sites, i18n key resolution (en/zh), CSS-class existence for every new class name, `lucide-react` dependency presence, and that Subscribers search/pagination logic is untouched.
- Not exercised: build/lint/test were not re-run in this review environment (no `node_modules`); the author reports focused vitest (34 tests), `i18n:check`, and `pnpm build` all green. A second independent reviewer's git subprocess was sandbox-denied and did not complete, so this verdict rests on the primary line-level review plus one component-scoped adversarial pass.

cnwenf (commented):
# Code Review — PR #1085 (Mininglamp-OSS/octo-web)

## Summary

This PR refreshes the group/subzone channel-setting panels into the reviewed card-based layout, migrates the simple name/notice/remark/group-nickname/subzone-name edits from the pushed full-screen `InputEdit` route into a reusable inline `ChannelSettingInlineEditRow`, collapses three thread section registrations into one `thread.overview`, and refines the member grid/list presentation. All three independent reviewers agree the migration is careful and behavior-preserving: permission guards, action semantics, and bridge actions are intact; the inline-edit draft-preservation and save-`false`/reject contracts are genuinely pinned by mutation-tested tests; i18n keys are in parity across locales; the build serves cleanly at the exact head; and security is cleared with no new XSS/SSRF/credential surface. After deduplication the consolidated findings are **1 disputed P1 (independently falsified as unreachable by two of three reviewers), 5 P2, 5 P3, and 2 nits** — none a confirmed blocker. The highest-value follow-ups are the `memberCount()` `??` fix, a remark save-failure test (guards a silent data-loss regression), and wiring the dmworkbase vitest suite into CI.

## Findings

13 findings after dedup (1 disputed P1, 5 P2, 5 P3, 2 nits). Multi-reviewer findings (higher confidence) are listed first.

### P2 — `memberCount()` uses `||` instead of `??` (`packages/dmworkbase/src/Components/Subscribers/vm.ts:76`) [source: octo-review-qwen3.7 + octo-review-glm5.2 + octo-review-qwen3.8]

`memberCount()` uses `||` for the fallback: `this.routeData.channelInfo?.orgData?.member_count || this.subscribers.length`. If the server returns `member_count: 0`, `||` treats `0` as falsy and falls through to `this.subscribers.length`. This count is interpolated into i18n strings shown to the user in the "View all N members" button and member list page title. Fix: replace `||` with `??`.

Validated: created SubscribersVM with `member_count: 0` and 3 subscribers. `vm.memberCount()` returned 3 instead of 0.

Corroboration: glm5.2 confirms a dependency-free probe of the verbatim expression — when `member_count` is absent (channelInfo not yet loaded / field omitted) the count falls back to the loaded page length, so a 500-member super-group can render `Members (20)` / `View all 20 members` until the server field lands; blast radius is bounded (server normally returns `member_count`; count only renders behind `hasMoreSubscribers()`). qwen3.8 confirms the authoritative `member_count` wins when present and the loaded count is a deliberate lower-bound otherwise, but notes none of the 2,072 existing tests pins it and `??` would also avoid coercing a present `member_count === 0`. Recommended fix (glm5.2):

```ts
return this.routeData.channelInfo?.orgData?.member_count ?? this.subscribers.length
```

(severity split: qwen3.7=P2, glm5.2=Nit, qwen3.8=Nit)

### P2 — A successful in-flight save can momentarily clobber an external refresh (`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:222`) [source: octo-review-qwen3.8 + octo-review-glm5.2]

`onSave` does `setSaving(true); const saved = await onSave(draft); if (saved !== false) { setCurrentValue(draft); … }`. The `editingRef` guard correctly stops an external `value` prop from overwriting the draft *while editing*, but an authoritative value that lands *during the in-flight save* is dropped from the display when `setCurrentValue(draft)` runs after the await (confirmed by a probe rendering the real component: the display shows the local draft, not the mid-save remote value). I am keeping this at P2 but flagging that its own evidence bounds it toward a nit: it requires a true concurrent edit inside the narrow save window, it is conventional optimistic-UI (a successful local save wins locally), and the display self-heals on the next authoritative prop change (`data.refresh()`/server echo), so the value is deferred, not lost. Note the group-profile name/notice handlers have no explicit success-path `data.refresh()`, so they rely on the bridge resolving. A small fix is to skip `setCurrentValue(draft)` when the prop changed mid-save (compare against a ref of the latest `value`).

Corroboration: glm5.2 independently reproduced the mechanism with an e2e probe (display shows the draft, not `Remote authoritative`, immediately after the save resolves; a later `Server converged` prop reconciles it) and adds the more user-visible wrinkle that the group-name and group-notice handlers (`channelSettingGroupProfileRows.tsx:66`, `:136`) don't call `data.refresh()` on success — unlike thread-name and nickname — so the `displayValue` external-group `<Tag>` (`:54`) disappears after a rename until an unrelated refresh updates `channelInfo.title`.

(severity split: qwen3.8=P2, glm5.2=Nit)

### P2 — CI does not run dmworkbase vitest suite (`.github/workflows/ci.yml`) [source: octo-review-qwen3.7 + octo-review-glm5.2]

`.github/workflows/ci.yml` build job runs `pnpm i18n:check`, docs-only vitest, `pnpm build`, `pnpm lint`, and `pnpm lint:wkmodal` — there is no dmworkbase vitest step. The new `ChannelSettingInlineEditRow.test.tsx` (278 lines) and modified `channelSettingSections.test.ts` are invisible to the pipeline, so a regression breaking them leaves CI green.

Corroboration: glm5.2 (`.github/workflows/ci.yml:91-104`) confirms there is no `pnpm test`/vitest step for `@octo/base` (dmworkbase) and notes this is a pre-existing CI gap (not introduced by this PR); the PR's stated validation ("focused vitest: 4 files, 34 tests passed") rests on suites CI never runs. The mutation probes show the tests are genuinely effective when run, so this is awareness-only, not a block; worth wiring `@octo/base` vitest into the build job separately.

(severity split: qwen3.7=P2, glm5.2=Nit/observation)

### P3 — Builder-level onSave/onStartEdit contracts (incl. new inline permission guards): no test coverage (`packages/dmworkbase/src/features/channelSetting/channelSettingGroupProfileRows.tsx:141`) [source: octo-review-qwen3.7 + octo-review-qwen3.8]

Group-name, group-notice, remark onSave failure paths and thread-name onStartEdit permission guard are all untested at the builder level. Mutation testing confirmed all four gaps leave the existing suite green.

Corroboration: qwen3.8 (`channelSettingGroupProfileRows.tsx:57`) — the group-name/notice `onStartEdit` guards (`if (!data.isManagerOrCreatorOfMe) { Toast.warning(…); return false; }`) and the thread-name `!canRenameThread` guard are not invoked by any test (the section tests assert row counts; `canRenameThread` is mocked `true`, making the blocking branch unreachable). A mutation removing the group-name check lets a non-manager open the editor with no failing test. Impact is UI/permission-display only — the server still enforces on save — hence a nit. A test calling `properties.onStartEdit()` as a non-manager would pin it.

(severity split: qwen3.7=P3, qwen3.8=Nit)

### Nit — Stale doc comment names a removed registration (`packages/dmworkbase/src/Service/threadPermission.ts:71`) [source: octo-review-glm5.2 + octo-review-qwen3.8]

The comment still references `module.tsx` 的 `thread.base.info`, which this PR removes (collapsed into `thread.overview` at priority 500). Confirmed by `git show HEAD:module.tsx | grep -c thread.base.info` = 0. Doc-only — the permission fn is still reached via `buildThreadOverviewSection → buildThreadInfoSection` — but the comment now contradicts the code. Update it to reference `thread.overview` (or drop the specific key).

Corroboration: qwen3.8 — the diff renames the thread rename-entry registration from `thread.base.info` to `thread.overview` (`packages/dmworkbase/src/module.tsx:1667`, also deleting `thread.md.setting`/`thread.webhook`); `rg 'thread.base.info'` now matches only that comment. Doc-only, no runtime impact.

### P1 (DISPUTED) — Thread name onSave returns void instead of false (`channelSettingThreadInfoSection.tsx:70`) [source: octo-review-qwen3.7 — independently falsified as unreachable by octo-review-glm5.2 + octo-review-qwen3.8]

Thread name `onSave` returns `undefined` (not `false`) when `threadInfo` is null: `if (!threadInfo) return;`. The `ChannelSettingInlineEditRow` component checks `if (saved !== false)` — `undefined !== false` is `true`, so it calls `setCurrentValue(draft)` and closes the editor, displaying the unsaved draft as the persisted value. No API call was made, but the UI shows a false-positive save until the next external refresh. Although `onStartEdit` also guards against null `threadInfo`, a thread deletion between start-edit and save would bypass that guard. The contract doc-string says "Resolve false or reject to keep the editor open" — this branch violates that contract.

Fix: `if (!threadInfo) return false;`

Validated: built `buildThreadInfoSection` with a channel ID lacking the separator (parseThreadChannelId returns null). Called `onSave('new name')` → result was `undefined` (not `false`).

> **Dispute note (2 of 3 reviewers):** glm5.2 and qwen3.8 each independently **falsified this as a reachable bug via probe** — `threadInfo` is a build-time closure constant, so the `if (!threadInfo) return` branch cannot coincide with `editing === true` (the editor only opens when `threadInfo` is non-null), making the early-return unreachable in practice. qwen3.7's residual concern is a thread-deletion race between start-edit and save. Consolidated view: contract-cleanliness fix (`return false`) is cheap and reasonable, but this is not a confirmed live bug — treat as disputed/low-risk, not a blocker.

(severity split: qwen3.7=P1, glm5.2=not-a-bug, qwen3.8=not-a-bug)

### P2 — Group name/notice `error.msg` without type narrowing (`channelSettingGroupProfileRows.tsx:68`) [source: octo-review-qwen3.7]

Group name and notice `onSave` catch handlers access `error.msg` without null-safe chaining: `.catch((error) => { Toast.error(error.msg); return false; })`. If `updateChannelSettingField` rejects with a non-object (string, null, undefined), `error.msg` throws a TypeError that `ChannelSettingInlineEditRow`'s `catch {}` silently swallows — the user sees no error feedback. In contrast, `channelSettingNicknameSection.tsx` in this same PR uses proper type narrowing (`typeof error === 'object' && error !== null && 'msg' in error`). This is an intra-PR inconsistency.

### P2 — Remark save-failure path is untested; a regression silently shows a failed edit as saved (`packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:220`) [source: octo-review-qwen3.8]

The remark inline edit is the one migrated row whose `onSave` uses `…catch((error) => { Toast.error(error?.msg); throw error; })` — it rethrows so the component's `await` throws and keeps the editor open. No test imports `buildGroupManagementRows` at all (it is touched only for row counts). A mutation deleting that `throw error` makes `onSave` resolve `undefined`, so `ChannelSettingInlineEditRow`'s `if (saved !== false)` (`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:221`) closes the editor and promotes the unsaved draft to the displayed value — a silent data-loss UX — with **zero** failing tests. The behavior is correct today; it is just unguarded. Worth a builder-level test asserting the remark `onSave` rejects (and the editor stays open) when `remarkChannelSetting` fails; also consider aligning remark with the other rows' `return false` contract so the component handles failure uniformly.

### P3 — Inconsistent error handling across migrated rows (`channelSettingGroupManagementRows.tsx:215`) [source: octo-review-qwen3.7]

Remark row re-throws after Toast while group name/notice return false. Functionally equivalent (editor stays open in both cases) but the three migrated rows use three different error-handling strategies. Consider unifying.

### P3 — Thread-name onSave failure path: no builder-level test (`channelSettingThreadInfoSection.tsx:79`) [source: octo-review-qwen3.7]

Mutation protocol confirmed: removing `return false` from the thread-name catch handler leaves all existing tests green. The component-level test uses a stub, not the real builder wiring.

### P3 — Dead `inputEditPush` parameter (`channelSettingGroupProfileRows.tsx:17`) [source: octo-review-qwen3.7]

`BuildGroupProfileRowsOptions` and `BuildGroupManagementRowsOptions` still declare `inputEditPush` as required. Callers pass it, but neither function reads it. The parameter was consumed before this PR and became dead from the migration.

### P3 — Dead i18n keys (`en-US.json:1056`) [source: octo-review-qwen3.7]

`subscribers.viewMore` and `subscribers.memberList` remain in both locale files with zero code references after migration to `viewAll` and `memberListWithCount`.

### Nit — remark error toast has no fallback message (`packages/dmworkbase/src/features/channelSetting/channelSettingGroupManagementRows.tsx:224`) [source: octo-review-glm5.2]

`Toast.error(error?.msg)` diverges from the sibling inline-edit rows, which fall back to the newly-added `saveFailedRetry` i18n key (e.g. `channelSettingThreadInfoSection.tsx`, `channelSettingNicknameSection.tsx`). A rejection without a structured `.msg` (raw network throw, non-SDK error) renders an empty/`undefined` toast while the editor stays open. Mirror the sibling pattern:

```ts
Toast.error(error?.msg || t('base.module.channelSettings.saveFailedRetry'))
```

- **Permission guards** — All five edit actions keep `onStartEdit` guards 1:1-equivalent to the pre-migration `onClick`/`inputEditPush` guards; `onStartEdit === false` blocks unauthorized edits; the server still enforces on save. No auth bypass. [qwen3.7 + glm5.2 + qwen3.8]
- **Draft preservation** — `useEffect([value])` with the `editingRef.current` guard correctly preserves mid-edit drafts across external value updates; mutation-pinned (removing the guard turns the suite red). [qwen3.7 + glm5.2 + qwen3.8]
- **Save-`false` / reject contract** — `ChannelSettingInlineEditRow` keeps the editor open when `onSave` returns `false` or rejects; mutation-pinned. Returning `false` (group-name/notice/thread-name/nickname) and re-throwing (remark) are functionally equivalent — both keep `editing=true`. [qwen3.7 + glm5.2 + qwen3.8]
- **i18n parity** — All new keys (`common.clear`, `module.channelSettings.saveFailedRetry`, `subscribers.viewAll`, `subscribers.memberListWithCount`) present in both `en-US` and `zh-CN` with matching `{{count}}`/`{{placeholder}}` sets; qwen3.7 reports 1321/1321 keys in parity; all `t("base.*")` calls resolve. [qwen3.7 + glm5.2 + qwen3.8]
- **Section consolidation** — `buildThreadOverviewSection` correctly merges info + md + webhook into 5 rows (and `undefined` for non-threads); row order matches the old priorities; `flatMap(section => section?.rows || [])` drops no previously-rendered row; registration merge preserves the priority slot and does not reorder thread sections. [qwen3.7 + glm5.2 + qwen3.8]
- **Nickname clear action** — Works on first panel entry; clear button sets draft to empty and saves with `allowEmpty=true`. [qwen3.7]
- **Thread-name `onSave` false-success branch is unreachable** — `threadInfo` is a build-time closure constant, so the early-return cannot coincide with `editing === true` (falsified as a bug by probe). [glm5.2 + qwen3.8]
- **Member-count "wrong total" not reproducible** — on the fallback path the real total is, by premise, unavailable client-side (falsified). [qwen3.8]
- **Security** — no `dangerouslySetInnerHTML`/`innerHTML`/new sinks/credentials/SSRF; user text (name/notice/remark/nickname/subzone) flows through React/Semi text nodes that auto-escape; no new fetch/URL/WebView surface. [glm5.2 + qwen3.8]
- **Scope boundaries** — no changes to `bridge/channelSetting/*Actions`, WKApp/WKSDK/datasource/Chat runtime paths, or permission/entry behavior — consistent with the stated scope. [glm5.2 + qwen3.8]

## Verification Results

| Code Review | APPROVED_WITH_RISK | 1 disputed P1 (falsified as unreachable by 2/3 reviewers) + 5 P2 + 5 P3 + 2 nits (deduped across 3 reviewers), no confirmed blockers |
| E2E Test | PASS | All three: `build:e2e` exit 0 + `vite preview` HTTP 200 at head `71a32ef2`, focused vitest green (9/9 + 11/11), and mutation probes killed (save-false/reject/external-value/nickname/permission contracts pinned); backend Docker stack not applicable — pure-frontend diff with no backend dependency |
| Security | CLEARED | Permission guards 1:1-equivalent; no XSS / new sinks / credentials / SSRF; user text auto-escaped |

PR #1095, opened 2026-07-27
feat(search): support pinyin in global search

Candice00711:
## Summary
- add Chinese, full-pinyin, and case-insensitive matching for global contact and group search
- support pinyin matching in sender, member, and group/thread filter pickers shared by chat and file tabs
- preserve server search, pagination, result ordering, permissions, and filter request contracts
- rebuild scoped indexes when the Space or readable channel pool changes

## Validation
- `pnpm --dir packages/dmworkbase vitest run src/Components/GlobalSearch/__tests__ src/bridge/globalSearch/globalSearchPinyin.test.ts` (103 tests)
- `pnpm i18n:check`
- `pnpm --dir apps/web build`
- `git diff --check`
- manually verified `quanneng` finds 全能接项目小组 in group search and the group/thread filter
- manually verified `jia` finds Chinese names in sender and member filters, including after switching between chat and file tabs

### DIFF
--- a/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
+++ b/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
@@ -274,4 +274,37 @@ describe("loadSenderCandidates (via searchSenders)", () => {
     mockState.contactsList = [];
     await expect(ds.searchSenders("jia")).resolves.toEqual([]);
   });
+
+  it("removes stale sender candidates when the contact pool shrinks in the same space", async () => {
+    mockState.commonDataSource = {
+      searchFriends: vi.fn().mockResolvedValue([]),
+    };
+    mockState.contactsList = [{ uid: "old-uid", name: "贾小明" }];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchSenders("jia")).resolves.toEqual(
+      expect.arrayContaining([expect.objectContaining({ uid: "old-uid" })])
+    );
+
+    mockState.contactsList = [];
+    await expect(ds.searchSenders("jia")).resolves.toEqual([]);
+    expect(ds.getSenders().some((sender) => sender.uid === "old-uid")).toBe(
+      false
+    );
+  });
+
+  it("removes renamed sender pinyin and excludes blacklisted contacts", async () => {
+    mockState.commonDataSource = {
+      searchFriends: vi.fn().mockResolvedValue([]),
+    };
+    mockState.contactsList = [{ uid: "user-1", name: "贾小明" }];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchSenders("jia")).resolves.toHaveLength(1);
+
+    mockState.contactsList = [{ uid: "user-1", name: "张小明" }];
+    await expect(ds.searchSenders("jia")).resolves.toEqual([]);
+    await expect(ds.searchSenders("zhang")).resolves.toHaveLength(1);
+
+    mockState.contactsList = [{ uid: "user-1", name: "张小明", status: 2 }];
+    await expect(ds.searchSenders("zhang")).resolves.toEqual([]);
+  });
 });
--- a/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
+++ b/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
@@ -21,6 +21,8 @@ import { describe, expect, it, vi, beforeEach } from "vitest";
 const mockState = vi.hoisted(() => ({
   commonDataSource: undefined as any,
   contactsList: [] as any[],
+  groups: [] as any[],
+  spaceId: "space-1",
   loginUid: "self-uid",
   loginName: "Me",
 }));
@@ -31,6 +33,9 @@ vi.mock("../../../App", () => ({
       return {
         commonDataSource: mockState.commonDataSource,
         contactsList: mockState.contactsList,
+        channelDataSource: {
+          groupSaveList: vi.fn().mockResolvedValue(mockState.groups),
+        },
       };
     },
     get loginInfo() {
@@ -41,6 +46,9 @@ vi.mock("../../../App", () => ({
       };
     },
     shared: {
+      get currentSpaceId() {
+        return mockState.spaceId;
+      },
       avatarUser: (uid: string) => `avatar://user/${uid}`,
       avatarChannel: (ch: any) =>
         `avatar://ch/${ch?.channelID ?? ""}/${ch?.channelType ?? ""}`,
@@ -52,24 +60,28 @@ vi.mock("../../../App", () => ({
   },
 }));
 
-vi.mock("wukongimjssdk", () => ({
-  Channel: class {
-    channelID: string;
-    channelType: number;
-    constructor(channelID: string, channelType: number) {
-      this.channelID = channelID;
-      this.channelType = channelType;
-    }
-  },
-  ChannelTypeGroup: 2,
-  ChannelTypePerson: 1,
-  WKSDK: {
+vi.mock("wukongimjssdk", () => {
+  const sdk = {
     shared: () => ({
       conversationManager: { conversations: [] },
       channelManager: { getChannelInfo: () => undefined },
     }),
-  },
-}));
+  };
+  return {
+    default: sdk,
+    Channel: class {
+      channelID: string;
+      channelType: number;
+      constructor(channelID: string, channelType: number) {
+        this.channelID = channelID;
+        this.channelType = channelType;
+      }
+    },
+    ChannelTypeGroup: 2,
+    ChannelTypePerson: 1,
+    WKSDK: sdk,
+  };
+});
 
 import { createGlobalSearchApiDataSource } from "../dataSource";
 
@@ -77,6 +89,8 @@ describe("loadSenderCandidates (via searchSenders)", () => {
   beforeEach(() => {
     mockState.commonDataSource = undefined;
     mockState.contactsList = [];
+    mockState.groups = [];
+    mockState.spaceId = "space-1";
   });
 
   it("§1: maps ChannelInfo[] from commonDataSource.searchFriends into ChannelSearchSender", async () => {
@@ -211,5 +225,57 @@ describe("loadSenderCandidates (via searchSenders)", () => {
 
     const ds = createGlobalSearchApiDataSource();
     await expect(ds.searchSenders("")).resolves.toBeDefined();
+  });
+
+  it("matches sender/member candidates by full pinyin without skipping remote search", async () => {
+    const searchFriends = vi.fn().mockResolvedValue([]);
+    mockState.commonDataSource = { searchFriends };
+    mockState.contactsList = [{ uid: "jia-uid", name: "贾小明" }];
+
+    const ds = createGlobalSearchApiDataSource();
+    const results = await ds.searchSenders("jia");
+
+    expect(searchFriends).toHaveBeenCalledWith("jia");
+    expect(results.map((sender) => sender.uid)).toContain("jia-uid");
+  });
+
+  it("matches group options by full pinyin", async () => {
+    mockState.groups = [
+      {
+        channel: { channelID: "group-1", channelType: 2 },
+        displayName: "全能接项目小组",
+      },
+    ];
+
+    const ds = createGlobalSearchApiDataSource();
+    const results = await ds.searchChannels("quanneng");
+
+    expect(results.map((channel) => channel.channelId)).toContain("group-1");
+  });
+
+  it("removes stale channel candidates when the readable pool changes", async () => {
+    mockState.groups = [
+      {
+        channel: { channelID: "group-1", channelType: 2 },
+        displayName: "全能接项目小组",
+      },
+    ];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchChannels("quanneng")).resolves.toHaveLength(1);
+
+    mockState.groups = [];
+    await expect(ds.searchChannels("quanneng")).resolves.toEqual([]);
+  });
+
+  it("clears cached pinyin candidates after switching spaces", async () => {
+    mockState.contactsList = [{ uid: "old-uid", name: "贾小明" }];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchSenders("jia")).resolves.toEqual(
+      expect.arrayContaining([expect.objectContaining({ uid: "old-uid" })])
+    );
+
+    mockState.spaceId = "space-2";
+    mockState.contactsList = [];
+    await expect(ds.searchSenders("jia")).resolves.toEqual([]);
   });
 });
--- a/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
+++ b/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
@@ -155,9 +155,7 @@ describe("loadSenderCandidates (via searchSenders)", () => {
 
   it("§3: surfaces contactsList when searchFriends is missing entirely", async () => {
     mockState.commonDataSource = {}; // no searchFriends method
-    mockState.contactsList = [
-      { uid: "erin-uid", name: "Erin" },
-    ];
+    mockState.contactsList = [{ uid: "erin-uid", name: "Erin" }];
 
     const ds = createGlobalSearchApiDataSource();
     const results = await ds.searchSenders("");
@@ -168,14 +166,12 @@ describe("loadSenderCandidates (via searchSenders)", () => {
   it("§4: cold panel (empty sender cache) still returns candidates from the real DS API — regression guard", async () => {
     // The exact scenario Jerry-Xin flagged: no prior search results have
     // warmed the sender cache. Before the fix, this returned only [self].
-    const searchFriends = vi
-      .fn()
-      .mockResolvedValue([
-        {
-          channel: { channelID: "frank-uid", channelType: 1 },
-          orgData: { displayName: "Frank" },
-        },
-      ]);
+    const searchFriends = vi.fn().mockResolvedValue([
+      {
+        channel: { channelID: "frank-uid", channelType: 1 },
+        orgData: { displayName: "Frank" },
+      },
+    ]);
     mockState.commonDataSource = { searchFriends };
 
     const ds = createGlobalSearchApiDataSource();
--- a/packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts
+++ b/packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts
@@ -13,6 +13,16 @@ import { ProviderListener } from "../../Service/Provider";
 import { debounce } from "../../Utils/rateLimit";
 import { t } from "../../i18n";
 import { addCurrentImChannelInfoListener, getCurrentImChannelInfo } from "../../im-runtime/currentChannelRuntime";
+import type { LegacyGlobalSearchResponse } from "../../Service/SearchService";
+import {
+  createEmptyGlobalSearchPinyinIndex,
+  extendGlobalSearchPinyinIndex,
+  globalSearchContactsToLegacy,
+  globalSearchGroupsToLegacy,
+  mergeGlobalSearchPinyinResults,
+  searchGlobalSearchPinyinIndex,
+  type GlobalSearchPinyinIndex,
+} from "./globalSearchPinyin";
 
 /** Legacy contacts/groups bridge retained while the aggregated tabs migrate. */
 export default class GlobalSearchVM extends ProviderListener {
@@ -31,7 +41,20 @@ export default class GlobalSearchVM extends ProviderListener {
   private unsubscribeChannelInfoListener?: () => void;
   public channel?: Channel; // 查询指定频道的消息
   private requestId = 0; // 请求计数器，用于处理竞态条件
+  private pinyinCandidateRequestId = 0;
+  private pinyinIndex: GlobalSearchPinyinIndex =
+    createEmptyGlobalSearchPinyinIndex();
+  private mounted = false;
   public searchError: string | null = null; // 搜索失败错误信息
+  private readonly contactsChangeListener = () => {
+    void this.refreshPinyinIndex();
+  };
+  private readonly spaceChangedListener = () => {
+    this.pinyinIndex = createEmptyGlobalSearchPinyinIndex();
+    this.initLoad();
+    void this.refreshPinyinIndex();
+    this.requestSearch();
+  };
   // tab数据列表
   public get tabList() {
     if (this.searchInChannel) {
@@ -96,6 +119,10 @@ export default class GlobalSearchVM extends ProviderListener {
   }
 
   didMount(): void {
+    this.mounted = true;
+    WKApp.dataSource.addContactsChangeListener(this.contactsChangeListener);
+    WKApp.mittBus.on("space-changed", this.spaceChangedListener);
+    void this.refreshPinyinIndex();
     this.requestSearch();
 
     this.channelInfoListener = (channelInfo: ChannelInfo) => {
@@ -121,6 +148,12 @@ export default class GlobalSearchVM extends ProviderListener {
   }
 
   didUnMount(): void {
+    this.mounted = false;
+    this.pinyinCandidateRequestId++;
+    this.handleInputChange.cancel();
+    this.handleComposingInputChange.cancel();
+    WKApp.dataSource.removeContactsChangeListener(this.contactsChangeListener);
+    WKApp.mittBus.off("space-changed", this.spaceChangedListener);
     this.unsubscribeChannelInfoListener?.();
     this.unsubscribeChannelInfoListener = undefined;
   }
@@ -134,12 +167,65 @@ export default class GlobalSearchVM extends ProviderListener {
     }
   }, 300);
 
+  // 中文输入法组合期间保留原有“不请求服务端”约束，但像通讯录搜索一样，
+  // 使用输入框里的原始字母即时匹配本地拼音索引。组合结束后该回调会自动
+  // 失效，最终选中的中文仍由 handleInputChange 走原服务端搜索链路。
+  public handleComposingInputChange = debounce((value: string) => {
+    if (!this.isComposing) return;
+    this.keyword = value;
+    this.initLoad();
+  }, 300);
+
   public initLoad() {
     this.page = 1;
     this.loadFinish = false;
     this.loadMoreing = false;
     this.searchResult = null;
+    this.applyPinyinMatches();
     this.notifyListener();
+  }
+
+  private async refreshPinyinIndex() {
+    if (this.searchInChannel) return;
+    const requestId = ++this.pinyinCandidateRequestId;
+    const spaceId = WKApp.shared.currentSpaceId;
+    const contacts = globalSearchContactsToLegacy(
+      WKApp.dataSource.contactsList || []
+    );
+    let groups: ChannelInfo[] = [];
+    try {
+      groups = await WKApp.dataSource.channelDataSource.groupSaveList();
+    } catch {
+      // The existing server search remains authoritative when local candidates
+      // cannot be loaded.
+    }
+    if (
+      !this.mounted ||
+      requestId !== this.pinyinCandidateRequestId ||
+      spaceId !== WKApp.shared.currentSpaceId
+    ) {
+      return;
+    }
+    this.pinyinIndex = extendGlobalSearchPinyinIndex(this.pinyinIndex, {
+      friends: contacts,
+      groups: globalSearchGroupsToLegacy(groups),
+    });
+    this.applyPinyinMatches();
+    this.notifyListener();
+  }
+
+  private applyPinyinMatches() {
+    if (this.searchInChannel || !this.keyword.trim()) return;
+    const localResult = searchGlobalSearchPinyinIndex(
+      this.keyword,
+      this.pinyinIndex
+    );
+    const current: LegacyGlobalSearchResponse = this.searchResult || {
+      friends: [],
+      groups: [],
+      messages: [],
+    };
+    this.searchResult = mergeGlobalSearchPinyinResults(current, localResult);
   }
 
   // 请求搜索
@@ -181,6 +267,14 @@ export default class GlobalSearchVM extends ProviderListener {
         } else {
           this.searchResult = res;
         }
+
+        if (!this.searchInChannel) {
+          this.pinyinIndex = extendGlobalSearchPinyinIndex(
+            this.pinyinIndex,
+            res
+          );
+        }
+        this.applyPinyinMatches();
 
         // 替换备注如果有备注的话
         this.searchResult.friends?.forEach((v: any) => {
--- a/packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts
+++ b/packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts
@@ -12,14 +12,21 @@ import { MessageContentTypeConst } from "../../Service/Const";
 import { ProviderListener } from "../../Service/Provider";
 import { debounce } from "../../Utils/rateLimit";
 import { t } from "../../i18n";
-import { addCurrentImChannelInfoListener, getCurrentImChannelInfo } from "../../im-runtime/currentChannelRuntime";
-import type { LegacyGlobalSearchResponse } from "../../Service/SearchService";
 import {
+  addCurrentImChannelInfoListener,
+  getCurrentImChannelInfo,
+} from "../../im-runtime/currentChannelRuntime";
+import type {
+  LegacyGlobalSearchContact,
+  LegacyGlobalSearchResponse,
+} from "../../Service/SearchService";
+import {
+  buildGlobalSearchPinyinIndex,
   createEmptyGlobalSearchPinyinIndex,
   extendGlobalSearchPinyinIndex,
   globalSearchContactsToLegacy,
-  globalSearchGroupsToLegacy,
-  mergeGlobalSearchPinyinResults,
+  refreshGlobalSearchGroupCandidates,
+  replaceGlobalSearchPinyinMatches,
   searchGlobalSearchPinyinIndex,
   type GlobalSearchPinyinIndex,
 } from "./globalSearchPinyin";
@@ -44,6 +51,11 @@ export default class GlobalSearchVM extends ProviderListener {
   private pinyinCandidateRequestId = 0;
   private pinyinIndex: GlobalSearchPinyinIndex =
     createEmptyGlobalSearchPinyinIndex();
+  private pinyinGroupCandidates: LegacyGlobalSearchContact[] = [];
+  private serverContactGroupResult: Pick<
+    LegacyGlobalSearchResponse,
+    "friends" | "groups"
+  > = { friends: [], groups: [] };
   private mounted = false;
   public searchError: string | null = null; // 搜索失败错误信息
   private readonly contactsChangeListener = () => {
@@ -51,8 +63,8 @@ export default class GlobalSearchVM extends ProviderListener {
   };
   private readonly spaceChangedListener = () => {
     this.pinyinIndex = createEmptyGlobalSearchPinyinIndex();
+    this.pinyinGroupCandidates = [];
     this.initLoad();
-    void this.refreshPinyinIndex();
     this.requestSearch();
   };
   // tab数据列表
@@ -87,9 +99,7 @@ export default class GlobalSearchVM extends ProviderListener {
   // 搜索标题
   public get searchTitle() {
     if (this.searchInChannel) {
-      const channelInfo = getCurrentImChannelInfo(
-        this.channel!
-      );
+      const channelInfo = getCurrentImChannelInfo(this.channel!);
       if (channelInfo) {
         return t("base.globalSearch.chatHistoryWith", {
           values: { name: channelInfo.title },
@@ -122,7 +132,6 @@ export default class GlobalSearchVM extends ProviderListener {
     this.mounted = true;
     WKApp.dataSource.addContactsChangeListener(this.contactsChangeListener);
     WKApp.mittBus.on("space-changed", this.spaceChangedListener);
-    void this.refreshPinyinIndex();
     this.requestSearch();
 
     this.channelInfoListener = (channelInfo: ChannelInfo) => {
@@ -181,6 +190,7 @@ export default class GlobalSearchVM extends ProviderListener {
     this.loadFinish = false;
     this.loadMoreing = false;
     this.searchResult = null;
+    this.serverContactGroupResult = { friends: [], groups: [] };
     this.applyPinyinMatches();
     this.notifyListener();
   }
@@ -192,7 +202,7 @@ export default class GlobalSearchVM extends ProviderListener {
     const contacts = globalSearchContactsToLegacy(
       WKApp.dataSource.contactsList || []
     );
-    let groups: ChannelInfo[] = [];
+    let groups: ChannelInfo[] | undefined;
     try {
       groups = await WKApp.dataSource.channelDataSource.groupSaveList();
     } catch {
@@ -206,9 +216,13 @@ export default class GlobalSearchVM extends ProviderListener {
     ) {
       return;
     }
-    this.pinyinIndex = extendGlobalSearchPinyinIndex(this.pinyinIndex, {
+    this.pinyinGroupCandidates = refreshGlobalSearchGroupCandidates(
+      this.pinyinGroupCandidates,
+      groups
+    );
+    this.pinyinIndex = buildGlobalSearchPinyinIndex({
       friends: contacts,
-      groups: globalSearchGroupsToLegacy(groups),
+      groups: this.pinyinGroupCandidates,
     });
     this.applyPinyinMatches();
     this.notifyListener();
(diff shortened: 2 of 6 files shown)


Jerry-Xin (changes requested):
Project relevance gate passed: the PR is in scope, but it has a blocking stale-index regression in the new pinyin search caches.

**🔴 Blocking**

- 🔴 The legacy global contacts/groups pinyin index is append-only, so removed or no-longer-readable contacts/groups can keep appearing in local pinyin results. `refreshPinyinIndex` in `packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts` rebuilds candidate arrays from the current `contactsList` and `groupSaveList()`, but then calls `extendGlobalSearchPinyinIndex` — and that extender (`packages/dmworkbase/src/bridge/globalSearch/globalSearchPinyin.ts`) only appends unseen `channel_type:channel_id` keys; it never removes or refreshes existing entries. `applyPinyinMatches` then merges those stale local hits into `searchResult` even when the server no longer returns them. This can surface stale contact/group rows after blacklist/removal, contact rename, or a readable-group-pool change, and it contradicts the PR's stated "rebuild scoped indexes when the Space or readable channel pool changes" behavior. The `space-changed` listener does reset the index, but the `contactsChangeListener` path only extends. Fix: on refresh, rebuild this index from the current authoritative scoped source (as the channel-picker path already does via its candidate signature), or make the extender update/remove by key.

- 🔴 The shared sender/member picker has the same stale-candidate behavior within a single space. In `packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts`, `searchSenders` always folds `senderCache` back into `senderPinyinIndex`, and both the cache and index are only cleared when `currentSpaceId` changes (`resetPinyinIndexesForSpace`). If `contactsList` changes in the same space (removal / blacklist), an old sender keeps matching pinyin queries in the sender and member filters until remount or a space switch. Note the channel picker already guards this correctly via `channelCandidateSignature` rebuild — the sender path needs equivalent same-space pool invalidation. Please also add test coverage for the same-space contact-pool-shrink case, not only the existing space-switch test.

**✅ Highlights**

- Remote sender search is preserved and still invoked before local pinyin matching is layered on, so the server remains authoritative.
- `pinyin-pro@3.28.0` is already a declared dependency (root + `packages/dmworkbase`) and present in the lockfile — no new dependency is introduced.
- Pinyin conversion is precomputed once per candidate at index-build time and cached in `searchText`; per-keystroke search is a plain substring scan, and a test asserts 10,000 names convert exactly once and the index is reused.
- Good coverage for case-insensitive pinyin, server-order preservation with local matches appended only when non-duplicate, space switching, remark-based matching, and the channel-picker signature rebuild.

lml2468 (dismissed):
## Review: #1095 — feat(search): support pinyin in global search

**意图**:全局搜索的联系人/群/发送人/成员**候选选择器**支持中文 / 全拼 / 大小写不敏感匹配;保留服务端搜索、分页、排序、权限、filter 请求契约;Space / 可读频道池变化时重建作用域索引。锚定 head `3581cc54`(rev-parse 确认 checkout==live head)/ `mergeable=true`。6 文件 +687/−28。

**结论:APPROVE ✅**

---

### Gate 1 — 规格符合度 ✅
新增 `globalSearchPinyin.ts`(拼音索引/匹配)+ VM/dataSource 接线,给 sender/member/group-thread 选择器加拼音匹配;**服务端搜索契约不变**(dataSource 仍「server-side friend search, keyword-aware」为主,拼音只在**已加载候选**上做客户端过滤/补充,测试「without skipping remote search」锁定)。无 over-build。

### Gate 2 — 代码质量 / 安全 ✅
- **依赖既有,无新增**:`pinyin-pro@^3.28.0` 已在 dmworkbase + root package.json(非本 PR 新加、无 lockfile 漂移);另复用既有 `Utils/pinYin`(legacy 拼音)+ `Utils/t2s`(繁→简)。`defaultPinyinConverter` 组合 legacy + pinyin-pro(toneType:none)双来源,覆盖面稳。
- **无授权变更**:纯客户端过滤已加载的候选(friends/groups/senders/members),不改服务端请求 body、不扩大搜索范围;权限仍由既有服务端搜索兜底。
- **无 XSS**:拼音只产出 `searchText` 供子串匹配,**不渲染**;全 PR **零 `dangerouslySetInnerHTML`/innerHTML** 引入(grep 确认);候选名走既有 React 文本转义。
- **作用域/竞态正确**:`pinyinCandidateRequestId` 陈旧请求守卫 + 切 Space 清缓存(测试「clears cached pinyin candidates after switching spaces」),无跨空间候选串台/泄漏。

**构建 / 测试 / i18n(实跑 @ `3581cc54`)**:`pnpm --filter @octo/web build` ✅ 3.52s;`pnpm i18n:check` ✅;`globalSearchPinyin.test` + `senderCandidates.test` **2 files / 19 tests 全绿**(含全拼匹配、群选项拼音、切空间清缓存、不跳过远程搜索)。

---

### 🟡 非阻断
- 大空间(数千成员)下拼音索引为 O(n) 转换,虽仅在 Space/频道池变化时重建 + 缓存,超大空间可留意构建耗时(可后续加构建上限/懒建)。低。

---

候选选择器拼音匹配用既有 `pinyin-pro` + 繁简归一,纯客户端过滤已加载候选(服务端搜索/契约/权限不变)、切空间清缓存 + requestId 守卫防串台、无 XSS,构建/i18n/19 测试通过。**可合并 —— APPROVE。**

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1095 (octo-web)

## Summary

This PR adds pinyin (romanized Chinese) matching to the global search panel and the sender/member/channel filter pickers. A new `globalSearchPinyin.ts` module builds a precomputed pinyin index from local contacts and groups, enabling instant local matching during IME composition and supplementing server search results with local pinyin hits. The `GlobalSearchVM` wires a contacts-change listener and space-changed event to keep the index fresh. The `GlobalSearchPanel` routes composing input to a new `handleComposingInputChange` handler that searches the local index without hitting the server. Tests cover the core index operations, merge semantics, performance at scale, and integration with the data source layer.

## Verification

Static analysis only at head `3581cc54b1169c14ffd1056692ed6241c9477d01`; build and tests not executed in this environment.

- ✅ **Lifecycle cleanup** — `didUnMount` cancels both debounced handlers, increments `pinyinCandidateRequestId` to invalidate in-flight async work, and unsubscribes from `contactsChangeListener` and `space-changed` event. The `mounted` flag is checked after the async `groupSaveList()` await in `refreshPinyinIndex()`.
- ✅ **Race condition guards** — `pinyinCandidateRequestId` + `spaceId` snapshot + `mounted` triple-check in `refreshPinyinIndex()` prevents stale async results from overwriting a newer index state. The existing `requestId` guard in `requestSearch()` is preserved.
- ✅ **Space switch clears all caches** — VM path: `spaceChangedListener` resets `pinyinIndex` to empty. Data source path: `resetPinyinIndexesForSpace()` clears `senderCache`, `senderPinyinIndex`, `channelPinyinIndex`, and `channelCandidateSignature`.
- ✅ **Composition flow** — `handleComposingInputChange` has `if (!this.isComposing) return` guard. `onCompositionEnd` sets `isComposing = false` and routes to `handleInputChange` (server search). The debounced composing handler won't fire server requests.
- ✅ **`debounce` has `.cancel()`** — confirmed at `Utils/rateLimit.ts:25`.
- ✅ **`addContactsChangeListener` / `removeContactsChangeListener`** — confirmed on `DataSource` class at `Service/DataSource/DataSource.ts:41-52`.
- ✅ **`space-changed` event** — registered in `App.tsx:7`, emitted by `Pages/Chat/vm.ts:103`.

## Findings

### P2 — `localContactSenders()` bypasses blacklist filter (`createGlobalSearchDataSource.ts:107`)

The new `localContactSenders()` function reads `WKApp.dataSource.contactsList` and maps every entry into a `ChannelSearchSender` without checking `ContactsStatus.Blacklist`. The sibling conversion path `globalSearchContactsToLegacy()` in `globalSearchPinyin.ts:218` explicitly filters `.filter((contact) => contact.status !== ContactsStatus.Blacklist)`. Both paths consume the same `contactsList` array, but the data source picker now surfaces blacklisted contacts as sender/member filter candidates while the VM contacts tab does not.

Diff-scope: **new** — `localContactSenders()` is added by this PR. The old `loadSenderCandidates` fallback had the same gap but only activated when the server `searchFriends` call failed. The PR promotes local contacts to always-on inclusion via `[...localContactSenders(), ...remote].forEach(rememberSender)`, amplifying the reach from rare-fallback to always-visible.

Fix direction: add the same blacklist guard:
```ts
if (contact?.status === ContactsStatus.Blacklist) return [];
```

### P2 — VM pinyin index is additive-only; stale entries persist within a session (`globalSearchPinyin.ts:155`)

`extendGlobalSearchPinyinIndex` builds a `seen` set from existing entries and skips any incoming item whose `channel_type:channel_id` key already exists. This means contacts or groups that are removed from the org, or that change their display name, remain in the VM's pinyin index until a space switch or component unmount/remount cycle. The `contactsChangeListener` triggers `refreshPinyinIndex()`, but that function calls `extendGlobalSearchPinyinIndex` which only appends — it cannot retract or update existing entries.

The data source path (`extendNamedPinyinSearchIndex`) handles this correctly: it compares `sourceText` and updates the entry when the source values change. The VM path lacks an equivalent update mechanism.

Diff-scope: **new** — both `extendGlobalSearchPinyinIndex` and its call sites in `GlobalSearchVM` are introduced by this PR.

Fix direction: either (a) rebuild the VM index from scratch in `refreshPinyinIndex()` (the pinyin conversion cost is bounded by the local contacts list size, which the 10K-entry perf test shows is acceptable), or (b) add an update/retraction mechanism to `extendGlobalSearchPinyinIndex` similar to the named index's `sourceText` comparison.

- **`searchGlobalSearchPinyinIndex` empty-keyword behavior** — returns `{ friends: [], groups: [] }` for empty keyword. This is correct because `applyPinyinMatches()` in the VM early-returns on `!this.keyword.trim()`, so the empty-keyword path is never reached.
- **`searchNamedPinyinIndex` empty-keyword behavior** — returns all entries for empty keyword. This is the correct picker behavior (show all candidates when no filter is applied).
- **`mergeGlobalSearchPinyinResults` dedup** — `uniqueByChannel` gives server results priority; local duplicates are filtered. Confirmed by test.
- **`channelCandidateSignature` rebuild** — correctly detects pool changes and rebuilds the channel pinyin index from scratch. Removals are handled.
- **`handleComposingInputChange` + `onCompositionEnd` interaction** — if composition ends before the 300ms debounce fires, the guard `!this.isComposing` prevents the composing handler from running; `onCompositionEnd` already triggered the server search path.
- **`pinyin-pro` + legacy `getPinyin` dual conversion** — both converters are run and concatenated with `\n` separator, maximizing match recall. The `toSimplized` (traditional→simplified) step runs first, handling traditional Chinese input.
- **Test coverage** — 103 tests per PR body; the `globalSearchPinyin.test.ts` covers index build, search, merge, extend, performance, and real group name matching. The `senderCandidates.test.ts` additions cover pinyin matching for senders, groups, stale pool cleanup, and space switching.

## Verdict: APPROVED

Well-structured feature addition with solid test coverage and correct lifecycle management. The two P2 items (blacklist filtering gap and additive-only VM index) are non-blocking consistency issues that can be addressed in a follow-up.

[Octo-Q] verdict: APPROVE — No P0/P1 issues found. Two P2 items (blacklist filter gap in localContactSenders, additive-only VM pinyin index staleness) are non-blocking consistency improvements.

yujiawei (changes requested):
# Code Review — PR #1095 (octo-web)

Reviewed at head `3581cc54b1169c14ffd1056692ed6241c9477d01` against merge-base `07c8e1045e9ba05143d7656e02609e79a07fca37` (6 files, +687/−28).

The core design is sound: pinyin is converted **once** at index-build time and cached in `searchText`, per-keystroke search is a plain substring scan, remote search is still invoked first and stays authoritative, and `pinyin-pro@3.28.0` was already a declared dependency. The problems below are all in the lifecycle of the new caches, not in the idea.

---

**Spec: ❌**

Judged against the four claims in this PR's own Summary (no separate design doc was linked). Two of them do not hold as written.

**Not delivered**

- *"support pinyin matching in … group/thread filter pickers"* — for a saved group that is **not** in the recent-conversation list, the option's `name` is the group ID, so there is no Chinese text to convert and pinyin can never match it. `loadReadableChannelOptions` reads the group name as:

  `packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts:90`
  ```ts
  name: g?.displayName || g?.name || channelId,
  ```

  but `groupSaveList()` builds `ChannelInfo` objects that carry the name on `title` / `orgData.displayName` and have **no** top-level `displayName` or `name` (`packages/dmworkdatasource/src/datasource.ts:102-116`), so both operands are `undefined` and `name` falls through to `channelId`. The conversation branch 20 lines above already reads the correct fields (`info?.orgData?.displayName || info?.title`), which is why manual verification passes for a group you have recently chatted in and would fail for one you have not.

  To be clear: **this naming defect predates this PR** — the old code applied its substring filter to the same broken `option.name`. What is new is that the PR adds a test asserting the behaviour works, and the test only passes because it mocks a shape the real data source never produces:

  `packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts:103-115`
  ```ts
  mockState.groups = [{ channel: { channelID: "group-1", channelType: 2 }, displayName: "全能接项目小组" }];
  const results = await ds.searchChannels("quanneng");
  expect(results.map((channel) => channel.channelId)).toContain("group-1");
  ```

  A green test over a fabricated payload is worse than no test, because it certifies a claim production does not meet. Please fix line 90 to `g?.orgData?.displayName || g?.title || g?.displayName || g?.name || channelId` and change the mock to the real `ChannelInfo` shape (`title` + `orgData.displayName`).

**Diverges from the stated approach**

- *"preserve server search, pagination, result ordering …"* — result ordering is not preserved. See P1-3 and P2-1 below: the entire local contact list is folded into `senderCache` **ahead of** the server-ranked candidates, and the empty-keyword list is then truncated at 50.
- *"rebuild scoped indexes when the Space or readable channel pool changes"* — the rebuild only happens for the channel picker and for `space-changed`. The contact/group index and the sender index are append-only, so a contact-pool change inside the same space never rebuilds anything (P1-1).

**Out-of-scope additions:** none found. The diff stays within global search.

---

## 2. Code quality

### P1-1 — The pinyin indexes are append-only, so renames and removals never take effect

`packages/dmworkbase/src/bridge/globalSearch/globalSearchPinyin.ts:146-171`
```ts
const additions = uniqueByChannel(incoming || [])
  .filter((item) => {
    const key = `${item.channel_type}:${item.channel_id}`;
    if (seen.has(key)) return false;   // <-- existing key: incoming item dropped entirely
    seen.add(key);
    return true;
  })
  .map((item) => ({ item, searchText: searchText(item, toPinyin) }));
return current.concat(additions);
```

`extendGlobalSearchPinyinIndex` keys on `channel_type:channel_id` and **discards** any incoming item whose key is already present — it never compares content and never drops entries. Note that the sibling function 80 lines above, `extendNamedPinyinSearchIndex:65-91`, does exactly the right thing (`if (current?.sourceText === sourceText) { current.item = item; return }` else re-index), which is good evidence this is an oversight rather than a decision.

Two concrete failures, both reachable through the listener this PR adds at `GlobalSearchVM.ts:123`:

- **Rename is ignored.** `contactsSync` is a *delta* sync: a renamed contact arrives as a record with the same `uid` and a new `name` (`packages/dmworkbase/src/Service/DataSource/DataSource.ts:20-30`), then fires `notifyContactsChange`. `refreshPinyinIndex` (`GlobalSearchVM.ts:188-215`) recomputes the candidate list correctly and then hands it to the append-only extender, which drops it. Rename 魏娇莹 → 李娇莹, search `lijiao` → no match; search `weijiao` → still matches, and the row renders the **old** name because the index holds the old object reference.
- **Removal is ignored.** Leave 全能接项目小组; `groupSaveList()` stops returning it, but the entry stays in `pinyinIndex.groups` for the lifetime of the view model. Searching `quanneng` still lists the group in the Groups tab, and clicking it navigates to a channel you can no longer read.

The same class of bug exists one layer down for senders — `senderCache` and `senderPinyinIndex` are only cleared on space change (`createGlobalSearchDataSource.ts:231-241`), never on a same-space contact-pool change:

`packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts:255-261`
```ts
const combined = Array.from(senderCache.values());
extendNamedPinyinSearchIndex(senderPinyinIndex, combined, ...);
```

Suggested fix: make `refreshPinyinIndex` **rebuild** from the authoritative scoped source (`buildGlobalSearchPinyinIndex` is already exported and does this) rather than extend, or teach the extender to update-by-key and prune keys absent from the incoming set — mirroring the `channelCandidateSignature` rebuild the channel picker already uses. Please add coverage for the *same-space pool shrink* and *rename* cases; the existing `removes stale channel candidates when the readable pool changes` test only covers the channel index, which is the one path that already worked.

### P1-2 — `initLoad()` from the composing path does not invalidate the in-flight request

`packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts:173-186`
```ts
public handleComposingInputChange = debounce((value: string) => {
  if (!this.isComposing) return;
  this.keyword = value;
  this.initLoad();          // no requestSearch() -> requestId is never bumped
}, 300);
```

Every other `initLoad()` caller (`handleInputChange:162`, `onTabClick:107`, `spaceChangedListener:52`) pairs it with `requestSearch()`, which is what increments `requestId` and thereby invalidates older responses. The new composing path is the only caller that breaks that invariant.

Failure scenario: type `wei` (no IME active) → `requestSearch()` is in flight. Start composing `quanneng` → after 300 ms `initLoad()` resets `page`/`loadFinish`/`searchResult` and shows local matches for `quanneng`. The `wei` response then arrives, passes `currentRequestId !== this.requestId` because nothing bumped the counter, and executes `this.searchResult = res` (`:269`). The panel now shows server rows for `wei` merged with local rows for `quanneng`.

A related consequence of assigning `this.keyword` mid-composition: `loadMore()` (`:343`, reachable from the list's `loadMore` at `GlobalSearchPanel.tsx:173/181/241`) and `onTabClick` both call `requestSearch()` unconditionally, so the raw half-typed pinyin is now sent to `/search/global`. That contradicts the constraint stated in this PR's own comment at `:170-172` ("组合期间保留原有'不请求服务端'约束"). Bumping `requestId` in `initLoad()` fixes the first half; the second needs the composing keyword kept out of the value `requestSearch` reads (e.g. a separate `localKeyword` field used only by `applyPinyinMatches`).

### P1-3 — Space check happens before the `await`, so results can land in the wrong space's cache

`packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts:251-266`
```ts
searchSenders: async (keyword: string) => {
  resetPinyinIndexesForSpace();                       // checked BEFORE the await
  const remote = await loadSenderCandidates(keyword);
  [...localContactSenders(), ...remote].forEach(rememberSender);
```

Interleaving: a request starts in space A → the user switches to space B → a B request runs `resetPinyinIndexesForSpace()` and clears everything → the A request resolves last and writes A's senders into B's cache and index. `searchChannels:267` has the identical shape.

`refreshPinyinIndex` in the view model already gets this right — it re-checks after the await (`GlobalSearchVM.ts:202-208`, `spaceId !== WKApp.shared.currentSpaceId`). The data-source path needs the same post-await guard.

### P1-4 — `localContactSenders()` does not exclude blacklisted contacts

`packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts:104-118`
```ts
const list: any[] = (WKApp.dataSource as any)?.contactsList ?? [];
return list.flatMap((contact) => { ... });   // no ContactsStatus.Blacklist filter
```

The sibling converter added in the same PR does filter it (`globalSearchPinyin.ts:213-216`, `contact.status !== ContactsStatus.Blacklist`). Because `contactsSync` never removes rows, a blacklisted contact stays in `contactsList` with `status === 2` forever, so blacklisting 张三 removes them from remote results but searching `zhang` still surfaces them in the sender and member pickers. Same one-line filter, please.

### P2-1 — Local contacts are ordered ahead of server-ranked candidates, and can truncate them away

`createGlobalSearchDataSource.ts:254`
```ts
[...localContactSenders(), ...remote].forEach(rememberSender);
```

`senderCache` is a `Map`, and `Map.set` on an existing key preserves the **original** insertion position, so once the whole contact list has been folded in, every server-ranked candidate sorts after every local contact. `searchNamedPinyinIndex` returns entries in map order and `slice(0, PAGE_SIZE_SENDERS)` cuts at 50 (`:262-265`). With ≥50 contacts, the empty-keyword call — the panel-open default — returns self plus the first 49 entries of `contactsList` and drops the server's recommended candidates entirely. Previously only keyword-matching remote results (plus a keyword-filtered local fallback) ever entered the cache.

Related: `getSenders()` (`:244`) now returns the entire contact list, and `GlobalSearchFilterPanel.tsx:115` seeds `senderOptions` from it while `:508` maps it to rendered rows with no cap.

Suggestion: keep the local snapshot as a *fallback* tier rather than a prefix — rank remote hits first, then append local-only matches (which is exactly what `mergeGlobalSearchPinyinResults` does correctly for the legacy path).

### P2-2 — Two converters per value, run synchronously over the full list, and the perf test does not cover the real one

`globalSearchPinyin.ts:41-49`
```ts
const legacyPinyin = getPinyin(simplified).toLowerCase();
const standardPinyin = pinyin(simplified, { toneType: "none", type: "array" }).join("").toLowerCase();
```

`Utils/pinYin.getPinyin` scans the entire ~400-entry dictionary with `indexOf` **per character** (`packages/dmworkbase/src/Utils/pinYin.ts:4-11`). Measured on this machine, `getPinyin` alone costs ≈13 ms / 500 names, ≈44 ms / 2 000, ≈109 ms / 5 000 — before `pinyin-pro` runs on top. And `namedSearchText:51-57` maps **every** value through the converter, including `sender.uid` (`:260`), so a UID is put through a Chinese pinyin dictionary for no possible benefit — roughly 4 converter passes per contact. All of this is synchronous, on the main thread, kicked off from `didMount` (`GlobalSearchVM.ts:125`) and from the first `searchSenders` call.

The test that appears to cover this injects a trivial mock converter:
```ts
const index = buildGlobalSearchPinyinIndex(largeSource, toPinyin);   // toPinyin = vi.fn(...)
expect(toPinyin).toHaveBeenCalledTimes(10_000);
```
so it proves memoization but measures nothing about the production path. Worth (a) skipping the converter for values with no CJK characters, (b) dropping the legacy leg — see P2-3 — and (c) either chunking the initial build or moving it off the first paint.

### P2-3 — The legacy converter leg is redundant and produces a wrong syllable

The legacy dictionary maps 能 to `nen`, not `neng`: `getPinyin("全能接项目小组")` returns `"QuanNenJieXiangMuXiaoZu"`. The `quanneng` test only passes because of the `pinyin-pro` leg. So the legacy leg contributes a false-positive spelling (`nen` matches 全能) and roughly all of the indexing cost, while `pinyin-pro` is both correct and handles 多音字. Consider dropping `getPinyin` here.

### P2-4 — `searchError` survives into the composing path

`initLoad()` does not clear `searchError`, and only `requestSearch()` sets it back to `null` (`:237`) — which the composing path never calls. After a failed search, starting an IME composition renders fresh local matches underneath a stale error banner.

### P2-5 — Per-keystroke work in `searchChannels`

`:267-288` calls `loadReadableChannelOptions()` on every keystroke (which awaits `groupSaveList()`, an HTTP `group/my` call — pre-existing) and then rebuilds an `O(n log n)` signature via `map().sort().join()` over the whole pool even when nothing changed. A cheaper invalidation key (count + a rolling hash, or reusing the response identity) would avoid sorting the full list per character.

### P2-6 — Full syllables only; no initials matching

Both converters emit whole syllables, so 魏娇莹 indexes as `weijiaoying` and the very common initials query `wjy` finds nothing. The PR only claims full pinyin, so this is scope-consistent — flagging it because it is likely the next bug report.

### P2-7 — Duplicate of an existing index in this repo

`packages/dmworkbase/src/features/channelSetting/channelSettingMemberSearch.ts` already implements the same "values + pinyin joined into `searchText`, substring match" index. The new generic `NamedPinyinSearchIndex` is a strict superset of it. Please fold the older one onto the new generic version (or note why not) so the two do not drift.

---

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ❌ and Quality Changes-Requested (four P1 items).

## 4. Suggested direction

1. `refreshPinyinIndex` → rebuild instead of extend; give the sender cache/index a same-space invalidation key equivalent to `channelCandidateSignature`. Cover rename and same-space pool shrink in tests.
2. Bump `requestId` in `initLoad()`, and keep the mid-composition keyword out of whatever `requestSearch` reads.
3. Re-check `currentSpaceId` **after** every `await` in `searchSenders` / `searchChannels`.
4. Filter `ContactsStatus.Blacklist` in `localContactSenders()`.
5. Rank remote hits ahead of the local snapshot instead of prefixing it, so the 50-item cut cannot drop server candidates.
6. Fix the group name lookup at `createGlobalSearchDataSource.ts:90` and rewrite the group test mock to the real `ChannelInfo` shape.

## 5. Additional observations (outside the diff)

- `loadReadableChannelOptions:90` reading a non-existent `displayName` predates this PR but is now load-bearing for the feature — worth fixing in this PR since the picker is in scope.
- `resetPinyinIndexesForSpace` is only reached from `searchSenders` / `searchChannels`; `getSender(uid)` (`:245`) can still resolve a selected chip to the previous space's display name until a search runs.
- The index stores references to the mutable server response objects, which `requestSearch` later rewrites (`GlobalSearchVM.ts:280-289` sets `channel_name = channel_remark`). `searchText` is frozen at insert time, so the indexed text and the rendered name can drift apart. Harmless today because `displayName()` already prefers `channel_remark`, but fragile.

### Verification performed / not performed

- Read the full diff plus every symbol it depends on; confirmed `addContactsChangeListener` / `removeContactsChangeListener`, the `space-changed` bus event, `WKApp.shared.currentSpaceId`, `groupSaveList`, `ContactsStatus`, and `pinyin-pro@^3.28.0` all exist as assumed.
- Benchmarked `Utils/pinYin.getPinyin` directly (numbers in P2-2).
- **Not run:** the vitest suite and `pnpm build` — dependencies are not installed in this review checkout, so the "103 tests" and build claims in the PR description are unverified here.
- **Not verified:** real IME event ordering across browsers (the composing findings are derived from the code, not observed); backend ranking and permission semantics of `/search/global` and `searchFriends`, i.e. whether the local merge can surface anything the server would deliberately withhold; `pinyin-pro`'s own conversion cost.

yujiawei (changes requested):
# Code Review — PR #1095 (octo-web)

**Re-review at head `84da11c2af40de6c09d1a20c8358d0ca5088a33d`** (base `07c8e1045e9ba05143d7656e02609e79a07fca37`, 6 files, +920/−45).

Housekeeping first: my previous review was written against `3581cc54` and the head advanced to `84da11c2` (`fix(search): rebuild scoped pinyin results`, `fix(search): invalidate stale sender candidates`) while it was in flight, so GitHub attached it to a commit it had not read. **This review supersedes it.** Two of its four P1s are genuinely fixed; two are not; and the fix commits introduce one new P1. Please treat the list below as the current one.

### Resolved since `3581cc54`

- ✅ **Append-only pinyin index.** `refreshPinyinIndex` now calls `buildGlobalSearchPinyinIndex` on the authoritative snapshot instead of extending it, and `refreshGlobalSearchGroupCandidates` correctly distinguishes "load failed" (`undefined` → keep last snapshot) from "authoritatively empty" (`[]` → drop everything). `replaceGlobalSearchPinyinMatches` rebases friends/groups on the latest server baseline so previously merged stale rows fall out. Rename and removal both take effect now, and the new tests cover both.
- ✅ **Same-space sender staleness.** `senderCandidateSignature` gives the sender cache/index the same-space invalidation the channel picker already had, with coverage for pool-shrink and rename.
- ✅ **Blacklisted contacts.** Filtered in `localContactSenders` and `loadSenderCandidates`, plus a `blacklistedUids` guard against a lagging remote response. Nice belt-and-braces.

---

**Spec: ❌**

**Not delivered** — unchanged from the previous head. *"support pinyin matching in … group/thread filter pickers"* still does not hold for a saved group that is not in the recent-conversation list:

`packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts:93`
```ts
name: g?.displayName || g?.name || channelId,
```

`groupSaveList()` builds `ChannelInfo` objects carrying the name on `title` / `orgData.displayName`, with **no** top-level `displayName` or `name` (`packages/dmworkdatasource/src/datasource.ts:102-116`), so both operands are `undefined` and `name` falls through to `channelId` — there is no Chinese text left to convert. The conversation branch 20 lines above already reads the right fields (`info?.orgData?.displayName || info?.title`), which is why manual verification passes for a group you have recently chatted in and would fail for one you have not.

This naming defect predates the PR — but the PR adds a test asserting the behaviour works, and it only passes because it mocks a shape the real data source never produces:

`packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts:103-115`
```ts
mockState.groups = [{ channel: { channelID: "group-1", channelType: 2 }, displayName: "全能接项目小组" }];
```

A green test over a fabricated payload is worse than no test, because it certifies a claim production does not meet. Please fix line 93 to `g?.orgData?.displayName || g?.title || g?.displayName || g?.name || channelId` and re-mock to the real `ChannelInfo` shape (`title` + `orgData.displayName`). Note `globalSearchGroupsToLegacy` gets this right (it falls through to `org.displayName`) — only the picker path is wrong.

**Diverges from the stated approach**

- *"preserve … result ordering"* — still not preserved; see P2-1.
- The fix commits trade away the PR's own stated performance property; see P1-2.

**Out-of-scope additions:** none. The diff stays within global search.

---

## 2. Code quality

### P1-1 — `initLoad()` from the composing path still does not invalidate the in-flight request

`packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts:182-186`
```ts
public handleComposingInputChange = debounce((value: string) => {
  if (!this.isComposing) return;
  this.keyword = value;
  this.initLoad();          // no requestSearch() -> requestId never bumped
}, 300);
```

Every other `initLoad()` caller (`handleInputChange`, `onTabClick`, `spaceChangedListener`) pairs it with `requestSearch()`, which is what increments `requestId` and thereby invalidates older responses. The composing path is the only caller that breaks that invariant, and it is unchanged at this head.

Failure scenario: type `wei` with no IME active → `requestSearch()` in flight. Start composing `quanneng` → after 300 ms `initLoad()` resets `page` / `loadFinish` / `searchResult` / `serverContactGroupResult` and renders local matches for `quanneng`. The `wei` response then arrives, passes `currentRequestId !== this.requestId` because nothing bumped the counter, sets `serverContactGroupResult` to `wei`'s friends/groups and `this.searchResult = res` — and the following `applyPinyinMatches()` re-merges local matches for `quanneng`. The panel ends up showing server rows for `wei` alongside local rows for `quanneng`.

Fix: bump `requestId` inside `initLoad()`.

Second half of the same problem: assigning `this.keyword` mid-composition means `loadMore()` (`:371`, reachable from `GlobalSearchPanel.tsx:173/181/241`) and `onTabClick` call `requestSearch()` with the raw half-typed pinyin, which contradicts the constraint stated in this PR's own comment at `:179-181` ("组合期间保留原有'不请求服务端'约束"). Keeping the composition text in a separate field that only `applyPinyinMatches` reads would settle both halves.

### P1-2 — NEW: the fix rebuilds the whole pinyin index on every search, defeating the memoisation the PR is built on

`packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts:250-256`
```ts
public requestSearch() {
  if (!this.searchInChannel) {
    void this.refreshPinyinIndex();
  }
  this.requestId++;
```

and `:223-226`
```ts
this.pinyinIndex = buildGlobalSearchPinyinIndex({
  friends: contacts,
  groups: this.pinyinGroupCandidates,
});
```

`buildGlobalSearchPinyinIndex` converts **every** name unconditionally — there is no cache keyed on the source text, unlike `extendNamedPinyinSearchIndex`, which compares `sourceText` and skips unchanged entries. Moving the rebuild onto `requestSearch()` therefore means the entire contacts + groups list is re-run through **both** converters on every debounced keystroke, every tab switch, every `loadMore()` page, and every space change.

`defaultPinyinConverter` (`globalSearchPinyin.ts:39-47`) runs `Utils/pinYin.getPinyin` — which scans the whole ~400-entry dictionary with `indexOf` **per character** (`packages/dmworkbase/src/Utils/pinYin.ts:4-11`) — and then `pinyin-pro` on top. Measured on this machine, `getPinyin` alone costs ≈13 ms / 500 names, ≈44 ms / 2 000, ≈109 ms / 5 000, before `pinyin-pro` runs. So a 5 000-contact directory pays >110 ms of synchronous main-thread work every 300 ms while the user types.

This is exactly the property the existing test was protecting:
```ts
expect(toPinyin).toHaveBeenCalledTimes(10_000);   // "converts 10,000 names once and reuses the index"
```
That test still passes because it asserts against a single `buildGlobalSearchPinyinIndex` call with an injected mock converter — it cannot observe that the *caller* now rebuilds per search. The invariant is gone while its guard is still green.

The same line also issues a `groupSaveList()` → `GET group/my` on every search request, so typing one query now costs several HTTP round-trips that previously did not happen.

The correctness fix is right; it just needs to keep the conversions cheap. Suggested: memoise conversions in a module-level `Map<sourceText, pinyin>` so a rebuild re-links cached strings instead of re-converting, and short-circuit `refreshPinyinIndex` when neither the contacts snapshot nor the group snapshot has changed (a signature, as the sender path now does). Please also add a test that asserts the converter is *not* re-invoked across two consecutive `requestSearch()` calls — otherwise this regresses again silently.

### P1-3 — Space-switch race is narrowed but still leaks across spaces

`createGlobalSearchDataSource.ts:267-283` now calls `resetScopedPinyinIndexes()` a second time **after** the await, which is a real improvement. But `remote` was resolved for the *old* space and is inserted after that reset:
```ts
resetScopedPinyinIndexes();
const remote = await loadSenderCandidates(keyword);
...
resetScopedPinyinIndexes();                     // clears B's cache
[...localContactSenders(), ...remote]           // ...then inserts A's remote senders
  .filter((sender) => !blacklistedUids.has(sender.uid))
  .forEach(rememberSender);
```

Interleaving: request starts in space A → user switches to B → A resolves → B's picker lists A's remote senders. And because the stored signature is recomputed from the live (B) contacts, a subsequent search sees no change and does not clear, so the leaked rows persist.

`searchChannels:295` has no post-await check at all, and `channelCandidateSignature` is only reset when `spaceChanged` is true — so an in-flight A request that resolves after the switch rebuilds the channel index from A's options.

The view model already does this correctly: capture the space id before the await and bail if it changed (`GlobalSearchVM.ts:212-217`). Please apply the same guard here rather than resetting twice. Downgraded from P1 to **P2** in effect — it needs a space switch mid-flight — but the one-line guard is cheap and the asymmetry with `refreshPinyinIndex` is worth closing.

### P2-1 — Local contacts still sort ahead of server-ranked candidates, and can truncate them away

`createGlobalSearchDataSource.ts:279`
```ts
[...localContactSenders(), ...remote]
```

`senderCache` is a `Map`, and `Map.set` on an existing key keeps its original position, so once the local snapshot is folded in, every server-ranked candidate sorts after every local contact. `searchNamedPinyinIndex` returns entries in map order and `slice(0, PAGE_SIZE_SENDERS)` cuts at 50 (`:288-292`). With ≥50 contacts the empty-keyword call — the panel-open default — returns self plus the first 49 entries of `contactsList` and drops the server's recommended candidates entirely. Previously only keyword-matching remote results, plus a keyword-filtered local fallback, ever entered the cache.

Suggestion: treat the local snapshot as a *fallback tier* — remote hits first, then append local-only matches. That is exactly what `mergeGlobalSearchPinyinResults` already does correctly for the legacy path.

### P2-2 — `resetScopedPinyinIndexes()` builds the whole sender list just to hash it

`createGlobalSearchDataSource.ts:234-240`
```ts
const nextSenderCandidateSignature = localContactSenders()
  .map((sender) => `${sender.uid}:${sender.name}`)
  .sort()
  .join("\n");
```

`localContactSenders()` allocates one object per contact and calls `WKApp.shared.avatarUser(uid)` for each, purely to derive a signature that only needs `uid` and `remark || name`. This now runs on `getSenders()` (`:257-259`), on `searchChannels` (`:295`), and **twice** per `searchSenders` — so a keystroke in the sender picker costs 2N allocations plus two `O(n log n)` sorts. Hash straight off `contactsList` instead, and consider a cheaper key than a sorted full join (count + rolling hash).

### P2-3 — Per-keystroke work in `searchChannels`

`:294-316` still calls `loadReadableChannelOptions()` on every keystroke — which awaits `groupSaveList()`, an HTTP `GET group/my` (pre-existing) — and then rebuilds an `O(n log n)` signature via `map().sort().join()` over the whole pool even when nothing changed.

### P2-4 — `searchError` survives into the composing path

`initLoad()` does not clear `searchError`, and only `requestSearch()` resets it to `null` (`:261`) — which the composing path never calls. After a failed search, starting an IME composition renders fresh local matches underneath a stale error banner.

### P2-5 — The legacy converter leg is redundant and produces a wrong syllable

The legacy dictionary maps 能 to `nen`, not `neng`: `getPinyin("全能接项目小组")` returns `"QuanNenJieXiangMuXiaoZu"`. The `quanneng` test passes only because of the `pinyin-pro` leg. So the legacy leg contributes a false-positive spelling (`nen` matches 全能) plus roughly all of the indexing cost measured in P1-2, while `pinyin-pro` is both correct and handles 多音字. Dropping `getPinyin` here would fix most of P1-2 on its own.

### P2-6 — Full syllables only; no initials matching

Both converters emit whole syllables, so 魏娇莹 indexes as `weijiaoying` and the very common initials query `wjy` finds nothing. The PR only claims full pinyin, so this is scope-consistent — flagging it because it is likely the next bug report.

### P2-7 — Duplicate of an existing index in this repo

`packages/dmworkbase/src/features/channelSetting/channelSettingMemberSearch.ts` already implements the same "values + pinyin joined into `searchText`, substring match" index. The new generic `NamedPinyinSearchIndex` is a strict superset. Please fold the older one onto it, or note why not, so the two do not drift.

### P2-8 — `extendGlobalSearchPinyinIndex` is now dead weight

With `refreshPinyinIndex` switched to `buildGlobalSearchPinyinIndex`, the only remaining caller of `extendGlobalSearchPinyinIndex` is the server-response path at `GlobalSearchVM.ts:300`, which re-introduces exactly the append-only accumulation the fix commit removed elsewhere — server hits for earlier keywords stay in the index for the life of the view model. Either drop it and let the authoritative rebuild own the index, or give it the same update-by-key semantics as `extendNamedPinyinSearchIndex`.

---

## 3. Overall verdict

**CHANGES_REQUESTED** — Spec ❌ and Quality Changes-Requested (two P1 items, one of them new at this head).

## 4. Suggested direction

1. Bump `requestId` in `initLoad()`, and keep the mid-composition keyword out of whatever `requestSearch` reads.
2. Memoise pinyin conversion by source text and short-circuit `refreshPinyinIndex` when neither snapshot changed — then add a test asserting the converter is not re-invoked across consecutive searches.
3. Capture `currentSpaceId` before each `await` in `searchSenders` / `searchChannels` and bail if it changed, instead of resetting twice.
4. Rank remote hits ahead of the local snapshot so the 50-item cut cannot drop server candidates.
5. Fix the group name lookup at `createGlobalSearchDataSource.ts:93` and rewrite the group test mock to the real `ChannelInfo` shape.
6. Derive the sender signature from `contactsList` directly rather than from mapped sender objects.

## 5. Process note

This is the fifth review pass on this PR (three at `3581cc54`, two of them since dismissed, plus mine, plus this one), and the head has moved twice mid-review. The remaining items are no longer independent bugs — they are all consequences of one question: **who owns the local candidate snapshot, and when is it allowed to be rebuilt?** Point-fixing them one round at a time is what produced the P1-2 perf regression while fixing the P1 correctness bug. I would suggest one consolidated pass that settles the ownership model — a single `LocalCandidateSnapshot` with an explicit invalidation key, a memoised converter, and the view model reading it rather than maintaining a parallel index — instead of another round of patches. Happy to review that as a whole.

## 6. Additional observations (outside the diff)

- `resetScopedPinyinIndexes` is not reached from `getSender(uid)` (`:261-265`), so a selected chip can still resolve to the previous space's display name until a search or `getSenders()` call runs.
- The index stores references to the mutable server response objects, which `requestSearch` later rewrites (`GlobalSearchVM.ts:308-316` sets `channel_name = channel_remark`). `searchText` is frozen at insert time, so indexed text and rendered name can drift. Harmless today because `displayName()` already prefers `channel_remark`, but fragile.

### Verification performed / not performed

- Read the full diff at both heads plus every symbol it depends on; confirmed `addContactsChangeListener` / `removeContactsChangeListener`, the `space-changed` bus event, `WKApp.shared.currentSpaceId`, `groupSaveList`, `ContactsStatus`, and `pinyin-pro@^3.28.0` all exist as assumed, and traced `groupSaveList`'s real `ChannelInfo` field layout to `packages/dmworkdatasource/src/datasource.ts`.
- Benchmarked `Utils/pinYin.getPinyin` directly (numbers in P1-2).
- **Not run:** the vitest suite and `pnpm build` — dependencies are not installed in this review checkout, so the test-count and build claims in the PR description are unverified here.
- **Not verified:** real IME event ordering across browsers (the composing findings are derived from code, not observed); backend ranking and permission semantics of `/search/global` and `searchFriends`, i.e. whether the local merge can surface anything the server would deliberately withhold; `pinyin-pro`'s own conversion cost, which adds to the P1-2 measurement rather than being included in it.

Candice00711:
Closing this PR to replace it with a clean, fresh-main implementation that preserves the existing search request, ordering, pagination, cache, and IME behavior while adding only scoped pinyin matching. The replacement PR will include consolidated lifecycle and regression coverage.

PR #1096, opened 2026-07-27
fix(html-doc): use concrete version for comment writes

### LINKED ISSUE #1093: fix: HTML anchored comments must write the rendered numeric version
## Description

HTML 文档页面按 `latest` 等读取别名展示具体版本后，评论写请求仍把读取参数作为 `version` 发送。HTML 评论写接口要求具体数字版本，因此后端解码在 `version` 处失败，后续 `anchor` 字段未被读取，最终可能形成正文已写入但锚点丢失的残缺评论。

上游 HTML 后端问题及完整链路见 Mininglamp-OSS/octo-docs-html#21。

## Root cause

评论列表读取与评论 mutation 共用了同一个字符串版本参数：

- 读取可使用 `latest` 等别名；
- 写入事件必须绑定用户当前看到的不可变数字版本；
- 当前创建评论把读取别名直接写入 POST body。

渲染结果中的 `window.__ODOC__.version` 已提供实际数字版本，应作为 mutation 版本来源。

## Proposed fix

- 保留列表查询的字符串版本参数。
- 为评论 mutation 单独传入渲染得到的数字版本。
- 收紧创建评论 input 类型，`version` 只允许 `number`。
- 实际数字版本缺失时禁止提交并显示错误，不静默回退到版本 1。

## Acceptance criteria

- [ ] 路由/读取版本为 `latest` 时，创建评论请求发送渲染得到的具体数字版本。
- [ ] 元素锚点和文本锚点均完整出现在 POST body。
- [ ] TypeScript 类型阻止 `latest`、`v2` 等字符串进入评论写接口。
- [ ] 渲染元数据缺少数字版本时不发送写请求，并给出可见错误。
- [ ] 自动化测试覆盖 `latest` 页面写入具体版本数字的链路。

## Linked issue

- Mininglamp-OSS/octo-docs-html#21

Ranwanglc:
## Summary

- separate the route/list version (`latest` or `vN`) from the concrete mutation version
- send `window.__ODOC__.version` as a number when creating HTML document comments
- block comment submission with a visible error when the rendered version is missing or invalid
- tighten the create-comment TypeScript contract and add regression coverage

## Root cause

The Web client reused the route selector (`latest` / `vN`) in comment mutation requests. `octo-docs-html` expects a concrete immutable numeric version, so decoding stopped before the later `anchor` field. The backend previously ignored that error and silently persisted an anchorless comment at version 1.

This is the Web half of the cross-repository fix. The defensive backend half is Mininglamp-OSS/octo-docs-html#23.

## Verification

- targeted Vitest: 3 files, 91 tests passed
- `git diff --check`: passed
- `@octo/web` production build reached bundling of the changed HTML-doc module, then failed on a pre-existing unrelated missing `lucide-react` dependency from `packages/dmworkmcp/src/pages/McpMarketListPage.tsx`
- docs typecheck reached the project and reported two pre-existing unrelated errors:
  - `packages/dmworkbase/src/i18n/format.ts`: `fractionalSecondDigits` typing
  - `packages/docs/src/members/sort.ts`: missing `grantedBy`

## Deployment prerequisites / required environment variables

None. This change adds no environment variables, schema changes, or migrations. Deploy together with Mininglamp-OSS/octo-docs-html#23 to complete the contract fix.

Closes #1093


lml2468 (dismissed):
## Review: #1096 — fix(html-doc): use concrete version for comment writes

**意图**:修 #1093 —— HTML 文档评论**写入**误用路由选择器(`latest`/`vN`)当版本,而 octo-docs-html 要的是具体不可变数字版本;解码在 `anchor` 前中止,后端静默存了个无 anchor、version=1 的评论。本 PR 把「路由/列表版本」与「具体写入版本」分离,写入发 `window.__ODOC__.version`(number),缺失/非法则挡提交并可见报错,并收紧 TS 契约。锚定 head `271b1bcc`(rev-parse 确认 checkout==live head)/ `mergeable=true`。8 文件 +75/−28。Web 半,配套 octo-docs-html#23。

**结论:APPROVE ✅**

### Gate 1 — 规格符合度 ✅
路由/列表版本(`latest`/`vN`,string)与写入版本(number)分离;createComment 发具体数字版本;缺失/非法版本挡提交 + 可见错误;TS 契约收紧。与描述一致,无 over-build。

### Gate 2 — 代码质量 ✅
- **版本分离正确**:`CreateCommentInput.version: number`(:129,写入具体数字),`listComments(version: string)`(:114,读用路由选择器);`instructionDoc.version = listVersion`(列表仍用路由版本,读路径不受影响)。
- **fail-closed 校验**:提交时 `const version = mutationVersion; if (typeof version !== 'number' || !Number.isInteger(version) || version <= 0) { setError(t('docs.comment.errorVersion')); return }` —— 版本缺失/非正整数时**挡提交并显式报错**,不再让坏版本落库产生无 anchor 评论。
- **具体版本来源**:`mutationVersion` 取自 render 页注入的 `window.__ODOC__.version`(`__ODOC__` 元数据类型 `version?: number`),非路由选择器。
- **无 XSS**:错误经 `<div role="alert">{error}</div>` React 文本节点。
- **契约收紧**:create 版本类型由宽松改为 `number`,tsc 层挡住再次误传路由选择器字符串。

**构建 / 测试 / i18n(实跑 @ `271b1bcc`)**:docs html(`HtmlDocCommentPanel` + `HtmlDocView` + `htmlDocComments`)**3 files / 91 tests 全绿**;`errorVersion` i18n 中英双语齐全(:413)。`@octo/web` build 失败是**既有无关问题**(`packages/dmworkmcp/src/pages/McpMarketListPage.tsx` 缺 `lucide-react`,该文件不在本 PR 改动内)—— 非本 PR 引入,PR body 已如实说明。

---

### 🟡 非阻断 / human-verify
1. 依赖 render 页始终注入 `window.__ODOC__.version`(number);缺失时已 fail-closed 报错(不是静默损坏),但请确认 render 页在所有版本路径都注入,否则评论创建会被挡(有明确提示,可接受)。
2. 跨仓:配套 octo-docs-html#23(后端防御半,拒绝无 anchor/坏版本)—— 两边一起落地才闭环。
3. 顺带提醒:`@octo/web` 生产构建当前因 dmworkmcp 的 `lucide-react` 缺失而红(既有、与本 PR 无关),建议另行修。

---

版本分离正确(写入用具体数字、列表用路由选择器)、缺失/非法版本 fail-closed 挡提交并可见报错、TS 契约收紧防再次误传、无 XSS,docs html 91 测试通过。**可合并 —— APPROVE**(合入前确认 render 页注入 `__ODOC__.version` + 配套后端 #23 一起上;`@octo/web` 的 lucide-react 构建红为既有无关问题)。

Jerry-Xin (dismissed):
The PR is relevant to `Mininglamp-OSS/octo-web` and correctly separates comment listing versions from concrete mutation versions.

💬 Non-blocking
- 🟡 Warning: In `packages/docs/src/html/HtmlDocCommentPanel.tsx:119-122`, `reload()` clears `error` after listing succeeds. If the initial comment-list request resolves after a user submits with an invalid `mutationVersion`, it can clear the visible `docs.comment.errorVersion` set at `:135-137`. The mutation is still blocked, so this is not a merge blocker, but preserving mutation errors separately from list-load errors would make the new visible-error behavior more robust.

✅ Highlights
- 🔵 Good: `packages/docs/src/html/HtmlDocView.tsx:728-735` passes the route/list selector separately from the rendered document version, which directly addresses the root cause.
- 🔵 Good: `packages/docs/src/html/HtmlDocCommentPanel.tsx:134-145` validates that mutation writes only use positive integer versions before posting.
- 🔵 Good: `packages/docs/src/html/htmlDocComments.ts:127-151` tightens the write contract to numeric `version`, and the updated tests cover both request shape and the `latest` route regression.

Verification note: I attempted to run the three targeted Vitest files locally, but this checkout’s pnpm environment could not find the `vitest` binary for `@octo/docs`. I reviewed the code and tests directly instead.

mochashanyao (dismissed):
[Octo-Q · automated review]

---

# Code Review — PR #1096 (octo-web)

Reviewed at head `271b1bcc`. Scope: html-doc comment write versioning — split route-selector (`listVersion`) from concrete mutation version (`mutationVersion`).

## Summary

This PR fixes a semantic issue in the html-doc comment system: previously the same string version (e.g. `"latest"` or `"v3"`) was sent for both listing comments (GET) and writing comments (POST). Writing with `version: "latest"` is ambiguous — the backend should receive a concrete numeric version to pin the comment to the exact document revision the commenter was viewing.

The fix splits the single `version: string` prop on `HtmlDocCommentPanel` into `listVersion: string` (for listing, unchanged semantics) and `mutationVersion?: number | null` (for writes, sourced from the rendered `window.__ODOC__.version` injected by the doc server). A guard in `submit()` rejects the POST when the concrete version is missing, non-integer, or non-positive — fail-closed behavior with a user-facing error message. The `CreateCommentInput.version` type in the data layer is widened from `string` to `number`. All call sites, tests, and i18n strings are updated consistently.

The change is well-scoped, the data flow separation is clean, and the guard logic is comprehensive. No P0/P1 issues found.

## Verification

- ✅ **Data flow traced** — `mutationVersion` originates from `parseOdocMeta(html).version` (`HtmlDocView.tsx:286`), extracted via regex from `window.__ODOC__` in the rendered HTML. The regex and JSON.parse are pre-existing; only the consumption point is new.
- ✅ **Guard correctness** — `submit()` at `HtmlDocCommentPanel.tsx:134` checks `typeof version !== 'number' || !Number.isInteger(version) || version <= 0`, covering undefined, null, NaN, string, fractional (1.5), zero, and negative values. Sets error state and returns before `setBusy(true)`, so no in-flight state leak.
- ✅ **Single call site** — `HtmlDocCommentPanel` is instantiated only in `HtmlDocView.tsx:729`, correctly passing `listVersion={version}` and `mutationVersion={meta?.version}`.
- ✅ **Data layer consistency** — `CreateCommentInput.version` (`htmlDocComments.ts:129`) is now `number`; `createComment` serializes it directly into the POST body. All test call sites updated from string to number.
- ✅ **Test coverage** — unit tests verify: concrete version POST (`mutationVersion={2}` → `version: 2`), guard rejection (`undefined`/`0`/`1.5` → no POST + error shown), and integration test exercises the full `__ODOC__` → `parseOdocMeta` → `mutationVersion` → POST body path.
- ✅ **i18n parity** — `errorVersion` added in both `en-US.json` and `zh-CN.json` at the same key path.
- ✅ **No stale callers** — grep confirms no other production code references the old `version` prop or calls `createComment` outside the updated paths.

Static analysis only at head `271b1bcc`; build and tests not executed in this environment.

## Findings

### P2 — `instructionDoc.version` still passes the route string (`packages/docs/src/html/HtmlDocCommentPanel.tsx:159`)

The "让 AI 处理" forward constructs `instructionDoc = { docId, slug, space, version: listVersion }`. When a user is on the `/latest` route, `listVersion` = `"latest"`, so the forwarded instruction embeds `version=latest` in the `octo-doc://agent-handle?` link. The downstream ring-3 (AI agent) receiver won't know which concrete version the commenter was viewing. This is pre-existing behavior (the old code also sent the route string), so not introduced by this PR — but now that `mutationVersion` carries the concrete number, it would be a natural improvement to include it in the forward payload too. Not a blocker.

### P2 — Missing integration test for version-less `__ODOC__` (`packages/docs/src/html/HtmlDocView.test.tsx:460`)

The integration test covers the happy path (`__ODOC__ = {"version":4}`) but not the realistic failure mode where `__ODOC__` exists without a `version` field (e.g. `{"slug":"x"}`). In production, older doc renders may omit `version`, making `meta.version` undefined and blocking all comment writes with the errorVersion message. The code handles this correctly (fail-closed), but an `HtmlDocView`-level test asserting the error is shown and no POST fires for a version-less `__ODOC__` blob would lock that contract against regression.

## Data Flow Trace

|---|---|---|
| `mutationVersion` prop | `HtmlDocView.tsx:735` → `meta?.version` | ✅ `meta` = `state.meta` when `status === 'ready'`; set at `:486` via `parseOdocMeta(html)` regex extraction of `window.__ODOC__` |
| `OctoDocMeta.version` | `parseOdocMeta()` at `:279` → regex `__ODOC__\s*=\s*(\{[\s\S]*?\});` → `JSON.parse(m[1])` | ✅ Typed `number \| undefined`; pre-existing regex, unchanged by PR |
| `listVersion` prop | `HtmlDocView.tsx:734` → `version` (component prop, default `'latest'`) | ✅ Unchanged route selector |
| `submit()` guard | `HtmlDocCommentPanel.tsx:134-138` | ✅ Checks typeof/Integer/positive before `setBusy(true)` |
| `createComment(slug, {version})` | `HtmlDocCommentPanel.tsx:141-145` | ✅ Passes validated `version` (number) to POST body |

## Blind-spot Checklist

- **C1 (dual path parity)**: N/A — no paired add/remove paths changed.
- **C2 (control-flow ordering)**: Version guard runs before `setBusy(true)`; no nested-reuse. ✅ Clear.
- **C3 (authorization boundary)**: No auth changes. ✅ Clear.
- **C4 (authorization lifecycle)**: N/A — no auth/status fields changed.
- **C5 (build/runtime parity)**: Integration test exercises real `parseOdocMeta` path. ✅ Clear.
- **C6 (governance docs)**: N/A — no governance/policy files changed.

## Cross-round Blocker Review

- `listComments(slug, listVersion)` contract preserved — still receives route string, no behavior change for listing.
- `AgentInstructionDoc.version` type still `string`, consistent with `listVersion` usage. No type mismatch.
- Error state correctly uses `role="alert"` for accessibility.
- i18n keys are consistent between en-US and zh-CN.
- The `HtmlDocCommentPanel.test.tsx` `it.each` guard test covers the three realistic failure modes (missing, zero, fractional).

## Verdict: APPROVED

Clean, well-scoped fix that correctly separates the listing version (route selector) from the mutation version (concrete revision number). The guard logic is comprehensive and fail-closed. Two non-blocking P2 suggestions for future improvement.

[Octo-Q] verdict: APPROVE — No P0/P1 issues. The version split is correct, the guard is comprehensive, and the data flow is well-traced. Two P2 suggestions (instructionDoc could carry concrete version; test gap for version-less __ODOC__) are non-blocking improvements.

yujiawei (dismissed):
# Code Review — PR #1096 (octo-web)

Reviewed at head SHA `271b1bcc9a9f46e5c377c9632e2ae52fa3639c97`, merge-base `07c8e1045e9ba05143d7656e02609e79a07fca37` (8 files, +75/−28). Verified locally: the four affected Vitest files pass (96 tests), and the `@octo/docs` typecheck reports only the two pre-existing failures already documented in the PR body (`packages/dmworkbase/src/i18n/format.ts`, `packages/docs/src/members/sort.ts` — neither is touched here).

## 1. Spec compliance

**Spec: ✅**

Checked against the five acceptance criteria in #1093:

| Criterion | Evidence |
|---|---|
| `latest` route sends the concrete numeric version | `HtmlDocView.tsx:734-735` splits `listVersion={version}` / `mutationVersion={meta?.version}`; `HtmlDocView.test.tsx:487-492` asserts the GET uses `version=latest` while the POST body carries `version: 4` |
| Element and text anchors reach the POST body | `htmlDocComments.ts:156` forwards the anchor object verbatim; element path covered in `HtmlDocView.test.tsx:494`, text path in `HtmlDocCommentPanel.test.tsx:128-157` |
| TypeScript blocks `latest` / `v2` from the write API | `htmlDocComments.ts:129` `version: number`; `HtmlDocCommentPanel.tsx:37` `mutationVersion?: number \| null` |
| No write and a visible error when the numeric version is absent | `HtmlDocCommentPanel.tsx:134-138` returns before `setBusy`; rendered through the existing `role="alert"` region at `:175-179`; `it.each` at `HtmlDocCommentPanel.test.tsx:161-186` covers `undefined` / `0` / `1.5` and asserts no POST |
| Automated coverage of the `latest` → concrete-number path | `HtmlDocView.test.tsx:455-495` |

- Nothing missing.
- Nothing added beyond scope: the diff is the prop split, the type tightening, the guard, one new i18n key in both locales, and tests. No new flags, endpoints, or env vars.
- No deviation from the stated approach. The route/list selector is deliberately kept as a string for `listComments` (`htmlDocComments.ts:112-116`), which is what the issue asks for.

I also confirmed the contract holds against the real producer, not just the test double: the render side always emits the field (`OverlayConfig.Version int \`json:"version"\`` with no `omitempty`) and the `/d/{slug}/v/{n}` handler populates it from the resolved version parameter, so the `/v/{version}` URL that `buildOctoDocUrl` (`HtmlDocView.tsx:242-245`) is the only thing this component ever fetches always carries a positive integer.

## 2. Code quality

**Quality: Approved** — no P0/P1. Four P2s, none of which should hold up the merge.

### P2-1 — an in-flight list refresh can wipe the version error
`HtmlDocCommentPanel.tsx:119-126` and `:134-138` share one `error` slot:

```tsx
const reload = useCallback(async () => {
  try {
    setThreads(await listComments(slug, listVersion))
    setError(null)        // <-- clears whatever error is displayed
```

The mount-time `reload()` runs independently of the composer, and the textarea is interactive while it is still in flight. Sequence: panel mounts → list GET in flight → user types and clicks **Send** while `mutationVersion` is unusable → `errorVersion` renders → the GET resolves → `setError(null)` silently removes the alert. The draft text and the enabled Send button survive, so nothing is lost, but the user is left with a comment that was never posted and no visible reason. Cheapest fix: keep the submit error in its own state slot, or have `reload` clear only the list error it owns.

### P2-2 — the `__ODOC__` parse is now load-bearing but is still best-effort
`HtmlDocView.tsx:276-286`:

```ts
const m = html.match(/__ODOC__\s*=\s*(\{[\s\S]*?\});/)
```

Its own comment says a parse miss "just means no header metadata" — that was true until this PR, because `meta.version` had no reader anywhere in the repo before now. The non-greedy capture stops at the first `};`, and the producer's script-safe encoder neutralizes `</script` and `<!--` but not `};`. A doc title or author display name containing `};` (for example a title with an inline code snippet) truncates the capture → `JSON.parse` throws → `meta` is `null` → `mutationVersion` is `undefined` → commenting is hard-blocked on that document, and the error text ("Refresh and try again") points at a remedy that cannot work. Worth hardening the version acquisition — anchor the match on `};</script>`, or fall back to the versions API when the inline parse misses.

### P2-3 / P2-4 — anchor assertions are weaker than the acceptance criterion
`HtmlDocView.test.tsx:494` and `HtmlDocCommentPanel.test.tsx:151-157` assert anchors with `toMatchObject({ kind, aid })` / `{ kind, text }`. The production path forwards the anchor verbatim so it is correct today, but a future regression that dropped `selector`, `label`, or `context_before` / `context_after` would keep these tests green while violating "both anchor kinds appear fully in the POST body." Pinning the full anchor object in at least one test would close that gap.

### Considered and cleared

- **Rejecting `version === 0`** (`HtmlDocCommentPanel.tsx:135`) is correct, not over-strict. The only surface that publishes `version: 0` is the draft render (`mode: "draft"`, "not yet a committed version"); this component never requests it — `buildOctoDocUrl` only builds `/d/{slug}/v/{version}`, and the version panel opens `/v/{n}`. Binding a comment to an uncommitted draft slot is exactly the silent-fallback behavior #1093 set out to remove.
- **`latest` list vs. pinned write version.** A comment written at v4 while `latest` advances to v5 still appears after the reload: the list request with `latest` returns the full comment history rather than a single-version snapshot, so there is no "posted but invisible" window here.
- **`AgentInstructionDoc.version` still receives `listVersion`** (`HtmlDocCommentPanel.tsx:159`). That is unchanged behavior and the right type: the field is a route coordinate serialized into `octo-doc://agent-handle?...&version=…` (`htmlDocAnchor.ts:97,144`), where a bare `4` would not be a valid route selector. If the AI work order should pin an immutable snapshot instead, that is a separate change to the ring-3 contract, not this PR.
- i18n: `docs.comment.errorVersion` is present in both `en-US.json` and `zh-CN.json`, and the parity guard in `src/i18n/i18n.test.ts` passes. Those are the only two locale files in this package.

## 3. Verdict

**APPROVE.** The change matches the issue's contract exactly, fails closed instead of corrupting data, and the new tests cover the paths that actually regressed in production. The four P2s are follow-ups; P2-1 and P2-2 are the two I would pick up next, since both turn a correct guard into a confusing dead end for the user.

## 4. Suggestions

1. Give the submit-time version error its own state slot so a concurrent list refresh cannot erase it (`HtmlDocCommentPanel.tsx:119-126`).
2. Harden the inline-metadata parse now that a write path depends on it, and give the user a real recovery route when it misses (`HtmlDocView.tsx:278-286`).
3. Assert the complete anchor object in one element-anchor and one text-anchor test.

## 5. Additional notes

- PR #920 also modifies `HtmlDocCommentPanel.tsx` and `HtmlDocView.tsx` (it consumes `__ODOC__` creator fields). Whichever lands second will need a rebase, and it should re-run the comment-panel suite rather than trusting a clean textual merge.
- This PR is behind `main` by a few commits, none of which touch `packages/docs/`, so no semantic conflict from the base.
- The backend half (octo-docs-html#23) must ship with this, as the PR body states: this client stops sending a string version, but nothing here prevents an older client from doing so.

lml2468 (approved):
## Re-review: #1096 — fix(html-doc): use concrete version for comment writes

**新提交 `08c48b0371c9`(`fix(html-doc): render lost comment anchors`)在我上轮 APPROVE @ `271b1bcc` 之上加了一条**丢失/未知锚点的健壮渲染修复**。锚定 head `08c48b0371c9`(rev-parse 确认 checkout==live head)/ `mergeable=true`。增量 6 文件 +79/−4。上轮的「版本分离」逻辑本次未触碰。

**结论:APPROVE ✅(增量为纯健壮性修复,消除 `"undefined"` 泄漏与无效 querySelector,测试到位)**

### 增量验证(`271b1bcc..08c48b037`)
- **修的真实缺陷**:octo-doc 对无法在当前版本重连的锚点返回 `kind:'lost'`(无 `text`/`aid`)。旧 `anchorLabel` 是 `if kind==='element' … else return "${anchor.text}"` —— 任何非 element 锚点(含 `lost` 与未来新 kind)都落入 text 分支,渲染出字面 `"undefined"`,并被送进 `resolveAnchorText` 做 `querySelector([data-odoc-aid="undefined"])`。
- **修法正确**:`anchorLabel` 改 `switch(anchor.kind)`——`element`/`text` 原样;`lost` → 本地化 `anchorLostWithLabel`(带后端 label)或 `anchorLost`;`default` → `anchorUnknown`(对**尚未进联合类型的 wire kind** 的防御,联合已穷尽故 default 为 never,运行时兜底)。
- **解析守卫两处对称**:
  - Panel:`canResolveAnchor = kind==='element' || 'text'`,否则不调 `resolveAnchorText`;`fallbackAnchorText` 仅对 `kind==='text'` 返回文本、否则 `null` → lost/unknown **不产生引用块、不泄漏 undefined**。
  - `resolveHtmlDocAnchorText`(HtmlDocView.tsx:134)新增 `if (anchor.kind !== 'element') return null`,挡住 lost/unknown 落到 `querySelector(anchor.aid)`(否则 `cssAttrValue(undefined)` 会抛)。
- **类型**:`LostAnchor{kind:'lost';reason?;label?;fingerprint?;fallback?}` 纳入 `Anchor` 联合;switch 各分支正确窄化(`lost` 分支可读 `.label`)。i18n `anchorLost`/`anchorLostWithLabel`/`anchorUnknown` 中英齐全。

### 上轮结论仍成立
版本分离(写入具体数字版本 / 列表用路由选择器)、fail-closed 版本校验、TS 契约收紧 —— 均不在本次增量内,保持不变。

**构建 / 测试 / i18n(实跑 @ `08c48b0371c9`)**:`pnpm --filter @octo/web build` ✅ 3.75s;`pnpm i18n:check` ✅;docs html `HtmlDocCommentPanel` + `HtmlDocView` + `htmlDocComments` **3 文件 93/93 全绿**(较上轮 91 增 2:lost 锚点显示本地化标签且保留后端 label、未知 wire kind 显示 `anchorUnknown` 且**不渲染 `undefined`**)。

---

### 🟡 非阻断 / 沿用上轮
- 依赖 render 页注入 `window.__ODOC__.version`(缺失 fail-closed 报错,可接受);跨仓配套 octo-docs-html#23 一起落地闭环。`@octo/web` 生产构建此前的 lucide-react 红为既有无关问题(本地此次 build 通过,以 CI 为准)。
- `LostAnchor.fingerprint`/`fallback`(如 `nearestHeading`)目前仅入类型、UI 未展示——预留字段,非缺陷;后续若要展示「附近标题」提示可增量利用。

---

增量把 lost/未知锚点从「渲染 `undefined` + 无效 querySelector」修为「本地化标签 + 解析守卫」,联合类型/兜底/i18n 齐备,上轮版本分离逻辑未受影响,构建/i18n/93 测试通过。**可合并 —— APPROVE。**

Jerry-Xin (approved):
Re-review at `08c48b0371c9` (🔄). Prior APPROVE at `271b1bcc9a9f` was auto-dismissed by the new push. One new commit only ("render lost comment anchors"), not a rebase (ahead 1 / behind 0). APPROVE.

The concrete-version comment-write fix from the prior head is untouched: the delta on `HtmlDocCommentPanel.tsx` does not modify `submit()`, the `listVersion`/`mutationVersion` split, the positive-integer guard, or the POST body. Read path still lists on `listVersion` (latest/vN route); write path still resolves the concrete integer `mutationVersion`, guards it (`typeof number && Number.isInteger && > 0`) before POST, and shows the i18n `errorVersion` otherwise. No stale-write / lost-update reintroduced.

The new commit is a companion display fix for comments whose backend anchor can no longer be reconciled:
- `htmlDocComments.ts`: adds a `LostAnchor` (`kind: 'lost'`) variant to the `Anchor` union.
- `HtmlDocCommentPanel.tsx`: `anchorLabel()` becomes a `switch` — `element`/`text` unchanged, `lost` → localized `anchorLostWithLabel` / `anchorLost`, and a `default` → `anchorUnknown` so an unsupported future wire kind renders a real label instead of `undefined`.
- `HtmlDocView.tsx`: `resolveHtmlDocAnchorText()` now early-returns `null` for any non-`element` kind, so a lost/unknown anchor never reaches `querySelector([data-odoc-aid="undefined"])`. The panel mirrors this with a `canResolveAnchor` gate before calling `resolveAnchorText?.()`, falling through to `fallbackAnchorText` (null) and rendering the label div rather than a broken quote.
- i18n en-US + zh-CN add the three new keys consistently.

Verification:
- `HtmlDocCommentPanel.test.tsx` + `HtmlDocView.test.tsx` pass (84/84) at this head. New tests assert the localized lost-anchor label, backend-label preservation, and that an unknown kind renders `anchorUnknown` with no `undefined` leaking to the DOM.
- Version-fix tests remain strong: POST body asserted `version: 2` (concrete integer, not mock-matched); the missing/zero/fractional trio each → no POST + `errorVersion`; list stays on the route selector.
- Null/undefined anchor still routed to the doc-level label; comment before a resolved version still blocked.

Non-blocking (carried over, not a regression): in `HtmlDocCommentPanel.tsx` `reload()` clears `error` on a successful list load, so a late initial list fetch could clear a visible `errorVersion`. The mutation is still blocked by the guard, so this remains visual-only and out of scope for this commit.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1096 (octo-web)

## Summary

This PR fixes the root cause of lost comment anchors in HTML documents (issue #1093) by separating the route/list version selector (`latest` or `vN`) from the concrete numeric mutation version. The route version continues to address the comment list endpoint, while comment writes now send the immutable numeric version extracted from `window.__ODOC__.version` injected by the backend into the rendered HTML. A submit-time guard blocks comment creation when the rendered version is missing or invalid, replacing the previous silent failure where the backend dropped the anchor field and persisted an anchorless comment at version 1.

The change also extends the anchor model with a `LostAnchor` kind for comments whose target the backend can no longer reconcile in the current document version, and an `anchorUnknown` fallback for unrecognized wire kinds. The `anchorLabel` function now uses an exhaustive switch, and `resolveHtmlDocAnchorText` correctly returns `null` for non-element/non-text anchors before attempting DOM queries.

## Verification

Static analysis only at head `08c48b03`; build and tests not executed in this environment.

- ✅ **Data flow: `mutationVersion`** — traced from `window.__ODOC__` (backend-injected script tag) → `parseOdocMeta` regex → `OctoDocMeta.version: number` → `state.meta` → `meta?.version` → `mutationVersion` prop → `submit()` guard → `createComment({ version: number })` → POST body. The version is always a concrete positive integer when the request fires.
- ✅ **Submit guard** — `HtmlDocCommentPanel.tsx:148-152` rejects `undefined`, `null`, `0`, negative, and fractional values (`typeof !== 'number' || !Number.isInteger(version) || version <= 0`). When `__ODOC__` is absent or lacks `version`, `meta?.version` is `undefined` and the guard fires correctly.
- ✅ **Anchor resolution guard** — `resolveHtmlDocAnchorText` (`HtmlDocView.tsx:136`) returns `null` before DOM query when `anchor.kind !== 'element'`, preventing `undefined` aid access on `LostAnchor`. The parallel guard in the panel (`:196`) prevents `resolveAnchorText` from being called with unsupported kinds.
- ✅ **Type contract** — `CreateCommentInput.version` changed from `string` to `number`; all test assertions updated; `listComments` still takes `version: string` (list route selector).
- ✅ **Anchor label exhaustiveness** — `anchorLabel` switch covers `element`, `text`, `lost` (with/without label), and `default` for unknown kinds. No `undefined` string can leak to the UI.
- ✅ **i18n parity** — four new keys added in both `en-US.json` and `zh-CN.json`.
- ✅ **No stale callers** — `HtmlDocCommentPanel` is rendered only in `HtmlDocView.tsx:730`; all test renders updated to `listVersion`/`mutationVersion`.

## Findings

No P0/P1 issues. One nit and one pre-existing observation below.

### Nit — `resolveAnchorText` guard not tested with callback present (`packages/docs/src/html/HtmlDocCommentPanel.tsx:196`)

The `canResolveAnchor` guard correctly prevents `resolveAnchorText` from being called with `lost`/unknown anchor kinds. However, the "unknown anchor" test (`HtmlDocCommentPanel.test.tsx` ~line 141) does not pass `resolveAnchorText` to the panel, so the branch where the callback IS provided alongside a `lost`/unknown anchor is not locked by a regression test. Low risk — the guard is a two-condition check — but a test rendering the panel with `resolveAnchorText` + a `lost` comment would close the gap.

### P2 (pre-existing) — `parseOdocMeta` regex fragility (`packages/docs/src/html/HtmlDocView.tsx:280`)

The non-greedy regex `/__ODOC__\s*=\s*(\{[\s\S]*?\});/` captures up to the first `};` in the HTML. Currently safe because `OctoDocMeta` contains only scalar fields, but a future backend addition of a nested object whose closing `}` is followed by `;` on the same line would truncate the capture and silently null out `meta?.version`, blocking all comment writes. Not introduced by this PR. A brace-balanced or JSON-aware extractor would be more robust. Not a merge blocker.

- `AgentInstructionDoc` still receives `listVersion` as `version: string` for the "让 AI 处理" link builder — correct, the forward flow uses the route selector, not the mutation version.
- `listComments` API contract unchanged: still sends `version` as a string query param.
- Test coverage for the submit guard covers three invalid-value shapes (`undefined`, `0`, `1.5`) via `it.each` — thorough.
- The two commits split cleanly: first commit does the version-type refactor, second adds the lost-anchor rendering.
- `createComment` exclusive contract (reply → `parent_id` only, root → `anchor` only) is preserved.

## Verdict: APPROVED

Well-structured cross-cutting fix with thorough test coverage. The version-type split, submit guard, and anchor-kind handling are all correct and well-tested. The single nit (test gap for `resolveAnchorText` guard with callback present) is non-blocking. The pre-existing regex observation is noted for future awareness only.





[Octo-Q] verdict: APPROVE — No P0/P1 issues. The version-type split, submit guard, and anchor-kind handling are all correct and well-tested. One nit (test gap for resolveAnchorText guard) is non-blocking.

yujiawei (approved):
# Code Review — PR #1096 (octo-web)

Reviewed at head SHA `08c48b0371c997abcc2e4bb91f7bd349b260de1b`, merge-base `07c8e1045e9ba05143d7656e02609e79a07fca37` (8 files, +154/−32, 2 commits).

**Verified locally at this head**, not taken on trust:
- `packages/docs`: `vitest run` on the three changed suites plus `src/i18n/i18n.test.ts` → **98 tests / 4 files passed**.
- `tsc --noEmit -p tsconfig.typecheck.json` → only the two pre-existing unrelated errors the PR body already discloses (`dmworkbase/src/i18n/format.ts` `fractionalSecondDigits`, `docs/src/members/sort.ts` `grantedBy`). Nothing new from this diff.
- `node scripts/i18n-scan.mjs check` → `0 candidates within baseline; locale keys healthy` (both locales carry all four new keys).
- The cross-repo contract this change depends on was read from the document backend's current `main`, not assumed (details in §5).

---

## 1. Spec compliance

**Spec: ✅**

Acceptance criteria from the linked issue (#1093), each checked against code rather than the PR description:

| AC | Status | Evidence |
|---|---|---|
| `latest` route ⇒ create-comment sends the concrete rendered numeric version | ✅ | `HtmlDocView.tsx:736` passes `mutationVersion={meta?.version}` (parsed from the rendered document at `:487`); `HtmlDocCommentPanel.tsx:144-153` posts that value. Regression test in `HtmlDocView.test.tsx` ("uses the rendered numeric version when posting from the latest route with an element anchor") asserts GET `version=latest` **and** POST `body.version === 4`. |
| element + text anchors appear intact in the POST body | ✅ | `htmlDocComments.ts:162-185` unchanged in that respect; asserted in `HtmlDocCommentPanel.test.tsx` (`anchor: {kind:'text',…}`) and `htmlDocComments.test.ts` (`anchor: {kind:'element', aid:'a42'}`). |
| TypeScript blocks `latest` / `v2` strings reaching the write path | ✅ | `htmlDocComments.ts:147` `version: number` on `CreateCommentInput`; the only caller is `HtmlDocCommentPanel.tsx:151`, which narrows to `number` first. |
| Missing concrete version ⇒ no write request + visible error | ✅ | `HtmlDocCommentPanel.tsx:144-148` guards `typeof` / `Number.isInteger` / `> 0` before `setBusy`, and `it.each(['missing','zero','fractional'])` asserts `role="alert"` contains `docs.comment.errorVersion` **and** that no POST was issued. |
| automated coverage of the latest-page → concrete-number path | ✅ | as above, `HtmlDocView.test.tsx`. |

- **Missing:** none.
- **Extra scope:** the second commit (`08c48b0`, "render lost comment anchors") is outside the linked issue's acceptance criteria — it adds `LostAnchor` (`htmlDocComments.ts:58-66`), three i18n keys, the `anchorLabel` switch with an `anchorUnknown` default (`HtmlDocCommentPanel.tsx:52-66`) and two resolvability guards. I checked whether this is speculative and it is not: the document backend really does emit `kind:"lost"` anchors (`internal/core/reconcile.go`: `Kind:"lost", Reason:"no_candidate"`), and the pre-change `anchorLabel` fell through to `` `“${anchor.text}”` `` for every non-element kind — i.e. it rendered a literal `“undefined”` to users. Real adjacent defect, no new flag/endpoint/persisted field, test-covered, so I am **not** treating it as a blocking scope violation. It does need the PR description updated (N-6): the Summary still describes only the version change, so release notes and any future bisect would miss half of the PR.
- **Deviation:** none. The route/list selector stays a string (`listVersion`), the mutation version is a separate concrete number, and the failure mode is fail-closed rather than the old silent version-1 fallback — exactly what the issue asked for.

---

## 2. Code quality

**Quality: Approved** — no P0/P1. Seven non-blocking notes; N-1 is the one I would fast-follow.

### N-1 (P2, highest value) — the write path now depends on a best-effort HTML regex, and document content can win that race

`HtmlDocView.tsx:277-287`:

```ts
// Pull the __ODOC__ blob the render page inlines. Best-effort: a parse miss just means
// no header metadata (header still renders with slug fallback).
function parseOdocMeta(html: string): OctoDocMeta | null {
  const m = html.match(/__ODOC__\s*=\s*(\{[\s\S]*?\});/)
```

That comment was true before this PR; it is stale now. `meta.version` is no longer decorative — `:736` makes it the value written to the backend. Two consequences, both reproduced locally against this exact regex:

1. **Content shadowing (the more interesting one).** The server injects its config immediately before `</body>` (`core.InjectOverlayCfg`), i.e. *after* all document content, and this regex takes the **first** match. A published document whose body contains a literal `window.__ODOC__ = {…};` — plausible for any agent-authored doc that documents this overlay — shadows the real config:

   ```
   body   → <pre>window.__ODOC__ = {"slug":"demo","version":1};</pre>
   real   → <script>window.__ODOC__ = {"slug":"real","version":7,…};</script></body>
   parsed → {"slug":"demo","version":1}
   ```

   `1` is a positive integer, so the new guard passes and the comment is written at **version 1** — precisely the outcome the linked issue exists to eliminate, reached through a different door. Pre-existing weakness, but this PR is what makes it reach a mutation.
2. **Fail-closed dead end.** The regex is non-greedy up to the first `};`, so a document *title* containing `};` truncates the match and `JSON.parse` throws → `meta = null` → every comment on that doc is refused with "Could not verify the current document version. Refresh and try again." Refreshing cannot help (the render effect only re-runs on `[effectiveSlug, version]`, `:504`), so the advice is misleading.

Suggested minimal hardening, in a file you are already touching: take the **last** match rather than the first (the injected block is always last), or scope extraction to the `<script>window.__ODOC__ = … ;</script>` shape. Two lines; worth doing before any document about this overlay is published through the pipeline.

### N-2 (P2) — `mutationVersion` is optional, so a future call site fails closed silently

`HtmlDocCommentPanel.tsx:37` (`mutationVersion?: number | null`) plus the strict runtime guard at `:145` means any new mount that forgets the prop still compiles, renders a normal-looking composer, and then refuses every submission. Today there is exactly one call site (`HtmlDocView.tsx:736`; both app entry points — `StandaloneDocPage.tsx:602` and `DocsHome.tsx:1606` — go through it), so this is latent rather than live. `mutationVersion: number | null` (required, explicitly nullable) forces the next caller to decide.

### N-3 (P2) — the new prop doc encodes a contract the backend does not honor

`HtmlDocCommentPanel.tsx:34`: `/** Route selector used for listing comments (`latest` or `vN`). */`. The backend parses that query param with `strconv.Atoi` plus a special case for `latest`, so `v1`/`v2` do **not** select version N — they fall back to latest. Only tests pass `vN`; both production call sites omit `version` entirely, so `listVersion` is always `latest` in the app. Recommend "`latest` or a numeric version string" so the comment doesn't cement a false contract for the next reader.

### N-4 (P2, pre-existing) — the AI-handoff link still carries the route selector

`HtmlDocCommentPanel.tsx:169`: `{ docId, slug, space, version: listVersion }` → `buildAgentInstruction` puts `version=latest` into the `octo-doc://agent-handle` link. Unchanged by this PR (it previously passed the same value), but now that a concrete version is in scope, the agent could be pinned to the snapshot the human actually commented on instead of re-resolving `latest` at click time. Follow-up, not a change request.

### N-5 (P2) — the create-comment input type now accepts a server-only anchor kind

`htmlDocComments.ts:66` widens `Anchor` to include `LostAnchor`, and `CreateCommentInput.anchor` (`:149`) reuses that union — so `{kind:'lost'}` is now type-legal on a write, even though `lost` is a read-side state the backend produces during reconciliation and a client should never post. Splitting the union (`type WritableAnchor = TextAnchor | ElementAnchor` for input, full `Anchor` for wire reads) keeps the write contract as tight as the `version: number` change just made it.

### N-6 (P2) — PR description does not cover the second commit

The Summary's four bullets describe only the version contract; the lost/unknown-anchor rendering, the new `LostAnchor` wire type and the three i18n keys are unmentioned. Please add a bullet, ideally with the reason it belongs here (the `“undefined”` label).

### N-7 (nit) — the same invariant is now guarded in two places

`HtmlDocView.tsx:136` (`if (anchor.kind !== 'element') return null`) and `HtmlDocCommentPanel.tsx:193` (`canResolveAnchor`) both encode "only element/text anchors are resolvable". Defensive duplication is defensible for a caller-supplied callback, but if a third resolvable kind appears, both sites must change and only one sits near the type. A one-line comment pointing at the union would help.

---

## 3. Verdict

**APPROVE.** The fix is correct, minimal in the right places, fails closed instead of silently writing version 1, and is covered by tests that would actually fail on regression (they assert both the absence of a POST and the concrete numeric body). Everything above is P2/nit; N-1 is a pre-existing parser weakness that this PR promotes onto the write path and is worth a fast follow-up rather than a block.

---

## 4. Suggested changes (concrete)

1. `HtmlDocView.tsx:280` — match the **last** `__ODOC__ = {…};` occurrence (or scope to the injected `<script>`), and refresh the now-stale "best-effort / only header metadata" comment above it to say the value is load-bearing for comment writes.
2. `HtmlDocCommentPanel.tsx:37` — `mutationVersion: number | null` (required, explicitly nullable).
3. `htmlDocComments.ts:149` — use a write-only anchor union that excludes `lost`.
4. `HtmlDocCommentPanel.tsx:34` — correct the `vN` claim in the `listVersion` doc comment.
5. PR body — add the lost-anchor rendering to the Summary.
6. Optional follow-up: pass the concrete version into `AgentInstructionDoc` (`:169`).

---

## 5. Additional observations (outside the diff, checked because the change depends on them)

- **The version source is genuinely concrete.** The `/d/{slug}/v/{version}` render path resolves `latest` server-side to the newest published `N` before injecting the config, and the field is a plain `int` with no `omitempty`, so a positive integer is always present on the surface this client fetches. The only path injecting `version: 0` is the draft render (`/d/{slug}/draft`), which `buildOctoDocUrl` never requests — and if it ever did, the new guard fails closed rather than writing garbage. The premise of the fix holds.
- **No deploy-ordering hazard.** The comment-create handler on the document backend's current `main` already decodes `version` as a JSON number, so a numeric body is accepted today; the companion backend PR (strict decoding + string/`vN`/`latest` compatibility) is not a prerequisite for this one. Landing this alone is safe and already removes the client-side source of anchorless comments.
- **Read/write version split is safe on the production path.** Reads keep the moving `latest` selector while writes pin the rendered number, which raises the question of whether a comment posted at version 4 can vanish from a reload after version 5 publishes. It cannot: for `latest` the backend returns the full history fold rather than a version-scoped snapshot, so comments created at earlier versions stay listed. The divergence would only matter for a caller passing a concrete list version — which, per N-3, no production caller does.
- **Republish race semantics look intentional.** If the document is republished between render and submit, the comment binds to the version the human actually saw and backend reconciliation handles the drift — which is what the `lost` anchor kind this PR now renders is for.
- **No other mutation carries a version in this module.** `htmlDocComments.ts` exposes only `listComments` (string query param, fine) and `createComment`; the HTML comment rail has no delete/reanchor/reaction path, so no second write site is left on the old string contract. The `deleteComment` hits under `packages/docs/src/comments/` belong to the separate collaborative-editor comment stack and are unrelated.
- **Formatting is not a finding here.** Prettier flags all three changed source files, but it flags them identically at the merge base and there is no formatting gate in CI — pre-existing, ignore.

### Coverage / what I did not verify

- No browser or device run: the `role="alert"` copy, the lost-anchor label and the Chinese/English strings were checked in jsdom and by reading the locale files, not visually.
- Full-repo `build` and the entire `packages/docs` suite were not run end-to-end; I ran the changed suites plus the i18n suite and the package typecheck. The PR's note about the unrelated `lucide-react` build failure in `packages/dmworkmcp` is consistent with CI, which is green on this head.
- Backend behavior was read from source at the document repo's current `main`; I did not exercise a live deployment, so "numeric version accepted" is verified by code and its tests, not by an HTTP round trip.
- Two advisory second-opinion passes were consulted; N-2, N-4 and N-5 originate there. Two of their claims did **not** survive verification and are deliberately excluded rather than reported: (a) that `t(key, { values: {…} })` fails to interpolate — this codebase's own `t` takes exactly that shape and interpolates `{{name}}` placeholders, and the existing locale strings use it; (b) that the string→number `version` change risks 400s during rollout — the backend has decoded that field as a JSON number since its initial import, so the old string was the failing case, not the new number.

PR #1106, opened 2026-07-27
fix(loop): prevent IME Enter submitting create issue modal

xiaokenbot:
## Summary
- guard the create-issue title Enter shortcut while an IME composition is active
- add a jsdom regression test for composing vs non-composing Enter

## Tests
- `pnpm --filter @octo/loop test -- src/ui/__tests__/CreateIssueModal.test.tsx`
- `pnpm exec tsc -p packages/dmloop/tsconfig.json --noEmit` (fails on existing Semi dependency type errors under node_modules)

Fixes WS-25

### DIFF
--- a/packages/dmloop/src/ui/CreateIssueModal.tsx
+++ b/packages/dmloop/src/ui/CreateIssueModal.tsx
@@ -282,7 +282,7 @@ export default function CreateIssueModal({
           onChange={(e) => setTitle(e.target.value)}
           placeholder={t("loop.field.titlePlaceholder")}
           onKeyDown={(e) => {
-            if (e.key === "Enter") submit();
+            if (e.key === "Enter" && !e.nativeEvent.isComposing) submit();
           }}
         />
         <AutoGrowTextarea
@@ -349,7 +349,11 @@ export default function CreateIssueModal({
               style={{ flex: 1 }}
             >
               {labels.map((label) => (
-                <Select.Option key={label.id} value={label.id} label={label.name}>
+                <Select.Option
+                  key={label.id}
+                  value={label.id}
+                  label={label.name}
+                >
                   <LabelChips labels={[label]} />
                 </Select.Option>
               ))}
--- a/packages/dmloop/src/ui/__tests__/CreateIssueModal.test.tsx
+++ b/packages/dmloop/src/ui/__tests__/CreateIssueModal.test.tsx
@@ -1,0 +1,155 @@
+/**
+ * @vitest-environment jsdom
+ */
+import React from "react";
+import ReactDOM from "react-dom";
+import { act, Simulate } from "react-dom/test-utils";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const hoisted = vi.hoisted(() => ({
+  createIssue: vi.fn().mockResolvedValue({ id: "issue-1" }),
+  listAssigneeCandidates: vi.fn().mockResolvedValue([]),
+  listProjectOptions: vi.fn().mockResolvedValue([]),
+  uploadAttachment: vi.fn(),
+  attachLabel: vi.fn(),
+  listLabels: vi.fn().mockResolvedValue([]),
+  onClose: vi.fn(),
+  onCreated: vi.fn(),
+}));
+
+vi.mock("@octo/base", () => ({
+  useI18n: () => ({
+    t: (key: string) => key,
+  }),
+}));
+
+vi.mock("@douyinfe/semi-ui", () => {
+  const Modal = ({ visible, children }: any) =>
+    visible ? <div>{children}</div> : null;
+  const Select: any = ({ children }: any) => <div>{children}</div>;
+  Select.Option = ({ children }: any) => <>{children}</>;
+  return {
+    Modal,
+    Select,
+    Avatar: ({ children }: any) => <span>{children}</span>,
+    Toast: { error: vi.fn() },
+  };
+});
+
+vi.mock("../../api/issueApi", () => ({
+  createIssue: (...args: any[]) => hoisted.createIssue(...args),
+  listAssigneeCandidates: (...args: any[]) =>
+    hoisted.listAssigneeCandidates(...args),
+}));
+
+vi.mock("../../api/directory", () => ({
+  listProjectOptions: (...args: any[]) => hoisted.listProjectOptions(...args),
+}));
+
+vi.mock("../../api/attachmentApi", () => ({
+  uploadAttachment: (...args: any[]) => hoisted.uploadAttachment(...args),
+}));
+
+vi.mock("../../api/labelApi", () => ({
+  attachLabel: (...args: any[]) => hoisted.attachLabel(...args),
+  listLabels: (...args: any[]) => hoisted.listLabels(...args),
+}));
+
+vi.mock("../../api/http", () => ({
+  currentWorkspaceName: () => "workspace",
+}));
+
+vi.mock("../AssigneePicker", () => ({
+  default: () => <div data-testid="assignee-picker" />,
+}));
+
+vi.mock("../LoopPropertyPill", () => ({
+  default: () => <button type="button" data-testid="property-pill" />,
+}));
+
+import CreateIssueModal from "../CreateIssueModal";
+
+let container: HTMLDivElement;
+
+beforeEach(() => {
+  hoisted.createIssue.mockReset().mockResolvedValue({ id: "issue-1" });
+  hoisted.listAssigneeCandidates.mockReset().mockResolvedValue([]);
+  hoisted.listProjectOptions.mockReset().mockResolvedValue([]);
+  hoisted.listLabels.mockReset().mockResolvedValue([]);
+  hoisted.onClose.mockReset();
+  hoisted.onCreated.mockReset();
+  container = document.createElement("div");
+  document.body.appendChild(container);
+});
+
+afterEach(() => {
+  act(() => {
+    ReactDOM.unmountComponentAtNode(container);
+  });
+  container.remove();
+});
+
+const flush = async (): Promise<void> => {
+  await act(async () => {
+    await Promise.resolve();
+    await Promise.resolve();
+  });
+};
+
+const render = async (): Promise<void> => {
+  act(() => {
+    ReactDOM.render(
+      <CreateIssueModal
+        visible
+        onClose={hoisted.onClose}
+        onCreated={hoisted.onCreated}
+      />,
+      container
+    );
+  });
+  await flush();
+};
+
+const titleInput = (): HTMLInputElement =>
+  container.querySelector<HTMLInputElement>(".loop-ci__title")!;
+
+const fillTitle = (): void => {
+  act(() => {
+    Simulate.change(titleInput(), { target: { value: "测试任务" } } as any);
+  });
+};
+
+const pressEnter = (isComposing: boolean): void => {
+  act(() => {
+    titleInput().dispatchEvent(
+      new KeyboardEvent("keydown", {
+        key: "Enter",
+        isComposing,
+        bubbles: true,
+        cancelable: true,
+      })
+    );
+  });
+};
+
+describe("CreateIssueModal title input Enter / IME composition", () => {
+  it("does NOT submit when Enter fires during IME composition", async () => {
+    await render();
+    fillTitle();
+
+    pressEnter(true);
+    await flush();
+
+    expect(hoisted.createIssue).not.toHaveBeenCalled();
+  });
+
+  it("submits when Enter fires outside composition", async () => {
+    await render();
+    fillTitle();
+
+    pressEnter(false);
+    await flush();
+
+    expect(hoisted.createIssue).toHaveBeenCalledTimes(1);
+  });
+});


lml2468 (approved):
## Review: #1106 — WS-25: fix IME Enter handling in create issue modal

**意图**:创建 issue 弹窗标题输入框,在中文/拼音输入法**组字（IME composition）**中按 Enter 选候选词时,不应误触发提交。锚定 head `4cf92180`(rev-parse 确认 checkout==live head)/ `mergeable=true`。2 文件 +161/−2。

**结论:APPROVE ✅**

### Gate 1 — 规格符合度 ✅
标题输入 `onKeyDown` 的 Enter 提交加 IME 组字守卫;+ 新增单测。Select.Option 那处是纯格式化(多行)无行为变化。最小改动,无 over-build。

### Gate 2 — 代码质量 ✅
- **标准 IME 修法**:`if (e.key === "Enter")` → `if (e.key === "Enter" && !e.nativeEvent.isComposing) submit()`(:285)。`nativeEvent.isComposing` 在组字期间为 true,故选候选词的 Enter 不再提交;组字结束后的 Enter 正常提交。用的是现代标准属性,正确。
- **作用域正确**:全文件仅此一处 Enter 提交(标题输入);下方 AutoGrowTextarea 不在 Enter 提交(允许换行),未误改。
- 无安全/XSS 面(纯 keydown 守卫)。
- **测试到位**:新测两分支 —— 组字中 Enter(`isComposing:true`)→ **不调用 createIssue**;非组字 Enter(`isComposing:false`)→ **createIssue 调用一次**。精确锁定修复。

**构建 / 测试(实跑 @ `4cf92180`)**:`pnpm --filter @octo/web build` ✅ 3.66s;`dmloop CreateIssueModal.test` **2 tests 全绿**。

---

### 🟡 非阻断
- `isComposing` 是现代标准属性,覆盖当前 Chrome/Safari/Firefox;仅在组字期只发 `keyCode 229`(不带 `isComposing`)的极老浏览器/IME 上不生效 —— 属可接受的边缘,现方案是业界标准修法。

---

标准 `!isComposing` 守卫加在唯一的 Enter 提交路径(标题输入)、textarea 换行不受影响、两分支各有测试、构建绿。**可合并 —— APPROVE。**

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review: octo-web PR#1106 — WS-25: fix IME Enter handling in create issue modal
**Head SHA:** `4cf9218014ff92da2279d5e4bf5f1b5270a33d7e`
**Base:** `7561e8ef` (main)
**Scope:** 1 commit, 2 files changed (+161 / -2)

---

## 1. Summary

This PR fixes a single bug: pressing Enter during IME composition (e.g. typing Chinese/Japanese text) in the Create Issue modal's title field would incorrectly trigger form submission. The fix adds `!e.nativeEvent.isComposing` to the existing Enter key guard. A new test file validates both the positive (Enter submits) and negative (Enter during IME does not submit) paths.

**Files changed:**
- `packages/dmloop/src/ui/CreateIssueModal.tsx` — 1 line logic change + cosmetic reformatting
- `packages/dmloop/src/ui/__tests__/CreateIssueModal.test.tsx` — new test file (155 lines)

---

## 2. Verification

### ✅ Core fix — `CreateIssueModal.tsx:285`

```tsx
// Before
if (e.key === "Enter") submit();
// After
if (e.key === "Enter" && !e.nativeEvent.isComposing) submit();
```

**Correct.** `KeyboardEvent.isComposing` is the standard DOM property (true when an IME composition session is active). This is the minimal, idiomatic guard. The existing `InlineEdit.tsx` in the same package uses a different approach (`composing` ref via `onCompositionStart`/`onCompositionEnd` events) — both are valid; `isComposing` is simpler and doesn't require extra state.

### ✅ Test coverage — `CreateIssueModal.test.tsx`

- **Negative test** (L136–143): Dispatches `KeyboardEvent("keydown", { key: "Enter", isComposing: true })` → verifies `createIssue` was NOT called. ✅
- **Positive test** (L146–153): Dispatches with `isComposing: false` → verifies `createIssue` was called once. ✅
- **Data flow verified:** `pressEnter` uses `dispatchEvent(new KeyboardEvent(...))` on the real DOM node; React's event delegation picks it up and wraps it in a synthetic event where `e.nativeEvent` IS the original `KeyboardEvent`, so `e.nativeEvent.isComposing` correctly reflects the constructor parameter.
- **Mocks** are thorough: `@octo/base`, `@douyinfe/semi-ui`, all API modules, child components — properly isolated with `vi.hoisted()` for spy access.

### ✅ Cosmetic change — `CreateIssueModal.tsx:349–355`

`<Select.Option>` reformatted to multi-line. No logic change. Pure whitespace/prettier.

---

## 3. Findings

### No P0/P1 findings.

### P2 — Pre-existing: Other Enter handlers in dmloop lack IME guard

The following `onKeyDown` handlers in the dmloop package check `e.key === "Enter"` without any IME composition guard. These are **pre-existing** (not introduced or modified by this PR) and therefore non-blocking, but represent the same bug class this PR fixes:

| File | Line | Context |
|------|------|---------|
| `SkillDetailPage.tsx` | 212 | `setEditingName(false)` on Enter — closes edit during IME |
| `SkillDetailPage.tsx` | 260 | `submitNewFile()` on Enter |
| `ModelPicker.tsx` | 54 | `canCreate` submit on Enter |
| `InlineEdit.tsx` | 103 | Has own IME guard via `composing` ref ✅ |
| `suggestionMenu.ts` | 201 | Menu selection on Enter |

**Recommendation:** Consider a follow-up pass to add `!e.nativeEvent.isComposing` (or the `composing` ref pattern already used in `InlineEdit.tsx`) to the remaining text-input Enter handlers. `SkillDetailPage.tsx:212` is the highest-risk candidate since it closes an edit mode on Enter during IME.

### P2 — Pre-existing: submit() `finally` resets submitting on early return

`CreateIssueModal.tsx:165–220` — The `submit()` function has early `return` paths (e.g. L183 attachment upload failure) but the `finally` block at L218 always resets `submitting = false`. This means a failed submit still clears the `submitting` guard, allowing immediate re-submission. Not introduced by this PR.

---

## 4. Data Flow Trace

| Consumer | Data | Source | Verified? |
|----------|------|--------|-----------|
| `submit()` guard | `title.trim()` | `useState("")` → `onChange` → `setTitle` | ✅ Test fills title via `Simulate.change` |
| `submit()` guard | `submitting` | `useState(false)` → set `true` on entry | ✅ Fresh state per test |
| IME guard | `e.nativeEvent.isComposing` | `KeyboardEvent` constructor param | ✅ Dispatched with correct value |
| `createIssue` call | args object | constructed in `submit()` from state | ✅ Mock captures call |

---

## 5. Blind-Spot Checklist (C1–C6)

| Check | Result | Evidence |
|-------|--------|----------|
| **C1 — Dual-path parity** | N/A | Single additive guard, no symmetric removal path |
| **C2 — Control-flow ordering / nesting** | Clear | `onKeyDown` handler on title `<input>` is not shared/reused. `AutoGrowTextarea` has no Enter handler |
| **C3 — Authorization boundary** | N/A | Frontend UI, no auth changes |
| **C4 — Auth lifecycle / container-member cascade** | N/A | No auth changes |
| **C5 — Build/runtime path correctness** | Clear | Test uses jsdom + vi.mock, consistent mock paths. No build/config changes |
| **C6 — Governance/docs self-consistency** | N/A | No docs/governance changes |

---

## 6. Cross-round Blocker Re-review (R6)

N/A — First review round for this PR.

---

## 7. Extra Observations

- **Consistency note:** The existing `InlineEdit.tsx` in the same package uses `compositionstart`/`compositionend` events with a ref-based flag. This PR's approach (`isComposing` on the native event) is simpler and equally correct. If the team prefers one pattern, consider documenting the preferred approach.
- **Browser support:** `KeyboardEvent.isComposing` is supported in all modern browsers (Chrome 60+, Firefox 44+, Safari 10.1+). No polyfill needed.

---

## Verdict

**[Octo-Q] verdict: APPROVE** — The fix is minimal, correct, and well-tested. No P0/P1 findings. The IME guard uses the standard `KeyboardEvent.isComposing` property and the test validates both the positive and negative paths. The P2 observations are all pre-existing issues not in this PR's scope.

Jerry-Xin (approved):
Correct, minimal IME Enter-handling fix with a meaningful regression test — LGTM.

## Summary
The create-issue title input previously submitted on any `Enter` keydown, including the `Enter` that CJK/IME users press to select a composition candidate. The fix guards the shortcut with `!e.nativeEvent.isComposing`, which is the right primitive (reads the native composition state directly, no compositionstart/end ref plumbing or listeners to leak).

## Verification
- Guard confirmed at `packages/dmloop/src/ui/CreateIssueModal.tsx:285`: `if (e.key === "Enter" && !e.nativeEvent.isComposing) submit();`
- Ran the new test at head: 2/2 pass.
- Mutation check: reverting the guard back to `if (e.key === "Enter") submit();` makes the "does NOT submit during IME composition" test fail (createIssue called once). This proves the test genuinely exercises the compose guard rather than only asserting a handler fired — it is not a false-green.
- Test dispatches a real native `KeyboardEvent` with `isComposing: true/false` that bubbles to the React 17 root delegated listener, so `e.nativeEvent.isComposing` is exercised end-to-end.

## Adjacent branches (checked, all clean)
- Only one key handler exists: the title `<input> onKeyDown`. The description field (`AutoGrowTextarea`) has no Enter-submit handler, so single-line-input Enter semantics are correct and no Shift/Ctrl/Cmd+Enter newline behavior was affected or broken.
- `submit()` already guards `!title.trim() || submitting`, so double-submit / empty-title are covered.
- The OK path (`disabled={!title.trim()}`) is independent of the Enter path; unchanged.
- No native compositionstart/end listeners were added — no add/removeEventListener symmetry or useEffect-cleanup concern introduced.

## Non-blocking notes
- The +161 diff is mostly the new 155-line test file. The only functional change is one line; the remaining change is a cosmetic prettier reformat of an unrelated `Select.Option` block in the same file. Harmless, but unrelated formatting churn in a targeted fix slightly muddies the diff — fine to keep.

Blocking items: none.

yujiawei (approved):
# Code Review — PR #1106 (octo-web)

Reviewed at head SHA `4cf9218014ff92da2279d5e4bf5f1b5270a33d7e` (single commit, base `main`).

## What I actually ran

I checked the branch out and exercised it rather than reading the diff alone:

| Check | Result |
|---|---|
| `vitest run src/ui/__tests__/CreateIssueModal.test.tsx` | 2/2 pass |
| Mutation check — reverted the guard to `if (e.key === "Enter") submit();` | composing test **fails** as designed (`createIssue` called once with `title: "测试任务"`) |
| Full `@octo/loop` suite | 21 files / 98 tests pass, no regressions |
| `tsc -p packages/dmloop/tsconfig.json --noEmit` | no new errors in `CreateIssueModal.tsx`; **2 new errors** in the new test file (see P2-3) |
| `prettier --check` on both changed files | clean |

The mutation check is the important one: the regression test is genuinely load-bearing, not a tautology that would pass either way.

## 1. Spec compliance

Stated scope (PR body + WS-25): guard the create-issue title Enter shortcut while an IME composition is active, and add a jsdom regression test.

**Spec: ✅**

- **Missing:** none. `packages/dmloop/src/ui/CreateIssueModal.tsx:285` guards exactly the handler described, and the regression test exists and is verified effective.
- **Extra:** one formatting-only hunk at `CreateIssueModal.tsx:349-356` (`Select.Option` reflowed onto multiple lines). Strictly outside WS-25, but it is a `prettier --write` byproduct that repairs a pre-existing violation — I verified the pre-PR file fails `prettier --check` and the post-PR file passes, because the original line was 81 columns against a `printWidth` of 80. No behavior change, no new API/flag/field. Not counted against the spec gate; see the nit below.
- **Deviation:** none. WS-25 did not prescribe a mechanism.

On mechanism choice, `e.nativeEvent.isComposing` is the correct access point. React's `SyntheticKeyboardEvent` does **not** expose `isComposing` — I confirmed this in the resolved `react-dom` build: `KeyboardEventInterface` (`react-dom/cjs/react-dom.development.js:6530-6560`) carries `key`/`code`/`location`/modifiers/`repeat`/`locale`/`charCode`/`keyCode`/`which` and nothing else. React's internal `isComposing` (same file, line 6785) is a module-level variable in the `beforeInput` plugin, never surfaced on the event object. So `!e.isComposing` would evaluate `!undefined === true` and silently reintroduce the bug. Reaching for `nativeEvent` here is right, not a smell.

## 2. Code quality

**Quality: Approved** — no P0/P1. The findings below are advisory.

### P2-1 — Residual IME gap on WebKit; guard is correct but not exhaustive
`packages/dmloop/src/ui/CreateIssueModal.tsx:285`
```tsx
if (e.key === "Enter" && !e.nativeEvent.isComposing) submit();
```
On Chromium and Gecko the Enter that commits a candidate reports `isComposing === true`, so this fixes those. WebKit has a long-standing ordering quirk where `compositionend` can precede the committing `keydown`, leaving `isComposing === false` on the event that leaks through. `apps/web/package.json:130-136` sets the production browserslist to `>0.2%, not dead`, so Safari is a shipped target and that path is reachable.

I am deliberately **not** treating this as blocking: the patch cannot make any browser worse than the pre-patch state (previously *every* Enter submitted, composing or not), so it is a strict improvement with a narrower residual surface. It is a follow-up, not a merge blocker. Concretely, when you do the follow-up: also bail on `e.nativeEvent.keyCode === 229`, or keep a short post-`compositionend` grace window, and confirm with manual QA on macOS Safari using a Pinyin IME — that specific combination is not something the current unit test can prove either way.

### P2-2 — The new regression test never runs in CI
`.github/workflows/ci.yml:87-95` is the only place tests execute, and it runs exactly two things:
```yaml
      - name: Skill Market tests
        run: pnpm --filter @dmwork/skillmarket test
      - name: Unified import/export unit tests
        run: >-
          pnpm --filter @octo/docs exec vitest run
          src/editor/EditorShell.test.tsx
          ...
```
There is no `@octo/loop` test step, and root `pnpm test` (`turbo run test`) is not invoked by the workflow. So this test protects nothing after merge — the next refactor of that handler will not be caught. The PR body's Tests section implies coverage the pipeline will not enforce. Worth a one-line CI addition (`pnpm --filter @octo/loop test`), here or in an immediate follow-up.

### P2-3 — Test file introduces 2 new TypeScript errors
```
src/ui/__tests__/CreateIssueModal.test.tsx(5,22): error TS7016: Could not find a declaration file for module 'react-dom'
src/ui/__tests__/CreateIssueModal.test.tsx(6,31): error TS7016: Could not find a declaration file for module 'react-dom/test-utils'
```
`packages/dmloop/package.json:25` declares `@types/react` but not `@types/react-dom`, so both new imports are implicitly `any`. Not build-breaking (the workflow has no typecheck step, and Vite/esbuild does not type-check), but these are two errors that did not exist before this diff and they will show up in every contributor's editor. Fix is one line: add `"@types/react-dom": "^17.0.25"` to `packages/dmloop` devDependencies, matching the pinned `react-dom@17.0.2`.

### P2-4 — Test harness runs React 17 while production ships React 18
`packages/dmloop/package.json:24-27` pins `react`/`react-dom` at `^17.0.2` as devDependencies (resolved: 17.0.2), but `apps/web/package.json:89,119` ships React 18.3.1. This test is the first jsdom component test in the package (`ReactDOM.render` + `react-dom/test-utils` `Simulate` appear nowhere else in `packages/dmloop/src`), so it establishes a pattern that validates behavior under React 17's event delegation rather than the React 18 root-scoped system that actually ships.

For *this* assertion the result still generalizes — `nativeEvent` is the raw DOM event in both versions — so I am not treating it as a correctness problem. But as the seed of a component-test convention it is the wrong footing: `ReactDOM.render` and `Simulate` are removed in React 19, and the repo root already has `@testing-library/react` and `@testing-library/jest-dom` available. Prefer those, and align the dev React version with 18, before this pattern gets copied.

### P2-5 — Same bug class remains in four sibling inputs (out of scope, for follow-up)
Not asking for these in this PR — WS-25 is scoped to the create-issue modal — but the same unguarded `Enter` on a text input exists at:

- `packages/dmloop/src/panel/ModelPicker.tsx:53-58` — `Enter` → `commit(trimmed)`
- `packages/dmloop/src/panel/SkillDetailPage.tsx:212` — `Enter` → `setEditingName(false)`, and `onBlur` commits
- `packages/dmloop/src/panel/SkillDetailPage.tsx:260` — `Enter` → `submitNewFile()`
- `packages/dmloop/src/panel/AutopilotDetailPage.tsx:54-55` — `Enter` → `e.currentTarget.blur()`, and `onBlur` → `commit()`

The last two are the same user-visible failure as WS-25: type a CJK name, press Enter to pick a candidate, the editor closes and commits the half-typed value.

Note also `packages/dmloop/src/ui/InlineEdit.tsx:95-108`, which already solves this with an explicit `composing` ref driven by `onCompositionStart`/`onCompositionEnd`. So the package now has two mechanisms for one problem. A tiny shared helper (`isImeCommitKey(e)`) applied across all five call sites would be the right consolidation — and it would give P2-1's WebKit fallback a single place to live.

### nit — scope of the `Select.Option` reflow
`CreateIssueModal.tsx:349-356` is unrelated to the fix and adds blame noise to an otherwise one-line correctness change. Since it repairs a genuine pre-existing Prettier violation I would not ask you to revert it; just flagging that a formatter sweep is easier to review as its own commit.

## 3. Overall verdict

**APPROVE.**

The fix is correct, minimal, uses the right API for React 17/18, and ships with a regression test I confirmed genuinely fails without it. Nothing here needs to change before merge. The two items I would most like to see land soon after are P2-2 (wire `@octo/loop` into CI, otherwise the test is decorative) and P2-1 (the WebKit fallback plus manual Safari QA).

## 4. Coverage note — what this review could not establish

- Real CJK IME behavior on macOS Safari and iOS WebKit. The unit test injects `isComposing` directly, so it cannot distinguish "guard works everywhere" from "guard works where `isComposing` is already reliable." This needs a human on a real browser.
- Whether WS-25's original report came from Chromium or WebKit — that would settle whether P2-1 is a live user-facing gap or theoretical.
- Runtime behavior of the label `Select` path: the test renders with `listLabels` resolving to `[]`, so the `labels.length > 0` branch containing the reformatted `Select.Option` is never mounted. The reflow is whitespace-only, so this is a gap in the test's reach rather than a risk in the diff.
- Electron packaging is untested here; it is Chromium, so it inherits the working case.

yujiawei:
Follow-up evidence on the P2-1 point from my review (the residual WebKit gap) — not a change request on this PR, which stands approved. I went looking for prior art and found that the repo's own dependency tree already documents this exact behavior and ships a fix for it, which makes the follow-up much more concrete than "add a keyCode 229 check".

`prosemirror-view@1.41.8` (pulled in via `@tiptap/pm`) gates all key handling behind a composition check — `dist/index.js:3144-3145`:

```js
editHandlers.keydown = (view, _event) => {
    let event = _event;
    view.input.shiftKey = event.keyCode == 16 || event.shiftKey;
    if (inOrNearComposition(view, event))
        return;
```

And `inOrNearComposition` (`dist/index.js:3450-3468`) spells out the Safari case verbatim:

```js
function inOrNearComposition(view, event) {
    if (view.composing)
        return true;
    // On Japanese input method editors (IMEs), the Enter key is used to confirm character
    // selection. On Safari, when Enter is pressed, compositionend and keydown events are
    // emitted. The keydown event triggers newline insertion, which we don't want.
    // We only ignore it once, as pressing Enter a second time *should* insert a newline.
    // Furthermore, the keydown event timestamp must be close to the compositionEndedAt timestamp.
    if (safari && Math.abs(event.timeStamp - view.input.compositionEndedAt) < 500) {
        view.input.compositionEndedAt = -2e8;
        return true;
    }
    return false;
}
```

Three things worth lifting from it for the follow-up:

1. **The Safari gap is real, not theoretical.** `isComposing` being supported since Safari 10.1 is about *property availability*; the problem is Safari's event *ordering* — `compositionend` lands before the committing `keydown`, so `isComposing` is already `false` on the event you're inspecting. Property support does not make the guard reliable there.
2. **The grace window is the mechanism, not `keyCode === 229`.** Track `compositionEndedAt` in an `onCompositionEnd` handler and ignore an `Enter` keydown whose `timeStamp` is within ~500ms of it.
3. **Consume the suppression exactly once** (the `compositionEndedAt = -2e8` sentinel), and gate it on Safari. Otherwise a deliberate second Enter stops working, and a mouse-confirmed candidate followed by a real Enter gets swallowed.

So a shared helper covering `nativeEvent.isComposing` plus that Safari grace window would close this for the title input and the sibling handlers I listed (`ModelPicker.tsx:53-58`, `SkillDetailPage.tsx:212,260`, `AutopilotDetailPage.tsx:54-55`) in one pass.

One correction to my own review while I'm here: `suggestionMenu.ts:201` should **not** be on that follow-up list. Its `onKeyDown` is invoked through ProseMirror's `handleKeyDown`, which is behind the `inOrNearComposition` early-return shown above — that call site is already protected upstream.

lejunliu:
## 用户复现反馈（2026-09-03）

- 反馈人：user:母世 (uid:f306c2785eea4a27a0d5048aabe20e5a)
- 来源：Octo Bug 反馈群 (gid:30ac52c2ccba4f8d9f152311f1f785b5)
- 环境：Web 1.0.12；Chrome 152.0.7977.65
- 现象：新建 Loop 任务时，中文输入法仍处于组词/候选阶段，按 Enter 确认候选词却直接提交并创建了尚未描述完成的任务。
- 期望：取消标题输入框 Enter 创建快捷键，仅允许点击「创建」按钮提交；至少 IME composition 期间绝不能触发创建。
- 用户希望修复后收到闭环通知。

🔑 闭环关键词：Loop, 新建任务, 标题输入框, 中文输入法, 候选词, 回车, Enter, IME, composition, 误创建, 自动提交, create issue modal

PR #1107, opened 2026-07-27
fix(channel-setting): refresh inline save and return states

Candice00711:
## Summary

- track inline-edit intent so clearing and retyping the original group nickname enables Save immediately
- preserve existing clear, cancel, failed-save retry, external refresh, and successful-save behavior
- remove the visible exit displacement when returning from group avatar and group QR code pages while preserving WKViewQueue cleanup

## Scope

- limited to `ChannelSettingInlineEditRow`, its regression tests, and ChannelSetting-scoped transition CSS
- no changes to APIs, bridge actions, permissions, WKApp, WKSDK, datasource, Chat, Conversation, Messages, or shared route logic

## Validation

- ChannelSetting focused Vitest: 8 files, 74 tests passed
- `pnpm i18n:check`: passed
- Prettier: passed
- Stylelint: 0 errors; 6 pre-existing warnings outside the changed rules
- `git diff --check`: passed
- Web production build: passed (14,970 modules transformed)
- manual verification passed for nickname clear/retype/save and avatar/QR-code return behavior

### DIFF
--- a/packages/dmworkbase/src/Components/ChannelSetting/index.css
+++ b/packages/dmworkbase/src/Components/ChannelSetting/index.css
@@ -87,6 +87,26 @@
     background: var(--wk-bg-surface);
 }
 
+/*
+ * 头像和二维码子页返回时仍保留 animationend，让 WKViewQueue 正常清理栈顶；
+ * 只取消可见位移，避免标题先切回父页、内容仍横向退场造成的闪动。
+ * 该规则不改变 RoutePage/WKViewQueue 的退栈和数据刷新逻辑。
+ */
+.wk-viewqueue-view-out:has(.wk-channelavatar),
+.wk-viewqueue-view-out:has(.wk-channelqrcode) {
+    background: var(--wk-bg-surface);
+    animation-name: wk-channelsetting-child-exit;
+    animation-duration: var(--wk-dur-fast);
+    animation-timing-function: var(--wk-ease);
+}
+
+@keyframes wk-channelsetting-child-exit {
+    from,
+    to {
+        transform: translate3d(0, 0, 0);
+    }
+}
+
 /* 仅作用于 ChannelSetting，避免影响其他 Sections/ListItem 调用方。 */
 .wk-channelsetting-content > .wk-sections {
     display: flex;
--- a/packages/dmworkbase/src/ui/ChannelSettingRows/__tests__/ChannelSettingInlineEditRow.test.tsx
+++ b/packages/dmworkbase/src/ui/ChannelSettingRows/__tests__/ChannelSettingInlineEditRow.test.tsx
@@ -88,6 +88,29 @@ describe("ChannelSettingInlineEditRow", () => {
     expect(onSave).toHaveBeenCalledWith("");
   });
 
+  it("enables saving after clearing and retyping the original nickname", () => {
+    render(
+      <ChannelSettingInlineEditRow
+        title="My nickname"
+        value="Alice"
+        allowEmpty
+        onSave={vi.fn(() => Promise.resolve())}
+      />
+    );
+
+    fireEvent.click(screen.getByRole("button", { name: "My nickname" }));
+    fireEvent.mouseDown(
+      screen.getByRole("button", { name: "My nickname-base.common.clear" })
+    );
+
+    const input = screen.getByDisplayValue("");
+    fireEvent.change(input, { target: { value: "Alice" } });
+
+    expect(
+      screen.getByRole("button", { name: "base.common.save" })
+    ).toBeEnabled();
+  });
+
   it("keeps the draft open when saving fails", async () => {
     const onSave = vi.fn(() => Promise.resolve(false));
 
@@ -109,6 +132,9 @@ describe("ChannelSettingInlineEditRow", () => {
     expect(screen.getByDisplayValue("Unsaved draft")).toBe(input);
     expect(
       screen.getByRole("button", { name: "base.common.cancel" })
+    ).toBeEnabled();
+    expect(
+      screen.getByRole("button", { name: "base.common.save" })
     ).toBeEnabled();
   });
 
@@ -167,6 +193,9 @@ describe("ChannelSettingInlineEditRow", () => {
 
     fireEvent.click(screen.getByRole("button", { name: "Group name" }));
     expect(screen.getByDisplayValue("Remote name")).toBeInTheDocument();
+    expect(
+      screen.getByRole("button", { name: "base.common.save" })
+    ).toBeDisabled();
   });
 
   it("keeps a successful save visible until the external value catches up", async () => {
--- a/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
+++ b/packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx
@@ -116,6 +116,7 @@ export function ChannelSettingInlineEditRow({
   const [editing, setEditing] = useState(false);
   const [currentValue, setCurrentValue] = useState(value);
   const [draft, setDraft] = useState(value);
+  const [dirty, setDirty] = useState(false);
   const [saving, setSaving] = useState(false);
   const editingRef = useRef(false);
 
@@ -132,12 +133,12 @@ export function ChannelSettingInlineEditRow({
 
   const exceeded = maxCount !== undefined && draft.length > maxCount;
   const emptyInvalid = !allowEmpty && draft.trim().length === 0;
-  const unchanged = draft === currentValue;
-  const saveDisabled = saving || exceeded || emptyInvalid || unchanged;
+  const saveDisabled = saving || exceeded || emptyInvalid || !dirty;
 
   const startEdit = () => {
     if (onStartEdit?.() === false) return;
     setDraft(currentValue);
+    setDirty(false);
     setEditing(true);
   };
 
@@ -156,7 +157,10 @@ export function ChannelSettingInlineEditRow({
     value: draft,
     placeholder,
     disabled: saving,
-    onChange: (next: string) => setDraft(next),
+    onChange: (next: string) => {
+      setDraft(next);
+      setDirty(true);
+    },
   };
 
   return (
@@ -166,7 +170,10 @@ export function ChannelSettingInlineEditRow({
         <TextArea
           {...inputProps}
           showClear
-          onClear={() => setDraft("")}
+          onClear={() => {
+            setDraft("");
+            setDirty(true);
+          }}
           autosize={{ minRows: 2, maxRows: 6 }}
         />
       ) : (
@@ -182,6 +189,7 @@ export function ChannelSettingInlineEditRow({
                   event.preventDefault();
                   event.stopPropagation();
                   setDraft("");
+                  setDirty(true);
                 }}
               >
                 <IconClear />
@@ -205,6 +213,7 @@ export function ChannelSettingInlineEditRow({
           disabled={saving}
           onClick={() => {
             setDraft(currentValue);
+            setDirty(false);
             setEditing(false);
           }}
         >
@@ -220,6 +229,7 @@ export function ChannelSettingInlineEditRow({
               const saved = await onSave(draft);
               if (saved !== false) {
                 setCurrentValue(draft);
+                setDirty(false);
                 setEditing(false);
               }
             } catch {


lml2468 (approved):
## Review: #1107 — fix(channel-setting): refresh inline save and return states

**意图**:①内联编辑改用 `dirty` 意图标记,使「清空后重打原昵称」也能立即启用保存;保留 clear/cancel/失败重试/外部刷新/成功保存行为。②去掉从群头像/群二维码子页返回时的可见位移,同时保留 WKViewQueue 清理。锚定 head `e8b84e0f`(rev-parse 确认 checkout==live head)/ `mergeable=true`。3 文件 +63/−4。#1075/#1085 同组件续。

**结论:APPROVE ✅**

### Gate 1 — 规格符合度 ✅
限定 `ChannelSettingInlineEditRow` + 其测试 + ChannelSetting 作用域 CSS;明确不改 API/bridge/权限/WKApp/WKSDK/datasource/Chat/Conversation。与描述一致,无 over-build。

### Gate 2 — 代码质量 ✅
- **`dirty` 意图标记正确**:由 `unchanged = draft===currentValue` 改为 `saveDisabled = ... || !dirty`;`dirty` 在 onChange/onClear/清除按钮置 true,在 startEdit/cancel/**成功保存**置 false,**失败保存不重置**(留 true 供重试)。修好了「清空后重打原值 → 旧逻辑 draft===currentValue → 保存被禁」的问题。
- **承 #1075 无回归**:`editingRef` value-sync 守卫(`if (!editingRef.current) setDraft(value)`,:129)未被触碰,编辑中远端刷新不覆盖草稿的修复保留;外部刷新后重进编辑 startEdit 置 dirty=false → 保存禁用(有测试)。
- **CSS 收敛**:`.wk-viewqueue-view-out:has(.wk-channelavatar/.wk-channelqrcode)` 用 no-op keyframe(from/to translate3d(0,0,0))**保留 animationend**(WKViewQueue 退栈清理照跑),只取消可见位移;全用语义 token(`--wk-dur-fast`/`--wk-ease`/`--wk-bg-surface` 均已定义);注释明确不改退栈/刷新逻辑。
- 无 XSS(React 文本 / CSS)。

**构建 / 测试 / i18n(实跑 @ `e8b84e0f`)**:`pnpm --filter @octo/web build` ✅ 3.88s;`pnpm i18n:check` ✅;`ChannelSettingInlineEditRow.test` **10 tests 全绿**(新增:清空+重打原值启用保存、失败保存保持启用、外部刷新禁用保存)。

---

### 🟡 非阻断
- `dirty` 语义下,用户输入一字符再删回原值也会使保存启用(draft===currentValue 但 dirty=true)→ 可能发一次值未变的冗余保存;后端幂等、无害,是「clear+retype 原值需可保存」需求的合理取舍。
- `:has()` 需现代浏览器(Electron/Chromium 当前支持)。

---

`dirty` 意图标记修好清空-重打原值的保存启用、承 #1075 的 editingRef 守卫无回归、CSS 作用域收敛且保留 WKViewQueue 清理生命周期、语义 token 齐全,构建/i18n/10 测试通过。**可合并 —— APPROVE。**

Jerry-Xin (approved):
Approve — clean, well-scoped fix for the inline-edit save/return state bugs, with regression tests that assert the actual resulting states.

## Summary
Introduces an explicit `dirty` intent flag to `ChannelSettingInlineEditRow` so Save enablement tracks edit intent instead of value equality, and adds a scoped no-op exit keyframe so returning from the avatar/QR child pages no longer shows a visible horizontal displacement while still preserving WKViewQueue cleanup.

## State-transition audit (all paths verified)
- **Enter edit** (`startEdit`): resets `draft = currentValue`, `dirty = false`, `editing = true` → Save correctly disabled until a real edit. Good.
- **Change / clear (input) / clear (textarea) / clear (suffix button)**: all set `dirty = true` → Save enabled. The original bug (clear + retype the same value left Save disabled because `draft === currentValue`) is fixed because enablement now keys off intent, not equality.
- **Cancel**: reverts `draft = currentValue`, `dirty = false`, `editing = false` → field returns to the persisted value, no stale draft. Good.
- **Save success** (`saved !== false`): `currentValue = draft`, `dirty = false`, `editing = false`, and `saving` cleared in `finally`. Optimistic display shows the new local value until the external `value` catches up, then the effect resyncs. Good.
- **Save failure / rejection**: editor stays open, `saving` cleared in `finally` (no stuck "saving" row), `dirty` stays true so retry is immediately available. Covered by dedicated tests.
- **External `value` update while editing**: effect updates `currentValue` but preserves the in-progress `draft` via `editingRef`; `dirty` is untouched (correct — unsaved edits still pending).
- **Re-enter after save**: `startEdit` seeds `draft` from the new `currentValue` with `dirty = false` → Save disabled until changed. Covered.

## Flag ownership / races
- `saving` is owned solely by this component and reset in `finally` across all three async outcomes (success, `false`, throw); no exit path leaves the row stuck.
- No event listeners are attached in this component, so no cleanup symmetry gap. A late `setSaving(false)` after unmount would only be a benign no-op state update; there is no cross-row identity change here that could misroute the update.
- The CSS change is a no-op keyframe (`translate3d(0,0,0)` from/to) scoped via `:has(.wk-channelavatar/.wk-channelqrcode)`; it cancels only the visible displacement and does not alter RoutePage/WKViewQueue retack or data-refresh logic.

## Validation performed
- Focused Vitest at head: `ChannelSettingInlineEditRow.test.tsx` — 10/10 passed.
- Tests assert real resulting states (Save enabled after failed save, Save disabled after re-enter with reverted value, saved value shown until external catches up), not merely that a handler fired.

## Non-blocking notes
- The `finally { setSaving(false) }` can run after unmount if the component is torn down mid-save; harmless today, but if a lint rule for post-unmount state updates is ever enabled this is the one spot to guard.

No blocking items.

mochashanyao (approved):
[Octo-Q · automated review]

---

# Code Review — PR #1107 (octo-web)
**Head SHA:** `e8b84e0f6947da43e868592d7205bc1480ae4f15`

## Summary

This PR has two major themes: (1) migrating channel setting fields (nickname, thread name, group name, group notice, remark) from a modal-based `inputEditPush` pattern to a new inline-edit component `ChannelSettingInlineEditRow`, and (2) removing the entire `dmworktodo` package plus all matter-service integrations (proxy routes, env vars, event bus handlers, UI panels, bridge types). The inline-edit migration is well-tested with edge cases (clear, failed save, draft preservation across rerenders). The matter removal is thorough and consistent across nginx config, Vite proxy, Docker entrypoint, i18n, and runtime code.

## Verification

- ✅ **No orphaned matter references** — grep for `showMatterPanel`, `toggle-matter-panel`, `MATTER_API_URL`, `/matter/api/v1`, `matterBridge`, `dmworktodo` returns zero hits in remaining source.
- ✅ **Deployment config consistent** — `docker-entrypoint.sh` removes `MATTER_API_URL` from envsubst; `nginx.conf.template` removes `$matter_api_url` and `/matter/api/v1/` location block; `vite.config.ts` removes `/matter/api/v1` proxy rule.
- ✅ **Inline edit component well-tested** — 307-line test file covers: clear+save, clear+retype, failed save keeps draft (both `false` return and thrown error), draft preserved across external value updates, successful save visible until external update, onStartEdit guard (rejection + void), disabled states, multiline clear+save.
- ✅ **Thread overview section aggregation** — `buildThreadOverviewSection` correctly combines info, md, and webhook sections; `module.tsx` replaces three separate registrations with one `thread.overview` entry.
- ✅ **lucide-react dependency verified** — `packages/dmworkbase/package.json` declares `"lucide-react": "^0.574.0"` and 131+ files already import from it; the new `QrCode` import in `channelSettingGroupProfileRows.tsx` is safe.
- ✅ **i18n keys added in both locales** — `common.clear`, `module.channelSettings.saveFailedRetry`, `subscribers.viewAll`, `subscribers.memberListWithCount` present in both `en-US.json` and `zh-CN.json`.
- ✅ **Error handling in inline-edit callers** — All `onSave` callbacks catch errors and return `false` (keeps editor open for retry) or extract `.msg` for user-visible Toast. `channelSettingGroupManagementRows.tsx` re-throws after Toast, which the InlineEditRow's catch block correctly swallows to keep the editor open.

Static analysis only at head `e8b84e0f`; build and tests not executed in this environment.

## Findings

### P2 — Dead `inputEditPush` parameter in options interfaces

**Diff-scope:** pre-existing interface shape, newly unused by this PR.

`channelSettingGroupProfileRows.tsx` and `channelSettingGroupManagementRows.tsx` both declare `inputEditPush` in their options interface and callers still pass it (e.g. `module.tsx` and `channelSettingNicknameSection.tsx` still pass `_inputEditPush` through the chain). The parameter is no longer destructured or consumed inside these functions.

Similarly, `channelSettingNicknameSection.tsx` and `channelSettingThreadInfoSection.tsx` prefix the parameter with underscore (`_inputEditPush`) to suppress lint warnings, signaling it's intentionally unused but kept for interface compatibility.

Not a bug — the parameter just flows through unused. Worth a follow-up cleanup to remove `inputEditPush` from the options interfaces and all call sites in a single pass.

### P2 — `extractMessageAttachments` and related matter helpers removed from Conversation component

**Diff-scope:** new (removed by this PR).

The `extractMessageAttachments` function (~86 lines) and related matter-integration code were removed from `packages/dmworkbase/src/Components/Conversation/index.tsx`. The component itself remains (still imported by `MessageCell.tsx` via `ConversationContext`). Grep confirms no remaining callers of the removed functions. Clean removal.

## Data Flow Traceback

- **`ChannelSettingInlineEditRow` draft state** — `draft` is initialized from `value` prop on mount; `editingRef.current` preserves draft across external `value` updates while editing; `useEffect` syncs `draft` from `value` only when not editing. Data flows correctly: external update → `draft` syncs → user edits → `draft` local → save → `onSave(draft)` → parent updates `value` prop → `draft` syncs via effect. ✅
- **`onSave` return contract** — `Promise<void | boolean>`: `false` keeps editor open, `void`/`true`/non-false closes it. All callers return `false` on error (after Toast), `undefined`/implicit on success. Contract consistent. ✅
- **`onStartEdit` guard** — Returns `boolean | void | Promise<boolean | void>`: `false` blocks edit start, `void`/`true` allows it. `channelSettingThreadInfoSection.tsx` uses it to check `threadInfo` existence and `canEdit` permission before allowing edit. ✅
- **Clear button data flow** — `onMouseDown` with `preventDefault` prevents Input blur (which would trigger save before clear). Sets draft to `""` and calls `onChange("")`. If `onClear` provided, also calls it. Correct event ordering. ✅
- **Matter removal completeness** — `showMatterPanel` state, `_onToggleMatterPanel`/`_onToggleMatterDetailPanel` handlers, mittBus subscriptions, CSS class `wk-chat-matter-detail-panel-open`, rendering blocks — all removed from `Chat/index.tsx`. No orphaned event emitters remain to trigger removed handlers. ✅

## Blind Spot Checklist

- **C1 (dual-path parity):** N/A — no symmetric add/remove or create/delete paths changed.
- **C2 (control-flow ordering):** N/A — no shared logic reused across multiple call sites with ordering risk.
- **C3 (authorization boundary):** N/A — no permission/auth changes. The `canEdit` check in `onStartEdit` is a UI-level guard, not an auth boundary.
- **C5 (build ≠ runtime):** N/A — no build/bundle changes. `lucide-react` icon swap from `require()` to ES import is standard and tree-shakeable.
- **C6 (governance doc consistency):** N/A — no governance/strategy docs changed.

## Verdict

No P0/P1 issues found. The inline-edit migration is well-tested with good edge-case coverage. The matter-service removal is thorough and consistent across all layers (runtime, config, proxy, i18n). The two P2 items are non-blocking cleanup opportunities.

[Octo-Q] verdict: APPROVE — no P0/P1 issues; PR is clean, well-tested, and consistent across all affected layers.

yujiawei (approved):
# Code Review — PR #1107 (octo-web)

**Head SHA reviewed:** `e8b84e0f6947da43e868592d7205bc1480ae4f15`
**Merge base:** `7561e8ef` · **Diff:** 3 files, +63/−4

> Note: this PR was already merged (`c010c6ff`, 2026-07-27T07:35:35Z) before this review pass completed. Publishing it anyway as the audit record; the two non-blocking items below are follow-up material, not merge blockers.

---

## 1. Spec compliance

**Spec: ✅**

- **Missing work:** none. Both stated goals are implemented: (a) an explicit edit-intent flag so clearing and retyping the original value re-enables Save (`packages/dmworkbase/src/ui/ChannelSettingRows/index.tsx:119`, `:136`), and (b) removal of the visible exit displacement when returning from the group avatar / QR-code child pages while keeping `animationend` alive (`packages/dmworkbase/src/Components/ChannelSetting/index.css:95-107`).
- **Over-build:** none. The diff touches exactly the three files the description claims — the component, its test file, and the ChannelSetting-scoped CSS. No API, bridge, permission, `WKApp`, SDK, datasource, Chat/Conversation/Messages, or shared route changes.
- **Divergence:** none. Every preserved behavior the description lists is verifiable in code and covered by a test: clear (`:174`, `:191`), cancel resets intent (`:215`), failed-save keeps intent so retry stays enabled (`:229-236`), external refresh keeps the in-progress draft via the `editingRef` guard (`:126-131`), successful save clears intent and closes the editor (`:231-233`).

### Validation I reproduced locally at this SHA
- `ChannelSettingInlineEditRow.test.tsx` — **10/10 passed**.
- **Mutation check:** restoring the pre-change `index.tsx` from the merge base makes exactly one test fail (`enables saving after clearing and retyping the original nickname`). The new test genuinely guards the fix rather than passing vacuously — this is the part worth calling out, since a regression test that passes both before and after is the common failure mode here.
- Stylelint on the changed stylesheet — **0 errors, 6 warnings, all pre-existing** (lines 33/45/249/276/292/303, none in the new rules). Matches the description exactly.
- Prettier — the two TS files are clean. The CSS file is reported unformatted both before and after this change, and the repo's `format` script only covers `ts,tsx,md`, so this is not attributable to the PR.

---

## 2. Code quality

**Quality: Approved**

### What holds up
- **State machine is complete.** All five transitions reset or preserve the intent flag correctly, and every async exit path (resolve, resolve-`false`, throw) clears `saving` in `finally` (`index.tsx:238-240`), so no path can leave the row stuck in a loading state.
- **The CSS approach is the correct one, and non-obviously so.** `WKViewQueue.pop()` only flips status to `Pop`; the actual removal from `queues` happens in `poped()`, which is reached only from the `onAnimationEnd` handler (`packages/dmworkbase/src/Components/WKViewQueue/index.tsx:115-130`, `:198-201`). Suppressing the displacement with a no-op keyframe (`from, to { translate3d(0,0,0) }`) keeps `animationend` firing, so the stack still unwinds. Using `animation: none` here would have silently broken pop.
- **Selector scoping verified.** `.wk-channelavatar` and `.wk-channelqrcode` are rendered only by the two child pages pushed from `channelSettingGroupProfileRows.tsx:94` and `:110`, and the left/right `WKViewQueue` instances in `WKLayout` are siblings rather than nested, so `:has()` cannot accidentally match the ChannelSetting entry itself or an ancestor view. All three tokens used (`--wk-dur-fast`, `--wk-ease`, `--wk-bg-surface`) are defined in `theme/semantic.css`.

### Non-blocking findings

**P2 — Intent-based enablement permits a genuine no-op write.**
`saveDisabled` no longer compares `draft` to `currentValue` (`index.tsx:136`), so typing one character and deleting it leaves Save enabled, and pressing it issues a real `updateChannelField` / remark PUT with an unchanged value. There is no client-side equality guard on either path. This is an unavoidable consequence of the requirement ("clear + retype the original must be savable") — the two rules are genuinely in tension, so the trade-off is defensible. Worth confirming on the server side that an identical group-name update does not emit a "group name changed" system message to every member; if it does, a cheap `draft === currentValue` short-circuit inside the Save handler (enabled button, no request) would give you both behaviors.

**P2 — Third copy of the same no-op-exit pattern; extract it.**
`Subscribers/list.css:14-26` and now `ChannelSetting/index.css:95-107` duplicate the same rule plus a near-identical private keyframe. `dmworksummary/src/index.css:97-104` solves the same problem with `animation: none !important`, which — per the `poped()` chain above — means popped summary views are never removed from `queues`. That divergence is a latent bug outside this diff, and it is exactly what a shared opt-out utility on `WKViewQueue` (e.g. a `wk-viewqueue-no-slide` class exposing one canonical no-op keyframe) would have prevented. Recommend extracting it and migrating the summary package in a follow-up.

**P2 (nit) — Test name no longer describes what it asserts.**
`disables saving for unchanged, empty, and over-limit values` now exercises "untouched", not "unchanged" — a reverted-to-original draft leaves Save enabled. Renaming it, and adding an explicit case for type-then-revert, would pin the new semantics deliberately instead of leaving it inferred.

**P2 (informational) — `:has()` support.**
The rule silently no-ops on engines without `:has()` (Firefox < 121, Chrome < 105), degrading to the current slide rather than breaking. The repo has no browserslist config and already ships two other `:has()`-based overrides, so this is a pre-existing baseline choice rather than something this PR introduces.

### Checked and found clean
- Save-button enablement while a request is in flight: the custom clear affordance is rendered only when `draft && !saving` (`index.tsx:171-173`), so it is not reachable mid-save and cannot wipe a draft that a failed save is about to restore.
- `setSaving(false)` in `finally` can land after unmount; benign no-op in React 18, no cross-row identity to misroute.
- No `dangerouslySetInnerHTML`, no user input reaching a selector or style value.

### Coverage gaps in this pass
Not exercised: real browser behavior of the new keyframe (verified by reading, plus the green `e2e-p0` job, not by driving the UI); server-side response to an identical-value field update; and the `animationend` bubbling characteristic of `WKViewQueue` — the handler is attached to the wrapper and animation events bubble, so a descendant animation completing inside the top view can also trigger `animationEnd()`. That fragility is pre-existing and untouched here, but it is the surface most likely to bite the next change in this area.

---

## 3. Verdict

**APPROVED** — Spec ✅ and Quality ✅. No P0/P1. The three P2 items are follow-up work.

PR #1109, opened 2026-07-27
feat(search): add scoped pinyin matching

Candice00711:
## Summary
- add full-pinyin and case-insensitive matching to the existing global contact/group results and sender/member/channel pickers
- preserve the existing literal/server request, pagination, cache, ordering, fallback, click, and IME paths; pinyin-only matches are deduplicated and appended after existing results
- reuse the existing empty-keyword `/search/global` response as the authoritative Space-scoped contact/group source, and limit local sender pinyin candidates to UIDs confirmed by the current Space server response
- rebuild pinyin indexes from each current candidate snapshot while keeping the sender identity cache across same-Space contact changes

## Validation
- 11 focused GlobalSearch test files, 111 tests passed
- full `dmworkbase` suite: 255 files, 2225 tests passed
- `pnpm i18n:check` equivalent passed
- Web production build passed
- E2E mock bundle build passed
- `git diff --check` passed

## Regression coverage
- no extra `/group/my` request when global search opens
- previous-Space account-level contacts cannot enter the current Space pinyin supplement
- sender chips retain server-derived names and avatars after same-Space contact changes
- contact rename, pool shrink, blacklist, deletion, and no-longer-followed states
- readable channel pool shrink; channel IDs and hidden aliases do not broaden matching
- existing server-first ordering, 50-item limit, and remote sender search calls
- in-flight sender/channel results cannot populate a different Space cache
- server result ordering remains unchanged and pinyin hits only append
- 10,000-name index converts once and reuses cached pinyin

### DIFF
--- a/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
+++ b/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
@@ -21,6 +21,9 @@ import { describe, expect, it, vi, beforeEach } from "vitest";
 const mockState = vi.hoisted(() => ({
   commonDataSource: undefined as any,
   contactsList: [] as any[],
+  groups: [] as any[],
+  groupSaveList: undefined as undefined | (() => Promise<any[]>),
+  spaceId: "space-1",
   loginUid: "self-uid",
   loginName: "Me",
 }));
@@ -31,6 +34,13 @@ vi.mock("../../../App", () => ({
       return {
         commonDataSource: mockState.commonDataSource,
         contactsList: mockState.contactsList,
+        channelDataSource: {
+          groupSaveList: vi.fn(() =>
+            mockState.groupSaveList
+              ? mockState.groupSaveList()
+              : Promise.resolve(mockState.groups)
+          ),
+        },
       };
     },
     get loginInfo() {
@@ -41,6 +51,9 @@ vi.mock("../../../App", () => ({
       };
     },
     shared: {
+      get currentSpaceId() {
+        return mockState.spaceId;
+      },
       avatarUser: (uid: string) => `avatar://user/${uid}`,
       avatarChannel: (ch: any) =>
         `avatar://ch/${ch?.channelID ?? ""}/${ch?.channelType ?? ""}`,
@@ -52,24 +65,28 @@ vi.mock("../../../App", () => ({
   },
 }));
 
-vi.mock("wukongimjssdk", () => ({
-  Channel: class {
-    channelID: string;
-    channelType: number;
-    constructor(channelID: string, channelType: number) {
-      this.channelID = channelID;
-      this.channelType = channelType;
-    }
-  },
-  ChannelTypeGroup: 2,
-  ChannelTypePerson: 1,
-  WKSDK: {
+vi.mock("wukongimjssdk", () => {
+  const sdk = {
     shared: () => ({
       conversationManager: { conversations: [] },
       channelManager: { getChannelInfo: () => undefined },
     }),
-  },
-}));
+  };
+  return {
+    default: sdk,
+    Channel: class {
+      channelID: string;
+      channelType: number;
+      constructor(channelID: string, channelType: number) {
+        this.channelID = channelID;
+        this.channelType = channelType;
+      }
+    },
+    ChannelTypeGroup: 2,
+    ChannelTypePerson: 1,
+    WKSDK: sdk,
+  };
+});
 
 import { createGlobalSearchApiDataSource } from "../dataSource";
 
@@ -77,6 +94,9 @@ describe("loadSenderCandidates (via searchSenders)", () => {
   beforeEach(() => {
     mockState.commonDataSource = undefined;
     mockState.contactsList = [];
+    mockState.groups = [];
+    mockState.groupSaveList = undefined;
+    mockState.spaceId = "space-1";
   });
 
   it("§1: maps ChannelInfo[] from commonDataSource.searchFriends into ChannelSearchSender", async () => {
@@ -141,9 +161,7 @@ describe("loadSenderCandidates (via searchSenders)", () => {
 
   it("§3: surfaces contactsList when searchFriends is missing entirely", async () => {
     mockState.commonDataSource = {}; // no searchFriends method
-    mockState.contactsList = [
-      { uid: "erin-uid", name: "Erin" },
-    ];
+    mockState.contactsList = [{ uid: "erin-uid", name: "Erin" }];
 
     const ds = createGlobalSearchApiDataSource();
     const results = await ds.searchSenders("");
@@ -154,14 +172,12 @@ describe("loadSenderCandidates (via searchSenders)", () => {
   it("§4: cold panel (empty sender cache) still returns candidates from the real DS API — regression guard", async () => {
     // The exact scenario Jerry-Xin flagged: no prior search results have
     // warmed the sender cache. Before the fix, this returned only [self].
-    const searchFriends = vi
-      .fn()
-      .mockResolvedValue([
-        {
-          channel: { channelID: "frank-uid", channelType: 1 },
-          orgData: { displayName: "Frank" },
-        },
-      ]);
+    const searchFriends = vi.fn().mockResolvedValue([
+      {
+        channel: { channelID: "frank-uid", channelType: 1 },
+        orgData: { displayName: "Frank" },
+      },
+    ]);
     mockState.commonDataSource = { searchFriends };
 
     const ds = createGlobalSearchApiDataSource();
@@ -211,5 +227,164 @@ describe("loadSenderCandidates (via searchSenders)", () => {
 
     const ds = createGlobalSearchApiDataSource();
     await expect(ds.searchSenders("")).resolves.toBeDefined();
+  });
+
+  it("matches sender/member candidates by full pinyin without skipping remote search", async () => {
+    const searchFriends = vi.fn().mockResolvedValue([]);
+    mockState.commonDataSource = { searchFriends };
+    mockState.contactsList = [{ uid: "jia-uid", name: "贾小明" }];
+
+    const ds = createGlobalSearchApiDataSource();
+    const results = await ds.searchSenders("jia");
+
+    expect(searchFriends).toHaveBeenCalledWith("jia");
+    expect(results.map((sender) => sender.uid)).toContain("jia-uid");
+  });
+
+  it("matches group options by full pinyin", async () => {
+    mockState.groups = [
+      {
+        channel: { channelID: "group-1", channelType: 2 },
+        title: "全能接项目小组",
+        orgData: { displayName: "全能接项目小组" },
+      },
+    ];
+
+    const ds = createGlobalSearchApiDataSource();
+    const results = await ds.searchChannels("quanneng");
+
+    expect(results.map((channel) => channel.channelId)).toContain("group-1");
+  });
+
+  it("removes stale channel candidates when the readable pool changes", async () => {
+    mockState.groups = [
+      {
+        channel: { channelID: "group-1", channelType: 2 },
+        displayName: "全能接项目小组",
+      },
+    ];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchChannels("quanneng")).resolves.toHaveLength(1);
+
+    mockState.groups = [];
+    await expect(ds.searchChannels("quanneng")).resolves.toEqual([]);
+  });
+
+  it("clears cached pinyin candidates after switching spaces", async () => {
+    mockState.contactsList = [{ uid: "old-uid", name: "贾小明" }];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchSenders("jia")).resolves.toEqual(
+      expect.arrayContaining([expect.objectContaining({ uid: "old-uid" })])
+    );
+
+    mockState.spaceId = "space-2";
+    mockState.contactsList = [];
+    await expect(ds.searchSenders("jia")).resolves.toEqual([]);
+  });
+
+  it("removes stale sender candidates when the contact pool shrinks in the same space", async () => {
+    mockState.commonDataSource = {
+      searchFriends: vi.fn().mockResolvedValue([]),
+    };
+    mockState.contactsList = [{ uid: "old-uid", name: "贾小明" }];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchSenders("jia")).resolves.toEqual(
+      expect.arrayContaining([expect.objectContaining({ uid: "old-uid" })])
+    );
+
+    mockState.contactsList = [];
+    await expect(ds.searchSenders("jia")).resolves.toEqual([]);
+    expect(ds.getSenders().some((sender) => sender.uid === "old-uid")).toBe(
+      false
+    );
+  });
+
+  it("removes renamed sender pinyin and excludes blacklisted contacts", async () => {
+    mockState.commonDataSource = {
+      searchFriends: vi.fn().mockResolvedValue([]),
+    };
+    mockState.contactsList = [{ uid: "user-1", name: "贾小明" }];
+    const ds = createGlobalSearchApiDataSource();
+    await expect(ds.searchSenders("jia")).resolves.toHaveLength(1);
+
+    mockState.contactsList = [{ uid: "user-1", name: "张小明" }];
+    await expect(ds.searchSenders("jia")).resolves.toEqual([]);
+    await expect(ds.searchSenders("zhang")).resolves.toHaveLength(1);
+
+    mockState.contactsList = [{ uid: "user-1", name: "张小明", status: 2 }];
+    await expect(ds.searchSenders("zhang")).resolves.toEqual([]);
+  });
+
+  it("preserves the existing server-first candidate order", async () => {
+    mockState.commonDataSource = {
+      searchFriends: vi.fn().mockResolvedValue([
+        {
+          channel: { channelID: "server-1", channelType: 1 },
+          orgData: { displayName: "服务端一" },
+        },
+        {
+          channel: { channelID: "server-2", channelType: 1 },
+          orgData: { displayName: "服务端二" },
+        },
+      ]),
+    };
+    mockState.contactsList = [{ uid: "local-1", name: "本地联系人" }];
+
+    const ds = createGlobalSearchApiDataSource();
+    const results = await ds.searchSenders("");
+
+    expect(results.map((sender) => sender.uid)).toEqual([
+      "self-uid",
+      "server-1",
+      "server-2",
+      "local-1",
+    ]);
+  });
+
+  it("does not write an in-flight sender response into a new space", async () => {
+    let resolveSearch: (value: any[]) => void = () => undefined;
+    mockState.commonDataSource = {
+      searchFriends: vi.fn().mockImplementation(
+        () =>
+          new Promise<any[]>((resolve) => {
+            resolveSearch = resolve;
+          })
+      ),
+    };
+    const ds = createGlobalSearchApiDataSource();
+    const pending = ds.searchSenders("old");
+
+    mockState.spaceId = "space-2";
+    mockState.contactsList = [];
+    resolveSearch([
+      {
+        channel: { channelID: "old-space-user", channelType: 1 },
+        orgData: { displayName: "旧空间用户" },
+      },
+    ]);
+
+    await expect(pending).resolves.toEqual([]);
+    expect(ds.getSenders().map((sender) => sender.uid)).toEqual(["self-uid"]);
+  });
+
+  it("does not write in-flight channel options into a new space", async () => {
+    let resolveGroups: (value: any[]) => void = () => undefined;
+    mockState.groupSaveList = () =>
+      new Promise<any[]>((resolve) => {
+        resolveGroups = resolve;
+      });
+    const ds = createGlobalSearchApiDataSource();
+    const pending = ds.searchChannels("quanneng");
+
+    mockState.spaceId = "space-2";
+    resolveGroups([
+      {
+        channel: { channelID: "old-space-group", channelType: 2 },
+        title: "全能接项目小组",
+        orgData: { displayName: "全能接项目小组" },
+      },
+    ]);
+
+    await expect(pending).resolves.toEqual([]);
   });
 });
--- a/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
+++ b/packages/dmworkbase/src/Components/GlobalSearch/__tests__/senderCandidates.test.ts
@@ -22,7 +22,6 @@ const mockState = vi.hoisted(() => ({
   commonDataSource: undefined as any,
   contactsList: [] as any[],
   groups: [] as any[],
-  groupSaveList: undefined as undefined | (() => Promise<any[]>),
   spaceId: "space-1",
   loginUid: "self-uid",
   loginName: "Me",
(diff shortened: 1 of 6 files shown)


Jerry-Xin (dismissed):
**APPROVE** — scoped pinyin matching is well-architected, and it correctly closes the two stale-index blockers raised on the prior global-search pinyin PR. Every scope index now rebuilds on a signature/reset key instead of appending, and the tests assert the *shrink/rebuild* behavior (not just "handler fired"). Build (`apps/web`) is green and all 27 focused tests pass at head `ab85d1d`.

### Verification of prior blockers (now fixed)
- **Prior blocker A (append-only contact index kept deleted contacts):** `GlobalSearchVM.contactsChangeListener` now routes through `rebuildPinyinIndexWithCurrentContacts()` → `rebuildGlobalSearchPinyinIndex(...)`, which rebuilds keyed on `channel_type:channel_id` from the *current* `contactsList` snapshot each time, so removed/renamed contacts drop out. Test `rebuilds authoritative snapshots without removed or renamed entries` locks this in (removed friend → `weijiao` returns `[]`).
- **Prior blocker B (sender/channel index only reset on space change):** `createGlobalSearchDataSource.resetScopedPinyinIndexes()` now also resets when `localContactSignature()` (uid+remark/name+status) changes *within the same space* — it clears `senderCache`, recreates `senderPinyinIndex`, and re-seeds self. Channels use the same `channelCandidateSignature` rebuild pattern consistently. Tests `removes stale sender candidates when the contact pool shrinks in the same space`, `removes renamed sender pinyin and excludes blacklisted contacts`, and `removes stale channel candidates when the readable pool changes` all assert the pool actually shrinks.

### ✅ Highlights
- Pinyin conversion happens once at build time and is cached in `searchText`; keystrokes are pure substring. Verified by the 10,000-name test asserting `toPinyin` is called exactly 10,000 times across 20 searches, and the rebuild test asserting unchanged entries are not reconverted.
- Server results stay authoritative: local pinyin matches are deduped by `channel_type:channel_id` and appended only after server rows; order-preservation and the 50-item sender limit are tested (`self, server-1, server-2, local-1`).
- Correctness coverage is real: full-pinyin, initials-vs-full, case-insensitivity (`WEIJIAO`), tone-none, remark-over-name, mixed CJK+latin, empty-query, and no-match paths are all exercised.
- Race safety: in-flight sender/channel responses are dropped if the space changed mid-request (`requestSpaceId !== currentSpaceId`), with dedicated tests for both.

### 🟡 Non-blocking notes
1. `pinyinGroupCandidates` (the readable-group pool feeding contact-change rebuilds in `GlobalSearchVM`) is only refreshed on mount and space change — there is no listener for join/leave of a group *within the same space*. A stale group can linger in the local pinyin supplement until the next remount/space switch. Low impact since groups are also server-authoritative and local hits are only an appended supplement, but a group-list-change listener (mirroring `contactsChangeListener`) would make freshness symmetric.
2. In `createGlobalSearchDataSource`, senders added to `senderCache` from server message rows (`searchMessages` → `rememberSender`) are not part of `localContactSignature()`, so they persist across same-space contact-pool changes until a space switch clears the cache. These are server-sourced identities so it is acceptable, but worth a comment noting the signature only tracks the local-contact pool.
3. Package-level `tsc --noEmit` reports test-file type noise (vitest globals, optional-member `possibly undefined`) for the new/edited `.test.ts` files. These do not appear in the `apps/web` build and do not affect runtime, but tightening the test tsconfig (or non-null asserting the data-source handles) would keep a standalone typecheck clean.

lml2468 (dismissed):
## Review: #1109 — feat(search): add scoped pinyin matching

**意图**:#1095 的干净重写(分支 `codex/global-search-pinyin-clean`)—— 给全局联系人/群结果与 sender/member/channel 选择器加全拼 + 大小写不敏感匹配;保留服务端请求/分页/缓存/排序/兜底/点击/IME;本地拼音匹配去重后**仅追加在既有结果之后**;联系人/源/真实群快照/Space 变化时**重建**作用域索引。锚定 head `ab85d1de`(rev-parse 确认 checkout==live head)/ `mergeable=true`。5 文件 +1048/−48。

**结论:APPROVE ✅ —— 我在 #1095 漏判/peers 提的 🔴 在此重写中逐条修复。**

### #1095 的阻塞项逐条验证(全修复)
- **🔴 索引只增不减(我上轮漏判) → 已修**:`rebuildGlobalSearchPinyinIndex` 从 `uniqueByChannel(incoming)`(**当前**源)重建 —— 不在当前源的条目被丢弃(陈旧移除);仅当 `cached.sourceText === sourceText` 才复用缓存拼音(改名→重建条目)。`contactsChangeListener` 另先 `createEmpty` 再 `refreshPinyinIndex`。改名/删除/同空间池收缩都正确 reconcile,不再残留已删对象。
- **🔴 黑名单未过滤 → 已修**:index build(:241)与 sender candidates(:116/:296)均 `status !== ContactsStatus.Blacklist` 过滤。
- **🔴 群名字段选取错误 → 已修**:真实 ChannelInfo 读 `displayName`/`orgData`(:255-266),legacy 走 `channel_remark||channel_name||channel_id`。
- **服务端排序保留**:`mergeGlobalSearchPinyinResults` + `seen` Set 去重,本地匹配**追加在服务端结果之后**,不重排(回归测试「server result ordering unchanged」)。
- **跨空间在飞守卫**:`indexedSpaceId`/`spaceChanged` 触发重建;in-flight `requestSpaceId !== currentSpaceId → return []`(:289/:335),不跨空间写。

### Gate 2 其余
- **无授权变更**:仍是已加载候选的客户端过滤 + 服务端搜索优先(回归测试「remote search remains invoked」);权限服务端兜底。
- **无 XSS**:纯匹配逻辑,不渲染;`pinyin-pro` 既有依赖。
- **性能**:每候选建索引时转一次、缓存 searchText;每键子串扫描;10k 名字只转一次(有测试)。

**构建 / 测试 / i18n(实跑 @ `ab85d1de`)**:`pnpm --filter @octo/web build` ✅ 4.04s;`pnpm i18n:check` ✅;`globalSearchPinyin.test` + `senderCandidates.test` **2 files / 27 tests 全绿**(较 #1095 的 19 增加了收缩/黑名单/改名/群字段/跨空间回归);PR 报全量 11 文件 116 测试。

---

### 🟡 非阻断
- 大空间下索引 O(n) 重建(仅 scope 变化时、有缓存),超大空间可留意构建耗时;有 10k-once 测试佐证。

---

#1095 的三条 🔴(append-only 陈旧索引【我上轮漏判、peers 判对】、黑名单未过滤、群名字段)在此重写逐条修复:改用**重建 reconciliation**(丢弃陈旧)+ 黑名单过滤 + 真实 ChannelInfo 字段;服务端排序/兜底/跨空间守卫保留,无授权/XSS 变更,构建/i18n/27 测试通过。**可合并 —— APPROVE**(此 PR 正是我在 #1095 漏判的 stale-index 🔴 的正解)。

yujiawei (changes requested):
# Code Review — PR #1109 (octo-web)

Reviewed at head SHA `ab85d1dee025ea00a42612d5813b723a954ff3e1` against merge-base `7561e8ef`. Repo checked out at that SHA; `pnpm install` + the full `packages/dmworkbase` suite run locally (**254 files / 2225 tests pass**, including the 27 new ones). Two findings below are backed by executable repros I wrote against this branch — their output is quoted inline.

---

## 1. Spec compliance

**Spec: ❌**

The PR states three goals: (a) add full-pinyin + case-insensitive matching to global contact/group results and the sender/member/channel pickers; (b) *"preserve the existing server request, pagination, cache, ordering, fallback, click, and IME behavior"*; (c) *"rebuild scoped indexes for contact/source changes, real group snapshots, and **Space changes**"*.

**Missing / diverging:**

- **(c) is not achieved for the contact leg.** The Space-scoping machinery (`resetScopedPinyinIndexes`, `createGlobalSearchDataSource.ts:253-270`) resets *indexes*, but the candidate pool it rebuilds from — `WKApp.dataSource.contactsList` — is not Space-scoped and is never cleared on a Space switch. See P0 below, with a failing repro.
- **(b) "preserve the existing cache" is violated.** `senderCache` previously only ever grew; it is now cleared whenever the contact pool changes (`createGlobalSearchDataSource.ts:262`), which discards server-derived sender identities that have nothing to do with contacts. See P1 below, with a failing repro.

**Over-built (outside the stated scope):**

- `createGlobalSearchDataSource.ts:93-97` changes the **display-name resolution priority** for channel options from `displayName || name || channelId` to `orgData.displayName || title || displayName || name || channelId`. This is a user-visible rename of picker entries, unrelated to pinyin matching, and it is not mentioned in the summary or covered by a test.
- `createGlobalSearchDataSource.ts:352` — `searchChannels` now indexes `[option.name, option.channelId]`. The previous filter was `o.name.toLowerCase().includes(kw)`, i.e. **name only**. Typing a substring that happens to occur inside an opaque channel ID now surfaces unrelated channels, and every channel ID is needlessly pushed through the traditional→simplified converter and pinyin-pro.

The pinyin/case-insensitive matching itself (goal a) is implemented, and the server-first ordering, 50-item cap, remote-call preservation, and IME guard are all intact — those parts of the summary check out.

---

## 2. Code quality

### P0 — Contacts from other Spaces leak into the current Space's search results

`packages/dmworkbase/src/bridge/globalSearch/GlobalSearchVM.ts:216-224`

```ts
private rebuildPinyinIndexWithCurrentContacts() {
  const contacts = globalSearchContactsToLegacy(
    WKApp.dataSource.contactsList || []
  );
```

`contactsList` is a single account-level array on `DataSource` (`Service/DataSource/DataSource.ts:17`). `contactsSync()` (`:20-31`) merges incrementally by `uid` and **never removes**, and its cursor `contactsMaxSyncVersion()` reads the last element of that cross-Space array. Meanwhile every request carries an injected `X-Space-Id` header (`Service/APIClient.ts:96-99`), so each `contactsSync` returns *the current Space's* contacts — which then accumulate on top of the previous Space's. Nothing resets `contactsList` on `space-changed`; the only reset is `clearLocalLoginState()` (`App.tsx:1021`), which immediately does `window.location.replace("/login")`.

Consequence: after visiting Space A then switching to Space B, the pinyin index — and therefore `searchResult.friends` via `applyPinyinMatches` (`GlobalSearchVM.ts:226-239`) — contains Space A's contacts. **The global-search contacts tab was previously 100% server-sourced and therefore Space-scoped; this PR makes it show contacts the current Space's backend deliberately did not return.**

Repro (the existing harness, with the one line that production does *not* perform removed):

```ts
mockState.contactsList = [{ uid: "space1-only-user", name: "贾小明" }];
const ds = createGlobalSearchApiDataSource();
expect((await ds.searchSenders("jia")).map(s => s.uid)).toContain("space1-only-user"); // ok

mockState.spaceId = "space-2";
// contactsList intentionally left as-is — this is what production does.
expect((await ds.searchSenders("jia")).map(s => s.uid)).toEqual([]);
```
```
- Expected: []
+ Received: [ "space1-only-user" ]
```

**The existing Space test cannot catch this.** `__tests__/senderCandidates.test.ts:273-283` asserts the guarantee, but line 281 (`mockState.contactsList = []`) hand-clears the pool the code under test reads from — so it verifies the mock, not the invalidation logic:

```ts
mockState.spaceId = "space-2";
mockState.contactsList = [];               // ← production never does this
await expect(ds.searchSenders("jia")).resolves.toEqual([]);
```

To be fair on attribution: the pre-existing `loadSenderCandidates` fallback (`:198-215`) already read the same unscoped list, so cross-Space exposure in the *sender picker* predates this PR. What is new is the **contacts tab** consumer, plus a claimed-and-tested Space guarantee that does not hold. Please either scope the candidate pool by Space at the source (clear/partition `contactsList` on `space-changed`), or filter local candidates against a Space membership set before indexing — and make the test assert it without pre-clearing the mock.

### P1 — `senderCache` is cleared by unrelated contact changes, so filter chips fall back to raw UIDs

`packages/dmworkbase/src/bridge/globalSearch/createGlobalSearchDataSource.ts:253-270`

```ts
const senderPoolChanged =
  nextSenderCandidateSignature !== senderCandidateSignature;
if (!spaceChanged && !senderPoolChanged) return;
...
senderCache.clear();
```

`senderCache` does not only hold contact-derived entries. It is also populated from `searchFriends` results (`:300`) and from message-search hits (`:386`, where `SearchResultMapper.senderFromHit` supplies `name` + `avatarUrl`). Keying its invalidation on a *contacts* signature therefore throws away server-derived identities that a contacts change says nothing about.

`getSender` (`:277-285`) then falls through to `{ uid, name: uid }` — no display name, no avatar. Both selected-chip memos read exactly that path:

- `Components/GlobalSearch/GlobalSearchFilterPanel.tsx:515-519` (`senderSelected`)
- `Components/GlobalSearch/GlobalSearchFilterPanel.tsx:563-567` (`memberSelected`)

Their deps are `[draft.senderUids, dataSource]`, so they recompute on the user's next filter edit — at which point every already-picked chip renders a raw UID. The `searchSenders` effect (`GlobalSearchFilterPanel.tsx:214-229`) only re-runs when `senderQuery` changes, so it does not heal this.

Repro:

```ts
await ds.searchSenders("");                       // warm the cache
// BEFORE sync: {"uid":"server-only","name":"服务端用户","avatarUrl":"http://a/x.png",...}

mockState.contactsList = [ ...previous, { uid: "contact-b", name: "乙" } ];  // routine contactsSync
ds.getSender("server-only");
// AFTER  sync: {"uid":"server-only","name":"server-only"}
```

A routine background `contactsSync` (fired on reconnect and on any contact add/rename/remove — `module.tsx:508,525,1371,1383`, `App.tsx:976`) is enough to trigger it.

Compounding this: `senderCandidateSignature` is initialized to `""` (`:244`) rather than seeded from the current list, so the **first** `getSender`/`getSenders` read after the data source is constructed always reports a change and wipes whatever `searchMessages` had already cached. Suggested fix: keep `senderCache` as a pure identity cache (never cleared by contact churn — evict only on Space change), and let the pinyin index be the only signature-invalidated structure. Seeding the signature at construction would also remove the spurious first-read wipe.

### P2 — `localContactSignature()` runs on every `getSender` / `getSenders` read

`createGlobalSearchDataSource.ts:128-140`, called from `:274` and `:278`.

Each read maps, sorts, and joins the entire contact list. Measured on this branch:

| contacts | 1 signature | 50 signatures |
|---|---|---|
| 1,000 | 0.44 ms | 13.7 ms |
| 5,000 | 1.12 ms | 58.3 ms |
| 10,000 | 2.83 ms | 120.6 ms |

Worth noting for calibration: this is *not* per result row. `SearchResultMapper.ts:384/425/470/523` always populate `item.sender`, and `ChannelSearchResults.tsx:38,61` use `item.sender || getSender(uid)`, so rendered rows short-circuit. Real exposure is the selected-chip memos — typically 1–5 uids — so this is a cleanup, not a freeze. Reference-comparing `contactsList` (it is replaced wholesale by `contactsSync`, never mutated in place) would make it O(1); `addContactsChangeListener`, which the VM already uses, is the cleaner signal.

### P2 — Two different display names for the same group

`createGlobalSearchDataSource.ts:93-97` resolves `orgData.displayName || title || displayName || name || channelId`, while `globalSearchPinyin.ts:263-270` resolves `orgData.remark || displayName || orgData.displayName || title || name || channelId` — from the same `ChannelInfo`. A group with `orgData.remark = "备注群"` and `title = "原名群"` renders as **"备注群"** in the groups tab and **"原名群"** in the channel picker, in the same search UI. Please extract one shared resolver.

### P2 — The group candidate snapshot is never refreshed while the panel is open

`GlobalSearchVM.ts:189-214`. `refreshPinyinIndex()` is only reached from `didMount` (`:136`) and `spaceChangedListener` (`:69`). There is no `groupSaveList` change listener, so joining, leaving, or renaming a group while the search panel is open leaves `pinyinGroupCandidates` stale. `refreshGlobalSearchGroupCandidates` handles pool shrink correctly — but the only thing that exercises it after mount is the unit test, never production. Contacts get this right via `addContactsChangeListener`; groups need the equivalent.

### P2 — In-channel search pays for a pinyin index it never uses

`GlobalSearchVM.ts:60-64`. `contactsChangeListener` calls `rebuildPinyinIndexWithCurrentContacts()` unconditionally, but `applyPinyinMatches()` (`:227`) early-returns whenever `searchInChannel` is true — and `refreshPinyinIndex()` (`:190`) already guards on it. So in-channel search converts the whole contact list to pinyin on every contacts change and discards the result. Add the same `if (this.searchInChannel) return;` guard to the listener.

### P2 — `buildGlobalSearchPinyinIndex` is exported but unreachable from production

`globalSearchPinyin.ts:139-152` has no non-test caller (production goes `createEmptyGlobalSearchPinyinIndex` → `rebuildGlobalSearchPinyinIndex`). Six of the headline assertions — including *"matches Chinese, full pinyin and case-insensitive pinyin"*, *"uses the visible remark for matching"*, and *"converts 10,000 names once"* — run against that entry point. They are equivalent by construction today, but the coverage claim is weaker than it reads. Either route production through it or test `rebuildGlobalSearchPinyinIndex` directly.

Relatedly, the *"converts 10,000 names once"* test injects a mock `toPinyin`, so it measures memoization, not conversion cost. The real converter chains `toSimplized` (`Utils/t2s.ts`, `indexOf` over a 2,553-char table per CJK char) with pinyin-pro. I measured `toSimplized` at ~10.5 ms for 10,000 three-character names — acceptable — but pinyin-pro's own cost at that scale is unmeasured by both the test and by me.

### P2 — Local matches are gated on a successful server response

`GlobalSearchVM.ts:227` requires `this.searchResult` to be non-null, and `initLoad()` (`:184`) sets it to `null`. When the server search fails, `searchError` is set and `searchResult` stays `null`, so a fully warm local index yields zero visible matches. Consistent with "the server stays authoritative," but surfacing local hits on failure is exactly where an offline-capable index earns its cost — worth a follow-up.

### P2 — Unguarded `getValues` outputs in `extendNamedPinyinSearchIndex`

`globalSearchPinyin.ts:73-75` calls `value.trim().toLowerCase()` on each `getValues` result. Every current caller passes required `string` fields, so this is safe today; a `?? ""` would make it robust to schema drift on the remote sender shape.

---

## 3. Overall verdict

**REQUEST_CHANGES** — the P0 (cross-Space contact exposure on a newly added surface, with a test that cannot detect it) and the P1 (sender chips degrading to raw UIDs after a routine background sync) both need fixing. Everything from P2 down is non-blocking.

## 4. Suggested direction

1. Scope the local contact pool by Space before indexing, and rewrite `senderCandidates.test.ts:273-283` so it does not pre-clear `contactsList` — the test should fail against the current implementation.
2. Split the two caches: `senderCache` = identity cache, evicted only on Space change; pinyin index = the only signature-invalidated structure. Seed `senderCandidateSignature` at construction.
3. Replace the read-path signature with a reference comparison on `contactsList`, or drive invalidation from `addContactsChangeListener`.
4. Extract one shared `ChannelInfo` → display-name resolver used by both the picker and the legacy groups path.
5. Add a `groupSaveList` refresh trigger, and guard `contactsChangeListener` on `searchInChannel`.
6. Revert or separately justify the two out-of-scope behavior changes (channel-option name priority; `channelId` matching in `searchChannels`).

## 5. Additional observations (not blocking)

- The `space-changed` listener resets `serverContactGroupResult` and re-runs `applyPinyinMatches`, which clears stale contacts/groups from the previous Space — but `searchResult.messages` is untouched and `requestSearch()` never re-checks its captured `spaceId` on response (`:249`, `:262`). So an in-flight search started in Space A still renders its messages in Space B. This predates the PR and the new listener slightly improves the contacts/groups half; flagging it because this PR is the first code to take Space transitions seriously here and is a natural place to finish the job.
- `didMount`/`didUnMount` listener registration is balanced (`Service/Provider.tsx:113,119`), and `WKApp.shared.currentSpaceId` is assigned *before* `mittBus.emit('space-changed')` (`Pages/Chat/vm.ts:101-102`), so the new listener reads the new Space ID correctly. Both checked; no issue.
- `loadMore()` behavior is preserved: `serverContactGroupResult` holds the page-1 baseline and is not overwritten on the `loadMoreing` branch, so `applyPinyinMatches` reproduces the previous friends/groups content rather than clobbering it. Verified by reading; no issue.

## 6. Review coverage

What I could not verify: no live deployment was available, so nothing here was confirmed against a running client — the Space-transition and chip-degradation behaviors are established from code plus the two repros above, not from a browser. pinyin-pro's conversion cost at 10k names is unmeasured. The `/search/global` response shape is taken from the TypeScript types rather than an observed payload. Whether Space is intended as a hard boundary for the contacts domain is a product question I inferred from the `X-Space-Id` injection and from this PR's own scoping code and test — if contacts are deliberately account-level, the P0 becomes "delete the misleading scoping machinery and test" rather than "fix the leak," and I would want a maintainer to confirm which.

Candice00711:
Closing in favor of a fresh implementation from the latest upstream/main. The replacement will reuse the established contacts pinyin-search pattern and keep the change additive-only: preserve the existing search requests, ordering, pagination, caching, permissions, IME, and click behavior, with only deduplicated pinyin matches appended.

Jerry-Xin (approved):
**APPROVE** — re-review at `28b730ced59d`. Clean single-commit delta on top of the prior head (`compare` reports `ahead:1, behind:0`, no rebase), and it materially tightens the scoped pinyin design without regressing the two RC #1095 blockers. Build is green in CI and all touched unit tests pass locally (27/27).

The commit replaces the listener-driven index machinery with a server-authoritative model: the pinyin candidate pool is now derived exclusively from the empty-keyword `searchLegacyGlobal` response and rebuilt via an immutable `rebuildNamedPinyinSearchIndex`, so removed/renamed/blacklisted entries can no longer linger. This is a stronger guarantee than the previous append-then-reset approach.

### ✅ Verified
- **RC #1095 blocker A (stale-on-change) stays fixed.** `GlobalSearchVM` no longer relies on `contactsChangeListener`; the pool is re-derived from the authoritative empty-keyword snapshot on each open/empty search and the index is fully rebuilt each request, so a shrunk/renamed pool drops stale rows. `GlobalSearchVM.test.ts` asserts the append then empty→re-query returns `[]`, and `senderCandidates.test.ts` asserts real shrink-to-`[]` on pool removal (not handler-fired).
- **RC #1095 blocker B (scope reset) stays fixed.** `resetSpaceScopedState` clears index + cache + `scopedFriendUids` on space-id change; `GlobalSearchVM.requestSearch` resets all pinyin state inline when `spaceId !== pinyinSpaceId`, and a late in-flight response is dropped by the `spaceId === currentSpaceId` re-check. Covered by the cross-Space tests in both suites.
- **Prior non-blocking notes addressed.** The sender/channel signature tracking is superseded by `scopedFriendUids` (only server-confirmed friends feed the local-contacts pinyin supplement, plus `beDeleted` / `follow === 0` / blacklist exclusion). The separate group-listener concern is now moot since the group pool comes from the server empty-keyword snapshot rather than an independent `groupSaveList` refresh path.
- **Scoping to visible name.** Channel candidate `name` narrowed to `displayName || name || channelId` (dropped hidden `title` / `orgData.displayName`); dedicated test confirms a hidden title/org name is not matchable while the visible name is.
- **Server-authoritative order preserved.** `appendUniqueByKey` keeps server results first and appends only new keys; friend/group order tests pass.
- **Perf.** Conversion happens once at index-build with per-key cache reuse when `sourceText` is unchanged (10k-name test converts exactly once; rebuild reuses); keystroke matching is pure `searchText.includes(kw)` substring — no per-keystroke O(n*m) reconversion.
- **Guards.** Pinyin supplement gated on `/^[a-z]+$/i`, so existing CJK / mixed / English direct searches are not polluted; empty/null pools guarded throughout.

### 💬 Non-blocking
- The pinyin converter was swapped from `pinyin-pro` to the in-repo `getPinyin` (first-match-in-dictionary heuristic). For polyphonic characters this can yield a different reading than the context-aware library, so some names may not match a full-pinyin query. It only affects the supplemental layer (server search stays authoritative) and current tests happen to use an unambiguous name, so this is a note, not a blocker — worth a follow-up test with a polyphonic surname if this path is important.

No blocking items. Approving.

